PR #1818, opened 2016-04-09
Appveyor builds -> s3

wiredfool:
A configuration to push appveyor builds up to s3.


wiredfool:
This does push to s3 on non-pr builds.

radarhere:
I presume that after the downloads after pushed to the s3 bucket, they're publicly accessible. Should documentation be added explaining how to access them for use?

wiredfool:
There's a direct url now, but it's not discoverable.

Issue #1514, opened 2015-10-30
PIL.Image.fromarray fails with uint16 arrays


dgua:
Sorry if this isn't the right place. 

When I try PIL.Image.fromarray with a uint16 array, it fails. Looking at the source code, uint16 is not supported. What would it take to add this data type?

radarhere:
This is the right place. Welcome to the official home of Pillow, where bug reports and contributions are accepted. In this case, you're asking for an enhancement.

As with many enhancements, it's a question of hoping that someone knowledge takes an interest and is willing to figure out the change. Or if you are so inclined, you are more than welcome to put it together and then create a pull request for review.

Would it be possible for you to provide some sample code? I understand what you're asking, I'd just be interested in having a short program to experiment against. While I'm likely not the right person for this one, I would be keen in checking out the problem at hand.

dgua:
On Nov 1, 2015, at 4:55:46, Andrew Murray <notifications@github.com> wrote:

> This is the right place. Welcome to the official home of Pillow, where bug reports and contributions are accepted. In this case, you're asking for an enhancement.
> 
> As with many enhancements, it's a question of hoping that someone knowledge takes an interest and is willing to figure out the change. Or if you are so inclined, you are more than welcome to put it together and then create a pull request for review.
> 
> Would it be possible for you to provide some sample code? I understand what you're asking, I'd just be interested in having a short program to experiment against. While I'm likely not the right person for this one, I would be keen in checking out the problem at hand.
> 
> —
> Reply to this email directly or view it on GitHub.
> 


OK, here is the code (Python):

from PIL import Image
# img_data is a np.array that contains the image (dtype=int16)
im = Image.fromarray(np.flipud(img_data))  
# This works
im.save('test.tiff’)  
# Now, ImageJ only supports uint16, not int16. So:
img_data2 = (img_data+32768).astype(dtype=np.uint16) 
im2 = Image.fromarray(np.flipud(img_data))  
# This generates an error


I am using the Anaconda distribution with Python 2.7.

Davide

mairsbw:
This problem still exists. I'm working with uint16 TIFFs for a remote sensing/GIS application. I can read the uint16 data from an image just fine, but can't write it back. Very unexpected that you can't roundtrip uint16 data with Pillow's Image module.

@radarhere If you could point me in the right direction I can likely code up this missing functionality as I would think it'd be trivial with the rest of the support available in this library.

[...]

PR #1819, opened 2016-04-09
Add support for writing 16-bit greyscale images

mairsbw:
Addresses #1514.

This has been tested and found working for writing back uint16 greyscale TIFFs little-endian TIFFs. A corresponding test has been added to the Tests directory with include an original test image collected by myself and its license is CC0 (it should be noted that some of the other Image tests failed when I ran them, but it may be a path issue since I'm running them on Windows versus a library problem).


mairsbw:
Looks like Travis has a configuration issue and missing a library (sphinx_rtd_theme).

wiredfool:
These are actually failing in the tests as well:  https://travis-ci.org/python-pillow/Pillow/jobs/121929495#L2747
```
FAIL: TestNumpy.test_numpy_to_image
  File "/home/travis/build/python-pillow/Pillow/Tests/test_numpy.py", line 51, in test_numpy_to_image
    self.assertRaises(TypeError, lambda: to_image(numpy.uint16))
AssertionError: TypeError not raised by <lambda>
Ran 674 tests in 22.647s
FAILED (SKIP=24, failures=1)
```

The docs, well for some reason breaking there isn't breaking the build, and a recent change broke them.

mairsbw:
Looks like my test [successfully passed in Travis](https://travis-ci.org/python-pillow/Pillow/jobs/121929495#L2410) anyways.

Is there anything else I need to do to get an r+ on this?

PIL/Image.py
```diff
@@ -2206,6 +2206,8 @@ def fromqpixmap(im):
     # first two members of shape are set to one
     # ((1, 1), "|b1"): ("1", "1"), # broken
     ((1, 1), "|u1"): ("L", "L"),
+    ((1, 1), "<u2"): ("L", "I;16"),
+    ((1, 1), ">u2"): ("L", "I;16B"),
     ((1, 1), "|i1"): ("I", "I;8"),
```
wiredfool: These two unpackers are only taking the upper byte and throwing away the lower byte, and packing this into a single channel 8 bit image.  In the process, they're interpreting the upper byte as a uint8, which is unlikely to be correct if it's a signed int type. 

I think that the correct mode, raw_mod tuples to use would be `('I', 'I;16S')`  and `('I', 'I;16BS')`. These two promote the image to a 32 bit image, which is enough range to losslessly contain the I16S datatype.

Tests/test_file_tiff.py
```diff
@@ -450,6 +450,26 @@ def test_save_tiff_with_jpegtables(self):
...
+        # And then all it's parameters should be the same as the original.
+        # We don't just compare the images because the 'readonly' value here will be different, so
+        # we instead compare all other attributes.
+        out_im = Image.open(outfile)
+        self.assertEqual(im.mode, out_im.mode)
+        self.assertEqual(im.size, out_im.size)
+        self.assertEqual(im.info, out_im.info)
+        self.assertEqual(im.category, out_im.category)
+        self.assertEqual(im.getpalette(), out_im.getpalette())
+        self.assertEqual(im.tobytes(), out_im.tobytes())
```
wiredfool: There's a helper function, self.assert_image_equal that covers the mode, size and content of the image.

Tests/test_file_tiff.py
```diff
@@ -450,6 +450,26 @@ def test_save_tiff_with_jpegtables(self):
...
+
```
wiredfool: This test doesn't actually test the code changes, as it doesn't run through any of the numpy code. It's only testing round tripping of a saved image.

PIL/Image.py
mairsbw: The actual datatype is a uint16, so there's no dataloss with the uint8 issue you talk about.

Tests/test_file_tiff.py
mairsbw: Why does numpy need to be involved? This is a uint16 TIFF that's being resaved. This test failed previously because it both threw a TypeError saying it could save uint16 TIFFs and because the data it saved was wrong, it only saved the image as a uint8 image. This test checks for both of those things.

Tests/test_file_tiff.py
mairsbw: Looks like that'll work, great. I'll use that instead.

PIL/Image.py
mairsbw: Additionally, I believe the "I" means that it's a color image, but in this case it's not. I don't quite understand all of these special strings that Pillow uses, so maybe I'm wrong here. I do believe this code is working as shown by my testcase. You can fully roundtrip a uint16 image with no dataloss.

mairsbw:
I also don't know how to get unittest to stop complaining about unclosed files. I've added code to close the images in my test case but it still complains. Does anyone know the right way to do that?

wiredfool:
1) The code section you changed is the fromarray definitions, which are the interface to numpy. 
2) #1514 is a bug/feature request for interoperation with numpy. 
3) I is a 32 bit single channel image, L is a single channel 8 bit image, color images are RGB, CMYK, or similar.
4) The code for the fromarray changes that you added do allow for a uint16 image to be passed from numpy using the array interface to pillow, but it ends up as an 8bit image using the most significant byte, throwing out the lower byte.  This is not something that I'd want to merge. 
5) The test roundtrips a 16bit image, which we do generally support, as indicated by the test passing on master without the changes. There are some specific 16 bit formats (like signed 16bit integers) that aren't supported.


I think fundamentally, you have a different problem than you're trying to solve here.

radarhere:
With regards to unclosed files, from my experience, this occurs for many other parts of Pillow, so don't consider it a blocker for this. If you like, you can create a new issue to discuss the problem.

mairsbw:
It sounds to me like there should be the addition of a 16-bit single-channel image type, but I still don't quite understand how numpy and Pillow communicate here. If I read in an image I'd expect the array to maintain the same datatype as the file.

@wiredfool Could you take a look at the two tests I now have in this branch in d6c9ac1? One tests just the round-tripping of a TIF (which fails in Pillow 3.1, which I had installed). Additionally it adds another tests that checks creating a 16-bit TIFF from a uint16 numpy array and making sure it is read back as the same data. I don't know if that second tests should be there or in test_numpy.py or one of the test_image_*.py files.

radarhere:
Hi. I'll note here that #1984 has been merged, which replaced #1824, relating to this.

wiredfool:
I'm not sure this really applies anymore -- but if it does, feel free to rebase and do a new pr.

PR #1821, opened 2016-04-10
Added TIFF Tags

radarhere:
See http://www.awaresystems.be/imaging/tiff/tifftags/private.html


mairsbw:
This excludes a lot of private tags used by the DNG format, even though TiffTags.py seems to include others for the DNG format. Should they not be added here?

radarhere:
Such as? I tried to add all that I could see from the link that I've used as reference.

mairsbw:
Looks like I misread the diff so it's only 50741.

PR #1822, opened 2016-04-10
Removed unused imports

radarhere:


aclark4life:
Nice! The ``docs`` import removal looks safe, but I wonder if someone was using those ``__future__`` imports. In any event, can we get (or do we already have) flake8 running on the entire code base so we can systematically remove unused imports (which flake8 reports) ?

aclark4life:
Let's try to merge this in for 3.5 after a comprehensive search for all unused imports (assuming it's not too much trouble). We don't have to have them all removed by then, but I'd like to start working on it at least (unless we're already done except for these :smile:)

hugovk:
@aclark4life Yes, pep8 and pyflakes are run on [Travis CI](https://github.com/python-pillow/Pillow/blob/master/.travis.yml#L77-L82), for example: https://travis-ci.org/python-pillow/Pillow/jobs/164405898#L3855

Landscape.io also does that sort of thing, click the badge in the README, or, for example: https://landscape.io/github/python-pillow/Pillow/560/messages

radarhere:
I've rebased and changed this PR to remove some more imports. I've removed some imports that are used only by comments. The division import in the test file was originally in use, but no longer.

Oh, and there's a drop in coverage now from the fewer number of lines.

@aclark4life when you say that you wonder if someone was using the print imports, do you mean someone relying on those imports being run by Pillow for the sake of an external script? Or are you just wondering why they were there in the first place?

aclark4life:
@radarhere I mean "is there anything else going on besides the obvious removal of an unused import" and/or "will there by any unintended side effects from this action". I assume someone imported ``print`` from ``__future__`` to use the Python 3 style print function? E.g. ``print()``. But I'm not sure.

radarhere:
docs/conf.py - `shlex` was unused from when it was added in https://github.com/python-pillow/Pillow/commit/dcefb93, presumably as part of the quickstart code

Testst/test_file_metadata.py - `division` was added in https://github.com/python-pillow/Pillow/commit/974bcc0, and the relevant operations were removed in https://github.com/python-pillow/Pillow/commit/8ed2d1e

Tests/test_tiff_ifdrational.py - `print_function` has been unused since the file was added.

Tests/test_file_tiff.py - `print_function` was added in https://github.com/python-pillow/Pillow/commit/4e2de9d, and the relevant lines were removed in https://github.com/python-pillow/Pillow/commit/4e754e9

PIL/WalImageFile.py - `print_function` was added in https://github.com/python-pillow/Pillow/commit/5076c35, and the relevant lines were removed in https://github.com/python-pillow/Pillow/commit/7f00312

PIL/PyAccess.py - `print_function` has been there since the file was added, and the relevant line was removed in https://github.com/python-pillow/Pillow/commit/4e754e9

PIL/PngImagePlugin.py - `print_function` was added in https://github.com/python-pillow/Pillow/commit/5076c35, and the relevant lines were removed in https://github.com/python-pillow/Pillow/commit/4e754e9

PIL/PcxImagePlugin.py - `print_function` was added in https://github.com/python-pillow/Pillow/commit/4e2de9d, and the relevant lines were removed in https://github.com/python-pillow/Pillow/commit/4e754e9

PR #1824, opened 2016-04-11
Fix tests for Image.fromarray()

### LINKED ISSUE #1514: PIL.Image.fromarray fails with uint16 arrays

mairsbw:
This also makes explicit the existing image support for different datatypes fed to `Image.fromarray()`.

This is part of the base work to resolve #1514 and #1819.


Tests/test_numpy.py
```diff
@@ -89,6 +115,24 @@ def test_16bit(self):
...
+            a.shape = TestNumpy.TEST_IMAGE_SIZE
+            Image.fromarray(a).save(tmpfile)
+            im_test = Image.open(tmpfile)
+            im_good = Image.open(image)
+            self.assert_image_equal(im_good, im_test)
+
+        _test_image(0, "Tests/images/uint16_1_0.tif")
+        _test_image(255, "Tests/images/uint16_1_255.tif")
+        _test_image(65280, "Tests/images/uint16_1_65280.tif")
+        _test_image(65535, "Tests/images/uint16_1_65535.tif")
```
wiredfool: I think all of these can be coalesced into a single save/load of a small image with data == [0,255, 65280, 65535].

Thinking about the way that these can fail, I'm not sure if this test would catch the file being saved and reloaded as an int32, as that should be what the storage mode is.

Tests/test_numpy.py
```diff
@@ -12,6 +12,7 @@
 
 
 class TestNumpy(PillowTestCase):
+    TEST_IMAGE_SIZE = (10, 10)
```
wiredfool: This should probably be a module level global, rather than a class member. 

So all of the TestNumpy.TEST_IMAGE_SIZE can be just TEST_IMAGE_SIZE

Tests/test_numpy.py
mairsbw: The problem here is that there are a ton of stages to saving and loading images with Pillow. On loading there's the initial reading of the fileformat, the conversion to an internal array, the conversion to a proper Pillow image class. Now on saving there's the conversion of the internal array to the format specified by the image class and then writing that in an appropriate format to disk.

I've been having a lot of trouble trying to get all of these different pieces in order because there definitely seems to be some lacking support, or incorrect behavior, in various pieces of the chain. This is why I'm interested in adding tests that just improve coverage and should be correct as we'll need a bunch more than this one to test that a uint16 image can be loaded as a uint16 array and then written back out as one.

So are you suggesting that this test be dropped, replaced, or extended with other tests? As I believe the latter is appropriate and what I plan to do once I get this first round of changes done.

mairsbw: Let me add the comment that I'd like to keep the image the same repeated uint16 value tiled over a 10x10 image for consistency with other tests. I'll condense this to a single image, however.

mairsbw:
I've incorporated your comments about using a single image to test uint16 reading and writing and moving that constant to module-scope. Hopefully these tests are more amenable.

radarhere:
You may already be aware, but I'll just point out here that there are merge conflicts in this branch.

mairsbw:
@radarhere I'll fix these tests if you think they'd be worth pulling in. Otherwise, I'll just close this PR.

wiredfool:
It looks like it's a minor merge issue, and the failing test is due to not having an int64 type in pillow:
```
ERROR: TestNumpy.test_numpy_to_image
  File "/home/erics/Pillow/Tests/test_numpy.py", line 58, in test_numpy_to_image
    self.assert_image(to_image(numpy.int), "I", (10, 10))
  File "/home/erics/Pillow/Tests/test_numpy.py", line 33, in to_image
    i = Image.fromarray(a)
  File "build/bdist.linux-x86_64/egg/PIL/Image.py", line 2170, in fromarray
    raise TypeError("Cannot handle this data type")
TypeError: Cannot handle this data type
```

in this:
```
        # Check non-fixed-size integer types
        self.assertRaises(TypeError, lambda: to_image(numpy.uint))
        self.assert_image(to_image(numpy.int), "I", (10, 10))
```

```
>>> import numpy
>>> a = numpy.array([0,1,2,3],numpy.int)
>>> a.__array_interface__['typestr']
'<i8'
>>> 
```

PR #1831, opened 2016-04-13
Fix the problem with icc_profile on tiff to png conversion

jespino:
Taking care of special cases for icc_profile information to always return a bytes variable.

### DIFF
--- a/PIL/TiffImagePlugin.py
+++ b/PIL/TiffImagePlugin.py
@@ -555,6 +555,10 @@ class ImageFileDirectory_v2(collections.MutableMapping):
             if legacy_api and self.tagtype[tag] in [5, 10]:
                 values = values,
             dest[tag], = values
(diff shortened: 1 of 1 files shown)


wiredfool:
I'd have to go back and see where we've discussed this before, but my memory of this is that the core problem in this section of code is that we're selecting on the count of the metadata as reported by the spec, not the count of the metadata that we've actually found in the file that we're reading. So images that are different from the spec but may be reasonably well formed otherwise will trigger an error. 

And, if there's one true thing about images, it's that the metadata varies.

jespino:
I'm not sure if the problem is more general (I don't know Pillow well to know how works the rest of the code) but this keep the tests working and allow transformations from tiff to png, that wasn't allowed. Maybe defined in anyway the como expected behavior for some headers? for example the ICC_PROFILE (this transform the icc_profile to the format expected by PNG, but I'm not sure this is enough general).

PR #1836, opened 2016-04-17
Removed unused test helper

radarhere:
It was added as part of https://github.com/python-pillow/Pillow/pull/832/commits/4c5a5c1f0eddef0d384f18d46657c92e954b5075 on #832, but then removed as part of https://github.com/python-pillow/Pillow/pull/832/commits/a0aff1a87faa72373010d0e5c71667c7dcc34e19 on that same PR.


hugovk:
Thanks.

PR #1837, opened 2016-04-18
Added OS X support for ImageGrab grabclipboard

radarhere:


PIL/ImageGrab.py
```diff
@@ -51,11 +51,31 @@ def grab(bbox=None):
...
     if sys.platform == "darwin":
-        raise NotImplementedError("Method is not implemented on OS X")
-    debug = 0  # temporary interface
-    data = Image.core.grabclipboard(debug)
-    if isinstance(data, bytes):
-        from PIL import BmpImagePlugin
-        import io
-        return BmpImagePlugin.DibImageFile(io.BytesIO(data))
-    return data
+        f, file = tempfile.mkstemp('.jpg')
```
hugovk: Don't use builtin `file`, and maybe use more descriptive name than `f`.

Perhaps `file` -> `filename` and `f` -> `fh`.

PIL/ImageGrab.py
```diff
@@ -51,11 +51,31 @@ def grab(bbox=None):
...
+        commands = [
+            "set theFile to (open for access POSIX file \""+file+"\" with write permission)",
+            "try",
+                "write (the clipboard as JPEG picture) to theFile",
+            "end try",
+            "close access theFile"
+        ]
+        script = ["osascript"]
+        for command in commands:
+            script += ["-e", command]
```
hugovk: Could `script` be something along the lines of this, or would it lose readability/make it worse?

```python
        script = ["osascript", -"e",  
            "set theFile to (open for access POSIX file \""+file+"\" with write permission)",
            "try",
                "write (the clipboard as JPEG picture) to theFile",
            "end try",
            "close access theFile"]
```

PIL/ImageGrab.py
radarhere: Okay, I've made the change. I used `fh` and `filepath`, since to my mind, `filename` is just the base name.

PIL/ImageGrab.py
radarhere: The code is actually adding "-e" before every line of `commands`, not just the first one. So if I combined it, this would be the result, which I felt was inelegant.
```python
script = ["osascript"
            "-e","set theFile to (open for access POSIX file \""+filepath+"\" with write permission)",
            "-e","try",
                "-e","write (the clipboard as JPEG picture) to theFile",
            "-e","end try",
            "-e","close access theFile"]
```

PR #1840, opened 2016-04-19
Improve handling of transparent PNG images

### LINKED ISSUE #1838: Transparency is lost when loading PNG with transparent black color
I noticed a regression from #1239 

### Expected behavior

``info['transparency']`` in PIL stores either the palette index of the full transparent color, or the opacity values for all palette entries. 
So it's 255 if all pixels with the value 255 are transparent, but it's a byte string if the palette values can have their own opacity.

Here is a paletted PNG with full alpha support (created with pngquant):

fade-rgba-fs8

You can see that ``info['transparency']`` contains a byte string, even after #1239:

``print(repr(Image.open('fade-rgba-fs8.png').info['transparency']))``

This byte string is important. Otherwise PIL is not able to keep the full alpha channel when saving PNGs.
See here how the byte string is used in the _save method:
https://github.com/python-pillow/Pillow/blob/master/PIL/PngImagePlugin.py#L720..L724

### Triggering the bug

The "fix" from #1239 removes the transparency information when the byte string only contains nulls. But this string is still valid. It means that all palette entries are full transparent.

You can create an image like that with:
``
Image.new("RGBA", (100, 100), (0, 0, 0, 0)).convert('P').save('transp.png')
``

You can see how Pillow with #1239 removes the transparency by open and saving the image with:
``
Image.open('transp.png').save('/tmp/bug.png')
``

### Solution

I suggest to revert the change. 
According to #1238, reportlab does not handle `transparency` with byte strings and #1239 tries to fix this by letting Pillow not return a byte string in this single case. But it only fixes it this specific case (all bytes null). The image above is one example where reportlab should still raise an exception. @juztin can you confirm this?

I can create a pull request with the revert together with a few new tests for the expected ``transparency`` behavior.

olt:
This pull request resolves  #1838

### DIFF
--- a/PIL/PngImagePlugin.py
+++ b/PIL/PngImagePlugin.py
@@ -74,8 +74,7 @@ _MODES = {
 }
 
 
-_simple_palette = re.compile(b'^\xff+\x00\xff*$')
-_null_palette = re.compile(b'^\x00*$')
+_simple_palette = re.compile(b'^\xff*\x00\xff*$')
 
 # Maximum decompressed size for a iTXt or zTXt chunk.
 # Eliminates decompression bombs where compressed chunks can expand 1000x
@@ -118,7 +117,7 @@ class ChunkStream(object):
             cid = s[4:]
             pos = self.fp.tell()
             length = i32(s)
-            
+
         if not is_cid(cid):
             raise SyntaxError("broken PNG file (chunk %s)" % repr(cid))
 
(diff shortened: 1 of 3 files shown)


olt:
The failed test is an unrelated error in the docs.

radarhere:
If you would like the tests to all pass, you can rebase on the current master. Or if you're happy to leave it, you can.

olt:
I'm OK with it, but I don't know about the person who merges this. Do you need green build before merge?

wiredfool:
Don't worry about it. I'll probably be the one merging, I just need to understand the PR first.

wiredfool:
merged with #1882

PR #1826, opened 2016-04-12
Added CMYK mode for opening EPS files

radarhere:
Fixes #1816

PR #1846, opened 2016-04-19
EPS + CMYK

wiredfool:
includes #1826 

Added a test to determine if we were getting reasonable raster data out of the test image. 

Oddly enough, it comes back as `RGB` from ghostscript, so there's some conversion going on in there that's not obvious. 

This still needs some docs, as things that aren't obvious are bound to bite us. Such as, what is the output format that we'd expect. What happens if it's b/w? Are there input formats that don't convert?

Also, there's this: 
```
$ identify -verbose Tests/images/pil_sample_cmyk.eps
Image: Tests/images/pil_sample_cmyk.eps
  Format: PS (PostScript)
  Class: DirectClass
  Geometry: 100x100+0+0
  Resolution: 72x72
  Print size: 1.38889x1.38889
  Units: Undefined
  Type: TrueColorMatte
  Endianess: Undefined
  Colorspace: RGB
  Depth: 16/8-bit
...
```

Which makes me wonder if this is actually a good test image.


Tests/test_file_eps.py
```diff
@@ -57,6 +57,22 @@ def test_invalid_file(self):
...
+        self.assertEqual(cmyk_image.size, (100, 100))
+        self.assertEqual(cmyk_image.format, "EPS")
+        
+        cmyk_image.load()
+        self.assertEqual(cmyk_image.mode, "RGB")
+
+        if 'jpeg_decoder' in dir(Image.core):
+            target = Image.open('Tests/images/pil_sample_rgb.jpg')
+            self.assert_image_similar(cmyk_image, target, 10)
+            cmyk_image.show()
```
radarhere: You're probably aware, but just to note - this debugging line is still there.

wiredfool: That's fixed now.

PR #1847, opened 2016-04-19
Revert 1813, plus second commit with tweaks. 

wiredfool:
This reverts #1813, reapplies the first commit. 

The second sets the returned image to BGRX mode, adds comments and removes the test with alpha.


wiredfool:
K, transparency nuked.

PR #1849, opened 2016-04-20
Common seek frame position check

radarhere:
When seeking, some of the multiframe image formats check that the destination frame is not the current frame, or that the destination frame is not out of range.

This moves those checks into ImageFile, also making the behaviour more consistent between formats.

### DIFF
--- a/PIL/ImImagePlugin.py
+++ b/PIL/ImImagePlugin.py
@@ -270,11 +270,7 @@ class ImImageFile(ImageFile.ImageFile):
         return self.info[FRAMES] > 1
 
     def seek(self, frame):
-
-        if frame < 0 or frame >= self.info[FRAMES]:
-            raise EOFError("seek outside sequence")
-
-        if self.frame == frame:
+        if not self._seek_check(frame):
             return
 
         self.frame = frame
@@ -292,7 +288,6 @@ class ImImageFile(ImageFile.ImageFile):
         self.tile = [("raw", (0, 0)+self.size, offs, (self.rawmode, 0, -1))]
 
     def tell(self):
-
         return self.frame
 
 #
--- a/PIL/MpoImagePlugin.py
+++ b/PIL/MpoImagePlugin.py
@@ -72,14 +72,13 @@ class MpoImageFile(JpegImagePlugin.JpegImageFile):
         return self.__framecount > 1
 
     def seek(self, frame):
-        if frame < 0 or frame >= self.__framecount:
-            raise EOFError("no more images in MPO file")
-        else:
-            self.fp = self.__fp
-            self.offset = self.__mpoffsets[frame]
-            self.tile = [
-                ("jpeg", (0, 0) + self.size, self.offset, (self.mode, ""))
-            ]
+        if not self._seek_check(frame):
+            return
+        self.fp = self.__fp
+        self.offset = self.__mpoffsets[frame]
+        self.tile = [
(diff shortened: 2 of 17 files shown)


PIL/SpiderImagePlugin.py
```diff
@@ -174,8 +174,8 @@ def tell(self):
     def seek(self, frame):
         if self.istack == 0:
             raise EOFError("attempt to seek in a non-stack file")
-        if frame >= self._nimages:
-            raise EOFError("attempt to seek past end of file")
+        if not self._seek_check(frame):
+            return
```
wiredfool: This was checking self._nimages, the common code checks self._n_frames, and this code isn't checked in the tests.

PIL/MpoImagePlugin.py
```diff
@@ -71,14 +71,13 @@ def is_animated(self):
...
-        if frame < 0 or frame >= self.__framecount:
-            raise EOFError("no more images in MPO file")
-        else:
-            self.fp = self.__fp
-            self.offset = self.__mpoffsets[frame]
-            self.tile = [
-                ("jpeg", (0, 0) + self.size, self.offset, (self.mode, ""))
-            ]
+        if not self._seek_check(frame):
+            return
```
wiredfool: The seek check here never fails in the testing. This one is checking __framecount, the core is checking n_frames

PIL/ImImagePlugin.py
```diff
@@ -269,11 +269,7 @@ def is_animated(self):
         return self.info[FRAMES] > 1
 
     def seek(self, frame):
-
-        if frame < 0 or frame >= self.info[FRAMES]:
-            raise EOFError("seek outside sequence")
-
-        if self.frame == frame:
+        if not self._seek_check(frame):
```
wiredfool: This one never passes the seek check in the tests.

PIL/MpoImagePlugin.py
radarhere: The n_frames property returns __framecount, so they are equivalent.

PIL/SpiderImagePlugin.py
radarhere: I'm going to presume that you meant 'self.n_frames', rather than 'self._n_frames', making this the same comment as you made on the MpoImagePlugin - if this is not the case, let me know. If it is the case, then the n_frames property method returns self._nimages, making them equivalent.

radarhere: This code isn't checked in the tests because there isn't a non-stack SPIDER image to test it with.

PIL/ImImagePlugin.py
radarhere: I've made a new commit to change the EOFError tests. See what you think.

aclark4life:
So we need a non-stack SPIDER image added for testing?

radarhere:
That would be ideal, yes - also just to inherently check those lines, not just for this PR.

PIL/TiffImagePlugin.py
```diff
@@ -959,8 +959,10 @@ def is_animated(self):
         return self._is_animated
 
     def seek(self, frame):
+        if not self._seek_check(frame):
+            return
         "Select a given frame as current image"
```
wiredfool: This was formerly a doc string, now it's a random string comment in the code.

radarhere: Okay, I've updated the commits to move it back to the start of the method.

PR #1856, opened 2016-04-22
Combined duplicate code in ImageTk

radarhere:


PIL/ImageTk.py
```diff
@@ -52,6 +52,18 @@ def _pilbitmap_check():
...
+        source = kw.pop("file")
+    except KeyError:
+        try:
+            data = kw.pop("data")
+        except KeyError:
+            return
+        from io import BytesIO
+        source = BytesIO(data)
+    return Image.open(source)
+
```
wiredfool: 1) This isn't tested in coveralls, other than it doesn't cause an import error
2) I don't think that this nested exception structure is any clearer than the if/elif construction in the original.  While the European Association of Fish Pathologists (first google hit, really) might prefer exceptions, Flat is better than nested. The import should be hoisted out, and I'm not sure that one Image.open is preferable to short circuit returns.

radarhere: Okay, I've refactored the code, and added a test for the new function.

PR #1860, opened 2016-04-25
Added register_extensions method

radarhere:
Rather than having multiple calls to `register_extension` from the same image plugins, I suggest a `register_extensions` method.


aclark4life:
Nice! Probably need a test or two then can merge.

radarhere:
Okay, I've added a test. See what people think.

radarhere:
Looks like the drop in coverage is simply due to the removed lines of code.

PR #1861, opened 2016-04-25
setup.py: add option to disable platform guessing

angeloc:
When cross-compiling (ex on Buildroot) platform guessing is not needed
cause the environment is correctly built externally.
This patch adds an option to disable platform guessing on Linux.


wiredfool:
See #1875

PR #1865, opened 2016-04-27
Image add support fromarray LA mode

pierriko:
when using grayscale image with alpha channel, at the moment I have to save them as:
```python
def save(path, data, quality=90):
    if len(data.shape) == 3 and data.shape[2] == 2:
        # Pillow "Cannot handle this data type" mode=LA
        img = Image.fromarray(data[:,:,0])
        alp = Image.fromarray(data[:,:,1])
        img.putalpha(alp)
    else:
        img = Image.fromarray(data)
    # in case of JPEG or WebP, set quality to 90%, else this option is ignored
    img.save(path, quality=quality)
```
(see: https://github.com/pierriko/atlaas/blob/master/python/atlaas/helpers/image.py )

This patch add support for mode LA in `Image.fromarray` and fix:
```
>>> img = np.zeros((40,40,2), 'uint8')
>>> Image.fromarray(img)
  File "/usr/lib/python2.7/dist-packages/PIL/Image.py", line 1929, in fromarray
    raise TypeError("Cannot handle this data type")
TypeError: Cannot handle this data type
```


radarhere:
Hi. Thanks for this. If you're able to add a test as well, that would be appreciated.

pierriko:
Hi !

Let me know if the test is ok with you.

Cheers,
Pierrick

radarhere:
I was imagining something that fits in with the existing tests for this - https://github.com/python-pillow/Pillow/blob/master/Tests/test_image_array.py.

Also, just to set expectations - I'm here to offer helpful hints. Someone else will ultimately review this PR for inclusion.

wiredfool:
see #1871

PR #1867, opened 2016-04-28
Reduce tests memory consumption

homm:
Currently tests consume 514 megabytes. This reduces this number to 16.


wiredfool:
Unfortunately, that test then passes on unpatched versions of the code. (e..g, 3.1.0)  The overflow is in the multiplication of both dimensions. 

If 512 megs is too much, we can move that behind a `check_*` name.

PR #1870, opened 2016-04-29
Updated Sphinx project description

radarhere:
See http://www.sphinx-doc.org/en/stable/config.html#confval-texinfo_documents for documentation of this code.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -272,7 +272,7 @@ man_pages = [
 #  dir menu entry, description, category)
 texinfo_documents = [
   (master_doc, 'PillowPILFork', u'Pillow (PIL Fork) Documentation',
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -272,7 +272,7 @@
 #  dir menu entry, description, category)
 texinfo_documents = [
   (master_doc, 'PillowPILFork', u'Pillow (PIL Fork) Documentation',
-   author, 'PillowPILFork', 'One line description of project.',
+   author, 'PillowPILFork', 'Pillow is the friendly PIL fork by `Alex Clark and Contributors.',
```
hugovk: The ` isn't needed before Alex.

radarhere: Thanks, good catch. I've updated the commit.

PR #1875, opened 2016-05-01
Disable-platform-guessing in setup.py

wiredfool:
Includes a bit of a refactor of #1861 + docs.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -202,7 +202,12 @@ Build Options
   the libraries are not found. Webpmux (WebP metadata) relies on WebP
   support. Tcl and Tk also must be used together.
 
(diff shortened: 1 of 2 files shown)


wiredfool:
@angeloc Could you check that this still works for you.

angeloc:
@wiredfool : Tested on Buildroot, it works like a charm!

Please merge!

angeloc:
@wiredfool: I would really have python-pillow included in buildroot, is there any possibility that a minor release will be released with this fix included?
Thanks!

wiredfool:
Unlikely before the scheduled release at the beginning of July, unless we have another issue that it can piggyback on. We generally only do point releases for security bugs or regressions, and this doesn't hit that criteria.

PR #1879, opened 2016-05-03
Replaced os.system with subprocess.call

radarhere:
Removes the theoretical need to escape the path argument.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,6 +13,7 @@ import platform as plat
 import re
 import struct
 import sys
+import subprocess
 
 from distutils.command.build_ext import build_ext
 from distutils import sysconfig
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -763,9 +763,12 @@ def add_multiarch_paths(self):
         tmpfile = os.path.join(self.build_temp, 'multiarch')
         if not os.path.exists(self.build_temp):
             os.makedirs(self.build_temp)
-        ret = os.system(
-            'dpkg-architecture -qDEB_HOST_MULTIARCH > %s 2> /dev/null' %
-            tmpfile)
+        devnull = open(os.devnull, 'wb')
```
wiredfool: I don't think there's really a point wiring up stderr to /dev/null, as it's just as effective to ignore it by not passing in a pipe to the stderr parameter.

radarhere: Okay, I've updated the commit.

PR #1881, opened 2016-05-03
Fixed point integer resample

homm:
This change speeds up resampling up to 60% and eliminates ugly `i2f` hack.

Source    | Operation               | Filter  | Before | After 
----------|-------------------------|---------|--------|-------
7712×4352 | **Resize to 16x16**     | Bilinear| 210    | 330   
          |                         | Bicubic | 109    | 177   
          |                         | Lanczos | 73.5   | 120   
          | **Resize to 320x180**   | Bilinear| 154    | 241   
          |                         | Bicubic | 88.6   | 139
          |                         | Lanczos | 60.4   | 96.5   
          | **Resize to 2048x1155** | Bilinear| 83.9   | 105   
          |                         | Bicubic | 55.3   | 75.1   
          |                         | Lanczos | 38.9   | 57.3

I'll add tests for other modes and maybe return old implementation for `IMAGING_TYPE_INT32` and `IMAGING_TYPE_FLOAT32`.

### DIFF
--- a/libImaging/Resample.c
+++ b/libImaging/Resample.c
@@ -7,11 +7,11 @@
 
 
 struct filter {
-    float (*filter)(float x);
-    float support;
+    double (*filter)(double x);
+    double support;
 };
 
-static inline float sinc_filter(float x)
+static inline double sinc_filter(double x)
 {
     if (x == 0.0)
         return 1.0;
@@ -19,7 +19,7 @@ static inline float sinc_filter(float x)
     return sin(x) / x;
 }
 
-static inline float lanczos_filter(float x)
+static inline double lanczos_filter(double x)
 {
     /* truncated sinc */
     if (-3.0 <= x && x < 3.0)
@@ -27,7 +27,7 @@ static inline float lanczos_filter(float x)
     return 0.0;
 }
 
-static inline float bilinear_filter(float x)
+static inline double bilinear_filter(double x)
 {
     if (x < 0.0)
         x = -x;
@@ -36,7 +36,7 @@ static inline float bilinear_filter(float x)
     return 0.0;
 }
 
-static inline float bicubic_filter(float x)
+static inline double bicubic_filter(double x)
 {
     /* https://en.wikipedia.org/wiki/Bicubic_interpolation#Bicubic_convolution_algorithm */
 #define a -0.5
@@ -83,7 +83,7 @@ ImagingPrecompute(int inSize, int outSize, struct filter *filterp,
     double *kk, *k;
 
     /* prepare for horizontal stretch */
-    filterscale = scale = (float) inSize / outSize;
+    filterscale = scale = (double) inSize / outSize;
     if (filterscale < 1.0) {
         filterscale = 1.0;
     }
--- a/libImaging/Resample.c
+++ b/libImaging/Resample.c
@@ -318,10 +318,10 @@ ImagingResampleHorizontal_32bpc(Imaging imIn, int xsize, struct filter *filterp)
 
     ImagingSectionEnter(&cookie);
     /* horizontal stretch */
-    for (yy = 0; yy < imOut->ysize; yy++) {
-        switch(imIn->type) {
-            case IMAGING_TYPE_INT32:
-                /* 32-bit integer */
+    
+    switch(imIn->type) {
+        case IMAGING_TYPE_INT32:
+            for (yy = 0; yy < imOut->ysize; yy++) {
                 for (xx = 0; xx < xsize; xx++) {
                     xmin = xbounds[xx * 2 + 0];
                     xmax = xbounds[xx * 2 + 1];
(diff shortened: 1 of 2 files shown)


wiredfool:
What are the units on the performance measurement?

homm:
Megapixels of source image per second. This is testing script: https://gist.github.com/homm/f9b8d8a84a57a7e51f9c2a5828e40e63

homm:
I split `ImagingResampleHorizontal` to two versions: for 8 bits per channel with `int` accumulators and for 32 bits per channel with `double` accumulator. I have tried to reduce code duplication and make separate `ImagingPrecompute` function.

New tests really good, because they check the consistency of result after manipulations: there should not be any shifts or dithering. And just for information, none of new tests are passed on master:

```
FAIL: CoreResampleConsistencyTest.test_32f
AssertionError: 1.00000011921 != 1.0 for pixel (0, 1)
AssertionError: nan != 3.40282306074e+38 for pixel (0, 0)
AssertionError: 1.17549351004e-38 != 1.17549393043e-38 for pixel (0, 0)
AssertionError: 1.19209289551e-07 != 1.19209303762e-07 for pixel (0, 0)

FAIL: CoreResampleConsistencyTest.test_32i
AssertionError: 11 != 12 for pixel (0, 0)
AssertionError: -2147483392 != 2147483647 for pixel (0, 0)
AssertionError: -11 != -12 for pixel (0, 0)
AssertionError: -2147483392 != -2147483648 for pixel (0, 0)
```

With the latest changes (separate `ImagingPrecompute`) new implementation is about 1% slower **than initial** for 8bpc and about 10% slower **than master** for 32bpc: this is a fee for the accuracy. Also, the bug with GCC 4.8 and below is still relevant for `I` mode, but I think this is the time then we can remove workaround: users who really faced with this problem can update compiler.

Could anyone help me with AppVeyor windows builds? Error message `command 'link.exe' failed with exit status 1120` doesn't make any sense and I don't see any other errors.

wiredfool:
I'll see what I can do on the windows builds. Probably a compile error, and for some reason we're not getting the traces.  The usual compile error is from a declaration not being at the beginning of a block, but I'm not seeing that from a quick scan of the diff.

wiredfool:
Error appears to be this: https://ci.appveyor.com/project/wiredfool/pillow/build/3.3.pre.290/job/n525alym7ebxiips#L1369
```
Resample.obj : error LNK2019: unresolved external symbol lround referenced in function ImagingResampleHorizontal_32bpc
build\lib.win-amd64-2.7\PIL\_imaging.pyd : fatal error LNK1120: 1 unresolved externals
 ```

**Comment by wiredfool** [comment]
If you rebase, you'll get the appveyor changes.

**Comment by homm** [comment]
> unresolved external symbol lround

ok, I'll check what we can use in Visual C

**Comment by homm** [comment]
LOL, new problems with windows: build hangs up on `test-installed.py`.

I'm ok with all other changes.

**Comment by homm** [comment]
@wiredfool Any thoughts about win buids?

**Comment by wiredfool** [comment]
I'll take a look, bot probably not tonight.

**Comment by wiredfool** [comment]
Looks like it's this test: https://github.com/python-pillow/Pillow/blob/master/Tests/test_image_filter.py#L39 that is hanging.

See: https://ci.appveyor.com/project/wiredfool/pillow/build/3.3.pre.295/job/wq9ms96g7w1amjvy

The code to change the way the tests run is in https://github.com/wiredfool/Pillow/commit/55b2c3bca5be73dedf71787e245aea4cef60b4d6, which I'm planning to pr/commit shortly.

**Comment by homm** [comment]
I found out that the problem is in replacing `malloc` with `calloc`. I have no idea why is it happens because `calloc` is already used in Pillow in many places. Moreover, as I understand, `TestImageFilter.test_crash` doesn't use resampling and don't call the code with `calloc` from this PR.

**Comment by homm** [comment]
Returning `malloc` makes builds green again. I don't understand anything )

**Comment by wiredfool** [comment]
@homm, I think I've got it:  https://github.com/wiredfool/Pillow/commit/c43a407d7325a8873f2b454385de2a5a076b5855

SIZE_MAX is not the constant we're looking for, in this case it's INT_MAX. With that change, calloc tests pass on windows.

File: libImaging/Resample.c [hunk 1]
```diff
@@ -65,125 +50,128 @@ static inline float bicubic_filter(float x)
...
+    int *k, *kk;
+    double *prekk;
+
+
+    kmax = ImagingPrecompute(imIn->xsize, xsize, filterp, &xbounds, &prekk);
+    if ( ! kmax) {
+        return (Imaging) ImagingError_MemoryError();
+    }
+    
+    kk = malloc(xsize * kmax * sizeof(int));
```
wiredfool: Need to check this against INT_MAX for overflow.  Also probably convert to calloc.

homm:
@wiredfool Done. Also rebased against master to pass the latest resampling accuracy tests.

PR #1768, opened 2016-03-14
adds option to ignore TCL/Tk framework installs on OS X

ddale:
Fixes issue #950. Pillow's setup.py file incorrectly assumes if you are on OS X, you must link with a TCL/TK framework installation. Some OS X python installations (Anaconda, for example) provide their own TCL/TK installation, which results in segfaults when Pillow tries to load _imagingtk (usually observed with a call to PhotoImage). A simple mechanism is needed to instruct Pillow's setup.py file to not use the the TCL/TK Frameworks installations on OS X. This pull request provides an environment variable IGNORE_OSX_TCLTK_FRAMEWORKS which, if defined, allows setup.py to build against non-frameworks installations of TCL/TK.

PR #1883, opened 2016-05-03
Alternative to PR #1768

wiredfool:
Same effect as #1768, but using a command line flag to match the other disable options. 

Also includes docs.


jakirkham:
This is great. Thanks @wiredfool.

wiredfool:
I'd like anyone who's interested in the "Which tk library to use" saga to check out PR #1932 and let me know if it works for you.

Issue #554, opened 2014-03-18
Use libimagequant for palette + alpha quantization


pornel:
Octtree seems to be the only method in Pillow that supports palette with alpha:

```python
Image.open("in.png").convert("RGBA").quantize(method=2).save("out.png")
```

However, the result is heavily posterized:

![out](https://f.cloud.github.com/assets/72159/2446741/c0aad3c0-ae8a-11e3-967c-d85b7753d3d8.png)

It would be great if you could [use pngquant's algorithm instead](http://pngquant.org/lib/) (it's available as an embeddable BSD-licensed C library):

![in-or8](https://f.cloud.github.com/assets/72159/2446746/d6e221de-ae8a-11e3-830a-8ee2ef0c71a7.png)

Both images are 256-color without dithering, but pngquant doesn't throw away any bits of precision. ([source image](https://f.cloud.github.com/assets/72159/2446751/f788ab06-ae8a-11e3-9501-c52e65b77421.png))

rr-:
I'm interested in this as well. I am willing to work on this, but I'm not sure where to begin with pngquant integration. For example, there seems to be no `install` target in libpngquant's `Makefile`, so I guess integrating it won't be as simple as doing `_find_include_file` inside `setup.py`?

wiredfool:
No, it's not going to be that simple. It's going to take integrating at the C level, likely in `_imaging.c` or one of the quantize c files.

rr-:
Clarification: I referred to wiring the library in terms of integrating it with the build process etc., rather than which C files to modify - in the latter case, I think I've got rough idea what to do (i.e. to add a new `method` inside `ImagingQuantize` function within `Quant.c`).

wiredfool:
Ah, then you'd need to know what library and header files are necessary, looks like `libimagequant.h` and `libimagequant` as a shared library. You'd need to check for those files using `_find_include_file` and `_find_library_file`. Then add a build flag and you're good to go. 

Looking at the windows side through, it's probably not going to be usable on anything but python 3.5, due to the compiler limitations there. That's ok by me, as this is an enhancement.

rr-:
[...]

PR #1889, opened 2016-05-05
Add libimagequant support for quantize()

rr-:
Implements issue #554

Manual test results:

```
Image.open("in.png").convert("RGBA").quantize(method=2).save("out.png")
```

> out

```
Image.open("in.png").convert("RGBA").quantize(method=3).save("out.png")
```

> out

Tested manually: `L`, `P`, `RGB` and `RGBA` mode. Tested transparent and opaque images.
Not tested animated GIFs but they should hardly matter. Not implemented the above tests in the code - the automated test I've added is based on existing quantization tests, which aren't very exhaustive.

---

I haven't done anything special to make custom palettes work, but:

1. There's no dedicated support for them in libimagequant - see: https://github.com/pornel/pngquant/issues/194
- I don't think I need to do anything, actually: everything seems to be dealt with inside `Image.py`, in quantization algorithm agnostic way.

---

Not sure how to integrate things with Travis, so I haven't touched that.

---

One more thing: if I put `libimagequant.so.0` in `/usr/local/lib`, `setup.py` will happily build and discover `libimagequant`, but Python itself will fail to find it and I'll need to `export LD_LIBRARY_PATH=/usr/local/lib/:$LD_LIBRARY_PATH` for it to work. Problem's gone if I move the library to `/usr/lib`. Not sure if it's an issue with my change, or just the way my environment works.

### DIFF
--- a/depends/install_imagequant.sh
+++ b/depends/install_imagequant.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+# install libimagequant
+
+git clone https://github.com/pornel/pngquant
+
+pushd pngquant
+
+make -C lib shared
+sudo cp lib/libimagequant.so* /usr/lib/
+sudo cp lib/libimagequant.h /usr/include/
+
+popd
--- a/depends/install_imagequant.sh
+++ b/depends/install_imagequant.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 # install libimagequant
 
-git clone https://github.com/pornel/pngquant
+git clone -b 2.6.0 https://github.com/pornel/pngquant
 
 pushd pngquant
 
--- a/libImaging/Quant.c
+++ b/libImaging/Quant.c
@@ -27,6 +27,7 @@
 
 #include "QuantTypes.h"
 #include "QuantOctree.h"
+#include "QuantPngQuant.h"
 #include "QuantHash.h"
 #include "QuantHeap.h"
 
@@ -1483,7 +1484,7 @@ ImagingQuantize(Imaging im, int colors, int mode, int kmeans)
         strcmp(im->mode, "RGB") != 0 && strcmp(im->mode, "RGBA") !=0)
         return ImagingError_ModeError();
 
-    /* only octree supports RGBA */
+    /* only octree and imagequant supports RGBA */
     if (!strcmp(im->mode, "RGBA") && mode != 2 && mode != 3)
        return ImagingError_ModeError();
 
--- a/libImaging/Quant.c
+++ b/libImaging/Quant.c
@@ -1484,7 +1484,7 @@ ImagingQuantize(Imaging im, int colors, int mode, int kmeans)
         return ImagingError_ModeError();
 
     /* only octree supports RGBA */
-    if (!strcmp(im->mode, "RGBA") && mode != 2)
+    if (!strcmp(im->mode, "RGBA") && mode != 2 && mode != 3)
        return ImagingError_ModeError();
 
     p = malloc(sizeof(Pixel) * im->xsize * im->ysize);
@@ -1503,8 +1503,10 @@ ImagingQuantize(Imaging im, int colors, int mode, int kmeans)
            should be done by a simple copy... */
 
         for (i = y = 0; y < im->ysize; y++)
-            for (x = 0; x < im->xsize; x++, i++)
+            for (x = 0; x < im->xsize; x++, i++) {
                 p[i].c.r = p[i].c.g = p[i].c.b = im->image8[y][x];
+                p[i].c.a = 255;
+            }
 
     } else if (!strcmp(im->mode, "P")) {
         /* palette */
@@ -1517,6 +1519,7 @@ ImagingQuantize(Imaging im, int colors, int mode, int kmeans)
                 p[i].c.r = pp[v*4+0];
                 p[i].c.g = pp[v*4+1];
                 p[i].c.b = pp[v*4+2];
+                p[i].c.a = pp[v*4+3];
             }
 
     } else if (!strcmp(im->mode, "RGB") || !strcmp(im->mode, "RGBA")) {
@@ -1565,6 +1568,21 @@ ImagingQuantize(Imaging im, int colors, int mode, int kmeans)
         result = quantize_octree(
             p,
             im->xsize*im->ysize,
+            colors,
+            &palette,
+            &paletteLength,
+            &newData,
+            withAlpha
+            );
+        break;
+    case 3:
+        if (!strcmp(im->mode, "RGBA")) {
+            withAlpha = 1;
+        }
+        result = quantize_pngquant(
+            p,
+            im->xsize,
+            im->ysize,
             colors,
             &palette,
             &paletteLength,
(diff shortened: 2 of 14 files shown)


libImaging/QuantPngQuant.c
```diff
@@ -0,0 +1,141 @@
...
+    uint32_t width,
+    uint32_t height,
+    uint32_t quantPixels,
+    Pixel **palette,
+    uint32_t *paletteLength,
+    uint32_t **quantizedPixels,
+    int withAlpha)
+{
+    return 0;
+}
```
rr-: I put this signature here to avoid linker errors in case `libimagequant` is *not* found. I didn't want to pollute `Quant.c` with `#ifdef`. Plus it's good to yield an error (no matter how obscure) rather than pretend a method doesn't exist altogether (like it happens with `n_frames` and `is_animated` attributes, that need to be retrieved with `getattr()`...)

libImaging/Quant.c
```diff
@@ -1503,8 +1503,10 @@ ImagingQuantize(Imaging im, int colors, int mode, int kmeans)
            should be done by a simple copy... */
 
         for (i = y = 0; y < im->ysize; y++)
-            for (x = 0; x < im->xsize; x++, i++)
+            for (x = 0; x < im->xsize; x++, i++) {
                 p[i].c.r = p[i].c.g = p[i].c.b = im->image8[y][x];
+                p[i].c.a = 255;
```
rr-: Without this, `libimagequant` acts very weird. Come to think of it, not having this line is actually a bug:

- Pillow supports 32-bit palettes
- We run this method with an image using a 32-bit palette
- We send this image to a quantizer that supports such 32-bit palettes
- ...it receives random garbage in the alpha channel

libImaging/Quant.c
```diff
@@ -1517,6 +1519,7 @@ ImagingQuantize(Imaging im, int colors, int mode, int kmeans)
                 p[i].c.r = pp[v*4+0];
                 p[i].c.g = pp[v*4+1];
                 p[i].c.b = pp[v*4+2];
+                p[i].c.a = pp[v*4+3];
```
rr-: Same story as above.

rr-:
Please note that the Travis build has failed because libpngquant is not installed. I'm not sure how can I test changes to Travis scripts.

depends/install_imagequant.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+# install libimagequant
+
+git clone https://github.com/pornel/pngquant
```
wiredfool: We should probably be tracking a version tag here rather than master.

wiredfool:
+1 at this point. Thanks for your work on this. 

We're going to need documentation about the new dependency, and add it to the methods mentioned in the quantization docs. (assuming they exist). 

I'd also like to see tests that show a sanity check on the output images, probably using the hopper.png image.

rr-:
The tests are included [here](https://github.com/python-pillow/Pillow/pull/1889/files#diff-17a8d6b02c5b9414bade1a4d0c7d2de6R18).

Please note that the Travis build (probably still) fails, but I don't fully understand why.

wiredfool:
Remaining test failure was a known issue with the J2k plugin hanging.

rr-:
So what do we do from now on?

wiredfool:
I need to review, but I'm kind of buried right now.

libImaging/QuantPngQuant.c
```diff
@@ -0,0 +1,141 @@
+/* Copyright (c) 2010 Oliver Tonnhofer <olt@bogosoft.com>, Omniscale
+//
```
wiredfool: Where's the original of this file from?

rr-: There's no original, I copy pasted the header from another file, probably one of `quant*.c`. Sorry.

libImaging/QuantPngQuant.c
```diff
@@ -0,0 +1,125 @@
...
+#include "QuantPngQuant.h"
+
+#ifdef HAVE_LIBIMAGEQUANT
+#include "libimagequant.h"
+
+int
+quantize_pngquant(
+    Pixel *pixelData,
+    uint32_t width,
+    uint32_t height,
```
wiredfool: Height and Width should be signed ints, as that's what they are in the imaging structure and in the libpngquant calls.

wiredfool:
More notes from review:

1) I'd like to make sure that the output image is within some epsilon of the original. 
2) Running one w/alpha through would be good as well, and it's probably worth running `P` and `L`.
3) Not specific to your code, but none of the quantization will work for more than a 2Gpx image due to integer overflow. Additionally, In storage, we're only allocating blocks up to 4MB, so anything more than that is row oriented storage. Here though, we're allocating in 4*w*h in 2 places per conversion. 
4) Also not specific, there's a lot of copies of the image floating around, and I'm unclear why it's coming back as a uint32_t* when it's a palette (char *) image at that point. 
5) Need to add 2 things to the docs: 
  * The licenses aren't strictly compatible, so distributing binaries of the combination requires care and licence review
  * The state of windows compiler support in the library means that it's likely only compatible with python 3.5, since it requires a later version of the compiler than python is compiled with.

rr-:
1. (and 2) What method of comparison do you suggest?
    1. We'll hit pretty big epsilons with individual pixels
    2. Average color on the whole image doesn't tell us much
    3. Dividing the image into blocks and sampling average color of such blocks will result in lengthy test code (I'd go with this approach if we don't have better ideas)
3. We'll hit OOM much sooner by trying to allocate `MAX_INT*MAX_INT` pixels
4. I used `uint32_t*` that was given to me by `ImagingQuantize()` and proceeded like other methods (for ex. `quantize_octree`), though I agree it doesn't make much sense to reallocate it like this
5. Could you please show specific places that I could change?

I changed the signature and fixed the header (used `Incremental.c` as a reference).

wiredfool:
> (ad 1 and 2) What method of comparison do you suggest?
Something like: `self.assert_image_similar(converted, orig, epsilon)`

That gives an abs average color difference per pixel. Put in something small, see what the number is, verify that the image looks good, and then set the epsilon to just over that previous error. It will tell us a general level of error, and if something goes really off, that error will skyrocket and trigger an error.

> (ad 3) We'll hit OOM much sooner by trying to allocate MAX_INT*MAX_INT pixels, or I misunderstood something. (unless there's support for some kind of "pipe" images)

Errors will happen on x*y>2^31, which is 2 gigapixel (or less, for when we're allocating 4 bytes/pixel).  That's only 46k pixels on a side, which is not out of the realm of reason. Don't worry about it here though, as that's just for future. I just didn't want to forget about it. 

> (ad 4) I used uint32_t* that was given to me by ImagingQuantize() and proceeded like other methods (for ex. quantize_octree), though I agree it doesn't make much sense to reallocate it like this. I could change Quant.c to expose uint8_t* directly, but I think it's better to file a separate PR for this.

Again, that's probably for a later PR and cleanup. 

> (ad 5) Could you please show specific places that I could change?

It's probably worth pulling out a section in the Installation docs. I can probably write something for that and include it.

rr-:
I've added assertions for image similarity, including the existing tests. (Sorry for the delay.) It seems all that's left to do are the tweaks to the documentation.

rr-:
:3

PR #1891, opened 2016-05-06
Avoid unnecessary Image operations

radarhere:
- Avoid unnecessarily loading an image twice
- In Image's convert method, if image is already in the destination mode, do not convert

### DIFF
--- a/PIL/Image.py
+++ b/PIL/Image.py
@@ -875,8 +875,8 @@ class Image(object):
                     mode = "RGB"
             else:
                 return self.copy()
-
-        self.load()
+        else:
+            self.load()
 
         if matrix:
             # matrix conversion
@@ -1062,10 +1062,10 @@ class Image(object):
         :returns: An :py:class:`~PIL.Image.Image` object.
         """
 
-        self.load()
         if box is None:
             return self.copy()
 
+        self.load()
         return self._new(self._crop(self.im, box))
 
     def _crop(self, im, box):
@@ -1390,9 +1390,10 @@ class Image(object):
                     im = im.convert(self.mode)
             im = im.im
 
-        self.load()
         if self.readonly:
             self._copy()
+        else:
+            self.load()
 
         if mask:
             mask.load()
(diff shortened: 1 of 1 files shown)


PIL/Image.py
```diff
@@ -1589,11 +1591,10 @@ def putpixel(self, xy, value):
         :param value: The pixel value.
         """
 
-        self.load()
         if self.readonly:
             self._copy()
             self.pyaccess = None
```
homm: `self._copy()` already sets `pyaccess` to None

homm:
```python
def _mutable(self):
    if self.readonly:
        self._copy()
    else:
        self.load()
```

homm:
First of all, `im.load()` is always noop when called in second and subsequent times. It's very cheap call. So, it's nice to avoid the double calling for some operations, but it is not necessary. I would agree, that we can avoid calling nearby `self.readonly`, but as of `Image.convert()` method, it is too complicated. I'd prefer to have one call in the beginning. Something like this:

```python
def convert(...):
    self.load()

    if not mode and self.mode == "P":
        # determine default mode
        if self.palette:
            mode = self.palette.mode
        else:
            mode = "RGB"

    if not mode or (mode == self.mode and matrix is None):
        return self.copy()
        
...
```

PIL/Image.py
radarhere: I have created a separate PR for this, #2714

radarhere:
Okay, I've updated the commits. See what you think.

PIL/Image.py
```diff
@@ -592,6 +592,12 @@ def _copy(self):
         self.pyaccess = None
         self.readonly = 0
 
+    def _mutable(self):
+        if self.readonly:
```
wiredfool: I don't like the name of this, as it shows in the calling code. The bare call to `self._mutable` while clearly having a side effect, doesn't clearly explain what's going on. 

I'd be happier with something like `_ensure_mutable`.

wiredfool:
merged in #2767

PR #1896, opened 2016-05-07
Added support for GIF comment extension

### LINKED ISSUE #1895: Maybe not an issue but a need for support
Hello, gif similarly to png with itxt or ztxt can contains arbitrary plain text metadata *(comment extension)*.

However, looking at the source code and the documentation of Pillow, I wasn’t able to find any relevant function for reading or writing them.

radarhere:
Resolves #1895

### DIFF
--- a/PIL/GifImagePlugin.py
+++ b/PIL/GifImagePlugin.py
@@ -198,6 +198,11 @@ class GifImageFile(ImageFile.ImageFile):
                         # correct, but it seems to prevent the last
                         # frame from looking odd for some animations
                         self.disposal_method = dispose_bits
+                elif i8(s) == 254:
+                    #
+                    # comment extension
+                    #
+                    self.info["comment"] = block
                 elif i8(s) == 255:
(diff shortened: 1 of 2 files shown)


ytrezq:
A last thing : how would I be able to use this code to write comments ?

radarhere:
im = Image.open('test.gif')
    im.info['comment'] = b"Comment text"
    im.save('test.gif')

ytrezq:
I know this isn’t related to pillow but what is theoretical maximum comment size ?

ytrezq:
I know this isn’t related to pillow but does ɢɪꜰ extensions support compression ?

radarhere:
I highly suspect that there is no maximum size of a GIF file. Here is a link to the GIF89a specification - https://www.w3.org/Graphics/GIF/spec-gif89a.txt - a brief search doesn't show anything, and if it's not in the specification, then I wouldn't think that there is any kind of universal limit.

PR #1900, opened 2016-05-08
Changed BUILDME to be easier to execute

radarhere:


wiredfool:
Honestly, given that we have a makefile and a large set of documentation shipped, I think this file can actually just go away.

PR #1793, opened 2016-04-01
Added Pillow to LICENSE

radarhere:
As discussed in #1507, this PR is created for inspection and review.

PR #1901, opened 2016-05-08
Added reference to Pillow to COPYING

radarhere:
Keeping it up to date with the changes from #1793, and in sync with the rest of the LICENSE file.

### DIFF
--- a/docs/COPYING
+++ b/docs/COPYING
@@ -1,7 +1,14 @@
-The Python Imaging Library is
+The Python Imaging Library (PIL) is
 
-Copyright (c) 1997-2009 by Secret Labs AB
-Copyright (c) 1995-2009 by Fredrik Lundh
+    Copyright © 1997-2011 by Secret Labs AB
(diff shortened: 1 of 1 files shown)


aclark4life:
Thx

wiredfool:
So, the thing about this one is that "MIT-like license" means less than nothing when applied to legal language. 

Specifically, this clause: 

  Permission to use, copy, modify, and distribute this software and its associated documentation for any purpose and without fee is hereby granted

Is not MIT like. I'm not entirely sure what it does mean, but it dates back to the PIL license, and it is significantly less free than any of the usual open source licenses.

aclark4life:
It also broke the build :-) (Not sure why)

hugovk:
The build broke due to a third-party issue: https://github.com/eliben/pycparser/issues/147, perfectly timed for release day :)

Here's a workaround: PR https://github.com/python-pillow/Pillow/pull/2143.

PR #1905, opened 2016-05-11
Do not save incorrect key

radarhere:
Originally, this PR aimed to add a warning for the JpegImagePlugin 'progression' info key, resolving the FIXME. However, my test was failing, I suspect due to the warning being caught as part of Pillow's save process.

I have instead updated this to remove setting the incorrect key when saving, so that Pillow is not actively creating incorrect files.


radarhere:
I've since realised that, actually, nothing's changed since this code was originally written, so there's no new reason for this PR to be incorporated, other than the passage of time. If anyone is so inclined, feel free to close this PR. Or I might.

Issue #1850, opened 2016-04-21
mempcpy undeclared on installation (alpine:edge docker)


ColtonProvias:
### What did you do?

Attempted to install Pillow on Alpine Linux

### What did you expect to happen?

Successful installation via pip3.

### What actually happened?

Compile-time error with mempcpy missing.  The issue occurs with both the latest version on PyPI and the latest master.

```
Collecting pillow
  Using cached Pillow-3.2.0.tar.gz
Installing collected packages: pillow
  Running setup.py install for pillow
    Complete output from command /usr/bin/python3.5 -c "import setuptools, tokenize;__file__='/tmp/pip-build-3z4zmxf1/pillow/setup.py';exec(compile(getattr(tokenize, 'open', open)(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" install --record /tmp/pip-9a36m_ws-record/install-record.txt --single-version-externally-managed --compile:
    Single threaded build, not installing mp_compile:1 processes
    running install
    running build
    running build_py
    creating build
    creating build/lib.linux-x86_64-3.5
    creating build/lib.linux-x86_64-3.5/PIL
    copying PIL/BdfFontFile.py -> build/lib.linux-x86_64-3.5/PIL
    copying PIL/MicImagePlugin.py -> build/lib.linux-x86_64-3.5/PIL
    copying PIL/ImageStat.py -> build/lib.linux-x86_64-3.5/PIL
    copying PIL/ImageFont.py -> build/lib.linux-x86_64-3.5/PIL
    copying PIL/ImtImagePlugin.py -> build/lib.linux-x86_64-3.5/PIL
    copying PIL/features.py -> build/lib.linux-x86_64-3.5/PIL
    copying PIL/DcxImagePlugin.py -> build/lib.linux-x86_64-3.5/PIL
    copying PIL/PcxImagePlugin.py -> build/lib.linux-x86_64-3.5/PIL
    copying PIL/GimpGradientFile.py -> build/lib.linux-x86_64-3.5/PIL
    copying PIL/GdImageFile.py -> build/lib.linux-x86_64-3.5/PIL
    copying PIL/GribStubImagePlugin.py -> build/lib.linux-x86_64-3.5/PIL
    copying PIL/PalmImagePlugin.py -> build/lib.linux-x86_64-3.5/PIL
    copying PIL/FontFile.py -> build/lib.linux-x86_64-3.5/PIL
[...]

PR #1906, opened 2016-05-11
Include Python.h before wchar.h so _GNU_SOURCE is set consistently

hugovk:
Fixes https://github.com/python-pillow/Pillow/issues/1850

---

https://github.com/python-pillow/Pillow/issues/1850#issuecomment-218428866:

> We figured out what the real problem is. What happens is:

> First will `wchar.h`be included which will pull in `string.h` without `_GNU_SOURCE` set. Later will `Python.h` explicitly set `_GNU_SOURCE` (via `pyconfig.h`) and after that is `string.h`pulled in again but this time with `_GNU_SOURCE` set.

> This (rightfully) confuses the fortify-headers. You cannot really include some system headers with without _GNU_SOURCE and some with. It will cause problem.

> Either `Python.h` should not set `_GNU_SOURCE` (eg bug in python) or it is required to be included first, before any system headers, (eg bug is in Pillow). Workaround/fix is trivial:

````diff
diff --git a/_imagingcms.c b/_imagingcms.c
index ad5b845..117d8a7 100644
--- a/_imagingcms.c
+++ b/_imagingcms.c
@@ -25,9 +25,9 @@ kevin@cazabon.com\n\
 http://www.cazabon.com\n\
 "
 
-#include "wchar.h"
 
 #include "Python.h"
+#include "wchar.h"
 #include "datetime.h"
 
 #include "lcms2.h"
````

### DIFF
--- a/_imagingcms.c
+++ b/_imagingcms.c
@@ -25,9 +25,8 @@ kevin@cazabon.com\n\
 http://www.cazabon.com\n\
 "
 
+#include "Python.h" // Include before wchar.h so _GNU_SOURCE is set
 #include "wchar.h"
-
-#include "Python.h"
 #include "datetime.h"
 
 #include "lcms2.h"


ncopa:
It is correct to include `Python.h` before any system headers. :+1:

PR #1908, opened 2016-05-12
Add more detail to the appveyor test runs

wiredfool:
Allows us to see where a test is hanging on appveyor.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -26,7 +26,7 @@ build_script:
 test_script:
 - cd c:\pillow
 - '%PYTHON%\Scripts\pip.exe install nose'
-- '%PYTHON%\python.exe test-installed.py'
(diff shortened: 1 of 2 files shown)


radarhere:
I understand that you've added the verbose flag to AppVeyor, and I like the change to test-installed.py.

Why the s flag though? The man reads as 'don't add user site directory to sys.path'. What is the expected difference of this on AppVeyor builds?

wiredfool:
It's from nosetests, short for `--nocapture`: don't capture stdout and let it flow unbuffered.

radarhere:
Right, because if it was an argument applying to Python it would have to come before the reference to test-installed.py. Thanks for the explanation.

homm:
👍

PR #1909, opened 2016-05-12
Get ICC profile from `encoderinfo` while PNG saving

homm:


radarhere:
I presume this is your fix for #1910?

homm:
@radarhere No, this is not related to the data type of ICC profile, only to saving in PNG.

homm:
**Motivation**

Currently JPEG and WEBp both support `image.save('out_file', icc_profile=icc_profile)`, while PNG always get icc_profile from `image.info['icc_profile']`. This PR allows set `icc_profile` while saving and also discard `icc_profile` while saving without removing `image.info['icc_profile']`:

```python
image.save('out_file', icc_profile=None)
```

PR #1912, opened 2016-05-15
Premultiplied luminosity

homm:
This adds new `La` mode and test for `RGBA` and `LA` resampling.


homm:
I haven't added anything to the docs because I don't want to support additional modes officially. In fact `RGBa` and `La` modes are just workarounds for fast computing with very low precision (see skipped tests in this PR).

homm:
rebased

Issue #1660, opened 2016-01-12
PillowPatch.py


v-python:
The first section of the code below will patch old versions of Pillow (3.0.0 at least) with the newer ImageDraw.text .textsize .multiline_text and .multiline_textsize from the repository. The second section of the code below patches in a new trio of APIs:  ImageFont.getBB to get a bounding box for a single line of text, ImageDraw.textInfo to return a bounding box and analysis results for single or multiline text, and ImageDraw.textAtPos to actually draw single or multiline text at a specific position, or at various positions relative to the image on which you are drawing.

I have no experience with anything except Issues on GitHub. I barely know what a Pull request is. I'd be happy to go through any specific incantations that are required if someone tells me how.  Meantime, here's the code, which I store in a file named PillowPatch.py (for now), and import after Pillow, so that I can use this functionality.

Markdown compatible version of the code in the next comment.

v-python:
Markdown interprets Python code in interesting manners.

```python
from PIL import Image, ImageDraw, ImageFont

def text(self, xy, text, fill=None, font=None, anchor=None):
    if self._multiline_check(text):
        return self.multiline_text(xy, text, fill, font, anchor)

    ink, fill = self._getink(fill)
    if font is None:
        font = self.getfont()
    if ink is None:
        ink = fill
    if ink is not None:
        try:
            mask, offset = font.getmask2(text, self.fontmode)
            xy = xy[0] + offset[0], xy[1] + offset[1]
        except AttributeError:
            try:
                mask = font.getmask(text, self.fontmode)
            except TypeError:
                mask = font.getmask(text)
        self.draw.draw_bitmap(xy, mask, ink)

def multiline_text(self, xy, text, fill=None, font=None, anchor=None,
                   spacing=4, align="left"):
    widths = []
    max_width = 0
    lines = self._multiline_split(text)
    line_spacing = self.textsize('A', font=font)[1] + spacing
    for line in lines:
        line_width, line_height = self.textsize(line, font)
        widths.append(line_width)
        max_width = max(max_width, line_width)
    left, top = xy
    for idx, line in enumerate(lines):
        if align == "left":
            pass  # left = x
        elif align == "center":
            left += (max_width - widths[idx]) / 2.0
        elif align == "right":
            left += (max_width - widths[idx])
        else:
            assert False, 'align must be "left", "center" or "right"'
        self.text((left, top), line, fill, font, anchor)
        top += line_spacing
        left = xy[0]

##
[...]

PR #1915, opened 2016-05-19
Text positioning fixes for odd fonts

MicahChambers:
PR for https://github.com/python-pillow/Pillow/issues/1660

### DIFF
--- a/PIL/ImageDraw.py
+++ b/PIL/ImageDraw.py
@@ -344,7 +344,7 @@ class ImageDraw(object):
                 line_bounding_boxes)
 
 
-    def draw_at_pos(self, text, font=None, line_height=None,
+    def text_at_pos(self, text, font=None, line_height=None,
                     line_height_percent=None, origin=(0, 0), align_x='exact',
                     align_y='exact', justify_x='left', fill=None):
         """
--- a/Tests/test_imagefont.py
+++ b/Tests/test_imagefont.py
@@ -180,6 +180,36 @@ try:
 
                 self.assert_image_similar(im, target_img, .5)
 
+        def test_render_text_at_pos(self):
+            ttf = ImageFont.truetype(FONT_PATH, FONT_SIZE)
+
+            # Test that text() correctly connects to multiline_text()
+            # and that align defaults to left
+            im = Image.new(mode='RGB', size=(300, 100))
+            draw = ImageDraw.Draw(im)
+            draw.text_at_pos((0, 0), TEST_TEXT, font=ttf)
+
+            target = 'Tests/images/multiline_text.png'
+            target_img = Image.open(target)
+
+            self.assert_image_similar(im, target_img, .5)
+
+            # Test align center and right
+            for align_x in ('left', 'center', 'right', 'exact'):
+                for align_y in ('top', 'middle', 'bottom', 'exact'):
+                    for justify_x in ('left', 'right', 'center'):
+                        im = Image.new(mode='RGB', size=(300, 100))
+                        ext = '%s_%s_%s' % (align_x, align_y, justify_x)
+                        draw = ImageDraw.Draw(im)
+                        draw.text_at_pos(TEST_TEXT, font=ttf, align_x=align_x,
+                                         align_y=align_x, justify_x=justify_x)
+                        del draw
+
+                        target = 'Tests/images/text_at_pos_'+ext+'.png'
+                        target_img = Image.open(target)
+
+                        self.assert_image_similar(im, target_img, .5)
+
         def test_unknown_align(self):
             im = Image.new(mode='RGB', size=(300, 100))
             draw = ImageDraw.Draw(im)


PIL/ImageDraw.py
```diff
@@ -305,6 +304,120 @@ def multiline_textsize(self, text, font=None, spacing=4):
...
+        txWid = 0
+        tyMax = None
+        tyMin = None
+        if len( lines ) > 1:
+            if lineHeight is None:
+                if lineHeightPercent is None:
+                    lineHeightPercent = 100
+                try:
+                    lineHeight = font.font.height
+                except Exception as exc:
```
hugovk: `exc` is assigned to but unused.

Do we want to catch *all* exceptions, or would it be better just to catch some and let others go up?

hugovk:
@MicahChambers Thanks for the PR.

Please can you summarise what this actually fixes, and include unit tests?

PIL/ImageDraw.py
```diff
@@ -305,6 +304,120 @@ def multiline_textsize(self, text, font=None, spacing=4):
...
+    # can also be 'top', 'middle', or 'bottom'. justifyX can be 'left', 'center',
+    # or 'right'. Other parameters like textInfo.
+    def textAtPos( self, text, font=None, lineHeight=None, lineHeightPercent=None,
+                   origin=( 0, 0 ), alignX='exact', alignY='exact', justifyX='left',
+                   fill=None ):
+        if font is None:
+            font = self.getfont()
+        ink, fill = self._getink( fill )
+        if ink is None:
+            inx = fill
```
hugovk: `inx` is assigned to but unused.

PIL/ImageDraw.py
```diff
@@ -305,6 +304,120 @@ def multiline_textsize(self, text, font=None, spacing=4):
...
+    # above the baseline.
+    # if text contains "\n" it is treated as multiple lines of text.
+    #    yMin becomes really negative.
+    #    lineHeight and lineHeightPercent become relevant.
+    #    lineHeightPercent is only used if LineHeight is None, and defaults to 100%.
+    #    Use of a lineHeight < sum( font.getmetrics ) may result in text overlap.
+    #    The best lineHeight would be that returned from the font.font.height
+    #    attribute, but in versions of PIL in which that isn't accessible,
+    #    sum( font.getmetrics()) is used instead.
+    def textInfo( self, text, font=None, lineHeight=None, lineHeightPercent=None ):
```
hugovk: Please use snake_case naming like `line_height`, `line_height_percent`, `tx_max`, `lx_max` etc.

For the def, `text_info` (or maybe `textinfo` because there's already a `textsize`)

PIL/ImageDraw.py
```diff
@@ -305,6 +304,120 @@ def multiline_textsize(self, text, font=None, spacing=4):
...
+        if font is None:
+            font = self.getfont()
+        lines = text.split('\n')
+        txMax = 0
+        txMin = 0
+        txWid = 0
+        tyMax = None
+        tyMin = None
```
hugovk: What does the `t` indicate in these? Can they be named more descriptively? And let's use width rather than wid.

PIL/ImageDraw.py
```diff
@@ -305,6 +304,120 @@ def multiline_textsize(self, text, font=None, spacing=4):
...
+            if lineHeight is None:
+                if lineHeightPercent is None:
+                    lineHeightPercent = 100
+                try:
+                    lineHeight = font.font.height
+                except Exception as exc:
+                    lineHeight = sum( font.getmetrics())
+                lineHeight = int( lineHeight * lineHeightPercent / 100 )
+
+        lineBBs = []
```
hugovk: What's a line BB? Please use a more descriptive name (in snake_case).

PIL/ImageDraw.py
```diff
@@ -305,6 +304,120 @@ def multiline_textsize(self, text, font=None, spacing=4):
...
+        for line in lines:
+            lxMin, lyMin, lxMax, lyMax = font.getBB( line )
```
hugovk: What does the `l` indicate in these? Can they be named more descriptively (in snake_case)?

PIL/ImageDraw.py
```diff
@@ -305,6 +304,120 @@ def multiline_textsize(self, text, font=None, spacing=4):
...
+            lineBBs.append(( lxMin, lyMin, lxMax, lyMax, line ))
+            lxWid = lxMax - lxMin
```
hugovk: Let's use width rather than wid

PIL/ImageDraw.py
```diff
@@ -305,6 +304,120 @@ def multiline_textsize(self, text, font=None, spacing=4):
...
+                   fill=None ):
+        if font is None:
+            font = self.getfont()
+        ink, fill = self._getink( fill )
+        if ink is None:
+            inx = fill
+            if ink is None:
+                return
+        if justifyX not in ('left', 'center', 'right'):
+            raise ValueError('Unknown justifyX value "%s".' % justifyX )
```
hugovk: These two validation lines could be at the top of this function: if validation fails, can raise immediately without making any font or ink calls.

PIL/ImageDraw.py
```diff
@@ -305,6 +304,120 @@ def multiline_textsize(self, text, font=None, spacing=4):
...
+        txMin, tyMin, txMax, tyMax, txWid, lineHeight, lineBBs = self.textInfo(
+            text, font, lineHeight, lineHeightPercent )
+        if alignX == 'exact':
+            ox = 0
```
hugovk: What is the `o` in `ox` and `oy`? Can they be named more descriptively (in snake_case)?

PIL/ImageFont.py
```diff
@@ -169,6 +169,15 @@ def font_variant(self, font=None, size=None, index=None, encoding=None):
                             encoding=self.encoding if encoding is None else
                             encoding)
 
+    def getBB( self, text ):
```
hugovk: What's a BB? Could this be named more descriptively? A docstring (`"""TODO"""`) would be useful.

PIL/ImageDraw.py
```diff
@@ -305,6 +304,120 @@ def multiline_textsize(self, text, font=None, spacing=4):
...
+    # origin, yMin is negative pixels below the baseline, yMax is positive pixels
+    # above the baseline.
+    # if text contains "\n" it is treated as multiple lines of text.
+    #    yMin becomes really negative.
+    #    lineHeight and lineHeightPercent become relevant.
+    #    lineHeightPercent is only used if LineHeight is None, and defaults to 100%.
+    #    Use of a lineHeight < sum( font.getmetrics ) may result in text overlap.
+    #    The best lineHeight would be that returned from the font.font.height
+    #    attribute, but in versions of PIL in which that isn't accessible,
+    #    sum( font.getmetrics()) is used instead.
```
hugovk: Could this long comment box go into a docstring?
```python
"""
TODO
"""
```

PIL/ImageFont.py
v-python: BB = Bounding Box

PIL/ImageDraw.py
v-python: typo. Should be ink.

PIL/ImageDraw.py
v-python: t = text, x & y are Cartesian coordinate directions.

PIL/ImageDraw.py
v-python: l is for line

PIL/ImageDraw.py
v-python: o is for origin

MicahChambers:
Thanks for the review! I'll look over it tonight.

PIL/ImageDraw.py
MicahChambers: @v-python what got raised here? An AttributeError I'm guessing?

PIL/ImageDraw.py
MicahChambers: :+1:

MicahChambers:
@v-python @hugovk I'll go through and do a little cleanup. I'll

1) change variable names to be more descriptive
2) Add tests
3) add docstrings describing the arguments to things.

The reason I didn't do this before when I was dumping @v-python 's code into a PR is that I have submitted PR's to open source projects in the past and gotten no response -- so I didn't want to spend a bunch of time if it wasn't going to get picked up. @hugovk you can probably ignore this for a week and I'll ping you when things are more "pythonic"

PIL/ImageFont.py
hugovk: bbox is a good compromise for bounding box and is already used elsewhere in the codebase.

PIL/ImageDraw.py
v-python: Yes. When integrated with PILLOW, that try/except may not be needed. The code was dealing with whether or not font.font.height was defined, which was a recent addition, but of course not quite as recent as this code is being added, so you can probably scoot line 333 to the left and drop the try and the whole except block.

MicahChambers: Ahh I see. Thanks.

MicahChambers:
@hugovk @v-python I need to add tests, would it make sense to just replace the current draw text function?

hugovk:
Which function?

It's better to add new tests than replace existing ones. We want to keep existing ones to guard against regressions (but of course update any where functionality has changed).

MicahChambers:
Oh, I mean, `draw_at_pos` is basically doing the same thing as `text()`, so maybe this should just be an updated of `text()` to draw lines more accurately?

v-python:
Current text() tests might be revised to test draw_at_pos(). But IIRC, text doesn't let you specify a baseline... it positions the top left corner, which is not totally useless, but it makes it really hard to line things up that don't want to all be the same font or same font size. So updating text() to have the function of draw_at_pos() would require a significant rework of the code of either or both... and it would lose the ease-of-use of having things be aligned on a particular baseline.

v-python:
I'm sorry this effort seems to have stalled, and bit-rotted. But now it seems Pillow doesn't even support text on Windows anymore.  #3407

PR #1916, opened 2016-05-20
Extra font info

MicahChambers:
Added some utilities for checking what characters are actually included in a loaded font. I personally find it frustrating to only be able to discover errors by looking at the rendered text.

### DIFF
--- a/PIL/ImageFont.py
+++ b/PIL/ImageFont.py
@@ -29,6 +29,7 @@ from PIL import Image
 from PIL._util import isDirectory, isPath
 import os
 import sys
+import six
 
 
 class _imagingft_not_installed(object):
@@ -168,6 +169,7 @@ class FreeTypeFont(object):
                             index=self.index if index is None else index,
                             encoding=self.encoding if encoding is None else
                             encoding)
+
     def hasglyphs(self, text):
         """
         Returns true if all the characters in the string exist in the font,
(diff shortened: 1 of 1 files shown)


hugovk:
@MicahChambers Thanks for the PR.

The tests are failing for Python 3 because of `unichr`: https://travis-ci.org/python-pillow/Pillow/builds/131615297

PIL/ImageFont.py
```diff
@@ -168,6 +168,20 @@ def font_variant(self, font=None, size=None, index=None, encoding=None):
                             index=self.index if index is None else index,
                             encoding=self.encoding if encoding is None else
                             encoding)
+    def hasglyphs(self, text):
```
hugovk: Please add a newline between these defs.

MicahChambers:
Thanks for the review. I'll look over it tonight!

PIL/ImageFont.py
MicahChambers: :+1:

MicahChambers:
Hmm. Not sure what is causing this segfault. Reading the freetype documentation I'm thinking listing the characters might be a bad idea: https://www.freetype.org/freetype2/docs/reference/ft2-base_interface.html#FT_Get_First_Char 

Basically malformed charmaps happen and can result in 4 billion iterations....

I'm going to investigate to see if there is a way to get the all the glyphs then reverse map to the primary character mapping to each glyph.... If possible.

PR #1920, opened 2016-05-21
Tests for resampling accuracy

homm:


homm:
That tests check resampling algorithm and filters accuracy. They are very strict and require certain values as a result but I think there are no reasons why we can get different values across platforms or compilers.

The confusing part is that we are testing diagonal-symmetric sample and sometimes get not symmetric result, like `da` and `d9` values in `test_reduce_lanczos`:

```python
data = ('e1 e0 e4 d7'
        'e0 df e3 d6'
        'e4 e3 e7 da'
        'd7 d6 d9 ce')
```

But this has the simple explanation: we are resampling images in two passes: first horizontal, then vertical. The result of the first pass is stored in the 8-bit temporary image. When we are resampling in the second time, we are working with the rounded result of the first pass.

PR #1932, opened 2016-05-27
RF: run-time loading of Tcl / Tk functions

matthew-brett:
Discover Tcl / Tk functions at run-time by looking in libs loaded by
Tkinter Python module.

Use header excerpts from Tcl / Tk so we don't need an installed Tcl / Tk
to build.

### DIFF
--- a/PIL/_tkinter_finder.py
+++ b/PIL/_tkinter_finder.py
@@ -1,0 +1,20 @@
+""" Find compiled module linking to Tcl / Tk libraries
+"""
+import sys
+
+if sys.version_info[0] > 2:
+    from tkinter import _tkinter as tk
+else:
+    from Tkinter import tkinter as tk
+
+if hasattr(sys, 'pypy_find_executable'):
+    # Tested with packages at https://bitbucket.org/pypy/pypy/downloads.
+    # PyPies 1.6, 2.0 do not have tkinter built in.  PyPy3-2.3.1 gives an
+    # OSError trying to import tkinter. Otherwise:
+    try: # PyPy 5.1, 4.0.0, 2.6.1, 2.6.0
+        TKINTER_LIB = tk.tklib_cffi.__file__
+    except AttributeError:
+        # PyPy3 2.4, 2.1-beta1; PyPy 2.5.1, 2.5.0, 2.4.0, 2.3, 2.2, 2.1
+        TKINTER_LIB = tk.tkffi.verifier.modulefilename
+else:
+    TKINTER_LIB = tk.__file__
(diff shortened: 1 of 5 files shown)


matthew-brett:
I've tested on OSX Python 2, 3; Linux Python 2, 3 and Windows Python 2.

matthew-brett:
pypy builds failing - any suggestions what to do about these?

matthew-brett:
I looked for a while, but I can't see any reasonable way of locating the built tkinter interface `.so` file on pypy3.   The ``.so`` file results from a `cffi` build, and the library is available via a cffi interface in `tikinter.tklib`, but I can't see any way to get to the library `.so` filename or the pointer to the handle. 

So, unless someone has a good idea, I think this approach doesn't work for pypi3.

matthew-brett:
It's not too hard to solve this for PyPy 5.1 (Python 2 version, that has a recent `cffi` package version).   The problem is with PyPy3 2.4 which has a very old `cffi` package version : 

https://groups.google.com/d/msg/python-cffi/gkPqq1u52E4/ReKJWJiPCgAJ

Will you be supporting Python 3.2 for much longer?   How about PyPy3 (which is an implementation of Python 3.2)?

wiredfool:
It looks like there hasn't been a pypy3 release in almost 2 years. I'm guessing that it's likely that the improvements from pypy will make it in on the next release, and then this will likely work. 

Is it possible that a newer version of cffi would be installable alongside pypy3?

As for 3.2 support, I believe that there are still LTS distros where 3.2 is the python3 version. I'd rather not drop support for them without a good reason, though paring down the build matrix is something that might happen first. 3.2 is probably the first to go when it does, as it's the least like the others.

matthew-brett:
I think I've got this working for PyPy3 and other old PyPies.  I've tested lots of PyPy and PyPy3 versions from https://bitbucket.org/pypy/pypy/downloads and the finder works for all versions that have a compiled tkinter.  I'll check on the cffi mailing list that this is likely to work for future cffi versions.

matthew-brett:
Tests all passing - from my point of view, this one is ready to merge.

matthew-brett:
From discussion at https://groups.google.com/forum/#!msg/python-cffi/gkPqq1u52E4/ReKJWJiPCgAJ - algorithm for finding Tcl / Tk libraries should continue to work for future pypi versions. 

This branch also works with pypi3 5.2.0-alpha1 (released a few days ago).

wiredfool:
@crjh @ddale @jakirkham -- Can you check this PR without using the `--disable-osx-tcltk-framework` flag and confirm that it works for you?

PR #1933, opened 2016-05-28
Resample vertical pass

homm:
This PR implements separate vertical pass of resampling instead of two transposes.

Originally [I've replaced](https://github.com/python-pillow/Pillow/commit/40f9f48) two passes with transpose for the following reasons:

1. It was harder to heavily change the code in two branches simultaneously.
2. It makes downscaling faster, although upscaling became slightly slower.

Since then there are many tests for resampling were written and many optimizations were implemented. Now it is time to return upscaling performance.

Result on aws ec2 c4.large with 1920×1280 RGB image. Megapixels of source image per second.

            | Destination   | Filter  | 3.2    | FPI    | Vertical pass
------------|---------------|---------|--------|--------|--------------
Downscaling | **16x16**     | Bilinear| 229.31 | 381.97 | 386.89
            |               | Bicubic | 120.64 | 203.59 | 204.98
            |               | Lanczos |  81.33 | 135.74 | 136.20
            | **320x180**   | Bilinear| 133.08 | 161.96 | 193.30
            |               | Bicubic |  77.00 | 109.46 | 123.08
            |               | Lanczos |  51.36 |  77.48 |  83.27
Upscaling   | **2048x1155** | Bilinear|  25.78 |  27.81 |  40.14
            |               | Bicubic |  20.60 |  23.75 |  31.83
            |               | Lanczos |  17.04 |  19.50 |  25.00
            | **5120x2880** | Bilinear|   5.55 |   5.74 |   8.91
            |               | Bicubic |   4.54 |   4.98 |   7.25
            |               | Lanczos |   3.80 |   4.33 |   5.93

I also include results of stable version and fixed-point separetely to show what FPI doesn't affect upscaling too much, while vertical pass does. Vertical pass faster for upscaling in average by 40% and also makes downscaling to reasonable sizes 10% faster in average.

This last two improvements together makes all resampling operations about 60% faster.

### DIFF
--- a/PIL/ImageMode.py
+++ b/PIL/ImageMode.py
@@ -39,6 +39,7 @@ def getmode(mode):
         for m, (basemode, basetype, bands) in Image._MODEINFO.items():
             _modes[m] = ModeDescriptor(m, bands, basemode, basetype)
         # extra experimental modes
+        _modes["RGBa"] = ModeDescriptor("RGBa", ("R", "G", "B", "a"), "RGB", "L")
         _modes["LA"] = ModeDescriptor("LA", ("L", "A"), "L", "L")
         _modes["La"] = ModeDescriptor("La", ("L", "a"), "L", "L")
         _modes["PA"] = ModeDescriptor("PA", ("P", "A"), "RGB", "L")
--- a/PIL/ImageMode.py
+++ b/PIL/ImageMode.py
@@ -40,6 +40,7 @@ def getmode(mode):
             _modes[m] = ModeDescriptor(m, bands, basemode, basetype)
         # extra experimental modes
         _modes["LA"] = ModeDescriptor("LA", ("L", "A"), "L", "L")
+        _modes["La"] = ModeDescriptor("La", ("L", "a"), "L", "L")
         _modes["PA"] = ModeDescriptor("PA", ("P", "A"), "RGB", "L")
         # mapping modes
         _modes["I;16"] = ModeDescriptor("I;16", "I", "L", "L")
--- a/libImaging/Resample.c
+++ b/libImaging/Resample.c
@@ -91,7 +91,7 @@ ImagingPrecompute(int inSize, int outSize, struct filter *filterp,
     /* determine support size (length of resampling filter) */
     support = filterp->support * filterscale;
 
-    /* maximum number of coofs */
+    /* maximum number of coeffs */
     kmax = (int) ceil(support) * 2 + 1;
 
     // check for overflow
@@ -102,7 +102,9 @@ ImagingPrecompute(int inSize, int outSize, struct filter *filterp,
     if (outSize > INT_MAX / (2 * sizeof(double)))
         return 0;
 
-    /* coefficient buffer */
+    /* coefficient buffer. kmax elements for each of outSize pixels.
+       Only xmax number of coefficients are initialized (xmax <= kmax),
+       other coefficients should be 0, so we are using calloc.  */
     kk = calloc(outSize * kmax, sizeof(double));
     if ( ! kk)
         return 0;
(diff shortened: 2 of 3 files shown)


PIL/ImageMode.py
```diff
@@ -40,6 +40,7 @@ def getmode(mode):
             _modes[m] = ModeDescriptor(m, bands, basemode, basetype)
         # extra experimental modes
         _modes["LA"] = ModeDescriptor("LA", ("L", "A"), "L", "L")
+        _modes["La"] = ModeDescriptor("La", ("L", "a"), "L", "L")
```
homm: This required for `Image.merge('La', ...)`. But there is no 'RGBa', so I used 'RGBX', which is. Modes support in Pillow is a mess ((

libImaging/Resample.c
```diff
@@ -107,7 +107,7 @@ ImagingPrecompute(int inSize, int outSize, struct filter *filterp,
     if ( ! kk)
         return 0;
 
-    xbounds = calloc(outSize * 2, sizeof(int));
+    xbounds = malloc(outSize * 2 * sizeof(int));
```
homm: Returned malloc because actually remaining memory can be not initialized only in one place.

homm:
As a bonus: It will be easier for me to maintain Pillow-SIMD, because Pillow-SIMD does two separate passes from the start.

PIL/ImageMode.py
homm: @wiredfool If you don't mind about adding `La` in `_modes`, I'd like to merge it. I don't see any other controversial points in this RP.

wiredfool: I went through the review, hit this, started wondering why we didn't have `RGBa` as a mode here, and then determined that modes are a mess.

homm: I can add `RGBa` here as a temporary solution.

libImaging/Resample.c
wiredfool: I don't understand this reasoning.

homm: I've added the [comment](https://github.com/python-pillow/Pillow/pull/1933/files#diff-b1e715e96fb950759e657d15a0b3f4f9R105) why we need calloc.

PIL/ImageMode.py
homm: Added.

libImaging/Resample.c
wiredfool: It's the calloc/malloc change I don't understand, specifically: "memory can be not initialized only in one place."

PIL/ImageMode.py
wiredfool: works for me

libImaging/Resample.c
homm: malloc: just give me mem. calloc: ensure that if zero-initialized.

In case of `kk` we request `outSize * kmax` elements, but assign only `outSize * xmax` elements, where `xmax` <= `kmax`. And it is better than all other elements (between xmax and kmax) would be zero (because letter it is simpler to iterate over all `kk` elements). We can make it manually after [this loop](https://github.com/python-pillow/Pillow/pull/1933/files#diff-b1e715e96fb950759e657d15a0b3f4f9R130) or we can use calloc.

In the case of `xbounds` we always assign values to all elements. It is not required that memory should be zero-initialized.

wiredfool: Is there any harm in using calloc?  IIRC, it's already pointed to one bug.

homm: What exactly do you mean? ;)
`calloc` is just a function and it does what it should.

I still don't understand why calloc failed on windows while malloc didn't. In my opinion it should depend on pointer size and SIZE_MAX check was correct while the test with fixed size wasn't. But I can't waste much time to investigate real reasons especially when I don't have windows PC. The INT_MAX fix perfectly fits me.

homm: I've returned malloc anywhere and clearing values [with loop](https://github.com/uploadcare/pillow-simd/blob/12c8cf94f244b3fc1039e71c7a664b97a2045554/libImaging/Resample.c#L138).

wiredfool:
Are you ready for this to be merged?

homm:
Yes

Issue #1917, opened 2016-05-20
Image.rotate(90*n, expand=True) results in translation and expansion


BryghtShadow:
### What did you do?
Apply rotation of 90*N degrees to an image, with expand set to True.
### What did you expect to happen?
Because the rotation is a multiple of 90 degrees, I expected the image to be rotated with no translations. I expected the result to match a rotation via `Image.transpose`.
### What actually happened?
* For 0 degrees, no visible difference (Working as intended)
* For -90 degrees, the image is translated by (1+x, 1+y), but expands only in height. This results in the right-most column being cropped.
* For +90 degrees, the image is translated by (1+x, 1+y), but expands only in width. This results in the bottom-most row being cropped.
* For &plusmn;90*(2, 3, ..., N-1, N) degrees, the image is translated by (1+x, 1+y), and expands to fit the image.

### What versions of Pillow and Python are you using?
Pillow 3.2.0; Python 3.4.4 and Python 3.5.1

```python
from PIL import Image, ImageDraw

im = Image.new('RGBA', (11,13), 'white')
# Add markers to corners to check for any cropping
w, h = im.size
draw = ImageDraw.Draw(im)
draw.line((0, 0, 1, 1), fill='cyan')
draw.line((0, h-1, 1, h-2), fill='magenta')
draw.line((w-1, h-1, w-2, h-2), fill='yellow')
draw.line((w-1, 0, w-2, 1), fill='black')
del draw

im.show()

print('  deg :  old  ->  new')
for i in range(-4, 5):
    deg = i * 90
    im2 = im.rotate(deg, expand=True)
    print('{:>5} : {}x{} -> {}x{}'.format(deg, im.width, im.height, im2.width, im2.height))
    im2.show()
```

manushifva:
Use `expand=1`. Don't use `expand=True`.

PR #1941, opened 2016-06-03
Cleanup transforms

homm:
Internal transform API was very confusing. For example, `rotate` method was calling `self.transform` or `self.im.rotate` depends on `expand` argument, but `self.im.rotate` internally also calls `ImagingTransformAffine`, so it was excessive.

 * First of all, this PR fixes rotations bug #1917 in two ways:
    * It really uses very cheap `transpose` operations if it is possible regardless of filter. https://github.com/python-pillow/Pillow/pull/1941/commits/8203a43d262d349f71bcfa3da46e5f3e30fe99e0
    * It rounds coefficients in the transform matrix, so even if we are not using cheap transpose, we still get right dimensions because `cos(90°)` now is exactly 0, not 6.123e-17. https://github.com/python-pillow/Pillow/pull/1941/commits/7687ce829e50635838707f5153079fb4d34cd41e
 * `Image.rotate` always uses `self.transform`, so `im.rotate` was removed from core library. https://github.com/python-pillow/Pillow/pull/1941/commits/709078efd14cfcffaedf59e883f56aa16d61926e https://github.com/python-pillow/Pillow/pull/1941/commits/1f8c2527baaff6357f31761b6340d2016920781f
 * The order of coefficients in internal API now match matrix representation and order of external API. https://github.com/python-pillow/Pillow/pull/1941/commits/2b77b1cec7fae15112be03dd5c1bfbb0169c3e0b https://github.com/python-pillow/Pillow/pull/1941/commits/3d622d60cf7c6c609cfc48b774ee0d8aef2d1d85
 * Internal `ImagingTransformAffine`, `ImagingTransformPerspective` and `ImagingTransformQuad` was replaced with one entry point to all transformations: `ImagingTransform`. Former `ImagingTransform` was  renamed to `ImagingGenericTransform` and removed from extern symbols (`Imaging.h` file). https://github.com/python-pillow/Pillow/pull/1941/commits/6be3df2a432bdc61d0705ebacfc878b93cc86b9c https://github.com/python-pillow/Pillow/pull/1941/commits/9902d2e1c50712a7fa1d308fab124bf5c71f018e

### DIFF
--- a/PIL/Image.py
+++ b/PIL/Image.py
@@ -1869,9 +1869,7 @@ class Image(object):
         h = box[3] - box[1]
 
         if method == AFFINE:
-            # change argument order to match implementation
-            data = (data[2], data[0], data[1],
-                    data[5], data[3], data[4])
+            data = data[0:6]
 
         elif method == EXTENT:
             # convert extent to an affine transform
@@ -1879,7 +1877,7 @@ class Image(object):
             xs = float(x1 - x0) / w
             ys = float(y1 - y0) / h
             method = AFFINE
-            data = (x0 + xs/2, xs, 0, y0 + ys/2, 0, ys)
+            data = (xs, 0, x0 + xs/2, 0, ys, y0 + ys/2)
 
         elif method == PERSPECTIVE:
             # change argument order to match implementation
--- a/PIL/Image.py
+++ b/PIL/Image.py
@@ -1880,7 +1880,6 @@ class Image(object):
             data = (xs, 0, x0 + xs/2, 0, ys, y0 + ys/2)
 
         elif method == PERSPECTIVE:
-            # change argument order to match implementation
             data = data[0:8]
 
         elif method == QUAD:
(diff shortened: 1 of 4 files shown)


homm:
rebased

PIL/Image.py
```diff
@@ -1863,28 +1865,24 @@ def transform(self, size, method, data=None, resample=NEAREST, fill=1):
...
             # convert extent to an affine transform
             x0, y0, x1, y1 = data
             xs = float(x1 - x0) / w
             ys = float(y1 - y0) / h
             method = AFFINE
-            data = (x0 + xs/2, xs, 0, y0 + ys/2, 0, ys)
+            data = (xs, 0, x0 + xs/2, 0, ys, y0 + ys/2)
+
         elif method == PERSPECTIVE:
             # change argument order to match implementation
```
wiredfool: This comment doesn't apply anymore, right?

homm: Yes, removed

homm:
Links to commits in the description are updated.

Issue #1939, opened 2016-06-01
Preferring PyQt5 causes errors with PyQt4 applications


ericfrederich:
### What did you do?
Installed both PyQt4 and PyQt5.
Ran a program which imports PyQt4 and uses pillow.

### What did you expect to happen?
I expected the application to run fine with PyQt4.

### What actually happened?
```
  File "/home/user/venv/lib/python3.5/site-packages/PIL/ImageQt.py", line 26, in <module>
    from PyQt5.QtGui import QImage, qRgba, QPixmap
RuntimeError: the PyQt5.QtCore and PyQt4.QtCore modules both wrap the QObject class
```

### What versions of Pillow and Python are you using?

* Python 3.5.1
* pillow from either PyPI or GitHub master branch

Don't think the version of PyQt would matter much, but
* PyQt5: 5.6
* PyQt4: 4.11.4
* sip: 4.18
* Qt4: 4.8.5
* Qt5: 5.6.0

Easy to reproduce...
```python
from PyQt4.QtCore import *
from PIL.ImageQt import fromqpixmap
```
A simple solution would be to change the line https://github.com/python-pillow/Pillow/blob/master/PIL/ImageQt.py#L29 from
`execpt ImportError:` to `except (ImportError, RuntimeError):`

This fix wouldn't help someone with the reverse problem though depending on the order that their programs get imported.

If Pillow is going to support both PyQt4 and PyQt5 at the same time, perhaps the order in which PyQt versions are preferred should be configurable!

wiredfool:
Seems like a reasonable fix.

sorki:
Adding RuntimeError indeed fixed my issue.

hugovk:
I've made a PR: https://github.com/python-pillow/Pillow/pull/1942.

ericfrederich:
Thanks.  This change is welcome.
I think long-term though maybe the precedence should be configurable.

Imagine a Qt4 program which imports Pillow prior to anything from Qt5.

I haven't put any thought into how a configurable import would work... maybe there is an elegant solution.

wiredfool:
Frankly, it's not worth complicating things for that much of an edge case. We can support either, but not both. If someone is really running both qt4 and qt5 in the same app, they can handle the complexity of maintaining a vendor branch or monkey patching or whatever.

ericfrederich:
@wiredfool, I agree it is an edge case, and other ways to solve it.
[...]

PR #1942, opened 2016-06-07
Fix for program importing PyQt4 when PyQt5 also installed

hugovk:
Fixes https://github.com/python-pillow/Pillow/issues/1939 using suggested solution.


hugovk:
@ericfrederich Thanks, PR updated.

PR #1946, opened 2016-06-11
Added catch for UnsupportedOperation in ImageFile load

### LINKED ISSUE #1945: Can't save JPEG2000 to a BytesIO object
### What did you do?
I tried to save a JPEG2000 image into a BytesIO object. 

It should have saved the image file into the BytesIO object.

### What actually happened?
The script fails("python.exe has stopped working"). 

Python 3.3.4 (32-bit) and Pillow 3.2.0
I also tried using Python 3.5.1, but the script still fails.

```
from PIL import Image
from io import BytesIO

#Any image will reproduce the bug.
img=Image.open(r"some-image.png")

print("Saving ordinary JPEG to a BytesIO object. This works fine.")
bio=BytesIO()
img.save(bio,"JPEG")

print("Saving JPEG2000 to a file. This also works fine.")
img.save(r"some-other-image.jpf","JPEG2000")

print("Saving JPEG2000 to a BytesIO object. This fails.")
bio2=BytesIO()
img.save(bio2,"JPEG2000")
```

radarhere:
Resolves #1945

Added a catch for UnsupportedOperation in load. Mirrors the catch for this exception when saving at https://github.com/python-pillow/Pillow/blob/master/PIL/ImageFile.py#L461

### DIFF
--- a/PIL/ImageFile.py
+++ b/PIL/ImageFile.py
@@ -170,7 +170,7 @@ class ImageFile(Image.Image):
                             self.map, self.size, d, e, o, a
                             )
                     readonly = 1
-                except (AttributeError, EnvironmentError, ImportError):
+                except (AttributeError, EnvironmentError, ImportError, io.UnsupportedOperation):
                     self.map = None
 
         self.load_prepare()


radarhere:
Further investigation has lead me to the conclusion that this does not fix the issue after all.

PR #1954, opened 2016-06-14
Do not do unnecessary passes on resizing

homm:
This is a relatively rare case, but why not?

From some point of view, this breaks the consistency because resampling filter is applied to one pass and isn't for another. But Pillow already returns a copy of the image if both dimensions are the same, and there is no way to apply resampling filter to the image from python code without affecting dimension.


homm:
This looks pretty harmless for the 3.3 release if someone has a minute for review. And maybe still useful for some applications.

PR #1958, opened 2016-06-16
Simplified condition in init method

radarhere:
This is minor, so if anyone would like to close, feel free to do so.

In the `init` method, there is a check for `_initialized >= 2`. However, `_initialized` never becomes > 2. So I have changed it to `== 2`, just for clarity.

### DIFF
--- a/PIL/Image.py
+++ b/PIL/Image.py
@@ -388,7 +388,7 @@ def init():
     """
 
     global _initialized
-    if _initialized >= 2:
+    if _initialized == 2:
         return 0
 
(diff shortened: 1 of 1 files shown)


homm:
I think than the condition `_initialized >= 2` is more correct because it works regardless of number of initialization steps. Also it is consistent with [preinit](https://github.com/radarhere/Pillow/blob/b5ef9e4fd15c1b6285b0be55b1260d336fdd0aca/PIL/Image.py#L353).

PR #1959, opened 2016-06-16
New filters! ⭐⭐⭐

homm:
This adds two new filters to `resize` method.

BOX is event faster than LINEAR because it has a smaller window (0.5) and produces very sharp images. Applicable for significant downscaling.

HAMMING costs like LINEAR, also produces very sharp images and applicable for downscaling at any range. Doesn't look good for upscaling, though.

### DIFF
--- a/libImaging/Imaging.h
+++ b/libImaging/Imaging.h
@@ -233,7 +233,6 @@ extern void ImagingError_Clear(void);
 #define IMAGING_TRANSFORM_BILINEAR 2
 #define IMAGING_TRANSFORM_HAMMING 5
 #define IMAGING_TRANSFORM_BICUBIC 3
-#define IMAGING_TRANSFORM_MITCHELL 6
 #define IMAGING_TRANSFORM_LANCZOS 1
 
 typedef int (*ImagingTransformMap)(double* X, double* Y,
--- a/libImaging/Imaging.h
+++ b/libImaging/Imaging.h
@@ -229,9 +229,12 @@ extern void ImagingError_Clear(void);
 
 /* standard filters */
 #define IMAGING_TRANSFORM_NEAREST 0
-#define IMAGING_TRANSFORM_LANCZOS 1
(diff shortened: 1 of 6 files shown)


homm:
This PR is not critical for me and has relatively large change for review, so it can wait until 3.4. The most valuable part from point of view is filters comparison table in documentation. So if we decide to move this PR to 3.4 I can make another PR with the table only.

wiredfool:
Ok, bumping to 3.4.

homm:
Needs rebasing.

libImaging/Imaging.h
```diff
@@ -229,9 +229,11 @@ extern void ImagingError_Clear(void);
 
 /* standard filters */
 #define IMAGING_TRANSFORM_NEAREST 0
-#define IMAGING_TRANSFORM_LANCZOS 1
+#define IMAGING_TRANSFORM_BOX 4
 #define IMAGING_TRANSFORM_BILINEAR 2
+#define IMAGING_TRANSFORM_HAMMING 5
 #define IMAGING_TRANSFORM_BICUBIC 3
+#define IMAGING_TRANSFORM_LANCZOS 1
```
wiredfool: Is there any logical order here?  It doesn't appear to be numeric, alphabetical, or order added.

homm: From less expensive and quality to more. Same as in the comparison table.

homm: @wiredfool Are you happy with this explanation?

wiredfool: Yeah, though I expect someone is going to helpfully clean up the order in a PR in a year or two.

wiredfool:
@homm Can you write up basic release notes for this in docs/release/3.4.0.rst?

homm:
Yes, sure. But I'd prefere to delay this on the weak before the release, to add all notes at once :)

wiredfool:
I'd rather have it in there so that we don't miss it.

PR #1963, opened 2016-06-18
Replaced del with pop

radarhere:
y = x['test']
    del x['test']

can be replaced with

    y = x.pop('test')

This applies to lists as well as dictionaries.


hugovk:
Out of curiosity I timed these two options and the two-line version was about 5% quicker, but in the scheme of things I doubt that'll ever make much difference in real code, especially as these are used during IO.

Merging.

PR #1965, opened 2016-06-19
Changed equals method on Image class to short circuit.

mattBoros:

 * I was having performance issues with the equals method, and short circuiting caused it to speed up by about 200x.
* I tried to run the unit tests but even on the master branch they failed. I'm probably doing something wrong I'm sorry :(

### DIFF
--- a/PIL/Image.py
+++ b/PIL/Image.py
@@ -586,14 +586,17 @@ class Image(object):
     def __eq__(self, other):
         if self.__class__.__name__ != other.__class__.__name__:
             return False
-        a = (self.mode == other.mode)
-        b = (self.size == other.size)
-        c = (self.getpalette() == other.getpalette())
-        d = (self.info == other.info)
-        e = (self.category == other.category)
-        f = (self.readonly == other.readonly)
-        g = (self.tobytes() == other.tobytes())
-        return a and b and c and d and e and f and g
+
+        properties = ['mode', 'size', 'info', 'category', 'readonly']
+
+        props_eq = all(self.__getattribute__(p) == other.__getattribute__(p)
+                       for p in properties)
+        if not props_eq:
+            return False
+
+        funcs_eq = (self.getpalette() == other.getpalette()) \
+                   and (self.tobytes() == other.tobytes())
+        return funcs_eq
 
     def __ne__(self, other):
         eq = (self == other)


radarhere:
Travis and AppVeyor both pass the changes that you've made, so no problem!

Just for discussion, I'll say that a simpler version of this change is in #1891. Clearly, this PR has put more thought into the change however.

PIL/Image.py
```diff
@@ -586,14 +586,17 @@ def _dump(self, file=None, format=None):
...
-        f = (self.readonly == other.readonly)
-        g = (self.tobytes() == other.tobytes())
-        return a and b and c and d and e and f and g
+
+        properties = ['mode', 'size', 'info', 'category', 'readonly']
+
+        props_eq = all(self.__getattribute__(p) == other.__getattribute__(p)
+                       for p in properties)
+        if not props_eq:
+            return False
```
hugovk: The build "failure" is from coverage, [because this `return False` line isn't hit](https://coveralls.io/builds/6660498/source?filename=PIL%2FImage.py#L595) by any tests.

Not a huge thing, but would be good to make sure this is really working for all Python versions. Please could you add a unit test that should return false here when comparing two images? 

Somewhere in here: https://github.com/python-pillow/Pillow/blob/master/Tests/test_image.py

PIL/Image.py
```diff
@@ -586,14 +586,17 @@ def _dump(self, file=None, format=None):
...
+
+        funcs_eq = (self.getpalette() == other.getpalette()) \
+                   and (self.tobytes() == other.tobytes())
```
hugovk: Minor [pep8](https://travis-ci.org/python-pillow/Pillow/jobs/138750658#L4694) thing:
598: E127 continuation line over-indented for visual indent

Also I don't think it needs the `\` because it's all inside brackets.

hugovk: I've not timed it, but I imagine `tobytes` is the slowest thing here and a further optimisation could be:
```python
if self.getpalette() != other.getpalette():
    return False

return self.tobytes() == other.tobytes()
```

hugovk: Actually, `x and y` is already a short-circuit operation (in both [Py2](https://docs.python.org/2/library/stdtypes.html#boolean-operations-and-or-not) and [Py3](https://docs.python.org/3/library/stdtypes.html#boolean-operations-and-or-not)) so how about having an `__eq__` like in PR https://github.com/python-pillow/Pillow/pull/1891 but with `getpalette()` second to last and `tobytes()` last?

mattBoros: Sure! That sounds good.

Issue #1955, opened 2016-06-15
Expose registered file extensions


lggomez:
Is there any way of getting the file extensions supported by the library?
Usually it's a good practice to expose the supported types or extensions, specially in libraries that handle files, so the consumer can have a way to obtain or check its target extensions based on the ones provided by the library

I'm not very familiar with Pillow's code but this method in Image.py should do the trick:

``` python
def registered_extensions():
    """
    Returns a dictionary containing all file extensions belonging
    to registered plugins
    """
    if not bool(EXTENSION):
        init()
    return EXTENSION
```

Anyway, it's just a suggestion so you shall take it with a grain of salt. Keep up the good work on this fork

radarhere:
From my understanding, `preinit()` should not be necessary when immediately followed by `init()`. It's just a version of `init()` that aims to be faster by only covering the most popular formats.

lggomez:
Indeed, after checking I see that using init() will be enough, I corrected it

Do you consider this a worthy addition? If so I can make the pull request

radarhere:
Potentially, yes. I won't be the one to ultimately merge it, but you should certainly feel free to create a PR as a way of moving this forward.

radarhere:
This has been done in #2343.

PR #1968, opened 2016-06-20
Expose registered file extensions in Image

lggomez:
This request applies the proposal made in #1955

- Add a new method in Image (registered_extensions) that exposes the
  internal EXTENSION dictionary to consumers of the library

### DIFF
--- a/Tests/test_image.py
+++ b/Tests/test_image.py
@@ -184,6 +184,24 @@ class TestImage(PillowTestCase):
         img_colors = sorted(img.getcolors())
         self.assertEqual(img_colors, expected_colors)
 
+    def test_registered_extensions_uninitialized(self):
+        # Act
+        ext = Image.registered_extensions()
+
+        # Assert
+        self.assertEqual(bool(ext), True)
+
+    def test_registered_extensions(self):
+        # Arrange
+        # Open an image to trigger plugin registration
+        Image.open('Tests/images/rgb.jpg')
+
+        # Act
+        ext = Image.registered_extensions()
+
+        # Assert
+        self.assertEqual(bool(ext), True)
+
     def test_effect_mandelbrot(self):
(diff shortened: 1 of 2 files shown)


hugovk:
@lggomez Thanks for the PR. Please could you add a unit test to make sure this behaves sensibly?

See https://github.com/python-pillow/Pillow/blob/master/Tests/test_image.py

aclark4life:
@hugovk Does the unit test added suffice? looks like there is still an issue with coveralls, though I can't tell what it is from look at this: https://coveralls.io/builds/6670496 (other than coverage decreased).

Tests/test_image.py
```diff
@@ -184,6 +184,24 @@ def test_alpha_composite(self):
         img_colors = sorted(img.getcolors())
         self.assertEqual(img_colors, expected_colors)
 
+    def test_registered_extensions_uninitialized(self):
+        # Act
+        ext = Image.registered_extensions()
+
+        # Assert
+        self.assertEqual(bool(ext), True)
```
hugovk: Can use `self.assertTrue(bool(ext))` here and below instead.

Tests/test_image.py
```diff
@@ -184,6 +184,24 @@ def test_alpha_composite(self):
...
+    def test_registered_extensions(self):
+        # Arrange
+        # Open an image to trigger plugin registration
+        Image.open('Tests/images/rgb.jpg')
+
+        # Act
+        ext = Image.registered_extensions()
+
+        # Assert
+        self.assertEqual(bool(ext), True)
```
hugovk: Can you also test some actual expected values in `ext`? (Also above.)

hugovk:
This has been merged in https://github.com/python-pillow/Pillow/pull/2343.

Issue #471, opened 2014-01-05
Support for reading PDFs


mgalgs:
Are there any plans to add read support for the PDF format?

radarhere:
Just throwing the idea out there - rather than Ghostscript, this could also be done using ImageMagick, which has a more permissive license from my reading.

It's also worth noting that a consequence of the EPS conversion method would be that transparency is lost.

radarhere:
> It looks like it could be hacked in through ghostscript and the EpsImagePlugin without too much effort.

Just to note this here - I previously attempted to add support for opening PDFs, and [encountered resistance](https://github.com/python-pillow/Pillow/pull/1969#issuecomment-238076315) to the idea of using other software to convert PDFs to another format. So perhaps this requires a more thorough approach.

radarhere:
It is possible to use Pillow to extract images from a PDF - https://github.com/python-pillow/Pillow/issues/3491#issuecomment-2671268101

PR #1969, opened 2016-06-20
Added PDF read for macOS

radarhere:
OS X has a command line tool called 'sips' installed by default. It is able to convert the first page of a PDF to an image. So this PR adds PDF read functionality for OS X by converting a PDF into a temporary file and then loading the file as a PngImageFile. Not quite sure about this approach, so I thought I'd put it up here for feedback before adding anything like documentation or tests.

This could be considered a partial implementation of #471

### DIFF
--- a/PIL/PdfImagePlugin.py
+++ b/PIL/PdfImagePlugin.py
@@ -23,6 +23,12 @@
 from PIL import Image, ImageFile
 from PIL._binary import i8
 import io
+import sys
+if sys.platform == "darwin":
+    from PIL._util import isPath
+    import os
+    import tempfile
+    import subprocess
 
 __version__ = "0.4"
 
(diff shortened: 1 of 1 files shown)


PIL/PdfImagePlugin.py
```diff
@@ -247,9 +253,37 @@ def write(self, value):
...
+        os.write(fh, fp.read())
+        os.close(fh)
+    with open(os.devnull, 'w') as fp:
+        subprocess.call(['sips', '-s', 'format', 'png', filename, '--out', filepath], stdout=fp)
+
+    if os.stat(filepath).st_size == 0:
+        raise SyntaxError("PDF file could not be converted")
+
+    im = Image.open(filepath)
+    im.load()
```
homm: Why we need this?

homm: Ah, ok. Because we are deleting source file.

homm:
In my opinion, this is not PDF image files reader, Instead this is PDF converter. Of course, it can be very useful, but I don't think that Pillow is the best place for this code. In my opinion it is better to create separate package.

aclark4life:
@homm What do you think about #471 in general?

aclark4life:
I'd probably prefer a more generic approach to reading PDFs (i.e. cross plat, not macOS only)

homm:
@aclark4life I'm for adding PDF reading, but not with such approach.

radarhere:
Sure. Feel free to close. The thinking was just that perhaps adding functionality for one platform was better than no functionality.

aclark4life:
Thanks

PR #1970, opened 2016-06-20
Ignore large text blocks in PNG if LOAD_TRUNCATED_IMAGES is enabled

homm:
Sometimes large text block is just large text block while the image is still valid. We can safely ignore this blocks and still read the image. Because we are losing some information, I do this only when the `LOAD_TRUNCATED_IMAGES` flag is enabled.

### DIFF
--- a/PIL/PngImagePlugin.py
+++ b/PIL/PngImagePlugin.py
@@ -311,6 +311,11 @@ class PngStream(ChunkStream):
                               comp_method)
         try:
             icc_profile = _safe_zlib_decompress(s[i+2:])
+        except ValueError:
+            if ImageFile.LOAD_TRUNCATED_IMAGES:
+                icc_profile = None
+            else:
+                raise
         except zlib.error:
             icc_profile = None  # FIXME
         self.im_info["icc_profile"] = icc_profile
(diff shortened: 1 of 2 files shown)


wiredfool:
I'm not clear that it's appropriate to overload the LOAD_TRUNCATED_IMAGES flag with ignoring the protection against large text blocks. In this patch, we'd be purposefully ignoring data, rather than trying to scavenge more even though we're getting errors.

homm:
> with ignoring the protection against large text blocks

This patch doesn't ignore the protection. It ignores text blocks. It's absolutely different. As before, `LOAD_TRUNCATED_IMAGES` means "load the image despite of decoding errors". This doesn't introduce any new memory issues, because compressed block is already in the memory and decompressed has not allocated.

We don't have audible API for loading control. The `LOAD_TRUNCATED_IMAGES` global variable is currently the only way to affect the decoding somehow. In my application users upload valid PNG images with large icc profiles or text blocks which are not gzip bomb. I want to process this images as usual, but I don't want to increase `MAX_TEXT_CHUNK`. Actually I prefer to decrease it, because I don't need this chunks at all and still be able to process user images. But even when users are upload valid images with gzip bombs, why we shouldn't be able to process valid images just ignoring the bomb?

wiredfool:
Ok, fair enough.

Issue #630, opened 2014-04-18
Bug: "IOError: cannot identify image file" with some JPG images


cepreu2github:
This 3 images always produce error "IOError: cannot identify image file", when you try PIL.Image.open() on them.
There not good images. djpeg says "Corrupt JPEG data: 1 extraneous bytes before marker 0xdb", when processing them.
But many other libraries and applications handle this images without any errors (java.awt.image.BufferedImage, GIMP, Chrome, GPicView, Firefox etc...). I suppose, that Pillow must process them successfully also.

![2490631_original](https://cloud.githubusercontent.com/assets/4133611/2738437/b55a518c-c68e-11e3-9b67-0c6bc3cc2256.jpg)
![2490632_original](https://cloud.githubusercontent.com/assets/4133611/2738438/b55f126c-c68e-11e3-883e-4e4d1782a84d.jpg)
![2490633_original](https://cloud.githubusercontent.com/assets/4133611/2738439/b5604150-c68e-11e3-845b-774c4de800d9.jpg)

PR #1977, opened 2016-06-22
Skip extra 0xff00 in jpeg

kkopachev:
Some jpeg images might contain 0xFF00 bytes. Currently code just raises an error in this case. Example image is attached to PR and used in test.
imagemagick are able to read and show this file just right, with warning:

> identify Tests/images/jpeg_ff00_header.jpg 
Tests/images/jpeg_ff00_header.jpg JPEG 300x225 300x225+0+0 8-bit sRGB 86.5KB 0.000u 0:00.009
identify: Corrupt JPEG data: 4 extraneous bytes before marker 0xdb `Tests/images/jpeg_ff00_header.jpg' @ warning/jpeg.c/JPEGWarningHandler/352.

I looked up some other decoders and they just skip 0xff00:
https://golang.org/src/image/jpeg/reader.go L562
http://geezer.osdevbrasil.net/software/dumpjpeg.c (search for 0xFF00)
https://github.com/eugeneware/jpeg-js/blob/master/lib/decoder.js#L604

Related issue: https://github.com/python-pillow/Pillow/issues/630

I can not share original image with which this issue occurred, so I modified flower2.jpg from Pillow's test suite.


cerber717:
👍

kkopachev:
Test failures seems to be not related to the changes.

willtheorangeguy:
Sounds good. I have been having a problem too.

kkopachev:
Would be great if this could be included in next release

homm:
What is the license of the test image?

kkopachev:
I took [image from existing tests ](https://github.com/python-pillow/Pillow/blob/master/Tests/images/flower2.jpg) and added 4 bytes. Not sure of a license of original. I assume new image will have same license.

PR #1986, opened 2016-06-25
Fix issue converting P mode to LA

### LINKED ISSUE #1979: Converting palette mode to grayscale+alpha deforms image
### What did you do?
Convert a png/gif image in "P" (palette) mode to "LA" (grayscale + alpha).
```python
from PIL import Image

image = Image.open('palette_image.png')
print('original mode: %s' % image.mode)

# converting straight to grayscale+alpha leads to deformed image
image = image.convert('LA')
image.save('out.la.png')
```

palette_image
palette_image.png

I would like to get something like this:
out rgba la
out.rgba.la.png

### What actually happened?
I got this:
out la
out.la.png

python==2.7.11
Pillow==3.2.0
bug also occurs with pillow master

### Workaround
Convert to RGBA before converting to LA.
```python
from PIL import Image

image = Image.open('palette_image.png')
print('original mode: %s' % image.mode)

# converting to RGBA first as a workaround
image = image.convert('RGBA')
image = image.convert('LA')
image.save('out.rgba.la.png')
```

didrix:
Fixes #1979.

* Convert P to LA without deforming image

### DIFF
--- a/Tests/test_image_convert.py
+++ b/Tests/test_image_convert.py
@@ -123,6 +123,26 @@ class TestImageConvert(PillowTestCase):
         self.assertNotIn('transparency', p.info)
         p.save(f)
 
+    # ref https://github.com/python-pillow/Pillow/issues/1979
+
+    def test_p_la(self):
+        def L(rgb):
+            return int((rgb[0] * 299 + rgb[1] * 587 + rgb[2] * 114) / 1000)
+
+        def convert_to_gray(im, pixel):
+            if im.mode == 'P':
+                color_id = im.getpixel(pixel)
+                color = im.getpalette()[color_id * 3:(color_id+1)*3]
+                return L(color)
+            elif im.mode == 'LA':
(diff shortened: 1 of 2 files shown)


wiredfool:
Could you add a test for this, either using the image from the issue or one of selection of hopper images we have in the Tests/images directory.

didrix:
@wiredfool test added.

Tests/test_image_convert.py
```diff
@@ -123,6 +123,26 @@ def test_trns_RGB(self):
         self.assertNotIn('transparency', p.info)
         p.save(f)
 
+    # ref https://github.com/python-pillow/Pillow/issues/1979
+
+    def test_p_la(self):
+        def L(rgb):
+            return int((rgb[0] * 299 + rgb[1] * 587 + rgb[2] * 114) / 1000)
```
homm: You just need to check for alpha channel, right? No need to check conversion to L itsef.

Issue #1859, opened 2016-04-23
Manylinux wheel recipe


matthew-brett:
I've been working on making a recipe for building [manylinux](https://www.python.org/dev/peps/pep-0513/) wheels for Pillow.

As y'all may have seen, Manylinux is a specification and a matching docker image that allows us to build binary wheels for Linux that we expect to run on almost any Linux system (see the spec above).

As of pip version 8.1, pip will install wheels with the `manylinux1` tag.  There are manylinux wheels for numpy, scipy and cython already uploaded to pypi.

I made a recipe for building manylinux wheels for Pillow - here : https://github.com/matthew-brett/manylinux-builds/blob/master/build_pillows.sh

Does that recipe look right?  It passes all tests on a simple test run I did for Python 2.7 and Pillow 3.2.

I just uploaded the built wheels to a testing server at https://nipy.bic.berkeley.edu/manylinux

Y'all can test with:

```
# Get latest pip
python -m pip install --upgrade pip
pip install -f https://nipy.bic.berkeley.edu/manylinux pillow
```

radarhere:
Curious, how did you arrive at that list of Pillow versions? It skips a few, and I'm just wondering why.

matthew-brett:
No very serious reason - I was just trying to replicate the versions for the wheels already built and used for travis-ci Ubuntu 12.04 installs over at http://travis-wheels.scikit-image.org/  .  It would be easy to add all the versions available on pypi.

matthew-brett:
I put up logs of running the tests on all the built wheels as installed onto a Debian sid machine.

Test script: http://nipy.bic.berkeley.edu/manylinux/tests/test_pillow.py
Logs:

http://nipy.bic.berkeley.edu/manylinux/tests/27_pillow.log
http://nipy.bic.berkeley.edu/manylinux/tests/34_pillow.log
http://nipy.bic.berkeley.edu/manylinux/tests/35_pillow.log

AbdealiJK:
I was wondering if this is planned to be implemented soon ? Or if an ETA could be provided ?

matthew-brett:
We have a recipe up as a pull-request here : https://github.com/python-pillow/pillow-wheels/pull/36

That repo does build good 64-bit wheels, but there are some test errors for 32-bit wheels.

[...]

PR #1990, opened 2016-06-28
test_imagecms: Reduce precision of extended info

AbdealiJK:
Fixes (partially) https://github.com/python-pillow/Pillow/issues/1859 .

 * Reduce precision in some tests that seem to be failing in 32bit.

In 32bit computers(When making wheels) it was found that the
precision of the extended information in test_extended_information() was
found to be lesser than what was being tested. This commit adds a helper
function which reduces the precision using `round()` and testing the
rounded numbers rather than the original numbers.

### DIFF
--- a/Tests/test_imagecms.py
+++ b/Tests/test_imagecms.py
@@ -260,9 +260,19 @@ class TestImageCms(PillowTestCase):
         o = ImageCms.getOpenProfile(SRGB)
         p = o.profile
 
+        def assert_truncated_tuple_equal(tup1, tup2, digits=10):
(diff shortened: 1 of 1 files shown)


AbdealiJK:
I was initially using `round(val, 10)` but it seems like the behaviour of `round()` is a little different in different jobs - https://travis-ci.org/python-pillow/Pillow/builds/140821124

So, I stuck to the old fashioned `int(val * 1e10) / 1e10`.
http://stackoverflow.com/questions/783897/truncating-floats-in-python gave a lot of interesting options, but this seemed the simplest and easiest.

PR #1991, opened 2016-06-29
Ability to skip crc checks for ancillary chunks

kkopachev:
I encountered that some PNG files could not be open with Pillow since  they have crc mismatch for iTXt chunks. 
Browsers and imagemagick can open them just fine (though imagemagick shows a warning)
libpng defines a way to set it's behavior to throw an error or warning for ancillary chunks.

So I am proposing use of `LOAD_TRUNCATED_IMAGES` to skip crc checks for ancillary chunks. No need to calculate if we're going to ignore an error.

### DIFF
--- a/PIL/PngImagePlugin.py
+++ b/PIL/PngImagePlugin.py
@@ -140,7 +140,7 @@ class ChunkStream(object):
 
         # Skip CRC checks for ancillary chunks if allowed to load truncated images
         # 5th byte of first char is 1 [specs, section 5.4]
-        if ImageFile.LOAD_TRUNCATED_IMAGES and (ord(cid[0]) >> 5 & 1):
+        if ImageFile.LOAD_TRUNCATED_IMAGES and (i8(cid[0]) >> 5 & 1):
             self.crc_skip(cid, data)
             return
 
--- a/PIL/PngImagePlugin.py
+++ b/PIL/PngImagePlugin.py
@@ -138,6 +138,12 @@ class ChunkStream(object):
     def crc(self, cid, data):
         "Read and verify checksum"
 
+        # Skip CRC checks for ancillary chunks if allowed to load truncated images
+        # 5th byte of first char is 1 [specs, section 5.4]
+        if ImageFile.LOAD_TRUNCATED_IMAGES and (ord(cid[0]) >> 5 & 1):
+            self.crc_skip(cid, data)
+            return
+
         try:
             crc1 = Image.core.crc32(data, Image.core.crc32(cid))
             crc2 = i16(self.fp.read(2)), i16(self.fp.read(2))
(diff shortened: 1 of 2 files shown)


PIL/PngImagePlugin.py
```diff
@@ -138,6 +138,12 @@ def call(self, cid, pos, length):
     def crc(self, cid, data):
         "Read and verify checksum"
 
+        # Skip CRC checks for ancillary chunks if allowed to load truncated images
+        # 5th byte of first char is 1 [specs, section 5.4]
+        if ImageFile.LOAD_TRUNCATED_IMAGES and (i8(cid[0]) >> 5 & 1):
```
homm: I'd prefer to ignore the error after it occurs. Of course, it is meaningless to do the check, if we are going to accept the result in any case. But in the future, we can add some warnings or other decoder info.

homm:
Does `libpng` and friend open images with error in data chunks?

PIL/PngImagePlugin.py
kkopachev: Hm, why bother doing this now, if in both cases code has to be changed to throw a warning?
But anyway, I can keep crc check if you'd like. Would you like to ignore checksum mismatch or incomplete checksum as well?

kkopachev:
From what I can tell, in `libpng` user of it can ask to throw errors for crc corruptions in any chunks, or just ancillary or all of them. Not sure what happens if it encounters errors in data chunks, though. My thought here was that since Pillow does not expose any methods to change decoder behavior, but only LOAD_TRUNCATED_IMAGES, using that flag would be pretty safe bet to let read files at least with ancillary chunks broken.
Not sure where my original image come from, but looks like there was an issue with `libpng` where it sometimes could produce invalid CRC for iTXt chunk.

Tests/test_file_png.py
```diff
@@ -319,6 +320,35 @@ def test_verify_struct_error(self):
...
 
+    def test_verify_ignores_crc_error(self):
+        # check ignores crc errors in ancillary chunks
+
+        chunk_data = chunk(b'tEXt', b'spam')
+        broken_crc_chunk_data = chunk_data[:-1] + b'q'  # break CRC
+
+        image_data = HEAD + broken_crc_chunk_data + TAIL
+        self.assertRaises(SyntaxError,
+                          lambda: PngImagePlugin.PngImageFile(BytesIO(image_data)))
```
homm:
```python
self.assertRaises(SyntaxError, PngImagePlugin.PngImageFile, BytesIO(image_data))
```

kkopachev: Fixed. FYI there are few more lambda usages in that file in same context.

homm:
I haven't met such images in a wild. But I'm always for decoding more with `LOAD_TRUNCATED_IMAGES`.

PR #1995, opened 2016-06-30
Move test_j2k_overflow to check_j2k_overflow

wiredfool:
It is a DOS on 32bit machines


homm:
Oh, thank you! Could you also look at `LargeMemoryNumpyTest` and `LargeMemoryTest`?

wiredfool:
I don't think those two tests run by default. They're not showing up in the travis runs anyway.

wiredfool:
(note that I've cancelled the tests, because we're backed up and these don't run anyway)

PR #2007, opened 2016-07-01
Correct ValueError

### LINKED ISSUE #2006: ValueError when getting tiff tags
### What did you do?

I tried to extract metadata from a .tif file

get the metadata
### What actually happened?

ValueError: too many values to unpack (expected 1)

pillow 3.3.0
python 3.5

Please include code that reproduces the issue and whenever possible, an image that demonstrates the issue. The best reproductions are self-contained scripts with minimal dependencies.

```
from PIL import Image
img = Image.open("myfil.tif")
dict(img.tag)
```

impact27:
Fixes #2006.

When `info.length` is 1 and `len(values) > 1`, the code used to throw: 

```
ValueError: too many values to unpack (expected 1)
```

For example: tag `IMAGEJ_META_DATA_BYTE_COUNTS` (50838) can sometimes have two values (I have .tif files with metadata split between UTF8 and UTF16 => one byte count for each)

### DIFF
--- a/PIL/TiffImagePlugin.py
+++ b/PIL/TiffImagePlugin.py
@@ -550,7 +550,7 @@ class ImageFileDirectory_v2(collections.MutableMapping):
 
         dest = self._tags_v1 if legacy_api else self._tags_v2
 
-        if info.length == 1:
+        if info.length > 0 and len(values) == 1:
             if legacy_api and self.tagtype[tag] in [5, 10]:
                 values = values,
             dest[tag], = values


impact27:
Maybe it would be better to leave info.length == 1 instead of info.length > 0?

wiredfool:
Good idea, but I think it's the wrong approach. I think the implied test case is useful/necessary, but we need to treat the case of getting a different type from the file than the spec differently than this patch.

PR #2008, opened 2016-07-02
Resampling lookups, trailing empty coefficients, precision

homm:
### Performance

* The saturated lookup table is used for truncating channels values to `UINT8` before output instead of two if operators.
* Coefficients with zero weights are truncated from coefficients table.

Both changes affect mostly the cases with short coefficients rows length and when the channels truncating contributes significantly compared to convolution itself. I.e. changes affect upscaling and reasonable downscaling.

Source    | Operation           | Filter  | Before | After | Speedup
----------|---------------------|---------|--------|-------|--------
2560×1600 | Resize to 1920x1200 | Bilinear|  55.87 | 73.01 | 30%
          |                     | Bicubic |  43.64 | 52.96 | 21%
          |                     | Lanczos |  34.51 | 40.34 | 17%
          | Resize to 7712x4352 | Bilinear|   6.71 |  8.89 | 32%
          |                     | Bicubic |   5.51 |  6.95 | 26%
          |                     | Lanczos |   4.62 |  5.51 | 19%

### Precision

Negative coefficients used to round up. This affected only very large coefficient rows, i.e. significant downscaling.

First error in the test:
Fixed point without fix: **42 542**
Fixed point with fix: **999 534**
Float coefficients: **2 840 156**


homm:
Failed tests are not related to resizing at all. There are two bugs: in PyAccess module on CPython and in PyAccess tests. I've fixed tests (PR impending), looking at the module.

homm:
Now green. Fixes in https://github.com/python-pillow/Pillow/pull/2009

wiredfool:
Merged in #2091

Issue #2005, opened 2016-07-01
Pillow shouldn't discard alpha channel while saving RGBA as JPEG


homm:
We shouldn't allow to save `RGBA` images as JPEG and discard the alpha channel. This can be convenient in some cases, but in most cases this is confusing. See #1345. Explicit is better than implicit.

```python
In [8]: Image.open('../i/grayscale-alpha.png').resize((200, 200), Image.BICUBIC).save('../i/grayscale-alpha.jpg')
IOError: cannot write mode LA as JPEG

In [9]: Image.open('../i/color-alpha.png').resize((200, 200), Image.BICUBIC).save('../i/color-alpha.jpg')
```

PR #2010, opened 2016-07-03
Show warning when trying to save RGBA image as JPEG

homm:
Fixes https://github.com/python-pillow/Pillow/issues/2005

### DIFF
--- a/PIL/JpegImagePlugin.py
+++ b/PIL/JpegImagePlugin.py
@@ -534,7 +534,6 @@ RAWMODE = {
     "1": "L",
     "L": "L",
     "RGB": "RGB",
-    "RGBA": "RGB",
     "RGBX": "RGB",
     "CMYK": "CMYK;I",  # assume adobe conventions
     "YCbCr": "YCbCr",
--- a/PIL/JpegImagePlugin.py
+++ b/PIL/JpegImagePlugin.py
@@ -534,6 +534,7 @@ RAWMODE = {
     "1": "L",
     "L": "L",
     "RGB": "RGB",
+    "RGBA": "RGB",
     "RGBX": "RGB",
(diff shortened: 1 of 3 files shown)


wiredfool:
The problem here is that it's liable to break thumbnailing scripts that read PNGs, and this behavior dates from the PIL fork. (so it's well established).  We'd be replacing something that does questionable things to the alpha channel with an exception.

homm:
PNG have [16 base color formats](http://www.schaik.com/pngsuite/pngsuite_bas_png.html). 6 of them fails PNG → JPEG conversation on master:

```python
from PIL import Image
for fn in glob('../i/basn*.png'):
    try:
        Image.open(fn).save(fn + '.jpg')
    except IOError:
        print '>>>', fn
```

They are:
```
>>> ../i/basn0g16.png
>>> ../i/basn3p01.png
>>> ../i/basn3p02.png
>>> ../i/basn3p04.png
>>> ../i/basn3p08.png
>>> ../i/basn4a08.png
```

We can leave the alpha channel discarding for now, but I afraid this means that we leave it forever.

wiredfool:
We can make it a specific deprecation warning, and anyone who wants to can register that as an exception. Then we can switch it over in 2-4 releases. (probably after stretch freezes, or April '17)

homm:
Ok, I'll add warning then.

Sorry, what is stretch freezes?

wiredfool:
Debian Stretch, the next stable release. I'm leaning to include a deprecation warning in one iteration of Debian stable before making a hard error.

homm:
Ok, caught your point.

homm:
Turned the error to the warning.

wiredfool:
This should probably get a release note as a future deprecation warning.

PR #2011, opened 2016-07-03
Add a scale function to expand or contract a PIL image by a factor 

vlmath:
Add the scale function to provide an easy way to expand and contract image by a certain factor given in parameter.
This function "only" calculate the target size, and call the PIL.Image.Image.resize() function.

### DIFF
--- a/PIL/Image.py
+++ b/PIL/Image.py
@@ -1547,6 +1547,26 @@ class Image(object):
 
         return self._new(self.im.resize(size, resample))
 
+    def scale(self, factor, resample = NEAREST):
+        """
+        Returns a rescaled image by a specific factor given in parameter.
+        A factor greater than 1 expands the image, between 0 and 1 contracts the
+        image.
+
+        :param factor: The expansion factor, as a float.
+        :param resample: An optional resampling filter. Same values possible as
+           in the PIL.Image.resize function.
+        :returns: An :py:class:`~PIL.Image.Image` object.
+        """
+        if factor == 1:
+            return self._new(self.im)
+        elif factor <= 0:
+            raise ValueError("the factor must be greater than 0")
+        else:
+            size = (int(round(factor * self.width)),
+                    int(round(factor * self.height)))
+            return self.resize(size, resample)
+
     def rotate(self, angle, resample=NEAREST, expand=0):
         """
         Returns a rotated copy of this image.  This method returns a
--- a/PIL/Image.py
+++ b/PIL/Image.py
@@ -1547,26 +1547,6 @@ class Image(object):
 
         return self._new(self.im.resize(size, resample))
 
-    def scale(self, factor, resample = NEAREST):
-        """
-        Returns a rescaled image by a specific factor given in parameter.
-        A factor greater than 1 expands the image, between 0 and 1 contracts the
-        image.
-
-        :param factor: The expansion factor, as a float.
-        :param resample: An optional resampling filter. Same values possible as
-           in the PIL.Image.resize function.
-        :returns: An :py:class:`~PIL.Image.Image` object.
-        """
-        if factor == 1:
-            return self._new(self.im)
-        elif factor <= 0:
-            raise ValueError("the factor must be greater than 0")
-        else:
-            size = (int(round(factor * self.width)),
-                    int(round(factor * self.height)))
-            return self.resize(size, resample)
-
     def rotate(self, angle, resample=NEAREST, expand=0):
         """
         Returns a rotated copy of this image.  This method returns a
--- a/PIL/ImageOps.py
+++ b/PIL/ImageOps.py
@@ -178,6 +178,27 @@ def crop(image, border=0):
         )
 
 
+def scale(self, factor, resample=Image.NEAREST):
+        """
+        Returns a rescaled image by a specific factor given in parameter.
+        A factor greater than 1 expands the image, between 0 and 1 contracts the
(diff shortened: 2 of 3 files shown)


homm:
I do not find this feature very useful. This can be solved in one line without extending API.

vlmath:
I'am using this kind of function every day, and dont like to make two time the sames calculations (new width and new heigth calculation) in my code, but ok, as you think. I will overload the Image.Image class I think instead ^^

PIL/Image.py
```diff
@@ -1547,6 +1547,26 @@ def resize(self, size, resample=NEAREST):
 
         return self._new(self.im.resize(size, resample))
 
+    def scale(self, factor, resample = NEAREST):
```
homm: Spaces around "=".

homm:
Anyway, tests required to make coveralls happy.

wiredfool:
Take a look at `ImageOps.fit`.  I think it's a more complete version of this.

homm:
`ImageOps.fit` operates different, but I think `ImageOps` module is the best place for `scale`.

vlmath:
To understand clearly, should I move it, make the tests, or it will not me merged because because it's not very useful ? Not sure to understand correctly.

homm:
For `ImageOps` module this is quite reasonable. I am just against the blowing of the core Image API.

vlmath:
OK, no problems.
Since it's my first contribution in opensource, I have to remake a merge request with the commits on ImageOps, right ? Thank's

homm:
You can just move the function in subsequent commits in the same request.

PIL/ImageOps.py
```diff
@@ -178,6 +178,27 @@ def crop(image, border=0):
         )
 
 
+def scale(self, factor, resample=Image.NEAREST):
+        """
```
homm: Indention should be 4 spaces, not 8.

Tests/test_imageops.py
```diff
@@ -78,6 +78,22 @@ def test_pil163(self):
         ImageOps.equalize(i.convert("P"))
         ImageOps.equalize(i.convert("RGB"))
 
+    def test_scale(self):
+        # Test the scaling function
+        i = hopper("L").resize((50,50))
```
homm: Should be one space after the comma in arguments list. There and below.

Tests/test_imageops.py
```diff
@@ -78,6 +78,22 @@ def test_pil163(self):
         ImageOps.equalize(i.convert("P"))
         ImageOps.equalize(i.convert("RGB"))
 
+    def test_scale(self):
+        # Test the scaling function
+        i = hopper("L").resize((50, 50))
+        
+        with self.assertRaises(ValueError):
+            ImageOps.scale(i,-1)
```
homm: Here and below too.

PIL/ImageOps.py
```diff
@@ -178,6 +178,27 @@ def crop(image, border=0):
         )
 
 
+def scale(self, factor, resample=Image.NEAREST):
```
wiredfool: Should we default to quality rather than speed here?

wiredfool:
Lgtm

PIL/ImageOps.py
vlmath: We should. I use this default value to be coherent with the Image.resize function. For me, all resample default parameter should be Bicubic or bilinear

PIL/ImageOps.py
```diff
@@ -178,6 +178,27 @@ def crop(image, border=0):
...
+    A factor greater than 1 expands the image, between 0 and 1 contracts the
+    image.
+
+    :param factor: The expansion factor, as a float.
+    :param resample: An optional resampling filter. Same values possible as
+       in the PIL.Image.resize function.
+    :returns: An :py:class:`~PIL.Image.Image` object.
+    """
+    if factor == 1:
+        return self._new(self.im)
```
radarhere: I applaud the thought, but `resize` has it's own check for the image being the same size - https://github.com/python-pillow/Pillow/blob/master/PIL/Image.py#L1539 - so I'm going to suggest that this is redundant? Feel free to disagree.

vlmath: I agree with you. I think it's faster to return the value here rather make the calculations of the destination size, pass the relay to the resize function, and then return the value because factor = 1.
Not tested in practice however.

wiredfool: Actually, that leads to the issue that were not in the image object, so no self, do the def lone is wrong as well

radarhere: @vlmath yeah, it's just a question of lines of code vs speed - you're probably fine, I guess I just felt the need to put that out there.

PIL/ImageOps.py
```diff
@@ -178,6 +178,27 @@ def crop(image, border=0):
...
+    A factor greater than 1 expands the image, between 0 and 1 contracts the
+    image.
+
+    :param factor: The expansion factor, as a float.
+    :param resample: An optional resampling filter. Same values possible as
+       in the PIL.Image.resize function.
+    :returns: An :py:class:`~PIL.Image.Image` object.
+    """
+    if factor == 1:
+        return Image.Image._new(image.im)
```
homm: `_new` is not a static method. You need to call it on the instance: `image._new()`.
Or you can use public method: `return image.copy()`.
Or you can remove this branch as @radarhere suggested.

PR #2012, opened 2016-07-03
Remove "end of file" banner from all files

homm:


hugovk:
Not a big deal, but I personally find these useful.

wiredfool:
What do you find useful about it?

hugovk:
Glanceability. If you see it on the screen you know immediately you're at the end of the file, there's no more code to follow.

homm:
@hugovk Yea, but how this helps you? File ends regardless of the comment.

screen shot 2016-07-04 at 12 00 36

For me for example it is absolutely clear that `root.mainloop()` is the last line of code. Maybe there is some UI problem in you text editor/IDE that you don't see the end of file?

wiredfool:
Clearly he's not using vi...

hugovk:
It's like this in TextWrangler (and similar in Notepad++ and PhpStorm):

image


image

homm:
> and similar in Notepad++ and PhpStorm

In PyCharm it looks like this. I don't think that PhpStorm looks different.

screen shot 2016-07-04 at 15 31 56

Anyway, even if the last line is rendered on the bottom of the screen. Is it a real problem for you? Do you want to add such comments to all other files? ) PIL is the first project where I see something like this.

hugovk:
Yeah, I think you're right about PhpStorm. Anyway, TextWranger and Notepad++ are my main Python editors :) I have been adding it to other files, and if wanted, I could add to others. It's not a [convention unique to this project](https://github.com/search?q=%22end+of+file%22&type=Code&utf8=%E2%9C%93) but like I said, not a big deal, just something I personally find useful.

wiredfool:
I'd say merge

PR #2013, opened 2016-07-04
Docs: Add documentation for hacking on Pillow

mgalgs:
- New page in the docs for hacking on pillow

It's not clear from the existing documentation how to set up a local
environment for hacking on pillow.  Add some docs for this purpose so
that new contributors can hit the ground running more quickly.

### DIFF
--- a/docs/hacking.rst
+++ b/docs/hacking.rst
@@ -1,0 +1,31 @@
+===================
+ Hacking on Pillow
+===================
+
+Found a bug?  Want to submit a feature?  This page will help you get
+started with doing development on Pillow.
+
+First, get the source::
+
+    $ git clone https://github.com/python-pillow/Pillow.git
+    $ cd Pillow
+
+Install pre-requisits::
+
+    $ virtualenv env
+    $ . env/bin/activate
+    $ pip install nose
+
+Build the ``develop`` target::
+
+    $ python setup.py develop
+
+Check that all the tests pass in your environment::
+
+    $ nosetests -vx Tests/test_*.py
+
+Assuming all the tests pass successfully you're all set!  Go ahead and hack
+away.  When you're ready to submit your work back to the Pillow project,
+just `fork and submit a pull request
+<https://help.github.com/articles/using-pull-requests/>`_ as usual.  You
+will probably be asked to submit a new test along with your pull request.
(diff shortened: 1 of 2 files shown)


homm:
I'm not native English speaker, so my opinion is not so important, but personally I don't like the word "hacking". 

Also, I think the best place for this is [CONTRIBUTING.md](https://github.com/python-pillow/Pillow/blob/master/.github/CONTRIBUTING.md).

wiredfool:
FWIW, I really don't like the suggestion for the `develop` option to `setup.py`. It fails badly when you wind up using more than one python (e.g. 2.x, 3.x), which you have to do when developing on Pillow so as not to have a ton of follow up commits that fix support for 3.x, 3.2, and then 2.6. 

On unix, it's 
```
~/venv/bin/activate
make install && make test
```

With possibly make install-req in there as well. (but that installs a bunch of doc building stuff as well)

or 
``` 
~/venv/bin/activate
python setup.py install
python ./test-installed.py
```

I think that we're 95% of the way there in the installation docs, perhaps adding the default `python setup.py install ` bit instead of pip in the install section and something about test runners.

docs/hacking.rst
```diff
@@ -0,0 +1,31 @@
...
+
+Found a bug?  Want to submit a feature?  This page will help you get
+started with doing development on Pillow.
+
+First, get the source::
+
+    $ git clone https://github.com/python-pillow/Pillow.git
+    $ cd Pillow
+
+Install pre-requisits::
```
radarhere: requisites

docs/hacking.rst
```diff
@@ -0,0 +1,31 @@
...
+    $ python setup.py develop
+
+Check that all the tests pass in your environment::
+
+    $ nosetests -vx Tests/test_*.py
+
+Assuming all the tests pass successfully you're all set!  Go ahead and hack
+away.  When you're ready to submit your work back to the Pillow project,
+just `fork and submit a pull request
+<https://help.github.com/articles/using-pull-requests/>`_ as usual.  You
```
radarhere: This redirects to https://help.github.com/articles/about-pull-requests/

hugovk:
Closing this, it's out of date and hasn't been updated after review.

PR #2015, opened 2016-07-05
Speedup paste with masks to 20–25%

homm:
This speedups paste using any types of masks (third argument of `im.paste()` function).

Mask mode | Previous | This PR
-----|---------------|------------
RGBA |  92.16 Mpx/s  | 117.89 Mpx/s
RGBa | 131.62 Mpx/s  | 164.27 Mpx/s
L    |  83.17 Mpx/s  | 106.29 Mpx/s

This achieved only by loops unrolling. This is 100% compatible with the previous implementation because `pixelsize` is always 4 when `im->image8` is not set. [Explanation in Storage.c](https://github.com/python-pillow/Pillow/blob/master/libImaging/Storage.c#L250-L258).


homm:
I'll add tests for whole `Paste.c` later.

homm:
Bump

Tests/test_image_paste.py
```diff
@@ -0,0 +1,112 @@
...
+        for mode in ('L', 'RGB'):
+            im = Image.new(mode, (200, 200), 'red')
+            im2 = hopper(mode)
+
+            im.paste(im2, (0, 0))
+
+            im = im.crop((0, 0) + im2.size)
+            self.assert_image_equal(im, im2)
+
+    def test_image_mask_1(self):
```
hugovk: Other than checking these run without exception, is there something that can/should be explicitly asserted in this and the following tests?

Tests/test_image_paste.py
```diff
@@ -0,0 +1,112 @@
+from helper import unittest, PillowTestCase, hopper
```
hugovk: `unittest` imported but unused.

homm:
Tests are rewritten. All tests pass on the master, so changes are correct.

PR #2021, opened 2016-07-09
Converted documentation

radarhere:
Carrying on the work of #2018


wiredfool:
These need to be moved to docstrings to be picked up by sphinx.

radarhere:
Thanks. I've now converted them to docstrings.

Issue #1722, opened 2016-02-09
TiffImagePlugin: ValueError: need more than 0 values to unpack


edudar:
An attempt to get EXIF data for auto-rotation of http://images.dpchallenge.com/images_challenge/0-999/622/1200/Copyrighted_Image_Reuse_Prohibited_457512.jpg leads to the following error:
```
  File "/opt/anaconda2/lib/python2.7/site-packages/PIL/JpegImagePlugin.py", line 391, in _getexif
    return _getexif(self)
  File "/opt/anaconda2/lib/python2.7/site-packages/PIL/JpegImagePlugin.py", line 426, in _getexif
    exif = dict(_fixup_dict(info))
  File "/opt/anaconda2/lib/python2.7/site-packages/PIL/JpegImagePlugin.py", line 407, in _fixup_dict
    return dict([(k, _fixup(v)) for k, v in src_dict.items()])
  File "/opt/anaconda2/lib/python2.7/_abcoll.py", line 414, in items
    return [(key, self[key]) for key in self]
  File "/opt/anaconda2/lib/python2.7/site-packages/PIL/TiffImagePlugin.py", line 854, in __getitem__
    self._setitem(tag, handler(self, data, legacy), legacy)
  File "/opt/anaconda2/lib/python2.7/site-packages/PIL/TiffImagePlugin.py", line 534, in _setitem
    dest[tag], = values
ValueError: need more than 0 values to unpack
```

wiredfool:
The general problem here is that the count of values in the exif don't match the Tiff metadata spec as we have it encoded, and this portion barfs if the item doesn't match. 

We've fixed portions of this before, but clearly not this part.

mgalgs:
@edudar this seems to have been fixed by c50503a14d5b04dd42030adb77b05732e800b153 in #1988.  At least that fixed my issue that had a very similar symptom to this.  Care to check?

homm:
This is not fixed in #1988 Rather will be fixed in #2024.

hugovk:
PR https://github.com/python-pillow/Pillow/pull/2024 is merged, and will be released in Pillow 3.4.0 next week.

PR #2024, opened 2016-07-12
Skip empty values in ImageFileDirectory

homm:
I'm not sure is there any valid cases where we can skip value. Without this fix Pillow raises:

```
  File "/home/homm/Code/Pillow/Tests/test_file_tiff_metadata.py", line 220, in test_expty_values
    info = info.as_dict()
  File "/home/homm/Code/Pillow/PIL/TiffImagePlugin.py", line 481, in as_dict
    return dict(self)
  File "/home/homm/Code/Pillow/PIL/TiffImagePlugin.py", line 500, in __getitem__
    self[tag] = handler(self, data, self.legacy_api)  # check type
  File "/home/homm/Code/Pillow/PIL/TiffImagePlugin.py", line 514, in __setitem__
    self._setitem(tag, value, self.legacy_api)
  File "/home/homm/Code/Pillow/PIL/TiffImagePlugin.py", line 556, in _setitem
    dest[tag], = values
ValueError: need more than 0 values to unpack
```

First I considering returning None from `__getitem__`. But then I think what if an application expects exact value type (int, float, but not None). So for me, the better solution is just skip empty values.


homm:
Fixes #1722

homm:
@wiredfool any thoughts?

wiredfool:
Leaning to yes, but I want to look at all the tiff stuff in one pass when I can get it in my head.

Issue #2019, opened 2016-07-08
Memory is slowly leaking


dmpetrov:
Memory is slowly leaking when process many image files one by one.

Leak size:
100K images --> ~100MB
300K images --> ~200MB
1M images --> ~1GB

Here you can find reproducing code examples: https://github.com/PyWavelets/pywt/issues/180

hugovk:
@dmpetrov Please can you give a minimal code example reproducing the problem that uses Pillow but no third-party libraries like numpy or pywt? Thanks!

dmpetrov:
```
import os
import PIL
from PIL import Image

dir = "/Volumes/Seagate/storage/kaggle/avito/images"
onlyfiles = []
dirs = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14]
for d in dirs:
    filedir = os.path.join(dir, str(d))
    files = [f for f in os.listdir(filedir) if os.path.isfile(os.path.join(filedir, f))]
    filepaths = map( lambda x: os.path.join(filedir, x), files )
    onlyfiles += list(filepaths)

s = 0
for i in range(len(onlyfiles)):
    fname = onlyfiles[i]
    image = PIL.Image.open(fname)
    # Do something
    s += image.size[0]*image.size[1]

    if i % 10000 == 0:
        print('iteration ', i)

input("Press Enter to continue...")
```

Input data: image dataset (10M small images) from Kaggle competition https://www.kaggle.com/c/avito-duplicate-ads-detection

Result: It starts with 221MB for the file names.
500 images - 374MB
1M images - 510MB

gc() inside the loop did not help.

damaestro:
Do we know what release this started in? I'd like to pin to a lower release to mitigate any production impact without having to set memory scaling alarm/actions.

homm:
@dmpetrov I can't reproduce this on Ubuntu 14.04 with Python 2.7 on Pillow master. Could you tell your Pillow, Python, OS, libjpeg versions?

@damaestro This is not confirmed yet.

homm:
@dmpetrov I'v tried Python 2.7 and 3.4 under OS X. I can confirm the leak in 3.4.

homm:
So, this is a minimal test case, which I came to:

https://gist.github.com/homm/c6a79ce7b445f47a74a4d296742a5af8

As you can there is no Pillow at all! I'm pretty sure this is pure Python 3 memory leak.

[...]

PR #2032, opened 2016-07-17
Close files after loading

homm:
Draft for #2019

This PR closes files at least for files which we are opening from filename at least for not sequence formats. This breaks consistency across different codecs, though.

Is there any reliable place where we can close files for sequence formats?

### DIFF
--- a/PIL/DcxImagePlugin.py
+++ b/PIL/DcxImagePlugin.py
@@ -59,6 +59,7 @@ class DcxImageFile(PcxImageFile):
             self._offset.append(offset)
 
         self.__fp = self.fp
+        self._exclusive_fp = False
         self.seek(0)
 
     @property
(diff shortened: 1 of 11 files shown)


wiredfool:
Will have more to say later, but mostly afk for a week.  

As far as I know, the only place to deal with closing the file for sequences is in the explicit close o r the destructor.

This may interact with handles_eof flag.

wiredfool:
I've dumped my thoughts on open file handling into a design doc: https://github.com/wiredfool/Pillow/blob/open-files/docs/reference/open_files.rst

It looks like _exclusive_fd is actually a flag that indicates that we might have a multiple frame image. 

It doesn't appear that there's a good test for if there's > 1 image in a file in general, other than checking n_frames, which at least in the case of the GIF image plugin, has the effect of loading all of the frames in the image (which incidentally changes the image bytes). I haven't checked the others.

condemil:
Fixes #2032

condemil:
Fixes #2361

PR #2047, opened 2016-08-01
TiffImagePlugin: __next is not updated properly

db4:
Fixes `ValueError: seek of closed file` in the attached test case


db4:
Looks like this is not enough. I still have `ValueError: seek of closed file` in a similar case.

wiredfool:
Can you share the other image where you're getting similar behavior?

db4:
I've added another test image that triggers the bug.  I still have no idea how to fix it.

db4:
The test case is here: https://github.com/python-pillow/Pillow/pull/2047/commits/25804cc8c753eb542f66310c0ef6fefff99d97ab

wiredfool:
The general problem illustrated by the second test in this PR is that once you load the last frame in the image, the file is closed, so you can't go back and reload earlier frames. It's a general design decision, and one that probably needs to be looked at in concert with the other file closing issues.

radarhere:
The test in this PR now passes in Pillow 4.1.

hugovk:
Should the tests be included in the test suite?

PR #2052, opened 2016-08-05
Fix image loading when rotating by 0 deg

homm:
This fixes image rotation in cases when `angle % 360.0 == 0` and image is not loaded yet.

### DIFF
--- a/PIL/Image.py
+++ b/PIL/Image.py
@@ -1572,7 +1572,7 @@ class Image(object):
 
         # Fast paths regardless of filter
         if angle == 0:
-            return self._new(self.im)
+            return self.copy()
         if angle == 180:
             return self.transpose(ROTATE_180)
         if angle == 90 and expand:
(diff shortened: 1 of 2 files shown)


homm:
@wiredfool from my point of view, this is critical issue worth 3.3.1, but this is fully up to you.

tisdall:
I did an upgrade and just got stung by this...

homm:
@tisdall You can fix it on your side by explicit `im.load()` before the rotationg.

tisdall:
I actually changed it to test for `angle % 360.0 == 0` before calling rotate.  I just wanted to point out there was at least one person affected by this issue to hopefully add weight to the 3.3.1 suggestion.  Also, I was very lucky to notice this issue mentioned in the change log as the traceback is very cryptic and I likely would have spent a long time trying to sort out the real issue.

PR #2053, opened 2016-08-06
Add BGRa for compatibility with cairo

arjennienhuis:
Cairo uses BGRa format as its `CAIRO_FORMAT_ARGB32`

> each pixel is a 32-bit quantity, with alpha in the upper 8 bits, then red, then green, then blue. The 32-bit quantities are stored native-endian. Pre-multiplied alpha is used. (That is, 50% transparent red is 0x80800000, not 0x80ff0000.)

This patch makes it possible to this:
```python
Image.frombuffer("RGBA", (im.width, im.height), data, "raw", "BGRa", im.stride, 1)
```
instead of this:
```python
im = Image.frombuffer("RGBA", (im.width, im.height), data, "raw", "RGBa", im.stride, 1)
r, g, b, a = im.split()
im = Image.merge("RGBA", (b, g, r, a))
```


wiredfool:
Question: Since the 32 bit quantities are native endian, are there any modes that would be required for similar streamlining on a bigendian machine?

arjennienhuis:
Yes, this needs tests. Can you give me a pointer how I do that? Can I just test the high level interfaces like `Image.frombuffer()` and `tobytes()`?

arjennienhuis:
For bigendian it would be "aRGB" I guess. I have no machine to test this on though. What uses bigendian that I could find?

homm:
I'm against adding new modes just for compatibility with something. This is just extra code to support, no one will do this. I'm closing in favor of #2055.

homm:
By the way, even in the current versions you can do the followings:

```python
Image.frombuffer("RGBa", (im.width, im.height), data, "raw", "BGRa", im.stride, 1).convert('RGBA')
```

PR #2054, opened 2016-08-06
Updated redirect URL

radarhere:


hugovk:
Blogspot is a little odd in that it redirects URLs to local domains.

For me in Finland, both http://littlecms2.blogspot.com/2009/07/less-is-more.html and http://littlecms2.blogspot.com.au/2009/07/less-is-more.html are redirected to https://littlecms2.blogspot.fi/2009/07/less-is-more.html

radarhere:
Ah. Thanks for checking.

PR #2055, opened 2016-08-07
Add unpacking from BRGa to RGBA

homm:
Fixes #2053

Makes possible the next without adding extra core modes:

```python
Image.frombuffer("RGBA", (im.width, im.height), data, "raw", "BGRa", im.stride, 1)
```

PR #2057, opened 2016-08-07
Add (un)packing between RGBA and BGRa

arjennienhuis:
Cairo uses BGRa format as its `CAIRO_FORMAT_ARGB32`

> each pixel is a 32-bit quantity, with alpha in the upper 8 bits, then red, then green, then blue. The 32-bit quantities are stored native-endian. Pre-multiplied alpha is used. (That is, 50% transparent red is 0x80800000, not 0x80ff0000.)

This patch makes it possible to do this:
```python
Image.frombuffer("RGBA", (im.width, im.height), data, "raw", "BGRa", im.stride, 1)
```

This pull request adds to #2055


Tests/test_mode_bgra.py
```diff
@@ -0,0 +1,23 @@
+from helper import unittest, PillowTestCase
```
homm: I suppose, there is no need in the separate module. Please, move pack tests to `test_lib_pack.py: TestLibPack.test_pack` like unpack tests.

libImaging/Pack.c
```diff
@@ -315,6 +319,21 @@ ImagingPackABGR(UINT8* out, const UINT8* in, int pixels)
     }
 }
 
+void
+ImagingPackBGRa(UINT8* out, const UINT8* in, int pixels)
+{
+    int i;
+    /* BGRa, reversed bytes with premultiplied alplha */
+    for (i = 0; i < pixels; i++) {
+	int alpha = out[3] = in[A];
```
homm: Indention issues. Please, use spaces for all new C code.

homm:
👍

Issue #350, opened 2013-09-30
1 bit images to 1 bit nd arrays


wiredfool:
The conversion of 1 bit images to an ndarray is listed as broken in test_numpy.py, Image.py lines 217 and 1931.

pv:
Ping; this is causing crashes for people.

aclark4life:
@pv Can you elaborate?

pv:
cf https://github.com/scipy/scipy/issues/5285, https://github.com/numpy/numpy/issues/5861 for repro

matthew-brett:
Another ping about this one - it's still causing confusing segfaults for numpy users - see e.g. numpy/numpy#7877

PR #2058, opened 2016-08-07
BF: fix conversion of bit images to numpy arrays

matthew-brett:
Numpy cannot form arrays from bits.  To convert bit images to numpy,
convert bits to bytes, using numpy.

Fixes gh-350.

### DIFF
--- a/PIL/Image.py
+++ b/PIL/Image.py
@@ -246,7 +246,7 @@ else:
 
 _MODE_CONV = {
     # official modes
-    "1": ('|b1', None),  # broken
+    "1": ('|b1', None),  # Bits need to be extended to bytes
     "L": ('|u1', None),
     "LA": ('|u1', 2),
     "I": (_ENDIAN + 'i4', None),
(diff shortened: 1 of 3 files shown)


homm:
You do not need numpy to convert bytes to bits, you can do the same with Pillow: `self.tobytes('raw', 'L')`.

matthew-brett:
Ah yes, that's much cleaner - thanks.

PIL/Image.py
```diff
@@ -616,16 +616,21 @@ def _repr_png_(self):
         return b.getvalue()
 
     def __getattr__(self, name):
-        if name == "__array_interface__":
-            # numpy array interface support
-            new = {}
-            shape, typestr = _conv_type_shape(self)
-            new['shape'] = shape
-            new['typestr'] = typestr
+        if name != "__array_interface__":
```
homm: Why we can't just create `__array_interface__` property? /cc @wiredfool

wiredfool: Sounds like a good idea to me.

matthew-brett:
I converted the previous `__getattr__` logic to an `__array_interface__` property, as suggested.   This is a slight change in API, in that you can no longer set the `__array_interface__` attribute on `Image`.

PR #2061, opened 2016-08-08
Added return for J2k (and fpx) Load to return a pixel access object

### LINKED ISSUE #2059: JPEG2000 files do not return pixel access object on calling the load() function
### What did you do?
I tried to load pixels of a j2k file but got None instead of the pixel access object.
Same for other jpeg 2000 file extensions like jpk. 
Since `j2k.show()` shows the image perfectly, the problem is only in the load function. 

I expected it to return the pixel access object like it does for other file types.
### What actually happened?
the load() function returned None
Python 3.4.3 (v3.4.3:9b73f1c3e601, Feb 24 2015, 22:43:06) [MSC v.1600 32 bit (Intel)] on win32
Pillow 3.3.0
```
j2k = Image.open('./MyImage.j2k')
pix = j2k.load()
assert pix is not None #Assertion fails
```

wiredfool:
Fixes #2059 .

Also fixes fpx files, but we have no actual tests for it so,  no new tests here.

### DIFF
--- a/Tests/test_file_jpeg2k.py
+++ b/Tests/test_file_jpeg2k.py
@@ -35,7 +35,7 @@ class TestFileJpeg2k(PillowTestCase):
 
         im = Image.open('Tests/images/test-card-lossless.jp2')
         px = im.load()
-        self.assertTrue(px)
+        self.assertEqual(px[0,0], (0, 0, 0))
         self.assertEqual(im.mode, 'RGB')
         self.assertEqual(im.size, (640, 480))
         self.assertEqual(im.format, 'JPEG2000')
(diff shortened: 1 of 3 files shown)


Tests/test_file_jpeg2k.py
```diff
@@ -34,7 +34,8 @@ def test_sanity(self):
         self.assertRegexpMatches(Image.core.jp2klib_version, '\d+\.\d+\.\d+$')
 
         im = Image.open('Tests/images/test-card-lossless.jp2')
-        im.load()
+        px = im.load()
+        self.assertTrue(px)
```
homm: Maybe a bit more strict check there? For example that the object implements getitem: `px[0,0]`.

homm:
3.3.1?

wiredfool:
Not for this one. This is a long running minor bug that's never actually worked correctly.

PR #2074, opened 2016-08-14
optionally use pkg-config (when present) to detect *_ROOTs

garbas:
This change is [Reviewable](https://reviewable.io/reviews/python-pillow/pillow/2074)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -98,6 +98,11 @@ def _lib_include(root):
     # map root to (root/lib, root/include)
     return os.path.join(root, "lib"), os.path.join(root, "include")
 
+def _cmd_exists(cmd):
+    return any(
+        os.access(os.path.join(path, cmd), os.X_OK) 
+        for path in os.environ["PATH"].split(os.pathsep)
+    )
 
 def _read(file):
     return open(file, 'rb').read()
@@ -119,6 +124,20 @@ TIFF_ROOT = None
 FREETYPE_ROOT = None
 LCMS_ROOT = None
 
+
+def _pkg_config(name):
+    try:
+        command = [
+            'pkg-config',
+            '--libs-only-L', name,
+            '--cflags-only-I', name,
+        ]
+        if not DEBUG:
+            command.append('--silence-errors')
+        libs = subprocess.check_output(command).decode('utf8').split(' ')
+        return libs[1][2:].strip(), libs[0][2:].strip()
+    except:
+        pass
 
 class pil_build_ext(build_ext):
     class feature:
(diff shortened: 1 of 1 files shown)


garbas:
rebased on current master to re-trigger CI builds

garbas:
everything builds ok, except pypi which is anyway broken in current master as well.

garbas:
Is there anything else I can do?

wiredfool:
Looking at the Travis logs, it appears that pkg-config is installed, but it's failing to find anything. I'm not seeing it do anything on my local machine either. 

How is this supposed to be setup? What environments is this intended to improve?

radarhere:

garbas:
just rebased ... lets see

garbas:
@wiredfool ``pkg-config`` no idea how travis setup is configured. but ``pkg-config`` tool is the tool that Pillow should use anyway to look for headers of libraries. i believe that is the most prevalent way of looking for library headers.

In my case i'm using [Nix](https://nixos.org/nix) where header files are not located ``/usr/`` (actually there is no ``/usr``), but in some other locations exposed via system variables at build time so that tools like ``pkg-config`` can look at different places.

this patch just adds that by default it will use ``pkg-config`` (only if present) to look for header files. so you not seeing anything it is a good thing since everything should work as now, except on systems like ``nix`` being able to install Pillow is now possible.

and while writing this comment looks like all tests pass. yaaay!

setup.py
```diff
@@ -119,6 +124,29 @@ def _read(file):
 FREETYPE_ROOT = None
 LCMS_ROOT = None
 
+# ask pkg-config
+if _cmd_exists('pkg-config'):
+    roots = dict(
+        JPEG_ROOT="libjpeg",
+        JPEG2K_ROOT="libopenjp2",
+        TIFF_ROOT="libtiff-4",
```
wiredfool: We support both libtiff-4 (e.g. 3.9.x) and libtiff-5 (4.0.x)

wiredfool:
When running `make clean`, I get the following output:
```
$  make clean
python setup.py clean
Package libtiff-4 was not found in the pkg-config search path.
Perhaps you should add the directory containing `libtiff-4.pc'
to the PKG_CONFIG_PATH environment variable
No package 'libtiff-4' found
Package libtiff-4 was not found in the pkg-config search path.
Perhaps you should add the directory containing `libtiff-4.pc'
to the PKG_CONFIG_PATH environment variable
No package 'libtiff-4' found
Package libimagequant was not found in the pkg-config search path.
Perhaps you should add the directory containing `libimagequant.pc'
to the PKG_CONFIG_PATH environment variable
No package 'libimagequant' found
Package libimagequant was not found in the pkg-config search path.
Perhaps you should add the directory containing `libimagequant.pc'
to the PKG_CONFIG_PATH environment variable
No package 'libimagequant' found
Package libjpeg was not found in the pkg-config search path.
Perhaps you should add the directory containing `libjpeg.pc'
to the PKG_CONFIG_PATH environment variable
No package 'libjpeg' found
Package libjpeg was not found in the pkg-config search path.
Perhaps you should add the directory containing `libjpeg.pc'
to the PKG_CONFIG_PATH environment variable
No package 'libjpeg' found
running clean
removing 'build/temp.linux-x86_64-2.7' (and everything under it)
rm PIL/*.so || true
rm: cannot remove `PIL/*.so': No such file or directory
rm -r build || true
find . -name __pycache__ | xargs rm -r || true
```

So it appears that you're running this any time that the setup runs.

setup.py
garbas: should i make it first search for ``libtiff-5`` and then ``libtiff-4``? or is ``libtiff-4`` preferred as a default?

garbas:
@wiredfool tnx for taking the time for testing this. 

1. i will change that pkg-config is only called when build is triggered (not at every setup.py invocation)
2. i will hide all the debug output of pkg-config behind --debug flag
3. if pkg-config does not find libraries current steps should be taken and reported.

@wiredfool did i understand your comments correctly and is my todo list (above) complete? a :+1: is all i'm looking for before is start working on this ;)

wiredfool:
Looks good. 

For debug reporting, I'd add success or failure for each item so it's clear where we're finding the libraries.

aclark4life:
Weeeee, thanks @garbas ! Is this too risky for 3.4? If not, I say GO GO GO.

wiredfool:
@aclark4life I was still waiting on the changes from my review.

garbas:
@wiredfool @aclark4life rebased on master and fixed above mentioned issues.

garbas:
i also squashed the commits

wiredfool:
Merged in #2212

garbas:
@wiredfool thank you!

PR #2078, opened 2016-08-16
Store TIFF DPI as integer

### LINKED ISSUE #2077: Error when saving TIFF as JPEG with original DPI

```python

        im = Image.open(infile)

        reloaded.load()

```

masterfool:
Fixes #2077 which prevents TIFF from being saved as JPEG with its original DPI.

Stores the TIFF DPI as integers

### DIFF
--- a/PIL/TiffImagePlugin.py
+++ b/PIL/TiffImagePlugin.py
@@ -52,6 +52,7 @@ from numbers import Number, Rational
 
 import io
 import itertools
+import math
 import os
 import struct
 import sys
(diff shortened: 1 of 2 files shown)


arjennienhuis:
I think it makes more sense to round when saving the image. Now you always lose precision, even when not converting the image to JPEG.

masterfool:
Good idea. I've updated the pull request.

wiredfool:
-1 

Tiff resolution is stored as a rational value and a resolution unit.  See page 18/19 here: https://partners.adobe.com/public/developer/en/tiff/TIFF6.pdf

If there's an issue with jpegs, then either:

1) The jpeg spec doesn't allow rational dpi, and we're enforcing that in a confusing way 
or 
2) We're not properly allowing a rational dpi in the jpeg save routine.

arjennienhuis:
https://www.w3.org/Graphics/JPEG/jfif3.pdf

page 5

integer dpi or dots per cm

wiredfool:
That certainly looks like #1 is the issue, and a better way to enforce integer dpi in jpegs would do it.

arjennienhuis:
see #2084

wiredfool:
Closing in (eventual) favor of #2084

PR #2084, opened 2016-08-17
Round float dpi when saving JPEG

### LINKED ISSUE #2077: Error when saving TIFF as JPEG with original DPI

arjennienhuis:
Fixes #2077


wiredfool:
I think that this is better done in the python realm at: https://github.com/python-pillow/Pillow/blob/master/PIL/JpegImagePlugin.py#L695

radarhere:

PR #2087, opened 2016-08-22
Fixes TIFFImagePlugin ICC color profile saving.

cskau:
In the TIFF code saving icc_profile is conditional on tag_v2 being set which doesn't make sense to me. I believe this is merely an indentation typo.

I've been trying to save TIFFs with im.info['icc_profile'] set and compression=raw, but unfortunately this results in TIFFs without ICC color profiles.
With the attached patch TIFFs with said conditions will be saved with the profile set in im.info['icc_profile'].

Note: There are a number of different conditions that need to be met for code to succeed in saving with the profile since it branches between using libtiff and ImageFile._save(..), and the libtiff code does not currently save the ICC color profile.
For instance setting compression=tiff_lzw will result in using libtiff and no profile will be saved.

### DIFF
--- a/PIL/TiffImagePlugin.py
+++ b/PIL/TiffImagePlugin.py
@@ -1366,10 +1366,10 @@ def _save(im, fp, filename):
                 ifd[key] = im.tag_v2[key]
                 ifd.tagtype[key] = im.tag_v2.tagtype[key]
 
-        # preserve ICC profile (should also work when saving other formats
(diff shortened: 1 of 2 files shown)


radarhere:

cskau:
@radarhere, thank you for the tip.
It took me a couple of tries but it looks like we got there.

wiredfool:
Can you add a test case that fails on master but passes on this PR?

cskau:
Test added.

PR #2088, opened 2016-08-22
Updated lcms2 to 2.8

radarhere:


radarhere:
So this has failed because the SourceForge URL isn't working. However, at least some, if not all, of the SourceForge URLs in config.py aren't working either for me. Any thoughts from anyone?

wiredfool:
We should pull from the depends repo instead, and have docs for refreshing and verifying.

hugovk:
Searching the web for "iweb.dl.sourceforge.net" gives other files which also now 404.

Some if not all of these SourceForge files are already in https://github.com/python-pillow/pillow-depends. Should we use those instead?

wiredfool:
iweb is a sourceforge mirror that worked when these were first added, but may need updating now.  I believe that sourceforge has undergone some major changes recently(possibly for the better), so it's not surprising that the mirroring has shifted. 

It would probably make sense to check and consolidate so that we have at least one known good source in our depends repo.

hugovk:
See PR https://github.com/python-pillow/Pillow/pull/2095.

aclark4life:
Do we have a label for "Some checks were not successful"? :smile: (appveyor)

radarhere:
I've rebased this after the merge of https://github.com/python-pillow/Pillow/pull/2095

It requires lcms2 2.8 to be present. So https://github.com/python-pillow/pillow-depends/pull/15 will need to be merged before this can pass. I have left the 2.7 zip present there as well, so that unrelated Pillow PRs are not affected.

radarhere:
I merged the pillow-depends PR, but this PR is still failing.

PR #2083, opened 2016-08-17
Ignore PyPy numpy errors

arjennienhuis:
Fixes #2076

Just ignore the numpy tests on PyPy

PR #2090, opened 2016-08-25
Pr 2083

wiredfool:
More specific version of PR #2083.  

* We will be reminded about this later
* Only skips failing tests

### DIFF
--- a/Tests/test_numpy.py
+++ b/Tests/test_numpy.py
@@ -1,4 +1,5 @@
 from __future__ import print_function
+import sys
 from helper import unittest, PillowTestCase, hopper
 
 from PIL import Image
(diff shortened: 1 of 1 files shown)


Tests/test_numpy.py
```diff
@@ -14,6 +15,13 @@
 
 TEST_IMAGE_SIZE = (10, 10)
 
+# Numpy on pypy as of pypy 5.3.1 is corrupting the numpy.array(Image)
+# call such that it's returning a object of type numpy.ndarray, but
```
hugovk: "a object" -> "an object"

Tests/test_numpy.py
```diff
@@ -14,6 +15,13 @@
...
+# the repr is that of a PIL.Image. Size and shape are 1 and (), not the
+# size and shape of the array. This causes failures in several tests.
+SKIP_NUMPY_ON_PYPY = hasattr(sys, 'pypy_version_info') and (
+    sys.pypy_version_info <= (5,3,1,'final',0))
```
hugovk: Spaces after commas

PR #2095, opened 2016-08-29
Windows build: fetch dependencies from pillow-depends

hugovk:
We cache build dependencies in https://github.com/python-pillow/pillow-depends, which is already cloned during Windows builds, so for Windows builds, grab them from there instead of downloading them fresh from the internet, especially as mirrors can break (e.g. see https://github.com/python-pillow/Pillow/pull/2088).

The second commit has the main changes: https://github.com/python-pillow/Pillow/commit/5a0104f64d5baf9277271722ad6b68092f1e36a2.

Before, things in the `libs` list were downloaded and extracted. Things in the `bin_libs` were just extracted, but `extract_binlib` was commented out. Instead, move them all to the `bin_libs` list (and then remove unused things like `fetch_libs` and rename things like `extract_binlib` to `extract_libs`).

As a before-and-after check, AppVeyor passes and each equivalent Python-version build job has the same number of run and skipped tests.


wiredfool:
2 Comments:

1) This makes the winbuild scripts rely on the appveyor context to run, or at least a non-scripted step of pulling in the depends repo. 
2) The list of urls was minimal documentation for where we were getting the files at a specific point in time.

hugovk:
1) Should the git clone stuff (https://github.com/python-pillow/Pillow/blob/master/appveyor.yml#L19-L21) be moved inside build_dep.py (to write into build_deps.cmd)? Downloading from URLs won't work at the moment, given the broken SF mirrors.

2) I've reinstated the documenting URLs.

PR #2100, opened 2016-09-03
Flake8 and whitespace fixes

radarhere:


hugovk:
Generally speaking, I wouldn't bother editing files just to remove whitespace.

(My editor's configured to do so for those I edit, and we have https://github.com/python-pillow/Pillow/blob/master/.editorconfig#L15 for editors that use it.)

radarhere:
Fair enough - my editor works similarly these days, although I'd thought that I was the exception, and had felt odd about submitting commits with these apparently irrelevant whitespace changes. So thanks for the thoughts!

PR #2102, opened 2016-09-03
Convert DPI to ints when saving as JPEG

### LINKED ISSUE #2077: Error when saving TIFF as JPEG with original DPI

radarhere:
Fixes #2077. Alternative to PR #2084.

Uses the test suggested in the issue.

### DIFF
--- a/PIL/JpegImagePlugin.py
+++ b/PIL/JpegImagePlugin.py
@@ -593,7 +593,7 @@ def _save(im, fp, filename):
 
     info = im.encoderinfo
 
-    dpi = info.get("dpi", (0, 0))
+    dpi = [int(round(x)) for x in info.get("dpi", (0, 0))]
 
     quality = info.get("quality", 0)
     subsampling = info.get("subsampling", -1)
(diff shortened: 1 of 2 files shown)


PIL/JpegImagePlugin.py
```diff
@@ -593,7 +593,7 @@ def _save(im, fp, filename):
 
     info = im.encoderinfo
 
-    dpi = info.get("dpi", (0, 0))
+    dpi = [int(x) for x in info.get("dpi", (0, 0))]
```
arjennienhuis: I would round to nearest integer.

Ideally it would round to the nearest integer dpi or the nearest integer "dot per centimeter"

radarhere: Is this essentially what you're referring to? https://github.com/python-pillow/Pillow/blob/master/PIL/BmpImagePlugin.py#L104

wiredfool: @radarhere I wouldn't worry about meters/cm, but rounding seems reasonable.

radarhere: So the rounding to the nearest integer that I'm already doing? Apologies if confused.

wiredfool:
```
>>> int(3.9)
3
>>> int(round(3.9))
4
```


<aside>  what a lovely bikeshed we've got here </aside>

radarhere: @wiredfool thanks. updated the PR.

Issue #1713, opened 2016-02-04
Initial setup of python-pillow.org domain


aclark4life:
Discuss initial setup of python-pillow.org domain.

aclark4life:
For my purposes, I'd like to see:

- security@python-pillow.org going to @wiredfool and myself
- Something like team@python-pillow.org possibly going to everyone on the @python-pillow/pillow-team (if folks think that would be useful).
- Website moved to python-pillow.org

Normally I'd suggest Google Apps but I don't think it's free anymore. Anyone else have suggestions about how to support the above?

hugovk:
You can set up a domain for the existing gh-pages branch with a CNAME file:

https://help.github.com/articles/setting-up-a-custom-domain-with-github-pages/

wiredfool:
I can easily throw this up on my hosting account, essentially for free.

hugovk:
It's alive!

http://python-pillow.org

https://github.com/python-pillow/python-pillow.github.io/commit/cb1ade721d0ea21e40a414a6a0ec4641f9c38234

aclark4life:
Yep, working on it; will have updates soon.

radarhere:
Out of interest, what is left to be done in this issue?

aclark4life:
@radarhere I've tested and ruled out Bitrix24 for hosting, now testing Zoho. I've got aclark@python-pillow.org forwarding working, but stuck on security@python-pillow.org. Will try to finish soon. @wiredfool 's offer to host is appreciated, but I want a generic "white box" service that can be easily managed by anyone.

hugovk:
http://python-pillow.org/ is hosted on GitHub Pages right now: https://github.com/python-pillow/python-pillow.github.io 

Is that sufficient for static pages? Or did you want to host something else?

Advantages with GitHub Pages: it's free, it's open, it deploys directly from the GitHub repo, it's managed by all of us, others can make PRs and so on.

aclark4life:
@hugovk I think so, yes

aclark4life:
And … finally realized I can add python-pillow.org to my (grandfathered) Google Apps account. Thanks to: https://news.ycombinator.com/item?id=8534231. ``security@python-pillow.org`` now configured to send mail to Eric & myself. :cake:

PR #2106, opened 2016-09-06
Update security email address

hugovk:
Re: https://github.com/python-pillow/Pillow/issues/1713#issuecomment-204597429


aclark4life:
Thx

PR #2110, opened 2016-09-14
Allow pathlib.Path in Image.open on Python 2.7

patricksnape:

 * Allow using `pathlib.Path` for `Image.open` on Python 2.7

Although the pathlib backport for Python 2.7 may be deprecated:
https://pypi.python.org/pypi/pathlib/
It is still used by many projects. Therefore, changing to a
Try/Except pattern for checking for pathlib is not any more
obtrusive that the current >= Python 3.4 check and allows users
to use the backport without issue.

### DIFF
--- a/PIL/Image.py
+++ b/PIL/Image.py
@@ -2279,10 +2279,14 @@ def open(fp, mode="r"):
     filename = ""
     if isPath(fp):
         filename = fp
-    elif sys.version_info >= (3, 4):
-        from pathlib import Path
-        if isinstance(fp, Path):
-            filename = str(fp.resolve())
+    else:
+        try:
+            from pathlib import Path
+            if isinstance(fp, Path):
+                filename = str(fp.resolve())
+        except ImportError:
+            pass
+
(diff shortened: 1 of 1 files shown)


PIL/Image.py
```diff
@@ -2279,10 +2279,14 @@ def open(fp, mode="r"):
     filename = ""
     if isPath(fp):
         filename = fp
-    elif sys.version_info >= (3, 4):
```
hugovk: Before, the pathlib stuff was only run when `isPath(fp)` is False (and Python >= 3.4).

Now, it's always run, whether `isPath(fp)` is False or True.

Should the `try`/`except` be put in an `else` case, so it's only run when `isPath(fp)` is False?

patricksnape: To maintain consistency with the previous behaviour you are totally right.

patricksnape:
@hugovk Thanks for the review - that update has been made (force pushed)

hugovk:
Thanks!

PR #2113, opened 2016-09-17
[SgiImagePlugin] Save uncompressed SGI/BW/RGB/RGBA files

jbltx:
**Issue:**
SGI files are only readable

-  Add a _save method for uncompressed version of SGI File (not RLE)
-  Able to save Image as .sgi, .bw, .rgb, .rgba
- Support "L", "RGB" and "RGBA" modes (like read method)

**Note:**
A future PR could be possible to add RLE feature

### DIFF
--- a/PIL/SgiImagePlugin.py
+++ b/PIL/SgiImagePlugin.py
@@ -86,25 +86,24 @@ def _save(im, fp, filename):
     if im.mode != "RGB" and im.mode != "RGBA" and im.mode != "L":
         raise ValueError("Unsupported SGI image mode")
 
-    im = im.transpose(Image.FLIP_TOP_BOTTOM)                                    # Flip the image, since the origin of SGI
-                                                                                # file is the bottom-left corner
+    im = im.transpose(Image.FLIP_TOP_BOTTOM)
 
-    magicNumber = 474                                                           # Define the file as SGI File Format
-    rle = 0                                                                     # Run-Length Encoding Compression - Unsupported
-    bpc = 1                                                                     # Byte-per-pixel precision, 1 = 256bits per pixel
-    dim = 3                                                                     # Number of dimensions (x,y,z)
-    x, y = im.size                                                              # X Dimension / Y Dimension
+    magicNumber = 474
+    rle = 0
+    bpc = 1
+    dim = 3
+    x, y = im.size
     if im.mode == "L" and y == 1:
         dim = 1
     elif im.mode == "L":
         dim = 2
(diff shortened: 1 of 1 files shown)


radarhere:
Hi. Thanks for putting this contribution forward. It would be helpful if you would also be able to add test cases to the test suite. This would decrease the chance of this functionality being broken by another change in the future.

Also, would you've amended the description with 'hack version'. What's your reasoning behind that?

jbltx:
Woops, I let my old edit in the comment at the top, if you can edit it and remove this "hack" it could be nice. (It was used as keyword to find my edited file in several libs)

radarhere:
Due to the nature of GitHub, I can't edit the code in this PR. This is a page to discuss whether or not to include commits from your repository. I have certain permissions on the python-pillow Pillow repository, but none on your version of it.

So just for information, the various options would be -
- You could add a commit your master branch. New commits will automatically appear here.
- I could submit a PR to your repository, so you could choose to accept or reject commits that I made. Once accepted, those commits from me would similarly appear here.
- Be aware that both of those options would still leave the phrase 'hack version' in the commit history. That's not a problem as far as Pillow is concerned, but if you wanted to remove all evidence of that, you could reset these commits `git reset --soft HEAD^` and commit them again, either still on master in this PR, or with a new branch in a new PR.

hugovk:
@jbltx The inline comments you removed in https://github.com/python-pillow/Pillow/pull/2113/commits/3ab47512a0a5f3e18f71e9323d369b9369fe4bb4 would still be useful for future maintainers. How about putting them on the preceding lines?

jbltx:
No problem i can comment my code again and commit it this morning.

jbltx:
I have finished RLE compression and 8/16bits per channel support for save method, but  i encountered somethig I dont get in the documentation : 

> You can use the raw decoder to read images where data is packed in any standard machine data type, using one of the following raw modes[...]

In the table given there's `F;16B` to decode 16bits data (unsigned short), but this raw mode doesn't exist when I set my tile descriptor:

```py
self.tile.append(("raw", (0, 0)+self.size, offset, ('F;16B', 0, -1)))
```

Did I miss something ?

Thanks

wiredfool:
@jbltx Sorry, I missed your last post when it happened. I can take a look at that but I'll need a little more context of what's going on there. 

I've added tests, squashed, rebased, and gotten this ready to go into the next release as it is, since it's useful in it's current (fixed anyway) state.

PR #2114, opened 2016-09-17
ImageColor.getrgb hexadecimal RGBA

homm:
Support of [hexadecimal RGBA](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#rgba) notations.

### DIFF
--- a/Tests/test_imagecolor.py
+++ b/Tests/test_imagecolor.py
@@ -7,51 +7,68 @@ from PIL import ImageColor
 class TestImageColor(PillowTestCase):
 
     def test_hash(self):
+        # short 3 components
         self.assertEqual((255, 0, 0), ImageColor.getrgb("#f00"))
         self.assertEqual((0, 255, 0), ImageColor.getrgb("#0f0"))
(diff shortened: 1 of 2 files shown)


Tests/test_imagecolor.py
```diff
@@ -6,32 +6,96 @@
 
 class TestImageColor(PillowTestCase):
 
-    def test_sanity(self):
+    def test_hash(self):
         self.assertEqual((255, 0, 0), ImageColor.getrgb("#f00"))
-        self.assertEqual((255, 0, 0), ImageColor.getrgb("#ff0000"))
+        self.assertEqual((0, 255, 0), ImageColor.getrgb("#0f0"))
+        self.assertEqual((0, 0, 255), ImageColor.getrgb("#00f"))
```
hugovk: Maybe add some uppercase here and later? e.g.
`self.assertEqual((0, 0, 255), ImageColor.getrgb("#00F"))`

homm: Done below

PR #2115, opened 2016-09-17
Read value when checking for JPEG progressive key

### LINKED ISSUE #2064: unintuitive save properties
> **progressive**
> If present, indicates that this image should be stored as a progressive JPEG file.

It is not completely obvious that ```progressive = False``` evaluates to it being set. I guess it's difficult to change this whilst keeping backwards compatibility. (I doubt someone passes ```progressive = False``` with the intention to enable it.)

radarhere:
Resolves #2064

### DIFF
--- a/PIL/JpegImagePlugin.py
+++ b/PIL/JpegImagePlugin.py
@@ -682,13 +682,16 @@ def _save(im, fp, filename):
                       o8(len(markers)) + marker)
             i += 1
 
+    # "progressive" is the official name, but older documentation
+    # says "progression"
+    # FIXME: issue a warning if the wrong form is used (post-1.1.7)
+    progressive = info.get("progressive", False) or\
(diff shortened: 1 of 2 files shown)


wiredfool:
Maybe add a test to test_file_jpeg.py:test_progressive_compat ?

radarhere:
Fair call, thanks. I've added tests.

wiredfool:
OK, LGTM.  However, it looks like optimize has the same issue. Could probably use something in the docs as well, but I'm going to merge this and we can deal with the rest of it in the next pr.

pdknsk:
I've [commented](https://github.com/python-pillow/Pillow/issues/2064#issuecomment-249258740) on the issue, but it appears there are no subscribers, so I'm mentioning it here.

PR #2120, opened 2016-09-18
Replace deprecated functions with new ones

NotSoSuper:
```
bot3-7 (err): /usr/local/lib/python3.5/dist-packages/PIL/ImageFont.py:410: DeprecationWarning: decodestring() is a deprecated alias, use decodebytes()
bot3-7 (err):   ''')), Image.open(BytesIO(base64.decodestring(b'''
bot3-7 (err): /usr/local/lib/python3.5/dist-packages/PIL/ImageFont.py:434: DeprecationWarning: decodestring() is a deprecated alias, use decodebytes()
bot3-7 (err):   '''))))
```

Changes proposed in this pull request:

 * replace deprecated functions that actually call and return the new function in the code

PR #2121, opened 2016-09-19
Use modern base64 interface over deprecated

hugovk:
Replaces and closes https://github.com/python-pillow/Pillow/pull/2120.

 * Deprecated `base64.decodestring` has started issuing deprecation warnings; see #2120.
 * Replace deprecated `base64.decodestring` from the legacy interface with `base64.b64decode` from the modern interface. https://docs.python.org/3.5/library/base64.html
 * `base64.standard_b64decode` from the modern interface also passes: https://travis-ci.org/hugovk/Pillow/builds/160945046


NotSoSuper:
Thanks for finding a better way.

hugovk:
@NotSoSuper Thanks for reporting this in the first place!

PR #2122, opened 2016-09-19
Allow ImageFont getsize to read bytes

### LINKED ISSUE #2117: TypeError: expected string  [draw.textsize(font)]
### That piece of code:
```python
W, H = (1285,720)
text = "Hello world"
img = Image.open("img.png")
draw = ImageDraw.Draw(img)
font = ImageFont.truetype("/usr/share/fonts/MyriadProCondBold.ttf", 90)
w, h = draw.textsize(text.encode('utf-8'), font=font)
draw.text(((W-w)/2,520),text,(255,255,255),font=font)
```
### Output:
```python
w, h = draw.textsize(text.encode('utf-8'), font=font)
  File "/usr/lib/python3.5/site-packages/PIL/ImageDraw.py", line 273, in textsize
    return font.getsize(text)
  File "/usr/lib/python3.5/site-packages/PIL/ImageFont.py", line 141, in getsize
    size, offset = self.font.getsize(text)
TypeError: expected string
```

radarhere:
Resolves #2117

### DIFF
--- a/PIL/ImageFont.py
+++ b/PIL/ImageFont.py
@@ -138,6 +138,8 @@ class FreeTypeFont(object):
         return self.font.ascent, self.font.descent
 
     def getsize(self, text):
+        if bytes is not str and isinstance(text, bytes):
+            text = text.decode()
         size, offset = self.font.getsize(text)
         return (size[0] + offset[0], size[1] + offset[1])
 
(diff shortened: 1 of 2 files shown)


homm:
I'm not sure that we should fix it. We can work with fonts in terms of text, not bytes. As I understand, `font.render` also doesn't works with bytes.

radarhere:
Yes, I suspected that it wasn't suitable. Feel free to close.

wiredfool:
I don't think that this is appropriate.

PR #2129, opened 2016-09-25
Use truth value when checking for optimize option on save

radarhere:
See #2064

### DIFF
--- a/docs/handbook/image-file-formats.rst
+++ b/docs/handbook/image-file-formats.rst
@@ -219,20 +219,20 @@ The :py:meth:`~PIL.Image.Image.save` method supports the following options:
     image quality.
 
 **optimize**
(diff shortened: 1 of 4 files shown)


docs/handbook/image-file-formats.rst
```diff
@@ -219,19 +219,19 @@ The :py:meth:`~PIL.Image.Image.save` method supports the following options:
     image quality.
 
 **optimize**
-    If present, indicates that the encoder should make an extra pass over the
-    image in order to select optimal encoder settings.
+    If present and non-empty, indicates that the encoder should make an extra
```
wiredfool: I'd say present and true.

radarhere: Okay, I've updated the PR.

PR #2138, opened 2016-09-29
Make Image.crop an immediate operation, not lazy.

### LINKED ISSUE #1077: pillow Image.crop function cause python.exe crash
So here is what i encountered:
assume i have an image file named " bmg1.bmp", if i run the script below

```python
from PIL import Image
aoi = (1, 1, 10, 10) # an example of aoi
img = Image.open('bmg1.bmp')
img2 = img.crop(aoi)
img2.show()
```
everything is fine , but if I run:
```python
from PIL import Image
aoi = (1, 1, 10, 10)
img = Image.open('bmg1.bmp')
img = img.crop(aoi)
img.show()
```

same error happens if i do:
```python
from PIL import Image
aoi = (10, 10, 15, 15)
img = Image.open('bmg1.bmp').crop(aoi)
img.show()
```
the python.exe will crash, i am not sure if this is CPython's bug or pillow's bug. 

i doubt the crop function seems to return a Image object that is different from the original one, thus run show() on it gives errors, while does not on the original one.
both python 27 and python 34 have this problem, hope that will be resolved soon.

thanks!

wiredfool:
Fixes #1077 .

 * `Image.crop` is now immediate, and returns a new image object
 * All the important fields are copied over, using the _new function, including the palette
 * `_ImageCrop` is now gone.
 * Fixes an access violation that looks like a use after free on windows.

((probably should have release notes))

### DIFF
--- a/PIL/Image.py
+++ b/PIL/Image.py
@@ -1020,10 +1020,7 @@ class Image(object):
         4-tuple defining the left, upper, right, and lower pixel
         coordinate.
 
-        This is a lazy operation.  Changes to the source image may or
-        may not be reflected in the cropped image.  To break the
-        connection, call the :py:meth:`~PIL.Image.Image.load` method on
-        the cropped copy.
+        Note: Prior to Pillow 3.4.0, this was a lazy operation. 
 
         :param box: The crop rectangle, as a (left, upper, right, lower)-tuple.
         :rtype: :py:class:`~PIL.Image.Image`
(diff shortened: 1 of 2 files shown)


aclark4life:
Thanks @homm

aclark4life:
Still a coveralls check "not successful"; not sure what's going on there. Otherwise, @homm's review is good enough for me.

wiredfool:
Pretty sure that the coveralls is because I deleted code, so fewer lines covered.

PR #2139, opened 2016-09-29
Force reloading palette when using mmap in ImageFile.

lambdafu:
When ImageFile uses mmap, it replaces self.im.  However, this resets self.im.palette, which may have already been set from self.palette raw data in a multipage TIFF loader. To force Pillow to reload self.palette into self.im, set the dirty flag.

I added a testcase.  multipage-mmap.tiff is needed, because multipage.tiff is encoded in a way that does not allow mmap to be used.


wiredfool:
My two favorite bug nexuses, tiff and palettes. Yay.

lambdafu:
And after debugging this, I know why :) It's a bit messy, but the tests will help eventual refactoring efforts for sure.

PR #2149, opened 2016-10-04
CRITICAL: Allow lists as arguments for Image.new()

homm:
Fixes the broken backward compatibility introduced in #2146

I believe that this breaks is A LOT of code and the fix should be released as 3.4.1.

@aclark4life @hugovk @wiredfool

### DIFF
--- a/PIL/Image.py
+++ b/PIL/Image.py
@@ -1020,7 +1020,7 @@ class Image(object):
         4-tuple defining the left, upper, right, and lower pixel
         coordinate.
 
-        Note: Prior to Pillow 3.4.0, this was a lazy operation. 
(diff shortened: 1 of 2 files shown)


wiredfool:
Aw crud.

PIL/Image.py
```diff
@@ -1993,7 +1993,7 @@ def _check_size(size):
     :returns: True, or raises a ValueError
     """
 
-    if not isinstance(size, tuple):
+    if not isinstance(size, (list, tuple)):
         raise ValueError("Size must be a tuple")
```
hugovk: "Size must be a tuple" 
-> 
"Size must be a tuple or list"

homm: Not necessary. We can require tuple and accept both.

PR #2155, opened 2016-10-05
Add center and translate option to Image.rotate.

lambdafu:
Hi,

freeimage has a RotateEx function that allows to set a rotation center and a subsequent translation. Due to the affine matrix used for the interpolation, this is super-easy to implement in Pillow, too, and really makes a lot of sense to provide in the API, as calculating the transformation matrix is not something that every user will be able to do easily.

- Add center=(x,y) argument to Image.rotate
- Add translate=(x,y) argument to Image.rotate
- Documentation and tests for these.

### DIFF
--- a/PIL/Image.py
+++ b/PIL/Image.py
@@ -1554,7 +1554,7 @@ class Image(object):
 
         return self._new(self.im.resize(size, resample))
 
-    def rotate(self, angle, resample=NEAREST, expand=0):
+    def rotate(self, angle, resample=NEAREST, expand=0, center=None, translate=None):
         """
         Returns a rotated copy of this image.  This method returns a
         copy of this image, rotated the given number of degrees counter
@@ -1568,10 +1568,14 @@ class Image(object):
            (cubic spline interpolation in a 4x4 environment).
            If omitted, or if the image has mode "1" or "P", it is
            set :py:attr:`PIL.Image.NEAREST`. See :ref:`concept-filters`.
+        :param center: Optional center of rotation (a 2-tuple).  Origin is
+           the upper left corner.  Default is the center of the image.
+        :param translate: Am optional final translation.
         :param expand: Optional expansion flag.  If true, expands the output
            image to make it large enough to hold the entire rotated image.
            If false or omitted, make the output image the same size as the
-           input image.
+           input image.  Note that the expand flag assumes rotation around
+           the center and no translation.
         :returns: An :py:class:`~PIL.Image.Image` object.
         """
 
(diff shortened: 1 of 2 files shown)


PIL/Image.py
```diff
@@ -1554,7 +1554,7 @@ def resize(self, size, resample=NEAREST):
 
         return self._new(self.im.resize(size, resample))
 
-    def rotate(self, angle, resample=NEAREST, expand=0):
+    def rotate(self, angle, center=None, translate=None, resample=NEAREST, expand=0):
```
hugovk: Should `center` and `translate` be added as last parameters, in case existing code already calls `rotate` assuming the third and fourth parameters are `resample` and `expand`?

lambdafu: Yes, that's a good idea.  I pushed a squashed change to that effect. Thanks!

PIL/Image.py
```diff
@@ -1568,10 +1568,14 @@ def rotate(self, angle, resample=NEAREST, expand=0):
            (cubic spline interpolation in a 4x4 environment).
            If omitted, or if the image has mode "1" or "P", it is
            set :py:attr:`PIL.Image.NEAREST`. See :ref:`concept-filters`.
+        :param center: Optional center of rotation (a 2-tuple).  Origin is
+           the upper left corner.  Default is the center of the image.
+        :param translate: Am optional final translation.
```
radarhere: I'm guessing 'An'.

lambdafu:
Is there anything I can do to move this along?

PIL/Image.py
```diff
@@ -1568,10 +1568,14 @@ def rotate(self, angle, resample=NEAREST, expand=0):
            (cubic spline interpolation in a 4x4 environment).
            If omitted, or if the image has mode "1" or "P", it is
            set :py:attr:`PIL.Image.NEAREST`. See :ref:`concept-filters`.
+        :param center: Optional center of rotation (a 2-tuple).  Origin is
+           the upper left corner.  Default is the center of the image.
+        :param translate: An optional final translation.
```
wiredfool: Doc param order isn't matching the function parameter order.

PIL/Image.py
```diff
@@ -1587,32 +1591,59 @@ def rotate(self, angle, resample=NEAREST, expand=0):
         if angle == 270 and expand:
```
wiredfool: The fastpath cases aren't properly handled except for the angle == 0 case, which is correctly still a no-op. If the caller has passed a center or translate for the other three cases, the result is going to be different than the fast path.

wiredfool:
Sorry that this has taken so long to be reviewed.  I think this is a reasonable addition to the api, and I'd be happy to merge when I'm sure that we're getting the mechanics right.

lambdafu:
I agree it's undertested.

PR #2162, opened 2016-10-10
Fix coefficients calculation

### LINKED ISSUE #2161: Pillow 3.4 breaks Image.thumbnail
### What did you do?
```python
    img = Image.open("test.png")
    img.thumbnail((256, 256))
    img.save("output.png")
```
Create a thumbnail.

### What actually happened?
It creates a thumbnail with random transparent lines all over it.

Python 3.5 and Pillow 3.4.1. With Pillow 3.3.3 it works perfectly.

test

output

homm:
Fixes #2161.

### DIFF
--- a/libImaging/Resample.c
+++ b/libImaging/Resample.c
@@ -176,6 +176,9 @@ precompute_coeffs(int inSize, int outSize, struct filter *filterp,
         k = &kk[xx * kmax];
         for (x = 0; x < xmax; x++) {
             double w = filterp->filter((x + xmin - center + 0.5) * ss);
+            k[x] = w;
+            ww += w;
+            
             // We can skip extreme coefficients if they are zeroes.
             if (w == 0) {
                 // Skip from the start.
@@ -185,15 +188,11 @@ precompute_coeffs(int inSize, int outSize, struct filter *filterp,
                     // But `w` will not be 0, because it based on `xmin`.
                     xmin += 1;
                     xmax -= 1;
-                    // Prevent setting the `k[-1]` item.
-                    continue;
                 } else if (x == xmax - 1) {
                     // Truncate the last coefficient for current `xx`.
                     xmax -= 1;
                 }
             }
-            k[x] = w;
-            ww += w;
         }
         for (x = 0; x < xmax; x++) {
             if (ww != 0.0)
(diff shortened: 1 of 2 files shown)


homm:
Please note, this is PR for `3.4` branch. We also need it in master.

hugovk:
Looks good:
output

Is there a test you could add to guard against future regressions?

homm:
@hugovk test added

libImaging/Resample.c
```diff
@@ -184,7 +184,6 @@ precompute_coeffs(int inSize, int outSize, struct filter *filterp,
                 } else if (x == xmax - 1) {
                     xmax -= 1;
                 }
-                continue;
             }
             k[x] = w;
```
wiredfool: Is this bug a case of using uninitialized memory?  It kind of looks that way from the non-deterministic pattern of the alpha in the bug.

Would initializing all the coefficients to 0 have prevented this?

homm: > Is this bug a case of using uninitialized memory?

Yes

> Would initializing all the coefficients to 0 have prevented this?

Yes. Actually, I was certain that all memory is initialized:

1. `kk` has size of `outSize` * `kmax` doubles;
2. `xbounds` has size of two `outSize` ints;
3. We iterate `xx` from 0 to `outSize` in outer loop, so we initialize whole `xbounds`;
4. We iterate `x` in inner loop from 0 to `xmax` and initialize it with some coefficient `w`;
5. Then iterate `x` from current `x`(`xmax`) to `kmax`, and initialize it with zeroes.

So, `kk` and `xbounds` should be guaranteed initialized. The bug was on the fourth step: there was a `continue`, so on some conditions, some `kk` elements were not initialized.

homm: Hmm, but such uninitialized values should be outside of `xmax`. I need 10 minutes to investigate why this leads to the bug.

wiredfool: Yeah, I'm not seeing how that happens. xmax should be decremented on any 0 coefficient, and then the last loop should go from that decremented xmax to kmax. 

It does look like a calloc instead of malloc would solve all the initialization problems in one less math prone statement:
```
    kk = calloc(outSize * kmax , sizeof(double));
```

homm: > It does look like a calloc instead of malloc would solve all the initialization problems

This is true for this case. Of course, it would be better to initialize memory one extra time than get a bug. But I think it is unnecessary for now when we exactly know what everything is initialized. I also eliminated the last `continue` from this code.

homm:
@wiredfool Are you happy with the current fix. I'm afraid this is a perfect candidate for v3.4.2. Sorry for this.

PR #2171, opened 2016-10-19
Refer user to installation docs on jpeg/zlib ValueError

hugovk:
We get a [lot of issues](https://github.com/python-pillow/Pillow/search?q=is+required+unless+explicitly+disabled+using&type=Issues&utf8=%E2%9C%93) about `ValueError: jpeg is required unless explicitly disabled using --disable-jpeg, aborting` or `ValueError: zlib is required unless explicitly disabled using --disable-zlib, aborting` which are usually solved by pointing them in the direction of the docs.

So let's include it in the error message.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -518,7 +518,9 @@ class pil_build_ext(build_ext):
                 if f in ('jpeg', 'zlib'):
                     raise ValueError(
                         '%s is required unless explicitly disabled'
-                        ' using --disable-%s, aborting' % (f, f))
+                        ' using --disable-%s, aborting. See https://pillow.'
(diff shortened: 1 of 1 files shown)


wiredfool:
I'd like to see setup.py cleaned up to be more pythonic. (Not necessarily big change all the things changes, but as we do things we should clean it up.).

I think something like this makes the code clearer and cleaner: https://github.com/wiredfool/Pillow/commit/96453efb73931d90a4278533cb06328ac6ceb24a

hugovk:
wiredfool@96453ef looks good (maybe rename `DefaultDependencyException` to `RequiredDependencyException`), want to PR it?

PR #2172, opened 2016-10-22
Fix BytesWarning discovered while running tests

jdufresne:

Fixes warnings that look like: "BytesWarning: Comparison between bytes and string"

Discovered using the command:

python -b -m nose -vx Tests/test_*.py

### DIFF
--- a/PIL/ImtImagePlugin.py
+++ b/PIL/ImtImagePlugin.py
@@ -69,7 +69,7 @@ class ImtImageFile(ImageFile.ImageFile):
                 s = s + self.fp.readline()
                 if len(s) == 1 or len(s) > 100:
                     break
-                if s[0] == b"*":
+                if s[0] == ord(b"*"):
                     continue  # comment
 
                 m = field.match(s)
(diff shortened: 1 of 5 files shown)


PIL/ImtImagePlugin.py
```diff
@@ -69,7 +69,7 @@ def _open(self):
                 s = s + self.fp.readline()
                 if len(s) == 1 or len(s) > 100:
                     break
-                if s[0] == b"*":
+                if s[0] == ord(b"*"):
```
wiredfool: Not a ding against the PR, but the ImtImagePlugin is essentially untested, or at the most generous interpretation, it's tested in that it doesn't open any of the non-imt images that we try to open.

hugovk: Hang on, found a file and Pillow can open and show it: http://www.realtimerendering.com/resources/GraphicsGems/gemsiv/mrsfoley.im

> * copyright (c) by Alan Paeth, 1993
> * general redistribution for public, non-profit use is granted by the author

And a bit of info [here](https://groups.google.com/forum/#!search/width$20height$20%22pixel$20n8%22/comp.graphics/VXyVCT5i8pQ/MiUhnFcpK_8J).

PR #2175, opened 2016-10-25
Fix too-short title underlines and malformed table

hugovk:
e.g. before https://travis-ci.org/python-pillow/Pillow/jobs/170394881#L3453


radarhere:
Considering that I caused for some of this, thanks!

PR #1361, opened 2015-07-30
Test on Python 3.5 dev and 3.6 nightly

hugovk:
I noticed Travis CI now supports nightly and dev releases of Python:

> **Nightly build support**

> Travis CI supports a special version name `nightly`, which points to a recent development version of CPython build.

> **On-demand installations**

> For a limited number of Python development releases, on-demand installation is available.

> Currently, these are: 3.5.0b2, 3.5.0b3, and 3.5-dev (built nightly).

http://docs.travis-ci.com/user/languages/python/#Nightly-build-support

Right now, `nightly` is Python 3.6.0a0. I'd suggest the 3.5 betas aren't as useful as `3.5-dev`, but let's include `3.5-dev` and `nightly`. It could help us identify breaking changes in new Python versions before main releases. Of course, we don't want them to actually break our builds, so they are in the `allow_failures` section.

See also discussion in https://github.com/python-pillow/Pillow/pull/1354.

PR #2177, opened 2016-10-25
Build: Test on Python 3.5 dev and 3.6 dev

hugovk:
The `nightly` build was `3.6-dev`, but has now moved on to `3.7-dev` (right now 3.7.0a0). Until 3.6 is released later this year, let's  explicitly test against `3.6-dev`.

We already test against `3.5`, is there any value in also testing against `3.5-dev`?

`3.5-dev`, `3.6-dev` and `nightly` are all in `allow_failures` so if they fail (possibly due to errors in Python itself, possibly due to some change we'll need to support) they won't fail our builds, but we'll know about them.

(`nightly` was first added in https://github.com/python-pillow/Pillow/pull/1361)

---

[Travis CI docs](https://docs.travis-ci.com/user/languages/python/#Choosing-Python-versions-to-test-against) say:

> Travis CI supports Python versions 2.6, 2.7, 3.2, 3.3, 3.4, 3.5, as well as recent development versions.
> 
> ``` yaml
> language: python
> python:
>   - "2.6"
>   - "2.7"
>   - "3.2"
>   - "3.3"
>   - "3.4"
>   - "3.5"
>   - "3.5-dev" # 3.5 development branch
>   - "3.6-dev" # 3.6 development branch
>   - "nightly" # currently points to 3.7-dev
> # command to install dependencies
> install: "pip install -r requirements.txt"
> # command to run tests
> script: nosetests
> ```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,8 @@ python:
   - 3.2
   - 3.3
   - 3.4
+  - 3.5-dev
+  - 3.6-dev
   - nightly
 
 install:
@@ -127,6 +129,8 @@ after_script:
 matrix:
   fast_finish: true
   allow_failures:
+    - python: 3.5-dev
+    - python: 3.6-dev
     - python: nightly
 
 env:


wiredfool:
If we're going to be adding extra items that slow down our tests, especially when we have a bunch stacked up around release time, then I'd rather have them be ones that add actionable data through real diversity that might prevent releasing a bug, rather than adding additional may-fail pythons.

PR #2182, opened 2016-10-27
Fix for ImageQt Segfault

### LINKED ISSUE #1370: Pillow 2.9 causes PySide application to crash (whereas 2.8.2 does not)
Hello, the program below causes a crash on line 24 (setPixmap() line). It does not crash when I use Pillow 2.8.2. What might be the problem?

``` python
import sys
from PySide import QtGui, QtCore
from PIL import Image
from PIL import ImageQt


class Example(QtGui.QWidget):

    def __init__(self):
        super(Example, self).__init__()

        pil_image = Image.open("test.png")

        print("Creating ImageQt.")
        qimage = ImageQt.ImageQt(pil_image)

        print("Converting to QPixmap.")
        pixmap1 = QtGui.QPixmap.fromImage(qimage)
        print("Converted QPixmap.")

        hbox = QtGui.QHBoxLayout(self)

        lbl = QtGui.QLabel(self)
        lbl.setPixmap(pixmap1.copy())

        hbox.addWidget(lbl)
        self.setLayout(hbox)

        self.setGeometry(300, 300, 280, 170)
        self.show()


def main():
    app = QtGui.QApplication(sys.argv)
    ex = Example()
    sys.exit(app.exec_())


if __name__ == '__main__':
    main()
```

wiredfool:
Fixes #1370 .

- Retain a reference to the data passed into the QImage constructor.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,7 +49,7 @@ install:
   - pushd depends && ./install_imagequant.sh && popd
 
 before_script:
-# QT needs a display for some of the tests, and it's only run on the system site packages install
+# Qt needs a display for some of the tests, and it's only run on the system site packages install
   - "export DISPLAY=:99.0"
   - "sh -e /etc/init.d/xvfb start"
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,6 +48,11 @@ install:
   # libimagequant
   - pushd depends && ./install_imagequant.sh && popd
 
+before_script:
+# QT needs a display for some of the tests, and it's only run on the system site packages install
+  - "export DISPLAY=:99.0"
+  - "sh -e /etc/init.d/xvfb start"
+
 script:
   - if [ "$TRAVIS_PYTHON_VERSION" != "nightly" ]; then coverage erase; fi
   - python setup.py clean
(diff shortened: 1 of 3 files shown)


hugovk:
Fails on 2.7_with_system_site_packages, which is the only one with Qt (PyQt4) enabled (info:  https://github.com/python-pillow/Pillow/pull/814):

`TestToQImage.test_segfault ... : cannot connect to X server`

The other builds skip this test:

`TestToQImage.test_segfault ... SKIP: Qt bindings are not installed`

wiredfool:
Yeah, I saw that. We apparently have no effective tests on the Qt stuff on travis, since it requires context that we don't have there.

wiredfool:
Ok, added an X server to travis. It only took 5 tries to get the yml syntax right, which are now squashed.

.travis.yml
```diff
@@ -48,6 +48,11 @@ install:
   # libimagequant
   - pushd depends && ./install_imagequant.sh && popd
 
+before_script:
+# QT needs a display for some of the tests, and it's only run on the system site packages install
```
hugovk: "Qt" rather than "QT". 

And is this needed for all build jobs, or can we limit it to just 2.7_with_system_site_packages?

wiredfool: No. It's not needed. However. It takes .01 seconds to start, (https://travis-ci.org/python-pillow/Pillow/jobs/171335292#L1933)  and that's far less time than I spent faffing around to try to get the yml syntax right.

PR #2184, opened 2016-10-31
Replace type() equality checks with isinstance

jdufresne:
- Use more Pythonic "isinstance()" in places where a variable's type is being checked.


Tests/test_imagemath.py
```diff
@@ -9,7 +9,7 @@ def pixel(im):
     if hasattr(im, "im"):
         return "%s %r" % (im.mode, im.getpixel((0, 0)))
     else:
-        if isinstance(im, type(0)):
+        if isinstance(im, bool):
```
jdufresne: Fixed in latest version. Thanks.

hugovk:
Thanks!

PR #2187, opened 2016-11-01
Avoid deprecated 'U' mode when opening files

jdufresne:
<s>Instead, use [io.open()](https://docs.python.org/2/library/io.html#io.open) wrapper to handle universal newlines for all Python versions.</s> (Different approach taken, see below.)

Fixes warning "DeprecationWarning: 'U' mode is deprecated" in tests.

### DIFF
--- a/Tests/test_file_eps.py
+++ b/Tests/test_file_eps.py
@@ -1,7 +1,6 @@
 from helper import unittest, PillowTestCase, hopper
 
 from PIL import Image, EpsImagePlugin
-from PIL._util import py3
 import io
 
 # Our two EPS test files (they are identical except for their bounding boxes)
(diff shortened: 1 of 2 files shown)


wiredfool:
IIRC, there was some pain around trying this previously.

jdufresne:
Do you have any information on the previous attempt? If an edge case does exist, it doesn't look to be captured in a test.

wiredfool:
I think it's from the pr around https://github.com/python-pillow/Pillow/commit/9b35a4538d4cb121806f41c4338ac50a83107e5c

Looking around that era, it does look like the U was just carried forward at that time, rather than being added then.

jdufresne:
Looking at the [EPS spec](http://partners.adobe.com/public/developer/en/ps/5002.EPSF_Spec.pdf), it looks like Python's "universal new lines" will always be incorrect. From the spec: 

> 2.9 Miscellaneous Constraints

> ...

> Lines must be terminated with one of the following combinations of  characters: CR, LF, CR LF, or LF CR.

> CR is the carriage return character and LF is the line feed character (decimal  ASCII 13 and 10, respectively).

[Python's universal newlines](https://docs.python.org/3/glossary.html#term-universal-newlines) does not handle the '\n\r' case.

Therefore, I think the `PSFile` wrapper should be used in _all_ cases.

I have updated the PR with this chage as well as updated the tests to now test against all combinations of line endings.

jdufresne:
As a nice side effect, this change also fixes the orphaned open file that is never explicitly closed. Thus leading to fewer `ResourceWarning`s during tests with warnings enabled.

jdufresne:
Rebased to latest master.

hugovk:
I suggest we merge this after this quarterly release so it has some time to sit in master.

jdufresne:
Rebased to resolve merge conflicts.

PR #2188, opened 2016-11-01
Remove unused, open files at top level of tests.

jdufresne:
The data read from the file was unused. The files remained opened and
were never explicitly closed.

Fixes some instances of warnings during tests:

"ResourceWarning: unclosed file ..."


hugovk:
Thanks!

Issue #1637, opened 2016-01-04
Drop support for Python 3.2?


hugovk:
It was suggested [here](https://github.com/python-pillow/Pillow/issues/1563#issuecomment-168641684) to drop support for Python 3.2:

> I'm kind of willing to let Python 3.2 go, It's 5 years old with three newer more compatible successors. I'd say drop it and if there are major complaints we can revisit.

Anyone for or against this?

hugovk:
Here's some stats for Python 3.2.

The last release Pillow 3.0.0 was downloaded from PyPI 1,047,167 times of which 83% was the zipped source. Excluding source, Python 3.2 accounts for 2% of the packages downloaded:

![image](https://cloud.githubusercontent.com/assets/1324225/12089054/6c3a3bc2-b2e9-11e5-97c7-9ed134e8a765.png)

![image](https://cloud.githubusercontent.com/assets/1324225/12089066/7c13c20c-b2e9-11e5-9094-0d8fa4889182.png)

This [2014](http://www.randalolson.com/2015/01/30/python-usage-survey-2014/) [survey](https://docs.google.com/forms/d/1DqxkNi4GvyTCu54usSdE1DjW29zw1tc52iMeH3z4heg/viewanalytics) said 2.8% of 6,746 respondents regularly use Python 3.2:
![image](https://cloud.githubusercontent.com/assets/1324225/12089281/d5c8c692-b2eb-11e5-8999-e9277f8604eb.png)

This [2015 survey](https://astrofrog.github.io/blog/2015/05/09/2015-survey-results/) said <1% of 786 respondents use Python 3.2 as their *primary* installation:
![image](https://cloud.githubusercontent.com/assets/1324225/12089282/dbc81f0c-b2eb-11e5-906b-a5e652fb5b23.png)

wiredfool:
To be clear, I was referring to Windows 3.2 builds, not overall Pillow compatibility with 3.2. Pillow still works with 3.2, and I'm not aware of anything that we're chomping at the bit to add if we dropped 3.2. (Though, not being able to use u'' strings in tests has tripped me up. )

hugovk:
Fair enough.

Windows packages account for 3,211/3,771 of the Python 3.2 packages for Pillow 3.0.0.

wiredfool:
My guess is that we're going to trip up more ci systems than actual users.

Issue #1087, opened 2015-01-23
Release Pillow 3.0 after 2.9?


hugovk:
One [CPython core developer](http://www.curiousefficiency.org/posts/2014/08/python-4000.html) wrote last year:

> My current expectation is that Python 4.0 will merely be "the release that comes after Python 3.9". That's it. No profound changes to the language, no major backwards compatibility breaks - going from Python 3.9 to 4.0 should be as uneventful as going from Python 3.3 to 3.4 (or from 2.6 to 2.7). I even expect the stable Application Binary Interface (as first defined in PEP 384) to be preserved across the boundary.

We could consider following the Python core versioning scheme. After 2.9 comes 3.0, not 2.10. Just another release. It fits well with the "release whatever is ready each quarter". 

I know version numbers aren't a decimal number (that second dot in the full x.y.z is a giveaway), but it's that little bit clearer to avoid double digit version numbers. No need to wonder if 2.11 is later that 2.2, no bugs caused by dodgy coding (see why [Windows 9 was skipped](https://searchcode.com/?q=startswith+Windows+9)).

(Python 2.7 is a bit of a special case because it's hanging around for maintenance releases until at least [2020](https://www.python.org/dev/peps/pep-0373/) and will soon hit [2.7.10](http://comments.gmane.org/gmane.comp.python.devel/148310) as can't really roll over to 2.8/2.9.)

Is the release after 2.9 big or important enough to be given a 3.0 version number? Well, 10 releases of Pillow is a great achievement in itself, especially with quarterly releases. 

Having said that, if anything "big" comes along before 2.9, it might make sense to bump up to 3.0. For example Python 3 support warranted the bump to Pillow 2.0.

Thoughts?

aclark4life:
I'm not sure about other features/changes/etc, but if renaming the ``Pillow`` distribution to ``pillow`` proves to be desirable, then I'd be comfortable doing it then.

And in general, I'm fine with "3.0 comes after 2.9" vs. 2.10.

nicktimko:
How about also renaming modules/packages to lower-case versions a la Python 2's `Queue`, `SimpleHTTPServer`, `StringIO`...

hugovk:
I'm not so keen on renaming just to follow conventions, as in this case it would break lots of existing code that uses Pillow, and to support multiple versions require lots of ugly code like this:

    try: 
        import Xyz
    except ImportError:
        import xyz

It's basically an API break that can be avoided.

[...]

PR #2192, opened 2016-11-04
Drop support for Python 2.6

jdufresne:
I would humbly like to suggest Pillow drop support for Python 2.6.

Python 2.6 has had its last release and been end of life for over [3 years](https://www.python.org/dev/peps/pep-0361/). By removing support, Pillow can move forward with the greater Python community by adopting newer language features and be more forward compatible. pip, the Python package manager, has recently announced they are [dropping 2.6 support](https://mail.python.org/pipermail/distutils-sig/2016-November/029785.html).

I've tried my best to remove as much 2.6 specific code as I can, including the `'Programming Language :: Python :: 2.6'` trove classifier from `setup.py`.

Interestingly, there are some existing uses of `str.format()` [automatic](https://github.com/jdufresne/Pillow/blob/5aeb0ed9727a8c51c98d11788c0135c54e88b1b7/PIL/IcnsImagePlugin.py#L317) [numbering](https://github.com/jdufresne/Pillow/blob/5aeb0ed9727a8c51c98d11788c0135c54e88b1b7/Tests/test_image_resample.py#L205) in the code. Those parts are already incompatible with 2.6.

Thanks for considering.

Code changes in this PR include: 

* Drop unittest2 requirement
* Use set literals
* Use dict/set comprehension
* Use `str.format()` automatic numbering

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -17,7 +17,7 @@ Notes
 
 .. note:: Pillow >= 2.0.0 < 3.5.0 supports Python versions 2.6, 2.7, 3.2, 3.3, 3.4, 3.5
 
-.. note:: Pillow >= 3.5.0 supports Python versions 2.7, 3.2, 3.3, 3.4, 3.5
+.. note:: Pillow >= 3.5.0 supports Python versions 2.7, 3.3, 3.4, 3.5
 
 Basic Installation
 ------------------
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -15,7 +15,9 @@ Notes
 
 .. note:: Pillow < 2.0.0 supports Python versions 2.4, 2.5, 2.6, 2.7.
 
-.. note:: Pillow >= 2.0.0 supports Python versions 2.6, 2.7, 3.2, 3.3, 3.4, 3.5
+.. note:: Pillow >= 2.0.0 < 3.5.0 supports Python versions 2.6, 2.7, 3.2, 3.3, 3.4, 3.5
+
+.. note:: Pillow >= 3.5.0 supports Python versions 2.7, 3.2, 3.3, 3.4, 3.5
 
 Basic Installation
 ------------------
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,6 @@ python:
   - 3.5
   - 2.7
   - "2.7_with_system_site_packages" # For PyQt4
-  - 3.2
   - 3.3
   - 3.4
   - nightly
@@ -26,9 +25,7 @@ install:
     # Pyroma tests sometimes hang on PyPy; skip
   - if [ $TRAVIS_PYTHON_VERSION != "pypy" ]; then travis_retry pip install pyroma; fi
 
-  # Coverage 4.0 doesn't support Python 3.2
-  - if [ "$TRAVIS_PYTHON_VERSION" == "3.2" ]; then travis_retry pip install coverage==3.7.1; fi
-  - if [ "$TRAVIS_PYTHON_VERSION" != "3.2" ]; then travis_retry pip install coverage; fi
+  - "travis_retry pip install coverage"
 
   # docs only on python 2.7
   - if [ "$TRAVIS_PYTHON_VERSION" == "2.7" ]; then travis_retry pip install -r requirements.txt ; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,6 @@ python:
   - "pypy3"
   - 3.5
   - 2.7
-  - 2.6
   - "2.7_with_system_site_packages" # For PyQt4
   - 3.2
   - 3.3
@@ -24,10 +23,8 @@ install:
   - "travis_retry pip install cffi"
   - "travis_retry pip install nose"
   - "travis_retry pip install check-manifest"
-    # Pyroma tests sometimes hang on PyPy and Python 2.6; skip for those
-  - if [ $TRAVIS_PYTHON_VERSION != "pypy" && $TRAVIS_PYTHON_VERSION != "2.6" ]; then travis_retry pip install pyroma; fi
-
-  - if [ "$TRAVIS_PYTHON_VERSION" == "2.6" ]; then travis_retry pip install unittest2; fi
+    # Pyroma tests sometimes hang on PyPy; skip
+  - if [ $TRAVIS_PYTHON_VERSION != "pypy" ]; then travis_retry pip install pyroma; fi
 
   # Coverage 4.0 doesn't support Python 3.2
   - if [ "$TRAVIS_PYTHON_VERSION" == "3.2" ]; then travis_retry pip install coverage==3.7.1; fi
--- a/PIL/BmpImagePlugin.py
+++ b/PIL/BmpImagePlugin.py
@@ -73,7 +73,7 @@ class BmpImageFile(ImageFile.ImageFile):
         read, seek = self.fp.read, self.fp.seek
         if header:
             seek(header)
-        file_info = dict()
+        file_info = {}
         file_info['header_size'] = i32(read(4))  # read bmp header size @offset 14 (this is part of the header size)
         file_info['direction'] = -1
         # --------------------- If requested, read header at a specific position
--- a/PIL/IcoImagePlugin.py
+++ b/PIL/IcoImagePlugin.py
@@ -139,7 +139,7 @@ class IcoFile(object):
         """
         Get a list of all available icon sizes and color depths.
         """
-        return set((h['width'], h['height']) for h in self.entry)
+        return {(h['width'], h['height']) for h in self.entry}
 
     def getimage(self, size, bpp=False):
         """
--- a/PIL/JpegImagePlugin.py
+++ b/PIL/JpegImagePlugin.py
@@ -408,7 +408,7 @@ def _fixup_dict(src_dict):
         except: pass
         return value
 
-    return dict((k, _fixup(v)) for k, v in src_dict.items())
+    return {k: _fixup(v) for k, v in src_dict.items()}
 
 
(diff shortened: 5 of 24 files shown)


hugovk:
See https://github.com/python-pillow/Pillow/issues/1637 for some stats form January this year.

A quick check of PyPI downloads for Pillow's last release, 3.4.2 made 18 days ago, suggests only 0.68% (1,401/204,807) were for a Python 2.6 package.

\*(`vanity pillow -p 3\.4\.2` and `vanity pillow -p 3\.4\.2.*2[\.]?6`)

hugovk:
https://gist.github.com/juanpabloaj/dffc6900f80abcfe8ce121a39cffa743 has more recent stats:

> # Total of pip packages downloaded separated by Python versions
>
> From June 26, 2016 (python 3.5.2 release) to Aug. 31, 2016.
>
> Python versions from 2.6 to 3.5
> Python versions from 2.6 to 3.5
>
> Without 2.7
> 
Without 2.7
...
> Here's a log scale version of the same data.
> 
Here's a log scale version of the same data

wiredfool:
Not sure where the 3.1 data comes from, since we don't actually produce a binary for it and IIRC we've never been compatible with it. 

I'd be willing to have a discussion on dropping 3.2 and 2.6. They're both ancient, and they both are significant additions to the test matrix in terms of platforms that break. 

From what I can tell, 2.6 isn't shipped anywhere, and is only in use by people who don't care about updates, or can't update for some external reason. They're implicitly or explicitly taking on backporting duties for a whole bunch of other things so, I'd be willing to push the job down to them. (unless @manisandro objects, as the only possible people shipping 2.6 would be with some enterprise supported RH type product).

3.2 was shipped with a currently supported Ubuntu LTS release (12.04), but clearly not many people are downloading from pypi on that platform.

manisandro:
@wiredfool No objections here (though I only can speak for Fedora side since I'm not involved with RHEL).

hugovk:
The graphs on this page that include 3.1 are for all PyPI packages, not just Pillow.

wiredfool:
Ah, thanks @hugovk.  If that's the case, then we probably account for some small percentage of those downloads due to pip installing packages when running our 3.2 ci.

hugovk:
Yep, I bet CIs in general account for a large number of 2.6 and 3.2 downloads.

I'd also be fine with removing 2.6 and especially 3.2. If it's trivial to support, it doesn't matter. But definitely 3.2 is the most troublesome and more and more libraries are dropping 3.2 support, soon it will be harder for us too. 

2.6 doesn't seem that difficult to support (at the moment), still a few edge cases. But if it's not being used much, and no longer supported, there's less point in even doing so.

Ping also @python-pillow/pillow-team.

aclark4life:
👍 2.7 should probably be the only 2.x we support at this point.

homm:
> I'd be willing to have a discussion on dropping 3.2 and 2.6. 

👍

hugovk:
Here's some stats of Pillow downloads, for 3.2.0 (released 1st April and the most recent release to be out the full three months before the next release) and the most recent 3.4.2 (released 18 days ago).

Pillow 3.2.0
image

Pillow 3.4.2
image

image

Notes:
 * These percentages are excluding source-only packages. If source were included, source alone accounts for 87% for Pillow 3.2.0 and 74% for Pillow 3.4.2.
 * Python 3.2 has gone to zero for 3.4.2 because we don't make binary packages for that any more.

homm:
I believe than 90% of Python 2.6 users is CIs

jdufresne:
Would you like me to update this PR/commit to drop 3.2 at the same time? I have no problem making that change. Thanks.

aclark4life:
@jdufresne Unless someone objects, I'd say yes let's get 2.6 & 3.2 removal done in one PR

jdufresne:
> Unless someone objects, I'd say yes let's get 2.6 & 3.2 removal done in one PR

Done in the updated commit. All additional feedback welcome. Thanks.

docs/installation.rst
```diff
@@ -340,44 +342,40 @@ current versions of Linux, macOS, and Windows.
...
 | Windows Server 2012 R2           |Yes          | 2.7,3.3,3.4                  | 3.4.1 (CI target)              |x86,x86-64             |
 +----------------------------------+-------------+------------------------------+--------------------------------+-----------------------+
-| Windows 8 Pro                    |Yes          | 2.6,2.7,3.2,3.3,3.4a3        | 2.2.0                          |x86,x86-64             |
+| Windows 8 Pro                    |Yes          | 2.7,3.3,3.4a3                | 2.2.0                          |x86,x86-64             |
 +----------------------------------+-------------+------------------------------+--------------------------------+-----------------------+
-| Windows 8.1 Pro                  |Yes          | 2.6,2.7,3.2,3.3,3.4          | 2.4.0                          |x86,x86-64             |
+| Windows 8.1 Pro                  |Yes          | 2.7,3.3,3.4                  | 2.4.0                          |x86,x86-64             |
 +----------------------------------+-------------+------------------------------+--------------------------------+-----------------------+
```
hugovk: Might it still be useful leaving the historical version support here? If someone still uses Python 2.6 or 3.2, they can still see here what versions were good on what platforms. Of course, we won't bother testing 2.6 or 3.2 for future ones.

hugovk:
It might be worth bumping the next Pillow release to 4.0.0 to make it abundantly clear it's not backwards compatible with Python 2.6 and 3.2. 

It'd follow suit with Pillow 2.0.0 dropping Python versions 2.4 and 2.5, and also be much easier to remember.

Thoughts?

jdufresne:
> 

Might it still be useful leaving the historical version support here? If someone still uses Python 2.6 or 3.2, they can still see here what versions were good on what platforms. Of course, we won't bother testing 2.6 or 3.2 for future ones.

Reverted this doc change in the latest version.

> It might be worth bumping the next Pillow release to 4.0.0 to make it abundantly clear it's not backwards compatible with Python 2.6 and 3.2. 

I agree. I think it helps make it more clear to library consumers. I wasn't sure what the exact version policy was when I originally wrote the change. I have made this change in the latest version. If there is any disagreement about changing the version, let me know and I can revert it.

homm:
1. @jdufresne It would be better to remove 2.6 and 3.2 at least in two separate commits.
2. A good practice is not to modify changelog and versions in Pull requests. This information not related to code changes itself, it is more an administrative question. It's more conveniently to see version changes as separate commits in master. Also, as you can see, the conflicts probability is high.
3. `3.0` doesn't introduce significant changes or backward incompatibilities. In [the previous discussion](https://github.com/python-pillow/Pillow/issues/1087) about versioning policy, there was a decision to do "aggressive but fair deprecation" not tied to version numbers. Also, was decided to remove feature after one year after depreciation. I suppose we must adhere to the decisions made or change decisions in the more explicit way.

depends/ubuntu_12.04.sh
```diff
@@ -2,7 +2,7 @@
 
 #
 # Installs all of the dependencies for Pillow for Ubuntu 12.04
-# for both system Pythons 2.7 and 3.2
```
wiredfool: This is incorrect, as the system python3 is version 3.2

appveyor.yml
```diff
@@ -1,68 +1,68 @@
-version: 3.5.pre.{build}
```
wiredfool: Please don't bulk change line endings.

jdufresne:
Integrated feedback in latest version. All additional feedback is welcome.

>  It would be better to remove 2.6 and 3.2 at least in two separate commits.

Split changes into two commits.

> A good practice is not to modify changelog and versions in Pull requests.

I have removed the changes to `CHANGES.rst` and the version bump. I'll leave these as administrative tasks.

> This is incorrect, as the system python3 is version 3.2

Reverted this change.

> Please don't bulk change line endings.

Reverted this changed (was only related to the version bump)

jdufresne:
> 3.0 doesn't introduce significant changes or backward incompatibilities. In the previous discussion about versioning policy, there was a decision to do "aggressive but fair deprecation" not tied to version numbers. Also, was decided to remove feature after one year after depreciation. I suppose we must adhere to the decisions made or change decisions in the more explicit way.

Sorry, it isn't clear if this is asking me to make an explicit modification to my PR to make it more merge friendly for the project. If you are requesting a specific change, can you please state it explicitly?

aclark4life:
@jdufresne Hang tight, I think @homm voiced concern about @hugovk's suggestion about bumping to 4.0.0. While I tend to like to use "big" version number changes as @hugovk suggests (as an indicator of some significant change), @homm's point about adhering to our own policies is valid.

Aside from the ticket that @homm referenced, do we have our policies written down somewhere (I assume so, but if not, we should)? I'm inclined to not worry too much about it and just drop them and move on, but let's see if we can make everyone happy by (re)making decisions about when to drop old features and what our version number policy is before we move forward.

wiredfool:
This post still reflects my thinking: https://github.com/python-pillow/Pillow/issues/1087#issuecomment-121655443

For this particular PR, we're not in a place where the packagers really matter anymore, as either they're on newer python or they're only backporting bugfixes. So, if they're maintaining an old python, they're not maintaining our compatibility with that old python for those CVE bugfixes which affect them.

jdufresne:
> This is also an invalid change. This is reporting a test result, which was with the system python3 on ubuntu 12.04

Thanks for the review. Fixed in latest version.

wiredfool:
Meged in #2192

hugovk:
This needs to go in the release notes as well.
https://github.com/python-pillow/Pillow/tree/master/docs/releasenotes

hugovk:
Bikeshed discussion about version number: https://github.com/python-pillow/Pillow/issues/2244

PR #2194, opened 2016-11-04
Close file handle in TiffImagePlugin when image is closed.

### LINKED ISSUE #2181: tiff fd leak
### What did you do?

```
PYTHONPATH=. python fdleaktest.py
```

with fdleaktest.py:

``` python
from PIL import Image
d={}
for i in range(2048):
    try:
        d[i] = Image.open("Tests/images/multipage-lastframe.tif")
        d[i].load()
        d[i].close()
    except IOError as e:
        print "Failed to open %i files" % i
        raise e
```

That the script completes without error.
### What actually happened?

```
Failed to open 1021 files
  File "../fdleak/test-3.py", line 10, in <module>
IOError: [Errno 24] Too many open files: 'Tests/images/multipage-lastframe.tif'
```

master branch abc4d55f08a61be8e1fb7815b56423c0dff268c1 and Python 2.7.12

Note: png (for example) works fine.

Note: Under Windows, this prevents such image files from being removed (because Windows does not allow to unlink opened files).

lambdafu:
Let's see if this fixes #2181

### DIFF
--- a/PIL/Image.py
+++ b/PIL/Image.py
@@ -550,10 +550,6 @@ class Image(object):
         had their file read and closed by the
         :py:meth:`~PIL.Image.Image.load` method.
         """
-        try:
-            self.fp.close()
-        except Exception as msg:
-            logger.debug("Error closing: %s", msg)
 
         # Instead of simply setting to None, we're setting up a
         # deferred error that will better explain that the core image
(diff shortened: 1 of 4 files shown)


lambdafu:
Well, at least we have a test case now...

lambdafu:
This is looking A-OK to me now!

wiredfool:
Cherry picked test into #2330

PR #2133, opened 2016-09-26
Speed up GIF save optimization step

wiredfool:
Fixes #2093 .

Changes proposed in this pull request:

 * Only optimize a gif's palette if: 
   - We have to (L mode)
   - It's got less than 128 colors (otherwise, we're just padding with zeros anyway)
   - It changed
   - The image is smaller than 1/4 mpix. (otherwise, the potential savings is lost in the noise)
 * Do it in C, using the existing conversion machinery, and some monkeying with palettes and convert.

PR #2196, opened 2016-11-05
Unified different GIF optimize conditions

radarhere:
PR #2133 changed the condition for optimizing GIFs, but it only changed it in one location out of two.

This is a suggestion for fixing #2195. Happy for any feedback, particularly from @wiredfool, since it was his PR.

### DIFF
--- a/PIL/GifImagePlugin.py
+++ b/PIL/GifImagePlugin.py
@@ -434,18 +434,16 @@ def _get_local_header(fp, im, offset, flags):
         # optimize the block away if transparent color is not used
         transparent_color_exists = True
 
-        if _get_optimize(im, im.encoderinfo):
-            used_palette_colors = _get_used_palette_colors(im)
-
+        used_palette_colors = _get_optimize(im, im.encoderinfo)
+        if used_palette_colors is not None:
             # adjust the transparency index after optimize
-            if len(used_palette_colors) < 256:
-                for i in range(len(used_palette_colors)):
-                    if used_palette_colors[i] == transparency:
-                        transparency = i
-                        transparent_color_exists = True
-                        break
-                    else:
-                        transparent_color_exists = False
+            for i in range(len(used_palette_colors)):
+                if used_palette_colors[i] == transparency:
+                    transparency = i
+                    transparent_color_exists = True
+                    break
+                else:
+                    transparent_color_exists = False
 
     if "duration" in im.encoderinfo:
         duration = int(im.encoderinfo["duration"] / 10)
@@ -552,9 +550,28 @@ def _save_netpbm(im, fp, filename):
 # --------------------------------------------------------------------
 # GIF utilities
 
-def _get_optimize(im, info):
-    return im.mode in ("P", "L") and info and info.get("optimize", 0)
+# Force optimization so that we can test performance against
+# cases where it took lots of memory and time previously.
+_FORCE_OPTIMIZE = False
 
+def _get_optimize(im, info):
+    if im.mode in ("P", "L") and info and info.get("optimize", 0):
+        # Potentially expensive operation.
+
+        # The palette saves 3 bytes per color not used, but palette
+        # lengths are restricted to 3*(2**N) bytes. Max saving would
+        # be 768 -> 6 bytes if we went all the way down to 2 colors.
+        # * If we're over 128 colors, we can't save any space.
+        # * If there aren't any holes, it's not worth collapsing.
+        # * If we have a 'large' image, the palette is in the noise.
+
+        # create the new palette if not every color is used
+        used_palette_colors = _get_used_palette_colors(im)
+        if _FORCE_OPTIMIZE or im.mode == 'L' or \
+           (len(used_palette_colors) <= 128 and
+            max(used_palette_colors) > len(used_palette_colors) and
+            im.width * im.height < 512 * 512):
+            return used_palette_colors
 
 def _get_used_palette_colors(im):
     used_palette_colors = []
@@ -586,10 +603,6 @@ def _get_header_palette(palette_bytes):
         palette_bytes += o8(0) * 3 * actual_target_size_diff
     return palette_bytes
 
-# Force optimization so that we can test performance against
-# cases where it took lots of memory and time previously. 
-_FORCE_OPTIMIZE = False
-
 def _get_palette_bytes(im, palette, info):
     if im.mode == "P":
         if palette and isinstance(palette, bytes):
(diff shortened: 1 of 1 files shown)


PIL/GifImagePlugin.py
```diff
@@ -434,18 +434,16 @@ def _get_local_header(fp, im, offset, flags):
...
-            if len(used_palette_colors) < 256:
-                for i in range(len(used_palette_colors)):
-                    if used_palette_colors[i] == transparency:
-                        transparency = i
-                        transparent_color_exists = True
-                        break
-                    else:
-                        transparent_color_exists = False
+            for i in range(len(used_palette_colors)):
+                if used_palette_colors[i] == transparency:
```
wiredfool: Apart from anything else in this PR, this should use enumerate() instead of range and [].  It's pythonic.

radarhere: Okay, thanks, I've added a commit.

wiredfool:
in general looks ok, I'd like to see testing, since review clearly didn't pickup this error the last time through. I also think that once we get working tests that cover the range of behavior that we want, a refactor would be a good idea so that we're not doing that horrid mutation of the core image object in the middle of a _get* method. Perhaps pull out optimize as a method, or rename it to compact_palette, or similar.

PIL/GifImagePlugin.py
```diff
@@ -552,9 +550,28 @@ def _save_netpbm(im, fp, filename):
...
+        # create the new palette if not every color is used
+        used_palette_colors = _get_used_palette_colors(im)
+        if _FORCE_OPTIMIZE or im.mode == 'L' or \
+           (len(used_palette_colors) <= 128 and
+            max(used_palette_colors) > len(used_palette_colors) and
+            im.width * im.height < 512 * 512):
+            return used_palette_colors
 
 def _get_used_palette_colors(im):
     used_palette_colors = []
```
homm: This can be replaced with:
```
return [i for i, count in enumerate(im.histogram()) if count]
```

PIL/GifImagePlugin.py
```diff
@@ -552,9 +550,28 @@ def _save_netpbm(im, fp, filename):
...
+        # be 768 -> 6 bytes if we went all the way down to 2 colors.
+        # * If we're over 128 colors, we can't save any space.
+        # * If there aren't any holes, it's not worth collapsing.
+        # * If we have a 'large' image, the palette is in the noise.
+
+        # create the new palette if not every color is used
+        used_palette_colors = _get_used_palette_colors(im)
+        if _FORCE_OPTIMIZE or im.mode == 'L' or \
+           (len(used_palette_colors) <= 128 and
+            max(used_palette_colors) > len(used_palette_colors) and
```
homm: It is really hard to read the code in the if clause with such indention.
Do we have any formatting policy for this case?

PIL/GifImagePlugin.py
```diff
@@ -434,18 +434,16 @@ def _get_local_header(fp, im, offset, flags):
...
-                        transparent_color_exists = True
-                        break
-                    else:
-                        transparent_color_exists = False
+            for i, palette_color in enumerate(used_palette_colors):
+                if palette_color == transparency:
+                    transparency = i
+                    transparent_color_exists = True
+                    break
+                else:
```
homm: This should be `else` clause for `for` loop (ie when break isn't called). Or, better, `transparent_color_exists = False` should become before the loop.

PIL/GifImagePlugin.py
```diff
@@ -602,79 +615,64 @@ def _get_palette_bytes(im, palette, info):
...
+        # Internally, we require 768 bytes for a palette.
+        new_palette_bytes = (palette_bytes +
+                             (768 - len(palette_bytes)) * b'\x00')
+        m_im.putpalette(new_palette_bytes)
+        m_im.palette = ImagePalette.ImagePalette("RGB",
+                                               palette=palette_bytes,
+                                               size=len(palette_bytes))
+
+        # oh gawd, this is modifying the image in place so I can pass by ref.
+        # REFACTOR SOONEST
```
homm: )))

PIL/GifImagePlugin.py
```diff
@@ -602,79 +615,64 @@ def _get_palette_bytes(im, palette, info):
...
+        im.frombytes(m_im.tobytes())
 
-        # create the new palette if not every color is used
-        if _FORCE_OPTIMIZE or im.mode == 'L' or \
-               (len(used_palette_colors) <= 128 and  
-                max(used_palette_colors) > len(used_palette_colors) and
-                im.width * im.height < 512 * 512):
```
homm: I like this formatting )

homm:
I can't realize what exactly wrong with the image from #2195 and can't reproduce the same with other GIF images, but I really want this to be fixed. This is a blocker for me to use 3.4.

wiredfool:
I'll take a look at #2195, it must have slipped by me when it came up first.

PIL/GifImagePlugin.py
radarhere: I personally don't think that replacement is clear. You should feel free to create a separate PR though, as it doesn't have anything to do with this change.

PIL/GifImagePlugin.py
radarhere: I've created a separate PR with my suggestion - #2283

PIL/GifImagePlugin.py
```diff
@@ -602,79 +615,64 @@ def _get_palette_bytes(im, palette, info):
         else:
             source_palette = bytearray([i//3 for i in range(768)])
 
-    used_palette_colors = palette_bytes = None
+    palette_bytes = None
 
-    if _get_optimize(im, info):
-        used_palette_colors = _get_used_palette_colors(im)
+    used_palette_colors = _get_optimize(im, info)
```
wiredfool: This line is horribly confusing. `_get_optimize()` does not sound like it should return either a palette modification map or None. And there's no documentation to that effect.

PIL/GifImagePlugin.py
```diff
@@ -602,79 +615,64 @@ def _get_palette_bytes(im, palette, info):
...
+    if used_palette_colors is not None:
+        palette_bytes = b""
```
wiredfool: I think this block should be hoisted into a separate method -- optimize_palette or something similar.

wiredfool:
Commented on #2195, there's enough to make a synthetic test image that triggers that issue.

PR #2197, opened 2016-11-05
Replaced range(len())

radarhere:
Prompted by this comment - https://github.com/python-pillow/Pillow/pull/2196#discussion_r86660305

### DIFF
--- a/PIL/PalmImagePlugin.py
+++ b/PIL/PalmImagePlugin.py
@@ -80,13 +80,12 @@ _Palm8BitColormapValues = (
 
 # so build a prototype image to be used for palette resampling
 def build_prototype_image():
-    image = Image.new("L", (1, len(_Palm8BitColormapValues),))
+    image = Image.new("L", (1, len(_Palm8BitColormapValues)))
     image.putdata(list(range(len(_Palm8BitColormapValues))))
     palettedata = ()
-    for i in range(len(_Palm8BitColormapValues)):
-        palettedata = palettedata + _Palm8BitColormapValues[i]
-    for i in range(256 - len(_Palm8BitColormapValues)):
-        palettedata = palettedata + (0, 0, 0)
+    for colormapValue in _Palm8BitColormapValues:
+        palettedata += colormapValue
+    palettedata += (0, 0, 0)*(256 - len(_Palm8BitColormapValues))
     image.putpalette(palettedata)
     return image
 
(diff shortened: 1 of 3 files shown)


PIL/PalmImagePlugin.py
```diff
@@ -83,10 +83,10 @@ def build_prototype_image():
     image = Image.new("L", (1, len(_Palm8BitColormapValues),))
     image.putdata(list(range(len(_Palm8BitColormapValues))))
     palettedata = ()
-    for i in range(len(_Palm8BitColormapValues)):
-        palettedata = palettedata + _Palm8BitColormapValues[i]
+    for colormapValue in _Palm8BitColormapValues:
+        palettedata += colormapValue
     for i in range(256 - len(_Palm8BitColormapValues)):
-        palettedata = palettedata + (0, 0, 0)
+        palettedata += (0, 0, 0)
```
wiredfool: This could probably be: 
`palettedata += (0,0,0)*(256 - len(_Palm8BitColormapValues)`

And the above loop could probably be:
`palettedata = sum(_Palm8BitColormapValues, ())`, but I'm not sure if that's clear or if it's just code golf. 

(though, if sum were renamed append...  ` def append(seq_of_tuples): return sum(seq_of_tuples,())`

PIL/PalmImagePlugin.py
```diff
@@ -80,13 +80,10 @@
...
 def build_prototype_image():
-    image = Image.new("L", (1, len(_Palm8BitColormapValues),))
+    image = Image.new("L", (1, len(_Palm8BitColormapValues)))
     image.putdata(list(range(len(_Palm8BitColormapValues))))
-    palettedata = ()
-    for colormapValue in _Palm8BitColormapValues:
-        palettedata += colormapValue
-    for i in range(256 - len(_Palm8BitColormapValues)):
-        palettedata += (0, 0, 0)
+    palettedata = sum(_Palm8BitColormapValues, ()) +\
```
wiredfool: Looking at that sum, I think it's an optimization/obscuration too far. I think it's just doing the same thing as before, but it's all hidden behind a badly named function.  It's still making N tuples, each one longer than the last.  But, it's only <=256, so it's not that bad.

radarhere: Okay, I've updated the PR.

PR #2198, opened 2016-11-05
Use more Pythonic super().__init__() instead of calling parent class directly

jdufresne:


wiredfool:
I'm not clear that super(Class, instance) is any more pythonic than the explicit parent call, at least in 2.x. 

1) it replaces an explicit MRO with an implicit MRO controlled by the child. 
2) It's ugly.

jdufresne:
Take a look at [Raymond Hettinger's article "Python’s super() considered super! "](https://rhettinger.wordpress.com/2011/05/26/super-considered-super/).

> it replaces an explicit MRO with an implicit MRO controlled by the child. 

The MRO is still explicit. It is defined on the `class` line. The MRO and `super()` use a deterministic algorithm that, once understood, should not introduce surprises.

> It's ugly.

I suppose that is subjective. As someone who uses `super()` regularly, I don't feel this way. But if the Pillow project has a coding style regarding `super()`, so be it.

wiredfool:
>The MRO is still explicit. It is defined on the class line. The MRO and super() use a deterministic >algorithm that, once understood, should not introduce surprises.

The MRO of super() is controlled at runtime, (see super() considered super, Raymond Hettinger) the MRO of the explicit parent call is determined when the code is authored. Those are two different things.

jleclanche:
Yes please, this really should be merged.

wiredfool:
I'm still -1, given the ugly syntax required and the fact that a child can mess with a parent's method resolution order with super.

jleclanche:
@wiredfool that's exactly the point. When subclassing PIL classes, one expects those PIL classes to be calling super() appropriately.

jdufresne:
This was tagged as "Needs Rebase" by @radarhere , but was also rejected by @wiredfool. If I put in the work to rebase this will it be reconsidered for inclusion?

radarhere:
If you don't feel inspired to rebase, then I wouldn't recommend doing so, particularly with an issue like this, where there is not a lot of nuance in the implementation.

wiredfool:
I think the tag was essentially a mechanical marking of all the PRs with merge errors.

homm:
@wiredfool

> it replaces an explicit MRO with an implicit MRO controlled by the child.

Certainly, but this is desired behaviour. Python is very flexible language and there should be a way to insert your class in the middle of MRO. Super allows this. 
As of ugly syntax, you are dfenately right, but this it the thing we all have to put up with for a compatibility.

jdufresne:
> If you don't feel inspired to rebase, then I wouldn't recommend doing so, particularly with an issue like 
this, where there is not a lot of nuance in the implementation.

I have not problem doing the work, I like contributing. So I feel plenty inspired. I just don't want to waste everyone's time (including my own) if it is known ahead of time that it will be rejected. FWIW, I still think using super() is preferable.

jdufresne:
Rebased to latest master.

jdufresne:
Rebased to latest master.

radarhere:
Just to note - Python 2 is no longer supported in the next Pillow release.

jdufresne:
Updated to use the `super().__init__(...)` style syntax of Python 3. This will obviously fail to Python 2 until it is removed from the test matrix. Feel free to pull this change/commit into #4109 if that is easier.

hugovk:
Any remaining objections to merging this, now the Python 2.7 ugliness has gone?

PR #2200, opened 2016-11-06
Replace try/except/fail pattern with TestCase.assertRaises()

jdufresne:
Replace pattern with the builtin support for asserting exceptions.

### DIFF
--- a/Tests/test_file_jpeg2k.py
+++ b/Tests/test_file_jpeg2k.py
@@ -174,15 +174,8 @@ class TestFileJpeg2k(PillowTestCase):
     def test_unbound_local(self):
         # prepatch, a malformed jp2 file could cause an UnboundLocalError
         # exception.
-        try:
-            jp2 = Image.open('Tests/images/unbound_variable.jp2')
-            self.assertTrue(False, 'Expecting an exception')
-        except SyntaxError as err:
-            self.assertTrue(True, 'Expecting a syntax error')
-        except IOError as err:
-            self.assertTrue(True, 'Expecting an IO error')
-        except UnboundLocalError as err:
-            self.assertTrue(False, "Prepatch error")
+        with self.assertRaises(IOError):
+            Image.open('Tests/images/unbound_variable.jp2')
 
 if __name__ == '__main__':
     unittest.main()
--- a/Tests/test_imagepath.py
+++ b/Tests/test_imagepath.py
@@ -63,7 +63,9 @@ class TestImagePath(PillowTestCase):
         self.assertEqual(list(p), [(0.0, 1.0)])
 
     def test_overflow_segfault(self):
-        try:
+        # Some Pythons fail getting the argument as an integer, and it falls
+        # through to the sequence. Seeing this on 32-bit Windows.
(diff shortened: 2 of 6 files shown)


Tests/test_file_jpeg2k.py
```diff
@@ -174,15 +174,8 @@ def test_16bit_jp2_roundtrips(self):
     def test_unbound_local(self):
         # prepatch, a malformed jp2 file could cause an UnboundLocalError
         # exception.
-        try:
-            jp2 = Image.open('Tests/images/unbound_variable.jp2')
-            self.assertTrue(False, 'Expecting an exception')
-        except SyntaxError as err:
-            self.assertTrue(True, 'Expecting a syntax error')
-        except IOError as err:
-            self.assertTrue(True, 'Expecting an IO error')
```
hugovk: This one's a bit odd in that the original allows either a SyntaxError or IOError...

Tests/test_imagepath.py
```diff
@@ -63,7 +63,9 @@ def test_path(self):
         self.assertEqual(list(p), [(0.0, 1.0)])
 
     def test_overflow_segfault(self):
-        try:
+        # Some pythons fail getting the argument as an integer, and it falls
+        # through to the sequence. Seeing this on 32bit windows.
+        with self.assertRaises((TypeError, MemoryError)):
```
hugovk: "pythons" -> "Pythons", "32bit windows" -> "32-bit Windows"

Tests/test_file_jpeg2k.py
```diff
@@ -174,15 +174,8 @@ def test_16bit_jp2_roundtrips(self):
...
-        except UnboundLocalError as err:
-            self.assertTrue(False, "Prepatch error")
+        with self.assertRaises(IOError):
+            Image.open('Tests/images/unbound_variable.jp2')
```
hugovk: The old code expects either a SyntaxError or IOError.

jdufresne: The reason I didn't continue to catch the `SyntaxError` is because [`Image.open()` catches it](https://github.com/python-pillow/Pillow/blob/5aeb0ed9727a8c51c98d11788c0135c54e88b1b7/PIL/Image.py#L2302) and then [converts it to an `IOError`](https://github.com/python-pillow/Pillow/blob/5aeb0ed9727a8c51c98d11788c0135c54e88b1b7/PIL/Image.py#L2318-L2319). So it should _never_ occur inside the test. With that, I think catching it in the test is incorrect. If you disagree, can you please explain why it should continue to be caught in the test?

jdufresne:
> "pythons" -> "Pythons", "32bit windows" -> "32-bit Windows"

Fixed in latest version.

Tests/test_file_jpeg2k.py
hugovk: Perhaps something like [builtins.open](https://github.com/python-pillow/Pillow/blob/5aeb0ed9727a8c51c98d11788c0135c54e88b1b7/PIL/Image.py#L2282) could still raise something, but I think your change is fine. Thanks.

PR #2211, opened 2016-11-10
Changed behaviour of default box argument for paste method to match docs

### LINKED ISSUE #2207: Image.paste box=None doesn't work as documented
The docs say that box having a value of `None` is "same as (0, 0)".  But [the code clearly shows](https://github.com/python-pillow/Pillow/blob/master/PIL/Image.py#L1311) that it's actually equal to (0, 0, self.size[0], self.size[1]).

The difference is an exception occurs if the image being pasted in (the one in the first argument) is not the same size as the destination image.  If instead you pass `box=(0, 0)`, you don't get an exception when the pasted image is a different size as it's just pasted at the top left regardless of it's size.

```
img = Image.new(
    mode='RGBA',
    size=(100, 100),
    color=(255,255,255,0)
)

small_img = Image.new(
    mode='RGBA',
    size=(50, 50),
    color=(255,255,255,0)
)

img.paste(small_img, box=(0,0))  #  this works fine
img.paste(small_img)  # this throws an exception
```

It looks like it's safe to just change:
```
    if box is None:
        # cover all of self
        box = (0, 0) + self.size
```
into:
```
    if box is None:
        box = (0, 0)
```

radarhere:
Resolves #2207

### DIFF
--- a/PIL/Image.py
+++ b/PIL/Image.py
@@ -1307,8 +1307,7 @@ class Image(object):
             box = None
 
         if box is None:
-            # cover all of self
-            box = (0, 0) + self.size
+            box = (0, 0)
 
         if len(box) == 2:
             # upper left corner given; get size from image or mask
@@ -1321,7 +1320,7 @@ class Image(object):
                 raise ValueError(
                     "cannot determine region size; use 4-item box"
                     )
-            box = box + (box[0]+size[0], box[1]+size[1])
+            box += (box[0]+size[0], box[1]+size[1])
 
         if isStringType(im):
             from PIL import ImageColor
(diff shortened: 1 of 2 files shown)


wiredfool:
Ok, looks like this dates back to the initial PIL import. 

* If the size of the two images is the same, then this is a noop. 
* If the images are different, then earlier code would error, as the size doesn't match. 
* The proposed code will paste in the upper left. This is functionally equivalent to `box = (0,0) + im.size` rather than `box = (0,0) + self.size`

I think this is a safe change. Tests would be good, potentially as simple as not erroring when a small image is pasted over a larger one.

tisdall:
@radarhere - it looks like you made a commit on the master that conflicts with this branch (I don't think github notifies of PR's that have conflicts, so I'm sending this message).  Thanks for making the PR!

PR #2214, opened 2016-11-11
Use TextIOWrapper.detach() instead of NoCloseStream

jdufresne:
Usage and this pattern is discussed in Python bug: https://bugs.python.org/issue21363

[detach() docs](https://docs.python.org/3/library/io.html#io.BufferedIOBase.detach)


jdufresne:
Rebased on latest mater.

PR #2216, opened 2016-11-12
If WalImageFile opens a file pointer, close

radarhere:
If a file pointer is passed in to the method, leave it open.
If the method opens the file pointer itself, then close when it is no longer needed.

### DIFF
--- a/PIL/WalImageFile.py
+++ b/PIL/WalImageFile.py
@@ -56,6 +56,7 @@ def open(filename):
         # load pixel data
         fp.seek(offset)
 
+        Image._decompression_bomb_check(size)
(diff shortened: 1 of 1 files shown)


PIL/WalImageFile.py
```diff
@@ -61,6 +63,8 @@ def open(filename):
     fp.seek(offset)
```
radarhere: Okay, I've added one. See what you think.

PIL/WalImageFile.py
```diff
@@ -61,6 +63,8 @@ def open(filename):
...
 
     im = Image.frombytes("P", size, fp.read(size[0] * size[1]))
```
radarhere: Okay, I've added a check.

PR #1390, opened 2015-08-24
Removed use of spaces in TIFF kwargs, deprecated in 2.7

radarhere:
The discussion in #1087 indicates that this is not to be merged yet, to avoid skipping over deprecation warnings.

However, I thought I'd put this up with a 'Do Not Merge' label, so that it can be merged in the future.

PR #2219, opened 2016-11-12
Drop spaces deprecation warning

jdufresne:
Deprecated in 800480db526b3c763fc821db4d29775ec9ba3c73. Included in 2.7 release. Almost 2 years ago.


wiredfool:
This needs to be in till Jessie is frozen, which should be in the first half of next year.

radarhere:
This is a duplicate of #1390

jdufresne:
Thanks for catching that. Closing.

PR #2220, opened 2016-11-13
Remove long deprecated ImageDraw methods

jdufresne:
Deprecated in 89ccf66ff7f2646c3e649d6a3bba0a7e1c38a99e since Sep 29,
2015, version 3.0.0.

I think enough time has passed to now remove these.

### DIFF
--- a/PIL/ImageDraw.py
+++ b/PIL/ImageDraw.py
@@ -31,7 +31,6 @@
 #
 
 import numbers
-import warnings
 
 from PIL import Image, ImageColor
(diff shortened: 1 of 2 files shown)


wiredfool:
Once again, we want these in one iteration of debian stable, so after jessie freezes.

radarhere:
Debian Stretch full freeze is on the 5th of February, before the next Pillow release, so this may be able to be merged now.

Just to connect the two issues, I'll note that #1390 is also related to the Debian timeline.

PR #2221, opened 2016-11-14
Add support for another type of BMP bitfield

jmerdich:
 *  Add support for a weird XBGR-bitfield BMP I found.

There's an example file [here](https://dl.dropboxusercontent.com/u/11688373/Gompei%20(1).bmp), though I don't have the rights to commit it here. It decodes fine in ImageMagick, Chrome, and Firefox, though fails in several others (GIMP and feh tested). 

As far as generating a file for test cases, not sure if I should add one directly or commit to the bmp test generator, but I can revise as needed.


wiredfool:
I'd certainly like to see a test on this one. The `g/rgb32bf.bmp` seems like it would be a file with the same structure, but headers look pretty different. I'm open to just generating one for here, but getting it into the bmpsuite would be useful for others as well.

jmerdich:
Also, I popped [a PR into bmpsuite](https://github.com/jsummers/bmpsuite/pull/5).

Issue #1907, opened 2016-05-12
Stroke color for font


controversial:
Is there a way in `ImageDraw` or `ImageFont` to draw text with a border color? If not, there probably should be. [It can be faked](http://stackoverflow.com/a/8050556/4414003) but it'd be nice if this functionality was built-in.

radarhere:
The only solutions that I'm aware of are the workaround already suggested, or finding a font file that has a shadow built in.

HinTak:
https://github.com/rougier/freetype-py have 4 examples on this.

radarhere:
Closing as a duplicate of #2209. I know this issue was first, but that one has more detail.

radarhere:
I've created PR #3978 to resolve this.

PR #2224, opened 2016-11-15
Add a feature to ImageDraw.text() to stroke text

### LINKED ISSUE #2209: ImageFont/ImageDraw should support text stroking
Hey,

Something that I find myself doing quite often when drawing text to an image is to stroke the text with a high-contrast outline.
I find this helps improve readability over cluttered backgrounds.

Neither pillow.ImageDraw nor pillow.ImageFont currently includes a function while allows ImageDraw to draw stroked text with a specific color outline.

It'd be pretty cool if ImageDraw.text() would include two additional method parameters.

- One to indicate the size of the outline in pixels
- One to indicate the colour of the outline (similar to the current fill parameter)

For example:

``draw.text((10, 10), "hello", font=font, outline=BLACK, outline_size=2)``

Irishsmurf:
Fixes #2209 and #1907 

 * Adds an `outline` argument to ImageDraw.text() & ImageDraw.multiline_text()
 * Outline variable (Default, None) will take a colour, and stroke a 1px border around the text

This change adds an additional method argument to the draw() method
which draws the text in black, before drawing it in the color selected
in the fill parameter.

`ImageDraw.text((5, 5), caption, BLACK, outline='White', font=FONT)`

Alternatively - I can do this is to break out ImageDraw.text and ImageDraw.stroke_text().
Feedback welcome.


PIL/ImageDraw.py
```diff
@@ -237,10 +236,17 @@ def text(self, xy, text, fill=None, font=None, anchor=None,
...
+                for offset in range(0, 2):
+                    tmp_xy = list(xy)
+                    for difference in [-1, 1]:
+                        tmp_xy[offset] = xy[offset] + difference
+                        self.draw.draw_bitmap(tmp_xy, mask, color)
             self.draw.draw_bitmap(xy, mask, ink)
 
     def multiline_text(self, xy, text, fill=None, font=None, anchor=None,
-                       spacing=4, align="left"):
+                       outline=None, spacing=4, align="left"):
```
hugovk: Put `outline` at the end, because changing the order of parameters can break existing user code.

Tests/test_imagefont.py
```diff
@@ -164,7 +175,7 @@ def test_render_multiline_text(self):
             # to multiline_text()
             draw.text((0, 0), TEST_TEXT, fill=None, font=ttf, anchor=None,
                       spacing=4, align="left")
-            draw.text((0, 0), TEST_TEXT, None, ttf, None, 4, "left")
+            draw.text((0, 0), TEST_TEXT, None, ttf, None, None, 4, "left")
```
hugovk: This is an example of existing user code that is broken by changing the order of parameters. This change shouldn't be necessary.

Irishsmurf: Got it.

I've gone ahead and refactored the ImageDraw.text() method signature to no longer take in **kwargs which I believe will allow for greater flexibility in future.

Other than passing the Align and Spacing args to multiline_text, I don't believe it's using anything else (Atleast it's not documented)
Moved to explicitly declaring the keyword arguments

PIL/ImageDraw.py
Irishsmurf: Done.

PIL/ImageDraw.py
```diff
@@ -217,13 +217,13 @@ def _multiline_split(self, text):
...
-             *args, **kwargs):
+    def text(self, xy, text, fill=None, font=None, anchor=None, spacing=4, 
+             align="left", outline=None):
         if self._multiline_check(text):
             return self.multiline_text(xy, text, fill, font, anchor,
-                                       *args, **kwargs)
-
+                                       spacing, align, outline)
         ink, fill = self._getink(fill)
+        print("Outline: %s" % outline)
```
hugovk: Remove print

Irishsmurf: Whoops.
Addressed.

radarhere:
I'm not necessarily sold either way on this question, but have you considered updating `textsize` to account for the different dimensions that would be generated?

wiredfool:
I think that this is the wrong approach. At the very least, it's not adjusting the spacing due to the wider letters and the stroke is fixed at one pixel. 
 
The right approach is going to take hooking into the freetype layer using FT_Stroker, as seen here: http://stackoverflow.com/a/28078293 .

radarhere:
I've created PR #3978 instead.

PR #2230, opened 2016-11-17
Lambdafu/tiff nframes

### LINKED ISSUE #2229: libtiff closes fp prematurely after n_frames, causing seek to fail
### What did you do?

```python
    TiffImagePlugin.READ_LIBTIFF = True
    with Image.open("/tmp/multipage.tiff") as im:
        for idx in range(im.n_frames):
                im.seek(idx)
                im.load()
```

The frames should be loaded one after another.

### What actually happened?

Loading the second (and any further) frame fails with ValueError('I/O operation on closed file',).

17d38d09050ea6a546fcb3ddfce189c2b82784fe

lambdafu:
Fixes #2229 

This includes a test for #2229 and in another commit a fix.  The fix causes the fp_leak test to fail, so that test case is adjusted accordingly (I think the Tiff plugin can only make a promise to close after close(), not after load(), but maybe we need another fix to close any fps zealously after load() for single-frame TIFFs).  The fix depends on #2194 to properly close all fps after close().

 * Don't close fp after load() when __next is None (this is what causes n_frames to close the fp
 * Now fp is not closed after load even for single-frame TIFFs.  This may be a regression.
 * Does not close even after close() without #2194.

### DIFF
--- a/PIL/TiffImagePlugin.py
+++ b/PIL/TiffImagePlugin.py
@@ -1096,10 +1096,6 @@ class TiffImageFile(ImageFile.ImageFile):
 
         self.tile = []
         self.readonly = 0
-        # libtiff closed the fp in a, we need to close self.fp, if possible
-        if hasattr(self.fp, 'close'):
-            if not self.__next:
-                self.fp.close()
         self.fp = None  # might be shared
 
         if err < 0:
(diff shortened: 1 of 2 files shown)


wiredfool:
I want to cross link this with #2032, (and https://github.com/wiredfool/Pillow/blob/open-files/docs/reference/open_files.rst)  #2047, and probably some others. 

Basically, unless we're looking at something that's clearly independent of the `Image.open` file path, I'd like to agree on what Pillow should be doing, then get it overall doing that, and then fix up the individual image plugins.

wiredfool:
Pulled test into #2330.

PR #2234, opened 2016-11-19
Updated comments to use print as a function

radarhere:
I know this doesn't have any meaningful effect, but hey, it will save a few keystrokes for any Python 3 users who decide to use these comments.


homm:
Please add `from __future__ import print_function` in all of these files

radarhere:
Commented?

homm:
No, not commented

radarhere:
Okay, added.

PR #2240, opened 2016-11-22
Use minimal scale for jpeg drafts

homm:
This fixes `.draft()` when requested aspect ratio doesn't match the original one.

```python
from PIL import Image
im = Image.open('dice_illusion.jpg')
im.draft(None, (150, 45))
im = im.resize((150, 45), Image.BICUBIC)
im.save('_out.jpg', quality=90)
```

Source image
dice_illusion

Results before/after
_out _out fixed

### DIFF
--- a/Tests/test_image_draft.py
+++ b/Tests/test_image_draft.py
@@ -2,35 +2,63 @@ from helper import unittest, PillowTestCase, fromstring, tostring
 
 from PIL import Image
 
-CODECS = dir(Image.core)
-FILENAME = "Tests/images/hopper.jpg"
-DATA = tostring(Image.open(FILENAME).resize((512, 512)), "JPEG")
-
-
-def draft(mode, size):
-    im = fromstring(DATA)
-    im.draft(mode, size)
-    return im
-
 
 class TestImageDraft(PillowTestCase):
-
     def setUp(self):
-        if "jpeg_encoder" not in CODECS or "jpeg_decoder" not in CODECS:
+        codecs = dir(Image.core)
+        if "jpeg_encoder" not in codecs or "jpeg_decoder" not in codecs:
             self.skipTest("jpeg support not available")
 
+    def draft_roundtrip(self, in_mode, in_size, req_mode, req_size):
+        im = Image.new(in_mode, in_size)
+        data = tostring(im, 'JPEG')
+        im = fromstring(data)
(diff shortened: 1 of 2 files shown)


homm:
Also added protection from second `.draft()` call. Otherwise:

```
$ time python ./Tests/test_image_draft.py
.E*** Error in `python': free(): invalid next size (fast): 0x000000000128c990 ***
Aborted (core dumped)
```

Tests/test_image_draft.py
```diff
@@ -2,36 +2,73 @@
...
-        self.assertEqual(draft("RGB", (32, 32)).size, (64, 64))
+        for in_size, req_size, out_size in [
+            ((435, 361), (2048, 2048), (435, 361)),  # bigger
+            ((435, 361), (435, 361), (435, 361)),  # same
+            ((128, 128), (64, 64), (64, 64)),
+            ((128, 128), (32, 32), (32, 32)),
+            ((128, 128), (16, 16), (16, 16)),
+
+            # large requested width
+            ((435, 361), (218, 128), (435, 361)),  # almost 2x
```
wiredfool: Does this mean that if we request a draft image of (218, 128), we're getting something larger than that?

homm:
> Does this mean that if we request a draft image of (218, 128), we're getting something larger than that?

Yes. And this is absolutely correct. The main idea behind the draft method is: it shouldn't truncate more information than defined. There are some details that should clarify.

First of all, this fix doesn't change the behavior when aspect ratios are the same or near the same. The draft image will be the same size as requested or larger. [Look at this](https://github.com/python-pillow/Pillow/pull/2240/files#diff-b6ed928019f4042acb209ac7ec041354R23), for example. We don't lose more information than define in the `.draft()` call.

Second, in can look weird but we can't return `(218, 181)` when requested size is `(218, 181)`. This also matches the previous behavior but required further explanation. The original image size is `(435, 361)`. When we downscale it in two times, the result image size should be `(217.5, 180.5)`. Of course, we can't return fractional pixels. So the actually returned size will be `(218, 181)`. But requested size `(218, 181)` is still greater than `(217.5, 180.5)`, so we are returning original images without downscaling to do not lose more information than defined.

Third, and the only what is really changed. When we can scale one dimension more than other, we don't scale as much as possible. Instead, we are trying do not lose more information than define :-)

PR #2245, opened 2016-11-22
FIX: fixed isOleFile issue when passing file content with len < MINIM…

Nazarii:
Fixes # .
fixed isOleFile: file() argument 1 must be encoded string without NULL bytes, not str

When calling isOleFile('Hello there') we will receive an error as method determines argument as a file but should return False instead of raising an error.

 * 
 * 
 * 

…AL_OLEFILE_SIZE

### DIFF
--- a/PIL/OleFileIO.py
+++ b/PIL/OleFileIO.py
@@ -453,6 +453,9 @@ def isOleFile(filename):
         # filename is a bytes string containing the OLE file to be parsed:
         header = filename[:len(MAGIC)]
     else:
+        if len(filename) < MINIMAL_OLEFILE_SIZE:
+            return False
+
         # string-like object: filename of file on disk
         header = open(filename, 'rb').read(len(MAGIC))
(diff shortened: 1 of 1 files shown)


hugovk:
@Nazarii This fails an existing test:
```
TestOleFileIo.test_isOleFile ... FAIL
FAIL: TestOleFileIo.test_isOleFile
  File "/home/travis/build/python-pillow/Pillow/Tests/test_olefileio.py", line 13, in test_isOleFile
    self.assertTrue(OleFileIO.isOleFile(ole_file))
AssertionError: False is not true
```
https://travis-ci.org/python-pillow/Pillow/jobs/178012505

This PR would also need a test which demonstrates the error, fails on current master and passes with your change. 

However, we may be removing olefile from this repo, see https://github.com/python-pillow/Pillow/pull/2199. Even if we don't, it might this would probably laso have to go to the main repo: https://github.com/decalage2/olefile

PR #2248, opened 2016-11-23
Use Image._new() instead of _makeself()

homm:


wiredfool:
Is there any reason to have this line anymore?

```
  _makeself = _new  # compatibility

```

It's a supposedly private function, and it doesn't show up anywhere else in our code.

homm:
I don't know, it's up to you )

hugovk:
`_makeself` was inherited from the initial PIL fork:

https://github.com/python-pillow/Pillow/blame/adfbe8323aa9daf453872fd60561a865b95beb24/PIL/Image.py#L897
https://github.com/python-pillow/Pillow/blame/adfbe8323aa9daf453872fd60561a865b95beb24/PIL/Image.py#L498

And `_makeself()` was a function in PIL 1.1.1, but was replaced with `_makeself = _new  # compatibility` in PIL 1.1.2:

(big slow-to-load diff:)

https://github.com/hugovk/PIL/commit/1f91cfd36155d207a594aea1c7379992bce59b45#diff-3fef5f960d5bd8bdfb9c6fadcf96af22L302

wiredfool:
Ah. I'd say it should go then.

homm:
Removed

PR #2249, opened 2016-11-23
EMF: support negative bounding box coordinates

glexey:
Similar to placeable WMF, EMF bounding box coordinates should be interpreted as signed integer, otherwise  opening EMF file with negative (x0,y0) fails.


wiredfool:
* If you need a signed int, just unpack a signed int rather than muck about with reading an unsigned int and  then doing math. This also applies to the old short method. 
* I'm seriously not in favor of open tests without a load. The basic test should really be one of loading an image and comparing it to a known target. 
* Is there a spec for either of these image formats that you're working off of?

glexey:
Eric,
* Well, I just followed the style of code that was already there. But, I agree with you. Done.
* Done.
* Here's some WMF/EMF reference documentation (last one is the useful one for this particular case):
https://winprotocoldoc.blob.core.windows.net/productionwindowsarchives/MS-WMF/[MS-WMF].pdf
http://wvware.sourceforge.net/caolan/index.html
http://wvware.sourceforge.net/caolan/ora-wmf.html

wiredfool:
@glexey This particular format has had little or no attention since the PIL days. So when we open the hood, I want to fix anything that's less than straight forward and add real tests. Thanks for contributing.

PR #2262, opened 2016-11-29
Allow 0 size images, Fixes #2259

### LINKED ISSUE #2259: Image of size zero is not getting initialized.
I want an image with zero size. I have a lot of use for it.

```py
>>> from PIL import Image as im
>>> import numpy as np
>>> im.fromarray(np.empty((0, 0), dtype=np.uint8))
  File "/usr/local/lib/python3.4/dist-packages/PIL/Image.py", line 2187, in fromarray
    return frombuffer(mode, size, obj, "raw", rawmode, 0, 1)
  File "/usr/local/lib/python3.4/dist-packages/PIL/Image.py", line 2114, in frombuffer
    _check_size(size)
  File "/usr/local/lib/python3.4/dist-packages/PIL/Image.py", line 2001, in _check_size
    raise ValueError("Width and Height must be > 0")
ValueError: Width and Height must be > 0
```

This was working previously. But after this [commit](https://github.com/python-pillow/Pillow/commit/445451c0b9347b50e0f603db33f196e207de470d), in the latest version (i.e. 3.4.2) it is throwing an error.

As @manthey points out in the commit, one can create a zero sized image by cropping an existing one. So this too should be allowed. 

Thanks.

wiredfool:
Fixes #2259 .

 * Allow 0x0 images to be created. Restores pre 3.4 behavior

### DIFF
--- a/PIL/Image.py
+++ b/PIL/Image.py
@@ -1994,8 +1994,8 @@ def _check_size(size):
         raise ValueError("Size must be a tuple")
     if len(size) != 2:
         raise ValueError("Size must be a tuple of length 2")
-    if size[0] <= 0 or size[1] <= 0:
-        raise ValueError("Width and Height must be > 0")
+    if size[0] < 0 or size[1] < 0:
+        raise ValueError("Width and Height must be => 0")
 
     return True
 
--- a/Tests/test_image.py
+++ b/Tests/test_image.py
@@ -256,7 +256,11 @@ class TestImage(PillowTestCase):
         with self.assertRaises(ValueError):
             Image.new('RGB', (0,))  # Tuple too short
         with self.assertRaises(ValueError):
-            Image.new('RGB', (0,0))  # w,h <= 0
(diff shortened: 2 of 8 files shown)


PIL/Image.py
```diff
@@ -1994,8 +1994,8 @@ def _check_size(size):
         raise ValueError("Size must be a tuple")
     if len(size) != 2:
         raise ValueError("Size must be a tuple of length 2")
-    if size[0] <= 0 or size[1] <= 0:
-        raise ValueError("Width and Height must be > 0")
+    if size[0] < 0 or size[1] < 0:
+        raise ValueError("Width and Height must be => 0")
```
hugovk: Change `=>` to `>=`, and might as well be a small `h` too: 

> "Width and height must be >= 0"

homm:
There are more tests required. We should check not only 0x0 construction but also basic operations against such images.

PR #2265, opened 2016-11-30
Allow saving .ico files of 256x256 instead of 255x255

### LINKED ISSUE #2264: .ico files are saved as 255x255
### What did you do?

Loading an .ico of the size 256x256 and then saving it results in an .ico file of the size 255x255. I tested it on an example .ico file: http://www.axialis.com/tutorials/iw/down.ico

The icon file should stay the same after saving. No other adjustments have been made. See the code below.

### What actually happened?

The saved .ico file ended up as 255x255 file. Because Windows doesn't support 255x255 size for .ico files, it is downscaled to 128x128 when used as desktop icons. [Example](http://i.imgur.com/fm7FTfw.png). The Chromium and Chrome icons are default ones, the Chrome Canary one was saved with Pillow.

Pillow 3.4.2, Python 3.5.2

```python
from PIL import Image
im = Image.open('down.ico')
im.save('saved.ico')
im_saved = Image.open('saved.ico')

In [2]: im.size
Out[2]: (256, 256)

In [3]: im_saved.size
Out[3]: (255, 255)
```

`(255, 255)` appears on the list of sizes at https://github.com/python-pillow/Pillow/blob/master/PIL/IcoImagePlugin.py#L45 .

hugovk:
Fixes #2264.

Allow saving icon files of 256x256 pixels.

Note that bWidth and bHeight use 0 to mean 256 pixels.

* http://fileformats.wikia.com/wiki/Icon
* https://www.daubnet.com/en/file-format-ico
* https://git.merproject.org/mer-core/qtbase/commit/5316befba2b2f63a4c4a4185a09bc6b149d3f842?view=parallel


hugovk:
See also #2267 that fixes #2266.

Issue #1276, opened 2015-06-14
PIL is 20!


hugovk:
On Monday 15th June 2015, the Python Imaging Library (PIL), from which Pillow is a fork, turns 20 years old!

From the [Image-SIG mailing list](https://mail.python.org/pipermail/image-sig/2005-June/003377.html) in June 2005:

> just fyi:

> $ cd pil
$ grep 1995-06 */*.c
libImaging/Convert.c: * 1995-06-15 fl   created
libImaging/Except.c: * 1995-06-15 fl   Created
libImaging/Geometry.c: * 1995-06-15 fl  Created
libImaging/Histo.c: * 1995-06-15 fl   Created.
libImaging/Storage.c: * 1995-06-15 fl   Created

> thanks to everyone who's used PIL for cool stuff for the last decade,
and everyone who's contributed cool stuff to PIL !

> cheers /F 

And checking again today:

```bash
[hugo:~/github/Pillow] master(3)+* ± grep 1995-06 */*.c
libImaging/Convert.c: * 1995-06-15 fl   created
libImaging/Except.c: * 1995-06-15 fl   Created
libImaging/Geometry.c: * 1995-06-15 fl  Created
libImaging/Histo.c: * 1995-06-15 fl   Created.
libImaging/Storage.c: * 1995-06-15 fl   Created
```

Thanks to original author Fredrik Lundh, and all the other contributors and users over the decades!

radarhere:
Congratulations to all users and contributors! And here's to many more years of helping Python to work with images.

aclark4life:
Awesome! :beers:

PR #2271, opened 2016-12-01
Updated copyright year

radarhere:
The next release will be next year.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -48,8 +48,8 @@ master_doc = 'index'
 
 # General information about the project.
 project = u'Pillow (PIL Fork)'
-copyright = u'1995-2016, Fredrik Lundh and Contributors, Alex Clark and Contributors'
-author = u'Fredrik Lundh and Contributors, Alex Clark and Contributors'
+copyright = u'1995-2011 Fredrik Lundh, 2010-2017 Alex Clark and Contributors'
+author = u'Fredrik Lundh, Alex Clark and Contributors'
 
 # The version info for the project you're documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the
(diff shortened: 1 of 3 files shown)


docs/conf.py
```diff
@@ -48,7 +48,7 @@
 
 # General information about the project.
 project = u'Pillow (PIL Fork)'
-copyright = u'1995-2016, Fredrik Lundh and Contributors, Alex Clark and Contributors'
+copyright = u'1995-2017, Fredrik Lundh and Contributors, Alex Clark and Contributors'
```
wiredfool: This should probably be split to 1995-2012 and 2012-2017.

wiredfool:
Don't merge till next year.

docs/conf.py
hugovk: * PIL was born in 1995: https://github.com/python-pillow/Pillow/issues/1276

* Pillow was forked from PIL (1.1.7 final): https://github.com/python-pillow/Pillow/commit/9a640e3157150e35f17354517e8c6304fc428060#diff-e4eb329834da3d36278b1b7d943b3bc9

* PIL 1.1.7 was released November 15, 2009: http://www.pythonware.com/products/pil/

* Pillow was forked in July 2010: https://github.com/python-pillow/Pillow/milestone/13

* So, 1995-2009 and 2010-2017?

hugovk: Here's the PIL copyright in the README from when it was forked: https://github.com/python-pillow/Pillow/blob/9a640e3157150e35f17354517e8c6304fc428060/README#L98-L99

wiredfool: I agree with @aclark4life. Include the PIL copyright as it was, and add additional year ranges of other contributors, which at this point can be AC + Contributors.

radarhere:
Happy new year.

PR #2273, opened 2016-12-03
Moving tests requiring libtiff to test_file_libtiff

wiredfool:
Moves tests requiring libtiff to test_file_libtiff.

Tests now all pass using 
`python setup.py build_ext --disable-tiff install && make test`


wiredfool:
rebased

PR #2281, opened 2016-12-09
Add note about aspect ratio to documentation

wilsonge:
Adds a note about aspect ratio's into the docs for the thumbnail example. This is described in the main function description further down but isn't noted in the example which can lead to panic attacks when you're skim reading :)

### DIFF
--- a/docs/reference/Image.rst
+++ b/docs/reference/Image.rst
@@ -26,7 +26,9 @@ Open, rotate, and display an image (using the default viewer)
     im.rotate(45).show()
 
 The following script creates nice 128x128 thumbnails of all JPEG images in the
-current directory.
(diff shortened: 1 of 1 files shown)


homm:
Technically, the result can be less than 128x128 even if the source image is square: in the case when the source image itself less than 128x128. So, my suggestion is something like this:

> The following script creates nice thumbnails of all JPEG images in the
 current directory preserving aspect ratios with 128x128 max resolution.

wilsonge:
Oops missed the response here. Amended to your suggestion :)

homm:
@wiredfool Is it ok?

wiredfool:
Yeah, it looks good to me.

PR #2284, opened 2016-12-11
updated version of Add complex text support #1682

Fahad-Alsaidi:
This pull request adds support for languages that require complex text layout.

We are using the Raqm library, that wraps FriBidi (for bidirectional
text support) and HarfBuzz (for text shaping), and does proper BiDi and script
itemization:
https://github.com/HOST-Oman/libraqm

This should fix #1089, #2255

@andreymal

PR #2289, opened 2016-12-15
switch to ubuntu trusty 14.04

Fahad-Alsaidi:
This required for Adding CTL support #2284.

### DIFF
--- a/Tests/test_file_icns.py
+++ b/Tests/test_file_icns.py
@@ -4,6 +4,12 @@ from PIL import Image
 
 import sys
 
+# The the default version of PyPy in Trusty is PyPy 5.4.1.
+# There is a known error in it: TypeError: PyPy does not yet implement the new buffer interface (issue #2163).
+
+SKIP_5_4_1_PYPY = hasattr(sys, 'pypy_version_info') and (
+    sys.pypy_version_info >= (5, 4, 1, 'final', 0))
+
 # sample icon file
 TEST_FILE = "Tests/images/pillow.icns"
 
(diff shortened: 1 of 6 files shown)


Fahad-Alsaidi:
I need help to fix this:
```

ERROR: TestImagePath.test_path

  File "/home/travis/build/python-pillow/Pillow/Tests/test_imagepath.py", line 62, in test_path

    p = ImagePath.Path(tuple(arr.tostring()))

```

Tests/test_file_icns.py
```diff
@@ -4,6 +4,12 @@
 
 import sys
 
+# The the default version of PyPy in Trusty is PyPy 5.4.1.
+# There is a known error in it: TypeError: PyPy does not yet implement the new buffer interface (issue #2163).
+
+SKIP_5_4_1_PYPY = hasattr(sys, 'pypy_version_info') and (
+    sys.pypy_version_info >= (5, 4, 1, 'final', 0))
```
wiredfool: Lets make this a little tighter -- either just 5.4.1 or 5.4.0 < v <5.4.1. 

The default case should be that we have to reaffirm this skip for each pypy that fails, not that we never run these tests again.

wiredfool:
The getargs thing is a real bug in the C layer.

wiredfool:
Ok, I was wrong, that's the same underlying issue. 

I've confirmed that there's a regression in pypy 5.4 and 5.6, where it worked in 5.3. There's a workaround in #2294.

wiredfool:
Closed in favor of #2294

hugovk:
Merged via PR https://github.com/python-pillow/Pillow/pull/2294.

PR #2296, opened 2016-12-20
Fix Access to be reloadable -- Issue #2268

### LINKED ISSUE #2268: Loading the PIL.Image module twice during the same process causes AccessInit: hash collision
### What did you do?

We have an embedded c++ Python program that runs scripts in a clean Python environment by always creating a fresh environment with Py_InitializeEx and Py_Finalize. I tried adding some image processing with PIL.Image but that causes the application to crash on the second run of the script. I traced this down to the import throwing an AccessInit: hash collision exception.

The module should load properly after the Python state has been cleared and re-created.

### What actually happened?

The module failed to load on the second run.

Pillow 3.4.2
Python 2.7.11 32-bit
Windows 10

### Code to reproduce

```cpp
#include "Python.h"

int main(int argc, char* argv[])
{
	Py_SetPythonHome( "c:\\Python27" );

	Py_InitializeEx( 0 );
	Py_DECREF(PyImport_ImportModule( "PIL.Image" ));
	Py_Finalize();

	Py_InitializeEx( 0 );
	Py_DECREF(PyImport_ImportModule( "PIL.Image" ));
	Py_Finalize();

	return 0;
}
```

Output:
```
AccessInit: hash collision: 22 for both 1 and 1
```

wiredfool:
Fixes Access to be reloaded if the python interpreter is restarted when embedded. Fixes #2268.

Test from cgohlke

### DIFF
--- a/Tests/test_image_access.py
+++ b/Tests/test_image_access.py
@@ -8,6 +8,7 @@ except ImportError:
 
 from PIL import Image
 import sys
+import os
 
 class AccessTest(PillowTestCase):
     # initial value
@@ -291,17 +292,22 @@ int main(int argc, char* argv[])
 
         compiler = ccompiler.new_compiler()
         compiler.add_include_dir(sysconfig.get_python_inc())
-        compiler.add_library_dir(sysconfig.get_config_var('LIBDIR'))
+        
+        libdir = sysconfig.get_config_var('LIBDIR') or sysconfig.get_python_inc().replace('include', 'libs')
+        print (libdir)
+        compiler.add_library_dir(libdir)
         objects = compiler.compile(['embed_pil.c'])
         compiler.link_executable(objects, 'embed_pil')
+
+        env = os.environ.copy()
+        env["PATH"] = sys.prefix + ';' + env["PATH"]
         
         # do not display the Windows Error Reporting dialog
         ctypes.windll.kernel32.SetErrorMode(0x0002)
         
-        process = subprocess.Popen(['embed_pil.exe'])
+        process = subprocess.Popen(['embed_pil.exe'], env=env)
         process.communicate()
         self.assertEqual(process.returncode, 0)
-
 
 if __name__ == '__main__':
     unittest.main()
--- a/Tests/test_image_access.py
+++ b/Tests/test_image_access.py
@@ -1,4 +1,4 @@
-from helper import unittest, PillowTestCase, hopper
+from helper import unittest, PillowTestCase, hopper, on_appveyor
 
 try:
     from PIL import PyAccess
@@ -252,7 +252,9 @@ class TestCffi(AccessTest):
 
 class TestEmbeddable(unittest.TestCase):
     @unittest.skipIf(not sys.platform.startswith('win32') or
-                     sys.version_info[:2] in ((3, 3), (3, 4)),
+                     sys.version_info[:2] in ((3, 3), (3, 4)) or
+                     on_appveyor(),   # failing on appveyor when run from
+                                      # subprocess, not from shell
                      "requires Python 2.7 or >=3.5 for Windows")
     def test_embeddable(self):
         import subprocess
--- a/Tests/test_image_access.py
+++ b/Tests/test_image_access.py
@@ -250,10 +250,14 @@ class TestCffi(AccessTest):
 
 
 class TestEmbeddable(unittest.TestCase):
-    @unittest.skipIf(not sys.platform.startswith('win32'), "requires Windows")
+    @unittest.skipIf(not sys.platform.startswith('win32') or
+                     sys.version_info[:2] in ((3, 3), (3, 4)),
+                     "requires Python 2.7 or >=3.5 for Windows")
     def test_embeddable(self):
         import subprocess
-        from distutils import ccompiler
+        import ctypes
+        import setuptools
+        from distutils import ccompiler, sysconfig
 
         with open('embed_pil.c', 'w') as fh:
             fh.write("""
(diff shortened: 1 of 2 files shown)


wiredfool:
@cgohlke I'm not getting this reloading working with my poking here?  Any suggestions?

cgohlke:
Works for me on Python 2.7.

I don't know about the AppVeyor failure, but for Python 3 and to catch crashes, the test need to be adjusted. This worked for me on Python 2.7 and >= 3.5:

```diff
diff --git a/Tests/test_image_access.py b/Tests/test_image_access.py
index 900f39e..3d9a3c0 100644
--- a/Tests/test_image_access.py
+++ b/Tests/test_image_access.py
@@ -7,7 +7,7 @@ except ImportError:
     pass
 
 from PIL import Image
-
+import sys
 
 class AccessTest(PillowTestCase):
     # initial value
@@ -246,5 +246,56 @@ class TestCffi(AccessTest):
                 self.assertEqual(px[i, 0], 0)
 
 
+class TestEmbeddable(unittest.TestCase):
+    @unittest.skipIf(not sys.platform.startswith('win32') or
+                     sys.version_info[:2] in ((3, 3), (3, 4)),
+                     "requires Python 2.7 or >=3.5 for Windows")
+    def test_embeddable(self):
+        import subprocess
+        import ctypes
+        from distutils import ccompiler
+
+        with open('embed_pil.c', 'w') as fh:
+            fh.write("""
+#include "Python.h"
+
+int main(int argc, char* argv[])
+{
+    char *home = "%s";
+#if PY_MAJOR_VERSION >= 3
+    wchar_t *whome = Py_DecodeLocale(home, NULL);
+    Py_SetPythonHome(whome);
+#else
+    Py_SetPythonHome(home);
+#endif
+
+    Py_InitializeEx(0);
+    Py_DECREF(PyImport_ImportModule("PIL.Image"));
+    Py_Finalize();
+
+    Py_InitializeEx(0);
+    Py_DECREF(PyImport_ImportModule("PIL.Image"));
+    Py_Finalize();
+
+#if PY_MAJOR_VERSION >= 3
+    PyMem_RawFree(whome);
+#endif
+
+    return 0;
+}
+        """ % sys.prefix.replace('\\', '\\\\'))
+
+        compiler = ccompiler.new_compiler()
+        objects = compiler.compile(['embed_pil.c'])
+        compiler.link_executable(objects, 'embed_pil')
+        
+        # do not display the Windows Error Reporting dialog
+        ctypes.windll.kernel32.SetErrorMode(0x0002)
+        
+        process = subprocess.Popen(['embed_pil.exe'])
+        process.communicate()
+        self.assertEqual(process.returncode, 0)
+
+
 if __name__ == '__main__':
     unittest.main()
```

cgohlke:
Re AppVeyor failures: 

Errorcode -1073741819 is [NTSTATUS value](https://msdn.microsoft.com/en-us/library/cc704588.aspx) 0xC0000135: STATUS_DLL_NOT_FOUND {Unable To Locate Component}.

I'm not sure where the Python DLL is located on AppVeyor. Try adding the Python directory to the PATH environment and pass it to subprocess.Popen, e.g.


```Python
...

import os
env = os.environ.copy()
env["PATH"] = sys.prefix + ';' + env["PATH"]

process = subprocess.Popen(['embed_pil.exe'], env=env)
...
```

cgohlke:
I can reproduce error 0xC0000135 when PIL can not be imported (not found in sys.path). Another possibility is that there is a problem with the msvc9 manifest in the executable or one of the PIL extensions.

cgohlke:
As mentioned before, this PR works for me. 

I think the difference is that AppVeyor tests `eggs` while I test in-place builds or installations from `wheels`. First, `embed_pil.exe` does not seem to know about the location of the egg (as far as I can reproduce on my system). Second, even when making the location of the egg known to `embed_pil.exe`, the second import of `PIL.Image` fails because another package used by the egg import system (`win32com`) fails to function (at least on my system). So far I have not been able to make `embed_pil.exe` work with Pillow eggs.

@tribetiki, could you confirm that your issue #2268 is solved using the binaries at <http://www.lfd.uci.edu/~gohlke/pythonlibs/#pillow>, which contain this fix?

If yes, how about disabling the test for this issue on AppVeyor or make AppVeyor test wheels instead of eggs?

tribetiki:
@cgohlke, I installed the binaries from http://www.lfd.uci.edu/~gohlke/pythonlibs/tuth5y6k/Pillow-4.0.0-cp27-cp27m-win32.whl. It seems the simple c++ embedding test from #2268 works a bit differently: loading the module twice works without exceptions, but I get an access violation exception from tk85.dll during cleanup (after the program exits). This happens even when I load PIL.Image only once. Not sure what is causing this.

cgohlke:
> It seems the simple c++ embedding test from #2268 works a bit differently: loading the module twice works without exceptions, but I get an access violation exception from tk85.dll during cleanup (after the program exits). This happens even when I load PIL.Image only once. Not sure what is causing this.

I can't reproduce the access violation exception. It looks like a separate issue. What happens if you change the embedding test from `import PIL.Image` to `import FixTk`? Also, which Python build and version are you using?

wiredfool:
I've rebased these on top of @cgohlke's recent prs regarding tk.

cgohlke:
To stop `winbuild/build.py` from installing eggs, try to use `python -m pip install .` instead of `python setup.py install`.

wiredfool:
Removed in-progress changes, squashed the reasonable changes to the tests, and disabled the test on appveyor.  It does work when the embed_pil.exe executable is run manually from the shell on Appveyor, it's just not working with the environment that we're getting in subprocess.

PR #2308, opened 2016-12-25
Remove executable bit from OleFileIO.py

jwilk:


jwilk:
I've added another commit by @radarhere to remove the shebang.

Historical background: the +x bit and the shebang used to (kinda) make sense:
```console
$ python PIL/OleFileIO.py 
olefile version 0.42b 2015-01-25 - Philippe Lagadec

Launched from the command line, this script parses OLE files and prints info.

Usage: olefile.py [-d] [-c] <file> [file2 ...]

Options:
-d : debug mode (displays a lot of debug information, for developers only)
-c : check all streams (for debugging purposes)

For more information, see http://www.decalage.info/olefile
```
But now that OleFileIO was unvendorized in 52b1f46b12433bcc4a4c02396b0875f800866787, executing this file no longer does anything.

PR #2310, opened 2016-12-27
allow frame durations of less than 10 milliseconds

gboeing:

Allow saving animated GIFs with frame durations of less than 10 milliseconds to make smoother animations. Currently, when pillow saves a GIF it divides the duration attribute by 10 then converts to int. This rounds any duration less than 10 to 0, which falls back to the default duration. The changes in this PR remove the division by 10 (and the concomitant multiplication by 10) so durations less than 10 can be saved. Demonstrated [here](https://github.com/gboeing/lorenz-system/blob/master/lorenz-system-attractor-animate.ipynb).

### DIFF
--- a/PIL/GifImagePlugin.py
+++ b/PIL/GifImagePlugin.py
@@ -187,7 +187,7 @@ class GifImageFile(ImageFile.ImageFile):
                     flags = i8(block[0])
                     if flags & 1:
                         self.info["transparency"] = i8(block[3])
-                    self.info["duration"] = i16(block[1:3]) * 10
+                    self.info["duration"] = i16(block[1:3])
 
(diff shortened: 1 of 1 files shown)


wiredfool:
I don't think that's doing what you think it's doing:

from https://en.wikipedia.org/wiki/GIF#Animated_GIF
```
The animation delay for each frame is specified in the GCE in hundredths of a second. Some economy of data is possible where a frame need only rewrite a portion of the pixels of the display, because the Image Descriptor can define a smaller rectangle to be rescanned instead of the whole image. Displays that do not support animated GIFs show only the first frame.
```

We do durations in milliseconds, probably for compatibility across image formats. Gif only does hundredths, so you've effectively changed the parameter to hundredths.

PR #1384, opened 2015-08-21
GIF parameters

radarhere:
Train of thought started by #1355. Combines PR #1378 and #1379

GifImagePlugin currently uses only the params passed directly into Image's `save` method. It doesn't retrieve any params from the image itself. It effectively does not roundtrip properties. The reason that this has gone unnoticed is that `duration` and `loop` have only started to be saved recently, there is no test to roundtrip `transparency` that I can see, `version` has been fixed when saving and `background` is the one exception to this rule.

So this PR merges the params passed in with those of the image itself, with the parameters passed in having priority.

***

The second part of this is the GIF version. GifImagePlugin has the output version number fixed as 87a, but may use features that are only valid for 89a.

http://www.w3.org/Graphics/GIF/spec-gif89a.txt
"The specification given here defines version 89a, which is an extension of version 87a."

The features that require 89a are -

Graphic Control Extension
Comment Extension (unused by Pillow)
Plain Text Extension (unused by Pillow)
Application Extension
"An encoder should use the earliest possible version number that includes all the blocks used in the Data Stream."

I think that this sounded reasonable in 1990, but today, both versions of GIF would be supported on a given platform. My suggestion is - if 89a features are used, then 89a. Otherwise, if the input has a version, use that. Otherwise, 87a.

This PR also makes that change.

PR #1320, opened 2015-06-30
Merged gifmaker into GifImagePlugin

radarhere:
Resolves #1271

This is the first step in saving multiframe images (#733), by merging the functionality of the gifmaker script into GifImagePlugin.

Happy to accept other ideas about how to do `save_all`. In this pull request, I'm imagining `im.save('test.gif', save_all=True)`

PR #2317, opened 2016-12-29
Combine duration of identical GIF frames when saving

### LINKED ISSUE #2314: saving animated gif drops frames
### What did you do?

Saved a set of 100 png files as an animated gif then opened the gif

I expected the gif to have 100 frames, same as the number of png files

### What actually happened?

The gif has only 70 frames

pillow 3.4.2, python 3.5.2

### Demo

There are 100 png files in the folder `save_folder`, named 000.png through 099.png. The first 16 images are identical to one another, and the final 16 images are identical to one another. The other 68 images are all different from each other. In other words, the animation would show the first image for 16 frames, then run through the middle 68 frames before coming to the concluding image which it shows for 16 frames. But when I save the gif, it is missing 15 of the first 16 images and 15 of the final 16 images. Thus there is no "pause" at the beginning or end of the animation.

```python
images = [Image.open(image) for image in glob.glob('{}/*.png'.format(save_folder))]
len(images)
```
returns: 100

```python
gif = images[0]
gif.save(fp=gif_filepath, format='gif', save_all=True, append_images=images[1:])
Image.open(gif_filepath).n_frames
```
returns: 70

radarhere:
Resolves #2314

If identical frames exist in a GIF when saving, combine their durations.

Also removes a copy operation in this refactoring. This operation was last modified in https://github.com/python-pillow/Pillow/pull/1384/commits/e88e90b8ede189221199c429340d3242fc9dc318 and before that in https://github.com/python-pillow/Pillow/pull/1320/commits/2c4fe7281f4c58ab9f5f32addd835678c1363e35. I can't see a reason for it.

### DIFF
--- a/PIL/GifImagePlugin.py
+++ b/PIL/GifImagePlugin.py
@@ -378,7 +378,7 @@ def _save(im, fp, filename, save_all=False):
                         first_frame = None
 
                     # delta frame
-                    delta = ImageChops.subtract_modulo(im_frame, previous.copy())
+                    delta = ImageChops.subtract_modulo(im_frame, previous)
                     bbox = delta.getbbox()
 
                     if bbox:
--- a/PIL/GifImagePlugin.py
+++ b/PIL/GifImagePlugin.py
@@ -302,7 +302,7 @@ except ImportError:
 RAWMODE = {
     "1": "L",
     "L": "L",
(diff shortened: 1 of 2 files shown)


wiredfool:
Closing in favor of #2374

PR #2319, opened 2016-12-29
Updated pngquant to 2.8.2

radarhere:


depends/install_imagequant.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 # install libimagequant
 
-archive=pngquant-2.6.0
+archive=pngquant-2.8.2
```
wiredfool: This didn't actually work due to a change in the project/makefile:
https://travis-ci.org/python-pillow/Pillow/jobs/187587442#L2240

radarhere: Thanks for noticing. I've created PR #2354.

PR #2320, opened 2016-12-29
Updated tested versions of libjpeg

radarhere:
libjpeg 9b has been a part of Pillow since PR #1673 and https://github.com/python-pillow/pillow-wheels/pull/29


aclark4life:
Thanks

PR #2329, opened 2017-01-01
Refactor out postprocessing hack to load_end in PcdImageFile

wiredfool:

 * Move single purpose hack in imagefile.load to PcdImagePlugin.load_end()


PIL/PcdImagePlugin.py
```diff
@@ -42,6 +42,7 @@ def _open(self):
             raise SyntaxError("not a PCD file")
 
         orientation = i8(s[1538]) & 3
+        self.tile_post_rotate = None
         if orientation == 1:
             self.tile_post_rotate = 90  # hack
```
radarhere: Would you also like to remove the '# hack' comment here?

PR #2330, opened 2017-01-01
In progress update to #2032

wiredfool:
Fixes #2032.

 * Close any single frame files

### DIFF
--- a/Tests/test_file_tiff.py
+++ b/Tests/test_file_tiff.py
@@ -490,6 +490,7 @@ class TestFileTiffW32(PillowTestCase):
         tmpfile = self.tempfile("temp.tif")
         import os
 
+        # this is an mmaped file. 
         with Image.open("Tests/images/uint16_1_4660.tif") as im:
             im.save(tmpfile)
 
@@ -499,7 +500,12 @@ class TestFileTiffW32(PillowTestCase):
         self.assertRaises(Exception, lambda: os.remove(tmpfile))
         im.load()
         self.assertTrue(fp.closed)
-        # this should not fail, as load should have closed the file. 
+
+        # this closes the mmap
+        im.close()
+        
+        # this should not fail, as load should have closed the file pointer,
+        # and close should have closed the mmap
         os.remove(tmpfile)
 
 if __name__ == '__main__':
--- a/Tests/test_file_tiff.py
+++ b/Tests/test_file_tiff.py
@@ -1,6 +1,7 @@
 import logging
 from io import BytesIO
 import struct
+import sys
 
 from helper import unittest, PillowTestCase, hopper, py3
 
@@ -468,6 +469,21 @@ class TestFileTiff(PillowTestCase):
 
         self.assertEqual(b'Dummy value', reloaded.info['icc_profile'])
 
+@unittest.skipUnless(sys.platform.startswith('win32'), "Windows only")
+class TestFileTiffW32(PillowTestCase):
+    def test_fd_leak(self):
+        tmpfile = self.tempfile("temp.tif")
+        import os
+
+        with Image.open("Tests/images/uint16_1_4660.tif") as im:
+            im.save(tmpfile)
+
+        im = Image.open(tmpfile)
+        im.load()
+        self.assertRaises(Exception, lambda: os.remove(tmpfile))
+
+        im.close()
+        os.remove(tmpfile)
 
 if __name__ == '__main__':
     unittest.main()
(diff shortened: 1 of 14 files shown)


homm:
My recent thoughts about the problem:

* `exclusive_fp` is good name for variable in `PIL.Image.open` function. This refers to the fact that the file pointer was created inside the function and could be easily freed in case of error.
* `_exclusive_fp` if not such good name for the image property. The pointer could be exclusive while we still can't close it after the initial loading.
* If the pointer is really **not** exclusive (file is opened not in the `PIL.Image.open` function), we shouldn't do anything at all.
* If the pointer **is exclusive** we must close the pointer at some point. In worst case in `__del__` method.
* There is a huge issue with `__del__` method in Python 2.7: objects with`__del__` could not be collected in case of circular references. As a result, the hypothetical memory leak can turn into real memory leak. On the other hand, there is no memory leak related to not properly closed file pointers in Python 2.7. So we can relatively safe implement `__del__` method only for Python >= 3.4 (see [gc.garbarge](https://docs.python.org/3.6/library/gc.html#gc.garbage)).
* For most of the image plugins we can close file pointers much earlier than in `__del__` method. To distinguish this cases, we need separate flag in the `PIL.Image.Image` class. In current implementation (this PR), this flag is combined with `_exclusive_fp`. I believe we should separate them.

In conclusion:

* In `PIL.Image.open` always set `_exclusive_fp` without any conditions:
    ```python
    im._exclusive_fp = exclusive_fp
    ```
* Add flag to the Image class:
    ```python
    _close_exclusive_after_loading = True
    ```
* Set it in all places [like this](https://github.com/python-pillow/Pillow/pull/2330/files#diff-22ef6cdf231992915ec315132b2d4aecR67):
    ```python
    self._close_exclusive_after_loading = False
    ```
* Add to `PIL.ImageFile.ImageFile.load` method [another check](https://github.com/python-pillow/Pillow/pull/2330/files#diff-6b6c4410a4159082c840f529897c991aR250):
    ```python
    if self._exclusive_fp and self._close_exclusive_after_loading:
        self.fp.close()
    ```
* Add to `PIL.Image.Image` for Python >= 3.4 `__del__` method.
    ```python
     def __del__(self):
        if self.fp and self._exclusive_fp:
            self.fp.close()
        self.fp = None
    ```

wiredfool:
So, to sum up:

* If we've opened the file, we have to close it. 
  * If it's multiframe, leave it open til `__del__`
  * If not, close it when we've loaded. 
  * If we don't load, then close it in the `__del__` method.
* If we're passed an open file, we leave it open.

homm:
Right

wiredfool:
Ok, I think that's the right way to handle the file closing.

okor:
I just wanted to thank both of you for working on this. I know it's a lot of hard work. And I know the Thumbor crowd will deeply appreciate this fix. Ping me if I can help : )

wiredfool:
Ok. I now understand the test failure on windows. 

For images where we have a memory mapped file, even if we close the file pointer, we've still got a map open to the underlying file until the image is closed. We can't close that map until we're not going to access the image pixels again. 

"""
class TestFileTiffW32(PillowTestCase):
    def test_fd_leak(self):
        tmpfile = self.tempfile("temp.tif")
        import os

        with Image.open("Tests/images/uint16_1_4660.tif") as im:
            im.save(tmpfile)

        im = Image.open(tmpfile)
        fp = im.fp
        self.assertFalse(fp.closed)
        self.assertRaises(Exception, lambda: os.remove(tmpfile))
        im.load()
        self.assertTrue(fp.closed)
   
       # this will allow the remove to work
        im2 = im.copy()
        im = None

        # this should not fail, as load should have closed the file. 
        os.remove(tmpfile)
"""

So, I believe that this test is actually hitting a different issue (can't remove a mapped file on windows), and if we opened a jpeg file (which can't be mmaped) it would succeed and the file closing issue would apparently pass.

wiredfool:
Ok, I think we're finally good here.

wiredfool:
@homm, this look good to you?

Tests/test_file_tiff.py
```diff
@@ -468,6 +469,44 @@ def test_saving_icc_profile(self):
 
         self.assertEqual(b'Dummy value', reloaded.info['icc_profile'])
 
+    def test_close_on_load(self):
+        # same as test_fd_leak, but runs on unixlike os
+        tmpfile = self.tempfile("temp.tif")
```
homm: It is not clear: is it a path or file-like object?

homm:
I'd like to see tests for both cases: when pf is exclusive and not

Tests/test_file_tiff.py
wiredfool: Tempfile returns a path.

```
    def tempfile(self, template):
        assert template[:5] in ("temp.", "temp_")
        fd, path = tempfile.mkstemp(template[4:], template[:4])
        os.close(fd)

        self.addCleanup(self.delete_tempfile, path)
        return path
```

PR #2340, opened 2017-01-06
Removed unnecessary import

radarhere:
Resolves a TODO in _binary to replace `from struct import ...` with `import struct`

Cherry-picks an isolated commit from #2339 that achieves the same goal in JpegImagePlugin.

### DIFF
--- a/PIL/JpegImagePlugin.py
+++ b/PIL/JpegImagePlugin.py
@@ -38,7 +38,6 @@ import array
 import struct
 import io
 import warnings
-from struct import unpack_from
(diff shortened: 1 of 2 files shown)


jleclanche:
lgtm

wiredfool:
That todo was satisfied by the changes in https://github.com/radarhere/Pillow/commit/aeb1e61aa753beacf442763f0df23af8fb9c781c , which replaced a whole bunch python stuff with `struct.un/pack`.

radarhere:
Thanks for catching that. Okay, I have updated this PR to remove the TODO instead.

PR #2347, opened 2017-01-08
Moved iCCP chunk before PLTE chunk when saving as PNG

### LINKED ISSUE #2201: PNG: iCCP must precede PLTE
```Python
from PIL import Image
img = Image.new(size = (1, 1), mode = 'P')
img.save('file.png', icc_profile = 'ICC') # bogus
```
```Bash
$ pngcheck -v file.png 
File: file.png (883 bytes)
  chunk IHDR at offset 0x0000c, length 13
    1 x 1 image, 8-bit palette, non-interlaced
  chunk PLTE at offset 0x00025, length 768: 256 palette entries
  chunk iCCP at offset 0x00331, length 24:  must precede PLTE
ERRORS DETECTED in file.png
```
[Chunk ordering as per the PNG spec.](https://www.w3.org/TR/PNG/#5ChunkOrdering)

radarhere:
Fixes #2201

### DIFF
--- a/PIL/PngImagePlugin.py
+++ b/PIL/PngImagePlugin.py
@@ -715,6 +715,8 @@ def _save(im, fp, filename, chunk=putchunk, check=0):
           b'\0',                                # 11: filter category
           b'\0')                                # 12: interlace flag
 
+    chunks = [b"cHRM", b"gAMA", b"sBIT", b"sRGB", b"tIME"]
+
     icc = im.encoderinfo.get("icc_profile", im.info.get("icc_profile"))
     if icc:
         # ICC profile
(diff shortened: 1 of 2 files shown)


homm:
Tests?

radarhere:
Okay, I've added a test.

wiredfool:
So with this patch are we handling the ordering correctly for all of the chunks now?

radarhere:
For all of the chunks that are written specifically, yes, the iCCP change covers it.

However, there is writing of arbitrary chunks at https://github.com/python-pillow/Pillow/blob/master/PIL/PngImagePlugin.py#L762. So I have added another commit to this PR to restrict that process to writing allowed chunk names, in the correct order, that are not specifically written elsewhere in the file,.

PR #2349, opened 2017-01-09
Replaced absolute PIL imports with relative imports

### LINKED ISSUE #2342: Cleanup: Replace absolute PIL imports with relative imports
There's a ton of stuff like this in all the plugins:

`from PIL import Image, ImageFile, ImagePalette, _binary`

The ones shipped with PIL should be using relative imports instead.

This would also be a good occasion to replace all this kind of nonsense:

```py
from PIL import Image, ImageFile, ImagePalette, _binary
i8 = _binary.i8
i16 = _binary.i16le
i32 = _binary.i32le
o8 = _binary.o8
o16 = _binary.o16le
o32 = _binary.o32le
```

With this:

```py
from ._binary import i8, i16, i32, o8, o16, o32
```

radarhere:
Resolves #2342

The following files cannot use relative imports because they may be run directly, as shown by their `if __name__ == "__main__":` line -
- IcnsImagePlugin
- ImageCms
- ImageShow
- SpiderImagePlugin

### DIFF
--- a/PIL/BdfFontFile.py
+++ b/PIL/BdfFontFile.py
@@ -19,8 +19,7 @@
 
 from __future__ import print_function
 
-from PIL import Image
-from PIL import FontFile
+from . import Image, FontFile
 
 
 # --------------------------------------------------------------------
--- a/PIL/BmpImagePlugin.py
+++ b/PIL/BmpImagePlugin.py
@@ -24,17 +24,12 @@
 #
 
 
-from PIL import Image, ImageFile, ImagePalette, _binary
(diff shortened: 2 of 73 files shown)


radarhere:
This makes me think about the usefulness of a 'It's fine, it's fine, the coverage only dropped because there's fewer lines' label.

wiredfool:
Coveralls is only approximately useful, as there are often 'changes' in coverage to unrelated files. 

Perhaps a better label would be 'It's fine, coveralls is drunk. Again.'

PR #2355, opened 2017-01-15
Imagecms preserve aux channels

### LINKED ISSUE #1662: Converting Image Profile to sRGB
Hi,

when i am converting a Image with an embedded "Adobe RGB (1998)" profile to sRGB, i only receive an fully transparent image. On the other Hand, when i am converting an image with an embedded "Linear Color Space" profile, everything works as expected. I don't know why some profiles convert to only transparent pictures. It would be great, if we could work out this issue.

Source Code:

``` python
import PIL.Image
import PIL.ImageCms
import tempfile

img = PIL.Image.open('input.png')
icc = tempfile.mkstemp(suffix='.icc')[1]
with open(icc, 'w') as f:
    f.write(img.info.get('icc_profile'))
srgb = PIL.ImageCms.createProfile('sRGB')
img = PIL.ImageCms.profileToProfile(img, icc, srgb)
img.save('output.png')
```

Tested Python Versions:
- 2.7.10

Tested Pillow Versions:
- 3.0.0
- 3.1.0

Tested LittleCms Versions:
- 2.7

Source and Sample Image:
[source.zip](https://github.com/python-pillow/Pillow/files/92095/source.zip)

gunjambi:
Fixes #1662.

 * After a not-in-place CMS operation, propagate auxiliary (i.e. non-color) channel data to the result image. In practice, this preserves image alpha channel during a color space conversion.

### DIFF
--- a/_imagingcms.c
+++ b/_imagingcms.c
@@ -291,6 +291,97 @@ findLCMStype(char* PILmode)
     }
 }
 
+#define Cms_Min(a, b) ((a) < (b) ? (a) : (b))
+
+static int
+pyCMSgetAuxChannelChannel (cmsUInt32Number format, int auxChannelNdx)
+{
+    int numColors = T_CHANNELS(format);
+    int numExtras = T_EXTRA(format);
+
+    if (T_SWAPFIRST(format) && T_DOSWAP(format)) {
+        // reverse order, before anything but last extra is shifted last
+        if (auxChannelNdx == numExtras - 1)
+            return numColors + numExtras - 1;
+        else
+            return numExtras - 2 - auxChannelNdx;
+    }
+    else if (T_SWAPFIRST(format)) {
+        // in order, after color channels, but last extra is shifted to first
+        if (auxChannelNdx == numExtras - 1)
+            return 0;
+        else
+            return numColors + 1 + auxChannelNdx;
+    }
+    else if (T_DOSWAP(format)) {
+        // reverse order, before anything
+        return numExtras - 1 - auxChannelNdx;
+    }
+    else {
+        // in order, after color channels
+        return numColors + auxChannelNdx;
+    }
+}
+
+static void
+pyCMScopyAux (cmsHTRANSFORM hTransform, Imaging imDst, const Imaging imSrc)
+{
+    cmsUInt32Number dstLCMSFormat;
+    cmsUInt32Number srcLCMSFormat;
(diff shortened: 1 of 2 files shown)


hugovk:
Please can you check the failing Windows build for Python 3?

https://ci.appveyor.com/project/Python-pillow/pillow/build/4.1.pre.2031

Please also use `lower_case_with_underscores` for function names and variables in Python code.

_imagingcms.c
```diff
@@ -291,6 +291,97 @@ findLCMStype(char* PILmode)
...
+    dstChunkSize = (T_CHANNELS(dstLCMSFormat) + T_EXTRA(dstLCMSFormat)) * channelSize;
+
+    for (e = 0; e < numExtras; ++e) {
+        int y;
+        int dstChannel = pyCMSgetAuxChannelChannel(dstLCMSFormat, e);
+        int srcChannel = pyCMSgetAuxChannelChannel(srcLCMSFormat, e);
+
+        for (y = 0; y < ySize; y++) {
+            int x;
+            void* pDstExtras = imDst->image[y] + dstChannel * channelSize;
```
wiredfool: I think these should be char * or uint8 *

gunjambi:
Ack, I'll look into the issues probably next weekend or so.

However, I noticed there has been some relevant development recently on littlecms side too. The latest release of last July added support for alpha channel preservation via a new cmsFLAGS_COPY_ALPHA flag. Too bad this wasn't available when I debugged this, would've saved me some effort.

Now the question is, should I pursue with this change? Using littlecms would probably be in general the better solution as it would keep this aux channel cruft from polluting pillow codebase. The downside is that it requires upgraded littlecms and since the feature was added as an additional flag, any user building with the older lcms would still have silently broken alpha channel.

Alternatively, we could run with this code for now to wait for latest littlecms version to spread, and then after a year or two tear this code out and use cmsFLAGS_COPY_ALPHA, but it's doubtful if anyone has any interest to do this then.

So, opinions? Just scrap this and postpone until lcms dependency is upgraded to 2.8? We can probably salvage the tests.

wiredfool:
If the feature is useful, I think it's useful to have this in as a polyfill. Lcms 2.8 isn't in any of the Ubuntu releases yet, though it looks like it's going to make Debian Stretch, and it's been in Fedora for a while. 

I'd certainly defer to the library implementation if it's available.

Issue #1902, opened 2016-05-10
ImageTk int overflow error


vlachoudis:
I am running on a 32bit Fedora-23 virtual machine via virtualbox, version 3.0.0-4
When I try to convert an image to tkimage I get an overflow error
since the tk.interpaddr() returns a 64bit address which cannot be converted to int.

The same code in 64bit Fedora-23 save version works ok.

```
File "/usr/local/flair/lib/ZoomImage.py", line 162, in zoomImage
    self.tkimage = ImageTk.PhotoImage(self.photo)
  File "/usr/lib/python2.7/site-packages/PIL/ImageTk.py", line 115, in __init__
    self.paste(image)
  File "/usr/lib/python2.7/site-packages/PIL/ImageTk.py", line 182, in paste
    _imagingtk.tkinit(tk.interpaddr(), 1)
OverflowError: Python int too large to convert to C long
```

```
>>> from Tkinter import *
>>> tk=Tk()
>>> tk.interpaddr()
2161299816L
```

wiredfool:
It's not actually a 64 bit address, but it is in the upper 2gb of a 32 bit address. Which is too big to fit into a Py_ssize_t on 32 bit, which is limited to 2^31. 

The value is actually a PyInt_FromLong, from _tkinter.c:
```
static PyObject *
Tkapp_InterpAddr(PyObject *self, PyObject *args)
{

    if (!PyArg_ParseTuple(args, ":interpaddr"))
        return NULL;

    return PyInt_FromLong((long)Tkapp_Interp(self));
}
```

Which is interpreted as a an int->Py_ssize_t in _imagingtk.c 
```
static PyObject*
_tkinit(PyObject* self, PyObject* args)
{
    Tcl_Interp* interp;

    Py_ssize_t arg;
    int is_interp;
    if (!PyArg_ParseTuple(args, "ni", &arg, &is_interp))
        return NULL;

    if (is_interp)
        interp = (Tcl_Interp*) arg;
    else {
        TkappObject* app;
	/* Do it the hard way.  This will break if the TkappObject
	   layout changes */
        app = (TkappObject*) arg;
        interp = app->interp;
    }
```

Given the evidence here, I don't think that it's actually possible to reliably (in all cases, with all corners of 32/64 signed/unsigned silliness) pass a pointer through as an interpreted int. It can be passed as a PyInt, and then it needs to be interpreted as something like:


```
    void *arg;
    PyObject *int_obj;
[...]

PR #2359, opened 2017-01-17
Fix for issue #1902 Applied patch: matplotlib/matplotlib/commit/a9155…

wiredfool:
Applied patch: https://github.com/matplotlib/matplotlib/commit/a91559b82cf23ee407cd57580653015fc7dc35f0 to fix issue #1902

 * Use PyLong_AsVoidPtr instead of converting a pyint to a py_ssizet, which misses the top half of the address space on 32 bit machines.

### DIFF
--- a/PIL/ImageTk.py
+++ b/PIL/ImageTk.py
@@ -277,6 +277,8 @@ class BitmapImage(object):
 
 
 def getimage(photo):
+    """ This function is unimplemented """
+    
     """Copies the contents of a PhotoImage to a PIL image memory."""
     photo.tk.call("PyImagingPhotoGet", photo)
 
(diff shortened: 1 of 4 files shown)


wiredfool:
Tests:  I can't find a way to trigger this in a 32 bit docker image on a 64 bit machine -- it requires a 32 bit vm at the very least, and it's only getting triggered on Debian, not Ubuntu in my testing. 

Waiting on #2368

wiredfool:
Pypy 5.6 is segfaulting in `_imagingtk.c:_tkinit`:

```
    /* This will bomb if interp is invalid... */
    TkImaging_Init(interp);
```

It's consistent back to 2.5.0, which is well before the Tk reorg.

PR #2363, opened 2017-01-18
Ensure that the `err_code` variable is initialized before using.

alexkiro:
### What did you do?

Called ``image = image.convert("RGB")``

### What actually happened?

Unfortunately I cannot retrieve the actual image that causes this to occur, but there is a scenario where the ``err_code`` variable is not initialized.

```
  File "PIL/Image.py", line 841, in convert
  File "PIL/ImageFile.py", line 238, in load
UnboundLocalError: local variable 'err_code' referenced before assignment
```

 - Pillow 3.3.3
 - Python 2.7.12

### DIFF
--- a/PIL/ImageFile.py
+++ b/PIL/ImageFile.py
@@ -178,7 +178,7 @@ class ImageFile(Image.Image):
                     self.map = None
 
         self.load_prepare()
-
+        err_code = 0
         if not self.map:
             # sort tiles in file order
             self.tile.sort(key=_tilesort)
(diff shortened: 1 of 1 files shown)


wiredfool:
I'd really like to see the image where that happens. We've had another report of it, but have still never seen an image that triggers the error. 

From looking at the code, it's got to be a case where a value error is thrown and silenced from each tile that's been defined. 

I doubt that passing a no-error code into this is quite the right approach.

alexkiro:
> I'd really like to see the image where that happens. We've had another report of it, but have still never seen an image that triggers the error.

I'll see if I can catch a live one then :+1: and send it to you.

hugovk:
See https://github.com/python-pillow/Pillow/issues/2383#issuecomment-275932234

PR #2365, opened 2017-01-19
Default to inch-interpretation for missing ResolutionUnit in TiffImagePlugin

lambdafu:

 * Default to inch-interpretation for Tiff images missing ResolutionUnit tag, thus creating an im.info['dpi'] attribute.
 * Keep old im.info['resolution'] attribute in such cases for compatibility.
 * Test case for this.


lambdafu:
Here is a document saying that inch is the right default: http://www.awaresystems.be/imaging/tiff/tifftags/resolutionunit.html

PR #2288, opened 2016-12-13
Libtiff metadata writing fixes

wiredfool:
Fixes #1765 (among others, likely)

Libtiff's metadata support is tricky and dangerous. This patch is 1/3 of the work to clean up our support for it. 

1) Understand and support writing basic metadata through libtiff. This includes casting numbers to the correct format and passing in arrays correctly for known items. (e.g., this fixes #1765, writing ints to dpi, and writing the colormap for `P` mode images)

2) Consider supporting custom metadata tags safely. This is going to require reflection from libtiff to ensure that we're setting items correctly.

3) Coordinate this work with the python metadata implementation in two places: When reading images where the type conflicts with the spec, and when writing (in python) metadata of incorrect types. 

While this is not finished, I believe that this is a material improvement in the support for tiff metadata, and can be merged (after I'm happy and rebased out some things)  for 3.5.

PR #2368, opened 2017-01-20
Docker images for additional travis runs

### LINKED ISSUE #2152: Add 32 bit linux to the Travis test matrix
We should add at least a canary 32 bit build to the travis test matrix to prevent errors from slipping through.

wiredfool:
Fixes #2152.

This is a big, in progress refactor and addition. 

* I've added 4 new test items in the matrix, Alpine, Precise (12.04), Xenial (16.04), and Trusty X86 (14.04). They're all testing py2.7 for now. I particularly want the x86 part of the matrix, because we've tripped over that recently. (and there's a pr requiring x86 testing: #2359) Precise and Xenial wind up testing different library versions, also useful and necessary for the tiff metadata pr #2288.
* I've added a repo of docker files, so they can be pulled from the dockerhub.  I'm not convinced that it's tons faster to do that rather than just install from bare images.
* ~I've converted a few random test docstrings so that the -v listing shows all the test names that have been run.~ 
* I've re-orged the .travis.yml file, refactoring some of it into scripts so that it's easier to gate on docker/non-docker. This unfortunately removes the travis_retry verb, since it's implemented in travis, not shell. 
* ~I'm going to at least rebase the many commits this prior to merge.~

And the big one -- Tests are failing:
* ~There are actual failing tests on the trusty one.~
* ~Precise is hanging, I think on Jpeg2k. It doesn't do it OMM, where I'm building the docker image. (go figure).~

### DIFF
--- a/.travis/after_success.sh
+++ b/.travis/after_success.sh
@@ -1,0 +1,44 @@
+#!/bin/bash
+
+# gather the coverage data
+sudo apt-get -qq install lcov
+lcov --capture --directory . -b . --output-file coverage.info
+#  filter to remove system headers
+lcov --remove coverage.info '/usr/*' -o coverage.filtered.info
+#  convert to json
+gem install coveralls-lcov
+coveralls-lcov -v -n coverage.filtered.info > coverage.c.json
+
+coverage report
+pip install coveralls-merge
+coveralls-merge coverage.c.json
+
+if [ "$TRAVIS_PYTHON_VERSION" == "2.7" ]; then
+	pip install pep8 pyflakes
+	pep8 --statistics --count PIL/*.py
+	pep8 --statistics --count Tests/*.py
+	pyflakes *.py       | tee >(wc -l)
+	pyflakes PIL/*.py   | tee >(wc -l)
+	pyflakes Tests/*.py | tee >(wc -l)
+fi
+
+# Coverage and quality reports on just the latest diff.
+# (Installation is very slow on Py3, so just do it for Py2.)
+if [ ${TRAVIS_PYTHON_VERSION:0:1} == "2" ]; then depends/diffcover-install.sh; fi
+if [ ${TRAVIS_PYTHON_VERSION:0:1} == "2" ]; then depends/diffcover-run.sh; fi
+
+# after_all
+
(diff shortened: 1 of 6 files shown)


wiredfool:
So, this is working now.  Though, I'd feel better if the tk tests were failing.  Though, it looks like this would have caught the 3.4.0 issue.

.travis/after_success.sh
```diff
@@ -0,0 +1,44 @@
...
+coveralls-merge coverage.c.json
+
+if [ "$TRAVIS_PYTHON_VERSION" == "2.7" ]; then
+	pip install pep8 pyflakes
+	pep8 --statistics --count PIL/*.py
+	pep8 --statistics --count Tests/*.py
+	pyflakes *.py       | tee >(wc -l)
+	pyflakes PIL/*.py   | tee >(wc -l)
+	pyflakes Tests/*.py | tee >(wc -l)
+fi
```
wiredfool: So, I could have sworn that this was in the .travis.yml file that I was pulling from, but I can't find it now, and I've rebased enough that it's lost.

wiredfool:
> Would it be worth testing and comparing both approaches?

That was written before I pulled the compiled dependencies into the images. That saves 15-30 seconds. It's disappointing that on a local cached run, it's 10-15 seconds, but the runs on travis are measured in minutes. But there's downloading and post script setup stuff.  Apart from that I think having the testing images on dockerhub is useful for others to pull and test against when debugging failures. e.g: `cd ubuntu-trusty-x86 && make shell` puts you in the docker image with all the dependencies. 

>  it's been possible to identify and improve some `make -jX` commands

Yeah, I noticed that. `make install` shouldn't really be threaded, though it probably doesn't matter that much as it doesn't start a ton of subtasks. 

The `after_success.sh` script can be easily put back in -- except that we're not running coverage on the docker ones, so it's not actually doing anything there. `Install.sh` and `script.sh`, those are either/or and will be a big mess if we test on docker for each item.

.travis/after_success.sh
wiredfool: Ok, I've just taken these off the dockers. (Note that `TRAVIS_PYTHON_VERSION` in the docker runs is the python version in travis, not in the docker image, so it's pretty much irrelevant at this point, and could fundamentally mismatch)

wiredfool:
Also worth noting -- I'm planning on adding a virtualenv parameter to the docker args so that we can test against alternate pythons on the same image. (e.g. system py2 and py3)

PR #2371, opened 2017-01-22
Consider all frames when optimizing a GIF image

radarhere:
https://github.com/python-pillow/Pillow/issues/2314#issuecomment-274265583 raises a bug.

From what I can see, the problem is that not all of the frames in the image are being considered when optimizing. This PR fixes that.

### DIFF
--- a/PIL/GifImagePlugin.py
+++ b/PIL/GifImagePlugin.py
@@ -353,40 +353,42 @@ def _save(im, fp, filename, save_all=False):
         else:
             duration = None
         frame_count = 0
+        im_frames = []
         for imSequence in [im]+append_images:
             for im_frame in ImageSequence.Iterator(imSequence):
-                encoderinfo = im.encoderinfo.copy()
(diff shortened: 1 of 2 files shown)


wiredfool:
I think what's going on here for subsequent frames is: 
1) The image difference doesn't take palette into account, it's essentially the `P` mode bytes interpreted as `L`. 
2) I don't think we're writing a local color table if the palette is different than the palette from the first image. 

I don't think stuffing all of the palettes into the first palette is going to be the right solution.

wiredfool:
The test, at least, should be cherry picked into whatever solution comes up.

radarhere:
I deleted this quickly after your initial comment, because I agreed that local color tables were a better idea.

However, in my further investigation of this, I came to a different conclusion about what the bug was - see PR #2375.

wiredfool:
I think that 1 is still an issue, but I'm not sure on 2.

PR #2283, opened 2016-12-11
Minor code cleanup in GifImagePlugin

radarhere:
Inspired by @homm's comments in #2196

Issue #2373, opened 2017-01-24
Release Pillow 4.1.0 on April 1, 2017


wiredfool:
This is the release thread for 4.1.0, scheduled for 2017-04-01. 

I'd like to have a couple of goals for this release. 

1) I'd like to cover all of the image formats that we supposedly support with tests for correctness of loaded (/saved) images, or have a loud warning that this is untested code, either in code or docs. There have been three cases in the last few months where I've opened the hood on a file format or a feature and found it nearly untested and/or broken. (Sun RLE was broken, SGI raster files only tested that we rejected unsupported compression, and ImageTk has never had tests where we loaded the tk interpreter.)  I'd like to not need the "Hasn't worked in 20 years" tag anymore. 

 - [x] BufrStubImagePlugin: https://github.com/python-pillow/Pillow/pull/2441
 - [x] ContainerIO: https://github.com/python-pillow/Pillow/pull/2443
 - [x] FitsStubImagePlugin: https://github.com/python-pillow/Pillow/pull/2430
 - [ ] GdImagePlugin: https://github.com/python-pillow/Pillow/pull/1817
 - [x] GribStubImagePlugin: https://github.com/python-pillow/Pillow/pull/2440
 - [x] Hdf5StubImagePlugin: https://github.com/python-pillow/Pillow/pull/2431
 - [x] MspImagePlugin: #1938 (#2384)
 - [x] PixarImagePlugin: #2428 
 - [x] PsdImagePlugin: https://github.com/python-pillow/Pillow/pull/2427
 - [x] XVThumbImagePlugin: https://github.com/python-pillow/Pillow/pull/2442
 - [ ] FPXImagePlugin
 - [ ] BitDecode.c
 - [ ] McIdasImagePlugin.py
 - [ ] MicImagePlugin.py
 - [ ] IptcImagePlugin.py (load is never tested)

2) Finish several longstanding issues -- File Closing, Palette, and Tiff Metadata handling come to mind. 

 - [x] File closing #2330
 - [ ] Palette
 - [ ] Tiff metadata

3) Refactor the mess that is GifImagePlugin. 

 - [x] GifImagePlugin

wiredfool:
For point 1, just from the review of #2425:
 - BufrStubImagePlugin
 - ContainerIO
 - FitsStubImagePlugin
 - GribStubImagePlugin
 - Hdf5StubImagePlugin
 - PixarImagePlugin
 - PsdImagePlugin
[...]

PR #2374, opened 2017-01-26
Simplified GifImagePlugin code

radarhere:
Includes #2283 and #2317

The following breaking changes are also included -
- Removes second return value from `getheader` and `_get_palette_bytes`, since nothing in the code requires it
- Removes `_get_used_palette_colors` by merging it in to it's parent method, since it is only called once
- Removes `getdata`, since it is only called once, and this removes the need for the `Collector` class


wiredfool:
Unfortunately, `getheader` and `getdata` are part of the public api, and from a quick search, they are used in code in the wild. 

What I'd really like to see is the elements of _save and the other helpers broken into smaller, functional, testable bits. 

`_get_palette_bytes` should have it's core modifications pulled out into a `compress_image_palette` method on `Image`, and it should return a new image, not mutate the existing image in place. (though, that might be better as a remap palette method, I'm not 100% on the right level of granularity there). That's going to require an explicit call, probably right after the convert_mode call in `_save`. (though, it's only going to be required in the case where the original frame was 'P' mode, because any other mode is going to already be compressed).  At that point, _get_palette_bytes could simply do exactly what the name suggests. 

 `getheader` and `getdata` can be reimplemented as combinations of a new _get_global_header method, and getdata possibly can be internally renamed to `_get_frame_data`. 

Then we can start to make a clear flow for the the _save method.  

Currently, if `saveall==True` and we've only got one frame, we actually generate the frame twice, once on the first run through in the saveall code, then we throws that away and write it in the single frame code.

radarhere:
Okay, I have updated this PR to restore `getheader` and `getdata`. I have separated out their internals into `get_global_header` and `_write_frame_data` respectively.

`_save` now has most of its code separated into `_write_single_frame` and `_write_multiple_frames`.

Your comment about generating the frame twice if `save_all` is true is already taken care of by a commit from #2317 that is included in this PR.

See what you think of that.

wiredfool:
I've pushed a bunch more refactorings on top of this. 

The biggest aim was to get rid of the REFACTOR SOONEST bit where the image is modified when getting the palette. It's now pulled out into a place where we explicitly do stuff to the palette and images. 

There's a bunch more documentation as well. 

I'm also explicitly handling the case where we're comparing two images that may have the same image bits but different palettes. 

I don't think that there are any changes to tests, so we're still passing everything we were doing before.

radarhere:
Thanks very much.

hugovk:
@radarhere This conflicts after merging https://github.com/python-pillow/Pillow/pull/2424.

radarhere:
@hugovk yep, rebased now.

Tests/test_file_gif.py
```diff
@@ -507,6 +505,35 @@ def test_save_I(self):
...
+        try:
+            h = GifImagePlugin.getheader(im, passed_palette)
+            d = GifImagePlugin.getdata(im)
+
+            import cPickle as pickle
+            # Enable to get target values on pre-refactor version
+            # with open('Tests/images/gif_header_data.pkl', 'wb') as f:
+            #    f.write(pickle.dumps((h,d)))
+            with open('Tests/images/gif_header_data.pkl', 'rb') as f:
+                (h_target,d_target) = pickle.loads(f.read())
```
hugovk: You need to be careful with the pickle protocol version, pickled files are not necessarily compatible with other Python versions..

Python 2 defaults to version 0:
https://docs.python.org/2/library/pickle.html#data-stream-format

Python 3 defaults to version 3 and is not compatible with Python 2:
https://docs.python.org/3/library/pickle.html#data-stream-format

It'd be safer to explicitly use protocol version 0, 1 or 2 in the save and load.

(Also PEP8 spaces after commas and colons.)

wiredfool: So, ignoring pep8 for a moment, this is a big PR and needs real review. 

This fixes at least a couple longstanding bugs, splits out image modification into explicit steps and retains compatibility with the earlier version of the legacy images.  Now that getheader/getdata are tests, I think we've got everything covered so we don't break anything. There is still at least one corner case where I'm not sure _why_ we're doing what we're doing, (the initial palette thing for 'RGB' base mode images). 

- Is this pythonic?  ey, maybe. could be better. 
- Is it clearer than it was?
- Is there anything that's confusing in the code?

wiredfool: And even worse, cPickle isn't on py3.

hugovk: Yeah, can do this:
```
try:
    # Python 2
    import cPickle as pickle
except ImportError:
    # python 3
    import pickle
```

wiredfool: Yeah, but it's not really worth that much bother. `pickle` is in both, it's clear, it works, and cPickle is only going to save milliseconds on py2.7 test runs.

wiredfool:
Needs release notes for changes and clarifications to the palette handling in save.
  * ImagePalette.ImagePalette now accepted
  * No longer modifies image in place. 

#2373

radarhere:
Due to the 'REBASE ME' message in the commit log, I have rebased this on the latest master.

wiredfool:
I was going to interactively rebase those show commands out of the test where they were added to clean up the history.   Basically, it looks like it should be fixup'd into f9d0deb5648bcca2d267b9662aae0665ca77d2aa.

wiredfool:
I've rebased that commit out, and reworded the earlier merge failure one in my gif branch. I'll force push to yours if ok.

radarhere:
Sure, go for it.

radarhere:
Thanks for all your work on this @wiredfool

wiredfool:
You're welcome.

PR #2375, opened 2017-01-26
Fixed bug when comparing frames in GifImagePlugin

radarhere:
https://github.com/python-pillow/Pillow/issues/2314#issuecomment-274265583 raises a bug.

In the first iteration of the frames loop, `getheader` is called, which in turn calls `_get_palette_bytes`. In this test case, that changes the image, and so changes the resulting delta to say that there are no differences between the first and the second frames.

So, this PR makes a copy of the frame before `getheader` is called, and that copy is used for the comparison.


wiredfool:
Test cherry picked into #2374

Issue #722, opened 2014-06-25
Test coverage >= 90%


aclark4life:
I'm not sure where the cutoff is, but it would be nice to get test coverage high enough to change the coveralls icon from red to not red. Currently we're at 72%.

radarhere:
The coveralls icon is now yellow, and says 80%!

hugovk:
Great!

https://coveralls.io/builds/9892857 says 79.977%, rounded up to 80% for the badge.

One last push and we can up the target to green 90% :)

hugovk:
https://coveralls.io/builds/10203543 now says 80.631%, rounded up to 81%!

![image](https://cloud.githubusercontent.com/assets/1324225/23106017/0dda807c-f6f0-11e6-9048-959de6dec4e8.png)

Let's push onwards and upwards to 90% for a green!

radarhere:
As of #3637, we are now using the Codecov badge. Note that it does not have the same boundaries for the badge color. According to https://docs.codecov.io/docs/coverage-configuration, it will gradually transition from red at 70% to green at 100%.

radarhere:
90%!

PR #2381, opened 2017-01-29
Test matrix convert

hugovk:
libImaging/Matrix.c was completely untested, so test using the example in the docs:

https://pillow.readthedocs.io/en/4.0.x/reference/Image.html#PIL.Image.Image.convert

Matrix.c is now 96.55% covered.

Helps for https://github.com/python-pillow/Pillow/issues/722, takes overall coverage from 79.977% to 80.155%.


wiredfool:
This leaves out a test for correctness, which is what I'm pushing for in #2373.   As a demonstration of the sort of obvious error that's not getting caught -- look at the test results from this patch: https://github.com/wiredfool/Pillow/commit/3a507b7c0b18e8a96c047557574d66416ed447f3  here: https://travis-ci.org/wiredfool/Pillow/builds/196540352 . They're all green. ( or at least, some are green and I expect they will all be in time)

96% coverage of that file sounds good, especially since the one line that's not covered is an error condition that should probably be filtered in the python code. So we've tested everything. Sort of.

That 96% test coverage is misleading, in that it reflects that branches were taken or code was run, but not that we actually agree with the results. There could be overflow errors, copy paste errors, or rounding errors, just to think of a few off the top of my head.  If we pass in an image with the identity matrix, do we get that image back?  Are we sure that all of the parameters of the matrix are in the right place? Did we transpose two of the factors?  If I were to reimplement this function using smid, asm, or some other method, would these tests be enough to determine if the new implementation was correct?

hugovk:
I've added a new test with the identity matrix. The output image is binary identical to the input.

I've added a similarity test for the XYZ converted image.

Here's the input image:
original-rgb

Here's the L and RGB-converted images:
converted-l converted-rgb

The colours are not the same as the input. I wonder if this is a bug in the production code, or expected result with that particular matrix? At least it's not completely mangled.

About parameter ordering, do you think it'd be better/safer as `im.convert(mode, matrix)` instead of `im.convert(mode=mode, matrix=matrix)`? (They are the first two parameters: `Image.convert(mode=None, matrix=None, dither=None, palette=0, colors=256)` https://pillow.readthedocs.io/en/4.0.x/reference/Image.html#PIL.Image.Image.convert)

wiredfool:
XYZ is a different color space that's typically used as an intermediate on the way to L*ab space. It's as different from RGB as CMYK is. I'd expect the luminance channel to be pretty close to an L image, but the other two channels are going to be on different axes than you might expect. (https://en.wikipedia.org/wiki/CIE_1931_color_space#Meaning_of_X.2C_Y.2C_and_Z)

The parameter order I'm concerned about is these lines:
https://github.com/wiredfool/Pillow/commit/3a507b7c0b18e8a96c047557574d66416ed447f3#diff-14ac03b70f5707b278e0ff12a809d479L61

The identity will nail down three of the parameters, the other factors would be nailed down with accurate XYZ conversions, or a specific test matrix. The only ones that the image conversions wouldn't nail down would be the constant offsets in the last column.

homm:
@wiredfool It is very cool to have correctness tests for core functionality. But when there are no such tests, I'd like to have at least coverage tests, which shows that the code at least runs without segmentation faults. Lack of correctness tests is not a reason to reject this tests. Just a several TODOs in test's code should me fine.

wiredfool:
So, I was able to generate this: 

from: 
hopper-xyz-separate

```
convert hopper.png -set colorspace RGB \( -clone 0 -colorspace XYZ +append \) -delete 0 -append -set colorspace RGB hopper-xyz.png
```

That's more complicated than it should be because it's really easy to wind up converting back to RGB if you're not being careful. However, it looks like a pretty good match for the XYZ conversion above.

PR #2382, opened 2017-01-29
Test linear and radial gradient effects

hugovk:
`ImagingFillLinearGradient` and `ImagingFillRadialGradient` in [Fill.c](https://github.com/python-pillow/Pillow/blob/6e7553fb0f12025306b2819b9b842adf6b598b2e/libImaging/Fill.c#L61-L104) are called by `_linear_gradient` and `_radial_gradient` in [_imaging.c](https://github.com/python-pillow/Pillow/blob/296fb5fe1da4e44a473862b8ba46669a98e5ec15/_imaging.c#L669-L689) and exposed as [`linear_gradient` and `radial_gradient`](https://github.com/python-pillow/Pillow/blob/296fb5fe1da4e44a473862b8ba46669a98e5ec15/_imaging.c#L3415-L3416), but not used anywhere in the Python layer.

Expose them in Image.py and test them.

Increases coverage of Fill.c from 64.1% to 94.87%, and Image.py and _imaging.c by a bit. Helps with https://github.com/python-pillow/Pillow/issues/722.

(Alternatively, they could be removed as they've probably not been used in over 20 years, if ever.)


Tests/test_image.py
```diff
@@ -315,6 +315,32 @@ def test_offset_not_implemented(self):
...
+
+            # Act
+            im = Image.linear_gradient(mode)
+
+            # Assert
+            self.assertEqual(im.size, (256, 256))
+            self.assertEqual(im.mode, mode)
+            self.assertEqual(im.getpixel((0, 0)), 0)
+            self.assertEqual(im.getpixel((255, 255)), 255)
+            im2 = Image.open('Tests/images/linear_gradient_{}.png'.format(
```
wiredfool: Is the P gradient any different than convert('P') of the 'L' file?

hugovk: No, they're identical.
```python
>>> from PIL import Image
>>> l = Image.open("Tests/images/linear_gradient_L.png")
>>> l2p = l.convert("P")
>>> l2p.save("Tests/images/linear_gradient_L2P.png")
```
```bash
$ diff linear_gradient_L2P.png linear_gradient_P.png
$
```

wiredfool: Ok, can we just have one of them in the test image archive?

wiredfool: What I meant is that we only need one target image, not that we only need to test one mode. 

e.g. `target = Image.open('Tests/images/linear_gradient.png').convert(mode)`

hugovk: Done!

Issue #2231, opened 2016-11-17
Sun Raster files with RLE encoding and depth 1 fail to load


mrenters:
### What did you do?
Tried to open a 1-bit RLE encoded Sun Raster file
### What did you expect to happen?
Expected image to load
### What actually happened?
Got a 'buffer overrun when reading image file' error
### What versions of Pillow and Python are you using?
Python 2.7.8, Pillow 3.3.1

Please include code that reproduces the issue and whenever possible, an image that demonstrates the issue. The best reproductions are self-contained scripts with minimal dependencies. If you are using a framework such as plone, django, or buildout, try to replicate the issue just using Pillow. 

Use image in attached ZIP file or create a new image in GIMP, 640x400 with white background. Select black brush, draw a line, convert Image->Mode to Indexed, and select black and white (1-bit) palette. Save image as Sun Raster file with RLE encoding.

[sunraster.zip](https://github.com/python-pillow/Pillow/files/597607/sunraster.zip)

```
>>> from PIL import Image
>>> im = Image.open('sunraster.im1')
>>> print im.format,im.size, im.mode
SUN (640, 400) 1
>>> im.load()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/rh/python27/root/usr/lib/python2.7/site-packages/PIL/ImageFile.py", line 240, in load
    raise_ioerror(err_code)
  File "/opt/rh/python27/root/usr/lib/python2.7/site-packages/PIL/ImageFile.py", line 59, in raise_ioerror
    raise IOError(message + " when reading image file")
IOError: buffer overrun when reading image file
>>>

```

wiredfool:
This is an error that's triggered from within the decoder, where it's being requested to write more into a buffer than remaining space in that buffer. It's checking, failing, and returning an error message.

RLE is run length encoding, in this case there's a marker for RLE, then a byte of how many, then a byte for the value. e.g. `0x80 0xff 0x00` is the code for 255 zeros 

What's going on here, at least initially, is that the Sun RLE decoder is expecting that when you're unpacking 255 items into a buffer, that there's at least 255 bytes of space in that buffer. 

```
                if (state->x + n > state->bytes) {
                    /* FIXME: is this correct? */
                    state->errcode = IMAGING_CODEC_OVERRUN;
                    return -1;
                }
```

[...]

PR #1938, opened 2016-05-31
RFC: Pure Python Decoder objects

wiredfool:
File format decoders are currently complicated to put together.
1. There's no registry, we're just doing a getattr on the image.core object. This PR adds a registry, then falls back to the current method. 
2. The decoders are all in C, a fast but unsafe language. Python is significantly slower, but safe. This is a tradeoff that we should be able to make. (in either direction)
3. Currently, if we do implement a decoder in python, the `load` phase is stuffed into the `open` phase.  This breaks the convention of open being a relatively quick operation and by deferring the heavy work to the load phase. As an added bonus, parsing errors here will no longer lead to "cannot identify image file" errors. 

This pr builds on the Python fd changes from #1934, but the underlying concept doesn't need to rely on that. It's still likely that the interface chosen there will change, requiring changes here. 

Changes proposed in this pull request:
- Add `Image.register_decoder` and `Image.register_encoder` to map decoder name -> a callable that returns a de/encoder.
- Adds the `ImageFile.PyDecoder` class that implements the interface used by the C layer decoder object. (Undone, an encoder object)
- Refactors the DDS image Plugin to use the new registry and decoder class.

PR #2384, opened 2017-01-29
More MSP tests

hugovk:

 * Test opening MSP file with a bad checksum. The test file was created by Pillow, temporarily hacked to save with a bad checksum.
 * Test Windows v2 MSP files. Fetches files via https://github.com/python-pillow/pillow-depends/pull/19.
 * Helps towards #722: 
   *  0% -> 82.14% MspDecode.c
   *  95.12% -> 100% MspImagePlugin.py
   * 83.81% -> 85.97% decode.c 
   * 79.977% -> 80.131% overall


Tests/test_file_msp.py
```diff
@@ -24,14 +27,43 @@ def test_invalid_file(self):
...
+                          lambda: MspImagePlugin.MspImageFile(bad_checksum))
+
+    def test_open_windows_v1(self):
         # Arrange
         # Act
         im = Image.open(TEST_FILE)
 
         # Assert
         self.assertEqual(im.size, (128, 128))
         self.assert_image_equal(im, hopper("1"), 4)
```
wiredfool: Why is the message here '4'?

Tests/test_file_msp.py
```diff
@@ -24,14 +27,43 @@ def test_invalid_file(self):
...
+        self.assertIsInstance(im, MspImagePlugin.MspImageFile)
+
+    @unittest.skipIf(not os.path.exists(EXTRA_DIR),
+                     "Extra image files not installed")
+    def test_open_windows_v2(self):
+        # Arrange
+        ImageFile.LOAD_TRUNCATED_IMAGES = True
```
wiredfool: Why are we suppressing errors here and not checking for correctness below?

Tests/test_file_msp.py
hugovk: Mistake. Removed.

Tests/test_file_msp.py
hugovk: The files are truncated and otherwise give:
```
  File "Tests/test_file_msp.py", line 60, in test_open_windows_v2
    im.load()
  File "/usr/local/lib/python2.7/site-packages/PIL/ImageFile.py", line 226, in load
    "(%d bytes not processed)" % len(b))
IOError: image file is truncated (0 bytes not processed)
```
These are the only example files I can find on the internet. I can't find any other software to generate them, or open these, and I've only got Windows 7 not Windows 1 or 2 :)

`im.show()`ing them gives reasonable-looking images, although with a black bar below the main image. The size fetching inside MspImageFile.py is the same for both Windows 1 and 2 formats, which looks sound for Windows 1, but MspDecode.c is only used for Windows 2 to do the actual decoding.

I've added some `getpixel` assertions.

wiredfool: Ah. It's YAFRLE implementation. 

That sort of thing really sounds like an RLE decoding error in MspDecode.c, as RLE is prone to underrun or overrun if it's interpreted incorrectly. (e.g. https://github.com/python-pillow/Pillow/issues/2231). I'd call this probably broken unless I can prove otherwise.

wiredfool:
- [ ] confirm correctness or failure of msp parsing re: truncated images

hugovk:
Unless the fix is simple this MSP code is probably a good candidate for removal, seeing as it's quite never worked in 20 years and no-one has complained about it, the software to create these files (Windows 1 and 2) have been unsupported since 2001, it was very hard to find any existing test files and it looks like no-one is using it anymore. Thoughts?

wiredfool:
So, basically, the MSP decoder wan't handling all the cases properly. There's a structure that was being ignored, the scan line map. That map gave an encoded byte length for each scan line, where some of them were 0. Those scan lines were not included in the decoded image. (Not that that method was actually mentioned in the documentation I could find.)

If we assume that the background color is white and add that for each 0 length line, all of the images from this PR (and all the images I could find) decompress to something that looks roughly intentional. Admittedly, some of the ones that I found look a lot like b/w dither/halftoning corresponding to colors, but the marbles one looks pretty good. 

This PR is now included in the Pure Python Decoder PR, as it's a low volume not terribly performance critical decoder, and I didn't really relish implementing the scan line map in C.

hugovk:
Merged in https://github.com/python-pillow/Pillow/pull/1938.

PR #2388, opened 2017-01-30
Add disposal option for saving GIF

linnil1:
After Reading [GIF89a spec](https://www.w3.org/Graphics/GIF/spec-gif89a.txt), 
I add disposal option which is important for GIF animation with transparent images.

This change is independent to other options.

You can see the difference of disposal option between two images below.
dispose_none
dispose_prev

### DIFF
--- a/PIL/GifImagePlugin.py
+++ b/PIL/GifImagePlugin.py
@@ -453,15 +453,24 @@ def _get_local_header(fp, im, offset, flags):
         duration = int(im.encoderinfo["duration"] / 10)
     else:
         duration = 0
+
+    if "disposal" in im.encoderinfo:
+        disposal = int(im.encoderinfo['disposal'])
+        if not 0<= disposal <=3:
+            disposal = 2
+    else:
+        disposal = 2
+
     if transparent_color_exists or duration != 0:
-        transparency_flag = 1 if transparent_color_exists else 0
+        packed_flag = 1 if transparent_color_exists else 0
+        packed_flag |= disposal << 2
         if not transparent_color_exists:
             transparency = 0
 
         fp.write(b"!" +
                  o8(249) +                # extension intro
                  o8(4) +                  # length
-                 o8(transparency_flag) +  # packed fields
+                 o8(packed_flag) +        # packed fields
                  o16(duration) +          # duration
                  o8(transparency) +       # transparency index
                  o8(0))
--- a/PIL/GifImagePlugin.py
+++ b/PIL/GifImagePlugin.py
@@ -454,14 +454,9 @@ def _get_local_header(fp, im, offset, flags):
     else:
         duration = 0
 
-    if "disposal" in im.encoderinfo:
-        disposal = int(im.encoderinfo['disposal'])
-        if not 0<= disposal <=3:
-            disposal = 2
-    else:
-        disposal = 2
+    disposal = int(im.encoderinfo.get('disposal', 0))
 
-    if transparent_color_exists or duration != 0:
+    if transparent_color_exists or duration != 0 or disposal:
         packed_flag = 1 if transparent_color_exists else 0
         packed_flag |= disposal << 2
         if not transparent_color_exists:
--- a/docs/handbook/image-file-formats.rst
+++ b/docs/handbook/image-file-formats.rst
@@ -127,6 +127,14 @@ are available::
 **palette**
     Use the specified palette for the saved image.
 
+**disposal**
+    Indicates the way in which the graphic is to be treated after being displayed.
+
+    * 0 - No disposal specified.
+    * 1 - Do not dispose.
+    * 2 - Restore to background color.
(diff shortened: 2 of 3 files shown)


hugovk:
Thanks, and for including documentation. Please could you include tests?
```
-------------
Diff Coverage
Diff: origin/master...HEAD, staged, and unstaged changes
-------------
PIL/GifImagePlugin.py (57.1%): Missing lines 458-460
-------------
Total:   7 lines
Missing: 3 lines
Coverage: 57%
-------------
-------------
Diff Quality
Quality Report: pyflakes
Diff: origin/master...HEAD, staged, and unstaged changes
-------------
docs/handbook/image-file-formats.rst (100%)
PIL/GifImagePlugin.py (100%)
-------------
Total:   19 lines
Violations: 0 lines
% Quality: 100%
-------------
-------------
Diff Quality
Quality Report: pep8
Diff: origin/master...HEAD, staged, and unstaged changes
-------------
docs/handbook/image-file-formats.rst (100%)
PIL/GifImagePlugin.py (90.9%):
    459: E225 missing whitespace around operator
-------------
Total:   19 lines
Violations: 1 line
% Quality: 94%
-------------
```
https://travis-ci.org/python-pillow/Pillow/jobs/196714202#L4705

PIL/GifImagePlugin.py
```diff
@@ -453,15 +453,24 @@ def _get_local_header(fp, im, offset, flags):
...
         duration = 0
+
+    if "disposal" in im.encoderinfo:
+        disposal = int(im.encoderinfo['disposal'])
+        if not 0<= disposal <=3:
+            disposal = 2
+    else:
+        disposal = 2
+
     if transparent_color_exists or duration != 0:
```
wiredfool: I don't think that this is the correct case anymore -- unless disposal is only valid with a transparent color or duration.

PIL/GifImagePlugin.py
```diff
@@ -453,15 +453,24 @@ def _get_local_header(fp, im, offset, flags):
         duration = int(im.encoderinfo["duration"] / 10)
     else:
         duration = 0
+
+    if "disposal" in im.encoderinfo:
```
wiredfool: Could this whole thing be as simple as:
```
disposal = int(im.encoderinfo.get('disposal', 2))
```
and then range check?

docs/handbook/image-file-formats.rst
```diff
@@ -127,6 +127,14 @@ are available::
 **palette**
     Use the specified palette for the saved image.
 
+**disposal**
+    Indicates the way in which the graphic is to be treated after being displayed.
+
+    * 0 - No disposal specified.
+    * 1 - Do not dispose.
+    * 2 - Restore to background color.
+    * 3 - Restore to previous content.
```
wiredfool: I'm not clear on what any of these options actually mean, without resorting to wikipedia.

wiredfool:
As an additional interface issue -- this would wind up with the same disposal for each frame. That doesn't sound especially flexible for the construction of animated gifs. We've recently added a per-frame duration. I can see this winding up going a similar direction. 

And, in a similar vein, if we read in an animated gif and tell it to save all, we're going to be using the passed in disposal for all frames, not using the original value.

docs/handbook/image-file-formats.rst
hugovk: Yep, I had to check against the spec: https://www.w3.org/Graphics/GIF/spec-gif89a.txt

```
iv) Disposal Method - Indicates the way in which the graphic is to
            be treated after being displayed.

            Values :    0 -   No disposal specified. The decoder is
                              not required to take any action.
                        1 -   Do not dispose. The graphic is to be left
                              in place.
                        2 -   Restore to background color. The area used by the
                              graphic must be restored to the background color.
                        3 -   Restore to previous. The decoder is required to
                              restore the area overwritten by the graphic with
                              what was there prior to rendering the graphic.
                     4-7 -    To be defined.
```

PIL/GifImagePlugin.py
linnil1: My fault. Thanks for suggestion

PIL/GifImagePlugin.py
```diff
@@ -454,14 +454,9 @@ def _get_local_header(fp, im, offset, flags):
     else:
         duration = 0
 
-    if "disposal" in im.encoderinfo:
-        disposal = int(im.encoderinfo['disposal'])
-        if not 0<= disposal <=3:
```
wiredfool: We'll probably still need this clause.

linnil1: This range check is not needed.
Because the higher bits is reserved, this option will not pollute another data.
And if it is higher than 0xffff(ushort), than will raise struct.error

wiredfool:
- [ ] update docs
 - [ ] Use named constants?

wiredfool:
See #2902

PR #2394, opened 2017-02-07
Additional docker builds

wiredfool:
More Docker images. 

 * Added Arch, Debian
 * Removed Nightly

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,22 +9,21 @@ notifications:
 
 matrix:
   fast_finish: true
-  allow_failures:
-    - python: nightly
   include:
     - python: "pypy"
     - python: "pypy3"
     - python: '3.6'
     - python: '2.7'
     - env: DOCKER="alpine"
+    - env: DOCKER="arch" # contains PyQt5
     - env: DOCKER="ubuntu-trusty-x86"
     - env: DOCKER="ubuntu-xenial-amd64"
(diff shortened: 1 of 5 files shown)


wiredfool:
I suppose his is somewhat unfair, but apart from the one known failure in Arch/qt5, it works on my machine.

I think I broke all the other docker images by updating and pushing them after upgrading my workstations docker version.

wiredfool:
So it looks like the docker failures are fixed by clearing out all of the images on the workstation and rebuilding them.

wiredfool:
Pillow apparently has trouble with Qt5:

* Saving 1bit images as a png through a qimage fail. We were testing the write, not the read.
* qpixmap operations are totally broken
* Our tests didn't track the class reorganization of qt4->5

wiredfool:
I think we're good here. 

* Arch and Xenial are now both testing QT5, our pixmap tests pass on Xenial but not Arch, so I'm going to chalk it up to a packaging issue. I'm going to file a bug with Arch when my account comes through. 
* 1 bit images don't save as png properly on either QT4 or 5. I don't think that's our issue.
* This PR is blocking https://github.com/python-pillow/docker-images/pull/3

wiredfool:
The QT5 changes are now also blocking other testing, since Xenial was updated to test it. 

Merging.

Issue #2401, opened 2017-02-13
Creating a multiframe image object


blochl:
Hi!

* I know that the ```save_all=True``` option saves all the frames of the image object for supported formats.
* I know that opening a multiframe image of a supported format creates a multiframe image object.

But how to create a new multiframe image object? Say from a 3D Numpy array? ```Image.fromarray``` accepts only 2D arrays...

Thanks!
Leonid.

radarhere:
You can create a series of single frame images, and then combine them using `append_images` - https://github.com/python-pillow/Pillow/blob/master/docs/releasenotes/3.4.0.rst#append-images-to-gif

blochl:
Hi, thanks for your answer. But it does not seem to work.
Code example:
```python
import numpy as np
from PIL import Image
a = np.random.rand(100,100,100)
a = ((a - a.min())/(a.max() - a.min()) * 255).astype(np.uint8)
imlist = []
for m in a:
    imlist.append(Image.fromarray(m))

imlist[0].save("test.tif", save_all=True, compression="tiff_deflate",
               append_images=imlist[1:])
```
Expected results: a 100-frame tiff.
Actual results: 1-frame tiff.

radarhere:
`append_images` is currently available only for GIF. Changing that, your example works for me.

blochl:
But is there currently a way to save multiframe tiffs (tiff image stacks) in Pillow?

radarhere:
You may not like this answer.

    import numpy as np
    from PIL import Image
    a = np.random.rand(100,100,100)
    a = ((a - a.min())/(a.max() - a.min()) * 255).astype(np.uint8)
    imlist = []
    for m in a:
        imlist.append(Image.fromarray(m))

    # By saving it as a GIF and then opening it and saving it as a TIFF
    imlist[0].save("test.gif", save_all=True, append_images=imlist[1:])
    Image.open("test.gif").save("test.tif", save_all=True)
    print(Image.open("test.tif").n_frames) #  100

    # To improve this somewhat, it can be done using BytesIO
    import io
    output = io.BytesIO()
    imlist[0].save(output, format="GIF", save_all=True, append_images=imlist[1:])
    Image.open(output).save("test.tif", save_all=True)
    print(Image.open("test.tif").n_frames) #  100

martmists:
Same result happens here, even though I'm trying to save to a .gif
when I print a list of frames to console:
[...]

PR #2404, opened 2017-02-16
Allow to save tiff stacks from separate images

blochl:
This is a quick solution that will allow to save tiff stacks from separate images, _e.g._ from Numpy arrays.
Previously, tiff stacks could be saved only from multiframe images.
This behavior is similar to what is possible now with GIFs.

Example:
```python
import numpy as np
from PIL import Image
a = np.ones((100,100,100), dtype=np.uint8)
imlist = []
for m in a:
    imlist.append(Image.fromarray(m))

imlist[0].save("test.tif", compression="tiff_deflate", save_all=True,
               append_images=imlist[1:])
```
...This should result in a 100-frame, 100x100 tiff stack. (Previously: would save only the first frame, despite ```save_all``` and ```append_images```.)

Please see the discussion here: #2401

### DIFF
--- a/PIL/TiffImagePlugin.py
+++ b/PIL/TiffImagePlugin.py
@@ -1737,17 +1737,28 @@ class AppendingTiffWriter:
                 self.rewriteLastLong(offset)
 
 def _save_all(im, fp, filename):
-    if not hasattr(im, "n_frames"):
+    encoderinfo = im.encoderinfo.copy()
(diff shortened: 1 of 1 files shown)


radarhere:
Hi. Thanks for this. Would you be able to add tests as well?

blochl:
Please see #2406 - test added (and is passing on 4.0.0).
I deleted my fork from before, so had to open a new one.

PR #2406, opened 2017-02-16
Allow to save tiff stacks from separate images

blochl:
This is a quick solution that will allow to save tiff stacks from separate images, _e.g._ from Numpy arrays. Previously, tiff stacks could be saved only from multiframe images.
This behavior is similar to what is possible now with GIFs. Note however, that for correct results, all the appended images should have the same ```encoder{info,config}``` properties.

Example:
```python
import numpy as np
from PIL import Image
a = np.ones((100,100,100), dtype=np.uint8)
imlist = []
for m in a:
    imlist.append(Image.fromarray(m))

imlist[0].save("test.tif", compression="tiff_deflate", save_all=True,
               append_images=imlist[1:])
```
...This should result in a 100-frame, 100x100 tiff stack. (Previously: would save only the first frame, despite ```save_all``` and ```append_images```.)

Please see the discussion here: #2401
And a previous pull request here: #2404 (I deleted my fork previously, so opening a new one)

### DIFF
--- a/PIL/TiffImagePlugin.py
+++ b/PIL/TiffImagePlugin.py
@@ -1737,17 +1737,28 @@ class AppendingTiffWriter:
                 self.rewriteLastLong(offset)
 
 def _save_all(im, fp, filename):
-    if not hasattr(im, "n_frames"):
+    encoderinfo = im.encoderinfo.copy()
+    encoderconfig = im.encoderconfig
+    append_images = encoderinfo.get("append_images", [])
+    if not hasattr(im, "n_frames") and not len(append_images):
         return _save(im, fp, filename)
 
     cur_idx = im.tell()
     try:
         with AppendingTiffWriter(fp) as tf:
-            for idx in range(im.n_frames):
-                im.seek(idx)
-                im.load()
-                _save(im, tf, filename)
-                tf.newFrame()
+            for ims in [im]+append_images:
+                ims.encoderinfo = encoderinfo
+                ims.encoderconfig = encoderconfig
+                if not hasattr(ims, "n_frames"):
+                    nfr = 1
+                else:
+                    nfr = ims.n_frames
+
+                for idx in range(nfr):
+                    ims.seek(idx)
+                    ims.load()
+                    _save(ims, tf, filename)
+                    tf.newFrame()
     finally:
         im.seek(cur_idx)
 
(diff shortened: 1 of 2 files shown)


hugovk:
@blochl Please can you check the failing test?

```
TestFileTiff.test_tiff_save_all ... ERROR
ERROR: TestFileTiff.test_tiff_save_all
  File "/home/travis/build/python-pillow/Pillow/Tests/test_file_tiff.py", line 477, in test_tiff_save_all
    mp.seek(0, os.SEEK_SET)
ValueError: I/O operation on closed file.
```

https://travis-ci.org/python-pillow/Pillow/jobs/202371600#L2750

blochl:
@hugovk , thanks, indeed I changed that part after the test have passed. Updated.

PIL/TiffImagePlugin.py
```diff
@@ -1737,17 +1737,28 @@ def fixOffsets(self, count, isShort=False, isLong=False):
...
-                im.load()
-                _save(im, tf, filename)
-                tf.newFrame()
+            for ims in [im]+append_images:
+                ims.encoderinfo = encoderinfo
+                ims.encoderconfig = encoderconfig
+                if not hasattr(ims, "n_frames"):
+                    nfr = 1
+                else:
+                    nfr = ims.n_frames
```
wiredfool: Can we do this in terms of just appending all frames of the current image to the new image? Getting the number of frames (at least in some cases) involves processing the entire image twice (namely, GIFs) while throwing away the results.

blochl: The best would be to be able to append images with different specifications as well, but even this is better than the current situation, when one can save a multiframe tiff only from another multiframe image. That's why I wrote "a quick solution" in the commit message.

PIL/TiffImagePlugin.py
```diff
@@ -1737,17 +1737,28 @@ def fixOffsets(self, count, isShort=False, isLong=False):
                 self.rewriteLastLong(offset)
 
 def _save_all(im, fp, filename):
-    if not hasattr(im, "n_frames"):
+    encoderinfo = im.encoderinfo.copy()
+    encoderconfig = im.encoderconfig
+    append_images = encoderinfo.get("append_images", [])
+    if not hasattr(im, "n_frames") and not len(append_images):
         return _save(im, fp, filename)
```
radarhere: The way that you've setup writing multiple frames, for a single frame image, there's no difference in outcome between multiple and single saving. Sure, a few more methods are called, but they're not expensive. So I would think this check is unnecessary?

PR #2409, opened 2017-02-17
Clarify that draft is only implemented for JPEG and PCD

wiredfool:

 * Documentation, clarifying that draft doesn't do anything for most images.

### DIFF
--- a/PIL/Image.py
+++ b/PIL/Image.py
@@ -1051,7 +1051,7 @@ class Image(object):
         in place.  If the image has already been loaded, this method has no
         effect.
 
-        Warning: This method is not implemented for most images. It is
+        Note: This method is not implemented for most images. It is
         currently implemented only for JPEG and PCD images.
 
         :param mode: The requested mode.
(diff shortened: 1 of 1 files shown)


PIL/Image.py
```diff
@@ -1051,6 +1051,9 @@ def draft(self, mode, size):
         in place.  If the image has already been loaded, this method has no
         effect.
 
+        Warning: This method is not implemented for most images. It is
```
homm: why warning?

wiredfool: Seemed like it was appropriate. People have been confused before.

homm: Yeah, but I mean why this is "warning"? This is not what you should afraid of. I believe that it's better to clarify things instead in the second sentence:

> For example, you can use this method to convert a color JPEG to greyscale while loading it, or to extract a 128x192 version from a PCD file.

Could be something like this:

> This method currently implemented only for two formats. You can load JPEG images with DCT scaling or convert a color JPEG to greyscale while loading it, or to extract a 128x192 version from a PCD file.

PR #2412, opened 2017-02-17
Remove unused private function

wiredfool:

 * Remove unused, undocumented private `Image._expand` function

### DIFF
--- a/PIL/Image.py
+++ b/PIL/Image.py
@@ -1055,12 +1055,6 @@ class Image(object):
         """
         pass
 
-    def _expand(self, xmargin, ymargin=None):
-        if ymargin is None:
-            ymargin = xmargin
-        self.load()
-        return self._new(self.im.expand(xmargin, ymargin, 0))
-
     def filter(self, filter):
         """
         Filters this image using the given filter.  For a list of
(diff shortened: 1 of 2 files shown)


homm:
Despite the fact that the method name starts with underscore, this is not true private method. I would be very careful removing methods from the main Pillow's object.

Another question do we need to remove it at all? As I can see, there are `ImagingExpand` and `_expand_image` functions in C layer, which are used only by `Image.im.expand()`. I believe one of the following is true: this function can be easily replaced with another size-manipulation methods (`crop`?) and should be completely removed; or it should be exposed in `Image`.

hugovk:
About the first paragraph, we can remove underscore-prefixed methods without deprecation warnings:

> “Private” instance variables that cannot be accessed except from inside an object don’t exist in Python. However, there is a convention that is followed by most Python code: a name prefixed with
an underscore (e.g. `_spam`) should be treated as a non-public part of the API (whether it is a function, a method or a data member). It should be considered an implementation detail and subject to change without notice.

https://docs.python.org/2/tutorial/classes.html#private-variables-and-class-local-references

PR #2413, opened 2017-02-18
Replaced broken URL

radarhere:
http://partners.adobe.com/public/developer/en/ps/5002.EPSF_Spec.pdf no longer works, so this PR replaces it with a web.archive.org URL.

Also replaces a http URL with https, in the spirit of #2403

### DIFF
--- a/PIL/EpsImagePlugin.py
+++ b/PIL/EpsImagePlugin.py
@@ -321,7 +321,7 @@ class EpsImageFile(ImageFile.ImageFile):
             # EPS can contain binary data
             # or start directly with latin coding
             # more info see:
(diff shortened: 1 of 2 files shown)


homm:
There is at least one another copy which can be easily googled:
https://www-cdf.fnal.gov/offline/PostScript/5002.PDF

Should we refer to live site instead of archive?

radarhere:
Hmm, I'm not convinced of this. For one thing, the Collider Detector at Fermilab have no obligation to continue hosting the file.

wiredfool:
Looks like partners.adobe.com was completely reorganized -- I've noticed that the TIFF spec has been moved as well.

Issue #2396, opened 2017-02-10
Pillow downloads count


radarhere:
While we may just wait for this to be resolved, it looks to me like the downloads count is broken.

![downloads](https://cloud.githubusercontent.com/assets/3112309/22841419/a68c4412-f025-11e6-929f-e7ab37483986.png)

hugovk:
We're using a shields.io badge, here's their issue: https://github.com/badges/shields/issues/716

PR #2434, opened 2017-03-02
Remove commented download-count badge

hugovk:
The commenting out of https://github.com/python-pillow/Pillow/pull/2432 didn't work, so let's just remove it until the badge is fixed.

---

image


radarhere:
Just to link, this relates to Issue #2396

PR #2441, opened 2017-03-04
Test BufrStubImagePlugin for correctness

hugovk:

 * Test image from NOAA: http://www.ftp.ncep.noaa.gov/data/nccf/com/gfs/prod/gfs.2017030406/ ( "The information on government servers are in the public domain, unless specifically annotated otherwise... etc." https://www.esrl.noaa.gov/gmd/about/disclaimer.html)
 * Test file operations work as expected, given it's a stub with no handler available 
* Only need to read 4 bytes to accept, not 8

Helps for #2373 and #722.


Tests/test_file_bufrstub.py
```diff
@@ -1,21 +1,45 @@
...
         self.assertRaises(SyntaxError,
                           lambda:
                           BufrStubImagePlugin.BufrStubImageFile(invalid_file))
 
+    def test_load(self):
+        # Arrange
+        im = Image.open(TEST_FILE)
+
+        # Act / Assert: stub cannot load without an implemented handler
+        self.assertRaises(IOError, lambda: im.load())
```
radarhere: Thanks for all your work on this, but yeah, unnecessary lambda.

PR #2442, opened 2017-03-04
Test XVThumbImagePlugin for correctness

hugovk:

 * First test image created using netpbm: `brew install netpbm`,  `pngtopam hopper.png | pamtoxvmini > hopper.p7`, second by editing hopper.p7 in a text editor
 * Test file open work as expected
 * Includes a Python 3 fix, a general fix, and a refactor of XVThumbImagePlugin

Helps for #2373 and #722.


PIL/XVThumbImagePlugin.py
```diff
@@ -58,14 +58,14 @@ def _open(self):
             s = self.fp.readline()
             if not s:
                 raise SyntaxError("Unexpected EOF reading XV thumbnail file")
-            if s[0] != b'#':
+            if i8(s[0]) != 35:  # ie. when not a comment: '#'
```
wiredfool: This seems like a step backwards in terms of readability

Tests/test_file_xvthumb.py
```diff
@@ -1,13 +1,35 @@
...
 
 class TestFileXVThumb(PillowTestCase):
 
+    def test_open(self):
+        # Act
+        im = Image.open(TEST_FILE)
+
+        # Assert
+        self.assertEqual(im.format, "XVThumb")
+        self.assert_image_similar(im, hopper("P"), 49)
```
wiredfool: epsilon=49?  That's quite high for a reasonable image. Is this really right?

hugovk: `im` looks like this:
image

PIL/XVThumbImagePlugin.py
hugovk: Yeah, it's not great. What would be the way to do it?
Python 2: `s[0] == "#"`
Python 3: `s[0] == 35`

wiredfool: right. one of the things that python3 got wrong.

Tests/test_file_xvthumb.py
wiredfool: 49 is not exactly uncorrelated, but it's not a recognizable match either. FWIW, ~85 is totally uncorrelated. 

`assert_image_similar` doesn't actually do palettes.  It converts the data channels to 'L' mode, and then does the diffs on that. The image palettes aren't taken into account. 

`assert_image_equal` does, but only because it runs through `im.tobytes()` rather than `im.getdata()`.

In this case, the correct method is probably to convert both of them to `RGB` first or to quantize `hopper('RGB')` to the funky RGB;332 palette.

hugovk: Here's what `im.convert("RGB")` looks like.
It has epsilon of 42.0398 against `hopper("RGB")`:
image

How do we quantise it to funky palette?

wiredfool: It's the palette at the top of XVImagePlugin. We should be able to use Image.quantize with (a version of) that palette to get something that's very close to an XV image. 

((and incidentally, that's probably a way to save it as well. Quantize to that palette, then call getdata and write that to the file))

hugovk: Here's what `hopper().quantize(palette=im)` looks like.
It has an epsilon of 8.6859 against `im`.
image

wiredfool: That's a lot better. The difference looks to be in the dither, which is a judgement call for quantization. 

((you know, I bet the original was a truncation.)) 

Anyway, I think this is good enough.

PR #2447, opened 2017-03-13
Test MicImagePlugin for correctness

hugovk:

 * Image created with Microsoft Image Composer 1.5 from hopper.png

Helps for #2373 and #722.

Needs work, see review comments.


Tests/test_file_mic.py
```diff
@@ -1,10 +1,46 @@
...
 
+    def test_sanity(self):
+        im = Image.open(TEST_FILE)
+        im.load()
+        self.assertEqual(im.mode, "RGBA")
+        self.assertEqual(im.size, (128, 128))
+        self.assertEqual(im.format, "MIC")
+
+        im2 = hopper("RGBA")
+        self.assert_image_similar(im, im2, 123.5)
```
hugovk: Epsilon is 123.4919! `im.show()` gives:

im.show()

`im2.show()` looks like:

im2.show()


hopper.mic opened in Microsoft Image Composer looks like:
hopper.mic in MIC

wiredfool: There certainly appear to be gamma or colorspace issues. I'd almost believe sRGB/naive colorspace. But that's not enough to account for the epsilon. You might want to look at the alpha, it's not going to show up in `im.show()` but if it's 0 vs 255, then that would account for enough of the rest.

wiredfool: Ok, there's a gamma of 2.2 encoded into the file, as per some of the defaults in the Tiff spec, but there's no indication in the tags that there is a gamma applied. With these changes, we get a much tighter epsilon=10:

```
        lut = ImagePalette.make_gamma_lut(1/2.2)
        im = Image.merge('RGBA', [chan.point(lut) for chan in im.split()])

        im2 = hopper("RGBA")
        self.assert_image_similar(im, im2, 10)
```

This is perhaps not the place to do it, but it does explain the results. Also, I think we need some form of 'apply gamma' function on an image, because splitting and recombining should be something we encapsulate, not require.

wiredfool:
The big question now is if Mic images have an embedded 2.2 gamma in general, and we should apply it, or if it's just this one.

hugovk:
I couldn't find any MIC images when I hunted earlier, but can convert more using Microsoft Image Composer if that'd help.

http://fileformats.archiveteam.org/wiki/Microsoft_Image_Composer gives Konvertor, XnView and ThumbsPlus under software, unfortunately all closed source.

Tests/test_file_mic.py
wiredfool: I think we should go with this in the test, and note somewhere in the docs that there may be an embedded gamma of 2.2

hugovk:
By the way, looks like these CodeCov comments are only coming from old branches that don't contain codecov.yml, so they'll dry up soon.

wiredfool:
Yeah, I gathered that.

PR #2457, opened 2017-03-23
Update setup.py

lishuo9527:
When I Install Pillow with pip:

`pip install Pillow`

It will response:
```
  File "setup.py", line 138
    required = {'jpeg','zlib'}
                      ^
SyntaxError: invalid syntax
```
so, I make a change :
```
   required = ['jpeg','zlib']
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -131,7 +131,7 @@ class pil_build_ext(build_ext):
         features = ['zlib', 'jpeg', 'tiff', 'freetype', 'lcms', 'webp',
                     'webpmux', 'jpeg2000', 'imagequant']
 
-        required = {'jpeg', 'zlib'}
+        required = ['jpeg', 'zlib']
 
         def __init__(self):
             for f in self.features:


wiredfool:
It's set notation, and supported in all currently supported versions of python -- 2.7 and 3.3 onwards.

lishuo9527:
ok

lishuo9527:
Ok. I found the versions of python is 2.6 on my machine .
 
But why not use ` required = ['jpeg','zlib']` instead of  `required = {'jpeg','zlib'}` ?

hugovk:
Because that is not the only line of 2.6 incompatible code and we've decided not to support 2.6 any longer because CPython 2.6 itself is no longer supported as of 2013: https://en.wikipedia.org/wiki/CPython#Version_history

See https://github.com/python-pillow/Pillow/pull/2192 for more on why 2.6 was dropped.

You have two options:

* Upgrade your Python to 3.6 or 2.7
 * Install Pillow < 4

Upgrading Python is the better choice, as more and more other packages are dropping 2.6, inducing pip: https://mail.python.org/pipermail/distutils-sig/2016-November/029785.html

lishuo9527:
Thanks!

PR #2465, opened 2017-03-26
Added checksum verification to download-and-extract.sh

radarhere:


hugovk:
Before we add more, I was actually going to ask about the checksum verification in [winbuild](https://github.com/python-pillow/Pillow/blob/bbc205573d2bf17306b5a4ab2e283fdbc3a21de5/winbuild/build_dep.py#L17): it must be tedious manually generating the hash for every new update, and then updating it in two or three places.

Is this manual work adding us much value? Have we ever had a checksum mismatch caused by download error? The files are (all?) coming from our own pillow-depends repo, so we control what goes up. Do we not trust the files we download are really the ones we request?

wiredfool:
I threw that in there in the first cut because we were doing some non-https downloads or downloads from somewhat sketchy companies (sf) where they were known to do things to their downloads at times. They never got checked because I just never got that far. 

If we're downloading from github over https, the manifest and the files are essentially under the same trust boundary and either could be attacked with the same opportunity. So I'm not sure that it really buys us much.

PR #2466, opened 2017-03-26
Run faster Dockers last so slower jobs don't hold up the build

hugovk:
Run the faster Docker jobs last (which are all under two minutes), so the slower ones don't hold up the build.

Similarly, this is why the slow PyPy jobs are first.

This PR:
* Ran for 9 min 40 sec https://travis-ci.org/python-pillow/Pillow/builds/215293752
* Ran for 9 min 55 sec https://travis-ci.org/hugovk/Pillow/builds/215289384

Older master builds:
* Ran for 10 min 8 sec https://travis-ci.org/python-pillow/Pillow/builds/215172329
* Ran for 12 min 32 sec https://travis-ci.org/python-pillow/Pillow/builds/214990466
* Ran for 9 min 58 sec https://travis-ci.org/python-pillow/Pillow/builds/214990372
* Ran for 11 min 45 sec https://travis-ci.org/python-pillow/Pillow/builds/214012098 
* Ran for 10 min 35 sec https://travis-ci.org/python-pillow/Pillow/builds/214011829

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ notifications:
 
 # Run slow PyPy* first, to give them a headstart and reduce waiting time.
 # Run latest 3.x and 2.x next, to get quick compatibility results.
-# Then run the remainder.
+# Then run the remainder, with fastest Docker jobs last.
 
 matrix:
   fast_finish: true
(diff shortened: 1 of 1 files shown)


wiredfool:
I don't know. There are times when I want the greatest diversity of runs quickly to pick out if a solution is going to fail or pass, and there are times (particularly around releases) when I'd just like to get the total runtime as low as possible.

PR #2472, opened 2017-04-01
DPI is a tuple

hugovk:
Fix merged PR https://github.com/python-pillow/Pillow/pull/2449, re: https://github.com/python-pillow/Pillow/commit/977f3192349282ae1a7108102415a7ad1baaa305#commitcomment-21546844

DPI should be an (xres, yres) tuple.

Confirmed in [code](https://github.com/python-pillow/Pillow/search?l=Python&q=dpi&type=&utf8=%E2%9C%93) and [docs](https://pillow.readthedocs.io/en/4.0.x/handbook/image-file-formats.html?highlight=dpi#jpeg):

> The `open()` method may set the following info properties if available:
> ...
> **dpi**
> A tuple representing the reported pixel density in pixels per inch, if the file is a jfif file and the units are in inches.
> ...
> The `save()` method supports the following options:
> ...
> **dpi**
A tuple of integers representing the pixel density, `(x,y)`.

Note: one test image (Tests/images/exif_gps.jpg) only has an X value, and so raised exceptions, so if there's no Y value, use the X twice.


PIL/JpegImagePlugin.py
```diff
@@ -119,8 +119,16 @@ def APP(self, marker):
 
     # If DPI isn't in JPEG header, fetch from EXIF
     if "dpi" not in self.info and "exif" in self.info:
-        x_resolution = self._getexif()[0x011A]
-        self.info["dpi"] = x_resolution[0] / x_resolution[1]
+        exif = self._getexif()
+        x_resolution = exif[0x011A]
```
wiredfool: This could fail with a keyerror?

hugovk: Theoretically, yes. I'll wrap in another try/except later.

hugovk: Updated, but see below.

hugovk:
It's better to check the spec before implementing things...

http://www.sno.phy.queensu.ca/~phil/exiftool/TagNames/EXIF.html says:

* Exif.Image.XResolution: "The number of pixels per \<ResolutionUnit> in the \<ImageWidth> direction. When the image resolution is unknown, 72 [dpi] is designated."
* Exif.Image.YResolution: "The number of pixels per \<ResolutionUnit> in the \<ImageLength> direction. The same value as \<XResolution> is designated."
* Exif.Image.ResolutionUnit: "The unit for measuring \<XResolution> and \<YResolution>. The same unit is used for both \<XResolution> and \<YResolution>. If the image resolution is unknown, 2 (inches) is designated."

And http://www.sno.phy.queensu.ca/~phil/exiftool/TagNames/EXIF.html gives these for resolution unit:
> (the value 1 is not standard EXIF) 
1 = None 
2 = inches 
3 = cm

So I'll update this again: 
* we don't need to check for YResolution, it's the same as XResolution - is that correct?
* if no value, default to 72,72 dpi
* should also check the unit: if inches, do nothing; if cm, convert; else use default

wiredfool:
@hugovk , you happy with this?

hugovk:
@wiredfool Yes.

Feel free to squash and merge if you'd like to tidy up the commits.

image

wiredfool:
someone is screen shotting on a high dpi monitor...

hugovk:
Yes, that was a surprisingly large image!

PR #2487, opened 2017-04-12
Remove broken downloads badge

hugovk:
Let's remove the downloads badge as it's broken and showing 0 downloads/month.

---

image

---

https://pillow.readthedocs.io/en/4.1.x/index.html

See also https://github.com/python-pillow/Pillow/issues/2396 and https://github.com/badges/shields/issues/716.


aclark4life:
Thanks

PR #2488, opened 2017-04-13
Update Image.py

p0123n:
Trying to save some memory when many pictures are opened at a time.

Fixes # Closing a previously opened resource.

 * Memory usage fix


p0123n:
seems like need to finalize

PR #2493, opened 2017-04-19
Undef PySlice_GetIndicesEx, see https://bugs.python.org/issue29943

### LINKED ISSUE #2479: Pillow 4.1.0 wheels for Windows and Python 3.6 seem to be broken
I’m using Pillow in a project. Our AppVeyor tests started failing today, and it seems that `Pillow-4.1.0-cp36-cp36m-win32.whl` is broken. This happens with our 3.6 tests only; 2.7 works fine. The most basic import fails with a (vague) DLL error:

```pytb
  File "C:\projects\nikola\nikola\plugins\task\galleries.py", line 43, in <module>
    from PIL import Image  # NOQA
  File "c:\python36\lib\site-packages\PIL\Image.py", line 56, in <module>
    from . import _imaging as core
ImportError: DLL load failed: The specified procedure could not be found.
```

wiredfool:
https://bugs.python.org/issue29943 

Fixes #2479.

From @cgohlke https://github.com/python-pillow/Pillow/issues/2479#issuecomment-295367838

### DIFF
--- a/_webp.c
+++ b/_webp.c
@@ -1,5 +1,6 @@
 #define PY_SSIZE_T_CLEAN
 #include <Python.h>
+#include "Imaging.h"
 #include "py3.h"
 #include <webp/encode.h>
 #include <webp/decode.h>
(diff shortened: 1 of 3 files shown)


wiredfool:
I've verified that this works as expected on Linux, compiling the Pillow package on 3.6.1 and running on 3.6.0.

PR #2502, opened 2017-05-05
Add new2 function to pildriver.py

phuslu:
Currently `new` function creates a grayscale image in stack, it will break `paste` a colorful(i.e. `RGBA`) image.

Repro Steps
```
pildriver.py save new.png paste open test1.png 0 200 paste open test2.png 0 0 new 400 400 0
```
We will get a **grayscale**  new.png rather than a **RGBA** one.

This commit is to fix this by adding a new `new2` function.
```
pildriver.py save new.png paste open test1.png 0 200 paste open test2.png 0 0 new2 RGBA 400 400 0
```

### DIFF
--- a/Scripts/pildriver.py
+++ b/Scripts/pildriver.py
@@ -122,6 +122,17 @@ class PILDriver(object):
         color = int(self.do_pop())
         self.push(Image.new("L", (xsize, ysize), color))
 
+    def do_new2(self):
+        """usage: new2 <string:mode> <int:xsize> <int:ysize> <int:color>:
+
+        Create and push a image of given mode, size and color.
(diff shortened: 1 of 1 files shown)


wiredfool:
Looks like a good idea. I'd like a better name than new2, perhaps new_mode or something that gives a clue about what the difference is between new and new2. 

It appears that we have no tests on pildriver (or likely, any of the scripts). It would be a good thing to add some tests to document that it's doind what we expect.

hugovk:
Closing this as these scripts now live in their own repo at https://github.com/python-pillow/pillow-scripts.

PRs welcome over there.

PR #2507, opened 2017-05-06
Updated references to point to existing files

radarhere:


homm:
These tutorials not related o tests. Most likely `hopper.jpg` doesn't exist in current directory as well as `lena.jpg`. I think `lena.jpg` is more recognizable name.

Also, some paths start with `Pillow/Tests/images`, other with `Tests/images`.

docs/handbook/tutorial.rst
```diff
@@ -499,9 +499,9 @@ Reading from a tar archive
 
 ::
 
-    from PIL import TarIO
+    from PIL import Image, TarIO
 
-    fp = TarIO.TarIO("Imaging.tar", "Imaging/test/lena.ppm")
+    fp = TarIO.TarIO("Tests/images/hopper.tar", "hopper.jpg")
```
hugovk: Is the second format intentionally jpg rather than ppm?

hugovk:
These are just examples and not real code, but I approve of the change to a file which does exist in the codebase, and disassociates with sexist imagery; eg. https://www.theatlantic.com/technology/archive/2016/02/lena-image-processing-playboy/461970/.

For the paths I'm not too bothered: either leave as is; update so they'd all work from the `Pillow` directory (it's probably an unlikely use case to checkout this whole repo first); or remove paths entirely so you'd have to run it from the `images` directory, or probably more likely, move the file to your project directory.

docs/handbook/tutorial.rst
wiredfool: Well, it does actually work. So there's that going for it.

Issue #2508, opened 2017-05-08
Image.show() convert to non-transparent format before mis-displaying


HinTak:
See discussion starting with:

https://github.com/rougier/freetype-py/pull/55#issuecomment-299511358

`Image.open()` this:
https://cloud.githubusercontent.com/assets/327203/25754502/be994452-31bf-11e7-85da-9cc7fb2811bd.png
then `show()` result in losing the transparency layer.

The problem is that PIL saves to a non-transparent capable PPM format first, before displaying the temporary file.

HinTak:
I have testing the proposed change
( https://github.com/python-pillow/Pillow/pull/2509 ) and it is indeed using png as intermediate now.
But the transparency information is still lost and an RGBA png is saved as RGB-only png.
So somehow the transparency information is lost between the Image class and the `PIL/PngImagePlugin.py`.
@rougier .

HinTak:
I am testing on git HEAD 1e0fbc0f1a524413e6fe77d0cb8de6b1be781302

HinTak:
This saves with transparency:
```
from PIL import Image
a = Image.open('emoji-color-cairo.png')
a.save('out.png')
```

so it is a bit curious why `a.show()` save a non-transparent copy, and where the transparency is lost.
```

HinTak:
Added more commits with the down-convert problem, now the group of changes works as intended.

PR #2509, opened 2017-05-08
use PNG as intermediate format for viewing

HinTak:
This is part way towards fixing https://github.com/python-pillow/Pillow/issues/2508

Since both Apple's Preview and unix's imagemagick are capable of displaying png, (and all the linux distros which distributes xv distribute xv patched with png support), it is better to use PNG as an intermediate for displaying.

 * PNG as intermediate format for viewing

### DIFF
--- a/PIL/ImageShow.py
+++ b/PIL/ImageShow.py
@@ -115,7 +115,7 @@ if sys.platform == "win32":
 elif sys.platform == "darwin":
 
     class MacViewer(Viewer):
-        format = "BMP"
+        format = "PNG"
 
         def get_command(self, file, **options):
             # on darwin open returns immediately resulting in the temp
@@ -142,6 +142,8 @@ else:
         return None
 
     class UnixViewer(Viewer):
+        format = "PNG"
+
         def show_file(self, file, **options):
             command, executable = self.get_command_ex(file, **options)
             command = "(%s %s; rm -f %s)&" % (command, quote(file),


homm:
> it is better to use PNG as an intermediate for displaying.

Why? PNG is significant slower, especially for encoding

HinTak:
Because of https://github.com/python-pillow/Pillow/issues/2508 - one want to save to a transparency-capable format, before displaying.

homm:
Sorry I missed this.

As I know there is 32-bit format for BMP, but none of tested programs consider extra bits as alpha channel. So, yes, we need to change format to something else. If it should be PNG, I'd like to suggest to add `compress_level=1` to the save parameters. It will speedup compression about 2 times for large files.

HinTak:
That's okay. I am not quite getting the event between `save()` and `show()` - the former do save to RGBA png, the latter does RGB png after this change, so the two code    path is slightly different. I am all for displaying quicker! In fact I am not entirely sure why one cannot just display the original (if the viewer is capable, and the file handle is still around).

HinTak:
I am not sure about (1) whether there are RGBA mode image that the png plugin cannot save, (2) the handshake seems a bit ugly - would be nice to set encoding option in a neater way.

But other than these concerns, the fix seems to work okay for my purpose, the file I mentioned above.

hugovk:
It might be tricky to fully test on the CI, but can you add unit tests for your changes? 

https://travis-ci.org/python-pillow/Pillow/jobs/230619667#L4942

HinTak:
How do you propose this being auto-tested? It mainly affects (interactive?) viewing...

PIL/PngImagePlugin.py
```diff
@@ -687,6 +687,9 @@ def _save(im, fp, filename, chunk=putchunk, check=0):
             mode = "%s;%d" % (mode, bits)
 
     # encoder options
+    if (im.info.get('showing', False)):
+        im.encoderinfo['compress_level'] = 1
+        im.info.pop('showing', None)
```
wiredfool: This is not an appropriate separation of concerns. The PNG encoder should not care that it's being called from within the show method. 

A better change would be to the internal `Image._dump` command. It should be altered to take `**options` to be passed into the save command
```
    def _dump(self, file=None, format=None, **options):
        import tempfile
        suffix = ''
        if format:
            suffix = '.'+format
        if not file:
            f, file = tempfile.mkstemp(suffix)
            os.close(f)

        self.load()
        if not format or format == "PPM":
            self.im.save_ppm(file, **options)
        else:
            if not file.endswith(format):
                file = file + "." + format
            self.save(file, format, **options)
        return file
```
(and renaming file to something that does not shadow the builtin)

PIL/ImageShow.py
```diff
@@ -69,8 +69,9 @@ def show(self, image, **options):
             # FIXME: auto-contrast if max() > 255?
         else:
             base = Image.getmodebase(image.mode)
-        if base != image.mode and image.mode != "1":
+        if base != image.mode and image.mode != "1" and image.mode != "RGBA":
             image = image.convert(base)
```
wiredfool: Is this going to break windows?

HinTak: No. See first commit - I only change the intermediate format for Unix and Mac os X.

PIL/PngImagePlugin.py
HinTak: I have not figured out how **option are used, and there aren't examples...

PIL/ImageShow.py
wiredfool: Yes, but you're no longer squashing RGBA->RGB for BMPs on Windows.

PIL/PngImagePlugin.py
wiredfool: `**options` turn into `encoderinfo` in the save method, so you can call save like: `im.save('foo.png', compress_level=1)` to pass the compression level into the PNG encoder.

HinTak: What do you mean by "(and renaming file to something that does not shadow the builtin)"?

wiredfool: `file` is a python builtin. It's bad form to have a local variable with that name. path, or temp_file are better.

But, looking now, it's in the function signature, which means finding any other place where it might have been called with kwargs.  At least it's an internal function so we can change it.

PIL/ImageShow.py
HinTak: I'll check if RGBA saves as BMP . Would be surprised if it doesn't...

PIL/PngImagePlugin.py
HinTak: That's the part I haven't worked out - what's encoderinfo attached to...

wiredfool: The keyword options to save turn into im.encoderinfo.

PIL/ImageShow.py
wiredfool: IIRC, BMP + alpha is somewhat inconsistently implemented in the viewer. I'm not familiar with what it does on windows.

HinTak: I just tried saving with and without alpha like this to BMP:

```
>>> from PIL import Image
>>> a = Image.open("emoji-color-cairo.png")
>>> a.save("one.bmp")
>>> b = a.convert("RGB")
>>> b.save("two.bmp")
```
It results in two substantially different bmp in size:

```
-rw-rw-r--. 1 Hin-Tak Hin-Tak 204854 May 10 20:35 one.bmp
-rw-rw-r--. 1 Hin-Tak Hin-Tak 153654 May 10 20:35 two.bmp
```

xv shows them to be visually the same, but imagemagick's `display` shows the former with alpha like the origi. So BMP + alpha is, as you said, works on some viewer but not others. On windows, according to `PIL/ImageShow.py` line 108 (a few lines above my first change, which switches for darwin and unix), it launches `start "Pillow" /WAIT file`. This seems to be some sort of built-in viewer in Pillow?

HinTak: IMHO, as long as the windows viewer does not choke on the alpha-enabled BMP, it is no worse than the current situation, which is already wrong.

hugovk: `'start "Pillow" /WAIT "%s" '`

No, it's not a built-in viewer, `"Pillow"` is a required title:

```
Syntax
      START "title" [/D path] [options] "command" [parameters]

Key:
   title       Text for the CMD window title bar (required.)
   path        Starting directory.
   command     The command, batch file or executable program to run.
   parameters  The parameters passed to the command.
```
https://ss64.com/nt/start.html

It just tells Windows to open an image file by giving the path to the file, and it'll open it in whatever program is associated with that file type. Usually IrfanView for me, but we need it in a format that will open in something default like MS Paint.

PIL/PngImagePlugin.py
HinTak: okay, I have reverted the last one, and put your suggestion in, and it seems to work okay. I see how it gets it now: `options` at for caller of save() becomes  `params` inside, which becomes `self.encoderinfo`, which becomes `im.encoderinfo` in the plugins.

PIL/ImageShow.py
HinTak: That's reasonably easy to test - I can just copy the bmp files to a windows machine and see. I'll try that sometimes.

HinTak:
I have tried putting `compress_level` at `Image._dump()` and also at its caller in `ImageShow.save_image()`, and both of them work. I don't think I can push back further as it gets messy whether the options should be processed by save_image() or show() . I think either of them make sense (if you think of `_dump` being a quicker save). That said, `compress_level` is understood only my the png backend, so in a way one has to put a backend-specific knowledge somewhere, and philosophically it isn't better than the secret handshake I initially came up with, those it the format is neater.

PIL/Image.py
```diff
@@ -588,11 +588,11 @@ def _dump(self, file=None, format=None):
 
         self.load()
         if not format or format == "PPM":
-            self.im.save_ppm(file)
+            self.im.save_ppm(file, **options)
```
wiredfool: Sorry, I was wrong here, options aren't actually accepted for the save_ppm internal method.

wiredfool: And looking at it, _dump is potentially used in a few places that aren't debugging related, and that helper is only compiled in when WITH_DEBUG is defined.

PIL/ImageShow.py
```diff
@@ -87,7 +87,7 @@ def get_command(self, file, **options):
 
     def save_image(self, image):
         """Save to temporary file, and return filename"""
-        return image._dump(format=self.get_format(image))
+        return image._dump(format=self.get_format(image), compress_level=1)
```
wiredfool: That could be a class varaible:
```
format = 'BMP'
save_options = {}

# or
format = 'PNG'
save_options = {'compress_level', 1}
```

PIL/Image.py
HinTak: I doubt that. Both production (as installed by distros) and dev code uses `_dump`. Besides, debug code uses an entirely different code path is confusing.

wiredfool: @HinTak see #2522

PIL/ImageShow.py
HinTak: Look - I don't want to re-engineer a "perfect" solution and rewrite the whole thing. There was a problem: down-convert strip transparency and mid-displays. The compress_level thing is just a nice addition: if it would works, good, but display slowly and correctly is still better than fast and wrongly.

What I mean is that if the compress_level change takes too much effect/discussion, I 'd say just do without it. So this series is logically three changes, use transparency-capable intermediate format, do not down-convert, and make transparency-format works well. If the 3Rd cannot be done, or takes too much discussion, I 'd say just live without it.

PIL/Image.py
HinTak: Okay. Well, that's unfortunate. But again, I am not volunteering to rewrite a perfect replacement. There is a visual problem I am trying to fix.

HinTak:
FWIW, I did find that `save_ppm` exists is a bit odd, when one can do `save(format='PPM')` . presumably it is just historical. `save_ppm` happens first, and nobody bother to rewrite. I am not volunteering to rewrite either. I'd just like to focus on merging a fix that will fix the transparecy mis-display problem,  even if the solution is slow. Slow correctness is better than fast wrongness.

HinTak:
you decide. either adding **option to save_ppm, or not called it with options. I don't mind either way, but I do not want to be drawn into a discussion on how to rewrite this code "better".

HinTak:
I put the alpha-enabled bmp onto a web site and download it onto one of the local commnity library public computers runing windows 7. They boot off a network image every morning and get wiped on powered down, and don't have much beyond basics, and users don't have any rights installing software or changng settings.

I think win7's default is 'Windows Photo Viewer'. I was also offered 3 others for 'open with': MS Paint, Office 2010 (Office Picture Manager), and Opera Browsers. I also found MS IE 11 and firefox can display bmp. I also tried inseting into a new word document; Windows Explorer can show thumbnails. So that's 8 applications I tried on windows 7, all of them display it without alppha (i.e. wrongly), but none of them complain about not being able to open it.

So it is definitely no-harm doing RGBA bmp.

As for the compress_level issue - to be honest, I offered 3 solutions already. The first wasn't very good; my favourite is the 2nd, a one-line change, but I guess you like the 3rd better. Any of them does the job. So I'll say just merge/cherry--pick any of them. If you feel like you want to rewrite the relevant parts of PIL, please feel free to do so, without me.

wiredfool:
I'm not clear on why you've gotten hostile. I'm not asking for anything. My earlier code that passed **options to save_ppm is going to fail in some cases. I've pointed out for future reference that there's a C debugging flag that enables functionality that's not used in a debugging context. This is a general 'We should clean this up' note -- we've got lots of this because it's old code that has accreted, rather than designed. 

But as you say, I can cherry-pick.

HinTak:
As I mentioned a few times, I prefer not to be involved and hear about future plans on rewrite and clean-ups. Please stay on topic - incoprate the changes (or modification therof you deem appropriate) and close this pull. Thank you.

wiredfool:
As I said, I don't understand your hostility. You are free to turn off notifications and not respond.

HinTak:
This is a pull request - negative comments are treated as "please do more work until it can be merged'. Hence your talking-to-yourself on futures plans on rewrite and clean-ups are inappropriate and quite unwelcomed.

In any case, you "improvement" had broken your own compress_level=1 suggestion. So I'll need to open another bug report. (note I'll switch off notification for "bug reports", vs pull requests).

PR #2517, opened 2017-05-10
RFC: Specify Version in one place

wiredfool:

 * Improves the release process by having one place to change the version number for Pillow, then including that value elsewhere, when required. 
 * Does not update the appveyor version number. I'm wondering if just a build is ok there.

### DIFF
--- a/PIL/__init__.py
+++ b/PIL/__init__.py
@@ -11,8 +11,10 @@
 
 # ;-)
 
-VERSION = '1.1.7'  # PIL version
-PILLOW_VERSION = '4.2.0.dev0'  # Pillow
+from . import version
+
+
+PILLOW_VERSION =  version.__version__
 
 __version__ = PILLOW_VERSION
 
--- a/PIL/__init__.py
+++ b/PIL/__init__.py
@@ -13,7 +13,7 @@
 
 from . import version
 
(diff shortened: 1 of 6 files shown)


hugovk:
It's nice to have the version number in AppVeyor, it makes things 100% clear what's building, but it's not essential: we don't have it on Travis CI.

PIL/__init__.py
```diff
@@ -11,8 +11,10 @@
 
 # ;-)
 
-VERSION = '1.1.7'  # PIL version
```
hugovk: So we're ditching the old PIL version.

The build fails because PIL/Image.py still tries to import `VERSION`, but doesn't look like Image.py or anything that import it actually needs it.

wiredfool: That was an inadvertent change.

hugovk:
Windows isn't happy.

wiredfool:
Yeah, I know. I've got to fire up the windows box and have a look.

wiredfool:
Apparently codecov really wants a comment on this PR. 

Anyway, I've got it building on all the CI versions. Removed the version from Appveyor.yml. Works on clang/osx as well, so major platforms are all covered.

PR #2520, opened 2017-05-11
Features detection

wiredfool:
Fixes detection of webp transparency, adds detection of webpmux, and splits out the modules from features.  Also actually tests the results.

This should now agree with the similar items that are printed during setup. 

Edit:

* python-tk is now installed on travis, since apparently the trusty images don't have it. 
* Keying off of PIL._tkinter_finder to discover if tkinter is available. 
* This is a stepping stone to the complex font PR: #2284

### DIFF
--- a/.travis/install.sh
+++ b/.travis/install.sh
@@ -3,7 +3,7 @@
 set -e
 
 sudo apt-get update
-sudo apt-get -qq install libfreetype6-dev liblcms2-dev\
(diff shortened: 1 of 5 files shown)


wiredfool:
Ok, I'm not sure what's going on with tkinter -- It's working on my versions of 2.7 w/system site packages, and the verbose error is not helpful.

Issue #2514, opened 2017-05-10
Wrong error handling in C Extensions (_webp.c)


adw1n:
### What versions of Pillow are you using?
Pillow version 4.2.0.dev0 -  branch `master` from 10min ago (commit 1e0fbc0f1a524413e6fe77d0cb8de6b1be781302).

### Issue
You made a mistake in the way you handle exceptions returned by basic functions like `PyArg_ParseTuple`. 
Examples:
https://github.com/python-pillow/Pillow/blob/master/_webp.c#L32
https://github.com/python-pillow/Pillow/blob/master/_webp.c#L148
Instead of doing:
```C
return NULL;
```
you are doing:
```C
Py_RETURN_NONE;
```
Why does it matter?
a) error may not be signalized to the user
In python >=3.5 error will be signalized because of https://bugs.python.org/issue23571 but instead of `TypeError`,  `SystemError` will be thrown.
example:
``` python
import PIL._webp
PIL._webp.WebPDecode()
TypeError: function takes exactly 1 argument (0 given)

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
SystemError: <built-in function WebPDecode> returned a result with an error set
```
expected (correct) behaviour would be:
```python
import PIL._webp
PIL._webp.WebPDecode()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: function takes exactly 1 argument (0 given)
```

For example this code in python 2.7.10 and 3.4.3 does not raise an error at all!
```
import PIL._webp
PIL._webp.WebPEncode()
```
b) error does not get cleared
Python 2.7.10 or 3.4.3
```python
>>> import PIL._webp
>>> a=PIL._webp.WebPDecode()
>>> type(a)
TypeError: function takes exactly 1 argument (0 given)
>>> type(a)
<type 'NoneType'>
```

Unit test:
```python
import unittest
import PIL._webp


class TestWebpErrorHandling(unittest.TestCase):
    def test_WebPEncode_invalid_num_of_args_raises_type_error(self):
        with self.assertRaises(TypeError):
            PIL._webp.WebPEncode()
    def test_WebPDecode_invalid_num_of_args_raises_type_error(self):
        with self.assertRaises(TypeError):
            PIL._webp.WebPDecode()


if __name__ == '__main__':
    unittest.main()
```

After running the test using Python 3.4.3:
```
EE
======================================================================
[...]

PR #2521, opened 2017-05-11
Fix for issue #2514 - error handling mistake in _webp.c

adw1n:
This fixes (issue #2514) error handling for when functions are called with an invalid number of arguments or invalid types of arguments.
As for the tests, I've used the test provided in the issue.

### DIFF
--- a/Tests/test_file_webp.py
+++ b/Tests/test_file_webp.py
@@ -77,6 +77,11 @@ class TestFileWebp(PillowTestCase):
 
         self.assertRaises(IOError, lambda: hopper("L").save(temp_file))
 
+    def test_WebPEncode_with_invalid_args(self):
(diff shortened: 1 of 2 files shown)


hugovk:
Looks like you forgot to commit the tests.

adw1n:
I've left them out, because I felt like using them would be like testing the CPython itself. Under normal circumstances you wouldn't probably write that test. Maybe I shouldn't have mentioned it.

Granted, before the commit the test included in the issue would show errors and now it passes.

wiredfool:
It's more like testing that _webp is conforming to the expectations of the CPython api. (as evidenced by the change in Pillow changing the status of the test)

adw1n:
OK. I added the tests that I had.

PR #2525, opened 2017-05-12
Update ImageFile.py

joshblum:

 * Fixes UnboundLocalError in ImageFile.py

```
  [...]
  File "/usr/local/lib/python3.5/dist-packages/PIL/ImageFile.py", line 387, in feed
    fp.close()  # explicitly close the virtual file
UnboundLocalError: local variable 'fp' referenced before assignment
```

### DIFF
--- a/PIL/ImageFile.py
+++ b/PIL/ImageFile.py
@@ -380,11 +380,8 @@ class Parser(object):
 
             # attempt to open this file
             try:
-                try:
-                    fp = io.BytesIO(self.data)
+                with io.BytesIO(self.data) as fp:
                     im = Image.open(fp)
-                finally:
-                    fp.close()  # explicitly close the virtual file
             except IOError:
                 # traceback.print_exc()
                 pass  # not enough data
@@ -432,12 +429,11 @@ class Parser(object):
         if self.data:
             # incremental parsing not possible; reopen the file
             # not that we have all data
-            try:
-                fp = io.BytesIO(self.data)
-                self.image = Image.open(fp)
-            finally:
-                self.image.load()
-                fp.close()  # explicitly close the virtual file
+            with io.BytesIO(self.data) as fp:
+                try:
+                    self.image = Image.open(fp)
+                finally:
+                    self.image.load()
         return self.image
 
 
--- a/Tests/test_imagefile.py
+++ b/Tests/test_imagefile.py
@@ -95,6 +95,11 @@ class TestImageFile(PillowTestCase):
     def test_raise_ioerror(self):
         self.assertRaises(IOError, lambda: ImageFile.raise_ioerror(1))
 
+    def test_raise_typeerror(self):
+        with self.assertRaises(TypeError):
+            parser = ImageFile.Parser()
+            parser.feed(1)
+
     def test_truncated_with_errors(self):
         if "zip_encoder" not in codecs:
             self.skipTest("PNG (zlib) encoder not available")


radarhere:
Are you able to provide a test to reproduce the problem?

joshblum:
I'm not sure a test is appropriate. The `UnboundLocalError` is only triggered when the file is not able to be opened, so an exception is still thrown, but now the stack trace is clean since PIL does not throw an additional exception.

Before:
```
In [1]: from PIL import ImageFile

In [2]: p = ImageFile.Parser()

In [3]: p.feed('')
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
/usr/local/lib/python3.5/dist-packages/PIL/ImageFile.py in feed(self, data)
    383                 try:
--> 384                     fp = io.BytesIO(self.data)
    385                     im = Image.open(fp)

TypeError: a bytes-like object is required, not 'str'

UnboundLocalError                         Traceback (most recent call last)
<ipython-input-3-01b100519976> in <module>()
----> 1 p.feed('')

/usr/local/lib/python3.5/dist-packages/PIL/ImageFile.py in feed(self, data)
    385                     im = Image.open(fp)
    386                 finally:
--> 387                     fp.close()  # explicitly close the virtual file
    388             except IOError:
    389                 # traceback.print_exc()

UnboundLocalError: local variable 'fp' referenced before assignment
```

After:
```
In [1]: from PIL import ImageFile

In [2]:  p = ImageFile.Parser()

In [3]: p.feed('')
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-3-01b100519976> in <module>()
----> 1 p.feed('')

/usr/local/lib/python3.5/dist-packages/PIL/ImageFile.py in feed(self, data)
    381             # attempt to open this file
    382             try:
--> 383                 with io.BytesIO(self.data) as fp:
    384                     im = Image.open(fp)
    385             except IOError:

TypeError: a bytes-like object is required, not 'str'
```

hugovk:
The change means the user gets `TypeError` rather than `UnboundLocalError`, which makes more sense.

A test would guard against regressions.

---

As it happens, this isn't a problem in Python 2.7:
```python
Python 2.7.13 (default, Dec 18 2016, 07:03:39)
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.42.1)] on darwin
>>> from PIL import ImageFile
>>> p = ImageFile.Parser()
>>> p.feed('')
>>> p
<PIL.ImageFile.Parser object at 0x1090dc510>
>>>
```

```python
Python 3.6.1 (default, Apr  4 2017, 09:40:21)
[GCC 4.2.1 Compatible Apple LLVM 8.1.0 (clang-802.0.38)] on darwin
>>> from PIL import ImageFile
>>> p = ImageFile.Parser()
>>> p.feed('')
  File "/private/tmp/py3/lib/python3.6/site-packages/PIL/ImageFile.py", line 384, in feed
    fp = io.BytesIO(self.data)
TypeError: a bytes-like object is required, not 'str'

  File "/private/tmp/py3/lib/python3.6/site-packages/PIL/ImageFile.py", line 387, in feed
    fp.close()  # explicitly close the virtual file
UnboundLocalError: local variable 'fp' referenced before assignment
>>>
```

joshblum:
I believe the error would also happen in python 2.7 if someting was fed to `io.BytesIO` that threw  `TypeError` like `p.feed(1)`. Will add a test to guard against regressions.

joshblum:
@hugovk any other changes you would like made or addition to the tests?

hugovk:
How about also a test for `p.feed('')`? Can skip for Python 2.7 with something like [this](https://github.com/python-pillow/Pillow/blob/310adf087fc5a1e1cc3b6cce9d74a4e58ca8cb88/Tests/test_image.py#L54).

joshblum:
@hugovk i don't think it's necessary since throwing the `TypeError` triggers the `UnboundLocalError` which is what we want to test against. Is that ok?

joshblum:
@hugovk what do you think? :)

hugovk:
@joshblum Merged, thanks!

joshblum:
Thanks!!

PR #2528, opened 2017-05-13
Use Codecov for coverage

hugovk:
Because Coveralls has sometimes been a bit iffy and they rarely respond to support requests (eg. ["Changes Unknown when pulling X on branch into ** on master**"](https://github.com/lemurheavy/coveralls-public/issues/351) / ["First build on master"](https://github.com/lemurheavy/coveralls-public/issues/610)), I'd like to trial using Codecov for a while.

For example: https://codecov.io/gh/python-pillow/Pillow

We could use both Coveralls and Codecov in parallel for a while, to make sure they agree, and fine tune Codecov's notifications, and then decide which one to keep.

Here's some example PRs:
* https://github.com/hugovk/Pillow/pull/17: increase coverage
* https://github.com/hugovk/Pillow/pull/18: decrease coverage
* https://github.com/hugovk/Pillow/pull/19: increase and decrease

Thoughts?

### DIFF
--- a/.travis/after_success.sh
+++ b/.travis/after_success.sh
@@ -10,8 +10,10 @@ gem install coveralls-lcov
 coveralls-lcov -v -n coverage.filtered.info > coverage.c.json
 
 coverage report
+pip install codecov
 pip install coveralls-merge
 coveralls-merge coverage.c.json
+codecov
 
 if [ "$DOCKER" == "" ]; then
 	pip install pep8 pyflakes
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,1 @@
+comment: off


wiredfool:
Fine with me, as long as it doesn't spam each PR thread.

hugovk:
[By default](https://docs.codecov.io/v4.3.6/docs/pull-request-comments) it only adds one comment and edits it on changes, but I've added a config file to disable commenting entirely.

Instead we can enable the GitHub status API, like we have for Travis, AppVeyor and Coveralls. That may need a webhook enabling at https://docs.codecov.io/v4.3.6/docs/commit-status by someone with more repo permissions than me.

wiredfool:
It looks like the github integration is for github enterprise only. (https://github.com/integrations/codecov ) I'm not seeing any other way to get the status back into github, other than the comments.

hugovk:
Let's just merge and see how it works on new PRs, and I'll get in touch with them if it's not working.

hugovk:
@wiredfool I checked with Codecov, and the GitHub integration isn't just for GitHub Enterprise. Please could you install it? https://github.com/integration/codecov

wiredfool:
I've enabled it, we'll see how it goes.

PR #2531, opened 2017-05-15
Fixed Tox URL

radarhere:


hugovk:
(I'm going to blame this Codecov spam on the fact that codecov.yml wasn't merged (https://github.com/python-pillow/Pillow/pull/2528) when the build started, but was when it ended. But let's see the next PR.)

PR #2539, opened 2017-05-23
Remove unused functions

hugovk:
Remove some unused C functions (added in 1996, 2006 and 2000) which are not called by any other C or Python code.

Helps a bit towards #722.

### DIFF
--- a/outline.c
+++ b/outline.c
@@ -135,25 +135,11 @@ _outline_close(OutlineObject* self, PyObject* args)
     return Py_None;
 }
 
-static PyObject*
-_outline_transform(OutlineObject* self, PyObject* args)
-{
-    double a[6];
-    if (!PyArg_ParseTuple(args, "(dddddd)", a+0, a+1, a+2, a+3, a+4, a+5))
-        return NULL;
-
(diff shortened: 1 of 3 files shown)


outline.c
```diff
@@ -135,25 +135,11 @@ _outline_close(OutlineObject* self, PyObject* args)
...
 }
 
-static PyObject*
-_outline_transform(OutlineObject* self, PyObject* args)
-{
-    double a[6];
-    if (!PyArg_ParseTuple(args, "(dddddd)", a+0, a+1, a+2, a+3, a+4, a+5))
-        return NULL;
-
-    ImagingOutlineTransform(self->outline, a);
```
homm: Is it also not used anymore? Should we instead add public API for `outline.transform`?

hugovk: I don't think it's ever been used. I've added public APIs for some other functions, but I'm not sure what's expected of this one, so be my guest! :)

wiredfool:
These are actually exposed in the python object layer, so it's possible that these are called from elsewhere. `new_array` is probably safe to drop, The other two I'm less sure of, as they appear to provide some useful functionality.

Issue #2491, opened 2017-04-19
Check ImageFile.py decoder.cleanup on load


wiredfool:
Suspect that if we get an `IndexError` or `struct.error` here, we leak the decoder:

```python
                    while True:
                        try:
                            s = read(self.decodermaxblock)
                        except (IndexError, struct.error):  # truncated png/gif
                            if LOAD_TRUNCATED_IMAGES:
                                break
                            else:
                                raise IOError("image file is truncated")

                        if not s:  # truncated jpeg
                            self.tile = []

                            # JpegDecode needs to clean things up here either way
                            # If we don't destroy the decompressor,
                            # we have a memory leak.
                            decoder.cleanup()

                            if LOAD_TRUNCATED_IMAGES:
                                break
                            else:
                                raise IOError("image file is truncated "
                                              "(%d bytes not processed)" % len(b))

                        b = b + s
                        n, err_code = decoder.decode(b)
                        if n < 0:
                            break
                        b = b[n:]

                # Need to cleanup here to prevent leaks in PyPy
                decoder.cleanup()
```

radarhere:
I think this has been resolved by https://github.com/python-pillow/Pillow/pull/3023/commits/add2746ac61980421455dc722d0edede2478cbfa

PR #2541, opened 2017-05-24
Fix truncated png loading

homm:
This PR fixes two problems:

### PngImagePlugin

Pillow can't open some truncated PNG images even with `ImageFile.LOAD_TRUNCATED_IMAGES = True`.

`SyntaxError: broken PNG file (chunk '\x00\x00\x00\x00')`

Example image: https://ucarecdn.com/99b9dc77-abd0-4077-887b-2b7a0a753d76/587593.png

### ImagingZipDecodeCleanup

This is general memory leak while loading truncated PNG files regardless of LOAD_TRUNCATED_IMAGES flag. The reason is `inflate` operates normally if source stream is truncated and returns no errors. For `ImagingZipDecode` function this look like normal behavior. But on the python level we see than the file if over and we should call `inflateEnd` and free `context->previous` somehow. This PR adds `ImagingZipDecodeCleanup` function for this.

### DIFF
--- a/libImaging/ZipDecode.c
+++ b/libImaging/ZipDecode.c
@@ -281,7 +281,7 @@ ImagingZipDecode(Imaging im, ImagingCodecState state, UINT8* buf, int bytes)
 
 
 int ImagingZipDecodeCleanup(ImagingCodecState state){
-    /* called to fee the decompression engine when the decode terminates
+    /* called to free the decompression engine when the decode terminates
        due to a corrupt or truncated image
     */
     ZIPSTATE* context = (ZIPSTATE*) state->context;
--- a/libImaging/ZipDecode.c
+++ b/libImaging/ZipDecode.c
@@ -85,6 +85,7 @@ ImagingZipDecode(Imaging im, ImagingCodecState state, UINT8* buf, int bytes)
         err = inflateInit(&context->z_stream);
         if (err < 0) {
             state->errcode = IMAGING_CODEC_CONFIG;
+            free(context->previous);
             return -1;
         }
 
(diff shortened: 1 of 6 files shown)


homm:
The example:

I've prepared the test file. It is just 128k truncated PNG file
https://ucarecdn.com/c7dd476a-cb01-41f2-a93b-46cdc2fe7c06/

```python
$ ipython

In [1]: from PIL import ImageFile, Image; ImageFile.LOAD_TRUNCATED_IMAGES = True
In [2]: %timeit -n 2000 im = Image.open('trunc'); im.load()
2000 loops, best of 3: 13.7 ms per loop
```

The process memory constantly grows during the test up to 390MB. With the fix the memory consumption remains 60Mb. This gives memory leak 330Mb / 6000runs * 1024 ≈ 56Kb per image.

libImaging/ZipDecode.c
```diff
@@ -274,4 +279,20 @@ ImagingZipDecode(Imaging im, ImagingCodecState state, UINT8* buf, int bytes)
 
 }
 
+
+int ImagingZipDecodeCleanup(ImagingCodecState state){
+    /* called to fee the decompression engine when the decode terminates
```
radarhere: free?

homm: Thanks. That was copypaste from JpegDecoder.c

radarhere: Hmm. I went ahead and created #2542 before realising you'd bundled it into this PR. I'll leave the PR open and see what happens, if you don't mind, just because it's an independent change.

wiredfool:
Does #2491 interact with this?  I think the truncated PNG might be a way to make that one actually trigger.

homm:
[Ошибка в роботах](https://trello.com/c/JWJX20gz/44--)

homm:
> Does #2491 interact with this?

Not necessarily. In this cases `decoder.cleanup()` is called, but ZIP image handler was empty. I'm not sure how we can add tests for the memory leak: it outside of Python code. I'll try to add small affected image for the loading bug. The one in the description is too big (2.5 Mb).

wiredfool:
I think there's a test case that checks RSS.  Looks like check_imaging_leaks and check_webp_leaks.

PR #2544, opened 2017-05-26
Freebsd compat

melvyn-sopacua:
When openjpeg 1.5 is installed (Calligra 2.x, for example) and OpenJPEG
2.1 is installed, the subdirectory is added twice.

Since the 2nd time is ignored, it is ranked after the more general
/usr/local/include and openjpeg.h from the 1.5 version is picked up.

Fix this in a more general way:
- If a directory is added that already is in path
- But where is specified

Then remove the subdir from path and insert it at the spot requested.

A FIXME is added to the code bit explaining the real issue, but lack
investigative spirit / time to dive down the rabbit hole.

 * See above

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -63,6 +63,9 @@ def _add_directory(path, subdir, where=None):
         else:
             _dbg('Inserting path %s', subdir)
             path.insert(where, subdir)
+    elif subdir in path and where is not None:
+        path.remove(subdir)
+        path.insert(where, subdir)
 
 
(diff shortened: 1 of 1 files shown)


wiredfool:
I've checked this on FreeBSD and it works OMM, I've also checked that it doesn't mess up on Ubuntu 14.04 with the OpenJpeg 1.5 version installed (but, we didn't error out there in Master).

PR #2548, opened 2017-05-27
Remove unused function

hugovk:
Remove an unused C function.

Replaces and closes #2539.

Helps a little bit towards #722.


wiredfool:
#2567 Release notes -- removing internal but accessible function.

PR #2553, opened 2017-05-29
More unit tests for ImageMorph

hugovk:
Changes proposed in this pull request:

 * More unit tests
 * Fix a bug found by new tests
 * Helps towards #722

PR #2554, opened 2017-05-29
More unit tests for ImageMorph

hugovk:
Same as PR https://github.com/python-pillow/Pillow/pull/2553 but includes a failing roundtrip test.

 * More unit tests
 * Fix a bug found by new tests
 * Helps towards #722

---

The test in https://github.com/python-pillow/Pillow/commit/42da0a56731b91311b4791d1dc603b991064c2d5 is [failing](https://travis-ci.org/python-pillow/Pillow/jobs/237061871#L3179) because [ImageMorph.py expects a length of 8192](https://github.com/python-pillow/Pillow/blob/master/PIL/ImageMorph.py#L237-L239) but it's really 30605:

```python
        if len(self.lut) != 8192:
            self.lut = None
            raise Exception('Wrong size operator file!')
```

@dov Is there a reason for this magic number, or is there a problem with the unit test? Thanks!


hugovk:
Rebased on master, which includes the fix in https://github.com/python-pillow/Pillow/issues/2561, and now passes.

PR #2549, opened 2017-05-27
Remove deprecated code

hugovk:
And document it in release notes.

Also remove some old docs about deprecated code that's already been removed.

PR #2555, opened 2017-05-29
Changed warning to error when saving RGBA images as JPEG

radarhere:
As stated in the release notes - https://github.com/python-pillow/Pillow/blob/master/docs/releasenotes/3.4.0.rst - saving RGBA images as JPEG 'will become an error in Pillow 4.2'.


hugovk:
This is also part of PR https://github.com/python-pillow/Pillow/pull/2549.

radarhere:
Ah, yes, thanks for pointing that out.

PR #2558, opened 2017-05-29
Replace the truncating of zero coefficients with math power

homm:
The Initial idea: while resizing we are excluding zero coefficients on the edges. For example, if `xmin = 103` and `xmax = 103 + 7` and we have coefficients `[0, -0.15, -0.3, 0,5, 0.53, -0.34, 0]`, it is equivalent of `xmin = 104`, `xmax = 104 + 5`, `[-0.15, -0.3, 0,5, 0.53, -0.34]`.

Currently, the following code is responsible for this:
https://github.com/python-pillow/Pillow/blob/5a359fbf287d6962def2d0d482d2570625c90bf6/libImaging/Resample.c#L183-L195

But it is possible mathematically prove, that we can just do not calculate these coefficients, instead of excluding them.

```
We need to find:
w = filterp->filter((x + xmin - center + 0.5) * ss)
, where xmin = (int) floor(center - support)

let center - support == n + λ, where 0.5 <= λ < 1
, then xmin == floor(center - support) == n
, then xmin - center + 0.5 == n - n - λ - support + 0.5 == 0.5 - λ - support
, where -0.5 < 0.5 - λ <= 0

if support == filterp->support * filterscale
, and ss == 1.0 / filterscale
, then ss == filterp->support / support

(xmin - center + 0.5) * ss == (0.5 - λ - support) * (filterp->support / support)
== (0.5 - λ) * filterp->support / support - filterp->support

(0.5 - λ) * filterp->support / support will always be less than 0
```

We know what each value of `filterp->filter` for parameters less than `- filterp->support` will be 0.

So we shouldn't consider `0.5 <= λ < 1`. `xmin  = floor(center - support)` can be safely replaced by `xmin  = round(center - support)`. Same for `xmax`.

### DIFF
--- a/Tests/test_image_resample.py
+++ b/Tests/test_image_resample.py
@@ -158,14 +158,12 @@ class TestImagingCoreResampleAccuracy(PillowTestCase):
 
     def test_enlarge_hamming(self):
         for mode in ['RGBX', 'RGB', 'La', 'L']:
-            case = self.make_case(mode, (4, 4), 0xe1)
-            case = case.resize((8, 8), Image.HAMMING)
-            data = ('e1 e1 ea d1'
-                    'e1 e1 ea d1'
-                    'ea ea f4 d9'
-                    'd1 d1 d9 c4')
+            case = self.make_case(mode, (2, 2), 0xe1)
+            case = case.resize((4, 4), Image.HAMMING)
+            data = ('e1 d2'
+                    'd2 c5')
             for channel in case.split():
-                self.check_case(channel, self.make_sample(data, (8, 8)))
+                self.check_case(channel, self.make_sample(data, (4, 4)))
 
     def test_enlarge_bicubic(self):
         for mode in ['RGBX', 'RGB', 'La', 'L']:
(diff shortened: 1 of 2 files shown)


homm:
Working on this PR I found the problem with HAMMING filter. Please, look at the graph:

screen shot 2017-05-29 at 14 12 50

It is obviously, that function not ends after [-1:1]. But the filter window should be strictly limited in that range. I've added the check and fixed the test.

wiredfool:
Ok, to summarize:

`filterp->support` is the window size of the filter, so by definition, filterp->filter(y) of anything outside the range of `[-filterp->support, filterp->support] is zero. 

The expression for y at x=0 can be reduced to `[negative] - filterp->support`, when 0.5<λ<=1, therefore we're outside the filter window.  If λ <0.5, then round(n+λ) is equivalent to floor(n+λ).

So, we can use round() instead of floor() and the excluded items will always be zero coefficients.

Correct?

homm:
> the excluded items will always be zero coefficients.

Exactly

PR #2560, opened 2017-05-30
Minor test changes

radarhere:
* Replace deprecated assert_ with assertTrue
 * Changed a file opening to use with

### DIFF
--- a/Tests/test_image.py
+++ b/Tests/test_image.py
@@ -390,7 +390,7 @@ class TestRegistry(PillowTestCase):
     def test_encode_registry(self):
 
         Image.register_encoder('MOCK', mock_encode)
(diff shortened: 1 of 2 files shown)


Tests/test_image.py
```diff
@@ -390,7 +390,7 @@ class TestRegistry(PillowTestCase):
     def test_encode_registry(self):
 
         Image.register_encoder('MOCK', mock_encode)
-        self.assert_('MOCK' in Image.ENCODERS)
+        self.assertTrue('MOCK' in Image.ENCODERS)
```
hugovk: Or assertIn?

radarhere: Yes, you're right thanks.

PR #2561, opened 2017-05-30
Fixed wrong expected size of mrls read from disk.

dov:
Fixes the failure of  the mrl round trip test in https://github.com/python-pillow/Pillow/pull/2554 .

 * 
 * 
 *


hugovk:
Thanks, I added this commit to PR https://github.com/python-pillow/Pillow/pull/2554 and it passed.

PR #2568, opened 2017-06-04
Removed deprecated universal newlines mode from Python 3 code

radarhere:
Universal newlines mode is deprecated, according to https://docs.python.org/3.2/library/functions.html#open


wiredfool:
See also #2187

radarhere:
Thanks. I'll close this then.

Issue #273, opened 2013-07-08
16-bit TIFF files cannot be converted to array


tdsmith:
Example data file: https://dl.dropboxusercontent.com/u/3720/junk/12bit.tif (a 16-bit TIFF container with pixel values spanning a 12-bit range, which is common in scientific imaging)

Opening the file with Image.open works:

```
>>> im, im.mode
(<PIL.TiffImagePlugin.TiffImageFile image mode=I;16 size=1344x1024 at 0x1074A2518>, 'I;16')
```

Attempting to convert the Image object to an array just encapsulates it:

```
>>> np.array(im)
array(<PIL.TiffImagePlugin.TiffImageFile image mode=I;16 size=1344x1024 at 0x1074A2518>, dtype=object)
```

This is a 1x1 array containing an Image object. I was expecting to have a 1344x1024 array containing 16-bit integer values.

Issue #861, opened 2014-08-20
Can't create I;16S TIFF images


xan-der:
Moving to a more standard install of MapProxy, using Pillow.  Pillow can load 16 bit grayscale TIFF images, but can't create them.   Similar bug was in PIL.  Steps to reproduce:

```python
# python
Python 2.6.6 (r266:84292, Jan 22 2014, 09:42:36) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-4)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from PIL import Image;
>>> filename = "GMTED.tiff";
>>> img = Image.open(filename);
>>> img.mode
'I;16S'
>>> img.size
(256, 256)
>>> img1 = Image.core.new(img.mode, img.size);
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: unrecognized mode
>>> 
```

In the old PIL code, the changes below would fix the problem.  I will need to look at the Pillow code to see if it is similar enough to make the same mods.

```
  Imaging.h - increase buffer for mode to avoid overrun in Storage.c
    char mode[6+1]; /* Band names ("1", "L", "P", "RGB", "RGBA", "CMYK", "YCbCr", "BGR;xy") */
  Access.c - add I;16S to list and update hash code to avoid collision with new value
    #define ACCESS_TABLE_SIZE 25
    #define ACCESS_TABLE_HASH 5543
    ADD("I;16S", line_16, get_pixel_16L, put_pixel_16L);
  Pack.c - add I;16S to storage modes using "copy2"
    {"I;16S",   "I;16S",      16,     copy2},
  Storage.c - change multiple strcmp(mode, ...) to single strncmp(mode, "I;16", 4) so all cases are covered
    } if (strncmp(mode, "I;16", 4) == 0) {
  Unpack.c - add I;16S to the "storage modes" part of the unpackers array
    {"I;16S",   "I;16S",      16,     copy2},
  Convert.c - Add converter references
    { "I", "I;16S", I_I16L },
    { "I;16S", "I", I16L_I },
    { "L", "I;16S", L_I16L },
    { "I;16S", "L", I16L_L },
  ImageMode.py - Add I;16S to mapping modes
        _modes["I;16S"] = ModeDescriptor("I;16S", "I", "L", "L")
```

Gael-de-Sailly:
Same error with a 6001x6001 16-bit TIFF image from SRTM. What should I do ?

[...]

Issue #2228, opened 2016-11-17
Eliminate experimental modes


homm:
Every time when I open Pillow's C sources, I absolutely scared by all these experimental modes. Do I need to worry how not to break them? Do I need to implement features for all of them? How can I write tests for them?

In my opinion, an experiment must have the following characteristics: hypothesis; subject; deadline or exit condition; metric; outcomes. When someone adds another `else if` in `ImagingNewPrologueSubtype`, this is not an experiment. This is just another unfinished feature which someone else needs to support.

My suggestions are:

* Remove following modes with `IMAGING_TYPE_SPECIAL` type: `I;16`, `I;16L`, `I;16B`, `I;16N`, `BGR;15`, `BGR;16`, `BGR;24`, `BGR;32`
* Remove `RGBX` mode as full equivalent of `RGB`
* Keep widespread `1`, `P`, `L`, `LA`, `RGB` and `RGBA` modes
* Keep alternative linear (or near linear) color spaces `CMYK` and `YCbCr`
* Keep derivatives with multiplicated alpha `La` and `RGBa`
* Apparently, keep wide modes `F` and `I`
* Create full list of the supported modes, monitor correctness of new or modified operations with all of them and include tests

Questionable modes:
* Alternative color spaces with signed channels `LAB` and `HSV`
* Palette with alpha `PA`. As I know palette supports `RGBA` modes, so it can handle "palette with alpha" in a different way

wiredfool:
Experimental is a word, a more accurate word would be legacy. It's released, in the wild, and I'm pretty sure that a lot of it is used.

From what I can tell, there are 4 classes of support for modes:

* Full support - Anything that Pillow can do can be done in these modes. `RGB/RGBA` would be examples here. This mostly covers the `1` mode as well, but it's a little bit special.
* Partial support - There's basic support, we can read, write and store (in memory) images in this format, but there are holes in the coverage for other operations. (e.g. cropping works, but not necessarily resampling). This is the case with the `I;16` modes, and the partial width `RGB/BGR` modes
* Pack/unpack support - These are the raw modes, and they're only useful in transit to one of the storage modes. 
* Unsupported - Pixel formats that we just don't do, like multiband high bit depth and signed I;16.

I totally agree on putting together a list of the supported modes and monitoring the correctness of operations with them. Tests are always good. 

[...]

PR #2577, opened 2017-06-14
Allow access of signed 16-bit TIFF pixel data as numpy arrays.

tomgoddard:
Small additions to access I;16S TIFF pixel data as numpy arrays.
Code to convert to other pixel types such as 32-bit integers is not included.
This pull request doesn't fix all the deficiencies in handling 16-bit signed TIFF data but it is a start.  I use this in a patched Pillow version in a visualization package ChimeraX for light microscopy data.

Partial fix to issue #861

Related to closed issue #273 addressed by merged pull request #347

 * This enables opening a TIFF image containing mode I;16S pixels and seeking in multipage TIFF:

```python
  i = Image.open('test_i16s.tif')
  i.seek(1)
  import numpy
  a = numpy.array(i)
```

 The current Pillow master branch and 4.1.1 release gives an error on the seek and a zero dimensional array when trying to get the pixels.  This pull request gives the correct results.  Here is a multipage I;16S tiff example file.

    http://www.cgl.ucsf.edu/home/goddard/temp/test_i16s.tif

Here is the error using from the seek() call using Pillow 4.1.1

```python
>>> i = Image.open('test_i16s.tif')
>>> i.seek(1)
  File "/usr/local/lib/python3.6/site-packages/PIL/TiffImagePlugin.py", line 950, in seek
    self.im = Image.core.new(self.mode, self.size)
ValueError: unrecognized mode
>>> i.mode
'I;16S'
```

and the incorrect result of the numpy array call in Pillow 4.1.1

```python
>>> a = numpy.array(i)
>>> a
array(<PIL.TiffImagePlugin.TiffImageFile image mode=I;16S size=300x512 at 0x109D3D588>, dtype=object)
>>> a.shape
()

```


hugovk:
@tomgoddard Please can you add unit tests?

wiredfool:
This is discussed in #2228

tomgoddard:
I added conversion code between I;16S and modes I and L and added I;16S to the test_mode_i16.py unit test.

tomgoddard:
I added tests of the I;16S mode a week ago.  Is there more I need to do to move this along?

homm:
In general I against adding modes with `im->type = IMAGING_TYPE_SPECIAL`. I'd prefer to have less fully supported modes than more modes with 1-2 applicable operations. In my opinion Pillow must support all possible external modes and convert them to one of the appropriate internal modes. Also Pillow should allow to export images as bytes or numpy arrays in wide range of modes.

For example, in case of `I;16S` TIFFs, Pillow should load them in `I` mode (which is already signed, I believe) and provide some way to construct numpy array in `I;16S` format. As I see there is a problem with exporting numpy arrays in arbitrary mode, because `Image.__array_interface__` acts based on current mode and there in no way to specify another. Additionally, `I;16S` mode is one of the `I;16` family, which for some reasons, was not added on Storage.c and Convert.c.

I think we should accept this.

libImaging/Convert.c
```diff
@@ -714,6 +725,15 @@ I16B_I(UINT8* out_, const UINT8* in, int xsize)
 }
 
 static void
+I16S_I(UINT8* out_, const UINT8* in, int xsize)
+{
+    int x;
+    INT32* out = (INT32*) out_;
+    for (x = 0; x < xsize; x++, in += 2)
+      *out++ = (INT16)(in[0] + ((int) in[1] << 8));
```
homm: wrong indention

libImaging/Convert.c
```diff
@@ -774,6 +813,19 @@ I16B_L(UINT8* out, const UINT8* in, int xsize)
             *out++ = in[1];
 }
 
+static void
+I16S_L(UINT8* out, const UINT8* in, int xsize)
+{
+    int x;
+    for (x = 0; x < xsize; x++, in += 2)
+        if (in[1] & 0x80)
+	    *out++ = 0;  /* Negative -> 0 */
```
homm: wrong indention

wiredfool:
I've been looking at this and adding more tests, access, and whatnot.

I'm now leaning against this approach. 

Prior to this PR, I think the core issue is in the tiff `OPEN_INFO` specification, specifically this: 
```
    (II, 1, (1,), 1, (16,), ()): ("I;16", "I;16"),
    (MM, 1, (1,), 1, (16,), ()): ("I;16B", "I;16B"),
    (II, 1, (2,), 1, (16,), ()): ("I;16S", "I;16S"),
    (MM, 1, (2,), 1, (16,), ()): ("I;16BS", "I;16BS"),
```

The last to elements are (`mode`, `rawmode`). These modes have been here since The Fork, but we've never had a mode for either of the S modes. We _have_ had rawmodes in `Unpack.c` for these formats to unpack to an `I` mode the whole time. 

Also, mirroring this is the `Image._fromarray_typemap`, which maps _all_ of the `I;16` modes to `I` mode:
```
    ((1, 1), "<u2"): ("I", "I;16"),
    ((1, 1), ">u2"): ("I", "I;16B"),
    ((1, 1), "<i2"): ("I", "I;16S"),
    ((1, 1), ">i2"): ("I", "I;16BS"),
```

Though, we do optimistically offer a bunch of rawmodes in the `Image._MODE_CONV` map to transfer a whole host of rawmodes to numpy. Unfortunately, this is only ever accessed with `self.mode`, which is not a rawmode. 
``` 
    "I;16S": ('<i2', None),
    "I;16BS": ('>i2', None),
    "I;16LS": ('<i2', None),
    "I;32": ('<u4', None),
    "I;32B": ('>u4', None),
    "I;32L": ('<u4', None),
    "I;32S": ('<i4', None),
    "I;32BS": ('>i4', None),
    "I;32LS": ('<i4', None),
```

Also, there's this one other mode, which is broken as well:
```
    (MM, 1, (2,), 1, (32,), ()): ("I;32BS", "I;32BS"),
```

There's got to be a better solution to this than adding a supported mode which doesn't work with any of the Image operations that we do.

Option 1 is just to fix the TiffImagePlugin to not advertise modes we don't support. We could losslessly read an `I;16S` into an `I` image, do any image manipulations on it, and export it to numpy as an int32. If they wish to change the type in numpy, that's an easy operation. 

Alternately, we could do something along the lines of an adapter that provides an `__array_interface__` of a specified rawmode from an Image. Where now, the interface looks like `np.array(img)`, it would be something like: `np.array(img.as_array_mode('I;16S'))`. This would essentially do the image type conversion in Pillow. rather than in numpy.

I'm hesitant to add a lightly supported mode like I;16S that has such a potential for numerical error.

tomgoddard:
For my uses being able to access I;16S TIFF files as int32 arrays would be fine.

Does the comment "We have had rawmodes in Unpack.c for these formats to unpack to an I mode the whole time.” mean there is some way to read I;16S data currently?  The following code using Image.convert(‘I’) on an I;16S TIFF image gives an error in Pillow 4.2.1

path = '/Users/goddard/Downloads/test_i16s.tif'
from PIL import Image
i = Image.open(path).convert(‘I')


  File "test.py", line 3, in <module>
    i = Image.open(path).convert('I')
  File "/Users/goddard/ucsf/chimerax/ChimeraX.app/Contents/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/PIL/Image.py", line 860, in convert
    self.load()
  File "/Users/goddard/ucsf/chimerax/ChimeraX.app/Contents/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/PIL/TiffImagePlugin.py", line 1018, in load
    return super(TiffImageFile, self).load()
  File "/Users/goddard/ucsf/chimerax/ChimeraX.app/Contents/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/PIL/ImageFile.py", line 188, in load
    self.load_prepare()
  File "/Users/goddard/ucsf/chimerax/ChimeraX.app/Contents/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/PIL/ImageFile.py", line 261, in load_prepare
    self.im = Image.core.new(self.mode, self.size)
ValueError: unrecognized mode

wiredfool:
These lines in TiffImagePlugin should be changed to make it work:
```
-    (II, 1, (2,), 1, (16,), ()): ("I;16S", "I;16S"),
-    (MM, 1, (2,), 1, (16,), ()): ("I;16BS", "I;16BS"),
+    (II, 1, (2,), 1, (16,), ()): ("I", "I;16S"),
+    (MM, 1, (2,), 1, (16,), ()): ("I", "I;16BS"),
```

wiredfool:
(I'll note that it's likely completely untested, but I believe that there might be a way to trigger that code in currently shipping Pillow by doing raw conversions from a buffer or bytes.)

tomgoddard:
Thanks!  This two line change solves the problem in my software reading 16-bit signed TIFF microscopy images.  With this change numpy.array(image) produces 32-bit int arrays.  It would be better if 16-bit int arrays were directly produced but 32-bit int is a perfectly workable solution.  The advantage of 16-bit is that these can be giga-byte size multiimage stacks (3-dimensional microscopy) so efficiently reading into a native 16-bit int array is preferred for performance reasons.

wiredfool:
I'm closing this, there's a successor PR (#2743) for it including some other testing, but the other approach was merged (#2748)

PR #2579, opened 2017-06-17
Added Ubuntu 14.04 and 16.04 tested Pillow version

radarhere:
I set up some VMs and started testing.


wiredfool:
Those are covered in the section above, under CI targets. There's a different expectation of coverage on those, as we're verifying it on every PR.

radarhere:
Okay, thanks.

PR #2410, opened 2017-02-17
Crop decompression

wiredfool:
Fixes #2402 .

Changes proposed in this pull request:

 * Add decompression bomb check for image.crop, since it can enlarge images.
 * Refactor out checks to _crop, so that we can apply them to any core image object.
 * Gif disposal is where the bug in #2402 was, it was requesting a 1GP dispose_extents.

Still need to: 
~- [ ] Generate a gif with extra large extents~
 - [x] Consider making decompression bombs an error at some level, as now the gif from #2402 kicks out a warning, then dies.

PR #2583, opened 2017-06-21
RFC: Decompression bomb error

wiredfool:
Noted in #2410.

 *  Added a DecompressionBombError for if the requested size is 2*Image.MAX_PIXELS

Note that this will cause new exceptions, so we should make sure we really want to do this.


wiredfool:
#2832 need release notes.

Issue #1911, opened 2016-05-13
Paste issues


simonschofield:
### What did you do?
I am using pillow image paste to construct large collage images from many small ones. The large target image is initially set to have an overall alpha of 0, so is like an empty transparent Photoshop layer if you like. I am pasting smaller images with various alphas onto this
You can see such images at www.simonschofield.net.
### What did you expect to happen?
I expect that the paste function mimics the Duff Porter OVER composite algorithm
### What actually happened?
I have noticed two things wrong with paste.
1/ If you paste onto a fully transparent area of the target canvas image it seems to blend with a mid-grey creating dark halos around light objects. This is not in accordance with Duff Porter: it should set the colour to that of the incoming pasted pixel, and set the alpha to the same as the incoming pixel.
2/ Multiple applications of Alphas of pasted pixels to the same point in the large canvas image should should always contribute (i.e. always get greater, stopping at 255). I have noticed that pasted images with low alpha seem to lessen the alpha of the existing pixels in the canvas image, making those areas more transparent.

Also, is paste the only way of compositing a small image into a larger one?

### What versions of Pillow and Python are you using?
Pillow is about v3
Python 3.3.3 64bit

Please include code that reproduces the issue and whenever possible, an image that demonstrates the issue. The best reproductions are self-contained scripts with minimal dependencies.

```python
code goes here
```

wiredfool:
You might want to look at `Image.alpha_composite`: http://pillow.readthedocs.io/en/3.2.x/reference/Image.html?highlight=alpha_composite#PIL.Image.alpha_composite

simonschofield:
Thanks, sounds promising. Will do. Is this a new feature, I don't remember seeing it in my in my version?


________________________________

simonschofield:
Ahh but it says the source must have the same size as the target image, I want to paste smaller images into a big image.


________________________________

simonschofield:
Ignore last email. I was looking at blend by mistake


________________________________

simonschofield:
No, alpha composite need two images of the same size. I want to composite image A at point P within image B


________________________________

simonschofield:
The documentation admits my second point

"Where the mask is 0, the current value is preserved. Intermediate values will mix the two images together, including their alpha channels if they have them."

This is not useful in a standard paste operation and should be altered to accumulate the alpha or at least have the choice with a flag


________________________________

wiredfool:
[...]

PR #2595, opened 2017-06-26
RFC: Image.Image.alpha_composite

wiredfool:
See #1911 

 * Adds an alpha_composite instance method on Image that does the composition 'in place', rather than returning a whole new image. 
 * Specifies a bounding box in the destination/background image, and a source point in the overlay.
 * Currently done in python, but this is an initial implementation that could be pushed into the C layer to eliminate copies and crops.

I'm doing this as a test to see if the interface makes sense as a companion of Image.Image.paste, and Image.alpha_composite.

### DIFF
--- a/PIL/Image.py
+++ b/PIL/Image.py
@@ -1373,45 +1373,51 @@ class Image(object):
         else:
             self.im.paste(im, box)
 
-    def alpha_composite(self, im, box=None, source=(0,0)):
+    def alpha_composite(self, im, dest=(0,0), source=(0,0)):
         """ 'In-place' analog of Image.alpha_composite
 
         :param im: image to composite over this one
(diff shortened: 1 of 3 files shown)


PIL/Image.py
```diff
@@ -1373,6 +1373,53 @@ def paste(self, im, box=None, mask=None):
         else:
             self.im.paste(im, box)
 
+    def alpha_composite(self, im, dest=(0,0), source=(0,0)):
+        """ 'In-place' analog of Image.alpha_composite
+
+        :param im: image to composite over this one
+        :param dest: Optional 2 tuple specificing the upper left corner
```
radarhere: specifying?

wiredfool:
@homm, @hugovk  -- + or - on this api?  Parameters make sense? Names appropriate?

Issue #148, opened 2013-03-20
Automatic scaling of Image.MAXBLOCK


acdha:
There's a venerable old PIL bug which caused `Image.save(optimize=True)` to fail with JPEGs when the file size is over a certain threshold. The earliest reference I could find dates back to 1999:

http://mail.python.org/pipermail/image-sig/1999-December/000942.html

The workaround commonly used is to either hardcode something like this:

```
ImageFile.MAXBLOCK = <empirically determined constant (i.e. wild guess)>
```

or the more dynamic approach used in e.g. jdriscoll/django-imagekit#50:

```
ImageFile.MAXBLOCK = max(img.size) ** 2
```

This seems like exactly the kind of thing Pillow could just fix once and for all.

agschwender:
One of the users of Pilbox reported that they were receiving an `encoder error -2 when writing image file` error. I did some digging into it and it turned out it was receiving a `Suspension not allowed here`. Looking at the code it seemed as if it might be caused by a combination of `quality=keep` and `optimize=True`.  Specifically, it looked as if it would not reach the condition that increased the buffer size. 

```
if "optimize" in info or "progressive" in info or "progression" in info:
    if quality >= 95:
        bufsize = 2 * im.size[0] * im.size[1]
    else:
        bufsize = im.size[0] * im.size[1]
```

Here's a reference to the issue: https://github.com/agschwender/pilbox/issues/29

Apreche:
I know this bug is marked as closed, but I have come across a jpg file uploaded by a user that triggered this error. I am using Python 2.7.6 and Pillow 2.8.1. The error also exists in 2.8.0 and 2.7.0. 

Here are steps to reproduce.

1. Open the JPEG file
2. Extract the icc_profile data
3. Thumbnail the jpeg, but don't make the thumbnail much smaller than the original.
4. Save it as a jpeg with optimize=True, the original icc_profile, and quality=95

Expected Result: Save a thumbnailed image
Actual Result: Prints the message "Suspension not allowed here" and raises an "IOError: encoder error -2" 

Here is some code to reproduce the error.

```Python
from PIL import Image
from StringIO import StringIO

i = Image.open('bad.jpg')
icc_profile = i.info.get('icc_profile')
i.thumbnail((127, 275), Image.ANTIALIAS)
output = StringIO()
i.save(output, format='JPEG', quality=95, optimize=True, icc_profile=icc_profile)
```

[...]

PR #2596, opened 2017-06-27
Adjust buffer size when icc_profile > MAXBLOCK (issue #148)

Darou:
We were encountering some errors when saving specific JPEG images.
The error was shown in stderr as:
  IOError: encoder error -2 when writing image file

And on stdout it printed:
    Suspension not allowed here

The problem was the bufsize not contemplating the icc_profile block.

Fixes # .

 * 
 * 
 *


wiredfool:
LGTM, if you rebase on top of current master everything should pass, as I've fixed the breaking test.

Tests/test_file_jpeg.py
```diff
@@ -133,6 +133,21 @@ def test(n):
...
+        # https://github.com/python-pillow/Pillow/issues/148
+        # Sometimes the meta data on the icc_profile block is biger than
+        # Image.MAXBLOCK or the image size.
+        im = Image.open('Tests/images/icc_profile_big.jpg')
+        f = self.tempfile("temp.jpg")
+        icc_profile = im.info["icc_profile"]
+        try:
+            im.save(f, format='JPEG', progressive=True,quality=95,
+                    icc_profile=icc_profile, optimize=True)
+            assert True
```
hugovk: This redundant line can be removed

Tests/test_file_jpeg.py
```diff
@@ -133,6 +133,21 @@ def test(n):
...
+        except IOError:
+            assert False, "Failed saving image with icc larger than image size"
```
hugovk: This can be replaced with unit test's self.fail https://stackoverflow.com/a/4319870/724176

Tests/test_file_jpeg.py
```diff
@@ -133,6 +133,21 @@ def test(n):
...
+
+
     def test_optimize(self):
```
hugovk: Just one line break between defs

Tests/test_file_jpeg.py
```diff
@@ -133,6 +133,19 @@ def test(n):
         test(ImageFile.MAXBLOCK+1)  # full buffer block plus one byte
         test(ImageFile.MAXBLOCK*4+3)  # large block
 
+    def test_large_icc_meta(self):
+        # https://github.com/python-pillow/Pillow/issues/148
+        # Sometimes the meta data on the icc_profile block is biger than
```
radarhere: bigger

PR #2599, opened 2017-06-28
Added threshold parameter to ImageDraw.floodfill

nediamond:

 * Add threshold parameter to ImageDraw.floodfill for filling of regions with non-homologous, but similar, coloring.

### DIFF
--- a/PIL/ImageDraw.py
+++ b/PIL/ImageDraw.py
@@ -315,7 +315,7 @@ def getdraw(im=None, hints=None):
     return im, handler
 
 
-def floodfill(image, xy, value, border=None):
+def floodfill(image, xy, value, border=None, thresh=0):
     """
     (experimental) Fills a bounded region with a given color.
 
@@ -326,6 +326,10 @@ def floodfill(image, xy, value, border=None):
         pixels with a color different from the border color.  If not given,
         the region consists of pixels having the same color as the seed
         pixel.
+    :param thresh: Optional threshold value which specifies a maximum 
+        tolerable difference of a pixel value from the 'background' in 
+        order for it to be replaced. Useful for filling regions of non-
+        homogenous, but similar, colors.
     """
     # based on an implementation by Eric S. Raymond
     pixel = image.load()
@@ -348,7 +352,7 @@ def floodfill(image, xy, value, border=None):
                     except IndexError:
                         pass
                     else:
-                        if p == background:
+                        if _color_diff(p, background) <= thresh:
                             pixel[s, t] = value
                             newedge.append((s, t))
             edge = newedge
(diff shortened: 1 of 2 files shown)


nediamond:
I think the logs show that the one failure is due to internet reliability.

hugovk:
It was during the CI setup, getting something from AWS. I restated the job and it passed.

PIL/ImageDraw.py
```diff
@@ -326,13 +326,17 @@ def floodfill(image, xy, value, border=None):
         pixels with a color different from the border color.  If not given,
         the region consists of pixels having the same color as the seed
         pixel.
+    :param thresh: Optional threshold value which specifies a maximum 
+        tolerable difference of a pixel value from the 'background' in 
+        order for it to be replaced. Useful for filling regions of non-
+        homogenous, but similar, colors.
```
radarhere: homogeneous

Tests/test_imagedraw.py
```diff
@@ -400,6 +400,28 @@ def test_floodfill_border(self):
         self.assert_image_equal(
             im, Image.open("Tests/images/imagedraw_floodfill2.png"))
 
+
+    @unittest.skipIf(hasattr(sys, 'pypy_version_info'),
```
wiredfool: We shouldn't be skipping pypy, there's nothing here that should cause it to error out. If there is, it indicates a deeper problem somewhere.

PIL/ImageDraw.py
```diff
@@ -366,3 +370,10 @@ def floodfill(image, xy, value, border=None):
                             pixel[s, t] = value
                             newedge.append((s, t))
             edge = newedge
+
+  
+def _color_diff(rgb1, rgb2):
+    """
+    Uses distance formula to calculate difference between two rgb values.
+    """
+    return ((rgb1[0]-rgb2[0])**2 + (rgb1[1]-rgb2[1])**2 + (rgb1[2]-rgb2[2])**2)**.5
```
wiredfool: Classic optimization technique, when comparing pythagorean distances many times, compare (dx^2 + dy^2) to thresh^2, instead of taking the square root each time. It's basically a free optimization that doesn't change the outcome. 

I'm not sure of the speed here, but just comparing linear distance is actually a pretty close approximation. (within 1/8 if you take a cube with corners at x,y,z = +-(9/8)*threshold,0 rather than a sphere of radius threshold.  9/8 gives you roughly equal error at the corners and center)

Tests/test_imagedraw.py
nediamond: Sorry, I just used that decorator because I saw that the other floodfill test required it.  Will remove.

PIL/ImageDraw.py
nediamond: That optimization makes sense, I think it would make the code a bit more confusing though.  The linear distance you talk about sounds like it would make sense to use for speed and cleanness, but I'm not exactly sure what formula you're referring to.  But from my personal testing, speed hasn't been too big of an issue with this code.

PR #2606, opened 2017-07-03
Use "sudo pip install Pillow" to allow installation

ozmium:
The command "pip install Pillow" will not work, and will just produce "Permission denied" installation error messages. So use "sudo pip install Pillow" instead.

Also give link to pip website.

Fixes # .

 * 
 * 
 *

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -24,9 +24,9 @@ Basic Installation
 
     The following instructions will install Pillow with support for most common image formats. See :ref:`external-libraries` for a full list of external libraries supported.
 
-Install Pillow with :command:`pip`::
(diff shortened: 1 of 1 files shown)


hugovk:
It's not needed on macOS, and definitely not needed on Windows.

ozmium:
The document doesn't explain how to workaround this common error: https://stackoverflow.com/questions/31512422/pip-install-r-oserror-errno-13-permission-denied

wiredfool:
This is a general issue with anything that someone is installing with pip, and should be dealt with elsewhere. Sudo, while it may work, is not necessarily the best, or even a good solution to the problem.

PR #2612, opened 2017-07-06
[etc] Fix built-in name.

kijeong:
 * Fix the built-in name "dict" to "dict_" in PIL/PdfImagePlugin.py
It can be seen an error when: python -c "from Cython.Build import cythonize; cythonize(['PIL/PdfImagePlugin.py'])"

### DIFF
--- a/PIL/PdfImagePlugin.py
+++ b/PIL/PdfImagePlugin.py
@@ -37,11 +37,11 @@ __version__ = "0.4"
 #  4. page
 #  5. page contents
 
-def _obj(fp, obj, **dict_):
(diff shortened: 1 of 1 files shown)


hugovk:
I agree builtin names should be avoided, and perhaps `dictionary` is better than `dict_`.

(There's a chance someone may already be calling this and naming `dict` as the parameter, but they shouldn't be, as `_obj` is an internal.)

kijeong:
I agree that 'dictionary' as name is better for PDF dictionary objects parameter.

PR #2621, opened 2017-07-13
Docs: added the description of the filename attribute to images.rst

dasdachs:
Images created with `open` have a filename attribute, but the documentation does not mention it. This lead to a short, but interesting discussion on [Stackoverflow](https://stackoverflow.com/questions/45087638/get-image-filename-from-image-pil/).


wiredfool:
(note -- originally against 4.2.x, cherrypicked into master rather than the release/bugfix branch)

PR #2633, opened 2017-07-18
Delete transparency info when img.convert'ing RGB/L to RGBA

olt:
`info['transparency]` was not removed when an RGB or L image
was converted to RGBA. This could result in unexpected behavior
when saving the resulting image.

Other image conversions already delete or update the transparency
info. There is a shortcut for RGB/L to RGBA which missed this.

### DIFF
--- a/PIL/Image.py
+++ b/PIL/Image.py
@@ -877,8 +877,10 @@ class Image(object):
             if self.mode in ('L', 'RGB') and mode == 'RGBA':
                 # Use transparent conversion to promote from transparent
                 # color to an alpha channel.
-                return self._new(self.im.convert_transparent(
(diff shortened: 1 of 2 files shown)


homm:
> This could result in unexpected behavior when saving the resulting image.

While in general this looks good, could you provide simple example with code which leads to unexpected behavior.

homm:
@olt I'd like to see this in the further version, but also I'd like to know what exactly the issue is. Could you provide an example of unexpected behavior?

wiredfool:
I think this is roughly the same issue as #2702.  RGBA images shouldn't have any transparency or palette info, as all of information that should affect the image interpretation is in the pixels themselves. (excepting color profiles.  but that's a different can of worms).

PR #2638, opened 2017-07-22
[SGI] Read RLE compressed images

jbltx:

 * Add Run-Length Encoding support to read compressed SGI images (.sgi, .rgb, ...) 
 * Edit Unit test for `test_rle` in **test_file_sgi.py**


jbltx:
I'm working on reading 16bits RLE compressed SGI images, i will push it soon. I also need to fix RLE decoding with images using alpha channels before merging branches.

jbltx:
Well, i am a bit stuck with the `shuffle` function from the `ImagingCodecState` instance... For images with 3 channels it works fine when I send a full line to the shuffle like this : 
```c
state->buffer = {R0, G0, B0, R1, G1, B1, ...}; /* UINT8* */
state->shuffle((UINT8*) im->image[state->y], state->buffer, state->xsize);

/* in Python */
list(img.getdata())[0]
=> (R0, G0, B0) : same values
```
but with four channels, the data returned in the PIL image is wrong
```c
state->buffer = {R0, G0, B0, A0,  R1, G1, B1, A1, ...}; /* UINT8* */
state->shuffle((UINT8*) im->image[state->y], state->buffer, state->xsize);

/* in Python */
list(img.getdata())[0]
=> (≠R0, ≠G0, ≠B0, ≠A0) : not same values
```
The decoder returns a valid `"RGBA"` rawmode and image data contains 4 bands too. I don't get what is wrong with the shuffle method.

wiredfool:
Which shuffler are you getting from get_unpack?  Does the code look like it does the right thing?

jbltx:
It's the copy4 unpacker, which just makes a memcpy on my buffers.

I just saved an other RGBA images in Photoshop and it works fine, I think my old image used for my test may differ a bit from the SGI file format documentation, but applications like XnView can read it and show it properly.

wiredfool:
I'd believe that you've got a weird image before the copy4 unpacker not working properly. 

Many of these old formats acquired interesting variants that aren't well reflected in the specs.

jbltx:
Strange things happened, with larger images. 

I'm not sure if the `buffer` parameter in _decode function is well set, because I met a segfault with an rleoffset equals to **2106536** and an image of more than **3000000** bytes.

I thought this buffer parameter contained bytes from the end of the file header to the EOF.
It seems to be set using `PyArg_ParseTuple` function in **decode.c** file.

Any informations about it ?
Thanks.

```
Program received signal SIGSEGV, Segmentation fault.
ImagingSgiRleDecode (im=<optimized out>, state=0x7ffff508fb78, buf=0xd15224 "", 
    bytes=<optimized out>) at libImaging/SgiRleDecode.c:100
100	            printf("count 0 -> %d\n", buf[rleoffset]);
(gdb) bt
#0  ImagingSgiRleDecode (im=<optimized out>, state=0x7ffff508fb78, buf=0xd15224 "", 
    bytes=<optimized out>) at libImaging/SgiRleDecode.c:100
#1  0x00007ffff5ade137 in _decode (decoder=0x7ffff508fb58, args=<optimized out>)
    at decode.c:130
#2  0x00000000004c468a in PyEval_EvalFrameEx ()
#3  0x00000000004c9d8f in PyEval_EvalFrameEx ()
#4  0x00000000004c2765 in PyEval_EvalCodeEx ()
#5  0x00000000004ca099 in PyEval_EvalFrameEx ()
#6  0x00000000004c2765 in PyEval_EvalCodeEx ()
#7  0x00000000004c2509 in PyEval_EvalCode ()
#8  0x00000000004f1def in ?? ()
#9  0x00000000004ec652 in PyRun_FileExFlags ()
#10 0x00000000004eae31 in PyRun_SimpleFileExFlags ()
#11 0x000000000049e14a in Py_Main ()
#12 0x00007ffff7810830 in __libc_start_main (main=0x49dab0 <main>, argc=2, 
    argv=0x7fffffffde18, init=<optimized out>, fini=<optimized out>, 
    rtld_fini=<optimized out>, stack_end=0x7fffffffde08) at ../csu/libc-start.c:291
#13 0x000000000049d9d9 in _start ()
```

wiredfool:
The `buf` parameter is a chunk of the file of length `bytes`, I think it's in the range of 64k or so. The C decoders are generally required to decode incrementally to avoid allocating multiple image sized buffers for very large images. There are some exceptions to this in the code, especially in the _pulls_fd code, where the decoder is responsible for reading the data from the open file. 

If you look at some of the other decoders, there's a multiple stage process, where the initial call to the decoder does the setup and decodes the first image data, then subsequent calls decode additional data, then once the image is fully decoded, any cleanup happens. (note that there's also a hook in the decoder to do that cleanup on a failed image)

jbltx:
okey I finally used the file descriptor instead of chunks, it's the easiest way to decompress rows. I'm not sure what is wrong with AppVeyor tests however.

jbltx:
Everything seems fine now, one job on Travis has failed but it was a "Too Many Requests" svn error on Pillow-depends repository.

hugovk:
I restated that build job and all is green.

PIL/SgiImagePlugin.py
```diff
@@ -44,54 +58,89 @@ class SgiImageFile(ImageFile.ImageFile):
...
+                            offset, (layer, 0, orientation)))
+                    offset += pagesize
         elif compression == 1:
-            raise ValueError("SGI RLE encoding not supported")
+            self.tile = [("sgi_rle", (0, 0) + self.size,
+                          headlen, (rawmode, orientation, bpc))]
 
 
 def _save(im, fp, filename):
     if im.mode != "RGB" and im.mode != "RGBA" and im.mode != "L":
```
wiredfool: If you're adding support for 16bpc, then `I` or `I;16` mode images should probably be allowed through here.

PIL/SgiImagePlugin.py
```diff
@@ -123,28 +177,59 @@ def _save(im, fp, filename):
...
-                         (z, len(im.getbands())))
-
     for channel in im.split():
-        fp.write(channel.tobytes())
+        rawchannel = channel.tobytes()
+        if bpc == 1:
+            fp.write(rawchannel)
+        else:
+            for pixel in rawchannel:
+                fp.write(o16(i8(pixel) * 256))
```
wiredfool: Should probably do this in C by converting to `I`, and using  ImageMath to multiply by 256.

(though, somewhere there should be a MSB promotion to `I` or `I;16`, as that's a common issue)

libImaging/SgiRleDecode.c
```diff
@@ -0,0 +1,168 @@
...
+{
+    UINT8 *ptr;
+    SGISTATE *c;
+
+    /* Get all data from File descriptor */    
+    c = (SGISTATE*)state->context;
+    _imaging_seek_pyFd(state->fd, 0L, SEEK_END);
+    c->bufsize = _imaging_tell_pyFd(state->fd);
+    c->bufsize -= SGI_HEADER_SIZE;
+    ptr = malloc(sizeof(UINT8) * c->bufsize);
```
wiredfool: This is a potentially large malloc. Is there a way to read this file piece by piece? 

* sizeof(uint8) should be 1 on all reasonable platforms
* Need to check the return value and handle any errors.

libImaging/SgiRleDecode.c
```diff
@@ -0,0 +1,168 @@
...
+        state->y = im->ysize - 1;
+    else
+        state->ystep = 1;
+
+    /* Allocate memory for RLE tables and rows */
+    free(state->buffer);
+    state->buffer = malloc(sizeof(UINT8) * 2 * im->xsize * im->bands);
+    c->tablen = im->bands * im->ysize;
+    c->starttab = calloc(c->tablen, sizeof(UINT32));
+    c->lengthtab = calloc(c->tablen, sizeof(UINT32));
```
wiredfool: Need to overflow check any of these items, especially if they're potentially attacker controlled.

libImaging/SgiRleDecode.c
```diff
@@ -0,0 +1,168 @@
...
+    state->count = 0;
+    state->y = 0;
+    if (state->ystep < 0)
+        state->y = im->ysize - 1;
+    else
+        state->ystep = 1;
+
+    /* Allocate memory for RLE tables and rows */
+    free(state->buffer);
+    state->buffer = malloc(sizeof(UINT8) * 2 * im->xsize * im->bands);
```
wiredfool: Should probably just calloc this one as well. `calloc(im->xsize, im->bands * 2 )`

* sizeof(uint8) should be 1

wiredfool:
Closing in favor of #2769

PR #2647, opened 2017-07-28
Added file pointer save to SpiderImagePlugin

radarhere:
I'm actually not sure whether the test should be included - while it demonstrates the need for this PR, the test isn't standard for other formats.

### DIFF
--- a/PIL/SpiderImagePlugin.py
+++ b/PIL/SpiderImagePlugin.py
@@ -267,16 +267,10 @@ def _save(im, fp, filename):
         raise IOError("Error creating Spider header")
 
     # write the SPIDER header
-    try:
-        fp = open(filename, 'wb')
-    except:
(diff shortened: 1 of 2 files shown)


wiredfool:
I'm not clear what's going on here, and why this plugin is different from the others. It does seem like this would be a breaking change as it effectively changes the type of fp from a filename to an open file handle.

radarhere:
fp was always an open file handle. The current code ignores the passed in file handle, and instead opens a new file handle from the filename. This change instead uses the passed in file handle.

It shouldn't be a breaking change. File pointer saving for the SPIDER format is not currently supported, so any saves being done are using a filename. The passed in fp is an open file pointer for that filename, as set in Image.save - https://github.com/python-pillow/Pillow/blob/master/PIL/Image.py#L1916

PR #2648, opened 2017-08-03
Fixed bc5 decoding

ifeherva:
BC5 decoding uses only 2 channels instead of 4. The current algorithm did not initialize the RGBA struct which resulted in random values in the output image. This commit should initialize the decoded RGBA struct, thus setting B and A values to default (0).

 * Initialize target buffer to zero before filling.


jleclanche:
Looks good to merge.

PR #2653, opened 2017-08-06
Remove Landscape/health badge

radarhere:
See #2584


hugovk:
There's always one more! Thanks!

PR #2657, opened 2017-08-06
Remove unused im.copy2 and core.copy methods

homm:


homm:
Add to release notes when https://github.com/python-pillow/Pillow/pull/2661 will be merged

PR #2659, opened 2017-08-07
Added tests for SgiImagePlugin

radarhere:
While I have added a test for it, I'm skeptical about the error raised at https://github.com/python-pillow/Pillow/blob/master/PIL/SgiImagePlugin.py#L136. In the current state of the plugin, it's impossible to trigger without breaking the image as I do here - from what I can see, the number of modes are static values retrieved from https://github.com/python-pillow/Pillow/blob/master/PIL/Image.py#L211.

Any thoughts from @wiredfool, who added it in https://github.com/python-pillow/Pillow/pull/2325/commits/31c204eae4e158bbaa6cda4afe7a2f8e2fcabe6e?

### DIFF
--- a/Tests/test_file_sgi.py
+++ b/Tests/test_file_sgi.py
@@ -55,6 +55,9 @@ class TestFileSgi(PillowTestCase):
         for mode in ('L', 'RGB', 'RGBA'):
             roundtrip(hopper(mode))
 
+        # Test 1 dimension for an L mode image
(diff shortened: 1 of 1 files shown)


wiredfool:
That test was essentially a sanity check. I don't expect that it will ever go wrong, but we need to have the same number of bands as we have z planes in the image.

Issue #2672, opened 2017-08-11
FLI reports one too many frames


adambiser:
Sorry, I don't have any sample code or files to demonstrate this, but it's also not a huge issue.

The [n_frames function](https://github.com/python-pillow/Pillow/blob/a4dafe78df5fa050131bddac0b2378d11ea8d9b8/PIL/FliImagePlugin.py#L112) for FliImagePlugin determines the number of frames in the file by reading frames until EOF is reached.  However, from what I've seen in the FLI files I've tested and according to [this page](http://steve.hollasch.net/cgindex/formats/fli.html), "There is one extra frame at the end of a FLI which contains the difference between the last frame and the first frame." There is a frame count given in the FLI header which does not include the extra frame and I'm curious why that value is not used.

Since ["All flic files will contain a ring frame, including a single-frame flic."](http://www.martinreddy.net/gfx/anim/FLC.txt), the [is_animated function](https://github.com/python-pillow/Pillow/blob/a4dafe78df5fa050131bddac0b2378d11ea8d9b8/PIL/FliImagePlugin.py#L129) will always report true since it is checking for the presence of a second frame in the file which would be the ring frame for single-frame files. Granted, having a single-frame FLI is rather pointless...

radarhere:
I have created at PR to implement your idea of using the value from the FLI header - #2674. So thanks for that.

With regards to the issue of 'one too many frames', currently, `is_animated` [returns false for the test image](https://github.com/python-pillow/Pillow/blob/master/Tests/test_file_fli.py#L28), and `n_frames` [reports that the test image has 1 frame](https://github.com/python-pillow/Pillow/blob/master/Tests/test_file_fli.py#L27). So I wouldn't conclude that Pillow is reporting one too many frames?

adambiser:
Sorry for not trying that test FLI file. Its header count and the cycling to find n_frames both give 1. So obviously there are some exceptions to the rule.

The FLI files I've been dealing with cannot be shared do to copyright, but a.fli found at https://samples.libav.org/fli-flc/ (also used by me in issue 2649) reports an n_frame of 385 while the header reports 384. seek(384) gives the same image as seek(0).

I don't have another single-frame file on hand, so I can't provide another example for it.

hugovk:
> There is one extra frame at the end of a FLI which contains the difference between the last frame and the first frame.

Maybe it was counting this for animated.

adambiser:
[...]

PR #2674, opened 2017-08-12
Use frame count from FLI header

radarhere:
Suggested in #2672, documented at http://steve.hollasch.net/cgindex/formats/fli.html


hugovk:
Other than the frame count, do the tests in https://github.com/python-pillow/Pillow/pull/2650 pass with this?

radarhere:
Yes

PR #2679, opened 2017-08-12
Fast filters

homm:
```
               Ubuntu 14.04    i2f or 16.04       Optimized
Smooth        12.47 Mpx/s     28.81 Mpx/s       44.71 Mpx/s
Sharpen       12.34 Mpx/s     29.94 Mpx/s       44.74 Mpx/s
Smooth More    4.41 Mpx/s     12.49 Mpx/s       17.61 Mpx/s
```


PIL/Image.py
```diff
@@ -1112,7 +1112,9 @@ def filter(self, filter):
             raise TypeError("filter argument should be ImageFilter.Filter " +
                             "instance or class")
 
-        if self.im.bands == 1:
+        multiband = getattr(filter, 'is_multiband', False)
```
homm: All filters used to work only with single-channel images. As this is public API and there are can be custom filters in a wild, we can't just send multi-channel images to all filters. So `is_multiband` flag indicates that the filter is ready for multi-channel images.

_imaging.c
```diff
@@ -852,8 +852,12 @@ _filter(ImagingObject* self, PyObject* args)
         return ImagingError_ValueError("bad kernel size");
     }
 
+    for (i = 0; i < kernelsize; ++i) {
+        kerneldata[i] /= divisor;
```
homm: Precompute by dividing coefficients to divisor rather than each result pixel.

homm:
Oh my god! I clash with the [same bug in Intel x64 CPUs](https://stackoverflow.com/q/26585977/253146) with int to float conversion, as I did when [was working on resampling](https://github.com/python-pillow/Pillow/pull/977#issuecomment-60832877) for Pillow 2.7.

In essence: GCC version < 4.9 (at least Ubuntu 14.04, Debian 7 Wheezy, Red Hat Linux 7) generates code which runs about 2.5 times slower than GCC >= 4.9. This can be fixed using [ugly hack with assembler](https://github.com/python-pillow/Pillow/blob/90ee223305baef8c7f2d8923849fe39cd9c255f4/libImaging/Antialias.c#L93-L105). We decided to include this hack in Pillow 2.7 (released Jan 1, 2015). After rewriting the resampling using fixed point arithmetics, this hack [become obsolete and was removed](https://github.com/python-pillow/Pillow/pull/1881#issuecomment-217396518) in Pillow 3.3 (Jul 1, 2016). Unlike for resampling, I don't see a way to implement kernel filters on FPA. So, I suggest to include this hack again.

homm:
I've just tried SSE4 for multiband images, with no specific optimizations. Much faster, but not enough )

```
                Simple SSE4     OpenCV
Smooth         137.25 Mpx/s    163.68 Mpx/s
Sharpen        133.55 Mpx/s    163.81 Mpx/s
Smooth More     45.71 Mpx/s     64.01 Mpx/s
```

PIL/Image.py
homm: @wiredfool As of naming things is one of the hard problems in programming, I need your help there )

wiredfool: I think that's a reasonable name, at least, I haven't thought of anything better.

homm: The another option is subclass `MultibandFilter`.

wiredfool: That's probably cleaner, especially with an eye towards the typing PR.

homm:
@wiredfool implemented through MultibandFilter

PR #2686, opened 2017-08-17
Use ImagingNewDirty when possible

homm:
Use ImagingNewDirty when possible. I've checked all occurrences, in all cases all pixels are filled unconditionally.

Fix indention in affected files.


libImaging/Effects.c
```diff
@@ -121,14 +121,14 @@ ImagingEffectSpread(Imaging imIn, int distance)
...
 
     if (!imOut)
-    return NULL;
+        return NULL;
 
 #define SPREAD(type, image)\
-    for (y = 0; y < imIn->ysize; y++)\
-    for (x = 0; x < imIn->xsize; x++) {\
+    for (y = 0; y < imOut->ysize; y++)\
+        for (x = 0; x < imOut->xsize; x++) {\
```
homm: In this case `imIn->xsize` == `imOut->xsize` and `imIn->ysize` = `imOut->ysize`. imOut more explicit points to actual action.

homm:
I realized that 25% of uncovered lines are lines with `ImagingError` and there is no way to cover them

PR #2687, opened 2017-08-17
[WIP]: Type annotations for Image.py

### LINKED ISSUE #2625: Support for typing?
I've been looking at applying typing to some code which uses PIL/pillow.

Is there any feeling regarding the addition of type annotations within Pillow, or if these belong outside the code in eg. typeshed? This influences whether I might work on a PR against Pillow itself, or generate stubs for typeshed.

On the basis of supporting python 2 and 3, I expect annotations would be in comments, and there would be an import of the typing module near the top, to define required types.

neiljp:
Fixes #2625 .

 * Add type annotations for many functions (almost all)
 * Where mypy suggests, and in other places, add annotations for variables (often empty containers or initially-None)
 * Introduce typing names within `if False:` block, as well as type aliases to simplify annotations
 * Some imports, which may be temporary or unnecessary (based on how I'm running mypy)

Annotation caveats:
* Some of these annotations may be incorrect, and some still use 'dynamic annotations' (ie. Any); any improvements in that direction would be welcome!
* The documentation is great in most cases, though in some it is a little vague or does conflict!

There are a few cases where I don't follow why the code is why it is, or where a small change might allow mypy to 'pass' the problem (eg infer types).

I'm currently using a development branch of mypy which is being used in Zulip, but I'm happy to use another version if that'd be easier. mypy doesn't pass this as it stands, with the options:
```
-2 --ignore-missing-imports --follow-imports=silent --check-untyped-defs
```


PIL/Image.py
```diff
@@ -1049,6 +1108,7 @@ def crop(self, box=None):
         return self._new(self._crop(self.im, box))
 
     def _crop(self, im, box):
+        # type: (Image, LURD) -> Image
```
wiredfool: This is actually a core Imaging object, which is not an Image.Image but rather a PIL._imaging object.

PIL/Image.py
```diff
@@ -1146,6 +1212,7 @@ def getbbox(self):
         return self.im.getbbox()
 
     def getcolors(self, maxcolors=256):
+        # type: (int) -> Optional[List[Tuple[int, int]]]
```
wiredfool: Why isn't this an Optional[int]?

wiredfool:
I'm having trouble getting mypy to run in a manner that looks remotely correct. 

Starting in a py35 virtual environment, just so that I'm not dealing with any python2 stuff. Pip installed mypy (for now), and running with the command line options above. This branch of pillow is installed in the virtualenv with `make install`.  From outside the Pillow directory, `from PIL import Image` works correctly as expected. 

However, mypy can't find the module:
```
(vpy35)erics@builder-1204-x64:~/test$ mypy -2 --ignore-missing-imports --follow-imports=silent --check-untyped-defs -m PIL.Image
mypy: can't find module 'PIL.Image'
(vpy35)erics@builder-1204-x64:~/test$ mypy -2 --ignore-missing-imports --follow-imports=silent --check-untyped-defs -m PIL
mypy: can't find module 'PIL'
```

Running from within the PIL directory, I get this:
```
(vpy35)erics@builder-1204-x64:~/Pillow$ mypy -2 --ignore-missing-imports --follow-imports=silent --check-untyped-defs PIL/Image.py | more
/home/erics/Pillow/PIL/BmpImagePlugin.py:128: error: INTERNAL ERROR -- please report a bug at https://github.com/python/mypy/issues version: 0.521
PIL/Image.py:75: error: Module 'PIL' has no attribute '_imaging'
PIL/Image.py:406: error: Missing return statement
[ ... many more lines]
```

I'm mostly worried about the `Module 'PIL' has no attribute '_imaging'` error, as that indicates that mypy isn't able to get the core imaging functions from the C module which has been installed.(*) This is not altogether surprising due to the source file layout and the path tricks needed to be able to access the PIL module from within the Pillow directory. (see `test-installed.py` for one of the workarounds) 


* Also leading to the question, can mypy infer types from C extensions, where there is an explicit PyArg_ParseTuple call?

PIL/Image.py
neiljp: Optional[int] is equivalent to Union[int, None], which isn't the case here?

PIL/Image.py
neiljp: It took me a while to realise what was happening (the code is new to me) but as per my comment below this is clearer now.

PIL/Image.py
wiredfool: Ah, so it's that the parameter is always an int, not that it's sometimes required as a named parameter.

neiljp:
I've not run mypy using module imports before, so I don't have a response from experience for that case. To be precise, most recently the exact command I'm (now) running is:
```
mypy PIL/Image.py -2
```
from the main repo directory, though the absence of those extra flags does have the side effect of checking the imports.

I've not actually built the C code, since as far as I know mypy only checks python and doesn't delve into eg. built object code. On that note, and related to your comments above, I've actually been attempting to make some stubs for the 'core' as '_imaging.pyi'.

I'm thinking that the type aliases/imports might best be in a separate module?

Is this something you think might be merged, down the line?

PIL/Image.py
neiljp: Right, I have to think about this too, with function parameters, as `Optional` can be a misleading term!

PIL/Image.py
```diff
@@ -632,6 +677,7 @@ def _repr_png_(self):
 
     @property
     def __array_interface__(self):
+        # type: () -> Dict[Text, Any]
```
bjd2385: According to the [NumPy documentation on `__array_interface__`](https://docs.scipy.org/doc/numpy/reference/arrays.interface.html), and looking over the source file (particularly at `_conv_type_shape`), I think this particular `Any` could be constrained further to `Tuple[Tuple[int, ...], Text, int, bytes]`.

PIL/Image.py
```diff
@@ -667,6 +715,7 @@ def __setstate__(self, state):
         self.frombytes(data)
 
     def tobytes(self, encoder_name="raw", *args):
```
bjd2385: Nothing to do with this particular PR, but from the typing, it seems that the function `_getencoder` is rather...strange. Like why do we need to _tweak_ arguments ?

PIL/Image.py
```diff
@@ -735,6 +786,7 @@ def tobitmap(self, name="image"):
             ])
 
     def frombytes(self, data, decoder_name="raw", *args):
```
bjd2385: This has the same format as `getencoder`, wherein we _tweak_ arguments.

PIL/Image.py
```diff
@@ -1092,12 +1154,14 @@ def draft(self, mode, size):
         pass
```
bjd2385: The signature could probably be `(Text, Tuple[int, ...]) -> Something`...but why does the docstring say it's implemented for JPEG && PCD images when it's just a `pass`?

PIL/Image.py
```diff
@@ -1122,6 +1186,7 @@ def filter(self, filter):
         return merge(self.mode, ims)
 
     def getbands(self):
+        # type: () -> Tuple[Any]
```
bjd2385: Instead of `Tuple[Any]`, couldn't this be `Tuple[str, ...]`, or...I'm so confused by  `ImageMode.py`, and why are there so many `global` variables? It should be far easier to narrow down what this method should return as a type.

PIL/Image.py
```diff
@@ -1219,6 +1289,7 @@ def getim(self):
         return self.im.ptr
 
     def getpalette(self):
+        # type: () -> Optional[List[int]]
```
bjd2385: I don't believe this accepts any arguments?

PIL/Image.py
```diff
@@ -1236,6 +1307,7 @@ def getpalette(self):
             return None  # no palette
 
     def getpixel(self, xy):
+        # type: (Coord) -> Tuple[Any]
```
bjd2385: Could this `Any` be constrained to a type of `float` or `int`? For example, `Tuple[Union[float, int], ...]`.

neiljp: The issue here is that we're returning a tuple, but whose length varies - though actually also the 'color' type, which I'm not sure how to type offhand?

So, if we alias `Color = [something]`, then ideally the function would return `Union[Color, Tuple[Color, Color], Tuple[Color, Color, Color], ...]`... but where we go all the way to the maximum number of layers in the image! Since I'm not sure if that's possible to quantify, I stuck with `Tuple` for now, though it should probably be `Union[Color, Tuple]` for now, for some value (type) of Color.

PIL/Image.py
```diff
@@ -504,26 +535,30 @@ class Image(object):
...
+        # type: () -> None
         # FIXME: take "new" parameters / other image?
         # FIXME: turn mode and size into delegating properties?
         self.im = None
         self.mode = ""
         self.size = (0, 0)
-        self.palette = None
-        self.info = {}
+        self.palette = None  # type: Optional[ImagePalette.ImagePalette]
+        self.info = {}  # type: Dict[Text, Any]
```
neiljp: We can certainly work on this, though I'm not sure exactly what it might represent. If the set of keys/values are limited, a `TypedDict` might be plausible in the future, even :)

PIL/Image.py
```diff
@@ -538,12 +573,15 @@ def _new(self, im):
 
     # Context Manager Support
     def __enter__(self):
+        # type: () -> Image
         return self
 
     def __exit__(self, *args):
+        # type: (*Any) -> None
```
neiljp: I'm not too familiar with __exit__, but I believe in python2 that `*args` can only be one type?
If not, that's certainly plausible, particularly if changing the function signature (which I've been avoiding).

PIL/Image.py
neiljp: Do you mean that to be a Union? From the docs only, and given that this needs to potentially interoperate with other code which exposes all the dictionary keys, I'd imagine this at least needs to include `None` and `List` as well?

This may be where it's sometimes easier to use `Any`, at least initially ;)

PIL/Image.py
neiljp: Re the type, I'd imagine this is `(Mode, Size)` based on the type aliases, ie. `(Text, Tuple[int, int])`; or could the tuple be bigger?

PIL/Image.py
neiljp: This was due to the same point I raised elsewhere regarding Tuples of unknown length.

neiljp:
@bjd2385 Thanks for the comments. I've replied to what I could, and have pushed the changes I've made, which include:
* pulling the aliases into a separate file - maybe rename to _aliases.py, or _types.py?
* adding some aliases
* attempt to document _imaging via a stubs file (very much WIP)
* add 'ignore' annotations for ne/eq, as they seem unclear in general stubs
* __repr__ return type to `str` from Text (it's not unicode ever?)
* changes to imports which hopefully improve the situation

On other projects I'm used to committing solo in PRs; will it be a problem if I force-push, or might you fork from my branch to work/commit?

PIL/Image.py
neiljp: **update*: so, probably `Tuple[Mode, ...]`, which is the same as you specified, except that Mode is (in an updated commit I made) an alias for Text, not str.

PIL/Image.py
bjd2385: I think `Union[Color, Tuple[Color, ...]]`looks right. But what do you mean by alias `Color = [something]`? As in `Color` refers to a list of some values?

PIL/Image.py
```diff
@@ -1250,6 +1322,7 @@ def getpixel(self, xy):
         return self.im.getpixel(xy)
 
     def getprojection(self):
+        # type: () -> Tuple[List[Any], List[Any]]
```
bjd2385: I think we can constrain this return type to `Tuple[List[int], List[int]]` ~~, but I'm confused a little by what `_binary.i8` actually does.~~ I wasn't aware that in Python 2, `bytes is str`.

PIL/Image.py
```diff
@@ -1262,6 +1335,7 @@ def getprojection(self):
         return [i8(c) for c in x], [i8(c) for c in y]
 
     def histogram(self, mask=None, extrema=None):
+        # type: (Optional[Image], Optional[Any]) -> List[int]
```
bjd2385: Instead of `Optional[Any]`, which resolves to the same thing as `Any` I believe, perhaps this should be `Optional[ImagingCore]` ? Also, the docstring doesn't have a corresponding `:param` mentioning this parameter.

PIL/Image.py
```diff
@@ -1426,6 +1503,7 @@ def alpha_composite(self, im, dest=(0,0), source=(0,0)):
         self.paste(result, box)
 
     def point(self, lut, mode=None):
+        # type: (Union[List, Callable[[Any], Any]], Optional[Text]) -> Image
```
bjd2385: Out of curiosity -- i'm not familiar with what you're describing with `[Any]`. Is it meant to be `List[Any]`?

PIL/Image.py
```diff
@@ -2142,6 +2237,7 @@ def __transformer(self, box, image, method, data,
         self.im.transform2(box, image.im, method, data, resample, fill)
```
bjd2385: Where is this method definition? I'd like to know what `fill` is, why it defaults to a `1`. I don't see where it's used in `_imaging.c/_transform2` either; i.e. I see it's parsed, but I don't see it being used anywhere in that function.

PIL/Image.py
```diff
@@ -1209,6 +1278,7 @@ def getextrema(self):
         return self.im.getextrema()
 
     def getim(self):
+        # type: () -> Any
```
wiredfool: A PyCapsule is an internal python type: https://docs.python.org/3.6/c-api/capsule.html

PIL/Image.py
wiredfool: Backwards compatibility. Plugins for PIL were added in the Python 1.6 era, IIRC.

PIL/Image.py
wiredfool: It's an iterable of length 2. There are places in calling code (i.e. django) where a 2 element list is passed in. 

We've hit this usage before when making things tighter with the type/size checking.

PIL/Image.py
wiredfool: Should return a tuple of single character strings, where len==number of bands/channels.

PIL/Image.py
wiredfool: +1 on `any` here for now, this is an implementation of an interface defined elsewhere for datatype->datatype conversion, it's not something that's going to get hit in ordinary development., so type hints for IDEs aren't going to be essential.

PIL/Image.py
wiredfool: There's a specific set of what can be called on a context manager. We can change the signature to that, as it's the only way that this should ever get called.

PIL/Image.py
wiredfool: Info can be almost anything. It's pulled from the metadata in images, so bytes, string, tuples of int/float/?, float, ints are certainly possible. You can pretty much fit a ham sandwich in tiff metadata, and that's the source for a lot of this.

neiljp: Right, so leave it at `Any` for now? That said, I think I read that it could be better to use `object` where it could be 'anything', but is that going to work for all the types in the python2 that PIL[low] supports?

PIL/Image.py
neiljp: Right, sorry, I forgot the context...was context-manager.

PIL/Image.py
neiljp: Maybe `object`?

PIL/Image.py
neiljp: The 'ABC's aren't solid in my mind, but this could be `Sequence`? (__getitem__, __len__)

PIL/Image.py
neiljp: I'm not sure there's a way to specify a 'char' in the type (ie. length==1), but we can certainly add a type alias for `Band = Text`, which documents it nicely?

Aliases of the same underlying type are interchangeable (eg. `Band` and `Mode`), but we could go beyond this and make them separate types.

PIL/Image.py
neiljp: See the previous added code at the top of `Image.py`, now in `aliases.py`.

We could add `Color = float`, for example.

PIL/Image.py
bjd2385: So Pillow targets almost every Python distribution? Not just the currently supported versions?

PIL/Image.py
neiljp: I couldn't find an explicit type for these so far. This may be an issue for mypy or typeshed, or perhaps we just leave it at `Any` for now?

PIL/Image.py
neiljp: You're referring to the extrema parameter? I've looked at this again, and added an `Extrema` alias to aliases.py; let me know what you think.

PIL/Image.py
neiljp: For a `Callable`, a fully-typed version is `Callable[[PARAM_LIST], RETURNED]`.

The related PEP (484?) or the mypy docs are pretty good for this :)

PIL/Image.py
bjd2385: Ohhh, okay, that makes sense! I have no idea how to annotate that then. May as well be `Any`.

PIL/Image.py
bjd2385: That looks right to me, i.e. `Sequence`.

I know for some types it's a little ambiguous; like, when should I only use `Iterable` instead of `Sequence`? I see for the former we have `__abstractmethods__ = frozenset({'__getitem__', '__len__'})`, and then for the latter `__abstractmethods__ = frozenset({'__iter__'})`, but I have yet to see an implementation of an interable in Python that doesn't use `__getitem__` with `__iter__` to retrieve values.

PIL/Image.py
bjd2385: @neiljp Okay, that makes sense. I've never written it that way and PyCharm/Mypy didn't complain, but I'll try it in the future.

PIL/Image.py
neiljp: Well, while it's possible to infer length while iterating, these are supposed to be length==2 so it makes sense to assume it might be tested (which it currently is not?)

@wiredfool Re django etc, does this mean that all such `Size` usage should not be a tuple?

PIL/Image.py
neiljp: What was the alternative form you used?

PIL/Image.py
wiredfool: No, but the interface dates from that era, and there could be private image plugins out in the wild using the interface.  So we can't just change the interface and touch up our plugins, if we make changes, we need to do backwards compatibility and deprecation periods.

PIL/Image.py
wiredfool: Yes, leave it as any.

PIL/Image.py
wiredfool: It means that at one point we checked for a tuple, and broke people on release.

Otoh, if we document it as a tuple, people being typesafe will get it right, and eventually people using list because it works will get linted out as pep8 starts pushing typing.  

I think our current stance is we specify a tuple, but we don't reject a 2list (anymore)

PIL/Image.py
bjd2385: @wiredfool as you can probably tell, I'm still new to this, and contributing to open source. I'm trying though.

wiredfool:
> On other projects I'm used to committing solo in PRs; will it be a problem if I force-push, or might you fork from my branch to work/commit?

I'd rather not have rebases or force-pushes when there's a lot of discussion going on, as it wipes out the history. 

> I'm thinking that the type aliases/imports might best be in a separate module?

I think that's a reasonable approach, similar to a header file in C.

> Is this something you think might be merged, down the line?

I think so. There are really ~three~four criteria I see:
1) Is it a future maintenance burden? Does it mess up the source files or add tons of effort?
2) Is it complete/correct. (or at least, complete enough to be useful, and what is complete is correct)
3) Are there tests or other (automated) acceptance criteria that will notify us when we break something. 
4) Documentation. 

I think I'm comfortable saying that we'll hit #1. The impact to the source doesn't look bad from here, and I think on an incremental basis, it's not going to add too much effort to any individual feature. 

#2 is an ongoing effort. 

#3 would be a hook into our testing system. A new test case, e.g. `Tests/test_types.py`  Gate it on typing and mypy being installed, run mypy on any source files, and have a pass / fail based on the results. This needs to fail on a new method in Image.Image with missing types or incorrect types.  If we don't have this, the type support is likely to decay. 

#4 Documentation may not need to be huge, but there should be some.

PIL/Image.py
wiredfool: No, it doesn't appear to be used.  I'm going to say historical anomaly, as it's passed in and used in ImageTransform, but here in transform2 it's hardwired to 1.

PIL/Image.py
wiredfool: @bjd2385 No worries. You clearly know more about type systems than I do. 

There's a lot of history in this code base, and some parts of it look strange until you try to work out what it would take to modernize it without breaking everyone else's code. Sometimes updating works, sometimes it doesn't. (tiff metadata is a poster child for that)

PIL/Image.py
```diff
@@ -504,26 +530,30 @@ class Image(object):
...
+        # type: () -> int
         return self.size[0]
 
     @property
     def height(self):
+        # type: () -> int
         return self.size[1]
 
     def _new(self, im):
+        # type: (Image) -> Image
```
wiredfool: That's (ImagingCore) -> Image

Anything that's assigned to self.im had better be an imaging core.

(realizing that mypy should us that...)

PIL/Image.py
bjd2385: @wiredfool could that be removed in a future PR then?

PIL/Image.py
```diff
@@ -599,7 +635,7 @@ def _dump(self, file=None, format=None, **options):
             self.save(file, format, **options)
         return file
 
-    def __eq__(self, other):
+    def __eq__(self, other):  # type: ignore # type: (Image) -> bool
         return (self.__class__.__name__ == other.__class__.__name__ and
```
wiredfool: The __eq__ and other __[comparison]__ methods take any object, as they're the implementation of `img == a`, and `img == True` or `img == 1`. They're valid comparisons even if the other object isn't an image. False, but valid.

PIL/Image.py
bjd2385: @wiredfool Regarding `ImagingCore`, what about [this type](https://github.com/python-pillow/Pillow/pull/2687#discussion-diff-134087659R1338)?

PIL/Image.py
```diff
@@ -655,6 +695,7 @@ def __getstate__(self):
             self.tobytes()]
 
     def __setstate__(self, state):
+        # type: (List) -> None
```
wiredfool: Pretty sure this is actually (tuple), as this is the pickle interface.

PIL/Image.py
wiredfool: @bjd2385 Documentation is pulled from Image.Image, but the implementations of the file formats inherit from Image -> ImageFile -> JpegImagePlugin.

PIL/Image.py
```diff
@@ -1092,12 +1147,14 @@ def draft(self, mode, size):
...
 
     def _expand(self, xmargin, ymargin=None):
+        # type: (int, Optional[int]) -> Image
         if ymargin is None:
             ymargin = xmargin
         self.load()
         return self._new(self.im.expand(xmargin, ymargin, 0))
 
     def filter(self, filter):
+        # type: (ImageFilter.Filter) -> Image
```
wiredfool: Can be a callable returning a filter as well. (ducktyped at that, that implements filter(ImagingCore))

PIL/Image.py
neiljp: @wiredfool Right, so specifying Tuple is good for `Size`, as it is now.

PIL/Image.py
neiljp: I think the problem I read about was the return type (for some reason), rather than the parameter, but we can certainly look at this again.

PIL/Image.py
neiljp: Ah, yes, I worked backwards from __getstate__ passing out a list :)

neiljp:
> I'd rather not have rebases or force-pushes when there's a lot of discussion going on, as it wipes out the history.

Unfortunately I unthinkingly just did one earlier; ok this time?

>     I'm thinking that the type aliases/imports might best be in a separate module?
> 
> I think that's a reasonable approach, similar to a header file in C.

That's in now, of course, though I'm not sure of the best name.

>     Is this something you think might be merged, down the line?
> 
> I think so. There are really threefour criteria I see:
> 
>     Is it a future maintenance burden? Does it mess up the source files or add tons of effort?
>     Is it complete/correct. (or at least, complete enough to be useful, and what is complete is correct)
>     Are there tests or other (automated) acceptance criteria that will notify us when we break something.
>     Documentation.
> 
> I think I'm comfortable saying that we'll hit 1. The impact to the source doesn't look bad from here, and I think on an incremental basis, it's not going to add too much effort to any individual feature.
:+1: 
> 2 is an ongoing effort.
:+1:
> 3 would be a hook into our testing system. A new test case, e.g. Tests/test_types.py Gate it on typing and mypy being installed, run mypy on any source files, and have a pass / fail based on the results. This needs to fail on a new method in Image.Image with missing types or incorrect types. If we don't have this, the type support is likely to decay.
Right, happy to work on this and possibly compare to eg. Zulip.
> 4 Documentation may not need to be huge, but there should be some.
:+1:

PIL/Image.py
```diff
@@ -1426,6 +1496,7 @@ def alpha_composite(self, im, dest=(0,0), source=(0,0)):
         self.paste(result, box)
 
     def point(self, lut, mode=None):
+        # type: (Union[List, Callable[[Any], Any]], Optional[Mode]) -> Image
```
wiredfool: Also ImagePointHandler. And the Callable should be something like Callable[[Numeric],Numeric], though the two numeric values don't necessarily need to be the same type. (i.e., float->int is possible)

PIL/Image.py
```diff
@@ -1784,6 +1862,7 @@ def rotate(self, angle, resample=NEAREST, expand=0, center=None,
         ]
 
         def transform(x, y, matrix):
+            # type: (float, float, Sequence[float]) -> Tuple[float, float]
```
wiredfool: Doesn't the Matrix here have to be a Tuple due to the expansion?

PIL/Image.py
```diff
@@ -1814,6 +1893,7 @@ def transform(x, y, matrix):
         return self.transform((w, h), AFFINE, matrix, resample)
 
     def save(self, fp, format=None, **params):
+        # type: (Union[Text, pathlib.Path, BytesIO], Optional[Text], **Any) -> None
```
wiredfool: I think this can also be a StringIO.StringIO in py2, but that's a py2 only type.

PIL/Image.py
wiredfool: Probably. It's technically accessible part of the public interface, but as part of the core object it's something that we can change reasonably carefully. 

I'm kind of wondering what became of `im.transform`, since we're just left with `im.transform2`

PIL/Image.py
```diff
@@ -2223,6 +2317,7 @@ def _check_size(size):
 
 
 def new(mode, size, color=0):
+    # type: (Mode, Size, Optional[Union[int, float, Tuple, Text]]) -> Image
```
wiredfool: This union might make sense aliased/named as a Colorspec. (and the Tuple is going to be 1..4 of ints at this point)

PIL/aliases.py
```diff
@@ -0,0 +1,12 @@
...
+
+# Type aliases; names subject to change
+LURD = Tuple[int, int, int, int]  # left, up(per), right, down = x0, y0, x1, y1
+XY = Tuple[int, int]
+Coord = XY
+Size = XY  # NOTE: All XY aliases will be interchangeable
+Matrix4 = Tuple[float, float, float, float]
+Matrix12 = Tuple[float, float, float, float, float, float, float, float, float, float, float, float]
+Mode = str
+Extrema = Union[Tuple[Any, Any], Tuple[Tuple[Any, Any], ...]]  # Any -> float?
```
wiredfool: I think the Extrema can be tuples of either int or floats.  (and potentially nested)

wiredfool: Pixels are going to either be:

Single int/float, of up to 32 bit width, or tuples of 1 to 4 8 bit ints

I'm not sure if the type system can or should distinguish uint8 vs int32 at the python level.

PIL/Image.py
neiljp: @wiredfool I'd actually just made this change, after fixing some imports too; I take it you've seen `_imaging.pyi`?

PIL/Image.py
neiljp: ...and it seems to work :+1:

PIL/Image.py
neiljp: Also that's a `List[float]`?
We can just use float here, since both are possible?

PIL/Image.py
neiljp: a) The code uses a `List` :)
b) I tested with 2.7 and 3.5, and `n=[3,4];(c,d)=n;c;d` gives what you'd hope for.
c) This is only used within `rotate`, so if it passes mypy we're probably good, in terms of defining a public interface at least.

PIL/Image.py
neiljp: Is that as the filename to open or the file object to 'write' to? StringIO is unicode, at least by 2.6/7?

PIL/Image.py
neiljp: * Colorspec is preferable to eg. simply `Color`?
* Are all 'color' uses going to be of that type, eg. for getpixel/putpixel and no doubt others?
* I'm not sure if we gain anything with having `Union[int, float]` as part of that, cf `float`?

Either way, something like:
`Union[int, float, Tuple[int], Tuple[int, int], Tuple[int, int, int], Tuple[int, int, int, int], Text]` 
or (simplified, if the Tuple values are still `int`s?):
`Union[float, Tuple[int], Tuple[int, int], Tuple[int, int, int], Tuple[int, int, int, int], Text]`?

PIL/aliases.py
neiljp: As per above, likely Extrema can be simply of floats?

As above, would this cover it?
`Union[float, Tuple[int], Tuple[int, int], Tuple[int, int, int], Tuple[int, int, int, int], Text]` ?

Is Text a special case for `new` ?

neiljp:
I managed to extract the commits from my rebased branch and they seem to have gone through ok :)

PIL/Image.py
wiredfool: StringIO.StringIO is binary capable in py2 (and removed in py3), but it was replaced with the io.*Io classes. It's the file to write to, and in open, read from. But I have no idea how that's going to interact with mypy, as it's only defined in py2.  This may be another case where it's unlikely that people who are using typing are going to be using StringIO, as they're going to be modernizing other sections of their code, and io.BytesIO is a drop in replacement.

PIL/Image.py
wiredfool: No, not all colors are going to take the text equivalents, though the draw, fill, (and maybe paste) are going to. The pyaccess and putpixel are much more constrained to the underlying type of the image, and getpixel is going to be exactly related to the underlying pixel type.

PIL/aliases.py
wiredfool: In the C layer, Extrema match line 11 above, where Any is either int/float, and each one has to be consistent and match the image type. So, Tuple[int, float] is invalid. Actually, due to a current limitation, multiband images are restricted to being 8 bit ints, so the actual type could be:
```
Extrema = Union[Tuple[float, float], Tuple[int,int], Tuple[Tuple[int,int],...]]
```

For pixels, text is a special case of input. I'd have to check the code to see if the single band float came back in a tuple, but I expect that it does.

PIL/Image.py
```diff
@@ -695,9 +698,9 @@ def __getstate__(self):
             self.tobytes()]
 
     def __setstate__(self, state):
-        # type: (List) -> None
+        # type: (Tuple) -> None
         Image.__init__(self)
-        self.tile = []
+        self.tile = []  # type: List[Tuple] ## FIXME More detail? Should tile be in __init__ & _new?
```
wiredfool: Tile is a messy type, and really would be good to document. `ImageFile` is probably the core of where that happens, along with all of the decoders.

PIL/Image.py
neiljp: I think this is resolved by using `BinaryIO`, which is defined in `typing`; I've come across this before while trying to resolve an issue in Zulip. Also see [stack overflow](https://stackoverflow.com/questions/41748896/the-correct-way-to-annotate-a-file-type-in-python).
I've made this change in an upcoming set of commits.

PIL/aliases.py
neiljp: Let's mark it as that for now :+1:

PIL/Image.py
neiljp: I've added some comments with some notes of potential values, but it's not used in the current round of actual type-checking, within Image.py.

neiljp:
I think I've followed most of the changes except for the Color ones, and added others besides, perhaps most noticeably a huge chunk of the stubs in _imaging.pyi.

Most of the remaining things to tag/discuss in Image.py should now be tagged with 'FIXME TYPING' in a comment, though of course any/all of the typing that's already there can be checked too!

Note the related PR for code changes; I'm keeping this one for just annotations if at all possible, whereas the other has small commits which improve the code and/or allow type inference more easily.

PIL/_imaging.pyi
```diff
@@ -0,0 +1,104 @@
...
+    size = ...  # type: Size  ## close enough?
+    bands = ...  # type: int
+    id = ... ### use? size_t? int?
+    ptr = ... ###
+
+    def getpixel(self, xy: XY) -> Optional[Any]: ...
+    def putpixel(self, xy: XY, value: Any) -> None: ...
+    def pixel_access(self, readonly: int) -> Any: ...
+
+    def convert(self, mode: Mode, dither: int=..., paletteimage: Any=...) -> ImagingCore: ...  ### what is paletteimage?
```
wiredfool: ImagingCore, see `Image.Quantize`

PIL/_imaging.pyi
```diff
@@ -0,0 +1,104 @@
...
+    def convert2(self, im1: ImagingCore, im2: ImagingCore) -> None: ...
+    def convert_matrix(self, mode: Mode, matrix: Union[Matrix4, Matrix12]) -> ImagingCore: ...
+    def convert_transparent(self, mode: Mode, t: Union[Tuple[int, int, int], int]) -> ImagingCore: ...  ### Name t better?
+    def copy(self) -> ImagingCore: ...
+    def crop(self, bbox: LURD) -> ImagingCore: ...
+    def expand(self, xmargin: int, ymargin: int, mode_index: int=0) -> Any: ...  ### Include default?
```
wiredfool: Should be ImagingCore

PIL/_imaging.pyi
```diff
@@ -0,0 +1,104 @@
...
+    def filter(self, size: Size, divisor: float, offset: float, kernel: Any) -> ImagingCore: ...  ### what is kernel?
```
wiredfool: A list of float32, length size[x] * size[y]

PIL/_imaging.pyi
```diff
@@ -0,0 +1,104 @@
...
+    def histogram(self, tuple2: Tuple[float, float]=..., mask: ImagingCore=...) -> List[int]: ...  ### Do these have default values? Do they take None? What is tuple2?
```
wiredfool: Tuple2 is the extrema for the channel that's being processed, It's optional, and either ints or floats.

PIL/_imaging.pyi
```diff
@@ -0,0 +1,104 @@
...
+
+    def modefilter(self, i: int) -> Any: ...
+
+#    def offset(self): ... ### Function removed?
+
+    def paste(self, core: ImagingCore, box: LURD, coremask: ImagingCore) -> None: ...
+    def point(self, lut: Any, mode: Optional[Mode]) -> ImagingCore: ...  ### Clarify Any
```
wiredfool: It's a list of floats or ints.

PIL/_imaging.pyi
```diff
@@ -0,0 +1,104 @@
...
+    def point_transform(self, scale: float=1.0, offset: float=0.0) -> ImagingCore: ...  ### Include defaults or not?
+    def putdata(self, data: Any, scale: float=1.0, offset: float=0.0) -> None: ...  ### Include defaults or not?
+
+    def quantize(self, colors: int=..., method: int=..., kmeans: int=...) -> ImagingCore: ...
+
+    def rankfilter(self, size: int, rank: int) -> ImagingCore: ...
+
+    def resize(self, size: Size, resample: int=...) -> ImagingCore: ...
+    def transpose(self, method: int) -> ImagingCore: ...
+    def transform2(self, box: LURD, core: ImagingCore, method: int, data: Any, resample: int=..., fill: int=...) -> None: ...  ### What is data?
```
wiredfool: A list of float. They're the matrix entries for the transform, n=6 or 8

PIL/_imaging.pyi
```diff
@@ -0,0 +1,104 @@
...
+
+    def isblock(self) -> int: ...
+
+    def getbbox(self) -> Optional[LURD]: ...
+    def getcolors(self, maxcolors: int) -> List[Tuple[int, Any]]: ...  ### Any is a 'Color'?
```
wiredfool: Any is Union[Int, Float, Tuple[int,int], Tuple[int,int,int],Tuple[int,int,int,int]]

PIL/_imaging.pyi
```diff
@@ -0,0 +1,104 @@
...
+    def getextrema(self) -> Tuple[float, float]: ...
+    def getprojection(self) -> Tuple[Any, Any]: ...  ### Need types
```
wiredfool: Tuple[List[int],List[int]] where len(list[i]) = size[i]

PIL/_imaging.pyi
```diff
@@ -0,0 +1,104 @@
...
+
+    def getband(self, band: int) -> ImagingCore: ...
+    def putband(self, image: ImagingCore, band_index: int) -> None: ...
+    def split(self) -> List[ImagingCore]: ...
+    def fillband(self, band: int, color: int) -> None: ...
+
+    def setmode(self, mode: Mode) -> None: ...
+
+    def getpalette(self, mode: Mode="RGB", rawmode: Mode="RGB") -> Any: ... ### returns bytearray?
```
wiredfool: Returns a PyBytes object.

PIL/_imaging.pyi
```diff
@@ -0,0 +1,104 @@
...
+    def getpalettemode(self) -> Mode: ...
+    def putpalette(self, rawmode: Mode, palettes) -> None: ...  ### palettes is a bytesarray?
```
wiredfool: It's a 'y#', so a read only bytes like object.

PIL/_imaging.pyi
```diff
@@ -0,0 +1,104 @@
...
+    def putpalettealpha(self, index: int, alpha: int=...) -> None: ...
+    def putpalettealphas(self, alphas) -> None: ...  ### alphas is a bytesarray?
```
wiredfool: That should be a read only bytes like object as above, but it's actually a read only null capable stringish type.  That's a bug.

PIL/_imaging.pyi
```diff
@@ -0,0 +1,104 @@
...
+def blend(core1: ImagingCore, core2: ImagingCore, alpha: float) -> ImagingCore: ...
+def fill(mode: Mode, size: Size, color: Any) -> Any: ...
+def new(mode: Mode, size: Size) -> ImagingCore: ...
+
+def map_buffer(target_data: Any, size: Size, decoder_name: Text,
+               bbox: Optional[Any], offset: int, args: Tuple[Mode, int, int]
+	      ) -> ImagingCore: ...  ### what are data & bbox?
+
+def crc32(buffer: Any, hilo: Tuple[int, int]=(0,0)) -> Tuple[int, int]: ...
+### buffer is bytesarray? Should Tuple[int, int] be another named type?
```
wiredfool: Buffer is read only byteslike object. I wouldn't worry about that tuple being another named type, I don't think it's use elsewhere.

PIL/_imaging.pyi
```diff
@@ -0,0 +1,104 @@
...
+# List of functions is incomplete
+
+def alpha_composite(core1: ImagingCore, core2: ImagingCore) -> ImagingCore: ...
+def blend(core1: ImagingCore, core2: ImagingCore, alpha: float) -> ImagingCore: ...
+def fill(mode: Mode, size: Size, color: Any) -> Any: ...
+def new(mode: Mode, size: Size) -> ImagingCore: ...
+
+def map_buffer(target_data: Any, size: Size, decoder_name: Text,
+               bbox: Optional[Any], offset: int, args: Tuple[Mode, int, int]
+	      ) -> ImagingCore: ...  ### what are data & bbox?
```
wiredfool: Bbox appears to be unused. 

Target_data is something that exports a python Buffer interface

PIL/_imaging.pyi
```diff
@@ -0,0 +1,104 @@
...
+def effect_noise(size: Size, sigma: float=128) -> ImagingCore: ...
+def linear_gradient(mode: Mode) -> ImagingCore: ...
+def radial_gradient(mode: Mode) -> ImagingCore: ...
+def wedge(mode: Mode) -> ImagingCore: ...
+
+class ImagingCore:
+    mode = ...  # type: str
+    size = ...  # type: Size  ## close enough?
+    bands = ...  # type: int
+    id = ... ### use? size_t? int?
```
wiredfool: It's a PyInt that's a pointer, where ptr is the same thing as a PyCapsule.

PIL/_imaging.pyi
```diff
@@ -0,0 +1,104 @@
...
+                      extent: Tuple[float, float, float, float]=(-3,-2.5,2,2.5),
+		      quality: int=100) -> ImagingCore: ...
+def effect_noise(size: Size, sigma: float=128) -> ImagingCore: ...
+def linear_gradient(mode: Mode) -> ImagingCore: ...
+def radial_gradient(mode: Mode) -> ImagingCore: ...
+def wedge(mode: Mode) -> ImagingCore: ...
+
+class ImagingCore:
+    mode = ...  # type: str
+    size = ...  # type: Size  ## close enough?
```
wiredfool: yep

PIL/Image.py
```diff
@@ -25,18 +25,17 @@
...
     # Just required for typing, or gradual module inclusion while adding annotation?
-    from io import BytesIO
     import pathlib
-#    from . import ImagingPalette  ## This will need a stub?
+#    from . import ImagingPalette  ## FIXME TYPING This will need a stub?
     from . import ImagePalette
     from .ImageFilter import Filter
     from .ImageFile import PyDecoder
-    PyEncoder = Any  ## FIXME: PyEncoder is not defined anywhere? Needs stub/class?
+    PyEncoder = Any  ## FIXME TYPING: PyEncoder is not defined anywhere? Needs stub/class?
```
wiredfool: I think that didn't get merged in because we didn't have an encoder as a sample that was going to use it. I'll dig a bit on that one.

PIL/Image.py
```diff
@@ -782,7 +782,7 @@ def tobitmap(self, name="image"):
             ])
 
     def frombytes(self, data, decoder_name="raw", *args):
-        # type: (Any, Text, *Any) -> None
+        # type: (Any, Text, *Any) -> None  ### FIXME TYPING: What type is data? bytearray?
```
wiredfool: Byteslike object

PIL/Image.py
```diff
@@ -1129,7 +1129,8 @@ def _crop(self, im, box):
 
     def draft(self, mode, size):
         # type: (Mode, Size) -> Any
-        # TODO: return type clarification
+        # FIXME TYPING: return type clarification
+        # FIXME TYPING: thumbnail calls as (None, size), so is it Optional[Mode]?
```
wiredfool: Yes, optional mode. The JPEG one is the only one I see, and it returns self, which is not really consistent with the rest of the API.

PIL/Image.py
```diff
@@ -1597,7 +1598,7 @@ def putalpha(self, alpha):
         self.im.putband(alpha.im, band)
 
     def putdata(self, data, scale=1.0, offset=0.0):
-        # type: (Sequence[Any], float, float) -> None
+        # type: (Sequence[Any], float, float) -> None  ## FIXME TYPING: Sequence[bytes] ?
```
wiredfool: It's a sequence, bytes or int/float for `I` and `F` images

PIL/Image.py
```diff
@@ -1649,6 +1650,7 @@ def putpalette(self, data, rawmode="RGB"):
 
     def putpixel(self, xy, value):
         # type: (Coord, Union[int, Tuple]) -> Any
+        ## FIXME TYPING: ImagingCore.putpixel returns None, so this can too?
```
wiredfool: PyAccess methods don't return anything, so None would be appropriate

PIL/Image.py
```diff
@@ -2130,7 +2132,7 @@ def thumbnail(self, size, resample=BICUBIC):
     # FIXME: the different transform methods need further explanation
     # instead of bloating the method docs, add a separate chapter.
     def transform(self, size, method, data=None, resample=NEAREST, fill=1):
-        # type: (Size, int, Optional[Any], int, Any) -> Image
+        # type: (Size, int, Optional[Any], int, Any) -> Image  ## FIXME TYPING: What are Any here?
```
wiredfool: data -> list[float]

Fill, I think is unused, but remains to not break things.

PIL/Image.py
```diff
@@ -2618,7 +2620,7 @@ def open(fp, mode="r"):
     preinit()
 
     def _open_core(fp, filename, prefix):
-#        # type: (Any, Text, Text): -> Optional[Image]
+#        # type: (Any, Text, Text): -> Optional[Image]  # FIXME TYPING: Is Any BinaryIO here?
```
wiredfool: It's a binary file like object. So, probably.

PIL/Image.py
neiljp: Maybe not urgent, but it's useful to see the types of the 'registries'.

PIL/Image.py
neiljp: So, `bytes`?

Or is this supposed to accept str (python2 only?) and int too?

PIL/Image.py
neiljp: Have made the Optional change, but have left a comment for the Any/return-type based on your response, at least for now.

PIL/Image.py
neiljp: `List[float]` and `int` respectively.

PIL/_imaging.pyi
neiljp: Set as `Any` for now?

PIL/_imaging.pyi
neiljp: Also set mode to `Mode` for now; possibly this should just be `bytes`? It's never going to be unicode, so that's the closest 2+3 type that overlaps?

PIL/_imaging.pyi
neiljp: As I asked elsewhere, 'byteslike' can be `bytes` ?

PIL/_imaging.pyi
neiljp: So `bytearray`, perhaps?

PIL/_imaging.pyi
neiljp: Set to `bytes`.

Bug to be filed separately?

PIL/_imaging.pyi
neiljp: So `bytes`?

PIL/_imaging.pyi
neiljp: So `bytes`?

PIL/Image.py
bjd2385: @wiredfool that's a good point.

PIL/_imaging.pyi
neiljp: I've made this `Union[List[int], List[float]]`, rather than `List[Union[int, float]]`; right?

PIL/_imaging.pyi
neiljp: I've set this to `List[float]`. It's not so important, but is this more accurately 'kerneldata' as a label?

PIL/_imaging.pyi
neiljp: They have default values, but can they be `None`? I'm not completely following the C interface code in all the cases.

PIL/Image.py
neiljp: I've now tried to work on this in the latest 3 commits.

PIL/_imaging.pyi
wiredfool: https://docs.python.org/3.6/c-api/buffer.html 

It's more than bytearray.

neiljp: The only place I could see it used in the code itself was ImageFile.py, after a `mmap`; this suggests more `bytes` (as input, so constant), perhaps, at least in this instance?

Of course, I'm not sure how it's used elsewhere, historically, or otherwise, so this could just be left as `Any`.

PIL/Image.py
```diff
@@ -570,18 +603,21 @@ def close(self):
 
     if sys.version_info >= (3, 4, 0):
         def __del__(self):
+            # type: () -> None
             if (hasattr(self, 'fp') and hasattr(self, '_exclusive_fp')
                and self.fp and self._exclusive_fp):
```
wiredfool: The `fp/_exclusive_fp` stuff could be pulled into ImageFile. It may be more appropriate there from a typing POV, but it's unfortunate that mypy can't infer members based on `hasattr`.

PIL/Image.py
```diff
@@ -570,18 +603,21 @@ def close(self):
```
wiredfool: This line is erroring out with: 
```PIL/Image.py:606: error: Incompatible types in assignment (expression has type "deferred_error", variable has type "Optional[ImagingCore]")```

It's probably fixable, as we can type that as a Union, but there's also:
 
```PIL/Image.py:80: error: Incompatible types in assignment (expression has type "_imaging_not_installed", variable has type Module)```

These are both cases where we're explicitly assigning in an object that returns verbose errors if accessed. I'm not sure how a type system deals with intentional type breakage like this.

File: PIL/Image.py [hunk 51]
```diff
@@ -600,6 +636,7 @@ def _dump(self, file=None, format=None, **options):
         return file
 
     def __eq__(self, other):
+        # type: (object) -> bool
         return (self.__class__.__name__ == other.__class__.__name__ and
                 self.mode == other.mode and
```
wiredfool: We're getting errors here now, that object doesn't have all of these attributes, and we're getting that because the type system isn't picking up that we've narrowed the type of other to an Image with the first comparison.

PIL/Image.py
```diff
@@ -270,6 +285,7 @@ def isImageType(t):
 
 
 def _conv_type_shape(im):
+    # type: (Image) -> Tuple[Size, Tuple[Text, Optional[int]]]  # NOTE: Size inverted compared to normal
```
neiljp: Yes, I worked on this last night and just committed my version to the PR, with the same annotation as you, though I see you changed the code too.

This appears to be resolved through simply using:
```
    typ, extra = _MODE_CONV[im.mode]
    if extra is None:
        return (im.size[1], im.size[0]), typ
    else:
        return (im.size[1], im.size[0], extra), typ
```
I assume this is since mypy cannot cope with tuple addition (or that it's not as clear-cut as it appears). That said, this looks cleaner to me, in terms of tuple structure, and does remove the need for the shape variable.

PIL/Image.py
neiljp: In Zulip the approach appears to be to mark the more dynamic cases with simply `# type: ignore`, often with an additional comment afterwards as to the reason; personally I'd suggest doing the same with use of `Any`, so that later searches can find both 'ignore' and 'Any', and find any reasoning in a nearby location.

PIL/Image.py
neiljp: Right. I know that mypy can use `isinstance` to distinguish in cases like this. Since all the variables that are being tested are in `Image`, would adding `isinstance(other, Image)` before or after the first test be reasonable from a code sense? It does resolve the typing issue, at least here.

PIL/Image.py
neiljp: Again, this might be a case of using `ignore` for now and adding an issue to do some type-inspired tidying ;)

neiljp:
@wiredfool This seems to have reached a semi-reasonable state now, and there's now a conflict. Will commits go in as-is, or squashed? I could rebase to deal with the conflict, but you didn't want to do that?

PIL/_imaging.pyi
bjd2385: Isn't `Optional[Any]` <=> `Union[Any, None]` <=> `Any`?

bjd2385:
I get the following errors and warnings while running `mypy PIL/Image.py -2 --strict-optional --disallow-untyped-defs --ignore-missing-imports --warn-return-any` on this PR.

```bash
PIL/Image.py:596: error: Cannot determine type of 'fp'
PIL/Image.py:607: error: Incompatible types in assignment (expression has type "deferred_error", variable has type "Optional[ImagingCore]")
PIL/Image.py:2484: error: "object" has no attribute "__array_interface__"
PIL/Image.py:2509: error: "object" has no attribute "tobytes"
PIL/Image.py:2511: error: "object" has no attribute "tostring"
PIL/Image.py:2513: error: Argument 3 to "frombuffer" has incompatible type "object"; expected Sequence[str]
PIL/Image.py:2522: warning: Returning Any from function with declared return type "PIL.Image.Image"
PIL/Image.py:2531: warning: Returning Any from function with declared return type "PIL.Image.Image"
```

PIL/_imaging.pyi
```diff
@@ -0,0 +1,91 @@
...
+PILLOW_VERSION = ...
+DEFAULT_STRATEGY = ...
+FILTERED = ...
+HUFFMAN_ONLY = ...
+RLE = ...
+FIXED = ...
+
+# List of functions is incomplete
+
+def alpha_composite(core1: ImagingCore, core2: ImagingCore) -> ImagingCore: ...
```
bjd2385: These `ImagingCore` types need to be enclosed in `''` or `""`, or the type definition (class) `ImagingeCore` needs to be moved to the top of the file.

PIL/_imaging.pyi
```diff
@@ -0,0 +1,91 @@
...
+    def paste(self, core: ImagingCore, box: LURD, coremask: ImagingCore) -> None: ...
+    def point(self, lut: Union[List[int], List[float]], mode: Optional[Mode]) -> ImagingCore: ...
+    def point_transform(self, scale: float=1.0, offset: float=0.0) -> ImagingCore: ...
+    def putdata(self, data: Any, scale: float=1.0, offset: float=0.0) -> None: ...
+    def quantize(self, colors: int=..., method: int=..., kmeans: int=...) -> ImagingCore: ...
+    def rankfilter(self, size: int, rank: int) -> ImagingCore: ...
+    def resize(self, size: Size, resample: int=...) -> ImagingCore: ...
+    def transpose(self, method: int) -> ImagingCore: ...
+    def transform2(self, box: LURD, core: ImagingCore, method: int,
+                   data: list[float], resample: int=..., fill: int=...) -> None: ...
```
bjd2385: This should be `List[float]` (capitalize the first letter).

PIL/Image.py
```diff
@@ -2358,6 +2466,7 @@ def frombuffer(mode, size, data, decoder_name="raw", *args):
 
 
 def fromarray(obj, mode=None):
+    # type: (object, Optional[Mode]) -> Image
```
bjd2385: I think this should be a `np.ndarray` type (i.e. `(np.ndarray,Optional[Mode])->Image`).

PIL/_imaging.pyi
neiljp: Yes for type analysis, but my experience is that including `Any` is just as much a note to look at it later, so `Optional[Any]` is just like `Optional`, but it says we might care about what the `Any` is (or becomes). That's just my take on it, though.

PIL/_imaging.pyi
neiljp: True! Strange; I ran with (and without) --strict and it didn't find this, which I'm surprised by!

PIL/_imaging.pyi
neiljp: Another case for mypy note noticing this, at least here. Which version are you running?

PIL/Image.py
neiljp: This could be reasonable - but that depends:
a) np is not a given prerequisite for PIL[low], right?
b) should it be a prerequisite for type-checking? (checking compatibility)
c) if b, what version?

bjd2385: @neiljp Good points!

PIL/aliases.py
```diff
@@ -0,0 +1,13 @@
+from typing import Tuple, Union
+from typing import Any
```
bjd2385: `Any` isn't used anywhere in this file.

PIL/_imaging.pyi
bjd2385: I don't understand why it runs without errors. In a normal program these types would have to be defined before you could use them, e.g. in the following little example.
```python
try:
    # breaks
    def bar(x: A) -> A:
        print(x)
        return x
except NameError as err:
    print(err)
    def bar(x: 'A') -> 'A':
        print(x)
        return x


class A:
    def __init__(self, val: int) -> int:
        self.val = val

    def __repr__(self) -> str:
        return f'A({self.val})'


# fine
def foo(x: A) -> A:
    print(x)
    return x


if __name__ == '__main__':
    bar(A(3))
    foo(A(2))

# A(3)
# A(2)
```
This isn't necessary in stub files? It could be a bug in PyCharm too; it highlights them as undefined.

PIL/Image.py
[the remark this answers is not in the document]
bjd2385: There are a few things Mypy will miss, and this looks similar to the first example [here](http://mypy.readthedocs.io/en/latest/common_issues.html#no-errors-reported-for-obviously-wrong-code).

PIL/Image.py
bjd2385: Looking over the codebase, I think we could fix this by simply operating on `self.im` within a type's context, so that it never changes type (similar to using a `Maybe` monad in Haskell). It would greatly simplify the type signature so we wouldn't have to use `Optional[ImagingCore,deferred_error]`.

wiredfool: I understand most of those words, but... I'm not sure how to apply it.

PIL/Image.py
bjd2385: @neiljp I think this is something @wiredfool could resolve. As of now, though, `object` should probably be `Any`.

wiredfool: Numpy is certainly an optional dependency for testing, we skip the appropriate tests if it's not installed. This is a small stable interface to numpy, I'm not aware of it changing in the time that we've worked with it. So the exact version isn't terribly important.

PIL/Image.py
wiredfool: The class checking was added here: https://github.com/python-pillow/Pillow/commit/1aee9bfdef1d73007d8e35ab883c8c936c1da602  I'm not clear if we're checking class names here because an isInstance would be a performance hit here or if this is just a complicated way of doing the class check.

wiredfool:
Go ahead and rebase it. Most of the current comments are hidden by later code revisions anyway.

PIL/Image.py
neiljp: I've found the following, but haven't yet tested with it:
https://github.com/machinalis/mypy-data/tree/master/numpy-mypy

PIL/Image.py
neiljp: I see you merged this approach in my other PR?

The difference is maybe that it doesn't do an exact class match, but rather based on the base class? So derived classes will match with `isinstance`. Depending on the perceived slowdown/importance, as long as there is an isinstance check above the tests which use Image attributes, it should be fine (and mypy shouldn't then complain).

neiljp:
This seems to be the status at the moment; for reference, I'm now running with just '--strict' and then using `|grep Image.py|grep -v "Call to untyped function"|grep -v warning`. With some more changes to the code, which I have in a separate branch, I'm down to about 190 errors. That perhaps sounds like a lot, but a *significant* number of the errors can be traced back to `Image.im` being `Optional`, and the lack of explicit checks (`if` statements, or assertions) to avoid `None`. Possibly we can just add a `type: Any` for now? (that reduces the outstanding amount to 120 for me)

Other outstanding issues include: (and some won't appear in a less-strict mode)
* Some names not being defined
* Some modules not being accessible
* The module/_imaging_not_installed Union issue
* mypy not identifying the custom string type functions as distinguishing types
* Some remaining issues with the correct types (see requests for feedback in Image.py)
* Clarification on what to do with numpy-related functions
* Lack of _exclusive_fp in Image itself (discussed briefly?)
* hasattr not being supported by mypy (?)

aliases.py:
* It's unclear if all types are correct?
* Maybe add (and use elsewhere) a 'Band' type?

_imaging.pyi: (from _imaging.c)
* Need to reorder functions and ImagingCore? (mypy bug?)
* Many/most functions have types, but some can still be improved.
* Consistency in function parameter names could be better
* Should the default values go in the stubs?
* Do some parameters just have default parameters, or are some Optional?

Hopefully that provides a semi-complete list that we can work forward from. The rebase has nothing adjusted except for the rebase off master and some extra commits, though there are quite a lot of commits there, so I can squash some together if that makes it easier.

Would you like the extra code changes in a new PR/branch, or onto the same branch as the others?

bjd2385:
Out of curiosity, what is holding up a merge of this PR at this point? I'm still becoming familiar with GitHub and making contributions.

neiljp:
@radarhere @wiredfool 
I appreciate this is likely not mergeable as it stands, but this has been quiet since the rebase so just checking in to see how you feel about this PR.

wiredfool:
Looknig at the criteria here: https://github.com/python-pillow/Pillow/pull/2687#issuecomment-323544730
> 1.  Is it a future maintenance burden? Does it mess up the source files or add tons of effort?
2. Is it complete/correct. (or at least, complete enough to be useful, and what is complete is correct)
3. Are there tests or other (automated) acceptance criteria that will notify us when we break something.
4. Documentation.

Right now I think we're missing on 2 and 3. When I look at the output of `mypy PIL/Image.py -2`, I'm seeing a ton of output, and it appears to be more along the lines of issues in the typing system rather than underlying issues with the code. Until that output is managable, then I'm not sure how we'd either:
* ensure that nothing is broken on changes
* have useful output for consumers of the library. 

Some of this can be traced to the typing system having trouble with dynamic usage, where either we're dispatching off of a type because we're flexible about the item passed in (e.g. filename or open file like object) or we're ducktyping and it's not picking that up. But for it to be useful, we've got to find a way to deal with it.

wiredfool:
https://github.com/wiredfool/Pillow/tree/pr_2687 has a couple more commits to move closer.

wiredfool:
Potential item of interest: http://mypy-lang.blogspot.ie/2017/11/dropbox-releases-pyannotate-auto.html

wiredfool:
https://github.com/wiredfool/Pillow/tree/pr_2687 now passes `mypy --ignore-missing-imports -2 src/PIL/Image.py` with no errors with the latest released mypy as of this weekend. 

Getting to this point took more effort than I would have liked -- including some 'single static type' style rewriting and some selective ignoring of constructs that are difficult.  Python's ducktyping does not play well with it's type system, in that there are many cases where an attribute is checked, and then that constraint is ignored in the then/else clause. There are also lots of places where a 2 or 4-tuple of ints is checked for length, and then interpreted as an Tuple[Any,...].  Finally, I'm unclear if it's even possible to have a type that is 'SupportsBufferInterface' because as far as I can tell, that's a C api level attribute. 

Never the less, that branch now passes the tests, and in theory can remain that way. I'm not sure that it's entirely right, I'm not sure that it's the cleanest way to do it, but it does work.

hugovk:
See https://github.com/python-pillow/Pillow/pull/2941 for a continuation of this.

PR #2689, opened 2017-08-19
Added tests

radarhere:


Tests/test_file_icns.py
```diff
@@ -82,6 +82,23 @@ def test_jp2_icon(self):
...
+            im = icns_file.getimage()
+            self.assertEqual(im.mode, "RGBA")
+            self.assertEqual(im.size, (1024, 1024))
+
+            im = icns_file.getimage((512, 512))
+            self.assertEqual(im.mode, "RGBA")
+            self.assertEqual(im.size, (512, 512))
+
+    def test_not_an_icns_file(self):
+        with open('Tests/images/invalid.icns', 'rb') as fp:
```
homm: I'd prefer `io.BytesIO(b'invalid')` for such simple cases

Tests/test_file_icns.py
```diff
@@ -82,6 +82,23 @@ def test_jp2_icon(self):
...
+            self.assertRaises(SyntaxError,
+                lambda: IcnsImagePlugin.IcnsFile(fp))
```
homm: assertRaises could be used without lambdas:

```python
self.assertRaises(SyntaxError, IcnsImagePlugin.IcnsFile, fp)
```

I find such format cleaner.

PIL/WmfImagePlugin.py
```diff
@@ -153,7 +153,7 @@ def _load(self):
 
 
 def _save(im, fp, filename):
-    if _handler is None or not hasattr("_handler", "save"):
+    if _handler is None or not hasattr(_handler, "save"):
```
homm: OMG )

Tests/test_imagedraw.py
```diff
@@ -436,18 +437,22 @@ def test_square(self):
         expected.load()
         img, draw = self.create_base_image_draw((10, 10))
         draw.polygon([(2, 2), (2, 7), (7, 7), (7, 2)], BLACK)
+        del draw
```
homm: Why is it important?

PIL/WmfImagePlugin.py
wiredfool: You know something when you fix an "it could never work" bug that dates back to PIL.

wiredfool:
LGTM, pending @homm's comments.

Tests/test_imagedraw.py
radarhere: It's not. I just thought it would be nice to clean up after the test code. I can remove the commit if you like.

Tests/test_file_icns.py
radarhere: Okay, I've updated the commit.

Tests/test_file_icns.py
radarhere: Okay, I've added a commit.

Tests/test_imagedraw.py
wiredfool: They're just going to go out of scope at the end of the method, like the image.

radarhere: Okay, I've removed the commit.

PIL/ImageGrab.py
```diff
@@ -72,7 +72,6 @@ def grabclipboard():
         os.unlink(filepath)
         return im
     else:
-        debug = 0  # temporary interface
         data = Image.core.grabclipboard(debug)
```
homm: It's not clear for me. `debug` var is never defined but is still used there. Possible should be `Image.core.grabclipboard(0)`.

Tests/test_image.py
```diff
@@ -70,6 +70,10 @@ def test_invalid_image(self):
             im = io.BytesIO(b'')
         self.assertRaises(IOError, lambda: Image.open(im))
 
+    def test_bad_mode(self):
+        im = hopper()
+        self.assertRaises(ValueError, lambda: Image.open(im, "bad mode"))
```
homm: `im` there is `PIL.Image.Image` object. Does `Image.open(im)` make any sense?
I believe it should be something like:
```python
def test_bad_mode(self):
    self.assertRaises(ValueError, Image.open, "filename", "bad mode")
```

PIL/ImageGrab.py
radarhere: Thanks for catching that! I've removed this change from here, and created PR #2708

Tests/test_image.py
radarhere: Okay, I've updated the PR.

PR #2690, opened 2017-08-20
[WIP] Code alteration supplementing annotation in #2687

neiljp:
Changes proposed in this pull request: (so far)

 * Renaming a loop variable, which should be clearer, and ensures it isn't reused for a different purpose (and different type)
 * Rewriting _dump to first get a good filename, and then use it; again the idea is to ensure variables have a known (more limited, often non-None) type.

These supplement #2687, but I'm keeping the code changes separate as they potentially change behaviour, whereas annotation should not.

### DIFF
--- a/PIL/Image.py
+++ b/PIL/Image.py
@@ -605,7 +605,8 @@ class Image(object):
         return filename
 
     def __eq__(self, other):
-        return (self.__class__.__name__ == other.__class__.__name__ and
+        return (isinstance(other, Image) and
+                self.__class__.__name__ == other.__class__.__name__ and
                 self.mode == other.mode and
                 self.size == other.size and
                 self.info == other.info and
--- a/PIL/Image.py
+++ b/PIL/Image.py
@@ -1791,9 +1791,13 @@ class Image(object):
         w, h = self.size
 
         if translate is None:
-            translate = [0, 0]
+            post_trans = (0, 0)
+        else:
+            post_trans = translate
         if center is None:
-            center = [w / 2.0, h / 2.0]
+            rotn_center = (w / 2.0, h / 2.0)  # FIXME These should be rounded to ints?
+        else:
+            rotn_center = center
 
         angle = - math.radians(angle)
         matrix = [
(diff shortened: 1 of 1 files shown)


PIL/Image.py
```diff
@@ -583,21 +583,27 @@ def _copy(self):
 
     def _dump(self, file=None, format=None, **options):
```
hugovk: Maybe the builtin file parameter can also be renamed? The underscore indicates this is an internal function, so no one should be using this. Then line 595 can be removed.

PIL/Image.py
```diff
@@ -1697,7 +1705,7 @@ def resize(self, size, resample=NEAREST):
 
         self.load()
 
-        size = tuple(size)
+        size = (size[0], size[1])  # Truncate input into 2-tuple
```
wiredfool: This should probably just be a call to _check_size.

PIL/Image.py
neiljp: This wouldn't resolve the type-related issue, which is that `file` is effectively an `Optional[Text]` (`Union[None, Text]`).

Assigning to it isn't treated (by mypy) as removing the `None`-ness, so use of the variable later assumes it still may be `None`. Alternatively the code can be enclosed within a conditional checking for `None`-ness, but here that would probably be a duplicating of the subsequent code since format would need to be checked in each case.

PIL/Image.py
neiljp: I wasn't sure exactly what was happening here, but mypy was concerned that a parameter annotated as a `Tuple[int, int]` became a less-defined general tuple. This seemed like a more explicit way of saying that size must be a 2-tuple at run-time.

Should functions accepting something which is a size, all be calling _check_size()? Should I amend this/these commits to have _check_size(), add another, or is it good as-is? (and maybe handle general _check_size(size) adding in a separate PR?)

neiljp:
Happy to rebase; was avoiding doing so since in the related PR @wiredfool was indicating to avoiding doing so.

neiljp:
Rebased and force-pushed. Dropped the tuple fix in resize, as that code changed and I need to look at it again.

wiredfool:
Merging this, there may still be more to do but this is at a good enough place.

PR #2692, opened 2017-08-21
Load 16-bit RGB and RGBA TIFF image

homm:


PIL/TiffImagePlugin.py
```diff
@@ -201,6 +201,17 @@
     (II, 2, (1,), 1, (8, 8, 8, 8), (999,)): ("RGBA", "RGBA"),  # Corel Draw 10
     (MM, 2, (1,), 1, (8, 8, 8, 8), (999,)): ("RGBA", "RGBA"),  # Corel Draw 10
 
+    (II, 2, (1,), 1, (16, 16, 16), ()): ("RGB", "RGB;16L"),
```
homm: I know, that L means line interleave according to current mode syntax:

> raw mode syntax is "<mode>;<bits><flags>" where "bits" defaults depending on mode (1 for "1", 8 for "P" and "L", etc), and "flags" should be given in alphabetical order.  if both bits and flags have their default values, the ; should be left out
> flags: "I" inverted data; "R" reversed bit order; "B" big endian byte order (default is little endian); "L" line interleave, "S" signed, "F" floating point

But there is already the confusion with "bits": `RGB;16B` means 16-bit per channel big endian, while `RGB;16` means 16 bits per pixel. So, without renaming `RGB;16B` to `RGB;48B`, `RGB;16L` is the best name for this mode, because the mode `16 bits per pixel line interleave` is not possible.

PIL/TiffImagePlugin.py
```diff
@@ -230,7 +241,12 @@
     (MM, 8, (1,), 1, (8, 8, 8), ()): ("LAB", "LAB"),
 }
 
-PREFIXES = [b"MM\000\052", b"II\052\000", b"II\xBC\000"]
+PREFIXES = [
+    b"MM\x00\x2A",  # Valid TIFF header with big-endian byte order
+    b"II\x2A\x00",  # Valid TIFF header with little-endian byte order
+    b"MM\x2A\x00",  # Invalid TIFF header, assume big-endian
+    b"II\x00\x2A",  # Invalid TIFF header, assume little-endian
```
homm: According to [the specs](https://www.itu.int/itudoc/itu-t/com16/tiff-fx/docs/tiff6.pdf) (page 13), there are only two valid headers: `"MM\x00*"` and `"II*\x00"`. But I met files with wrong headers, where second two bytes were not aligned according to the reported ending. However, these files were successfully opened by other viewers and were absolutely correct in anything else.

As of `"II\xBC\x00"` prefix, I have no Idea what does it mean. I haven't find it in specs or anywhere else in web. It was there from original PIL and I don't know original motivation of adding it.

homm:
[https://uploadcare.com/widget/configure/2.x/ отдает версию 3, а не последнюю 2](https://trello.com/c/swN6v7JG/36-https-uploadcarecom-widget-configure-2x-%D0%BE%D1%82%D0%B4%D0%B0%D0%B5%D1%82-%D0%B2%D0%B5%D1%80%D1%81%D0%B8%D1%8E-3-%D0%B0-%D0%BD%D0%B5-%D0%BF%D0%BE%D1%81%D0%BB%D0%B5%D0%B4%D0%BD%D1%8E%D1%8E-2)

homm:
[[3] Pillow: Открывать странные битые тиффы](https://trello.com/c/zNN5taNL/167-3-pillow-%D0%BE%D1%82%D0%BA%D1%80%D1%8B%D0%B2%D0%B0%D1%82%D1%8C-%D1%81%D1%82%D1%80%D0%B0%D0%BD%D0%BD%D1%8B%D0%B5-%D0%B1%D0%B8%D1%82%D1%8B%D0%B5-%D1%82%D0%B8%D1%84%D1%84%D1%8B)

homm:
Should be rebased after #2694 merge

homm:
Tests for this PR are strictly depend on #2694, so I merged it here. After merging #2694 it should become smaller again.

PIL/TiffImagePlugin.py
wiredfool: I'm ok with this.

wiredfool:
Ok, pending #2694. 

Looking at these PRs has gotten me thinking about proper wide pixel multiband support, and I'm thinking that the best solution might be storing the image in separate planes as an array of single band images of mode L/I/F.  All of the image ops/kernels would pretty much work out of the box on the individual channels. It would just be pack/unpack/shuffle that would need low level changes to multiplex from the single stream on the decoder/encoder to the multiple planes.

PR #2694, opened 2017-08-22
Packing tests

homm:
Decided to implement all and improved tests for packing and unpacking.

Added many missed tests. At the end all possible modes should be covered. Most tests improved, for example with line interleave.

### DIFF
--- a/Tests/test_lib_pack.py
+++ b/Tests/test_lib_pack.py
@@ -71,73 +71,6 @@ class TestLibPack(PillowTestCase):
 
             return im.getpixel((0, 0))
 
-        def unpack_1(mode, rawmode, value):
-            assert mode == "1"
-            im = None
-
-            if py3:
-                im = Image.frombytes(
-                    mode, (8, 1), bytes([value]), "raw", rawmode, 0, 1)
-            else:
-                im = Image.frombytes(
-                    mode, (8, 1), chr(value), "raw", rawmode, 0, 1)
-
-            return tuple(im.getdata())
-
-        X = 255
-
-        self.assertEqual(unpack_1("1", "1", 1), (0, 0, 0, 0, 0, 0, 0, X))
-        self.assertEqual(unpack_1("1", "1;I", 1), (X, X, X, X, X, X, X, 0))
-        self.assertEqual(unpack_1("1", "1;R", 1), (X, 0, 0, 0, 0, 0, 0, 0))
-        self.assertEqual(unpack_1("1", "1;IR", 1), (0, X, X, X, X, X, X, X))
-
-        self.assertEqual(unpack_1("1", "1", 170), (X, 0, X, 0, X, 0, X, 0))
-        self.assertEqual(unpack_1("1", "1;I", 170), (0, X, 0, X, 0, X, 0, X))
-        self.assertEqual(unpack_1("1", "1;R", 170), (0, X, 0, X, 0, X, 0, X))
-        self.assertEqual(unpack_1("1", "1;IR", 170), (X, 0, X, 0, X, 0, X, 0))
-
-        self.assertEqual(unpack("L", "L;2", 1), 0)
-        self.assertEqual(unpack("L", "L;4", 1), 0)
-        self.assertEqual(unpack("L", "L", 1), 1)
-        self.assertEqual(unpack("L", "L;I", 1), 254)
-        self.assertEqual(unpack("L", "L;R", 1), 128)
-        self.assertEqual(unpack("L", "L;16", 2), 2)  # little endian
-        self.assertEqual(unpack("L", "L;16B", 2), 1)  # big endian
-
-        self.assertEqual(unpack("LA", "LA", 2), (1, 2))
-        self.assertEqual(unpack("LA", "LA;L", 2), (1, 2))
-
-        self.assertEqual(unpack("RGB", "RGB", 3), (1, 2, 3))
-        self.assertEqual(unpack("RGB", "RGB;L", 3), (1, 2, 3))
-        self.assertEqual(unpack("RGB", "RGB;R", 3), (128, 64, 192))
-        self.assertEqual(unpack("RGB", "RGB;16B", 6), (1, 3, 5))  # ?
-        self.assertEqual(unpack("RGB", "BGR", 3), (3, 2, 1))
-        self.assertEqual(unpack("RGB", "RGB;15", 2), (8, 131, 0))
-        self.assertEqual(unpack("RGB", "BGR;15", 2), (0, 131, 8))
-        self.assertEqual(unpack("RGB", "RGB;16", 2), (8, 64, 0))
-        self.assertEqual(unpack("RGB", "BGR;16", 2), (0, 64, 8))
-        self.assertEqual(unpack("RGB", "RGB;4B", 2), (17, 0, 34))
-
-        self.assertEqual(unpack("RGB", "RGBX", 4), (1, 2, 3))
-        self.assertEqual(unpack("RGB", "BGRX", 4), (3, 2, 1))
-        self.assertEqual(unpack("RGB", "XRGB", 4), (2, 3, 4))
-        self.assertEqual(unpack("RGB", "XBGR", 4), (4, 3, 2))
-
-        self.assertEqual(unpack("RGBA", "RGBA", 4), (1, 2, 3, 4))
-        self.assertEqual(unpack("RGBA", "RGBa", 4), (63, 127, 191, 4))
-        self.assertEqual(unpack("RGBA", "BGRA", 4), (3, 2, 1, 4))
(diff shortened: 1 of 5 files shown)


wiredfool:
From the bigendian machine:
```
FAIL: TestLibUnpack.test_F_float
  File "/home/erics/Pillow/Tests/test_lib_pack.py", line 318, in test_F_float
    1.539989614439558e-36, 4.063216068939723e-34)
  File "/home/erics/Pillow/Tests/test_lib_pack.py", line 78, in assert_unpack
    self.assertEqual(pixel, im.getpixel((x, 0)))
AssertionError: 1.539989614439558e-36 != 2.387939260590663e-38

FAIL: TestLibUnpack.test_I
  File "/home/erics/Pillow/Tests/test_lib_pack.py", line 254, in test_I
    self.assert_unpack("I", "I", 4, 0x04030201, 0x08070605)
  File "/home/erics/Pillow/Tests/test_lib_pack.py", line 78, in assert_unpack
    self.assertEqual(pixel, im.getpixel((x, 0)))
AssertionError: 67305985 != 16909060

FAIL: TestLibUnpack.test_I16
  File "/home/erics/Pillow/Tests/test_lib_pack.py", line 347, in test_I16
    self.assert_unpack("I;16", "I;16N", 2, 0x0201, 0x0403, 0x0605)
  File "/home/erics/Pillow/Tests/test_lib_pack.py", line 78, in assert_unpack
    self.assertEqual(pixel, im.getpixel((x, 0)))
AssertionError: 513 != 258
```

homm:
It's interesting. I've fixed `test_I16`, it's my fault. The problem with `test_F_float` and `test_I`, because in both cases just `copy4` is used. So one of the following is correct: rawmode `F` and `I` actually means `F;32N` and `I;32N` (native) **or** we should replace `copy4` with `unpackI32` unpacker.

Could you test the latest commit?

wiredfool:
I'm getting this now:
```
FAIL: TestImageArray.test_fromarray
  File "/home/erics/Pillow/Tests/test_image_array.py", line 50, in test_fromarray
    self.assertEqual(test("I"), ("I", (128, 100), True))
AssertionError: Tuples differ: ('I', (128, 100), False) != ('I', (128, 100), True)

First differing element 2:
False
True

- ('I', (128, 100), False)
?                   ^^^^

+ ('I', (128, 100), True)
?                   ^^^
```

There's also an ongoing error that's also in the latest releases:
```
FAIL: TestImageFile.test_parser
  File "/home/erics/Pillow/Tests/test_imagefile.py", line 66, in test_parser
    self.assert_image_similar(im1, im2.convert('L'), 20)
  File "/home/erics/Pillow/Tests/helper.py", line 110, in assert_image_similar
    ave_diff, epsilon))
AssertionError:  average pixel value difference 84.7002 > epsilon 20.0000
```

homm:
@wiredfool What do you think about this in general?

> So one of the following is correct: rawmode `F` and `I` actually means `F;32N` and `I;32N` (native) **or** we should replace `copy4` with `unpackI32` unpacker.

Should be consider `F` as `F;32N` and add at least comments, or fix it and continue investigate why tests failed?

wiredfool:
I think that for any sort of resampling or image ops, it would be completely goofy to store non-native endian in `I` or `F` storage. It has to be native mode.

Historically, the native versions were added for integer modes because of libtiff, which returns multibyte pixels in the native format, so it's just a copy operation to add it to storage. Our tiff implementation needed to decode the endianness to get it right.

homm:
> I think that for any sort of resampling or image ops, it would be completely goofy to store non-native endian in I or F storage.

For sure. In Pillow objects we store all data only in native order. The question is how to interpret the data from external sources in 'F' and 'I' rawmodes. There are two sources which can contain ramodes:

* in our file decoders. In this case than rawmodes definitely means non-native endian. As the vast majority of systems is little endian, it should be.
* in user code with `Image.frombytes()`. Users can get data arrays from numpy, for example, and apply them with `Image.frombytes()`. Depending on numpy behavior, user can expect that we will decode bytes as native endings.

For me the best solution is align those modes with current rawmode format (where "[default is little endian](https://github.com/python-pillow/Pillow/blob/master/libImaging/Unpack.c#L1176)") and add appropriate section in Release Notes.

I've installed QEMU with powerpc architecture, it is damn slow but works. I see the error in `TestImageArray.test_fromarray`, but I have no errors in `TestImageFile.test_parser`.

Tests/test_lib_pack.py
```diff
@@ -23,135 +24,412 @@ def pack(mode, rawmode, in_data=(1, 2, 3, 4)):
...
-
-        self.assertEqual(unpack("CMYK", "CMYK", 4), (1, 2, 3, 4))
-        self.assertEqual(unpack("CMYK", "CMYK;I", 4), (254, 253, 252, 251))
-
-        self.assertRaises(ValueError, lambda: unpack("L", "L", 0))
-        self.assertRaises(ValueError, lambda: unpack("RGB", "RGB", 2))
-        self.assertRaises(ValueError, lambda: unpack("CMYK", "CMYK", 2))
+    def assert_pack(self, mode, rawmode, data, *pixels):
+        """
+        data - eather raw bytes with data or just number of bytes in rawmode.
```
radarhere: either

Tests/test_lib_pack.py
```diff
@@ -23,135 +24,412 @@ def pack(mode, rawmode, in_data=(1, 2, 3, 4)):
...
+        self.assert_pack("RGBX", "R", 1, (1,0,8,9), (2,0,8,9), (3,0,8,0))
+        self.assert_pack("RGBX", "G", 1, (6,1,8,9), (6,2,8,9), (6,3,8,9))
+        self.assert_pack("RGBX", "B", 1, (6,7,1,9), (6,7,2,0), (6,7,3,9))
+        self.assert_pack("RGBX", "X", 1, (6,7,0,1), (6,7,0,2), (0,7,0,3))
+
+
+class TestLibUnpack(PillowTestCase):
+    def assert_unpack(self, mode, rawmode, data, *pixels):
+        """
+        data - eather raw bytes with data or just number of bytes in rawmode.
```
radarhere: either

homm:
Changed packers and unpackers for `I` and `F` modes to platform-independent format. All tests passed on my big-endian qemu-ppc system.

wiredfool:
Ok, mostly afk this weekend, but will check on mon.

wiredfool:
(fwiw, my mini is at ~ 70 seconds for a test run, which is roughly equivalent to a raspberry pi)

Ok, I get your point now. 

We should not change the meaning of the `I` and `F` modes now from `copy4` to byteswap. This will break anyone on PPC who's using `I` or `F` images.  It does trip one test on my PPC which you're probably not hitting:

```
FAIL: TestNumpy.test_to_array
  File "/home/erics/Pillow/Tests/test_numpy.py", line 185, in test_to_array
    _to_array(*mode)
  File "/home/erics/Pillow/Tests/test_numpy.py", line 166, in _to_array
    self._test_img_equals_nparray(img, np_img)
  File "/home/erics/Pillow/Tests/test_numpy.py", line 122, in _test_img_equals_nparray
    self.assert_deep_equal(px[x, y], np[y, x])
  File "/home/erics/Pillow/Tests/helper.py", line 66, in assert_deep_equal
    self.assertEqual(a, b, msg)
AssertionError: 26 != 436207616
```

re:  "default is little endian" comment

There's something of a disconnect between the storage modes, which are limited in number and expressiveness and fixed to native endian and the pack/unpack raw modes, where the flagged details default to LE if not specified. The disconnect comes when there's identically named storage and raw modes with conflicting defaults. 

There _are_ modes that do what you're proposing, `I;32` and `F;32` that are the default LE versions of the unpackers corresponding to the `F;32B` and `I;32B` modes.

homm:
> It does trip one test on my PPC which you're probably not hitting

You are right, I didn't install numpy.

> We should not change the meaning of the I and F modes now from copy4 to byteswap.

Ok, reverted. Could you check again?

Tests/test_lib_pack.py
```diff
@@ -146,31 +146,34 @@ def test_HSV(self):
...
 
     def test_I(self):
-        self.assert_pack("I", "I", 4, 0x04030201, 0x08070605)
         self.assert_pack("I", "I;16B", 2, 0x0102, 0x0304)
         self.assert_pack("I", "I;32S",
             b'\x83\x00\x00\x01\x01\x00\x00\x83',
             0x01000083, -2097151999)
 
         if sys.byteorder == 'little':
+            self.assert_pack("I", "I", 4, 0x04030201, 0x08070605)
```
wiredfool: I think you missed moving the unpack version of these tests.

homm: Indeed. Done.

wiredfool:
Ok, tests are passing here.

PR #2698, opened 2017-08-27
Fix sampling factor for subsampling=2 while JPEG saving

### LINKED ISSUE #2063: jpeg sampling factor is wrong
`2x2` is `4:2:0`, [not `4:1:1`](https://github.com/python-pillow/Pillow/blob/master/libImaging/JpegEncode.c#L197). This is wrong in many places, most importantly [here](https://github.com/python-pillow/Pillow/blob/master/PIL/JpegImagePlugin.py#L615).

Documented [here](https://github.com/libjpeg-turbo/libjpeg-turbo/blob/master/turbojpeg.h#L106-L107), and elsewhere.

May I also suggest to add the `NxN` representation to the docs, which is easier to comprehend.

homm:
Fixes #2063

`.save(subsampling=2)` actually sets sampling factor `4:2:0`. As of `.save(subsampling="4:1:1")` is just alias to `.save(subsampling=2)`, it also meant `4:2:0`.

Fixed all documentation and comments to correct sampling factor.
Allow new `.save(subsampling="4:2:0")` value which actually means `4:2:0`.
Leave old `.save(subsampling="4:1:1")` value as alias to `4:2:0`.

### DIFF
--- a/PIL/JpegImagePlugin.py
+++ b/PIL/JpegImagePlugin.py
@@ -635,7 +635,11 @@ def _save(im, fp, filename):
         subsampling = 0
     elif subsampling == "4:2:2":
         subsampling = 1
+    elif subsampling == "4:2:0":
+        subsampling = 2
     elif subsampling == "4:1:1":
+        # For compatibility. Before Pillow 4.3, 4:1:1 actually meant 4:2:0.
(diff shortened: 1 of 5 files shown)


homm:
This doesn't affect current users, only documentation and comments in code.

PR #2699, opened 2017-08-28
Convert CMYK to RGB like Google Chrome

homm:
There is no easy way to convert CMYK to RGB without complicated color management. But there are some approximations. Pillow used to use the simplest one, which, for example, could be [found there](http://www.rapidtables.com/convert/color/rgb-to-cmyk.htm). It is not too accurate and produces very dark images. On the other hand, I found [this formulas](https://www.easyrgb.com/en/math.php) which produces far better result.

```
C = ( C * ( 1 - K ) + K )
M = ( M * ( 1 - K ) + K )
Y = ( Y * ( 1 - K ) + K )

R = ( 1 - C ) * 255
G = ( 1 - M ) * 255
B = ( 1 - Y ) * 255
```
or in one equation:
```
R = 255 - K - C * ( 255 - K ) / 255
G = 255 - K - M * ( 255 - K ) / 255
B = 255 - K - Y * ( 255 - K ) / 255
```

Original CMYK image opened in Safari (and probably how it should looks like with correct color management)
_out_cmyk

Converted to RGB using the new formula and also CMYK image looks absolutely the same in Google Chrome
_out_new

Converted to RGB using the old formula
_out_old

### DIFF
--- a/libImaging/Convert.c
+++ b/libImaging/Convert.c
@@ -536,12 +536,15 @@ rgb2cmyk(UINT8* out, const UINT8* in, int xsize)
 static void
 cmyk2rgb(UINT8* out, const UINT8* in, int xsize)
 {
-    int x;
-    for (x = 0; x < xsize; x++, in += 4) {
-        *out++ = CLIP(255 - (in[0] + in[3]));
-        *out++ = CLIP(255 - (in[1] + in[3]));
-        *out++ = CLIP(255 - (in[2] + in[3]));
-        *out++ = 255;
+    int x, nk, tmp;
+    for (x = 0; x < xsize; x++) {
+        nk = 255 - in[3];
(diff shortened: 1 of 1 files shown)


homm:
[Setfill operation issue](https://trello.com/c/bGNXhY6Z/89-setfill-operation-issue)

homm:
Another example:

_out_cmyk _out_new _out_old

znerol:
I'm all for better CMYK handling. Wouldn't it be easier to better to simply read the icc profile from the image and then use `ImageCms.profileToProfile` if little CMS is available?

wiredfool:
I think doing things intelligently when we have lcms and image profiles is a good change in general, but it's going to be intrusive and a noticeable change in behavior for people relying on the library. 

This PR makes sense in absence of that.

PR #2703, opened 2017-08-31
Fix comment about palette class in quantize method

kkopachev:
Changes documentation to reflect correct expected class

### DIFF
--- a/PIL/Image.py
+++ b/PIL/Image.py
@@ -983,7 +983,7 @@ class Image(object):
                        2 = fast octree
                        3 = libimagequant
         :param kmeans: Integer
-        :param palette: Quantize to the :py:class:`PIL.ImagingPalette` palette.
+        :param palette: Quantize to the palette of given :py:class:`PIL.Image.Image`.
         :returns: A new image
 
         """


wiredfool:
Palette is an image, see the remainder of the method:

```
        if palette:
             # use palette from reference image
             palette.load()
             if palette.mode != "P":
                 raise ValueError("bad mode for palette image")
             if self.mode != "RGB" and self.mode != "L":
                 raise ValueError(
                     "only RGB or L mode images can be quantized to a palette"
                     )
             im = self.im.convert("P", 1, palette.im)
             return self._new(im)
```

kkopachev:
@wiredfool My attempt was to change doc so it says that palette is an image.
With current docs I was under impression that I have to pass `PIL.ImagingPalette` as an argument, which supposed to be extracted from `im.palette`. However, `quantize` wants `PIL.Image` object woth attached palette, not palette object itself. 
Do you see my intent here?

wiredfool:
Sorry, I got that exactly backwards.

wiredfool:
Thanks for catching this.

PR #2707, opened 2017-08-31
Fix resize with box when no resize required

homm:
I found many errors with box argument when x0 & y0 coordinates are 0. In all following cases result resolution should be 400x400:

```python
>>> im.size
<<< (1034, 1280)

>>> im.resize((400, 400), Image.LANCZOS, (0, 0, 400, 400)).size
<<< (1034, 1280)

>>> im.resize((400, 400), Image.LANCZOS, (0, 0, 400, 1024)).size
<<< (1034, 400)

>>> im.resize((400, 400), Image.LANCZOS, (0, 0, 1024, 400)).size
<<< (400, 403)
```

I'm opening PR just to let know what the problem exists and I'm working on the fix.

### DIFF
--- a/Tests/test_image_resample.py
+++ b/Tests/test_image_resample.py
@@ -344,6 +344,7 @@ class CoreResamplePassesTest(PillowTestCase):
             cropped = im.crop(box).resize(im.size, Image.BILINEAR)
         self.assert_image_similar(with_box, cropped, 0.1)
 
+
 class CoreResampleCoefficientsTest(PillowTestCase):
     def test_reduce(self):
         test_color = 254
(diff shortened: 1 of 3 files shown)


Tests/test_image_resample.py
```diff
@@ -458,6 +459,86 @@ def test_formats(self):
                 cropped = im.crop(box).resize((32, 32), resample)
                 self.assert_image_similar(cropped, with_box, 0.4)
 
+    def test_passthrough(self):
+        "When no resize is required"
```
hugovk: Is there a reason for these to be unused strings and not docstrings or comments? 

Docstrings print out during test run, if I remember, so comments would be better for consistency.

homm: Thanks. [Fixed](https://github.com/python-pillow/Pillow/commit/a7cf91499a65978509a7e341c3215444a1c05777)

PR #2708, opened 2017-09-01
Removed debugging interface for Image.core.grabclipboard

radarhere:
This is `debugging; will be removed in future versions` since PIL.

It doesn't have a Python interface, so it seems like it can be removed.


wiredfool:
Do you want to kill the `#if 0` block below as well?

radarhere:
Okay, sure, done.

PR #2709, opened 2017-09-01
Alpha composite fixes

homm:
Pending changes in this PR:

- [x] compile-time const `PRECISION_BITS`
- [x] fix self as background  in `Image.alpha_composite`
- [x] accept lists as arguments in `Image.alpha_composite`


homm:
I missed that @radarhere already added test to alpha_composite's wrong arguments. In this case I done with it.

homm:
I accidentally pushed branch to `python-pillow:alpha-composite-box` (instead of `uploadcare:alpha-composite-box`), then removed it. Appveroy and Travis started unwanted builds and for some reasons this builds are attached to this PR

PR #2719, opened 2017-09-08
Fixes for Issues #2278 and #2006, value error in exif/tiff ifd

### LINKED ISSUE #2006: ValueError when getting tiff tags

### LINKED ISSUE #2278: AttributeError: 'tuple' object has no attribute 'ljust' when trying to save image
### What did you do?

Ran a snippet of code, see below.

The script to complete without error.

### What actually happened?

Stack trace, see below.

Pillow==3.4.2
(also fails with Pillow==3.5.0.dev0)
Python 3.5.2

Please include code that reproduces the issue and whenever possible, an image that demonstrates the issue. The best reproductions are self-contained scripts with minimal dependencies. If you are using a framework such as plone, django, or buildout, try to replicate the issue just using Pillow. 

```python
from io import BytesIO
from contextlib import closing
from PIL import Image

image = Image.open('phototest.tif')
print(image)

with closing(BytesIO()) as f:
    image.save(f, image.format or 'JPEG')
    value = f.getvalue()
    

```

Where the image can be found [here](https://github.com/tesseract-ocr/tesseract/blob/master/testing/phototest.tif?raw=true), or download it with this command:

```
curl  "https://github.com/tesseract-ocr/tesseract/blob/master/testing/phototest.tif?raw=true" > phototest.tif
```

The output is:

```
<PIL.TiffImagePlugin.TiffImageFile image mode=1 size=640x480 at 0x7F37D8546E10>
  File "try.py", line 9, in <module>
    image.save(f, image.format or 'JPEG')
  File "/usr/local/lib/python3.5/dist-packages/PIL/Image.py", line 1698, in save
  File "/usr/local/lib/python3.5/dist-packages/PIL/TiffImagePlugin.py", line 1487, in _save
    offset = ifd.save(fp)
  File "/usr/local/lib/python3.5/dist-packages/PIL/TiffImagePlugin.py", line 758, in save
    entries.append((tag, typ, count, data.ljust(4, b"\0"), b""))
AttributeError: 'tuple' object has no attribute 'ljust'
```

wiredfool:
Fixes #2278, fixes #2006, closes #2007 .

 * Tiff tags with unknown type/quantity now default to being bare values if they are 1 element, where previously they would be a single element tuple. This is only with the new api, not the legacy api.  This normalizes the handling of fields, so that the metadata with inferred or image specified counts are handled the same as metadata with count specified in the tiff spec. This is a change in behavior. 
 * The PhotoshopInfo, XMP, and JPEGTables tags now have a type and a count.
 * The ImageJMetaDataByteCounts now have an arbitrary number of items, as there can be multiple items, one for UTF-8, and one for UTF-16

 * [x] Need specific test for ImageJ meta tag
 * [x] Need specific test for PhotoshopInfo meta tag


wiredfool:
I'm happy with this now, need to rerun tests when aws/s3/us-east-1 comes back on line.

PR #2725, opened 2017-09-09
Drop support for EOL Python 3.3

hugovk:
Do we want to drop support for Python 3.3?

It's expected to reach end of life on September 29, 2017, after which there will be no more security releases. 

The last, 3.3.7, will be source-only (like the one before):

> Assuming no new security issues arise prior to the EOL date, 3.3.7 will likely be the final release of 3.3.  And you really shouldn't be using 3.3 at all at this point; while downstream distributors are, of course, free to provide support of 3.3 to their customers, in a little over two months when EOL is reached python-dev will no longer accept any issues or make any changes available for 3.3.  If you are still using 3.3, you really owe it to your applications, to your users, and to yourself to upgrade to a more recent release of Python 3, preferably 3.6!  Many, many fixes, new features, and substantial performance improvements await you.

* Benefits: reduces maintenance burden somewhat, some code can be removed, and some slots on the CI, I expect not many are still using 3.3, it might be good to nudge them up.
* Drawbacks: Perhaps some are still stuck on 3.3, Pillow should still work, it might not be too difficult to keep maintenance for a bit longer, fewer compatibility issues between 3.3 and 3.4+ than between 3.2 and 3.3+.

https://en.wikipedia.org/wiki/CPython#Version_history
https://www.python.org/dev/peps/pep-0398/#x-end-of-life
https://mail.python.org/pipermail/python-dev/2017-July/148584.html
http://blog.python.org/2017/09/python-337rc1-is-now-available-prior-to.html

If so, TODO:

* [ ] Similar changes to https://github.com/python-pillow/pillow-wheels


homm:
Maybe we need postpone this to 4.4.0 Jan 1 release?

wiredfool:
Unlike 2.6 and 3.2, I don't think that dropping 3.3 really gains us much. I can't think of any places in the code where we're specifically calling out 3.3 vs 3.4. 

On the other hand, Ubuntu Trusty is at 3.4, and that's the oldest supported LTS version out there. Centos OTOH, I'm not sure. They were the last running on 2.6.

hugovk:
> 20. What is the support ''end of life''
for each CentOS release?


...
> CentOS 5 CentOS-5 updates until March 31, 2017 

> CentOS 6 CentOS-6 updates until November 30,
2020

> CentOS 7 CentOS-7 updates until June 30, 2024



https://wiki.centos.org/FAQ/General#head-fe8a0be91ee3e7dea812e8694491e1dde5b75e6d



CentOS 5 ships with Python 2.4.

CentOS 6 ships with Python 2.6.

CentOS 7 ships with Python 2.7.

hugovk:
Let's drop this then until we have more beneficial benefits.

PR #2732, opened 2017-09-12
Cleanup Filter.c

homm:
I think this is fine to have in .h files the small inline functions which could be used in several modules. At least there are a lot in [SSE headers](https://github.com/gcc-mirror/gcc/blob/master/gcc/config/i386/emmintrin.h).

### DIFF
--- a/libImaging/ImagingUtils.h
+++ b/libImaging/ImagingUtils.h
@@ -30,8 +30,8 @@
     (MULDIV255(in1, (255 - mask), tmp1) + in2)
 
 
-/* This is work around bug in GCC prior 4.9 in 64 bit mode.
-   GCC generates code with partial dependency which 3 times slower.
+/* This is to work around a bug in GCC prior 4.9 in 64 bit mode.
+   GCC generates code with partial dependency which is 3 times slower.
    See: http://stackoverflow.com/a/26588074/253146 */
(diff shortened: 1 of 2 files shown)


libImaging/ImagingUtils.h
```diff
@@ -28,3 +28,18 @@
 
 #define PREBLEND(mask, in1, in2, tmp1)\
     (MULDIV255(in1, (255 - mask), tmp1) + in2)
+
+
+/* This is work around bug in GCC prior 4.9 in 64 bit mode.
+   GCC generates code with partial dependency which 3 times slower.
```
radarhere: 'which is 3 times slower'

libImaging/ImagingUtils.h
```diff
@@ -28,3 +28,18 @@
 
 #define PREBLEND(mask, in1, in2, tmp1)\
     (MULDIV255(in1, (255 - mask), tmp1) + in2)
+
+
+/* This is work around bug in GCC prior 4.9 in 64 bit mode.
```
radarhere: 'This is to work around a bug'

wiredfool:
+1, I'd rather have small inline functions and macros in header files than potentially duplicated all over the code.

Issue #2664, opened 2017-08-09
Release Pillow 4.3.0 on October 1, 2017


homm:

homm:
In this release I'm going to work on performance of most common operations. While [Pillow-SIMD is the fastest library](https://blog.uploadcare.com/the-fastest-production-ready-image-resize-out-there-part-0-7c974d520ad9) for image convolutional resize on CPU, there is still a room for optimizations without using SIMD.

- [ ] #2655 Fast image allocation affects almost every operation. Most of them does not require expensive computations and limited only by memory bandwidth, especially with multithreading
- [ ] #2661 New `Image.getchannel()` method returns single band from image, which is common operation, especially for alpha channel
- [ ] #1989 Using RGBX rawmode for JPEG image loading and saving eliminates unnecessary color conversion for most common image format
- [ ] Methods `image.split()` and `Image.merge()` use several calls of low-level functions `im.getband()` and `im.putband()` which is not effective enough. We need specific low-level functions for this operations which will be 2-2.5 times faster.
- [ ] All filters from `ImageFilter.py` module are single channel. They are applied to multichannel images by splitting original image to bands, applying filter to each band and merging bands again. Applying filters in one pass should be much more effective. This also unlocks SIMD optimizations.
- [ ] Many functions in `Pack.c` and `Unpack.c` operate byte-by-byte (for example `ImagingUnpackRGB`). This should be more effective to read 4 bytes at once, apply transformation and write back 4 bytes. The same should be much much more effective using 128-bit SIMD registers.

I also worried about [VIPS benchmarks](https://github.com/jcupitt/libvips/wiki/Speed-and-memory-use). Pillow doesn't show good result, because ≈70% of execution time is the filter applying (which is one of the most ineffective operation in Pillow). Funny thing is according to my tests Pillow is **already faster** than Python-VIPS in concurrent mode on 4-cores, even with such ineffective filters. My aim to next release is to make Pillow-SIMD on **single core** as fast as Python-VIPS on **all 6 cores** of Xeon E5-1650, used for tests :-> I also want to add VIPS benchmarks to [pillow-perf](https://github.com/python-pillow/pillow-perf) despite the fact that Python-VIPS is very hard to install.

wiredfool:
Sounds good to me.  I'm still mostly away from the keyboard for a few more days, but I'm hoping to have a chance to go through these prs within a week or so.

wiredfool:
* [ ] release notes for #2699

[...]

PR #2735, opened 2017-09-13
Deprecate ImageOps undocumented functions

homm:
Todo: add release notes for ImageFilter.BoxBlur

### DIFF
--- a/PIL/ImageFilter.py
+++ b/PIL/ImageFilter.py
@@ -160,6 +160,26 @@ class GaussianBlur(MultibandFilter):
         return image.gaussian_blur(self.radius)
 
 
+class BoxBlur(MultibandFilter):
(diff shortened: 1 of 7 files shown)


hugovk:
How long would this be deprecated before removing?

wiredfool:
It's hard to tell from code searches, but it doesn't appear that many people are using this. 

Most of the hits are from our tests, or a blanket "from PIL.ImageOps import *" in PIL_compat template code.

homm:
> How long would this be deprecated before removing?

This is not so huge change. I believe 2-3 versions or Pillow 5, if it comes earlier.

homm:
release notes added

wiredfool:
Still need to call out the deprecated functions in the release notes #2664

PR #2737, opened 2017-09-14
Rename ImagingCopyInfo to ImagingCopyPalette

homm:
Explicit is better than implicit. Potentially `ImagingCopyInfo` could copy something in addition to palette but this is not true for a long time.


homm:
Or maybe `ImagingCopyPalette` is better name?

wiredfool:
Yeah, info is a content free name.

Issue #2619, opened 2017-07-13
BUG: `np.array(arg: PIL.Image.Image)` results array of reversed image axis


mikigom:
I tried to convert `PIL.Image.Image` object to `numpy.ndarray` by `np.array` but it returns reversed axis.

```bash
$ python
Python 3.6.1 |Anaconda 4.4.0 (64-bit)| (default, May 11 2017, 13:09:58) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-1)] on linux
```

```python
>>> import PIL
>>> from PIL import Image
>>> import numpy as np
>>> np.__version__
'1.12.1'
>>> PIL.__version__
'4.1.1'
>>> image_Image = Image.new('RGBA', (7, 13))
>>> image_Image.size
(7, 13)
>>> image_np = np.array(image_Image)
>>> image_np.shape
(13, 7, 4) # (7, 13, 4) is expected
```

wiredfool:
Whatever it is, it hasn't changed recently. I suspect that the ordering is based on a row vs column major ordering, as there's the potential for memory mapping for certain modes.

radarhere:
http://scikit-image.org/docs/dev/user_guide/numpy_images.html#numpy-indexing
> Be careful: in NumPy indexing, the first dimension (camera.shape[0]) corresponds to rows, while the second (camera.shape[1]) corresponds to columns

vs

https://pillow.readthedocs.io/en/4.2.x/reference/Image.html#PIL.Image.new
> size – A 2-tuple, containing (width, height) in pixels.

An example of the way that numpy uses `shape` can be found at http://scikit-image.org/docs/dev/user_guide/numpy_images.html#color-images. In that example, there is a 'cat' image that is 451 pixels wide and 300 pixels high, and is described by `shape` as `(300, 451, 3)`

PR #2744, opened 2017-09-19
Bug about Image.fromarray

Bowenwu1:
Fixes # .Image.fromarray
When I use Pillow like that "img = Image.fromarray(matrix)" I find it will change my image size(exchange width and height). I can gain a proper result by use this function like that
"w, h = matrix.shape
    matrix = np.reshape(matrix, (h, w))
    img = Image.fromarray(matrix)
"

So, I think there must be a mistake such that exchange width and height.

 *  Exchange shape[1] and shape[0]
---
I am not a native English speaker and hope you can understand me.

### DIFF
--- a/PIL/Image.py
+++ b/PIL/Image.py
@@ -2429,7 +2429,7 @@ def fromarray(obj, mode=None):
     if ndim > ndmax:
         raise ValueError("Too many dimensions: %d > %d." % (ndim, ndmax))
 
-    size = shape[1], shape[0]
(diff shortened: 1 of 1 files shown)


wiredfool:
See this comment: https://github.com/python-pillow/Pillow/issues/2619#issuecomment-315519657

Numpy indexes arrays in the opposite order to python, row-major vs. column-major.

Bowenwu1:
Thank you!
It's quite weird!

PR #2749, opened 2017-09-20
Added executable flag to shell scripts

radarhere:


wiredfool:
I'm not sure that's a great idea.  The shebang line is iffy, they have to be run with the specific bash flags, and at least the build script is appveyor specific.

radarhere:
Okay, feel free to close if you like.

PR #2761, opened 2017-09-27
Add support for animated WebP files

### LINKED ISSUE #2755: KeyError: 'WEBP' when trying to write animated WebP file
### What did you do?
Attempt to save an image sequence to a *.webp file using save_all=True.

Write an animated WebP file. Looking at WebPImagePlugin.py, it's clear support hasn't been added for writing WebP image sequences yet, not sure if this is on the roadmap already or is terribly difficult to add, if it's pretty straightforward I might be able to contribute a PR for it.

### What actually happened?
Receive an error: "KeyError: 'WEBP'"

Pillow 4.2.1, Python 3.6.0

Code to reproduce:

```python
from PIL import Image
im = Image.open('image.png')
im.save('image.webp', save_all=True)
```

jd20:
Fixes #2755 .

 * Extend the C API to support WebPAnimEncoder / WebPAnimDecoder
 * Use the AnimEncoder / AnimDecoder API's, when mux support is compiled, except for when writing single-frame images (because internally the WebPAnimXXX code doesn't support RGB, only RGBA)
 * Also add support for XMP metadata chunks, and reading/writing metadata chunks to animated webp files
* Added support for specifying some additional encoder options, and updated docs accordingly. I've not exposed every possible option, as there are way too many, if someone else would like to tackle that feel free to (maybe encapsulate them in a WebPConfig object, so the C method arguments don't become ridiculously long).
* Add some tests to cover the new functionality

### Notes
* The WebPAnimEncoder / WebPAnimDecoder API is a bit more high-level than the Muxer API, meaning you can't specify frame fragments and offsets, but is a lot simpler overall (renders the final canvas for you, and takes care of disposal modes, blending, and everything). I suppose if someone really wanted ultimate control, they could re-implement the lower-level API, but it would be a lot more work (and seems you lose the benefit of libwebp doing things like keyframe insertion and compression heuristics for you).
* Internally, the WebPAnimXXX API's use RGBA for everything, even if the source and final output data is YUV. So, be aware of this, if performance is a huge concern. For most use cases, there shouldn't be much noticeable difference. I've put in some simple hacks, to still support the case of reading / writing images without alpha ("RGB" mode), but internally libwebp is still converting these to RGBA.
* The old behavior for unsupported modes (anything not "RGB" or "RGBA") was to fail with an IOError, I've changed it so that other modes are silently converted to the proper mode (with or without alpha). This should make working with WebP and Pillow more user-friendly, just something to be aware of.

## Testing
AppVeyor's reporting errors for me even before my changes, but I was able to successfully run the full tests, both with Python2 and Python3, with and without --disable-webpmux, using libwebp 0.6.0 and Python 2.7.13 / 3.6.0 on MacOS 10.12.6.


jd20:
Not sure the cause of the AppVeyor errors, but the Travis CI errors regarding for loop initial declarations I'll submit a fix for soon.

jd20:
WebP 0.5.0 added the WebPAnimEncoder and Decoder API, so will need an additional version check for that. Probably need to break up the version checks into: HAVE_WEBPMUX and HAVE_WEBPANIM, and bring metadata support back into WebPDecode_wrapper, for the case where you have a really old verison of libwebp, but still need the mux features.

wiredfool:
* Both currently supported Ubuntu LTS releases are still on webp 0.4.x, and those are mainstream platforms. Centos 6, the designated old library distro, is there as well. 
* Windows is c99 (I think) but the result is that all declarations of variables have to be at the beginning of the block. So `    WebPAnimEncoderOptions enc_options;` at _webp.c:69 needs to be up with the other declarations. I see a few others from a quick scan. 
* We've recently been being tighter about allowed modes for saving items (e.g., transparency isn't allowed in JPEG anymore), but in this case, it looks like it's always lossless conversion to RGB(A), rather than dropping a transparency channel. That should be acceptable.

_webp.c
```diff
@@ -8,8 +8,491 @@
...
+        int i_icc_size = (int)icc_size;
+        int i_exif_size = (int)exif_size;
+        int i_xmp_size = (int)xmp_size;
+        WebPData icc_profile = { icc_bytes, i_icc_size };
+        WebPData exif = { exif_bytes, i_exif_size };
+        WebPData xmp = { xmp_bytes, i_xmp_size };
+
+        mux = WebPMuxCreate(&webp_data, 1);
+        if (mux == NULL) {
+            fprintf(stderr, "ERROR: Could not re-mux to add metadata.\n");
```
wiredfool: These lines should set a PythonError and return NULL

_webp.c
```diff
@@ -8,8 +8,491 @@
...
+
+    WebPAnimDecoderObject* decp = (WebPAnimDecoderObject *)self;
+    if (!WebPAnimDecoderGetNext(decp->dec, &buf, &timestamp)) {
+        fprintf(stderr, "Error! Failed to read next frame.\n");
+        Py_RETURN_NONE;
+    }
+
+    // HACK: If original mode was RGB, we need to strip alpha before passing back, this
+    // is needed because internally WebPAnimDecoder doesn't suppor ta non-alpha mode
+    uint32_t size = decp->info.canvas_width * 4 * decp->info.canvas_height;
```
wiredfool: This should probably be a py_ssize_t, and there should be a check for overflow, or some sort of guarantee that this will never overflow due to limits on info.canvas_width.

wiredfool: Actually, looking at this, I'm not sure why we're not just returning this as RGBA or RGBX

jd20:
Thanks for the comments, I can work on those fixes today, good point about RGBX, I didn't realize Pillow supported it, that's why I thought I needed to convert to RGB first.

jd20:
Also, the original webp code used fprintf for libwebp errors, and I more or less just extended that pattern not knowing better, should I replace all those cases in _webp.c with real Python exceptions?

wiredfool:
>Also, the original webp code used fprintf for libwebp errors, and I more or less just extended that pattern not knowing better, should I replace all those cases in _webp.c with real Python exceptions?

Yes, go ahead and upgrade them. 

The other thing that I'd like to look at is making the webp decoding work in a similar manner to the other decoders -- namely doing the bulk of the work in the load phase rather than the open phase. That's one of the things that the long neglected webp decoder PR did. I'm not sure if the best course is to merge this first, or to try to do both at the same time.

jd20:
The PR I submitted is doing lazy loading, if using the WebPAnimDDecoder API. I copy all the data to memory and parse out canvas info and metadata chunks in open(), but the actual image data doesn't get decoded until load() is called.

wiredfool:
Ok, I wasn't quite precise there -- Reading in the entire file in open is not really in spec, especially for potentially large animations. It also winds up doing the image decode as a chunk, then writing to a string, which is then run through the raw decoder. 

This is a drawback of the original webp implementation, and not really your issue, but it's tightly related to what you're building. We've added some features to the decoder api that may make it easier to fix this, but I need to dig a bit to check.

jd20:
Right, it's not streaming from the file, decoding chunks as it goes along. The WebPAnimDecoder API doesn't support that either unfortunately (I can decode frame-by-frame, but not anything more granular than that). You really have to drop down to the Demux API to get the kind of control needed. So there'd be a lot more work required, to get it fully optimal.

jd20:
Pushed latest changes, which addresses all the issues in comments (except doing block decoding, that's too out of scope for me at this time). Travis build is passing now too, AppVeyor has been failing for some unrelated reason, I'm not familiar enough to know how to fix.

wiredfool:
Appveyor's failing on a variable definition after a command in the block. 

```
diff --git a/_webp.c b/_webp.c
index dcb2e29..fc3ac52 100644
--- a/_webp.c
+++ b/_webp.c
@@ -34,6 +34,8 @@ static const char* const kErrorMessages[-WEBP_MUX_NOT_ENOUGH_DATA + 1] = {
 };
 
 PyObject* HandleMuxError(WebPMuxError err, char* chunk) {
+    char message[100];
+    
     assert(err <= WEBP_MUX_NOT_FOUND && err >= WEBP_MUX_NOT_ENOUGH_DATA);
 
     // Check for a memory error first
@@ -42,7 +44,6 @@ PyObject* HandleMuxError(WebPMuxError err, char* chunk) {
     }
 
     // Create the error message
-    char message[100];
     if (chunk == NULL) {
         sprintf(message, "could not assemble chunks: %s", kErrorMessages[-err]);
     } else {
```

jd20:
Ok, now I see the appveyor errors for _webp.c, going through those...

jd20:
Ok, appveyor test is green now, any idea how to fix the codecov reports?

radarhere:
I wouldn't worry overly about 'No report found to compare against'. Coverage tools are not always perfect.

hugovk:
Please can you check the new tests are being run on CI? There's virtually no coverage on WebPImagePlugin.py. 

https://coveralls.io/builds/13503860/source?filename=PIL%2FWebPImagePlugin.py

I had a quick check on 3.6, and it looks like it's built okay (some warnings, nice to fix), and it's there in the selftest, but the unit tests are skipping:

TestFileWebpAlpha.test_read_rgba ... SKIP: WebP support not installed
TestFileWebpAlpha.test_write_lossless_rgb ... SKIP: WebP support not installed
TestFileWebpAlpha.test_write_rgba ... SKIP: WebP support not installed
TestFileWebpAlpha.test_write_unsupported_mode_PA ... SKIP: WebP support not installed

https://travis-ci.org/python-pillow/Pillow/jobs/281460879

hugovk:
They are passing for Windows on AppVeyor (only checked 2.7), but we don't yet collect coverage on there.

jd20:
@hugovk From what I can see in the build output, the summary from setup.py is that WEBP/WEBPMUX support was present, but then the summary from selftest.py is that the feature checks for webp failed, not sure why that would be the case.

hugovk:
Maybe: instead of importing from the internal _WEBP in the unit tests,  import the public WebPImagePlugin.py interface instead? Generally we should only test public stuff, and we want to test how end users should use the API.

jd20:
@hugovk The majority of the webp tests do use WebPImagePlugin.py, the only direct usage of the _webp module is to 1) check presence of webp features, for which there's no public equivalent, and 2) couple of the original webp tests which already existed, like test_version and test_WebPEncode_with_invalid_args. These tests could be dropped if we don't think they're useful, but I didn't create them originally so I left them alone when adding the new tests.

Tests/test_features.py
```diff
@@ -35,6 +35,11 @@ def check_webp_mux(self):
         self.assertEqual(features.check('webp_mux'),
                          _webp.HAVE_WEBPMUX)
 
+    @unittest.skipUnless(HAVE_WEBP, True)
+    def check_webp_anim(self):
```
hugovk: If we want this to run on the CI, which may be useful for debugging, call it test_ instead of check_. Could change the other two too.

hugovk:
Here's the problem (using this branch on macOS Sierra):

```
$ python
Python 2.7.13 (default, Dec 18 2016, 07:03:39)
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.42.1)] on darwin
>>> from PIL import _webp
ImportError: dlopen(PIL/_webp.so, 2): Symbol not found: _sprintf_s
  Referenced from: PIL/_webp.so
  Expected in: flat namespace
 in PIL/_webp.so
>>> 
```

After replacing `_sprintf_s` with `sprintf`:
```
$ python
Python 2.7.13 (default, Dec 18 2016, 07:03:39)
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.42.1)] on darwin
>>> from PIL import _webp
>>> 
```

I'll make a PR into your branch.

---

Without the fix, setup.py said "WEBP support available":
https://travis-ci.org/python-pillow/Pillow/jobs/281460879#L2635

But moments later selftest.py said "WEBP support not installed":
https://travis-ci.org/python-pillow/Pillow/jobs/281460879#L2654

That's because setup.py looks for library headers and things, but selftest.py uses Pillow's `features.check("webp")`, which is essentially doing the same `try: from PIL import _webp...` and fails. Let's use `features` for consistency.

hugovk:
> WebP 0.5.0 added the WebPAnimEncoder and Decoder API, so will need an additional version check for that. Probably need to break up the version checks into: HAVE_WEBPMUX and HAVE_WEBPANIM, and bring metadata support back into WebPDecode_wrapper, for the case where you have a really old verison of libwebp, but still need the mux features.

> Both currently supported Ubuntu LTS releases are still on webp 0.4.x, and those are mainstream platforms. Centos 6, the designated old library distro, is there as well.

* Ubuntu 16.04 LTS has EOL in April 2021
* Ubuntu 14.04 LTS has EOL in April 2019
* CentOS 6 has EOL in November 2020

https://linuxlifecycle.com/

Right now we're only installing and therefore testing code with WebP >= 0.5.0, which means we could very well break the untested, legacy code during the next four years. Is it worth installing an old WebP on the CI to also test the legacy code?

We can see the untested code here:

* 86% coverage https://codecov.io/gh/hugovk/Pillow/src/e75c38676017f8b686baaf3f261c31ce5f268a95/PIL/WebPImagePlugin.py

* 68% coverage https://codecov.io/gh/hugovk/Pillow/src/e75c38676017f8b686baaf3f261c31ce5f268a95/_webp.c

I think it'd be useful to put a note at the top of the code to mention when different bits can be removed/deprecated, as we're otherwise we're likely to forget. What do you think?

wiredfool:
@hugovk All of the dockers (except for alpine) are installing webp from the distro packages. 

Part of the point of the docker runs is to have a diversity of library versions so that we're not just running against the latest and greatest that no one is using because they're all on distro packages.

wiredfool:
> That's because setup.py looks for library headers and things, but selftest.py uses Pillow's features.check("webp"), which is essentially doing the same try: from PIL import _webp... and fails. Let's use features for consistency.

Those are two distinct checks. One is for what we should be compiling, one is what successfully compiled.

hugovk:
> @hugovk All of the dockers (except for alpine) are installing webp from the distro packages.
>
> Part of the point of the docker runs is to have a diversity of library versions so that we're not just running against the latest and greatest that no one is using because they're all on distro packages.

Aha, great! Checking the [CentOS 6 build](https://travis-ci.org/hugovk/Pillow/jobs/281694590#L1055), it has webpmux but not animation, as expected, and skips the TestFileWebpAnimation tests as expected. But the legacy mode `_open` is not covered:
https://coveralls.io/builds/13510797/source?filename=PIL%2FWebPImagePlugin.py#L35

And looks like we're only collecting coverage from the first 8 jobs: https://coveralls.io/builds/13510797

Checking logs, here's a problem with coveralls-merge on CentOS that doesn't occur for standard 2.7:
```
Successfully built coveralls-merge docopt pycparser
Installing collected packages: docopt, coveralls, coveralls-merge, asn1crypto, enum34, pycparser, cffi, ipaddress, cryptography, pyOpenSSL
Successfully installed asn1crypto-0.23.0 cffi-1.11.0 coveralls-1.2.0 coveralls-merge-0.0.2 cryptography-2.0.3 docopt-0.6.2 enum34-1.1.6 ipaddress-1.0.18 pyOpenSSL-17.3.0 pycparser-2.18
  File "/home/travis/virtualenv/python2.7.13/bin/coveralls-merge", line 11, in <module>
    sys.exit(main())
  File "/home/travis/virtualenv/python2.7.13/lib/python2.7/site-packages/coveralls_merge/core.py", line 84, in main
    m.collect()
  File "/home/travis/virtualenv/python2.7.13/lib/python2.7/site-packages/coveralls_merge/core.py", line 51, in collect
    self.merge(self.parse(f))
  File "/home/travis/virtualenv/python2.7.13/lib/python2.7/site-packages/coveralls_merge/core.py", line 36, in parse
    return json.loads(src.read())
  File "/opt/python/2.7.13/lib/python2.7/json/__init__.py", line 339, in loads
    return _default_decoder.decode(s)
  File "/opt/python/2.7.13/lib/python2.7/json/decoder.py", line 364, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/opt/python/2.7.13/lib/python2.7/json/decoder.py", line 382, in raw_decode
    raise ValueError("No JSON object could be decoded")
ValueError: No JSON object could be decoded
```

wiredfool:
I'm going to bet that the error there is because there's no data coming in from the coveralls run, not that there's something funky in coveralls-merge.

hugovk:
Yep, some divergence already with lcov:


Setting up lcov (1.10-1build1) ...
Setting up libgd-gd2-perl (1:2.46-3.1build1) ...
Capturing coverage data from .
Found gcov version: 4.8.4
Scanning . for .gcda files ...
geninfo: ERROR: no .gcda files found in .!

Versus:

Setting up lcov (1.10-1build1) ...
Setting up libgd-gd2-perl (1:2.46-3.1build1) ...
Capturing coverage data from .
Found gcov version: 4.8.4
Scanning . for .gcda files ...
Found 1 data files in .
Processing _webp.gcda
Finished .info-file creation

One thing: Docker and native both share after_success.sh script but not install.sh or script.sh. Are the Dockers built with lcov/CFLAGS="-coverage? Where is the Pillow build defined for Docker?

jd20:
@hugovk I've pushed the formatting changes you suggested, the only part I didn't understand was to use the features module in setup.py. I'm really not familiar with how setup.py works, is it as simple as replacing the _find_include_file and _find_library_file calls with features.check('webp')? I thought this code wasn't just checking for existence of webp support, but also determining which files to bundle up during setup process?

Tests/test_file_webp_lossless.py
```diff
@@ -4,37 +4,39 @@
...
-            from PIL import _webp
-        except:
+        if not HAVE_WEBP:
             self.skipTest('WebP support not installed')
+            return
 
         if (_webp.WebPDecoderVersion() < 0x0200):
             self.skipTest('lossless not included')
 
+        # WebPAnimDecoder only retuns RGBA or RGBX, never RGB
```
radarhere: returns

hugovk:
@jd20 Let's not worry about the features stuff too much, it was a bit of a red herring, the main problem there was the sprintf thing. It'd be nice if we can figure out why we're not collecting/building coverage properly for Docker builds, and then we know the legacy code remains tested properly.

wiredfool:
@hugovk That's a discussion that's probably better held in the docker-images repo.

Tests/test_file_webp_lossless.py
jd20: Thanks, fixed the typo (was also in test_file_webp.py too).

jd20:
Anything else I can do, to help push this PR along?

jd20:
Just checking in again, if there's anything I can do to move this PR along?

wiredfool:
I think you've pushed it as far as possible. I need to review where we were on this and make one last pass before merge.

PR #2676, opened 2017-08-12
Fast split()

homm:
This introduces `im.split()` core function and also speedups old `im.getband()` core function.

There are two sets of benchmarks: for current master and for conjunction with #2655
```
Slow allocation
               Master            Fast GetBand      im.split()
split LA      647.17 Mpx/s      804.72 Mpx/s      997.79 Mpx/s
split RGB     427.97 Mpx/s      538.45 Mpx/s      691.43 Mpx/s
split RGBA    331.82 Mpx/s      405.38 Mpx/s      519.33 Mpx/s

Fast allocation
               Fast allocation   Fast GetBand      im.split()
split LA      739.13 Mpx/s      961.76 Mpx/s     1364.44 Mpx/s
split RGB     492.85 Mpx/s      617.69 Mpx/s      981.31 Mpx/s
split RGBA    373.55 Mpx/s      486.30 Mpx/s      713.16 Mpx/s
```

PR #2764, opened 2017-09-28
Fix expression result unused warning

radarhere:
Fixes
```
libImaging/Bands.c:76:9: warning: expression result unused [-Wunused-value]
        (Imaging) ImagingError_ModeError();
        ^         ~~~~~~~~~~~~~~~~~~~~~~~~
```
by removing the cast.

Introduced in https://github.com/python-pillow/Pillow/pull/2676/files#diff-7dd4690995a96a7444dfaff941412c02R83

### DIFF
--- a/libImaging/Bands.c
+++ b/libImaging/Bands.c
@@ -73,7 +73,7 @@ ImagingSplit(Imaging imIn, Imaging bands[4])
 
     /* Check arguments */
     if (!imIn || imIn->type != IMAGING_TYPE_UINT8) {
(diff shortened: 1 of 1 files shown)


homm:
Formally, every function result should be used in C. So, strictly speaking `ImagingError_ModeError();` is also unused expression. The correct way to fix it is cast the result to void:

```C
(void) ImagingError_ModeError();
```

radarhere:
Okay, I've updated the commit.

Issue #2763, opened 2017-09-28
ReadTheDocs badge is red


hugovk:
### What did you do?

Check the ReadTheDocs badge on the README.

### What did you expect to happen?

It to be green.

### What actually happened?

It's red.

![image](https://user-images.githubusercontent.com/1324225/30955885-f152d2d2-a43d-11e7-99d2-9ec82adaaaf7.png)

hugovk:
Quite a few of the recent build failed at https://readthedocs.org/projects/pillow/builds/

![image](https://user-images.githubusercontent.com/1324225/30955996-4da08dcc-a43e-11e7-92da-b29fff44518a.png)

Here's the failing step of a failing build:

```
python /home/docs/checkouts/readthedocs.org/user_builds/pillow/envs/latest/bin/pip install --use-wheel -U --cache-dir /home/docs/checkouts/readthedocs.org/user_builds/pillow/.cache/pip Pygments==2.2.0 setuptools==28.8.0 docutils==0.13.1 mock==1.0.1 pillow==2.6.1 alabaster>=0.7,<0.8,!=0.7.5 commonmark==0.5.4 recommonmark==0.4.0 sphinx==1.5.3 sphinx-rtd-theme<0.3 readthedocs-sphinx-ext<0.6
Collecting Pygments==2.2.0
  Using cached Pygments-2.2.0-py2.py3-none-any.whl
Requirement already up-to-date: setuptools==28.8.0 in /home/docs/checkouts/readthedocs.org/user_builds/pillow/envs/latest/lib/python2.7/site-packages
Collecting docutils==0.13.1
  Using cached docutils-0.13.1-py2-none-any.whl
Requirement already up-to-date: mock==1.0.1 in /home/docs/checkouts/readthedocs.org/user_builds/pillow/envs/latest/lib/python2.7/site-packages
Collecting pillow==2.6.1
Collecting alabaster!=0.7.5,<0.8,>=0.7
  Using cached alabaster-0.7.10-py2.py3-none-any.whl
Requirement already up-to-date: commonmark==0.5.4 in /home/docs/checkouts/readthedocs.org/user_builds/pillow/envs/latest/lib/python2.7/site-packages
Requirement already up-to-date: recommonmark==0.4.0 in /home/docs/checkouts/readthedocs.org/user_builds/pillow/envs/latest/lib/python2.7/site-packages
Collecting sphinx==1.5.3
[...]

PR #1776, opened 2016-03-23
Clean up reqs

aclark4life:
- pip freeze
- Combine docs/requirements.txt with requirements.txt (No need for separate reqs if top-level calls sub-level.)

PR #2766, opened 2017-09-28
Unfreeze requirements.txt

hugovk:
requirements.txt was frozen in March 2016 in https://github.com/python-pillow/Pillow/pull/1776.

It causes some errors when making docs in a new VM (see https://github.com/python-pillow/Pillow/issues/2763#issuecomment-332805043), so how about just unfreezing them?


wiredfool:
Works for me -- We're not very sensitive to anything here, since it's 95% doc related, nose being the remainder, and that's test.

Issue #2772, opened 2017-10-01
Image.new segfaults on win-amd64


cgohlke:
Using current git master (https://github.com/python-pillow/Pillow/commit/007adc3134317ce3ed8737a887973399db70c779) on win-amd64-py3.6, `Image.new("1", (100, 100))` segfaults at [Fill.c#L53](https://github.com/python-pillow/Pillow/blob/master/libImaging/Fill.c#L53). This was working with yesterday's git master. All tests pass on win32-py3.6.
```
Python 3.6.2 (v3.6.2:5fd33b5, Jul  8 2017, 04:57:36) [MSC v.1900 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import faulthandler
>>> faulthandler.enable()
>>> from PIL import Image
>>> Image.new("1", (100, 100))
Windows fatal exception: access violation

Current thread 0x00000544 (most recent call first):
  File "D:\Build\Pillow\Pillow-git\PIL\Image.py", line 2287 in new
  File "<stdin>", line 1 in <module>
```

Call Stack:
```
 	vcruntime140.dll!00007fff1e54c840()	Unknown	Non-user code. Cannot find or open the PDB file.
>	_imaging.cp36-win_amd64.pyd!ImagingFill(ImagingMemoryInstance * im, const void * colour) Line 53	C	Symbols loaded.
 	_imaging.cp36-win_amd64.pyd!_fill(_object * self, _object * args) Line 617	C	Symbols loaded.
 	python36.dll!_PyCFunction_FastCallDict(_object * func_obj, _object * * args, __int64 nargs, _object * kwargs) Line 236	C	Symbols loaded.
 	[Inline Frame] python36.dll!_PyCFunction_FastCallKeywords(_object *) Line 294	C	Symbols loaded.
 	python36.dll!call_function(_object * * * pp_stack, __int64 oparg, _object * kwnames) Line 4809	C	Symbols loaded.
 	python36.dll!_PyEval_EvalFrameDefault(_frame * f, int throwflag) Line 3297	C	Symbols loaded.
 	[Inline Frame] python36.dll!PyEval_EvalFrameEx(_frame *) Line 718	C	Symbols loaded.
[...]

PR #2773, opened 2017-10-02
Fix bulletproof code from libc

cgohlke:
Fixes https://github.com/python-pillow/Pillow/issues/2772 (Image.new segfaults on win-amd64). 

Tested on Windows with Python 2.7 and 3.6, 32-bit and 64-bit.


hugovk:
libImaging\Storage.c(434) : error C2065: 'ssize_t' : undeclared identifier
libImaging\Storage.c(434) : error C2146: syntax error : missing ')' before identifier 'arena'
C:\

PR #2775, opened 2017-10-02
Return bool from isblock function

cgohlke:
Maybe I am missing something, but it looks like the `isblock` function should return a `bool`. 

This also fixes a compiler warning and a potential issue on 64-bit Windows:
```
_imaging.c(1759): warning C4311: 'type cast': pointer truncation from 'char *' to 'long'
```

### DIFF
--- a/_imaging.c
+++ b/_imaging.c
@@ -1756,7 +1756,14 @@ _box_blur(ImagingObject* self, PyObject* args)
 static PyObject*
 _isblock(ImagingObject* self, PyObject* args)
 {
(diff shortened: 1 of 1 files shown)


wiredfool:
This is used in one place in the code, ImageTk.py.  Boolean or integer is fine from the caller's point of view. The compiler's complaint is the cast truncation. We could probably do something along the lines of 

```
  return PyBool_fromLong(self->image->block != NULL);
```

We'd get the new reference without explicitly dealing with Py_True and Py_False.

PR #2784, opened 2017-10-04
Remove deprecated PIL.OleFileIO in favour of olefile Python package

hugovk:
The vendored version was removed in #2199, on 13 Dec 2016, in favour of the olefile Python package and replaced with a deprecation warning that PIL.OleFileIO would be removed in a future version.

We've had four quarterly releases since then (4.0.0, 4.1.0, 4.2.0, 4.3.0), shall we now finish this off?

cc: @decalage2 @jdufresne

### DIFF
--- a/PIL/OleFileIO.py
+++ b/PIL/OleFileIO.py
@@ -1,4 +1,0 @@
-raise ImportError(
-    'PIL.OleFileIO is deprecated. Use the olefile Python package '
-    'instead. This module will be removed in a future version.'
-)


wiredfool:
Is this going to improve anyone's experience in using Pillow?

hugovk:
If they're still importing PIL/OleFileIO.py, then rather than the deprecation warning they'll get an import error and will finally, after a year of warnings, have to update their code to import from the real module.

wiredfool:
Well, yes, that's what the PR does. But does it improve their experience? It takes formerly working code and intentionally breaks it. And unlike some of our other deprecations, it removes all evidence that this was here without pointing to the problem that they are now having.

jdufresne:
> And unlike some of our other deprecations, it removes all evidence that this was here without pointing to the problem that they are now having.

Would a mention in `CHANGES.rst` handle this concern?

hugovk:
We can keep this file longer (forever?), or replace it with an import error that says the same thing and then remove it later.

jdufresne:
> or replace it with an import error that says the same thing and then remove it later.

I think this is a good compromise. It would still break code requiring someone to take action but would do so with a more helpful message.

hugovk:
https://github.com/python-pillow/Pillow/pull/2833 has been merged, let's revisit this in the future.

PR #2785, opened 2017-10-04
ImageDraw.textsize: fix zero length error

### LINKED ISSUE #2783: ImageDraw.textsize zero length error
### What did you do?
Attempted to get the `textsize` of an empty string ("")
Returns a tuple which is the size of the text
### What actually happened?
Produced this error
```py
  File "/usr/local/lib/python3.6/site-packages/PIL/ImageDraw.py", line 263, in textsize
    return font.getsize(text, direction, features)
  File "/usr/local/lib/python3.6/site-packages/PIL/ImageFont.py", line 111, in getsize
    return self.font.getsize(text)
SystemError: <built-in method getsize of ImagingFont object at 0x2c72760> returned NULL without setting an error
```
Pillow 4.30
Python 3.6

I believe this can be traced to the latest 4.3 release and this commit: https://github.com/python-pillow/Pillow/commit/fd0e7be55a77dd1d499c0b9df99d24e23e8c15e3

The following code is producing the error: https://github.com/python-pillow/Pillow/blob/46845bafc544bc117de2743c3d0a509b312f0442/_imaging.c#L2304-L2322

In previous versions, the function would not error. If it's not intended to get the size of an empty string, maybe changing the behavior of [`ImageDraw._multiline_split`](https://github.com/python-pillow/Pillow/blob/46845bafc544bc117de2743c3d0a509b312f0442/PIL/ImageDraw.py#L203-L206) would fix the issue.

**Code**
```py
import PIL.Image, PIL.ImageDraw
i = PIL.Image.new('RGB', (2000, 1000))
d = PIL.ImageDraw.Draw(i)
d.textsize("")
```

hugovk:
Fixes #2783.

 * If the string is empty, its size is (0, 0).

### DIFF
--- a/PIL/ImageDraw.py
+++ b/PIL/ImageDraw.py
@@ -217,7 +217,8 @@ class ImageDraw(object):
             ink = fill
         if ink is not None:
             try:
-                mask, offset = font.getmask2(text, self.fontmode, *args, **kwargs)
+                mask, offset = font.getmask2(text, self.fontmode,
+                                             *args, **kwargs)
                 xy = xy[0] + offset[0], xy[1] + offset[1]
             except AttributeError:
                 try:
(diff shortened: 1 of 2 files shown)


wiredfool:
Lets fix the underlying issue. See #2788

hugovk:
Fixed by https://github.com/python-pillow/Pillow/pull/2788 instead.

PR #2787, opened 2017-10-04
4.4.0.dev0 version bump

radarhere:


hugovk:
Does this need documenting in https://github.com/python-pillow/Pillow/blob/master/RELEASING.md?

PR #2789, opened 2017-10-04
Allow the olefile dependency to be optional

jdufresne:
Support for plugins requiring olefile will not be loaded if it is not installed. Allows library consumers to avoid installing this dependency if they choose. Some library consumers have little interest in the format support and would like to keep dependencies to a minimum.

 * olefile support is no longer a required dependency
 * if it is exists, image support requiring olefile continues to be available


wiredfool:
This needs release notes and potentially additional documentation. 

Release note ping: #2832

PR #2805, opened 2017-10-17
Fix #2804: sets encoderinfo for images when saving multi-page PDF

### LINKED ISSUE #2804: Error while saving multiple JPEGs as a multi-page PDF
### What did you do?

I wanted to make a PDF out of multiple JPEG images, I loaded the JPEGs and try to save the first one as a PDF with the other images appended.

I expected a multi-page PDF, each page being one of my input images.

### What actually happened?

I got the following error:

```
  File "[...]/lib/python3.5/site-packages/PIL/Image.py", line 1933, in save
  File "[...]/lib/python3.5/site-packages/PIL/PdfImagePlugin.py", line 60, in _save_all
    _save(im, fp, filename, save_all=True)
  File "[...]/lib/python3.5/site-packages/PIL/PdfImagePlugin.py", line 187, in _save
    Image.SAVE["JPEG"](im, op, filename)
  File "[...]/lib/python3.5/site-packages/PIL/JpegImagePlugin.py", line 609, in _save
    info = im.encoderinfo
AttributeError: 'JpegImageFile' object has no attribute 'encoderinfo'
```

Pillow 4.3.0 on Python 3.5

### How to reproduce this and more

I had the problem with JPEGs of mine but even with two very simple JPEGs generated with imagemagick I get the errors. What's interesting is if I try to make a TIFF first I can then make a PDF.

```python
# The bug should be on any two random JPEGs, I generated the following to test:
# > convert -size 100x100 xc:#990000 a.jpeg
# > convert -size 100x100 xc:#009900 b.jpeg

from PIL import Image

Image.open('a.jpeg').save('test.pdf', save_all=True, append_images=[Image.open('b.jpeg')])
# AttributeError: 'JpegImageFile' object has no attribute 'encoderinfo'

a = Image.open('a.jpeg')
b = Image.open('b.jpeg')
a.save('test.pdf', save_all=True, append_images=[b]) # ERROR 
a.save('test.gif', save_all=True, append_images=[b]) # SUCCESS
a.save('test.pdf', save_all=True, append_images=[b]) # ERROR
a.save('test.tiff', save_all=True, append_images=[b]) # SUCCESS
a.save('test.pdf', save_all=True, append_images=[b]) # SUCCESS
```

Test JPEGs you can use:

b
a

ixio:
Fixes #2804 

Seeing that saving as a TIFF before saving as a PDF fixed the problem (see issue #2804 for more details), I tried to copy some of the _save_all method in TiffImagePlugin.py into the _save_all method of PdfImagePlugin.py and it seems to work.

### DIFF
--- a/PIL/PdfImagePlugin.py
+++ b/PIL/PdfImagePlugin.py
@@ -139,7 +139,7 @@ def _save(im, fp, filename, save_all=False):
         for append_im in append_images:
             if append_im.mode != im.mode:
                 append_im = append_im.convert(im.mode)
-                append_im.encoderinfo = im.encoderinfo.copy()
+            append_im.encoderinfo = im.encoderinfo.copy()
             ims.append(append_im)
     numberOfPages = 0
     for im in ims:
--- a/Tests/test_file_pdf.py
+++ b/Tests/test_file_pdf.py
@@ -80,6 +80,13 @@ class TestFilePdf(PillowTestCase):
         self.assertTrue(os.path.isfile(outfile))
         self.assertGreater(os.path.getsize(outfile), 0)
 
+        # Append JPEG images
+        jpeg = Image.open("Tests/images/flower.jpg")
+        jpeg.save(outfile, save_all=True, append_images=[jpeg.copy()])
+
+        self.assertTrue(os.path.isfile(outfile))
+        self.assertGreater(os.path.getsize(outfile), 0)
+
 
 if __name__ == '__main__':
     unittest.main()


hugovk:
A couple of general points: please use more descriptive commit messages (and PR titles), and include unit tests with the fix. Thank you!

ixio:
Thanks for the info. I wasn't sure what my commit message should look like (it's not detailed in the contribution guide) but I hope the new one does the trick.

I've also added a test in the commit (or should have I made a second commit?). The test is very specific to my use case #2804 that was failing but it works well (fails on master and succeeds on my PR).

wiredfool:
Before we merge this, I'd like to be able to answer if this addresses the root cause of the issue, or if it's just a bandaid over something.

ixio:
Well I can't be sure because I'm not really familiar with your code base or image processing in general. 

However the error message is "AttributeError: 'JpegImageFile' object has no attribute 'encoderinfo'" because it tries to get the encoderinfo of one of the appended images and it fails as it doesn't have this attribute (in my test case with jpegs).

Since I noticed that saving the JPEGs as a TIFF before saving as PDF solved the problem, I copied the way the _save_all method in TiffImagePlugin.py adds encoderinfo to the appended images.

While I'm not certain, I am fairly confident that this is the correct fix for this particular problem as it looks pretty straightforward.

ixio:
Oops sorry I spoke too soon, this does look like a bandaid as there is that code already in place:

```python
    if save_all:
        append_images = im.encoderinfo.get("append_images", [])
        for append_im in append_images:
            if append_im.mode != im.mode:
                append_im = append_im.convert(im.mode)
                append_im.encoderinfo = im.encoderinfo.copy()
ims.append(append_im)
```

I'll look into it and get back to you.

ixio:
Okay I fixed what might just have been an extra typo tab. My initial fix would've been a bandaid, just like wiredfool wondered.

Or is there a reason otherwise for not saving the encoderinfo in the case the modes of the original image and the appended image are different?

PR #2820, opened 2017-10-31
fix for #272 on SmartOS. Stripping whitespace from string.whitespace.

### LINKED ISSUE #272: Image.open raise UnicodeDecodeError if locale setting modified
``` python
import locale
from PIL import Image

locale.setlocale(locale.LC_ALL, "polish")
im = Image.open('test.jpg')

  File "E:\DOWNLOADS\jpegtran\test.py", line 6, in <module>
    im = Image.open('test.jpg')
  File "G:\Python27\lib\site-packages\PIL\Image.py", line 1982, in open
    preinit()
  File "G:\Python27\lib\site-packages\PIL\Image.py", line 300, in preinit
    from PIL import PpmImagePlugin
  File "G:\Python27\lib\site-packages\PIL\PpmImagePlugin.py", line 27, in <module>
    b_whitespace = string.whitespace.encode()
UnicodeDecodeError: 'ascii' codec can't decode byte 0xa0 in position 6: ordinal
not in range(128)
```

markmiscavage:
Fixes #272 
Strips the whitespace that comes through on the `string.whitespace` call on SmartOS systems. 

 * 
 * 
 *

### DIFF
--- a/PIL/PpmImagePlugin.py
+++ b/PIL/PpmImagePlugin.py
@@ -24,7 +24,7 @@ __version__ = "0.2"
 #
 # --------------------------------------------------------------------
 
-b_whitespace = string.whitespace
+b_whitespace = string.whitespace.strip()
 try:
     import locale
     locale_lang, locale_enc = locale.getlocale()


hugovk:
At least this breaks all the existing CI builds.

---

Examples on Mac:
```python
Python 2.7.14 (default, Sep 25 2017, 09:53:22)
[GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.37)] on darwin
>>> import string
>>> string.whitespace
'\t\n\x0b\x0c\r '
>>> string.whitespace.strip()
''
>>>
```
```python
Python 3.6.3 (default, Oct  4 2017, 06:09:15)
[GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.37)] on darwin
>>> import string
>>> string.whitespace
' \t\n\r\x0b\x0c'
>>> string.whitespace.strip()
''
>>>
```

What do you get for `string.whitespace`?

---

https://github.com/python-pillow/Pillow/issues/272#issuecomment-339436387:

> I appear to still be getting this error. I'm only seeing this currently on SmartOS base-64 17.3.0. 
> 
> PIL/PpmImagePlugin.py line 33 is failing with `UnicodeDecodeError: 'utf8' codec can't decode byte 0xa0 in position 6: invalid start byte`,  but the try/except is just skipping that error to fail with the same error seen above. 
> 
> b_whitespace = string.whitespace.strip() seems to fix the problem for me. 

```python
try:
    import locale
    locale_lang, locale_enc = locale.getlocale()
    if locale_enc is None:
        locale_lang, locale_enc = locale.getdefaultlocale()
    b_whitespace = b_whitespace.decode(locale_enc)  # Line 33
except:
    pass
```

That bare `except` isn't helping much. Anyway, what do you get for `locale.getdefaultlocale()`? Example on Mac: `('en_GB', 'UTF-8')`.

wiredfool:
The spec (http://netpbm.sourceforge.net/doc/ppm.html) defines whitespace as: 
```
A "white space" character is space, CR, LF, TAB, VT, or FF (I.e. what the ANSI standard C isspace() function calls white space).
```
Which are essentially the old pre-locale default whitespace characters. So, we should probably just assign a fixed list, rather than pulling the locale dependent whitespace characters from the string module.

e.g.:
```
b_whitespace = (cnr(32), chr(10), chr(13), chr(9), chr(11), chr(12))
```

markmiscavage:
`Python 2.7.14 (default, Oct  3 2017, 16:34:09) 
[GCC 4.9.4] on sunos5
(InteractiveConsole)
>>> import string
>>> string.whitespace
'\t\n\x0b\x0c\r \xa0'
>>> import locale
>>> locale.getdefaultlocale()
('en_US', 'UTF-8')`

Notice the space after the \r in string.whitespace, thats what was causing the problem for me. Switching it to a fixed list sounds sane to me.

hugovk:
I had a quick test with https://github.com/hugovk/Pillow/commit/5867ce0bc3e179636aef0ee863e7e44163553c3d but it's failing selftest.py on Python 3: https://travis-ci.org/hugovk/Pillow/builds/295933112

selftest and the full tests pass for me locally.

wiredfool:
Looks like it needs to be done as a bytes object: https://github.com/wiredfool/Pillow/blob/pr_2820/PIL/PpmImagePlugin.py#L27

hugovk:
https://github.com/wiredfool/Pillow/blob/pr_2820/PIL/PpmImagePlugin.py#L27 as a PR: https://github.com/python-pillow/Pillow/pull/2831.

Issue #2832, opened 2017-11-05
Release 5.0.0 Issue: 2018-01-01


wiredfool:
Release issue placeholder for the ~4.4.0~ 5.0.0 Release.

wiredfool:
if we're going to merge the PyTest PR, this is going to turn into a major (5.0) release, due to pytest's lack of support for EOL python 3.3.  There are other potential breaking/major changes that might be worth merging.

* #2257, the pillow scripts rename/install issue. 
* #2583, Decompression bomb warnings -> error
* #2339, Reworking the exception types that we throw.

homm:
As part of the major release, I also would like to see #2198, because MRO could change call order.

wiredfool:
I dislike #2198, especially WRT python2 required syntax, but also in terms of the MRO at runtime vs our build/test time.

wiredfool:
I don't think I'm going to get to #2339 -- There's a lot of subsequently changed imports that are merge errors just to get to the point where it can merge, and then there are the reviews for if we're doing it correctly.

cgohlke:
How about discontinuing `bdist_egg` and `bdist_wininst` binaries for Windows?

wiredfool:
I'm OK with that, it'll need to be added to the release notes. Basically that means that they need pip of a reasonable recent version?

wiredfool:
I've pushed the version bump, but have not tagged yet. I'm waiting on travis to get to the wheel builds, but it's being really slow about that.

wiredfool:
Looks like we're getting a max of 2 concurrent builds right now, for comparison, appveyor finished before travis, which basically never happens.

hugovk:
Travis is now green. 

https://travis-ci.org/python-pillow/pillow-wheels/builds/323791158

wiredfool:
Tagged 5.0.0

cgohlke:
[Windows binaries](https://www.lfd.uci.edu/~gohlke/download/Pillow-5.0.0-win-dist.zip) are ready.

wiredfool:
Thanks @cgohlke

wiredfool:
And we're uploaded and released

radarhere:
Maybe this is just caching, but it looks to me like the 'Make sure the default version for Read the Docs is the latest release version' step in the RELEASING checklist hasn't been done?

[...]

PR #2833, opened 2017-11-05
Replace PIL.OleFileIO warning with descriptive ImportError

hugovk:
The vendored version was removed in #2199, on 13 Dec 2016, in favour of the olefile Python package and replaced with a deprecation warning that PIL.OleFileIO would be removed in a future version.

We've had four quarterly releases since then (4.0.0, 4.1.0, 4.2.0, 4.3.0), shall we now ~finish this off~ replace it with an import error that says the same thing and then remove it at a later date?

See https://github.com/python-pillow/Pillow/pull/2784 for removal at a future date.

cc: @decalage2 @jdufresne

### DIFF
--- a/PIL/OleFileIO.py
+++ b/PIL/OleFileIO.py
@@ -1,12 +1,4 @@
-import warnings
-
-warnings.warn(
+raise ImportError(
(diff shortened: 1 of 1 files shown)


wiredfool:
Release notes needed #2832

PR #2838, opened 2017-11-06
How about yum install? [ci skip]

aclark4life:

 * Add ``sudo yum install libjpeg-devel libwebp-devel freetype-devel openjpeg2-devel libtiff-devel libimagequant-devel lcms2-devel`` to docs/installation.rst

Seems like the whole ``installation.rst`` could use a rewrite, but in the meantime I came across @python-pillow/pillow-team not having specific Redhat/CentOS dependency installation instructions, so I'm adding them here.

Also even though I have ``libimagequant-devel`` installed, I don't get expected support (as far as I can tell)::

```

PIL SETUP SUMMARY
version      Pillow 4.3.0
platform     linux2 2.7.5 (default, Aug  4 2017, 00:39:18)
             [GCC 4.8.5 20150623 (Red Hat 4.8.5-16)]
--- JPEG support available
--- OPENJPEG (JPEG2000) support available (2.1)
--- ZLIB (PNG/ZIP) support available
*** LIBIMAGEQUANT support not available
--- LIBTIFF support available
--- FREETYPE2 support available
--- RAQM support available
--- LITTLECMS2 support available
--- WEBP support available
*** WEBPMUX support not available
```

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -316,7 +316,7 @@ Or for Python 3::
 
 .. Note:: ``redhat-rpm-config`` is required on Fedora 23, but not earlier versions.
 
-Prerequisites are installed on **Ubuntu 14.04 LTS** with::
+Prerequisites can be installed on **Ubuntu 14.04 LTS** with::
 
     $ sudo apt-get install libtiff5-dev libjpeg8-dev zlib1g-dev \
         libfreetype6-dev liblcms2-dev libwebp-dev libharfbuzz-dev libfribidi-dev \
@@ -324,7 +324,12 @@ Prerequisites are installed on **Ubuntu 14.04 LTS** with::
 
 Then see ``depends/install_raqm.sh`` to install libraqm.
 
-Prerequisites are installed on **Fedora 23** with::
+Prerequisites can be installed on **RedHat/CentOS 7.4** with::
+
+    $ sudo yum install libjpeg-devel libwebp-devel freetype-devel openjpeg2-devel \
+        libtiff-devel libimagequant lcms2-devel
+
+Prerequisites can be installed on **Fedora 23** with::
 
     $ sudo dnf install libtiff-devel libjpeg-devel zlib-devel freetype-devel \
         lcms2-devel libwebp-devel tcl-devel tk-devel libraqm-devel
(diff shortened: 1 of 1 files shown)


wiredfool:
We do have 4 redhat variants in the dockerfile repo (amazon, centos 6, fedora 24 and 26) with known working sets of packages and package manager invocations. While we don't have RH or Centos 7, those would possibly be good to add there. 

A

docs/installation.rst
```diff
@@ -316,15 +316,20 @@ Or for Python 3::
...
         libfreetype6-dev liblcms2-dev libwebp-dev libharfbuzz-dev libfribidi-dev \
         tcl8.6-dev tk8.6-dev python-tk
 
 Then see ``depends/install_raqm.sh`` to install libraqm.
 
-Prerequisites are installed on **Fedora 23** with::
+Prerequisites can be installed on **RedHat/CentOS 7.4** with::
+
+    $ sudo yum install libjpeg-devel libwebp-devel freetype-devel openjpeg2-devel \
+        libtiff-devel libimagequant lcms2-devel
```
wiredfool: This may be why your libimagequant isn't working, since this isn't the dev package.

aclark4life: @wiredfool That's a typo, I definitely have the ``devel`` package installed.

aclark4life:
Also for completeness'-sake, I'll mention I didn't see any ``webpmux`` libs avail

hugovk:
Here comes CentOS 7: https://github.com/python-pillow/docker-images/pull/16.

At some point we could switch out EOL Fedora 24 for something newer. Fedora 25 and 26 are still supported, 27 is due out next week: 
image
https://en.wikipedia.org/wiki/Fedora_(operating_system)#Releases

wiredfool:
Libimagequant-devel puts it's headers in a subdirectory on centos 7: 
`CFLAGS=-I/usr/include/imagequant python setup.py build_ext --debug`

hugovk:
If we want to support webp 0.3.0 then it's good to test it on at least one build.

wiredfool:
> Here comes CentOS 7: python-pillow/docker-images#16.

Fedora 25 as well, python-pillow/docker-images#17.  Fedora 27 doesn't appear to be available yet in docker form.

wiredfool:
Need to test on current centos/redhat to check to see if we're getting: 
 [ ] all the dependencies
 [ ] if any are actually different than fedora. 

Docwise, Fedora should probably be updated to 27, rather than 23, as that's out of support.

If the packages aren't actually any different, I'm not sure it's worth calling out the dnf/yum difference, as people who are compiling pillow should be able to determine their preferred package manager for their platform.

wiredfool:
Covered in 2919

PR #2852, opened 2017-11-13
Fillcolor parameter for Transform

### LINKED ISSUE #2837: Affine transformation fill parameter seems to do nothing
I would like to use PIL to perform an affine warp on an image, but I want to specify a custom fill value for undefined pixels. The `transform` function seems to have a `fill=1` argument, but it is undocumented and looking further in the code it seems to be unused. 

Is there any way to get a custom fill value in an affine transform?

wiredfool:
Fixes #2837 .

 * Adds fillcolor parameter to transform to fill the target image prior to the transform
 * Pipes through the existing fill parameter in `_imaging.transform2` rather than hardwiring it to true.
 
Note: This _extends_ the api and retains the currently mostly ignored `fill` parameter. This `fill` parameter is currently passed into the ImageTransformHandler api, which has user supplied code that we don't control. If we repurpose that parameter, we risk breaking existing code.

### DIFF
--- a/Tests/test_image_transform.py
+++ b/Tests/test_image_transform.py
@@ -267,16 +267,5 @@ class TestImageTransformPerspective(TestImageTransformAffine):
     # Repeat all tests for AFFINE transformations with PERSPECTIVE
     transform = Image.PERSPECTIVE
 
-class TestFill(unittest.TestCase):
-    def test_fill(self):
-        im = hopper('RGB')
-        (w, h) = im.size
-        transformed = im.transform(im.size, Image.EXTENT,
-                                   (0, 0,
-                                    w*2, h*2),
-                                   Image.BILINEAR,
-                                   fill = 'red')
-        transformed.show()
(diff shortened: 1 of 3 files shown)


Tests/test_image_transform.py
```diff
@@ -256,6 +267,16 @@ class TestImageTransformPerspective(TestImageTransformAffine):
...
+class TestFill(unittest.TestCase):
+    def test_fill(self):
+        im = hopper('RGB')
+        (w, h) = im.size
+        transformed = im.transform(im.size, Image.EXTENT,
+                                   (0, 0,
+                                    w*2, h*2),
+                                   Image.BILINEAR,
+                                   fill = 'red')
+        transformed.show()
```
radarhere: You've left debugging code in?

wiredfool: oops

wiredfool: and two copies of the same test.

PR #2853, opened 2017-11-13
Fix setup.py on debian

### LINKED ISSUE #2851: Can't build Pillow on Debian armhf
### What did you do?

On a Debian armhf machine, I tried running: (keeping only the relevant parts)

```
$ python setup.py install
...
Looking for zlib
Checking for include file zlib.h in /tmp/pip-rje8ft8s-build/libImaging
Checking for include file zlib.h in /usr/include
Found zlib.h
Couldn't find library z in ['/usr/lib', '/usr/lib']
...
The headers or library files could not be found for zlib,
a required dependency when compiling Pillow from source.

Please see the install instructions at:
   https://pillow.readthedocs.io/en/latest/installation.html

...
```

The problem is that [the code which detects the library dirs to search in](https://github.com/python-pillow/Pillow/blob/master/setup.py#L313-L341) just doesn't work at all on Debian.

It suffers from many problems, but let's start with the ARM-related protion:

```python
                    elif platform_ in ["arm", "armv7l"]:
                        _add_directory(library_dirs,
                                       "/usr/lib/arm-linux-gnueabi")
                        break
```

This is missing the `/usr/lib/arm-linux-gnueabihf` directory for the armhf architecture (ARM 32 bits with hardware floating point)

More fundamentally, `platform_` in this case comes from `arch_tp = (plat.processor(), plat.architecture()[0])`. On a Debian ARM system, that's `('', '32bit')`, which means we never even pass in the above condition anyway.

But in fact, this whole architecture detection code suffers from the same problem, because even on a less exotic system, for example x86_64, then `arch_tp = ('', '64bit')`. Basically Python's `platform.processor()` always seems to return an empty string on Debian.

This means that the conditional below can never evaluate to `True`, even in the case it was designed for:

```python
            if arch_tp == ("x86_64", "32bit"):
                # 32-bit build on 64-bit machine.
                _add_directory(library_dirs, "/usr/lib/i386-linux-gnu")
```

The only reason Pillow builds at all on Debian x86_64 and i386 is because those two conditions are two broad:

```python
                    if platform_ in ["x86_64", "64bit"]:
                        _add_directory(library_dirs, "/lib64")
                        _add_directory(library_dirs, "/usr/lib64")
                        _add_directory(library_dirs,
                                       "/usr/lib/x86_64-linux-gnu")
                        break
                    elif platform_ in ["i386", "i686", "32bit"]:
                        _add_directory(library_dirs, "/usr/lib/i386-linux-gnu")
                        break
```

The `64bit` and `32bit` also match AArch64 and ARM machines, where the above multiarch libdirs really shouldn't be added as they are the wrong ones.

---

For reference, the Debian package of Pillow uses this patch (among others) : 

```patch
diff --git a/setup.py b/setup.py
index 4948faf..02676ba 100755
--- a/setup.py
+++ b/setup.py
@@ -311,61 +311,6 @@ class pil_build_ext(build_ext):
                 _add_directory(include_dirs, "/usr/X11/include")

         elif sys.platform.startswith("linux"):
-            arch_tp = (plat.processor(), plat.architecture()[0])
-            if arch_tp == ("x86_64", "32bit"):
-                # 32-bit build on 64-bit machine.
-                _add_directory(library_dirs, "/usr/lib/i386-linux-gnu")
-            else:
-                for platform_ in arch_tp:
-
-                    if not platform_:
-                        continue
-
-                    if platform_ in ["x86_64", "64bit"]:
-                        _add_directory(library_dirs, "/lib64")
-                        _add_directory(library_dirs, "/usr/lib64")
-                        _add_directory(library_dirs,
-                                       "/usr/lib/x86_64-linux-gnu")
-                    elif platform_ in ["i386", "i686", "32bit"]:
-                        _add_directory(library_dirs, "/usr/lib/i386-linux-gnu")
-                    elif platform_ in ["aarch64"]:
-                        _add_directory(library_dirs, "/usr/lib64")
-                        _add_directory(library_dirs,
-                                       "/usr/lib/aarch64-linux-gnu")
-                    elif platform_ in ["arm", "armv7l"]:
-                        _add_directory(library_dirs,
-                                       "/usr/lib/arm-linux-gnueabi")
-                    elif platform_ in ["ppc64"]:
-                        _add_directory(library_dirs, "/usr/lib64")
-                        _add_directory(library_dirs,
-                                       "/usr/lib/ppc64-linux-gnu")
-                        _add_directory(library_dirs,
-                                       "/usr/lib/powerpc64-linux-gnu")
-                    elif platform_ in ["ppc"]:
-                        _add_directory(library_dirs, "/usr/lib/ppc-linux-gnu")
-                        _add_directory(library_dirs,
-                                       "/usr/lib/powerpc-linux-gnu")
-                    elif platform_ in ["s390x"]:
-                        _add_directory(library_dirs, "/usr/lib64")
-                        _add_directory(library_dirs,
-                                       "/usr/lib/s390x-linux-gnu")
-                    elif platform_ in ["s390"]:
-                        _add_directory(library_dirs, "/usr/lib/s390-linux-gnu")
-                else:
-                    raise ValueError(
-                        "Unable to identify Linux platform: `%s`" % platform_)
-
-                # XXX Kludge. Above /\ we brute force support multiarch. Here we
-                # try Barry's more general approach. Afterward, something should
-                # work ;-)
                 self.add_multiarch_paths()

                 # termux support for android.
```

That's right, they just completely remove the attempt at automatically detecting things, and directly use the `dpkg-architecture` code path (in `self.add_multiarch_paths()`), which otherwise gets skipped by the bad conditionals.

It seems like this part of `setup.py` needs to be reworked. 😕

wiredfool:
Fixes #2851 .

Apparently Debian strips out all the platform detection code when building .debs, and we're missing support for some arm variants. But since that support is missing, we pop an error when we don't detect the linux version. 

 * Promote the add_multiarch_paths to run before the manual detection, if we know we're on a debian derivative. This is going to continue to fail on non-ubuntu/non-debian debian derivatives. 
 * Refactor a series of elifs into a data structure.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -312,7 +312,15 @@ class pil_build_ext(build_ext):
 
         elif sys.platform.startswith("linux"):
             arch_tp = (plat.processor(), plat.architecture()[0])
-            if arch_tp == ("x86_64", "32bit"):
+            # This should be correct on debian derivatives.
+            if plat.dist()[0].lower() in ('debian', 'ubuntu'):
+                # If this doesn't work, don't just silently patch
+                # downstream because it's going to break when people
+                # try to build pillow from source instead of
+                # installing from the system packages.
+                self.add_multiarch_paths()
+
+            elif arch_tp == ("x86_64", "32bit"):
                 # 32-bit build on 64-bit machine.
                 _add_directory(library_dirs, "/usr/lib/i386-linux-gnu")
             else:
@@ -362,11 +370,6 @@ class pil_build_ext(build_ext):
                 else:
                     raise ValueError(
                         "Unable to identify Linux platform: `%s`" % platform_)
-
-                # XXX Kludge. Above /\ we brute force support multiarch. Here we
-                # try Barry's more general approach. Afterward, something should
-                # work ;-)
-                self.add_multiarch_paths()
 
                 # termux support for android.
                 # system libraries (zlib) are installed in /system/lib
--- a/setup.py
+++ b/setup.py
@@ -321,52 +321,34 @@ class pil_build_ext(build_ext):
                 self.add_multiarch_paths()
 
             elif arch_tp == ("x86_64", "32bit"):
-                # 32-bit build on 64-bit machine.
+                # Special Case: 32-bit build on 64-bit machine.
                 _add_directory(library_dirs, "/usr/lib/i386-linux-gnu")
             else:
+                libdirs = {
+                    'x86_64':  ["/lib64", "/usr/lib64", "/usr/lib/x86_64-linux-gnu"],
+                    '64bit':   ["/lib64", "/usr/lib64", "/usr/lib/x86_64-linux-gnu"],
+                    'i386':    ["/usr/lib/i386-linux-gnu"],
+                    'i686':    ["/usr/lib/i386-linux-gnu"],
+                    '32bit':   ["/usr/lib/i386-linux-gnu"],
+                    'aarch64': ["/usr/lib64", "/usr/lib/aarch64-linux-gnu"],
+                    'arm':     ["/usr/lib/arm-linux-gnueabi"],
+                    'armv71':  ["/usr/lib/arm-linux-gnueabi"],
+                    'ppc64':   ["/usr/lib64", "/usr/lib/ppc64-linux-gnu",
+                                "/usr/lib/powerpc64-linux-gnu"],
+                    'ppc':     ["/usr/lib/ppc-linux-gnu",
+                                "/usr/lib/powerpc-linux-gnu"],
+                    's390x':   ["/usr/lib64", "/usr/lib/s390x-linux-gnu"],
+                    's390':    ["/usr/lib/s390-linux-gnu"],
+                    }
+                    
                 for platform_ in arch_tp:
-
+                    dirs = libdirs.get(platform_, None)
                     if not platform_:
                         continue
-
-                    if platform_ in ["x86_64", "64bit"]:
-                        _add_directory(library_dirs, "/lib64")
-                        _add_directory(library_dirs, "/usr/lib64")
-                        _add_directory(library_dirs,
-                                       "/usr/lib/x86_64-linux-gnu")
-                        break
-                    elif platform_ in ["i386", "i686", "32bit"]:
-                        _add_directory(library_dirs, "/usr/lib/i386-linux-gnu")
-                        break
-                    elif platform_ in ["aarch64"]:
-                        _add_directory(library_dirs, "/usr/lib64")
-                        _add_directory(library_dirs,
-                                       "/usr/lib/aarch64-linux-gnu")
-                        break
-                    elif platform_ in ["arm", "armv7l"]:
-                        _add_directory(library_dirs,
-                                       "/usr/lib/arm-linux-gnueabi")
-                        break
-                    elif platform_ in ["ppc64"]:
-                        _add_directory(library_dirs, "/usr/lib64")
-                        _add_directory(library_dirs,
-                                       "/usr/lib/ppc64-linux-gnu")
-                        _add_directory(library_dirs,
-                                       "/usr/lib/powerpc64-linux-gnu")
-                        break
-                    elif platform_ in ["ppc"]:
-                        _add_directory(library_dirs, "/usr/lib/ppc-linux-gnu")
-                        _add_directory(library_dirs,
-                                       "/usr/lib/powerpc-linux-gnu")
-                        break
-                    elif platform_ in ["s390x"]:
-                        _add_directory(library_dirs, "/usr/lib64")
-                        _add_directory(library_dirs,
-                                       "/usr/lib/s390x-linux-gnu")
-                        break
-                    elif platform_ in ["s390"]:
-                        _add_directory(library_dirs, "/usr/lib/s390-linux-gnu")
-                        break
+                    for path in dirs:
+                        _add_directory(library_dirs, path)
+                    break
+                
                 else:
                     raise ValueError(
                         "Unable to identify Linux platform: `%s`" % platform_)


hugovk:
`+            # Ubunbu <= Precise fails, >= Trusty succeeds.`

Do we care about <= Precise? Trusty is the oldest LTS supported.

image

https://en.wikipedia.org/wiki/Ubuntu_version_history#Table_of_versions

wiredfool:
For some reason, I thought it was good till next year, even after migrating off of it.  It's still in the docker builds, but we could drop that, as we've likely got library coverage for old libraries on the centos build.

wiredfool:
Ok, I've removed debian from the tests, and force rebased this one on top of that.

gerardgilliland:
Wiredfool,

You commented, there's got to be a better way ... But at least you have left enough tracks to help me.
I am attempting to install hyperspy
And thanks to the help of Josh Taillon, Petras Jokubauskas, Magnus Nord, and bochecha
I am down to the last module (I think or at least I hope).

I have the same problem as bochecha:
Can't build Pillow on Debian armhf
https://github.com/python-pillow/Pillow/issues/2851
I can't build Pillow on a Raspberry Pi ARM processor which I believe falls under his category.
The bottom line(s):
  The headers or library files could not be found for jpeg,
  a required dependency when compiling Pillow from source.
And I think the following should say arm rather than 386
    File "/tmp/pip-build-3866ru_p/pillow/setup.py", line 577, in build_extensions
      raise RequiredDependencyException(f)

I believe your fix includes 2851. 
Having never gone beyond the standard install,
How do I avoid waiting for an unknown time for this to show up when I attempt to install ?
This tells me how to contribute (using the master)
https://github.com/python-pillow/Pillow/blob/master/.github/CONTRIBUTING.md
by going to https://github.com/python-pillow/Pillow/pulls and clicking “New Pull Request” 
But I think, master plus your fixes are at:
https://github.com/wiredfool
under Contribution activity 
with the release and date: Release 4.4.0 Issue: 2018-01-01
Could you please point me to any link that would be:
 Pillow master, plus ARM processor selection fix, plus Pillow fixes
That would occur within the next day or two?

Thank you and Peace,
Gerard Gilliland

hugovk:
You can test this out with

git clone https://github.com/wiredfool/Pillow/
cd Pillow
git checkout debian-arm

Then building the source as detailed in the docs.

gerardgilliland:
Hugo,

Thank you for sending exactly what I needed !!

You wrote:

You can test this out with:

 

git clone https://github.com/wiredfool/Pillow/

cd Pillow

git checkout debian-arm

Then building the source as detailed in the docs <https://pillow.readthedocs.io/en/4.3.x/installation.html#building-from-source> .

I added a folder  /home/pi/Hyperspy/Pillow

And followed your (above) instructions which populated the local Pillow folder. (YEA)

then

pi@raspberrypi:~/Hyperspy $ CFLAGS=”-I/home/pi/Hyperspy/Pillow” pip3 install Pillow –no-cache-dir

but it still downloads the old version 4.3.0

 

I moved the Pillow folder to /home/pi/ (out of the Hyperspy folder to the same level)

 

pi@raspberrypi:~ $ pip3 install Pillow –no-cache-dir

 

and tried:

cd Pillow

pi@raspberrypi:~/Pillow $ pip3 install Pillow –no-cache-dir

but it still downloads the old version 4.3.0

 

Obviously, I can NOT follow the instructions in the above link docs.

Could you please help me with the command to build Pillow?  SORRY.

 

Thank you and Peace,

Gerard

hugovk:
Once you have the prerequisites in place, don't 'pip install pillow', which installs the last release from PyPI, but instead run 'python setup.py install' from the cloned Pillow directory.

gerardgilliland:
Hugo,

Thank you for your instructions – they are spot on.

I am going to need your help again !! 

I’m sorry.

See attached installPillow.txt

It started out good but ended with:

 

The headers or library files could not be found for jpeg,

a required dependency when compiling Pillow from source.

 

Reading: https://pillow.readthedocs.io/en/latest/installation.html

it looks like I need to install libjpeg which I am NOT capable of doing.

I did find libjpeg.so but don’t know what to do with it.

 

The internet (link is in the attached file) said by installing python-imaging, libjpeg would be installed as part installation.

But it didn’t seem to help.

Maybe I’m hung up on libjpeg and should be trying other options.

I hope the bochecha / Wiredfool fix includes the libjpeg install and it is my failure to select the proper prerequisites.

 

Thank you and Peace,

Gerard

hugovk:
Merging.

@gerardgilliland Would you like to open a new issue?

https://github.com/python-pillow/Pillow/issues/new

gerardgilliland:
Hugo:
You asked: “Would I like to open a new issue ?”

With your help, Yes. 

I know how to install and run 
git clone https://github.com/wiredfool/Pillow/
cd Pillow
git checkout debian-arm
and run 
python setup.py install
from the cloned Pillow directory on Raspberry Pi (with your help).
However, the above fails to build Pillow.
The headers or library files could not be found for jpeg,
a required dependency when compiling Pillow from source.

However, No.

I am simply testing the app and the original failure defined by bochecha still exists.
In other words I was getting this same message the same day #2851 was opened.
And the same other errors as well.
And the same errors with wiredfool’s new code that you linked me to.
So, I would be causing confusion.
I think when other testers on different systems check in to run the same code you linked me to, 
They would have the same error as I am getting.
Because I believe bochecha still “Can't build Pillow on Debian armhf”.
If #2853 can’t be changed even though is doesn’t solve the problem,
Then I believe bochecha will open a new issue.
And I will certainly comment on it, that I have the same problem on a Raspberry Pi.

So, Maybe.

I will look at wiredfool’s code in the cloned Pillow folder.
And see if I can get the 
python setup.py install
to run to completion.

That is my perspective.  Please correct me where I am wrong.

Thank you and Peace,
Gerard

setup.py
```diff
@@ -312,62 +312,47 @@ def build_extensions(self):
 
         elif sys.platform.startswith("linux"):
             arch_tp = (plat.processor(), plat.architecture()[0])
-            if arch_tp == ("x86_64", "32bit"):
-                # 32-bit build on 64-bit machine.
+            # This should be correct on debian derivatives.
+            if plat.dist()[0].lower() in ('debian', 'ubuntu'):
```
bochecha: `platform.dist()` is deprecated and will be removed in Python 3.7

wiredfool: Great. Some batteries included I guess.

Issue #2842, opened 2017-11-08
Installation issues on Windows/Python 3.7


Mattwmaster58:
### What did you do?
Admin CMD console > `python -m pip install pillow --trusted-host='pypi.python.org'`
### What versions of Pillow and Python are you using?
Python Version 3.7, latest pip
Latest Pillow version (4.3?)
Windows 10 Pro

I think the most relevant part from this error is this part:
```
The headers or library files could not be found for zlib,
 a required dependency when compiling Pillow from source.
```
and yet typing `import zlib` into the python console produces no errors. 
Is it b/c the Appdata dir is `c:\users\marcus family\appdata\local\` half the time and the other times it's `C:\Users\MARCUS~1\AppData\Local\`?

Here is the full output from `python -m pip install pillow --trusted-host='pypi.python.org"`
<details>
  <summary>Here is the full output from this command</summary>
<code>
Collecting pillow
  Downloading Pillow-4.3.0.tar.gz (13.9MB)
    100% |████████████████████████████████| 13.9MB 7.9MB/s
Requirement already satisfied: olefile in c:\users\marcus family\appdata\local\programs\python\python37\lib\site-packages (from pillow)
Installing collected packages: pillow
  Running setup.py install for pillow ... error
    Complete output from command "C:\Users\Marcus Family\AppData\Local\Programs\Python\Python37\python.exe" -u -c "import setuptools, tokenize;__file__='C:\\Users\\MARCUS~1\\AppData\\Local\\Temp\\pip-build-9mcv5007\\pillow\\setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" install --record C:\Users\MARCUS~1\AppData\Local\Temp\pip-au15nuhi-record\install-record.txt --single-version-externally-managed --compile:
    Single threaded build for windows
    running install
    running build
    running build_py
    creating build
    creating build\lib.win-amd64-3.7
    creating build\lib.win-amd64-3.7\PIL
    copying PIL\BdfFontFile.py -> build\lib.win-amd64-3.7\PIL
    copying PIL\BmpImagePlugin.py -> build\lib.win-amd64-3.7\PIL
    copying PIL\BufrStubImagePlugin.py -> build\lib.win-amd64-3.7\PIL
    copying PIL\ContainerIO.py -> build\lib.win-amd64-3.7\PIL
[...]

PR #2855, opened 2017-11-14
Warn if trying to install for Py3.7 on Windows

hugovk:
Fixes https://github.com/python-pillow/Pillow/issues/2842.


wiredfool:
So, is this going to warn us as we try to do CI on windows? (assuming that we get the VS2015 builds working)

hugovk:
Yes, it should do on 3.7.

PR #2858, opened 2017-11-15
Correct error in ImageDraw documentation

meribold:
The first parameter of several functions is described as "four points". It should be "two points" or "four coordinates".


hugovk:
Thank you!

PR #2861, opened 2017-11-17
Update docs to reflect that font can be loaded from a stream, not jus…

robin-norwood:
…t by filename

I want to load my font from an io.BytesIO stream. It turns out that ImageFont already supports this, but the docs weren't clear on that matter.

Fixes # .

 * Just a docs update

### DIFF
--- a/PIL/ImageFont.py
+++ b/PIL/ImageFont.py
@@ -233,14 +233,14 @@ def load(filename):
 def truetype(font=None, size=10, index=0, encoding="",
              layout_engine=None):
     """
-    Load a TrueType or OpenType font from a file or stream, and create
-    a font object.
-    This function loads a font object from the given file or stream,
-    and creates a font object for a font of the given size.
+    Load a TrueType or OpenType font from a file or file-like object,
+    and create a font object.
+    This function loads a font object from the given file or file-like
+    object, and creates a font object for a font of the given size.
 
     This function requires the _imagingft service.
 
-    :param font: A filename or binary stream containing a TrueType font.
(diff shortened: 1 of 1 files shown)


PIL/ImageFont.py
```diff
@@ -233,15 +233,16 @@ def load(filename):
 def truetype(font=None, size=10, index=0, encoding="",
              layout_engine=None):
     """
-    Load a TrueType or OpenType font file, and create a font object.
-    This function loads a font object from the given file, and creates
-    a font object for a font of the given size.
+    Load a TrueType or OpenType font from a file or stream, and create
+    a font object.
```
wiredfool: I think we can just change this instance to `file like object`, and leave the next one as file.

PIL/ImageFont.py
```diff
@@ -233,15 +233,16 @@ def load(filename):
...
+    This function loads a font object from the given file or stream,
+    and creates a font object for a font of the given size.
 
     This function requires the _imagingft service.
 
-    :param font: A truetype font file. Under Windows, if the file
-                     is not found in this filename, the loader also looks in
-                     Windows :file:`fonts/` directory.
+    :param font: A filename or binary stream containing a TrueType font.
+                     Under Windows, if the file is not found in this filename,
```
wiredfool: Here are think we should just change binary stream to file like object.

PIL/ImageFont.py
hugovk: Nitpick: `file-like object` for both.

robin-norwood:
Changed the wording as requested.

PR #2864, opened 2017-11-18
Add eog support for Ubuntu Image Viewer

NafisFaysal:
## Context:
*  As a Ubuntu user, ```~PIL.Image.Image.show``` method shows the image, but the default Image Viewer of Ubuntu using the command ```eog``` in the terminal. By default, Pillow looking for the commands ```xv``` and ```display``` .

 *  Add ```eog``` support for the Ubuntu. 
 *  Now Ubuntu user can see the image and it is very handy for debugging and tests.

### DIFF
--- a/PIL/ImageShow.py
+++ b/PIL/ImageShow.py
@@ -101,6 +101,7 @@ class Viewer(object):
 
 # --------------------------------------------------------------------
 
+
 if sys.platform == "win32":
 
     class WindowsViewer(Viewer):
@@ -161,14 +162,15 @@ else:
             command = executable = "display"
             return command, executable
 
+    if which("display"):
+        register(DisplayViewer)
+
     class EogViewer(UnixViewer):
         def get_command_ex(self, file, **options):
             command = executable = "eog"
             return command, executable
 
-    if which("display"):
-        register(DisplayViewer)
-    elif which("eog"):
+    if which("eog"):
         register(EogViewer)
 
     class XVViewer(UnixViewer):
--- a/PIL/ImageShow.py
+++ b/PIL/ImageShow.py
@@ -156,17 +156,20 @@ else:
 
     # implementations
 
-    # The default Image Viewer of Ubuntu using the command "eog" in the terminal.
-    # Reference: https://help.gnome.org/users/eog/stable/commandline.html.en
-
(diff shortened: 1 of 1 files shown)


PIL/ImageShow.py
```diff
@@ -156,12 +156,16 @@ def show_file(self, file, **options):
 
     # implementations
 
+    # The default Image Viewer of Ubuntu using the command "eog" in the terminal.
+    # Reference: https://help.gnome.org/users/eog/stable/commandline.html.en
+
     class DisplayViewer(UnixViewer):
         def get_command_ex(self, file, **options):
-            command = executable = "display"
+            command = executable = "display" or "eog"
```
hugovk: Doesn't this just evaluate to "display"?

NafisFaysal: yap just evaluate to display  :smiley:

hugovk: I understand that, but what's the point of changing this line? `command == display` with or without this change.

NafisFaysal: If I got you, I keep display for the other *nix user.

hugovk: On every Unix, `command` will always be `"display"`. It will never be `"eog"`. 

Have you tested this PR? Does it open eog?

NafisFaysal: Yah I test it on my Ubuntu machine, If I add ```eog``` only that time ```PIL.Image.Image.show``` work otherwise ```PIL.Image.Image.show``` don't work :)

hugovk: Maybe a new class EogViewer.

```python
    class EogViewer(UnixViewer):
        def get_command_ex(self, file, **options):
            command = executable = "eog"
            return command, executable
```
And then:
```python
    if which("display"):
        register(DisplayViewer)
    elif which("eog"):
        register(EogViewer)
```

Please test it works.

NafisFaysal: @hugovk  thanks but coincidentally I was writing the same code after your comment ```command = executable = "display" or "eog"```

NafisFaysal:
After update the code now, everything is fine and working.

hugovk:
I've added a commit to reorder/refactor for consistency with the viewers.

PR #2872, opened 2017-11-23
removed olefile from requirements.txt

stuertz:
as it is no longer a requirement. Its optional.

see also https://github.com/python-pillow/Pillow/commit/9175706300dc431fd8573305a1ef8e78fd55bc21#diff-2eeaed663bd0d25b7e608891384b7298

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -13,7 +13,6 @@ Jinja2
 MarkupSafe
 nose
 nose-cov
-olefile
 pycodestyle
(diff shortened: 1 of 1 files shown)


wiredfool:
Yes, it's optional, but it should stay in `requirements.txt`. The `requirements.txt` is dev time requirements, most of the other stuff there is for building docs and running the tests. It's not going to get pulled in when people are installing from pip.

PR #2873, opened 2017-11-25
Removed debugging code

radarhere:


wiredfool:
I'm not sure that this is debugging code as much as preflight for determining if it's possible to save the image in the specified format.

Issue #363, opened 2013-10-04
Test_file_webp fails on Windows


cgohlke:
Current master fails in test_file_webp. This test passed with version 2.2.1. Probably due to PR #359. Tested on win-amd64-py3.3:

```
running test_file_webp ...
Tests\test_file_webp.py:66: assert_image_similar(image, target, 20.0) failed:
- average pixel value difference 42.4100 > epsilon 20.0000
Tests\test_file_webp.py:93: assert_image_similar(image, pil_image, 1.0) failed:
- average pixel value difference 13.0000 > epsilon 1.0000
```

Issue #353, opened 2013-09-30
Platform tests for 2.2.0


wiredfool:
I've tested, and are working:
- Ubuntu 10.04LTS w/ py 2.6.5 (system python)
- Raspian Wheezy w/ py 2.7.3 (system python)
- Raspian Wheezy w/ py 3.2.3 (alternate system python)

Issue #343, opened 2013-09-28
Drop support for LCMS version 1


aclark4life:
Due to security issue: http://bugs.debian.org/cgi-bin/bugreport.cgi?bug=718682

Issue #332, opened 2013-09-24
Document 'import Image' hack-around


aclark4life:
http://reinout.vanrees.org/weblog/2013/09/24/pil-pillow-image-import.html

PR #2880, opened 2017-12-06
 Updated libwebp to 0.6.1

radarhere:
A test failed after the update with an average pixel value different of 21.3764, so I increased the threshold to 25.

### DIFF
--- a/Tests/test_file_webp_alpha.py
+++ b/Tests/test_file_webp_alpha.py
@@ -119,7 +119,7 @@ class TestFileWebpAlpha(PillowTestCase):
         image.getdata()
         target = Image.open(file_path).convert("RGBA")
 
-        self.assert_image_similar(image, target, 10.0)
+        self.assert_image_similar(image, target, 25.0)
 
 
 if __name__ == '__main__':
--- a/depends/install_webp.sh
+++ b/depends/install_webp.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 # install webp
 
-archive=libwebp-0.6.0
+archive=libwebp-0.6.1
 
 ./download-and-extract.sh $archive https://raw.githubusercontent.com/python-pillow/pillow-depends/master/$archive.tar.gz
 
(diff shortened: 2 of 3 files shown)


Tests/test_file_webp_alpha.py
```diff
@@ -119,7 +119,7 @@ def test_write_unsupported_mode_PA(self):
         image.getdata()
         target = Image.open(file_path).convert("RGBA")
 
-        self.assert_image_similar(image, target, 10.0)
+        self.assert_image_similar(image, target, 25.0)
```
wiredfool: That's a pretty big jump for a .0.1 release, are we sure that we're not missing something there?

radarhere: https://github.com/webmproject/libwebp/blob/master/NEWS
> - 11/24/2017: version 0.6.1
>   This is a binary compatible release.
>   * lossless performance and compression improvements + a new 'cruncher' mode
>     (-m 6 -q 100)
>   * ARM performance improvements with clang (15-20% w/ndk r15c, issue #339)
>   * webp-js: emscripten/webassembly based javascript decoder
>   * miscellaneous bug & build fixes (issue #329, #332, #343, #353, #360, #361,
>     #363)
>   Tool updates / additions:
>     added webpinfo - prints file format information (issue #330)
>     gif2webp - loop behavior modified to match Chrome M63+ (crbug.com/649264);
>                '-loop_compatibility' can be used for the old behavior

Here's a copy of the generated file from my local machine using 0.6.1.
transparent

wiredfool:
This is now a blocker for the fedora-27 docker tests. https://github.com/python-pillow/docker-images/pull/20 and #2895

Tests/test_file_webp_alpha.py
wiredfool: I have no idea where those issues actually reside on the web, other than not on github, and not in the repo it's cloned from. 

Anyway, it looks like there was a YUV->RGB change in 0.52 that could potentially account for that.

Issue #2503, opened 2017-05-05
Invalid colors in TIF format


AlexLazov:
Some TIFF images are displayed incorrectly.
I'm using the latest version Pillow.

![dd03](https://cloud.githubusercontent.com/assets/26875658/25739924/54979196-31ae-11e7-81c3-caee6e65bb8b.jpg)

Images TIF:
https://yadi.sk/d/SpI9WGU-3Hh6KQ

wiredfool:
This looks something like a color space issue, almost as if we're interpreting YCrCb (the default jpeg space) as RGB (which is what the image is tagged as).  I'm not seeing any place where we're obviously going wrong, so ... more digging required.

AlexLazov:
I already understood this. If you open using the library, everything is correctly displayed (READ_LIBTIFF = True). By the way, the problem was discussed before.

wiredfool:
I'm not seeing the previous discussion in github's search.

AlexLazov:
We can solve the problem like this.
In the condition "_if READ_LIBTIFF or self._compression in ["tiff_ccitt", "group3", ..._" add "_jpeg_". Then everything will work.
File TiffImagePlugin.py, line 1194.

``if READ_LIBTIFF or self._compression in ["tiff_ccitt", "group3",
                                                     "group4", "tiff_jpeg",
                                                     "tiff_adobe_deflate",
                                                     "tiff_thunderscan",
                                                     "tiff_deflate",
                                                     "tiff_sgilog",
                                                     "tiff_sgilog24",
                                                     "tiff_raw_16","jpeg"]:``

hugovk:
Fixed in https://github.com/python-pillow/Pillow/pull/2899.

PR #2882, opened 2017-12-07
Add jpg value in TiffImagePlugin in compression list

Thorleon:
Fix for wrong colors when jpeg embedded in tiff
https://github.com/python-pillow/Pillow/issues/2503


wiredfool:
It looks like there's a test that's either too specific, or it's flagging an actual change in behavior. 

I'd also say that at this point, it's probably better to just use libtiff if there's a compressed image at all, given that we know that there are some odd corners in the compression algorithms that are implemented in Pillow.

wiredfool:
Closing in favor of #2899

PR #2885, opened 2017-12-10
Restored deleted lines

radarhere:
Restores lines deleted in https://github.com/python-pillow/Pillow/commit/eccf9f624b60e32e4f24c0afd20f172795914bd9 and https://github.com/python-pillow/Pillow/commit/db88c6c3161f174ede97910ffda66a3e08e662a1


wiredfool:
I'm merging this, but I think we should take the changes and split off anything that's too old, possibly pre-fork, and dump it into the docs somewhere with a reference.

PR #2888, opened 2017-12-15
Fixed typos

radarhere:


wiredfool:
I'm not sure that's a typo, I think it's more that I totally missed their branding.

PR #2890, opened 2017-12-16
Make PngImagePlugin.add_text() zip argument type bool

jdufresne:
Always used as a `bool`, but was previously defined as 0/1. Use modern idiomatic Python by using the `bool` type for `bool` arguments.

### DIFF
--- a/PIL/PngImagePlugin.py
+++ b/PIL/PngImagePlugin.py
@@ -249,7 +249,7 @@ class PngInfo(object):
             self.add(b"iTXt", key + b"\0\0\0" + lang + b"\0" + tkey + b"\0" +
                      value)
 
-    def add_text(self, key, value, zip=0):
+    def add_text(self, key, value, zip=False):
         """Appends a text chunk.
 
         :param key: latin-1 encodable text key name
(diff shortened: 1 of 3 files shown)


PIL/PngImagePlugin.py
```diff
@@ -249,7 +249,7 @@ def add_itxt(self, key, value, lang="", tkey="", zip=False):
             self.add(b"iTXt", key + b"\0\0\0" + lang + b"\0" + tkey + b"\0" +
                      value)
 
-    def add_text(self, key, value, zip=0):
+    def add_text(self, key, value, zip=False):
```
wiredfool: As an aside, this is a really badly named parameter, as it shadows a builtin. However, since it's (nominally) a named parameter in a public interface, changing it is potentially going to cause more pain than it's worth.

jdufresne: I agree, but I wasn't attempting to change the public interface. I agree, that may be worthwhile to avoid the builtin.

wiredfool: Yep.

PR #2894, opened 2017-12-19
Test: avoid random noise failure

hugovk:
Just occasionally (e.g. the [last master build](https://travis-ci.org/python-pillow/Pillow/jobs/318081653#L3650)) `test_effect_noise` will fail because the random noise will put the same value in these pixels:

```python
self.assertNotEqual(im.getpixel((0, 0)), im.getpixel((0, 1)))
```

 * Roll again if we get the same value, 5 attempts 
 * flake8

(The randomness comes from libImaging/Effects.c. If it were Python code we could do `from random import Random; random = Random(1538)`.)

### DIFF
--- a/Tests/test_image.py
+++ b/Tests/test_image.py
@@ -348,24 +348,19 @@ class TestImage(PillowTestCase):
         # Arrange
         size = (100, 100)
         sigma = 128
-        # To reduce chance of randomly having the same value twice
-        attempts = 5
 
         # Act
-        while attempts > 0:
-            im = Image.effect_noise(size, sigma)
+        im = Image.effect_noise(size, sigma)
 
-            # Assert
-            self.assertEqual(im.size, (100, 100))
-            self.assertEqual(im.mode, "L")
-            p0 = im.getpixel((0, 0))
-            p1 = im.getpixel((0, 1))
-            if p0 == p1:
-                # Let's roll again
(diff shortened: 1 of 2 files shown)


Tests/test_image.py
```diff
@@ -347,14 +347,24 @@ def test_effect_noise(self):
         # Arrange
         size = (100, 100)
         sigma = 128
+        # To reduce chance of randomly having the same value twice
```
wiredfool: Instead of this approach, maybe we should do something like assert_image_similar, but with a minimum or expected difference.

hugovk: The image will be something like this: 
image

For `assert_image_similar` the epsilon is 84-87. So a new `assert_image_different(self, a, b, minimum, msg=None)`?

Or how about taking 3 or more pixels and assert they're not all the same?

wiredfool: There should be a pretty good relationship between sigma and the epsilon that we see, so checking more than three pixels will lead to a better test, and could potentially uncover sigma being treated incorrectly. 

However, this is a pretty low impact feature, so I'm not sure how much testing is really necessary.

hugovk: Replaced it with an `assert_not_all_same([p0, p1, p2, p3, p4])`

PR #2911, opened 2017-12-27
RFC: Source layout reorg

wiredfool:
We've had a bunch of issues with running tests and whatnot from the project directory because our PIL directory doesn't contain the compiled extensions unless you develop in-place. (which I object to, because it leads to problems with having the correct binaries for multiple python versions.)

This PR places all the source in a src directory below the project directory. 

This:
* Makes it so that you don't need the --installed on selftest. 
* Pytests' xdist extension works out of the box
* Individual tests can be run with `python Tests/test_foo.py` 
* `python setup.py develop` still works.

In short, it makes `make install` work in a sane manner, without breaking other people's workflow. 

Unfortunately, it also:
* Moves everything. 
* Breaks history for individual files unless you use --follow

I'd rather do this at a major release than some other time, so time is short to say yes or no to this.


aclark4life:
@wiredfool Thanks for doing this! I'm OK with breaking changes for the sake of moving forward, with less confusion and less historical baggage.

hugovk:
Pytest strongly recommend this https://docs.pytest.org/en/latest/goodpractices.html and point to https://blog.ionelmc.ro/2014/05/25/python-packaging/#the-structure for more benefits.

Not really an issue for me as I usually use a Mac, but is this going to make things harder to develop on Windows, where it's difficult to set up the compiler? As it is, if you don't need to modify C code, you can edit Python code and tests and it'll happily use the compiled C stuff from the the last release pip installed.

aclark4life:
@hugovk We can probably clear that up by asking @cgohlke to comment

radarhere:
No objections. I'll just say the obvious though - this will break almost every other PR.

wiredfool:
@hugovk It looks like the first _build_ on windows did succeed, it's just that it failed to move the PIL directory out of the way to force test runs against the installed version because the directory isn't there anymore.  Similarly, the tests on linux failed due to the PIL directory having been moved, and coverage not being able to find the PIL directory.

I think this matches at least one of the pytest recommended source layouts, certainly better than the old layout. 

@radarhere Yes, it does break PRs, but it also solves a long running pain point. We're at the lowest inventory of PRs in a long while, and we can do this one after the big ones we're merging in this window.

wiredfool:
This branch works on all the docker images OMM, with the pytest branch merged into this one.

aclark4life:
@radarhere Yeah, actually I appreciate you pointing that because I wasn't thinking about it; but I think it's OK too, because it should be easy for PR authors to update their PRs after this change.

hugovk:
Shall we move_tkmini.h and tkImaging.c from Tk/ as well? 

https://github.com/python-pillow/Pillow/tree/master/Tk

wiredfool:
Yeah, that would make sense.

PR #2922, opened 2018-01-02
Skip outline if the draw operation fills with the same colour

radarhere:
If the outline colour in a draw operation is the same as the fill colour, skip drawing the outline.


hugovk:
Do we have regression tests to cover these changes where the outline and fill colours are the same?

src/PIL/ImageDraw.py
```diff
@@ -110,7 +110,7 @@ def _getink(self, ink, fill=None):
                 if self.palette and not isinstance(ink, numbers.Number):
                     ink = self.palette.getcolor(ink)
                 ink = self.draw.draw_ink(ink, self.mode)
-            if fill is not None:
+            if fill is not None and ink != fill:
```
wiredfool: Fill and ink are variable type values, e.g. 'white' and (255,255,255) are equivalent, but not equal. You can only really compare them for equality after the ImageColor.getcolor call.

radarhere: Okay, thanks. I've removed that line change.

radarhere:
Okay, I've added tests.

PR #2936, opened 2018-01-06
Replaced exec with getattr in ImageCms

radarhere:


wiredfool:
I think we're better off dropping the __main__ portion all together.  There are plenty of ways to do docs without executing a module.

radarhere:
Okay, sure. I've created a new PR.

Issue #2718, opened 2017-09-08
`Pillow` vs `cv2` resize


TropComplique:
Results of reading and resizing can be different in `cv2` and `Pilllow`. This creates problems when you want to reuse a model (neural network) trained using `cv2` with `Pillow`.
#### Case 1
Different results of resizing an image.
```
import cv2
from PIL import Image
import numpy as np

# read images
a = cv2.imread('images/dogs-and-cats.jpg', cv2.IMREAD_COLOR)
b = Image.open('images/dogs-and-cats.jpg')

# they are equal
print((np.asarray(b)[:, :, [2, 1, 0]] == a).all())
# True

# resize images
b = b.resize((300, 300), Image.NEAREST)
a = cv2.resize(a, (300, 300), interpolation=cv2.INTER_NEAREST)

# results of resizing are different, but visually they are equal
b = np.asarray(b)[:, :, [2, 1, 0]]
print((b - a).mean())
# 71.1218444444
```
#### Case 2
Different results of reading an image.
```
import cv2
from PIL import Image
import numpy as np

a = cv2.imread('images/cat_train.jpg', cv2.IMREAD_COLOR)

b = Image.open('images/cat_train.jpg')
b = np.asarray(b)[:, :, [2, 1, 0]]

# results of reading are not equal
print((b == a).all())
# False
print((b == a).mean())
# 0.999996744792
```
I use: `Python 3.6`, `Pillow 4.2.1`, `opencv-python 3.3`

wiredfool:
Case 1: Nearest neighbor is a fast, low quality, best effort interpolation. That said, I believe that our tests show our implementation is reasonably correct. a mean difference of 71 would fall into the range of seriously visually different, and probably the result of either a rotation or channel swapping. 

Case 2. That appears to be 1 pixel level difference in 10^6 pixels. That's pretty close, and well within what I'd consider reasonable for a lossy format with potentially different implementations and settings on the decoder.

Feel free to point out where you feel that the implementation is incorrect.

TropComplique:
For case 1 `Image.fromarray(a)` and `Image.fromarray(b)` show the same image.
Also, their values are kinda close:
```
print(a[41:55, 60, :])
# [[255 255 255]
#  [117 124 144]
#  [ 96 117 145]
#  [ 70  90 121]
#  [ 68  88 119]
#  [ 78  96 127]
#  [ 70  88 119]
#  [ 71  87 116]
#  [ 71  89 118]
[...]

PR #2939, opened 2018-01-07
Geometry: affine_fixed breaks Minecraft Overviewer since Pillow 3.4.0

lubosz:
Commit 5232361718bae0f0ccda76bfd5b390ebf9179b18 introduced changes to the affine_fixed function
which lead to Minecraft Overviewer issue https://github.com/overviewer/Minecraft-Overviewer/issues/1370

This patch reverts the change.

Not sure if this is the right fix in terms of the original patch, but I wanted to point it out. so a proper fix can be discussed.

### DIFF
--- a/src/libImaging/Geometry.c
+++ b/src/libImaging/Geometry.c
@@ -862,10 +862,8 @@ affine_fixed(Imaging imOut, Imaging imIn,
 /* use 16.16 fixed point arithmetics */
 #define FIX(v) FLOOR((v)*65536.0 + 0.5)
 
-    a0 = FIX(a[0]); a1 = FIX(a[1]);
-    a3 = FIX(a[3]); a4 = FIX(a[4]);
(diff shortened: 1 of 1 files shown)


hugovk:
Latest comment in https://github.com/overviewer/Minecraft-Overviewer/issues/1370#issuecomment-355799643:

> Okay, so looks like they've fixed a bug that we may have either been relying on or working around ourselves already by fudging coordinates, so it'd probably be wise to look at fixing this in Overviewer

Let's give them some time to investigate in Overviewer first.

homm:
@lubosz Why do you think the old implementation is more correct?

Possibly related: #2718

lubosz:
Thanks for the quick response.

@homm I didn't think, just bisected. It may turn out overviewer was working around the original bug and this can be just closed. I will take a look into overviewer.

wiredfool:
I'm leaning to Works as Intended here, and closing this PR.

PR #2941, opened 2018-01-08
Continuation of WIP: Type Annotations

wiredfool:
Fixes #2687

* make typecheck, i.e. `mypy -2 --ignore-missing-imports src/PIL/Image.py` works


radarhere:
Just to make the link, the starting issue for this topic is #2625

wiredfool:
I'll look at updating these, but the big question is now -- Should we merge this? 

@python-pillow/pillow-team ?

aclark4life:
With all the work done by you and @neiljp (and others?) and at least one other comment indicating this is something people want, I'd say merge-away for April 1 release.

hugovk:
I've not used typing in Python before, but no objections. 

Of course, in the future we'll need to make sure new/changed code updates the types. Some instructions in CONTRIBUTING.md would be helpful, and we'll need to make sure contributors' PRs keep things updated, or update them ourselves.

I think it would be useful to check the typing on the CI. I'd suggest a new Travis job that just does the type checking, and no other tests. That way it would run and pass or fail quickly, and wouldn't muddy the waters of whether the actual unit tests pass or fail.

src/PIL/Image.py
```diff
@@ -24,6 +24,22 @@
 # See the README file for information on usage and redistribution.
 #
 
+if False:
```
hugovk: Is this `if False:` going to remain? Does it need to?

A comment to explain why would help.

src/PIL/Image.py
```diff
@@ -65,7 +82,7 @@ def __getattr__(self, id):
                            PILLOW_VERSION))
 
 except ImportError as v:
-    core = _imaging_not_installed()
+    core = _imaging_not_installed() # type: ignore
```
hugovk: pep8: two spaces before inline comment, here and elsewhere

src/PIL/Image.py
```diff
@@ -706,21 +743,23 @@ def tobytes(self, encoder_name="raw", *args):
 
         :param encoder_name: What encoder to use.  The default is to
                              use the standard "raw" encoder.
-        :param args: Extra arguments to the encoder.
+        :param *args: Extra arguments to the encoder. May be one tuple,
+                      or individual arguments
```
hugovk: End sentence with `.` for consistency

src/PIL/Image.py
```diff
@@ -2116,25 +2243,31 @@ def transform(self, size, method, data=None, resample=NEAREST,
...
 
         if data is None:
             raise ValueError("missing method data")
 
         im = new(self.mode, size, fillcolor)
         if method == MESH:
             # list of quads
-            for box, quad in data:
+            data_mesh = None # type: List[Tuple[LURD,LURD]]
+            data_mesh = data # type: ignore
```
hugovk: What's going on here, setting `data_mesh` twice?

neiljp: Yes, as I mentioned above. If you wish to maintain the previous comment of what it "should" be, ideally, then perhaps just put it as the 'post-ignore' comment?

src/PIL/Image.py
```diff
@@ -2225,8 +2362,9 @@ def toqpixmap(self):
 
 class ImagePointHandler(object):
     # used as a mixin by point transforms (for use with im.point)
-    pass
-
+    def point(self, s):
+        # type (T) -> T
```
neiljp: What is `T` here? This doesn't look familiar from when I was working on this. My instinct is a `TypeVar`, but `T` is not very descriptive?

src/PIL/Image.py
```diff
@@ -2436,16 +2584,23 @@ def fromarray(obj, mode=None):
...
         else:
-            obj = obj.tostring()
+            obj_bytes = obj.tostring()
 
-    return frombuffer(mode, size, obj, "raw", rawmode, 0, 1)
+    # Typing: At this point, obj should be an item exporting a buffer interface
+    # and obj_bytes, if it exists, is a bytes exporting a buffer interface.
+    # Unfortunately at this point, there's no protocol for the buffer interface
+    # and due to it's c-level implementation, I don't see how to make one.
+    return frombuffer(mode, size, obj_bytes or obj, "raw", rawmode, 0, 1) # type: ignore
```
neiljp: Maybe file an issue in typeshed re this, if there's not one already? If typeshed is the best place - maybe even python/typing? (for more general language/typing queries)

src/PIL/Image.py
```diff
@@ -1364,48 +1436,68 @@ def paste(self, im, box=None, mask=None):
...
-            im = ImageColor.getcolor(im, self.mode)
-
-        elif isImageType(im):
-            im.load()
-            if self.mode != im.mode:
-                if self.mode != "RGB" or im.mode not in ("RGBA", "RGBa"):
-                    # should use an adapter for this!
-                    im = im.convert(self.mode)
-            im = im.im
+            box_lurd = box_coord + (box_coord[0]+size[0], box_coord[1]+size[1]) # type: ignore
```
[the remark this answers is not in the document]
hugovk: No, I've never seen it before.

.gitignore
```diff
@@ -78,3 +78,4 @@ Tests/images/README.md
 Tests/images/msp
 Tests/images/picins
 Tests/images/sunraster
+/.mypy_cache/
```
homm: This section of .gitignore is for "Extra test images installed from pillow-depends/test_images". Better place is "Unit test / coverage reports"

src/PIL/ImageFilter.py
```diff
@@ -19,8 +19,8 @@
 
 
 class Filter(object):
-    pass
-
+    def filter(self, image):
+        return image.im
```
homm: The image argument is already an image core object.

src/PIL/Image.py
wiredfool: Basically, the first sets the type, and the second actually sets the variable, but we're forcing the typechecker to accept that we know that data is the correct type. This sort of idiom is required(?) when we're unpacking a datastructure that's goat a bunch of type overloads.

src/PIL/Image.py
```diff
@@ -1432,17 +1524,22 @@ def alpha_composite(self, im, dest=(0,0), source=(0,0)):
...
-        if source == (0,0) + im.size:
+        if source_lurd == (0,0) + im.size:
             overlay = im
         else:
-            overlay = im.crop(source)
+            overlay = im.crop(source_lurd)
 
         # target for the paste
-        box = dest + (dest[0] + overlay.width, dest[1] + overlay.height)
+        box = None # type: LURD
```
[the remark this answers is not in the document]
wiredfool: Yes, lines of that form were necessary when I was doing this. There may be a better way of doing it, but this is the "define the type, then set it" when we're setting with a complex expression or splitting apart something that is a multiple type union.

src/PIL/Image.py
```diff
@@ -1432,17 +1524,22 @@ def alpha_composite(self, im, dest=(0,0), source=(0,0)):
         if min(dest) < 0:
             raise ValueError("Destination must be non-negative")
 
+        source_lurd = None # type: LURD
+
         if len(source) == 2:
-            source = source + im.size
+            source_lurd = source + im.size # type: ignore
```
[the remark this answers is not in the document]
wiredfool: IIRC, I couldn't get the typechecker to accept that we'd checked that this was a 2-tuple + a 2-tuple yeilding a 4-tuple.

src/PIL/Image.py
```diff
@@ -1261,11 +1326,12 @@ def getpalette(self):
             if bytes is str:
                 return [i8(c) for c in self.im.getpalette()]
             else:
-                return list(self.im.getpalette())
+                return list(self.im.getpalette()) # type: ignore
```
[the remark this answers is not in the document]
wiredfool: There are several ignores here that may be mypy bugs. I did put those in when I'd spent enough time on it, and it's clear what types we had but not clear how to force mypy to deal with it.

wiredfool:
>One thing I am really intrigued by, is the experience of anyone involved of the process - have you identified any latent issues in doing this? is this primarily good for documentation? pushing you towards simpler/better implementations? ...or just an interesting experiment that may as well get merged at this point?

I wouldn't say that it's pushing things to be a simpler implementation, because in many cases we're stuck with the interface we have, and that interface was not designed to be type safe. There are a few idioms, like a single type static assignment that work, and are at least one way to do things. But they aren't simple, they don't seem pythonic, and they obscure the flow of what's going on. 

I could see where this might guide IDEs to make better suggestions. I can't recall if this process actually uncovered any bugs in Pillow, but my feeling at the end of it was more that this was an interesting exercise than a leap forward in code quality.

src/PIL/Image.py
and-semakin: it should be this:

```python
import typing

if typing.TYPE_CHECKING:
    ...
```

Stadly:
What's the status on this? Type annotations for Pillow would be great!

hugovk:
This has conflicts and a lot has changed in typing since this was opened. Let's close in favour of a more incremental approach along the lines of https://github.com/python-pillow/Pillow/issues/2625#issuecomment-1765150919. Thanks anyway for all your work here.

PR #2947, opened 2018-01-10
Codecov: Avoid "Missing base report"

hugovk:
This is an attempt to avoid "Missing base report" due to committing CHANGES.rst with "[CI skip]".

For example, PR https://github.com/python-pillow/Pillow/pull/2946 points to https://codecov.io/gh/python-pillow/Pillow/compare/4bde7258d4eddb221ce905b89d6a57c61fc89fa1...8f88d6b60acb68ecf526a3e9d5ec178a75d702ec:

image

---

Also, the file can be called codecov.yml or .codecov.yml, so let's hide it away with the other dotfiles.

https://docs.codecov.io/docs/codecov-yaml#section-can-i-name-the-file-codecov-yml-

PR #2948, opened 2018-01-10
Codecov: Avoid "Missing base report"

hugovk:
Same as https://github.com/python-pillow/Pillow/pull/2947 but without the magic word.


hugovk:
(I only clicked merge on #2947 but GitHub recognised they both have the same commit hash so marked this as merged too.)

PR #2952, opened 2018-01-12
Added __main__ check to mp_compile.py

radarhere:
I just feel like this is good practice.

### DIFF
--- a/mp_compile.py
+++ b/mp_compile.py
@@ -77,4 +77,5 @@ def install():
               "%s processes" % MAX_PROCS)
 
 
-install()
+if __name__ == "__main__":
+    install()


hugovk:
I think the whole point of mp_compile.py is to be run on import, and not from the command-line.

It's used here:

https://github.com/python-pillow/Pillow/blob/e92e730597204adbc104ad165d40a37d6187febd/setup.py#L24-L26

mp_compile.py
```diff
@@ -77,4 +77,5 @@ def install():
               "%s processes" % MAX_PROCS)
 
 
-install()
+if __name__ == "__main__":
+    install()
```
wiredfool: The whole point of this module is that it monkeypatches the installer when it's imported. It can't run as a separate script, so the main clause isn't appropriate.

radarhere:
Okay, thanks.

PR #2961, opened 2018-01-17
Changed assert statements to unittest calls

radarhere:


Tests/helper.py
```diff
@@ -210,7 +210,7 @@ def skipKnownBadTest(self, msg=None, platform=None,
             self.skipTest(msg or "Known Bad Test")
 
     def tempfile(self, template):
-        assert template[:5] in ("temp.", "temp_")
+        self.assertIn(template[:5], ("temp.", "temp_"))
```
wiredfool: I'm not sure about this one, it's more of an immediate test coding error, not something that should cause a test suite failure.

wiredfool: Though, I'm struggling to determine why that's actually the case. It's not like we actually care about it anymore.

radarhere: Okay, I've removed it from the commit.

PR #2978, opened 2018-01-25
Rename appveyor.yml as .appveyor.yml

hugovk:
Let's group it with the other dotfiles for consistency and less clutter.

> AppVeyor supports dot-file-style YAML named `.appveyor.yml` as is. 

https://www.appveyor.com/docs/build-configuration/#yaml-file-alternative-naming


radarhere:
You will also want to change https://github.com/python-pillow/Pillow/blob/master/MANIFEST.in#L27

hugovk:
Thanks, updated.

PR #2979, opened 2018-01-25
Remove redundant travis_after_all stuff

hugovk:
The stuff done in https://github.com/python-pillow/Pillow/pull/1067 to trigger a python-pillow/pillow-wheels build after a successful python-pillow/Pillow (non-PR, master) build hasn't been working for quite some time.

Let's just remove it.

* [x] I will also delete the old, unused branch named "[latest](https://github.com/python-pillow/pillow-wheels/tree/latest)" from pillow-wheels.

Instead, I propose we enable a [cron job](https://docs.travis-ci.com/user/cron-jobs/) on pillow-wheels.

How about daily, but not if there's been a build in the last 24h?

image

image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,29 +61,3 @@ script:
 
 after_success:
   - .travis/after_success.sh
-
-after_failure:
-  - |
-      if [ "$TRAVIS_REPO_SLUG" = "python-pillow/Pillow" ] && [ "$TRAVIS_BRANCH" = "master" ] && [ "$TRAVIS_PULL_REQUEST" = "false" ]; then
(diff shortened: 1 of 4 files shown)


hugovk:
Cron job enabled as suggested.

image

https://travis-ci.org/python-pillow/pillow-wheels/settings

hugovk:
Switched the cron from daily to weekly. Daily is a bit much, and ~12 runs between releases should be enough to catch something.

wiredfool:
I'm not sure this is doing anything, because it appears that it's attempting to build the same commit each time.

hugovk:
That's what https://github.com/python-pillow/Pillow/pull/2979 was doing: restarting the latest commit on a branch.

PR #2980, opened 2018-01-25
WIP Allow new-style buffer interfaces to be read/write

### LINKED ISSUE #2973: frombuffer constructed images are disconnected from the original data
### What did you do?

I wanted to use PIL to render into an externally allocated buffer without any memcpy, and thought that the `frombuffer` constructor would allow me to do that.

To test this I first created a np.array and then an image with the `Image.frombuffer()` and expected the reference to the data to be bidirectional (the documentation isn't clear on that). I.e. I expected that I can change a numpy array by creating an image based on it's data and then using paint operators. 

This did not work though. Is this a bug or a feature? Is there any particular reason why this isn't supported?

See below for a code example.

I expected the assert in the code below to succeed. I.e. that the `draw.rectangle()` call would change the data in the `data` variable.

### What actually happened?

The assert failed.

4.3.0

```python
from PIL import Image
from PIL import ImageDraw
import numpy as np

width,height=1,8
data = np.array([ord('A')]*(width*height),dtype=np.uint8)            #  An 1x8 chunk composed of 'A'

img = Image.frombuffer('L',(width,height),
                       data,'raw','L',0,1)
assert(img.getpixel((0,0)) == ord('A'))     # Assert that the data to
                                            # img conversion works
draw = ImageDraw.Draw(img)
draw.rectangle((0,0,width,height),ord('B')) # Draw B's in the entire image
assert(img.getpixel((0,0)) == ord('B'))     # Assert that the drawing worked
print 'data=',data
assert(data[0] == ord('B'))                 # Assert that the underlying
                                            # data was changed
```

wiredfool:
Fixes #2973.

 * Allow buffers to be read/write for at least the array interface, and possibly some mmapd files in the future. This opens up bidirectional communication with numpy arrays through shared memory. 
 
I believe that the reason that it wasn't read/write was probably historical and tied to the old buffer interface, as mmap and the new style buffer interface don't play nicely on 2.7. As I'm not sure of the reason that it's this way, I'm a little cautious of opening this up without reasonably serious checking. 

(also, tests, docs, etc)


vodp:
When this pull request will be merged? My graphics program is heavily affected by memcpy and this feature is the issue killer!

radarhere:
Looking at the code here, #8330 has added a slight variation on this.

PR #2892, opened 2017-12-16
Trim trailing whitespace from all source files throughout project

jdufresne:
Many editors trim whitespace by default. When newcomers are submitting patches, it is often undesirable to see inadvertent whitespace changes in the diff as it is distracting and unrelated. To make contributing easier, clean up all whitespace in one go across the project. Will allow for cleaner diffs in the future. Contributors will not need to worry about automatic yet inadvertent whitespace changes.

PR #2985, opened 2018-01-27
Trim trailing whitespace

Metallicow:
Evict foul whitespace demons.

### DIFF
--- a/Tests/helper.py
+++ b/Tests/helper.py
@@ -16,7 +16,7 @@ logger = logging.getLogger(__name__)
 HAS_UPLOADER = False
 
 if os.environ.get('SHOW_ERRORS', None):
-    # local img.show for errors. 
+    # local img.show for errors.
     HAS_UPLOADER=True
     class test_image_results:
         @classmethod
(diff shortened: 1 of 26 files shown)


hugovk:
This was going to be done in https://github.com/python-pillow/Pillow/pull/2892, but doesn't look like it happened, so merging this.

Metallicow:
@hugovk Im not sure about this,... but is there a way you can set up a repo to do this automatically or maybe even a bot(probably would cause more noise than helpful)?

I ask because usually when I start to tinker with code, I always seem to have to do a repo trimming first to reduce noise in my actual commits because my editor is set up to trim. If only everyone trimed automatically would solve this issue.. but alas not all editors are equal and some may just be making a one-off commit or are new to programming and not realize it.

hugovk:
It's probably possible with a pre-commit hook or something, but I doubt it's worth the effort for open-source projects.

If anyone happens to have [EditorConfig](http://editorconfig.org/) enabled, we have an [.editorconfig](https://github.com/python-pillow/Pillow/blob/master/.editorconfig) file with `trim_trailing_whitespace = true`.

PR #2998, opened 2018-02-06
Simplify version checks

hugovk:
* Named version attributes are available from Python 2.7+.

* Because 3.4 is our minimum supported 3.x, we can simplify these:
  * `sys.version_info >= (3, 4, 0)`
  * `sys.version_info >= (3, 3)`
* Into:
  * `sys.version_info.major >= 3`


radarhere:
There is also https://github.com/python-pillow/Pillow/blob/master/Tests/helper.py#L268

PR #2999, opened 2018-02-09
Corrected winbuild documentation

radarhere:
Changed `docs/build.rst` to `winbuild/build.rst`, to better point towards https://github.com/python-pillow/Pillow/blob/master/winbuild/build.rst

### DIFF
--- a/winbuild/README.md
+++ b/winbuild/README.md
@@ -1,7 +1,7 @@
 Quick README
 ------------
 
-For more extensive info, see the Windows build instructions `winbuild/build.rst`.
+For more extensive info, see the [Windows build instructions](build.rst).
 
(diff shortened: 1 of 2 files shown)


hugovk:
Rather than code formatting, how about adding a hyperlink to winbuild/build.rst?

hugovk:
Also, whilst we're at it, there's a broken link at https://github.com/python-pillow/Pillow/blob/master/winbuild/build.rst:

> For most people, the :doc:`installation instructions <installation>` should be sufficient

radarhere:
Thanks. Done and done.

hugovk:
Thanks. Relative links would be nice, but I don't think it matters too much.

PR #3001, opened 2018-02-11
Removed command option from Image.Image.show()

pythondude325:
The command option for Image.show() was still in the docs, fixed it.

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -1954,7 +1954,7 @@ class Image(object):
         if frame != 0:
             raise EOFError
 
-    def show(self, title=None, command=None):
+    def show(self, title=None):
         """
         Displays this image. This method is mainly intended for
         debugging purposes.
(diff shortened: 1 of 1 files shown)


radarhere:
I'm confused by your comment that the 'command option for Image.show() was still in the docs, fixed it' - this PR only modifies code, not documentation.

pythondude325:
From what I understand, the documentation is autogenerated from the code, Is it not?

radarhere:
You know what, I was wrong in my previous comment. The documentation is autogenerated from the code - the docstrings are pulled and used to populate http://pillow.readthedocs.io/

Generally speaking, the Pillow team do not like to make breaking API changes. You might say that no one would be using this argument, since it has no effect. While that is true for most cases, it is possible that someone `register`s their own viewer, and uses the `command` argument in their code.

pythondude325:
OK, thanks for the information.

PR #3009, opened 2018-02-20
Enable pip cache in appveyor build

thijstriemstra:
Enable pip cache in the appveyor build so packages are cached.


thijstriemstra:
can anyone let me know if this trivial change is ok?

hugovk:
Thanks.

For future reference, here's a link to the docs: https://www.appveyor.com/docs/build-cache/

PR #3023, opened 2018-03-01
Certain corrupted jpegs can result in no data read

### LINKED ISSUE #3022: Corrupt jpeg opens as all black image
### What did you do?
```python
from PIL import Image
from PIL import ImageFile

im = Image.open('pic263134.jpg')
im.save('1.png')
```
Image could be saved up to corruption point. 
### What actually happened?
All black image is saved. Tried saving to different output formats - it's all black anyway.

Python2. Pillow 5.0.0 and 3.0.0

Original image:
original_image

`djpeg` can output image correctly (up to corruption point)

Looks like it's problem with decoding step, since after opening, all image bands contain 0-only bytes:
```python
sum(im.split()[0].point(lambda x: 0 if x else 1).getdata())
# returns 3097440
```

kkopachev:
On truncated jpeg, decoder can suspend waiting for additional bytes in
buffer. For some input files, decoder suspends on jpeg_start_decompress
stage. If at this point file reader reaches EOF, py code never gets back
to jpeg decoder and we end up with no bytes to result image. This leaves
us with some amount of potentially useful bytes undecoded and thrown
away.
Libjpeg docs suggest that in such situation, more appropriate would be
to add EOI marker to the end of buffer, which will allows decoder
to finish. https://github.com/libjpeg-turbo/libjpeg-turbo/blob/0dd9a2c1fd6c/libjpeg.txt#L1803-L1809
Docs also mention that adding EOI markers is what non-suspending code
does anyway.

Fixes #3022 .

?w=1

 * On unexpected EOF of JPEG, when LOAD_TRUNCATED_IMAGES is set, add EOI marker to the buffer to let decoder finish processing. 

~~Proposed changes does not seem to be generic enough to make it through, so I am seeking advice of there this could be fixed better.~~


kkopachev:
Added few tests with truncated jpeg. Ready for review. Image is randomly generated. and few bytes cut from its tail.

kkopachev:
ping?

PR #3025, opened 2018-03-02
Improved readability of Other Platforms table

radarhere:
Splits the Ubuntu Linux 12.04 LTS section of the Other Platforms table into multiple rows, except for the first column, to improve readability.

Before

before

After

after


hugovk:
Maybe also the multiline parts of the preceding Continuous Integration Targets table?

radarhere:
Yep, good call thanks. Done.

Issue #2940, opened 2018-01-07
DNS changes for python-pillow.org


aclark4life:
I've just **moved python-pillow.org from Cloudflare to AWS Route 53** which presents at least one problem, as far as I can tell:

- Cloudflare has a nice CNAME flattening feature and I'm not sure what the equivalent is in AWS Route 53.

In the meantime, I've created a CNAME for www.python-pillow.org and am wondering if anyone has any suggests about how to handle this situation e.g. redirect to www will probably suffice … I think I can do that with S3 and if I don't hear any better ideas, that's probably what I'll do.

Oh, and SSL probably needs some thought/discussion too. (I want to support it.)

wiredfool:
A redirect is probably simplest, since github is hosting the project website.

aclark4life:
@python-pillow/pillow-team Never mind, for now I'm just going to redirect www & python-pillow.org to pillow.readthedocs.io. "One less thing." (I don't want a public facing "marketing" website anymore, unless it's awesome and awesome will take some effort & time.)

PR #3045, opened 2018-03-16
Updated URL

radarhere:
After https://github.com/python-pillow/Pillow/issues/2940#issuecomment-358326331, python-pillow.org redirects to http://pillow.readthedocs.io/en/latest/

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -763,7 +763,7 @@ try:
           long_description=_read('README.rst').decode('utf-8'),
           author='Alex Clark (Fork Author)',
           author_email='aclark@aclark.net',
-          url='https://python-pillow.org',
+          url='http://python-pillow.org',
           classifiers=[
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -763,7 +763,7 @@ def debug_build():
           long_description=_read('README.rst').decode('utf-8'),
           author='Alex Clark (Fork Author)',
           author_email='aclark@aclark.net',
-          url='https://python-pillow.org',
+          url='http://pillow.readthedocs.io/en/latest/',
```
jdufresne: This URL could be https://

aclark4life:
@radarhere Actually let's make it ``http://python-pillow.org`` for now (which redirects to RTD and is probably future-proof-ish.) Thanks

aclark4life:
Thanks @radarhere and @jdufresne

PR #3055, opened 2018-03-25
Fix incorrect image type checking in _imagingmorph module

homm:
Proof: https://github.com/python-pillow/Pillow/blob/master/src/PIL/ImageMorph.py#L202


radarhere:
Standard question - tests?

homm:
Without this fix, tests lead to segmentation error.

homm:
@hugovk technically, yes. Actually, most RuntimeErrors should be changed to ValueErrors across different modules. So there are two ways:

1. Revert it
2. Change all to ValueErrors and add release notes

I choose the first right now )

PR #3056, opened 2018-03-26
3D Color lookup tables

homm:
Support for https://en.wikipedia.org/wiki/3D_lookup_table transformations.

This implementation works at speed ≈34Mpx/s on my CPU.

The SSE4 implementation works at about 100Mpx/s, AVX2 mostly at 150Mpx/s, but there are some inexplicable slowdowns in some cases.


Tests/test_color_lut.py
```diff
@@ -0,0 +1,129 @@
+from __future__ import division
+from helper import unittest, PillowTestCase
+
+import time
```
radarhere: Unused import. Although you may tell me to save these comments for when this is less of a work-in-progress.

src/libImaging/ColorLUT.c
```diff
@@ -0,0 +1,161 @@
...
+#include <math.h>
+
+
+/* 8 bits for result. Table can overflow [0, 1.0] range,
+   so we need extra bits for overflow and negative values.
+   NOTE: This value should be the same as in _imaging/_prepare_lut_table() */
+#define PRECISION_BITS (16 - 8 - 2)
+#define PRECISION_ROUNDING (1<<(PRECISION_BITS-1))
+
+/* 8 — scales are multiplyed on byte.
```
radarhere: multiplied

src/libImaging/ColorLUT.c
```diff
@@ -0,0 +1,161 @@
...
+    /* This float to int conversion doesn't have rounding
+       error compensation (+0.5) for two reasons:
+       1. As we don't hit the highest value,
+          we can use one extra bit for precision.
+       2. For every pixel, we interpolate 8 elements from the table:
+          current and +1 for every dimension and they combinations.
+          If we hit the upper cells from the table,
+          +1 cells will be outside of the table.
+          With this compensation we never hit the upper cells
+          but this also doesn't introduce any noticable difference. */
```
radarhere: noticeable

src/libImaging/ColorLUT.c
```diff
@@ -0,0 +1,161 @@
...
+ImagingColorLUT3D_linear(Imaging imOut, Imaging imIn, int table_channels,
+                         int size1D, int size2D, int size3D,
+                         INT16* table)
+{
+    /* This float to int conversion doesn't have rounding
+       error compensation (+0.5) for two reasons:
+       1. As we don't hit the highest value,
+          we can use one extra bit for precision.
+       2. For every pixel, we interpolate 8 elements from the table:
+          current and +1 for every dimension and they combinations.
```
radarhere: Guessing that 'they' should be 'their'?

homm:
I think this mostly done. The only thing I'd like to add is ".cube" file format parser for "ImageFilter.Color3DLUT" class, but this is a relatively small feature.

src/PIL/ImageFilter.py
```diff
@@ -297,3 +299,150 @@ class SMOOTH_MORE(BuiltinFilter):
...
+                    size[0], size[1], size[2], len(table)))
+        self.table = table
+
+    @staticmethod
+    def _check_size(size):
+        try:
+            _, _, _ = size
+        except ValueError:
+            raise ValueError("Size should be an integer either "
+                             "tuple of three integers.")
```
radarhere: I presume you meant 'Size should be either an integer or a tuple of three integers'.

Although I'm not convinced they need to be integers, since you convert the values to integers in the next line after this block?

homm: With duck typing, "should be an integer" means that it should be something that can be used as an integer. So I'm converting values to check that they are really integers or can be converted to int.

homm:
Oh, I've found ".cube" files with newlines in the middle of the header, so parser should be fixed.

homm:
Done

homm:
The more I'm thinking about cubes, the more I want to move some functionality to the separate library. I think `Color3DLUT.generate` is ok in Pillow but `Color3DLUT.from_cube_file` should be removed.

homm:
@wiredfool Do you like class name and function names and the API itself?

homm:
@wiredfool I understand that this is a rather long patch and you may not have a time for reviewing. So it's Ok if you check only the most significant part, the API in `ImageFilter.py`. It is "only" 97 lines and most of them are docstrings.

homm:
I'm currently working on the [Pillow LUT tools](https://github.com/homm/pillow-lut-tools) which strongly depend on API from this PR. That is why I need API stabilization to move on.

wiredfool:
I’ll try to look at it this weekend.

wiredfool:
Looks reasonable to me. I'm a little unclear on the interaction between 3 channel vs 4 channel vs 3d table. Does it mean that the input is always a 3-tuple, and the output can be 3 or 4?  (e.g., RGB->RGB, or RGB-> CMYK), or can we go 4 Channel -> 3 (e.g. CMYK -> RGB)?

homm:
> Does it mean that the input is always a 3-tuple, and the output can be 3 or 4?

Yes

> or can we go 4 Channel -> 3

Not yet. 4D transformations are required for this.

homm:
Also, we can do 4 channels → 4 channels, where only 3 channels transformed from the source, and fourth is taken from LUT if it is 4 channel itself, or from the source image. This allows RGBA → RGBA transformations transparently without splitting off alpha channel.

wiredfool:
Is the 4d transform a long term future potential enhancement, or a quick knock it out thing when everything settles, or just not a thing that you're looking at?

wiredfool:
Ok, lgtm. I’d say it’s safe to build on this.

homm:
4d transformation is totally achievable and very interesting thing, but I suppose it will be about 2.5 times slower and cosume more RAM. It makes possibly not only CMYK → RGB transformations, but also RGBX transformations where X is some layer with additinal parameter. It could be, for example, spotlights in some places, where colors should be shifted more then in other places. This makes possible to implement such things as Instagram filters in one pass.

However, my next goal is an improvement for Python → C layer and optional numpy support as a table source. This should significantly reduce overhead and reduce memory consumption.

PR #3057, opened 2018-03-26
Enabling background colour parameter on rotate

storesource:
Enabling the user to choose the background colour of the final rotated image rather than just black. 
parameter added: backgroundcolor reusing transform()'s fillcolor parameter

If image was rotated by x degrees other than 90,180,270,
the background is black with no option to customise will cause problems when the rotated image's black background can be a problem for further use of the image eg. training data for ML.

Adding a backgroundcolor feature in rotate() will relieve the user of the extra effort of dealing with an unwanted colour

 * parameter added: backgroundcolor in rotate()


src/PIL/Image.py
```diff
@@ -1851,7 +1852,7 @@ def transform(x, y, matrix):
                                              matrix)
             w, h = nw, nh
 
-        return self.transform((w, h), AFFINE, matrix, resample)
+        return self.transform((w, h), AFFINE, matrix, resample, fillcolor=backgroundcolor)
```
hugovk: Let's keep the name fillcolor for the new parameter as well.

src/PIL/Image.py
```diff
@@ -1771,6 +1771,7 @@ def rotate(self, angle, resample=NEAREST, expand=0, center=None,
         :param center: Optional center of rotation (a 2-tuple).  Origin is
            the upper left corner.  Default is the center of the image.
         :param translate: An optional post-rotate translation (a 2-tuple).
+        :param backgroundcolor: An optional color for the color outside the transformed image
```
hugovk: End with a full stop.

hugovk: Perhaps "rotated" instead of "transformed"? We don't need to care how the transform happened.

Tests/test_image_rotate.py
```diff
@@ -96,6 +96,11 @@ def test_center(self):
         self.rotate(im, im.mode, 45, center=(0, 0))
         self.rotate(im, im.mode, 45, translate=(im.size[0]/2, 0))
         self.rotate(im, im.mode, 45, center=(0, 0), translate=(im.size[0]/2, 0))
+        
+    def test_rotate_with_fill(self):
+        im = hopper()
+        target = Image.open('Tests/images/hopper45withfill.png')
+        self.assert_image_equal(im.rotate(45, fillcolor='white'), target)
```
hugovk: Hmm, tests are failing due to "got different content", I guess from  minor differences from the Hopper photo.

Is it any better with with plain colour fields, something like this?

```python
    def test_rotate_no_fill(self):
        # Arrange
        im = Image.new('RGB', (100, 100), 'green')
        target = Image.open('Tests/images/rotate_45_no_fill.png')

        # Act
        im = im.rotate(45)

        # Assert
        self.assert_image_equal(im, target)

    def test_rotate_with_fill(self):
        # Arrange
        im = Image.new('RGB', (100, 100), 'green')
        target = Image.open('Tests/images/rotate_45_with_fill.png')

        # Act
        im = im.rotate(45, fillcolor='white')

        # Assert
        self.assert_image_equal(im, target)
```

storesource: I had not added anything for no fillcolor because I guessed that scenario is already handled in the other tests.
But I will do so if required.

hugovk:
And it's green! Please could you remove the unused test images?

storesource:
Thanks for your help @hugovk . You're very patient!
Needed some suggestions for below so that my other check-ins will be more professional:

* The IDE you use
* IDE plugin for code coverage so I can add more tests that will be required in future and now
* from . import _imaging as core
E   ImportError: cannot import name '_imaging' . 
What's the best way to deal with this issue with Image.py?
I did read the comment near the import statement I made changes accordingly but I was curious as to why the tests don't run directly on PyCharm

hugovk:
You're welcome, and thanks for the contribution!

* I've started using PyCharm (free for [open](https://www.jetbrains.com/pycharm/download/) [source](https://www.jetbrains.com/buy/opensource/)), otherwise I use BBEdit on Mac and Notepad++ on Windows.
* I've not used any plugins for code coverage in any IDEs. I run tests from the command line, see https://github.com/python-pillow/Pillow/tree/master/Tests#execution for full instructions. Normally I just run the single test file I'm adding to and let the CI check all the rest with coverage, and you can run `coverage` on a single file if you want. I've not run tests from the IDE, just debugging some simple test programs, but it could be useful to figure out how to do that.
* Something like `python setup.py install`, see https://pillow.readthedocs.io/en/latest/installation.html#building-from-source

PR #3062, opened 2018-03-31
.gitignore review

homm:
There are two points:

Ignore rules shouldn't be too broad to not affect unrelated files. This is very confusing when you are creating a file and don't see it in the git. As I know, git doesn't provide information why a particular file is ignored and it is a bit hard to find the exact rule. 

Compiled files is a frequent source of errors. For example, If you switch to 4.3.x branch, compile Pillow and return to 5.0.x, there will be PIL folder with `.so` and `.pyc` files.  It will be importable from python but will be not visible to git because `.so` and `.pyc` files are ignored everywhere. With this PR you can see compiled files in unexpected places.


.gitignore
```diff
@@ -1,25 +1,26 @@
...
 # C extensions
-*.so
+/src/PIL/*.so
 
 # Distribution / packaging
+/build/
+/dist/
+/src/*.egg-info/
+/env/
 .eggs/
```
homm: To be honest, I'd rather remove all rules stating from this until the end of the section. I haven't ever seen folders with this names. If someone believes that something should be left, please leave comments on the corresponding line.

.gitignore
```diff
@@ -1,25 +1,26 @@
 # Byte-compiled / optimized / DLL files
 __pycache__/
-*.py[cod]
+/*.py[cod]
+/src/PIL/*.py[cod]
+/Tests/*.py[cod]
```
radarhere: Other possible directories to include here are 'winbuild' and 'docs'. I feel like there's a good argument to be made not to include winbuild, since it is only used by AppVeyor, but I would suggest including 'docs', since a user may wish to use the documentation on their local machine, yes?

homm:
@wiredfool The example isn't included in the documentation anywhere, it just in the repo without any links to it. I'll revert this, just want to note.

hugovk:
With this .gitignore I get:
```
➤ Untracked files
#
#      untracked:  [5] PIL
#      untracked:  [6] Pillow.egg-info
#      untracked:  [7] Tests/__pycache__
#      untracked:  [8] __pycache__
#      untracked: [10] build
#      untracked: [11] dist
#      untracked: [12] src/PIL/__pycache__
#      untracked: [13] src/PIL/_imaging.cpython-36m-darwin.so
#      untracked: [14] src/PIL/_imaging.so
#      untracked: [15] src/PIL/_imagingcms.cpython-36m-darwin.so
#      untracked: [16] src/PIL/_imagingcms.so
#      untracked: [17] src/PIL/_imagingft.cpython-36m-darwin.so
#      untracked: [18] src/PIL/_imagingft.so
#      untracked: [19] src/PIL/_imagingmath.cpython-36m-darwin.so
#      untracked: [20] src/PIL/_imagingmath.so
#      untracked: [21] src/PIL/_imagingmorph.cpython-36m-darwin.so
#      untracked: [22] src/PIL/_imagingmorph.so
#      untracked: [23] src/PIL/_imagingtk.cpython-36m-darwin.so
#      untracked: [24] src/PIL/_imagingtk.so
#      untracked: [25] src/PIL/_webp.cpython-36m-darwin.so
#      untracked: [26] src/PIL/_webp.so
#      untracked: [27] src/Pillow.egg-info
```

Do we want all those to be visible/potentially committed?

aclark4life:
FWIW I don't care that much for ``.gitignore`` in general. Mine are usually empty because I like to "see what's going on". On rare occasions, I've added things to ``.gitignore`` and have certainly worked on a lot of projects where other developers seem to care about ``.gitignore``.

File under bikeshed, I guess.

homm:
@hugovk

> Do we want all those to be visible/potentially committed?

Not potentially committed, but visible.

> [5] PIL

This is what I talking about and why sources was moved to ./src subdir ) If you have compiled files in this folder, they will be used in from PIL import Image. Now you can see this folder and can erese it.

> [6] Pillow.egg-info

This is artifact from pre-5.0 versions of Pillow and now you see than you can saifly remove it.

As of other files, they are all should be ignored and I don't know why you are seeing them. Which version of git have you installed? My is: 

```
$ git --version
git version 1.9.1
```

and I don't see these files.

depends/install_extra_test_images.sh
```diff
@@ -7,3 +7,5 @@ rm -r test_images
 svn checkout https://github.com/python-pillow/pillow-depends/trunk/test_images
 
 cp -r test_images/* ../Tests/images
+
+rm -r test_images
```
radarhere: After I resolved a conflict, this is now failing on Xenial builds, with
> rm: remove write-protected regular file 'test_images/.svn/pristine/5d/5d6d293b404eb2b4ade0869394040ae1435eead2.svn-base'?

radarhere: This is no longer necessary, thanks to #4175

.gitignore
```diff
@@ -1,42 +1,51 @@
...
+/docs/*.py[cod]
+/winbuild/*.py[cod]
 
 # C extensions
-*.so
+/src/PIL/*.so
```
nulano, line 10: If `.so` is included, perhaps also include `.dll`? Although they are hidden by my global `.gitignore`, this might not be the case for others.

.gitignore
nulano: `build` is used by setuptools as an intermediate build directory (often not cleaned properly), and `dist` is where it puts wheels/eggs.

`env` is probably for virtual environments, and likely unnecessary (one can just put a `*` gitignore in the subdirectory if they need it).

I would guess the other directories are for older versions of setuptools/distutils.

hugovk:
Let's close this, thanks anyway for the suggestion!

PR #3064, opened 2018-04-02
Release notes for 5.1.0

wiredfool:


docs/handbook/image-file-formats.rst
```diff
@@ -767,6 +767,13 @@ PIL reads and writes X bitmap files (mode ``1``).
 Read-only formats
 -----------------
 
+BLP
+^^^
+
+BLP is the Blizzard Mipmap Format is a texture format used in World of
```
radarhere: This needs rewording. 'is the'... 'is a'

docs/releasenotes/5.1.0.rst
```diff
@@ -0,0 +1,36 @@
+5.1.0
+-----
+
+New File Format
+================
```
radarhere: The number of equals signs does not match the number of characters in the heading.

PR #3072, opened 2018-04-04
Remove unnecessary `#if 1` directive

jdufresne:
The preprocessor directive `#if 1` always evaluates as true, so the enclosed code is always compiled.

The directive has existed since the original fork from PIL, 9a640e3157150e35f17354517e8c6304fc428060.


hugovk:
And it's been there since at lease PIL 1.1.1, from October 2000.

https://github.com/hugovk/PIL/blame/master/libImaging/Draw.c#L928

I guess it was originally added because it was "experimental graphics stuff", to make it easy to compile out. Twenty years is long enough!

PR #3076, opened 2018-04-05
 Support Python 3.7

hugovk:
Python 3.7.0 is due out on 2018-06-15, before the next quarterly Pillow release.

https://www.python.org/dev/peps/pep-0537/#schedule

`3.7-dev` is the latest available on Travis CI, let's replace it with `3.7` when that's available.

### DIFF
--- a/docs/releasenotes/5.2.0.rst
+++ b/docs/releasenotes/5.2.0.rst
@@ -1,0 +1,21 @@
+5.0.0
+-----
+
+API Additions
+=============
+
+Image.rotate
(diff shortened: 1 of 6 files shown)


aclark4life:
LGTM

docs/releasenotes/5.2.0.rst
```diff
@@ -0,0 +1,21 @@
+5.0.0
```
radarhere: 5.2.0

PR #3084, opened 2018-04-09
Health fixes

radarhere:


src/PIL/IcnsImagePlugin.py
```diff
@@ -310,7 +310,7 @@ def _save(im, fp, filename):
 
     # create the temporary set of pngs
     iconset = tempfile.mkdtemp('.iconset')
-    provided_images = {im.width:im for im in
+    provided_images = {im.width: im for im in
                        im.encoderinfo.get("append_images", [])}
```
homm: I more prefer a style when the loop and body on different lines:

```python
provided_images = {
    im.width: im
    for im in im.encoderinfo.get("append_images", [])
}

radarhere: Okay, I've added a commit.

PR #3085, opened 2018-04-09
Use __version__ internally, drop old 1.1.7 number, and deprecate other constants

hugovk:
Fixes #3082.

This can be as in addition to or instead of #3083.

Changes proposed in this pull request:

 * In the codebase, use the more conventional `__version__` rather than `PILLOW_VERSION` or `VERSION`
 * Ditch the old PIL version number from `VERSION` and put the Pillow version in there instead. It's been nearly a decade since the fork, and no end users will need to know the old PIL 1.1.7. It just causes confusion.
 * Further, mark `PILLOW_VERSION` and `VERSION` as deprecated and we can drop them in the next major release.

PR #3083, opened 2018-04-09
Rename PIL.version to PIL._version and remove it from module

homm:
Partially fixes #3082.

It is really confusing for people to find three different constants in PIL module: `VERSION`, `PILLOW_VERSION`, `__version__` and `version`. While `VERSION` is here for historical reasons and does matter, `version` is module is not required to end user at all. My suggestion is to hide it from users by removing from namespace:

```python
In [1]: from PIL import _version
ImportError: cannot import name _version  # Python 2 only

In [2]: import PIL

In [3]: PIL._version
AttributeError: 'module' object has no attribute '_version'
```

Issue #3082, opened 2018-04-08
Confusion getting PIL / Pillow version string


gnbl:
I was trying to figure out which version of pillow (PIL) is installed, and stumbled on this:

```cmd
Python 3.6.5 (v3.6.5:f59c0932b4, Mar 28 2018, 17:00:18) [MSC v.1900 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import PIL
>>> PIL.version
<module 'PIL.version' from 'C:\\Users\\me\\AppData\\Local\\Programs\\Python\\Python36\\lib\\site-packages\\PIL\\version.py'>
>>> help(PIL.version)
Help on module PIL.version in PIL:

NAME
    PIL.version - # Master version for Pillow

VERSION
    5.1.0

FILE
    c:\users\me\appdata\local\programs\python\python36\lib\site-packages\pil\version.py


>>> help(PIL)
Help on package PIL:

NAME
    PIL

DESCRIPTION
    # The Python Imaging Library.
    # $Id$
    #
    # package placeholder
    #
    # Copyright (c) 1999 by Secret Labs AB.
    #
    # See the README file for information on usage and redistribution.
    #

PACKAGE CONTENTS
    BdfFontFile
    BlpImagePlugin
    ...

>>> PIL.VERSION
'1.1.7'
>>> PIL.version.__version__
'5.1.0'
>>> PIL.PILLOW_VERSION
'5.1.0'
>>> PIL.__version__
'5.1.0'
>>>
```

- it's confusing that PIL.VERSION returns the version string of the former PIL instead of Pillow's
- there does not seem to be documentation on this version number (why this, will it ever change, ..) e.g.  at https://pillow.readthedocs.io/en/5.1.x/about.html#why-a-fork
- it's confusing that PIL.version is a module and does not return the version information directly or hints on how to get it
- the package information header is essentially useless (placeholder, does not even mention Pillow, nor the version)
- PIL.version module documentation comment could explain how to access the version information

Others also have trouble finding it: https://stackoverflow.com/questions/44739775/how-to-get-python-pil-version

Source files:
- https://github.com/python-pillow/Pillow/blob/master/src/PIL/version.py
- https://github.com/python-pillow/Pillow/blob/master/src/PIL/__init__.py#L2

aclark4life:
[...]

PR #2027, opened 2016-07-15
Expose Pillow package version as PIL.__version__

techtonik:
Fixes this:
```
>>> import PIL
>>> from PIL import ImageGrab
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/dist-packages/PIL/ImageGrab.py", line 22, in <module>
    raise ImportError("ImageGrab is OS X and Windows only")
ImportError: ImageGrab is OS X and Windows only
>>> PIL.version
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'module' object has no attribute 'version'
>>> PIL.__version__
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'module' object has no attribute '__version__'
>>> PIL.get_version()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'module' object has no attribute 'get_version'
>>> 
```

PR #3090, opened 2018-04-11
 Deprecate PILLOW_VERSION and VERSION

hugovk:
For #3082.

Replaces PR https://github.com/python-pillow/Pillow/pull/3085.

 * Internally, use more conventional `__version__` rather than `PILLOW_VERSION`
 * Mark `PILLOW_VERSION` and `VERSION` as deprecated, we'll remove them in the next major version. 
   * It's been nearly a decade since the fork, and no end users will need to know the old PIL 1.1.7. It just causes confusion.
   * With that gone, no need for `PILLOW_VERSION`, use `__version__` instead.

---

https://github.com/python-pillow/Pillow/pull/3083#issuecomment-380353788:

> Don't break the single version in c & python,

Check

> don't break the world by removing the PIL 1.1.7 version until we do a major release

Check


src/PIL/__init__.py
```diff
@@ -13,6 +13,7 @@
 
 from . import version
 
+# PILLOW_VERSION and VERSION are deprecated and will be removed in Pillow 6.0.0. Use __version__ instead.
 VERSION = '1.1.7'  # PIL Version
 PILLOW_VERSION = version.__version__
```
homm: Could you rebase on the current master to see the full picture?

hugovk: Rebased!

src/PIL/ImageCms.py
```diff
@@ -951,5 +951,5 @@ def versions():
 
     return (
         VERSION, core.littlecms_version,
-        sys.version.split()[0], Image.VERSION
+        sys.version.split()[0], Image.__version__
```
homm: It used to be 1.1.7 here.

hugovk: Yes, reverted.

docs/releasenotes/5.2.0.rst
```diff
@@ -0,0 +1,22 @@
+5.2.0
+-----
+
+API Changes
+===========
+
+Deprecations
+^^^^^^^^^^^^
+
+Two version constants – ``VERSION`` (the old PIL version 1.1.7) and
```
homm: ``PIL.VERSION`` and so on

hugovk: Added.

wiredfool:
You know, there's one other thing. 

When we ask for a version number, we should be able to ask for one thing. And unfortunately, that's going to be PILLOW_VERSION for a good long time. 

Otherwise, we're going to get this conversation:

>>> What version are you running?

>> How do I do that?

> Depends on what version you're running, either PIL.PILLOW_VERSION or PIL.\_\_version\_\_

aclark4life:
That conversation is probably OK because that’s what deprecation is for. PILLOW_VERSION was a hack we should probably fix…

hugovk:
PIL.__version__ has already been around for 2 years since 3.4.0, in 7 quarterly releases so far. 

(Commit https://github.com/python-pillow/Pillow/commit/7bfb2527bb0b832f830f87c37bd165028118e442 in PR https://github.com/python-pillow/Pillow/pull/2027.) 

So I think we're fine asking for PIL.__version__.

docs/releasenotes/5.2.0.rst
```diff
@@ -0,0 +1,28 @@
...
+
+These version constants have been deprecated and will be removed in the next
+major release:
+
+* ``PIL.VERSION`` (old PIL version 1.1.7)
+* ``PIL.PILLOW_VERSION``
+* ``PIL.Image.VERSION``
+* ``PIL.Image.PILLOW_VERSION``
+
+Use ``__version__`` instead.
```
homm: PIL.__version__ ?

hugovk: Good idea, updated.

homm:
You know, I'm for the depreciating VERSION and leaving PILLOW_VERSION as is. I don't think this is so terrible to have fallback for code which works from the beging of the library. Anyway, we already reduced the number of "version" constants from four to two in next release.

homm:
@hugovk Needs rebase

aclark4life:
@hugovk Sounds good… I think saying we're going to get rid of ``PILLOW_VERSION`` at some point but not specifying _which_ future version will allow folks plenty of time to comment if they care. As the person who probably added ``PILLOW_VERSION``, I think it should eventually go…

hugovk:
Rebased and updated.

docs/releasenotes/5.2.0.rst
```diff
@@ -0,0 +1,28 @@
...
+-----
+
+API Changes
+===========
+
+Deprecations
+^^^^^^^^^^^^
+
+These version constants have been deprecated. ``VERSION`` will be removed in
+Pillow 6.0.0, and ``PILLOW_VERSION`` will be removed in a future release.
```
homm: Doesn't it mean that PILLOW_VERSION will be removed in a future release i.e. in 5.3, before 6.0? For me, "and ``PILLOW_VERSION`` will be removed after that." looks more clear.

hugovk: It is ambiguous. Updated!

jwilk:
I was using ``PILLOW_VERSION`` to [distinguish between the original PIL and Pillow](https://github.com/jwilk/scanhelper/blob/ecad238aae9e00eee4cc867cb28dc8006e310ec6/lib/cli.py#L92).
Is there another way to distinguish between the two?

radarhere:
PIL doesn't have `__version__`, so you can use that as a direct continuation of your current check

PR #3091, opened 2018-04-11
Transforming LUTs

homm:
This adds the new method `Color3DLUT.transform()` which allows altering table values with a callback.

This also improves the performance in three times of `Color3DLUT.generate()` in two ways: First, using future division instead of direct `float` function call. Second, avoid of extra copying of the table in the constructor.

Also, PR adds autogenerated docs for `Color3DLUT`.

### DIFF
--- a/docs/reference/ImageFilter.rst
+++ b/docs/reference/ImageFilter.rst
@@ -40,10 +40,10 @@ image enhancement filters:
 .. autoclass:: PIL.ImageFilter.Color3DLUT
 .. autoclass:: PIL.ImageFilter.BoxBlur
 .. autoclass:: PIL.ImageFilter.GaussianBlur
+.. autoclass:: PIL.ImageFilter.UnsharpMask
 .. autoclass:: PIL.ImageFilter.Kernel
+.. autoclass:: PIL.ImageFilter.RankFilter
 .. autoclass:: PIL.ImageFilter.MedianFilter
 .. autoclass:: PIL.ImageFilter.MinFilter
 .. autoclass:: PIL.ImageFilter.MaxFilter
 .. autoclass:: PIL.ImageFilter.ModeFilter
-.. autoclass:: PIL.ImageFilter.RankFilter
-.. autoclass:: PIL.ImageFilter.UnsharpMask
--- a/docs/reference/ImageFilter.rst
+++ b/docs/reference/ImageFilter.rst
@@ -37,12 +37,13 @@ image enhancement filters:
 * **SMOOTH**
 * **SMOOTH_MORE**
 
-.. autoclass:: PIL.ImageFilter.GaussianBlur
+.. autoclass:: PIL.ImageFilter.Color3DLUT
 .. autoclass:: PIL.ImageFilter.BoxBlur
-.. autoclass:: PIL.ImageFilter.UnsharpMask
+.. autoclass:: PIL.ImageFilter.GaussianBlur
 .. autoclass:: PIL.ImageFilter.Kernel
-.. autoclass:: PIL.ImageFilter.RankFilter
 .. autoclass:: PIL.ImageFilter.MedianFilter
 .. autoclass:: PIL.ImageFilter.MinFilter
 .. autoclass:: PIL.ImageFilter.MaxFilter
 .. autoclass:: PIL.ImageFilter.ModeFilter
+.. autoclass:: PIL.ImageFilter.RankFilter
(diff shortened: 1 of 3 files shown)


homm:
`transform()` and  `generate()` are companions. They work together to create or adjust LUTs using some functions. The good example of how they can be used together is the [color adjustment](https://github.com/homm/pillow-lut-tools/blob/213b98e55127d3a2ec1d2368af41730485ce3044/pillow_lut/generators.py#L302-L305).

src/PIL/ImageFilter.py
```diff
@@ -442,6 +442,15 @@ def transform(self, callback, with_normals=False, channels=None,
                           target_mode=target_mode or self.mode,
                           _copy_table=False)
 
+    def __repr__(self):
+        r = ["{0} from {1}".format(self.__class__.__name__,
+                                   self.table.__class__.__name__)]
+        r.append("size={0}x{1}x{2}".format(*self.size))
+        r.append("channels={0}".format(self.channels))
```
radarhere: Minor styling issue - would it not be clearer to include all three items in the initial declaration of the array, rather than using two `append` statements?

src/PIL/ImageFilter.py
```diff
@@ -442,6 +442,15 @@ def transform(self, callback, with_normals=False, channels=None,
                           target_mode=target_mode or self.mode,
                           _copy_table=False)
 
+    def __repr__(self):
+        r = ["{0} from {1}".format(self.__class__.__name__,
```
hugovk: In all these, we can have simple formatters without numbering: https://pyformat.info/#simple

src/PIL/ImageFilter.py
homm: Done

src/PIL/ImageFilter.py
homm: Done

docs/reference/ImageFilter.rst
```diff
@@ -37,12 +37,13 @@ image enhancement filters:
 * **SMOOTH**
 * **SMOOTH_MORE**
 
-.. autoclass:: PIL.ImageFilter.GaussianBlur
+.. autoclass:: PIL.ImageFilter.Color3DLUT
 .. autoclass:: PIL.ImageFilter.BoxBlur
-.. autoclass:: PIL.ImageFilter.UnsharpMask
+.. autoclass:: PIL.ImageFilter.GaussianBlur
```
homm: Oh, I tried to order classes alphabetically and was happy that BoxBlur and GaussianBlur come together because they are very similar. But now I see that actually, Color3DLUT should be between them ) So it's still not alphabetically.

Also, it better to return RankFilter and UnsharpMask on its place.

src/PIL/ImageFilter.py
```diff
@@ -371,20 +379,81 @@ def generate(cls, size, callback, channels=3, target_mode=None):
...
+                         function should take is ``self.channels`` or
+                         ``3 + self.channels`` if ``with_normals`` flag is set.
+                         Should return a tuple of ``self.channels`` or
+                         ``channels`` elements if it is set.
+        :param with_normals: If true, ``callback`` will be called with
+                             coordinates in the color cube as the first
+                             three arguments. Otherwise, ``callback``
+                             will be called only with actual color values.
+        :param channels: Number of channels in the resulted lookup table.
+        :param target_mode: Passed to the constructor of the resulted
```
hugovk: "resulted" -> "resulting" (3 places)

homm: Thanks!

homm:
I think it is ready for landing. @hugovk @radarhere Any notes?

src/PIL/ImageFilter.py
```diff
@@ -371,20 +379,81 @@ def generate(cls, size, callback, channels=3, target_mode=None):
...
+        return cls((size1D, size2D, size3D), table, channels=channels,
+                   target_mode=target_mode, _copy_table=False)
+
+    def transform(self, callback, with_normals=False, channels=None,
+                  target_mode=None):
+        """Transforms the table values using provided callback and returns
+        a new LUT with altered values.
+
+        :param callback: A function which takes old lookup table values
+                         and returns a new. The number of arguments which
```
radarhere: 'and returns a new'. And returns a new set of values?

homm: Fixed. Thanks!

PR #3098, opened 2018-04-14
Raise error if it is occurred during conversion in getlist

homm:
From [`PyInt_AsLong()`](https://docs.python.org/3/c-api/long.html#c.PyLong_AsLong) and [`PyFloat_AsDouble()`](https://docs.python.org/3/c-api/float.html#c.PyFloat_AsDouble) docs:

```
Returns -1 on error. Use PyErr_Occurred() to disambiguate.
```
```
This method returns -1.0 upon failure, so one should call PyErr_Occurred() to check for errors.
```

### DIFF
--- a/src/_imaging.c
+++ b/src/_imaging.c
@@ -379,12 +379,12 @@ getlist(PyObject* arg, Py_ssize_t* length, const char* wrong_length, int type)
     PyObject* seq;
     PyObject* op;
 
-    if (!PySequence_Check(arg)) {
+    if ( ! PySequence_Check(arg)) {
         PyErr_SetString(PyExc_TypeError, must_be_sequence);
         return NULL;
     }
 
-    n = PyObject_Length(arg);
+    n = PySequence_Size(arg);
     if (length && wrong_length && n != *length) {
         PyErr_SetString(PyExc_ValueError, wrong_length);
         return NULL;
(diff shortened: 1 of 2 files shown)


src/_imaging.c
```diff
@@ -393,13 +393,12 @@ getlist(PyObject* arg, Py_ssize_t* length, const char* wrong_length, int type)
...
     list = calloc(n, type & 0xff);
-    if (!list)
+    if ( ! list)
         return PyErr_NoMemory();
 
     seq = PySequence_Fast(arg, must_be_sequence);
-    if (!seq) {
+    if ( ! seq) {
         free(list);
-        PyErr_SetString(PyExc_TypeError, must_be_sequence);
```
homm: It's superfluous, [because](https://docs.python.org/3/c-api/sequence.html#c.PySequence_Fast):

```
If the object is not a sequence, raises TypeError with m as the message text.
```

src/_imaging.c
```diff
@@ -379,12 +379,12 @@ getlist(PyObject* arg, Py_ssize_t* length, const char* wrong_length, int type)
...
     PyObject* op;
 
-    if (!PySequence_Check(arg)) {
+    if ( ! PySequence_Check(arg)) {
         PyErr_SetString(PyExc_TypeError, must_be_sequence);
         return NULL;
     }
 
-    n = PyObject_Length(arg);
+    n = PySequence_Size(arg);
```
homm: Usually, `PyObject_Length` does more because it also could tell a size of dicts, for example, but we already checked that the object is Sequence and using `PySequence_` function here.

PR #3107, opened 2018-04-21
Corrected undefined behaviour

### LINKED ISSUE #3030: Undefined behavior in Effects.c
In `libImaging/Effects.c` the function `ImagingEffectNoise` has an undefined behavior in [line 109](https://github.com/python-pillow/Pillow/blob/master/src/libImaging/Effects.c#L109):

```c
out[x] = (unsigned char) (128 + sigma * this);
```

This is detected when running the test `test_image.py` and compiling the library with UndefinedBehaviorSanitizer. See https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html to see how to detect this.

This code is easy to fix with something like:

```c
int d = 128 + sigma * this;
out[x] = d<0? 0: (d > UCHAR_MAX: UCHAR_MAX: d);
```

radarhere:
Resolves #3030 

This uses the code change suggested in that issue, with a syntax correction.

### DIFF
--- a/src/libImaging/ImagingUtils.h
+++ b/src/libImaging/ImagingUtils.h
@@ -30,6 +30,8 @@
     (MULDIV255(in1, (255 - mask), tmp1) + in2)
 
 
+#define CLIP(v) ((v) <= 0 ? 0 : (v) < 256 ? (v) : 255)
+
 /* This is to work around a bug in GCC prior 4.9 in 64 bit mode.
    GCC generates code with partial dependency which is 3 times slower.
    See: http://stackoverflow.com/a/26588074/253146 */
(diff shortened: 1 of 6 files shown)


hugovk:
How about using the CLIP macro? And the four existing CLIP definitions could be moved to ImagingUtils.h.

radarhere:
Sure, thanks, done.

src/libImaging/ImagingUtils.h
```diff
@@ -30,6 +30,8 @@
     (MULDIV255(in1, (255 - mask), tmp1) + in2)
 
 
+#define CLIP(v) ((v) <= 0 ? 0 : (v) < 256 ? (v) : 255)
```
homm: Better name is `CLIP8`

PR #3108, opened 2018-04-21
Updated libwebp to 1.0.0

radarhere:


homm:
@radarhere Do you have any idea why coverage decreased?

radarhere:
Yes, it's webp wasn't installing correctly. I've fixed this by pushing the libwebp package from downloads.webmproject.org to pillow-depends, which includes `configure`, instead of the one from github.

PR #3126, opened 2018-05-09
Added seek 0 and Jpeg for multiframe MPO

### LINKED ISSUE #1631: No n_frames, or bad value for n_frames
When I feed the test file  flower2.jpg  into this code (from #1630)

```python
im = Image.open( fn )
imgcnt = im.n_frames
colors = im.getcolors( im.width * im.height )
if args.hist:
    for cnt, col in colors:
        allcolors[ col ] += cnt
    for iz in range( 1, imgcnt ):
        im = Image.open( fn ) # does getcolors implicitly close????
                              # without the open, get "seek of closed
                              # file" error on line below.
        im.seek( iz )
        colors = im.getcolors( im.width * im.height )
        for cnt, col in colors:
            allcolors[ col ] += cnt
```

I get "AttributeError: n_frames"

But other .jpg files do not get that error... this one: http://nevcal.com/temporary/20151110-105826gl.jpg has no problem with the attribute error on that line, but it gets a value of 2, apparently handles the seek OK, but dies in the second call to getcolors, with "OSError: image file is truncated (0 bytes not processed)".

jmetz:
Fixes #1631 issue when loading multiple frames of MPO.

 * Added seek(0) and JpegImagePlugin.JpegImageFile._open(self) after each seek

This works to reinitialise the fp.

### DIFF
--- a/src/PIL/MpoImagePlugin.py
+++ b/src/PIL/MpoImagePlugin.py
@@ -80,6 +80,8 @@ class MpoImageFile(JpegImagePlugin.JpegImageFile):
             ("jpeg", (0, 0) + self.size, self.offset, (self.mode, ""))
         ]
         self.__frame = frame
+        self.fp.seek(0)
+        JpegImagePlugin.JpegImageFile._open(self)
 
     def tell(self):
         return self.__frame


src/PIL/MpoImagePlugin.py
```diff
@@ -80,6 +80,8 @@ def seek(self, frame):
             ("jpeg", (0, 0) + self.size, self.offset, (self.mode, ""))
         ]
         self.__frame = frame
+        self.fp.seek(0)
+        JpegImagePlugin.JpegImageFile._open(self)
```
radarhere: Would `super(MpoImageFile, self)._open()` not be better?

jmetz: @radarhere - I was simply reusing the existing syntax used above (https://github.com/jmetz/Pillow/blob/ebb78aecb94227e00e259931df70581f7f5b56cf/src/PIL/MpoImagePlugin.py#L46 ) . 
Could switch to `super` version though (should I change both?) if that's preferred.

radarhere: Ah. Fair enough then! Well, we'll see how #3149 goes

radarhere: Wait, I've realised there is already discussion about this question in #2198. Ignore my comment, your code is fine.

hugovk:
I've not checked this PR in any detail, but at least the CIs need to pass. Right now `TestFileMpo::test_image_grab` is failing. 

Also new unit tests to verify the fix and protect against regressions.

radarhere:
Closing. #3588 replaces this, and is [confirmed to be a solution by the author of this PR](https://github.com/python-pillow/Pillow/issues/1631#issuecomment-455205959).

PR #3133, opened 2018-05-17
AppVeyor: upgrade to PyPy 6.0.0

hugovk:
It was released in April 2018:

https://morepypy.blogspot.com/2018/04/pypy27-and-pypy35-v60-dual-release.html

We can also re-enable some tests which were a problem with older versions of PyPy.

### DIFF
--- a/Tests/test_imagedraw.py
+++ b/Tests/test_imagedraw.py
@@ -360,8 +360,6 @@ class TestImageDraw(PillowTestCase):
         ImageDraw.floodfill(im, (W, H), red)
         self.assert_image_equal(im, im_floodfill)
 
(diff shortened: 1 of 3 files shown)


Tests/test_imagedraw.py
```diff
@@ -360,8 +360,6 @@ def test_floodfill(self):
         ImageDraw.floodfill(im, (W, H), red)
         self.assert_image_equal(im, im_floodfill)
 
-    @unittest.skipIf(hasattr(sys, 'pypy_version_info'),
-                     "Causes fatal RPython error on PyPy")
```
radarhere: The sys import in test_imagedraw is no longer necessary after this change

PR #3141, opened 2018-05-26
Fixed typo

radarhere:


hugovk:
Of course, coverage didn't really drop from this typo fix.

What happened: the last master build didn't complete on AppVeyor, which meant some files weren't included in the report at all, so coverage was higher for those (83.58%).

83.43% was the value for the last successful master build.

PR #3142, opened 2018-05-26
Changed ellipse point calculations to be more evenly distributed

### LINKED ISSUE #1913: ImageDraw Ellipse: Asymmetric Circle for all even diameters
### What did you do?

used ImageDraw Ellipse with two equal diameters to draw a circle.

round, symmetric circle
### What actually happened?

round and symmetric circles for all odd diameters but unsymmetric artifacts for all even diameters

Pillow 3.2.0, Python 3
bildschirmfoto 2016-05-16 um 13 56 04

``` python

img = Image.new('1', (8, 8), "white")
draw = ImageDraw.Draw(img)

draw.ellipse((1, 1, 6, 6), fill = 'black', outline ='black')

img.show()
```

radarhere:
Resolves #1913

The float values calculated by Pillow create a symmetric ellipse.

However, the issue is converting those float values to ints. Since PIL, `FLOOR` and `+ 0.5` have been used. This produces ellipses that may not be symmetric, since a 0.5 values are consistently rounded up, which has the effect of leaning the ellipse towards the bottom and right.

Instead, I suggest using `FLOOR` when the point is towards the bottom or right, and `CEIL` when it is towards the top or left. This should have the effect of leaning the ellipse towards it's center instead, producing a more symmetric shape.

### DIFF
--- a/Tests/test_imagedraw.py
+++ b/Tests/test_imagedraw.py
@@ -169,6 +169,16 @@ class TestImageDraw(PillowTestCase):
         self.assert_image_similar(
             im, Image.open("Tests/images/imagedraw_ellipse_edge.png"), 1)
 
+    def test_ellipse_symmetric(self):
+        for bbox in [
+            (25, 25, 76, 76),
+            (25, 25, 75, 75)
+        ]:
+            im = Image.new("RGB", (101, 101))
+            draw = ImageDraw.Draw(im)
+            draw.ellipse(bbox, fill="green", outline="blue")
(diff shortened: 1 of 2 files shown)


hugovk:
Please could you show some example before and after ellipses?

radarhere:
Sure. The following code generates the following images.

```

im = Image.new('1', (22, 230), "white")
draw = ImageDraw.Draw(im)

x = 1
y = 1
for i in range(2, 20):
	draw.ellipse((x, y, x+i, y+i))
	
	y += i+2
im.save('out.png')
```

Before
before
After
after

PR #3146, opened 2018-05-30
Allow float values in getrgb hsl color string

### LINKED ISSUE #2119: Feature Request: PIL.ImageColor.getrgb("hsl(hue,sat%,lig%.)") allow values to be decimal/float numbers
The values in doing PIL.ImageColor.getrgb("hsl(hue, sat%, lig%)") must currently be integer, but this limits the resulting RGB color resolution, since a difference of 1 as minimum step for arguments does sometimes results in a difference for RGB value of up to about 5, thus not allowing the function to control the value fine grained.

radarhere:
Resolves #2119

### DIFF
--- a/src/PIL/ImageColor.py
+++ b/src/PIL/ImageColor.py
@@ -87,7 +87,7 @@ def getrgb(color):
             int((int(m.group(3)) * 255) / 100.0 + 0.5)
             )
 
-    m = re.match(r"hsl\(\s*(\d+)\s*,\s*(\d+)%\s*,\s*(\d+)%\s*\)$", color)
+    m = re.match(r"hsl\(\s*(\d+\.?\d*)\s*,\s*(\d+\.?\d*)%\s*,\s*(\d+\.?\d*)%\s*\)$", color)
     if m:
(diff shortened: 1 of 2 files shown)


src/PIL/ImageColor.py
```diff
@@ -87,13 +87,13 @@ def getrgb(color):
             int((int(m.group(3)) * 255) / 100.0 + 0.5)
             )
 
-    m = re.match(r"hsl\(\s*(\d+)\s*,\s*(\d+)%\s*,\s*(\d+)%\s*\)$", color)
+    m = re.match(r"hsl\(\s*(\d+(\.\d*)?)\s*,\s*(\d+(\.\d*)?)%\s*,\s*(\d+(\.\d*)?)%\s*\)$", color)
```
hugovk: Do we need brackets around these new decimal matchers?

If they're not being used in a `m.group(X)`, they can be omitted to make the regex more readable (readable in relative terms!).

radarhere: Okay, I've updated the commit, see what you think.

PR #3147, opened 2018-05-30
Fix transform fill color for alpha images

fozcode:
This is a small code change to fix missing parameters when the `transform` function recurses.  Without this, alpha images (LA and RGBA) always have a transparent background regardless of the passed in fill color.

 * Pass through the `fillcolor` parameter when recursing

Tests all pass on Python 2.7 and 3.5.

### DIFF
--- a/Tests/test_image_rotate.py
+++ b/Tests/test_image_rotate.py
@@ -109,6 +109,19 @@ class TestImageRotate(PillowTestCase):
         im = im.rotate(45, fillcolor='white')
         self.assert_image_equal(im, target)
 
+    def test_alpha_rotate_no_fill(self):
+        # Alpha images are handled differently internally
+        im = Image.new('RGBA', (10, 10), 'green')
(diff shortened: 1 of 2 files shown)


hugovk:
Please could you add unit tests that make use of this change, and fail without the change and pass with it?

Have a look at:
https://github.com/python-pillow/Pillow/blob/master/Tests/test_image_transform.py

fozcode:
Yes of course, I'll get going on that now.  Thanks for the pointer for where to look.

fozcode:
I've added tests using rotation, because that makes for much simpler test code than using `transform` directly and having to implement a manual transformation.  Hope that's OK.

Before the fix, `test_alpha_rotate_no_fill` passes but `test_alpha_rotate_with_fill` fails.  
After the fix, both tests pass.

I think coverage is being a bit harsh failing with -0.15% for a 2 word code change.  All I can think of is to test an LA image as well as RGBA if you think it necessary.

PR #3148, opened 2018-05-30
Added getrgb HSB/HSV color strings

### LINKED ISSUE #2118: PIL.ImageColor.getrgb add support for HSB/HSV format
The current `PIL.ImageColor.getrgb` does not support [HSB/HSV](https://en.wikipedia.org/wiki/HSL_and_HSV) format, only the related HSL, but for easier conversion of some coordinates it would be nice to have support for HSB/HSV also.

radarhere:
Resolves #2118

### DIFF
--- a/docs/reference/ImageColor.rst
+++ b/docs/reference/ImageColor.rst
@@ -31,6 +31,11 @@ The ImageColor module supports the following string formats:
   (black=0%, normal=50%, white=100%). For example, ``hsl(0,100%,50%)`` is pure
   red.
 
+* Hue-Saturation-Value (HSV) functions, given as ``hsv(hue, saturation%,
+  value%)`` where hue and saturation are the same as HSL, and value is between
+  0% and 100% (black=0%, normal=100%). For example, ``hsv(0,100%,100%)`` is
+  pure red.
+
 * Common HTML color names. The :py:mod:`~PIL.ImageColor` module provides some
   140 standard color names, based on the colors supported by the X Window
   system and most web browsers. color names are case insensitive. For example,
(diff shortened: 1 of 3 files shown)


src/PIL/ImageColor.py
```diff
@@ -101,6 +101,20 @@ def getrgb(color):
             int(rgb[2] * 255 + 0.5)
             )
 
+    m = re.match(r"hs[bv]\(\s*(\d+\.?\d*)\s*,\s*(\d+\.?\d*)%\s*,\s*(\d+\.?\d*)%\s*\)$", color)
```
hugovk: Here, and in others, looks like the initial opening, and final closing brackets are also redundant.

Shall we remove these to make the regex a bit shorter? The `m.group` calls can then begin from 0.

docs/reference/ImageColor.rst
```diff
@@ -31,6 +31,13 @@ The ImageColor module supports the following string formats:
   (black=0%, normal=50%, white=100%). For example, ``hsl(0,100%,50%)`` is pure
   red.
 
+* Hue-Saturation-Value (HSV) functions, given as ``hsv(hue, saturation%,
+  value%)`` where hue and saturation are the same as HSL, and value is a value
+  between 0% and 100% (black=0%, normal=100%). For example,
```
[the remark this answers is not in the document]
radarhere: Okay, updated.

src/PIL/ImageColor.py
radarhere: The initial and final closing brackets are escaped - they are part of the string, e.g. `hsb(0,100%,50%)`

hugovk: Haha, we'd better keep them then!

PR #3155, opened 2018-06-05
Corrected test

radarhere:
Instead of raising the `Exception` at https://github.com/python-pillow/Pillow/blob/master/src/PIL/TiffImagePlugin.py#L484 as intended, this test was raising a `NameError` because `ImageFileDirectory_v2` is not imported.

### DIFF
--- a/Tests/test_file_tiff.py
+++ b/Tests/test_file_tiff.py
@@ -63,8 +63,11 @@ class TestFileTiff(PillowTestCase):
         im.load()
 
     def test_set_legacy_api(self):
-        with self.assertRaises(Exception):
-            ImageFileDirectory_v2.legacy_api = None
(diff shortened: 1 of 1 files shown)


hugovk:
Good spot, asserting against `Exception` is a bit too general.

Should we also assert the exception message to be sure it's the right one?

```python
    def test_set_legacy_api(self):
        ifd = TiffImagePlugin.ImageFileDirectory_v2()
        with self.assertRaises(Exception) as e:
            ifd.legacy_api = None
        self.assertEqual(str(e.exception),
                         "Not allowing setting of legacy api")
```

radarhere:
Okay, sure, I've updated the commit with your code.

hugovk:
The AppVeyor failure is unrelated, I've created https://github.com/python-pillow/Pillow/issues/3156 for that.

Issue #2309, opened 2016-12-25
Unhelpful error if libwebp is not installed


khazhyk:
### What did you do?
Tried to open a webp file.
### What did you expect to happen?
The webp file opened, or PIL would notify me that it was compiled without webp support.
### What actually happened?
Got generic 'OSError: cannot identify image file'
### What versions of Pillow and Python are you using?
Python 3.5, Pillow 3.4.2

Judging by https://github.com/python-pillow/Pillow/issues/1930 a more specific error doesn't seem unreasonable, and could speed up troubleshooting for others in the future.

aclark4life:
@khazhyk Can you send a pull request with the changes you want? Also what platform did you install on? I know that if you compile Pillow you'll see webp missing… but maybe you installed a pre-built wheel.

khazhyk:
I had installed (not from wheel) a while ago and just so happened to recently to start using webp.

From what I understand by the code, a plugin will fail to import (and register) if the native extension isn't compiled (and logs to debug), which means by the time we reach open() we can't distinguish between something that matches no plugins and something that would match a plugin if it's native portion had been compiled. It seems like that could be changed to allow for that, although I don't fully understand PIL's design or rationale for what implications that would have.

radarhere:
I've created a PR as a possible solution to this.

hugovk:
Fixed by https://github.com/python-pillow/Pillow/pull/3169.

PR #3169, opened 2018-06-09
Add warnings if image file identification fails due to lack of support

radarhere:
Suggestion for #2309


radarhere:
I've fixed the error that started occurring in AppVeyor - it was due to combining this PR with a new test from #3341

PR #3172, opened 2018-06-10
Add convert_mode param when saving

### LINKED ISSUE #2663: Convert image to supported mode before saving
We need explicit method to automatically convert image to one of the supported modes for saving formats. Affects #2609 and others. Maybe it should be part of ImagePlugins.

This is a good start point:

```python
def fix_mode(image, new_format, fill_color='white'):
    old_mode = image.mode

    if old_mode in ('RGBA', 'LA') and new_format == 'GIF':
        alpha = image.getchannel('A')
        # Convert the image into P mode but only use 255 colors
        # in the palette out of 256.
        image = image.convert(old_mode[:-1]) \
            .convert('P', palette=Image.ADAPTIVE, colors=255)
        # Set all pixel values below 128 to 255, and the rest to 0.
        mask = Image.eval(alpha, lambda px: 255 if px < 128 else 0)
        # Paste the color of index 255 and use alpha as a mask.
        image.paste(255, mask)
        # The transparency index is 255.
        image.info['transparency'] = 255
        return image

    if old_mode == 'CMYK' and new_format in ('PNG', 'WEBP', 'GIF'):
        return image.convert('RGB')

    if old_mode == 'I' and new_format in ('JPEG', 'GIF', 'WEBP'):
        image = image.point([i//256 for i in range(65536)], 'L')
        old_mode = image.mode

    if old_mode == 'P' and new_format in ('JPEG', 'WEBP'):
        # PIL loses palette attribute during crop. So we use low-level api.
        alpha = 'A' in image.im.getpalettemode()
        image = image.convert('RGBA' if alpha else 'RGB')
        old_mode = image.mode

    if old_mode in ('RGBA', 'LA') and new_format == 'JPEG':
        background = Image.new(old_mode[:-1], image.size, fill_color)
        background.paste(image, image.getchannel('A'))
        image = background
        old_mode = image.mode

    # Pillow doesn't support L modes for webp for now.
    if old_mode in ('L', 'LA') and new_format == 'WEBP':
        image = image.convert('RGB' + old_mode[1:])
        old_mode = image.mode

    return image
```

radarhere:
Resolves #2663


radarhere:
Okay, I've added a commit. See what you think

src/PIL/Image.py
```diff
@@ -1929,11 +1923,23 @@ def save(self, fp, format=None, **params):
 
         if format.upper() not in SAVE:
             init()
+        save_all = False
+        if 'save_all' in params:
+            save_all = params.pop('save_all')
```
homm: This could be `save_all = params.pop('save_all', False)`

src/PIL/Image.py
```diff
@@ -1929,11 +1923,23 @@ def save(self, fp, format=None, **params):
...
         if save_all:
             save_handler = SAVE_ALL[format.upper()]
         else:
             save_handler = SAVE[format.upper()]
 
+        if params.get('convert_mode'):
+            plugin = sys.modules[save_handler.__module__]
+            convertedIm = self._convert_mode(plugin, params)
```
homm: This variable name violates pep8: https://www.python.org/dev/peps/pep-0008/?#function-and-variable-names

src/PIL/Image.py
```diff
@@ -1949,6 +1955,57 @@ def save(self, fp, format=None, **params):
             if open_fp:
                 fp.close()
 
+    def _convert_mode(self, plugin, params):
```
homm: I'd prefer to have a function which takes a list of modes, rather than a plugin. It could be useful in other places, not only for saving.

src/PIL/Image.py
```diff
@@ -1949,6 +1955,57 @@ def save(self, fp, format=None, **params):
...
+            if 'A' in self.im.getpalettemode():
+                preferred_modes.append('RGBA')
+            preferred_modes.append('RGB')
+        else:
+            preferred_modes = {
+                'RGBA': ['RGB'],
+                'LA': ['RGBA', 'P', 'L'],
+                'L': ['RGB'],
+                'I': ['L', 'RGB'],
+                'CMYK': ['RGB']
```
homm: There should be a full list of possible modes, for example 'YCbCr' and others.

homm (dismissed):

src/PIL/Image.py
radarhere: Good call, thanks. I've made that change.

src/PIL/Image.py
radarhere: Okay, I've changed that.

src/PIL/Image.py
radarhere: Okay, done.

src/PIL/Image.py
radarhere: Okay, I've updated the list, see what you think

homm:
In general, this is great. The only drawback for me is binding of supported modes for the save function via the parent's module attribute. In particular, this prevents different sets of supported modes in one module. Isn't better to explicitly assign the supported modes to the save function?

```python
def _save(im, fp, filename):
   pass

_save._supported_modes = ['RGB', 'RGBA', 'P', 'I', 'LA', 'L', '1']
```

PR #3178, opened 2018-06-12
TGA: Read and write LA data

danpla:


Tests/test_file_tga.py
```diff
@@ -60,8 +57,16 @@ def test_save_rle(self):
...
+    def test_save_l_transparency(self):
+        in_file = "Tests/images/l_trns.tga"
+        im = Image.open(in_file)
+
+        test_file = self.tempfile("temp.tga")
+        im.save(test_file)
+
+        # There are 559 transparent pixels.
+        im = im.convert('RGBA')
+        self.assertEqual(im.getchannel('A').getcolors()[0][0], 559)
```
hugovk: * This test opens l_trns.tga and saves it to a temp file.
* It then converts the image to RGBA and asserts the number of pixels.

It's good it's making sure the save runs without exception, but it doesn't verify the save.

How about opening the saved file and asserting something there? Have a look at the other tests here.

docs/handbook/image-file-formats.rst
```diff
@@ -927,11 +934,6 @@ PSD
 PIL identifies and reads PSD files written by Adobe Photoshop 2.5 and 3.0.
 
 
-TGA
-^^^
-
-PIL reads 24- and 32-bit uncompressed and run-length encoded TGA files.
```
hugovk: Do we not need to mention the 24- and 32-bit any more?

docs/handbook/image-file-formats.rst
```diff
@@ -558,6 +558,13 @@ For more information about the SPIDER image processing package, see the
 .. _SPIDER homepage: https://spider.wadsworth.org/spider_doc/spider/docs/spider.html
 .. _Wadsworth Center: https://www.wadsworth.org/
 
+TGA
```
hugovk: Thanks for re-ordering these!

PR #3186, opened 2018-06-15
TGA: Add support for writing RLE data

danpla:


danpla:
AppVeyor's PyPy build fails with en exception from pytest. It's probably not my failure, as the tga tests was OK.

danpla:
Never mind, that was a AppVeyor's glitch.

Issue #3089, opened 2018-04-11
5.1.1 release


wiredfool:
* Build osx wheels with xcode 6.4, supporting older OSX versions. (https://github.com/python-pillow/Pillow/issues/3068) 
* Include #3067 ?
* ~Wheels for 3.7, if we can convince the pillow-wheels repo to do that easily? (https://github.com/python-pillow/Pillow/issues/3074)~

wiredfool:
I'd appreciate any help on this.

matthew-brett:
I believe it's dangerous to build 3.7 wheels before the official release, as the binary API may change.

homm:
> Include #3067 ?

I'm for it. Compilation problems are that can be fixed in minor updates

paregorios:
Respectfully, from the user gallery: I'd prefer to see Pillow 5.1.1 not get held up for Python 3.7 official release, which is iirc still a couple months away. Because #3068. Thanks.

wiredfool:
Ok, #3067 and xcode6.4 wheels.

hugovk:
I've done the first steps of https://github.com/python-pillow/Pillow/blob/master/RELEASING.md#point-release on https://github.com/python-pillow/Pillow/tree/5.1.x:

---

## Point Release

Released as needed for security, installation or critical bug fixes.

* [x] Make necessary changes in ``master`` branch.
* [x] Update `CHANGES.rst`.
* [x] Cherry pick individual commits from ``master`` branch to release branch e.g. ``2.9.x``.
* [x] Check [Travis CI](https://travis-ci.org/python-pillow/Pillow) to confirm passing tests in release branch e.g. ``2.9.x``.
* [x] Checkout release branch e.g.:
```
    git checkout -t remotes/origin/2.9.x
```
* [x] In compliance with https://www.python.org/dev/peps/pep-0440/, update version identifier in `src/PIL/_version.py`
* [x] Run pre-release check via `make release-test`.
* [ ] Create tag for release e.g.:
```
    $ git tag 2.9.1
    $ git push --tags
```
* [ ] Create source distributions e.g.:
```
    $ make sdist
```
* [ ] Create [binary distributions](#binary-distributions)

---

Plus release notes. 

[...]

PR #3188, opened 2018-06-16
Merge 5.1.1 into master

radarhere:


hugovk:
This should only be merged if 5.1.1 is released (https://github.com/python-pillow/Pillow/issues/3089), which looks unlikely with 5.2.0 due out in two weeks.

Issue #3031, opened 2018-03-06
Undefined behavior in QuantOctree.c


sk-:
In `libImaging/QuantOctree.c` the function `ImagingEffectNoise` has an undefined behavior in [line 151](https://github.com/python-pillow/Pillow/blob/master/src/libImaging/QuantOctree.c#L151):

Count can actually be `0` as in the test `test_file_png.py` and the values need to be clamped to the range `(0, MAX_UCHAR)`.

This is detected when running the test `test_file_png.py` and compiling the library with UndefinedBehaviorSanitizer. See https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html to see how to detect this.

aclark4life:
@sk- Send a PR?

wiredfool:
Probably need to add the clip macro.

sk-:
The clip macro is not enough, as the count can also be `0`. Also note that there's no reusable clip macro, as it is defined 4 times in C files (see https://github.com/python-pillow/Pillow/search?utf8=%E2%9C%93&q=CLIP&type=)

homm:
> Also note that there's no reusable clip macro, as it is defined 4 times in C files (
@sk- This no longer the case ) CLIP8 macro was added.

hugovk:
(In https://github.com/python-pillow/Pillow/pull/3107)

tkoeppe:
OK, at least in `subtract_color_buckets` we should flat out do nothing if the subtrahend is empty.

hugovk:
> OK, at least in `subtract_color_buckets` we should flat out do nothing if the subtrahend is empty.

Merged in https://github.com/python-pillow/Pillow/issues/3031.

radarhere:
At the time of this PR's creation, this is line 151 and following - https://github.com/python-pillow/Pillow/blob/262b6d15ccc5af3e31f6b6a8092ff9808bd991ff/src/libImaging/QuantOctree.c#L151-L155

This undefined behaviour was actually fixed in #3388

PR #3196, opened 2018-06-21
[QuantOctree.c] Remove erroneous attempt to average over an empty range

tkoeppe:
Partially addresses #3031.


tkoeppe:
Thanks a lot!

Issue #2071, opened 2016-08-12
UnboundLocalError: local variable 'err_code' referenced before assignment


shayanelhami:
On this line 
[ImageFile.py Line 238](https://github.com/python-pillow/Pillow/blob/master/PIL/ImageFile.py#L238)

`if not self.map and not LOAD_TRUNCATED_IMAGES and err_code < 0:`

It happens when there is a ValueError exception on line 194:

``` python
    try:
        decoder.setimage(self.im, extents)
    except ValueError:
        continue
```

which leave no opportunity for err_code to be set.  
### What did you do?

Say I have a 128x128 png with one zip tile:
![128x128](https://cloud.githubusercontent.com/assets/1285365/17634055/9429cb88-60c8-11e6-91c2-0dac8d38a086.png)

and I use this code:

``` python
from PIL import Image
im = Image.open("128x128.png")
im.size = (32, 32)
im.save("32x32.png")
```
### What did you expect to happen?

Instead of an UnboundLocalError it should at least raise a meaningful error.
### What versions of Pillow and Python are you using?

Python 2.7.12, Pillow commit 533f4145

homm:
We should make read-only property and raise on this:

```python
im.size = (32, 32)
```

wiredfool:
Agreed.

homm:
Actually, the size assigning is used at least [for icns files](https://github.com/python-pillow/Pillow/blob/master/Tests/test_file_icns.py#L47).

We'll need to create some smart property which will accept only allowed sizes and only before loading.

radarhere:
@homm what do you mean by 'only allowed sizes'?

PeterJCLaw:
Not sure if it's the same issue, but I've seen a similar symptom when trying to load a (corrupted) GIF file. The GIF in question causes a `ValueError` at the same location and then also the `UnboundLocalError`. My scenario didn't involve any assignment of the size of the image, just doing `Image.open("bad.gif").load()` was enough.

alexwlchan:
@radarhere Rather than changing all the internal setters to use `size`, could you use `@size.setter` and implement the custom logic there? Something like:

```
@property
def size(self):
    return self._size

@size.setter
def size(self, value)
    # validation logic goes here...
    self._size = value
```

Then all the existing assignments continue to work without modifications, but we get the nice validation logic.

I’d offer a patch, but I confess that it’s not completely obvious to me what the conditions are that should trip those validation checks.

[...]

Issue #3044, opened 2018-03-16
ValueError: tile cannot extend outside image


fowler0503:
### What did you do?
I'm attempting to load a satellite image from Planet.com by running:
image = Image.open(image_file_name)
pix = image.load()
When my code attempts to run image.load() I receive the above referenced error

### What did you expect to happen?
I'm expected the tif image to load to be parsed for vectors

### What actually happened?
I receive the error: ValueError: tile cannot extend outside image

### What versions of Pillow and Python are you using?
Python 3.6.3 :: Anaconda, Inc
Pillow: 1.1.7

Please include **code** that reproduces the issue and whenever possible, an **image** that demonstrates the issue. Please upload images to GitHub, not to third-party file hosting sites. If necessary, add the image to a zip or tar archive.

The best reproductions are self-contained scripts with minimal dependencies. If you are using a framework such as plone, Django, or buildout, try to replicate the issue just using Pillow. 
[PlanetImage.tif.zip](https://github.com/python-pillow/Pillow/files/1820366/PlanetImage.tif.zip)
[PlanetImage.tif.zip](https://github.com/python-pillow/Pillow/files/1820368/PlanetImage.tif.zip)

```python
>>> from PIL import Image, ImageDraw
>>> image_file_name=('PlanetImage.tif')
>>> image = Image.open(image_file_name)
>>> pix = image.load()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/jwfowler/anaconda3/envs/ship_classify/lib/python3.6/site-packages/PIL/TiffImagePlugin.py", line 1053, in load
    return super(TiffImageFile, self).load()
  File "/Users/jwfowler/anaconda3/envs/ship_classify/lib/python3.6/site-packages/PIL/ImageFile.py", line 206, in load
    decoder.setimage(self.im, extents)
ValueError: tile cannot extend outside image

```

wiredfool:
The tiff that you're loading has tiles that extend outside the image boundaries. Basically, the reported size of the image is wrong with respect to the encoded data.

There are two things that can be done, either set the image size to the maximum tile extents, or strip out the tile elements that are outside the reported image size:
```
>>> im = Image.open('issue_3044.tif')
>>> im.tile
[('raw', (0, 0, 128, 128), 2829, ('RGB', 0, 1)), 
('raw', (128, 0, 256, 128), 51981, ('RGB', 0, 1)), 
...
[...]

PR #3203, opened 2018-06-23
Changed Image size property to be read-only by default

radarhere:
Suggestion for #2071

### DIFF
--- a/docs/releasenotes/5.3.0.rst
+++ b/docs/releasenotes/5.3.0.rst
@@ -39,6 +39,22 @@ and size, new method ``ImageOps.pad`` pads images to fill a requested aspect
 ratio and size, filling new space with a provided ``color`` and positioning the
 image within the new area through a ``centering`` argument.
 
+Image Size
+==========
+
+If you attempt to set the size of an image directly, e.g.
+``im.size = (100, 100)``, you will now receive an ``AttributeError``. This is
+not about removing existing functionality, but instead about raising an
+explicit error to prevent later consequences. The ``resize`` method is the
+correct way to change an image's size.
+
+The exceptions to this are:
+* The ICO and ICNS image formats, which use ``im.size = (100, 100)`` to select
+a subimage.
+* The TIFF image format, which now has a ``DeprecationWarning`` for this
+action, as direct image size setting was previously necessary to work around an
+issue with tile extents.
+
 Other Changes
 =============
 
--- a/src/PIL/IcnsImagePlugin.py
+++ b/src/PIL/IcnsImagePlugin.py
@@ -265,12 +265,32 @@ class IcnsImageFile(ImageFile.ImageFile):
     def _open(self):
         self.icns = IcnsFile(self.fp)
         self.mode = 'RGBA'
+        self.info['sizes'] = self.icns.itersizes()
         self.best_size = self.icns.bestsize()
         self.size = (self.best_size[0] * self.best_size[2],
                      self.best_size[1] * self.best_size[2])
-        self.info['sizes'] = self.icns.itersizes()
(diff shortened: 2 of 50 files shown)


hugovk:
Will this break existing code? Is it okay to break existing code in this case?

If not, should we have a deprecation warning first?

src/PIL/IcnsImagePlugin.py
```diff
@@ -265,13 +265,33 @@ class IcnsImageFile(ImageFile.ImageFile):
...
         # Just use this to see if it's loaded or not yet.
         self.tile = ('',)
 
+    @property
+    def size(self):
+        return self._size
+
+    @size.setter
+    def size(self, value):
+        infoSize = value
```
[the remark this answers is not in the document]
radarhere: Okay, done

src/PIL/IcnsImagePlugin.py
```diff
@@ -265,13 +265,33 @@ class IcnsImageFile(ImageFile.ImageFile):
...
+        if not infoSize in self.info['sizes'] and len(infoSize) == 2:
```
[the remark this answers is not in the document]
radarhere: Ok, done

src/PIL/IcoImagePlugin.py
```diff
@@ -263,6 +263,17 @@ def _open(self):
         self.size = self.ico.entry[0]['dim']
         self.load()
 
+    @property
+    def size(self):
+        return self._size
+
+    @size.setter
+    def size(self, value):
+        if not value in self.info['sizes']:
```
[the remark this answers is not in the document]
radarhere: Okay.

radarhere:
I think that this is about explicitly raising an error at when a user performs incorrect behaviour, rather than allowing their incorrect behaviour to produce bugs later on.

In the issue that started this, the user was receiving an error when saving. This just moves the error one step earlier.

The only situation I can think of where we might not want to immediately throw an error is for people writing their own image format plugins - they will now need to assign to `self._size` rather than `self.size`. That's a fairly simple change though.

radarhere:
In light of https://github.com/python-pillow/Pillow/issues/3044#issuecomment-375207867, where setting the size directly on a TIFF image is given as a solution, I have changed setting the size of a TIFF image to be only deprecated.

hugovk:
Let's also mention it in the release notes.

docs/releasenotes/5.3.0.rst
```diff
@@ -39,6 +39,22 @@ and size, new method ``ImageOps.pad`` pads images to fill a requested aspect
 ratio and size, filling new space with a provided ``color`` and positioning the
 image within the new area through a ``centering`` argument.
 
+Image Size
+==========
+
+If you attempt to set the size of an image directly. e.g.
+``im.size = (100, 100)``, you will now receive an AttributeError. This is not
```
hugovk: Let's use code formatting for the error (and below).

docs/releasenotes/5.3.0.rst
```diff
@@ -39,6 +39,22 @@ and size, new method ``ImageOps.pad`` pads images to fill a requested aspect
 ratio and size, filling new space with a provided ``color`` and positioning the
 image within the new area through a ``centering`` argument.
 
+Image Size
+==========
+
+If you attempt to set the size of an image directly. e.g.
```
hugovk: Comma after "directly"

docs/releasenotes/5.3.0.rst
radarhere: Okay, done

docs/releasenotes/5.3.0.rst
radarhere: Ok

PR #3205, opened 2018-06-25
Added transparency to matrix conversion

### LINKED ISSUE #3150: Save fails after converting 16-bit RGB to grayscale
### What did you do?
Downloaded test-PNG from "pngsuite" (see below), converted to grayscale using the matrix-option, saved the result.

that the script would succeed like with almost all other PNGs from "pngsuite"

### What actually happened?
Save fails with

```
  File "C:/base2_nl/py-grayscale/minimalExampleBug.py", line 5, in <module>
    im.save('tbbn2c16-out.png')
  File "C:\Python36\lib\site-packages\PIL\Image.py", line 1935, in save
  File "C:\Python36\lib\site-packages\PIL\PngImagePlugin.py", line 790, in _save
    transparency = max(0, min(65535, transparency))
TypeError: '<' not supported between instances of 'tuple' and 'int'
```

Pillow 5.1.0
Python 3.6.5

### Minimal example to reproduce the error
```python
from PIL import Image

im = Image.open('tbbn2c16.png') # from here: http://www.schaik.com/pngsuite/tbbn2c16.png
im = im.convert("L",  (0.22, 0.72, 0.06, 0))
im.save('tbbn2c16-out.png') # fails
```

BTW in this case transparency is a tuple with the following value before the exception:
```python
transparency  = (65535,65535,65535)
```

Probably line 790 would have produced the expected result in Python 2.7, but is invalid in Python 3.x when mixing integers and tuple parameters.

radarhere:
Resolves #3150

### DIFF
--- a/Tests/test_image_convert.py
+++ b/Tests/test_image_convert.py
@@ -187,6 +187,7 @@ class TestImageConvert(PillowTestCase):
         def matrix_convert(mode):
             # Arrange
             im = hopper('RGB')
+            im.info['transparency'] = (255, 0, 0)
             matrix = (
                 0.412453, 0.357580, 0.180423, 0,
                 0.212671, 0.715160, 0.072169, 0,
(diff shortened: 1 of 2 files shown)


radarhere:
Rebased.

Issue #3213, opened 2018-06-29
Unaligned access crash


DerDakon:
The test suite crashes when run on sparc machines at several places because of SIGBUS errors, i.e. unaligned memory accesses. This usually affects also other platforms (ARM, Itanium), but to a lesser degree (often only slowdown and no crashes).

I have a unfinished patch, currently targetting 4.3.0 as that is what is my current problem in Gentoo. A port to master and PR will follow later once I get the testsuite to pass again.

Further details and preliminary patch at https://bugs.gentoo.org/659460

aclark4life:
@DerDakon Sparc?! Well that makes my day. Anyway, feel free to open a PR with that patch …

DerDakon:
The first one is already there…

aclark4life:
Ah #3219 I assume, thanks

DerDakon:
No, #3219 is totally irrelevant for this and was just spotted while trying to fix those. #3215 has a few fixes, but they are not enough. My current working set can be found in [this branch](https://github.com/DerDakon/Pillow/tree/unaligned-access).

PR #3215, opened 2018-06-30
Replace copy operations with memcpy()

DerDakon:
This replaces trivial instances where a copy from one pointer to the other involves no further calculations or casts. The compiler will optimize this to whatever the platform offers.

This is a first step in direction of a fix for #3213.


DerDakon:
Abandoned in favor of #3225, which is the (hopefully) complete solution for that issue.

PR #3220, opened 2018-07-01
5.2.0 release notes

aclark4life:
@python-pillow/pillow-team If anyone wants to add anything to the 5.2.0 release notes, now would be the time! Else these can probably go out as-is.


aclark4life:
Not clear in this PR is that I also just added 885c9cb206f47b8edd02d5d61a191f61c1f1d92c

hugovk:
Thanks, made a slight update to mention the other two version-related PRs.

PR #3225, opened 2018-07-02
Use explicit memcpy() to avoid unaligned memory accesses

### LINKED ISSUE #3213: Unaligned access crash

DerDakon:
Fixes #3213.

This replaces casts of incoming memory with explicit memcpy(). Some platforms, especially Sparc, but also ARM and Itanium, do not like unaligned memory accesses and may lack from slowdowns or crashes as result of those if the incoming memory is not aligned to the expected access type. Using memcpy() will avoid this: the compiler will optimize this to direct unaligned accesses on platforms that support those operations, and will do whatever copy operation is needed on platforms that do not.

### DIFF
--- a/src/libImaging/Unpack.c
+++ b/src/libImaging/Unpack.c
@@ -32,7 +32,6 @@
 
 #include "Imaging.h"
 
-
 #define R 0
 #define G 1
 #define B 2
@@ -327,11 +326,11 @@ static void
 unpackLA(UINT8* _out, const UINT8* in, int pixels)
 {
     int i;
-    UINT32* out = (UINT32*) _out;
     /* LA, pixel interleaved */
     for (i = 0; i < pixels; i++) {
-        out[i] = MAKE_UINT32(in[0], in[0], in[0], in[1]);
-        in += 2;
+        UINT32 iv = MAKE_UINT32(in[0], in[0], in[0], in[1]);
+        memcpy(_out, &iv, sizeof(iv));
+        in += 2; _out += 4;
     }
 }
 
@@ -339,10 +338,10 @@ static void
 unpackLAL(UINT8* _out, const UINT8* in, int pixels)
 {
     int i;
-    UINT32* out = (UINT32*) _out;
     /* LA, line interleaved */
-    for (i = 0; i < pixels; i++) {
-        out[i] = MAKE_UINT32(in[i], in[i], in[i], in[i+pixels]);
+    for (i = 0; i < pixels; i++, _out+=4) {
+        UINT32 iv = MAKE_UINT32(in[i], in[i], in[i], in[i+pixels]);
+        memcpy(_out, &iv, sizeof(iv));
     }
 }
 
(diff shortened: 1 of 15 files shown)


homm:
Personally, I don't like it because, while compiler *will likely* optimize `memcpy` calls to just one move operation, it *doesn't have to* optimize it.

Do you have examples of such fixes in other wide-spread libraries?

DerDakon:
Is [python](https://github.com/python/cpython/pull/6123) common enough? For fixed size copies you must try very hard for the compiler not to optimize this on e.g. x86, this is one of the most common intrinsics that are in use.

DerDakon:
Ping?

DerDakon:
Ping?

DerDakon:
Ping?

DerDakon:
What is the way forward? I bet this needs updating, but I'm only going to do this if it would be accepted afterwards.

DerDakon:
ping?

DerDakon:
I would rebase it if there is a chance that this get's merged in time then, otherwise it is too much work for nothing.

hugovk:
There's been no objections from @homm or anyone else in a year, so let's merge this for Monday's release, if possible. Thank you!

src/libImaging/Unpack.c
```diff
@@ -480,86 +479,78 @@ void
 ImagingUnpackRGB(UINT8* _out, const UINT8* in, int pixels)
 {
     int i = 0;
-#ifdef __sparc
-    /* SPARC CPUs cannot read integers from nonaligned addresses. */
```
hugovk: Are the removals of these exceptions for SPARC intentional?

They were only added recently, in https://github.com/python-pillow/Pillow/pull/3858.

DerDakon: Yes, by using memcpy the compiler will care for the correct access pattern.

hugovk: OK, thanks.

cc @kulikjak fyi.

DerDakon:
I have tested this on my sparc machine and I see some test errors. I guess test errors and I suspect them being endianess issues. I guess this is still better than crashes. I'll probably need a few days to look into this.

DerDakon:
Ok, I was even faster than expected. When I compare the test results of my patch on sparc with current git master on hppa I have only these additional failures:

Tests/test_imagefont.py::TestImageFont::test_variation_get FAILED
Tests/test_imagefont.py::TestImageFont::test_variation_set_by_axes FAILED
Tests/test_imagefont.py::TestImageFont::test_variation_set_by_name FAILED

Please decide if they are likely caused by this patches or not, I have no clue.

DerDakon:
I have tested it on my hppa machine (also big endian) with and without this patch, and the font errors are also present there in both runs. So I don't see any regressions with that patch.

hugovk:
Thanks for testing. As no regressions found, let's merge it.

PR #3218, opened 2018-07-01
Update docstrings

hugovk:
Fixes #3082.

Changes proposed in this pull request:

 * about.rst: add last PIL release version and date
 * PIL/\_\_init__.py: update package and version module docstrings

```python
Python 3.7.0 (default, Jun 29 2018, 20:13:13)
[Clang 9.1.0 (clang-902.0.39.2)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import PIL
>>> help(PIL)
```
Gives:
image
...
image

PR #3233, opened 2018-07-04
Fix code for PYTHONOPTIMIZE

### LINKED ISSUE #3231: Pillow cannot be loaded in python optimize (2) mode
### What did you do?
I was loading the PIL module with 'import PIL' when running Python with PYTHONOPTIMZE=2 (or python -OO).

I Expected the PIL/Pillow Library to be imported.

### What actually happened?
I received an exception:
AttributeError: 'NoneType' object has no attribute 'format'

Pillow 5.2.0 / Python 3.6.5

### Reason:

PYTHONOPTIMIZE=2 / python -OO strips the _ _ doc _ _ variable.

## Example:

create a file called test.py with only import PIL

```python
import PIL
```

Execute it with python in optimize mode
```sh
python -OO test.py
```
You will receive the following exception:
```
  File "test.py", line 1, in <module>
    import PIL
  File "/usr/lib/python3.6/site-packages/PIL/__init__.py", line 27, in <module>
    __doc__ = __doc__.format(__version__)  # include version in docstring
AttributeError: 'NoneType' object has no attribute 'format'
```

hugovk:
Fixes #3231.

 * Fix bug introduced by https://github.com/python-pillow/Pillow/pull/3218 with `PYTHONOPTIMIZE=2`, because that level strips away docstrings resulting in '`NoneType' object has no attribute 'format'`

---

https://ziade.org/2015/11/25/should-i-use-pythonoptimize/ says:

> So yeah, **even if you don't use yourself those options, it's good practice to make sure that your python code is tested with all possible values for PYTHONOPTIMIZE.**
> 
> It's easy enough, just run your tests with **-O** and **-OO** and without, and make sure your code does not depend on docstrings or assertions.

---

 * Test two build jobs (chosen arbitrarily) with the two different `PYTHONOPTIMIZE` levels to ensure this stays fixed

 * Running `PYTHONOPTIMIZE=1` (same as `-O`) identified code depending on an assert. Convert it to an exception.

 * Running `PYTHONOPTIMIZE=2` (same as `-OO`) identified code depending on docstrings in `pycparser`, imported by `cffi`. `pycparser` doesn't support `PYTHONOPTIMIZE=2`:  https://github.com/eliben/pycparser/pull/198#issuecomment-317001670.

   * When `PYTHONOPTIMIZE=2`, skip tests that require `cffi`

   * ImageTk.py only needs `cffi` for PyPy. Import it only where needed, to avoid problems when `PYTHONOPTIMIZE=2`.

### DIFF
--- a/Tests/test_image_access.py
+++ b/Tests/test_image_access.py
@@ -14,7 +14,6 @@ else:
         import cffi
     except ImportError:
         cffi = None
-        pass
 
 
 class AccessTest(PillowTestCase):
--- a/Tests/test_image_access.py
+++ b/Tests/test_image_access.py
@@ -1,15 +1,20 @@
 from helper import unittest, PillowTestCase, hopper, on_appveyor
 
-try:
-    from PIL import PyAccess
-    import cffi
-except ImportError:
-    cffi = None
-    pass
-
 from PIL import Image
 import sys
 import os
+
+# CFFI imports pycparser which doesn't support PYTHONOPTIMIZE=2
+# https://github.com/eliben/pycparser/pull/198#issuecomment-317001670
+if os.environ.get("PYTHONOPTIMIZE") == "2":
+    cffi = None
+else:
+    try:
+        from PIL import PyAccess
+        import cffi
+    except ImportError:
+        cffi = None
+        pass
 
 
 class AccessTest(PillowTestCase):
--- a/Tests/test_image_access.py
+++ b/Tests/test_image_access.py
@@ -2,8 +2,9 @@ from helper import unittest, PillowTestCase, hopper, on_appveyor
 
 try:
     from PIL import PyAccess
+    import cffi
 except ImportError:
-    # Skip in setUp()
+    cffi = None
     pass
 
 from PIL import Image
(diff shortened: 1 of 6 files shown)


Tests/test_image_access.py
```diff
@@ -1,15 +1,21 @@
...
+# https://github.com/eliben/pycparser/pull/198#issuecomment-317001670
+if os.environ.get("PYTHONOPTIMIZE") == "2":
+    cffi = None
+else:
+    try:
+        from PIL import PyAccess
+        import cffi
+    except ImportError:
+        cffi = None
+        pass
```
radarhere: Unless you feel otherwise, `pass` is unnecessary here

hugovk: Thanks, removed.

src/PIL/ImageTk.py
```diff
@@ -192,7 +185,11 @@ def paste(self, im, box=None):
                 from . import _imagingtk
                 try:
                     if hasattr(tk, 'interp'):
-                        # Pypy is using a ffi cdata element
+                        # Required for PyPy, which always has CFFI installed
+                        from cffi import FFI
+                        ffi = FFI()
+
+                        # Pypy is using an FFI cdata element
```
radarhere: PyPy

hugovk: Thanks, fixed.

hugovk:
Rebased on master to resolve conflicts.

src/PIL/__init__.py
```diff
@@ -24,8 +24,6 @@
 
 del _version
 
-__doc__ = __doc__.format(__version__)  # include version in docstring
```
radarhere: You're not interested in keeping the version in the docstring?
```
if __doc__ is not None:
    __doc__ = __doc__.format(__version__)  # include version in docstring
```

hugovk:
Merging as it's important for this to be in the next release (5.3.0).

PR #3234, opened 2018-07-05
Add Python 3.7 and Xenial to Travis CI

hugovk:
The main thing here is adding Python 3.7 (and removing `3.7-dev`), which requires Xenial.

Whilst we're at it, we can test most Python versions on both Trusty and Xenial.

If we don't want all, we should probably do at least a few on both.

 * Travis CI has Ubuntu 16.04 Xenial images available
 * Switch to Xenial as the base
 * Test on both Ubuntu 14.04 Trusty and Xenial where supported:
   * PyPy2/3 and 3.4 only support Trusty
   * Python 3.7 only supports Xenial due to OpenSSL requirements (https://github.com/travis-ci/travis-ci/issues/9069)

| Python | Trusty | Xenial |
| ------ | ------ | ------ |
| PyPy2  | yes    | no     |
| PyPy3  | yes    | no     |
| 3.7    | no     | yes    |
| 3.6    | yes    | yes    |
| 3.5    | yes    | yes    |
| 3.4    | yes    | no     |
| 2.7    | yes    | yes    |


.travis.yml
```diff
@@ -12,13 +14,24 @@ matrix:
...
+    - python:  "2.7_with_system_site_packages" # For PyQt4
+      dist: trusty
+    - python: '3.6'
     - python: '3.6'
+      dist: trusty
     - python: '3.5'
+    - python: '3.5'
+      dist: trusty
     - python: '3.4'
+      dist: trusty
```
radarhere: You've missed 3.4 on Xenial.

radarhere: Nevermind, I see now.

hugovk: PyPy2/3 and 3.4 only support Trusty on Travis.

Here's how they look: https://travis-ci.org/hugovk/Pillow/builds/397828420

hugovk:
Any comments on this? If not, I'll merge it next week.

It adds 4 new jobs to the build, increasing build time from ~16m to ~22m.

Trusty 14.04 goes [EOL in 2019-04](https://linuxlifecycle.com) and Python 3.4 goes [EOL on 2019-03-16](https://en.wikipedia.org/wiki/CPython#Version_history), so we can remove 4-6 jobs relatively soon, and should start testing on Xenial 16.04.

(Hopefully by then PyPy/3 will be available on something newer, otherwise we could keep those on Trusty a bit longer.)

PR #3241, opened 2018-07-07
Capture pkg-config output even if only one component returned

kkopachev:
Changing pkg-config output parser in case it returns only one component
(only -I or only -L)

Currently an issue with libtiff in official python container:
```
> docker run --rm -it --entrypoint=bash python:3
# pkg-config --libs-only-L libtiff-4 --cflags-only-I libtiff-4
 -I/usr/include/x86_64-linux-gnu 
# pip install pillow --global-option="build_ext" --global-option="--disable-zlib" \
 --global-option="build_ext" --global-option="--disable-jpeg" \
 --global-option="build_ext" --global-option="--enable-tiff"

...
    The headers or library files could not be found for tiff,
    which was requested by the option flag --enable-tiff
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -154,8 +154,8 @@ def _pkg_config(name):
         ]
         if not DEBUG:
             command.append('--silence-errors')
-        libs = subprocess.check_output(command).decode('utf8').split(' ')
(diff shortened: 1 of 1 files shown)


kkopachev:
Fixed by #3395

PR #3242, opened 2018-07-07
Add feature: ImageOps.colorize to support three colors

tsennott:
 * Added optional arguments to allow for a third color (`mid`) to be specified in `ImageOps.colorize()`, and the ability to choose where this color is applied in the [black, white] range (`midpoint`)
* While re-writing `colorize`, I formatted that function to pep8

 Comments:
* I was using `ImageOps.colorize()` to try to generate a image resembling a cyanotype, and found the two colors for mapping limiting. See attached images for a sense of how the change allows a more dynamic mapping
* I added the test `test_imageops.test_colorize()` to test the functionality and it is passing on Travis. However I'm not sure how to automatically test the gradient mapping without a test image, so I attached an image showing a stoplight gradient with a slight offset in the midpoint, demonstrating that the effect works smoothly. Test code to generate this is `ImageOps.colorize3(image_gradient, 'red', 'green', 'yellow', 100)` where `image_gradient` is a black to white gradient image
* I tested the docs and they seem to be building fine with my updated docstring

Two-color:
test_2color
Three color:
test_3color
Gradient test:
test_gradient_offset

### DIFF
--- a/Tests/test_imageops.py
+++ b/Tests/test_imageops.py
@@ -94,6 +94,22 @@ class TestImageOps(PillowTestCase):
         newimg = ImageOps.scale(i, 0.5)
         self.assertEqual(newimg.size, (25, 25))
 
+    def test_colorize(self):
+        # Test the colorizing function
+
+        # Grab test image
+        i = hopper("L").resize((15, 16))
+
+        # Test original 2-color functionality
+        ImageOps.colorize(i, 'green', 'red')
+
+        # Test new three color functionality (cyanotype colors)
+        ImageOps.colorize(i,
+                          (32, 37, 79),
+                          (255, 255, 255),
+                          (35, 52, 121),
+                          40)
+
 
 if __name__ == '__main__':
     unittest.main()
--- a/src/PIL/ImageOps.py
+++ b/src/PIL/ImageOps.py
@@ -136,28 +136,50 @@ def autocontrast(image, cutoff=0, ignore=None):
     return _lut(image, lut)
 
 
-def colorize(image, black, white):
+def colorize(image, black, white, mid=None, midpoint=128):
     """
-    Colorize grayscale image.  The **black** and **white**
-    arguments should be RGB tuples; this function calculates a color
-    wedge mapping all black pixels in the source image to the first
-    color, and all white pixels to the second color.
+    Colorize grayscale image.
+    This function calculates a color wedge mapping all
+    black pixels in the source image to the first
+    color, and all white pixels to the second color. If
+    mid is specified, it uses three color mapping.
+    The **black** and **white**
+    arguments should be RGB tuples; optionally you can use
+    three color mapping by also specifying **mid**, and
+    optionally, **midpoint** (which is the integer value
+    in [0, 255] corresponding to the midpoint color,
+    default 128).
 
     :param image: The image to colorize.
     :param black: The color to use for black input pixels.
     :param white: The color to use for white input pixels.
+    :param mid: The color to use for midtone input pixels.
+    :param midpoint: the int value [0, 255] for the mid color.
     :return: An image.
     """
     assert image.mode == "L"
     black = _color(black, "RGB")
+    if mid is not None:
+        mid = _color(mid, "RGB")
     white = _color(white, "RGB")
     red = []
     green = []
     blue = []
-    for i in range(256):
-        red.append(black[0]+i*(white[0]-black[0])//255)
-        green.append(black[1]+i*(white[1]-black[1])//255)
-        blue.append(black[2]+i*(white[2]-black[2])//255)
+    if mid is None:
+        for i in range(256):
+            red.append(black[0] + i * (white[0] - black[0]) // 255)
+            green.append(black[1] + i * (white[1] - black[1]) // 255)
+            blue.append(black[2] + i * (white[2] - black[2]) // 255)
+    else:
+        for i in range(0, midpoint):
+            red.append(black[0] + i * (mid[0] - black[0]) // 255)
+            green.append(black[1] + i * (mid[1] - black[1]) // 255)
+            blue.append(black[2] + i * (mid[2] - black[2]) // 255)
+        for i in range(0, 256 - midpoint):
+            red.append(mid[0] + i * (white[0] - mid[0]) // 255)
+            green.append(mid[1] + i * (white[1] - mid[1]) // 255)
+            blue.append(mid[2] + i * (white[2] - mid[2]) // 255)
+
     image = image.convert("RGB")
     return _lut(image, red + green + blue)
 
--- a/src/PIL/ImageOps.py
+++ b/src/PIL/ImageOps.py
@@ -154,10 +154,11 @@ def colorize(image, black, white, mid=None, midpoint=128):
     :param black: The color to use for black input pixels.
     :param white: The color to use for white input pixels.
     :param mid: The color to use for midtone input pixels.
-    :param midpoint: the int value [0, 255] for the mid color.
+    :param midpoint: the int value in [1, 254] for the mid color.
     :return: An image.
     """
     assert image.mode == "L"
+    assert 1 <= midpoint <= 254
 
     # Define colors from arguments
     black = _color(black, "RGB")
--- a/src/PIL/ImageOps.py
+++ b/src/PIL/ImageOps.py
@@ -147,8 +147,8 @@ def colorize(image, black, white, mid=None, midpoint=128):
     arguments should be RGB tuples; optionally you can use
     three color mapping by also specifying **mid**, and
     optionally, **midpoint** (which is the integer value
-    in [0, 255] corresponding to the midpoint color,
-    default 128).
+    in [1, 254] corresponding to where the midpoint color
+    should be mapped (0 being black and 255 being white).
 
     :param image: The image to colorize.
     :param black: The color to use for black input pixels.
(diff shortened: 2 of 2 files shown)


tsennott:
Not sure why the Travis build didn't pass, looks like everything except `TestDefaultFontLeak` for pypy3 passed, and even that one passed on my fork, so not sure what's going on. Let me know if i can help diagnose! Thanks,
Tim

radarhere:
When you say that 'I'm not sure how to automatically test the gradient mapping without a test image'
- if you mean that you don't know how to access your test image from your test code, you are able to add an image to the Tests/images directory as part of your PR. See #2922 for an example.
- if you mean that you don't have an image to be distributed under the Pillow license, you could use one of the existing images in the test suite

hugovk:
I restated the Travis job and it passed this time.

Tests/test_imageops.py
```diff
@@ -94,6 +94,22 @@ def test_scale(self):
...
         self.assertEqual(newimg.size, (25, 25))
 
+    def test_colorize(self):
+        # Test the colorizing function
+
+        # Grab test image
+        i = hopper("L").resize((15, 16))
+
+        # Test original 2-color functionality
+        ImageOps.colorize(i, 'green', 'red')
```
hugovk: It would be good to `assertEqual` some values from the returned image. For example, check the colour of a pixel on the left, on the right, and in the centre.

And/or `assert_image_equal` against a pre-colourised image.

Tests/test_imageops.py
```diff
@@ -94,6 +94,22 @@ def test_scale(self):
...
+
+        # Test new three color functionality (cyanotype colors)
+        ImageOps.colorize(i,
+                          (32, 37, 79),
+                          (255, 255, 255),
+                          (35, 52, 121),
+                          40)
```
hugovk: Same here, but in a new test case using a fresh input image.

Tests/test_imageops.py
```diff
@@ -94,6 +94,22 @@ def test_scale(self):
         newimg = ImageOps.scale(i, 0.5)
         self.assertEqual(newimg.size, (25, 25))
 
+    def test_colorize(self):
+        # Test the colorizing function
+
+        # Grab test image
+        i = hopper("L").resize((15, 16))
```
hugovk: Let's use the more standard `im` variable name. We can update the other test too.

Tests/test_imageops.py
tsennott: ok sounds good, working on that now

Tests/test_imageops.py
tsennott: do you mean i should split out the two tests (2 color, 3 color) into two different test methods in this class?

Tests/test_imageops.py
tsennott: ok great, sounds good, the `i` was confusing to me too

Tests/test_imageops.py
tsennott: i wasn't sure and i liked keeping them in one test, so i did that for now. let me know if i should change it though

tsennott:
@radarhere @hugovk thanks very much for your comments. I pushed a new commit with test images and it looks like it passed. If anything else should change just let me know. Thanks again, excited to be contributing to this great project.

source image
bw_gradient
comparison images
bw_gradient_2color
bw_gradient_3color

tsennott:
So I'm thinking of adding an additional feature, whereby you could optionally specify not just the `midpoint` but also the `blackpoint` and `whitepoint`, where pixels below and above the latter just get mapped as the custom `black` and `white` colors (basically flatten the lookup table at either end). This gives a lot more flexibility in the resulting image. I have it working on my local copy and it has made the cyanotype image much closer to what i wanted. See below. 

I was also thinking I could add an optional `preset` string argument with a few preset colormaps that would define the settings for you. The "cyanotype" example from above would be one, and I could come up with a few others that look cool and would allow users to apply them without specifying any values except the image to use and what preset to apply. I would list the possible preset values in the docstring.

What do you think about these two ideas? Is this making it too complicated? I think they are valuable additions that make this a really useful feature, but curious for input. 

Proposed new function prototype:
`def colorize(image, black=None, white=None, preset=None, mid=None, blackpoint=0, whitepoint=255, midpoint=127)`
Since there are a lot of defaulted arguments now, I would handle the case where users passed insufficient arguments in with an explanative exception.


Example "cyanotype" image:
cyanotype

tsennott:
OK, so I elected to implement `blackpoint` and `whitepoint`, but thought the presets idea was overkill so didn't do that.

I also changed the tests to evaluate pixel values (+/- 1) instead of comparing images, since I think this is more robust to future changes. In addition, I added additional test cases, so I think all the use cases are handled. They are all in one method, since they are very similar, but could be split up if that's better.

Let me know what you think. Thanks again.
Tim

src/PIL/ImageOps.py
```diff
@@ -136,28 +136,101 @@ def autocontrast(image, cutoff=0, ignore=None):
...
+        # Define ranges
+        range_low = range(0, blackpoint)
+        range_map = range(0, whitepoint - blackpoint)
+        range_high = range(0, 256 - whitepoint)
+
+        # Map
+        for i in range_low:
+            red.append(black[0])
+            green.append(black[1])
+            blue.append(black[2])
```
hugovk: This loop is the same in both `if` and `else` blocks. Let's remove the duplication and move it before the `if`.

Also, `range_low` is only used here, we can remove the variable entirely and do `for i in range(0, blackpoint):` directly.

src/PIL/ImageOps.py
```diff
@@ -136,28 +136,101 @@ def autocontrast(image, cutoff=0, ignore=None):
...
+        for i in range_map:
+            red.append(black[0] + i * (white[0] - black[0]) // len(range_map))
+            green.append(black[1] + i * (white[1] - black[1]) // len(range_map))
+            blue.append(black[2] + i * (white[2] - black[2]) // len(range_map))
+        for i in range_high:
+            red.append(white[0])
+            green.append(white[1])
+            blue.append(white[2])
```
hugovk: Similarly, this loop is the same in both `if` and `else` blocks. Let's remove the duplication and move it after the `else` block.

And `range_high` can be deleted and used directly as `for i in range(0, 256 - whitepoint):`.

src/PIL/ImageOps.py
```diff
@@ -136,28 +136,101 @@ def autocontrast(image, cutoff=0, ignore=None):
...
+
+    # Initial asserts
     assert image.mode == "L"
+    assert 0 <= whitepoint <= 255
+    assert 0 <= blackpoint <= 255
+    assert 0 <= midpoint <= 255
+    assert blackpoint <= whitepoint
+    if mid is not None:
+        assert blackpoint <= midpoint
+        assert whitepoint >= midpoint
```
hugovk: Perhaps these two could be a single `assert blackpoint <= midpoint <= whitepoint`?

src/PIL/ImageOps.py
```diff
@@ -136,28 +136,101 @@ def autocontrast(image, cutoff=0, ignore=None):
...
+    :param midpoint: an int value [0, 255] for the midtone mapping.
     :return: An image.
     """
+
+    # Initial asserts
     assert image.mode == "L"
+    assert 0 <= whitepoint <= 255
+    assert 0 <= blackpoint <= 255
+    assert 0 <= midpoint <= 255
+    assert blackpoint <= whitepoint
```
hugovk: It's adding assumptions, but perhaps `assert 0 <= blackpoint <= midpoint <= whitepoint <= 255` instead of these four? But only if those assumptions are correct.

And if so, the following asserts aren't needed.

Tests/test_imageops.py
```diff
@@ -94,6 +95,89 @@ def test_scale(self):
...
+        im_3c_offset = ImageOps.colorize(im,
+                                         black='red',
+                                         white='green',
+                                         mid='blue',
+                                         blackpoint=50,
+                                         whitepoint=200,
+                                         midpoint=100)
+
+        # Define function for approximate equality of tuples
+        def tuple_approx_equal(actual, target, thresh):
```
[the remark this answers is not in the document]
tsennott: agree

src/PIL/ImageOps.py
tsennott: agree

src/PIL/ImageOps.py
tsennott: agree

src/PIL/ImageOps.py
tsennott: i took your two suggestions and put them in an if/else mid is None, to handle the 2-color case where only `blackpoint` and `whitepoint` have been user-defined, and could be incompatible with the default `midpoint` which this is not used in that case. new code:
```
if mid is None:
    assert 0 <= blackpoint <= whitepoint <= 255
else:
    assert 0 <= blackpoint <= midpoint <= whitepoint <= 255
```

src/PIL/ImageOps.py
tsennott: see other comments on asserts. new commit coming soon.

hugovk:
Do you think `tuple_approx_equal` would be better as `assert_tuple_approx_equal` with the `assertTrue` moved inside, or left as is?

And please could you also add release notes for 5.3.0?

We'll need a new copy of https://github.com/python-pillow/Pillow/blob/master/docs/releasenotes/5.2.0.rst as 5.3.0.rst (and added to index.rst).

Thanks!

PR #3245, opened 2018-07-09
Lookup ld.so.cache instead of hardcoding search paths

### LINKED ISSUE #3244: Pillow setup.py find_library_file logic issue
### What did you do?
Install Pillow by source code
Install successfully.
### What actually happened?
Build error. 

      File "<string>", line 1, in <module>
      File "/tmp/pip-req-build-_pl57v2h/setup.py", line 807, in <module>
        raise RequiredDependencyException(msg)
    __main__.RequiredDependencyException:

    The headers or library files could not be found for zlib,
    a required dependency when compiling Pillow from source.

Pillow：
Pillow-5.2.0
Python:
Python 3.7.0

detail see: https://github.com/docker-library/python/issues/305

Because the `library_dirs` lose `/usr/lib/x86_64-linux-gnu` or `/lib/x86_64-linux-gnu`

pslacerda:
Fixes #3244

 * Search library dirs in `/etc/ld.so.cache` instead of hardcoded assumptions;
 * The `_find_strings_in_binary()` parses the cache and extract directories. 

It worked on my Fedora 28 workstation, but with little or even no modification it can works on FreeBSD, NetBSD and Android. Darwin architecture also has the `dyld` command that is similar to `ldconfig` on Linux.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -91,8 +91,8 @@ def _find_library_dirs_ldconfig():
             }
         abi_type = mach_map.get(machine, 'libc6')
 
-        # XXX assuming GLIBC's ldconfig (with option -p)
-        # XXX Alpine Linux uses musl that can't print cache
+        # Assuming GLIBC's ldconfig (with option -p)
+        # Alpine Linux uses musl that can't print cache
         args = ['/sbin/ldconfig', '-p']
         expr = r'.*\(%s.*\) => (.*)' % abi_type
         env = dict(os.environ)
--- a/setup.py
+++ b/setup.py
@@ -10,7 +10,6 @@
 from __future__ import print_function
 
 import os
-import platform as plat
 import re
 import struct
 import subprocess
@@ -395,9 +394,9 @@ class pil_build_ext(build_ext):
                 # system libraries (zlib) are installed in /system/lib
                 # headers are at $PREFIX/include
                 # user libs are at $PREFIX/lib
-                _add_directory(library_dirs,
-                                os.path.join(os.environ['ANDROID_ROOT'],
-                                            'lib'))
+                _add_directory(
+                    library_dirs, os.path.join(os.environ["ANDROID_ROOT"], "lib")
+                )
 
         elif sys.platform.startswith("netbsd"):
             _add_directory(library_dirs, "/usr/pkg/lib")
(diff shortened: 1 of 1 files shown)


pslacerda:
The build fails only at Alpine Linux docker image because Alpline's ldconfig (http://www.musl-libc.org/) don't has a cache. Related to https://github.com/docker-library/python/issues/111.

Note that `ctypes.util.find_library()` first try `ldconfig` then `gcc` and `ld` if the former fails.

pslacerda:
I just searched a bit about `ldconfig` and its cache.

Seems that the cache always [has the same filename](https://github.com/bminor/glibc/blob/09533208febe923479261a27b7691abef297d604/elf/ldconfig.c#L54), and I don't expect it to on a directory other than `/etc`, except probably `/var` like BSD (but that don't use `glibc`). And the cache format [is pretty stable](https://github.com/bminor/glibc/blob/315304d24a5149a3670217514256b4ae959089e3/elf/cache.c#L159), apparently there is currently two formats and probably the old one nobody cares anymore.

pslacerda:
Any thoughts on this? I had a hard time trying to improve it.

setup.py
```diff
@@ -74,6 +74,57 @@ def _dbg(s, tp=None):
...
+            'x86_64-64': 'libc6,x86-64',
+            'ppc64-64': 'libc6,64bit',
+            'sparc64-64': 'libc6,64bit',
+            's390x-64': 'libc6,64bit',
+            'ia64-64': 'libc6,IA-64',
+            }
+        abi_type = mach_map.get(machine, 'libc6')
+
+        # XXX assuming GLIBC's ldconfig (with option -p)
+        # XXX Alpine Linux uses musl that can't print cache
```
hugovk: Why the `XXX`s?

hugovk:
I'm not so familiar with this area so would prefer someone else to review it.

setup.py
hugovk: But why prefix these comments with `XXX`?

pslacerda: Just an awareness note.

hugovk: Let's remove the `XXX`s then, they're just normal comments.

pslacerda: For sure. Any more comments so I can do it together.

pslacerda:
Very cool that upstream developers can commit into pull requests.

Happy New year!

PR #3246, opened 2018-07-09
[QuantOctree.c] Remove further attempts to average over empty buckets

### LINKED ISSUE #3031: Undefined behavior in QuantOctree.c

tkoeppe:
Fixes #3031.

* Skip cube updates for empty buckets in "add_lookup_buckets".

* Don't explicitly set pixel value in create_palette_array;
  the array is already zero-initialized, and zero is as good as
  any value for the average over nothing.

*Important:* Please review these two changes very carefully. There is obviously some sort of logic error if you ever need to *consume* the "average over an empty range", so what I'd like you to confirm is that the values that we stop writing in this change have never actually been read, or at least if they've been read they've not been used further.

### DIFF
--- a/src/libImaging/QuantOctree.c
+++ b/src/libImaging/QuantOctree.c
@@ -296,6 +296,7 @@ void add_lookup_buckets(ColorCube cube, ColorBucket palette, long nColors, long 
    long i;
    Pixel p;
    for (i=offset; i<offset+nColors; i++) {
+      if (palette[i].count == 0) continue;
       avg_color_from_color_bucket(&palette[i], &p);
       set_lookup_value(cube, &p, i);
    }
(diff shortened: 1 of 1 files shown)


radarhere:
@tkoeppe you're happy for this to be closed then?

tkoeppe:
@radarhere: I've lost the detailed understanding of the problem, but if those other fixes make it so that we don't need to return early here anymore, then yes please, do close this. Thank you!

PR #3250, opened 2018-07-13
Added line joints

### LINKED ISSUE #2840: Polyline joints are ugly
When drawing a Polyline using PIL.ImageDraw.Draw.line(xy, fill=None, width=0), the segment joints are two rectangles that join together in the center of their widths. This leaves an ugly gap. To make it look pretty it could either extend the rectangle sides until they meet, or draw a radiused joint. Extending the sides could lead to very long extensions on highly acute angles, but a radiused joint works well in every case I can think of.

I suggest changing the function to:

`PIL.ImageDraw.Draw.line(xy, fill=None, width=0, radiused=False)`

and then for each point in xy[1:-2] draw an ellipse centered at the point and whose width and height are the same as the line width. Here is a snippet of how I did it externally

```
draw.line(plot,'black', Weight)
for joint in plot[1:-2]:
    draw.ellipse((joint[0]-Weight/2,joint[1]-Weight/2,
                  joint[0]+Weight/2,joint[1]+Weight/2),'black')
```

radarhere:
Resolves #2840 by adding a `joint` argument to ImageDraw.Draw.line.

The issue suggests curved joints, instead of extending line edges until they meet. However, I'm suggesting letting users do either, and let them work with the results. I'm also adding simple angled edges. See the test images for illustrations.

Also, for the curved joints, instead of using ellipses as suggested in the issue, I'm using pieslices, as it is faster.


src/PIL/ImageDraw.py
```diff
@@ -149,11 +150,114 @@ def ellipse(self, xy, fill=None, outline=None):
...
+                        math.degrees(math.atan2(
+                            end[0] - start[0], start[1] - end[1]
+                        )) % 360
+                        for start, end in ((xy[i-1], point), (point, xy[i+1]))
+                    ]
+                    if angles[0] == angles[1]:
+                        # This is a straight line, so no joint is required
+                        continue
+
+                    def coordAtAngle(coord, angle):
```
hugovk: PEP 8: `coord_at_angle`

radarhere: Okay, I've updated the commit.

radarhere:
I think that part of the problem is the difference between `line` and `polygon`.

```
im = Image.new("RGB", (500, 300))
draw = ImageDraw.Draw(im)

draw.line([(100, 250), (100, 150)], '#fff', 100)
draw.polygon([(50, 150), (150, 150), (150, 50), (50, 50)], '#f00')

draw.line([(250, 150), (350, 150)], '#fff', 100)
draw.polygon([(350, 100), (450, 100), (450, 200), (350, 200)], '#f00')

im.show()
```

draw

`polygon` is giving a different extra pixel in both rotations, making it hard to correct for.

radarhere:
I have now changed this PR to only be about curved joints. It no longer adds 'straight' or 'extend' joints.

hugovk:
👍

image

docs/reference/ImageDraw.rst
```diff
@@ -171,19 +171,20 @@ Methods
...
     :param fill: Color to use for the line.
-    :param width: The line width, in pixels. Note that line
-        joins are not handled well, so wide polylines will not look good.
+    :param width: The line width, in pixels.
 
         .. versionadded:: 1.1.5
 
         .. note:: This option was broken until version 1.1.6.
+    :param joint: Joint type between a sequence of lines. It can be "curve",
+                  for rounded edges, or None.
```
hugovk: Should we add `versionadded`?

radarhere: Done

PR #1343, opened 2015-07-13
Removed deprecated code

radarhere:
Removing different segments of code that have been deprecated long enough.

Currently under discussion in #1087.

PR #3256, opened 2018-07-14
Docs: Remove redundant module skipping

hugovk:
The skipping originally applied to ImageFileIO (#375), and ImageFileIO was later removed (#1343) but this line was missed.

(It wasn't obvious as the skipping line comes before the module.)


radarhere:
Ah. Thanks for that.

PR #3260, opened 2018-07-18
Added GitHub release to release checklist

radarhere:
5.2.0 wasn't added as a release to https://github.com/python-pillow/Pillow/releases until I did it just now - so this PR adds it to the release checklist.


hugovk:
Well spotted!

I'd probably put it last on each checklist, or at least after uploading the source and binaries, as I think it's best to make those available as soon as possible.

radarhere:
Okay. I've moved it to after the binary items in the checklist, but I have kept it before appending dev0 to the version number - it seems natural to me to complete the release before preparing for the next one.

Issue #2716, opened 2017-09-06
Image loading by file name memory leak is not properly fixed


homm:
`__del__` doesn't work, because in the most cases `self.fp` is [None after initial loading](https://github.com/python-pillow/Pillow/blob/master/PIL/ImageFile.py#L249) and actual fp pointer is stored in `self.__fp`

radarhere:
I have created PR #3261 to try and address this.

radarhere:
@homm is this now fixed on master?

aclark4life:
@hugovk @radarhere If you still think this can go in and no objection from @wiredfool or @homm I'd probably just merge it. (I see the PR was merged.)

hugovk:
The PR is merged, closing this issue let's re-open or create a new issue needed.

Issue #2786, opened 2017-10-04
Testing ImageDraw.Draw: "ResourceWarning: unclosed file"


hugovk:
### What did you do?

```
python3 Tests/test_imagedraw.py TestImageDraw.test_valueerror
```

### What did you expect to happen?

```
.
----------------------------------------------------------------------
Ran 1 test in 0.040s

OK
```

### What actually happened?

```
/usr/local/Cellar/python3/3.6.2/Frameworks/Python.framework/Versions/3.6/lib/python3.6/unittest/case.py:605: ResourceWarning: unclosed file <_io.BufferedReader name='Tests/images/chi.gif'>
  testMethod()
.
----------------------------------------------------------------------
Ran 1 test in 0.040s

OK
```

### What versions of Pillow and Python are you using?

Latest Pillow master and Python 3.6.2.
Does not happen with Python 2.7.13.

```python
    def test_valueerror(self):
        im = Image.open("Tests/images/chi.gif")

        draw = ImageDraw.Draw(im)   # comes from here
        # draw.line(((0, 0)), fill=(0, 0, 0))  # not needed to reproduce
```
Or without the test suite:
```python
from PIL import Image, ImageDraw
import warnings

warnings.simplefilter("always")

im = Image.open("Tests/images/chi.gif")
draw = ImageDraw.Draw(im)   # comes from here
```

wiredfool:
When running on python3.5-dbg:
```
python test-installed.py
/home/erics/Pillow/Tests/test_file_webp_metadata.py:21: ResourceWarning: unclosed file <_io.BufferedReader name='Tests/images/flower.webp'>
  image = Image.open(file_path)
/home/erics/Pillow/Tests/test_file_webp_metadata.py:60: ResourceWarning: unclosed file <_io.BufferedReader name='Tests/images/flower2.webp'>
  image = Image.open(file_path)
/usr/lib/python3.5/contextlib.py:66: ResourceWarning: unclosed file <_io.FileIO name='Tests/images/hopper.ppm' mode='rb' closefd=True>
  next(self.gen)
/home/erics/vpy35-dbg/lib/python3.5/site-packages/Pillow-4.3.0-py3.5-linux-x86_64.egg/PIL/TiffImagePlugin.py:1053: ResourceWarning: unclosed file <_io.TextIOWrapper name='Tests/images/16bit.cropped.tif' mode='r' encoding='UTF-8'>
  return super(TiffImageFile, self).load()
[...]

PR #3261, opened 2018-07-20
Resolve __fp when closing and deleting

radarhere:
See #2716


hugovk:
@homm Does this fix https://github.com/python-pillow/Pillow/issues/2716?

radarhere:
Next question after that is - does this fix #2019?

radarhere:
This resolves most of #2786

Issue #3239, opened 2018-07-06
CI: Intermittent PyPy "error: release unlocked lock" on AppVeyor


hugovk:
Occasionally the PyPy build on AppVeyor errors. It passes after restarting the build. Here's the most recent five, all in the last five days. Of the last 19, 5 failed due to this, 13 passed, one was a valid fail.

<details>
  <summary>1</summary>

```
Tests/test_image_resample.py::TestImagingCoreResampleAccuracy::test_enlarge_bilinear ERROR [ 61%]
=================================== ERRORS ====================================
_ ERROR at teardown of TestImagingCoreResampleAccuracy.test_enlarge_bilinear __
self = <contextlib.GeneratorContextManager object at 0x077b5c70>, type = None
value = None, traceback = None
    def __exit__(self, type, value, traceback):
        if type is None:
            try:
>               self.gen.next()
c:\pypy2-v6.0.0-win32\lib-python\2.7\contextlib.py:24: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
c:\pypy2-v6.0.0-win32\lib-python\2.7\contextlib.py:24: in __exit__
    self.gen.next()
c:\pypy2-v6.0.0-win32\lib-python\2.7\logging\__init__.py:1331: in removeHandler
    _releaseLock()
c:\pypy2-v6.0.0-win32\lib-python\2.7\logging\__init__.py:232: in _releaseLock
    _lock.release()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
self = <_RLock owner=None count=0>
    def release(self):
        """Release a lock, decrementing the recursion level.
    
            If after the decrement it is zero, reset the lock to unlocked (not owned
            by any thread), and if any other threads are blocked waiting for the
            lock to become unlocked, allow exactly one of them to proceed. If after
            the decrement the recursion level is still nonzero, the lock remains
            locked and owned by the calling thread.
    
            Only call this method when the calling thread owns the lock. A
            RuntimeError is raised if this method is called when the lock is
            unlocked.
    
            There is no return value.
    
            """
        if self.__owner != _get_ident():
            raise RuntimeError("cannot release un-acquired lock")
        self.__count = count = self.__count - 1
        if not count:
[...]

PR #3272, opened 2018-07-26
Fix builds with --parallel

hsoft:
Python 3.5's distutils added support for parallel builds, which means
that we don't need to monkeypatch it anymore. But more importantly, this
monkeypatch made build fail (hang in fact) whenever `--parallel` was
passed to `python setup.py build`.

This commit fixes the problem by not applying the monkeypatch on
python 3.5+ and preserve the old behavior (parallel build by default) by
injecting a `parallel` option when it's not specified.

### DIFF
--- a/mp_compile.py
+++ b/mp_compile.py
@@ -1,5 +1,7 @@
 # A monkey patch of the base distutils.ccompiler to use parallel builds
 # Tested on 2.7, looks to be identical to 3.3.
+# Only applied on Python < 3.5 because otherwise, it conflicts with Python's
+# own newly-added support for parallel builds.
 
 from __future__ import print_function
 from multiprocessing import Pool, cpu_count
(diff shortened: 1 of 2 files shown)


hsoft:
Hum, the test failure on appveyor doesn't look related to my changes. Or could it be?

hugovk:
AppVeyor failure is probably https://github.com/python-pillow/Pillow/issues/3239. I've restated the build.

mp_compile.py
```diff
@@ -1,5 +1,7 @@
 # A monkey patch of the base distutils.ccompiler to use parallel builds
 # Tested on 2.7, looks to be identical to 3.3.
+# Only applied on python < 3.5 because otherwise, it conflicts with python's
+# own newly-added support for parallel builds.
```
radarhere: Minor style point: Python, not python

setup.py
```diff
@@ -205,6 +205,12 @@ def finalize_options(self):
         if self.debug:
             global DEBUG
             DEBUG = True
+        if sys.version_info >= (3, 5) and not self.parallel:
+            # For python < 3.5, we monkeypatch distutils to have parallel
```
radarhere: Minor style point: Python, not python

PR #3273, opened 2018-07-28
Changed Kernel to subclass BuiltinFilter

radarhere:
Changed Kernel to subclass BuiltinFilter, instead of the other way around.

Before, BuiltinFilter overrides Kernel's init method with a simple pass statement -
```
class Kernel(MultibandFilter):
    name = "Kernel"
    def __init__(self, size, kernel, scale=None, offset=0):
        ...
    def filter(self, image):
        ...

class BuiltinFilter(Kernel):
    def __init__(self):
        pass
```

After, Kernel adds an init method -
```
class BuiltinFilter(MultibandFilter):
    def filter(self, image):
        ...

class Kernel(BuiltinFilter):
    name = "Kernel"
    def __init__(self, size, kernel, scale=None, offset=0):
        ...
```

This should have no effect, it just seems like a nicer way to organise the code.

### DIFF
--- a/Tests/test_image_filter.py
+++ b/Tests/test_image_filter.py
@@ -94,14 +94,14 @@ class TestImageFilter(PillowTestCase):
         self.assertEqual(rankfilter.size, 1)
         self.assertEqual(rankfilter.rank, 2)
 
+    def test_builtinfilter_p(self):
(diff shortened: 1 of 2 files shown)


radarhere:
I have added tests to satisfy the coverage diff.

PR #3274, opened 2018-07-29
Improve PDF document info

radarhere:
Using https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/pdf_reference_archives/PDFReference.pdf as a reference, this PR suggests two changes -

- By default, set the 'title' field to the filename, minus the extension.
- Add creation and modification dates. Defaults to the current time. This can be read and written as a time.struct_time object

### DIFF
--- a/Tests/test_file_pdf.py
+++ b/Tests/test_file_pdf.py
@@ -4,6 +4,7 @@ import io
 import os
 import os.path
 import tempfile
+import time
 
 
 class TestFilePdf(PillowTestCase):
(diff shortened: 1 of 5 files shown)


radarhere:
To be clear, adding a default title is my own suggestion.

Adding the creation and modification dates was done using the PDF as a reference. Section 9.2.1 / Page 596 lists the fields as possible values in the Document Information Dictionary. Section 3.8.2 / Page 120 discusses the format of the date values.

PR #3282, opened 2018-08-04
Updated tested versions in installation docs

radarhere:
- Added 3.7 to list of OS X CI tested Python versions, since pillow-wheels now tests 3.7 - https://github.com/python-pillow/pillow-wheels/pull/90
- Added 9c to list of tested libjpeg version, which was added in #3136


hugovk:
Thanks!

PR #3283, opened 2018-08-04
Improved GIF documentation

### LINKED ISSUE #3255: `loop` parameter to `Image.save()` with `save_all=True` does not have correct default
### What did you do?

I called `im.save(filename, save_all=True, append_images=images)`

I expected to get a GIF image that looped forever, as per [the docs](https://pillow.readthedocs.io/en/latest/handbook/image-file-formats.html#saving-sequences).

### What actually happened?

I got a GIF image that played once and stopped.

4.3.0

### More comments.

The workaround is to call `im.save(filename, save_all=True, append_images=images, loop=0)`

The code causing this behavior is [this line](https://github.com/python-pillow/Pillow/blob/master/src/PIL/GifImagePlugin.py#L538
) in Pillow.

If you wanted to make the code correspond to the documentation, then [this](https://github.com/rec/Pillow/commit/e12d989f66d7c7928a384c957aed1733e0e30946) might be the sort of change you wanted and I could send this out as a pull request.

Or, the docs could change to match the code.

Given the workaround, this isn't any sort of blocking issue for me, but felt I should report it now I found it!  Thanks for an excellent library...

radarhere:
Resolves #3255 by improving documentation regarding the default value of 'loop' when saving. Also improves documentation for other options when reading and saving GIFs.

### DIFF
--- a/docs/handbook/image-file-formats.rst
+++ b/docs/handbook/image-file-formats.rst
@@ -84,7 +84,14 @@ The :py:meth:`~PIL.Image.Image.open` method sets the following
     of the GIF, in milliseconds.
 
 **loop**
-    May not be present. The number of times the GIF should loop.
+    May not be present. The number of times the GIF should loop. 0 means that
+    it will loop forever.
+
+**comment**
+    May not be present. A comment about the image.
+
+**extension**
+    May not be present. Contains application specific information.
 
 Reading sequences
 ~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


docs/handbook/image-file-formats.rst
```diff
@@ -146,6 +140,38 @@ are available::
...
+    the palette can be passed in as an
+    :py:class:`PIL.ImagePalette.ImagePalette` object.
+
+**optimize**
+    If present and true, attempt to compress the palette by
+    eliminating unused colors. This is only useful if the palette can
+    be compressed to the next smaller power of 2 elements.
+
+Note that if the image you are saving comes from an existing GIF, it may have
+the following properties in it's :py:attr:`~PIL.Image.Image.info` dictionary.
```
hugovk: "it's" -> "its"

radarhere: Thanks, okay, updated.

PR #3286, opened 2018-08-04
 Moved QApplication and QGuiApplication testing into separate threads

radarhere:
Using pytest to run all of the tests in a macOS 10.13 environment with PyQt5 installed, I find that test_imagetk.py exits with an 'Abort trap: 6'.

Investigating, I conclude that this is because QGuiApplication is not responding to the 'quit' method as it should, and then Tkinter dies when it runs at the same time.

For a simple example to demonstrate the problem, in running the following code, QGuiApplication does not quit, and I can't figure out how to persuade it to do so, short of exiting Python.

```
from PyQt5.QtGui import QGuiApplication
app = QGuiApplication([])
app.quit()
```

The simplest solution would be to rename the three test scripts that use QGuiApplication, so that they run after test_imagetk.py in the test suite. However, this is not very future proof. So instead, this PR uses multiprocessing to isolate QGuiApplication and QApplication.

### DIFF
--- a/Tests/test_image_fromqpixmap.py
+++ b/Tests/test_image_fromqpixmap.py
@@ -1,5 +1,5 @@
 from helper import unittest, PillowTestCase, hopper
-from test_imageqt import PillowQtTestCase, PillowQPixmapTestCase
+from test_imageqt import PillowQPixmapTestCase
 
 from PIL import ImageQt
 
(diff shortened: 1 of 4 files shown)


radarhere:
Closing in favour of #3288, as I find that with this change, Python is quitting unexpectedly.

PR #3288, opened 2018-08-05
Changed order of tests

radarhere:
Using pytest to run all of the tests in a macOS 10.13 environment with PyQt5 installed, I find that test_imagetk.py exits with an 'Abort trap: 6'.

Investigating, I conclude that this is because QGuiApplication is not responding to the 'quit' method as it should, and then Tkinter dies when it runs at the same time.

For a simple example to demonstrate the problem, in running the following code, QGuiApplication does not quit, and I can't figure out how to persuade it to do so, short of exiting Python.

```python
from PyQt5.QtGui import QGuiApplication
app = QGuiApplication([])
app.quit()
```

This PR renames the three test scripts that use QGuiApplication and QApplication, so that they run after test_imagetk.py in the test suite.


hugovk:
I can also reproduce on Mac after installing PyQt5 (`python3 -m pip install PyQt5`), with `python -m pytest -vx --cov PIL --cov-report term Tests`, and this PR fixes it.

PR #3294, opened 2018-08-06
Performance improvement of ImageDraw.floodfill() 

yo1995:
- Feature improvement: improved performance of `ImageDraw.floodfill()` with  Python built-in `set()` datatype. reduce run time while maintaining memory consumption.

 * under `ImageDraw Module`: `ImageDraw.floodfill()`
    - changed original *list* datatype to *set* in order to handle pixels in a more efficient manner.

## unittest result

pytest-result

## test result

test-result

by implementing this change, the floodfill() method is around 40% faster on both of the test cases. The result highly depends on the amount of the pixels to fill. The pixels to be filled are only iterated once, rather than 4 times with original method.

## test-case-files

[PR-testcases.zip](https://github.com/python-pillow/Pillow/files/2262254/PR-testcases.zip)

run test.sh to see comparison between methods.

### DIFF
--- a/src/PIL/ImageDraw.py
+++ b/src/PIL/ImageDraw.py
@@ -325,21 +325,21 @@ def getdraw(im=None, hints=None):
 
 def floodfill(image, xy, value, border=None, thresh=0):
     """
-        (experimental) Fills a bounded region with a given color.
+    (experimental) Fills a bounded region with a given color.
 
-        :param image: Target image.
-        :param xy: Seed position (a 2-item coordinate tuple). See
-            :ref:`coordinate-system`.
-        :param value: Fill color.
-        :param border: Optional border value.  If given, the region consists of
-            pixels with a color different from the border color.  If not given,
-            the region consists of pixels having the same color as the seed
-            pixel.
-        :param thresh: Optional threshold value which specifies a maximum
-            tolerable difference of a pixel value from the 'background' in
-            order for it to be replaced. Useful for filling regions of non-
-            homogeneous, but similar, colors.
-        """
+    :param image: Target image.
+    :param xy: Seed position (a 2-item coordinate tuple). See
+        :ref:`coordinate-system`.
+    :param value: Fill color.
+    :param border: Optional border value.  If given, the region consists of
+        pixels with a color different from the border color.  If not given,
+        the region consists of pixels having the same color as the seed
+        pixel.
+    :param thresh: Optional threshold value which specifies a maximum
+        tolerable difference of a pixel value from the 'background' in
+        order for it to be replaced. Useful for filling regions of non-
+        homogeneous, but similar, colors.
+    """
     # based on an implementation by Eric S. Raymond
     # amended by yo1995 @20180806
     pixel = image.load()
--- a/src/PIL/ImageDraw.py
+++ b/src/PIL/ImageDraw.py
@@ -352,7 +352,7 @@ def floodfill(image, xy, value, border=None, thresh=0):
     except (ValueError, IndexError):
         return  # seed point outside image
     edge = {(x, y)}
-    full_edge = set()  # use a set to record each unique pixel processed
+    full_edge = set()  # use a set to keep record of current and previous edge pixels to reduce memory consumption
     while edge:
         new_edge = set()
         for (x, y) in edge:  # 4 adjacent method
@@ -364,6 +364,7 @@ def floodfill(image, xy, value, border=None, thresh=0):
                 except (ValueError, IndexError):
                     pass
                 else:
+                    full_edge.add((s, t))
                     if border is None:
                         fill = _color_diff(p, background) <= thresh
                     else:
@@ -371,8 +372,7 @@ def floodfill(image, xy, value, border=None, thresh=0):
                     if fill:
                         pixel[s, t] = value
                         new_edge.add((s, t))
-                        full_edge.add((s, t))
-        full_edge = edge  # do not record useless pixels to reduce memory consumption
+        full_edge = edge  # discard pixels proceeded
         edge = new_edge
 
 
--- a/src/PIL/ImageDraw.py
+++ b/src/PIL/ImageDraw.py
@@ -388,7 +388,7 @@ def floodfill(image, xy, value, border=None, thresh=0):
                             new_edge.add((s, t))
                             full_edge.add((s, t))
             full_edge = edge
-            edge = newedge
+            edge = new_edge
 
 
 def _color_diff(rgb1, rgb2):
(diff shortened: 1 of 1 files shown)


yo1995:
first time to submit PR to larger project. if any other info is needed, please reply.
found this issue when using ImageDraw.floodfill() method and found it somehow slower than expectation. tried to improve from my own perspective with Python3 built-in type *set()*.

src/PIL/ImageDraw.py
```diff
@@ -350,34 +351,43 @@ def floodfill(image, xy, value, border=None, thresh=0):
...
                     except IndexError:
                         pass
                     else:
                         if p != value and p != border:
                             pixel[s, t] = value
-                            newedge.append((s, t))
+                            new_edge.add((s, t))
+                            full_edge.add((s, t))
+            full_edge = edge
             edge = newedge
```
hugovk: Thanks for the PR! Before we can check the performance improvements, the PR should be passing for both CIs: Travis CI for Linux and AppVeyor for Windows. Right now they are both failing because this `newedge` is not defined. Rename it to `new_edge` as well?

```console
$ flake8 src/PIL/ImageDraw.py
src/PIL/ImageDraw.py:372:80: E501 line too long (89 > 79 characters)
src/PIL/ImageDraw.py:391:20: F821 undefined name 'newedge'
```

(I wonder why your tests passed? Was it truly testing this modified code or another version?)

yo1995: sorry... a typo was found in the committed file 😓  I should just paste the tested version into it...

hugovk:
Here's what I get running this on macOS High Sierra with Python 3.7.0 and 2.7.15:
```bash
#!/usr/bin/env bash

set -o xtrace

python3 ff.py rect-test-1000.jpg
python3 ff.py rect-test-2500.jpg

python2 ff.py rect-test-1000.jpg
python2 ff.py rect-test-2500.jpg
```
```python
import time
import sys
import psutil
import os


if __name__ == '__main__':
    img_path = sys.argv[1]
    img1 = Image.open(img_path)
    t1 = time.time()
    ImageDraw.floodfill(img1, (1, 1), (0, 255, 0), None, 0)
    print('time used: ' + str(time.time() - t1))
    print('mem used: ' + str(psutil.Process(os.getpid()).memory_info().rss))
    prefix = "master"  # master or pr
    img1.save("{}-{}-{}.png".format(prefix, sys.version_info[0:3], img_path))
```


Time used | master | PR | PR / master
-- | -- | -- | --
python3 ff.py rect-test-1000.jpg | 3.05395722 | 1.6223352 | 0.53122394
python3 ff.py rect-test-2500.jpg | 18.1347847 | 11.3400948 | 0.62532283
python2 ff.py rect-test-1000.jpg | 2.03690004 | 1.62980819 | 0.80014147
python2 ff.py rect-test-2500.jpg | 14.1166871 | 11.3599179 | 0.80471557

Time used, shows PR uses x0.5-0.8 time


Mem used | master | PR |  PR / master
-- | -- | -- | --
python3 ff.py rect-test-1000.jpg | 15753216 | 16420864 | 1.0423817
python3 ff.py rect-test-2500.jpg | 37142528 | 37617664 | 1.01279224
python2 ff.py rect-test-1000.jpg | 22257664 | 22523904 | 1.01196172
python2 ff.py rect-test-2500.jpg | 43634688 | 44048384 | 1.0094809


Mem used, shows PR uses x1.00-1.04 memory

radarhere:
Hi. Thanks very much for this PR.

* `full_edge = edge  # do not record useless pixels to reduce memory consumption` - Could you explain this line in more detail?

    From what I can see, `full_edge = edge` followed by `edge = new_edge` means that `full_edge` becomes `edge` from the iteration of the while loop. Then, within the for loop, you’re adding in any new pixels that have been filled in.

    For example, running `ImageDraw.floodfill(im, (250, 250), ImageColor.getrgb("red"))` and printing `full_edge` each loop, the first two loops give -
    `{(250, 250)}`
    `{(249, 250), (251, 250), (250, 251), (250, 249)}` - no longer having (250, 250)

    So you’re deliberately abandoning all pixels but the current and previous edge in order to save memory, yes? If so, `full_edge` has a comment that it is ‘a set to record each unique pixel processed’, which isn't accurate.

* You’re appending to `full_edge` if the pixel is filled in. Would it not make more sense to append even if it is not filled in, to save checking the pixel twice? Or are you going to tell that the speed increase is not worth the memory?

yo1995:
- to first problem
Yeah that's exactly the thing I want to do. How about change the comment --
"use a set to record each unique pixel processed" -> "use a set to keep record of current and previous edge pixels to reduce memory consumption"
"do not record useless pixels to reduce memory consumption" -> "discard pixels proceeded"

- to second problem
At first sight I think your idea is better. I tested the idea you've described and found that the run-time reduced by around 0.1~0.3s in the previous test cases and two newly added ones. The numbers of iterations are overall reduced, though they highly depends on the distribution of pixels to fill. 

One of the cases that I found it slower than my commit is the bmp below in .zip.I guess it's something to do with the "in" and "add" methods of Python Set objects. Still, the compromise is minor, in my test ~0.17s for this picture.
[maze.zip](https://github.com/python-pillow/Pillow/files/2342572/maze.zip)

Below attached the test cases for this issue. If it's OK, I'll revise the comments and change the code to your described method, then commit again. 😄 
[PR-testcases-2-1.zip](https://github.com/python-pillow/Pillow/files/2342585/PR-testcases-2-1.zip)

radarhere:
Okay. Thanks for your investigation. I mean, the problem with this, as with many parts of an image processing library I imagine, is that there is such a wide variety of use cases, and no way of knowing what is more typical.

So yes, do whatever you conclude is best, feel free to take or ignore my suggestion as you see fit.

src/PIL/ImageDraw.py
```diff
@@ -364,15 +364,15 @@ def floodfill(image, xy, value, border=None, thresh=0):
...
                     if border is None:
                         fill = _color_diff(p, background) <= thresh
                     else:
                         fill = p != value and p != border
                     if fill:
                         pixel[s, t] = value
                         new_edge.add((s, t))
-                        full_edge.add((s, t))
-        full_edge = edge  # do not record useless pixels to reduce memory consumption
+        full_edge = edge  # discard pixels proceeded
```
radarhere: Typo? processed, not proceeded

yo1995: 🤦‍♂️ just fixed it. blindly copy-paste from my own comment without double check...

yo1995: hmmm... strange, editted comment led to Build execution time exceeded. Maybe I should add a tag in commit message to skip the build test. What's tag should I use?

hugovk:
I restated it and it's green now.

src/PIL/ImageDraw.py
radarhere: Thanks. In case you're interested for the future, to skip the tests, include ' [ci skip]' in the commit message.

PR #3298, opened 2018-08-09
Read/Save RGB webp as RGB (instead of RGBX)

kkopachev:
Issue being fixed: 
Animated WebP [support added in 5.0](https://github.com/python-pillow/Pillow/pull/2761) made RGB WebP images being read as RGBX. From my understanding, there is no difference in storage for RGB vs RGBX, but having mode=RGBX makes it requirement to convert image to RGB before saving to other formats, like png. It throws `cannot write mode RGBX as PNG`. This behavior change was not mentioned in [release notes](https://pillow.readthedocs.io/en/latest/releasenotes/5.0.0.html), so I assume this can be declared BC break.

see https://github.com/thumbor/thumbor/issues/1099 for example bug.

 * Read RGB WebP images as RGB instead of RGBX
 * On WebP save, recognize more transparency modes


src/PIL/WebPImagePlugin.py
```diff
@@ -288,7 +297,8 @@ def _save(im, fp, filename):
     xmp = im.encoderinfo.get("xmp", "")
 
     if im.mode not in _VALID_WEBP_LEGACY_MODES:
-        alpha = im.mode == 'P' and 'A' in im.im.getpalettemode()
+        alpha = 'A' in im.mode or 'a' in im.mode \
+                or im.mode == 'P' and 'A' in im.im.getpalettemode()
```
radarhere: While this is presumably what you meant to infer with the line break, could I interest you in adding brackets?
```
alpha = 'A' in im.mode or 'a' in im.mode \
         or (im.mode == 'P' and 'A' in im.im.getpalettemode())
```
Maybe it's just me, but I didn't instinctively know what `bool or bool and bool` was going to result in.

src/PIL/WebPImagePlugin.py
```diff
@@ -240,16 +242,23 @@ def _save_all(im, fp, filename):
 
                 # Make sure image mode is supported
                 frame = ims
+                rawmode = ims.mode
                 if ims.mode not in _VALID_WEBP_MODES:
-                    alpha = ims.mode == 'P' and 'A' in ims.im.getpalettemode()
-                    frame = ims.convert('RGBA' if alpha else 'RGBX')
+                    alpha = 'A' in ims.mode or 'a' in ims.mode \
+                            or (ims.mode == 'P' and 'A' in ims.im.getpalettemode())
+                    rawmode = 'RGBA' if alpha else 'RGBX'
```
homm: Is it possible to use `RGB` instead? I'm not sure, probably `ims.convert('RGB`)` has more wide range of supported source modes, than `ims.convert('RGBX`)`. There should no be difference in other aspects.

kkopachev: That seems to make sense. I made the change. Few lines later, when we exporting to bytes it still uses `RGBX` rawmode.

fxuls:
I'm getting the same error when I try saving a webp Image object as a png. Is there a way for me to get this fix now or do I have to wait until the next update so I can update Pillow through pip? Sorry I'm still learning git and I've just relied on pip to manage all my dependencies up until now.

hugovk:
@fxuls This hasn't been merged or released yet, so for now you would need to build this from source using https://github.com/kkopachev/Pillow/tree/webp-rgb.

Follow https://pillow.readthedocs.io/en/5.2.x/installation.html#building-from-source but instead of "Download and extract the compressed archive from PyPI" do this:

```sh
git clone https://github.com/kkopachev/Pillow/tree/webp-rgb
cd Pillow
```

hugovk:
@fxuls Update: This has been merged and is due for release on 2018-10-01. If you want to test now,  build from source using `git clone https://github.com/python-pillow/Pillow`

Issue #3284, opened 2018-08-04
CI: AppVeyor builds failing on PyPy


hugovk:
### What did you do?

Do an AppVeyor Windows build on PyPy.

### What did you expect to happen?

Pass, as it did 6 days:

* https://ci.appveyor.com/project/Python-pillow/pillow/build/4029/job/9dsqymki6r0ydy6h

### What actually happened?

It failed. Here's that same PR rebuilt:

* https://ci.appveyor.com/project/Python-pillow/pillow/build/4040/job/jkogih4sp9nq3yut

```
Collecting coverage>=3.7.1 (from pytest-cov)
Collecting scandir; python_version < "3.5" (from pathlib2>=2.2.0; python_version < "3.6"->pytest)
  Downloading https://files.pythonhosted.org/packages/50/a4/141939a8d213b2cf1b1d6b2704c6f6c07d4f1903df532af887c8e64e6815/scandir-1.8.tar.gz
Building wheels for collected packages: scandir
  Running setup.py bdist_wheel for scandir: started
  Running setup.py bdist_wheel for scandir: finished with status 'error'
  Complete output from command c:\vp\pypy2\bin\pypy.exe -u -c "import setuptools, tokenize;__file__='c:\\users\\appveyor\\appdata\\local\\temp\\1\\pip-install-tb8im8\\scandir\\setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" bdist_wheel -d c:\users\appveyor\appdata\local\temp\1\pip-wheel-7oh920 --python-tag pp260:
  c:\pypy2-v6.0.0-win32\lib-python\2.7\distutils\extension.py:133: UserWarning: Unknown Extension options: 'optional'
    warnings.warn(msg)
  running bdist_wheel
  running build
  running build_py
  creating build
  creating build\lib.win32-2.7
  copying scandir.py -> build\lib.win32-2.7
  running build_ext
  building '_scandir' extension
  creating build\temp.win32-2.7
  creating build\temp.win32-2.7\Release
  C:\Users\appveyor\AppData\Local\Programs\Common\Microsoft\Visual C++ for Python\9.0\VC\Bin\cl.exe /c /nologo /Ox /MD /W3 /GS- /DNDEBUG -Ic:\vp\pypy2\include /Tc_scandir.c /Fobuild\temp.win32-2.7\Release\_scandir.obj
[...]

PR #3299, opened 2018-08-10
Revert "CI: Temporarily pin pytest to prevent scandir non-compilation"

hugovk:
Reverts python-pillow/Pillow#3285

Fixes https://github.com/python-pillow/Pillow/issues/3284

Re: https://github.com/pytest-dev/pytest/issues/3750#issuecomment-412093501


radarhere:
Thanks

Issue #3312, opened 2018-08-23
Values passed to bomb_decompression_check in _crop are not correct


dinkolubina:
### What did you do?
 im = Image.new("RGB", (100, 100))
 im.crop((-9990, -9990, 9999, 9999))

### What did you expect to happen?
a crop to be preformed

### What actually happened?
>>> i.crop((-9990, -9990, 9999, 9999))
/home/dinko/anaconda3/envs/klarnai/lib/python3.6/site-packages/PIL/Image.py:2546: DecompressionBombWarning: Image size (99980001 pixels) exceeds limit of 89478485 pixels, could be decompression bomb DOS attack.
  DecompressionBombWarning)
<PIL.Image.Image image mode=RGB size=19989x19989 at 0x7F7E00163D68>

_bomb_decompression_check checks if the size of the image is larger than a certain value.
In Image._crop which is called from Image.crop x1 and y1 are passed to check the size of the crop.
abs(x1 - x0) and abs(y1-y0) should be passed instead, since that represents the size of the crop not just x1 and y1.

### What versions of Pillow and Python are you using?
python 3.6.5

The issue is easily reproducible if you pass values that go over the MAX_IMAGE_PIXELS, we tried to cut something that had a minus, although the cropped rectangle area is under the MAX_IMAGE_PIXELS, we saw that only x1 and y1 were passed, which represent the lower right location, but not the actual width and height

PR #3313, opened 2018-08-23
Fix _crop and tests

dinkolubina:
Issue being fixed:
DecompressionBombWarning/Error caused by wrong values passed to it in Image._crop

_bomb_decompression_check checks if the size of the image is larger than a certain value.
In Image._crop which is called from Image.crop x1 and y1 are passed to check the size of the crop.
abs(x1 - x0) and abs(y1-y0) should be passed instead, since that represents the size of the crop not just x1 and y1.

* Use absolute value of (x1-x0) and (y1-y0) instead of the existing check
* Pas those values which represent the width and height instead just the location of x1, y1, which is not correct in the context of _decompression_bomb_check

Pull request is solving this issue: https://github.com/python-pillow/Pillow/issues/3312

### DIFF
--- a/Tests/test_decompression_bomb.py
+++ b/Tests/test_decompression_bomb.py
@@ -6,7 +6,6 @@ TEST_FILE = "Tests/images/hopper.ppm"
 
 ORIGINAL_LIMIT = Image.MAX_IMAGE_PIXELS
 
-from PIL.Image import DecompressionBombWarning, DecompressionBombError
 
 class TestDecompressionBomb(PillowTestCase):
 
@@ -67,7 +66,7 @@ class TestDecompressionCrop(PillowTestCase):
         im = Image.new("RGB", (100, 100))
 
         good_values = ((-9999, -9999, -9990, -9990),
-                        (-999, -999, -990, -990))
+                       (-999, -999, -990, -990))
 
         warning_values = ((-160, -160, 99, 99),
                           (160, 160, -99, -99))
@@ -76,14 +75,15 @@ class TestDecompressionCrop(PillowTestCase):
                         (99909, 99990, -99999, -99999))
 
         for value in good_values:
-            self.assertEqual(im.crop(value).size, (9,9))
+            self.assertEqual(im.crop(value).size, (9, 9))
 
         for value in warning_values:
-            self.assert_warning(DecompressionBombWarning, im.crop, value)
+            self.assert_warning(Image.DecompressionBombWarning, im.crop, value)
 
         for value in error_values:
-            with self.assertRaises(DecompressionBombError):
+            with self.assertRaises(Image.DecompressionBombError):
                 im.crop(value)
+
 
 if __name__ == '__main__':
     unittest.main()
--- a/Tests/test_decompression_bomb.py
+++ b/Tests/test_decompression_bomb.py
@@ -6,6 +6,7 @@ TEST_FILE = "Tests/images/hopper.ppm"
 
 ORIGINAL_LIMIT = Image.MAX_IMAGE_PIXELS
 
+from PIL.Image import DecompressionBombWarning, DecompressionBombError
 
 class TestDecompressionBomb(PillowTestCase):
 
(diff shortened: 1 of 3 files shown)


Tests/test_image_crop.py
```diff
@@ -103,6 +103,30 @@ def test_crop_zero(self):
         self.assertEqual(cropped.size, (10, 10))
         self.assertEqual(cropped.getdata()[2], (0, 0, 0))
 
+    def test_crop_decompression_checks(self):
```
hugovk: Maybe this test case would be better in `Tests/test_decompression_bomb.py`?

src/PIL/Image.py
```diff
@@ -1104,12 +1104,9 @@ def _crop(self, im, box):
 
         x0, y0, x1, y1 = map(int, map(round, box))
 
-        if x1 < x0:
-            x1 = x0
-        if y1 < y0:
-            y1 = y0
+        absolute_values = (abs(x1-x0), abs(y1-y0))
```
hugovk: The outer parentheses can be removed (and spaces added around the minuses).

Tests/test_decompression_bomb.py
```diff
@@ -61,6 +62,28 @@ def testEnlargeCrop(self):
         self.assert_warning(Image.DecompressionBombWarning,
                             self.src.crop, box)
 
+    def test_crop_decompression_checks(self):
```
dinkolubina: Moved the test to test_decompression_bomb.py

src/PIL/Image.py
```diff
@@ -1104,12 +1104,9 @@ def _crop(self, im, box):
 
         x0, y0, x1, y1 = map(int, map(round, box))
 
-        if x1 < x0:
-            x1 = x0
-        if y1 < y0:
-            y1 = y0
+        absolute_values = (abs(x1 - x0), abs(y1 - y0))
```
dinkolubina: left the parentheses since _decompression_bomb_check expects a tuple (Image.size), added spaces around minuses

Tests/test_decompression_bomb.py
```diff
@@ -61,6 +62,28 @@ def testEnlargeCrop(self):
...
+                          (160, 160, -99, -99))
+
+        error_values = ((-99909, -99990, 99999, 99999),
+                        (99909, 99990, -99999, -99999))
+
+        for value in good_values:
+            self.assertEqual(im.crop(value).size, (9,9))
+
+        for value in warning_values:
+            self.assert_warning(DecompressionBombWarning, im.crop, value)
```
dinkolubina: Changed to `assert_warning` instead `assertWarns` as per instructions to stop the test from failing in Python 2.

Tests/test_decompression_bomb.py
```diff
@@ -61,6 +62,28 @@ def testEnlargeCrop(self):
...
+
+        for value in error_values:
+            with self.assertRaises(DecompressionBombError):
```
dinkolubina: Changed per instructions (moved the assertRaises in the for loop)

dinkolubina:
I rebased after changing the code. Didn't know that the comments will be deleted, sorry about that. I have addressed the comments individually in the recent commit.

src/PIL/Image.py
hugovk: It's a tuple with or without parentheses:

```
⌂86% [hugo:~/github/pypistats] master* ± p
Python 3.7.0 (default, Jun 29 2018, 20:13:13)
[Clang 9.1.0 (clang-902.0.39.2)] on darwin
>>> absolute_values = (1, 2)
>>> type(absolute_values)
<class 'tuple'>
>>> absolute_values = 1, 2
>>> type(absolute_values)
<class 'tuple'>
>>>
⌂140% [hugo:~/github/pypistats] master* 21s ± p2
Python 2.7.15 (default, Jun 17 2018, 12:46:58)
[GCC 4.2.1 Compatible Apple LLVM 9.1.0 (clang-902.0.39.2)] on darwin
>>> absolute_values = (1, 2)
>>> type(absolute_values)
<type 'tuple'>
>>> absolute_values = 1, 2
>>> type(absolute_values)
<type 'tuple'>
>>>
```

dinkolubina: Thanks for the suggestion, didn't know that.

PR #3314, opened 2018-08-23
Fix None value of comment

### LINKED ISSUE #3315: TypeError: object of type 'NoneType' has no len()
### TypeError: object of type 'NoneType' has no len()

This image has a value "comment" is None. It raises that error.  

This is my PR 
https://github.com/python-pillow/Pillow/pull/3314

https://github.com/python-pillow/Pillow/blob/master/src/PIL/GifImagePlugin.py#L527
```python
>    if "comment" in im.encoderinfo and 1 <= len(im.encoderinfo["comment"]) <= 255:
E                  TypeError: object of type 'NoneType' has no len()

```

https://github.com/python-pillow/Pillow/blob/master/src/PIL/GifImagePlugin.py#L697

```python
    def _get_global_header(im, info):
        """Return a list of strings representing a GIF header"""
    
        # Header Block
        # http://www.matthewflickinger.com/lab/whatsinagif/bits_and_bytes.asp
    
        version = b"87a"
                if ((extensionKey == "duration" and info[extensionKey] == 0) or
>                  (extensionKey == "comment" and not (1 <= len(info[extensionKey]) <= 255))):
E                  TypeError: object of type 'NoneType' has no len()
```

```python
{'background': 160,
 'comment': None,
 'duration': 500,
 'optimize': True,
 'version': b'GIF89a'}
```

This code is reproducing that issue
```python
from PIL import Image
from io import BytesIO

filename = 'test.gif'


with open(filename, "rb") as fp:

    img = Image.open(fp)
    file_obj = BytesIO()

    print(img.info)

    getattr(img, "is_animated", False)

    print(img.info)

    img.save(file_obj, format="GIF")
```
test2

tytarenko:
Fixes #3315.

### DIFF
--- a/src/PIL/GifImagePlugin.py
+++ b/src/PIL/GifImagePlugin.py
@@ -528,8 +528,8 @@ def _write_local_header(fp, im, offset, flags):
                  o8(transparency) +       # transparency index
                  o8(0))
 
-    if "comment" in im.encoderinfo and \
-       and im.encoderinfo["comment"] is not None \
+    if "comment" in im.encoderinfo \
+        and im.encoderinfo["comment"] is not None \
         and 1 <= len(im.encoderinfo["comment"]) <= 255:
         fp.write(b"!" +
                  o8(254) +                # extension intro
--- a/src/PIL/GifImagePlugin.py
+++ b/src/PIL/GifImagePlugin.py
@@ -529,7 +529,8 @@ def _write_local_header(fp, im, offset, flags):
                  o8(0))
 
     if "comment" in im.encoderinfo and \
-       1 <= len(im.encoderinfo["comment"]) <= 255:
+       and im.encoderinfo["comment"] is not None \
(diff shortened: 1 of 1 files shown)


src/PIL/GifImagePlugin.py
```diff
@@ -529,7 +529,8 @@ def _write_local_header(fp, im, offset, flags):
                  o8(0))
 
     if "comment" in im.encoderinfo and \
-       1 <= len(im.encoderinfo["comment"]) <= 255:
+       and im.encoderinfo["comment"] is not None \
```
radarhere: If you look at these two lines, you have two `and`s in a row.

radarhere: Thanks

hugovk:
Thanks for the report and PR. Closing this, #3315 fixed via https://github.com/python-pillow/Pillow/pull/3324.

PR #3320, opened 2018-08-28
Improved pillow-wheels RELEASING instructions

radarhere:
* Only update the Pillow submodule of pillow-wheels, rather than also updating multibuild
* Add the modified submodule before commiting


hugovk:
Thanks!

PR #3327, opened 2018-09-01
Added orientation, compression and id_section as TGA save keyword arguments

### LINKED ISSUE #3226: Sanitize TGA Image.info
Currently, the TGA image plugin sets `Image.info` "compression" to "tga_rle" on `load()`, while `save()` takes "rle" (bool) option. Both methods should use the same option for consistency. It probably makes sense to keep "rle" as TGA format doesn't support other compression methods, but "compression" may be more consistent with BMP and TIFF plugins. Neither of the options is documented, so there is no danger of breaking backward compatibility.

Also, it's not very clear whether `save()`  method should "inherit" info like TIFF and PNG plugins do:

https://github.com/python-pillow/Pillow/blob/4407cb65079a7d1150277e3b9a144996f56357c9/src/PIL/TiffImagePlugin.py#L1399-L1400

Currently, TGA plugin only inherits "orientation" but doesn't allow to specify it as a keyword to `save()`, and "id_section" is ignored altogether.

radarhere:
Resolves #3226

### DIFF
--- a/src/PIL/TgaImagePlugin.py
+++ b/src/PIL/TgaImagePlugin.py
@@ -20,6 +20,8 @@
 from . import Image, ImageFile, ImagePalette
 from ._binary import i8, i16le as i16, o8, o16le as o16
 
+import warnings
+
 __version__ = "0.3"
 
 
@@ -163,6 +165,10 @@ def _save(im, fp, filename):
     id_section = im.encoderinfo.get("id_section",
                                     im.info.get("id_section", ""))
     idlen = len(id_section)
+    if idlen > 255:
+        idlen = 255
+        id_section = id_section[:255]
+        warnings.warn("id_section has been trimmed to 255 characters")
 
     if colormaptype:
         colormapfirst, colormaplength, colormapentry = 0, 256, 24
--- a/src/PIL/TgaImagePlugin.py
+++ b/src/PIL/TgaImagePlugin.py
@@ -55,7 +55,7 @@ class TgaImageFile(ImageFile.ImageFile):
         # process header
         s = self.fp.read(18)
 
-        idlen = i8(s[0])
+        id_len = i8(s[0])
 
         colormaptype = i8(s[1])
         imagetype = i8(s[2])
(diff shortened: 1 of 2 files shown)


src/PIL/TgaImagePlugin.py
```diff
@@ -166,11 +174,12 @@ def _save(im, fp, filename):
...
         flags = 0
 
-    orientation = im.info.get("orientation", -1)
+    orientation = im.encoderinfo.get("orientation",
+                                     im.info.get("orientation", -1))
     if orientation > 0:
         flags = flags | 0x20
 
-    fp.write(b"\000" +
+    fp.write(o8(idlen) +
```
danpla: There is a small problem: o8 is `bytes((i & 255,))`, which will not raise an exception like `struct.pack()`, making it possible for `fp.write()` below to write `id_section` bigger than 255 bytes, resulting in a broken image.

radarhere: Okay, I've added a commit to trim `id_section`.

src/PIL/TgaImagePlugin.py
```diff
@@ -151,11 +151,19 @@ def _save(im, fp, filename):
     except KeyError:
         raise IOError("cannot write mode %s as TGA" % im.mode)
 
-    rle = im.encoderinfo.get("rle", False)
-
+    if "rle" in im.encoderinfo:
```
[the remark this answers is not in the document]
radarhere: Yeah. I'm just not keen to break the backwards compatibility of this one. If this PR is merged, you're welcome to submit a separate PR for that.

src/PIL/TgaImagePlugin.py
```diff
@@ -151,11 +153,23 @@ def _save(im, fp, filename):
...
+                                         im.info.get("compression"))
+        rle = compression == "tga_rle"
     if rle:
         imagetype += 8
 
+    id_section = im.encoderinfo.get("id_section",
+                                    im.info.get("id_section", ""))
+    idlen = len(id_section)
+    if idlen > 255:
+        idlen = 255
```
hugovk: Let's `snake_case` this variable.

src/PIL/TgaImagePlugin.py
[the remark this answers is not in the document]
hugovk: Sounds sensible to keep it, we've no idea what code might be using `im.encoderinfo["rle"]`.

We'd need deprecation warnings first before removing.

src/PIL/TgaImagePlugin.py
radarhere: Okay, I've added a commit. Note that the commit also changes the `idlen` variable in `_open`, for consistency.

PR #3338, opened 2018-09-07
Convert int values of RATIONAL TIFF tags to floats

### LINKED ISSUE #1765: Save compressed TIFF only accepts float values for dpi
There is a trap in setting the resolution for compressed TIFF images.  

```python
# pillow v3.1.1

import PIL.Image
im = PIL.Image.new('CMYK', (100, 100))

dpi = 100.0  # works
dpi = 100  # fails

im.save('test.tif', format='TIFF',
        dpi=(dpi, dpi),
        compression='tiff_lzw',
        #compression='tiff_deflate',
        #compression='tiff_adobe_deflate'
        )
```

Integers fail silently resulting in the resolution value missing from the saved file. Floats work as expected.

radarhere:
Resolves #1765

#2288 is a PR to significantly improve writing libtiff metadata. This PR is a quick fix for #1765, if we would like to resolve that issue in the short term.

### DIFF
--- a/Tests/test_file_libtiff.py
+++ b/Tests/test_file_libtiff.py
@@ -231,6 +231,16 @@ class TestFileLibTiff(LibTiffTestCase):
 
         TiffImagePlugin.WRITE_LIBTIFF = False
 
+    def test_int_dpi(self):
+        # issue #1765
+        im = hopper('RGB')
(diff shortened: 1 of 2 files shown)


hugovk:
Thanks.

PR #3355, opened 2018-09-15
Removed unnecessary line

radarhere:


homm:
Is there is a chance this could be exported from setup.py?

radarhere:
`feature` is a property of the `pil_build_ext` class in setup.py. There’s no point at which the parent `build_ext` class has `feature` assigned to it, and there’s no method where it is returned.

Nothing else in Pillow imports `pil_build_ext` either.

PR #3359, opened 2018-09-17
Give correct extrema for I;16 format images

bz2:
Currently gives None for a 16 bit greyscale image rather than the true min and max values in the 0-65536 range. The internal `ImagingGetProjection` function already supports `I;16` but the `_getextrema` needs to know to unpack the result.


bz2:
The appveyor failure looks spurious to me?

hugovk:
Yes, it's unrelated, see https://github.com/python-pillow/Pillow/issues/3363.

hugovk:
https://github.com/python-pillow/Pillow/issues/3363 has been fixed, you can rebase/merge master if you'd like to include it.

bz2:
@hugovk @radarhere Thanks! Rebase done, gating rerunning now.

Issue #3236, opened 2018-07-05
Release Pillow 5.3.0 on October 1, 2018


aclark4life:
# [Release Checklist](https://github.com/python-pillow/Pillow/blob/master/RELEASING.md)

## Main Release

* [x] Open a release ticket e.g. https://github.com/python-pillow/Pillow/issues/3236
* [x] Develop and prepare release in ``master`` branch.
* [x] Check [Travis CI](https://travis-ci.org/python-pillow/Pillow) and [AppVeyor CI](https://ci.appveyor.com/project/python-pillow/Pillow) to confirm passing tests in ``master`` branch.
* [x] Check that all of the wheel builds [Pillow Wheel Builder](https://github.com/python-pillow/pillow-wheels) pass the tests in Travis CI.
* [x] In compliance with [PEP 440](https://www.python.org/dev/peps/pep-0440/), update version identifier in `src/PIL/_version.py`
* [x] Update `CHANGES.rst`.
* [x] Run pre-release check via `make release-test` in a freshly cloned repo.
* [x] Create branch and tag for release e.g.:
```
    $ git branch 5.3.x
    $ git tag 5.3.0
    $ git push --all
    $ git push --tags
```
* [x] Create source distributions e.g.:
```
    $ make sdist
```
* [x] Create [binary distributions](#binary-distributions)
* [x] Upload all binaries and source distributions e.g. ``twine upload dist/Pillow-5.3.0-*``
* [x] Create a [new release on GitHub](https://github.com/python-pillow/Pillow/releases/new)
* [x] In compliance with [PEP 440](https://www.python.org/dev/peps/pep-0440/), append `.dev0` to version identifier in `src/PIL/_version.py`

### Windows
* [x] Contact @cgohlke for Windows binaries via release ticket e.g. https://github.com/python-pillow/Pillow/issues/1174.
* [x] Download and extract tarball from @cgohlke and ``twine upload *``.

### Mac and Linux
* [x] Use the [Pillow Wheel Builder](https://github.com/python-pillow/pillow-wheels):
```
    $ git clone https://github.com/python-pillow/pillow-wheels
    $ cd pillow-wheels
    $ git submodule init
    $ git submodule update
    $ cd Pillow
    $ git fetch --all
    $ git checkout [[release tag]]
    $ cd ..
    $ git commit -m "Pillow -> 5.3.0" Pillow
    $ git push
```
[...]

PR #3364, opened 2018-09-18
Added ImageOps pad method

### LINKED ISSUE #3339: Suggestion: option to pad instead of crop in ImageOps.fit
I recently wanted to use Pillow to generate thumbnails of a directory of images. I wanted each thumbnail to be exactly 300x200, but preserve the aspect ratio of the original image by padding one of the dimensions.

I thought [`PIL.ImageOps.fit`](https://pillow.readthedocs.io/en/latest/reference/ImageOps.html?highlight=ImageOps#PIL.ImageOps.fit) might do exactly what I wanted, but it crops rather than padding.

Would there be interest in an option to pad rather than crop? Or another method? I'd be happy to contribute if so.

PIL 1.1.7, Python 3.6.5

radarhere:
Resolves #3339

Added `contain` to ImageOps, with the arguments and docstring based on the `fit` method. It expands a given image until it matches at least one of the dimensions of a provided size, and fills in the remaining space with a provided color.


danvk:
This looks great @radarhere. Thank you!

hugovk:
> There are only two hard things in Computer Science: cache invalidation and naming things.

I wonder if there's a better name than `contain`?

Perhaps `pad`?

At least "pad image" is a common result on Stack Overflow and Google.

radarhere:
Okay, I've renamed it.

hugovk:
I've added this to the release notes TODO on https://github.com/python-pillow/Pillow/issues/3236.

PR #3367, opened 2018-09-19
CI: Retry SVN checkout to mitigate temporary network problems

hugovk:
Here's a recent one-off build failure on Travis CI during `svn checkout` in the installation phase:

```
svn: E175002: Unexpected HTTP status 429 'Too Many Requests' on '/python-pillow/pillow-depends/!svn/vcc/default'
cp: cannot stat 'test_images/*': No such file or directory
The command "if [ "$DOCKER" == "" ]; then .travis/install.sh; fi" failed and exited with 1 during .
```
https://travis-ci.org/python-pillow/Pillow/jobs/430241230#L2608

Let's make the build a bit more resilient by retrying up to 3 times using `travis_retry`:

https://docs.travis-ci.com/user/common-build-problems/#travis_retry

It's possible subsequent attempts right after a `429 'Too Many Requests'` won't work, but let's give this a shot. This should help with other temporary network problems.

---

Checking other network installation things, `wget` already defaults to 20 retries:

https://www.gnu.org/software/wget/manual/wget.html#Download-Options

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,7 @@ services:
   - docker
 
 install:
-  - if [ "$DOCKER" == "" ]; then .travis/install.sh; fi
+  - if [ "$DOCKER" == "" ]; then travis_retry .travis/install.sh; fi
 
 before_install:
   - if [ "$DOCKER" ]; then travis_retry docker pull pythonpillow/$DOCKER:$DOCKER_TAG; fi
--- a/depends/install_extra_test_images.sh
+++ b/depends/install_extra_test_images.sh
@@ -3,7 +3,7 @@
 
 rm -rf test_images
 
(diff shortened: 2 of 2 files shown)


hugovk:
Oh, that didn't work, `travis_retry` isn't available in subshells or scripts. The last commit moves it up to `.travis.yml`.

Upside: it covers the `git clone` commands in there too, and any other temporary failures that may occur.

Downside: it can take a couple of minutes to run `install.sh`, if install fails for a valid reason (eg. a PR breaks some install), it'll take longer to re-run and fail.

Maybe it's not worth it, I've not seen any `svn checkout` errors before.

hugovk:
Closing in favour of https://github.com/python-pillow/Pillow/pull/3368.

PR #3376, opened 2018-09-26
Catch CalledProcessError when checking for Ghostscript

radarhere:
`CalledProcessError` is thrown in https://travis-ci.org/python-pillow/docker-images/jobs/421960744#L2136. This PR catches it.

### DIFF
--- a/src/PIL/EpsImagePlugin.py
+++ b/src/PIL/EpsImagePlugin.py
@@ -61,7 +61,7 @@ def has_ghostscript():
             with open(os.devnull, 'wb') as devnull:
                 subprocess.check_call(['gs', '--version'], stdout=devnull)
             return True
-        except OSError:
+        except (OSError, subprocess.CalledProcessError):
             # no ghostscript
             pass
     return False


hugovk:
In this case, `gs` is found, but it threw an error:

```
gs: /usr/lib/libc.so.6: version `GLIBC_2.28' not found (required by /usr/lib/libgs.so.9)
```

https://superuser.com/a/537694/83235 says of a similar error:

> ```/lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.14' not found```
>
> That means the program was compiled against glibc version 2.14, and it requires that version to run, but your system has an older version installed.  You'll need to either recompile the program against the version of glibc that's on your system, or install a newer version of glibc (the "libc6" package in Debian).

glibc 2.28 is [newly released on 2018-08-01](https://www.gnu.org/software/libc/).

The ghostscript we have must have been compiled against glibc 2.28.

radarhere:
Okay, thanks.

Just for reference, `has_ghostscript` is a method that is not called by anything else in the main library, only from tests - this is what it was added for in #601 - so as far as our CI testing is concerned, you're right, we should throw and track down this error, but if a user wanted to use that method, it's not perfectly obvious to me what should happen. They have ghostscript, but it's not a working version of ghostscript. Should that return `False`, or report the error?

Closing, as I think it's better to report the error.

Issue #2949, opened 2018-01-11
TypeError occurred when converting transparent WebP to transparent GIF


blueset:
### What did you do?

```python
from PIL import Image
from io import BytesIO

image = Image.open('test.webp')
destination = BytesIO()
image.save(destination, 'gif', transparency=0)
```

### What did you expect to happen?

Transparent WebP image is converted to transparent GIF, and stored in `destination`.

### What actually happened?

```python
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.6/site-packages/PIL/Image.py", line 1930, in save
    save_handler(self, fp, filename)
  File "/usr/local/lib/python3.6/site-packages/PIL/GifImagePlugin.py", line 468, in _save
    _write_single_frame(im, fp, palette)
  File "/usr/local/lib/python3.6/site-packages/PIL/GifImagePlugin.py", line 376, in _write_single_frame
    for s in _get_global_header(im_out, im.encoderinfo):
  File "/usr/local/lib/python3.6/site-packages/PIL/GifImagePlugin.py", line 715, in _get_global_header
    o8(background) + o8(0),
  File "/usr/local/lib/python3.6/site-packages/PIL/_binary.py", line 27, in o8
    return bytes((i & 255,))
TypeError: unsupported operand type(s) for &: 'tuple' and 'int'
```

### What versions of Pillow and Python are you using?

```python
>>> PIL.VERSION
'1.1.7'
>>> PIL.PILLOW_VERSION
'5.0.0'
>>> sys.version
'3.6.4 (default, Dec 19 2017, 15:26:29) \n[GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.39.2)]'
```

```shell
$ brew info webp
webp: stable 0.6.1 (bottled), HEAD
Image format providing lossless and lossy compression for web images
https://developers.google.com/speed/webp/
/usr/local/Cellar/webp/0.6.0_2 (36 files, 1.9MB)
  Poured from bottle on 2017-11-27 at 16:55:56
/usr/local/Cellar/webp/0.6.1 (38 files, 2.0MB) *
  Poured from bottle on 2017-12-07 at 11:16:14
From: https://github.com/Homebrew/homebrew-core/blob/master/Formula/webp.rb
==> Dependencies
Required: libpng ✔
Recommended: jpeg ✔
Optional: libtiff ✔, giflib ✘
```

### Attachments

[...]

PR #3385, opened 2018-09-29
Change tuple background to global color table index when saving as GIF

radarhere:
Suggestion for #2949

WebP sets `im.info['background']` to an RGBA tuple. However, GIF uses background as an ['Index into the Global Color Table'](https://www.w3.org/Graphics/GIF/spec-gif89a.txt). This PR converts the value from the tuple to a global color table index when saving as a GIF.

I will note that this isn't perfect - `ImagePalette.getcolor` ignores the alpha part of the RGBA tuple, and I think there a few things wrong with the way that we are handling GIF backgrounds. However, I don't think there is anything wrong with the code that I am adding in.

### DIFF
--- a/src/PIL/GifImagePlugin.py
+++ b/src/PIL/GifImagePlugin.py
@@ -711,10 +711,17 @@ def _get_global_header(im, info):
         if im.info.get("version") == b"89a":
             version = b"89a"
 
+    background = 0
+    if "background" in info:
+        background = info["background"]
+        if isinstance(background, tuple):
+            # WebPImagePlugin stores an RGBA value in info["background"]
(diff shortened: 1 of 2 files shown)


src/PIL/GifImagePlugin.py
```diff
@@ -704,11 +704,15 @@ def _get_global_header(im, info):
         if im.info.get("version") == b"89a":
             version = b"89a"
 
+    background = 0
+    if "background" in info:
+        background = info["background"]
+        if isinstance(background, tuple):
```
homm: Since this is a kind of hack, please add some comments about why we are doing this.

radarhere: Okay, done

PR #3197, opened 2018-06-21
Avoid undefined behaviour due to division by zero

andreas-schwab:

PR #3388, opened 2018-09-29
Avoid undefined behaviour due to division by zero

hugovk:
Continuation of and closes #3197.

* Same as #3197 but incorporating review comment to re-use an existing variable
* And rebased on master


tkoeppe:
Thank you very much!

PR #3099, opened 2018-04-14
Numpy support for LUTs 

homm:
This dramatically improves the performance of arguments conversion when a source table supports buffer interface (numpy && arrays in Python > 3).

```python
from PIL import Image, ImageFilter
import array
import numpy

lut_33 = ImageFilter.Color3DLUT.generate(33, lambda r, g, b: (r, g, b))
lut_33_numpy = ImageFilter.Color3DLUT.generate(33, lambda r, g, b: (r, g, b))
lut_33_numpy.table = numpy.array(lut_33_numpy.table, numpy.float32)
lut_33_array = ImageFilter.Color3DLUT.generate(33, lambda r, g, b: (r, g, b))
lut_33_array.table = array.array('f', lut_33_array.table)
lut_65 = ImageFilter.Color3DLUT.generate(65, lambda r, g, b: (r, g, b))
lut_65_numpy = ImageFilter.Color3DLUT.generate(65, lambda r, g, b: (r, g, b))
lut_65_numpy.table = numpy.array(lut_65_numpy.table, numpy.float32)
lut_65_array = ImageFilter.Color3DLUT.generate(65, lambda r, g, b: (r, g, b))
lut_65_array.table = array.array('f', lut_65_array.table)

%timeit Image.new('RGB', (1,1)).filter(lut_33)
%timeit Image.new('RGB', (1,1)).filter(lut_33_array)
%timeit Image.new('RGB', (1,1)).filter(lut_33_numpy)
%timeit Image.new('RGB', (1,1)).filter(lut_65)
%timeit Image.new('RGB', (1,1)).filter(lut_65_array)
%timeit Image.new('RGB', (1,1)).filter(lut_65_numpy)
```

|Data type | master  | buffer interface |
|-|-|-|
|list 33³  | 0.84 ms | 0.88 ms |
|array 33³ | 3.25 ms | 0.33 ms |
|numpy 33³ | 6.39 ms | 0.33 ms |
|list 65³  | 6.37 ms | 6.56 ms |
|array 65³ | 30.1 ms | 2.54 ms |
|numpy 65³ | 51.5 ms | 2.48 ms |

PR #3392, opened 2018-10-02
Update tests

hugovk:
The neater test-skipping idea is from https://github.com/python-pillow/Pillow/pull/3099.


hugovk:
Rebased.

PR #3393, opened 2018-10-02
Fixes for issues reported by static code analysis

frenzymadness:
Hello.

With our instance of static analysis software, we found some issues which may cause some troubles with un-terminated buffers and potential leaked storage.

Pages with problems description:
* [CWE-120: Buffer Copy without Checking Size of Input ('Classic Buffer Overflow')](https://cwe.mitre.org/data/definitions/120.html)
* [CWE-772: Missing Release of Resource after Effective Lifetime](https://cwe.mitre.org/data/definitions/772.html)

This PR contains fixes for almost all of the reported issue (some were false-positives) and it doesn't change any functionality. Also, all test should pass.


hugovk:
Thank you.

PR #3395, opened 2018-10-04
Improve configuration

diorcety:
Fix pkg-config call: If there is multiple spaces between pkg-config output elements the indexing is not correct ([1] can be an empty string and the value is in [2])
The second commit allow to define *_ROOT using environment variables


kkopachev:
there is also #3241

PR #3397, opened 2018-10-04
Fixed documentation on bilevel conversion

### LINKED ISSUE #3394: Incorrect output when converting greyscale to bilevel
### What did you do?
converted test image to bilevel
Most of the image should be black, only the very bottom should be white
### What actually happened?
half is black, half is white

* OS: Windows 10
* Python: 3.7.0 (v3.7.0:1bf9cc5093, Jun 27 2018, 04:06:47) [MSC v.1914 32 bit (Intel)]
* Pillow: 5.3.0

Documentation states 
If dither is NONE, all non-zero values are set to 255 (white)

included image and code show that this is not the case, instead any value under 128 becomes white while any value over 128 becomes black.
Considering the behavior of other programs, it is likely that the documentation is wrong rather than the behavior.

[graydient.zip](https://github.com/python-pillow/Pillow/files/2441111/graydient.zip)

Files have been zipped due to github not accepting bmp images.

```python
Image.open("input.bmp").convert("1", dither=Image.NONE).save("output.bmp")
```

eseifert:
The cutoff when converting an image to bilevel (mode "1") is 128.

Fixes #3394.

 * Document the correct threshold used to convert to a bilevel image in ```Image.convert()```.

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -873,9 +873,9 @@ class Image(object):
         The default method of converting a greyscale ("L") or "RGB"
         image into a bilevel (mode "1") image uses Floyd-Steinberg
         dither to approximate the original image luminosity levels. If
-        dither is NONE, all non-zero values are set to 255 (white). To
-        use other thresholds, use the :py:meth:`~PIL.Image.Image.point`
-        method.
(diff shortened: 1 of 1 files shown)


hugovk:
Thanks!

Issue #1455, opened 2015-10-02
Python 3.5/3.6/3.7 on Appveyor


wiredfool:
We should be testing on Python 3.5 on Appveyor. This is going to take adding the new compiler info into the config.py structures and testing it on Visual Studio 2015.

wiredfool:
Also, we should save n days of nightlies or merges to master,

radarhere:
Maybe this was already the case when you created the issue, but Python 3.5.0 is now up on AppVeyor.

wiredfool:
Yes, the real issue is getting all the settings right for a build using VS2015.

aclark4life:
Seems like @cgohlke made a comment about AppVeyor recently, not sure if this is related.

wiredfool:
Really, this time. I want this to work.

hugovk:
VS 2015 can be used for Python 3.5-3.7:


Python version	Visual Studio
Python 2.7	VS 2008 and VS 2010
Python 3.4	VS 2010
Python 3.5 and 3.6	VS 2015
Python 3.7 (master)	VS 2015 or VS 2017

http://pythondev.readthedocs.io/windows.html

hugovk:
Brilliant, thanks and https://github.com/python-pillow/Pillow/pull/3399 merged!

The second bit can be done with "artifacts":

https://www.appveyor.com/docs/packaging-artifacts/

https://ci.appveyor.com/project/python-pillow/pillow/settings/artifacts

They're kept for 6 months, which should be plenty. They can be copied to external storage like S3, which might be useful later, but for now AppVeyor's own storage should suffice.

Which files and folders should be included?

hugovk:
Hang on, we already have artifacts enabled!

Added in https://github.com/python-pillow/Pillow/pull/1818 and https://github.com/python-pillow/Pillow/pull/1927:

https://github.com/python-pillow/Pillow/blob/07645bd82878a12af70b00698805f650c2548c60/.appveyor.yml#L94-L120

Here's a `Pillow-5.4.0.dev0-py3.7-win32.egg` and `Pillow-5.4.0.dev0-cp37-cp37m-win32.whl` from a recent build:

https://ci.appveyor.com/project/python-pillow/pillow/build/job/i8yyorldevfyotcr/artifacts

PR #3399, opened 2018-10-08
Added Python 3.5-3.7 AppVeyor builds

radarhere:
See #1455

### DIFF
--- a/winbuild/config.py
+++ b/winbuild/config.py
@@ -3,12 +3,13 @@ import os
 SF_MIRROR = 'http://iweb.dl.sourceforge.net'
 PILLOW_DEPENDS_DIR = 'C:\\pillow-depends\\'
 
-pythons = {  # '26': 7,
-           '27': 7,
-           'pypy2': 7,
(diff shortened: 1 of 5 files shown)


winbuild/config.py
```diff
@@ -3,12 +3,13 @@
...
 
-pythons = {  # '26': 7,
-           '27': 7,
-           'pypy2': 7,
-           # '32': 7,
-           '33': 7.1,
-           '34': 7.1}
+pythons = {'27': {'compiler':7, 'vc':2008},
+           'pypy2': {'compiler':7, 'vc':2008},
+           '33': {'compiler':7.1, 'vc':2010},
```
hugovk: Shall we remove this line? It's not used and EOL Python 3.3 was removed in 5.0.0 (2018-01-01).

hugovk:
Great, thank you for doing this!

winbuild/config.py
radarhere: Okay, done

PR #3406, opened 2018-10-14
Add negative index to PixelAccess

Nazime:
Add a negative indexing to the PixelAccess so we can get the last pixels easily

```python
from PIL import Image

im = Image.open("image.png")
pix = im.load()
print(pix[-1, -1])
```

### DIFF
--- a/Tests/test_image_access.py
+++ b/Tests/test_image_access.py
@@ -68,10 +68,10 @@ class TestImagePutPixel(AccessTest):
         im1 = hopper()
         im2 = Image.new(im1.mode, im1.size, 0)
 
-        width, heigth = im1.size
-        self.assertEqual(im1.getpixel((0, 0)), im1.getpixel((-width, -heigth)))
+        width, height = im1.size
+        self.assertEqual(im1.getpixel((0, 0)), im1.getpixel((-width, -height)))
         self.assertEqual(im1.getpixel((-1, -1)),
-                         im1.getpixel((width-1, heigth-1)))
+                         im1.getpixel((width-1, height-1)))
 
         for y in range(-1, -im1.size[1]-1, -1):
             for x in range(-1, -im1.size[0]-1, -1):


src/_imaging.c
```diff
@@ -471,7 +471,12 @@ getpixel(Imaging im, ImagingAccess access, int x, int y)
       INT32 i;
       FLOAT32 f;
     } pixel;
-
+    if (x < 0){
+        x = im->xsize + x;
+    }
+    if (y < 0){
```
hugovk: Please add a space between `)` and `{`, here and above.

src/PIL/PyAccess.py
```diff
@@ -88,8 +88,12 @@ def __getitem__(self, xy):
         :returns: a pixel value for single band images, a tuple of
           pixel values for multiband images.
         """
-
-        (x, y) = self.check_xy(xy)
+        (x, y) = xy
+        if (x < 0):
+            x = self.xsize + x
+        if (y < 0):
```
hugovk: Please remove redundant parentheses, here and above.

src/_imaging.c
```diff
@@ -471,7 +471,12 @@ getpixel(Imaging im, ImagingAccess access, int x, int y)
       INT32 i;
       FLOAT32 f;
     } pixel;
-
+    if (x < 0){
```
hugovk: Let's keep that line of whitespace after `pixel;`

src/_imaging.c
```diff
@@ -471,7 +471,12 @@ getpixel(Imaging im, ImagingAccess access, int x, int y)
...
+        x = im->xsize + x;
+    }
+    if (y < 0){
+        y = im->ysize + y;
+    }
```
hugovk: And add a line of whitespace after this.

radarhere:
Thanks for this.

You’ve added getting the negative index - it would be good to also add setting.

[`im.load()` returns either PyAccess (Python) or PixelAccess (C)](https://github.com/python-pillow/Pillow/blob/master/src/PIL/Image.py#L842-L843 ). So that’s why you are finding that you need to change the C code.

Nazime:
What do you mean by add setting?

radarhere:
You have added support for `print(pix[-1, -1])`, but not `pix[-1, -1] = (0, 0, 0)` - the first being 'get', the second being 'set'.

Nazime:
Oh yes, I'll add that too.

Tests/test_image_access.py
```diff
@@ -64,6 +64,43 @@ def test_sanity(self):
 
         self.assert_image_equal(im1, im2)
 
+    def test_sanity_negative_index(self):
+        im1 = hopper()
+        im2 = Image.new(im1.mode, im1.size, 0)
+
+        width, heigth = im1.size
```
radarhere: height

Nazime: Oups didn't see it

hugovk:
Let' remember to add this to the release notes.

PR #3412, opened 2018-10-17
CI: Fix diff-cover, move linting to own job, name the jobs

hugovk:

 * Fix diff-cover checks: 
   * Fix install `no such option: --use-wheel`
   * It tried to use the old pep8 tool, which is now pycodestyle

 * Move the linting to its own job, but do it inside `after_success` so it won't fail the build. This makes it easier to find. I chose 3.6 as the base Python, as it's the newest that doesn't need to download the Python archive, so is faster. It runs first to get fast feedback.

 * Name the non-Docker build jobs, so we can clearly see which are Trusty and which are Xenial.

# Before

image

# After

image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,13 +6,18 @@ cache: pip
 notifications:
   irc: "chat.freenode.net#pil"
 
-# Run slow PyPy* first, to give them a headstart and reduce waiting time.
+# Run fast lint first to get fast feedback.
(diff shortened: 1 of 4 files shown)


hugovk:
A downside of that is it would be run last:

https://travis-ci.org/hugovk/Pillow/builds/443097000

Let's not worry about it, I'm going to make a follow-up PR to let it fail on errors, and fix some flake8 things and exclude a lot of other rules.

PR #3341, opened 2018-09-08
Close existing WebP fp before setting new fp

radarhere:
Resolves #3340

PR #3415, opened 2018-10-18
Quick fix: revert #3341 due to regression

homm:
```python

with open('input/256.webp', 'rb') as blob:
    Image.open(blob).load()
    Image.open(blob).load()
```

```
/home/homm/Code/Pillow/src/PIL/Image.pyc in open(fp, mode)
   2611 
   2612     try:
-> 2613         fp.seek(0)
   2614     except (AttributeError, io.UnsupportedOperation):
   2615         fp = io.BytesIO(fp.read())

ValueError: I/O operation on closed file
```


homm:
@radarhere @hugovk I reverting #3341. Warning is better than ValueError.

Please, look for a better way to fix it, probably with `_exclusive_fp` check.

PR #3434, opened 2018-10-24
Fix issues with GIF transparency

### LINKED ISSUE #3433: GIF transparency not applied correctly
The attached animated GIF image is decoded incorrectly (testing with master).  This was exported by Photoshop and is decoded correctly by ffmpeg and other decoders.

The file is 3x1 pixels and three frames: XBX GBG XBX (X: transparent, G: green, B: blue).
Pillow decodes it as: XBX GXG XBX, leaving pixel 1x0 in the second frame transparent when it should be blue.

The issue seems to be that the transparent color index isn't handled correctly.  It should leave the existing pixel in the buffer unchanged.  Instead, the decoder overwrites the pixel with the transparent index, changing the blue pixel to transparent.  There's a special case for dispose mode 1 in load_end, but it's a little odd.  (The dispose mode only affects the next frame, but this is affecting the current frame, and it doesn't do anything with other dispose modes.)

I'll send a PR shortly to fix this.  It adds two tests using this image: one for this issue, and one for an issue that came up while implementing this.  (Reading img.n_pixels changed the image, since the image wasn't being reset when decoding frame 0 after rewinding.  This didn't matter before, since the decoder overwrote the whole image, but it matters now.)  The load_end special case is removed, since the compositing happens in the decoder.  Background dispose clears to the transparency color if one is set, which matches other decoders.

If anyone's interested in trying to make the decoder work everywhere, it might be useful to check other projects for comprehensive GIF tests.  There's probably one in Chromium.

test

zewt:
Fixes #3433.


hugovk:
Thank you for the bug report and PR!

Please can you check the CI failures? 

Python 3 is failing because it doesn't have xrange. 

Python 2 is failing in Tests/test_file_webp_animated.py.

zewt:
I'm looking at it now.  The test system silently turns tests off if some features aren't being built and I didn't expect GIF tests inside WebP tests.  Will take a bit (need to build a newer libwebp).

src/libImaging/GifDecode.c
```diff
@@ -264,29 +264,33 @@ ImagingGifDecode(Imaging im, ImagingCodecState state, UINT8* buffer, int bytes)
...
+	if(context->transparency == -1)
+	{
+	    if (i == 1) {
+		if (state->x < state->xsize-1) {
+		    /* Single pixel, not at the end of the line. */
+		    *out++ = p[0];
+		    state->x++;
+		    continue;
+		}
+	    } else if (state->x + i <= state->xsize) {
```
homm: Please, always use 4 spaces as indention everywhere in added code

zewt: I'm using the same indentation style as the rest of the file: 8-space tabs with 4-space blocks (vim: noet sw=4 ts=8).

hugovk:
PR to fix lint failures: https://github.com/zewt/Pillow/pull/1.

src/PIL/GifImagePlugin.py
```diff
@@ -144,7 +144,7 @@ def _seek(self, frame):
             self.disposal_method = 0
         else:
             # ensure that the previous frame was loaded
-            if not self.im:
+            if self.tile:
```
radarhere: Okay, so I realise now that this commit is solving the same problem as #3716.

To compare the two solutions, #3716 adds a new load every time a backwards seek is made. This solution adds a new load for each frame seeked forward.

Adding some print statements to demonstrate, I get -

```python
>>> from PIL import Image
>>> im = Image.open("Tests/images/iss634.gif")
>>> im.seek(10)
This adds a new load, when seeking forward to 2
This adds a new load, when seeking forward to 3
This adds a new load, when seeking forward to 4
This adds a new load, when seeking forward to 5
This adds a new load, when seeking forward to 6
This adds a new load, when seeking forward to 7
This adds a new load, when seeking forward to 8
This adds a new load, when seeking forward to 9
This adds a new load, when seeking forward to 10
>>> im.seek(5)
#3716 adds a new load, when seeking back to 5
This adds a new load, when seeking forward to 1
This adds a new load, when seeking forward to 2
This adds a new load, when seeking forward to 3
This adds a new load, when seeking forward to 4
This adds a new load, when seeking forward to 5
```

The tests added in this PR still pass with the change from #3716. So I'm in favour of #3716. Feel free to add your thoughts.

zewt: I don't entirely follow (don't have time to dig into it again right now), but if you seek forward in a GIF, you need to decode each frame, since the decoding is incremental.

radarhere: No worries if you don't have time right now, and If you think that this needs to be there instead of or regardless of my change, that's fine. I'm just saying that the tests don't show evidence of that.

src/PIL/GifImagePlugin.py
```diff
@@ -259,9 +258,10 @@ def _seek(self, frame):
                 # do not dispose or none specified
                 self.dispose = None
             elif self.disposal_method == 2:
-                # replace with background colour
-                self.dispose = Image.core.fill("P", self.size,
-                                               self.info["background"])
+                # replace with transparency if there is one, otherwise
+                # background colour
```
radarhere, line 262: Why potentially use the transparency? I don't see this in the spec.

radarhere:
I've cherry-picked two commits from here into #5333

radarhere:
I've removed the tests from here that pass without this PR.

src/PIL/GifImagePlugin.py
radarhere: Ok, I've separated this out into #5386

src/PIL/GifImagePlugin.py
```diff
@@ -309,6 +313,17 @@ def _seek(self, frame):
...
+    def load_prepare(self):
+        super(GifImageFile, self).load_prepare()
+
+        if self.__frame == 0:
+            # On the first frame, clear the buffer.  Use the transparency index
+            # if we have one, otherwise use the background index.
+            default_color = self.info.get(
+                "transparency", self.info.get("background", 0)
+            )
+            self.im.paste(default_color, (0, 0, self.size[0], self.size[1]))
```
radarhere, line 325:
```suggestion
        if not self.im:
            # On the first frame, clear the buffer.  Use the transparency index
            # if we have one, otherwise use the background index.
            default_color = self.info.get(
                "transparency", self.info.get("background", 0)
            )
            self.im = Image.core.fill(self.mode, self.size, default_color)

        super(GifImageFile, self).load_prepare()
```
This saves the parent `load_prepare` from creating an image only to have it overwritten here. Instead, this just creates the image and the parent does not.

radarhere: I've created #5557 with this functionality, minus the background color fallback.

src/PIL/GifImagePlugin.py
radarhere: I've found https://legacy.imagemagick.org/Usage/anim_basics/#dispose
> is cleared to transparency
> ...
> There is some thinking that rather than clearing the overlaid area to the transparent color, this disposal should clear it to the 'background' color meta-data setting stored in the GIF animation
> ...
> However all modern web browsers clear just the area that was last overlaid to transparency, as such this is now accepted practice, and what IM now follows.

radarhere:
I've pulled the remaining parts of this into #5557

PR #3435, opened 2018-10-24
Install project using pip in ReadTheDocs build

radarhere:
Possible solution for #2763 

Given the intermittent nature of the problem, I can't say anything definitively, but here is my evidence that it might solve it - testing builds of the 5.3.x branch against this, 5.3.x fails 3 out of 6, and this fails 0 out of 6.

readthedocs

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,2 @@
+python:
+    pip_install: true


hugovk:
Thanks, let's give it a shot!

PR #3440, opened 2018-10-25
Fix flake8

hugovk:
# Before

```console
$ flake8 --statistics --count
...
8     E122 continuation line missing indentation or outdented
5     E127 continuation line over-indented for visual indent
31    E128 continuation line under-indented for visual indent
5     E131 continuation line unaligned for hanging indent
6     E201 whitespace after '('
48    E265 block comment should start with '# '
13    E402 module level import not at top of file
72    E501 line too long (80 > 79 characters)
1     E711 comparison to None should be 'if cond is None:'
22    E722 do not use bare 'except'
3     E741 ambiguous variable name 'I'
11    F401 'PIL.ImageDraw' imported but unused
10    F821 undefined name 'unicode'
20    F841 local variable 'msg' is assigned to but never used
2     W605 invalid escape sequence '\.'
257
```

# After

```console
$ flake8 --statistics --count
42    E501 line too long (80 > 79 characters)
22    E722 do not use bare 'except'
3     E741 ambiguous variable name 'I'
10    F821 undefined name 'unicode'
77
```

I added some targeted `# noqa: <code>` comments which prevent QA tools inspecting them for a specific error code. I'm generally against these, but I think some exceptions are acceptable so we can have other valid errors flagged. 

For example six examples of these whitespace errors are allowed only in cases where a large, aligned matrix is shown:
```
E127 continuation line over-indented for visual indent
E128 continuation line under-indented for visual indent
E131 continuation line unaligned for hanging indent
```

And a single `local variable is assigned to but never used` allowed only where it's specifically needed for a test case:

```python
        # Yeah. Watch some JIT optimize this out.
        pattern = None	        pattern = None  # noqa: F841
```


Tests/test_image.py
```diff
@@ -139,7 +139,6 @@ def test_comparison_with_other_type(self):
         # Act/Assert
         # Shouldn't cause AttributeError (#774)
         self.assertFalse(item is None)
-        self.assertFalse(item == None)
```
radarhere: This is actually testing the `__eq__` method, whereas `is` does not.

```
class C:
    def __eq__(self, other):
        print("method tested")


a = C()
b = C()
print("is check")
a is b
print("equal check")
a == 0
```

Instead of deleting the line, my suggested fix would be to change it to testing `== 0`. This calls `__eq__`, without triggering flake8.

radarhere: Oh, wait, that's exactly what the next line does. Never mind then.

hugovk: Thanks for checking :)

Issue #3441, opened 2018-10-27
Jpeg load performance regression for Pillow 5.1+ from wheels


kkopachev:
### What did you do?
Performance regression when opening (and loading) a jpeg file in Pillow, installed from wheels on x86_64 linux. Affects loading of jpeg files. Regression happened updating from Pillow 5.0.0 to 5.1.0 (up to 5.3.0)

### What are your OS, Python and Pillow versions?

* OS: linux, x86_64
* Python: 2.7, 3.7
* Pillow: 5.1.0+

I am not sure where this issue belongs to, seems to be related to wheels building somehow.

Original issue was created in thumbor repo https://github.com/thumbor/thumbor/issues/1102 saying there was a noticeable performance regression. Upon further investigation, I was able to come up with simple reproduction steps.

```
docker run --rm -it python:2 bash
root@9e6e1f13f44a:/# wget -q wget https://upload.wikimedia.org/wikipedia/commons/b/b4/JPEG_example_JPG_RIP_100.jpg
root@9e6e1f13f44a:/# pip install pillow==5.0.0
Collecting pillow==5.0.0
  Downloading https://files.pythonhosted.org/packages/89/bd/1d9a10f3e8157b7df275740b0782a892a0db387f8286620110c41e5146c7/Pillow-5.0.0-cp27-cp27mu-manylinux1_x86_64.whl (5.8MB)
    100% |████████████████████████████████| 5.9MB 1.5MB/s 
Installing collected packages: pillow
Successfully installed pillow-5.0.0
You are using pip version 18.0, however version 18.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
root@9e6e1f13f44a:/# wget https://upload.wikimedia.org/wikipedia/commons/b/b4/JPEG_example_JPG_RIP_100.jpg
root@9e6e1f13f44a:/#
root@9e6e1f13f44a:/#
root@9e6e1f13f44a:/# python -m timeit -n1000 -r5 -s'from PIL import Image' 'Image.open("JPEG_example_JPG_RIP_100.jpg").load()'
1000 loops, best of 5: 2.58 msec per loop
root@9e6e1f13f44a:/#
root@9e6e1f13f44a:/#
root@9e6e1f13f44a:/# pip install pillow==5.1.0
Collecting pillow==5.1.0
[...]

PR #3442, opened 2018-10-29
Updated libimagequant to 2.12.2

radarhere:


hugovk:
Perhaps we should hold off merging lib updates until we know more about https://github.com/python-pillow/Pillow/issues/3441.

hugovk:
We now know https://github.com/python-pillow/Pillow/issues/3441 is to do with build flags, let's merge these. Thanks!

PR #3454, opened 2018-11-07
Do not play sound when running screencapture command

radarhere:
This PR changes the ImageGrab test to run screencapture silently. Without it, a sound plays when running the test.


hugovk:
Confirmed with Python 3.7 / macOS High Sierra.

PR #3464, opened 2018-11-16
Check quality_layers type

### LINKED ISSUE #3463: Segmentation Fault while saving jpeg2000 image using wrong param
### What did you do?
Save a jpeg2000 image.
`Image.fromarray(np.zeros((64,64), dtype=np.uint8)).save('img.jp2', quality_layers='40')`
Throw an exception.
### What actually happened?
`[1]    18024 segmentation fault (core dumped)  python3`

* OS: Ubuntu 16.04
* Python: 3.5.2
* Pillow: 5.3.0

I know that I should pass a number to `quality_layers` instead of a string. But it should not cause python segmentation fault. It is better to throw a exception in python.

```python
import numpy as np
from PIL import Image
Image.fromarray(np.zeros((64,64), dtype=np.uint8)).save('img.jp2', quality_layers='40')
```

radarhere:
Resolves #3463

### DIFF
--- a/Tests/test_file_jpeg2k.py
+++ b/Tests/test_file_jpeg2k.py
@@ -105,6 +105,22 @@ class TestFileJpeg2k(PillowTestCase):
         im.load()
         self.assertEqual(im.size, (160, 120))
 
+    def test_layers_type(self):
+        outfile = self.tempfile('temp_layers.jp2')
+        for quality_layers in [
+            [100, 50, 10],
+            (100, 50, 10),
+            None
+        ]:
+            test_card.save(outfile, quality_layers=quality_layers)
+
+        for quality_layers in [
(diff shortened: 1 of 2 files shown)


hugovk:
Do we need to raise against sequence containing a string? eg. `quality_layers=["40"]`

Out of curiosity: I know it's against the documentation, but what currently happens  with `quality_layers=40`?

radarhere:
Okay, I've updated to include checking for numbers.

From my investigation, `quality_layers=40` does nothing, as it doesn't pass `PySequence_Check` - https://github.com/python-pillow/Pillow/blob/e3d5494a513b7454efc7eb10241b1ce2b28f939d/src/encode.c#L1052

PR #3482, opened 2018-11-29
Qualify reference to LOAD_TRUNCATED_IMAGES

distributist:
Setting ImageFile.LOAD_TRUNCATED_IMAGES = True had no effect in loading JPEG images with "extra" data after the end of image marker in a Zope application. Tracing the code showed that although ImageFile.LOAD_TRUNCATED_IMAGES was True, the unqualified reference in ImageFile.py to LOAD_TRUNCATED_IMAGES returned False, and so the load failed.

* Add ImageFile. qualifier to LOAD_TRUNCATED_IMAGES references

### DIFF
--- a/src/PIL/ImageFile.py
+++ b/src/PIL/ImageFile.py
@@ -221,13 +221,13 @@ class ImageFile(Image.Image):
                                 s = read(self.decodermaxblock)
                             except (IndexError, struct.error):
                                 # truncated png/gif
-                                if LOAD_TRUNCATED_IMAGES:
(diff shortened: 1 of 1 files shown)


distributist:
Ok, this change works in my development environment, but fails with an AttributeError in my production environment. I had my doubts about such a reference, but when it did in my tests, I assumed it was valid. I have no idea why it would work in one, and not the other. 

This is an old, complex application, so I'm not going to open an issue, as I have a feeling a simple reproduction would be difficult, as this seems to work for others who have tried it. I'm going to close this, and will open a new request if I am able to work on this further.

PR #3493, opened 2018-12-12
Added libjpegturbo to check_feature

### LINKED ISSUE #3492: How to tell at run time whether libjpeg-turbo version of libjpeg is used?
tl;dr:

Is there some way to accomplish:  `PIL.Image.libjpeg_turbo_is_enabled()`?

The full story:

Is there a way to tell from a pre-built Pillow whether it was built against `libjpeg-turbo` or not?

This is assuming that all I have is `libjpeg.so.X.X` and no way to tell where it came from.

I see there is a symbol in the library:
```
nm _imaging.cpython-36m-x86_64-linux-gnu.so | grep -I turbo
000000000007e5a0 D libjpeg_turbo_version
```
but I don't know how to access its value from python.

If there is a way to tell the same from from shell using `ldd`/`nm` or other linker tools, it'd do too.

The intention is to be able to tell a user at run-time to re-build Pillow after installing `libjpeg-turbo` to gain speed. The problem is that It's not enough to build Pillow against  `libjpeg-turbo`. Given how conda/pip dependencies work, a new prebuilt package of `Pillow` could get swapped in as a dependency for some other package and the user won't know that they now run a less efficient `Pillow` unless they watch closely any install/update logs.

Currently the only solution I can think of (in conda env) is to take the output of:

    cd ~/anaconda3/envs/pytorch-dev/lib/python3.6/site-packages/PIL
    ldd  _imaging.cpython-36m-x86_64-linux-gnu.so | grep libjpeg

which wold give me something like:

    libjpeg.so.8 => ~/anaconda3/envs/pytorch-dev/lib/libjpeg.so.8

And then to try to match it to:

    grep libjpeg  ~/anaconda3/envs/pytorch-dev/conda-meta/libjpeg-turbo-2.0.1-h470a237_0.json

which may work. There is a problem with this approach

It's very likely that conda is going to reinstall `jpeg` since many packages depend on it, and when it does, there is going to be 2 libjpeg libs.

    ldd  _imaging.cpython-36m-x86_64-linux-gnu.so | grep libjpeg
        libjpeg.so.8 => /home/stas/anaconda3/envs/pytorch-dev/lib/libjpeg.so.8 (0x00007f92628c8000)
        libjpeg.so.9 => /home/stas/anaconda3/envs/pytorch-dev/lib/./libjpeg.so.9 (0x00007f9261c4e000)

And now I can no longer tell which is which, since I can no longer tell which of the two Pillow will load at run time. Well, I can go one step further and check /proc/<pid>/maps to get the library, but it's getting more and more convoluted. And I won't even know how to do the same on non-linux platform. And this is just for the conda setup, for pip setup it'd be something else.

Also what happens if `libjpeg-turbo` and `libjpeg` are the same version?

Perhaps there is an easier way? Any chance to have `PIL.Image.libjpeg_turbo_is_enabled()`?

Thank you.

radarhere:
Resolves #3492, by allowing the following code to check for libjpegturbo.

```python
from PIL import features
features.check_feature('libjpegturbo')
```

### DIFF
--- a/src/PIL/features.py
+++ b/src/PIL/features.py
@@ -51,7 +51,8 @@ features = {
     "webp_anim": ("PIL._webp", 'HAVE_WEBPANIM'),
     "webp_mux": ("PIL._webp", 'HAVE_WEBPMUX'),
     "transp_webp": ("PIL._webp", "HAVE_TRANSPARENCY"),
-    "raqm": ("PIL._imagingft", "HAVE_RAQM")
+    "raqm": ("PIL._imagingft", "HAVE_RAQM"),
+    "libjpegturbo": ("PIL._imaging", "HAVE_LIBJPEGTURBO")
 }
 
 
--- a/src/PIL/features.py
+++ b/src/PIL/features.py
@@ -52,7 +52,7 @@ features = {
     "webp_mux": ("PIL._webp", 'HAVE_WEBPMUX'),
     "transp_webp": ("PIL._webp", "HAVE_TRANSPARENCY"),
     "raqm": ("PIL._imagingft", "HAVE_RAQM"),
-    "libjpegturbo": ("PIL._imaging", "HAVE_LIBJPEGTURBO"),
+    "libjpeg_turbo": ("PIL._imaging", "HAVE_LIBJPEGTURBO"),
 }
 
 
--- a/src/PIL/features.py
+++ b/src/PIL/features.py
@@ -52,7 +52,7 @@ features = {
     "webp_mux": ("PIL._webp", 'HAVE_WEBPMUX'),
     "transp_webp": ("PIL._webp", "HAVE_TRANSPARENCY"),
     "raqm": ("PIL._imagingft", "HAVE_RAQM"),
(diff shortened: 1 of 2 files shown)


src/PIL/features.py
```diff
@@ -51,7 +51,8 @@ def get_supported_codecs():
     "webp_anim": ("PIL._webp", 'HAVE_WEBPANIM'),
     "webp_mux": ("PIL._webp", 'HAVE_WEBPMUX'),
     "transp_webp": ("PIL._webp", "HAVE_TRANSPARENCY"),
-    "raqm": ("PIL._imagingft", "HAVE_RAQM")
+    "raqm": ("PIL._imagingft", "HAVE_RAQM"),
+    "libjpegturbo": ("PIL._imaging", "HAVE_LIBJPEGTURBO")
```
hugovk: Nit: append a comma so the next addition is only a one-line diff: 

```suggestion
    "libjpegturbo": ("PIL._imaging", "HAVE_LIBJPEGTURBO"),
```

stas00:
It works, thank you.

hugovk:
Thanks both!

PR #3498, opened 2018-12-17
Cache EXIF information

Glandos:
In case we need to call _getexif from the outside, since it's already called while parsing.

This give a great improvement in my test using [sigal](saimn/sigal): parsing a library of 379 photos (without resizing, only metadata reading) went from 2.36295 seconds down to 1.92881 (mean time over 3 rounds).

Caching costs memory though. But this small PR is here to start the discussion.

### DIFF
--- a/src/PIL/JpegImagePlugin.py
+++ b/src/PIL/JpegImagePlugin.py
@@ -455,6 +455,12 @@ def _getexif(self):
     # and is likely to be replaced with something better in a future
     # version.
 
+    # Use the cached version if possible
+    try:
+        return self.info["parsed_exif"]
+    except KeyError:
+        pass
+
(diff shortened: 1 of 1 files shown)


homm:
There is another `_getexif` method in `WebPImageFile` class. Maybe it is better to cache parsed exif in `im.info['_parsed_exif']` or something similar?

Glandos:
Good catch.
It could be a better idea to patch `JpegImagePlugin._getexif` instead.

I don't know how do you want to have it: a new pull request, or new commits into this one?

radarhere:
I would keep it as part of this PR.

Glandos:
I did a forced push to avoid showing revert commits.

hugovk:
Thanks!

PR #3502, opened 2018-12-18
Size updates

surgan12:
This PR adds the option of taking only one input i.e an int for square images. in various functions


surgan12:
Reviews ?

PR #3505, opened 2018-12-24
Added Tidelift badge

radarhere:
URLs taken from https://tidelift.com/lifter/pypi/Pillow/tasks/2439


hugovk:
Is there a badge with rounded corners, like the others? If not, we can ask for one and/or use a Shields image for the badge.

radarhere:
Experimenting, I found that appending '?style=flat' worked, changing it from

default

to

new

PR #3512, opened 2018-12-28
Changed docker tag to master

radarhere:


.travis.yml
```diff
@@ -73,7 +73,7 @@ install:
     fi
 
 before_install:
-  - if [ "$DOCKER" ]; then travis_retry docker pull pythonpillow/$DOCKER:$DOCKER_TAG; fi
```
wiredfool: This should probably remain, but we should define the docker tag somewhere in the environment.

radarhere: Okay, I've updated the commit to just change the tags to master.

PR #3520, opened 2018-12-31
Adding missing mime types and extensions

pirate486743186:
for reference

https://en.wikipedia.org/wiki/BMP_file_format
https://en.wikipedia.org/wiki/ICO_(file_format)
https://en.wikipedia.org/wiki/Pcx
https://en.wikipedia.org/wiki/Netpbm_format
https://en.wikipedia.org/wiki/.tga

there's also "image/x-ms-bmp"


pirate486743186:
as is, it's actually wrong https://github.com/python-pillow/Pillow/pull/3501#pullrequestreview-188572018

pirate486743186:
I think, in my noob mind, that this is now mergable as is.
I moved the rest of the mime types here #3531

hugovk:
This will need tests, similar to those in https://github.com/python-pillow/Pillow/pull/3525.

src/PIL/BmpImagePlugin.py
```diff
@@ -347,6 +347,6 @@ def _save(im, fp, filename):
 Image.register_open(BmpImageFile.format, BmpImageFile, _accept)
 Image.register_save(BmpImageFile.format, _save)
 
-Image.register_extension(BmpImageFile.format, ".bmp")
+Image.register_extensions(BmpImageFile.format, [".bmp", ".dib"])
```
radarhere: https://www.imagemagick.org/script/formats.php

> DIB is a BMP file without the BMP header.

Since the BMP header is used to accept files for BmpImageFile, adding support is not as simple here as adding another extension.

radarhere:
I've just rebased this on master to include #3585

src/PIL/BmpImagePlugin.py
pirate486743186: Ok, removed that and created new issue.
#3689

radarhere:
My thoughts - the docs don't mention mime types. A commit has been added to update the documented list of PPM types, but there's no consistent documentation of file extensions. My instinct is that if something isn't documented, then a change doesn't need to be added to the release notes.

PR #3521, opened 2018-12-31
Update 5.4.0 release notes

hugovk:
A few things from https://github.com/python-pillow/Pillow/blob/master/CHANGES.rst#540-unreleased.

### DIFF
--- a/docs/releasenotes/5.4.0.rst
+++ b/docs/releasenotes/5.4.0.rst
@@ -38,7 +38,7 @@ For example, to get or set the farthest pixel in the lower right of an image::
 New custom TIFF tags
 ^^^^^^^^^^^^^^^^^^^^
 
-TIFF images can now be saved with custom integer, float and sting TIFF tags::
+TIFF images can now be saved with custom integer, float and string TIFF tags::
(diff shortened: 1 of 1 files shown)


docs/releasenotes/5.4.0.rst
```diff
@@ -1,9 +1,29 @@
...
+    im.show()  # Shows a single frame
+
+Check for libjpeg-turbo
+^^^^^^^^^^^^^^^^^^^^^^^
+
+You can check if Pillow has been built against the libjpeg-turbo version of the
+libjpeg library::
+
+    from PIL import features
+    features.check_feature("libjpeg_turbo")  # True or False
```
radarhere: Maybe you feel like this is an edge case because it would require Pillow to not be installed correctly, but [`None` is also a possible return value](https://github.com/python-pillow/Pillow/blob/master/src/PIL/features.py#L69)

hugovk: Yeah, this is just an example and not thorough docs. It could also raise `ValueError`. Thanks though!

PR #3523, opened 2019-01-01
Added updating Pillow in docker-images to RELEASING

radarhere:


hugovk:
Good idea!

How about putting this as a Bash script in that repo? And a similar one for Pillow Wheel Builder?

RELEASING.md
```diff
@@ -107,3 +107,19 @@ Released as needed privately to individual vendors for critical security-related
 ## Documentation
 
 * [ ] Make sure the default version for Read the Docs is the latest release version, i.e. ``5.2.0`` rather than ``latest`` e.g. https://pillow.readthedocs.io/en/5.2.x/
+
+## Docker images
+
+* [ ] Update Pillow for the [docker images](https://github.com/python-pillow/docker-images)
```
hugovk:
```suggestion
* [ ] Update Pillow for the [Docker images](https://github.com/python-pillow/docker-images):
```

wiredfool:
FWIW, this isn't really a requirement for release of pillow, only for updates to the docker images, and those don't have to change frequently unless they're under active development.  

I'd consider pulling in a source archive from pip in travis for this, but it is very useful to have a Pillow directory for docker development and pillow bug squashing.

hugovk:
Okay, let's move those instructions to that repo's README:

https://github.com/python-pillow/Pillow/pull/3534 --> https://github.com/python-pillow/docker-images/pull/49

PR #3529, opened 2019-01-02
Fix build for Termux

### LINKED ISSUE #3526: Pillow 5.4.0 cannot be installed on Termux
Pillow is upgraded to version 5.4.0

* OS:  Termux
* Python: 3.7.2
* Pillow: 5.4.0

I think the error happens in the function __find_library_dirs_ldconfig_, where the executable "/sbin/ldconfig" is called. This executable  does not exist on Termux, making the function return None.

```python
[...]
        args = ['/sbin/ldconfig', '-p']
[...]
    null = open(os.devnull, 'wb')
    try:
        with null:
            p = subprocess.Popen(args,
                                 stderr=null,
                                 stdout=subprocess.PIPE,
                                 env=env)
    except OSError:  # E.g. command not found
        return None
[...]
```
Previous versions can be installed without a problem (as long as the necessary dev packages are installed).

pslacerda:
Now `_find_library_dirs_ldconfig` returns an empty list when `ldconfig` is missing, then `setup.py` can continue guessing library directories.

Fixes #3526.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -111,7 +111,7 @@ def _find_library_dirs_ldconfig():
                                  stdout=subprocess.PIPE,
                                  env=env)
     except OSError:  # E.g. command not found
-        return None
+        return []
     [data, _] = p.communicate()
     if isinstance(data, bytes):
         data = data.decode()


pslacerda:
@valpogus I not really build Pillow on Termux because I didn't found zlib. I simply removed the error.

The directories at https://github.com/python-pillow/Pillow/issues/1957#issuecomment-304086698 are standard? My device isn't rooted so maybe you can look it for me if the headers are there and maybe add headers search paths specific for Termux. What is supposed to be inside `/data/data/com.termux/files/usr/include/`?

landfillbaby:
yup. it works now

hugovk:
Thank you for the PR!

PR #3530, opened 2019-01-02
Added `image.entropy()` method

fish2000:
This calculates the entropy for the image, based on the histogram.

Because this uses image histogram data directly, the existing C function underpinning the `image.histogram()` method was abstracted into a macro, and a new C function was added that uses this macro.

The new `image.entropy()` method is based on `image.histogram()`, and will accept the same arguments to calculate the histogram data it will use to assess the entropy of the image.

The algorithm and methodology is based on existing Python code:
* https://git.io/fhmIU

... A test case in the `Tests/` directory, and doctest lines in `selftest.py`, have both been added and checked.

 * The addition of an `image.entropy()` method,
 * The abstraction of the prologue of the C function `_histogram` into a macro, and
 * The use of that macro in both `_histogram` and `_entropy`.
 * Minor documentation addenda in the docstrings for both `image.entropy()` and `image.histogram()`

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -1361,6 +1361,7 @@ class Image(object):
         bi-level image (mode "1") or a greyscale image ("L").
 
         :param mask: An optional mask.
+        :param extrema: An optional tuple of manually-specified extrema.
         :returns: A list containing pixel counts.
         """
         self.load()
@@ -1372,6 +1373,36 @@ class Image(object):
                 extrema = self.getextrema()
             return self.im.histogram(extrema)
         return self.im.histogram()
+
+    def entropy(self, mask=None, extrema=None):
+        """
+        Returns the histogram entropy. The histogram is returned as
+        a list of pixel counts, one for each pixel value in the source
+        image. If the image has more than one band, the histograms for
+        all bands are concatenated (for example, the histogram for an
+        "RGB" image contains 768 values).
+
+        A bilevel image (mode "1") is treated as a greyscale ("L") image
+        by this method.
+
+        If a mask is provided, the method returns a histogram for those
+        parts of the image where the mask image is non-zero. The mask
+        image must have the same size as the image, and be either a
+        bi-level image (mode "1") or a greyscale image ("L").
+
+        :param mask: An optional mask.
+        :param extrema: An optional tuple of manually-specified extrema.
+        :returns: A float value measuring the image entropy.
+        """
+        self.load()
+        if mask:
+            mask.load()
+            return self.im.entropy((0, 0), mask.im)
+        if self.mode in ("I", "F"):
+            if extrema is None:
+                extrema = self.getextrema()
+            return self.im.entropy(extrema)
+        return self.im.entropy()
 
     def offset(self, xoffset, yoffset=None):
         raise NotImplementedError("offset() has been removed. "
--- a/src/_imaging.c
+++ b/src/_imaging.c
@@ -1175,59 +1175,62 @@ _getpixel(ImagingObject* self, PyObject* args)
     return getpixel(self->image, self->access, x, y);
 }
 
+#define HISTOGRAM_METHOD_PROLOGUE(HISTO)                                            \
+    ImagingHistogram HISTO;                                                         \
+    union {                                                                         \
+        UINT8 u[2];                                                                 \
+        INT32 i[2];                                                                 \
+        FLOAT32 f[2];                                                               \
+    } extrema;                                                                      \
+    void* ep;                                                                       \
+    int i0, i1;                                                                     \
+    double f0, f1;                                                                  \
+                                                                                    \
+    PyObject* extremap = NULL;                                                      \
+    ImagingObject* maskp = NULL;                                                    \
+    if (!PyArg_ParseTuple(args, "|OO!", &extremap, &Imaging_Type, &maskp))          \
+    return NULL;                                                                    \
+                                                                                    \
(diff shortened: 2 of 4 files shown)


src/PIL/Image.py
```diff
@@ -1373,6 +1373,35 @@ def histogram(self, mask=None, extrema=None):
...
+
+        A bilevel image (mode "1") is treated as a greyscale ("L") image
+        by this method.
+
+        If a mask is provided, the method returns a histogram for those
+        parts of the image where the mask image is non-zero. The mask
+        image must have the same size as the image, and be either a
+        bi-level image (mode "1") or a greyscale image ("L").
+
+        :param mask: An optional mask.
```
radarhere: Would you be interested in also adding a description of the 'extrema' parameter?

fish2000:
The tests are failing, due to float-comparison issues… I will rewrite the test code to deal with that and re-push.

src/PIL/Image.py
fish2000: Indeed yes!… descriptions for `extrema` have been added for both `image.histogram()` and `image.entropy()`… the latest push has them.

fish2000:
I’m on Mac OS X, and I have no idea how to fix the Appveyor build scripts – can someone who knows windows help me out, or point me in the right direction?

Evidently it needs to link the C math library when linking the result of compiling `_imaging.c` in order to find the symbol for the `_log2` function call.

radarhere:
I have created https://github.com/fish2000/Pillow/pull/1

homm:
Any reason to implement it in C rather than in Python? I'm comparing with this implementation:

```python
def entr(hist):
    from math import log
    fs = 1.0 / sum(hist)
    fentropy = 0.0
    for h in hist:
        if h != 0:
            h = h * fs
            fentropy += h * log(h, 2)
    return -fentropy
```

For 1024 × 640 image I get:

```
In: %timeit im.histogram()
100 loops, best of 3: 2.04 ms per loop

In: %timeit im.entropy()
100 loops, best of 3: 2.07 ms per loop

In: %timeit entr(im.histogram())
100 loops, best of 3: 2.31 ms per loop
```

The bigger image, the smaller the difference between Python and C implementation. Does it worth it?

src/_imaging.c
```diff
@@ -1175,59 +1178,62 @@ _getpixel(ImagingObject* self, PyObject* args)
     return getpixel(self->image, self->access, x, y);
 }
 
+#define HISTOGRAM_METHOD_PROLOGUE(HISTO)                                            \
```
homm: An internal function will look better than macros.

fish2000: I had used a macro to be as minimally invasive toward the existent code (without resorting to copy-pasting).

I am happy to refactor it as e.g. a `static` function resident in the `_imaging.c` TU, tho – can you do me a favor, and point me towards an exemplary internal-use function?

fish2000:
> The bigger image, the smaller the difference between Python and C implementation. Does it worth it?

Short answer: yes, I believe so.

Many histogram-entropy functions are implemented in Python, using the PIL/Pillow `histogram()` method data – e.g. 

* The [ever-popular `easy-thumbnails`](https://github.com/SmileyChris/easy-thumbnails/blob/c611c5c5e3542fbf99c75404edd71f67d3e34ec5/easy_thumbnails/utils.py#L18-L28) Django module;
* My own version, in [my `instakit` package](https://github.com/fish2000/instakit/blob/44893861972b16e279ad0fe3f9dc1f84b39004b4/instakit/processors/squarecrop.py#L11-L23);
* A [popular snippet](https://snipplr.com/view/20341/image-entropy-from-pil-histogram/), freely available;
* The [reference implementation from SciPy](https://github.com/scipy/scipy/blob/v1.2.0/scipy/stats/_distn_infrastructure.py#L2515-L2556);
* An [attempt to squeeze more speed from the same methodology](https://github.com/astrofrog/fast-histogram/blob/e208d3d6ef52469b39fa31371065394883a433db/fast_histogram/_histogram_core.c#L154-L173), using the NumPy C-API;

One of PIL/Pillow’s primary uses is in web applications – processing images uploaded by users, avatar images, `favicon.ico` source material, minute changes to a minuscule images’ scale calculated on-the-fly, and so forth. In-memory, ad-hoc performance with smaller images is important in these cases.

Here’s the long answer: the case I would make for moving this algorithm into Pillow’s `_imaging.c`, where it can take full advantage of Pillow’s internal structures, is that the version I have submitted heap-allocates only one `PyObject` per invocation. Comparable Python implementations need to create at least one Python object for each entry in the histogram – 768 for an `RGB`-mode image – and add at least one frame to the interpreter stack, in the best possible scenario.

Using a C function, utilizing the Pillow internal structures (e.g. `ImagingHistogramInstance`) in an application where the function is iteratively called across 1-dimensional image slices, makes a big difference in the context of a web application, with their attendant expectations of efficient concurrency. That’s my take on it, at any rate, and my original motivation, basically. Yes!

src/_imaging.c
fish2000: @homm sweet! I can rewrite this and push a new change - I will most likely make a new pull request for it. Thank you!

fish2000:
This PR has been superseded by #3608, which uses @homm's static extrema-parsing function code instead of a prologue macro.

src/_imaging.c
fish2000: @homm See PR #3608, which incorporates this code. Thanks again!

PR #3541, opened 2019-01-04
Simplified has_ghostscript

radarhere:


hugovk:
I find the original a bit easier to read than having to think about two `is not falsey` and then an `else` negating one of the `is not`s.

radarhere:
Okay.

PR #3545, opened 2019-01-05
Avoid potential downscaling artifacts by cropping

### LINKED ISSUE #3419: .draft on certain jpegs create incorrect color line at border
### What did you do?
Found this issue posted on chromium bug tracker https://bugs.chromium.org/p/chromium/issues/detail?id=890745 and related [issue reported to libjpeg-turbo](https://github.com/libjpeg-turbo/libjpeg-turbo/issues/297) and obviously reproduced in Pillow. Both links have good explanation of what's going on and original image.

```python
from PIL import Image
im = Image.open('1.jpg')
im.draft('RGB', (75, 54))
# <PIL.JpegImagePlugin.JpegImageFile image mode=RGB size=150x108 at 0x7F46264B73D0>
im.save('2.jpg')
```
Scaled image with the line:

I tried with libjpeg-turbo and libjpeg62, they both show same issue. 
I know that the problem is coming from how libjpeg handles scaling, and it's a feature, so I just want to have your attention and you might want to workaround issue as chromium team did, basically disabling libjpeg scaling for images with dimensions not multiple of MCU size.
Or do you think that users of Pillow should do that themselves? 

@homm

radarhere:
Resolves #3419 

The problem in that issue is that JPEGs are composed of MCUs (Minimum Coded Units). If a JPEG image is not made up of a whole number of MCUs, the encoder must fill in the leftover part of the last MCUs with something - it is not reliable what. When decoding the image, downscaling can make this unreliable image data become visible.

The docstring for `draft` states that it ['returns a version of the image that as closely as possible matches the given mode and size'](https://github.com/python-pillow/Pillow/blob/2e17e0f618eb8b9e3decd0a586e79cdfdbea4baf/src/PIL/Image.py#L1154-L1156).

Since accuracy of size is by no means guaranteed with when using `draft`, my suggestion to resolve the potential artifacts is to crop them out.

### DIFF
--- a/Tests/test_file_jpeg.py
+++ b/Tests/test_file_jpeg.py
@@ -581,6 +581,11 @@ class TestFileJpeg(PillowTestCase):
         # OSError for unidentified image.
         self.assertEqual(im.info.get("dpi"), (72, 72))
 
+    def test_partial_mcu_draft(self):
+        im = Image.open("Tests/images/flower2.jpg")
+        im.draft(im.mode, (100, 100))
+        self.assertEqual(im.size, (146, 109))
+
 
 @unittest.skipUnless(sys.platform.startswith('win32'), "Windows only")
 class TestFileCloseW32(PillowTestCase):
--- a/src/PIL/JpegImagePlugin.py
+++ b/src/PIL/JpegImagePlugin.py
@@ -385,6 +385,7 @@ class JpegImageFile(ImageFile.ImageFile):
             a = mode, ""
 
         if size:
+            original_size = self.size
             scale = min(self.size[0] // size[0], self.size[1] // size[1])
             for s in [8, 4, 2, 1]:
                 if scale >= s:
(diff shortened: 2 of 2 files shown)


kkopachev:
This seems like a braking change. 
Since `.draft` is used in `thumbnail`, which is supposed to preserve aspect ratio. 

```python
from PIL import Image
im = Image.open('132776_a_657140.jpg')
im.thumbnail((300, 215))
im.save('2.jpg')
```

Using original image from the chromium ticket (600x430) this creates a 300x215 image, but the image itself is distorted. `thumbnail` method calculates target size based on `im.size`. In this case target size is exactly twice smaller then original - 300x215, then it calls `draft`, which internally reduces image size to `300x211`, then it calls `resize` method, which scales image back up to 215 height.
 

hard to tell, but right image looks a little more stretched on height.

radarhere:
Okay, fair enough, I hadn't considered the use of `draft` within `thumbnail`.

PR #3551, opened 2019-01-05
Remove unused C hashtable functions

### LINKED ISSUE #3546: Suspicious expression in C code
Is the expression [`if (h->keyDestroyFunc || h->keyDestroyFunc) {`](https://github.com/python-pillow/Pillow/blob/master/src/libImaging/QuantHash.c#L317) in `QuantHash.c` correct? Should it be `if (h->keyDestroyFunc || h->valDestroyFunc) {`?

hugovk:
Fixes #3546.

Alternative to and closes https://github.com/python-pillow/Pillow/pull/3550.

 * https://github.com/python-pillow/Pillow/issues/3546 points out a suspicious expression which checks the same thing twice:
   `if (h->keyDestroyFunc || h->keyDestroyFunc) {`

 * One option is to fix it. See PR #3550 for:
   `if (h->keyDestroyFunc || h->valDestroyFunc) {`

 * However, the expression can never be True (and [isn't covered by tests](https://codecov.io/gh/python-pillow/Pillow/src/7447b44edfe152a8c6857f417f85219a3c450625/src/libImaging/QuantHash.c#L318)), because the only place these functions are set are in `hashtable_set_value_destroy_func` and `hashtable_set_key_destroy_func`

* These functions are never called, and aren't exposed to the Python layer, so can be removed

* There's a number of other unused and unexposed functions in this file which can be likewise removed

* Helps towards #722: QuantHash.c coverage 59% -> 84%

---

The file header says it was created in 1998, and this line has been there since at least [PIL 1.1.1](https://github.com/hugovk/PIL/blame/master/libImaging/QuantHash.c#L345) (released in 2000), so we can use [`Hasn't worked in 20 years`](https://github.com/python-pillow/Pillow/labels/Hasn%27t%20worked%20in%2020%20years) label!

### DIFF
--- a/src/libImaging/QuantHash.c
+++ b/src/libImaging/QuantHash.c
@@ -182,34 +182,6 @@ static int _hashtable_insert(HashTable *h,HashKey_t key,HashVal_t val,int resize
    }
 }
 
-static int _hashtable_lookup_or_insert(HashTable *h,HashKey_t key,HashVal_t *retVal,HashVal_t newVal,int resize) {
-   HashNode **n,*nv;
-   HashNode *t;
-   int i;
-   uint32_t hash=h->hashFunc(h,key)%h->length;
-
-   for (n=&(h->table[hash]);*n;n=&((*n)->next)) {
-      nv=*n;
-      i=h->cmpFunc(h,nv->key,key);
-      if (!i) {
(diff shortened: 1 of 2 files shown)


hugovk:
I prefer this to #3550.

PR #3552, opened 2019-01-06
Removed distutils.version from TIFF

radarhere:
Resolves #3547

PR #3561, opened 2019-01-08
Fix MicImagePlugin unclosed file warning

radarhere:
Note that the warning only appears for Python 3.4.

### DIFF
--- a/src/PIL/MicImagePlugin.py
+++ b/src/PIL/MicImagePlugin.py
@@ -88,7 +88,9 @@ class MicImageFile(TiffImagePlugin.TiffImageFile):
         except IndexError:
             raise EOFError("no such frame")
 
+        prev_fp = self.fp
         self.fp = self.ole.openstream(filename)
+        prev_fp.close()
 
         TiffImagePlugin.TiffImageFile._open(self)
 
--- a/Tests/test_file_mic.py
+++ b/Tests/test_file_mic.py
@@ -54,6 +54,9 @@ class TestFileMic(PillowTestCase):
         self.assertRaises(EOFError, im.seek, 99)
         self.assertEqual(im.tell(), 0)
 
+    def test_unclosed_file(self):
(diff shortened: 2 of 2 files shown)


hugovk:
Does that mean it's not a problem on non-3.4?

radarhere:
This is a strange one.

You can see in #3552 that my first commit failed - it reports an unclosed file in test_file_mpo.py, which seems completely unrelated to the code changes. Inspecting the error, the unclosed file in question is a MIC file, which doesn't make sense either. My second commit there is the code change in this PR, which fixes the PR.

When testing this in Python 3.4, if this new test is failing with the master branch, and then you remove `test_invalid_file`, which should be separate and doesn't use the same files, the new test starts passing.

So I can't say anything definitely - without this code change, the new test fails in Python 3.4, but not in other Python versions.

src/PIL/MicImagePlugin.py
```diff
@@ -88,7 +88,9 @@ def seek(self, frame):
         except IndexError:
             raise EOFError("no such frame")
 
+        prev_fp = self.fp
```
jdufresne: The `__init__` method does the following:

```py
self.__fp = self.fp
...
self.seek(0)
```

So, this will always result in `self.__fp` being closed immediately. At that point, is the `self.__fp` attribute useful?

---

Why not just close the file on this line instead of using the `prev_fp` temporary variable?

radarhere: I don't think `__fp` is used even in the current master. I imagine that if we ever get a Mic test file with multiple frames, it will become useful, and code would change.

---

Closing the file pointer before `openstream` results in `ValueError: seek of closed file`.

radarhere:
Closing, in light of #3581

radarhere:
Just for reference, I will note that this is one of the Python 3.4 exceptions mentioned in #2786

PR #3566, opened 2019-01-10
Added I;16 PNG save

### LINKED ISSUE #2970: Saving grayscale 16 bit PNG fails
### What did you do?
Trying to save a grayscale 16 bit PNG.

Obtain a 16 bit PNG file.

### What actually happened?
An error.

Pillow 5.00
Python 3.5.4

Code:
```python
import numpy as np
from PIL import Image

ar = np.ones((32, 32), dtype=np.uint16)
im = Image.fromarray(ar)
im.save('foo.png')
```
Output:

> OSError: cannot write mode I;16 as PNG

radarhere:
Resolves #2970


Tests/test_file_png.py
```diff
@@ -103,6 +103,9 @@ def test_sanity(self):
         hopper("I").save(test_file)
         Image.open(test_file)
 
+        hopper("I;16").save(test_file)
+        Image.open(test_file)
```
hugovk: This is just a sanity test that it can save and re-open with no exceptions. 

Is it worth adding a new test method to actually check the file is saved and loaded correctly?

hugovk: @radarhere ^ What do you think?

radarhere: Okay, I've changed the tests, see what you think.

Under the hood, this is both I was being converted to I;16B. I am now also converting I;16 to I;16B. The downside is that when they are reloaded, they are both reloaded as I, since there is no difference.

radarhere:
The docs don't list the modes at the different bit levels, so no. I have added another commit though that inclues LA and I as possible modes in the docs, since they were missing.

PR #3568, opened 2019-01-11
Make ContainerIO.isatty() return a bool, not int

jdufresne:
Better follows the interface of `IOBase.isatty`:

https://docs.python.org/3/library/io.html#io.IOBase.isatty

### DIFF
--- a/Tests/test_file_container.py
+++ b/Tests/test_file_container.py
@@ -16,7 +16,7 @@ class TestFileContainer(PillowTestCase):
         im = hopper()
         container = ContainerIO.ContainerIO(im, 0, 0)
 
-        self.assertEqual(container.isatty(), 0)
+        self.assertFalse(container.isatty())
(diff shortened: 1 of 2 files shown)


Tests/test_file_container.py
```diff
@@ -16,7 +16,7 @@ def test_isatty(self):
         im = hopper()
         container = ContainerIO.ContainerIO(im, 0, 0)
 
-        self.assertEqual(container.isatty(), 0)
+        self.assertIs(container.isatty(), False)
```
hugovk: I know `assertIs` also checks the value is `False` by identity, but perhaps `assertFalse` is good enough here?

```suggestion
        self.assertFalse(container.isatty())
```

jdufresne: Switched to `assertFalse` in the latest revision.

PR #3573, opened 2019-01-13
Python del is a statement not a function

jdufresne:
https://docs.python.org/3/reference/simple_stmts.html#the-del-statement


hugovk:
Thank you.

PR #3575, opened 2019-01-13
Use bare 'raise' to re-raise exceptions

jdufresne:
Slightly more concise and nicer syntax


radarhere:
This code was originally added in https://github.com/python-pillow/Pillow/commit/5a8fd796e9e4a6ec9d705c837f992baca0753d90

radarhere:
I have created PR #3583 to revert this.

Issue #2818, opened 2017-10-30
Switching to context managers in Tests


bjd2385:
Looking through `Tests`, I see a lot of places where I would typically use a context manager. But is it worth a quick PR to change the styling?

For example, I see a lot of places that use `im = Image.open(some_file); im.load()` instead of 

```python
with Image.open(some_file) as im:
    im.load()
    ...
```

I know it's not a huge concern, but I also see that these are almost never followed by `im.close()`, which the CM would implicitly call.

By the way, I wrote a quick Bash script to output all of the instances in `Tests` that I'm talking about.

```
Tests/32bit_segfault_check.py
Tests/bench_cffi_access.py
Tests/bench_get.py
Tests/check_fli_overflow.py
        im = Image.open(TEST_FILE)
Tests/check_icns_dos.py
Tests/check_imaging_leaks.py
Tests/check_j2k_dos.py
Tests/check_j2k_leaks.py
Tests/check_j2k_overflow.py
Tests/check_jpeg_leaks.py
Tests/check_large_memory_numpy.py
Tests/check_large_memory.py
Tests/check_libtiff_segfault.py
            im = Image.open(TEST_FILE)
Tests/check_png_dos.py
            im = Image.open(TEST_FILE)
            im = Image.open(TEST_FILE)
            im2 = Image.open(b)
Tests/check_webp_leaks.py
Tests/createfontdatachunk.py
Tests/helper.py
Tests/import_all.py
Tests/make_hash.py
Tests/README.rst
Tests/test_000_sanity.py
Tests/test_binary.py
Tests/test_bmp_reference.py
                im = Image.open(f)
                im = Image.open(f)
                im = Image.open(f)
                compare = Image.open(get_compare(f))
Tests/test_box_blur.py
Tests/test_decompression_bomb.py
Tests/test_features.py
Tests/test_file_bmp.py
        reloaded = Image.open(outfile)
        reloaded = Image.open(output)
        reloaded = Image.open(output)
Tests/test_file_bufrstub.py
        im = Image.open(TEST_FILE)
        im = Image.open(TEST_FILE)
Tests/test_file_container.py
Tests/test_file_cur.py
        im = Image.open(TEST_FILE)
Tests/test_file_dcx.py
        im = Image.open(TEST_FILE)
        im = Image.open(TEST_FILE)
        im = Image.open(TEST_FILE)
        im = Image.open(TEST_FILE)
        im = Image.open(TEST_FILE)
[...]

PR #3577, opened 2019-01-13
Improve handling of file resources

jdufresne:
Better follow Python's file object semantics. User code is responsible for closing resources (usually through a context manager) in a deterministic way.

To achieve this, remove `__del__` functions. These functions used to close open file handlers in an attempt to silence Python `ResourceWarnings`. However, using `__del__` has the following drawbacks:

- `__del__` isn't called until the object's reference count reaches 0. Therefore, resource handlers remain open or in use longer than necessary.

- The `__del__` method isn't guaranteed to execute on system exit. See the Python documentation:

  https://docs.python.org/3/reference/datamodel.html#object.__del__

  > It is not guaranteed that `__del__()` methods are called for objects that still exist when the interpreter exits.

- Exceptions that occur inside `__del__` are ignored instead of raised. This has the potential of hiding bugs. This is also in the Python documentation:

  > Warning: Due to the precarious circumstances under which `__del__()` methods are invoked, exceptions that occur during their execution are ignored, and a warning is printed to sys.stderr instead.

Instead, always close resource handlers when they are no longer in use. This will close the file handler at a specified point in the user's code and not wait until the interpreter chooses to. It is always guaranteed to run. And, if an exception occurs while closing the file handler, the bug will not be ignored.

Now, when code receives a `ResourceWarning`, it will highlight an area that is mishandling resources. It should not simply be silenced, but fixed by closing resources with a context manager.

All warnings that were emitted during tests have been cleaned up. To enable warnings, I passed the `-Wa` CLI option to Python. This exposed some mishandling of resources in `ImageFile.__init__()` and
`SpiderImagePlugin.loadImageSeries()`, they too were fixed.

### DIFF
--- a/Tests/test_file_dcx.py
+++ b/Tests/test_file_dcx.py
@@ -1,6 +1,8 @@
+import unittest
+
 from PIL import DcxImagePlugin, Image
 
-from .helper import PillowTestCase, hopper
+from .helper import PillowTestCase, hopper, is_pypy
 
 # Created with ImageMagick: convert hopper.ppm hopper.dcx
 TEST_FILE = "Tests/images/hopper.dcx"
@@ -11,18 +13,34 @@ class TestFileDcx(PillowTestCase):
         # Arrange
 
         # Act
-        im = Image.open(TEST_FILE)
+        with Image.open(TEST_FILE) as im:
 
-        # Assert
-        self.assertEqual(im.size, (128, 128))
-        self.assertIsInstance(im, DcxImagePlugin.DcxImageFile)
-        orig = hopper()
-        self.assert_image_equal(im, orig)
+            # Assert
+            self.assertEqual(im.size, (128, 128))
+            self.assertIsInstance(im, DcxImagePlugin.DcxImageFile)
+            orig = hopper()
+            self.assert_image_equal(im, orig)
 
+    @unittest.skipIf(is_pypy(), "Requires CPython")
     def test_unclosed_file(self):
         def open():
             im = Image.open(TEST_FILE)
             im.load()
+
+        self.assert_warning(ResourceWarning, open)
+
+    def test_closed_file(self):
+        def open():
+            im = Image.open(TEST_FILE)
+            im.load()
+            im.close()
+
+        self.assert_warning(None, open)
+
+    def test_context_manager(self):
+        def open():
+            with Image.open(TEST_FILE) as im:
+                im.load()
 
         self.assert_warning(None, open)
 
@@ -32,34 +50,34 @@ class TestFileDcx(PillowTestCase):
 
     def test_tell(self):
         # Arrange
-        im = Image.open(TEST_FILE)
+        with Image.open(TEST_FILE) as im:
 
-        # Act
-        frame = im.tell()
+            # Act
+            frame = im.tell()
 
-        # Assert
-        self.assertEqual(frame, 0)
+            # Assert
+            self.assertEqual(frame, 0)
 
     def test_n_frames(self):
-        im = Image.open(TEST_FILE)
-        self.assertEqual(im.n_frames, 1)
-        self.assertFalse(im.is_animated)
+        with Image.open(TEST_FILE) as im:
+            self.assertEqual(im.n_frames, 1)
+            self.assertFalse(im.is_animated)
 
     def test_eoferror(self):
-        im = Image.open(TEST_FILE)
-        n_frames = im.n_frames
+        with Image.open(TEST_FILE) as im:
+            n_frames = im.n_frames
 
-        # Test seeking past the last frame
-        self.assertRaises(EOFError, im.seek, n_frames)
-        self.assertLess(im.tell(), n_frames)
+            # Test seeking past the last frame
+            self.assertRaises(EOFError, im.seek, n_frames)
+            self.assertLess(im.tell(), n_frames)
 
-        # Test that seeking to the last frame does not raise an error
-        im.seek(n_frames - 1)
+            # Test that seeking to the last frame does not raise an error
+            im.seek(n_frames - 1)
 
     def test_seek_too_far(self):
         # Arrange
-        im = Image.open(TEST_FILE)
-        frame = 999  # too big on purpose
+        with Image.open(TEST_FILE) as im:
+            frame = 999  # too big on purpose
 
         # Act / Assert
         self.assertRaises(EOFError, im.seek, frame)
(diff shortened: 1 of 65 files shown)


radarhere:
See also #2818

jdufresne:
Would you like me to add a "Fixes: ..." to the commit message?

radarhere:
No, that's not necessary.

jdufresne:
I have rebased and fixed the unused import.

jdufresne:
Rebased :slightly_smiling_face:

radarhere:
#2019 is a memory leak issue that I believe should be fixed on master. My concern with this PR is that it would trigger those memory leaks, deliberately causing a reported issue. However, I imagine that you would hold that this is a problem with the end code, rather than with Pillow. Let me know if I've misunderstood.

jdufresne:
I'm not sure I fully follow how this would produce a memory leak. Sorry, the thread of that issue is a bit hard for me to know exactly where the issue will be. Are you able to provide sample code that you believe will now result in a memory leak? Or is this already captured in a test?

Is it this?

```py
s = 0
for i in range(len(onlyfiles)):
    fname = onlyfiles[i]
    image = PIL.Image.open(fname)
    # Do something
    s += image.size[0]*image.size[1]
```

If so, then yeah. I would argue that the user isn't using this interface correctly and should instead do the following to _deterministically_ close the resource after finished.


```py
s = 0
for i in range(len(onlyfiles)):
    fname = onlyfiles[i]
    with PIL.Image.open(fname) as image:
        # Do something
        s += image.size[0]*image.size[1]
```

If that is still memory leak then I can take a deeper look.

If you think this is too much of a backwards incompatible change, maybe it would make sense to wait until the next major release to include this.

radarhere:
Yes, I'm just discussing the first code sample you provided, and agree that the context manager should resolve the situation.

The next release is actually a major release, so if everyone is fine with this PR, then go for it, I just wanted to make sure that people were aware of this consequence.

hugovk:
Let's first confirm that it really is the case that memory leaks.

And if so, we'll need to communicate clearly the recommended way of doing things (we need to update the docs anyway).

Would it help to announce the change ahead of making it? If so, a major release is also planned for 2020-01-01 (dropping Python 2.7).

jdufresne:
> Let's first confirm that it really is the case that memory leaks.

After rereading that full linked thread, it looks like it was concluded that the leak occurred in Python and has since been fixed. Am I understanding this right? If I understand, then no new action needs to be taken for this PR w/r/t a memory leak. Do you agree?

> And if so, we'll need to communicate clearly the recommended way of doing things (we need to update the docs anyway).

Good call. I have updated the docs in the latest revision. This includes many changes to `open_files.rst`, a paragraph in the release notes, and some updates in the turorial to promote best practices.

> Would it help to announce the change ahead of making it? If so, a major release is also planned for 2020-01-01 (dropping Python 2.7).

I'm open to this, no preference. Please let me know what you'd like and I'll adjust the PR.

hugovk:
Hi, sorry for the delay on this. This obviously didn't make the last Pillow 6.0.0 release.

> > Would it help to announce the change ahead of making it? If so, a major release is also planned for 2020-01-01 (dropping Python 2.7).
> 
> I'm open to this, no preference. Please let me know what you'd like and I'll adjust the PR.

Let's go for this.

Please could you add something to the [deprecations and removals](https://github.com/python-pillow/Pillow/blob/master/docs/deprecations.rst) and any other relevant docs? Can we add some deprecation warnings in code too?

One thing, that page says:

> Deprecated features are only removed in major releases after an appropriate period of deprecation has passed.

If we deprecate now, it's 6 months until Pillow 6.0.0. Is that long enough? I estimate Pillow 7.0.0 will be when Python 3.5 is removed (it's [EOL on 2020-09-13](https://devguide.python.org/#status-of-python-branches)) in the 2020-10-01 release.

---

PS I guess most of the merge conflicts are from formatting with Black, so many can probably be resolved by running Black on this branch.

jdufresne:
Great! I have rebased and fixed all newly discovered warnings. I updated the docs given that 6.0.0 has been released.

---

> Can we add some deprecation warnings in code too?

After thinking about this some more, what do you see a deprecation warning providing that isn't also provided by Python 3's resource warning? I guess this isn't a true backwards incompatible change in the sense that the resultant behavior is the same (CPython closes the file when the reference count hits 0) except it now emits a resource warning (which would otherwise be a deprecation warning). So the end result is the same, just it is a resource warning instead of a deprecation warning. WDYT?

hugovk:
Oops, I made some typos! This is what I meant to write (in **bold**):

> If we deprecate now, it's 6 months until Pillow ~6.0.0~ **7.0.0**. Is that long enough? I estimate Pillow ~7.0.0~ **8.0.0** will be when Python 3.5 is removed (it's [EOL on 2020-09-13](https://devguide.python.org/#status-of-python-branches)) in the 2020-10-01 release.

Let's make a new PR to put the text in the deprecations section saying this is deprecated in 6.1.0, and will be removed in 7.0.0. We merge that new PR for Monday's release. 

And then _this_ PR can be merged during Q4.

Sorry for the mixup!

---

Let's just go for the deprecations in docs. Thank you!

hugovk:
> Let's make a new PR to put the text in the deprecations section saying this is deprecated in 6.1.0, and will be removed in 7.0.0. We merge that new PR for Monday's release.

Please see https://github.com/python-pillow/Pillow/pull/3929.

hugovk:
Any idea why centos_6_amd64 is failing?

radarhere:
The centos failure is an intermittent problem that should be fixed by #4057

Tests/test_file_dcx.py
```diff
@@ -11,19 +14,35 @@ def test_sanity(self):
...
-        self.assertIsInstance(im, DcxImagePlugin.DcxImageFile)
-        orig = hopper()
-        self.assert_image_equal(im, orig)
+            # Assert
+            self.assertEqual(im.size, (128, 128))
+            self.assertIsInstance(im, DcxImagePlugin.DcxImageFile)
+            orig = hopper()
+            self.assert_image_equal(im, orig)
 
+    @unittest.skipIf(sys.version_info.major < 3 or is_pypy, "Requires CPython 3")
```
jdufresne: As Python 2 support is being dropped, should I go ahead and remove this version check now?

nulano: It's already done in #4109.

hugovk: Yeah, go for it.

jdufresne: Done.

PR #3578, opened 2019-01-13
Add linting to the tox matrix

jdufresne:
Allows contributors to easily run the lint stage with a single command: `tox -e lint`. This creates shorter round trips with CI in case of an error.

Update the Travis configuration to use the new environment to run lint.This allows using a single/common source of truth to reduce differences between Travis and local testing.

The tox environment uses the skip_install feature as static analysis doesn't requiring install the Pillow package itself to run. This feature requires tox 1.9 (2015-02-24), so a minimum tox version was added to the configuration.

https://tox.readthedocs.io/en/latest/config.html#conf-skip_install

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,7 +67,7 @@ services:
 install:
   - |
     if [ "$LINT" == "true" ]; then
-      pip install -U check-manifest flake8
+      pip install tox
     elif [ "$DOCKER" == "" ]; then
       .travis/install.sh;
     fi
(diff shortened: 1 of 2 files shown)


hugovk:
I fixed the merge conflict (from dropping Python 3.4 https://github.com/python-pillow/Pillow/pull/3596) and it made the lint step fail, but that was expected as some lint errors had slipped into master. I've fixed those (https://github.com/python-pillow/Pillow/pull/3613), so this is good to merge. Thanks!

PR #3579, opened 2019-01-13
Improve pytest configuration to allow specific tests as CLI args

jdufresne:
The previous test configuration made it difficult to run a single test with the pytest CLI. There were two major issues: 

- The `Tests` directory was not a package. It now includes a `__init__.py` file and imports from other tests modules are done with relative imports.

- `setup.cfg` always specified the `Tests` directory. So even if a specific test were specified as a CLI arg, this configuration would also always include all tests. This configuration has been removed to allow specifying a single test on the command line.

Contributors can now run specific tests with a single command such as:

```
$ tox -e py37 -- Tests/test_file_pdf.py::TestFilePdf.test_rgb
```

This makes it easy and faster to iterate on a single test failure and is very familiar to those that have previously used tox and pytest.

When running tox or pytest with no arguments, they still discover and runs all tests in the Tests directory.

### DIFF
--- a/Tests/bench_cffi_access.py
+++ b/Tests/bench_cffi_access.py
@@ -1,4 +1,4 @@
-from helper import unittest, PillowTestCase, hopper
+from .helper import unittest, PillowTestCase, hopper
 
 # Not running this test by default. No DOS against Travis CI.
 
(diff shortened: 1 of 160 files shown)


hugovk:
I often use `pytest -k test_remap_palette`, and as expected still works.

PR #3583, opened 2019-01-14
Revert "Use bare 'raise' to re-raise exceptions"

radarhere:
Revert #3575

Now that the PR is merged to master, '2.7_with_system_site_packages Trusty' is failing.

Investigating, I find that the following script raises `NameError: name 'first' is not defined` on Python 3, but  `NameError: name 'second' is not defined` on Python 2. So the original code was not identical to the new code.

```
try:
	first
except Exception:
	try:
		second
	except Exception:
		pass
	raise
```

### DIFF
--- a/Tests/helper.py
+++ b/Tests/helper.py
@@ -140,14 +140,14 @@ class PillowTestCase(unittest.TestCase):
                 (msg or '') +
                 " average pixel value difference %.4f > epsilon %.4f" % (
                     ave_diff, epsilon))
-        except Exception:
+        except Exception as e:
             if HAS_UPLOADER:
                 try:
                     url = test_image_results.upload(a, b)
                     logger.error("Url for test images: %s" % url)
                 except Exception:
                     pass
-            raise
+            raise e
 
     def assert_image_similar_tofile(self, a, filename, epsilon, msg=None,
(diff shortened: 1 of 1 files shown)


hugovk:
Any idea why original PR passed?

jdufresne:
Never mind, I see this is in the `Tests` directory. So I think adding a test for it is not necessary.

radarhere:
@hugovk testing, I find that it passes normally because [`PILLOW_TEST_API_KEY` is None](https://github.com/wiredfool/test-image-results/blob/e6ff809f60993753fff9bc3600495bd245a4a84f/client/test_image_results/client.py#L8
), so [`test_image_results.upload` `return`s](https://github.com/wiredfool/test-image-results/blob/e6ff809f60993753fff9bc3600495bd245a4a84f/client/test_image_results/client.py#L29), rather than attempting to upload and generating an error. Presumably, `PILLOW_TEST_API_KEY` is only present on the master branch.

hugovk:
Ah yes, it's set at https://travis-ci.org/python-pillow/Pillow/settings and is only available for master builds, to prevent malicious PRs from doing `echo  $PILLOW_TEST_API_KEY`.

PR #3588, opened 2019-01-17
Change size of MPO image to match frame

### LINKED ISSUE #1631: No n_frames, or bad value for n_frames

radarhere:
Resolves #1631, replaces #3126

MPO files contain multiple images. The images within the current Pillow test suite are 'Multi-Frame Image'. The images in the issue are not, and instead have a 'Large Thumbnail'. This 'Large Thumbnail' does not have same size as the first image.


soupault:
Is this going to be backported?

radarhere:
No. It will be a part of Pillow 6.0.0, to be released in a few days. What's your reason for being interested in it being backported, rather than just waiting for the next release?

soupault:
@radarhere thank you for the input!

>What's your reason for being interested in it being backported

Well, we're typically a bit conservative about the versions of the dependencies (https://github.com/scikit-image/scikit-image/blob/master/requirements/default.txt#L5) 🙂. Anyway, if your development process doesn't imply having LTS, we'll try to figure something out on our side.

PR #3608, opened 2019-01-25
Added an `image.entropy()` method (second revision)

fish2000:
This calculates the entropy for the image, based on the histogram.

Because this uses image histogram data directly, the existing C function underpinning the `image.histogram()` method was abstracted into a static function to parse extrema tuple arguments, and a new C function was added to calculate image entropy, making use of the new static extrema function.

The extrema-parsing function was written by @homm, based on the macro abstraction I wrote, during the discussion of [my first entropy-method pull request](https://github.com/python-pillow/Pillow/pull/3530): https://github.com/python-pillow/Pillow/pull/3530#discussion_r245379871

The original PR had code with issues with `log2` and the use of the standard math library – @radarhere corrected these: https://github.com/fish2000/Pillow/pull/1

The new `image.entropy()` method is based on `image.histogram()`, and will accept the same arguments to calculate the histogram data it will use to assess the entropy of the image.

The algorithm and methodology is based on existing Python code:

* https://github.com/fish2000/instakit/blob/44893861972b16e279ad0fe3f9dc1f84b39004b4/instakit/processors/squarecrop.py#L11-L25

... A test case in the `Tests/` directory, and doctest lines in `selftest.py`, have both been added and checked.

* Adding the `<math.h>` include to `_imaging.c`,
* The addition of an `image.entropy()` method to the `Image` Python class,
* The abstraction of the extrema-parsing logic of of the C function `_histogram` into a static function, and
* The use of that static function in both the `_histogram` and `_entropy` C functions.
* Minor documentation addenda in the docstrings for both the `image.entropy()` and `image.histogram()` methods were also added.

This resolves my earlier version of this feature, PR #3530.

### DIFF
--- a/Tests/test_image_entropy.py
+++ b/Tests/test_image_entropy.py
@@ -2,9 +2,7 @@ from .helper import PillowTestCase, hopper
 
 
 class TestImageEntropy(PillowTestCase):
-
     def test_entropy(self):
-
         def entropy(mode):
             return hopper(mode).entropy()
 
--- a/Tests/test_image_entropy.py
+++ b/Tests/test_image_entropy.py
@@ -1,0 +1,19 @@
+from .helper import PillowTestCase, hopper
+
+
+class TestImageEntropy(PillowTestCase):
+
+    def test_entropy(self):
+
+        def entropy(mode):
+            return hopper(mode).entropy()
+
+        self.assertAlmostEqual(entropy("1"), 0.9138803254693582)
+        self.assertAlmostEqual(entropy("L"), 7.06650513081286)
+        self.assertAlmostEqual(entropy("I"), 7.06650513081286)
+        self.assertAlmostEqual(entropy("F"), 7.06650513081286)
+        self.assertAlmostEqual(entropy("P"), 5.0530452472519745)
+        self.assertAlmostEqual(entropy("RGB"), 8.821286587714319)
+        self.assertAlmostEqual(entropy("RGBA"), 7.42724306524488)
+        self.assertAlmostEqual(entropy("CMYK"), 7.4272430652448795)
+        self.assertAlmostEqual(entropy("YCbCr"), 7.698360534903628)
--- a/src/_imaging.c
+++ b/src/_imaging.c
@@ -86,6 +86,9 @@
 
 #include "py3.h"
 
+#define _USE_MATH_DEFINES
+#include <math.h>
+
 /* Configuration stuff. Feel free to undef things you don't need. */
 #define WITH_IMAGECHOPS /* ImageChops support */
 #define WITH_IMAGEDRAW /* ImageDraw support */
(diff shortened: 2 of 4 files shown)


fish2000:
What’s the best way to re-trigger the Travis CI tests? At first, it showed all tests as having passed – but then one of them later changed to the red, with a failure in a WebP-related test orthogonal to the code in this PR:

WebP Test Failure

Also, should I report whatever that failure was as a bug, if necessary, or is this a transient glitch?

Please pardon my Travis ignorance.

fish2000:
Actually nevermind – closing and re-opening (as suggested by some rando on Stack Overflow) actually fixed the Travis problem. ALL IS WELL!

fish2000:
@homm @radarhere Might I get a review on this, if either of you get a moment?

src/_imaging.c
```diff
@@ -1175,59 +1178,69 @@ _getpixel(ImagingObject* self, PyObject* args)
...
+    ImagingHistogram h;
+    PyObject* list;
+    int i;
+    union hist_extrema extrema;
+    union hist_extrema* ep;
+
+    PyObject* extremap = NULL;
+    ImagingObject* maskp = NULL;
+    if (!PyArg_ParseTuple(args, "|OO!", &extremap, &Imaging_Type, &maskp))
+    return NULL;
```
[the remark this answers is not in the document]
fish2000: I just fixed this indent, as well as one other “`return NULL`” that needed it (see the latest force-push). @radarhere Thanks!!

fish2000:
Hey guys, is something going on with Travis tests? Two Xenial tests have failed due to `import` errors on the test that I wrote – but I couldn’t see any difference with those `import` statements and those existing test modules that have all thoroughly passed.

Two other tests failed due to a GPG signature verification problem that looks like it’s coming from the package manager.

hugovk:
You may need to rebase on master, after https://github.com/python-pillow/Pillow/pull/3579 was merged.

I'll restart those that are from bad GPG verification.

Tests/test_image_entropy.py
```diff
@@ -0,0 +1,23 @@
+from helper import unittest, PillowTestCase, hopper
```
hugovk: And this should help:

```suggestion
from .helper import unittest, PillowTestCase, hopper
```

fish2000:
@hugovk Thanks –– I am never sure when best to rebase. I appreciate the tip!

Tests/test_image_entropy.py
fish2000: Thanks also for this – the imports are now relative in the latest force-push.

fish2000:
Sweet, all the tests are passing!!

src/_imaging.c
```diff
@@ -1242,11 +1255,61 @@ _histogram(ImagingObject* self, PyObject* args)
...
+        sum += h->histogram[idx];
+    }
+
+    /* Next, normalize the histogram data, */
+    /* using the histogram sum value */
+    fsum = (double)sum;
+    fentropy = 0.0;
+    for (idx = 0; idx < length; idx++) {
+        p = (double)h->histogram[idx] / fsum;
+        fentropy += p != 0.0 ? (p * log(p) * M_LOG2E) : 0.0;
```
radarhere: Minor: It wouldn't be better to made this an `if` statement, instead of adding zero under some circumstances?
```suggestion
        if (p != 0.0) {
            fentropy += p * log(p) * M_LOG2E;
        }
```

fish2000: That sounds fine to me – the only question I have is that OK, since I am in no way a code-generation expert: which of the two will produce the most efficient set of operations, _aprés_-compilation?

fish2000: … I actually figured this out using [godbolt](https://godbolt.org) – the `if` statement has less branching and seems to be more generally optimizable by the latest GCC and recent Clangs. Merging!

fish2000:
screen shot 2019-02-06 at 4 01 56 pm

… OK so it looks like `/usr/bin/xvfb-run` segfaulted in one of the `QImage` tests during the Docker Arch run – but otherwise everything else on the latest Travis build-flight passed it would seem, erm yes.

radarhere:
The seg fault was fixed in https://github.com/python-pillow/docker-images/pull/55

fish2000:
@radarhere awesome, it looks like everything re-ran in the green… if you had anything to do with that, I appreciate it, thanks!

fish2000:
Any chance of getting this into version 6.0.0? – q.v. issue #3618 supra.

Tests/test_image_entropy.py
```diff
@@ -0,0 +1,23 @@
...
+        self.assertAlmostEqual(entropy("F"), 7.06650513081286)
+        self.assertAlmostEqual(entropy("P"), 5.0530452472519745)
+        self.assertAlmostEqual(entropy("RGB"), 8.821286587714319)
+        self.assertAlmostEqual(entropy("RGBA"), 7.42724306524488)
+        self.assertAlmostEqual(entropy("CMYK"), 7.4272430652448795)
+        self.assertAlmostEqual(entropy("YCbCr"), 7.698360534903628)
+
+
+if __name__ == '__main__':
+    unittest.main()
```
radarhere: These lines can be removed, in light of #3631

fish2000: Got ’em – they’re out in the latest push, which is rebased off of the latest `master`. Yes!

src/_imaging.c
```diff
@@ -1242,11 +1254,63 @@ _histogram(ImagingObject* self, PyObject* args)
...
+    fentropy = 0.0;
+    for (idx = 0; idx < length; idx++) {
+        p = (double)h->histogram[idx] / fsum;
+        if (p != 0.0) {
+            fentropy += p * log(p) * M_LOG2E;
+        }
+    }
+
+    /* Finally, allocate a PyObject* for return */
+    entropy = PyFloat_FromDouble(-fentropy);
```
radarhere: Minor note, and feel free to ignore - `entropy` doesn't need to be a separate variable. You could just `return PyFloat_FromDouble(-fentropy);` after `ImagingHistogramDelete(h);`

fish2000: I have no trouble with this one either way - but last week my personal laptop was accidentally damaged and I am currently not able to work on this (!) - if you would like this change in there could you please add it? Either way I hope I can still make the 6.0.0 deadline, yes!

hugovk:
Thank you! This will be released on Monday.

Please could you add release notes to https://github.com/python-pillow/Pillow/blob/master/docs/releasenotes/6.1.0.rst#api-additions?

hugovk:
Release notes added in https://github.com/python-pillow/Pillow/pull/3930.

PR #3616, opened 2019-01-28
_util.isPath returns True for pathlib.Path objects

wbadart:
Now, for functions which accept either a path or file object, the
predicate will pass on Paths and not attempt to call .read on them
before opening.

The pathlib module was added in 3.4 but os.path functions did not start
accepting path-like objects until 3.6, so that is the version after
which this implementation is defined.

Added a unit test to make sure isPath accepts Path objects. The unit
test is skipped if python version is not 3.6 or later.

### DIFF
--- a/Tests/test_util.py
+++ b/Tests/test_util.py
@@ -4,8 +4,6 @@ from helper import unittest, PillowTestCase
 
 from PIL import _util
 
-py36 = sys.version_info.major >= 3 and sys.version_info.minor >= 6
-
 
 class TestUtil(PillowTestCase):
 
@@ -39,7 +37,7 @@ class TestUtil(PillowTestCase):
         # Assert
         self.assertTrue(it_is)
 
-    @unittest.skipIf(not py36, 'os.path support for Paths added in 3.6')
+    @unittest.skipIf(not _util.py36, 'os.path support for Paths added in 3.6')
     def test_path_obj_is_path(self):
         # Arrange
         from pathlib import Path
--- a/Tests/test_util.py
+++ b/Tests/test_util.py
@@ -39,10 +39,10 @@ class TestUtil(PillowTestCase):
     def test_path_obj_is_path(self):
         # Arrange
         from pathlib import Path
-        fp = Path('filename.ext')
+        test_path = Path('filename.ext')
 
         # Act
-        it_is = _util.isPath(fp)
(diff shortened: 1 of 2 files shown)


Tests/test_util.py
```diff
@@ -1,7 +1,11 @@
+import sys
+
 from helper import unittest, PillowTestCase
 
 from PIL import _util
 
+py36 = sys.version_info.major >= 3 and sys.version_info.minor >= 6
```
hugovk: This is True for 3.6, 3.7, ... 4.6, 4.7... but False for 4.0, 4.1, ... 4.5.

How about something like sys.version_info[0:2] >= (3, 6)?

src/PIL/_util.py
```diff
@@ -2,13 +2,20 @@
 import sys
 
 py3 = sys.version_info.major >= 3
+py36 = py3 and sys.version_info.minor >= 6
```
hugovk: Same comment here. 

Would it be better to define it once here and import it elsewhere?

Tests/test_util.py
wbadart: I do like that better. Stay tuned, tweak incoming...

Tests/test_util.py
```diff
@@ -35,6 +35,18 @@ def test_is_path(self):
         # Assert
         self.assertTrue(it_is)
 
+    @unittest.skipIf(not _util.py36, 'os.path support for Paths added in 3.6')
+    def test_path_obj_is_path(self):
+        # Arrange
+        from pathlib import Path
+        fp = Path('filename.ext')
```
hugovk: `fp` is often used for file pointers, shall we rename it to something else?

```suggestion
        test_path = Path('filename.ext')
```

Tests/test_util.py
```diff
@@ -35,6 +35,18 @@ def test_is_path(self):
...
+
+        # Act
+        it_is = _util.isPath(fp)
```
hugovk:
```suggestion
        it_is = _util.isPath(test_path)
```

Tests/test_util.py
wbadart: Sure. I only used `fp` to be consistent with the other path/strinType tests, I'm certainly not attached to the name.

PR #3623, opened 2019-01-30
Fix Flake8

hugovk:
There's a new release of Flake8 (3.7.1), including a new release of pyflakes (2.1.0), which brings in some new tests which fail:

https://travis-ci.org/python-pillow/Pillow/jobs/486309539#L224

```
./winbuild/build_dep.py:310:9: E117 over-indented (comment)
```
Simple comment indentation.

```
./Tests/test_file_libtiff.py:196:9: F723 syntax error in type comment 'dummy value'
```
It now checks mypy type comments. This one isn't a real mypy type comment, so change `type:` to `Type:`.

```
./src/PIL/PdfParser.py:396:9: F821 undefined name 'file'
./src/PIL/PdfParser.py:396:9: F821 undefined name 'Union'
```

The `Union` one can be fixed with `from typing import Union`, but that introduces a dependency on the mypy `typing` module. We have issue https://github.com/python-pillow/Pillow/issues/2625 to add typing support, so I've removed it from here.

### DIFF
--- a/Tests/test_file_libtiff.py
+++ b/Tests/test_file_libtiff.py
@@ -205,7 +205,7 @@ class TestFileLibTiff(LibTiffTestCase):
         #     4: "long",
         #     5: "rational",
         #     12: "double",
-        # type: dummy value
+        # Type: dummy value
         values = {2: 'test',
                   3: 1,
                   4: 2**20,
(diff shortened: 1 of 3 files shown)


hugovk:
Coverage dropped because AppVeyor didn't run:

https://ci.appveyor.com/project/Python-pillow/pillow/history

(It passed on my fork: https://ci.appveyor.com/project/hugovk/pillow/builds/21993434)

Closing and re-opening in case it triggers it.

hugovk:
✅ That sorted it.

Issue #3270, opened 2018-07-24
Can't get "icc_profile"


fangcunguolin:
I used the following code to get ICC profile, but ICC returned None. There is no "icc_profile" in im.info
```python
image = Image.open('test.jpg')
icc = image.info.get('icc_profile')
for x in im.info:
      print(x)
```
The JPG picture taken by the camera has "EXIF, DPI, MP" in im.info. After Photoshop saves and embeds ICC, there are "JFIF, jfif_version, DPI, jfif_unit, jfif_density, icc_profile".

I took photos with Nikon and Sony cameras, without any software processing, and did not find ICC profile using Pillow. But when GIMP and Photoshop are opened, you can check the use of Adobe RGB (1998) configuration files.

python 3.6  
Pliiow 5.2

radarhere:
Are you able to upload a sample image here so that we can inspect it?

fangcunguolin:
@radarhere This is the picture I used for testing. This picture was taken directly by Sony a7riii, and was not processed by Photoshop or other software. Thank you！
![_dsc4949](https://user-images.githubusercontent.com/33274131/43180348-33b8b282-900a-11e8-82b8-243c7f4caefd.JPG)

fangcunguolin:
![image](https://user-images.githubusercontent.com/33274131/43180467-eb75fbd2-900a-11e8-93ea-64af9efedb47.png)

bestvibes:
+1 same issue for me, shooting with a Sony A6000 and my file browser shows the same color profile as above but `image.info.get('icc_profile')` returns `None`. Please help!

radarhere:
@bestvibes are you able to provide a sample image so that it can be inspected?

kkopachev:
I've checked the attached image. Exiftool reports there is no embedded ICC profile. 
However, Exif data contains 
```
[EXIF] ColorSpace: Uncalibrated
[EXIF] InteropIndex: R03 - DCF option file (Adobe RGB)
```

[There is an article ](https://ninedegreesbelow.com/photography/embedded-color-space-information.html) explaining that some cameras don't bother embedding color profiles into files, but only indicate in Exif data what profile to be used. Presumably, image editing software should be smart enough to figure that out and save image with Adobe RGB profile embedded, so image is usable.

There is also old bug reported in firefox https://bugzilla.mozilla.org/show_bug.cgi?id=1322030 stating FF does not automatically apply Adobe RGB profile if EXIF contains InteropIndex=R03.

[...]

PR #3625, opened 2019-01-31
Added EXIF class

### LINKED ISSUE #520: Make EXIF plugin
Exif support (labeled experimental since 2003) is somewhat adequate for reading, but there's no clear way to change a value and then write it out.  We currently support writing Exif as byte strings, with no way to make them apart from manually setting up nested `TiffImagePlugin.ImageFileDirectory` instances, then calling save into a file pointer.
-  `Image.getexif()` should exist, and return an Exif object. 
-  The Exif object should export a dictionary interface.
-  The Exif object should have a `toBytes()` method returning the correct bytes.
-  `Exif.load(bytes).toBytes()` should roundtrip. 
-  This should work: 

```python
exif = Exif()
exif['Orientation'] = 1  
im.save('temp.jpg', exif=exif)
```

### LINKED ISSUE #2081: No way to read Parallax tag from stereo MPO files
Stereo MPO files (such as those produced with the Fujifilm Finepix W3 camera) store the parallax value in a tag in the EXIF information of the _second_ image in the MPO file, but it appears that Pillow can only read the EXIF information from the first image at present.

The tag can be manually read via exiftool as follows:

``` bash
$ exiftool DSCF3195.MPO -mpimage2 -b > DSCF3195-r.jpg
$ exiftool  DSCF3195-r.jpg -Parallax
Parallax                        : 5.625
```

I have several sample MPO files available:
http://valen.darkstarsword.net/photos/stereo/misc/originals/

I would like to make use of this tag in my Stereo Cropping Tool ( https://github.com/DarkStarSword/StereoCropper ) to start with the correct parallax value.

radarhere:
Resolves #520

In addition to the methods described in that issue, I have also added `get_ifd()`, for reading specified subdirectories -
* The Interop EXIF subdirectory, to help #3270. Reference documentation -https://www.awaresystems.be/imaging/tiff/tifftags/interoperabilityifd.html
* The Makernote subdirectory, to resolve #2081

---

Regarding #2081 -

MPO files can have EXIF data for separate images (Page 9 of http://www.cipa.jp/std/documents/e/DC-007_E.pdf). This PR changes `im.info["exif"]` to be specific to the current MPO frame.

Within the EXIF data of the second image in files in #2081, there is a Makernote EXIF tag containing parallax data. However, the MakerNote tag does not have consistent structure across MPO images - it is manufacturer specific. See https://www.awaresystems.be/imaging/tiff/tifftags/privateifd/exif/makernote.html and http://www.exiv2.org/makernote.html.

The issue provides Fujifilm MPO images. The specification of the Fujifilm MakerNote tag can be found at http://www.ozhiker.com/electronics/pjmt/jpeg_info/fujifilm_mn.html#Fujifilm_Tags, and the parallax tag value can be found at https://www.sno.phy.queensu.ca/~phil/exiftool/TagNames/FujiFilm.html. I have added one of the images from the issue.

Within the test suite, there are Nintendo MPO images. See https://3dbrew.org/wiki/MPO and http://owl.phy.queensu.ca/~phil/exiftool/TagNames/Nintendo.html for information about their MakerNote tag.

The test added in this PR reads parallax data from both types of files.

### DIFF
--- a/src/PIL/ImageFile.py
+++ b/src/PIL/ImageFile.py
@@ -28,10 +28,17 @@
 #
 
 from . import Image
-from ._util import isPath
+from ._util import isPath, py3
 import io
 import sys
 import struct
+
+try:
+    # Python 3
+    from collections.abc import MutableMapping
+except ImportError:
+    # Python 2.7
+    from collections import MutableMapping
 
 MAXBLOCK = 65536
 
@@ -292,6 +299,12 @@ class ImageFile(Image.Image):
             raise EOFError("attempt to seek outside sequence")
 
         return self.tell() != frame
+
+    def getexif(self):
+        exif = Exif()
+        if "exif" in self.info:
+            exif.load(self.info["exif"])
+        return exif
 
 
 class StubImageFile(ImageFile):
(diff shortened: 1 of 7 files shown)


src/PIL/PngImagePlugin.py
```diff
@@ -696,8 +696,14 @@ def load_end(self):
     def _getexif(self):
         if "exif" not in self.info:
             self.load()
-        from .JpegImagePlugin import _getexif
-        return _getexif(self)
+        if "exif" not in self.info:
+            return None
+        return dict(self.getexif())
+
+    def getexif(self):
```
hugovk: 🚲🏡time: I know #520 suggested it, but is `getexif` of `get_exif` the better name?

aclark4life: @hugovk Depends on what you mean by "better"? If we have some convention of using ``get_`` everywhere, then the latter. Else, perhaps the former. Personally, the first one reads better to me but I'm not entirely sure why.

src/PIL/ImageFile.py
```diff
@@ -672,3 +687,182 @@ def set_as_raw(self, data, rawmode=None):
...
+            self._data.update(ifd)
+            self._ifds[0x8769] = ifd
+
+        # get gpsinfo extension
+        ifd = self._get_ifd_dict(0x8825)
+        if ifd:
+            self._data[0x8825] = ifd
+            self._ifds[0x8825] = ifd
+
+    def toBytes(self, offset=0):
```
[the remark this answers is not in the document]
radarhere: No. Thanks for spotting, I've added a commit to remove it.

src/PIL/PngImagePlugin.py
hugovk: In this class, there's a `getchunks`. Across the codebase there's more `getx` than `get_x`. Let's keep this :)

PR #3631, opened 2019-02-03
Remove unnecessary unittest.main() boilerplate from test files

jdufresne:
With the introduction and use of pytest, it is simple and easy to execute specific tests in isolation through documented command line arguments. Either by specifying the module path or through the `-k
EXPRESSION` argument. There is no longer any need to provide the boilerplate:

```py
if __name__ == '__main__':
    unittest.main()
```

To every test file. It is simply noise.

The pattern remains in test files that aren't named with `test_*` as those files are not discovered and executed by pytest by default.

### DIFF
--- a/Tests/README.rst
+++ b/Tests/README.rst
@@ -13,28 +13,20 @@ Install::
 Execution
 ---------
 
-**If Pillow has been built in-place**
-
 To run an individual test::
 
-    python Tests/test_image.py
+    pytest Tests/test_image.py
 
-Run all the tests from the root of the Pillow source distribution::
+Or::
 
-    pytest -vx Tests
-
-Or with coverage::
-
-    pytest -vx --cov PIL --cov-report term Tests
-    coverage html
-    open htmlcov/index.html
-
-**If Pillow has been installed**
-
-To run an individual test::
-
-    pytest -k Tests/test_image.py
+    pytest -k test_image.py
 
 Run all the tests from the root of the Pillow source distribution::
 
     pytest
+
+Or with coverage::
+
+    pytest --cov PIL --cov-report term
+    coverage html
+    open htmlcov/index.html
--- a/Tests/test_000_sanity.py
+++ b/Tests/test_000_sanity.py
@@ -1,4 +1,4 @@
-from .helper import unittest, PillowTestCase
+from .helper import PillowTestCase
 
 import PIL
 import PIL.Image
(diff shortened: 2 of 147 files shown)


hugovk:
I used to use that to run like:
```console
$ python Tests/test_file_bmp.py
......
Ran 6 tests in 0.098s

OK
```

But not after https://github.com/python-pillow/Pillow/pull/3579: 
```console
$ python Tests/test_file_bmp.py
  File "Tests/test_file_bmp.py", line 1, in <module>
    from .helper import unittest, PillowTestCase, hopper
ModuleNotFoundError: No module named '__main__.helper'; '__main__' is not a package
```

However, this works:

```console
$ pytest Tests/test_file_bmp.py
========================================================== test session starts ===========================================================
platform darwin -- Python 3.7.2, pytest-4.2.0, py-1.5.4, pluggy-0.7.1
rootdir: /Users/hugo/github/Pillow, inifile: pytest.ini
plugins: requests-mock-1.5.2, xdist-1.26.0, parallel-0.0.9, forked-1.0.1, cov-2.6.1, flaky-3.5.3
collected 6 items

Tests/test_file_bmp.py ......                                                                                                      [100%]
===Flaky Test Report===


===End Flaky Test Report===

======================================================== 6 passed in 0.24 seconds ========================================================
```

jdufresne:
Just to be clear, are you suggesting this PR be closed (and perhaps that #3579 be fixed or reveerted) or something else?

Now that pytest has been adopted, I think it makes sense to encourage its canonical use as the main entry point for tests. Having multiple ways to run tests is, IMO, a bit confusing to new contributors. As your examples above shows, it is also a bit of a maintenance problem as different test runners expect different environment conditions. As such, I think using the command `pytest Tests/test_file_bmp.py` should be the documented way to run single test files. WDYT?

hugovk:
I'm not suggesting this should be closed, it was more to note how it's different.

Whilst it would be nice to run via `python ...` with no other dependencies, I think running via `pytest ...` is fine now, and agree it's good to standardise.

I agree we should document it, please could you update https://github.com/python-pillow/Pillow/blob/master/Tests/README.rst?

jdufresne:
Thanks for pointing that out. I've updated the test `README.rst` and was able to simplify it a bit. There is now a single set of instructions that works for both an installed and inplace environment.

PR #3658, opened 2019-02-15
Fast MPO open

Glandos:

 * Multiply the performance of opening an MPO file by two.

Before:
```py
%timeit Image.open('Tests/images/frozenpond.mpo')
1000 loops, best of 5: 1.51 ms per loop
```
After:
```py
%timeit Image.open('Tests/images/frozenpond.mpo')
1000 loops, best of 5: 779 µs per loop
```

I thought about multiple strategies to avoid a double call to `JpegImageFile._open` that is costly in this use case:
* `__init__` with JpegInstanceFile object. This is the cleanest, by very difficult to achieve, since an instance has at least three inherited classes with many attributes that can be initialized.
* Delegate. Same as above. A delegate is nice, but it should be done in `ImageFile` itself, and it renders every method call more costly.
* Downcasting. This is maybe the dirtiest method from a computer science point-of-view. This is also the most efficient and easy to read.

I chose the last, but I'm completely opened to constructive critics.

I've explained some performance figures in https://github.com/saimn/sigal/pull/363. The gain is really obvious when you have a range of images that are recognized as MPO, but the only thing you want is reading their EXIF data.

### DIFF
--- a/src/PIL/MpoImagePlugin.py
+++ b/src/PIL/MpoImagePlugin.py
@@ -46,10 +46,10 @@ class MpoImageFile(JpegImagePlugin.JpegImageFile):
     def _open(self):
         self.fp.seek(0)  # prep the fp in order to pass the JPEG test
         JpegImagePlugin.JpegImageFile._open(self)
-        self._after_jpeg_open(self._getmp())
+        self._after_jpeg_open()
 
-    def _after_jpeg_open(self, mpheader):
-        self.mpinfo = mpheader
+    def _after_jpeg_open(self, mpheader=None):
+        self.mpinfo = mpheader if mpheader is not None else self._getmp()
         self.__framecount = self.mpinfo[0xB001]
         self.__mpoffsets = [mpent['DataOffset'] + self.info['mpoffset']
                             for mpent in self.mpinfo[0xB002]]
--- a/src/PIL/MpoImagePlugin.py
+++ b/src/PIL/MpoImagePlugin.py
@@ -46,7 +46,10 @@ class MpoImageFile(JpegImagePlugin.JpegImageFile):
     def _open(self):
         self.fp.seek(0)  # prep the fp in order to pass the JPEG test
         JpegImagePlugin.JpegImageFile._open(self)
-        self.mpinfo = self._getmp()
+        self._after_jpeg_open(self._getmp())
+
+    def _after_jpeg_open(self, mpheader):
+        self.mpinfo = mpheader
         self.__framecount = self.mpinfo[0xB001]
         self.__mpoffsets = [mpent['DataOffset'] + self.info['mpoffset']
                             for mpent in self.mpinfo[0xB002]]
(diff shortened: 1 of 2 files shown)


radarhere:
Just to throw out another possible method, by no means necessarily any better than your ideas, you could create the new instance using `__new__` and then copy over existing attributes.

```python
mpo_im = MpoImageFile.__new__(MpoImageFile)
mpo_im.copyJpegAttributes(im)
im = mpo_im
```

Putting together a quickly test, I find that it reduces the time taken by about a third.

Glandos:
I tried something like this:
```py
    @staticmethod
    def adopt_copy(jpeg_instance, mpheader=None):
        mpo_im = MpoImageFile.__new__(MpoImageFile)
        mpo_im.__dict__ = jpeg_instance.__dict__.copy()
        mpo_im._after_jpeg_open(mpheader)
        return mpo_im
```
Running `cProfile` on both show that my first version is slightly faster, but the difference is really small compared to other methods called in `_open`.

And the advantage of the latter method is that it is a (shallow in some parts) copy.

I don't know if you thought about something else for `copyJpegAttributes`, but there are a lot of attributes in an `Image` instance, and I don't know if it's a good idea to list them all explicitly.

src/PIL/MpoImagePlugin.py
```diff
@@ -95,6 +98,22 @@ def _close__fp(self):
         finally:
             self.__fp = None
 
+    @staticmethod
+    def adopt(jpeg_instance, mpheader=None):
+        '''
```
hugovk: Minor thing fort consistency:
```suggestion
        """
```

src/PIL/MpoImagePlugin.py
```diff
@@ -95,6 +98,22 @@ def _close__fp(self):
...
+        Transform the instance of JpegImageFile into
+        instance of MpoImageFile.
+        After the call, the JpegImageFile is extended
+        to be an MpoImageFile.
+
+        This is essentially useful when opening a JPEG
+        file that reveals itself as an MPO, to avoid
+        double call to _open.
+        '''
```
hugovk:
```suggestion
        """
```

src/PIL/MpoImagePlugin.py
```diff
@@ -95,6 +98,22 @@ def _close__fp(self):
         finally:
             self.__fp = None
 
+    @staticmethod
+    def adopt(jpeg_instance, mpheader=None):
+        """
+        Transform the instance of JpegImageFile into
+        instance of MpoImageFile.
```
radarhere:
```suggestion
        an instance of MpoImageFile.
```

src/PIL/MpoImagePlugin.py
```diff
@@ -95,6 +98,22 @@ def _close__fp(self):
         finally:
             self.__fp = None
 
+    @staticmethod
+    def adopt(jpeg_instance, mpheader=None):
```
radarhere: At a high level I understand why you might want to give `mpheader` a default value, but at the moment, the second line of `_after_jpeg_open` would just throw a TypeError when given a `None` value. Any thoughts?

Glandos: You're right. So now `_after_jpeg_open` will use a default value too :)

Issue #3641, opened 2019-02-10
CI: Arch build failing


hugovk:
### What did you do?

Test the code on Travis CI.

### What did you expect to happen?

Pass, as it did 4 days ago on master:

* https://travis-ci.org/python-pillow/Pillow/builds/489433692

### What actually happened?

It failed. Here's the next build on master (with only an AppVeyor config change):

* https://travis-ci.org/python-pillow/Pillow/builds/490240861

```
Tests/test_qt_image_fromqpixmap.py::TestFromQPixmap::test_sanity_1 PASSED [ 98%]
Tests/test_qt_image_fromqpixmap.py::TestFromQPixmap::test_sanity_l PASSED [ 98%]
Tests/test_qt_image_fromqpixmap.py::TestFromQPixmap::test_sanity_p PASSED [ 98%]
Tests/test_qt_image_fromqpixmap.py::TestFromQPixmap::test_sanity_rgb PASSED [ 98%]
Tests/test_qt_image_fromqpixmap.py::TestFromQPixmap::test_sanity_rgba PASSED [ 98%]/usr/bin/xvfb-run: line 181:   715 Segmentation fault      (core dumped) DISPLAY=:$SERVERNUM XAUTHORITY=$AUTHFILE "$@" 2>&1
The command "if [ "$LINT" == "true" ]; then
  tox -e lint
elif [ "$DOCKER" == "" ]; then
  .travis/script.sh
elif [ "$DOCKER" ]; then
  # the Pillow user in the docker container is UID 1000
  sudo chown -R 1000 $TRAVIS_BUILD_DIR
  docker run -v $TRAVIS_BUILD_DIR:/Pillow pythonpillow/$DOCKER:$DOCKER_TAG
fi
" exited with 139.
cache.2
store build cache
0.00s1.21snothing changed, not updating cache
Done. Your build exited with 1.
```

Comparing last-pass and first-fail logs, the only notable difference is pytest has upgraded from 4.1.1 to 4.2.0:

![image](https://user-images.githubusercontent.com/1324225/52533795-c6074480-2d41-11e9-8466-50d50bcd5dd4.png)

Possibly related:
* https://bugs.python.org/issue34095

The https://github.com/python-pillow/docker-images builds are still [passing](https://travis-ci.org/python-pillow/docker-images/builds/491227210).

hugovk:
I'm hesitant to merge anything non-trivial until this is fixed. And I think this red build is preventing coverage reports on other PRs. 

[...]

PR #3663, opened 2019-02-18
Performance: Use subclass for parsing EXIF data

Glandos:
The old method (using dict() and _fixup_dict) create clones from IFD_v1, but it read all items when iterating, losing all benefits from the lazyness of IFD.
The new method use a custom subclass that try to prevent all iterations as much as possible, keeping all tags undecoded.

The return value isn't a dict anymore, but a subclass of ImageFileDirectory_v1 that doesn't support tov2(). All decoded tags are formatted as before.

With a test image, I'm going from:
```py
%timeit Image.open("Tests/images/exif-dpi-zerodivision.jpg") 1000 loops, best of 5: 485 µs per loop
```
to
```py
%timeit Image.open("Tests/images/exif-dpi-zerodivision.jpg") 1000 loops, best of 5: 257 µs per loop
```

With a more common JPEG image from my Panasonic TZ10 camera, I'm going from:
```py
%timeit Image.open("/home/photos/P1040526.JPG") 1000 loops, best of 5: 3.24 ms per loop
```
to
```py
%timeit Image.open("/home/photos/P1040526.JPG") 1000 loops, best of 5: 871 µs per loop
```

I hope that even if the return type of `_getexif` is not the same, its compatibility with the previous format will make it merged.

### DIFF
--- a/src/PIL/JpegImagePlugin.py
+++ b/src/PIL/JpegImagePlugin.py
@@ -493,7 +493,7 @@ class ExifImageFileDirectory(TiffImagePlugin.ImageFileDirectory_v1):
         raise NotImplementedError()
 
     def _setitem(self, tag, value, legacy_api):
-        super()._setitem(tag, value, legacy_api)
+        TiffImagePlugin.ImageFileDirectory_v1._setitem(self, tag, value, legacy_api)
         if legacy_api:
             val = self._tags_v1[tag]
             if not isinstance(val, (tuple, bytes)):
@@ -504,7 +504,7 @@ class ExifImageFileDirectory(TiffImagePlugin.ImageFileDirectory_v1):
         if isinstance(value, TiffImagePlugin.ImageFileDirectory_v2):
             self._tags_v1[tag] = value
         else:
-            super().__setitem__(tag, value)
+            TiffImagePlugin.ImageFileDirectory_v1.__setitem__(self, tag, value)
 
     def __getitem__(self, tag):
         if tag not in self._tags_v1:  # unpack on the fly
@@ -526,7 +526,7 @@ class ExifImageFileDirectory(TiffImagePlugin.ImageFileDirectory_v1):
             self._tagdata.update(other._tagdata)
             self.tagtype.update(other.tagtype)
         else:
-            super().update(*args, **kwds)
+            TiffImagePlugin.ImageFileDirectory_v1.update(self, *args, **kwds)
 
 
 def _getexif(self):
--- a/src/PIL/JpegImagePlugin.py
+++ b/src/PIL/JpegImagePlugin.py
@@ -479,7 +479,7 @@ def _fixup(value):
 
 
 class ExifImageFileDirectory(TiffImagePlugin.ImageFileDirectory_v1):
-    '''
+    """
     Specialization for parsing EXIF data:
     - Remove support for v2 to avoid useless computations
     - custom __setitem__ to supports IFD values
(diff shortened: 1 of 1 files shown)


radarhere:
Would you be able to add tests?

Glandos:
Since no external interfaces were modified, I don't see the point to add any tests.

Of course, we could imagine tests for this new class, but they should already be covered by the existing EXIF tests. In fact, existing tests helped me **a lot** to avoid making useless mistakes.

Could you be more precise on which part of the code could be tested?

radarhere:
I would think that adding a new class constitutes a new interface, and so I would suggest adding independent tests for it. Feel free to disagree, but on the other hand, if you think that this class should not be used by end users, then why have the `ExifImageFileDirectory.to_v2` method?

Normally, we would have an automatic report here about the change in code coverage as a result of this PR, so I might also be able to talk about keeping code coverage up generally, but #3641 is preventing that report.

Glandos:
My goal was to return lazy version of the current dict in `_getexif`, with some optimized version of `ImageFileDirectory_v1` that is currently used.

So _normally_, the output class should be transparent. However, the `to_v2` override is here to prevent misuse. I can also simply delete the method, so that it is impossible to call.

I don't know if it's possible (or even desirable) to hide this class declaration.

In case you really want tests for that new class, I will need some help, because there is currently no test for `ImageFileDirectory_v1`, there is only tests for `_getexif` that uses this class.

src/PIL/JpegImagePlugin.py
```diff
@@ -436,18 +436,62 @@ def _getmp(self):
...
+        if len(value) == 1 and not isinstance(value, dict):
+            return value[0]
+    except Exception:
+        pass
+    return value
+
 
-    return {k: _fixup(v) for k, v in src_dict.items()}
+class ExifImageFileDirectory(TiffImagePlugin.ImageFileDirectory_v1):
+    '''
```
```suggestion
    """
```

src/PIL/JpegImagePlugin.py
```diff
@@ -436,18 +436,62 @@ def _getmp(self):
...
+    Specialization for parsing EXIF data:
+    - Remove support for v2 to avoid useless computations
+    - custom __setitem__ to supports IFD values
+    - values are "fixed up" so that 1-size tuples are expanded
+    - custom update() to avoid iterating and expanding non parsed data
+
+    The goal is to use the lazyness of ImageFileDirectory_v1 in _getexif().
+    '''
```
```suggestion
    """
```

src/PIL/JpegImagePlugin.py
```diff
@@ -469,18 +469,62 @@ def _getmp(self):
...
-    return {k: _fixup(v) for k, v in src_dict.items()}
+class ExifImageFileDirectory(TiffImagePlugin.ImageFileDirectory_v1):
+    """
+    Specialization for parsing EXIF data:
+    - Remove support for v2 to avoid useless computations
+    - custom __setitem__ to supports IFD values
+    - values are "fixed up" so that 1-size tuples are expanded
+    - custom update() to avoid iterating and expanding non parsed data
+
+    The goal is to use the lazyness of ImageFileDirectory_v1 in _getexif().
```
hugovk: Typo:
```suggestion
    The goal is to use the laziness of ImageFileDirectory_v1 in _getexif().
```

Glandos:
It seems that the merge of #3625 make this request invalid, but not irrelevant.

I still have the following performance, when opening 1000 times the same JPEG image with a lot of EXIF tags produced by a camera:

Test

It seems that I can adapt this request to the new layout. Do you have any objections on that before I spend some of my spare time on it?

radarhere:
The image link from your last comment no longer works.

I've created PR #4031 as a new version of this PR - see what you think.

Glandos:
> The image link from your last comment no longer works.
I really dislike Github for their obscure retention policy.

> I've created PR #4031 as a new version of this PR - see what you think.
Many, many thanks for taking care of that. It is really different from mine, but the code has changed too, and I think it does the job.

PR #3664, opened 2019-02-18
Cache the result of _getexif()

Glandos:

 * Cache the result of the first call to `_getexif()`

When using Pillow to read EXIF data, the `_getexif()` is called twice per images, since it's already called in `APP()`. Caching the result improves a lot EXIF reading in that case, at the expense of a little bit more memory.

But maybe the `info["exif"]` could be drop in case of a sucessful call to `_getexif()`?

### DIFF
--- a/src/PIL/JpegImagePlugin.py
+++ b/src/PIL/JpegImagePlugin.py
@@ -430,7 +430,11 @@ class JpegImageFile(ImageFile.ImageFile):
         self.tile = []
 
     def _getexif(self):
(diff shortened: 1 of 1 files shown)


hugovk:
Closing as duplicate of https://github.com/python-pillow/Pillow/pull/3498.

Glandos:
Oh. I managed to forgot this.

Sorry for this, and don't take this duplicate as a tentative to get it merged faster.

Issue #3557, opened 2019-01-07
Error opening png file


Fogapod:
### What did you do?
I tried to open image

### What did you expect to happen?
Image to be opened without errors

### What actually happened?
Pillow raised error.
Traceback:
```
File "/usr/local/lib/python3.6/dist-packages/PIL/Image.py", line 915, in convert
    self.load()
  File "/usr/local/lib/python3.6/dist-packages/PIL/ImageFile.py", line 252, in load
    self.load_end()
  File "/usr/local/lib/python3.6/dist-packages/PIL/PngImagePlugin.py", line 680, in load_end
    self.png.call(cid, pos, length)
  File "/usr/local/lib/python3.6/dist-packages/PIL/PngImagePlugin.py", line 140, in call
    return getattr(self, "chunk_" + cid.decode('ascii'))(pos, length)
AttributeError: 'PngStream' object has no attribute 'chunk_eXIf
```

## What are your OS, Python and Pillow versions?

* OS: Ubuntu 18.04.1 LTS x86_64
* Python: 3.6
* Pillow: 5.4.1

## Demonstration code
```python
from PIL import Image


img = Image.open('50767927-ad61c580-128f-11e9-879e-502cf52fc129.png')
img.load()
```

![Image causing troubles](https://user-images.githubusercontent.com/22667809/50767927-ad61c580-128f-11e9-879e-502cf52fc129.png)

edit: Pillow version I use is 5.4.1, not 3.4.1

hugovk:
With:
* macOS High Sierra
* Python 3.7.2

Cannot reproduce with:
* Pillow 3.4.1
* Pillow 5.3.0

Can reproduce with:
* Pillow 5.4.0
* Pillow 5.4.1 (latest)

hugovk:
Git bisect points to 22837c37e279a5fb3fb7b482d81e2c4d44c8cdcc from https://github.com/python-pillow/Pillow/pull/3506:
```
22837c37e279a5fb3fb7b482d81e2c4d44c8cdcc is the first bad commit
commit 22837c37e279a5fb3fb7b482d81e2c4d44c8cdcc
Date:   Mon Dec 24 23:58:19 2018 +1100

    Read textual chunks located after IDAT chunks

:040000 040000 bbc4e38e995ac826609218792b542d89d25ea138 efdb2cef1a1a438728c23dd94d162d23a8579df7 M	Tests
[...]

PR #3674, opened 2019-02-23
Added EXIF support for PNG

### LINKED ISSUE #3644: Support for EXIF in PNG format
Based on http://ftp-osl.osuosl.org/pub/libpng/documents/pngext-1.5.0.html#C.eXIf, ~it looks like there is a push to officially add exif support into the PNG image format~, it looks like they have defined how to include EXIF data in a PNG 'data chunk' (a standard mechanism 'adding extra data' to PNG files).

Given the work EXIF-related started in #3625 (which addresses #520), is it reasonable to consider this feature request for Pillow?

radarhere:
Resolves #3644

The new image for the test suite is taken from #3557. PNG's `_open` method only reads up until the first IDAT chunk. In that image, the eXIf chunk appears after that, meaning that the EXIF data will not be read until the `load` method is called.

Rather than loading the whole image in `_open`, I suggest that `_getexif` loads the image in the EXIF chunk is not yet present. The downside of this is that `im.info["exif"]` is unreliable if the image has not been loaded, but as this is a new feature, that unreliability is not a step backwards at least.


radarhere:
Okay, done. Also, in docs/handbook/image-file-formats.rst I've added a version string and adjusted the wording of my new paragraph.

hugovk:
Thanks!

Issue #3642, opened 2019-02-10
Drop support for Python 2 on January 1, 2020


hugovk:
Python 2.7 reaches EOL on [2020-01-01](https://python2woop.pw/).

* https://devguide.python.org/#branchstatus

This will obviously be a bigger change than dropping other minor versions, so it's a good idea to start thinking about how to proceed and to communicate that sooner rather than later.

Looking at Pillow's PyPI stats, Python 3 accounts for 60% of downloads and is increasing.

![image](https://user-images.githubusercontent.com/1324225/52538433-3e3e2c00-2d7b-11e9-99a2-c0c23b50421c.png)

Here's the pip installs for pillow from PyPI for January 2019:

| category | percent | downloads |
|----------|--------:|----------:|
|        3 |  59.17% | 3,138,659 |
|        2 |  40.03% | 2,123,506 |
| null     |   0.79% |    42,021 |
| Total    |         | 5,304,186 |

Source: `pypistats python_major --last-month pillow`

| category | percent | downloads |
|----------|--------:|----------:|
|      2.7 |  40.01% | 2,122,431 |
|      3.6 |  34.54% | 1,832,073 |
|      3.5 |  11.78% |   625,070 |
|      3.7 |  10.00% |   530,448 |
|      3.4 |   2.75% |   146,034 |
| null     |   0.79% |    42,021 |
|      3.3 |   0.08% |     4,215 |
|      2.6 |   0.02% |     1,075 |
|      3.8 |   0.01% |       753 |
|      3.2 |   0.00% |        66 |
| Total    |         | 5,304,186 |

Source: `pypistats python_minor --last-month pillow`

Pip has started warning that Python 2 support will be [dropped in a future version](https://discuss.python.org/t/packaging-and-python-2/662) and a number of other projects have [pledged to drop it no later than 2020](https://python3statement.org/). The [latest Django 2.x](https://docs.djangoproject.com/en/2.1/faq/install/#what-python-version-can-i-use-with-django) has already dropped it and:

> Django 1.11 is the last version to support Python 2.7. Support for Python 2.7 and Django 1.11 ends in 2020.

What should Pillow do?

Ping @python-pillow/pillow-team and anyone else interested!

hsoft:
Why not now? It's not as if py2 users lose access to pillow. they're just stuck with an old version.

radarhere:
[...]

PR #3682, opened 2019-03-02
Document Python 2.7 will be dropped on 2020-01-01

hugovk:
For https://github.com/python-pillow/Pillow/issues/3642.

 * Document in deprecations.rst and release notes when Python 2.7 will be dropped

    * I suggest we repeat it in future 6.x release notes

    * When we update CHANGES.rst, let's include this at the top

 * Document in installation.rst that Pillow 7.x will drop Python 2.7


radarhere:
What gave you the RST warning?

hugovk:
It's in PyCharm:

image

PR #3683, opened 2019-03-02
Only close original fp in __del__ and __exit__ if original fp is exclusive

### LINKED ISSUE #3676: Some image files are closed overeagerly (.mpo in this case)
### What did you do?

Upload an MPO file into a Django installation.

I expected the image to either be saved or rejected.

### What actually happened?

Django crashed with a `FileNotFoundError` because the file was closed too early for some reason. I'm not sure anymore and I can't find anything relating to this anymore but I think that Pillow opens files with several frames differently (?) and this might cause the crash.

* OS: Ubuntu 18.04 LTS
* Python: 3.5.2 and 3.6.7
* Pillow: 5.4.1

https://github.com/matthiask/mpo-file-test

Here's a repository with an unit test hopefully demonstrating the issue. 

The problem only appears when using django-imagefield because django-imagefield not only verifies the image header but actually creates a scaled version of uploaded images. Some images' headers are fine but because of missing data e.g. at then end resizing still crashes. django-imagefield validates images a bit more but it also has to read the image data for this.

It's probably just an inconsistency in the way file pointers are handled. The `MpoImageFile` has a line setting `self.__fp = self.fp` and later the file is closed only iff `self.__fp == self.fp` (which is always true for some files but not e.g. for JPEG) meaning that Pillow closes a file handle it does not own.

The problem is circumvented for now in django-imagefield by working on a throwaway copy of the file (https://github.com/matthiask/django-imagefield/commit/567c4c235afa0fa3b9474c8099d97bc71dd89335)

cc @radarhere and @hugovk as requested here: https://github.com/python-pillow/Pillow/commit/71b64dcf7317495f413876f94175ada5eb7bb34d#commitcomment-32457451

radarhere:
Resolves #3676

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -578,11 +578,12 @@ class Image(object):
         return self
 
     def __exit__(self, *args):
-        if hasattr(self, "_close__fp"):
-            self._close__fp()
         if (hasattr(self, 'fp') and hasattr(self, '_exclusive_fp')
-           and self.fp and self._exclusive_fp):
-            self.fp.close()
+           and self._exclusive_fp):
+            if hasattr(self, "_close__fp"):
+                self._close__fp()
+            if self.fp:
+                self.fp.close()
         self.fp = None
 
     def close(self):
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -578,8 +578,7 @@ class Image(object):
         return self
 
     def __exit__(self, *args):
-        if (hasattr(self, 'fp') and hasattr(self, '_exclusive_fp')
-           and self._exclusive_fp):
+        if hasattr(self, 'fp') and getattr(self, '_exclusive_fp', False):
             if hasattr(self, "_close__fp"):
                 self._close__fp()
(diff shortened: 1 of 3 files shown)


hugovk:
This has a merge conflict.

radarhere:
Okay, I've resolved the merge conflict.

src/PIL/Image.py
```diff
@@ -578,11 +578,12 @@ def __enter__(self):
         return self
 
     def __exit__(self, *args):
-        if hasattr(self, "_close__fp"):
-            self._close__fp()
         if (hasattr(self, 'fp') and hasattr(self, '_exclusive_fp')
-           and self.fp and self._exclusive_fp):
-            self.fp.close()
+           and self._exclusive_fp):
```
jdufresne: This can be slightly simplified to:

```py
if hasattr(self, 'fp') and getattr(self, '_exclusive_fp', False):
```

radarhere: Thanks, I've added a commit.

PR #3684, opened 2019-03-02
Use the already imported io

fchapoton:
instead of importing again from io (lgtm suggestion)

Fixes # .

 * remove an import from io, as io is globally imported on top of the module
 * 
 *


hugovk:
Thanks!

PR #3691, opened 2019-03-06
Added DIB saving and loading through Image open

### LINKED ISSUE #3689: Add support for dib images
https://www.imagemagick.org/script/formats.php

> DIB is a BMP file without the BMP header.

radarhere:
Resolves #3689

### DIFF
--- a/docs/handbook/image-file-formats.rst
+++ b/docs/handbook/image-file-formats.rst
@@ -21,7 +21,7 @@ Fully supported formats
 BMP
 ^^^
 
-PIL reads and writes Windows and OS/2 BMP files containing ``1``, ``L``, ``P``,
+Pillow reads and writes Windows and OS/2 BMP files containing ``1``, ``L``, ``P``,
 or ``RGB`` data. 16-colour images are read as ``P`` images. Run-length encoding
 is not supported.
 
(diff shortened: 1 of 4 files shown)


radarhere:
Okay, I've pushed a commit to update the release notes and docs/handbook/image-file-formats.rst

docs/handbook/image-file-formats.rst
```diff
@@ -31,6 +31,14 @@ The :py:meth:`~PIL.Image.Image.open` method sets the following
 **compression**
     Set to ``bmp_rle`` if the file is run-length encoded.
 
+DIB
+^^^
+
+PIL reads and writes DIB files. DIB files are similar to BMP files, so see
```
hugovk: Hmm, maybe we should replace PIL with Pillow in this file?

docs/releasenotes/6.0.0.rst
```diff
@@ -32,6 +32,14 @@ Removed deprecated VERSION
 ``VERSION`` (the old PIL version, always 1.1.7) has been removed. Use ``__version__``
 instead.
 
+New File Format
+===============
+
+DIB File Format
```
hugovk: Shall we put this under "API Additions" (or "API Changes?") instead of a new header?

docs/handbook/image-file-formats.rst
radarhere: Okay, done, with the exception of two references to PIL versions - https://github.com/python-pillow/Pillow/blame/master/docs/handbook/image-file-formats.rst#L1139-L1142

docs/releasenotes/6.0.0.rst
radarhere: My thinking was to copy https://github.com/python-pillow/Pillow/blob/master/docs/releasenotes/5.1.0.rst#blp-file-format.

I have moved it to API Additions.

hugovk:
Thank you!

PR #3693, opened 2019-03-06
Added ability to set language for text rendering

iwsfutcmd:

 * Add `language` parameter to all text rendering methods

libraqm allows a language to be specified when formatting text, to indicate to the font that language-specific glyphs and ligatures are required. Pillow previously had no way to access this functionality, this pull request allows for it by implementing libraqm's `raqm_set_language()` function. Note that with `raqm_set_language()`, different chunks of the text string can be specified as different languages, but I wrote this to set the entire string as a single language as it would be difficult to implement in a non-confusing fashion to users and seemed unnecessary as if multiple languages are needed, the user can draw text multiple times.

I've included tests and documentation as well.


docs/reference/ImageDraw.rst
```diff
@@ -255,7 +255,7 @@ Methods
 
     Draw a shape.
 
-.. py:method:: PIL.ImageDraw.ImageDraw.text(xy, text, fill=None, font=None, anchor=None, spacing=0, align="left", direction=None, features=None)
+.. py:method:: PIL.ImageDraw.ImageDraw.text(xy, text, fill=None, font=None, anchor=None, spacing=0, align="left", direction=None, language=None, features=None)
```
hugovk: Please add the new `language` parameter to the end (to this and other methods).

Some people may already be calling them with positional rather than named arguments, and we don't want to break the API for them.

iwsfutcmd: fixed! thanks for the input.

hugovk: You're welcome! Still a few to do: please put `language` after `features` here.

docs/reference/ImageDraw.rst
```diff
@@ -274,6 +274,16 @@ Methods
 
                       .. versionadded:: 4.2.0
 
+    :param language: Language of the text. Different languages may use
```
hugovk: Move this section after `features`

iwsfutcmd: dang, thought I got all of them. anyways, done! thanks again.

iwsfutcmd:
> Please could you add this to the release notes?
> 
>     * https://github.com/python-pillow/Pillow/blob/master/docs/releasenotes/6.0.0.rst

Done! Also noticed that I forgot to implement the language parameter on the multiline text functions, this is now fixed, including documentation.

Also, the AppVeyor CI check seems to randomly fail sometimes for reasons outside of the control of anybody making a pull request (trouble connecting to a server to fetch requirements, etc.) Is there any way to retry such checks without making a new commit?

hugovk:
> Also, the AppVeyor CI check seems to randomly fail sometimes for reasons outside of the control of anybody making a pull request (trouble connecting to a server to fetch requirements, etc.) Is there any way to retry such checks without making a new commit?

I've clicked to rebuild and created https://github.com/python-pillow/Pillow/issues/3710 to improve build robustness.

Thanks, all green now, merging!

iwsfutcmd:
Thanks a bunch for all your support! And this was my first approved pull request!

PR #3699, opened 2019-03-07
Add option to set dither param on quantize

glasnt:
Default the option to `1`, as per original setting

 * Allow `Image.quantize` to supply it's own `dither` keyword argument. 
 * Defaults to `1`, to ensure backward compatibility.

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -1049,7 +1049,7 @@ class Image(object):
                 new_im.info['transparency'] = trns
         return new_im
 
-    def quantize(self, colors=256, method=None, kmeans=0, palette=None):
+    def quantize(self, colors=256, method=None, kmeans=0, palette=None, dither=1):
         """
         Convert the image to 'P' mode with the specified number
         of colors.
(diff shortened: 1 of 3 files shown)


src/PIL/Image.py
```diff
@@ -1049,7 +1049,7 @@ def convert_transparency(m, v):
                 new_im.info['transparency'] = trns
         return new_im
 
-    def quantize(self, colors=256, method=None, kmeans=0, palette=None):
+    def quantize(self, colors=256, method=None, dither=1, kmeans=0, palette=None):
```
hugovk: Please add the new parameter to the end (and in the docstring).

Some people may already be calling it with positional rather than named arguments, and we don't want to break the API for them.

hugovk:
Do you think it would be worthwhile adding a test to check setting the value really works? There are other `quantize` tests here:

* https://github.com/python-pillow/Pillow/blob/master/Tests/test_image_quantize.py

Also it'd be great to add this to the release notes:

* https://github.com/python-pillow/Pillow/blob/master/docs/releasenotes/6.0.0.rst

glasnt:
@hugovk I've tried to setup tests under the assumption that the two dithered images shouldn't match (they don't when I test manually on my own system), but I'm not sure my `assertNotEqual` check is valid (https://travis-ci.org/python-pillow/Pillow/jobs/503433710#L3771). Do you have any suggestions?

radarhere:
Your `assertNotEqual` line is fine - the problem is that you haven't supplied a `palette` argument to `quantize()`, so the line you have changed in `quantize()` is never reached and `dither` is never used. The same thing applies to `test_quantize_no_dither`.

glasnt:
@radarhere Ah. Yes, that would help :) I'll fix that up.

hugovk:
Thank you!

Issue #3543, opened 2019-01-04
Let's use Tidelift's Security Process


aclark4life:
https://tidelift.com/docs/lifting/security-process

aclark4life:
If nothing else, I'd prefer to rely on their domain vs python-pillow.org domain…

hugovk:
Please see PR https://github.com/python-pillow/Pillow/pull/3702.

radarhere:
On https://tidelift.com/lifter/pypi/Pillow/tasks/2614, it asks who should be emailed when a report is received.

![screen shot 2019-03-08 at 9 40 16 am](https://user-images.githubusercontent.com/3112309/53994352-3d37c900-4186-11e9-8429-2849536b09f5.png)

Should we just enter security@python-pillow.org, or is there another suggestion?

aclark4life:
@radarhere Yes, let's keep that address. It currently forwards to @wiredfool and myself, but we can and probably should add the entire team.

radarhere:
Okay. So is this issue resolved then?

PR #3702, opened 2019-03-07
Use Tidelift's security process

radarhere:
See #3543 and https://tidelift.com/lifter/pypi/Pillow/tasks/2614 Option 1


aclark4life:
Thanks @radarhere @hugovk

PR #3703, opened 2019-03-08
Adding an upper limit for blocks_max in _set_blocks_max

jkrshnmenon:

 * Fixing a potential integer overflow in ImagingMemorySetBlocksMax

The integer blocks_max is properly sanitized before being used in an arithmetic operation.
This PR should fix this issue by adding an upper limit to this variable.

If the error message is not appropriate, please do let me know and I will fix it accordingly.

### DIFF
--- a/Tests/32bit_overflow_check.py
+++ b/Tests/32bit_overflow_check.py
@@ -1,0 +1,6 @@
+from PIL import Image
+import sys
+
+
+if sys.maxsize < 2**32:
+    Image.core.set_blocks_max(2**29)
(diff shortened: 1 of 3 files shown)


radarhere:
You've created a separate test file for this, which means that it wouldn't be run in our standard tests. I would suggest adding your test commands below [the test that blocks_max should be greater than 0](https://github.com/python-pillow/Pillow/blob/406d1d4b9a66fb579f7e8a02bdc7a9213d470c28/Tests/test_core_resources.py#L107).

jkrshnmenon:
Oh, my bad. I didn't know how to add the appropriate test-case.
I've added the test command like you suggested. But I'm not sure if this test-case needs to be run for 64-bit systems. 
What would you suggest ?

radarhere:
I've created https://github.com/jkrshnmenon/Pillow/pull/1 as a suggestion.

hugovk:
Thank you.

PR #3705, opened 2019-03-09
Added support for BMP RGBA bitfield compression

### LINKED ISSUE #3444: Support for more bitfields in bmp
Image format: 32bpp BMP with bitfields

I am wondering if supporting bitfields with the following configuration

```python
                     (0xff, 0xff00, 0xff0000, 0xff000000),
```
is in scope

Currently, it is not supported
https://github.com/python-pillow/Pillow/blob/2fa54408d46411be9978e3b7ec6a5c70f44e09ab/src/PIL/BmpImagePlugin.py#L159-L166

The idea is that these bitfields allow one to store a contiguous C ordered `RGBA` array in memory and just dump it tofile. Doing so, I get a nice 20% boost in speed compared to pillow. But if I save my files in this format, PILLOW can't open them.

That said, pillow is kinda ubiquitous and saving in a BMP format that isn't supported by pillow is a little too dangerous for my taste.

I'm glad to submit a PR, but help navigating the Pillow database would be appreciated.

radarhere:
Resolves #3444


radarhere:
The mask modes aren't mentioned anywhere in documentation, so I wouldn't think that it needs including in the release notes.

PR #3708, opened 2019-03-11
Create GIF deltas from background colour of GIF frames if disposal mode is 2

### LINKED ISSUE #3665: Animated, transparent GIFs not calculating diff correctly
function `_write_multiple_frames` in `GifImagePlugin.py` has an error on line 445 ([line 445 in present repo](https://github.com/python-pillow/Pillow/blob/daa8b5133923561d2c6d4d1f7b2c9d9aa233007e/src/PIL/GifImagePlugin.py#L445), line 443 in released version 5.4.1).
Specifically, when the disposal_method is set to '2', it should calculate the diff against the background color, but currently it is calculating the diff against the previous frame
`previous = im_frames[-1]`,
and this produces substantial artifacts in the rendering whenever there are still portions of the image in adjacent frames.

sircinnamon:
Fixes #3665 
 All credit to @Artheau who figured out the method, I just tidied up and made sure it passed the tests+lint

 * Save a "background colour" image from the first frame image, and use that to create gif deltas if disposal mode is 2


radarhere:
Are you able to add tests, so that these changes won't be reverted by future changes?

sircinnamon:
Writing a test has shown me this solution is an improvement but is not complete. There is an issue where the palette of the first frame is forced upon the following frames and so any colours that "show up" later are mismatched. Going to keep working on it - I'll post my test so anyone else can check it out and compare:

```python
def test_dispose2_diff(self):
        out = self.tempfile('temp.gif')
        im_list = [
            Image.new('RGB', (100, 100), '#fff'),
            Image.new('RGB', (100, 100), '#999'),
            Image.new('RGB', (100, 100), '#000'),
        ]
        for img in im_list:
            d = ImageDraw.Draw(img)
            d.ellipse([(0,0),(100,100)], fill='#f00')

        im_list[0].save(
            out,
            save_all=True,
            append_images=im_list[1:],
            disposal=2
            )

        img = Image.open(out)
        top_left_pixels = []
        center_pixels = []

        rgb_img = img.convert('RGB')
        r, g, b = rgb_img.getpixel((1,1))
        top_left_pixels += [(r,g,b)]
        r, g, b = rgb_img.getpixel((50,50))
        center_pixels += [(r,g,b)]

        for i in range(2):
            img.seek(img.tell() + 1)
            r, g, b = rgb_img.getpixel((1,1))
            top_left_pixels += [(r,g,b)]
            r, g, b = rgb_img.getpixel((50,50))
            center_pixels += [(r,g,b)]
            for prev in top_left_pixels[:i]:
                # Check bacground changed every frame
                self.assertNotEqual((r,g,b), prev)
            for prev in center_pixels[:i]:
                # Check center remains the same every frame
                self.assertEqual((r,g,b), prev)
```
This creates a gif with a red circle surrounded but a changing white->grey->black corners filling out the square. then it loads the gif back in and pulls a pixel at the edge and at the center and checks the colours are expected (currently they are not).

sircinnamon:
@radarhere Added a test which very carefully checks that each frame of a gif correctly changes with disposal mode 2 and checks that the colours are right (because the palettes were screwing up).

The test showed that the previous solution was incomplete. Added a little more that forces the `include_color_table` flag for that disposal mode, and also (across all modes) if that flag is set, pad the table to be long enough to pass the `_get_color_table_size` check (less than 3 colours returns a 0 from that, my change just concats the palette with itself until its longer than 3 colours)

radarhere:
The only change required to the source code to made your tests pass is the expanding of the palette. Removing the rest of your changes don't actually cause the new tests to fail, meaning that the tests that you've added don't guard against regression.

sircinnamon:
You're right, I've renamed and slightly modified that test to represent that it's testing the palette and written a new one that tests that unchanged pixels still appear on subsequent frames

sircinnamon:
Is there anything else I can do for this PR or is it waiting on another PR? I'd love to get this merged in :)

hugovk:
@radarhere What do you think, is this good to merge once the conflicts are sorted?

radarhere:
Okay, I resolved the conflicts, pushed some commits, and I'm now happy with the changes to GifImagePlugin.

hugovk:
Thanks both!

PR #3709, opened 2019-03-11
Consistent DPI rounding

### LINKED ISSUE #3323: DPI is rounded wrong by 1
### What did you do?

Take this image

[test_image.zip](https://github.com/python-pillow/Pillow/files/2338635/test_image.zip)

Its info according to `magick`:
```
>magick identify -verbose 08.bmp
Image: 08.bmp
  Format: BMP3 (Microsoft Windows bitmap image (V3))
  Class: DirectClass
  Geometry: 518x244+0+0
  Resolution: 177.17x177.17
  Print size: 2.92375x1.37721
  Units: PixelsPerCentimeter
  Type: TrueColor
  Endianess: Undefined
  Colorspace: sRGB
```
As you can see, the PPC is 177.17, which converts to 177.17*2.54=450.0118 DPI

When use img.info["dpi"], it should show (450,450)

### What actually happened?

It shows (451,451)

Python 3.6.5
Pillow 5.2.0

### Repro code
```python
from PIL import Image

img = Image.open("08.bmp")
print(img.info["dpi"])
```

radarhere:
Resolves #3323 

Whether or not DPI is rounded, or how it is rounded, is inconsistent between image formats.

* JPEG does not round on load, and rounds to the nearest integer on save
* BMP rounds up on load, rounds down on save
* PNG rounds to the nearest integer on load and save
* TIFF does not round on load or save
* WMF rounds down on load

This PR changes all to round to the nearest integer.


hugovk:
Should this include tests?

radarhere:
I've added tests. Some of the added images are hex edited.

PR #3720, opened 2019-03-16
Describe multiframe features for TIFF/GIF

akuchling:
Changes proposed in this pull request, all documentation changes:

* TIFF: describe num_frames and seek()/tell()
* Link methods to PIL.Image class instead of the Python file object
* It doesn't seem like the Spider plug-in actually has an 'nimages' attribute
* Fix markup for exception name
* Remove reference link for two modules that no longer exist

### DIFF
--- a/docs/handbook/image-file-formats.rst
+++ b/docs/handbook/image-file-formats.rst
@@ -667,20 +667,17 @@ numbers are returned as a tuple of ``(numerator, denominator)``.
 
     .. deprecated:: 3.0.0
 
-
 Reading Multi-frame TIFF Images
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 The TIFF loader supports the :py:meth:`~PIL.Image.Image.seek` and
-:py:meth:`~PIL.Image.Image.tell` methods, taking and returning frame
-numbers within the image file.  Frames are numbered from 0 to
-`im.num_frames - 1`.  You can seek to the next frame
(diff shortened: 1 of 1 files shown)


akuchling:
Hm, did I not use the [ci skip] correctly in my commit message?  Or is that no longer required/supported?

radarhere:
You used '[ci skip]' correctly, and Travis and AppVeyor jobs have been skipped. The more recently added Azure jobs are not regarding it.

What makes you say that [FpxImagePlugin](https://github.com/python-pillow/Pillow/blob/master/src/PIL/FpxImagePlugin.py) and [MicImagePlugin](https://github.com/python-pillow/Pillow/blob/master/src/PIL/MicImagePlugin.py) no longer exist?

akuchling:
Re. missing modules: I didn't realize those docs weren't getting built because the olefile module wasn't installed.  I've removed that change from the branch.

hugovk:
Azure Pipelines have started work on adding support for `[ci skip]`:

* https://developercommunity.visualstudio.com/idea/420289/add-support-ci-skip-and-azure-skip-to-azure-pipeli.html?childToView=443074#comment-443074

radarhere:
Thanks for this. Regarding your description of TIFF seeking, you can actually rewind the file by seeking to any previous frame, not just the first one.

I realise that you were just copying the GIF documentation though, so that has been corrected in #3722

PR #3733, opened 2019-03-21
Format code with Black

hugovk:
I'd like to suggest formatting the code with Black:

> By using *Black*, you agree to cede control over minutiae of hand-formatting. In return, Black gives you speed, determinism, and freedom from *pycodestyle* nagging about formatting. You will save time and mental energy for more important matters.
>
> *Black* makes code review faster by producing the smallest diffs possible. Blackened code looks the same regardless of the project you’re reading. Formatting becomes transparent after a while and you can focus on the content instead.

https://black.readthedocs.io/en/stable/the_black_code_style.html

Black intentionally doesn't provides many options, and I've not used them here. The idea is you just run ` black` and it takes care of everything for you.

I added `#fmt:off`/`fmt:on` to retain formatting only where the shape of a list is important, for matrices.

Only the `src` directory has been done here for demonstration, let me know what you think.


jdufresne:
I like it :+1: 

I've seen more and more Python projects adopt black such that it now seems very natural style to me. I think it makes the process easier for new contributors as they won't need to stress or think about the project's coding style. Instead just let the tool handle it.

As a mere infrequent outside contributor, I'm for it.

radarhere:
Provided it is just run automatically over new code, I could see this as being a time saver. On the other hand, will new contributors be confused about why their code has rearranged itself?

Note that you'd also want to add this to [requirements.txt](https://github.com/python-pillow/Pillow/blob/master/requirements.txt)

hugovk:
> Provided it is just run automatically over new code, I could see this as being a time saver. On the other hand, will new contributors be confused about why their code has rearranged itself?

You do need to run the command, it's not in a pre-commit hook or anything (but if they wished they could create one, or add it to their editor on save). But it's  just a `pip install black && black <dir>`. It will fix lots of potential flake8 errors too.

On the CI, `black --check --diff` will show any fixes it would have made in case someone wants to make the changes manually and without the tool.

> Note that you'd also want to add this to [requirements.txt](https://github.com/python-pillow/Pillow/blob/master/requirements.txt)

Added.

hugovk:
@jdufresne Black has made this change (and another similar) in `Image.py`, which becomes too long and fails Flake8's E501. Any ideas how best to resolve it?

```diff
<                             unit_size, handler =\
<                                 TiffImagePlugin.ImageFileDirectory_v2._load_dispatch[
<                                     typ
<                                 ]
---
>                             unit_size, handler = TiffImagePlugin.ImageFileDirectory_v2._load_dispatch[
>                                 typ
>                             ]
```

radarhere:
My suggestion? Change `from . import TiffImagePlugin` to `from .TiffImagePlugin import ImageFileDirectory_v2` a few lines earlier.

hugovk:
Good idea, updated!

hugovk:
If no other comments, let's merge this and I'll make a new PR for the other directories.

radarhere:
Consider merging in #3898 first, before the changes from this PR make a simple revert unviable.

hugovk:
Yep, good idea.

PR #3749, opened 2019-03-26
Fixed deprecation warnings

radarhere:
The [Python 3.8 Travis job](https://travis-ci.org/python-pillow/Pillow/jobs/510918029) has started failing. A test for no unclosed file warning is failing on a deprecation warning that was recently added - https://github.com/python/cpython/pull/12473/commits/0895793194cd5f1153d2a286ed3a4fc876149b16 

This PR fixes the situation.

### DIFF
--- a/src/_imaging.c
+++ b/src/_imaging.c
@@ -71,6 +71,7 @@
  * See the README file for information on usage and redistribution.
  */
 
+#define PY_SSIZE_T_CLEAN
 #include "Python.h"
 
 #ifdef HAVE_LIBJPEG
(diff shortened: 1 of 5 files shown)


hugovk:
Thanks! Good to see the 3.8-dev build has saved us surprises later!

QuLogic:
Some of these look a bit suspicious to me. The types were changed from `int` to `Py_ssize_t`, but the `PyArg_ParseTuple` calls still specify `i`.

radarhere:
@QuLogic isn't that the change of behaviour that the deprecation warning is describing though?

QuLogic:
No, the `#define` only affects types that are `#` (e.g., `y#`); it does not affect `i` (which remains an `int`).

QuLogic:
I've written a patch to change the `PyArg_ParseTuple` calls to `n` where necessary, but I see that only `src/encode.c` was really changed. Seemingly-equivalent variables in `src/decode.c` were not changed, e.g. GIF bits [encode](https://github.com/python-pillow/Pillow/blob/master/src/encode.c#L410) vs. [decode](https://github.com/python-pillow/Pillow/blob/master/src/decode.c#L438).

So would it be better to move to the `n` type specifier or revert some of these to `int`?

PR #3752, opened 2019-03-26
Update Trove licence classifiers

hugovk:
* The "Standard PIL license" has been identified as "Historical Permission Notice and Disclaimer (HPND)" (https://github.com/python-pillow/Pillow/issues/1507#issuecomment-476373096)
 * This has now been added as a PyPI Trove classifier (https://github.com/pypa/warehouse/issues/5627) 

 * Sort the classifiers

 * Replace `License :: Other/Proprietary License` with `License :: OSI Approved :: Historical Permission Notice and Disclaimer (HPND)`

 * Remove the `license` parameter:
   * "this argument is meant for deviations from the Trove classifier, not instead of it" (https://github.com/pypa/packaging.python.org/pull/492)


hugovk:
Pyroma was failing because "Your package does not have license data". When there's a known Trove licence, `licence` can and should be omitted, according to https://github.com/pypa/packaging.python.org/pull/492.

I've reported this to pyroma (https://github.com/regebro/pyroma/issues/38) and relaxed our test in the meantime.

aclark4life:
LGTM, thanks

PR #3761, opened 2019-04-01
Improve exif_transpose

radarhere:
Sequel to #3687, incorporating #3625 

Now that Pillow can modify EXIF information, change the `exif_transpose` method to not just record internally that a transposition has occurred, but actually remove it from the EXIF data that specified it in the first place.

### DIFF
--- a/Tests/test_imageops.py
+++ b/Tests/test_imageops.py
@@ -1,7 +1,6 @@
 from .helper import PillowTestCase, hopper
 
 from PIL import Image
-from PIL import ImageFile
 from PIL import ImageOps
 
 try:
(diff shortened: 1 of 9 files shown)


radarhere:
Since some operations transform an ImageFile instance into an Image, I've added a commit to allow `exif_transpose` to work with Image instances as well.

radarhere:
I've realised that #520 suggested `Image.getexif()`, not `ImageFile.getexif()`. That is better, because after an ImageFile instance is changed into an Image instance by, for example, a `copy()` operation, `getexif()` can still be run. I've added a commit to move the class. Let me know if there are any thoughts otherwise.

PR #3778, opened 2019-04-08
Added ImageSequence all_frames

### LINKED ISSUE #3758: Transforms and enhancements across all frames
It would be nice if transforms and enhancements could be applied on animations.

radarhere:
Resolves #3758 

Adds a new method to apply a supplied function to all frames of an image at once. So, for example,

```python
im = Image.open('Tests/images/g4-multi.tiff')

ims = ImageSequence.all_frames(im, lambda im_frame: im_frame.rotate(180))

ims[0].save('out.tiff', save_all=True, append_images=ims[1:])
```

### DIFF
--- a/docs/releasenotes/6.1.0.rst
+++ b/docs/releasenotes/6.1.0.rst
@@ -11,6 +11,14 @@ An optional ``include_layered_windows`` parameter has been added to ``ImageGrab.
 defaulting to ``False``. If true, layered windows will be included in the resulting
 image on Windows.
 
+ImageSequence.all_frames
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+A new method to facilitate applying a given function to all frames in an image, or to
+all frames in a list of images. The frames are returned as a list of separate images.
+For example, ``ImageSequence.all_frames(im, lambda im_frame: im_frame.rotate(90))``
+could used to return all frames from an image, each rotated 90 degrees.
+
 Variation fonts
 ^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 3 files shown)


pirate486743186:
hmmm.... Maybe add multi-threading support?
The function could take a third parameter, the maximum number of threads to be used, default to one. 

If no time for this, then you could just silently add the third parameter anyway, so that future code is backwards compatible with the first version of the function.

radarhere:
While we are concerned with backwards compatibility, I don't think that we are concerned with forwards compatibility.

Nothing else in the Pillow runtime uses multiprocessing, so I'm not personally inclined to integrate it here. If anyone else feels otherwise, feel free to comment. Multiprocessing can't be applied to the process of splitting up a single image into frames, because it uses the same file pointer, but it could be applied to the supplied function. If a user wanted to do that, they could just -

```python
im = Image.open('Tests/images/iss634.gif')

ims = ImageSequence.all_frames(im)

from multiprocessing import Pool
def f(im_frame):
    return im_frame.rotate(180)
ims = Pool(5).map(f, ims)

ims[0].save('out.gif', save_all=True, append_images=ims[1:])
```

pirate486743186:
but isn't this a special case? You are basically processing multiple images at once. And it seams it would be used frequently IMHO.

docs/releasenotes/6.1.0.rst
```diff
@@ -11,6 +11,14 @@ An optional ``include_layered_windows`` parameter has been added to ``ImageGrab.
 defaulting to ``False``. If true, layered windows will be included in the resulting
 image on Windows.
 
+ImageSequence.all_frames
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+A new method to facilitate applying a given function to all frames in an image, or to
+all frames in a list of images. The frames are returned as a list of separate images.
+For example, ``ImageSequence.all_frames(im, lambda im_frame: im_frame.rotate(90))``
+could used to return all frames from an image, each rotated 90 degrees.
```
hugovk:
```suggestion
could be used to return all frames from an image, each rotated 90 degrees.
```

radarhere: Thanks

PR #3780, opened 2019-04-08
Update Py_UNICODE to Py_UCS4

### LINKED ISSUE #3344: Unable to draw an Ext-B+ char on Windows
### What did you do?

1\. Create a script file `char.py` for drawing.

```python
#!/usr/bin/env python3
"""Draw a text image from a font
"""
import sys
import os
import argparse

def color_hex_to_tuple(color):
    """Convert hex string RRGGBBAA to (R, G, B, A)
    """
    return (int(color[0:2], 16), int(color[2:4], 16), int(color[4:6], 16), int(color[6:8], 16))

def draw_text(text, font, size=40, fgcolor='000000FF', bgcolor='FFFFFF00', output='text.png'):
    if not output:
        return
    if type(size) == str:
        size = int(size)
    if type(fgcolor) == str:
        fgcolor = color_hex_to_tuple(fgcolor)
    if type(bgcolor) == str:    
        bgcolor = color_hex_to_tuple(bgcolor)

    img = Image.new('RGBA', (size, size), bgcolor)
    draw = ImageDraw.Draw(img)
    fnt = ImageFont.truetype(font, size)
    dx, dy = fnt.getoffset(text)
    draw.text((-dx, -dy), text, font=fnt, fill=fgcolor)
    img.save(output)

def main():
    root = os.path.dirname(__file__)

    parser = argparse.ArgumentParser(
        formatter_class=argparse.RawDescriptionHelpFormatter,
        description=__doc__,
        )
    parser.add_argument('text',
        help="""The text to draw.""")
    parser.add_argument('font',
        help="""The font used to draw the text.""")
    parser.add_argument('--size', default=40,
        help="""Image size in pixels. Default: %(default)s""")
    parser.add_argument('--fgcolor', default='000000FF',
        help="""Foreground color in hex RRGGBBAA. Default: %(default)s""")
    parser.add_argument('--bgcolor', default='FFFFFF00',
        help="""Background color in hex RRGGBBAA. Default: %(default)s""")
    parser.add_argument('--output', default=os.path.join(root, 'text.png'),
        help="""Output file path. Default: %(default)s""")
    args = vars(parser.parse_args())
    draw_text(**args)

if __name__ == "__main__":
    main()
```

2\. Download [http://fonts.jp/hanazono/](HanaMin) font and put in the same directory.

3\. Try several ways to draw the text image:

3.1. Run below command on Windows:

    chcp 950
    char.py "人" HanaMinA.ttf

Result (successful): text

3.2. Run below command on Windows:

    chcp 65001
    char.py "人" HanaMinA.ttf

Result (successful): text

3.3. Run below command on Windows:

    chcp 950
    char.py "𠀀" HanaMinB.ttf

Result (bad): text

3.4. Run below command on Windows:

    chcp 65001
    char.py "𠀀" HanaMinB.ttf

Result (bad): text

3.5. Run below command on Linux:

    char.py "𠀀" HanaMinB.ttf

Result (successful): text

3.* should all draw the image successfully.

### What actually happened?
3.3 and 3.4 doesn't draw the image as expected.

PIL.PILLOW_VERSION = 5.2.0

Tested on Windows 7, SP1 and Linux Ubuntu 16.

### LINKED ISSUE #3777: 32-bit Unicode code points don't work on Windows
When rendering with TrueType fonts on Windows, Unicode code points outside the 16-bit range are rendered as a pair of unknown characters (usually a box with a question mark). I suspect this is caused by one of the following lines:

https://github.com/python-pillow/Pillow/blob/36db534c480ad4e1f33e02e15478ed122509f927/src/_imagingft.c#L322

https://github.com/python-pillow/Pillow/blob/36db534c480ad4e1f33e02e15478ed122509f927/src/_imagingft.c#L363

According to Python docs, `Py_UNICODE` is a typedef for `wchar_t`, which is 16-bit on Windows, but typically 32-bit on Linux. The docs suggest using the type `Py_UCS4` and function `PyUnicode_AsUCS4Copy` (introduced in Python 3.3).
https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_AsUCS4Copy

I would attempt changing this myself, but I'm not sure how to compile native Python libraries from source.

### Code to reproduce:
```
import sys

size = int(sys.argv[2]) if len(sys.argv) >= 3 else 48
fontname = sys.argv[1] if len(sys.argv) >= 2 else 'Symbola.ttf'
font = ImageFont.truetype(fontname, size=size)

text = ''' Unicode pictogram test:
Letters: P H BUS
Sans-serif: 𝗣 𝗛 𝗕𝗨𝗦
Crosses: † ✝
Amenity: 🚻 🖂 ⛲ ⛹ 🎓 🎒
Entertainment: 🎨 🎡 🎢 🎜 🎬 🎭 🎰 🎲
Sport: 🎳 🎾 🏓 🏸 🚩
Restaurants: 🍽 🍴 🍵 🍶 🍷 🍺 🍾
Transport: ⛽ 🚂 🚁 🚃 🚋 🚌 🚖 🛪
'''.splitlines()

image = Image.new('RGB', (1000, 600), '#fff')
draw = ImageDraw.Draw(image, 'RGBA')
for i, line in enumerate(text):
    draw.text((0, i * size), line, fill='#000', font=font)
image.save(fontname + '.png')
```
Font used: https://www.fontspace.com/unicode-fonts-for-ancient-scripts/symbola

### Result on Windows (Pillow 5.4.1, Python 3.7.0 32-bit, Windows 7 64-bit):
win-Symbola ttf
### Result on Linux (Pillow 5.4.1, Python 3.4.3, Linux Mint 17.3 64-bit):
linux-Symbola ttf

nulano:
`PyUnicode_AS_UNICODE` is now deprecated, replace it with `PyUnicode_AsUCS4Copy` or `PyUnicode_READ_CHAR`.

- Fixes #3344, #3777 (Python 3.x only).
- Add test for #3344, #3777.

Fixing the issues for Python 2.7 will likely require a lot of effort. Given the coming end of support, I don't think it makes sense to fix it there; therefore the test is set to skip on Windows for Python 2.x.


nulano:
I'm not sure why the new test fails on Centos-6 and Amazon-1. All other platforms pass.

Tests/test_imagefont.py
```diff
@@ -440,6 +440,18 @@ def test_unicode_pilfont(self):
...
+                     "requires Python 3.x on Windows")
+    def test_unicode_render(self):
+        # issue #3777
+        text = u"A\u278A\U0001F12B"
+        ttf = ImageFont.truetype("Tests/fonts/NotoSansSymbols-Regular.ttf",
+                                 FONT_SIZE, layout_engine=self.LAYOUT_ENGINE)
+        img = Image.new("RGB", (100, 60))
+        d = ImageDraw.Draw(img)
+        d.text((10, 10), text, font=ttf)
+        self.assert_image_similar_tofile(img, "Tests/images/unicode_extended.png", self.metrics['textsize'])
```
radarhere: Suggestion to allow Lint to pass

```suggestion
        self.assert_image_similar_tofile(
            img, "Tests/images/unicode_extended.png", self.metrics['textsize'])
```

nulano: I already fixed that, just forgot to commit and push.

Tests/test_imagefont.py
```diff
@@ -440,6 +440,21 @@ def test_unicode_pilfont(self):
...
+        text = u"A\u278A\U0001F12B"
+        target = "Tests/images/unicode_extended.png"
+
+        ttf = ImageFont.truetype("Tests/fonts/NotoSansSymbols-Regular.ttf",
+                                 FONT_SIZE, layout_engine=self.LAYOUT_ENGINE)
+        img = Image.new("RGB", (100, 60))
+        d = ImageDraw.Draw(img)
+        d.text((10, 10), text, font=ttf)
+
+        self.assert_image_similar_tofile(img, target, self.metrics['textsize'])
```
radarhere: Bear with me if this is a silly question, but why is the third argument `self.metrics['textsize']`?

nulano: I don't know what these values mean, I just guessed based on the other tests. Looks like changing it to `self.metrics['multiline']` would fix the failing builds. Do you happen to know the meaning of these values?

radarhere: Oh, I see what you mean - the other tests do have similar lines.

The function called is https://github.com/python-pillow/Pillow/blob/e8b2da587f721e9e7fee7135dee159b9330098f4/Tests/helper.py#L152

The third argument, epsilon, is a measure of how much different the first two images can be and still allow the test to pass.

nulano: ~~Yes, I found that, but I can't find a reason why these values were chosen and named as they were.~~
Duh, use `git blame`: https://github.com/python-pillow/Pillow/commit/20da797fcee7cfa22ee9637cb8b3b041e1fa8ecb sort-of explains this. I'll just change the value as I wrote earlier.

nulano:
I'm not sure why the Travis build got cancelled, but it did pass on my repository:
https://travis-ci.org/nulano/Pillow/builds/517255375

radarhere:
I restarted the Travis jobs, and they now pass.

nulano:
It looks like Codecov isn't receiving coverage from any Python 2.7 builds with Raqm installed (such as alpine: https://travis-ci.org/python-pillow/Pillow/jobs/517268209#L712):
https://codecov.io/gh/python-pillow/Pillow/src/a4fd4c2b/src/_imagingft.c#L370

hugovk:
Good spot!

The Alpine job is 7325.10 and is shown as received by Codecov:

https://codecov.io/gh/python-pillow/Pillow/commit/a4fd4c2b23d921ea8263738c2e6f87720774d5f2/build

But checking the logs (in the `if [ "$LINT" == "" ]; then` section), there's some problem with coveralls-merge, which combines coverage data for the Python and C files. I expect we're only getting the Python coverage data in this case. Scrolling back a few lines shows a missing dependency: `E: Unable to locate package lcov`

https://travis-ci.org/python-pillow/Pillow/jobs/517268209#L4166

hugovk:
From https://github.com/python-pillow/Pillow/issues/3835#issuecomment-507408755:

> Our tests didn't pick it up because ... whilst we do test PyPy3 on Travis CI, there is a ["implicit declaration of function ‘PyUnicode_AsUCS4Copy’" warning](https://travis-ci.org/python-pillow/Pillow/jobs/552842796#L2596) we didn't notice as all the [`test_imagefont` tests are skipped because there's no FreeType (or RAQM)](https://travis-ci.org/python-pillow/Pillow/jobs/552842796#L4209) on PyPy3. We should improve these testing gaps.

That's not quite accurate: FreeType is available on Travis CI for PyPy3.

The `test_imagefont` tests were skipped when this PR was merged:

* https://travis-ci.org/python-pillow/Pillow/jobs/552376944#L4197

But they were run on the previous merge to master:

* https://travis-ci.org/python-pillow/Pillow/jobs/552376428#L4192

Could it be that `PY_VERSION_HEX` isn't available on PyPy?

nulano:
> Could it be that `PY_VERSION_HEX` isn't available on PyPy?

It is available.

Because FreeType support is loaded dynamically on Linux, I guess it failed to dynamically link the `PyUnicode_AsUCS4Copy` function and assumed it is because of missing FreeType support.
https://travis-ci.org/python-pillow/Pillow/jobs/552842796#L3090

hugovk:
I reported this to PyPy and they've now added `PyUnicode_AsUCS4Copy()`:

> We are just following CPython with our emulation of the CPython C API. If CPython deprecates but keeps around functions, then they'll stay in PyPy too.
>
> `PyUnicode_AsUCS4Copy()` is not implemented in PyPy so far, even though it was present since CPython 3.3. I just added this function in 3ec1002a818c. Closing this as resolved, but please reopen if you hit more missing related functions.

* https://bitbucket.org/pypy/pypy/issues/3035/cpython-old-style-unicode-api-deprecations

Let's check again when the next PyPy release is out. The current version is 7.1.1:

* https://morepypy.blogspot.com/2019/04/pypy-711-bug-fix-release.html

nulano:
> Let's check again when the next PyPy release is out. The current version is 7.1.1:
> 
> * https://morepypy.blogspot.com/2019/04/pypy-711-bug-fix-release.html

PyPy3.6-7.2.0 has been released. The fix is ready, but depends on #4140 and #4109 being merged.

PR #3787, opened 2019-04-11
Fixed raqm layout bug

### LINKED ISSUE #3235: Clipping when rendering Khmer text using Pillow and libraqm
### What did you do?
I used Pillow (5.3.0) to draw Khmer text onto an image using libraqm (0.5.0) for complex layout support (`layout_engine=ImageFont.LAYOUT_RAQM`).

I expected the word "លោកុប្បត្តិ" to be rendered like this (output from MS Word):
khmer-word

### What actually happened?
The left offset of the text was too large and resulted in clipping on the right-hand side of the bounding box (shown in red):
khmer-pillow

Depending on how many glyphs of the word I chose to render, the offset became larger.

I am using the following versions on Ubuntu 18.04:

- Python 3.4.3
- Pillow 5.3.0.dev0 
- libraqm 0.5.0 (from https://github.com/python-pillow/pillow-depends)
- libfreetype6: 2.8.1-2ubuntu2 
- libharfbuzz0b: 1.7.2-1ubuntu1

Code to reproduce clipping effect:

```python
#!/usr/bin/python3
# -*- coding: utf-8 -*-
from PIL import Image
from PIL import ImageFont
from PIL import ImageDraw

img = Image.new("RGB", (512, 512), "white")
draw = ImageDraw.Draw(img)
font = ImageFont.truetype("Battambang-Regular.ttf", 60, layout_engine=ImageFont.LAYOUT_RAQM)
text = "លោកុប្បត្តិ"
draw.text((0, 0), text, (0, 0, 0), font=font)
img.save('out.png')
```

radarhere:
Resolves #3235

The new font is taken from http://www.khmeros.info/?q=en/download, which states that the font is public domain.


Yoki0304:
Could you please upload the font again? The url is no longer existing. Thanks a lot.

radarhere:
You can find it at https://github.com/python-pillow/Pillow/blob/master/Tests/fonts/KhmerOSBattambang-Regular.ttf

PR #3791, opened 2019-04-13
Fixed bug in decoding large images

### LINKED ISSUE #1475: OverflowError in PIL.Image.fromarray
When trying o generate a huge image (73067 x 28160) from a numpy ndarray, pillow throws me an 
"OverflowError: size does not fit in an int". Is pillow limited to maximum sizes of below 2**31bit?

I used:

``` python
im = PIL.Image.fromarray(array, mode='F')
```

radarhere:
Resolves #1475

### DIFF
--- a/Tests/check_large_memory.py
+++ b/Tests/check_large_memory.py
@@ -12,6 +12,12 @@ from .helper import unittest, PillowTestCase
 # 2.7 and 3.2.
 
 from PIL import Image
+
+try:
+    import numpy
+except ImportError:
+    numpy = None
+
 YDIM = 32769
 XDIM = 48000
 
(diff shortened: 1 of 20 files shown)


hugovk:
https://github.com/python-pillow/Pillow/issues/1475#issuecomment-146900183 suggests changing all `int`s to `py_ssize_t`.

What are the pros and cons of using `unsigned int` over `py_ssize_t`?

radarhere:
It didn't occur to me that `py_ssize_t` was also applicable here, thanks for spotting. I've changed the commit to use that.

I've also changed the other decoders as well, to fix a compilation warning.

elhuhdron:
Many PIL operations are working for me for image manipulations of images over 32 bit in size, which is fantastic. However, I still have a problem with fromarray even after this fix (tested using PIL version 6.1.0 and numpy version 1.16.4):

The test case is relatively simple:

```
from PIL import Image
import numpy as np

# this works
a = np.zeros((46340, 46340), dtype=np.uint8)
b = Image.fromarray(a)
del a,b

# this works
b = Image.new('L', (46341, 46341))
del b

# this does not work, fails with:
# MemoryError: Integer overflow in ysize
a = np.zeros((46341, 46341), dtype=np.uint8)
b = Image.fromarray(a)
```

This is likely because 46341*46341 > 2^31-1
Is this easily fixable?

radarhere:
@elhuhdron could you open a new issue with your problem?

elhuhdron:
Sure, I opened issue #3963

PR #3801, opened 2019-04-19
Remove Python.h where not needed

kerisquared:
* Removes Python.h header from a couple of python-free libImaging files.


radarhere:
Thanks

PR #3802, opened 2019-04-21
Added variation font support

### LINKED ISSUE #3789: Support for variable fonts?
Hi, <br>
are you planning to support variable fonts? Would be a great feature allowing people to work with it. DrawBot is the only Python tool that supports it as far as I know - Mac only, it has some dependencies. Most reliable for others on other platforms is recording it from Google Chrome and that the way I do it as well.  

https://en.wikipedia.org/wiki/Variable_fonts

radarhere:
Resolves #3789

See https://www.freetype.org/freetype2/docs/reference/ft2-multiple_masters.html for the API used, and https://docs.microsoft.com/en-us/typography/opentype/spec/otvaroverview for some general information.


jansindl3r:
to get a general idea what these fonts can do and how they behave you can check
https://v-fonts.com/
https://www.axis-praxis.org/specimens/__DEFAULT__

PR #3819, opened 2019-05-01
Include CPPFLAGS when searching for libraries

jefferyto:

 * This adds CPPFLAGS to the list of environment variables (currently CFLAGS and LDFLAGS) searched when looking for includes and library directories, as CPPFLAGS may also include `-I` options.


hugovk:
Thanks.

PR #3820, opened 2019-05-02
Error: "OSError: cannot identify image file"

anhnamxtanh:
I installed pillow-simd (in google colab) and tried open image, but I got an error bellow:

https://colab.research.google.com/drive/1g2Jk9WarGWsylIBqOyK_bFNfFZJ_5R26

How to fix it? Thanks!


radarhere:
Hi. You're created a pull request. Pull requests are for suggesting code changes to be included in a project, while reporting a problem, like you are doing, is what issues are for.

Also, this is Python Pillow, while you're experiencing a problem with Pillow SIMD. Testing in another Colaboratory session, I can reproduce your problem with Pillow SIMD, but not with Pillow. So I would suggest you create an issue over there - https://github.com/uploadcare/pillow-simd/issues/new

Issue #3351, opened 2018-09-13
Converting MCIDAS to PDF results in empty file


OskarPersson:
### What did you do?
Converted MCIDAS .ara-file to PDF

### What did you expect to happen?
A PDF would be created with content similar to the source

### What actually happened?
An empty PDF was created

### What versions of Pillow and Python are you using?
Pillow 5.2.0 and Python 2.7.15

[The tested file](https://github.com/python-pillow/Pillow/blob/9626e49c0bf571d83a99b08709ef4e0c8dc01e03/Tests/images/cmx3g8_wv_1998.260_0745_mcidas.ara)

```python
ara = Image.open("Tests/images/cmx3g8_wv_1998.260_0745_mcidas.ara")
rgb_ara = ara.convert('RGB')
rgb_ara.save('output.pdf')
```

radarhere:
If I run your code, the PDF is a white image 1,800 by 400.

However, this is also what I get if I `show` the image instead -

```
from PIL import Image
ara = Image.open("cmx3g8_wv_1998.260_0745_mcidas.ara")
ara.show()
```

I presume that you think the image is not just a while rectangle? What program are you using to view the original file?

OskarPersson:
I viewed it by converting it to a PNG and then opening the resulting file in Apples Preview app

```python
from PIL import Image
ara = Image.open("cmx3g8_wv_1998.260_0745_mcidas.ara")
ara.save("output.png")
```

radarhere:
I think this is related to #3159

radarhere:
I've created PR #3838 to resolve this.

radarhere:
Resolved by #3838

radarhere:
It turns out that this situation is more complicated. See https://github.com/python-pillow/Pillow/pull/3838#discussion_r292114051

radarhere:
Closing, as part of #3159

Issue #3159, opened 2018-06-07
Image conversion should scale pixel values accordingly


SolarLiner:
### What did you do?
I tried to convert grayscale images of different modes together

### What did you expect to happen?
Conversion should scale values; for example, converting from float to 8-bit should have the values scaled by 255, converting from 8-bit to 16-bit should have the values scaled by 65535/255, etc.

### What actually happened?
Values are being clamped
```python
>>> img = Image.open('16bit_image.png')
>>> img.mode
'I'
>>> numpy.array(img)
array([[51559, 52726, 50875, ..., 30493, 30991, 29907],
       [51743, 52185, 51221, ..., 30841, 29920, 30793],
       [51279, 50534, 51128, ..., 31532, 30852, 30651],
       ...,
       [28288, 27868, 28032, ..., 34367, 34235, 34312],
       [26900, 27567, 28120, ..., 36229, 34607, 33399],
       [27966, 28224, 27962, ..., 36223, 35851, 34477]], dtype=int32)
>>> numpy.array(img.convert('L'))
array([[255, 255, 255, ..., 255, 255, 255],
       [255, 255, 255, ..., 255, 255, 255],
       [255, 255, 255, ..., 255, 255, 255],
       ...,
       [255, 255, 255, ..., 255, 255, 255],
       [255, 255, 255, ..., 255, 255, 255],
       [255, 255, 255, ..., 255, 255, 255]], dtype=uint8)
```
Floating point data really doesn't go over well either
```python
>>> img_float = Image.fromarray(numpy.divide(numpy.array(img), 2**16-1))
>>> numpy.array(img_float)
array([[0.7867399 , 0.8045472 , 0.77630275, ..., 0.46529335, 0.47289234,
        0.45635158],
       [0.78954756, 0.79629207, 0.78158236, ..., 0.4706035 , 0.45654994,
        0.46987107],
       [0.78246737, 0.7710994 , 0.7801633 , ..., 0.48114747, 0.47077134,
        0.46770427],
       ...,
       [0.4316472 , 0.42523843, 0.4277409 , ..., 0.5244068 , 0.52239263,
        0.52356756],
       [0.41046768, 0.42064545, 0.4290837 , ..., 0.55281913, 0.52806896,
        0.50963604],
       [0.4267338 , 0.43067065, 0.42667276, ..., 0.5527275 , 0.5470512 ,
        0.5260853 ]], dtype=float32)
>>> img_oct = img_float.convert('L')
[...]

Issue #3011, opened 2018-02-20
Converting this 16-bit grayscale image to 'L' mode destroys it


ExplodingCabbage:
Here is a 16-bit grayscale image:

![test](https://user-images.githubusercontent.com/2358339/36452647-4647c7ae-168d-11e8-80b9-43bf34a2d8f3.png)

It clearly contains a gradient of grays.

If I open it with Pillow, convert it to 8-bit grayscale, and save it, like so...

```python
>>> from PIL import Image
>>> test_img = Image.open('test.png')
>>> test_img.mode
'I'
>>> test_img.convert('L').save('out.png')
```

... then I get this, which is mostly completely white:

![out](https://user-images.githubusercontent.com/2358339/36452736-86cd23d2-168d-11e8-9f50-43acc06e3d6a.png)

This conversion should work; according to http://pillow.readthedocs.io/en/4.2.x/handbook/tutorial.html#converting-between-modes

> The library supports transformations between each supported mode and the “L” and “RGB” modes. 

and according to http://pillow.readthedocs.io/en/latest/handbook/concepts.html#modes, `I` is a supported mode.

Notably, I *don't* see this same problem if I start by loading an 8-bit RGB image from a JPG and then do `.convert('I').convert('L')`, so it's not simply the case that `I`->`L` conversion is broken in general. I'm not what specifically leads to the breakage in this case.

wiredfool:
There's a longstanding behavioral issue with Pillow where conversions don't intelligently use the range of the target mode. 

So, in this case, you're taking an image with values from 0-65k and converting it to 0-255, with clipping. Most of the values are > 255, so they're all white. When you start with an 8 bit image, you convert 0-255 to 0-65k, but since there's no promotion, it's still just a 0-255 image. Converting back is then not a problem.

radarhere:
This looks related to #3159

radarhere:
I've created PR #3838 to resolve this.

radarhere:
Resolved by #3838

radarhere:
It turns out that this situation is more complicated.

jamesjjcondon:
FYI, my work around for my use case (large gray-scale images):
`x = np.linspace(0, 65535, 1000, dtype=np.uint16)
image = np.tile(x, (1000, 1)).T
plt.imshow(image)
plt.show()

im32 = image.astype(np.int32)
pil = Image.fromarray(im32, mode='I')`

Shouldn't loose precision.

[...]

Issue #3835, opened 2019-05-08
Release Pillow 6.1.0 on July 1, 2019


radarhere:
# Release notes needed
* [x] https://github.com/python-pillow/Pillow/pull/3608
* [x] https://github.com/python-pillow/Pillow/pull/3861
* [x] https://github.com/python-pillow/Pillow/pull/3886

# Release checklist

* [x] Open a release ticket e.g. https://github.com/python-pillow/Pillow/issues/3154
* [x] Develop and prepare release in `master` branch.
* [x] Check [Travis CI](https://travis-ci.org/python-pillow/Pillow) and [AppVeyor CI](https://ci.appveyor.com/project/python-pillow/Pillow) to confirm passing tests in `master` branch.
* [x] Check that all of the wheel builds [Pillow Wheel Builder](https://github.com/python-pillow/pillow-wheels) pass the tests in Travis CI.
* [x] In compliance with [PEP 440](https://www.python.org/dev/peps/pep-0440/), update version identifier in `src/PIL/_version.py`
* [x] Update `CHANGES.rst`.
* [x] Run pre-release check via `make release-test` in a freshly cloned repo.
* [x] Create branch and tag for release e.g.:
  ```bash
  git branch 5.2.x
  git tag 5.2.0
  git push --all
  git push --tags
  ```
* [x] Create source distributions e.g.:
  ```bash
  make sdist
  ```
* [x] Create [binary distributions](https://github.com/python-pillow/Pillow/blob/master/RELEASING.md#binary-distributions)

### Windows
* [x] Contact `@cgohlke` for Windows binaries via release ticket e.g. https://github.com/python-pillow/Pillow/issues/1174.
* [x] Download and extract tarball from `@cgohlke` and `twine upload *`.

### Mac and Linux
* [x] Use the [Pillow Wheel Builder](https://github.com/python-pillow/pillow-wheels):
  ```bash
  git clone https://github.com/python-pillow/pillow-wheels
  cd pillow-wheels
  ./update-pillow-tag.sh [[release tag]]
  ```
* [x] Download distributions from the [Pillow Wheel Builder container](http://a365fff413fe338398b6-1c8a9b3114517dc5fe17b7c3f8c63a43.r19.cf2.rackcdn.com/).
  ```bash
  wget -m -A 'Pillow-<VERSION>*' \
  http://a365fff413fe338398b6-1c8a9b3114517dc5fe17b7c3f8c63a43.r19.cf2.rackcdn.com
  ```
[...]

PR #3838, opened 2019-05-09
 Improved I mode conversion

radarhere:
Improves I mode conversion, scaling values, rather than clipping.

___________________________|Before                                                                                                                              |After
:-------------------------:|:----------------------------------------------------------------------------------------------------------------------------------:|:----------------------------------------------------------------------------------------------------------------------------------:
|#3011                     |issue3011_I_converted_to_L  |issue3011_I_converted_to_L
|#3351                     |issue3351_I_converted_to_RGB|issue3351_I_converted_to_RGB
|`hopper("I")`             |hopper_I                    |hopper_I
|`hopper("I;16")`          |hopper_I;16                 |hopper_I;16
|`hopper("F").convert("I")`|hopper_F_converted_to_I     |hopper_F_converted_to_I

Also resolves the first part of #3159.

This change the results for a number of tests though, most significantly for ImageMath. The value of 'I' is changed now, being a scaled version of other modes instead of a clipped version, so when ImageMath returns a result with 'I' in it, that number is also scaled.


docs/handbook/concepts.rst
```diff
@@ -25,9 +25,9 @@ Modes
 -----
 
 The ``mode`` of an image defines the type and depth of a pixel in the image.
-Each pixel uses the full range of the bit depth. So a 1-bit pixel has a range
-of 0-1, an 8-bit pixel has a range of 0-255 and so on. The current release
-supports the following standard modes:
+A 1-bit pixel has a range of 0-1, an 8-bit pixel or a 32-bit floating point
+pixel has a range of 0-255, and a 32-bit signed integer has a range of 0-65535.
+The current release supports the following standard modes:
```
hugovk: For brevity:
```suggestion
These modes are supported:
```

wiredfool:
This one is going to need release notes as a potentially breaking change

docs/handbook/concepts.rst
```diff
@@ -25,9 +25,9 @@ Modes
 -----
 
 The ``mode`` of an image defines the type and depth of a pixel in the image.
-Each pixel uses the full range of the bit depth. So a 1-bit pixel has a range
-of 0-1, an 8-bit pixel has a range of 0-255 and so on. The current release
-supports the following standard modes:
+A 1-bit pixel has a range of 0-1, an 8-bit pixel or a 32-bit floating point
+pixel has a range of 0-255, and a 32-bit signed integer has a range of 0-65535.
+These modes are supported:
```
wiredfool: Is that range actually correct?  32 bit signed is +-2 billion, that sounds more like 16bit unsigned.

src/libImaging/Convert.c
```diff
@@ -585,12 +585,7 @@ i2l(UINT8* out, const UINT8* in_, int xsize)
...
     INT32* in = (INT32*) in_;
     for (x = 0; x < xsize; x++, in++, out++) {
-        if (*in <= 0)
-            *out = 0;
-        else if (*in >= 255)
-            *out = 255;
-        else
-            *out = (UINT8) *in;
+        *out = (UINT8) (*in >> 8);
     }
```
wiredfool: I think that you’ve got the I;16 and I modes mixed up.  I suspect that part of the problem with the referenced issues was likely I;16 image data being loaded in an I(;32) image because the i16 modes are complicated and only partially supported, but apart from the conversion to 8 bit, it’s safe (but inefficient) to operate on 16bit data in 32 bit storage.

hugovk:
> This one is going to need release notes as a potentially breaking change

Noted in #3835. Or should we play it safe and hold off on this until major release 7.0.0 on 2020-01-01?

wiredfool:
As I noted on some of the other comments, I’m not sure that this pr is correct, because it Appears that the i16 and i32 modes got confused.

src/libImaging/Convert.c
radarhere: Thanks for checking, and what you're saying makes sense. So in order to fix this properly, am I going to have to switch Pillow to reading I;16 actually as I;16 not I then?

wiredfool: I think it's more difficult than that. 

I;16 and I;32 are not used quite like the 8 bit versions -- often times the actual image is only 15 or 10 bit depth, but there's no way to specify a 10 bit gray image, so it gets rounded up to the next byte. There are definitely files that I've seen with I;16 that are in the range of 0-8000 or so. So there's really no way to do a downconversion without extra info to tell you what the intent of the conversion is. Do you want to preserve the LSB, MSB, or autoscale?

The various I;16 modes (signed or unsigned) are not well supported, and really are only useful for bridging to numpy or file format conversions. I think it's basic things like resampling and filtering that don't work with I;16. But operating on a I;16 in 32 bit mode is fine, because none of the math is broken, so we went with that.

The reason that I never fixed this longstanding issue is that I never had a good handle on what I wanted the API to look like, and changing things as it is will break some users while fixing others. 

I think that something like a convert mode for downsampling would make sense:
```
img = i32.convert('L', overflow=Image.CLIP)
```
could work, where we had an enum of Clip, Shift or Autoscale. Autoscale is definitely a new case and more work. Clip is what we have, Shift is what you wrote (mostly).  (Though, this can work both ways, expanding is also a thing on upconversion). (naming is hard)

I think that fundamentally, you should be able to losslessly convert an I;16 to an I;32 and back by default, and similarly, I think that L->I;32->L should also work. I'd be ok with changing the defaults here on a major version, but probably not for a .x release.

radarhere: Okay, I've created PR #3898 to revert this.

docs/handbook/concepts.rst
jamesjjcondon: According to [numpy docs](https://www.numpy.org/devdocs/user/basics.types.html) 16-bit is 0 - 65535. uint32 is 0 to 4294967295. Ps thanks guys! In desperate need of 16-bit functionality so will be watching this closely.

veonua:
is the fix available in release 7.0.0?

radarhere:
@veonua the fix is not part of any release. It was reverted, as it did not perfectly solve the problem

veonua:
@wiredfool told that it can't be a part of a minor release, 
The  "imperfect" solution can not be merged with special experimental parameters like it was proposed  

> i32.convert('L', overflow=Image.SHIFT)

at this moment I'm facing an unexpected problem - around 5% of my pipeline output is blank images. 

please let me know if there is an alternative library that can simply convert image to grayscale (8bit) png

CrackerHax:
wen fix?

radarhere:
I don't know when this issue will be fixed. If you would like to, you could investigate and create a PR with a solution.

CrackerHax:
It would be far more productive if someone who already knew the code handled this.

tmatsuzawa:
is there a workaround to this problem?

radarhere:
There is not a general workaround. If you have a specific situation, you can open a new issue and we can offer a specific solution, something along the lines of https://github.com/python-pillow/Pillow/issues/7519

PR #3843, opened 2019-05-10
Remove deprecated license_file from setup.cfg

jdufresne:
Starting with wheel 0.32.0 (2018-09-29), the "license_file" option is
deprecated.

https://wheel.readthedocs.io/en/stable/news.html

The wheel will continue to include LICENSE, it is now included
automatically:

https://wheel.readthedocs.io/en/stable/user_guide.html#including-license-files-in-the-generated-wheel-file

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,8 +1,5 @@
 [aliases]
 test=pytest
 
-[metadata]
-license_file = LICENSE
-
 [flake8]
 max-line-length = 88


hugovk:
Checking the version we use: https://github.com/python-pillow/pillow-wheels uses multibuild for building wheels.

`travis_steps.sh` calls either `travis_linux_steps.sh` or `travis_osx_steps.sh`, both which install the latest version of wheel.

```sh
pip install --upgrade pip wheel
```

PR #3845, opened 2019-05-11
Add info to README re: Tidelift [ci skip]

aclark4life:
Link to https://tidelift.com/subscription/pkg/pypi-pillow.


aclark4life:
@python-pillow/pillow-team Any objections? I'd like to send some web traffic to Tidelift's marketing page (https://tidelift.com/subscription/pkg/pypi-pillow)

PR #3854, opened 2019-05-15
Fix double free in font loading failure

### LINKED ISSUE #3853: Crash on trying to load corrupted font as file handle
### What did you do?

When loading corrupted file by handle, PIL crashes:

```console
# python -c "from PIL import ImageFont; print(ImageFont.truetype(open('setup.py', 'rb')))"
double free or corruption (top)
Aborted
```

(Originally observed on actually corrupt TTF file, but it as well triggered by non font)

Raise an exception

### What actually happened?

Crash (SIGABRT)

* OS: Linux
* Python: 3.7.3
* Pillow: 6.0.0, reproduced with current git master

nijel:
The memory is freeed in font_dealloc, doing that again leads to SIGABRT.

Fixes #3853.

 * it removes double free in error handling path

I wanted to add test for this, but the error handling futher is broken as well (the invalid font raises IOError, which is later intepreted as missing file and loader tries to do `os.path.basename` on a file like object), so I ended up fixing the crash only for now.

### DIFF
--- a/Tests/test_imagefont.py
+++ b/Tests/test_imagefont.py
@@ -420,6 +420,10 @@ class TestImageFont(PillowTestCase):
         self.assertRaises(IOError, ImageFont.load_path, filename)
         self.assertRaises(IOError, ImageFont.truetype, filename)
 
+    def test_load_non_font_bytes(self):
+        with open("Tests/images/hopper.jpg", "rb") as f:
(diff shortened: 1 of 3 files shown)


hugovk:
Thank you for the report and the PR! The alternative PR https://github.com/python-pillow/Pillow/pull/3912 has been merged so let's close this. The fix will be in next week's release.

Issue #3125, opened 2018-05-09
The parameter direction of ImageDraw.text can't use Chinese sequence


xyl576807077:
### What did you do?
1. I try to draw chinese character by following test code
```python
from PIL import Image, ImageDraw, ImageFont
size = 32
font = ImageFont.truetype('./font/simsun.ttf', size)
img = Image.new('RGBA', (size, size), (0, 0, 0, 0))
draw = ImageDraw.Draw(img)
draw.text((0, 0), '我和你',  fill='black', direction='ttb')
```

2. Futhermore, if I don't use chinese, just english:
```python
from PIL import Image, ImageDraw, ImageFont, features
size = 32
img = Image.new('RGBA', (size, size), (255, 255, 255))
draw = ImageDraw.Draw(img)
draw.text((0, 0), 'abc', fill='black', direction='rtl')
```
### What did you expect to happen?
1. I want to get the image like that
```
我
和
你
```
2. I want to get the image
![aa](https://user-images.githubusercontent.com/11332927/39805870-aa08f402-53aa-11e8-9431-cd87621cbdf5.png)

### What actually happened?
1. The code can't run successfully. The log is
```
Traceback (most recent call last):
  File "draw_character.py", line 111, in <module>
    draw.text((0, 0), '我和你',  font=font, fill='black', direction='ttb')
  File "/home/xyl/anaconda3/lib/python3.6/site-packages/Pillow-5.0.0-py3.6-linux-x86_64.egg/PIL/ImageDraw.py", line 220, in text
  File "/home/xyl/anaconda3/lib/python3.6/site-packages/Pillow-5.0.0-py3.6-linux-x86_64.egg/PIL/ImageFont.py", line 166, in getmask2
RuntimeError: raqm_layout() failed.
```
The font file can be downloaded in here:  https://github.com/micmro/Stylify-Me/blob/master/.fonts/SimSun.ttf
2. I get the image like this:
![hehe](https://user-images.githubusercontent.com/11332927/39806044-59c970ce-53ab-11e8-8096-a4cd3362543f.png)

### What versions of Pillow and Python are you using?
```
python: 3.6.3
pillow: 5.0.0
```

lvhuiyang:
@xyl576807077  

Hi, I meet the same problem, I try use python2 env to solve the problem but not work.

Finally, I change the font to solve it. You can try to use the `Chinese` font to draw the image If your text is in Chinese. For example, I use the font called '华文细黑.ttf' then the program will work.

[...]

Issue #2656, opened 2017-08-06
Japanese font glyph with direction "ttb" is drawn to incorrect position


masushin:
### What did you do?

I tried to draw Japanese font glyph by following test code.

```python
from PIL import Image, ImageDraw, ImageFont

txt = Image.new('RGBA', (200,200), (255,255,255,255))
fnt = ImageFont.truetype("GenEiAntiqueN-Medium.otf", 80)
d = ImageDraw.Draw(txt)
d.text((60,20), "あい", font=fnt, fill=(0,0,0,255), direction="ttb")
txt.show()
```

### What did you expect to happen?
I expected to get the following image.

![ai](https://user-images.githubusercontent.com/18723885/29004138-ee250fee-7afd-11e7-9142-31ba68545270.png)

### What actually happened?

I got the following image.

![ai2](https://user-images.githubusercontent.com/18723885/29004148-0923991e-7afe-11e7-9182-b594afa87903.png)

Glyphs are drawn to deviate to the lower left.

Also, it seems that these glyphs are drawn overlapping.
(It may be better to deal with this after the first problem has been solved..)

The font file using in my code can be got from the following link.
http://okoneya.jp/font/GenEiAntique-3.0.zip

Incidentally, I tried it using other font "ヒラギノ 明朝".
It included in Japanese mac OS. And same problem appear.
So, I think it is not problem with a specific font file.

Please let me know if there are mistake in my code.

### What versions of Pillow and Python are you using?
Python 3.6.0
Pillow 4.2.1

wiredfool:
Do you have the raqm feature compiled in?  ( you would have had to build from source, as it's not in the binary builds)

masushin:
I installed libraqm using `depends/install_raqm_cmake.sh`.
It seemed that it is compiled from source. So, I think yes.

```
>>> from PIL import features
>>> print ("raqm:",features.check('raqm'))
raqm: True
```

masushin:
Similarly, This phenomenon appear in pillow 4.3.0.

wiredfool:
It looks like the layout we're getting is straight from what raqm is getting from harfbuzz. It's either a bug in harfbuzz, or raqm is driving it incorrectly for this font/script. 

I've looked at this against harfbuzz 1.01 (default ubuntu 16.04) and the latest 1.5.1.

@khaledhosny Can you take a look at this?

khaledhosny:
[...]

PR #3856, opened 2019-05-17
Corrected ttb text positioning

radarhere:
Helps #2656 and #3125

libraqm 0.4 and 0.5 throw an error for ttb - see https://github.com/HOST-Oman/libraqm/commit/7088743be4afe22e82282912bf414f5051046bb8. 0.3 works, but since there is no simple way to test for the difference between 0.3 and 0.4, for simplicity I'm saying that ttb requires libraqm 0.6 or later.


hugovk:
Thanks! 

Is there a way to get the libraqm version programatically? If not, shall we open an issue with libraqm to ask them to add some API? It may be useful in the future.

radarhere:
To be clear, this is testing to see if libraqm 0.6 is present, just not through a version string - I do it by testing for the existence of an unrelated function that was added in 0.6.

I've created https://github.com/HOST-Oman/libraqm/issues/103 requesting a version identifier.

hugovk:
Thanks! It could be that the unrelated feature is removed in 0.7 or some future version, and then breaking this fix.

src/_imagingft.c
```diff
@@ -498,20 +506,20 @@ text_layout_raqm(PyObject* string, FontObject* self, const char* dir, PyObject *
         goto failed;
     }
 
-    if (p_raqm.version == 1){
-        for (i = 0; i < count; i++) {
+    for (i = 0; i < count; i++) {
+        if (p_raqm.version == 1) {
```
khaledhosny: I think it would be faster to check the condition outside the loop.

khaledhosny: I think it would be faster to check the condition outside the loop.

Tests/test_imagefontctl.py
```diff
@@ -104,6 +104,22 @@ def test_text_direction_rtl2(self):
 
         self.assert_image_similar(im, target_img, .5)
 
+    def test_text_direction_ttb(self):
+        ttf = ImageFont.truetype(FONT_PATH, FONT_SIZE)
+
+        im = Image.new(mode='RGB', size=(100, 300))
+        draw = ImageDraw.Draw(im)
+        try:
+            draw.text((0, 0), 'English عربي', font=ttf, fill=500, direction='ttb')
```
khaledhosny: Wouldn’t it be better to actually test some CJK character here?

src/_imagingft.c
```diff
@@ -748,73 +797,80 @@ font_render(FontObject* self, PyObject* args)
...
                     }
                 }
-                source += glyph->bitmap.pitch;
             }
+            source += glyph->bitmap.pitch;
+        }
+        if (horizontal_dir) {
+            x += glyph_info[i].x_advance;
+        } else {
+            y -= glyph_info[i].y_advance;
```
khaledhosny: I think the condition is not needed, you can assign x and y unconditionally. In horizontal direction `y_advance` will be zero and in vertical direction `x_advance` will be zero.

src/_imagingft.c
radarhere: You're right. I've updated the commit

src/_imagingft.c
radarhere: Okay. I've updated the commit.

src/_imagingft.c
radarhere: I'm having trouble finding a small CJK font that could be included within the Pillow license. Do you have any thoughts?

Tests/test_imagefontctl.py
radarhere: I'm having trouble finding a small CJK font that could be included within the Pillow license. Do you have any thoughts?

radarhere: Actually, I've found one that's not too large.

khaledhosny: You can also subset any font with `pyftsubset` (from [FontTools](https://github.com/fonttools/fonttools)) to include just the few characters used in the test.

src/_imagingft.c
radarhere: This looks like a duplicate comment.

hugovk:
Now https://github.com/HOST-Oman/libraqm/pull/104 is merged and released in https://github.com/HOST-Oman/libraqm/releases/tag/v0.7.0, can we replace the version inference with a direct check for 0.7.0+?

radarhere:
Okay, done.

PR #3858, opened 2019-05-20
Fix SPARC memory alignment issues in Pack/Unpack functions

### LINKED ISSUE #3408: selftest fails in ImagingUnpackRGB
### What did you do?

Running the Pillow test suite with 64bit Python 2.7.14 on Solaris 11.4 (SPARC - only) we get a failure down in the ImagingUnpackRGB function

    $ pstack core |grep /builds
       [ /builds/jmcp/userland/trunk/components/python/pillow/build/prototype/sparc/usr/lib/python2.7/vendor-packages/PIL/Image.py:824 (load) ]
       [ /builds/jmcp/userland/trunk/components/python/pillow/build/prototype/sparc/usr/lib/python2.7/vendor-packages/PIL/ImageFile.py:269 (load_prepare) ]
       [ /builds/jmcp/userland/trunk/components/python/pillow/build/prototype/sparc/usr/lib/python2.7/vendor-packages/PIL/ImageFile.py:196 (load) ]

For Pillow v5.3.0, Image.py#824 is

      822         if self.im and self.palette and self.palette.dirty:
      823             # realize palette
      824             self.im.putpalette(*self.palette.getdata())

We expect the test suite to pass with each version of Python, and for Python2.7, in both 32- and 64-bit modes. The test suite passes with Python v3.x on both SPARC and X64, and with Python2.7 in both 32- and 64-bit modes on Solaris 11.4 on the X64 architecture.

### What actually happened?

The process terminated with a bus error:

    sparc-kz:Pillow-5.3.0 $ mdb core
    Loading modules: [ libc.so.1 libpython2.7.so.1.0 ld.so.1 ]
    python2.7:core> ::status
    debugging core file of python2.7 (64-bit) from slitheen
    file: /usr/bin/sparcv9/python2.7
    initial argv: /usr/bin/sparcv9/python2.7 ./selftest.py
    threading model: raw lwps
    status: process terminated by SIGBUS (Bus Error), addr=5cb4f8c067

* OS:  Solaris 11.4, on SPARC
* Python: 2.7.14 - 64bit
* Pillow: 5.1.0, 5.3.0

If desired, I can upload the sparcv9 corefile - it's 17Mb after gzip -9.

kulikjak:
Fixes #3408 

Unpacking and packing code is not valid for SPARC CPUs as it is working with 32bit integers on unaligned addresses. This fix reverts some 'optimizations' so that access is aligned properly.

It seems to me that other platforms can benefit from this as well (e.g. ARM will not be faster with unaligned access), but I cannot test it elsewhere and this works for us. I can however improve as needed.

### DIFF
--- a/src/libImaging/Pack.c
+++ b/src/libImaging/Pack.c
@@ -251,6 +251,15 @@ ImagingPackRGB(UINT8* out, const UINT8* in, int pixels)
 {
     int i = 0;
     /* RGB triplets */
+#ifdef __sparc
+    /* SPARC CPUs cannot read integers from nonaligned addresses. */
+    for (; i < pixels; i++) {
+        out[0] = in[R];
+        out[1] = in[G];
+        out[2] = in[B];
+        out += 3; in += 4;
+    }
+#else
     for (; i < pixels-1; i++) {
         ((UINT32*)out)[0] = ((UINT32*)in)[i];
         out += 3;
@@ -261,6 +270,7 @@ ImagingPackRGB(UINT8* out, const UINT8* in, int pixels)
         out[2] = in[i*4+B];
         out += 3;
     }
+#endif
 }
 
 void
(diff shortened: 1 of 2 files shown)


hugovk:
Thank you.

PR #3859, opened 2019-05-20
Updated resampling filter error messages

### LINKED ISSUE #3849: "Unknown resampling filter" error message can be improved
https://github.com/python-pillow/Pillow/blob/2766d943a1a39ea5beafd3cc10b115f5b608a9ab/src/PIL/Image.py#L1797

https://github.com/python-pillow/Pillow/blob/2766d943a1a39ea5beafd3cc10b115f5b608a9ab/src/PIL/Image.py#L2266

1. "unknown" feels vague/wrong. The resampling filter may very well be a valid one that Pillow is aware of _in general_, it's just that it's _unsupported_ for the current use-case.
2. The error could include both the sampling filter passed in _and_ what filters are supported/expected. Both of these are currently hard to gleam without reading the source code.

radarhere:
Resolves #3849

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -1794,7 +1794,18 @@ class Image(object):
         if resample not in (
                 NEAREST, BILINEAR, BICUBIC, LANCZOS, BOX, HAMMING,
         ):
-            raise ValueError("unknown resampling filter")
+            message = "Unknown resampling filter ("+str(resample)+")."
+
+            filters = [filter[1]+" (%d)" % filter[0] for filter in (
+                (NEAREST, "Image.NEAREST"),
+                (LANCZOS, "Image.LANCZOS"),
+                (BILINEAR, "Image.BILINEAR"),
+                (BICUBIC, "Image.BICUBIC"),
+                (BOX, "Image.BOX"),
+                (HAMMING, "Image.HAMMING")
+            )]
+            raise ValueError(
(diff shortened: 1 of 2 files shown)


src/PIL/Image.py
```diff
@@ -1794,7 +1794,18 @@ def resize(self, size, resample=NEAREST, box=None):
         if resample not in (
                 NEAREST, BILINEAR, BICUBIC, LANCZOS, BOX, HAMMING,
         ):
-            raise ValueError("unknown resampling filter")
+            message = "Unknown resampling filter ("+str(resample)+")."
```
hugovk:
```suggestion
            message = "Unknown resampling filter ({}).".format(resample)
```

src/PIL/Image.py
```diff
@@ -2263,7 +2274,22 @@ def __transformer(self, box, image, method, data,
             raise ValueError("unknown transformation method")
 
         if resample not in (NEAREST, BILINEAR, BICUBIC):
-            raise ValueError("unknown resampling filter")
+            if resample in (BOX, HAMMING, LANCZOS):
+                message = {
+                    BOX: "Image.BOX",
+                    HAMMING: "Image.HAMMING",
+                    LANCZOS: "Image.LANCZOS/Image.ANTIALIAS"
+                }[resample]+" (%d) cannot be used." % resample
```
hugovk:
```suggestion
                }[resample]+" ({}) cannot be used.".format(resample)
```

PR #3860, opened 2019-05-22
Replace non-ascii quotes from docstring

olt:
This gave me encoding warnings when running pytest on Python 2.7.

### DIFF
--- a/src/PIL/ImageFont.py
+++ b/src/PIL/ImageFont.py
@@ -284,8 +284,8 @@ class FreeTypeFont(object):
         """
         Create a bitmap for the text.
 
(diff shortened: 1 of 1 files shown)


olt:
Ugh, sorry for missing [ci skip] :-/

hugovk:
Thanks for the PR! In other docstrings, have we used backticks (`1`) or straight quotes ("1")?

olt:
> Thanks for the PR! In other docstrings, have we used backticks (`1`) or straight quotes ("1")?

Good point. I think double backticks is the right answer. I fixed that.

hugovk:
Thanks!

PR #3861, opened 2019-05-22
Improve encoding of TIFF tags

olt:
I improved the TIFF encoder to support more types (especially arrays). This is required for the GeoTIFF format which encodes geospatial information. 

- Pass tagtype from v2 directory to libtiff encoder, instead of
autodetecting type.
- Use explicit types. E.g. uint32_t for TIFF_LONG to fix issues on
platforms with 64bit longs.
- Add support for multiple values (arrays). Requires type in v2
directory and values must be passed as a tuple.
- Add support for signed types (e.g. TIFFTypes.TIFF_SIGNED_SHORT).

Note: I changed the `versionadded 3.0.0` note in the docs as this has been fixed with a3d45e9cefd05105a360588cd3087ccd5ae63b35 which is included in 5.4.

/cc @radarhere You are probably most familiar with this part.


olt:
I fixed the linting errors and (hopefully) the compile error on Windows.

olt:
All tests are now passing except the pyroma tests (see #3882)

hugovk:
I've just merged the pyroma fix (https://github.com/python-pillow/Pillow/pull/3882).

olt:
Squashed and rebased on master. All tests pass.

olt:
Rebased again. All ✅

hugovk:
@olt Thanks for the PR and for your patience in rebasing! There's one last conflict, if you can clear that then let's merge this for Monday's release.

src/libImaging/TiffDecode.c
```diff
@@ -520,15 +520,33 @@ int ImagingLibTiffEncodeInit(ImagingCodecState state, char *filename, int fp) {
...
     char field_name[10];
     uint32 n;
     int status = 0;
 
-    const TIFFFieldInfo info[] = {
-        { key, 0, 1, field_type, FIELD_CUSTOM, 1, 0, field_name }
+    // custom fields added with ImagingLibTiffMergeFieldInfo are only used for
+    // decoding, ignore readcount;
+    int readcount = 0;
+    // we support writing single value, or a variable number of values
```
radarhere:
```suggestion
    // we support writing a single value, or a variable number of values
```

hugovk:
Thanks!

Would be great to still add release notes for this one in https://github.com/python-pillow/Pillow/blob/master/docs/releasenotes/6.1.0.rst.

hugovk:
Release notes for this one added in https://github.com/python-pillow/Pillow/pull/3930.

PR #3864, opened 2019-05-23
Fixed bugs in calculating text size

### LINKED ISSUE #3745: Text clipping arabic fonts
### What did you do?
I used Pillow (5.4.1) to draw Arabic text onto a canvas using libraqm (0.5.0) 

I expected the arabic text to be rendered without clipping. 

### What actually happened?

The text was clipped. 

image

* OS: Ubuntu 18.04 
* Python: 3.6.8
* Pillow: 5.4.1

```python
%matplotlib inline 

font = ImageFont.truetype('ArefRuqaa-Regular.ttf', 70)
canvas = Image.new('RGB', (128, 128), color='white')
draw = ImageDraw.Draw(canvas)
draw.text((0, 0), 'لح', 'black', font, layout_engine=ImageFont.LAYOUT_RAQM)
canvas.save('test.png', format='PNG')

plt.imshow(plt.imread('test.png'))
```

I am using the Aref-Ruqaa font. The exact version I am using is here:
https://github.com/alif-type/aref-ruqaa/blob/master/ArefRuqaa-Regular.ttf

Using `font.getsize('لح')` gives me `(29, 121)` but as you can see from the image above, while the text is `121` pixels high it is still clipped from the top. It is also clipped on the right side, even though there is an offset of `(-2, 46)`. 

So the dimensions are correct but the actual drawn text is clipped. I was informed that the font is cascading so font global metrics cannot be used reliably for clipping. 

Is there any way to correct this clipping? 

Thanks!

radarhere:
Resolves #3745


src/_imagingft.c
```diff
@@ -630,7 +630,7 @@ font_getsize(FontObject* self, PyObject* args)
 
 
     for (x = i = 0; i < count; i++) {
-        int index, error;
+        int index, error, offset, x_advanced;
```
hugovk: What's the reason for moving these declarations before use?

radarhere: `x_advanced` is new, not moved. I'm happy to move the declarations down if you like, I'm just going with the convention established in the code - I moved the `offset` declaration up because it is no longer inside a conditional block.

radarhere:
With English characters, each subsequent character ends horizontally after all of the previous ones. Pillow has a variable `x`, and it moves it forward with each character. Pillow presumes the last `x` is the maximum x value.

From the issue, there are two characters - the backwards S at the bottom, and the line at the top. The backwards S ends horizontally after the line at the top - the horizontal clipping occurs because Pillow stops the size after the line at the top. So instead, check each character to see if it has the maximum x value.

The first diagram at https://www.freetype.org/freetype2/docs/tutorial/step2.html is helpful in understanding the glyph properties.

With regards to changing the vertical clipping, I presume that Pillow is not correctly processing the fact that the line at the top does not start at the baseline - there is a gap. Flipping the signs when applying the offset fixes this problem, with the only affect on existing tests being that a text string is moved vertically up.

Let me know if anything is still unclear.

PR #3870, opened 2019-05-25
Add __main__.py to output basic format and support information

jdufresne:
I recently came across a scenario where Pillow was installed in a VM
with limited OS libraries. To help debug, I was curious what formats
were supported. I think the following feature be useful for such scenarios:

If you like the idea, but want some alterations in the output format, let me know. I'm happy to adjust it.

---

To help debug and show supported formats, users can run:

    $ python -m PIL

to get basic format and support information about the installed version
of Pillow.

The new feature works as follows:

    $ python -m PIL
    Pillow 6.1.0.dev0
    Platform linux 3.7.3 (default, May 11 2019, 00:38:04)
             [GCC 9.1.1 20190503 (Red Hat 9.1.1-1)]

    Supported formats:

    BLP
      Extensions: .blp
      Features: open

    BMP (image/bmp)
      Extensions: .bmp
      Features: open, save

    BUFR
      Extensions: .bufr
      Features: open, save

    …

### DIFF
--- a/Tests/test_features.py
+++ b/Tests/test_features.py
@@ -1,3 +1,7 @@
+from __future__ import unicode_literals
+
+import io
+
 from .helper import unittest, PillowTestCase
 
 from PIL import features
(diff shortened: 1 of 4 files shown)


hugovk:
Good stuff! I've been thinking of something like this that we could ask people to run to help deal with issues.

Sometimes it's not always clear which Python or Pillow installation is running on the command line versus from a script.

* Would it be possible to make it so that it can be run from programmatically a script?

* And print the Pillow installation directory?

* @radarhere Can you think of any other useful info that could be included?

* Would any of the output of `python selftest.py` be useful?

radarhere:
No, no other ideas from me. I was going to comment that the word 'Python' should come before the version string, but it actually does in the code, so maybe that's something that was updated after original post.

jdufresne:
Changes:

- Moved to a function: `PIL.features.pilinfo()`. This can be called programmatically to write to a file like object. Opinions on this as an API are welcome. 
- Adjusted the output to include the header and output from `selftest.py`. This includes the PIL installation directory.The rest of the output has been modified to match this style. Opinions on this style welcome. It now looks like:


```
$ python -m PIL
Pillow 6.1.0.dev0
Python modules loaded from .../Pillow/src/PIL
Binary modules loaded from .../Pillow/src/PIL
Python 3.7.3 (default, May 11 2019, 00:38:04)
       [GCC 9.1.1 20190503 (Red Hat 9.1.1-1)]
--- PIL CORE support ok
--- TKINTER support ok
--- FREETYPE2 support ok
--- LITTLECMS2 support ok
--- WEBP support ok
--- WEBPMUX support ok
--- WEBP Animation support ok
--- JPEG support ok
--- OPENJPEG (JPEG2000) support ok
--- ZLIB (PNG/ZIP) support ok
--- LIBTIFF support ok
--- RAQM (Bidirectional Text) support ok
BLP
Extensions: .blp
Features: open
BMP image/bmp
Extensions: .bmp
Features: open, save
BUFR
Extensions: .bufr
Features: open, save
…

```

PR #3871, opened 2019-05-25
Remove unnecessary repetitive .upper()/.lower() calls

jdufresne:
Plugins are always registered with th ID all uppercase and the extension
all lowercase. These .upper()/.lower() calls amount to a noop.


radarhere:
My concern with this is that we support users writing their own plugins, so this could break existing code. If you did want to do this, this new requirement should be noted in https://pillow.readthedocs.io/en/latest/handbook/writing-your-own-file-decoder.html

jdufresne:
This probably isn't worth it then. The registration code is minimal and only runs once.

PR #3857, opened 2019-05-19
Update issue template for community profile checklist

radarhere:
While Pillow currently has an issue template, the community profile checklist is not detecting this.

Screen Shot 2019-05-19 at 10 23 30 pm

https://help.github.com/en/articles/about-issue-and-pull-request-templates
> To be included in the community profile checklist, issue templates must be located in the .github/ISSUE_TEMPLATE folder and contain valid name: and about: YAML front matter.

So this PR adds those variables. I've based the values here on one of the standards.

Screen Shot 2019-05-19 at 10 25 49 pm

PR #3903, opened 2019-06-13
Advice on general questions, bug reports and feature requests

hugovk:

 * Add guidance on writing issues, as discussed in https://github.com/python-pillow/Pillow/pull/3857#issuecomment-493889987
 * Point general questions towards Stack Overflow
 * Point to guidelines on writing self-contained examples bug reports
 * Advice is in `<!-- -->` comments so will be hidden if not deleted from the issue, to show the most pertinent info

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -10,8 +10,8 @@ Follow these guidelines to ensure your issue is handled properly.
 
 If you have a ...
 
-1.  **General question**: consider asking the question on Stack Overflow
-     with the python-imaging-library tag:
+1.  General question: consider asking the question on Stack Overflow
+    with the python-imaging-library tag:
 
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -2,6 +2,40 @@
...
 ---
+
+<!--
+Thank you for reporting an issue.
+
+Follow these guidelines to ensure your issue is handled properly.
+
+If you have a ...
+
+1.  **General question**: consider asking the question on Stack Overflow
```
radarhere: Since this is part of a comment, the markdown syntax will not be rendered. I don't know if you've done this deliberately anyway, or if it was just done on instinct. If you think it enhances readability for these three points regardless, then cool - I can certainly see that argument for the link dot points and 'code' and 'image'.

hugovk: Thanks, I've removed some of it.

PR #3907, opened 2019-06-16
Support for different charset encodings in PcfFontFile

yawor:

Currently `PcfFontFile` class outputs only bitmap pil fonts with iso8859-1 encoding, even though the PCF format supports Unicode. This makes in hard to work with Pillow with bitmap fonts in languages which use different character sets. With my PR it's possible to set different charset encoding in `PcfFontFile`'s class constructor. By default, it generates pil font file with iso8859-1 like before the change.
Generated pil font file still contains up to 256 characters, but the character set is different depending on the selected encoding.
The change works in both Python 2 and 3.
To use such font with `ImageDraw.text` method, the method must be called with a `str` (py2)/`bytes` (py3) object with the same encoding as the font file.


src/PIL/PcfFontFile.py
```diff
@@ -55,7 +55,9 @@ class PcfFontFile(FontFile.FontFile):
 
     name = "name"
 
-    def __init__(self, fp):
+    def __init__(self, fp, charset_encoding='iso8859-1'):
```
radarhere:
```suggestion
    def __init__(self, fp, charset_encoding="iso8859-1"):
```
Change to allow lint to pass.

yawor: Done

radarhere:
Could you add a test case, something that fails without your change and passes with your change?

yawor:
Should I copy `test_font_pcf.py` as a new test (for example `test_font_pcf_charsets.py` or something) and adapt it for different font and charset? Or should I add new tests in the original one?
Looking at the structure of the current test file for pcf, it seems that the former approach may be better.

radarhere:
I would say a new test, yes.

yawor:
To do the tests I need to also include another pcf file in the Fonts directory alongside properly generated pil/pbm ones for few charsets. Is there any policy regarding what I can place there? I've been testing this using xos4 Terminus fonts amongst others. It is licensed under SIL OFL 1.1. Is it OK to add it to the repository?

radarhere:
Sure. Just add another line to the list in https://github.com/python-pillow/Pillow/blob/master/Tests/fonts/LICENSE.txt

EletronicElephant:
Can confirm this bug exists. BTW, I'm wondering when it will be mergerd?

EletronicElephant:
Also, if I'm understanding the code correctly, this line again makes the whole program decodes only 256 chars from pcf file, which is not acceptable.

https://github.com/yawor/Pillow/blob/94b1a88e4983da96132ead4eebc147462f9ce54f/src/PIL/PcfFontFile.py#L222

I'm corrently working with Pillow to print Chinese Bitmap-fonts onto a screen. It's a hard time for me.

yawor:
@EletronicElephant this probably still waits for writing tests for this feature, about which I've totally forgot.
I've just added them to this PR.
Unfortunately this PR won't help with your problem. The issue is not with the PCF to PIL font converter, but with the PIL font format itself, which allows only 256 characters. To allow more than 256 characters, the format itself would need to be extended or changed.

EletronicElephant:
@yawor Bad news for me. I'll have to walk around for a new method... Thank you all the same ~

Tests/test_font_pcf_charsets.py
```diff
@@ -0,0 +1,129 @@
+# -*- coding: UTF-8 -*-
+from .helper import PillowTestCase
+
+from PIL import Image, FontFile, PcfFontFile
+from PIL import ImageFont, ImageDraw
+
+# from PIL._util import py3
```
hugovk: Not needed
```suggestion
```

Tests/test_font_pcf_charsets.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def test_sanity_iso8859_2(self):
+        self._test_sanity("iso8859-2")
+
+    def test_sanity_cp1250(self):
+        self._test_sanity("cp1250")
+
+    # def test_invalid_file(self):
+    #     with open("Tests/images/flower.jpg", "rb") as fp:
+    #         self.assertRaises(SyntaxError, PcfFontFile.PcfFontFile, fp)
```
hugovk: Is this test needed?

Tests/test_font_pcf_charsets.py
```diff
@@ -0,0 +1,129 @@
...
+    #     draw.text((0, 0), message, "black", font=font)
+    #     with Image.open("Tests/images/high_ascii_chars.png") as target:
+    #         self.assert_image_similar(im, target, 0)
+    #
+    # def test_high_characters(self):
+    #     message = "".join(chr(i + 1) for i in range(140, 232))
+    #     self._test_high_characters(message)
+    #     # accept bytes instances in Py3.
+    #     if py3:
+    #         self._test_high_characters(message.encode("latin1"))
```
hugovk: Are these tests needed?

yawor:
@hugovk regarding the lint issue, I can remove the `u` prefix from strings. Somehow I've overlooked that support for Py2.7 has been dropped when rebasing my changes on the latest master at that time.

Tests/test_font_pcf_charsets.py
yawor: Not really. It's already tested in test_font_pcf.py.

hugovk:
Thank you!

PR #3910, opened 2019-06-18
Changed bits value for XRGB unpackers

### LINKED ISSUE #3885: Unpacking raw XRGB
I found these bugs while inspecting `src/libImaging/Unpack.c` and it's history to see what raw modes I could use before falling back to generic and slow python code in the decoder for my image plugin.

1. `RGBX` for mode `RGB` uses `copy4`, which does not set the X part to 255 like the other RGB unpackers

2. `XRGB` is listed with 24 bits but `ImagingUnpackXRGB` expects 32 bits

PS- Ideally I would be able to use the RGBA component masks and color depth to query Pillow for a compatible raw mode, but there is no way to query or inspect that information.

radarhere:
Resolves #3885


hugovk:
Please could you redo the last commit without "ci skip"? It prevented the PR from being fully tested. Thanks!

radarhere:
Done

PR #3916, opened 2019-06-25
Fix memory leak

cgohlke:


hugovk:
Thanks again!

radarhere:
Thanks for catching

PR #3925, opened 2019-06-29
Split pytest flags

radarhere:
The next time that the Python 3.8 job is run, it will fail with

> ERROR: usage: pytest.py [options] [file_or_dir] [file_or_dir] [...]
> pytest.py: error: unrecognized arguments: -vx

This is connected to the recent release of pytest 5.0.

Splitting 'vx' into separate flags, I find that it passes again.

### DIFF
--- a/.travis/script.sh
+++ b/.travis/script.sh
@@ -7,7 +7,7 @@ make clean
 make install-coverage
 
 python selftest.py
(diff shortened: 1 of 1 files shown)


radarhere:
For the record, see https://github.com/pytest-dev/pytest/issues/5523 for the root of this problem

PR #3928, opened 2019-06-30
Respect the PKG_CONFIG environment variable when building

chewi:
This variable is commonly used by other build systems and using it can help with cross-compiling. Falls back to `pkg-config` as before.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -242,8 +242,9 @@ LCMS_ROOT = None
 
 def _pkg_config(name):
     try:
-        command_libs = ["pkg-config", "--libs-only-L", name]
-        command_cflags = ["pkg-config", "--cflags-only-I", name]
+        command = os.environ.get("PKG_CONFIG", "pkg-config")
+        command_libs = [command, "--libs-only-L", name]
+        command_cflags = [command, "--cflags-only-I", name]
         if not DEBUG:
             command_libs.append("--silence-errors")
             command_cflags.append("--silence-errors")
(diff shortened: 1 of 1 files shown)


DerDakon:
Just to clarify: this has been used in Gentoo for a while.

aclark4life:
@chewi @DerDakon LGTM, thanks. Any risk to approving this for tomorrow's release? I assume not, but if there is a risk, then we'll probably want to wait for 6.2.0… we definitely don't want to make a ``setup.py`` change and then have to release a 6.1.1 in a week 😄

chewi:
It's pretty straightforward, it falls back to the old behaviour, we've had it in Gentoo a while, and your CI has (almost) passed. I think we're good. :smile:

aclark4life:
@hugovk @radarhere OK if I merge this?

hugovk:
Fine by me. 

Is there some documentation on the PKG_CONFIG environment variable? From a quick search I could only find PKG_CONFIG_PATH. 

https://linux.die.net/man/1/pkg-config

DerDakon:
The latter is how to control pkg-config, i.e. where it looks for it's stuff. The former is how a build system picks up which pkg-config or replacement executable, which is a feature of the build systems. The Linux kernel does it: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/tools/build/feature/Makefile#n72 Others are probably doing the same. CMake has a different variable for the same purpose, but also looks ok PKG_CONFIG as a hint: https://cmake.org/cmake/help/latest/module/FindPkgConfig.html#variables-affecting-behavior

PR #3932, opened 2019-07-01
fix bad loop increments in p2i() and p2f()

DerDakon:
Fixes: 7a4af2b7671b14163f61c7f88889fad66e50d77b

@cgohlke


DerDakon:
@hugovk ready to land.

hugovk:
Thanks both!

PR #3935, opened 2019-07-02
Revert #3780 for PyPy3 as it hasn't been updated yet.

nulano:
Should fix the PyPy3 build failure introduced in 8d4bb339a6ec55ff5ddfc61aafc2dc50bb4cbf6c.

This disables the fix for #3777 for PyPy3, as PyPy's cpyext module hasn't been updated to support the new PyUnicode functions yet.

From [#3835 (comment)](https://github.com/python-pillow/Pillow/issues/3835#issuecomment-507408755):

> Our tests didn't pick it up because ... whilst we do test PyPy3 on Travis CI, there is a ["implicit declaration of function ‘PyUnicode_AsUCS4Copy’" warning](https://travis-ci.org/python-pillow/Pillow/jobs/552842796#L2596) we didn't notice as all the [`test_imagefont` tests are skipped because there's no FreeType (or RAQM)](https://travis-ci.org/python-pillow/Pillow/jobs/552842796#L4209) on PyPy3. We should improve these testing gaps.

The FreeType library was probably disabled due to a dynamic link error, not because it's not installed. FreeType works again with this PR: https://travis-ci.org/nulano/Pillow/jobs/553153042#L2976

### DIFF
--- a/Tests/test_imagefont.py
+++ b/Tests/test_imagefont.py
@@ -465,8 +465,10 @@ class TestImageFont(PillowTestCase):
             font.getsize(u"’")
 
     @unittest.skipIf(
-        sys.platform.startswith("win32") and sys.version.startswith("2"),
-        "requires Python 3.x on Windows",
+        sys.platform.startswith("win32") and (
+            sys.version.startswith("2") or hasattr(sys, "pypy_translation_info")
+        ),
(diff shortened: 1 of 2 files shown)


hugovk:
Thanks!

PR #3936, opened 2019-07-02
Revert "Update epsilon for FreeType 2.10 with eg. Unicode width 16" and skip test_unicode_extended on Python 2.x instead

nulano:
I suspect [#3835 (comment)](https://github.com/python-pillow/Pillow/issues/3835#issuecomment-507260313) is actually a real failure. When I tried the test on Windows without the #3780 fix, I got the same values:
`E AssertionError: 2.5 not greater than or equal to 14.728 : average pixel value difference 14.7280 > epsilon 2.5000`
https://ci.appveyor.com/project/nulano/pillow/builds/23676028/job/jjd897m8b68sx96a#L6647.

Reverts python-pillow/Pillow#3931
Skips test_unicode_extended on all platforms when running Python 2.

### DIFF
--- a/Tests/test_imagefont.py
+++ b/Tests/test_imagefont.py
@@ -463,10 +463,7 @@ class TestImageFont(PillowTestCase):
         with self.assertRaises(UnicodeEncodeError):
             font.getsize(u"’")
 
-    @unittest.skipIf(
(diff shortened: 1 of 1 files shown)


hugovk:
Please could you resolve the conflict?

nulano:
Sorry, forgot to push. Try now.

PR #3937, opened 2019-07-02
Added explicit braces

radarhere:
See warning - https://travis-ci.org/python-pillow/Pillow/jobs/552697865#L2627


hugovk:
Yep, it makes sense to always include braces, even for one-liners, to prevent bugs like this:

```diff
if (x)
    f1();
+   f2();
```

No bug:

```diff
if (x) {
    f1();
+   f2();
}
```

Let's start adding them for new code.

PR #3940, opened 2019-07-03
Remove check-manifest from .travis/script.sh

jdufresne:
check-manifest already exists as part of the lint tox target "lint"
which is run by Travis CI. There is no need to re-run the same check as
part of each test run.


hugovk:
Thanks.

PR #3941, opened 2019-07-03
Remove duplicate run of selftest.py

jdufresne:
The make target 'install-coverage' already runs selftest.py. Avoid
running the same script a second time to slightly speed up builds.


hugovk:
Thank you

PR #3945, opened 2019-07-03
Add braces to avoid dangling-else problem

lukegb:
When compiling under clang, it detects that this else block might suffer from the dangling else problem (i.e. the else is ambiguous as to whether it belongs to the first if, which has a nested second if) and issues a diagnostic.


radarhere:
Thanks, but this was already done by #3937

Issue #3626, opened 2019-02-02
ImageGrab cannot get correct resolution screenshot for Windows 10 zoom


HanqingTony:
### What did you do?
I have a large screen and i customed my win10 screen zooming on 125% , then i find that the function "ImageGrab" cannot work correctly, 
### What did you expect to happen?
I want get a screenshot of fullscreen, but it gave me part of it
### What actually happened?
First I used default parameters，the function should have been given me a 1080p picture as my screen resolution, but only a 864p one, just topleft part of my screen. then i used some parameters in order to force set the function to give me screenshot on 1080p but the result is a picture seem like previous 864p picture with terrible black bar on right and below
it is obvious that 864p*125% =1080p, i think the reason caused this problem may concerned about the customed zooming of windows.

### What are your OS, Python and Pillow versions?

* OS:  win10 x64 with 125% zooming
* Python: 3.70
* Pillow: 5.00

### Origin code：

```python
def scanScreen(tpl):
    capture=ImageGrab.grab()
    capture.save('capture.png') 
    pic=cv2.imread('capture.png')
    return findPosition(pic,tpl)

```

### Code can work correctly
I find another way to get my screenshot by use WinAPI，it is fast and always work correctly, code is here, hope helpful

```python

import win32gui, win32ui, win32con, win32api
def window_capture(filename):
  hwnd = 0 # window ID , '0' means the activated window 
  #get Divice Context
  hwndDC = win32gui.GetWindowDC(hwnd)
  #get mfcDC
  mfcDC = win32ui.CreateDCFromHandle(hwndDC)
  saveDC = mfcDC.CreateCompatibleDC()
  # create bigmap for picture saving
  saveBitMap = win32ui.CreateBitmap()
  # get Moniter info
  MoniterDev = win32api.EnumDisplayMonitors(None, None)
  w = MoniterDev[0][2][2]
  h = MoniterDev[0][2][3]
  # print w,h　　　#picture size
  # create bitmap space
  saveBitMap.CreateCompatibleBitmap(mfcDC, w, h)
  # save the screenshot to saveBitmap
  saveDC.SelectObject(saveBitMap)
  # from topleft（0，0）to bottomright（w，h）
  saveDC.BitBlt((0, 0), (w, h), mfcDC, (0, 0), win32con.SRCCOPY)
  saveBitMap.SaveBitmapFile(saveDC, filename)

```

radarhere:
[...]

Issue #3432, opened 2018-10-23
Can't capture the whole screen on Windows 10


Nevss:
when I use 

ImageGrab.grab()

I get this picture.

![bb138a4ccfd5a280s](https://user-images.githubusercontent.com/10334358/47361926-c5509100-d705-11e8-85d2-25d143b0096b.jpg)

I think  windows display zoom caused this.

Could someone help me?

Thanks.

radarhere:
Hi. Just to be clear, could you attach a regular screenshot for comparison?

Nevss:
ok

I mean , I want to shot the whole screen , like this ：


![wx20181026-165008](https://user-images.githubusercontent.com/10334358/47556424-dafdca80-d940-11e8-9a01-686fc3a4f0c0.png)

hugovk:
As a workaround, do either https://github.com/BoboTiG/python-mss or https://github.com/ponty/pyscreenshot help?

radarhere:
This should now have been fixed by #4000

Issue #2438, opened 2017-03-03
Enable DPI Scaling Awareness


yabbas:
### What did you do?
Tried to grab a screenshot with pyautogui.  This passes control over to Pillow's imagegrab.

### What did you expect to happen?
A screenshot of the entire screen.

### What actually happened?
The top left corner of the screen (1/4) was saved.

### What versions of Pillow and Python are you using?
Python 3.6, Pillow 4.00

Please include code that reproduces the issue and whenever possible, an image that demonstrates the issue. The best reproductions are self-contained scripts with minimal dependencies. If you are using a framework such as plone, django, or buildout, try to replicate the issue just using Pillow. 

```python
PIL.ImageGrab.grab(bbox=None)
<PIL.Image.Image image mode=RGB size=1600x900 at 0x132A0D5F898>
# screen resolution is actually 3200x1800
```

Fix - making the App DPI Aware:
https://github.com/asweigart/pyautogui/issues/33#issuecomment-236981496

davePlug:
This is still an issue...and now is becoming more pressing with even more 4k monitors around

radarhere:
#4000 has now been merged, using an awareness context of DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE.

PR #3950, opened 2019-07-05
Add option to capture all monitors on Windows

### LINKED ISSUE #1547: ImageGrab fails with multiple monitors
When calling ImageGrab.grab() passing in a bounding box that is outside the area of my primary monitor, I just get black.

For example, my primary monitor is 1920x1200, flanked on either side by monitors running at 1600x1200, making my total desktop size 5120x1200. Also, because my primary monitor is in the middle, the horizontal coordinates for the full virtual desktop go from -1600 to 3519, where 0 is the left-most pixel of my primary monitor. If I try to capture my rightmost monitor using the following code, all I get is a black image:

``` Python
from PIL import ImageGrab
img = ImageGrab.grab([1920, 0, 3519, 1199])
img.save("test.jpg")
```

Poking around the code, it looks like `ImageGrab.grab()` calls into `Image.core.grabscreen` which is an alias for [`PyImaging_GrabScreenWin32()`](https://github.com/python-pillow/Pillow/blob/2be12dec2b231d31400f44bfa855966484997c16/display.c#L323) in `display.c`. That function does retrieve a DC handle to the entire desktop, but the subsequent calls to `GetDeviceCaps` with `HORZRES` and `VERTRES` only return the x/y size of the primary monitor, not the entire desktop.

``` C
screen = CreateDC("DISPLAY", NULL, NULL, NULL);
// ...
width = GetDeviceCaps(screen, HORZRES);
height = GetDeviceCaps(screen, VERTRES);
// ...
if (!BitBlt(screen_copy, 0, 0, width, height, screen, 0, 0, SRCCOPY))
    goto error;
```

Another problem with the above code is that monitors to the left of or above the primary display have negative coordinates in the `screen` DC. So, for example, capturing the monitor to the left of my primary display (which has a resolution of 1600x1200) would need to call BitBlt with the following coordinates:

``` C
left = -1600
top = 0
width = 1600
height = 1200
BitBlt(screen_copy, 0, 0, width, height, screen, left, top, SRCCOPY)
```

Similarly, if I was trying to capture a monitor above my primary display, then `top` would be negative. Because of the negative coordinates issue, I don't see any way of fixing this without passing in `left, top, width, height` from the calling python code, which could be calculated easily from the `bbox` parameter. Then it's simply up to the caller to know the coordinates of the monitor they want to capture. If no `bbox` is provided, then the coordinates would default to the primary display (0, 0, HORZRES, VERTRES), keeping the current functionality unchanged so as not to break existing code that uses `ImageGrab.grab()`.

nulano:
Resolves #1547

~~Provides workaround for #2438, #3432, #3626:
For some reason grabbing a screenshot of the whole screen with the new `ImageGrab.grab(multimonitor=True)` ignores DPI scaling properly.~~
Edit: I didn't look at the screenshot closely enough.

### DIFF
--- a/docs/reference/ImageGrab.rst
+++ b/docs/reference/ImageGrab.rst
@@ -22,7 +22,11 @@ or the clipboard to a PIL image memory.
     :param bbox: What region to copy. Default is the entire screen.
                  Note that on Windows OS, the top-left point may be negative if ``all_screens=True`` is used.
     :param include_layered_windows: Includes layered windows. Windows OS only.
+
+        .. versionadded:: 6.1.0
     :param all_screens: Capture all monitors. Windows OS only.
+
+        .. versionadded:: 6.2.0
     :return: An image
 
 .. py:function:: PIL.ImageGrab.grabclipboard()
--- a/docs/reference/ImageGrab.rst
+++ b/docs/reference/ImageGrab.rst
@@ -11,7 +11,7 @@ or the clipboard to a PIL image memory.
 
 .. versionadded:: 1.1.3
 
-.. py:function:: PIL.ImageGrab.grab(bbox=None, include_layered_windows=False, multimonitor=False)
+.. py:function:: PIL.ImageGrab.grab(bbox=None, include_layered_windows=False, all_screens=False)
 
     Take a snapshot of the screen. The pixels inside the bounding box are
     returned as an "RGB" image on Windows or "RGBA" on macOS.
@@ -20,8 +20,9 @@ or the clipboard to a PIL image memory.
     .. versionadded:: 1.1.3 (Windows), 3.0.0 (macOS)
 
     :param bbox: What region to copy. Default is the entire screen.
+                 Note that on Windows OS, the top-left point may be negative if ``all_screens=True`` is used.
     :param include_layered_windows: Includes layered windows. Windows OS only.
-    :param multimonitor: Capture all monitors. Windows OS only.
+    :param all_screens: Capture all monitors. Windows OS only.
     :return: An image
 
 .. py:function:: PIL.ImageGrab.grabclipboard()
--- a/docs/reference/ImageGrab.rst
+++ b/docs/reference/ImageGrab.rst
@@ -11,7 +11,7 @@ or the clipboard to a PIL image memory.
 
 .. versionadded:: 1.1.3
 
-.. py:function:: PIL.ImageGrab.grab(bbox=None, include_layered_windows=False)
+.. py:function:: PIL.ImageGrab.grab(bbox=None, include_layered_windows=False, multimonitor=False)
 
     Take a snapshot of the screen. The pixels inside the bounding box are
     returned as an "RGB" image on Windows or "RGBA" on macOS.
@@ -21,6 +21,7 @@ or the clipboard to a PIL image memory.
 
     :param bbox: What region to copy. Default is the entire screen.
     :param include_layered_windows: Includes layered windows. Windows OS only.
+    :param multimonitor: Capture all monitors. Windows OS only.
     :return: An image
 
 .. py:function:: PIL.ImageGrab.grabclipboard()
--- a/src/PIL/ImageGrab.py
+++ b/src/PIL/ImageGrab.py
@@ -29,7 +29,7 @@ else:
     raise ImportError("ImageGrab is macOS and Windows only")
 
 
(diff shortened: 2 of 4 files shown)


src/PIL/ImageGrab.py
```diff
@@ -49,8 +51,10 @@ def grab(bbox=None, include_layered_windows=False):
             (size[0] * 3 + 3) & -4,
             -1,
         )
-    if bbox:
-        im = im.crop(bbox)
+        if bbox:
+            xo, yo = offset
```
radarhere: Could this be changed to `x0, y0`, rather than `xo, yo`?

nulano: sure

src/display.c
```diff
@@ -339,8 +339,16 @@ PyImaging_GrabScreenWin32(PyObject* self, PyObject* args)
...
-    height = GetDeviceCaps(screen, VERTRES);
+    if (multimonitor) {
+        x = GetSystemMetrics(SM_XVIRTUALSCREEN);
+        y = GetSystemMetrics(SM_YVIRTUALSCREEN);
+        width = GetSystemMetrics(SM_CXVIRTUALSCREEN);
+        height = GetSystemMetrics(SM_CYVIRTUALSCREEN);
+    } else {
+        // FIXME doesn't work with DPI scaling
+        width = GetSystemMetrics(SM_CXSCREEN);
+        height = GetSystemMetrics(SM_CYSCREEN);
```
radarhere: Changing from `GetDeviceCaps(screen, HORZRES)` to `GetSystemMetrics(SM_CXSCREEN)` is separate to the 'multimonitor' feature. Could you talk about why you've done this?

nulano: I was hoping it would resolve the DPI issue, but it seems they are equivalent. So I suppose there is no reason for this change.

nulano:
I have removed the FIXME comment now that it will be resolved by #4000 (it was in the wrong place anyway).

src/PIL/ImageGrab.py
```diff
@@ -29,16 +29,18 @@
     raise ImportError("ImageGrab is macOS and Windows only")
 
 
-def grab(bbox=None, include_layered_windows=False):
+def grab(bbox=None, include_layered_windows=False, multimonitor=False):
```
nulano: I just had a thought, would it be better to rename the `multimonitor` parameter to `primary_only` (and invert its value)?

radarhere: I think the problem with inverting it would be that this then changes the default behaviour - Pillow values backwards compatibility.

If you feel like the name needs to clearer, I might suggest `all_screens`.

nulano: I was thinking about making the default be None, which would use the old (platform dependent) behaviour, but could be overriden with True or False.

nulano: Alright, I got my hands on a mac with multiple monitors; it turns out I was wrong - mac also captures the primary monitor only.

I will rename the parameter to `all_screens` as you suggested.

docs/reference/ImageGrab.rst
```diff
@@ -20,7 +20,9 @@ or the clipboard to a PIL image memory.
     .. versionadded:: 1.1.3 (Windows), 3.0.0 (macOS)
 
     :param bbox: What region to copy. Default is the entire screen.
+                 Note that on Windows OS, the top-left point may be negative if ``all_screens=True`` is used.
     :param include_layered_windows: Includes layered windows. Windows OS only.
+    :param all_screens: Capture all monitors. Windows OS only.
```
hugovk: Please could you a `.. versionadded:: 7.0.0` for this param?

Compare with:

* https://pillow.readthedocs.io/en/stable/reference/ImageDraw.html#PIL.ImageDraw.PIL.ImageDraw.ImageDraw.line

* https://github.com/python-pillow/Pillow/blob/master/src/PIL/ImageDraw.py

radarhere: @hugovk to be clear, do you mean 7.0, the end of year release, or 6.2, the next one?

hugovk: Oops, I meant 6.2.0, thanks for catching that!

nulano: Done. I also added it for `include_layered_windows` (6.1.0).

hugovk:
Thanks!

PR #3953, opened 2019-07-05
Use JPEG APP14 marker color transform tag

### LINKED ISSUE #3952: Handle color space transformation correctly for JPEG image compressed in RGB color space
Pillow currently doesn't correctly read JPEG compressed images, which were compressed in RGB color space, i.e. which were not transformed into YCbCr color space upon compression.

hackermd:
Fixes #3952

 * Read Color Transform flag of JPEG APP14 Adobe marker
 * Set `mode` according to value of Color Transform flag

Further information on the APP14 marker:

- https://sno.phy.queensu.ca/~phil/exiftool/TagNames/JPEG.html#Adobe
- https://docs.oracle.com/javase/8/docs/api/javax/imageio/metadata/doc-files/jpeg_metadata.html


src/PIL/JpegImagePlugin.py
```diff
@@ -25,6 +25,7 @@
 # 2009-09-06 fl   Added icc_profile support (from Florian Hoech)
 # 2009-03-06 fl   Changed CMYK handling; always use Adobe polarity (0.6)
 # 2009-03-08 fl   Added subsampling support (from Justin Huff).
+# 2019-07-05 fl   Use Adobe APP14 marker ColorTransform tag.
```
radarhere: My presumption is that 'fl' here stands for 'Fredrik Lundh', as seen a few lines later in the copyright.

src/PIL/JpegImagePlugin.py
```diff
@@ -205,7 +209,16 @@ def SOF(self, marker):
     if self.layers == 1:
         self.mode = "L"
     elif self.layers == 3:
-        self.mode = "RGB"
+        try:
+            transform_flag = self.info["adobe_transform"]
+            if transform_flag == 0:  # Unknown (RGB or CMYK)
+                self.mode = "YCbCr"
+            elif transform_flag in (1, 2):  # YCbCr or YCCK
+                self.mode = "RGB"
```
```suggestion
                self.mode = "YCbCr"
```
Looks like you are setting these values the wrong way around, and that they are the correct way around in your comments.

src/PIL/JpegImagePlugin.py
```diff
@@ -205,7 +209,16 @@ def SOF(self, marker):
     if self.layers == 1:
         self.mode = "L"
     elif self.layers == 3:
-        self.mode = "RGB"
+        try:
+            transform_flag = self.info["adobe_transform"]
+            if transform_flag == 0:  # Unknown (RGB or CMYK)
+                self.mode = "YCbCr"
```
```suggestion
                self.mode = "RGB"
```

radarhere:
Could you add a test that fails before this PR, and then passes with this PR?

src/PIL/JpegImagePlugin.py
hackermd: Sorry, I didn't notice the copyright aspect. My intention was simply to add an item to the history list.

hackermd:
> Could you add a test that fails before this PR, and then passes with this PR?

Happy to add tests.

Please note that this PR would introduced a potentially breaking change: The value of `self.info['adobe_transform']` will now be different if the Adobe marker is found in the JPEG bitstream. Based on my (limited) understanding of the APP14 marker, I assume that the previously implementation resulted in an incorrect value for the Color Transform flag (due to a wrong offset) and that the change that would be introduced via this PR would fix it. @radarhere do you agree?

src/PIL/JpegImagePlugin.py
radarhere, line 28:
```suggestion
```

src/PIL/JpegImagePlugin.py
```diff
@@ -138,8 +138,11 @@ def APP(self, marker):
     elif marker == 0xFFEE and s[:5] == b"Adobe":
         self.info["adobe"] = i16(s, 5)
         # extract Adobe custom properties
+        # https://exiftool.org/TagNames/JPEG.html#Adobe
+        # https://docs.oracle.com/javase/8/docs/api/javax/imageio/metadata/doc-files/jpeg_metadata.html
         try:
-            adobe_transform = s[1]
+            index2_offset = 5
+            adobe_transform = s[index2_offset + 3]  # Index2 3
```
radarhere, line 145: This is actually incorrect as well. As per https://github.com/LuaDist/libjpeg/blob/6c0fcb8ddee365e7abc4d332662b06900612e923/jcmarker.c#L435-L443, the layout is
- 5 bytes for "Adobe" (as evidence a few lines earlier in this Pillow file in `s[:5] == b"Adobe":`
- 2 bytes for DCTEncodeVersion
- 2 bytes for APP14Flags0
- 2 bytes for APP14Flags1
- 1 byte for transform

This makes `s[11]` transform. 

I've created #5408 for this.

radarhere:
I agree that the old `adobe_transform` value was incorrect. As I mentioned in a comment, I think this PR's alternative is also incorrect, and I've created #5408 instead.

Correcting that value is a minor change, to my way of thinking.

hackermd:
Thanks @radarhere! I am closing this issue.

radarhere:
Thanks! Does that mean if my PR is merged, you're also happy for #3952 to be closed?

PR #3954, opened 2019-07-06
Introduce isort to automate import ordering and formatting

jdufresne:
Similar to the recent adoption of Black. isort is a Python utility to
sort imports alphabetically and automatically separate into sections. By
using isort, contributors can quickly and automatically conform to the
projects style without thinking. Just let the tool do it.

Uses the configuration recommended by the black project to avoid
conflicts of style.

### DIFF
--- a/Tests/32bit_segfault_check.py
+++ b/Tests/32bit_segfault_check.py
@@ -1,8 +1,8 @@
 #!/usr/bin/env python
 
-from PIL import Image
 import sys
 
+from PIL import Image
 
 if sys.maxsize < 2 ** 32:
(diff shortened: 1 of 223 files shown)


hugovk:
I'm fine with this, it helps see which imports are system, 3rd-party and local.

Most of the changes are minor, but I'm holding off until the coverage is sorted to check the [`test_imageqt.py`](https://github.com/python-pillow/Pillow/pull/3954/files#diff-951708f105bde59acf79738477e2af95) changes.

hugovk:
I restarted the Travis build and it's covered. Thanks!

PR #3980, opened 2019-07-19
Exif writing fixes: Rational boundaries and signed/unsigned types

kkopachev:
Related: #3973.

Changes proposed in this pull request:

 * Fix RATIONAL and SRATIONAL boundaries when writing IFDs
 * Chose signed/unsigned types based on passed value for unknown tags (fixes #4037)

Issue #3973, opened 2019-07-16
exif_transpose errors out on some images


okor:
### What did you do?
Used new `exif_transpose` method.

### What did you expect to happen?
I expected a transposed image to be returned, without error.

### What actually happened?
An error was raised.
```
/Users/okor/thumbor/lib/python2.7/site-packages/PIL/TiffImagePlugin.py:612: UserWarning: Metadata Warning, tag 282 had too many entries: 2, expected 1
  tag, len(values)))
/Users/okor/thumbor/lib/python2.7/site-packages/PIL/TiffImagePlugin.py:612: UserWarning: Metadata Warning, tag 283 had too many entries: 2, expected 1
  tag, len(values)))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/okor/thumbor/lib/python2.7/site-packages/PIL/ImageOps.py", line 549, in exif_transpose
    transposed_image.info["exif"] = exif.tobytes()
  File "/Users/okor/thumbor/lib/python2.7/site-packages/PIL/Image.py", line 3087, in tobytes
    return b"Exif\x00\x00"+head+ifd.tobytes(offset)
  File "/Users/okor/thumbor/lib/python2.7/site-packages/PIL/TiffImagePlugin.py", line 818, in tobytes
    count = len(data)
TypeError: object of type 'int' has no len()
```

Note that this error is not raised for _all_ images. Only some. I'm not sure how the EXIF data differs.

### What are your OS, Python and Pillow versions?

* OS: macOS Mojave 10.14.5
* Python: 2.7.16
* Pillow: 6.0.0

```python
from PIL import Image, ImageOps

im = Image.open('10_years_of_Wikipedia_by_Guillaume_Paumier.jpg')
im = ImageOps.exif_transpose(im)
```
[10_years_of_Wikipedia_by_Guillaume_Paumier.jpg.zip](https://github.com/python-pillow/Pillow/files/3398388/10_years_of_Wikipedia_by_Guillaume_Paumier.jpg.zip)

---

This error was encountered when attempting to replace `piexif` exif functionality with new Pillow functionality. The PR and conversation is here https://github.com/thumbor/thumbor/pull/1209

kkopachev:
There are multiple issues in there, all with saving back exif info.
Reproducible with pillow's own images:
```python
im = Image.open('Tests/images/exif_gps_typeerror.jpg')
# or use another one for different error
[...]

PR #3981, opened 2019-07-20
Corrected tag type

radarhere:
See https://www.awaresystems.be/imaging/tiff/tifftags/minsamplevalue.html and https://www.awaresystems.be/imaging/tiff/tifftags/exififd.html

### DIFF
--- a/src/PIL/TiffTags.py
+++ b/src/PIL/TiffTags.py
@@ -179,7 +179,7 @@ TAGS_V2 = {
     33432: ("Copyright", ASCII, 1),
     34377: ("PhotoshopInfo", BYTE, 1),
     # FIXME add more tags here
-    34665: ("ExifIFD", SHORT, 1),
+    34665: ("ExifIFD", LONG, 1),
     34675: ("ICCProfile", UNDEFINED, 1),
     34853: ("GPSInfoIFD", BYTE, 1),
     # MPInfo


hugovk:
Could this break anyone's code?

radarhere:
Going from SHORT to LONG certainly shouldn't.

As for going from LONG to SHORT, writing with Pillow rather than libtiff, `struct.error: ushort format requires 0 <= number <= (32767 *2 +1)` might be thrown if the value is too large. Hmm. Shall I just include the second fix then?

hugovk:
Let's do `SHORT` to `LONG` in this release and `LONG` to `SHORT` in 7.0.0.

radarhere:
Okay, I've updated the commit.

hugovk:
Thanks! Feel free to create the other PR and we can park it in the 7.0.0 milestone.

src/PIL/TiffTags.py
```diff
@@ -179,7 +179,7 @@ def lookup(tag):
     33432: ("Copyright", ASCII, 1),
     34377: ("PhotoshopInfo", BYTE, 1),
     # FIXME add more tags here
-    34665: ("ExifIFD", SHORT, 1),
+    34665: ("ExifIFD", LONG, 1),
     34675: ("ICCProfile", UNDEFINED, 1),
     34853: ("GPSInfoIFD", BYTE, 1),
```
kkopachev: This one should also be `LONG`

radarhere: This will be part of #3980 though, as per [your comment](https://github.com/python-pillow/Pillow/issues/3973#issuecomment-511944566), yes?

radarhere: Just to note here - the scope of #3980 has been reduced, and changing the type of GPSInfoIFD is now in #4009

PR #3982, opened 2019-07-20
Use OpenJPEG package

### LINKED ISSUE #1757: Ubuntu/Debian Jpeg2k dependency
From the [debian bug #816175](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=816175)

```
Openjpeg (jpeg2000) support is missing in the binary packages of pillow.
I could fix it locally by adding some openjpeg headers:
 apt-get source python-pil
 sudo apt-get build-dep python-pil
 sudo apt-get install libopenjp2-7-dev
 dpkg-buildpackage -uc -us
 sudo dpkg -i *deb

So I guess it's just a matter of adding libopenjp2-7-dev to pillow
Build-Depends.
```

We should check to see if this is the correct fix and if that's the case, we should put pointers to this package in the docs, the build scripts and elsewhere to use it rather than the download and build script for openjpeg.

radarhere:
Resolves #1757, together with https://github.com/python-pillow/docker-images/pull/67 - used package OpenJPEG instead of compiling it.

Also updates the Ubuntu version in the documentation from 14.04 to 16.04 - partly because 14.04 is now rather old, and partly because 'libopenjp2-7-dev' can't be found in 14.04.

Also adds 'harfbuzz-devel' and 'fribidi-devel' to the Fedora prerequistes, as seen in https://github.com/python-pillow/Pillow/blob/master/depends/fedora_23.sh

### DIFF
--- a/.travis/install.sh
+++ b/.travis/install.sh
@@ -3,9 +3,9 @@
 set -e
 
 sudo apt-get update
-sudo apt-get -qq install libfreetype6-dev liblcms2-dev python-tk\
-			 python-qt4 ghostscript libffi-dev libjpeg-turbo-progs cmake imagemagick\
-             libharfbuzz-dev libfribidi-dev
+sudo apt-get -qq install libfreetype6-dev liblcms2-dev python-tk python-qt4\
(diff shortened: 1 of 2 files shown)


hugovk:
Thanks.

PR #3988, opened 2019-07-25
Improved ImageFont documentation

radarhere:
See https://pillow.readthedocs.io/en/latest/reference/ImageFont.html for the current version

This PR
- fixes the documentation for the `getmask` and `getsize` methods of the ImageFont class. Comparing with the source code - https://github.com/python-pillow/Pillow/blob/master/src/PIL/ImageFont.py#L112 - the current documentation is incorrect regarding the arguments for those methods. The new docstrings I've added are copied and altered from the FreeTypeFont `getmask` and `getsize`.
- adds the TransposedFont class to the documentation

See https://pillow-radarhere.readthedocs.io/en/imagefont_docs/reference/ImageFont.html for the improved version.

### DIFF
--- a/docs/reference/ImageFont.rst
+++ b/docs/reference/ImageFont.rst
@@ -47,90 +47,11 @@ Functions
 Methods
 -------
 
-.. py:method:: PIL.ImageFont.ImageFont.getsize(text, direction=None, features=[], language=None)
-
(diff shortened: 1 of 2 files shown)


hugovk:
Thanks!

PR #3998, opened 2019-07-30
Fix bug in TIFF loading of BufferedReader

### LINKED ISSUE #3996: BufferedReader wrapping BytesIO breaks tiff loader
I struck an edge case trying to open a TIFF file from a BufferedReader that wraps a BytesIO object. The problem is that when the wrapped object is not a file, BufferedReader has a "fileno" attribute but calling it returns an error. This case is handled for BytesIO itself, and sidestepped by using "getvalue" rather than trying to open by fileno. I'll open a PR but documenting the issue here.

* OS: macOS 10.14.3
* Python: 3.7.3
* Pillow: 6.1.0

```python
import io
from PIL import Image

bytesio = io.BytesIO()
with open(tiff_filename, 'rb') as fp:
    bytesio.write(fp.read())

reader = io.BufferedReader(bytesio)
im = Image.open(reader)
im.load()
```
Result:
```
tempfile.tif: Cannot read TIFF header.
  File "/usr/local/lib/python3.7/site-packages/PIL/TiffImagePlugin.py", line 1099, in load
    return self._load_libtiff()
  File "/usr/local/lib/python3.7/site-packages/PIL/TiffImagePlugin.py", line 1191, in _load_libtiff
    raise IOError(err)
OSError: -2
```

chadawagner:
Fixes #3996.

 * There is an existing fp value that is safely created before the call to Image._getdecoder. The same value should just be used again in the logic surrounding the actual decode.
 * For the case of "we have something else" (no getvalue or fileno) it should seek to the beginning as it does in other cases.

### DIFF
--- a/Tests/test_file_libtiff.py
+++ b/Tests/test_file_libtiff.py
@@ -81,6 +81,19 @@ class TestFileLibTiff(LibTiffTestCase):
         self.assertEqual(im.size, (500, 500))
         self._assert_noerr(im)
 
+    def test_g4_non_disk_file_object(self):
+        """Testing loading from non-disk non-bytesio file object"""
+        test_file = "Tests/images/hopper_g4_500.tif"
+        s = io.BytesIO()
+        with open(test_file, "rb") as f:
+            s.write(f.read())
+            s.seek(0)
+        r = io.BufferedReader(s)
+        im = Image.open(r)
+
+        self.assertEqual(im.size, (500, 500))
+        self._assert_noerr(im)
+
     def test_g4_eq_png(self):
         """ Checking that we're actually getting the data that we expect"""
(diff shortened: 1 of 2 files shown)


radarhere:
Could you add a test that fails without your fix and passes with it?

Tests/test_file_libtiff.py
```diff
@@ -81,6 +81,19 @@ def test_g4_tiff_bytesio(self):
         self.assertEqual(im.size, (500, 500))
         self._assert_noerr(im)
 
+    def test_g4_non_disk_file_object(self):
+        """Testing loading from non-disk non-bytesio file object"""
```
radarhere:
```suggestion
        """Testing loading from non-disk non-BytesIO file object"""
```

PR #4000, opened 2019-07-31
Fix Screengrab DPI scaling on Windows 10 version 1607+

### LINKED ISSUE #2438: Enable DPI Scaling Awareness

nulano:
Fixes #2438 and its duplicates #3432, #3626 for Windows 10 version 1607+ (Anniversary Update).

This PR uses [`SetThreadDpiAwarenessContext`](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setthreaddpiawarenesscontext) to temporarily make Python DPI aware. This function returns the previous awareness context, which is then used to restore the old value, thus avoiding problems with other parts of Python. It was added in Windows 10 version 1607, which is why it is called dynamically.

The issue is not fixed for older versions of Windows. This is not a big problem for Windows 7, as *very* few people use DPI scaling there. Windows 8 and Windows 10 pre-1607 are now out of standard support by Microsoft.


src/display.c
```diff
@@ -329,6 +331,9 @@ PyImaging_GrabScreenWin32(PyObject* self, PyObject* args)
     HDC screen, screen_copy;
     DWORD rop;
     PyObject* buffer;
+    HANDLE dpiAwareness;
+	HMODULE user32 = LoadLibraryA("User32.dll");
+	Func_SetThreadDpiAwarenessContext SetThreadDpiAwarenessContext_function;
```
nulano: Can someone suggest a better name here? This the best I could do but it's ridiculously long.

hugovk: At least it's descriptive! :)

hugovk:
PR 4000! 🍰

PR #4003, opened 2019-08-02
Fix bug when merging identical images to GIF with a list of durations

### LINKED ISSUE #4002: GIF error for multiple duration values
### What did you do?
Using same images  and a list of durations to produce a gif

produce the gif the specified durations

### What actually happened?
 failed when saving

* OS: win10
* Python: 3.6
* Pillow: 6.1.0

to reproduce, add following testcase to `Tests/test_file_gif.py`, then run `pytest Tests/test_file_gif.py`
```python
    def test_totally_identical_frames(self):
        duration_list = [1000, 1500, 2000, 4000]

        out = self.tempfile("temp.gif")

        image_path = "Tests/images/bc7-argb-8bpp_MipMaps-1.png"
        im_list = [
            Image.open(image_path),
            Image.open(image_path),
            Image.open(image_path),
            Image.open(image_path),
        ]
        mask = Image.new("RGBA", im_list[0].size, (255, 255, 255, 0))

        frames = []
        for image in im_list:
            frames.append(Image.alpha_composite(mask, image))

        # duration as list
        frames[0].save(out,
                       save_all=True, append_images=frames[1:], optimize=False, duration=duration_list, loop=0,
                       transparency=0)

        reread = Image.open(out)

        # Assert that the first three frames were combined
        self.assertEqual(reread.n_frames, 1)

        # Assert that the new duration is the total of the identical frames
        self.assertEqual(reread.info["duration"], 8500)

```

err info is:

        if "duration" in im.encoderinfo:
            duration = int(im.encoderinfo["duration"] / 10)
            TypeError: unsupported operand type(s) for /: 'list' and 'int' 

https://github.com/python-pillow/Pillow/blob/f3f45cfec50a44970f325e6ef1a2e4eaf2caa89e/src/PIL/GifImagePlugin.py#L473-L477

seems like we should copy duration info from `im_frames[0]` to `im` when `len(im_frames) == 1`

djy0:
Fixes #4002

 * copy `duration` from `im_frames` to `im.encoderinfo`

### DIFF
--- a/Tests/test_file_gif.py
+++ b/Tests/test_file_gif.py
@@ -514,10 +514,16 @@ class TestFileGif(PillowTestCase):
             frames.append(Image.alpha_composite(mask, image))
 
         # duration as list
-        frames[0].save(out,
-                       save_all=True, append_images=frames[1:], optimize=False, duration=duration_list, loop=0,
-                       transparency=0)
-
+        frames[0].save(
+            out,
+            save_all=True,
+            append_images=frames[1:],
+            optimize=False,
+            duration=duration_list,
+            loop=0,
+            transparency=0,
+        )
+        
         reread = Image.open(out)
 
         # Assert that the first three frames were combined
--- a/Tests/test_file_gif.py
+++ b/Tests/test_file_gif.py
@@ -523,7 +523,6 @@ class TestFileGif(PillowTestCase):
             loop=0,
             transparency=0,
         )
-        
         reread = Image.open(out)
 
         # Assert that the first three frames were combined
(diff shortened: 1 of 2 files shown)


Tests/test_file_gif.py
```diff
@@ -495,6 +495,42 @@ def test_identical_frames(self):
...
+            save_all=True,
+            append_images=frames[1:],
+            optimize=False,
+            duration=duration_list,
+            loop=0,
+            transparency=0,
+        )
+        reread = Image.open(out)
+
+        # Assert that the first three frames were combined
```
radarhere: 'first three frames' - is this correct? Are there not four frames that are combined into one?

djy0: fixed

radarhere: Thanks

PR #4004, opened 2019-08-02
Improved HSV conversion

### LINKED ISSUE #3997: Converting P mode images to HSV fails with "ValueError: conversion not supported" in version 6.0.0 and later
### What did you do?
I want to take an RGB PNG as input, quantize it, convert to HSV and call getcolors() on the result to see a listing of (pixel count, (H,S,V)) values.  As a bigger-picture idea of what I'm trying to accomplish, the goal is to look for a certain hue and determine if the number of pixels that are "close" to that hue are above a certain threshold.  Eg: "Do more than 10% of the pixels in the image have a hue near 250?"  The information from getcolors helps quite a bit with answering that question and the quantization helps with the "near 250" part of the question.  The code below is what I've been using, and it works fine up through Pillow 5.4.1.

Pillow 5.4.1 produces this output when calling getcolors():
```python
[(3601, (244, 227, 146)),
 (3605, (155, 255, 128)),
 (21059, (117, 37, 171)),
 (44719, (63, 195, 136)),
 (6406, (26, 137, 250)),
 (4166, (11, 203, 237)),
 (1912, (194, 184, 255)),
 (5664, (159, 220, 255)),
 (411983, (0, 0, 255)),
 (1137, (0, 0, 0))]
```

### What actually happened?
Pillow 6.0.0 and later throws an exception as part of the convert() call:
```
ValueError                                Traceback (most recent call last)
script.py in <module>()
      1 from PIL import Image
      2 x = Image.open('PNG-Gradient_hex.png').quantize(colors=10,method=1)
----> 3 print x.convert(mode='HSV').getcolors()

PIL/Image.pyc in convert(self, mode, matrix, dither, palette, colors)
   1030                 # normalize source image and try again
   1031                 im = self.im.convert(getmodebase(self.mode))
-> 1032                 im = im.convert(mode, dither)
   1033             except KeyError:
   1034                 raise ValueError("illegal conversion")

ValueError: conversion not supported
```

* OS: macOS Mojave (10.14.5)
* Python: 2.7.16
* Pillow: 6.0.0 and later

```python
from PIL import Image
x = Image.open('PNG-Gradient_hex.png').quantize(colors=10,method=1)
print x.convert(mode='HSV').getcolors()
```

Sample input image:
PNG-Gradient_hex

radarhere:
Resolves #3997

The issue reports that you can't convert from P to HSV, whereas this previously worked.

If a direct conversion from a mode to another is not possible, then Pillow attempts to convert the image to its base mode first, and then convert from that intermediate image.

https://github.com/python-pillow/Pillow/blob/f3f45cfec50a44970f325e6ef1a2e4eaf2caa89e/src/PIL/Image.py#L1050-L1056

PR #3726 changed the base mode of P from RGB to P. So while Pillow previously worked around the absence of a direct method for converting from P to HSV, it now throws an error.

This PR fixes that by assigning direct methods for conversion from P to HSV, as well as 1 to HSV, L to HSV, LA to HSV, I to HSV, RGBA to HSV, RGBX to HSV and CMYK to HSV.

### DIFF
--- a/src/libImaging/Convert.c
+++ b/src/libImaging/Convert.c
@@ -101,6 +101,19 @@ bit2ycbcr(UINT8* out, const UINT8* in, int xsize)
     }
 }
 
+static void
+bit2hsv(UINT8* out, const UINT8* in, int xsize)
+{
+    int x;
+    for (x = 0; x < xsize; x++, out += 4) {
+        UINT8 v = (*in++ != 0) ? 255 : 0;
+        out[0] = 0;
+        out[1] = 0;
+        out[2] = v;
+        out[3] = 255;
+    }
+}
+
 /* ----------------- */
 /* RGB/L conversions */
 /* ----------------- */
@@ -176,6 +189,19 @@ l2rgb(UINT8* out, const UINT8* in, int xsize)
 }
 
 static void
+l2hsv(UINT8* out, const UINT8* in, int xsize)
+{
+    int x;
+    for (x = 0; x < xsize; x++, out += 4) {
+        UINT8 v = *in++;
+        out[0] = 0;
+        out[1] = 0;
+        out[2] = v;
+        out[3] = 255;
+    }
+}
+
+static void
 la2l(UINT8* out, const UINT8* in, int xsize)
 {
     int x;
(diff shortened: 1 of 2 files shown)


src/libImaging/Convert.c
```diff
@@ -631,6 +690,24 @@ i2rgb(UINT8* out, const UINT8* in_, int xsize)
...
+    INT32* in = (INT32*) in_;
+    for (x = 0; x < xsize; x++, in++, out+=4) {
+        out[0] = 0;
+        out[1] = 0;
+        if (*in <= 0)
+            out[2] = 0;
+        else if (*in >= 255)
+            out[2] = 255;
+        else
+            out[2] = (UINT8) *in;
```
hugovk: Let's bracket these single-liners for safety.

radarhere: Okay, done

PR #4009, opened 2019-08-04
Corrected tag types

radarhere:
As per https://github.com/python-pillow/Pillow/pull/3981#issuecomment-514552819


hugovk:
Temporarily adding the "Do not merge" label so I don't accidentally merge this in 6.2.0!

PR #1419, opened 2015-09-11
Tiff ImageFileDirectory Rewrite

wiredfool:
This is my take on #1059, which is a pretty full rewrite of the Tiff ImageFileDirectory for correctness.  This is not complete, as there are still several issues that need to be addressed. 

I had several issues with the PR, which I've attempted to fix here.
* The implementation of the legacy api did not pass the tests out of the box on current master
* Remerge/rebase was required
* Bad Exif tolerance was removed at one point. 
* TiffTags api has changed

This PR:
* Includes #1059
* Renames all accesses to the new api to _v2 suffixes
* Keeps a shell that works with the original unversioned interface
* Restores/adds checks for read length that would result in struct errors otherwise
* Versions the TiffTags data and retains legacy tags without type information
* Most of the current master branch tests pass, with these exceptions:
   - `test_file_tiff.py:test_page_number_x_0` segfaults
   - `test_file_tiff.py:test_xyres_tiff` had a portion where integer resolution was tested, this is stubbed out. 
  - There are two tests that don't have the im.tag.tags[tagname] populated, where they were before. This is arguably an internal interface that shouldn't be tested. If the tag had been read, it would be populated, and would be accessible. (note that this interface hasn't been documented as being private for very long)
  - There are several tests that test implementation specific details that can be dropped now. 

Needs
- [x] One test has been disabled due to a segfault: `test_file_tiff.py:test_page_number_x_0`
- [x] test_file_libtiff is segfaulting
- [x] One test has less coverage than it did: `test_file_tiff.py:test_xyres_tiff`
- [x] Rebase
- [x] Documentation of the appropriate interfaces to use. 
- [x] TiffTags
- [x] `test_file_tiff_metadata.py`
- [x] lingering test failures when run against master's test suite

My intention is that this land before 3.0.0. Please test and comment.

PR #4015, opened 2019-08-08
Corrected documentation of TIFF saving arguments

### LINKED ISSUE #4013: Unclear (incorrect?) documentation of "dpi" parameter for saving TIFF images
# What did you do?

I wanted to save a TIFF image.

On https://pillow.readthedocs.io/en/latest/handbook/image-file-formats.html#saving-tiff-images, I read the following documentation for the `dpi` parameter (below the `resolution`, `x_resolution`, `y_resolution` parameters):

> Either a Float, 2 tuple of (numerator, denominator) or a IFDRational. Resolution implies an equal x and y resolution, dpi also implies a unit of inches.

I read this as:

* I can specify _one_ numeric value for the `dpi` parameter ("Either **a** Float, **[a]** 2 tuple of (numerator, denominator) or **a** IFDRational."
* This one numeric value will specify the x and the y resolution in dots per inch
* The format of this one numeric value can either be
  - a `float`
  - a tuple representing one rational number (numerator, denominator)
  - an IFDRational number 

I wanted to save an image with an x- and a y-resolution of 300 dpi (numerator 300, denominator 1). I called

```python
my_image.save('my_filename.tif', dpi=(300, 1))
```

To get an image with an x- and a y-resolution of 300 dpi. 

### What actually happened?

I got an image with an x-resolution of 300 dpi and a y-resolution of 1 dpi

* OS: debian 10
* Python: 3.7.3
* Pillow:  6.1.0

radarhere:
Resolves #4013, which reports that the `dpi` argument in TIFF saving doesn't support '(numerator, denominator)' values as the documentation claims.

The documentation was added in #369

Looking at code, this functionality was removed in #1419, where 'A tuple of more than one rational tuples' 'flatten to floats' was removed, along with `ifd[X_RESOLUTION] = _cvt_res(dpi[0])` being changed to `ifd[X_RESOLUTION] = dpi[0]`.

While updating this, I also note that we round the dpi - that was added in #3709

Also removed in #1419 was the use of "inch", "cm" and "centimeter" for `resolution_unit`, so I'm fixing that as well.

### DIFF
--- a/docs/handbook/image-file-formats.rst
+++ b/docs/handbook/image-file-formats.rst
@@ -759,18 +759,21 @@ using the general tags available through tiffinfo.
     Strings
 
 **resolution_unit**
-    A string of "inch", "centimeter" or "cm"
+    An integer. 1 for no unit, 2 for inches and 3 for centimeters.
 
 **resolution**
+    Either an integer or a float, used for both the x and y resolution.
 
 **x_resolution**
+    Either an integer or a float.
 
 **y_resolution**
+    Either an integer or a float.
 
 **dpi**
-    Either a Float, 2 tuple of (numerator, denominator) or a
-    :py:class:`~PIL.TiffImagePlugin.IFDRational`. Resolution implies
-    an equal x and y resolution, dpi also implies a unit of inches.
+    A tuple of (x_resolution, y_resolution), with inches as the resolution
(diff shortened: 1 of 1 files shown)


hugovk:
Thanks.

PR #4020, opened 2019-08-12
Note a Windows limit on opening fonts

### LINKED ISSUE #3730: "OSError: cannot open resource" when trying to load more than exactly 509 ImageFonts
### What did you do?
For a data generator, I need to use lots of fonts in different sizes. Randomly one of that is used to generate a sample. Since I didn't want to load a font every time we generate a sample, I created a nested dictionary that dynamicalle loads fonts of a given size when it wasn't loaded previously. It worked fine at the beginning. Then I tried to generate some more samples (>1k) and got "OSError: cannot open resource" every time. I double checked the paths and tried to load the font with the same path and that worked just fine. It definitely isn't related to the path encoding bugs.

To reproduce that, you can use the code below.

It should be able to load fonts until the RAM is full ;)

### What actually happened?
OSError after loading exactly 509 fonts.
After clearing my dictionary as soon as the error occurs (temporary workaround), it works fine again.

* OS: Win10
* Python: 3.6
* Pillow: 5.4.1

```python
from PIL import ImageFont

test = []

for i in range(1000):
        test.append(ImageFont.truetype("fonts/AbyssinicaSIL-R.ttf", 15))
```

radarhere:
Resolves #3730

The issue reports `OSError: cannot open resource` when opening many fonts on Windows. This is a FreeType error that is thrown, presumably because on Windows, ['The C run-time libraries have a 512 limit for the number of files that can be open at any one time.'](https://docs.microsoft.com/en-us/cpp/c-runtime-library/file-handling?view=vs-2017)

This PR suggests resolving this situation by simply documenting the limitation.

Also improved some other FreeType documentation while I'm here.

### DIFF
--- a/src/PIL/ImageFont.py
+++ b/src/PIL/ImageFont.py
@@ -549,8 +549,12 @@ def truetype(font=None, size=10, index=0, encoding="", layout_engine=None):
     This function requires the _imagingft service.
 
     :param font: A filename or file-like object containing a TrueType font.
-                     Under Windows, if the file is not found in this filename,
-                     the loader also looks in Windows :file:`fonts/` directory.
+                 If the file is not found in this filename, the loader may also
+                 search in other directories, such as the :file:`fonts/`
+                 directory on Windows or :file:`/Library/Fonts/`,
+                 :file:`/System/Library/Fonts/` and :file:`~/Library/Fonts/` on
+                 macOS.
+
     :param size: The requested size, in points.
     :param index: Which font face to load (default is first available face).
     :param encoding: Which font encoding to use (default is Unicode). Common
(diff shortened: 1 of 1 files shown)


src/PIL/ImageFont.py
```diff
@@ -546,11 +546,20 @@ def truetype(font=None, size=10, index=0, encoding="", layout_engine=None):
     This function loads a font object from the given file or file-like
     object, and creates a font object for a font of the given size.
 
+    Note that Pillow uses FreeType to open font files. If you are opening many
```
hugovk: No need for "Note that":

```suggestion
   Pillow uses FreeType to open font files. If you are opening many
```

hugovk:
Thanks!

ismael-elatifi:
Good to document this issue. You could also add the BytesIO workaround in the doc.

PR #4025, opened 2019-08-16
Updated ImageShow documentation

radarhere:
ImageShow uses PNG for Unix platforms - see #2527 and https://github.com/python-pillow/Pillow/blob/master/src/PIL/ImageShow.py#L167


hugovk:
Thanks.

PR #4031, opened 2019-08-18
Lazily use ImageFileDirectory_v1 values from Exif

radarhere:
ImageFileDirectory_v1 is lazy, [performing operations when a value is accessed](https://github.com/python-pillow/Pillow/blob/cb7d8d622a3c17aea55d2f989f89300532161120/src/PIL/TiffImagePlugin.py#L968).

However, [accessing all of those values to create a dictionary](https://github.com/python-pillow/Pillow/blob/cb7d8d622a3c17aea55d2f989f89300532161120/src/PIL/Image.py#L3152) when [loading the data](https://github.com/python-pillow/Pillow/blob/cb7d8d622a3c17aea55d2f989f89300532161120/src/PIL/Image.py#L3184) undoes that speed increase.

#3663 set out to fix this by adding a new class - however, it had conflicts once #3625 was merged and a different class was added to deal with Exif data. This PR takes a different approach towards the original goal, for the new context.

It also uses this change by swapping calls in JpegImagePlugin and MpoImagePlugin from `_getexif` to `getexif`, switching from [JpegImagePlugin creating a dict of all the values](https://github.com/python-pillow/Pillow/blob/cb7d8d622a3c17aea55d2f989f89300532161120/src/PIL/JpegImagePlugin.py#L496) to using the now-lazy Exif class.

### DIFF
--- a/src/PIL/JpegImagePlugin.py
+++ b/src/PIL/JpegImagePlugin.py
@@ -158,7 +158,7 @@ def APP(self, marker):
     # If DPI isn't in JPEG header, fetch from EXIF
     if "dpi" not in self.info and "exif" in self.info:
         try:
-            exif = self._getexif()
+            exif = self.getexif()
             resolution_unit = exif[0x0128]
             x_resolution = exif[0x011A]
             try:
--- a/src/PIL/JpegImagePlugin.py
+++ b/src/PIL/JpegImagePlugin.py
@@ -485,19 +485,9 @@ def _fixup_dict(src_dict):
 
 
 def _getexif(self):
-    # Use the cached version if possible
-    try:
-        return self.info["parsed_exif"]
-    except KeyError:
-        pass
-
     if "exif" not in self.info:
         return None
-    exif = dict(self.getexif())
-
-    # Cache the result for future use
-    self.info["parsed_exif"] = exif
-    return exif
+    return dict(self.getexif())
 
 
 def _getmp(self):
--- a/src/PIL/MpoImagePlugin.py
+++ b/src/PIL/MpoImagePlugin.py
@@ -86,8 +86,6 @@ class MpoImageFile(JpegImagePlugin.JpegImageFile):
         self.offset = self.__mpoffsets[frame]
 
         self.fp.seek(self.offset + 2)  # skip SOI marker
-        if "parsed_exif" in self.info:
-            del self.info["parsed_exif"]
         if i16(self.fp.read(2)) == 0xFFE1:  # APP1
             n = i16(self.fp.read(2)) - 2
             self.info["exif"] = ImageFile._safe_read(self.fp, n)
--- a/src/PIL/MpoImagePlugin.py
+++ b/src/PIL/MpoImagePlugin.py
@@ -92,7 +92,7 @@ class MpoImageFile(JpegImagePlugin.JpegImageFile):
             n = i16(self.fp.read(2)) - 2
             self.info["exif"] = ImageFile._safe_read(self.fp, n)
 
-            exif = self._getexif()
+            exif = self.getexif()
             if 40962 in exif and 40963 in exif:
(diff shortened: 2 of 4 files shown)


src/PIL/JpegImagePlugin.py
```diff
@@ -158,7 +158,7 @@ def APP(self, marker):
     # If DPI isn't in JPEG header, fetch from EXIF
     if "dpi" not in self.info and "exif" in self.info:
         try:
-            exif = self._getexif()
+            exif = self.getexif()
```
Glandos: The main difference is that `getexif` from `Image` class does not cache the parsed Exif tags.
So this call is done and forget when the user will need it.

This is not what we should expect.

src/PIL/MpoImagePlugin.py
```diff
@@ -92,7 +92,7 @@ def seek(self, frame):
             n = i16(self.fp.read(2)) - 2
             self.info["exif"] = ImageFile._safe_read(self.fp, n)
 
-            exif = self._getexif()
+            exif = self.getexif()
```
Glandos: The main difference is that `getexif` from `Image` class does not cache the parsed Exif tags.
So this call is done and forget when the user will need it.

This is not what we should expect.

radarhere: I'm not following, sorry. Yes, using `_getexif` would create a cache of the parsed exif tags, but it would also iterate over all the values.

From my understanding, the point of these PRs is to reduce operations on load to potentially increase speed - we don't need to work out all of the EXIF values when a user might not even be interested in them. When JpegImagePlugin and MpoImagePlugin only need two EXIF values each, isn't decoding all the values overkill?

Glandos: Sorry for the unclear review.
Yes, calling `getexif` is better, but right now, when reading an image, the `_getexif` function is called, and the result is cached so that if the user called it once again, the computation will not happen a second time.
This is not the case with `getexif`. If I trace a call to `Image.open("my_image.jpg").getexif()`, then the dictionary loading and some decoding will happens twice.

In my opinion, the `Exif` object should be stored into the `Image` instance once loaded. And if the user needs to modify it, this object could be cloned.

I hope it is clearer now :smile:

radarhere: Okay, see what you think of this idea - I have pushed a commit that changes `im.getexif()` so that it returns a shared instance of the Exif class, meaning that `getexif` decoding of EXIF values is now cached. This actually makes the `_getexif` "parsed_exif" cache largely redundant, because `_getexif` uses `getexif`, so I have removed that. Hopefully, that should make all scenarios faster.

Glandos: Very nice now :) Iterating 1000 times over an image (`open().getexif()`), I go from 18.0s to 8.11s. Confirmed by `%timeit` in IPython, going from 3.69ms per loop to 1.68ms per loop.

The last step (for me) is to get rid of `_fixup_dict` call of the tag 0x8769, but this should be in another PR.

hugovk:
Thanks @radarhere for the PR and @Glandos for the review (and other PRs)!

Most of the changes are covered by existing tests. 👍

Would this benefit from adding simple tests to directly test some of the changed/added functions?

If I'm following this report correctly, it looks like at least this `__str__` isn't covered by any tests at all:

image


https://codecov.io/gh/python-pillow/Pillow/compare/a44e918a5b647d5eb75e0bf3fe3df4d6501cb78d...f3ed44a5662ea2728fae2139bbdde9419356302a/diff#D1-3299

radarhere:
Okay, I've added a test for `__str__`.

The externally visible changes are the removal of `info["parsed_exif"]` and the fact that `getexif()` now returns a shared instance. I've added a test for the shared instance.

hugovk:
Thanks all!

Issue #3773, opened 2019-04-05
 No preview of .ico file on Windows 10


MagTun:
### What did you do and what did you expect to happen?
I ran this code and tried to preview the `.ico` file created in Windows folder explorer, Windows Photo Viewer and  Windows Photo: 
```python
from PIL import Image
img = Image.new('RGB', (50, 50), color = (130, 173, 242))  
img.save('pillow.ico')
```
### What actually happened?
Although I can preview others `.ico` files on my computer, I can't preview the  `.ico` created by Pillow  (the file on the left is an `.ico` created by pillow, the one on the right is "genuine" `ico`)

- Windows folder explorer
![2019-04-05_18-02-08_explorer](https://user-images.githubusercontent.com/15843700/55640942-f6f97800-57cc-11e9-8488-d56a95b3d90c.png)

- Windows Photo Viewer  
![2019-04-05_17-55-58_explorer](https://user-images.githubusercontent.com/15843700/55640583-3b384880-57cc-11e9-9ad6-f29ef65db6e2.png)

- Windows Photo
![2019-04-05_17-55-09_explorer](https://user-images.githubusercontent.com/15843700/55640593-40959300-57cc-11e9-97db-89748cef4d82.png)

It was tested on 2 computers (same OS, Python and Pillow versions)
I can preview the `.ico` if I drag them in chrome. I can also preview them if I load them as an icon for a systray app. 

### What are your OS, Python and Pillow versions?

* OS: Windows 10
* Python:  3.6
* Pillow:   5.1.0

radarhere:
Could you upload an ICO file that previews correctly, and the Pillow generated one that doesn't?

MagTun:
Here is a [link ](https://drive.google.com/open?id=1_ppImgx4S7riMuBfNpWUJmBwF4J5z6U9)to the `.ico`. `pillow.ico` is the one that isn't working, `190.ico` is the one that is working (it comes from Windows shell32.dll) 

I can see the preview of the ico in google drive but I am still unable to see them in Windows folder explorer, Windows Photo Viewer, Windows Photo.

radarhere:
Thanks. Looking at the two files, pillow.ico has the sizes (16, 16), (24, 24), (32, 32) and (48, 48), while 190.ico just has (40, 40).

[...]

PR #4032, opened 2019-08-19
Documented OS support for ICO saved files

radarhere:
https://github.com/python-pillow/Pillow/issues/3773#issuecomment-509160651

> it should also mention that the current Pillow implementation saves ICOs internally as PNGs, not DIBs (support was only added in Windows Vista).

See https://github.com/python-pillow/Pillow/blob/cb7d8d622a3c17aea55d2f989f89300532161120/src/PIL/IcoImagePlugin.py#L72

and

https://en.wikipedia.org/wiki/ICO_(file_format)
> The ability to read PNG images from ICO and CUR format images was introduced in Windows Vista.

### DIFF
--- a/src/PIL/IcoImagePlugin.py
+++ b/src/PIL/IcoImagePlugin.py
@@ -263,6 +263,9 @@ class IcoImageFile(ImageFile.ImageFile):
 
     Handles classic, XP and Vista icon formats.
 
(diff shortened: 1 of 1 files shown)


radarhere:
Okay, I've updated the text to be clearer.

PR #4041, opened 2019-08-30
Simplified op declarations

radarhere:


src/PIL/TiffImagePlugin.py
```diff
@@ -370,42 +370,37 @@ def delegate(self, *args):
 
         return delegate
 
-    """ a = ['add','radd', 'sub', 'rsub','div', 'rdiv', 'mul', 'rmul',
-             'truediv', 'rtruediv', 'floordiv',
-             'rfloordiv','mod','rmod', 'pow','rpow', 'pos', 'neg',
-             'abs', 'trunc', 'lt', 'gt', 'le', 'ge', 'nonzero',
-             'ceil', 'floor', 'round']
-        print("\n".join("__%s__ = _delegate('__%s__')" % (s,s) for s in a))
-        """
```
hugovk: Did this do nothing?

src/PIL/TiffImagePlugin.py
```diff
@@ -370,42 +370,37 @@ def delegate(self, *args):
...
+        "lt",
+        "gt",
+        "le",
+        "ge",
+        "nonzero",
+        "ceil",
+        "floor",
+        "round",
+    ):
+        vars()["__" + op + "__"] = _delegate("__" + op + "__")
```
hugovk: Hmm, this makes it a bit easier to add new ops, but I think it's less readable.

A comment to make it clear what's happening may help, eg.

```python
# Equivalent to __add__ = _delegate("__add__") etc.
```

src/PIL/TiffImagePlugin.py
radarhere: I think it's intended as instructions for how to generate the declarations

src/PIL/TiffImagePlugin.py
radarhere: Yes, I could see arguments either way for this. I've added a commit with your comment.

hugovk: Thanks for the update. I'm +0 in this, so I'll leave it up to you (or someone else) :)

radarhere:
Closing due to debate over whether this is simpler or not.

PR #4050, opened 2019-09-06
Use double backticks in RST

hugovk:
reStructuredText uses double backticks for code. Markdown uses single backticks.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -88,7 +88,7 @@ Changelog (Pillow)
 - Updated TIFF tile descriptors to match current decoding functionality #3795
   [dmnisson]
 
-- Added an `image.entropy()` method (second revision) #3608
+- Added an ``image.entropy()`` method (second revision) #3608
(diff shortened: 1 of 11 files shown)


radarhere:
While I agree that it looks better, do you have a link for RST documentation saying this?

hugovk:
Here we are:

> Text enclosed by double-backquotes is treated as inline literals

* http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#inline-literals

> Interpreted text is text that is meant to be related, indexed, linked, summarized, or otherwise processed, but the text itself is typically left alone. Interpreted text is enclosed by single backquote characters

* http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#interpreted-text

Also https://github.com/ralsina/rst-cheatsheet/blob/master/rst-cheatsheet.rst

Single backticks are shown as italics on RTD, which is not the intended formatting for these.

PR #3978, opened 2019-07-19
Added text stroking

radarhere:
Resolves #2209. Replacement for #2224

Adds `strokeWidth` and `strokeFill` arguments, such that -

```python
from PIL import Image, ImageDraw, ImageFont

font = ImageFont.truetype("Tests/fonts/FreeMono.ttf", 40)
font.getsize_multiline("A", strokeWidth=2)
font.getsize("ABC\nAaaa", strokeWidth=2)

im = Image.new("RGB", (100, 100))
draw = ImageDraw.Draw(im)
draw.textsize("A", font, strokeWidth=2)
draw.multiline_textsize("ABC\nAaaa", font, strokeWidth=2)
draw.text((10, 10), "A", "#f00", font, strokeWidth=2, strokeFill="#0f0")
draw.multiline_text((10, 10), "A\nB", "#f00", font, strokeWidth=2, strokeFill="#0f0")
```

Uses FreeType's stroking. See https://www.freetype.org/freetype2/docs/reference/ft2-glyph_stroker.html for more information, and part of https://www.freetype.org/freetype2/docs/tutorial/example2.cpp for example code.

Also renames a FT_GlyphSlot variable from 'glyph' to 'glyphSlot'.

PR #4057, opened 2019-09-11
Increased tolerance for stroke test comparison

radarhere:
Since merging #3978, I've been seeing some intermittent failures from Travis and Azure.

For example,
- https://travis-ci.org/python-pillow/Pillow/jobs/583125917 - but if you rerun the build, there is a decent chance it will pass.
- https://dev.azure.com/python-pillow/pillow/_build/results?buildId=806
- https://dev.azure.com/python-pillow/pillow/_build/results?buildId=812
- https://dev.azure.com/python-pillow/pillow/_build/results?buildId=817

This PR increases the tolerance of that test.

### DIFF
--- a/Tests/test_imagedraw.py
+++ b/Tests/test_imagedraw.py
@@ -831,7 +831,7 @@ class TestImageDraw(PillowTestCase):
 
             # Assert
             self.assert_image_similar(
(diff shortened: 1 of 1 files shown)


hugovk:
Thanks.

PR #4062, opened 2019-09-13
Added ImageOps auto_transpose

### LINKED ISSUE #4053: TIFF orientation tag handled differently from most other software when loading G4 TIFF files
Most image software handles TIFF orientation tags  (tag 274) differently from Pillow when loading  binary G4 TIFF images. This means that only images having the orientation tag == 1 (The 0th row represents the visual top of the image, and the 0th column represents the visual left-hand side) are loaded correctly (in the same way as in other software), other images should be corrected manually. 
I am not sure whether Pillow regards the orientation tags while other software ignores or vice versa. At least  they are loaded transposed.
I use Pillow-6.1.0 with Python 3.8 on Windows.

I've developed the following procedure in my script to load TIFF files correctly (basically I only have files with orientation==8 from my office copier machine, so I have not tested other orientations). I think this something similar should be part of the TIFFImagePlugin

```python
from PIL import Image
from tkinter import filedialog, Tk
ORIENTATION = 274
root = Tk()
root.withdraw()
file_selected = filedialog.askopenfilename()
srcim = Image.open(filename)
orientation = srcim.tag[ORIENTATION][0]
format = os.path.splitext(filename)[1]
if format.lower() in [".tif", ".tiff"]:
     try:
        if orientation == 8:
            srcim = srcim.transpose(Image.ROTATE_90)
        elif orientation == 3:
            srcim = srcim.transpose(Image.ROTATE_180)
        elif orientation == 6:
            srcim = srcim.transpose(Image.ROTATE_270)
        elif orientation == 2:
            srcim = srcim.transpose(Image.FLIP_LEFT_RIGHT)
        elif orientation == 4:
            srcim = srcim.transpose(Image.FLIP_TOP_BOTTOM)
        elif orientation == 5:
            srcim = srcim.transpose(Image.ROTATE_90).transpose(Image.FLIP_LEFT_RIGHT)
        elif orientation == 7:
            srcim = srcim.transpose(Image.ROTATE_270).transpose(Image.FLIP_LEFT_RIGHT)
     except:
          print("Can't detect orientation")
```

radarhere:
Resolves #4053

The issue points out in addition to the EXIF orientation tag, there is also a TIFF orientation tag -  https://www.awaresystems.be/imaging/tiff/tifftags/orientation.html

Unfortunately, the auto transposing method for EXIF orientation was named `exif_transpose`, so I'm not able to just bundle the handling of the TIFF tag into that. So this PR adds `auto_transpose` to transpose either the EXIF or the TIFF orientation tag.


ailintom:
I guess, correcting the orientation of TIFF images as per Orientation tag should not in any way be linked to processing the EXIF tag. TIFF Orientation is a much more technical thing, like byte order or compression and should be applied automatically to all TIFF images that have Orientation differing from 1. It is a baseline tag, and it is not up to the user to decide whether they want to read TIFF images in the correct orientation just like it is not up to the user to decide whether they want to decode LZW compressed TIFFs or leave them compressed. (But some applications like the current version of Pillow simply do not support orientations other than 1, like some applications do not support some compressions).

And whether to apply EXIF orientation is up to the user or the developer. This correction should be initiated by a separate procedure call. (Like it is already implemented in Pillow).

ailintom:
Eventually, a file can have both TIFF orientation (TIFF tag 274) and EXIF  (TIFF tag 34665) set. In this case, Pillow should first read the TIFF file correctly, using TIFF tag 274, and then the developer can either call exif_transpose or not.

radarhere:
Okay. See #4063 instead

PR #4063, opened 2019-09-13
Use TIFF orientation

### LINKED ISSUE #4053: TIFF orientation tag handled differently from most other software when loading G4 TIFF files

radarhere:
Resolves #4053. Alternative to #4062

Transpose TIFF images on load according to the orientation tag -  https://www.awaresystems.be/imaging/tiff/tifftags/orientation.html


seb-odoo:
Hello @ailintom do you know if there is any plan to also use the orientation for JPEG files?

ailintom:
@seb-odoo JPEG files do not have an orientation tag comparable to the TIFF orientation tag. Unlike TIFF-files, 

> in JFIF files, the image orientation is always top-down

see https://www.w3.org/Graphics/JPEG/jfif3.pdf 
JPEG files however can have EXIF tags, including in the orientation EXIF tag, indicating the orientation of camera when shooting. You can transform the image according to the EXIF orientation tag using PIL by calling `ImageOps.exif_transpose`

Issue #3606, opened 2019-01-25
CI Review


wiredfool:
With the news that travis has been acquired, I've seen a bunch of discussion about CI platforms, leading to some questions:

* Is the current CI system covering what we need it to do now?
* Is it working with our long term goals?

One of the goals that I had originally with Appveyor was to be able to produce our windows binaries off of a CI system, rather than bothering @cgohlke for them every single time, as he's a single point of failure for the entire windows python ecosystem.   I'm not sure where we are on the windows testing WRT external dependency completeness or python version coverage.  Azure pipelines has now introduced free build time for open source projects, and I believe that the python core is now on that for their packages. I believe that they also do linux builds, but not macos. It may be worth investigating if their environment is any easier to get all the dependencies to build.

hugovk:
It's definitely worth looking into. Building Windows binaries would definitely be great.

More info from GitHub: https://github.blog/2018-09-10-azure-pipelines-now-available-in-github-marketplace/

I've seen a few other projects move over to Azure Pipelines, but haven't looked into it much. It can do Linux, Windows and macOS, and [CPython](https://github.com/python/cpython/), [pip](https://github.com/pypa/pip/) and [NumPy](https://github.com/numpy/numpy) are testing those with Azure.

I think our current CI system is doing quite a good job, it'd be nice for all three platforms to be tested in one place, and especially if Azure is faster (with 10 parallel builds).

Right now:
* Appveyor has just 1 parallel worker and takes ~1h15m for 12 jobs
* Travis CI has 5 parallel workers and takes ~26m for 24 jobs

From a [NumPy discussion](https://mail.python.org/pipermail/numpy-discussion/2018-September/078707.html):

> The "free for open source" tier is 10 concurrent jobs that can run up to 60
minutes each (although the website says 30 minutes), and no limit on
minutes per month.  Like the other CI services, it uses YAML files to
define the build and test process.  I found the documentation of the YAML
pipeline spec to be large, but still missing a number of small details and
explanations about how features worked together.  That said, the build
performance is *significantly* better than Travis CI and AppVeyor for all 3
platforms, and the convenience of all major platforms going through one
service is very nice.  Despite not liking the UI quite as much as Travis
CI, I think Numba will likely switch completely over to Azure assuming the
free offering continues to be this good.

What are our long term goals?

* Run tests time, with good coverage of supported platforms, Python versions and depedencies
* Build binaries

hugovk:
[...]

PR #4066, opened 2019-09-14
Test Docker with GitHub Actions

hugovk:
For https://github.com/python-pillow/Pillow/issues/3606. 

 * Run Docker tests using GitHub Actions
 * GitHub Actions testing now identical to Azure Pipelines, 13 jobs
 * Main difference: 
    * AZP 10 parallel jobs, so 3 wait. Whole run ~6 mins
    * GHA [20 parallel jobs](https://help.github.com/en/articles/about-github-actions#usage-limits), so none wait. Whole run ~3 mins

Shall we remove the AZP tests now or have both enabled for a while to compare and evaluate?

(Coverage still needed before removing Docker from Travis, can follow.)

### DIFF
--- a/.github/workflows/test-docker.yml
+++ b/.github/workflows/test-docker.yml
@@ -1,0 +1,40 @@
+name: Test Docker
+
+on: [push, pull_request]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        docker: [
+          alpine,
+          arch,
+          ubuntu-16.04-xenial-amd64,
+          ubuntu-18.04-bionic-amd64,
+          debian-9-stretch-x86,
+          debian-10-buster-x86,
+          centos-6-amd64,
+          centos-7-amd64,
+          amazon-1-amd64,
+          amazon-2-amd64,
+          fedora-29-amd64,
+          fedora-30-amd64,
+        ]
+        dockerTag: [master]
+
+    name: ${{ matrix.docker }}
+
+    steps:
+    - uses: actions/checkout@v1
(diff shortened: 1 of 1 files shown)


radarhere:
I'm fairly certain that you have referenced the wrong issue.

hugovk:
Thanks, fixed!

.github/workflows/test-docker.yml
```diff
@@ -0,0 +1,40 @@
...
+          amazon-2-amd64,
+          fedora-29-amd64,
+          fedora-30-amd64,
+        ]
+        dockerTag: [master]
+
+    name: ${{ matrix.docker }}
+
+    steps:
+    - uses: actions/checkout@v1
```
radarhere: You'd rather `@v1` instead of
```suggestion
    - uses: actions/checkout@master
```
?

hugovk: Yep:

> We strongly recommend that you include the version of the action you are using by specifying a Git ref, SHA, or Docker tag number. If you don't specify a version, it could break your workflows or cause unexpected behavior when the action owner publishes an update.
> 
> - Using the commit SHA of a released action version is the safest for stability and security.
> - Using the specific major action version allows you to receive critical fixes and security patches while still maintaining compatibility. It also assures that your workflow should still work.
> - Using the `master` branch of an action may be convenient, but if someone releases a new major version with a breaking change, your workflow could break.

https://help.github.com/en/articles/workflow-syntax-for-github-actions#jobsjob_idstepsuses

hugovk:
Okay to merge?

PR #4068, opened 2019-09-16
Corrected comment

radarhere:
Moved the comment so that it is above the correct line.

Added in https://github.com/python-pillow/Pillow/commit/6207b44ab1ff4a91d8ddc7


hugovk:
Well spotted!

PR #4074, opened 2019-09-21
Test ImageShow.show on CIs

radarhere:
This PR adds a test that actually runs the `ImageShow.show` method, opening the image. However, to be neat, it only runs on the CIs, where the shown images won't bother users.


radarhere:
Okay, done.

nulano:
Note, that on Windows this will always either succeed (AppVeyor), or hang (GitHub Actions). It can't fail, even if there is no application associated with the file format.

PR #4077, opened 2019-09-21
AppVeyor build improvements

nulano:
While trying to add PyPy3 to AppVeyor, I found and fixed several winbuild issues, including these:

Log changes:

 * `SetEnv.Cmd` and `vcvarsall.bat` both disable `echo`. Turning it back on improves readability of the build log.
 * `nmake` copyright message was messy (around 30 lines long). Adding `-nologo` turns it off.

Compare before: https://ci.appveyor.com/project/Python-pillow/pillow/builds/27337992/job/ni8rcc28utx9baq6?fullLog=true#L367
Compare after: https://ci.appveyor.com/project/nulano/pillow/builds/27571987/job/cojejreprth4khia?fullLog=true#L367

Build changes:

 * Build libjpeg in release mode, not debug mode, by adding `nodebug=1`. See `jpeg-9c\makefile.vc` line 9 for more info.
* Build openjp2 in release mode, not default mode, by adding `-DCMAKE_BUILD_TYPE=Release`. See e.g. https://codeyarns.com/2015/05/14/build-types-in-cmake/ for more info.

These changes also seem to decrease build times by about 10%:
Before: https://ci.appveyor.com/project/Python-pillow/pillow/builds/27337992
After: https://ci.appveyor.com/project/nulano/pillow/builds/27571987

### DIFF
--- a/winbuild/build.py
+++ b/winbuild/build.py
@@ -84,7 +84,8 @@ def vc_setup(compiler, bit):
         arch = "x86" if bit == 32 else "x86_amd64"
         script = (
             r"""
-call "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" %s"""
+call "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" %s
+echo on"""
             % arch
         )
     return script
--- a/winbuild/build_dep.py
+++ b/winbuild/build_dep.py
@@ -67,7 +67,6 @@ def extract_openjpeg(compiler):
         r"""
 rem build openjpeg
 setlocal
-@echo on
 cd %%BUILD%%
 mkdir %%INCLIB%%\openjpeg-2.0
 copy /Y /B openjpeg-2.0.0-win32-x86\include\openjpeg-2.0  %%INCLIB%%\openjpeg-2.0
@@ -140,7 +139,6 @@ setlocal
 """
         + vc_setup(compiler, bit)
         + r"""
-@echo on
 cd /D %%OPENJPEG%%%(inc_dir)s
 
 %%CMAKE%% -DBUILD_THIRDPARTY:BOOL=OFF -DBUILD_SHARED_LIBS:BOOL=OFF -DCMAKE_BUILD_TYPE=Release -G "NMake Makefiles" .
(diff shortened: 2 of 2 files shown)


hugovk:
The builds times are quite variable, but this is probably an improvement, and welcome for a very slow build.

Here's `winbuild-improve` builds:

50 min 25 sec
45 min 15 sec
Average: 47.8 min

Non-`winbuild-improve` builds:

1 hr 10 min
45 min 47 sec
45 min 40 sec
46 min 14 sec
45 min 12 sec
49 min 16 sec
45 min 5 sec
45 min 35 sec
54 min 29 sec
54 min 49 sec
Average: 50.2 min

Thanks for looking into it!

hugovk:
By the way, as AppVeyor is so slow (because we only get a single serial build job) what would be really great is to add Windows testing using GitHub Actions, where we get 20 parallel  jobs!

Testing just something recent like Python 3.7 would be a great start, and have its own `test-windows.yml` workflow file. For comparison, here's GHA for Travis CI:

* Lint: https://github.com/python-pillow/Pillow/pull/4058
* Linux: https://github.com/python-pillow/Pillow/pull/4081
* Docker: https://github.com/python-pillow/Pillow/pull/4066

nulano:
> By the way, as AppVeyor is so slow (because we only get a single serial build job) what would be really great is to add Windows testing using GitHub Actions, where we get 20 parallel jobs!
> 
> Testing just something recent like Python 3.7 would be a great start, and have its own `test-windows.yml` workflow file. For comparison, here's GHA for Travis CI:
> 
> * Lint: #4058
> * Linux: #4081
> * Docker: #4066

Windows builds are currently running on VS2015 (same as CPython), but AFAICT GitHub Actions only supports VS2017 and VS2019: https://help.github.com/en/articles/software-in-virtual-environments-for-github-actions#visual-studio-2017-enterprise

I have also noticed that GitHub is rejecting any push with commits affecting `.github/workflows` from my Git client (SourceTree), similarly to this issue with GitHub Desktop: https://github.community/t5/GitHub-Actions/Can-I-edit-the-main-workflow-github-actions-file-locally/td-p/26111
That's why I'm basing all my branches on cd880d6 (the last commit without GHA) for now.

PR #4080, opened 2019-09-22
Changed WindowsViewer format to PNG

### LINKED ISSUE #3695: Windows: image with alpha is not displayed correctly
### What did you do?
The pillow package has a method called [`Image.putalpha()`](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.putalpha) which is used to add or change the alpha channel of an image.

I tried to play with this method and found that I can not change the background color of an image. The original image is

owl

This is my code to add alpha to it

```python
from PIL import Image

im_owl = Image.open("owl.jpg")

alpha = Image.new("L", im_owl.size, 50)
im_owl.putalpha(alpha)

im_owl.show()
```
The produced image shows the effect of adding alpha channel to image.

### What actually happened?

The produced image is nothing different from the original image. I have tried with different value of alpha and see no difference.

* OS:  Windows 10 1803
* Python:  3.6.5
* Pillow: 5.3.0

After consulting the  pillow doc about `Image.show()` method, it seems that pillow will convert the Image object to bmp files and show it with default system image viewer.  That explains the reason why no difference between the original and alpha-composited image.

It would be great if the image is shown as its original format, not as bmp files.

radarhere:
Resolves #3695

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -2110,15 +2110,15 @@ class Image(object):
         Displays this image. This method is mainly intended for
         debugging purposes.
 
-        On Unix platforms, this method saves the image to a temporary
-        PNG file, and calls the **display**, **eog** or **xv**
-        utility, depending on which one can be found.
+        The image is first saved to a temporary file. By default, it will be in
+        PNG format.
 
-        On macOS, this method saves the image to a temporary PNG file, and
-        opens it with the native Preview application.
+        On Unix, the image is then opened using the **display**, **eog** or
+        **xv** utility, depending on which one can be found.
 
-        On Windows, it saves the image to a temporary BMP file, and uses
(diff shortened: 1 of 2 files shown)


radarhere:
Here is a screenshot from AppVeyor.

RGB

hugovk:
How do you get a screenshot from AppVeyor?

radarhere:
I used ImageGrab to get the screenshot, saved it to file, then used https://transfer.sh/ to extract it. I imagine you could also use AppVeyor artifacts to similarly get the file out.

nulano:
Another option, which I would have used, is remote desktop:
https://github.com/python-pillow/Pillow/blob/79a9981a639cc635f92e1c0a8b3a4ccaca6f9329/.appveyor.yml#L121-L125

PR #4083, opened 2019-09-23
Cleanup by resetting locale

radarhere:
https://bitbucket.org/pypy/pypy/issues/3079/systemerror-in-timestrptime-with-non#comment-54017374 notes that Pillow does not reset the locale at the end of one of our tests.


hugovk:
Thanks!

PR #4084, opened 2019-09-23
Test Windows with GitHub Actions

nulano:
For #3606 

Adds Windows tests to GitHub Actions.

Existing scripts `winbuild\build_dep.py` and `winbuild\build.py` expect MSVS2015, but GitHub Actions only comes with MSVS2017 (or MSVS2019). It does come with a backwards-compatibility component, but it is incompatible with the existing scripts. For now this uses build scripts hardcoded in `test-windows.yml`, I would like to update the `winbuild` scripts once I have all the dependencies working.

Dependency builds are split into separate steps to make logs easier to read. If kept, this would require another change in `winbuild\build_dep.py`.

## Available dependencies:

* [x] libjpeg
* [x] libjpeg-turbo
* [x] zlib
* [x] webp
* [x] libtiff
* [x] freetype
* [x] lcms2
* [x] openjpeg
* [x] ghostscript
* [x] ~libimagequant~
* [x] ~raqm~

_Strikethrough indicates potential licensing issues for binary distribution._

## TODO

* [x] Only Python 3.6.8 for now. TODO add 3.5.4 and 3.7.4
* [x] Only x86 for now. TODO add x64
* [x] Add PyPy3
* [x] Make sure dependencies are built for Release
* [x] Fix all LNK4098 warnings
* [ ] ~Update `winbuild\build_dep.py` and `winbuild\build.py`~ (in future PR)
* [x] Upload coverage. Depends on #4081 for token.
* [x] Build wheels

PR #4091, opened 2019-09-26
Add libimagequant to features.py

nulano:
From https://github.com/python-pillow/Pillow/pull/4084#issuecomment-535702633

 * Add libimagequant to `features.py`
 * Remove code duplication from `selftest.py` introduced in #3870

### DIFF
--- a/src/PIL/features.py
+++ b/src/PIL/features.py
@@ -56,6 +56,7 @@ features = {
     "transp_webp": ("PIL._webp", "HAVE_TRANSPARENCY"),
     "raqm": ("PIL._imagingft", "HAVE_RAQM"),
     "libjpeg_turbo": ("PIL._imaging", "HAVE_LIBJPEGTURBO"),
+    "libimagequant": ("PIL._imaging", "HAVE_LIBIMAGEQUANT"),
 }
 
 
@@ -94,7 +95,7 @@ def get_supported():
     return ret
 
 
-def pilinfo(out=None):
+def pilinfo(out=None, brief=False):
     if out is None:
         out = sys.stdout
 
@@ -113,11 +114,12 @@ def pilinfo(out=None):
     )
     print("-" * 68, file=out)
 
-    v = sys.version.splitlines()
-    print("Python {}".format(v[0].strip()), file=out)
-    for v in v[1:]:
-        print("       {}".format(v.strip()), file=out)
-    print("-" * 68, file=out)
+    if not brief:
+        v = sys.version.splitlines()
+        print("Python {}".format(v[0].strip()), file=out)
+        for v in v[1:]:
+            print("       {}".format(v.strip()), file=out)
+        print("-" * 68, file=out)
 
     for name, feature in [
         ("pil", "PIL CORE"),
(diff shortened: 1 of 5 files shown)


src/PIL/features.py
```diff
@@ -133,37 +135,41 @@ def pilinfo(out=None):
         ("zlib", "ZLIB (PNG/ZIP)"),
         ("libtiff", "LIBTIFF"),
         ("raqm", "RAQM (Bidirectional Text)"),
+        ("libimagequant", "LIBIMAGEQUANT (quantization method)"),
```
radarhere:
```suggestion
        ("libimagequant", "LIBIMAGEQUANT (Quantization method)"),
```

src/PIL/features.py
```diff
@@ -94,7 +95,7 @@ def get_supported():
     return ret
 
 
-def pilinfo(out=None):
+def pilinfo(out=None, brief=False):
```
radarhere: Suggestion: Instead of 'brief=False', what about 'include_python_version=True'?

hugovk: Or flip it and call it `verbose`?

nulano: @radarhere The significant difference is whether the list of supported image formats is printed.

Thinking about it, I don't think it would be a bad idea to include the Python version in `selftest.py`. The parameter could then be called `supported_formats=True`.

src/PIL/features.py
nulano: Done.

nulano:
I have renamed the `brief=False` parameter to `supported_formats=True`, and added the Python version to the `selftest.py` output.

The output is now (from https://travis-ci.org/nulano/Pillow/jobs/597002577#L2460):
```
python selftest.py
Pillow 6.2.0.dev0
Python 3.7.1 (default, Dec  5 2018, 18:09:53)
       [GCC 5.4.0 20160609]
Python modules loaded from /home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/Pillow-6.2.0.dev0-py3.7-linux-x86_64.egg/PIL
Binary modules loaded from /home/travis/.cache/Python-Eggs/Pillow-6.2.0.dev0-py3.7-linux-x86_64.egg-tmp/PIL
--- PIL CORE support ok
--- TKINTER support ok
--- FREETYPE2 support ok
--- LITTLECMS2 support ok
--- WEBP support ok
--- WEBPMUX support ok
--- WEBP Animation support ok
--- JPEG support ok
--- OPENJPEG (JPEG2000) support ok
--- ZLIB (PNG/ZIP) support ok
--- LIBTIFF support ok
--- RAQM (Bidirectional Text) support ok
--- LIBIMAGEQUANT (Quantization method) support ok
Running selftest:
--- 59 tests passed.
```

Running `python -m PIL` lists supported formats instead of the selftest message at the end, same as before.

src/PIL/features.py
nulano: > Thinking about it, I don't think it would be a bad idea to include the Python version in `selftest.py`. The parameter could then be called `supported_formats=True`.

Done, see comment below.

hugovk:
Thanks!

PR #4094, opened 2019-09-27
Round Spider record count

### LINKED ISSUE #4093: Spider format header bug for random image size
### What did you do?
I am trying to save `float` image in `SPIDER` format

I accepted it to work for any image size.

### What actually happened?
The saved spider file has corrupt header when the width of the image is not a power of 2. 

* OS: Windows 10 
* Python: 3.6
* Pillow: 6.1.1

```python
from PIL import Image
import numpy as np

filename= "im.spi"
width = 100  # change it to any natural number not a power of 2
im = np.random.random((64, width))-1
pil_image = Image.fromarray(im, 'F')
pil_image.save(filename, format='SPIDER')

im2 = Image.open(filename)
print(im2.size)
```

I was able to trace the bug to [this](https://github.com/python-pillow/Pillow/blob/master/src/PIL/SpiderImagePlugin.py#L239) line.
Changing the line from `labrec = 1024 / lenbyt` to `labrec = int(1024 / lenbyt)` fixes the bug for me.

nulano:
Fixes #4093.

### DIFF
--- a/Tests/test_file_spider.py
+++ b/Tests/test_file_spider.py
@@ -1,4 +1,5 @@
 import tempfile
+from io import BytesIO
 
 from PIL import Image, ImageSequence, SpiderImagePlugin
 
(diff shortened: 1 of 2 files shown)


hugovk:
Thanks!

PR #4117, opened 2019-10-06
Clarified image format documentation

### LINKED ISSUE #4095: save_all documentation fails to specify supported file types
Pillow's [save_all documentation](https://pillow.readthedocs.io/en/5.1.x/handbook/image-file-formats.html#saving) indicates that "If present and true, all frames of the image will be saved. If not, then only the first frame of a multiframe image will be saved."
However, the [release notes](https://pillow.readthedocs.io/en/3.1.x/releasenotes/3.0.0.html) for the same specify that this flag only applies to pdfs and gifs. 
Please specify supported file types to avoid misuse on multi-frame files of other types (e.g. multiframe jpeg, jpg, or png)

radarhere:
Resolves #4095

That issue saw a mistaken understanding that https://pillow.readthedocs.io/en/latest/handbook/image-file-formats.html#saving applied to all image formats, not just GIF. This PR tries to clarify that, and also for https://pillow.readthedocs.io/en/5.1.x/handbook/image-file-formats.html#saving-sequences

### DIFF
--- a/docs/handbook/image-file-formats.rst
+++ b/docs/handbook/image-file-formats.rst
@@ -113,8 +113,8 @@ to seek to the next frame (``im.seek(im.tell() + 1)``).
 Saving
 ~~~~~~
 
-When calling :py:meth:`~PIL.Image.Image.save`, the following options
-are available::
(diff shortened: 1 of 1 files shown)


hugovk:
Thanks!

PR #4119, opened 2019-10-06
Remove duplicate cleanup in test_decompression_bomb.py

jdufresne:
The same cleanup is done in the teardDown() method. There is no need to do it a 2nd time.


hugovk:
Thanks!

PR #4121, opened 2019-10-07
Use bytes literals instead of bytes(str)

jdufresne:
Bytes literals are available on all supported Python versions. Rather
than convert strings literals to bytes at runtime, simply use a bytes
literal.


hugovk:
Thanks.

PR #4122, opened 2019-10-07
Move several imports to the top-level of the file

jdufresne:
This better follows PEP 8 style guide:

https://www.python.org/dev/peps/pep-0008/#imports

> Imports are always put at the top of the file, just after any module
> comments and docstrings, and before module globals and constants.

This also avoids duplicate import code within the same file.


hugovk:
Thank you.

PR #4109, opened 2019-10-01
Drop support for EOL Python 2.7

hugovk:
For https://github.com/python-pillow/Pillow/issues/3642

Changes proposed in this pull request:

 * ~Stop testing Python 2.7~ -> https://github.com/python-pillow/Pillow/pull/4133
 * Remove compatibility code
 * Upgrade Python syntax with [pyupgrade](https://github.com/asottile/pyupgrade): ``pyupgrade `find . -name "*.py" -type f` --py3-plus``

PR #4136, opened 2019-10-12
Remove EOL (or nearly EOL) distro scripts from depends/

jdufresne:
The distro scripts in the depends directory are unmaintained. Som
reference long EOL distributions. For example, Fedora 23 went EOL
2016-12-20. These scripts aren't referenced in the documentation either.

With the planned removal of Python2 support, most of these scripts will
be outdated and incompatible as well.

This removes the last use of install_openjpeg.sh, so that has been
removed as well.


jdufresne:
Closing in favor of #4109.

PR #4139, opened 2019-10-12
Simpilify PillowTestCase.delete_tempfile for pytest

jdufresne:
As the test suite always runs with pytest now, self.currentResult is
always None. Using this, can remove unused code.


hugovk:
Thank you!

PR #4145, opened 2019-10-16
Fix Unicode support for PyPy

nulano:
Depends on #4109.

Reinstates #3780 changes reverted by #3935 for PyPy 7.2.0+.

This change also fixes Raqm support on PyPy (from https://github.com/python-pillow/Pillow/pull/4084#issuecomment-536663835).

There is a bug in PyPy that causes a failure of the test for #3777 when using Raqm, reported here: https://bitbucket.org/pypy/pypy/issues/3095/pyunicode_asucs4copy-fails-with-extended

### DIFF
--- a/src/_imagingft.c
+++ b/src/_imagingft.c
@@ -326,24 +326,12 @@ getfont(PyObject* self_, PyObject* args, PyObject* kw)
 static int
 font_getchar(PyObject* string, int index, FT_ULong* char_out)
 {
-#if (defined(PYPY_VERSION_NUM))
-    if (PyUnicode_Check(string)) {
-        Py_UNICODE* p = PyUnicode_AS_UNICODE(string);
-        int size = PyUnicode_GET_SIZE(string);
-        if (index >= size)
-            return 0;
-        *char_out = p[index];
-        return 1;
-    }
-#else
     if (PyUnicode_Check(string)) {
         if (index >= PyUnicode_GET_LENGTH(string))
             return 0;
         *char_out = PyUnicode_READ_CHAR(string, index);
         return 1;
     }
-#endif
-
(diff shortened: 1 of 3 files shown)


src/_imagingft.c
```diff
@@ -375,7 +351,7 @@ text_layout_raqm(PyObject* string, FontObject* self, const char* dir, PyObject *
         goto failed;
     }
 
-#if (PY_VERSION_HEX < 0x03030000) || (defined(PYPY_VERSION_NUM))
+#if (defined(PYPY_VERSION_NUM) && (PYPY_VERSION_NUM < 0x07020000))
```
nulano: I added a compile time PyPy version check here, to be removed in the future. What is Pillow's policy on supporting old versions of PyPy?

Aside: This fallback code fails on Windows anyway (due to the cast in the `p_raqm.set_text` call), see https://github.com/python-pillow/Pillow/pull/4084#issuecomment-536663835.

hugovk: For CPython we support the same versions as they do.

For PyPy, we generally only test the latest version, so maybe that too? If it's trivial to support older ones, we could do so, but I'm not sure of the value in testing many old versions. 

Does the PyPy project have a policy on which PyPy versions they support?

(PyPy plan to support Python 2.7 "forever" http://doc.pypy.org/en/latest/faq.html#how-long-will-pypy-support-python2, but we won't support PyPy Python 2.7 after 2020-01-01.)

hugovk:
https://github.com/python-pillow/Pillow/pull/4109 is merged, this is no longer blocked.

nulano:
Rebased.

hugovk:
Thanks!

PR #4147, opened 2019-10-17
Use default DPI when exif provides invalid x_resolution

### LINKED ISSUE #4146: Specific image can be opened by browser, but not by Pillow 6.2
### What did you do?
Tried to open a .jpg image with PIL

The photo can be opened in a browser and on Preview in Mac

### What actually happened?
```python
from PIL import Image
Image.open("/Users/dir/Desktop/3-full.jpg")
```
```Traceback (most recent call last):
  File "pillow_test.py", line 2, in <module>
    Image.open("/User/dir/Desktop/3-full.jpg")
  File "/User/dir/.pyenv/versions/3.6.6/lib/python3.6/site-packages/PIL/Image.py", line 2804, in open
    im = _open_core(fp, filename, prefix)
  File "/User/dir/.pyenv/versions/3.6.6/lib/python3.6/site-packages/PIL/Image.py", line 2790, in _open_core
    im = factory(fp, filename)
  File "/User/dir/.pyenv/versions/3.6.6/lib/python3.6/site-packages/PIL/JpegImagePlugin.py", line 789, in jpeg_factory
    im = JpegImageFile(fp, filename)
  File "/User/dir/.pyenv/versions/3.6.6/lib/python3.6/site-packages/PIL/ImageFile.py", line 106, in __init__
    self._open()
  File "/User/dir/.pyenv/versions/3.6.6/lib/python3.6/site-packages/PIL/JpegImagePlugin.py", line 376, in _open
    handler(self, i)
  File "/User/dir/.pyenv/versions/3.6.6/lib/python3.6/site-packages/PIL/JpegImagePlugin.py", line 165, in APP
    dpi = float(x_resolution[0]) / x_resolution[1]
ValueError: could not convert string to float: '\x00'
```

* OS: mac OS
* Python: 3.6.9
* Pillow: 6.2

For this particular image, I found out that the `x_resolution` is a byte string instead of an actual number. Therefore, we get the above error "ValueError: could not convert string to float: '\x00'"

```
(Pdb) x_resolution
'\x00\x04\x95\x0c\x00\x00\x00\x01'
```

https://github.com/python-pillow/Pillow/pull/2632 mentioned that getting dpi shouldn't be broken if the exif is invalid. Therefore, I propose the change of adding `ValueError` to the block:
```python
except (KeyError, SyntaxError, ZeroDivisionError):
            # SyntaxError for invalid/unreadable EXIF
            # KeyError for dpi not included
            # ZeroDivisionError for invalid dpi rational value
            self.info["dpi"] = 72, 72
```

beipang2:
Fixes #4146 Specific image can be opened by browser, but not by Pillow 6.2.

 * Each ValueError in the logic of retrieving DPI from EXIF
 * Added test


src/PIL/JpegImagePlugin.py
```diff
@@ -172,10 +172,11 @@ def APP(self, marker):
                 # 1 dpcm = 2.54 dpi
                 dpi *= 2.54
             self.info["dpi"] = int(dpi + 0.5), int(dpi + 0.5)
-        except (KeyError, SyntaxError, ZeroDivisionError):
+        except (KeyError, TypeError, SyntaxError, ValueError, ZeroDivisionError):
```
radarhere: Why did you add a TypeError? The test passes without it.

Tests/test_file_jpeg.py
```diff
@@ -634,6 +634,14 @@ def test_invalid_exif(self):
         # OSError for unidentified image.
         self.assertEqual(im.info.get("dpi"), (72, 72))
 
+    def test_invalid_exif_x_resolution(self):
+        # When no x or y resolution defined in EXIF
+        im = Image.open("Tests/images/invalid-exif-without-x-resolution.jpg")
+
+        # This should return the default, and not a ValueError or
+        # OSError for unidentified image.
```
radarhere:
```suggestion
        # OSError for an unidentified image.
```

Tests/test_file_jpeg.py
```diff
@@ -634,6 +634,14 @@ def test_invalid_exif(self):
         # OSError for unidentified image.
         self.assertEqual(im.info.get("dpi"), (72, 72))
 
+    def test_invalid_exif_x_resolution(self):
+        # When no x or y resolution defined in EXIF
```
radarhere:
```suggestion
        # When no x or y resolution is defined in EXIF
```

beipang2: Fixed

radarhere:
Thanks for reporting a problem, and then solving it!

PR #4174, opened 2019-10-28
Fix mismatched function name, add explanation

mixmastamyk:
Mention why this information is not available in the EXIF tag specified for this purpose.

 * Fix mismatched function name
 * Add explanation why current subsampling value is not set to the EXIF tag for this purpose.


hugovk:
Thanks!

PR #4182, opened 2019-10-29
Added UnidentifiedImageError

### LINKED ISSUE #3892: Add a specific exception if image not recognized?
Hello,

if the file provided as parameter to `Image.open()` is not recognized, an `IOError` is raised. It's nice but the exception is not really precise and, in python3.3+, it equals to `OSError` so it can be surprising depending on the used traceback.

The raised exception could be specific like `UnknownImageError`. It should inherit from `IOError` for backward compatibility.

The main modification would be in Image.py, line 2730:
```python
    for message in accept_warnings:
        warnings.warn(message)
-    raise IOError("cannot identify image file %r"
+    raise UnknownImageError("cannot identify image file %r"
                 % (filename if filename else fp))
```

What do you think about it? I will send a PR if you agree with the idea.

radarhere:
Resolves #3892

See how this goes. Pillow 7.0.0 seems like a good time for such changes.

As suggested there, this is a custom error, that inherits from the previous type for backwards compatibility.

See also the related issues #1643 and #2339

### DIFF
--- a/docs/releasenotes/7.0.0.rst
+++ b/docs/releasenotes/7.0.0.rst
@@ -66,8 +66,8 @@ API Additions
 Custom unidentified image error
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Pillow will now throw a custom UnidentifiedImageError when an image cannot be
-identified. For backwards compatibility, this will inherit from IOError.
+Pillow will now throw a custom ``UnidentifiedImageError`` when an image cannot be
+identified. For backwards compatibility, this will inherit from ``IOError``.
 
 Other Changes
 =============
--- a/docs/releasenotes/7.0.0.rst
+++ b/docs/releasenotes/7.0.0.rst
@@ -63,11 +63,11 @@ TODO
 API Additions
 =============
 
-TODO
-^^^^
+Custom unidentified image error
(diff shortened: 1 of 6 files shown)


hugovk:
This'll need release notes. 

Shall we add a simple regression test that the new one is an instance of the old one?

docs/releasenotes/7.0.0.rst
```diff
@@ -63,11 +63,11 @@ TODO
...
 =============
 
-TODO
-^^^^
-
-TODO
+Custom unidentified image error
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+Pillow will now throw a custom UnidentifiedImageError when an image cannot be
```
hugovk:
```suggestion
Pillow will now throw a custom ``UnidentifiedImageError`` when an image cannot be
```

docs/releasenotes/7.0.0.rst
```diff
@@ -63,11 +63,11 @@ TODO
...
+identified. For backwards compatibility, this will inherit from IOError.
```
hugovk:
```suggestion
identified. For backwards compatibility, this will inherit from ``IOError``.
```

radarhere: Thanks

PR #4160, opened 2019-10-21
Fix segfault: do not destroy glyph while its bitmap is used

cgohlke:
Fix segfault in `_imagingft.font_render` during `TestImageDraw::test_stroke`.

Fixes #4156 .

Changes proposed in this pull request:

Do not destroy glyph while its bitmap is used.

Issue #4152, opened 2019-10-20
Release Pillow 6.2.1 on October 21, 2019


hugovk:
## Purpose

Same as 6.2.0 and adds support for the newly-released Python 3.8 (https://github.com/python-pillow/Pillow/issues/4110).

## Point Release

Released as needed for security, installation or critical bug fixes.

* [x] Make necessary changes in `master` branch.
* [x] Update `CHANGES.rst`.
* [x] Check out release branch e.g.:
  ```bash
  git checkout -t remotes/origin/5.2.x
  ```
* [x] Cherry pick individual commits from `master` branch to release branch e.g. `5.2.x`.
* [x] Check [Travis CI](https://travis-ci.org/python-pillow/Pillow) to confirm passing tests in release branch e.g. `5.2.x`.
* [x] In compliance with [PEP 440](https://www.python.org/dev/peps/pep-0440/), update version identifier in `src/PIL/_version.py`
* [x] Run pre-release check via `make release-test`.
* [x] Create tag for release e.g.:
  ```bash
  git tag 5.2.1
  git push --tags
  ```
* [x] Create source distributions e.g.:
  ```bash
  make sdist
  ```

* [x] Create [binary distributions](https://github.com/python-pillow/Pillow/blob/master/RELEASING.md#binary-distributions)

### Windows
* [x] Contact `@cgohlke` for Windows binaries via release ticket e.g. https://github.com/python-pillow/Pillow/issues/1174.
* [x] Download and extract tarball from `@cgohlke` and `twine upload *`.

### Mac and Linux
* [x] Use the [Pillow Wheel Builder](https://github.com/python-pillow/pillow-wheels):
  ```bash
  git clone https://github.com/python-pillow/pillow-wheels
  cd pillow-wheels
  ./update-pillow-tag.sh [[release tag]]
  ```
* [x] Download distributions from the [Pillow Wheel Builder container](http://a365fff413fe338398b6-1c8a9b3114517dc5fe17b7c3f8c63a43.r19.cf2.rackcdn.com/).
  ```bash
  wget -m -A 'Pillow-<VERSION>-*' \
  http://a365fff413fe338398b6-1c8a9b3114517dc5fe17b7c3f8c63a43.r19.cf2.rackcdn.com
  ```

---

* [x] Upload all binaries and source distributions e.g. `twine upload dist/Pillow-5.2.0*`

* [x] Create a [new release on GitHub](https://github.com/python-pillow/Pillow/releases/new)

[...]

PR #4186, opened 2019-11-01
Enabled page heap verification

radarhere:
Suggested in https://github.com/python-pillow/Pillow/issues/3618#issuecomment-478872426

Helps prevent https://github.com/python-pillow/Pillow/issues/4152#issuecomment-544619189

Used https://docs.microsoft.com/en-us/windows-hardware/drivers/debugger/example-12---using-page-heap-verification-to-find-a-bug

As a demonstration, this will fail until #4160 is merged.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -85,6 +85,7 @@ build_script:
 test_script:
 - cd c:\pillow
 - '%PYTHON%\%PIP_DIR%\pip.exe install pytest pytest-cov'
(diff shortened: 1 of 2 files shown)


hugovk:
Very good! 

Do you know if this is possible with GitHub Actions?

radarhere:
Okay, added it to GHA as well.

nulano:
I was just working on this a week ago, but I wanted to add crash dumps, and got sidetracked with a ton of other work.

My WIP branch is here, if you would like to finish the dump part (it was exiting with the wrong exit code): https://github.com/nulano/Pillow/commits/heap

hugovk:
Thanks both, let's merge this now and further improvements always welcome in the future!

PR #4184, opened 2019-11-01
Ignore UserWarning in TestFileTiff::test_string_dimension

hugovk:
We should pay attention to and fix warnings in tests. Sometimes we will get deprecation warnings that we need to fix in our code, they shouldn't be lost in a haystack.

There's 5 warnings in the pytest summary, 4 of which are a `UserWarning` in a single test, like:

```
UserWarning: Possibly corrupt EXIF data.  Expecting to read 8587444226 bytes but only got 481. Skipping tag 63749
```

Let's either fix the test file, or ignore this specific warning, it's not relevant to this particular test.

# Before

```
/home/travis/virtualenv/python3.8.0/lib/python3.8/site-packages/Pillow-7.0.0.dev0-py3.8-linux-x86_64.egg/PIL/_imaging.py:3
  /home/travis/virtualenv/python3.8.0/lib/python3.8/site-packages/Pillow-7.0.0.dev0-py3.8-linux-x86_64.egg/PIL/_imaging.py:3: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
Tests/test_file_tiff.py::TestFileTiff::test_string_dimension
  /home/travis/virtualenv/python3.8.0/lib/python3.8/site-packages/Pillow-7.0.0.dev0-py3.8-linux-x86_64.egg/PIL/TiffImagePlugin.py:781: UserWarning: Possibly corrupt EXIF data.  Expecting to read 8587444226 bytes but only got 481. Skipping tag 63749
Tests/test_file_tiff.py::TestFileTiff::test_string_dimension
  /home/travis/virtualenv/python3.8.0/lib/python3.8/site-packages/Pillow-7.0.0.dev0-py3.8-linux-x86_64.egg/PIL/TiffImagePlugin.py:781: UserWarning: Possibly corrupt EXIF data.  Expecting to read 8589934082 bytes but only got 0. Skipping tag 278
Tests/test_file_tiff.py::TestFileTiff::test_string_dimension
  /home/travis/virtualenv/python3.8.0/lib/python3.8/site-packages/Pillow-7.0.0.dev0-py3.8-linux-x86_64.egg/PIL/TiffImagePlugin.py:781: UserWarning: Possibly corrupt EXIF data.  Expecting to read 83958528 bytes but only got 227. Skipping tag 34304
Tests/test_file_tiff.py::TestFileTiff::test_string_dimension
[...]

PR #4190, opened 2019-11-02
Ignore UserWarnings in TestFileTiff::test_string_dimension

radarhere:
For consideration, here's an alternative way to accomplish #4184

See https://docs.pytest.org/en/latest/warnings.html#pytest-mark-filterwarnings for documentation.


hugovk:
Much cleaner, thanks.

PR #4192, opened 2019-11-03
Correct str/bytes mixup in ContainerIO

jdufresne:
Image data is expected to be read as in bytes mode, not text mode so
ContainerIO should return bytes in all methods. The passed in file
handler is expected to be opened in bytes mode (as TarIO already does).

### DIFF
--- a/docs/handbook/tutorial.rst
+++ b/docs/handbook/tutorial.rst
@@ -468,17 +468,17 @@ Reading from an open file
     with open("hopper.ppm", "rb") as fp:
         im = Image.open(fp)
 
-To read an image from string data, use the :py:class:`~StringIO.StringIO`
-class:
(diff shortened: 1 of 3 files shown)


docs/handbook/tutorial.rst
```diff
@@ -471,14 +471,15 @@ Reading from an open file
 To read an image from string data, use the :py:class:`~StringIO.StringIO`
```
nulano: Should this also be changed?

jdufresne: :+1: thanks. Fixed in the latest revision.

radarhere:
I've created PR #4297 as an alternative - returning bytes if the file object is a bytestring object, and strings if it is not.

jdufresne:
Closing in favor of #4297. I think it makes more sense. Thanks!

PR #4194, opened 2019-11-04
GHA: Cache pip for Ubuntu, macOS and Windows

hugovk:

 * GitHub Actions now has caching
    * https://github.com/actions/cache

 * For the Ubuntu and macOS workflow, cache the pip directory (like we do on Travis CI) 

 * Unlike Travis config (`cache: pip`), we need to explicitly specify the directory to cache. The pip cache dir is different on different operating systems, so we need a conditional step for each
   * https://pip.pypa.io/en/latest/reference/pip_install/#caching

 * Also unlike Travis, we need to specify a unique key for the cache. Here it includes the workflow name, matrix OS and Python version.


hugovk:
Docs are now published:

* https://help.github.com/en/github/automating-your-workflow-with-github-actions/caching-dependencies-to-speed-up-workflows 

One thing:

> Once you create a cache, you cannot change the contents of an existing cache but you can create a new cache with a new key.

So we should update the key to include a hashFiles (see docs examples) of the files that list the pip installed stuff, so it will always fetch the same file, even if we change what we pip install. Those are listed in a few places, so we could just hash the bash files inside .travis and maybe the GHA workflow files too.

And also include some restore-keys so it can still find some useful cache when those hashed files change (see docs). A new cache will be uploaded for next time. 

Anyway, cache files are deleted if not accessed for 7 days, that's not very long.

hugovk:
Updated, ready for review.

hugovk:
Updated to include pip caching for Windows.

PR #4205, opened 2019-11-11
Include Pillow version in Windows wheel warning

hugovk:
For issues like https://github.com/python-pillow/Pillow/issues/4204, would it help to have the Pillow version number in the warning message?

Currently:

```
Pillow\setup.py:29: RuntimeWarning: Pillow does not yet support Python 3.8 and does not yet provide prebuilt Windows binaries. We do not recommend building from source on Windows.
```
xx

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ ZLIB_ROOT = None
 
 if sys.platform == "win32" and sys.version_info >= (3, 9):
     warnings.warn(
-        "Pillow {} does not yet support Python {}.{} and does not yet provide prebuilt"
+        "Pillow {} does not support Python {}.{} and does not provide prebuilt"
         "Windows binaries. We do not recommend building from source on Windows.".format(
             PILLOW_VERSION, sys.version_info.major, sys.version_info.minor
         ),
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ ZLIB_ROOT = None
 
 if sys.platform == "win32" and sys.version_info >= (3, 9):
     warnings.warn(
-        "Pillow {} does not support Python {}.{} and does not provide prebuilt"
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -24,11 +24,31 @@
...
+TIFF_ROOT = None
+ZLIB_ROOT = None
+
+
 if sys.platform == "win32" and sys.version_info >= (3, 9):
     warnings.warn(
-        "Pillow does not yet support Python {}.{} and does not yet provide "
-        "prebuilt Windows binaries. We do not recommend building from "
-        "source on Windows.".format(sys.version_info.major, sys.version_info.minor),
+        "Pillow {} does not yet support Python {}.{} and does not yet provide prebuilt"
```
```suggestion
        "Pillow {} does not support Python {}.{} and does not provide prebuilt"
```
If we're talking about a specific Pillow version, then I don't think 'yet' is true anymore - Pillow 6.2.0 will never support Python 3.8, for example.

hugovk: Good point, thanks!

setup.py
```diff
@@ -24,11 +24,31 @@
...
+TIFF_ROOT = None
+ZLIB_ROOT = None
+
+
 if sys.platform == "win32" and sys.version_info >= (3, 9):
     warnings.warn(
-        "Pillow does not yet support Python {}.{} and does not yet provide "
-        "prebuilt Windows binaries. We do not recommend building from "
-        "source on Windows.".format(sys.version_info.major, sys.version_info.minor),
+        "Pillow {} does not support Python {}.{} and does not provide prebuilt"
```
```suggestion
        "Pillow {} does not support Python {}.{} and does not provide prebuilt "
```
Without a space somewhere, 'prebuilt' and 'Windows' are combined into 'prebuiltWindows'

PR #4206, opened 2019-11-11
Enable copr builds and add packit config

dhodovsk:
Let us introduce [packit service](https://packit.dev) to you - the automation to integrate upstream open source projects into Fedora operating system.

After merging this PR, you are just a few steps away from RPM builds being automatically triggered on your PR's.
It means, that you'll be able to try and play with your change, packaged as an RPM.

But there is more. By using packit, you can for example enable adding new releases into Fedora Rawhide.

What are the next steps?
* Install [Packit-as-a-Service github application](https://github.com/marketplace/packit-as-a-service) in your repo
* In case you are first user, wait for account approval
* Enjoy the built RPMs!

For more info, please:
 * check out the documentation: https://packit.dev/docs/
 * contact @packit-service

The spec you see in this PR was fetched from your package's Fedora dist-git.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -1,0 +1,14 @@
+downstream_package_name: python-pillow
+jobs:
+- job: copr_build
+  metadata:
+    targets:
+    - fedora-30-x86_64
+    - fedora-31-x86_64
+    - fedora-rawhide-x86_64
+  trigger: pull_request
+specfile_path: python-pillow.spec
+synced_files:
+- python-pillow.spec
+- .packit.yaml
+upstream_package_name: Pillow
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -8,6 +8,7 @@ include *.sh
 include *.txt
 include LICENSE
 include Makefile
+include python-pillow.spec
 include tox.ini
 graft Tests
 graft src
@@ -20,6 +21,7 @@ exclude .appveyor.yml
 exclude .coveragerc
 exclude .codecov.yml
 exclude .editorconfig
+exclude .packit.yaml
 exclude .readthedocs.yml
 exclude azure-pipelines.yml
 global-exclude .git*
--- a/python-pillow.spec
+++ b/python-pillow.spec
@@ -1,0 +1,635 @@
+%global py2_incdir %(python2 -c 'import distutils.sysconfig; print(distutils.sysconfig.get_python_inc())')
+%global py2_libbuilddir %(python2 -c 'import sys; import sysconfig; print("lib.{p}-{v[0]}.{v[1]}".format(p=sysconfig.get_platform(), v=sys.version_info))')
+
+%if 0%{?fedora} || 0%{?rhel} > 7
+%global with_python3 1
+%global py3_incdir %(python3 -c 'import distutils.sysconfig; print(distutils.sysconfig.get_python_inc())')
+%global py3_libbuilddir %(python3 -c 'import sys; import sysconfig; print("lib.{p}-{v[0]}.{v[1]}".format(p=sysconfig.get_platform(), v=sys.version_info))')
+%endif
+
+%global srcname pillow
+# bootstrap building docs (pillow is required by docutils, docutils are
+#  required by sphinx; pillow build-requires sphinx)
+%global with_docs 1
+
+Name:           python-%{srcname}
+Version:        6.2.1
+Release:        1%{?dist}
+Summary:        Python image processing library
+
+# License: see http://www.pythonware.com/products/pil/license.htm
+License:        MIT
+URL:            http://python-pillow.github.io/
+Source0:        https://github.com/python-pillow/Pillow/archive/%{version}/Pillow-%{version}.tar.gz
+
+BuildRequires:  freetype-devel
+BuildRequires:  gcc
+BuildRequires:  ghostscript
+BuildRequires:  lcms2-devel
+BuildRequires:  libimagequant-devel
+BuildRequires:  libjpeg-devel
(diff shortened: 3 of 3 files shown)


hugovk:
Thanks for the PR! 

This would probably be fine from me. I doubt I'd personally use the RPM packages for testing, but if this makes things easier for downstream packaging (does it?) then that's good.

And it would be good to be in Rawhide to potentially get feedback on new things in master before our main releases and Fedora releases. 

Some questions.

What sort of maintenance would these files need in the future? Would we do that or someone else?

Please could you link to another repo that already uses this, so we can see how it'd look? 

The lint check is failing because check-manifest didn't find any rules for the new files. Should they be included in the sdist or left out? Please could you update the manifest file?

Please could you remove the Python 2 stuff? The next release will drop it. 

I don't have access to add apps to this repo. @wiredfool or @aclark4life, what do you think about this PR and could you add the app?

wiredfool:
* Half of the changes are a RH changelog of the RPM packages. I'm not clear that this repo is the appropriate place for that, nor is it likely to be maintained here going forward. 
* There's a hardcoded version in the spec file. We currently have a single version in the repo (src/PIL/_version.py). Inevitably, this will lead to version X.Y.Z being labeled with X-1.Q.W. For example, it's already out of date. 
* I'm not sure what constitutes a release where this would create an installable package that should be used by non-bleeding edge testers. We don't want to support anything except the most recent numerical releases. We wouldn't want people relying on random master snapshots. (Though, we can't prevent that with source installs, it's much less likely, as they need to at least assemble some dependencies and check out the code)

On the other hand, if this provides actual CI testing on non-x86 architectures (big-endian, ppc, ibm mainframes) then it's probably worth it. I'm not sure it's worth dropping our docker images for Fedora, as they do provide

hugovk:
Thanks for the PR! 

This would probably be fine from me. I doubt I'd personally use the RPM packages for testing, but if this makes things easier for downstream packaging (does it?) then that's good.

And it would be good to be in Rawhide to potentially get feedback on new things in master before our main releases. 

Some questions.

What sort of maintenance would these files need in the future? Would we do that or someone else?

Please could you link to another repo that already uses this, so we can see how it'd look? 

The lint check is failing because check-manifest didn't find any rules for the new files. Should they be included in the sdist or left out? Please could you update the manifest file?

Please could you remove the Python 2 stuff? The next release will drop it. 

I don't have access to add apps to this repo. @wiredfool or @aclark4life, what do you think about this PR and could you add the app? 

Thanks!

hugovk:
(Sorry for double posting, the first one didn't show up for some reason. Deleted!)

aclark4life:
Thanks @dhodovsk ! 

@hugovk @wiredfool Yeah let's leave this one open for a while and see if we get some overwhelming support and/or desire by users for Fedora RPMs … otherwise, not necessarily screaming "must have" to me for Pillow at this point.

TomasTomecek:
Dominika asked me to help her with answering your questions here. You asked good ones!

> And it would be good to be in Rawhide to potentially get feedback on new things in master before our main releases and Fedora releases.

That's exactly the intention here.

> What sort of maintenance would these files need in the future? Would we do that or someone else?

Our team (developers of the [packit project](https://packit.dev/)) would be glad to help.

For the RPM spec file - I understand it's pretty challenging to maintain that
file, so one of the solutions could be to work with the Fedora downstream
maintainer.

On the other hand, we can set this up in a way that the RPM spec file is
downloaded from Fedora git repositories all the time and not part of this PR.
This way you'd have the latest greatest (with all the ups and downs).

> Please could you link to another repo that already uses this, so we can see how it'd look?

Here's an example with RPM builds followed by testing in clean Fedora VMs:

https://github.com/ansible-community/ansible-bender/pull/190

> The lint check is failing because check-manifest didn't find any rules for the new files. Should they be included in the sdist or left out? Please could you update the manifest file?

They should be left out.

> Please could you remove the Python 2 stuff? The next release will drop it.

+1


>     * Half of the changes are a RH changelog of the RPM packages. I'm not clear that this repo is the appropriate place for that, nor is it likely to be maintained here going forward.

I'm assuming that the better variant would be then to download the RPM spec
file from Fedora and not store it here.

>     * There's a hardcoded version in the spec file. We currently have a single version in the repo (src/PIL/_version.py). Inevitably, this will lead to version X.Y.Z being labeled with X-1.Q.W. For example, it's already out of date.

When the spec file is used for building a PR, the version is replaced. Packit
gets version from git using `git describe`.

>     * I'm not sure what constitutes a release where this would create an installable package that should be used by non-bleeding edge testers. We don't want to support anything except the most recent numerical releases. We wouldn't want people relying on random master snapshots. (Though, we can't prevent that with source installs, it's much less likely, as they need to at least assemble some dependencies and check out the code)

You'll get an RPM for every PR with instructions how to install it - yep it's
really for people to test pull requests while they are being reviewed.

In packit, we don't support building the master branch for a project, yet.

> On the other hand, if this provides actual CI testing on non-x86 architectures (big-endian, ppc, ibm mainframes) then it's probably worth it. I'm not sure it's worth dropping our docker images for Fedora, as they do provide

These are the supported arches so far: fedora-rawhide-aarch64, fedora-rawhide-i386, fedora-rawhide-ppc64le, fedora-rawhide-x86_64

radarhere:
> Yeah let's leave this one open for a while and see if we get some overwhelming support and/or desire by users for Fedora RPMs … otherwise, not necessarily screaming "must have" to add to Pillow at this point.

Closing, as there hasn't been movement here in over six months.

PR #4208, opened 2019-11-15
Add Documentation and Github URLs for PyPi

andriyor:
Warehouse now uses the `project_urls` provided to display links in the sidebar, as well as including them in API responses to help automation tool find the source code for Pillow. For example, see Django's [setup.cfg](https://github.com/django/django/blob/master/setup.cfg) and [PyPI listing](https://pypi.org/project/Django/).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -859,7 +859,11 @@ try:
         license="HPND",
         author="Alex Clark (PIL Fork Author)",
         author_email="aclark@python-pillow.org",
-        url="http://python-pillow.org",
+        url="https://python-pillow.org/",
+        project_urls={
+            "Documentation": "https://pillow.readthedocs.io/en/stable/",
+            "Source Code": "https://github.com/python-pillow/Pillow",
+        },
         classifiers=[
             "Development Status :: 6 - Mature",
             "License :: OSI Approved :: Historical Permission Notice and Disclaimer (HPND)",  # noqa: E501
--- a/setup.py
+++ b/setup.py
@@ -859,10 +859,11 @@ try:
         license="HPND",
         author="Alex Clark (PIL Fork Author)",
         author_email="aclark@python-pillow.org",
-        url="https://python-pillow.org/",
+        url="https://python-pillow.org",
         project_urls={
             "Documentation": "https://pillow.readthedocs.io",
             "Source": "https://github.com/python-pillow/Pillow",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -859,7 +859,11 @@ def debug_build():
         license="HPND",
         author="Alex Clark (PIL Fork Author)",
         author_email="aclark@python-pillow.org",
-        url="http://python-pillow.org",
+        url="https://python-pillow.org/",
+        project_urls={
+            "Documentation": "https://pillow.readthedocs.io/en/stable/",
+            "Source Code": "https://github.com/python-pillow/Pillow",
```
hugovk: Let's use the [more common](https://github.com/hugovk/pypi-tools/issues/11) option:
```suggestion
            "Source": "https://github.com/python-pillow/Pillow",
```

setup.py
```diff
@@ -859,7 +859,11 @@ def debug_build():
         license="HPND",
         author="Alex Clark (PIL Fork Author)",
         author_email="aclark@python-pillow.org",
-        url="http://python-pillow.org",
+        url="https://python-pillow.org/",
+        project_urls={
+            "Documentation": "https://pillow.readthedocs.io/en/stable/",
```
hugovk: The README uses this more generic URL, let's use that in case we change the default in future:
```suggestion
            "Documentation": "https://pillow.readthedocs.io",
```

setup.py
```diff
@@ -859,7 +859,11 @@ def debug_build():
         license="HPND",
         author="Alex Clark (PIL Fork Author)",
         author_email="aclark@python-pillow.org",
-        url="http://python-pillow.org",
+        url="https://python-pillow.org/",
+        project_urls={
+            "Documentation": "https://pillow.readthedocs.io",
```
```suggestion
            "Documentation": "https://pillow.readthedocs.io",
            "Funding": "https://tidelift.com/subscription/pkg/pypi-pillow",
```
Like Django, we could also add in a funding link.

setup.py
```diff
@@ -859,7 +859,11 @@ def debug_build():
         license="HPND",
         author="Alex Clark (PIL Fork Author)",
         author_email="aclark@python-pillow.org",
-        url="http://python-pillow.org",
+        url="https://python-pillow.org/",
```
```suggestion
        url="https://python-pillow.org",
```
Nitpick: Would you be able to leave this as it is, unless there's a reason you've changed it?

PR #4209, opened 2019-11-16
Update the out-dated termux.sh

daquexian:
Termux does not have -dev packages any more(https://wiki.termux.com/wiki/No_more_-dev_packages) and pkg install -y instead of pkg -y install should be used

 * Use the right package name
 * Change pkg -y install (which doesn't work) to pkg install -y

### DIFF
--- a/depends/termux.sh
+++ b/depends/termux.sh
@@ -1,5 +1,5 @@
 #!/bin/sh
 
-pkg -y install python python-dev ndk-sysroot clang make \
-    libjpeg-turbo-dev 
+pkg install -y python ndk-sysroot clang make \
+    libjpeg-turbo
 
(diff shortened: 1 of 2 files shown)


hugovk:
Thanks! This file isn't used in the code or tests, or referred to in the docs.

However, the same command is mentioned in the docs at https://pillow.readthedocs.io/en/stable/installation.html#building-on-android

Shall we remove the file and update the docs instead?

wiredfool:
It’s useful to have shell scripts that can be run in addition to the documentation.

hugovk:
OK, let's update both!

radarhere:
I've created https://github.com/daquexian/Pillow/pull/1 to update the documentation.

daquexian:
@radarhere Thanks a lot!

hugovk:
Thank you!

PR #4216, opened 2019-11-20
Use Codecov's GitHub Action

hugovk:

 * Codecov GitHub Action is now available for all three operating systems
 * https://github.com/codecov/codecov-action
 * There was a problem including the Pillow token in `.codecov.yml`: some of the build jobs in my fork were sent to Pillow's Codecov, not mine
 * ~Plus fix so GHA macOS coverage is collected: needed to install lcov properly~ Done in https://github.com/python-pillow/Pillow/pull/4217

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -6,4 +6,6 @@ codecov:
   # https://docs.codecov.io/v4.3.6/docs/comparing-commits
   allow_coverage_offsets: true
 
+  token: 6dafc396-e7f5-4221-a38a-8b07a49fbdae
+
 comment: off
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -6,6 +6,4 @@ codecov:
   # https://docs.codecov.io/v4.3.6/docs/comparing-commits
   allow_coverage_offsets: true
 
-  token: 6dafc396-e7f5-4221-a38a-8b07a49fbdae
-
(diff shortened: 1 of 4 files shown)


hugovk:
Yeah, this doesn't work, for the same reason we added the token to `.codecov.yml` in the first place: GHA won't allow forks access to secrets, namely `secrets.CODECOV_TOKEN`.

So there's no GHA coverage here (yet, but all GHA jobs have finished):

* https://codecov.io/gh/python-pillow/Pillow/commit/1ff3a501ef31552e1ef279b8554655977f1c2e0e/build.

Specific issue at Codecov GitHub Action: 

* https://github.com/codecov/codecov-action/issues/29

General issue at GitHub Community:

* https://github.community/t5/GitHub-Actions/Make-secrets-available-to-builds-of-forks/m-p/30678/highlight/true#M508

Let's see what happens with the token re-added.

nulano:
How about putting the token in the `.yml` files (hopefully temporarily)? I think using the GitHub action is cleaner; coverage currently doesn't work on Windows, and the Linux builds are received as 'Untitled' anyway.

hugovk:
Yes, let's give it a go. Bullet point 3 will still be a problem, but the other bits are an improvement over the current situation.

PR #4201, opened 2019-11-07
GitHub Actions: Update to windows-2019

nulano:
Fixes #4188

Changes proposed in this pull request:

 * Run Windows tests on `windows-2019`
 * Use official prebuilt Ghostscript binary
 * Patch LCMS2 build scripts to work with VS2019 (1)
 * Use `DISTUTILS_USE_SDK=1` to help Python 3.5 and 3.6 find the compiler (1)
 * Fix PyPy3.6 pip warning causing early failures; FIXME PyPy can't find the compiler

(1) These two changes should probably be reverted once VS2017 is added to `windows-2019` as promised in https://github.com/actions/virtual-environments/issues/68#issuecomment-548534537

Issue #4188, opened 2019-11-01
GitHub Actions: Removing windows-2016 on 7 Nov


hugovk:
> The Windows Server 2016 virtual environment will be removed on November 7, 2019. You will need to update your workflow files to run on `windows-latest` which will run on the Windows Server 2019 environment instead.

https://github.com/actions/virtual-environments/issues/68

We use `windows-2016`:

https://github.com/python-pillow/Pillow/blob/64e1feece9fb56d41171d55a67f1d5969aef1dda/.github/workflows/test-windows.yml#L8

hugovk:
Simply flipping to `windows-latest` doesn't work:
. https://github.com/hugovk/Pillow/commit/31da3994b0dd0d74c788d64a436ceec4c7c5f13a/checks?check_suite_id=292424329

nulano:
I have a WIP rewrite of the Windows workflow in my repo, but I don't expect to have time to work on it this month. AFAICT from the log, it is having trouble finding `vcvarsall.bat`. There is another version in `C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\VC\Auxiliary\Build\vcvarsall.bat` which needs to be called with the `x86` parameter, and for Python 3.5 and PyPy3.6 with the `-vcvars_ver=14.0` (this ensures the same compiler distutils finds is also used for dependencies). The second parameter might not be needed on Windows Server 2019, as I think it only has the newer compiler.

You can see my build here: https://github.com/nulano/Pillow/runs/245421431#step:7:7

Another problem that might come up, is that I am not sure if the Windows 7.1A SDK is present in the newer VS version. Some of the dependencies require files which are removed in newer SDKs.

nulano:
I've updated the scripts: https://github.com/nulano/Pillow/runs/286210242

The following are no longer supported:
* LCMS2 hasn't been updated to support Visual Studio 2019.
* GhostScript requires v140 build tools, which are not installed on the `windows-2019`.
* distutils for Python 3.5 fails to find a compiler.
* distutils for Python 3.6 fails to find redistributable libraries and compiles with `/MT` instead of `/MD`. This can cause link and runtime errors.

(I believe PyPy3.6 uses distutils from Python 3.5, but it failed to install due to an intermittent network error.)

----

From https://github.com/actions/virtual-environments/issues/68#issuecomment-548534537:
> > Does that mean Visual Studio 2017 will be added to `windows-latest`?
> 
[...]

PR #4218, opened 2019-11-20
Use prebuilt GhostScript on AppVeyor

nulano:
#4201 changed to using official prebuilt GhostScript binaries on GitHub Actions (instead of compiling it), as it is hard to compile on `windows-2019`.

This PR does the same for AppVeyor to speed it up significantly.

Also see https://github.com/python-pillow/Pillow/issues/4188#issuecomment-551155313 for a (brief) discussion.


hugovk:
Thank you!

PR #4219, opened 2019-11-20
Fetch binary dependencies from pillow-depends on GitHub Actions

nulano:
Depends on https://github.com/python-pillow/pillow-depends/pull/31.

* Use NASM and GhostScript from `pillow-depends`
* Switch to using `actions/checkout` to fetch `pillow-depends`, instead of `ps curl`

These two changes should decrease the amount of intermittent network errors encountered in tests.

See build using https://github.com/python-pillow/pillow-depends/pull/31 here: https://github.com/nulano/Pillow/runs/312459799

### DIFF
--- a/.github/workflows/test-windows.yml
+++ b/.github/workflows/test-windows.yml
@@ -68,14 +68,11 @@ jobs:
 
     - name: Fetch dependencies
       run: |
-        curl -fsSL -o nasm.zip https://www.nasm.us/pub/nasm/releasebuilds/2.14.02/win64/nasm-2.14.02-win64.zip
-        7z x nasm.zip "-o$env:RUNNER_WORKSPACE\"
+        7z x ..\pillow-depends\nasm-2.14.02-win64.zip "-o$env:RUNNER_WORKSPACE\"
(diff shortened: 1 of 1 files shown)


hugovk:
I've merged https://github.com/python-pillow/pillow-depends/pull/31 and restarted GHA here.

.github/workflows/test-windows.yml
```diff
@@ -31,6 +31,11 @@ jobs:
     steps:
     - uses: actions/checkout@v1
 
+    - uses: actions/checkout@v1
+      with:
+        repository: python-pillow/pillow-depends
+        ref: master
```
hugovk:
```suggestion
        ref: master
        fetch-depth: 1
```

The repo isn't particularly deep, but we might be able to save some time (from the current ~16s) by only fetching the last commit. Docs:

* https://github.com/actions/checkout/blob/50fbc622fc4ef5163becd7fab6573eac35f8462e/action.yml#L17-L18

nulano: It seems to be slower, running for about ~18s: https://github.com/nulano/Pillow/runs/315950425

hugovk:
Thank you!

PR #4222, opened 2019-11-21
Import unittest from stdlib rather than helper.py

jdufresne:
The unittest in helper.py has not offered an interesting abstraction
since dbe9f85c7d8f760756ebf8129195470700e63e3d so import from the more
typical stdlib location.


hugovk:
Thanks!

PR #4231, opened 2019-11-24
Fix thumbnail geometry when DCT scaling is used

homm:
`Image.thumbnail()` method uses `.draft(None, size)` which could scale down an image with an integer ratio. The problem is if the original image size is not dividable by this ratio, the last line or row of the resulting image gets higher weight.

This could be fixed by using the `box` argument for resizing. For example, if the original image was 257×257 pixels and we want 30×30. The image will be scaled down 8 times to 33×33 and then resized to 30×30. But 257 / 8 = 32.125, not 33. Now using `box` we can extract an interesting region `(0, 0, 32.125, 32.125)`.

```python
im = Image.new('RGB', (257, 257), 'red')
im.paste(Image.new('RGB', (255, 255), 'white'), (1, 1))
 
thumb = fromstring(tostring(im, "JPEG", quality=95))
thumb.thumbnail((30, 30), Image.BICUBIC)
thumb.save('_out.thumb.png')
 
im.resize((30, 30), Image.BICUBIC).save('_out.ref.png')
```

Thumbnail, fixed version, reference without DCT scaling:

_out thumb _out thumb _out ref

Thumbnail, fixed version, reference without DCT scaling:

_out thumb _out fixed _out resize


src/PIL/JpegImagePlugin.py
```diff
@@ -409,7 +409,8 @@ def draft(self, mode, size):
             return
 
         d, e, o, a = self.tile[0]
-        scale = 0
+        scale = 1
```
homm: Both values are not handled: https://github.com/python-pillow/Pillow/blob/master/src/libImaging/JpegDecode.c#L238

While 1 is more correct.

src/PIL/JpegImagePlugin.py
```diff
@@ -432,7 +433,8 @@ def draft(self, mode, size):
         self.tile = [(d, e, o, a)]
         self.decoderconfig = (scale, 0)
 
-        return self
```
homm: This is breaking change. Previously, `draft()` could return `None` or `self`. This is very strange behavior, I don't think anyone actually used it. Now `draft()` returns the results of its work or `None`.

I'll add this in docs.

src/PIL/Image.py
```diff
@@ -1126,12 +1126,15 @@ def draft(self, mode, size):
         """
         Configures the image file loader so it returns a version of the
         image that as closely as possible matches the given mode and
-        size.  For example, you can use this method to convert a color
+        size. For example, you can use this method to convert a color
         JPEG to greyscale while loading it, or to extract a 128x192
         version from a PCD file.
 
+        If any changes are made, returns a tuple with the chosen `mode` and
```
```suggestion
        If any changes are made, returns a tuple with the chosen ``mode`` and
```

src/PIL/Image.py
```diff
@@ -1126,12 +1126,15 @@ def draft(self, mode, size):
...
+        `box` with coordinates of the original image within the altered one.
```
```suggestion
        ``box`` with coordinates of the original image within the altered one.
```

src/PIL/Image.py
homm: Thanks!

src/PIL/Image.py
```diff
@@ -2143,14 +2146,17 @@ def thumbnail(self, size, resample=BICUBIC):
...
             return
 
-        self.draft(None, size)
+        res = self.draft(None, size)
+        if res is not None:
+            box = res[1]
 
         if self.size != size:
-            im = self.resize(size, resample)
+            im = self.resize(size, resample, box=box)
```
```suggestion
            im = self.resize(size, resample, box)
```
I suggest not specifying the argument name - it's not necessary, and the variable name already clearly indicates the purpose.

homm: In truth, I think this is a case where we could use [Keyword-Only Arguments](https://www.python.org/dev/peps/pep-3102/).

It seems obvious that we can omit the argument name when you watch on it and so you know that it exists and what exactly it is doing. But once you omit it, you can't know exactly where your variable goes.

homm:
Rebased

docs/releasenotes/7.0.0.rst
```diff
@@ -58,6 +58,20 @@ and :py:meth:`~PIL.ImageOps.fit`` functions.
 See :ref:`concept-filters` to learn the difference. In short,
 ``Image.NEAREST`` is a very fast filter, but simple and low-quality.
 
+Image.draft() return value
```
homm: @radarhere Does this description look good for you?

radarhere: I've created https://github.com/uploadcare/pillow-simd/pull/60 with my suggestion.

PR #4232, opened 2019-11-25
Use context managers in tests

radarhere:
Resolves #2818 

The remaining ' = Image.open' lines in the test suite are either related to file pointer management or would involve restructuring the test code more significantly.

This is also good because it means that we're practicing what we preach - that opening an image should not be closed implicitly as of #3577

PR #4243, opened 2019-12-01
Add APNG support

### LINKED ISSUE #3483: Add apng support
apng is not what it used to. Now google and apple support it.

pmrowla:
Fixes #3483.

 * Implements support for reading APNG files

    * `seek()` `tell()` can be used for accessing frames, `seek()` rewinding is supported
     * If an APNG contains a default image that is not part of the actual animation sequence (i.e. `IDAT` chunks without an `fcTL` chunk), `im.info["default_image"]` will be set and `True`, and the default image can still be accessed as frame 0 (`seek(0)`).
      In this case, the first animation frame will be at `seek(1)`, and `n_frames` will contain the number of frames plus 1 to account for the default image.
    * If `im.info["default_image"]` is not present or is `False` it means the APNG did not contain a separate default image
    * `im.is_animated` will always be `True` for APNG files, even if the animation only contains a single frame, to differentiate single frame APNG files from standard PNG files.

* Implements partial support for writing APNG files

    * APNG animations can be written using `save_all` and `append_images` (same behavior as writing GIF sequences)
    * Loop and frame duration can also be specified (same behavior as GIF sequences)
    * APNG frame disposal method and frame blend method can be specified via `disposal` and `blend` parameters
   * All images from the `append_frames` sequence are expected to be the same size as the initial image, frame bounding box is calculated based on the differences between a given frame and the prior frame
    * Frame bounding boxes cannot be manually specified with this implemention
* Tests are based on the browser compatibility tests from https://philip.html5.org/tests/apng/tests.html, which are the only tests currently listed at: https://wiki.mozilla.org/APNG_Specification


pmrowla:
I've been doing some more testing and found a problem in the code for writing APNG sequences that occurs if it has to write multiple fdAT chunks in a row for one image frame. I'll try to get a fix and new tests pushed for that at some point in the next couple of days.

docs/handbook/image-file-formats.rst
```diff
@@ -551,6 +551,119 @@ The :py:meth:`~PIL.Image.Image.save` method supports the following options:
...
+where applicable:
+
+**default_image**
+    Specifies whether or not this APNG file contains a separate default image,
+    which is not a part of the actual APNG animation.
+
+    When an APNG file contains a default image, the initially loaded image (i.e.
+    the result of ``seek(0)``) will be the default image.
+    To account for the presence of the default image, the
+    py:attr:`~PIL.Image.Image.n_frames` property will be set to ``frame_count + 1``,
```
radarhere:
```suggestion
    :py:attr:`~PIL.Image.Image.n_frames` property will be set to ``frame_count + 1``,
```

Tests/test_file_png.py
```diff
@@ -620,16 +613,536 @@ def test_exif_argument(self):
...
-    @unittest.skipUnless(
-        HAVE_WEBP and _webp.HAVE_WEBPANIM, "WebP support not installed with animation"
-    )
-    def test_apng(self):
-        with Image.open("Tests/images/iss634.apng") as im:
+    # APNG browser support tests and fixtures via:
+    # https://philip.html5.org/tests/apng/tests.html
+    # (referenced from https://wiki.mozilla.org/APNG_Specification)
+    def test_apng_basic(self):
+        im = Image.open("Tests/images/apng/single_frame.png")
```
hugovk: Use context manager for these? cf. https://github.com/python-pillow/Pillow/pull/4232

docs/handbook/image-file-formats.rst
```diff
@@ -551,6 +551,119 @@ The :py:meth:`~PIL.Image.Image.save` method supports the following options:
     library before building the Python Imaging Library. See the `installation
     documentation <../installation.html>`_ for details.
 
+APNG sequences
+~~~~~~~~~~~~~~~~~
```
hugovk:
```suggestion
~~~~~~~~~~~~~~
```

docs/handbook/image-file-formats.rst
```diff
@@ -551,6 +551,119 @@ The :py:meth:`~PIL.Image.Image.save` method supports the following options:
...
+
+The PNG loader includes limited support for reading and writing APNG files.
```
hugovk:
```suggestion
The PNG loader includes limited support for reading and writing Animated Portable Network Graphics (APNG) files.
```

docs/handbook/image-file-formats.rst
```diff
@@ -551,6 +551,119 @@ The :py:meth:`~PIL.Image.Image.save` method supports the following options:
...
+When an APNG file is loaded, :py:meth:`~PIL.ImageFile.ImageFile.get_format_mimetype`
+will return ``"image/apng"``, and the :py:attr:`~PIL.Image.Image.is_animated` property
+will be ``True`` (even for single frame APNG files).
+The :py:meth:`~PIL.Image.Image.seek` and :py:meth:`~PIL.Image.Image.tell` methods
+are supported.
+
+``im.seek()`` raises an :py:exc:`EOFError` if you try to seek after the last frame.
+
+The following :py:attr:`~PIL.Image.Image.info` properties will be set for APNG frames,
```
hugovk:
```suggestion
These :py:attr:`~PIL.Image.Image.info` properties will be set for APNG frames,
```

Tests/test_file_png.py
```diff
@@ -620,16 +613,536 @@ def test_exif_argument(self):
         with Image.open(test_file) as reloaded:
             self.assertEqual(reloaded.info["exif"], b"Exif\x00\x00exifstring")
 
-    @unittest.skipUnless(
-        HAVE_WEBP and _webp.HAVE_WEBPANIM, "WebP support not installed with animation"
-    )
-    def test_apng(self):
-        with Image.open("Tests/images/iss634.apng") as im:
+    # APNG browser support tests and fixtures via:
```
hugovk: These tests double the size of this file, let's move them to their own `test_file_apng.py` or `test_file_png_apng.py`.

docs/handbook/image-file-formats.rst
```diff
@@ -551,6 +551,119 @@ The :py:meth:`~PIL.Image.Image.save` method supports the following options:
     library before building the Python Imaging Library. See the `installation
     documentation <../installation.html>`_ for details.
 
+APNG sequences
+~~~~~~~~~~~~~~~~~
+
+The PNG loader includes limited support for reading and writing APNG files.
+When an APNG file is loaded, :py:meth:`~PIL.ImageFile.ImageFile.get_format_mimetype`
+will return ``"image/apng"``, and the :py:attr:`~PIL.Image.Image.is_animated` property
+will be ``True`` (even for single frame APNG files).
```
radarhere: I'm not currently in favour of `is_animated` returning true for only one frames. You've said that this is to 'differentiate single frame APNG files from standard PNG files'. However, you can already tell the two apart through the mimetype. Thoughts?

pmrowla: I wrote it that way because in my head I was associating `is_animated` with "is the `seek()` method available", and for single frame APNG files `seek(1)` is valid if there is a separate default image in addition to the first (and only) animation frame.

But I'm fine with changing it to make the behavior more consistent with other image formats.

pmrowla:
Also, I'll try to push the testing & docs changes hugovk requested next week, but I might not be able to get to them until after the holidays.

docs/handbook/image-file-formats.rst
pmrowla: Just to double check this before I make any changes, my understanding is that the desired behavior is for `is_animated` to be true if `Image.n_frames` is greater than 1.

Meaning:

* If the APNG contains one frame and no default image, `is_animated` should be false (since `seek(0)` would return the first frame, and `seek(1)` is invalid)
* If the APNG contains one frame as well as a default image, `is_animated` should be true (since `seek(0)` would return the default image, and `seek(1)` is valid and returns the first frame)
* If the APNG contains more than one frame, `is_animated` should be true (the presence of a default image is irrelevant for this case)

Tests/test_file_png.py
pmrowla: The requested documentation changes are fixed in df42b63

Tests/test_file_png.py
pmrowla: Fixed in e7f47d2

pmrowla:
PR was rebased onto the current master, since 7.0.0 was just released. The failing test looks like a macos-pypy config issue, and I think is unrelated to this PR?

hugovk:
Yes, the macOS pypy3 failure is unrelated, it's also happening on `master`.

radarhere:
The issue with the failing macOS test has now been resolved.

docs/handbook/image-file-formats.rst
pmrowla: Resolved in ff7aea0, `is_animated` is now true when `n_frames` is greater than 1, and false otherwise (`n_frames` still depends on both animation frame count as well as presence of a separate default image)

radarhere:
For some reason, #4454 has caused this to fail.

pmrowla:
I know the APNG code uses `getbbox()` when writing frame sequences to get the bounding box for the actual modified region between consecutive frames (the GIF plugin does the same thing as well).

I should be able to take a look into this over the weekend.

pmrowla:
The issue was that subtract_modulo is used to find the delta between two animation frames, but frame data was not being converted to RGB before doing the subtract. The resulting delta always contained an alpha channel w/all zeroes, so `getbbox` would return empty after the change in #4454.

hugovk:
We recently switched from unittest to pytest, I've updated the tests here in https://github.com/python-pillow/Pillow/pull/4243/commits/92c9961ceac9fdfb1b3118ace8c1818acbe4305f.

hugovk:
Also added an unrelated fix to be able to restart a network error on GHA: https://github.com/python-pillow/Pillow/pull/4243/commits/91289f863be0a1ac05565fd57d712b883308a255

hugovk:
Thank you!

PR #4244, opened 2019-12-02
Fix EXIF rational serialization issue due to mismatch between legacy and new mode

kasper0406:
Note that I do not have any prior experience with the Pillow codebase nor EXIF parsing. Please be thorough when reviewing.

The test image is a copy of the image with id `17e017d9f38ada01 ` from the [Open Images Dataset](https://storage.googleapis.com/openimages/web/index.html).

**Current behavior:**
The added `test_exif_transpose_metadata_error` test fails with a type error trying to call `abs` on a tuple.

**New behavior:**
Test does not crash, but gives incorrect result:
```python
    def test_exif_transpose_metadata_error(self):
        with Image.open("Tests/images/exif_transpose_rational_mismatch.jpg") as img:
            transposed = ImageOps.exif_transpose(img)
            parsed_exif = transposed.getexif()
>           self.assertEqual(parsed_exif[318], [(313, 1000), (329, 1000)])
E           AssertionError: ((538976288, 8224), (313, 1000)) != [(313, 1000), (329, 1000)]
```

**Expected behavior:**
Test should succeed with the expected result.
Looking at the de-serialized EXIF data, I see the following:
```[...] 318: ((538976288, 8224), (313, 1000)), 319: ((329, 1000), [...]```
Hence, it looks like the bytes have been pushed somehow.

I do not believe that the offset EXIF information has been caused by my change. Any input on how and where to look to fix this is appreciated.

### DIFF
--- a/Tests/test_imageops.py
+++ b/Tests/test_imageops.py
@@ -307,3 +307,10 @@ def test_exif_transpose():
                     "Tests/images/hopper_orientation_" + str(i) + ext
                 ) as orientation_im:
                     check(orientation_im)
+
+
+def test_exif_transpose_metadata_error():
+    with Image.open("Tests/images/exif_transpose_rational_mismatch.jpg") as img:
+        transposed = ImageOps.exif_transpose(img)
+        parsed_exif = transposed.getexif()
+        assert parsed_exif[318] == [(313, 1000), (329, 1000)]
--- a/src/PIL/TiffImagePlugin.py
+++ b/src/PIL/TiffImagePlugin.py
@@ -258,8 +258,9 @@ def _accept(prefix):
 
 
 def _limit_rational(val, max_val):
-    inv = abs(val) > 1
-    n_d = IFDRational(1 / val if inv else val).limit_rational(max_val)
+    ifd_val = IFDRational(val[0], val[1]) if isinstance(val, tuple) else val
+    inv = abs(ifd_val) > 1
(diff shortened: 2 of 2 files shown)


radarhere:
More of a note for myself than anything - inspecting, I think the change should be applied [before `_write_dispatch` to be more general](https://github.com/python-pillow/Pillow/blob/b73e3dddcc9e6b3039db366b202254f693ccd13d/src/PIL/TiffImagePlugin.py#L808), but otherwise, it's fine. The reason that this is failing is an ongoing issue with saving TIFF tags with a count greater than 1.

radarhere:
The test image used here has a relatively large filesize. Is there a smaller file that you have that reproduces the problem?

kasper0406:
Unfortunately I do not have a smaller file reproducing this issue. I hit the bugs a few times when processing images from the [Open Images dataset](https://storage.googleapis.com/openimages/web/index.html).

radarhere:
This code is, as you say, trying to deal with the tuples creating by the legacy mode. I have found that my PR #4637 is actually an alternate solution to this problem - it switches the `Exif` away from the legacy mode, so that instead of tuples, `IFDRational`s are used, which can be used with `abs`.

I mean, my PR still needs to be reviewed by another member of the core team, but I thought I'd let you know it's there.

radarhere:
Thanks for your initiative in creating this. My PR has now been merged, so this is solved. The only difference is that `parsed_exif[318]` now equals `(0.313, 0.329)`.

PR #4248, opened 2019-12-04
Add La mode packing and unpacking

homm:


homm:
Simple use case:

```python
>>> Image.new('La', (10, 10)) == Image.new('La', (10, 10))                                                                  
ValueError: No packer found from La to La
```

PR #4255, opened 2019-12-07
Change default resize resampling filter from NEAREST to BICUBIC

homm:
I believe this is much convenient in most cases, which makes Pillow much user-friendly.


homm:
@python-pillow/pillow-team 
Guys, this is a very bold but I think necessary change. `Image.NEAREST` is the most useless resampling filter, this default in truth makes the filter argument mandatory in most cases. Most of the users are annoyed by this and this is what we can change is such great release. Do you have any thoughts?

wiredfool:
Makes sense to me.

aclark4life:
@homm Go for it, pending coordination with @hugovk and @radarhere and maybe some extra "we're doing this" type announcements preceding the release (presumably Jan 1), particularly if it could break existing usage.

homm:
Actually, there are `ImageOps.pad()`, `ImageOps.scale()` and `ImageOps.fit()` functions which also uses `Image.NEAREST` as default. I think we should change it too.

Another question is `Image.Image.rotate()` and `Image.Image.transform()` and in this case I'm not sure.

hugovk:
Normally we would want to put deprecation warnings where possible, but that would not have had much effect here.

The next release is a major, breaking release, so improving the default is fine. We're including it in the release notes, which I think is enough here.

Should we mention in the release notes something along the lines that performance is slower but with improved downscaling and upscaling quality?

homm:
@hugovk Maybe the link to filters description would be enough? https://pillow.readthedocs.io/en/stable/handbook/concepts.html#concept-filters

hugovk:
Yep, looks good.

docs/releasenotes/7.0.0.rst
```diff
@@ -47,6 +47,17 @@ Setting the size of TIFF images
 Setting the size of a TIFF image directly (eg. ``im.size = (256, 256)``) throws
 an error. Use ``Image.resize`` instead.
 
+Default resampling filter
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The default resampling filter is changed to high-quality convolution
+``Image.BICUBIC`` instead of ``Image.NEAREST`` for the :py:meth:`~PIL.Image.Image.resize`
+method and for :py:meth:`~PIL.ImageOps.pad``, :py:meth:`~PIL.ImageOps.scale``
```
radarhere:
```suggestion
method and the :py:meth:`~PIL.ImageOps.pad``, :py:meth:`~PIL.ImageOps.scale``
```
Just a suggestion.

docs/releasenotes/7.0.0.rst
```diff
@@ -47,6 +47,17 @@ Setting the size of TIFF images
 Setting the size of a TIFF image directly (eg. ``im.size = (256, 256)``) throws
 an error. Use ``Image.resize`` instead.
 
+Default resampling filter
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The default resampling filter is changed to high-quality convolution
```
radarhere:
```suggestion
The default resampling filter has been changed to the high-quality convolution
```
Just a suggestion.

docs/releasenotes/7.0.0.rst
```diff
@@ -47,6 +47,17 @@ Setting the size of TIFF images
...
+``Image.BICUBIC`` instead of ``Image.NEAREST`` for the :py:meth:`~PIL.Image.Image.resize`
```
radarhere:
```suggestion
``Image.BICUBIC`` instead of ``Image.NEAREST``, for the :py:meth:`~PIL.Image.Image.resize`
```

docs/releasenotes/7.0.0.rst
```diff
@@ -47,6 +47,17 @@ Setting the size of TIFF images
...
+method and for :py:meth:`~PIL.ImageOps.pad``, :py:meth:`~PIL.ImageOps.scale``
+and :py:meth:`~PIL.ImageOps.fit`` functions.
+``Image.NEAREST`` is still always used for images in "P" and "1" modes.
+See :ref:`concept-filters` to learn the difference. In short,
+``Image.NEAREST`` is a very fast and simple but low-quality filter.
```
radarhere:
```suggestion
``Image.NEAREST`` is a very fast filter, but simple and low-quality.
```

src/PIL/Image.py
```diff
@@ -1771,8 +1771,9 @@ def resize(self, size, resample=NEAREST, box=None):
            one of :py:attr:`PIL.Image.NEAREST`, :py:attr:`PIL.Image.BOX`,
            :py:attr:`PIL.Image.BILINEAR`, :py:attr:`PIL.Image.HAMMING`,
            :py:attr:`PIL.Image.BICUBIC` or :py:attr:`PIL.Image.LANCZOS`.
-           If omitted, or if the image has mode "1" or "P", it is
-           set :py:attr:`PIL.Image.NEAREST`.
+           Default filter is :py:attr:`PIL.Image.BICUBIC`.
+           If the image has mode "1" or "P", it is
+           always set :py:attr:`PIL.Image.NEAREST`.
```
```suggestion
           always set to :py:attr:`PIL.Image.NEAREST`.
```
I realise you're just copying what was always there.

homm:
@radarhere done

radarhere:
I've created #5411 to revert this for special image modes, since they don't currently support any resampling other than NEAREST.

PR #4256, opened 2019-12-07
Better thumbnail aspect ratio preservation

homm:
If the source image is 256×162, it's aspect ratio is 1.58. Before, after thumbnailing, for example, the resulting image had 33×20 size, which aspect ratio is 1.65. Now thumbnail returns 33×21 image, with aspect ration 1.57, which is much closer.

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -2137,11 +2137,11 @@ class Image:
         # preserve aspect ratio
         x, y = self.size
         if x > size[0]:
-            y = int(max(y * size[0] / x, 1))
-            x = int(size[0])
+            y = max(round(y * size[0] / x), 1)
+            x = size[0]
         if y > size[1]:
-            x = int(max(x * size[1] / y, 1))
-            y = int(size[1])
+            x = max(round(x * size[1] / y), 1)
+            y = size[1]
         size = x, y
 
         if size == self.size:
--- a/Tests/test_image_thumbnail.py
+++ b/Tests/test_image_thumbnail.py
@@ -5,37 +5,43 @@ from .helper import PillowTestCase, hopper
 
 class TestImageThumbnail(PillowTestCase):
     def test_sanity(self):
-
         im = hopper()
-        im.thumbnail((100, 100))
+        self.assertIsNone(im.thumbnail((100, 100)))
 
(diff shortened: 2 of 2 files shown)


src/PIL/Image.py
```diff
@@ -2137,11 +2137,11 @@ def thumbnail(self, size, resample=BICUBIC):
         # preserve aspect ratio
         x, y = self.size
         if x > size[0]:
-            y = int(max(y * size[0] / x, 1))
-            x = int(size[0])
+            y = max(round(y * size[0] / x), 1)
+            x = size[0]
```
radarhere: Just to be aware - by removing the `int` rounding on these second lines, if someone had existing code `im.thumbnail((150.2, 150.2))`, it would now break.

That's not unreasonable though.

hugovk: Example:

```python
from PIL import Image

print(im.size)

im.thumbnail((15.2, 15.2))
print(im.size)

im.thumbnail((10, 10))
print(im.size)
```

On 6.2.1:

```
(128, 128)
(15, 15)
(10, 10)
```

On master with this PR:

```
(128, 128)
  File "1.py", line 6, in <module>
    im.thumbnail((15.2, 15.2))
  File "/Users/hugo/github/Pillow/src/PIL/Image.py", line 2153, in thumbnail
    im = self.resize(size, resample)
  File "/Users/hugo/github/Pillow/src/PIL/Image.py", line 1822, in resize
    return self._new(self.im.resize(size, resample, box))
TypeError: integer argument expected, got float
```

Should we fix it something like this?

```diff
diff --git a/src/PIL/Image.py b/src/PIL/Image.py
index 613b3a36..981c8c81 100644
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -2138,10 +2138,10 @@ class Image:
         x, y = self.size
         if x > size[0]:
             y = max(round(y * size[0] / x), 1)
-            x = size[0]
+            x = round(size[0])
         if y > size[1]:
             x = max(round(x * size[1] / y), 1)
-            y = size[1]
+            y = round(size[1])
         size = x, y

         if size == self.size:
```

```
(128, 128)
(15, 15)
(10, 10)
```

radarhere: Okay, I've created PR #4272

PR #4198, opened 2019-11-06
WIP: Enable ImageGrab grab for Linux

vkresch:
Changes proposed in this pull request:

 *  Enable to use ImageGrab.py not only on mac and windows but on linux too
 *  Depends on ImageMagick `sudo apt-get install imagemagick`

Please point me to the parts which also needs to be change like the installation script.

PR #4260, opened 2019-12-09
ImageGrab.grab() for Linux with XCB

### LINKED ISSUE #2028: ImageGrab on Linux
### What did you do?

Got Ubuntu 16.04.

```python
>>> from PIL import ImageGrab
```

```python
>>>
```
### What actually happened?

```python
  File "/usr/lib/python2.7/dist-packages/PIL/ImageGrab.py", line 22, in <module>
    raise ImportError("ImageGrab is OS X and Windows only")
```

```python
>>> PIL.VERSION
'1.1.7'
>>> PIL.PILLOW_VERSION
'3.1.2'
```

So, what is the proper way for saving screens on Linux?

nulano:
Partially resolves #2028

Alternative to #4198

 * Add `ImageGrab.grab()` support for X11 (Linux) using the XCB library (only screenshots for now, clipboard can be added in the future)

There is also support for taking X11 screenshots on Windows and macOS (where X11 can be installed as a separate window manager). This proposed implementation uses the old methods if `xdisplay=None` is passed, but if `xdisplay=""` is passed (or on Linux), it will grab the default X11 screen (likely from the DISPLAY envvar). A different value can be passed in the `xdisplay` parameter to override the default X11 screen address.

Unlike [mss](https://github.com/BoboTiG/python-mss) which uses Xlib, this uses XCB. This provides more resilient error handling at the cost of potentially slightly lower OS compatibility (Xlib is older).

### DIFF
--- a/src/PIL/ImageGrab.py
+++ b/src/PIL/ImageGrab.py
@@ -22,38 +22,62 @@ import tempfile
 
 from . import Image
 
-if sys.platform not in ["win32", "darwin"]:
-    raise ImportError("ImageGrab is macOS and Windows only")
+if sys.platform == "win32":
+    pass
+elif sys.platform == "darwin":
+    import os
+    import tempfile
+    import subprocess
+elif not Image.core.HAVE_XCB:
+    raise ImportError("ImageGrab requires Windows, macOS, or the XCB library")
 
 
-def grab(bbox=None, include_layered_windows=False, all_screens=False):
-    if sys.platform == "darwin":
-        fh, filepath = tempfile.mkstemp(".png")
-        os.close(fh)
-        subprocess.call(["screencapture", "-x", filepath])
-        im = Image.open(filepath)
-        im.load()
-        os.unlink(filepath)
-        if bbox:
-            im_cropped = im.crop(bbox)
-            im.close()
-            return im_cropped
-    else:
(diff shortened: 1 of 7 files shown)


src/PIL/ImageGrab.py
```diff
@@ -79,11 +90,13 @@ def grabclipboard():
...
+    elif sys.platform == "win32":
+        data = Image.core.grabclipboard_win32()
         if isinstance(data, bytes):
             from . import BmpImagePlugin
             import io
 
             return BmpImagePlugin.DibImageFile(io.BytesIO(data))
         return data
+    else:
+        raise IOError("ImageGrab.grabclipboard() is macOS and Windows only")
```
radarhere:
```suggestion
        raise NotImplementedError("ImageGrab.grabclipboard() is macOS and Windows only")
```
Suggestion.

nulano: Good idea!

hugovk:
Thanks!

nulano:
Does [pillow-wheels](https://github.com/python-pillow/pillow-wheels) need to be updated? I'm guessing Windows and macOS don't need XCB support in distributed wheels, as this is quite an unusual (although valid) scenario.

PR #4283, opened 2019-12-20
Raise an error if BMP file size is too large when saving

### LINKED ISSUE #4274: Large BMP files crash (i.e. more than 2**32 pixels)
I started asking on StackOverflow about my original problem, and the code is pasted over there:
https://stackoverflow.com/questions/59353721/python-pillow-opening-1-bit-depth-files-with-8-bits

but after moving to a larger machine (256 GB RAM) I moved on from the original issue and onto an Exception in the code.

Basically I'm trying to save a 59GigaPixel single-channel (1 bit depth) BMP file, which is yielding this error:
```
  File "stitch_bmps.py", line 75, in <module>
    blank.save("59GP.bmp", "BMP")
  File "/home/nmz787/.local/lib/python3.7/site-packages/PIL/Image.py", line 2084, in save
  File "/home/nmz787/.local/lib/python3.7/site-packages/PIL/BmpImagePlugin.py", line 332, in _save
    + o32(offset)  # reserved
  File "/home/nmz787/.local/lib/python3.7/site-packages/PIL/_binary.py", line 91, in o32le
    return pack("<I", i)
struct.error: 'I' format requires 0 <= number <= 4294967295
```

Here's a small example to reproduce easily (requires at least 61GB of RAM, or RAM+swap):
```python
from PIL import Image
a = Image.new("1", (247246, 247246))
a.save("blank_59GP.bmp", "BMP")
```

* OS: Linux version 4.9.164-vs2.3.9.8-beng (root@ex64-jessie) (gcc version 4.9.2 (Debian 4.9.2-10+deb8u2) ) #1 SMP Wed Mar 20 20:18:21 GMT 2019
* PIP: 18.1
* Python: 3.7.3

radarhere:
Resolves #4274, by throwing an error if the file size is too large to fit into 4 bytes.

I set the size artificially in the test to avoid using a large amount of memory.

Also fixes up the comments, which were out of sync. Reference http://www.ece.ualberta.ca/~elliott/ee552/studentAppNotes/2003_w/misc/bmp_file_format/bmp_file_format.htm

### DIFF
--- a/Tests/test_file_bmp.py
+++ b/Tests/test_file_bmp.py
@@ -41,6 +41,13 @@ class TestFileBmp(PillowTestCase):
             self.assertEqual(im.size, reloaded.size)
             self.assertEqual(reloaded.format, "BMP")
 
+    def test_save_too_large(self):
+        outfile = self.tempfile("temp.bmp")
+        with Image.new("RGB", (1, 1)) as im:
+            im._size = (37838, 37838)
+            with self.assertRaises(ValueError):
+                im.save(outfile)
+
     def test_dpi(self):
         dpi = (72, 72)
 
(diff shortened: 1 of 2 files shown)


nmz787:
Thanks!

hugovk:
Thank you!

PR #4285, opened 2019-12-23
Accept io.StringIO as a path for Image.open

### LINKED ISSUE #4097: Loading image from bytes
### What did you do?
Convert string or bytes to PIL.Image

have the PIL.Image instance returned

### What actually happened?
Got a Traceback.

* OS: Windows 7 x64
* Python: 2.7 and also 3.7
* Pillow: 6.1.0

OK so this is the skinny. when running python 2.7 everything works as expected.
when I run the same code using Python 3.7 I get the following Traceback

```python
 File "C:\Program Files\Python37\lib\site-packages\PIL\Image.py", line 2822, in open
    raise IOError("cannot identify image file %r" % (filename if filename else fp))
OSError: cannot identify image file <_io.BytesIO object at 0x0000000003465F68>
```

Here is the test code to use.

```python

import sys
from PIL import Image
from io import BytesIO

# PNG data
LEFT_THUMB = (
    '\x89\x50\x4E\x47\x0D\x0A\x1A\x0A\x00\x00\x00\x0D\x49\x48\x44\x52\x00\x00'
    '\x00\x13\x00\x00\x00\x0B\x08\x06\x00\x00\x00\x9D\xD5\xB6\x3A\x00\x00\x01'
    '\x2E\x49\x44\x41\x54\x78\x9C\x95\xD2\x31\x6B\xC2\x40\x00\x05\xE0\x77\x10'
    '\x42\x09\x34\xD0\x29\x21\x82\xC9\x9C\x2E\x72\x4B\x87\x40\x50\xB9\xBF\x5B'
    '\x28\x35\xA1\xA4\x94\x76\x68\x1C\x1C\x74\xCD\x9A\xE8\x20\x0A\x12\xA5\x5A'
    '\xE4\x72\xC9\x75\x10\x6D\xDC\xCE\xF7\x03\x3E\xDE\x83\x47\xA4\x94\x68\x67'
    '\xB5\xD9\x4E\xBF\xBF\x3E\xE8\x78\x3C\x86\x6A\x3C\xCF\x43\x10\x04\x20\x6D'
    '\x6C\xB5\xD9\x4E\x93\xF8\x95\x5A\x96\x05\xC6\x98\x32\x56\x14\x05\x46\xA3'
    '\x11\xB4\x36\x14\xBD\x3C\xD3\x4E\xA7\x03\xC6\x18\x8E\xC7\x23\x9A\xA6\x51'
    '\xC2\x5C\xD7\x45\x9E\xE7\x27\xEC\x0C\x39\x8E\x03\xC6\x18\x0E\x87\x83\x32'
    '\x04\x00\xE7\x75\x1A\xE7\x7C\xF2\xF9\xFE\x46\x6D\xDB\x06\x63\x0C\xFB\xFD'
    '\x1E\x75\x5D\x2B\x43\x57\x58\xF9\xF3\xAB\xAD\xD7\x6B\x98\xA6\x09\x21\x04'
    '\x76\xBB\x1D\x84\x10\x37\x61\x86\x61\x9C\x30\x00\x70\x1C\x07\x49\x92\x80'
    '\x10\x82\x7E\xBF\x8F\xE5\x72\x79\x13\x78\x69\xF6\x70\x6F\x88\x5E\xAF\x37'
    '\x2B\xCB\x92\xC6\x71\x0C\x42\x08\xC2\x30\xC4\x7C\x3E\x57\x06\x2F\x98\xAE'
    '\xEB\x4F\xAE\xEB\x4E\x06\x83\xC1\x4C\x4A\x49\xA3\x28\x82\x94\x12\x61\x18'
    '\x2A\x37\x5B\x2C\x16\xE8\x76\xBB\xFF\x3F\xE3\x9C\x4F\x8A\xA2\xD0\xD2\x34'
    '\xA5\x59\x96\xA1\xAA\x2A\x65\xCC\xB2\x2C\x0C\x87\xC3\xEB\xD3\x9E\xC1\xAA'
    '\xAA\xEE\x38\xE7\x4A\x90\xAE\xEB\x00\x00\xDF\xF7\x1F\xFF\x00\x09\x7C\xA7'
    '\x93\xB1\xFB\xFA\x11\x00\x00\x00\x00\x49\x45\x4E\x44\xAE\x42\x60\x82'
)

PY3 = sys.version_info[0] > 2

if PY3:
    stream = BytesIO(LEFT_THUMB.encode())
else:
    stream = BytesIO(LEFT_THUMB)

image = Image.open(stream).convert("RGBA")
stream.close()
image.show()
```

Now This is where it goes a little bit sideways. when running Python 2.7 if I use cStringIO.StringIO everything works as expected.. But if I use io.StringIO I get the traceback listed above. I think it has got something to do with the first 16 bytes but I am not 100% sure.

any help is appreciated.

radarhere:
Resolves #4097 

The conclusion of discussion in the issue is that StringIO can't be used to pass bytes into Image.open.

```python
import io
from PIL import Image
im = Image.open(io.StringIO("'\x89\x50\x4E\x47\x0D"))
```
```
OSError: cannot identify image file <_io.StringIO object at 0x104ecdd70>
```

A user suggests that we raise an explicit error if someone passes StringIO in.

However, I like the idea of being able to use StringIO for something, rather than just rejecting it - so this PR adds functionality to use StringIO as a path instead.

```python
import io
from PIL import Image
im = Image.open(io.StringIO("Tests/images/hopper.jpg"))
```

### DIFF
--- a/Tests/test_image.py
+++ b/Tests/test_image.py
@@ -1,3 +1,4 @@
+import io
 import os
 import shutil
 import tempfile
@@ -90,6 +91,11 @@ class TestImage(PillowTestCase):
 
     def test_bad_mode(self):
         self.assertRaises(ValueError, Image.open, "filename", "bad mode")
+
+    def test_stringio(self):
+        with Image.open(io.StringIO("Tests/images/hopper.jpg")) as im:
+            self.assertEqual(im.mode, "RGB")
+            self.assertEqual(im.size, (128, 128))
 
     def test_pathlib(self):
         from PIL.Image import Path
--- a/docs/reference/open_files.rst
+++ b/docs/reference/open_files.rst
@@ -3,10 +3,10 @@
 File Handling in Pillow
 =======================
 
-When opening a file as an image, Pillow requires a filename, ``pathlib.Path``
-object, or a file-like object. Pillow uses the filename or ``Path`` to open a
-file, so for the rest of this article, they will all be treated as a file-like
-object.
+When opening a file as an image, Pillow requires a filename, ``io.StringIO``
+object, ``pathlib.Path`` object, or a file-like object. Pillow uses the
+filename, ``StringIO`` or ``Path`` to open a file, so for the rest of this
+article, they will all be treated as a file-like object.
 
 The following are all equivalent::
 
@@ -15,6 +15,9 @@ The following are all equivalent::
     import pathlib
 
     with Image.open('test.jpg') as im:
+        ...
+
+    with Image.open(io.StringIO('test.jpg')) as im2:
         ...
 
     with Image.open(pathlib.Path('test.jpg')) as im2:
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -2674,8 +2674,8 @@ def open(fp, mode="r"):
     :py:meth:`~PIL.Image.Image.load` method).  See
     :py:func:`~PIL.Image.new`. See :ref:`file-handling`.
 
-    :param fp: A filename (string), pathlib.Path object or a file object.
-       The file object must implement :py:meth:`~file.read`,
+    :param fp: A filename (string), io.StringIO object, pathlib.Path object
+       or a file object. The file object must implement :py:meth:`~file.read`,
        :py:meth:`~file.seek`, and :py:meth:`~file.tell` methods,
        and be opened in binary mode.
     :param mode: The mode.  If given, this argument must be "r".
@@ -2691,6 +2691,8 @@ def open(fp, mode="r"):
     filename = ""
     if isinstance(fp, Path):
         filename = str(fp.resolve())
+    elif isinstance(fp, io.StringIO):
+        filename = fp.getvalue()
     elif isPath(fp):
         filename = fp
 


hugovk:
I wonder how common a use case it is to use StringIO for paths.

The docs say it's "An in-memory stream for text I/O", so it's not really intended for that, and I don't think we need to explicitly support a use case where you first open a file that contains a path to an image (read as StringIO), and then pass that into Pillow.

If anything, I think raising an exception is better.

kdschlosser:
i agree with @hugovk on this as well.

It would be better to be able to use StringIO as a mechanism for passing in string data. because of the end life for python 2 it really would not be needed anymore.. But producing a traceback if using it.. instead of a deceptive error of "cannot identify image file" which leads one to think there is an issue with the data that is contained in the instance. 

I know it is a trivial problem and is easily fixed by the user.. But not having a proper error message caused me to open a support ticket and time to be spent for this when it could have been handled in a  much more graceful manner by providing a less cryptic message that sends a user on a wild goose chase, and would not have required any use of time by PIL developers or supporters.

I am not familiar with what the data looks like inside of these image files or what can and cannot be used. a better solution would be to create another function where a suer does not have to create the BytesIO object they could pass the literal or even any collected data from an external source and PIL would then handle loading that data into whatever container it sees fit to load it into instead of using the open function. a function called open means to me to open something up like a file.. now a function like from_bytes would make me want to pass either a literal or a bytearray and a function name of from_stream would be the passing of a file like object. if using from_bytes or from_stream in both cases the image data is technically already "opened" it simply needs to be loaded. providing a reverse of this in the Image class would also be useful. there is already such a thing to_bytes to return the bytes and there is save() to save it as a file.. but I do not think there is anything like to_stream. 

The reality is that there is a HUGE amount of information that is collected from outside sources like the internet... it is a hassle to have to create an instance of a BytesIO object and pass that and then holding a reference to it and then closing it later on or saving the data to a temp file and then pass the temp file name to the open and the make sure to delete the file when it is no longer used. because the simple fact that it ends up in memory whether it starts off that way or if it gets saved and that file then gets opened up by PIL it is still in memory. having separate entry points depending on the type of data being passed is the typical practice, but if you wanted to create a single entry point for the different types that can be done as well by using c style overloaded methods\functions that would ensure the proper method was used when a certain data type gets passed. I wrote a module that facilitates this. if only a portion of the overloaded method is depreciated you end up having far less code to write to deprecate the thing I also wrote a decorator that makes this process easy as well. 

I do choke every time i say this.. Python is a "beginners" programming language it is the first language a lot of developers learn with. Python has become as popular as it is because it is  "easy" to use. Kind of have to keep that in mind as well. 

my couple of pennies on this.. This is all suggestion or ideas.. I am not the person writing the code for this project... if there is a better way then by all means go that route. My only concern was that I am able to use StringIO in python 2 with PIL without error. but in python 3 it no longer works. so anyone porting their code to python 3 when python 2 is EOL and they  are using StringIO are going to end up bumping into this issue, it might be the gravy train on biscuit wheels at the moment.. that could change in the not to distant future.

PR #4286, opened 2019-12-23
Use the same Python version to build docs in Travis and  GHA

radarhere:
We're currently using [3.7 for Travis](https://github.com/python-pillow/Pillow/blob/master/.travis/test.sh#L8) and [3.8 for GHA](https://github.com/python-pillow/Pillow/blob/master/.github/workflows/test.yml#L87)


hugovk:
How about using 3.8 for both?

radarhere:
Okay, done.

PR #4287, opened 2019-12-23
Changed to publicly visible GHA link

radarhere:
https://github.com/python-pillow/Pillow/actions isn't actually publicly visible - if you open the link without being logged in to GitHub, you get a 404. So rather than linking to all of the action jobs, this PR does the next best thing, linking to just set of one action jobs - https://github.com/python-pillow/Pillow/commit/6e0f07bbe38def22d36ee176b2efd9ea74b453a6/checks

### DIFF
--- a/docs/about.rst
+++ b/docs/about.rst
@@ -12,7 +12,7 @@ The fork author's goal is to foster and support active development of PIL throug
 
 .. _Travis CI: https://travis-ci.org/python-pillow/Pillow
 .. _AppVeyor: https://ci.appveyor.com/project/Python-pillow/pillow
-.. _GitHub Actions: https://github.com/python-pillow/Pillow/actions
+.. _GitHub Actions: https://github.com/python-pillow/Pillow/commit/6e0f07bbe38def22d36ee176b2efd9ea74b453a6/checks
 .. _GitHub: https://github.com/python-pillow/Pillow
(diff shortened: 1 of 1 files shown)


hugovk:
Sound like a bug, and I can repro with another repo. 

The GH team applied a manual fix here, and I can access their actions now:

https://github.community/t5/GitHub-Actions/Actions-tab-missing-actions-URL-404-runs-work/td-p/36462

Please could you contact GH?

radarhere:
Done. Here's my message and the response.

> Hi. While an individual Actions job like https://github.com/python-pillow/Pillow/commit/6e0f07bbe38def22d36ee176b2efd9ea74b453a6/checks is visible when logged out of GitHub, the collective Actions like https://github.com/python-pillow/Pillow/actions is not. Is this intentional, or something that will be fixed as GitHub Actions continues to evolve? Thanks very much for your time.

> Thanks for reaching out and asking us about this! I think it's a great question that we don't currently have an answer to smile_cat As a next step, I've asked our Actions team for their input on this one. We'll write back to you as soon as we have an update from them, though I can't make any promises for when that will be.

hugovk:
https://github.com/python-pillow/Pillow/actions is now visible when logged out.

radarhere:
Cool, thanks for spotting.

PR #4290, opened 2019-12-23
Fixes djpeg load test

bluepython508:
* Test fails with `libjpeg-turbo` and `libjpeg-progs` on Ubuntu 16.04
* Epsilon reported is 4.18... - This PR sets it to 5.0

The test test_load_djpeg fails if the djpeg decoder is different from the one Pillow is compiled against.

### DIFF
--- a/Tests/test_file_jpeg.py
+++ b/Tests/test_file_jpeg.py
@@ -494,7 +494,7 @@ class TestFileJpeg(PillowTestCase):
     def test_load_djpeg(self):
         with Image.open(TEST_FILE) as img:
             img.load_djpeg()
-            assert_image_similar(img, Image.open(TEST_FILE), 0)
+            assert_image_similar(img, Image.open(TEST_FILE), 5)
 
     @unittest.skipUnless(cjpeg_available(), "cjpeg not available")
     def test_save_cjpeg(self):


radarhere:
Have you visually inspected the loaded image, to make sure that the loaded difference of 5 isn't because of some glaring difference? Or better yet, could you attach an image to show us what you're seeing?

```python
with Image.open(TEST_FILE) as img:
        img.load_djpeg()
	img.save("loaded_image.png")
```

bluepython508:
The image is visually the same.
img This is the image loaded with djpeg and saved to png.

homm:
`self.assert_image_similar` has an unpleasant feature: for multichannel images, it counts the sum of epsilons for each channel. So the real difference is about 1.4 for every value on each pixel. Counting the number of transformations in JPEG format, I believe this is a reasonable difference across different implementations.

wiredfool:
@homm That's sort of an issue, but unrelated to this. 

The main issue here is that there's no guarantee that djpeg and pillow are using the same libjpeg. I'm confused at the original `assert_image_similar(,,0)`, meaning, similar, but not different.

radarhere:
I was able to replicate this using our Ubuntu 16.04 docker image, removing our libjpeg-turbo-progs, and adding libjpeg-turbo8 and libjpeg-progs.

PR #4292, opened 2019-12-23
Added reading and writing of private PNG chunks

### LINKED ISSUE #4265: Allow non-standard chunks to be saved in PNG files
### What did you do?
I'm making a simple tool that's modifying sprites for ZDoom source port of Doom video game. It uses non-standard PNG chunks to store additional information, eg sprite offsets.

```python
import struct
from PIL import Image
from PIL.PngImagePlugin import PngInfo

img = Image.open('src.png')
pnginfo = PngInfo()
pnginfo.add(b'grAb', struct.pack('>II', 10, 49))
img.save('dst.png', pnginfo=pnginfo)
```

To see a 'grAb' chunk inside saved PNG file (like source image has)

### What actually happened?
Chunk have not been saved because there is a forced list of chunks that's being saved inside `src/PIL/PngImagePlugin.py`

* OS: GNU/Linux
* Python: 3.8.0
* Pillow: 6.2.1

As far as I understand PNG specification doesn't strictly forbid saving unknown chunks with "Safe-to-copy bit" set to 0. PIL is great tool for image manipulation automation, I think it would be great to give developers more ways of using it. Thanks!

radarhere:
Resolves #4265

Private PNG chunks can now be read as
```python
from PIL import Image
im = Image.open("Tests/images/exif.png")
im.private_chunks #  [(b'orNT', b'\x01')]
```

The issue states that the spec doesn't exclude writing unknown chunks - but I find it actually describes how private chunks are to be used.

From https://www.w3.org/TR/PNG/#5Chunk-naming-conventions, if the second letter is lowercase, the chunk is private.

From https://www.w3.org/TR/PNG/#14Ordering
> When copying an unknown safe-to-copy ancillary chunk, a PNG editor shall not move the chunk from before IDAT to after IDAT or vice versa.

Pillow doesn't automatically save private chunks, but it should give users the ability to keep the ordering. So I have set it so that `private_chunks` may have a third value to specify if a chunk is after IDAT when reading, and I have added an `after_idat` argument to `PngInfo.add` to specify when writing.

### DIFF
--- a/src/PIL/PngImagePlugin.py
+++ b/src/PIL/PngImagePlugin.py
@@ -228,15 +228,20 @@ class PngInfo:
     def __init__(self):
         self.chunks = []
 
-    def add(self, cid, data):
+    def add(self, cid, data, after_idat=False):
         """Appends an arbitrary chunk. Use with caution.
 
         :param cid: a byte string, 4 bytes long.
         :param data: a byte string of the encoded data
+        :param after_idat: for use with private chunks. Whether the chunk
+                           should be written after IDAT
 
         """
 
-        self.chunks.append((cid, data))
+        chunk = [cid, data]
+        if after_idat:
+            chunk.append(True)
+        self.chunks.append(tuple(chunk))
 
     def add_itxt(self, key, value, lang="", tkey="", zip=False):
         """Appends an iTXt chunk.
@@ -641,6 +646,7 @@ class PngImageFile(ImageFile.ImageFile):
         #
         # Parse headers up to the first IDAT or fDAT chunk
 
+        self.private_chunks = []
         self.png = PngStream(self.fp)
 
         while True:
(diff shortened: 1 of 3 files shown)


src/PIL/PngImagePlugin.py
```diff
@@ -216,15 +216,20 @@ class PngInfo:
...
+        :param after_idat: for use with private chunks. Whether the chunk
+                           should be written after IDAT
 
         """
 
-        self.chunks.append((cid, data))
+        chunk = [cid, data]
+        if after_idat:
+            chunk.append(True)
+        self.chunks.append(tuple(chunk))
```
homm: Code would simpler if you always save the flag.

```suggestion
        self.chunks.append((cid, data, after_idat))
```

radarhere: My thinking was that this is more backwards compatible.

If someone had this code before -

```python
from PIL import PngImagePlugin
info = PngImagePlugin.PngInfo()
info.add("priV", "\x00")
cid, data = info.chunks[0]
```

Then your suggestion would break it with `ValueError: too many values to unpack`. I will say that using the class in this way doesn't seem likely however.

homm: Since `PngInfo.chunks` is undocumented, I think we can sacrifice backward compatibility for convenience.

PR #1169, opened 2015-04-02
pcd: minor cleanup; remove draft method

benoit-pierre:
It's not supported, and the implementation does nothing.

PR #4293, opened 2019-12-24
Updated draft documentation

radarhere:
The PCD `draft` method was removed in #1169.

Meanwhile, MpoImageFile has a `draft` method because it inherits from JpegImageFile - https://github.com/python-pillow/Pillow/blob/fafeeefccda1d24f8ee33d2f2fd4e339ea2d64a3/src/PIL/MpoImagePlugin.py#L38


hugovk:
Thank you

PR #4297, opened 2019-12-25
Change ContainerIO return type to match file object mode

radarhere:
Alternative to #4192 

That PR changes ContainerIO to return bytes instead of strings.
This PR changes ContainerIO to return strings if the file object is a string object, and bytes if the file object is a bytestring object.

### DIFF
--- a/src/PIL/ContainerIO.py
+++ b/src/PIL/ContainerIO.py
@@ -93,12 +93,13 @@ class ContainerIO:
         :returns: An 8-bit string.
         """
         s = b"" if "b" in self.fh.mode else ""
+        newline_character = b"\n" if "b" in self.fh.mode else "\n"
         while True:
             c = self.read(1)
             if not c:
                 break
             s = s + c
-            if c == (b"\n" if "b" in self.fh.mode else "\n"):
+            if c == newline_character:
                 break
         return s
 
--- a/src/PIL/ContainerIO.py
+++ b/src/PIL/ContainerIO.py
@@ -82,7 +82,7 @@ class ContainerIO:
         else:
             n = self.length - self.pos
         if not n:  # EOF
-            return b""
+            return b"" if "b" in self.fh.mode else ""
         self.pos = self.pos + n
         return self.fh.read(n)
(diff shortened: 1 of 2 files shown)


src/PIL/ContainerIO.py
```diff
@@ -92,13 +92,13 @@ def readline(self):
...
         """
-        s = b""
+        s = b"" if "b" in self.fh.mode else ""
         while True:
             c = self.read(1)
             if not c:
                 break
             s = s + c
-            if c == b"\n":
+            if c in [b"\n", "\n"]:
```
jdufresne: This line results in a `BytesWarning` when running with the Python `-b` argument:

```
Tests/test_file_tar.py::TestFileTar::test_sanity
.../Pillow/.tox/py38/lib/python3.8/site-packages/PIL/ContainerIO.py:101: BytesWarning: Comparison between bytes and string
    if c in [b"\n", "\n"]:
```

jdufresne:
Overall, the approach makes sense. Do you have a use case in mind where the file would be opened in text mode?

src/PIL/ContainerIO.py
radarhere: Thanks. I've updated the commit.

radarhere:
No. Feel free to argue that bytes mode is correct. I'm just thinking about backwards compatibility.

src/PIL/ContainerIO.py
```diff
@@ -92,13 +92,13 @@ def readline(self):
...
         """
-        s = ""
+        s = b"" if "b" in self.fh.mode else ""
         while True:
             c = self.read(1)
             if not c:
                 break
             s = s + c
-            if c == "\n":
+            if c == (b"\n" if "b" in self.fh.mode else "\n"):
```
jdufresne: This could be pulled out of the loop so it doesn't need to be recalculated every iteration of the loop.

radarhere: Good idea. I've pushed a commit for this.

PR #4305, opened 2019-12-26
Replaced distutils with C version check

### LINKED ISSUE #4211: ImportError from distutils in exe built with cx_Freeze in virtualenv
Running a trivial Python program using Pillow built into a Windows EXE file using cx_Freeze out of a virtualenv fails with `ImportError: cannot import name 'dist' from 'distutils'`

Running the same directly using Python, without building the EXE, works.

I guess this is caused by https://github.com/python-pillow/Pillow/pull/3350 (which introduced a dependency on distutils.version) and the solution could possibly be https://github.com/python-pillow/Pillow/pull/3552, which was proposed already in https://github.com/python-pillow/Pillow/issues/3547 as a solution to another virtualenv-related issue.

The underlying cause seems to be that the distutils in the virtualenv is not normal distutils but something fake from virtualenv. Why it works outside of the built EXE is a mystery to me, I'm guessing some magic with sys.path or import hooks. Why it doesn't work in the EXE file seems clearer - distutils.version is simply not included.

To reproduce:
repro.cmd:
```cmd
virtualenv .
call Scripts\activate
pip install cx_freeze pillow
python setup.py bdist
call Scripts\deactivate
build\exe.win-amd64-3.7\pok.exe
```

setup.py:
```python
from cx_Freeze import setup, Executable

setup(
    options = {"build_exe": {
        "include_msvcr": True,
        "packages": ["PIL"],
        "zip_include_packages": "*",
        "zip_exclude_packages": "",
        "excludes": ["tkinter"]
        }
    },
    executables = [Executable("pok.py")],
)
```

pok.py:
```python
import PIL.Image
import PIL.TiffImagePlugin
```

`pip list`:
```
Package    Version
---------- -------
cx-Freeze  6.0
Pillow     6.2.1
pip        19.3.1
setuptools 41.6.0
wheel      0.33.6
```

Python version: Python 3.7.4 (tags/v3.7.4:e09359112e, Jul  8 2019, 20:34:20) [MSC v.1916 64 bit (AMD64)] on win32

radarhere:
Resolves #4211

This PR is another version of #3552, again the goal of removing distutils in order to resolve problems external to Pillow. That PR was not accepted, so I understand if this one is not either.

This time, I'm suggesting moving the Python-level version check into C.

It then duplicates the version check performed in TiffDecode.c -
https://github.com/python-pillow/Pillow/blob/661ceac71739023b55c2849bb80dd46b7ef586bd/src/libImaging/TiffDecode.c#L556-L561

However, the two checks are not for exact the same purpose. The check in TiffDecode.c is for ease of compilation because of the different return types between libtiff versions, whereas this one is about functionality.

### DIFF
--- a/src/PIL/TiffImagePlugin.py
+++ b/src/PIL/TiffImagePlugin.py
@@ -38,7 +38,6 @@
 #
 # See the README file for information on usage and redistribution.
 #
-import distutils.version
 import io
 import itertools
 import os
@@ -1559,11 +1558,10 @@ def _save(im, fp, filename):
             # Custom items are supported for int, float, unicode, string and byte
             # values. Other types and tuples require a tagtype.
             if tag not in TiffTags.LIBTIFF_CORE:
-                if TiffTags.lookup(tag).type == TiffTags.UNDEFINED:
-                    continue
-                if distutils.version.StrictVersion(
-                    _libtiff_version()
-                ) < distutils.version.StrictVersion("4.0"):
+                if (
+                    TiffTags.lookup(tag).type == TiffTags.UNDEFINED
+                    or not Image.core.libtiff_v4_or_greater
+                ):
                     continue
 
                 if tag in ifd.tagtype:
--- a/src/PIL/TiffImagePlugin.py
+++ b/src/PIL/TiffImagePlugin.py
@@ -263,10 +263,6 @@ def _limit_rational(val, max_val):
     return n_d[::-1] if inv else n_d
 
 
-def _libtiff_version():
-    return Image.core.libtiff_version.split("\n")[0].split("Version ")[1]
-
-
 ##
 # Wrapper for TIFF IFDs.
 
--- a/src/PIL/TiffImagePlugin.py
+++ b/src/PIL/TiffImagePlugin.py
@@ -1556,7 +1556,7 @@ def _save(im, fp, filename):
             if tag not in TiffTags.LIBTIFF_CORE:
                 if (
                     TiffTags.lookup(tag).type == TiffTags.UNDEFINED
-                    or not Image.core.libtiff_v4_or_greater
+                    or not Image.core.libtiff_support_custom_tags
                 ):
                     continue
 
--- a/src/_imaging.c
+++ b/src/_imaging.c
@@ -3969,13 +3969,13 @@ setup_module(PyObject* m) {
     PyDict_SetItemString(d, "libtiff_version", PyUnicode_FromString(ImagingTiffVersion()));
 
     // Test for libtiff 4.0 or later, excluding libtiff 3.9.6 and 3.9.7
-    PyObject* v4_or_greater;
+    PyObject* support_custom_tags;
 #if TIFFLIB_VERSION >= 20111221 && TIFFLIB_VERSION != 20120218 && TIFFLIB_VERSION != 20120922
-    v4_or_greater = Py_True;
+    support_custom_tags = Py_True;
 #else
-    v4_or_greater = Py_False;
+    support_custom_tags = Py_False;
 #endif
-    PyDict_SetItemString(d, "libtiff_v4_or_greater", v4_or_greater);
+    PyDict_SetItemString(d, "libtiff_support_custom_tags", support_custom_tags);
   }
 #endif
 
--- a/src/_imaging.c
+++ b/src/_imaging.c
@@ -82,6 +82,12 @@
 #include "zlib.h"
 #endif
 
+#ifdef HAVE_LIBTIFF
+#ifndef _TIFFIO_
+#include <tiffio.h>
+#endif
+#endif
+
 #include "Imaging.h"
 
 #define _USE_MATH_DEFINES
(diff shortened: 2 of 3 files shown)


src/PIL/TiffImagePlugin.py
```diff
@@ -1559,11 +1558,10 @@ def _save(im, fp, filename):
...
             # values. Other types and tuples require a tagtype.
             if tag not in TiffTags.LIBTIFF_CORE:
-                if TiffTags.lookup(tag).type == TiffTags.UNDEFINED:
-                    continue
-                if distutils.version.StrictVersion(
-                    _libtiff_version()
-                ) < distutils.version.StrictVersion("4.0"):
+                if (
+                    TiffTags.lookup(tag).type == TiffTags.UNDEFINED
+                    or Image.core.libtiff_mergefieldinfo_return != "true"
```
homm: Should be `True` constant, not string.

radarhere: Ok, done.

src/PIL/TiffImagePlugin.py
```diff
@@ -1559,11 +1558,10 @@ def _save(im, fp, filename):
             # Custom items are supported for int, float, unicode, string and byte
             # values. Other types and tuples require a tagtype.
             if tag not in TiffTags.LIBTIFF_CORE:
-                if TiffTags.lookup(tag).type == TiffTags.UNDEFINED:
-                    continue
-                if distutils.version.StrictVersion(
-                    _libtiff_version()
```
homm: `_libtiff_version` function isn't used anywhere? Maybe remove?

src/_imaging.c
```diff
@@ -82,6 +82,12 @@
 #include "zlib.h"
 #endif
 
+#ifdef HAVE_LIBTIFF
+#ifndef _TIFFIO_
+#include <tiffio.h>
```
homm: Why is that?

src/PIL/TiffImagePlugin.py
radarhere: Okay, done.

src/_imaging.c
radarhere: Coping the include from TiffDecode.h to bring in `TIFFLIB_VERSION`

src/_imaging.c
```diff
@@ -3961,6 +3967,15 @@ setup_module(PyObject* m) {
...
     PyDict_SetItemString(d, "libtiff_version", PyUnicode_FromString(ImagingTiffVersion()));
+
+    // Test for libtiff 4.0 or later, excluding libtiff 3.9.6 and 3.9.7
+    PyObject* v4_or_greater;
+#if TIFFLIB_VERSION >= 20111221 && TIFFLIB_VERSION != 20120218 && TIFFLIB_VERSION != 20120922
+    v4_or_greater = Py_True;
+#else
+    v4_or_greater = Py_False;
+#endif
+    PyDict_SetItemString(d, "libtiff_v4_or_greater", v4_or_greater);
```
homm: `PyDict_SetItemString` increases refs counters. Ok here.

nulano:
> However, the two checks are not for exact the same purpose. The check in TiffDecode.c is for ease of compilation because of the different return types between libtiff versions, whereas this one is about functionality.

It seems to me that the two checks are in fact used for the same purpose. If I understand #3350 properly, custom tags are only supported when it is possible to check `TIFFMergeFieldInfo` succeeded (i.e. when it has a return value), presumably to avoid segmentation faults if the old function silently fails. The function `ImagingLibTiffMergeFieldInfo` containing the original C version check is only called for custom tags, which are only provided from Python code if it detects the new version of libtiff. I would therefore suggest renaming `libtiff_v4_or_greater` to `libtiff_support_custom_tags`.

radarhere:
The reason custom tags aren't used when using libtiff < 4 isn't to be able to use the return status to ensure success - it's simply that on < 4 there was an error that I couldn't figure out how to fix.

Of course, that doesn't stop ensuring success from being a reason to avoid < 4 if others think that it needs to be. I have pushed a commit renaming it to `libtiff_custom_tags_status_return`.

nulano:
> The reason custom tags aren't used when using libtiff < 4 isn't to be able to use the return status to ensure success - it's simply that on < 4 there was an error that I couldn't figure out how to fix.
>
> Of course, that doesn't stop ensuring success from being a reason to avoid < 4 if others think that it needs to be. I have pushed a commit renaming it to `libtiff_custom_tags_status_return`.

I don't think the return value is necessary to be checked, that was just my best guess for what error you could have encountered. I now agree that the two checks have a different purpose. I therefore feel like the new name is even less descriptive than `libtiff_v4_or_greater`.

hugovk:
OK, revert the rename and merge? :)

nulano:
> OK, revert the rename and merge? :)

I would prefer `libtiff_support_custom_tags` as that describes how the value is used, but `libtiff_v4_or_greater` is also fine, as it describes what is being checked.

I do not like `libtiff_custom_tags_status_return`, as I now understand this is unrelated to the Python side check.

What is your preference @hugovk ?

hugovk:
`libtiff_support_custom_tags` is more descriptive of intent than `libtiff_v4_or_greater`.

PR #4311, opened 2019-12-28
Allow loading of WMF images at a given DPI

### LINKED ISSUE #4223: wmf file to bitmap was forced to 72 dpi
### What did you do?
I read in an wmf image and then save it as a png image. I found the resolution of the result image is rather low no matter how I set the quality parameters. 

When I looking into the source code ( src/PIL/WmfImagePlugin.py), I found the resolution was fixed to 72dpi.

* OS:  win10
* Python: 3.6.8
* Pillow: 6.1.0

radarhere:
Resolves #4223

WmfImagePlugin can load both WMF and EMF files. When loading WMFs, it sets the DPI to 72. This is an arbitrary number, because [WMFs are like SVGs](https://en.wikipedia.org/wiki/Windows_Metafile) - they don't have an intrinsic DPI, and can be scaled to any resolution. [The Python code just tells the C code what size is required.](https://github.com/python-pillow/Pillow/blob/a1635ac6c45ab4731700336d128ec720cec77857/src/PIL/WmfImagePlugin.py#L46-L56)

This PR suggests that after opening an WMF file, but not an EMF, the following code be used to set the DPI to 144, for example, by adding a `dpi` argument to WmfImagePlugin `load` -
```python
with Image.open("Tests/images/drawing.wmf") as im:
    im.load(144)
```

Feel free to suggest an alternative API.


radarhere:
I've updated the documentation - I've moved WMF from identify-only into read-only. The Windows read functionality was actually added in PIL 1.1.17, but I've removed the references to PIL versions.

radarhere:
Okay, done.

PR #4316, opened 2019-12-30
Added coverage for Docker GHA

### LINKED ISSUE #4212: Add coverage for Docker builds
For #3606.

We're testing a number of Docker images on Travis CI (`env: DOCKER=...` in `.travis.yml`).

We're also testing them on GitHub Actions (`.github/workflows/test-docker.yml`, added in https://github.com/python-pillow/Pillow/pull/4066) and Azure Pipelines (`azure-pipelines.yml`, https://github.com/python-pillow/Pillow/pull/3670).

However, the Travis builds are tested with coverage which is sent to Codecov, but GitHub Actions and Azure Pipelines do not.

To speed up builds, it would be good to enable coverage on GHA or AP so we can stop testing Docker on Travis.

I don't have a strong preference for GHA or AP. We've started using GHA for Linux, macOS and Windows testing, so AP definitely would have more capacity free and would make things faster overall, and with eggs in more baskets. But GHA would mean there's fewer CI systems to maintain, and it feels a bit easier.

radarhere:
Resolves #4212 if this works.

### DIFF
--- a/.github/workflows/test-docker.yml
+++ b/.github/workflows/test-docker.yml
@@ -42,3 +42,26 @@ jobs:
         # The Pillow user in the docker container is UID 1000
         sudo chown -R 1000 $GITHUB_WORKSPACE
         docker run -v $GITHUB_WORKSPACE:/Pillow pythonpillow/${{ matrix.docker }}:${{ matrix.dockerTag }}
+        sudo chown -R runner $GITHUB_WORKSPACE
+
+    - name: After success
+      if: success()
+      run: |
+        pip install wheel
+        sudo apt-get install -qq ruby-dev
+        PATH="$PATH:~/.local/bin"
+        .travis/after_success.sh
+      env:
+        MATRIX_OS: ${{ matrix.os }}
+        COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}
+
+    - name: Prepare coverage token
+      if: success() && github.repository == 'python-pillow/Pillow'
+      run: cp .github/codecov-upstream.yml .codecov.yml
+
+    - name: Upload coverage
+      if: success()
+      uses: codecov/codecov-action@v1
+      with:
+        token: ${{ secrets.CODECOV_TOKEN }}
+        name: ${{ matrix.os }} Python ${{ matrix.python-version }}
--- a/.github/workflows/test-docker.yml
+++ b/.github/workflows/test-docker.yml
@@ -52,7 +52,6 @@ jobs:
         PATH="$PATH:~/.local/bin"
         .travis/after_success.sh
       env:
-        MATRIX_OS: ${{ matrix.os }}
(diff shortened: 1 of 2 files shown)


.github/workflows/test-docker.yml
```diff
@@ -42,3 +42,26 @@ jobs:
...
+
+    - name: After success
+      if: success()
+      run: |
+        pip install wheel
+        sudo apt-get install -qq ruby-dev
+        PATH="$PATH:~/.local/bin"
+        .travis/after_success.sh
+      env:
+        MATRIX_OS: ${{ matrix.os }}
```
hugovk:
```suggestion
```

There's no `matrix.os` here, and it's used in `after_success.sh` like this, so we can remove it from here.

```sh
if [[ "$MATRIX_OS" == "macOS-latest" ]]; then
    brew install lcov
else
    sudo apt-get -qq install lcov
fi
```

.github/workflows/test-docker.yml
```diff
@@ -42,3 +42,26 @@ jobs:
...
+    - name: Prepare coverage token
+      if: success() && github.repository == 'python-pillow/Pillow'
+      run: cp .github/codecov-upstream.yml .codecov.yml
+
+    - name: Upload coverage
+      if: success()
+      uses: codecov/codecov-action@v1
+      with:
+        token: ${{ secrets.CODECOV_TOKEN }}
+        name: ${{ matrix.os }} Python ${{ matrix.python-version }}
```
hugovk:
```suggestion
        name: ${{ matrix.docker }}
```

image

The builds are getting to Codecov, but just as "Python". Let's just name them after the Docker image.

hugovk:
Thank you, looks good!

---

Total coverage with this PR is 87.99%:

* https://codecov.io/gh/python-pillow/Pillow/commit/0893515c98f401cccd0a9940dd6d9a950f49c7d9/build

The same branch, with Docker jobs removed from Travis CI, is also at 87.99%:

* https://github.com/hugovk/Pillow/commit/ced1098e36940c7f2bd7570981f78d13729127df
* https://codecov.io/gh/hugovk/Pillow/commit/ced1098e36940c7f2bd7570981f78d13729127df/build

---

I see there's an exception with coveralls-merge on GHA:

```
coveralls.exception.CoverallsException: Not on TravisCI. You have to provide either repo_token in .coveralls.yml or set the COVERALLS_REPO_TOKEN env var.
```

* https://github.com/python-pillow/Pillow/pull/4316/checks?check_run_id=367666705

But there's a different exception with coveralls-merge on Travis:

```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xf1 in position 99: invalid continuation byte
```

* https://travis-ci.org/python-pillow/Pillow/jobs/630891624#L2532

Let's not worry about that for now, at least it's the same end result for both.

---

Maybe we keep the Dockers in Travis until January?

PR #4320, opened 2019-12-31
Fix rounding error on RGB to L conversion

### LINKED ISSUE #3800: Incorrect Grayscale Conversion
### What did you do?

I'm having a weird inconsistency in the way Pillow is treating grayscale images, and the difference is enough to throw off some metrics I need to compute. After a lot of debugging I was able to trace the inconsistency back to the grayscale conversion. I have included a comparison with OpenCV which I have verified to be the correct conversion. By correct conversion I mean one that is consistent with other programs I have tried (they all agree with OpenCV)

It is very hard to detect this difference by just looking at the grayscale ouputs but I am including them for completeness. The difference becomes extremely apparent under aggressive JPEG compression.

##### Pillow Output (verified wrong)
pillow_output

##### OpenCV Output (verified correct)
opencv_output

##### Difference
diff

Here is the code used to compute these images:

```python
im = np.asarray(Image.open(args.input).convert('L'))
im2 = cv2.cvtColor(cv2.imread(args.input), cv2.COLOR_BGR2GRAY)

diff = im - im2

cv2.imwrite('pillow_output.png', im)
cv2.imwrite('opencv_output.png', im2)
cv2.imwrite('diff.png', diff)
```

**Things to Note**

* This difference is *only* apparent when I do a grayscale conversion. If I leave the images as color images they are identical
* OpenCV loads the images as BGR so line 2 is not a typo
* Maybe this is related to a difference in the chosen luma transform? 

* OS: Various Linux
* Python: 3.5 and 3.6
* Pillow: 6.0.0

homm:
Fixes #3800


homm:
This code is altered by me but I believe the error was there from the beginning. I think this is relatively safe for merge in 7.0.0.

radarhere:
Looks like you're solving the [second problem in the thread](https://github.com/python-pillow/Pillow/issues/3800#issuecomment-502750340), but [the original poster would like our results to exactly match OpenCV](https://github.com/python-pillow/Pillow/issues/3800#issuecomment-487989663). This PR brings Pillow closer, but doesn't exactly match it.

homm:
We don't have an aim "to do everything like OpenCV". The real aim is to do everything correctly and OpenCV is wrong very often.

```python
from PIL import Image
import cv2, numpy

im = Image.open('../imgs/56811035-ce7e0000-6805-11e9-9123-b297635e739e.png')
num_im = numpy.array(im)

ref = (num_im[:, :, 0] * 0.299 + num_im[:, :, 1] * 0.587 + num_im[:, :, 2] * 0.114 + 0.5).astype('uint8')
res_pillow = numpy.array(im.convert('I'))
res_cv2 = cv2.cvtColor(num_im, cv2.COLOR_RGB2GRAY)

print(numpy.count_nonzero(ref - res_pillow))
print(numpy.count_nonzero(ref - res_cv2))
```

So Pillow's result much closer to reference. I think this is because OpenCV uses [fewer significant bits](https://github.com/opencv/opencv/blob/11b020b9f9e111bddd40bffe3b1759aa02d966f0/modules/imgcodecs/src/utils.cpp#L54).

Anyway, now this is a question of precision, not rounding error which was truly an error.

src/libImaging/Convert.c
```diff
@@ -44,7 +44,8 @@
 #define L(rgb)\
     ((INT32) (rgb)[0]*299 + (INT32) (rgb)[1]*587 + (INT32) (rgb)[2]*114)
```
radarhere: If you've changed ImageColor, shouldn't this be changed as well?
https://github.com/python-pillow/Pillow/blob/fa1deca144210e519acc71e8fa9c8e004fc6f760/src/PIL/ImageColor.py#L137

homm: Not the "L" macro itself should be fixed. This macro is used for F mode for example, where no rounding needed. Instead, every place where "L" is used should be analyzed. This is a bit more work which needs more attention and I haven't much time today to do this before the release.

homm: Actually, I've fixed ImageColor.getcolor implementation.

Issue #4277, opened 2019-12-18
Unexpected, additional outline in ImageDraw.ellipse, even for outline=None


HansHirse:
### What did you do?

Drawing a polygon using [`ImageDraw.polygon`](https://pillow.readthedocs.io/en/stable/reference/ImageDraw.html#PIL.ImageDraw.PIL.ImageDraw.ImageDraw.polygon) and an ellipse using [`ImageDraw.ellipse`](https://pillow.readthedocs.io/en/stable/reference/ImageDraw.html#PIL.ImageDraw.PIL.ImageDraw.ImageDraw.ellipse) with no outline specified, with an opaque outline, and with a semi-transparent outline.

### What did you expect to happen?

[`ImageDraw.polygon`](https://pillow.readthedocs.io/en/stable/reference/ImageDraw.html#PIL.ImageDraw.PIL.ImageDraw.ImageDraw.polygon) and [`ImageDraw.ellipse`](https://pillow.readthedocs.io/en/stable/reference/ImageDraw.html#PIL.ImageDraw.PIL.ImageDraw.ImageDraw.ellipse) should behave the same for the mentioned cases.

### What actually happened?

The output created by the code below:

![Figure_1](https://user-images.githubusercontent.com/2692578/71064597-bd587980-216f-11ea-90f1-585e27110d11.png)

Looking at the polygon, if no outline is specified (top left image), we see that the black line is visible, which is one of the polygon's borders. Specifying an opaque outline (top center image), the black line's no longer visible. Setting a semi-transparent outline (top right image) reveals, that the outline is identical to the polygon's border.

Now, the same for the ellipse: If no outline is set (top left), an outline is shown nevertheless, most likely the same color as used for the `fill` parameter, but without incorporating an alpha value. Setting an opaque outline (top center) "overwrites" the unexpected existent outline, but when setting a semi-transparent outline, we see that the unexpected outline is still there.

This effect becomes even more obvious, when setting `width > 1` in `ellipse`, see the bottom row. The unexpected outline still seems to have `width = 1`, whereas the explicitly set outline has `width = 5`.

### What are your OS, Python and Pillow versions?

* OS: Windows 10 Prof.
* Python: 3.7.1
* Pillow: 6.2.1

```python
from matplotlib import pyplot as plt
from PIL import Image, ImageDraw


def example(outline_alpha=None, width=None):
    if outline_alpha is None:
        outline = None
    else:
        outline = (255, 255, 0, outline_alpha)
[...]

PR #4333, opened 2020-01-05
Only draw each polygon pixel once

radarhere:
Helps #4277 

When Pillow draws an ellipse, it considers it to be a 360-sided shape. Expressing that within an ordinary width and height means that those edges overlap.

In such a situation, the `polygon_generic` function will draw pixels more than once. This is usually unnoticed, since a solid pixel on top of another solid pixel is still just a solid pixel. However, if the pixels drawn have a partial opacity, then repeated drawing of a pixel changes the final opacity.

```python
im = Image.new('RGB', (45, 45), (255, 255, 255))
d = ImageDraw.Draw(im, 'RGBA')
d.ellipse([(0, 0), (45, 45)], (0, 0, 255, 100))
im.save('out.png')
```

out

As background, [the `polygon_generic` function uses a scanline algorithm](https://github.com/python-pillow/Pillow/blob/bfc478c0dbb2cb8265c7787e426e759fb7cdcc16/src/libImaging/Draw.c#L419), drawing each row in turn,. Within each row, it starts drawing a line at the edge of a polygon, stops at the next edge, starts at the following edge, and so on.

This PR updates the function to only draw each pixel once.

out


hugovk:
Thanks!

PR #4334, opened 2020-01-05
Allow explicit zero width to hide outline

radarhere:
Helps #4277 

ImageDraw operations `chord`, `ellipse`, `pieslice` and `rectangle` take optional arguments `outline` with a default of `None` and `width` with a default of `0`.

At present, specifying an outline and a width of zero stills draws the outline. The following three lines all produce the same result -

```python
draw.rectangle(BBOX2, fill="green", outline="yellow")
draw.rectangle(BBOX2, fill="green", outline="yellow", width=0)
draw.rectangle(BBOX2, fill="green", outline="yellow", width=1)
```

This PR changes the behaviour so that a zero width produces no outline. For backwards compatibility, it changes the `width` default to `1`, so that these two lines still produce the same result -

```python
draw.rectangle(BBOX2, fill="green", outline="yellow")
draw.rectangle(BBOX2, fill="green", outline="yellow", width=1)
```

but

```python
draw.rectangle(BBOX2, fill="green", outline="yellow", width=0)
```

does not.


hugovk:
Perhaps we should mention the default change in the release notes and docstring? 

Or maybe it's not needed because the functionality has been fixed?

radarhere:
...I'm inclined to suggest no?

Documenting that we have changed the default argument would make you presume that we changed the default behaviour, which we haven't. `draw.rectangle(BBOX2, fill="green", outline="yellow")` still has the same output before and after this PR. So that might actually introduce confusion.

Issue #4213, opened 2019-11-19
Big-endian test failures on IBM Z


hugovk:
### What did you do?

Run tests on IBM Z (`s390x`) on Travis CI, newly available:

* https://blog.travis-ci.com/2019-11-12-multi-cpu-architecture-ibm-power-ibm-z

Like this:

* https://github.com/python-pillow/Pillow/compare/master...hugovk:travis-arch

(The other new architectures, `arm64` and `ppc64le`, both passed when skipping the `pyqt5` install.)

### What did you expect to happen?

Tests pass.

### What actually happened?

7 tests fail.

* [x] `TestFileJpeg2k.test_16bit_monochrome_j2k_like_tiff`
* [x] `TestFileJpeg2k.test_16bit_monochrome_jp2_like_tiff`
* [x] `TestFilePng.test_sanity`
* [ ] `TestFileWebpAnimation.test_write_animation_L`
* [ ] `TestFileWebpAnimation.test_write_animation_RGB`
* [x] `TestImageGetExtrema.test_extrema`
* [x] `TestImageGetExtrema.test_true_16`

```
=================================== FAILURES ===================================
______________ TestFileJpeg2k.test_16bit_monochrome_j2k_like_tiff ______________
self = <Tests.test_file_jpeg2k.TestFileJpeg2k testMethod=test_16bit_monochrome_j2k_like_tiff>
    def test_16bit_monochrome_j2k_like_tiff(self):
    
        tiff_16bit = Image.open("Tests/images/16bit.cropped.tif")
        j2k = Image.open("Tests/images/16bit.cropped.j2k")
>       self.assert_image_similar(j2k, tiff_16bit, 1e-3)
Tests/test_file_jpeg2k.py:180: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
Tests/helper.py:153: in assert_image_similar
    raise e
Tests/helper.py:139: in assert_image_similar
    self.assertGreaterEqual(
E   AssertionError: 0.001 not greater than or equal to 254.937744140625 :  average pixel value difference 254.9377 > epsilon 0.0010
______________ TestFileJpeg2k.test_16bit_monochrome_jp2_like_tiff ______________
self = <Tests.test_file_jpeg2k.TestFileJpeg2k testMethod=test_16bit_monochrome_jp2_like_tiff>
[...]

PR #4336, opened 2020-01-05
Test on new Travis CPUs

hugovk:
Helps for #3606.

# New CPU architectures

Now we've moved a lot of testing to GitHub Actions, we should still use Travis CI for testing things which are unique to it, to get good coverage.

Travis has recently made available new CPU architectures, in addition to the existing `amd64`.

* In July, they added `arm64` (Arm support)
  https://blog.travis-ci.com/2019-10-07-multi-cpu-architecture-support

* In November, they added `ppc64le` (IBM Power) and `s390x` (IBM Z CPU)
https://blog.travis-ci.com/2019-11-12-multi-cpu-architecture-ibm-power-ibm-z

* `s390x` is particularly useful as it's big-endian, and we have a number of [big-endian issues](https://github.com/python-pillow/Pillow/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3ABig-endian)

# Changes proposed in this pull request

 * Run tests on the new `arm64`, `ppc64le` and `s390x` architectures

 * Skip install of pyqt5 because "ERROR: Could not find a version that satisfies the requirement pyqt5"

 * The known `s390x` test failures (https://github.com/python-pillow/Pillow/issues/4213) are skipped as expected failures when run on CI: we should still aim to fix them, but there's value in running the other tests

This draft* PR is to illustrate the new CPUs on each Python version. I don't intend we test them all on Python versions, just a single version for each CPU.

(\* Was meant to be a draft but I clicked the wrong button...)

# Cold cache

image

https://travis-ci.org/hugovk/Pillow/builds/632753494

# Warm cache

image

https://travis-ci.org/hugovk/Pillow/builds/632884754

Generally the cold cache build times are about the same for each Python version, with newer versions a bit slower than older ones, except 3.8 on `s390x` is many times slower. 

Warm cache times are closer, but let's optimise for cold caches, which is the case for PR branches.

# Suggestion

Avoid Python 3.8 for now, spread over other versions. For example:

* `s390x` on Python 3.5
* `arm64` on Python 3.6
* `ppc64le` on Python 3.7


wiredfool:
Oooh. Big Endian.

.travis.yml
```diff
@@ -6,16 +6,23 @@ notifications:
   irc: "chat.freenode.net#pil"
 
 # Run fast lint first to get fast feedback.
-# Run slow PyPy next, to give it a headstart and reduce waiting time.
-# Run latest 3.x next, to get quick compatibility results.
-# Then run the remainder.
+# Run slower CPUs next, to give them a headstart and reduce waiting time.
+# Then run the remainder, with fastest Docker jobs last.
```
radarhere: After #4335, there are no longer any Docker jobs.
```suggestion
# Then run the remainder.
```

Tests/helper.py
```diff
@@ -348,6 +349,10 @@ def on_ci():
     )
 
 
+def is_big_endian():
+    return platform.processor() == "s390x"
```
homm: There is a constant:
https://docs.python.org/3/library/sys.html#sys.byteorder

hugovk: Thanks, updated!

hugovk:
Also moved the scripts from `.travis/` into `.ci/` to because they're now not solely Travis CI, they're also used by GitHub Actions.

radarhere:
You mention in the initial description that you intended this as a draft PR. Is there still work to be done?

hugovk:
No, ready for review!

PR #4340, opened 2020-01-06
Word wrap LICENSE file

johnthagen:

 * Word wrap `LICENSE` file

When generating a LICENSE report using [`pip-licenses`](https://github.com/raimon49/pip-licenses), Pillow is one of the very few projects in the dependency graph that does not word wrap its LICENSE file. This causes a very large rightward drift when the license report is generated and shown to the user.

This PR simply wraps the LICENSE file lines to make it match most other LICENSE files and display better to a user reading a LICENSE report.


hugovk:
Thank you!

PR #4344, opened 2020-01-08
Renamed Jpeg2K variable to prevent masking Image reduce method

### LINKED ISSUE #4343: JPEG 2000 file cannot be Image.reduce() 'd
JPEG 2000 file cannot be Image.reduce() 'd

... because img.reduce becomes a integer 0 during loading somehow.

Can be repeated with

[this image](https://drive.google.com/open?id=1UbSkNdOuUCt50TTu7Zc4PnuP6f-5RkKK)

```
$ python -c 'import PIL;from PIL import Image; im = Image.open("pillow_bug.jp2"); print(PIL.__version__);print(im.reduce)'
7.1.0.dev0
0
```

radarhere:
Resolves #4343

The new Image `reduce` method does not work for Jpeg2K images, as [Jpeg2KImagePlugin has a `reduce` variable that masks it](https://github.com/python-pillow/Pillow/blob/ef4a0b2f4c9346db37140f102e80068abc280167/src/PIL/Jpeg2KImagePlugin.py#L179). The variable is not purely internal, but is actually tested behaviour -

https://github.com/python-pillow/Pillow/blob/ef4a0b2f4c9346db37140f102e80068abc280167/Tests/test_file_jpeg2k.py#L112-L116

This PR renames the `reduce` variable to `load_reduce` for the sake of moving forward, while also keeping the tested behaviour by checking if `reduce` is callable.

### DIFF
--- a/Tests/test_file_jpeg2k.py
+++ b/Tests/test_file_jpeg2k.py
@@ -142,6 +142,13 @@ def test_layers_type(tmp_path):
             test_card.save(outfile, quality_layers=quality_layers)
 
 
+def test_load_reduce():
+    with Image.open("Tests/images/test-card-lossless.jp2") as im:
+        im.load_reduce = 2
+        im.load()
+        assert im.size == (160, 120)
+
+
 def test_layers():
     out = BytesIO()
     test_card.save(out, "JPEG2000", quality_layers=[100, 50, 10], progression="LRCP")
--- a/Tests/test_image_reduce.py
+++ b/Tests/test_image_reduce.py
@@ -3,6 +3,8 @@ from PIL import Image, ImageMath, ImageMode
 
 from .helper import PillowTestCase, convert_to_comparable
 
+codecs = dir(Image.core)
+
 
 class TestImageReduce(PillowTestCase):
     # There are several internal implementations
@@ -236,3 +238,10 @@ class TestImageReduce(PillowTestCase):
         for factor in self.remarkable_factors:
             self.compare_reduce_with_reference(im, factor, 0, 0)
             self.compare_reduce_with_box(im, factor)
+
+    @pytest.mark.skipif(
+        "jpeg2k_decoder" not in codecs, reason="JPEG 2000 support not available"
+    )
+    def test_jpeg2k(self):
+        with Image.open("Tests/images/test-card-lossless.jp2") as im:
+            assert im.reduce(2).size == (320, 240)
--- a/src/PIL/Jpeg2KImagePlugin.py
+++ b/src/PIL/Jpeg2KImagePlugin.py
@@ -176,7 +176,7 @@ class Jpeg2KImageFile(ImageFile.ImageFile):
         if self.size is None or self.mode is None:
             raise SyntaxError("unable to determine size/mode")
 
-        self.reduce = 0
+        self.load_reduce = 0
         self.layers = 0
 
         fd = -1
@@ -200,13 +200,14 @@ class Jpeg2KImageFile(ImageFile.ImageFile):
                 "jpeg2k",
                 (0, 0) + self.size,
                 0,
-                (self.codec, self.reduce, self.layers, fd, length),
+                (self.codec, self.load_reduce, self.layers, fd, length),
             )
         ]
 
     def load(self):
-        if self.reduce:
-            power = 1 << self.reduce
+        reduce = self.reduce if not callable(self.reduce) else self.load_reduce
+        if reduce:
+            power = 1 << reduce
             adjust = power >> 1
             self._size = (
                 int((self.size[0] + adjust) / power),
@@ -216,7 +217,7 @@ class Jpeg2KImageFile(ImageFile.ImageFile):
         if self.tile:
             # Update the reduce and layers settings
             t = self.tile[0]
-            t3 = (t[3][0], self.reduce, self.layers, t[3][3], t[3][4])
+            t3 = (t[3][0], reduce, self.layers, t[3][3], t[3][4])
             self.tile = [(t[0], (0, 0) + self.size, t[2], t3)]
 
         return ImageFile.ImageFile.load(self)


Tests/test_file_jpeg2k.py
```diff
@@ -110,10 +110,11 @@ def test_prog_res_rt(self):
...
     def test_reduce(self):
-        with Image.open("Tests/images/test-card-lossless.jp2") as im:
-            im.reduce = 2
-            im.load()
-            self.assertEqual(im.size, (160, 120))
+        for var in ["reduce", "load_reduce"]:
+            with Image.open("Tests/images/test-card-lossless.jp2") as im:
+                setattr(im, var, 2)
+                im.load()
+                self.assertEqual(im.size, (160, 120))
```
hugovk: For readability, let's add a new test instead rather than cramming more in, for example:

```python
    def test_load_reduce(self):
        with Image.open("Tests/images/test-card-lossless.jp2") as im:
            im.load_reduce = 2
            im.load()
            self.assertEqual(im.size, (160, 120))
```

Somewhat related:
* https://testing.googleblog.com/2018/06/testing-on-toilet-keep-tests-focused.html
* https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html

radarhere: Okay, done

homm:
Oh my god (

How does this variable work? Shouldn't it be just `.draft()` method?

homm:
This property is related to `-r` option of opj_decompress: https://github.com/uclouvain/openjpeg/wiki/DocJ2KCodec#usage-1

It scales image in two times (exactly algorithm is unknown and is I see, it doesn't match JPEG DCT scaling or `.reduce()` method) for every reducing level.

```
$ opj_decompress -i ../imgs/radial.jp2 -o ../imgs/_out.r-2.opj.bmp -r 2
```

_out r-2 opj

The problem is in Pillow it not only scales the image, but it also crops it by the same times. So with `reduce = 2` you get 2048 / 4 / 4 = 128 pixels image instead of 2048 / 4 = 512.

```python
In [2]: im = Image.open('../imgs/radial.jp2') 
   ...: print(im) 
   ...: im.reduce = 2 
   ...: im.load() 
   ...: im.save('../imgs/_out.r-2.png') 
   ...:  

<PIL.Jpeg2KImagePlugin.Jpeg2KImageFile image mode=L size=2048x2048 at 0x7F3D58AABEF0>
```

_out r-2

Another problem is unlike console decoder, Pillow fails on images on some size.

```python
In [2]: from PIL import Image 
   ...: im = Image.open('../imgs/radial.rgb.jp2') 
   ...: print(im) 
   ...: im.reduce = 2 
   ...: im.load() 
   ...: im.save('../imgs/_out.r-2.png') 
   ...:  

<PIL.Jpeg2KImagePlugin.Jpeg2KImageFile image mode=RGBA size=2049x2047 at 0x7FB7867849B0>
OSError: broken data stream when reading image file
```

I've checked Pillow 2.7.0, it operated the same. So I'm considering this functionality as never worked and suggest to add deprecation warning on `Jpeg2KImageFile.scale` set and save the value in private property instead of renaming it.

In future, if someone wants to add this functionality, it should be added in `.draft()` method.

homm:
@al45tair This property was added by you at the very beginning. Can you confirm it's broken or I miss something?

al45tair:
@homm Honestly, I haven't looked at this code in ages, but from your description it does sound like it's broken. It should have done the same thing as the OpenJPEG command line; if it doesn't, it's wrong.

homm:
@al45tair do you want to try to fix this and reimplement as `.draft()` method?

al45tair:
I'm sorry, but I'm not working in this area any more and I don't really have the time to work on Pillow as well as everything else that's going on.

radarhere:
Regardless of any feelings about how well `.reduce` works for Jpeg2000 images, I think that the API should be restored for the sake of backwards compatibility. If people aren't sold on the idea of `load_reduce`, I'm happy to remove that from this PR.

homm:
I've created PR with suggestions from https://github.com/python-pillow/Pillow/pull/4344#issuecomment-572268018.

https://github.com/radarhere/Pillow/pull/5

homm:
The design decision is needed. In the current version, the very common core functionality is partially broken since the opened image could be in any format including jpeg2000:

```python
from PIL import Image
im = Image.open(image_path)
im.thumbnail((100, 100))
```

We can fix it by deprecating undocumented already partially broken `Jpeg2KImageFile.reduce` property. There is a small chance that someone uses it, so until the next big release we need to keep backward compatibility for this property. There is no problem to use the value which the user assigns to this property. The question is what we should return to the user.

### Option 1

This PR. Return the bound method until any value is set, otherwise, return given value.

```python
from PIL import Image
im = Image.open("Tests/images/16bit.cropped.j2k")
print(im.reduce)  # <bound method Image.reduce of ...>, backward incompatability
im.reduce = 5
print(im.reduce)  # 5, as expected
im.thumbnail((100, 100))  # TypeError since im.reduce is not an method
```

### Option 2

https://github.com/radarhere/Pillow/pull/5 Keep the value in private property, always return the method.

```python
from PIL import Image
im = Image.open("Tests/images/16bit.cropped.j2k")
print(im.reduce)  # <bound method Image.reduce of ...>, backward incompatability
im.reduce = 5
print(im.reduce)  # <bound method Image.reduce of ...>, backward incompatability
print(im._reduce)  # 5
im.thumbnail((100, 100))  # Ok, im.reduce is an method
```

So the truth is in both cases we introduce backward incompatibility but in my opinion, the second option is more predictable.

radarhere:
@homm looking into this, I want to point out that your Option 2 code isn't actually viable. Looks like any size value that causes it to [call `reduce`](https://github.com/python-pillow/Pillow/blob/95d0d1af96a12c92bc1ee69ac9c11ac11dbee3f9/src/PIL/Image.py#L1869) also causes `ValueError: box can't exceed original image size`.

That said, even if it is scenario where ultimate success is not possible, maybe Pillow still shouldn't throw a TypeError. So I've created PR #4474 as an alternative to this.

With that PR,

```python
from PIL import Image
im = Image.open("Tests/images/16bit.cropped.j2k")
print(im.reduce)  # <bound method Image.reduce of ...>, backward incompatability
im.reduce = 5
print(im.reduce)  # 5, as expected
im.thumbnail((100, 100))  # the reduce method is still called
```

homm:
> Looks like any size value that causes it to call reduce also causes ValueError: box can't exceed original image size.

This is due to another Jpeg2KImagePlugin bug. It reduces `self.size` on each `self.load()` call.

```python
In [12]: from PIL import Image 
    ...: im = Image.open("Tests/images/test-card-lossless.jp2") 
    ...: im.reduce = 1 
    ...: im
Out[12]: <PIL.Jpeg2KImagePlugin.Jpeg2KImageFile image mode=RGB size=640x480 at 0x7FF0B810AD90>

In [13]: im.load() 
    ...: im
Out[13]: <PIL.Jpeg2KImagePlugin.Jpeg2KImageFile image mode=RGB size=320x240 at 0x7FF0B810AD90>

In [14]: im.load() 
    ...: im
Out[14]: <PIL.Jpeg2KImagePlugin.Jpeg2KImageFile image mode=RGB size=160x120 at 0x7FF0B810AD90>

In [15]: im.load() 
    ...: im
Out[15]: <PIL.Jpeg2KImagePlugin.Jpeg2KImageFile image mode=RGB size=80x60 at 0x7FF0B810AD90>
```

radarhere:
Closed in favour of #4474

PR #4356, opened 2020-01-14
Updated release schedule

### LINKED ISSUE #4345: Align with CPython release
CPython is moving from an 18-month to an annual release cycle, with releases in October ([PEP 602](https://www.python.org/dev/peps/pep-0602/)). CPython 3.9.0 is currently scheduled for 2020-10-05 ([PEP 596](https://www.python.org/dev/peps/pep-0596/)).

To prevent double work of us making a 2020-10-01 release (eg. 7.2.0), and then a patch version (eg. 7.2.1) a few days later with new wheels, I'd like to suggest we make the October release when the new CPython is ready for us to support it.

---

As an example, let's look at the last CPython 3.8.0 release:

* 2019-10-01 Pillow 6.2.0 released (https://github.com/python-pillow/Pillow/issues/3968)
* 2019-10-14 Scheduled and actual release of CPython 3.8.0 ([PEP 569](https://www.python.org/dev/peps/pep-0569/))
* 2019-10-21 Dependencies in place and we released Pillow 6.2.1 (https://github.com/python-pillow/Pillow/issues/4110, https://github.com/python-pillow/Pillow/issues/4152)

Assuming no CPython slippage, and 7 days for our dependencies to be updated:

* 2020-10-05 CPython 3.9.0 scheduled release
* 2020-10-12 Pillow release?

Of course, CPython 3.9.0 could slip, and our dependencies might be ready sooner, or later, so we won't know for sure until things are ready.

---

Would some time later in October be acceptable? Or is it too long?

We could pick a cut-off date (eg. middle or last day of October), and if Py3.9 things aren't ready, we release a non-Py3.9 one, and make a second when with Py3.9 when we can.

Thoughts?

Ping @python-pillow/pillow-team and anyone else interested!

radarhere:
Resolves #4345, using the suggestion from https://github.com/python-pillow/Pillow/issues/4345#issuecomment-572229441

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -2,7 +2,7 @@
 
 ## Main Release
 
-Released on January 2nd, April 1st, July 1st and October 15th.
+Released quarterly on January 2nd, April 1st, July 1st and October 15th.
 
 * [ ] Open a release ticket e.g. https://github.com/python-pillow/Pillow/issues/3154
 * [ ] Develop and prepare release in `master` branch.
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -2,7 +2,7 @@
 
 ## Main Release
 
-Released quarterly on the first day of January, April, July, October.
+Released on January 2nd, April 1st, July 1st and October 15th.
 
 * [ ] Open a release ticket e.g. https://github.com/python-pillow/Pillow/issues/3154
 * [ ] Develop and prepare release in `master` branch.


RELEASING.md
```diff
@@ -2,7 +2,7 @@
 
 ## Main Release
 
-Released quarterly on the first day of January, April, July, October.
+Released on January 2nd, April 1st, July 1st and October 15th.
```
hugovk: I'd suggest to keep "quarterly", as even if not bang on, it is still four times per year, and in the same months.

```suggestion
Released quarterly on January 2nd, April 1st, July 1st and October 15th.
```

hugovk:
Wording looks good, thanks. Let's wait a bit before merging in case there's more feedback on https://github.com/python-pillow/Pillow/issues/4345.

hugovk:
Thanks!

PR #4357, opened 2020-01-14
Various test changes

radarhere:
Some commits simplifying test code, some commits adding new tests.

I think the only commits that need commentary are -
- Removed print: Removes a `print` statement that doesn't look necessary or helpful
- Removed unnecessary code: Removes code if `__version__` contains 'rc'. Pillow doesn't use 'rc' versions.

### DIFF
--- a/Tests/test_image_resample.py
+++ b/Tests/test_image_resample.py
@@ -502,13 +502,9 @@ class CoreResampleBoxTest(PillowTestCase):
             ((40, 50), (10, 0, 50, 50)),
             ((40, 50), (10, 20, 50, 70)),
         ]:
-            try:
-                res = im.resize(size, Image.LANCZOS, box)
-                self.assertEqual(res.size, size)
-                self.assert_image_equal(res, im.crop(box))
-            except AssertionError:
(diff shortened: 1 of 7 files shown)


wiredfool:
I’ve done rc versions in the past when I didn’t want to fill up with a .0.1 in a week due to breakage at the las minute.

radarhere:
Okay. Thanks for feedback, I've removed that commit.

Tests/test_image_resample.py
```diff
@@ -502,13 +502,9 @@ def test_passthrough(self):
...
-            try:
-                res = im.resize(size, Image.LANCZOS, box)
-                self.assertEqual(res.size, size)
-                self.assert_image_equal(res, im.crop(box))
-            except AssertionError:
-                print(">>>", size, box)
-                raise
+            res = im.resize(size, Image.LANCZOS, box)
+            self.assertEqual(res.size, size)
+            self.assert_image_equal(res, im.crop(box), ">>> {} {}".format(size, box))
```
homm: 👍

PR #1574, opened 2015-12-06
Consistent multiline spacing and tests, with extra space

hugovk:
This fixes the inconsistent multiline spacing reported in https://github.com/python-pillow/Pillow/issues/1540.

It includes the fix from @wiredfool and I updated the tests. I checked the baseline-to-baseline heights in the new images and they're equal.

In PR https://github.com/python-pillow/Pillow/pull/1573 I noticed the lines are a bit close together by default. So I set the spacing to four pixels by default instead of zero.

---

[Here's the PR](https://github.com/python-pillow/Pillow/pull/1573) without extra spacing by default, and so either that or this should be merged, not both.

PR #4368, opened 2020-01-17
Updated spacing defaults documentation

### LINKED ISSUE #4367: Documentation Error: Default Spacing in ImageDraw.text()
I'm not 100% sure this is the right place for this, as it's not a bug in PIL but an error in the online documentation. The headings for [ImageDraw.text()](https://pillow.readthedocs.io/en/stable/reference/ImageDraw.html#PIL.ImageDraw.PIL.ImageDraw.ImageDraw.text) and [ImageDraw.multiline_text()](https://pillow.readthedocs.io/en/stable/reference/ImageDraw.html#PIL.ImageDraw.PIL.ImageDraw.ImageDraw.multiline_text) both list the default spacing value as 0. But the actual default value is 4, which matches the listed defaults for [ImageDraw.textsize()](https://pillow.readthedocs.io/en/stable/reference/ImageDraw.html#PIL.ImageDraw.PIL.ImageDraw.ImageDraw.textsize) and [ImageDraw.multiline_textsize()](https://pillow.readthedocs.io/en/stable/reference/ImageDraw.html#PIL.ImageDraw.PIL.ImageDraw.ImageDraw.multiline_textsize) and can be verified with the following code:

```python
img = Image.new('RGB', (200, 200), 'white')
draw = ImageDraw.Draw(img)
text = 'This is sample text.\nIt has a newline in it.'
draw.text((0,0), text, fill='black')
draw.text((0,0), text, fill='black', spacing=4)
img.show()
```

The two copies of the text overlap each other perfectly; if the default value were not four, the lower line would be offset from its previous copy.

radarhere:
Resolves #4367 

Updates the defaults for `spacing` in the documentation, so that [`text`](https://github.com/python-pillow/Pillow/blame/master/docs/reference/ImageDraw.rst#L258) and [`multiline_text`](https://github.com/python-pillow/Pillow/blame/master/docs/reference/ImageDraw.rst#L309)
matches https://github.com/python-pillow/Pillow/blob/a520a435d6b6534a3db2c9457328b6a90733a793/src/PIL/ImageDraw.py#L263-L270 and https://github.com/python-pillow/Pillow/blob/a520a435d6b6534a3db2c9457328b6a90733a793/src/PIL/ImageDraw.py#L353-L360

For reference, itooks like #1574 just forgot to update the documentation.

### DIFF
--- a/docs/reference/ImageDraw.rst
+++ b/docs/reference/ImageDraw.rst
@@ -255,7 +255,7 @@ Methods
 
     Draw a shape.
 
-.. py:method:: PIL.ImageDraw.ImageDraw.text(xy, text, fill=None, font=None, anchor=None, spacing=0, align="left", direction=None, features=None, language=None, stroke_width=0, stroke_fill=None)
+.. py:method:: PIL.ImageDraw.ImageDraw.text(xy, text, fill=None, font=None, anchor=None, spacing=4, align="left", direction=None, features=None, language=None, stroke_width=0, stroke_fill=None)
 
     Draws the string at the given position.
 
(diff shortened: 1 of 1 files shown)


hugovk:
Thanks!

Issue #4193, opened 2019-11-03
Move towards pytest style tests


jdufresne:
Is the project interested in moving to use pytest conventions? This would mean something of the lines:

- Use `assert` instead of `unittest` assert methods
- Removing `unittest` classes
- Using fixtures for stdout/stderr capturing, temporary files, and more (and perhaps some internal ones too.)
- Using parameterized tests?

If so, I can help out here and there. As there are quite a lot of tests, it might make sense to this on a file by file basis or in pieces.

Thoughts?

mfonism:
I'm interested in helping out with this.

hugovk:
Well, no one has objected yet!

I had originally wanted to keep things fairly agnostic with regards third-party tools, to stick with the stdlib `unittest` for the tests.

But that was when there was nose, nose2 and pytest, in addition to unittest. Now:

* [Nose](https://github.com/nose-devs/nose) is in maintenance mode with no updates for 4 years (but still 3m downloads per month)
* [Nose2](https://github.com/nose-devs/nose2) (165k/m) is still going but mainly maintained for those "stuck" using it and they encourage people new to testing to try pytest
* [Pytest](https://codecov.io/gh/pytest-dev/pytest) (12m/m) is very popular and well-maintained, and arguably becoming a standard for unit tests

So I'd be up for giving it a go.

I'd suggest to try and keep the main usage of it fairly simple, so there's a low barrier for new contributors looking at tests. Although parameterised tests could certainly be used in a few cases, it's a good way to remove duplication. And if any other pytest things are really helpful, then let's have a look at that too.

> If so, I can help out here and there. As there are quite a lot of tests, it might make sense to this on a file by file basis or in pieces.

I agree, it definitely makes sense to do it in parts rather than one big bang.

Would you be able to do a smallish PR so we can see how it looks?

Thank you!

jdufresne:
A first pass as the easier tests to port: #4369

radarhere:
unittest is no longer in Pillow. Long live pytest

Is this issue resolved?

jdufresne:
Makes sense to me. Nice work everyone.

PR #4369, opened 2020-01-18

jdufresne:
To better follow conventional pytest style, this removes the outer
wrapper class in favor of a function for some tests. These tests were
picked as they are relatively simple and presented no barriers to a
quick port. The assert* methods are replaced with assert statements.
When necessary, a fixture is used to create a temporary directory.

This commit does not convert the entire test suite to this style as some
test classes use methods or other advanced features that are difficult
to automatically convert. The goal is to address these issues in
followup commits.

Refs #4193

---

To reviewers, I'm happy expand or shrink the scope of this change as desired. If there is a preference for this to be split up over multiple commits, let me know. I'm happy to make adjustments.

### DIFF
--- a/Tests/test_file_bufrstub.py
+++ b/Tests/test_file_bufrstub.py
@@ -1,42 +1,46 @@
+import pytest
 from PIL import BufrStubImagePlugin, Image
 
-from .helper import PillowTestCase, hopper
+from .helper import hopper
 
 TEST_FILE = "Tests/images/gfs.t06z.rassda.tm00.bufr_d"
 
 
-class TestFileBufrStub(PillowTestCase):
-    def test_open(self):
-        # Act
-        with Image.open(TEST_FILE) as im:
+def test_open():
+    # Act
+    with Image.open(TEST_FILE) as im:
 
-            # Assert
-            self.assertEqual(im.format, "BUFR")
+        # Assert
+        assert im.format == "BUFR"
 
-            # Dummy data from the stub
-            self.assertEqual(im.mode, "F")
-            self.assertEqual(im.size, (1, 1))
+        # Dummy data from the stub
+        assert im.mode == "F"
+        assert im.size == (1, 1)
 
-    def test_invalid_file(self):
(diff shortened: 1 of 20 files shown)


Tests/test_file_bufrstub.py
```diff
@@ -1,42 +1,46 @@
...
+    with Image.open(TEST_FILE) as im:
 
-        # Act / Assert: stub cannot save without an implemented handler
-        self.assertRaises(IOError, im.save, tmpfile)
+        # Act / Assert: stub cannot load without an implemented handler
+        with pytest.raises(IOError):
+            im.load()
+
+
+def test_save(tmp_path):
```
hugovk: Noting something new to me, this looks handy:

> The `tmp_path` fixture

https://docs.pytest.org/en/latest/tmpdir.html

hugovk:
I'm fine with this. Any comments from anyone else?

---

I tried timing running the changed tests in this commit (using pytest) vs. the previous commit (using unittest), and the numbers were more or less inconclusive.

* 3m49s pytest
* 3m58s unittest
* 3m41s unittest
* 3m47s unittest
* 4m04s pytest
* 4m05s pytest
* 4m05s unittest
* 4m56s unittest

```sh
for i in {1..100}; do pytest Tests/test_000_sanity.py Tests/test_binary.py Tests/test_box_blur.py Tests/test_core_resources.py Tests/test_file_bufrstub.py Tests/test_file_container.py Tests/test_file_gd.py Tests/test_file_gimppalette.py Tests/test_file_wal.py Tests/test_format_lab.py Tests/test_image_getbands.py Tests/test_image_getbbox.py Tests/test_image_getim.py Tests/test_image_getpalette.py Tests/test_image_putpalette.py Tests/test_image_tobytes.py Tests/test_imagestat.py Tests/test_locale.py Tests/test_main.py Tests/test_util.py; done
```

hugovk:
Alright, let's go ahead with this. Thanks!

PR #4370, opened 2020-01-19
Handle multiple and malformed JPEG APP13 markers

homm:
This is a regression compared to Pillow 5.3, where the APP13 marker was not analyzing and images were opening without errors.

I met JPEGs with two APP13 markers were the first APP13 is always 65504 bytes length and looks like the second is just a continuation.

One of them: https://user-images.githubusercontent.com/128982/72672639-0a30b800-3a6e-11ea-88c2-d2b8fa8ca623.jpeg

This fix does not join markers since I don't know is it applicable for all files. Instead, it does not fail on the reading errors and collects data from all successfully processed markers.


radarhere:
Is the image that you've linked to one that can be included as a test image?

homm:
Probably no. I'll prepare another image with the same chunks.

homm:
@radarhere Test added

src/PIL/JpegImagePlugin.py
```diff
@@ -132,10 +127,11 @@ def APP(self, marker):
...
                     }
                 photoshop[code] = data
-                offset = offset + size
-                if offset & 1:
-                    offset += 1
-            self.info["photoshop"] = photoshop
+                offset += size
+                offset += offset & 1  # align
+            except struct.error:
+                break  # no sufficient data
```
```suggestion
                break  # insufficient data
```

PR #4371, opened 2020-01-19
Fixed Python coverage on GitHub Actions

radarhere:
In GitHub Actions, test.yml and test-windows.yml are not reporting Python coverage. This PR fixes that.

For test.yml, the clearest evidence that something is wrong is that [macOS code is not covered](https://codecov.io/gh/python-pillow/Pillow/src/3aa5d56bae3b9cee79fa88ed6e5be7bb28fc96e4/src/PIL/ImageGrab.py#L30).

While [Travis has a line 'Generating coverage xml reports for Python'](https://travis-ci.org/python-pillow/Pillow/jobs/638762367#L5574), [GHA has 'Python coveragepy not found'](https://github.com/python-pillow/Pillow/runs/396564029#step:14:103).

Looking at codecov, [a way around the 'Python coveragepy not found' error is to provide a coverage.xml file](https://github.com/codecov/codecov-bash/blob/master/codecov#L1055-L1085).

So, in `after_success.sh`, we can replace `coverage report` with `coverage xml` to output the data to an XML file.

For test-windows.yml, it looks like essentially the 'After success' step was just omitted in #4084. Adding that, in combination with the first part of this PR, adds Python coverage.


hugovk:
Whilst we're here, Codecov no longer support `.codecov.yml`, only `codecov.yml`:


* https://docs.codecov.io/docs/codecov-yaml#section-can-i-name-the-file-codecov-yml

They used to support both:

* https://web.archive.org/web/20190430062713/https://docs.codecov.io/docs/codecov-yaml#section-can-i-name-the-file-codecov-yml-

Shall we switch back in this PR or another?

radarhere:
Ok, I've added a commit to switch the filename.

nulano:
Should [the shell be specified](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#using-a-specific-shell) for Windows? The default is powershell, I'm guessing PowerShell opened the `.ci/after_success.sh` script in bash based on the filename, but I'm not sure I would rely on this behavior to work indefinitely.

It looks like there is still no [Python coverage from the Docker jobs on GHA](https://codecov.io/gh/python-pillow/Pillow/src/6101a5d7e8bc74d33376da173ef13cc65ad12a10/src/PIL/Image.py), but [there is C coverage](https://codecov.io/gh/python-pillow/Pillow/src/6101a5d7e8bc74d33376da173ef13cc65ad12a10/src/_imaging.c) (You have to unselect all, then select e.g. GHA>amazon-1-amd64; I couldn't find a direct link). Looking back, it was already missing when #4316 was merged. It looks like it is [failing to find the Python source code](https://github.com/python-pillow/Pillow/pull/4371/checks?check_run_id=397606806#step:6:23).

hugovk:
Windows shell: sounds wise to specify the shell rather than relying on the default. GHA has been a bit of a moving target.

---

Docker: interesting, let's deal with that separately, please can you open a new issue?

Confirmed in this Docker-only build:

* https://github.com/hugovk/Pillow/commit/f75c33940fe394d97445b8b67dcf9f43fc62adf4/checks?check_suite_id=408797442
* https://codecov.io/gh/hugovk/Pillow/commit/f75c33940fe394d97445b8b67dcf9f43fc62adf4/build
* https://codecov.io/gh/hugovk/Pillow/tree/f75c33940fe394d97445b8b67dcf9f43fc62adf4

(At the "Docker build" step, it reports coverage on Python files like `/vpy3/lib/python3.8/site-packages/Pillow-7.1.0.dev0-py3.8-linux-x86_64.egg/PIL/BdfFontFile.py`, perhaps it's the path causing issues. Or perhaps generating `xml` would help there too.)

And that's a good tip about unselecting, thanks.

hugovk:
Codecov bot commenting means it didn't read `comment: off` in `[.]codecov.yml`. Maybe due to it being in transition in this PR. Let's hope it'll be fine when merged.

radarhere:
Ok, I've added a commit to specify the shell for Windows.

radarhere:
I've added a commit to fix Python coverage on Docker GHA. It can't find the Python files, so I'm moving them into place. However, it tries to look for PIL/_imaging.py, so I added `--ignore-errors` for the Docker jobs to work around it.

hugovk:
Thank you!

PR #4373, opened 2020-01-20
get_display_profile is only supported on Windows

radarhere:
If not on Windows, the ImageCms `get_display_profile` function checks for `_imagingcms.get_display_profile`. That method does not exist. It wasn't in PIL 1.1.17, and hasn't existed since the fork. So this PR cleans up the code, returning quickly on non-Windows platforms, rather than performing a failing check.

### DIFF
--- a/src/PIL/ImageCms.py
+++ b/src/PIL/ImageCms.py
@@ -254,20 +254,15 @@ def get_display_profile(handle=None):
     :returns: None if the profile is not known.
     """
 
-    if sys.platform == "win32":
-        from PIL import ImageWin
+    if sys.platform != "win32":
+        return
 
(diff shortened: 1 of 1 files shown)


src/PIL/ImageCms.py
```diff
@@ -254,20 +254,15 @@ def get_display_profile(handle=None):
     :returns: None if the profile is not known.
     """
 
-    if sys.platform == "win32":
-        from PIL import ImageWin
+    if sys.platform != "win32":
+        return
```
hugovk:
```suggestion
        return None
```
Let's explicitly return None, as this is a function that returns something.

PR #4302, opened 2019-12-26
Raise ValueError for io.StringIO in Image.open

radarhere:
Resolves #4097 

The conclusion of discussion in the issue is that StringIO can't be used to pass bytes into Image.open.

```python
import io
from PIL import Image
im = Image.open(io.StringIO("'\x89\x50\x4E\x47\x0D"))
```
```
OSError: cannot identify image file <_io.StringIO object at 0x104ecdd70>
```

A ValueError with a clearer message with now be raised in this situation, explaining that byte data is required instead of StringIO.

PR #4376, opened 2020-01-21
Use BytesIO instead of StringIO

radarhere:
Updates the tutorial in light of #4302, and in light of the fact that `StringIO.StringIO` doesn't exist in Python 3.

### DIFF
--- a/docs/handbook/tutorial.rst
+++ b/docs/handbook/tutorial.rst
@@ -464,17 +464,17 @@ Reading from an open file
     with open("hopper.ppm", "rb") as fp:
         im = Image.open(fp)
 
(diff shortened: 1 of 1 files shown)


hugovk:
The failures are an unrelated problem in the newly-released pip 20.0.0: https://github.com/pypa/pip/issues/7620

I expect there'll be a pip 20.0.1 soonish, shall we do something like this in the meantime?

```
python -m pip install --upgrade "pip!=20.0.0"
```

hugovk:
And we don't need the workaround for this docs PR, and can possibly wait a bit for a new pip release before needing it. Merging, thanks!

PR #4378, opened 2020-01-24
Add flags to Codecov for easier coverage checking/debugging

hugovk:

 * Add flags for each upload of coverage to Codecov
   * flags must match pattern `^[\w\,]+$`
 * This allows us to easily filter by group (see top-right corner), to see which groups cover different bits of code, or if one is missing
 * For example, we can see there's no GHA Windows coverage, and that only Travis CI has Qt coverage
   * https://codecov.io/gh/hugovk/Pillow/src/b55e5b6d5f8c3bdf0c24f7cf06140ef3104ff569/Tests/test_imageqt.py

image

cc @nulano @radarhere

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -81,7 +81,7 @@ test_script:
 
 after_test:
 - pip install codecov
-- codecov --file coverage.xml --name %PYTHON%
+- codecov --file coverage.xml --name %PYTHON% --flags AppVeyor
 
 matrix:
   fast_finish: true
--- a/.ci/after_success.sh
+++ b/.ci/after_success.sh
@@ -9,7 +9,7 @@ else
 fi
 
 if [[ $TRAVIS ]]; then
-    codecov
+    codecov --flags TravisCI
 fi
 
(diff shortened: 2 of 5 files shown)


nulano:
I think this could be really useful. I just have two suggestions:
* I would consider renaming the flag GHADocker to GHA_Docker (if that is possible), to make it easier to distinguish the CI and the platform. (The [docs](https://docs.codecov.io/docs/flags#section-flag-creation) suggest the requirement is `^[a-z0-9_]{1,45}$`)
* Would it make sense to split GHA into GHA_MacOS and GHA_Ubuntu (perhaps with the [`matrix.include` option](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#example-including-configurations-in-a-matrix-build))?

hugovk:
Good ideas, updated!

Build still in progress but shows:

image

https://codecov.io/gh/python-pillow/Pillow/src/070626d10ea9ecb4a26dc3d5d6336df6bde6b5aa/Tests/test_imageqt.py

PR #4384, opened 2020-01-26
Merge pull request #4384 from jdufresne/float

jdufresne:
In Python 3, the division operator is floating point division. No longer
need to coerce integers to floating point numbers before division.


hugovk:
Thanks!

PR #4387, opened 2020-01-26
Remove Azure Pipelines in favour of GitHub Actions

hugovk:
Now we're using GitHub Actions, shall we remove the Azure Pipelines tests?

AP came along first, what we have was a first pass to quickly set it up and test with the Docker images. GHA came later and we've got a pretty wide setup on GHA now.

For removing:

 * They're running Docker tests. We're running the same tests in GHA, but GHA has coverage.
 * Less to maintain. It's easier to maintain fewer systems, less knowledge needed about the ins and outs of each
 * AP and GHA are both based on the same Microsoft base, no difference in that regard. But AP is a bit complicated to set up for new contributors, I think GHA is basically automatically enabled
 * GHA feels easier to work with (or perhaps just more familiar to me)
 * Cleaner PRs, ~14 fewer checks
 * It was only doing the Docker tests
 * Greener and cheaper, uses less server electricity and time
 * Frees up capacity for others

Against removing:

 * More eggs in more baskets. If one goes down, the other is still running. (But we'd probably wouldn't merge unless we had basic testing green.)
 * Anything else?

---

From [another PR](https://github.com/python-pillow/Pillow/pull/4386):

image


radarhere:
Linking to #3606, where Azure Pipelines were discussed

PR #4389, opened 2020-01-26
Remove unnecessary make_bytes() function

jdufresne:
The function was introduced in 65112bad7e4a692ea01980a642e540ec4c0a2fcb
to handle the differences between Python 2 & 3 byte handling. Now that
Python 3 supports byte formatting, can drop the unnecessary
compatibility shim in favor of native features.


hugovk:
Thank you!

PR #4390, opened 2020-01-27
Fix ResourceWarning emitted during tests

jdufresne:
Appeared in the form:

    ResourceWarning: unclosed file <_io.BufferedReader name='Tests/images/invalid-exif-without-x-resolution.jpg'>

Enable all warnings to always display during tests to help catch these
warnings earlier.

### DIFF
--- a/.ci/test.sh
+++ b/.ci/test.sh
@@ -2,7 +2,7 @@
 
 set -e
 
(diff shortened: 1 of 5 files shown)


hugovk:
Thanks!

PR #4393, opened 2020-01-27
Remove deprecated pytest-runner support

jdufresne:
The pytest-runner package is deprecated and its use is discouraged. See:

https://github.com/pytest-dev/pytest-runner/

> pytest-runner depends on deprecated features of setuptools and relies
> on features that break security mechanisms in pip. For example
> 'setup_requires' and 'tests_require' bypass pip --require-hashes. See
> also https://github.com/pypa/setuptools/issues/1684.
>
> It is recommended that you:
>
> - Remove 'pytest-runner' from your 'setup_requires', preferably
>   removing the setup_requires option.
> - Remove 'pytest' and any other testing requirements from
>   'tests_require', preferably removing the setup_requires option.
> - Select a tool to bootstrap and then run tests such as tox

Running tests should go through the tox entrypoint.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,8 +15,7 @@ co:
     done
 
 coverage:
-	python3 selftest.py
-	python3 setup.py test
+	pytest -qq
 	rm -r htmlcov || true
 	coverage report
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,3 @@
-[aliases]
-test=pytest
-
 [flake8]
 extend-ignore = E203, W503
 max-line-length = 88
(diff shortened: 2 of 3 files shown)


wiredfool:
I use `make test`, and expect that it will run the current test directory against the currently installed version of pillow. I'm not in favor of requiring tox to run tests.

jdufresne:
`make test` continues to work. Do you experience otherwise for this PR?

`tox` is not strictly required to run tests. Running them in directory with `pytest` or `make test` is still supported.

The only command that has been removed is support for `python3 setup.py test`, because that is deprecated upstream.

jdufresne:
I changed the Makefile `coverage` recipe to use the `pytest` command directly instead of `tox`. Let me know if this works better for you. The command was borrowed from the `test` target.

wiredfool:
That's fine.

PR #4401, opened 2020-02-01
Fixed stroke on FreeType < 2.9

### LINKED ISSUE #4366: Offset Text Stroke in Linux
First, thank you for creating such an awesome library!  Also, I LOVE the new text stroking function!  My workarounds to implement this in previous versions were vastly inferior to the new functionality you added!  Bravo!

However, I am seeing a slight issue with the stroking.  It appears to be offset when running on linux.  I've only tried one version of linux (outlined below).  The same code works as expected in windows (as illustrated below).

### What did you do?
I tried to add a text overlay with stroking to an image.
The text stroking would appear, and not be offset.
### What actually happened?
The text stroking was offset, making the stroke appear to be a shadow.  See this image generated from my stand alone sample:
image

However, when using windows, I do not see the issue.  Here is another overlay done with windows using the same program:

image

Here the outlines are crisp and not offset.

* OS: Raspbian GNU/Linux Version 9 (stretch)
* Python: 2.7 & 3.7 (I haven't tried any other versions)
* Pillow: 6.2.2

```

def create_overlay(font_path):
    image_color = (128,128,128,255)
    # Create an image with background color inverse to the text color.
    image = Image.new('RGB', (640, 480), color=image_color)
    text = "Sample Overlay Text"
    text_color_tuple = 255,255,255,255
    outline_color_tuple = 0,0,0,255
    font = ImageFont.truetype(font_path, size=48)

    # Create the image to draw on.
    text_image = Image.new('RGBA', image.size, (255, 255, 255, 0))
    d = ImageDraw.Draw(text_image)
    # overlay location
    x, y = 0, 0

    # Draw overlay text.
    d.multiline_text(
        xy=(x, y),
        text=text,
        fill=text_color_tuple,
        font=font,
        align='left',
        stroke_width=2,
        stroke_fill=outline_color_tuple

    )
    return Image.alpha_composite(image.convert('RGBA'), text_image).convert('RGB')

if __name__ == '__main__':
    image = create_overlay('/home/pi/fonts/DejaVu/DejaVuSansMono.ttf')
    image.save("/home/pi/temp.jpg")
```

Obviously you will need to adjust the font location and the save location to get this to work, but it shouldn't require much modification.  Please let me know if you need a better sample, or any additional information.

Thanks!

radarhere:
Resolves #4366 

The issue reported that the stroke was incorrectly positioned when writing the text 'Sample Overlay Text'. Testing for that issue, I found that it was only happening for FreeType < 2.9 - all three of the values being used in calculating the ascender [(`bitmap.rows`, `glyph_slot->bitmap_top` and `PIXEL(glyph_info[i].y_offset)`)](https://github.com/python-pillow/Pillow/blob/32bfbca3c8b5da265adc5eee6d537ce4957ac941/src/_imagingft.c#L803-L804) were zero. I find that always passing in `FT_LOAD_RENDER` when loading the glyphs for those calculations fixes that.

However, FreeType < 2.9 is a scenario in our test environments, so why wasn't this picked up? It's because 'Sample Overlay Text' has descenders - 'p' and 'y' go below the baseline of the text. The current stroke text tests involve 'A', 'A\nB' and 'ABC\nAaaa' - no descenders.


hugovk:
Thank you!

PR #4404, opened 2020-02-04
Fix size calculation of Image.thumbnail()

orlnub123:
The current size calculation is fine if only one axis is smaller or if the image is wide, but it trips up when both axes are smaller and the image is tall. Here's an example:
```py
from PIL import Image

im = Image.new("L", (145, 100))
im.thumbnail((50, 50))
assert im.size == (50, 34)  # This is fine, the image is wide.

im = Image.new("L", (100, 145))
im.thumbnail((100, 50))
assert im.size == (34, 50)  # Also fine, only one axis is smaller.

im = Image.new("L", (100, 145))
im.thumbnail((50, 50))
assert im.size == (34, 50)  # Oh no! It returned (35, 50)!
```
Someone made a [pull request](https://github.com/python-pillow/Pillow/pull/2089) that fixed this in 2016, but it looks like he closed it shortly after due to a CI failure.

### DIFF
--- a/Tests/test_image_thumbnail.py
+++ b/Tests/test_image_thumbnail.py
@@ -58,6 +58,10 @@ def test_aspect():
     im.thumbnail((50, 50))
     assert im.size == (34, 50)  # ratio is 0.68
 
+    im = Image.new("L", (100, 30))  # ratio is 3.333333333333
+    im.thumbnail((75, 75))
+    assert im.size == (75, 23)  # ratio is 3.260869565217
+
 
 def test_float():
     im = Image.new("L", (128, 128))
--- a/Tests/test_image_thumbnail.py
+++ b/Tests/test_image_thumbnail.py
@@ -50,6 +50,14 @@ def test_aspect():
     im.thumbnail((33, 33))
     assert im.size == (21, 33)  # ratio is 0.6363636364
 
+    im = Image.new("L", (145, 100))  # ratio is 1.45
+    im.thumbnail((50, 50))
+    assert im.size ==(50, 34)  # ratio is 1.47058823529
+
+    im = Image.new("L", (100, 145))  # ratio is 0.689655172414
+    im.thumbnail((50, 50))
+    assert im.size == (34, 50)  # ratio is 0.68
+
 
 def test_float():
     im = Image.new("L", (128, 128))
--- a/Tests/test_image_thumbnail.py
+++ b/Tests/test_image_thumbnail.py
@@ -34,9 +34,9 @@ def test_aspect():
     im.thumbnail((100, 100))
     assert im.size == (100, 50)
 
-    im = Image.new("L", (256, 128))
-    im.thumbnail((100, 50))
-    assert im.size == (100, 50)
+    im = Image.new("L", (64, 64))
+    im.thumbnail((100, 100))
+    assert im.size == (64, 64)
 
     im = Image.new("L", (128, 128))
     im.thumbnail((100, 100))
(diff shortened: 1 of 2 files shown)


radarhere:
I imagine that @homm, as the author of #4256, would like to review this?

Tests/test_image_thumbnail.py
```diff
@@ -31,9 +31,9 @@ def test_aspect(self):
         im.thumbnail((100, 50))
         self.assertEqual(im.size, (100, 50))
 
-        im = Image.new("L", (128, 128))
+        im = Image.new("L", (64, 64))
         im.thumbnail((100, 100))
-        self.assertEqual(im.size, (100, 100))
+        self.assertEqual(im.size, (64, 64))
```
radarhere: Why change these lines? Both versions of them pass with or without your source changes.

src/PIL/Image.py
```diff
@@ -2236,20 +2236,19 @@ def thumbnail(self, size, resample=BICUBIC, reducing_gap=2.0):
...
-            x = round(size[0])
-        if y > size[1]:
-            x = max(round(x * size[1] / y), 1)
-            y = round(size[1])
-        size = x, y
-        box = None
-
-        if size == self.size:
+        x, y = size
+        if x >= self.width and y >= self.height:
```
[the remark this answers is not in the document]
orlnub123: Since the aspect ratio is being preserved I don't see how one side being smaller would make the image bigger. This does bring up something that I hadn't thought about: what happens when you pass in a size that's only fractionally smaller and it gets rounded up to the image's size?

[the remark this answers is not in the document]
homm: This will be handled by the next `if self.size != size:` statement, wouldn't?

orlnub123:
These latest commits have a fix related to the issue previously discussed. Rounding up the size contradicts the documentation which states:
> This method modifies the image to contain a thumbnail version of itself, __no larger than the given size__.

So now when you request a thumbnail with a maximum size of `(99.9, 99.9)`, you get back an image with the size of `(99, 99)` instead of `(100, 100)`.

As far as I can tell passing float values to size is deprecated, so that commit can be reverted when the switch is made.

homm:
I have an interesting question. Let's say we have an image `100×30` pixels and requested `75×75` thumbnail. The result will be `75×22.5` which will be rounded to `75×22`. It seems that there should be no difference in which direction we're rounding, up or down, both options should be equal.

But the thing is, `100/30 = 3.3(3)`, `75/22 = 3.4(09)` and `75/23 ≈ 3.261` and this is a bit closer to `100/30`, but we still have an answer `75×22`. Why?

orlnub123:
I'd like to point out that the current implementation had the same issue. It happens because the rounding is indiscriminate towards the aspect ratio. I've added a commit that fixes the issue, improvements are welcome.

src/PIL/Image.py
```diff
@@ -2240,13 +2240,20 @@ def thumbnail(self, size, resample=BICUBIC, reducing_gap=2.0):
         if x >= self.width and y >= self.height:
             return
 
+        def round_aspect(number):
+            if x / y >= aspect:
+                key = lambda n: abs(aspect - n / y)  # noqa: E731
+            else:
+                key = lambda n: abs(aspect - x / n)  # noqa: E731
```
homm: You can avoid this repeated condition if round_aspect receives a key as an argument:

``python
if x / y >= aspect:
    x = round_aspect(y * aspect, lambda n: abs(aspect - n / y))
else:
    y = round_aspect(x / aspect, lambda n: abs(aspect - x / n))
```

orlnub123: That's a good suggestion, thanks!

hugovk:
Thanks!

PR #4419, opened 2020-02-13

hugovk:
For #4193.

Some notes:

# Setup and teardown

Class `setUp` becomes module `setup_module`:

```python
class TestFileTar(PillowTestCase):

    def setUp(self):
        if "zip_decoder" not in codecs and "jpeg_decoder" not in codecs:
            self.skipTest("neither jpeg nor zip support available")
```
to:
```python
def setup_module():
    if "zip_decoder" not in codecs and "jpeg_decoder" not in codecs:
        pytest.skip("neither jpeg nor zip support available")

```

https://docs.pytest.org/en/latest/xunit_setup.html#module-level-setup-teardown

# Skip whole class/module

Before: skip a unittest class:

```python
@unittest.skipUnless(olefile_installed, "olefile package not installed")
class TestFileFpx(PillowTestCase):
    def test_invalid_file(self):
```

After: skip a pytest module use `pytestmark`:

```python
pytestmark = pytest.mark.skipif(
    not olefile_installed, reason="olefile package not installed"
)


def test_invalid_file():
```

Or use a list for multiple skips. From:

```python
@unittest.skipUnless(olefile_installed, "olefile package not installed")
@unittest.skipUnless(features.check("libtiff"), "libtiff not installed")
class TestFileMic(PillowTestCase):
    def test_sanity(self):
```

to:

```python
pytestmark = [
    pytest.mark.skipif(not olefile_installed, reason="olefile package not installed"),
    pytest.mark.skipif(not features.check("libtiff"), reason="libtiff not installed"),
]


def test_sanity():
```

https://pytest.readthedocs.io/en/latest/skipping.html#skip-all-test-functions-of-a-class-or-module


Tests/test_file_dds.py
```diff
@@ -12,146 +14,148 @@
...
-            self.assertEqual(im.size, (16, 16))
-            self.assertEqual(im.info["gamma"], 1 / 2.2)
 
-            with Image.open(
-                TEST_FILE_DX10_BC7_UNORM_SRGB.replace(".dds", ".png")
-            ) as target:
-                assert_image_equal(target, im)
+def test_unimplemented_dxgi_format():
+    with pytest.raises(NotImplementedError):
+        Image.open("Tests/images/unimplemented_dxgi_format.dds",)
```
radarhere:
```suggestion
        Image.open("Tests/images/unimplemented_dxgi_format.dds")
```

Tests/test_file_dds.py
```diff
@@ -12,146 +14,148 @@
...
 
-    def test_unimplemented_pixel_format(self):
-        self.assertRaises(
-            NotImplementedError,
-            Image.open,
-            "Tests/images/unimplemented_pixel_format.dds",
-        )
+def test_unimplemented_pixel_format():
+    with pytest.raises(NotImplementedError):
+        Image.open("Tests/images/unimplemented_pixel_format.dds",)
```
radarhere:
```suggestion
        Image.open("Tests/images/unimplemented_pixel_format.dds")
```

Tests/test_file_gimpgradient.py
```diff
@@ -1,122 +1,128 @@
...
 
-    def test_load_via_imagepalette(self):
-        # Arrange
-        from PIL import ImagePalette
+    # Act
+    ret = GimpGradientFile.sphere_increasing(middle, pos)
 
-        test_file = "Tests/images/gimp_gradient.ggr"
+    # Assert
+    assert round(abs(ret - 0.9682458365518543), 7) == 0
```
radarhere: The table just above https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertAlmostEqual doesn't list `abs` as part of the `assertAlmostEqual` operation?

hugovk: For clarity:

```diff
-        self.assertAlmostEqual(ret, 0.9682458365518543)
+    assert round(abs(ret - 0.9682458365518543), 7) == 0

```

---

You're right, that table says `assertAlmostEqual` checks

```python
round(a-b, 7) == 0
```

but the stdlib does

```python
round(abs(second-first), places) == 0
```

https://github.com/python/cpython/blob/276eb67c29d05a93fbc22eea5470282e73700d20/Lib/unittest/case.py#L870

essentially

```python
round(abs(b-a), 7) == 0
```

It can be mathematically proved `abs(b-a) == abs(a-b)`, but I'm not really sure if the `abs` is strictly necessary. At least in this case:

```python
>>> round(0.000000009, 7)
0.0
>>> round(-0.000000009, 7)
-0.0
>>> round(0.000000009, 7) == 0
True
>>> round(-0.000000009, 7) == 0
True
```

radarhere:
This can be merged once https://github.com/hugovk/Pillow/pull/41 is resolved.

PR #4420, opened 2020-02-14
Simplify command discovery with stdlib shutil.which()

jdufresne:
Use the builtin shutil.which() instead of reimplementing.

For the single use that used the output of the command, use
subprocess.run().

### DIFF
--- a/Tests/helper.py
+++ b/Tests/helper.py
@@ -4,6 +4,7 @@ Helper functions.
 
 import logging
 import os
+import shutil
 import subprocess
 import sys
 import tempfile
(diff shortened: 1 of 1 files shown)


Tests/helper.py
```diff
@@ -191,9 +192,12 @@ def open_withImagemagick(self, f):
             raise OSError()
 
         outfile = self.tempfile("temp.png")
-        if command_succeeds([IMCONVERT, f, outfile]):
-            return Image.open(outfile)
-        raise OSError()
+        r = subprocess.run(
```
radarhere: Why change this from `call` to `run`? If all we're using is the `returncode`, then `call` would be simpler - https://docs.python.org/3/library/subprocess.html#subprocess.call

jdufresne: The top of the linked Python docs say:

> The recommended approach to invoking subprocesses is to use the `run()` function for all use cases it can handle. For more advanced use cases, the underlying `Popen` interface can be used directly.

That is the only reason.

I agree that `.call()` is simpler, I'll change it.

PR #4424, opened 2020-02-17
Fix saving IM images in dir with long path

hugovk:
For #4193.

Converting `Tests/test_file_im.py` to use pytest surfaced a bug in saving IM format files.

When roundtripping the file, it failed opening the saved file, at this check:

https://github.com/python-pillow/Pillow/blob/29fee8fc43d59dd65b7c533def44548aa5b9bc12/src/PIL/ImImagePlugin.py#L155-L156

It's falling over the `Name: ` field, which is saved like this:

https://github.com/python-pillow/Pillow/blob/29fee8fc43d59dd65b7c533def44548aa5b9bc12/src/PIL/ImImagePlugin.py#L349-L350

The difference is the temp path+filename under the limit with `PillowTestCase.tempfile`:

```
Name: /var/folders/kt/j77sf4_n6fnbx6pg199rbx700000gn/T/tempx72ebkj5.im
```

And over the limit with pytest's `tmp_path`:

```
Name: /private/var/folders/kt/j77sf4_n6fnbx6pg199rbx700000gn/T/pytest-of-hugo/pytest-126/test_roundtrip0/temp.im
```

This PR trims the path+filename to fit, and I suggest to keep the last part which includes the filename and extension, rather than cropping it off.

```
Name: olders/kt/j77sf4_n6fnbx6pg199rbx700000gn/T/pytest-of-hugo/pytest-126/test_roundtrip0/temp.im
```


hugovk:
This passes on Linux and macOS, but fails on Windows (but not PyPy / Win):

```
!!!!!!!!!!!!!!!!!!!!!!!!!! stopping after 1 failures !!!!!!!!!!!!!!!!!!!!!!!!!!
================== 1 failed, 253 passed, 8 skipped in 46.57s ==================
Command exited with code 1
fp = 'C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\pytest-of-appveyor\\pytest-0\\test_roundtrip0\\temp.im'
format = 'IM', params = {}
filename = 'C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\pytest-of-appveyor\\pytest-0\\test_roundtrip0\\temp.im'
open_fp = True, save_all = False, ext = '.im'
save_handler = <function _save at 0x09CDD6E8>
        """
        Saves this image under the given filename.  If no format is
        specified, the format to use is determined from the filename
        extension, if possible.
    
        Keyword options can be used to provide additional instructions
        to the writer. If a writer doesn't recognise an option, it is
        silently ignored. The available options are described in the
        :doc:`image format documentation
        <../handbook/image-file-formats>` for each writer.
    
        You can use a file object instead of a filename. In this case,
        you must always specify the format. The file object must
        implement the ``seek``, ``tell``, and ``write``
        methods, and be opened in binary mode.
    
        :param fp: A filename (string), pathlib.Path object or file object.
        :param format: Optional format override.  If omitted, the
           format to use is determined from the filename extension.
           If a file object was used instead of a filename, this
           parameter should always be used.
        :param params: Extra parameters to the image writer.
        :returns: None
        :exception ValueError: If the output format could not be determined
           from the file name.  Use the format option to solve this.
        :exception IOError: If the file could not be written.  The file
           may have been created, and may contain partial data.
        """
    
        filename = ""
        open_fp = False
        if isPath(fp):
            filename = fp
            open_fp = True
        elif isinstance(fp, Path):
            filename = str(fp)
            open_fp = True
        if not filename and hasattr(fp, "name") and isPath(fp.name):
            # only set the name for metadata purposes
            filename = fp.name
    
        # may mutate self!
        self._ensure_mutable()
    
        save_all = params.pop("save_all", False)
        self.encoderinfo = params
        self.encoderconfig = ()
    
        preinit()
    
        ext = os.path.splitext(filename)[1].lower()
    
        if not format:
            if ext not in EXTENSION:
                init()
            try:
                format = EXTENSION[ext]
            except KeyError:
                raise ValueError("unknown file extension: {}".format(ext))
    
        if format.upper() not in SAVE:
            init()
        if save_all:
            save_handler = SAVE_ALL[format.upper()]
        else:
            save_handler = SAVE[format.upper()]
    
        if open_fp:
            if params.get("append", False):
                # Open also for reading ("+"), because TIFF save_all
                # writer needs to go back and edit the written data.
                fp = builtins.open(filename, "r+b")
            else:
>               fp = builtins.open(filename, "w+b")
E               PermissionError: [Errno 13] Permission denied: 'C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\pytest-of-appveyor\\pytest-0\\test_roundtrip0\\temp.im'
C:\Python38\lib\site-packages\pillow-7.1.0.dev0-py3.8-win32.egg\PIL\Image.py:2103: PermissionError
----------- coverage: platform win32, python 3.8.0-final-0 -----------
```

I also included in this PR the converted `test_file_bmp.py`, which also uses the pytest `tmp_path` fixture for roundtripping with no problem. 

Any ideas?

radarhere:
My personal opinion, in the case of `/private/var/folders/kt/j77sf4_n6fnbx6pg199rbx700000gn/T/pytest-of-hugo/pytest-126/test_roundtrip0/temp.im`, I find

```
Name: olders/kt/j77sf4_n6fnbx6pg199rbx700000gn/T/pytest-of-hugo/pytest-126/test_roundtrip0/temp.im
```

to be slightly odd. I would feel better about just trimming it down to the base name in this situation, because I don't think I need to know the absolute path of an image's original location.

```
Name: temp.im
```

hugovk:
Sounds sensible, updated.

Of course, it's technically possible for a long basename to exceed the limit. This is less likely, do we fix it, or let it be until someone stumbles over it? There's an easy workaround.

I doubt the IM format gets much use, pretty much the only references I could find online for the format were in PIL/Pillow.

PR #1054, opened 2014-12-27
Look for OSX and Linux fonts in common places (#370 fixes)

charleslaw:
Fix #370

PR #4428, opened 2020-02-17
Replace SimplePatcher with builtin unittest.mock module

jdufresne:
The class more or less duplicates the features of the mock module. Can avoid the duplication by using the stdlib.


hugovk:
Thanks, that was added long ago.

https://github.com/python-pillow/Pillow/pull/1054

PR #4429, opened 2020-02-17
Remove testing of Python-2-only cPickle module

jdufresne:
In Python 3, cPickle and pickle have been merged to a single module.

From https://docs.python.org/3/whatsnew/3.0.html#library-changes

> A common pattern in Python 2.x is to have one version of a module
> implemented in pure Python, with an optional accelerated version
> implemented as a C extension; for example, pickle and cPickle. This
> places the burden of importing the accelerated version and falling
> back on the pure Python version on each user of these modules. In
> Python 3.0, the accelerated versions are considered implementation
> details of the pure Python versions. Users should always import the
> standard version, which attempts to import the accelerated version and
> falls back to the pure Python version. The pickle / cPickle pair
> received this treatment.

Can now move the 'import pickle' to the top of the test file.

### DIFF
--- a/Tests/test_pickle.py
+++ b/Tests/test_pickle.py
@@ -1,3 +1,5 @@
+import pickle
+
 from PIL import Image
 
 from .helper import PillowTestCase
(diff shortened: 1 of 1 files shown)


hugovk:
Thanks!

PR #4432, opened 2020-02-17
Remove unnecessary skip test logic

jdufresne:
The `gif_encoder` and `gif_decoder` are always compiled and included in `src/_imaging.c`.


jdufresne:
Moved to #4434.

PR #4434, opened 2020-02-18
Streamline test skipping based on supported features

jdufresne:
This adds a new test decorator: `skip_unless_feature()`. The argument is the same as passed to `features.check()`. If the feature is not supported, the test will be skipped.

This removes several kinds of boilerplate copied and pasted around tests so test feature checking is handled and displayed more consistently.

Refs #4193

### DIFF
--- a/Tests/check_j2k_leaks.py
+++ b/Tests/check_j2k_leaks.py
@@ -3,22 +3,18 @@ from io import BytesIO
 
 from PIL import Image
 
-from .helper import PillowTestCase, is_win32
+from .helper import PillowTestCase, is_win32, skip_unless_feature
 
 # Limits for testing the leak
 mem_limit = 1024 * 1048576
 stack_size = 8 * 1048576
 iterations = int((mem_limit / stack_size) * 2)
-codecs = dir(Image.core)
 test_file = "Tests/images/rgb_trns_ycbc.jp2"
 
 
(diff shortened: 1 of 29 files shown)


src/PIL/features.py
```diff
@@ -76,14 +76,13 @@ def get_supported_features():
...
-        or feature in features
-        and check_feature(feature)
-    )
+    if feature in modules:
+        return check_module(feature)
+    if feature in codecs:
+        return check_codec(feature)
+    if feature in features:
+        return check_feature(feature)
+    raise ValueError("Unknown feature '%s'." % feature)
```
radarhere: I can see why this could be helpful, but it is backwards incompatible.

jdufresne: What do you think about changing this to emit a warning? Then the behavior would be the same, but then library users (internal and external) can see that a programming mistake was made.

Or would you rather I drop the commit?

jdufresne: I've dropped the backwards incompatible aspect of this change. We can always revisit or reconsider at a later date and it is not directly necessary for the other benefits presented. 

Thanks for the feedback.

jdufresne: Moved this to  #4438 so it can be considered and reviewed in isolation

hugovk:
Thanks!

PR #4436, opened 2020-02-18
Use pytest.importorskip to skip on a missing import dependency

hugovk:
For #4193.

 * First commit: Use pytest.importorskip to skip on a missing import dependency
    * https://docs.pytest.org/en/latest/skipping.html#skipping-on-a-missing-import-dependency
 * Second commit: Also convert the touched files to use pytest

### DIFF
--- a/Tests/test_file_webp_metadata.py
+++ b/Tests/test_file_webp_metadata.py
@@ -3,119 +3,120 @@ from io import BytesIO
 import pytest
 from PIL import Image
 
-from .helper import PillowTestCase
-
 _webp = pytest.importorskip("PIL._webp", reason="WebP support not installed")
 
 
-class TestFileWebpMetadata(PillowTestCase):
-    def setUp(self):
-        if not _webp.HAVE_WEBPMUX:
-            self.skipTest("WebPMux support not installed")
+def setup_module():
+    if not _webp.HAVE_WEBPMUX:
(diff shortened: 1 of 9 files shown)


radarhere:
Just so you're aware, this is an alternative to #4434

hugovk:
This actually covers something different to #4434: that skips tests when features are not available, this skips tests when imports are not possible :)

hugovk:
Ah, there was a bit of overlap for a bit of the WebP checking. Will update this to keep those as feature checking.

Tests/test_file_webp_metadata.py
```diff
@@ -1,114 +1,121 @@
 from io import BytesIO
 
+import pytest
 from PIL import Image
 
-from .helper import PillowTestCase, skip_unless_feature
+from .helper import skip_unless_feature
 
+_webp = pytest.importorskip("PIL._webp", reason="WebP support not installed")
```
radarhere: `_webp` is not actually used in this file.
```suggestion
```

Tests/test_file_webp_metadata.py
```diff
@@ -1,114 +1,121 @@
 from io import BytesIO
 
+import pytest
```
radarhere:
```suggestion
```

Which might also make this redundant.

PR #4440, opened 2020-02-21
Allow saving of zero quality JPEG images

### LINKED ISSUE #4439: Can't produce a Quality 0 JPEG
### What did you do?

```
im.save(output_path / i.with_suffix('.jpg').name, quality=0)
```

A JPEG with libjpeg quantization equivalent to zero quality to be output.

### What actually happened?

A JPEG with libjpeg quantization equivalent to quality 75 was output

* OS: Linux
* Python: 3.5
* Pillow: 

###  Comments

I do research in image compression and as part of that research I wanted to produce some quality 0 JPEG files, the quantization matrix for that should filled with 255, dumping the resulting JPEGs quantization tables with `djpeg` I see that they are quality 75, and from the source code it looks like `quality=0` indicates to use the defaulty quality which is 75. 

Quality zero is a completely valid quality for libjpeg, but it doesn't work in Pillow. I'm not sure how many people care about quality 0 JPEGs so it might not be worth fixing but I still figured I'd leave this here.

radarhere:
Resolves #4439

Since PIL, a JPEG quality of 0 has been used to indicate default quality. This PR changes that indicating value from 0 to -1, so that JPEG images of actual 0 quality can be produced.

### DIFF
--- a/Tests/test_file_jpeg.py
+++ b/Tests/test_file_jpeg.py
@@ -299,6 +299,10 @@ class TestFileJpeg(PillowTestCase):
         assert_image(im1, im2.mode, im2.size)
         self.assertGreaterEqual(im1.bytes, im2.bytes)
 
+        im3 = self.roundtrip(hopper(), quality=0)
+        assert_image(im1, im3.mode, im3.size)
+        self.assertGreater(im2.bytes, im3.bytes)
+
     def test_smooth(self):
         im1 = self.roundtrip(hopper())
         im2 = self.roundtrip(hopper(), smooth=100)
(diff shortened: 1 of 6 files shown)


hugovk:
The docs need updating too, at least the `quality` parameter here:

https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html#jpeg

hugovk:
And let's mention it in the release notes too.

radarhere:
Good catch, thanks. I've updated the documentation, and added release notes.

hugovk:
Thanks!

PR #4442, opened 2020-02-21
Changed docs badges to be closer to README badges

radarhere:
See https://pillow.readthedocs.io/en/latest/ vs https://github.com/python-pillow/Pillow/blob/master/README.rst ... vs https://pillow-radarhere.readthedocs.io/en/docs_badges/


hugovk:
# Screenshots

image

vs.

image

vs.

image

PR #4443, opened 2020-02-22
Converted most assert statements to pytest

radarhere:
Helps #4193 with a different approach - rather than converting individual files completely, this PR changes single assert lines.

### DIFF
--- a/Tests/check_libtiff_segfault.py
+++ b/Tests/check_libtiff_segfault.py
@@ -1,5 +1,6 @@
 import unittest
 
+import pytest
 from PIL import Image
 
 from .helper import PillowTestCase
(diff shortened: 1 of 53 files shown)


hugovk:
And do the rest of the conversion in follow-up PRs? It's a bit more churny, but sure.

Perhaps omit `test_file_bmp.py` and `test_file_im.py` from here, as they have full conversion in https://github.com/python-pillow/Pillow/pull/4424?

radarhere:
I just thought it was an easier way to approach the repetitive task of conversion.

Okay, omitted those files.

hugovk:
I restarted the failed AppVeyor builds after a network error:
```
Build started
ECHO %PYTHON%
C:/Python38-x64
git clone -q https://github.com/python-pillow/Pillow.git c:\pillow
git fetch -q origin +refs/pull/4443/merge:
git checkout -qf FETCH_HEAD
Restoring build cache
Cache 'C:\Users\appveyor\AppData\Local\pip\Cache' - Restored
Running Install scripts
curl -fsSL -o pillow-depends.zip https://github.com/python-pillow/pillow-depends/archive/master.zip
curl: (18) transfer closed with outstanding read data remaining
Command exited with code 18
```
If it happens again, we can look to hardening the curl.

PR #4446, opened 2020-02-22

hugovk:
For #4193.

### DIFF
--- a/Tests/test_file_eps.py
+++ b/Tests/test_file_eps.py
@@ -216,7 +216,7 @@ def test_readline(tmp_path):
         _test_readline(t, ending)
 
     def _test_readline_file_psfile(test_string, ending):
-        f = str(tmp_path / "temp.bufr")
+        f = str(tmp_path / "temp.txt")
         with open(f, "wb") as w:
             w.write(test_string.encode("latin-1"))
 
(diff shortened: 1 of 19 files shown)


Tests/test_file_eps.py
```diff
@@ -193,53 +210,49 @@ def _test_readline(self, t, ending):
...
+    def _test_readline_io_psfile(test_string, ending):
         f = io.BytesIO(test_string.encode("latin-1"))
         t = EpsImagePlugin.PSFile(f)
-        self._test_readline(t, ending)
+        _test_readline(t, ending)
 
-    def _test_readline_file_psfile(self, test_string, ending):
-        f = self.tempfile("temp.txt")
+    def _test_readline_file_psfile(test_string, ending):
+        f = str(tmp_path / "temp.bufr")
```
radarhere:
```suggestion
        f = str(tmp_path / "temp.txt")
```
Is there any particular reason why you changed this to '.bufr'? I'm imagining that you just picked it as an abbreviation of 'buffer', but I'm hesistant considering that BUFR is a separate file format that we deal with.

hugovk: That was a copy and paste mistake, thanks!

radarhere:
Thanks!

PR #4454, opened 2020-02-28
If present, only use alpha channel for bounding box

### LINKED ISSUE #4453: [ bug ] getbbox()
Simply won't crop exactly that image, meanwhile with others works fine.

When i move image slightly to right in GIMP and save - it is works again, plain resaving with gimp png defaults - does not helps.

* OS: Arch Linux Linux 5.4.22-1-lts
* Python: 3.8.1
* Pillow: community/python-pillow 6.2.1-1

```python
	from PIL import Image
    png_path = './22-108-20179696.png'
    im = Image.open(png_path)
    im = im.crop(im.getbbox())
    im.save('./test.png')

```

radarhere:
Resolves #4453 

Changes `getbbox` for modes RGBA, RGBa, PA, LA and La.

Testing, I find that the current functionality is to trim black transparent pixels. This PR changes it to trim any transparent pixels, regardless of color.

### DIFF
--- a/Tests/test_image_getbbox.py
+++ b/Tests/test_image_getbbox.py
@@ -10,29 +10,32 @@ def test_sanity():
 
 
 def test_bbox():
+    def check(im, fill_color):
+        assert im.getbbox() is None
+
+        im.paste(fill_color, (10, 25, 90, 75))
+        assert im.getbbox() == (10, 25, 90, 75)
+
+        im.paste(fill_color, (25, 10, 75, 90))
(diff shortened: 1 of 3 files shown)


hugovk:
Thanks for the release notes. Shall we use headings like "API Additions", "API Changes", etc?

* https://pillow.readthedocs.io/en/stable/releasenotes/index.html

hugovk:
Thanks!

PR #4455, opened 2020-02-28
Add JPEG comment to info dictionary

### LINKED ISSUE #4450: PIL cannot read JPEG comment
### What did you do?
I want PIL to read the JPEG comment (marker: 0xFF 0xFE).  
  
I took an image with an attached JPEG comment - verified with exiftools & IrfanView to exist.

```python
from PIL import Image, JpegImagePlugin
pic = Image.open(<path_to_pic_with_JPEG_comment>)
print(pic.info)
```

Show the JPEG comment in the dict.

* OS: W7x64
* Python: Python 3.8.1 x64
* Pillow: Pillow           7.0.0

I cannot attach an image via github ("Something went really wrong, ..."), so here is the file (5.61 KiB) (I downloaded it and verified it's byte-identical to the uploaded one): [](https://postimg.cc/BLrFc0kf)

radarhere:
Resolves #4450 

Adds JPEG comment to the `info` dictionary, in the same way that [GIF](https://github.com/python-pillow/Pillow/blob/1c1ad65a9646f73d80ce8f77894a447b03f203a6/src/PIL/GifImagePlugin.py#L213) and [GBR](https://github.com/python-pillow/Pillow/blob/1c1ad65a9646f73d80ce8f77894a447b03f203a6/src/PIL/GbrImagePlugin.py#L78) do.

### DIFF
--- a/docs/handbook/image-file-formats.rst
+++ b/docs/handbook/image-file-formats.rst
@@ -298,6 +298,11 @@ The :py:meth:`~PIL.Image.Image.open` method may set the following
 **exif**
     Raw EXIF data from the image.
 
+**comment**
+    A comment about the image.
+
+    .. versionadded:: 7.1.0
+
 
 The :py:meth:`~PIL.Image.Image.save` method supports the following options:
 
--- a/Tests/test_file_jpeg.py
+++ b/Tests/test_file_jpeg.py
@@ -60,6 +60,8 @@ class TestFileJpeg:
             )
             assert len(im.applist) == 2
 
+            assert im.info["comment"] == b"File written by Adobe Photoshop\xa8 4.0\x00"
+
     def test_cmyk(self):
         # Test CMYK handling.  Thanks to Tim and Charlie for test data,
         # Michael for getting me to look one more time.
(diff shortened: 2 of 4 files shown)


docs/handbook/image-file-formats.rst
```diff
@@ -298,6 +298,9 @@ The :py:meth:`~PIL.Image.Image.open` method may set the following
 **exif**
     Raw EXIF data from the image.
 
+**comment**
+    A comment about the image.
```
hugovk: Should we have a "New in version 7.1.0."?

radarhere: Good call. Done

hugovk:
The MinGW download failures look like this:

> Work has started on restoring the server, but it might take a while.

https://github.com/msys2/MSYS2-packages/issues/1884#issuecomment-595427161

Possible workaround here: https://github.com/msys2/MINGW-packages/issues/6256#issuecomment-595930128, but we can probably wait a bit.

Anyway, unrelated to this PR, and it passed earlier, and only docs changes were made since if failed.

hugovk:
Looks like the repo.msys2.org server is back up:
https://github.com/msys2/MSYS2-packages/issues/1884

I restarted the failed job and it passed:
https://ci.appveyor.com/project/Python-pillow/pillow/build/job/i1dpyag9iv286cmj

PR #4457, opened 2020-02-29
Fixed endian handling for I;16 getextrema

radarhere:
Helps #4213

Using the endian handling in https://github.com/python-pillow/Pillow/blob/b73e3dddcc9e6b3039db366b202254f693ccd13d/src/libImaging/Unpack.c#L1042-L1046 as inspiration, fixes `getextrema` I;16 operations for big-endian.


hugovk:
Thanks!

PR #4462, opened 2020-03-02
Convert to use pytest

hugovk:
For #4193.

### DIFF
--- a/Tests/test_file_tiff.py
+++ b/Tests/test_file_tiff.py
@@ -1,6 +1,5 @@
 import logging
 import os
-import unittest
 from io import BytesIO
 
 import pytest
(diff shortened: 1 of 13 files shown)


Tests/test_file_tiff.py
```diff
@@ -512,13 +510,16 @@ def test_tiled_planar_raw(self):
...
-            outfile = self.tempfile("temp.tif")
+    def test_palette(self, tmp_path):
+        def roundtrip(mode):
+            outfile = str(tmp_path / "temp.tif")
 
             im = hopper(mode)
             im.save(outfile)
 
+            for mode in ["P", "PA"]:
+                roundtrip(mode)
```
radarhere: This isn't right - you'll figure it out in a second of looking, but just to be explicit - this code never calls `roundtrip`, except from within `roundtrip`.

hugovk: Thanks, fixed in https://github.com/python-pillow/Pillow/pull/4462/commits/fbd4ca1ba0a93fb62da10e9aafd7dc1a965da2b0.

radarhere:
Thanks

Issue #4394, opened 2020-01-28
Re-disable Codecov PR comments


hugovk:
It should have been disabled by #4386, and there was no comment in that PR, but it's still commenting new PRs (https://github.com/python-pillow/Pillow/pull/4391#issuecomment-579146785).

hugovk:
Asked at Codecov: https://community.codecov.io/t/prs-are-commented-even-with-comment-off/941?u=hugovk

hugovk:
Nearly two weeks and no reply in the Codecov support forum.

Another option is to disable it for all python-pillow repos via the "team yaml":

https://docs.codecov.io/docs/codecov-yaml#section-team-yaml

But I don't have access to https://codecov.io/account/gh/python-pillow/yaml

> Only team owners may adjust this file.

hugovk:
Actually, looks like it's fixed, the last few PRs haven't received bot comments.

radarhere:
Thanks for asking codecov. I guess we'll see how this goes.

radarhere:
#4423 seems to demonstrate that this is still an issue.

hugovk:
Codecov have a possible fix now, let's see if it happens again:

> I located the comment. I suspect what happened was Codecov detected that the PR already had a comment and thus ignored the the setting in the YAML. It tried to edit it, failed, and created a new one.
>
> We have a new notify task that we are slowly rolling out that probably handles this better. I have whitelisted you for this, please give us 15-30 mins for the pods to roll and then let me know if you continue to see this behavior.

https://community.codecov.io/t/prs-are-commented-even-with-comment-off/941/9?u=hugovk

hugovk:
> Actually, looks like it's fixed, the last few PRs haven't received bot comments.

And again!

PR #4473, opened 2020-03-13
Codecov now supports tokenless upload for GHA

hugovk:
Codecov now supports tokenless uploads for GitHub Actions.

* https://community.codecov.io/t/whitelist-github-action-servers-to-upload-without-a-token/491/17

We can remove the workaround. Perhaps this will also help https://github.com/python-pillow/Pillow/issues/4394.

---

~Also use v2 of `actions/checkout`~

* ~https://github.com/actions/checkout/releases/tag/v2.0.0~

(Skipped for now, it failed on Windows)


radarhere:
Thanks

PR #4474, opened 2020-03-14
Prevent masking Image reduce method in Jpeg2K

### LINKED ISSUE #4343: JPEG 2000 file cannot be Image.reduce() 'd

radarhere:
Resolves #4343. Alternative to #4344

The new Image `reduce` method does not work for Jpeg2K images, as [Jpeg2KImagePlugin has a `reduce` variable that masks it](https://github.com/python-pillow/Pillow/blob/ef4a0b2f4c9346db37140f102e80068abc280167/src/PIL/Jpeg2KImagePlugin.py#L179). The variable is not purely internal, but is actually tested behaviour -

https://github.com/python-pillow/Pillow/blob/ef4a0b2f4c9346db37140f102e80068abc280167/Tests/test_file_jpeg2k.py#L112-L116

PR #4344 tried to address this by treating the setting of `reduce` as an integer, and the use of `reduce` as a method, as two different scenarios that were not to interact.

So instead, this PR changes Pillow so that if you set `reduce` as an integer, the user will then see `reduce` as an integer.

```python
with Image.open("Tests/images/test-card-lossless.jp2") as im:
        assert callable(im.reduce)

        im.reduce = 2
        assert im.reduce == 2
```

At that point, you can't call the `reduce` method externally. If Pillow tries to call the `reduce` method internally though, it will ignore this new use of `reduce`, and still call the Pillow method.

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -1866,7 +1866,11 @@ class Image:
             factor_y = int((box[3] - box[1]) / size[1] / reducing_gap) or 1
             if factor_x > 1 or factor_y > 1:
                 reduce_box = self._get_safe_box(size, resample, box)
-                self = self.reduce((factor_x, factor_y), box=reduce_box)
+                factor = (factor_x, factor_y)
+                if callable(self.reduce):
+                    self = self.reduce(factor, box=reduce_box)
+                else:
+                    self = Image.reduce(self, factor, box=reduce_box)
                 box = (
                     (box[0] - reduce_box[0]) / factor_x,
                     (box[1] - reduce_box[1]) / factor_y,
--- a/src/PIL/Jpeg2KImagePlugin.py
+++ b/src/PIL/Jpeg2KImagePlugin.py
@@ -176,7 +176,7 @@ class Jpeg2KImageFile(ImageFile.ImageFile):
         if self.size is None or self.mode is None:
             raise SyntaxError("unable to determine size/mode")
 
-        self.reduce = 0
+        self._reduce = 0
         self.layers = 0
 
         fd = -1
@@ -200,13 +200,21 @@ class Jpeg2KImageFile(ImageFile.ImageFile):
                 "jpeg2k",
                 (0, 0) + self.size,
                 0,
-                (self.codec, self.reduce, self.layers, fd, length),
+                (self.codec, self._reduce, self.layers, fd, length),
             )
         ]
 
+    @property
+    def reduce(self):
+        return self._reduce or super().reduce
+
+    @reduce.setter
+    def reduce(self, value):
+        self._reduce = value
+
     def load(self):
-        if self.reduce:
-            power = 1 << self.reduce
+        if self._reduce:
+            power = 1 << self._reduce
             adjust = power >> 1
             self._size = (
                 int((self.size[0] + adjust) / power),
@@ -216,7 +224,7 @@ class Jpeg2KImageFile(ImageFile.ImageFile):
         if self.tile:
             # Update the reduce and layers settings
             t = self.tile[0]
-            t3 = (t[3][0], self.reduce, self.layers, t[3][3], t[3][4])
+            t3 = (t[3][0], self._reduce, self.layers, t[3][3], t[3][4])
             self.tile = [(t[0], (0, 0) + self.size, t[2], t3)]
 
         return ImageFile.ImageFile.load(self)
--- a/src/PIL/Jpeg2KImagePlugin.py
+++ b/src/PIL/Jpeg2KImagePlugin.py
@@ -216,7 +216,7 @@ class Jpeg2KImageFile(ImageFile.ImageFile):
         self._reduce = value
 
     def load(self):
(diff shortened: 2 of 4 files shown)


src/PIL/Image.py
```diff
@@ -1866,7 +1866,11 @@ def resize(self, size, resample=BICUBIC, box=None, reducing_gap=None):
             factor_y = int((box[3] - box[1]) / size[1] / reducing_gap) or 1
             if factor_x > 1 or factor_y > 1:
                 reduce_box = self._get_safe_box(size, resample, box)
-                self = self.reduce((factor_x, factor_y), box=reduce_box)
+                factor = (factor_x, factor_y)
+                if callable(self.reduce):
+                    self = self.reduce(factor, box=reduce_box)
+                else:
+                    self = Image.reduce(self, factor, box=reduce_box)
```
homm: Ah, cool!

src/PIL/Jpeg2KImagePlugin.py
```diff
@@ -200,13 +200,21 @@ def _open(self):
...
                 (0, 0) + self.size,
                 0,
-                (self.codec, self.reduce, self.layers, fd, length),
+                (self.codec, self._reduce, self.layers, fd, length),
             )
         ]
 
+    @property
+    def reduce(self):
+        return self._reduce or super().reduce
```
homm: Probably some comment with a link to the issue is required.

src/PIL/Jpeg2KImagePlugin.py
```diff
@@ -200,13 +200,21 @@ def _open(self):
...
+        self._reduce = value
+
     def load(self):
-        if self.reduce:
-            power = 1 << self.reduce
+        if self._reduce:
+            power = 1 << self._reduce
             adjust = power >> 1
             self._size = (
                 int((self.size[0] + adjust) / power),
```
homm: This should be moved inside `if self.tile:` clause to prevent this bug: https://github.com/python-pillow/Pillow/pull/4344#issuecomment-599743249

src/PIL/Jpeg2KImagePlugin.py
```diff
@@ -216,7 +224,7 @@ def load(self):
         if self.tile:
```
homm:
```suggestion

```

src/PIL/Jpeg2KImagePlugin.py
```diff
@@ -200,13 +200,21 @@ def _open(self):
...
+        return self._reduce or super().reduce
+
+    @reduce.setter
+    def reduce(self, value):
+        self._reduce = value
+
     def load(self):
-        if self.reduce:
-            power = 1 << self.reduce
+        if self._reduce:
```
homm:
```suggestion
        if self.tile and self._reduce:
```

src/PIL/Jpeg2KImagePlugin.py
radarhere: Okay, done

radarhere:
@homm so you approve of this PR then?

homm:
What about deprecation warnings?

homm:
> What about deprecation warnings?

homm:
> What about deprecation warnings?

Otherwise is ok.

radarhere:
Ok, I've added a deprecation warning pointing users to use `load_reduce` instead.

homm:
I believe, `load_reduce` is not the way how it should work. There is already `.draft()` method exactly for this functionality, which should be used.

radarhere:
Correct me if I'm wrong - there isn't a `draft` method for Jpeg2kImagePlugin. It only exists for JpegImagePlugin.

homm:
Draft method exists for [`PIL.Image.Image`](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.draft). It "configures the image file loader so it returns a version of the image that as closely as possible matches the given mode and size". This is exactly what `Jpeg2KImageFile.reduce` was erroneously used for.

To be clear: I don't mean you have to implement `Jpeg2KImageFile.draft()` method to fix this bug, I just strongly against adding extra API for things that already exist.

radarhere:
Sure, sure. I'm just confused about the idea of adding a deprecation warning without being able to tell them what they should be doing instead.

I've removed `load_reduce` and the deprecation warning. How do you feel about merging this in for 7.1.0, to resolve #4343 and then maybe we introduce a deprecation warning and a Jpeg2K method at a later date?

homm:
Ok, go ahead.

radarhere:
Thanks for reviewing

PR #4477, opened 2020-03-16
Updated CI target Python versions

radarhere:
Alpine now has Python 3.8 - https://github.com/python-pillow/Pillow/runs/506966784?check_suite_focus=true#step:5:594

Arch has now Python 3.8 - https://github.com/python-pillow/Pillow/runs/506966786?check_suite_focus=true#step:5:587

Amazon 2 now has Python 3.7 - https://github.com/python-pillow/Pillow/runs/506966867?check_suite_focus=true#step:5:596


hugovk:
Great!

PR #4486, opened 2020-03-28
Removed PillowTestCase helper class

radarhere:
Helps #4193

Converts all uses of PillowTestCase to pytest, so that PillowTestCase is no longer needed.

### DIFF
--- a/Tests/bench_cffi_access.py
+++ b/Tests/bench_cffi_access.py
@@ -1,9 +1,8 @@
 import time
-import unittest
 
 from PIL import PyAccess
 
-from .helper import PillowTestCase, hopper
+from .helper import hopper
 
 # Not running this test by default. No DOS against Travis CI.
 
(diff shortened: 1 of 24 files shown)


hugovk:
image

This Codecov "+-0.01%" is newish (I think it was only a plus or a minus earlier), and a +- hundredth of a percent is pretty meaningless.

Shall we add a `threshold` to stop it?

> Allow the coverage to drop by `X%`, and posting a `success` status.

https://docs.codecov.io/docs/commit-status#section-threshold

Perhaps `0.01%` or `0.1%`? And if so, here or in another PR?

radarhere:
I've added the conversion of unittest mock to pytest monkeypatch, after receiving the suggestion from #4488

Regarding the threshold, yes, I guess we can. I think the diff hit coverage might be more meaningful statistic of the two. I've added that commit as well.

PR #4488, opened 2020-03-28
Removed unittest mock

radarhere:
Helps #4193

Suggestion to replace unittest `mock` with custom code, including a context manager.

### DIFF
--- a/Tests/test_imagefont.py
+++ b/Tests/test_imagefont.py
@@ -3,8 +3,9 @@ import distutils.version
 import os
 import re
 import shutil
+import sys
+from contextlib import contextmanager
 from io import BytesIO
-from unittest import mock
 
 import pytest
 from PIL import Image, ImageDraw, ImageFont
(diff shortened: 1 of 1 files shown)


coiax:
`unittest.mock` is an excellent toolkit for mocking and patching, and using it doesn't mean you have to use `unittest` style tests, surely?

radarhere:
No, it doesn't. I created this PR separate to other unittest related changes because I felt it was more debatable. So if you think this shouldn't be included, that's valid, and we'll see what others think.

nulano:
Why not use the [pytest alternative](https://docs.pytest.org/en/latest/monkeypatch.html)?

radarhere:
@nulano because I didn't see it? Thanks. I've added it to PR #4486

nulano:
@radarhere Apologies if I worded that poorly, it was merely a suggestion.

radarhere:
@nulano not at all, good suggestion

PR #1770, opened 2016-03-15
yapf-formatting for setup.py

aclark4life:
I know folks don't universally agree on using yapf, but I've been using it a lot lately elsewhere & propose we start using it on a file-by-file basis here.

PR #4492, opened 2020-03-29
Add missing if brackets

nulano:
Add missing brackets to if clause to avoid possible undefined behaviour.


hugovk:
Good spot, thanks!

(It says Travis is pending, but it [passed](https://travis-ci.org/github/python-pillow/Pillow/builds/668325118). See [also](https://travis-ci.community/t/fails-to-report-build-status-to-github-again/7880).)

There's been a suggestion earlier to using clang-format on the code (https://github.com/python-pillow/Pillow/pull/1770#issuecomment-196974405), I'll make a new issue.

PR #4494, opened 2020-03-29
Re-add deprecated PILLOW_VERSION to give projects more time to upgrade

hugovk:
`PILLOW_VERSION` was deprecated in Pillow 5.2.0 in July 2018 (https://github.com/python-pillow/Pillow/pull/3090, https://github.com/python-pillow/Pillow/pull/3782) and removed in 7.0.0 in January 2020 (https://github.com/python-pillow/Pillow/pull/4107).

However, it didn't raise a deprecation warning. The removal has caused problems for users of torchvision.

torchvision had already updated to use `__version__ ` in October 2019 (https://github.com/pytorch/vision/pull/1501) but when Pillow 7.0.0 came out on 2020-01-02, had not yet released it (https://github.com/pytorch/vision/issues/1712).

The fixed [torchvision 0.5.0](https://github.com/pytorch/vision/releases/tag/v0.5.0) was released on 2020-01-16.

However, some people are still having difficulties using torchvision 0.5.0 and Pillow 7.0.0 (https://github.com/pytorch/vision/issues/1712) and are pinning to old Pillow versions or editing the source. As it's not much code for us, I suggest re-adding `PILLOW_VERSION` until some unspecified future release.

---

As before, this still doesn't raise a deprecation warning. I don't know if it's possible to raise one when calling a constant (or calling something that looks like calling a constant), but that would be welcome for 7.1.0 or a later release.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,9 @@ Changelog (Pillow)
 7.1.0 (unreleased)
 ------------------
 
+- Added reading of earlier ImageMagick PNG EXIF data #4471
+  [radarhere]
+
 - Fixed endian handling for I;16 getextrema #4457
   [radarhere]
 
--- a/docs/deprecations.rst
+++ b/docs/deprecations.rst
@@ -12,6 +12,17 @@ Deprecated features
 Below are features which are considered deprecated. Where appropriate,
 a ``DeprecationWarning`` is issued.
 
+PILLOW_VERSION constant
+~~~~~~~~~~~~~~~~~~~~~~~
+
+.. deprecated:: 5.2.0
+
+``PILLOW_VERSION`` has been deprecated and will be removed in a future release. Use
+``__version__`` instead.
+
+It was initially removed in Pillow 7.0.0, but brought back in 7.1.0 to give projects
+more time to upgrade.
+
 ImageCms.CmsProfile attributes
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
@@ -68,13 +79,6 @@ Use instead:
 
     with Image.open("hopper.png") as im:
         im.save("out.jpg")
-
-PILLOW_VERSION constant
-~~~~~~~~~~~~~~~~~~~~~~~
-
-*Removed in version 7.0.0.*
-
-``PILLOW_VERSION`` has been removed. Use ``__version__`` instead.
 
 PIL.*ImagePlugin.__version__ attributes
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/releasenotes/7.1.0.rst
+++ b/docs/releasenotes/7.1.0.rst
@@ -27,6 +27,15 @@ Reading JPEG comments
 When opening a JPEG image, the comment may now be read into
 :py:attr:`~PIL.Image.Image.info`.
 
+PILLOW_VERSION constant
+^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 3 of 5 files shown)


wiredfool:
Does @property, or something like it work for modules?

hugovk:
No, but we can do something like https://stackoverflow.com/a/880550/724176:

```python
# __init__.py
import sys
import warnings


class _Deprecated_Version(object):
    def deprecated_version(self):
        warnings.warn(
            "PILLOW_VERSION is deprecated and will be removed in a future release. "
            "Use __version__ instead.",
            DeprecationWarning,
            stacklevel=2,
        )
        return __version__

    PILLOW_VERSION = property(deprecated_version)


sys.modules[__name__] = _Deprecated_Version()
```
```python
# 1.py
import PIL

print(dir(PIL))
print(PIL.PILLOW_VERSION)
```
```console
$ python -Wall 1.py
['PILLOW_VERSION', '__class__', '__delattr__', '__dict__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', '__weakref__', 'deprecated_version']
1.py:5: DeprecationWarning: PILLOW_VERSION is deprecated and will be removed in a future release. Use __version__ instead.
  print(PIL.PILLOW_VERSION)
7.1.0.dev0
```

hugovk:
However, the `sys.modules[__name__] = _Deprecated_Version()` in `__init__.py` means:
```pycon
>>> from PIL import Image
ImportError: cannot import name 'Image' from '<unknown module name>' (unknown location)
```

hugovk:
Alternatively, use PEP 562 to warn for at least Python 3.7+ (https://stackoverflow.com/a/57263518/724176), put this in both `__init__.py` and `Image.py`:

```python
if sys.version_info >= (3, 7):
    def __getattr__(name):
        if name == 'PILLOW_VERSION':
            warnings.warn(
                "PILLOW_VERSION is deprecated and will be removed in a future release. "
                "Use __version__ instead.",
                DeprecationWarning,
                stacklevel=2,
            )
            return __version__
        raise AttributeError("module '{}' has no attribute '{}'".format(__name__, name))
else:
    PILLOW_VERSION = __version__
```

```python
# 1.py
import PIL
from PIL import Image

print(PIL.PILLOW_VERSION)
print(Image.PILLOW_VERSION)
```

```console
$ python3.6 -Wall 1.py
7.1.0.dev0
7.1.0.dev0
```
```console
$ python3.7 -Wall 1.py
1.py:5: DeprecationWarning: PILLOW_VERSION is deprecated and will be removed in a future release. Use __version__ instead.
  print(PIL.PILLOW_VERSION)
7.1.0.dev0
1.py:6: DeprecationWarning: PILLOW_VERSION is deprecated and will be removed in a future release. Use __version__ instead.
  print(Image.PILLOW_VERSION)
7.1.0.dev0
```

Of last month's pip downloads, there were about 7.2m downloads for Python 3.5+. About 52% were 3.5-3.6, about 48% for 3.7+.

hugovk:
Thanks to @radarhere, updated via https://github.com/hugovk/Pillow/pull/46 with deprecation warnings via a class for some common uses for Python 3.5-3.6, and using PEP 562/`__getattr__` for 3.7+.

PR #4495, opened 2020-03-29
Winbuild rewrite

nulano:
As promised in https://github.com/python-pillow/Pillow/pull/4084#issuecomment-534555797.

A complete rewrite of `winbuild` scripts, based on the current `test-windows.yml` scripts, to unify Windows CI scripts and simplify local development. Instead of a single script to compile all dependencies, there is a separate script for each dependency. While this is slightly slower (due to `vcvarsall.bat`), I have found this to be very helpful when adding / changing dependencies. _The larger factor slowing down AppVeyor is support for more dependencies. It might be a good idea to remove a few more AppVeyor jobs here._

Another difference is that this version uses format strings instead of ifs to fill in build parameters, which is IMO easier to read. Additionally, all commands are wrapped in functions to avoid repeating parameters for similar commands.

This PR uses `vswhere` to automatically find VS using code based on distutils, it should therefore be fairly reliable and easy to use for other contributors, but it only supports VS versions starting with 2017. I have tested it on:
* GitHub Actions with VS 2019 Enterprise (`windows-2019` image), 
* AppVeyor with VS 2017 Community (VS 2017 image),
* Windows ltsc2019 in Docker with VS 2019 Build Tools (`mcr.microsoft.com/dotnet/framework/sdk:4.8-windowsservercore-ltsc2019` image),
* Windows 10 Home (1903) with VS 2017 Community (clean and minimal install),
* Windows 10 Pro with VS 2019 Community.

This PR switches AppVeyor to the Windows Server 2016 image (Windows ~8~ 10 (1607)) due to VS 2017 availability. This means Windows ~7~ 8.1 (Windows Server 2012 R2) support will no longer be automatically tested. I don't think this is a big issue, as ~both systems~ Windows 8.1 is now EOL; if you disagree, I can try to add VS 2015 support.
**Edit**: I had the Windows version numbers wrong.

Note that since VS 2015, all MSVC versions are supposed to be ABI compatible, so any such VS version can be used to compile Python 3.5+ packages.

While `build.rst` claims Python 3.6+ is required to create valid build scripts, this only affects `build_dep_all.cmd`, which is not used on GHA, so Python 3.5 is fine there.

Please let me know if you would like me to change anything.


hugovk:
(I restarted GHA/Test because Ubuntu/3.5 failed due to an unrelated network error installing deps:
`E: Failed to fetch http://azure.archive.ubuntu.com/ubuntu/pool/main/p/python3-stdlib-extensions/python3-tk_3.6.9-1~18.04_amd64.deb  Unable to connect to azure.archive.ubuntu.com:http:`)

hugovk:
Hm, now failing due to a Git checkout thing: `##[error]fatal: reference is not a tree: 8b1233b1a7786a04c1944bfab989189e413c4b09`, will probably be fixed by a rebase or something. Anyway, I can vouch that the GHA/Test ones were passing :)

nulano:
The error might be fixed in `checkout@v2`, see https://github.com/actions/checkout/issues/23#issuecomment-572688577. This PR updates to v2 for Windows, should I try to update it for the other workflows here as well, or in a separate PR?

I imagine this PR will take a while to review, but the error doesn't seem to be too frequent.

hugovk:
Let's update it here, would be good to turn this one green. You can also put it in another PR if you like.

hugovk:
Thanks for this! Couple of quick replies:

> _The larger factor slowing down AppVeyor is support for more dependencies. It might be a good idea to remove a few more AppVeyor jobs here._

Agreed, AppVeyor is the build bottleneck. Approx end-to-end running times:

* AppVeyor: 33 minutes for 6 jobs
* GitHub Actions: 12m for 33 jobs
* Travis CI: 8m for 9 jobs

What would you suggest removing?

> This PR switches AppVeyor to the Windows Server 2016 image (Windows 8) due to VS 2017 availability. This means Windows 7 (Windows Server 2012) support will no longer be automatically tested. I don't think this is a big issue, as both systems are now EOL; if you disagree, I can try to add VS 2015 support.

I think that's fine: "In general, we aim to support all current versions of Linux, macOS, and Windows".

https://pillow.readthedocs.io/en/stable/installation.html#platform-support

nulano:
> What would you suggest removing?

I would suggest removing Python38-x64 and Python35, and keeping Python38 and Python35-x64. This way the newest and oldest Python, and both architectures are being tested on a different system. This should get AV back to pre-PR speed.

I think MSYS2 can be moved to GHA once it is preinstalled (https://github.com/actions/virtual-environments/pull/632).

I'm not sure about PyPy. Since it is installed manually on AV, it can be updated as soon as a new version is released, unlike GHA where it takes at least a week (they mention in multiple issues that this is their standard rollout period). But it does take almost 8 min...

hugovk:
> > What would you suggest removing?
> 
> I would suggest removing Python38-x64 and Python35, and keeping Python38 and Python35-x64. This way the newest and oldest Python, and both architectures are being tested on a different system. This should get AV back to pre-PR speed.
> 
> I think MSYS2 can be moved to GHA once it is preinstalled ([actions/virtual-environments#632](https://github.com/actions/virtual-environments/pull/632)).

👍 

> I'm not sure about PyPy. Since it is installed manually on AV, it can be updated as soon as a new version is released, unlike GHA where it takes at least a week (they mention in multiple issues that this is their standard rollout period). But it does take almost 8 min...

We rarely need to have PyPy available as soon as it's released, so following the standard rollout should be fine. A week or a few shouldn't be much of a problem. And we're basically following standard rollouts for all the other CIs/versions.

nulano:
I have merged with master, removed Python38, Python35-x64, and PyPy3 from AppVeyor, and updated `install.rst`.

winbuild/build_prepare.py
```diff
@@ -0,0 +1,557 @@
...
+        "libs": [r"output\release-static\{architecture}\lib\*.lib"],
+    },
+    "freetype": {
+        "url": "https://download.savannah.gnu.org/releases/freetype/freetype-2.10.1.tar.gz",  # noqa: E501
+        "filename": "freetype-2.10.1.tar.gz",
+        "dir": "freetype-2.10.1",
+        "patch": {
+            r"builds\windows\vc2010\freetype.vcxproj": {
+                # freetype setting is /MD for .dll and /MT for .lib, we need /MD
+                "<RuntimeLibrary>MultiThreaded</RuntimeLibrary>": "<RuntimeLibrary>MultiThreadedDLL</RuntimeLibrary>",  # noqa E501
```
hugovk, line 179:
```suggestion
                "<RuntimeLibrary>MultiThreaded</RuntimeLibrary>": "<RuntimeLibrary>MultiThreadedDLL</RuntimeLibrary>",  # noqa: E501
```

winbuild/build_prepare.py
```diff
@@ -0,0 +1,557 @@
...
+                # freetype doesn't specify SDK version, MSBuild may guess incorrectly
+                '<PropertyGroup Label="Globals">': '<PropertyGroup Label="Globals">\n    <WindowsTargetPlatformVersion>$(WindowsSDKVersion)</WindowsTargetPlatformVersion>',  # noqa E501
```
hugovk, line 181:
```suggestion
                '<PropertyGroup Label="Globals">': '<PropertyGroup Label="Globals">\n    <WindowsTargetPlatformVersion>$(WindowsSDKVersion)</WindowsTargetPlatformVersion>',  # noqa: E501
```

winbuild/build_prepare.py
```diff
@@ -0,0 +1,557 @@
...
+    "lcms2": {
+        "url": SF_MIRROR + "/project/lcms/lcms/2.9/lcms2-2.9.tar.gz",
+        "filename": "lcms2-2.9.tar.gz",
+        "dir": "lcms2-2.9",
+        "patch": {
+            r"Projects\VC2017\lcms2_static\lcms2_static.vcxproj": {
+                # default is /MD for x86 and /MT for x64, we need /MD always
+                "<RuntimeLibrary>MultiThreaded</RuntimeLibrary>": "<RuntimeLibrary>MultiThreadedDLL</RuntimeLibrary>",  # noqa E501
+                # retarget to default toolset (selected by vcvarsall.bat)
+                "<PlatformToolset>v141</PlatformToolset>": "<PlatformToolset>$(DefaultPlatformToolset)</PlatformToolset>",  # noqa E501
```
hugovk, line 206:
```suggestion
                "<PlatformToolset>v141</PlatformToolset>": "<PlatformToolset>$(DefaultPlatformToolset)</PlatformToolset>",  # noqa: E501
```

winbuild/build_prepare.py
```diff
@@ -0,0 +1,557 @@
...
+        # "bins": [r"objs\{msbuild_arch}\Release\freetype.dll"],
+    },
+    "lcms2": {
+        "url": SF_MIRROR + "/project/lcms/lcms/2.9/lcms2-2.9.tar.gz",
+        "filename": "lcms2-2.9.tar.gz",
+        "dir": "lcms2-2.9",
+        "patch": {
+            r"Projects\VC2017\lcms2_static\lcms2_static.vcxproj": {
+                # default is /MD for x86 and /MT for x64, we need /MD always
+                "<RuntimeLibrary>MultiThreaded</RuntimeLibrary>": "<RuntimeLibrary>MultiThreadedDLL</RuntimeLibrary>",  # noqa E501
```
hugovk, line 204:
```suggestion
                "<RuntimeLibrary>MultiThreaded</RuntimeLibrary>": "<RuntimeLibrary>MultiThreadedDLL</RuntimeLibrary>",  # noqa: E501
```

winbuild/build_prepare.py
```diff
@@ -0,0 +1,557 @@
...
+                # retarget to default toolset (selected by vcvarsall.bat)
+                "<PlatformToolset>v141</PlatformToolset>": "<PlatformToolset>$(DefaultPlatformToolset)</PlatformToolset>",  # noqa E501
+                # retarget to latest (selected by vcvarsall.bat)
+                "<WindowsTargetPlatformVersion>8.1</WindowsTargetPlatformVersion>": "<WindowsTargetPlatformVersion>$(WindowsSDKVersion)</WindowsTargetPlatformVersion>",  # noqa E501
```
hugovk, line 208:
```suggestion
                "<WindowsTargetPlatformVersion>8.1</WindowsTargetPlatformVersion>": "<WindowsTargetPlatformVersion>$(WindowsSDKVersion)</WindowsTargetPlatformVersion>",  # noqa: E501
```

winbuild/build_prepare.py
```diff
@@ -0,0 +1,557 @@
...
+            cmd_copy(r"{winbuild_dir}\tiff.opt", "nmake.opt"),
+            cmd_nmake("makefile.vc", "clean"),
+            cmd_nmake("makefile.vc", "lib"),
+        ],
+        "headers": [r"libtiff\tiff*.h"],
+        "libs": [r"libtiff\*.lib"],
+        # "bins": [r"libtiff\*.dll"],
+    },
+    "libwebp": {
+        "url": "http://downloads.webmproject.org/releases/webp/libwebp-1.0.3.tar.gz",  # noqa: E501
```
hugovk, line 157:
```suggestion
        "url": "http://downloads.webmproject.org/releases/webp/libwebp-1.0.3.tar.gz",
```

nulano:
It seems @radarhere's rebase has removed some changes from merge commits. For example, #4342 changed `test-windows.yml`, so I resolved the conflict by updating `build_prepare.py` accordingly. The rebase removed this merge commit, and thus also this change. (Effectively reverting #4342 in this branch.) This is also what caused the lint failure. I'll compare my local copy with the rebase to restore these changes in a separate commit.

On that note, does anyone know how to do a merge like that properly to avoid this in the future? Or should I ask on StackOverflow?

hugovk:
@cgohlke Hi, please could you have a look at this PR? At least to check your it's okay with your workflow? Thank you!

cgohlke:
I don't see anything in this PR that could interfere with my build system. I don't rely on `winbuild` or build instructions.

docs/installation.rst
```diff
@@ -292,9 +292,7 @@ or from within the uncompressed source directory::
 Building on Windows
 ^^^^^^^^^^^^^^^^^^^
 
-We don't recommend trying to build on Windows. It is a maze of twisty
```
radarhere, line 295: Just to be clear - so you think that this PR improves the situation enough that it is now viable for users to compile from source by themselves?

nulano: While installing Visual Studio is not exactly easy, once that is done, compiling is as difficult as running the script. It is by no means a maze anymore. I have not had to do any extra configuration on my new laptop whatsoever.

That said, when writing it I was thinking more about developers (contributing PRs) than users. AFAIK most Windows users aren't used to compiling code at all, so perhaps it would be best to write it as "We recommend you use prebuilt wheels from PyPI. If you wish to compile Pillow manually, you can use the build scripts and notes ..."

nulano:
Thanks to v2 of actions/upload-artifact, I have enabled wheel building for PyPy (it was disabled due to .egg filename conflict).

I also updated the windows instructions as follows:
> We recommend you use prebuilt wheels from PyPI. If you wish to compile Pillow manually, you can use the build scripts in the winbuild directory used for CI testing and development. These scripts require Visual Studio 2017 or newer and NASM.

docs/installation.rst
hugovk: @radarhere any comments on the current wording? Okay to merge?

radarhere: Wording is fine, just wanted to check that we knew that we were stepping up our level of commitment to building from source on Windows.

hugovk:
@nulano Please could you resolve the conflicts? Then let's get this merged. Thanks!

nulano:
Merged with master; Travis failed because Python 3.5 on s390x was cancelled for some reason.

hugovk:
Thank you! It was hanging so I cancelled it, and have just restarted it. I'll give it a chance to finish, but it's not related to this PR so we can merge it without it too.

hugovk:
✅ Thanks again for this!

PR #4499, opened 2020-03-31
Update Tidelift UTM params

phated:
Heya Pillow folks! Hope things are well (as they can be).

We have been moving towards automation of marketing leads, so I've been reviewing the UTM parameters and marketing language for lifted projects. It looks like Pillow uses our old style params, so I'm submitting this and I've submitted https://github.com/python-pillow/python-pillow.github.io/pull/18 to the website.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -858,7 +858,7 @@ try:
         project_urls={
             "Documentation": "https://pillow.readthedocs.io",
             "Source": "https://github.com/python-pillow/Pillow",
-            "Funding": "https://tidelift.com/subscription/pkg/pypi-pillow",
+            "Funding": "https://tidelift.com/subscription/pkg/pypi-pillow?utm_source=pypi-pillow&utm_medium=pypi",
         },
         classifiers=[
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -858,7 +858,7 @@ def debug_build():
         project_urls={
             "Documentation": "https://pillow.readthedocs.io",
             "Source": "https://github.com/python-pillow/Pillow",
-            "Funding": "https://tidelift.com/subscription/pkg/pypi-pillow",
+            "Funding": "https://tidelift.com/subscription/pkg/pypi-pillow?utm_source=pypi-pillow&utm_medium=pypi",
```
hugovk: Something like this should fix the lint:
```suggestion
            "Funding": "https://tidelift.com/subscription/pkg/pypi-pillow?"
            "utm_source=pypi-pillow&utm_medium=pypi",
```

phated: Thanks, accepted

hugovk:
Thanks!

PR #4500, opened 2020-04-01
Update install docs and release notes for #4260

nulano:
Release notes for #4260 
Update installation docs for #4260 and #4029 

I'm assuming Windows and macOS don't need (and will not have) XCB support in distributed wheels, as this is quite an unusual (although valid) scenario; please correct this if necessary. (I don't think XCB even supports Windows outside MinGW.)


hugovk:
Thanks!

PR #4504, opened 2020-04-01
Fix buffer overflow in SGI-RLE decoding

hugovk:


utkarsh2102:
Hi @hugovk,
Is version 2.6.1 affected with this?
(because it doesn't look like the case to me!)

hugovk:
I don't know, but 2.6.1 is so old ([October 2014] (https://pypi.org/project/Pillow/2.6.1)), there will be some unfixed CVEs in there so it's advisable to upgrade. 

* https://cve.mitre.org/cgi-bin/cvekey.cgi?keyword=Pillow

utkarsh2102:
Yeah, thank you! :)

PR #4507, opened 2020-04-01
Fix 2 buffer overflows in TIFF decoding

hugovk:


hugovk:
The https://github.com/python-pillow/Pillow/pull/4507/commits/8d4fdf603c805021d2f7b9fcb9c2c7783c57953a commit to "Use actions/checkout@v2 to fix checkout after restarts" isn't helping for Windows:

image

Removing from this PR, can deal with it after the release.

hugovk:
This is showing some older failed GHA Windows jobs, but also showing the re-runs as green. To double check, here's the same GHA jobs passing on my fork: https://github.com/hugovk/Pillow/pull/47.

Merging.

Issue #4508, opened 2020-04-01
GHA: Add retry mechanism


hugovk:
The "Install Linux dependencies" step of test.yml is failing intermittently with timeouts on GitHub Actions.

```
Reading package lists...
E: Failed to fetch http://azure.archive.ubuntu.com/ubuntu/pool/main/b/blt/tk8.6-blt2.5_2.5.3+dfsg-4_amd64.deb  Could not connect to azure.archive.ubuntu.com:80 (52.177.174.250), connection timed out
```

It's possible to restart failures, but unlike Travis CI (which allows you to restart single jobs) and AppVeyor (which allows you to restart failed jobs), you can only re-run all the jobs on GHA.

At least it only re-runs the jobs in a single workflow, but it's still more than needed, and you may get a timeout in another job.

Let's add some sort of retry mechanism. Perhaps something similar to `travis_retry`

* https://docs.travis-ci.com/user/common-build-problems/#travis_retry
* https://stackoverflow.com/q/59356518/724176

Full log:

```
2020-04-01T08:21:29.9235373Z ##[section]Starting: Request a runner to run this job
2020-04-01T08:21:31.1331011Z Requesting a hosted runner in current repository's account/organization with labels: 'ubuntu-latest', require runner match: True
2020-04-01T08:21:31.2309561Z Labels matched hosted runners has been found, waiting for one of them get assigned for this job.
2020-04-01T08:21:31.3241895Z ##[section]Finishing: Request a runner to run this job
2020-04-01T08:22:34.1239751Z Current runner version: '2.165.2'
2020-04-01T08:22:34.1240696Z Prepare workflow directory
2020-04-01T08:22:34.1402410Z Prepare all required actions
2020-04-01T08:22:34.1411342Z Download action repository 'actions/checkout@v2'
2020-04-01T08:22:35.9113927Z Download action repository 'actions/cache@v1'
2020-04-01T08:22:36.0523705Z Download action repository 'actions/setup-python@v1'
2020-04-01T08:22:36.2206676Z Download action repository 'actions/upload-artifact@v1'
2020-04-01T08:22:36.4367930Z ##[group]Run actions/checkout@v2
2020-04-01T08:22:36.4368206Z with:
2020-04-01T08:22:36.4368401Z   repository: python-pillow/Pillow
2020-04-01T08:22:36.4368602Z   token: ***
2020-04-01T08:22:36.4368696Z   ssh-strict: true
[...]

PR #4514, opened 2020-04-02
Removed redundant arguments

radarhere:


hugovk:
The two failures are network errors, like https://github.com/python-pillow/Pillow/issues/4508. Merging.

PR #4339, opened 2020-01-06
Do not install PyQt5 5.14.1

radarhere:
PyQt5 5.14.1 was released 9 hours ago - https://pypi.org/project/PyQt5/5.14.1/#history

Since then, tests have been failing in master. This is presumably a temporary measure until PyQt5 5.14.2 is released.

PR #4521, opened 2020-04-03
Do not install PyQt5 5.14.2

hugovk:
Like https://github.com/python-pillow/Pillow/pull/4339 but for the new PyQt 5.14.2 which was released today (https://pypi.org/project/PyQt5/5.14.2/) and builds have started failing:

https://travis-ci.org/github/python-pillow/Pillow/jobs/670532178#L606


radarhere:
I've created PR #4527 as an alternative - it fixes the problem by upgrading pip first.

PR #4523, opened 2020-04-03
Implemented another ellipse drawing algorithm

xtsm:
This PR changes ellipse drawing algorithm from simply drawing 360-gon to something that resembles Bresenham's algo for circles. It works like 100x faster (on my machine, according to randomized test) and produces nicer results, especially for smaller ellipse sizes.

Arcs, pies, etc. are left untouched for now.

Upd: I've also changed some tests (e.g. ellipse_symmetric and ellipse_edge) according to new drawing behavior. If now they're not checking what they were supposed to check, please tell me.


xtsm:
Oops, I thought selftest.py actually does tests.

src/libImaging/Draw.c
```diff
@@ -969,6 +970,182 @@ ellipse(Imaging im, int x0, int y0, int x1, int y1,
...
+        }
+        s->cx = nx;
+        s->cy = ny;
+    }
+    return 0;
+}
+
+// quarter_* stuff can "draw" a quarter of an ellipse with thickness 1, great.
+// Now we use ellipse_* stuff to join all four quarters of two different sized
+// ellipses and recieve horizontal segments of a complete ellipse with
```
radarhere, line 1042:
```suggestion
// ellipses and receive horizontal segments of a complete ellipse with
```

src/libImaging/Draw.c
```diff
@@ -988,7 +1165,7 @@ int
 ImagingDrawEllipse(Imaging im, int x0, int y0, int x1, int y1,
                    const void* ink, int fill, int width, int op)
 {
-    return ellipse(im, x0, y0, x1, y1, 0, 360, ink, fill, width, CHORD, op);
+    return ellipseNew(im, x0, y0, x1, y1, ink, fill, width, op);
```
homm, line 1168: Is there any chance new implementation is used for other functions? At least for ImagingDrawChord?

homm:
All test images are circles, actually, not ellipses. Could you add ones?

src/libImaging/Draw.c
xtsm: Yeah, but it would involve another layer of processing that clips ellipse down to arc.
I've got it more or less done locally, maybe I'll commit that later. The code is gonna look funky though due to all the floating point geometry and special cases processing...

hugovk:
> All test images are circles, actually, not ellipses. Could you add ones?

@xtsm Little reminder for ^ this, thanks!

src/libImaging/Draw.c
xtsm: Arcs, chords and pies are _kinda_ working now.
At least compared to whatever was in master before.

I also added a test for various size ellipses. I added it long ago but forgot to push.

hugovk (commented):
I see elliptical images have been added (e.g. https://github.com/python-pillow/Pillow/commit/5830a641ccc039652c77efccee854e3f12bc9adc), thanks! Any more review comments?

hugovk:
Thanks!

Please also update the release notes if you think it's necessary: https://github.com/python-pillow/Pillow/blob/master/docs/releasenotes/8.0.0.rst

hugovk:
Release notes PR: https://github.com/python-pillow/Pillow/pull/4976

PR #4284, opened 2019-12-21
make doccheck: turn warnings into errors

hugovk:
Changes proposed in this pull request:

 * Catch issues like https://github.com/python-pillow/Pillow/pull/4271 and https://github.com/python-pillow/Pillow/pull/4282 on the CI, by turning the warnings from sphinx-build into errors

# Warnings as warnings

For example, a build of an commit before those two fixes, with normal warnings, has exit code 0 so the CI passes:

```
⌂90% [hugo:~/github/Pillow] 52113a82+* ± make doccheck
/Library/Developer/CommandLineTools/usr/bin/make -C docs html
sphinx-build -b html -d _build/doctrees   . _build/html
Running Sphinx v2.1.2
making output directory... done
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 72 source files that are out of date
updating environment: 72 added, 0 changed, 0 removed
reading sources... [100%] releasenotes/index
/Users/hugo/github/Pillow/docs/handbook/concepts.rst:168: WARNING: Malformed table.

+------------+-------------+-----------+-------------+
| Filter     | Downscaling | Upscaling | Performance |
|            | quality     | quality   |             |
+============+=============+===========+=============+
|``NEAREST`` |             |           | ⭐⭐⭐⭐⭐       |
+------------+-------------+-----------+-------------+
|``BOX``     | ⭐           |           | ⭐⭐⭐⭐        |
+------------+-------------+-----------+-------------+
|``BILINEAR``| ⭐           | ⭐         | ⭐⭐⭐         |
+------------+-------------+-----------+-------------+
|``HAMMING`` | ⭐⭐          |           | ⭐⭐⭐         |
+------------+-------------+-----------+-------------+
|``BICUBIC`` | ⭐⭐⭐         | ⭐⭐⭐       | ⭐⭐          |
+------------+-------------+-----------+-------------+
|``LANCZOS`` | ⭐⭐⭐⭐        | ⭐⭐⭐⭐      | ⭐           |
+------------+-------------+-----------+-------------+
/Users/hugo/github/pillow-security/src/PIL/Image.py:docstring of PIL.Image.Image.transform:16: WARNING: Unexpected indentation.
/Users/hugo/github/pillow-security/src/PIL/Image.py:docstring of PIL.Image.Image.transform:22: WARNING: Unexpected indentation.
/Users/hugo/github/pillow-security/src/PIL/ImageOps.py:docstring of PIL.ImageOps.pad:12: WARNING: Unexpected indentation.
[...]

PR #4524, opened 2020-04-03
GHA: Check docs

hugovk:

 * Revert https://github.com/python-pillow/Pillow/commit/fbb14f67a30b612454bc7cf9a5f73d6b35e7eb14 from https://github.com/python-pillow/Pillow/pull/4284 as `make doccheck` is now possible on GitHub Actions
 * Do it only for Python 3.8 on Ubuntu

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -95,6 +95,12 @@ jobs:
         name: errors
         path: Tests/errors
 
+    - name: Docs
+      if: matrix.python-version == 3.8
+      run: |
+        pip install sphinx-rtd-theme
+        make doccheck
+
     - name: After success
       if: success()
       run: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -96,7 +96,7 @@ jobs:
         path: Tests/errors
 
     - name: Docs
-      if: matrix.python-version == 3.8
+      if: startsWith(matrix.os, 'ubuntu') && matrix.python-version == 3.8
       run: |
         pip install sphinx-rtd-theme
         make doccheck


.github/workflows/test.yml
```diff
@@ -95,6 +95,12 @@ jobs:
...
         path: Tests/errors
 
+    - name: Docs
+      if: startsWith(matrix.os, 'ubuntu') && matrix.python-version == 3.8
+      run: |
+        pip install sphinx-rtd-theme
+        make doccheck
+
     - name: After success
       if: success()
```
radarhere, line 105: Could you explain to me how `if: success()` is different to the default behaviour? I'm trying to figure out Docs could interact with the general flow.

hugovk: Yeah, actually, it looks like it doesn't do anything different in this context.

* Each yml file is a workflow.

  * Each workflow has a number of jobs. (We're only using a single job in our workflows.)

    * Each job has a number of steps.


The docs only mention steps:

> **success**
> Returns true when none of the previous steps have failed or been canceled.

https://help.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#success

It's not entirely clear: perhaps it also checks previous steps in previous jobs?

Anyway, doesn't look like we actually need it.

nulano: IIRC a previous version of the docs explicitly mentioned that `success()` is the default `if` condition. There is now only this: https://help.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#job-status-check-functions

> You can use the following status check functions as expressions in if conditionals. If your if expression does not contain any of the status functions it will automatically result with success().

radarhere: Thanks for the explanations. I've created PR #4542 to remove them then, if they're not doing anything.

PR #4527, opened 2020-04-05
Update pip to fix PyQt5 install

radarhere:
Alternative to #4521. Improved fix over the previous solution from #4339

Credit to https://github.com/vidify/vidify/issues/49#issuecomment-598215649


hugovk:
Much better, thanks!

PR #4534, opened 2020-04-06
Retry on apt-get update problems

radarhere:
Helps #4508 

The issue reports that `apt-get update` is failing sometimes. This PR attempts to help by running the command up to two more times if there is a warning or error.

However, when I did manage to see `apt-get update` fail on this branch, the command ran again, but it also failed on the subsequent attempts. So be aware of that.


hugovk:
Thanks!

Issue #4535, opened 2020-04-06
ImageFont documentation does not match with current code


marado:
### What did you do?

Read the docs on 
https://pillow.readthedocs.io/en/stable/reference/ImageFont.html#PIL.ImageFont.ImageFont
and then the code by following the link provided there, to
https://pillow.readthedocs.io/en/stable/_modules/PIL/ImageFont.html#load

### What did you expect to happen?

Since the documentation claims that only an IOError is raised, I expected to see the code raising an IOError.

### What actually happened?

The code shows an OSError being raised, as a result of this commit: https://github.com/python-pillow/Pillow/commit/538d9e2e5d252bea23e12da08495945d776aed78#diff-0b53b789652f38dc633674dd3671e636L82

RioBaer:
Marcus, how do you mean what did I do? Please illuminate?
Regards Dan AKA BoomerNaut
Google Maps local guide level eight

hugovk:
Thanks for the report! I've made https://github.com/python-pillow/Pillow/pull/4536 to fix the docs, plus some related updates.

radarhere:
You can now see the updated documentation at https://pillow.readthedocs.io/en/latest/_modules/PIL/ImageFont.html#load. It will become 'stable' with the next release of Pillow on July 1st.

PR #4536, opened 2020-04-07
Replace IOError and WindowsError aliases with OSError

hugovk:
Fixes https://github.com/python-pillow/Pillow/issues/4535.

 * > _Changed in [Python] version 3.3:_ `EnvironmentError`, `IOError`, `WindowsError`, `socket.error`, `select.error` and `mmap.error` have been merged into `OSError`, and the constructor may return a subclass.
 * https://docs.python.org/3/library/exceptions.html#OSError
 * [pyupgrade](https://github.com/asottile/pyupgrade/#oserror-aliases) replaced some in https://github.com/python-pillow/Pillow/pull/4109
 * This PR replaces the rest
 
Two questions:

In `src/PIL/ImageFile.py`, this PR renames the `raise_ioerror` helper function to `raise_oserror`. Technically this could be considered a breaking change.

1. Should we keep the old `raise_ioerror` with a deprecation warning?
2. Should we name the new one `_raise_oserror` to indicate it is internal?

### DIFF
--- a/docs/releasenotes/7.0.0.rst
+++ b/docs/releasenotes/7.0.0.rst
@@ -85,7 +85,7 @@ Custom unidentified image error
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 Pillow will now throw a custom ``UnidentifiedImageError`` when an image cannot be
-identified. For backwards compatibility, this will inherit from ``IOError``.
+identified. For backwards compatibility, this will inherit from ``OSError``.
 
 New argument ``reducing_gap`` for Image.resize() and Image.thumbnail() methods
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/Tests/check_j2k_overflow.py
+++ b/Tests/check_j2k_overflow.py
@@ -5,5 +5,5 @@ from PIL import Image
 def test_j2k_overflow(tmp_path):
     im = Image.new("RGBA", (1024, 131584))
     target = str(tmp_path / "temp.jpc")
-    with pytest.raises(IOError):
+    with pytest.raises(OSError):
         im.save(target)
--- a/Tests/check_libtiff_segfault.py
+++ b/Tests/check_libtiff_segfault.py
@@ -9,6 +9,6 @@ def test_libtiff_segfault():
         libtiff >= 4.0.0
         """
 
-    with pytest.raises(IOError):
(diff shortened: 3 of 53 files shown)


docs/releasenotes/6.1.0.rst
```diff
@@ -62,7 +62,7 @@ file. ``ImageFont.FreeTypeFont`` has four new methods,
 :py:meth:`PIL.ImageFont.FreeTypeFont.set_variation_by_name` for using named styles, and
 :py:meth:`PIL.ImageFont.FreeTypeFont.get_variation_axes` and
 :py:meth:`PIL.ImageFont.FreeTypeFont.set_variation_by_axes` for using font axes
-instead. An ``IOError`` will be raised if the font is not a variation font. FreeType
+instead. An ``OSError`` will be raised if the font is not a variation font. FreeType
```
radarhere, line 65: Feel free to disagree with me, but at the time of 4.2.0 and 6.1.0, they were IOErrors, so for the sake of anyone using those versions, I would suggest leaving this?

hugovk: Good idea, I've rebased it out.

radarhere: Thanks, but...

docs/releasenotes/4.2.0.rst
```diff
@@ -42,7 +42,7 @@ Several deprecated items have been removed.
 * Before Pillow 4.2.0, attempting to save an RGBA image as JPEG would
   discard the alpha channel. From Pillow 3.4.0, a deprecation warning
   was shown. From Pillow 4.2.0, the deprecation warning is removed and
-  an :py:exc:`IOError` is raised.
+  an :py:exc:`OSError` is raised.
```
radarhere, line 45: ...there is the same situation here.

hugovk: Heh, will fixup too.

docs/releasenotes/7.0.0.rst
```diff
@@ -85,7 +85,7 @@ Custom unidentified image error
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 Pillow will now throw a custom ``UnidentifiedImageError`` when an image cannot be
-identified. For backwards compatibility, this will inherit from ``IOError``.
+identified. For backwards compatibility, this will inherit from ``OSError``.
```
hugovk, line 88: @radarhere What do you think about this one?

radarhere: I would think that's fine to change, because Pillow 7 dropped support for Python 2.7, so both errors are synonymous at this point.

Issue #4485, opened 2020-03-27
Wrong positioning when calculating height of line without descenders


jesusloplar:
As you may see in the picture below, the second and third line seems to be closer than the rest. I think this is because of the quoting marks found in the following line.

<img width="544" alt="Screenshot 2020-03-27 at 14 27 16" src="https://user-images.githubusercontent.com/46565164/77760702-11a8aa80-7037-11ea-9ea1-7a0224912c36.png">

hugovk:
Thank you for reporting an issue.

Follow these guidelines to ensure your issue is handled properly.

If you have a ...

1.  General question: consider asking the question on Stack Overflow
    with the python-imaging-library tag:

    * https://stackoverflow.com/questions/tagged/python-imaging-library

    Do not ask a question in both places.

    If you think you have found a bug or have an unexplained exception
    then file a bug report here.

2.  Bug report: include a self-contained, copy-pastable example that
    generates the issue if possible. Be concise with code posted.
    Guidelines on how to provide a good bug report:

    * https://stackoverflow.com/help/mcve

    Bug reports which follow these guidelines are easier to diagnose,
    and are often handled much more quickly.

3.  Feature request: do a quick search of existing issues
    to make sure this has not been asked before.

We know asking good questions takes effort, and we appreciate your time.
Thank you.

### What did you do?

### What did you expect to happen?

### What actually happened?

### What are your OS, Python and Pillow versions?

* OS: 
* Python: 
* Pillow: 

Please include **code** that reproduces the issue and whenever possible, an **image** that demonstrates the issue. Please upload images to GitHub, not to third-party file hosting sites. If necessary, add the image to a zip or tar archive.

The best reproductions are self-contained scripts with minimal dependencies. If you are using a framework such as Plone, Django, or Buildout, try to replicate the issue just using Pillow.

```python
code goes here
```

radarhere:
While providing a clear self-contained code example is absolutely the best way forward to help resolve your problem, I don't think this is going to be because of the quote marks. I think it will be because your second line 'currentGDP. Given that one of his senior' has no descenders - none of the letters go below the baseline. None of them are g, j, p, q or y.

jesusloplar:
Yes, you are completely right. Sorry about not having attached some code. You can see the script [here](https://github.com/jesusloplar/KindleHighlight2Picture/blob/master/note_to_picture.py). I start to draw the text on the image on the 66th line:

```
# Setting the font for the body
[...]

PR #4548, opened 2020-04-12
Added multiline text example

radarhere:
Suggested in https://github.com/python-pillow/Pillow/issues/4485#issuecomment-605428183


hugovk:
Thanks!

PR #4550, opened 2020-04-12
Replaced TiffImagePlugin DEBUG with logging

### LINKED ISSUE #2341: Cleanup: Remove all print() calls from the code
Scattered around the codebase is a bunch of print() calls.

Pillow should instead use `logging`; that is, if the calls are at all relevant. In [my latest PR](https://github.com/python-pillow/Pillow/pull/2339) I dropped a bunch of them already, most of which were commented out.
A bunch of them are behind a DEBUG flag as well, which can also be dropped if put behind `logging.debug` instead.

[Here](https://github.com/jleclanche/fireplace/blob/master/fireplace/logging.py) is an example of how to set up a consistent logging module across a library. But imho, it's not even necessary.

radarhere:
Resolves #2341 and https://github.com/python-pillow/Pillow/blob/97280a045cc2544ef8855dac2765e765565772cd/src/PIL/TiffImagePlugin.py#L54

### DIFF
--- a/src/PIL/TiffImagePlugin.py
+++ b/src/PIL/TiffImagePlugin.py
@@ -40,6 +40,7 @@
 #
 import io
 import itertools
+import logging
 import os
 import struct
 import warnings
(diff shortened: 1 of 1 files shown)


hugovk:
Thanks!

PR #4551, opened 2020-04-13
Fixed comparison warnings

radarhere:
* JpegEncode.c [before](https://travis-ci.org/github/python-pillow/Pillow/jobs/673981863#L2357-L2361) and [after](https://travis-ci.org/github/python-pillow/Pillow/jobs/674229403#L2275)
* Jpeg2KEncode.c [before](https://travis-ci.org/github/python-pillow/Pillow/jobs/673981863#L2432-L2440) and [after](https://travis-ci.org/github/python-pillow/Pillow/jobs/674229403#L2346)
* QuantPngQuant.c [before](https://travis-ci.org/github/python-pillow/Pillow/jobs/673981863#L2442-L2449) and [after](https://travis-ci.org/github/python-pillow/Pillow/jobs/674229403#L2348)

### DIFF
--- a/src/libImaging/Jpeg2KEncode.c
+++ b/src/libImaging/Jpeg2KEncode.c
@@ -50,7 +50,7 @@ static OPJ_SIZE_T
 j2k_write(void *p_buffer, OPJ_SIZE_T p_nb_bytes, void *p_user_data)
 {
     ImagingCodecState state = (ImagingCodecState)p_user_data;
-    int result;
+    unsigned int result;
 
     result = _imaging_write_pyFd(state->fd, p_buffer, p_nb_bytes);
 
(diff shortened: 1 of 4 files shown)


src/libImaging/Jpeg2KEncode.c
```diff
@@ -396,8 +396,8 @@ j2k_encode_entry(Imaging im, ImagingCodecState state)
         Py_ssize_t n;
         float *pq;
 
-        if (len) {
-            if (len > sizeof(params.tcp_rates) / sizeof(params.tcp_rates[0]))
+        if (len > 0) {
+            if ((unsigned)len > sizeof(params.tcp_rates) / sizeof(params.tcp_rates[0]))
```
radarhere, line 400: `len` [comes from `PySequence_Length`](https://github.com/python-pillow/Pillow/blob/345e97b593a86323a58d5b061d48937f0dde55be/src/libImaging/Jpeg2KEncode.c#L395), which can return [-1 on failure](https://docs.python.org/3/c-api/sequence.html#c.PySequence_Length). So if we change the condition to not just exclude empty sequence lengths, but also sequence lengths with errors, then the unsigned cast becomes safe to use.

PR #4560, opened 2020-04-16
Parse orientation from XMP tags

### LINKED ISSUE #4529: PIL failed to extract EXIF while MacOS Preview App can
### What did you do?
Hi, I'm trying to extract EXIF from the bellow PNG image, mainly to rotate it correctly based on the orientation information inside the EXIF tags.

test

with this simple script:

```python
import PIL

print(PIL.__version__)
from PIL import Image

image = 'test.png'
exif_info = Image.open(image).getexif()
print(exif_info)
```

but PIL failed to extract any EXIF information out of it. I can confirm that the EXIF information is there and can be detected via Preview app on MacOS:

image

As you can see, it said: **Orientation: 8 (Rotated 90° CW)**

`getexif()` return the EXIF information for the image, at least contain the orientation like MacOS Preview did.

Noted that some month ago I ran the same exact code and everything work fine, maybe somehow `getexif()` was broken in newer version of Pillow. I have tried to revert back to `6.0.0` but it does not solve the problem.

Also I have noticed that `getexif()` will almost fail with any file was rotated by Quick Look tool from MacOS

image

### What actually happened?
`getexif()` return an empty `dict` and failed to extract any EXIF

Output:
```bash
7.1.1
{}
```

* OS: MacOS Mojave 10.14.5
* Python: Python 3.7.6
* Pillow: 7.1.1

radarhere:
Resolves #4529

Reads orientation from [XMP tags](https://exiftool.org/TagNames/XMP.html#tiff) in `getexif()`. Test image supplied in the issue.

While I'm modifying `getexif()`, I've also included a commit so that parsing EXIF from ImageMagick zTXt chunks is not performed solely in PngImageFile - after `copy()`, a PngImageFile becomes just an ImageFile, and I don't think we should ignore the ImageMagick data afterwards.


hugovk:
Thanks!

PR #4568, opened 2020-04-19
Add append_images support for ICO

ziplantil:
This adds support for the append_images parameter when saving .ICO files. Its behavior is much the same as for .ICNS files, and can be used to provide alternative sprites for particular icon sizes instead of simply resizing the main image to fit that size.

### DIFF
--- a/Tests/test_file_ico.py
+++ b/Tests/test_file_ico.py
@@ -90,7 +90,7 @@ def test_only_save_append_images(tmp_path):
     im = hopper()
     provided_im = Image.new("RGBA", (32, 32), (255, 0, 0, 255))
     outfile = str(tmp_path / "temp_saved_multi_icon.ico")
-    im.save(outfile, sizes = [(32, 32), (64, 64)], append_images = [provided_im])
+    im.save(outfile, sizes=[(32, 32), (64, 64)], append_images=[provided_im])
 
     with Image.open(outfile) as reread:
         reread.size = (64, 64)
--- a/Tests/test_file_ico.py
+++ b/Tests/test_file_ico.py
@@ -85,6 +85,24 @@ def test_only_save_relevant_sizes(tmp_path):
         assert im_saved.info["sizes"] == {(16, 16), (24, 24), (32, 32), (48, 48)}
 
 
+def test_only_save_append_images(tmp_path):
+    """append_images should work to provide alternative sizes"""
+    im = hopper()
+    provided_im = Image.new("RGBA", (32, 32), (255, 0, 0, 255))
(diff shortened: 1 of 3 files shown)


src/PIL/IcoImagePlugin.py
```diff
@@ -63,9 +64,12 @@ def _save(im, fp, filename):
         fp.write(struct.pack("<H", 32))  # wBitCount(2)
 
         image_io = BytesIO()
-        # TODO: invent a more convenient method for proportional scalings
-        tmp = im.copy()
-        tmp.thumbnail(size, Image.LANCZOS, reducing_gap=None)
+        if size in alt_images:
+            tmp = alt_images[size]
+        else:
```
radarhere, line 69:
```suggestion
        tmp = alt_images.get(size)
        if not tmp:
```
Suggestion.

src/PIL/IcoImagePlugin.py
```diff
@@ -52,6 +52,7 @@ def _save(im, fp, filename):
     sizes = list(sizes)
     fp.write(struct.pack("<H", len(sizes)))  # idCount(2)
     offset = fp.tell() + len(sizes) * 16
+    alt_images = {im.size: im for im in im.encoderinfo.get("append_images", [])}
     for size in sizes:
```
nulano, line 56: It looks like if someone were to add an unusual sized image to append_images (e.g. (20,20)), it would not be saved unless it is added to the sizes parameter as well. I feel like this could be confusing behaviour. Perhaps adding `sizes.extend(alt_images.keys())` would be a good idea (sizes would have to be changed to a set to avoid duplicates).

ziplantil: That is correct. I believe the ICNS saving currently suffers from the same issue. I don't know if it should be fixed for .ICOs; the documentation would need to mention the role of append_images either way, since if the behavior is changed as you suggest, it would then also save other sizes than the ones the user requested (via the sizes parameter).

radarhere:
I've created https://github.com/ziplantil/Pillow/pull/1 with some suggestions.

Tests/test_file_ico.py
```diff
@@ -85,6 +85,20 @@ def test_only_save_relevant_sizes(tmp_path):
         assert im_saved.info["sizes"] == {(16, 16), (24, 24), (32, 32), (48, 48)}
 
 
+def test_only_save_append_images(tmp_path):
+    """append_images should work to provide alternative sizes"""
```
radarhere, line 89:
```suggestion
def test_save_append_images(tmp_path):
    # append_images should be used for scaled down versions of the image
```

radarhere (approved):

PR #4570, opened 2020-04-20
JPEG: make -1 and keep subsamplings identical for save

QSchulz:
Make the -1 subsampling behave like "keep" for the save method.

The doc claims that -1 is equivalent to keep. It is actually equivalent to 4:2:0 for some reason.

Note that -1 was representing both the -1 subsampling and the absence of it. To persist the current behavior which is compatible with multiple input file formats other than JPEG (which is not the case of the subsampling parameter), make the absence of subsampling parameter equal to -2.

get_sampling() remains unchanged since it is a public API.

Note that subsampling can be -1 and -2 in some cases, both results in an image with a subsampling of 4:2:0.

Also removing faulty test case since subsampling -1 cannot be used on a generated image but only JPEG file format.

 * im.save(subsampling=-1) behaves like im.save(subsampling="keep")
 * make a difference between -1 the picked subsampling parameter and the absence of subsampling parameter
 * remove one test case which cannot pass anymore with changed behavior for subsampling=-1
 * update doc to say 4:2:0 is the default when no subsampling is passed (current and future behavior)

Thanks!

### DIFF
--- a/Tests/test_file_jpeg.py
+++ b/Tests/test_file_jpeg.py
@@ -314,8 +314,6 @@ class TestFileJpeg:
             return layer[0][1:3] + layer[1][1:3] + layer[2][1:3]
 
         # experimental API
-        im = self.roundtrip(hopper(), subsampling=-1)  # default
-        assert getsampling(im) == (2, 2, 1, 1, 1, 1)
         im = self.roundtrip(hopper(), subsampling=0)  # 4:4:4
         assert getsampling(im) == (1, 1, 1, 1, 1, 1)
         im = self.roundtrip(hopper(), subsampling=1)  # 4:2:2
--- a/docs/handbook/image-file-formats.rst
+++ b/docs/handbook/image-file-formats.rst
@@ -334,7 +334,8 @@ The :py:meth:`~PIL.Image.Image.save` method supports the following options:
     If present, the image will be stored with the provided raw EXIF data.
 
 **subsampling**
-    If present, sets the subsampling for the encoder.
+    If present, sets the subsampling for the encoder. Otherwise, ``4:2:0`` is
+    the default subsampling.
 
     * ``keep``: Only valid for JPEG files, will retain the original image setting.
     * ``4:4:4``, ``4:2:2``, ``4:2:0``: Specific sampling values
--- a/src/PIL/JpegImagePlugin.py
+++ b/src/PIL/JpegImagePlugin.py
@@ -618,7 +618,7 @@ def _save(im, fp, filename):
     dpi = [round(x) for x in info.get("dpi", (0, 0))]
 
     quality = info.get("quality", -1)
-    subsampling = info.get("subsampling", -1)
+    subsampling = info.get("subsampling", -2)
     qtables = info.get("qtables")
 
     if quality == "keep":
@@ -628,13 +628,13 @@ def _save(im, fp, filename):
     elif quality in presets:
         preset = presets[quality]
         quality = -1
-        subsampling = preset.get("subsampling", -1)
+        subsampling = preset.get("subsampling", -2)
         qtables = preset.get("quantization")
     elif not isinstance(quality, int):
         raise ValueError("Invalid quality setting")
     else:
         if subsampling in presets:
-            subsampling = presets[subsampling].get("subsampling", -1)
+            subsampling = presets[subsampling].get("subsampling", -2)
         if isinstance(qtables, str) and qtables in presets:
             qtables = presets[qtables].get("quantization")
 
@@ -648,7 +648,7 @@ def _save(im, fp, filename):
         # For compatibility. Before Pillow 4.3, 4:1:1 actually meant 4:2:0.
         # Set 4:2:0 if someone is still using that value.
         subsampling = 2
-    elif subsampling == "keep":
+    elif subsampling == "keep" or subsampling == -1:
         if im.format != "JPEG":
             raise ValueError("Cannot use 'keep' when original image is not a JPEG")
         subsampling = get_sampling(im)


radarhere:
Okay, let me try and talk this through.

"keep" is not actually a single value. [Looking at the code](https://github.com/python-pillow/Pillow/blob/df886ed38694197665c8cce50216c877b477a30c/src/PIL/JpegImagePlugin.py#L651-L654), it tries to [get the existing sampling](https://github.com/python-pillow/Pillow/blob/df886ed38694197665c8cce50216c877b477a30c/src/PIL/JpegImagePlugin.py#L595-L606). If it can resolve it to one of 4:4:4, 4:2:2 or 4:2:0, it does so, but otherwise, it uses -1.

So at the moment, sometimes "keep" is the same as -1, sometimes it isn't.

What about the rest of the time? [-1 is actually not a setting at all, it's just the default behaviour](https://github.com/python-pillow/Pillow/blob/df886ed38694197665c8cce50216c877b477a30c/src/libImaging/JpegEncode.c#L180-L217).

So, I think there are two things to talk about here.

1. The default subsampling.

It sounds like you've experimentally found that the default subsampling is "4:2:0". However, as I've hopefully explained, Pillow hasn't explicitly set it to be that. Maybe you're using libjpeg-turbo, as https://github.com/libjpeg-turbo/libjpeg-turbo/blob/master/usage.txt talks about 'default subsampling level (2x2, AKA "4:2:0")'. I'm not sold that it's up to Pillow to document the behaviour of libjpeg variants though. If we wanted to say that the default is "4:2:0", then I think we should just change the Pillow default behaviour to tell libjpeg to use 4:2:0, rather than trusting libjpeg to do it. Feel free to disagree with me, but yes, I'm inclined to not update the documentation in that way.

2. The documentation says that -1 is equivalent to "keep", but this isn't always true

The effect of this PR on Pillow behaviour is that if you're explicitly passing in a subsampling of -1, then it becomes "keep", as per the documentation. This is a valid solution to the problem.

The way that I read the Pillow code, -1 is just an internal value, used to communicate between Python and C that no subsampling has been specified. My personal feeling is that if you want to have the default behaviour, just don't specify the subsampling, so I'm not convinced that -1 needed to explained to users at all. So an alternative solution would be to change the documentation to match the code, rather than changing the code to match the documentation.

I imagine that the number of people explicitly using the -1 value is small. Does the fact that Pillow values backwards compatibility point us towards keeping the behaviour the same and updating the documentation, or is it the problem the Pillow is not delivering on the promise to "keep" the subsampling when using the -1 value? I'm not sure about which choice is better, hopefully someone else has more thoughts on this.

QSchulz:
Thanks for taking the time to review this PR. Considering how much effort was put into your review, I should have put more context and explained a bit more. I'll try to remember that for next time.

> Okay, let me try and talk this through.
> 
> "keep" is not actually a single value. [Looking at the code](https://github.com/python-pillow/Pillow/blob/df886ed38694197665c8cce50216c877b477a30c/src/PIL/JpegImagePlugin.py#L651-L654), it tries to [get the existing sampling](https://github.com/python-pillow/Pillow/blob/df886ed38694197665c8cce50216c877b477a30c/src/PIL/JpegImagePlugin.py#L595-L606). If it can resolve it to one of 4:4:4, 4:2:2 or 4:2:0, it does so, but otherwise, it uses -1.
> 

That is my understanding as well.

> So, I think there are two things to talk about here.
> 
> 1. The default subsampling.
> 
> 
> It sounds like you've experimentally found that the default subsampling is "4:2:0". However, as I've hopefully explained, Pillow hasn't explicitly set it to be that. Maybe you're using libjpeg-turbo, as https://github.com/libjpeg-turbo/libjpeg-turbo/blob/master/usage.txt talks about 'default subsampling level (2x2, AKA "4:2:0")'. I'm not sold that it's up to Pillow to document the behaviour of libjpeg variants though. If we wanted to say that the default is "4:2:0", then I think we should just change the Pillow default behaviour to tell libjpeg to use 4:2:0, rather than trusting libjpeg to do it. Feel free to disagree with me, but yes, I'm inclined to not update the documentation in that way.
> 

Fully agree that if this is lib dependent, it shouldn't be documented. Or at least say that it's lib dependent but not mention the actual expected behaviour (which technically could change with a lib update).

`Otherwise, ``4:2:0`` is the default subsampling.` should thus be removed from my PR.

> 2. The documentation says that -1 is equivalent to "keep", but this isn't always true
> 
> 
> The effect of this PR on Pillow behaviour is that if you're explicitly passing in a subsampling of -1, then it becomes "keep", as per the documentation. This is a valid solution to the problem.
> 
> The way that I read the Pillow code, -1 is just an internal value, used to communicate between Python and C that no subsampling has been specified. My personal feeling is that if you want to have the default behaviour, just don't specify the subsampling, so I'm not convinced that -1 needed to explained to users at all. So an alternative solution would be to change the documentation to match the code, rather than changing the code to match the documentation.
> 

If -1 should represent the absence of parameter and is just internal, it shouldn't be documented that subsampling=-1 is the same as "keep". It shouldn't be documented at all, because that's an internal representation that shouldn't be made public. Now we have two different meaning for -1 value: absence of subsampling (whose behaviour isn't really documented) or subsampling=-1 (which is wrongly documented).

We could rework the doc to say that if subsampling is absent (or -1, if we want to keep -1 as a "public" value), the chosen subsampling is up to the system lib in use currently.

I also have an issue with -1 representing both the absence of subsampling and the possible -1 value, because we lose that information in the very first lines of the function.

> I imagine that the number of people explicitly using the -1 value is small. Does the fact that Pillow values backwards compatibility point us towards keeping the behaviour the same and updating the documentation, or is it the problem the Pillow is not delivering on the promise to "keep" the subsampling when using the -1 value? I'm not sure about which choice is better, hopefully someone else has more thoughts on this.

Either of them, I personally don't care. I just want the doc to reflect what the code is supposed to do (or vice-versa), which is not the case currently.

Let me know what we settle on so that I can send a v2.

radarhere:
I'm inclined to leave the code as is, and update the documentation instead. I've created PR #6067.

hugovk:
#6067 merged instead. Thanks anyway for the PR and raising this!

PR #4590, opened 2020-04-26
Added Python 3.9

radarhere:
Adds Python 3.9 Travis job.

Python 3.9 is still in alpha.


hugovk:
Thanks, the 3.9 beta is due out next month (2020-05-18) and 3.8 final on 2020-10-05.

https://www.python.org/dev/peps/pep-0596/

PR #4594, opened 2020-04-28
Fixes default offset for Exif

rodrigob:
test_imagefile.py checks exif instead of reloaded_exif; making the tests pass when they should not.


hugovk:
Thanks both!

PR #4596, opened 2020-04-29
Mention Ubuntu versions younger than 16.04

andreasnuesslein:
I suppose people will just try it either way but it's good to know that it still works with the same libs all the way through to 20.04

 * Update the docs to mention Ubuntu versions younger than 16.04

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -328,7 +328,7 @@ In Fedora, the command is::
 
 .. Note:: ``redhat-rpm-config`` is required on Fedora 23, but not earlier versions.
 
-Prerequisites for **Ubuntu 16.04 LTS or later** are installed with::
+Prerequisites for **Ubuntu 16.04 LTS - 20.04 LTS** are installed with::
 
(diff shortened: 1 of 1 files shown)


docs/installation.rst
```diff
@@ -328,7 +328,7 @@ In Fedora, the command is::
 
 .. Note:: ``redhat-rpm-config`` is required on Fedora 23, but not earlier versions.
 
-Prerequisites are installed on **Ubuntu 16.04 LTS** with::
+Prerequisites for **Ubuntu 16.04 LTS or later** are installed with::
```
radarhere, line 331:
```suggestion
Prerequisites for **Ubuntu 16.04 LTS - 20.04 LTS** are installed with::
```
Just a thought, since at some point in the future, these packages will stop being available for the latest Ubuntu.

andreasnuesslein: works for me.

radarhere:
Thanks

PR #3522, opened 2019-01-01
Improved ImageChops documentation

radarhere:
Resolves #2571

PR #4599, opened 2020-05-01
Fix ImageChops documentation

### LINKED ISSUE #2571: ImageChops: logical_and produces "Image has wrong mode" with 2 RGB images
### What did you do?
I tried to apply logical_and between two RGB images

AND function pixel-by-pixel, color-by-color

### What actually happened?
"Image has wrong mode" message. According to the docs, 
"At this time, most channel operations are only implemented for 8-bit images (e.g. “L” and “RGB”)."

Python 3.6.1, PIL 1.1.7

```python
from PIL import Image, ImageChops

def test_chops():
    img = Image.new("RGB", (100, 100), 0xffffff)
    new_img = ImageChops.logical_and(img, img)


def main():
    test_chops()
    return 0

main()
```

d0sboots:
Many methods were incorrectly documented as requiring mode "1". The remaining
ones require *both* images to be mode "1".

Documentation only, [ci skip]

Fixes #2571

### DIFF
--- a/src/PIL/ImageChops.py
+++ b/src/PIL/ImageChops.py
@@ -240,8 +240,10 @@ def subtract_modulo(image1, image2):
 def logical_and(image1, image2):
     """Logical AND between two images.
 
-    Both of the images must have mode "1". For an AND in RGB mode, use a
-    multiply() by a black-and-white mask.
+    Both of the images must have mode "1". If you would like to perform a
+    logical AND on an image with a mode other than "1", try
+    :py:meth:`~PIL.ImageChops.multiply` instead, using a black-and-white mask
+    as the second image.
 
     .. code-block:: python
 
--- a/src/PIL/ImageChops.py
+++ b/src/PIL/ImageChops.py
@@ -54,7 +54,7 @@ def invert(image):
 def lighter(image1, image2):
     """
     Compares the two images, pixel by pixel, and returns a new image containing
-    the lighter values. At least one of the images must have mode "1".
+    the lighter values.
 
     .. code-block:: python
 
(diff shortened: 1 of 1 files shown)


d0sboots:
I'm new-ish to GitHub pull workflow... is this waiting on some action from me?

hugovk:
@d0sboots Thanks for the PR! No action needed from you, just waiting for one of us maintainers to check it :)

radarhere:
Just to link, I'll note that this is reverting most of #3522

radarhere:
I agree that I just misunderstood `strcmp` with my PR. Thanks for correcting this.

src/PIL/ImageChops.py
```diff
@@ -244,8 +238,10 @@ def subtract_modulo(image1, image2):
 
 
 def logical_and(image1, image2):
-    """Logical AND between two images. At least one of the images must have
-    mode "1".
+    """Logical AND between two images.
+
+    Both of the images must have mode "1". For an AND in RGB mode, use a
+    multiply() by a black-and-white mask.
```
radarhere, line 244:
```suggestion
    Both of the images must have mode "1". If you would like to perform a
    logical AND on an image with a mode other than "1", try
    :py:meth:`~PIL.ImageChops.multiply` instead, using a black-and-white mask
    as the second image.
```

Suggestion. This advice isn't exclusive to RGB.

The ':py:meth' part is a way to link the method in the docs. For an example, see the [close method docstring](https://github.com/python-pillow/Pillow/blob/a28b2ea11c5c7e63adb9b29d136d412c862f7328/src/PIL/Image.py#L590) and the [way it is rendered](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.close).

d0sboots:
Thanks for the review!

Issue #4493, opened 2020-03-29
ClangFormat the C code


hugovk:
Run [ClangFormat](https://clang.llvm.org/docs/ClangFormat.html) on all of the C code to once and for all get rid of the mixed 8 tab/spaces -> all 4 spaces, get all the braces on a consistent style, enclose single lines in braces, and generally make it all consistent.

Re: https://github.com/python-pillow/Pillow/pull/1770#issuecomment-196974405

Will prevent bugs like https://github.com/python-pillow/Pillow/pull/4492.

Let's do this after the 2020-04-01 release (https://github.com/python-pillow/Pillow/issues/4354), when the PR list will be at a local minimum and to avoid merge conflicts.

There's four [pre-commit hooks](https://pre-commit.com/hooks.html) available, let's pick a good one to add to `.pre-commit-config.yaml` so the CI will prevent regressions.

We'll need to pick a [style](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) or define our own. Is there a predefined one close to the current style, or that is commonly used for Python extensions? What do other Python projects use?

hugovk:
I made 5 branches for each predefined ClangFormat style:

* https://github.com/python-pillow/Pillow/compare/master...hugovk:clang-format-chromium
* https://github.com/python-pillow/Pillow/compare/master...hugovk:clang-format-google
* https://github.com/python-pillow/Pillow/compare/master...hugovk:clang-format-llvm
* https://github.com/python-pillow/Pillow/compare/master...hugovk:clang-format-mozilla
* https://github.com/python-pillow/Pillow/compare/master...hugovk:clang-format-webkit

For each, I first generated the config file:

```sh
clang-format -style=chromium -dump-config > .clang-format
```

Then flipped the `SortIncludes` value from `true` to `false`, as it breaks the build otherwise. This can be looked into later.

Then changed any `*IndentWidth` value from 2 to 4 spaces.

And then finally ran the formatter:

```sh
clang-format -i `git ls-tree -r HEAD --name-only | grep "\.[ch]$"`
```

---

One thing to note, I wrote:

> Will prevent bugs like #4492.

This isn't actually true, ClangFormat doesn't insert brackets.

---

Any preferences?

nulano:
[...]

PR #4600, opened 2020-05-01
Replaced tabs with spaces in C code

radarhere:
Helps #4493

Not just a blanket find-and-replace, but also corrects indentation.


wiredfool:
Definitely needed something like this for ages.

Issue #4252, opened 2019-12-06
ICC Profile removed from TIFF when using any compression mode


joerambo:
### What did you do?

I'm trying to save a TIFF with 'tiff_lzw' compression and an ICC profile. When I set any compression mode other than 'raw' the ICC profile is removed completely. Other TIFF data set through tiffinfo is not stripped away.

### What did you expect to happen?

For the output TIFF to have an embedded ICC profile with compression.

### What actually happened?

Output TIFF has everything except for ICC profile.

### What are your OS, Python and Pillow versions?

* OS: macOS 10.13.6
* Python: 3.7.5
* Pillow: 6.2.1
* LibTIFF: 4.1.0

```python
from PIL import Image, TiffImagePlugin, features

TiffImagePlugin.DEBUG = True
print(f"LibTIFF enabled: {features.check('libtiff')}")

test_image_with_icc = Image.open('test_gm_icc.tiff')

ifd = TiffImagePlugin.ImageFileDirectory_v2()
ifd[TiffImagePlugin.ICCPROFILE] = test_image_with_icc.info.get("icc_profile")
ifd[TiffImagePlugin.ARTIST] = 'Test Artist'

print('--- saving lzw')
test_image_with_icc.save('test_pillow_lzw.tif', compression='tiff_lzw', tiffinfo=ifd)

print('--- saving raw')
test_image_with_icc.save('test_pillow_raw.tif', compression='raw', tiffinfo=ifd)
```

The example above uses an image with ICC profile already embedded, but same happens when I load an external .icc file. LZW drops it and RAW will add it in.

Inspecting LZW file shows tags from original and new tags added through tiffinfo except ICC:
```
$tiffdump test_pillow_lzw.tif

test_pillow_lzw.tif:
Magic: 0x4949 <little-endian> Version: 0x2a <ClassicTIFF>
Directory 0: offset 47410086 (0x2d36ba6) next 0 (0)
ImageWidth (256) SHORT (3) 1<6000>
ImageLength (257) SHORT (3) 1<4000>
BitsPerSample (258) SHORT (3) 3<8 8 8>
Compression (259) SHORT (3) 1<5>
Photometric (262) SHORT (3) 1<2>
DocumentName (269) ASCII (2) 47<./tests/output/test_200_ ...>
StripOffsets (273) LONG (4) 1<8>
SamplesPerPixel (277) SHORT (3) 1<3>
RowsPerStrip (278) SHORT (3) 1<4000>
[...]

PR #4605, opened 2020-05-03
Writing TIFF tags: improved BYTE, added UNDEFINED

### LINKED ISSUE #3677: Saving TIFF with compression throws Segmentation Fault 11 in 5.4.1, but not 5.3.0
### What did you do?
Opened a bitonal TIFF created with Adobe Photoshop CC (version 20.0.2) then saved it with Group4 compression using Pillow.

TIFF Image should be downloadable from Github repository here: [https://github.com/photosbyjeremy/images/blob/master/test_bitonal_text.tif.zip](https://github.com/photosbyjeremy/images/blob/master/test_bitonal_text.tif.zip)

I expected the image to save with compression.

### What actually happened?
In Pillow 5.4.1 there was a Segmentation Fault 11, but in Pillow 5.3.0 it works just fine.

screen shot 2019-02-25 at 2 09 47 pm

screen shot 2019-02-25 at 2 09 50 pm

* OS: macOS 10.14.3
* Python: 3.6.8 (and I actually tried 3.7.1 first, but didn't screenshot those attempts)
* Pillow: 5.4.1 and 5.3.0

```python
from PIL import Image
image = Image.open('test_bitonal_text.tif')
image.save('test_pillow_save.tif', compression='group4')
```

radarhere:
Resolves #3677

This PR has two changes.

Firstly, Pillow currently treats BYTE tags of count N as if they might have N strings. I do not believe this should be the case. To look at count values for character-related tags, [the BYTE tag GPSAltitudeRef](https://www.awaresystems.be/imaging/tiff/tifftags/privateifd/gps/gpsaltituderef.html) has count 1 for 1 character, [the ASCII tag GPSDateStamp](https://www.awaresystems.be/imaging/tiff/tifftags/privateifd/gps/gpsdatestamp.html) has count 11 for 11 characters, and [the UNDEFINED tag Flashpix Version](https://www.awaresystems.be/imaging/tiff/tifftags/privateifd/exif/flashpixversion.html) has count 4 for 4 characters. So, I think BYTE tags of count N have N characters, and are only single strings.

Secondly, this PR changes the following current code, to no longer force BYTE tags into ASCII tags.
https://github.com/python-pillow/Pillow/blob/d23df7227c419a010c4a82599856f83abd2633b5/src/encode.c#L732-L736

This second change was made because in trying to resolve the issue, I found that libtiff was throwing an error for the [XMLPACKET tag](https://www.awaresystems.be/imaging/tiff/tifftags/xmp.html). It seems reasonable that it would - it is a BYTE tag, and we are instead passing ASCII.

Together, these changes resolve the issue.

### DIFF
--- a/Tests/test_file_libtiff.py
+++ b/Tests/test_file_libtiff.py
@@ -299,13 +299,6 @@ class TestFileLibTiff(LibTiffTestCase):
                             )
                             continue
 
-                        if (
-                            libtiff
-                            and isinstance(value, bytes)
-                            and isinstance(tiffinfo, dict)
-                        ):
-                            value = value.decode()
-
                         assert reloaded_value == value
 
             # Test with types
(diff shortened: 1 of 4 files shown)


radarhere:
I've added another commit to this PR, since they both involve the same code, to resolve #4252. Pillow has been just been skipping UNDEFINED tags when writing with libtiff. This commit adds writing of UNDEFINED tags.

PR #4608, opened 2020-05-04
Added support for 1-D NumPy arrays

### LINKED ISSUE #4606: Image.fromarray() cannot handle 1-D arrays
Running Pillow ~~7.0.0~~ (EDIT: now 7.1.2) on Python 3.7.6 on Windows 10 x64

I am converting a float32 grayscale array to Pillow so that I can use some of the image operations. This is not strictly an actual image and I want to preserve the order of the grayscale values. Having to reshape the array to use these operations doesn't seem intended. The grayscale values are stored in index 3 of the dataset. When I try to load the array into the Pillow Image class, it throws an error because `__array_interface__['shape'][1]` is missing on a 1-D array. This is the Numpy intended behavior (see https://numpy.org/doc/stable/reference/generated/numpy.ndarray.shape.html). There is no check for this in `Image.fromarray()`.

Relevant code:
```python
rows = np.loadtxt (f, dtype = int, max_rows = 1)
columns = np.loadtxt (f, dtype = int, max_rows = 1)
inputPoints = np.loadtxt (f, delimiter=' ', max_rows = (rows * columns))
testArray = 255 * inputPoints[:,3]
print (testArray.__array_interface__)
print (len (testArray.__array_interface__['shape']))
grayImage = Image.fromarray (testArray, mode = 'L')
```

Output:

```python
{'data': (2080565403600, False), 'strides': None, 'descr': [('', '<f8')], 'typestr': '<f8', 'shape': (29118,), 'version': 3}
1
  File ".\test.py", line 333, in <module>
    grayImage = Image.fromarray (testArray, mode = 'L')
  File "D:\ProgramData\Anaconda3\envs\vtkvista37env\lib\site-packages\PIL\Image.py", line 2694, in fromarray
    size = shape[1], shape[0]
IndexError: tuple index out of range
```

Context in Image.py:

```python
    arr = obj.__array_interface__
    shape = arr["shape"]
    ndim = len(shape)
    strides = arr.get("strides", None)

    [some code for mode and dimensions]

    size = shape[1], shape[0]
```

I might suggest something like:

```python
    size = 1 if ndim == 1 else shape[1],  shape[0]
```

Thank you!

radarhere:
Resolves #4606

`fromarray` presumes that a NumPy array has a shape of at least 2 dimensions. This PR changes it so that if there is only 1 dimension, the other size dimension is 1, using the code suggestion from the issue.

### DIFF
--- a/Tests/test_numpy.py
+++ b/Tests/test_numpy.py
@@ -89,6 +89,11 @@ def test_3d_array():
     assert_image(Image.fromarray(a[:, :, 1]), "L", TEST_IMAGE_SIZE)
 
 
+def test_1d_array():
+    a = numpy.ones(5, dtype=numpy.uint8)
+    assert_image(Image.fromarray(a), "L", (1, 5))
+
+
 def _test_img_equals_nparray(img, np):
     assert len(np.shape) >= 2
     np_size = np.shape[1], np.shape[0]
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -2727,7 +2727,7 @@ def fromarray(obj, mode=None):
     if ndim > ndmax:
         raise ValueError("Too many dimensions: %d > %d." % (ndim, ndmax))
 
-    size = shape[1], shape[0]
+    size = 1 if ndim == 1 else shape[1], shape[0]
     if strides is not None:
         if hasattr(obj, "tobytes"):
             obj = obj.tobytes()


radarhere:
Looking at how 2d arrays are converted at the moment,

```
>>> from PIL import Image
>>> import numpy as np
>>> Image.fromarray(np.zeros((3, 4))).size
(4, 3)
>>> Image.fromarray(np.zeros((3, 1))).size
(1, 3)
```

This seems more expected to me

```
>>> Image.fromarray(np.zeros((3))).size
(1, 3)
```

Rather than this

```
>>> Image.fromarray(np.zeros((3))).size
(3, 1)
```

hugovk:
Thanks!

PR #4613, opened 2020-05-08
JPEG: Truncate icclist instead of setting to None

homm:
I faced with a slightly malformed file with two problems at once:

1. There was `APP2` marker with ICC profile after `SOF` marker which leads to exception on open:

```python
        elif marker == 0xFFE2 and s[:12] == b"ICC_PROFILE\0":
            # Since an ICC profile can be larger than the maximum size of
            # a JPEG marker (64K), we need provisions to split it into
            # multiple markers. The format defined by the ICC specifies
            # one or more APP2 markers containing the following data:
            #   Identifying string      ASCII "ICC_PROFILE\0"  (12 bytes)
            #   Marker sequence number  1, 2, etc (1 byte)
            #   Number of markers       Total of APP2's used (1 byte)
            #   Profile data            (remainder of APP2 data)
            # Decoders should use the marker sequence numbers to
            # reassemble the profile, rather than assuming that the APP2
            # markers appear in the correct sequence.
>           self.icclist.append(s)
E           AttributeError: 'NoneType' object has no attribute 'append'
```

2. The same `APP2` marker was included in the file twice: before and after `SOF`. Each `ICC_PROFILE` marker should include a marker sequence number and a number of markers. And both markers claim that they are the first and there is only one marker. They were literally identical. Despite this, viewers show one correct ICC profile.

I've reproduced both problems in the same test file. As we already ignore ICC profile markers after `SOF`, both problems have the same solution.


radarhere:
Minor note - I don't see how Problem 1 could happen without Problem 2. If the order was just SOF-ICC, then `self.icclist` would be empty when SOF is run, so SOF wouldn't set it to `None`. Because your test file has ICC-SOF-ICC, then `self.icclist` is populated first, SOF sets it to `None`, then when ICC runs again, the exception is thrown from trying to append to `None`.

homm:
> Minor note - I don't see how Problem 1 could happen without Problem 2.

Right, the first problem only possible with ICC-SOF-ICC sequence. And probably this is another bug: ICC is ignored after SOF. But the second problem is not related to SOF place, it happens for any ICC chanks with the same number.

PR #4614, opened 2020-05-08
Better _binary module use

homm:
* remove extra i8 calls where input is proved bytes[] or int. Most of i8 calls are not required in py3.
* use offset for all binary input functions instead of slicing which should be slightly faster especially when slicing "until the end" is used (`s[4:]`).

### DIFF
--- a/src/PIL/BmpImagePlugin.py
+++ b/src/PIL/BmpImagePlugin.py
@@ -47,7 +47,7 @@ def _accept(prefix):
 
 
 def _dib_accept(prefix):
-    return i32(prefix[:4]) in [12, 40, 64, 108, 124]
+    return i32(prefix) in [12, 40, 64, 108, 124]
 
 
 # =============================================================================
@@ -82,10 +82,10 @@ class BmpImageFile(ImageFile.ImageFile):
         # -------------------------------------------------- IBM OS/2 Bitmap v1
         # ----- This format has different offsets because of width/height types
         if file_info["header_size"] == 12:
-            file_info["width"] = i16(header_data[0:2])
-            file_info["height"] = i16(header_data[2:4])
-            file_info["planes"] = i16(header_data[4:6])
-            file_info["bits"] = i16(header_data[6:8])
+            file_info["width"] = i16(header_data, 0)
+            file_info["height"] = i16(header_data, 2)
+            file_info["planes"] = i16(header_data, 4)
+            file_info["bits"] = i16(header_data, 6)
             file_info["compression"] = self.RAW
             file_info["palette_padding"] = 3
 
@@ -94,22 +94,22 @@ class BmpImageFile(ImageFile.ImageFile):
         elif file_info["header_size"] in (40, 64, 108, 124):
             file_info["y_flip"] = header_data[7] == 0xFF
             file_info["direction"] = 1 if file_info["y_flip"] else -1
-            file_info["width"] = i32(header_data[0:4])
+            file_info["width"] = i32(header_data, 0)
             file_info["height"] = (
-                i32(header_data[4:8])
+                i32(header_data, 4)
                 if not file_info["y_flip"]
-                else 2 ** 32 - i32(header_data[4:8])
+                else 2 ** 32 - i32(header_data, 4)
             )
-            file_info["planes"] = i16(header_data[8:10])
-            file_info["bits"] = i16(header_data[10:12])
-            file_info["compression"] = i32(header_data[12:16])
+            file_info["planes"] = i16(header_data, 8)
+            file_info["bits"] = i16(header_data, 10)
+            file_info["compression"] = i32(header_data, 12)
             # byte size of pixel data
-            file_info["data_size"] = i32(header_data[16:20])
+            file_info["data_size"] = i32(header_data, 16)
             file_info["pixels_per_meter"] = (
-                i32(header_data[20:24]),
-                i32(header_data[24:28]),
+                i32(header_data, 20),
+                i32(header_data, 24),
             )
-            file_info["colors"] = i32(header_data[28:32])
+            file_info["colors"] = i32(header_data, 28)
             file_info["palette_padding"] = 4
             self.info["dpi"] = tuple(
                 int(x / 39.3701 + 0.5) for x in file_info["pixels_per_meter"]
@@ -119,7 +119,7 @@ class BmpImageFile(ImageFile.ImageFile):
                     for idx, mask in enumerate(
                         ["r_mask", "g_mask", "b_mask", "a_mask"]
                     ):
-                        file_info[mask] = i32(header_data[36 + idx * 4 : 40 + idx * 4])
+                        file_info[mask] = i32(header_data, 36 + idx * 4)
                 else:
                     # 40 byte headers only have the three components in the
                     # bitfields masks, ref:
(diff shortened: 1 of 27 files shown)


src/PIL/ImImagePlugin.py
```diff
@@ -223,14 +222,14 @@ def _open(self):
...
-                    if i8(palette[i]) != i:
+                    if palette[i] != i:
                         linear = 0
                 else:
                     greyscale = 0
             if self.mode in ["L", "LA", "P", "PA"]:
                 if greyscale:
                     if not linear:
-                        self.lut = [i8(c) for c in palette[:256]]
+                        self.lut = list(palette[:256])
```
homm, line 232:
```python
In [1]: [c for c in b'test']
Out[1]: [116, 101, 115, 116]

In [2]: list(b'test')
Out[2]: [116, 101, 115, 116]
```

homm:
I don't have enough permissions to restart build on Travis (

hugovk:
> I don't have enough permissions to restart build on Travis (

Maybe you do: try logging out and back in to Travis, that brought back the restart buttons for me recently.

homm:
Thanks, it works)

src/PIL/BmpImagePlugin.py
```diff
@@ -82,34 +82,34 @@ def _bitmap(self, header=0, offset=0):
         # -------------------------------------------------- IBM OS/2 Bitmap v1
         # ----- This format has different offsets because of width/height types
         if file_info["header_size"] == 12:
-            file_info["width"] = i16(header_data[0:2])
-            file_info["height"] = i16(header_data[2:4])
-            file_info["planes"] = i16(header_data[4:6])
-            file_info["bits"] = i16(header_data[6:8])
+            file_info["width"] = i16(header_data, 0)
```
radarhere, line 85:
```suggestion
            file_info["width"] = i16(header_data)
```
Zero is the default value for the offset. Throwing the idea out there of not explicitly setting the argument?

src/PIL/BmpImagePlugin.py
```diff
@@ -82,34 +82,34 @@ def _bitmap(self, header=0, offset=0):
...
             file_info["palette_padding"] = 3
 
         # --------------------------------------------- Windows Bitmap v2 to v5
         # v3, OS/2 v2, v4, v5
         elif file_info["header_size"] in (40, 64, 108, 124):
-            file_info["y_flip"] = i8(header_data[7]) == 0xFF
+            file_info["y_flip"] = header_data[7] == 0xFF
             file_info["direction"] = 1 if file_info["y_flip"] else -1
-            file_info["width"] = i32(header_data[0:4])
+            file_info["width"] = i32(header_data, 0)
```
radarhere, line 97:
```suggestion
            file_info["width"] = i32(header_data)
```

src/PIL/EpsImagePlugin.py
```diff
@@ -312,14 +312,14 @@ def _find_offset(self, fp):
             fp.seek(0, io.SEEK_END)
             length = fp.tell()
             offset = 0
-        elif i32(s[0:4]) == 0xC6D3D0C5:
+        elif i32(s, 0) == 0xC6D3D0C5:
```
radarhere, line 315:
```suggestion
        elif i32(s) == 0xC6D3D0C5:
```

src/PIL/GbrImagePlugin.py
```diff
@@ -29,7 +29,7 @@
 
 
 def _accept(prefix):
-    return len(prefix) >= 8 and i32(prefix[:4]) >= 20 and i32(prefix[4:8]) in (1, 2)
+    return len(prefix) >= 8 and i32(prefix, 0) >= 20 and i32(prefix, 4) in (1, 2)
```
radarhere, line 32:
```suggestion
    return len(prefix) >= 8 and i32(prefix) >= 20 and i32(prefix, 4) in (1, 2)
```

src/PIL/GifImagePlugin.py
```diff
@@ -232,12 +232,12 @@ def _seek(self, frame):
                 s = self.fp.read(9)
 
                 # extent
-                x0, y0 = i16(s[0:]), i16(s[2:])
-                x1, y1 = x0 + i16(s[4:]), y0 + i16(s[6:])
+                x0, y0 = i16(s, 0), i16(s, 2)
```
radarhere, line 235:
```suggestion
                x0, y0 = i16(s), i16(s, 2)
```

src/PIL/JpegImagePlugin.py
```diff
@@ -121,10 +121,10 @@ def APP(self, marker):
                 data = s[offset : offset + size]
                 if code == 0x03ED:  # ResolutionInfo
                     data = {
-                        "XResolution": i32(data[:4]) / 65536,
-                        "DisplayedUnitsX": i16(data[4:8]),
-                        "YResolution": i32(data[8:12]) / 65536,
-                        "DisplayedUnitsY": i16(data[12:]),
+                        "XResolution": i32(data, 0) / 65536,
```
radarhere, line 124:
```suggestion
                        "XResolution": i32(data) / 65536,
```

src/PIL/PngImagePlugin.py
```diff
@@ -369,14 +369,14 @@ def chunk_IHDR(self, pos, length):
 
         # image header
         s = ImageFile._safe_read(self.fp, length)
-        self.im_size = i32(s), i32(s[4:])
+        self.im_size = i32(s, 0), i32(s, 4)
```
radarhere, line 372:
```suggestion
        self.im_size = i32(s), i32(s, 4)
```

src/PIL/PngImagePlugin.py
```diff
@@ -450,15 +450,15 @@ def chunk_sRGB(self, pos, length):
...
+        self.im_info["srgb"] = s[0]
         return s
 
     def chunk_pHYs(self, pos, length):
 
         # pixels per unit
         s = ImageFile._safe_read(self.fp, length)
-        px, py = i32(s), i32(s[4:])
-        unit = i8(s[8])
+        px, py = i32(s, 0), i32(s, 4)
```
radarhere, line 460:
```suggestion
        px, py = i32(s), i32(s, 4)
```

src/PIL/BmpImagePlugin.py
nulano: I think specifying the offset explicitly is easier to read in this case. I think the only time I would omit the offset is when the whole value is to be converted.

src/PIL/EpsImagePlugin.py
wiredfool: I’m in favor of being explicit for clarity of what is being converted.

src/PIL/IcnsImagePlugin.py
```diff
@@ -70,7 +69,7 @@ def read_32(fobj, start_length, size):
                 byte = fobj.read(1)
                 if not byte:
                     break
-                byte = i8(byte)
+                byte = byte[0]
```
radarhere, line 72: Is it just me, or is `byte = byte[0]` not as clear as `byte = i8(byte)` about what is happening here?

homm: You are right that `i8()` looks more clear at first glance. Mostly because indexing had changed from Py2 to Py3 for bytes. However, this saves one function call and type checking on each value. Also, IDEs handle types correctly and show variables as int.

homm: As an alternative, it is possible to change i8 and use it everywhere for consistency:

```python
def i8(c, o=0):
    return c[o]
```

radarhere:
I've created a rebase version of this as #5156

Issue #4611, opened 2020-05-07
ImageGrab.grabclipboard() for RGBA images on Windows?


teamclouday:
### What did you do?
I'm trying to read RGBA image from Windows clipboard, and convert the image to RGB format with a white background. The RGBA image was copyed from chrome browser.  

### What did you expect to happen?
The returned image from ```ImageGrab.grabclipboard()``` should be in RGBA mode.  

### What actually happened?
The returned image from ```ImageGrab.grabclipboard()``` is already in RGB mode, with black background.  

### What are your OS, Python and Pillow versions?

* OS: Windows 10
* Python: 3.6.8
* Pillow: 7.0.0

```python
new_image = ImageGrab.grabclipboard()
new_image_converted = Image.new("RGB", new_image.size, (255, 255, 255))
new_image_converted.paste(new_image)
```

teamclouday:
I think the problem comes from this line of code: https://github.com/python-pillow/Pillow/blob/a9cd55bc3990da462a5cfdb2798d706bf0057f98/src/display.c#L484  
Seems that ```CF_DIB``` does not support the alpha channel.  
Is it possible to use ```CF_DIBV5```?  
[link](https://docs.microsoft.com/en-us/windows/win32/dataxchg/standard-clipboard-formats)

teamclouday:
It seems that Windows automatically convert RGBA image from chrome to RGB image in clipboard.  
So it is not the problem with PIL.

nulano:
Chrome does actually copy RGBA images properly, I can paste them just fine in Paint.NET.
Reading a few StackOverflow questions, it seems that there are multiple options for how applications store RGBA images in the clipboard:
* `CF_DIB` is supposed to be RGBX according to the docs (RGB with padding), but some applications treat it as RGBA.
* `CF_DIBV5` supports multiple formats, including RGBA (the documentation is a bit ambiguous), or even PNG.
[...]

PR #4615, opened 2020-05-09
Improve ImageGrab.grabclipboard() on Windows

nulano:
Helps #4611

Adds support for the following clipboard formats on Windows in addition to the already supported `CF_DIB`:

 * `CF_DIBV5`
 * `'PNG'` used by e.g. MS Word 2016 and Paint.NET,
 * `CF_HDROP` used by most browsers (except Chrome), file explorers, etc.

*Support for `CF_HDROP` was already hinted at in the documentation, but there was merely a FIXME in place of the code.*

Used format is selected out of available formats based on the priority specified by the clipboard owner.

While I could see the argument that CF_HDROP is more complicated and should therefore only be a fallback, based on my testing, it is almost always the preferred format, and preserves the original image data (e.g. transparency, tags, ...), unlike bitmaps.

### DIFF
--- a/src/PIL/ImageGrab.py
+++ b/src/PIL/ImageGrab.py
@@ -93,15 +93,20 @@ def grabclipboard():
         os.unlink(filepath)
         return im
     elif sys.platform == "win32":
-        import io
-
         fmt, data = Image.core.grabclipboard_win32()
-        if isinstance(data, str):
-            if fmt == "file":
-                with open(data, "rb") as f:
-                    im = Image.open(io.BytesIO(f.read()))
-                return im
+        if fmt == "file":  # CF_HDROP
+            import struct
+
+            o = struct.unpack_from("I", data)[0]
+            if data[16] != 0:
+                files = data[o:].decode("utf-16le").split("\0")
+                return files[: files.index("")]
+            else:
+                files = data[o:].decode("mbcs").split("\0")
+                return files[: files.index("")]
         if isinstance(data, bytes):
+            import io
+
             data = io.BytesIO(data)
             if fmt == "png":
                 from . import PngImagePlugin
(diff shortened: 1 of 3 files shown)


src/PIL/ImageGrab.py
```diff
@@ -93,12 +93,29 @@ def grabclipboard():
...
+        if fmt == "file":  # CF_HDROP
+            import struct
+
+            o = struct.unpack_from("I", data)[0]
+            if data[16] != 0:
+                files = data[o:].decode("utf-16le").split("\0")
+                return files[: files.index("")]
+            else:
+                files = data[o:].decode("mbcs").split("\0")
+                return files[: files.index("")]
```
radarhere, line 106:
```suggestion
            else:
                files = data[o:].decode("mbcs").split("\0")
            return files[: files.index("")]
```
Replaced two identical lines with one.

nulano: Good idea.

src/PIL/ImageGrab.py
```diff
@@ -93,12 +93,28 @@ def grabclipboard():
...
+            data = io.BytesIO(data)
+            if fmt == "png":
+                from . import PngImagePlugin
+
+                return PngImagePlugin.PngImageFile(data)
+            elif fmt == "DIB":
+                from . import BmpImagePlugin
+
+                return BmpImagePlugin.DibImageFile(data)
+        return None
```
radarhere, line 118: This might be just a curious question - why change this from `return data` to `return None`?

nulano: It should be dead code in both cases, I just think None is a safer fallback then returning undefined data.

hugovk:
Thanks!

PR #4617, opened 2020-05-10
Removed duplicate return statements

radarhere:


hugovk:
Good spot! These have been there since at least PIL 1.1.1 (Oct 20, 2000)!

https://github.com/hugovk/PIL/blame/master/libImaging/Except.c#L45-L53

PR #4618, opened 2020-05-10
Added braces to C

radarhere:
See #4493

### DIFF
--- a/src/libImaging/GetBBox.c
+++ b/src/libImaging/GetBBox.c
@@ -36,7 +36,7 @@ ImagingGetBBox(Imaging im, int bbox[4])
 #define GETBBOX(image, mask)\
     for (y = 0; y < im->ysize; y++) {\
         has_data = 0;\
-        for (x = 0; x < im->xsize; x++)\
+        for (x = 0; x < im->xsize; x++) {\
             if (im->image[y][x] & mask) {\
                 has_data = 1;\
                 if (x < bbox[0])\
@@ -44,6 +44,7 @@ ImagingGetBBox(Imaging im, int bbox[4])
                 if (x >= bbox[2])\
                     bbox[2] = x+1;\
             }\
+        }\
         if (has_data) {\
             if (bbox[1] < 0)\
             bbox[1] = y;\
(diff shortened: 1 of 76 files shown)


src/libImaging/GetBBox.c
```diff
@@ -94,21 +96,24 @@ ImagingGetProjection(Imaging im, UINT8* xproj, UINT8* yproj)
...
                     xproj[x] = 1;\
                 }\
-                if (has_data)\
-                    yproj[y] = 1;\
-                }
+            }\
+            if (has_data) {\
+                yproj[y] = 1;\
+            }\
+        }
```
hugovk, line 108: Just a note.

Before:
Visually, both `if` blocks are indented as if they are both in the `for` loop.

But it wouldn't make sense for both to be in the `for` loop, so it was functionally correct before, and the change makes it clearer.

src/libImaging/JpegEncode.c
```diff
@@ -261,17 +264,19 @@ ImagingJpegEncode(Imaging im, ImagingCodecState state, UINT8* buf, int bytes)
...
                 context->extra_offset += n;
-                if (context->extra_offset >= context->extra_size)
+                if (context->extra_offset >= context->extra_size) {
                     state->state++;
-                else
+                } else {
                     break;
+                }
             } else
                 state->state++;
```
hugovk, line 281:
```suggestion
            } else {
                state->state++;
            }
```

hugovk:
A few other `else`s:

```console
grep -r " else$" . --include "*.c"
./src/libImaging/BoxBlur.c:    else
./src/libImaging/BoxBlur.c:    else
./src/libImaging/BoxBlur.c:    else
./src/libImaging/RawEncode.c:        } else
./src/libImaging/JpegEncode.c:            } else
./src/libImaging/XbmEncode.c:                } else
./src/libImaging/GifDecode.c:    } else
./src/_webp.c:        } else
./src/_webp.c:        } else
```

The `if`s are harder to `grep`, this finds some plus many false positives:

`grep -r " if.*[^{]$" . --include "*.c"`

radarhere:
Ok, I've done a pass over the code using your `grep` commands, and pushed the commit.

PR #4620, opened 2020-05-11
Fix repeatedly loading .gbr

ElinksFr:
.gbr files are not usable since on each `load()` we try to load the image from the file buffer.
This lead to an error on the second `load()` since the buffer is empty (we reached the end of the file).

 * read from the file only if `self.im is None`
 * a basic test doing 2 `crop` in a row that would crash without the fix

### DIFF
--- a/Tests/test_file_gbr.py
+++ b/Tests/test_file_gbr.py
@@ -15,3 +15,10 @@ def test_gbr_file():
     with Image.open("Tests/images/gbr.gbr") as im:
         with Image.open("Tests/images/gbr.png") as target:
             assert_image_equal(target, im)
+
+
+def test_multiples_operation():
+    with Image.open("Tests/images/gbr.gbr") as im:
+        rect = (0, 0, 10, 10)
+        im.crop(rect)
+        im.crop(rect)
--- a/Tests/test_file_gbr.py
+++ b/Tests/test_file_gbr.py
@@ -17,8 +17,9 @@ def test_gbr_file():
             assert_image_equal(target, im)
 
 
-def test_multiples_operation():
+def test_multiple_load_operations():
     with Image.open("Tests/images/gbr.gbr") as im:
-        rect = (0, 0, 10, 10)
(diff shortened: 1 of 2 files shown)


Tests/test_file_gbr.py
```diff
@@ -15,3 +15,10 @@ def test_gbr_file():
     with Image.open("Tests/images/gbr.gbr") as im:
         with Image.open("Tests/images/gbr.png") as target:
             assert_image_equal(target, im)
+
+
+def test_multiples_operation():
+    with Image.open("Tests/images/gbr.gbr") as im:
+        rect = (0, 0, 10, 10)
+        im.crop(rect)
+        im.crop(rect)
```
radarhere, line 24:
```suggestion
        im.load()
        im.load()
```
Simplify the test case?

Tests/test_file_gbr.py
```diff
@@ -15,3 +15,10 @@ def test_gbr_file():
     with Image.open("Tests/images/gbr.gbr") as im:
         with Image.open("Tests/images/gbr.png") as target:
             assert_image_equal(target, im)
+
+
+def test_multiples_operation():
```
radarhere, line 20:
```suggestion
def test_multiple_load_operations():
```

Tests/test_file_gbr.py
ElinksFr: As you want. I am not fine connoiseur of this library.

I used `crop` as an example instead of `load` because of the lazy loading (and because in my use case I do not call `load()` explicitly)

But since the fix is clearly on the the `load` method it would propably be clearer.

radarhere: Ok, sure. I've pushed some changes, including this, and also rearranged your change to the plugin itself to match some of our existing code -
https://github.com/python-pillow/Pillow/blob/bc1dc14b592dbc7d303bc7a82daa13fcdc547369/src/PIL/IcoImagePlugin.py#L291-L294

PR #4622, opened 2020-05-13
Fix broken string in test error message

hugovk:
Hasn't cropped up because the test is passing :)

# Demo

```python
average_diff = 2
max_average_diff = 1
band = "R"

# broken
msg = "average pixel value difference {:.4f} > expected {:.4f} "
"for '{}' band".format(average_diff, max_average_diff, band)

print(msg)

# fixed
msg = (
    "average pixel value difference {:.4f} > expected {:.4f} "
    "for '{}' band".format(average_diff, max_average_diff, band)
)

print(msg)
```
```
average pixel value difference {:.4f} > expected {:.4f}
average pixel value difference 2.0000 > expected 1.0000 for 'R' band
```

### DIFF
--- a/Tests/test_image_reduce.py
+++ b/Tests/test_image_reduce.py
@@ -174,8 +174,10 @@ def assert_compare_images(a, b, max_average_diff, max_diff=255):
         average_diff = sum(i * num for i, num in enumerate(ch_hist)) / (
             a.size[0] * a.size[1]
         )
-        msg = "average pixel value difference {:.4f} > expected {:.4f} "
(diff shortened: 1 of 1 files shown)


radarhere:
Nice catch

hugovk:
Found by the new Flake8 3.8.

There's a few other minor things too, but I've not updated those (or pre-commit) yet as Black doesn't like it, so they can wait a bit.

https://github.com/psf/black/issues/195#issuecomment-627783652

PR #4628, opened 2020-05-16
Added is_animated and n_frames to all images

### LINKED ISSUE #4609: is_animated should be defined for all images
### What did you do?

I have a user with a photo. I want to check if the photo is animated. I ran the code:
```python
from PIL import Image

if (Image.open(user.photo.file).is_animated):
````

I expect `is_animated` to be defined for all images.
https://stackoverflow.com/a/50808096/1412564
https://github.com/python-pillow/Pillow/pull/1319

### What actually happened?
If the image if not a GIF, I got an exception:

'JpegImageFile' object has no attribute 'is_animated'

* OS: Ubuntu 18.04
* Python: 3.6
* Pillow: 7.1.2

### PR DESCRIPTION
**Pull request by radarhere**
Resolves #4609 

Sets `is_animated` and `n_frames` for the base `Image` class so that they are present for all images.

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -538,6 +538,10 @@ class Image:
         self.readonly = 0
         self.pyaccess = None
         self._exif = None
+        if "is_animated" not in dir(self):
+            self.is_animated = False
+        if "n_frames" not in dir(self):
+            self.n_frames = 1
 
     @property
     def width(self):
--- a/Tests/test_image.py
+++ b/Tests/test_image.py
@@ -65,6 +65,8 @@ class TestImage:
         assert repr(im)[:45] == "<PIL.Image.Image image mode=L size=100x100 at"
         assert im.mode == "L"
         assert im.size == (100, 100)
+        assert not im.is_animated
+        assert im.n_frames == 1
 
         im = Image.new("RGB", (100, 100))
         assert repr(im)[:45] == "<PIL.Image.Image image mode=RGB size=100x100 "

### CODE REVIEW

File: src/PIL/Image.py [hunk 1]
```diff
@@ -538,6 +538,10 @@ def __init__(self):
         self.readonly = 0
         self.pyaccess = None
         self._exif = None
+        if "is_animated" not in dir(self):
+            self.is_animated = False
+        if "n_frames" not in dir(self):
+            self.n_frames = 1
```
radarhere, line 544: These properties can't just be always set regardless of whether or not they are already present, because for [some formats, they are property decorators without setters](https://github.com/python-pillow/Pillow/blob/4634eafe3c695a014267eefdce830b4a825beed7/src/PIL/GifImagePlugin.py#L92-L102).

hugovk:
Are there any in there which _can_ be animated but don't currently have `is_animated`? 

* https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html

For example, `(getattr(image, "is_animated", False)` could be returning `False` for an animated format when we just haven't implemented support. There's some old and obscure ones in there.

You gave the example of FitsStubImagePlugin in https://github.com/python-pillow/Pillow/issues/4609#issuecomment-629570395. Would we be claiming animated FITS images are not animated?

radarhere:
Yes, I guess we would be claiming that.

I was persuaded by the fact that `seek()` and `tell()` exist for all images already. Seeking past the first frame and throwing an `EOFError` already indicates something about the number of images in the file. Maybe if we could go back and do things differently, that wouldn't be the case though.

Happy for this to be closed if that is the stance we want to take.

nulano:
I have created #4739 as an alternative to instead document the current situation.

PR #4630, opened 2020-05-16
Use _accept function in example plugin

radarhere:
For an example in existing code, see
https://github.com/python-pillow/Pillow/blob/bc1dc14b592dbc7d303bc7a82daa13fcdc547369/src/PIL/GifImagePlugin.py#L39-L40
and
https://github.com/python-pillow/Pillow/blob/bc1dc14b592dbc7d303bc7a82daa13fcdc547369/src/PIL/GifImagePlugin.py#L875

### DIFF
--- a/docs/handbook/writing-your-own-file-decoder.rst
+++ b/docs/handbook/writing-your-own-file-decoder.rst
@@ -17,8 +17,8 @@ itself. Such plug-ins usually have names like
 Pillow decodes files in 2 stages:
 
 1. It loops over the available image plugins in the loaded order, and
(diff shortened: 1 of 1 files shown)


hugovk:
From the same file, should `accept` be `_accept`?

> 1. It loops over the available image plugins in the loaded order, and calls the plugin's `accept` function with the first 16 bytes of the file. If the `accept` function returns true, the plugin's `_open` method is called to set up the image metadata and image tiles. The `_open` method is not for decoding the actual image data.

radarhere:
Good call, yes, that change would potentially reduce confusion.

PR #4642, opened 2020-05-23
Move MSYS2 to GitHub Actions

### LINKED ISSUE #2848: python setup.py install: Jpeg2KDecode.c: undefined reference to `opj_stream_create'
### What did you do?

Attempted to install Pillow on Windows using `setup.py` in a MSYS2 (MINGW32) environment
```
$ python setup.py install
```
### What actually happened?
Compilation failed with:
```
build/temp.mingw-2.7/libImaging/Jpeg2KEncode.o:Jpeg2KEncode.c:(.text+0xfd9): undefined reference to `opj_end_compress'
collect2.exe: error: ld returned 1 exit status
error: command 'C:\\msys32\\mingw32\\bin/i686-w64-mingw32-gcc.exe' failed with exit status 1
```
[pillow-compilation.log](https://github.com/python-pillow/Pillow/files/1464226/pillow-compilation.log)

* Pillow - `git clone https://github.com/python-pillow/Pillow`
* Python 2.7.13

nulano:
For #3606 and https://github.com/python-pillow/Pillow/pull/4495#issuecomment-605699021.

Move MSYS2 from AppVeyor to faster GHA, now that it is available.

 * Move MSYS2/MinGW from AppVeyor to faster GHA.
 * Use `_WIN32` instead of `_MSC_VER` in `_imagingft.c` to fix FreeType compile.
 * Add `libraqm-0` as alternate Raqm filename for MinGW.
* Don't add `OPJ_STATIC` on MinGW to fix #2848.  Inspired by the patch from MSYS2 repo: https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-python-pillow/002-shared-openjpeg.patch

TODO: `libtiff` is failing to open files for some reason (I tried both 4.1.0 and 4.0.10), disabled for now.
*Looking at the MSYS2 repo, they don't seem to have a patch for `libtiff`, yet somehow that version passes tests.* Edit: https://trac.wxwidgets.org/ticket/17286 may be related.


Tests/test_imagefont.py
```diff
@@ -660,6 +661,7 @@ def test_variation_get(self):
             {"name": b"Size", "minimum": 0, "maximum": 300, "default": 0}
         ]
 
+    @pytest.mark.skipif(is_mingw(), reason="epsilon too high for meaningful test")
```
nulano, line 664: I'm not sure what is causing the difference, but I am getting epsilon around 33 for both variation tests. Checking manually, I see there is a significantly different kerning after the first letter, but the output looks correct to me.

Note: MSYS2 is currently using FreeType version 2.10.2.

hugovk: Please could you attach images for each case?

nulano: Sure.

test_variation_set_by_name:
a.png (MSYS):
a

b.png (target - variation_adobe.png):
b


test_variation_set_by_axes
a.png (MSYS):
a

b.png (target - variation_adobe_axes.png):
b

Basic and Raqm layout are pixel identical. In fact, it looks like there is a bug in the tests where the used `layout_engine` is not being explicitly set and defaults to Raqm in both cases.

Tests/helper.py
```diff
@@ -288,6 +288,12 @@ def is_pypy():
     return hasattr(sys, "pypy_translation_info")
 
 
+def is_mingw():
+    import sysconfig
```
hugovk, line 292: Let's import this at the top with the others.

nulano:
I decided to test Cygwin, and aside from Raqm (which was not found) and two test which should have been skipped, everything passes on this PR. It turns out that `libraqm-0` is the correct name for Cygwin, but _WIN32 is not defined on Cygwin. Adding that name to the `dlopen` branch finds the file, but for some reason Pillow is still refusing to use it.

hugovk:
Thank you!

Issue #4651, opened 2020-05-27
ImageFont module: getsize is slow?


cristianocca:
### What did you do?
See code sample.

### What did you expect to happen?
`getsize` should be fast. I'm trying to use this to implement a text-wrap mechanism (for SVGs), but the moment I add a single call to this in a for loop, timers go up 10 to 20 times slower.

Calling `getsize` on small text multiple times adds a significant overhead.  I wonder if it can be improved somehow, or it is really as good as it can get.  For example, 50 calls takes about 50ms (with short text), and increases significantly the bigger the text.

### What actually happened?
`getsize` is slow.

### What are your OS, Python and Pillow versions?

* OS: Windows
* Python: 3.7
* Pillow: 7.1.2

```python
font = ImageFont.truetype(font_file, font_size)

# Run the following in a loop
font.getsize(text)
```

nulano:
You are using Windows, so I'm assuming your default `layout_engine` is basic layout. With a quick look at the code, I have noticed that basic layout renders all glyphs before looking at their advance values. I wonder whether this is actually necessary?
https://github.com/python-pillow/Pillow/blob/a78de17346da722bc015cf35af0189adcdc0e2b9/src/_imagingft.c#L556
 I would assume the advance value only depends on hinting, which should probably run without the `FT_LOAD_RENDER` flag. In fact, the FreeType docs mention the flag `FT_LOAD_BITMAP_METRICS_ONLY` which seems appropriate here and conflicts with `FT_LOAD_RENDER`. https://www.freetype.org/freetype2/docs/reference/ft2-base_interface.html#ft_load_xxx

cristianocca:
I'm pretty sure I observed the same speed on MacOS. I wonder if this is different/faster on Linux distros?

nulano:
On Linux and Mac it is more likely that you have Raqm installed; this is rare on Windows. The default behaviour if you don't explicitly set `ImageFont.truetype(layout_engine=...)` is to use Raqm if available, and basic layout otherwise. The two layout engines use a completely different code path.

nulano:
Testing with `FreeMono.ttf` at size 20 for text "Hello world" I get a 15% improvement in `getsize` by swapping the flags as I mentioned above, with a pixel-identical image. The change also passes the test suite. I'll submit a PR for this. I do not see any other potential changes that would improve `getsize` speed without a significant redesign.

radarhere:
[...]

Issue #891, opened 2014-09-07
Error when using bitmap font embedded inside truetype font.


jackyyf:
Truetype font has a feature which allows you to embed bitmap font for specific font size, to optimize the rendering results on screens. But using such fonts in Pillow will generate some strange result.

Example:
![Bitmap font vs TrueType font](http://i.imgur.com/tzJlIvp.png)

In this case I'm using a modified version of SimSun, which contains embedded bitmap fonts, for all size between 8px and 24px. In the image, first line is using 24px font and the second line is 28px.

[TestCase](https://dl.dropboxusercontent.com/u/93107683/Pillow-issue.tar.bz2) is available (Contains SimSun font, testcode and the result above).
This test can be reproduced on both windows and linux.

PR #4652, opened 2020-05-27
Do not render glyphs during text layout

nulano:
For #4651 

> I would assume the advance value only depends on hinting, which should probably run without the FT_LOAD_RENDER flag. In fact, the FreeType docs mention the flag FT_LOAD_BITMAP_METRICS_ONLY which seems appropriate here and conflicts with FT_LOAD_RENDER. https://www.freetype.org/freetype2/docs/reference/ft2-base_interface.html#ft_load_xxx

~~While the `FT_LOAD_BITMAP_METRICS_ONLY` will likely be ignored due to `FT_LOAD_NO_BITMAP`, it should serve as a reminder for when #891 is resolved properly. *(I know the solution, I intend to submit a PR later. Essentially, Pillow should check the bitdepth of the returned bitmap instead of guessing.)*~~ Edit: Removed for now as it is not supported on all FreeType versions.

Testing I get the following results (in s) on my Windows 10 laptop with the code below.
| 7.1.2 (PyPI) | master (local compile) | PR (local compile) |
| --- | --- | --- |
| 0.088 | 0.0895 | 0.079 |
```python
    from timeit import repeat

    print(
        'getsize(L) (x100): ',
        min(repeat(
            'font.getsize("Hello world")',
            'from PIL import ImageFont; font = ImageFont.truetype("Tests/fonts/FreeMono.ttf", 20, layout_engine=ImageFont.LAYOUT_BASIC)',
            number=100, repeat=30
        )),
    )
```

### DIFF
--- a/src/_imagingft.c
+++ b/src/_imagingft.c
@@ -553,7 +553,7 @@ text_layout_fallback(PyObject* string, FontObject* self, const char* dir, PyObje
         return 0;
     }
 
-    load_flags = FT_LOAD_RENDER|FT_LOAD_NO_BITMAP;
+    load_flags = FT_LOAD_NO_BITMAP;
     if (mask) {
         load_flags |= FT_LOAD_TARGET_MONO;
     }


radarhere:
Thanks!

hugovk:
Test on macOS, also improved:

| 7.1.2 (PyPI)	| master before merge (local compile) | master after merge (local compile) |
| - | - | - |
| 0.035 | 0.015 | 0.011 |

PR #4664, opened 2020-06-01
Use FT_LOAD_TARGET_MONO in getsize

### LINKED ISSUE #4177: Truetype fonts not being rendered correctly with the TEXT method
### What did you do?
Created an image suitable for the Pimoroni InkyHat display
Added italic text onto the image object at different places

The whole of the last character of the text to be displayed as it did to render correctly on Stretch on the Pi Zero
Looking at the font from a Mac (Mojave), the characters look fine.

### What actually happened?
Depending on the length of the text, some or all of the last character is missing when using some italic fonts.  If the text length is around 105 - 110 pixels then you can lose some / all of the very last character.

**Padding out the text string with spaces so it is longer than 110 pixels fixes the issue**

* OS: Raspbian Buster Lite (10.1) Kernel 4.19.75+ #1270 Tue Sep 24 18:38:54 BST 2019 armv6l GNU/Linux (running on a Pi Zero)
* Python: 3.7.3 - not tested under Python 2
* Pillow: 6.2.1 (installed using pip3)

The fonts used are available from [Font Squirrel](https://www.fontsquirrel.com) - the two are:

Open Sans [here](https://www.fontsquirrel.com/fonts/open-sans)
Ubuntu [here](https://www.fontsquirrel.com/fonts/ubuntu?q%5Bterm%5D=ubuntu&q%5Bsearch_check%5D=Y)

```python

image_obj = Image.new('P', (212, 104), color = (255,255,255))
screen_image = ImageDraw.Draw(image_obj)

osbi = ImageFont.truetype('Fonts/OpenSans-BoldItalic.ttf', 15)
screen_image.text((10,10), 'Network Status test',fill = (0,0,0) , font = osbi)
screen_image.text((10,30), 'Bit longer no iffy ending',fill = (0,0,0) , font = osbi)

ubuntu = ImageFont.truetype('Fonts/Ubuntu-BI.ttf', 15)
screen_image.text((10,60), 'Network Status test',fill = (0,0,0) , font = ubuntu)
screen_image.text((10,80), 'Bit longer no iffy ending',fill = (0,0,0) , font = ubuntu)

image_obj.save('italic-test.png')
```
italic-test

nulano:
Fixes #4177.

* Adds the `mask` parameter to the internal `getsize` function. I did not add it to the public function, as I'm not sure what is the best way to update the API in a reasonable and backwards-compatible manner.
* Test is disabled on FreeType-2.3 due to different metrics. This affects Centos-6 and Amazon-1. It seems to use the same hinting algorithm for both targets (is hinting even enabled on those systems?), so I can't reproduce the original issue there. All other versions give identical output.
* Raqm is unaffected, as that layout function ignores the `mask` parameter. I did not include it in tests, as the epsilon among systems is greater than epsilon to a blank image.
* This PR adds the first test for mode 1 text functions. Perhaps more should be added in the future.


src/_imagingft.c
```diff
@@ -614,11 +616,11 @@ font_getsize(FontObject* self, PyObject* args)
     /* calculate size and bearing for a given string */
 
     PyObject* string;
-    if (!PyArg_ParseTuple(args, "O|zOz:getsize", &string, &dir, &features, &lang)) {
+    if (!PyArg_ParseTuple(args, "O|izOz:getsize", &string, &mask, &dir, &features, &lang)) {
```
hugovk, line 619: Can we add the new parameter to the end, to maintain compatibility?

People have been calling `ImageFont.font.getsize` directly. A quick example:

```python
font = ImageFont.truetype(fontname, fontsize)
limits = font.getsize(text)
```
* https://github.com/brumar/SupperWrapper/blob/b8c58c9f8ac44bec25ca7397f58c568e270e264f/superWrapper.py#L123
* via https://github.com/search?l=Python&q=font.getsize+ImageFont&type=Code

nulano: The linked and quoted example is the Python-layer function. It looks confusing as the code refers to the `FreeTypeFont` object as `font`, but updated function would have to be called with `font.font.getsize(...)`. The same is true for the search results in the first few pages.

There are three functions concerning the size of text:
```python
ImageDraw.Draw.textsize(...)
ImageFont.FreeTypeFont.getsize(...)  # the Python version; I did not update this
ImageFont.FreeTypeFont.font.getsize(...)  # the C version; updated by this PR
```

The equivalent code would be:
```python
font = ImageFont.truetype(...)
font.getsize(...)  # the Python function
font.font.getsize(...)  # the C function
```

I did not update the Python function as I think it would confusing to have the arguments in a different order, but not backwards compatible to add it at the front (suggestions are welcome).

If you still think the C function is a concern, would something with `PyArg_ParseTupleAndKeywords` be a good solution?

hugovk: Thanks, I think we're fine to change this then. The C API is internal and doesn't need to maintain backwards compatibility.

hugovk:
Thank you!

PR #4670, opened 2020-06-05
Use --freeze feature of pre-commit autoupdate

ericfrederich:
This prevents owners of repos used in pre-commit from force updating an
existing tag to contain malicious code.

The changes here were created simply by running

    pre-commit autoupdate --freeze


radarhere:
Thanks for contributing

PR #4673, opened 2020-06-06
Remove redundant noqa

hugovk:

 * Add https://github.com/asottile/yesqa to pre-commit to remove redundant `noqa` markers
 * Run pre-commit hooks which can edit code first, they may fix things later hooks would fail on
 * `pre-commit autoupdate`: only updated `pre-commit/pre-commit-hooks`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,16 +8,27 @@ repos:
         files: \.py$
         types: []
 
+  - repo: https://github.com/timothycrosley/isort
+    rev: 4.3.21
+    hooks:
+      - id: isort
+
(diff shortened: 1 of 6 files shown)


radarhere:
yesqa doesn't support Python 3.5 - https://github.com/asottile/yesqa/blob/master/setup.cfg. Is it worth waiting until Pillow 7.3 to include this?

hugovk:
I don't think we need to wait, this is only a dev thing.

yesqa is run by the CI lint job (has a newer supported Python version), and by anyone who is developing locally _and_ has installed pre-commit for this repo, which I guess is a small set of people.

If it's a problem for anyone, we can remove it or adjust it somewhow.

PR #4675, opened 2020-06-07
Fixed drawing a 1px high polygon

### LINKED ISSUE #4674: Odd behaviour of ImageDraw.Draw.polygon
### What did you do?
I drew two polygons, each specified by only two vertices that are both on a vertical or horizontal line.

Expected to get a one pixel wide vertical and a one pixel wide horizontal line.

### What actually happened?
The vertical line shows but not the horizontal one.

* OS: Ubuntu 18.04.4
* Python: 3.7.6
* Pillow: 7.1.1

```python
import numpy as np
img = Image.new("L", size=(2,2), color=0)
ImageDraw.Draw(img).polygon([(0,0),(0,1)], outline=1, fill=1)
np.array(img)
```

gives

```python
array([[1, 0],
       [1, 0]], dtype=uint8)
```
but 
```python
import numpy as np
img = Image.new("L", size=(2,2), color=0)
ImageDraw.Draw(img).polygon([(0,0),(1,0)], outline=1, fill=1)
np.array(img)
```
results in 
```python
array([[0, 0],
       [0, 0]], dtype=uint8)
```

radarhere:
Resolves #4674

When determining the `ymin` and `ymax` of a polygon, Pillow doesn't consider edges if they are horizontal. Sounds fair enough, because why would a horizontal line contribute to the height? Except, the definition of 'horizontal' here is that all of the points in the edge have the same `y` value, which in the case of a polygon that is 1px high, applies to all edges.

So this PR fixes that by allowing these 'horizontal' lines to expand the height of a polygon.


hugovk:
Thank you!

PR #4680, opened 2020-06-08
Update setup.py

### LINKED ISSUE #2860: Error upon pip installation in cygwin terminal run on Windows
I ran the command pip install Pillow
I expected the Pillow Package to install but instead got the following error

`Command "/usr/bin/python3 -u -c "import setuptools, tokenize;__file__='/tmp/pip-build-rgaobc15/Pillow/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" install --record /tmp/pip-0t2ptbmp-record/install-record.txt --single-version-externally-managed --compile" failed with error code 1 in /tmp/pip-build-rgaobc15/Pillow/`

I am using pip version  9.0.1 python version 3.6.3 and I'm trying to install PIllow version 4.3.0

Meithal:
This lets you locally install Pillow on Windows if you built it following the instructions of winbuild with pip install --editable ..\Pillow

 * You currently can't install Pillow on windows if you don't follow the instructions in winbuild/readme
 * Even if you do you still can't install it as setup script doesn't look here
 * This makes the setup script look into the winbuild build output if you followed the instructions in the readme

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -391,9 +391,12 @@ class pil_build_ext(build_ext):
                     _dbg("Looking for `%s` using pkg-config." % lib_name)
                     root = pkg_config(lib_name)
 
-            winbuil_root = os.path.join('.', 'winbuild', 'build')
-            if root is None and os.path.isdir(winbuil_root):
-                root = os.path.join(winbuil_root, 'lib'), os.path.join(winbuil_root, 'inc')
+            winbuild_root = os.path.join(".", "winbuild", "build")
+            if root is None and os.path.isdir(winbuild_root):
+                root = (
+                    os.path.join(winbuild_root, "lib"),
+                    os.path.join(winbuild_root, "inc")
+                )
 
             if isinstance(root, tuple):
                 lib_root, include_root = root
--- a/setup.py
+++ b/setup.py
@@ -391,6 +391,10 @@ class pil_build_ext(build_ext):
                     _dbg("Looking for `%s` using pkg-config." % lib_name)
                     root = pkg_config(lib_name)
 
+            winbuil_root = os.path.join('.', 'winbuild', 'build')
+            if root is None and os.path.isdir(winbuil_root):
+                root = os.path.join(winbuil_root, 'lib'), os.path.join(winbuil_root, 'inc')
+
             if isinstance(root, tuple):
                 lib_root, include_root = root
             else:
--- a/setup.py
+++ b/setup.py
@@ -395,7 +395,7 @@ class pil_build_ext(build_ext):
             if root is None and os.path.isdir(winbuild_root):
                 root = (
                     os.path.join(winbuild_root, "lib"),
-                    os.path.join(winbuild_root, "inc")
+                    os.path.join(winbuild_root, "inc"),
                 )
 
             if isinstance(root, tuple):


setup.py
```diff
@@ -391,6 +391,10 @@ def build_extensions(self):
                     _dbg("Looking for `%s` using pkg-config." % lib_name)
                     root = pkg_config(lib_name)
 
+            winbuil_root = os.path.join('.', 'winbuild', 'build')
+            if root is None and os.path.isdir(winbuil_root):
+                root = os.path.join(winbuil_root, 'lib'), os.path.join(winbuil_root, 'inc')
```
radarhere, line 396:
```suggestion
            winbuild_root = os.path.join('.', 'winbuild', 'build')
            if root is None and os.path.isdir(winbuild_root):
                root = os.path.join(winbuild_root, 'lib'), os.path.join(winbuild_root, 'inc')
```

nulano:
I do not think it is a good idea to hardcode the build directory, because as noted in the readme, the directory can be changed with a parameter (`prepare_build.py --dir=<path>`). This is useful to maintain different sets of built dependencies depending on your current branch or Python version. This PR would likely break that behaviour.

The [extended readme](https://github.com/python-pillow/Pillow/blob/master/winbuild/build.rst#building-pillow) suggests you use `build_pillow.cmd install` or `build_pillow.cmd bdist_wheel` and install the wheel (both are tested in the CI systems). If you wish to install in `--editable` mode, https://stackoverflow.com/a/19048754/1648883 suggests to use `develop` instead of `install`: `build_Pillow.cmd develop`. Does either of these options not work for you?

Meithal:
You can pass a parameter to prepare_build but how can you fetch it from setup.py? should prepare_build write that to a separate file, which setup.py would read? that may lead to some transversal vulnerabilities; or setup.py invokes directly prepare_build through subprocess.run, so you would have to document that feature in setup.py; or maybe a simple switch similar to DEBUG = False, something like winbuild_dir = None
Assuming that people would use prepare_build in its default settings seems the least disruptive way.

As for `develop`, I think it is the same thing as pip -e, and installing a wheel/egg would clone it and it is not the same thing than wanting to develop the checked out branch.

Also yes I am not using cygwin but I encounter the same issues as people using cygwin (can't find zlib, etc.), people with cygwin should use the solution proposed in the issue #2860 above my post.

nulano:
The current `prepare_build.py` script passes the lib and inc directories to `setup.py` by setting the LIB and INCLUDE enviroment variables, which are read on these lines:
https://github.com/python-pillow/Pillow/blob/7dd6faa927384d4bc315b16ab54615e35ee36bb6/setup.py#L418-L426
You can see the generated script here: https://github.com/python-pillow/Pillow/runs/749638754?check_suite_focus=true#step:8:323 The whole point of this generated `build_pillow.cmd` script is to pass these values to the setup script.

This script allows you to install Pillow with `winbuild\build\build_pillow.cmd install`; you can also use `develop` instead if you wish. This script wraps setup.py instead of pip to allow for directly invoking each step (i.e. build_ext, bdist_wheel, install, develop, ...). Pip just [wraps these steps](https://pip.pypa.io/en/stable/reference/pip_install/#build-system-interface) in a nice package manager.

---

The can't find zlib issue is just a symptom of a lot of other issues. Typically this happens because the given library is not installed, or Pillow doesn't know where to look for it on the given system. Winbuild uses the script mentioned above to tell Pillow where to look; Cygwin doesn't use winbuild and so #2860 is an unrelated issue.

Meithal:
It seems that the issue is not related indeed, should I open a new one? If I run `build_pillow install` it will copy all the files into lib.win-amd64-3.7\PIL. What I really need is to have `pip -e` semantics so I can debug the code from other local projects, `prepare_build.py` does the job but only partially.

To be able to run `pip install --editable ..\Pillow` from a neighbor project (without editing environment variables or setup.py itself) this PR would be useful.

A proper way to have #2860 fixed would be to suprocess.run `pacman -su gcc-core, devel versions of zlib, jpeg, tiff, freetype, fribidi, harfbuzz, imagequant, png, raqm` at some point? (that's only for msys2)

nulano:
> It seems that the issue is not related indeed, should I open a new one? If I run `build_pillow install` it will copy all the files into lib.win-amd64-3.7\PIL. What I really need is to have `pip -e` semantics so I can debug the code from other local projects, `prepare_build.py` does the job but only partially.
> 
> To be able to run `pip install --editable ..\Pillow` from a neighbor project (without editing environment variables or setup.py itself) this PR would be useful.

`winbuild\build\build_pillow.cmd develop` should do what you want.

> A proper way to have #2860 fixed would be to suprocess.run `pacman -su gcc-core, devel versions of zlib, jpeg, tiff, freetype, fribidi, harfbuzz, imagequant, png, raqm` at some point? (that's only for msys2)

I disagree. Pillow should not attempt to install these libraries, the user should choose which versions they wish to install, and optionally skip some of these. [The docs](https://pillow.readthedocs.io/en/stable/installation.html#external-libraries) detail the process on various systems, although Cygwin and MSYS are missing at the moment. Improving the docs would close the issue.

Meithal:
I was trying to install PIL in development mode but it didn't work, I wanted to keep the code that used PIL and PIL itself separate so I would know if it was an issue on my side or PIL side.

nulano:
Are you trying to install Pillow in a virtual environment? In that case make sure you run `prepare_build.py` with the virtual environment activated.

Otherwise you can just use `build_pillow.cmd --inplace develop` to do the same as `pip install --editable ..\Pillow` would do with your change - I have tested it locally.

Meithal:
> Are you trying to install Pillow in a virtual environment? In that case make sure you run prepare_build.py with the virtual environment activated.
You mean `..\Pillow\winbuild\prepare_build` from the other project?
Again I try to keep both things separated.

nulano:
> > Are you trying to install Pillow in a virtual environment? In that case make sure you run prepare_build.py with the virtual environment activated.
> 
> You mean `..\Pillow\winbuild\prepare_build` from the other project?

No, I was asking about [virtual environments](https://docs.python.org/3/tutorial/venv.html).

---

Your original post indicated that you wish to

>This lets you locally install Pillow on Windows if you built it following the instructions of winbuild with pip install --editable ..\Pillow

I am pointing out that this is already possible with `winbuild\build\build_pillow.cmd --inplace develop` which has the same effect.

If you wish to use it in a script, you should run it in a separate process to avoid it modifying your environment variables:
```cmd
cmd.exe /c "..\Pillow\winbuild\build\build_pillow.cmd --inplace develop"
```

radarhere:
@Meithal did you come to any conclusions about this?

radarhere:
See #4836

Issue #3176, opened 2018-06-11
Saving BLP files


Promises:
Any chance for saving BLP1 files being added in the near future?

hugovk:
There's partial support for reading BLP1: 
https://pillow.readthedocs.io/en/5.1.x/handbook/image-file-formats.html?highlight=blp1#blp

There's no plans for further support, but pull requests are welcome. Please see: 
https://github.com/python-pillow/Pillow/blob/master/.github/CONTRIBUTING.md.

Promises:
i would make a pull request if I had any idea of how images work :P

radarhere:
PR #6069 resolves this. By default, BLP images will be saved as BLP2, but you can save as BLP1 with `im.save("out.blp", blp_version="BLP1")`.

PR #4685, opened 2020-06-12
Correctly read JPEG compressed BLP images

### LINKED ISSUE #4684: BLP1 files are not handled
### What did you do?
I tried to open a BLP1 file
I expected to see the image
### What actually happened?
I saw a black square

* OS: Windows10 x64
* Python: 3.7
* Pillow: 7.2

You first need a blp1 picture, I think it is a proprietary format so i'm not sure I can put one here
To easily get one go on https://www.epicwar.com, download a map then open it with a MPQ viewer, every map has a thumbnail in the blp1 format.
```python
from PIL import Image
with Image.open("file.blp") as pil:
    convert = pil.convert('RGB')
    convert.show()
    convert.save((work_folder / name).with_suffix('.jpg'))
```

Meithal:
Fixes #4684

 * At this moment trying to open a BLP1 file is a black square
 * This is because BLP1 are a 4channel files in the BGRA format and BLP1 tries to open it in RGB format
 * This open the BLP1 file in the correct format, + counter basic protection where picture is made completely transparent because war3 engine ignores alpha channel


Meithal:
Also this can break some code, if you want to save the picture in JPG, you will have to convert it to RGB mode before.

hugovk:
> Also this can break some code, if you want to save the picture in JPG, you will have to convert it to RGB mode before.

Please could you give some more info/examples of what this could break?

Also please can you add tests?

Meithal:
Hello, a test would be having a reference blp file and comparing if the extracted file matches the original one? How is it possible to get a blp file without using blizzard tools, do we need to be able to generate a blp with pillow in first place? #3176

hugovk:
You can see the existing BLP tests in https://github.com/python-pillow/Pillow/blob/master/Tests/test_file_blp.py

They test an opened BLP image is as expected by comparing against a PNG version.

The BLP tests and files were added in https://github.com/python-pillow/Pillow/pull/3007, and see the discussion there, that the files must be able to be shared under the Pillow licence, so it's most likely we cannot include files copyright someone else (unless specific permission is given).

hugovk:
@Meithal Hi! This PR is now a year old, did you get a chance to look into adding tests?

Meithal:
I'll try to tackle it this weekend.

radarhere:
I've removed the alpha channel, which [I think you weren't sure about in the first place.](https://github.com/python-pillow/Pillow/pull/4685/commits/bed2ed7438b512e4cfb18bc3a3bb18f4a53d8a87#diff-30059d5b5f19c6de1f1b7fcb21e982958ef9432c2795ef68fb5470fb9eccb9ddR334) I did this for simplicity, and because there wasn't a test image demonstrating the need.

@Meithal to be clear, is the image that you've included here one that can be included under our license and distributed with Pillow? There was a long discussion about this in #3007

Meithal:
I'm not sure, it should have the same license than the other ones, it is generated automaticly by blizzard worldedit and included as is in a map archive.

I have been trying to find an example of blp1 file with a transparent component to include it in the test, and cover those paths but got lost in a rabbit hole doing that (I was looping through thousands of existing archives), I think your patches removed the alpha support, it may still be useful to not do so, maybe add an extra function that keeps it so older code doesn't break ?

radarhere:
You may be interested to know that #6069 currently adds support for alpha in BLP1 when `_blp_compression == 1`.

PR #4693, opened 2020-06-14
Avoid implying that _show always uses xv

radarhere:
This code in Image.py implies that `_show` always uses xv.
https://github.com/python-pillow/Pillow/blob/41a75210b5460640d41e4f0ded56f588b88c97fb/src/PIL/Image.py#L3143-L3151

This PR rearranges the code into the following to avoid that implication.

```python
def _show(image, **options):
    # override me, as necessary
    from . import ImageShow

    ImageShow.show(image, **options)


_showxv = _show
```

`title` still has a default value of `None` because
https://github.com/python-pillow/Pillow/blob/41a75210b5460640d41e4f0ded56f588b88c97fb/src/PIL/ImageShow.py#L38

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -3142,13 +3142,12 @@ def register_encoder(name, encoder):
 
 def _show(image, **options):
     # override me, as necessary
-    _showxv(image, **options)
-
-
-def _showxv(image, title=None, **options):
     from . import ImageShow
 
-    ImageShow.show(image, title, **options)
+    ImageShow.show(image, **options)
+
+
+_showxv = _show
 
 
 # --------------------------------------------------------------------


nulano:
This change would mean that users replacing `Image._showxv` with custom code would no longer have an effect. I'm not sure how common that is, just wanted to note this change.

radarhere:
That's true. As you seem to be aware though. I really don't know why you would do that. If you were going to customise this functionality, `_show` is the more likely target than `_showxv`. If someone is replacing Pillow functions with their own code though, unless it is common, my instinct would be that it's on them to keep their code working.

This is a minor cleanup, so it can be dropped if there is debate over it.

nulano:
I'm just commenting on the comment above the function: `# Simple display support.  User code may override this.`

I think this change is fine, perhaps it should just be noted somewhere. Also note that this makes it into an unused internal function, although [it seems to be used quite often from other projects](https://github.com/search?q=Image._showxv%3D&type=Code), so I wonder whether it should also be deprecated.

radarhere:
Oh. I didn't actually notice that comment. I only saw `# override me, as necessary`.

hugovk:
Both `_show` and `_showxv` begin with underscores indicating they're internal and private, but `# Simple display support. User code may override this.` suggests otherwise!

A quick search has "9 available code results" for `Image._showxv`:

* https://github.com/search?l=Python&q=%22Image._showxv%22&type=Code

Compared with 16,074 for `Image._show`:

* https://github.com/search?l=Python&q=%22Image._show%22&type=Code

Idea: shall we put a deprecation warning on `_showxv` to remove it in the future, and remove that comment?

radarhere:
Sure. See #4713 and #4714

PR #4646, opened 2020-05-25
Deprecate Image.show(command="...")

nulano:
For #4643 

Changes proposed in this pull request:

 * Deprecate the `Image.show(command="...")` parameter and remove its documentation, because it hasn't been implemented since the PIL fork, and implementing it is non-trivial.

PR #4695, opened 2020-06-14
Use command argument in Image.show

### LINKED ISSUE #4643: Can't set command used for show()
I'm trying to change the command used to display the image with show() from `display` to `pqiv` but it seems to just ignore any options I put. 

```python
f = Image.open(img)
upscaled = f.resize((1000,700), resample=Image.BOX)
upscaled.show(command='pqiv')
```

* OS: Void Linux 5.4
* Python: 3.8
* Pillow: 6.2.2

radarhere:
See #4643

The issue reports that the `command` argument is not used by Image `show()`. It should be a way to provide a custom command to use on the command line to show the image. PR #4646 suggests deprecating it as a solution.

As an alternative, this PR actually implements the `command` argument.

### DIFF
--- a/Tests/test_image.py
+++ b/Tests/test_image.py
@@ -12,6 +12,7 @@ from .helper import (
     assert_image_similar,
     assert_not_all_same,
     hopper,
+    is_mingw,
     is_win32,
 )
 
@@ -608,6 +609,24 @@ class TestImage:
                 im.load()
 
             assert not fp.closed
+
+    def test_show_command(self, tmp_path):
+        # Arrange
+        temp_file = str(tmp_path / "opened")
+        with open(temp_file, "a") as f:
+            f.close()
+        assert os.path.exists(temp_file)
+        im = hopper()
+
+        # Act
+        im.show(
+            command=("del" if (is_win32() and not is_mingw()) else "rm")
+            + " "
+            + temp_file
+        )
+
+        # Assert
+        assert not os.path.exists(temp_file)
 
     @pytest.mark.parametrize(
         "test_module", [PIL, Image],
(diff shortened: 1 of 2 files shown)


hugovk:
Just a thought: the test uses a `command` to delete a file.

Are we opening up a potential security hole, and might it be better to leave this broken and deprecate and eventually remove instead?

radarhere:
You can use `subprocess.Popen` instead, to ensure that you're just running a single executable with the image file. I just can't find a way to successfully test that.

But one might still think that any form of the `command` argument - allowing an arbitrary executable to be run - is still unusual behaviour from a security perspective. So sure, this can be closed if deprecation is the agreed solution.

nulano:
Using `subprocess.Popen` doesn't allow for passing arguments to the running executable with `command=`, does it? I'm sure there are viewers where that could be necessary.

The reason I don't like the `command` argument is that it seems to be just a generic way to ask Pillow to run an arbitrary command on the file. If that is the intended behaviour, why not just call the relevant function `exec` or something similar? And that obviously sounds like a potential security issue.

I don't think registering a custom Viewer with `ImageShow.register` is too much to ask, especially if the documentation is improved (e.g. #4697); the one in this PR is only 3-4 lines anyway.

The only reason `command` would be useful is to select a viewer out of the available viewers, but arbitrary options can already be passed to custom viewers to filter them by calling `ImageShow.show(image, **extra_args)`, or `XYZViewer.show(image)` can be called directly.

hugovk:
I'm in favour of deprecating rather than fixing for security reasons: we have an alternative, it's not been working for a long time (~decades?), and no-one has reported it until now.

PR #4696, opened 2020-06-14
Updated harfbuzz to 2.6.7

radarhere:
I've found that since harfbuzz 2.6.5, the test_imagefont test_variation_set_by_* tests started failing. Looking further, I find that it is just the assertions for AdobeVFPrototype.ttf that are failing, while TINY5x3GX.ttf passes.

Visually inspecting the difference, you get

_______|variation_adobe.png                                                                                                                   |variation_adobe_name.png                                                                                                                   |variation_adobe_axes.png                                                                                                                   |
:-----:|:------------------------------------------------------------------------------------------------------------------------------------:|:-----------------------------------------------------------------------------------------------------------------------------------------:|:-----------------------------------------------------------------------------------------------------------------------------------------:|
|Before|variation_adobe_older_harfbuzz|variation_adobe_older_harfbuzz_name|variation_adobe_older_harfbuzz_axes|
|After |variation_adobe               |variation_adobe_name               |variation_adobe_axes               |

Looking at the harfbuzz release notes for 2.6.5, my complete guess is that 'AAT is now always preferred for horizontal scripts when both AAT and OT layout tables exist at the same time.' might be culprit, AAT standing for Apple Advanced Typography while OT would be OpenType.

Rather than simply increase the epsilon to allow for the difference, I've created separate images in the test suite for before and after this change, accepting either as a passing test case.


nulano:
This is probably what was causing the issues for the MinGW builds. Can you try removing the skip mark?

radarhere:
Yep, ok, removed.

khaledhosny:
The before images seem to lack kerning, so something was broken and is fixed now. I don't think AAT has anything to do with this since these are OpenType fonts.

hugovk:
Thanks both!

PR #4701, opened 2020-06-16
Cache Windows builds

nulano:
Add cache for Windows builds on GHA to cut build times almost in half. Using key based on `test-windows.yml`, `build_prepare.py`, and the current Python location (unique per Python minor version). The only other variable affecting builds should be the VS install location, which is unlikely to change.

~I disabled libimagequant on GHA to simplify this, it is still enabled on AppVeyor.~ To make sure caching is available for PRs (see note below), libimagequant is now always built (takes 5s), but it is disabled in the Build Pillow and Build Wheel steps if building wheels.

There is a step to delete all intermediate files (~180MB on my system), the remaining size is ~30MB on my system, which seems to be getting compressed down to ~8MB on GHA per build configuration.

Note that for security reasons, PRs only have access to caches from `upstream/master`.
See [run on my repo](https://github.com/nulano/Pillow/runs/775085358?check_suite_focus=true) or re-run the Windows job to see the effect here.


.github/workflows/test-windows.yml
```diff
@@ -143,14 +171,12 @@ jobs:
 
     - name: Build wheel
       id: wheel
-      if: "github.event_name == 'push'"
```
radarhere, line 146: This change means that wheels will be built on PRs. What's your thinking behind this?

nulano: I just figured there is no reason not to. Before it was split based on whether libimagequant was being built, with this PR the build is the same (otherwise it would never cache due to security reasons). If you think the artifact size is a concern, I can remove it.

radarhere:
So the downside of this is the fact that libimagequant isn't tested anymore. Perhaps that's a minor downside, and perhaps AppVeyor is enough, but the Windows GHA jobs are slightly faster than the macOS and Linux GHA jobs, so Windows isn't a bottleneck when running the jobs either.

Others may disagree with me, it just seems like an odd move in the history of our jobs transitioning from AppVeyor to GHA.

hugovk:
I agree, it would be nice to have libimagequant on GHA if possible.

> I disabled libimagequant on GHA to simplify this, it is still enabled on AppVeyor.

How complicated would it be to keep it?

nulano:
I figured out how to keep it enabled - compile it always (takes 5s), but then disable it in the Build Pillow step. I'm already working on this.

.github/workflows/test-windows.yml
nulano: This change is now reverted.

hugovk:
So imagequant is now available for PRs, but not for pushes (eg. normal builds on mater)?

What's the reasoning for disabling for pushes?

nulano:
This is the same as the current behaviour. It is disabled for pushes because that is when wheels are selected to build, and imagequant is GPL. (This is not a problem for Raqm because it is LGPL and linked at runtime.)

The idea is to make it easier to test builds for contributors (especially before #4495 was done). Another option would be to build wheels only on tag creation (e.g. release)

hugovk:
thanks!

PR #4704, opened 2020-06-16
Updated lcms2 to 2.11

radarhere:


nulano:
Note: LCMS 2.11 includes additional project files for VS2019, but all changes would be removed by `build_prepare.py` anyway, so there is no reason to switch.

PR #4706, opened 2020-06-18
Do not ignore viewer if ImageShow.register order is zero

radarhere:
Suggested by https://github.com/python-pillow/Pillow/issues/4643#issuecomment-645868757

If you set `order` to zero when registering an ImageShow viewer, like so, `ImageShow.register(viewer, order=0)`, the viewer is silently ignored.

This PR changes that to treat it as a negative number instead, inserting it at the beginning of the list of viewers.


hugovk:
Thank you!

PR #2946, opened 2018-01-10
Only extract first Exif segment

hugovk:
Fixes https://github.com/python-pillow/Pillow/issues/2944.

The image in the report appears to have two [APP1 (0xFFE1/65505) aka Exif](http://www.ozhiker.com/electronics/pjmt/jpeg_info/app_segments.html) segments. 

The first looks correct (you can see "Canon PowerShot" in there):

```
[...]

PR #4707, opened 2020-06-18
Replaced primitive "magic number" inside of JpegImagePlugin._accept() function by more correct version

Cykooz:
Current version of function `JpegImagePlugin._accept()` uses a very primitive "magic number" to check that file may be a JPEG-file:
```python
def _accept(prefix):
    return prefix[0:1] == b"\377"
```
In some cases this leads to reading an entire file to decide it's not a JPEG-file. (Actually, it's a more complicated problem. Maybe I'll do another pull-request.)

Other libraries for decode JPEG-files uses more complex value of  "magic number" - `b'\xFF\xD8\xFF'`. This value is also given on Wikipedia (https://en.wikipedia.org/wiki/JPEG).


Tests/test_file_jpeg.py
```diff
@@ -706,6 +706,26 @@ def test_icc_after_SOF(self):
         with Image.open("Tests/images/icc-after-SOF.jpg") as im:
             assert im.info["icc_profile"] == b"profile"
 
+    def test_reading_not_whole_file_for_define_it_type(self):
```
radarhere, line 709: 'for define it type'... sorry, this confuses me. What is this name intended to communicate?

radarhere:
I've created https://github.com/Cykooz/Pillow/pull/1 with some suggestions.

radarhere:
Could you give a brief description of the 'more complicated problem' you allude to?

Cykooz:
My English is not very well, but I'll try to explain.

My initial problem is that Pillow, in some cases, reads an entire file just to determine its type.
I found that this problem comes from `JpegImagePlugin`. Look at this "infinite" loop
https://github.com/python-pillow/Pillow/blob/db7186b962dc94cdc30196fb6be064232f4d7080/src/PIL/JpegImagePlugin.py#L357-L386

This loop may read an entire file if content of file is something like this:
- many FF bytes;
- many FF00 "words".

Correct "magic number" doesn't fix this problem completely. But it reduces the likelihood of encountering a problem. I am not an expert in JPEG-format and I don't know how many bytes need to read from a file to decide it's not a JPEG file.

Tests/test_file_jpeg.py
Cykooz: Hmm... English is not my native language. May be more correct to name this test as "test_not_read_entire_file_to_determine_its_type".

radarhere:
> My initial problem is that Pillow, in some cases, reads an entire file just to determine its type.

I wouldn't phrase it exactly like this. It's not in order to determine its type. Pillow is attempting to open the file, hits an error and then gives up on the idea that it is a valid JPEG. So a large amount of the file has to be read to determine its validity.

Cykooz:
>  So a large amount of the file has to be read to determine its validity.

In my app, reading a file is a very expensive operation. Reading 3-4 Mb data from a file may take several seconds. This may be used in DoS attacks.
But I will not argue if you are sure that a JPEG file can contain a lot of "garbage" at the beginning of the file.

radarhere:
We guard against images with large dimensions. We don't presently guard against images with large file sizes.

Setting aside the matter of 0xFF bytes for one second, part of the problem is that Pillow reads data from before the SOS marker on the initial open. Without changing that, the only way to solve your dilemma would be if valid JPEGs only allowed a limited number of markers before SOS.

Considering that
- we have encountered an image with multiple APP1 segments - #2946 - and we would presumably like to be flexible enough to handle such a file
- and going through our test suite, I can see that APP1 markers can come before the SOS marker

I don't see a way forward there.

PR #4711, opened 2020-06-21
Install NumPy with OpenBLAS

radarhere:
GHA macOS PyPy is currently failing on master. This fixes it by installing NumPy with OpenBLAS.


hugovk:
Thank you!

PR #4714, opened 2020-06-21
Deprecated _showxv

radarhere:
My attempt to deprecate `Image._showxv`, without raising a `DeprecationWarning` when calling `Image.show` or `Image._show`.

Inspired by https://github.com/python-pillow/Pillow/pull/4693#issuecomment-647110867

### DIFF
--- a/Tests/test_image.py
+++ b/Tests/test_image.py
@@ -5,7 +5,7 @@ import tempfile
 
 import PIL
 import pytest
-from PIL import Image, ImageDraw, ImagePalette, UnidentifiedImageError
+from PIL import Image, ImageDraw, ImagePalette, ImageShow, UnidentifiedImageError
 
 from .helper import (
     assert_image_equal,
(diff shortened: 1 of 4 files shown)


hugovk:
Would something like this be simpler?

```diff
 def _show(image, **options):
-    _showxv(image, **options)
+    from . import ImageShow
+
+    ImageShow.show(image, **options)


 def _showxv(image, title=None, **options):
     from . import ImageShow

+    warnings.warn(
+        "_showxv is deprecated and will be removed in a future release. "
+        "Use Image.show instead.",
+        DeprecationWarning,
+        stacklevel=2,
+    )
     ImageShow.show(image, title, **options)
```

Also adds `stacklevel=2` for the warning to show the line of the user's own script calling `_showxv` (where action is required), instead of the line in `Image.py`.

radarhere:
I'm just thinking that if a user has overridden `_showxv` and expects their custom function to be called by `im.show`, changing the code as you have suggested would break their code, instead of showing a deprecation warning.

hugovk:
Thanks!

PR #4720, opened 2020-06-23
Call _accept instead of duplicating code

radarhere:
`_open` sometimes contains the same check as `_accept`. This PR changes some plugins to call `_accept` directly, instead of just duplicating the code.

This is already done by some other plugins. For example,
https://github.com/python-pillow/Pillow/blob/2456601e6b15849c5a4a72548a73d29647771a37/src/PIL/CurImagePlugin.py#L38-L45

### DIFF
--- a/src/PIL/BmpImagePlugin.py
+++ b/src/PIL/BmpImagePlugin.py
@@ -263,7 +263,7 @@ class BmpImageFile(ImageFile.ImageFile):
         # read 14 bytes: magic number, filesize, reserved, header final offset
         head_data = self.fp.read(14)
         # choke if the file does not have the required magic bytes
-        if head_data[0:2] != b"BM":
+        if not _accept(head_data[0:2]):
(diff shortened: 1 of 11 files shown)


src/PIL/BmpImagePlugin.py
```diff
@@ -263,7 +263,7 @@ def _open(self):
         # read 14 bytes: magic number, filesize, reserved, header final offset
         head_data = self.fp.read(14)
         # choke if the file does not have the required magic bytes
-        if head_data[0:2] != b"BM":
+        if not _accept(head_data[0:2]):
```
hugovk, line 266: Just pass the data and let `_accept` do whatever it needs to? (Like in the other files.)

```suggestion
        if not _accept(head_data):
```

```python
def _accept(prefix):
    return prefix[:2] == b"BM"
```

radarhere: Yep. Missed this one.

PR #4722, opened 2020-06-23
Use hypot function

### LINKED ISSUE #4718: Line width is broken for super long lines
Code for two images with diagonal lines differing only in coordinates of the points: 

```python

im_size = 1600

im1 = Image.new ('RGB', (im_size, im_size), (255, 255, 255))
draw1 = ImageDraw.Draw (im1)
point1 = ((-16000, -16000), (16000, 16000))
draw1.line (point1, fill = (0, 255, 0), width = 2)
im1.save ("./im1.png", "PNG", optimize = True)

im2 = Image.new ('RGB', (im_size, im_size), (255, 255, 255))
draw2 = ImageDraw.Draw (im2)
point2 = ((-16400, -16400), (16400, 16400))
draw2.line (point2, fill = (0, 255, 0), width = 2)
im2.save ("./im2.png", "PNG", optimize = True)
```

Python 3.8.2, Pillow 7.1.2 on Windows 10

Result: 
im1
im2

Seems something strange happens if the coordinate difference in either dimension goes beyond 32768. 

Not adding `width = 2` would be fine. 

I hope this will get fixed. My work involves mass auto-generated coordinates, and managing them within a bounded region without changing their appearance on the image seems too difficult a task.

radarhere:
Resolves #4718

Uses `hypot` function to fix issue where co-ordinates greater than int size produced unexpected output. Credit to https://github.com/python-pillow/Pillow/pull/4721#issuecomment-648009596

Also takes this opportunity to go through Tests/test_imagedraw.py and replace `assert_image_similar(im, Image.open(expected), ...)` with `assert_image_similar_tofile(im, expected, ...)`.


hugovk:
Thanks!

Issue #4569, opened 2020-04-20
Absent pixels when drawing font


gaoxinge:
### What did you do?

- I use pillow to draw font.

### What actually happened?

![1](https://user-images.githubusercontent.com/16648345/79721975-d63b8c00-8315-11ea-802d-094c8f1c084d.png)

- I observe there are some absence pixels on the top of numbers: `0`, `2`, `3`, `8`, `9` .

### What are your OS, Python and Pillow versions?

* OS: win10
* Python: 3.6.1
* Pillow: 7.1.1

```python
from PIL import Image, ImageFont, ImageDraw

ttf_path = "vista-hei-all.ttf"
text = "0123456789"
text_size = 35

font = ImageFont.truetype(ttf_path, text_size)
text_width, text_height = font.getsize(text)
canvas = Image.new("RGBA", [text_width, text_height], (255, 0, 0, 0))
draw = ImageDraw.Draw(canvas)
draw.text((0, 0), text, font=font, fill="#000000")
canvas.show()
```

radarhere:
Are you able to replicate this problem with a font that you can link to?

gaoxinge:
@radarhere I place a zip here.

- [vista-hei-all.zip](https://github.com/python-pillow/Pillow/files/4502112/vista-hei-all.zip)

radarhere:
As a note for future investigation of this, here's a clear way to demonstrate that this is a bug - removing the '0' from '0123456789' causes the missing pixels to be drawn without a problem.

nulano:
I have seen this before in some previous tests. I believe the cause is #185, as the solution there seems to be to count the text distance from the bottom instead of from the top. (Confusingly, this patch made the `ascender` variable refer to the descender height.) This is wrong and works only because most fonts tend to have a gap at the top. I believe I have fixed this properly in my WIP PR for https://github.com/python-pillow/Pillow/issues/4511#issuecomment-612940904, by enabling hinting for `FT_GetGlyphCBox` in `_imageft.getsize`.

gaoxinge:
@nulano Nice!

gaoxinge:
I have run the code below on pillow version 7.1.1 and 8.0.1 respectively, and the bug seems to be fixed.

```python
from PIL import Image, ImageFont, ImageDraw

ttf_path = "vista-hei-all.ttf"
text = "0123456789"
text_size = 35

font = ImageFont.truetype(ttf_path, text_size)
text_width, text_height = font.getsize(text)
[...]

Issue #4511, opened 2020-04-01
RTL Direction not having any effect


CovertError:
### What did you do?
 I'm trying to put text on top of an image and I also set the direction to RTL when doing this before I used to get some library not installed called libraqm but no with it install it still doesn't work I'm testing with the Arabic language 
### What did you expect to happen?
 the text goes RTL
### What actually happened?
the text went LTR
### What are your OS, Python and Pillow versions?
* OS: Ubuntu 18
* Python: 3.6
* Pillow: latest version

```python
def PhotoAlterer(name):
    image = Image.open('check.jpeg')

    draw = ImageDraw.Draw(image)

    font = ImageFont.truetype('din-next-lt-w23-medium.ttf', size=60)

    text_to_be_reshaped = '%s' % name
    reshaped_text = arabic_reshaper.reshape(text_to_be_reshaped)
    (x, y) = (1050, 427)
    color = 'rgb(255, 255, 255)'  # white color
    draw.text((x, y), reshaped_text , fill=color, font=font, direction="rtl")

    image.save('pImg/%s.png' % name)

```

radarhere:
Could you adjust the code that you've provided so that it's self-contained? At the moment, I don't know the value of `name` or what `arabic_reshaper` does.

radarhere:
This is the sequel to #4496

So from that, it seems reasonable to presume that `arabic_reshaper` is https://pypi.org/project/arabic-reshaper/, and that 'محمد' is a valid value for `text_to_be_reshaped`.

```python
from PIL import Image, ImageDraw, ImageFont
import arabic_reshaper

def PhotoAlterer(name):
    image = Image.open('check.jpeg')

    draw = ImageDraw.Draw(image)

    font = ImageFont.truetype('din-next-lt-w23-medium.ttf', size=60)

    text_to_be_reshaped = '%s' % name
    reshaped_text = arabic_reshaper.reshape(text_to_be_reshaped)
    (x, y) = (1050, 427)
    color = 'rgb(255, 255, 255)'  # white color
    draw.text((x, y), reshaped_text , fill=color, font=font, direction="rtl")

    image.save('pImg/%s.png' % name)

PhotoAlterer('محمد')
```

You may be interested in https://github.com/python-pillow/Pillow/issues/3081#issuecomment-557918892, but more likely, you're looking for https://github.com/python-pillow/Pillow/issues/3125#issuecomment-504790947

[...]

Issue #4483, opened 2020-03-23
font.getsize() changes after installing libraqm


seasker:
I am using PIL to render text on images with position annotation. The text size changes after installing libraqm. 
```python
from PIL import Image, ImageFont, ImageDraw
font=ImageFont.truetype('Microsoft/msyhl.ttc',size=14,layout_engine=ImageFont.LAYOUT_RAQM)
print(font.getsize('a'))  
print(font.getsize('b'))  
print(font.getsize('c'))  
print(font.getsize('abc'))  
```
Result without libraqm
```python
(7, 16)
(8, 16)
(7, 16)
(22, 16)

```
The sum of character's width 7+8+7=22 is correct. 
Result after installing libraqm
```python
(8, 16)
(9, 16)
(7, 16)
(23, 16)
```
The sum of character's width 8+9+7=24 doesn't equal to 23. 
 Also I find that the text size changes differently on different font after installing libraqm.
Why are these happening after installing libraqm?

radarhere:
Well, libraqm is a library for complex text layout. So it doesn't seem unreasonable that it would want to layout text slightly differently.

If you don't want to use raqm, instead of
```python
font=ImageFont.truetype('Microsoft/msyhl.ttc',size=14,layout_engine=ImageFont.LAYOUT_RAQM)
```
you can use `ImageFont.LAYOUT_BASIC`,
```python
font=ImageFont.truetype('Microsoft/msyhl.ttc',size=14,layout_engine=ImageFont.LAYOUT_RAQM)
```

Regarding the fact that 8+9+7 doesn't equal to 23, it is certainly possible in other languages that the width of a combination of two characters is not the same as the width of the two characters separately. It's also conceivable if the English font were slanted - one letter might easily stray into the space of the previous one. I imagine that this is less obvious in your font.

If you'd like to follow this further, it would be helpful to provide a copy of a font that we could use to reproduce this. However, I don't think it's necessarily true that 'a', 'b' and 'c' should combine to make the width of 'abc'.

nulano:
The font `msyhl.ttc` in the original post appears to be a component of Windows Vista or newer (but I don't see the light version on my Windows 7 Professional, only on my Windows 10 Home): https://docs.microsoft.com/en-us/typography/font-list/microsoft-yahei

---

[...]

Issue #3977, opened 2019-07-18
Font spacing changes


NathanWailes:
### What did you do?

1. I'm using Pillow to create many individual images which I then stitch together to create a lyrics video for a song.
    - The app is online at https://www.rhymecraft.guru
1. One of the features of the videos created by my app is that, while showing an entire line of lyrics on-screen at once, the currently-pronounced syllable can be highlighted.
    - You can see an example here: https://www.youtube.com/watch?v=9u9ylseLAfU
1. The way I achieve this is by:
    1. first calculating where the X, Y coordinates of the line would start if the entire line of lyrics was drawn at once
    1. then drawing the text *preceding* the highlighted syllable and recording the width of that text,
    1. then drawing the highlighted syllable, using the original X, Y coordinates and the width of the preceding text to calculate where the highlighted syllable's X, Y coordinates should start,
    1. then proceeding in the same way to draw the text following the highlighted syllable.

### What did you expect to happen?

I expected that the current syllable would change color without any other visible changes.

### What actually happened?

The text spacing always doesn't seem to work totally right, so the lyrics sometimes get spaced out.

You can see it:
- caused by the word 'of' at 0:47: https://www.youtube.com/watch?v=9u9ylseLAfU&t=47s
- in the word 'paradise' at 1:03: https://www.youtube.com/watch?v=9u9ylseLAfU&t=1m3s

### What are your OS, Python and Pillow versions?

* OS: Windows 10 (local machine), Ubuntu (Digital Ocean)
* Python: 3.7
* Pillow: 5.0.0

Example code:

```python
from PIL import ImageDraw, Image, ImageFont

img = Image.new('RGBA', (200, 60), color=(0, 0, 0))
draw = ImageDraw.Draw(img)
font = ImageFont.truetype('arial.ttf', size=48)
first_font_color = (255, 0, 0)
second_font_color = (255, 255, 255)

draw.text((0, 0), 'paradise', first_font_color, font=font)

draw.text((0, 0), 'par', second_font_color, font=font)
par_width, par_height = draw.textsize('par', font=font)
draw.text((0 + par_width, 0), 'adise', second_font_color, font=font)

img.save('example.png')
```

There's first a red 'paradise' created, then a white 'paradise' created in the same spot.  If you zoom in you'll notice that the red one shows up underneath the white one starting with the 'a' and then that continues for the 'dise':

[...]

PR #4724, opened 2020-06-23
Implement anchor for truetype text functions

### LINKED ISSUE #4553: Vertical text alignment ignores yOffset
While working on implementing the `anchor` parameter of `ImageDraw.text` (see https://github.com/python-pillow/Pillow/issues/4511#issuecomment-612940904), I noticed a bug with the handling of y_offset with raqm.

For the following code:

```python
font = ImageFont.truetype("segoeui.ttf", 60, layout_engine=ImageFont.LAYOUT_RAQM)
def test(text):
    im = Image.new("RGB", (200,200), "white")
    d = ImageDraw.Draw(im)
    d.line(((0,100),(200,100)), "black")
    d.line(((100,0),(100,200)), "black")
    d.text((100,100),text,"black", font=font)
    im.show()
test("a" + ("\u030C" * 20))    # a + 20 * Unicode Character 'COMBINING CARON' (U+030C)
```

I expect to get the following image:
image

But I get the following image (with master):
image

This is because when raqm support was added, the `yoffset` variable was not updated to look at the `glyphinfo[i].y_offset` value: https://github.com/python-pillow/Pillow/blob/97280a045cc2544ef8855dac2765e765565772cd/src/_imagingft.c#L665-L674

Since `font_render` looks at the `bitmap.top` value anyway (see below), which should be equal to `bbox.yMax`, I believe `yoffset` is supposed to calculate the same value as `y_max`, so it can just be removed and `y_max` used instead (in second snippet below).
https://github.com/python-pillow/Pillow/blob/97280a045cc2544ef8855dac2765e765565772cd/src/_imagingft.c#L801-L804
https://github.com/python-pillow/Pillow/blob/97280a045cc2544ef8855dac2765e765565772cd/src/_imagingft.c#L710-L712

I am planning to include this change in the PR implementing the `anchor` parameter, but I wanted to document this as a separate issue.

nulano:
I am splitting this PR into multiple parts. I have marked this as draft until all parts are finished and will close this afterwards.

1. [x] Refactor font_getsize and font_render (#4910)
2. [x] Implement anchor for TrueType fonts (#4930)
3. [x] Add `getlength` and `getbbox` functions for TrueType fonts (#4959)

---

The following is the original text of this PR:

---

* For #4651, #3977, https://github.com/python-pillow/Pillow/issues/4483#issuecomment-617319843, #3921, likely others.
  Add new function `getlength` and `getbbox` which compute the advance length and rendered bounding box of text respectively.

  I did not update `getsize` (and `getoffset`) and the related `ImageDraw` functions, as I don't think this is possible while maintaining backwards compatibility. The issue is that while `getoffset` is correct, `getsize` returns the width and vertical distance to the last drawn pixel (i.e. the bottom parameter of the bounding box). The horizontal is sometimes equal to the advance length (now in `getlength`), but breaks with e.g. slanted fonts or odd diacritics (see `test_combine` for double breve). This is because the bbox is always extended at least up to the advance length. The returned vertical distance is from the ascender line to the bottom-most pixel, which is mostly useless without looking at `getoffset` as well, which is confusing.

  `getbbox` is suggested by @wiredfool in https://github.com/python-pillow/Pillow/pull/784#issuecomment-48213245, `getlength` is inspired by my thoughts in https://github.com/python-pillow/Pillow/issues/3977#issuecomment-617349188 where I also described the problems mentioned above.
  I will gladly improve the API if someone can give a good suggestion.
  - [x] ~TODO `ImageDraw.Draw.textsize` docs should at least point to the new functions, or they should also be added to `ImageDraw`.~ I have added a note summarising the paragraph above to `Draw.textsize`.
  - [ ] TODO Multiline text equivalent to `getbbox`. I would like to get feedback for the above before adding a multiline version.
  
  <p></p>
  `getbbox` is merely a better public api for the *private version* of `getsize`, while `getlength` is 3x faster for basic layout (2x for raqm with simple text) because it only performs the layout step and sums the advance, without the need to reload all glyphs (it would be nice to modify the layout to cache the loaded glyhs to speed up internal `getsize` as well). Testing with the following code, I get these times (Windows 10 laptop):

  | layout_engine | getsize (7.1.2 from PyPI) | getsize (master) | getsize (PR) | getlength (PR) |
  | --- | --- | --- | --- | --- |
  | LAYOUT_BASIC | 0.092 | 0.080 | 0.082 | 0.027 |
  | LAYOUT_RAQM | 0.107 | 0.108 | 0.108 | 0.053 |

  <details><summary>Code (click to expand)</summary><p>

  ```python
    from timeit import repeat
    print(
        'getsize(L) (x100): ',
        min(repeat(
            'font.getsize("Hello world")',
            'from PIL import ImageFont; font = ImageFont.truetype("Tests/fonts/FreeMono.ttf", 20, layout_engine=ImageFont.LAYOUT_BASIC)',
            number=100, repeat=30
        )),
    )
    if hasattr(FreeTypeFont, 'getlength'):
        print(
            'getlength(L) (x100): ',
            min(repeat(
                'font.getlength("Hello world")',
                'from PIL import ImageFont; font = ImageFont.truetype("Tests/fonts/FreeMono.ttf", 20, layout_engine=ImageFont.LAYOUT_BASIC)',
                number=100, repeat=30
            )),
        )
  ```
  </p></details>

---

* For https://github.com/python-pillow/Pillow/issues/4511#issuecomment-612940904, #4553, #4569, #2293, many others.
  Implement `anchor` parameter for text rendering and fix related issues.

  My suggestion for the anchor API is already documented: https://pillow--4724.org.readthedocs.build/en/4724/handbook/text-anchors.html

  From https://github.com/python-pillow/Pillow/issues/4569#issuecomment-617175808:
  >  I believe the cause is #185, as the solution there seems to be to count the text distance from the bottom instead of from the top. (Confusingly, this patch made the ascender variable refer to the descender height.) This is wrong and works only because most fonts tend to have a gap at the top. I believe I have fixed this properly in my WIP PR for #4511 (comment), by enabling hinting for FT_GetGlyphCBox in _imagingft.getsize.

  There were many places where the original text code was merely rounding glyph positions, often incorrectly. My solution is to instead let FreeType perform rounding using its hinting engine with `FT_Set_Transform` and asking for hinting in `FT_GetGlyphCBox`. The limitation is that FreeType converts the value to 16.16 precision internally, thus limiting max line length to 32768 pixels. I don't think this should be an issue (outside of updating the leak tests), but it is possible to extend this range by splitting the text into multiple runs in Pillow code.

  Additionally, for #4553 and related issues, all glyphs are now checked when calculating offsets for the baseline. See `test_imagefontctl.py:test_combine` for other issues of the same class.

  These changes required updating most test images for `test_imagefontctl.py`. I believe most of the changes improve kerning or fix #4553 issues. In `test_direction_ttb`, the bottom of the text was even clipped!

---

* Remove the additional offset caused by stroking text.
  This was incompatible with the anchor implementation and I believe it to be a bug; I don't think users should need to offset text just to stroke a single word in the middle of a line.

---

- [ ] TODO Kerning is still broken (since #2576). I see two options which would work well with `getlength`, such that users can expect its sum to be consistent (see #3977): 
1) Remove kerning for basic layout, since it has been broken for years anyway.
2) Add support for `-kern` as the only feature supported by basic layout.

(The problem is that kerning is scaled twice, see below.)
https://github.com/python-pillow/Pillow/blob/c9745f4b517127bd5090cc381aa1e67c0d780499/src/_imagingft.c#L578-L585


nulano:
Can I change something here to make this easier to review? I'm hoping this could be merged in time for 8.0.0. See also https://github.com/python-pillow/Pillow/issues/3346#issuecomment-683471416.

hugovk:
Thanks for the PR, it's a big one! That also makes it harder to review. Would it be possible to split it into smaller PRs?

nulano:
I think I can split this into three incremental steps (PR2 depends on PR1 changes etc.):
1) Refactor and fix some text rendering issues (clipping, rounding issues, etc.)
2) Implement anchor parameter
3) Add getbbox and getlength

However, I don't think I can write tests for (1) specifically. Would it be OK to merge that under the assumption that the next PR in the sequence adds tests? For example, the test for #4553 makes heavy use of the anchor parameter, [see here](https://github.com/python-pillow/Pillow/blob/57242d39722d02fa6f9c562caf6006e8b0db51c2/Tests/test_imagefontctl.py#L288-L352). I can also add this test to (1) with the anchor test cases commented out, but I think this will only partially test the (1) changes.

hugovk:
Yeah, that'll probably be okay. At least for refactoring, you don't always need new tests because generally the idea is not to change what is done, but how it's done. Thank you!

nulano:
I have now rebased the first part in PR #4910. While writing the changes comment, I realized I had misread the FreeType documentation for `FT_Set_Transform`, so it took me a bit more time than expected to fix and test a different approach.

src/PIL/ImageFont.py
```diff
@@ -215,13 +215,123 @@ def getmetrics(self):
...
+                         <https://www.w3.org/International/articles/language-tags/>`
+                         Requires libraqm.
+
+        :return: Width for horizontal, height for vertical text.
+        """
+        return (
+            self.font.getlength(text, mode == "1", direction, features, language) / 64
+        )
+
+    def getbbox(
```
gofr, line 264: Would it be pedantic to say this should be called `getcbox`, since this uses FreeType's `FT_Glyph_Get_CBox`, not `FT_Outline_Get_BBox`, if some future version of Pillow would ever add a separate version for that?

nulano: The idea is to return the bounding box of the drawn text, i.e. the rectangle of pixels that may be modified by `draw.text`, not just the union of glyph `CBox`es. In fact, this function doesn't even return the tightest possible bound on the glyph `CBox`es, since it also includes the pen line (i.e. the origin and glyph advance points).

I think the reason for calling it `getbbox` is that this is what it is most often called in Pillow code (see https://pillow.readthedocs.io/en/latest/search.html?q=bbox). Another possible name would be `getbounds` to avoid confusion with the similarly named concept in FreeType? However, this doesn't seem to be used anywhere else in code yet (see https://pillow.readthedocs.io/en/latest/search.html?q=bound).

nulano:
Part two, implementing the anchor parameter, is now rebased in #4930.

nulano:
Part three, the finale, is now rebased in #4959, also adding `ImageDraw` versions of the two functions.

nulano:
All parts of this PR have now been merged, closing.

Issue #4725, opened 2020-06-23
Error on saving APNG


dhsdshdhk:
### What did you do?
I opened a PNG image that behaves sort of like a GIF , then I tried to save.

### What did you expect to happen?
No error, image saved normally.

### What actually happened?

`Traceback (most recent call last):                                                                                        File "<stdin>", line 1, in <module>                                                                                     File "C:\Python38\lib\site-packages\PIL\Image.py", line 2100, in save                                                     self._ensure_mutable()                                                                                                File "C:\Python38\lib\site-packages\PIL\Image.py", line 617, in _ensure_mutable                                           self._copy()                                                                                                          File "C:\Python38\lib\site-packages\PIL\Image.py", line 610, in _copy                                                     self.load()                                                                                                           File "C:\Python38\lib\site-packages\PIL\ImageFile.py", line 262, in load                                                  self.load_end()                                                                                                       File "C:\Python38\lib\site-packages\PIL\PngImagePlugin.py", line 921, in load_end                                         self._prev_im.paste(dispose, self.dispose_extent)                                                                   ValueError: images do not match  `

### What are your OS, Python and Pillow versions?

* OS: Windows 10
* Python: 3.8.3
* Pillow: 7.1.2

```python
from PIL import Image
im = Image.open('problem.png')
im.save('out.png')

```
![problem](https://user-images.githubusercontent.com/28062763/85478800-4b844100-b593-11ea-8662-478de5d8c41b.png)

hugovk:
A quick check: it's coming from `PngImageFile.load_end` because `dispose.mode` is RGBA and `self._prev_im.mode` is P here:


```python
            if dispose:
                self._prev_im.paste(dispose, self.dispose_extent)
```
And they must match in `ImagingPaste` in `Paste.c` (the x and y sizes all match):
```c
    if (xsize != imIn->xsize || ysize != imIn->ysize ||
        pixelsize != imIn->pixelsize) {
        (void) ImagingError_Mismatch();
        return -1;
    }
```


ping @pmrowla

pmrowla:
I'll look into this when I get a chance, but probably won't be until this weekend at the earliest

radarhere:
I've created PR #4742 to resolve this. Feel free to comment @pmrowla if you have any thoughts.

dhsdshdhk:
Thank you @radarhere :)

PR #4726, opened 2020-06-24
APNG: Fix setting disposal

hugovk:

 * This looks like a bug: should assign, rather than a comparison that has no effect
 * I don't have a test, but they pass before and after
 * https://github.com/python-pillow/Pillow/issues/4725 isn't fixed by this 

cc @pmrowla


pmrowla:
this is definitely a bug, I can add a test case whenever I get to #4725

radarhere:
I've created https://github.com/hugovk/Pillow/pull/52 to add a test.

PR #4739, opened 2020-06-27
Document is_animated and n_frames

nulano:
For #4609. Alternative to #4628.

Instead of defining `is_animated` and `n_frames`, the suggested 'workaround' can be documented.

Created in the style proposed in #4735.
Also fixes 3 nitpicky warnings:

```
-C:\Git\Pillow\docs\handbook\image-file-formats.rst:564: WARNING: py:attr reference target not found: PIL.Image.Image.is_animated
-C:\Git\Pillow\docs\handbook\image-file-formats.rst:564: WARNING: py:attr reference target not found: PIL.Image.Image.n_frames
-C:\Git\Pillow\docs\handbook\image-file-formats.rst:584: WARNING: py:attr reference target not found: PIL.Image.Image.n_frames
```


docs/reference/Image.rst
```diff
@@ -330,6 +330,34 @@ Instances of the :py:class:`Image` class have the following attributes:
...
+    :type: bool
+
+    ``True`` if this image has more than one frame, or ``False`` otherwise.
+
+    This attribute is only defined by image plugins that support animated images.
+    Plugins may leave this attribute undefined if they don't support loading
+    animated images, even if the given format supports animated images.
+
+    To check whether an image is animated regardless of its format, use
+    ``getattr(image, "is_animated", False)``.
```
hugovk, line 343: The problem of https://github.com/python-pillow/Pillow/pull/4628#issuecomment-650542773 still applies: this will return `False` for some animated images (and below return 1 for `n_frames`).

radarhere, line 343:
```suggestion
    Given that this attribute is not present for all images, to check if Pillow is
    aware of multiple frames in an image, regardless of the image's format, use
    ``getattr(image, "is_animated", False)``.
```

You're right. What about this?

hugovk: :+1: That's better!

docs/reference/Image.rst
```diff
@@ -330,6 +330,34 @@ Instances of the :py:class:`Image` class have the following attributes:
...
+    :type: int
+
+    The number of frames in this image.
+
+    This attribute is only defined by image plugins that support animated images.
+    Plugins may leave this attribute undefined if they don't support loading
+    animated images, even if the given format supports animated images.
+
+    To check the number of frames in an image regardless of its format, use
+    ``getattr(image, "n_frames", 1)``.
```
radarhere, line 357:
```suggestion
    Given that this attribute is not present for all images, to find out the number of
    frames that Pillow is aware of in an image, regardless of the image's format, use
    ``getattr(image, "n_frames", 1)``.
```

docs/reference/Image.rst
```diff
@@ -330,6 +330,36 @@ Instances of the :py:class:`Image` class have the following attributes:
...
+
+    ``True`` if this image has more than one frame, or ``False`` otherwise.
+
+    This attribute is only defined by image plugins that support animated images.
+    Plugins may leave this attribute undefined if they don't support loading
+    animated images, even if the given format supports animated images.
+
+    Given that this attribute is not present for all images use
+    ``getattr(image, "is_animated", False)`` to check if Pillow is aware of multiple
+    frames in an image regardless of its format.
```
nulano, line 344: I've commited @radarhere's suggestion with IMO slightly better word order. WDYT?

PR #4740, opened 2020-06-27
Add gitignore to winbuild build directory

nulano:
The default path is currently ignored with the [`build/` line](https://github.com/python-pillow/Pillow/blob/b4e7202dcb7235642f6e6a050c9485070ae6ea4f/.gitignore#L13), but this would be removed by #3062. Additionally, the directory is configurable with a parameter.

Adding a `*` gitignore to the generated build directory prevents it from being committed by accident in these cases.

### DIFF
--- a/winbuild/build_prepare.py
+++ b/winbuild/build_prepare.py
@@ -553,5 +553,6 @@ if __name__ == "__main__":
 
     print()
 
+    write_script(".gitignore", ["*"])
     build_dep_all()
     build_pillow()


radarhere:
Hi. What is your thinking about when people would use "PILLOW_BUILD"? I'm wondering if you meant for it to allow users to specify a directory external to Pillow, and if that's the cause, is it potentially unwanted to make a gitignore change to a different project?

hugovk:
> The default path is currently ignored with the [`build/` line](https://github.com/python-pillow/Pillow/blob/b4e7202dcb7235642f6e6a050c9485070ae6ea4f/.gitignore#L13), but this would be removed by #3062. 

This isn't a problem until #3062 is merged, and I don't think it will. 

> Additionally, the directory is configurable with a parameter.

Do we know if anyone is actually changing it with a parameter, and do we know if this is a problem they need fixing?

nulano:
> What is your thinking about when people would use "PILLOW_BUILD"?

I use it when switching python version or when updating the build steps of a dependency to avoid having to rebuild dependencies when switching back.

> Do we know if anyone is actually changing it with a parameter, and do we know if this is a problem they need fixing?

I am; it is a *very minor* inconvenience :).

> is it potentially unwanted to make a gitignore change to a different project?

This only affects the files in the generated build directory (which is deleted at the start of the script). It would act kind of like a git submodule, in that the "included project" (Pillow build) has its own ignore file. I feel it is far less likely that someone *really* wants to commit a bunch of binary files and/or scripts configured for their local environment, and in that case they can just delete the ignore file (possibly using the script they used to generate it).

> I'm wondering if you meant for it to allow users to specify a directory external to Pillow,

I just figured it might be a bit less confusing to treat the path relative to the current working directory instead of the script directory.

hugovk:
Thanks!

PR #4748, opened 2020-06-30
Simplify GHA config

hugovk:
Simplify some GitHub Actions config:

* Don't need a matrix of one version for linting
* Since pip 20.1, we can use `pip cache dir` to combine macOS/Ubuntu cache config ([example](https://github.com/actions/cache/blob/master/examples.md#using-pip-to-get-cache-location))
* Also update some actions to v2

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -24,7 +24,7 @@ jobs:
           lint-pip-
 
     - name: pre-commit cache
-      uses: actions/cache@v1
+      uses: actions/cache@v2
       with:
         path: ~/.cache/pre-commit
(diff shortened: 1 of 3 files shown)


.github/workflows/lint.yml
```diff
@@ -6,9 +6,6 @@ jobs:
   build:
 
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        python-version: ["3.8"]
 
     name: Python ${{ matrix.python-version }}
```
radarhere, line 10:
```suggestion
    name: Python 3.8
```

hugovk: Let's just call it "Lint". Will update, thanks!

PR #4749, opened 2020-06-30
Allow ImageOps.autocontrast to specify low and high cutoffs separately

### LINKED ISSUE #4731: Autocontrast cutoff improvement
### What did you do?
I am trying to improve image contrast.

I expect to be able to control the cutoff values for each side of the histograms. 

### What actually happened?
The cutoff value of PIL.ImageOps.autocontrast() cutoff the histogram from both side. What should happen is the user should be given the custom control to specify the percentage cutoff of both the right and the left side separately

* OS: Windows 10
* Python: Python 3.8.2 
* Pillow: PIL 7.1.2

```python

img = Image.open("ky.jpg").convert("L")

edited = ImageOps.autocontrast(img, cutoff=4

edited.save("pil_1.jpg")
```

millionhz:
Fixes #4731 

 ### Autocontrast enhancement:

Added the ability to change the hi and low cutoff separately without causing backwards comparability issue

The user should be allowed to change the values of the left and right side of the histogram separately so:

 * added a tuple to control the high and low values separately (cutoff[0] is for low end and cutoff[1] is for hi end)
 * to ensure backwards comparability I added a way to turn integer to a tuple so the program works both ways

### DIFF
--- a/src/PIL/ImageOps.py
+++ b/src/PIL/ImageOps.py
@@ -88,12 +88,8 @@ def autocontrast(image, cutoff=0, ignore=None):
                     h[ix] = 0
         if cutoff:
             # cut off pixels from both ends of the histogram
-            if isinstance(cutoff, int):
+            if not isinstance(cutoff, tuple):
                 cutoff = (cutoff, cutoff)
-            elif isinstance(cutoff, tuple):
-                pass
-            else:
-                raise ValueError("the cutoff can only be a integer or tuple")
             # get number of pixels
             n = 0
             for ix in range(256):
--- a/src/PIL/ImageOps.py
+++ b/src/PIL/ImageOps.py
@@ -88,12 +88,18 @@ def autocontrast(image, cutoff=0, ignore=None):
                     h[ix] = 0
         if cutoff:
             # cut off pixels from both ends of the histogram
+            if isinstance(cutoff, int):
+                cutoff = (cutoff, cutoff)
+            elif isinstance(cutoff, tuple):
+                pass
+            else:
+                raise ValueError("the cutoff can only be a integer or tuple")
             # get number of pixels
             n = 0
             for ix in range(256):
                 n = n + h[ix]
             # remove cutoff% pixels from the low end
-            cut = n * cutoff // 100
+            cut = n * cutoff[0] // 100
             for lo in range(256):
                 if cut > h[lo]:
                     cut = cut - h[lo]
@@ -104,7 +110,7 @@ def autocontrast(image, cutoff=0, ignore=None):
                 if cut <= 0:
                     break
             # remove cutoff% samples from the hi end
-            cut = n * cutoff // 100
+            cut = n * cutoff[1] // 100
             for hi in range(255, -1, -1):
                 if cut > h[hi]:
                     cut = cut - h[hi]
--- a/Tests/test_imageops.py
+++ b/Tests/test_imageops.py
@@ -300,3 +300,9 @@ def test_exif_transpose():
                     "Tests/images/hopper_orientation_" + str(i) + ext
                 ) as orientation_im:
                     check(orientation_im)
+
+def test_autocontrast_cutoff():
+    # Test the cutoff argument of autocontrast
+    with Image.open("Tests/images/bw_gradient.png") as img:
+        assert  ImageOps.autocontrast(img, cutoff=10).getdata() == ImageOps.autocontrast(img, cutoff=(10,10)).getdata()
+        assert  ImageOps.autocontrast(img, cutoff=10).getdata() != ImageOps.autocontrast(img, cutoff=(1,10)).getdata()
--- a/Tests/test_imageops.py
+++ b/Tests/test_imageops.py
@@ -304,5 +304,5 @@ def test_exif_transpose():
 def test_autocontrast_cutoff():
     # Test the cutoff argument of autocontrast
     with Image.open("Tests/images/bw_gradient.png") as img:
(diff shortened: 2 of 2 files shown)


src/PIL/ImageOps.py
```diff
@@ -88,12 +88,18 @@ def autocontrast(image, cutoff=0, ignore=None):
                     h[ix] = 0
         if cutoff:
             # cut off pixels from both ends of the histogram
+            if isinstance(cutoff, int):
+                cutoff = (cutoff, cutoff)
+            elif isinstance(cutoff, tuple):
+                pass
+            else:
+                raise ValueError("the cutoff can only be a integer or tuple")
```
radarhere, line 96:
```suggestion
            if not isinstance(cutoff, tuple):
                cutoff = (cutoff, cutoff)
```

millionhz:
I just want to ask one other thing, it is irrelevant. 
Writing text on images is slow, what is the reason behind it and can it be made faster?

radarhere:
Using one of the fonts from the test suite,

```python
import time
times = []
for i in range(1000):
	im = Image.new("RGB", (300, 130))
	draw = ImageDraw.Draw(im)
	font = ImageFont.truetype("Tests/fonts/FreeMono.ttf", 120)
	start = time.time()
	draw.text((10, 0), "test", "#f00", font)
	times.append(time.time() - start)
print("Average time to draw text: "+str(sum(times) / len(times)))
```

I get an average time of 0.0024 seconds. So when you say it is slow, are you getting a very different result than me? Do you need to provide a self-contained example to demonstrate? Do you feel it is slow compared to something else? A new issue might be the best place to have that discussion.

millionhz:
I see, actually what I am doing is using a multi-line string of a few thousand characters so I think its not the code which is slowing the drawing process down, its the few thousand characters. 
[sample image i generated](https://github.com/millionhz/textify/blob/master/images/woman.jpg) 

Also, actually im pretty new to programming and github, so is there anything i am supposed to do to resolve this pull request

radarhere:
As long as all the tests pass (or close enough in case of codecov), and there are no merge conflicts or outstanding suggestions or comments from someone, there's nothing you need to do, no. It's just about when one of the members of this project decide to merge the request.

hugovk:
Thanks! Let's remember to add release notes too.

navneeth:
I am trying to add an enhancement to the autocontrast implementation (use of a mask). However I notice that the test in the current build does not pass (master).
`
pytest Tests/test_imageops.py
` 

`
================================================================================================================= FAILURES ==================================================================================================================
_________________________________________________________________________________________________________ test_autocontrast_cutoff __________________________________________________________________________________________________________

    def test_autocontrast_cutoff():
        # Test the cutoff argument of autocontrast
        with Image.open("Tests/images/bw_gradient.png") as img:

            def autocontrast(cutoff):
                return ImageOps.autocontrast(img, cutoff).histogram()

>           assert autocontrast(10) == autocontrast((10, 10))

Tests/test_imageops.py:312:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
Tests/test_imageops.py:310: in autocontrast
    return ImageOps.autocontrast(img, cutoff).histogram()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

image = <PIL.PngImagePlugin.PngImageFile image mode=L size=256x10 at 0x7F3A5DAAEFD0>, cutoff = (10, 10), ignore = None

    def autocontrast(image, cutoff=0, ignore=None):
        """
        Maximize (normalize) image contrast. This function calculates a
        histogram of the input image, removes **cutoff** percent of the
        lightest and darkest pixels from the histogram, and remaps the image
        so that the darkest pixel becomes black (0), and the lightest
        becomes white (255).

        :param image: The image to process.
        :param cutoff: How many percent to cut off from the histogram.
        :param ignore: The background pixel value (use None for no background).
        :return: An image.
        """
        histogram = image.histogram()
        lut = []
        for layer in range(0, len(histogram), 256):
            h = histogram[layer : layer + 256]
            if ignore is not None:
                # get rid of outliers
                try:
                    h[ignore] = 0
                except TypeError:
                    # assume sequence
                    for ix in ignore:
                        h[ix] = 0
            if cutoff:
                # cut off pixels from both ends of the histogram
                # get number of pixels
                n = 0
                for ix in range(256):
                    n = n + h[ix]
                # remove cutoff% pixels from the low end
>               cut = n * cutoff // 100
E               TypeError: unsupported operand type(s) for //: 'tuple' and 'int'

`

Appears that the test for backward compatibilty of cutoff does not pass (autocontrast(10) == autocontrast((10, 10)) ) in current release. 

Any pointers & I can fix it.

Issue #4586, opened 2020-04-25
7.1.2: gcc 10 warnings


kloczek:
```
src/_imaging.c: In function ‘_set_blocks_max’:
src/_imaging.c:3743:26: warning: comparison of integer expressions of different signedness: ‘int’ and ‘long unsigned int’ [-Wsign-compare]
 3743 |     else if ( blocks_max > SIZE_MAX/sizeof(ImagingDefaultArena.blocks_pool[0])) {
      |                          ^
src/path.c: In function ‘alloc_array’:
src/path.c:59:15: warning: comparison of integer expressions of different signedness: ‘Py_ssize_t’ {aka ‘long int’} and ‘long unsigned int’ [-Wsign-compare]
   59 |     if (count > (SIZE_MAX / (2 * sizeof(double))) - 1 ) {
      |               ^
src/libImaging/Resample.c: In function ‘precompute_coeffs’:
src/libImaging/Resample.c:200:17: warning: comparison of integer expressions of different signedness: ‘int’ and ‘long unsigned int’ [-Wsign-compare]
  200 |     if (outSize > INT_MAX / (ksize * sizeof(double))) {
      |                 ^
src/libImaging/Draw.c: In function ‘allocate’:
src/libImaging/Draw.c:1065:31: warning: comparison of integer expressions of different signedness: ‘int’ and ‘long unsigned int’ [-Wsign-compare]
 1065 |             if (outline->size > INT_MAX / sizeof(Edge)) {
      |                               ^
src/libImaging/GetBBox.c: In function ‘ImagingGetExtrema’:
src/libImaging/GetBBox.c:189:25: warning: pointer targets in assignment from ‘char *’ to ‘unsigned char *’ differ in signedness [-Wpointer-sign]
  189 |                   pixel = im->image[y] + x * sizeof(v);
      |                         ^
src/libImaging/JpegEncode.c: In function ‘ImagingJpegEncode’:
src/libImaging/JpegEncode.c:248:37: warning: comparison of integer expressions of different signedness: ‘int’ and ‘size_t’ {aka ‘long unsigned int’} [-Wsign-compare]
  248 |         if (context->rawExifLen + 5 >  context->destination.pub.free_in_buffer){
      |                                     ^
src/libImaging/RankFilter.c: In function ‘ImagingRankFilter’:
[...]

PR #4752, opened 2020-07-01
Fixed comparison warnings

radarhere:
Helps #4586

Jpeg2KDecode.c - [before](https://travis-ci.org/github/python-pillow/Pillow/jobs/703615504#L2422-L2435) and [after](https://travis-ci.org/github/python-pillow/Pillow/jobs/703893359#L2337-L2344)
path.c - [before](https://travis-ci.org/github/python-pillow/Pillow/jobs/703615504#L2319-L2323) and [after](https://travis-ci.org/github/python-pillow/Pillow/jobs/703893359#L2237)

### DIFF
--- a/src/libImaging/Jpeg2KDecode.c
+++ b/src/libImaging/Jpeg2KDecode.c
@@ -742,8 +742,8 @@ j2k_decode_entry(Imaging im, ImagingCodecState state)
            swapped), bail. */
         if (tile_info.x0 >= tile_info.x1
             || tile_info.y0 >= tile_info.y1
-            || tile_info.x0 < image->x0
-            || tile_info.y0 < image->y0
+            || tile_info.x0 < (OPJ_INT32)image->x0
+            || tile_info.y0 < (OPJ_INT32)image->y0
             || tile_info.x1 - image->x0 > im->xsize
             || tile_info.y1 - image->y0 > im->ysize) {
             state->errcode = IMAGING_CODEC_BROKEN;
--- a/src/path.c
+++ b/src/path.c
@@ -56,7 +56,7 @@ alloc_array(Py_ssize_t count)
         PyErr_NoMemory();
         return NULL;
     }
-    if (count > (SIZE_MAX / (2 * sizeof(double))) - 1 ) {
+    if ((unsigned long long)count > (SIZE_MAX / (2 * sizeof(double))) - 1 ) {
         PyErr_NoMemory();
         return NULL;
     }


src/path.c
```diff
@@ -56,7 +56,7 @@ alloc_array(Py_ssize_t count)
         PyErr_NoMemory();
         return NULL;
     }
-    if (count > (SIZE_MAX / (2 * sizeof(double))) - 1 ) {
+    if ((unsigned long)count > (SIZE_MAX / (2 * sizeof(double))) - 1 ) {
```
radarhere, line 59: Immediately before this, the code `return`s if `count < 0`, so it is safe to cast it to unsigned here.

src/libImaging/Jpeg2KDecode.c
```diff
@@ -742,8 +742,8 @@ j2k_decode_entry(Imaging im, ImagingCodecState state)
            swapped), bail. */
         if (tile_info.x0 >= tile_info.x1
             || tile_info.y0 >= tile_info.y1
-            || tile_info.x0 < image->x0
-            || tile_info.y0 < image->y0
+            || tile_info.x0 < (OPJ_INT32)image->x0
+            || tile_info.y0 < (OPJ_INT32)image->y0
```
radarhere, line 746: `x0` and `y0` are [signed, as set by OpenJPEG](https://github.com/uclouvain/openjpeg/blob/cbee7891a0ee664dd83ca09553d2e30da716a883/src/lib/openjp2/openjpeg.h#L1515). So the tile offset can be theoretically negative.
Part of this block is to check '[if the tile is outside the image area](https://github.com/python-pillow/Pillow/blob/0defffaece306ffa3ac910867c692fe94bfaddbc/src/libImaging/Jpeg2KDecode.c#L740)'. If the tile offset is negative, that is definitely outside the image area.

src/path.c
```diff
@@ -56,7 +56,7 @@ alloc_array(Py_ssize_t count)
         PyErr_NoMemory();
         return NULL;
     }
-    if (count > (SIZE_MAX / (2 * sizeof(double))) - 1 ) {
+    if ((unsigned long long)count > (SIZE_MAX / (2 * sizeof(double))) - 1 ) {
```
radarhere, line 59: Immediately before this, [if `count < 0` the code `return`s](https://github.com/python-pillow/Pillow/blob/f15c64dbb667190b81fbe86a67dda67200a601cd/src/path.c#L55-L58), so it is safe to cast to unsigned here.

src/libImaging/Jpeg2KDecode.c
nulano: Would this cause issues for images with `x0` or `y0` greater than 2**31, since [image->x0 is unsigned](https://github.com/uclouvain/openjpeg/blob/cbee7891a0ee664dd83ca09553d2e30da716a883/src/lib/openjp2/openjpeg.h#L680-L683)? The following suggestion would catch both cases:

```suggestion
            || tile_info.x0 < 0
            || tile_info.y0 < 0
            || (OPJ_UINT32)tile_info.x0 < image->x0
            || (OPJ_UINT32)tile_info.y0 < image->y0
```

One could argue that such an image could have no valid tiles, but I do not see another check specifically for that case.

src/libImaging/Jpeg2KDecode.c
```diff
@@ -742,8 +742,8 @@ j2k_decode_entry(Imaging im, ImagingCodecState state)
...
             || tile_info.x1 - image->x0 > im->xsize
             || tile_info.y1 - image->y0 > im->ysize) {
```
nulano, line 748:
```suggestion
            || (OPJ_INT32)(tile_info.x1 - image->x0) > im->xsize
            || (OPJ_INT32)(tile_info.y1 - image->y0) > im->ysize) {
```

The previous 4 lines verify that `tile_info.x1 > tile_info.x0` and `tile_info.x0 >= image->x0`, so by transitivity `tile_info.x1 > image->x0` and so the result fits into the unsigned type of `tile_info.x1` (`y` is analogous).

PR #4753, opened 2020-07-01
Minor fix: remove str(bytes) call

Conchylicultor:
Some environments have strict mode to catch potential str<>bytes error. This is triggered by this line:

```
TiffImagePlugin.py3", line 996, in _open
    logger.debug("- ifh: {}".format(ifh))
BytesWarning: str() on a bytes instance
```


radarhere:
Could you provide a command to trigger the error?

nulano:
It appears to be the `-b` parameter: https://docs.python.org/3/using/cmdline.html#cmdoption-b

Running the test suite locally I get 332 warnings with it enabled, all caused by these two lines.

radarhere:
Thanks

PR #4756, opened 2020-07-02
Fix TiffDecode comparison warnings

nulano:
For #4586 

 * `rows_per_strip` and `tile_length` are already checked to be less than `INT_MAX`, so it is safe to cast them to `INT32`
 * add a check for `tile_width` to make sure `row_byte_size` does not overflow and make it safe to cast to `INT32`
 * `x`, `y`, `tile_y`, and `strip_row` are compared with signed ints or the values above only, it is safe to change their type to `INT32`

Before: https://travis-ci.org/github/python-pillow/Pillow/jobs/704275379#L2399-L2419
After: https://travis-ci.org/github/nulano/Pillow/jobs/704337888#L2314

### DIFF
--- a/src/libImaging/TiffDecode.c
+++ b/src/libImaging/TiffDecode.c
@@ -344,12 +344,19 @@ int ImagingLibTiffDecode(Imaging im, ImagingCodecState state, UINT8* buffer, Py_
     }
 
     if (TIFFIsTiled(tiff)) {
-        UINT32 x, y, tile_y, row_byte_size;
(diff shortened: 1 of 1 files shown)


radarhere:
For clarity - in your description, you've said `tiles_per_strip`, but you mean `rows_per_strip`.

nulano:
> For clarity - in your description, you've said `tiles_per_strip`, but you mean `rows_per_strip`.

Oops, thank you. Updated comment.

Issue #4764, opened 2020-07-07
Release Pillow 8.0.0 on October 15, 2020


hugovk:
A major bump because the next release will drop EOL Python 3.5 (https://github.com/python-pillow/Pillow/pull/4746). Will also be the first version to support Python 3.9.

Needs release notes:

* [x] https://github.com/python-pillow/Pillow/pull/4523
* [x] https://github.com/python-pillow/Pillow/pull/4955
* [x] https://github.com/python-pillow/Pillow/pull/4959
* [x] https://github.com/python-pillow/Pillow/pull/4964

---

# Release Checklist

## Main Release

Released quarterly on January 2nd, April 1st, July 1st and October 15th.

* [x] Open a release ticket e.g. https://github.com/python-pillow/Pillow/issues/3154
* [x] Develop and prepare release in `master` branch.
* [x] Check [GitHub Actions](https://github.com/python-pillow/Pillow/actions),
      [Travis CI](https://travis-ci.org/github/python-pillow/Pillow) and
      [AppVeyor](https://ci.appveyor.com/project/python-pillow/Pillow) to confirm
      passing tests in `master` branch.
* [x] Check that all of the wheel builds [Pillow Wheel Builder](https://github.com/python-pillow/pillow-wheels) pass the tests in Travis CI.
* [x] In compliance with [PEP 440](https://www.python.org/dev/peps/pep-0440/), update version identifier in `src/PIL/_version.py`
* [x] Update `CHANGES.rst`.
* [x] Run pre-release check via `make release-test` in a freshly cloned repo.
* [x] Create branch and tag for release e.g.:
  ```bash
  git branch 5.2.x
  git tag 5.2.0
  git push --all
  git push --tags
  ```
* [x] Create source distributions e.g.:
  ```bash
  make sdist
  ```
* [x] Create [binary distributions](https://github.com/python-pillow/Pillow/blob/master/RELEASING.md#binary-distributions)
* [x] Upload all binaries and source distributions e.g. `twine upload dist/Pillow-5.2.0*`
* [x] Create a [new release on GitHub](https://github.com/python-pillow/Pillow/releases/new)
* [x] In compliance with [PEP 440](https://www.python.org/dev/peps/pep-0440/), increment and append `.dev0` to version identifier in `src/PIL/_version.py`

### Windows
* [x] Contact `@cgohlke` for Windows binaries via release ticket e.g. https://github.com/python-pillow/Pillow/issues/1174.
[...]

Issue #4354, opened 2020-01-13
Release Pillow 7.1.0 on April 1, 2020


hugovk:
Needs release notes:
* [x] https://github.com/python-pillow/Pillow/pull/4230
* [x] https://github.com/python-pillow/Pillow/pull/3907 <- PR https://github.com/python-pillow/Pillow/pull/4502
* [x] https://github.com/python-pillow/Pillow/pull/4260
* [x] https://github.com/python-pillow/Pillow/pull/4243

# Release Checklist

## Main Release

Released quarterly on January 2nd, April 1st, July 1st and October 15th.

* [x] Open a release ticket e.g. https://github.com/python-pillow/Pillow/issues/3154
* [x] Develop and prepare release in `master` branch.
* [x] Check [Travis CI](https://travis-ci.org/python-pillow/Pillow) and [AppVeyor CI](https://ci.appveyor.com/project/python-pillow/Pillow) to confirm passing tests in `master` branch.
* [x] Check that all of the wheel builds [Pillow Wheel Builder](https://github.com/python-pillow/pillow-wheels) pass the tests in Travis CI.
* [x] In compliance with [PEP 440](https://www.python.org/dev/peps/pep-0440/), update version identifier in `src/PIL/_version.py`
* [x] Update `CHANGES.rst`.
* [x] Run pre-release check via `make release-test` in a freshly cloned repo.
* [x] Create branch and tag for release e.g.:
  ```bash
  git branch 5.2.x
  git tag 5.2.0
  git push --all
  git push --tags
  ```
* [x] Create source distributions e.g.:
  ```bash
  make sdist
  ```
* [x] Create [binary distributions](https://github.com/python-pillow/Pillow/blob/master/RELEASING.md#binary-distributions)
* [x] Upload all binaries and source distributions e.g. `twine upload dist/Pillow-5.2.0*`
* [x] Create a [new release on GitHub](https://github.com/python-pillow/Pillow/releases/new)
* [x] In compliance with [PEP 440](https://www.python.org/dev/peps/pep-0440/), increment and append `.dev0` to version identifier in `src/PIL/_version.py`

### Windows
* [x] Contact `@cgohlke` for Windows binaries via release ticket e.g. https://github.com/python-pillow/Pillow/issues/1174.
* [x] Download and extract tarball from `@cgohlke` and `twine upload *`.

### Mac and Linux
* [x] Use the [Pillow Wheel Builder](https://github.com/python-pillow/pillow-wheels):
[...]

PR #4770, opened 2020-07-09
ClangFormat the C code

### LINKED ISSUE #4493: ClangFormat the C code

hugovk:
Fixes #4493. 

I found a "clang-format style that approximates Python's [PEP 7](https://www.python.org/dev/peps/pep-0007/)":

* https://github.com/pganssle/zoneinfo/blob/master/.clang-format

(The zoneinfo repo was the reference implementation for `zoneinfo` in the stdlib (PEP 615), and is now a backport for Python 3.6+.)

This style has also been used elsewhere, as is or as the basis:

* https://github.com/search?q=%22A+clang-format+style+that+approximates+Python%27s+PEP+7%22&type=Code

Including CPython's new PEG generator (written by Guido and two others), which is the same as the zoneinfo one but with `ColumnLimit` increased from 79 to 95:

* https://github.com/python/cpython/blob/master/Tools/peg_generator/.clang-format

Here's how it looks applied to Pillow:

```sh
clang-format -i `git ls-tree -r HEAD --name-only | grep "\.[ch]$"`
```

1. The first commit is the same config file as in zoneinfo
2. The second formats the codebase with the config
3. The third increases `ColumnLimit` from 79 to 88, to match Black
4. The fourth formats again with the 88 limit

I suggest we use this, and welcome feedback and any further tweaks.

Before merge, to make the history cleaner, I'll squash the config to a single commit, and the formatting to another single commit.

Also this will need adding to the CI.


wiredfool:
I really dislike this form of function definition, especially since almost all other forms of braces are inline with the end of the starting line:

```
static PyObject*
_putdata(ImagingObject* self, PyObject* args)
{
```

The rest looks pretty reasonable. I think this is forcing braces around single line blocks (if/while), which is good. If it's not, then I'd say that it should.

nulano:
I think this ignores the following part of PEP 7:
> * Breaking long lines: if you can, break after commas in the outermost argument list. Always indent continuation lines appropriately, e.g.:
> ```c
> PyErr_Format(PyExc_TypeError,
>             "cannot create '%.100s' instances",
>             type->tp_name);
> ```

This is also a requirement in Python code in PEP 8 and Black.

hugovk:
> I really dislike this form of function definition, especially since almost all other forms of braces are inline with the end of the starting line:
> 
> ```c
> static PyObject*
> _putdata(ImagingObject* self, PyObject* args)
> {
> ```

This config sets `BreakBeforeBraces: Stroustrup`, overriding the Google value of `BreakBeforeBraces: Attach`.

The options ([docs with examples](https://clang.llvm.org/docs/ClangFormatStyleOptions.html)):

* Attach: Always attach braces to surrounding context.
* Linux: Like Attach, but break before braces on function, namespace and class definitions.
* Stroustrup: Like Attach, but break before function definitions.
* Allman: Always break before braces.
* GNU: Always break before braces and add an extra level of indentation to braces of control statements, not to those of class, function or other definitions.

Attach is the only one not to break before that brace:

```c
static PyObject *
_putdata(ImagingObject *self, PyObject *args) {
```

I've changed config to `BreakBeforeBraces: Attach` in ba2242157529e29fef96fbcad4d32c162876e6b9 and formatted with it in 4466b71e26c7db1780f6366c6ac5e8c208578498. Is it better?

hugovk:
> I think this is forcing braces around single line blocks (if/while), which is good. If it's not, then I'd say that it should.

I was hoping it would, to prevent bugs like https://github.com/python-pillow/Pillow/pull/4492, but alas it's out of scope for clang-format (https://stackoverflow.com/a/36871872/724176).

@radarhere fixed up the existing code in https://github.com/python-pillow/Pillow/pull/4618.

hugovk:
> I think this ignores the following part of PEP 7:
> 
> > * Breaking long lines: if you can, break after commas in the outermost argument list. Always indent continuation lines appropriately, e.g.:
> > 
> > ```c
> > PyErr_Format(PyExc_TypeError,
> >             "cannot create '%.100s' instances",
> >             type->tp_name);
> > ```
> 
> I think this is also a requirement in Python code in Black.

Looks like this can be achieved with (https://stackoverflow.com/a/47097008/724176):

```
AlignAfterOpenBracket: AlwaysBreak
BinPackArguments: false
BinPackParameters: false
```

Docs: https://clang.llvm.org/docs/ClangFormatStyleOptions.html
Config: f714356374b84fe6018b7ad4af576c45c0fbc7a5
Formatted: a48cc59e29433052411dab4482c99639d8d7379f

hugovk:
I've rebased, squashed the style modifications to a single commit, and reformatted.

src/_imaging.c
```diff
@@ -94,40 +94,38 @@
...
+#define S16(v) ((v) < 32768 ? (v) : ((v)-65536))
 
 /* -------------------------------------------------------------------- */
 /* OBJECT ADMINISTRATION                        */
 /* -------------------------------------------------------------------- */
 
 typedef struct {
-    PyObject_HEAD
-    Imaging image;
+    PyObject_HEAD Imaging image;
```
nulano, line 120: This doesn't look right. Perhaps the following should get formatted properly?
```suggestion
    PyObject_HEAD;
    Imaging image;
```
This change can be done in all files with a simple grep.

nulano (commented):
> Let's do this after the 2020-04-01 release (#4354), when the PR list will be at a local minimum and to avoid merge conflicts.

Maybe after 8.0.0 (#4764)?

src/_imaging.c
radarhere: Adding in the semicolon causes syntax errors on Windows (AppVeyor and GHA).

nulano: Ah great, leave it to MSVC to apply arbitrary restrictions from a standard it doesn't even fully support yet.

After a bit of effort, I have found the following hack which works with both MSVC and GCC 4.1 and has no effect on structure size (see https://godbolt.org/z/76W7f4):
```suggestion
    PyObject_HEAD int:0;
    Imaging image;
```
The extra code can be hidden in a define such as `#define PILLOW_HEAD PyObject_HEAD int:0` and used similarly to my original comment. However, I'm not sure whether it is worth the extra effort.

radarhere:
I don't have strong feelings about this.
We're coming up on a release. Is it a good time to merge?

src/_imaging.c
wiredfool: I think we should leave this if possible, PyObject_HEAD is the standard way to do it, and hiding it in a macro is additional complication that doesn’t make sense except for formatting.

wiredfool:
Let’s merge after release. Fewer potential merge errors and rebase required then.

PR #4746, opened 2020-06-30
Drop support for EOL Python 3.5

hugovk:
Python 3.5 reaches EOL on **2020-09-13**:

* https://devguide.python.org/#status-of-python-branches

The next Pillow release 8.0.0 release is due out after that, on **2020-10-15**.

This removes testing for Python 3.5 on the CIs, updates docs and metadata. I'll send another PR later with syntax updates, but first it's good to reduce the CI usage to speed up PR builds.

Edit: It also removes some redundant PyPy compatibility code for old and unsupported PyPy versions.

---

Notes:

* AppVeyor tests on two Python versions, the lowest supported (was 3.5, now 3.6) and the highest.

* Travis CI and GitHub Actions tested `PYTHONOPTIMIZE=2` on the 3.5 job (and `PYTHONOPTIMIZE=1` on 3.6). Now it tests `PYTHONOPTIMIZE=2` on the 3.7 job.

* Travis CI tested `s390x` on 3.5 (chosen arbitrarily). Now tests `s390x` on 3.8, as `arm64` is in 3.6 and `ppc64le` is on 3.7.

PR #4772, opened 2020-07-09
Use f-strings in winbuild

nulano:
For #4746 

 * Use f-strings in winbuild

The use of format strings in the winbuild script is somewhat complicated and I doubt an automated script could upgrade all uses properly, so I did it manually. Do not merge before #4746.


hugovk:
Thanks!

PR #4774, opened 2020-07-09
Fix documentation references: standard library

nulano:

 * Add `intersphinx_mapping` for standard library. A good place to check it out is the [attribute types in `ImageCms.CmsProfile`](https://pillow--4774.org.readthedocs.build/en/4774/reference/ImageCms.html#cmsprofile).
 * Fix typo in PSDraw
 * Use cross-reference instead of code for `None`. I used a grep to replace all instances in rst files, but I can see there are still a lot of references which are neither code or cross-reference. I also did not update docstrings due to line length limit. While linking to the docs for None might be a bit too much, it would be good to decide on a consistent standard to use for all standard library types. There are likely other types that I missed that should be replaced if this is the agreed-upon style.

<details><summary>Fixes 7 nitpicky warnings: (click to expand)</summary><p>

```diff
-C:\Git\Pillow\docs\handbook\tutorial.rst:468: WARNING: py:class reference target not found: io.BytesIO
-C:\Git\Pillow\docs\handbook\writing-your-own-file-decoder.rst:94: WARNING: py:exc reference target not found: struct.error
-c:\git\pillow\src\PIL\Image.py:docstring of PIL.Image.frombytes:10: WARNING: py:class reference target not found: io.BytesIO
-C:\Git\Pillow\docs\reference\ImageCms.rst:: WARNING: py:class reference target not found: datetime.datetime
-c:\git\pillow\src\PIL\PSDraw.py:docstring of PIL.PSDraw.PSDraw:1: WARNING: py:attr reference target not found: sys.stdout
-c:\git\pillow\src\PIL\TiffImagePlugin.py:docstring of PIL.TiffImagePlugin.IFDRational:1: WARNING: py:class reference target not found: numbers.Rational
-c:\git\pillow\src\PIL\TiffImagePlugin.py:docstring of PIL.TiffImagePlugin.ImageFileDirectory_v2:1: WARNING: py:class reference target not found: collections.abc.MutableMapping
```
</p></details>

### DIFF
--- a/src/PIL/PSDraw.py
+++ b/src/PIL/PSDraw.py
@@ -26,7 +26,7 @@ from . import EpsImagePlugin
 class PSDraw:
     """
     Sets up printing to the given file. If **fp** is omitted,
-    :py:attr:`sys.stdout` is assumed.
+    :py:data:`sys.stdout` is assumed.
     """
 
     def __init__(self, fp=None):
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -34,6 +34,8 @@ extensions = [
     "sphinx_removed_in",
 ]
 
+intersphinx_mapping = {"python": ("https://docs.python.org/3", None)}
+
 # The suffix(es) of source filenames.
 # You can specify multiple suffix as a list of string:
 # source_suffix = ['.rst', '.md']
--- a/docs/handbook/image-file-formats.rst
+++ b/docs/handbook/image-file-formats.rst
@@ -857,7 +857,7 @@ The :py:meth:`~PIL.Image.Image.save` method can take the following keyword argum
 **compression**
     A string containing the desired compression method for the
     file. (valid only with libtiff installed) Valid compression
-    methods are: ``None``, ``"tiff_ccitt"``, ``"group3"``,
+    methods are: :data:`None`, ``"tiff_ccitt"``, ``"group3"``,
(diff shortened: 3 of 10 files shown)


hugovk:
For reference:

> # `sphinx.ext.intersphinx` – Link to other projects’ documentation
>
> This extension can generate automatic links to the documentation of objects in other projects.
>
> Usage is simple: whenever Sphinx encounters a cross-reference that has no matching target in the current documentation set, it looks for targets in the documentation sets configured in `intersphinx_mapping`. A reference like `:py:class:`zipfile.ZipFile` can then link to the Python documentation for the ZipFile class, without you having to specify where it is located exactly.




https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html

src/PIL/PSDraw.py
```diff
@@ -26,7 +26,7 @@
 class PSDraw:
     """
     Sets up printing to the given file. If **fp** is omitted,
```
hugovk, line 28: Let's also put this in code formatting?
```suggestion
    Sets up printing to the given file. If ``fp``is omitted,
```

nulano: I'd prefer to do it in a separate PR, and keep this one only about enabling the intersphinx mapping specifically.

I imagine a grep for `\*\*[^ ]+\*\*` could find a lot of instances like this and would deserve its own PR anyway.

Issue #4769, opened 2020-07-09
Python 3.9-dev build failure


hugovk:
Between these two builds the `3.9-dev` build started failing on Travis CI:

* Yesterday: https://travis-ci.org/github/python-pillow/Pillow/jobs/706015838
* Today: https://travis-ci.org/github/hugovk/Pillow/jobs/706476038

```
Finished processing dependencies for Pillow==7.3.0.dev0
python3 selftest.py
Traceback (most recent call last):
  File "/home/travis/build/hugovk/Pillow/selftest.py", line 6, in <module>
    from PIL import Image, features
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 664, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 627, in _load_backward_compatible
  File "<frozen zipimport>", line 259, in load_module
  File "/home/travis/virtualenv/python3.9-dev/lib/python3.9/site-packages/Pillow-7.3.0.dev0-py3.9-linux-x86_64.egg/PIL/Image.py", line 94, in <module>
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 664, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 627, in _load_backward_compatible
  File "<frozen zipimport>", line 259, in load_module
  File "/home/travis/virtualenv/python3.9-dev/lib/python3.9/site-packages/Pillow-7.3.0.dev0-py3.9-linux-x86_64.egg/PIL/_imaging.py", line 8, in <module>
  File "/home/travis/virtualenv/python3.9-dev/lib/python3.9/site-packages/Pillow-7.3.0.dev0-py3.9-linux-x86_64.egg/PIL/_imaging.py", line 7, in __bootstrap__
TypeError: exec_module() missing 1 required positional argument: 'module'
Makefile:59: recipe for target 'install-coverage' failed
make: *** [install-coverage] Error 1
Traceback (most recent call last):
  File "/home/travis/virtualenv/python3.9-dev/lib/python3.9/site-packages/_pytest/config/__init__.py", line 495, in _importconftest
    return self._conftestpath2mod[key]
[...]

PR #4777, opened 2020-07-10
Travis CI: Allow 3.9-dev to fail

hugovk:
We should figure out https://github.com/python-pillow/Pillow/issues/4769, but shall we not let pre-releases fail PRs?

image

* https://docs.travis-ci.com/user/customizing-the-build#jobs-that-are-allowed-to-fail

(For some visibility we could use a conditional: so PRs can pass but `master` would still fail?)

* https://docs.travis-ci.com/user/customizing-the-build#conditionally-allowing-jobs-to-fail

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,6 +46,9 @@ matrix:
       env: PYTHONOPTIMIZE=2
       services: xvfb
 
+  allow_failures:
+  - python: "3.9-dev"
+
 install:
   - |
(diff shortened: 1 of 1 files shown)


radarhere:
I think this is a reasonable move to reduce the likelihood that all PRs spontaneously start failing - like they are now.

Personally, I feel like you still notice the failing dev build when creating a PR - but also happy for the conditional to be added.

Issue #2806, opened 2017-10-18
Support for HEIF


pythonicrubyist:
Do you have any plans to support HEIF image format, recently introduced by apple?

wiredfool:
I would welcome support, but I'm not aware of any appropriately licensed HEIF or HEVC libraries to do encoding/decoding.

CounterPillow:
The [reference HEVC encoder/decoder](https://hevc.hhi.fraunhofer.de/) is licensed as BSD 3-clause (but obviously extremely slow and tuned for PSNR), [the openHEVC decoder](https://github.com/OpenHEVC/openHEVC) is licensed as LGPL2.1.

Patent licenses are the bigger issues; the current patent licensing situation with 3 competing pools means that it'd be practically impossible for any FOSS project to produce binaries including a HEVC decoder which don't violate some patents.

So even if someone were to do the work to implement it, you wouldn't find it in most Pillow builds, because nobody wants to get harassed by the patent trolls.

wiredfool:
The HEIF container decoder is also not a free license: https://github.com/nokiatech/heif/blob/master/LICENSE.TXT .

CounterPillow:
There's an initial HEIF support patch for FFmpeg floating around [on their mailing list](https://ffmpeg.org/pipermail/ffmpeg-devel/2017-August/215003.html), but as the post itself says, it's early work and the whole format is batshit insane. The container is based off of MP4 for a start.

I don't think it'll be possible to support HEIF in Pillow in a sane manner based on the container and codec complexity alone, and even if someone were to make a demuxer+decoder in one library with a license Pillow could use, you wouldn't be able to build a product on it for around 30 more years, assuming the patents remain unlicenseable for anyone without big money lawyers.

wiredfool:
I'm definitely -1 on batshit insane. 

I think that the only way it could work is if we shipped support that was dynamically linked and based off of OS libraries, e.g. CoreImage on OSX. But it's not clear that would work, and it would only be the one platform at this point. 

I'm going to close this, if the world comes around to the point that it's reasonable, we can look at it.

rspeed:
What about [libheif](https://github.com/strukturag/libheif)? It takes care of the packaging while using [libde265](http://www.libde265.org) for reading and [x265](http://x265.org/) for writing. All three libraries are LGPL.

[...]

PR #4782, opened 2020-07-11
HEIF plugin

david-poirier-csn:
Hey, this is a *DRAFT* version of a HEIF plugin using `pyheif`. Just running it up the flagpole ;)
- Is this wanted?
- Am I on the right track with the plugin design?

If so I'll proceed and add exif, color profile, doco and tests.

#2806


radarhere:
I think that given the licensing issues, this is the only way forward, and presuming it works, then thanks very much for thinking of this. There is certainly a large number of people requesting this feature over in #2806

@wiredfool, did you have any thoughts on this approach?

david-poirier-csn:
Re: the licensing issues, `libheif` now supports AVIF (using libaom) in addition to HEIC and thus so does `pyheif` (if built from source, the binary wheels don't *yet*). In a nutshell HEIF is the container format, HEIC is a HEIF file containing at least one HEVC (h.265) keyframe, and AVIF is a HEIF file containing at least one AV1 keyframe. Chrome and Firefox are both on the cusp of supporting AVIF. So this is a "twofer" with potential to be even more than that if other image formats adopt the HEIF container format.

On another note I should point out that `pyheif` is Python 3.6+ only, and Linux + MacOS only. I don't know if `libheif` can be built on Windows without significant effort.

hugovk:
> On another note I should point out that `pyheif` is Python 3.6+ only, and Linux + MacOS only. I don't know if `libheif` can be built on Windows without significant effort.

Python 3.5 will be EOL and dropped in the next release: https://github.com/python-pillow/Pillow/pull/4746.

nulano:
> I don't know if libheif can be built on Windows without significant effort.

A brief search turned up with an [ImageMagick libheif repository](https://github.com/ImageMagick/libheif) which claims to be intended for building libheif on Windows, so it should be possible:

> ## ImageMagick / libheif
> ### About
> libheif for the Windows build of ImageMagick

If this gets merged, I am willing to *attempt* to contribute scripts for Windows builds to pyheif (e.g. using GitHub Actions).

jolyonralph:
Just a quick comment to the first question you asked - yes it IS wanted! 

Thank you!

src/PIL/HeifImagePlugin.py
```diff
@@ -0,0 +1,36 @@
...
+    return check != pyheif.heif_filetype_no
+
+
+class HeifImageFile(ImageFile.ImageFile):
+    format = "HEIF"
+    format_description = "HEIF Image File Format"
+
+    def _open(self):
+        b = self.fp.read()
+        if not _accept(b):
```
wiredfool, line 21: In general we should not be reading the whole file in _open. It does appear that pyheif does that, which is somewhat disappointing. I'm not sure if it's possible to get around that, except for the initial magic check.

src/PIL/HeifImagePlugin.py
```diff
@@ -0,0 +1,36 @@
...
+            raise ValueError("Not a HEIF image file")
+
+        heif_file = pyheif.read(b)
+        self._size = heif_file.size
+        self.mode = heif_file.mode
+        self.fp = io.BytesIO(heif_file.data)
```
wiredfool, line 27: The contract on _open is that we don't do expensive operations, so decoding the entire images and storing a copy violates that. 

Unfortunately, it doesn't appear that pyheif makes it easy to get the metadata that we require for the _open command without decoding the entire image.

wiredfool (commented):
Sorry this is a stale review. 2020. 

I think this is a reasonable approach from a licensing point of view, and it should be shippable that way. I'm not sure that we'd ever get libheif in the binary builds, but this would be the sort of thing that would work at arms length if someone were to install pyheif separately. 

I'm a concerned with the amount of work that's done in open. I think it might be worth seeing if the pyheif project would be amenable to having a basic call into the library for size/mode/etc without the load of a full decode.

src/PIL/HeifImagePlugin.py
david-poirier-csn: Thanks, I thought I did have a metadata-only call but maybe I'm not using it. I'll check it out.

myers:
See https://pypi.org/project/pyheif-pillow-opener/

david-poirier-csn:
Haha yes, I think it will be easier to register like this from `pyheif`, rather than try to integrate with `Pillow` due to H.265 patent issues etc

I will still try to implement @wiredfool's advice, much appreciated

homm:
To whom it may concern. I've made [heif-image-plugin](https://pypi.org/project/heif-image-plugin/) where fixed many issues including described here. For example:

* `im.info['icc_profile']` is now accessible 
* works without decoding whole image on open
* Opens AVIF files as well

Issue #4775, opened 2020-07-09
Docs: nitpicky warnings


nulano:
After merging #4739, #4771, #4773, #4774, there are still 18 nitpicky warnings (down from 400!) that I don't know how to fix:

```
C:\Git\Pillow\docs\handbook\image-file-formats.rst:186: WARNING: py:attr reference target not found: PIL.Image.Image.tile
C:\Git\Pillow\docs\handbook\image-file-formats.rst:711: WARNING: py:meth reference target not found: PIL.Image.Image.convert2byte
C:\Git\Pillow\docs\handbook\image-file-formats.rst:762: WARNING: py:attr reference target not found: PIL.Image.Image.tag
C:\Git\Pillow\docs\handbook\image-file-formats.rst:775: WARNING: py:attr reference target not found: PIL.Image.Image.tag_v2
C:\Git\Pillow\docs\handbook\image-file-formats.rst:784: WARNING: py:attr reference target not found: PIL.Image.Image.tag
C:\Git\Pillow\docs\handbook\image-file-formats.rst:823: WARNING: py:attr reference target not found: PIL.TiffImagePlugin.ImageFileDirectory_v2.tagtype
C:\Git\Pillow\docs\handbook\image-file-formats.rst:851: WARNING: py:attr reference target not found: PIL.TiffImagePlugin.ImageFileDirectory_v2.tagtype
C:\Git\Pillow\docs\handbook\writing-your-own-file-decoder.rst:29: WARNING: py:meth reference target not found: _open
C:\Git\Pillow\docs\handbook\writing-your-own-file-decoder.rst:29: WARNING: py:attr reference target not found: tile
C:\Git\Pillow\docs\handbook\writing-your-own-file-decoder.rst:39: WARNING: py:meth reference target not found: _open
C:\Git\Pillow\docs\handbook\writing-your-own-file-decoder.rst:106: WARNING: py:attr reference target not found: tile
C:\Git\Pillow\docs\handbook\writing-your-own-file-decoder.rst:109: WARNING: py:attr reference target not found: tile
C:\Git\Pillow\docs\handbook\writing-your-own-file-decoder.rst:137: WARNING: py:attr reference target not found: tile
C:\Git\Pillow\docs\reference\ImageMath.rst:63: WARNING: py:func reference target not found: PIL.ImageMath.convert
C:\Git\Pillow\docs\reference\ImageMath.rst:63: WARNING: py:func reference target not found: PIL.ImageMath.float
[...]

PR #4785, opened 2020-07-14
Updated references

radarhere:
Suggestions for #4775 

- Corrected reference
    - Replaced 'PIL.Image.Image.convert2byte' with 'PIL.SpiderImagePlugin.SpiderImageFile.convert2byte'
- Removed domain references from image expressions
    - 'convert()', 'float()' and 'int()' are not Python methods in this context
- Removed domain references when referring to hypothetical code
    - In the context of writing your own file decoder, `_open` and `tile` are only hypothetical.
- Removed domain reference for external method
    - `CDC.GetHandleAttrib()` is in PythonWin, not Pillow

### DIFF
--- a/src/PIL/ImageWin.py
+++ b/src/PIL/ImageWin.py
@@ -88,8 +88,8 @@ class Dib:
         Copy the bitmap contents to a device context.
 
         :param handle: Device context (HDC), cast to a Python integer, or an
-                       HDC or HWND instance.  In PythonWin, you can use the
-                       :py:meth:`CDC.GetHandleAttrib` to get a suitable handle.
+                       HDC or HWND instance.  In PythonWin, you can use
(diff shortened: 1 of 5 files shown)


src/PIL/ImageWin.py
```diff
@@ -88,8 +88,8 @@ def expose(self, handle):
         Copy the bitmap contents to a device context.
 
         :param handle: Device context (HDC), cast to a Python integer, or an
-                       HDC or HWND instance.  In PythonWin, you can use the
-                       :py:meth:`CDC.GetHandleAttrib` to get a suitable handle.
```
nulano, line 92: It would be ideal to link to PythonWin docs for the function, but I can't find them, so this is probably fine.

PR #4788, opened 2020-07-15
Add MIME type to PsdImagePlugin

### LINKED ISSUE #4787: PSD Plugin does not register a MIME type
The [`PSDImagePlugin`](https://github.com/python-pillow/Pillow/blob/master/src/PIL/PsdImagePlugin.py) does not register a MIME type as I'd expect it to. The correct MIME for PSD images, according to IANA, is ["image/vnd.adobe.photoshop"](https://www.iana.org/assignments/media-types/image/vnd.adobe.photoshop).

Is there a reason this isn't registered?

samamorgan:
Fixes #4787

 * Register "image/vnd.adobe.photoshop" as PsdImageFile MIME type

### DIFF
--- a/Tests/test_file_psd.py
+++ b/Tests/test_file_psd.py
@@ -12,6 +12,7 @@ def test_sanity():
         assert im.mode == "RGB"
         assert im.size == (128, 128)
         assert im.format == "PSD"
+        assert im.get_format_mimetype() == "image/vnd.adobe.photoshop"
 
         im2 = hopper()
         assert_image_similar(im, im2, 4.8)
--- a/src/PIL/PsdImagePlugin.py
+++ b/src/PIL/PsdImagePlugin.py
@@ -307,3 +307,5 @@ def _maketile(file, mode, bbox, channels):
 Image.register_open(PsdImageFile.format, PsdImageFile, _accept)
 
 Image.register_extension(PsdImageFile.format, ".psd")
+
+Image.register_mime(PsdImageFile.format, "image/vnd.adobe.photoshop")


hugovk:
Please could you add tests similar to https://github.com/python-pillow/Pillow/pull/3525? 

Also, https://wordpress.stackexchange.com/a/212334 and https://www.whatisfileextension.com/psd/ suggest there's more in use:

```
image/vnd.adobe.photoshop

application/x-photoshop

application/photoshop

application/psd

image/psd
```

Should these be added too?

samamorgan:
@radarhere Thanks for the quick check-in!

I'm not sure a mechanism exists for multiple mimetypes in this library? I'll look for examples of this in existing code and emulate that if possible. That being said, if there isn't a mechanism for this, I'd prefer we use "image/vnd.adobe.photoshop" as it's the registered IANA type (noted in issue #4787).

I'll work on tests!

samamorgan:
Looking at all of the image plugins, I only see one mimetype registered for each, so without major restructures around how mimetypes work in this library, I think we'll have to choose one and go with it.

radarhere:
Just for the record, while there isn't a mechanism for multiple simultaneous mimetypes, there is a way to [customise a mimetype based on the image](https://github.com/python-pillow/Pillow/blob/2b441edf38b4a72ff1978c35215c46140776dc48/src/PIL/SgiImagePlugin.py#L98) if there is a more specific mimetype that can be used. However, that doesn't look to be the case here - it would appear that there is no distinction between the various Photoshop mimetypes.

hugovk:
(The s390x failure on Travis is an intermittent Travis problem.)

Thanks!

PR #4791, opened 2020-07-16
Fix expected failures on big endian systems without CI

kulikjak:

 * remove `on_ci` checks from endianess related `xfail` decorators. It doesn't matter whether tests are running on CI or not for those tests to fail (we are running those tests without CI).

### DIFF
--- a/Tests/test_file_jpeg2k.py
+++ b/Tests/test_file_jpeg2k.py
@@ -8,7 +8,6 @@ from .helper import (
     assert_image_equal,
     assert_image_similar,
     is_big_endian,
-    on_ci,
     skip_unless_feature,
 )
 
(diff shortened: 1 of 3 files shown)


radarhere:
These were added in #4336. I presume the intention by @hugovk was to not prevent our CI from running other tests, but also not to hide the failures from end users. However, I'm guessing that you feel that marking them as xfail is sufficient?

hugovk:
Yes, that's about right, and I think xfail is sufficient as they'll still be shown in the logs:

```
XFAIL Tests/test_file_jpeg2k.py::test_16bit_monochrome_jp2_like_tiff
  Fails on big-endian
XFAIL Tests/test_file_jpeg2k.py::test_16bit_monochrome_j2k_like_tiff
  Fails on big-endian
XFAIL Tests/test_file_palm.py::test_p_mode
  Palm P image is wrong
XFAIL Tests/test_file_png.py::TestFilePng::test_sanity
  Fails on big-endian
XFAIL Tests/test_file_webp_animated.py::test_write_animation_L
  Fails on big-endian
```
https://travis-ci.org/github/python-pillow/Pillow/jobs/708663515#L4544

It's good people are testing big-endian locally too, maybe someone will be able to fix them? :)

Thanks for the PR!

Issue #4730, opened 2020-06-26
Antivirus Avast detect  test  image crash_2.tif as with  exploit 


danieltomasz:
### What did you do?
My Antivirus (Avast Shield alert) keep reporting and removing file crash_2.tif to my virus chest because it has vulnerability  with TIFF:CVE-2015-5097 [Expl] . 
This image resides on path `/private/var/folders/..../unpacked/Pillow-7.1.2/Tests/images/crash_2.tif`

### What did you expect to happen?

Any file distributed with Pillow will not be reported as an exploit.

### What are your OS, Python and Pillow versions?

* OS: MacOS 10.15.5
* Python: 3.8.2
* Pillow: 7.1.2

```python
code goes here
```

wiredfool:
They're not wrong, but it's not helpful.   Anything that's named `crash_` is a reproduction of a crashing bug in Pillow, most of them have CVEs.

hugovk:
And this one was specifically for CVE-2020-10379 "Fix two buffer overflows in TIFF decoding":

* https://pillow.readthedocs.io/en/latest/releasenotes/7.1.0.html#security

danieltomasz:
> They're not wrong, but it's not helpful. Anything that's named `crash_` is a reproduction of a crashing bug in Pillow, most of them have CVEs.

I am not importing Pillow directly, it was installed as a dependency for other packages. If I could isolate more I will post it here.

radarhere:
It's a vulnerability for a past version of Pillow. It is not a vulnerability for Pillow 7.1.2 that you have.

radarhere:
Also, to be clear, the image file can be removed by your antivirus software and Pillow will continue to work without a problem. The image would only be used if you ran a specific Pillow test script.

danieltomasz:
If this is solved already, the issue could be closed ( I filled it to let others know about the issue) I removed the aforementioned  file and could use Pillow without problems

Zen-CODE:
I've also just hit this issue when building pillow for a Kivy python deployment on iOS. Got it whilst trying to update from version 6.2.0 to 7.20. And sorry, deleting the file after deployment is not a solution. It's a workaround.

Consider if someone is building a Kivy app on MacOS for the first time and they sees this Virus warning immediately when they build?  It's a terrible impression, and they will probably cancel and uninstall everything. Please could you consider a proper solution, that avoids this warning? i.e. removing the files or updating with non-vulnerable ones?

[...]

PR #4792, opened 2020-07-16
Download CVE images instead of storing them on disk

radarhere:
Proof of concept solution for #4730 

The strategy suggested here is to remove the images from Pillow, and instead store them in pillow-depends. To ensure that they are still run each time though, instead of optionally downloading that repository, the tests could download each image as needed.

Obviously the URL need to be changed in this PR, and it could be more widely implemented. Just wanted to create this to provide clarity.

### DIFF
--- a/Tests/check_tiff_crashes.py
+++ b/Tests/check_tiff_crashes.py
@@ -14,16 +14,20 @@
 # version.
 
 
+import urllib.request
+
 from PIL import Image
 
 repro_read_strip = (
-    "images/crash_1.tif",
-    "images/crash_2.tif",
+    "crash_1.tif",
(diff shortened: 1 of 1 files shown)


wiredfool:
Tests shouldn't rely on the network being available -- The tests need to be deterministic based on what's on the disk. 

Either:
1) It should stay, it's important
2) It should go in the additional test images repo
3) It should be in a pw protected zip/xor'd file with a known password like 'sudo_ignore_this_avast_this_is_a_false_positive'

(fwiw, tests that rely on network have been a pain over the last months as I've been on connections that are flakey to one level or other, and cell hotspots aren't much better)

PR #4793, opened 2020-07-16
Removed Docker images running Python 3.5

radarhere:
As suggested in https://github.com/python-pillow/Pillow/pull/4746#issuecomment-659313021


docs/installation.rst
```diff
@@ -405,7 +403,7 @@ These platforms are built and tested for every change.
 +----------------------------------+--------------------------+-----------------------+
 | macOS 10.15 Catalina             | 3.6, 3.7, 3.8, PyPy3     |x86-64                 |
 +----------------------------------+--------------------------+-----------------------+
-| Ubuntu Linux 16.04 LTS           | 3.5, 3.6, 3.7, 3.8, PyPy3|x86-64                 |
+| Ubuntu Linux 16.04 LTS           | 3.6, 3.7, 3.8, PyPy3     |x86-64                 |
```
hugovk, line 406: Thanks!

Could we add Xenial, Bionic and Focal whilst we're here?

```suggestion
| Ubuntu Linux 16.04 LTS (Xenial)  | 3.6, 3.7, 3.8, PyPy3     |x86-64                 |
```

Issue #4796, opened 2020-07-16
Replace distutils


hugovk:
There's an effort underway to move the functionality of distutils into Setuptools and other libraries, with a goal to remove distutils from the standard library.

Re: 

> (Though replacing distutils with setuptools wherever possible is worthwhile irrespective of the distutils injection strategy.)

* https://github.com/pypa/setuptools/issues/2261#issuecomment-658214508

> `setuptools` will be the canonical implementation of "distutils" in the future. Some of the `distutils` hierarchy already is duplicated in `setuptools` (like commands), the parts that are likely to remain supported will likely move into the `setuptools` (I suspect that will include things like exceptions) namespace, and some stuff that is no longer necessary will be deprecated and removed.
> 
> If you need something from the `distutils` namespace and it's not available in `setuptools`, continue to use it until we provide an alternative (I would say "raise an issue" but I am not sure if we're going to just do a bulk migration or not), but if there's already an equivalent in `setuptools`, switching over to using `setuptools` now is the best thing you can do (and you _should_ report any issues you encounter when doing this).

* https://github.com/pypa/setuptools/issues/2261#issuecomment-658315975

> ... `setuptools` is not exposing its version of `distutils` as `setuptools.distutils` (and to the extent that it exists under the `setuptools` namespace, that location is not necessarily stable). `setuptools` replaces the top-level `distutils` module with its own version of `distutils`, so `import distutils` pulls the version from `setuptools`.
> 
> This is one phase of an ongoing project to remove `distutils` from the standard library (and probably eventually retire the project entirely in favor of `setuptools`).

* https://github.com/numpy/numpy/pull/16851#issuecomment-658321642

---

Here's where we're using disutils:

| File | Use | Replacement |
| - | - | - |
| `Tests/test_image_access.py` and `setup.py` | `from distutils import ccompiler, sysconfig` | #4809, #4814, #4817 |
| `Tests/test_imagefont.py` | `distutils.version.StrictVersion` | [`packaging.version.parse`](https://stackoverflow.com/a/11887885/724176) #4797 |
| `setup.py` | `from distutils.command.build_ext import build_ext` | `from setuptools.command.build_ext import build_ext` #4829 |
| `setup.py` | `from distutils import cygwinccompiler` | #4890 |

radarhere:
I'm not certain that we need to remove `from distutils import cygwinccompiler`. Some thoughts.

[...]

PR #4797, opened 2020-07-16
Replace distutils.version with packaging.version.parse

hugovk:
For https://github.com/python-pillow/Pillow/issues/4796.

 * distutils will eventually be removed from the stdlib
 * Replace `distutils.version` with `packaging.version.parse`
 * > `packaging.version.parse` is a third-party utility but is used by [setuptools][4] (so you probably already have it installed) and is conformant to the current [PEP 440][5]; it will return a `packaging.version.Version` if the version is compliant and a `packaging.version.LegacyVersion` if not. The latter will always sort before valid versions.
 * https://stackoverflow.com/a/11887885/724176

  [4]: https://pypi.python.org/pypi/setuptools
  [5]: https://www.python.org/dev/peps/pep-0440/

### DIFF
--- a/Tests/test_imagefont.py
+++ b/Tests/test_imagefont.py
@@ -1,5 +1,4 @@
 import copy
-import distutils.version
 import os
 import re
 import shutil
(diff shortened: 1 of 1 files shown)


radarhere:
Thanks

PR #4798, opened 2020-07-17
Remove long-deprecated Image.py functions

hugovk:
Our [current approach](https://pillow.readthedocs.io/en/latest/deprecations.html) has been to deprecate for at least a year, and then remove in the next major version. I think this has been working fine (except for `PILLOW_VERSION` which didn't raise deprecation warnings; so we reinstated it with warnings).

`Image.tostring` and `fromstring` were deprecated in 2013, and `Image.offset` in 2001. They've raised either a `DeprecationWarning`, `Exception` or `NotImplementedError` for a long time. Let's remove them.

---

# im.tostring and fromstring

## 2000: added

`Image.tostring`, `im.tostring` and `im.fromstring` date back to at least PIL 1.1.1 (Oct 2000), probably earlier.

## 2013: DeprecationWarning

In Pillow 2.0.0 (March 2013), when adding support for Python 3, `tobytes` was added, and `tostring` was deprecated, raising `DeprecationWarning: tostring() is deprecated. Please call tobytes() instead.` And similarly with `fromstring`/`frombytes` (https://github.com/python-pillow/Pillow/pull/35 & https://github.com/python-pillow/Pillow/pull/71).

## 2015: Exception

In Pillow 3.0.0 (Oct 2015), the `DeprecationWarning` was replaced with an `Exception` (https://github.com/python-pillow/Pillow/pull/1343).

## 2016: NotImplementedError

In Pillow 3.3.0 (July 2016), the `Exception` was replaced with a `NotImplementedError` (https://github.com/python-pillow/Pillow/pull/1862).

`NotImplementedError` means this, and doesn't quite fit:

> In user defined base classes, abstract methods should raise this exception when they require derived classes to override the method, or while the class is being developed to indicate that the real implementation still needs to be added.
>
> Note: It should not be used to indicate that an operator or method is not meant to be supported at all – in that case either leave the operator / method undefined or, if a subclass, set it to None.

https://docs.python.org/3/library/exceptions.html#NotImplementedError

# im.offset

## 2001: NotImplementedError

`Image.offset` followed a similar path, except it was documented as deprecated in PIL 1.1.2 (May 2001), raised a `DeprecationWarning` in 1.1.5 (March 2005), then as above: `Exception` in Pillow 3.0.0 (Oct 2015), `NotImplementedError` in 3.3.0 (July 2016).

### DIFF
--- a/docs/deprecations.rst
+++ b/docs/deprecations.rst
@@ -68,8 +68,8 @@ raised a ``DeprecationWarning`` since 1.1.5,
 an ``Exception`` since Pillow 3.0.0
 and ``NotImplementedError`` since 3.3.0.
 
-im.fromstring and tostring
-~~~~~~~~~~~~~~~~~~~~~~~~~~
+Image.fromstring, im.fromstring and im.tostring
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 .. deprecated:: 2.0.0
 .. versionremoved:: 8.0.0
--- a/docs/deprecations.rst
+++ b/docs/deprecations.rst
@@ -55,6 +55,33 @@ Removed features
 Deprecated features are only removed in major releases after an appropriate
 period of deprecation has passed.
 
+im.offset
+~~~~~~~~~
+
+.. deprecated:: 1.1.2
+.. versionremoved:: 8.0.0
+
+``im.offset()`` has been removed, call ``ImageChops.offset()`` instead.
+
+It was documented as deprecated in PIL 1.1.2,
+raised a ``DeprecationWarning`` since 1.1.5,
+an ``Exception`` since Pillow 3.0.0
+and ``NotImplementedError`` since 3.3.0.
+
+im.fromstring and tostring
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. deprecated:: 2.0.0
+.. versionremoved:: 8.0.0
+
+* ``Image.fromstring()`` has been removed, call ``frombytes()`` instead.
+* ``im.fromstring()`` has been removed, call ``frombytes()`` instead.
+* ``im.tostring()`` has been removed, call ``tobytes()`` instead.
+
+They issued a ``DeprecationWarning`` since 2.0.0,
+an ``Exception`` since 3.0.0
+and ``NotImplementedError`` since 3.3.0.
+
 ImageCms.CmsProfile attributes
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/docs/deprecations.rst
+++ b/docs/deprecations.rst
@@ -74,9 +74,9 @@ im.fromstring and tostring
 .. deprecated:: 2.0.0
 .. versionremoved:: 8.0.0
 
-* ``Image.fromstring()`` has been removed, call ``frombytes()`` instead.
-* ``im.fromstring()`` has been removed, call ``frombytes()`` instead.
-* ``im.tostring()`` has been removed, call ``tobytes()`` instead.
+* ``Image.fromstring()`` has been removed, call :py:func:`.Image.frombytes()` instead.
+* ``im.fromstring()`` has been removed, call :py:meth:`~PIL.Image.Image.frombytes()` instead.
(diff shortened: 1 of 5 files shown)


docs/deprecations.rst
```diff
@@ -55,6 +55,33 @@ Removed features
...
+
+im.fromstring and tostring
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. deprecated:: 2.0.0
+.. versionremoved:: 8.0.0
+
+* ``Image.fromstring()`` has been removed, call ``frombytes()`` instead.
+* ``im.fromstring()`` has been removed, call ``frombytes()`` instead.
+* ``im.tostring()`` has been removed, call ``tobytes()`` instead.
```
radarhere, line 79:
```suggestion
* ``Image.fromstring()`` has been removed, call ``Image.frombytes()`` instead.
* ``im.fromstring()`` has been removed, call ``im.frombytes()`` instead.
* ``im.tostring()`` has been removed, call ``im.tobytes()`` instead.
```
I'm just slightly concerned that someone would read through your version and think that the two mentions of ``frombytes()`` are referring to the same function.

nulano:
```suggestion
* ``Image.fromstring()`` has been removed, call :py:func:`.Image.frombytes()` instead.
* ``im.fromstring()`` has been removed, call :py:meth:`~PIL.Image.Image.frombytes()` instead.
* ``im.tostring()`` has been removed, call :py:meth:`~PIL.Image.Image.tobytes()` instead.
```

How about providing links instead?
Also need to add `PIL.Image.Image.frombytes` to `Image.rst`, see my other comment.

docs/reference/Image.rst
```diff
@@ -263,8 +261,6 @@ This flips the input image by using the :data:`FLIP_LEFT_RIGHT` method.
 
 .. automethod:: PIL.Image.Image.verify
 
-.. automethod:: PIL.Image.Image.fromstring
```
nulano, line 266: `PIL.Image.Image.frombytes` is undocumented. Perhaps change this to `frombytes` instead of removing it.

docs/deprecations.rst
nulano:
```suggestion
* ``Image.fromstring()`` has been removed, call :py:func:`.Image.frombytes()` instead.
* ``im.fromstring()`` has been removed, call :py:meth:`~PIL.Image.Image.frombytes()` instead.
* ``im.tostring()`` has been removed, call :py:meth:`~PIL.Image.Image.tobytes()` instead.
```

How about also linking to the new functions?
Also need to add documentation for `Image.Image.frombytes`, see my other comment.

*PS: I'm not sure why my comment disappeared before, I may have deleted it accidentally.*

docs/deprecations.rst
```diff
@@ -55,6 +55,33 @@ Removed features
...
+
+``im.offset()`` has been removed, call ``ImageChops.offset()`` instead.
+
+It was documented as deprecated in PIL 1.1.2,
+raised a ``DeprecationWarning`` since 1.1.5,
+an ``Exception`` since Pillow 3.0.0
+and ``NotImplementedError`` since 3.3.0.
+
+im.fromstring and tostring
+~~~~~~~~~~~~~~~~~~~~~~~~~~
```
radarhere, line 72: `Image.fromstring` is left out by this heading.
```suggestion
Image.fromstring, im.fromstring and im.tostring
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

docs/reference/Image.rst
```diff
@@ -151,6 +151,7 @@ This crops the input image with the provided coordinates:
 .. automethod:: PIL.Image.Image.draft
 .. automethod:: PIL.Image.Image.effect_spread
 .. automethod:: PIL.Image.Image.entropy
+.. automethod:: PIL.Image.Image.frombytes
 .. automethod:: PIL.Image.Image.filter
```
radarhere, line 155: If this is going for alphabetical order. then...
```suggestion
.. automethod:: PIL.Image.Image.filter
.. automethod:: PIL.Image.Image.frombytes
```

PR #4799, opened 2020-07-17
GHA: Test Python 3.9-dev

nulano:
GHA now provides Python 3.9 beta 4 ~if the full version number `3.9.0-beta.4` is specified~, add it to the matrix.
(https://github.com/actions/setup-python/issues/20#issuecomment-660026867)

~Or perhaps this can wait until beta 5 is available (possibly later this week?).~

Edit: Actually, specifying `3.9-dev` also seems to work fine.

### DIFF
--- a/.github/workflows/test-windows.yml
+++ b/.github/workflows/test-windows.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "pypy3"]
+        python-version: ["3.6", "3.7", "3.8", "3.9-dev", "pypy3"]
         architecture: ["x86", "x64"]
         include:
           - architecture: "x86"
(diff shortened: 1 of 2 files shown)


hugovk:
No mention of `3.9-dev` in https://github.com/actions/setup-python/issues/20#issuecomment-660026867 (or the links), but great it works!

nulano:
I found it by searching through the action's sources before they were done implementing support (the converted value was not yet supported by the backend when I tried it before):

https://github.com/actions/setup-python/blob/7010ec794f55ed6e9a6b0fa34cdfcc5bbc6137b2/src/find-python.ts#L139-L147

nulano:
GHA have just released beta 5 and it has already been picked up by the `3.9-dev` identifier in #4802 tests: https://github.com/python-pillow/Pillow/pull/4802/checks?check_run_id=894275587#step:7:8

PR #4802, opened 2020-07-20
Documented MAX_IMAGE_PIXELS

radarhere:
Helps #4775

As well as being a reasonable addition to the documentation in its own right.

### DIFF
--- a/docs/reference/Image.rst
+++ b/docs/reference/Image.rst
@@ -53,10 +53,16 @@ Functions
         To protect against potential DOS attacks caused by "`decompression bombs`_" (i.e. malicious files
         which decompress into a huge amount of data and are designed to crash or cause disruption by using up
         a lot of memory), Pillow will issue a ``DecompressionBombWarning`` if the number of pixels in an
-        image is over a certain limit, :py:data:`PIL.Image.MAX_IMAGE_PIXELS`. If desired, the warning can be
-        turned into an error with ``warnings.simplefilter('error', Image.DecompressionBombWarning)`` or
-        suppressed entirely with ``warnings.simplefilter('ignore', Image.DecompressionBombWarning)``. See
-        also `the logging documentation`_ to have warnings output to the logging facility instead of stderr.
+        image is over a certain limit, :py:data:`PIL.Image.MAX_IMAGE_PIXELS`.
+
+        This threshold can be changed by setting :py:data:`PIL.Image.MAX_IMAGE_PIXELS`. It can be disabled
+        by setting ``Image.MAX_IMAGE_PIXELS = None``.
+
+        If desired, the warning can be turned into an error with
+        ``warnings.simplefilter('error', Image.DecompressionBombWarning)`` or suppressed entirely with
+        ``warnings.simplefilter('ignore', Image.DecompressionBombWarning)``. See also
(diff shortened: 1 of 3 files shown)


docs/reference/Image.rst
```diff
@@ -374,6 +374,10 @@ Constants
 ---------
 
 .. data:: NONE
+.. data:: MAX_IMAGE_PIXELS
+
+    Set to 89,478,485, approximately 0.25gb for a 24 bit (3 bpp) image.
+    See :py:meth:`~PIL.Image.open` for more information about how this is used.
```
hugovk, line 380: It's not actually used directly by `Image.open`, but by `_decompression_bomb_check` called by `_open_core` by `open`, so not so easy to find.

Let's combine the info from the [4.2.0](https://pillow.readthedocs.io/en/stable/releasenotes/4.2.0.html#new-decompressionbomb-warning) and [5.0.0](https://pillow.readthedocs.io/en/stable/releasenotes/5.0.0.html#decompression-bombs-now-raise-exceptions) release notes. 

Maybe something along these lines?

> Pillow emits warnings or errors for images that are unexpectedly large and may be a denial of service. 
>
> A ``DecompressionBombWarning`` is issued for images larger than the ``MAX_IMAGE_PIXELS`` threshold, and  a ``DecompressionBombError`` is raised for images that are twice as big.
>
> The default threshold can be changed or disabled by setting
``Image.MAX_IMAGE_PIXELS = None``.

hugovk: Ah, sorry, I misunderstood the first version!

Before, https://github.com/python-pillow/Pillow/commit/fa4d1493e26891b82f5f315c9afa683fdc8e0bd1 had:

```
See :py:meth:`~PIL.Image.open` for more information about how this is used.
```

... and was linking not to the `Image.open` _source_, but to [_documentation_](https://pillow--4802.org.readthedocs.build/en/4802/reference/Image.html#PIL.Image.open) which has this warning:

image

In this warning, shall we mention and link to the constant?

```diff
-is over a certain limit.
+is over a certain limit, [MAX_IMAGE_PIXELS](link to the constant).
```

And let's also keep the link from the constant to `Image.open` docs.

And perhaps we should combine most of this info in a single place?

hugovk:
Also there's a little typo on this page that could be fixed at the same time:

```diff
     :param fp: A filename (string), pathlib.Path object or a file object.
        The file object must implement ``file.read``,
-       ``file.seek`, and ``file.tell`` methods,
+       ``file.seek``, and ``file.tell`` methods,
        and be opened in binary mode.
```

docs/reference/Image.rst
radarhere: Ok, I'm now linking to the constant. I'm confused though about how I can keep the link, while also trying to combine this info in one place?

radarhere:
I've added a commit to fix the typo.

docs/reference/Image.rst
hugovk: Yes, like this is good 👍

Shall we also mention `DecompressionBombError` in the first bit?

nulano:
The current revision is missing the information that user code can change this value. Looks good otherwise.

radarhere:
Ok, I've re-added that text, now in ``open()``.

hugovk:
Thank you!

PR #4803, opened 2020-07-20
Fail tests if bytes/bytearray are interpolated into str

lukegb:
This uses the [Python -b flag](https://docs.python.org/3/using/cmdline.html#cmdoption-b) (twice) to enable raising an exception if bytes/bytearray are compared with str.

 * Fail tests if bytes/bytearray are interpolated into str.
 * Fix case of TiffImagePlugin interpolating bytes into str without conversion.


lukegb:
Hmm, it's not clear to me how this managed to decrease the coverage, unless it's on a per-character basis.

nulano:
The coverage decrease is unrelated to this PR.

Looking at Codecov, there is a decrease in `test_image_quantize` code checking for `libimagequant` availability: Since imagequant is always present for PR tests, the not-present code was not executed, unlike the master build where it was hit on the Windows runners. There are two possible solutions to this that I can think of:
* Enable `libimagequant` on push builds -- this would include it in built wheels and make them undistributable due to GPL, a different trigger (e.g. manual or create label) could be used to build wheels without it.
* Add another test configuration which always runs without `libimagequant`. What I like about this idea, is that adding a configuration with a minimal set of dependencies would check that all tests are being skipped properly when a dependency of the test is missing.

There is also another line in `Dib.c` (line 246) that was hit on MINGW on master but not on this PR; I have no idea why.

hugovk:
Thanks!

Issue #3961, opened 2019-07-09
OSS-Fuzz integration


guidovranken:
I've written a fuzzer for Pillow. It is compatible with [Google OSS-Fuzz](https://github.com/google/oss-fuzz).

1. Are you interested in integrating the Pillow fuzzer into OSS-Fuzz? I will need one or more e-mail addresses, linked to a Google account, of people who ordinarily deal with security reports.
2. My fuzzer has so far found 2 distinct denial-of-service bugs that can be triggered by loading crafted data into ```ImageFile.Parser()```. Should I report these in the issue tracker, or via Tidelift?

Also pinging @kcc @Dor1s in case they want to object against Pillow integration into OSS-Fuzz.

Dor1s:
SGTM

wiredfool:
Works for me, I’d be one getting security issues.  

I’ve got a google account, but don’t receive email there. I’d prefer notifications go to the address on my GitHub profile.

guidovranken:
We can add two e-mail addresses for you: the one on your Github profile at which you receive e-mail notifications, and the one linked to your Google account, which you can use to login to the OSS-Fuzz dashboard at https://oss-fuzz.com/

I'm having a little trouble compiling Pillow with custom CFLAGS, which will be necessary for OSS-Fuzz integration. Would you mind commenting on how I might get the following to work?

```sh
CC=clang CXX=clang++ CFLAGS="-fsanitize=fuzzer-no-link" CXXFLAGS="-fsanitize=fuzzer-no-link" LDSHARED=clang ./setup.py build_ext --inplace
```

wiredfool:
This does build with the flags OMM (mojave), but linking fails because of python's symbols being missing.  (there's a little futzing to make zlib show up. )

``` CC=clang CXX=clang++ CFLAGS="-I/usr/local/opt/zlib/include -fsanitize=fuzzer-no-link" CXXFLAGS="-fsanitize=fuzzer-no-link" LDFLAGS="-L/usr/local/opt/zlib/lib" LDSHARED=clang python ./setup.py build_ext --inplace
running build_ext
building 'PIL._imaging' extension
creating build/temp.macosx-10.13-x86_64-2.7
creating build/temp.macosx-10.13-x86_64-2.7/src
creating build/temp.macosx-10.13-x86_64-2.7/src/libImaging
Building using 4 processes
[...]

PR #4806, opened 2020-07-21
Initialize offset memory for PyImagingPhotoPut

nqbit:
When using image mode "1" or "L" block.offset[3] is never initialized. In some use cases the block is passed to Tk_PhotoPutBlock (https://github.com/tcltk/tk/blob/master/generic/tkImgPhoto.c#L2752) and block.offset[3] is stored in alphaoffset which is later referenced.

This change initializes block.offset[3] to 0 when using mode "1" or "L".

### DIFF
--- a/src/Tk/tkImaging.c
+++ b/src/Tk/tkImaging.c
@@ -114,7 +114,7 @@ PyImagingPhotoPut(ClientData clientdata, Tcl_Interp* interp,
 
     if (strcmp(im->mode, "1") == 0 || strcmp(im->mode, "L") == 0) {
         block.pixelSize = 1;
-        block.offset[0] = block.offset[1] = block.offset[2] = 0;
+        block.offset[0] = block.offset[1] = block.offset[2] = block.offset[3] = 0;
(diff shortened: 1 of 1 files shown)


radarhere:
Are you aware of a situation that shows this bug? We would be interested in adding such a scenario as a test.

nqbit:
This was caught with a memory sanitizer, see the following example:

```python
import numpy as np
example_image_array = np.full((100,100), 123, dtype=np.uint8)
example_image = Image.fromarray(example_image_array)
example_image_tk = ImageTk.PhotoImage(image=example_image)
```

Is there currently any support for memory sanitization in the test framework?

radarhere:
The closest we have to memory sanitization testing is #3961.

As such, while I agree that your change makes logical sense, I'm having trouble running a sanitizer over the code, since Tkinter requires a display, and my usual approach to this problem would be display-less docker.

nqbit:
Thank you for helping to get this merged. One option for display-less configuration is xvfb (X virtual frame-buffer). https://en.wikipedia.org/wiki/Xvfb

PR #4812, opened 2020-07-23
Fix build on macOS and clang when zip.h is present on system 

PeterPetrik:
zip.h is part of zlib and under some circumstances it could break a build of pillow . renaming the Zip.h to never happen this


radarhere:
Hi. Could you elaborate a bit on 'under some circumstances'?

PeterPetrik:
well, I package QGIS and build PIL as one of the requirements. I have a zlib includes in the `$STAGE_PATH` folder and `export CFLAGS="-I$STAGE_PATH/include"`, and PIL compilation didn't go well since it tried to pick zip.h file instead of Zip.h file (looks like clangs on Mac happily ignored the case-sensitivity)

radarhere:
Might I suggest ZipCodecs.h instead? ZIPSTATE isn't the only thing in there.

hugovk (approved):

hugovk:
(Ah, I see nothing actually changed from the rebase :) Will go and approve the other PR.)

hugovk:
Thanks both!

PR #4813, opened 2020-07-24
Fix IFDRational and Exif equality for NaN values

luphord:
Fixes equality of `TiffImagePlugin.IFDRational` instances and therefore Exif data.

In Pillow 7.2.0 Exif RATIONALs and SIGNED_RATIONALs are read as `TiffImagePlugin.IFDRational`, instead of as a tuple with a numerator and a denominator. This caused a regression in one of my projects where I compared Exif data for equality. Reading exactly the same Exif data twice and comparing them via `==` yields `False` as the `TiffImagePlugin.IFDRational.__eq__` is broken.

`TiffImagePlugin.IFDRational` redundantly stores `numerator` / `denominator` and there ratio as `_val`. Equality comparison against other `TiffImagePlugin.IFDRational` instances worked only if `numerator` and `denominator` had no common divisors. E.g. `TiffImagePlugin.IFDRational(15, 5)` would not compare equal to itself: 15/5 == 3/1, but 15 != 3 and 5 != 1.

 * Fix `TiffImagePlugin.IFDRational.__eq__` to compare stored `Fraction` instance against other `Fraction` instance (instead of other `TiffImagePlugin.IFDRational` instance)
* Treat `NaN`-case (i.e. denominator==0) as a special case. Usually `NaN != NaN`, but that would break Exif equality. Therefore, 
`NaN == NaN` if numerators are equal.
* Test cases for `TiffImagePlugin.IFDRational` and Exif equality.

### DIFF
--- a/Tests/test_tiff_ifdrational.py
+++ b/Tests/test_tiff_ifdrational.py
@@ -49,6 +49,11 @@ def test_nonetype():
     assert xres and yres
 
 
+def test_nan():
+    # usually NaN != NaN, but this would break exif self-equality
+    assert IFDRational(123, 0) == IFDRational(123, 0)
+
+
 def test_ifd_rational_save(tmp_path):
     methods = (True, False)
     if not features.check("libtiff"):
--- a/src/PIL/TiffImagePlugin.py
+++ b/src/PIL/TiffImagePlugin.py
@@ -354,6 +354,9 @@ class IFDRational(Rational):
 
     def __eq__(self, other):
         if isinstance(other, IFDRational):
+            if self.denominator == 0 and other.denominator == 0:
+                # in this case self._val and other._val would be NaN
+                return self.numerator == other.numerator
             other = other._val
         return self._val == other
 


Tests/test_file_jpeg.py
```diff
@@ -241,6 +241,15 @@ def test_exif_gps(self):
         # Assert
         assert exif[gps_index] == expected_exif_gps
 
+    def test_exif_equality(self):
+        # in 7.2.0 Exif rationals are read as TiffImagePlugin.IFDRational
+        # which broke self-equality of Exif data
```
radarhere, line 246:
```suggestion
        # In 7.2.0, Exif rationals were changed to be read as
        # TiffImagePlugin.IFDRational. This class had a bug in __eq__,
        # breaking the self-equality of Exif data
```
Suggestion.

Tests/test_tiff_ifdrational.py
```diff
@@ -43,6 +50,12 @@ def test_nonetype():
     assert xres and yres
 
 
+def test_nan():
+    # usually NaN != NaN, but this would break exif self-equality
+
+    assert IFDRational(123, 0) == IFDRational(123, 0)
```
radarhere, line 56: You're not actually loading EXIF values of NaN, right? This is just for theoretical purposes?

luphord: Actually, it's a very practical issue: I am loading Exif from a [test image](https://raw.githubusercontent.com/luphord/imgwrench/master/tests/images/town.jpg). `exif[34853][6]` (GPS tag altitude) happens to be `0/0`, i.e. NaN. I do not care about this particular value, but I do care about `exif == exif_loaded_again`.

I use this equality assertion in test cases to ensure that loaded Exif data was saved to a transformed image unchanged. Alternatively, I could compare the Exif raw byte sequences. But then again: If the byte sequences are equal, why should the deserialized objects not compare equal?

radarhere: I'm not sold on the idea that NaN should equal NaN. If the GPS altitude is NaN, then one GPS altitude is not equal to the other, so maybe the EXIF data shouldn't be considered equal? I hear what you're saying, I'm just uncomfortable with the idea of causing our subclass of `Rational` to violate a standard like that.

If anyone else has input, feel free to jump in.

luphord: I see your point and understand that you do not want to deviate from standard behaviour. And maybe this is and edge case with little relevance. Does anyone have experience with NaNs in EXIF data?

Apart from this I wonder if the standard NaN != NaN behaviours is really what you want here:
- Up to version 7.2.0 loading EXIF from the *same image* resulted in equal objects, now it does not anymore.
- Equal EXIF byte streams do not necessarily imply equal objects anymore.

Regardless, I am open to removing the NaN-check from this PR. I believe that fixing `numerator` / `denominator` equality adds value on its own. We would then have equality if there are no NaNs present (which is a reasonably simple statement in my opinion).

radarhere: I've created #4888 to try and get the changes in that are agreed on, at least.

hugovk:
Thank you for the PR!

I think we should keep the usual NaN != NaN behaviour.

#4888 has been merged, does that still help your use case? Okay to close this?

luphord:
#4888 has been merged, so this PR contains only NaN == NaN as additional change. As NaN == NaN is not desired, this PR can now be closed. Thanks for the review and feedback!

PR #4814, opened 2020-07-25
Replaced distutils in test_image_access.py

radarhere:
For #4796, similar to #4809

Replaces
```python
from distutils import sysconfig
```
with
```python
import sysconfig
```
and
```python
sysconfig.get_python_inc()
```
with
```python
sysconfig.get_config_var("INCLUDEPY")
```
Also removes a `print` statement.

However, this test is skipped on CIs, so this PR passing is not good evidence that my changes are correct. Instead, I created https://github.com/radarhere/Pillow/commit/73f8d40df22f1203c01f9103c978835234d9c36d to check that the values with and without distutils are the same - passing on [AppVeyor](https://ci.appveyor.com/project/radarhere/pillow/builds/34296603) and [MSYS2 MINGW32](https://github.com/radarhere/Pillow/runs/909551283).

### DIFF
--- a/Tests/test_image_access.py
+++ b/Tests/test_image_access.py
@@ -2,7 +2,8 @@ import ctypes
 import os
 import subprocess
 import sys
-from distutils import ccompiler, sysconfig
+import sysconfig
+from distutils import ccompiler
 
 import pytest
 from PIL import Image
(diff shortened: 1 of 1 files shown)


radarhere:
In a move that's up for debate, I've added another commit, so that distutils is no longer explicitly imported - instead, I'm importing the distutils function from where it is imported in setuptools.

hugovk:
Thank you!

PR #4823, opened 2020-07-28
Do not try to close fp if fp is empty

### LINKED ISSUE #4821: EpsImagePlugin logs "Error closing: 'NoneType' object has no attribute 'close'"
### What did you do?
```
import io, logging, PIL.Image
img = PIL.Image.open(io.BytesIO(b"%!PS-Adobe-3.0 EPSF-3.0\n%%BoundingBox:   36   36  576  756\n"))
img.load()
```

(GhostScript must be in PATH)

Nothing.

### What actually happened?

`DEBUG:PIL.Image:Error closing: 'NoneType' object has no attribute 'close'`

* OS: Windows 10
* Python: Python 3.6.8 (tags/v3.6.8:3c6b436a57, Dec 24 2018, 00:16:47) [MSC v.1916 64 bit (AMD64)] on win32
* Pillow: 7.2.0

### What causes the problem?
Here: https://github.com/python-pillow/Pillow/blob/893489c57227f76cff8816de167ad329e5862756/src/PIL/EpsImagePlugin.py#L155 `Ghostscript()` calls `out_im.close()` but `out_im.load()` was already called here: https://github.com/python-pillow/Pillow/blob/893489c57227f76cff8816de167ad329e5862756/src/PIL/EpsImagePlugin.py#L145, so `Image.close()` encounters `self.fp == None`

radarhere:
Resolves #4821 

The issue reports that a debug message is logged if trying to `im.close()` after `im.load()`. This is because `ImageFile.load()` has https://github.com/python-pillow/Pillow/blob/893489c57227f76cff8816de167ad329e5862756/src/PIL/ImageFile.py#L278

And `im.close()` has https://github.com/python-pillow/Pillow/blob/893489c57227f76cff8816de167ad329e5862756/src/PIL/Image.py#L594-L600

So instead, this PR changes it to only close `fp` if `fp` is not empty, like `im.__exit__` already does.

https://github.com/python-pillow/Pillow/blob/893489c57227f76cff8816de167ad329e5862756/src/PIL/Image.py#L574-L580


urapadmin:
Not sure where to report this, I just add it here:
Image.close() seems to have the same issue like __exit__ and needs the same fix.

radarhere:
You say that `close()` needs this fix. This PR is indeed about applying that fix to `close()`. It isn't yet part of a Pillow release, but it is in the master branch for the next release.
I don't know what issue you're describing in `__exit__`.
Please provide some code to demonstrate.

urapadmin:
my apologies for the confusion. I just dealt with the bug in a workaround and misread the explanation above. All is fine and I am happy about the upcoming fix.

PR #4824, opened 2020-07-29
Remove outdated make upload-test and upload commands

hugovk:
The linked docs have moved.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -90,16 +90,5 @@ sdist:
 test:
 	pytest -qq
 
-# https://docs.python.org/3/distutils/packageindex.html#the-pypirc-file
-upload-test:
-#       [test]
-#       username:
-#       password:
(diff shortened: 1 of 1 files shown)


radarhere:
Using the Wayback Machine as reference - https://web.archive.org/web/20190402091726/https://docs.python.org/3/distutils/packageindex.html#the-pypirc-file - and Googling, I would have thought the link should be https://python.readthedocs.io/en/stable/distutils/packageindex.html#the-pypirc-file.

hugovk:
Actually, we're not using `python setup.py ... upload` for releases, we use twine in `RELEASING.md` as per current best practices.

Shall we instead remove `upload-test` and `upload` from the `Makefile`?

hugovk:
Updated to remove them.

PR #4831, opened 2020-08-01
Updated phony targets

radarhere:


hugovk:
On another project, instead of listing all the phonies at the top, we moved each one besides its target. That way they don't become out of sync when adding/copying/pasting/deleting.

For example, from the [docs](https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html):

```makefile 
.PHONY: clean
clean:
        rm *.o temp
```

Shall we do the same?

hugovk:
Thanks!

PR #4832, opened 2020-08-01
Removed 32-bit MinGW job

radarhere:
Removes 32-bit MinGW in light of https://www.msys2.org/news/#2020-05-17-32-bit-msys2-no-longer-actively-supported

### DIFF
--- a/.github/workflows/test-windows.yml
+++ b/.github/workflows/test-windows.yml
@@ -171,7 +171,7 @@ jobs:
       CHERE_INVOKING: 1
 
     timeout-minutes: 30
-    name: MSYS2 MINGW64
+    name: MSYS2 MinGW 64-bit
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/test-windows.yml
```diff
@@ -163,25 +163,15 @@ jobs:
...
       run:
         shell: bash.exe --login -eo pipefail "{0}"
     env:
-      MSYSTEM: ${{ matrix.mingw }}
+      MSYSTEM: MINGW64
       CHERE_INVOKING: 1
 
     timeout-minutes: 30
-    name: MSYS2 ${{ matrix.mingw }}
+    name: MSYS2 MINGW64
```
nulano, line 174:
```suggestion
    name: MSYS2 MinGW 64-bit
```

Can use proper name now that it is no longer used in the env block above.

.github/workflows/test-windows.yml
```diff
@@ -231,4 +221,4 @@ jobs:
           python3 -m pip install codecov
           bash <(curl -s https://codecov.io/bash) -F GHA_Windows
         env:
-          CODECOV_NAME: MSYS2 ${{ matrix.mingw }}
+          CODECOV_NAME: MSYS2 MINGW64
```
nulano, line 224:
```suggestion
          CODECOV_NAME: MSYS2 MinGW 64-bit
```

PR #4833, opened 2020-08-02
Added formats parameter to init

### LINKED ISSUE #4826: Load a restricted subset of image formats
Image parsing libraries are, and always have been, complicated pieces of software, with a requirement to (a) handle complicated input formats with a vast number of edge cases, and (b) operate as quickly and efficiently as possible.

This is a combination that leads to a lot of security bugs.

Pillow is no exception to this rule.

In many cases, when building an application, it makes sense to reduce the possible set of security bugs by only handling the input formats that you actually need to handle.

For example, if I only need to handle JPEG files, there is no reason for me to want to run a PostScript interpreter.  If I can design my program so it is *impossible* for the PostScript interpreter to be invoked, no matter what garbage input my program receives, then there's a huge category of security issues I don't need to worry about.

So, if I know what format the input file uses, is there a supported way to load it using *only* the required plugin?

There doesn't appear to be any documented way of doing this with Pillow.  As far as I can see, the only documented way of reading an image file is to use PIL.Image.open.  And if the input is malformed, then PIL.Image.open will potentially invoke *any* of the available format plugins, and will do so in an unpredictable order (depending on whether PIL.Image.init() or PIL.Image.preinit() is invoked first.)

An *undocumented* way would be to do this:
```python
PIL.Image.init()
image = PIL.Image.OPEN['JPEG'][0](fp, filename)
```
But this is pretty cryptic and I wouldn't assume this will always work with future Pillow versions.  This also skips the _decompression_bomb_check that is performed by PIL.Image.open.

radarhere:
Resolves #4826

Throwing out this suggestion. The issue talks about a way to limit the formats considered when opening an image.

At the moment, I think the best way is
```python
from PIL import Image
Image.init()
Image.ID = ['JPEG']
```

This PR suggests adding a `formats` argument to `init()` -

```python
from PIL import Image
Image.init(formats=["JPEG"])
```

With a `True` value to reset to all formats after having limited it.

```python
from PIL import Image
Image.init(formats=True)
```

### DIFF
--- a/Tests/test_image.py
+++ b/Tests/test_image.py
@@ -93,21 +93,24 @@ class TestImage:
             assert len(Image.ID) == 0
             for ext in (".jpg", ".png"):
                 with pytest.raises(UnidentifiedImageError):
-                    Image.open("Tests/images/hopper" + ext)
+                    with Image.open("Tests/images/hopper" + ext):
+                        pass
 
             # Only JPEG
             Image.init(["JPEG"])
             assert len(Image.ID) != 0
             Image.open("Tests/images/hopper.jpg")
             with pytest.raises(UnidentifiedImageError):
-                Image.open("Tests/images/hopper.png")
+                with Image.open("Tests/images/hopper.png"):
+                    pass
 
         finally:
             # All formats
             Image.init(True)
             assert len(Image.ID) != 0
             for ext in (".jpg", ".png"):
-                Image.open("Tests/images/hopper" + ext)
+                with Image.open("Tests/images/hopper" + ext):
+                    pass
 
     def test_width_height(self):
         im = Image.new("RGB", (1, 2))
--- a/Tests/test_image.py
+++ b/Tests/test_image.py
@@ -86,6 +86,29 @@ class TestImage:
         # with pytest.raises(MemoryError):
         #   Image.new("L", (1000000, 1000000))
 
+    def test_init(self):
+        try:
+            # No formats
+            Image.init([])
+            assert len(Image.ID) == 0
+            for ext in (".jpg", ".png"):
+                with pytest.raises(UnidentifiedImageError):
+                    Image.open("Tests/images/hopper" + ext)
+
+            # Only JPEG
+            Image.init(["JPEG"])
+            assert len(Image.ID) != 0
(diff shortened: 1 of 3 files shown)


Tests/test_image.py
```diff
@@ -86,6 +86,29 @@ def test_sanity(self):
...
         #   Image.new("L", (1000000, 1000000))
 
+    def test_init(self):
+        try:
+            # No formats
+            Image.init([])
+            assert len(Image.ID) == 0
+            for ext in (".jpg", ".png"):
+                with pytest.raises(UnidentifiedImageError):
+                    Image.open("Tests/images/hopper" + ext)
```
nulano, line 96:
```suggestion
                    with Image.open("Tests/images/hopper" + ext):
                        pass
```

Apply #4232 to this test. Another option is to use e.g. `assert im.mode == "RGB"` instead of `pass`.

Tests/test_image.py
```diff
@@ -86,6 +86,29 @@ def test_sanity(self):
...
+
+            # Only JPEG
+            Image.init(["JPEG"])
+            assert len(Image.ID) != 0
+            Image.open("Tests/images/hopper.jpg")
+            with pytest.raises(UnidentifiedImageError):
+                Image.open("Tests/images/hopper.png")
```
nulano, line 103:
```suggestion
                with Image.open("Tests/images/hopper.png"):
                    pass
```

Tests/test_image.py
```diff
@@ -86,6 +86,29 @@ def test_sanity(self):
...
+
+        finally:
+            # All formats
+            Image.init(True)
+            assert len(Image.ID) != 0
+            for ext in (".jpg", ".png"):
+                Image.open("Tests/images/hopper" + ext)
```
nulano, line 110:
```suggestion
                with Image.open("Tests/images/hopper" + ext):
                    pass
```

nulano:
The `ID` value is also used in `features.pilinfo. Perhaps that reference should also be changed to `EVERY_ID`.

radarhere:
Good thought. I've added a commit for that.

radarhere:
Closing in favour of #4837

PR #4836, opened 2020-08-02
Add develop instructions for winbuild

nulano:
Alternative to #4680, documents how to install Pillow in develop mode using the winbuild scripts.

### DIFF
--- a/winbuild/build.rst
+++ b/winbuild/build.rst
@@ -81,6 +81,9 @@ Pillow for the selected version of Python.
 ``winbuild\build\build_pillow.cmd bdist_wheel`` will build wheels
 instead of installing Pillow.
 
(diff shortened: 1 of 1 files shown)


radarhere:
Bear with me - why does it need to be `inplace`?

nulano:
From https://github.com/python-pillow/Pillow/pull/4680#issuecomment-641539312:

> Edit: Trying this locally, I am getting an odd build error, likely due to the double invocation of build_ext. The correct command is winbuild\build\build_pillow.cmd --inplace develop.

`build_pillow.cmd` calls `build_ext`, while `develop` calls `build_ext --inplace`. When I tried it without `--inplace`, the second `build_ext` invocation failed, likely because `build_ext` was initialized without the `--inplace` parameter.

PR #4837, opened 2020-08-02
Add formats parameter to open

### LINKED ISSUE #4826: Load a restricted subset of image formats

nulano:
Resolves #4826.
Alternative to #4833.

Instead of restricting the set of available formats for all open operation, this PR allows to restrict the set on a per-file basis, e.g.:

```python
from PIL import Image
Image.open("hopper.png", formats=["PNG"])
```

This gives dependent libraries and programs finer control over which formats are used when.

A sample use case would be a web server which lets users upload avatars only in PNG, but lets admins upload any image file to embed in the website content, without any synchronisation concerns with the `Image.init` function.


hugovk (approved):
Please also add to release notes. Thanks!

PR #4838, opened 2020-08-03
Fix exception handling when saving images

liZe:
The e variable is already used in the for loop, use exc to store the exception.

### DIFF
--- a/src/PIL/ImageFile.py
+++ b/src/PIL/ImageFile.py
@@ -507,7 +507,7 @@ def _save(im, fp, tile, bufsize=0):
     try:
         fh = fp.fileno()
         fp.flush()
-    except (AttributeError, io.UnsupportedOperation) as e:
+    except (AttributeError, io.UnsupportedOperation) as exc:
         # compress to Python file-compatible object
         for e, b, o, a in tile:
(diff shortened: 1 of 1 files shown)


radarhere:
Thanks for this. Is there a test situation that you know of that triggers the bug, so that we can add it to the test suite and prevent this from re-appearing in the future?

liZe:
> Thanks for this. Is there a test situation that you know of that triggers the bug, so that we can add it to the test suite and prevent this from re-appearing in the future?

Sure!

```python
from io import BytesIO
from PIL import Image
Image.new('P', (1, 1)).save(BytesIO(), 'JPEG2000')
```

This should raise the OSError, not `TypeError: exception causes must derive from BaseException`.

radarhere:
@liZe you've deleted your Pillow repository. I'd like to add the test situation you've mentioned though, so I've recreated this PR as #4885. Let us know if you have anything further to add.

liZe:
> I'd like to add the test situation you've mentioned though, so I've recreated this PR as #4885. Let us know if you have anything further to add.

Sorry, I didn’t check, I thought it was already merged.

#4885 is OK for me!

Issue #3331, opened 2018-09-04
Autocontrast cutoff behaves unexpectedly


jnweiger:
### What did you do?
python-3.6 with Pillow-5.2.0
![leaf-gr](https://user-images.githubusercontent.com/1108546/45062236-50b78980-b0a8-11e8-8ffe-63d449e80e05.jpg)
Download this image as leaf-gr.jpg then start python and do
```python
from PIL import Image, ImageOps
im = Image.open('leaf-gr.jpg')
ImageOps.autocontrast(im, cutoff=2).show()
```
![grafik](https://user-images.githubusercontent.com/1108546/45062365-da675700-b0a8-11e8-8b13-4bda846d4edd.png)

### What did you expect to happen?
The background is two light gray values. I expect both to become brighter.
The foreground is darker gray values, I expect these to touch black.

Histogram stretching until the outermost gray values hit 0 and 255 respectively.
Then some more linear stretching so that the outermost gray values would go 2 percent beyond, ca -5 and 260, then clip everything back to be within 0..255 

### What actually happened?
The background is half black 0x000000 and half white 0xffffff. That is unexpected. 
The darker gray of the background is in any case more than 2% percent brighter than the darkest point of the foreground. (Or we must be measuring very non-inear)

### What versions of Pillow and Python are you using?
python-3.6 with Pillow-5.2.0

radarhere:
> The background is two light gray values. I expect both to become brighter.
> The foreground is darker gray values, I expect these to touch black.

If you don't specify `cutoff=2`, then the result matches this part of the expectation.

https://pillow.readthedocs.io/en/stable/reference/ImageOps.html#PIL.ImageOps.autocontrast
> This function calculates a histogram of the input image (or mask region), removes `cutoff` percent of the lightest and darkest pixels from the histogram, and remaps the image so that the darkest pixel becomes black (0), and the lightest becomes white (255).

I think your expectation is that the cutoff would apply after scaling, whereas Pillow is applying it before scaling. This behaviour is what is described in the documentation, so I don't think there's anything to do here.

PR #4843, opened 2020-08-06
ENH: Autocontrast method enhancement - mask specification for computation

navneeth:
This is a draft of an enhancement to autocontrast that allows the specification of a mask to constrain the contrast stretching.  
`autocontrast(image, mask=a_mask)`

Perhaps it Fixes # #3331. 

 * adding an optional mask argument. if none is specified, behaviour defaults to how it was before. 
 * added tests in Tests/test_imageops.py
 * also added tests for cutoff argument, where the tuple case was not being checked. 

This is my first contribution to pillow. So do let me know if this suffices.

### DIFF
--- a/Tests/test_imageops.py
+++ b/Tests/test_imageops.py
@@ -1,5 +1,5 @@
 import pytest
-from PIL import Image, ImageOps, features
+from PIL import Image, ImageOps, ImageDraw, features
 
 from .helper import (
     assert_image_equal,
@@ -24,7 +24,9 @@ def test_sanity():
     ImageOps.autocontrast(hopper("RGB"))
 
     ImageOps.autocontrast(hopper("L"), cutoff=10)
+    ImageOps.autocontrast(hopper("L"), cutoff=(2, 10))
     ImageOps.autocontrast(hopper("L"), ignore=[0, 255])
+    ImageOps.autocontrast(hopper("L"), mask=hopper("L"))
 
     ImageOps.colorize(hopper("L"), (0, 0, 0), (255, 255, 255))
     ImageOps.colorize(hopper("L"), "black", "white")
@@ -311,3 +313,30 @@ def test_autocontrast_cutoff():
 
         assert autocontrast(10) == autocontrast((10, 10))
         assert autocontrast(10) != autocontrast((1, 10))
+
+
+def test_autocontrast_mask():
+    # Test the mask argument of autocontrast
+    with Image.open("Tests/images/bw_gradient.png") as img:
+
+        rect_mask = Image.new("L", img.size, 0)
+        draw = ImageDraw.Draw(rect_mask)
+        x0, y0 = img.size[0]//4, img.size[1]//4
+        x1, y1 = 3*img.size[0]//4, 3*img.size[1]//4
+        draw.rectangle((x0, y0, x1, y1), fill=255)
+
+        assert ImageOps.autocontrast(img, mask=rect_mask) != ImageOps.autocontrast(img)
+
+    # Test the autocontrast with a rectangular mask
+    with Image.open("Tests/images/iptc.jpg") as img:
+
+        rect_mask = Image.new("L", img.size, 0)
+        draw = ImageDraw.Draw(rect_mask)
+        x0, y0 = img.size[0]//2, img.size[1]//2
+        x1, y1 = img.size[0]-40, img.size[1]
+        draw.rectangle((x0, y0, x1, y1), fill=255)
+
+        result = ImageOps.autocontrast(img, mask=rect_mask)
+        result_nomask = ImageOps.autocontrast(img)
+
+        assert result_nomask != result
--- a/src/PIL/ImageOps.py
+++ b/src/PIL/ImageOps.py
@@ -61,10 +61,10 @@ def _lut(image, lut):
 # actions
 
 
-def autocontrast(image, cutoff=0, ignore=None):
+def autocontrast(image, cutoff=0, ignore=None, mask=None):
     """
     Maximize (normalize) image contrast. This function calculates a
-    histogram of the input image, removes **cutoff** percent of the
+    histogram of the input image (or mask region), removes **cutoff** percent of the
     lightest and darkest pixels from the histogram, and remaps the image
     so that the darkest pixel becomes black (0), and the lightest
     becomes white (255).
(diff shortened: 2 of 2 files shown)


src/PIL/ImageOps.py
```diff
@@ -74,9 +74,15 @@ def autocontrast(image, cutoff=0, ignore=None):
...
+    :param mask: histogram used in contrast operation is computed using pixels
+                 within the mask. If no mask is given the entire image is used
+                 for histogram computation.
     :return: An image.
     """
-    histogram = image.histogram()
+    if mask:
+        histogram = image.histogram(mask)
+    else:
+        histogram = image.histogram()
```
hugovk, line 85: Just pass the `mask` in? It defaults to `None`:

`    def histogram(self, mask=None, extrema=None):`

```suggestion
    histogram = image.histogram(mask)
```

src/PIL/ImageOps.py
```diff
@@ -74,9 +74,15 @@ def autocontrast(image, cutoff=0, ignore=None):
                    high ends. Either a tuple of (low, high), or a single
                    number for both.
     :param ignore: The background pixel value (use None for no background).
+    :param mask: histogram used in contrast operation is computed using pixels
```
hugovk, line 77:
```suggestion
    :param mask: Histogram used in contrast operation is computed using pixels
```

Tests/test_imageops.py
```diff
@@ -311,3 +313,30 @@ def autocontrast(cutoff):
...
+
+        rect_mask = Image.new("L", img.size, 0)
+        draw = ImageDraw.Draw(rect_mask)
+        x0, y0 = img.size[0]//4, img.size[1]//4
+        x1, y1 = 3*img.size[0]//4, 3*img.size[1]//4
+        draw.rectangle((x0, y0, x1, y1), fill=255)
+
+        assert ImageOps.autocontrast(img, mask=rect_mask) != ImageOps.autocontrast(img)
+
+    # Test the autocontrast with a rectangular mask
```
hugovk, line 330: Please could you split this into a second test `def`? That way we know this part doesn't depend on something happening on the previous part.

Tests/test_imageops.py
```diff
@@ -311,3 +313,30 @@ def autocontrast(cutoff):
...
+    # Test the mask argument of autocontrast
+    with Image.open("Tests/images/bw_gradient.png") as img:
+
+        rect_mask = Image.new("L", img.size, 0)
+        draw = ImageDraw.Draw(rect_mask)
+        x0, y0 = img.size[0]//4, img.size[1]//4
+        x1, y1 = 3*img.size[0]//4, 3*img.size[1]//4
+        draw.rectangle((x0, y0, x1, y1), fill=255)
+
+        assert ImageOps.autocontrast(img, mask=rect_mask) != ImageOps.autocontrast(img)
```
hugovk, line 328: This test asserts that an autocontrasted image with a mask is different an autocontrasted image without a mask. Okay.

But for all we know the masked image could be a blank black image. Could you add something to actually test for correctness of the masked output?

Tests/test_imageops.py
```diff
@@ -311,3 +313,30 @@ def autocontrast(cutoff):
...
+        rect_mask = Image.new("L", img.size, 0)
+        draw = ImageDraw.Draw(rect_mask)
+        x0, y0 = img.size[0]//2, img.size[1]//2
+        x1, y1 = img.size[0]-40, img.size[1]
+        draw.rectangle((x0, y0, x1, y1), fill=255)
+
+        result = ImageOps.autocontrast(img, mask=rect_mask)
+        result_nomask = ImageOps.autocontrast(img)
+
+        assert result_nomask != result
```
hugovk, line 342: Same comment here about testing for correctness.

src/PIL/ImageOps.py
navneeth: Agree. good catch.

Tests/test_imageops.py
navneeth: Makes sense. I have split it in new PR.

Tests/test_imageops.py
navneeth: Hmm. That will require some deeper thought. What you are suggesting is a functionality
test for the autocontrast method. 
I noticed that the baseline autocontrast did not have such a functional test in place.

1. So I guess the test would apply to both ?
2. I would likely have to add a custom test image for such a test. Any guidelines on this ?

How about we close this PR and follow up with the functional test in another PR ?

navneeth:
Thank you for the thorough and sensible review. I genuinely appreciate what you do.
cheers.  
Just submitted an updated PR with the changes.

Tests/test_imageops.py
```diff
@@ -311,3 +313,32 @@ def autocontrast(cutoff):
...
+
+
+def test_autocontrast_mask_toy_input():
+    # Test the mask argument of autocontrast
+    with Image.open("Tests/images/bw_gradient.png") as img:
+
+        rect_mask = Image.new("L", img.size, 0)
+        draw = ImageDraw.Draw(rect_mask)
+        x0, y0 = img.size[0] // 4, img.size[1] // 4
+        x1, y1 = 3 * img.size[0] // 4, 3 * img.size[1] // 4
```
hugovk, line 325: Let's also put these on their own lines for clarity:

```suggestion
        x0 = img.size[0] // 4
        y0 = img.size[1] // 4
        x1 = 3 * img.size[0] // 4
        y1 = 3 * img.size[1] // 4
```

Tests/test_imageops.py
hugovk: We could test an arbitrary pixel is something expected. For example:

```diff
diff --git a/Tests/test_imageops.py b/Tests/test_imageops.py
index 3d5223d9..c5ad670a 100644
--- a/Tests/test_imageops.py
+++ b/Tests/test_imageops.py
@@ -342,3 +342,5 @@ def test_auto_contrast_mask_real_input():
         result_nomask = ImageOps.autocontrast(img)

         assert result_nomask != result
+        assert result.getpixel((20, 30)) == (239, 239, 242)
+        assert result_nomask.getpixel((20, 30)) == (132, 118, 96)
```

No need to close and create a new PR, we can update this one.

navneeth: Just added code to test for explicit pixel value in the test. Rather than look for a pixel value it compares the stats (median) of the autocontrast (with mask) and autocontrast (without mask). This logic draws from the theory of autocontrast and should scale well. PR 01aeaa4

navneeth:
Update the PR with code to test for explicit pixel value in the test. Rather than look for a pixel value it compares the stats (median) of the autocontrast (with mask) and autocontrast (without mask). This logic draws from the theory of autocontrast and should scale well. Please see PR 2d3a841

hugovk:
I've updated this PR with release notes:

* https://pillow--4843.org.readthedocs.build/en/4843/releasenotes/8.0.0.html#api-additions

PR #4845, opened 2020-08-09
Add release notes for 8.0.0 and a template

hugovk:

 * Add release notes for 8.0.0
 * Add a release note template for future copy/paste, with the sections in the usual order
 * Exclude the template from `make doc`, to silence a `make doc` error as it's not linked in the toctree
 * Put `PIL.ImageChops.offset` as `autodoc`, as it has a good docstring
 * A few other minor doc updates related to autolinking to classes, methods or functions

### DIFF
--- a/docs/releasenotes/8.0.0.rst
+++ b/docs/releasenotes/8.0.0.rst
@@ -54,7 +54,7 @@ ImageOps.autocontrast cutoffs
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 Previously, the ``cutoff`` parameter of :py:func:`.ImageOps.autocontrast` could only
-be a single integer, used as the percent to cut off from the histogram on the low and
+be a single number, used as the percent to cut off from the histogram on the low and
 high ends.
 
 Now, it can also be a tuple ``(low, high)``.
--- a/docs/releasenotes/8.0.0.rst
+++ b/docs/releasenotes/8.0.0.rst
@@ -1,0 +1,73 @@
+8.0.0
+-----
+
+Backwards Incompatible Changes
+==============================
+
+Python 3.5
+^^^^^^^^^^
+
+Pillow has dropped support for Python 3.5, which reached end-of-life on 2020-09-13.
+
+im.offset
+^^^^^^^^^
+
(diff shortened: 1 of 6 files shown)


docs/releasenotes/8.0.0.rst
```diff
@@ -0,0 +1,73 @@
...
+:py:class:`.PsdImagePlugin.PsdImageFile` MIME type.
+
+API Additions
+=============
+
+ImageOps.autocontrast cutoffs
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Previously, the ``cutoff`` parameter of :py:func:`.ImageOps.autocontrast` could only
+be a single integer, used as the percent to cut off from the histogram on the low and
```
radarhere, line 57:
```suggestion
be a single number, used as the percent to cut off from the histogram on the low and
```
It could also be a float?

hugovk: Indeed it can, thanks!

WhyNotHugo:
There's a breaking change that I believe is a side effect of dropping `tostring()`, that should probably be documented.

I've seen and used this pattern in many applications:

```python
from io import BytesIO

from PIL import Image

image = Image.open("someimage.jpeg")
# do something to image
io = BytesIO()
io.write(image)
```

This used to work up to 7.2.0, but doesn't work in `master` any more. It's probably worth including a mention and workaround in the release notes.

I've also seen other file-like objects used, like Django's [HTTPResponse](https://docs.djangoproject.com/en/3.1/ref/request-response/#django.http.HttpResponse).

hugovk:
That doesn't work for me in 7.2.0 either:

```pycon
Python 3.8.3 (v3.8.3:6f8c8320e9, May 13 2020, 16:29:34)
[Clang 6.0 (clang-600.0.57)] on darwin
>>> from io import BytesIO
>>> from PIL import Image
>>> Image.__version__
'7.2.0'
>>> image = Image.open("Tests/images/hopper.jpg")
>>> image.show()
>>> io = BytesIO()
>>> io.write(image)
TypeError: a bytes-like object is required, not 'JpegImageFile'
```

WhyNotHugo:
Never mind, my bad! I figured I had a check elsewhere for `hasattr(output, 'tostring')`. Sorry for the noise!

Issue #4818, opened 2020-07-27
Writing your own image plugin documentation missing information


altimore:
Hello,

First, thank you guys for the work you put in.
I hope this is the good place to ask.

Today i got to work on files in an exotic format generated by an old RIP software.

I checked the doc and it seems there is exactly what i need in the handbook chapter called ["Writing your own image plugin"](https://pillow.readthedocs.io/en/stable/handbook/writing-your-own-file-decoder.html).

However, the documentation is not explaining everything so I'm left with questions and i hope i could get some pointers here :

1. What should be the recommended way of loading the new plugin ? Should we clone the pillow repo and copy it in the PIL folder ? or should we call it in some way to add it in the register before calling `Image.open()` ?

2. the decoders documentation specify that we should set the data to the PIL's internal pixel layout when writing our own, I tried to check in the source code and it seems that the function `set_as_raw` from the `PyDecoder` class is the way to go but it doesn't describe the layout itself nor the way to fill it.

### What did you do? 
Search the documentation, and browse the source code to try to find the information.

### What did you expect to happen?
It would be nice if the page/example could provide a bit more details.

### What are your OS, Python and Pillow versions?

* OS: Arch Linux
* Python: 3.8.3
* Pillow: 7.2.0

hugovk:
I can quickly answer the first question:

> 1. What should be the recommended way of loading the new plugin ? Should we clone the pillow repo and copy it in the PIL folder ? or should we call it in some way to add it in the register before calling `Image.open()` ?

The second way is better. For example, with [`SpamImagePlugin.py`](https://pillow.readthedocs.io/en/stable/handbook/writing-your-own-file-decoder.html#example) in my directory, I can import it in my code, and it'll register the plugin and extension:

`1.py`:

```python
from PIL import Image
import SpamImagePlugin

with Image.open("1.spam") as im:
    print(im)
    print(im.format)
    print(im.format_description)
```
`1.spam`:
```spam
SPAM 1 1 1 1
```
```console
$ python 1.py
<SpamImagePlugin.SpamImageFile image mode=1 size=1x1 at 0x1023D5E50>
SPAM
Spam raster image
```

radarhere:
[...]

PR #4847, opened 2020-08-10
Improved example plugin documentation

radarhere:
- Combine two `register_extension` calls into `register_extensions`
- Added an example importing SpamImagePlugin, to help #4818

### DIFF
--- a/docs/handbook/writing-your-own-file-decoder.rst
+++ b/docs/handbook/writing-your-own-file-decoder.rst
@@ -159,6 +159,8 @@ can be used with most uncompressed file formats, such as PPM, BMP, uncompressed
 TIFF, and many others. To use the raw decoder with the
 :py:func:`PIL.Image.frombytes` function, use the following syntax::
 
+.. code-block:: python
+
     image = Image.frombytes(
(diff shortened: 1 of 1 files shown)


docs/handbook/writing-your-own-file-decoder.rst
```diff
@@ -103,6 +105,13 @@ Note that the image plugin must be explicitly registered using
 :py:func:`PIL.Image.register_open`. Although not required, it is also a good
 idea to register any extensions used by this format.
 
+Once the plugin has been imported, it can be used::
+
+    from PIL import Image
```
hugovk, line 110: Shall we add explicit Python code blocks in this file too?

```suggestion
Once the plugin has been imported, it can be used:

.. code-block:: python

    from PIL import Image
```

Looks like things are already autocoloured for Python at https://pillow.readthedocs.io/en/stable/handbook/writing-your-own-file-decoder.html but being explicit will also colour it here on GitHub: https://github.com/python-pillow/Pillow/blob/master/docs/handbook/writing-your-own-file-decoder.rst

PR #4850, opened 2020-08-11
Docs: Use code formatting instead of emphasis where applicable

nulano:
Followup to https://github.com/python-pillow/Pillow/pull/4774#discussion_r452840274

Replace instances of `**code**` with ``` ``code`` ```, and fix various code formatting related typos.
Document `ImageFile.LOAD_TRUNCATED_IMAGES`, `ImageFile.ERRORS`, `ImageFileDirectory_v1.tagtypes` and `ImageFont.LAYOUT_*` which are now linked to.

### DIFF
--- a/src/PIL/ImageFile.py
+++ b/src/PIL/ImageFile.py
@@ -40,6 +40,7 @@ MAXBLOCK = 65536
 SAFEBLOCK = 1024 * 1024
 
 LOAD_TRUNCATED_IMAGES = False
+"""Whether loading of truncated image files is currently enabled. User code may change this."""
 
 ERRORS = {
     -1: "image buffer overrun error",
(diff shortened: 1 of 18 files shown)


src/PIL/ImageFile.py
```diff
@@ -40,6 +40,7 @@
 SAFEBLOCK = 1024 * 1024
 
 LOAD_TRUNCATED_IMAGES = False
+"""Whether loading of truncated image files is currently enabled. User code may change this."""
```
radarhere, line 43:
```suggestion
"""Whether or not to load truncated image files. User code may change this."""
```
Fixes the lint error that the line is too long. You could also break your text into two lines, but here's my suggestion.

hugovk:
Thank you!

Issue #4857, opened 2020-08-12
_imaging.py doesn't have PILLOW_VERSION attribute after building with all dependencies with VS2019


newfeatureengineer:
### What did you do?
Built pillow from source using the script in winbuild. 

### What did you expect to happen?

I expected the package to work.

### What actually happened?

Package seems broken.

### What are your OS, Python and Pillow versions?

* OS: Windows 10
* Python: 3.7.7
* Pillow: 7.2.0

1. Extract pillow-7.2.0.tar.gz
2. cd to winbuild
3. python build_prepare.py
4. cd to build
5. build_dep_all  (had to modify `winbuild\build\lcms2-2.11\Projects\VC2017\lcms2_static\lcms2_static.vcxproj` to point to the installed windows SDK version (10.0.19041.0) for it to compile, but otherwise, everything worked fine).
6. build_pillow
7. back in the root of the extracted package, run `python setup.py install`
*   shows the [following output](https://pastebin.pl/view/ed23290b)
8. python selftest.py
Fails with 

```
RuntimeWarning: The _imaging extension was built for another version of Pillow or PIL:
Core version: None
Pillow version: 7.2.0
Traceback (most recent call last):
  File "selftest.py", line 6, in <module>
    from PIL import Image, features
  File "c:\venv\lib\site-packages\pillow-7.2.0-py3.7-win-amd64.egg\PIL\Image.py", line 100, in <module>
ImportError: The _imaging extension was built for another version of Pillow or PIL:
Core version: None
Pillow version: 7.2.0
```

Because `PIL\_imaging.py` loads from `_imaging.cp37-win_amd64.pyd` with no attribute named `PILLOW_VERSION`.

radarhere:
Hi. What version of setuptools are you using? My guess is that you are using setuptools > 47.3.1, which has a bug at the moment. If you downgrade to setuptools 47.3.1, hopefully your problem will be solved.

A PR to fix the issue in setuptools is currently open at https://github.com/pypa/setuptools/pull/2297

hugovk:
Snap :)

nulano:
Yes, this looks like the setuptools issue. GHA also fails with recent versions: https://github.com/nulano/Pillow/runs/866809418?check_suite_focus=true#step:21:803

I've created #4858 to fix the LCMS2 issue, thanks for the notice.

radarhere:
The setuptools PR has now been merged and released, so you could also upgrade setuptools to the 49.3.2

[...]

PR #4784, opened 2020-07-13
3.9-dev: Upgrade setuptools

hugovk:
The setuptools fix https://github.com/pypa/setuptools/pull/2249 for python-pillow/Pillow#4769  has been released in [setuptools v49.1.3](https://github.com/pypa/setuptools/releases/tag/v49.1.3), so we can require that version or newer.

Created as draft as there are further build problems from new changes in setuptools, will add comments.

PR #4858, opened 2020-08-12
Fix LCMS2 patch

nulano:
Fixes error from #4857:

> had to modify winbuild\build\lcms2-2.11\Projects\VC2017\lcms2_static\lcms2_static.vcxproj to point to the installed windows SDK version (10.0.19041.0) for it to compile, but otherwise, everything worked fine

Also make sure all patches are actually applied to avoid this happening again.


nulano:
GHA 3.9-dev jobs have now started failing due to #4784 with the release of 3.9-rc1.

hugovk:
thanks!

PR #4868, opened 2020-08-14
Removed duplicate package

radarhere:
mingw-w64-x86_64-python3-numpy is mentioned twice in test-windows.yml.

https://github.com/python-pillow/Pillow/blob/cb79b1fa8900adc1617e9518bb8a83d671dc9246/.github/workflows/test-windows.yml#L227-L229

### DIFF
--- a/.github/workflows/test-windows.yml
+++ b/.github/workflows/test-windows.yml
@@ -218,22 +218,22 @@ jobs:
       - name: Install Dependencies
         run: |
           pacman -S --noconfirm \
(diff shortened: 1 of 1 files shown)


hugovk:
Can we also alphabetise them? Would help avoid future duplicates.

nulano:
The non-Python packages are sorted according to the order in `features.pilinfo`, and IIRC the Python packages are sorted according to the order in which they were reported in test skips.

+1 for alphabetising the Python packages; no opinion on non-Python packages.

PR #4869, opened 2020-08-15
Moved CVE images into password-protected zip

### LINKED ISSUE #4730: Antivirus Avast detect  test  image crash_2.tif as with  exploit

radarhere:
Resolves #4730 

Using a suggestion from https://github.com/python-pillow/Pillow/pull/4792#issuecomment-659335356, this PR moves images into a password-protected zip, that is then read at the time of the test.


hugovk:
Replaced by https://github.com/python-pillow/Pillow/pull/4929, thanks for the PRs!

PR #4874, opened 2020-08-18
Add dominant color method

### LINKED ISSUE #4634: Identify dominant colors
This is a feature request. 

Would it be within the scope of pillow to have a function that identifies dominant colors within an image? The [relevant question](https://stackoverflow.com/questions/3241929/python-find-dominant-most-common-color-in-an-image#3244061) on StackOverflow has been viewed ~50k times. 

One of the suggestions is the [colorthief](https://github.com/fengsp/color-thief-py/) library that identifies an image's palette and dominant colors with only `Pillow` (and `math`, but it's not necessary). Could that code be ported to Pillow?

Pillow's own `getcolors()` doesn't solve the problem.

392781:
Fixes #4634.

 * Adds a method to determine the dominant colors in an image using k-means clustering
 
I added some very basic tests but seeing as I'm very new at open source contributing I am open to suggestions to improve them.

### DIFF
--- a/Tests/test_image_getdominantcolors.py
+++ b/Tests/test_image_getdominantcolors.py
@@ -2,13 +2,9 @@ from .helper import hopper
 
 
 def test_getdominantcolors():
-    def getdominantcolors(mode, numcolors=None):
+    def getdominantcolors(mode):
         im = hopper(mode)
-
-        if numcolors:
-            colors = im.getdominantcolors(numcolors)
-        else:
-            colors = im.getdominantcolors()
+        colors = im.getdominantcolors()
         return len(colors)
 
     assert getdominantcolors("F") == 3
--- a/Tests/test_image_getdominantcolors.py
+++ b/Tests/test_image_getdominantcolors.py
@@ -1,0 +1,22 @@
+from .helper import hopper
+
+
+def test_getdominantcolors():
+    def getdominantcolors(mode, numcolors=None):
+        im = hopper(mode)
+
+        if numcolors:
+            colors = im.getdominantcolors(numcolors)
+        else:
+            colors = im.getdominantcolors()
+        return len(colors)
+
+    assert getdominantcolors("F") == 3
+    assert getdominantcolors("I") == 3
+    assert getdominantcolors("L") == 3
+    assert getdominantcolors("P") == 3
+    assert getdominantcolors("RGB") == 3
+    assert getdominantcolors("YCbCr") == 3
+    assert getdominantcolors("CMYK") == 3
+    assert getdominantcolors("RGBA") == 3
+    assert getdominantcolors("HSV") == 3
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -1304,8 +1304,7 @@ class Image:
         elif self.mode in ("RGBA", "CMYK"):
             channels = 4
         else:
-            tb = sys.exc_info()[2]
-            raise ValueError("Unsupported image mode").with_traceback(tb)
+            raise ValueError("Unsupported image mode")
 
         def euclidean(p1, p2):
             if channels == 1:
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -1272,7 +1272,7 @@ class Image:
             return self.im.getband(band)
         return self.im  # could be abused
 
-    def getdominantcolors(self, numcolors=3, maxiter=50, threshold=1):
+    def getdominantcolors(self, numcolors=3, maxiter=50, threshold=1, quality=1.0):
         """
         Returns a list of dominant colors in an image using k-means
         clustering.
@@ -1284,10 +1284,20 @@ class Image:
         :param threshold: Early stopping condition for the algorithm.
         Higher values correspond with increased color differences. The
         default is set to 1 (corresponding to 1 pixel difference).
+        :param quality: Used for scaling an image to speed up calculations.
+        The default value is 1.0.
         :returns: An unsorted list of pixel values.
         """
 
-        if self.mode in ("F", "I", "L", "P"):
+        # Checking if # of pixels is greater than a 1080p image.
+        if quality >= 1.0 and self.width * self.height >= 2073600:
+            recommended_quality = 300000 / self.width * self.height
+            message = "Lower quality recommended: {:.4}".format(recommended_quality)
+            warnings.warn(message)
+        elif quality != 1.0:
+            self.thumbnail((quality * self.width, quality * self.height))
+
+        if self.mode in ("F", "L", "I", "P"):
             channels = 1
         elif self.mode in ("RGB", "YCbCr", "LAB", "HSV"):
             channels = 3
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -1272,6 +1272,77 @@ class Image:
             return self.im.getband(band)
         return self.im  # could be abused
 
+    def getdominantcolors(self, numcolors=3, maxiter=50, threshold=1):
+        """
+        Returns a list of dominant colors in an image using k-means
+        clustering.
+
+        :param numcolors: Number of dominant colors to search for.
+        The default number is 3.
+        :param maxiter: Maximum number of iterations to run the
+        algorithm.  The default limit is 50.
+        :param threshold: Early stopping condition for the algorithm.
+        Higher values correspond with increased color differences. The
+        default is set to 1 (corresponding to 1 pixel difference).
+        :returns: An unsorted list of (pixel) values.
+        """
+
+        def euclidean(p1, p2):
+            return sum([(p1[i] - p2[i]) ** 2 for i in range(channels)])
+
+        if self.mode in ("1", "L", "P"):
+            channels = 1
+        elif self.mode in ("RGB", "YCbCr", "LAB", "HSV"):
+            channels = 3
+        elif self.mode in ("RGBA", "CMYK"):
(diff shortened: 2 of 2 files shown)


Tests/test_image_getdominantcolors.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+def test_getdominantcolors():
+    def getdominantcolors(mode, numcolors=None):
+        im = hopper(mode)
+
+        if numcolors:
+            colors = im.getdominantcolors(numcolors)
+        else:
+            colors = im.getdominantcolors()
```
radarhere, line 11:
```suggestion
    def getdominantcolors(mode):
        im = hopper(mode)
        colors = im.getdominantcolors()
```
Just removing unused code, since this function is never called with `numcolors`.

src/PIL/Image.py
```diff
@@ -1272,6 +1272,98 @@ def getdata(self, band=None):
...
+
+        if self.mode in ("F", "I", "L", "P"):
+            channels = 1
+        elif self.mode in ("RGB", "YCbCr", "LAB", "HSV"):
+            channels = 3
+        elif self.mode in ("RGBA", "CMYK"):
+            channels = 4
+        else:
+            tb = sys.exc_info()[2]
+            raise ValueError("Unsupported image mode").with_traceback(tb)
```
radarhere, line 1298: Why raise the error with traceback?

src/PIL/Image.py
```diff
@@ -1272,6 +1272,98 @@ def getdata(self, band=None):
...
+        clustering.
+
+        :param numcolors: Number of dominant colors to search for.
+        The default number is 3.
+        :param maxiter: Maximum number of iterations to run the
+        algorithm.  The default limit is 50.
+        :param threshold: Early stopping condition for the algorithm.
+        Higher values correspond with increased color differences. The
+        default is set to 1 (corresponding to 1 pixel difference).
+        :returns: An unsorted list of (pixel) values.
```
radarhere, line 1287:
```suggestion
        :returns: An unsorted list of pixel values.
```
I suspect you copied this from the `getcolors` docstring. The brackets there were part of indicating the tuple contents. This function doesn't necessarily return a list of tuples.

src/PIL/Image.py
```diff
@@ -1272,6 +1272,98 @@ def getdata(self, band=None):
...
+
+        def euclidean(p1, p2):
+            if channels == 1:
+                return (p1 - p2) ** 2
+            return sum([(p1[i] - p2[i]) ** 2 for i in range(channels)])
+
+        w, h = self.size
+
+        pixels_and_counts = []
+        for count, color in self.getcolors(w * h):
```
radarhere, line 1308:
```suggestion
        pixels_and_counts = []
        for count, color in self.getcolors(self.width * self.height):
```
Just because you don't use `w` or `h` anywhere else, this looks simpler to me.

392781:
One thing to note about this is that k-means clustering takes a few seconds to compute for larger images which is bothersome.  I'm currently working on improving the speed of the algorithm.  

One catch-all method of doing this is to rescale the image for faster results.  This solution is simple and doesn't require a lot of bloated code on top of the algorithm.  The cost is that there is a loss of pixel information through rescaling so the dominant colors will come out different than original image.

Another is to do a bit of math on the color spaces used and throw out some less necessary pixels (kind of how [Color-Thief does it](https://github.com/fengsp/color-thief-py/blob/master/colorthief.py#L68) on the RGBA space.  The benefit is avoiding some more computation.  The cost is readability of the method and the need to cover all supported color spaces.  One way to sidestep this is to convert to a more familiar color space such as HSV, RGB, or RGBA; however, I'm not that knowledgable about these spaces and the effects of converting between them.

I'm hesitant to implement option #2 because I do not want this method to become too bloated or more complicated than it already is... But these are some ideas to think about.  For now I'll test out resizing as an option and maybe play around with converting modes.

hugovk:
The `Image` module is pretty big and I wonder if this would be better suited in another one?

* [`ImageColor`](https://pillow.readthedocs.io/en/stable/reference/ImageColor.html)
* [`ImageOps`](https://pillow.readthedocs.io/en/stable/reference/ImageOps.html)
* [`ImageStat`](https://pillow.readthedocs.io/en/stable/reference/ImageStat.html)

Tests/test_image_getdominantcolors.py
```diff
@@ -0,0 +1,18 @@
+from .helper import hopper
+
+
+def test_getdominantcolors():
+    def getdominantcolors(mode):
+        im = hopper(mode)
```
hugovk, line 6: I see what you mean about it being slow, `pytest Tests/test_image_getdominantcolors.py` takes ~5.67s on my Mac.

Resizing to `(10, 10)` reduces it to 0.23s.
Resizing to `(20, 20)` reduces it to 0.41s.

Do you think 10x10 might be too small to be useful?

```suggestion
        im = hopper(mode)
        im.thumbnail((10, 10))
```

Tests/test_image_getdominantcolors.py
```diff
@@ -0,0 +1,18 @@
...
+
+    assert getdominantcolors("F") == 3
+    assert getdominantcolors("I") == 3
+    assert getdominantcolors("L") == 3
+    assert getdominantcolors("P") == 3
+    assert getdominantcolors("RGB") == 3
+    assert getdominantcolors("YCbCr") == 3
+    assert getdominantcolors("CMYK") == 3
+    assert getdominantcolors("RGBA") == 3
+    assert getdominantcolors("HSV") == 3
```
hugovk, line 18: We can improve these tests so they're testing more than just they return a thing of length three.

We can also use `@pytest.mark.parametrize` so each test case is run independently so cannot depend on an earlier case.

Here's an example:

```python
import pytest

from .helper import hopper


@pytest.mark.parametrize(
    "test_mode, expected",
    [
        ("F", [28.8104386146252, 66.26185757773263, 127.53211228743844]),
        ("I", [136, 94, 40]),
        ("L", [25, 63, 127]),
        ("P", [11, 71, 159]),
        ("RGB", [(172, 117, 94), (53, 44, 55), (95, 127, 185)]),
        ("YCbCr", [(130, 108, 155), (123, 163, 105), (47, 131, 131)]),
        ("CMYK", [(201, 210, 199, 0), (159, 127, 69, 0), (82, 137, 160, 0)]),
        ("RGBA", [(31, 24, 37, 255), (129, 125, 150, 255), (87, 67, 72, 255)]),
        ("HSV", [(140, 131, 85), (177, 70, 46), (97, 131, 186)]),
    ],
)
def test_getdominantcolors(test_mode, expected):
    def getdominantcolors(mode):
        im = hopper(mode)
        im.thumbnail((10, 10))
        colors = im.getdominantcolors()
        return colors

    assert getdominantcolors(test_mode) == expected
```

What do you think?

Do those expected return values look right?

Might they be different on other systems? Especially the first one. If so, we can adjust it somewhat, possibly have a special case for that.

It would be good to add further `test_X` functions to verify the other parameters of `im.getdominantcolors`, plus warnings and exceptions.

nulano:
I'm not sure ImageOps is the right place. It is a set of functions that return a processed variation of the source image.
> The ImageOps module contains a number of ‘ready-made’ image processing operations.

---

ImageColor doesn't seem right either:
> The ImageColor module contains color tables and converters from CSS3-style color specifiers to RGB tuples. This module is used by PIL.Image.new() and the ImageDraw module, among others.

---

ImageStat looks like the closest match, but the `PIL.ImageStat.Stat` class caches results and the proposed function takes parameters. Maybe make this a module function in ImageStat (e.g. `PIL.ImageStat.dominant_colors(im, ...)`)?
> The ImageStat module calculates global statistics for an image, or for a region of an image.

Tests/test_image_getdominantcolors.py
392781: Small sizes like that give a very rough approximation especially if you're looking for a palette that's more than 3 colors.  The colors returned tend to be much darker.  I found that `(100,100)` gives somewhat decent results.

src/PIL/Image.py
```diff
@@ -1272,6 +1272,106 @@ def getdata(self, band=None):
...
+
+        if self.mode in ("F", "L", "I", "P"):
+            channels = 1
+        elif self.mode in ("RGB", "YCbCr", "LAB", "HSV"):
+            channels = 3
+        elif self.mode in ("RGBA", "CMYK"):
+            channels = 4
+        else:
+            tb = sys.exc_info()[2]
+            raise ValueError("Unsupported image mode").with_traceback(tb)
```
radarhere, line 1308:
```suggestion
            raise ValueError("Unsupported image mode")
```

radarhere, line 1308:
```suggestion
            raise ValueError("Unsupported image mode")
```

src/PIL/Image.py
```diff
@@ -1272,6 +1272,105 @@ def getdata(self, band=None):
...
+        elif quality != 1.0:
+            self.thumbnail((quality * self.width, quality * self.height))
+
+        if self.mode in ("F", "L", "I", "P"):
+            channels = 1
+        elif self.mode in ("RGB", "YCbCr", "LAB", "HSV"):
+            channels = 3
+        elif self.mode in ("RGBA", "CMYK"):
+            channels = 4
+        else:
```
radarhere, line 1306:
```suggestion
        channels = self.im.bands
        if channels not in (1, 3, 4):
```

Issue #2802, opened 2017-10-16
SystemError: new style getargs format but argument is not a tuple


spaceexperiment:
I am trying to generate thumb from png images, but i get this error sometime.
it's fixed by opening the image in photoshop and saving it again.

i am using sorl.thumbnail, but the error seems to be from PIL

```python
im = get_thumbnail(p.image, '100x100', crop='center')
```

here is the error i am getting

```python
 File "/.../lib/python3.5/site-packages/sorl/thumbnail/engines/pil_engine.py", line 103, in _colorspace
    mask = Image.new('L', image.size, color=transparency)
  File "/.../lib/python3.5/site-packages/PIL/Image.py", line 2287, in new
    return Image()._new(core.fill(mode, size, color))
SystemError: new style getargs format but argument is not a tuple
```

wiredfool:
You're passing something in for color, and it's not getting interpreted correctly.  Given that you're passing in 'transparency', you should make sure that it's a valid colorspec. I'd suggest that it should be a single int, 0<= color <=255. (while there are other options, transparency is not likely to be a string with a css color, e.g. "white")

homm:
@spaceexperiment Which version of Pillow do you use?

spaceexperiment:
@homm I am using 4.2.1, tried 4.3.0, still the same.
I was able to make things work by converting all my images to PNG and convert them to mode 'RGBA' , i noticed for example the issue was when for example the image PNG and the color mode was 'P'. converting  it to 'RGBA' fixed it.

I am assuming sorl.thumbnail is passing the transparency color for all image even if the image don't support it ?, but i am  not really sure.

I opened the same issue here [https://github.com/jazzband/sorl-thumbnail/issues/521](https://github.com/jazzband/sorl-thumbnail/issues/521)

PR #4875, opened 2020-08-18
Throw error on incorrect type in Image.putpixel

applemonkey496:
For `Image` objects with mode `L` (greyscale), when running `img.putpixel`, if the argument `xy` is not a tuple with types (int, int), the code raises a rather cryptic error:
```
```
Again, if the argument **is** a tuple, but not with the types `(int, int)`, this error will come up, which is likely to confuse many people. I propose adding an `if` statement in the `putpixel` function to catch this case, and then to throw a clearer error message:
```
Expected argument xy to be a tuple (int, int) but got { xy }
```
Where `xy` is replaced by whatever acutal value of `xy` caused the error.
____________
 * Related: [#2802](https://github.com/python-pillow/Pillow/issues/2802)
 * I got the idea to submit this pull request after reading [this issue](https://stackoverflow.com/a/43656642/11151112) on StackOverflow

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -1733,11 +1733,13 @@ class Image:
         ):
             # RGB or RGBA value for a P image
             value = self.palette.getcolor(value)
-            
-        if self.mode = "L":
+
+        if self.mode == "L":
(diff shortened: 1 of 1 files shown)


radarhere:
Closing in favour of #4882

PR #4882, opened 2020-08-19
Raise proper TypeError in putpixel

### LINKED ISSUE #4876: Throw error on incorrect type in Image.putpixel
*This text is pasted from my PR, #4875*

For `Image` objects with mode `L` (greyscale), when running `img.putpixel`, if the argument `xy` is not a tuple with types (int, int), the code raises a rather cryptic error:
```
```
Again, if the argument **is** a tuple, but not with the types `(int, int)`, this error will come up, which is likely to confuse many people. I propose adding an `if` statement in the `putpixel` function to catch this case, and then to throw a clearer error message:
```
Expected argument xy to be a tuple (int, int) but got { xy }
```
Where `xy` is replaced by whatever acutal value of `xy` caused the error.

**See the [Pull Request](https://github.com/python-pillow/Pillow/pull/4875).**
____________
 * Related: [#2802](https://github.com/python-pillow/Pillow/issues/2802)
 * I got the idea to submit this pull request after reading [this answer](https://stackoverflow.com/a/43656642/11151112) on StackOverflow

nulano:
Fixes #4876:

* The existing code silently ignores an `OverflowError` for 8bpp images, hoping that the int will become a tuple in the next few lines.
* It also fails to check that the given parameter is a tuple before trying to unpack it, giving the less than useful message `SystemError: new style getargs format but argument is not a tuple`.
* Last but not least, passing a non-int value for non-8bpp non-float images (e.g. `I;16`) gives an even less useful message `SystemError: <method 'putpixel' of 'ImagingCore' objects> returned NULL without setting an error`.

This PR resolves all of these issues by explicitly raising a TypeError if the parameter can't be converted to an int, or a tuple for an 8bpp image.

### DIFF
--- a/Tests/test_image_access.py
+++ b/Tests/test_image_access.py
@@ -328,6 +328,36 @@ class TestCffi(AccessTest):
             assert im.convert("RGB").getpixel((0, 0)) == (255, 0, 0)
 
 
+class TestImagePutPixelError(AccessTest):
+    def test_putpixel_error_message(self):
+        for mode, reason, accept_tuple in [
+            ("L", "color must be int or tuple", True),
+            ("LA", "color must be int or tuple", True),
+            ("RGB", "color must be int or tuple", True),
+            ("RGBA", "color must be int or tuple", True),
+            ("I", "color must be int", False),
+            ("I;16", "color must be int", False),
+            ("BGR;15", "color must be int", False),
+        ]:
+            im = hopper(mode)
+
+            for v in ["foo", 1.0, None]:
+                with pytest.raises(TypeError, match=reason):
+                    im.putpixel((0, 0), v)
+
+            if not accept_tuple:
+                with pytest.raises(TypeError, match=reason):
+                    im.putpixel((0, 0), (10,))
+
+            with pytest.raises(OverflowError):
+                im.putpixel((0, 0), 2 ** 80)
+
+        for mode in ["BGR;15"]:
+            im = hopper(mode)
+            with pytest.raises(ValueError, match="unrecognized image mode"):
+                im.putpixel((0, 0), 0)
+
+
 class TestEmbeddable:
     @pytest.mark.skipif(
         not is_win32() or on_ci(),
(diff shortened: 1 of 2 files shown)


Tests/test_image_access.py
```diff
@@ -329,33 +329,36 @@ def test_p_putpixel_rgb_rgba(self):
...
-
-            with pytest.raises(OverflowError):
-                im.putpixel((0, 0), 2 ** 80)
-
-        for mode in ["BGR;15"]:
-            im = hopper(mode)
-            with pytest.raises(ValueError, match="unrecognized image mode"):
-                im.putpixel((0, 0), 0)
+    IMAGE_MODES1 = ["L", "LA", "RGB", "RGBA"]
+    INVALID_TYPES1 = ["foo", 1.0, None]
```
hugovk, line 333: Group these together?
```suggestion
    IMAGE_MODES1 = ["L", "LA", "RGB", "RGBA"]
    IMAGE_MODES2 = ["I", "I;16", "BGR;15"]
    INVALID_TYPES1 = ["foo", 1.0, None]
    INVALID_TYPES2 = [*INVALID_TYPES1, (10,)]    
```

Tests/test_image_access.py
```diff
@@ -329,33 +329,36 @@ def test_p_putpixel_rgb_rgba(self):
...
+    @pytest.mark.parametrize("mode", IMAGE_MODES1)
+    def test_putpixel_type_error1(self, mode):
+        im = hopper(mode)
+        for v in self.INVALID_TYPES1:
+            with pytest.raises(TypeError, match="color must be int or tuple"):
+                im.putpixel((0, 0), v)
+
+    IMAGE_MODES2 = ["I", "I;16", "BGR;15"]
+    INVALID_TYPES2 = [*INVALID_TYPES1, (10,)]
+
```
hugovk, line 344:
```suggestion
```

hugovk:
Thanks! Do you think this should go in the release notes?

https://pillow.readthedocs.io/en/latest/releasenotes/8.0.0.html

nulano:
Probably not. Release notes typically mention API changes or additions, and significant bugfixes. *Fixed error message when an invalid color is specified* does not sound very noteworthy, especially since most users will probably only encounter the error message once during development of their application.

Issue #4237, opened 2019-11-28
OSError: -2 when decoding a tiff_lzw file


kmilos:
This is with pillow 6.2.1 on 

Python 3.7.3 | packaged by conda-forge | (default, Jul  1 2019, 22:01:29) [MSC v.1900 64 bit (AMD64)]

Seems libtiff is found without a problem, but decoding fails:

```python
In [5]: dat = img.getdata()
have fileno, calling fileno version of the decoder.
Traceback (most recent call last):

  File "<ipython-input-5-d82315236633>", line 1, in <module>
    dat = img.getdata()

  File "C:\Users\kmilos\.conda\envs\dev\lib\site-packages\PIL\Image.py", line 1304, in getdata
    self.load()

  File "C:\Users\kmilos\.conda\envs\dev\lib\site-packages\PIL\TiffImagePlugin.py", line 1097, in load
    return self._load_libtiff()

  File "C:\Users\kmilos\.conda\envs\dev\lib\site-packages\PIL\TiffImagePlugin.py", line 1209, in _load_libtiff
    raise IOError(err)

OSError: -2


In [6]: dat = img.getdata()

In [7]: print(dat)
<ImagingCore object at 0x000001FC04A55DB0>
```

Also note that on the second attempt I get an object back, bit if I try to convert to an array, it is filled with zeros, so issue #3863 is not really closed for me.

No problems if I use 'TIFF' or 'TIFF-FI' formats via imageio, but 'TIFF-PIL' fails with the same error.

radarhere:
Could you provide the image that is causing the problem?

kmilos:
Happens also with this tiff_adobe_deflate compressed [mandrill.zip](https://github.com/python-pillow/Pillow/files/3901288/mandrill.zip), so not limited to tiff_lzw...

radarhere:
Loading the mandrill image, I'm unable to replicate using Pillow 6.2.1 with Python 3.7.5 on macOS 10.15.
```python
from PIL import Image
im = Image.open('mandrill.tif')
im.load()
```

kmilos:
On Windows it is line 1189 of tiffimageplugin.py where the problem occurs

`n, err = decoder.decode(b"fpfp")`

results in n=-1 and err=-2

radarhere:
I'm also unable to replicate with 6.2.1 on AppVeyor Windows or master on GHA Windows.

kmilos:
[...]

PR #4890, opened 2020-08-30
MSYS2 fixes

nulano:

 * Remove old include directory `C:\\msys64\\mingw32\\include\\libimagequant`.

   It uses a hardcoded path to MSYS2, it is hardcoded to 32-bit, and it is no longer necessary, as the include is now in `C:\\msys64\\mingw32\\include\\libimagequant.h`.

 * Remove GCC workaround for MSYS, see https://github.com/python-pillow/Pillow/issues/4796#issuecomment-683370940

   > The issue was that AppVeyor was missing the MSYSTEM=MINGW32 envvar, and getting confused by the default MSYSTEM=MSYS2.

 * Fix LibTiff support. The issue is that MSYS2 builds LibTiff with `USE_WIN32_FILEIO`, which changes the meaning of the `fd` parameter for `TIFFFdOpen`, see https://github.com/python-pillow/Pillow/issues/4237#issuecomment-569403417. I've added a workaround by converting the parameter before passing it to LibTiff if building on MinGW.

   It might be possible to detect the system version of LibTiff by including the `tif_config.h` header to fix&nbsp;#4237 (by enabling the workaround if `USE_WIN32_FILEIO` was enabled), but it appears to be a private header (while it is available in MSYS2, it is not available in Ubuntu). See also [passing build using CMake on Windows](https://github.com/nulano/Pillow/runs/1046684238) from my branch https://github.com/nulano/Pillow/compare/mingw-setup...nulano:tiff-test.

   Another possible fix would involve absorbing `TIFFFdOpen` from `tif_unix.c` into Pillow to avoid this issue by using the stable `TiffClientOpen`.

 * [Document installation on MSYS2](https://pillow--4890.org.readthedocs.build/en/4890/installation.html#building-on-windows-using-msys2-mingw)


docs/installation.rst
```diff
@@ -299,6 +301,42 @@ If you wish to compile Pillow manually, you can use the build scripts
...
+    pacman -S \
+        mingw-w64-x86_64-libjpeg-turbo \
+        mingw-w64-x86_64-zlib \
+        mingw-w64-x86_64-libtiff \
+        mingw-w64-x86_64-freetype \
+        mingw-w64-x86_64-lcms2 \
+        mingw-w64-x86_64-libwebp \
+        mingw-w64-x86_64-openjpeg2 \
+        mingw-w64-x86_64-libimagequant \
+        mingw-w64-x86_64-libraqm
```
radarhere, line 332:
```suggestion
        mingw-w64-x86_64-freetype \
        mingw-w64-x86_64-lcms2 \
        mingw-w64-x86_64-libimagequant \
        mingw-w64-x86_64-libjpeg-turbo \
        mingw-w64-x86_64-libraqm
        mingw-w64-x86_64-libtiff \
        mingw-w64-x86_64-libwebp \
        mingw-w64-x86_64-openjpeg2 \
        mingw-w64-x86_64-zlib \
```
Sorted

nulano: The order is the same as in the External Libraries section above. I think it is more helpful to follow that order for easier cross-referencing if someone doesn't want a full install.

lazka:
You could use https://github.com/marketplace/actions/setup-msys2

hugovk:
Thanks!

Issue #4827, opened 2020-07-30
py -3.9 -m pip install -U pillow fails with pip 20.2


djhenderson:
`py -3.9 -m pip install -U pillow`

fails, includes these messages:

```
  ERROR: Failed building wheel for pillow
  Running setup.py clean for pillow
Failed to build pillow
DEPRECATION: Could not build wheels for pillow which do not use PEP 517. pip will fall back to legacy 'setup.py install' for these. pip 21.0 will remove support for this functionality. A possible replacement is to fix the wheel build issue reported above. You can find discussion regarding this at https://github.com/pypa/pip/issues/8368.
Installing collected packages: pillow
    Running setup.py install for pillow ... error
    ERROR: Command errored out with exit status 1:
```

using: 

```
> py -3.9 -m pip --version
pip 20.2 from C:\Users\Doug\AppData\Local\Programs\Python\Python39\lib\site-packages\pip (python 3.9)
```

```
> py -3.9
Python 3.9.0b5 (tags/v3.9.0b5:8ad7d50, Jul 20 2020, 18:35:09) [MSC v.1924 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>>
```

On windows 10 pro 1909, full patched as of 2020-07-30.

Dinoosawruss:
I'm experiencing the same issue

hugovk:
Pillow doesn't yet provide prebuilt wheels for Python 3.9, and it's attempting to build from source, which is fiddly on Windows. The 3.9 wheels are scheduled for Pillow 8 in October after Python 3.9.0 is out. (Unofficial Windows wheels for 3.9 are here: https://www.lfd.uci.edu/~gohlke/pythonlibs/#Pillow) 

You should be getting this warning:

https://github.com/python-pillow/Pillow/blob/48c7d84bf771c3cf9bf9c0aba96b4ce12c449c8c/setup.py#L40-L46

Do you get it? Please could you post the full traceback?

If not, do you get the warning with pip 20.1?

hugovk:
I tested Python 3.9 with pip 20.1 and 20.2 on GitHub Actions, on Windows, and got the warning at the top of each traceback:

* https://github.com/hugovk/test/runs/931018810?check_suite_focus=true
* https://github.com/hugovk/test/runs/931018847?check_suite_focus=true

Also tested with wheel installed, similar thing:

* https://github.com/hugovk/test/runs/931032374?check_suite_focus=true
[...]

PR #4891, opened 2020-08-30
Revert #4832: Removed 32-bit MinGW job

nulano:
Reverts #4827.
Update CI targets in `installation.rst`.

While working on #4890 I reread https://www.msys2.org/news/#2020-05-17-32-bit-msys2-no-longer-actively-supported and realized it is not referring to MSYS2 MinGW 32-bit, but the 32-bit version of MSYS2.

> 32-bit mingw-w64 packages are still supported, this is about the POSIX emulation layer, i.e. the runtime, Bash, MinTTY...

I.e. both MSYS2 MinGW 64-bit (`mingw-w64-x86_64`) and MSYS2 MinGW 32-bit (`mingw-w64-i686`) are still supported on 64-bit Windows.

> After this date, we don't plan on building updated msys-i686 packages nor releasing i686 installers anymore. This is due to increasingly frustrating difficulties with limited 32-bit address space, high penetration of 64-bit systems and Cygwin (our upstream) starting their way to drop 32-bit support as well.

I.e. MSYS2 (the installer) is no longer supported on 32-bit Windows.

To clarify a bit, the MSYS2 subsystem is based on Cygwin, while the MinGW subsystems provide native Windows executables cross-compiled using MSYS2.


hugovk:
Welcome back to the CI, MSYS2 MinGW 32-bit!

PR #4892, opened 2020-08-30
Update pre-commit

radarhere:


hugovk (commented):
As a note, the new release of Black has something called a "magic comma", which tells Black to explode a collection to multiple lines:

* https://black.readthedocs.io/en/stable/the_black_code_style.html#the-magic-trailing-comma

Some of the ones here could remain single lines by removing that trailing comma, but I'm fine with these expanding. Feel free to change any if you like.

src/PIL/PngImagePlugin.py
```diff
@@ -1110,7 +1110,10 @@ def _write_multiple_frames(im, fp, chunk, rawmode):
 
     # animation control
     chunk(
-        fp, b"acTL", o32(len(im_frames)), o32(loop),  # 0: num_frames  # 4: num_plays
+        fp,
+        b"acTL",
+        o32(len(im_frames)),
+        o32(loop),  # 0: num_frames  # 4: num_plays
```
hugovk, line 1116: Yup, before [this](https://github.com/python-pillow/Pillow/commit/512b060a53cc189770646414ce91e552eb01fe2b#diff-07cbf160cd247acb9b67060ff3d81ed9L1074), used to be like this:
 
```suggestion
        fp,
        b"acTL",
        o32(len(im_frames)),  # 0: num_frames
        o32(loop),  # 4: num_plays
```

hugovk (approved):

PR #4905, opened 2020-09-07
Docs: Give supported Python versions its own header

hugovk:

 * Right now, linking to the table is really a link to "Notes":
https://pillow.readthedocs.io/en/stable/installation.html#notes
 * It's important enough to deserve its own header, not just "Notes"
 * I think keeping it near the top is good
 * This was the only "Note" under "Notes". If we get more notes in the future, they can be back after "Warnings"

https://pillow--4905.org.readthedocs.build/en/4905/installation.html#python-support

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -10,11 +10,10 @@ Warnings
 
 .. warning:: Pillow >= 2.1.0 no longer supports "import _imaging". Please use "from PIL.Image import core as _imaging" instead.
 
-Notes
------
+Python Support
+--------------
 
-.. note:: Pillow is supported on the following Python versions
-
+Pillow supports these Python versions.
 
 +----------------------+-------+-------+-------+-------+-------+-------+-------+-------+-------+-------+-------+-------+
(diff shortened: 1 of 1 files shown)


docs/installation.rst
```diff
@@ -10,11 +10,10 @@ Warnings
 
 .. warning:: Pillow >= 2.1.0 no longer supports "import _imaging". Please use "from PIL.Image import core as _imaging" instead.
 
-Notes
------
-
-.. note:: Pillow is supported on the following Python versions
+Python Support
+--------------
```
radarhere, line 14:
```suggestion
Python Version Support
----------------------
```
Feel free to disregard this suggestion. My thinking is that this is not about whether or not we support Python, but about what version of Python we support.

radarhere (approved):

docs/installation.rst
hugovk: I think I'll keep it for brevity, and it should be ambiguous enough given the section contents and that this is a Python Imaging Library :) But thanks!

PR #4909, opened 2020-09-08
Clarify that modes are raw strings, not constants

MauveRanger:
Fixes # a small mistake I made while reading the docs. I was trying to initialize an image and didn't know whether modes were constants (like in tkinter.CENTER) or raw strings (like in open(filename, 'rwb').

 *  change one line in concepts.rst

### DIFF
--- a/docs/handbook/concepts.rst
+++ b/docs/handbook/concepts.rst
@@ -24,7 +24,7 @@ To get the number and names of bands in an image, use the
 Modes
 -----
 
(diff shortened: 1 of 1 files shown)


hugovk:
Please could you update the PR to target the `master` branch?

(I tried, but it included a load of irrelevant commits, so switched it back.)

MauveRanger:
Ok. I'm not entirely sure how to do that, but I will try.

PR #4912, opened 2020-09-09
Clarify that modes are strings, not constants

MauveRanger:
I misunderstood this section of the documentation the first time I read it. I hope this change will clarify how to request a mode such as from Image.new("RGBA", size).

 * Add that modes are strings, as opposed to constants used like tkinter.NW


hugovk:
Thank you!

PR #4914, opened 2020-09-10
Fixed loading profile with non-ASCII path on Windows

### LINKED ISSUE #3145: ImageFont and non latin characters in a path to font
### What did you do?

```
>>> from PIL.ImageFont import truetype
>>> import os.path
>>> fname = 'C:\\temp\\Шрифты\\font.ttf'
>>> os.path.exists(fname)
True
>>> truetype(fname)
  File "C:\Users\jenyay\.virtualenvs\outwiker-lFR0QK22\lib\site-packages\PIL\ImageFont.py", line 261, in truetype
  File "C:\Users\jenyay\.virtualenvs\outwiker-lFR0QK22\lib\site-packages\PIL\ImageFont.py", line 144, in __init__
    self.font = core.getfont(font, size, index, encoding, layout_engine=layout_engine)
OSError: cannot open resource
```

If `fname` contains only Latin characters, then everything works.

Pillow 5.0.1
Windows 7, Windows 10.
In Linux is OK.

### LINKED ISSUE #4897: ImageCms.getOpenProfile does not support Unicode path
### What did you do?

I tried to load an ICC file from a non-Latin path.

It should load the profile.

### What actually happened?

It throws

```
F:\sync\code\python\_gists>_test.py
  File "C:\Program Files\Python3\lib\site-packages\PIL\ImageCms.py", line 395, in getOpenProfile
    return ImageCmsProfile(profileFilename)
  File "C:\Program Files\Python3\lib\site-packages\PIL\ImageCms.py", line 162, in __init__
    self._set(core.profile_open(profile), profile)
OSError: cannot open profile file

The above exception was the direct cause of the following exception:

  File "F:\sync\code\python\_gists\_test.py", line 3, in <module>
    lsrgb_profile = ImageCms.getOpenProfile('D:\\新\\sRGB.icc')
  File "C:\Program Files\Python3\lib\site-packages\PIL\ImageCms.py", line 397, in getOpenProfile
    raise PyCMSError(v) from v
PIL.ImageCms.PyCMSError: cannot open profile file
```

* OS:  Win7x64
* Python: 3.8.5 x64
* Pillow: 7.2.0

```python
from PIL import Image, ImageCms

lsrgb_profile = ImageCms.getOpenProfile('D:\\新\\sRGB.icc')
```

Similar and fixed issue: https://github.com/python-pillow/Pillow/issues/3145

radarhere:
Resolves #4897 - in the same way that #3785 resolved #3145, by loading data from a non-Unicode path on Windows into memory.

Also corrects a message in the test suite from that PR.

### DIFF
--- a/Tests/test_imagecms.py
+++ b/Tests/test_imagecms.py
@@ -1,6 +1,7 @@
 import datetime
 import os
 import re
+import shutil
 from io import BytesIO
 
 import pytest
@@ -434,6 +435,19 @@ def test_extended_information():
     assert p.version == 2.0
     assert p.viewing_condition == "Reference Viewing Condition in IEC 61966-2-1"
     assert p.xcolor_space == "RGB "
+
+
+def test_non_ascii_path(tmp_path):
+    skip_missing()
+    tempfile = str(tmp_path / ("temp_" + chr(128) + ".icc"))
+    try:
+        shutil.copy(SRGB, tempfile)
+    except UnicodeEncodeError:
+        pytest.skip("Non-ASCII path could not be created")
+
+    o = ImageCms.getOpenProfile(tempfile)
+    p = o.profile
+    assert p.model == "IEC 61966-2-1 Default RGB Colour Space - sRGB"
 
 
 def test_profile_typesafety():
--- a/Tests/test_imagefont.py
+++ b/Tests/test_imagefont.py
@@ -111,12 +111,12 @@ class TestImageFont:
         with open(FONT_PATH, "rb") as f:
             self._render(f)
 
-    def test_non_unicode_path(self, tmp_path):
+    def test_non_ascii_path(self, tmp_path):
         tempfile = str(tmp_path / ("temp_" + chr(128) + ".ttf"))
         try:
             shutil.copy(FONT_PATH, tempfile)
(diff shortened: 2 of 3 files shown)


hugovk:
What does "non-Unicode" mean here? ASCII?

fireattack:
Yeah, was about to comment the title before :p I think he meant non-ASCII.

nulano:
LCMS2 [calls `fopen` internally](https://github.com/mm2/Little-CMS/blob/d6eb05ae6f7a6b81f11314fd36b580ed2dd2f5a5/src/cmsio0.c#L388), same as FreeType, which on Windows expects a string encoded with the current ANSI codepage: https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/fopen-wfopen?view=vs-2019#remarks

Python returns strings in the `utf-8` encoding for the `s` parameter in `PyArg_ParseTuple`. These two encodings are only compatible in the ASCII range unless the current codepage is UTF-8, but that is [not yet widely supported](https://en.wikipedia.org/wiki/Unicode_in_Microsoft_Windows#UTF-8).

So it seems non-ASCII is the intended message.

hugovk:
Thanks!

PR #4915, opened 2020-09-11
Winbuild: Enable FT_CONFIG_OPTION_USE_HARFBUZZ

nulano:
I have finally figured out how to resolve the circular dependency between FreeType and HarfBuzz on Windows.

From `ftoption.h`:
> FreeType uses the HarfBuzz library to improve auto-hinting of OpenType fonts.  If available, many glyphs not directly addressable by a font's character map will be hinted also.
>
>Define this macro if you want to enable this 'feature'.

While I was not able to find a font and text combination that shows a difference, if such a combination exists, someone using these scripts locally or downloading GHA artifacts could benefit from this.

The difference to compile times is imperceptible, `_imagingft.cp37-win_amd64.pyd` is 0.6MB larger, the GHA artifact increases from 13MB to 15.2MB (17%). I don't expect this to go over the (unspecified) GHA limit, but there is an action to remove old artifacts before the 90 timeout if necessary: https://github.com/c-hive/gha-remove-artifacts


hugovk (approved):

hugovk:
Thanks!

PR #4923, opened 2020-09-16
Imaging.h: confusion with system

ax3l:
The file `libImaging/Imaging.h` gets installed [on Ubuntu flat into the public Python includes](https://packages.ubuntu.com/search?searchon=contents&keywords=Imaging.h&mode=exactfilename&suite=focal&arch=any).

When building a newer version of Pillow from source and hinting the Python includes "too early" in includes, e.g. in [package managers](https://github.com/spack/spack/issues/18589#issuecomment-693165252), this can confuse the two files and pick up the external file over the internal one. With different versions, this mismatch can lead to build errors, e.g. undefined macros.

The most robust way to avoid this is to prefix the internal include accordingly, so that the relative path to the including file has to match as well.


radarhere:
I've created https://github.com/ax3l/Pillow/pull/1 as a suggestion to expand on this - prefixing more header includes.

wiredfool:
I think this is a bug in the debian packaging. `Imaging.h` is an internal file, not a supported set of external interfaces. 

Following the debian rules for splitting and naming packages, if the header is actually required (e.g, for other debian packages), it should be in a package named python-pillow-dev, not in the primary python-pillow package. (e.g, libfoo and libfoo-dev for compilation).

radarhere:
If it's not a problem though, is it worth doing just because it makes it clearer to the reader where the header file is located?

ax3l:
I think even if it were in a `-dev` package, we would have the same problem.

radarhere (approved):

hugovk (approved):

ax3l:
Thank you for the update, LGTM.

nulano:
I tried to remove the `libImaging` directory from the includes added in `setup.py` so future includes can't omit the prefix by accident, but this causes a failure building `_imagingft.c` on systems without libraqm. Fixing this requires changing
```c
#include <raqm.h>
```
to
```c
#include "libImaging/raqm.h"
```
so I'm not sure it is worth it. Thoughts?

https://github.com/ax3l/Pillow/compare/fix-imagingHinclude...nulano:fix-imagingHinclude

ax3l:
Oh, but that means instead of the internally shipped `raqm.h` it was picking up a system one already. Yes, would add this.

hugovk:
Merged in `master` to include https://github.com/python-pillow/Pillow/pull/4978 to fix the unrelated CI failure.

PR #4927, opened 2020-09-20
Allow tuples with one item to give single color value in getink

### LINKED ISSUE #4925: SystemError on Image.new / fill for 32-bit mode with color tuple
The following works smoothly:

```python
>>> PIL.Image.new('L', (200,100), (255,)) # accepts tuple
```

However, for 32-bit modes, this does not work:
```python
>>> PIL.Image.new('F', (200,100), (255,)) # fails okay
  File "/lib/python3.6/site-packages/PIL/Image.py", line 2613, in new
    return im._new(core.fill(mode, size, color))
TypeError: must be real number, not tuple
>>> PIL.Image.new('I', (200,100), (255,)) # fails badly
  File "/lib/python3.6/site-packages/PIL/Image.py", line 2613, in new
    return im._new(core.fill(mode, size, color))
SystemError: <built-in function fill> returned NULL without setting an error
```

At least the error message in the latter case should be improved.

But why not be more tolerant in both cases (I or F), like above (L or P)?

(Seen with Pillow 7.2.0 on Python 3.6.7 Linux x86_64.)

radarhere:
Resolves #4925

Normally, a user would run `Image.new('F', (100,100), 255)` and `Image.new('I', (100,100), 255)`, providing a single value for `color` for those modes.

This PR allows users to also provide that single value inside a tuple, `Image.new('F', (200,100), (255,))` and `Image.new('I', (200,100), (255,))`, which would have previously thrown an error.

### DIFF
--- a/src/_imaging.c
+++ b/src/_imaging.c
@@ -518,6 +518,9 @@ getink(PyObject* color, Imaging im, char* ink)
        be cast to either UINT8 or INT32 */
 
     int rIsInt = 0;
+    if (PyTuple_Check(color) && PyTuple_Size(color) == 1) {
+        color = PyTuple_GetItem(color, 0);
+    }
     if (im->type == IMAGING_TYPE_UINT8 ||
         im->type == IMAGING_TYPE_INT32 ||
         im->type == IMAGING_TYPE_SPECIAL) {
@@ -533,7 +536,7 @@ getink(PyObject* color, Imaging im, char* ink)
                 return NULL;
             }
         } else {
-            PyErr_SetString(PyExc_TypeError, "color must be int");
+            PyErr_SetString(PyExc_TypeError, "color must be int or tuple");
             return NULL;
         }
     }
--- a/src/_imaging.c
+++ b/src/_imaging.c
@@ -536,7 +536,7 @@ getink(PyObject* color, Imaging im, char* ink)
                 return NULL;
             }
         } else {
-            PyErr_SetString(PyExc_TypeError, "color must be int or tuple");
+            PyErr_SetString(PyExc_TypeError, "color must be int or single-element tuple");
             return NULL;
         }
     }
--- a/Tests/test_image.py
+++ b/Tests/test_image.py
@@ -498,6 +498,12 @@ class TestImage:
         with pytest.raises(ValueError):
             Image.core.fill("RGB", (2, -2), (0, 0, 0))
 
+    def test_one_item_tuple(self):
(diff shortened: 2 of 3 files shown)


src/_imaging.c
```diff
@@ -533,7 +536,7 @@ getink(PyObject* color, Imaging im, char* ink)
                 return NULL;
             }
         } else {
-            PyErr_SetString(PyExc_TypeError, "color must be int");
+            PyErr_SetString(PyExc_TypeError, "color must be int or tuple");
```
nulano, line 539:
```python
>>> PIL.Image.new('I', (200,100), (255,255,255))
...
TypeError: color must be int or tuple
```
is not very helpful.

The following suggestion would avoid confusion if a user specifies a tuple with two or more values:
```suggestion
            PyErr_SetString(PyExc_TypeError, "color must be int or single-element tuple");
```

The `IMAGING_TYPE_UINT8` branch provides a more specific error message for an invalid tuple below (from `PyArg_ParseTuple`).

src/_imaging.c
```diff
@@ -518,6 +518,9 @@ getink(PyObject* color, Imaging im, char* ink)
        be cast to either UINT8 or INT32 */
 
     int rIsInt = 0;
+    if (PyTuple_Check(color) && PyTuple_Size(color) == 1) {
+        color = PyTuple_GetItem(color, 0);
+    }
```
nulano, line 523: I think this currently makes `L` mode accept `tuple(tuple(10))` and `RGB` mode accept `tuple(tuple(255,255,255))`, which is probably not the intended behaviour.
```suggestion
    /* IMAGING_TYPE_UINT8 unpacks tuples later */
    if (im->type != IMAGING_TYPE_UINT8 && PyTuple_Check(color) && PyTuple_Size(color) == 1) {
        color = PyTuple_GetItem(color, 0);
    }
```

nulano (approved):
I think the current version will accept a tuple-in-tuple for `L` and `RGB` modes, LGTM otherwise.

hugovk (approved):

NightMachinary:
My code broke, probably because of this PR:
```python
            image = PIL.ImageOps.pad(
                image,
                (max(image.width, min_width), max(image.height, min_height)),
                color=(255, 255, 255),
            )
```

How do you now provide an RGB tuple?

radarhere:
I don't think this PR should have broken your code. What mode is your image? What error are you receiving?

PR #4929, opened 2020-09-22
Moved CVE images to pillow-depends

### LINKED ISSUE #4730: Antivirus Avast detect  test  image crash_2.tif as with  exploit

radarhere:
Resolves #4730. Alternative to #4869

This PR is part of moving those images to [pillow-depends](https://github.com/python-pillow/pillow-depends). https://github.com/python-pillow/pillow-depends/pull/32 is the other part, and this PR fails without it.

This PR also changes the test so that it is run as part of the test suite, but skips the test if run locally without the test images.


nulano:
Would it make sense to put the test images in a subdirectory to make them easier to remove from a checked-out copy? All future CVE examples could be put in there as well.

For example `Tests/images/cve/crash_1.tif`, similarly to https://github.com/python-pillow/Pillow/blob/c2367400fa8aab47d7752f3780681fced044f0d2/Tests/test_file_msp.py#L10-L11 https://github.com/python-pillow/Pillow/blob/c2367400fa8aab47d7752f3780681fced044f0d2/Tests/test_file_msp.py#L59-L61

hugovk:
Yeah, we could put them in a `cve` subdir. 

Normally we can't disclose CVE details until the fix is released and announced, but I expect we'd first put them in the main repo, and only move them to the other repo after someone reports a problem with their virus scanner?

Or let's keep it this way, and use a `cve` subdir if the need actually arises to be able easily delete them. Most people don't have test_images from the other repo anyway.

nulano:
> Normally we can't disclose CVE details until the fix is released and announced, but I expect we'd first put them in the main repo, and only move them to the other repo after someone reports a problem with their virus scanner?
> 
> Or let's keep it this way, and use a `cve` subdir if the need actually arises to be able easily delete them. Most people don't have test_images from the other repo anyway.

Fair points, I suppose these can be moved into a subdir if/when someone reports another file flagged by a virus scanner.

hugovk:
https://github.com/python-pillow/pillow-depends/pull/32 is merged, and CIs restarted here.

hugovk:
Thank you!

PR #4930, opened 2020-09-22
Implement anchor for TrueType fonts

### LINKED ISSUE #4511: RTL Direction not having any effect

### LINKED ISSUE #4553: Vertical text alignment ignores yOffset

nulano:
Second part of #4724, implementing anchor parameter to fix #4511.

 * Implement `anchor` parameter for text positioning.

   See proposed documentation for details: https://pillow--4930.org.readthedocs.build/en/4930/handbook/text-anchors.html

 * Add tests for #4553 and related issues (fixed in #4910). Closes #4553.

 * Remove the additional offset caused by stroking text.

   This was incompatible with the anchor implementation and I believe it to be a bug; I don't think users should need to offset text just to stroke a single word in the middle of a line.


Tests/test_imagefont.py
```diff
@@ -750,6 +753,93 @@ def test_variation_set_by_axes(self):
...
+            "ma",
+            "mt",
+            "mm",
+            "mb",
+            "md",
+        ),
+    )
+    def test_anchor(self, anchor):
+        name, text = "quick", "Quick"
+        path = "Tests/images/test_anchor_%s_%s.png" % (name, anchor)
```
hugovk, line 773:
```suggestion
        path = f"Tests/images/test_anchor_{name}_{anchor}.png"
```

Tests/test_imagefont.py
```diff
@@ -750,6 +753,93 @@ def test_variation_set_by_axes(self):
...
+            ("rm", "center"),
+            ("rm", "right"),
+            # test vertical anchors
+            ("ma", "center"),
+            # ("mm", "center"),  # duplicate
+            ("md", "center"),
+        ),
+    )
+    def test_anchor_multiline(self, anchor, align):
+        target = "Tests/images/test_anchor_multiline_%s_%s.png" % (anchor, align)
```
hugovk, line 807:
```suggestion
        target = f"Tests/images/test_anchor_multiline_{anchor}_{align}.png"
```

Tests/test_imagefontctl.py
```diff
@@ -206,3 +207,118 @@ def test_language():
...
+
+@pytest.mark.parametrize("anchor", ("lt", "mm", "rb", "sm"))
+def test_anchor_ttb(anchor):
+    if parse_version(features.version_module("freetype2")) < parse_version("2.5.1"):
+        # FreeType 2.5.1 README: Miscellaneous Changes:
+        # Improved computation of emulated vertical metrics for TrueType fonts.
+        pytest.skip("FreeType <2.5.1 has incompatible ttb metrics")
+
+    text = "f"
+    path = "Tests/images/test_anchor_ttb_%s_%s.png" % (text, anchor)
```
hugovk, line 220:
```suggestion
    path = f"Tests/images/test_anchor_ttb_{text}_{anchor}.png"
```

Tests/test_imagefontctl.py
```diff
@@ -206,3 +207,118 @@ def test_language():
...
+def test_combine(name, text, dir, anchor, epsilon):
+    if (
+        parse_version(features.version_module("freetype2")) < parse_version("2.5.1")
+        and dir == "ttb"
+    ):
+        # FreeType 2.5.1 README: Miscellaneous Changes:
+        # Improved computation of emulated vertical metrics for TrueType fonts.
+        pytest.skip("FreeType <2.5.1 has incompatible ttb metrics")
+
+    path = "Tests/images/test_combine_%s.png" % name
```
hugovk, line 284:
```suggestion
    path = f"Tests/images/test_combine_{name}.png"
```

hugovk (commented):
f-strings!

Tests/test_imagefont.py
nulano: Ah yes, this was started before 3.5 support was dropped.

docs/handbook/text-anchors.rst
```diff
@@ -0,0 +1,140 @@
+
+.. _text-anchors:
+
+Text anchors
+============
+
+The ``anchor`` parameter determines alignment of drawn text relative to the ``xy`` parameter.
+The default alignment is top left, specifically ``la`` (left--ascender) for horizontal text
+and ``lt`` (left--top) for vertical text.
```
hugovk, line 9: Nit: is there a reason for the double dash?

```suggestion
and ``lt`` (left--top) for vertical text.
```

docs/handbook/text-anchors.rst
```diff
@@ -0,0 +1,140 @@
...
+
+This parameter is only supported by OpenType/TrueType fonts.
+Other fonts may ignore the parameter and use the default (top left) alignment.
+
+Specifying an anchor
+^^^^^^^^^^^^^^^^^^^^
+
+An anchor is specified with a two character string. The first character is the
```
hugovk, line 17:
```suggestion
An anchor is specified with a two-character string. The first character is the
```

docs/handbook/text-anchors.rst
```diff
@@ -0,0 +1,140 @@
...
+  :align: center
+
+.. image:: ../resources/anchor_vertical.svg
+  :alt: Vertical text
+  :align: center
+
+Horizontal anchor alignment
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+``l`` --- left
```
hugovk, line 64: Nit: and for these triple dashes?

docs/handbook/text-anchors.rst
```diff
@@ -0,0 +1,140 @@
...
+Examples
+^^^^^^^^
+
+The following image shows several examples of anchors for horizontal text.
+In each section the ``xy`` parameter was set to the center shown by the intersection
+of the two lines.
+
+.. comment: Image generated with ../example/anchors.py
+
+.. image:: ../example/anchors.png
```
hugovk, line 135: Minor suggestion:

This image may be a bit easier to read if the "Sample" text was replaced with "anchor=ma" etc. That is, move "anchor=ma" etc. to the position where "Sample" is now.

Then there's no need to keep glancing back and forth between each pair.

hugovk (approved):

docs/handbook/text-anchors.rst
nulano: 1) A colon or shorter dash made the description too close and a little harder to read.
2) According to most style guides, the em-dash is an alternative to the colon, see https://en.wikipedia.org/wiki/Dash#Usage_2

(side note: `--` is an en-dash and `---` is an em-dash with Sphinx by default: https://www.sphinx-doc.org/en/master/usage/configuration.html?highlight=em-dash#confval-smartquotes_action)

docs/handbook/text-anchors.rst
nulano: I wanted to somehow make it clear that e.g. "left-ascender" means left horizontally and ascender vertically, and an en-dash is sometimes used in complex situations in place of a hyphen, but I think this is more confusing then helpful. I'll change these to hyphens, i.e. `-`.

docs/handbook/text-anchors.rst
nulano: Good suggestion, but "anchor=ma" doesn't contain any descenders, so "md" and "mb" would look identical. I changed it to "ma example" instead.

src/PIL/ImageDraw.py
```diff
@@ -373,6 +375,16 @@ def multiline_text(
         stroke_width=0,
         stroke_fill=None,
     ):
+        if direction == "ttb":
+            raise ValueError("ttb direction is unsupported for multiline text")
```
nulano, line 379: I didn't implement anchor handling for vertical text due to #4945, so instead this raises an error to avoid confusion due to unexpected anchor values.

hugovk (approved):

Issue #4900, opened 2020-09-03
Approve & Install Mergify


aclark4life:
OK give me your best Mergify pitch, @hugovk ! Thanks

<img width="928" alt="Screen Shot 2020-09-03 at 11 43 49 AM" src="https://user-images.githubusercontent.com/72164/92137266-f35f8b80-edda-11ea-85a2-13beb53e4c43.png">

hugovk:
Was just about to open an issue for this :)

Use case: there's a PR I want to merge, but the CIs are still running. Right now, I need to hang around for ~10 minutes before clicking merge (well, I _could_ merge before they finish, but it's nearly always best to make sure they pass).

Sometimes I'll just approve it now, and check back later.

With Mergify, I can add a label like `automerge`, and it'll automatically merge when the CIs pass. No more hanging around, let the robots do the work for us.

We can define other rules too, and I'd suggest: only merge a PR that has been explicitly approved.

aclark4life:
@wiredfool Any thoughts?

aclark4life:
@hugovk Just Pillow repo needed right?

hugovk:
It'd be of most use for the Pillow repo. And would be great to have docker-images and pillow-wheels too, they're updated less often but their CIs take [~30 mins](https://travis-ci.org/github/python-pillow/docker-images/builds/727648140) and [~60 mins](https://travis-ci.org/github/python-pillow/pillow-wheels/builds/726264804).

I don't need it for the other repos.

aclark4life:
Done!

<img width="1131" alt="Screen Shot 2020-09-17 at 12 12 40 PM" src="https://user-images.githubusercontent.com/72164/93498116-56731700-f8df-11ea-9607-c24cfe5aa7c8.png">

hugovk:
Thanks! I'll create the config files.

hugovk:
* [ ] https://github.com/python-pillow/docker-images/pull/84
* [ ] https://github.com/python-pillow/pillow-wheels
* [ ] https://github.com/python-pillow/Pillow

PR #4931, opened 2020-09-22
Add Mergify config

hugovk:
Fixes https://github.com/python-pillow/Pillow/issues/4900.

 * Approval, label and Travis CI rules like in Like https://github.com/python-pillow/docker-images/pull/84 and https://github.com/python-pillow/pillow-wheels/pull/161
 * AppVeyor rule is similar
 * GHA: https://doc.mergify.io/conditions.html#github-actions

> GitHub Actions works slightly differently. To match a status check when using GitHub Action, only the job name is used.
> 
> image
>
> In the example above, it would be `A job to say hello`:
>
```yml
conditions:
  - status-success=A job to say hello
```

So it's a bit more maintenance for GHA. We could rename the jobs so they don't include Python versions etc., but it's useful to see which one fails at a glance. Let's see how it goes and iterate if necessary.

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -4,47 +4,11 @@ pull_request_rules:
       - "#approved-reviews-by>=1"
       - label=automerge
       - status-success=Lint
+      - status-success=Test Successful
+      - status-success=Docker Test Successful
+      - status-success=Windows Test Successful
       - status-success=continuous-integration/appveyor/pr
       - status-success=continuous-integration/travis-ci/pr
-
-      # Test
-      - status-success=macOS-latest Python 3.6
-      - status-success=macOS-latest Python 3.7
-      - status-success=macOS-latest Python 3.8
-      - status-success=macOS-latest Python 3.9-dev
-      - status-success=macOS-latest Python pypy3
-      - status-success=ubuntu-latest Python 3.6
-      - status-success=ubuntu-latest Python 3.7
-      - status-success=ubuntu-latest Python 3.8
-      - status-success=ubuntu-latest Python 3.9-dev
-      - status-success=ubuntu-latest Python pypy3
-
-      # Test Docker
-      - status-success=alpine
-      - status-success=amazon-1-amd64
-      - status-success=amazon-2-amd64
-      - status-success=arch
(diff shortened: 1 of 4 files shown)


nulano:
It might be possible to use [dependant jobs in GHA](https://docs.github.com/en/actions/learn-github-actions/managing-complex-workflows#creating-dependent-jobs) to simplify this config a bit. Each workflow set (`test.yml`, `test-docker.yml`, `test-windows.yml`) can have another dummy job added that depends on the success of the rest of the set named something like `Test Windows / All Successful` which could allow this config to be just three GHA lines.

I'm not sure whether there is anything else that could be shown in these dummy jobs such as perhaps a warnings summary, but it might be worth adding just to simplify this config. OTOH the set of jobs doesn't change often, so this might be fine as is.

hugovk:
Interesting idea! 

Would they still need enumerating, something like this?

```yml
jobs:
  build:
    ...
  success:
    needs: [
      MSYS2 MinGW 32-bit,
      MSYS2 MinGW 64-bit,
      Python 3.6 x64,
      Python 3.6 x86,
      Python 3.7 x64,
      Python 3.7 x86,
      Python 3.8 x64,
      Python 3.8 x86,
      Python 3.9-dev x64,
      Python 3.9-dev x86,
      Python pypy3 x86,
    ]

```

Or would it be possible to use a matrix?

nulano:
It only needs to list `build`:
```yml
jobs:
  build:
    ...
  success:
    needs: build
    ...
```
Take a look at an example I made: [test.yml](https://github.com/nulano/Pillow/actions/runs/268268455/workflow#L112), [test-windows.yml](https://github.com/nulano/Pillow/actions/runs/268268456/workflow#L272)

hugovk:
That looks good. Please could you make a test PR against my fork? I've put this config there:

https://github.com/hugovk/Pillow/commit/5b6c6fc6337a34f985711573719fa3f41a486c49

hugovk:
The tests worked (https://github.com/hugovk/Pillow/pull/60, https://github.com/hugovk/Pillow/pull/61), updated!

PR #4934, opened 2020-09-23
Added overview to README and ReadTheDocs index

### LINKED ISSUE #4933: Main pages do not state what Pillow actually is, a short overview would be really useful
https://pillow.readthedocs.io/en/stable/ begins with 
"Pillow is the friendly PIL fork by Alex Clark and Contributors. PIL is the Python Imaging Library by Fredrik Lundh and Contributors.
Pillow for enterprise is available via the Tidelift Subscription. Learn more."

And then comes a big table of contents beginning with the installation. Finally the project has a good overview at https://pillow.readthedocs.io/en/stable/handbook/overview.html

The situation is similar at this repo's main `readme.md`, I need 3 clicks and know where to look to get an explanation about what this library does. I think this can be confusing especially for beginners

radarhere:
Resolves #4933

The issue reports that the [ReadTheDocs Pillow index page](https://pillow.readthedocs.io/en/latest/) and the [GitHub README](https://github.com/python-pillow/Pillow/blob/master/README.md) don't describe what Pillow is in much detail.

This PR copies the 'Overview' section from https://pillow.readthedocs.io/en/latest/handbook/overview.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,14 @@ As of 2019, Pillow development is
     </tr>
 </table>
 
+## Overview
+
+The Python Imaging Library adds image processing capabilities to your Python interpreter.
+
+This library provides extensive file format support, an efficient internal representation, and fairly powerful image processing capabilities.
+
+The core image library is designed for fast access to data stored in a few basic pixel formats. It should provide a solid foundation for a general image processing tool.
+
 ## More Information
 
 - [Documentation](https://pillow.readthedocs.io/)
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -57,6 +57,15 @@ Pillow for enterprise is available via the Tidelift Subscription. `Learn more <h
    :target: https://pypi.org/project/Pillow/
    :alt: Number of PyPI downloads
 
+Overview
+========
+
+The Python Imaging Library adds image processing capabilities to your Python interpreter.
+
+This library provides extensive file format support, an efficient internal representation, and fairly powerful image processing capabilities.
+
+The core image library is designed for fast access to data stored in a few basic pixel formats. It should provide a solid foundation for a general image processing tool.
+
 .. toctree::
    :maxdepth: 2
 


hugovk (approved):

docs/index.rst
```diff
@@ -57,6 +57,15 @@ Pillow for enterprise is available via the Tidelift Subscription. `Learn more <h
...
 
+Overview
+========
+
+The Python Imaging Library adds image processing capabilities to your Python interpreter.
+
+This library provides extensive file format support, an efficient internal representation, and fairly powerful image processing capabilities.
+
+The core image library is designed for fast access to data stored in a few basic pixel formats. It should provide a solid foundation for a general image processing tool.
+
```
matyasf, line 68: I'd move this up, right below the title. Here users need to scroll a lot to find it.

README.md
```diff
@@ -80,6 +80,14 @@ As of 2019, Pillow development is
...
 </table>
 
+## Overview
+
+The Python Imaging Library adds image processing capabilities to your Python interpreter.
+
+This library provides extensive file format support, an efficient internal representation, and fairly powerful image processing capabilities.
+
+The core image library is designed for fast access to data stored in a few basic pixel formats. It should provide a solid foundation for a general image processing tool.
+
```
matyasf, line 90: I'd move this up, right below the title. Here users need to scroll a lot to find it.

docs/index.rst
radarhere: If others have thoughts, feel free to disagree with me, but I don't see the need for this
- for ReadTheDocs, if you're on a desktop, you don't need to scroll to see this new text in its current position
- the current first line in both ReadTheDocs and the README includes 'Python Imaging Library'. I think those three words convey the idea of this repository fairly well. If a visitor is interested after that, they are interested enough to scroll further. If they're not, then I don't think the 'Overview' is going to convince them otherwise.

hugovk: Yep, I think this PR is an improvement. Thanks both!

Issue #4937, opened 2020-09-27
MSYS2 Access Violation


radarhere:
The 'MSYS2 MinGW 64-bit' GitHub Actions job has started crashing with 'Windows fatal exception: access violation' - https://github.com/python-pillow/Pillow/runs/1160314080?check_suite_focus=true

```
Tests/test_imagetk.py::test_kw PASSED                                    [ 92%]
Windows fatal exception: access violation

Current thread 0x0000069c (most recent call first):
  File "C:/msys64/mingw64/lib/python3.8/site-packages/pillow-8.0.0.dev0-py3.8-mingw.egg/PIL/ImageTk.py", line 197 in paste
  File "C:/msys64/mingw64/lib/python3.8/site-packages/pillow-8.0.0.dev0-py3.8-mingw.egg/PIL/ImageTk.py", line 115 in __init__
  File "D:/a/Pillow/Pillow/Tests/test_imagetk.py", line 61 in test_photoimage
  File "C:/msys64/mingw64/lib/python3.8/site-packages/_pytest/python.py", line 184 in pytest_pyfunc_call
  File "C:/msys64/mingw64/lib/python3.8/site-packages/pluggy/callers.py", line 187 in _multicall
  File "C:/msys64/mingw64/lib/python3.8/site-packages/pluggy/manager.py", line 84 in <lambda>
  File "C:/msys64/mingw64/lib/python3.8/site-packages/pluggy/manager.py", line 93 in _hookexec
  File "C:/msys64/mingw64/lib/python3.8/site-packages/pluggy/hooks.py", line 286 in __call__
  File "C:/msys64/mingw64/lib/python3.8/site-packages/_pytest/python.py", line 1479 in runtest
  File "C:/msys64/mingw64/lib/python3.8/site-packages/_pytest/runner.py", line 135 in pytest_runtest_call
  File "C:/msys64/mingw64/lib/python3.8/site-packages/pluggy/callers.py", line 187 in _multicall
  File "C:/msys64/mingw64/lib/python3.8/site-packages/pluggy/manager.py", line 84 in <lambda>
  File "C:/msys64/mingw64/lib/python3.8/site-packages/pluggy/manager.py", line 93 in _hookexec
  File "C:/msys64/mingw64/lib/python3.8/site-packages/pluggy/hooks.py", line 286 in __call__
  File "C:/msys64/mingw64/lib/python3.8/site-packages/_pytest/runner.py", line 217 in <lambda>
[...]

PR #4944, opened 2020-10-04
Skip test_photoimage on MinGW due to segmentation fault

hugovk:
For https://github.com/python-pillow/Pillow/issues/4937.

 * Temporarily skip the `test_photoimage` on MinGW, it otherwise causes a segmentation fault


radarhere:
Thanks

PR #4947, opened 2020-10-05
Stop flattening EXIF IFD into getexif()

### LINKED ISSUE #3973: exif_transpose errors out on some images

radarhere:
Resolves #3973 and #5273

At the moment, the `Exif` class [flattens the EXIF IFD into the rest of the Exif data](https://github.com/python-pillow/Pillow/blob/68c67403435f7a75e09c7bbd7385643a5eb82ff0/src/PIL/Image.py#L3335-L3337). Losing information in this way is not ideal.

Here is the new test added to demonstrate the difference.
```python
def test_exif_ifd(self):
    im = Image.open("Tests/images/flower.jpg")
    exif = im.getexif()

    reloaded_exif = Image.Exif()
    reloaded_exif.load(exif.tobytes())
    assert reloaded_exif.get_ifd(0x8769) == exif.get_ifd(0x8769)
```

To try and prevent this, while maintaining as much backwards compatibility as possible, this PR stops this flattening only for `getexif()`, while keeping it in place for the older `_getexif()`. So this is backwards incompatible, but only for code written after the release of 6.0.0.

Only one existing test had to be changed to match this breaking change.


src/PIL/Image.py
```diff
@@ -3346,91 +3351,100 @@ def tobytes(self, offset=8):
             head = b"MM\x00\x2A\x00\x00\x00\x08"
         ifd = TiffImagePlugin.ImageFileDirectory_v2(ifh=head)
         for tag, value in self.items():
+            if tag in [0x8769, 0x8225] and not isinstance(value, dict):
+                value = self.get_ifd(tag)
```
kkopachev, line 3355: This does not work for interop since it's inside of Exif IFD. 
```
im = Image.open('10_years_of_Wikipedia_by_Guillaume_Paumier.jpg')
good_exif = im.getexif()

bad_exif = Exif()
bad_exif.load(good_exif.tobytes())

print("Good interop dictionary size: %d" % len(good_exif.get_ifd(0xA005)))
print(" Bad interop dictionary size: %d" % len(bad_exif.get_ifd(0xA005)))

```
I suspect same for makernote too, but since makernote is vendor specific, I think it would be safer to make it read only property.

kkopachev:
Also consider that Exif might contain multiple IFDs one after another. Currently only first IFD is preserved on save.

src/PIL/Image.py
```diff
@@ -3346,91 +3351,100 @@ def tobytes(self, offset=8):
...
+                    self._ifds[tag] = self._get_ifd_dict(self[tag])
+            elif tag in [0xA005, 0x927C]:
+                # interop, makernote
+                if 0x8769 not in self._ifds:
+                    self.get_ifd(0x8769)
+                tag_data = self._ifds[0x8769][tag]
+                if tag == 0x927C:
+                    from .TiffImagePlugin import ImageFileDirectory_v2
+
+                    if self._ifds[0x8769][tag][:8] == b"FUJIFILM":
```
kkopachev, line 3373:
```suggestion
                    if tag_data[:8] == b"FUJIFILM":
```

src/PIL/Image.py
radarhere: Ok, I've added another commit to cover Interop.

radarhere:
> Also consider that Exif might contain multiple IFDs one after another. Currently only first IFD is preserved on save.

I'm not following. What do you mean, 'only first IFD'?

kkopachev:
I was referring to IFD1 - thumbnail image info which might be present in the file. Accessible through `self._info.next` after reading IFD0. 
Not sure how important it is to preserve it, thought we're here already and it might affect code design anyway.

src/PIL/Image.py
kkopachev: I've being thinking through this on and off for some time.
Thinking of consumer of this class, it would make sense when reading Exif IFD to get Interop value decoded into a dict at that time. Or shimmed with `ImageFileDirectory_v2` instance. Otherwise caller must know Inerop tag ID, must know to call `get_ifd`, etc. So that's a bit inconvenient and adds extra knowledge to consumer. 
Same for Makernote, however saving it back we don't want to encode it back to vendor-specific format and probably worth just copying whatever bytes were there originally.
On the other hand Exif is quite complicated and some knowledge from consumer is expected anyway.
:shrug: 
What do you think of this?

kkopachev: Another benefit of using `ImageFileDirectory_v2` as a shim for sub-IFDs is that we get lazy loading.

src/PIL/JpegImagePlugin.py
```diff
@@ -485,7 +485,7 @@ def _fixup_dict(src_dict):
 def _getexif(self):
     if "exif" not in self.info:
         return None
-    return dict(self.getexif())
+    return self.getexif()._get_merged_dict()
```
kkopachev, line 488: `def _fixup_dict(src_dict):` defined above seems unused used

src/PIL/Image.py
```diff
@@ -3346,91 +3351,107 @@ def tobytes(self, offset=8):
...
-                        else:
-                            data = data[:size]
-
-                        if len(data) != size:
-                            warnings.warn(
-                                "Possibly corrupt EXIF MakerNote data.  "
-                                f"Expecting to read {size} bytes but only got "
-                                f"{len(data)}. Skipping tag {ifd_tag}"
+        if tag not in self._ifds:
+            if tag in [0x8769, 0x8825]:
```
kkopachev, line 3368: I think `__getitem__` method should be updated as well to call `get_ifd` for Exif IFD. I can't comment on non-changed code, but currently it's line 3474 should be `if tag in [0x8769, 0x8825]:`

src/PIL/JpegImagePlugin.py
radarhere: Ok, I've suggested that change in a separate PR, #4957

src/PIL/Image.py
radarhere: While I can see that this would be more useful for most cases, if you wanted to actually read the IFD offset, you wouldn't be able to anymore (apart from looking directly into `_data` or `_info`)

radarhere:
> I was referring to IFD1 - thumbnail image info which might be present in the file. Accessible through `self._info.next` after reading IFD0.
> Not sure how important it is to preserve it, thought we're here already and it might affect code design anyway.

Couldn't that easily lead to a situation where the thumbnail image no longer matches the main image, if the main image is modified by Pillow, but the thumbnail image isn't?

src/PIL/Image.py
kkopachev: I see your point. 
I think getting offset of Exif IFD is useless by itself, unless you want manually parse IFD structure yourself, in which case you could do so with raw exif bytes.
Additionally, I think code should be consistent: either decode both GPS and Exif IFDs into dictionaries on read in `__getitem__`, or return offset in both cases.

kkopachev:
> Couldn't that easily lead to a situation where the thumbnail image no longer matches the main image, if the main image is modified by Pillow, but the thumbnail image isn't?

That's a great point. In that case it's better to ignore IFD1 for writing.

hugovk:
@kkopachev How does this PR look now?

kkopachev:
@hugovk Apart from [inconsistent treatment](https://github.com/python-pillow/Pillow/pull/4947#discussion_r503539024) of GPS IFD in `__getitem__`, it looks good :shipit:

radarhere:
Ok, I've added a commit to treat it consistently.

kkopachev (approved):

hugovk:
Thanks both!

PR #4955, opened 2020-10-07
Add support for CBDT and COLR fonts

nulano:
Proper fix for #891, partial support for #3346

Changes proposed in this pull request:

 * properly detect glyph bitmap mode instead of guessing
 * use `FT_Convert_Bitmap` instead of a custom solution
 * add support for Google CBDT (if FreeType is compiled with libpng) and Microsoft COLR fonts

Apple SBIX and W3C SVG fonts are not yet supported by FreeType.

Example:

```python
from PIL import Image, ImageDraw, ImageFont

def test(font, out_name):
    fnt = ImageFont.truetype(font, size=109, layout_engine=ImageFont.LAYOUT_RAQM)
    im = Image.new("RGBA", (600, 600), (100, 100, 100, 100))
    draw = ImageDraw.Draw(im)
    draw.text((0, 32), "a\u263A", fill="#faa2", embedded_color=True, font=fnt)
    draw.text((0, 132), "a\u263A", fill="#faa8", embedded_color=True, font=fnt)
    draw.text((0, 232), "a\u263A", fill="#faa", embedded_color=True, font=fnt)
    draw.text((0, 332), "\U0001F3AC\U0001F44B\U0001F3FB\U0001F44B\U0001F3FF", fill="white", embedded_color=True, font=fnt)
    draw.text((0, 432), "a\u263A", fill="#faa2", font=fnt)
    im.show()
    im.save(f"testemoji_{out_name}.png")

test(r"C:\Users\Nulano\Downloads\NotoColorEmoji.ttf", "cbdt")
test("seguiemj.ttf", "colr")
```

NotoColorEmoji.ttf (CBDT):
testemoji_cbdt

seguiemj.ttf (COLR):
testemoji_colr

PR #4964, opened 2020-10-12
Drop support for old PyPy versions

nulano:
Is it time to remove compatibility code for PyPy3<7.2.0 [released a year ago on 2019-10-14](https://morepypy.blogspot.com/2019/10/pypy-v72-released.html)? See also https://github.com/python-pillow/Pillow/pull/4145#discussion_r335667230.

 * Remove compatibility code in `_imagingft.c` for PyPy3<7.2.0.

 * Pin Travis to `pypy3.6-7.3.1`. The simple `pypy3` is currently pointing to the over-year-and-a-half-old 7.1.1-beta0 version, see https://docs.travis-ci.com/user/languages/python/#pypy-support. This version also needs the setuptools update.

   Travis does not yet support 7.3.2 released two weeks ago, see https://travis-ci.community/t/add-pypy-7-3-2-support/10007.

 * PyPy3.6-7.3.2 is now available on GHA and fixes the issue in `test_imagefont.test_unicode_extended` on Windows. Remove `pytest.mark.xfail`. This will also be needed for `test_cbdt[_mask]` tests from #4955. 

   I don't remember if this was needed on MacOS with PyPy 7.3.1 or older, but the wheels for the 8.0.0 release (#4764) should ideally (I'm assuming) target the latest 7.3.2 version.

I did not remove the distutils envars from https://github.com/python-pillow/Pillow/pull/4746#discussion_r453307511 which are no longer necessary as of 7.3.2. I think it makes sense to keep these until Python 3.6 support is dropped at which point users will have to upgrade PyPy anyway (PyPy 7.3.2 is the first to support Python 3.7).

### DIFF
--- a/.ci/install.sh
+++ b/.ci/install.sh
@@ -33,9 +33,10 @@ pip install pyroma
 pip install test-image-results
 pip install numpy
 
-# TODO Remove when 3.8 / 3.9 includes setuptools 49.3.2+:
+# TODO Remove when 3.8 / 3.9 / PyPy3 includes setuptools 49.3.2+:
 if [ "$GHA_PYTHON_VERSION" == "3.8" ]; then pip install -U "setuptools>=49.3.2" ; fi
 if [ "$GHA_PYTHON_VERSION" == "3.9" ]; then pip install -U "setuptools>=49.3.2" ; fi
+if [ "$TRAVIS_PYTHON_VERSION" == "pypy3.6-7.3.1" ]; then pip install -U "setuptools>=49.3.2" ; fi
 
 if [[ $TRAVIS_PYTHON_VERSION == 3.* ]]; then
   # arm64, ppc64le, s390x CPUs:
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ matrix:
     - python: "3.8"
       arch: s390x
 
-    - python: "pypy3"
+    - python: "pypy3.6-7.3.1"
       name: "PyPy3 Xenial"
     - python: "3.9-dev"
       name: "3.9-dev Xenial"
--- a/Tests/test_imagefont.py
+++ b/Tests/test_imagefont.py
@@ -15,7 +15,6 @@ from .helper import (
     assert_image_equal_tofile,
     assert_image_similar,
     assert_image_similar_tofile,
-    is_pypy,
     is_win32,
     skip_unless_feature,
 )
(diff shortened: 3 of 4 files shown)


radarhere:
A year doesn't seem very long when talking in the context of end-of-life. The more interesting thing to me is that PyPy 7.2.0 is the first to support 3.6 without the word 'beta' included as a warning.

> I don't remember if this was needed on MacOS with PyPy 7.3.1 or older, but the wheels for the 8.0.0 release (#4764) should ideally (I'm assuming) target the latest 7.3.2 version.

I would have thought the wheels should target the earliest support version, rather than the latest, to ensure compatibility. But the wheels are generated at pillow-wheels, not here, so I don't think that's relevant.

hugovk:
I asked PyPy on [Twitter](https://twitter.com/hugovk/status/1315565380878962690):

> @pypyproject CPython lists supported (3.6-3.9) and EOL versions:
>
> https://devguide.python.org/#status-of-python-branches
> https://devguide.python.org/devcycle/#end-of-life-branches
>
> Does PyPy have a similar concept? Or is it just the latest release, i.e. now only PyPy 7.3.2 (for Python 2.7, 3.6, and 3.7) is supported?
>
> https://morepypy.blogspot.com/2020/09/pypy-732-triple-release-python-27-36.html

Their [reply](https://twitter.com/pypyproject/status/1315571999515836417):

> It's the latter, we only really support the most current release.

---

I suggest in general we aim to support the latest release of PyPy?

(Excluding any Python versions that are EOL by CPython, i.e. Python 2.7.)

And as Pillow 8.0.0 is a major release, it's okay to remove old compat code?

radarhere (approved):

hugovk (approved):
Please also add to the release notes.

nulano:
> A year doesn't seem very long when talking in the context of end-of-life. The more interesting thing to me is that PyPy 7.2.0 is the first to support 3.6 without the word 'beta' included as a warning.

I seem to recall a comment from PyPy about a similar issue where they mentioned something along the lines of "... you can remove it after a year or so ..." and wanted to link to it later, but I'm having trouble finding it.

> > I don't remember if this was needed on MacOS with PyPy 7.3.1 or older, but the wheels for the 8.0.0 release (#4764) should ideally (I'm assuming) target the latest 7.3.2 version.
> 
> I would have thought the wheels should target the earliest support version, rather than the latest, to ensure compatibility. But the wheels are generated at pillow-wheels, not here, so I don't think that's relevant.

The relevant part is that I'm not sure if the tests can pass on 7.3.1 on MacOS, and that change is made here.

> I suggest in general we aim to support the latest release of PyPy?

Makes sense to me. Additionally, 7.1.x is now two major versions behind.

Issue #4960, opened 2020-10-10
Warning not obvious when installing on Windows for future Python version


jimmyxp01:
I'm trying to install pillow. But I'm getting this error. I didn't understand what to do. help me ,
### What are your OS, Python and Pillow versions?

* OS: Windows 10
* Python: 3.9
* Pillow: latest 

```python
Running setup.py install for Pillow ... error
    ERROR: Command errored out with exit status 1:
     command: 'C:\Users\Jitu\AppData\Local\Programs\Python\Python39\python.exe' -u -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'C:\\Users\\Jitu\\AppData\\Local\\Temp\\pip-install-q4hs2n1b\\pillow\\setup.py'"'"'; __file__='"'"'C:\\Users\\Jitu\\AppData\\Local\\Temp\\pip-install-q4hs2n1b\\pillow\\setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' install --record 'C:\Users\Jitu\AppData\Local\Temp\pip-record-xvh0leeq\install-record.txt' --single-version-externally-managed --compile --install-headers 'C:\Users\Jitu\AppData\Local\Programs\Python\Python39\Include\Pillow'
         cwd: C:\Users\Jitu\AppData\Local\Temp\pip-install-q4hs2n1b\pillow\
    Complete output (174 lines):
    C:\Users\Jitu\AppData\Local\Temp\pip-install-q4hs2n1b\pillow\setup.py:42: RuntimeWarning: Pillow 7.2.0 does not support Python 3.9 and does not provide prebuilt Windows binaries. We do not recommend building from source on Windows.
      warnings.warn(
    running install
    running build
    running build_py
    creating build
    creating build\lib.win-amd64-3.9
    creating build\lib.win-amd64-3.9\PIL
    copying src\PIL\BdfFontFile.py -> build\lib.win-amd64-3.9\PIL
    copying src\PIL\BlpImagePlugin.py -> build\lib.win-amd64-3.9\PIL
    copying src\PIL\BmpImagePlugin.py -> build\lib.win-amd64-3.9\PIL
    copying src\PIL\BufrStubImagePlugin.py -> build\lib.win-amd64-3.9\PIL
[...]

PR #4965, opened 2020-10-12
Moved warning to end of execution

radarhere:
Suggestion to help https://github.com/python-pillow/Pillow/issues/4960#issuecomment-706528502

To make the `RuntimeWarning` (that future versions of Python are not supported by a given Pillow release) more obvious, this PR raises it within the context of `atexit`, so that the warning is shown at the end of the output.


hugovk:
Let's give this a shot, thanks!

PR #4966, opened 2020-10-12
Removed OSError in favour of DecompressionBombError for BMP

### LINKED ISSUE #4939: [Feature Request] Making BMP pixel size limit configurable
Hello, I have been a happy user of Pillow for a long time, but new to the developer community. Thank you for making such amazing work available to the world. 

I am writing to ask if **the community would be open to the idea of making the hard-coded BMP [file size limit](https://github.com/python-pillow/Pillow/blob/5ce2fac3d6cfcf9af0f84502d9bd0c2dfcfcb398/src/PIL/BmpImagePlugin.py#L166) configurable by clients**. I need to work with large bmp files whose pixel counts exceed the limit, which does not appear configurable by design. Therefore, when I try opening the large bmp files using PIL, I get `OSError: Unsupported BMP Size`. It appears I need to modify the source code to lift the hard-coded limit.  

I understand, as documented, the motivation is to mitigate the risk with DOS attacks. However, I suppose there might be clients, just like me, that use PIL in an environment where there's a relatively limited DOS risk and are willing to accept the risk in any case.

If the idea makes sense, I'd love to work on a PR.

Look forward to your advice and thoughts.

FYI: the relevant code.

image

source:

https://github.com/python-pillow/Pillow/blob/5ce2fac3d6cfcf9af0f84502d9bd0c2dfcfcb398/src/PIL/BmpImagePlugin.py#L166

radarhere:
Resolves #4939

At the moment, a very large BMP file does not raise a DecompressionBombError, but instead an OSError. This PR changes it so that the more standard DecompressionBombError is raised.

### DIFF
--- a/Tests/test_decompression_bomb.py
+++ b/Tests/test_decompression_bomb.py
@@ -60,6 +60,10 @@ class TestDecompressionBomb:
         with pytest.raises(Image.DecompressionBombError):
             Image.open("Tests/images/decompression_bomb.gif")
 
+    def test_exception_bmp(self):
+        with pytest.raises(Image.DecompressionBombError):
+            Image.open("Tests/images/bmp/b/reallybig.bmp")
+
 
 class TestDecompressionCrop:
     @classmethod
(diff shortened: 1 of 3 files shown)


hugovk (approved):
Thanks! Let's also put this in the release notes.

PR #4968, opened 2020-10-12
Docs: Add dark mode

hugovk:
Dark mode is the new thing!

Not only that, it can improve accessibility, ease eye strain, and save battery power.

The `prefers-color-scheme` media query allows the theme to detect whether the user has requested a dark or light mode, often via an operating system toggle or time-of-day trigger.

* https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme
* https://web.dev/prefers-color-scheme/
* https://caniuse.com/?search=prefers-color-scheme

# Theme review

We're using the [Read the Docs Sphinx Theme](https://github.com/readthedocs/sphinx_rtd_theme), but it doesn't support dark theme:

* https://github.com/readthedocs/sphinx_rtd_theme/issues/224

Checking some other themes, they don't support dark mode yet:

* https://github.com/python/python-docs-theme/issues/43 (PR: https://github.com/python/python-docs-theme/pull/44)
* https://github.com/pradyunsg/furo/issues/13

PdJ Sphinx Theme is a specific dark theme, but is always dark and doesn't yet toggle via `prefers-color-scheme`: 

* https://github.com/jucacrispim/sphinx_pdj_theme/issues/5

# Alternative

As suggested at https://github.com/readthedocs/sphinx_rtd_theme/issues/224#issuecomment-703867448, the popular [Dark Reader browser extension](https://darkreader.org/) allows export of the dark mode styles it computes for a website.

The CSS here is exported from Dark Reader, wrapped in `@media (prefers-color-scheme: dark) { ... }` and added to our config.

# Demo

https://pillow--4968.org.readthedocs.build/en/4968/

I've checked it in Chrome, Firefox, Safari and Opera on Mac, and in Chrome and Samsung Internet on Android, not found any problems. 

Would be good to check it on Windows too.

### DIFF
--- a/docs/resources/css/dark.css
+++ b/docs/resources/css/dark.css
@@ -1988,4 +1988,9 @@
     .wy-body-for-nav .keep-us-sustainable a {
         color: rgb(222, 219, 215);
     }
+
+    /* For black-on-white/transparent images at handbook/text-anchors.html */
+    #text-anchors img {
+        filter: invert(1);
+    }
 }
--- a/docs/resources/css/dark.css
+++ b/docs/resources/css/dark.css
@@ -1991,6 +1991,6 @@
 
     /* For black-on-white/transparent images at handbook/text-anchors.html */
     #text-anchors img {
-        filter: invert(1) brightness(0.85);
+        filter: invert(1) brightness(0.85) hue-rotate(-60deg);
     }
 }
--- a/docs/resources/css/dark.css
+++ b/docs/resources/css/dark.css
@@ -1991,6 +1991,6 @@
 
     /* For black-on-white/transparent images at handbook/text-anchors.html */
     #text-anchors img {
-        filter: invert(1);
+        filter: invert(1) brightness(0.85);
     }
 }
--- a/docs/resources/css/dark.css
+++ b/docs/resources/css/dark.css
@@ -1,0 +1,1991 @@
+@media (prefers-color-scheme: dark) {
+    html {
+        background-color: #181a1b !important;
+    }
+
+    html, body, input, textarea, select, button {
+        background-color: #181a1b;
+    }
+
+    html, body, input, textarea, select, button {
+        border-color: #736b5e;
+        color: #e8e6e3;
+    }
+
+    a {
+        color: #3391ff;
+    }
+
+    table {
+        border-color: #545b5e;
+    }
+
+    ::placeholder {
+        color: #b2aba1;
+    }
+
+    input:-webkit-autofill,
+    textarea:-webkit-autofill,
+    select:-webkit-autofill {
+        background-color: #555b00 !important;
+        color: #e8e6e3 !important;
+    }
+
+    ::selection {
+        background-color: #004daa !important;
+        color: #e8e6e3 !important;
+    }
+
+    ::-moz-selection {
+        background-color: #004daa !important;
+        color: #e8e6e3 !important;
+    }
+
+    /* Invert Style */
+    .jfk-bubble.gtx-bubble, embed[type="application/pdf"] {
+        filter: invert(100%) hue-rotate(180deg) contrast(90%) !important;
+    }
+
+    /* Override Style */
+    .vimvixen-hint {
+        background-color: #7b5300 !important;
+        border-color: #d8b013 !important;
+        color: #f3e8c8 !important;
+    }
+
+    ::placeholder {
+        opacity: 0.5 !important;
+    }
+
+    /* Variables Style */
+    :root {
+        --darkreader-neutral-background: #181a1b;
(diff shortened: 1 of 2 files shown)


pradyunsg:
Godot has a really nice theme built on top of RTD-Sphinx-Theme, but someone needs to factor it out to be reusable -- https://github.com/godotengine/godot-docs/issues/3873.

hugovk:
Thanks for testing!


> Internet Explorer 11 (has no effect), 

Yeah, IE11 doesn't support it: https://caniuse.com/?search=prefers-color-scheme

> Is it possible to use a different image based on the selected theme? I'm thinking something like adding both a bright theme image and a dark theme image, adding a class to the dark image, and using css to hide one of the images.

Yes, that's one way. But will both images be loaded by the browser? Alternatively, figure out how to do this in Sphinx: https://stackoverflow.com/a/56030447/724176

But from a maintenance point of view, it'll be easier to keep one set of images.

The easiest thing is just to stick a white background in the SVGs themselves.

Or add a background to `dark.css` that's lighter-than-black and darker-than-white. Let's try this, any suggestions for a colour?

hugovk:
> Godot has a really nice theme built on top of RTD-Sphinx-Theme, but someone needs to factor it out to be reusable -- [godotengine/godot-docs#3873](https://github.com/godotengine/godot-docs/issues/3873).

Looks good!

Would be nice to get something "good enough" out as `/stable` in this week's release, then replace it with something else when there's something else that's ready :)

nulano:
> The easiest thing is just to stick a white background in the SVGs themselves.
> 
> Or add a background to `dark.css` that's lighter-than-black and darker-than-white. Let's try this, any suggestions for a colour?

I've played with it a bit, and can't find a color where both the page and image have readable text.

However, I have found that adding a `style="filter: invert"` looks good just by itself.
It seems to be supported [almost everywhere](https://caniuse.com/css-filters) (except for IE11).

<details><summary>Image</summary>

image

</details>

hugovk:
How does https://github.com/python-pillow/Pillow/pull/4968/commits/461d9362270c0aaac7715d10c399b35d4bb1b63a look? In addition to the two black-on-transparent SVGs, it inverts the two black-on-white PNGs:

https://pillow--4968.org.readthedocs.build/en/4968/handbook/text-anchors.html#specifying-an-anchor

<details>
<summary>Details</summary>

image

image

image

</details>

docs/resources/css/dark.css
```diff
@@ -0,0 +1,1996 @@
...
+        color: rgb(184, 178, 169);
+    }
+
+    .wy-body-for-nav .keep-us-sustainable a {
+        color: rgb(222, 219, 215);
+    }
+
+    /* For black-on-white/transparent images at handbook/text-anchors.html */
+    #text-anchors img {
+        filter: invert(1);
```
nulano, line 1994: I think it's a bit too bright, certainly more than the rest of the page. I think the following is a good balance:
```suggestion
        filter: invert(1) brightness(0.85);
```

docs/resources/css/dark.css
```diff
@@ -0,0 +1,1996 @@
...
+        color: rgb(184, 178, 169);
+    }
+
+    .wy-body-for-nav .keep-us-sustainable a {
+        color: rgb(222, 219, 215);
+    }
+
+    /* For black-on-white/transparent images at handbook/text-anchors.html */
+    #text-anchors img {
+        filter: invert(1) brightness(0.85);
```
nulano, line 1994: Also adding a -60° hue shift gives an easier to read colour scheme (the other two images are black-and-white and unaffected) that more closely matches the original.
```suggestion
        filter: invert(1) brightness(0.85) hue-rotate(-60deg);
```
<details>

image

</details>

nulano (approved):
There is also an image in 6.2.0 release notes, but it already uses a dark colour scheme. I can't see any other images in the docs.

image

hugovk:
Let's go with this! We can review other themes as and when they get dark mode.

pradyunsg:
FYI -- Furo's got a nice dark mode now. Just saying. ;)

hugovk:
Thanks! Here's a demo (just rebuilt with latest Furo 2020.10.15b13):

* https://pillow-hugovk.readthedocs.io/en/furo/

I gave it a quick test before, and there was something a bit strange on mobile with the advert always showing at the bottom of the page rather than hidden in the left menu. Of course, I didn't take a screenshot and cannot reproduce now (no ad being served), but will report to the tracker if I see it again!

pradyunsg:
Yes please! I'd also appreciate if you could put down the html for that page somewhere, coz it's dependent on when RTD decides to serve an ad.

PR #4975, opened 2020-10-13
Avoid using pkg_resources in PIL.features.pilinfo

nulano:
Instead of using `pkg_resources`, which AFAICT is [part of `setuptools`](https://setuptools.readthedocs.io/en/latest/pkg_resources.html#parsing-utilities), check the LCMS version (check added in #4963) manually.

While `setuptools` is very likely to already be installed, Pillow doesn't actually require it for runtime, only compilation and testing. Additionally, the `pilinfo` function is useful for finding configuration issues and should have as few external dependencies as possible.

Since this version string is being set in the C code, it is safe to break it back into ints and perform a tuple comparison instead.
This PR also adds a comment to make sure the two checks don't get out of sync.

### DIFF
--- a/src/PIL/features.py
+++ b/src/PIL/features.py
@@ -224,8 +224,6 @@ def pilinfo(out=None, supported_formats=True):
         If ``True``, a list of all supported image file formats will be printed.
     """
 
-    from pkg_resources import parse_version
-
     if out is None:
         out = sys.stdout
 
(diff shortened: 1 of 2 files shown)


hugovk:
Yep, `pkg_resources` is part of setuptools, and is likely to be installed (it's used by `setup.py`), some users of flit (or poetry?) might not have setuptools.

PR #4924, opened 2020-09-16
chown pip wheels cache in Travis

radarhere:
The TravisCI Python 3.8 s390x job has started failing - https://travis-ci.org/github/python-pillow/Pillow/jobs/727457309

```
Building wheels for collected packages: black, regex, typed-ast, MarkupSafe
  WARNING: Building wheel for black failed: [Errno 13] Permission denied: '/home/travis/.cache/pip/wheels/95'
  WARNING: Building wheel for regex failed: [Errno 13] Permission denied: '/home/travis/.cache/pip/wheels/c1/d2'
  WARNING: Building wheel for typed-ast failed: [Errno 13] Permission denied: '/home/travis/.cache/pip/wheels/90'
  WARNING: Building wheel for MarkupSafe failed: [Errno 13] Permission denied: '/home/travis/.cache/pip/wheels/0c'
Failed to build black regex typed-ast MarkupSafe
ERROR: Could not build wheels for black which use PEP 517 and cannot be installed directly
```

This PR fixes it by using `sudo chown $USER ~/.cache/pip/wheels`. This fix was suggested at https://travis-ci.community/t/permission-issue-while-building-wheels-for-various-python-packages/7822/12

PR #4978, opened 2020-10-14
chown pip wheels cache on ppc64le

radarhere:
master has started failing on [the ppc64le job](https://travis-ci.org/github/python-pillow/Pillow/jobs/735677534). This fixes it, as with #4924 for s390x.


hugovk (approved):
Thanks!

This is https://github.com/RonnyPfannschmidt/iniconfig/issues/16 upstream, will mention this workaround there too.

PR #4980, opened 2020-10-14
GHA: Add 3.10-dev

nulano:
GHA have added support for Python 3.10.0-a1, add it to the matrix:

 * #3749 missed `display.c` which caused two new failures on Windows: `test_file_wmf.py` and `test_imagewin.py`
 * Due to https://github.com/pypa/wheel/issues/354, temporarily disable installing numpy and building Windows wheels on Python 3.10.


src/display.c
```diff
@@ -777,7 +777,7 @@ PyImaging_DrawWmf(PyObject* self, PyObject* args)
     char* ptr;
 
     char* data;
-    int datasize;
+    Py_ssize_t datasize;
```
radarhere, line 780: I find this change is unnecessary? I've created https://github.com/nulano/Pillow/pull/11 to demonstrate.

radarhere: See reply at https://github.com/nulano/Pillow/pull/11#issuecomment-720086493

radarhere:
Regarding numpy, https://github.com/numpy/numpy/pull/18389#issuecomment-819640687
> 1.21.x will be the first to support Python 3.10

PR #4985, opened 2020-10-16
Remove soon-EOL Fedora 31

hugovk:
Fedora 31 will be EOL on 2020-12-02, before the next scheduled Pillow 8.1.0 on 2020-01-02.

* https://fedorapeople.org/groups/schedule/f-31/f-31-all-tasks.html

See also https://github.com/python-pillow/docker-images/pull/88.


radarhere:
Thanks

PR #4986, opened 2020-10-17
Revert "chown pip wheels cache on ppc64le as well"

radarhere:
Reverts #4978 - Travis now passes without it


hugovk:
Thanks! It'll be because iniconfig has a wheel now: https://github.com/RonnyPfannschmidt/iniconfig/issues/17#issuecomment-710288144

PR #4988, opened 2020-10-17
Fix internal Sphinx link to installation doc

jdufresne:
Fixes linkcheck warnings:

    handbook/image-file-formats.rst:563: [broken] ../installation.html:
    releasenotes/4.2.0.rst:7: [broken] ../installation.html:

### DIFF
--- a/docs/handbook/image-file-formats.rst
+++ b/docs/handbook/image-file-formats.rst
@@ -561,8 +561,8 @@ The :py:meth:`~PIL.Image.Image.save` method supports the following options:
 .. note::
 
     To enable PNG support, you need to build and install the ZLIB compression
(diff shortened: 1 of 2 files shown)


hugovk:
Thanks, why does it says it's broken?

The links work before:

* https://pillow.readthedocs.io/en/stable/releasenotes/4.2.0.html#added-complex-text-rendering

And after:

* https://pillow--4988.org.readthedocs.build/en/4988/releasenotes/4.2.0.html#added-complex-text-rendering

jdufresne:
I think because the link is relative. If `linkcheck` runs before making the HTML, the `.html` document doesn't exist yet.

In the end, I think it is more robust to let Sphinx handle the linking anyway.

PR #4989, opened 2020-10-17
De-zigzag JPEG's DQT when loading; deprecate convert_dict_qtables

### LINKED ISSUE #4962: Copying a JPEG quantization table as a list breaks the value order
### What did you do?

```python
from PIL import Image

im = Image.open('any_existing.jpg')
im.save('output.jpg', qtables=[im.quantization[0]])
im2 = Image.open('output.jpg')
assert im2.quantization[0] == im.quantization[0]
```

The assert passes.

### What actually happened?

It fails.

If you replace the second line by one of the following, it does pass:

```python
im.save('output.jpg', qtables=im.quantization)
im.save('output.jpg', qtables={0: im.quantization[0]})
```

The reason is that JPEGs store their quantization table in zigzag order. When reading JPEGs, [Pillow loads the table into the `quantization` property unchanged](https://github.com/python-pillow/Pillow/blob/804c76ff126496011797cd93cc05fe4461023cac/src/PIL/JpegImagePlugin.py#L235) which means the values are in zigzag order. 

But when _writing_, Pillow passes the `qtables` value to libjpeg's `jpeg_add_quant_table`, which takes values in normal order (since libjpeg version 6a).

Pillow has the built-in [`convert_dict_qtables`](https://github.com/python-pillow/Pillow/blob/804c76ff126496011797cd93cc05fe4461023cac/src/PIL/JpegImagePlugin.py#L586) to convert this dict of zigzagged arrays stored in `quantization` back into normal-order arrays. It uses that internally when _saving_ using a dict value, so if you use `qtables='keep'`, or `qtables=im.quantization` things work as expected.

But if you write the quantization data from an existing JPEG as a list, `jpeg_add_quant_table` ends up re-zigzagging the already zigzagged data.

I think it would be less confusing if Pillow already de-zigzagged the values when it reads them in the `DQT` function.

Note also that the existing `test_qtables` test is already affected by this:

https://github.com/python-pillow/Pillow/blob/804c76ff126496011797cd93cc05fe4461023cac/Tests/test_file_jpeg.py#L460

But the only thing that line really does is check that `save` doesn't raise any exceptions. It does not check that the input and output images are the same.

gofr:
Re-order the JPEG quantization tables to normal order when loading. This wastes a few CPU cycles if you don't need them.
But it has the advantage of hiding the zigzag order JPEG implementation detail of these tables completely from Pillow users.

This difference has led to cases where:
* arrays in zigzag order were taken from a dict and passed directly as a qtables parameter, causing them to be "zigzagged"
  again by libjpeg.
* dicts with lists in normal order being passed to JpegImagePlugin.convert_dict_qtables, causing them to be unnecessarily "de-zigzagged".

Fixes #4962

 * De-zigzag JPEG quantization tables when reading from file, causing all internal formats to consistently represent them in normal array order.
 * Make JpegImagePlugin.convert_dict_qtables a deprecated no-op.


src/PIL/JpegImagePlugin.py
```diff
@@ -585,9 +585,11 @@ def _getmp(self):
...
 
 def convert_dict_qtables(qtables):
-    qtables = [qtables[key] for key in range(len(qtables)) if key in qtables]
-    for idx, table in enumerate(qtables):
-        qtables[idx] = [table[i] for i in zigzag_index]
+    warnings.warn(
+        "convert_dict_qtables is deprecated and will be removed in a future"
+        " release. Conversion is no longer needed.",
+        DeprecationWarning,
+    )
```
radarhere, line 592: A thought (feel free to disagree) - if we're considering this to be a public method that requires deprecation, then should we maybe just leave the functionality there (still adding in the deprecation warning)?

gofr: I searched GitHub for this function: https://github.com/search?q=convert_dict_qtables+language%3Apython+-filename%3AJpegPresets.py. There are not many uses. Only the first two pages contain interesting results. The rest is just copies of Pillow itself. They all take the `.quantization` of an existing image, convert it to a nested list with this function and then manipulate that nested list in some way.

If Pillow de-zigzags the tables when loading (as the rest of this PR does), `.quantization` no longer needs any conversion step. That means those existing uses of the `convert_dict_qtables` function will actually break their code. The order of the numbers in the tables will be wrong. They'd need to remove the function call to fix their code. Turning `convert_dict_qtables` into a no-op means their code will continue to work unchanged. They'll only need to remove the function call to get rid of the warning.

Most people don't need to mess with quantization tables. But there's obviously more code than just what I found on GitHub and I don't know how representative these uses are. I guess there could also be cases where people are getting qtables from somewhere _other_ than straight from an image's `.quantization`. Maybe they serialized them for later re-use? In that case they no longer have a helper function that converts them back.

I guess Pillow could temporarily add a new `really_convert_dict_qtables` for those who still need it. :wink:

hugovk: @radarhere What do you think? Shall we merge this PR as is?

(And can explicitly replace "a future version" with "Pillow 10 (2023-01-02)")

Tests/test_file_jpeg.py
```diff
@@ -458,7 +458,8 @@ def _n_qtables_helper(n, test_file):
 
             # valid bounds for baseline qtable
             bounds_qtable = [int(s) for s in ("255 1 " * 32).split(None)]
-            self.roundtrip(im, qtables=[bounds_qtable])
+            im2 = self.roundtrip(im, qtables=[bounds_qtable])
+            assert im2.quantization[0] == bounds_qtable
```
radarhere, line 462:
```suggestion
            assert im2.quantization == {0: bounds_qtable}
```
Just increasing the specificity.

src/PIL/JpegImagePlugin.py
```diff
@@ -585,9 +585,11 @@ def _getmp(self):
 
 
 def convert_dict_qtables(qtables):
-    qtables = [qtables[key] for key in range(len(qtables)) if key in qtables]
-    for idx, table in enumerate(qtables):
-        qtables[idx] = [table[i] for i in zigzag_index]
+    warnings.warn(
+        "convert_dict_qtables is deprecated and will be removed in a future"
+        " release. Conversion is no longer needed.",
```
radarhere, line 590:
```suggestion
        "convert_dict_qtables is deprecated and will be removed in Pillow 10"
        "(2023-01-02). Conversion is no longer needed.",
```

src/PIL/JpegImagePlugin.py
radarhere: Sure - I can see now why `convert_dict_qtables` should do nothing, so no objections from me.

hugovk (approved):

hugovk:
Let's also add it to https://pillow.readthedocs.io/en/stable/deprecations.html

hugovk:
Thanks also for fixing the order of the other deprecations in the docs!

PR #4990, opened 2020-10-18
Added --remove-dist-egg option

### LINKED ISSUE #4981: Release process creates a dist/*.egg file
### What did you do?

Ran `make release-test` as part of the 8.0.0 release checks.

Release tests ran. Nothing created in `dist/`

### What actually happened?

Release tests ran. `dist/Pillow-8.1.0.dev0-py3.9-macosx-10.9-x86_64.egg` was created.

* OS: macOS Mojave
* Python: 3.9
* Pillow: 8.0.0

This comes from the `python3 setup.py install` command. I moved the file away before proceeding, to make sure it wasn't uploaded to PyPI as there wasn't one for the last release:

* https://pypi.org/project/Pillow/7.2.0/#files

But now I look again, these have a single egg:

| Release | egg |
|-|-|
| https://pypi.org/project/Pillow/7.1.2/#files | Pillow-7.1.2-py3.8-macosx-10.9-x86_64.egg |
| https://pypi.org/project/Pillow/7.1.1/#files | Pillow-7.1.1-py3.8-macosx-10.9-x86_64.egg |
| https://pypi.org/project/Pillow/7.1.0/#files | Pillow-7.1.0-py3.8-macosx-10.9-x86_64.egg |  

We don't need to release a single egg file for a single Python version, right? When we have all the wheels?

If we don't need it, can we prevent `python3 setup.py install` creating it, or alternatively, update `RELEASING.md` to remove it?

radarhere:
Resolves #4981

Adds a ` --remove-dist-egg` flag to setup.py. It removes any created `dist/*.egg` file. This is also added to `make release-test`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -90,7 +90,7 @@ release-test:
 	python3 setup.py develop
 	python3 selftest.py
 	python3 -m pytest Tests
-	python3 setup.py install
+	python3 setup.py install --remove-dist-egg
 	python3 -m pytest -qq
 	check-manifest
 	pyroma .
--- a/setup.py
+++ b/setup.py
@@ -857,8 +857,12 @@ ext_modules = [
 with open("README.md") as f:
     long_description = f.read()
 
+remove_dist_egg = "--remove-dist-egg" in sys.argv
+if remove_dist_egg:
+    while "--remove-dist-egg" in sys.argv:
+        sys.argv.remove("--remove-dist-egg")
 try:
-    setup(
+    distribution = setup(
         name=NAME,
         version=PILLOW_VERSION,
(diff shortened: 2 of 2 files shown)


hugovk:
To try and keep the monster `setup.py` a bit leaner, do you think `rm dist/*.egg` in `make release-test` would be enough?

Prefixing like `-rm dist/*.egg` tells make to ignore any errors (e.g. file not found): https://stackoverflow.com/a/2670143/724176

radarhere:
My thinking was to avoid removing any other eggs that might also be in the 'dist' directory - but I realise that's not really a situation that's likely to occur.

Ok, see #4992

PR #5000, opened 2020-10-22
Add release notes for 8.0.1

hugovk:
For https://github.com/python-pillow/Pillow/issues/4764.

 * Release notes for the FreeType security update.

### DIFF
--- a/docs/releasenotes/8.0.1.rst
+++ b/docs/releasenotes/8.0.1.rst
@@ -1,0 +1,23 @@
+8.0.1
+-----
+
+Security
+========
+
+Update FreeType used in binary wheels to `2.10.4`_ to fix CVE-2020-15999_:
+
+  - A heap buffer overflow has been found  in the handling of embedded PNG bitmaps,
+    introduced in FreeType version 2.6.
+
+      https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-15999
+
+    If you use option ``FT_CONFIG_OPTION_USE_PNG`` you should upgrade immediately.
+
+Before Pillow 8.0.0 bitmap fonts were disabled with ``FT_LOAD_NO_BITMAP``, but it is not
+clear if this prevents the exploit and we recommend updating to Pillow 8.0.1.
+
+Pillow 8.0.0 and earlier are potentially vulnerable releases, including the last release
+to support Python 2.7, namely Pillow 6.2.2.
+
+.. _2.10.4: https://sourceforge.net/projects/freetype/files/freetype2/2.10.4/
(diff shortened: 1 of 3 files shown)


docs/releasenotes/8.0.1.rst
```diff
@@ -0,0 +1,23 @@
...
+    introduced in FreeType version 2.6.
+
+      https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-15999
+
+    If you use option ``FT_CONFIG_OPTION_USE_PNG`` you should upgrade immediately.
+
+Before Pillow 8.0.0 bitmap fonts were disabled with ``FT_LOAD_NO_BITMAP``, but it is not
+clear if this prevents the exploit and we recommend updating to Pillow 8.0.1.
+
+Pillow 8.0.0 and earlier are potentially vulnerable releases, including the last release
```
radarhere, line 19: Technically speaking, it is only Pillow 2.9.0 onwards that are potentially vulnerable. Freetype 2.6 wasn't available before that.

radarhere (approved):

docs/releasenotes/8.0.1.rst
nulano: Also, `FT_LOAD_NO_BITMAP` was added in Pillow 2.8.0, before the release of FreeType 2.6, in #1072.

docs/releasenotes/8.0.1.rst
```diff
@@ -0,0 +1,23 @@
...
+
+  - A heap buffer overflow has been found  in the handling of embedded PNG bitmaps,
+    introduced in FreeType version 2.6.
+
+      https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-15999
+
+    If you use option ``FT_CONFIG_OPTION_USE_PNG`` you should upgrade immediately.
+
+Before Pillow 8.0.0 bitmap fonts were disabled with ``FT_LOAD_NO_BITMAP``, but it is not
+clear if this prevents the exploit and we recommend updating to Pillow 8.0.1.
```
nulano, line 17: "Before Pillow 8.0.0" sounds a bit confusing to me.

I think it would be best to split this into two parts, one stating that 8.0.0 is vulnerable, and one stating that <=7.2.0 might also be.

How is the following formulation?
```suggestion
We strongly recommend updating to Pillow 8.0.1 if you are using Pillow 8.0.0, which improved support for bitmap fonts.

In Pillow 7.2.0 and earlier bitmap fonts were disabled with ``FT_LOAD_NO_BITMAP``, but it is not
clear if this prevents the exploit and we recommend updating to Pillow 8.0.1.
```

PR #4993, opened 2020-10-19
Moved string_dimension image to pillow-depends

radarhere:
Resolves #4970 by moving the image flagged by antivirus to pillow-depends, as #4929 did.

The other half of this is https://github.com/python-pillow/pillow-depends/pull/34

PR #5001, opened 2020-10-22
CI: Replace test images from pillow-depends

nulano:
For https://github.com/python-pillow/Pillow/issues/4764#issuecomment-714547692:

> [GHA Window CI](https://github.com/python-pillow/Pillow/actions/runs/322256390) failed with "Overwrite D:\a\Pillow\Pillow\Tests\images\string_dimension.tiff (Yes/No/All)?", will be because of #4993 / [python-pillow/pillow-depends#34](https://github.com/python-pillow/pillow-depends/pull/34), so will include that too: [f886bc9](https://github.com/python-pillow/Pillow/commit/f886bc973bb972a796756d9c0a728795d203dd97).

Do not prompt to overwrite when copying test images from pillow-depends on Windows CIs so the above will not be needed for future point releases or when working on an old branch.

Successful test from my repo (does not include #4993): https://github.com/nulano/Pillow/runs/1293222859?check_suite_focus=true#step:10:32

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -22,7 +22,7 @@ install:
 - curl -fsSL -o pillow-depends.zip https://github.com/python-pillow/pillow-depends/archive/master.zip
 - 7z x pillow-depends.zip -oc:\
 - mv c:\pillow-depends-master c:\pillow-depends
-- xcopy /s c:\pillow-depends\test_images\* c:\pillow\tests\images
+- xcopy /S /Y c:\pillow-depends\test_images\* c:\pillow\tests\images
 - 7z x ..\pillow-depends\nasm-2.14.02-win64.zip -oc:\
 - ..\pillow-depends\gs9533w32.exe /S
 - path c:\nasm-2.14.02;C:\Program Files (x86)\gs\gs9.53.3\bin;%PATH%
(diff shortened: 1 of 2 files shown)


hugovk:
Thanks for the PR! I'm going to cancel the Travis build for now to clear the way for the release, because Travis is again giving us less than 5 jobs, it's been 2 and is actually zero right now... We'll restart it after the release is out!

nulano:
> Thanks for the PR! I'm going to cancel the Travis build for now to clear the way for the release, because Travis is again giving us less than 5 jobs, it's been 2 and is actually zero right now... We'll restart it after the release is out!

I think Travis should give preference to the earlier build, but it can't hurt to run this later 👍 .

https://www.traviscistatus.com#week shows less than 600 running jobs with over 4000 in the queue...

hugovk:
Yeah, as soon as that first one ends the version bump will be pushed as a branch _and_ tag build (will have to cancel those), and _then_ the big wheels build...

I'm actually also running the first one on my fork https://travis-ci.com/github/hugovk/Pillow/builds/191711135 and once that passes, I'll cancel the first one too.

I've already migrated my builds from https://travis-ci.org to https://travis-ci.com, and for some reason they give 7 parallel jobs there! We'll have to migrate the main builds at some point, but the full history is not yet being migrated: https://travis-ci.community/t/question-regarding-build-history-after-org-com-migration/7906?u=hugovk

nulano:
I wonder if it would be worth migrating at least the x86 builds to GHA for 20 parallel jobs, and so far no backlog that I have seen. Looking at the multibuild scripts I do not see any obvious problems. I might give it a try later this week if I have time.

hugovk:
It's probably worth moving it all over, but any increment would help!

hugovk:
Thanks!

Issue #5008, opened 2020-10-26
PGM and PPM maximum grey/color value is ignored


gofr:
### What did you do?

```python
from PIL import Image


header = b"""\
P5
15 7
11
"""
data = [
    0,  0,  0,  0,  0,  0,  0,  0,  0,  0,   0,  0,  0,  0,  0,
    0,  3,  3,  3,  3,  0,  7,  7,  7,  0,  11,  0,  0,  0,  0,
    0,  3,  0,  0,  3,  0,  0,  7,  0,  0,  11,  0,  0,  0,  0,
    0,  3,  3,  3,  3,  0,  0,  7,  0,  0,  11,  0,  0,  0,  0,
    0,  3,  0,  0,  0,  0,  0,  7,  0,  0,  11,  0,  0,  0,  0,
    0,  3,  0,  0,  0,  0,  7,  7,  7,  0,  11, 11, 11, 11,  0,
    0,  0,  0,  0,  0,  0,  0,  0,  0,  0,   0,  0,  0,  0,  0
]
with open('pil.pgm', 'wb') as fp:
    fp.write(header)
    fp.write(bytes(data))

img = Image.open('pil.pgm')
img.show()
img.save('pil.png')
```

### What did you expect to happen?

Both the image preview and saved PNG show a 15x7 greyscale image with the letters "PIL", where the "P" is dark grey, the "I" light grey and the "L" is white. This is what I see if I load the "pil.pgm" in other software, such as Ubuntu's file manager, image viewer, ImageMagick and GIMP.

### What actually happened?

The PIL letters are all really dark grey.

Pillow only uses the "11" [maxval](http://netpbm.sourceforge.net/doc/pgm.html) to determine how many bytes to use for each pixel. E.g. a value of `10000` makes it use 2 bytes per pixel. It does not use it as the maximum value. It uses 255 as the maximum instead.

Similarly, with 16-bit grey values, the maximum is always 65535.
The same happens for the [RGB P6 format](http://netpbm.sourceforge.net/doc/ppm.html). The maxval is only used to determine the bit depth.

### What are your OS, Python and Pillow versions?

* OS: Ubuntu 20.04
* Python: 3.8.5
* Pillow: built from 8dfc0364fef32eaa11fcbf4abe765c072840e27e

radarhere:
I've created PR #6119 to help this. It adds support for arbitrary maxval for P6, and maxval less than 255 for P5 - that fixes the specific code example you've provided.

Issue #4999, opened 2020-10-20
P3 ppm file format support (the non binary version of ppm that is very common)


wegfawefgawefg:
**What did you do?**
try to open the non binary versions of ppm files p1,p2,p3

**What did you expect to happen?**
to load the file

**What actually happened?**
it errors. you cant load p1,p2,p3 with pillow because pillow does not support them.

There are a lot of ppm p3 format files on the internet
Many graphics programming tutorials work with the non binary version.

I have found others complaining about this.
https://stackoverflow.com/questions/4101576/importing-ppm-images-with-python-and-pil-module

Its a dead simple image format and support for loading a text ppm and writing one 
could be added really easily.

i volunteer if anyone is interested.

CodeMarbles15:
Hmm, definitely should be added. Easy file format.

radarhere:
This looks like a duplicate of #860

wegfawefgawefg:
that issue is 6 years old and dead so.

lets make ascii ppm's happen!

wegfawefgawefg:
thanks you're a god

PR #5009, opened 2020-10-26
Add basic support for decoding plain text PPM formats

gofr:
"Fixes" #4999.

* Only adds decode support, not encode support
* Lacks tests
* Lacks error handling
* Is also affected by #5008


gofr:
Closing this one since #5242 is a better version of this.

PR #5011, opened 2020-10-26
Add save_all notation to webp

RadixSeven:
I just tried to use the WEBP writer to save an animation and was unable to without `save_all=true`. But that parameter was not mentioned in the documentation. This correction adds it.

Fixes
-----
* Add `save_all` as an option in the docs for WEBP

### DIFF
--- a/docs/handbook/image-file-formats.rst
+++ b/docs/handbook/image-file-formats.rst
@@ -951,6 +951,10 @@ When calling :py:meth:`~PIL.Image.Image.save` to write a WebP file, the
 following options are available when the ``save_all`` argument is present and
 true.
 
+**save_all**
+    If present and true, all frames of the image will be saved. If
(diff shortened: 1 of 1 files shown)


radarhere:
Thanks for trying to address the problem.

I wouldn't say that `save_all` isn't mentioned though - it's mentioned in the line just before your addition -
> When calling save() to write a WebP file, the following options are available when the save_all argument is present and true.

So combined with your change -
> When calling save() to write a WebP file, the following options are available when the save_all argument is present and true.
>
> **save_all**
    If present and true, all frames of the image will be saved. If not, then only the first frame of a multiframe image will be saved.

It is then saying that ``save_all`` can be used when ``save_all` is present... I find this to be confusing. I've created #5016 as an alternative - let us know if you have any thoughts on that.

RadixSeven:
I like #5016 better, so I'm closing this PR.

PR #5013, opened 2020-10-27
Updated CI targets

radarhere:
Updated CI targets after #5010 - Travis CI now tests [Python 3.9 on Ubuntu 16.04](https://travis-ci.org/github/python-pillow/Pillow/jobs/739128295)


hugovk:
Thanks!

PR #5016, opened 2020-10-28
Describe effect of WebP save_all

radarhere:
Alternative to #5011.

At the moment in https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html#saving-sequences, it doesn't actually describe what `save_all` does - it just mentions it.

This PR fixes that.


RadixSeven:
I think this is a good alternative to #5011 - which I have closed.

hugovk:
Thanks both!

PR #5014, opened 2020-10-27
AppVeyor: Test on Python 3.9 final

hugovk:
Fixes #4953.

Changes proposed in this pull request:

 * AppVeyor now supports Python 3.9 final
 * It requires the Visual Studio 2019 image (also works for 3.6)
 * And the setuptools bump
 * https://github.com/appveyor/ci/issues/3541

(It's like buses. You wait for ages and then [two](https://github.com/python-pillow/Pillow/pull/5010) come along at once.)

PR #5018, opened 2020-10-29
Updated CI targets

radarhere:
#5014 stopped testing Python 3.8 on Visual Studio 2017, [which is Windows Server 2016](https://www.appveyor.com/docs/windows-images-software/), so this PR removes that row from the CI targets.

That PR added Python 3.9 on Visual Studio 2019, [which is Windows Server 2019](https://www.appveyor.com/docs/windows-images-software/), which is already listed in the CI targets.


hugovk:
Thanks!

PR #5019, opened 2020-10-29
Invoke pip using Python

radarhere:
Another PR in the vein of #4465


.appveyor.yml
```diff
@@ -33,7 +33,7 @@ install:
         c:\pillow\winbuild\build\build_dep_all.cmd
         $host.SetShouldExit(0)
 - path C:\pillow\winbuild\build\bin;%PATH%
-- '%PYTHON%\%PIP_DIR%\pip.exe install -U "setuptools>=49.3.2"'
```
hugovk, line 36: Can the `PIP_DIR` var be removed?

radarhere: Good call, I've added a commit

PR #5037, opened 2020-11-11
init() if one of the formats is unrecognised

### LINKED ISSUE #5036: Unpopular `formats` argument of `PIL.Image.open` cause exceptions
In `open()`, `preinit()`  only initialize popular formats, 
`_open_core()` will raise if the  given  formats have not been initialized. The `init()` failback become unreachable. 
https://github.com/python-pillow/Pillow/blob/2d6e51e26c7683f067d29bd69f4fdf8eef003643/src/PIL/Image.py#L2901-L2934

### What did you do?

```python

from PIL import Image


def main():
    with open('any.webp', 'rb') as f:
        try:
            im = Image.open(f, formats=('WEBP',))
        except Exception:
            print(Image.ID)  # No WEBP in the list
            raise
        else:
            print(im.size)


if __name__ == '__main__':
    main()

```

Output the size
```console
(640, 480)
```

### What actually happened?
```console
['BMP', 'DIB', 'GIF', 'TIFF', 'JPEG', 'PPM', 'PNG']
  File "pil_debug.py", line 18, in <module>
    main()
  File "pil_debug.py", line 9, in main
    im = Image.open(f, formats=('WEBP',))
  File "lib\site-packages\PIL\Image.py", line 2929, in open
    im = _open_core(fp, filename, prefix, formats)
  File "lib\site-packages\PIL\Image.py", line 2909, in _open_core
    factory, accept = OPEN[i]
KeyError: 'WEBP'
```

* OS: Windows 10
* Python: 3.8
* Pillow: 8.0.1

radarhere:
Resolves #5036 

```python
from PIL import Image
im = Image.open('Tests/images/hopper.webp', formats=('WEBP',))
```
`Image.open` calls `preinit()` first and then expects to loop through them. However, `formats` overrides this, and so `WEBP` is checked, which throws a `KeyError` because `init()` has not been called yet, as `Image.open()` only calls it if the `preinit()` formats do not match.

This PR resolves that.

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -2905,6 +2905,8 @@ def open(fp, mode="r", formats=None):
 
     def _open_core(fp, filename, prefix, formats):
         for i in formats:
+            if i not in OPEN:
+                init()
             try:
                 factory, accept = OPEN[i]
                 result = not accept or accept(prefix)


hugovk:
Can this be tested too?

radarhere:
The key part of the change here is causing a module to be imported, so that `Image.register_open` is run.

https://github.com/python-pillow/Pillow/blob/2d6e51e26c7683f067d29bd69f4fdf8eef003643/src/PIL/WebPImagePlugin.py#L345

The problem is that conftest

https://github.com/python-pillow/Pillow/blob/2d6e51e26c7683f067d29bd69f4fdf8eef003643/Tests/conftest.py#L4-L9

calls `Image.init()` at the start of the tests

https://github.com/python-pillow/Pillow/blob/2d6e51e26c7683f067d29bd69f4fdf8eef003643/src/PIL/features.py#L216-L230

So the plugins are already imported. I'm not seeing a way to re-import a Python module.

I think it would be possible to mock a plugin, but I feel like that would be overly complex.

hugovk:
Okay, thanks!

PR #5029, opened 2020-11-03
Remove excess Travis CI jobs

hugovk:
For https://github.com/python-pillow/Pillow/issues/5028.

Changes proposed in this pull request:

 * Remove jobs which are already run on GitHub Actions

# TODO

 * Check coverage drops and see what's tested on Travis that's not in GHA
    * Ideally add to GHA
    * Or keep a single Travis job for a bit longer
    * Or take the drop

 * There's some redundant bits of scripts that can be removed now or later, e.g. `TRAVIS_PYTHON_VERSION` checks

Issue #5028, opened 2020-11-02
The new pricing model for Travis CI


hugovk:
Travis CI announced a new pricing model today:

* https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing

It's not entirely clear how this affects open source, but https://github.com/zopefoundation/meta/issues/38#issuecomment-720638731 looks like the best interpretation:

* You get a free trial of 10k credits to begin (equates to 1,000 Linux minutes; macOS costs more)
* When the trial runs out, they want you to pick a paid plan
* They will offer free credits for OSS "reviewed and allocated on a case by case basis", email support to apply
* Once these run out, you can email support again and ask for more (does this mean not being able to build during the review process?)

It looks like we should move more stuff from Travis CI to GitHub Actions.

cc @python-pillow/pillow-team

# Pillow

* https://travis-ci.com/python-pillow/Pillow

This repo, and the most active.

![image](https://user-images.githubusercontent.com/1324225/97916885-4c588b00-1d5c-11eb-80e1-a5642bb6b8b6.png)

* Let's ditch lint, it's on GHA
* The new CPU architectures Arm64, ppc64le, s390x are unique to Travis, let's keep them there
* 3.6-3.9 (AMD64) builds: let's remove these, we have them on GHA. Is there anything unique to these that aren't included in the equivalent GHA jobs?

# pillow-wheels

* https://travis-ci.com/python-pillow/pillow-wheels

The biggest build, less active. Essential for releases.

![image](https://user-images.githubusercontent.com/1324225/97917346-051eca00-1d5d-11eb-987e-45826b142e40.png)

With great timing, @nulano (thank you!) had already done the work and opened this PR to move all the jobs to GHA, except for Arm64 which need to stay there: https://github.com/python-pillow/pillow-wheels/pull/169.

Let's get that reviewed and merged.

# docker-images

* https://travis-ci.com/github/python-pillow/docker-images

A big build, less active.

![image](https://user-images.githubusercontent.com/1324225/97917243-de609380-1d5c-11eb-9381-aadbd653ce20.png)

Let's shift it all over to GHA. We'll need to set credentials as secrets to upload to Docker Hub. Or alternatively upload to [GitHub Packages](https://docs.github.com/en/free-pro-team@latest/actions/guides/publishing-docker-images#publishing-images-to-github-packages).

# Also

* https://travis-ci.com/github/python-pillow/Sane
* https://travis-ci.com/github/python-pillow/pillow-perf

[...]

PR #5050, opened 2020-11-21
Remove unused Travis CI config

hugovk:
Follow up to https://github.com/python-pillow/Pillow/pull/5029, for https://github.com/python-pillow/Pillow/issues/5028.

 * Remove some unused Travis config after reducing testing in #5029 

 * Removes the diffcover stuff, not run since #5029, and I've not checked it in a while, so let's just remove it (unless anyone else wants it), and we have linting on the CI anyway

PS Due to a git mistake I accidentally pushed this change directly to `upstream/master`, so I reset and force-pushed.

### DIFF
--- a/.ci/after_success.sh
+++ b/.ci/after_success.sh
@@ -7,7 +7,3 @@ if [[ $MATRIX_DOCKER ]]; then
 else
   coverage xml
 fi
-
-if [[ $TRAVIS ]]; then
-    codecov --flags TravisCI
-fi
--- a/.ci/after_success.sh
+++ b/.ci/after_success.sh
@@ -11,9 +11,3 @@ fi
 if [[ $TRAVIS ]]; then
     codecov --flags TravisCI
 fi
-
-if [ "$TRAVIS_PYTHON_VERSION" == "3.9" ]; then
-    # Coverage and quality reports on just the latest diff.
-    depends/diffcover-install.sh
-    depends/diffcover-run.sh
-fi
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -9,7 +9,7 @@ Please send a pull request to the master branch. Please include [documentation](
 - Fork the Pillow repository.
 - Create a branch from master.
 - Develop bug fixes, features, tests, etc.
-- Run the test suite. You can enable [Travis CI](https://travis-ci.com/account/repositories) and [AppVeyor](https://ci.appveyor.com/projects/new) on your repo to catch test failures prior to the pull request, and [Codecov](https://codecov.io/gh) to see if the changed code is covered by tests.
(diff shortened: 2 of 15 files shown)


hugovk:
https://github.com/python-pillow/Pillow/pull/5088 will remove `.travis.yml`, we can remove the rest of the old Travis config.

.ci/after_success.sh
```diff
@@ -7,13 +7,3 @@ if [[ $MATRIX_DOCKER ]]; then
 else
   coverage xml
 fi
-
-if [[ $TRAVIS ]]; then
-    codecov --flags TravisCI
-fi
```
radarhere, line 13: My idea from #5100 was `codecov --flags GitHubActions`. Why remove it instead?

hugovk: We already have this in `test.yml`, setting `GHA_X` flags via `-F`:


```yml
    - name: After success
      run: |
        .ci/after_success.sh

    - name: Upload coverage
      run: bash <(curl -s https://codecov.io/bash) -F ${{ matrix.codecov-flag }}
      env:
        CODECOV_NAME: ${{ matrix.os }} Python ${{ matrix.python-version }}
```

You can see them on `master` here:

https://codecov.io/gh/python-pillow/Pillow/commit/5c600adf4191e864568c1988cca4b045c7d05435/build

https://codecov.io/gh/python-pillow/Pillow/src/5c600adf4191e864568c1988cca4b045c7d05435/src/PIL/Image.py

.github/CONTRIBUTING.md
```diff
@@ -9,15 +9,15 @@ Please send a pull request to the master branch. Please include [documentation](
 - Fork the Pillow repository.
 - Create a branch from master.
 - Develop bug fixes, features, tests, etc.
-- Run the test suite. You can enable [Travis CI](https://travis-ci.com/account/repositories) and [AppVeyor](https://ci.appveyor.com/projects/new) on your repo to catch test failures prior to the pull request, and [Codecov](https://codecov.io/gh) to see if the changed code is covered by tests.
+- Run the test suite. You can enable [GitHub Actions](http://github.com/MY-USERNAME/Pillow/actions) and [AppVeyor](https://ci.appveyor.com/projects/new) on your repo to catch test failures prior to the pull request, and [Codecov](https://codecov.io/gh) to see if the changed code is covered by tests.
```
radarhere, line 12:
```suggestion
- Run the test suite. You can enable GitHub Actions (https://github.com/MY-USERNAME/Pillow/actions) and [AppVeyor](https://ci.appveyor.com/projects/new) on your repo to catch test failures prior to the pull request, and [Codecov](https://codecov.io/gh) to see if the changed code is covered by tests.
```
Since this isn't an actual link, is it worth showing it by itself? Also, https

.ci/after_success.sh
radarhere: Thanks for the explanation

Issue #5053, opened 2020-11-21
Docs: 4th level menu items not showing in dark mode


MarkSetchell:
The latest (coloured) documentation is completely unusable on macOS in "Dark Mode". You can't see anything in the table of contents (outlined in red in attached screen-grab)!

<img width="1193" alt="Screenshot 2020-11-21 at 18 21 14" src="https://user-images.githubusercontent.com/12937410/99884573-320d2100-2c27-11eb-977f-1e4c259a05a4.png">

hugovk:
What browser and OS versions are you using?

Here's how it looks for me with Chrome 86 on macOS Mojave:

![image](https://user-images.githubusercontent.com/1324225/99884781-6ee21380-2c39-11eb-8413-90c8a999e7ed.png)

hugovk:
Also https://pillow.readthedocs.io/en/stable/ is currently the 8.0.1 docs, but your title says 7.0.0. Please could you also empty your browser cache and reload?

MarkSetchell:
Hi Hugo,

Not sure many Mac users will use Chrome - I think they are FAR more likely to be using Safari on a Mac!

I use Safari 13.1.2 on a macOS Catalina 10.15.6 in Dark Mode

Thank you
Mark

hugovk:
This is how it looks for me with Safari 14.0:

![image](https://user-images.githubusercontent.com/1324225/99885978-a5239100-2c41-11eb-8599-dcdacb2ddb75.png)

Please could you clear your browser cache and reload?

MarkSetchell:
Hi Hugo 

I cleared the cache but it doesn't seem to make any difference on Safari 13. I will probably upgrade to Big Sur macOS 11 in the next 3-5 weeks so maybe that will help.

Best regards
Mark

hugovk:
Is it still showing 7.0.0?

![image](https://user-images.githubusercontent.com/1324225/99886372-a5715b80-2c44-11eb-8d22-f6c4e92a44a0.png)

It should be 8.0.1.

hugovk:
@jeremy-rifkin Please can you show what version of the docs you're on? Is it 7.0.0 or 8.0.1? The release notes page is a good place to check the latest mentioned.

https://pillow.readthedocs.io/en/stable/releasenotes/index.html

Does clearing the browser cache make any difference?

nulano:
[...]

PR #5056, opened 2020-11-24
Don't add light mode bg to 3rd and 4th level menu items

hugovk:
Fixes https://github.com/python-pillow/Pillow/issues/5053.

 * Don't add a light mode background colour to fourth-level menu items
 * Let them inherit styles instead
 * Fixes unreadable menu items in dark mode

# Example

https://pillow--5056.org.readthedocs.build/en/5056/reference/ImageFont.html#functions

# Dark mode

image

# Light mode

image


nulano:
This does fix the readability issue, but the menu looks less uniform without the extra background highlight. Trying the opposite color manually `#363636` looks good, although I don't see an obvious way to add it based on the dark/light selection. If that is too complicated then this is obviously worth removing at least for now.

hugovk:
How's this with CSS? For some reason the RTD PR build hasn't updated `dark.css` or included `light.css` in the docs, even though `light.css` is now on the server.

Anyway, it looks like this locally:

| image| image|

nulano:
The screenshots look good.

It's hard to test with the RtD build not updated, but I think this might break Internet Explorer compatibility (since it doesn't support the `@media` tag). However, I don't think it's worth worrying about a minor style inconsistency on an unsupported browser, so probably fine.

PR #5061, opened 2020-11-25
GHA: Install Raqm on macOS

nulano:
Before: https://github.com/python-pillow/Pillow/runs/1447978107#step:10:34
```
```

After: https://github.com/python-pillow/Pillow/runs/1452096925?check_suite_focus=true#step:10:34
```
--- RAQM (Bidirectional Text) support ok, loaded 0.7.0
```

### DIFF
--- a/.github/workflows/macos-install.sh
+++ b/.github/workflows/macos-install.sh
@@ -2,7 +2,7 @@
 
 set -e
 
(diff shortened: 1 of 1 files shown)


hugovk:
Thanks!

PR #5088, opened 2020-12-11
Move non-x86 tests to GHA

nulano:
For https://github.com/python-pillow/Pillow/issues/5028#issuecomment-721428190.
Companion to https://github.com/python-pillow/docker-images/pull/98.
Includes https://github.com/python-pillow/Pillow/pull/5091.

Move tests of arm64v8, ppc64le, s390x from Travis CI to GHA Docker tests using aptman/qus.

Other changes:

* (https://github.com/python-pillow/Pillow/pull/5091) Add xfail to libtiff tests that are failing on big-endian (see also https://github.com/python-pillow/Pillow/issues/4213#issuecomment-721425032).
* (https://github.com/python-pillow/Pillow/pull/5091) Add xfail to `test_libimagequant_quantize` failing on ppc64le (not sure why).
* FreeType 2.6 is no longer tested, remove FreeType<2.7 metrics from tests (see also #5075).

The new tests will fail until the new images get pushed, see run in my branch for now: https://github.com/nulano/Pillow/actions/runs/416245841

## Speed comparison (approximate)

| Architecture | Travis CI | GHA / Create Docker image | GHA / Test in Docker image | GHA Total |
|-|-|-|-|-|
| arm64    | 7 min           | 15 min 50 sec | 8 min 15 sec (1.18x Travis)   | 24 min 5 sec |
| ppc64le | 4 min 20 sec | 16 min (*)       | 9 min 15 sec (2.15x Travis)   | 25 min 15 sec (*) |
| s390x | 2 min 15 sec    | 14 min (*)        | 11 min 45 sec (5.22x Travis) | 25 min 45 sec (*) |

*: The ppc64le and s390x image builds currently fail a bit early due to new test failures, the install time will be slightly longer.

The total is a bit faster than I saw in https://github.com/python-pillow/Pillow/issues/5028#issuecomment-721428190 mostly because it installs WebP from `apt-get` instead of compiling locally. More importantly, only the test times should matter, which are in line with cache-miss Windows tests and slightly slower than macOS tests.

## Dependency versions:
Travis CI (removed):
```
--- PIL CORE support ok, compiled for 8.1.0.dev0
*** TKINTER support not installed
--- FREETYPE2 support ok, loaded 2.6.1
--- LITTLECMS2 support ok, compiled for 2.6
--- WEBP support ok, loaded 1.1.0
--- WEBP Transparency support ok
--- WEBPMUX support ok
--- WEBP Animation support ok
[...]

PR #5100, opened 2020-12-18
Travis CI is no longer used

radarhere:
As of #5088, Travis CI is no longer used for this repository.

### DIFF
--- a/.ci/test.sh
+++ b/.ci/test.sh
@@ -5,6 +5,6 @@ set -e
 python -bb -m pytest -v -x -W always --cov PIL --cov Tests --cov-report term Tests
 
 # Docs
-if [ "$TRAVIS_PYTHON_VERSION" == "3.9" ] && [ "$TRAVIS_CPU_ARCH" == "amd64" ]; then
+if [ "$GHA_PYTHON_VERSION" == "3.9" ] && [ "$GHA_OS" == "ubuntu-latest" ]; then
     make doccheck
 fi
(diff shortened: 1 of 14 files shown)


.ci/test.sh
```diff
@@ -5,6 +5,6 @@ set -e
 python -bb -m pytest -v -x -W always --cov PIL --cov Tests --cov-report term Tests
 
 # Docs
-if [ "$TRAVIS_PYTHON_VERSION" == "3.9" ] && [ "$TRAVIS_CPU_ARCH" == "amd64" ]; then
+if [ "$GHA_PYTHON_VERSION" == "3.9" ] && [ "$GHA_OS" == "ubuntu-latest" ]; then
```
hugovk, line 8: Already tested on GHA at https://github.com/python-pillow/Pillow/blob/9e3ad5e9971fc0ec300e181ef014a6f467572b2d/.github/workflows/test.yml#L104-L108

hugovk (commented):
This is pretty similar to https://github.com/python-pillow/Pillow/pull/5050, except https://github.com/python-pillow/Pillow/pull/5050 also:

> Removes the diffcover stuff, not run since #5029, and I've not checked it in a while, so let's just remove it (unless anyone else wants it), and we have linting on the CI anyway

Would you like to keep diffcover?

radarhere:
Oh, sorry for the duplicate.

hugovk:
No problem!

PR #5102, opened 2020-12-18
Add tests for ImagePath.Path

hugovk:
For https://github.com/python-pillow/Pillow/issues/722.

 * Adds some extra tests for `ImagePath.Path`
 * Bumps `path.c` coverage from 64.52% to 75.63%
 * ["Pillow (PIL) examples"](https://hhsprings.bitbucket.io/docs/programming/examples/python/PIL/ImagePath.html) was handy to check some example calls

### DIFF
--- a/Tests/test_imagepath.py
+++ b/Tests/test_imagepath.py
@@ -1,4 +1,5 @@
 import array
+import math
 import struct
 
 import pytest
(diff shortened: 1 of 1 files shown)


Tests/test_imagepath.py
```diff
@@ -1,80 +1,181 @@
...
+    with pytest.raises(ValueError) as e:
+        ImagePath.Path(coords)
+
+    assert str(e.value) == "wrong number of coordinates"
+
+
+@pytest.mark.parametrize(
+    "coords, expected",
+    [
+        ([0, 1, 2, 3], (0.0, 1.0, 2.0, 3.0)),
```
radarhere, line 91: I don't think this line adds any coverage - but you'll probably just tell me that it's to be more systematic

hugovk: Yeah, turns out this line didn't add any. The next one was to add coverage. Let's keep both to be systematic :)

PR #5108, opened 2020-12-19
Fix dereferencing potential null pointer

cgohlke:


src/_imaging.c
```diff
@@ -3180,6 +3180,10 @@ _draw_polygon(ImagingDrawObject* self, PyObject* args)
 
     /* Copy list of vertices to array */
     ixy = (int*) calloc(n, 2 * sizeof(int));
+    if (ixy == NULL) {
+        free(xy);
+        return NULL;
```
radarhere, line 3185:
```suggestion
        ImagingError_MemoryError();
        return NULL;
```

PR #5115, opened 2020-12-20
Fix use of a versus an in ImageCms.rst

HollowMan6:

 *  an color -> a color in ImageCms.rst


hugovk:
Thanks!

PR #5120, opened 2020-12-22
Block TIFFTAG_SUBIFD

### LINKED ISSUE #5119: Conversion of TIFF with CCITT Group 4 Fax Encoding sometimes crashes under Windows
### What did you do?
I'm trying to convert various TIFF files to TIFF with LZW compression. I have now repeatedly run into TIFF files with CCITT Group 4 Fax Encoding which will make Python crash under Windows. 

I expect an exception to be raised from the PIL TIFFImagePlugin, instead of a hard crash. 

### What actually happened?
On Windows, the code below with the TIFF file provided will crash the entire Python process without raising exceptions. Under WSL, an OSError is raised, however:

```
saving image
TIFFWriteDirectoryTagSubifd: Illegal value for SubIFD tag.
TIFFWriteDirectoryTagSubifd: Illegal value for SubIFD tag.
  File "main.py", line 33, in <module>
    img2tif(par_dir / "fail.tif", par_dir / "out")
  File "main.py", line 27, in img2tif
    im.save(str(img_out), compression="tiff_lzw")  # Python crashes here
  File "/home/jnik/.local/lib/python3.8/site-packages/PIL/Image.py", line 2151, in save
  File "/home/jnik/.local/lib/python3.8/site-packages/PIL/TiffImagePlugin.py", line 1631, in _save
    raise OSError(f"encoder error {s} when writing image file")
OSError: encoder error -9 when writing image file
```

* OS: Windows 10 Pro 2004, 19041.685 and WSL Debian (`Linux 4.19.128-microsoft-standard #1 SMP Tue Jun 23 12:58:10 UTC 2020 x86_64 GNU/Linux`)
* Python: Tested under 3.8.5 and 3.9.1
* Pillow: 8.0.1

### Code and further information

```python
# -----------------------------------------------------------------------------
# Imports
# -----------------------------------------------------------------------------
from pathlib import Path

from PIL import Image


# -----------------------------------------------------------------------------
# Image conversion
# -----------------------------------------------------------------------------


def img2tif(img_file: Path, out_dir: Path) -> None:
    img_out: Path = out_dir / img_file.name
    img_out = img_out.with_suffix(".tif")
    print(img_out)
    print("opening image")
    im = Image.open(img_file)
    print("loading image")
    im.load()
    print("saving image")
    im.save(str(img_out), compression="tiff_lzw")  # Python crashes here
    print("saved image")  # Never reached


if __name__ == "__main__":
    par_dir = Path(__file__).parent
    img2tif(par_dir / "fail.tif", par_dir / "out")

```
I have attached both the code and the offending TIFF image here: [tif_fail.zip](https://github.com/python-pillow/Pillow/files/5724335/tif_fail.zip)
Note that opening the offending file in e.g. IrfanView and saving it again will fix the issue. So the TIFF file is definitely damaged somehow. I don't expect Pillow to handle the file, but rather fail more gracefully.

radarhere:
Resolves #5119

The image from the issue is crashing when trying to set [TIFFTAG_SUBIFD](https://www.awaresystems.be/imaging/tiff/tifftags/subifds.html) - specifying offsets to more data.

However, given that we aren't writing data to these arbitrary locations, we might as well as just block the tag.

### DIFF
--- a/Tests/test_file_tiff.py
+++ b/Tests/test_file_tiff.py
@@ -4,7 +4,7 @@ from io import BytesIO
 import pytest
 
 from PIL import Image, TiffImagePlugin
-from PIL.TiffImagePlugin import RESOLUTION_UNIT, X_RESOLUTION, Y_RESOLUTION
+from PIL.TiffImagePlugin import RESOLUTION_UNIT, SUBIFD, X_RESOLUTION, Y_RESOLUTION
 
 from .helper import (
     assert_image_equal,
@@ -160,6 +160,14 @@ class TestFileTiff:
                 with Image.open(outfile) as reloaded:
                     reloaded.load()
                     assert (round(dpi), round(dpi)) == reloaded.info["dpi"]
+
+    def test_subifd(self, tmp_path):
+        outfile = str(tmp_path / "temp.tif")
+        with Image.open("Tests/images/g4_orientation_6.tif") as im:
+            im.tag_v2[SUBIFD] = 10000
+
+            # Should not segfault
+            im.save(outfile)
 
     def test_save_setting_missing_resolution(self):
         b = BytesIO()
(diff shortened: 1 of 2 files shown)


hugovk:
Thank you!

Issue #4926, opened 2020-09-16
Image.putalpha should handle 16/32-bit to RGBA conversion


bertsky:
This is similar to #2705 (which covered P→PA conversion):

IMHO images in 16-bit (`I;16`) or 32-bit (`I` / `F`) modes should be automatically converted to RGBA when `putalpha` is attempted on them.

Current (7.2.0) behaviour:
```python
>>> img = PIL.Image.new('F', (200,100), 1.0)
>>> img.putalpha(PIL.Image.new('L', (200,100), 255))
Traceback (most recent call last):
  File "/lib/python3.6/site-packages/PIL/Image.py", line 1627, in putalpha
    self.im.setmode(mode)
ValueError: image has wrong mode

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/lib/python3.6/site-packages/PIL/Image.py", line 1630, in putalpha
    im = self.im.convert(mode)
ValueError: conversion from F to LA not supported

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/lib/python3.6/site-packages/PIL/Image.py", line 1637, in putalpha
    raise ValueError("illegal image mode") from e
ValueError: illegal image mode
```

If this is too difficult or unwise for other reasons (like quantization), then at least the documentation should state which base modes are supported.

In the very least, the above ugly multi-catch error message should be avoided.

wiredfool:
I don't agree that there should be an automatic conversion, as we currently don't do implicit conversions that would lose data, which would be the case going from I;16 or F to an 8 bit image format. 

The multiple errors are something that should probably be cleaned up though.

bertsky:
> I don't agree that there should be an automatic conversion, as we currently don't do implicit conversions that would lose data, which would be the case going from I;16 or F to an 8 bit image format.

Not necessarily – 16-bit grayscale content could well be preserved in a 24-bit color image without too much loss in precision. But I agree that it would be unexpected (and need some form of color management to get back to an equivalent luminance in the end).

And since reporting this, I have also learned that Pillow does not even convert correctly (loosing only precision), but clips off the significant part of the dynamic range – without warning. See:
- #3011 (conversion from 16-bit to 8-bit clips)
- #3159 and #4044 (conversion from 32-bit 8-bit clips)

Given that, I am actually glad Pillow does raise an exception here at least.

radarhere:
[...]

PR #5121, opened 2020-12-22
Improve handling of PPM headers

Piolie:
Addresses #5104. That discussion didn't get any replies, so I decided to open this PR to make my intentions more explicit.

 * remove some dead code;
 * handle non-decimal ASCII in headers properly;
 * handle comment strings that can start in the middle of a token and end with LF or CR;
 * make the exception types raised more consistent (although the resolution of issues _raised_ in #2339 and related is still pending);
 * add two tests; change the exception type expected in other two.


src/PIL/PpmImagePlugin.py
```diff
@@ -49,25 +49,39 @@ class PpmImageFile(ImageFile.ImageFile):
...
-        return s
+            token += c
+        return token
 
     def _open(self):
-
-        # check magic
-        s = self.fp.read(1)
-        if s != b"P":
-            raise SyntaxError("not a PPM file")
```
radarhere, line 69: Why remove this block?

Piolie: The [docs](https://pillow.readthedocs.io/en/stable/handbook/writing-your-own-file-decoder.html) state that Pillow:

> ...loops over the available image plugins in the loaded order, and calls the plugin’s `_accept` function with the first 16 bytes of the file. If the `_accept` function returns true, the plugin’s `_open` method is called...

I think the relevant lines are these:

https://github.com/python-pillow/Pillow/blob/2d7c543a2f2d853c8b69425040d36411eb0e2e39/src/PIL/Image.py#L2911-L2912

The `_accept` method already checks the magic numbers. I removed that block because it looked like dead code to me. That assumption might be wrong. My intention was to discuss that and other points before submitting.

I see that other plugins (BMP and JPEG at least) also recheck the magic, either manually (PPM) or calling `_accept`, so I'm probably missing something.

gofr: `_accept` is only guaranteed to be called if you're going through `Image.open()` and relying on the plugin registration. I can also do this to load an image:

```python
import io
from PIL import PpmImagePlugin
PpmImagePlugin.PpmImageFile(fp=io.BytesIO(b'P5\n2 2\n255\n\x80\x00\x00\xFF')).show()
```

That calls `_open` directly, without calling `_accept` first.

src/PIL/PpmImagePlugin.py
```diff
@@ -49,25 +49,39 @@ class PpmImageFile(ImageFile.ImageFile):
...
+        def _ignore_comment():  # ignores rest of the line; stops at CR, LF or EOF
+            while True:
+                c = self.fp.read(1)
+                if c in b"\r\n":
+                    break
+
+        while True:  # read until non-whitespace is found
+            c = self.fp.read(1)
+            if c == b"#":  # found comment, ignore it
+                _ignore_comment()
```
gofr, line 62: This function is also used to read the magic number, after it has read the initial "P". `P# comment` is not a valid magic number.

src/PIL/PpmImagePlugin.py
```diff
@@ -83,29 +97,21 @@ def _open(self):
...
-                s = self.fp.readline()
-            s = int(self._token(s))
-            if ix == 0:
-                xsize = s
-            elif ix == 1:
-                ysize = s
+            try:  # check token sanity
+                token = int(self._read_token())
+            except ValueError:
+                raise SyntaxError("Non-decimal-ASCII found in header")
```
gofr, line 103: (See https://github.com/python-pillow/Pillow/discussions/5104#discussioncomment-255699 for why this should remain a `ValueError`.)

And I don't think this exception message is necessarily better than what the original exception gives me. It doesn't tell me which bytes were the problem. I always prefer to see what value exactly it was it choked on. That makes it easier for me to see what I did wrong, or perhaps show enough info to file a PIL bug if it's something strange.

Or actually, it _does_ tell me, because this will show me 2 exceptions: first the original `ValueError` with its stack, then "During handling of the above exception, another exception occurred:" and the `SyntaxError` with its stack.

I don't think that amount of detail is useful for a PIL user. Re-raising might still be worth it to get a prettier message, but then I think it should do something like this to avoid the double stack:

```python
try:
    token = self._read_token()
    token = int(token)
except ValueError:
    raise ValueError(f"Better message with {token} in it") from None
```

Though this would need some additional work. If something is really wrong with the file or with PIL, this `token` could be any amount of data. The message for the built-in exception already truncates the value. This re-raised message does _not_ truncate the message, so a messed up token could give me a messed up exception message too...

So I'd say this isn't worth the trouble.

gofr: But if you bring back the token size limit in some way then it's probably safe enough to do.

src/PIL/PpmImagePlugin.py
gofr: Adding an `allow_comments=True` parameter to the function is maybe the easiest way to fix this?

Piolie: I was thinking of writing a short `_read_magic` method (which would essentially be a rebranded `_token`) to first read the magic number. The rest would be handled by `_read_token` with an added limit to the token length. What do you say?

src/PIL/PpmImagePlugin.py
Piolie: So that's what I had missed!

src/PIL/PpmImagePlugin.py
Piolie: I would use `_read_token` before the `try` because it can raise `ValueError` by itself if the file ends while reading the header.
It would be:
````python
token = self._read_token()  # if EOF is reached we get ValueError("Reached EOF while reading header")
try:
    token = int(token)
except ValueError:
    raise ValueError(f"Non-decimal-ASCII found in header: {token.decode('ascii')}") from None
````

File: src/PIL/PpmImagePlugin.py (hunk 2 above)
Comment (gofr) [reply src/PIL/PpmImagePlugin.py:62]: Sounds good to me.

File: src/PIL/PpmImagePlugin.py (hunk 3 above)
Comment (gofr) [reply src/PIL/PpmImagePlugin.py:103]: Right, yes.

But I don't think the `.decode('ascii')` should be there. If the token contains non-printable characters they'll disappear, or if it contains non-ASCII characters that'll raise a `UnicodeDecodeError`. I think it should just print the bytes value, unless `_read_token()` already raises its own exception if the token is not ASCII.

File: src/PIL/PpmImagePlugin.py [hunk 4]
```diff
@@ -49,25 +49,39 @@ class PpmImageFile(ImageFile.ImageFile):
...
+                    break
+
+        while True:  # read until non-whitespace is found
+            c = self.fp.read(1)
+            if c == b"#":  # found comment, ignore it
+                _ignore_comment()
+                continue
+            if c in B_WHITESPACE:  # found whitespace, ignore it
+                if c == b"":  # reached EOF
+                    raise EOFError("Reached EOF while reading header")
```
gofr, line 66: This looks fine to me, but the exception handler you found in https://github.com/python-pillow/Pillow/discussions/5104#discussioncomment-261675 suggests `EOFError` is only expected in case you "got header but not the first frame". Does anyone think this would be a problem?

Piolie: From what we discussed in that thread I had made my mind to revert to `ValueError` everywhere (except when checking the magic number).

Tests/test_file_ppm.py
```diff
@@ -49,6 +49,38 @@ def test_pnm(tmp_path):
...
             assert_image_equal(im, reloaded)
 
+def test_not_ppm(tmp_path):
+    path = str(tmp_path / "temp.djvurle")
+    with open(path, "wb") as f:
+        f.write(b"PyXX")
+
+    with pytest.raises(UnidentifiedImageError):
+        Image.open(path)
+
```
gofr, line 59: The lint check wants two blank lines between each test function.

src/PIL/PpmImagePlugin.py
```diff
@@ -83,29 +111,22 @@ def _open(self):
...
                     break
-            elif ix == 2:
-                # maxgrey
-                if s > 255:
+            elif ix == 2:  # token is maxval
+                if token > 255:
                     if not mode == "L":
-                        raise ValueError(f"Too many colors for band: {s}")
-                    if s < 2 ** 16:
+                        raise ValueError(f"Too many colors for band: {token}")
```
gofr, line 128: You didn't change this, so you don't have to do this at all, but the code coverage warning here can be fixed by adding a test with an image like this: `b"P6\n1 1\n1000\n"`.

I'd be happy to add that in a separate PR if you don't want to.

Tests/test_file_ppm.py
```diff
@@ -49,6 +49,38 @@ def test_pnm(tmp_path):
         with Image.open(f) as reloaded:
             assert_image_equal(im, reloaded)
 
+def test_not_ppm(tmp_path):
+    path = str(tmp_path / "temp.djvurle")
+    with open(path, "wb") as f:
+        f.write(b"PyXX")
```
gofr, line 55: If you change `PyXX` to something like `PyInvalid` you'll also cover the 6-byte length check.

gofr (commented):
I'm just a random contributor so I don't dare to "Approve", but apart from the minor notes I added this looks good to me. :+1:

Piolie:
I hurried to push and forgot to lint and a `raise from None`. I'll fix that and the other tests as noted by @gofr.

gofr (approved):

Tests/test_file_ppm.py
```diff
@@ -50,6 +50,55 @@ def test_pnm(tmp_path):
...
 
 
+def test_not_ppm(tmp_path):
+    path = str(tmp_path / "temp.djvurle")
+    with open(path, "wb") as f:
+        f.write(b"PyInvalid")
+
+    with pytest.raises(UnidentifiedImageError):
+        with Image.open(path):
+            pass
```
radarhere, line 60: So this is testing that not any file starting with P is considered a PPM file?

Piolie: Almost. Not any file starting with `Py` (accounting for some stated "PIL extensions").

In my other PR for the plain PPM decoder (unhappily built on top of this one), I replaced that test with:
```python
def test_not_ppm(tmp_path):
    with pytest.raises(SyntaxError):
        PpmImagePlugin.PpmImageFile(fp=BytesIO(b"PyInvalid"))

File: src/PIL/PpmImagePlugin.py [hunk 9]
```diff
@@ -135,7 +157,7 @@ def _save(im, fp, filename):
     elif im.mode == "RGBA":
         rawmode, head = "RGB", b"P6"
     else:
-        raise OSError(f"cannot write mode {im.mode} as PPM")
+        raise OSError(f"Cannot write mode {im.mode} as PPM")
```
radarhere, line 160:
```suggestion
        raise OSError(f"cannot write mode {im.mode} as PPM")
```
Considering that "cannot write mode" is used in 9 other plugins, it would be nice not to have this one be unique.

src/PIL/PpmImagePlugin.py
```diff
@@ -49,26 +49,52 @@ class PpmImageFile(ImageFile.ImageFile):
     format = "PPM"
     format_description = "Pbmplus image"
 
-    def _token(self, s=b""):
+    def _read_magic(self, magic=b""):
         while True:  # read until next whitespace
             c = self.fp.read(1)
-            if not c or c in b_whitespace:
+            if c in B_WHITESPACE:
```
radarhere, line 55: You've removed the `not c` condition. The purpose of that would have been to avoid an infinite loop if `c` was empty.
i mean, technically, an empty string is in `B_WHITESPACE`, so it will pass, but your version seems less clear.

Piolie: Since the [Library Reference](https://docs.python.org/3/reference/expressions.html#in) states that:
> Empty strings are always considered to be a substring of any other string, so "" in "abc" will return `True`.

I considered the check to be redundant. Maybe a short comment like `# also checks for empty string` would make it clearer? Change it as you see fit.

src/PIL/PpmImagePlugin.py
radarhere: This was addressed by https://github.com/python-pillow/Pillow/pull/5121/commits/d2ad27d70a00efa98560cf5036326abcae850d5e

src/PIL/PpmImagePlugin.py
radarhere: https://github.com/python-pillow/Pillow/pull/5121/commits/5d0ad5e2e9a5065d69a00890342334f374277d1c addressed this by changing to `ValueError`, and then https://github.com/python-pillow/Pillow/pull/5121/commits/73fed77c0c58efd2ce11b8c0e96858bd03655163 added `from None`.

radarhere: Considering we'd had an issue in the past about multiple errors being raised (#4926), I think this is a step backwards. I've pushed a commit to remove this.

src/PIL/PpmImagePlugin.py
radarhere: I've pushed a commit to use `not c`.

src/PIL/PpmImagePlugin.py
radarhere: The SyntaxError was restored in https://github.com/python-pillow/Pillow/pull/5121/commits/d2ad27d70a00efa98560cf5036326abcae850d5e

src/PIL/PpmImagePlugin.py
radarhere: A test was added for this in https://github.com/python-pillow/Pillow/commit/bc5ecfb79c3dca0553e2b863ba7d306f2e0fd434

src/PIL/PpmImagePlugin.py
radarhere: This was addressed in https://github.com/python-pillow/Pillow/pull/5121/commits/5d0ad5e2e9a5065d69a00890342334f374277d1c

Tests/test_file_ppm.py
radarhere: This was done in https://github.com/python-pillow/Pillow/pull/5121/commits/bc5ecfb79c3dca0553e2b863ba7d306f2e0fd434

src/PIL/PpmImagePlugin.py
```diff
@@ -49,26 +49,46 @@ class PpmImageFile(ImageFile.ImageFile):
...
-        while True:  # read until next whitespace
+    def _read_magic(self):
+        magic = b""
+        # read until whitespace or limit of 6 characters
+        for _ in range(6):
             c = self.fp.read(1)
             if not c or c in b_whitespace:
                 break
-            if c > b"\x79":
-                raise ValueError("Expected ASCII value, found binary")
```
radarhere, line 58: This was removed because, as point 4 of #5104 states, b"\x79" is "y"... which is confusing, because that 's still ASCII. The intention was probably to the limit the start of the magic to "Py" and nothing alphabetically after that.

Tests/test_file_ppm.py
```diff
@@ -50,15 +50,70 @@ def test_pnm(tmp_path):
...
+
+
+def test_nondecimal_header(tmp_path):
+    path = str(tmp_path / "temp.ppm")
+    with open(path, "wb") as f:
+        f.write(b"P6\n128\x00")
+
+    with pytest.raises(ValueError):
+        with Image.open(path):
+            pass
```
radarhere, line 79: I'm not quite on board with this one. The idea here is that b"\x00" is not caught by https://github.com/python-pillow/Pillow/blob/92c26a77ca53a2bfbd8804f009c6c8755d0e5a43/src/PIL/PpmImagePlugin.py#L57-L58
even though \x00 is not greater than \x79, and only for it to be raised as a different ValueError later when it turns out that b"128\x00" is not an integer.

There isn't a valid scenario where the token/magic needs to be something other than a number or a letter. I've pushed a commit to change this test into a check that a ValueError is raised when a token is not an integer.

PR #5122, opened 2020-12-22
Fixed comparison warnings

radarhere:
Helps #4586

- Fixed comparisons between int and unsigned long
- Applied @nulano's suggestions from #4752, fixing comparisons between unsigned int and int


nulano (approved):

hugovk:
Thanks!

PR #5149, opened 2020-12-28
Fix Out of bounds read when saving GIF of xsize=1

wiredfool:
Triggered in Tests/test_file_gif.py::test_optimize

There is a OOB read of size 1 when the width of the image is 1. 

Valgrind output:
```
Tests/test_file_gif.py::test_optimize **70** 
**70** **********************************************************************
**70** Tests/test_file_gif.py::test_optimize
**70** **********************************************************************
==70== Invalid read of size 1
==70==    at 0x87A7194: ImagingGifEncode (GifEncode.c:237)
==70==    by 0x877D519: _encode (encode.c:141)
==70==    by 0x64BF47: method_vectorcall_VARARGS (descrobject.c:300)
```

### DIFF
--- a/src/libImaging/GifEncode.c
+++ b/src/libImaging/GifEncode.c
@@ -241,8 +241,6 @@ ImagingGifEncode(Imaging im, ImagingCodecState state, UINT8* buf, int bytes)
                     break;
                 }
 
-                this = state->buffer[state->x++];
-
                 if (this == context->last) {
                     context->count++;
                 } else {
--- a/src/libImaging/GifEncode.c
+++ b/src/libImaging/GifEncode.c
@@ -233,6 +233,13 @@ ImagingGifEncode(Imaging im, ImagingCodecState state, UINT8* buf, int bytes)
                         }
 
                 }
+                /* Potential special case for xsize==1 */
+                if (state->x < state->xsize) {
+                    this = state->buffer[state->x++];
+                } else {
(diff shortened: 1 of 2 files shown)


src/libImaging/GifEncode.c
```diff
@@ -233,6 +233,13 @@ ImagingGifEncode(Imaging im, ImagingCodecState state, UINT8* buf, int bytes)
...
                 }
+                /* Potential special case for xsize==1 */
+                if (state->x < state->xsize) {
+                    this = state->buffer[state->x++];
+                } else {
+                    EMIT_RUN(label0);
+                    break;
+                }
 
                 this = state->buffer[state->x++];
```
radarhere, line 244:
```suggestion
```

I don't have valgrind setup, so I'm not on top of this - but did you mean to remove this line (so that it could be replaced by your new ones)? Because doing so causes the tests to pass.

wiredfool: Yeah, that was a bad rebase, as this was originally (accidentally) written on top of a branch that was a year old. (I mean, last update was Dec 23, how out of date could it be?)

wiredfool: You should have an invite from a couple days back for a docker image that will do valgrind.

radarhere: Ah, yes, I see. Thanks

PR #5150, opened 2020-12-28
Support for ignoring tests when running valgrind

wiredfool:
Before we can add valgrind as an experimental docker test target, we need to have all of the tests passing or known ignored.

This adds:

* a pytest configuration for ignoring the error when pytest-valgrind isn't installed. 
* Ignores the tiff_additional_metadata test, because we're sending some illogical values to it. 

WIP -- there are more tests to investigate.

### DIFF
--- a/Tests/conftest.py
+++ b/Tests/conftest.py
@@ -24,7 +24,7 @@ def pytest_configure(config):
         warnings.simplefilter("error")
         try:
             getattr(pytest.mark, "valgrind_known_error")
-        except:
+        except Exception:
             config.addinivalue_line(
                 "markers",
                 "valgrind_known_error: Tests that have known issues with valgrind",
--- a/Tests/conftest.py
+++ b/Tests/conftest.py
@@ -10,3 +10,18 @@ def pytest_report_header(config):
             return out.getvalue()
     except Exception as e:
         return f"pytest_report_header failed: {e}"
+
+def pytest_configure(config):
+    # We're marking some tests to ignore valgrind errors and XFAIL them.
+    # Ensure that the mark is defined even in cases where pytest-valgrind isn't installed
+
+    import pytest
+    import warnings
+
+    with warnings.catch_warnings():
+        warnings.simplefilter("error")
+        try:
+            getattr(pytest.mark, "valgrind_known_error")
+        except:
+            config.addinivalue_line("markers",
+                                    "valgrind_known_error: Tests that have known issues with valgrind")
--- a/Tests/conftest.py
+++ b/Tests/conftest.py
@@ -11,17 +11,22 @@ def pytest_report_header(config):
     except Exception as e:
         return f"pytest_report_header failed: {e}"
 
+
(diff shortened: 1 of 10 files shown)


Tests/conftest.py
```diff
@@ -10,3 +10,18 @@ def pytest_report_header(config):
...
+    # Ensure that the mark is defined even in cases where pytest-valgrind isn't installed
+
+    import pytest
+    import warnings
+
+    with warnings.catch_warnings():
+        warnings.simplefilter("error")
+        try:
+            getattr(pytest.mark, "valgrind_known_error")
+        except:
```
hugovk, line 25: ?
```suggestion
        except AttributeError:
```

Tests/test_file_libtiff.py
```diff
@@ -185,6 +185,7 @@ def test_write_metadata(self, tmp_path):
             for field in requested_fields:
                 assert field in reloaded, f"{field} not in metadata"
 
+    @pytest.mark.valgrind_known_error(reason="Known Invalid Metadata")
```
hugovk, line 188:
```suggestion
    @pytest.mark.valgrind_known_error(reason="Known invalid metadata")
```

Tests/conftest.py
```diff
@@ -10,3 +10,18 @@ def pytest_report_header(config):
             return out.getvalue()
     except Exception as e:
         return f"pytest_report_header failed: {e}"
+
+def pytest_configure(config):
+    # We're marking some tests to ignore valgrind errors and XFAIL them.
+    # Ensure that the mark is defined even in cases where pytest-valgrind isn't installed
+
+    import pytest
+    import warnings
```
hugovk, line 19: These could go at the top of the file.

Tests/conftest.py
```diff
@@ -10,3 +10,18 @@ def pytest_report_header(config):
             return out.getvalue()
     except Exception as e:
         return f"pytest_report_header failed: {e}"
+
```
radarhere, line 13:
```suggestion


```
Lint fix

Tests/conftest.py
```diff
@@ -10,3 +10,18 @@ def pytest_report_header(config):
...
+    import pytest
+    import warnings
+
+    with warnings.catch_warnings():
+        warnings.simplefilter("error")
+        try:
+            getattr(pytest.mark, "valgrind_known_error")
+        except:
+            config.addinivalue_line("markers",
+                                    "valgrind_known_error: Tests that have known issues with valgrind")
```
radarhere, line 27:
```suggestion
            config.addinivalue_line(
                "markers",
                "valgrind_known_error: Tests that have known issues with valgrind"
            )
```
Lint fix

Tests/conftest.py
wiredfool: I don't think it's an attribute error, as it's catching a warning thrown by pytest which hooks the getattr to specifically avoid an attribute error.

PR #5151, opened 2020-12-29
Updated openjpeg to 2.4.0

radarhere:


nulano:
The release notes include "Many bug fixes (including security fixes)".

Is this worth mentioning in the Pillow release notes (e.g. wheels now include OpenJpeg 2.4.0 which has several security fixes)?

For reference, the full changelog is here: https://github.com/uclouvain/openjpeg/blob/v2.4.0/CHANGELOG.md#v240-2020-12-28

radarhere:
Ok, sure - I've created PR #5152, since that's technically about the pillow-wheels PR, not this change.
I don't know that it's anything unusual though - looking at https://github.com/uclouvain/openjpeg/blob/v2.4.0/NEWS.md, every release since 2.1.1 has had 'security fixes'.

PR #5155, opened 2020-12-30
Document and add tests for SBIX color font support

nulano:
In #4955 I missed that FreeType 2.5.1 added SBIX font support, this PR adds tests and corrects the documentation / 8.0.0 release notes. The font is from https://github.com/RoelN/ChromaCheck (MIT license), I couldn't find any other fonts that were under 15MB.

Also some minor formatting cleanup of the other color font tests.

FreeType release notes: https://sourceforge.net/projects/freetype/files/freetype2/2.5.1/


Tests/test_imagefont.py
```diff
@@ -918,6 +918,46 @@ def test_cbdt_mask(self):
...
+            font = ImageFont.truetype(
+                "Tests/fonts/chromacheck-sbix.woff",
+                size=300,
+                layout_engine=self.LAYOUT_ENGINE,
+            )
+
+            im = Image.new("RGB", (400, 400), "white")
+            d = ImageDraw.Draw(im)
+
+            d.text((50, 50), "\uE901", embedded_color=True, font=font)
```
radarhere, line 933:
```suggestion
            d.text((50, 50), "\uE901", font=font, embedded_color=True)
```
Minor suggestion - because `font` comes before `embedded_color` in the argument list.

Tests/test_imagefont.py
```diff
@@ -918,6 +918,46 @@ def test_cbdt_mask(self):
...
+
+            with Image.open("Tests/images/chromacheck-sbix.png") as expected:
+                assert_image_similar(im, expected, 1)
+        except IOError as e:
+            assert str(e) in ("unimplemented feature", "unknown file format")
+            pytest.skip("freetype compiled without libpng or unsupported")
```
radarhere, line 939: What exactly does 'or unsupported' mean?

nulano: The tests are a direct copy of the `test_cbdt...` tests, I no longer remember what is the difference. :/ It might be a left-over from before I added the version check, but I'm not sure. Or it might be referring to a version of FreeType compiled with libpng included but CBDT (or in this case SBIX and WOFF) disabled.

I'll change it to `...or without SBIX support`.

hugovk:
> The font is from [RoelN/ChromaCheck](https://github.com/RoelN/ChromaCheck) (MIT license), I couldn't find any other fonts that were under 15MB.

I was going to say 15 MB is too big (the sdist is currently 45 MB) and we could move it instead to https://github.com/python-pillow/pillow-depends, but `chromacheck-sbix.woff` is only 740B?

nulano:
> > The font is from [RoelN/ChromaCheck](https://github.com/RoelN/ChromaCheck) (MIT license), I couldn't find any other fonts that were under 15MB.
> 
> I was going to say 15 MB is too big (the sdist is currently 45 MB) and we could move it instead to https://github.com/python-pillow/pillow-depends, but `chromacheck-sbix.woff` is only 740B?

I meant that all other fonts I could find were 15 MB or larger, not that this specific font is large :smile:. I would have preferred to use a font more complicated than just a single rectangle, but couldn't find anything small enough.

hugovk:
Ah great! Well let's get the merge conflict resolved and this merged!

nulano:
> Ah great! Well let's get the merge conflict resolved and this merged!

Resolved.

hugovk (approved):

nulano:
> GitHub App like Mergify are not allowed to merge pull request where .github/workflows is changed.

Because I changed (fixed) a comment in the workflow... 😕

PR #5159, opened 2020-12-30
Makefile updates

wiredfool:
* Set the default goal to something unsurprising.
* Remove obsolete target (co)
* Add the lint target.


Makefile
```diff
@@ -109,3 +102,9 @@ test:
 .PHONY: readme
 readme:
 	python3 setup.py --long-description | markdown2 > .long-description.html && open .long-description.html
+
+
+.PHONY: lint
+lint:
+	tox --help > /dev/null || pip install tox
```
hugovk, line 109: We use pip like this elsewhere in the Makefile:


```suggestion
	tox --help > /dev/null || python3 -m pip install tox
```

Let's also put the make commands in roughly alphabetical order to make them easier to find, and add a help entry for lint:

```sh
$ make help
Welcome to Pillow development. Please use `make <target>` where <target> is one of
  clean              remove build products
  coverage           run coverage test (in progress)
  doc                make html docs
  docserve           run an http server on the docs directory
  html               to make standalone HTML files
  inplace            make inplace extension
  install            make and install
  install-coverage   make and install with C coverage
  install-req        install documentation and test dependencies
  install-venv       install in virtualenv
  release-test       run code and package tests before release
  test               run tests on installed pillow
  upload             build and upload sdists to PyPI
  upload-test        build and upload sdists to test.pythonpackages.com
```

wiredfool:
```
install-venv:
        virtualenv .
        bin/pip install -r requirements.txt
```

It's 1:1 for `python3 -m pip install` vs `pip install`, though to be fair, I think this is an artifact of the original usage, and I'd argue against both this and the inplace target on a general objection to mucking up the source tree. (though, I do understand that it's potentially used for some coverage or automated thing. )

Do we have a handle on what people are actually using from this makefile?

hugovk: Other than the release checklist commands (`release-test`, `sdist`) I mainly only use `doc` and `clean`.

hugovk: https://snarky.ca/why-you-should-use-python-m-pip/ is the reason for `python3 -m pip`, mostly important for the docs as we often get bug reports caused by installing for a different interpreter.

Similarly, I have earlier had problems with `python` in by the Makefile pointing to Python 2, not 3.

hugovk (approved):
isort could be handy for lint-fix too

wiredfool:
Yeah, anything that just fixes it is good.

Makefile
wiredfool: Also, I'm not a fan of sorting the rules alphabetically, I'd rather group them by purpose. 

(especially as there are some new ones coming in that are 1 public, and several dependent ones)

Makefile
```diff
@@ -47,7 +40,9 @@ help:
 	@echo "  install            make and install"
 	@echo "  install-coverage   make and install with C coverage"
 	@echo "  install-req        install documentation and test dependencies"
-	@echo "  install-venv       install in virtualenv"
+	@echo "  install-venv       (deprecated) install in virtualenv"
+	@echo "  lint	       	    run the lint checks"
+	@echo "  lint-fix 	    run black to (mostly) fix lint issues."
```
radarhere, line 45:
```suggestion
	@echo "  lint-fix 	        run black to (mostly) fix lint issues."
```
Match up with the other lines.

Makefile
```diff
@@ -47,7 +40,9 @@ help:
 	@echo "  install            make and install"
 	@echo "  install-coverage   make and install with C coverage"
 	@echo "  install-req        install documentation and test dependencies"
-	@echo "  install-venv       install in virtualenv"
+	@echo "  install-venv       (deprecated) install in virtualenv"
+	@echo "  lint               run the lint checks"
+	@echo "  lint-fix           run black to (mostly) fix lint issues."
```
radarhere, line 45:
```suggestion
	@echo "  lint-fix           run black and isort to (mostly) fix lint issues."
```

Issue #5145, opened 2020-12-28
Support PySide6 for ImageQt


kvnxiao:
Support PySide6 https://www.qt.io/blog/qt-for-python-6-released

by updating the qt_versions in `ImageQt.py` to allow using Qt6 with Pillow

hugovk:
@kvnxiao Please could you test out PR #5161?

PR #5161, opened 2020-12-30
Add support for PySide6

hugovk:
Fixes https://github.com/python-pillow/Pillow/issues/5145.

There are two sets of Python bindings for Qt, called PyQt and PySide. 

> Why are there two packages?
>
> PyQt has been developed by Phil Thompson of Riverbank Computing Ltd. for a very long time — supporting versions of Qt going back to 2.x. Back in 2009 Nokia, who owned the Qt toolkit at the time, wanted to have Python bindings for Qt available under the LGPL license (like Qt itself). Unable to come to agreement with Riverbank (who would lose money from this, so fair enough) they then released their own bindings as PySide (also, fair enough).

https://www.learnpyqt.com/tutorials/pyqt5-vs-pyside2/

We support:

* PySide2 for Qt 5
* PyQt5 for Qt 5

Qt 6 has been recently released: https://www.qt.io/blog/qt-6.0-released

* PyQt6 is still in development: https://riverbankcomputing.com/software/pyqt/download Let's add it when it's available.
* PySide6 for Qt 6 is out: https://www.qt.io/blog/qt-for-python-6-released (they bumped the PySide version to match the Qt version)

 * Add support for PySide6. Take precedence over PyQt5 and PySide2, if they're also installed.

### DIFF
--- a/Tests/test_imageqt.py
+++ b/Tests/test_imageqt.py
@@ -14,7 +14,9 @@ def test_rgb():
     # typedef QRgb
     # An ARGB quadruplet on the format #AARRGGBB,
     # equivalent to an unsigned int.
-    if ImageQt.qt_version == "5":
+    if ImageQt.qt_version == "side6":
+        from PySide6.QtGui import qRgb
+    elif ImageQt.qt_version == "5":
         from PyQt5.QtGui import qRgb
     elif ImageQt.qt_version == "side2":
         from PySide2.QtGui import qRgb
--- a/Tests/test_qt_image_qapplication.py
+++ b/Tests/test_qt_image_qapplication.py
@@ -7,7 +7,10 @@ from .helper import assert_image_equal, hopper
 if ImageQt.qt_is_installed:
     from PIL.ImageQt import QPixmap
 
-    if ImageQt.qt_version == "5":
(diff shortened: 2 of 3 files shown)


radarhere:
Is it concerning that if I [`pip install PySide6`](https://github.com/radarhere/Pillow/commit/ae90510f207a9faaa0104160cfbc13bfade4ba63), Ubuntu starts [segfaulting](https://github.com/radarhere/Pillow/runs/1631785200?check_suite_focus=true#step:10:44)? When I [`pip install PySide2`](https://github.com/radarhere/Pillow/commit/ea1f468cb5a473a9aa4d40b4bc40b254f8e28e26), this is not [what happens](https://github.com/radarhere/Pillow/runs/1631790350?check_suite_focus=true).

hugovk:
It'll be missing dependencies for Ubuntu, for example we have `libxcb-xinerama0 pyqt5-dev-tools` for PyQt5.

I've created https://github.com/python-pillow/docker-images/pull/102 to add PySide6 and its dependencies to the ubuntu-20.04-focal-amd64 image.

radarhere:
Thanks

PR #5171, opened 2021-01-02
Add #5159 to the release notes

radarhere:
Document #5159

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,6 +76,7 @@ install-req:
 
 .PHONY: install-venv
 install-venv:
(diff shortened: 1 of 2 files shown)


hugovk:
I don't think we've put Makefile stuff in the release notes before, but no objections to adding.

To really deprecate, should we echo a warning when the command is run? I doubt we need to do the full deprecation thing for Makefile (deprecations.rst, wait until major bump before removal).

radarhere:
This PR was because it was a deprecation. If it's not significant enough, since it's not core functionality, this can be closed.

hugovk (approved):

PR #5174, opened 2021-01-02
Fix for Read Overflow in PCX Decoding

radarhere:
Fixes CVE-2020-35655, by no longer trusting the size provided by the image for a buffer.


radarhere:
I just found that this bug made the Atheris Hall of Fame - https://github.com/google/atheris/blob/54accd93d19386f6217f4d5ec8d13c94af27538f/hall_of_fame.md

PR #5181, opened 2021-01-04
Docs: add alpha channel to supported hex color specifiers

macjohnny:
Updates the docs with an example how to specify hexadecimal rgb color values with an alpha channel / opacity.
Related code:

https://github.com/python-pillow/Pillow/blob/d374015504df2c7f0316154deade1bc1fa5075df/src/PIL/ImageColor.py#L48-L65

### DIFF
--- a/docs/reference/ImageColor.rst
+++ b/docs/reference/ImageColor.rst
@@ -16,8 +16,17 @@ Color Names
 
 The ImageColor module supports the following string formats:
 
-* Hexadecimal color specifiers, given as ``#rgb`` or ``#rrggbb``. For example,
-  ``#ff0000`` specifies pure red.
+* Hexadecimal color specifiers, given as ``#rgb``, ``#rgba``, ``#rrggbb`` or ``#rrggbbaa``, 
+  with the following placeholders: 
+    - ``r``: red
+    - ``g``: green
+    - ``b``: blue
+    - ``a``: alpha / opacity (can be used in combination with ``mode="RGBA"`` in :py:mod:`~PIL.ImageDraw`)
+  
+  Examples:
+   - ``#ff0000`` specifies pure red.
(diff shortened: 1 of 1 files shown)


radarhere:
I've created https://github.com/macjohnny/Pillow/pull/1 to fix the failing doccheck.

macjohnny:
@radarhere thanks for the fix, the tests are now fine (except for the code coverage tests, but this should be unrelated)

macjohnny:
@radarhere is there anything I need to do before this PR can be merged?

radarhere:
"need? No, all tests pass, apart from coverage (which is irrelevant to this change). The real obstacle is just a matter of myself or one of the other Pillow members sitting down and reviewing.

I've created https://github.com/macjohnny/Pillow/pull/2 with some of my personal thoughts.

macjohnny:
> I don't understand 'can be used in combination with mode="RGBA" in ImageDraw'. I find this vague - which method in ImageDraw?

Here I wanted to refer to the mode parameter in the `ImageDraw.Draw()` method
https://pillow.readthedocs.io/en/stable/reference/ImageDraw.html#PIL.ImageDraw.Draw
as an example where alpha opacity can be taken into account.
If you have any suggestion how to add a comment in a more general way, let me know.

radarhere:
`ImageDraw.Draw()` can accept an "RGBA" mode, yes, and other `ImageDraw` methods can accept translucent colors as `fill` or `outline`.

But you don't need to use `ImageColor` first. I could pass a hexadecimal color specifier to it directly.
```python
d = ImageDraw.Draw(im, "RGBA")
d.ellipse((0, 0, 100, 100), "#f005")
im.show()
```

So I find documenting that `ImageColor` can be used as a converter before using `ImageDraw` to be odd.

I'm going to merge this is in now. If you would like to try and add a different reference to ImageDraw, you can create another PR with the suggestion.

PR #5183, opened 2021-01-05
Only draw each rectangle outline pixel once

### LINKED ISSUE #5182: Incorrect border color in corners when drawing rectangles with border color with opacity < 1
### What did you do?
Draw a rectangle with a color with opacity < 1 for border (e.g. `outline='#0000ffaa'`)

The border color should be homogenous, and not altered by the `fill` color, if set.

### What actually happened?
image
The border color is different in the corners, probably due to overlapping line segments.
Moreover, when a `fill` color is set, the border color is changed due to the overlap of the border and fill rectangle.

* OS: Windows 10
* Python: 3.8
* Pillow: 8.0.1

#### Code
```python
import PIL.Image
import PIL.ImageDraw

image = PIL.Image.new(mode='RGB', size=(500, 500), color='#ffffff')
draw = PIL.ImageDraw.ImageDraw(image, mode='RGBA')
draw.rectangle(xy=((100,100), (200,200)),
               fill='#ffff00',
               outline='#0000ffaa',
               width=10)
draw.rectangle(xy=((205,100), (305,200)),
               outline='#0000ffaa',
               width=10)

image.save('image.png')
```
#### Result
image

### Related Sources
https://github.com/python-pillow/Pillow/blob/d374015504df2c7f0316154deade1bc1fa5075df/src/PIL/ImageDraw.py#L255-L258
https://github.com/python-pillow/Pillow/blob/d374015504df2c7f0316154deade1bc1fa5075df/src/_imaging.c#L3171-L3180
https://github.com/python-pillow/Pillow/blob/d374015504df2c7f0316154deade1bc1fa5075df/src/libImaging/Draw.c#L724-L729

radarhere:
Resolves #5182

`ImagingDrawRectangle` draws four individual lines to make up a rectangle, including each corner twice.
This becomes apparent when the lines are drawn with partial opacity, and the repeated drawing of the corner pixels changes the final opacity.

This PR updates the function to only draw each corner pixel once, similar to #4333


macjohnny (approved):
Thanks for the quick fix!

hugovk:
@radarhere Thanks for the fix!

@macjohnny Thanks for the report!

PR #5184, opened 2021-01-05
Fixed comparison warning

radarhere:
https://github.com/python-pillow/Pillow/runs/1649086275#step:9:118
> src/libImaging/TiffDecode.c:241:22: warning: comparison of integers of different signs: 'int' and 'uint32' (aka 'unsigned int') [-Wsign-compare]
    if (state->xsize != img.width || state->ysize != img.height) {
        ~~~~~~~~~~~~ ^  ~~~~~~~~~
src/libImaging/TiffDecode.c:241:51: warning: comparison of integers of different signs: 'int' and 'uint32' (aka 'unsigned int') [-Wsign-compare]
    if (state->xsize != img.width || state->ysize != img.height) {
                                     ~~~~~~~~~~~~ ^  ~~~~~~~~~~
src/libImaging/TiffDecode.c:582:41: warning: comparison of integers of different signs: 'int' and 'unsigned int' [-Wsign-compare]
                for (tile_y = 0; tile_y < current_tile_length; tile_y++) {
                                 ~~~~~~ ^ ~~~~~~~~~~~~~~~~~~~
3 warnings generated.

### DIFF
--- a/src/libImaging/TiffDecode.c
+++ b/src/libImaging/TiffDecode.c
@@ -503,8 +503,8 @@ ImagingLibTiffDecode(
     isYCbCr = photometric == PHOTOMETRIC_YCBCR;
 
     if (TIFFIsTiled(tiff)) {
-        INT32 x, y, tile_y;
-        UINT32 tile_width, tile_length, current_tile_length, current_line,
+        INT32 x, y, tile_y, current_tile_length;
+        UINT32 tile_width, tile_length, current_line,
             current_tile_width, row_byte_size;
         UINT8 *new_data;
 


wiredfool:
I'm not sure that this is the right approach for a couple of reasons. 

1) We should probably not be using the U/INT32 for these in general, they should really be tsize_t or whatever they are in libtiff, because they change over time. Some of the libtiff items have definitely changed from u/int32 -> u/int64 over the time that I've been digging into valgrind. 

2) If we're casting like that, we need to make sure that we're not going to under/overflow in the cast. Basically, it's silencing the warning, but it doesn't actually change the behavior.

radarhere:
Thanks.
I've updated to remove the casting change, and kept the second change.

radarhere:
This looks like it became part of #5364.

PR #5194, opened 2021-01-09
PyModule_AddObject fix for Python 3.10

### LINKED ISSUE #5193: Python 3.10: segfault due to invalid pointer
On 3.10 alpha 4, pillow fail to run and segfault with "invalid pointer". All projects depending on Pillow also fail to build.

```
python3 -c 'from PIL import Image'
free(): invalid pointer
Aborted (core dumped)
```

* OS: Fedora Linux Rawhide
* Python: 3.10 alpha4
* Pillow: 8.1.0

radarhere:
Resolves #5193

The first commit adds `python3 -c "from PIL import Image"` to .ci/test.sh - for whatever reason, this fails in Python 3.10, while our pytest command does not.

The second commit replaces instances like

```c
#ifdef HAVE_XCB
    PyModule_AddObject(m, "HAVE_XCB", Py_True);
#else
    PyModule_AddObject(m, "HAVE_XCB", Py_False);
#endif
```

with

```c
    PyObject *have_xcb;
#ifdef HAVE_XCB
    have_xcb = Py_True;
#else
    have_xcb = Py_False;
#endif
    Py_INCREF(have_xcb);
    PyModule_AddObject(m, "HAVE_XCB", have_xcb);
```

This is per the example in https://docs.python.org/3/c-api/module.html#c.PyModule_AddObject, which calls `Py_INCREF` before `PyModule_AddObject`.

### DIFF
--- a/.ci/test.sh
+++ b/.ci/test.sh
@@ -2,4 +2,6 @@
 
 set -e
 
+python3 -c "from PIL import Image"
+
 python3 -bb -m pytest -v -x -W always --cov PIL --cov Tests --cov-report term Tests
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -45,6 +45,7 @@ test_script:
 - cd c:\pillow
 - '%PYTHON%\%EXECUTABLE% -m pip install pytest pytest-cov'
 - c:\"Program Files (x86)"\"Windows Kits"\10\Debuggers\x86\gflags.exe /p /enable %PYTHON%\%EXECUTABLE%
+- '%PYTHON%\%EXECUTABLE% -c "from PIL import Image"'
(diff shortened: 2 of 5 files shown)


.ci/test.sh
```diff
@@ -2,4 +2,6 @@
 
 set -e
 
+python3 -c "from PIL import Image"
```
hugovk, line 5: Should we also do this in test-docker.yml and test-windows.yml?

radarhere: Fair enough. I've pushed a commit for test-windows.yml - but the change doesn't go in test-docker.yml, so I've created https://github.com/python-pillow/docker-images/pull/104

hugovk (approved):

PR #5198, opened 2021-01-09
Add CIFuzz Github Action

wiredfool:
Draft, to trigger the CI.

This PR adds CI support to run the OSS-Fuzz routines on fresh changes.

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,24 @@
+name: CIFuzz
+on: [push,pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
+      id: build
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'pillow'
+        dry-run: false
+    - name: Run Fuzzers
+      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'pillow'
+        fuzz-seconds: 600
+        dry-run: false
+    - name: Upload Crash
+      uses: actions/upload-artifact@v1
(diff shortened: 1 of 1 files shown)


wiredfool:
This is failing on a missing item in apt-get, base images need updated.

radarhere:
What's the missing apt-get item?

wiredfool:
I've fixed it upstream, but there are other errors, I need to track down why it's not finding the fuzzers.

wiredfool:
Ok, this is blocked upstream because there isn't actually python support for cifuzz, and it's unclear if it's a small or large problem. https://github.com/google/oss-fuzz/issues/5195

radarhere:
For the record, https://github.com/google/oss-fuzz/issues/5195 has been theoretically resolved.

wiredfool:
Ok, current status here is that 
* the fuzzer is working now
* [ ] Need to get creds to upload the repro cases
* [ ] Fail on a crash to prevent PRs from progressing
* [ ] Remove .yml change so that we're just running on .c/h changes.   (Tho, I will say that ~20% of the security bugs wind up getting fixed with a python change, but I'm not really into adding a 20 minute runtime check to all of our test runs)

hugovk:
> * [ ]  Remove .yml change so that we're just running on .c/h changes.   (Tho, I will say that ~20% of the security bugs wind up getting fixed with a python change, but I'm not really into adding a 20 minute runtime check to all of our test runs)

Add something like this?

```yml
on:
  push:
    paths:
    - '**.c'
    - '**.h'
```

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths

wiredfool:
Yah, that's not the hard part, just a checkbox as the last thing before the merge.

(and the yml change is actually something I added in the valgrind test, not here. memory is going. )

wiredfool:
Ok, I think I understand what's going on here -- it's only a failure if it fails on the branch but not the original.

Issue #1888, opened 2016-05-05
Tracking Issue for high bit depth multichannel images


wiredfool:
Pillow (and PIL) is currently able to open 8 bit per channel multi-channel images (such as RGB) but is able to open higher bit depth images (e.g. I16, I32, or Float32 images) if they are single channel (e.g., grayscale). 

## Previous References

This has been requested many times: #1828, #1885, #1839, #1602, and farther back.

## Requirements

- We should be able to support common GIS formats as well as high bit depth RGB(A) images. 
- At least 4 channels, but potentially more (see #1839)
- Different pixel formats, including I16, I32, and Float.

## Background Reference Info

The rough sequence for image loading is:

- Image file is opened
- Each of the ImagePlugin _accept functions have a chance to look at the first few bytes to determine if they should attempt to open the file
- The `*ImagePlugin._open` method is called giving the image plugin a chance to read more of the image and determine if it still wants to consider it a valid image of it's particular type. If it does, it passes back a tile definition which includes a decoder and an image size. 
- If there is a successful _open call, at some point later `*ImagePlugin._load` may be called on the image, which runs the decoder producing a set of bytes in a raw mode. This is where things like compression are handled, but the output of the decoder is not necessarily what we're storing in our internal structures. 
- The image is unpacked (`Unpack.c`) from the raw mode (e.g. `I16;BS`) into a storage (`Storage.c`) mode (`I`).
- It's now possible to operate on the image (e.g. crop, pixel access, etc)

 There are 3 (or 4) image data pointers, as defined in Imaging.h:
```
struct ImagingMemoryInstance {

    /* Format */
    char mode[IMAGING_MODE_LENGTH];	/* Band names ("1", "L", "P", "RGB", "RGBA", "CMYK", "YCbCr", "BGR;xy") */
    int type;		/* Data type (IMAGING_TYPE_*) */
    int depth;		/* Depth (ignored in this version) */
    int bands;		/* Number of bands (1, 2, 3, or 4) */
    int xsize;		/* Image dimension. */
    int ysize;

    /* Colour palette (for "P" images only) */
    ImagingPalette palette;

    /* Data pointers */
    UINT8 **image8;	/* Set for 8-bit images (pixelsize=1). */
    INT32 **image32;	/* Set for 32-bit images (pixelsize=4). */

    /* Internals */
    char **image;	/* Actual raster data. */
    char *block;	/* Set if data is allocated in a single block. */

[...]

Issue #569, opened 2014-03-25
Parameters for Image.open()


al45tair:
There are a few file formats that would benefit from the ability to pass parameters through the `Image.open()` method. This needs to be thought about carefully because there are clearly some parameters that are generic (e.g. `mode`, `size` and `dpi` arguments, which could cause conversion or resizing, all of which can in some instances be supported directly by the image file plugin), while others are very format specific (e.g. the number of quality layers to load in a JPEG 2000 image).

As of today, some image file plugins are essentially relying on the laziness of PIL’s loading machinery so that the user can call `open` and subsequently set some properties on the returned object before `load` gets called. This is, IMO, pretty nasty.

(I might be up for fixing this myself, but I want to file a bug report before doing any coding because I want the opinions of others on how this should work.)

homm:
> We already have the mode argument

This is not truth. Mode is not "image mode", this is "file read mode" and it is always "r". First lines of `open()`:

```python
if mode != "r":
    raise ValueError("bad mode %r" % mode)
```

I strongly disagree about passing `mode` and `size` in constructor because:

```
There should be one-- and preferably only one --obvious way to do it.
Although that way may not be obvious at first unless you're Dutch.
```

Currently that way is opening image and calling `draft()`. This have many advantages: you can open image, check it dimensions, mode, some options and even verify sometimes before loading an image into memory.

I agree what there should be some way to configure decoder, but `format_args={'JPEG2000': {'layers': 4}}` looks too ugly. Maybe something like this?

```python
i = Image.open(image)
i.draft(mode, size)
i.configure_decoder(jpeg2000={'layers': 4})
```

al45tair:
I'm not sure I understand your point about the `mode` argument. Maybe you’re confused and think I was proposing using it for the image mode, but indeed I was not.

As far as the `draft` method goes, it looks like you’re right that that’s the way to go here — however, this is not documented particularly well and the documentation could do with an overhaul in this regard. Right now there is code out there that is not using `draft` but instead pokes about directly in the Image object to set the desired mode and size.

I like the `configure_decoder` method also; it has less visual noise than using an explicit dictionary.

radarhere:
I created #1349 as a suggestion, but did not progress due to lack of a specific use for the API.

Shivahk027:
im = Image.open(fp, mode='r', format_args={ 'JPEG2000': { 'layers': 4 } })
in this what is fp? ,what is the value for fp?is that mandatory?

[...]

PR #5201, opened 2021-01-11
Add AVIF plugin (decoder + encoder using libavif)

fdintino:
This adds support for AVIF encoding and decoding, including AVIF image sequences. 

I've added tests, and integrated libavif into the windows, linux, and mac CI builds. I haven't done anything to integrate with the docker-images repo.

I chose libavif rather than libheif because the former has been embraced by AOMedia and it's what Chromium uses. Packaging support is spotty at the moment, but I expect that to change soon (currently it's in Debian testing, Fedora rawhide, Ubuntu hirsute, and Alpine edge).

I haven't written any documentation yet. I wanted to get feedback on the general direction before I took that on.

A few notes on the implementation here:

- The plugin currently only supports encoding 8-bit images, and all images are decoded to 8-bit RGB(A). I wasn't totally clear on how best to deal with higher bit depths (somewhat related issue: #1888)
- The RGB to YUV conversion isn't exposed to python at all. Chroma subsampling and the presence of an alpha channel make it non-trivial to return decoded images as YCbCr. It would not be difficult to permit encoding from a YCbCr source.
- Since there isn't a way to pass parameters to Image.open (#569), I'm using module globals in AvifImagePlugin.py to make decoder codec choice and chroma upsampling configurable. I suspect there's a better way to do this.

The star.avifs test file is licensed as [CC-BY](https://github.com/link-u/avif-sample-images#twinkle-star---image-sequence-test)

I linted the C code with the new clang-format settings, but made the following change so that it didn't make `PyObject_HEAD` and the threading macros look wonky:

```diff
diff --git a/.clang-format b/.clang-format
index be32e6d1..300f8e54 100644
--- a/.clang-format
+++ b/.clang-format
@@ -18,3 +18,7 @@ SpaceBeforeParens: ControlStatements
 SpacesInParentheses: false
 TabWidth: 4
 UseTab: Never
+StatementMacros:
+  - PyObject_HEAD
+  - Py_BEGIN_ALLOW_THREADS
+  - Py_END_ALLOW_THREADS
```


Tests/helper.py
```diff
@@ -206,6 +207,7 @@ def _test_leak(self, core):
         start_mem = self._get_mem_usage()
         for cycle in range(self.iterations):
             core()
+            gc.collect()
```
radarhere, line 210: Did you want to talk about why you added this?

fdintino: I accidentally left this in here while I was debugging. I'll remove it.

src/_avif.c
```diff
@@ -0,0 +1,819 @@
...
+    if (strcmp(mode, "RGBA") == 0) {
+        rgb.format = AVIF_RGB_FORMAT_RGBA;
+        channels = 4;
+    } else {
+        rgb.format = AVIF_RGB_FORMAT_RGB;
+        channels = 3;
+    }
+
+    avifRGBImageAllocatePixels(&rgb);
+    memcpy(rgb.pixels, rgb_bytes, size);
```
wiredfool, line 339: Please document in a comment that this is safe for r/w, and potentially add an explict check that the rgb_bytes/rgb.pixels is large enough.

src/_avif.c
```diff
@@ -0,0 +1,819 @@
...
+    self->decoder = NULL;
+    self->size = size;
+    self->data = PyMem_New(uint8_t, size);
+    if (self->data == NULL) {
+        PyErr_SetString(PyExc_MemoryError, "PyMem_New() failed");
+        Py_DECREF(self);
+        return NULL;
+    }
+
+    memcpy(self->data, avif_bytes, size);
```
wiredfool, line 482: Document here as well.

src/_avif.c
```diff
@@ -0,0 +1,819 @@
...
+    if (result != AVIF_RESULT_OK) {
+        PyErr_Format(
+            exc_type_for_avif_result(result),
+            "Conversion from YUV failed: %s",
+            avifResultToString(result));
+        avifRGBImageFreePixels(&rgb);
+        return NULL;
+    }
+
+    size = rgb.rowBytes * rgb.height;
```
wiredfool, line 620: Is this guaranteed to not overflow, even in the face of invalid input?

.ci/install.sh
```diff
@@ -19,7 +19,8 @@ set -e
 
 sudo apt-get -qq install libfreetype6-dev liblcms2-dev python3-tk\
                          ghostscript libffi-dev libjpeg-turbo-progs libopenjp2-7-dev\
-                         cmake imagemagick libharfbuzz-dev libfribidi-dev
+                         cmake imagemagick libharfbuzz-dev libfribidi-dev\
+                         ninja-build build-essential nasm
```
wiredfool, line 23: I believe build-essential is already installed

.ci/install.sh
```diff
@@ -29,6 +30,7 @@ python3 -m pip install -U pytest
 python3 -m pip install -U pytest-cov
 python3 -m pip install pyroma
 python3 -m pip install test-image-results
+python3 -m pip install meson
```
wiredfool, line 33: Does meson need to be added to the requirements.txt?

Tests/test_file_avif.py
```diff
@@ -0,0 +1,537 @@
...
+
+
+if hasattr(os, "sched_getaffinity"):
+    MAX_THREADS = len(os.sched_getaffinity(0))
+else:
+    MAX_THREADS = os.cpu_count()
+
+
+@skip_unless_feature("avif")
+class TestAvifLeaks(PillowLeakTestCase):
```
wiredfool, line 524: I'd prefer not iterating a leak test in the standard test suite, as that can be expensive from a time POV. It's ok for the initial cut, but I'd rather not have it long term.

wiredfool (changes requested):

nulano:
Adding libavif to MSYS2 fails to compile due to a few missing defines (`AVIF_CHROMA_UPSAMPLING_AUTOMATIC`, `AVIF_CHROMA_UPSAMPLING_FASTEST`, `AVIF_CHROMA_UPSAMPLING_BEST_QUALITY`): https://github.com/nulano/Pillow/runs/1683386633?check_suite_focus=true#step:5:77

.ci/install.sh
fdintino: I don't think so. It's a build dependency of dav1d so it's used when building libavif in CI, but it's not otherwise required to run tests.

fdintino:
@nulano it looks like those defines were only added in libavif 0.8.3. I'll figure some `#if` version checks around their usage.

src/_avif.c
fdintino: libavif currently restricts images to a maximum of 2^28 pixels. If the dimensions are larger than 16384x16384 then the function that sets `decoder->image->width` and `decoder->image->height` fails. So I suppose that a 4-channel 16384x16384 8-bit image could overflow on a 32-bit platform. I'm not certain because the codecs used by libavif have their own overflow limit checks. For instance, dav1d enforces a maximum of 2^26 pixels on 32-bit systems. Should I add a check against `PY_SSIZE_T_MAX` to be sure?

Tests/helper.py
fdintino: Actually, I realized now why I added this: without it the leak tests are non-deterministic. I could pad the memory limit to counteract the fact that it may not have hit the gc limit, but forcing garbage collection after each iteration ensures that the test is deterministic.

src/_avif.c
fdintino: Added here

https://github.com/python-pillow/Pillow/blob/b84a8e0d47d9e8807d207aba9c9eeb455646ef62/src/_avif.c#L619-L622

src/_avif.c
fdintino: I wasn't entirely sure what you wanted documented for this line. I added this, let me know if it's what you had in mind:

https://github.com/python-pillow/Pillow/blob/b84a8e0d47d9e8807d207aba9c9eeb455646ef62/src/_avif.c#L484-L485

fdintino:
@nulano Is it okay if I cherry-pick your MSYS commit into this PR?

src/_avif.c
wiredfool: Basically, I'm the one who will get a CVE on this if there's a problem, and I'd like really clear guidelines about what the assumptions are for sizes of things and where they come from for dangerous operations like memset, malloc, and pointer reads/writes.  This isn't so much for now, but a couple years down the line, things need to be clear. This will be fuzzed, this will be run under valgrind, so hopefully there won't be problems. 

I've basically had to reverse engineer how SgiRleDecode works over the last month or so, and I'd like to be preventing that sort of experience in the future.

nulano (commented):
> @nulano Is it okay if I cherry-pick your MSYS commit into this PR?

Of course, cherry-pick away!

I have a few nitpicks for `winbuild/build_prepare`, I haven't looked at the rest yet.

src/_avif.c
fdintino: Does raising a `MemoryError` if `rgb.height > PY_SSIZE_T_MAX / row_bytes` (as I have in the latest PR push) suffice to address that concern?

winbuild/build_prepare.py
```diff
@@ -312,6 +317,42 @@ def cmd_msbuild(
...
+                [
+                    "-DBUILD_SHARED_LIBS=OFF",
+                    "-DAVIF_CODEC_AOM=ON",
+                    "-DAVIF_LOCAL_AOM=ON",
+                    "-DAVIF_CODEC_DAV1D=ON",
+                    "-DAVIF_LOCAL_DAV1D=ON",
+                ],
+                "..",
+            ),
+            cmd_nmake(),
```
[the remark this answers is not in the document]
fdintino: How about `cmd_rmdir("build")` before the `cmd_mkdir("build")` instead?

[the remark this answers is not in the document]
nulano: That's fine.

fdintino:
@radarhere @wiredfool @nulano I think I've addressed all feedback (except for the requests for docs on building), but I've left it up to you all to resolve conversations (or not).

Is this PR generally on the right track? I've held off on writing docs until I've gotten a signal one way or the other.

src/_avif.c
fdintino: I was able to avoid a memcpy here by having `PyArg_ParseTuple` pass in a `PyBytesObject` and incrementing the reference in the new / decrementing in the dealloc. That also avoids an unnecessary malloc during decoding.

fdintino:
Since it's been a month since I asked my question without response, I'll try to reframe it as more specific questions that might be more answerable.

1. Is the general structure of this plugin acceptable? I tried to hew closely to the conventions elsewhere in the repo, so I assume so, but would appreciate confirmation.
2. Is the test coverage sufficient? The gaps are all in the error handling. I'd be happy to try to add test cases for those to make it more complete.
3. I'd appreciate feedback on the encoder settings. For instance: should `yuv_format` be renamed `subsampling` to be consistent with the Jpeg plugin? Should I offer a jpeg-like 0-100 `quality` setting that maps the min/max quantizer? (The colorist library has [such a quality setting](https://github.com/joedrago/colorist/blob/0f6f274bbe5828e14e42ad4f5d03c97f3826373e/lib/src/format_avif.c#L284-L292) that still allows qmin/qmax as an advanced override option). Should I eliminate any options? (My vote would be to get rid of `qmin_alpha` and `qmax_alpha`).
4. What would you like to see, CI-wise, with this pull request? While waiting for feedback I worked on putting [this plugin in its own package](https://github.com/fdintino/pillow-avif-plugin), which builds manylinux wheels in the same manner as the pillow-wheels repo. I setup builds for all the codecs supported by libavif, on all platforms, and also added cached dependencies (see [pillow-avif-plugin-depends](https://github.com/fdintino/pillow-avif-plugin-depends)). That includes a crate vendor tarball for rav1e, as I've found crates.io to be too unreliable for frequent CI builds. Would you want to wait until this PR is wrapped up and merged before I opened a pull request against pillow-wheels?
    - Note: I think licensing shouldn't be an issue for anything: AOM, rav1e, dav1d, SVT-AV1 and libyuv are all BSD-2 licensed, libgav1 is Apache, and they all are covered under the Alliance for Open Media Patent License.
    - It might be overkill to include all codecs in the manylinux and windows wheels. In particular, SVT-AV1 is far from ready for prime-time. My recommendation would be to include AOM, dav1d, and rav1e by default. AOM, being the reference implementation, is the most complete, while dav1d and rav1e are the fastest (putting aside SVT-AV1).

Tests/test_file_avif.py
```diff
@@ -0,0 +1,683 @@
...
+
+    def test_no_resource_warning(self, tmp_path):
+        with Image.open(TEST_AVIF_FILE) as image:
+            temp_file = str(tmp_path / "temp.avif")
+            pytest.warns(None, image.save, temp_file)
+
+    def test_file_pointer_could_be_reused(self):
+        with open(TEST_AVIF_FILE, "rb") as blob:
+            Image.open(blob).load()
+            Image.open(blob).load()
```
radarhere, line 161:
```suggestion
            with Image.open(blob) as im:
                im.load()
            with Image.open(blob) as im:
                im.load()
```
Using context managers, as per https://pillow.readthedocs.io/en/stable/releasenotes/7.0.0.html#image-del

Tests/test_file_avif.py
```diff
@@ -0,0 +1,683 @@
...
+        with Image.open("Tests/images/chi.gif") as im:
+            original_value = im.convert("RGB").getpixel((1, 1))
+
+            # Save as AVIF
+            out_avif = str(tmp_path / "temp.avif")
+            im.save(out_avif, save_all=True)
+
+        # Save as GIF
+        out_gif = str(tmp_path / "temp.gif")
+        Image.open(out_avif).save(out_gif)
```
radarhere, line 173:
```suggestion
        with Image.open(out_avif) as im:
            im.save(out_gif)
```

radarhere:
This might be a libavif bug, but I find that if I run this PR, libavif has [stopped working](https://github.com/radarhere/Pillow/runs/2201531959#step:9:525) for macOS.

https://github.com/radarhere/Pillow/runs/2201531959#step:8:1174
> /Users/runner/work/Pillow/Pillow/depends/libavif-0.8.4/ext/libyuv/include/libyuv/row.h:750:5: error: 'LIBYUV_UNLIMITED_DATA' is not defined, evaluates to 0 [-Werror,-Wundef]
> #if LIBYUV_UNLIMITED_DATA
>     ^
> 1 error generated.

wiredfool:
We're going to need to add the required libraries to the docker images as well, and we're going to need to add these to the oss-fuzz builder to get fuzzer support.  

Might as well make a PR to the Pillow-wheels for whatever needs to happen on build. That will also be potentially helpful for getting the dependencies into oss-fuzz.

jcea:
I am looking forward to have AVIF support in Pillow. What are the showstoppers to merge this PR?.

Thanks for your work!!

vanchung1995:
Hello everybody. I'm wondering that how I can custom avif plugin to get return value instead of saving to file?
And why in some case, image in avif has larger size (in KB) than origin image like the bellow image. Origin image has 12.2kb but avif image has 29.7KB??
I used this code: 
Screenshot from 2021-09-08 16-25-14


c58755e537a7def987b6

radarhere:
As a general answer, if you would like Pillow to place the image contents in a variable rather than saving to file,
```python
from PIL import Image
import io
fp = io.BytesIO()
im.save(fp, 'PNG')
print(fp.getvalue())
```

vanchung1995:
> As a general answer, if you would like Pillow to place the image contents in a variable rather than saving to file,
> 
> ```python
> from PIL import Image
> import io
> fp = io.BytesIO()
> im = Image.new("RGB", (100, 100))
> im.save(fp, 'PNG')
> print(fp.getvalue())
> ```

Thanks so much. But what about the size of image? I realize that all my 300*200 images in jpeg has double size when convert to avif? How to custom to decrease size of images?

fdintino:
This isn't really the right forum for these questions. That said: in general AVIF images will compress smaller than JPEG, but that might not always be the case for all images. There are docs included in this pull request listing the settings you can pass to `.save()`. You might find that decreasing `quality` will give you the desired file size. You can find those docs here:

https://github.com/python-pillow/Pillow/blob/66660cad793d43e6ed6a8dd7d9dd722065ead676/docs/handbook/image-file-formats.rst#avif

barboteur:
Hello,

I've just made some tests and effectively, I have to use a quality of 40 in the AVIF format to get a smaller size than WEBP with the default quality used by pillow.
And with these quality settings, the image is better in AVIF than in WEBP.

vanchung1995:
thanks so much all. I tried quality param in save method and I realize same as @barboteur . 40 is a better score

Tests/check_avif_leaks.py
```diff
@@ -0,0 +1,42 @@
...
+test_file = "Tests/images/avif/hopper.avif"
+
+pytestmark = [
+    pytest.mark.skipif(is_win32(), reason="requires Unix or macOS"),
+    skip_unless_feature("avif"),
+]
+
+
+def test_leak_load():
+    from resource import RLIMIT_AS, RLIMIT_STACK, setrlimit
```
mara004, line 22: Why are you importing this inside a function, and that twice (l. 22, l. 32)? I'd rather move this import to the top...

fdintino: That's a good question. This was copied from the leak test for jpeg2000 [here](https://github.com/python-pillow/Pillow/blob/95cff6e959bb3c37848158ed2145d49d49806a31/Tests/check_j2k_leaks.py#L22). I suspect this was done because those imports will fail on windows (judging by the skip pytest mark above it). But those could be handled at the top of the file with an `except ImportError`

fdintino:
A quick update on this pull request: I haven't updated or resolved the merge conflicts in the pillow-depends or pillow-wheels repositories because I haven't received any comments here for over a year. Though it's promising to see that it is still occasionally merged against the main branch.

As I mentioned in my previous comment, until this gets merged I will be maintaining this as a separate plugin ([pillow-avif-plugin](https://github.com/fdintino/pillow-avif-plugin)). The CI process there is based off of the one in pillow-wheels. I've made an attempt to keep up with changes there (by, for instance, adding musllinux wheels), but it's not identical so it would need some tweaks before those changes could be incorporated into python-pillow/pillow-wheels#193

jcea:
I am very interested in getting AVIF support in stock pillow.

NewUserHa:
.AVIF support lossless.
Does this support `lossless = True` like .webp, and `loop = 1`?

docs/installation.rst
```diff
@@ -308,11 +330,11 @@ Many of Pillow's features require external libraries:
 
         sudo pkg install python3
 
-    Prerequisites are installed on **FreeBSD 10 or 11** with::
+    Prerequisites are installed on **FreeBSD 11 or 12** with::
```
radarhere, line 333: Why update the versions here? Does libavif not work on FreeBSD 10?

fdintino: FreeBSD 10.4 was EoL in 2018 and libavif didn't even exist until 2020. I just tried running FreeBSD 10 in virtualization and it appears that https://pkg.freebsd.org/ no longer even hosts the 10.x binaries, so it's not even possible to get `pkg` running without hunting for a mirror that has kept the old versions around. If someone _does_ still have FreeBSD 10 running they would presumably be able to install libavif through ports. But the `pkg install` command below would not work.

fdintino:
@radarhere since you've been pushing to this branch with upstream merges (thanks for that, by the way) I wanted to give you a heads up that I would like to update this PR to have parity [with the fixes and changes I've made to pillow-avif-plugin](https://github.com/fdintino/pillow-avif-plugin/blob/master/CHANGELOG.rst) so that we don't step on each others' toes. I think there are other changes in that repository that would address the CI failures in the latest commits.

vrabaud:
Hi, I am a libavif developer: do you need any help on that PR ?

vrabaud:
@fdintino , libavif could use those artefacts for other projects so that would be welcome. Then again, couldn't caching be used here ?

fdintino:
@vrabaud I have caching for pillow-avif-plugin, but many of my commits are upgrades for libavif or one of its transitive dependencies, which invalidates the caches. With caching there would still be a maintenance burden of compiling and upgrading libavif, rav1e, SVT-AV1, etc. And even if it makes sense to take on that burden for the wheel release workflows, there is still utility in being able to use pre-built libraries for the workflows that run tests.

vrabaud:
@fdintino , that makes sense. We'll gladly accept a PR for storing artefacts then !

stalkerg:
Can somebody from the core team explain why this very important PR was not reviewed and didn't merge? 
What we (the community) can do to unlock it? 
@wiredfool

aclark4life:
> With caching there would still be a maintenance burden of compiling and upgrading libavif, rav1e, SVT-AV1, etc. And even if it makes sense to take on that burden for the wheel release workflows, there is still utility in being able to use pre-built libraries for the workflows that run tests.

@stalkerg Looks like a WIP where the benefit of including in Pillow is still TBD, with specific reference to "maintenance burden". At this point I'd say it's up to @radarhere and @fdintino to find a way forward, if they still believe that forward is the right direction. @wiredfool may have some additional comments. For onlookers, maybe a review of adding AVIF support to Pillow at a high level would help. How strong is the demand? How many are willing to help support it? is the maintenance burden manageable? If the results of the high level review indicate more than just you, @fdintino, @vrabaud (thanks!) and @radarhere are interested in this, that would be good to know.

jcea:
I am very interested in AVIF support and waiting for it for years. So +1 for me.

cgohlke:
Besides the maintenance burden, including AVIF would double/triple the size of the Pillow wheels, which might or might not be relevant. I ran into that issue with the [imagecodecs](https://pypi.org/project/imagecodecs/) package, which bundles libavif.

mara004:
What is the advantage of supporting AVIF in Pillow in the first place, rather than through an external plugin?

fdintino:
I don't want to overstate the maintenance burden. I'm a single person and it hasn't been terribly difficult for me to stay on top of libavif updates for pillow-avif-plugin. I'll probably continue to do that for a while—at least as long as I'm working at The Atlantic. We use thumbor, and have it configured to auto-transcode images to AVIF for requests that Accept image/avif. But I certainly wouldn't mind help.

What makes libavif unusual compared to the other pillow dependencies is that it doesn't actually have its own AVIF codec. There are several different AVIF encoding and decoding libraries, that offer different features and have varying performance and compression quality. The value of libavif is the ability to have a single unified API to use AOM, or rav1e, or dav1d, depending on the requirements. I've chosen to ship all supported codecs with the pillow-avif-plugin wheel, but I don't think that pillow would have to do the same. If I had to narrow it down, I would recommend rav1e for encoding and dav1d for decoding. This ought to reduce the impact on the wheel size, and make CI maintenance easier.

I think having AVIF in core pillow might give shape to the implementation of features like high bitrate multichannel images (#1888). There are other features (like supporting YCbCr mode with alpha) that would be nice-to-have, and which would be useful in the implementation of support for next-generation image formats like HEIF and JPEF XL.

Yay295:
Would it be possible to check if an AVIF codec is available and use it, instead of bundling any of them?

fdintino:
@Yay295 That is actually what this pull request does, with the exception of the wheels for windows (which are built in this repo, not in pillow-wheels). The discussions around maintenance all have to do with the pillow-wheels repository. But they're not entirely theoretical; the CI for my project pillow-avif-plugin is based on the one from pillow-wheels, and I publish wheels for all the same platforms and architectures as pillow.

@aclark4life If I could correct one thing: this is not a work-in-progress. This pull request was turned into [pillow-avif-plugin](https://github.com/fdintino/pillow-avif-plugin), which I consider to be complete, with a full CI workflow. I know of several organizations that use it in production. The only thing I haven't stayed on top of is resolving conflicts in this pull request. I'm disinclined to invest too much time into that until I have a clearer sense of whether or not it will be merged. Even less so for my now very outdated pillow-wheels PR.

src/_avif.c
fdintino: Realized it would probably be better to have you resolve these conversations, to confirm that the feedback has indeed been addressed.

Yay295:
#7418 might be relevant then. I'm not sure if that changes anything.

stalkerg:
```
avif.lib(codec_rav1e.c.obj) : error LNK2001: unresolved external symbol _rav1e_packet_unref
avif.lib(codec_rav1e.c.obj) : error LNK2001: unresolved external symbol _rav1e_frame_fill_plane
build\lib.win32-cpython-311\PIL\_avif.cp311-win32.pyd : fatal error LNK1120: 17 unresolved externals
```

Maybe there is a version mismatch?

fdintino:
It is an issue with how the libavif static library is built. I’m working on a pull request to that project that will fix it.

fdintino:
I was mistaken. There _were_ issues with how the libavif dependencies were being linked (the fixes for which were just merged in AOMediaCodec/libavif#1579). But the reason that the appveyor build was failing is that the python 3.11 environment is 32-bit. If we care to support this I can open a pull request to rav1e to add a win32 deploy artifact. For the time being I'm skipping libavif setup in 32-bit windows environments.

jnoring-pw:
What's the status of this work? 

A gentle nudge: browser support for AVIF is approaching 95%, and I'd really like to see built-in support in this library. Is there anything I can do to expedite this piece of work so there can be official support?

jcea:
I am very interested too!!!!!

stalkerg:
@wiredfool  which things should be fixed to merge it? Seems like most your requests was addressed.

sdbds:
@wiredfool 
Three years have passed, and seeing the efforts of the contributors, I hope this PR can be merged.

bigcat88:
Just a suggestion: make support for the AFIV format using `libehif` instead of `libavif`
It will be still use `aom` library and `rav1e` like in this PR, so this can be changed right in this PR.

Cons:
  * it has GPL3 license

Pros:
  * in future will be very easy to add suport for HEIC(H.265) and new upcoming VVC(H.266) format
  * there is branch with CUDA support for decoding
  * API tried to be unified between different codecs format(easy maintain)
  * many other nice things and the author are very nice to the community PR's
  
All these will reduce maintenance work to add the new image formats support, imho.

jnoring-pw:
I'd prefer not to make this MR any more late for an image format with questionable performance and poor browser support. HEIC has abysmal support in browsers--it's currently at ~13%, and if anything I see that number decreasing over time.

Also, GPL3 is a big "no thank you" from me. If this MR merges with that license I'll stick with `pillow-avif-plugin`.

gsantner:
./subscribe

hope this gets merged soon and pillow to support more modern formats **by default**.

cefege:
Looking forward to the merge, been waiting for years.

cotwitch:
🙉 YEARS 🙊

aclark4life:
OK folks, we hear you! @fdintino are you still around? Anyone else available to pick this up?

src/_avif.c
```diff
@@ -0,0 +1,1084 @@
...
+    } else if (!strcmp(upsampling_str, "best")) {
+        upsampling = AVIF_CHROMA_UPSAMPLING_BEST_QUALITY;
+    } else if (!strcmp(upsampling_str, "nearest")) {
+        upsampling = AVIF_CHROMA_UPSAMPLING_NEAREST;
+    } else if (!strcmp(upsampling_str, "bilinear")) {
+        upsampling = AVIF_CHROMA_UPSAMPLING_BILINEAR;
+    } else {
+        PyErr_Format(PyExc_ValueError, "Invalid upsampling option: %s", upsampling_str);
+        return NULL;
+    }
```
radarhere, line 756: As GitHub Actions says above,
> variable ‘upsampling’ set but not used [-Wunused-but-set-variable]

So either this partially implemented feature should be removed, or it needs to finish being implemented?

src/PIL/Image.py
```diff
@@ -1548,7 +1548,9 @@ def getexif(self) -> Exif:
 
         # XMP tags
         if ExifTags.Base.Orientation not in self._exif:
-            xmp_tags = self.info.get("XML:com.adobe.xmp")
+            xmp_tags = self.info.get("XML:com.adobe.xmp") or self.info.get("xmp")
+            if isinstance(xmp_tags, bytes):
+                xmp_tags = xmp_tags.decode("utf-8")
```
radarhere, line 1553: Could a test be added for this change?

fdintino:
@aclark4life I'm still around. I only pushed a rebase a week ago. I'll take a look at @radarhere's pull request.

winbuild/build_prepare.py
```diff
@@ -121,6 +121,9 @@ def cmd_msbuild(
     "TIFF": "4.6.0",
     "XZ": "5.6.3",
     "ZLIB": "1.3.1",
+    "MESON": "1.5.1",
```
radarhere, line 124:
```suggestion
    "MESON": "1.5.2",
```

.github/workflows/test-windows.yml
```diff
@@ -86,6 +86,8 @@ jobs:
         choco install nasm --no-progress
         echo "C:\Program Files\NASM" >> $env:GITHUB_PATH
 
+        python -m pip install meson
+
```
radarhere, line 90: Testing, I don't find this to be necessary - https://github.com/radarhere/Pillow/commit/2d4a4de12c8902c184168cd59b5dd7f03cdb6357 / https://github.com/radarhere/Pillow/actions/runs/11400689885

Tests/check_avif_leaks.py
```diff
@@ -0,0 +1,44 @@
...
+
+    setrlimit(RLIMIT_STACK, (stack_size, stack_size))
+    setrlimit(RLIMIT_AS, (mem_limit, mem_limit))
+    for _ in range(iterations):
+        with Image.open(test_file) as im:
+            im.load()
+            test_output = BytesIO()
+            im.save(test_output, "AVIF")
+            test_output.seek(0)
+            test_output.read()
```
radarhere, line 44:
```suggestion
        test_output = BytesIO()
        with Image.open(test_file) as im:
            im.save(test_output, "AVIF")
        test_output.seek(0)
        test_output.read()
```
This seems neater.

aclark4life:
> @fdintino are you taking PayPal donations? I want to send over my love from the Django community.

@codingjoe Can you explain why this is good for Django? Would love to have a good answer I can point other folks too. Thank you 🙏

codingjoe:
> > @fdintino are you taking PayPal donations? I want to send over my love from the Django community.
> 
> @codingjoe Can you explain why this is good for Django? Would love to have a good answer I can point other folks too. Thank you 🙏

Sure, I maintain [django-pictures](https://github.com/codingjoe/django-pictures), a responsive cross-browser image library. I guess every good web engineer wants to serve fast, responsive, high-res images. And since [AVIF](https://caniuse.com/avif) is part of [Baseline 2024](https://web.dev/baseline/2024), folks are eager to get started. I know I am 😁

Tests/test_file_avif.py
```diff
@@ -0,0 +1,830 @@
...
+        temp_file = str(tmp_path / "temp.avif")
+        frame1 = Image.new("RGB", (100, 100))
+        frame2 = Image.new("RGB", (150, 150))
+        with pytest.raises(ValueError):
+            frame1.save(temp_file, save_all=True, append_images=[frame2], duration=100)
+
+    def test_heif_raises_unidentified_image_error(self) -> None:
+        with pytest.raises(UnidentifiedImageError):
+            with Image.open("Tests/images/avif/rgba10.heif"):
+                pass
```
radarhere, line 733: What is the relevance of this test?

Tests/test_file_avif.py
```diff
@@ -0,0 +1,830 @@
...
+        im_png = Image.open(buf_png)
+        buf_out = BytesIO()
+        im_png.save(buf_out, format="AVIF", quality=100)
+
+        assert_image_similar(im_png.convert("RGBA"), Image.open(buf_out), 1)
+
+    def test_decoder_strict_flags(self) -> None:
+        # This would fail if full avif strictFlags were enabled
+        with Image.open("Tests/images/avif/chimera-missing-pixi.avif") as im:
+            assert im.size == (480, 270)
```
radarhere, line 620: Isn't it possible that someone could build libavif with those strict flags enabled, and then build Pillow from source to use it? So in that scenario, this test would fail?

src/PIL/AvifImagePlugin.py
```diff
@@ -0,0 +1,285 @@
...
+            exif_data.load(exif)
+        except SyntaxError:
+            pass
+        else:
+            orientation_tag = next(
+                k for k, v in ExifTags.TAGS.items() if v == "Orientation"
+            )
+            exif_orientation = exif_data.get(orientation_tag) or 0
+
+    xmp = info.get("xmp", im.info.get("xmp") or im.info.get("XML:com.adobe.xmp"))
```
radarhere, line 169: You're saving XMP and EXIF data from the `info` dictionary.

With the exception of JPEG and XMP, potentially to be reversed by #8483, none of other plugins currently do this.

docs/installation/building-from-source.rst
```diff
@@ -210,9 +233,10 @@ Many of Pillow's features require external libraries:
 
     Prerequisites are installed on **FreeBSD 10 or 11** with::
 
-        sudo pkg install jpeg-turbo tiff webp lcms2 freetype2 openjpeg harfbuzz fribidi libxcb
+        sudo pkg install jpeg-turbo tiff webp lcms2 freetype2 openjpeg harfbuzz fribidi libxcb libavif
 
-    Then see ``depends/install_raqm_cmake.sh`` to install libraqm.
+    See ``depends/install_raqm_cmake.sh`` to install libraqm and
+    ``depends/install_libavif.sh`` to install libavif.
```
radarhere, line 239: As a reader, I find this confusing. It makes me wonder if it I should `pkg install libavif` or `depends/install_libavif.sh`.

Tests/test_file_avif.py
fdintino: `strictFlags` are passed at runtime, they're not set at compile time. By default it strictly enforces various specification rules, but two in particular are often found in the wild and so are regularly disabled:

https://github.com/python-pillow/Pillow/blob/8b8bbba0f77c4ed5d1479d3f6e0c1b79dfb4f867/src/_avif.c#L797-L802

[Chromium](https://chromium.googlesource.com/chromium/src/+/1ade7e92edd967c0d54c64f262f9c149b48d0a96/third_party/blink/renderer/platform/image-decoders/avif/avif_image_decoder.cc#660) disables these two flags and [WebKit](https://github.com/WebKit/WebKit/blob/main/Source/WebCore/platform/image-decoders/avif/AVIFImageReader.cpp#L43-L47) disables `AVIF_STRICT_PIXI_REQUIRED`.
The `test_decoder_strict_flags` test verifies that the strictFlags are set appropriately so that these somewhat common images can load without throwing an error.

Tests/test_file_avif.py
fdintino: It is it ended to test this bit of logic:

https://github.com/python-pillow/Pillow/blob/ee3c46af4259858d2b4b018aa68e5b7d424d794e/src/PIL/AvifImagePlugin.py#L36-L44

If the ftyp box is `avif` or `avis` then this is definitely an AVIF image, and so it is sensible to return a string from the accept function to trigger an error. But `mif1` and `msf1` are also valid values for AVIF, but also for HEIF. The initial bytes passed to the accept function is not long enough to determine which format it is. In order to be interoperable with pyheif I cannot raise an error in the accept, and must instead raise a SyntaxError in `_open` if AVIF is not compiled. `_open` would only be called if there isn't another PIL image plugin that accepts HEIF files.

fdintino:
@codingjoe I've never solicited donations for my open-source work, but I certainly wouldn't mind if someone was motivated to give something. The best place to send me personal messages and questions is fdintino@gmail.com, which is also what you can use to find me on PayPal. Thanks!

.github/workflows/test-windows.yml
radarhere: I removed this in https://github.com/fdintino/Pillow/pull/3

winbuild/build_prepare.py
radarhere: I made this change in https://github.com/fdintino/Pillow/pull/3

sheecegardezi:
> > @fdintino are you taking PayPal donations? I want to send over my love from the Django community.
> 
> @codingjoe Can you explain why this is good for Django? Would love to have a good answer I can point other folks too. Thank you 🙏

literly every single embedded mapping library (think google maps) ... atm png and jpeg are defecto formats for the map tiles (chunks)

winbuild/build_prepare.py
```diff
@@ -396,6 +400,57 @@ def cmd_msbuild(
         ],
         "bins": [r"*.dll"],
     },
+    "rav1e": {
+        "url": (
+            f"https://github.com/xiph/rav1e/releases/download/v{V['RAV1E']}/"
+            f"rav1e-{V['RAV1E']}-windows-msvc-generic.zip"
```
radarhere, line 406:
```suggestion
            f"https://github.com/xiph/rav1e/releases/download/v{V['RAV1E']}/FILENAME"
```
See #8498

src/PIL/Image.py
radarhere: A test has been added in https://github.com/fdintino/Pillow/pull/5

winbuild/build_prepare.py
radarhere: This change was made in https://github.com/fdintino/Pillow/pull/5

.github/workflows/wheels-dependencies.sh
```diff
@@ -5,7 +5,7 @@ if [ -z "$IS_MACOS" ]; then
     export MB_ML_LIBC=${AUDITWHEEL_POLICY::9}
     export MB_ML_VER=${AUDITWHEEL_POLICY:9}
 fi
-export PLAT=$CIBW_ARCHS
+export PLAT="${CIBW_ARCHS:-$AUDITWHEEL_ARCH}"
```
radarhere, line 8: Testing, I don't find this change to be necessary. What was your thinking here?

doublex:
Is there a chance to merge this pull request into Pillow 11.1?
Kind regards!

docs/installation/building-from-source.rst
radarhere: I've suggested removing the reference to `install_libavif.sh` from this sentence in https://github.com/fdintino/Pillow/pull/6

src/PIL/AvifImagePlugin.py
```diff
@@ -0,0 +1,294 @@
...
+    if isinstance(exif, Image.Exif):
+        exif = exif.tobytes()
+
+    exif_orientation = 0
+    if exif:
+        exif_data = Image.Exif()
+        try:
+            exif_data.load(exif)
+        except SyntaxError:
+            pass
```
radarhere, line 181: Could you talk about why invalid EXIF data is something to be ignored, and not something that should stop the save operation?

radarhere: I've created https://github.com/fdintino/Pillow/pull/8 to not catch the error.

src/PIL/AvifImagePlugin.py
radarhere: I've created https://github.com/fdintino/Pillow/pull/8 to remove this.

doublex:
@radarhere 
Thank you so much for all of your efforts!

fdintino:
Since this PR is so far along in review, I figure I should run any non-trivial changes past the reviewers before undertaking them:

libaom is a very capable codec, but it has absolutely awful performance and compression efficiency with default settings. This gets reported as an issue to pillow-avif-plugin, and [my advice](https://github.com/fdintino/pillow-avif-plugin/issues/65#issuecomment-2513477035) is usually to use rav1e. How would people feel about changing the order of encoder codecs such that rav1e is preferred to aom?

codingjoe:
> Since this PR is so far along in review, I figure I should run any non-trivial changes past the reviewers before undertaking them:
> 
> libaom is a very capable codec, but it has absolutely awful performance and compression efficiency with default settings. This gets reported as an issue to pillow-avif-plugin, and [my advice](https://github.com/fdintino/pillow-avif-plugin/issues/65#issuecomment-2513477035) is usually to use rav1e. How would people feel about changing the order of encoder codecs such that rav1e is preferred over aom?

I can second that. rav1e isn't only faster in my experience, but some might consider it more open…

radarhere:
If it is more performant, then I raise no objection to changing the order of preference.

radarhere:
Just for awareness, do you know why libavif doesn't also prefer rav1e over aom?

winbuild/build_prepare.py
```diff
@@ -312,6 +317,42 @@ def cmd_msbuild(
...
+    "libavif": {
+        "url": "https://github.com/AOMediaCodec/libavif/archive/v0.8.4.zip",
+        "filename": "libavif-0.8.4.zip",
+        "dir": "libavif-0.8.4",
+        "build": [
+            cmd_cd("ext"),
+            'cmd.exe /c "aom.cmd"',
+            "if errorlevel 1 echo AOM build failed! && exit /B 1",
+            'cmd.exe /c "dav1d.cmd"',
+            "if errorlevel 1 echo dav1d build failed! && exit /B 1",
```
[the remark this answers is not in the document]
radarhere: The code this is commenting on is no longer present.

fdintino:
libaom is the most configurable and tunable of the encoding codecs, and with the right parameters it is very competitive—if not superior—to the other codecs, particularly for video. But for whatever reason the default encoder options are suboptimal. One of rav1e's strengths is that it has very sensible defaults. And some of the features aom has that rav1e lacks, like film grain synthesis, are less relevant when encoding still images.

vrabaud:
Hi, libavif maintainer here. dav1d is indeed what we recommend for decoding. For encoding, libaom just saw some recent patches to greatly improve quality performance for images, cf https://aomedia.googlesource.com/aom/+/2d2f644e475ce348722ea8b199b49baea285ed04 and all the patches above dealing with SSIMULACRA2 and will become the best recommednation for encoding in libavif. The work that got merged into libaom is by the same great people who did it for SVT-AV1, cf https://github.com/AOMediaCodec/libavif/issues/2412.

Eskuero:
Nobody argued for it so it may be pointless but I will add an extra argument against SVT as the default. They seem focused only on video so you end up hitting some of their [barriers](https://gitlab.com/AOMediaCodec/SVT-AV1/-/blob/master/Source/Lib/Globals/enc_settings.c#L39) when using images with non standard resolutions. To name a few that I have hit myself:

- Width can't be bigger than 16384 pixels
- Height can't be bigger than 8704 pixels.
- Height * width can't be bigger than 21.012.480 pixels so per example a 8000*3000 can't be encoded either
- Neither dimension can be odd 

So you end with checks like this
```
# Decide codec to use
if image.width < 64 or image.height < 64 or image.height > 8704 or image.width > 16384 or (image.width * image.height) > 21012480 or (image.width % 2) == 1 or (image.height) % 2 == 1:
	image_codec = "aom"
else:
	image_codec = "svt"
```
To avoid exceptions like:

> Svt[error]: Error Instance 1: Source Width must be even for YUV_420 colorspaces                                                         
....
RuntimeError: Failed to encode image: Encoding of color planes failed


------------------------------------------------------------------------------

That being said on a personal note for anybody's whose usercase is the same as mine: archiving with lossless quality and max compression, I found svt to always beat the other codecs, with aom coming next and rav1e dead last (probably because [they do not support lossless](https://github.com/xiph/rav1e/issues/151).

```
[eskuero@4cx4]$ pacman -Q | grep -E "rav1e|svt-av1|aom"
aom 3.11.0-1
rav1e 0.7.1-1
svt-av1 2.2.1-1
(venv) [eskuero@4cx4]$ time python convert.py svt
Codec svt output is 19.49% of the original size

real    0m10.453s
user    0m38.610s
sys     0m1.916s
(venv) [eskuero@4cx4]$ time python convert.py aom
Codec aom output is 32.96% of the original size

real    0m14.470s
user    3m11.148s
sys     0m0.261s
(venv) [eskuero@4cx4]$ time python convert.py rav1e
Codec rav1e output is 33.23% of the original size

real    2m42.493s
user    15m44.806s
sys     0m0.458s
(venv) [eskuero@4cx4]$ du -sh *
2.0M    aom.avif
2.1M    rav1e.avif
6.1M    sample.png
1.2M    svt.avif
```

convert.py
```
from PIL import Image
import pillow_avif
import sys
import os

os.environ['SVT_LOG'] = '1'
image_codec = sys.argv[1]

image = Image.open("sample.png")
image.save(f"{image_codec}.avif", "AVIF", quality = 100, speed = 0, codec = image_codec)

file_size_og = os.path.getsize("sample.png")
file_size_new = os.path.getsize(f"{image_codec}.avif")
percent = round(file_size_new / file_size_og * 100, 2)

print(f"Codec {image_codec} output is {percent}% of the original size")
```

jnoring-pw:
Does it make sense to follow suit with whatever the default is in `libavif`, which is the library doing the dispatching to a specific codec implementation? I feel like that project is going to have the best pulse on what is ideal for defaults.

In my own testing, I stuck with libaom for encoding. What I think is probably most important is A) stability in the default encode/decode choices, B) the ability to customize the decoder, and C) documentation.  On point C, what I've found most difficult about AVIF hasn't been the pillow plugin (a thousand thanks to @fdintino for writing this!), but even figuring out the various settings that can be plumbed through, and what they even mean in real-world terms.

doublex:
> what I've found most difficult about AVIF hasn't been the pillow plugin

Yup!

fdintino:
I just ran some benchmarks with libavif linked against the aom main branch, and in all of my tests it comes out ahead of rav1e (without having to pass any codec-specific options). So I've reverted the commit that switched the encoder codec priority order.

winbuild/build_prepare.py
```diff
@@ -386,6 +388,27 @@ def cmd_msbuild(
         ],
         "bins": [r"*.dll"],
     },
+    "libavif": {
+        "url": f"https://github.com/AOMediaCodec/libavif/archive/v{V['LIBAVIF']}.zip",
+        "filename": f"libavif-{V['LIBAVIF']}.zip",
+        "dir": f"libavif-{V['LIBAVIF']}",
```
radarhere, line 394:
```suggestion
```
This is no longer needed, thanks to https://github.com/python-pillow/Pillow/pull/8567

radarhere: I've created https://github.com/fdintino/Pillow/pull/13

vrabaud:
> I just ran some benchmarks with libavif linked against the aom main branch, and in all of my tests it comes out ahead of rav1e (without having to pass any codec-specific options). So I've reverted the commit that switched the encoder codec priority order. Thanks for the input and context, @vrabaud .
> 
> Vincent, do you know when that's likely to end up in a libavif release? It looks like the SSIMULACRA2 commits just missed making it into the last aom release, but I know libavif has at times used a commit SHA for the aom dependency.

libavif will not depend on a sha1 this time, so we'll wait for the next libaom release, in Q1 2025.

.ci/install.sh
radarhere: This line is no longer present.

radarhere:
check_avif_leaks.py is almost identical to check_j2k_leaks.py. Could you talk about why you added it? Or rather, since we don't have such a file for every format, why you think that AVIF is in particular danger of leaking memory?

Tests/helper.py
radarhere: This line is no longer present.

fdintino:
> check_avif_leaks.py is almost identical to check_j2k_leaks.py. Could you talk about why you added it? Or rather, since we don't have such a file for every format, why you think that AVIF is in particular danger of leaking memory?

When I was first writing this plugin it was useful for me, personally, to guard against introducing memory leaks. The code is mature enough now that I think it's served its purpose, and I wouldn't have any objections against its removal.

src/_avif.c
```diff
@@ -0,0 +1,970 @@
...
+    } else if (strcmp(subsampling, "4:4:4") == 0) {
+        image->yuvFormat = AVIF_PIXEL_FORMAT_YUV444;
+    } else {
+        PyErr_Format(PyExc_ValueError, "Invalid subsampling: %s", subsampling);
+        return NULL;
+    }
+
+    image->colorPrimaries = AVIF_COLOR_PRIMARIES_UNSPECIFIED;
+    image->transferCharacteristics = AVIF_TRANSFER_CHARACTERISTICS_UNSPECIFIED;
+    image->matrixCoefficients = AVIF_MATRIX_COEFFICIENTS_BT601;
```
radarhere, line 306: Why isn't this `AVIF_MATRIX_COEFFICIENTS_UNSPECIFIED`, like the two 'UNSPECIFIED' lines before it?

And why does an empty ICC profile lead to these being overridden by `AVIF_COLOR_PRIMARIES_BT709` and `AVIF_TRANSFER_CHARACTERISTICS_SRGB` on lines 410-411? Seems odd.

fdintino: These default values are set to match the ones found in libavif's avifenc.c. Here are the inline comments that precede these settings ([link](https://github.com/tongyuantongyu/libavif/blob/407672cf9a7c4109fc6a2d6a0cd06c748fa2f403/apps/avifenc.c#L799-L805
)):

> By default, the color profile itself is unspecified, so CP/TC are set (to 2) accordingly.
> However, if the end-user doesn't specify any CICP, we will convert to YUV using BT601
> coefficients anyway (as MC:2 falls back to MC:5/6), so we might as well signal it explicitly.
> See: ISO/IEC 23000-22:2019 Amendment 2, or the comment in avifCalcYUVCoefficients()

Here are the comments referenced in `avifCalcYUVCoefficients`

https://github.com/AOMediaCodec/libavif/blob/5eaa316da9b101d5fdfe6939738670ef5610e9d5/src/colr.c#L150-L163

The same is true for the defaults set in the absence of an icc profile. [The avifenc.c comments for that](https://github.com/tongyuantongyu/libavif/blob/407672cf9a7c4109fc6a2d6a0cd06c748fa2f403/apps/avifenc.c#L1240-L1247) have this to say:

> The final image has no ICC profile, the user didn't specify any CICP, and the source
> image didn't provide any CICP. Explicitly signal SRGB CP/TC here, as 2/2/x will be
> interpreted as SRGB anyway.

I think your question amounts to whether or not this is explicitly setting default values just for the sake of being explicit, or if the unspecified values would actually result in different behavior. I don't know, but perhaps a libavif maintainer (@wantehchang or @vrabaud ?) could chime in.

wantehchang: I suggest merging these three lines (304-306) with the if-else statement that handles icc_bytes (lines 390-412) as follows:

```
    if (PyBytes_GET_SIZE(icc_bytes)) {
        self->icc_bytes = icc_bytes;
        Py_INCREF(icc_bytes);

        result = avifImageSetProfileICC(
            image, (uint8_t *)PyBytes_AS_STRING(icc_bytes), PyBytes_GET_SIZE(icc_bytes)
        );
        if (result != AVIF_RESULT_OK) {
            PyErr_Format(
                exc_type_for_avif_result(result),
                "Setting ICC profile failed: %s",
                avifResultToString(result)
            );
            avifImageDestroy(image);
            avifEncoderDestroy(encoder);
            Py_XDECREF(self->icc_bytes);
            PyObject_Del(self);
            return NULL;
        }
        // colorPrimaries and transferCharacteristics are ignored when an ICC profile is present.
        // So set them to UNSPECIFIED.
        image->colorPrimaries = AVIF_COLOR_PRIMARIES_UNSPECIFIED;
        image->transferCharacteristics = AVIF_TRANSFER_CHARACTERISTICS_UNSPECIFIED;
    } else {
        image->colorPrimaries = AVIF_COLOR_PRIMARIES_BT709;
        image->transferCharacteristics = AVIF_TRANSFER_CHARACTERISTICS_SRGB;
    }
    image->matrixCoefficients = AVIF_MATRIX_COEFFICIENTS_BT601;
```

The comment I added answers Frankie's question -- we set colorPrimaries and transferCharacteristics to UNSPECIFIED only when they are ignored. Otherwise we set them to the default values explicitly.

We may want to duplicate the `image->matrixCoefficients = AVIF_MATRIX_COEFFICIENTS_BT601;` line in both the if and else blocks for clarity.

radarhere: Thanks for your response. The suggested change has now been made.

Tosh0kan:
Is the AVIF plugin still needed or has PIL integrated it or made their own solution?

stalkerg:
@Tosh0kan 
> Is the AVIF plugin still needed or has PIL integrated it or made their own solution?

it's still needed, and here people try to finished review of this big PR. After this PR will be merged, the plugin will not needed. 
AVIF is a big challenge for the PIL and many people are involve.

Tosh0kan:
> @Tosh0kan 
> > Is the AVIF plugin still needed or has PIL integrated it or made their own solution?
> 
> it's still needed, and here people try to finished review of this big PR. After this PR will be merged, the plugin will not needed. 
> AVIF is a big challenge for the PIL and many people are involve. 

Thank you for the reply, and I see. Sorry if I sounded impatient, that was not my intention. I was just trying to make sense of the whole PR but couldn't, and just wanted a clear answer.

Best regards

src/_avif.c
```diff
@@ -0,0 +1,944 @@
...
+    memset(&rgb, 0, sizeof(rgb));
+    avifRGBImageSetDefaults(&rgb, image);
+
+    rgb.depth = 8;
+
+    if (decoder->alphaPresent) {
+        rgb.format = AVIF_RGB_FORMAT_RGBA;
+    } else {
+        rgb.format = AVIF_RGB_FORMAT_RGB;
+        rgb.ignoreAlpha = AVIF_TRUE;
```
radarhere, line 813: I wouldn't have thought ignoring the alpha was necessary, because the format is set a line earlier? See https://github.com/AOMediaCodec/libavif/blob/4e9b26e496f56d71b9284bab8cc7d23f34c0512d/include/avif/avif.h#L1002-L1003

src/_avif.c
```diff
@@ -0,0 +1,944 @@
...
+            "Failed to decode frame %u: %s",
+            decoder->imageIndex + 1,
+            avifResultToString(result)
+        );
+        return NULL;
+    }
+
+    image = decoder->image;
+
+    memset(&rgb, 0, sizeof(rgb));
```
wantehchang, line 804: Delete this `memset` call. The `avifRGBImageSetDefaults` call in the next line is the preferred way to initialize `rgb`.

src/_avif.c
wantehchang: Andrew is right. `rgb.ignoreAlpha` is used only when `avifRGBFormatHasAlpha(rgb.format)` is true. Since `rgb.format` is `AVIF_RGB_FORMAT_RGB`, `rgb.ignoreAlpha` is not used and does not need to be set. Just leave it initialized to the default (`AVIF_FALSE`).

src/_avif.c
radarhere: Thanks. That change has now been made.

src/_avif.c
radarhere: Thanks. That change has now been made.

src/_avif.c
```diff
@@ -0,0 +1,932 @@
...
+    if (image->icc.size) {
+        icc = PyBytes_FromStringAndSize((const char *)image->icc.data, image->icc.size);
+    }
+
+    ret = Py_BuildValue(
+        "IIIsSSIS",
+        image->width,
+        image->height,
+        decoder->imageCount,
+        decoder->alphaPresent == AVIF_TRUE ? "RGBA" : "RGB",
```
wantehchang, line 750: Nit: Omit `== AVIF_TRUE`

src/_avif.c
```diff
@@ -0,0 +1,932 @@
...
+        return NULL;
+    }
+
+    image = decoder->image;
+
+    avifRGBImageSetDefaults(&rgb, image);
+
+    rgb.depth = 8;
+    rgb.format =
+        decoder->alphaPresent == AVIF_TRUE ? AVIF_RGB_FORMAT_RGBA : AVIF_RGB_FORMAT_RGB;
```
wantehchang, line 798: Nit: Omit `== AVIF_TRUE`

src/_avif.c
radarhere: I don't mind removing it, but let me ask - if this isn't the appropriate use case, then why does `AVIF_TRUE` exist at all?

wantehchang: Hi Andrew,

I am afraid that there is a misunderstanding. `decoder->alphaPresent` is an expression with a "boolean" type (emulated with the `int` type in C). When testing a boolean expression, it is not necessary to compare it with the constant "true".

We can illustrate this point with an exaggerated example in C++. It is not necessary to write `if ((a > b) == true) {...}`. We can simply write `if (a > b) {...}`.

radarhere: It's been removed.

src/_avif.c
radarhere: It's been removed.

src/PIL/AvifImagePlugin.py
```diff
@@ -0,0 +1,286 @@
...
+    speed = info.get("speed", 6)
+    max_threads = info.get("max_threads", _get_default_max_threads())
+    codec = info.get("codec", "auto")
+    if codec != "auto" and not _avif.encoder_codec_available(codec):
+        msg = "Invalid saving codec"
+        raise ValueError(msg)
+    range_ = info.get("range", "full")
+    tile_rows_log2 = info.get("tile_rows", 0)
+    tile_cols_log2 = info.get("tile_cols", 0)
+    alpha_premultiplied = bool(info.get("alpha_premultiplied", False))
```
radarhere, line 169: Do we need `alpha_premultiplied` as an argument? To my simple way of thinking, it would cause the saved image to no longer be accurate to the Pillow image being saved. We do have a separate mode for premultiplied alpha, `RGBa`, see https://pillow.readthedocs.io/en/stable/handbook/concepts.html#modes

radarhere (approved):
Ubuntu Jammy has libavif 0.9.3, so I guess it makes sense to support versions before 1.0.0.

I would like someone else from the core team to review this for any license implications
- I've realised that this has essentially copied `avifImageGetExifOrientationFromIrotImir` and `avifImageExtractExifOrientationToIrotImir` from libavif. Granted, they are relatively simple functions.
- By introducing new dependencies - not just libavif, but also various codecs - there are different licenses to consider. Remember that we don't distribute libimagequant with our wheels

https://pillow.readthedocs.io/en/stable/installation/building-from-source.html#building-from-source
> Libimagequant is licensed GPLv3, which is more restrictive than the Pillow license, therefore we will not be distributing binaries with libimagequant support enabled.

src/PIL/AvifImagePlugin.py
radarhere: I've created https://github.com/fdintino/Pillow/pull/23

radarhere: I receive a [comment](https://github.com/fdintino/Pillow/pull/23#issuecomment-2685813536) on my PR that
> I think if the underlying pixel data in the AVIF image was RGBa then it would make sense, because you would then presumably get RGBa back from the decoder. But because it is converting to YUV, and because alphaPremultiply is specified separately for the RGB and YUV image, with the former only existing so that libavif can do alpha multiplying and unmultiplying if necessary, I'm not so sure. To the extent that the RGBa mode is used in pillow, I imagine it is mostly for image compositing operations. But for an AVIF image it really only serves to get more efficient compression on images with detailed alpha planes, in a way that doesn't compromise perceptual quality. I think it wouldn't be obvious that the intended way to enable (what amounts to) a lossy compression flag would be to convert to a different mode, particularly when that mode is different from what is actually stored in the image.

It's not the default behaviour of the plugin, so I'm not going to fight too hard against this. I've closed my PR.

.github/workflows/wheels-dependencies.sh
```diff
@@ -50,6 +50,7 @@ LIBWEBP_VERSION=1.5.0
 BZIP2_VERSION=1.0.8
 LIBXCB_VERSION=1.17.0
 BROTLI_VERSION=1.1.0
+LIBAVIF_VERSION=1.1.1
```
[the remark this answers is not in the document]
fdintino: It failed one of our [linux aarch64 builds](https://github.com/fdintino/Pillow/actions/runs/13571382279/job/37937027397). I've opened a bug report to libyuv [here](https://libyuv.issues.chromium.org/issues/399856238)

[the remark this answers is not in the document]
y-guyon: > It failed one of our [linux aarch64 builds](https://github.com/fdintino/Pillow/actions/runs/13571382279/job/37937027397). I've opened a bug report to libyuv [here](https://libyuv.issues.chromium.org/issues/399856238)

https://github.com/AOMediaCodec/libavif/releases/tag/v1.2.1 contains the fix and was released.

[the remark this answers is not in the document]
radarhere: Thanks. I've created https://github.com/fdintino/Pillow/pull/26

src/PIL/AvifImagePlugin.py
```diff
@@ -0,0 +1,284 @@
...
+    )
+
+    # Add each frame
+    frame_idx = 0
+    frame_duration = 0
+    cur_idx = im.tell()
+    is_single_frame = total == 1
+    try:
+        for ims in [im] + append_images:
+            # Get # of frames in this image
```
hugovk, line 233:
```suggestion
            # Get number of frames in this image
```

src/_avif.c
```diff
@@ -0,0 +1,952 @@
...
+
+// Decoder type
+typedef struct {
+    PyObject_HEAD avifDecoder *decoder;
+    Py_buffer buffer;
+} AvifDecoderObject;
+
+static PyTypeObject AvifDecoder_Type;
+
+#if AVIF_VERSION < 1000000
```
hugovk, line 23:
```suggestion
#if AVIF_VERSION < 1000000  // 1.0.0
```

src/_avif.c
```diff
@@ -0,0 +1,952 @@
...
+_encoder_codec_available(PyObject *self, PyObject *args) {
+    char *codec_name;
+    if (!PyArg_ParseTuple(args, "s", &codec_name)) {
+        return NULL;
+    }
+    int is_available = _codec_available(codec_name, AVIF_CODEC_FLAG_CAN_ENCODE);
+    return PyBool_FromLong(is_available);
+}
+
+#if AVIF_VERSION >= 80200
```
hugovk, line 185:
```suggestion
#if AVIF_VERSION >= 80200  // 0.8.2
```

src/_avif.c
```diff
@@ -0,0 +1,952 @@
...
+    if (width == 0 || height == 0) {
+        PyErr_SetString(PyExc_ValueError, "invalid canvas dimensions");
+        error = 1;
+        goto end;
+    }
+    image->width = width;
+    image->height = height;
+
+    image->depth = 8;
+#if AVIF_VERSION >= 90000
```
hugovk, line 321:
```suggestion
#if AVIF_VERSION >= 90000  // 0.9.0
```

src/_avif.c
```diff
@@ -0,0 +1,952 @@
...
+        error = 1;
+        goto end;
+    }
+
+    int is_aom_encode = strcmp(codec, "aom") == 0 ||
+                        (strcmp(codec, "auto") == 0 &&
+                         _codec_available("aom", AVIF_CODEC_FLAG_CAN_ENCODE));
+    encoder->maxThreads = is_aom_encode && max_threads > 64 ? 64 : max_threads;
+
+#if AVIF_VERSION >= 1000000
```
hugovk, line 337:
```suggestion
#if AVIF_VERSION >= 1000000  // 1.0.0
```

src/_avif.c
```diff
@@ -0,0 +1,952 @@
...
+    }
+    if (speed < AVIF_SPEED_SLOWEST) {
+        speed = AVIF_SPEED_SLOWEST;
+    } else if (speed > AVIF_SPEED_FASTEST) {
+        speed = AVIF_SPEED_FASTEST;
+    }
+    encoder->speed = speed;
+    encoder->timescale = (uint64_t)1000;
+
+#if AVIF_VERSION >= 110000
```
hugovk, line 357:
```suggestion
#if AVIF_VERSION >= 110000  // 0.11.0
```

src/_avif.c
```diff
@@ -0,0 +1,952 @@
...
+        encoder->tileRowsLog2 = normalize_tiles_log2(tile_rows_log2);
+        encoder->tileColsLog2 = normalize_tiles_log2(tile_cols_log2);
+    }
+#else
+    encoder->tileRowsLog2 = normalize_tiles_log2(tile_rows_log2);
+    encoder->tileColsLog2 = normalize_tiles_log2(tile_cols_log2);
+#endif
+
+    if (advanced != Py_None) {
+#if AVIF_VERSION >= 80200
```
hugovk, line 369:
```suggestion
#if AVIF_VERSION >= 80200  // 0.8.2
```

src/_avif.c
```diff
@@ -0,0 +1,952 @@
...
+
+        frame->width = width;
+        frame->height = height;
+        frame->colorPrimaries = image->colorPrimaries;
+        frame->transferCharacteristics = image->transferCharacteristics;
+        frame->matrixCoefficients = image->matrixCoefficients;
+        frame->yuvRange = image->yuvRange;
+        frame->yuvFormat = image->yuvFormat;
+        frame->depth = image->depth;
+#if AVIF_VERSION >= 90000
```
hugovk, line 540:
```suggestion
#if AVIF_VERSION >= 90000  // 0.9.0
```

src/_avif.c
```diff
@@ -0,0 +1,952 @@
...
+    if (!decoder) {
+        PyErr_SetString(PyExc_MemoryError, "Can't allocate decoder");
+        PyBuffer_Release(&buffer);
+        PyObject_Del(self);
+        return NULL;
+    }
+#if AVIF_VERSION >= 80400
+    decoder->maxThreads = max_threads;
+#endif
+#if AVIF_VERSION >= 90200
```
hugovk, line 695:
```suggestion
#if AVIF_VERSION >= 80400  // 0.8.4
    decoder->maxThreads = max_threads;
#endif
#if AVIF_VERSION >= 90200  // 0.9.2
```

hugovk (approved):
Thanks for all your work on this! Let's see if we can get this into Pillow 11.2.0...

---

> Ubuntu Jammy has libavif 0.9.3, so I guess it makes sense to support versions before 1.0.0.

We have version checks for 0.8.4 and 0.9.2, are those widespread?

We're starting with a clean slate, so we have the opportunity to pick something modern-ish as a baseline.

* 0.8.4 - 2020-11-23
* 0.9.2 - 2021-06-23
* 0.9.3 - 2021-10-20
* 1.0.0 - 2023-08-24
* 1.2.1 - 2025-03-17 (latest)

https://github.com/AOMediaCodec/libavif/blob/main/CHANGELOG.md

fdintino:
I think it's a good question about what the libavif baseline should be. The latest debian currently has libavif 0.11.1, so I've always tested pillow-avif-plugin against that version (as well as against the latest version). I'm happy with whatever baseline you all decide.

Your question reminded me of something. When I went to integrate some of the improvements @radarhere made on this PR to my own project, pillow-avif-plugin, I realized that there was a backwards incompatible change. Should we want to support libavif versions before 1.0.0, we'll want to fix it. I've created fdintino/Pillow#27 with that fix.

Tests/test_file_avif.py
```diff
@@ -0,0 +1,765 @@
...
+                image, "Tests/images/avif/hopper_avif_write.png", 11.5
+            )
+
+    def test_write_rgb(self, tmp_path: Path) -> None:
+        """
+        Can we write a RGB mode file to avif without error?
+        Does it have the bits we expect?
+        """
+
+        temp_file = str(tmp_path / "temp.avif")
```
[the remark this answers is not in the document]
radarhere: I've pushed a commit for this.

radarhere:
> We're starting with a clean slate, so we have the opportunity to pick something modern-ish as a baseline.

CentOS and Amazon Linux don't have an libavif package
As I said, Ubuntu 22.04 has 0.9.3
Debian 12 has 0.11.1
Ubuntu 24.04 has 1.0.4
Alpine 3.21 has 1.0.4
Fedora 40 has 1.0.4
Arch has 1.2.0

In a vacuum, I like the idea of only starting with the stable version of 1.0.0.

Tests/test_file_avif.py
```diff
@@ -0,0 +1,765 @@
...
+            assert reloaded.info["exif"] == exif_data
+
+    def test_exif_invalid(self, tmp_path: Path) -> None:
+        with Image.open(TEST_AVIF_FILE) as im:
+            test_file = str(tmp_path / "temp.avif")
+            with pytest.raises(SyntaxError):
+                im.save(test_file, exif=b"invalid")
+
+    @pytest.mark.parametrize(
+        "rot, mir, exif_orientation",
```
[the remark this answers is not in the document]
radarhere: https://github.com/AOMediaCodec/libavif/blob/main/doc/avifenc.1.md

> --irot ANGLE : Add irot property (rotation) in 0..3. Makes (90 * ANGLE) degree rotation anti-clockwise.
> 
> --imir AXIS : Add imir property (mirroring). 0=top-to-bottom, 1=left-to-right.

They're either short for rotation and mirroring, or for irot and imir.

hugovk:
Yep, let's do 1.0.0.

fdintino:
@radarhere Could you take a look at fdintino/Pillow#28 and let me know if you are okay with its inclusion here? Thanks.

hugovk:
@fdintino Thank you very much for your work and your persistence! This will be in today's Pillow 11.2.0 release.

Stormheg:
I've been quietly following this pull request since 2022 or so. Super excited to see this land in Pillow 😄 

Congratulations to all involved - what an amazing effort! ❤️

wantehchang:
@radarhere Andrew: I merged the libavif pull request https://github.com/AOMediaCodec/libavif/pull/2723 yesterday. You should be able to remove your `CMAKE_POLICY_VERSION_MINIMUM=3.5` workaround for libyuv's CMakeLists.txt problem with CMake 4.0.0 now.

radarhere:
Thanks. I've tested it, and it's still required for now. I see that's in libavif main, so that's something to look forward to with the next libavif release.

wantehchang:
@radarhere Ah, you are using a release tag of libavif. I will look into creating a libavif v1.2.2 patch release with the CMake 4.0.0 build fixes.

codingjoe:
@radarhere can you remind me? Why didn't we ship Pillow binaries w/o AVIF support?
It does make application deployment hard. Especially, since tools like [pipenv don't support --no-binary](https://github.com/pypa/pipenv/issues/5362).

It raises the bar substantially for most users. Which is sad, considering the effort y'all put into this. Besides, AVIF is part of Baseline 2024 and, with the proliferation of client decoding, is a direly needed replacement for WebP for most websites.

aclark4life:
> why didn't we ship Pillow binaries w/o AVIF support?

We _did_ ship 11.2.1 w/o AVIF support but even without it, [we still call `register_open`](https://github.com/python-pillow/Pillow/issues/8886#issuecomment-2814038881). Looks like we may [try again in 11.3](https://github.com/python-pillow/Pillow/issues/8856#issuecomment-2827278596).

codingjoe:
> > why didn't we ship Pillow binaries w/o AVIF support?
> 
> We _did_ ship 11.2.1 w/o AVIF support but even without it, [we still call `register_open`](https://github.com/python-pillow/Pillow/issues/8886#issuecomment-2814038881). Looks like we may [try again in 11.3](https://github.com/python-pillow/Pillow/issues/8856#issuecomment-2827278596).

Thanks, I appreciate the swift response. This was such a long journey, I didn't keep up with everything. And now I am trying to do the django-pictures integration. The link to the tickets helps a lot, I should have found it myself. Sorry for the trouble. Anyhow, now I know how to update the documentation to help people out.

radarhere:
Pillow 11.3.0 has now been released with AVIF support in the wheels.

codingjoe:
So, we got some Hot Wheels Cars GIF by Mattel

Sorry for the dad joke 🥸

PR #5206, opened 2021-01-14
Allow PixelAccess to use Python __int__ when parsing x and y

### LINKED ISSUE #5048: PixelAccess class does not support Numpy int indexing
Hi developers, I happen to find out that PixelAcess class fails to use Numpy int as index, while Python built-in int, Python built-in float and even Numpy float all work fine. I'm using Python 3.6.1, Numpy 1.15.0, PIL 7.2.0.

An example is provided as follows.

code:
`import numpy as np
from PIL import Image
import os

base = os.path.dirname(os.path.realpath(__file__))
im_path = os.path.join(base, 'xxxx.bmp')
pix = Image.open(im_path).load()

default_float_coor = [1.1, 9.9]
default_int_coor = [1,10]
numpy_float_coor = np.array([1.1, 9.9])
numpy_int_coor = np.array([1, 10])

print(type(pix))
print('pix at default_float_coor: ', pix[default_float_coor[0], default_float_coor[1]])
print('pix at default_int_coor: ', pix[default_int_coor[0], default_int_coor[1]])
print(type(numpy_float_coor[0]))
print('pix at numpy_float_coor: ', pix[numpy_float_coor[0], numpy_float_coor[1]])
print(type(numpy_int_coor[0]))
print('pix at numpy_int_coor: ', pix[numpy_int_coor[0],numpy_int_coor[1]])`

Output:
`<class 'PixelAccess'>
pix at default_float_coor:  (71, 71, 75)
pix at default_int_coor:  (70, 70, 70)
<class 'numpy.float64'>
pix at numpy_float_coor:  (71, 71, 75)
<class 'numpy.int32'>
  File "e:/HanLab_MyLab/Updated Codes/PreProcessing/draft.py", line 22, in <module>
    print('pix at numpy_int_coor: ', pix[numpy_int_coor[0],numpy_int_coor[1]])
TypeError: an integer is required`

np.int64 fails too as I trialed.
It would add to much convenience if Numpy int could serve as index.
Thx.

radarhere:
Resolves #5048

The issue reports that
```python
px = im.load()
px[numpy.int32(5), numpy.int32(5)]
```
returns
```
```

This PR calls `__int__()` on the object passed in, which should provide support for other kinds of objects as well.

### DIFF
--- a/Tests/test_image_access.py
+++ b/Tests/test_image_access.py
@@ -23,6 +23,11 @@ else:
     except ImportError:
         cffi = None
 
+try:
+    import numpy
+except ImportError:
+    numpy = None
+
 
 class AccessTest:
     # initial value
@@ -108,6 +113,13 @@ class TestImagePutPixel(AccessTest):
                 pix2[x, y] = pix1[x, y]
 
         assert_image_equal(im1, im2)
+
+    @pytest.mark.skipif(numpy is None, reason="NumPy not installed")
+    def test_numpy(self):
+        im = hopper()
+        pix = im.load()
+
+        assert pix[numpy.int32(1), numpy.int32(2)] == (18, 20, 59)
 
 
 class TestImageGetPixel(AccessTest):
--- a/Tests/test_image_access.py
+++ b/Tests/test_image_access.py
@@ -71,6 +71,10 @@ class TestImagePutPixel(AccessTest):
         pix1 = im1.load()
         pix2 = im2.load()
 
+        for x, y in ((0, "0"), ("0", 0)):
+            with pytest.raises(TypeError):
+                pix1[x, y]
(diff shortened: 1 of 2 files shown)


Seikegn:
Thank you for making life easier!

wiredfool:
Do we need to do anything on PyAccess for this?

radarhere:
Testing, I find that PyAccess works without this PR. The following code produces no error.
```python
from PIL import Image
import numpy
Image.USE_CFFI_ACCESS = True
pix = im.load()
print(pix.__class__)  # <class 'PIL.PyAccess._PyAccess32_3'>
pix[numpy.int32(1), numpy.int32(2)]
```

PR #4379, opened 2020-01-25
Ensure that errors directory exists before trying to upload from it

radarhere:
Saw this in a failed test in #4378 

```
2020-01-25T02:08:26.0031734Z ##[error]Process completed with exit code 1.
2020-01-25T02:08:26.0213052Z ##[group]Run actions/upload-artifact@v1
2020-01-25T02:08:26.0213188Z with:
2020-01-25T02:08:26.0213274Z   name: errors
2020-01-25T02:08:26.0213352Z   path: Tests/errors
2020-01-25T02:08:26.0213415Z env:
2020-01-25T02:08:26.0213494Z   pythonLocation: d:\a\Pillow\pypy3.6
2020-01-25T02:08:26.0213574Z ##[endgroup]
2020-01-25T02:08:26.2089374Z ##[error]Path does not exist d:\a\Pillow\Pillow\Tests\errors
2020-01-25T02:08:26.2448668Z ##[error]Exit code 1 returned from process: file name 'c:\runners\2.164.0\bin\Runner.PluginHost.exe', arguments 'action "GitHub.Runner.Plugins.Artifact.PublishArtifact, Runner.Plugins"'.
2020-01-25T02:08:26.2467880Z Cleaning up orphan processes
```

This PR fixes the error that the path does not exist, using `mkdir -p`.

PR #5219, opened 2021-01-22
Replaced mkdir command with Python

radarhere:
I recently pushed a [commit](https://github.com/radarhere/Pillow/commit/adf125309e789f5a9568f212d4676010a6852c63) to GitHub Actions. It failed for a perfectly valid reason, but when failing, [test-windows reported](https://github.com/radarhere/Pillow/runs/1751631505?check_suite_focus=true) -

Screen Shot 2021-01-23 at 9 28 54 am

The '-p' flag is meant to [prevent this error](https://github.com/python-pillow/Pillow/pull/4379), but I guess this doesn't work on Windows (https://en.wikipedia.org/wiki/Mkdir#Options)

So instead, this PR replaces 'mkdir' with a Python command. It also changes test.yml, just for symmetry.

See https://github.com/radarhere/Pillow/runs/1751747364?check_suite_focus=true for my failing case combined with this commit to produce a correct 'Prepare to upload errors'

### DIFF
--- a/.github/workflows/test-windows.yml
+++ b/.github/workflows/test-windows.yml
@@ -174,7 +174,7 @@ jobs:
       if: failure()
       run: |
         mkdir -p Tests/errors
-      shell: pwsh
+      shell: bash
 
     - name: Upload errors
       uses: actions/upload-artifact@v2
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -92,7 +92,6 @@ jobs:
       if: failure()
       run: |
         mkdir -p Tests/errors
-      shell: pwsh
 
     - name: Upload errors
       uses: actions/upload-artifact@v2


hugovk:
I wonder if it failed due to using the `pwsh` shell? Does it work when not specifying a shell? Or using another one?

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#using-a-specific-shell

radarhere:
If I [do not specify the shell](https://github.com/radarhere/Pillow/commit/c8ae086aa910892443b1d9abe56e3b1ce60ce994), then it [fails the same](https://github.com/radarhere/Pillow/runs/2055287704).
if I [specify the cmd shell](https://github.com/radarhere/Pillow/commit/5b37c6361a6fc5fdad54b80f2876bfaa2a5f5762), then it [fails with "The syntax of the command is incorrect."](https://github.com/radarhere/Pillow/runs/2055289592)

hugovk:
> If I [specify the bash shell](https://github.com/radarhere/Pillow/commit/77229fe0f421dd5a8e7d9b4b31f7d765684a7b7a), that does [work](https://github.com/radarhere/Pillow/runs/2055467873)... it seems odd to using bash on Windows, but if you think that is simpler, I can update this.

Yes, let's go for this one. Thanks!

nulano:
Checking the help documents for mkdir in cmd and powershell, I have found that `mkdir <path> -Force` works.

hugovk:
Good to know, thanks! This'll do for now :)

Issue #4591, opened 2020-04-26
Copyright clarifications


doko42:
During a review, Debian found some missing/unclear copyright issues in the pillow sources.
See https://bugs.debian.org/952899

For the Debian packages, I already remove Tests/icc and the testsuite gracefully skips these tests when this directory is not found, So I could extend that for other test cases as well. However I'd like to run as many tests as possible.

Could you have a look, what could be addressed upstream?

wiredfool:
For reference, this is the included bug report:

```
Source: pillow
Version: 6.2.1-2
Severity: serious
Justification: Policy 2.3, 4.5, 12.5

During review in binary-NEW, I found the following issues with
d/copyright.  Since the issues already exist in the archive I did not
REJECT, but they should be fixed.

Note that I am filing an identical list of problems against
src:pillow-python2 as the orig.tar is now in the archive twice.

- Needs "Alex Clark **and contributors**" from LICENSE

- Copyright information for Tests/fonts not in d/copyright.

- Google suggests that Tests/images/bmp is GPL-3 and also that the files
  are generated using C code.  If that's right, we would need to at
  least include that code for generating the images in
  debian/missing-sources/ (and ideally regenerate them).

- Tests/images/pillow3.icns looks like it contains a colour space
  copyright 1998 Hewlett-Packard Company; unclear how licensed.

- Tests/test_file_fli.py says that Tests/images/a.fli came from the web
  and I cannot see any free license; probably needs to be filtered out.

- Tests/test_file_mcidas.py says something similar for some other
  images; this should be accounted for.

- Several files matching {docs/example,src/PIL}/*ImagePlugin.py and
  src/libImaging/BcnDecode.c are public domain/CC0.

- Most files matching
  src/PIL/{ImageMorph,MspImagePlugin,SpiderImagePlugin,_binary}.py and
  src/libImaging/{Jpeg2K*,Quant*,SgiRleDecode.c,UnsharpMask.c} have
  different copyright holders.

- src/Tk/_tkmini.h has different copyright holders and license; possibly
  not DFSG-free due to "GOVERNMENT USE" section.

- src/libImaging/{QuantOctree.c,raqm.h,nmake.opt} have different
  licenses and copyright holders.
```

wiredfool:
The only two I can address off of the top of my head are the ICNS/Colorspace issue, which Debian has complained about before in other files, and the BMP test images.  

[...]

PR #5220, opened 2021-01-23
Removed CC0 notices

radarhere:
One of the comments in https://github.com/python-pillow/Pillow/issues/4591#issuecomment-619537811 is
> Several files matching {docs/example,src/PIL}/*ImagePlugin.py and src/libImaging/BcnDecode.c are public domain/CC0.

While the original authors may have released their code into the public domain, we do modify these files over time. Removing these notices simplifies our licensing situation - IANAL, but there shouldn't be anything stopping us from effectively taking this public domain code and stamping it with our more restrictive license instead.


wiredfool:
I'm not really comfortable stripping copyright/licensing info from files. While it's probably strictly legal, I don't think that it would go over well with the contributors.

PR #5231, opened 2021-01-27
Install pytest and pytest-cov using pip rather than pacman

radarhere:
Fixes an error that has started appearing on MSYS jobs. Seen in #5230 (https://github.com/python-pillow/Pillow/pull/5230/checks?check_run_id=1779961003#step:6:32) and #5232, and will presumably be seen in master with the next merge.

```
Running selftest:
--- 58 tests passed.
  File "C:/msys64/mingw64/lib/python3.8/runpy.py", line 185, in _run_module_as_main
    mod_name, mod_spec, code = _get_module_details(mod_name, _Error)
  File "C:/msys64/mingw64/lib/python3.8/runpy.py", line 144, in _get_module_details
    return _get_module_details(pkg_main_name, error)
  File "C:/msys64/mingw64/lib/python3.8/runpy.py", line 111, in _get_module_details
    __import__(pkg_name)
  File "C:/msys64/mingw64/lib/python3.8/site-packages/pytest/__init__.py", line 5, in <module>
    from _pytest.assertion import register_assert_rewrite
  File "C:/msys64/mingw64/lib/python3.8/site-packages/_pytest/assertion/__init__.py", line 9, in <module>
    from _pytest.assertion import rewrite
  File "C:/msys64/mingw64/lib/python3.8/site-packages/_pytest/assertion/rewrite.py", line 33, in <module>
    from _pytest._version import version
ModuleNotFoundError: No module named '_pytest._version'
```

### DIFF
--- a/.github/workflows/test-windows.yml
+++ b/.github/workflows/test-windows.yml
@@ -249,8 +249,6 @@ jobs:
               ${{ matrix.package }}-python3-olefile \
               ${{ matrix.package }}-python3-pip \
               ${{ matrix.package }}-python3-pyqt5 \
-              ${{ matrix.package }}-python3-pytest \
-              ${{ matrix.package }}-python3-pytest-cov \
               ${{ matrix.package }}-python3-setuptools \
               ${{ matrix.package }}-freetype \
               ${{ matrix.package }}-ghostscript \
@@ -263,7 +261,7 @@ jobs:
               ${{ matrix.package }}-openjpeg2 \
               subversion
 
-          python3 -m pip install pyroma
+          python3 -m pip install pyroma pytest pytest-cov
 
           pushd depends && ./install_extra_test_images.sh && popd
 


nulano:
It seems the issue isn't `pytest-cov`, but `pytest`. Not sure why, but this change also makes `pip` upgrade pytest to 6.2.2, which is actually what is fixing this issue. See also: https://github.com/msys2/MINGW-packages/pull/7690#issuecomment-767000081

radarhere:
Keeping [pytest and pytest-cov pacman with pytest>=6.2.2](https://github.com/radarhere/Pillow/commit/66294c2f7628de149b04c9165083cf6a3d22d222) still [failed somehow](https://github.com/radarhere/Pillow/runs/1782315979).

Switching [both pytest and pytest-cov from pacman to pip](https://github.com/radarhere/Pillow/commit/c88e3d67245295121d0da44bbcd07fdf479f89af) did [pass](https://github.com/radarhere/Pillow/runs/1782519307), so that is an alternative to this PR - not a problem, but I also don't see the advantage.

nulano:
Haven't tested it, but I suspect `python3 -m pip install pyroma pytest>=6.2.2` is interpreted as `python3 -m pip install pyroma pytest` with output redirected to the file `=6.2.2`, which doesn't update pytest because it is already installed (although pip thinks it's version 0.0.0).

> Switching both pytest and pytest-cov from pacman to pip did pass, so that is an alternative to this PR - not a problem, but I also don't see the advantage.

A (very slightly) shorter install time would be one advantage, since pytest wouldn't have to be installed twice.

radarhere:
Ah - [`python3 -m pip install pyroma "pytest>=6.2.2"`](https://github.com/radarhere/Pillow/commit/6f1568dd7ca8e68a4011a366e05a33548f280e44) does [work.](https://github.com/radarhere/Pillow/actions/runs/517932295)

Ok, I've changed this PR to swapping both pacman lines for pip.

hugovk:
In general, is there an advantage to installing via pacman instead of pip?

nulano:
For binary packages, the advantage is much shorter install times (no need to compile), and sometimes there might be a patch applied to the pacman version needed to get a package to work. I wouldn't expect much of a difference for pure-python packages such as pytest (pytest and pytest-cov don't have any patches, just an install script for the `pytest.exe` binaries). So for pytest, it might be better to use pip simply because it is updated more frequently.

My understanding is that MSYS2 provides pacman packages only for those libraries that require patches, and their dependencies. Pytest is likely only available because it is a dependency of something else.

PR #5238, opened 2021-01-30
Updated libwebp to 1.2.0

radarhere:


nulano:
For reference, this is the changelog: https://fossies.org/diffs/libwebp/1.1.0_vs_1.2.0/ChangeLog-diff.html

Issue #5235, opened 2021-01-29
Drawing polygon gets different result with reversed xy


hrdrq:
### What did you do?
I found that sometimes using reversed `xy` will get different images
(Just few pixels though)

### What did you expect to happen?
Using reversed `xy` will get the same image

### What actually happened?
Got different images.
In the example below, the pixels `(33, 127)` and `(34, 127)` are different.

In `im1`, `(33, 127)` and `(34, 127)` has no color.
In `im2`, `(33, 127)` and `(34, 127)` has color.

The coordinates near them are:
```
 (33, 136),
 (35, 131),
 (35, 130),
 (35, 129),
 (35, 128),
 (35, 127),
 (33, 127),
 (32, 127),
 (30, 127),
 (30, 126),
```
I think `(33, 127)` and `(34, 127)` should have color,
so `im2` is correct. (which uses the reversed `xy`)

### What are your OS, Python and Pillow versions?

* OS: macOS Catalina 10.15.7
* Python: 3.9.1
* Pillow: 8.0.1

```python
from PIL import Image, ImageDraw

[...]

PR #5240, opened 2021-01-31
Fix polygon drawing if there are continuous horizontal edges

hrdrq:
Fixes https://github.com/python-pillow/Pillow/issues/5235

- In `polygon_generic()`, collect horizontal edges and sort them by `xmin`.
- `draw_horizontal_lines` only needs to loop horizontal edges.

### Simple problem reproduction
```py
xy = [(2, 6), (12, 6), (12, 12), (8, 12), (8, 8), (4, 8), (2, 8)]
image = Image.new('RGB', (16, 16))
draw = ImageDraw.Draw(image)
draw.polygon(xy, 'white')
image.show()
```
problem


src/libImaging/Draw.c
```diff
@@ -434,7 +446,7 @@ draw_horizontal_lines(
                 }
             }
 
-            (*hline)(im, xmin, e[i].ymin, xmax, ink);
+            (*hline)(im, xmin, e[i]->ymin, xmax, ink);
```
radarhere, line 449: Could you explain your thinking behind this change?

hrdrq: @radarhere 
Hi, I'm thinking using `Edge **edge_table` to store "horizontal edges" and sort them.
(Currently `edge_table` only store "non-horizontal edges", and there is space for horizontal ones)

The reason why I want to sort them is that, if the order of continuous horizontal edges is (x-value) descending, 
the edge whose x is bigger will be skipped and never be drawn.

Since it's pointer of pointer, it's needed to use `->` instead of `.`

https://github.com/python-pillow/Pillow/pull/5240/files#diff-09507fd39d242b056864e569bedf77dda352959bad5f0fa7acd077c8d9371dcaR460

hrdrq: An example of the case:
https://github.com/python-pillow/Pillow/issues/5235#issuecomment-770226489

radarhere:
I've created #5347 as an alternative to this, that I hope can be considered to be a simpler change.

hrdrq:
Got it. Thank you

hugovk:
Thanks for the issue and PR, we went with https://github.com/python-pillow/Pillow/pull/5347. It'll be out in this week's release.

PR #5242, opened 2021-02-02
 Add support for decoding plain PPM formats

### LINKED ISSUE #860: 'IOError: cannot identify image file' when opening .pbm, .pgm and .ppm in an ascii file encoding
Netpbm file formats can be encoded in either binary or ASCII. When opening a .pbm, .pgm or .ppm in the ASCII format, the following error is thrown:

```
IOError: cannot identify image file
```

All Netpbm formats in binary encoding open perfectly.

### LINKED ISSUE #4999: P3 ppm file format support (the non binary version of ppm that is very common)

Piolie:
Resolves #860
Resolves #4999
Builds upon @gofr's PR draft #5009.

 * only adds decode support, not encode support;
 * includes several tests;
 * does error handling;
 * is also affected by #5008. It would be easy to fix this. However, it would make the results of the `raw` and `plain` formats inconsistent, so I decided to leave it as is;
 * in plain PBM the inter-token whitespace is optional, so the decoder differs from the one for the other formats;
 * there is also a hack to handle the fact that the 32-bit PGM extension expects signed ints, unlike the 8- and 16-bit ones, which are unsigned.

I made the changes on top of my other, still open, PR (#5121).

gofr's PR could potentially read the whole file at once (if there are no `LF`s) which for very large files could take up lots of memory. To avoid this, I tried to read one char at a time and although the implementation was easy, the performance was poor. I finally opted to read 1 MB blocks and process each one instead. The code ended up being more complex,* so there's that. On the bright side, it could easily be adapted not to `_pull_fd`, if it's any use.

*Strictly speaking, the PPM format doesn't allow comments in the raster section. However, these are commonly used. Ignoring them even when they span several blocks or split tokens in half was the hardest part.


src/PIL/PpmImagePlugin.py
```diff
@@ -49,71 +53,264 @@ class PpmImageFile(ImageFile.ImageFile):
...
+#
+# --------------------------------------------------------------------
+
+
+class PpmPlainDecoder(ImageFile.PyDecoder):
+    _pulls_fd = True
+
+    def _read_block(self, block_size=10 ** 6):
+        return bytearray(self.fd.read(block_size))
+        # return self.fd.read(block_size)
```
radarhere, line 162:
```suggestion
```
Is this unintentionally left in? Or were you trying to communicate something to future developers?

src/PIL/PpmImagePlugin.py
```diff
@@ -49,71 +53,264 @@ class PpmImageFile(ImageFile.ImageFile):
...
-        if s != b"P":
-            raise SyntaxError("not a PPM file")
-        magic_number = self._token(s)
-        mode = MODES[magic_number]
+    def _open(self):
+        magic_number = self._read_magic()
+        try:
+            decoder, mode = MODES[magic_number]
+        except KeyError:
+            raise SyntaxError("Not a PPM image file") from None
```
radarhere, line 101: Why raise it from `None`?

src/PIL/PpmImagePlugin.py
Piolie: I was unintentional, sorry.

src/PIL/PpmImagePlugin.py
Piolie: I was considering the underlying cause (`KeyError`) not relevant to the end user.

src/PIL/PpmImagePlugin.py
```diff
@@ -135,7 +331,7 @@ def _save(im, fp, filename):
     elif im.mode == "RGBA":
         rawmode, head = "RGB", b"P6"
     else:
-        raise OSError(f"cannot write mode {im.mode} as PPM")
+        raise OSError(f"Cannot write mode {im.mode} as PPM")
```
radarhere, line 334:
```suggestion
        raise OSError(f"cannot write mode {im.mode} as PPM")
```
Considering that "cannot write mode" is seen in 9 other plugins, it would be nice not to have this one be unique

Piolie: Sure thing.

src/PIL/PpmImagePlugin.py
radarhere: The change this comment was addressing was part of #5121

src/PIL/PpmImagePlugin.py
```diff
@@ -123,7 +128,176 @@ def _open(self):
...
+        elif self.mode == "I":
+            if rawmode == "I;16B":
+                decoded_data = self._decode_blocks(channels=1, depth=16)
+            elif rawmode == "I;32B":
+                decoded_data = self._decode_blocks(channels=1, depth=32)
+        elif self.mode == "RGB":
+            decoded_data = self._decode_blocks(channels=3, depth=8)
+
+        self.set_as_raw(bytes(decoded_data), rawmode)
+        return -1, 0
```
radarhere, line 300:
```suggestion
        return 0, 0
```
I think this should be zero, as per https://github.com/python-pillow/Pillow/blob/180c89f7559c7faeddd1a215e2a2937dc3afb4d9/src/PIL/ImageFile.py#L668-L674

src/PIL/PpmImagePlugin.py
```diff
@@ -123,7 +128,176 @@ def _open(self):
...
+                decoded_data.append(token)
+                if tokens_read == total_tokens:  # finished!
+                    invert = bytes.maketrans(b"01", b"\xFF\x00")
+                    decoded_data = decoded_data.translate(invert)
+                    return decoded_data
+
+    def _decode_blocks(self, channels=1, depth=8):
+        decoded_data = bytearray()
+        if depth == 32:
+            maxval = 2**31 - 1  # HACK: 32-bit grayscale uses signed int
```
radarhere, line 221: I can see that your test image agrees with you. I'd like to understand why this is though. Where did you test image come from?

radarhere: If I use ImageMagick, `convert Tests/images/hopper.png -depth 32 out.pgm`,`maxval` is `2 ** 32 - 1`.

I'm also able to get the same result without explicitly setting the depth -
```
python3 -c "from PIL import Image;im = Image.open('Tests/images/hopper.png');im.convert('F').save('out.tiff')"
convert out.tiff out.pgm
cat out.pgm | head -n 3
```

radarhere: I've realised that this workaround can be avoided by using the original `maxval`. I've pushed a commit for this.

Piolie: Sorry for the late reply. It's been a while since I worked on this.
Just for the record: I don't remember how I got to the conclusion that that "hack" was needed. I think the test image I manually generated myself—so of course my hack would work ;P. I'm glad that using `maxval` makes all this avoidable.

djbuijs:
I'm looking for this feature as well. What's holding up the merge? I'd be happy to help bring this over the finish-line if there's anything left to do.

radarhere:
After #6242, it has become apparent that the hopper_32bit had a maximum value [outside of the 0-65535 range](https://en.wikipedia.org/wiki/Netpbm#Description). So I've removed those test images.

radarhere:
@djbuijs I've now resolved the merge conflict, and added support for arbitrary maxval in plain formats to continue the work of #6119. Those were the issues that I was conscious of. What's left is just reviewing it to make sure that everything is correct and there's no straightforward improvements to be made.

djbuijs:
Thank you @radarhere!

PR #5250, opened 2021-02-05
Changed Image.open formats parameter to be case-insensitive

Piolie:
* Make `formats` parameter in `Image.open` accept aNy cAsE.

Before this commit
```python
Image.open("Tests/images/hopper.png", formats=['png'])
```
would fail, because the format is not upper case (`'PNG'`). This PR makes `'PNG'`, `'Png'`, `'pNG'` and all other variants equivalent in the `formats` list (or tuple).

### DIFF
--- a/Tests/test_image.py
+++ b/Tests/test_image.py
@@ -94,7 +94,7 @@ class TestImage:
         with pytest.raises(TypeError):
             Image.open(PNGFILE, formats=123)
 
-        for formats in [["jPeG"], ("JpEg",)]:
+        for formats in [["JPEG"], ("JPEG",), ["jpeg"], ["Jpeg"], ["jPeG"], ["JpEg"]]:
             with pytest.raises(UnidentifiedImageError):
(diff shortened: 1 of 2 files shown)


Tests/test_image.py
```diff
@@ -94,7 +94,7 @@ def test_open_formats(self):
         with pytest.raises(TypeError):
             Image.open(PNGFILE, formats=123)
 
-        for formats in [["JPEG"], ("JPEG",)]:
+        for formats in [["jPeG"], ("JpEg",)]:
```
radarhere, line 97:
```suggestion
        for formats in [["JPEG"], ("JPEG",), ["jpeg"], ["Jpeg"], ["jPeG"], ["JpEg"]]:
```
Just a suggestion.
My preference would be to keep the original test cases if possible.
I'm also adding in 'jpeg' and 'Jpeg', since I would think those are the most likely non-uppercase versions.

Issue #5217, opened 2021-01-21
quantize() has worse results on an identical image in RGBA mode instead of RGB


boringhexi:
### What did you do?
RGB image (no alpha) named `test-input.png`
![test-input](https://user-images.githubusercontent.com/36807140/105379145-8c504100-5bda-11eb-8e7c-94501bab67e6.png)

Ran the following code on it.
```python3
from PIL import Image

image = Image.open("test-input.png")

# The input image is converted from RGB to RGBA and back to RGB, holding onto a copy each step of the way.
image1_rgb = image.convert("RGB")
image2_rgba = image1_rgb.convert("RGBA")
image3_back2rgb = image2_rgba.convert("RGB")

# Then each held-onto copy is separately quantized and saved to an output image.
image1_rgb.quantize(colors=16).save("out01_rgb.png")
image2_rgba.quantize(colors=16).save("out02_rgba.png")
image3_back2rgb.quantize(colors=16).save("out03_back2rgb.png")
```

### What did you expect to happen?
Same result across all 3 outputs, since an identical image is used in each case. Changing the mode from RGB to RGBA should not make a difference in the quantization results, since the colors are unchanged by the mode change.

### What actually happened?
The 1st and 3rd outputs, both created from an Image in `RGB` mode, have decent quantization results.
![out01_rgb](https://user-images.githubusercontent.com/36807140/105379847-434cbc80-5bdb-11eb-8f77-b3b2b7a6addd.png) ![out03_back2rgb](https://user-images.githubusercontent.com/36807140/105379860-45168000-5bdb-11eb-9b73-a724276bd208.png)

However, the 2nd output, created from the same exact image in `RGBA` mode, has far worse results, with noticeable gray blotches and loss of detail.
![out02_rgba](https://user-images.githubusercontent.com/36807140/105379924-55c6f600-5bdb-11eb-80f5-5cd1f065c0ea.png)

All 3 outputs have 16 colors as expected, there is not a shortage of colors like in #5204

### What are your OS, Python and Pillow versions?
* OS: Ubuntu 20.04 (Kubuntu)
* Python: 3.9.1 (via pyenv)
* Pillow: 8.1.0

gofr:
This difference is expected. RGB and RGBA have different quantization method defaults. RGB images use `MEDIANCUT`, RGBA images use `FASTOCTREE`.

The defaults aren't documented though. I think they should be.

radarhere:
The documentation has now been updated in #5352

[...]

PR #5251, opened 2021-02-06
Document Image.quantize "method" defaults

gofr:
The method uses different defaults for different image modes.

Relates to issue #5217. Maybe enough to fix it, since I think it's just expected behavior that's not documented well enough.

And um, confusing use of the word "method" in the commit message. I mean the `method` parameter to the `Image.quantize` function.

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -1053,12 +1053,10 @@ class Image:
         of colors.
 
         :param colors: The desired number of colors, <= 256
-        :param method: :data:`MEDIANCUT` (median cut),
-                       :data:`MAXCOVERAGE` (maximum coverage),
-                       :data:`FASTOCTREE` (fast octree),
-                       :data:`LIBIMAGEQUANT` (libimagequant; check support using
-                       :py:func:`PIL.features.check_feature`
-                       with ``feature="libimagequant"``).
+        :param method: :data:`MEDIANCUT` (default),
+                       :data:`MAXCOVERAGE`,
+                       :data:`FASTOCTREE` (default for RGBA images),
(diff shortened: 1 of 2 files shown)


src/PIL/Image.py
```diff
@@ -1053,12 +1053,10 @@ def quantize(self, colors=256, method=None, kmeans=0, palette=None, dither=1):
...
-        :param method: :data:`MEDIANCUT` (median cut),
-                       :data:`MAXCOVERAGE` (maximum coverage),
-                       :data:`FASTOCTREE` (fast octree),
-                       :data:`LIBIMAGEQUANT` (libimagequant; check support using
-                       :py:func:`PIL.features.check_feature`
-                       with ``feature="libimagequant"``).
+        :param method: :data:`MEDIANCUT` (default),
+                       :data:`MAXCOVERAGE`,
+                       :data:`FASTOCTREE` (default for RGBA images),
+                       :data:`LIBIMAGEQUANT`.
```
nulano, line 1059: Why remove the mention of `PIL.features.check_feature(feature="libimagequant")`?

gofr: Because in the online documentation these methods link to https://pillow.readthedocs.io/en/stable/reference/Image.html#quantization-methods which already mentions exactly what it said here. I thought it would be better not to duplicate information.

I've added another commit to remove even more here and make this link to the existing section instead.

gofr: @nulano can you have another look? Are you ok with my changes here?

radarhere:
I've created #5352 as an alternative. It has the same information as yours, just arranged differently.

PR #5259, opened 2021-02-10
Fixed asserting that no warnings were raised

radarhere:
The following test is intended to check that no warnings are raised when opening and closing an image.
https://github.com/python-pillow/Pillow/blob/cc4ed21f191760fbf36417d79eca3ed3555245ea/Tests/test_file_fli.py#L40-L46

If I [change it like so](https://github.com/radarhere/Pillow/commit/418069e0ecce94c29c0e5cd5d2195d45e2b11d68)
```python
def test_closed_file():
    def open():
        Image.open(static_test_file)

    pytest.warns(ResourceWarning, open)
```
It [passes (except in PyPy).](https://github.com/radarhere/Pillow/actions/runs/554673724) So we know that a ResourceWarning is raised by my modification.

But if I [change it back to assert that there are no warnings,](https://github.com/radarhere/Pillow/commit/e7ebc16499d32e4bf0774b8cca968a68beac9013)
```python
def test_closed_file():
    def open():
        Image.open(static_test_file)

    pytest.warns(None, open)
```
It [still passes.](https://github.com/radarhere/Pillow/actions/runs/554674612)

This is because we are not using pytest correctly.

https://docs.pytest.org/en/stable/warnings.html
> To record with func:pytest.warns without asserting anything about the warnings, pass None as the expected warning type:
>
> ```python
> with pytest.warns(None) as record:
>    warnings.warn("user", UserWarning)
>    warnings.warn("runtime", RuntimeWarning)
> 
> assert len(record) == 2
> ```

This PR corrects our usage in various tests.

### DIFF
--- a/Tests/test_bmp_reference.py
+++ b/Tests/test_bmp_reference.py
@@ -20,7 +20,7 @@ def test_bad():
     either"""
     for f in get_files("b"):
 
-        def open(f):
+        with pytest.warns(None) as record:
             try:
                 with Image.open(f) as im:
                     im.load()
@@ -28,7 +28,7 @@ def test_bad():
                 pass
 
         # Assert that there is no unclosed file warning
-        pytest.warns(None, open, f)
+        assert len(record) == 0
 
 
 def test_questionable():
--- a/Tests/test_bmp_reference.py
+++ b/Tests/test_bmp_reference.py
@@ -28,7 +28,7 @@ def test_bad():
                 pass
 
         # Assert that there is no unclosed file warning
-        assert len(record) == 0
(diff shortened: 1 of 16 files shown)


nulano:
From https://docs.pytest.org/en/stable/warnings.html#custom-failure-messages:

> If no warnings are issued when calling `f`, then `not record` will evaluate to `True`.

Would it be clearer to use `assert not record` instead?

radarhere:
Ok, I've pushed another commit for that change

hugovk:
Good catch, thanks!

PR #5261, opened 2021-02-11
Added context managers

radarhere:
This is because of https://pillow.readthedocs.io/en/stable/releasenotes/7.0.0.html#image-del
> Implicitly closing the image’s underlying file in Image.__del__ has been removed. Use a context manager or call close() instead to close the file in a deterministic way.


Tests/test_file_dds.py
```diff
@@ -170,7 +171,8 @@ def test_short_header():
         img_file = f.read()
 
     def short_header():
-        Image.open(BytesIO(img_file[:119]))
+        with Image.open(BytesIO(img_file[:119])):
+            pass
```
hugovk, line 175: let's stick a directive to skip coverage on this line, it's intentionally unreachable (from an exception in `Image.open`)

hugovk (approved):

PR #5268, opened 2021-02-16
Moved RGB fix inside ImageQt class

### LINKED ISSUE #5266: ImageQt does not work as expected in PyQt5
### What did you do?

Attempted to use ImageQt to load a Pillow image in a PyQt5 application

I expected the image to load correctly. There are no errors, but it does not load the image in PyQt5 correctly.

### What actually happened?

The image loaded as a mostly white image with kind of ghost image of the actual photo (see screenshot). I have attached the PyQt5 and PySide6 code, including screenshots from when I ran both files.

**Note: The same code works in PySide6, but not in PyQt5.**

* OS: MacOS Mojave and Windows 10
* Python: Python 3.9
* Pillow: 8.0.0
pink_flower

```python
import sys

from PIL import Image, ImageQt
from PyQt5.QtGui import QPixmap, QImage
from PyQt5.QtWidgets import QWidget, QLabel
from PyQt5.QtWidgets import QVBoxLayout, QApplication


class ImageViewer(QWidget):

    def __init__(self):
        QWidget.__init__(self)
        self.setWindowTitle("PyQt Image Viewer")

        # Open up image in Pillow
        image = Image.open("pink_flower.jpg")
        qt_image = ImageQt.ImageQt(image)
        pixmap = QPixmap.fromImage(qt_image)

        self.image_label = QLabel('')
        self.image_label.setPixmap(pixmap)

        self.main_layout = QVBoxLayout()
        self.main_layout.addWidget(self.image_label)
        self.setLayout(self.main_layout)


if __name__ == "__main__":
    app = QApplication(sys.argv)
    viewer = ImageViewer()
    viewer.show()
    app.exec_()
```

[pyqt_pillow_issue.zip](https://github.com/python-pillow/Pillow/files/5976581/pyqt_pillow_issue.zip)

radarhere:
Resolves #5266. A workaround suggested in that issue is to change the user code. This PR causes the original code mentioned there to work.

https://github.com/python-pillow/Pillow/blob/fdd8b68b83fd1afdbbe5d0a0460ac1413299dcc5/src/PIL/ImageQt.py#L203-L205

This 'strange bug' is because at https://doc.qt.io/qt-5/qimage.html
> QImage::Format_RGB32... is stored using a 32-bit RGB format (0xffRRGGBB).

QImage is expecting the 4th channel to be populated with 0xff (255). Calling `convert("RGBA")` in Pillow does this.

https://github.com/python-pillow/Pillow/blob/fdd8b68b83fd1afdbbe5d0a0460ac1413299dcc5/src/libImaging/Convert.c#L421-L431

Without 255 values, as seen in the issue, the display in Qt is very washed out.

This PR just documents the fix slightly better, and moves it from `toqpixmap()` to `_toqclass_helper()`, so that it is also run when calling the `ImageQt.ImageQt` class (`toqixmap()` still works, since `toqpixmap()` calls `toqimage()`, which returns an instance of the `ImageQt.ImageQt` class).

### DIFF
--- a/Tests/test_qt_image_qapplication.py
+++ b/Tests/test_qt_image_qapplication.py
@@ -2,18 +2,26 @@ import pytest
 
 from PIL import ImageQt
 
-from .helper import assert_image_equal, hopper
+from .helper import assert_image_equal, assert_image_equal_tofile, hopper
 
 if ImageQt.qt_is_installed:
     from PIL.ImageQt import QPixmap
 
     if ImageQt.qt_version == "6":
+        from PyQt6.QtCore import QPoint
+        from PyQt6.QtGui import QImage, QPainter, QRegion
         from PyQt6.QtWidgets import QApplication, QHBoxLayout, QLabel, QWidget
     elif ImageQt.qt_version == "side6":
+        from PySide6.QtCore import QPoint
+        from PySide6.QtGui import QImage, QPainter, QRegion
         from PySide6.QtWidgets import QApplication, QHBoxLayout, QLabel, QWidget
     elif ImageQt.qt_version == "5":
+        from PyQt5.QtCore import QPoint
+        from PyQt5.QtGui import QImage, QPainter, QRegion
         from PyQt5.QtWidgets import QApplication, QHBoxLayout, QLabel, QWidget
     elif ImageQt.qt_version == "side2":
+        from PySide2.QtCore import QPoint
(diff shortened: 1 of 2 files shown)


nulano:
The pypy3.7 run on macos has crashed with a segmentation fault in pytest, IIRC last time that happened the solution was to clear pip cache.

PR #5270, opened 2021-02-17
Added context managers

radarhere:
Sequel to #5261


nulano:
There are also the `helper.assert_image_equal_tofile` and `assert_image_similar_tofile` functions that look ideal for this.

radarhere:
Ok, I've created #5275

PR #5280, opened 2021-02-24
Fix suspicious sequence of types castings

cgohlke:
The suspicious type casting is: `memsize -> 32-bit integer -> memsize`

### DIFF
--- a/src/libImaging/Jpeg2KEncode.c
+++ b/src/libImaging/Jpeg2KEncode.c
@@ -385,7 +385,7 @@ j2k_encode_entry(Imaging im, ImagingCodecState state) {
         float *pq;
 
         if (len > 0) {
-            if ((unsigned)len >
+            if ((ssize_t)len >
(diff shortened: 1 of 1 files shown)


radarhere:
Hi. For my sake at least, could you provide a more detailed explanation?

cgohlke:
https://github.com/python-pillow/Pillow/blob/585683ce258cb5e3362bf22e165d3a739efc27c6/src/libImaging/Jpeg2KEncode.c#L383-L391

```C
    Py_ssize_t len
    ...
    if ((unsigned)len > sizeof() / sizeof()) {
```
On common 64-bit platforms, `Py_ssize_t` is signed 64-bit, `sizeof()` returns unsigned 64-bit `size_t`, and `unsigned` is 32-bit. The 64 bit `len` is explicitly cast to 32 bit and then back to 64 bit for the comparison to 64 bit. To fix this, `len` should be cast to unsigned 64 bit `size_t` for the comparison.

PR #5289, opened 2021-02-26
Added IPythonViewer

### LINKED ISSUE #5286: Displaying images in Google Colab
### What did you do?
`Image.show()`

Display the Image

### What actually happened?
<PIL.PngImagePlugin.PngImageFile image mode=RGB size=450x253 at 0x7FFB7B401C50>

* OS: Windows
* Python: 3.9
* Pillow: 7.0.0

```python
display(image)
```

radarhere:
Resolves #5286 by adding a new `ImageShow.Viewer` - `IPythonViewer`. It calls the `display()` method to allow `ImageShow` to work with IPython... and Google Colab.

Here is my demonstration of it working in Google Colab.
Screen Shot 2021-02-26 at 11 09 38 pm
```python
from PIL import Image, ImageShow
im.show()
```
```python
from PIL import ImageShow
class IPythonViewer(ImageShow.Viewer):
    def show_image(self, image, **options):
        display(image)
        return 1

try:
    from IPython.display import display
except ImportError:
    pass
else:
    ImageShow.register(IPythonViewer)

from PIL import Image
im.show()

# Cleanup
ImageShow._viewers = [v for v in ImageShow._viewers if v.__class__ != IPythonViewer]
```

### DIFF
--- a/Tests/test_imageshow.py
+++ b/Tests/test_imageshow.py
@@ -62,4 +62,20 @@ def test_viewer():
 
 def test_viewers():
     for viewer in ImageShow._viewers:
-        viewer.get_command("test.jpg")
+        try:
+            viewer.get_command("test.jpg")
+        except NotImplementedError:
+            pass
+
+
+def test_ipythonviewer():
+    pytest.importorskip("IPython", reason="IPython not installed")
+    for viewer in ImageShow._viewers:
+        if isinstance(viewer, ImageShow.IPythonViewer):
+            test_viewer = viewer
+            break
+    else:
+        assert False
+
+    im = hopper()
+    assert test_viewer.show(im) == 1
--- a/docs/releasenotes/8.2.0.rst
+++ b/docs/releasenotes/8.2.0.rst
@@ -21,10 +21,17 @@ TODO
 API Additions
 =============
 
-TODO
-^^^^
+ImageShow.IPythonViewer
+^^^^^^^^^^^^^^^^^^^^^^^
 
-TODO
+If IPython is present, this new ``ImageShow.Viewer`` subclass will be
+registered. It displays images on all IPython frontends. This will be helpful
+to users of Google Colab, allowing ``im.show()`` to display images.
+
+It is lower in priority than the other default Viewer instances, so it will
+only be used by ``im.show()`` or ``ImageShow.show()`` if none of the other
+viewers are available. This means that the behaviour of ``ImageShow`` will stay
+the same for most Pillow users.
 
 Security
 ========
--- a/src/PIL/ImageShow.py
+++ b/src/PIL/ImageShow.py
@@ -228,12 +228,12 @@ if sys.platform not in ("win32", "darwin"):  # unixoids
 
 class IPythonViewer(Viewer):
     def show_image(self, image, **options):
-        display(image)
+        ipython_display(image)
         return 1
 
 
 try:
-    from IPython.display import display
+    from IPython.display import display as ipython_display
 except ImportError:
     pass
 else:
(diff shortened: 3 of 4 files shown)


src/PIL/ImageShow.py
```diff
@@ -218,6 +218,20 @@ def get_command_ex(self, file, title=None, **options):
...
 
 
+class IPythonViewer(Viewer):
+    def show_image(self, image, **options):
+        display(image)
+        return 1
+
+
+try:
+    from IPython.display import display
```
nulano, line 228:
```suggestion
    from IPython.display import display as _display
```

Would it be reasonable to add a prefix to avoid confusion when using code completion tools? `ImageShow.show` and `ImageShow.display` are very similar and could be confusing to someone using it for the first time.

src/PIL/ImageShow.py
```diff
@@ -218,6 +218,20 @@ def get_command_ex(self, file, title=None, **options):
...
+except ImportError:
+    pass
+else:
+    register(IPythonViewer)
```
nulano, line 232: Any reason to put `IPythonViewer` before the unix viewers but after the other Windows/macOS ones? I think it would be best to put this at the top (or use `register(..., order=0)`) to help consistency in IPython (e.g. when combined with tools that offer PDF exports, etc.).

Tests/test_imageshow.py
```diff
@@ -62,4 +62,20 @@ def test_viewer():
 
 def test_viewers():
     for viewer in ImageShow._viewers:
-        viewer.get_command("test.jpg")
+        try:
+            viewer.get_command("test.jpg")
+        except NotImplementedError:
+            pass
```
nulano, line 68:
```suggestion
            assert viewer.show(hopper()) == 1
```
Would remove the need for a separate test.

nulano (approved):
This should be very useful! Even if it takes a while until the changes reach Colab, it is not difficult to upgrade Pillow manually anyway.

src/PIL/ImageShow.py
radarhere: Okay, sure. I went with `ipython_display`, to be even clearer.

src/PIL/ImageShow.py
radarhere: Yep, no reason. I've moved it up.

Tests/test_imageshow.py
radarhere: The separate test
- confirms that the IPythonViewer is present when IPython is available
- runs IPythonViewer even if another Viewer instance has priority

Also, IPython seems to be a special case, where running its Viewer doesn't disrupt the test process - thinking generically for the sake of the future, just because a Viewer isn't based on a command line instruction doesn't mean that we want it to run in the middle of the test suite?

docs/releasenotes/8.2.0.rst
```diff
@@ -21,10 +21,17 @@ TODO
...
-^^^^
+ImageShow.IPythonViewer
+^^^^^^^^^^^^^^^^^^^^^^^
 
-TODO
+If IPython is present, this new ``ImageShow.Viewer`` subclass will be
+registered. It displays images on all IPython frontends. This will be helpful
+to users of Google Colab, allowing ``im.show()`` to display images.
+
+It is lower in priority than the other default Viewer instances, so it will
```
nulano, line 31: > It is lower in priority than the other default Viewer instances

Please correct me if I'm wrong, but this is false. The new viewer has the highest priority AFAICT

radarhere: You're right! Thanks for catching that. However, the documentation correctly states my intention, so I've updated the code to match it.

Issue #5287, opened 2021-02-26
Release Pillow 8.2.0 on April 1, 2021


radarhere:
Some things to check before release:

* [x] https://github.com/python-pillow/Pillow/pull/5364#issuecomment-810635018: fuzz crash
* [x] https://github.com/python-pillow/Pillow/pull/5291: release notes
* [X] https://github.com/python-pillow/pillow-wheels/pull/181 is needed for the Raqm/FriBiDi changes
* [x] Fuzzers are good
* [x] https://github.com/python-pillow/pillow-security/pull/9 in the pillow-security repo. Release notes should be basically the commit messages

## Main Release

Released quarterly on January 2nd, April 1st, July 1st and October 15th.

* [x] Open a release ticket e.g. https://github.com/python-pillow/Pillow/issues/3154
* [x] Develop and prepare release in `master` branch.
* [x] Check [GitHub Actions](https://github.com/python-pillow/Pillow/actions) and [AppVeyor](https://ci.appveyor.com/project/python-pillow/Pillow) to confirm passing tests in `master` branch.
* [x] Check that all of the wheel builds [Pillow Wheel Builder](https://github.com/python-pillow/pillow-wheels) pass the tests in Travis CI and GitHub Actions.
* [x] In compliance with [PEP 440](https://www.python.org/dev/peps/pep-0440/), update version identifier in `src/PIL/_version.py`
* [x] Update `CHANGES.rst`.
* [x] Run pre-release check via `make release-test` in a freshly cloned repo.
* [x] Create branch and tag for release e.g.:
  ```bash
  git branch 5.2.x
  git tag 5.2.0
  git push --all
  git push --tags
  ```
* [x] Create and check source distribution:
  ```bash
  make sdist
  twine check dist/*
  ```
* [x] Create [binary distributions](https://github.com/python-pillow/Pillow/blob/master/RELEASING.md#binary-distributions)
* [x] Check and upload all binaries and source distributions e.g.:
  ```bash
  twine check dist/*
  twine upload dist/Pillow-5.2.0*
  ```
* [x] Publish the [release on GitHub](https://github.com/python-pillow/Pillow/releases)
* [x] In compliance with [PEP 440](https://www.python.org/dev/peps/pep-0440/), increment and append `.dev0` to version identifier in `src/PIL/_version.py`

### Windows
* [x] Contact `@cgohlke` for Windows binaries via release ticket e.g. https://github.com/python-pillow/Pillow/issues/1174.
[...]

PR #5291, opened 2021-02-26
Modify Gif.h and GifEncode.c to use LZW encoding

### LINKED ISSUE #5278: GIF writing should use LZW encoding
### What did you do?
Create a GIF
It should use LZW encoding, as most GIF applications do.
### What actually happened?
It used Fredrik Lundh's cleverly compatible but suboptimal encoding from 1999.

* OS: Windows, Linux
* Python: 3.7, 3.8, 3.9, any other
* Pillow: Doesn't matter; does this on everything since 1999.

This isn't a bug report, it's an enhancement request. I'm pretty sure Fredrik did it this way to avoid the then-current LZW patent problems with Unisys. Those ended in 2004 after all Unisys patents on LZW expired. I'm puzzled why, 17 years later, LZW has been barely mentioned anywhere regarding GIFs from PIL. I see a mention in #617 and #4644. Both of these say GIFs are currently "uncompressed"; that's not quite right. Fredrik had a clever sort-of-run-length-encoding technique. I think using LZW still won't fix all the concerns in #617, I think PIL could use more work for animated GIFs but would need to look into that more.

I see in #617 "Filesize concerns" an invite from @wiredfool "If you’d like to contribute a gif compressor under the PIL license, go for it."

I have one ready to go, well tested at my end in Windows and Linux (WSL2 and Raspios). It's a mod to Gif.h and a major overhaul of GifEncode.c. A drop-in replacement that does not require any other changes.

PIL currently writes all GIFs as 256-color, with full 8 bits in the uncompressed pixels (as in tobytes()). My GifEncode.c writes them this way, but it is ready to go for shorter codes (smaller color tables) when the rest of PIL is (probably GifImagePlugin.py and others).

raygard:
Fixes #5278 .

 * Modify Gif.h and GifEncode.c to use LZW encoding

After this has been done and the Tests/images/hopper_resized.gif file has been recreated with LZW encoding (per 72a098b mod to test_image_resize.py by @radarhere), that same test can be used to validate against regression, "making this an appropriate test to ensure that we don't break your code in the future." (quoting @radarhere)

I have tested this extensively. The best test I can provide now is this: [test_giflzw.py](https://gist.github.com/raygard/4c0c531f7106f60fdba90e60d0e13f16) . As described in the doc comment there, you need a build with the current (older) GifEncode.c named PIL and another with the new version named PIL_giflzw. You then run the program on image files (GIFs, but some other types will also work). It will open and save each file using both the old and the new encoders, to in-memory io.BytesIO buffers and optionally to actual files. It then re-opens them and retrieves the unencoded pixel bytes using .tobytes() and compares. I've never had a compare fail to match.

The new encoder is a little slower than the old, but typically yields GIF files about 70% the size of the old encoder.

If anyone is equipped to run a static or dynamic checker against my code (e.g. valgrind or anything else), I would welcome that. It currently compiles clean on both Windows and Linux.

The whole modules can be found in https://github.com/raygard/Pillow/blob/giflzw/src/libImaging/Gif.h and https://github.com/raygard/Pillow/blob/giflzw/src/libImaging/GifEncode.c .


wiredfool:
If you can rebase on current master, there's now a Valgrind github action.

src/libImaging/GifEncode.c
```diff
@@ -21,136 +22,199 @@
...
+        /* code bits to pack */
+        UINT32 n = st->buf_bits_left < st->code_bits_left
+                        ? st->buf_bits_left : st->code_bits_left;
+        st->code_buffer |=
+                        (st->code & ((1 << n) - 1)) << (8 - st->buf_bits_left);
+        st->code >>= n;
+        st->buf_bits_left -= n;
+        st->code_bits_left -= n;
+        if (st->code_bits_left)
+            goto check_buf_bits;
```
radarhere, line 160:
```suggestion
        if (st->code_bits_left) {
            goto check_buf_bits;
        }
```
Just a style change, see #4492

raygard: Done. Thanks for catching that. I tried to conform to the current Pillow style but missed that one.

wiredfool:
LGTM. 

One more thing that I'd ask -- can you link any references for the LZW format that you used (as comments), so if we do have to dig into this, we have a reference to what you were trying to do at the time? We'll also need to note something in the release notes. (#5287)

I'd like to get this into the 8.2.0 release, I think it's looking good enough at this point.

raygard:
I have added a comment referencing the original GIF LZW spec from Compuserve. I hope that's what you wanted.

hugovk:
@raygard Please could you also add this to the release notes?

https://github.com/python-pillow/Pillow/blob/master/docs/releasenotes/8.2.0.rst

raygard:
@hugovk Not sure if I did this correctly, as I just copied the latest release notes into my own ``giflzw`` branch before adding this, under "Other Changes":

```
GIF writer uses LZW encoding
^^^^^^^^^^^^^^^^^^^^^^^^^^^^

GIF files are now written using LZW encoding, which will generate smaller files,
typically about 70% of the size generated by the older encoder.

The pixel data is encoded using the format specified in the [Compuserve GIF
standard](https://www.w3.org/Graphics/GIF/spec-gif89a.txt). The older encoder
used a variant of run-length encoding that was compatible but less efficient.
```

I see that git/github says it has conflicts to be resolved, but it looks to me like a simple insertion of a few lines. I don't understand git enough to see why it causes a conflict.

raygard:
@hugovk I don't know why the link does not render correctly when I view the release notes at https://github.com/python-pillow/Pillow/blob/master/docs/releasenotes/8.2.0.rst. 

I see literally:

\[Compuserve GIF standard\](https://www.w3.org/Graphics/GIF/spec-gif89a.txt). 

Instead of  [Compuserve GIF standard](https://www.w3.org/Graphics/GIF/spec-gif89a.txt). 

No big deal, but wonder why.

hugovk:
Oh yeah, it's because it's reStructuredText not Markdown, so has a different syntax:

https://docutils.sourceforge.io/docs/user/rst/quickref.html#hyperlink-targets

I'll fix it. Thanks for checking!

raygard:
That’s on me for not noticing it’s .rst not .md. I’ve written a lot more .rst than .md and should have noticed! Thank you for fixing it. 

BTW there is some code and documentation at https://github.com/raygard/giflzw and https://raygard.github.io/giflzw/ regarding the development of this LZW encoder. In particular, there is an explanation and justification for the Duff's-device-like structure at https://raygard.github.io/giflzw/pages/about_the_code.html .

Issue #4516, opened 2020-04-02
RGBA Image.resize is lossy because of RGBa conversion


crmauceri:
### What did you do?
I have a RGB-D image (RGB plus depth channel). I load it as a RGBA image in PIL for image processing. However after the Image.resize operation, the values in the RGB channels that correspond to `A==0` are set to zero. 

### What did you expect to happen?
I would expect the RGB values to stay the same, and only the size of the image to change.

### What actually happened?
Any pixels where the alpha channel was zero, became zero in all channels. 

Here's an example visualization. The code to reproduce it is below. 
<img width="667" alt="RGBA_visualization" src="https://user-images.githubusercontent.com/12262946/78306123-76f91180-74f7-11ea-8a99-e1748511acde.png">

### Why did this happen? 
It appears this is because the Image.resize function calls self.convert('RGBa') and after resizing the image then converts back to RGBA. RGBa is "true color with premultiplied alpha" while RGBA is "true color with transparency mask" [1](https://pillow.readthedocs.io/en/4.0.x/handbook/concepts.html#modes). It appears to be a lossy operation. 

### What are your OS, Python and Pillow versions?

* OS: MacOSX 10.14
* Python: 3.7
* Pillow: 6.2.1

### MCVE
I'm using this example image as a stand in for my RGB image.  

![glass.png](https://i.stack.imgur.com/fE2Su.png)

The depth channel is stored as a seperate png image, but for this MCVE I've used a numpy array 

```python
from PIL import Image
import matplotlib.pyplot as plt
import numpy as np

img_path = 'glass.png'
img = Image.open(img_path).convert('RGB')

# Load depth from file
#depth_path = 'glass_d.png'
#depth = Image.open(depth_path).convert('L')

# Toy depth for code example
depth_arr = np.zeros((180,180))
depth_arr[50:150, 50:100] = 255
depth = Image.fromarray(depth_arr).convert('L')

# Add the depth as alpha channel
img.putalpha(depth)

# Make a copy and resize to demonstrate problem
img_resize = img.copy()
img_resize = img_resize.resize((250, 250), Image.NEAREST) 

# Visualization of result 
plt.figure()
plt.subplot('221')
plt.imshow(img)
plt.title('Image before resize')

plt.subplot('222')
plt.imshow(np.array(img)[:, :, :3])
plt.title('RGB channels before resize')

plt.subplot('223')
plt.imshow(img_resize)
plt.title('Image after resize')

plt.subplot('224')
[...]

PR #5304, opened 2021-03-03
Do not premultiply alpha when resizing with Image.NEAREST resampling

### LINKED ISSUE #5300: RGBA Image.resize with Nearest Neighbor Interpolation introduces new RGB Values
### What did you do?

Resizing an RGBA image with Image.resize and nearest neighbor interpolation.

Image resized to given size, no new RGB values introduced.

### What actually happened?

New RGB values introduced.

* OS: Linux
* Python: 3.8.6
* Pillow: 6.2.1

### Python Example

```python
import numpy as np
from PIL import Image

image = np.asarray([[[1, 1, 0, 128]], [[2, 2, 0, 64]]], dtype=np.uint8)

print("Input")
print(image)
print("")

image = Image.fromarray(image)
image = image.resize((1, 1), Image.NEAREST)
image = np.asarray(image)

print("Output")
print(image)
```

```
Input
[[[  1   1   0 128]]
 [[  2   2   0  64]]]

Output
[[[ 3  3  0 64]]]
```

nulano:
Fixes #5300.

Test is based on the previous tests that make sure premultiplied alpha is used with BILINEAR resampling.

From https://pillow.readthedocs.io/en/stable/handbook/concepts.html#PIL.Image.NEAREST:

> Pick one nearest pixel from the input image. Ignore all other input pixels.

I would interpret that as meaning that each pixel in the output will be equal to one of the original pixels, i.e. not introducing new values.

### DIFF
--- a/Tests/test_image_transform.py
+++ b/Tests/test_image_transform.py
@@ -145,9 +145,8 @@ class TestImageTransform:
 
     def _test_nearest(self, op, mode):
         # create white image with half transparent,
-        # with the black half transparent.
         # do op,
-        # the image should be white with half transparent
+        # the image should remain white with half transparent
         transparent, opaque = {
             "RGBA": ((255, 255, 255, 0), (255, 255, 255, 255)),
             "LA": ((255, 0), (255, 255)),
--- a/Tests/test_image_transform.py
+++ b/Tests/test_image_transform.py
@@ -143,6 +143,42 @@ class TestImageTransform:
 
         self._test_alpha_premult(op)
 
+    def _test_nearest(self, op, mode):
+        # create white image with half transparent,
+        # with the black half transparent.
+        # do op,
+        # the image should be white with half transparent
+        transparent, opaque = {
+            "RGBA": ((255, 255, 255, 0), (255, 255, 255, 255)),
+            "LA": ((255, 0), (255, 255)),
+        }[mode]
+        im = Image.new(mode, (10, 10), transparent)
+        im2 = Image.new(mode, (5, 10), opaque)
+        im.paste(im2, (0, 0))
+
+        im = op(im, (40, 10))
+
+        colors = im.getcolors()
+        assert colors == [
+            (20 * 10, opaque),
+            (20 * 10, transparent),
+        ]
+
+    @pytest.mark.parametrize("mode", ("RGBA", "LA"))
+    def test_nearest_resize(self, mode):
+        def op(im, sz):
+            return im.resize(sz, Image.NEAREST)
+
+        self._test_nearest(op, mode)
+
+    @pytest.mark.parametrize("mode", ("RGBA", "LA"))
+    def test_nearest_transform(self, mode):
+        def op(im, sz):
+            (w, h) = im.size
+            return im.transform(sz, Image.EXTENT, (0, 0, w, h), Image.NEAREST)
+
+        self._test_nearest(op, mode)
+
     def test_blank_fill(self):
         # attempting to hit
         # https://github.com/python-pillow/Pillow/issues/254 reported
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -1910,7 +1910,7 @@ class Image:
         if self.mode in ("1", "P"):
             resample = NEAREST
 
-        if self.mode in ["LA", "RGBA"]:
+        if self.mode in ["LA", "RGBA"] and resample != NEAREST:
             im = self.convert(self.mode[:-1] + "a")
             im = im.resize(size, resample, box)
             return im.convert(self.mode)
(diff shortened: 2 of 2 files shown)


Tests/test_image_transform.py
```diff
@@ -143,6 +143,46 @@ def op(im, sz):
...
+        # do op,
+        # the image should be white with half transparent
+        if mode == "RGBA":
+            transparent = (255, 255, 255, 0)
+            opaque = (255, 255, 255, 255)
+        elif mode == "LA":
+            transparent = (255, 0)
+            opaque = (255, 255)
+        else:
+            assert False, "bad test"
```
radarhere, line 158:
```suggestion
```
You wouldn't rather avoid adding lines that deliberately don't have coverage? If this method was used with another mode, I feel like it would be hard to miss this conditional block.

nulano: I put that in to stop my IDE complaining about uninitialized values, but you're right, not a great solution. I replaced the if with a dict lookup.

radarhere: Thanks

radarhere:
So, in what I presume is an extreme situation to allow for ease of testing, your test demonstrates that converting the image to pre-multiplied alpha and then back to normal alpha causes a loss of information when the alpha is zero - because any color with zero alpha becomes black when converted to pre-multiplied alpha.

It may not be related, but I find that if I take one of the values from the issue, and convert it to RGBa and then back again,
```pycon
>>> from PIL import Image
>>> im = Image.new("RGBA", (1, 1), (2, 2, 0, 64))
>>> im.convert("RGBa").load()[0, 0]
(1, 1, 0, 64)
>>> im.convert("RGBa").convert("RGBA").load()[0, 0]
(3, 3, 0, 64)
```
the roundtrip changes the value. That's odd.

nulano:
>It may not be related, but I find that if I take one of the values from the issue, and convert it to RGBa and then back again, [...] the roundtrip changes the value. That's odd.



It's not really that odd when you consider what the operation is doing. Premultiplying with alpha 64, you are rescaling each color from 0-255 to 0-64 range, effectively losing two least significant bits. Therefore there are four different values for each RGBA color band that all map to the same value in RGBa. Converting back to RGBA then chooses the "middle" value.

```python
>>> from PIL import Image
>>> Image.new("RGBA", (1,1), (1,1,0,64)).convert("RGBa").load()[0,0]
(0, 0, 0, 64)
>>> Image.new("RGBA", (1,1), (2,2,0,64)).convert("RGBa").load()[0,0]
(1, 1, 0, 64)
>>> Image.new("RGBA", (1,1), (3,3,0,64)).convert("RGBa").load()[0,0]
(1, 1, 0, 64)
>>> Image.new("RGBA", (1,1), (4,4,0,64)).convert("RGBa").load()[0,0]
(1, 1, 0, 64)
>>> Image.new("RGBA", (1,1), (5,5,0,64)).convert("RGBa").load()[0,0]
(1, 1, 0, 64)
>>> Image.new("RGBA", (1,1), (6,6,0,64)).convert("RGBa").load()[0,0]
(2, 2, 0, 64)
>>> Image.new("RGBa", (1,1), (1,1,0,64)).convert("RGBA").load()[0,0]
(3, 3, 0, 64)
```

radarhere:
Thanks. So that's why NEAREST should not roundtrip through pre-multiply conversion - since it changes the non-alpha channel values.

But have you figured out what the original purpose of it was? In other words, why it's done for any resampling methods? It's not just a shortcut method, as it actually produces different results. I would like to understand the reason so that I know why it's ok to disregard that reason in this case.

nulano:
The reason to use RGBa is explained by `@wiredfool` here: https://github.com/python-pillow/Pillow/issues/4516#issuecomment-608330507

When using NEAREST downsampling, the conversion only discards some pixels keeping the ones nearest to the result image grid.
All other downsampling methods use some sort of averaging of a group of pixels near the output pixel, which can cause the colors of transparent pixels to change the colors of non-transparent pixels. See the tests above this one for an example (`test_alpha_premult_resize`).

radarhere (approved):

radarhere:
> All other resampling methods use some sort of averaging of a group of pixels near the output pixel

https://pillow.readthedocs.io/en/stable/handbook/concepts.html#PIL.Image.BOX
> **PIL.Image.BOX**
> Each pixel of source image contributes to one pixel of the destination image with identical weights. For upscaling is equivalent of NEAREST.

So that makes BOX an exception, yes?

nulano:
> > All other resampling methods use some sort of averaging of a group of pixels near the output pixel
> 
> https://pillow.readthedocs.io/en/stable/handbook/concepts.html#PIL.Image.BOX
> 
> > **PIL.Image.BOX**
> > Each pixel of source image contributes to one pixel of the destination image with identical weights. For upscaling is equivalent of NEAREST.
> 
> So that makes BOX an exception, yes?

Ah, yes, but only for upsampling. Downsampling produces an average of multiple pixels, with each included pixel having the same weight.

PR #5305, opened 2021-03-03
Fix -Wformat error in TiffDecode

lukegb:
When building under an environment with -Werror=format, TiffDecode.c no longer builds:

```
src/libImaging/TiffDecode.c: In function ‘_tiffReadProc’:
src/libImaging/TiffDecode.c:59:58: error: format ‘%d’ expects argument of type ‘int’, but argument 3 has type ‘toff_t’ {aka ‘long unsigned int’} [-Werror=format=]
src/libImaging/TiffDecode.c:59:67: error: format ‘%d’ expects argument of type ‘int’, but argument 4 has type ‘toff_t’ {aka ‘long unsigned int’} [-Werror=format=]
```

### DIFF
--- a/src/libImaging/TiffDecode.c
+++ b/src/libImaging/TiffDecode.c
@@ -56,7 +56,7 @@ _tiffReadProc(thandle_t hdata, tdata_t buf, tsize_t size) {
     dump_state(state);
 
     if (state->loc > state->eof) {
-        TIFFError("_tiffReadProc", "Invalid Read at loc %d, eof: %d", state->loc, state->eof);
+        TIFFError("_tiffReadProc", "Invalid Read at loc %lu, eof: %lu", state->loc, state->eof);
         return 0;
     }
     to_read = min(size, min(state->size, (tsize_t)state->eof) - (tsize_t)state->loc);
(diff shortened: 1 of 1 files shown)


radarhere:
Building normally, I find that there is currently a
>  warning: format specifies type 'unsigned long' but the argument has type 'toff_t' (aka 'unsigned long long') [-Wformat]

With your change, it becomes
> warning: format specifies type 'int' but the argument has type 'toff_t' (aka 'unsigned long long') [-Wformat]

src/libImaging/TiffDecode.c
```diff
@@ -56,7 +56,7 @@ _tiffReadProc(thandle_t hdata, tdata_t buf, tsize_t size) {
     dump_state(state);
 
     if (state->loc > state->eof) {
-        TIFFError("_tiffReadProc", "Invalid Read at loc %d, eof: %d", state->loc, state->eof);
+        TIFFError("_tiffReadProc", "Invalid Read at loc %lu, eof: %lu", state->loc, state->eof);
```
radarhere, line 59:
```suggestion
        TIFFError("_tiffReadProc", "Invalid Read at loc %llu, eof: %llu", state->loc, state->eof);
```
This resolves the warnings for me.

lukegb:
Hmm, interesting.

I suspect there's a disparity in what underlying type these are being mapped to on your system vs mine - IIRC there are special format specifiers or macros for this case, but failing that it might make sense to explicitly cast to a different type like is being done above.

lukegb:
I tried to reply by email but GitHub apparently dropped it. Oh well.

PR #5315, opened 2021-03-06
Simplified code

radarhere:
I think this change simplifies the code. Feel free to disagree.


src/PIL/Image.py
```diff
@@ -2396,18 +2396,11 @@ def getdata(self):
         :returns: An :py:class:`~PIL.Image.Image` object.
         """
 
-        if self.mode == "LA":
+        if self.mode in ("LA", "RGBA"):
             return (
-                self.convert("La")
+                self.convert(self.mode.replace("A", "a"))
```
nulano, line 2401:
```suggestion
                self.convert(self.mode[:-1] + "a")
```
To match `resize()`.

wiredfool:
I don't like this, while it does remove a duplicated bit of code, it obscures what's happening. 

In my opinion, 
* Modes are essentially enums that happen to be strings. Doing string.replace is kind of weird on an enum. 
* It reduces greppability. It's nice to be able to grep the code and find everywhere `La` is referenced. 

If really want to do something to remove the duplicate code, perhaps a dict with a mapping of `{'LA': 'La', 'RGBA':'RGBa'}` would be appropriate.

radarhere:
I've pushed a commit to use the mapping option, and in two other locations as well.

hugovk (approved):

PR #5323, opened 2021-03-11
Only import numpy when necessary

### LINKED ISSUE #5320: Kodi v19 crashes when PIL imports numpy
This happens on Arch Linux with all addons and skins that use PIL image libraries.
Information on TheMovieDb Helper Kodi addon where its dev did a workaround here:
https://github.com/jurialmunkey/plugin.video.themoviedb.helper/issues/445

A Kodi dev mentioned that this is a PIL / Numpy issue not a Kodi issue so I should ask here:
https://github.com/xbmc/xbmc/issues/19324

Thank you for any possible solutions.

radarhere:
Resolves #5320

Moves `ImageFilter` numpy import into the `ImageFilter.Color3DLUT` constructor, so that numpy is not imported only to be unused.

### DIFF
--- a/src/PIL/ImageFilter.py
+++ b/src/PIL/ImageFilter.py
@@ -16,11 +16,6 @@
 #
 import functools
 
-try:
-    import numpy
-except ImportError:  # pragma: no cover
-    numpy = None
-
 
 class Filter:
     pass
@@ -368,6 +363,13 @@ class Color3DLUT(MultibandFilter):
         copy_table = kwargs.get("_copy_table", True)
         items = size[0] * size[1] * size[2]
         wrong_size = False
+
+        numpy = None
+        if hasattr(table, "shape"):
+            try:
+                import numpy
+            except ImportError:  # pragma: no cover
+                pass
 
         if numpy and isinstance(table, numpy.ndarray):
             if copy_table:


basilgello:
And still adding an optional parameter to Color3DLut constructor disabling numpy usage might be perfect solution for those who want to stop using numpy within Pillow (for example if a buggy version installed in the system is out of their control).

radarhere:
But that wouldn't help your situation, because you're not the one calling `Color3DLUT`, right?

basilgello:
But for situations a 3rd party developer does call 3DLut but realizes nu,py is buggy?

radarhere:
I'm not sold on the idea that Pillow should be adding workarounds for uncommon numpy bugs.

That said, I am willing to accept that from an efficiency standpoint, it could be better for Pillow not to import numpy if it isn't actually going to be used.

I've updated this PR, so that it now does a rudimentary check to see if `table` is a numpy object. This creates a clear path for Pillow users to not import numpy in ImageFilter - just don't pass a numpy object to Color3DLut.

basilgello:
Sounds great for me! Let me check it then, and I think we can mark the issue solved on Pillow side
-- 
Vasyl Gello
==================================================
Certified SolidWorks Expert

Mob.:+380 (98) 465 66 77

***@***.***

Skype: vasek.gello
==================================================
호랑이는 죽어서 가죽을 남기고 사람은 죽어서 이름을 남긴다

hugovk:
@basilgello Hi! Have you checked this PR yet? If it's good we can include it in Thursday's release. Thanks!

basilgello:
Yes, LGTM and is probably a low-risk change! Thanks for the heads-up!

cgcgcgcgcg:
The python-pillow update on 2021-04-23 fixed my issue with Kodi 19, but one day later python-numpy was updated and crashes started again.
So I'm back to using Flatpak Kodi until this is sorted.

radarhere:
Let us know if you feel there is anything for us to do - but from discussion in #5320, I would think we've now done all we can.

PR #5326, opened 2021-03-14
Improved efficiency when creating GIF disposal images

radarhere:
Two changes.

1. Do not call `self.im.copy()`, only to then call `self._crop` - `self._crop` already creates a copy of the image.
2. Instead of creating a solid color image of `self.size` and then cropping it, just create the solid color image at the necessary size.


hugovk:
Thanks!

PR #3555, opened 2019-01-07
Add TIFF compression codecs: LZMA, Zstd, WebP

cgohlke:
Windows binaries are linked to a libtiff library, which has LZMA (Python >=3.5), WebP, and ZStd compression codecs enabled. However, Pillow currently fails to open TIFF files that use these compressions.

PR #5336, opened 2021-03-17
Updated list of TIFF compression methods

### LINKED ISSUE #5335: Clarify saving LZW-compressed TIFFs in the documentation
The [documentation about saving TIFF images](https://pillow.readthedocs.io/en/latest/handbook/image-file-formats.html#saving-tiff-images) says that

> Valid compression methods are: None, "tiff_ccitt", "group3", "group4", "tiff_jpeg", "tiff_adobe_deflate", "tiff_thunderscan", "tiff_deflate", "tiff_sgilog", "tiff_sgilog24", "tiff_raw_16"

However, I can create LZW-compressed TIFFs by passing a "tiff_lzw" value:

```python
image.save("image.tif", compression="tiff_lzw")
```

Since LZW is a commonly used compression method, I think it would be good to clarify what's going on here:

* Does it accidentally happen to work but with problems I'm not seeing?
* Is the support dependent on the user's software configuration?
* Is the value "tiff_lzw" officially supported and the fact that it isn't in the documentation is an oversight?

radarhere:
Resolves #5335, that states that the "compression" list at https://pillow.readthedocs.io/en/latest/handbook/image-file-formats.html#saving-tiff-images is incorrect.

I've added "tiff_lzw", "jpeg", "packbits", "lzma", "zstd" and "webp" from https://github.com/python-pillow/Pillow/blob/8064e04e963e3061918564d3c5fff35074577f29/src/PIL/TiffImagePlugin.py#L109-L128
and removed "tiff_jpeg" (since it is obsolete, see #4627, let's stop advertising it).

### DIFF
--- a/docs/handbook/image-file-formats.rst
+++ b/docs/handbook/image-file-formats.rst
@@ -876,10 +876,10 @@ The :py:meth:`~PIL.Image.Image.save` method can take the following keyword argum
 **compression**
     A string containing the desired compression method for the
     file. (valid only with libtiff installed) Valid compression
-    methods are: :data:`None`, ``"tiff_ccitt"``, ``"group3"``,
-    ``"group4"``, ``"tiff_jpeg"``, ``"tiff_adobe_deflate"``,
-    ``"tiff_thunderscan"``, ``"tiff_deflate"``, ``"tiff_sgilog"``,
-    ``"tiff_sgilog24"``, ``"tiff_raw_16"``
+    methods are: :data:`None`, ``"tiff_ccitt"``, ``"group3"``, ``"group4"``,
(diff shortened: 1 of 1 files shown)


hugovk:
If there's no special meaning in the order, let's alphabetise them.

radarhere:
Ok, done

kmilos:
Just out of curiosity, any reason for the `tiff_` prefix while you're at it? I don't think there's anything TIFF specific about lzw, deflate, etc...

radarhere:
At this point, backwards compatibility? PIL had some with the `tiff_` prefix and some without, and I guess #268 and #3555 just went different ways on this.

I've added a commit removing `tiff_deflate` from the list, as suggested in https://github.com/python-pillow/Pillow/issues/5335#issuecomment-801742691

kmilos:
Great, thanks!

PR #5349, opened 2021-03-21
Add ImageShow support for GraphicsMagick

latosha-maltba:
GraphicsMagick[0] is a "stable/conservative" fork of ImageMagick[1]. It provides the same commands and is almost always interchangable with the corresponding imageMagick command. The ImageMagick command `<im-command> <arguments>` becomes `gm <im-command> <arguments>`, e.g. `convert file.jpg file.png` becomes `gm convert file.jpg file.png`.

This pull request does two things:

* Update the test suite to run with either ImageMagick or GraphicsMagick (only `test_file_palm.py` seems to use it)
* Update `ImageShow` to support `gm display`

[0] http://www.graphicsmagick.org/
[1] http://www.graphicsmagick.org/mission.html

### DIFF
--- a/Tests/helper.py
+++ b/Tests/helper.py
@@ -257,16 +257,22 @@ def netpbm_available():
     return bool(shutil.which("ppmquant") and shutil.which("ppmtogif"))
 
 
-def convert_available():
-    return imagemagick_available() or graphicsmagick_available()
+def magick_command():
+    if sys.platform == "win32":
+        imagemagick = os.environ.get("MAGICK_HOME", "")
+        if imagemagick:
+            imagemagick = [os.path.join(imagemagick, "convert.exe")]
+            graphicsmagick = [os.path.join(imagemagick, "gm.exe"), "convert"]
+        else:
+            graphicsmagick = None
+    else:
+        imagemagick = ["convert"]
+        graphicsmagick = ["gm", "convert"]
 
-
-def imagemagick_available():
-    return bool(IMCONVERT and shutil.which(IMCONVERT[0]))
-
-
-def graphicsmagick_available():
-    return bool(GMCONVERT and shutil.which(GMCONVERT[0]))
+    if imagemagick and shutil.which(imagemagick[0]):
+        return imagemagick
+    elif graphicsmagick and shutil.which(graphicsmagick[0]):
+        return graphicsmagick
 
 
 def on_appveyor():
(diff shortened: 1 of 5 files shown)


hugovk:
Thank you for the PR, please could you also include this in the release notes?

https://github.com/python-pillow/Pillow/blob/master/docs/releasenotes/8.2.0.rst

latosha-maltba:
Sure!

Sidenote: Maybe its point worth adding this suggestion (i.e. when providing a pull request also update the releasenotes) to .github/CONTRIBUTING.md?

latosha-maltba:
The Docs fail in *Test / ubuntu-latest Python 3.9 (pull_request)* and I guess the culprit is the warning `WARNING: py:class reference target not found: PIL.ImageShow.GmDisplayViewer`.

Sadly I have no clue how to correctly reference the newly added class. The usual :py:class:`PIL.ImageShow.GmDisplayViewer` does not seem to work (might be because of `autoclass::` since in the same file :py:class:`PIL.ImageShow.Viewer` does work?). Help would be appreciated.

nulano:
Looking at the generated docs, I see that the link to the new viewer references `PIL.ImageShow.UnixViewer.GmDisplayViewer`: https://pillow--5349.org.readthedocs.build/en/5349/reference/ImageShow.html#PIL.ImageShow.UnixViewer.GmDisplayViewer

I'm not sure whether there is an obvious way to fix this without sacrificing the formating.

latosha-maltba:
@nulano Thanks for the hint and solving the mystery.  It seems expected and unexpected that Sphinx uses its own class hierarchy instead of the one of Python.

In the end I settled with :py:class:`PIL.ImageShow.GmDisplayViewer <PIL.ImageShow.UnixViewer.GmDisplayViewer>` so we have the full-qualified name (PIL.ImageShow.GmDisplayViewer) pointing to the correct (sub)class.

hugovk:
> Sidenote: Maybe its point worth adding this suggestion (i.e. when providing a pull request also update the releasenotes) to .github/CONTRIBUTING.md?

Good idea! We already had it in https://github.com/python-pillow/Pillow/blob/master/docs/releasenotes/index.rst but makes sense to include it in the contrib guide too. Please see PR https://github.com/python-pillow/Pillow/pull/5357. Thanks!

PR #5351, opened 2021-03-22
Deprecated categories

radarhere:
Images have a ``category`` property. It is set to ``Image.NORMAL``
https://github.com/python-pillow/Pillow/blob/aa35f6b572c3491976e2336b28f481dad0a2a353/src/PIL/Image.py#L538

Unless the image format is MIC and there are multiple frames, in which case it is ``Image.CONTAINER``.
https://github.com/python-pillow/Pillow/blob/aa35f6b572c3491976e2336b28f481dad0a2a353/src/PIL/MicImagePlugin.py#L70-L71

There is also the unused attribute ``Image.SEQUENCE``.
https://github.com/python-pillow/Pillow/blob/aa35f6b572c3491976e2336b28f481dad0a2a353/src/PIL/Image.py#L190-L193

This minimal support across formats shows the feature has clearly not kept up with our other changes around multi-frames images, and can be deprecated in favour of `is_animated`.

### DIFF
--- a/Tests/test_image.py
+++ b/Tests/test_image.py
@@ -1,6 +1,7 @@
 import io
 import os
 import shutil
+import sys
 import tempfile
 
 import pytest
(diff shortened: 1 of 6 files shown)


hugovk:
I'm fine with deprecating it.

It was a pain when we initally removed `Image.PILLOW_VERSION` without active deprecation warnings (only warnings in docs).

I expect calls to `im.category` are less common, but shall we nevertheless include deprecation warnings?

radarhere:
Ok, sure. I've added deprecation warnings.

However, not for Python 3.6, since `__getattr__` was only added in Python 3.7 - but Python 3.6 reaches EOL at the end of this year, so even if a user switched across to Python 3.7 at the start of 2022, they would still have a full year of seeing the warnings before the feature was removed.

hugovk:
Sounds good, thank you!

PR #5352, opened 2021-03-22
Document quantize method defaults and mode support

radarhere:
Alternative to #5251. The same information is added, just arranged differently.


nulano:
This looks more useful IMO.

PR #5359, opened 2021-03-24
Compile LibTIFF with CMake on Windows

nulano:
Alternative to #5243, see the discussion there for more information.

Switches winbuild to compile libtiff using CMake which enables `USE_WIN32_FILEIO` on all win32 platforms by default.
Compiling with NMake will not be supported when the next version of libtiff is released, see https://gitlab.com/libtiff/libtiff/-/merge_requests/223.


kmilos:
Great - I reckon 8.2 is a good time to make the switch and I/O API change as any, cmake has been supported by libtiff for a while already...

setup.py
```diff
@@ -727,9 +727,7 @@ def build_extensions(self):
         if feature.tiff:
             libs.append(feature.tiff)
             defs.append(("HAVE_LIBTIFF", None))
-            # FIXME the following define should be detected automatically
-            #       based on system libtiff, see #4237
-            if PLATFORM_MINGW:
+            if sys.platform == "win32":
                 defs.append(("USE_WIN32_FILEIO", None))
```
radarhere, line 731: So if this PR "switches winbuild to compile libtiff using CMake which enables USE_WIN32_FILEIO on all win32 platforms by default"... why do we need this def?

radarhere:
> Compiling with NMake will not be supported when the next version of libtiff is released, it might be a good idea to wait for that release before merging this PR.

The downside of waiting is that when it is released (which could happen at any point in time), we will be failing to support the latest libtiff until our next scheduled release.

setup.py
kmilos: Have to keep Pillow's `src/libImaging/TiffDecode.c` in sync w/ the way libtiff was built, i.e. because of the whole mess behind #4237 (this fix was intoduced in introduced #4890); maybe better leave the comment for posterity?

Unfortunately [libtiff doesn't expose this build time defined macro in a public header](https://gitlab.com/libtiff/libtiff/-/issues/173#note_503495775) like it does for some other (similar) stuff...

kmilos:
Btw, I found it interesting that the [vcpkg build of libitiff turns off Win32 I/O for UWP only](https://github.com/microsoft/vcpkg/blob/a1e67195f84af0bf3d2c75b61128e4d10684347a/ports/tiff/cmakelists.patch#L23)... Maybe it really is not available on the phone?

All other Windows builds I know (libtiff cmake & autotools default, mingw, vcpkg, conda-forge, ...) have it enabled.

setup.py
kmilos: Alternatively we could do away with this case and define completely, and just use _WIN32 in TiffDecode.c directly if we take a leap of faith and assume all Windows libtiff builds will have Win32 I/O enabled, like e.g. [Poppler does since forever](https://github.com/freedesktop/poppler/blame/42dde686bf5a674401850b2d3fdd2bc7467e9a66/goo/TiffWriter.cc#L165).

nulano: LibTIFF also checks this define in `tiffio.h` when typedefing `thandle_t`. While all alternatives have the same binary size, it is probably safest to define the macro to make sure. It is also probably simpler to set it in `setup.py` instead of just `TiffDecode.c`.

I added a longer explanatory comment to this line:

> This define needs to be defined if-and-only-if it was defined
when compiling LibTIFF. LibTIFF doesn't expose it in `tiffconf.h`,
so we have to guess; by default it is defined in all Windows builds.
See #4237, #5243, #5359 for more information.

hugovk:
Thanks all!

PR #5366, opened 2021-03-29
Remove redundant check (addition to #5364)

kkopachev:
It's duplicated inside following call of TIFFReadTile

Since adding tiff reading using libtiff's TIFFReadRGBA* functions got extracted into it's own code path, there is no need to double check tiles. For more context https://github.com/python-pillow/Pillow/pull/5364#discussion_r603494127
Original commit to fix the CVE added a test image as well: https://github.com/python-pillow/Pillow/commit/cbdce6c5d054fccaf4af34b47f212355c64ace7a


kkopachev:
@wiredfool would you mind incorporating this too for the release?

wiredfool:
Ok, I see where this cam in in the manual merge I did.

PR #5264, opened 2021-02-12
Embed sort function in Quant.c to get same results in Windows and Linux

raygard:
Fixes #5263

Changes proposed in this pull request:

 * Embed sort in Quant.c in place of system qsort(). Avoids different results from different implementations of qsort() on different platforms.
 * quantize() and quantize2() can be static; they are not called outside Quant.c

PR #5367, opened 2021-03-30
Avoid unstable nature of qsort in Quant.c

### LINKED ISSUE #5263: Image.resize() can give different results in Windows than in Linux
### What did you do?
Resized a file in Linux and saved as GIF. Resized same file in Windows and saved as GIF.

I should get identical files in Windows and in Linux.

### What actually happened?
Files have different contents, sometimes different sizes. The differences are not visible to my eye, but I want reproducible and identical results when possible.

* OS: Ubuntu 20.04.1 LTS (GNU/Linux 4.19.128-microsoft-standard x86_64) (WSL2); Windows 10 Pro 64-bit build 18363.1316
* Python: 3.8.5 in Linux; 3.9.1 in Windows
* Pillow: 8.2.0.dev0 -- my own builds from recent source in both Linux and Windows

### Why does this happen?
I tracked the problem down to the use of `qsort()` in`build_distance_tables()` in `Quant.c`. `qsort()` is not a stable sort, so the different implementions in Windows and Linux can sort equal keys differently, leading to different quantization results. `build_distance_tables()` is called by `quantize()` and `quantize2()` which are called by `ImagingQuantize()` which is called by `_quantize()` in `_imaging.c`, which is called by `quantize()` and `convert()` in `Image.py`, and ultimately by `resize()`. (`convert()` is also called a lot of other places; I don't know if others will end up using `quantize()`.)

### How can this be corrected?
Use a sort routine embedded in `Quant.c` instead of calling the standard library `qsort()`. Not difficult. See (edited: I created a branch on my fork) https://github.com/raygard/Pillow/blob/quant_sort/src/libImaging/Quant.c for my fix. I've been using this for several months. (The sort routine is very well tested.)

I'm new at this and don't yet know how to create a pull request. (But I'm going to try...)

### How to reproduce my results and test my fix

See the code below. Run this on some not-too-small image (static, not animation), in Linux and Windows, using the current release or development version of PIL. I used `Pillow/Tests/images/old-style-jpeg-compression.png` for this test. The program will resize and save as GIF, and display the CRC of the resulting file.

Then rebuild PIL with my `Quant.c` in Linux and Windows and try it again. In my case, I get these results:
```
Linux, before:
ray@radon:/mnt/e/rdg/wsl/pil$ python3 test_quant_sort.py Pillow/Tests/images/old-style-jpeg-compression.png
fn='Pillow/Tests/images/old-style-jpeg-compression.png' im.mode='RGB' im.size=(4160, 870)
crc='a8b873ca'

Windows, before:
15:48:33.98 ++e:\rdg\pil>py -3 test_quant_sort.py Pillow-master\Tests\images\old-style-jpeg-compression.png
fn='Pillow-master\\Tests\\images\\old-style-jpeg-compression.png' im.mode='RGB' im.size=(4160, 870)
crc='481f4453'

Linux, after:
ray@radon:/mnt/e/rdg/wsl/pil$ python3 test_quant_sort.py Pillow/Tests/images/old-style-jpeg-compression.png
fn='Pillow/Tests/images/old-style-jpeg-compression.png' im.mode='RGB' im.size=(4160, 870)
crc='9160fc08'

Windows, after:
15:49:38.65 ++e:\rdg\pil>py -3 test_quant_sort.py Pillow-master\Tests\images\old-style-jpeg-compression.png
fn='Pillow-master\\Tests\\images\\old-style-jpeg-compression.png' im.mode='RGB' im.size=(4160, 870)
crc='9160fc08'
```

```python
import sys, io, os, binascii
from PIL import Image

def main():
    fn = sys.argv[1]
    with Image.open(fn).convert('RGB') as im:
        print(f'{fn=} {im.mode=} {im.size=}')
        factor = 0.2
        newsize = tuple(int(x*factor) for x in im.size)
        newim = im.resize(newsize, resample=Image.LANCZOS)
        b = io.BytesIO()
        newim.save(b, format='GIF')
        crc = '%08x' % (binascii.crc32(b.getvalue()) & 0xffffffff)
        print(f'{crc=}')
        newim.save(f'{os.path.splitext(os.path.basename(fn))[0]}_resized.gif')
        newim.close()

main()
```

radarhere:
Resolves #5263

https://en.wikipedia.org/wiki/Quicksort
> Efficient implementations of Quicksort are not a stable sort, meaning that the relative order of equal sort items is not preserved.

In other words, if `qsort` is given two equal items to sort, it doesn't have to return them in the same order on one platform vs another. This has lead to #5263.

#5264 solves this by adding a custom stable sort instead of `qsort`.

This PR is an alternative (and feel free to favour #5264 instead if you so choose). It instead keeps `qsort` but ensures that no two items are equal, by adding an index to the items, so that even if one dimension is equal, the second one will not be. It also reverts #5363 in the process.

Both PRs have the same code added to the test suite.


raygard:
I have looked over the code changes to ``Quant.c``. Looks correct to me. I have made sure it builds in Windows. I assume it's been built in Linux and assume it has been tested.

A small quibble: the Shell sort in #5264 is not stable. The object was to get the same sort order results on any platform, and incorporating the sort in ``Quant.c`` accomplishes that (assuming a deterministic sort!). Making the sort stable also makes it consistent, so that also works.

The Shell sort is only a little slower than Quicksort until the array gets fairly large (and can be faster, e.g. if the data are already nearly in order). Generally the compare function contributes heavily, so a more complex compare (as here) can slow things down. I have not tried to profile or time these two approaches so cannot say which will be faster in practice.

I'm the new guy here and @radarhere is a longtime member, so I feel I have to tread lightly. But I am curious as to the motivation. This seems to add more complexity than just using an embedded sort, and you've gone to rather a lot of effort to do it. Is there a concern about the performance, or maybe about the correctness of my Shell sort?

This approach is a sort of use of the decorate-sort-undecorate idiom. You need to allocate a new array of structs. I don't know how much memory this might take in the worst case, but based on the ``    if (nPaletteEntries > UINT32_MAX / nPaletteEntries) {`` check on lines 1182 and 1389, it appears there can be up to 2**16 palette entries, so up to ``UINT32_MAX`` elements of ``DistanceWithIndex`` allocated. That's about 32GB! I don't know what other limits there are on ``nPaletteEntries`` so that may not be a problem in practice.

But I see that there are only `nEntries` elements sorted at a time, and that you "undecorate" those as you sort each batch, so I suspect there is a way to only "decorate" `nEntries` at a time as well, just before each call to `qsort()`, so you'd need a lot less of `dwi`.

I also don't know if the added time to "decorate" and "undecorate" the ``DistanceWithIndex`` elements is of any significance. Probably minuscule compared with the rest of the work in quantization.

The name ``_sort_ulong_ptr_keys()`` is now no longer appropriate for the compare function.

Your construction of the ``DistanceWithIndex`` elements uses compound literals, which I think came in C99. Aside from `//` -style comments, are there other non-C90 features used in PIL? Does it matter?

You've not made ``quantize()`` and ``quantize2()`` static, as I had, but maybe that's best. They are not currently called outside ``Quant.c`` but maybe they could be in the future.

I'm sure you'll choose one or the other of these approaches, and I am fine with either, but if you go with keeping the system ``qsort()``, I think it would be good to see if you can maybe build the sort keys inside the loop that contains the call to ``qsort()`` and need only ``nEntries`` elements of ``dwi``.

raygard:
I have definitely not thought this through, but would this suffice? Build ``avgDistSortKey`` as was done originally in ``build_distance_tables()``, then:

```c
    for (i = 0; i < nEntries; i++) {
        for (j = 0; j < nEntries; j++) {
            dwi[j] = (DistanceWithIndex){
                avgDistSortKey[i * nEntries + j],
                j
            };
        };
        qsort(
            dwi,
            nEntries,
            sizeof(DistanceWithIndex),
            _sort_ulong_ptr_keys);
        for (j = 0; j < nEntries; j++) {
            avgDistSortKey[i * nEntries + j] = dwi[j].distance;
        }
    }
```

Only need ``nEntries`` elements in ``dwi``.  This does build in Windows and seems to work in a quick test.

radarhere:
> You've not made `quantize()` and `quantize2()` static, as I had, but maybe that's best. They are not currently called outside `Quant.c` but maybe they could be in the future.

I've created #5374, to separate out that change from this discussion.

radarhere:
My hope with this PR is that is adds less complexity - the nuances of a sort function seem more complicated to me than temporarily replacing a single variable with a struct. If you count it up, fewer lines are modified. And if our compare function is leaving room for interpretation, then I'd rather clarify that to address what I think of as the root of the problem.

I can appreciate that your opinion is probably that inserting a custom sort function is simpler from a big picture perspective. I guess my other concern is that we're potentially re-inventing the wheel by doing so.

Don't feel you have to tread lightly (I mean, politeness is still good... you know what I mean). I don't have strong opinions here. If your PR is merged instead, then onto the next issue.

You make a good point about the memory allocation. I've pushed a commit for that and to rename the function.

raygard:
@radarhere With your change to reduce memory (which reduces complexity as well), I have withdrawn my PR. Your newer version generates exactly the same results as my similar suggestion above; I tested. So I feel good about this version.

I do get your concerns about the sort function. My Shell sort was adapted for use as ``qsort()`` in the [uClibc](https://en.wikipedia.org/wiki/UClibc) stdlib (https://git.uclibc.org/uClibc/tree/libc/stdlib/stdlib.c or https://cgit.uclibc-ng.org/cgi/cgit/uclibc-ng.git/tree/libc/stdlib/stdlib.c near line 700) and various other places, and has been used for over 30 years, so I'm pretty confident in it :). Use it wherever you need a compact ``qsort()``.

raygard:
@radarhere , I would like to see this get merged in the upcoming release. Then I will not have to make my own Windows build anymore. Are there any blockers keeping this from getting merged?

radarhere:
Just a case of waiting for someone else from the Pillow team to review this.

hugovk:
Are there any discernible performance impacts from this, for example resizing large images?

radarhere:
Using https://photojournal.jpl.nasa.gov/jpeg/PIA24472.jpg (large enough that it triggers a DecompressionBombWarning), and
```python
import timeit
from PIL import Image

def test():
	im = Image.open("PIA24472.jpg")
	im.resize((int(im.width/2), int(im.height/2)))
	im.close()

print(timeit.timeit(test, number=100))
```

With this PR, I get 108.989, 108.575 and 108.075
With master, I get 109.324, 109.241 and 107.471

So no obvious differences in terms of speed.

hugovk:
How about resizing to something small, say `(10, 10)`?

radarhere:
On master, 62.347, 61.433 and 61.384.
With this PR, 62.190, 61.543 and 61.649.

So still no obvious difference.

hugovk:
Thanks for checking!

PR #5373, opened 2021-03-31
Fix pytest valgrind warnings

wiredfool:
Potential fix for pytest-valgrind mark not defined for non-valgrind docker tests.

### DIFF
--- a/Tests/conftest.py
+++ b/Tests/conftest.py
@@ -19,13 +19,11 @@ def pytest_configure(config):
     # We're marking some tests to ignore valgrind errors and XFAIL them.
     # Ensure that the mark is defined
     # even in cases where pytest-valgrind isn't installed
-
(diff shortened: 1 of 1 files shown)


hugovk:
Thanks! Confirmed fixed in a dummy commit: https://github.com/hugovk/Pillow/runs/2244123927?check_suite_focus=true

wiredfool:
There are a few XFAILed here because they're undefined behaviour and I haven't tracked them down yet. (some of which are deep in JPEG guts). The recent addition of the fuzzer test -- which runs all the images through open/load has retriggered some of them that are xfailed (I think) but I haven't had a chance to get to them. 

My idea for now is to prevent additional items from making valgrind worse. Which I'm failing at a bit with the fuzzer test, but to be fair, that's increased testing, not increased code to be tested.

PR #5379, opened 2021-04-03
Allow converting an image to a numpy array to raise errors

### LINKED ISSUE #5084: np.array(img) should not hide errors for corrupted images
### What did you do?
I spent a lot of time debugging a code which loads png images from disk and converts them to numpy arrays to do some processing.

If I take any png and corrupt it, say, by truncating,
```
$ dd if=good.png of=corrupted.png bs=1024 count=10
$ python
```
 I expect the following code to fail with some exception because image pixels can not be read into an array.
```pycon
>>> from PIL import Image
>>> import numpy as np
>>> img = Image.open('corrupted.png') 
>>> arr = np.array(img)
```

### What actually happened?
Instead, I got an object array with a single element:
```pycon
>>> print(arr)
<PIL.PngImagePlugin.PngImageFile image mode=RGB size=384x288 at 0x7F38258C6240>
```

and now the code fails later with a strange error message, like:
```pycon
>>> arr / 255
TypeError: unsupported operand type(s) for /: 'PngImageFile' and 'int'
```

#### img.verify() behaves strangely, too
Funny and counter-intuitive thing, if I try instead doing:  
```pycon
>>> img = Image.open('good.png') 
>>> img.verify()
>>> arr = np.array(img)
>>> arr / 255
TypeError: unsupported operand type(s) for /: 'PngImageFile' and 'int'
```
I still got an error presumably because calling verify() makes image unusable, but again, np.array() doesn't care. So now the code fails for all good images.

* OS: ubuntu-18.04
* Python: 3.6
* Pillow: 8.0.1

radarhere:
Resolves #5084

The following code runs without raising an error.
```python
from PIL import Image
import numpy as np
im = Image.open("Tests/images/truncated_jpeg.jpg")
np.array(im)
```

Even though an OSError is raised if `__array_interface__` is accessed directly.
```python
from PIL import Image
im = Image.open("Tests/images/truncated_jpeg.jpg")
im.__array_interface__
```
This is a [bug in numpy](https://gitter.im/numpy/numpy?at=6067b4ef3153ce63a3b6dcad).

This PR works around that bug by replacing `__array_interface__` with `__array__`.

### DIFF
--- a/Tests/test_image_array.py
+++ b/Tests/test_image_array.py
@@ -4,26 +4,32 @@ from PIL import Image
 
 from .helper import hopper
 
+numpy = pytest.importorskip("numpy", reason="NumPy not installed")
+
 im = hopper().resize((128, 100))
 
 
 def test_toarray():
     def test(mode):
-        ai = im.convert(mode).__array_interface__
-        return ai["version"], ai["shape"], ai["typestr"], len(ai["data"])
+        ai = numpy.array(im.convert(mode))
+        return ai.shape, ai.dtype.str, ai.nbytes
 
-    # assert test("1") == (3, (100, 128), '|b1', 1600))
-    assert test("L") == (3, (100, 128), "|u1", 12800)
+    # assert test("1") == ((100, 128), '|b1', 1600))
+    assert test("L") == ((100, 128), "|u1", 12800)
 
     # FIXME: wrong?
-    assert test("I") == (3, (100, 128), Image._ENDIAN + "i4", 51200)
+    assert test("I") == ((100, 128), Image._ENDIAN + "i4", 51200)
     # FIXME: wrong?
-    assert test("F") == (3, (100, 128), Image._ENDIAN + "f4", 51200)
+    assert test("F") == ((100, 128), Image._ENDIAN + "f4", 51200)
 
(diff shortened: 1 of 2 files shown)


hugovk:
Thanks for chasing up. Let's open an issue at NumPy as well?

https://gitter.im/numpy/numpy?at=6067b4ef3153ce63a3b6dcad says:

> Its a bug, but I couldn't just rip the bandaid off it because of shapely. Should do that, but a PR to do a deprecation would be welcome also of course ;)

Maybe they won't have it deprecated _and_ fixed by the next Pillow release in 3 months but let's at least give them a chance, and review this before release. I've added it to the 8.3.0 milestone.

radarhere:
Ok, raised at https://github.com/numpy/numpy/issues/18723

radarhere:
NumPy 1.21 will deprecate the error catching, raising a warning in this situation.

hugovk:
Thanks for following up!

radarhere:
https://github.com/numpy/numpy/pull/20835 has been merged into NumPy, so NumPy 1.23.should raise an error.

radarhere:
NumPy 1.23 has been released, so I've created #6394 to revert this.

PR #5384, opened 2021-04-05
Added TransferFunction and ReferenceBlackWhite to core TIFF tags

radarhere:
They have been commented out until now, as they cause "crashes". Testing, I find that they started working after https://github.com/python-pillow/Pillow/pull/5153

Also, removed a duplicate word in an error message while I'm here.


hugovk:
Are the Valgrind errors unrelated?

radarhere:
Yes. They're the same three failing tests from master - https://github.com/python-pillow/Pillow/runs/2248415838

radarhere:
The valgrind errors are gone now - fixed in master.

PR #5397, opened 2021-04-09
Valgrind fixes

wiredfool:
Fixes most of the valgrind XFAILs that we've had, including fixing the non-xfailed test_fuzzers.

### DIFF
--- a/Tests/test_file_webp_metadata.py
+++ b/Tests/test_file_webp_metadata.py
@@ -4,7 +4,8 @@ import pytest
 
 from PIL import Image
 
-from .helper import skip_unless_feature
+from .helper import skip_unless_feature, mark_if_feature_version
+
 
 pytestmark = [
     skip_unless_feature("webp"),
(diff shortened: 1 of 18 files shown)


wiredfool:
This PR works and valgrind passes on libjpeg-turbo 1.5.2 (ubuntu 18.04), but it's failing on libjpeg-turbo 2.0.3 (ubuntu-20.04).

wiredfool:
It also passes with manylinux-wheels with libjpeg 9.0, which is what we're shipping. 

Do we actually ever test with the versions of the libraries we're shipping, other than when we actually make the wheels?

Tests/test_file_webp_metadata.py
```diff
@@ -1,7 +1,5 @@
 from io import BytesIO
 
-import pytest
```
wiredfool, line 3: It's probably not worth fixing lint until we pass the: "Does it work" and "Is it a good idea" tests.

PR #5399, opened 2021-04-09
Fix PDF file format documentation

### LINKED ISSUE #5398: Docs: Make it clear that both the `append_images` and `save_all` params are needed for PDFs
The following code will not work unless the `save_all` param is set to True
```python
def imgs_to_pdf(img_arrays, path_or_byte_buf, resolution=300):
    '''
        Convert multiple images into one, multi-paged PDF
    '''
    imgs = []
    for im_array in img_arrays:
        im = Image.fromarray(im_array)
        im = im.convert('RGB')
        imgs.append(im)
    Image.save(path_or_byte_buf, format='PDF',  append_images=imgs, resolution=resolution) # Must include `save_all` param but docs do not specifically say this.
```
What to do:
* Update the docs to make it clear that both params are needed if you want the `append_images` functionality to work.

cholojuanito:
Made sure it is documented that both the `append_images` and `save_all` params must be set. Just setting `append_images` does not work.
Documentation only, [ci skip]

Fixes #5398.

 * Updated documentation about the `append_images` param for saving PDFs
 * Also made it clear that PIL.Image.Image objects must be pass in as the image list for `append_images`

### DIFF
--- a/docs/handbook/image-file-formats.rst
+++ b/docs/handbook/image-file-formats.rst
@@ -1242,8 +1242,9 @@ The :py:meth:`~PIL.Image.Image.save` method can take the following keyword argum
     .. versionadded:: 3.0.0
 
 **append_images**
-    A list of images to append as additional pages. Each of the
-    images in the list can be single or multiframe images.
+    A list of :py:obj:`PIL.Image`s to append as additional pages. Each of the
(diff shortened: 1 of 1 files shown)


radarhere:
I've created https://github.com/cholojuanito/Pillow/pull/2 to fix up the documentation syntax.

Issue #4864, opened 2020-08-13
Support DXT1, DXT3 and DXT5 DDS encoding


SerVB:
Hi! Currently DDS is a read-only format: <https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html?highlight=format#dds>. This issue asks to add support for saving DDS images.

radarhere:
I've created #5402 to add RGB saving at least.

radarhere:
I've pushed another commit to the PR to add RGBA as well, resolving this issue.

SerVB:
Thanks a lot for your efforts. Which DDS formats does writing support? Ideally, it would be great to have all: DXT1, DXT3, and DXT5

radarhere:
Ah, ok. At the moment, it only saves uncompressed data - so not DXT1, DXT3, DXT5  or DX10. That's a level more complex, and not something I'm going to attempt.

We can leave this issue open for those additional formats though if you want.

SerVB:
Let's leave it open.

jrd:
Can someone point me the documentation to the DDS (DXTn) format/spec please? I could try to implement the DXT1 (and therefore DXT3 and 5) compression write support if possible…

radarhere:
A while ago, I attempted to implement DXT1 - https://github.com/radarhere/Pillow/tree/dds. Feel free to take that and build on it if you want.

I hit a wall when I realised that it was a lossy encoding, so I wasn't going to be a single perfect solution, and using someone else's logic would probably be better. https://github.com/nothings/stb/blob/master/stb_dxt.h seemed good, but I didn't return to this.

You asked for documentation though. Here are links that I found helpful
- https://docs.microsoft.com/en-us/windows/win32/direct3d9/opaque-and-1-bit-alpha-textures
- https://docs.microsoft.com/en-us/windows/win32/directshow/working-with-16-bit-rgb
- https://docs.microsoft.com/en-us/windows/win32/direct3d10/d3d10-graphics-programming-guide-resources-block-compression#bc1

frederickjjoubert:
Did this get implemented? I would like to save a file as a `.dds` with `BC3 / DXT5` compression.

radarhere:
No, this hasn't been implemented yet.

rekterakathom:
I am also wishing for ``BC2 / DXT3`` and / or ``BC3 / DXT5``.

[...]

PR #5402, opened 2021-04-12
Add DDS saving

radarhere:
Helps #4864 by adding DDS saving. https://docs.microsoft.com/en-us/windows/win32/direct3ddds/dds-header was used as a reference.

Also, a minor cleanup change. Uses this existing line
https://github.com/python-pillow/Pillow/blob/45003b78ed4003f6bf68e3a921b78539a64c7337/src/PIL/DdsImagePlugin.py#L50
to make
https://github.com/python-pillow/Pillow/blob/45003b78ed4003f6bf68e3a921b78539a64c7337/src/PIL/DdsImagePlugin.py#L130-L131
clearer.
```python
if pfflags & DDPF_RGB:
    # Texture contains uncompressed RGB data
```

### DIFF
--- a/Tests/test_file_dds.py
+++ b/Tests/test_file_dds.py
@@ -5,7 +5,7 @@ import pytest
 
 from PIL import DdsImagePlugin, Image
 
-from .helper import assert_image_equal, assert_image_equal_tofile
+from .helper import assert_image_equal, assert_image_equal_tofile, hopper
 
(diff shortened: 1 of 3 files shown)


radarhere:
I've pushed another commit for RGBA saving.

While you would look at the spec and conclude that you should be able to specify the order of the channels arbitrarily using the bit masks, checking the results with ImageMagick I didn't find this to work. Instead, I've swapped the RGBA channels before saving so that it becomes ARGB, the example for the bit masks - https://docs.microsoft.com/en-us/windows/win32/direct3ddds/dds-pixelformat.

PR #5404, opened 2021-04-14
Changed error type to allow for incremental WebP parsing

### LINKED ISSUE #4745: RuntimeError in ImageFile.Parser.feed for WebP
### What did you do?
Use `ImageFile.Parser.feed` to get some image metadata without full image download.
Get `ImageFile.Parser.image` instance
### What actually happened?
      File "./reproduce.py", line 7, in <module>
        p.feed(new_data)
      File "/home/test/env/lib/python3.8/site-packages/PIL/ImageFile.py", line 411, in feed
        im = Image.open(fp)
      File "/home/test/env/lib/python3.8/site-packages/PIL/Image.py", line 2885, in open
        im = _open_core(fp, filename, prefix)
      File "/home/test/env/lib/python3.8/site-packages/PIL/Image.py", line 2867, in _open_core
        im = factory(fp, filename)
      File "/home/test/env/lib/python3.8/site-packages/PIL/ImageFile.py", line 107, in __init__
        self._open()
      File "/home/test/env/lib/python3.8/site-packages/PIL/WebPImagePlugin.py", line 60, in _open
        self._decoder = _webp.WebPAnimDecoder(self.fp.read())
    RuntimeError: could not create decoder object

* OS: ArchLinux. Same for Ubuntu 20.04.
* Python: 3.8.2
* Pillow: 7.1.2

This code worked for JPEG but failed for WebP:
```python
from PIL import ImageFile

p = ImageFile.Parser()
with open("./img.webp", "rb") as f:
    new_data = f.read(1024)
    while not p.image and new_data:
        p.feed(new_data)
        new_data = f.read(1024)
    print(p.image.size)

```

radarhere:
Resolves #4745

Running code from the issue on a test image,
```python
from PIL import ImageFile

p = ImageFile.Parser()
with open("Tests/images/hopper.webp", "rb") as f:
    new_data = f.read(1024)
    while not p.image and new_data:
        p.feed(new_data)
```
an error is triggered.
https://github.com/python-pillow/Pillow/blob/45003b78ed4003f6bf68e3a921b78539a64c7337/src/_webp.c#L398

This is effectively an error being raised because the the image is truncated, and most truncation errors are OSError. So if we change it to an OSError, then it is handled by `feed()`.
https://github.com/python-pillow/Pillow/blob/45003b78ed4003f6bf68e3a921b78539a64c7337/src/PIL/ImageFile.py#L413-L418

This fixes the problem.

### DIFF
--- a/Tests/test_imagefile.py
+++ b/Tests/test_imagefile.py
@@ -82,6 +82,19 @@ class TestImageFile:
             p.feed(data)
             assert (48, 48) == p.image.size
 
+    @skip_unless_feature("webp")
+    @skip_unless_feature("webp_anim")
+    def test_incremental_webp(self):
+        with ImageFile.Parser() as p:
+            with open("Tests/images/hopper.webp", "rb") as f:
+                p.feed(f.read(1024))
+
+                # Check that insufficient data was given in the first feed
+                assert not p.image
+
+                p.feed(f.read())
+            assert (128, 128) == p.image.size
+
     @skip_unless_feature("zlib")
     def test_safeblock(self):
         im1 = hopper()
--- a/src/_webp.c
+++ b/src/_webp.c
@@ -395,7 +395,7 @@ _anim_decoder_new(PyObject *self, PyObject *args) {
         }
         PyObject_Del(decp);
     }
-    PyErr_SetString(PyExc_RuntimeError, "could not create decoder object");
+    PyErr_SetString(PyExc_OSError, "could not create decoder object");
     return NULL;
 }
 


wiredfool:
We should really discourage using this interface. While it's faster in some cases, when it doesn't work, it doesn't work in O(n^2).

radarhere:
That inefficiency should be apparent to the user though - our API doesn't do anything to obscure that.

Are you saying that we shouldn't improve `Parser` any further, closing the issue as a 'wontfix'?

wiredfool:
I'm not sure.  

I'm pretty sure for most of the more recent formats that use external libraries, like jpeg2k and webp, probably the avif are going to to require the whole image eventually. It's also not uncommon that tiff style images have IFDs that are at the end of the file. 

In that case, you're calling the parser with the first 1k, then 2k, then 3k, then 4k, and so on. 

Metadata only, maybe ok.

hugovk:
> We should really discourage using this interface. While it's faster in some cases, when it doesn't work, it doesn't work in O(n^2).

Shall we merge this improvement and then deprecate the API?

It can then be made private and/or removed in Pillow 10.0.0 (2023-01-02).

hugovk (approved):

PR #5408, opened 2021-04-15
Corrected JPEG APP14 transform value

radarhere:
Helps #3953

In the APP14 marker, https://github.com/LuaDist/libjpeg/blob/6c0fcb8ddee365e7abc4d332662b06900612e923/jcmarker.c#L435-L443
```
/* Emit an Adobe APP14 marker */
{
  /*
   * Length of APP14 block	(2 bytes)
   * Block ID			(5 bytes - ASCII "Adobe")
   * Version Number		(2 bytes - currently 100)
   * Flags0			(2 bytes - currently 0)
   * Flags1			(2 bytes - currently 0)
   * Color transform		(1 byte)
```
There are 5 bytes of "Adobe", 2 bytes of Version, 4 of flags and then the 'transform' value. This makes 'transform' the 12th value, at index 11 - not index 1, as we have had since PIL.

Using exiftool to confirm the results of my new test,
```
$ exiftool -v Tests/images/pil_sample_rgb.jpg | grep ColorTransform 
  | ColorTransform = 1
$ exiftool -v Tests/images/pil_sample_cmyk.jpg | grep ColorTransform
  | ColorTransform = 2
```

Also changes an exception type to be more specific - `IndexError` is what would be raised if `s` were too short.


hugovk:
Looks good, is the `except IndexError:` block easy to test as well?

radarhere:
Ok, I've added a commit.

hugovk:
Thanks!

PR #5411, opened 2021-04-16
For special image modes, revert default resize resampling to NEAREST

### LINKED ISSUE #4402: Bicubic resize of an uint16 image
### What did you do?
I moved an existing Python app from a machine with Pillow 6.2.1 to a new machine which had Pillow 7.0.0 installed.  The app failed to resize a PIL.Image.

I expected it to work as before

### What actually happened?
resize threw an exception ValueError: image has wrong mode

* OS: ubuntu 18.04
* Python: 3.6.9
* Pillow: 7.0.0

```python
import numpy as np
from PIL import Image

imgdata = np.array( [[484, 451, 481, 465, 488, 442],
 [823, 897, 971, 920, 855, 908],
 [758, 919, 854, 920, 850, 854],
 [840, 873, 950, 893, 872, 881],
 [937, 940, 835, 879, 895, 774],
 [841, 914, 867, 886, 922, 952]]).astype(np.uint16)

pimg = Image.fromarray(imgdata)
scale_factor = 0.5
size = tuple((np.array(pimg.size) * scale_factor).astype(int))
newim = pimg.resize(size)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-1-5c0c4827f11b> in <module>
     12 scale_factor = 0.5
     13 size = tuple((np.array(pimg.size) * scale_factor).astype(int))
---> 14 newim = pimg.resize(size)

e:\program files\python37\lib\site-packages\PIL\Image.py in resize(self, size, resample, box, reducing_gap)
   1871                 )
   1872 
-> 1873         return self._new(self.im.resize(size, resample, box))
   1874 
   1875     def reduce(self, factor, box=None):

ValueError: image has wrong mode


```

radarhere:
Helps #4402

#4255 changed the default `resample` argument for `resize()` to BICUBIC, with the goal of being more user-friendly. However, it turned out to not be more user-friendly for IMAGING_TYPE_SPECIAL modes, where we [don't currently support resampling](https://github.com/python-pillow/Pillow/blob/9b6fe1b039e4a70d59a4f0115df813090bb230f6/src/libImaging/Resample.c#L569-L570), except [for NEAREST](https://github.com/python-pillow/Pillow/blob/9b6fe1b039e4a70d59a4f0115df813090bb230f6/src/_imaging.c#L1808-L1822) - leading to the "image has wrong mode" in the issue.

This PR switches back to NEAREST if the image is IMAGING_TYPE_SPECIAL, so that the default `resample` value once again works for those modes.

Looking at the [code](https://github.com/python-pillow/Pillow/blob/9b6fe1b039e4a70d59a4f0115df813090bb230f6/src/libImaging/Storage.c#L123-L161), IMAGING_TYPE_SPECIAL applies to mode with a custom number of bits: a semicolon in their mode.

### DIFF
--- a/Tests/test_image_resize.py
+++ b/Tests/test_image_resize.py
@@ -250,3 +250,7 @@ class TestImageResize:
         for mode in "1", "P":
             im = hopper(mode)
             assert im.resize((20, 20), Image.NEAREST) == im.resize((20, 20))
+
+        for mode in "I;16", "I;16L", "I;16B", "BGR;15", "BGR;16":
+            im = hopper(mode)
+            assert im.resize((20, 20), Image.NEAREST) == im.resize((20, 20))
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -1849,7 +1849,7 @@ class Image:
             min(self.size[1], math.ceil(box[3] + support_y)),
         )
 
-    def resize(self, size, resample=BICUBIC, box=None, reducing_gap=None):
+    def resize(self, size, resample=None, box=None, reducing_gap=None):
         """
         Returns a resized copy of this image.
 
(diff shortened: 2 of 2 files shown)


homm (approved):

axu2:
Just to add to this conversation, I also think NEAREST is the best default for upscaling low resolution text while keeping sharpness. https://en.wikipedia.org/wiki/Comparison_gallery_of_image_scaling_algorithms

PR #5412, opened 2021-04-17
Convert to float for comparison with float in IFDRational __eq__

radarhere:
This one is up for debate.

Consider this behaviour.
```pycon
>>> 7 / 5
1.4
>>> IFDRational(7, 5) == 1.4
False
```
I feel like this is unexpected. Why is this happening? Well,
```pycon
>>> IFDRational(7, 5)._val
Fraction(7, 5)
>>> from fractions import Fraction
>>> Fraction(7, 5) == 1.4
False
```
So despite the fact that
```pycon
>>> float(Fraction(7, 5)) == 1.4
True
```
`Fraction(7, 5)` is not equal to 1.4.

This isn't the case for all values.

```pycon
>>> Fraction(5, 4) == 1.25
True
```

This PR 'fixes' this behaviour in `IFDRational` by casting `Fraction` to `float` when comparing in `IFDRational.__eq__()`.

As I said, open for debate.

### DIFF
--- a/Tests/test_file_tiff.py
+++ b/Tests/test_file_tiff.py
@@ -146,25 +146,24 @@ class TestFileTiff:
             "Tests/images/hopper_float_dpi_" + str(resolutionUnit) + ".tif"
         ) as im:
             assert im.tag_v2.get(RESOLUTION_UNIT) == resolutionUnit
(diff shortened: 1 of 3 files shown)


kmilos:
LGTM

kmilos (approved):

PR #5417, opened 2021-04-19
Added ImageOps contain()

### LINKED ISSUE #5415: Function similar to pad but returns resized image without padding
I'd sure like a function that looks like ImageOps.pad
https://pillow.readthedocs.io/en/stable/_modules/PIL/ImageOps.html#pad

but just returns the image after resizing without adding the padding.

What would be a good name for such a function? Where would it best be placed?

ImageOps.fit is already taken.

EDIT: Looking at #3364 `contain` seems like a decent candidate. Seems a refactor would be required. @radarhere thoughts?

radarhere:
Resolves #5415

Adds a `contain()` method to ImageOps. I would describe it as the inverse of `fit()`.

Running this code for each method in turn for comparison -

```python
ImageOps.fit(im, (200, 300))
```

|       | Fit                                                                                                        | Pad                                                                                                        | Contain                                                                                                        |
| ----- | ---------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------- |
| Size  |200 x 300                                                                                                   | 200 x 300                                                                                                  | 200 x 200                                                                                                      |
| Image |fit|pad|contain|

### DIFF
--- a/Tests/test_imageops.py
+++ b/Tests/test_imageops.py
@@ -37,6 +37,9 @@ def test_sanity():
     ImageOps.pad(hopper("L"), (128, 128))
     ImageOps.pad(hopper("RGB"), (128, 128))
 
+    ImageOps.contain(hopper("L"), (128, 128))
+    ImageOps.contain(hopper("RGB"), (128, 128))
+
     ImageOps.crop(hopper("L"), 1)
     ImageOps.crop(hopper("RGB"), 1)
 
@@ -97,6 +100,21 @@ def test_fit_same_ratio():
     with Image.new("RGB", (1000, 755)) as im:
         new_im = ImageOps.fit(im, (1000, 755))
         assert new_im.size == (1000, 755)
+
+
+def test_contain():
+    # Same ratio
+    im = hopper()
+    new_size = (im.width * 2, im.height * 2)
+    new_im = ImageOps.contain(im, new_size)
+    assert new_im.size == new_size
+
+    for new_size in [
+        (im.width * 4, im.height * 2),
+        (im.width * 2, im.height * 4),
+    ]:
+        new_im = ImageOps.contain(im, new_size)
+        assert new_im.size == (im.width * 2, im.height * 2)
 
 
 def test_pad():
--- a/Tests/test_imageops.py
+++ b/Tests/test_imageops.py
@@ -102,19 +102,11 @@ def test_fit_same_ratio():
         assert new_im.size == (1000, 755)
 
 
-def test_contain():
-    # Same ratio
+@pytest.mark.parametrize("new_size", ((256, 256), (512, 256), (256, 512)))
+def test_contain(new_size):
     im = hopper()
-    new_size = (im.width * 2, im.height * 2)
     new_im = ImageOps.contain(im, new_size)
-    assert new_im.size == new_size
-
-    for new_size in [
-        (im.width * 4, im.height * 2),
-        (im.width * 2, im.height * 4),
-    ]:
-        new_im = ImageOps.contain(im, new_size)
-        assert new_im.size == (im.width * 2, im.height * 2)
+    assert new_im.size == (256, 256)
 
 
 def test_pad():
--- a/docs/releasenotes/8.3.0.rst
+++ b/docs/releasenotes/8.3.0.rst
@@ -1,0 +1,49 @@
+8.3.0
+-----
+
+Deprecations
+============
+
+TODO
+^^^^
+
+TODO
+
+API Changes
+===========
+
+TODO
+^^^^
+
+TODO
+
+API Additions
+=============
+
+ImageOps.contain
+^^^^^^^^^^^^^^^^
+
+Returns a resized version of the image, set to the maximum width and height within
+``size``, while maintaining the original aspect ratio.
+
+To compare it to other ImageOps methods:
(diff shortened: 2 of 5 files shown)


radarhere:
3.10-dev failures seem to be due to https://github.com/pytest-dev/pytest/issues/8539, which has been resolved, but is not available in a pytest release yet.

axu2:
Looks great, exactly what I was looking for.

axu2:
OK, I actually copy pasted the code and the contain method seems to change the aspect ratio.

axu2:
Moreover, I think the same problem exists with the pad function as well.

I am passing images that are skinnier than the requested `size`.

axu2:
OK, I actually tried to run the pad function myself and I got some weird issues.

`im1` is a file thats 2037 wide 3056 tall. When I run this I get:

```
>>> ImageOps.pad(im1, (824,1200)).show() 
  File "...\venv\lib\site-packages\PIL\ImageOps.py", line 262, in pad
    image = image.resize((new_width, size[1]), resample=method)
  File "...\venv\lib\site-packages\PIL\Image.py", line 1888, in resize
    return self._new(self.im.resize(size, resample, box))
ValueError: height and width must be > 0
```

axu2:
Sorry, I left a bunch of comments that I quickly deleted, since I had made an error.

src/PIL/ImageOps.py
```diff
@@ -236,6 +236,34 @@ def colorize(image, black, white, mid=None, blackpoint=0, whitepoint=255, midpoi
     return _lut(image, red + green + blue)
 
 
+def contain(image, size, method=Image.BICUBIC):
+    """
+    Returns a sized version of the image, expanded to fill the requested aspect ratio
+    and size.
+
+    :param image: The image to size and crop.
```
hugovk, line 244:
```suggestion
    :param image: The image to resize and crop.
```

src/PIL/ImageOps.py
```diff
@@ -236,6 +236,34 @@ def colorize(image, black, white, mid=None, blackpoint=0, whitepoint=255, midpoi
...
+    :param size: The requested output size in pixels, given as a
+                 (width, height) tuple.
+    :param method: What resampling method to use. Default is
```
hugovk, line 247:
```suggestion
    :param method: Resampling method to use. Default is
```

src/PIL/ImageOps.py
```diff
@@ -236,6 +236,34 @@ def colorize(image, black, white, mid=None, blackpoint=0, whitepoint=255, midpoi
     return _lut(image, red + green + blue)
 
 
+def contain(image, size, method=Image.BICUBIC):
+    """
+    Returns a sized version of the image, expanded to fill the requested aspect ratio
+    and size.
```
hugovk, line 242: Maybe something like this is clearer?
```suggestion
    Returns a resized version of the image, expanded to fill the requested size by cropping, and with no padding.
```

Tests/test_imageops.py
```diff
@@ -99,6 +102,21 @@ def test_fit_same_ratio():
         assert new_im.size == (1000, 755)
 
 
+def test_contain():
```
hugovk, line 105: I think these three cases can be refactored using `pytest.mark.parametrize`?

hugovk (commented):
Let's also include this in the docs (`autofunction` in `docs/reference/ImageOps.rst`?) and release notes.

Tests/test_imageops.py
radarhere: I've pushed a commit for this

src/PIL/ImageOps.py
radarhere: There isn't any cropping in this method.
I've pushed a commit with
> Returns a resized version of the image, set to the maximum width and height within the requested size, while maintaining the original aspect ratio.

Feel free to suggest further changes.

hugovk: Sounds good, thanks! (I forgot hopper.png is square when looking at the examples.)

docs/releasenotes/8.3.0.rst
```diff
@@ -0,0 +1,64 @@
...
+
+ImageOps.contain
+^^^^^^^^^^^^^^^^
+
+Returns a resized version of the image, set to the maximum width and height within
+``size``, while maintaining the original aspect ratio.
+
+To compare it to other ImageOps methods:
+- :py:meth:`~PIL.ImageOps.fit` expands an image until is fills ``size``, cropping the
+parts of the image that do not fit.
```
hugovk, line 46:
```suggestion
To compare it to other ImageOps methods:

- :py:meth:`~PIL.ImageOps.fit` expands an image until is fills ``size``, cropping the
parts of the image that do not fit.
```

Is a newline needed to activate bullet points in rst?

image

https://docutils.sourceforge.io/docs/user/rst/quickref.html#bullet-lists

radarhere: Yep, I guess newline, but also some indentation. Should be right now.

hugovk:
Thank you!

axu2:
🥳

axu2:
Oh one last comment, just realized this function is similar to Image.thumbnail except it will upscale the image, if needed.

Also, for upscaling low resolution text, PIL.NEAREST might be a better default choice.
https://en.wikipedia.org/wiki/Comparison_gallery_of_image_scaling_algorithms

PR #5424, opened 2021-04-20
Add Raspberry Pi OS to platform support

thijstriemstra:


radarhere:
You're saying that you installed Pillow 8.2.0 in Python 2.7 and were able to successfully run the test suite? That doesn't sound right.

thijstriemstra:
>  That doesn't sound right.

True. I've run older Pillows on 2.7 on raspberry pi os but the table structure isn't suitable for that. Should I remove 2.7 or lower pillow version, or keep as is?

radarhere:
Take a look at the top of the table - macOS 10.14 and later. Multiple Pillow versions are listed for a single operating system there, by splitting rows.

thijstriemstra:
Thanks for the feedback @radarhere, I updated the pr.

docs/installation.rst
```diff
@@ -532,6 +532,10 @@ These platforms have been reported to work at the versions mentioned.
 +----------------------------------+------------------------------+--------------------------------+-----------------------+
 | Raspbian Stretch                 | 2.7, 3.5                     | 4.0.0                          |arm                    |
 +----------------------------------+------------------------------+--------------------------------+-----------------------+
+| Raspberry Pi OS                  | 2.7                          | 6.2.2                          |arm                    |
+|                                  +------------------------------+--------------------------------+-----------------------+
+|                                  | 3.6, 3.7, 3.8, 3.9           | 8.2.0                          |arm                    |
```
radarhere, line 537:
```suggestion
| Raspberry Pi OS                  | 3.6, 3.7, 3.8, 3.9           | 8.2.0                          |arm                    |
|                                  +------------------------------+--------------------------------+                       |
|                                  | 2.7                          | 6.2.2                          |                       |
```
Suggestion to swap the rows so that the latest version of Pillow is at the top, and to merge the two 'arm' cells

radarhere:
I think it would also be good to include what version of Raspberry Pi OS you are using. Buster?

thijstriemstra:
Yea, it's buster but it'll always be referenced as raspberry pi os from now on afaik.

radarhere (approved):

PR #5430, opened 2021-04-22
Switched to saving 1-bit PDFs with DCTDecode

### LINKED ISSUE #1775: PDFs and 1 bit pixels
I posted this question [on Stack Overflow](http://stackoverflow.com/questions/36121914/pillow-pdf-and-1-bit-pixels) before realizing that perhaps I should post it here.

I'm trying to create a PDF of a complicated black-and-white rectangular grid, nominally 7 inches wide by 10 inches tall.  I'm using Pillow to do the conversion.  If I specify a PNG, everything works fine (that is, the aspect ratio is 7 by 10).  If I specify a PDF, the output is 200 inches wide and 0.04 inches tall.  What's going on?

Incidentally, if I set `mode='RGB'`, then the output PDF dimensions are correct, but Pillow performs a JPEG encoding of the image. JPEG converts my sharp black and white grid into smoother shades of gray (which is anathema for my application).

```python
import PIL
from PIL import Image

# Let's consider a 7" x 10" rectangle of pixels
inches_wide=7
inches_tall=10
dpi=200
num_pixel_rows=inches_tall*dpi
num_pixel_cols=inches_wide*dpi

# Make an image
mode='1'
img = Image.new( mode, (num_pixel_cols,num_pixel_rows))
img.info['dpi'] = (dpi,dpi)

# Saving it as PNG, we get a black rectangle, as expected.
img.save('big_rectangle.png','png',resolution=dpi)
# Saving it as a PDF, we get a 200" x 0.04" all-white line?!
img.save('big_rectangle.pdf','pdf',resolution=dpi)

# Versioning info:
print 'PILLOW',PIL.PILLOW_VERSION
print 'PIL',PIL.VERSION
#  My output:
#  PILLOW 3.1.1
#  PIL 1.1.7
```

radarhere:
Resolves #1775 - the remaining part of that issue is a request for compression when saving 1-bit PDFs.

The issue reports that
```python
from PIL import Image
img = Image.new("1", (1400, 2000))
img.save("out.pdf")
```
generates a 5mb file.

Changing the filter to "DCTDecode", it becomes 34kb instead.

I've modified a test to ensure that saving `hopper("1")` as a PDF, that currently creates a 34kb file, stays below 15kb.

### DIFF
--- a/src/PIL/PdfImagePlugin.py
+++ b/src/PIL/PdfImagePlugin.py
@@ -124,7 +124,7 @@ def _save(im, fp, filename, save_all=False):
             decode = None
 
             if im.mode == "1":
-                filter = "ASCIIHexDecode"
+                filter = "DCTDecode"
                 colorspace = PdfParser.PdfName("DeviceGray")
                 procset = "ImageB"  # grayscale
                 bits = 1
@@ -161,12 +161,6 @@ def _save(im, fp, filename, save_all=False):
             op = io.BytesIO()
 
             if filter == "ASCIIHexDecode":
-                if bits == 1:
-                    # FIXME: the hex encoder doesn't support packed 1-bit
-                    # images; do things the hard way...
-                    data = im.tobytes("raw", "1")
-                    im = Image.new("L", im.size)
-                    im.putdata(data)
                 ImageFile._save(im, op, [("hex", (0, 0) + im.size, 0, im.mode)])
             elif filter == "DCTDecode":
                 Image.SAVE["JPEG"](im, op, filename)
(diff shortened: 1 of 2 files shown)


src/PIL/PdfImagePlugin.py
```diff
@@ -161,12 +161,6 @@ def _save(im, fp, filename, save_all=False):
             op = io.BytesIO()
 
             if filter == "ASCIIHexDecode":
-                if bits == 1:
-                    # FIXME: the hex encoder doesn't support packed 1-bit
-                    # images; do things the hard way...
-                    data = im.tobytes("raw", "1")
-                    im = Image.new("L", im.size)
-                    im.putdata(data)
```
radarhere, line 169: This code is now unused, so I have removed it.

PR #5437, opened 2021-04-25
Replace sys.stdout with sys.stdout.buffer when saving

### LINKED ISSUE #5436: Can't save PNG to sys.stdout
### What did you do? 
Just tried to run simple code (ImageDraw)

### What actually happened? 
Exception looks very basic:
```  File "/data/user/0/ru.iiec.pydroid3/files/aarch64-linux-android/lib/python3.8/site-packages/PIL/Image.py", line 2164, in save
  File "/data/user/0/ru.iiec.pydroid3/files/aarch64-linux-android/lib/python3.8/site-packages/PIL/PngImagePlugin.py", line 1225, in _save
    fp.write(_MAGIC)
TypeError: write() argument must be str, not bytes
```

* OS: Android app, "pydroid3" (I don't have a computer now to check it)
* Python: 3.8.3
* Pillow: 8.1.2

```python
import sys

# list of dots, in the following format: [x, y, x, y, x, y,...]
first = (146, 399, 163, 403, 170, 393, 169, 391, 166, 386, 170, 381, 170) #.... long tuple

with Image.open("draw image.jpg") as img:

    draw = ImageDraw.Draw(img)
    for i in range(len(first)//2 - 1):
    	draw.line(first[2*i : 2*i + 4])

    # write to stdout
    img.save(sys.stdout, "PNG")
```
Sorry if I didn't understand the concept... This is my first issue ever...

radarhere:
Resolves #5436

https://pillow.readthedocs.io/en/stable/reference/ImageDraw.html#example-draw-a-gray-cross-over-an-image
```python

with Image.open("hopper.jpg") as im:

    draw = ImageDraw.Draw(im)
    draw.line((0, 0) + im.size, fill=128)
    draw.line((0, im.size[1], im.size[0], 0), fill=128)

    # write to stdout
    im.save(sys.stdout, "PNG")
```

However, running this example code gives
```
  File "<stdin>", line 6, in <module>
  File "PIL/Image.py", line 2172, in save
  File "PIL/PngImagePlugin.py", line 1223, in _save
    fp.write(_MAGIC)
TypeError: write() argument must be str, not bytes
```

This PR suggests fixing this by changing `im.save` so that if it is passed `sys.stdout` (`sys.stdout.mode` is 'w'), then it is replaced with `sys.stdout.buffer` (`sys.stdout.buffer.mode` is 'wb').

It also changes the `PSDraw()` default from `sys.stdout` to `sys.stdout.buffer`, and simplifies EPS saving by abandoning the idea of wrapping `fp` just so that it can write strings.

### DIFF
--- a/Tests/test_file_png.py
+++ b/Tests/test_file_png.py
@@ -713,13 +713,20 @@ class TestFilePng:
             with pytest.raises(EOFError):
                 im.seek(1)
 
-    def test_save_stdout(self):
+    @pytest.mark.parametrize("buffer", (True, False))
+    def test_save_stdout(self, buffer):
         old_stdout = sys.stdout.buffer
 
-        class MyStdOut:
-            buffer = BytesIO()
+        if buffer:
 
-        sys.stdout = mystdout = MyStdOut()
+            class MyStdOut:
+                buffer = BytesIO()
+
+            mystdout = MyStdOut()
+        else:
+            mystdout = BytesIO()
+
+        sys.stdout = mystdout
 
         with Image.open(TEST_PNG_FILE) as im:
             im.save(sys.stdout, "PNG")
@@ -727,7 +734,9 @@ class TestFilePng:
         # Reset stdout
         sys.stdout = old_stdout
 
-        reloaded = Image.open(mystdout.buffer)
+        if buffer:
+            mystdout = mystdout.buffer
+        reloaded = Image.open(mystdout)
         assert_image_equal_tofile(reloaded, TEST_PNG_FILE)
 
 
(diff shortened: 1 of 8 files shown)


hugovk:
Looks good, one question:

The [docs for `sys.stdout`](https://docs.python.org/3/library/sys.html#sys.stdout) includes this:

> **Note:** To write or read binary data from/to the standard streams, use the underlying binary `buffer` object. For example, to write bytes to `stdout`, use `sys.stdout.buffer.write(b'abc')`.

OK, good. It continues:

> However, if you are writing a library (and do not control in which context its code will be executed), be aware that the standard streams may be replaced with file-like objects like `io.StringIO` which do not support the `buffer` attribute.

Is this something we need to accomodate?

radarhere:
Nice catch. Sure, I've added a commit to allow for that.

hugovk:
Thank you!

PR #5447, opened 2021-04-27
Fixed Python 3.10 at alpha.7

radarhere:
Fix the failures on Python 3.10 in master - https://github.com/python-pillow/Pillow/runs/2431607308 - by changing the Python 3.10 version to be static at alpha.7

The failures do not come from Pillow, but from pytest - https://github.com/pytest-dev/pytest/issues/8539


hugovk:
I asked about release ETA at https://github.com/pytest-dev/pytest/issues/8539#issuecomment-827505576, no reply yet so let's go with pinning. Thanks!

PR #5460, opened 2021-05-02
Remove spikes when drawing thin pieslices

### LINKED ISSUE #5432: `pieslice` draws very thin sectors incorrectly
### What did you do?
I drew a circle sector with a very small central angle using `pieslice`.

I expected to get a shape approximating a line from the edge of the (imagined) disk to its center.

### What actually happened?
The (approximate) line extends past the disk's center.

The effect is more pronounced for smaller central angles and less pronounced for arcs starting/ending near 0° and 180°.

The following images show 1° arcs drawn using `pieslice` and a small red circle marking the center of the corresponding disk. The script creating them is given at the end.

These results are as expected (base angles 0°, 180°):
test0 test180

The following demonstrate the problem (base angles 15°, 45°, 90°, 270°):
test15 test45 test90 test270

* OS: archlinux
* Python: 3.9.3
* Pillow: 8.1.2

```python
# This code creates several files, only a few selected ones are shown above.

r = 100 # radius in px
angle = 1 # in degrees
stepsize = 15

for angle_offset in range(0, 360+stepsize, stepsize):
    img = Image.new(mode="RGB", size=(2*r, 2*r), color=(255, 255, 255))

    # draw a sector
    draw = ImageDraw.Draw(img)
    draw.pieslice((0, 0, 2*r, 2*r), start=angle_offset, end=angle_offset+angle, fill=(0, 0, 0))

    # draw a small circle to mark the center
    draw.ellipse((r-1, r-1, r+1, r+1), fill=(255, 0, 0))

    img.save(f"test{angle_offset}.png")
```

xtsm:
Fixes #5432 .

### DIFF
--- a/src/libImaging/Draw.c
+++ b/src/libImaging/Draw.c
@@ -1347,6 +1347,22 @@ pie_init(clip_ellipse_state *s, int32_t a, int32_t b, int32_t w, float al, float
     s->root->l = lc;
     s->root->r = rc;
     s->root->type = ar - al < 180 ? CT_AND : CT_OR;
+
+    // add one more semiplane to avoid spikes
+    if (ar - al < 90) {
+        clip_node *old_root = s->root;
+        clip_node *spike_clipper = s->nodes + s->node_count++;
+        s->root = s->nodes + s->node_count++;
+        s->root->l = old_root;
+        s->root->r = spike_clipper;
+        s->root->type = CT_AND;
+
+        spike_clipper->l = spike_clipper->r = NULL;
+        spike_clipper->type = CT_CLIP;
+        spike_clipper->a = (xl + xr) / 2.0;
+        spike_clipper->b = (yl + yr) / 2.0;
+        spike_clipper->c = 0;
+    }
 }
 
 void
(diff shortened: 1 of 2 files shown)


src/libImaging/Draw.c
```diff
@@ -1347,6 +1347,22 @@ pie_init(clip_ellipse_state *s, int32_t a, int32_t b, int32_t w, float al, float
...
+        clip_node *spike_clipper = s->nodes + s->node_count++;
+        s->root = s->nodes + s->node_count++;
+        s->root->l = old_root;
+        s->root->r = spike_clipper;
+        s->root->type = CT_AND;
+
+        spike_clipper->l = spike_clipper->r = NULL;
+        spike_clipper->type = CT_CLIP;
+        spike_clipper->a = (xl + xr) / 2.0;
+        spike_clipper->b = (yl + yr) / 2.0;
```
radarhere, line 1363: Here's my explanation of what is happening here.
This is taking the average of the two points on the outside of the circle. Going through the center of the circle as well, that gives us a line down the middle of the pieslice.
Then by swapping x and y (normally y would be with a), and multiplying one of them by -1, it gives us a perpendicular line. Because the line also goes through the center of the circle, that cuts off the spike.

xtsm: yeah, that's one way to put it
line equation Ax+By+C=0 can be also written as rn + C = 0, r(x, y), n(A, B) - i.e. n(A, B) is a normal vector for line

PR #5462, opened 2021-05-03
Fix build with libraqm < 0.7.0

adrian-the-git:
gcc's cpp doesn't short-circuit when RAQM_VERSION_ATLEAST is undefined
so building fails on systems with old libraqm (e.g. Debian 10 for
raspberry pi where wheels w/libraqm support may not be available).
RAQM_VERSION_ATLEAST was introduced with 0.7.0, so the test is
effectively equivalent.

### DIFF
--- a/src/_imagingft.c
+++ b/src/_imagingft.c
@@ -277,7 +277,7 @@ text_layout_raqm(
             direction = RAQM_DIRECTION_LTR;
         } else if (strcmp(dir, "ttb") == 0) {
             direction = RAQM_DIRECTION_TTB;
-#if !defined(RAQM_VERSION_ATLEAST) || !RAQM_VERSION_ATLEAST(0, 7, 0)
+#if !defined(RAQM_VERSION_ATLEAST)
             PyErr_SetString(
                 PyExc_ValueError,
                 "libraqm 0.7 or greater required for 'ttb' direction");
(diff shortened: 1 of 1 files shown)


radarhere (approved):

radarhere:
Yes, I see that the Debian package does have an older version - https://packages.debian.org/buster/libraqm-dev

nulano (approved):
I only tested MSVC which seems to give a warning I didn't notice (compiling Pillow with MSVC generates a lot of other warnings).

src/_imagingft.c
```diff
@@ -277,7 +277,7 @@ text_layout_raqm(
             direction = RAQM_DIRECTION_LTR;
         } else if (strcmp(dir, "ttb") == 0) {
             direction = RAQM_DIRECTION_TTB;
-#if !defined(RAQM_VERSION_ATLEAST) || !RAQM_VERSION_ATLEAST(0, 7, 0)
+#if !defined(RAQM_VERSION_ATLEAST)
```
nulano, line 280: Does it make sense to add a clarifying comment?
```suggestion
#if !defined(RAQM_VERSION_ATLEAST)
            /* RAQM_VERSION_ATLEAST was added in Raqm 0.7.0 */
```

PR #5446, opened 2021-04-27
Removed TIFF DPI rounding

radarhere:
Helps #4129 by removing TIFF DPI rounding when loading and saving.

Because the dpi values are no longer converted to int, they are now IFDRational - so in order to allow existing pickling tests to continue passing, I have added `__getstate__` and `__setstate__` to IFDRational to allow it to be pickled.

Issue #4129, opened 2019-10-11
Since 6.0, setting float resolution with `dpi` argument converts to an integer


Bolayniuss:
Since `Pillow==6.0`, when doing this:
```python
pillow_image.save(
    file_path,
    dpi=(812.8, 812.8)
)
```

the resolution of the of the saved file is `(813, 813)`, even if the dpi value is a `PIL.TiffImagePlugin.IFDRational ` object. (I use Photoshop CS6 to check the resolution)

before `6.0` it was `(812.8, 812.8)`.

I've found that using the `resolution` argument solve this issue but then the `img.info` is a bit weird as it gives two different values for the resolution and dpi 

```python
{'dpi': (813, 813), 'resolution': (812.8000797477443, 812.8000797477443)}
```

radarhere:
If you look at PR #3709, you'll see that rounded was deliberately added, to be consistent about the way that DPI was treated across image formats.

Bolayniuss:
Alright, I see that it's done an purpose. But I still don't understand why? Why DPI should be magically rounded, especially when a float was previously expected. Why DPI should be constraint to integer, a float resolution is not just an abstract concept, it's something that can exist.

I am also confused be the deferent behavior when setting the `resolution` instead of `dpi`.

Bolayniuss:
Is there any news about this issue. I am currently dealing with this issue by keeping Pillow under v6 but it will not be manageable any longer. First because of new version of python (since p3.8 is not "officially" supported by Pillow < 6.2.1), Secondly because a lot of our dependencies expect a Pillow version >= 6, it was not a huge deal until now, but PIP change it's way to deal with dependencies and we have conflict now.

Thanks to keep me updated.

awbacker:
Curious if this will ever be fixed, or adjusted to maybe support setting of a `Decimal` at least, to alert the code that it shouldn't round.  The HP Indigo is 812.8 DPI at least in halftone, and that is an extremely common printer.  

I've hacked this floating point resolution into golang and node packages (which are much more primitive imaging libraries), but this one is beyond me for now.

[From a forum thread I found](https://printplanet.com/threads/resolution.19975/):
> Most drum scanners were metric, so instead of scanning at a true 150 lpi they digitized files at 60 LCM or 120 PPCM
> NOTE: Halftone resolution was commonly half of the digital image resolution or a digital image had 2 pixels per halftone line.
> A 60 LCM halftone was actually digitized at 120 PPCM
[...]

PR #5475, opened 2021-05-07
Fix: Allow TIFF plugin to write DPI values with floating point

romaroman:
Fixes inability to write TIFF file with float DPI values. It's crucial to set these settings for industrial printing purposes because they often work on specific DPIs.

### DIFF
--- a/src/PIL/TiffImagePlugin.py
+++ b/src/PIL/TiffImagePlugin.py
@@ -1519,8 +1519,8 @@ def _save(im, fp, filename):
     dpi = im.encoderinfo.get("dpi")
     if dpi:
         ifd[RESOLUTION_UNIT] = 2
-        ifd[X_RESOLUTION] = int(dpi[0] + 0.5)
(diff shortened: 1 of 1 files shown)


radarhere:
Hi. If you had created this pull request two weeks ago, you would have started the ball rolling to help #4129 with this change - however, you've been beaten to the punch by #5446.

If that PR is merged in the next two months, then it will be part of Pillow 8.3.0 when it is released on July 1.

romaroman:
> Hi. If you had created this pull request two weeks ago, you would have started the ball rolling to help #4129 with this change, and I'd look at this and suggest that we should remove the TIFF rounding on loading as well - however, PR #5446 already removes TIFF DPI rounding on both loading and saving.

Thanks for info, I’ll close this PR then and will just use modified package for the moment

radarhere:
#5446 has been merged. So if you would like this change immediately, you can compile from Pillow master. Otherwise, this change will be part of Pillow 8.3.0 when it is due to be released on July 1

PR #5488, opened 2021-05-11
Use METH_NOARGS when no arguments are required

### LINKED ISSUE #5487: Calling conventions of some Python foreign functions mismatch with their C implementations
For example, Python foreign function `ImagingCore.isblock()` takes no arguments, as in its C implementation, arguments passed from the Python side will not be used:

https://github.com/python-pillow/Pillow/blob/d393cfb65ae7d99b4b8e1cb8c0ced4cf79d12835/src/_imaging.c#L2097-L2100

parameter `args` is not used in the function body.

However, its calling convention given by Python/C API when declaring the foreign function is `METH_VARARGS (0x0001)`, which should be `METH_NOARGS` instead in this case:

https://github.com/python-pillow/Pillow/blob/d393cfb65ae7d99b4b8e1cb8c0ced4cf79d12835/src/_imaging.c#L3460

This mismatch makes parameter of any type can be passed to this foreign function designed to be parameter free without getting a TypeError:

```Python
from PIL import Image
im = Image.open("./test.png")
ic = im.getdata()
ic.isblock('str', 1) # passing anything is legal because of the mismatch
```

The same mismatch can be found for foreign functions `ImagingCore.split()`, `ImagingCore.getbbox()`, `ImagingCore.chop_invert()`, `ImagingCore.getextrema()`, `ImagingCore.getpalettemode()`, `ImagingCore.getprojection()`.

We also found a mismatch in NumPy, which has been fixed in https://github.com/numpy/numpy/pull/18679. 

Please let me know if I miss something or get something wrong here.

radarhere:
This PR first replaces '1' with [METH_VARARGS](https://docs.python.org/3/c-api/structures.html#METH_VARARGS) in various locations, improving clarity. Changing code like this -

https://github.com/python-pillow/Pillow/blob/d393cfb65ae7d99b4b8e1cb8c0ced4cf79d12835/src/_imaging.c#L3421-L3425

to match how it is specified elsewhere in Pillow -

https://github.com/python-pillow/Pillow/blob/d393cfb65ae7d99b4b8e1cb8c0ced4cf79d12835/src/_imagingft.c#L1228-L1230

Then, this PR resolves #5487. The issue identifies 7 functions where we can use [METH_NOARGS](https://docs.python.org/3/c-api/structures.html#METH_NOARGS ) when no arguments are being used, rather than always using [METH_VARARGS](https://docs.python.org/3/c-api/structures.html#METH_VARARGS). This update is similar to #4441

### DIFF
--- a/src/_imaging.c
+++ b/src/_imaging.c
@@ -1086,7 +1086,7 @@ _getpalette(ImagingObject *self, PyObject *args) {
 }
 
 static PyObject *
-_getpalettemode(ImagingObject *self, PyObject *args) {
+_getpalettemode(ImagingObject *self) {
     if (!self->image->palette) {
         PyErr_SetString(PyExc_ValueError, no_palette);
         return NULL;
@@ -2095,12 +2095,12 @@ _box_blur(ImagingObject *self, PyObject *args) {
 /* -------------------------------------------------------------------- */
 
 static PyObject *
-_isblock(ImagingObject *self, PyObject *args) {
+_isblock(ImagingObject *self) {
     return PyBool_FromLong(self->image->block != NULL);
 }
 
 static PyObject *
-_getbbox(ImagingObject *self, PyObject *args) {
+_getbbox(ImagingObject *self) {
     int bbox[4];
     if (!ImagingGetBBox(self->image, bbox)) {
         Py_INCREF(Py_None);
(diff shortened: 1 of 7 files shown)


hugovk:
Thanks!

PR #5494, opened 2021-05-14
Add OSError info for load function

kratel:
Problem: When a truncated image is loaded, without `LOAD_TRUNCATED_IMAGES` set to `True` the load command fails without specifying which file failed to load.

In the current output the filename is not specified.
```python
>>> from PIL import Image
>>> test_filename = "Tests/images/truncated_image.png"
>>> try:
...     with Image.open(test_filename) as im:
...             im.load()
... except OSError as oe:
...     print(oe)
...     print(oe.args)
...     print(oe.filename)
...     print(oe.filename2)
... 
image file is truncated
('image file is truncated',)
None
None
```

Assuming you don't want to enable the  `LOAD_TRUNCATED_IMAGES` flag.

This makes handling this type of error difficult since when you catch it, you can't recover the name of the troublesome file that caused it. A workaround would be to catch the error and then return the filename that was used to call the function. However other libraries that use Pillow might make this call, and therefore the user does not make the call to that function directly and might not be able to do this. It would be simpler to pass in the filename when generating the error. That way when it is raised back up the user can still track down the troublesome image file.

I wasn't sure what `errno` to use, so I left it as `None`.

Error with more info:
```python
>>> from PIL import Image
>>> test_filename = "Tests/images/truncated_image.png"
>>> try:
...     with Image.open(test_filename) as im:
...             im.load()
... except OSError as oe:
...     print(oe)
...     print(oe.args)
...     print(oe.filename)
...     print(oe.filename2)
... 
[Errno None] image file is truncated: 'Tests/images/truncated_image.png'
(None, 'image file is truncated')
Tests/images/truncated_image.png
None
```

### DIFF
--- a/Tests/test_imagefile.py
+++ b/Tests/test_imagefile.py
@@ -120,12 +120,14 @@ class TestImageFile:
     @skip_unless_feature("zlib")
     def test_truncated_with_errors(self):
         with Image.open("Tests/images/truncated_image.png") as im:
-            with pytest.raises(OSError):
+            with pytest.raises(OSError) as oseinfo:
                 im.load()
+            assert "Tests/images/truncated_image.png" in str(oseinfo.value)
 
             # Test that the error is raised if loaded a second time
-            with pytest.raises(OSError):
+            with pytest.raises(OSError) as oseinfo:
                 im.load()
+            assert "Tests/images/truncated_image.png" in str(oseinfo.value)
 
     @skip_unless_feature("zlib")
     def test_truncated_without_errors(self):
--- a/src/PIL/ImageFile.py
+++ b/src/PIL/ImageFile.py
@@ -240,7 +240,8 @@ class ImageFile(Image.Image):
                                 if LOAD_TRUNCATED_IMAGES:
                                     break
                                 else:
-                                    raise OSError("image file is truncated") from e
+                                    raise OSError(None, "image file is truncated",
+                                                  self.filename) from e
 
                             if not s:  # truncated jpeg
                                 if LOAD_TRUNCATED_IMAGES:
--- a/src/PIL/ImageFile.py
+++ b/src/PIL/ImageFile.py
@@ -240,8 +240,9 @@ class ImageFile(Image.Image):
                                 if LOAD_TRUNCATED_IMAGES:
                                     break
                                 else:
-                                    raise OSError(None, "image file is truncated",
-                                                  self.filename) from e
+                                    raise OSError(
+                                        None, "image file is truncated", self.filename
+                                    ) from e
 
                             if not s:  # truncated jpeg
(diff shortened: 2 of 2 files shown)


radarhere:
Here are some thoughts.
- If another library is using Pillow and obscures which file caused the problem, I think that's the responsibility of that library. If you are purely using Pillow, it is absolutely clear which file caused the error - we don't process multiple files at once... I think ever.
- Is there any reason why this error is special? Surely there are other errors that occur during `load()` that this could apply to equally well, and inserting the filename into a bunch of errors seems like unnecessarily complexity.
- It's also not guaranteed that an image will have a meaningful filename - if you open an image from a file pointer, `filename` is an empty string.
```pycon
>>> with open("Tests/images/hopper.png", "rb") as fp:
...   with Image.open(fp) as im:
...     print(im.filename)
... 

>>>
```
- All of that said, there is actually an instance already where we pass the filename back with the error 
https://github.com/python-pillow/Pillow/blob/186145e5a0ced68b4b12b9ba52a05a006e46a1ac/src/PIL/Image.py#L2977-L2979

kratel:
I did see that error and saw that it returned a filename. That is why I was confused that the errors in `load()` did not return filenames. 

There is nothing particularly special about that error. It's just the one I ran into during a coding project. I agree that if it makes sense to add this enrichment, then it would follow that it should be done in other places where `OSError`s are created. 

I misunderstood the `Image.__init__` constructor, I thought `self.filename` would always have something. I could fix my error to do this instead to pass a valid filename.
```python
self.filename if self.filename else self.fp.name
```

I only wrote this for one OSError because it was the one I encountered. I can definitely add it to the other errors if that makes sense to do. If the responsibility is placed on external libraries that use Pillow then I guess it isn't needed.

I thought it made sense since OSError was being used. In the case of `open()` it returns the filename anyway even though it processes a single file every time.
```python
 >>> open("doesnotexist.txt")
FileNotFoundError: [Errno 2] No such file or directory: 'doesnotexist.txt'
```

kratel:
Hey @radarhere , should I add this for other errors? Or is this not necessary and I should just go ahead and close?

radarhere:
My personal opinion is that it's not necessary.

@hugovk did you have any thoughts?

hugovk:
If this was to be merged, it would need further work to deal with the file pointer case. And then it would make sense to update the other places too.

But is this necessary? The calling code does know the filename it's calling Pillow with, and if it's a filename or pointer, so it's not really necessary.

Would it help? Possibly. The stdlib `open` comparison is interesting, but on balance I'm thinking the extra complexity might not be worth it.

kratel:
Fair enough and thanks for the comments, will close the PR then.

PR #5497, opened 2021-05-14
Linked if blocks into if-else blocks

radarhere:
Minor cleanup

### DIFF
--- a/src/libImaging/Draw.c
+++ b/src/libImaging/Draw.c
@@ -1004,9 +1004,7 @@ clip_tree_do_clip(
         end->type = -1;
         end->next = NULL;
         *ret = start;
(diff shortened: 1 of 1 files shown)


hugovk:
I think the original is a bit clearer: 

return early means I know there's nothing more to run for that block, let's just get out of here quick, I don't even need to read any further (no "is there more logic further down this function I need to read?")

And then we don't need to add `else` afterwards, because when we made it here, we didn't return early, we're just continuing.

radarhere:
Ok, sure. Thanks for reviewing

wiredfool:
FWIW, I’m with Hugo on this one.  Early return is a good pattern.

PR #5500, opened 2021-05-16
Add support for multiple component transformation to JPEG2000

### LINKED ISSUE #5485: JPEG2000 irreversible option misleading
The [documentation](https://pillow.readthedocs.io/en/latest/handbook/image-file-formats.html#jpeg-2000) for the J2K plugin says:
> **irreversible**
If `True`, use the lossy Irreversible Color Transformation followed by DWT 9-7. Defaults to `False`, which means to use the Reversible Color Transformation with DWT 5-3.

I think the "color transformation" here is referring to the multiple component transformation (MCT), however looking at the source code, MCT isn't being applied, so no color transform is occurring (as I understand it `irreversible` in openjpeg only controls which DWT is used). This is visible in the output codestream:

```python
from io import BytesIO
import numpy as np
from PIL import Image
from struct import unpack

# https://github.com/python-pillow/Pillow/blob/master/Tests/images/xmp_test.jpg
im = Image.open("xmp_test.jpg")
assert im.mode == "RGB"
ref = np.asarray(im)
b = BytesIO()
b.name = "foo.j2k"  # workaround so no JP2
im.save(b, format="JPEG2000", irreversible=False)

# Parse J2K codestream
codestream = b.getvalue()
print(f"len: {len(codestream)}")
if codestream[0:2] == b'\xff\x4f':
    if codestream[2:4] == b'\xff\x51':
        lsiz = unpack('>H', codestream[4:6])[0]
        ssiz = codestream[42]
        if ssiz & 0x80:
            print(f"precision: {(ssiz & 0x7F) + 1}")
            print("signed: true")
        else:
            print(f"precision: {ssiz + 1}")
            print("signed: false")

        cod_offset = lsiz + 4
        if codestream[cod_offset:cod_offset + 2] == b'\xff\x52':
            # 0 for none, 1 for applied to components 0, 1, 2
            print(f"MCT: {codestream[cod_offset + 8]}")
            # 0 for DWT 9-7 (irreversible), 1 for DWT 5-3 (reversible)
            print(f"DWT: {codestream[cod_offset + 13]}")

im = Image.open(b)
assert np.array_equal(ref, np.asarray(im))
```
Produces:
```
len: 3506536
precision: 8
signed: false
MCT: 0
DWT: 1
```

For comparison, with Jpeg2KEncode.c modified to use MCT:
```
len: 2682646
precision: 8
signed: false
MCT: 1
DWT: 1
```
With the transform into YCbCr the compression ratio is higher, as expected, while still being reversible.

* OS: Ubuntu 20.04
* Python: 3.9
* Pillow: current master d393cfb

scaramallion:
Closes #5485

 * Add `mct` option for JPEG2000 that allows the use of multiple component transformation when encoding
 * Add `no_jp2` option for JPEG2000 to allow specifying the format when not saving to file
 * Updates handbook with new options and revises `irreversible` entry

OpenJPEG's `tcp_mct` is defined [here](https://github.com/uclouvain/openjpeg/blob/b6b8d28b3a85b74ff5415565cff2c20c019ca3c5/src/lib/openjp2/openjpeg.h#L522)

### DIFF
--- a/Tests/test_file_jpeg2k.py
+++ b/Tests/test_file_jpeg2k.py
@@ -271,7 +271,6 @@ def test_mct():
         test_card.save(out, "JPEG2000", mct=val, no_jp2=True)
         out.seek(0)
         with Image.open(out) as im:
-            im.load()
             assert_image_similar(im, test_card, 1.0e-3)
             assert out.getvalue()[59] == val
 
@@ -283,7 +282,6 @@ def test_mct():
 
         out.seek(0)
         with Image.open(out) as im:
-            im.load()
             assert_image_similar(im, jp2, 1.0e-3)
             assert out.getvalue()[53] == 0
 
--- a/Tests/test_file_jpeg2k.py
+++ b/Tests/test_file_jpeg2k.py
@@ -209,6 +209,37 @@ def test_layers():
         assert_image_similar(im, test_card, 0.4)
 
 
+def test_use_jp2():
+    out = BytesIO()
+    test_card.save(out, "JPEG2000", use_jp2=False)
+    out.seek(0)
+    assert out.read(2) == b"\xff\x4f"
+
+
+def test_mct():
+    # Three component
+    for val in (0, 1):
+        out = BytesIO()
+        test_card.save(out, "JPEG2000", mct=val, use_jp2=False)
+        out.seek(0)
+        with Image.open(out) as im:
+            im.load()
+            assert_image_similar(im, test_card, 1.0e-3)
+            assert out.getvalue()[59] == val
+
+    # Single component
+    for val in (0, 1):
+        out = BytesIO()
+        with Image.open("Tests/images/16bit.cropped.jp2") as jp2:
+            jp2.save(out, "JPEG2000", mct=val, use_jp2=False)
+
+        out.seek(0)
+        with Image.open(out) as im:
+            im.load()
+            assert_image_similar(im, jp2, 1.0e-3)
+            assert out.getvalue()[53] == 0
+
+
 def test_rgba():
     # Arrange
     with Image.open("Tests/images/rgb_trns_ycbc.j2k") as j2k:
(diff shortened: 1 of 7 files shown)


scaramallion:
Failure in appveyor build not related to PR

Tests/test_file_jpeg2k.py
```diff
@@ -168,6 +168,85 @@ def test_layers():
...
+
+
+def test_mct():
+    # Three component
+    for val in (0, 1):
+        out = BytesIO()
+        test_card.save(out, "JPEG2000", mct=val, no_jp2=True)
+        out.seek(0)
+        with Image.open(out) as im:
+            im.load()
```
radarhere, line 233:
```suggestion
```
I don't think this is necessary.

Tests/test_file_jpeg2k.py
```diff
@@ -168,6 +168,85 @@ def test_layers():
...
+
+    # Single component should have MCT disabled
+    for val in (0, 1):
+        out = BytesIO()
+        with Image.open("Tests/images/16bit.cropped.jp2") as jp2:
+            jp2.save(out, "JPEG2000", mct=val, no_jp2=True)
+
+        out.seek(0)
+        with Image.open(out) as im:
+            im.load()
```
radarhere, line 245:
```suggestion
```

scaramallion:
Not sure why that PyPy build failed. Similar to [this](https://github.com/python-pillow/Pillow/runs/2482254315?check_suite_focus=true)?

src/encode.c
```diff
@@ -1182,11 +1182,12 @@ PyImaging_Jpeg2KEncoderNew(PyObject *self, PyObject *args) {
     OPJ_PROG_ORDER prog_order;
     char *cinema_mode = "no";
     OPJ_CINEMA_MODE cine_mode;
+    char mct;
```
nulano, line 1185: Optional parameters should have a default value.
```suggestion
    char mct = 0;
```

scaramallion: Done

hugovk (approved):
I'm okay with this going in tomorrow's 9.1.0. If so, let's add verisonadded for the new params in image-file-formats.rst, and mention in release notes.

scaramallion:
Thanks!

Issue #5499, opened 2021-05-15
Support BC5 DDS encoding


pepperoni505:
BC5 is currently not supported by Pillow. There is already some code in the repo that can read BC5 files. I would have made a PR for this but I'm not familiar with how this project is laid out. 

```python
code goes here
```

radarhere:
Hi. You may be interested to know that there is a PR in review that adds support for saving uncompressed DDS textures - #5402.

Where is the code that can read BC5 files? I look at https://github.com/python-pillow/Pillow/blob/master/src/PIL/DdsImagePlugin.py and I can only see code to read uncompressed textures, DXT1, DXT3, DXT5 and DX10? If you can't see it, attaching an BC5 file here that you can open with Pillow would be another way of answering my question.

pepperoni505:
> Hi. You may be interested to know that there is a PR in review that adds support for saving uncompressed DDS textures - #5402.
> 
> 
> 
> Where is the code that can read BC5 files? I look at https://github.com/python-pillow/Pillow/blob/master/src/PIL/DdsImagePlugin.py and I can only see code to read uncompressed textures, DXT1, DXT3, DXT5 and DX10? If you can't see it, attaching an BC5 file here that you can open with Pillow would be another way of answering my question.

Hello. Sorry, I should have been more specific. I was talking about code in the following file: `src/libImaging/BcnDecode.c`. I don't know how easy it would be to add it into `src/PIL/DdsImagePlugin.py`.

pepperoni505:
I tried modifying `src/PIL/DdsImagePlugin.py` to accept BC5 files, but it doesn't fully work. Here is a diff:
```diff
diff --git a/src/PIL/DdsImagePlugin.py b/src/PIL/DdsImagePlugin.py
index 1a7fe003..83c6c16d 100644
--- a/src/PIL/DdsImagePlugin.py
+++ b/src/PIL/DdsImagePlugin.py
@@ -97,6 +97,9 @@ DXT5_FOURCC = 0x35545844
 DXGI_FORMAT_R8G8B8A8_TYPELESS = 27
 DXGI_FORMAT_R8G8B8A8_UNORM = 28
 DXGI_FORMAT_R8G8B8A8_UNORM_SRGB = 29
+DXGI_FORMAT_BC5_TYPELESS = 82
+DXGI_FORMAT_BC5_UNORM = 83
+DXGI_FORMAT_BC5_SNORM = 84
[...]

PR #5501, opened 2021-05-16
Added DDS BC5_UNORM reading

radarhere:
Helps #5499

Using the test image from that issue, all that was missing from Pillow's C implementation was that the alpha channel was set to zero. This PR fixes that, and declares the format to be available at the Python level.

See https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpcr2/98e2dbc9-3a1d-4ab2-93e3-152eeebd17a3 for documentation that `DXGI_FORMAT_BC5_UNORM` is 0x00000053 = 83.


hugovk:
Let's also add this to release notes 👍

radarhere:
Sure, done

PR #5513, opened 2021-05-26
Added ICO saving in BMP format

### LINKED ISSUE #2512: Support saving uncompressed ICO
Currently _Pillow_ basically writes multiple _PNGs_ into an _ICO_ container, which is not supported by some applications that can read _ICO_. I'm not familiar with the _ICO_ format, but I think `biCompression` [here](https://msdn.microsoft.com/en-us/library/vs/alm/dd183376(v=vs.85).aspx) references the possible sub-formats.

radarhere:
Resolves #2512

ICO can either be a container for [PNG or BMP](https://en.wikipedia.org/wiki/ICO_(file_format)).

Pillow has been saving ICOs only in the PNG format. This PR adds support for saving ICOs with the BMP format, largely by reversing the process that we use to load the BMP format. It would be called with ``im.save("out.ico", bitmap_format="bmp")``

### DIFF
--- a/src/PIL/IcoImagePlugin.py
+++ b/src/PIL/IcoImagePlugin.py
@@ -30,6 +30,7 @@ from math import ceil, log
 from . import BmpImagePlugin, Image, ImageFile, PngImagePlugin
 from ._binary import i16le as i16
 from ._binary import i32le as i32
+from ._binary import o32le as o32
 
 #
 # --------------------------------------------------------------------
@@ -53,6 +54,7 @@ def _save(im, fp, filename):
     sizes = list(sizes)
     fp.write(struct.pack("<H", len(sizes)))  # idCount(2)
     offset = fp.tell() + len(sizes) * 16
+    bmp = im.encoderinfo.get("bitmap_format") == "bmp"
     provided_images = {im.size: im for im in im.encoderinfo.get("append_images", [])}
     for size in sizes:
         width, height = size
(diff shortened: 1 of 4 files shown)


src/PIL/IcoImagePlugin.py
```diff
@@ -62,17 +64,30 @@ def _save(im, fp, filename):
...
+
+        image_io = BytesIO()
+        if bmp:
+            tmp.save(image_io, "dib")
+
+            if bits != 32:
+                and_mask = Image.new("1", tmp.size)
+                ImageFile._save(
+                    and_mask, image_io, [("raw", (0, 0) + tmp.size, 0, ("1", 0, -1))]
+                )
```
radarhere, line 84: Our 1, L, P and RGB modes don't have any transparency, so the AND mask can just be constant.

JamesTheAwesomeDude: Will transparency still work for BMP-formatted ICO images?

radarhere: RGBA images saved to BMP-formatted ICO will have transparency, because 32-bit images don't require the AND mask (see https://en.wikipedia.org/wiki/ICO_(file_format) for more information).

hugovk:
Looks good, let's also add to release notes.

radarhere:
Done.

hugovk:
Thank you!

Issue #5370, opened 2021-03-31
Loading large TIFF fails with OSError: -9


martinleopold:
### What did you do?

Load a large TIFF created previously with Pillow.

### What did you expect to happen?

Load without error.

### What actually happened?

        Traceback (most recent call last):
          File "large_tiff_bug.py", line 15, in <module>
            img.load()
          File "/Users/mlg/code/stocks-map/venv/stocks-map/lib/python3.8/site-packages/PIL/TiffImagePlugin.py", line 1088, in load
            return self._load_libtiff()
          File "/Users/mlg/code/stocks-map/venv/stocks-map/lib/python3.8/site-packages/PIL/TiffImagePlugin.py", line 1192, in _load_libtiff
            raise OSError(err)
        OSError: -9

### What are your OS, Python and Pillow versions?

* OS: macOS 11.2.3
* Python: 3.8.8
* Pillow: 8.1.2
* PIL.features.version_codec('libtiff'): 4.2.0

Code:
```python
from PIL import Image

width  = 30_000
height = 35_792

print(f'creating test image: {width:,} x {height:,} = {width * height:,}; {width * height * 2:,} bytes')
img = Image.new( 'I;16', (width, height) ) # I'm using 16 bit greyscale 
img.save( '_tmp.tiff', compression='tiff_adobe_deflate' ) # add zip compression, so filesize doesn't get out of hand

print()
print(f'loading test image')
Image.MAX_IMAGE_PIXELS = 1_800_000_000
img = Image.open('_tmp.tiff')
img.load()
```

I believe we are hitting the size limit limit [here](https://github.com/python-pillow/Pillow/blob/master/src/libImaging/TiffDecode.c#L489), because reducing `height` to `35_791` in the above code makes it work.

30.000 x 35.792 x 2 bytes = 2.147.520.000 bytes -> error -9 ([IMAGING_CODEC_MEMORY](https://github.com/python-pillow/Pillow/blob/master/src/libImaging/Imaging.h#L686))
30.000 x 35.791 x 2 bytes = 2.147.460.000 bytes  -> works

This leads me to believe that INT_MAX is 2.147.483.647, so it's an `int32` (32 bit, signed), where it could be 64 bit (on supported platforms) and unsigned to maximize the size of loadable images.

[...]

PR #5514, opened 2021-05-27
Limit TIFF strip size when saving with libtiff

kmilos:
Adresses #5370 partly.

Doesn't really "fix" the loading of any existing >2GB single strip images (would need to handle specially by buffered read in decoder), but reduces their incidence as recommended by the TIFF spec.

### DIFF
--- a/Tests/test_file_libtiff.py
+++ b/Tests/test_file_libtiff.py
@@ -9,7 +9,7 @@ from ctypes import c_float
 import pytest
 
 from PIL import Image, ImageFilter, TiffImagePlugin, TiffTags, features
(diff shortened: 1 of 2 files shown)


kmilos:
Still work to do: StripOffsets and StripByteCounts are assumed to be just single values instead of being derived and stored as arrays if multiple strips are present.

kmilos:
Changed to only fix compressed saving (or when libtiff writer is forced), as those cannot be read otherwise.

Uncompressed internal writer still uses a single strip (updating the strip offsets still TODO), as those files can be read when #5517 is merged.

PR #5515, opened 2021-05-28
Fix differing background colors in right sidebar in dark mode docs

PH-KDX:
Fixes differently colored blocks in the right sidebar in the dark mode docs.

For comparison, see the following two images:
Before fix (current situation):
before_fix
After fix:
after_fix

Change proposed in this pull request:

 * Make the `background-color` of the top part of the right sidebar in the docs darker when in dark mode to match the rest of the right sidebar

### DIFF
--- a/docs/resources/css/dark.css
+++ b/docs/resources/css/dark.css
@@ -1275,7 +1275,7 @@
 
     .wy-body-for-nav {
         background-image: initial;
-        background-color: rgb(26, 28, 29);
+        background-color: rgb(24, 26, 27);
     }
 
     .wy-nav-side {
@@ -1335,7 +1335,7 @@
 
     .wy-body-for-nav {
         background-image: initial;
-        background-color: rgb(26, 28, 29);
+        background-color: rgb(24, 26, 27);
     }
 
     @media screen and (min-width: 1100px) {


Null-B:
wow that is soo
cool

docs/resources/css/dark.css
```diff
@@ -1275,7 +1275,7 @@
 
     .wy-body-for-nav {
         background-image: initial;
-        background-color: rgb(26, 28, 29);
+        background-color: rgb(24, 26, 27);
```
radarhere, line 1278: How did you come up with this new value?

PH-KDX:
Got it from the bottom half with Inspect Element; it also has a better contrast with the main text block. Several members of the Python Discord helped decide on it.

docs/resources/css/dark.css
PH-KDX: Got it from the bottom half with Inspect Element; it also has a better contrast with the main text block. Several members of the Python Discord helped decide on it.

PH-KDX:
What's the current status on this?

hugovk:
The current status is... merged! Thanks for the PR!

PH-KDX:
Perfect, thanks!

PR #5517, opened 2021-05-31
Explicitly enable strip chopping for large uncompressed TIFFs

### LINKED ISSUE #5370: Loading large TIFF fails with OSError: -9

kmilos:
Helps #5370 (only works for uncompressed images)

### DIFF
--- a/src/libImaging/TiffDecode.c
+++ b/src/libImaging/TiffDecode.c
@@ -543,7 +543,7 @@ ImagingLibTiffDecode(
     Imaging im, ImagingCodecState state, UINT8 *buffer, Py_ssize_t bytes) {
     TIFFSTATE *clientstate = (TIFFSTATE *)state->context;
     char *filename = "tempfile.tif";
-    char *mode = "r";
+    char *mode = "rC";
     TIFF *tiff;
     uint16_t photometric = 0;  // init to not PHOTOMETRIC_YCBCR
     uint16_t compression;


radarhere:
Would you be able to add a test, that fails without your change and passes with it?

kmilos:
Hm, [I just pushed a test to my branch](https://github.com/kmilos/Pillow/commit/1fc144089e8fdfed42df05e2ecc896280ab287df), but it's not showing up here?!

kmilos:
Ah, only internal `TiffImagePlugin.ImageFileDirectory_v2` is ever used to parse the TIFF IFDs (in both `Image.open()` and `TiffImagePlugin._load_libtiff()`, there's no way to actually see how libtiff is re-interpreting the IFD as far as I can tell, so not really sure how to test this...

Tests/test_file_libtiff.py
```diff
@@ -9,7 +9,7 @@
 import pytest
 
 from PIL import Image, ImageFilter, TiffImagePlugin, TiffTags, features
-from PIL.TiffImagePlugin import SUBIFD
+from PIL.TiffImagePlugin import SUBIFD, STRIPOFFSETS
```
radarhere, line 12:
```suggestion
from PIL.TiffImagePlugin import STRIPOFFSETS, SUBIFD
```
Lint fix.

kmilos: Saw that, but didn't bother since the test is not working...

radarhere:
https://linux.die.net/man/3/tifffdopen
> The default behaviour, for backwards compatibility, is to enable strip chopping.

Is it possible that this PR doesn't actually do anything?

kmilos:
I only know this: https://gitlab.com/libtiff/libtiff/-/blob/29219a5bd14040d60e376324a52ef91b68808143/libtiff/tif_open.c#L254-257

But it does seem the flags are initialized by default to do the chopping: https://gitlab.com/libtiff/libtiff/-/blob/29219a5bd14040d60e376324a52ef91b68808143/libtiff/tif_open.c#L161-164 

However, there was also recent (not yet released) PR that seems to indicate the default wasn't working on libtiff CMake builds (which is what Pillow uses lately AFAIK): https://gitlab.com/libtiff/libtiff/-/merge_requests/271

If this is the behaviour Pillow was expecting, doesn't hurt to be explicit ;)

radarhere:
Ok, if the default has been wrong in previous libtiff versions, then it seems like this would be good, as it will fix that.

radarhere:
libtiff 4.4.0 has now been released, fixing the default behaviour.

PR #5522, opened 2021-06-04
Fix lcms2_static target for msbuild

kmilos:
Fixes CI failures for Windows

For some unknown reason the "Build" target is indeed non-existant for the lcms2_static project, but VC2017 only (even when specified)! `msbuild Projects\VC2019\lcms2.sln /t:lcms2_static ` works :/


radarhere (approved):

radarhere:
Thanks

PR #5539, opened 2021-06-12
Modified docstring for GaussianBlur

kdpenner:

 * The argument for `GaussianBlur` is `radius`. The canonical name for the parameter is standard deviation. The value for the standard normal distribution at a given radius is a function of standard deviation.
 * `GaussianBlur` doesn't convolve the image with a Gaussian kernel. It uses a kernel that when applied iteratively approximates a Gaussian kernel.

The changes improve the precision of the documentation.

### DIFF
--- a/src/PIL/ImageFilter.py
+++ b/src/PIL/ImageFilter.py
@@ -149,9 +149,10 @@ class ModeFilter(Filter):
 
 
 class GaussianBlur(MultibandFilter):
-    """Gaussian blur filter.
+    """Blurs the image with an extended box filter, which approximates a
+    Gaussian kernel.
 
-    :param radius: Blur radius.
+    :param radius: Standard deviation of the Gaussian kernel.
     """
 
     name = "GaussianBlur"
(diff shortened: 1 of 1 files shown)


radarhere:
@homm you wrote the C implementation in #961. Did you want to quickly review this?

src/PIL/ImageFilter.py
```diff
@@ -149,9 +149,10 @@ def filter(self, image):
 
 
 class GaussianBlur(MultibandFilter):
-    """Gaussian blur filter.
+    """Blurs the image with an extended box filter, which approximates a
```
homm, line 152: "with a sequence of extended box filters"?

src/PIL/ImageFilter.py
```diff
@@ -149,9 +149,10 @@ def filter(self, image):
...
+    Gaussian kernel.
```
homm, line 153: Is it worth to mention that approximation is very accurate?

src/PIL/ImageFilter.py
kdpenner: Done

src/PIL/ImageFilter.py
kdpenner: Instead of "very accurate" I prefer to give a notion of the accuracy. In this case it depends on a few variables so I put in the link in the C header.

hugovk (approved):

homm (approved):

PR #5549, opened 2021-06-18
Ensure palette string matches RGB mode

### LINKED ISSUE #4830: PIL doesn't seem to handle the color map correctly in some color indexed TGA files
Please see my question in [Stack Overflow](https://stackoverflow.com/questions/63151906/is-there-any-other-attribute-that-affetcts-how-colors-are-displayed-on-an-indexe) where I explain everything in greater detail and more importantly, where Mark Setchell explains why he thinks there is a bug in PIL when handling some color indexed TGA files.

Basically, I am working with a sample color indexed TGA file to test a script that would invert its colors.

But before any of that, the issue is that PIL doesn't seem to handle the color map correctly when open it and the values of the color palette end up being wrong.

These are the sample files: [CCM8.TGA](https://www.fileformat.info/format/tga/sample/c36e10e31828483d832ef67561e20d5a/download) and [UCM8.TGA](https://www.fileformat.info/format/tga/sample/34402cd832c3456eb82628798f54709a/download).
And you will see that if you open show any of them:
```python
from PIL import Image
Image.open('CCM8.TGA').show()
```
PIL shows:
pQ6IH
When it should be:
IpPCy

As I said, I wrote all the details on that post. If you want me to submit more information here, please let me know.

I am using Pillow 7.0 on Fedora 32 and Python 3.8.

radarhere:
Resolves #4830

The image from that issue has a [BGR;15 palette](https://github.com/python-pillow/Pillow/blob/384a4bf01ec8d631dcde2c3246bd08e37fa1b7cd/src/PIL/TgaImagePlugin.py#L115-L117). So when PngImagePlugin tries to calculate the number of colors to use when saving for `show()`,

https://github.com/python-pillow/Pillow/blob/384a4bf01ec8d631dcde2c3246bd08e37fa1b7cd/src/PIL/PngImagePlugin.py#L1195

the assumption, that it can divide by 3 because this is RGB, is false.

This PR adds a change to Image, so that when it sets the palette mode to RGB, there is actually RGB data in the palette string.

### DIFF
--- a/Tests/test_file_tga.py
+++ b/Tests/test_file_tga.py
@@ -65,9 +65,14 @@ def test_sanity(tmp_path):
                         roundtrip(original_im)
 
 
-def test_palette_depth_16():
+def test_palette_depth_16(tmp_path):
     with Image.open("Tests/images/p_16.tga") as im:
         assert_image_equal_tofile(im.convert("RGB"), "Tests/images/p_16.png")
+
+        out = str(tmp_path / "temp.png")
+        im.save(out)
+        with Image.open(out) as reloaded:
+            assert_image_equal_tofile(reloaded.convert("RGB"), "Tests/images/p_16.png")
 
 
 def test_id_field():
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -830,7 +830,7 @@ class Image:
                 arr = bytes(
                     value for (index, value) in enumerate(arr) if index % 4 != 3
                 )
-            self.im.putpalette(mode, arr)
+            palette_length = self.im.putpalette(mode, arr)
             self.palette.dirty = 0
             self.palette.rawmode = None
             if "transparency" in self.info:
@@ -841,7 +841,7 @@ class Image:
                 self.palette.mode = "RGBA"
             else:
                 self.palette.mode = "RGB"
-                self.palette.palette = self.im.getpalette()
+                self.palette.palette = self.im.getpalette()[: palette_length * 3]
 
         if self.im:
             if cffi and USE_CFFI_ACCESS:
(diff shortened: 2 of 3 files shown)


radarhere:
The first commit is failing, as another test complains that the palette is now longer than it should be.

This is because I'm calling `getpalette()`, which has a [fixed palette length](https://github.com/python-pillow/Pillow/blob/384a4bf01ec8d631dcde2c3246bd08e37fa1b7cd/src/_imaging.c#L1064-L1066).

The C layer `ImagingPaletteInstance` [doesn't actually have a palette length](https://github.com/python-pillow/Pillow/blob/384a4bf01ec8d631dcde2c3246bd08e37fa1b7cd/src/libImaging/Imaging.h#L141-L150), so instead, I've returned the length from `putpalette` and used that in Python to trim the output.

radarhere:
The CI Fuzz failure is related. See https://github.com/radarhere/Pillow/actions/runs/951376400 from [just a comment change in C.](https://github.com/radarhere/Pillow/commit/e1dcbe8c383e88bc713afa0d6d2def20a2c84fff)

hugovk:
> The CI Fuzz failure is related. See [radarhere/Pillow/actions/runs/951376400](https://github.com/radarhere/Pillow/actions/runs/951376400) from [just a comment change in C.](https://github.com/radarhere/Pillow/commit/e1dcbe8c383e88bc713afa0d6d2def20a2c84fff)

You mean unrelated?

radarhere:
Yes, unrelated :p

radarhere:
I've created https://github.com/google/oss-fuzz/pull/5950 to fix the CI Fuzz failure.

radarhere:
PR has been merged, and CI Fuzz is working again.

PR #5554, opened 2021-06-24
Added tag data for IFD groups

### LINKED ISSUE #5553: EXIF tag `InteropVersion` has wrong type
## What did you do?

Open an image with Pillow and save it with its exif data.

## What did you expect to happen?

The content of the exif field `InteropVersion` must be the same in both files.

## What actually happened?

The type of `InteropVersion`  field value is set to `TiffTags.BYTE` (1)  by Pillow where it should be of type `TiffTags.UNDEFINED` (7) to conform to standard. Some software might rely on that field to assess the validity of the file.

## What are your OS, Python and Pillow versions?

- OS: Ubuntu 20.04
- Python: 3.8.5
- Pillow: 8.2.0 + fix from #5490

```python
from PIL import Image
#open a jpeg image
im = Image.open("original.jpg")
# save it with original parameters, but it doesn't change the problem to use default values
im.save("copy.jpg", exif=im.getexif(), quality=95, progressive=True, subsampling="4:4:4")
```

Original image:
pixelated

Result image:
original_copy

Extract of the EXIF data from the original file (using `exiftool -D -s original.jpg`):
```
40962 ExifImageWidth                  : 1920
40963 ExifImageHeight                 : 692
    1 InteropIndex                    : R98 - DCF basic file (sRGB)
    2 InteropVersion                  : 0100
```
In the saved file  (using `exiftool -D -s copy.jpg`):
```
40962 ExifImageWidth                  : 1920
40963 ExifImageHeight                 : 692
    1 InteropIndex                    : R98 - DCF basic file (sRGB)
    2 InteropVersion                  : 48 49 48 48
```

radarhere:
Resolves #5553

At the moment, our [list of tag data](https://github.com/python-pillow/Pillow/blob/70ef50cf72992ab6c9330412d6f1340c593e2d8f/src/PIL/TiffTags.py#L74) is 1 dimensional. It doesn't account for the fact that tags actually exist in different groups - see https://exiftool.org/TagNames/EXIF.html

This PR adds `TAGS_V2_GROUPS`, which is used when the tag in question is part of a group.

#5553 mentions that the wrong type is used for InteropVersion, part of the InteropIFD. The type information for that tag is added here.

### DIFF
--- a/Tests/test_file_tiff_metadata.py
+++ b/Tests/test_file_tiff_metadata.py
@@ -376,3 +376,17 @@ def test_too_many_entries():
 
     # Should not raise ValueError.
     pytest.warns(UserWarning, lambda: ifd[277])
+
+
+def test_tag_group_data():
+    base_ifd = TiffImagePlugin.ImageFileDirectory_v2()
+    interop_ifd = TiffImagePlugin.ImageFileDirectory_v2(group=40965)
+    for ifd in (base_ifd, interop_ifd):
+        ifd[2] = "test"
+        ifd[256] = 10
+
+    assert base_ifd.tagtype[256] == 4
+    assert interop_ifd.tagtype[256] != base_ifd.tagtype[256]
+
+    assert interop_ifd.tagtype[2] == 7
+    assert base_ifd.tagtype[2] != interop_ifd.tagtype[256]
--- a/src/PIL/TiffImagePlugin.py
+++ b/src/PIL/TiffImagePlugin.py
@@ -465,7 +465,7 @@ class ImageFileDirectory_v2(MutableMapping):
 
     """
 
-    def __init__(self, ifh=b"II\052\0\0\0\0\0", prefix=None):
+    def __init__(self, ifh=b"II\052\0\0\0\0\0", prefix=None, group=None):
         """Initialize an ImageFileDirectory.
 
         To construct an ImageFileDirectory from a real file, pass the 8-byte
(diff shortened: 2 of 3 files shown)


kmilos:
How is the possibility of a tag belonging to more than one group (==IFD?) handled?

For example, there is significant overlap between [TIFF/EP](http://www.barrypearson.co.uk/top2009/downloads/TAG2000-22_DIS12234-2.pdf) and [Exif](https://www.cipa.jp/std/documents/download_e.html?DC-008-Translation-2019-E) tags, but [in TIFF/EP they are part of the main image IFD0, while in Exif they are obviously part of ExifIFD](https://en.wikipedia.org/wiki/TIFF/EP#Differences_from_TIFF_and_Exif).

radarhere:
This PR doesn't force the tags to be unique across groups, no.

When saving the data in `tobytes()`, the current `tag` is passed to `ImageFileDirectory_v2`.
https://github.com/python-pillow/Pillow/blob/a2709be7f424d6a7a2bb196cfe39abdbbc495238/src/PIL/TiffImagePlugin.py#L832

Then, when looking up the info of a tag, this PR retrieves info according to the group. If `group` is not `None`, then it doesn't retrieve from `TAGS_V2`, and if it is not, then it only pulls from the group.

https://github.com/python-pillow/Pillow/blob/a2709be7f424d6a7a2bb196cfe39abdbbc495238/src/PIL/TiffTags.py#L45-L48

So both IFD0 and EXIFIFD could have the same tag - which group is used comes from how the data is structured coming into `tobytes()`.

PR #5564, opened 2021-06-29
Do not use background or transparency index for new color

### LINKED ISSUE #4202: Odd color change in GIF after overlaying text
## What did you do?
Tried to overlay some text over a GIF file

I expected the outcome to look almost identical to the original in terms of quality.

### What actually happened?
The final result quality text looks odd, and the GIF looks like it had it's color inverted

* OS: Windows
* Python: 3.8
* Pillow: 6.2.1

Code used to modify the GIF from a URL: 
```python
image = Image.open(BytesIO(await request.read()))
frames = []
for frame in ImageSequence.Iterator(image):
    frame_bytes = BytesIO()
    draw = ImageDraw.Draw(frame)
    draw_text(frame, draw, top_text, 'top')
    draw_text(frame, draw, bottom_text, 'bottom')
    frame.save(frame_bytes, format='GIF')
    frames.append(Image.open(frame_bytes))
    del draw
edited_image_bytes = BytesIO()
frames[0].save(edited_image_bytes, format='GIF', save_all=True, append_images=frames[1:])
```
draw_text and draw_text_with_outline methods: https://hastebin.com/xoxecosawu.py

Files: [Gifs.zip](https://github.com/python-pillow/Pillow/files/3822773/Gifs.zip)

radarhere:
Resolves #4202

When the code from that issue calls `getcolor()` to find a new color to draw with, it picks 255. However, the image from that issue has an `info["transparency"]` of 255. So Pillow draws with the transparency index, which produces unexpected results.

This PR skips the background and transparency index when calling `getcolor()` to prevent this.

### DIFF
--- a/Tests/test_imagepalette.py
+++ b/Tests/test_imagepalette.py
@@ -49,23 +49,27 @@ def test_getcolor():
         palette.getcolor("unknown")
 
 
-def test_getcolor_not_special():
-    im = Image.new("P", (1, 1))
-    for index, palette in {
+@pytest.mark.parametrize(
+    "index, palette",
+    [
         # Test when the palette is not full
-        0: ImagePalette.ImagePalette(),
+        (0, ImagePalette.ImagePalette()),
         # Test when the palette is full
-        255: ImagePalette.ImagePalette("RGB", list(range(256)) * 3),
-    }.items():
-        # Do not use transparency index as a new color
-        im.info["transparency"] = index
-        index1 = palette.getcolor((0, 0, 0), im)
-        assert index1 != index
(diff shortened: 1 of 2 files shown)


Tests/test_imagepalette.py
```diff
@@ -49,6 +49,25 @@ def test_getcolor():
         palette.getcolor("unknown")
 
 
+def test_getcolor_not_special():
+    im = Image.new("P", (1, 1))
+    for index, palette in {
```
hugovk, line 54: Replace loop with `@pytest.mark.parametrize`?

hugovk (approved):

PR #5567, opened 2021-06-30
Use snprintf instead of sprintf

radarhere:
Resolves
https://github.com/python-pillow/Pillow/blob/b5c15e93f29bf4fe5071071b0ccc9450ab1920c8/src/libImaging/Convert.c#L1598
and
https://github.com/python-pillow/Pillow/blob/b5c15e93f29bf4fe5071071b0ccc9450ab1920c8/src/libImaging/Convert.c#L1649

### DIFF
--- a/src/libImaging/Convert.c
+++ b/src/libImaging/Convert.c
@@ -1594,9 +1594,8 @@ convert(
 #ifdef notdef
         return (Imaging)ImagingError_ValueError("conversion not supported");
 #else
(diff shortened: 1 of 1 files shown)


wiredfool (changes requested):
Need to truncate the mode prior to passing to sprintf

radarhere:
This is the fix for CVE-2021-34552

PR #5568, opened 2021-06-30
Add support for reading DPI information from JPEG2000 images

rogermb:

 * Currently, Pillow does not read the DPI information for JPEG2000 images, unlike it does for other image formats.
This PR seeks to remedy this issue by also parsing the "resc" JPEG2000 header box if it is present.
 * I didn't want to copy & paste the same box header reading code a third time due to the hierarchical format of the JPEG2000 header, so I wrapped the common header reading code in a small helper class (`BoxReader`), see the first commit.
 * However, this only adds support for reading DPI information, as OpenJPEG currently lacks support for writing the required "res " and "resc" header boxes. There are already years-old open issues addressing this shortcoming: [OpenJPEG/#378](https://github.com/uclouvain/openjpeg/issues/378), [OpenJPEG/#788](https://github.com/uclouvain/openjpeg/issues/788).

This is my first contribution to Pillow, so please let me know if I messed something up :)


rogermb:
I'm not sure what's up with the pypy-3.6 x86 build failing on Windows. It worked [on my fork](https://github.com/rogermb/Pillow/runs/2953128585) and I haven't even modified the file that seems to be causing problems, `Tests/test_image_getextrema.py`.

Perhaps it's just a random failure and running the Windows build again will fix it?

radarhere:
I ran the job again and it is passing now

src/PIL/Jpeg2KImagePlugin.py
```diff
@@ -114,11 +178,10 @@ def _parse_jp2_header(fp):
                     mode = "RGB"
                 elif nc == 4:
                     mode = "RGBA"
-                break
         elif tbox == b"colr":
-            meth, prec, approx = struct.unpack_from(">BBB", content)
-            if meth == 1:
-                cs = struct.unpack_from(">I", content, 3)[0]
+            meth, prec, approx = header.read_fields(">BBB")
+            if meth == 1 and unkc == 0:
```
radarhere, line 183: https://www.ics.uci.edu/~dan/class/267/papers/jpeg2000.pdf
> Colourspace Unknown. This field specifies if the actual colourspace of the image data is known. This field is encoded as a 1-byte unsigned integer. Legal values for this field are 0, if the colourspace of the image is known and correctly specified the colourspace boxes within the file, or 1, if the colourspace of the image is not known. A value of 1 will be used in cases such as the transcoding of legacy images where the actual colourspace of the image data is not known. In those cases, while the colourspace interpretation methods specified in the file may not accurately reproduce the image with respect to some original, the image should be treated as if the methods do accurately reproduce the image.

The way I read that, even if `unkc` is 1, we should still use `colr` regardless, so this change isn't needed. Any thoughts?

rogermb: The reason for this change was to preserve the previous behavior of the implementation: If an "ihdr" box was found with unkc != 0, we'd break from the loop without reading from "colr": https://github.com/python-pillow/Pillow/blob/3346389ace0537b17360928ca59bff0a50f94041/src/PIL/Jpeg2KImagePlugin.py#L103-L117

---

That being said, I agree with your reading of the spec. As I interpret it: while the colorspace information may be inaccurate, we are to treat it as if it were accurate, and thus read it from "colr".

However, if this is the case, then we may as well stop trying to infer `mode` from the number of channels (`nc` in "ihdr") if `unkc` is set, because then we should only assign `mode` based on the values in the "colr" header.

Unless... would we want to accept unsupported `cs` / `nc` values in the "colr" header if `unkc` is set? That's the behavior you'd get if you'd just remove `and unkc == 0` from the commented-on if statement. I'm really unsure about the correct behavior in such a situation.

Nevertheless, I think that a change that alters how the colorspace information is parsed is outside the scope of this PR 😄

src/PIL/Jpeg2KImagePlugin.py
```diff
@@ -154,9 +154,6 @@ def _parse_jp2_header(fp):
             if reader.read_fields(">4s")[0] == b"jpx ":
                 mimetype = "image/jpx"
 
-    if header is None:
-        raise SyntaxError("Could not find JP2 header")
```
rogermb: You're right, this is unreachable, `reader.has_next_box()` will never return False (because `reader = BoxReader(fp)` is initialized without a length), so the only possible outcome for a malformed image not containing a "jp2h" header is that `reader.next_box_type()` will fail with some kind of exception.

While not exactly elegant, that's definitely better than having an if statement that misleads the reader. 👍

src/PIL/Jpeg2KImagePlugin.py
```diff
@@ -33,12 +33,12 @@ def __init__(self, fp, length=-1):
         self.remaining_in_box = -1
 
     def _can_read(self, num_bytes):
+        if self.has_length and self.fp.tell() + num_bytes > self.length:
+            # Outside box: ensure we don't read past the known file length
+            return False
         if self.remaining_in_box >= 0:
```
rogermb, line 39: This could technically also be an elif for symmetry and whatnot -- not that it matters, since we return from the previous if branch anyway. 👍 on spotting and fixing the bug where we could read past the parent box length!

src/PIL/Jpeg2KImagePlugin.py
```diff
@@ -131,11 +131,8 @@ def _res_to_dpi(num, denom, exp):
     """Convert JPEG2000's (numerator, denominator, exponent-base-10) resolution,
     calculated as (num / denom) * 10^exp and stored in dots per meter,
     to floating-point dots per inch."""
-    if num == 0 or denom == 0:
-        raise SyntaxError(
-            f"Invalid JP2 resolution information: ({num} / {denom}) * 10^{exp}"
-        )
-    return (254 * num * (10 ** exp)) / (10000 * denom)
+    if denom != 0:
+        return num / denom * (10 ** exp) * 0.0254
```
rogermb, line 135: I should've probably clarified this in a code comment: the reason for my roundabout way of calculating the resolution was to remain precise by working with ints for as long as possible, and to only have a single (division) operation which introduces floating point errors.

While your code is more is more elegant and more easily readable, there are now 3 operations that can introduce slight numeric errors.

I'm not sure whether my approach is overkill, though. It's not like those last few mantissa bits are ever realistically going to matter.

rogermb:
Also, thank you for the additional test cases 🚀

src/PIL/Jpeg2KImagePlugin.py
radarhere: Thanks for the explanation. Yep, fair enough trying to keep the behaviour the same.

src/PIL/Jpeg2KImagePlugin.py
radarhere: Ah, ok. I was thinking of being consistent with how this value is used [elsewhere](https://github.com/python-pillow/Pillow/blob/9f42a17b2d03631a8131ae0f7ebc27bed6309482/src/PIL/PngImagePlugin.py#L503) in Pillow.

I'll switch back to your version.

rogermb: Awesome, thanks! 😄

rogermb:
LGTM! Thank you for the code improvements 😄

PR #5572, opened 2021-07-01
Make Image.__array__ take optional dtype argument

### LINKED ISSUE #5571: Pillow 8.3 and NumPy
Throws exception with Pillow 8.3: `TypeError: __array__() takes 1 positional argument but 2 were given`

````python
with PIL.Image.open(filepath) as img:
    numpy.array( img, dtype=numpy.float32 )
````

### PR DESCRIPTION
**Pull request by t-vi**
This is required by the numpy protocol.

Fixes #5571

### DIFF
--- a/Tests/test_image_array.py
+++ b/Tests/test_image_array.py
@@ -31,7 +31,7 @@ def test_toarray():
     assert test("RGBA") == ((100, 128, 4), "|u1", 51200)
     assert test("RGBX") == ((100, 128, 4), "|u1", 51200)
 
-    test_with_dtype(numpy.float)
+    test_with_dtype(numpy.float64)
     test_with_dtype(numpy.uint8)
 
     with Image.open("Tests/images/truncated_jpeg.jpg") as im_truncated:
--- a/Tests/test_image_array.py
+++ b/Tests/test_image_array.py
@@ -14,6 +14,10 @@ def test_toarray():
         ai = numpy.array(im.convert(mode))
         return ai.shape, ai.dtype.str, ai.nbytes
 
+    def test_with_dtype(dtype):
+        ai = numpy.array(im, dtype=dtype)
+        assert ai.dtype == dtype
+
     # assert test("1") == ((100, 128), '|b1', 1600))
     assert test("L") == ((100, 128), "|u1", 12800)
 
@@ -26,6 +30,9 @@ def test_toarray():
     assert test("RGB") == ((100, 128, 3), "|u1", 38400)
     assert test("RGBA") == ((100, 128, 4), "|u1", 51200)
     assert test("RGBX") == ((100, 128, 4), "|u1", 51200)
+
+    test_with_dtype(numpy.float)
+    test_with_dtype(numpy.uint8)
 
     with Image.open("Tests/images/truncated_jpeg.jpg") as im_truncated:
         with pytest.raises(OSError):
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/PIL/Image.py [hunk 1]
```diff
@@ -700,7 +700,10 @@ def __array__(self):
         class ArrayData:
             __array_interface__ = new
 
-        return np.array(ArrayData())
+        arr = np.array(ArrayData())
+        if dtype is not None:
+            arr = arr.astype(dtype)
+        return arr
```
radarhere, line 706:
```suggestion
        return np.array(ArrayData(), dtype)
```
This seems simpler to me.

t-vi:
To demystify what is needed by __array__, here are the two calls numpy does, one with dtype one without. (And in particular, you don't need to implement `numpy.array` 's signature.) 

https://github.com/numpy/numpy/blob/45c57078fd599194b49148fb66808f03490ee423/numpy/core/src/multiarray/ctors.c#L2453-L2458

This has been very stable and the reason the bug shows up in Pillow 8.3 is that it implemented only the version without the dtype passed in when switching to providing  `__array__` in https://github.com/python-pillow/Pillow/pull/5379  .

I looked at numpy's ndarray `__array__` for inspiration that providing a non-identical dtype casts/copies rather than failing.

Tests/test_image_array.py
```diff
@@ -27,6 +31,9 @@ def test(mode):
     assert test("RGBA") == ((100, 128, 4), "|u1", 51200)
     assert test("RGBX") == ((100, 128, 4), "|u1", 51200)
 
+    test_with_dtype(numpy.float)
```
radarhere, line 34:
```suggestion
    test_with_dtype(float)
```
Running the test with `numpy.float`, I get
> DeprecationWarning: `np.float` is a deprecated alias for the builtin `float`. To silence this warning, use `float` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.float64` here.

t-vi: We could use numpy.float64 if that suppresses the warning, too.
The reason I would hesitate to use float is that I would have to research what the minimum version of numpy is that supports this.
But I'll follow your advice if you say that PIL is happy to just be compatible with recent numpy.

t-vi:
I'm not entirely sure what to do with the codecov. It seems implausible to me that the line it points out in the Image.__array__ isn't covered by tests and I don't know what to think about the idea of tests being covered by tests.

Tests/test_image_array.py
hugovk: Checking with Python 3.6 (the lowest version we support) and NumPy 1.12.0 (the lowest to support Python 3.6), this script runs fine:

```python
import numpy as np

np.array([1, 2, 3], dtype=np.float)
np.array([1, 2, 3], dtype=float)
print("ok")
```

NumPy 1.12.0 was released Jan 16, 2017, so I guess that's old enough.

And checking NumPy download stats at [pepy.tech](https://pepy.tech/project/numpy?versions=*&versions=1.21.*&versions=1.12.*&versions=1.11.*), those older versions account for a tiny proportion:

image

radarhere:
I expect that was just because some CI jobs had an intermittent failure and were not returning their coverage results - everything looks fine with your latest commit.

t-vi:
OK, I leave the it to you to take this forward, then. Thank you for your suggestions.

hugovk (approved):

Issue #5546, opened 2021-06-18
`exif_transpose` overwrites the original image exif


artemisart:
### What did you do?

Use `ImageOps.exif_tranpose` on an image with an EXIF orientation code (not in [None, 0, 1]).

### What did you expect to happen?

`exif_tranpose()` should be immutable and returns a new image with the correct orientation.

### What actually happened?

`exif_tranpose()` returns a new image but also modifies the original image exif info in-place (deletes the orientation, key 0x112).
It's a pretty big problem if your work with rotated images because it breaks the original images, if you e.g. call again exif_tranpose() on it the EXIF orientation value is not there but the image content is not correctly rotated.

### What are your OS, Python and Pillow versions?

* OS: ubuntu 18.04
* Python: 3.8.8
* Pillow: 8.2.0

### Code

```python
import PIL, PIL.ImageOps
# if you have an image with exif
pil = PIL.Image.open(file)
# otherwise
import requests
resp = requests.get('https://raw.githubusercontent.com/recurser/exif-orientation-examples/master/Portrait_3.jpg', stream=True)
resp.raw.decode_content = True
pil = PIL.Image.open(resp.raw)

print(pil.getexif().get(0x0112))  # 3
correct_image = PIL.ImageOps.exif_transpose(pil)
print(pil.getexif().get(0x0112))  # None, but should be 3
```

### Proposed fix

Replace the first line of exif_tranpose in https://pillow.readthedocs.io/en/stable/_modules/PIL/ImageOps.html#exif_transpose with
```diff
from copy import deepcopy  # copy may be enough, I didn't try it

-    exif = image.getexif()
+    exif = deepcopy(image.getexif())
```

radarhere:
Yep, good point. I've created PR #5547 to resolve this - my method was not to copy the `Image.Exif` instance of the original image, but to create a new instance from the output image.

PR #5574, opened 2021-07-02
Remove `image.copy()` op from `exif_transpose()`

glenn-jocher:
This version of the function avoids image copy and a dictionary build for the majority use case of missing exif tags / no action needed on call.

As implemented in YOLOv5 in https://github.com/ultralytics/yolov5/pull/3852

### DIFF
--- a/src/PIL/ImageOps.py
+++ b/src/PIL/ImageOps.py
@@ -568,27 +568,24 @@ def solarize(image, threshold=128):
 
 def exif_transpose(image):
     """
-    If an image has an EXIF Orientation tag, return a new image that is
-    transposed accordingly. Otherwise, return a copy of the image.
+    Transpose an image accordingly if it has an EXIF Orientation tag
 
     :param image: The image to transpose.
     :return: An image.
     """
     exif = image.getexif()
     orientation = exif.get(0x0112)
-    method = {
-        2: Image.FLIP_LEFT_RIGHT,
-        3: Image.ROTATE_180,
-        4: Image.FLIP_TOP_BOTTOM,
-        5: Image.TRANSPOSE,
-        6: Image.ROTATE_270,
-        7: Image.TRANSVERSE,
-        8: Image.ROTATE_90,
-    }.get(orientation)
-    if method is not None:
-        transposed_image = image.transpose(method)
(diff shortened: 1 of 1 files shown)


radarhere:
I count 3 changes here

1. Moving `method` inside a new `if` block.

It seems to me like the extra code complexity isn't worth the very minor speed increase.

2. You reverted #5547, deleting the orientation tag from the original image.

As #5546 explains in full, this means that while original image hasn't been transposed, its orientation tag is gone now.

```python
from PIL import Image
im = Image.open("hopper.jpg")
exif = im.getexif()
orientation_before = exif.get(0x0112)
im.exif_transpose()
orientation_after = exif.get(0x0112)
```

The `orientation_after` will be empty, but the image data in `im` is still the same. I don't think this is desired.

3. Removing the `copy()` operation if a transpose is not performed.

Because `transpose` returns a copy of the image, this keeps the output consistent. Without it,
```python
from PIL import Image
im = Image.open("hopper.jpg")
im2 = im.exif_transpose()
```
sometimes, `im` will be the same object as `im2`, and sometimes it won't. That's a source of confusion for the user.

For example, afterwards, if I rotate `im2`,
```python
im2.rotate(90)
```
I don't know whether or not `im` has been rotated, because it might the same image instance, or it might not.

glenn-jocher:
CI is failing. I think this appears to be due to the CI itself expecting copies of images returned from the function rather than a problem in the function.

radarhere:
The specific failure seen in AppVeyor at the moment isn't because this isn't `copy()`ing the image. It's a test from #5547. So it's because this is deleting the orientation tag from the original image's EXIF data - what I'm calling point 2 in these changes.

glenn-jocher:
@radarhere thanks for the detailed analysis. I understand the need for continuity of behavior within the PIL package, so in this case it seems we want slightly different behaviors, as in our case we want to prioritize speed, so to use the official function we'd probably want to have an if-statement outside the function to determine if it required transposing first, which would duplicate the code inside the function.

I think probably we'll implement our updated version in YOLOv5 with a comment referencing the source, and then I'll close this PR.

glenn-jocher:
@radarhere PR is closed. Thank you for your feedback!

PR #5588, opened 2021-07-07
Ensure TIFF RowsPerStrip is multiple of 8 for JPEG compression

### LINKED ISSUE #5586: Recent change in TIFF rows per strip break saving with JPEG compression
### What did you do?

Tried to save an image with TIFF JPEG compression.

It to create a file.

### What actually happened?

It threw `OSError: encoder error -2 when writing image file`.  More specifically, I see `JPEGSetupEncode: RowsPerStrip must be multiple of 8 for JPEG.`.

* OS: Ubuntu 20.04
* Python: 3.9
* Pillow: 8.3.0

```python
import PIL.Image
image = PIL.Image.new('RGB', (1760, 1046))
image.save('/tmp/junk.tiff', compression='tiff_jpeg')
```

The error was specifically introduced in https://github.com/python-pillow/Pillow/pull/5514/files#diff-6ad43f85f1a075181d4d8cfcd97ae27bba1eccf5c3db5a3457160f98218eba6eR1562-R1565

kmilos:
Fixes #5586

### DIFF
--- a/src/PIL/TiffImagePlugin.py
+++ b/src/PIL/TiffImagePlugin.py
@@ -1577,6 +1577,9 @@ def _save(im, fp, filename):
     # aim for 64 KB strips when using libtiff writer
     if libtiff:
         rows_per_strip = min((2 ** 16 + stride - 1) // stride, im.size[1])
+        # JPEG encoder expects multiple of 8 rows
+        if compression == "jpeg":
+            rows_per_strip = min(((rows_per_strip + 7) // 8) * 8, im.size[1])
     else:
         rows_per_strip = im.size[1]
     strip_byte_counts = stride * rows_per_strip
--- a/Tests/test_file_libtiff.py
+++ b/Tests/test_file_libtiff.py
@@ -968,10 +968,11 @@ class TestFileLibTiff(LibTiffTestCase):
             assert str(e.value) == "-9"
         TiffImagePlugin.READ_LIBTIFF = False
 
-    def test_save_multistrip(self, tmp_path):
+    @pytest.mark.parametrize("compression", ("tiff_adobe_deflate", "jpeg"))
+    def test_save_multistrip(self, compression, tmp_path):
         im = hopper("RGB").resize((256, 256))
         out = str(tmp_path / "temp.tif")
-        im.save(out, compression="tiff_adobe_deflate")
+        im.save(out, compression=compression)
 
         with Image.open(out) as im:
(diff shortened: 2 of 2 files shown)


kmilos:
This works only for 4:4:4 data (no subsampling). AFAICT, we don't support subsampling directly when saving to TIFF, but the tag might be copied over from another info structure, so might be an idea to handle it?

kmilos:
Thanks, that leaves YCbCr mode TIFF saving which it seems never worked w/ JPEG compression?

src/PIL/TiffImagePlugin.py
```diff
@@ -1577,6 +1577,9 @@ def _save(im, fp, filename):
     # aim for 64 KB strips when using libtiff writer
     if libtiff:
         rows_per_strip = min((2 ** 16 + stride - 1) // stride, im.size[1])
+        # JPEG encoder expects multiple of 8 rows
+        if compression == "jpeg":
+            rows_per_strip = min(((rows_per_strip + 7) // 8) * 8, im.size[1])
```
radarhere, line 1582:
```suggestion
            rows_per_strip = min(((rows_per_strip + 7) // 8) * 8, rows_per_strip)
```
This seems better to me.

radarhere, line 1582:
```suggestion
            rows_per_strip = ((rows_per_strip + 7) // 8) * 8
```
Because `rows_per_strip` has already been set to be `im.size[1]` at minimum, and the calculation here only ever rounds up, the `min` part is unnecessary.

kmilos: I believe it is needed, precisely because it rounds up. This is to support the rare case of small images where im.size[1]<8 and you end up as a single strip. Libtiff/jpeg codec have no problem handling the single/last strip that is not factor of 8, as it is padded internally.

radarhere: Ok, thanks

radarhere:
```suggestion
            rows_per_strip = min(((rows_per_strip + 7) // 8) * 8, rows_per_strip)
```
I think this is clearer?

kmilos: That won't work for the corner case where you have u multi-strip w/ e.g. rows_per_strip=2 and im.size[1]=6. For JPEG, you want to end up as single strip of 6 rows.

Issue #4668, opened 2020-06-05
PIL convert('ycbcr') gives different result from formula


BarCodeReader:
### What did you do?
compare PIL convert('ycbcr') function with formula listed on wikipedia
### What did you expect to happen?
they should have the same result
### What actually happened?
they have different result
### What are your OS, Python and Pillow versions?

* OS: Windows 10
* Python: 3.7.4
* Pillow: 7.0.0

```python
# The formula of RGB to YCbCr is from Wiki: https://en.wikipedia.org/wiki/YCbCr
from PIL import Image
import numpy as np

# use PIL library
img = Image.open('any_image_you_have_in_hand.png').convert('RGB')
img, cb, cr = img.convert('YCbCr').split()
img = np.array(img).astype(np.float32) / 255.
print(img[:48,:48])

# use formula
def rgb_to_y(img_array):
    # formula is from wikipedia
    R,G,B = img_array[0], img_array[1], img_array[2]
    return 16. + (65.481 * R + 128.553 * G + 24.966 * B) / 255.

img2 = Image.open('any_image_you_have_in_hand.png').convert('RGB')
img2 = np.array(img2).transpose((2, 0, 1)).astype(np.float32) # c,h,w
img2 = rgb_to_y(img2) / 255.
print(img2[:48,:48])
```
different results will be generated...and I have no idea which one to use...

radarhere:
Here is the formula that Pillow advertises as the one it uses for this conversion.

https://github.com/python-pillow/Pillow/blob/61a35f94cf8a217db3e67d32db943b05e593e781/src/libImaging/ConvertYCbCr.c#L17-L21

This is also listed on that wikipedia page, under [JPEG Conversion](https://en.wikipedia.org/wiki/YCbCr#JPEG_conversion).

BarCodeReader:
> Here is the formula that Pillow advertises as the one it uses for this conversion.
> 
> https://github.com/python-pillow/Pillow/blob/61a35f94cf8a217db3e67d32db943b05e593e781/src/libImaging/ConvertYCbCr.c#L17-L21
> 
> This is also listed on that wikipedia page, under [JPEG Conversion](https://en.wikipedia.org/wiki/YCbCr#JPEG_conversion).

Thanks for this point.
Actually in wikipedia we have different formular, may I know why PIL choose this instead of the formula I used in my code? Is there a reason or I missed something?

radarhere:
Just to show that Pillow is fully aware of this -
[...]

PR #5597, opened 2021-07-09
Added tags when saving YCbCr TIFF

### LINKED ISSUE #5589: YCbCr mode TIFF saving not functional
This is w/ Pillow 8.3.1.

1. Uncompressed saving (i.e. not using libtiff_encode) doesn't set tag 530 to (1,1), and mandatory tag 532:

```python
import PIL.Image
image = PIL.Image.new('YCbCr', (1760, 1046))
image.save('test_ycbcr_nocomp.tif')
```

  Resulting file is invalid - compare to tags in tiff_strip_ycbcr_jpeg_1x1_sampling.tif

2. When using compression and/or libtiff_encode, even if added somehow, tag 530 doesn't get stored correctly and encoding fails:

```python
import PIL.Image
image = PIL.Image.new('YCbCr', (1760, 1046))
image.save('test_ycbcr_libtiff.tif', compression='jpeg')
```

See also https://github.com/python-pillow/Pillow/pull/5588#issuecomment-875490034

radarhere:
Resolves #5589

Ensures that tags 530 and 532 are present when saving a YCbCr TIFF.

### DIFF
--- a/src/PIL/TiffImagePlugin.py
+++ b/src/PIL/TiffImagePlugin.py
@@ -93,6 +93,7 @@ SUBIFD = 330
 EXTRASAMPLES = 338
 SAMPLEFORMAT = 339
 JPEGTABLES = 347
+YCBCRSUBSAMPLING = 530
 REFERENCEBLACKWHITE = 532
 COPYRIGHT = 33432
 IPTC_NAA_CHUNK = 33723  # newsphoto properties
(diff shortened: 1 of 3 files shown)


src/PIL/TiffImagePlugin.py
```diff
@@ -1593,6 +1594,13 @@ def _save(im, fp, filename):
     # no compression by default:
     ifd[COMPRESSION] = COMPRESSION_INFO_REV.get(compression, 1)
 
+    if im.mode == "YCbCr" and not libtiff:
+        for tag, value in {
+            YCBCRSUBSAMPLING: (1, 1),
+            REFERENCEBLACKWHITE: [15, 235, 128, 240, 128, 240],
```
kmilos, line 1600: I think this should probably be by default the full range [0, 255, 128, 255, 128, 255], as that is what JPEG uses?

radarhere: I got the numbers the "CCIR Recommendation 601.1 headroom/footroom" from http://www.verypdf.com/document/tiff6/txtidx0009.htm

You think that the "no headroom/footroom" option is better though?

kmilos: Yes, 601 is for video really, while JPEG uses full range. See also https://github.com/python-pillow/Pillow/issues/4668

kmilos: So there is of course the (valid?) assumption that we're almost always dealing w/ full range image data.

radarhere: Thanks. I've updated the commit

PR #5609, opened 2021-07-13
Fixed error saving APNG with duplicate frames and different duration times

### LINKED ISSUE #5608: type list doesn't define __round__ method
### What did you do?

I saved .apng file

Running with no exception

### What actually happened?

I got "type list doesn't define \_\_round\_\_ method" error in function "_write_multiple_frames"
https://github.com/python-pillow/Pillow/blob/83c05aaf8daa930086257eb38b254f06808d13e5/src/PIL/PngImagePlugin.py#L1060

https://github.com/python-pillow/Pillow/blob/83c05aaf8daa930086257eb38b254f06808d13e5/src/PIL/PngImagePlugin.py#L1152

* OS: MacOS
* Python: 3.8.2
* Pillow: 8.3.1

thak1411:
Resolves #5608 

This issue is caused by the 'duration' variable being overwritten.

https://github.com/python-pillow/Pillow/blob/83c05aaf8daa930086257eb38b254f06808d13e5/src/PIL/PngImagePlugin.py#L1120

After being overwritten, the duration variable is not updated.

https://github.com/python-pillow/Pillow/blob/83c05aaf8daa930086257eb38b254f06808d13e5/src/PIL/PngImagePlugin.py#L1083-L1084

### DIFF
--- a/Tests/test_file_apng.py
+++ b/Tests/test_file_apng.py
@@ -433,7 +433,9 @@ def test_apng_save_duration_loop(tmp_path):
 
     # test removal of duplicated frames
     frame = Image.new("RGBA", (128, 64), (255, 0, 0, 255))
-    frame.save(test_file, save_all=True, append_images=[frame], duration=[500, 250])
+    frame.save(
+        test_file, save_all=True, append_images=[frame, frame], duration=[500, 100, 150]
+    )
     with Image.open(test_file) as im:
         im.load()
         assert im.n_frames == 1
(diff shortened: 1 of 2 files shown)


dogemad:
Yes. There is a problem in that the 'duration', which should be transmitted as an integer to the next layer by redefining the variable names in lines 1062 and 1120 to be the same, is transmitted as a list.

radarhere:
I've created https://github.com/thak1411/Pillow/pull/1 with some suggestions.

thak1411:
thanks @dogemad @radarhere !

PR #5613, opened 2021-07-15
PSD layer count may be negative

### LINKED ISSUE #5612: Pillow fails/hangs on PSD hidden background layer
Depending on file content, Pillow:

- fails with "cannot identify image file";
- fails with "Truncated File Read";
- just hangs inside the `PsdImagePlugin._layerinfo` function because it thinks there are 65534 layers.

See attachment: [pillow-psd-fail.zip](https://github.com/python-pillow/Pillow/files/6815914/pillow-psd-fail.zip)

GIMP works fine with these files.

radarhere:
Resolves #5612

https://www.adobe.com/devnet-apps/photoshop/fileformatashtml/#50577409_pgfId-1031423
> Layer count. If it is a negative number, its absolute value is the number of layers and the first alpha channel contains the transparency data for the merged result.

But at the moment, we're reading it in as H, [unsigned short](https://docs.python.org/3/library/struct.html#format-characters).

https://github.com/python-pillow/Pillow/blob/696b82e44032500197734c4dc4669d7c3f8b9371/src/PIL/_binary.py#L70-L71

https://github.com/python-pillow/Pillow/blob/696b82e44032500197734c4dc4669d7c3f8b9371/src/PIL/PsdImagePlugin.py#L23

https://github.com/python-pillow/Pillow/blob/696b82e44032500197734c4dc4669d7c3f8b9371/src/PIL/PsdImagePlugin.py#L182

This PR changes it to signed.


hugovk:
Thanks!

PR #5618, opened 2021-07-18
Added WalImageFile class

radarhere:
Resolves FIXME

https://github.com/python-pillow/Pillow/blob/696b82e44032500197734c4dc4669d7c3f8b9371/src/PIL/WalImageFile.py#L42-L43


hugovk (approved):
Let's include this in the release notes.

PR #5620, opened 2021-07-18
Added "transparency" argument to EpsImagePlugin load()

### LINKED ISSUE #5619: EPS files are loaded without transparency
EPS files can have transparency, and GhostScript have the option to keep it. However, if we use Pillow to open an EPS-file, then the alpha-channel is discarded. To my understanding, this is because Pillow is currently using the `ppm` (portable pixel map)-driver, which creates RGB-images, not RGBA-images. 

I therefore propose to add an option to specify which driver to use with GhostScript (or add a flag that can be enabled to use the `pngalpha` driver instead of the `ppmraw` driver)

To my understanding, this can be fixed by modifying line 120 in `src/PIL/EpsImagePlugin.py`: https://github.com/python-pillow/Pillow/blob/20b8a8377372b1f318ae275acd5e11bad4722e2e/src/PIL/EpsImagePlugin.py#L120
into
```python
        "-sDEVICE=pngalpha",  # png driver
```

I made the change locally on my machine, and it seemed to work well. 

* OS: Windows 10
* Python: 3.8
* Pillow: 8.2.0

### Endnote: Script to generate a postscript file and convert it with Pillow
```python
import turtle
from PIL import Image

pen = turtle.Turtle()
for side in range(4):
    pen.forward(50)
    pen.right(90)

turtle.getscreen().getcanvas().postscript(file="square.eps")
img = Image.open('square.eps')
img.save("image.png")
```

radarhere:
Resolves #5619

At the moment, EpsImagePlugin can load images at different scales like so
https://github.com/python-pillow/Pillow/blob/426b730d3b0810fca59d4e0bcbc670916b890398/Tests/test_file_eps.py#L48-L49

This PR adds another argument, "transparency",
```python
with Image.open(FILE1) as image12: 
     image1.load(transparency=True)
```
to switch from "ppmraw"
https://github.com/python-pillow/Pillow/blob/426b730d3b0810fca59d4e0bcbc670916b890398/src/PIL/EpsImagePlugin.py#L120
to "pngalpha", generating an RGBA image with a transparent background, instead of an RGB image with a white background.


hugovk (approved):
Let's also add this to release notes.

PR #5621, opened 2021-07-19
Accept any path-like object on `Image.{open,save}` and `truetype`, not just `pathlib.Path`

ntninja:
Also tries to fix some bugs where PIL will crash in case of binary paths, by using `os.fsdecode`.

 * Accept any path-like object (as introduced with Python 3.6) where it would previously only accept `str`, `bytes` and `pathlib.Path`
 * Fix some crashes when using binary paths found during testing/review
 * Consistently use the word “filepath” when refering to a filesystem location, rather than the previous mix of the words “file” and “filename”


radarhere:
> Fix some crashes when using binary paths found during testing/review

Are you saying that you've found bugs, separate to the path-like object changes? Could you provide some more detail about this?

Tests/test_image.py
```diff
@@ -138,21 +154,26 @@ def test_stringio(self):
...
-        with Image.open(Path("Tests/images/multipage-mmap.tiff")) as im:
+    @pytest.mark.parametrize("PathCls", (
+        pathlib.Path,
+        MyStrPathLike,  # Returns `str` on `os.fspath`
+        MyBytesPathLike,  # Returns `bytes` on `os.fspath`
+        str,
+        os.fsencode,  # Converts path to `bytes`
+    ))
+    def test_path_like(self, tmp_path, PathCls):
+        with Image.open(PathCls("Tests/images/multipage-mmap.tiff")) as im:
```
ntninja, line 165: @radarhere: This test when using `MyBytesPathLike` or `os.fsencode` (ie `bytes`) as `PathCls` causes…

src/PIL/Image.py
```diff
@@ -2204,7 +2201,7 @@ def save(self, fp, format=None, **params):
 
         preinit()
 
-        ext = os.path.splitext(filename)[1].lower()
+        ext = os.fsdecode(os.path.splitext(filename)[1].lower())
```
ntninja, line 2204: … the extension format lookup to fail here (b/c all dict items are strings), thereby raising `ValueError`.

(Fix is to ensure that the extension is always a string as above.)

src/PIL/ImageFont.py
```diff
@@ -856,7 +857,7 @@ def freetype(font):
     except OSError:
         if not isPath(font):
             raise
-        ttf_filename = os.path.basename(font)
+        ttf_filename = os.fsdecode(os.path.basename(font))
```
ntninja, line 860: Similarily, this lookup code also obviously expects `str` paths for everything!

ntninja:
@radarhere: Any idea why that Windows test is failing?

radarhere:
It's an intermittent failure on master.  Nothing to do with your changes.
I've rerun the job, and it passes now.

Tests/test_util.py
```diff
@@ -39,6 +54,22 @@ def test_is_not_path(tmp_path):
...
+    class PathAndFileThingy(io.TextIOWrapper):
+        def __fspath__(self):
+            return self.name
+
+    with PathAndFileThingy(open(tmp_path / "temp.ext", "wb")) as test_obj:
+        # Act
+        it_is_not = _util.isPath(test_obj)
+
+        # Assert
+        assert not it_is_not
```
radarhere, line 70:
```suggestion
    with open(tmp_path / "temp.ext", "wb") as fp:
		with PathAndFileThingy(fp) as test_obj:
			# Act
			it_is_not = _util.isPath(test_obj)

			# Assert
			assert not it_is_not
```
Adds a context manager for the file pointer.

ntninja: `io.TextIOBase` takes ownership of the provided binary I/O object, so it will be automatically closed when closing the “PathAndFileThingy” at the end of the with-block.

src/PIL/ImageFont.py
radarhere: Could you add tests for these ImageFont changes?

ntninja: Sorry, but I don't even know what this stuff is for exactly (don't have Windows at all). I just fixed it because it is obviously affected by the same class of bug as pointed out by the previous comment. If that is a problem, I can revert it through.

src/PIL/ImageFont.py
```diff
@@ -199,6 +199,7 @@ def load_from_bytes(f):
 
         if isPath(font):
             if sys.platform == "win32":
+                font = os.fspath(font)
```
radarhere, line 202: I've created https://github.com/ntninja/Pillow/pull/1 to add a test for this.

src/PIL/ImageFont.py
radarhere: I think this change isn't necessary. I think that `basename` already converts it.
```pycon
>>> from pathlib import Path
>>> import os
>>> filename = os.path.basename(Path("a/b/c"))
>>> filename
'c'
>>> type(filename)
<class 'str'>
```

ntninja: No it is still needed for binary paths (see the added tests):

```py
>>> os.path.basename(b"abcdef/def")
b'def'
>>> os.fsdecode(os.path.basename(b"abcdef/def"))
'def'
```

src/PIL/Image.py
```diff
@@ -2150,22 +2150,22 @@ def transform(x, y, matrix):
 
     def save(self, fp, format=None, **params):
         """
-        Saves this image under the given filename.  If no format is
-        specified, the format to use is determined from the filename
-        extension, if possible.
+        Saves this image under the given filepath.  If no format is
+        specified, the format to use is determined from the filepath's
+        file extension, if possible.
```
hugovk, line 2155:
```suggestion
        Saves this image under the given file path. If no format is
        specified, the format to use is determined from the path's
        file extension, if possible.
```

src/PIL/Image.py
```diff
@@ -2150,22 +2150,22 @@ def transform(x, y, matrix):
...
 
         Keyword options can be used to provide additional instructions
         to the writer. If a writer doesn't recognise an option, it is
         silently ignored. The available options are described in the
         :doc:`image format documentation
         <../handbook/image-file-formats>` for each writer.
 
-        You can use a file object instead of a filename. In this case,
+        You can use a file object instead of a filepath. In this case,
```
hugovk, line 2163:
```suggestion
        You can use a file object instead of a file path. In this case,
```

src/PIL/Image.py
```diff
@@ -2150,22 +2150,22 @@ def transform(x, y, matrix):
...
         you must always specify the format. The file object must
         implement the ``seek``, ``tell``, and ``write``
         methods, and be opened in binary mode.
 
-        :param fp: A filename (string), pathlib.Path object or file object.
+        :param fp: A filepath (str or bytes), path-like object or file object.
```
hugovk, line 2168:
```suggestion
        :param fp: A file path (str or bytes), path-like object or file object.
```

src/PIL/Image.py
```diff
@@ -2923,7 +2920,7 @@ def open(fp, mode="r", formats=None):
     :py:meth:`~PIL.Image.Image.load` method).  See
     :py:func:`~PIL.Image.new`. See :ref:`file-handling`.
 
-    :param fp: A filename (string), pathlib.Path object or a file object.
+    :param fp: A filepath (str or bytes), path-like object or a file object.
```
hugovk, line 2923:
```suggestion
    :param fp: A file path (str or bytes), path-like object or a file object.
```

src/PIL/ImageFont.py
```diff
@@ -796,10 +797,10 @@ def load(filename):
 
 
 def truetype(font=None, size=10, index=0, encoding="", layout_engine=None):
-    """
-    Load a TrueType or OpenType font from a file or file-like object,
+    r"""
+    Load a TrueType or OpenType font from a filepath or file-like object,
```
hugovk, line 801:
```suggestion
    Load a TrueType or OpenType font from a file path or file-like object,
```

src/PIL/ImageFont.py
```diff
@@ -796,10 +797,10 @@ def load(filename):
...
     and create a font object.
-    This function loads a font object from the given file or file-like
+    This function loads a font object from the given filepath or file-like
```
hugovk, line 803:
```suggestion
    This function loads a font object from the given file path or file-like
```

src/PIL/ImageFont.py
```diff
@@ -809,12 +810,12 @@ def truetype(font=None, size=10, index=0, encoding="", layout_engine=None):
 
     This function requires the _imagingft service.
 
-    :param font: A filename or file-like object containing a TrueType font.
-                 If the file is not found in this filename, the loader may also
-                 search in other directories, such as the :file:`fonts/`
+    :param font: A filepath or file-like object containing a TrueType font.
+                 If the file is not found at this filepath, the loader may also
```
hugovk, line 814:
```suggestion
    :param font: A file path or file-like object containing a TrueType font.
                 If the file is not found at this path, the loader may also
```

hugovk (commented):
Docs suggestion: use "file path" or "path" instead of "filepath" which isn't used much:

* https://pillow.readthedocs.io/en/stable/search.html?q=filepath&check_keywords=yes&area=default
* https://docs.python.org/3/search.html?q=filepath&check_keywords=yes&area=default

radarhere:
I've made a suggestion for this over at https://github.com/ntninja/Pillow/pull/2

ntninja:
> I've made a suggestion for this over at [ntninja#2](https://github.com/ntninja/Pillow/pull/2)

I think we've resolved this now. Can we merge this then?

PR #5552, opened 2021-06-23
Improved ImagePalette

radarhere:
This PR has several changes
- the ImagePalette `colors` dictionary is currently empty when the object is created, rather than being filled with palette entries. This fixes that
- it changes the palette to be empty by default. This means that new colors can be appended more easily, rather than finding that all 256 entries are already occupied
- if a new color is added when 256 entries are taken, search through the image. If there is a palette index that isn't actually used, use that for the new color
- helps #5375 by determining the border palette index on the new image only after the palette has been attached

Edit: Thanks to later commits, resolves #2803

PR #5641, opened 2021-07-28
Deprecate ImagePalette size parameter

### LINKED ISSUE #5636: Incorrect documentation for ImagePalette size parameter (and maybe not needed at all)
The documentation for the `ImagePalette` initializer in version 8.3.1 says the `palette` parameter must be "of length `size` times the number of colors in `mode`". Therefore, for an RGB image, I would expect `len(palette) == size * 3`. However, the code asserts that `len(palette) == size`, so I believe the code and documentation are inconsistent. (The same problem existed in 8.2.0 before some ImagePalette improvements were made, so this wasn't introduced with that change.)

Furthermore, it isn't clear to me that the `size` parameter is needed at all. It isn't stored on `self`, and the only place it's used in the initializer is to assert that its value is `0` or `len(palette)`, so it doesn't seem to provide any benefit. The only reason to keep it that I can think of is to maintain backwards compatibility with existing code that explicitly passes the parameter.

radarhere:
Resolves #5636

The `size` parameter was originally added in #537. Before #5552, it was there to override a default length check.

https://github.com/python-pillow/Pillow/blob/10615a7da7a5254de9696f4f3f14d229098b67cc/src/PIL/ImagePalette.py#L45-L48

Afterwards though, the default length check was gone, so `size` is now just adding an arbitrary constraint on the user - if provided, it throws an error if `palette` parameter is not the length of the `size` parmeter.

https://github.com/python-pillow/Pillow/blob/3462b5fd74e5e86539e7fa639adf4a52d53bf57c/src/PIL/ImagePalette.py#L43-L44

### DIFF
--- a/docs/deprecations.rst
+++ b/docs/deprecations.rst
@@ -92,6 +92,17 @@ dictionary. The :py:attr:`~PIL.JpegImagePlugin.convert_dict_qtables` method no l
 performs any operations on the data given to it, has been deprecated and will be
 removed in Pillow 10.0.0 (2023-01-02).
 
+ImagePalette size parameter
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. deprecated:: 8.4.0
+
+The ``size`` parameter will be removed in Pillow 10.0.0 (2023-01-02).
+
+Before Pillow 8.3.0, ImagePalette required palette data of particular lengths by
+default, and the size parameter could be used to override that. Pillow 8.3.0 removed
+the default required length, also removing the need for the size parameter.
+
 Removed features
 ----------------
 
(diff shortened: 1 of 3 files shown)


docs/deprecations.rst
```diff
@@ -92,6 +92,17 @@ dictionary. The :py:attr:`~PIL.JpegImagePlugin.convert_dict_qtables` method no l
...
 removed in Pillow 10.0.0 (2023-01-02).
 
+ImagePalette size parameter
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. deprecated:: 8.4.0
+
+The ``size`` parameter will be removed in Pillow 10.0.0 (2023-01-02).
+
+Before Pillow 8.3.0, ImagePalette required palette data of particular lengths by
```
hugovk, line 102:
```suggestion
Before Pillow 8.3.0, ``ImagePalette`` required palette data of particular lengths by
```

hugovk (approved):

PR #5646, opened 2021-07-30
Speed up rotating square images by 90 or 270 degrees

radarhere:
`rotate()` may use `transpose()` for a faster result. For 90 or 270 degrees, it makes sure that the `expand` is true, since rotating an image sideways may change the size... except if the image is a square, so I've created this PR.

The test does not fail without this change, since this is just a speed improvement.

https://github.com/python-pillow/Pillow/blob/6cb127c8cdb95cfc5f8e34e6cc686b565526a933/src/PIL/Image.py#L2070-L2080


hugovk:
What sort of speed increase does this give?

radarhere:
```python
import timeit
setup = '''
from PIL import Image
'''
print(timeit.Timer('im.rotate(90)', setup=setup).repeat(number=10000))

# Without change: [0.23164604200000002, 0.23040333300000004, 0.22858733399999998, 0.22947383300000002, 0.228850459]
# With change: [0.10277529199999999, 0.10330620800000001, 0.10458945799999997, 0.10447408300000005, 0.10545916600000005]
```

So it makes it twice as fast.

homm:
It will be much more significant for larger images.

hugovk:
Thanks, this is worth including in the release notes as well.

PR #5651, opened 2021-08-03
Hide FriBiDi shim symbols to avoid conflict with real FriBiDi library

### LINKED ISSUE #5637: Segmentation fault on import when gi is imported
### What did you do?
```python
from gi.repository import Gtk
from PIL import ImageOps
```
No errors
### What actually happened?
```
[1]    85601 segmentation fault (core dumped)  python azote/main.py
```

* OS: Arch Linux
* Python: 3.9.6
* Pillow: 8.3.1

EDIT: shorter code causing segfault

nulano:
Fixes #5637

In #5637, when Gtk is imported, the FriBiDi library gets loaded which replaces the shim symbols due to dynamic linking. Trying to load `_imagingft` then calls `load_fribidi` which attempts to overwrite its symbols and segfaults. Hiding the symbols prevents them from being replaced by the real library and segfaulting. Hiding is the default with MSVC.

### DIFF
--- a/src/thirdparty/fribidi-shim/fribidi.c
+++ b/src/thirdparty/fribidi-shim/fribidi.c
@@ -12,7 +12,7 @@
 
 
 /* FriBiDi>=1.0.0 adds bracket_types param, ignore and call legacy function */
-FriBidiLevel fribidi_get_par_embedding_levels_ex_compat(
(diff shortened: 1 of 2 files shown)


wyatt8740:
Just want to state here as well that this PR fixes the fribidi-related segfaults that I was having with Pillow 8.3.1.

hugovk (approved):

PR #5655, opened 2021-08-05
Allow saving 1 and L mode TIFF with PhotometricInterpretation 0

### LINKED ISSUE #4804: Tag 262 (PhotometricInterpretation) is changed when saving as a TIFF
### What did you do?
save an uncompressed tiff file to group 4 compressed tiff file
I expected to get a group 4 compressed tiff file and kept all other tags the same as original file
### What actually happened?
I got a group 4 compressed tiff file, but Tag 262 (PhotometricInterpretation) was changed from 0 to 1

* OS: Window 10
* Python: Python 3.6.8
* Pillow:  7.2.0

```python
from PIL import Image

img = Image.open('uncompressed.tiff')
print(img.tag)
img.save('group4.tiff', compression="group4", tiffinfo=img.tag)
img = Image.open('group4.tiff')
print(img.tag)
```

[pillowIssue.zip](https://github.com/python-pillow/Pillow/files/4949280/pillowIssue.zip)

radarhere:
Resolves #4804

TIFF tag 262 is [PhotometricInterpretation](https://www.awaresystems.be/imaging/tiff/tifftags/photometricinterpretation.html). Pillow currently does not allow this value to be set when saving. Changing the value for mode 1 images means that white becomes black and vice versa, so to keep the output the same, we need to invert the pixel values.

This PR allows for that specific scenario, inverting the values when saving.

### DIFF
--- a/src/PIL/TiffImagePlugin.py
+++ b/src/PIL/TiffImagePlugin.py
@@ -48,7 +48,7 @@ from collections.abc import MutableMapping
 from fractions import Fraction
 from numbers import Number, Rational
 
-from . import Image, ImageFile, ImagePalette, TiffTags
+from . import Image, ImageFile, ImageOps, ImagePalette, TiffTags
 from ._binary import o8
 from .TiffTags import TYPES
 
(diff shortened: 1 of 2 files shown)


src/PIL/TiffImagePlugin.py
```diff
@@ -1568,7 +1570,15 @@ def _save(im, fp, filename):
...
 
-    ifd[PHOTOMETRIC_INTERPRETATION] = photo
+    if PHOTOMETRIC_INTERPRETATION not in ifd:
+        ifd[PHOTOMETRIC_INTERPRETATION] = photo
+    elif im.mode == "1" and ifd[PHOTOMETRIC_INTERPRETATION] == 0:
+        inverted_im = im.copy()
+        px = inverted_im.load()
+        for y in range(inverted_im.height):
+            for x in range(inverted_im.width):
+                px[x, y] = 0 if px[x, y] == 255 else 255
```
kmilos, line 1580: This only handles bilevel, but grayscale (mode "L") images can also have WhiteIsZero interpretation, so might be an idea to cover them as well here (maybe use 255-px[x ,y] and also check for "L" above?)

kmilos (approved):
LGTM

PR #1365, opened 2015-07-31
Fixed PSDraw stdout Python 3 compatibility

radarhere:
If no file pointer is provided, PSDraw uses `sys.stdout`. This causes some bugs in Python 3.

To demonstrate, I've added a test.

PR #5665, opened 2021-08-09
Do not return in ImageFile when saving to stdout

### LINKED ISSUE #5657: Saving images to stdout using some file formats does not write the image correctly
### What did you do?

I created a image and saved it to stdout using PPM format.

The image is written correctly.

### What actually happened?

Only the PPM header is written, and the image data is not.

* OS: Gentoo Linux
* Python: Tested with both 3.8.11 and 3.9.6
* Pillow: 8.3.1

```python
import sys
from PIL import Image

im = Image.frombytes('RGB', (4, 4), b'\0' * 48)
im.save(sys.stdout, 'ppm')
```

infmagic2047:
Fixes #5657.

The removed block of code is bad for the following reasons:
1. The return statement makes the function write nothing when fp is stdout, which is obviously unwanted.
2. The fp.flush() statement is duplicate of the line below, and is therefore useless.

### DIFF
--- a/Tests/test_file_jpeg.py
+++ b/Tests/test_file_jpeg.py
@@ -1,5 +1,6 @@
 import os
 import re
+import sys
 from io import BytesIO
 
 import pytest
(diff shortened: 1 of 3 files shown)


radarhere:
When I try this, it doesn't fix the issue for me. Could you add a test to demonstrate?

infmagic2047:
Added tests for JPEG and PPM formats.

Tests/test_file_ppm.py
```diff
@@ -80,3 +83,30 @@ def test_mimetypes(tmp_path):
...
+    with Image.open(TEST_FILE) as im:
+        im.save(sys.stdout, "PPM")
+
+    # Reset stdout
+    sys.stdout = old_stdout
+
+    if buffer:
+        mystdout = mystdout.buffer
+    reloaded = Image.open(mystdout)
+    assert_image_equal_tofile(reloaded, TEST_FILE)
```
radarhere, line 112:
```suggestion
    with Image.open(mystdout) as reloaded:
        assert_image_equal_tofile(reloaded, TEST_FILE)
```
Added a context manager.

Tests/test_file_jpeg.py
```diff
@@ -870,6 +871,33 @@ def test_getxmp(self):
...
+            im.save(sys.stdout, "JPEG")
+            im_roundtrip = self.roundtrip(im)
+
+        # Reset stdout
+        sys.stdout = old_stdout
+
+        if buffer:
+            mystdout = mystdout.buffer
+        reloaded = Image.open(mystdout)
+        assert_image_equal(reloaded, im_roundtrip)
```
radarhere, line 899:
```suggestion
        with Image.open(mystdout) as reloaded:
            assert_image_equal(reloaded, im_roundtrip)
```

radarhere:
This is reverting part of #1365. That PR added `return` to prevent the encoder writing bytes to `sys.stdout`.

Looking back now, I think that was incorrect for this reason, that it did prevent the body of the image from being output.

PR #5670, opened 2021-08-12
pre-commit autoupdate --freeze

radarhere:


hugovk:
Whilst we're on the topic:

---

@aclark4life Please could you enable these repo with https://pre-commit.ci ?

https://github.com/python-pillow/Pillow
https://github.com/python-pillow/pillow-wheels

It's a free service for open source projects. It'll give us two benefits:

1. It will automatically update the linting config every so often. It can be set to weekly, monthly or quarterly. We don't need weekly, quarterly should be fine.

2. It checks PRs and will autofix them. For example, if someone hasn't formatted with Black, this will do it automatically, without us needing to explain to them how to install and run Black, or us checking out the branch and doing it for them.

---

@radarhere Please could add this to the end of `.pre-commit-config.yaml`?

```yaml

ci:
  autoupdate_schedule: quarterly
```

---

I'm not sure if https://pre-commit.ci/ supports freezing the `rev` (added in https://github.com/python-pillow/Pillow/pull/4670), but I don't recall seeing any other projects doing it. For example, pre-commit repos and pip don't freeze.

hugovk (approved):

PR #5683, opened 2021-08-23
ubuntu prerequisites are still true for ubuntu

andreasnuesslein:
Ubuntu 21.04 needs the same prerequisites.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -390,7 +390,7 @@ In Alpine, the command is::
 
 .. Note:: ``redhat-rpm-config`` is required on Fedora 23, but not earlier versions.
 
-Prerequisites for **Ubuntu 16.04 LTS - 20.04 LTS** are installed with::
+Prerequisites for **Ubuntu 16.04 LTS - 21.04 LTS** are installed with::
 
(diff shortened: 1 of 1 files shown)


hugovk:
Thanks, good to hear!


This change makes it look a bit like 21.04 is LTS, and I think we've only listed LTS here before, so maybe we'll keep it as is (until the next LTS is out)?

andreasnuesslein:
oh my bad, I didn't even look at the `LTS` bit. yes what you say makes sense :+1:

we could just put "Prerequisites for Ubuntu 16.04 and newer" or even just "Prerequisites for Ubuntu". At this point nobody ought to use an Ubuntu older than 16.04, or even 18.04 for that matter. If I was using Ubuntu from 5 years ago I would not be asuming that current docs are still relevant for me.

radarhere:
It has been suggested that it be "Ubuntu 16.04 LTS or later" - https://github.com/python-pillow/Pillow/pull/4596#discussion_r422623881. My two cents is that at some point in the future, libjpeg8-dev will stop being a valid package name. I also think it is fine as is.

andreasnuesslein:
okiedokie :+1: I'll report back in a year ;)

PR #5694, opened 2021-08-29
Added width argument to ImageDraw polygon

### LINKED ISSUE #5689: Add width to polygon
Feature request: add `width` as an option to method `polygon()`.

As you can see, `rectangle()` has the `width` as a parameter, and many other similar methods. Why can't `polygon()`?

radarhere:
Resolves #5689

If we just drew thicker lines when creating a polygon, then since the centre of the line is the edge, half of each line would be drawn outside the filled polygon.

This PR uses masking to trim the outer half.

Because the image might have any mode, and there are limited modes that can be used as a mask,
- I draw the filled polygon on a 1-bit image, `fill_im`
- I draw the outline of the polygon on a 1-bit image, `ink_im`, which extends outside the filled polygon
- I use `fill_im` as a mask to trim the polygon outline to just the inner part, giving a 1-bit `mask`.
- Then I create an image with the destination mode, draw the outline, and use `mask` to paste only the inner part on the original image.


hugovk:
Thanks! Let's add release notes too.

radarhere:
Sure. See #5847

PR #5696, opened 2021-08-30
Copy Python palette to new image in quantize()

### LINKED ISSUE #5695: PNG palette quantized save functionality different between 8.1.0 and 8.3.1
### What did you do?
Create a palette, quantize an image using this palette and save it.

The palette information is stored correctly. I.e. the list of RGB values is the same in the saved image.
E.g. [0, 0, 0, 209, 254, 76, 217, 3, 3, 10, 66, 192,...]

### What actually happened?
The list of RGB values is corrupted.
E.g. [0, 0, 0, 1, 1, 1, 2, 2, 2, 3, 3, 3,...]

* OS: Tested on Windows and Azure Databricks (Ubuntu)
* Python: 3.6 & 3.8
* Pillow: 8.3.1 has the issue

```python
from PIL import Image
from itertools import cycle
import numpy as np

palette = [0, 0, 0, 209, 254, 76, 217, 3, 3, 10, 66, 192]
# palette must contain 768 values (256 * 3)
# repeat the colours until we get there
color_iter = cycle(palette)
expanded_palette_format = []
while True:
    expanded_palette_format.append(next(color_iter))
    if len(expanded_palette_format) == 768:
        break

assert len(expanded_palette_format) == 768
palette_image = Image.new('P', (16, 16))
palette_image.putpalette(expanded_palette_format)

# a new image that we want to apply the palette to
im_to_palette = np.array(
    [
        [[0, 0, 0,],[209, 254, 76], [217, 3, 3], [10, 66, 192]],
        [[0, 0, 0,],[209, 254, 76], [217, 3, 3], [10, 66, 192]],
        [[0, 0, 0,],[209, 254, 76], [217, 3, 3], [10, 66, 192]],
        [[0, 0, 0,],[209, 254, 76], [217, 3, 3], [10, 66, 192]]
    ],
    np.uint8
    )
im = Image.fromarray(im_to_palette).quantize(palette=palette_image, dither=0)

im.save(r"test.png")
saved_im = Image.open(r"test.png")

assert im.getpalette() == saved_im.getpalette()
# AssertionError
```

Please note that if I specifically pass the 'GIF' argument on save, it has the expected behaviour.
```python
im.save(r"test.png", "GIF")
saved_im = Image.open(r"test.png")

assert im.getpalette() == saved_im.getpalette()
```

radarhere:
Resolves #5695

[quantize()](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.quantize) accepts `palette` as an argument, but only applies the palette to the C layer.

https://github.com/python-pillow/Pillow/blob/d50052a75c4d0d17d3d37c24a4558c9e0736f0e8/src/PIL/Image.py#L1123-L1133

This PR copies the palette into the Python layer as well.

### DIFF
--- a/Tests/test_image_quantize.py
+++ b/Tests/test_image_quantize.py
@@ -63,6 +63,7 @@ def test_quantize_no_dither():
 
     converted = image.quantize(dither=0, palette=palette)
     assert_image(converted, "P", converted.size)
+    assert converted.palette.palette == palette.palette.palette
 
 
 def test_quantize_dither_diff():
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -1130,7 +1130,9 @@ class Image:
                     "only RGB or L mode images can be quantized to a palette"
                 )
             im = self.im.convert("P", dither, palette.im)
-            return self._new(im)
+            new_im = self._new(im)
+            new_im.palette = palette.palette.copy()
+            return new_im
 
         im = self._new(self.im.quantize(colors, method, kmeans))
 


paulds8:
Looking at the change, I don't quite understand how adding the 'GIF' format to the save worked though. My expectation was that the issue would be somewhere in `save()` but if this works I'm happy.

radarhere:
When writing the PNG PLTE chunk, the palette data is being retrieved from the C layer (`.im`), but limited to the number of `colors`

https://github.com/python-pillow/Pillow/blob/d50052a75c4d0d17d3d37c24a4558c9e0736f0e8/src/PIL/PngImagePlugin.py#L1283-L1288

`colors` is set here (unless there is no palette or you're passing a "bits" argument, but those are not the case)

https://github.com/python-pillow/Pillow/blob/d50052a75c4d0d17d3d37c24a4558c9e0736f0e8/src/PIL/PngImagePlugin.py#L1206

This is determining the number of colors from the ImagePalette object in the Python layer

https://github.com/python-pillow/Pillow/blob/d50052a75c4d0d17d3d37c24a4558c9e0736f0e8/src/PIL/ImagePalette.py#L80-L89

So in this PR, I am making sure that ImagePalette is also correct, meaning that the number of `colors` is correct when saving to PNG.

Issue #5569, opened 2021-07-01
Add support for Python 3.10


hugovk:
Tracking issue for getting the pieces into place to support Python 3.10 in Pillow 8.4.0, scheduled for 2020-10-15.

* [x] Wait for 3.10.0, expected [Monday, 2021-10-04](https://www.python.org/dev/peps/pep-0619/)
* [x] Wait for 3.10.0 on GHA: https://github.com/actions/setup-python/issues/249
* [x] Wait for 3.10.0 on AppVeyor: https://github.com/appveyor/ci/issues/3741
* [ ] Wait for 3.10.0 on Travis CI (https://github.com/python-pillow/pillow-wheels): https://travis-ci.community/t/add-python-3-10/12220
* [x] ([#5748](https://github.com/python-pillow/Pillow/pull/5748)) Replace `3.10-dev` with `3.10` on GHA
* [x] ([#5805](https://github.com/python-pillow/Pillow/pull/5805)) Replace `3.9` with `3.10` on AppVeyor
* [x] ([#5570](https://github.com/python-pillow/Pillow/pull/5570)) Add Trove classifier, update `setup.py` version check, add to `tox.ini`, update [Python support table](https://pillow.readthedocs.io/en/stable/installation.html#python-support)
* [x] ([#5713](https://github.com/python-pillow/Pillow/pull/5713)) Update [CI targets table](https://pillow.readthedocs.io/en/latest/installation.html#continuous-integration-targets)
* [x] ([python-pillow/pillow-wheels#215](https://github.com/python-pillow/pillow-wheels/pull/215) [python-pillow/pillow-wheels#217](https://github.com/python-pillow/pillow-wheels/pull/217)) Update https://github.com/python-pillow/pillow-wheels

(3.9 issue for comparison: https://github.com/python-pillow/Pillow/issues/4953)

wiredfool:
I've seen this mentioned elsewhere and it makes sense -- Shipping a py 3.10 compatible/tagged wheel before the actual release can help with other libraries sorting their compatibility. It could reduce the support load here as well, as we wouldn't get the confusion from people attempting to compile for 3.10 in the window between it coming out and us (well, you) getting the new release out. This assumes that we can run the RC in the testing setup, and I'm not 100% clear that that's actually the case. 

In the unlikely event that the 3.10 RC differs from 3.10 final in a significant manner, we can roll a point release.

hugovk:
[...]

PR #5702, opened 2021-09-05
Install numpy on Python 3.10

radarhere:
I think I just missed this in https://github.com/hugovk/Pillow/pull/71 / #5666


hugovk:
Thanks!

(Cross ref https://github.com/python-pillow/Pillow/issues/5569)

PR #5710, opened 2021-09-12
Moved _info function into docstring

radarhere:
In selftest.py, the shell commands in the docstring refers a number of times to an earlier function, `_info`. This PR suggests merging that function into the docstring as a lambda instead.

This should help give users the impression that they can just be up-and-running from shell, rather than "Wait, I run these shell commands from the example... but they refer to an external function?"

I also removed the `load()` command from `_info`, as I found it wasn't necessary.

### DIFF
--- a/selftest.py
+++ b/selftest.py
@@ -14,17 +14,13 @@ except AttributeError:
     pass
 
 
-def _info(im):
-    im.load()
-    return im.format, im.mode, im.size
-
-
 def testimage():
     """
     PIL lets you create in-memory images with various pixel types:
 
     >>> from PIL import Image, ImageDraw, ImageFilter, ImageMath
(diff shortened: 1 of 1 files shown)


hugovk:
I find `def`s clearer, and [PEP 8](https://www.python.org/dev/peps/pep-0008/#programming-recommendations) discourages lambdas:

> Always use a def statement instead of an assignment statement that binds a lambda expression directly to an identifier:

```python
# Correct:
def f(x): return 2*x
```
```python
# Wrong:
f = lambda x: 2*x
```

> The first form means that the name of the resulting function object is specifically 'f' instead of the generic '\<lambda>'. This is more useful for tracebacks and string representations in general. The use of the assignment statement eliminates the sole benefit a lambda expression can offer over an explicit def statement (i.e. that it can be embedded inside a larger expression)

PR #5744, opened 2021-10-01
Make TIFF strip size configurable

### LINKED ISSUE #5740: Multi-strip usage when encoding group4 TIF breaks use in PDF
Hi,

since #5514, group4 tiff images are not anymore saved as a single strip. This breaks the use-case of utilizing PIL for encoding images as group4 for inclusion in PDF documents, as the PDF format only supports a single group4 encoded strip per image.

Please allow for a way to force single-strip group4 encoding to restore this functionality, maybe by allowing to specify the maximum strip length?

(I realize this sounds a lot like https://xkcd.com/1172/ :smile:)

kmilos:
Fixes #5740

 * Add a way to override the strip size by redefining `PIL.TiffImagePlugin.STRIP_SIZE` (in bytes)
 * Do not round up `rows_per_strip`, requested strip size should not be exceeded

### DIFF
--- a/src/PIL/TiffImagePlugin.py
+++ b/src/PIL/TiffImagePlugin.py
@@ -58,6 +58,7 @@ logger = logging.getLogger(__name__)
 READ_LIBTIFF = False
 WRITE_LIBTIFF = False
 IFD_LEGACY_API = True
+STRIP_SIZE = 65536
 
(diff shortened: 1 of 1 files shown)


kmilos:
Thanks for fixing up the PR!

PR #5750, opened 2021-10-09
Fixed Python errors when saving a (0, 0) TIFF image

radarhere:
At the moment, saving a zero by zero image to TIFF gives a `ZeroDivisionError`.
```pycon
>>> from PIL import Image
>>> Image.new("RGB", (0, 0)).save("out.tiff")
  File "PIL/Image.py", line 2235, in save
  File "PIL/TiffImagePlugin.py", line 1586, in _save
    strips_per_image = (im.size[1] + rows_per_strip - 1) // rows_per_strip
ZeroDivisionError: integer division or modulo by zero
```

This PR ensures that `rows_per_strip`, and then later `strip_byte_counts`, are not zero, allowing the code to progress forward to a `SystemError` instead.

```pycon
>>> Image.new("RGB", (0, 0)).save("out.tiff")
  File "PIL/Image.py", line 2240, in save
  File "PIL/TiffImagePlugin.py", line 1759, in _save
  File "PIL/ImageFile.py", line 531, in _save
```

### DIFF
--- a/Tests/test_file_libtiff.py
+++ b/Tests/test_file_libtiff.py
@@ -986,3 +986,10 @@ class TestFileLibTiff(LibTiffTestCase):
         with Image.open(out) as im:
             # Assert that there are multiple strips
             assert len(im.tag_v2[STRIPOFFSETS]) > 1
+
+    @pytest.mark.parametrize("compression", ("tiff_adobe_deflate", None))
+    def test_save_zero(self, compression, tmp_path):
+        im = Image.new("RGB", (0, 0))
+        out = str(tmp_path / "temp.tif")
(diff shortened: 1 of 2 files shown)


radarhere:
cc @kmilos

kmilos:
This was actually raising two different kinds of error in 8.1 and before: SystemError indeed if uncompressed (no libtiff), but a different

```
RuntimeError: Error setting from dictionary


_TIFFVSetField: out.tiff: Bad value 0 for "RowsPerStrip" tag.
```

when using libtiff...

So what's the behaviour you want, both raising the same SystemError now I guess?

PR #5751, opened 2021-10-09
Fix multiline text drawing and height calculation

### LINKED ISSUE #1646: Multiline text height not being calculated correctly

**The issue**

**Current code**
```py
```

```
170
244
```

**Proposed Change**
```py
```

axd1x8a:
Fixes #1646  .

 * Fixed size calculation and drawing of multiline text
 
 ### Example
 ```python

font = ImageFont.truetype("tnr.ttf", size=80, encoding="unic")
text = "{{{{\n}}}}"

size = font.getsize_multiline(text)
print(size) # (152, 132) / (152, 168)
image = Image.new("RGB", size, color="#FFF")
draw = ImageDraw.Draw(image)
draw.multiline_text((0, 0), text, "#000", font)
image.save("a.png")
```
### Then / Now
 
a a

### DIFF
--- a/Tests/test_imagefont.py
+++ b/Tests/test_imagefont.py
@@ -585,7 +585,7 @@ class TestImageFont:
         assert t.getsize_multiline("A") == (12, 16)
         assert t.getsize_multiline("AB") == (24, 16)
         assert t.getsize_multiline("a") == (12, 16)
-        assert t.getsize_multiline("ABC\n") == (36, 36)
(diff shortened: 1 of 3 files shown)


FeeeeK:
Most tests fail because they are not designed to account for these changes, for example before the height of the blank line was equal to the height of `A` string, although `font.getsize()` returns `(0, 0)`.
Most of the test images need to be changed to fix this.

PR #5768, opened 2021-10-15
Drop support for soon-EOL Python 3.6

hugovk:

 * The Python 3.6 EOL is scheduled for 2021-12-23, before Pillow 9.0.0's 2022-01-02 release
   * https://devguide.python.org/#status-of-python-branches
   * https://endoflife.date/python

 * Ran `pyupgrade **/*.py --py36-plus` to upgrade Python syntax

 * No changes running `pyupgrade **/*.py --py37-plus` (other than removing stuff that will be removed for `PILLOW_VERSION` in another PR)

 * ~Remove Docker jobs which test on Python 3.6:~
   * ~`centos-7-amd64`~
   * ~`centos-8-amd64`~
   * ~`centos-stream-8-amd64`~
     * ~Perhaps we could upgrade to 3.7 in a CentOS~
     * ~Or check when CentOS Stream 9 is out and what Python it'll have~
   * ~`ubuntu-18.04-bionic-amd64`~
     * ~If we want to test 18.04, we can add it add `ubuntu-18.04` as an `include:` to the regular `test.yml`~
   * (Updated to use 3.9: https://github.com/python-pillow/docker-images/pull/123 / https://github.com/hugovk/Pillow/pull/76) 

~Not removed the `if sys.version_info >= (3, 7):` stuff around the `PILLOW_VERSION` warnings, will remove that in the next PR.~ (Now removed, `PILLOW_VERSION` since removed in https://github.com/python-pillow/Pillow/pull/5776.)


cclauss:
Why `pyupgrade --py36-plus` and not `pyupgrade --py37-plus`?

PR #5790, opened 2021-10-23
Document #5788: "Use title for display"

nulano:
Updates documentation for #5788 similarly to the XV documentation.


radarhere:
Ok, I've created PR #5792

PR #5798, opened 2021-10-28
WebP: Fix memory leak during decoding on failure

ilai-deutel:
When creating the `WebPAnimDecoder` object, we create a `WebPAnimDecoderObject` and populate its data using `WebPDataCopy()`.

Subsequently, if either `WebPAnimDecoderNew()` or `WebPAnimDecoderGetInfo()` fails, data is not currently deallocated, which may cause memory leaks.

This PR clears the decoder object's data in that situation.


radarhere:
Thanks

PR #5811, opened 2021-11-04
Windows: Enable win-arm64 for MSVC

gaborkertesz-linaro:
This patch enabled win-arm64 as a new platform for Windows.
Platform query and documentation is updated accordingly.

### DIFF
--- a/winbuild/build.rst
+++ b/winbuild/build.rst
@@ -24,7 +24,7 @@ Download and install:
 * `CMake 3.12 or newer <https://cmake.org/download/>`_
   (also available as Visual Studio component C++ CMake tools for Windows)
 
-* `NASM <https://www.nasm.us/pub/nasm/releasebuilds/?C=M;O=D>`_
+* x86/x64: `NASM <https://www.nasm.us/pub/nasm/releasebuilds/?C=M;O=D>`_
 
 Any version of Visual Studio 2017 or newer should be supported,
 including Visual Studio 2017 Community, or Build Tools for Visual Studio 2019.
@@ -42,8 +42,8 @@ behaviour of ``build_prepare.py``:
   If ``PYTHON`` is unset, the version of Python used to run
   ``build_prepare.py`` will be used. If only ``PYTHON`` is set,
   ``EXECUTABLE`` defaults to ``python.exe``.
-* ``ARCHITECTURE`` is used to select a ``x86`` or ``x64`` build. By default,
-  uses same architecture as the version of Python used to run ``build_prepare.py``.
+* ``ARCHITECTURE`` is used to select a ``x86``, ``x64`` or ``ARM64``build.
+  By default, uses same architecture as the version of Python used to run ``build_prepare.py``.
   is used.
 * ``PILLOW_BUILD`` can be used to override the ``winbuild\build`` directory
   path, used to store generated build scripts and compiled libraries.
--- a/winbuild/build_prepare.py
+++ b/winbuild/build_prepare.py
@@ -1,4 +1,5 @@
 import os
+import platform
 import shutil
 import struct
 import subprocess
@@ -93,6 +94,7 @@ SF_MIRROR = "http://iweb.dl.sourceforge.net"
 architectures = {
     "x86": {"vcvars_arch": "x86", "msbuild_arch": "Win32"},
     "x64": {"vcvars_arch": "x86_amd64", "msbuild_arch": "x64"},
+    "ARM64": {"vcvars_arch": "x86_arm64", "msbuild_arch": "ARM64"},
 }
 
 header = [
@@ -490,7 +492,10 @@ if __name__ == "__main__":
     python_dir = os.environ.get("PYTHON")
     python_exe = os.environ.get("EXECUTABLE", "python.exe")
     architecture = os.environ.get(
-        "ARCHITECTURE", "x86" if struct.calcsize("P") == 4 else "x64"
+        "ARCHITECTURE",
+        "ARM64"
+        if platform.machine() == "ARM64"
+        else ("x86" if struct.calcsize("P") == 4 else "x64"),
     )
     build_dir = os.environ.get("PILLOW_BUILD", os.path.join(winbuild_dir, "build"))
     sources_dir = ""
--- a/winbuild/build_prepare.py
+++ b/winbuild/build_prepare.py
@@ -225,21 +225,21 @@ deps = {
         "filename": "lcms2-2.13.tar.gz",
         "dir": "lcms2-2.13",
         "patch": {
-            r"Projects\VC2017\lcms2_static\lcms2_static.vcxproj": {
+            r"Projects\VC2019\lcms2_static\lcms2_static.vcxproj": {
                 # default is /MD for x86 and /MT for x64, we need /MD always
                 "<RuntimeLibrary>MultiThreaded</RuntimeLibrary>": "<RuntimeLibrary>MultiThreadedDLL</RuntimeLibrary>",  # noqa: E501
                 # retarget to default toolset (selected by vcvarsall.bat)
-                "<PlatformToolset>v141</PlatformToolset>": "<PlatformToolset>$(DefaultPlatformToolset)</PlatformToolset>",  # noqa: E501
+                "<PlatformToolset>v142</PlatformToolset>": "<PlatformToolset>$(DefaultPlatformToolset)</PlatformToolset>",  # noqa: E501
                 # retarget to latest (selected by vcvarsall.bat)
-                "<WindowsTargetPlatformVersion>10.0.17134.0</WindowsTargetPlatformVersion>": "<WindowsTargetPlatformVersion>$(WindowsSDKVersion)</WindowsTargetPlatformVersion>",  # noqa: E501
+                "<WindowsTargetPlatformVersion>10.0</WindowsTargetPlatformVersion>": "<WindowsTargetPlatformVersion>$(WindowsSDKVersion)</WindowsTargetPlatformVersion>",  # noqa: E501
             }
         },
         "build": [
             cmd_rmdir("Lib"),
-            cmd_rmdir(r"Projects\VC2017\Release"),
-            cmd_msbuild(r"Projects\VC2017\lcms2.sln", "Release", "Clean"),
+            cmd_rmdir(r"Projects\VC2019\Release"),
+            cmd_msbuild(r"Projects\VC2019\lcms2.sln", "Release", "Clean"),
             cmd_msbuild(
-                r"Projects\VC2017\lcms2.sln", "Release", "lcms2_static:Rebuild"
+                r"Projects\VC2019\lcms2.sln", "Release", "lcms2_static:Rebuild"
             ),
             cmd_xcopy("include", "{inc_dir}"),
         ],


gaborkertesz-linaro:
[win-arm64_pytests.txt](https://github.com/python-pillow/Pillow/files/7474294/win-arm64_pytests.txt)
[win-arm64_selftests.txt](https://github.com/python-pillow/Pillow/files/7474295/win-arm64_selftests.txt)

nulano:
Could you please post the full Pytest output? In particular, I would be interested in seeing the skips and the list of loaded dependencies (at the start of the output).

gaborkertesz-linaro:
[x64_pytests_full.txt](https://github.com/python-pillow/Pillow/files/7475627/x64_pytests_full.txt)
[win-arm64_pytests_full.txt](https://github.com/python-pillow/Pillow/files/7475631/win-arm64_pytests_full.txt)

I attach the x64 results as a reference also as it looks that there 3 common failing tests.

winbuild/build.rst
```diff
@@ -42,8 +42,8 @@ behaviour of ``build_prepare.py``:
   If ``PYTHON`` is unset, the version of Python used to run
   ``build_prepare.py`` will be used. If only ``PYTHON`` is set,
   ``EXECUTABLE`` defaults to ``python.exe``.
-* ``ARCHITECTURE`` is used to select a ``x86`` or ``x64`` build. By default,
-  uses same architecture as the version of Python used to run ``build_prepare.py``.
+* ``ARCHITECTURE`` is used to select a ``x86``, ``x64`` or ``x86_arm64`` (for win-arm64) build.
```
kmilos, line 45: `x86_arm64` seems very odd to see... why not simply `arm64`?

gaborkertesz-linaro: I agree it's strange.
It's coming from vcvarsall.bat (attached) and the fact the win-arm64 MSVC compiler is basically a cross-compiler, not native arm64. The host is working in x86 emulation mode.
[vcvarsall.bat.txt](https://github.com/python-pillow/Pillow/files/7483176/vcvarsall.bat.txt)

kmilos: Ah, got it. Thanks.

nulano: Since this is an internal value only used by Pillow, there is nothing preventing the use of `arm64`. In fact, while `vcvarsall` uses `x86_arm64`, MSBuild uses `ARM64` (according to this PR, I haven't checked myself). The `x64` value already disagrees with `vcvarsall`'s `x86_amd64`. So I would also vote for either `arm64` or `ARM64`. Of the two, I would probably prefer `ARM64` to match `platform.machine()`, but either is fine.

hugovk:
Is this testable on GitHub Actions?

gaborkertesz-linaro:
> Is this testable on GitHub Actions?

Not yet unfortunetaly.

winbuild/build.rst
gaborkertesz-linaro: Good point, updated!

gaborkertesz-linaro:
I've investigated the following test: _test_image_access.py::TestEmbeddable::test_embeddable_

It's only enabled local Windows run: https://github.com/python-pillow/Pillow/blob/main/Tests/test_image_access.py#L399
However it fails for me on x64 and win-arm64 as well.

I tend to think this test should be removed or fix for all platforms as it should be a platform agnostic test as embedding into C/C++ code should be universal.

I see several issues:

1. First error: LINK : fatal error LNK1104: cannot open file 'python310.lib'
- Library directory list doesn't include the Python libraries: https://github.com/python-pillow/Pillow/blob/main/Tests/test_image_access.py#L437
- I was able to fix that by adding the missing libraries in test_image_access.py.
2. Next error: "python310.dll not found"
- I was able to fix that by adding the required directory to the path in test_image_access.py also.
3. Next error: "ModuleNotFoundError: No module named 'encodings'"
- I run Python in venv and the compiled C code set Python home to the venv's home directory, but changing that to the Python executable's home directory fixed that in: https://github.com/python-pillow/Pillow/blob/main/Tests/test_image_access.py#L413 
- Next error: https://github.com/python-pillow/Pillow/blob/main/Tests/test_image_access.py#L416 returns with error, which results _AssertionError: assert 3221225477 == 0_

nulano:
> 4. Next error: https://github.com/python-pillow/Pillow/blob/main/Tests/test_image_access.py#L416 returns with error, which results _AssertionError: assert 3221225477 == 0_

For reference, 3221225477 == 0xC0000005, i.e. access violation. From MSDN:
> Specifically, a C0000005 error is an access violation error caused by a buffer overrun.

gaborkertesz-linaro:
> > 1. Next error: https://github.com/python-pillow/Pillow/blob/main/Tests/test_image_access.py#L416 returns with error, which results _AssertionError: assert 3221225477 == 0_
> 
> For reference, 3221225477 == 0xC0000005, i.e. access violation. From MSDN:
> 
> > Specifically, a C0000005 error is an access violation error caused by a buffer overrun.
Thanks!

However my main question is:
Would you welcome if I investigate this issue further? I'm hesitating a bit as this test is disabled by most cases. If you would prefer to see this to be fixed, I'll investigate it further.

nulano:
> Would you welcome if I investigate this issue further?

As far as I'm concerned, I wasn't around when the test was added, and I haven't seen it pass on Windows in the past 2 years. I've also not looked into it further myself.

winbuild/build.rst
```diff
@@ -42,8 +42,8 @@ behaviour of ``build_prepare.py``:
   If ``PYTHON`` is unset, the version of Python used to run
   ``build_prepare.py`` will be used. If only ``PYTHON`` is set,
   ``EXECUTABLE`` defaults to ``python.exe``.
-* ``ARCHITECTURE`` is used to select a ``x86`` or ``x64`` build. By default,
-  uses same architecture as the version of Python used to run ``build_prepare.py``.
+* ``ARCHITECTURE`` is used to select a ``x86``, ``x64`` or ``ARM64``build.
+  By default, uses same architecture as the version of Python used to run ``build_prepare.py``.
   is used.
```
nulano, line 47: > From https://github.com/python-pillow/Pillow/pull/5811#issuecomment-961081944, win-arm64_pytests_full.txt:
> ```
> --- PIL CORE support ok, compiled for 9.0.0.dev0
> *** TKINTER support not installed
> *** FREETYPE2 support not installed
> *** LITTLECMS2 support not installed
> --- WEBP support ok, loaded 1.2.1
> --- WEBP Transparency support ok
> --- WEBPMUX support ok
> --- WEBP Animation support ok
> --- JPEG support ok, compiled for libjpeg-turbo 2.1.1
> --- OPENJPEG (JPEG2000) support ok, loaded 2.4.0
> --- ZLIB (PNG/ZIP) support ok, loaded 1.2.11
> --- LIBTIFF support ok, loaded 4.3.0
> *** RAQM (Bidirectional Text) support not installed
> *** LIBIMAGEQUANT (Quantization method) support not installed
> *** XCB (X protocol) support not installed
> ```

Based on this and the other PR you opened, it is my understanding that not all dependencies build successfully on ARM64. This should probably be noted in the readme, perhaps as follows:

```suggestion
  is used. Note that the ``ARM64`` build is not fully supported and some dependencies may fail to build. Pull requests are welcome.
```

gaborkertesz-linaro: Yes, the current status not all optional external libraries are built.
However FreeType and libimagequant changes are upstreamed, I'm waiting for new releases to integrate, while littlecms2 is in review.

FreeType:
https://gitlab.freedesktop.org/freetype/freetype/-/commit/80bd4cba7eb25ecae313a6ac376ad6b363655243
https://gitlab.freedesktop.org/freetype/freetype/-/commit/e6e6cbf1648d4a776da0857921872f2fbc853205
https://github.com/mesonbuild/wrapdb/pull/205

libimagequant:
https://github.com/ImageOptim/libimagequant/commit/43c1c60e0302400b41ca9e69a2f0d2c710700647

littlecms2:
https://github.com/mm2/Little-CMS/pull/288

gaborkertesz-linaro:
> > Would you welcome if I investigate this issue further?
> 
> As far as I'm concerned, I wasn't around when the test was added, and I haven't seen it pass on Windows in the past 2 years. I've also not looked into it further myself.

Wouldn't it be a right time to remove this test to avoid further confusion?

gaborkertesz-linaro:
@radarhere @nulano @hugovk Sorry for the reminder, but can I do anything to close this PR?

hugovk (approved):

winbuild/build_prepare.py
```diff
@@ -257,10 +259,10 @@ def cmd_msbuild(
         "libs": [r"bin\*.lib"],
     },
     "libimagequant": {
-        # commit: Merge branch 'master' into msvc (matches 2.16.0 tag)
-        "url": "https://github.com/ImageOptim/libimagequant/archive/f41ee301ff3a407b16991af3dbe03910919bbdc3.zip",  # noqa: E501
-        "filename": "libimagequant-f41ee301ff3a407b16991af3dbe03910919bbdc3.zip",
-        "dir": "libimagequant-f41ee301ff3a407b16991af3dbe03910919bbdc3",
```
radarhere, line 263: Why switch to the master branch here? As the comment says, previously this used the msvc branch, so the 2.17.0 equivalent would be https://github.com/ImageOptim/libimagequant/commit/e4c1334be0eff290af5e2b4155057c2953a313ab

winbuild/build.rst
nulano, line 47: This change seems to have disappered in a force-push. Is there any reason not to include it?

```suggestion
  is used. Note that the ``ARM64`` build is not fully supported and some dependencies may fail to build. Pull requests are welcome.
```

winbuild/build_prepare.py
radarhere: I've updated to the 2.17.0 version of the msvc branch in #5916

winbuild/build.rst
gaborkertesz-linaro: Sorry I didn't mention that LittleCMS is the last dependency and VS2019 update is merged, just not released yet:
https://github.com/mm2/Little-CMS/pull/288
Patch to use VS2019 project: https://github.com/gaborkertesz-linaro/Pillow/commit/0af6042c5c0a412c2614147458ce2d4832436bc8

So if either VS2019 project is accepted or VS2017 will be updated for ARM64 as well and LCMS2 is released, all external dependencies will be enabled.

radarhere: lcms2 2.13 has now been released.

gaborkertesz-linaro: Thanks for the heads up, I'm going to rebase my changes and run the tests!

gaborkertesz-linaro:
[01_build_prepare.log](https://github.com/python-pillow/Pillow/files/7972070/01_build_prepare.log)
[02_build_dep_all.log](https://github.com/python-pillow/Pillow/files/7972071/02_build_dep_all.log)
[03_build_pillow_install.log](https://github.com/python-pillow/Pillow/files/7972072/03_build_pillow_install.log)
[04_selftest.log](https://github.com/python-pillow/Pillow/files/7972073/04_selftest.log)
[05_Tests.log](https://github.com/python-pillow/Pillow/files/7972074/05_Tests.log)

nulano:
Could you please re-run the tests with `C:\kg\python_test\github_packages\Pillow\winbuild\build\bin` added to `PATH`? It seems fribidi.dll wasn't found, likely because of this. Looks mostly OK otherwise. The commands should probably be something like:

```cmd
PATH %PATH%;C:\kg\python_test\github_packages\Pillow\winbuild\build\bin
"C:\kg\python_test\venv_310_new\Scripts\python.exe" -m pytest -v Tests
```

gaborkertesz-linaro:
[05_Tests2.log](https://github.com/python-pillow/Pillow/files/7977537/05_Tests2.log)
Tests are repeated with Pillow\winbuild\build\bin added to the path.

nulano (approved):
Thank you, looks mostly good.

The first three failures seem to be a difference in floating point rounding:
```
FAILED Tests/test_color_lut.py::TestTransformColorLut3D::test_3_to_4_channels
    At index 4 diff: 0.04000000000000001 != 0.04
FAILED Tests/test_color_lut.py::TestTransformColorLut3D::test_4_to_4_channels
    At index 4 diff: 0.04000000000000001 != 0.04
FAILED Tests/test_color_lut.py::TestTransformColorLut3D::test_with_normals_3_channels
    At index 12 diff: 0.15999999999999992 != 0.16000000000000003
```

The next two could be caused by a similar difference, being a small margin above the epsilon (but hard to be sure without visually inspecting the images):
```
FAILED Tests/test_file_wmf.py::test_load_raw - AssertionError:  average pixel value difference 2.0479 > epsilon 2.0000
FAILED Tests/test_file_wmf.py::test_load_set_dpi - AssertionError:  average pixel value difference 2.6547 > epsilon 2.1000
```

The next failed test I have never seen pass on Windows:
```
FAILED Tests/test_image_access.py::TestEmbeddable::test_embeddable - distutil...
```

And the last one could be due to how the tests are being run. Are you running the tests directly in a desktop session, or are you using some sort of terminal without a desktop, e.g. SSH? Either way this seems like a separate issue unrelated to ARM.
```
FAILED Tests/test_imagegrab.py::TestImageGrab::test_grab - OSError: screen gr...
```

gaborkertesz-linaro:
> And the last one could be due to how the tests are being run. Are you running the tests directly in a desktop session, or are you using some sort of terminal without a desktop, e.g. SSH? Either way this seems like a separate issue unrelated to ARM.

I connect to my local PC via remote desktop and run the tests in Windows Terminal, so I'd expect this shouldn't prevent screengrab. Probably as a follow-up taks this could be investigated.

gaborkertesz-linaro:
Thanks for the thorough and persistent support!
If anybody ever interested about Pillow and dependencies porting to WoA, I've created a confluence page summary:
https://linaro.atlassian.net/wiki/spaces/WOAR/pages/28654632982/A+tale+of+a+dependency+chain+Pillow+and+the+crew

nulano:
> If anybody ever interested about Pillow and dependencies porting to WoA, I've created a confluence page summary: https://linaro.atlassian.net/wiki/spaces/WOAR/pages/28654632982/A+tale+of+a+dependency+chain+Pillow+and+the+crew

A small clarification: Pillow doesn't use libpng, HarfBuzz, FriBiDi directly. These are only listed in build_prepare because they are used by the other libraries.

Specifically, libpng is used by FreeType to support color fonts in CBDT/SBIX format, and HarfBuzz/FriBiDi are dependencies of Raqm. Raqm used to be built directly, but as of Pillow 8.2.0 a modified version is included in the sources instead and is now the default on Windows. For details see: https://pillow.readthedocs.io/en/stable/installation.html#external-libraries

gaborkertesz-linaro:
> > If anybody ever interested about Pillow and dependencies porting to WoA, I've created a confluence page summary: https://linaro.atlassian.net/wiki/spaces/WOAR/pages/28654632982/A+tale+of+a+dependency+chain+Pillow+and+the+crew
> 
> A small clarification: Pillow doesn't use libpng, HarfBuzz, FriBiDi directly. These are only listed in build_prepare because they are used by the other libraries.
> 
> Specifically, libpng is used by FreeType to support color fonts in CBDT/SBIX format, and HarfBuzz/FriBiDi are dependencies of Raqm. Raqm used to be built directly, but as of Pillow 8.2.0 a modified version is included in the sources and is now the default on Windows. For details see: https://pillow.readthedocs.io/en/stable/installation.html#external-libraries

Thanks, I'll fix these issues!
Why libpng is needed to be imported directly by Pillow, while FreeType imports for itself anyway?

nulano:
> Why libpng is needed to be imported directly by Pillow, while FreeType imports for itself anyway?

I don't understand your question. Doing a grep for libpng in Pillow only shows build_prepare.py and test-windows.yml for me, the latter containing a comment that libpng is only used by FreeType.

Are you asking why build_prepare generates a script to compile the indirect dependencies? The VS2010 build of FreeType doesn't build libpng directly, libpng must already be present and compiled somewhere, so build_prepare takes care of that.

gaborkertesz-linaro:
Thanks, yes, that was my concern that FreeTypes includes libpng dependency in its source and I'd expect that ideally it should take care of build dependencies:
https://gitlab.freedesktop.org/freetype/freetype/-/blob/master/subprojects/libpng.wrap
However I see the released package doesn't include the subprojects.

PR #5812, opened 2021-11-04
Windows: Try to build all enabled optional dependencies

gaborkertesz-linaro:
This patch enabled win-arm64 as a new platform for Windows.
Platform query and documentation is updated accordingly.

### DIFF
--- a/winbuild/build_prepare.py
+++ b/winbuild/build_prepare.py
@@ -404,7 +404,7 @@ def write_script(name, lines):
     lines = [line.format(**prefs) for line in lines]
     print("Writing " + name)
     with open(name, "w") as f:
-        f.write("\n\r".join(lines))
+        f.write("\n".join(lines))
     if verbose:
         for line in lines:
             print("    " + line)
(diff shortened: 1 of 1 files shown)


gaborkertesz-linaro:
Tested on x64 and win-arm64.

build_dep_all.cmd before this change:
[build_all_orig.cmd.txt](https://github.com/python-pillow/Pillow/files/7474313/build_all_orig.cmd.txt)

build_dep_all.cmd with this change:
[build_all_mod.cmd.txt](https://github.com/python-pillow/Pillow/files/7474314/build_all_mod.cmd.txt)

I'm open for any recommendation if this solution won't fit to your CI/CD system and I would accept rejection, but this patch helps my current task, so I thought it can be useful for others later on.

gaborkertesz-linaro:
Steps are done as tests:

python \winbuild\build_prepare.py
winbuild\build\build_dep_all.cmd

Output
`Pillow dependencies build result:
  libjpeg: successful
  zlib: successful
  libtiff: successful
  libwebp: successful
  libpng: successful
  freetype: failed
  lcms2: failed
  openjpeg: successful
  libimagequant: failed
  harfbuzz: failed
  fribidi: successful`

winbuild\build\build_pillow.cmd install

python selftest.py
pytest Tests

nulano:
The expectation is that all dependencies should build correctly or be explicitly disabled. As is, this patch would break that assumption. While `build_dep_all.cmd` is not used on GitHub Actions, it is used on AppVeyor, where a silent failure would be even worse as it could go undetected for a long time (as happened before this script was in use).

There are also internal dependencies between the dependencies themselves, for example it makes no sense to build HarfBuzz and FriBiDi if FreeType failed, and FreeType is configured to require libpng.

If you only require some of the dependencies, you are welcome to run the required cmd files yourself with `cmd /c build_dep_xyz.cmd` as I often do when updating the build of some dependency, and as mentioned in the documentation:

> To build all dependencies, run ``winbuild\build\build_dep_all.cmd``, or run the individual scripts to build each dependency separately.

gaborkertesz-linaro:
Ok, understood thanks for the quick feedback!

PR #5824, opened 2021-11-09
Added rounding when converting P and PA

### LINKED ISSUE #5728: Conversion P to L wrong?
```python
from PIL import Image

img = Image.open('p.png')
print('P->L: ', img.convert('L').getpixel((0, 0)))
print('P->RGB: ', img.convert('RGB').getpixel((0, 0)))
print('P->RGB->L: ', img.convert('RGB').convert('L').getpixel((0, 0)))
```

p.png:  p

Old
PIL: 6.2.2
P->L: 175
P->RGB: (177, 175, 175)
P->RGB->L: 175

7.0.0+
After commit [#4320](https://github.com/python-pillow/Pillow/pull/4320)
PIL:  8.3.2
P->L:  **175**
P->RGB:  (177, 175, 175)
P->RGB->L:  **176**

radarhere:
Resolves #5728 

The issue discusses a pixel in an image with the palette value `(177, 175, 175)`, and asks why converting from P to L doesn't convert it to 176. That conversion uses the L macro -

https://github.com/python-pillow/Pillow/blob/a3d1e2f85db890c6a1623e3902711863f55cc54e/src/libImaging/Convert.c#L1012-L1016

https://github.com/python-pillow/Pillow/blob/a3d1e2f85db890c6a1623e3902711863f55cc54e/src/libImaging/Convert.c#L43

Applying the L macro,

`(177*299 + 175*587 + 175*114) / 1000` = 175.598

Except the `p2l` code above rounds it down. This PR applies proper rounding to bring it to 176.

### DIFF
--- a/Tests/test_image_convert.py
+++ b/Tests/test_image_convert.py
@@ -93,7 +93,7 @@ def test_trns_p(tmp_path):
     f = str(tmp_path / "temp.png")
 
     im_l = im.convert("L")
-    assert im_l.info["transparency"] == 0  # undone
+    assert im_l.info["transparency"] == 1  # undone
     im_l.save(f)
 
     im_rgb = im.convert("RGB")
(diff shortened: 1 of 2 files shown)


Tests/test_image_convert.py
```diff
@@ -93,7 +93,7 @@ def test_trns_p(tmp_path):
     f = str(tmp_path / "temp.png")
 
     im_l = im.convert("L")
-    assert im_l.info["transparency"] == 0  # undone
+    assert im_l.info["transparency"] == 1  # undone
```
radarhere, line 96: The first palette entry is `(0, 1, 2)`. `(0*299 + 1*587 + 2*114) / 1000` = 0.815, so this is now rounded to 1.

src/libImaging/Convert.c
```diff
@@ -1013,7 +1013,7 @@ p2l(UINT8 *out, const UINT8 *in, int xsize, const UINT8 *palette) {
     int x;
     /* FIXME: precalculate greyscale palette? */
     for (x = 0; x < xsize; x++) {
-        *out++ = L(&palette[in[x] * 4]) / 1000;
+        *out++ = round(L(&palette[in[x] * 4]) / 1000.0);
```
homm, line 1016: Floating point arithmetic slower than integer. It's better to use `L24` macro here, it already contains rounding.

radarhere: Thanks. I've updated the commit.

PR #5624, opened 2021-07-21
Prefer command line tools SDK on macOS

radarhere:
Resolves #5622

PR #5828, opened 2021-11-11
Only prefer command line tools SDK on macOS over the default

### LINKED ISSUE #5827: Allow setting specific SDK on macOS
Hey all

I help package things for conda-forge, and pillow 8.4 recently got stuck in limbo due to https://github.com/python-pillow/Pillow/pull/5624 breaking the build we (need to) use there. We rely on the fact that we can control the MacOS SDK being used (which is explicitly set up by conda-build), but after #5624, the sdk from the system gets preference and ends up breaking (mismatched compiler vs. lib).

I reverted it to get https://github.com/conda-forge/pillow-feedstock/pull/101 out the door, but carrying patches indefinitely is not great.

I haven't studied #5622, but I hope we can find a solution that suits everyone. This could (for example) work with an environment variable that gets precedence if non-empty (but many other solutions would likely be possible).

radarhere:
Resolves #5827

#5624 changed setup.py on macOS to prefer the command line tools SDK over `xcrun --show-sdk-path`. Apparently that was too far however, as conda-forge would like to specify the SDK path.

So instead, this PR scales back #5624 to only use the command line tools SDK when `xcrun --show-sdk-path` returns the default path of `/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -532,16 +532,24 @@ class pil_build_ext(build_ext):
                 _add_directory(include_dirs, "/usr/X11/include")
 
             # SDK install path
-            sdk_path = "/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk"
-            if not os.path.exists(sdk_path):
-                try:
-                    sdk_path = (
-                        subprocess.check_output(["xcrun", "--show-sdk-path"])
-                        .strip()
-                        .decode("latin1")
-                    )
-                except Exception:
-                    sdk_path = None
+            try:
+                sdk_path = (
+                    subprocess.check_output(["xcrun", "--show-sdk-path"])
+                    .strip()
+                    .decode("latin1")
+                )
+            except Exception:
+                sdk_path = None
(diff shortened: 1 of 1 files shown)


hugovk:
`build_extenstions()` is 500 lines long, shall we put this changed chunk into its own function?

radarhere:
Ok, sure, done.

h-vetinari:
I tested this PR in https://github.com/conda-forge/pillow-feedstock/pull/102, and it works fine! 🥳 

Thanks a lot for the quick handling, much appreciated! :)

hugovk:
Installs fine and tests pass for me on macOS Monterey 12.0.1 with defaults:

```console
$  xcrun --show-sdk-path
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk
```

Thanks both!

Issue #5729, opened 2021-09-22
TGA buffer overrun


Davnit:
Currently only top-left (0x20) and bottom-left (0x00) are supported, and TGA images with other orientations fail to open. Additional orientations are top-right (0x30) and bottom-right (0x10).

Attached is a sample file with the 0x30 (top-right) orientation.
[icons_clan.zip](https://github.com/python-pillow/Pillow/files/7207834/icons_clan.zip)

radarhere:
http://paulbourke.net/dataformats/tga/ doesn't list these additional orientations, but page 12 of https://www.dca.fee.unicamp.br/~martino/disciplinas/ea978/tgaffs.pdf does.

I don't suppose that image is one that you're happy for us to include in our test suite, and distribute under the Pillow license?

Davnit:
Yea they seem to be relatively obscure. Not even GIMP opens it correctly.

I do not own the rights to that image so unfortunately permission is not mine to give. It's just the one I found that broke.

radarhere:
I've created PR #5829 to add these new orientations.

However, running that over the original image you supplied triggers a "buffer overrun when reading image file" by hitting https://github.com/python-pillow/Pillow/blob/24f0bbf5ecd2b4bc230de7ad8f5c23c7c8377fdc/src/libImaging/TgaRleDecode.c#L80

radarhere:
Do you happen to know where the original image came from?

Davnit:
As far as I know it was created by someone a long time ago who has no recollection of how or what they did in making it. I only noticed the unusual rotation value and then while researching it found out it wasn't unique.

Contextually though, it was designed to work with a [wrapper format](https://bnetdocs.org/document/25/icons-bni) used by clients connecting to Blizzard's old Battle.net service. None of the official images from Blizzard used this rotation though.

radarhere:
#5829 is now merged, so the next Pillow release (due on January 1) will support the additional orientations.

radarhere:
Pillow 9.0 has been released.

radarhere:
@Davnit is this resolved now that the additional orientations have been added, or do you think your sample image should be able to be read by Pillow?

Davnit:
It'd be nice for it to be readable, but that's probably a separate issue. I don't entirely understand which aspect of that image is causing the overrun though.

radarhere:
Ok, I've figured it out.

https://en.wikipedia.org/wiki/Truevision_TGA
[...]

PR #5829, opened 2021-11-11
Added support for top right and bottom right TGA orientations

radarhere:
Helps #5729

While http://paulbourke.net/dataformats/tga/ states that 'Bit 4' of the 'Image Specification' is 'reserved.  Must be set to 0.', page 12 of https://www.dca.fee.unicamp.br/~martino/disciplinas/ea978/tgaffs.pdf states that 'Bit 4 is for left-to-right ordering'.

So this PR allows it to flip a TGA image horizontally.


hugovk:
Thanks! Also a candidate for release notes?

radarhere:
Sure. See #5847

PR #5833, opened 2021-11-13
Fix pushes_fd method signature

hoodmane:
Python getters are supposed to take a `void* closure` argument.

I am coming from Pyodide. Web assembly has no native support for function pointer casting -- a function declared with a certain signature must be called with the same signature, or it will fail with `RuntimeError: function signature mismatch`. Emscripten has support for emulating function pointer casts which we are currently using, but it imposes a significant performance and stack space cost and we are trying to remove the pointer cast emulation pyodide/pyodide#1677. To do this, we need to patch Python packages so that they have function signatures that match the calling conventions they declare.

### DIFF
--- a/src/decode.c
+++ b/src/decode.c
@@ -235,7 +235,7 @@ _setfd(ImagingDecoderObject *decoder, PyObject *args) {
 }
 
 static PyObject *
-_get_pulls_fd(ImagingDecoderObject *decoder) {
+_get_pulls_fd(ImagingDecoderObject *decoder, void *closure) {
     return PyBool_FromLong(decoder->pulls_fd);
 }
 
--- a/src/encode.c
+++ b/src/encode.c
@@ -299,7 +299,7 @@ _setfd(ImagingEncoderObject *encoder, PyObject *args) {
 }
 
 static PyObject *
-_get_pushes_fd(ImagingEncoderObject *encoder) {
+_get_pushes_fd(ImagingEncoderObject *encoder, void *closure) {
     return PyBool_FromLong(encoder->pushes_fd);
 }
 


hoodmane:
I also removed all of the `(getter)` casts, but I'm not sure if that will work correctly since strictly speaking most of the function pointers are defined with type e.g., 
```C
PyObject *
get_some_attr(SpecificTypeOfObject *decoder, void *closure)
```
whereas the first argument is expected to be a `PyObject *`. So depending on what error / warning flags are set, this may generate a bunch of warnings or a compile error.

Probably it's better to put the casts back, but it's unfortunate that we can't get better type safety here -- it's so easy to mess up these signatures.

radarhere:
It is indeed generating a bunch of warnings - https://github.com/python-pillow/Pillow/runs/4202395176#step:9:26

> src/_imaging.c:3571:14: warning: initialization of ‘PyObject * (*)(PyObject *, void *)’ {aka ‘struct _object * (*)(struct _object *, void *)’} from incompatible pointer type ‘PyObject * (*)(ImagingObject *, void *)’ {aka ‘struct _object * (*)(struct <anonymous> *, void *)’} [-Wincompatible-pointer-types]
 3571 |     {"mode", _getattr_mode},
>       |              ^~~~~~~~~~~~~
> src/_imaging.c:3571:14: note: (near initialization for ‘getsetters[0].get’)

hoodmane:
Okay I put the casts back.

hugovk:
Please could you rebase or redo those two commits, to clean the history up a little bit?

One way:
```sh
git reset --soft HEAD~2
git push --force
```

radarhere (approved):

hugovk:
Thanks!

Issue #5700, opened 2021-09-01
Major (non-linear) performance regression with `ImageDraw.polygon`


ntninja:
### What did you do?

See the sample program below for a minimal test case – it simply draws a polygon from a list of input coordinates.

### What did you expect to happen?

Time taken to render the polygon line to be linear (and very small) even for 100'000s of points.

### What actually happened?

Apparently in any Pillow version after 7.0.0 the time taken to draw coordinate lists appears to increase non-linear with the number of input coordinates. When processing larger input files we regularily end up with 100'000s or perhaps even 1'000'000s points to be (nicely) drawn by Pillow for file visualization. With 7.0.0 this was no problem, with drawing times hardly ever exceeding a second at most, but since 7.1.1 (and likely also 7.1.0) the drawing suddenly turned into the slowest part of our production pipeline for larger files.

### What are your OS, Python and Pillow versions?

* OS: Linux x86-64 with Conda
* Python: 3.6 or 3.9
* Pillow: 7.1.1 (and any later version, 7.1.0 is likely affected too ) – using [conda-forge build](https://github.com/conda-forge/pillow-feedstock/)

Very likely related: https://github.com/python-pillow/Pillow/pull/4333

```python
#!/usr/bin/env python3                                                              
import json                                                                         
                                                                                    
from PIL import Image, ImageDraw, ImageFont                                         
                                                                                    
oversampling_factor = 2                                                             
width = 707                                                                         
height = 90                                                                         
scolor = "#000000"                                                                  
background_color = "#FFFFFF"                                                        
                                                                                    
image = Image.new(                                                                  
    "RGB",                                                                          
    (width * oversampling_factor, height * oversampling_factor),                    
    background_color,                                                               
)                                                                                   
                                                                                    
draw = ImageDraw.Draw(image)                                                        
with open("pillow_slow_polygon_coords.json", "r") as f:                             
    draw.polygon(json.load(f), scolor)                                              
                                                                                    
image = image.resize((width, height), Image.ANTIALIAS)                              
image.save("pillow_slow_polygon_output.png")
```

Referenced JSON file: [pillow_slow_polygon_coords.json](https://github.com/python-pillow/Pillow/files/7090994/pillow_slow_polygon_coords.json.txt) – this is audio waveform data but the issue is likely the same for other data as only the number of coordinates appear to matter.

radarhere:
Testing, yes, #4333 is responsible for this.

radarhere:
[...]

PR #5835, opened 2021-11-15
Only prevent repeated polygon pixels when drawing with transparency

radarhere:
#4333 prevented polygon pixels from being drawn more than once, so that the transparencies of pixels would not be combined to create an unintended result.

However, #5700 reported that this results in a speed decrease.

This PR switches back to the earlier, faster method of drawing pixels repeatedly - when drawing without transparency. This means that the slower method to avoid the transparency problem is not used when it does not need to be used.


ntninja:
Thank you! I will test this shortly and report back!

ntninja:
I can confirm this works in our setup, thank you!

hugovk:
@ntninja Thank you for testing!

@radarhere Thank you for the PR! Can we add tests to cover the `(hasAlpha == 1)` branch?

radarhere:
Ah, I had presumed that the test from #4333 would still cover this situation, but #4523 changed ellipses so that they are no longer polygons.

Ok, I've added a test.

hugovk:
Thanks both!

Issue #672, opened 2014-05-22
Tiled tif write support


jamesra:
I would like to write some BigTiff 4GB+ images as tiled tiffs to export data to an Omero server.  Pillow appears to support reading tiled tiff but not writing.

Being optimistic I naively added the TileWidth and TileLength tags to a tifinfo dictionary passed to the save function the result is an invalid tif file.  My test code can be found on github here:

https://github.com/jamesra/random/blob/master/pillow_tif_tiles.py

TiffInfo shows the tags correctly in the resulting tif file, but it fails to load in Photoshop or any of the built-in Windows viewers I tried.

OniDaito:
Hi all. I've just started looking at the source as I'm really keen on BigTIFF (not necessarily tiled tiff) support. Has there been any movement on this at all or shall I just crack on with looking at libtiff? Cheers

jamesra:
I haven’t had a chance to make progress.

It has been a while but I vaguely remember a Pillow API issue where I could not read/write specific tiles.  I believe Pillow wanted to read or write the entire file.  I need a method where I can update individual tiles.  Like many BigTIFF users our images are too large to entirely load into memory and get decent performance.

On Jun 23, 2015, at 9:19 AM, Benjamin Blundell <notifications@github.com<mailto:notifications@github.com>> wrote:


Hi all. I've just started looking at the source as I'm really keen on BigTIFF (not necessarily tiled tiff) support. Has there been any movement on this at all or shall I just crack on with looking at libtiff? Cheers

radarhere:
Does the image size limit imposed by #2583 mean that this scenario is no longer viable?

bjtho08:
Will this ever be implemented in Pillow?

aclark4life:
@bjtho08 Probably no time soon, unfortunately, unless you are offering to do it for us 😄 And it also may be a "wontfix" according to @radarhere 3 years ago, sorry.

PR #5839, opened 2021-11-18
Block tile TIFF tags when saving

### LINKED ISSUE #5838: Three TIFF tags cause display issues in GIS software
When I run my GeoTIFF though Pillow while trying to preserve the TIFF info I end up with 3 additional TIFF tags in the output (`StripOffsets`, `RowsPerStrip`, `StripByteCounts`). The resulting output GeoTIFF is now displayed incorrectly in the GIS software `ESRI ArcMap` (all cell values are now `NoData`), but is displayed correctly with `Pillow` or `Matplotlib`. The input GeoTIFF image is produced with `ESRI ArcMap` and does not have the 3 extra TIFF tags.

This could of course be an issue with `ESRI ArcMap`, but I believe the 3 Pillow generated TIFF tags might be wrong. They show up as:
```
StripOffsets = (810,)
RowsPerStrip = (40,)
StripByteCounts = (6400,)
```

Input GeoTIFF:
[dem_input.zip](https://github.com/python-pillow/Pillow/files/7557972/dem_input.zip)

```python
from PIL import Image

EXTRA_TIFF_TAGS = {"StripOffsets" : 273, "RowsPerStrip" : 278, "StripByteCounts" : 279}

image_input = Image.open("dem_input.tif")
print([TAG in image_input.tag._tagdata for TAG in EXTRA_TIFF_TAGS.values()])   # [False, False, False]

image_input.save("dem_output.tif", tiffinfo = image_input.tag)

image_output = Image.open("dem_output.tif")
print([TAG in image_output.tag._tagdata for TAG in EXTRA_TIFF_TAGS.values()])   # [True, True, True]

print([image_output.tag[TAG] for TAG in EXTRA_TIFF_TAGS.values()])   # [(810,), (40,), (6400,)]
```

radarhere:
Resolves #5838

It is [suggested in that issue](https://github.com/python-pillow/Pillow/issues/5838#issuecomment-972714494) that Pillow should automatically remove tile TIFF tags - rather than having [both tile and strip tags saved to the same image](https://github.com/python-pillow/Pillow/issues/5838#issuecomment-972699979).

So this PR removes TILEWIDTH, TILELENGTH, TILEOFFSETS and TILEBYTECOUNTS.

### DIFF
--- a/Tests/test_file_libtiff.py
+++ b/Tests/test_file_libtiff.py
@@ -920,6 +920,23 @@ class TestFileLibTiff(LibTiffTestCase):
         with Image.open("Tests/images/tiff_strip_planar_16bit_RGBa.tiff") as im:
             assert_image_equal_tofile(im, "Tests/images/tiff_16bit_RGBa_target.png")
 
+    @pytest.mark.parametrize("compression", (None, "jpeg"))
+    def test_block_tile_tags(self, compression, tmp_path):
+        im = hopper()
+        out = str(tmp_path / "temp.tif")
+
+        tags = {
+            TiffImagePlugin.TILEWIDTH: 256,
+            TiffImagePlugin.TILELENGTH: 256,
+            TiffImagePlugin.TILEOFFSETS: 256,
+            TiffImagePlugin.TILEBYTECOUNTS: 256,
+        }
+        im.save(out, exif=tags, compression=compression)
+
+        with Image.open(out) as reloaded:
+            for tag in tags.keys():
+                assert tag not in reloaded.getexif()
+
     def test_old_style_jpeg(self):
         with Image.open("Tests/images/old-style-jpeg-compression.tif") as im:
             assert_image_equal_tofile(im, "Tests/images/old-style-jpeg-compression.png")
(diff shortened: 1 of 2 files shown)


kmilos:
I haven't checked this - so Pillow never ever writes out tiles ATM, even when using libtiff?

LGTM otherwise

kmilos:
I ask because the alternative could be: if you detect tile-based tags, one could force libtiff writing and remove the stripe-based ones instead?

radarhere:
When saving, whether using libtiff or not, [strip tags are always set](https://github.com/python-pillow/Pillow/blob/d6e42f330710ddb854ba56eda7a025b4025fab1e/src/PIL/TiffImagePlugin.py#L1621-L1641).

In encoding TIFF using libtiff, you can see a call to [`TIFFWriteScanline`](https://github.com/python-pillow/Pillow/blob/d6e42f330710ddb854ba56eda7a025b4025fab1e/src/libImaging/TiffDecode.c#L935-L936) ("[It is not possible to write scanlines to a file that uses a tiled organization](https://linux.die.net/man/3/tiffwritescanline)").

And for another opinion, https://github.com/python-pillow/Pillow/issues/672#issue-34124260
> Pillow appears to support reading tiled tiff but not writing.

kmilos:
Indeed, so this PR is a "must" then. 👍

hugovk:
Thanks!

PR #5842, opened 2021-11-19
ImageViewer that allows users to provide commands

m-shinder:
Currently all viewers ignores `command` parameter.
But sometimes it useful to be able to provide specific command.
Especially on unix, where PIL can't guess "which viewer user prefer?"
Writing own viewer for every project where you need custom command is weird, So i think, it should be in PIL by default

 * Add Viewer that rely on provided command.
   This viewer should not be registered now, since Image.show() is throwing `DeprecationWarning` on `command` parameter
   Should warning be removed?

### DIFF
--- a/Tests/test_imageshow.py
+++ b/Tests/test_imageshow.py
@@ -78,7 +78,7 @@ def test_commandviewer():
         basecmd = "rm -f {file}"
     viewer = ImageShow.CommandViewer()
     im = hopper()
-    viewer.get_command(im, command=basecmd)
+    viewer.get_command(viewer.save_image(im), command=basecmd)
 
 
 def test_ipythonviewer():
--- a/Tests/test_imageshow.py
+++ b/Tests/test_imageshow.py
@@ -78,7 +78,8 @@ def test_commandviewer():
         basecmd = "rm -f {file}"
     viewer = ImageShow.CommandViewer()
     im = hopper()
-    assert viewer.show(im, command=basecmd) == 1
+    try:
+        viewer.get_command(im, command=basecmd)
 
 
 def test_ipythonviewer():
--- a/Tests/test_imageshow.py
+++ b/Tests/test_imageshow.py
@@ -78,8 +78,7 @@ def test_commandviewer():
         basecmd = "rm -f {file}"
     viewer = ImageShow.CommandViewer()
     im = hopper()
-    try:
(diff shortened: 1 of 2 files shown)


radarhere:
Hi. I suggested implementing the `command` argument in `im.show()` in #4695. That was declined because it [felt redundant](https://github.com/python-pillow/Pillow/pull/4695#issuecomment-644113271) and because it presented a [potential security problem.](https://github.com/python-pillow/Pillow/pull/4695#issuecomment-643800048)

m-shinder:
Running external programs is always security risk.
But i think, if user writes a command for `CommandViewer` he knows what he is doing.
And since `CommandViewer` not registered, there is no way to run it implicitly.
On my opinion the current approach is not better cause
 * It's just throwing away any utilities that not known by PIL (and it's infeasible to make PIL know about everything)
 * It's not giving ability to choose, I'm using gnome, but have imagemagick installed, 
   and PIL chooses `display` over `eog` just because.
 * If you want to use other program, you should write own Viewer, and you will **have to** run some arbitrary code.

The only difference between current solution and mine, is that mine has a bit easier way to provide command
Also, it's not breaking any of current PIL plans, im.show()'s `command` still can be deprecated,
and CommandViewer can be opaque fallback on systems where no known viewers installed

radarhere:
I think that by building the functionality into Pillow, we are implying that we think it is safe.

https://pillow.readthedocs.io/en/latest/reference/Image.html#PIL.Image.Image.show
> This method is mainly intended for debugging purposes.

We've covered Windows, macOS and for Unix we've covered `display`, `gm`, `eog` and `xv`. That feels like pretty good coverage for a debugging tool. I don't think we need to be aware of every tool - just at least one that is present on your system.

If it's choice you're after, I would think this should work -

```python
from PIL import ImageShow
ImageShow.EogViewer().show(im)
```

That seems simpler to me than.

```python
from PIL import ImageShow
ImageShow.CommandViewer().show(im, command="eog")
```
or
```python
from PIL import ImageShow
ImageShow.register(CommandViewer)
im.show(command="eog")
```

If a user wants arbitrary control, they could even just use `_dump`, which saves the image to a temporary file -
```python
import subprocess
subprocess.Popen(["eog", im._dump()])
```

radarhere:
Closing, as I think this has been considered in #4695

PR #5846, opened 2021-11-21
Updated link to CSS color specification

radarhere:
https://drafts.csswg.org/css-color-4/ can now also be viewed at https://www.w3.org/TR/css-color-4/


hugovk:
The current one is a bit nicer because they have dark theme enabled :)

The colour tables are identical in both.

radarhere:
Ah, so the drafts site has a dark theme, but the live one doesn't? So... the dark theme isn't ready yet? :p

PR #5847, opened 2021-11-21
Documentation for #5829 and #5694

radarhere:
Adds release notes for #5829 and #5694 - as well as updating the documentation for ``polygon`` as a result of #5694.


hugovk:
Thanks!

PR #5861, opened 2021-11-30
GHA: Add workflow_dispatch to enable manually triggering builds

hugovk:

 * This will add a button to allow us to manually trigger a new build for a (whole) workflow, similar to Travis CI
 * https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#manual-events

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,4 +1,5 @@
 name: CIFuzz
+
 on:
   push:
     paths:
@@ -8,6 +9,7 @@ on:
     paths:
       - "**.c"
       - "**.h"
+  workflow_dispatch:
 
 jobs:
   Fuzzing:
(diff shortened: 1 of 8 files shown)


nulano:
Good idea. I would suggest changing the condition in `test-windows.yml` for building wheels from `if: "github.event_name == 'push'"` to `if: "github.event_name != 'pull_request'"` as well as the related conditions throughout the workflow - this would enable anyone to trigger a Windows build and download the wheels for any specific revision in their fork without the GPL imagequant dependency - that seems to me more useful than manually triggering a build with imagequant enabled but not generating wheels.

hugovk:
Updated, how does https://github.com/python-pillow/Pillow/pull/5861/commits/915d014136d769f07654109fab383796ddccf4ed look?

nulano (approved):
Looks good: https://github.com/nulano/Pillow/actions/runs/1533419665

Issue #5697, opened 2021-09-01
Release Pillow 8.3.2 on September 2, 2021


hugovk:
Point release to include Python 3.10 wheels (https://github.com/python-pillow/Pillow/issues/5569) and some regressions (https://github.com/python-pillow/Pillow/issues/5569#issuecomment-907086699).

## Point Release

Released as needed for security, installation or critical bug fixes.

* [x] Make necessary changes in `master` branch.
* [x] Update `CHANGES.rst`.
* [x] Check out release branch e.g.:
  ```bash
  git checkout -t remotes/origin/5.2.x
  ```
* [x] Cherry pick individual commits from `master` branch to release branch e.g. `5.2.x`, then `git push`.
* [ ] Check [GitHub Actions](https://github.com/python-pillow/Pillow/actions) and [AppVeyor](https://ci.appveyor.com/project/python-pillow/Pillow) to confirm passing tests in release branch e.g. `5.2.x`.
* [x] In compliance with [PEP 440](https://www.python.org/dev/peps/pep-0440/), update version identifier in `src/PIL/_version.py`
* [x] Run pre-release check via `make release-test`.
* [x] Create tag for release e.g.:
  ```bash
  git tag 5.2.1
  git push
  git push --tags
  ```
* [x] Create and check source distribution:
  ```bash
  make sdist
  twine check dist/*
  ```
* [x] Create [binary distributions](https://github.com/python-pillow/Pillow/blob/master/RELEASING.md#binary-distributions)
* [x] Check and upload all binaries and source distributions e.g.:
  ```bash
  twine check dist/*
  twine upload dist/Pillow-5.2.1*
  ```
* [x] Publish the [release on GitHub](https://github.com/python-pillow/Pillow/releases)

### Windows
* [x] Contact `@cgohlke` for Windows binaries via release ticket e.g. https://github.com/python-pillow/Pillow/issues/1174.
* [x] Download and extract tarball from `@cgohlke` and copy into `dist/`

### Mac and Linux
* [x] Use the [Pillow Wheel Builder](https://github.com/python-pillow/pillow-wheels):
  ```bash
  git clone https://github.com/python-pillow/pillow-wheels
  cd pillow-wheels
  ./update-pillow-tag.sh [[release tag]]
  ```
* [x] Download wheels from the [Pillow Wheel Builder release](https://github.com/python-pillow/pillow-wheels/releases)
  and copy into `dist/`

## Publicize Release

[...]

PR #5867, opened 2021-12-04
Updated freetype to 2.11.1

radarhere:


hugovk (approved):

radarhere:
This should fix https://github.com/python-pillow/Pillow/issues/5697#issuecomment-911819922

Issue #5795, opened 2021-10-25
Problem with using ImageTk on Cygwin


friedrichromstedt:
While attempting to use `ImageTk` on Cygwin (64-bit, Windows 10, Python 3.8), I ran into an unhandled Exception. Just attempting to show some black square Pillow Image using a tkinter `Canvas` causes this failure. I've tried several versions of Pillow, from `8.4.0` down to `5.4.1`, all with the same result.

Further description of the dysfunction including a minimal example script can be found at https://github.com/friedrichromstedt/bughunting-02. This repo provides a summary along with detailed information on my work regarding the subject.

I've reported to Cygwin upstream https://sourceware.org/pipermail/cygwin/2021-October/249643.html, but up to now this did not lead to a solution. It might be that the problem is rooted somewhere outside of Pillow, but I would appreciate any pointer.

friedrichromstedt:
Any further ideas how to track down this issue?

DWesl:
https://github.com/python-pillow/Pillow/blob/a215f725ba98efeb48194460b8796be3acd6173f/src/_imagingtk.c#L54-L55
and
https://github.com/python-pillow/Pillow/blob/95cff6e959bb3c37848158ed2145d49d49806a31/src/Tk/tkImaging.c#L380-L397
look to be the only places that could fail on import.

```python
import tkinter
tk = tkinter.Tk()
canvas = tkinter.Canvas(tk)
canvas.pack()
photoimage = tkinter.PhotoImage(file="/path/to/image.png")
canvas.create_image((0, 0), image=photoimage, anchor='nw')
tk.mainloop()
```
works fine.

I'm not sure how to check further what's going on beyond this.

An interesting comment &mdash;
https://github.com/python-pillow/Pillow/blob/95cff6e959bb3c37848158ed2145d49d49806a31/src/Tk/tkImaging.c#L224-L230
&mdash; suggests that this is one of the times that "Cygwin uses Windows linking semantics" is relevant: I suspect Pillow is currently trying to import symbols defined in the Tcl/Tk shared libraries from Python's _tkinter shared library.  It would be interesting to have the Cygwin build of Pillow use the Windows code path and see if that fixes the issue.

nulano:
> An interesting comment —

Note that this is in an `#if defined(_WIN32) || ...` block, so this is used on Cygwin (according to [this](http://jdebp.info/FGA/predefined-macros-platform.html#Cygwin), Cygwin defines `_WIN32`).

[...]

PR #5878, opened 2021-12-10
CI: Add a Cygwin run to GitHub Actions

DWesl:
Requested for the _imagingtk fix.

 * Add a Cygwin run to the GitHub Actions Suite


.github/workflows/test-cygwin.yml
```diff
@@ -0,0 +1,53 @@
...
+            python38-wheel python38-pip python38-olefile
+            python38-numpy python38-pytest python38-sphinx
+            python38-packaging python38-cffi
+            libfribidi-devel zlib-devel libX11-xcb-devel tcl-tk-devel
+          install-dir: 'C:\tools\cygwin'
+      - name: Clean up path
+        uses: egor-tensin/cleanup-path@v2
+        with:
+          dirs: 'C:\tools\cygwin\bin;C:\tools\cygwin\lib\lapack'
+      - name: Ensure python dependencies installed
```
radarhere, line 38:
```suggestion
      - name: Ensure Python dependencies are installed
```
Suggestion.

.github/workflows/test-cygwin.yml
```diff
@@ -0,0 +1,53 @@
+name: Test Cygwin
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
```
hugovk, line 9: Let's not limit to `main`, and also add a button to trigger builds via `workflow_dispatch`

```suggestion
on: [push, pull_request, workflow_dispatch]
```

.github/workflows/test-cygwin.yml
```diff
@@ -0,0 +1,53 @@
...
+
+jobs:
+  build:
+    runs-on: windows-2019
+    name: Cygwin Python 3.8
+
+    steps:
+      - name: Checkout Pillow
+        uses: actions/checkout@v2
+      - name: Set up Cygwin
```
hugovk, line 19: Please separate each step with a newline, like the other workflows:
```suggestion

      - name: Set up Cygwin
```

hugovk (commented):
Thanks for the PR!

Please could you also add this to the "Continuous Integration Targets" table in `docs/installation.rst`?

https://pillow.readthedocs.io/en/latest/installation.html#continuous-integration-targets

And not necessary for this PR, but do you think it's possible to include test coverage and upload to Codecov?

.github/workflows/test-cygwin.yml
DWesl: Done

.github/workflows/test-cygwin.yml
DWesl: Done

.github/workflows/test-cygwin.yml
DWesl: Done

.github/workflows/test-cygwin.yml
```diff
@@ -0,0 +1,65 @@
...
+        run: |
+          C:\tools\cygwin\bin\xvfb-run -s '-screen 0 1024x768x24' .ci/test.sh
+
+      - name: After success
+        run: |
+          C:\tools\cygwin\bin\dash .ci/after_success.sh
+
+      - name: Upload coverage
+        run: |
+          C:\tools\cygwin\bin\bash -c 'bash <(curl -s https://codecov.io/bash) -F'
```
hugovk, line 65: The bash uploader is deprecated and will stop working on Feb 1, 2022:

https://docs.codecov.com/docs/about-the-codecov-bash-uploader

Maybe use https://docs.codecov.com/docs/codecov-uploader or https://github.com/codecov/codecov-action?

DWesl: Is the current version better?  The Windows and Linux codecov uploaders will also need to be updated, but that's a separate PR, I think.

DWesl:
Is there some way to tell GitHub Actions "I'm trying to set up a Workflow; only run the Workflows that have actually changed"?

radarhere:
I would create a separate branch to this one (just to keep things neater here) and [remove any workflow files you don't want to run](https://github.com/python-pillow/Pillow/tree/main/.github/workflows). Then when you push, you should only the workflow you want to run at https://github.com/DWesl/Pillow/actions

DWesl:
> I would create a separate branch to this one (just to keep things neater here) and [remove any workflow files you don't want to run](https://github.com/python-pillow/Pillow/tree/main/.github/workflows). Then when you push, you should only the workflow you want to run at https://github.com/DWesl/Pillow/actions

I can do that, though it's *probably* fine now, and will report the two or three ImageTk errors from #5795 just fine.  If not, off I go to a new branch to tinker with things.

docs/installation.rst
```diff
@@ -460,6 +460,8 @@ These platforms are built and tested for every change.
 +----------------------------------+----------------------------+---------------------+
 | CentOS Stream 8                  | 3.9                        | x86-64              |
 +----------------------------------+----------------------------+---------------------+
+| Cygwin                           | 3.8                        | x86-64              |
```
kmilos, line 463: Could move this under Windows 2019 below, like there is a separate line for MinGW?

radarhere:
I've created https://github.com/DWesl/Pillow/pull/1 with some suggestions.

.ci/test.sh
```diff
@@ -2,6 +2,6 @@
 
 set -e
 
-python3 -c "from PIL import Image"
+python3$1 -c "from PIL import Image"
 
-python3 -bb -m pytest -v -x -W always --cov PIL --cov Tests --cov-report term Tests $REVERSE
+python3$1 -bb -m pytest -v -x -W always --cov PIL --cov Tests --cov-report term Tests $REVERSE
```
hugovk, line 7: What's this `$1` for?

radarhere: In [an earlier commit](https://github.com/python-pillow/Pillow/pull/5878/commits/7fd3e9977e7dd0c04e4e72fdc65f59253f0e6281) in this PR, it was used to explicitly call Python with the Python minor version (`python3.8`).
My contribution to the PR removed this.
I think it's only back by accident.

DWesl: An early CI run was failing due to one of the dependencies pulling in python3.9 by accident (so python3 was python3.9, with almost no requirements installed, instead of python3.8, with most dependencies installed.  The dependencies got resolved a while back, so this can go away.

DWesl:
I get four of those ghostscript errors when running locally.  This appears to have broken between three and two commits back, around February 7.  The [list of Cygwin package updates](https://cygwin.com/pipermail/cygwin-announce/2022-February/date.html) from around that time does not include ghostscript, just [lcms2](https://cygwin.com/pipermail/cygwin-announce/2022-February/010455.html) and [findutils](https://cygwin.com/pipermail/cygwin-announce/2022-February/010452.html).

Out of curiosity, how many of the other CI runs are using [lcms2 2.13, released in late January](https://littlecms.com/blog/2022/01/28/lcms2-2.13/)?

radarhere:
At least the windows Ci jobs are running lcms2 2.13.1. See #6017

DWesl:
It looks like MacOS is as well.  So it's something else.

nulano:
> It looks like MacOS is as well. So it's something else. It doesn't look like Windows or MacOS have ghostscript installed, so some interaction between ghostscript and recent lcms?

IIRC it is an issue in LCMS 2.13: https://github.com/python-pillow/Pillow/issues/6013#issuecomment-1039609802

DWesl:
Merge reverts [5588f57](https://github.com/python-pillow/Pillow/pull/5878/commits/5588f572de4cd386f576fc761e50d373cf1ca239).  

Cygwin recently updated [liblcms2 to 2.13.1](https://cygwin.com/pipermail/cygwin/2022-April/251256.html). so let's see if the ghostscript failures go away.

.github/workflows/test-cygwin.yml
```diff
@@ -0,0 +1,104 @@
...
+        run: |
+          bash.exe xvfb-run -s '-screen 0 1024x768x24' .ci/test.sh
+
+      - name: Prepare to upload errors
+        if: failure()
+        run: |
+          dash.exe -c "mkdir -p Tests/errors"
+
+      - name: Upload errors
+        uses: actions/upload-artifact@v2
```
radarhere, line 81:
```suggestion
        uses: actions/upload-artifact@v3
```

.github/workflows/test-cygwin.yml
```diff
@@ -0,0 +1,104 @@
...
+        with:
+          name: errors
+          path: Tests/errors
+
+      - name: After success
+        run: |
+          bash.exe .ci/after_success.sh
+
+      - name: Upload coverage
+        uses: codecov/codecov-action@v2
```
radarhere, line 92:
```suggestion
        uses: codecov/codecov-action@v3
```

.github/workflows/test-cygwin.yml
```diff
@@ -0,0 +1,104 @@
...
+
+    name: Python 3.${{ matrix.python-minor-version }} ${{ matrix.architecture }}
+
+    steps:
+      - name: Fix line endings
+        run: |
+          git config --global core.autocrlf input
+
+      - name: Checkout Pillow
+        uses: actions/checkout@v2
```
radarhere, line 24:
```suggestion
        uses: actions/checkout@v3
```

.github/workflows/test-cygwin.yml
```diff
@@ -0,0 +1,104 @@
...
+
+      - name: Install Cygwin
+        uses: cygwin/cygwin-install-action@v1
```
radarhere, line 27:
```suggestion
        uses: cygwin/cygwin-install-action@v2
```

DWesl:
```
    error: command 'gcc' terminated by signal 11
ERROR: Command errored out with exit status 1.
```
`kill -l` says 11 is `SIGSEGV`, which I think implies this is a segfault in GCC.  I don't know how to debug that.  The last time I ran into segfaults was with unaligned vector loads, but I'm not sure where in 32-bit GCC that would be relevant.

hugovk:
The x86 jobs don't install NumPy and take around 10 minutes.

The x86_64 jobs do install NumPy and take around 27 minutes, of which about 14 minutes is installing NumPy, and it's building wheels from source each time.

Are there any wheels available for older NumPy versions that could be installed using `--prefer-binary`?

And (especially if there's no wheels on PyPI) can we cache the pip dir using `actions/cache` to avoid rebuilding from source every time? See https://github.com/actions/cache/blob/main/examples.md#python---pip

.github/workflows/test-cygwin.yml
```diff
@@ -0,0 +1,118 @@
...
+          
+      - name: Get pip cache dir
+        id: pip-cache
+        run: |
+          bash.exe -c 'echo "::set-output name=dir::$(python -m pip cache dir)"'
+          
+      - name: pip cache
+        uses: actions/cache@v3
+        with:
+          path: ${{ steps.pip-cache.outputs.dir }}
```
hugovk, line 55: I expect this directory is constant, so let's put the string here and remove the previous step.

.github/workflows/test-cygwin.yml
```diff
@@ -0,0 +1,118 @@
...
+          key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
+          restore-keys: |
+            ${{ runner.os }}-pip-
```
hugovk, line 58: Let's include `cygwin` just to make sure it doesn't clash with another workflow.

```suggestion
          key: ${{ runner.os }}-cygwin-pip-${{ hashFiles('**/requirements.txt') }}
          restore-keys: |
            ${{ runner.os }}-cygwin-pip-
```

.github/workflows/test-cygwin.yml
```diff
@@ -0,0 +1,118 @@
...
+      - name: Get pip cache dir
+        id: pip-cache
+        run: |
+          bash.exe -c 'echo "::set-output name=dir::$(python -m pip cache dir)"'
+          
+      - name: pip cache
+        uses: actions/cache@v3
+        with:
+          path: ${{ steps.pip-cache.outputs.dir }}
+          key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
```
hugovk, line 56: We don't use `requirements.txt`. Instead, can pick something that has the majority of the dependencies, like `.ci/install.sh` or this yml file?

DWesl: `.ci/install.sh` has all the `pip` dependencies; adding the workflow file would be useful for caching the Cygwin installation as well (that would also need architecture and python minor version in the key, but that's doable).

.github/workflows/test-cygwin.yml
DWesl: Constant except for user name, which is hopefully also constant.

.github/workflows/test-cygwin.yml
```diff
@@ -0,0 +1,119 @@
...
+          bash.exe -c 'cygpath -wa $(python -m pip cache dir)'
+          bash.exe -c 'echo "::set-output name=dir::$(cygpath -wa $(python -m pip cache dir))"'
+          
+      - name: pip cache
+        uses: actions/cache@v3
+        with:
+          path: 'C:\cygwin\home\runneradmin\.cache\pip'
+          key: ${{ runner.os }}-cygwin-pip-${{ hashFiles('.ci/install.sh') }}
+          restore-keys: |
+            ${{ runner.os }}-cygwin-pip-
```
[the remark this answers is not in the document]
DWesl: The wheels can also be used only for the same Cygwin patch version.  I'm not sure how to represent that in the key here.

.github/workflows/test-cygwin.yml
```diff
@@ -0,0 +1,113 @@
...
+        run: |
+          bash.exe -c "python3.${{ matrix.python-minor-version }} -m pip install -U 'numpy!=1.21.*'"
+
+      - name: Check imports
+        run: |
+          bash.exe -c "python3.${{ matrix.python-minor-version }} -c 'import numpy as np; print(np.__version__)'"
+
+      - name: Build
+        run: |
+          bash.exe .ci/build_cygwin.sh
```
nulano, line 74: Can avoid needing a separate file by passing the shell parameter directly.

```suggestion
        run: |
          python3 -m coverage erase
          make clean
          CFLAGS="-coverage -Werror=implicit-function-declaration" python3 -m pip install -v --global-option="build_ext" .
          python3 selftest.py
        shell: bash.exe --login -eo pipefail "{0}"
```

.github/workflows/test-cygwin.yml
```diff
@@ -0,0 +1,113 @@
...
+            ImageMagick jpeg python3${{ matrix.python-minor-version }}-cffi
+            python3${{ matrix.python-minor-version }}-devel
+            python3${{ matrix.python-minor-version }}-numpy
+            python3${{ matrix.python-minor-version }}-cython liblapack-devel gcc-g++
+            python3${{ matrix.python-minor-version }}-sip
+            python3${{ matrix.python-minor-version }}-tkinter ghostscript
+            libfreetype-devel libimagequant-devel libjpeg-devel liblcms2-devel
+            libopenjp2-devel libraqm-devel libtiff-devel libwebp-devel libxcb-devel
+            libxcb-xinerama0 netpbm perl qt5-devel-tools xorg-server-extra zlib-devel
+            subversion make
```
nulano, line 40: Put in alphabetical order?

```suggestion
            ghostscript gcc-g++ ImageMagick jpeg make
            python3${{ matrix.python-minor-version }}-cffi
            python3${{ matrix.python-minor-version }}-devel
            python3${{ matrix.python-minor-version }}-numpy
            python3${{ matrix.python-minor-version }}-cython
            python3${{ matrix.python-minor-version }}-sip
            python3${{ matrix.python-minor-version }}-tkinter
            libfreetype-devel libimagequant-devel libjpeg-devel liblapack-devel
            liblcms2-devel libopenjp2-devel libraqm-devel libtiff-devel
            libwebp-devel libxcb-devel libxcb-xinerama0 netpbm perl
            qt5-devel-tools xorg-server-extra zlib-devel subversion
```

.github/workflows/test-cygwin.yml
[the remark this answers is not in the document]
hugovk: That's okay, we'll get a new wheel built when the new patch version comes along and they can both sit in the cache, and the cache will get cleared eventually.

---

Cache is looking good now, thanks!

First time, no cache yet, did a 16m build of NumPy, and uploaded the cache:

https://github.com/python-pillow/Pillow/runs/6128416734?check_suite_focus=true

I restarted the build, it fetched the cache, and took just 11s to install the pre-build wheel:

https://github.com/python-pillow/Pillow/runs/6134328119?check_suite_focus=true

👍

.github/workflows/test-cygwin.yml
```diff
@@ -0,0 +1,113 @@
...
+          restore-keys: |
+            ${{ runner.os }}-cygwin-${{ matrix.architecture }}-pip3.${{ matrix.python-minor-version }}-
+
+      - name: Build system information
+        run: |
+          bash.exe -c "python3 .github/workflows/system-info.py"
+
+      - name: Install dependencies
+        run: |
+          bash.exe .ci/install.sh
```
hugovk, line 61: Let's silence these warnings:

```
WARNING: You are using pip version 21.3.1; however, version 22.0.4 is available.
You should consider upgrading via the '/usr/bin/python3 -m pip install --upgrade pip' command.
```

```suggestion
          bash.exe -c "python3.${{ matrix.python-minor-version }} -m pip install -U pip"
          bash.exe .ci/install.sh
```

.github/workflows/test-cygwin.yml
nulano: (I see you committed a change for this - feel free to mark as resolved. The button does not show up for me.)

radarhere: The button shows up for me!

nulano: AFAIK it shows up for team members and PR author, but for some reason it does not show for the author of the comment.

.github/workflows/test-cygwin.yml
```diff
@@ -0,0 +1,124 @@
...
+      - name: Install a different NumPy
+        if: matrix.architecture == 'x86_64'
+        shell: dash.exe -l "{0}"
+        run: |
+          python3.${{ matrix.python-minor-version }} -m pip install -U 'numpy!=1.21.*'
+
+      - name: Check imports
+        shell: dash.exe -l "{0}"
+        run: |
+          python3.${{ matrix.python-minor-version }} -c 'import numpy as np; print(np.__version__)'
```
radarhere, line 76:
```suggestion
          python3 -c 'import numpy as np; print(np.__version__)'
```

.github/workflows/test-cygwin.yml
```diff
@@ -0,0 +1,124 @@
...
+            ${{ runner.os }}-cygwin-${{ matrix.architecture }}-pip3.${{ matrix.python-minor-version }}-
+
+      - name: Build system information
+        shell: bash.exe --login -eo pipefail "{0}"
+        run: |
+          python3 .github/workflows/system-info.py
+
+      - name: Install dependencies
+        run: |
+          dash.exe -c "python3.${{ matrix.python-minor-version }} -m pip install -U pip"
```
radarhere, line 64:
```suggestion
          dash.exe -c "python3 -m pip install -U pip"
```

.github/workflows/test-cygwin.yml
```diff
@@ -0,0 +1,124 @@
...
+          bash.exe .ci/install.sh
+
+      - name: Install a different NumPy
+        if: matrix.architecture == 'x86_64'
+        shell: dash.exe -l "{0}"
+        run: |
+          python3.${{ matrix.python-minor-version }} -m pip install -U 'numpy!=1.21.*'
```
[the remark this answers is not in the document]
DWesl: Simpler, but occasionally there's packaging problems pulling in a different python.  I have no idea how often this happens, since I just install all python versions available, but it's caused problems once on the NumPy CI in the six months to a year since they added a Cygwin run.  Are you okay with that risk?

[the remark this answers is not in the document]
radarhere: Ok, we can keep your version if you feel it is more stable.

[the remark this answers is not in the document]
DWesl: I changed the other references to `python3` to reflect that choice.  The other option is to tell alternatives to make sure it point python3 at the correct version of python, but I don't know how to do that.

DWesl:
Another GCC failure with signal 11, which I have no idea how to deal with.  DWesl/pillow#5 assumes this is due to a `fork()` failure, which causes python to misplace the `coverage` module somehow. DWesl/pillow#4 is attempting something else, but also causes python to forget where it put the `coverage` module.

radarhere:
From what I can see, pip has just stopped working in Cygwin. [Simplifying the GitHub Action](https://github.com/radarhere/Pillow/blob/pip/.github/workflows/test-cygwin.yml), you can see that even that basic example [doesn't work.](https://github.com/radarhere/Pillow/runs/6239589920?check_suite_focus=true) It doesn't look like it's particular to coverage either.

DWesl:
That is a really weird problem, and not one I can reproduce locally.

DWesl:
32-bit Cygwin Python 3.8 seems to have started failing between 3453b0e and 143e57b.  Specifying precisely which version of python to use shouldn't cause problems, and it definitely shouldn't cause a GCC crash.

[There are Cygwin updates around that time](https://cygwin.com/pipermail/cygwin-announce/2022-April/date.html#10539), but `neomutt`, `perl`, and `pcre2` seem unlikely to cause GCC to fail with signal 11.

DWesl/pillow#5 assumes this is some kind of new `fork` failure (it doesn't look like the usual ones), but the usual fix for that doesn't seem to be working.

Oh, and now the failure has migrated to Python 3.9, and 3.8 works fine.  Is this a new manifestation of the old "[sometimes Windows decides it hates how Cygwin works](https://cygwin.com/faq.html#faq.using.fixing-fork-failures)" problems, or is this another bug that depends on the architecture of the CI runner?

This is such a weirdly specific failure, and reproduced somewhat consistently whenever the workflow managed to get that far, but there's no apparent cause in the code, nor in the part of the environment I can control.  The change in the length of the command line, or of the template generating the file that the CI runner actually runs, is about the only thing similar between the start and end of the run of failures; however, I saw similar failures on the test branch where I ironed out the syntax for alternatives.  None of this should be causing *any* problems, but any problems I can think of should be producing at least three failures, not merely one.  I have no idea how to go about debugging this.

The best idea I have is that roughly a third of the 32-bit Windows GitHub Actions runners are older machines than Cygwin's GCC maintainer assumed when compiling 11.2, but I don't think there's been *that* many updates to 32-bit processors recently, and I don't know if there's a way to convince [cygwin-install-action](https://github.com/cygwin/cygwin-install-action) to install old versions of GCC to test the theory.

hugovk:
Would it make sense to only test 64-bit Cygwin?

Or at least to get this PR merged, and 32-bit could be investigated in a new PR?

DWesl:
Given that [the Cygwin website recommends not using 32-bit Cygwin if possible](https://cygwin.com/), and that [the Cygwin 3.3.0 announcement said 3.4.0 will not include a 32-bit component](https://cygwin.com/pipermail/cygwin-announce/2021-October/010268.html), it would probably make sense to omit regular testing for 32-bit Cygwin.  Should I update the matrix?

hugovk:
That's quite a warning!

32 bit Cygwin: 
Address space is a very limiting factor for Cygwin. These days, a full 32 bit Cygwin distro is not feasible anymore, and will in all likelihood fail in random places due to an issue with the fork(2) system call. Therefore we recommend using 32 bit Cygwin only in limited scenarios, with only a minimum of necessary packages installed, and only if there's no way to run 64 bit Cygwin instead. You have been warned. If you're still sure you really need a 32 bit Cygwin, and there's absolutely no way around it, you may run the [setup-x86.exe installer. The [signature](https://cygwin.com/setup-x86.exe.sig) can be used to verify the validity of this binary.](https://user-images.githubusercontent.com/1324225/166678247-7e4acd81-ffd5-4238-a56a-ee0b36ad8e32.png)

Yeah, let's just ditch 32 bit.

The community is gradually dropping support for 32 bit anyway, for example see [NumPy](https://mail.python.org/archives/list/numpy-discussion@python.org/thread/6B3VJZ3I5JV4PFJUSRPTN4ABZE7UPAEI/#3O37HETOD637FIH5FWR53QP3ROWHAXHE).

.ci/install.sh
```diff
@@ -13,17 +13,23 @@ aptget_update()
...
-sudo apt-get -qq install libfreetype6-dev liblcms2-dev python3-tk\
-                         ghostscript libffi-dev libjpeg-turbo-progs libopenjp2-7-dev\
-                         cmake meson imagemagick libharfbuzz-dev libfribidi-dev
+if [[ $(uname) != CYGWIN* ]]; then
+    sudo apt-get -qq install libfreetype6-dev liblcms2-dev python3-tk\
+                             ghostscript libffi-dev libjpeg-turbo-progs libopenjp2-7-dev\
+                             cmake meson imagemagick libharfbuzz-dev libfribidi-dev
+fi
+
+if [[ $(uname -mo) != i*86" Cygwin" ]]; then
```
hugovk, line 28: Is this a 32-bit check? If so, let's revert it.

.github/workflows/test-cygwin.yml
```diff
@@ -0,0 +1,130 @@
...
+    runs-on: windows-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-minor-version: [7, 8, 9]
+        architecture: ["x86_64"]
+
+    timeout-minutes: 40
+
+    name: Python 3.${{ matrix.python-minor-version }} ${{ matrix.architecture }}
```
hugovk, line 16:
```suggestion
    name: Python 3.${{ matrix.python-minor-version }}
```

.github/workflows/test-cygwin.yml
```diff
@@ -0,0 +1,130 @@
...
+on: [push, pull_request, workflow_dispatch]
+
+jobs:
+  build:
+    runs-on: windows-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-minor-version: [7, 8, 9]
+        architecture: ["x86_64"]
```
hugovk, line 12:
```suggestion
```

.github/workflows/test-cygwin.yml
```diff
@@ -0,0 +1,130 @@
...
+        run: |
+          git config --global core.autocrlf input
+
+      - name: Checkout Pillow
+        uses: actions/checkout@v3
+
+      - name: Install Cygwin
+        uses: cygwin/cygwin-install-action@v2
+        with:
+          platform: ${{ matrix.architecture }}
```
hugovk, line 29:
```suggestion
          platform: x86_64
```

.github/workflows/test-cygwin.yml
```diff
@@ -0,0 +1,130 @@
...
+        with:
+          dirs: 'C:\cygwin\bin;C:\cygwin\lib\lapack'
+          
+      - name: pip cache
+        uses: actions/cache@v3
+        with:
+          path: 'C:\cygwin\home\runneradmin\.cache\pip'
+          key: ${{ runner.os }}-cygwin-${{ matrix.architecture }}-pip3.${{ matrix.python-minor-version }}-${{ hashFiles('.ci/install.sh') }}
+          restore-keys: |
+            ${{ runner.os }}-cygwin-${{ matrix.architecture }}-pip3.${{ matrix.python-minor-version }}-
```
hugovk, line 55:
```suggestion
          key: ${{ runner.os }}-cygwin-pip3.${{ matrix.python-minor-version }}-${{ hashFiles('.ci/install.sh') }}
          restore-keys: |
            ${{ runner.os }}-cygwin-pip3.${{ matrix.python-minor-version }}-
```

.github/workflows/test-cygwin.yml
```diff
@@ -0,0 +1,130 @@
...
+      - name: Build system information
+        run: |
+          dash.exe -c "python3 .github/workflows/system-info.py"
+
+      - name: Install dependencies
+        run: |
+          bash.exe .ci/install.sh
+
+      - name: Install a different NumPy
+        if: matrix.architecture == 'x86_64'
```
hugovk, line 74:
```suggestion
```

.github/workflows/test-cygwin.yml
```diff
@@ -0,0 +1,130 @@
...
+      - name: After success
+        run: |
+          bash.exe .ci/after_success.sh
+
+      - name: Upload coverage
+        uses: codecov/codecov-action@v3
+        with:
+          file: ./coverage.xml
+          flags: GHA_Cygwin
+          name: Cygwin Python 3.${{ matrix.python-minor-version }} ${{ matrix.architecture }}
```
hugovk, line 122:
```suggestion
          name: Cygwin Python 3.${{ matrix.python-minor-version }}
```

docs/installation.rst
```diff
@@ -483,7 +483,9 @@ These platforms are built and tested for every change.
 | Windows Server 2022              | 3.7, 3.8, 3.9, 3.10, 3.11, | x86, x86-64         |
 |                                  | PyPy3                      |                     |
 |                                  +----------------------------+---------------------+
-|                                  | 3.9/MinGW                  | x86, x86-64         |
+|                                  | 3.9 (MinGW)                | x86, x86-64         |
+|                                  +----------------------------+---------------------+
+|                                  | 3.7, 3.8, 3.9 (Cygwin)     | x86, x86-64         |
```
hugovk, line 488:
```suggestion
|                                  | 3.7, 3.8, 3.9 (Cygwin)     | x86-64              |
```

hugovk (changes requested):

hugovk (approved):
Thank you!

.github/workflows/test-cygwin.yml
```diff
@@ -0,0 +1,128 @@
...
+          
+      - name: pip cache
+        uses: actions/cache@v3
+        with:
+          path: 'C:\cygwin\home\runneradmin\.cache\pip'
+          key: ${{ runner.os }}-cygwin-pip3.${{ matrix.python-minor-version }}-${{ hashFiles('.ci/install.sh') }}
+          restore-keys: |
+            ${{ runner.os }}-cygwin-pip3.${{ matrix.python-minor-version }}-
+            
+      - name: Ensure correct python minor version used in scripts
```
radarhere, line 56:
```suggestion
      - name: Ensure correct Python minor version is used in scripts
```

.github/workflows/test-cygwin.yml
```diff
@@ -0,0 +1,128 @@
...
+      - name: Install a different NumPy
+        shell: dash.exe -l "{0}"
+        run: |
+          python3 -m pip install -U 'numpy!=1.21.*'
+
+      - name: Check imports
+        shell: dash.exe -l "{0}"
+        run: |
+          python3 -c 'import numpy as np; print(np.__version__)'
+
```
radarhere, line 81:
```suggestion

```

You're free to leave this in, but my feeling is that this was put in here for debugging purposes when there were problems with NumPy?

.github/workflows/test-cygwin.yml
```diff
@@ -0,0 +1,107 @@
...
+      - name: Install a different NumPy
+        shell: dash.exe -l "{0}"
+        run: |
+          python3 -m pip install -U 'numpy!=1.21.*'
+
+      - name: Build
+        shell: bash.exe -eo pipefail -o igncr "{0}"
+        run: |
+          bash.exe .ci/build.sh
```
nulano, line 72:
```suggestion
        run: |
          bash.exe -eo pipefail -o igncr .ci/build.sh
```
Probably not necessary to invoke bash twice?
By omitting the shell param, it will be invoked via PowerShell, but at lesat it seems a little easier to read.

DWesl: Alternately, we could do [like the Linux tests](https://github.com/python-pillow/Pillow/blob/9c580ce64902b23579e04a7890ba309438c152b2/.github/workflows/test.yml#L66-L68):
```suggestion
        shell: bash.exe -eo pipefail -o igncr "{0}"
        run: |
          .ci/build.sh
```
which may run into trouble if git doesn't set up the permissions properly.  Do you have a preference as to which goes in?

nulano: Your suggestion looks like the better option to me (assuming it works as expected).

radarhere (approved):

hugovk:
@DWesl Thank you for your work on this, and everyone else for the reviews!

PR #5884, opened 2021-12-13
unnecessary values

### LINKED ISSUE #5877: Pillow is saving unnecessary values in paletted pngs
Pillow is currently saving unnecessarily long palette information even in case we are saving just 2color png. 

### What did you do?

Pillow is saving all 256 colors in palette even if only 2 colors are defined thus increasing the size of the pngs significantly. 
We use paletted pngs to save space and bandwidth but we have to manually remove palette from all pngs saved by Pillow. (In old version we would simply remove img.palette.palette altogether and png would be saved properly. In new versions it's no longer possible so we have to cut it to correct size)
Python code provided should show exact way how palettes are being increased in size by this issue.

I'd expect that output png would save only palette required. (2colors instead of 256colors)

### What actually happened?

Pillow keeps saving unnecessary data that increases size of the paletted output pngs.

* OS: Ubuntu 20
* Python: 3.9.1
* Pillow: Pillow==8.4.0

```python

img = Image.new("RGB", (64, 64), (255,255,255))
# get a drawing context
draw = ImageDraw.Draw(img)
draw.line((0, 0) + img.size, fill=(0,255,0))
img.save('line.png', "PNG")  # Size 354 bytes
palette_img = img.quantize(colors=2)
palette_img.save('line_palette.png', "PNG")  # Size 878 bytes
# Saving with adaptive and optimize doesn't solve bloated png output:
pil_adaptive = img.convert("P", palette=Image.ADAPTIVE, colors=2)
pil_adaptive.save('line_palette_adaptive.png', "PNG", optimize=True)  # Size 878 bytes
# Remove all unneeded balast from pallete output, keeping only 6 bytes needed
# for pallete. (2x 3 rgb values)
palette_img.palette.palette = palette_img.palette.palette[:6]
palette_img.save('line_palette_no_balast.png', "PNG")  # Size 113 bytes
```

kero-ly:
Fixes #5877 .

Pillow is saving all 256 colors in palette even if only 2 colors are defined thus increasing the size of the pngs significantly.
We use paletted pngs to save space and bandwidth but we have to manually remove palette from all pngs saved by Pillow. (In old version we would simply remove img.palette.palette altogether and png would be saved properly. In new versions it's no longer possible so we have to cut it to correct size)
Python code provided should show exact way how palettes are being increased in size by this issue.
 * 
 * 
 *

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -1111,7 +1111,8 @@ class Image:
         from . import ImagePalette
 
         mode = im.im.getpalettemode()
-        im.palette = ImagePalette.ImagePalette(mode, im.im.getpalette(mode, mode))
+        palette = im.im.getpalette(mode, mode)[: colors * len(mode)]
+        im.palette = ImagePalette.ImagePalette(mode, palette)
 
         return im
 


hugovk:
Closing: this is a duplicate of PR https://github.com/python-pillow/Pillow/pull/5879, but without tests.

PR #5897, opened 2021-12-20
Added ImageShow support for xdg-open

m-shinder:
`xdg-open` is reliable way to open image with preferred application.
It works on linux and BSD with any desktop environment.

Debugging process should be comfortable.


radarhere (approved):

hugovk:
This is registered first and so will be used first. 

Is it better than the other viewers?

radarhere:
It's not a viewer in the normal sense - it is instructing the system to open the file with whatever program it thinks is appropriate. I suppose the macOS equivalent is 'open'.

Issue #3927, opened 2019-06-30
Test suite fails 16 tests on hppa


DerDakon:
### What did you do?
Built and run Pillow master (290cd170eccd52b285cfbe379b5915a2d46281ee) on Gentoo/HPPA (big endian!) and ran the test suite:

### What actually happened?
Tests/test_file_gif.py::TestFileGif::test_extents FAILED 
Tests/test_file_ico.py::TestFileIco::test_unexpected_size FAILED
Tests/test_file_jpeg.py::TestFileJpeg::test_photoshop FAILED
Tests/test_file_libtiff.py::TestFileLibTiff::test_16bit_RGB_tiff FAILED
Tests/test_file_libtiff.py::TestFileLibTiff::test_old_style_jpeg FAILED
Tests/test_file_libtiff.py::TestFileLibTiff::test_strip_cmyk_16l_jpeg FAILED
Tests/test_file_libtiff.py::TestFileLibTiff::test_strip_ycbcr_jpeg_1x1_sampling FAILED
Tests/test_file_libtiff.py::TestFileLibTiff::test_strip_ycbcr_jpeg_2x2_sampling FAILED
Tests/test_file_libtiff.py::TestFileLibTiff::test_tiled_ycbcr_jpeg_1x1_sampling FAILED
Tests/test_file_libtiff.py::TestFileLibTiff::test_tiled_ycbcr_jpeg_2x2_sampling FAILED
Tests/test_file_png.py::TestFilePng::test_sanity FAILED
Tests/test_image_getextrema.py::TestImageGetExtrema::test_extrema FAILED
Tests/test_image_getextrema.py::TestImageGetExtrema::test_true_16 FAILED
Tests/test_imagedraw.py::TestImageDraw::test_arc_width_pieslice_large FAILED
Tests/test_imagedraw.py::TestImageDraw::test_ellipse_width_large FAILED
Tests/test_imagedraw.py::TestImageDraw::test_rectangle_I16 FAILED

Not all of those are regressions, here is the fail list for 6.0.0:

Tests/test_file_libtiff.py::TestFileLibTiff::test_old_style_jpeg FAILED
Tests/test_file_libtiff.py::TestFileLibTiff::test_strip_ycbcr_jpeg_1x1_sampling FAILED
Tests/test_file_libtiff.py::TestFileLibTiff::test_strip_ycbcr_jpeg_2x2_sampling FAILED
[...]

PR #5900, opened 2021-12-21
Fixed I;16 conversion on big endian

radarhere:
Helps #3927 and #4213 - reverting part of #4457


radarhere:
This was replaced by #5901

PR #5920, opened 2022-01-01
Fixed ImagePath.Path array handling

radarhere:


radarhere:
CIFuzz failure is also occurring in main.

radarhere:
CIFuzz failure is now fixed thanks to https://github.com/python-pillow/Pillow/commit/525c840b0f8334118ae9a032cc7be2bffd4b8d46

PR #5923, opened 2022-01-02
Restrict builtins for ImageMath.eval()

radarhere:
To limit `ImageMath` to working with images, Pillow will now restrict the builtins available to `ImageMath.eval()`. This will help prevent problems arising if users evaluate arbitrary expressions, such as ``ImageMath.eval("exec(exit())")``.

PR #5924, opened 2022-01-02
CVEs TBD

radarhere:
Noting that CVEs are TBD for #5920 and #5923


hugovk (approved):

radarhere:
#5940 is the follow up to this.

PR #5935, opened 2022-01-06
Add missing ImageModes

### LINKED ISSUE #5933: `BGR;[15,16,34,32]` modes are documented but not part of `ImageMode`
I found a few image modes that are listed as partially supported in the docs ([here](https://pillow.readthedocs.io/en/stable/handbook/concepts.html)), but that are not listed in `ImageMode`:

```python
from PIL import ImageMode

# note: I'm using the private variable here because it supports `in`
# and leads to nicer code
# afaik correct access is done via ImageMode.getmode(mode)

assert "BGR;15" in ImageMode._modes
assert "BGR;16" in ImageMode._modes
assert "BGR;24" in ImageMode._modes
assert "BGR;32" in ImageMode._modes
```

Is there any reason why these have not been added? If not, would a PR adding those be welcome?

FirefoxMetzger:
Fixes #5933

This PR adds the modes mentioned in the docs to ImageMode. It also adds the mode `BGR` to be used as a basetype for the missing modes.

---

Sidenote: To get pillow 9.1.0.dev0 to build on Windows I had to do some modifications to the script in `winbuild\README.md`, e.g., because the script/snippet builds for python3.8, but calls python3.7 to prepare - which I don't have. I will leave my modified script here, since I don't know where else to put it, but if there is interest in a dedicated DOC PR for this, I can do that, too:

```cmd
set PYTHON=C:\Users\<username>\AppData\Local\Programs\Python\Python38
set PILLOW_REPO=<git-clone-location>\Pillow

cd /D %PILLOW_REPO%\winbuild
%PYTHON%\python.exe build_prepare.py -v --depends=%PILLOW_REPO%\depends
build\build_dep_all.cmd
build\build_pillow.cmd install

path %PILLOW_REPO%\winbuild\build\bin;%PATH%
%PYTHON%\python.exe -m pip install -U -e .
%PYTHON%\python.exe selftest.py
%PYTHON%\python.exe -m pytest -vx --cov PIL --cov Tests --cov-report term --cov-report xml Tests
%PILLOW_REPO%\winbuild\build\build_pillow.cmd bdist_wheel
```

Sidenote2: I didn't manage to get the full test suite to pass after building pillow. One test (`Tests/test_image_access.py::TestEmbeddable::test_embeddable`) keeps failing. However, I think this is because the test can't find `python38.lib` on my system rather than something related to the PR.

### DIFF
--- a/src/PIL/ImageMode.py
+++ b/src/PIL/ImageMode.py
@@ -52,6 +52,11 @@ def getmode(mode):
             "HSV": ("RGB", "L", ("H", "S", "V")),
             # extra experimental modes
             "RGBa": ("RGB", "L", ("R", "G", "B", "a")),
+            "BGR": ("BGR", "L", ("B", "G", "R")),
+            "BGR;15": ("RGB", "L", ("B", "G", "R")),
+            "BGR;16": ("RGB", "L", ("B", "G", "R")),
+            "BGR;24": ("RGB", "L", ("B", "G", "R")),
+            "BGR;32": ("RGB", "L", ("B", "G", "R")),
             "LA": ("L", "L", ("L", "A")),
             "La": ("L", "L", ("L", "a")),
             "PA": ("RGB", "L", ("P", "A")),
--- a/src/PIL/ImageMode.py
+++ b/src/PIL/ImageMode.py
@@ -52,7 +52,6 @@ def getmode(mode):
             "HSV": ("RGB", "L", ("H", "S", "V")),
             # extra experimental modes
             "RGBa": ("RGB", "L", ("R", "G", "B", "a")),
-            "BGR": ("BGR", "L", ("B", "G", "R")),
             "BGR;15": ("RGB", "L", ("B", "G", "R")),
             "BGR;16": ("RGB", "L", ("B", "G", "R")),
             "BGR;24": ("RGB", "L", ("B", "G", "R")),


nulano:
Did you read the instructions for the Windows build, or just the sample script? I think most of your issues with it are explained in the longer documentation: https://github.com/python-pillow/Pillow/blob/main/winbuild/build.rst

For example, you do not need to pass the `--depends=...` parameter, and you don't need to `pip install -e .` if you `build_pillow.cmd --inplace develop` instead of `build_pillow.cmd install`.

> One test (Tests/test_image_access.py::TestEmbeddable::test_embeddable) keeps failing.

A known issue, sort of, https://github.com/python-pillow/Pillow/pull/5811#issuecomment-965363094.

FirefoxMetzger:
> Did you read the instructions for the Windows build, or just the sample script?

@nulano Neither, because I missed the full instructions. I found [Readme.md](https://github.com/python-pillow/Pillow/blob/main/winbuild/README.md) and it appeared to get me close enough, so I didn't follow the links therein. The full instructions indeed make the script cleaner and I should have read them; my bad.

If you don't think it is worthwhile to change or add to the existing example then let's not worry about it. That's why I put it as a side note to begin with; mainly to see if it is useful and/or if there is any interest in it :)

> A known issue, sort of, #5811 (comment). You might notice at the bottom of this PR it says All checks have passed. This includes testing on Windows, but this test is skipped if you look at the log.

Then it's all good from my side 👍

src/PIL/ImageMode.py
```diff
@@ -52,6 +52,11 @@ def getmode(mode):
             "HSV": ("RGB", "L", ("H", "S", "V")),
             # extra experimental modes
             "RGBa": ("RGB", "L", ("R", "G", "B", "a")),
+            "BGR": ("BGR", "L", ("B", "G", "R")),
```
radarhere, line 55: I don't see the need for this? BGR is not listed in the docs or in [Storage.c](https://github.com/python-pillow/Pillow/blob/0648692b2c18cd42f5d20aef9ecf3d65c9f98941/src/libImaging/Storage.c#L131)

FirefoxMetzger: @radarhere Sure, we can remove it. In this case, what should be the base mode for the respective `BGR;X`? That mode itself?

radarhere: You've already listed the base mode for `BGR;*` as just `RGB`.

FirefoxMetzger: @radarhere Oh, right. That was unintentional on my part. Initially I thought the basemode for `BGR;*` should be `BGR` and added mode `BGR` correspondingly.

However, since `RGB` is the desired mode, I will keep it and drop mode `BGR`.

PR #5950, opened 2022-01-10
Moved xdg-open to lowest priority

### LINKED ISSUE #5945: Image.open on Ubuntu 20.04 gives the error "No images found" in the viewer
The viewer is opened, but a "No images found" error is displayed.  Pillow version `pillow-9.0.0`.  This may be the same bug: https://stackoverflow.com/questions/63579415/why-does-pil-fail-to-display-images-with-error-no-images-found-in-file-tmp

Code to reproduce:

```python
from PIL import Image

im = Image.open(r"/var/lib/app-info/icons/ubuntu-focal-updates-universe/48x48/inkscape_inkscape.png")
im.show()
```

Changing the code in the `UnixViewer` class in `ImageShow.py` seems to fix it.  I'm not sure why the current code is writing the filename to a file and then reading it back from stdin, using `shell=True` in `Popen`.   Deleting the `rm -f $im"` command in the `Popen` call seems to fix the problem.  It also seems to work using `cat | xargs -L1` in the Popen command instead of setting the shell variable `im`.

This simpler code replacement also seems to work, without setting `shell=True`:

```python
def show_file(self, file, **options):
    """Display given file"""
    command = self.get_command_ex(file, **options)[0]
    subprocess.Popen([command, file])
    return 1
```

radarhere:
Resolves #5945

In the issue, two users find that ImageShow no longer successfully passes images to `eog`. It would seem that #5897 calling `xdg-open` has introduced some delay that means that images are removed before the viewing application is finished with them.

We could add in a `sleep` to try and solve this, but I would think it is preferable to minimise the race condition.

This PR instead moves `xdg-open` to the lowest priority, so that `eog` can run directly.

### DIFF
--- a/src/PIL/ImageShow.py
+++ b/src/PIL/ImageShow.py
@@ -243,8 +243,6 @@ class XVViewer(UnixViewer):
 
 
 if sys.platform not in ("win32", "darwin"):  # unixoids
-    if shutil.which("xdg-open"):
-        register(XDGViewer)
     if shutil.which("display"):
         register(DisplayViewer)
     if shutil.which("gm"):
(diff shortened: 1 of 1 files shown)


radarhere:
I've created PR #6010 instead of this.

PR #5953, opened 2022-01-11
Removed unused constant

radarhere:
`ImageMath.VERBOSE` has been unused since PIL, so this PR removes it.

While I'm here, I've also included two cleanup changes.
- https://github.com/python-pillow/Pillow/blob/9a9afd36b41c3ff511ec2788ad657c0757e24b3e/src/PIL/ImageMath.py#L68-L73 The `ValueError` cannot be reached, so I've removed it.
- https://github.com/python-pillow/Pillow/blob/9a9afd36b41c3ff511ec2788ad657c0757e24b3e/src/PIL/ImageMath.py#L76-L83 `im1.size` will equal `size` at the end of this code, meaning that the last three lines can be simplified to a single line.

### DIFF
--- a/src/PIL/ImageMath.py
+++ b/src/PIL/ImageMath.py
@@ -19,8 +19,6 @@ import builtins
 
 from . import Image, _imagingmath
 
-VERBOSE = 0
-
 
 def _isconstant(v):
     return isinstance(v, (int, float))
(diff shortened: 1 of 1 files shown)


hugovk:
Docs failure is unrelated, from the new sphinx-issues 3.0.0 release: https://github.com/sloria/sphinx-issues/issues/125.

Let's see if it gets a quick fix, otherwise we can pin it around it.

PR #5954, opened 2022-01-11
Added enums

### LINKED ISSUE #5875: Change from constants to enums
### Feature request

Example set of constants that would fit into an enum really well:

https://pillow.readthedocs.io/en/stable/_modules/PIL/Image.html#Image.transpose

That could be refactored to:
```python
class TransposeType(enum.IntEnum):
    FLIP_LEFT_RIGHT = 0
    FLIP_TOP_BOTTOM = 1
    [etc]

FLIP_LEFT_RIGHT = TransposeType.FLIP_LEFT_RIGHT
FLIP_TOP_BOTTOM = TransposeType.FLIP_TOP_BOTTOM
[etc]
```

Advantages:
- easier to check if a value is as expected (check type of enum value)

Disadvantages:
- have to research if any existing usages can break

Smaller factors:
- (some) refactoring required
- (negligible) performance overhead

radarhere:
Resolves #5875, by replacing various sets of constants with `IntEnum` classes. This would seem to be the exact scenario for which enums were added to Python.

I have retained backwards compatibility by including lines like `globals().update(enum.__members__)`.

I have updated internal usage to use the enums, but not the tests or documentation.


hugovk:
This is a good way to group things together, but for backwards compatibility we have each one doubled up:

```python
Image.ROTATE_90
Image.Transpose.ROTATE_90
Image.ROTATE_180
Image.Transpose.ROTATE_180
Image.ROTATE_270 
Image.Transpose.ROTATE_270
# etc.
```

Although `dir(Image)` only lists the extra `Transpose`.


Is the idea be that end user code should be updated to use the enum form?

Do we keep the old constants "forever"?

Or do we want to deprecate the old constants and remove them in Pillow 10 (15 months' deprecation period)?

melonmouse (commented):
To me, this change looks great! I'm a guest in this codebase, so can't judge it completely - thats the only reason I'm commenting and not approving.

One small remark would be that some of the introduced enum classes have names that are a bit generic (like Encoding) and might lead to naming collisions later.

radarhere:
When you talk about naming collisions, I find it hard to imagine that we'd want to add another module-level variable called `Format`. Or are you thinking of a user running `from PIL.BlpImagePlugin import Format`?

I was trying to avoid users writing
```pycon
>>> from PIL import BlpImagePlugin
>>> BlpImagePlugin.BlpFormat.JPEG
```
Instead,
```pycon
>>> from PIL import BlpImagePlugin
>>> BlpImagePlugin.Format.JPEG
```
seems cleaner to me.

radarhere:
Ok, I've updated the tests, documentation, and deprecated the original constants.

I've also deprecated the ["backwards compatibility" constants for Resampling](https://pillow.readthedocs.io/en/stable/reference/Image.html#resampling-filters), and deprecated `NEAREST` for use in Dither, [which isn't documented](https://pillow.readthedocs.io/en/stable/reference/Image.html#dither-modes).

hugovk:
Let's also add this to the "Deprecations and removals" page and release notes.

https://pillow.readthedocs.io/en/latest/deprecations.html

radarhere:
Ok, done. See https://pillow--5954.org.readthedocs.build/en/5954/deprecations.html#constants and https://pillow--5954.org.readthedocs.build/en/5954/releasenotes/9.1.0.html#deprecations

PR #5958, opened 2022-01-13
Corrected memory allocation

radarhere:
Addresses https://github.com/python-pillow/Pillow/commit/1e092419b6806495c683043ab3feb6ce264f3b9c#commitcomment-63846366

I failed to acknowledge that `malloc` allocates a block of "size", while `calloc` allocates "nitems" blocks of "size".


stratakis:
Pinging here. Does this address the issue from 1e09241 ?

radarhere:
That too much memory is allocated? Yes, it should fix that.

PR #5959, opened 2022-01-15
Deprecated show_file "file" argument in favour of "path"

radarhere:
First pointed out in https://github.com/python-pillow/Pillow/pull/3450#discussion_r230547142, 
https://github.com/python-pillow/Pillow/blob/fd2b07c454b20e1e9af0cea64923b21250f8f8d6/src/PIL/ImageShow.py#L108 the `file` argument mirrors the builtin `file`.

This PR deprecates it in favour of `path`.

### DIFF
--- a/Tests/test_imageshow.py
+++ b/Tests/test_imageshow.py
@@ -79,3 +79,18 @@ def test_ipythonviewer():
 
     im = hopper()
     assert test_viewer.show(im) == 1
+
+
+@pytest.mark.skipif(
+    not on_ci() or is_win32(),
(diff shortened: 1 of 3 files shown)


hugovk:
Let's also add it to https://pillow.readthedocs.io/en/stable/deprecations.html#deprecated-features

And can we add a note to https://pillow.readthedocs.io/en/stable/reference/ImageShow.html#PIL.ImageShow.Viewer.show_file as well?

radarhere:
Ok, done.

hugovk:
Thanks!

PR #5963, opened 2022-01-16
Forbid Lambda expressions in ImageMath.eval()

feliperalmeida:
## Description
This is an additional fix for CVE-2022-22817 (addressed in https://github.com/python-pillow/Pillow/pull/5923). As [commented](https://github.com/python-pillow/Pillow/pull/5923#issuecomment-1008715242) in the original PR, it's still possible to execute arbitrary code via lambda expressions - e.g.: `ImageMath.eval("(lambda: exit())()")`.

### Original Fix
The original fix checks all `co_names` in the compiled `expression` passed to `ImageMath.eval()` against builtins names. However, lambda expressions generates anonymous functions and they're not listed in the `co_names` structure. 

## Changes in this PR
This PR forbids Lambda expressions being passed to `ImageMath.eval()` by verifying all functions names in the literals section of the bytecode (`co_consts`). 

## Trade-offs 
Lambda expressions will no longer be supported in `ImageMath.eval()`. If such support is desired, all code objects inside `co_consts` can be verified in the same way as the original fix does (iterating through `co_names` and verifying against builtins names). Please let me know in the comments I can rewrite this fix.

### DIFF
--- a/Tests/test_imagemath.py
+++ b/Tests/test_imagemath.py
@@ -52,9 +52,10 @@ def test_ops():
     assert pixel(ImageMath.eval("float(B)**33", images)) == "F 8589934592.0"
 
 
-def test_prevent_exec():
+@pytest.mark.parametrize("expression", ("exec('pass')", "(lambda: None)()"))
+def test_prevent_exec(expression):
     with pytest.raises(ValueError):
-        ImageMath.eval("exec('pass')")
+        ImageMath.eval(expression)
 
 
 def test_logical():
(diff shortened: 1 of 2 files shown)


radarhere:
Hi. I've just merged #6009 instead of this.
Please don't take that to mean that we completely ignored your PR - it's just that this is a CVE, so it seemed better to not draw too much attention to it by discussing it openly. We considered this privately, and went with #6009, so that users can still use lambdas if they want. That fix will be shortly be part of Pillow 9.0.1.
If you have any further thoughts, let us know.

hugovk:
@feliperalmeida And thanks for the PR!

In the future, when it comes to security-related issues, please could you check and follow the security policy of the project? If there's none available, it's good practice to ask how to disclose.

The Pillow one is here:

https://github.com/python-pillow/Pillow/security/policy

Thanks again!

PR #5972, opened 2022-01-18
Raise an error when performing a negative crop

### LINKED ISSUE #5957: Provide an error message when user gives incorrect arguments to Image.crop()
### What did you do?

I called `Image.crop()`, but I got confused about what the arguments are meant to be.

It's meant to be a `(left, upper, right, lower)`-tuple, but I supplied a `(left, upper, width, height)`-tuple instead. The error message I got was:

```
```

which took me a while to figure out, because all the crops looked like they were inside the image – I think possibly this is a side effect of `width` being lower than `left`, and `height` being lower than `upper`?

I'd get an error message explaining my mistake.
Maybe the `crop()` function could throw an error like:

```
ValueError: crop bounds should be a (left, upper, right, lower)-tuple, but right=100 was less than left=50
```

because I don't think cropping ever makes sense if `right > left` and `lower > upper` aren't satisfied.

### What actually happened?

I got confused.

* OS: macOS 10.15.7
* Python: Python 3.9.9
* Pillow: Pillow==9.0.0

### Steps to reproduce

This is a 100×100 pixel image:

gradient

```python
from PIL import Image

im = Image.open("gradient.png")
print(im.size)

# What I think I'm doing:
# left=0, upper=0, width=25, height=25
#
# This works, because box[2] > box[0] and box[3] > box[1]
crop1 = im.crop(box=(0, 0, 25, 25))
crop1.save("crop1.png")

# What I think I'm doing:
# left=50, upper=50, width=25, height=25
#
# What I should actually have done:
# left=50, upper=50, right=75, upper=75
crop2 = im.crop(box=(50, 50, 25, 25))
crop2.save("crop2.png")
```

In case it's helpful, here's the full stack trace:

<details><summary>Stack trace</summary>

```
  File "/Users/alexwlchan/Library/Python/3.9/lib/python/site-packages/PIL/ImageFile.py", line 495, in _save
    fh = fp.fileno()
AttributeError: '_idat' object has no attribute 'fileno'

  File "./example.py", line 19, in <module>
    crop2.save("crop2.png")
  File "/Users/alexwlchan/Library/Python/3.9/lib/python/site-packages/PIL/Image.py", line 2212, in save
  File "/Users/alexwlchan/Library/Python/3.9/lib/python/site-packages/PIL/PngImagePlugin.py", line 1348, in _save
    ImageFile._save(im, _idat(fp, chunk), [("zip", (0, 0) + im.size, 0, rawmode)])
  File "/Users/alexwlchan/Library/Python/3.9/lib/python/site-packages/PIL/ImageFile.py", line 503, in _save
    e.setimage(im.im, b)
```

</details>

radarhere:
Resolves #5957

At the moment, mixing up the order of the `box` values for `crop()`, so that a negative crop is performed, doesn't raise an error.
```python
from PIL import Image
im = Image.new("RGB", (10, 10))
im.crop((10, 10, 0, 0))
```
gives a (0, 0) image.

In the issue, this was done unintentionally by a user, and because no error was raised, they didn't realise their mistake.

This PR raises an error is the right of a crop region is less than the left, or if the bottom of the region is less than the top.

I am presuming that it is uncommon for users to want a negative crop, and more common for users to mix up the order of the values.

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -1145,6 +1145,11 @@ class Image:
         if box is None:
             return self.copy()
 
+        if box[2] < box[0]:
+            raise ValueError("Region right less than region left")
+        elif box[3] < box[1]:
+            raise ValueError("Region lower less than region upper")
+
         self.load()
         return self._new(self._crop(self.im, box))
 
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -1146,9 +1146,9 @@ class Image:
             return self.copy()
 
         if box[2] < box[0]:
-            raise ValueError("Region right less than region left")
+            raise ValueError("Coordinate 'right' is less than 'left'")
         elif box[3] < box[1]:
-            raise ValueError("Region lower less than region upper")
+            raise ValueError("Coordinate 'lower' is less than 'upper'")
 
         self.load()
         return self._new(self._crop(self.im, box))
--- a/Tests/test_decompression_bomb.py
+++ b/Tests/test_decompression_bomb.py
@@ -86,21 +86,12 @@ class TestDecompressionCrop:
             pytest.warns(Image.DecompressionBombWarning, src.crop, box)
 
     def test_crop_decompression_checks(self):
-
         im = Image.new("RGB", (100, 100))
 
-        good_values = ((-9999, -9999, -9990, -9990), (-999, -999, -990, -990))
-
-        warning_values = ((-160, -160, 99, 99), (160, 160, -99, -99))
-
-        error_values = ((-99909, -99990, 99999, 99999), (99909, 99990, -99999, -99999))
-
-        for value in good_values:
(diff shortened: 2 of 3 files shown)


alexwlchan (approved):
Looks good to me, thanks!

src/PIL/Image.py
```diff
@@ -1145,6 +1145,11 @@ def crop(self, box=None):
         if box is None:
             return self.copy()
 
+        if box[2] < box[0]:
+            raise ValueError("Region right less than region left")
+        elif box[3] < box[1]:
+            raise ValueError("Region lower less than region upper")
```
hugovk, line 1151: Looks good, maybe this is clearer?

```suggestion
        if box[2] < box[0]:
            raise ValueError("Coordinate 'right' is less than 'left'")
        elif box[3] < box[1]:
            raise ValueError("Coordinate 'lower' is less than 'upper'")
```

hugovk:
Thanks!

PR #5973, opened 2022-01-18
Always save TIFF with contiguous planar configuration

### LINKED ISSUE #3622: PlanarConfiguration save
### What did you do?
- Open a file using Pillow Image.open `im = Image.open(..blah)`
- downscale the image using Image.thumnail `im.thumbnail(..blah)`
   - (Args: (512, 512), Image.ANTIALIAS) 
- Save the image using save function, only passing new file name `im.save('file_two.tif/tiff')` - Same issue for tiff and tif.
Image would be saved, scaled down, and viewable.

### What actually happened?
Image was downscaled, but corrupt, appears red and pixellated, cannot be opened in some image viewers.

Note: I cannot repeat this on Mac OS, haven't tried this on Debian... yet.

* OS: Windows 10 Version 10.0.17763 Build 17763
* Python:  Python 3.6.8 
* Pillow: Pillow 5.4.1

Corrupted Image
```python
from PIL import Image
import os


mode = input("Enter mode")

files = [f for f in os.listdir('./{}'.format(mode)) if os.path.isfile(os.path.join('./{}'.format(mode), f))]

im = Image.open('./{}/{}'.format(mode, files[0]))
im.show()

input()
im.thumbnail((512, 512), Image.ANTIALIAS)

im.save('./PILissue.tiff')
```

radarhere:
Resolves #3622

If a TIFF image is opened with "separate" [planar configuration](https://www.awaresystems.be/imaging/tiff/tifftags/planarconfiguration.html), it is successfully read.

However, if that same image is saved, Pillow tries to keep the "separate" planar configuration from the input file, and the output image is broken.

This PR simply changes Pillow to always save "contiguous", and the output image is fixed.

### DIFF
--- a/Tests/test_file_tiff.py
+++ b/Tests/test_file_tiff.py
@@ -594,6 +594,17 @@ class TestFileTiff:
         with Image.open(infile) as im:
             assert_image_equal_tofile(im, "Tests/images/tiff_adobe_deflate.png")
 
+    def test_planar_configuration_save(self, tmp_path):
+        infile = "Tests/images/tiff_tiled_planar_raw.tif"
+        with Image.open(infile) as im:
+            assert im._planar_configuration == 2
+
+            outfile = str(tmp_path / "temp.tif")
+            im.save(outfile)
+
+            with Image.open(outfile) as reloaded:
+                assert_image_equal_tofile(reloaded, infile)
+
     def test_palette(self, tmp_path):
         def roundtrip(mode):
             outfile = str(tmp_path / "temp.tif")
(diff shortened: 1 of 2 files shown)


kmilos:
The forced libtiff writer cannot preserve it either?

radarhere:
I don't know why the saved image is broken in main. However,

https://www.awaresystems.be/imaging/tiff/tifftags/planarconfiguration.html
> The specification adds a warning that PlanarConfiguration=2 is not in widespread use and that Baseline TIFF readers are not required to support it.

So it seems like a good idea to avoid the "separate" setting for the sake of compatibility of our created images.

kmilos:
Yeah, it's ok to play it safe for this case. I was just wondering if there was a remote chance of preserving the original image configuration, but unfortunately don't have the time to experiment right now...

Issue #5966, opened 2022-01-17
[9.0.0 Regression] Pillow claims GIF frame to be mode RGB instead of P


doko42:
Forwarding from https://bugs.debian.org/1003661

img2pdf's autopkg tests fail with pillow 9.0.0, as seen at
https://tracker.debian.org/pkg/img2pdf

=========================== short test summary info ============================
FAILED src/img2pdf_test.py::test_general[animation.gif-internal] - AssertionE...
FAILED src/img2pdf_test.py::test_general[animation.gif-pikepdf] - AssertionEr...
ERROR src/img2pdf_test.py::test_gif_animation[internal] - TypeError: object i...
ERROR src/img2pdf_test.py::test_gif_animation[pikepdf] - TypeError: object is...

The img2pdf autopkgtest has found a regression in Pillow. Minimal reproducer
with the attached gif:

python3 -c 'from PIL import Image;img=Image.open("animation.gif");img.seek(1);print(img.mode)'

With 9.0.0 this will output "RGB" even though frame 1 is a palette based image.
With Pillow 8.4.0 above one-liner will correctly print "P".

tracked this down to this commit: b383a175be70da0df33babea3d480efd9768e937

doko42:
[https://bugs.debian.org/cgi-bin/bugreport.cgi?att=1;bug=1003661;filename=animation.gif;msg=8](url)

radarhere:
Hi. This is not a bug, but a deliberate and necessary change.

https://pillow.readthedocs.io/en/stable/releasenotes/9.0.0.html#convert-subsequent-gif-frames-to-rgb-or-rgba
> Since each frame of a GIF can have up to 256 colors, after the first frame it is possible for there to be too many colors to fit in a P mode image. To allow for this, seeking to any subsequent GIF frame will now convert the image to RGB or RGBA, depending on whether or not the first frame had transparency.

josch:
Hi, I'm the author of img2pdf and version 9.0.0 breaks the tests due to the new handling of GIF images.

If this change is here to stay, what can software do that wants to really figure out whether frame X of an image is P or RGB?

radarhere:
The first frame will be P (and, technically, it could also be L, but that's not new behaviour).
Any frame that you seek to after the first frame will be RGB or RGBA.

If you would like to know what the mode of an image is, you can just get the `mode` attribute. That will update when you seek to the new frame.
```pycon
>>> from PIL import Image
[...]

PR #5974, opened 2022-01-18
WIP: src/PIL/GifImagePlugin.py: only convert image if palettes actually changes

josch:
This is a WIP pull request which avoids converting image frame of GIF images in case all earlier palettes of the GIF are actually the same.

I wrote this because the suggestions by @radarhere in https://github.com/python-pillow/Pillow/issues/5966#issuecomment-1015155331 doesn't work for some images which already get converted in the `_seek` function. Thus, this pull request introduces a new variable that keeps track of the last used palette and only converts images if any of the former frames changed the palette (including the global palette).

What do you think?


josch:
Some dispose-related tests are still failing and I'm working on that (suggestions welcome!).

The advantages of only changing to the new behaviour if needed (i.e. if the palette does change from one frame to the next) are:

 - faster because we have to run `convert` less often
 - on all my test input the palette images were smaller than the images after they were converted to RGB(A)
 - software like img2pdf does not need to add special-casing to be compatible with both Pillow 9.0.0 and Pillow before 9.0.0. Other pieces of software might be in a similar situation.

radarhere:
> software like img2pdf does not need to add special-casing to be compatible with both Pillow 9.0.0 and Pillow before 9.0.0

Could you provide some more detail on this point?

josch:
> > software like img2pdf does not need to add special-casing to be compatible with both Pillow 9.0.0 and Pillow before 9.0.0
> 
> Could you provide some more detail on this point? If I'm understanding correctly, the img2pdf tests are failing just because they expect an indexed colorspace and are instead finding RGB. That doesn't sound like a problem for end users, just a problem for the test suite.

Yes, this is a minor argument and only becomes the larger argument if there is more software or users having similar expectations. The bigger argument is, that the files produced by img2pdf are now bigger because they are not palette images anymore even though they could be palette images because all frames share the same palette.

The larger file size and the additional processing cost (due to converting) are my main arguments.

> [#5966 (comment)](https://github.com/python-pillow/Pillow/issues/5966#issuecomment-1015151962)
> 
> > what can software do that wants to really figure out whether frame X of an image is P or RGB?
> 
> At the moment in the main branch, if you seek past the first frame of a GIF, it will be either RGB or RGBA.
> 
> My primary concern with this PR is that it would no longer be predictable what mode the image would be in that scenario. It might have the same palette and be P, it might not and be RGB. Surely that sounds like it would make situations like the one happening in your test suite worse, not better.

But it would be predictable. If the input GIF has the same palette for all frames, then al frames would be P. I'd just add different test cases for the different situations.

Really, the img2pdf testsuite is of little concern here. If subsequent images have the same palette as all the frames before, then that's an advantage for filesize (because we can retain the palette when converting) and for processing speed (because frames don't have to be converted from P to RGB).

radarhere:
> the files produced by img2pdf are now bigger

Just thinking aloud here for a second. Presuming you were talking about [img2pdf's animation.gif](https://raw.githubusercontent.com/josch/img2pdf/main/src/tests/input/animation.gif), I ran
```python
from PIL import Image
with Image.open("animation.gif") as im:
    im.save("out.gif", save_all=True)
```
and found that the saved GIF is actually smaller with Pillow 9.0 than it is with Pillow 8.4.0. That seemed to go against your point, but then I realised - you're not talking about Pillow resaving a GIF, you're talking about img2pdf saving a GIF.

wiredfool:
What is img2pdf actually doing with the multiple frames of gif images?

josch:
> What is img2pdf actually doing with the multiple frames of gif images?

For images that contain more than one frame (so not only for GIF images) img2pdf outputs one page per frame unless the `--first-frame-only` option is given.

Thus, I very much welcome the fixes to Pillow concerning the automatic conversion to RGB if 256 colors are not enough to represent a frame. I think that's awesome! I just want to retain the palette in those cases were it does *not* change because storing an RGB image (even after paeth filter and zlib compression) in PDF requires more space than the palette based image.

Sorry, that I haven't pushed more fixes to this branch yet. I'll find more time to work on this during the weekend.

josch:
Hi,
I worked on it a bit more but I'm stuck with four tests failing:

```
FAILED Tests/test_file_gif.py::test_transparent_dispose - assert 2 == 1
FAILED Tests/test_file_gif.py::test_dispose_previous_first_frame - AssertionError: got different content
FAILED Tests/test_file_gif.py::test_dispose2_diff - assert (0, 0, 0, 255) == (255, 255, 255, 0)
FAILED Tests/test_file_gif.py::test_missing_background - AssertionError: got different content
```

Maybe somebody has some time to look into this and find out where I made a mistake?

Thanks!

mara004:
Hi,
I was wondering whether there is any progress on this topic? (I would like to see img2pdf's animated gif tests fixed with Pillow >= 9)
Thanks!

radarhere:
I've created PR #6150 as an alternative to this. In it, I've added a setting to control this behaviour. Running this following code with that PR should fix the img2pdf tests.
```python
from PIL import GifImagePlugin
GifImagePlugin.PALETTE_TO_RGB = GifImagePlugin.ModeStrategy.DIFFERENT_PALETTE_ONLY 
```
However, I'm having trouble confirming that, due to https://gitlab.mister-muffin.de/josch/img2pdf/issues/132. If you would like to confirm that my change resolves this?

mara004:
Thanks, that sounds great. I installed Pillow from your pull request:
```python3
gh pr checkout 6150
pip3 install . -v
```
and then set the configuration:
```python3
GifImagePlugin.PALETTE_TO_RGB = GifImagePlugin.ModeStrategy.DIFFERENT_PALETTE_ONLY 
```
I can confirm that all GIF tests in question pass now (in my local, patched img2pdf variant).

josch:
Thanks for testing @mara004 -- I tried testing it too but now other tests fail. Minimal reproducer by using this image:

https://gitlab.mister-muffin.de/josch/img2pdf/src/branch/main/src/tests/input/animation.gif

And then running:

```
src/img2pdf.py -v src/tests/input/animation.gif -o out.pdf
[...]
ERROR:__main__:error: local variable 'frame_dispose_extent' referenced before assignment
  File "/home/josch/annex/personal/projects/img2pdf/src/img2pdf.py", line 3980, in main
    convert(
  File "/home/josch/annex/personal/projects/img2pdf/src/img2pdf.py", line 2280, in convert
    ) in read_images(
  File "/home/josch/annex/personal/projects/img2pdf/src/img2pdf.py", line 1654, in read_images
    imgdata.seek(img_page_count)
  File "/usr/lib/python3/dist-packages/PIL/GifImagePlugin.py", line 147, in seek
    self._seek(f)
  File "/usr/lib/python3/dist-packages/PIL/GifImagePlugin.py", line 319, in _seek
    self.dispose_extent = frame_dispose_extent
UnboundLocalError: local variable 'frame_dispose_extent' referenced before assignment
```

Can you confirm @mara004?

@radarhere thanks a lot for #6150 -- I'll continue to test your work. To work around the pikepdf issue in img2pdf you can either downgrade to 4.2.0 as @mara004 already suggested or apply the following patch:

```diff
--- a/src/img2pdf.py
+++ b/src/img2pdf.py
@@ -1108,7 +1110,10 @@ class pdfdoc(object):
 
         # the /OpenAction array must contain the page as an indirect object
         if self.engine == Engine.pikepdf:
-            initial_page = self.writer.make_indirect(initial_page)
+            if isinstance(initial_page, pikepdf.Page):
+                initial_page = self.writer.make_indirect(initial_page.obj)
+            else:
+                initial_page = self.writer.make_indirect(initial_page)
 
         if self.magnification == Magnification.fit:
             catalog[PdfName.OpenAction] = PdfArray([initial_page, PdfName.Fit])
```

mara004:
No, I can't reproduce your issue. For me converting `animation.gif` works just fine, and all tests pass.
Your traceback suggests that the issue is happening down in Pillow when seeking to the next frame? Maybe we installed a different state of the pull request?

josch:
@radarhere I can now confirm the results of @mara004 and #6150 lets the img2pdf testsuite pass just fine after setting `GifImagePlugin.PALETTE_TO_RGB = GifImagePlugin.ModeStrategy.DIFFERENT_PALETTE_ONLY`. Thus, I'm closing this pull request in favor of #6150. Thanks!

PR #5980, opened 2022-01-21
Connected discontiguous polygon corners

### LINKED ISSUE #4347: ImageDraw.Draw.polygon output is sometimes discontiguous
### What did you do?
I drew a filled quadrilateral.

The output would be a contiguous quadrilateral.

### What actually happened?
The output was discontiguous.

* OS: macOS 10.14.6
* Python: Python 3.7.4 (default, Jul 11 2019, 01:08:00) [Clang 10.0.1 (clang-1001.0.46.4)] on darwin
* Pillow: 6.1.0

```python3
img = Image.new("1", (128, 128))
coords = [(29.0, 74.0), (62.0, 57.0), (99.0, 54.0), (66.0, 71.0)]
drawer = ImageDraw.Draw(img)
drawer.polygon(coords,1,1)
img.show()
```

Screen Shot 2020-01-08 at 4 58 26 PM

radarhere:
Resolves #4347

The issue reports that for a particular polygon, the corners are not visually joined to the main shape.

image

Pillow draws this polygon is by plotting on each row where each of the edges intersect, and then going through row by row and drawing horizontal lines. If I adjust Pillow to just draw the intersections of the edges on each row, it is clearer what has happened - no point in any of the lines is wrong, the angles are just such that the corner doesn't connect to the rest of the shape visually.

points

When hitting such corners, this PR finds the two edges making up the corner, calculates where those edges intersected with the adjacent row, and then stretches out the corner to connect to it.

The test includes the polygon from the issue, and two variations I put together.
Screen Shot 2022-01-21 at 11 49 24 am

Here is the test image without this fix.

Screen Shot 2022-01-21 at 11 49 45 am


radarhere:
I've created https://github.com/google/oss-fuzz/pull/7175 in an attempt to fix the CIFuzz error.

radarhere:
Ok, the CIFuzz error was instead fixed by https://github.com/google/oss-fuzz/pull/7199

PR #5989, opened 2022-01-26
Upgraded raqm to 0.8.0

radarhere:
The build commands also had to be changed, as libraqm changed to meson - https://github.com/HOST-Oman/libraqm/pull/97


khaledhosny:
FWIW, now can now use the Apache 2.0-lincensed SheenBidi instead of LGPL-licensed FriBiDi (which I recall the being an issue for Pillow), so you might want to consider switching libraries.

PR #5999, opened 2022-01-30
Switched from windows-2019 to windows-latest in GitHub Actions

radarhere:
In test.yml, we have

https://github.com/python-pillow/Pillow/blob/5fdf29f13b30b763b5b9f399ec227d76fb51f291/.github/workflows/test.yml#L12-L13

but in test-windows.yml and test-mingw.yml, we have

https://github.com/python-pillow/Pillow/blob/5fdf29f13b30b763b5b9f399ec227d76fb51f291/.github/workflows/test-windows.yml#L7
https://github.com/python-pillow/Pillow/blob/5fdf29f13b30b763b5b9f399ec227d76fb51f291/.github/workflows/test-mingw.yml#L7

This PR switches from windows-2019 to windows-latest, to ensure that we are always running the latest version, and don't have to upgrade it.

See https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners for more information about this topic.

### DIFF
--- a/.github/workflows/test-mingw.yml
+++ b/.github/workflows/test-mingw.yml
@@ -4,7 +4,7 @@ on: [push, pull_request, workflow_dispatch]
 
 jobs:
   build:
-    runs-on: windows-2019
+    runs-on: windows-latest
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


hugovk:
It has been the case that when a new one came out they don't flip it to `latest` for some time.

But that should be fine, unless we specifically need to test a given version.

Thanks!

PR #6002, opened 2022-01-31
Fix lcms2 URL

hugovk:
As pointed out here: https://github.com/python-pillow/Pillow/commit/5fdf29f13b30b763b5b9f399ec227d76fb51f291#r65695669


radarhere:
Thanks

PR #6005, opened 2022-01-31
Unix viewers

### LINKED ISSUE #5976: XDGViewer and GmDisplayViewer fail to show images
### What did you do?
1. I created script to check Unix image viewers available in PIL.ImageShow module. The script example for XDGViewer is provided
below.

2. I run the script for all Linux/Unix viewers except XVViewer (I didn't manage to obtain it)

I expected to open my image with each single viewer I tried to use:
XDGViewer, DisplayViewer, GmDisplayViewer, EogViewer

### What actually happened?

The image failed to open with

a) XDGViewer
b) GmDisplayViewer

* OS:  Ubuntu, 20.04.3 LTS (Focal Fossa)
* Python: 3.9.9
* Pillow: 9.1.0.dev0

```python
from PIL import Image
from PIL.ImageShow import XDGViewer

if __name__ == "__main__":
    viewer = XDGViewer()
    img_file = "path/to/image_file"
    with Image.open(img_file) as im:
        viewer.show(im)
```
cosmonaut

alv2017:
Fixes #5976

 *  The major changes affected **UnixViewer** class. I used Python threads in order to control image opening by external Linux/Unix viewers. I used a viewer thread and a viewer monitoring thread that monitors the image viewing process and it is also responsible for removal of the temporary images.
 * I also changed how the way **get_command_ex(self, file, **options):** method works: the command returned from the method is the exact command opening the image file provided as an argument. This is in order to keep the CL stuff as simple as possible.
 * I added additional method **get_executable(self, file, **options):** that returns an executable of the specific Unix Viewer.
 * Set Viewer class property **format** to "PNG", and remove this property from the Viewer child classes.
 * Set Viewer class property **options** to {"compress_level": 1}, and remove this property from the Viewer child classes.
 * Move XDGViewer to the bottom of the viewers list. This viewer will be utilized only in case if all the previous viewers failed.
 * Remove, skip or timeout tests that are attempting to open actual image.


src/PIL/ImageShow.py
```diff
@@ -282,12 +302,11 @@ class XVViewer(UnixViewer):
     This viewer supports the ``title`` parameter.
     """
 
-    def get_command_ex(self, file, title=None, **options):
+    def get_command_ex(self, file, **options):
         # note: xv is pretty outdated.  most modern systems have
         # imagemagick's display command instead.
-        command = executable = "xv"
-        if title:
-            command += f" -name {quote(title)}"
```
radarhere, line 290: Could you explain why you think `xv` doesn't need a title argument?

src/PIL/ImageShow.py
```diff
@@ -241,20 +262,19 @@ class XDGViewer(UnixViewer):
...
 class DisplayViewer(UnixViewer):
     """
     The ImageMagick ``display`` command.
-    This viewer supports the ``title`` parameter.
     """
 
-    def get_command_ex(self, file, title=None, **options):
-        command = executable = "display"
-        if title:
-            command += f" -name {quote(title)}"
```
radarhere, line 257: Could you explain why you think display doesn't need a title argument?

radarhere:
> * The major changes affected **UnixViewer** class. I used Python threads in order to control image opening by external Linux/Unix viewers. I used a viewer thread and a viewer monitoring thread that monitors the image viewing process and it is also responsible for removal of the temporary images.

For clarity - this is motivated by [wanting to show multiple images at a time](https://github.com/python-pillow/Pillow/issues/5976#issuecomment-1018444936), and to navigate between them after 20 seconds.

Minor note - you understand that our CI jobs are now running indefinitely? :p

> * I also changed how the way **get_command_ex(self, file, **options):** method works: the command returned from the method is the exact command opening the image file provided as an argument. This is in order to keep the CL stuff as simple as possible.

Pillow values backwards compatibility, and this would be a backwards incompatible change. Could you provide some more detail about why this is needed?

src/PIL/ImageShow.py
alv2017: The **-name** option: sets string displayed in image window's title bar. I thought that this option is not particularly useful and removed it. Now I realize that this had an effect on documentation, including the ImageShow.show() method.

src/PIL/ImageShow.py
alv2017: Same thing as for xv, on top of it I think that the option is used incorrectly here. Please see the doc for X Windows: https://linux.die.net/man/7/x. I think that the **-title** option would be more relevant here.

alv2017:
**Minor note**: This is motivated by the idea that the Python program doesn't terminate till the user closes the opened image in the viewer. In short the termination of the thread requires a manual user action. Unfortunately, what works for the user, doesn't work for CI. :) I would be curious to see failing CI jobs, because all the unit tests that I run on my machine execute successfully.

I also think that the user has to have an explicit control of the processes created by the program.

**get_command_ex()** method: I do not agree that there is a backward compatibility issue here: the interface is preserved:
we take object, file and options as input and we return command and executable as output. I think that this is a cleaner and more consistent approach.

alv2017:
Hi, Andrew, thanks for the feedback, and please let me know

1) Do you agree/disagree with the taken approach: viewers have to be managed by the Python program that started them.
2) What would be your solution?

radarhere:
Yes, #6010 conflicted with this.

You asked for my thoughts, so...
- I still think you're breaking backwards compatibility on `get_command_ex`, as you're changing the purpose of the first part of the tuple that is returned.
- I don't mind `get_executable`, but I'm also not clear on the need for it.
- My only weird feeling about moving `format` and `options` up to Viewer is that `IPythonViewer` doesn't use them, but that's minor. Fyi - before #4080, the `format` for `WindowsViewer` was BMP.
- I'm not inclined to rearrange `XDGViewer`. The user from #5897 thought it was a good idea to have it first, so that the preferred application is opened. A user from #5976 thinks that the viewer should be lightweight instead. I don't personally favour reverting a change if there's not unity on what to do.
- I haven't checked out `title` instead of `name` for `display`, but I don't think there's a compelling reason here to drop the `title` argument completely.
- If there is a way to keep the temporary images around until the viewer is done with them, without pausing Python execution for the user, yes, that sounds interesting. You may have to adjust this in light of #6010. I haven't experimented much, but do you need to jump to `threading`, or can the same effect be achieved with just `subprocess`?

As something of an aside, in the documentation, it is mentioned that `show()` is intended for debugging.

https://pillow.readthedocs.io/en/stable/handbook/overview.html#image-display
> For debugging, there’s also a show() method which saves an image to disk, and calls an external display utility.

You're talking about opening multiple images at once, keeping them open longer than 20 seconds and swapping between them - sounds to me like you're trying to use this for more than just debugging? Nothing wrong with that, just a thought.

alv2017:
@radarhere thank you for your reply!
The conflict occurred because I rebased incorrectly.
I will take into consideration your concerns and submit a new PR for the viewers.

radarhere:
The next version of this PR is #6021

radarhere:
The final form of this was #6045

PR #6006, opened 2022-02-01
Support Python distributions where _tkinter is compiled in

lukegb:
In some cases, the Tk library may have been directly compiled in or bundled into the main executable by the time Pillow runs, in which case `__file__` isn't available (nor would it make sense to use, anyway).

If `__file__` is missing, then set the library path to None and rely on our Tk loader being able to find the function pointers within the main binary - we know we probably have it because we've managed to import it already.


radarhere:
> In some cases

Could you go into a bit more detail about when this occurs?

radarhere (approved):

PR #6021, opened 2022-02-04
Unix Viewers - 2nd attempt 

### LINKED ISSUE #5945: Image.open on Ubuntu 20.04 gives the error "No images found" in the viewer

alv2017:
Fixes #5945, #5976

 *  Viewer class properties **format** and **options** have been moved to the parent Viewer class, those properties were removed from the Viewer child classed. **format** is set to **"PNG"** and **options** are set to **{"compress_level": 1}** The exception is IPythonViewer class, for this class **format** was set to **None** and options are set to **{}**.
* Python threads were used to control opening/closing of images by external Linux/Unix viewers.
* Within current setup, when working with Linux viewers the Python program remains opened till the user closes the image in the viewer.
* Implementation description: there are viewer thread and monitoring thread. Viewer thread is responsible for opening an image in a viewer; the monitoring thread is responsible for removal of temporary images after user closed the image in the viewer.
* Corresponding documentation is updated.
* Affected tests are updated: Tests/test_imageshow.py, Tests/helper.py

### DIFF
--- a/Tests/helper.py
+++ b/Tests/helper.py
@@ -27,8 +27,7 @@ if os.environ.get("SHOW_ERRORS", None):
     class test_image_results:
         @staticmethod
         def upload(a, b):
-            a.show()
-            b.show()
+            return None
 
 elif "GITHUB_ACTIONS" in os.environ:
     HAS_UPLOADER = True
@@ -314,6 +313,10 @@ def is_ppc64le():
 
 def is_win32():
     return sys.platform.startswith("win32")
+
+
+def is_macos():
+    return sys.platform.startswith("darwin")
 
 
 def is_pypy():
--- a/Tests/test_imageshow.py
+++ b/Tests/test_imageshow.py
@@ -44,7 +44,6 @@ def test_viewer_show(order):
 @pytest.mark.skip(
     reason="""Current implementation of some viewers requires manual closing of an image,
         because of that the tests calling show() method will hang infinitely.
-        Please also note that this test duplicates test_viewer_show() test.
     """
 )
 def test_show():
--- a/Tests/test_imageshow.py
+++ b/Tests/test_imageshow.py
@@ -54,7 +54,7 @@ def test_show():
 def test_viewer():
     viewer = ImageShow.Viewer()
 
-    assert viewer.get_format(None) is None
+    assert viewer.get_format(None) == "PNG"
 
     with pytest.raises(NotImplementedError):
         viewer.get_command(None)
--- a/Tests/test_imageshow.py
+++ b/Tests/test_imageshow.py
@@ -2,7 +2,7 @@ import pytest
 
 from PIL import Image, ImageShow
 
-from .helper import hopper, is_win32, on_ci
+from .helper import hopper, is_macos, is_win32, on_ci
 
 
 def test_sanity():
@@ -41,9 +41,11 @@ def test_viewer_show(order):
     ImageShow._viewers.pop(0)
 
 
-@pytest.mark.skipif(
-    not on_ci() or is_win32(),
-    reason="Only run on CIs; hangs on Windows CIs",
+@pytest.mark.skip(
+    reason="""Current implementation of some viewers requires manual closing of an image,
+        because of that the tests calling show() method will hang infinitely.
+        Please also note that this test duplicates test_viewer_show() test.
+    """
 )
 def test_show():
     for mode in ("1", "I;16", "LA", "RGB", "RGBA"):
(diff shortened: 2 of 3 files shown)


alv2017:
I ran all the workflows in my repository, all the workflows except Windows passed.

hugovk:
The Windows failure was a bug in `actions/setup-python`, now fixed: https://github.com/actions/setup-python/issues/328. I've restarted the Windows jobs.

alv2017:
It failed on  Python pypy-3.7 x64:
Process completed with exit code -1073741819.
No files were found with the provided path: Tests/errors. No artifacts will be uploaded.

radarhere:
I reran the Windows jobs again, and they have all passed now.

alv2017:
@hugovk, @radarhere 

Please let me know if this PR has a chance of being accepted. Thanks!

Tests/test_imageshow.py
```diff
@@ -41,9 +41,11 @@ def show_image(self, image, **options):
     ImageShow._viewers.pop(0)
 
 
-@pytest.mark.skipif(
-    not on_ci() or is_win32(),
-    reason="Only run on CIs; hangs on Windows CIs",
+@pytest.mark.skip(
+    reason="""Current implementation of some viewers requires manual closing of an image,
+        because of that the tests calling show() method will hang infinitely.
+        Please also note that this test duplicates test_viewer_show() test.
```
radarhere, line 47: I don't think it is a duplicate. `test_viewer_show()` registers a dummy viewer, to check that `Viewer.show()` calls `show_image()`. `test_show()`  runs an actual viewer.

alv2017:
@radarhere I corrected the test skip message.

src/PIL/ImageShow.py
```diff
@@ -234,150 +235,80 @@ def show_file(self, path=None, **options):
...
+
+        path = quote(path)
+        command = self.get_command(path, **options)
+
+        kwargs = {
+            "stdout": subprocess.PIPE,
+        }
+
+        th = threading.Thread(
+            target=subprocess.run, args=(command.split(),), kwargs=kwargs, name=path
```
radarhere, line 247: We've recently had to address the possibility of [the file path having a space in it](https://github.com/python-pillow/Pillow/pull/3450#issuecomment-1009608418). This would `split()` incorrectly in such a situation, yes?

src/PIL/ImageShow.py
```diff
@@ -234,150 +235,80 @@ def show_file(self, path=None, **options):
...
+        return command, executable
 
 
 class GmDisplayViewer(UnixViewer):
     """The GraphicsMagick ``gm display`` command."""
 
     def get_command_ex(self, file, **options):
         executable = "gm"
-        command = "gm display"
+        command = f"gm display {quote(file)}"
```
radarhere, line 285: As I've said before, I consider this to be a backwards incompatible change.

Tests/helper.py
```diff
@@ -27,8 +27,7 @@
     class test_image_results:
         @staticmethod
         def upload(a, b):
-            a.show()
-            b.show()
```
radarhere, line 31: What is the reason for this change?

alv2017: If it is hit, the tests will hang infinitely. I do not see any good reason for a.show(), b.show(): we are interested in return value of the **upload** function, and return value here is None.

src/PIL/ImageShow.py
alv2017: I think that it is more convenient to keep a viewer opening shell command at one place. As excuse I will add, that get_command_ex() is a helper function, and the interface of the function was preserved.

src/PIL/ImageShow.py
alv2017: 1) I do not think that there is a a split problem, because the file(path) is quoted, and the path always be represented as a single entity.
2) I wouldn't be concerned about users having a temporary path "foo bar baz", I think that they will be able to sort this out.

alv2017: The solution was to run the viewer command as a shell command. User input has been sanitized in **get_command_ex** stage.  As a result (a) path problem solved; (b) possible security problem solved;
```python
        kwargs = {"shell": True, "stdout": subprocess.PIPE, "stderr": subprocess.PIPE}

        th = threading.Thread(
            target=subprocess.run, args=(command,), kwargs=kwargs, name=path
        )
```

Tests/helper.py
nulano: Huh? This is only used if there is an environment variale `SHOW_ERRORS` set, which means the user is asking to be shown failed image comparisons. Removing this breaks that functionality. Also, the result of the upload function is irrelevant as there is a check in place to see if it is defined. If it isn't defined, it is never called anyway, so if this were to be removed, you would remove the whole if block.

radarhere:
I've created #6039 as an alternate version of this. The primary difference is that it doesn't cause Python to hang while waiting for the image viewer to close. See what you think.

hugovk:
https://github.com/python-pillow/Pillow/pull/6045 has been merged instead. Thank you for a way for the suggestion!

PR #6030, opened 2022-02-06
Added get_photoshop_blocks() to parse Photoshop TIFF tag

### LINKED ISSUE #6025: Photoshop EXIF tags are not read
Hello, Hello!

I need to detect if a file has Photoshop clipping paths present using the corresponding EXIF tags [(0x07d0 - 0xbb5)](https://exiftool.org/TagNames/Photoshop.html) However Pillow seems to ignore these tags.

```python
from PIL import Image

img=Image.open('mypsdTiff.tiff')
psd_paths = range(0x07d0, 0x0bb5)
exif_tags = list(img.getexif().keys())
print ("Yay!" if any(path in exif_tags for path in psd_paths) else "Nay!")
```

Result: "Nay!"
Expected Result: "Yay!"

The tag in the file is the 0x07d0. I did test it out with exiftool.

I have the feeling I have to add this range to PIL.TiffTags?

OS: Windows 11 (and Server 2008)
Python Version: 3.9 (and 3.10)
Pillow version: 9.01

radarhere:
Resolves #6025

TIFF tag 34377 is a ["Collection of Photoshop 'Image Resource Blocks'"](https://www.awaresystems.be/imaging/tiff/tifftags/photoshop.html). This PR adds a method, `getphotoshopblocks()`, to determine what blocks are present in an image. https://www.awaresystems.be/imaging/tiff/tifftags/docs/photoshopthumbnail.html was useful in writing this code.

This PR does not go so far as to format the information within each block. Instead, the bytes are simply deposited within a dictionary under a "data" key. This will allow us to add in more useful information in the future without breaking backwards compatibility. https://www.adobe.com/devnet-apps/photoshop/fileformatashtml/#50577409_pgfId-1037727 will be a good resource at that time.

### DIFF
--- a/src/PIL/TiffImagePlugin.py
+++ b/src/PIL/TiffImagePlugin.py
@@ -41,6 +41,7 @@
 import io
 import itertools
 import logging
+import math
 import os
 import struct
 import warnings
@@ -49,6 +50,8 @@ from fractions import Fraction
 from numbers import Number, Rational
 
 from . import Image, ImageFile, ImageOps, ImagePalette, TiffTags
+from ._binary import i16be as i16
+from ._binary import i32be as i32
 from ._binary import o8
 from .TiffTags import TYPES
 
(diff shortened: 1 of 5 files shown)


src/PIL/TiffImagePlugin.py
```diff
@@ -1128,6 +1131,20 @@ def getxmp(self):
         """
         return self._getxmp(self.tag_v2[700]) if 700 in self.tag_v2 else {}
 
+    def getphotoshopblocks(self):
```
hugovk, line 1134: Naming: I see this follows the other one here -- `getxmp` -- but how about `get_photoshop_blocks` for this new one?

Docs: let's at least add a docstring and release notes. Would any other documentation be useful?

(Docstring goes here:)

image

https://pillow--6030.org.readthedocs.build/en/6030/reference/plugins.html#PIL.TiffImagePlugin.TiffImageFile.getphotoshopblocks

hugovk:
Thank you for the other release notes too!

PR #6033, opened 2022-02-07
Add support for 1-bit images to ImageOps.invert()

dov:
Currently `ImageOps.invert(Img)` throws an error when called with a binary image. This pull request fixes that.

### DIFF
--- a/src/PIL/ImageOps.py
+++ b/src/PIL/ImageOps.py
@@ -520,6 +520,11 @@ def invert(image):
     :param image: The image to invert.
     :return: An image.
     """
+    # Special case for 1-bit images using python translate() method
+    if image.mode == '1':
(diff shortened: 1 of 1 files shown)


radarhere:
Hi. I've created PR #6034 as an alternative to this. Rather than adding dedicated code to `invert()`, I adjust `_lut`, and so allow for several ImageOps method to support 1 mode images.

radarhere:
Since you've determined that #6034 is [faster](https://github.com/python-pillow/Pillow/pull/6034#issuecomment-1032269430), do you mind if this is closed?

dov:
> Since you've determined that #6034 is [faster](https://github.com/python-pillow/Pillow/pull/6034#issuecomment-1032269430), do you mind if this is closed?

Sure. I'm closing it.

PR #6039, opened 2022-02-09
Wait until executable has finished to remove file on Unix

radarhere:
Helps #5945 and #5976. Alternative to #6021

Three changes here.
1. As pointed out in #6021, `ImageShow.show()` does not return `True` and `False` like the docstring states, it returns 0 and 1. So this PR updates the docstring.
2. As suggested out in https://github.com/python-pillow/Pillow/pull/6005#discussion_r796396077, `display -title` looks preferable to `display -name`.

Screen Shot 2022-02-09 at 4 17 44 pm
Screen Shot 2022-02-09 at 4 18 04 pm

3. Rather than waiting an arbitrary 20 seconds on Unix, wait until the application has finished showing the file to remove it. #6021 also does this, but in that version, Python script hangs until the application is closed. This version does not.
Since `xdg-open` does not wait until the application is closed, this PR instead uses `xdg-mime` to determine what application would be selected by `xdg-open`, and executes it directly.
For the record - I'm also not opposed to the [suggested alternative](https://github.com/python-pillow/Pillow/issues/5945#issuecomment-1028845574) of just abandoning the idea of manually deleting temporary files and leaving it up to the operating system.

### DIFF
--- a/src/PIL/ImageShow.py
+++ b/src/PIL/ImageShow.py
@@ -50,12 +50,12 @@ def show(image, title=None, **options):
     :param image: An image object.
     :param title: Optional title. Not all viewers can display the title.
     :param \**options: Additional viewer options.
-    :returns: ``1`` if a suitable viewer was found, ``0`` otherwise.
+    :returns: ``True`` if a suitable viewer was found, ``False`` otherwise.
     """
     for viewer in _viewers:
         if viewer.show(image, title=title, **options):
-            return 1
-    return 0
+            return True
+    return False
 
 
 class Viewer:
--- a/src/PIL/ImageShow.py
+++ b/src/PIL/ImageShow.py
@@ -50,7 +50,7 @@ def show(image, title=None, **options):
     :param image: An image object.
     :param title: Optional title. Not all viewers can display the title.
     :param \**options: Additional viewer options.
-    :returns: ``True`` if a suitable viewer was found, ``False`` otherwise.
+    :returns: ``1`` if a suitable viewer was found, ``0`` otherwise.
     """
     for viewer in _viewers:
         if viewer.show(image, title=title, **options):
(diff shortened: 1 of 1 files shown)


nulano:
> ImageShow.show() does not return True and False like the docstring states, it returns 0 and 1. So this PR updates the docstring.

Would it be a breaking change to instead change the returned value? I think it is quite unlikely that any code depends on the exact type of the returned value, since bool is a subclass of int:

```python
>>> False == 0
True
>>> True == 1
True
>>> True + True
2
>>> type(True)
<class 'bool'>
>>> issubclass(bool, int)
True
```

radarhere:
You make a good argument. Ok, I've pushed that change.

alv2017:
@radarhere I'm upvoting the idea of leaving file deletion to OS. It seems to be a valid solution.

https://www.fosslinux.com/41739/linux-tmp-directory-everything-you-need-to-know.htm

radarhere:
@alv2017 while we may end up choosing that option, I'm still interested to know - do you have any objections to this PR as it currently stands?

alv2017:
@radarhere, I will make my objections tomorrow :)

radarhere:
Here is my recording of it working with GmDisplayViewer for me.

https://user-images.githubusercontent.com/3112309/153544946-2787b2ae-adfd-4f85-a1c5-912dc12d6ce1.mov

When you say that it fails, could you give some more detail? When you say that XDGViewer fails, what image viewer is it opening?

radarhere:
Ok, the premise of my change - that commands like `eog` stay open until the image viewer closed - turns out to only be valid if the image viewer is being opened for the first time. Otherwise, the command finishes immediately.

I've created PR #6045 instead.

PR #6045, opened 2022-02-11
Do not automatically remove temporary ImageShow files on Unix

radarhere:
Helps #5945 and #5976. Alternative to #6021

Three changes here.
1. As pointed out in #6021, `ImageShow.show()` does not return `True` and `False` like the docstring states, it returns 0 and 1. This PR updates the return values to match the docstring.
2. As suggested out in https://github.com/python-pillow/Pillow/pull/6005#discussion_r796396077, `display -title` looks preferable to `display -name`.

Screen Shot 2022-02-09 at 4 17 44 pm
Screen Shot 2022-02-09 at 4 18 04 pm

3. Given that there have been [requests](https://github.com/python-pillow/Pillow/issues/5945#issuecomment-1028760062) for [multiple images to be shown at once](https://github.com/python-pillow/Pillow/issues/5976#issuecomment-1017214685) and [to be viewed after 20s](https://github.com/python-pillow/Pillow/issues/5976#issuecomment-1018444936), a possible solution to these requests is to not manually remove the temporary ImageShow files on Unix.

### DIFF
--- a/docs/releasenotes/9.1.0.rst
+++ b/docs/releasenotes/9.1.0.rst
@@ -48,6 +48,13 @@ breaking backwards compatibility.
 Other Changes
 =============
 
+ImageShow temporary files on Unix
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+When calling :py:meth:`~PIL.Image.Image.show` or using :py:mod:`~PIL.ImageShow`,
+a temporary file is created from the image. On Unix, Pillow will no longer delete these
+files, and instead leave it to the operating system to do so.
+
 Image._repr_pretty_
 ^^^^^^^^^^^^^^^^^^^
 
--- a/Tests/test_imageshow.py
+++ b/Tests/test_imageshow.py
@@ -51,6 +51,16 @@ def test_show():
         assert ImageShow.show(im)
 
 
+def test_show_without_viewers():
+    viewers = ImageShow._viewers
+    ImageShow._viewers = []
+
+    im = hopper()
+    assert not ImageShow.show(im)
+
+    ImageShow._viewers = viewers
+
+
 def test_viewer():
     viewer = ImageShow.Viewer()
 
--- a/src/PIL/ImageShow.py
+++ b/src/PIL/ImageShow.py
@@ -126,16 +126,6 @@ class Viewer:
         os.system(self.get_command(path, **options))
         return 1
 
-    def _remove_path_after_delay(self, path):
-        subprocess.Popen(
-            [
-                sys.executable,
-                "-c",
-                "import os, sys, time; time.sleep(20); os.remove(sys.argv[1])",
-                path,
-            ]
-        )
-
 
 # --------------------------------------------------------------------
 
@@ -190,7 +180,14 @@ class MacViewer(Viewer):
             else:
                 raise TypeError("Missing required argument: 'path'")
         subprocess.call(["open", "-a", "Preview.app", path])
-        self._remove_path_after_delay(path)
+        subprocess.Popen(
+            [
+                sys.executable,
+                "-c",
+                "import os, sys, time; time.sleep(20); os.remove(sys.argv[1])",
+                path,
+            ]
+        )
         return 1
 
 
@@ -235,7 +232,6 @@ class XDGViewer(UnixViewer):
             else:
                 raise TypeError("Missing required argument: 'path'")
         subprocess.Popen(["xdg-open", path])
-        self._remove_path_after_delay(path)
         return 1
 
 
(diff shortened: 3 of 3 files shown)


alv2017 (commented):
@radarhere 

1. I have tested the functionality (Ubuntu 20.04): show() function worked well for all the viewers except GmDisplayViewer. In case of single file the image was sometimes shown and sometimes not shown. For multiple images, in my case 3 images, only two images showed up. I do not have any good explanation for such behavior of GmDisplayViewer... 

2. Good news: the temporary image files get deleted on reboot (Ubuntu 20.04)

3. Code: 
(a) I think there is a backwards compatibility issue with viewers interface, I'm talking here about get_command() function. This function is redundant for Mac and Unix viewers. There is also a consistency issue if one compares Windows viewer implementation vs Mac and Unix.
(b) I would say that DRY principle is violated for Unix viewers image_show() method. 
(c) Unix viewers get_command_ex() method became redundant, and it doesn't reflect the actually implemented functionality.
(d) get_command() method does not reflect actually implemented functionality. 
**Summary:** the key thing here is to make a decision about viewers interface: either implement the new approach to all the viewers, or stick with the old interface, and apply it to all the viewers.

radarhere:
3. a) When you say there is a backwards compatibility issue with `get_command()`, the only change I've made to `get_command()` is swapping "title" for "name" in `DisplayViewer`. Do you mean that `get_command()` should not remove the file on Unix? I've added a commit for that.
Why is the function redundant for Mac and Unix viewers?
What is the consistency issue?
b) Did you mean `show_image()` on Viewer?
c) Could you elaborate on this?
d) Does my latest commit address this?

alv2017:
@radarhere 

In the Viewer class viewers interface is defined. As per this definition the **get_command()** is expected to return specific viewer's display command, the return value of the **get_command()** function is expected to be used by **show_file()** method.
In my opinion, current PR violates this logic.

Question: What is the purpose of **get_command()** and **get_command_ex()** in this PR?

radarhere:
I am confused by your comment.

I have proposed minimal changes to `get_command` and `get_command_ex`
- I swapped "title" for "name" in `DisplayViewer` as per [your suggestion](https://github.com/python-pillow/Pillow/pull/6005#discussion_r796396077)
- I removed the `rm` command from `get_command` in UnixViewer

Are you just having a conversation about ImageShow in general, rather than reviewing my changes here? It's ok if you are, I just don't feel like that's clear.

If you're asking why `get_command()` and `get_command_ex()` exist at all, they have been present since PIL, so we did not add them in. I personally find it hard to think of a use case for them, but as they have some functionality and don't cause problems, I'm more inclined to leave them as is. If you think they should be removed, please create a separate PR that adds a deprecation warning to those methods (like #5959). If that change is accepted, then they can be removed in a future Pillow release.

alvtech2017:
@radarhere 
1) I just expressed concerns about the part of the code that I have some doubts. (namely repetitive show_file methods, and seemingly redundant get_command() method.

2) Please note, that I overlooked the changes you did on Feb 2. This might be one of the reasons my comment seems confusing to you. :)

3) In general I'm Ok with the changes you made, and  yes you are right, it will be easier to explain what I mean by submitting a pull request.

alv2017:
Sorry, I forgot to switch to my regular account :smile:

hugovk (approved):
> 1. `ImageShow.show()` does not return `True` and `False` like the docstring states, it returns 0 and 1.

```pycon
>>> True == 1
True
>>> False == 0
True
>>>
```

;)

Returning `True` or `False` is nicer. 👍 

---

Let's include in the release notes too the fact they're no longer automatically* removed, so people know it's their job to clear it out if for example they show lots of images and don't reboot often.


(* Shall we rename  "manually" -> "automatically" in the PR title?)

docs/releasenotes/9.1.0.rst
```diff
@@ -48,6 +48,13 @@ breaking backwards compatibility.
 Other Changes
 =============
 
+ImageShow temporary files
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+When calling :py:meth:`~PIL.Image.Image.show` or using :py:mod:`~PIL.ImageShow`,
+a temporary file is created from the image. Pillow will no longer delete these files,
+and instead leave it to the operating system to do so.
```
hugovk, line 56: Only on Linux, right?

radarhere: You're right! Thanks for catching.

hugovk:
Thanks!

alv2017:
I suggest to close those issues as well: #5976, #5945

PR #6050, opened 2022-02-13
Only skip test if libimagequant is earlier than 4 on ppc64le

radarhere:
`test_libimagequant_quantize` is currently [skipped](https://github.com/python-pillow/Pillow/blob/cac305f8d2db5eca3ce18f8d334a9ff35f9c7fde/Tests/test_image_quantize.py#L20-L21) on ppc64le.

And for good reason, because if you [remove the skip](https://github.com/radarhere/Pillow/commit/12e2e97599209177550a8c382d64f6e03b910b29), it [fails](https://github.com/radarhere/Pillow/runs/5167378728).

Or rather, it did, because [that failure used libimagequant 2.12](https://github.com/radarhere/Pillow/runs/5167378728#step:6:82), and since https://github.com/python-pillow/docker-images/pull/138 effectively upgraded libimagequant on ppc64le to 4, it now passes, as seen in this PR.

### DIFF
--- a/Tests/test_image_quantize.py
+++ b/Tests/test_image_quantize.py
@@ -1,8 +1,9 @@
 import pytest
+from packaging.version import parse as parse_version
 
-from PIL import Image
+from PIL import Image, features
 
-from .helper import assert_image_similar, hopper, is_ppc64le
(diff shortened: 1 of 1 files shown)


hugovk:
Nice!

PR #6057, opened 2022-02-15
Merged _MODE_CONV typ into ImageMode as typestr

### LINKED ISSUE #5990: Refactor _MODE_CONV into ImageModes
Currently, Image.py has a private variable called `_MODE_CONV` that looks like this:

https://github.com/python-pillow/Pillow/blob/8e4a618b550b1a67e31c5362552db74b2cc72841/src/PIL/Image.py#L211-L239

I find this constant quite useful, and I was wondering if this info can be made part of `ImageMode` and exposed publically, so that I can (confidently) use it in downstream code. I think it would also make this codebase cleaner overall :D

For my use case, I am pre-allocating buffers into which I later want to decode image data, i.e., I have something similar to [`_conv_type_shape`](https://github.com/python-pillow/Pillow/blob/8e4a618b550b1a67e31c5362552db74b2cc72841/src/PIL/Image.py#L242) and would like to keep it in sync with the modes supported by pillow (without having to update manually).

radarhere:
Resolves #5990

https://github.com/python-pillow/Pillow/issues/5990#issuecomment-1039873428 pointed out that the second value in [`_MODE_CONV`](https://github.com/python-pillow/Pillow/blob/7b3f28047fe008d70bda1938e3178da388ecaa64/src/PIL/Image.py#L211-L239), [`extra`](https://github.com/python-pillow/Pillow/blob/7b3f28047fe008d70bda1938e3178da388ecaa64/src/PIL/Image.py#L243), is the number of bands in the mode. Rather than writing those out in `_MODE_CONV`, this could instead be derived from ImageMode.

The user in the issue would also like the NumPy typestr from `_MODE_CONV` available through a public API, and suggested ImageMode. From an organisational point of view for Pillow's code, that seems like a more logical location for that information, and it would be neater, since it then means that `_MODE_CONV` can be removed. The NumPy typestr could then be accessed by a user through `ImageMode.getmode(mode).typestr`.

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -203,10 +203,7 @@ ENCODERS = {}
 # --------------------------------------------------------------------
 # Modes
 
-if sys.byteorder == "little":
-    _ENDIAN = "<"
-else:
-    _ENDIAN = ">"
+_ENDIAN = "<" if sys.byteorder == "little" else ">"
 
 
 def _conv_type_shape(im):
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -208,7 +208,7 @@ _ENDIAN = "<" if sys.byteorder == "little" else ">"
 
 def _conv_type_shape(im):
     m = ImageMode.getmode(im.mode)
-    shape = (im.size[1], im.size[0])
+    shape = (im.height, im.width)
     extra = len(m.bands)
     if extra != 1:
         shape += (extra,)
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -208,44 +208,14 @@ if sys.byteorder == "little":
 else:
     _ENDIAN = ">"
 
-_MODE_CONV = {
-    # official modes
-    "1": "|b1",  # Bits need to be extended to bytes
-    "L": "|u1",
-    "LA": "|u1",
-    "I": _ENDIAN + "i4",
-    "F": _ENDIAN + "f4",
-    "P": "|u1",
-    "RGB": "|u1",
-    "RGBX": "|u1",
-    "RGBA": "|u1",
-    "CMYK": "|u1",
-    "YCbCr": "|u1",
-    "LAB": "|u1",  # UNDONE - unsigned |u1i1i1
-    "HSV": "|u1",
-    # I;16 == I;16L, and I;32 == I;32L
-    "I;16": "<u2",
-    "I;16B": ">u2",
-    "I;16L": "<u2",
-    "I;16S": "<i2",
-    "I;16BS": ">i2",
-    "I;16LS": "<i2",
-    "I;32": "<u4",
-    "I;32B": ">u4",
-    "I;32L": "<u4",
-    "I;32S": "<i4",
-    "I;32BS": ">i4",
-    "I;32LS": "<i4",
-}
-
 
 def _conv_type_shape(im):
-    typ = _MODE_CONV[im.mode]
-    extra = len(im.getbands())
-    if extra == 1:
-        return (im.size[1], im.size[0]), typ
-    else:
-        return (im.size[1], im.size[0], extra), typ
+    m = ImageMode.getmode(im.mode)
+    shape = (im.size[1], im.size[0])
+    extra = len(m.bands)
+    if extra != 1:
+        shape += (extra,)
+    return shape, m.typestr
 
 
 MODES = ["1", "CMYK", "F", "HSV", "I", "L", "LAB", "P", "RGB", "RGBA", "RGBX", "YCbCr"]
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -210,38 +210,39 @@ else:
 
 _MODE_CONV = {
     # official modes
-    "1": ("|b1", None),  # Bits need to be extended to bytes
-    "L": ("|u1", None),
-    "LA": ("|u1", 2),
-    "I": (_ENDIAN + "i4", None),
-    "F": (_ENDIAN + "f4", None),
-    "P": ("|u1", None),
-    "RGB": ("|u1", 3),
-    "RGBX": ("|u1", 4),
-    "RGBA": ("|u1", 4),
-    "CMYK": ("|u1", 4),
-    "YCbCr": ("|u1", 3),
-    "LAB": ("|u1", 3),  # UNDONE - unsigned |u1i1i1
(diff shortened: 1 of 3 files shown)


src/PIL/Image.py
```diff
@@ -208,43 +208,14 @@ def isImageType(t):
...
-    typ, extra = _MODE_CONV[im.mode]
-    if extra is None:
-        return (im.size[1], im.size[0]), typ
-    else:
-        return (im.size[1], im.size[0], extra), typ
+    m = ImageMode.getmode(im.mode)
+    shape = (im.size[1], im.size[0])
+    extra = len(m.bands)
+    if extra != 1:
+        shape += (extra,)
```
FirefoxMetzger, line 217:
```suggestion
    shape = (im.size[1], im.size[0])  # im.size[::-1]
    shape = shape if len(m.bands) == 1 else (*shape, len(m.bands))
```

Not sure if you find this syntax neater. Both options look viable to me.

src/PIL/ImageMode.py
```diff
@@ -36,43 +39,52 @@ def getmode(mode):
...
-            "BGR;24": ("RGB", "L", ("B", "G", "R")),
-            "BGR;32": ("RGB", "L", ("B", "G", "R")),
-            "LA": ("L", "L", ("L", "A")),
-            "La": ("L", "L", ("L", "a")),
-            "PA": ("RGB", "L", ("P", "A")),
+            "RGBa": ("RGB", "L", ("R", "G", "B", "a"), None),
+            "BGR;15": ("RGB", "L", ("B", "G", "R"), None),
+            "BGR;16": ("RGB", "L", ("B", "G", "R"), None),
+            "BGR;24": ("RGB", "L", ("B", "G", "R"), None),
+            "BGR;32": ("RGB", "L", ("B", "G", "R"), None),
```
FirefoxMetzger, line 63:
```suggestion
            "BGR;32": ("RGB", "L", ("B", "G", "R"), "u4"),
```

Not sure if this is big-endian or little-endian (nor if pillow would allow conversion from this mode to array), but I guess it should be a 4-byte uint.

src/PIL/ImageMode.py
```diff
@@ -36,43 +39,52 @@ def getmode(mode):
...
-            "BGR;15": ("RGB", "L", ("B", "G", "R")),
-            "BGR;16": ("RGB", "L", ("B", "G", "R")),
-            "BGR;24": ("RGB", "L", ("B", "G", "R")),
-            "BGR;32": ("RGB", "L", ("B", "G", "R")),
-            "LA": ("L", "L", ("L", "A")),
-            "La": ("L", "L", ("L", "a")),
-            "PA": ("RGB", "L", ("P", "A")),
+            "RGBa": ("RGB", "L", ("R", "G", "B", "a"), None),
+            "BGR;15": ("RGB", "L", ("B", "G", "R"), None),
+            "BGR;16": ("RGB", "L", ("B", "G", "R"), None),
```
FirefoxMetzger, line 61:
```suggestion
            "BGR;16": ("RGB", "L", ("B", "G", "R"), "u2"),
```

Not sure if this is big-endian or little-endian (nor if pillow would allow conversion from this mode to array), but I guess it should be a 2-byte uint.

src/PIL/ImageMode.py
```diff
@@ -13,18 +13,21 @@
 # See the README file for information on usage and redistribution.
 #
 
+from PIL import Image
```
FirefoxMetzger, line 16: This creates a circular import (not sure why it didn't crash CI). `PIL.Image` now uses `ImageMode.getmode` in `_conv_dtype_shape` and `PIL.ImageMode` uses `PIL.Image._ENDIAN`. As a result, both modules import each other.

Here is the import in the pillow module:

https://github.com/python-pillow/Pillow/blob/7b3f28047fe008d70bda1938e3178da388ecaa64/src/PIL/Image.py#L50

Do you think this will become problematic?

FirefoxMetzger (commented):
I left some small, perhaps opinionated, comments. Looking good from my side otherwise.

src/PIL/ImageMode.py
radarhere: Ok, I've updated the entry for this.

src/PIL/ImageMode.py
radarhere: Ok, I've updated the entry for this.

src/PIL/ImageMode.py
radarhere: Thanks for catching! I've pushed a commit to avoid the circular dependency.

src/PIL/Image.py
```diff
@@ -203,48 +203,16 @@ def isImageType(t):
...
-    "I;16BS": (">i2", None),
-    "I;16LS": ("<i2", None),
-    "I;32": ("<u4", None),
-    "I;32B": (">u4", None),
-    "I;32L": ("<u4", None),
-    "I;32S": ("<i4", None),
-    "I;32BS": (">i4", None),
-    "I;32LS": ("<i4", None),
-}
+_ENDIAN = "<" if sys.byteorder == "little" else ">"
```
FirefoxMetzger, line 206: You know better than me for what this is used later on, but if you only need this to set the native byte order for numpy, you could also use `_ENDIAN= "="` (`=` for native) and have numpy do the rest. Of course, you could also set it directly in the typestring, e.g. `=u4`.

radarhere: While that is listed at https://numpy.org/doc/stable/reference/generated/numpy.dtype.byteorder.html, it is not listed under "typestr" at https://numpy.org/doc/stable/reference/arrays.interface.html#object.__array_interface__. So I'm skeptical that "=" would be valid.

src/PIL/Image.py
radarhere: It does seem to be a matter of personal preference, but `shape = shape` looks odd to me.

I have clarified the first line slightly by changing it to `(im.height, im.width)`.

src/PIL/Image.py
FirefoxMetzger: Good catch. So far I used them interchangeably since I never noticed something being broken, but I agree it is better to err on the side of caution and not use it.

src/PIL/Image.py
FirefoxMetzger: that's fair.

hugovk:
Thanks both!

PR #6061, opened 2022-02-16
Added rawmode argument to Image.getpalette()

### LINKED ISSUE #6059: Add rawmode argument to Image.getpalette()
[putpalette](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.putpalette) has a `rawmode` argument which accepts `"RGBA"`, allowing a palette with some transparent colors. Unfortunately [getpalette](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.getpalette) does not have a similar argument.

In most but not all of the images I'm working with, the first two elements of the palette are both (0, 0, 0), with the first element corresponding to the transparent color (0, 0, 0, 0) and the second corresponding to (0, 0, 0, 255). I was able to figure this out through experimentation, but would prefer to be more precise, as I have a few images where this is not the case.

Current code:

```python
def recolor_poodle(original_path, new_path):
    image = Image.open(original_path)
    image = image.convert("P")
    unused_palette = 127
    ImageDraw.floodfill(image, (31, 21), unused_palette)
    original_palette = image.getpalette()
    new_palette = [0, 0, 0, 0]
    for palette_index in range(1, len(original_palette) // 3):
        old_rgb = [
            original_palette[3 * palette_index + channel_index]
            for channel_index in range(3)
        ]
        channels = [hex(channel)[2:].rjust(2, "0") for channel in old_rgb]
        # padding
        old_color = f"{channels[0]}{channels[1]}{channels[2]}"
        new_color = OLD_TO_NEW.get(old_color)
        if new_color is None:
            new_palette += old_rgb + [255]
        else:
            new_palette += [
                int(new_color[0:2], base=16),
                int(new_color[2:4], base=16),
                int(new_color[4:6], base=16),
                255,
            ]

    arr = numpy.asarray(image)

    im2 = Image.fromarray(arr, mode="P")
    im2.putpalette(new_palette, rawmode="RGBA")
    im2.save(new_path)
```

Desired code:

```python
def recolor_poodle(original_path, new_path):
    image = Image.open(original_path)
    image = image.convert("P")
    unused_palette = 127
    ImageDraw.floodfill(image, (31, 21), unused_palette)
    original_palette = image.getpalette(rawmode="RGBA")
    new_palette = []
    for palette_index in range(len(original_palette) // 4):
        # Skip mapping transparent colors.
        if original_palette[4 * palette_index + 3] == 0:
            new_palette += [0, 0, 0, 0]
            continue
        old_rgb = [
            original_palette[4 * palette_index + channel_index]
            for channel_index in range(3)
        ]
        channels = [hex(channel)[2:].rjust(2, "0") for channel in old_rgb]
        # padding
        old_color = f"{channels[0]}{channels[1]}{channels[2]}"
        new_color = OLD_TO_NEW.get(old_color)
        if new_color is None:
            new_palette += old_rgb + [255]
        else:
            new_palette += [
                int(new_color[0:2], base=16),
                int(new_color[2:4], base=16),
                int(new_color[4:6], base=16),
                255,
            ]

    arr = numpy.asarray(image)

    im2 = Image.fromarray(arr, mode="P")
    im2.putpalette(new_palette, rawmode="RGBA")
    im2.save(new_path)
```

radarhere:
Resolves #6059

At the moment, `getpalette()` only returns the palette in RGB mode. This PR adds a `rawmode` argument that allows the user to specify the mode that the palette shoud be returned as.

I've also allowed the `rawmode` to be `None`, in case the user would like the palette returned without transformation.


CookiePLMonster:
Would it make sense to also allow converting via a `.tobytes()` call, much like it is done for Images?

radarhere:
Just for clarity - you've talking about `ImagePalette.tobytes()`, and you're suggesting that users be able to call `ImagePalette.tobytes(rawmode)` (encoders are for images, not palettes).

Bear with me as I try and put this into a coherent thought. I don't think there's an inherit pull for `Image.tobytes()` and `ImagePalette.tobytes()` to mirror each other. For one thing, the first argument of `Image.tobytes()` is "encoder" as you say, which is not something that applies to an ImagePalette. For another, these aren't the only `tobytes()` methods - there's also `ImageCmsProfile.tobytes()`, `Dib.tobytes()`, `ImageFileDirectory_v2.tobytes()` and `Exif.tobytes()`.

After this PR, If you wanted to convert the data from an ImagePalette to a particular mode, you could
```python
im.putpalette(palette)
im.getpalette(rawmode)
```

I would think of ImagePalette as being for the analysis and modification of a palette.

I can imagine one wanting to
1. Load an image
2. Modify the palette data in ImagePalette
3. Attach the modified palette to the image proper through `load()` (or one of the many other methods that call `load()` internally)
4. Convert the palette data to a certain mode, because you want to write it to file, alongside the image data.

But are you saying you have a scenario where you would like to skip step 3? So you have a situation where you would like to load an image, modify the palette, take a copy of the palette for external purposes, and no longer use the image with it's modified palette?

CookiePLMonster:
Fair enough - I only suggested `.tobytes()` for consistency of the API, but your point of encoders not applying to palettes makes sense. I think the current approach is fine then.

> But are you saying you have a scenario where you would like to skip step 3? So you have a situation where you would like to load an image, modify the palette, take a copy of the palette for external purposes, and no longer use the image with it's modified palette?

Not sure, I imagine that if there is a need for this the user may simply deep copy the palette. But to be precise, my idea was more along the lines of:
1. Load image
2. Get a palette as RGBA regardless of whether the image has alpha or not
3. Write results to file

However, sounds like a modified `getpalette()` satisfies that requirement just fine too.

src/PIL/Image.py
```diff
@@ -1401,19 +1401,24 @@ def getim(self):
         self.load()
         return self.im.ptr
 
-    def getpalette(self):
+    def getpalette(self, rawmode="RGB"):
         """
         Returns the image palette as a list.
 
+        :param rawmode: The mode in which to return the palette. ``None`` will
+           return the palette in its current mode.
```
hugovk, line 1409: Let's add a "new in version" like in https://github.com/python-pillow/Pillow/blob/caa0deb4c76442b3f3036dbf4d59b2922deaaaab/src/PIL/ImageOps.py#L83

And should this go in release notes?

hugovk (approved):

hugovk:
Thanks!

PR #6064, opened 2022-02-17
Added multiple image merging example

ruhulmus:


docs/example/img_merge.py
```diff
@@ -0,0 +1,15 @@
...
+image2 = Image.open("img/img2.jpg")
+
+# resize, first image
+image1 = image1.resize((426, 240))
+image1_size = image1.size
+image2_size = image2.size
+new_image = Image.new("RGB", (2 * image1_size[0], image1_size[1]), (250, 250, 250))
+new_image.paste(image1, (0, 0))
+new_image.paste(image2, (image1_size[0], 0))
+new_image.save("img/merged_image.jpg", "JPEG")
```
radarhere, line 14:
```suggestion
new_image.save("img/merged_image.jpg")
```
Specifying the format isn't necessary when there is a file extension.

docs/example/img_merge.py
```diff
@@ -0,0 +1,15 @@
...
+
+# Read the two images
+image1 = Image.open("img/img1.jpg")
+image2 = Image.open("img/img2.jpg")
+
+# resize, first image
+image1 = image1.resize((426, 240))
+image1_size = image1.size
+image2_size = image2.size
+new_image = Image.new("RGB", (2 * image1_size[0], image1_size[1]), (250, 250, 250))
```
radarhere, line 11:
```suggestion
new_image = Image.new("RGB", (2 * image1_size[0], image1_size[1]))
```
If your purpose is to combine two RGB images, then you don't need a background color.

radarhere:
Hi. Could you talk a bit about why you think this is a valuable addition? It could be interesting to have your perspective.

Also, have you seen https://pillow.readthedocs.io/en/stable/handbook/tutorial.html? That is typically where this sort of example lives, rather than docs/example. We don't explicitly have this situation covered there. The closest we have I think is https://pillow.readthedocs.io/en/stable/handbook/tutorial.html#rolling-an-image

ruhulmus (commented):
added new_image.show()

radarhere:
> added new_image.show()

I don't follow this comment at all. If I was meant to, please elaborate.

hugovk:
Also, we cannot include copyright images without permission from the copyright holders.


https://www.alamy.com/stock-photo-common-kingfisher-alcedo-atthis-bird-perched-on-a-branch-with-green-135484692.html

https://www.alamy.com/mountain-slide-of-austria-image429226842.html

ruhulmus:
@hugovk
I understand .. but I have downloaded those images for free.
you can check from here :

https://unsplash.com/photos/vUNQaTtZeOo

https://unsplash.com/photos/_RBcxo9AU-U

hugovk:
Please could you also answer these questions? https://github.com/python-pillow/Pillow/pull/6064#issuecomment-1042687513

docs/example/img_merge.py
radarhere: You marked this as resolved. Could you explain why you disagree with my suggestion?

docs/example/img_merge.py
radarhere: You marked this as resolved. Could you explain why you disagree with my suggestion?

radarhere:
I've created #6129 as an alternative to this.

radarhere:
#6129 has been merged instead. See https://pillow.readthedocs.io/en/latest/handbook/tutorial.html#merging-images. Let us know if you think there is any useful information still missing.

PR #6065, opened 2022-02-17
macOS: 'brew install openjpeg' for JPEG2000 support

hugovk:
Shall we include JPEG2000 as well?

---

Whilst we're here, a line or two down we tell the user to pip install Pillow from PyPI, or instead build from source:

```rst
Now install Pillow with::

    python3 -m pip install --upgrade Pillow

or from within the uncompressed source directory::

    python3 -m pip install .
```

But this is the section on [Building from source](https://pillow.readthedocs.io/en/stable/installation.html#building-on-macos), we already had the [Basic installation](https://pillow.readthedocs.io/en/stable/installation.html#basic-installation) near the top of the page.

Shall we change it to this?

```rst
Now install from within the uncompressed source directory::::

    python3 -m pip install .
```

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -291,7 +291,7 @@ tools.
 The easiest way to install external libraries is via `Homebrew
 <https://brew.sh/>`_. After you install Homebrew, run::
 
-    brew install libtiff libjpeg webp little-cms2
+    brew install libjpeg libtiff little-cms2 openjpeg webp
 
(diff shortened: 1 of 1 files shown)


radarhere:
Good point about building from source.

But there are three other instances where we tell people to `pip install --upgrade Pillow` when "Building from source", and I wonder if it's an extra step to tell them to download the source code in each section. How about using `--no-binary :all:` instead? See PR #6066

radarhere (approved):

PR #6066, opened 2022-02-17
Use --no-binary when installing from source

radarhere:
When "Building From Source" as this section of the documentation is supposed to do, make sure that the user is not installing a wheel - use `--no-binary :all:`


hugovk:
Good idea!

PR #6069, opened 2022-02-18
Added PyEncoder

### LINKED ISSUE #3176: Saving BLP files

### LINKED ISSUE #4059: PyEncoder doesn't exist
### What did you do?

Attempted to implement a Python based encoder for an image plugin.

From the documentation of 'register_encoder', I expected ImageFile.PyEncoder to exist.

```python
def register_encoder(name, encoder):
    """
    Registers an image encoder.  This function should not be
    used in application code.

    :param name: The name of the encoder
    :param encoder: A callable(mode, args) that returns an
                    ImageFile.PyEncoder object

    .. versionadded:: 4.1.0
    """
    ENCODERS[name] = encoder
```

### What actually happened?

It does not exist in any namespace that I can find.  Is this an oversight or is there a new way to implement Python based encoders suitable for addition to register_encoder?

* OS: Windows 10
* Python: 3.7.3
* Pillow: 6.0.0

```python

class MyEncoder(ImageFile.PyEncoder):
    pass
```

radarhere:
Resolves #4059
Resolves #3176

#1938 added `Image.register_encoder()`, but stopped short of creating `ImageFile.PyEncoder` to go with it.

First, however, some cleanup - tests from #4885 and #3190 were blindly appended to the end of Tests/test_imagefile.py. This incorrectly made them a part of the unrelated `TestPyDecoder` class. So I have moved them up to the `TestImageFile` class.

I've implemented a PyEncoder for BLP.

### DIFF
--- a/src/PIL/ImageFile.py
+++ b/src/PIL/ImageFile.py
@@ -49,7 +49,11 @@ ERRORS = {
     -8: "bad configuration",
     -9: "out of memory error",
 }
-"""Dict of known error codes returned from :meth:`.PyDecoder.decode`."""
+"""
+Dict of known error codes returned from :meth:`.PyDecoder.decode`,
+:meth:`.PyEncoder.encode` :meth:`.PyEncoder.encode_to_pyfd` and
+:meth:`.PyEncoder.encode_to_file`.
+"""
 
 
 #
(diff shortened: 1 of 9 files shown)


src/PIL/ImageFile.py
```diff
@@ -670,6 +650,32 @@ def setimage(self, im, extents=None):
         ):
             raise ValueError("Tile cannot extend outside image")
 
+
+class PyDecoder(PyCodec):
```
hugovk, line 654: Do these non-overridden methods needs testing?

radarhere: Ok, there should now be specific tests, and I've also implemented a PyEncoder for BLP to test it in the general sense.

PR #6070, opened 2022-02-18
(Discussion) Use setuptools / importlib entry points to load installed decoders

akaIDIOT:
*Note that this pull request is currently a proof of concept to aid in the discussion on whether this type of feature would suit Pillow.*

* Define a `PIL.Image.decoder` entry point group;
* Load / import decoders defined in this way on init.

This proof of concept was tested locally in Python 3.10 with the following addition to @K0lb3's [implementation of an ASTC-decoder for Pillow](https://github.com/K0lb3/astc_decomp/):

```ini
[options.entry_points]
PIL.Image.decoder =
     astc = astc_decomp:ASTCDecoder
```

The addition is purely packaging metadata, that can be discovered by `importlib.metadata.entry_points` as shown in the diff. Older versions of Python could be supported through the `importlib_metadata` backport, not currently implemented in the PR. Note that this definition contains the exact same information as it would normally register to `DECODERS`: a name and a type reference.

This PoC primarily poses the question **"does using packaging metadata entry points suit Pillow?"**. This was driven by my own surprise at having to import a module like `astc_decomp` without using that module for anything to allow it to register itself into `PIL.Image.DECODERS`. Considering that many Python projects use some kind of plugin system and setuptools / importlib expose a stdlib system for this, why could Pillow not do something similar?

There's more than just the primary question here that might need some discussion and / or fiddling to smoothly support something like this:

- Should it work in Python < 3.10 (throught the importlib backport)?
- Is importing a declared module (expecting that module to register itself, as they'd be doing right now) enough, or should the registration be explicit as per the entry point declaration (see diff)?
- At what point in time should this listing be done?
    - At the module level in `Image.py` won't work, this would halt execution of the module while 'plugins' would want to import things from that module;
    - Is this something for `preinit()` or for `init()`?
    - `init()` is not always called early enough it seems, my code would call `Image.frombytes(..., 'astc', ...)`, at which point `init()` was not (being) called yet, causing it to complain that the decoder was not available (or am I doing something wrong?);
    - An `EntryPoint` need not be loaded at `init()`, it could be 'wrapped' with a lazy loader that registers it, but only actually loads it when it's first used;
- Is `PIL.Image.decoder` the right entry point group (too specific maybe)?

I'm willing to put in work to implement more of this idea and add tests and documentation if needed, but I'm currently mostly interested in other people's thoughts on this; is this a good or a bad idea? :)

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -44,6 +44,12 @@ try:
 except ImportError:
     ElementTree = None
 
+try:
+    from importlib.metadata import entry_points
+except ImportError:
+    # NB: potentially try to load entry_points from backported importlib_metadata?
+    entry_points = None
+
 # VERSION was removed in Pillow 6.0.0.
 # PILLOW_VERSION was removed in Pillow 9.0.0.
 # Use __version__ instead.
@@ -373,6 +379,12 @@ def init():
             __import__(f"PIL.{plugin}", globals(), locals(), [])
         except ImportError as e:
             logger.debug("Image: failed to import %s: %s", plugin, e)
+
+    if entry_points:
+        for decoder in entry_points(group='PIL.Image.decoder'):
+            # or, alternatively, register a decoder directly from here:
+            # DECODERS[decoder.name] = decoder.load()
+            decoder.load()
 
     if OPEN or SAVE:
         _initialized = 2
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -403,6 +403,9 @@ def _getdecoder(mode, decoder_name, args, extra=()):
     elif not isinstance(args, tuple):
         args = (args,)
 
+    if _initialized < 2:
+        init()
+
     try:
         decoder = DECODERS[decoder_name]
     except KeyError:
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -381,7 +381,7 @@ def init():
             logger.debug("Image: failed to import %s: %s", plugin, e)
 
     if entry_points:
-        for decoder in entry_points(group='PIL.Image.decoder'):
+        for decoder in entry_points(group="PIL.Image.decoder"):
             # or, alternatively, register a decoder directly from here:
             # DECODERS[decoder.name] = decoder.load()
             decoder.load()


akaIDIOT:
I think that's a way of putting it, yeah. For things I write, I wouldn't be installing a package like that if there wasn't a reason to use it, and the import of that module can easily get optimized away by an IDE (at least PyCharm does that, I know flake8 is tough on unused imports), so converting the intention makes sense to me.

akaIDIOT:
Judging from the silence here, I'm guessing there's no real interest in using or integrating something like this in Pillow. 

For what it's worth, I'm not sold on the wording "intention to register" from an end user's perspective. As I came across an image encoded in a particular way, I had an intention to decode or render; how Pillow handles that internally is none of my concern at that point. The decoder project stated that the module needed to be imported, so that's clear enough, but it left me with a script that relied on import side effects, importing a module that it wasn't using (confusing IDE's to boot).

I get that changing something like this isn't to be taken lightly by a project with the install base of the likes of Pillow, but to me at least, "the spooky action at a distance" is the lesser of two evils compared to import side-effects that enable use of that module from code that doesn't (need to) reference that module. The two *could* co-exist, but that could be confusing to both developers and users of course…

nulano:
> The decoder project stated that the module needed to be imported, so that's clear enough, but it left me with a script that relied on import side effects, importing a module that it wasn't using (confusing IDE's to boot).

Just as an example of how plugins can avoid a seemingly unused import, pillow_heif defines a function `register_heif_opener` to register it as a Pillow plugin:
https://github.com/bigcat88/pillow_heif/blob/eb23ef58bad682a285c07ac95cc571b54d8c5e17/pillow_heif/as_opener.py#L92-L101

I don't have an opinion on whether that is better than an import side-effect, just showing that it is a possible alternative.

K0lb3:
As the author of the mentioned module,
I have to say that while the solution presented here is sure cool and would be nice to use,
I still disagree with it for the following two reasons:
First, as already mentioned, it's not really possible to track where the module is imported from,
so e.g. malware might be able to abuse this entry point to replace pillow's own decoders.
Name conflicts would also be hard annoying to solve.
Second, it's not possible to know which module is used as decoder just by looking at the code.
While good coders will likely make a comment somewhere, that module xyz is required for the decoder, this isn't the case in a lot of cases.

That said,
back when I created the decoder I would have wished for something like this system.

akaIDIOT:
Thanks for weighing in everyone, don't think this idea was meant to be :)

PR #6072, opened 2022-02-19
Replace requirements.txt with extras

hugovk:
Using `extras_require` means we can install just those requirements needed for a given task at the same time as installing Pillow, rather using a separate command that installs the whole bunch.

For example, to install things needed for testing:

```sh
pip install .[tests]

# Or my Mac terminal needs quotes
pip install ".[tests]"
```

Or for building docs:

```sh
pip install .[docs]

# Or my Mac terminal needs quotes
pip install ".[docs]"
```

---

This also removes two non-existent commands from the make help, and adds some capitalisation.

---

Whilst editing `Makefile` I noticed `install-venv` is deprecated but not in the deprecations docs:

```makefile
	@echo "  install-venv       (deprecated) install in virtualenv"

...

.PHONY: install-venv
install-venv:
	echo "'install-venv' is deprecated and will be removed in a future Pillow release"
	virtualenv .
```

Deprecated in https://github.com/python-pillow/Pillow/pull/5171 / https://github.com/python-pillow/Pillow/pull/5159 / Pillow 8.1.0 / 2 Jan 2021, so strictly speaking would have been eligible for removal in 9.0.0 / Jan 2022.

But this is arguably a dev helper rather than a formal API. Shall we wait for Pillow 10 (and add it to the deprecations docs?), or is it safe to remove it now?


requirements.txt
```diff
@@ -1,18 +0,0 @@
-# Development, documentation & testing requirements.
-black
```
radarhere, line 2: Was it a deliberate decision to remove black?

hugovk: Yes, it's since been added to pre-commit and is defined and run via that.

One exception: `make lint-fix` runs Black and isort without pre-commit. We could add the `--help > /dev/null 2>&1 || ... pip install` thing? Although isort was never in `requirements.txt`

radarhere: See what you think of https://github.com/hugovk/Pillow/pull/80

setup.cfg
```diff
@@ -35,6 +35,27 @@ project_urls =
 [options]
 python_requires = >=3.7
 
+[options.extras_require]
+docs =
+    olefile
```
radarhere, line 40: I feel like I should be able to work this out, but I can't - why is olefile needed for docs?

hugovk: Without it the autodoc fails for `FpxImagePlugin` and `MicImagePlugin`:

```
WARNING: autodoc: failed to import module 'FpxImagePlugin' from module 'PIL'; the following exception was raised:
No module named 'olefile'
looking for now-outdated files... WARNING: autodoc: failed to import module 'MicImagePlugin' from module 'PIL'; the following exception was raised:
No module named 'olefile'
...
/home/docs/checkouts/readthedocs.org/user_builds/pillow-hugovk/checkouts/extras_require/docs/reference/plugins.rst:60: WARNING: py:mod reference target not found: PIL.FpxImagePlugin
/home/docs/checkouts/readthedocs.org/user_builds/pillow-hugovk/checkouts/extras_require/docs/reference/plugins.rst:164: WARNING: py:mod reference target not found: PIL.MicImagePlugin
```

https://readthedocs.org/api/v2/build/16137387.txt
https://readthedocs.org/projects/pillow-hugovk/builds/16137387/

And iirc we just get the header and not this:

image

radarhere:
> Whilst editing `Makefile` I noticed `install-venv` is deprecated but not in the deprecations docs:
> 
> ```makefile
> 	@echo "  install-venv       (deprecated) install in virtualenv"
> 
> ...
> 
> .PHONY: install-venv
> install-venv:
> 	echo "'install-venv' is deprecated and will be removed in a future Pillow release"
> 	virtualenv .
> ```
> 
> Deprecated in #5171 / #5159 / Pillow 8.1.0 / 2 Jan 2021, so strictly speaking would have been eligible for removal in 9.0.0 / Jan 2022.
> 
> But this is arguably a dev helper rather than a formal API. Shall we wait for Pillow 10 (and add it to the deprecations docs?), or is it safe to remove it now?

You've [suggested before](https://github.com/python-pillow/Pillow/pull/5171#issuecomment-753441352) that the Makefile is a lower priority, so I don't mind. See https://github.com/hugovk/Pillow/pull/81

I think the bigger change is removing `install-req` without deprecation - but I can imagine an argument that changes to how you compile Pillow are easier to adapt to than changes in the core API, and that if you're compiling Pillow yourself, there is a measure of acceptance that this might not be a completely stable experience.

Shall we make a note of these changes in the release notes at least?

setup.cfg
radarhere: Thanks

hugovk:
Yep, added to release notes in the last commit.

docs/releasenotes/9.1.0.rst
```diff
@@ -18,11 +18,34 @@ Rather than returning a ``SystemError``, passing the incorrect types of coordina
...
+
+    # Install with dependencies for tests:
+    python3 -m pip install .[tests]
+
+    # Or for building docs:
+    python3 -m pip install .[docs]
+
+    # Or for all:
+    python3 -m pip install .[docs,tests]
+
```
radarhere, line 40:
```suggestion

On macOS, the last argument must be wrapped in quotes, e.g. `python3 -m pip install ".[tests]"`

```

My suggestion to document [this.](https://github.com/python-pillow/Pillow/pull/6072#issue-1144747636) But feel free to reword and improve it.
> \# Or my Mac terminal needs quotes
> pip install ".[docs]"

radarhere (approved):

docs/releasenotes/9.1.0.rst
hugovk: Yes, let's include something. I left it out as I'm not sure if it's a "must be" or "may need to be" depending on the terminal, are they needed on your Mac too?

(Also needs double backticks for RST.)

radarhere: They are needed on my Mac.
I've updated the suggestion to say "may", and with double backticks.

hugovk: Thanks!

PR #6077, opened 2022-02-21
GIF seek performance improvements

### LINKED ISSUE #6105: GIF's n_frames is (about 60x) slower than it should be
While idling around on SO and answering a question I noticed that `n_frames` is surprisingly slow for GIF:

```python
# setup code based on ImageIO 
# because it's short and we can avoid measuring disk IO
import imageio.v3 as iio
from PIL import Image
import io

img = iio.imread("imageio:newtonscradle.gif", index=None)
img_bytes = iio.imwrite("<bytes>", img, format="GIF")
```

The desired way to do things:

```python
%%timeit

with Image.open(io.BytesIO(img_bytes)) as file:
    n_frames = file.n_frames
```

Timing for this approach:

```
13.2 ms ± 94.3 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

This is much more time than it should take to check a series of block headers. So I checked, and - indeed - `n_frames` parses pixel data, which (in pillow9) involves a conversion to RGB(A) which is what makes this slow. Here is an alternative approach that only reads headers and is - more or less - a drop-in replacement:

```python
%%timeit
# based on SO: https://stackoverflow.com/a/7506880/6753182

with Image.open(io.BytesIO(img_bytes)) as file:
    def skip_color_table(flags):
        if flags & 0x80: 
            file.fp.seek(3 << ((flags & 7) + 1), 1)

    current_fp = file.fp.tell()
    total_frames = file.tell()  # start counting from the current frame

    # seek to beginning of next block
    buffer_start = file.tile[0][2]
    file.fp.seek(buffer_start)
    while True:
        size = file.fp.read(1)
        if size and size[0]:
            file.fp.seek(size[0], 1)
        else:
            break
    
    # count the remaining blocks
    while True:
        block = file.fp.read(1)
        if block == b';': 
            break
        if block == b'!': 
            file.fp.seek(1, 1)
        elif block == b',':
            total_frames += 1
            file.fp.seek(8, 1)
            skip_color_table(ord(file.fp.read(1)))
            file.fp.seek(1, 1)
        else: raise RuntimeError("unknown block type")
        
        # skip to next block instead of loading pixels
        while True:
            l = ord(file.fp.read(1))
            if not l: break
            file.fp.seek(l, 1)

    file.fp.seek(current_fp)
```

The timings:

```
211 µs ± 529 ns per loop (mean ± std. dev. of 7 runs, 1000 loops each) 
```

So if we only check block headers we can be about 60x faster than we are now. If desired, and there is somebody willing to review, I can submit a PR that adds this some time next week :)

(tests were done on Windows 11 with a AMD Ryzen 7 5800X 8-Core Processor and 2x Kingston KHX3200C16D4/32GX 32GB)

radarhere:
Resolves #6105

1. When checking whether or not a GIF is "animated", meaning that it has more than 1 frame, Pillow gets the current frame, and then seeks to frame 1. If it works, the image is animated, and if it doesn't, it is not.

https://github.com/python-pillow/Pillow/blob/54be93c8b0c1a0d7e1337614bde75a35a9a7a6ae/src/PIL/GifImagePlugin.py#L106-L121

Except it might be the case that the user has already seeked forward. `current` might be greater than 1. If so, then Pillow doesn't need to seek anywhere - we already know the image is animated.

This PR uses that knowledge as a minor performance improvement.

2. At the moment, `_seek` updates the frame position, applies disposal to the previous frame, and then checks to see if there's any more data in the file.

This PR rearranges that order, so if there is no more data, the image can stay as it is.

3. `is_animated` and `n_frames` seek forward through the GIF image to determine information about how many frames are present. During those seeks, the actual images are irrelevant. So don't perform image operations like loading or making sure that transparency and disposal are applied


AnonymouX47:
Just tested it out... superb!

AnonymouX47:
`n_frames` was almost immediate for a 2188-frame 1354x768 GIF... Makes a whole world of difference.
Thanks

AnonymouX47:
Is there any possibility this improvements will be included in release 9.1.0?

radarhere:
I would consider it likely, yes.

AnonymouX47:
Ok. Thanks.

AnonymouX47:
Well done! Sorry for bringing this up (again) but I was thinking (while working on my dependent project :smiley:)...

With these changes, I noticed that when `n_frames` is invoked, it still seeks back to frame 0 before seeking forward to the original *current* frame.

Since the image data is not updated (though I noticed `self.tile = []` isn't within the conditional block), shouldn't it be possible and **safe** to simply reset `self.__offset` to that of the original *current* frame?

hugovk:
Is the performance improvement quantifiable?

radarhere:
Running tests from #6075 with iss634.gif,

1. If GIF has already seeked past first frame, it is animated - https://github.com/python-pillow/Pillow/pull/6077/commits/73cf28cc9486cf3deda876bb24cdf1c701af0036
```python
from PIL import Image
import timeit

def test_is_animated_2():
    img = Image.open("iss634.gif")
    img.seek(41)
    img.is_animated

print(timeit.timeit(stmt=test_is_animated_2, number=25))
```
Before This Commit: 0.991317542
After This Commit:  0.5407172090000001

2. If next byte ends the GIF, stay on the current frame - https://github.com/python-pillow/Pillow/pull/6077/commits/590c616cfdd47e9771281ca75fa57733f4af8b14
```python
from PIL import Image
import timeit

def test_seek_2():
    with Image.open("iss634.gif") as im:
        im.seek(41)
        try:
            im.seek(42)
        except EOFError:
            pass

print(timeit.timeit(stmt=test_seek_2, number=25))
```
Before This Commit: 0.9614665830000001
After This Commit:  0.514499292

3. Do not update images during n_frames or is_animated seeking - https://github.com/python-pillow/Pillow/pull/6077/commits/f8546768d6da71d9a88fbadd6df6c8448270966b
```python
from PIL import Image
import timeit

def test_is_animated_n_frames():
    img = Image.open("iss634.gif")
    img.is_animated
    img.n_frames

print(timeit.timeit(stmt=test_is_animated_n_frames, number=25))
```
Before This Commit: 0.527664791
After This Commit:  0.134304667

AnonymouX47:
For what it's worth, [here](https://mega.nz/file/x4p1ma4B#PpYiUtwWFl5kUKOkyGavV9lXLuD4qsHB9uLRs1rfZvM) is the large GIF i mentioned earlier.

FirefoxMetzger (approved):
I looked it over and I don't see any other quick and significant wins. There is one small thing where instead of reading the data of the current block and discarding it, we could seek instead, but this only affects slow file systems, so not sure if it's worth the added complexity:

```diff
--- a/src/PIL/GifImagePlugin.py
+++ b/src/PIL/GifImagePlugin.py
@@ -55,11 +55,15 @@ class GifImageFile(ImageFile.ImageFile):

     global_palette = None

-    def data(self):
+    def data(self, read_data=True):
         s = self.fp.read(1)
-        if s and s[0]:
+        if not read_data and s and s[0]:
+            self.fp.seek(s[0], 1)
+            return b""
+        elif s and s[0]:
             return self.fp.read(s[0])
-        return None
+        else:
+            return None

     def _open(self):

@@ -100,7 +104,7 @@ class GifImageFile(ImageFile.ImageFile):
                     self._seek(self.tell() + 1, False)
             except EOFError:
                 self._n_frames = self.tell() + 1
-            self.seek(current)
+            self._seek(current, False)
         return self._n_frames

     @property
@@ -159,7 +163,7 @@ class GifImageFile(ImageFile.ImageFile):
         if self.__offset:
             # backup to last frame
             self.fp.seek(self.__offset)
-            while self.data():
+            while self.data(read_data=False) is not None:
                 pass
             self.__offset = 0

@@ -243,7 +247,7 @@ class GifImageFile(ImageFile.ImageFile):
                         block = self.data()
                         if len(block) >= 3 and block[0] == 1:
                             info["loop"] = i16(block, 1)
-                while self.data():
+                while self.data(read_data=False) is not None:
                     pass

             elif s == b",":
```

Otherwise the PR looks good to me 👍 

---

Note that merging this will close: #6105

radarhere:
Thanks for the approval, but yeah - testing your diff with runs of 1000 loops, I don't see any consistent improvements, so I'm not personally inclined. You do mention slow filesystems though, and I wouldn't describe my machine like that.

AnonymouX47:
Just curious, is there anything hindering these changes from being merged? :thinking:

AnonymouX47:
> Just waiting for another Pillow core developer to give their seal of approval.

Okay, I guessed as much.

> If you need these changes sooner, you can always compile https://github.com/radarhere/Pillow/tree/gif from source yourself.

Done that (for myself) since but the issue is... I'm concerned about the users of my project as Pillow is a core dependency and the issue concerned here affects performance and therefore user experience.

For some insight, https://github.com/AnonymouX47/term-img/issues/10 was the inception of these changes.

Thanks.

hugovk (approved):
> There are no problems that I'm aware of, no. Just waiting for another Pillow core developer to give their seal of approval. Judging from the past habits, I expect this, or a version of this, to be merged by the next release on April 1.

🦭 of ✅!

Yep, this will be in 9.1.0 on April 1.

Issue #6085, opened 2022-02-22
Inconsistent GIF result after resizing


Kusefiru:
I am trying to resize a `.gif` file. For that, I open my file with Pillow, then for each frame, resize that frame, and save the resulting frames as a new .gif file.

Here is the sample code:

```python
from PIL import Image

image = Image.open("test.gif")

frames = []
for i in range(image.n_frames):
    image.seek(i)
    frames.append(image.resize((512, 512), resample=Image.BICUBIC))

frames[0].save("result.gif", save_all=True, append_images=frames[1:], duration=100, loop=0)
```

Here is the used `test.gif` file:
![test.gif](https://user-images.githubusercontent.com/19759836/155106468-edcdfc5c-eb80-476c-b8fa-539f65b94386.gif)

As the resize is done using the `Image.BICUBIC` resampling method, I expect the resulting file to have a blurred appearance.

On **Pillow 9.0.1**, the first frame appears pixel crisp:
![result](https://user-images.githubusercontent.com/19759836/155107345-f69a5e7e-efb0-40e9-b2a4-98d9bb1d2367.gif)

On my previous install, **Pillow 8.4.0**, all the frames are pixel crisp:
![result_8 4 0](https://user-images.githubusercontent.com/19759836/155107328-69dd7907-a29e-4f02-8fb9-03f706aa0859.gif)

What could explain such behaviour (is there a missing option in my `save` call)?

FirefoxMetzger:
This is actually unrelated to doing a resize or the save call. Instead, it happens during the loading of the GIF. The lack of crisp-ness, aka, the graininess of each frame is a typical dithering artifact, in this case [Floyd–Steinberg dithering](https://en.wikipedia.org/wiki/Floyd%E2%80%93Steinberg_dithering). It currently happens under the hood once you seek beyond the first frame of a GIF:

https://github.com/python-pillow/Pillow/blob/4638c4fba651025851bf1f41a28583ae11d73c3a/src/PIL/GifImagePlugin.py#L169-L177

I don't think there is much you can do in pillow 9.0.1 considering that the value was hard-coded. It might be possible to expose the dithering method as a kwarg instead of hardcoding it (which could then be set to `None`), but I don't know the internal plugin API well enough to know where such a kwarg would live. `seek` seems natural, but might be the wrong place to do so. We'd have to wait for @radarhere to tell what's feasible.

radarhere:
[...]

Issue #5929, opened 2022-01-04
Reading GIF frame-by-frame changes mode during read in Pillow 9


FirefoxMetzger:
I encountered some new behavior in pillow v9 that - to me - is a bit counter-intuitive (image attached):

```python
from PIL import Image
import numpy as np

img = Image.open("newtonscradle.gif")

img.seek(0)
foo = np.array(img.getdata())
assert foo.shape == (30000, )
assert img.mode == "P"

img.seek(1)
bar = np.array(img.getdata())

assert bar.shape == (30000, 3)
assert img.mode == "RGB"
```

Technically, this is not a bug, since it seems to be intended behavior according to [the v9 changelog](https://pillow.readthedocs.io/en/stable/releasenotes/9.0.0.html#convert-subsequent-gif-frames-to-rgb-or-rgba). At the same time, I find it a bit counter-intuitive that the first frame of a GIF now has mode `P` while each subsequent frame has mode `RGB`/`RGBA`. 

I was wondering if it would be possible to make this more consistent and also convert the first frame (`img.seek(0)`) of any GIF to RGB/RGBA, so that the mode of all frames is consistent.

Supplementary:

![newtonscradle.gif](https://user-images.githubusercontent.com/4402489/148028883-224cc59c-ba9e-4a48-a45a-cd3f440cbd6a.gif)

radarhere:
I presume there is a contingent of Pillow users who just read the first frame of images, and don't bother seeking past that. #5857 had no effect on those users.

For Pillow users who do read animated GIFs, I expect them to be ok with this change, in the face of being able to read many GIF images correctly for the first time.

This is just a judgement call, but for my part, I'm a bit reluctant to change the experience of the first group of users for something that isn't an error. P mode more accurately reflects what is in that first image.

FirefoxMetzger:
> For Pillow users who do read animated GIFs, I expect them to be ok with this change, in the face of being able to read many GIF images correctly for the first time.

I, too, like the idea behind this change, because it means better support for GIFs that use a different palette for each image and try to be smart on what areas they do/don't dispose of.

[...]

PR #6086, opened 2022-02-22
Fixed loading L mode GIF with transparency

### LINKED ISSUE #6084: Pillow no longer reads grayscale multi-frame gifs
The following breaks in pillow 9.0.1 but works fine if I downgrade to pillow 8.4.0. My guess is that the new loader that converts to RGB/RGBA when seeking doesn't account for the possibility of a grayscale GIF.

```python

from PIL import Image
import numpy as np

img = Image.open("gray_test.gif")
im1 = np.asarray(img)  # just showing that the first frame works
img.seek(2)  # break
```

Stack Trace:

```
  File "C:\Users\Sebastian\AppData\Local\Programs\Python\Python38\lib\site-packages\PIL\GifImagePlugin.py", line 133, in seek
    self._seek(f)
  File "C:\Users\Sebastian\AppData\Local\Programs\Python\Python38\lib\site-packages\PIL\GifImagePlugin.py", line 151, in _seek
    self.load()
  File "C:\Users\Sebastian\AppData\Local\Programs\Python\Python38\lib\site-packages\PIL\ImageFile.py", line 264, in load
    self.load_end()
  File "C:\Users\Sebastian\AppData\Local\Programs\Python\Python38\lib\site-packages\PIL\GifImagePlugin.py", line 371, in load_end
    self.im.putpalettealpha(self._frame_transparency, 0)
ValueError: image has no palette
```

Picture for testing:

gray_test

radarhere:
Resolves #6084

In `GifImageFile.load_end()`, if there is transparency, `putpalettealpha()` is called, without considering the possibility that the image might not have a palette to modify.
https://github.com/python-pillow/Pillow/blob/4638c4fba651025851bf1f41a28583ae11d73c3a/src/PIL/GifImagePlugin.py#L367-L371

A similar situation happens in `_seek()` as well.

https://github.com/python-pillow/Pillow/blob/4638c4fba651025851bf1f41a28583ae11d73c3a/src/PIL/GifImagePlugin.py#L169-L171

This PR fixes the problem by calling the C layer `convert_transparent()` instead if the image mode is L.

Test image was created using a modified Pillow.


src/PIL/GifImagePlugin.py
```diff
@@ -167,9 +167,14 @@ def _seek(self, frame):
         if self.__frame == 1:
             self.pyaccess = None
             if "transparency" in self.info:
+                if self.mode == "P":
+                    self.im.putpalettealpha(self.info["transparency"], 0)
+                    self.im = self.im.convert("RGBA", Image.Dither.FLOYDSTEINBERG)
+                else:
+                    self.im = self.im.convert_transparent(
+                        "RGBA", self.info["transparency"]
+                    )
```
FirefoxMetzger, line 176: This is probably me not understanding correctly, but does this mean pillow will now promote gray GIFs to RGBA?

radarhere: Once the user seeks past the first frame, it will become RGBA if there transparency, yes. If there is no transparency, it will become RGB.
If you would like to suggest that something different happen here, that is fair. Yes, we do also have a LA mode, L with alpha, that is not being used here. Maybe I need to sit down and rethink https://pillow.readthedocs.io/en/stable/releasenotes/9.0.0.html#convert-subsequent-gif-frames-to-rgb-or-rgba in light of this, #5929, #5974 and #6085.

FirefoxMetzger: > Once the user seeks past the first frame, it will become RGBA if there transparency, yes. If there is no transparency, it will become RGB.
If you would like to suggest that something different happen here, that is fair.

Hmm, I think gray GIFs (with no pallette) are a niche, so it is probably okay to promote everything to RGB/RGBA. 

It does, however, come as a bit of a surprise, so I think it would be good to have this documented somewhere. Before it was "we unpack palettes to account for the possibility of more than 256 colors per frame", whereas now it is "we promote everything to RGB/RGBA". If I wouldn't know about this change and would see my gray GIF being served as RGB I would likely waste hours chasing a bug in my code that doesn't exist.

> Maybe I need to sit down and rethink https://pillow.readthedocs.io/en/stable/releasenotes/9.0.0.html#convert-subsequent-gif-frames-to-rgb-or-rgba in light of this, https://github.com/python-pillow/Pillow/issues/5929, https://github.com/python-pillow/Pillow/pull/5974 and https://github.com/python-pillow/Pillow/issues/6085.

I still like the idea overall. GIF - in my mind - is more of a video format than an image format, so doing a similar style of reading (where frames build on top of subsequent ones) seems quite natural. At the same time, I'd wish for this to be done consistently. From what I can tell it is currently something like: the first frame is P, L, or LA(?) and subsequent frames are RGB/RGBA but might also be P in the future (what about L/LA?).

Could the conversion be turned into an optional kwarg? That way users that want/need it can switch to this behavior and can enjoy it consistently from the first frame, whereas other users can get sequences of P, L, or LA frames.

radarhere: Just keeping the changes here about L mode GIFs - I've pushed commits so that the image becomes LA instead.

radarhere:
When seeking past first frame in an L mode GIF, If there is transparency present, the image now becomes LA.

I've also added this functionality to `convert()` in general, taking note of transparency when converting from L to LA.

src/libImaging/Convert.c
```diff
@@ -1634,29 +1634,15 @@ ImagingConvertTransparent(Imaging imIn, const char *mode, int r, int g, int b) {
         return (Imaging)ImagingError_ModeError();
     }
 
-    if (!((strcmp(imIn->mode, "RGB") == 0 || strcmp(imIn->mode, "1") == 0 ||
-           strcmp(imIn->mode, "I") == 0 || strcmp(imIn->mode, "L") == 0) &&
-          strcmp(mode, "RGBA") == 0))
-#ifdef notdef
```
radarhere, line 1640: This seems redundant, so I've removed it.

src/PIL/GifImagePlugin.py
radarhere: I've created PR #6150 as a solution to this, adding a setting to choose how they would like GifImagePlugin to behave. See what you think.

PR #6092, opened 2022-02-26
Various _accept changes

radarhere:
- Do not read data until necessary in GbrImagePlugin
- Raise SyntaxError if data is not as expected in XbmImagePlugin
- Moved FLI flags check into _accept
  - Since the [prefix is 16 characters](https://github.com/python-pillow/Pillow/blob/14b9b597ffa228ac56b91dc2782199f432defa14/src/PIL/Image.py#L3064), [this check on the 15th and 16th characters](https://github.com/python-pillow/Pillow/blob/14b9b597ffa228ac56b91dc2782199f432defa14/src/PIL/FliImagePlugin.py#L49) can be run earlier
- Make code clearer by matching _accept condition.
  - The `dword` condition here has the same purpose as the `prefix[:4]` check, so have them match for clarity.
    https://github.com/python-pillow/Pillow/blob/14b9b597ffa228ac56b91dc2782199f432defa14/src/PIL/WmfImagePlugin.py#L115
    https://github.com/python-pillow/Pillow/blob/14b9b597ffa228ac56b91dc2782199f432defa14/src/PIL/WmfImagePlugin.py#L68-L71
- Use `_accept` in `_open` for DDS and FLEX
- Simplify code by using `_accept`, instead of writing out the `_accept` condition a second time, in ICNS and TIFF

### DIFF
--- a/src/PIL/FtexImagePlugin.py
+++ b/src/PIL/FtexImagePlugin.py
@@ -95,7 +95,7 @@ class FtexImageFile(ImageFile.ImageFile):
 
     def _open(self):
         if not _accept(self.fp.read(4)):
-            raise SyntaxError("not a FTEX file")
+            raise SyntaxError("not an FTEX file")
         struct.unpack("<i", self.fp.read(4))  # version
         self._size = struct.unpack("<2i", self.fp.read(8))
         mipmap_count, format_count = struct.unpack("<2i", self.fp.read(8))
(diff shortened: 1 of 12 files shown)


src/PIL/FtexImagePlugin.py
```diff
@@ -94,7 +94,8 @@ class FtexImageFile(ImageFile.ImageFile):
     format_description = "Texture File Format (IW2:EOC)"
 
     def _open(self):
-        struct.unpack("<I", self.fp.read(4))  # magic
+        if not _accept(self.fp.read(4)):
+            raise SyntaxError("not a FTEX file")
```
hugovk, line 98: Nit: be guided by pronunciation:
```suggestion
            raise SyntaxError("not an FTEX file")
```

hugovk (approved):

PR #6102, opened 2022-03-02
Added support for reading BMP images with RLE8 compression

### LINKED ISSUE #3425: RLE8 compression support for bitmap images
### What did you do?
Attempted to open an RLE8-compressed bitmap image.
The image to display correctly.
### What actually happened?
An IOError("Unsupported BMP compression (%d)" % file_info['compression']) was thrown.

We try to open the image in our main program:
```python
def load_image(self, path, canvas):
        try:
            image = Image.open(path)
```
This calls the BmpImageFile class from BmpImagePlugin.py, which attempts to display the file according to the compression types listed in the dictionary on line 63: 
```python
COMPRESSIONS = {
        'RAW': 0,
        'RLE8': 1,
        'RLE4': 2,
        'BITFIELDS': 3,
        'JPEG': 4,
        'PNG': 5
    }
    RAW, RLE8, RLE4, BITFIELDS, JPEG, PNG = 0, 1, 2, 3, 4, 5
```
After reading header data, the program continues; on [line 158](https://github.com/python-pillow/Pillow/blob/master/src/PIL/BmpImagePlugin.py#L158), an if statement checks:
```python
if file_info['compression'] == self.BITFIELDS:
```
which evaluates to false. So, the program then jumps to [line 190](https://github.com/python-pillow/Pillow/blob/master/src/PIL/BmpImagePlugin.py#L190), where it checks whether the file's compression type is RAW:
```python
elif file_info['compression'] == self.RAW:
```
 which also evaluates to false. This causes the program to jump the code block to [line 193](https://github.com/python-pillow/Pillow/blob/master/src/PIL/BmpImagePlugin.py#L193), where the exception is thrown, and causes our main program to fail.
```python
raise IOError("Unsupported BMP compression (%d)" %
```
The code that ultimately throws this exception is found at line 194, in the [BmpImagePlugin source.](https://github.com/python-pillow/Pillow/blob/master/src/PIL/BmpImagePlugin.py)

* OS: Windows 10
* Python: 3.6.3
* Pillow: 5.2.0

[testImages.zip](https://github.com/python-pillow/Pillow/files/2498557/testImages.zip)

radarhere:
Resolves #3425

See https://docs.microsoft.com/en-us/windows/win32/gdi/bitmap-compression for documentation.

While my initial commit fixes test1.bmp from the issue, it doesn't fix test3.bmp. Investigating, I found that there were rows with too much data present. So I've added another commit to trim that out.

### DIFF
--- a/Tests/test_file_bmp.py
+++ b/Tests/test_file_bmp.py
@@ -147,6 +147,25 @@ def test_rle8():
                 im.load()
 
 
+@pytest.mark.parametrize(
+    "file_name,length",
+    (
+        # EOF immediately after the header
+        ("Tests/images/hopper_rle8.bmp", 1078),
+        # EOF during delta
+        ("Tests/images/bmp/q/pal8rletrns.bmp", 3670),
+        # EOF when reading data in absolute mode
+        ("Tests/images/bmp/g/pal8rle.bmp", 1064),
+    ),
+)
+def test_rle8_eof(file_name, length):
+    with open(file_name, "rb") as fp:
+        data = fp.read(length)
+        with Image.open(io.BytesIO(data)) as im:
+            with pytest.raises(ValueError):
+                im.load()
+
+
 def test_offset():
     # This image has been hexedited
     # to exclude the palette size from the pixel data offset
--- a/src/PIL/BmpImagePlugin.py
+++ b/src/PIL/BmpImagePlugin.py
@@ -283,8 +283,11 @@ class BmpRleDecoder(ImageFile.PyDecoder):
         data = bytearray()
         x = 0
         while len(data) < self.state.xsize * self.state.ysize:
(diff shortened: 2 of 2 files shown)


Tests/test_bmp_reference.py
```diff
@@ -40,6 +40,7 @@ def test_questionable():
         "rgb32fakealpha.bmp",
         "rgb24largepal.bmp",
         "pal8os2sp.bmp",
+        "pal8rletrns.bmp",
```
radarhere, line 43: When running this locally, I was told to add this by the test.
https://github.com/python-pillow/Pillow/blob/38f4660d9ecfab74af8bad292d31e080072b25b3/Tests/test_bmp_reference.py#L50

PR #6107, opened 2022-03-02
Fixed reading 9 to 15-bit PGM

### LINKED ISSUE #5403: 10 bit pgm file not read correctly
### What did you do?
Load 10 bit pgm image and saved it as png image

Pixel values scaled to use the 16 bit range 
(http://www.libpng.org/pub/png/spec/1.2/PNG-Chunks.html)

### What actually happened?
pixel values stored in the least significant bytes and hence the image is too dark

* OS:  Ubuntu 18.04.5 LTS
* Python: 3.6
* Pillow: 8.2.0

```python
from PIL import Image

input_path_pgm = 'path-to-10-bit-pgm-image'
output_path_png = 'path-to-16-bit-png-image'
image = Image.open(input_path_pgm)
image.save(output_path_png, format='PNG')
```

radarhere:
Resolves #5403

The image from the issue is 10-bit PGM. As per http://netpbm.sourceforge.net/doc/pgm.html, this bit depth can be determined from the "maximum gray value (Maxval)" in the header (`s` or [`# maxgrey`](https://github.com/python-pillow/Pillow/blob/92c26a77ca53a2bfbd8804f009c6c8755d0e5a43/src/PIL/PpmImagePlugin.py#L104) in PpmImagePlugin) of 1023 (2 ** 10 - 1).

This PR adds an unpacker for that. The test image included was created with ImageMagick using `convert hopper.png -depth 10 10_bit_binary.pgm`.


radarhere:
As [requested](https://github.com/python-pillow/Pillow/issues/5403#issuecomment-1058028889), I've also now added 9, 12 and 14-bit.

wiredfool:
I feel a little bad for the prime numbered bits 11, 13, which have been sadly neglected, here and in the wild.

PR #6119, opened 2022-03-09
Added support for PPM arbitrary maxval

### LINKED ISSUE #5008: PGM and PPM maximum grey/color value is ignored

### LINKED ISSUE #5403: 10 bit pgm file not read correctly

radarhere:
Resolves #5008
Resolves #5403

#5008 points out that the PPM [`maxval`](https://github.com/python-pillow/Pillow/blob/397a9409959bccd0130d0b51820f2a94e3bf153d/src/PIL/PpmImagePlugin.py#L114-L123) does not just indicate the bit depth of an image. It is an arbitrary number, that ["Must be less than 65536 and more than zero."](http://netpbm.sourceforge.net/doc/ppm.html)

The example from the issue has a `maxval` of 11 (not 2 ** 11, but just 11), and https://en.wikipedia.org/wiki/Netpbm#PGM_example shows an example where the `maxval` is 15. So when it is 15, the values range between 0 and 15, from black to white.

This PR 
- allows this value to be arbitrary for P5 (grayscale). That fixes the code example in #5008, and #5403, which features a P5 image at 10-bit depth, or 1023 maxval (2 ** 10 - 1).
- allows this value to be arbitrary for P6 (RGB). When the value is more than 255, that would require more than 16-bits of RGB data. As we do not have such a mode, this PR simply scales down the pixels to fit into 16-bit.

### DIFF
--- a/Tests/test_file_ppm.py
+++ b/Tests/test_file_ppm.py
@@ -13,16 +13,64 @@ TEST_FILE = "Tests/images/hopper.ppm"
 
 def test_sanity():
     with Image.open(TEST_FILE) as im:
-        im.load()
         assert im.mode == "RGB"
         assert im.size == (128, 128)
-        assert im.format, "PPM"
+        assert im.format == "PPM"
         assert im.get_format_mimetype() == "image/x-portable-pixmap"
+
+
+def test_arbitrary_maxval():
+    # P5 L mode
+    fp = BytesIO(b"P5 3 1 4 \x00\x02\x04")
+    with Image.open(fp) as im:
+        assert im.size == (3, 1)
+        assert im.mode == "L"
+
+        px = im.load()
+        assert tuple(px[x, 0] for x in range(3)) == (0, 128, 255)
+
+    # P5 I mode
+    fp = BytesIO(b"P5 3 1 257 \x00\x00\x00\x80\x01\x01")
+    with Image.open(fp) as im:
+        assert im.size == (3, 1)
+        assert im.mode == "I"
+
+        px = im.load()
+        assert tuple(px[x, 0] for x in range(3)) == (0, 32640, 65535)
+
+    # P6 with maxval < 255
+    fp = BytesIO(b"P6 3 1 17 \x00\x01\x02\x08\x09\x0A\x0F\x10\x11")
+    with Image.open(fp) as im:
+        assert im.size == (3, 1)
+        assert im.mode == "RGB"
+
+        px = im.load()
+        assert tuple(px[x, 0] for x in range(3)) == (
+            (0, 15, 30),
+            (120, 135, 150),
+            (225, 240, 255),
+        )
+
+    # P6 with maxval > 255
+    # Scale down to 255, since there is no RGB mode with more than 8-bit
+    fp = BytesIO(
+        b"P6 3 1 257 \x00\x00\x00\x01\x00\x02"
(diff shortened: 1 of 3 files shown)


Tests/test_file_ppm.py
```diff
@@ -13,16 +13,64 @@
...
     with Image.open(TEST_FILE) as im:
-        im.load()
         assert im.mode == "RGB"
         assert im.size == (128, 128)
-        assert im.format, "PPM"
+        assert im.format == "PPM"
         assert im.get_format_mimetype() == "image/x-portable-pixmap"
 
 
+def test_arbitrary_maxval():
```
hugovk, line 22: Can these mode be parametrised? Looks like one input bytestring and two things to assert.

radarhere: Ok, done

hugovk: Thanks!

PR #6138, opened 2022-03-16
Search pkgconf system libs/cflags

jameshilliard:
We need to search the system paths as well from pkg-config for some packages to be found properly.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -252,8 +252,8 @@ def _cmd_exists(cmd):
 def _pkg_config(name):
     try:
         command = os.environ.get("PKG_CONFIG", "pkg-config")
-        command_libs = [command, "--libs-only-L", name]
-        command_cflags = [command, "--cflags-only-I", name]
+        command_libs = [command, "--libs-only-L", "--keep-system-libs", name]
+        command_cflags = [command, "--cflags-only-I", "--keep-system-cflags", name]
         if not DEBUG:
             command_libs.append("--silence-errors")
             command_cflags.append("--silence-errors")
--- a/setup.py
+++ b/setup.py
@@ -250,28 +250,32 @@ def _cmd_exists(cmd):
 
 
 def _pkg_config(name):
-    try:
-        command = os.environ.get("PKG_CONFIG", "pkg-config")
-        command_libs = [command, "--libs-only-L", "--keep-system-libs", name]
-        command_cflags = [command, "--cflags-only-I", "--keep-system-cflags", name]
-        if not DEBUG:
-            command_libs.append("--silence-errors")
-            command_cflags.append("--silence-errors")
-        libs = (
-            subprocess.check_output(command_libs)
-            .decode("utf8")
-            .strip()
-            .replace("-L", "")
-        )
-        cflags = (
-            subprocess.check_output(command_cflags)
-            .decode("utf8")
-            .strip()
-            .replace("-I", "")
-        )
-        return libs, cflags
-    except Exception:
-        pass
+    command = os.environ.get("PKG_CONFIG", "pkg-config")
+    for keep_system in (True, False):
+        try:
+            command_libs = [command, "--libs-only-L", name]
+            command_cflags = [command, "--cflags-only-I", name]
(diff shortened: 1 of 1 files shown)


radarhere:
> some packages

Could you elaborate on this a bit? It's potentially helpful to have this information to refer back to.

jameshilliard:
> Could you elaborate on this a bit?

Zlib and anything that has headers in the normal system libs/include folders seems to not get their headers picked up without this change.

I think this issue is mostly going to be something people hit when cross compiling since the prefix based system libs/include folder would probably work in most of the usual cases(searching in `sys.prefix` will not work when cross compiling since it points to the host toolchain prefix rather than the target toolchain sysroot prefix):
https://github.com/python-pillow/Pillow/blob/0cf072db39bf1868575c625fc4403f388e5d47d4/setup.py#L494-L495

When cross compiling we try and make sure pkg-config points to the correct target sysroot however.

jameshilliard:
@radarhere FYI there's some additional background/discussion around this issue [here](https://lore.kernel.org/buildroot/CA+_SqVYy6DXFsQ26VmAwhsjGjo232fqAH=WQbnMk9cJQwg=JzQ@mail.gmail.com/).

jameshilliard:
@radarhere Tests should be passing now, I've re-based on `main` with #6229.

radarhere:
Yes, the tests will pass now, but the failure revealed a problem - on main, [libtiff is detected](https://github.com/python-pillow/Pillow/runs/6089745828#step:8:37). On this branch, [it is not.](https://github.com/python-pillow/Pillow/runs/6103156939#step:8:37)

jameshilliard:
> On this branch, [it is not.](https://github.com/python-pillow/Pillow/runs/6103156939#step:8:37)

Weird, on ubuntu 20.04 with `libtiff-dev` installed it seems to get detected fine for me.

jameshilliard:
FYI there's a repo setting you can change so that you don't have to manually approve most workflows in pull requests:
Screen Shot 2022-04-22 at 12 17 36 PM

radarhere:
Ok, I [stopped `_pkg_config` from catching exceptions](https://github.com/radarhere/Pillow/commit/8b654dcef5559d09b235fc5d2c262b6704955103), and it turned out there was an error occurring - "[Unknown option --keep-system-libs](https://github.com/radarhere/Pillow/runs/6139685074?check_suite_focus=true#step:7:31)".

If I try on my macOS machine,
```
$ pkg-config --keep-system-libs
Unknown option --keep-system-libs
$ pkg-config --keep-system-cflags
Unknown option --keep-system-cflags
```

So I guess those options are only available in certain environments. What is surprising though is that you say the flag works for you on your copy of Ubuntu 20.04, but it doesn't in GitHub Actions. Any ideas why that might be?

Perhaps a solution would be to try `--keep-system-libs` first, and then to try again without it if it doesn't work? https://github.com/jameshilliard/Pillow/pull/1

jameshilliard:
> What is surprising though is that you say the flag works for you on your copy of Ubuntu 20.04, but it doesn't in GitHub Actions. Any ideas why that might be?

Yeah, not sure why that is happening.

> Perhaps a solution would be to try `--keep-system-libs` first, and then to try again without it if it doesn't work?

That seems to work for me, merged to my branch here.

radarhere:
The version number is was helpful, thanks - I think you are using pkgconf, not pkg-config.

https://pkgs.org/download/pkg-config
pkgconf

There is apparently a "[compatibility interface](https://fedoraproject.org/wiki/Changes/pkgconf_as_system_pkg-config_implementation#Detailed_Description)" that means you can run it as pkg-config, hence the confusion.

jameshilliard:
Ah, yeah, guess that explains why the fallback is needed for some systems.

radarhere (approved):

jameshilliard:
Rebased on [`python-pillow:main`](https://github.com/python-pillow/Pillow).

jameshilliard:
@radarhere This good to merge or anything else I should do?

PR #6141, opened 2022-03-18
Do not always require ctypes in tests

radarhere:
https://github.com/python-pillow/Pillow/blob/f3a3b20c51035969e585583e7340c80bf28caef3/Tests/test_image_access.py#L8

This import at the top level of test_image_access.py, but is only used on [non-CI Windows](https://github.com/python-pillow/Pillow/blob/f3a3b20c51035969e585583e7340c80bf28caef3/Tests/test_image_access.py#L402). This PR moves the import so that if a user doesn't have setuptools installed, that doesn't stop them from running our test suite.

test_image_access.py and test_imagewin_pointers.py similarly always import ctypes for tests that only run on Windows. I have moved those imports as well.

test_file_libtiff.py imports ctypes to calculate libtiff will transform a specific value into. I've just run that calculation now, recorded the value and removed the import.

### DIFF
--- a/Tests/test_image_access.py
+++ b/Tests/test_image_access.py
@@ -1,4 +1,3 @@
-import ctypes
 import os
 import subprocess
 import sys
@@ -404,6 +403,8 @@ class TestEmbeddable:
         "not from shell",
     )
     def test_embeddable(self):
+        import ctypes
+
         with open("embed_pil.c", "w") as fh:
             fh.write(
                 """
--- a/Tests/test_file_libtiff.py
+++ b/Tests/test_file_libtiff.py
@@ -4,7 +4,6 @@ import itertools
 import os
 import re
 from collections import namedtuple
-from ctypes import c_float
 
 import pytest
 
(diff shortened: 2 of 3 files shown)


radarhere:
Ok, I've taken out the setuptools change.

Tests/test_image_access.py
```diff
@@ -404,6 +403,8 @@ class TestEmbeddable:
         "not from shell",
     )
     def test_embeddable(self):
+        import ctypes
```
hugovk, line 406: Any reason to move this here (and in the other file)? Is it sometimes slow to import? 

PEP 8 says to put them at the top, but I think it's fine to move slow imports to where they're needed, if they're not always run.

radarhere: ctypes is an optional module, so it may not be supported by the user's Python installation. If they are on macOS or Linux, we shouldn't block them from running our test suite because a dependency for Windows tests are missing.

If that sounds obscure, my personal motivation for this is that when trying to add musllinux to pillow-wheels, [I found that ctypes was failing](https://github.com/radarhere/pillow-wheels/runs/5582403106#step:4:6554). After [my attempt to fix ctypes failed](https://github.com/multi-build/docker-images/pull/30), I thought I'd try and remove ctypes instead since it wasn't really needed for that build.

I also discovered I'm not the only one who has thought about optionally importing ctypes. [Most of the time, setuptools only imports ctypes where it is needed.](https://github.com/pypa/setuptools/pull/3178)

If you think this isn't a generally helpful thing though, we can close it and I can go back to trying to fix ctypes on musllinux.

hugovk: Ah, if it's optional, that's a good reason. And if it helps unblock musllinux, that's also a very good reason! Thanks!

radarhere: I was able to fix ctypes for musllinux after all, in https://github.com/multi-build/docker-images/pull/32

Issue #5944, opened 2022-01-08
Trying to save a (0, 0) image creates an unreadable file


NewUserHa:
### What did you do?
```python
from PIL import Image
Image.new("RGB",(0,0)).save('test.png') # `.png` or `.jpg` or `.tiff`
```

### What did you expect to happen?
no file should be created on disk because there's an `SystemError: tile cannot extend outside image` error.
(https://github.com/python-pillow/Pillow/issues/5931#issuecomment-1007142950 for jpg, libjpeg seems not support size 0 0)

### What actually happened?
throw `SystemError: tile cannot extend outside image`
a ~33 bytes file is on disk, which is unable to read via pillow and other viewer No matter the extension is `.png` or `.jpg` or `.tiff`

### What are your OS, Python and Pillow versions?
* OS: win10 1803
* Python: 3.8.10
* Pillow: 9.0.0

radarhere:
I've created PR #5947 to resolve this.

NewUserHa:
rather than removing file, why not create file only if there's no error?

radarhere:
Ok, i've updated #5947 to use that method, rather than my original suggestion of removing the saved image after an error.

radarhere:
I've restored my original idea of removing the file after the error as a new PR, #6134. I'm skeptical about the idea of forcing files to live in memory simply to handle this edge case.

radarhere:
Yes, if the user is asking Pillow to overwrite an existing file with the image, a corrupted image will still be saved.
But I think that this is a rare situation, and forcing all users to keep their images in memory to handle this is an extreme solution.

I don't understand your comment about `created = not os.path.exists(filename)`. It is before `builtins.open` because it is checking whether or not Pillow is creating the file. If Pillow is creating the file, then I feel ok about removing it. If the file was already on disk, then I think the user would not expect their file to be deleted after Pillow failed to save to it.

NewUserHa:
what about: "before modifying on disk, check if it is saveable" OR
"use temp file then move to the place, like that the browsers download files does"?

ok. I thought it was to check if the file is `created` by the pillow

radarhere:
It's unclear what "check if it is saveable" means without a significant change to how Pillow plugins work.

Using a temporary file is possible, but again, it seems like a sizeable change to how Pillow operates, and this issue seems like a rare situation. Feel free to argue otherwise if you think this is a common situation, and my PR doesn't address it.

NewUserHa:
[...]

Issue #5931, opened 2022-01-04
size 0,0 image use `.save()`/`.tobytes()` throw `SystemError: tile cannot extend outside image`


NewUserHa:
### What did you do?
Create a size (0,0) image instead of None, for convenient in coding,
```python
x.width
x.height
# other stuffs
x.tobytes()
```
to get rid of
```python
if x:
    x.width
    x.height
   # other stuffs
    x.tobytes()
```

### What did you expect to happen?
Normal, probably return `b''`

### What actually happened?
```python
---------------------------------------------------------------------------
SystemError                               Traceback (most recent call last)
<ipython-input-0-d240cdef7328> in <module>
----> 1 Image.new("RGB",(0,0)).tobytes()

c:\program files\python38\lib\site-packages\PIL\Image.py in tobytes(self, encoder_name, *args)
    720         # unpack data
    721         e = _getencoder(self.mode, encoder_name, args)
--> 722         e.setimage(self.im)
    723
    724         bufsize = max(65536, self.size[0] * 4)  # see RawEncode.c

SystemError: tile cannot extend outside image
```
### What are your OS, Python and Pillow versions?

* OS: Win10 1803
* Python: 3.8.10
* Pillow: 9.0.0

```python
from PIL import Image
Image.new("RGB",(0,0)).tobytes()
```

NewUserHa:
`.save` throw the same error as well.

radarhere:
Just to be clear, what format are trying to save the image as?

NewUserHa:
the `.save` is `.jpg`

radarhere:
When I try and save a (0, 0) JPEG, if I push past the Pillow error, I get "Empty JPEG image (DNL not supported)" which is coming from libjpeg itself - https://github.com/freedesktop/libjpeg/blob/b4f886fcbb692e20fb4da4e6e3ead57f79444734/jcmaster.c#L54-L57. So it doesn't look like it is possible to create a (0, 0) JPEG file.

I've created PR #5938 to fix the error coming from `tobytes()`.

NewUserHa:
I searched a few minutes, it seems like .jpg, .png don't support size 0x0 image. but no info for .tiff and other formats.

PR #6159, opened 2022-03-28
Raise ValueError when trying to save empty JPEG

radarhere:
Suggested in https://github.com/python-pillow/Pillow/issues/5944#issuecomment-1079934111

The JPEG format [does not support saving an image with zero width or height](https://github.com/python-pillow/Pillow/issues/5931#issuecomment-1007142950). This PR raises a ValueError for the situation, rather than the SystemError that is currently raised.

### DIFF
--- a/Tests/test_file_jpeg.py
+++ b/Tests/test_file_jpeg.py
@@ -68,6 +68,13 @@ class TestFileJpeg:
             assert im.format == "JPEG"
             assert im.get_format_mimetype() == "image/jpeg"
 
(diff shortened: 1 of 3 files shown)


hugovk (approved):
Looks good.

Optional suggestion: this could even be the very first check before the `OSError`, as it's arguably a more fundamental problem than changing the mode.

radarhere:
Ok, done.

hugovk:
Thanks!

PR #5062, opened 2020-11-25
Dynamically link FriBiDi instead of Raqm

nulano:
This is likely to fix issues #3066 and #4225. 

Includes Raqm and a FriBiDi shim in `src/thirdparty`. Raqm is modified to use the shim ~~and for compatibility with C89~~ (edit: reverted because Raqm is no longer built by default).

I've checked that Raqm/FriBiDi/HarfBuzz are being loaded correctly in all of the Windows, Ubuntu, and Docker GHA jobs.
~~I'm not sure what is missing to get HarfBuzz detected on macOS~~ edit: merging master fixed this somehow. ~~Travis is failing to compile Raqm because it has HarfBuzz 1.0.1 which is too old~~ edit: disabled dynamic loading by default.

Is this the way forward to resolving the Raqm/FreeType crashes? I haven't tried building wheels to see if the issues are actually resolved.

This effectively removes dynamic loading as the default option. It now needs to be explicitly enabled with `--vendor-raqm --vendor-fribidi`, e.g. when building wheels. A side effect of these changes is that Windows users now need `fribidi.dll` instead of `libraqm.dll`.

~~TODO:~~
* ~~I intended to add a detailed message for which Raqm version is selected in `setup.py`.~~
* ~~PR in pillow-wheels adding HarfBuzz and enabling dynamic linking.~~ https://github.com/python-pillow/pillow-wheels/pull/181

PR #6165, opened 2022-03-31
Fix loading FriBiDi on Alpine

nulano:
Testing https://github.com/python-pillow/pillow-wheels/pull/271, I find the tests pass, but Raqm is not enabled even when I install FriBiDi.

On Alpine, fribidi is at `/usr/lib/libfribidi.so.0`, and I can see the second `dlopen` call succeeds when using `strace`. However, the shim is still rejecting the SO. This is because the `dlerror` call returns the error for the first `dlopen` call.

This could be fixed by calling `dlerror` before retrying to clear the error, but since all `dlsym` return values are checked to be non-null, there is actually no need to check `dlerror` at all.

See PR adding tests in https://github.com/python-pillow/pillow-wheels/pull/274

### DIFF
--- a/src/thirdparty/fribidi-shim/fribidi.c
+++ b/src/thirdparty/fribidi-shim/fribidi.c
@@ -7,7 +7,6 @@
 #endif
 
 #define FRIBIDI_SHIM_IMPLEMENTATION
-
 #include "fribidi.h"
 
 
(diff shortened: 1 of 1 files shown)


hugovk (approved):
Should this be reported/contributed upstream?

nulano:
> Should this be reported/contributed upstream?

No, this my code written for Pillow (in #5062). I added `dlerror` because that was used in the previous implementation (dynamically loading Raqm in `_imagingfg.c`) which this was based on.

hugovk:
Thanks!

PR #6167, opened 2022-03-31
Add support for unpacking 16-bit BGRA

gmarkall:
Adds support for unpacking 16-bit BGRA format. This format is used by ImageMagick 6 for many platforms - the addition of this support enables converting directly from `Magick::PixelPacket*` into PIL images.

I have used this new functionality in an example image-processing library I wrote to accompany a talk on Python APIs for CUDA-accelerated applications (see e.g. https://github.com/gmarkall/numba-accelerated-udfs/blob/main/filigree/api.py#L47), but I think it is generally useful for other applications to be able to work directly with ImageMagick 6 data.

### DIFF
--- a/src/libImaging/Unpack.c
+++ b/src/libImaging/Unpack.c
@@ -1065,6 +1065,30 @@ unpackBGRA(UINT8 *_out, const UINT8 *in, int pixels) {
     }
 }
 
+static void
+unpackBGRA16L(UINT8 *_out, const UINT8 *in, int pixels) {
+    int i;
+    /* 16-bit RGBA, little-endian order, reversed words */
+    for (i = 0; i < pixels; i++) {
+        UINT32 iv = MAKE_UINT32(in[5], in[3], in[1], in[7]);
+        memcpy(_out, &iv, sizeof(iv));
+        in += 8;
+        _out += 4;
+    }
+}
+
+static void
+unpackBGRA16B(UINT8 *_out, const UINT8 *in, int pixels) {
+    int i;
+    /* 16-bit RGBA, big-endian order, reversed words */
+    for (i = 0; i < pixels; i++) {
+        UINT32 iv = MAKE_UINT32(in[4], in[2], in[0], in[6]);
+        memcpy(_out, &iv, sizeof(iv));
+        in += 8;
+        _out += 4;
+    }
+}
+
 /* Unpack to "CMYK" image */
 
 static void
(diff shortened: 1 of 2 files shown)


nulano:
Related to #1888.

It seems this PR scales 16-bit images to 8-bit by discarding the lower 8 bits. Note that this was considered controversial for I mode images in #3011.

gmarkall:
> It seems this PR scales 16-bit images to 8-bit by discarding the lower 8 bits.

Yes, it does - I just followed the same pattern used in `unpackRGBA16L` and `unpackRGBA16B` which seem to do the same: https://github.com/python-pillow/Pillow/blob/dae82d16ba23548c74aa2a5aa820baec19bb7a51/src/libImaging/Unpack.c#L1010-L1030

(in fact I now realise I have a copy/paste error in the comment from those functions - will fix shortly)

gmarkall:
> (in fact I now realise I have a copy/paste error in the comment from those functions - will fix shortly)

... actually I didn't - apologies!... I found the comments I added confusing, but they're consistent with the other comments in the file as far as I can tell.

src/libImaging/Unpack.c
```diff
@@ -1065,6 +1065,30 @@ unpackBGRA(UINT8 *_out, const UINT8 *in, int pixels) {
     }
 }
 
+static void
+unpackBGRA16L(UINT8 *_out, const UINT8 *in, int pixels) {
+    int i;
+    /* 16-bit RGBA, little-endian order, reversed words */
```
radarhere, line 1071:
```suggestion
    /* 16-bit RGBA, little-endian order, reversed bytes */
```
A suggestion, so that it matches https://github.com/python-pillow/Pillow/blob/5d070222d21138d2ead002fd33fdf5adcb708941/src/libImaging/Unpack.c#L1059

src/libImaging/Unpack.c
```diff
@@ -1065,6 +1065,30 @@ unpackBGRA(UINT8 *_out, const UINT8 *in, int pixels) {
...
+        memcpy(_out, &iv, sizeof(iv));
+        in += 8;
+        _out += 4;
+    }
+}
+
+static void
+unpackBGRA16B(UINT8 *_out, const UINT8 *in, int pixels) {
+    int i;
+    /* 16-bit RGBA, big-endian order, reversed words */
```
radarhere, line 1083:
```suggestion
    /* 16-bit RGBA, big-endian order, reversed bytes */
```

radarhere (approved):

src/libImaging/Unpack.c
gmarkall: However, aren't the words (16-bit values R, G, and B) reversed in this case, but not the bytes (8-bit values, which would be half of one of the channels I guess)?

radarhere: A is staying the same place, so I don't really think it's completely reversed at all. How about "rearranged channels"? See https://github.com/gmarkall/Pillow/pull/1

gmarkall: Merged, many thanks!

PR #6171, opened 2022-04-01
DOC: the constants were deprecated in 9.1.0 not 9.2.0

tacaswell:

 * Correct the deprecated note on  https://pillow.readthedocs.io/en/stable/deprecations.html#constants to match https://pillow.readthedocs.io/en/stable/releasenotes/9.1.0.html#constants


hugovk:
Thanks, good catch!

Note: this doesn't affect the planned removal date, Pillow 10.0.0 (2023-07-01).

---

We did have the pre-commit pins frozen, but I don't think we really need to do that: I've not seen any other projects use it and we already trust those projects, for example by taking new tags, frozen or not.

hugovk (approved):

PR #6176, opened 2022-04-02
Consider transparency from each frame when saving GIF

### LINKED ISSUE #6162: Background is white instead of being transparent
### What did you do?
Overlay add png images on a gif file frames
Get a gif with the pngs added on each layer and a transparent background
### What actually happened?
I'm getting the gif as I want it but with a white background

* OS: MacOs 11.6.5
* Python: python3.7
* Pillow: 9.1

All the png files have a transparent background and also the gif has a transparent background !

Here is an image from the end result gif (the file is too big i couldn't upload it here):
Screen Shot 2022-03-29 at 13 59 33

````python

background   = Image.open('images/1.png')#.convert('RGBA')
animated_gif = Image.open("images/2.gif")
layer1 = Image.open('images/3.png')
layer2 = Image.open('images/4.png')
all_frames = []


for gif_frame in ImageSequence.Iterator(animated_gif):


    new_frame = background.copy()
    layerr1 = layer1.copy()
    layerr2 = layer2.copy()

    gif_frame = gif_frame.convert('RGBA')
    
    new_frame.paste(layerr2, mask=layerr2)
    new_frame.paste(gif_frame, mask=gif_frame) 
    new_frame.paste(layerr1, mask=layerr1)
    all_frames.append(new_frame)


all_frames[0].save("image2.gif",transparency = 255, background = 255, save_all=True, append_images=all_frames[1:], duration=50, loop=0)

````

Please help me with this, this is the first time I'm using pillow !

### PR DESCRIPTION
**Pull request by radarhere**
Resolves #6162

When saving a GIF, `_normalize_mode` may set `info["transparency"]` for a frame.
https://github.com/python-pillow/Pillow/blob/a575ec33d2f0a04de6cde880a4163de61106b343/src/PIL/GifImagePlugin.py#L482

When the code comes back to `_write_multiple_frames`, the transparency is used, if the user hasn't overridden it from `encoderinfo`... but it is only copied from the first frame.
https://github.com/python-pillow/Pillow/blob/a575ec33d2f0a04de6cde880a4163de61106b343/src/PIL/GifImagePlugin.py#L573-L576

This PR allows each frame to have a different transparency from `_normalize_mode`.

### DIFF
--- a/Tests/test_file_gif.py
+++ b/Tests/test_file_gif.py
@@ -619,7 +619,7 @@ def test_dispose2_background(tmp_path):
         assert im.getpixel((0, 0)) == (255, 0, 0)
 
 
-def test_transparency_in_second_frame():
+def test_transparency_in_second_frame(tmp_path):
     with Image.open("Tests/images/different_transparency.gif") as im:
         assert im.info["transparency"] == 0
 
@@ -628,6 +628,15 @@ def test_transparency_in_second_frame():
         assert "transparency" not in im.info
 
         assert_image_equal_tofile(im, "Tests/images/different_transparency_merged.png")
+
+        out = str(tmp_path / "temp.gif")
+        im.save(out, save_all=True)
+
+        with Image.open(out) as reread:
+            reread.seek(reread.tell() + 1)
+            assert_image_equal_tofile(
+                reread, "Tests/images/different_transparency_merged.png"
+            )
 
 
 def test_no_transparency_in_second_frame():
--- a/src/PIL/GifImagePlugin.py
+++ b/src/PIL/GifImagePlugin.py
@@ -573,10 +573,14 @@ def _write_multiple_frames(im, fp, palette):
             im_frame = _normalize_mode(im_frame.copy())
             if frame_count == 0:
                 for k, v in im_frame.info.items():
+                    if k == "transparency":
+                        continue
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by raygard** [comment]
Where you now have:

```python
            if "transparency" in im_frame.info:
                encoderinfo.setdefault("transparency", im_frame.info["transparency"])
            im_frame = _normalize_palette(im_frame, palette, encoderinfo)
```

Does it hurt anything to put the `im_frame = ...` in front?

```python
            im_frame = _normalize_palette(im_frame, palette, encoderinfo)
            if "transparency" in im_frame.info:
                encoderinfo.setdefault("transparency", im_frame.info["transparency"])
```

I ask because it will help my issue #6260 with transparency if they can be swapped around like this.

radarhere:
`_remap_palette`, as the name suggests, reorders the palette. However, it doesn't update the transparency index, meaning that they can become out of sync. This is what happens in #6260. I've added a commit here to fix that.

Also using the swap mentioned in the previous comment, that means that when `_normalize_palette` [calls `_remap_palette`](https://github.com/python-pillow/Pillow/blob/22d797fd81d444fef3d907f4a5913623736db63d/src/PIL/GifImagePlugin.py#L522-L526), the updated transparency can then be used for the frame.

raygard:
Looks good! Your approach is much cleaner than mine was. I hope this can get into the next release.

PR #6189, opened 2022-04-08
Update Image.py

dust-to-dust:
In this pull request:
Add support for difference-aware hash(dHash) calculation. 
DHash value can be used to quickly determine the similarity of two pictures.

Fixes # .

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -2692,6 +2692,78 @@ class Image:
         if not ImageQt.qt_is_installed:
             raise ImportError("Qt bindings are not installed")
         return ImageQt.toqpixmap(self)
+    
+    def dhash(self) -> str:
+    """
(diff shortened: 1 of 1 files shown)


hugovk:
Thanks for the suggestion.

There's already a dhash library that can be used with Pillow and others, why would it be better to have it built in to Pillow?

* https://pypi.org/project/dhash/

Also these:

* https://pypi.org/project/imgsimilarity/
* https://pypi.org/project/dhashpy/
* and more https://pypi.org/search/?q=dhash

There's probably a well-maintained library amongst those, and I'd tend to prefer to let them continue doing a good job rather than stretching the Pillow's maintenance even thinner.

(_If_ this were to be added to Pillow, it would also need tests, docs and release notes. Also Pillow doesn't have any Python dependencies, so we can't require NumPy to be installed.)

dust-to-dust:
I was thoughtless. I didn't take these into consideration. I hope Pillow will become better and stronger. Thank you for your work.

PR #6184, opened 2022-04-05
Add deprecations helper

hugovk:
I noticed in a project using Pillow that the deprecation warning resulted in a very long traceback:

```
=================================== FAILURES ===================================
  ___________________________________ test_pil ___________________________________
  
      def test_pil():
          image_manager = PILImageManager("RGB")
          osm = OSMManager(image_manager=image_manager)
          image, bounds = osm.create_osm_image((30, 31, -117, -116), 9)
          wh_ratio = float(image.size[0]) / image.size[1]
  >       image2 = image.resize((int([80](https://github.com/hugovk/osmviz/runs/5822426177?check_suite_focus=true#step:5:80)0 * wh_ratio), 800), Image.ANTIALIAS)
  
  test/functional/test_manager.py:11: 
  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
  
  name = 'ANTIALIAS'
  
      def __getattr__(name):
          deprecated = "deprecated and will be removed in Pillow 10 (2023-07-01). "
          categories = {"NORMAL": 0, "SEQUENCE": 1, "CONTAINER": 2}
          if name in categories:
              warnings.warn(
                  "Image categories are " + deprecated + "Use is_animated instead.",
                  DeprecationWarning,
                  stacklevel=2,
              )
              return categories[name]
          elif name in ("NEAREST", "NONE"):
              warnings.warn(
                  name
                  + " is "
                  + deprecated
                  + "Use Resampling.NEAREST or Dither.NONE instead.",
                  DeprecationWarning,
                  stacklevel=2,
              )
              return 0
          old_resampling = {
              "LINEAR": "BILINEAR",
              "CUBIC": "BICUBIC",
              "ANTIALIAS": "LANCZOS",
          }
          if name in old_resampling:
  >           warnings.warn(
                  name
                  + " is "
                  + deprecated
                  + "Use Resampling."
                  + old_resampling[name]
                  + " instead.",
                  DeprecationWarning,
                  stacklevel=2,
              )
  E           DeprecationWarning: ANTIALIAS is deprecated and will be removed in Pillow 10 (2023-07-01). Use Resampling.LANCZOS instead.
  
  .tox/py310/lib/python3.10/site-packages/PIL/Image.py:[82](https://github.com/hugovk/osmviz/runs/5822426177?check_suite_focus=true#step:5:82): DeprecationWarning
[...]

PR #6178, opened 2022-04-02
Deprecated PhotoImage.paste() box parameter

radarhere:
Resolves #6163

In [PhotoImage.paste()](https://github.com/python-pillow/Pillow/blob/a575ec33d2f0a04de6cde880a4163de61106b343/src/PIL/ImageTk.py#L178), the `box` parameter is unused.

This PR deprecates it.

PR #6195, opened 2022-04-11
Cleanup: various

hugovk:
Various cleanups, see commit messages for details.

https://github.com/python-pillow/Pillow/commit/6a648c9ce7e06c04051852d48cc68c527bf4f0f7 should be the only functional change: should these be tuples (comma added) or not (remove parenthesis)?

### DIFF
--- a/src/PIL/ImageColor.py
+++ b/src/PIL/ImageColor.py
@@ -121,12 +121,13 @@ def getrgb(color):
 def getcolor(color, mode):
     """
     Same as :py:func:`~PIL.ImageColor.getrgb`, but converts the RGB value to a
-    greyscale value if the mode is not color or a palette image. If the string
+    greyscale value if ``mode`` is not color or a palette image. If the string
     cannot be parsed, this function raises a :py:exc:`ValueError` exception.
 
     .. versionadded:: 1.1.4
 
     :param color: A color string
+    :param mode: Convert result to this mode
     :return: ``(graylevel [, alpha]) or (red, green, blue[, alpha])``
     """
     # same as getrgb, but converts the result to the given mode
--- a/src/PIL/ImageColor.py
+++ b/src/PIL/ImageColor.py
@@ -133,7 +133,7 @@ def getcolor(color, mode):
     # same as getrgb, but converts the result to the given mode
     color, alpha = getrgb(color), 255
     if len(color) == 4:
-        color, alpha = color[0:3], color[3]
+        color, alpha = color[:3], color[3]
 
     if Image.getmodebase(mode) == "L":
         r, g, b = color
--- a/src/PIL/ImageColor.py
+++ b/src/PIL/ImageColor.py
@@ -45,7 +45,7 @@ def getrgb(color):
 
     # check for known string formats
     if re.match("#[a-f0-9]{3}$", color):
-        return (int(color[1] * 2, 16), int(color[2] * 2, 16), int(color[3] * 2, 16))
+        return int(color[1] * 2, 16), int(color[2] * 2, 16), int(color[3] * 2, 16)
 
     if re.match("#[a-f0-9]{4}$", color):
         return (
@@ -56,7 +56,7 @@ def getrgb(color):
         )
 
     if re.match("#[a-f0-9]{6}$", color):
-        return (int(color[1:3], 16), int(color[3:5], 16), int(color[5:7], 16))
+        return int(color[1:3], 16), int(color[3:5], 16), int(color[5:7], 16)
 
     if re.match("#[a-f0-9]{8}$", color):
         return (
@@ -68,7 +68,7 @@ def getrgb(color):
 
     m = re.match(r"rgb\(\s*(\d+)\s*,\s*(\d+)\s*,\s*(\d+)\s*\)$", color)
     if m:
-        return (int(m.group(1)), int(m.group(2)), int(m.group(3)))
+        return int(m.group(1)), int(m.group(2)), int(m.group(3))
 
     m = re.match(r"rgb\(\s*(\d+)%\s*,\s*(\d+)%\s*,\s*(\d+)%\s*\)$", color)
     if m:
@@ -114,7 +114,7 @@ def getrgb(color):
 
     m = re.match(r"rgba\(\s*(\d+)\s*,\s*(\d+)\s*,\s*(\d+)\s*,\s*(\d+)\s*\)$", color)
     if m:
-        return (int(m.group(1)), int(m.group(2)), int(m.group(3)), int(m.group(4)))
+        return int(m.group(1)), int(m.group(2)), int(m.group(3)), int(m.group(4))
     raise ValueError(f"unknown color specifier: {repr(color)}")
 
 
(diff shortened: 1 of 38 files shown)


src/PIL/ImageFont.py
```diff
@@ -634,6 +634,7 @@ def getmask2(
         should have mode ``RGBA``. Otherwise, it should have mode ``1``.
 
         :param text: Text to render.
+        :param fill: A fill function.
```
nulano, line 637: It's not clear to me what is the use for this parameter. It defaults to a function in the private interface, yet AFAICT it is not passed a different value anywhere within Pillow. My best guess is that it was supposed to be used with `ImageDraw.setfill`, deprecated before the fork and removed in Pillow 3.0.0. I think it was undocumented because it was inteded to be private? This docstring is also not particularly helpful.

Would it make more sense to deprecate it instead?

hugovk: Yeah, it probably would be a good idea to deprecate it.

The definition:

```python
    def getmask2(
        self,
        text,
        mode="",
        fill=Image.core.fill,
        direction=None,
        features=None,
        language=None,
        stroke_width=0,
        anchor=None,
        ink=0,
        *args,
        **kwargs,
    ):
```

The tricky thing is that it's the 3rd of many parameters. We can remove it in Pillow 10, a breaking change, but we would want people to be able to modify their code during the 9.x deprecation period and have it still work in 10 with no further changes.

We're not setting `fill` and so use keyword params for all those after `fill`:

```python
        return self.getmask2(
            text,
            mode,
            direction=direction,
            features=features,
            language=language,
            stroke_width=stroke_width,
            anchor=anchor,
            ink=ink,
        )[0]
```

And:

```python
                mask, offset = font.getmask2(
                    text,
                    mode,
                    direction=direction,
                    features=features,
                    language=language,
                    stroke_width=stroke_width,
                    anchor=anchor,
                    ink=ink,
                    *args,
                    **kwargs,
                )
```

So our code won't break. 

We don't know how others are calling `getmask2`. They're _probably_ not setting `fill` but is it safe to assume so?

Well, maybe in Pillow 10 we can set those later arguments to be [keyword-only arguments](https://peps.python.org/pep-3102/).

If they're not using `fill` they must be using the later ones as keyword-only arguments. 

* No warning. 
* They don't need to change their code.

If they are using `fill`, they could be using later ones as either positional or keyword-only.

* In any case, we warn `fill` is deprecated and to use keyword-only for the later ones.
* They can change their code during the deprecation period and it'll still work when the Pillow 10 breaking change happens.

src/PIL/ImageTk.py
```diff
@@ -184,6 +184,7 @@ def paste(self, im, box=None):
         :param im: A PIL image. The size must match the target region.  If the
                    mode does not match, the image is converted to the mode of
                    the bitmap image.
+        :param box: Deprecated.
```
hugovk, line 187: I know this was just removed in https://github.com/python-pillow/Pillow/pull/6178 to prevent people using it, but I think we should keep it and simply say it is deprecated, partly to prevent a docs warning if nothing else.

src/PIL/ImageFont.py
nulano: I don't think it is necessary to make them keyword-only, and this decision can be deferred until the parameter is actually removed anyway.

* If they aren't using `fill`, they are passing them as keyword args and don't need to change their code.
* If they are using `fill` (I expect it is very uncommon), they will have to change to keyword args until it is removed whether or not they are keyword-only.

After the parameter is removed, both can change to use positional args moving forward.

However, it might be worth making them keyword-only due to how many there are. I remember I struggled with keeping them backwards-compatible when working on some previous PRs (e.g. #4930). In that case, it would be worth deprecating positional args for all of the font-related functions.

hugovk: Ah yes, good point. 👍 

> However, it might be worth making them keyword-only due to how many there are. I remember I struggled with keeping them backwards-compatible when working on some previous PRs (e.g. https://github.com/python-pillow/Pillow/pull/4930). 

Yeah, some of them have grown a lot of arguments over the years!

> In that case, it would be worth deprecating positional args for all of the font-related functions.

If I'm calling `getmask2("my text", "RGBA")` now, how would we detect that to warn?

And not warn for `getmask2(text="my text", mode="RGBA")`?

nulano: I would keep `text` as the only positional argument since it is mandatory.

To detect whether `mode` is passed as positional, the function can be changed to `getmask2(text, _mode=None, fill=Image.core.fill, ..., *args, mode="", **kwargs)`. If `_mode` is set, it was passed as a positional argument and should raise a warning. Otherwise it was passed as a keyword argumnent.

hugovk: Yes, I see.

Would you like to put together a PR to deprecate `fill` from `getmask2`? It's pretty clear that parameter isn't much use. 

And for "deprecating positional args for all of the font-related functions": it'd be nice to see a draft PR for a single function first, to get an idea, in case we decide not to do it and avoid wasted effort. Mention which other functions you'd like to be included. No need to do all the docs and stuff, and maybe just a partial test. Feel free to open an issue for further discussion :)

nulano: Sure, but I would prefer to wait until #6184 is merged/rejected first :)

radarhere: #6184 has now been merged.

nulano: > Would you like to put together a PR to deprecate fill from getmask2? It's pretty clear that parameter isn't much use.

See #6220.

> And for "deprecating positional args for all of the font-related functions"

Will do a bit later.

src/PIL/ImageTk.py
radarhere, line 187:
```suggestion
        :param box: Deprecated. This parameter will be removed in Pillow 10
                    (2023-07-01).
```
Shall we include the removal date in the docs?

hugovk: Yes good idea!

src/PIL/ImageFont.py
radarhere, line 637:
```suggestion
```
This has now been added in #6220 instead.

I wondered why it didn't conflict, until I realised it was because this is before `mode`, when it should be after.

src/PIL/ImageColor.py
```diff
@@ -114,19 +114,20 @@ def getrgb(color):
...
     Same as :py:func:`~PIL.ImageColor.getrgb`, but converts the RGB value to a
-    greyscale value if the mode is not color or a palette image. If the string
+    greyscale value if ``mode`` is not color or a palette image. If the string
     cannot be parsed, this function raises a :py:exc:`ValueError` exception.
 
     .. versionadded:: 1.1.4
 
     :param color: A color string
+    :param mode: Convert result to this mode
     :return: ``(graylevel [, alpha]) or (red, green, blue[, alpha])``
```
radarhere, line 131:
```suggestion
    :return: ``(graylevel[, alpha]) or (red, green, blue[, alpha])``
```
Another "various" change for you, making `(graylevel[, alpha])` match the formatting of `(red, green, blue[, alpha])`

PR #6196, opened 2022-04-11
Cleanup: adjust names

hugovk:
Class, function and argument names were only changed for `_private` or tests.

Otherwise, variable names internal to functions were changed for any.

See commit messages for details.

### DIFF
--- a/Tests/helper.py
+++ b/Tests/helper.py
@@ -324,7 +324,7 @@ def is_mingw():
     return sysconfig.get_platform() == "mingw"
 
 
-class cached_property:
+class CachedProperty:
     def __init__(self, func):
         self.func = func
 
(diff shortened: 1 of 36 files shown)


Tests/helper.py
```diff
@@ -45,7 +45,7 @@ def upload(a, b):
 
 else:
     try:
-        import test_image_results
+        import TestImageResults
```
nulano, line 48: I believe this is importing https://pypi.org/project/test-image-results/, https://github.com/wiredfool/test-image-results.

If anything, it might make sense to just remove it.

hugovk: Good point, I've reverted it.

radarhere (approved):

Issue #6154, opened 2022-03-24
Image.show() causes xwayland to crash


xAspirus:
```py
def image_split_into_4(im: Image.Image):
    return (
        im.crop((0          , 0           , im.width//2, im.height//2)), # tl
        im.crop((im.width//2, 0           , im.width   , im.height//2)), # tr
        im.crop((0          , im.height//2, im.width//2, im.height   )), # bl
        im.crop((im.width//2, im.height//2, im.width   , im.height   )), # br
    )

im = Image.open('test.png')
x = image_split_into_4(im)
x[0].show()
x[1].show()
x[2].show()
x[3].show()
```

will cause xwayland server to crash

hugovk:
### What are your OS, Python and Pillow versions?

* OS: 
* Python: 
* Pillow: 

Did it output any error traceback? Please paste it in full.

Does it happen with all images or only certain ones?

radarhere:
What viewers are detected by Pillow?
```pycon
>>> from PIL import ImageShow
>>> ImageShow._viewers
[<PIL.ImageShow.MacViewer object at 0x1050706d0>]
```

radarhere:
When Pillow "shows" an image on Linux, it is running `xdg-open`, `display`, `gm`, `eog` or `xv` on the command line. If there is a crash after that, it is because of a problem in the Linux software, not within Pillow.

It is possible that the PNG file being saved by Pillow is corrupt in some way, or that we have chosen an obscure method of writing the file that is not fully supported by the Linux software. This could result in Linux being unable to display the image correctly, but it shouldn't result in a crash.

If your problem is easily reproducible, that would be interesting. If it is a common problem, then we should see if there is anything we can do to help from our side. But I think the most likely reason for this situation is that there is a problem in your installation.

radarhere:
@xAspirus did you have any more information for us?

radarhere:
Closing, unless more information can be provided.

Issue #6181, opened 2022-04-05
When I exited my virtual environment it seemed to work


Aditya-76:
When I exited my virtual environment it seemed to work

_Originally posted by @akackon in https://github.com/python-pillow/Pillow/issues/3851#issuecomment-657049322_

Why is that the case? I want to use pillow within the virtual environment. How do I fix it?

hugovk:
Thank you for reporting an issue.

Follow these guidelines to ensure your issue is handled properly.

If you have a ...

1.  General question: consider asking the question on Stack Overflow
    with the python-imaging-library tag:

    * https://stackoverflow.com/questions/tagged/python-imaging-library

    Do not ask a question in both places.

    If you think you have found a bug or have an unexplained exception
    then file a bug report here.

2.  Bug report: include a self-contained, copy-pastable example that
    generates the issue if possible. Be concise with code posted.
    Guidelines on how to provide a good bug report:

    * https://stackoverflow.com/help/mcve

    Bug reports which follow these guidelines are easier to diagnose,
    and are often handled much more quickly.

3.  Feature request: do a quick search of existing issues
    to make sure this has not been asked before.

We know asking good questions takes effort, and we appreciate your time.
Thank you.

### What did you do?

### What did you expect to happen?

### What actually happened?

### What are your OS, Python and Pillow versions?

* OS: 
* Python: 
* Pillow: 

Please include **code** that reproduces the issue and whenever possible, an **image** that demonstrates the issue. Please upload images to GitHub, not to third-party file hosting sites. If necessary, add the image to a zip or tar archive.

The best reproductions are self-contained scripts with minimal dependencies. If you are using a 

```python
code goes here
```

radarhere:
> Why is that the case?

I don't expect that Pillow was broken because the user's environment was virtual. I expect it was just because they had installed Pillow in their normal environment and not in their virtual environment.

radarhere:
If you think you have installed Pillow, but you can't import it, then I suggest you install Pillow in the same way that you run your script.

What does that mean? If you had run
```
pip install Pillow
python3 script.py
```
and then found that `python3` couldn't detect Pillow, it might be because `pip` is connected to a different environment than `python3` is.

So in https://pillow.readthedocs.io/en/stable/installation.html#basic-installation, we recommend
> python3 -m pip install --upgrade pip
> python3 -m pip install --upgrade Pillow

That way, when you run `python3 script.py`, Pillow will be found.

[...]

PR #6203, opened 2022-04-12
Use actions/stale to close 'Awaiting OP Action' with no response

hugovk:
We often get issues where we ask the OP for more info but they never reply.

Most recently, no reply in a week after four comments from us: https://github.com/python-pillow/Pillow/issues/6181. Or already closed: https://github.com/python-pillow/Pillow/issues/6154, no reply after five comments from us.

We often label them ["Awaiting OP Action"](https://github.com/python-pillow/Pillow/issues?q=label%3A%22Awaiting+OP+Action%22+sort%3Aupdated-desc+) for our own convenience.

GitHub has an action to automatically close "stale" issues: https://github.com/actions/stale

There's a whole bunch of config. It can:

* add a "stale" label and/or a comment after some time, 
* and close the issue after more time.

For example, we could have this:

* We manually add the "Awaiting OP Action". 
* If no-one replies after, say, a week, it comments that it will be closed after another couple of days, and if still no action, just closes it.
* If someone does reply, the action removes "Awaiting OP Action".

Or, as in this PR, just keep it simple:

* We manually add the "Awaiting OP Action". 
* If no-one replies after, say, a week, the issue is closed with a comment.
* If someone does reply, the action removes "Awaiting OP Action".

And of course, we can always re-open issues.

---

The second commit here sets `debug-only: true` to do a dry run, so no issues will be closed, and also to set `days-before-close: 1`, so we can test it sooner and see the logs.

Let's keep https://github.com/python-pillow/Pillow/issues/6181 open for testing.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -16,7 +16,7 @@ jobs:
 
     steps:
     - name: "Check issues"
-      uses: actions/stale@v4
+      uses: actions/stale@v5
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
         debug-only: true
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,26 @@
+name: Close stale issues
+
+on:
+  schedule:
+  - cron: "10 0 * * *"
+  workflow_dispatch:
+
+permissions:
+  issues: write
+
+jobs:
+  stale:
+    if: github.repository_owner == 'python-pillow'
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: "Check issues"
+      uses: actions/stale@v4
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        only-issue-labels: "Awaiting OP Action"
+        close-issue-message: "Closing this issue as no feedback has been received."
+        days-before-stale: -1
+        days-before-close: 7
+        labels-to-remove-when-unstale: "Awaiting OP Action"
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -19,8 +19,9 @@ jobs:
       uses: actions/stale@v4
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
+        debug-only: true
         only-issue-labels: "Awaiting OP Action"
         close-issue-message: "Closing this issue as no feedback has been received."
         days-before-stale: -1
-        days-before-close: 7
+        days-before-close: 1
         labels-to-remove-when-unstale: "Awaiting OP Action"


.github/workflows/stale.yml
```diff
@@ -0,0 +1,27 @@
...
+
+jobs:
+  stale:
+    if: github.repository_owner == 'python-pillow'
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: "Check issues"
+      uses: actions/stale@v4
```
radarhere, line 19:
```suggestion
      uses: actions/stale@v5
```

radarhere (approved):
Sure, let's give this a try

hugovk:
I triggered a run:

https://github.com/python-pillow/Pillow/actions/runs/2154611321

A few notes.

---

Checking the first issue it processed:

image

It skipped it because it didn't have the "Awaiting OP Action" label. ✅ 

---

Checking the first PR it processed:

image

Hmm, we don't really want it to process any PRs. I don't think I see a config setting that says skip PRs.

First of all, it only has permission to write/close issues, not PRs, so we're safe there.

```yml
permissions:
  issues: write
```

Two further things to do:

```yml
        only-issue-labels: "Awaiting OP Action"
```

This "Override `only-labels` for issues only". So we replace it with this, so it'll quickly skip PRs without the special label:

```yml
        only-labels: "Awaiting OP Action"
```

Second, we can make `days-before-close: 1` specific to issues and set `-1` to disable for PRs:

```yml
        days-before-issue-close: 1
        days-before-pr-close: -1
```

I'll PR these updates.

---

Checking #6181:

image

* It has the label ✅ 
* It's not yet old enough to be stale ✅

Let's check tomorrow.

hugovk:
After triggering a build with https://github.com/python-pillow/Pillow/pull/6204:

https://github.com/python-pillow/Pillow/runs/5990629124?check_suite_focus=true

First issue skipped due to no special label ✅ 

image

---

First PR skipped due to no special label ✅ 

image

---


#6181 as before ✅ 

image

hugovk:
Checking the cron run 12 hours ago (and noting got the cron wrong, I meant to do 10:00 UTC instead of 00:10 UTC! It doesn't really matter :)

https://github.com/python-pillow/Pillow/runs/5999629985?check_suite_focus=true

First issue and PR skipped as before ✅ 

image

#6181:

image

```
This issue is not stale
This issue should be stale based on the last update date the 12-04-2022 (2022-04-12T07:13:51Z)
This issue should not be marked as stale based on the option days-before-stale (​[https://github.com/actions/stale#days-before-stale​)](https://github.com/actions/stale#days-before-stale%E2%80%8B)) (-1)
```

This is as before, but we want something to have happened by now.

I think what is happening: it's not stale. It's eligible to be stale based on age, but not based on that `days-before-stale: -1`.

I think it needs to get into the stale state first before it can consider closing.

So I think we want:

```yml
days-before-stale: 7 # (or 1 in the test)
days-before-issue-close: 0
```

That should set it to stale after some time, and then hopefully make it available for closure immediately. It's possible it will wait until the next batch run though, but let's see.

I'm going to edit the `yml` file directly as below, trigger a run and report back!

```yml
days-before-stale: 1
days-before-issue-close: 0
```

hugovk:
-> https://github.com/python-pillow/Pillow/commit/fb361fc89ebabc8a12bd645947e88eeca40932bc

Looks good, it would have closed it right away!

image

This was in debug mode so didn't actually close anything.

I'll remove `debug-only: true` and trigger another run, it should close it.

Then all being well I'll bump `days-before-stale: 1` to `days-before-stale: 7` and we should be all set!

hugovk:
-> https://github.com/python-pillow/Pillow/commit/317ea9a8bbc061a8276f789d8fec626b84618673

https://github.com/python-pillow/Pillow/actions/runs/2161332162

image

Success! Issue autoclosed:

https://github.com/python-pillow/Pillow/issues/6181#issuecomment-1098026769

And finally I've bumped it to seven days:

https://github.com/python-pillow/Pillow/commit/0329b8fa1d7ad3ae39a171a903b328a56046003a

hugovk:
Oh, I'd already added that in this very PR :)

PR #6220, opened 2022-04-16
Deprecate FreeTypeFont.getmask2 fill parameter

nulano:
First part for https://github.com/python-pillow/Pillow/pull/6195#discussion_r847410876

Deprecate undocumented FreeTypeFont.getmask2 fill parameter that expects an internal function by default.

### DIFF
--- a/Tests/test_imagefont.py
+++ b/Tests/test_imagefont.py
@@ -977,6 +977,11 @@ class TestImageFont:
 
         assert_image_similar_tofile(im, "Tests/images/colr_bungee_mask.png", 22)
 
+    def test_fill_deprecation(self):
+        font = self.get_font()
+        with pytest.warns(DeprecationWarning):
+            font.getmask2("Hello world", fill=Image.core.fill)
+
 
 @skip_unless_feature("raqm")
 class TestImageFont_RaqmLayout(TestImageFont):
--- a/Tests/test_imagefont.py
+++ b/Tests/test_imagefont.py
@@ -981,8 +981,8 @@ class TestImageFont:
         font = self.get_font()
         with pytest.warns(DeprecationWarning):
             font.getmask2("Hello world", fill=Image.core.fill)
-        with pytest.raises(TypeError):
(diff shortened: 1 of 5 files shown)


nulano:
The ubuntu-20.04 docker build failed due to a Tiff timeout and MacOS PyPy3.7 failed with a segfault while collecting tests.

radarhere:
I've rerun the tests and they pass now.

Tests/test_imagefont.py
```diff
@@ -977,6 +977,14 @@ def test_colr_mask(self):
 
         assert_image_similar_tofile(im, "Tests/images/colr_bungee_mask.png", 22)
 
+    def test_fill_deprecation(self):
+        font = self.get_font()
+        with pytest.warns(DeprecationWarning):
+            font.getmask2("Hello world", fill=Image.core.fill)
+        with pytest.raises(TypeError):
+            with pytest.warns(DeprecationWarning):
```
radarhere, line 985:
```suggestion
        with pytest.warns(DeprecationWarning):
            with pytest.raises(TypeError):
```
Testing with main, the test only fails if the checks are in this order.

nulano: You are right, fixed. I thought I tested this, but I guess I was confused by the previous check...

src/PIL/ImageFont.py
```diff
@@ -674,6 +683,10 @@ def getmask2(
                  :py:mod:`PIL.Image.core` interface module, and the text offset, the
                  gap between the starting coordinate and the first marking
         """
+        if fill is not _UNSPECIFIED:
+            deprecate("fill", 10)
+        else:
+            fill = Image.core.fill
```
hugovk, line 689: Can we flip so the conditional is positive?

```suggestion
        if fill is _UNSPECIFIED:
            fill = Image.core.fill
        else:
            deprecate("fill", 10)
```

hugovk (approved):
Thank you!

PR #6221, opened 2022-04-18
Run Tidelift Align on Pipfile changes

radarhere:
Updated Pipfile to Python 3.10

### DIFF
--- a/.github/workflows/tidelift.yml
+++ b/.github/workflows/tidelift.yml
@@ -4,9 +4,11 @@ on:
     - cron: "30 2 * * *"  # daily at 02:30 UTC
   push:
     paths:
+      - "Pipfile*"
       - ".github/workflows/tidelift.yml"
   pull_request:
     paths:
+      - "Pipfile*"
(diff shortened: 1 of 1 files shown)


hugovk:
Does the lockfile need updating?

hugovk:
And because these files are only used by https://github.com/python-pillow/Pillow/blob/main/.github/workflows/tidelift.yml should we edit that to also run when these files change?

radarhere:
Good call, done.

aclark4life:
```
Error: No API key found. Please authenticate with the Tidelift API as described at https://ptm.tl/docs-apikeys.
```
Hmmm

radarhere:
I'm going to abandon my original idea of updating, and instead just keep @hugovk's idea of running Tidelift Align for Pipfile changes.

hugovk:
PR https://github.com/python-pillow/Pillow/pull/6233 is the same as this but from a branch in this repo, instead of from a personal fork.

This one failed because it can't find the API key, likely a GitHub security precaution to prevents forks accessing secrets.

PR #6224, opened 2022-04-18
Increase wait time of temporary file deletion on Windows

AlexTedeschi:
Increase wait time to 4 seconds from 2 seconds for the deletion of the temporary file made from .show(). The Windows default image viewer at times can take longer than 2 seconds to open and display an image causing the image to never display for the user.

Fixes # Line 144, Increase wait time

 * Increase wait time to 4 seconds from 2 seconds for deletion of temporary image file

### DIFF
--- a/src/PIL/ImageShow.py
+++ b/src/PIL/ImageShow.py
@@ -141,7 +141,7 @@ class WindowsViewer(Viewer):
     def get_command(self, file, **options):
         return (
             f'start "Pillow" /WAIT "{file}" '
-            "&& ping -n 2 127.0.0.1 >NUL "
+            "&& ping -n 4 127.0.0.1 >NUL "
             f'&& del /f "{file}"'
         )
 


radarhere:
Turns out I missed a spot in https://github.com/multi-build/multibuild/pull/468, and so also needed https://github.com/multi-build/multibuild/pull/469

radarhere:
Given that it's taken until now for this PR to be created, 2 seconds has presumably been enough for most Windows users of Pillow so far. Not saying we shouldn't merge this, I just think it might be of interest to know - do you have any ideas why your system is different? Is your machine slow? Are you dealing with very large images?

AlexTedeschi:
What is weird is that it was enough for quite a while. But I am running a program for research that is reverse annotating images and putting the location found as text on top of the image. I had no changed anything and then 2 days later it was all of a sudden too slow. I feel any intensive program may run cause this issue to arise. It broke at the same time on both of my computers, both of which are high end, so I am guessing some library updated and got larger causing it to slow down. When I switched it to 4, it worked perfectly on both machines. I felt this change would be nice cause it would only help those who run into the same issue as I did and shouldn’t hurt others since its just the wait time on the deletion of the temporary image.

radarhere:
Given that the timeout on macOS is 20 seconds, 4 seconds does not seem unreasonable.

PR #6235, opened 2022-04-21
Fixed enum typo

Iniquitatis:


radarhere:
Thanks for finding that. I've created https://github.com/Iniquitatis/Pillow/pull/1 to fix this in another location as well.

hugovk:
Thanks both!

PR #6242, opened 2022-04-25
Raise ValueError for invalid PPM maxval

### LINKED ISSUE #6240: ZeroDivisionError is raised when the `maxval` of the PPM metadata is 0
Hi,

Thank you for developing and maintaining this project.

Recently I am fuzzing this library for fun, and I found the test case which raises ZeroDivisionError.

### What did you do?

```python3
from PIL import Image
import io

# Load hopper.ppm. Assumes the current directory is the root of this repository.
with open('Tests/images/hopper.ppm', 'rb') as f:
    data = bytearray(f.read())

# Sets the maximum value 0
data[49] = 0x30
data[50] = 0x30
data[51] = 0x30

# Raises ZeroDivisionError
with Image.open(io.BytesIO(data)) as img:
    img.load()
```

The file is broken, so some error should be raised. But `ZeroDivisionError` may be a little confusing.

### What actually happened?

```
  File "issue1.py", line 15, in <module>
    img.load()
  File "/usr/local/lib/python3.8/site-packages/PIL/ImageFile.py", line 234, in load
    err_code = decoder.decode(b"")[1]
  File "/usr/local/lib/python3.8/site-packages/PIL/PpmImagePlugin.py", line 152, in decode
    value = min(out_max, round(value / maxval * out_max))
ZeroDivisionError: division by zero
```

* OS: Debian GNU/Linux 11 (bullseye)
* Python: 3.8.13
* Pillow: 9.1.0

I am happy to create the pull request to fix this issue, but I am not very sure which exception should be raised here.

radarhere:
Resolves #6240

http://netpbm.sourceforge.net/doc/ppm.html states that maxval "Must be less than 65536 and more than zero". The issue found that a maxval of zero raised a ZeroDivisionError instead.

Instead, if `maxval` is outside of the valid range, this PR raises a ValueError, "Invalid maxval".

### DIFF
--- a/src/PIL/PpmImagePlugin.py
+++ b/src/PIL/PpmImagePlugin.py
@@ -116,6 +116,10 @@ class PpmImageFile(ImageFile.ImageFile):
                     break
             elif ix == 2:  # token is maxval
                 maxval = token
+                if not 0 < maxval < 65536:
+                    raise ValueError(
+                        "maxval must be greater than 0 and less than 65536"
+                    )
                 if maxval > 255 and mode == "L":
                     self.mode = "I"
 
--- a/Tests/test_file_ppm.py
+++ b/Tests/test_file_ppm.py
@@ -145,6 +145,19 @@ def test_truncated_file(tmp_path):
             im.load()
 
 
+@pytest.mark.parametrize("maxval", (0, 65536))
+def test_invalid_maxval(maxval, tmp_path):
+    path = str(tmp_path / "temp.ppm")
+    with open(path, "w") as f:
+        f.write("P6\n3 1 " + str(maxval))
+
+    with pytest.raises(ValueError) as e:
+        with Image.open(path):
+            pass
+
+    assert str(e.value) == "maxval must be greater than 0 and less than 65536"
+
+
 def test_neg_ppm():
     # Storage.c accepted negative values for xsize, ysize.  the
     # internal open_ppm function didn't check for sanity but it


src/PIL/PpmImagePlugin.py
```diff
@@ -116,6 +116,8 @@ def _open(self):
                     break
             elif ix == 2:  # token is maxval
                 maxval = token
+                if maxval >= 65536 or maxval == 0:
+                    raise ValueError("Invalid maxval")
```
hugovk, line 120: Shall we also say what the valid range is?

src/PIL/PpmImagePlugin.py
```diff
@@ -116,6 +116,8 @@ def _open(self):
                     break
             elif ix == 2:  # token is maxval
                 maxval = token
+                if maxval >= 65536 or maxval == 0:
+                    raise ValueError("Maxval not greater than 0 and less than 65536")
```
hugovk, line 120: This is a little ambiguous, it could be what is wrong or what it should be.

Maybe something along the lines of "Maxval must be greater than 0 and less than 65536" or "Maxval must be between 1-65535"?

(And lowercase maxval even if at start of the sentence, the case is important for a variable.)

src/PIL/PpmImagePlugin.py
```diff
@@ -116,6 +116,10 @@ def _open(self):
                     break
             elif ix == 2:  # token is maxval
                 maxval = token
+                if maxval >= 65536 or maxval == 0:
```
SimonSegerblomRex, line 119: Slightly improved readability:
```python3
if not 0 < maxval < 65536:
```

radarhere: Fair enough. I've updated the commit

src/PIL/PpmImagePlugin.py
radarhere: Ok, updated.

hugovk:
Thanks!

PR #6244, opened 2022-04-26
Corrected screencapture argument in ImageGrab.grab()

axt-one:
I have found a bug in ImageGrab.grab().

File "src/PIL/ImageGrab.py", line 36,
`args += ["-R", f"{left},{right},{right-left},{bottom-top}"]`
seems to be a mistake and ImageGrab.grab() does not return expected results.

As ImageGrab.grab() is implemented with "screencapture" subprocess which takes an option -R "x,y,width,height" to capture a rectangle area, it must be `args += ["-R", f"{left},{top},{right-left},{bottom-top}"]`.


radarhere:
Thanks for finding this.

PR #6247, opened 2022-04-27
Updated URL

radarhere:
https://www.exiv2.org/tags.html does not work.
https://exiv2.org/tags.html does.


hugovk (approved):
See also https://github.com/radarhere/Pillow/pull/16

PR #6254, opened 2022-05-01
Support more affine expression forms in im.point()

benrg:
In modes I and F, Image.point only supported affine expressions of the forms (lambda x:) x * a, x + a, and x * a + b. Expressions like 1 - x had to be written x * -1 + 1.

This rewrite, though still limited to affine transformations, supports far more expression forms, including 1 - x, (2 * x + 1) / 3, etc.


radarhere:
Thanks. I've created https://github.com/benrg/Pillow/pull/1 with some suggestions.

radarhere (approved):

Issue #6268, opened 2022-05-04
[JPEG 2000] OverflowError is raised when the size is too large


ks888:
Hi, I am recently fuzzing this library and found the test case which raises OverflowError.

### What did you do?

```python

from PIL import Image
import io

# assumes the current directory is the root of this repository.
with open('Tests/images/16bit.cropped.j2k', 'rb') as f:
    data = bytearray(f.read())

# Change the main header of codestream
data[8:12] = b"\xff\x00\x00\x00"  # xsiz
data[12:16] = b"\x00\x00\x00\x00"  # ysiz
data[16:20] = b"\x00\x00\x00\x00"  # xosiz
data[20:24] = b"\x00\x00\x00\x01"  # yosiz

# OverflowError is raised
with Image.open(io.BytesIO(data)) as img:
    img.load()

```

### What did you expect to happen?

Maybe ValueError is better.

### What actually happened?

```
Traceback (most recent call last):
  File "issue4.py", line 17, in <module>
    img.load()
  File "/usr/local/lib/python3.8/site-packages/PIL/Jpeg2KImagePlugin.py", line 278, in load
    return ImageFile.ImageFile.load(self)
  File "/usr/local/lib/python3.8/site-packages/PIL/ImageFile.py", line 201, in load
    self.load_prepare()
  File "/usr/local/lib/python3.8/site-packages/PIL/ImageFile.py", line 271, in load_prepare
    self.im = Image.core.new(self.mode, self.size)
OverflowError: signed integer is greater than maximum
```

The size of ImageFile is `(4278190080, -1)`. The first element of size is too large for `PyArg_ParseTuple`. Also, the test case passes through the bomb check because the second element is negative.

### What are your OS, Python and Pillow versions?

* OS: Debian GNU/Linux 11 (bullseye)
* Python: 3.8.13
* Pillow: 9.1.0

radarhere:
It doesn't address your main concern, but I've created #6269 to prevent the negative height.

radarhere:
After thinking this through, I've updated #6269 - it now prevents zero or negative height. That should resolve this.

radarhere:
As this is a security problem, we've just released the fix for this as part of Pillow 9.1.1.

If you find any further places where `_decompression_bomb_check` is skipped, or any other security problems, please follow our [security policy](https://github.com/python-pillow/Pillow/security/policy).

ks888:
OK. Thank you for the advice.

PR #6269, opened 2022-05-04
Do not allow negative height in JPEG2000 images

radarhere:
#6268 found that JPEG2000 images can have negative heights.

This PR sets a minimum height of zero instead.


hugovk (approved):

hugovk:
Failing `Tests/test_imagetk.py::test_bitmapimage` on macOS/3.7 is unrelated.

PR #6270, opened 2022-05-04
Adjust BITSPERSAMPLE to match SAMPLESPERPIXEL

### LINKED ISSUE #6266: PIL.UnidentifiedImageError on TIFF file (regression from 9.0.1 to 9.1.0)
### What did you do?

The attached TIFF file, which was generated by [Hugin](http://hugin.sourceforge.net/), fails to open in Pillow 9.1.0. It works fine in 9.0.1 and prior versions.

[example.zip](https://github.com/python-pillow/Pillow/files/8612938/example.zip)

With 9.0.1:
```pycon
>>> import PIL
>>> PIL.__version__
'9.0.1'
>>> from PIL import Image
>>> Image.open('eqr.tiff')
<PIL.TiffImagePlugin.TiffImageFile image mode=RGBA size=512x256 at 0x1073FC4C0>
```

### What actually happened?

With 9.1.0:
```pycon
>>> import PIL
>>> PIL.__version__
'9.1.0'
>>> from PIL import Image
>>> Image.open('eqr.tiff')
  File "/private/tmp/test/lib/python3.9/site-packages/PIL/Image.py", line 3123, in open
    raise UnidentifiedImageError(
PIL.UnidentifiedImageError: cannot identify image file 'eqr.tiff'
```

* OS: Tested on macOS and Linux
* Python: Tested on 3.9.10 and 3.9.12
* Pillow: 9.1.0

radarhere:
Resolves #6266

In TiffImagePlugin, there is code to try and match BITSPERSAMPLE to the length that would be expected from the PHOTOMETRIC_INTERPRETATION tag and the EXTRASAMPLES tag.

https://github.com/python-pillow/Pillow/blob/9d988dab6aeb4346960a24b2ec423d8f16f8cddc/src/PIL/TiffImagePlugin.py#L1356-L1365

However, immediately following that an error is raised if the BITSPERSAMPLE length doesn't match SAMPLESPERPIXEL.

https://github.com/python-pillow/Pillow/blob/9d988dab6aeb4346960a24b2ec423d8f16f8cddc/src/PIL/TiffImagePlugin.py#L1367-L1372

So adjustments to match `bps_count` are wasted if that doesn't match `samples_per_pixel`.

This changes the code to try and match `samples_per_pixel` instead.


kevinmehall:
Tested this and confirmed that the affected images can be opened and the alpha channel is preserved.

Thanks!

PR #6292, opened 2022-05-13
Always use GIF89a for comments

### LINKED ISSUE #6291: GIF files with comments must be GIF89a
### What did you do?
Ran the GIF tests in Pillow-9.1.0, looked at output from `test_comment_over_255(tmp_path)` test.

The output GIF file should be GIF89a because it contains a comment.

### What actually happened?
The file temp.gif has a signature of GIF87a.

* OS: Windows 10
* Python: 3.10.4
* Pillow: 9.1.0

For code, see `def test_comment_over_255(tmp_path):` in `Tests\test_file_gif.py`.

The generated image file is:

temp

Discussion:

This is due to an obsolete line in `_get_global_header(im, info)` in `GifImagePlugin.py`:
```python
                extensionKey == "comment" and not (1 <= len(info[extensionKey]) <= 255)
```

The `len(info[extensionKey])` should be simply tested for 0 to indicate no comment is wanted.
Suggested change is
```python
                extensionKey == "comment" and len(info[extensionKey]) == 0
```

I have created a pull request #6292 for this.

As a side note, I think the entire feature test for GIF 87a vs. 89a can be changed from:

```python
    version = b"87a"
            if (extensionKey == "duration" and info[extensionKey] == 0) or (
                extensionKey == "comment" and len(info[extensionKey]) == 0
            ):
                continue
            version = b"89a"
            break
    else:
        if im.info.get("version") == b"89a":
            version = b"89a"
```

to:

```python
    version = b"87a"
    if (im.info.get("version") == b"89a") or info and (
            "transparency" in info or
            "loop" in info or
            ("duration" in info and info["duration"] != 0) or
            ("comment" in info and len(info["comment"]) != 0)):
        version = b"89a"
```

I think that's correct but would be happy for someone to look it over.

raygard:
Fix bug that allows GIFs with long comments to be written as GIF87a.

Fixes #6291 .

 * Change version feature test in `_get_global_header()` to use GIF89a for long comments.
 * Add test in `test_comment_over_255()` to check this.


src/PIL/GifImagePlugin.py
```diff
@@ -915,7 +915,7 @@ def _get_global_header(im, info):
     for extensionKey in ["transparency", "duration", "loop", "comment"]:
         if info and extensionKey in info:
             if (extensionKey == "duration" and info[extensionKey] == 0) or (
-                extensionKey == "comment" and not (1 <= len(info[extensionKey]) <= 255)
+                extensionKey == "comment" and len(info[extensionKey]) == 0
```
radarhere, line 918:
```suggestion
                extensionKey == "comment" and info[extensionKey]
```
This seems simpler to me?

radarhere:
This forces GIF89a for all comments, not just "long comments". So wouldn't it be more straightforward to modify `test_comment`, rather than `test_comment_over_255`?

src/PIL/GifImagePlugin.py
raygard: The test needs to determine that `info[comment]` is an empty string to keep `version` as `87a`. Your suggestion `info[extensionKey]` does the opposite. I was attempting to extrapolate from `not (1 <= len(info[extensionKey]) <= 255)` (which originated with PR #1896 but was superseded by PR #3479) to retain the length test. I think it probably could be `info[extensionKey] == ""` or even `not info[extensionKey]`. The 255 was there to begin with, I think, to fit with the maximum length of the single data `block` that was allowed before PR #3479.

radarhere: You're right - I meant `not info[extensionKey]`. That was just a mistake on my part.

radarhere:
Ok, thanks. That's perfectly fair.

radarhere (approved):

radarhere:
I've created https://github.com/raygard/Pillow/pull/1 with a suggestion from me, to add the check to `test_comment()` as well, and a version of your suggestion from https://github.com/python-pillow/Pillow/issues/6291#issue-1235628952 to simplify the version check.

PR #6294, opened 2022-05-13
Separate multiple GIF comment blocks in a frame with newlines

### LINKED ISSUE #6293: GIF: multiple comment blocks for a frame are run together
### What did you do?
Read a GIF file with two or more comments in a single frame

They would somehow be kept separate in the `.info` dict

### What actually happened?
The comments are run together, e.g. separate comments `foo` and `bar` become `foobar`

* OS: Windows 10
* Python: 3.10.4
* Pillow: 9.1.0

Example image:

multiple_comments

```python
from PIL import Image
with Image.open("multiple_comments.gif") as im:
    print(f'{im.info["comment"]=}')
```
prints 
```
im.info["comment"]=b'Test comment 1Test comment 2'
```
I made PR #6294 to address this, that will separate the comments with `\r\n`.

I admit that such cases are rare, but I have a few found in the wild. Unfortunately most of them have the multiple comments after the last image block, and at present I see no way in Pillow to access comments beyond the last image block. I do have one with 10 comment blocks before the image, a logo from my former employer:

b6483cc24863f387173dd6fb5a150ede_image003

raygard:
If more than one comment is in a GIF frame, separate them with \r\n in the info dict.

Fixes #6293 .

 * When reading GIF file, multiple comments in a frame are separated with \r\n, rather than being concatenated.
 * Add a test to check this for the first frame of an image.
 * Add an image to the test set that contains two comment blocks before the first frame (as well as several other comment blocks.)


radarhere:
If I suggest `\n` instead of `\r\n`, how do you feel about that?

raygard:
As mentioned in the corresponding issue, I found all the cases of more than one comment per GIF that I could find easily in my files. I see one where a single `\n` is used, but several that use `\r\n` to separate lines within a comment. I'm fine with any way you want to go on this issue. It's far less important to me than PR #6300, the one that fixes the comment-per-frame problem, that will (along with PR #6211) probably fix the problem @smccombie has with animated GIFs in MS Teams.

PR #6300, opened 2022-05-14
Only write GIF comments at the beginning of the file

### LINKED ISSUE #6299: GIF comment blocks should be placed near front of file, once
### What did you do?

Opened an animated GIF, wrote it out with a `comment=` argument to `.save`.

I expected the comment to be written once, preferably at the beginning of the file, per advice in (GIF89a spec)[https://www.w3.org/Graphics/GIF/spec-gif89a.txt] 

> it is recommended that Comment Extensions do not interfere with Control or Data blocks; they should be located at the beginning or at the end of the Data Stream to the extent possible.

### What actually happened?
The comment was repeated in every frame.

* OS: Windows 10
* Python: 3.10.4
* Pillow: 9.1.0

dispose_prev

```python
with Image.open("dispose_prev.gif") as im:
    frames = ImageSequence.all_frames(im)
    frames[0].save("testout.gif", save_all=True, append_images=frames[1:], comment="Test comment")
    with Image.open("testout.gif") as img:
        for k, frame in enumerate(ImageSequence.Iterator(img)):
            if "comment" in frame.info:
                print(f'{k=} {frame.info["comment"]=}')
```

Output with 9.1.0:

```
k=0 frame.info["comment"]=b'Test comment'
k=1 frame.info["comment"]=b'Test comment'
k=2 frame.info["comment"]=b'Test comment'
k=3 frame.info["comment"]=b'Test comment'
k=4 frame.info["comment"]=b'Test comment'
```
#### Discussion:

Pillow's handling of GIF comments needs to be changed. A comment can be specified with `comment="..."` arg to `.save()` or by putting a comment in the `.info` dict of the Image object being saved. It is now being written in every frame, and placed after the Graphic Control Extension block (if one is written) and before the Image Descriptor. The placement is not prohibited by the standard, but is contrary to the "Recommendations" in part 24e of the standard and also to the GIF Grammar in Appendix B.

I have written  PR #6300 to resolve this. It will write the comment once only, before the first Graphic Control Block (if any) and the first Image Descriptor. After PR #6211 is merged, this should cause the comment to appear just after the "NETSCAPE2.0" looping control block, if one is written.

raygard:
Helps https://github.com/python-pillow/Pillow/discussions/6207

Place GIF comment after Global Color table. Should go after "NETSCAPE" looping extension after pull #6211.

Fixes #6299.

 * Move writing of comment block from `_write_local_header()` to `_get_global_header()`
 * Add a test `test_write_comment()` to validate that a comment is written to the first frame only
 * Add a test `test_write_no_comment()` to validate that an empty `comment=""` arg to `.save` leaves no comments in saved file
 * Tests use image file `multiple_comments.gif` added in PR #6294 but could use any animated GIF instead — CHANGED to `dispose_prev.gif’`

 The change here to  `_get_global_header()` overlaps the change in PR #6211. Please be sure the merge is done so that the `if "comment" in info and len(info["comment"]):` code in this PR is placed after the `if "loop" in info:` code in PR #6211. The comment must come after the "NETSCAPE2.0" block.


raygard:
I updated the tests to use `dispose_prev.gif` so that the automated tests would complete.

raygard:
@radarhere I am fine with your changes to this. Thanks.

radarhere:
https://github.com/raygard/Pillow/pull/2 keeps the last comment read in `info`. The thinking is that if GIF comments are not intended to exist on a per-frame basis, then it would seem more correct to keep them when seeking to the the next frame - this way, if there is only a comment at the start of a GIF file, it is present for throughout subsequent seek operations.

With that change, the following code also works before and after this PR, helping backwards compatibility.

```python
from PIL import Image
im = Image.new("RGB", (1, 1))
im2 = Image.new("RGB", (1, 1), "#f00")
im.save("out.gif", save_all=True, append_images=[im2], comment=b"comment")

reloaded = Image.open("out.gif")
reloaded.seek(1)
assert reloaded.info["comment"] == b"comment"
```

PR #6310, opened 2022-05-19
Helpful error message on encode/decode

boxed:

 * A nicer error message when encoding/decoding images. The code knows what encoders/decoders are available. It should just say what is valid so the user can fix their problem.

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -400,7 +400,8 @@ def _getdecoder(mode, decoder_name, args, extra=()):
         # get decoder
         decoder = getattr(core, decoder_name + "_decoder")
     except AttributeError as e:
-        raise OSError(f"decoder {decoder_name} not available") from e
+        valid_decoders = ', '.join(DECODERS.keys())
+        raise OSError(f"decoder {decoder_name} not available. Valid decoders: {valid_decoders}") from e
     return decoder(mode, *args + extra)
 
 
@@ -423,7 +424,8 @@ def _getencoder(mode, encoder_name, args, extra=()):
         # get encoder
         encoder = getattr(core, encoder_name + "_encoder")
     except AttributeError as e:
-        raise OSError(f"encoder {encoder_name} not available") from e
+        valid_encoders = ', '.join(ENCODERS.keys())
+        raise OSError(f"encoder {encoder_name} not available. Valid encoders: {valid_encoders}") from e
     return encoder(mode, *args + extra)
 
 
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -400,8 +400,10 @@ def _getdecoder(mode, decoder_name, args, extra=()):
         # get decoder
         decoder = getattr(core, decoder_name + "_decoder")
     except AttributeError as e:
-        valid_decoders = ', '.join(DECODERS.keys())
-        raise OSError(f"decoder {decoder_name} not available. Valid decoders: {valid_decoders}") from e
+        valid_decoders = ", ".join(DECODERS.keys())
+        raise OSError(
+            f"decoder {decoder_name} not available. Valid decoders: {valid_decoders}"
+        ) from e
     return decoder(mode, *args + extra)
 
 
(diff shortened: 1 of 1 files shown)


boxed:
Researching more it looks like the entire code paths for DECODERS/ENCODERS/register_encoder/register_decoders and Image.to_bytes is just some unused code that doesn't work anymore? ENCODERS doesn't seem like it can ever be filled with "jpeg" for example, or anything else...

radarhere:
#6069 added complete support for `register_encoder` in Pillow 9.1.0 - https://pillow.readthedocs.io/en/stable/releasenotes/9.1.0.html#added-pyencoder. So that's not code that doesn't work anymore, it's new code.
It is called from BlpImagePlugin
https://github.com/python-pillow/Pillow/blob/6bb408c9622617a6d37d1746716721004188f743/src/PIL/BlpImagePlugin.py#L484

`register_decoder` is called from BlpImagePlugin, MspImagePlugin, SgiImagePlugin and PpmImagePlugin.

`tobytes()` is part of the public API - https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.tobytes - so if you say it is unused, that doesn't make sense to me.

radarhere:
So `DECODERS` and `ENCODERS` are for the Python codecs.

Within `_getencoder`, you will see that if it fails to find a Python codec in `ENCODERS`, it moves on to check the C codecs - https://github.com/python-pillow/Pillow/blob/6bb408c9622617a6d37d1746716721004188f743/src/PIL/Image.py#L424

Here is a list of the C codecs -
https://github.com/python-pillow/Pillow/blob/6bb408c9622617a6d37d1746716721004188f743/src/_imaging.c#L4025-L4061

So your suggested error message would only report the Python codecs, and ignore the possible C codecs that could be used.

boxed:
Hmm. Yea I see your point. I guess a good error message would then also have to do something like

```py
c_encoders = [x for x in dir(core) if x.endswith('_encoder')]
```
And join ENCODERS with this c_encoders before producing the error message.

radarhere:
This may just be my opinion, but doesn't a list of up to 21 different possible decoders make for a pretty long error message?

boxed:
Why is the length relevant? If it contains information that lets you solve the problem you are happy. If it does not, then you are unhappy.

I gave up on using this API totally because I couldn't figure out a valid encoding name. I used `image.save(f, format='jpeg')` instead because that worked.

radarhere:
> I gave up on using this API totally because I couldn't figure out a valid encoding name. I used `image.save(f, format='jpeg')` instead because that worked.

While I don't know exactly what your goal was, that may have been the actual solution to your problem. The [documentation for `tobytes()`](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.tobytes) states that
> This method returns the raw image data from the internal storage. For compressed image data (e.g. PNG, JPEG) use [save()](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.save), with a BytesIO parameter for in-memory data.

Also, just to point this out, if I try

```python
from PIL import Image
im = Image.new("RGB", (1, 1))
im.tobytes("jpg")
```

I get

```
  File "PIL/Image.py", line 455, in _getencoder
    encoder = getattr(core, encoder_name + "_encoder")
AttributeError: module 'PIL._imaging' has no attribute 'jpg_encoder'. Did you mean: 'jpeg_encoder'?

The above exception was the direct cause of the following exception:

  File "demo.py", line 3, in <module>
    im.tobytes("jpg")
  File "PIL/Image.py", line 768, in tobytes
    e = _getencoder(self.mode, encoder_name, args)
  File "PIL/Image.py", line 457, in _getencoder
    raise OSError(f"encoder {encoder_name} not available") from e
OSError: encoder jpg not available
```

Which is something of a helpful error message already.

boxed:
It's not a helpful message at all. The computer knows what the right answers is, why doesn't it tell me? Computer error messages can be better than "if you don't know what's wrong, I'm not going to tell you"!

radarhere:
I've created #6415 as an alternative to this, adding notes to the documentation on where to find other decoders.

PR #6313, opened 2022-05-20
Improved image file formats documentation

radarhere:
Addresses concerns raised in https://github.com/python-pillow/Pillow/discussions/6273#discussioncomment-2694757

1. https://pillow.readthedocs.io/en/latest/handbook/image-file-formats.html#gif
> The library writes LZW encoded files in GIF87a by default, unless GIF89a features are used or GIF89a is already in use.

This could be interpreted to mean that LZW encoding is not used for GIF89a. This is not the case (and @raygard would know, as he actually wrote #5291).

2. The ["Reading local images"](https://pillow.readthedocs.io/en/latest/handbook/image-file-formats.html#reading-local-images) example relies on the ability of the user to set `im.size`. This was removed in #3203, so I've updated the example to use `crop()` instead.

3. https://pillow.readthedocs.io/en/latest/handbook/image-file-formats.html
> So if [copy()](https://pillow.readthedocs.io/en/latest/reference/Image.html#PIL.Image.Image.copy) is called on an image, or another method internally creates a copy of the image, the `fp` (file pointer), along with any methods and attributes specific to a format. The [format](https://pillow.readthedocs.io/en/latest/reference/Image.html#PIL.Image.Image.format) attribute will be None.

The first sentence seems to just stop, so I've reworked it.

### DIFF
--- a/docs/handbook/image-file-formats.rst
+++ b/docs/handbook/image-file-formats.rst
@@ -101,8 +101,8 @@ GIF
 ^^^
 
 Pillow reads GIF87a and GIF89a versions of the GIF file format. The library
-writes LZW encoded files in GIF87a by default, unless GIF89a features
-are used or GIF89a is already in use.
+writes files in GIF87a by default, unless GIF89a features are used or GIF89a is
+already in use. Files are written with LZW encoding.
 
 GIF files are initially read as grayscale (``L``) or palette mode (``P``)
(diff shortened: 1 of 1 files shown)


raygard:
Thanks! This looks much better. The "read local image" example works now. I didn't realize you could do it like that; it may come in handy.

PR #6326, opened 2022-05-24
Decrement reference count

### LINKED ISSUE #6319: Memory leak in function `fname2char` due to new reference is not decreased (static analyzer report)
1. A new reference is returned and assigned to `bytes`.
https://github.com/python-pillow/Pillow/blob/68e39cba431680105c0e0f1a54371b7028e8da9a/src/Tk/tkImaging.c#L371

2. Function returns without decreasing the refcnt of `bytes`.
https://github.com/python-pillow/Pillow/blob/68e39cba431680105c0e0f1a54371b7028e8da9a/src/Tk/tkImaging.c#L375

Internal Report ID: c6d448

radarhere:
Resolves #6319

### DIFF
--- a/src/Tk/tkImaging.c
+++ b/src/Tk/tkImaging.c
@@ -372,7 +372,9 @@ fname2char(PyObject *fname) {
     if (bytes == NULL) {
         return NULL;
     }
-    return PyBytes_AsString(bytes);
+    char* libname = PyBytes_AsString(bytes);
+    Py_DECREF(bytes);
+    return libname;
 }
 
 #include <dlfcn.h>


nulano:
I believe this change will cause undefined behaviour by reading deallocated memory. 

From https://docs.python.org/3/c-api/bytes.html#c.PyBytes_AsString (emphasis mine):

> char *PyBytes_AsString([PyObject](https://docs.python.org/3/c-api/structures.html#c.PyObject) *o)
> Part of the [Stable ABI](https://docs.python.org/3/c-api/stable.html#stable).
> Return a pointer to the contents of o. **The pointer refers to the internal buffer of o**, which consists of len(o) + 1 bytes. The last byte in the buffer is always null, regardless of whether there are any other null bytes. The data must not be modified in any way, unless the object was just created using PyBytes_FromStringAndSize(NULL, size). **It must not be deallocated.** If o is not a bytes object at all, [PyBytes_AsString()](https://docs.python.org/3/c-api/bytes.html#c.PyBytes_AsString) returns NULL and raises [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError).

My understanding of this is that the returned buffer will be freed with the decref call. 

This would not be picked up by the Valgrind CI run as it is missing a display:

> Tests/test_imagetk.py::test_kw SKIPPED (TCL Error: no display name and no $DISPLAY environment variable) [ 63%]
Tests/test_imagetk.py::test_photoimage SKIPPED (TCL Error: no display name and no $DISPLAY environment variable) [ 63%]
Tests/test_imagetk.py::test_photoimage_blank SKIPPED (TCL Error: no display name and no $DISPLAY environment variable) [ 63%]
Tests/test_imagetk.py::test_box_deprecation SKIPPED (TCL Error: no display name and no $DISPLAY environment variable) [ 63%]
Tests/test_imagetk.py::test_bitmapimage SKIPPED (TCL Error: no display name and no $DISPLAY environment variable) [ 63%]

radarhere:
Thanks

PR #6329, opened 2022-05-24
Inline fname2char to fix memory leak

### LINKED ISSUE #6319: Memory leak in function `fname2char` due to new reference is not decreased (static analyzer report)

nulano:
Fixes #6319.

### DIFF
--- a/src/Tk/tkImaging.c
+++ b/src/Tk/tkImaging.c
@@ -364,17 +364,6 @@ load_tkinter_funcs(void) {
  * tkinter dynamic library (module).
  */
 
-/* From module __file__ attribute to char *string for dlopen. */
-char *
-fname2char(PyObject *fname) {
-    PyObject *bytes;
-    bytes = PyUnicode_EncodeFSDefault(fname);
-    if (bytes == NULL) {
-        return NULL;
-    }
-    return PyBytes_AsString(bytes);
-}
-
 #include <dlfcn.h>
 
 void *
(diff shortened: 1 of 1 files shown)


nulano:
I'm not sure what happened with valgrind, but there are no errors reported for the modified file.
```
==328== More than 10000000 total errors detected.  I'm not reporting any more.
==328== Final error counts will be inaccurate.  Go fix your program!
==328== Rerun with --error-limit=no to disable this cutoff.  Note
==328== that errors may occur in your program without prior warning from
==328== Valgrind, because errors are no longer being displayed.
```
The two failed tests are clearly unrelated:
```
FAILED Tests/test_file_pdf.py::test_redos[\r] - Failed: Timeout >1.0s
FAILED Tests/test_file_pdf.py::test_redos[\n] - Failed: Timeout >1.0s
```

wiredfool:
I'm not surprised that the redos ones might have failed, because the timeouts are probably calibrated for an ordinary run, not the 10x slower valgrind run.

radarhere:
I've rerun valgrind and it is passing now.

PR #6335, opened 2022-05-26
Once Exif data is parsed, do not reload unless it changes

### LINKED ISSUE #6334: TIFF EXIF data is not updated when saving image
### What did you do?
Tried to add 'Artist' and 'Copyright' exif fields to artwork.tiff
Exif to be saved and viewable for artwork.tiff.
### What actually happened?
Exif data was not updated and loading the image with Pillow again also didn't show the updated data.

* OS: Linux
* Python: 3.10.4
* Pillow: 9.1.1

```python
img = Image.open("artwork.tiff")
exif = img.getexif()
exif[0x8298] = "test"
exif[0x013B] = "test"
img.save("artwork.tiff", exif=exif)
```

[artwork.zip](https://github.com/python-pillow/Pillow/files/8775245/artwork.zip)
The tiff in the zip above was created by converting the svg I will attach below using ImageMagick.
[artwork-00.svg](https://user-images.githubusercontent.com/8126341/170388737-9f2c3100-f3b9-4342-b483-79eee8a6f226.svg)

Pillow saved the image successfully, no failure messages were shown.

radarhere:
Resolves #6334

The issue features a user calling `im.getexif()`, modifying the exif data, and then saving to a TIFF file.

However, when saving TIFF [makes sure that the exif data has been read with `getexif()`](https://github.com/python-pillow/Pillow/blob/c34d78934fac390409db169690e5b8b35a654ad1/src/PIL/TiffImagePlugin.py#L1208), which then [loads the data with `Exif.load_from_fp`](https://github.com/python-pillow/Pillow/blob/c34d78934fac390409db169690e5b8b35a654ad1/src/PIL/Image.py#L1396). This reloads the exif data, overriding the user's changes.

The solution is to prevent the exif data from being loaded again if it has already been parsed.

While `Exif.load()` [checks that it is not reloading the same input](https://github.com/python-pillow/Pillow/blob/c34d78934fac390409db169690e5b8b35a654ad1/src/PIL/Image.py#L3472-L3474), this is not done for `Exif.load_from_fp()`. To add that check specifically for `load_from_fp` would require storing the file pointer to ensure that it is the same file being read, and storing the file pointer would mean that it needs to be considered as part of the file handling lifecycle.

Instead, this PR adds a `_loaded` attribute. Once set, `getexif()` does not load the exif data anymore, but just returns the existing object. The attribute is cleared when the exif data is changed by seek operations so that the updated exif data can be loaded.


gverm:
@radarhere Just tested this PR and it also fixes the issue.

bigcat88:
@radarhere in my plugin `pillow-heif` i got a problem with new code.
there are separate exif info for each frame in HEIF, and with this changes, when changing to next frame `exif` thinks that it is already loaded.

```python3
        if self._exif is None:
            self._exif = Exif()
            self._exif._loaded = False
        elif self._exif._loaded:               # <--- this is always true when use ImageSequence.Iterator in Pillow 9.2
            return self._exif
        self._exif._loaded = True
```

should I reset in plugin `self._exif._loaded` this value when changing frame throw `seek` operation?

radarhere:
Hi. You could set `self._exif._loaded = False` in the plugin, sure.

Without knowing exactly what your code does, I might alternatively suggest calling `self._reload_exif()`, so that if the user has previously called `getexif()`, that object will be updated at the same time.

PR #6337, opened 2022-05-27
Fix "P"->"PA" image conversion 

### LINKED ISSUE #6336: Mismatch alpha channel when converting "P" to "PA" compared to "P" to "RGBA"
### What did you do?
I opened a PNG "P" image and converted it to "PA" and "RGBA".

The alpha channel for every pixel is the same for the "PA" and "RGBA" conversion

### What actually happened?
The alpha channel in the "PA" image is different from the "RGBA" image 

* OS: Microsoft Windows 10.0.19043 Pro
* Python: 3.10.4
* Pillow: 9.1.1

```python
def output(bytes):
    return ''.join(['\\x%02x' % b for b in bytes])

assert Image.open("tiny.png").mode == "P"

img_PA = Image.open("tiny.png").convert("PA")
img_RGBA = Image.open("tiny.png").convert("RGBA")

print(output(img_PA.tobytes()[slice(1, None, 2)]))   # \x00\x00\xff\xff\x00\x00\xd0\xb7\x00\x00\x38\x00\x00\x00\x15\xc0
print(output(img_RGBA.tobytes()[slice(3, None, 4)])) # \x00\x00\xff\xff\x00\x00\xff\xff\x00\x00\xff\xff\x00\x00\xff\xff

```
The tiny image used:
tiny

RedShy:
Resolves #6336

I fixed the function `p2pa()` in the file `Convert.c` by multiplying `in[0]` with 4 like other similar conversion functions do in the same source file.
I also added a test to cover the conversion "P" -> "PA".


Lucas-C:
Great, precise fix!

Looks legitimate to me.

Have you checked that the transparency of the resulting image can be seen with some image viewer ? (a web browser or your OS default one)

RedShy:
> Have you checked that the transparency of the resulting image can be seen with some image viewer ? (a web browser or your OS default one)

Yes. I ran this code with and without the fix I added
```python
Image.open("black.png").convert("PA").convert("RGBA").save("converted.png")
```
The original image is half black, half transparent
orig_black
Without the fix, the resulting image is all transparent:
old_black
With the fix the image is like the original: half transparent and half black
new_black

radarhere:
Some history.

#3728 added `p2pa` with `in[0] * 4`.
During the release of Pillow 6.0.0 though, this function [caused a problem](https://github.com/python-pillow/Pillow/issues/3618#issuecomment-478677444). Part of the fix was to [change it to `in[0]`](https://github.com/python-pillow/Pillow/commit/4f3b44960f103bc201a742332dbf9a77de1eaa87). I made that change without being able to iteratively test the fix myself though, so it is believable that that line did not actually need changing to fix Pillow 6.0.0.

radarhere (approved):

PR #6352, opened 2022-06-06
Added apply_transparency()

### LINKED ISSUE #6348: Transparency is not applied to RGBA palette when opening PNG
### What did you do?
I need to get a palette with transparency information from a PNG image. When saving this is correctly applied to the output image, when loading only an opaque palette is returned.

I expect that image.info["transparency"] is applied to a returned RGBA palette, i.e. alpha cleared for a single transparent colour, or alpha replaced with the array of transparency values.

### What actually happened?
A completely opaque palette is returned

* OS: Windows 10
* Python: 3.9.5 (64 bit)
* Pillow: 9.1.1

```python
from PIL import Image

img = Image.new("P", (256, 256))
pixel = []
for y in range(256):
    for x in range(256):
        v = ((y // 16) & 1) ^ ((x//16) & 1)
        pixel.append(v)
img.putdata(pixel)
pal = []
for i in range(256):
    if i == 0:
        pal.extend([0,0,0])
    else:
        pal.extend([255,255,255])
img.putpalette(pal, "RGB")

img.save("test_A.png", transparency=0)
img.save("test_B.png", transparency=bytes([0] + [255] * 255))

pal = []
for i in range(256):
    if i == 0:
        pal.extend([0,0,0,0])
    else:
        pal.extend([255,255,255,255])
img.putpalette(pal, "RGBA")
img.save("test_C.png")

for f in ["test_A.png", "test_B.png", "test_C.png"]:
    img = Image.open(f)
    pal = img.getpalette("RGBA")
    print (img.info, img.mode, pal[0:8])
```
This sets the transparency in 3 different ways and saves images to the current directory, then loads them back. 
The images show the expected transparency if opened in a program like GIMP.

Output is:
```
{'transparency': 0} P [0, 0, 0, 255, 255, 255, 255, 255]
{'transparency': 0} P [0, 0, 0, 255, 255, 255, 255, 255]
{'transparency': 0} P [0, 0, 0, 255, 255, 255, 255, 255]
```
Expected output should be something like:
```
{'transparency': 0} P [0, 0, 0, 0, 255, 255, 255, 255]
{'transparency': 0} P [0, 0, 0, 0, 255, 255, 255, 255]
{'transparency': 0} P [0, 0, 0, 0, 255, 255, 255, 255]
```

radarhere:
Resolves #6348

Adds `apply_transparency()`, a method to take a P mode image with a "transparency" key in the `im.info` dictionary, remove the "transparency" key and apply the transparency to the palette instead.

### DIFF
--- a/docs/reference/Image.rst
+++ b/docs/reference/Image.rst
@@ -123,6 +123,7 @@ methods. Unless otherwise stated, all methods return a new instance of the
 
 
 .. automethod:: PIL.Image.Image.alpha_composite
+.. automethod:: PIL.Image.Image.apply_transparency
 .. automethod:: PIL.Image.Image.convert
 
 The following example converts an RGB image (linearly calibrated according to
--- a/Tests/test_image.py
+++ b/Tests/test_image.py
@@ -840,6 +840,35 @@ class TestImage:
         im = Image.new("RGB", size)
         assert im.tobytes() == b""
 
+    def test_apply_transparency(self):
+        im = Image.new("P", (1, 1))
+        im.putpalette((0, 0, 0, 1, 1, 1))
+        assert im.palette.colors == {(0, 0, 0): 0, (1, 1, 1): 1}
+
+        # Test that no transformation is applied without transparency
+        im.apply_transparency()
+        assert im.palette.colors == {(0, 0, 0): 0, (1, 1, 1): 1}
+
+        # Test that a transparency index is applied
+        im.info["transparency"] = 0
+        im.apply_transparency()
+        assert "transparency" not in im.info
+        assert im.palette.colors == {(0, 0, 0, 0): 0, (1, 1, 1, 255): 1}
+
+        # Test that existing transparency is kept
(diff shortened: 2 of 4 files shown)


hugovk:
Thanks, for a new API, let's add this to docs (an `automethod` in `docs/reference/Image.rst`?) and release notes.

radarhere:
Ok, done

docs/reference/Image.rst
```diff
@@ -202,6 +202,7 @@ This helps to get the bounding box coordinates of the input image:
 .. automethod:: PIL.Image.Image.getexif
 .. automethod:: PIL.Image.Image.getextrema
 .. automethod:: PIL.Image.Image.getpalette
+.. automethod:: PIL.Image.Image.apply_transparency
```
hugovk, line 205: The rest are alphabetical, shall we put this up near line 125?

hugovk:
Thanks!

PR #6354, opened 2022-06-08
Revert "Skip test_realloc_overflow unless libtiff 4.0.4 or higher"

radarhere:
Reverts https://github.com/python-pillow/Pillow/commit/1a2866


radarhere:
The initial commit was added to workaround https://alas.aws.amazon.com/cve/html/CVE-2022-34266.html. The problem was fixed in [libtiff-4.0.3-35.amzn2.0.2](https://alas.aws.amazon.com/AL2/ALAS-2022-1814.html), so the workaround was able to be reverted.

Issue #6312, opened 2022-05-19
Not possible to take screenshot under Wayland


KhazAkar:
### What did you do?
I've tried to make screenshot using Pillow on Wayland compositor, such as GNOME running in Wayland mode
### What did you expect to happen?
Screenshot of full display
### What actually happened?
```
Traceback (most recent call last):
  Python Shell, prompt 2, line 1
    # Used internally for debug sandbox under external interpreter
  File "/home/khazakar/Project/Test/test_venv/lib/python3.8/site-packages/PIL/ImageGrab.py", line 68, in grab
    size, data = Image.core.grabscreen_x11(xdisplay)
builtins.OSError: X get_image failed: error 8 (73, 0, 1315)
```
### What are your OS, Python and Pillow versions?

* OS: Ubuntu 20.04, GNOME Wayland
* Python: 3.8.10
* Pillow:  9.1.0

```python
from PIL import ImageGrab
ImageGrab.grab()
```

Is there wayland support on roadmap anytime soon?

nulano:
> Is there wayland support on roadmap anytime soon?

AFAIK no one is working on this, contributions are welcome.

From https://stackoverflow.com/q/72216896/1648883, it sounds like `gnome-screenshot` is one of few utilities that can take a screenshot on Wayland. From a cursory look, https://gitlab.gnome.org/GNOME/gnome-screenshot/-/blob/master/src/screenshot-backend-shell.c appears to be the implementation.

radarhere:
I don't suppose it would be sufficient to just check for [`gnome-screenshot`](https://linux.die.net/man/1/gnome-screenshot) and call that, in the same way that ImageGrab [calls `screencapture` on macOS](https://github.com/python-pillow/Pillow/blob/30a0e448c181a81b435b77655d2c86549fbb4fec/src/PIL/ImageGrab.py#L30-L37)?

KhazAkar:
> I don't suppose it would be sufficient to just check for [`gnome-screenshot`](https://linux.die.net/man/1/gnome-screenshot) and call that, in the same way that ImageGrab [calls `screencapture` on macOS](https://github.com/python-pillow/Pillow/blob/30a0e448c181a81b435b77655d2c86549fbb4fec/src/PIL/ImageGrab.py#L30-L37)?

Not so much, since you can don't have GNOME stack running and all and you have more than 1 wayland-enabled DE :)

nulano:
[...]

PR #6361, opened 2022-06-12
Use gnome-screenshot on Linux if available

radarhere:
Helps #6312

On Linux, if XCB is not available and `gnome-screenshot` can be found, use `gnome-screenshot` to take a screenshot.

### DIFF
--- a/src/PIL/ImageGrab.py
+++ b/src/PIL/ImageGrab.py
@@ -15,14 +15,13 @@
 # See the README file for information on usage and redistribution.
 #
 
+import os
+import shutil
+import subprocess
 import sys
+import tempfile
 
 from . import Image
-
-if sys.platform == "darwin":
-    import os
-    import subprocess
-    import tempfile
 
 
 def grab(bbox=None, include_layered_windows=False, all_screens=False, xdisplay=None):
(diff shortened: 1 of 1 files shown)


src/PIL/ImageGrab.py
```diff
@@ -62,6 +61,18 @@ def grab(bbox=None, include_layered_windows=False, all_screens=False, xdisplay=N
                 left, top, right, bottom = bbox
                 im = im.crop((left - x0, top - y0, right - x0, bottom - y0))
             return im
+        elif not Image.core.HAVE_XCB and shutil.which("gnome-screenshot"):
```
nulano, line 64: I'm not sure `HAVE_XCB` should be required here. This is only set if Pillow is compiled with XCB, but it says nothing about whether the display supports screenshots. Anyone installing Plilow from a wheel would be unable to use gnome-screenshot. Linux users can still force XCB to be used even with gnome-screenshot present by passing `xdisplay=''`. I think release notes/documentation should be sufficient here.

```suggestion
        elif shutil.which("gnome-screenshot"):
```

radarhere: Ok, I've added a commit for documentation.

Issue #3900, opened 2019-06-12
FPX fill decoder not available


faheemziker:
### What did you do?

opened a fpx file using FpxImagePlugin and tried to save the same file as jpg.
the file is compressed fpx file.

```python
Image.open("/opt/project/test.fpx")
im.save("/opt/project/fpx_jpg.jpg")
```

### What did you expect to happen?
it should save without any error

### What actually happened?
it gives me exception
```pycon
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/dist-packages/PIL/Image.py", line 1961, in save
    self.load()
  File "/usr/local/lib/python2.7/dist-packages/PIL/FpxImagePlugin.py", line 227, in load
    return ImageFile.ImageFile.load(self)
  File "/usr/local/lib/python2.7/dist-packages/PIL/ImageFile.py", line 213, in load
    args, self.decoderconfig)
  File "/usr/local/lib/python2.7/dist-packages/PIL/Image.py", line 457, in _getdecoder
    raise IOError("decoder %s not available" % decoder_name)
IOError: decoder fill not available
```

### What are your OS, Python and Pillow versions?

* OS: ubuntu
* Python: 2.7
* Pillow: latest

Test file
https://github.com/python-pillow/Pillow/files/3281304/test.fpx.zip

```python
Image.open("/opt/project/test.fpx")
im.save("/opt/project/fpx_jpg.jpg")
```

radarhere:
You have hit https://github.com/python-pillow/Pillow/blob/a986fed5b445bc9586476167b08d46e19cba1bbc/src/PIL/FpxImagePlugin.py#L175

faheemziker:
How can we write fill decoder implementation any hint ??

radarhere:
As far as hints go, this looks like 4.2.2.2 of http://graphcomp.com/info/specs/livepicture/fpx.pdf. For information on Pillow decoders, see https://pillow.readthedocs.io/en/stable/handbook/writing-your-own-file-decoder.html#writing-your-own-file-decoder-in-c

radarhere:
If a fix is created, is the image that you attached able to be included as a test image under the Pillow license?

faheemziker:
i need to ask my our project owner but i think there should be no issue but lets wait for confirmation

faheemziker:
i am also trying its fix for placing single color tile in case of fill. but still having issues. can you ping here if you have the fix

[...]

PR #6372, opened 2022-06-16
Added FPX fill decoder

radarhere:
Helps #3900

The FPX image in the issue uses ["fill" compression](https://github.com/python-pillow/Pillow/blob/b3bc4a7bb4e7ab007af7d19b5091b6d4ee3c13a4/src/PIL/FpxImagePlugin.py#L172-L174), or as it is known in the [specification](https://graphcomp.com/info/specs/livepicture/fpx.pdf), "single color compression".

The specification lays out that "single color compression" is an apt name.
> If the compression type is set to single color compression (0x1), the compression subtype identifies the actual color value of all pixels in the tile.

This PR adds a PyDecoder to fill the tile with the color.


faheemziker:
i have already fixed this issue ill create a pr in which we can see expected output

radarhere:
@faheemziker thanks, that would be good. If you have an image that could be added to our test suite and distributed under the Pillow license, that would also be useful

radarhere:
@faheemziker any update on that PR?

radarhere:
Closing, as apparently this does not produce the correct output.

PR #6378, opened 2022-06-19
Improved GIF optimize condition

### LINKED ISSUE #6362: GIF: _get_optimize() missed opportunities
### What did you do?

Copied a GIF using a recent 9.2.0dev0 build:

```python
from PIL import Image
with Image.open('stripes.gif') as im:
    im.save('out.gif', save_all=True, optimize=True, disposal=1)
```
stripes

The output file would have optimized local color tables.

### What actually happened?

All the local color tables had 256 colors where 8 colors would suffice.

* OS: Windows 10 Pro
* Python: 3.10.4
* Pillow: 9.2.0dev0

#### Discussion

`stripes.gif` is similar in every way that matters to the `aniship.gif` image I used in issue #6260. I created it so it could be freely distributed as desired. I realize it's unusual. It has no GCT and the LCT is the same on every frame. But `aniship.gif` was found "in the wild" so `stripes.gif` not an entirely artificial example.

In `GifImagePlugin.py`, `_get_optimize()` uses certain criteria to decide if the palette is worth trying to optimize (remapping to only the used colors). The optimization is done only if the image is smaller than 262144 pixels, there are no more than 128 used colors, and there are "holes" in the palette, the latter being the case if `max(used_palette_colors) > len(used_palette_colors)`.

I think the "holes" test should be `max(used_palette_colors) >= len(used_palette_colors)`, as e.g. [1, 2, 3, 4, 5] has a hole at the bottom and [0, 1, 2, 4, 5] has a hole in the middle, but would be seen now as having no hole.

But I'm not clear on the point regarding holes. I would think the condition for optimizing would be if the palette actually got smaller. The number of colors in the current palette would (I think) be a power of two, so you would want to check if the new (used colors) palette is not more than half the size of the current palette. I'm sure I must be missing something here, but how about getting the current number of colors and testing if the number of colors used is not more than half of that. For example:

```python
            num_palette_colors = len(im.palette.palette) // 4 if len(im.palette.palette) % 3 else len(im.palette.palette) // 3
            if optimise or (
                len(used_palette_colors) <= 128
                and (max(used_palette_colors) >= len(used_palette_colors)
                or len(used_palette_colors) <= num_palette_colors // 2)
            ):
```

This assumes that im.palette.palette exists, is not None and is actually a byte string or array of RGB or RGBA colors. If its length is not a multiple of 3 I assume it's RGBA because they would always be a multiple of 4 in size. Are these valid assumptions? If not, is there something we can do along these lines?

Would this also obviate the need for `len(used_palette_colors) <= 128`?

With this change to `_get_optimize()`, The `out.gif` is reduced from 36653 bytes to 22517 bytes.

If this change is worth pursuing and my logic is close to correct, I can create a PR for this. What have I missed?

raygard:
Palette can be optimized if number of colors can be reduced by half or more.

Fixes #6362 .

 * Fix test for palette "hole" by changing comparison `max(used_palette_colors) > len(used_palette_colors)` to use `>=`
 * Also allow optimize if number of used colors is not more than half the number of current colors
 * Added a test for the latter that fails with current `_get_optimize()` but passes with this change.
 
I note that this change should obviate the condition `len(used_palette_colors) <= 128` but some tests fail if I remove it. I would rather leave it than try to update those tests in `test_file_gif.py`.
As noted in #6362, this assumes that im.palette.palette exists, is not None and is actually a byte string or array of RGB or RGBA colors. If this is possibly not the case please let me know!

### DIFF
--- a/src/PIL/GifImagePlugin.py
+++ b/src/PIL/GifImagePlugin.py
@@ -824,9 +824,14 @@ def _get_optimize(im, info):
                 if count:
                     used_palette_colors.append(i)
 
+            num_palette_colors = len(im.palette.palette) // 4 if im.palette.mode == 'RGBA' else len(im.palette.palette) // 3
+            # Round up to power of 2 but at least 4
+            num_palette_colors = max(4, 1 << (num_palette_colors - 1).bit_length())
             if optimise or (
+
                 len(used_palette_colors) <= 128
-                and max(used_palette_colors) > len(used_palette_colors)
+                and max(used_palette_colors) >= len(used_palette_colors)
+                or len(used_palette_colors) <= num_palette_colors // 2
             ):
                 return used_palette_colors
 
--- a/src/PIL/GifImagePlugin.py
+++ b/src/PIL/GifImagePlugin.py
@@ -824,11 +824,14 @@ def _get_optimize(im, info):
                 if count:
                     used_palette_colors.append(i)
 
-            num_palette_colors = len(im.palette.palette) // 4 if im.palette.mode == 'RGBA' else len(im.palette.palette) // 3
+            num_palette_colors = (
+                len(im.palette.palette) // 4
+                if im.palette.mode == "RGBA"
+                else len(im.palette.palette) // 3
+            )
             # Round up to power of 2 but at least 4
             num_palette_colors = max(4, 1 << (num_palette_colors - 1).bit_length())
             if optimise or (
-
                 len(used_palette_colors) <= 128
                 and max(used_palette_colors) >= len(used_palette_colors)
                 or len(used_palette_colors) <= num_palette_colors // 2
--- a/Tests/test_file_gif.py
+++ b/Tests/test_file_gif.py
@@ -180,6 +180,21 @@ def test_optimize_full_l():
     assert im.mode == "L"
 
 
+def test_optimize_if_palette_can_be_reduced_by_half():
+    with Image.open("Tests/images/test.colors.gif") as im:
+        # Reduce because original is too big for _get_optimize()
+        im = im.resize((591, 443))
+        imrgb = im.convert("RGB")
+        out = BytesIO()
+        imrgb.save(out, "GIF", optimize=False)
+        with Image.open(out) as reloaded:
+            assert len(reloaded.palette.palette) // 3 == 256
+        out = BytesIO()
+        imrgb.save(out, "GIF", optimize=True)
+        with Image.open(out) as reloaded:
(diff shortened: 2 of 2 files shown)


src/PIL/GifImagePlugin.py
```diff
@@ -824,9 +824,17 @@ def _get_optimize(im, info):
                 if count:
                     used_palette_colors.append(i)
 
+            num_palette_colors = (
+                len(im.palette.palette) // 4
+                if im.palette.mode == "RGBA"
+                else len(im.palette.palette) // 3
+            )
+            # Round up to power of 2 but at least 4
+            num_palette_colors = max(4, 1 << (num_palette_colors - 1).bit_length())
```
radarhere, line 833: Why at least 4?

raygard: Perhaps I didn't think it through carefully enough. The idea is that this condition would trigger only when the number of used colors is low enough to reduce the GIF color table. For example, if the palette now has 30 colors, the GIF would have 32 without optimization, so if the used colors are 16 or fewer, then it is worth remapping down. So I was thinking that the right side of the `len(used_palette_colors) <= num_palette_colors // 2` condition should never be less than 2 because that's the smallest size GIF color table.

Now I see that this could mean attempting to optimize if the actual palette is 1 or 2 colors and and only one is used (an odd case but not impossible), which is probably silly. Maybe it should instead have a test to prevent optimization if there are fewer than 3 colors in the palette? Reducing a 4-color palette to 2 is the minimum reasonable case for optimization. Maybe replace

```python
            # Round up to power of 2 but at least 4
            num_palette_colors = max(4, 1 << (num_palette_colors - 1).bit_length())
```
with something like

```python
            # Round up to power of 2 (0 -> 2 but that's OK)
            num_palette_colors = 1 << (num_palette_colors - 1).bit_length()
            # Suppress if less than 4
            num_palette_colors = 0 if num_palette_colors < 4 else num_palette_colors
```

Thoughts?

radarhere: I've added another commit to https://github.com/raygard/Pillow/pull/3 with my suggestion. See what you think

raygard: Looks good! I merged into my own repo and the changes apparently flow automatically into this PR. I'm still learning how github works...
Thank you for helping me get this working right and clean. I would not have been able to fix the tests myself.
Next up: I have mods about ready (I hope) to finally allow the "LZW minimum code size" (part 22 of the GIF spec) to be less than 8 in files written by Pillow.

PR #6386, opened 2022-06-20
Handle PCF fonts files with less than 256 characters

dawidcrivelli:
Fixes loading fonts where a lot of basic characters are missing. 

Here is an example [clR4x6.pcf.gz](https://github.com/python-pillow/Pillow/files/8941407/clR4x6.pcf.gz), from the X11 misc fixed-sized fonts, found on Ubuntu at `/usr/share/fonts/X11/misc/clR4x6.pcf.gz`


radarhere:
We like to have tests with our PRs, so that we can guard against this change being accidentally reverted in the future.

Is the font one that we can include in our test suite and distribute under Pillow's license?

dawidcrivelli:
I believe so, it's available from X.org: https://www.x.org/releases/individual/font/font-schumacher-misc-1.0.0.tar.gz , but I'm not an expert on open-source licenses.

The copyright notice says:
```
Copyright 1989 Dale Schumacher, dal@syntel.mn.org
               399 Beacon Ave.
               St. Paul, MN  55104-3527

Permission to use, copy, modify, and distribute this software and
its documentation for any purpose and without fee is hereby
granted, provided that the above copyright notice appear in all
copies and that both that copyright notice and this permission
notice appear in supporting documentation, and that the name of
Dale Schumacher not be used in advertising or publicity pertaining to
distribution of the software without specific, written prior
permission.  Dale Schumacher makes no representations about the
suitability of this software for any purpose.  It is provided "as
is" without express or implied warranty.
```

hugovk:
With whitespace adjustments, the font's licence on the left, MIT-CMU on the right:

image

radarhere:
I've pushed a commit that adds a test, using a hexedited version of https://github.com/python-pillow/Pillow/blob/main/Tests/fonts/10x20-ISO8859-1.pcf.

I've also changed this to no longer catch the exception, but instead to not look beyond the number of characters in the first place.

radarhere (approved):

PR #6394, opened 2022-06-24
Reverted to __array_interface__ with the release of NumPy 1.23

### LINKED ISSUE #5815: Enhancement: Change `Image.tobytes` return type to ByteArray or add a `Image.tobytearray` method
Currently the method `Image.tobytes` returns a byte string. There is nothing inherently bad about this, except that the returned bytes are _read only_ (ofc) since bytes are immutable.

The consequence of this is that `Image.__array__` does an extra copy when calling 

https://github.com/python-pillow/Pillow/blob/a3d1e2f85db890c6a1623e3902711863f55cc54e/src/PIL/Image.py#L678

that (I think) can be avoided. The data is copied from the immutable bytes string into a new mutable memory block around which the array is constructed.

This copy can be avoided in a few ways:

1. Chance the code in `Image.__array__` to (roughly) `np.frombuffer(self.tobytes(), dtype=dtype).reshape(shape)`. The downside is that the result is read-only (`result.flags['WRITEABLE'] == False`) which may be undesirable.
2. Change the code in `Image.tobytes` to return a `bytearray`. This may require work, depending on what `encoder` returns (didn't dive that far into the code), and it may be undesirable because type(Image.tobytes()) changes.
3. Introduce a new `Image.tobytearray` method that essentially mirrors `Image.tobytes`, except that it creates a mutable buffer. This resolves the problems that come with changing `Image.tobytes` return type, but may still need more work.

Would this be something interesting and/or useful to do?

radarhere:
Resolves #5815

#5379 was created to workaround a bug in NumPy. That bug has now been fixed in NumPy 1.23, so this PR reverts the src change from #5379.

### DIFF
--- a/Tests/test_image_array.py
+++ b/Tests/test_image_array.py
@@ -1,4 +1,5 @@
 import pytest
+from packaging.version import parse as parse_version
 
 from PIL import Image
 
@@ -34,9 +35,10 @@ def test_toarray():
     test_with_dtype(numpy.float64)
     test_with_dtype(numpy.uint8)
 
-    with Image.open("Tests/images/truncated_jpeg.jpg") as im_truncated:
-        with pytest.raises(OSError):
-            numpy.array(im_truncated)
+    if parse_version(numpy.__version__) >= parse_version("1.23"):
+        with Image.open("Tests/images/truncated_jpeg.jpg") as im_truncated:
+            with pytest.raises(OSError):
+                numpy.array(im_truncated)
 
 
 def test_fromarray():
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -671,14 +671,9 @@ class Image:
             raise ValueError("Could not save to PNG for display") from e
         return b.getvalue()
 
-    class _ArrayData:
-        def __init__(self, new):
-            self.__array_interface__ = new
-
-    def __array__(self, dtype=None):
+    @property
+    def __array_interface__(self):
         # numpy array interface support
-        import numpy as np
-
         new = {}
         shape, typestr = _conv_type_shape(self)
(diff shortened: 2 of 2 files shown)


radarhere:
The MinGW 32-bit GitHub Actions job has started failing not just here, but also [on main, with 'test_adobe_deflate_tiff'.](https://github.com/python-pillow/Pillow/runs/7050714055?check_suite_focus=true#step:6:580)

Comparing the failing job and the last passing job, [the failing job has libtiff-4.4.0-2](https://github.com/python-pillow/Pillow/runs/7050714055?check_suite_focus=true#step:4:65), while [the passing one has libtiff-4.4.0-1.](https://github.com/python-pillow/Pillow/runs/7018324483?check_suite_focus=true#step:4:65)

The package is actually up to [4.4.0-4 at present](https://packages.msys2.org/package/mingw-w64-i686-libtiff), and those [more recent changes mention libdeflate](https://github.com/msys2/MINGW-packages/commits/master/mingw-w64-libtiff). So I expect this will solve itself whenever the MinGW image is updated.

hugovk:
Shall we skip the bad test for now, so it doesn't interfere with next week's release?

nulano:
AFAIK the virtual environments seem to have roughly weekly releases, with the latest one being from the 19th (probably right before the libtiff rebuild). So I'd expect the error clear up before the release. Perhaps disable the test if it is still failing on Wed/Thu?

hugovk:
Fine by me, if failing builds aren't too distracting in the meantime :)

radarhere:
The MinGW 32-bit job has started passing again - https://github.com/python-pillow/Pillow/runs/7102756365

PR #6402, opened 2022-06-30
Added GIF decompression bomb check

radarhere:


ajakk:
So, should this get a CVE as a DoS vulnerability fix?

radarhere:
I don't think so. None of our other `_decompression_bomb_check` calls have warranted CVEs.

risicle:
It's happened :D

 CVE-2022-45198

PR #6411, opened 2022-07-03
Please, implement compatibility with pypy

Ad101DP:
When installing with pypy it throws an error, please fix it


hugovk:
Pillow is already compatible with PyPy: tested on PyPy and wheels are built for PyPy.

This PR has no affect on that.

Please open a detailed issue if you're having problems with Pillow on PyPy.

PR #6413, opened 2022-07-04
Set permissions for GitHub actions

neilnaveen:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: neilnaveen <42328488+neilnaveen@users.noreply.github.com>

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -2,6 +2,9 @@ name: Lint
 
 on: [push, pull_request, workflow_dispatch]
 
+permissions:
+  contents: read
+
 jobs:
   build:
 
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -7,8 +7,14 @@ on:
       - main
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 jobs:
   update_release_draft:
+    permissions:
+      contents: write  # for release-drafter/release-drafter to create a github release
+      pull-requests: write  # for release-drafter/release-drafter to add label to PR
     if: github.repository == 'python-pillow/Pillow'
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/test-cygwin.yml
+++ b/.github/workflows/test-cygwin.yml
@@ -99,6 +99,8 @@ jobs:
           name: Cygwin Python 3.${{ matrix.python-minor-version }}
 
   success:
+    permissions:
+      contents: none
     needs: build
     runs-on: ubuntu-latest
     name: Cygwin Test Successful
--- a/.github/workflows/test-docker.yml
+++ b/.github/workflows/test-docker.yml
@@ -2,6 +2,9 @@ name: Test Docker
 
 on: [push, pull_request, workflow_dispatch]
 
+permissions:
+  contents: read
+
 jobs:
   build:
 
@@ -82,6 +85,8 @@ jobs:
         name: ${{ matrix.docker }}
 
   success:
+    permissions:
+      contents: none
     needs: build
     runs-on: ubuntu-latest
     name: Docker Test Successful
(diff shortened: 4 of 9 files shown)


hugovk:
The Tidelift Align workflow is failing because it needs to read an API key from the repo secrets.

Please could you fix it?

neilnaveen:
> The Tidelift Align workflow is failing because it needs to read an API key from the repo secrets.
> 
> Please could you fix it?

At the moment I am busy, could you please cherrypick and merge this commit? Thanks.

hugovk:
There's no rush, whenever you're next free :)

radarhere:
@hugovk I don't think Tidelift Align is failing because of the changes made here. I think it's failing because it's running from a fork, [as has happened before.](https://github.com/python-pillow/Pillow/pull/6221#issuecomment-1104958958)

If I [take the changes made here](https://github.com/python-pillow/Pillow/commit/4dc128877c54c6942c80114c0c39ff374ee7b6ad) to the Tidelift Align workflow, and run it under python-pillow, [it passes.](https://github.com/python-pillow/Pillow/runs/7249405430?check_suite_focus=true)

hugovk:
Oh yes, this line skips running for forks (https://github.com/python-pillow/Pillow/commit/836f740f7b3b107e3b20e8317b514da79584622e):

https://github.com/python-pillow/Pillow/blob/fdf4084a7f8ea49bca25158930a8af80cf685b81/.github/workflows/tidelift.yml#L17

But not for PRs from forks to upstream.

It would be nice to only run for upstream pushes, or PRs from upstream forks (kind of like the inverse of https://github.com/has2k1/plotnine/blob/69a5289f625202411e2e543aad9c7f2cb9954fc7/.github/workflows/testing.yml#L10-L12), but we have enough guards so it only runs when certain files are changed, and this file is rarely changed, so it's probably not worth the bother.

Let's merge this then, thanks!

jayaddison:
Note: GitHub Actions recently updated the [default `GITHUB_TOKEN` token permissions to be read-only](https://github.blog/changelog/2023-02-02-github-actions-updating-the-default-github_token-permissions-to-read-only/), and that might allow simplifying some of the permissions in the workflow files here.

(ensuring that read-only permissions are in place makes sense; it seems ungreat to me that many projects had to expand their workflow definitions with additional vendor-specific configuration to gain better confidence about that though.  so if it's possible to clean some of that up and simplify back to a more portable configuration, that could be worthwhile)

radarhere:
Thanks for the update. I think there is still value here, as removing these might lower our 'Token-Permissions' score at https://deps.dev/pypi/pillow/, since they might not be aware of that.

jayaddison:
Ah, ok - good to know.  Hmm.. is `deps.dev` a significant factor in uptake/continued usage of `Pillow`?

radarhere:
Tidelift, who kindly support Pillow, would like us to do well in a few of the OpenSSF scorecard checks.

'Token-Permissions' isn't actually one of the checks that Tidelift mentions, but that's the origin of the thinking.

hugovk:
The new default only applies to new repos, we still have the read/write default:

> This change will not impact any existing enterprises, organizations or repositories.

jayaddison:
Ek, thanks @hugovk.  I should read more carefully :|

PR #6416, opened 2022-07-04
Fix when `sys.executable` is empty or null

bryant1410:
There are exceptional cases in which `sys.executable` is empty or null (see [the docs](https://docs.python.org/3/library/sys.html#sys.executable)). I think it's better to just not delete the image in these cases, instead of crashing.

### DIFF
--- a/src/PIL/ImageShow.py
+++ b/src/PIL/ImageShow.py
@@ -178,10 +178,11 @@ class MacViewer(Viewer):
             else:
                 raise TypeError("Missing required argument: 'path'")
         subprocess.call(["open", "-a", "Preview.app", path])
-        if sys.executable:
+        exectable = sys.executable or shutil.which("python3")
+        if executable:
             subprocess.Popen(
                 [
-                    sys.executable,
+                    executable,
                     "-c",
                     "import os, sys, time; time.sleep(20); os.remove(sys.argv[1])",
                     path,
--- a/src/PIL/ImageShow.py
+++ b/src/PIL/ImageShow.py
@@ -178,7 +178,7 @@ class MacViewer(Viewer):
             else:
                 raise TypeError("Missing required argument: 'path'")
         subprocess.call(["open", "-a", "Preview.app", path])
-        exectable = sys.executable or shutil.which("python3")
+        executable = sys.executable or shutil.which("python3")
         if executable:
             subprocess.Popen(
                 [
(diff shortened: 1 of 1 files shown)


radarhere:
Just curious - did you figure this out because you are in an environment where `sys.executable` is `None`? What type of environment is that?

bryant1410:
Yes. I'm not sure how much details I can share, but it's kind of a custom way of running Python where the final program is compiled, so there's no Python interpreter anymore (AFAIK).

What about calling `os.remove` directly?

radarhere:
The subprocess doesn't just call `os.remove`, it also calls `time.sleep(20)`. This behaviour started in #6010, so that the current Python process doesn't hang for 20 seconds. This period is intended to give time for Preview to open the image.

However, given that #6045 dropped removing the image altogether on Unix, the current form of this PR is not unreasonable.

radarhere:
I've created https://github.com/bryant1410/Pillow/pull/1 as a suggestion - if `sys.executable` is not available, how about falling back to `python3` instead?

bryant1410:
> I've created [bryant1410#1](https://github.com/bryant1410/Pillow/pull/1) as a suggestion - if `sys.executable` is not available, how about falling back to `python3` instead?

Sounds good to me! It works locally.

src/PIL/ImageShow.py
```diff
@@ -178,14 +178,16 @@ def show_file(self, path=None, **options):
...
         subprocess.call(["open", "-a", "Preview.app", path])
-        subprocess.Popen(
-            [
-                sys.executable,
-                "-c",
-                "import os, sys, time; time.sleep(20); os.remove(sys.argv[1])",
-                path,
-            ]
-        )
+        exectable = sys.executable or shutil.which("python3")
```
radarhere, line 181:
```suggestion
        exectuable = sys.executable or shutil.which("python3")
```

radarhere (approved):

beaux-riel:
I’m still experiencing the issue where the temporary file is deleted before Preview can open it on Mac. Has/will this be resolved?

radarhere:
This PR hasn't been released yet. It will be released as part of Pillow 9.3.0, due out towards the end of October.

However, it doesn't sound like you're experiencing the same problem. Are you saying that your images are taking longer than 20 seconds to open in Preview? Are they very large?

PR #6433, opened 2022-07-13
Docs: Replace version tables with RST csv-table and add 3.11

hugovk:
The existing version support tables at https://pillow.readthedocs.io/en/stable/installation.html#python-support are a bit time consuming to edit when adding a new version, for example 3.11:

```rst
+----------------------+-----+-----+-----+-----+-----+-----+-----+-----+
|        Python        |3.10 | 3.9 | 3.8 | 3.7 | 3.6 | 3.5 | 3.4 | 2.7 |
+======================+=====+=====+=====+=====+=====+=====+=====+=====+
| Pillow >= 9.0        | Yes | Yes | Yes | Yes |     |     |     |     |
+----------------------+-----+-----+-----+-----+-----+-----+-----+-----+
| Pillow 8.3.2 - 8.4   | Yes | Yes | Yes | Yes | Yes |     |     |     |
+----------------------+-----+-----+-----+-----+-----+-----+-----+-----+
| Pillow 8.0 - 8.3.1   |     | Yes | Yes | Yes | Yes |     |     |     |
+----------------------+-----+-----+-----+-----+-----+-----+-----+-----+
| Pillow 7.0 - 7.2     |     |     | Yes | Yes | Yes | Yes |     |     |
+----------------------+-----+-----+-----+-----+-----+-----+-----+-----+
| Pillow 6.2.1 - 6.2.2 |     |     | Yes | Yes | Yes | Yes |     | Yes |
+----------------------+-----+-----+-----+-----+-----+-----+-----+-----+
| Pillow 6.0 - 6.2.0   |     |     |     | Yes | Yes | Yes |     | Yes |
+----------------------+-----+-----+-----+-----+-----+-----+-----+-----+
| Pillow 5.2 - 5.4     |     |     |     | Yes | Yes | Yes | Yes | Yes |
+----------------------+-----+-----+-----+-----+-----+-----+-----+-----+

+------------------+-----+-----+-----+-----+-----+-----+-----+-----+-----+
|      Python      | 3.6 | 3.5 | 3.4 | 3.3 | 3.2 | 2.7 | 2.6 | 2.5 | 2.4 |
+==================+=====+=====+=====+=====+=====+=====+=====+=====+=====+
| Pillow 5.0 - 5.1 | Yes | Yes | Yes |     |     | Yes |     |     |     |
+------------------+-----+-----+-----+-----+-----+-----+-----+-----+-----+
| Pillow 4         | Yes | Yes | Yes | Yes |     | Yes |     |     |     |
+------------------+-----+-----+-----+-----+-----+-----+-----+-----+-----+
| Pillow 2 - 3     |     | Yes | Yes | Yes | Yes | Yes | Yes |     |     |
+------------------+-----+-----+-----+-----+-----+-----+-----+-----+-----+
| Pillow < 2       |     |     |     |     |     | Yes | Yes | Yes | Yes |
+------------------+-----+-----+-----+-----+-----+-----+-----+-----+-----+
```
image

One option would be to flip it so higher Python versions are on the right. But it still needs a bunch of character dancing, and would also be a bit tricky to edit when rebalancing the two tables - moving some rows from the top table to the bottom table.

# CSV table

Another option is to the `.. csv-table::` directive.

Docs:
* https://sublime-and-sphinx-guide.readthedocs.io/en/latest/tables.html#csv-files
* https://docutils.sourceforge.io/docs/ref/rst/directives.html#csv-table

We can then put CSV directly in the RST file, which means there's fewer structure characters that need juggling. For example (and using the comma to help display it more like a table when editing RST):

```rst
.. csv-table:: Newer versions [table]
   :widths: auto
   :header: Python               , 3.11, 3.10, 3.9 , 3.8 , 3.7 , 3.6 , 3.5 , 3.4 , 2.7

            Pillow >= 9.3        , Yes , Yes , Yes , Yes , Yes ,     ,     ,     ,
            Pillow 9.0 - 9.2     ,     , Yes , Yes , Yes , Yes ,     ,     ,     ,
            Pillow 8.3.2 - 8.4   ,     , Yes , Yes , Yes , Yes , Yes ,     ,     ,
            Pillow 8.0 - 8.3.1   ,     ,     , Yes , Yes , Yes , Yes ,     ,     ,
            Pillow 7.0 - 7.2     ,     ,     ,     , Yes , Yes , Yes , Yes ,     ,
            Pillow 6.2.1 - 6.2.2 ,     ,     ,     , Yes , Yes , Yes , Yes ,     , Yes
            Pillow 6.0 - 6.2.0   ,     ,     ,     ,     , Yes , Yes , Yes ,     , Yes
            Pillow 5.2 - 5.4     ,     ,     ,     ,     , Yes , Yes , Yes , Yes , Yes

.. csv-table:: Older versions [table]
   :widths: auto
   :header: Python           , 3.6 , 3.5 , 3.4 , 3.3 , 3.2 , 2.7 , 2.6 , 2.5 , 2.4

            Pillow 5.0 - 5.1 , Yes , Yes , Yes ,     ,     , Yes ,     ,     ,
            Pillow 4         , Yes , Yes , Yes , Yes ,     , Yes ,     ,     ,
            Pillow 2 - 3     ,     , Yes , Yes , Yes , Yes , Yes , Yes ,     ,
            Pillow < 2       ,     ,     ,     ,     ,     , Yes , Yes , Yes , Yes
```
image

This is still a bit of a pain to edit.

# CSV file

Or we can load them directly from CSV files:

```rst
.. csv-table:: Newer versions [CSV file]
   :file: newer-versions.csv
   :header-rows: 1
   :widths: auto

.. csv-table:: Older versions [CSV file]
   :file: older-versions.csv
   :header-rows: 1
   :widths: auto
```
```csv
Python,"3.10",3.9,3.8,3.7,3.6,3.5,3.4,2.7
Pillow >= 9.0,Yes,Yes,Yes,Yes,,,,
Pillow 8.3.2 - 8.4,Yes,Yes,Yes,Yes,Yes,,,
Pillow 8.0 - 8.3.1,,Yes,Yes,Yes,Yes,,,
Pillow 7.0 - 7.2,,,Yes,Yes,Yes,Yes,,
Pillow 6.2.1 - 6.2.2,,,Yes,Yes,Yes,Yes,,Yes
Pillow 6.0 - 6.2.0,,,,Yes,Yes,Yes,,Yes
Pillow 5.2 - 5.4,,,,Yes,Yes,Yes,Yes,Yes
```
```csv
Python,3.6,3.5,3.4,3.3,3.2,2.7,2.6,2.5,2.4
Pillow 5.0 - 5.1,Yes,Yes,Yes,,,Yes,,,
Pillow 4,Yes,Yes,Yes,Yes,,Yes,,,
Pillow 2 - 3,,Yes,Yes,Yes,Yes,Yes,Yes,,
Pillow < 2,,,,,,Yes,Yes,Yes,Yes
```

image

Then it's pretty easy to edit the CSV files in a spreadsheet.

One thing we'd need to watch out for: when opening `newer-versions.csv`, Excel has an annoying habit of converting 3.10 to 3.1, even if it's in the CSV as `"3.10"` or `'3.10`.

Putting this as a draft to compare the different options. Will delete the `[CSV file]`/`[table]` bits as well, they're just for visibility during testing.

These CSV tables are also render more compactly than the original, which is a nice plus.

Shall we switch to one of these?


hugovk:
Demo: https://pillow--6433.org.readthedocs.build/en/6433/installation.html#python-support

Note "Newer versions [table]" also has 3.11. I left that in by accident when testing out, will remove it from the final PR.

radarhere:
My suggestion would be a mix of the two, as the inline version still works without spaces. This would avoid the separate CSV files, while still keeping the easy of editing - https://github.com/radarhere/Pillow/commit/395469d6112b85c8639a8c1b1945414ad6227a95 / https://pillow-radarhere.readthedocs.io/en/docs-csv-table/installation.html#python-support

hugovk:
Yep, it also works without spaces. I think the spaces help make sure the raw text match the output, to make it easier to edit and see you're making the right changes.

But my idea for using separate CSV files is they can be much more easily manipulated using a spreadsheet program to make it easier to add/move rows, and between tables as well if necessary.

radarhere:
Ok. I would personally still just edit the raw text anyway, but it is easier for others, then sure.

hugovk:
Thanks, updated to use CSV files.

I also added Python 3.11 for Pillow >= 9.3, and then rebalanced the tables so 2.7 support is only in the older table.

# Before

image

# After

image

docs/installation.rst
```diff
@@ -15,35 +15,15 @@ Python Support
...
-+------------------+-----+-----+-----+-----+-----+-----+-----+-----+-----+
+.. csv-table:: Newer versions
+   :file: newer-versions.csv
+   :header-rows: 1
+   :widths: auto
+
+.. csv-table:: Older versions
+   :file: older-versions.csv
+   :header-rows: 1
+   :widths: auto
```
radarhere, line 26:
```suggestion

.. csv-table:: Older versions
   :file: older-versions.csv
   :header-rows: 1
```
We can leave it in if you prefer, but I don't think specifying the width is necessary

https://docutils.sourceforge.io/docs/ref/rst/directives.html#table
> "auto" delegates the determination of column widths to the backend (LaTeX, the HTML browser, ...). Default for the [html5 writer](https://docutils.sourceforge.io/docs/user/html.html#html5-polyglot)

https://pillow-radarhere.readthedocs.io/en/docs-csv-table/installation.html

PR #4662, opened 2020-06-01
Installed libxcb-xinerama0 to fix PyQt5

radarhere:
Resolves #4659. See the issue for my explanation of why this is the solution.

PR #6438, opened 2022-07-14
Install libxcb-shape0

radarhere:
Tests/test_qt_image_qapplication.py has started aborting in ubuntu-latest on main - https://github.com/python-pillow/Pillow/runs/7337854934?check_suite_focus=true#step:8:2131

Starting [QApplication with `QT_DEBUG_PLUGINS` enabled](https://github.com/radarhere/Pillow/commit/a540fb9d2c42a994a4d91fdcaab661cba45e67b0), I find the message ["libxcb-shape.so.0: cannot open shared object file: No such file or directory"](https://github.com/radarhere/Pillow/runs/7339328878?check_suite_focus=true#step:4:258)

So this PR installs libxcb-shape0.

A similar thing has happened before, with #4662

### DIFF
--- a/.ci/install.sh
+++ b/.ci/install.sh
@@ -43,7 +43,7 @@ if [[ $(uname) != CYGWIN* ]]; then
 
     # PyQt6 doesn't support PyPy3
     if [[ $GHA_PYTHON_VERSION == 3.* ]]; then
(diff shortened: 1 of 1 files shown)


hugovk:
Thanks, good find!

PR #6444, opened 2022-07-16
Support saving multiple MPO frames

radarhere:
Addresses #6373

Adds `save_all` and `append_images` to the MPO format.

I put this together by studying the way that Pillow reads in multiple MPO frames, and by looking at the [MPO specification.](https://web.archive.org/web/20190227081740/http://www.cipa.jp/std/documents/e/DC-007_E.pdf)

### DIFF
--- a/Tests/test_file_mpo.py
+++ b/Tests/test_file_mpo.py
@@ -5,15 +5,19 @@ import pytest
 
 from PIL import Image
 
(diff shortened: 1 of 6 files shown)


hugovk (approved):
Thanks, let's also add release notes for this one.

PR #6449, opened 2022-07-17
Added DDS BC6 reading

### LINKED ISSUE #6344: DDS BC6 decoding has a broken blue channel
I attempted to use the code in https://github.com/python-pillow/Pillow/blob/main/src/libImaging/BcnDecode.c for a project in order to decode BC6 textures. The red and green channels decoded correctly but the blue channel ended up being mangled somehow.

Here's the red channel of the image as displayed in renderdoc:
20220601_16h17m43s_grim

and here's the blue channel:

20220601_16h17m52s_grim

This results in an image that looks like this:

20220601_15h56m13s_grim

* OS: N/A
* Python: N/A
* Pillow: https://github.com/python-pillow/Pillow/commit/b261e179078a36adb030e029c9834d8c00dec4e0

ShadelessFox:
Fixes #6344

|Before|After|
|---|---|
|||


src/libImaging/BcnDecode.c
```diff
@@ -750,8 +750,10 @@ decode_bc6_block(rgb32f *col, const UINT8 *src, int sign) {
         }
     }
     if (info->tr) { /* apply deltas */
-        for (i = 3; i < numep; i++) {
-            endpoints[i] = (endpoints[i] + endpoints[0]) & mask;
+        for (i = 3; i < numep; i += 3) {
+            endpoints[i + 0] = (endpoints[i + 0] + endpoints[0]) & mask;
```
radarhere, line 754:
```suggestion
            endpoints[i] = (endpoints[i] + endpoints[0]) & mask;
```

radarhere:
Thanks very much for figuring this out!

src/libImaging/BcnDecode.c
ShadelessFox: I wanted to keep the code style the same, but okay.

ShadelessFox:
I have several images that can be used as a reference. I'm unsure about their origin. If that's okay, I can attach them.

I originally wanted to add tests, but there are no python bindings for BC6H. Are there C API tests of some kind?

ShadelessFox:
Thanks, I will look into that later.

I was looking at [Khronos' specs](https://registry.khronos.org/DataFormat/specs/1.1/dataformat.1.1.html#_bc6h), and it says the following:

> If the mode has transformed endpoints, the values from `E0` are used as a base to offset all other endpoints, wrapped at the number of endpoint bits. For example, `R1 = (R0 + R1) & ((1 << EPB) - 1)`.

Where `E0` is `(R0, G0, B0)`. The provided reference shows the operation just for a single component, and others were implied.

ShadelessFox:
I hope you don't mind this mess. I can squash commits if you want.

REDxEYE:
All this started from an idea of previewing textures in a java application, both I and @ShadelessFox ended up implementing BC1-BC7 decoders, to see who can get it working first. @ShadelessFox found an issue with delta decoding and I found an issue with the signed format.

ShadelessFox:
Let me know if this PR requires more polishing before it can be merged.

ShadelessFox:
Hi @radarhere,

Is there anything I must do before this PR can be merged?

radarhere:
I'm not aware of any problems at the moment, and you've added tests, so it's just a matter of myself or someone else finding the time to review.

The Valgrind failure happening here isn't due to your code. I've raised a separate issue for it - #6468

radarhere:
Interestingly, I found that Table 58 at https://registry.khronos.org/DataFormat/specs/1.1/dataformat.1.1.html#_bc6h has some mistakes in it (e.g. Mode 1, Bit 4 should be G35 as per Table 57).

Everything in this PR makes sense to me, except for `bc6_gamma_correct`. That doesn't seem to be mentioned in the specification?

ShadelessFox:
It's not the part of the specification. I had to perform gamma correction so the output image looks _fine_ without washed-out colors due to the latter float16 -> int8 conversion required by Pillow. AFAIK Pillow does not support floating point storage?

radarhere:
There is an open issue for multichannel floating point images - #1888. I'm not following why that should lead to washed-out colors though?

If I remove the gamma correction, then the output matches what I get when saving the DDS image as a PNG from macOS Preview. I've created https://github.com/ShadelessFox/Pillow/pull/6

radarhere (approved):

PR #6457, opened 2022-07-22
Add support for ATI1/2(BC4/BC5) DDS files

REDxEYE:
This commit adds support for loading DDS with ATI1 and ATI2 fourcc pixel format

 * Add support for ATI1 dds files
 * Add support for ATI2 dds files


REDxEYE:
I haven't found the ATI1 texture yet, I will add a test for ATI1 when I find texture

Test texture is taken from Zombie Army 4

REDxEYE:
Quick question, how should I go about DDS cube maps? Does pillow support layers or something like it?
If not, should I flatten multi-side/layer textures into one line?

radarhere:
> Quick question, how should I go about DDS cube maps? Does pillow support layers or something like it? If not, should I flatten multi-side/layer textures into one line?

Pillow has a concept of 'frames'. While used for the different images for in an animated GIF, they are also used for the different layers in a PSD. You might want to look at https://github.com/python-pillow/Pillow/blob/main/src/PIL/PsdImagePlugin.py

radarhere:
> Test texture is taken from Zombie Army 4

Is that something that can be distributed under Pillow's license as part of our test suite?

REDxEYE:
> > Test texture is taken from Zombie Army 4
> 
> Is that something that can be distributed under Pillow's license as part of our test suite?

I'll replace them with selfmade textures in next commit

REDxEYE:
@radarhere Do you have any estimations of when this PR will be merged? I have other PR in progress that needs changes from this one

radarhere (approved):

radarhere:
All looks good. My only suggestion is to rearrange the order of the formats - https://github.com/REDxEYE/Pillow/pull/1/files

I think this can be merged after that.

REDxEYE:
> All looks good. My only suggestion is to rearrange the order of the formats - https://github.com/REDxEYE/Pillow/pull/1/files
> 
> I think this can be merged after that.

Good idea, I forgot to sort them. I'm also working on a full rewrite of DDS plugin, using proper enums and flags with support for more pixel formats, but i'm stuck at floating point formats, is there any plans on adding RGBAF32 format or something?

radarhere:
Sounds like you're talking about #1888

hugovk:
Let's also add this to the release notes. Thanks!

https://pillow.readthedocs.io/en/latest/releasenotes/9.3.0.html

radarhere:
Ok, I've created #6465 to add to the release notes.

Issue #6369, opened 2022-06-15
Windows Wheels from Christoph Gohlke Going Away Forever


wiredfool:
According to Christoph Gohlke, https://www.lfd.uci.edu/~gohlke/pythonlibs/  his lab has lost funding and the unofficial python wheels collection, and likely our official windows wheels, is going away before the end of next month.

hugovk:
> **Funding for the Laboratory for Fluorescence Dynamics has ceased. This service will be discontinued before July 2022.**

I'm very sorry to hear this. Thank you @cgohlke, for providing an invaluable service to the Python community for 12+ years. And thank you for all your work helping with Pillow too.

Do you think you'll be able to make a last set of Windows wheels for our July 1 release?

aclark4life:
End of an era, thanks @wiredfool for reporting and thank you @cgohlke for [being there for Pillow since the beginning](https://pypi.org/project/Pillow/1.7.5/#files)!! Not to mention dozens of other projects … 🙏 

@hugovk We probably better figure something out sooner rather than later … can you or @wiredfool or @radarhere remind me how far we got on wheels doing it ourselves last time? Seems like we got close …

aclark4life:
@hugovk @cgohlke Maybe we can push the July release up a week or so if that would help… then we'd have 3 months or so to figure out something else assuming we don't need to do a bug fix release 🤞🤷

hugovk:
We are actually building wheels on GitHub Actions for Windows (for example, see the "dist-main" link at https://github.com/python-pillow/Pillow/actions/runs/2495168122), but as I recall there's one or two things needed for them to be on par with the release ones.

@nulano, do you know what's needed?

@aclark4life We would also like to put out wheels for 3.11 beta, see https://github.com/python-pillow/Pillow/issues/6367, and need to check if multibuild is ready to do 3.11 for Linux/macOS.

(If we wanted to overhaul our wheel building, we could look at the newish https://github.com/pypa/cibuildwheel which can do wheels for all of Windows/Linux/macOS, but that's a bit longer term thing. That could also let us automate uploading releases to PyPI.)

cgohlke:
> Do you think you'll be able to make a last set of Windows wheels for our July 1 release?

Sure.

nulano:
> We are actually building wheels on GitHub Actions for Windows [...] but as I recall there's one or two things needed for them to be on par with the release ones.

[...]

PR #6458, opened 2022-07-23
Switch to GitHub Actions artifacts for Windows wheels

radarhere:
Change to the RELEASING procedure for #6369


stonebig:
would it be possible to get a Pillow wheel for pypy-3.9-v7.3.9 ?

radarhere:
PyPy 3.9 is still in beta, so we don't plan on releasing wheels for it yet.

PR #6470, opened 2022-08-01
Save 1 mode PDF using CCITTFaxDecode filter

### LINKED ISSUE #6453: When I convert TIFF to PDF, the PDF size is 10 times that of TIFF
When I convert TIFF to PDF, the PDF size is 10 times that of TIFF
Why is that?

radarhere:
Resolves #6453

The issue requests the CCITTTaxDecode filter be using to save 1 mode PDFs, to reduce file size.

### DIFF
--- a/src/PIL/PdfImagePlugin.py
+++ b/src/PIL/PdfImagePlugin.py
@@ -25,7 +25,7 @@ import math
 import os
 import time
 
-from . import Image, ImageFile, ImageSequence, PdfParser, TiffImagePlugin, __version__
+from . import Image, ImageFile, ImageSequence, PdfParser, __version__
 
 #
 # --------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


src/PIL/PdfImagePlugin.py
```diff
@@ -123,8 +124,26 @@ def _save(im, fp, filename, save_all=False):
...
+                        PdfParser.PdfDict(
+                            {
+                                "K": -1,
+                                "BlackIs1": True,
+                                "Columns": width,
+                                "Rows": height,
+                            }
+                        )
+                    ]
+                )
```
radarhere, line 146: https://opensource.adobe.com/dc-acrobat-sdk-docs/standards/pdfstandards/pdf/PDF32000_2008.pdf#page=38 indicates that K should be <0 for Group 4 encoding

On the next page of that document, it describes BlackIs1, that allows the colors to be inverted.

src/PIL/PdfImagePlugin.py
```diff
@@ -161,6 +180,11 @@ def _save(im, fp, filename, save_all=False):
 
             if filter == "ASCIIHexDecode":
                 ImageFile._save(im, op, [("hex", (0, 0) + im.size, 0, im.mode)])
+            elif filter == "CCITTFaxDecode":
+                original_strip_size = TiffImagePlugin.STRIP_SIZE
+                TiffImagePlugin.STRIP_SIZE = math.ceil(im.width / 8) * im.height
```
radarhere, line 185: In writing this PR, I found that it was necessary to save the image as a single strip. This has actually been pointed out before, in #5740

src/PIL/TiffImagePlugin.py
```diff
@@ -1684,7 +1684,8 @@ def _save(im, fp, filename):
     stride = len(bits) * ((im.size[0] * bits[0] + 7) // 8)
     # aim for given strip size (64 KB by default) when using libtiff writer
     if libtiff:
-        rows_per_strip = 1 if stride == 0 else min(STRIP_SIZE // stride, im.size[1])
+        im_strip_size = encoderinfo.get("strip_size", STRIP_SIZE)
```
radarhere, line 1687: Rather than updating `STRIP_SIZE` and needing to reset it afterwards, I've added "strip_size" as a TIFF encoder argument.

PR #6484, opened 2022-08-06
Allow default ImageDraw font to be set

### LINKED ISSUE #6480: Support setting a default font
### What did you do?

Was printing CJK characters to an image but got an `AttributeError`. No doubt the default font wasn't supporting such characters.

Understand I can use Unicode-friendly fonts everytime I do `text()`. However, can I somehow change the default font so that I won't need to change everywhere in my program `text()` with a font?

### What actually happened?

Checking the docs there's `PIL.ImageFont.load_default()` to 
> Load a “better than nothing” default font.
Well, it is very clear.
But can I provide some "better than nothing" default for my own use?

* OS: Ubuntu
* Python: 3.8.13
* Pillow: 9.2.0

```python

im = Image.new("RGB", (400, 300))
draw = ImageDraw.Draw(im)
draw.text((10, 10), "你好")

im.show()
```

It gives:
```
AttributeError: 'ImageFont' object has no attribute 'getmask2'
```

radarhere:
Resolves #6480

By raising `ImageDraw.ImageDraw.font` from an instance variable to a class variable, a default ImageDraw font can be set by the user.

```python
from PIL import ImageDraw, ImageFont
ImageDraw.ImageDraw.font = ImageFont.truetype("Tests/fonts/FreeMono.ttf", 120)
```

From then on, ImageDraw operations would use that font as the default, instead of the `ImageFont.load_default()` font.

### DIFF
--- a/docs/releasenotes/9.3.0.rst
+++ b/docs/releasenotes/9.3.0.rst
@@ -31,7 +31,7 @@ Allow default ImageDraw font to be set
 
 Rather than specifying a font when calling text-related ImageDraw methods, or
 setting a font on each ImageDraw instance, the default font can now be set for
-all future ImageDraw operations.
+all future ImageDraw operations::
 
     from PIL import ImageDraw, ImageFont
     ImageDraw.ImageDraw.font = ImageFont.truetype("Tests/fonts/FreeMono.ttf")
--- a/docs/releasenotes/9.3.0.rst
+++ b/docs/releasenotes/9.3.0.rst
@@ -26,6 +26,16 @@ TODO
 API Additions
 =============
 
+Allow default ImageDraw font to be set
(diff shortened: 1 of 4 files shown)


radarhere:
Ok, done

docs/releasenotes/9.3.0.rst
```diff
@@ -26,6 +26,16 @@ TODO
 API Additions
 =============
 
+Allow default ImageDraw font to be set
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Rather than specifying a font when calling text-related ImageDraw methods, or
+setting a font on each ImageDraw instance, the default font can now be set for
+all future ImageDraw operations.
```
hugovk, line 34: Code formatting:

```suggestion
all future ImageDraw operations::
```

Or `.. code-block:: python`

PR #6486, opened 2022-08-06
Support reading BC4U and DX10 BC1 images

REDxEYE:

 * Refactored DDS reader/writter
 * Add Tile namedtuple for code readability
 * Add support for reading/writing single channel DDS textures


REDxEYE:
Is it possible to make pre-commit.ci not to reformat some parts of the code?

radarhere:
Yes. There is [`# fmt: off`](https://github.com/python-pillow/Pillow/blob/3b0f046c5a26eec8b5c9901e719f82500a4000fa/src/PIL/PalmImagePlugin.py#L14) and [`# fmt: on`](https://github.com/python-pillow/Pillow/blob/3b0f046c5a26eec8b5c9901e719f82500a4000fa/src/PIL/PalmImagePlugin.py#L80). See https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#code-style for documentation.

src/PIL/Image.py
```diff
@@ -216,6 +217,15 @@ class Quantize(IntEnum):
...
 
+Tile = NamedTuple(
+    "Tile",
+    [
+        ("encoder_name", str),
+        ("extents", Tuple[int, int, int, int]),
+        ("offset", int),
+        ("tile_args", Tuple),
+    ],
+)
```
radarhere, line 228: Could you say a few words about why this is helpful?

src/PIL/DdsImagePlugin.py
```diff
@@ -121,131 +289,165 @@ def _open(self):
...
+        + struct.pack("<IIIII", DDSCAPS.TEXTURE, 0, 0, 0, 0)
     )
-    if im.mode == "RGBA":
-        r, g, b, a = im.split()
-        im = Image.merge("RGBA", (a, r, g, b))
-    ImageFile._save(im, fp, [("raw", (0, 0) + im.size, 0, (im.mode[::-1], 0, 1))])
+    if im.mode == 'LA':
+        ImageFile._save(im, fp, [Image.Tile("raw", (0, 0) + im.size, 0, ('LA', 0, 1))])
+    else:
+        ImageFile._save(im, fp, [Image.Tile("raw", (0, 0) + im.size, 0, (im.mode[::-1], 0, 1))])
```
radarhere, line 447:
```suggestion
    mode = "LA" if im.mode == "LA" else im.mode[::-1]
    ImageFile._save(im, fp, [Image.Tile("raw", (0, 0) + im.size, 0, (mode, 0, 1))])
```
This is simpler, and should fix the lint error.

src/PIL/Image.py
REDxEYE: I've added it to understand what tile structure is. I didn't found any doc for it, so i scouted sources to understand what everything mean

src/PIL/DdsImagePlugin.py
```diff
@@ -19,97 +20,264 @@
...
+    P208 = 130
+    V208 = 131
+    V408 = 132
+    SAMPLER_FEEDBACK_MIN_MIP_OPAQUE = 133
+    SAMPLER_FEEDBACK_MIP_REGION_USED_OPAQUE = 134
+    INVALID = -1
+
+    @classmethod
+    def _missing_(cls, value: object):
+        return cls.INVALID
```
radarhere, line 191: Would you be able to give a short explanation why this is helpful?

src/PIL/DdsImagePlugin.py
```diff
@@ -19,97 +20,264 @@
...
+    IA44 = 112
+    P8 = 113
+    A8P8 = 114
+    B4G4R4A4_UNORM = 115
+    P208 = 130
+    V208 = 131
+    V408 = 132
+    SAMPLER_FEEDBACK_MIN_MIP_OPAQUE = 133
+    SAMPLER_FEEDBACK_MIP_REGION_USED_OPAQUE = 134
+    INVALID = -1
```
radarhere, line 187: Where did this list come from?

src/PIL/DdsImagePlugin.py
REDxEYE: This can be removed safely. I copied this enum from my other project and forgot to strip unneeded stuff

src/PIL/DdsImagePlugin.py
REDxEYE: https://learn.microsoft.com/en-us/windows/win32/api/dxgiformat/ne-dxgiformat-dxgi_format

src/PIL/DdsImagePlugin.py
REDxEYE: Ye, both can be removed

src/PIL/DdsImagePlugin.py
```diff
@@ -9,103 +9,254 @@
...
+    BC7_TYPELESS = 97
+    BC7_UNORM = 98
+    BC7_UNORM_SRGB = 99
+    AYUV = 100
+    Y410 = 101
+    Y416 = 102
+    NV12 = 103
+    P010 = 104
+    P016 = 105
+    _420_OPAQUE = 106
```
radarhere, line 172: Just to note - this should be `420_OPAQUE`, but you've prefixed it with an underscore because otherwise Python will throw a SyntaxError.

REDxEYE: Is there any better way around this?

radarhere: No, I don't think so, not while using an enum.

radarhere:
I extracted the support for single channel images, and it has been merged as PR #6820.

REDxEYE:
Awesome!

REDxEYE:
I've updated branch and added BC5U support

radarhere:
I created #7358 to separately add BC5U reading, and that has been merged.

REDxEYE:
Anything I need to improve (besides rebase) to get this PR merged?

radarhere:
Your intention here is, correct me if I'm wrong, not primarily to add new functionality, but to reorganise the plugin's API into a form that you feel is more logical. From my perspective, Pillow values backwards compatibility, and that is at odds with this attempt.

You're welcome to see if other members of the Pillow team feel differently. I'm not against enums, I'm just reluctant to go against one of our guiding principles.

REDxEYE:
Ah, I see. I can refactor it to use variables instead of enums. 
What about the other parts of the refactor?

radarhere:
@hugovk did you have any thoughts on this?

src/PIL/ImageFile.py
```diff
@@ -31,6 +31,7 @@
 import itertools
 import struct
 import sys
+from typing import List
```
hugovk, line 34: Let's add `from __future__ import annotations` as the first import and remove this:
```suggestion
```

src/PIL/ImageFile.py
```diff
@@ -521,13 +522,13 @@ def _save(im, fp, tile, bufsize=0):
         fp.flush()
 
 
-def _encode_tile(im, fp, tile, bufsize, fh, exc=None):
-    for e, b, o, a in tile:
-        if o > 0:
-            fp.seek(o)
-        encoder = Image._getencoder(im.mode, e, a, im.encoderconfig)
+def _encode_tile(im, fp, tile: List[Image.Tile], bufsize, fh, exc=None):
```
hugovk, line 525:
```suggestion
def _encode_tile(im, fp, tile: list[Image.Tile], bufsize, fh, exc=None):
```

src/PIL/Image.py
```diff
@@ -38,6 +38,7 @@
 from collections.abc import Callable, MutableMapping
 from enum import IntEnum
 from pathlib import Path
+from typing import List, NamedTuple, Tuple
```
hugovk, line 41: Let's add `from __future__ import annotations` and then:
```suggestion
from typing import NamedTuple
```

src/PIL/Image.py
```diff
@@ -206,6 +207,15 @@ class Quantize(IntEnum):
...
 
+Tile = NamedTuple(
+    "Tile",
+    [
+        ("encoder_name", str),
+        ("extents", Tuple[int, int, int, int]),
+        ("offset", int),
+        ("tile_args", Tuple),
+    ],
+)
```
hugovk, line 218:
```suggestion
class Tile(NamedTuple):
    encoder_name: str
    extents: tuple[int, int, int, int]
    offset: int
    tile_args: tuple
```

And let's name it `_Tile` so it's not part of the public API?

src/PIL/Image.py
```diff
@@ -692,6 +702,7 @@ def __getstate__(self):
 
     def __setstate__(self, state):
         Image.__init__(self)
+        self.tile: List[Tile] = []
```
hugovk, line 705:
```suggestion
        self.tile: list[Tile] = []
```

src/PIL/DdsImagePlugin.py
```diff
@@ -9,103 +9,255 @@
...
 
 from . import Image, ImageFile
+from ._binary import i32le as i32
 from ._binary import o32le as o32
 
 # Magic ("DDS ")
 DDS_MAGIC = 0x20534444
 
-# DDS flags
-DDSD_CAPS = 0x1
```
hugovk, line 23: Please don't delete these, it's a backwards incompatible change in case someone was using `DdsImagePlugin.DDSD_CAPS` etc.

(I've not reviewed the rest of this file.)

Tests/images/l.dds
hugovk, line 1: Would it be better to rename these files like `mode-l.dds`, `mode-l.png`, `mode-la.png` etc. to better describe them, group in directory listing?

hugovk (changes requested):
The test changes are fine, here's a naming suggestion.

The `Image.py` and `ImageFile.py` changes are mostly fine, some typing and naming suggestions there.

For `DdsImagePlugin.py`, let's not break backwards compatibility.

REDxEYE:
Understood. 
I agree with all suggestions, I will apply them as soon as I can.

src/PIL/DdsImagePlugin.py
```diff
@@ -124,114 +353,169 @@ def _open(self):
...
+                        "bcn", extents, data_offs, (7, self.pixel_format)
+                    )
+                elif dxgi_format == DXGI_FORMAT.BC7_UNORM_SRGB:
+                    self._mode = "RGBA"
                     self.pixel_format = "BC7"
                     self.info["gamma"] = 1 / 2.2
-                    n = 7
+                    tile = Image._Tile(
+                        "bcn", extents, data_offs, (7, self.pixel_format)
+                    )
```
radarhere, line 498: What is the thinking behind this change, where instead of specifying different `n` values, you write out the entire tile each time? It would seem to be increasing the amount of duplicate code.

REDxEYE: Hm, yeah, you're right. I will refactor it

REDxEYE: After second though, the way i structured if-elif chain does not allow me to easily refactor it back to n= thing

radarhere: See what you think of https://github.com/REDxEYE/Pillow/pull/6

src/PIL/DdsImagePlugin.py
```diff
@@ -124,114 +335,162 @@ def _open(self):
...
-            # Texture contains uncompressed L or LA data
-            if pfflags & DDPF_ALPHAPIXELS:
-                self._mode = "LA"
-            else:
-                self._mode = "L"
-
-            self.tile = [("raw", (0, 0) + self.size, 0, (self.mode, 0, 1))]
-        elif pfflags & DDPF_RGB:
+        if flags & DDSD.CAPS:
+            header.seek(20, io.SEEK_CUR)
```
radarhere, line 350: https://learn.microsoft.com/en-us/windows/win32/direct3ddds/dds-header#members states that `DDSD_CAPS` is "Required in every .dds file."

So why do you think this will appear in some files but not others?

REDxEYE: Just a safety check. What if we get weird dds that have no caps flag and no caps section?

nulano: The website continues with:
> When you write .dds files, you should set the DDSD_CAPS and DDSD_PIXELFORMAT flags, and for mipmapped textures you should also set the DDSD_MIPMAPCOUNT flag. However, when you read a .dds file, you should not rely on the DDSD_CAPS, DDSD_PIXELFORMAT, and DDSD_MIPMAPCOUNT flags being set because some writers of such a file might not set these flags.

It's not clear to me whether this means readers should always act as if the flag is set, or if a file can be readable without this data. Either way, if the flag is missing, it just means the caps section is set to unknown data, not that it is missing (that would require the `header_size` value to be less than 124). I don't see anywhere the flags value or caps section is actually used when reading a file in Pillow, so I don't see what purpose the seek call has?

REDxEYE:
Just found weird DDS file with DDPF_RGB flag, but RGBBitsCount = 8. So need to handle them too

src/PIL/DdsImagePlugin.py
radarhere: I've added a commit to https://github.com/REDxEYE/Pillow/pull/7 to remove the seek call.

src/PIL/DdsImagePlugin.py
```diff
@@ -124,114 +335,144 @@ def _open(self):
...
                 self.fp.read(16)
-                if dxgi_format in (DXGI_FORMAT_BC5_TYPELESS, DXGI_FORMAT_BC5_UNORM):
+                if dxgi_format in (
+                    DXGI_FORMAT.BC1_UNORM,
+                    DXGI_FORMAT.BC1_UNORM_SRGB,
+                    DXGI_FORMAT.BC1_TYPELESS,
+                ):
+                    self._mode = "RGBA"
+                    self.pixel_format = "BC1"
+                    n = 1
```
radarhere, line 426: Our tradition is to add tests for new code, to try and prevent it breaking in the future. Do you have an image that can be used to test this?

REDxEYE: Ye, but they are from that game and developers do not allow usage of their assets. So I need to craft one

REDxEYE: Oh, sorry, replied to wrong comment

radarhere: I've added a commit to https://github.com/REDxEYE/Pillow/pull/7. I copied the existing test image dxt1-rgb-4bbp-noalpha_MipMaps-1.dds and modified it to be BC1_UNORM and BC1_TYPELESS. However, I have removed BC1_UNORM_SRGB. See what you think.

src/PIL/DdsImagePlugin.py
```diff
@@ -9,103 +9,314 @@
...
+    BC4S = i32(b"BC4S")
+    BC4U = i32(b"BC4U")
+    BC5S = i32(b"BC5S")
+    BC5U = i32(b"BC5U")
+    ATI1 = i32(b"ATI1")
+    ATI2 = i32(b"ATI2")
+    MULTI2_ARGB8 = i32(b"MET1")
+
+
+# Backward compat layer
```
radarhere, line 266:
```suggestion
# Backward compatibility layer
```

Tests/test_file_dds.py
```diff
@@ -316,13 +347,35 @@ def test_save_unsupported_mode(tmp_path):
...
+
+
 @pytest.mark.parametrize(
     ("mode", "test_file"),
     [
         ("L", "Tests/images/linear_gradient.png"),
         ("LA", "Tests/images/uncompressed_la.png"),
         ("RGB", "Tests/images/hopper.png"),
         ("RGBA", "Tests/images/pil123rgba.png"),
+        ("L", TEST_FILE_ATI1),
```
radarhere, line 378: Did you have any particular reason for adding `test_open`, and for adding "L" to `test_save`? These tests don't fail without this PR, and it's not immediately apparent to me that it adds more coverage?

REDxEYE: for test_save, I probably wanted to test L saving and didn't notice that L was already tested

REDxEYE: about test_open, I don't really remember, but i think coverage tool was complaining about some path not being covered

radarhere: Hmm. That shouldn't be the case. For L, we have uncompressed_l.dds, for LA uncompressed_la.dds, for RGB hopper.dds and for RGBA uncompressed_rgb.dds.

I've created https://github.com/REDxEYE/Pillow/pull/9. See whether or not you'd like to merge. If I'm wrong, and there is missing coverage, it can be undone in some way.

radarhere: Thanks. The PR has been merged, and coverage is fine.

radarhere:
The benefits of the refactoring in this PR from my current understanding
- adds `Image._Tile` to help code clarity
- adds enums to DdsImagePlugin (without removing existing constants)
- previously, only flags were used to determine the type of uncompressed data. Now the flags and the number of bits are used, to be more certain about the data type

Other benefits we are getting at the same time
- reading uncompressed "RGB" images with 8 bits
- reading DX10 BC1 UNORM and TYPELESS images
- reading BC4U images

radarhere:
> Just found weird DDS file with DDPF_RGB flag, but RGBBitsCount = 8

I previously presumed that this would be greyscale, but now I see no reason why the masks shouldn't be applied, so that this image would entirely different shades of red, for example. Is that the case with the weird file you found? Is it entirely red, or entirely green or entirely blue?

radarhere (approved):

REDxEYE:
A bit offtop. I may work on adding support for PVR textures in some time. I already implemented ETC decoder(one of they compression formats) https://github.com/REDxEYE/TextureDecoder/blob/master/library/src/decoders/etc/etc1.cpp

Tests/test_file_dds.py
```diff
@@ -29,11 +32,20 @@
...
+    "image_path",
+    (
+        TEST_FILE_DXT1,
+        # hexeditted to use DX10 FourCC
+        TEST_FILE_DX10_BC1,
+        TEST_FILE_DX10_BC1_TYPELESS,
+    ),
+)
+def test_sanity_dxt1_bc1(image_path):
     """Check DXT1 images can be opened"""
```
radarhere, line 45:
```suggestion
    """Check DXT1 and BC1 images can be opened"""
```

src/PIL/DdsImagePlugin.py
hugovk: How about `OPAQUE_420`? Some might think it's private with a leading underscore.

src/PIL/DdsImagePlugin.py
```diff
@@ -10,104 +10,314 @@
...
+    BC7_TYPELESS = 97
+    BC7_UNORM = 98
+    BC7_UNORM_SRGB = 99
+    AYUV = 100
+    Y410 = 101
+    Y416 = 102
+    NV12 = 103
+    P010 = 104
+    P016 = 105
+    _420_OPAQUE = 106
```
hugovk, line 174:
```suggestion
    OPAQUE_420 = 106
```

hugovk (approved):
Thanks!

Let's also add release notes for this (can be in a followup PR) and retitle the PR for the most important thing ("Add support for ..."?).

src/PIL/DdsImagePlugin.py
radarhere: I'm ok with this. @REDxEYE ?

radarhere:
I've retitled the PR, and added release notes. The release notes also cover #7603

docs/releasenotes/10.2.0.rst
```diff
@@ -0,0 +1,54 @@
...
+TODO
+
+API Additions
+=============
+
+Added DdsImagePlugin enums
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+``DDSD``, ``DDSCAPS``, ``DDSCAPS2``, ``DDPF``, ``DXGI_FORMAT`` and ``D3DFMT``
+enums have been added to DdsImagePlugin.
```
radarhere, line 35:
```suggestion
enums have been added to :py:class:`PIL.DdsImagePlugin`.
```

docs/releasenotes/10.2.0.rst
```diff
@@ -0,0 +1,56 @@
...
+
+TODO
+
+API Additions
+=============
+
+Added DdsImagePlugin enums
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+:py:class:`PIL.DdsImagePlugin.DDSD`, :py:class:`PIL.DdsImagePlugin.DDSCAPS`,
```
hugovk, line 34: Let's add the tilde to all these, so it only shows the enum name not its full path. 

https://pillow--6486.org.readthedocs.build/en/6486/releasenotes/10.2.0.html#added-ddsimageplugin-enums


```suggestion
:py:class:`~PIL.DdsImagePlugin.DDSD`, :py:class:`~PIL.DdsImagePlugin.DDSCAPS`,
```

docs/releasenotes/10.2.0.rst
```diff
@@ -0,0 +1,56 @@
...
+:py:class:`PIL.DdsImagePlugin.DDSCAPS2`, :py:class:`PIL.DdsImagePlugin.DDPF`,
```
hugovk, line 35:
```suggestion
:py:class:`~PIL.DdsImagePlugin.DDSCAPS2`, :py:class:`~PIL.DdsImagePlugin.DDPF`,
```

docs/releasenotes/10.2.0.rst
```diff
@@ -0,0 +1,56 @@
...
+:py:class:`PIL.DdsImagePlugin.DXGI_FORMAT` and :py:class:`PIL.DdsImagePlugin.D3DFMT`
```
hugovk, line 36:
```suggestion
:py:class:`~PIL.DdsImagePlugin.DXGI_FORMAT` and :py:class:`~PIL.DdsImagePlugin.D3DFMT`
```

hugovk:
Thank you for your patience! This will be in the next release, set for 2nd January 2024.

PR #6489, opened 2022-08-10
Restored __array__ for libraries that do not support __array_interface__

### LINKED ISSUE #6422: 9.2.0 breaks compatibility to pass PIL.Image to tensorflow (tf.reshape)
On passing it tf.reshape throws:
```
Attempt to convert a value (<PIL.Image.Image image mode=RGB size=100x100 at 0x153B2CF8EE0>) with an unsupported type (<class 'PIL.Image.Image'>) to a Tensor.
```

radarhere:
Resolves #6422

The issue has discovered that tensorflow supports converting Pillow images using `__array__`, but not `__array_interface__`. One solution would be to say that tensorflow should support `__array_interface__`, and an issue has been opened for that - https://github.com/tensorflow/tensorflow/issues/56723

This PR is an alternative, where we add `__array__`, to provide support from our end. Conceivably, there may be other libraries that support `__array__`, but not `__array_interface__`. Testing, I find that NumPy prefers `__array_interface__` over `__array__`, so this should not affect NumPy use.

### DIFF
--- a/Tests/test_image_array.py
+++ b/Tests/test_image_array.py
@@ -94,3 +94,10 @@ def test_fromarray_palette():
 
     # Assert that the Python and C palettes match
     assert len(out.palette.colors) == len(out.im.getpalette()) / 3
+
+
+def test_array():
+    i = im.convert("L")
+    a = numpy.array(i)
+
+    assert (a == i.__array__()).all()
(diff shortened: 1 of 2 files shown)


radarhere:
This was resolved first by tensorflow - https://github.com/tensorflow/tensorflow/commit/8b3369b3cb4c3ecc3e0059a550c2d7cad136d371

PR #6491, opened 2022-08-10
Corrected some docstrings that are using enumeration parameters

hecent:
Converted things like Dither.FLOYDSTEINBERG to PIL.Image.Dither.FLOYDSTEINBERG to match other docstrings and to as that is the correct code.

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -892,12 +892,12 @@ class Image:
            should be 4- or 12-tuple containing floating point values.
         :param dither: Dithering method, used when converting from
            mode "RGB" to "P" or from "RGB" or "L" to "1".
-           Available methods are :data:`Dither.NONE` or :data:`Dither.FLOYDSTEINBERG`
+           Available methods are :data:`PIL.Image.Dither.NONE` or :data:`PIL.Image.Dither.FLOYDSTEINBERG`
            (default). Note that this is not used when ``matrix`` is supplied.
         :param palette: Palette to use when converting from mode "RGB"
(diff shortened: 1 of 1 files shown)


hugovk:
Thanks for the PR, here's an example:

<table>
<tr>
	<td><a href="https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.quantize">Before</a>
	<td><a href="https://pillow--6491.org.readthedocs.build/en/6491/reference/Image.html#PIL.Image.Image.quantize">After</a>
<tr>
	<td>image
	<td>image
</table>

In both cases, they link to the same things. For example, the first link:

* https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Quantize.MEDIANCUT
* https://pillow--6491.org.readthedocs.build/en/6491/reference/Image.html#PIL.Image.Quantize.MEDIANCUT

But the PR makes the links appear more verbose.

hecent:
The PR was to make the links the same format as other docstrings like `resize()` or `transform()` and so that they better represent what you are actually supposed to write when passing the options as parameters:
If you didn't know better it seems as though they should be written as `Image.quantize(method=Quantize.MEDIANCUT)` from reading the docs (which I did), and is incorrect.

(https://pillow.readthedocs.io/en/stable/reference/Image.html)
image

radarhere:
#6557 has been merged instead, resolving the inconsistency by replacing `PIL.Image.Resampling.NEAREST` with `Resampling.NEAREST`.

PR #6504, opened 2022-08-14
Allow RGB and RGBA values for PA image putpixel

### LINKED ISSUE #6502: Possible missing "PA" check
https://github.com/python-pillow/Pillow/blob/5a087ccbb9b48cba04276fd1b7d0cec93488b7c3/src/PIL/PyAccess.py#L59-L62

This code saves a reference to the image palette, but only for "P" mode images. I think it should probably check for "PA" here too. This saved palette is used later in this class, but again only for "P", not "PA".

https://github.com/python-pillow/Pillow/blob/5a087ccbb9b48cba04276fd1b7d0cec93488b7c3/src/PIL/PyAccess.py#L91-L97

I haven't checked if this palette reference is used outside of this class, though it is supposed to be private so hopefully not.

radarhere:
Resolves #6502

This is continuing the work of #3519, "Allow RGB and RGBA values for P image putpixel", allowing the behaviour for PA images as well, so that the following code can run.
```python
from PIL import Image
im = Image.new("PA", (1, 1))
im.putpixel((0, 0), (255, 0, 0))
```


hugovk (approved):
Looks good, ready for merge once conflict fixed.

PR #6509, opened 2022-08-16
Added direct conversion from PA to P

radarhere:
#6506 asked how Pillow converted images from PA to P.

The answer was because PA to P conversion doesn't exist, Pillow instead [falls back](https://github.com/python-pillow/Pillow/blob/87ecd01fc042bb67b11bfcff3db99fc2b61e702d/src/PIL/Image.py#L1034-L1040) to converting from PA to RGB, and then RGB to P.

This PR adds direct conversion from PA to P. There should be no functional change, only an increase in efficiency.

### DIFF
--- a/src/libImaging/Convert.c
+++ b/src/libImaging/Convert.c
@@ -1027,6 +1027,14 @@ pa2l(UINT8 *out, const UINT8 *in, int xsize, ImagingPalette palette) {
 }
 
 static void
+pa2p(UINT8 *out, const UINT8 *in, int xsize, ImagingPalette palette) {
+    int x;
+    for (x = 0; x < xsize; x++, in += 4) {
+        *out++ = in[0];
+    }
+}
+
+static void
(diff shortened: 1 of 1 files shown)


src/libImaging/Convert.c
```diff
@@ -1209,6 +1217,8 @@ frompalette(Imaging imOut, Imaging imIn, const char *mode) {
         convert = alpha ? pa2l : p2l;
     } else if (strcmp(mode, "LA") == 0) {
         convert = alpha ? pa2la : p2la;
+    } else if (strcmp(mode, "P") == 0) {
+        convert = pa2p;
```
radarhere, line 1221: Strange that Codecov doesn't think this is tested - when I run the test suite on my local machine, it is tested by [test_image_convert.py's test_sanity.](https://github.com/python-pillow/Pillow/blob/87ecd01fc042bb67b11bfcff3db99fc2b61e702d/Tests/test_image_convert.py#L14-L33)

PR #6519, opened 2022-08-22
Parametrize Tests/test_image_paste.py

Yay295:
Just changed the tests to use pytest parameters instead of looping through the parameters inside each method.

### DIFF
--- a/Tests/test_image_paste.py
+++ b/Tests/test_image_paste.py
@@ -1,3 +1,5 @@
+import pytest
+
 from PIL import Image
 
 from .helper import CachedProperty, assert_image_equal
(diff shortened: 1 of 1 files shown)


Tests/test_image_paste.py
```diff
@@ -101,226 +103,303 @@ def gradient_RGBa(self):
...
-                ],
-            )
+    @pytest.mark.parametrize(
+        "mode",
+        [
+            "RGBA",
+            "RGB",
+            "L",
+        ],
+    )
```
hugovk, line 113: This (and the others below) are fine as one-liners:
```suggestion
    @pytest.mark.parametrize("mode", ["RGBA", "RGB", "L"])
```

hugovk:
Thank you!

PR #6521, opened 2022-08-22
Changed contain function

bibinhashley:
Fixes # .

 * I was using ImageOps.pad for resizing an image by keeping the aspect ratio. But when I used pad function for an image with size (4307, 6030) to resize it to (50,70), it extended right side only. When I manually resized it, that image didn't need any extension or padding to keep the aspect ratio because both sizes are proportional.
 * So I checked the function in pillow repo, and found that pad function is using contain function and in contain function, finding new height or width for keeping aspect ratio was using this code : 
 `new_height = int(image.height / image.width * size[0])`
`new_width = int(image.width / image.height * size[1])`
 * I think this code has a problem, actually instead of "int", it should use math.ceil so that the next highest number will be used as size(width or height)
 * In my image by running contain function by running ImageOps.contain, it gave an output image with size (49,70), not (50,70).

### DIFF
--- a/src/PIL/ImageOps.py
+++ b/src/PIL/ImageOps.py
@@ -18,9 +18,10 @@
 #
 
 import functools
+import math
 import operator
 import re
-import math
+
 from . import Image
 
 #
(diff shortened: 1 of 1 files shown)


hugovk:
(For future reference, new version at https://github.com/python-pillow/Pillow/pull/6522)

PR #6524, opened 2022-08-22
Temporarily skip valgrind failure

radarhere:
Temporarily skip the valgrind failure introduced by #6470, so that any new failing tests can be seen more clearly.

See https://github.com/radarhere/Pillow/runs/7962954391?check_suite_focus=true for a passing valgrind build with this change.


radarhere:
#6699 reverts this skip.

PR #6534, opened 2022-08-25
Use fixtures instead of classes in test_imagefont

nulano:

* Use fixtures instead of classes to make `test_imagefont.py` a bit easier to read.
* Parameterize tests for `TransposedFont`.
* Combine `test_find_linux_font` and `test_find_macos_font` into a single parameterized test.

### DIFF
--- a/Tests/test_imagefont.py
+++ b/Tests/test_imagefont.py
@@ -28,497 +28,527 @@ TEST_TEXT = "hey you\nyou are awesome\nthis looks awkward"
 pytestmark = skip_unless_feature("freetype2")
 
 
-class TestImageFont:
-    LAYOUT_ENGINE = ImageFont.Layout.BASIC
+def test_sanity():
+    assert re.search(r"\d+\.\d+\.\d+$", features.version_module("freetype2"))
 
-    def get_font(self):
-        return ImageFont.truetype(
-            FONT_PATH, FONT_SIZE, layout_engine=self.LAYOUT_ENGINE
-        )
 
-    def test_sanity(self):
-        assert re.search(r"\d+\.\d+\.\d+$", features.version_module("freetype2"))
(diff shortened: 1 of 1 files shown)


Tests/test_imagefont.py
```diff
@@ -28,497 +28,527 @@
...
-    LAYOUT_ENGINE = ImageFont.Layout.BASIC
+def test_sanity():
+    assert re.search(r"\d+\.\d+\.\d+$", features.version_module("freetype2"))
+
+
+@pytest.fixture(
+    scope="module",
+    params=[
+        ImageFont.Layout.BASIC,
+        pytest.param(ImageFont.Layout.RAQM, marks=skip_unless_feature("raqm")),
```
hugovk, line 39: Can we do this for consistency?
```suggestion
        pytest.param(ImageFont.Layout.BASIC),
        pytest.param(ImageFont.Layout.RAQM, marks=skip_unless_feature("raqm")),
```

hugovk (commented):
One quick suggestion, not reviewed it all yet.

---

For next time: it would probably be easier to review this as separate commits, rather than a big, single change. Don't worry about it this time, unless it's easy to re-do. Thanks!

Tests/test_imagefont.py
nulano: Good idea, seems to work fine.

nulano:
> For next time: it would probably be easier to review this as separate commits, rather than a big, single change. 

Just in case you aren't aware of it, I find the split view with "Hide whitespace" enabled the easiest to read for reviewing this:

image

> Don't worry about it this time, unless it's easy to re-do. Thanks!

Not easy to split for this PR, but I will try to make smaller commits for next time.

hugovk:
Thanks!

PR #6540, opened 2022-08-26
Change pixel size of "LA", "La", and "PA" from 4 to 2

### LINKED ISSUE #6503: Change pixel size of "LA", "La", and "PA" from 4 to 2
These image modes currently use 4 bytes for each pixel, but they only need two. I've started looking into this, but there are a lot more changes than I expected, so I'm creating this issue for documentation. The files I've found that need to be changed are:
* /src/libImaging/Access.c
  * `ImagingAccessInit()` contains a mapping to three functions (also defined in this file) for each of these modes.
* /src/libImaging/Bands.c
  * Lots of special casing for "LXXA", and some functions assume an image with 2 bands uses 4 bytes per pixel.
* /src/libImaging/Convert.c
  * All of the conversion functions for these modes.
* /src/libImaging/Geometry.c
  * `bicubic_filter32LA()` and `bilinear_filter32LA()`.
* /src/libImaging/GetBBox.c
  * `ImagingGetBBox()` creates a bit mask for the alpha channel.
* /src/libImaging/Imaging.h
  * `IMAGING_PIXEL_LA` and `IMAGING_PIXEL_PA` are defined in this file, though they aren't used anywhere.
* /src/libImaging/Jpeg2KDecode.c
  * `j2ku_graya_la()` is currently used for both "LA" and "RGBA", so it will first have to be duplicated and renamed for "RGBA", and then the original can be modified.
* /src/libImaging/Jpeg2KEncode.c
  * `j2k_pack_la()`.
* /src/libImaging/Pack.c
  * The references to `packLA()` can be replaced with `copy2()`, and `packLAL()` needs to be modified. `packLA()` is no longer used after this change, so it can be removed.
* /src/libImaging/Paste.c
  * `fill_mask_L()` uses a mask for the alpha channel, and `ImagingPaste()` calls the same function for "LA" and "RGBA".
* /src/libImaging/Storage.c
  * The image pixel size and line size are set in `ImagingNewPrologueSubtype()`.
* /src/libImaging/Unpack.c
  * `unpackLA()` and `unpackLAL()`.
* /src/PIL/PyAccess.py
  * The `mode_map` uses `_PyAccess32_2()` for these modes.

Changing `im->pixelsize` from 4 to 2 also means the image data will be stored in `im->image8` instead of `im->image32`. `im->image` also exists and always holds the image data. The only difference between these pointers is their type: `image` is `char`, `image8` is `UINT8`, and `image32` is `INT32`. Some functions check for `im->image8` being set and do different things based on that.

Yay295:
Fixes #6503


Yay295:
hmm, everything passed when I ran it on my computer. I think the failure in test_image.py might be an endianness issue. I added a lot more modes to the test because they all passed for me, but I guess they still have issues on other systems.

Yay295:
It looks like the Valgrind errors are because the image is created with space for 4-byte pixels, but then only the first three are assigned to a value. Later, the pixel size is used to loop over the bytes, and since the last byte hadn't been assigned to anything, Valgrind complains about it. It's not actually a problem though since those bytes won't be used.

I think the easiest fix would be to zero-out those bytes when the image is allocated. Another solution would be to go through all of the places where the pixel size is checked and make sure it ignores that last byte if it's not being used. This would probably be faster since it would skip a lot of computation, but it would also complicate the code having to figure out the right size everywhere.

Yay295:
I'm not sure what to do about this error on the s390s machine:

```
__________________ TestImageTransform.test_nearest_resize[LA] __________________

self = <Tests.test_image_transform.TestImageTransform object at 0x40085c28f0>
mode = 'LA'

    @pytest.mark.parametrize("mode", ("RGBA", "LA"))
    def test_nearest_resize(self, mode):
        def op(im, sz):
            return im.resize(sz, Image.Resampling.NEAREST)
    
>       self._test_nearest(op, mode)

Tests/test_image_transform.py:186: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <Tests.test_image_transform.TestImageTransform object at 0x40085c28f0>
op = <function TestImageTransform.test_nearest_resize.<locals>.op at 0x40fae7dfc0>
mode = 'LA'

    def _test_nearest(self, op, mode):
        # create white image with half transparent,
        # do op,
        # the image should remain white with half transparent
        transparent, opaque = {
            "RGBA": ((255, 255, 255, 0), (255, 255, 255, 255)),
            "LA": ((255, 0), (255, 255)),
        }[mode]
        im = Image.new(mode, (10, 10), transparent)
        im2 = Image.new(mode, (5, 10), opaque)
        im.paste(im2, (0, 0))
    
        im = op(im, (40, 10))
    
        colors = im.getcolors()
>       assert colors == [
            (20 * 10, opaque),
            (20 * 10, transparent),
        ]
E       assert [(200, (255, ..., (255, 255))] == [(200, (255, ...00, (255, 0))]
E         At index 0 diff: (200, (255, 0)) != (200, (255, 255))
E         Full diff:
E         - [(200, (255, 255)), (200, (255, 0))]
E         + [(200, (255, 0)), (200, (255, 255))]

Tests/test_image_transform.py:176: AssertionError
```

It's not actually returning a wrong result; it's just backwards. Probably a combination of the hash results being different now that it's two bytes instead of four, though still stored in an INT32.

Yay295:
Actually, it would probably work if I changed it to a set comparison instead of an array comparison, since all of the colors should be unique anyways.

Yay295:
Everything's passing! Except code coverage went down by 0.08%.

Yay295:
Is there anything else that needs to be done before this can be merged?

Tests/test_image_transform.py
```diff
@@ -173,10 +173,10 @@ def _test_nearest(self, op, mode):
         im = op(im, (40, 10))
 
         colors = im.getcolors()
-        assert colors == [
+        assert set(colors) == {
             (20 * 10, opaque),
             (20 * 10, transparent),
-        ]
+        }
```
radarhere, line 179: This is unrelated to the changes made, yes?

Tests/test_image.py
```diff
@@ -528,16 +527,46 @@ def test_one_item_tuple(self):
...
-        # Arrange
-        wrong_mode = "RGB"
-
-        # Act / Assert
+    @pytest.mark.parametrize(
+        "wrong_mode",
+        (
+            "bad",
+            "I;16N",
+            # "LAB",
```
radarhere, line 535:
```suggestion
```

Tests/test_image.py
```diff
@@ -528,16 +527,46 @@ def test_one_item_tuple(self):
...
+            ("P", 0, 255),
+            ("PA", (0, 255), (255, 255)),
+            ("L", 0, 255),
+            ("LA", (0, 255), (255, 255)),
+            ("La", (0, 255), (255, 255)),
+            ("F", 0, 255),
+            ("I", 0, 255),
+            # ("I;16", 0, 255),
+            # ("I;16L", 0, 255),
+            # ("I;16B", 0, 255),
```
radarhere, line 559:
```suggestion
```

Tests/test_image.py
```diff
@@ -547,22 +576,52 @@ def test_linear_gradient(self, mode):
...
-        # Arrange
-        wrong_mode = "RGB"
-
-        # Act / Assert
+    @pytest.mark.parametrize(
+        "wrong_mode",
+        (
+            "bad",
+            "I;16N",
+            # "LAB",
```
radarhere, line 590:
```suggestion
```

Tests/test_image.py
```diff
@@ -547,22 +576,52 @@ def test_linear_gradient(self, mode):
...
+            ("P", 255, 0),
+            ("PA", (255, 255), (0, 255)),
+            ("L", 255, 0),
+            ("LA", (255, 255), (0, 255)),
+            ("La", (255, 255), (0, 255)),
+            ("F", 255, 0),
+            ("I", 255, 0),
+            # ("I;16", 255, 0),
+            # ("I;16L", 255, 0),
+            # ("I;16B", 255, 0),
```
radarhere, line 614:
```suggestion
```

src/libImaging/Resample.c
```diff
@@ -286,88 +286,30 @@ void
 ImagingResampleHorizontal_8bpc(
     Imaging imOut, Imaging imIn, int offset, int ksize, int *bounds, double *prekk) {
     ImagingSectionCookie cookie;
-    int ss0, ss1, ss2, ss3;
-    int xx, yy, x, xmin, xmax;
-    INT32 *k, *kk;
+    int xx, yy, b, x, xmin, xmax, ss;
+    UINT8 *out, *in;
+    INT32 *k;
```
radarhere, line 291:
```suggestion
    INT32 *k, *kk;
```

src/libImaging/Resample.c
```diff
@@ -286,88 +286,30 @@ void
...
 
     // use the same buffer for normalized coefficients
-    kk = (INT32 *)prekk;
+    const INT32 *kk = (INT32 *)prekk;
```
radarhere, line 294:
```suggestion
    kk = (INT32 *)prekk;
```
Compiling, I get `warning: initializing 'unsigned char *' with an expression of type 'const unsigned char *' discards qualifiers`.
This fixes it.

src/libImaging/Resample.c
```diff
@@ -378,79 +320,30 @@ void
 ImagingResampleVertical_8bpc(
     Imaging imOut, Imaging imIn, int offset, int ksize, int *bounds, double *prekk) {
     ImagingSectionCookie cookie;
-    int ss0, ss1, ss2, ss3;
-    int xx, yy, y, ymin, ymax;
-    INT32 *k, *kk;
+    int xx, yy, b, y, ymin, ymax, ss;
+    UINT8 *out, *in;
+    INT32 *k;
```
radarhere, line 325:
```suggestion
    INT32 *k, *kk;
```

src/libImaging/Resample.c
```diff
@@ -378,79 +320,30 @@ void
...
 
     // use the same buffer for normalized coefficients
-    kk = (INT32 *)prekk;
+    const INT32 *kk = (INT32 *)prekk;
```
radarhere, line 328:
```suggestion
    kk = (INT32 *)prekk;
```

src/libImaging/Unpack.c
```diff
@@ -413,24 +413,14 @@ unpackL4IR(UINT8 *out, const UINT8 *in, int pixels) {
...
-unpackLAL(UINT8 *_out, const UINT8 *in, int pixels) {
+unpackLAL(UINT8 *out, const UINT8 *in, int pixels) {
     int i;
+    UINT8 *in1 = in;
+    UINT8 *in2 = in + pixels;
     /* LA, line interleaved */
-    for (i = 0; i < pixels; i++, _out += 4) {
-        UINT32 iv = MAKE_UINT32(in[i], in[i], in[i], in[i + pixels]);
-        memcpy(_out, &iv, sizeof(iv));
+    for (i = 0; i < pixels; ++i) {
```
radarhere, line 421:
```suggestion
    for (i = 0; i < pixels; i++) {
```
Just to be consistent with the other loops.

Tests/test_image_transform.py
Yay295: No, changing the size of the pixels also changed what they get hashed to, so the order of the colors isn't the same. We can ignore the order by comparing them as a set, or sorting both lists before comparing them. All of the colors should be unique so I decided to compare them as a set.

src/libImaging/Resample.c
Yay295: I don't see that warning, but I made `k` `const` as well which should also fix this issue.

Tests/test_image_transform.py
Yay295: Sorting is used in this other test https://github.com/python-pillow/Pillow/blob/a877d7dab2c4c6701d20520187dca94e496b2e7b/Tests/test_image.py#L341-L342 but I don't see that being done anywhere else.

src/libImaging/Unpack.c
```diff
@@ -413,24 +413,14 @@ unpackL4IR(UINT8 *out, const UINT8 *in, int pixels) {
...
-        _out += 4;
-    }
-}
-
-static void
-unpackLAL(UINT8 *_out, const UINT8 *in, int pixels) {
+unpackLAL(UINT8 *out, const UINT8 *in, int pixels) {
     int i;
+    UINT8 *in1 = in;
+    UINT8 *in2 = in + pixels;
```
radarhere, line 419:
```suggestion
    const UINT8 *in1 = in;
    const UINT8 *in2 = in + pixels;
```
Fixes
```
warning: initializing 'unsigned char *' with an expression of type 'const unsigned char *' discards qualifiers [-Wincompatible-pointer-types-discards-qualifiers]
```

src/libImaging/Geometry.c
```diff
@@ -898,35 +783,28 @@ ImagingScaleAffine(
...
+            }
+        }
+        yo += a[4];
     }
-
     ImagingSectionLeave(&cookie);
 
-#undef AFFINE_SCALE
-
     free(xintab);
```
radarhere, line 808: To be clear, changes like [this](https://github.com/python-pillow/Pillow/commit/5b94bf1ce863fa7d319e2fe6c8f7505f6bb5ab45) in your PR have nothing to do the fact that you've changed the pixelsize of LA/La/PA, right? You're just tidying up while you're in this general area?

In this case, your change casts the image32 pointers into UINT8 instead of INT32, resulting in this new warning when I compile it.
```
src/libImaging/Geometry.c:983:50: warning: cast to 'unsigned char *' from smaller integer type 'char' [-Wint-to-pointer-cast]
                        out[x * pixelsize + b] = (UINT8 *)imIn->image[yin][xin * pixelsize + b];
                                                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/libImaging/Geometry.c:983:48: warning: incompatible pointer to integer conversion assigning to 'unsigned char' from 'unsigned char *'; dereference with * [-Wint-conversion]
                        out[x * pixelsize + b] = (UINT8 *)imIn->image[yin][xin * pixelsize + b];
                                               ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Yay295: No, there is a change here. Previously it only supported `UINT8` and `INT32`, neither of which is `UINT16`. I changed it so that it always uses `UINT8` now, and uses the image pixel size to know how many bytes to copy. The diff doesn't line things up correctly, but compare for example:
`memset(out + x0, 0, (x1 - x0) * sizeof(pixel));`
`memset(out + x0 * pixelsize, 0, (x1 - x0) * pixelsize);`
Since it always uses `UINT8` now, it didn't need to be a macro. That's unfortunately what confuses the diff.

That error is valid though; I just pushed a fix.

Yay295:
Closing as I believe #6547 is a better way to go, and this change will be much easier if that change is made.

PR #6548, opened 2022-08-29
Fixed GIF remapping to palette with duplicate entries

radarhere:
Addresses #6389

`remap_palette` cannot map multiple palette indexes to the same output index, by its very definition. It accepts a list of palette indexes that it transforms to the position in the list. There cannot be more than one palette index at a given position in the list.

If a palette is supplied when saving a GIF image, `im.save("out.gif", palette=...)`, and that palette has duplicate entries, then the following GifImagePlugin code will try to prepare a list that maps the first of the duplicated entries to multiple locations, leaving the rest of the duplicate entries to be not strictly defined in the behaviour of `remap_palette`.
https://github.com/python-pillow/Pillow/blob/ad7be550aa539dc4c79da928dad071cf0150b731/src/PIL/GifImagePlugin.py#L519-L526

This PR fixes that, by leaving the duplicated indexes as they are - setting `index` to `None`, the next loop in GifImagePlugin
https://github.com/python-pillow/Pillow/blob/ad7be550aa539dc4c79da928dad071cf0150b731/src/PIL/GifImagePlugin.py#L527-L532

will tell `remap_palette` to map the index to itself, leaving it untouched.

### DIFF
--- a/src/PIL/GifImagePlugin.py
+++ b/src/PIL/GifImagePlugin.py
@@ -523,6 +523,8 @@ def _normalize_palette(im, palette, info):
                 index = im.palette.colors[source_color]
             except KeyError:
                 index = None
+            if index in used_palette_colors:
+                index = None
             used_palette_colors.append(index)
         for i, index in enumerate(used_palette_colors):
             if index is None:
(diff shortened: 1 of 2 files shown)


src/PIL/GifImagePlugin.py
```diff
@@ -523,6 +523,8 @@ def _normalize_palette(im, palette, info):
                 index = im.palette.colors[source_color]
             except KeyError:
                 index = None
+            if index in used_palette_colors:
+                index = None
```
Yay295, line 527: I think the `if` could be inside the `try`, since if there's an exception it will have already been set to `None`.

radarhere: I actually have a different idea of how to simplify the code. In my latest commit, I replace the `try` block with `get()`.

PR #6549, opened 2022-08-30
Removed support for Python versions before interpaddr()

radarhere:
https://github.com/python-pillow/Pillow/blob/ad7be550aa539dc4c79da928dad071cf0150b731/src/_imagingtk.c#L26-L28

I saw "for new versions", and wondered how long this message has been there. It turns out, it has been there since the fork from PIL.

The code is used when ImageTk catches an AttributeError, which has also been present since the fork.

https://github.com/python-pillow/Pillow/blob/ad7be550aa539dc4c79da928dad071cf0150b731/src/PIL/ImageTk.py#L71-L85

`interpaddr()` added in Python 1.5.2. The last version of PIL supported Python 1.5.2, so it seems likely that the AttributeError catch was there for Python versions < 1.5.2.

We no longer support such early versions of Python, so this code can be removed.

### DIFF
--- a/src/PIL/ImageTk.py
+++ b/src/PIL/ImageTk.py
@@ -68,21 +68,18 @@ def _pyimagingtkcall(command, photo, id):
         # may raise an error if it cannot attach to Tkinter
         from . import _imagingtk
 
-        try:
-            if hasattr(tk, "interp"):
-                # Required for PyPy, which always has CFFI installed
-                from cffi import FFI
+        if hasattr(tk, "interp"):
+            # Required for PyPy, which always has CFFI installed
+            from cffi import FFI
 
-                ffi = FFI()
+            ffi = FFI()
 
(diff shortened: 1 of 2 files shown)


hugovk:
Wow! Here "new versions" means since April 1999, 23 years ago!

Here's where `interpaddr()` was committed to CPython on 29 Apr 1998:

* https://github.com/python/cpython/commit/9d1b7ae65b103525a2f28ebcb43d2c580a04d2d4

And added to the 1.5.2 NEWS file on 11 Aug 1998:

* https://github.com/python/cpython/commit/ab9d6f0be97d77b167dd14d99eae622023354554

And 1.5.2 was released on 13 April 1999:

* https://www.python.org/download/releases/1.5/

I think we can safely drop this without a deprecation warning :)

PR #6551, opened 2022-08-30
Removed support for PyPy before Python 3.6

nulano:
Continuing from https://github.com/python-pillow/Pillow/pull/6549, this PR removes support for old PyPy.

The function `interpaddr` was added to PyPy over 4 years ago: https://foss.heptapod.net/pypy/pypy/-/commit/8f15f623274f84bae5a4c3be914abb96b604a254

From the tags on that commit, I can see it was included in `release-pypy3.6-v7.0.0`, the [first Python 3.6 PyPy release](https://doc.pypy.org/en/latest/release-v7.0.0.html).
Since Python 3.5 is no longer supported by Pillow, the old code can be removed without deprecation.

### DIFF
--- a/src/PIL/ImageTk.py
+++ b/src/PIL/ImageTk.py
@@ -68,18 +68,7 @@ def _pyimagingtkcall(command, photo, id):
         # may raise an error if it cannot attach to Tkinter
         from . import _imagingtk
 
-        if hasattr(tk, "interp"):
-            # Required for PyPy, which always has CFFI installed
-            from cffi import FFI
-
-            ffi = FFI()
-
-            # PyPy is using an FFI CDATA element
-            # (Pdb) self.tk.interp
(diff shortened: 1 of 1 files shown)


hugovk:
👍 

There's also some PyPy3.5 tags on that commit.

And to look at it from the other angle: Python 3.7 is the oldest version we support.

[PyPy3.7 stable was released on 2021-04-08](https://www.pypy.org/posts/2021/04/pypy-v734-release-of-python-27-and-37.html), a good few years after the 2017-11-05 commit.

(We don't support pre-release PyPy, but [PyPy3.7 alpha was 2020-09-25](https://www.pypy.org/posts/2020/09/pypy-732-triple-release-python-27-36-3980901335490872787.html) and [beta was 2020-11-21](https://www.pypy.org/posts/2020/11/pypy-733-triple-release-python-37-36-3446596804408262749.html), also both long after the commit.)

Issue #6482, opened 2022-08-05
Size of multiple raw formats (.dng, .NEF) is reported incorrectly


gverm:
### What did you do?
I opened the images using Pillow and printed their detected size.

### What did you expect to happen?
The size to be reported correctly. This causes conversions to JPEG to be way too small.

### What actually happened?
Size was reported incorrectly.

### What are your OS, Python and Pillow versions?

* OS: EndeavourOS
* Python: 3.10
* Pillow: 9.2.0 (Also tested -git as of 05/08/22)

```python
>>> from PIL import Image
>>> img = Image.open("sample.dng")
>>> img.width, img.height
(256, 195)
>>> img = Image.open("sample2.NEF")
>>> img.width, img.height
(160, 120)
```

Interestingly the `file` command also shows the wrong sizes:
```
$ file sample.dng
sample.dng: TIFF image data, little-endian, direntries=60, height=195, bps=734, compression=none, PhotometricIntepretation=RGB, manufacturer=NIKON CORPORATION, model=NIKON Z 7, orientation=upper-right, width=256

$ file sample2.NEF
sample2.NEF: TIFF image data, little-endian, direntries=28, height=120, bps=352, compression=none, PhotometricIntepretation=RGB, manufacturer=NIKON CORPORATION, model=NIKON D7200, orientation=upper-left, width=160
```
While inspecting the files with Dolphin shows the correct sizes:
![image](https://user-images.githubusercontent.com/8126341/183224642-8fee460a-a29d-4155-bb1a-c0ad4325be39.png)

radarhere:
When I inspect the image with `exiftool -v`, I see that sample2.NEF has an image that is 160x120.

However, I also see that it has [child images](https://www.awaresystems.be/imaging/tiff/tifftags/subifds.html) - two reduced resolution images, called JpgFromRaw and OtherImage, and one full resolution image that is 6016x4016.

So I don't know that Pillow is necessarily doing anything wrong. Would you like to instead request that Pillow provide an API to access those child images?

gverm:
> However, I also see that it has [child images](https://www.awaresystems.be/imaging/tiff/tifftags/subifds.html) - two reduced resolution images, called JpgFromRaw and OtherImage, and one full resolution image that is 6016x4016.

That's interesting, I didn't know about that, thanks for explaning.

> So I don't know that Pillow is necessarily doing anything wrong. Would you like to instead request that Pillow provide an API to access those child images?

That would be great.

[...]

PR #6569, opened 2022-09-10
Added reading of TIFF child images

radarhere:
Helps #6482 by adding reading of TIFF [child images](https://www.awaresystems.be/imaging/tiff/tifftags/subifds.html) - a series of offsets to child IFDs.

There are two test files here.
- One was created using tifffile, with https://forum.image.sc/t/tifffile-ome-tiff-generation-is-taking-too-much-ram/41865/9 as a basis. For that image we can just seek to the child IFD and load the TIFF image.
- The other test image uses the outdated [JPEGInterchangeFormat](https://www.awaresystems.be/imaging/tiff/tifftags/jpeginterchangeformat.html). Because ["No new TIFF writer code should ever attempt to use this tag"](https://www.awaresystems.be/imaging/tiff/tifftags/jpeginterchangeformat.html), I didn't think I was going to find a way to create a small file in this format, and instead simply appended a JPEG to a TIFF file with Pillow, setting the offset to the JPEG image accordingly.


src/PIL/TiffImagePlugin.py
```diff
@@ -1148,6 +1148,39 @@ def tell(self):
...
 
+    def get_child_images(self):
+        if SUBIFD not in self.tag_v2:
+            return []
+        child_images = []
+        exif = self.getexif()
+        offset = None
+        for im_offset in self.tag_v2[SUBIFD]:
+            # reset buffered io handle in case fp
+            # was passed to libtiff, invalidating the buffer
```
radarhere, line 1159: Copied from https://github.com/python-pillow/Pillow/blob/d843759ca9e0295032c5123508798cfdc9aaf6ce/src/PIL/TiffImagePlugin.py#L1113-L1115

PR #6500, opened 2022-08-13
Corrected BMP and TGA palette size when saving

radarhere:
Resolves #6494
Resolves #6572

#6060 contained a commit entitled "[Allow getpalette() to return less than 256 colors](https://github.com/python-pillow/Pillow/pull/6060/commits/948c064b282f80492f5b88d3f06a599b76516edf)"

That meant that when saving BMP and TGA images, the palette might be smaller than 256 colors. However, the plugins always specify the length of the palette as 256 for P images.

This PR changes them to match the length of the palette.

PR #6573, opened 2022-09-12
Corrected BMP palette size when saving

### LINKED ISSUE #6572: Invalid 8-bit BMP files written when not all palette colours are used
### What did you do?

I loaded an image (BMP or PNG) in 8-bit format. (i.e. with a palette of up to 256 colours)
The image did use a palette with _less than_ 256 colours.

Then I edited the image and saved it again, as 8-bit BMP.

EDIT: This may be a similar problem as #6500.

The saved image should be viewable by Windows' default image viewer.

### What actually happened?

The saved image was corrupt and could not be displayed.

Reason: The offset calculation for the `bfSize` (file offset 0x02) and `bfOffBits` (file offset 0x0A) assumes that the palette is 256 colours large (0x400 bytes). However there are only 0x200 bytes written for the palette. (4 bytes * 128 colours)

The fields `biClrUsed` (file offset 0x2E) and `biClrImportant` (file offset 0x32) are also incorrectly set to 256 instead of 128.

* OS: Windows 10 1909
* Python: Python 3.7.6
* Pillow: 9.2.0

I attached an archive with two tests: [bmp-tests.zip](https://github.com/python-pillow/Pillow/files/9551375/bmp-tests.zip)
- Test 1: BMP image with 256 colours being used (works fine)
- Test 2: BMP image with 128 colours being used (broken after resaving)

```python
import PIL.Image

with PIL.Image.open("Test1_AllColors.bmp") as img1:
    img1.save("t1-good.bmp")

with PIL.Image.open("Test2_128Colors.bmp") as img2:
    img2.save("t2-bad.bmp")
```

radarhere:
Resolves #6572

#6060 contained a commit entitled "[Allow getpalette() to return less than 256 colors](https://github.com/python-pillow/Pillow/pull/6060/commits/948c064b282f80492f5b88d3f06a599b76516edf)"

That meant that when saving BMP images, the palette might be smaller than 256 colors. However, BmpImagePlugin always specifics the number of colors in the palette as 256 for P images.

This PR changes it to match the length of the palette.

### DIFF
--- a/Tests/test_file_bmp.py
+++ b/Tests/test_file_bmp.py
@@ -58,6 +58,18 @@ def test_save_to_bytes():
         assert reloaded.format == "BMP"
 
 
+def test_small_palette(tmp_path):
+    im = Image.new("P", (1, 1))
+    colors = [0, 0, 0, 125, 125, 125, 255, 255, 255]
+    im.putpalette(colors)
+
+    out = str(tmp_path / "temp.bmp")
+    im.save(out)
+
+    with Image.open(out) as reloaded:
+        assert reloaded.getpalette() == colors
+
+
 def test_save_too_large(tmp_path):
     outfile = str(tmp_path / "temp.bmp")
     with Image.new("RGB", (1, 1)) as im:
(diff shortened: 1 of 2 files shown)


radarhere:
I've made this part of #6500

PR #6576, opened 2022-09-13
Fixed seeking to an L frame in a GIF

### LINKED ISSUE #6571: Exception using "seek" on GIF that contains frames with different modes
I have an animated GIF that has about 60 frames, some of the frames are encoded in 'L' mode and others are in 'RGB' mode. Many actions, including a simple "seek" command" fail with a "ValueError: images do not match" exception when `self.im.paste(frame_im, self.dispose_extent)` is issued in GifImagePlugin.py. This happens when "self.im" and "frame_im" have different modes.

This might be a protection technique for not being able to process such files with Pillow.

radarhere:
Resolves #6571

When pasting an L frame onto an RGB(A) GIF, convert the frame to RGB(A) first.

The test image was created using a modified Pillow, so that there is no global palette, and the image is comprised of a frame with a palette, and then two frames without. This means that the first frame is P mode, the second is L mode, combining to make RGB, and then there is an L mode frame pasted onto that RGB frame.


radarhere:
I've added a commit to address https://github.com/python-pillow/Pillow/pull/6610#issuecomment-1267310192

That comment found a GIF where on a later frame, the palette is "not needed" because the frame is grayscale. The problem is that Pillow then concludes that since there isn't a palette, it should fallback to the global palette.

https://github.com/python-pillow/Pillow/blob/243402e78e2bf2b7af478c6d891816e372b5c3f9/src/PIL/GifImagePlugin.py#L275-L301

This is incorrect, as the frame is specifying custom palette behaviour, it's just that custom behaviour is to have a grayscale palette.

PR #6585, opened 2022-09-16
Remove duplicate parameter EXIF from JPEG

melonmouse:
Small improvement of the docs. Have a nice day :)

### DIFF
--- a/docs/handbook/image-file-formats.rst
+++ b/docs/handbook/image-file-formats.rst
@@ -376,7 +376,7 @@ The :py:meth:`~PIL.Image.open` method may set the following
     The ICC color profile for the image.
 
 **exif**
-    Raw EXIF data from the image.
+    If present, the image will be stored with the provided raw EXIF data.
 
 **comment**
     A comment about the image.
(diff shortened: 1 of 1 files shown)


radarhere:
The first mention of EXIF is listed under `The open() method may set the following info properties if available:`
The second is listed under `The save() method supports the following options:`

This is not a duplication. It is mentioned twice because it can be used in two different ways.

melonmouse:
Ah thanks for the clarification! I guess I read over those because "The save/load method supports the following options:" looks to me to be a part of the header above/below it - rather than being their semantic header. Would it make sense to elevate those to headers with a fontsize in between the parameter names and the image format header?

radarhere:
```
The :py:meth:`~PIL.Image.Image.save` method supports the following options:
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
renders as a `<h4>`.

Screen Shot 2022-09-16 at 7 53 17 pm


I personally feel like that's a bit much?

This also turns it into a `<section>` and adds an anchor tag.

melonmouse:
How about the following? I think the thing you propose is already a big win for readability of the docs, but maybe this is a bit neater.

```
The :py:meth:`~PIL.Image.Image.save` method
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

The :py:meth:`~PIL.Image.Image.save` method supports the following options:
```

radarhere:
What do you think of PR #6632 / https://pillow--6632.org.readthedocs.build/en/6632/handbook/image-file-formats.html#jpeg?

radarhere:
#6632 has now been merged.

Issue #6537, opened 2022-08-25
API changes for Resampling modes


99991:
With Pillow 9.2.0, I get the following deprecation notice:

```
DeprecationWarning: BILINEAR is deprecated and will be removed in Pillow 10 (2023-07-01). Use Resampling.BILINEAR instead.
```

Is it really necessary to introduce this breaking API change? Can the old names not be kept for backwards compatibility? Thousands of repositories will need to update their code, wasting countless man-hours.

You can get a rough idea on how many repositories are affected from the following two websites, which search a subset of GitHub repositories:

* https://grep.app/search?q=Image.BILINEAR
* https://sourcegraph.com/search?q=context:global+Image.BILINEAR&patternType=standard

```python
from PIL import Image
img = Image.new("RGB", (5, 5))
img.resize((7, 7), Image.BILINEAR)
```

radarhere:
To provide context, the deprecations being discussed were added in #5954. The intention was not to make an arbitrary change, but to make use of a newish Python language feature, enums.

https://github.com/python-pillow/Pillow/pull/5954#issuecomment-1009970094
> Do we keep the old constants "forever"?
>
> Or do we want to deprecate the old constants and remove them in Pillow 10 (15 months' deprecation period)?

radarhere:
I think in the scheme of things, this is a very minor change for users to make when upgrading to a new version. In Javascript programming, I'm used to finding old npm packages abandoned and replaced with new ones, and in iOS programming, I'd be surprised if Apple went a year without changing any constants.

If your argument is that Pillow is extremely widely used, I'll try and use [NumPy](https://github.com/numpy/numpy) as a point of reference. That is another Python library, downloaded twice as much as Pillow. In NumPy 1.20.0, they [deprecated a series of aliases](https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations).

https://grep.app/search?q=np.int seems to show that at least 20 times the amount of code is affected by that compared to https://grep.app/search?q=Image.BILINEAR.

I make this comment not in the interest of taking a stance, but in the interest of giving you a quick response.

99991:
> In NumPy 1.20.0, they [deprecated a series of aliases](https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations).

I believe that the argument for deprecating inbuilt types in NumPy is a bit stronger since it was know for a long time to be an actual problem https://github.com/numpy/numpy/pull/6103

[...]

PR #6586, opened 2022-09-16
Add inverse mappings for exif tags

melonmouse:
A nicer solution would be to switch to an enum with the right values, but this is the quick win.

Tested locally, ran through pycodestyle. To me it seems the code is trivial enough to not warrant adding a unit test.

### DIFF
--- a/docs/reference/ExifTags.rst
+++ b/docs/reference/ExifTags.rst
@@ -4,25 +4,45 @@
 :py:mod:`~PIL.ExifTags` Module
 ==============================
 
-The :py:mod:`~PIL.ExifTags` module exposes two dictionaries which
+The :py:mod:`~PIL.ExifTags` module exposes four dictionaries which
 provide constants and clear-text names for various well-known EXIF tags.
 
 .. py:data:: TAGS
     :type: dict
 
-    The TAG dictionary maps 16-bit integer EXIF tag enumerations to
-    descriptive string names.  For instance:
+    The TAGS dictionary maps 16-bit integer EXIF tag enumerations to
+    descriptive string names. For instance:
 
         >>> from PIL.ExifTags import TAGS
         >>> TAGS[0x010e]
         'ImageDescription'
+        
+.. py:data:: TAG_CODES
+    :type: dict
+
+    The TAG_CODES dictionary maps descriptive string names to 16-bit integer EXIF
+    tag enumerations. For instance:
+
+        >>> from PIL.ExifTags import TAG_CODES
+        >>> TAG_CODES['ImageDescription']
+        0x010e
 
 .. py:data:: GPSTAGS
     :type: dict
 
     The GPSTAGS dictionary maps 8-bit integer EXIF gps enumerations to
-    descriptive string names.  For instance:
+    descriptive string names. For instance:
 
         >>> from PIL.ExifTags import GPSTAGS
         >>> GPSTAGS[20]
         'GPSDestLatitude'
+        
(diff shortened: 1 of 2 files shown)


docs/reference/ExifTags.rst
```diff
@@ -4,25 +4,45 @@
...
 
         >>> from PIL.ExifTags import GPSTAGS
         >>> GPSTAGS[20]
         'GPSDestLatitude'
+        
+.. py:data:: GPS_CODES
+    :type: dict
+
+    The GPS_CODES dictionary maps descriptive string names to 8-bit integer EXIF
+    gps enumerations. For instance:
```
radarhere, line 44:
```suggestion
    GPS enumerations. For instance:
```

radarhere:
Could you describe an example use case for this feature? The currently dictionary allows for the codes to be translated into something that can be understood by humans. What is your particular scenario where you want to go the other way?

radarhere:
> A nicer solution would be to switch to an enum with the right values, but this is the quick win.

Could you expand on this?

melonmouse:
Suppose a user wants to set an exif tag, I presume they would want to set the data using a human readable tag name.

melonmouse:
> > A nicer solution would be to switch to an enum with the right values, but this is the quick win.
> 
> Could you expand on this?

Using a dict to store these values feels hacky in the first place.

```python
from enum import IntEnum


class ExifTags(IntEnum):
    InteropIndex = 0x0001
    ProcessingSoftware = 0x000B
    [...]
```

Which gives us just the right access:
```python
$ExifTags(0x0001).name
InteropIndex

$ExifTags.InteropIndex.value
1
```

melonmouse:
To elaborate a bit more.

Currently:
```python
exif_data = Image.Exif()
exif_data[0x123456] = '2020:01:01 00:00:00'
image.save(filename, exif=exif_data.tobytes())
```

After this pullrequest:
```python
exif_data = Image.Exif()
exif_data[TAG_CODES['DateTimeOriginal']] = '2020:01:01 00:00:00'
image.save(filename, exif=exif_data.tobytes())
```

After bigger (breaking?) enum change:
```python
exif_data = Image.Exif()
exif_data[Exif.DateTimeOriginal] = '2020:01:01 00:00:00'
image.save(filename, exif=exif_data.tobytes())
```

(I did not test the above code, maybe I'm understanding how exif data is handled by PIL incorrectly)

radarhere:
> Here your IDE will autocomplete the tagname. May need to update the code of ExifTags a bit to do this right. Could provide a TAGS dict generated from the Enum for backward compatibility.

I'm in favour of this, but when we previously had this conversation in #5875, I implemented it, [it was suggested that the original attributes be deprecated](https://github.com/python-pillow/Pillow/pull/5954#issuecomment-1009970094), and now there's #6537 where a user doesn't like that we changed the API. So I might hold off on this for a moment.

melonmouse:
Whenever adding anything to an API, it is important to think about a potential future cost of deprecating it again.

Deprecation of the old approach would be costly indeed. It is always a balance between quality of a library and the amount of maintenance for existing users. It is never possible to please everyone, but we can try our best. For a change like the one suggested in this thread, backward compatibility should win. Adding the Enum in a backward compatible way is possible however. I'm sorry about not noticing the backward compatibility break in #5954.

In any case, thank you for your hard work on PIL!

radarhere:
I've created PR #6630 to use enums as a possible alternative to this.

radarhere:
#6630 has been merged.

PR #6596, opened 2022-09-18
Improved palette handling in ImageOps

### LINKED ISSUE #6595: Method ImageOps.pad doesn't copy palette of original image
I think i found a bug with `ImageOps.pad` function.

I have an image which mode is `P` and this image has a specific palette. So when you do pad for this image it will return an image that is displayed as completely black.

So this is origin image
spidy

And this image is after applying `ImageOps.pad`
pad_spidy

This images all have the same unique values of pixels - ```(array([0, 1, 2, 3, 4, 5], dtype=uint8),
 array([0, 1, 2, 3, 4, 5], dtype=uint8))```
 
 But when convert origin and pad image to rgb it will have different uniq value -  ```(array([0, 1, 2, 3, 4, 5], dtype=uint8),
array([  0,  13,  35,  80,  83,  98, 148, 150, 171, 177, 212, 227, 255],
       dtype=uint8))```

### What did you do?
Download spidy image. It is image with mode `P` and has a specific palette. Read this image via Pillow then use `ImageOps.pad` and you will see full black image.

I expect to see pad colorful spidy image.

### What actually happened?
In pad palette of origin image don't copy, so you see full black image after padding. 

* OS: Ubuntu 20.04
* Python: 3.9
* Pillow: 9.1.1

### Reproduce code:
```python
import numpy as np

with Image.open("spidy.png") as origin_image:
    plt.imshow(origin_image)
    plt.show()

pad_image = ImageOps.pad(origin_image, (224, 224))
pad_image.save('pad_spidy.png')
plt.imshow(pad_image)
plt.show()

pad_uniq = np.unique(np.asarray(pad_image))
origin_uniq = np.unique(np.asarray(origin_image))

print(pad_uniq, origin_uniq)

pad_rgb_image = pad_image.convert('RGB')
origin_rgb_image = origin_image.convert('RGB')

pad_uniq_val = np.unique(np.asarray(pad_rgb_image))
origin_uniq_val = np.unique(np.asarray(origin_rgb_image))

print(pad_uniq_val, origin_uniq_val)
```

PososikTeam:
Resolves #6595

This pull request fix `ImageOps.pad` method. In previous version of this method it doesn't copy the palette, and whats why it might change the image to full black image.

 * Add copy origin image palette to pad image.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,9 @@ Changelog (Pillow)
 9.3.0 (unreleased)
 ------------------
 
+- Corrected BMP and TGA palette size when saving #6500
+  [radarhere]
+
 - Do not call load() before draft() in Image.thumbnail #6539
   [radarhere]
 
--- a/src/PIL/ImageOps.py
+++ b/src/PIL/ImageOps.py
@@ -293,7 +293,7 @@ def pad(image, size, method=Image.Resampling.BICUBIC, color=None, centering=(0.5
         out = Image.new(image.mode, size, color)
         palette = image.palette.copy()
         if palette:
-            out.putpalette(palette.palette)
+            out.putpalette(palette)
         if resized.width != size[0]:
             x = int((size[0] - resized.width) * max(0, min(centering[0], 1)))
             out.paste(resized, (x, 0))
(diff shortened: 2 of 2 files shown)


PososikTeam:
Is i need to fix the tests? And why it's fail? I think my change is too small to not pass tests.

radarhere (approved):

radarhere:
https://github.com/PososikTeam/Pillow/pull/1 has been merged, containing suggestions from me
- Removing the `copy()` operation, as it is not necessary, and not all images have palettes
- Switched to using `getpalette()` in both `pad()` and `expand()`, allowing the test I added to pass
- Stop `expand()` from only copying the palette for P images, allowing for the PA mode as well

PR #6606, opened 2022-09-21
Expand .ani and .cur functionality

jlwoolf:
Currently the BmpImagePlugin checks if a file is a .cur file by checking if the bits are 32 and a the header is 22. However, the header for cursor files is not guaranteed to be 22.

The CurImagePlugin selects the largest cursor image from the file. The largest cursor does not have to be at the front of the file. So the start of the DIB header file varies depending on the file. 

The fix below initializes a parameter in the CurImagePlugin (if it is in fact a .cur file) so that transparency is considered. It might be worth modifying CurImagePlugin to work similarly to IcoImagePlugin however. I'm looking into that right now, but this is a simple fix.


jlwoolf:
Turns out it was much easier to implement .cur file saving (as it shares a lot with ICO). So I remodified CurImagePlugin to add that functionality. I reverted the BmpImagePlugin changes.

jlwoolf:
Alright, made some significant modifications. I believe I have mostly resolved all the .cur file issues, but tests need to be written for a few. I have also further expanded this functionality to work with .ani files. You can now read and write them. I also plan to work on tests for them in the future.

src/PIL/CurImagePlugin.py
```diff
@@ -15,61 +15,209 @@
...
+                "nb_color": s[2],  # No. of colors in image (0 if >=8bpp)
+                "reserved": s[3],
+                "x_hotspot": i16(s, 4),
+                "y_hotspot": i16(s, 6),
+                "size": i32(s, 8),
+                "offset": i32(s, 12),
+            }
 
-        offset = self.fp.tell()
+            # See Wikipedia
```
radarhere, line 143: Could you append a link to a specific page?

Tests/test_file_cur.py
```diff
@@ -1,30 +1,133 @@
...
 
 from PIL import CurImagePlugin, Image
 
-TEST_FILE = "Tests/images/deerstalker.cur"
-
 
-def test_sanity():
-    with Image.open(TEST_FILE) as im:
+def test_deerstalker():
+    with Image.open("Tests/images/cur/deerstalker.cur") as im:
```
radarhere, line 9: Any particular reason for this change?

src/PIL/CurImagePlugin.py
```diff
@@ -15,61 +15,209 @@
...
 
 def _accept(prefix):
-    return prefix[:4] == b"\0\0\2\0"
+    return prefix[:4] == _MAGIC
 
 
 ##
 # Image plugin for Windows Cursor files.
+class CurFile(IcoImagePlugin.IcoFile):
+    def __init__(self, buf: BytesIO):
```
radarhere, line 112:
```suggestion
    def __init__(self, buf: BytesIO()):
```
Fixes https://github.com/python-pillow/Pillow/actions/runs/3138116114/jobs/5099677056#step:11:233
> /opt/hostedtoolcache/Python/3.10.7/x64/lib/python3.10/site-packages/PIL/CurImagePlugin.py:docstring of PIL.CurImagePlugin.CurFile:1: WARNING: py:class reference target not found: _io.BytesIO

Tests/test_file_cur.py
jlwoolf: I moved all the .cur files into a folder so that the Images folder was not cluttered by the ones I planned to add. At the same time I decided to rename the test since I was about to test a few more cursors (from windows 98 and up)

jlwoolf:
@radarhere 
Got busy and haven't had a chance to check up on this until. This repository is quite large with quite a few checks (understandably) so I am unsure what problems I need to fix for this merger. Any guidance would be much appreciated.

akx:
Oh, this apparently seems to be similar to my work in #6965 and https://github.com/akx/Pillow/tree/cur-support...

src/PIL/AniImagePlugin.py
```diff
@@ -0,0 +1,446 @@
...
+
+from PIL import BmpImagePlugin, CurImagePlugin, Image, ImageFile
+from PIL._binary import i32le as i32
+from PIL._binary import o8
+from PIL._binary import o16le as o16
+from PIL._binary import o32le as o32
+
+
+def _accept(prefix: bytes) -> bool:
+    return prefix[:4] == b"RIFF"
```
radarhere, line 15:
```suggestion
    return prefix.startswith(b"RIFF")
```
See #8761

PR #6610, opened 2022-09-23
Consider all frames when selecting mode for PNG save_all

### LINKED ISSUE #6608: Image.show() defaults to PNG ignoring animated GIFs
When using show() for an animated GIF with version 9.2.0, Pillow converts the image to PNG by default thus eliminating the animation.

I suggest the following change in "ImageShow.py" to preserve the animation:

```python
def save_image(self, image):
    """Save to temporary file and return filename."""

    if image.format == 'GIF' and image.is_animated:
        suffix = "." + image.format
        f, filename = tempfile.mkstemp(suffix)
        os.close(f)

        image.save(filename, save_all=True)

        return filename
    else:
        image_format = self.get_format(image)
        return image._dump(format=image_format, **self.options)
```

radarhere:
Resolves #6608

At the moment, when saving multiple frames with PngImagePlugin, the mode of the first image is used.
https://github.com/python-pillow/Pillow/blob/af8260ee5505642ff7d97ff7d53660ca9a609cab/src/PIL/PngImagePlugin.py#L1224

However, this is not necessarily ideal. For a GIF, the first frame might be P, and subsequent frames might be RGB. Converting the RGB frames to P could result in loss of information.

This PR loops through the frames to be saved, checks if any are RGB, RGBA or P, and upgrades the mode accordingly.

### DIFF
--- a/Tests/test_file_apng.py
+++ b/Tests/test_file_apng.py
@@ -648,6 +648,16 @@ def test_seek_after_close():
         im.seek(0)
 
 
+@pytest.mark.parametrize("mode", ("RGBA", "RGB", "P"))
+def test_different_modes_in_later_frames(mode, tmp_path):
+    test_file = str(tmp_path / "temp.png")
+
+    im = Image.new("L", (1, 1))
+    im.save(test_file, save_all=True, append_images=[Image.new(mode, (1, 1))])
+    with Image.open(test_file) as reloaded:
+        assert reloaded.mode == mode
+
+
 def test_constants_deprecation():
     for enum, prefix in {
         PngImagePlugin.Disposal: "APNG_DISPOSE_",
--- a/src/PIL/PngImagePlugin.py
+++ b/src/PIL/PngImagePlugin.py
@@ -1089,28 +1089,28 @@ class _fdat:
         self.seq_num += 1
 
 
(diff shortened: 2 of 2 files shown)


seidnerj:
This fix seems to not be working correctly for some files. See the attached animated GIF and the resulting animated PNG. Some of the frames in the PNG file are corrupted.

bB4Y4h
bB4Y4h

seidnerj:
See attached a corrupted frame and the corresponding original frame (albeit with fewer colors than in the original animated GIF since there are only 256 available in a single frame).

original frame
corrupted frame

radarhere:
That actually isn't a problem related to this PR - you've found a GIF that Pillow isn't reading correctly in the first place.

Run this code over your GIF, and you will see what I'm talking about.
```python
from PIL import Image
im = Image.open("in.gif")
im.seek(3)
im.save("out.gif")
```

I'll investigate and see if I can put together a fix in a new PR.

seidnerj:
Just my luck, running into 4 separate issues all with the same set if GIFs 🤦‍♂️

Thanks a lot!

radarhere:
I've added a commit to #6576 to fix your GIF.

seidnerj:
Many thanks!

PR #6611, opened 2022-09-23
Show all frames in ImageShow

radarhere:
Helps #6608

The issue requests that multiple frames be displayed when using ImageShow. This PR adds the `save_all` argument.

### DIFF
--- a/src/PIL/ImageShow.py
+++ b/src/PIL/ImageShow.py
@@ -136,7 +136,7 @@ class WindowsViewer(Viewer):
     """The default viewer on Windows is the default system application for PNG files."""
 
     format = "PNG"
-    options = {"compress_level": 1}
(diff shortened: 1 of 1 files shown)


src/PIL/ImageShow.py
```diff
@@ -136,7 +136,7 @@ class WindowsViewer(Viewer):
     """The default viewer on Windows is the default system application for PNG files."""
 
     format = "PNG"
-    options = {"compress_level": 1}
+    options = {"compress_level": 1, "save_all": True}
```
nulano, line 139: Photos, the default Windows viewer for images, does not support APNG and only shows the first frame. However, users can change their default viewer to one that does support APNG, so this is still a worthwhile change.

PR #6612, opened 2022-09-23
Ensure dependencies are installed when building docs

radarhere:
Moves the sphinx requirements from https://github.com/python-pillow/Pillow/blob/af8260ee5505642ff7d97ff7d53660ca9a609cab/.github/workflows/test.yml#L102 to https://github.com/python-pillow/Pillow/blob/af8260ee5505642ff7d97ff7d53660ca9a609cab/docs/Makefile#L45-L47 so that `make doc` can ensure that all of the dependencies are installed.

Also ensures that Pillow and olefile are installed, two more requirements for building the docs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,6 @@ coverage:
 .PHONY: doc
 doc:
 	python3 -c "import PIL" > /dev/null 2>&1 || python3 -m pip install .
-	python3 -c "import olefile" > /dev/null 2>&1 || python3 -m pip install olefile
 	$(MAKE) -C docs html
 
 .PHONY: doccheck
--- a/Makefile
+++ b/Makefile
@@ -17,11 +17,13 @@ coverage:
 
 .PHONY: doc
 doc:
+	python3 -c "import PIL" > /dev/null 2>&1 || python3 -m pip install .
+	python3 -c "import olefile" > /dev/null 2>&1 || python3 -m pip install olefile
 	$(MAKE) -C docs html
 
 .PHONY: doccheck
 doccheck:
-	$(MAKE) -C docs html
+	$(MAKE) doc
 # Don't make our tests rely on the links in the docs being up every single build.
 # We don't control them.  But do check, and update them to the target of their redirects.
 	$(MAKE) -C docs linkcheck || true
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -17,11 +17,13 @@ coverage:
 
 .PHONY: doc
 doc:
+	python3 -c "import PIL" > /dev/null 2>&1 || python3 -m pip install .
+	python3 -c "import olefile" > /dev/null 2>&1 || python3 -m pip install olefile
```
hugovk, line 21: We're building Pillow anyway, shall we just use the docs extra?

```suggestion
	python3 -c "import PIL" > /dev/null 2>&1 || python3 -m pip install ".[docs]"
```

docs/Makefile
```diff
@@ -44,6 +44,10 @@ clean:
 
 install-sphinx:
 	$(PYTHON) -c "import sphinx" > /dev/null 2>&1 || $(PYTHON) -m pip install sphinx
+	$(PYTHON) -c "import sphinx_copybutton" > /dev/null 2>&1 || $(PYTHON) -m pip install sphinx-copybutton
+	$(PYTHON) -c "import sphinx_issues" > /dev/null 2>&1 || $(PYTHON) -m pip install sphinx-issues
+	$(PYTHON) -c "import sphinx_removed_in" > /dev/null 2>&1 || $(PYTHON) -m pip install sphinx-removed-in
+	$(PYTHON) -c "import sphinxext_opengraph" > /dev/null 2>&1 || $(PYTHON) -m pip install sphinxext-opengraph
 	$(PYTHON) -c "import furo" > /dev/null 2>&1 || $(PYTHON) -m pip install furo
```
hugovk, line 51: Rather than invoking the interpreter 6 times, shall we just do a single (quiet) pip install?
```suggestion
	$(PYTHON) -m pip install --quiet sphinx sphinx-copybutton sphinx-issues sphinx-removed-in sphinxext-opengraph furo
```

Makefile
radarhere: If Pillow was already installed, but olefile wasn't, your version wouldn't install it. Unless you'd like to add olefile to the docs/Makefile install command?

hugovk: Yes, makes more sense to move it to docs/Makefile.

hugovk (approved):

PR #6621, opened 2022-09-26
Add "concurrency" to GitHub Actions

Yay295:
This will cancel outdated jobs in favor of new jobs. So if you push some changes, and then push some more changes while the first changes are still being tested, the first tests will be cancelled in favor of the new tests.

Also if you reset to the previous commit and force push that, it will actually un-cancel the previously cancelled tests for that commit.

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency
https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

### DIFF
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -10,6 +10,10 @@ on:
 permissions:
   contents: read
 
+concurrency: 
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   update_release_draft:
     permissions:
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -8,6 +8,10 @@ on:
 permissions:
   issues: write
 
+concurrency: 
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   stale:
     if: github.repository_owner == 'python-pillow'
--- a/.github/workflows/tidelift.yml
+++ b/.github/workflows/tidelift.yml
@@ -1,4 +1,5 @@
 name: Tidelift Align
+
 on:
   schedule:
     - cron: "30 2 * * *"  # daily at 02:30 UTC
@@ -14,6 +15,10 @@ on:
 
 permissions:
   contents: read
+
+concurrency: 
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
 
 jobs:
   build:
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -14,6 +14,10 @@ on:
 permissions:
   contents: read
 
+concurrency: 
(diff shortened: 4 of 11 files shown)


hugovk (approved):
I'm good with this, we just added this to the CI at work.

We have something similar with [Travis CI](https://app.travis-ci.com/github/python-pillow/pillow-wheels/settings):

image

And [AppVeyor](https://ci.appveyor.com/project/python-pillow/pillow/settings):

image

The key difference is already-running builds are not cancelled for Travis (and not configurable) or AppVeyor (configurable but we haven't done so).

We can allow already-running builds to continue with GHA using `cancel-in-progress: true` or omitting it.

But I'm tempted to let it cancel, we have a pretty big CI matrix nowadays.

.github/workflows/stale.yml
```diff
@@ -8,6 +8,10 @@ on:
 permissions:
   issues: write
 
+concurrency: 
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
```
radarhere, line 15:
```suggestion
jobs:
```
I don't think this is necessary, as stale only runs once a day.

.github/workflows/tidelift.yml
```diff
@@ -15,6 +16,10 @@ on:
 permissions:
   contents: read
 
+concurrency: 
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
```
radarhere, line 23:
```suggestion
jobs:
```
I don't think this is necessary, as Tidelift Align only runs once a day.

.github/workflows/release-drafter.yml
```diff
@@ -10,6 +10,10 @@ on:
 permissions:
   contents: read
 
+concurrency: 
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
```
radarhere, line 17:
```suggestion
jobs:
```
Feel free to give thoughts on this, but... I kind of don't want to cancel the release drafter?

hugovk: It doesn't really as it only takes about 15s, but it's safe to cancel as we only need it to run once: 

when it runs it looks at the current state of merged PRs and overwrites the previous draft.

.github/workflows/stale.yml
hugovk: This and `tidelift.yml` can also be triggered manually via `workflow_dispatch:`, so potentially could have more than one in the queue at once.

But I've no strong opinion either way.

.github/workflows/release-drafter.yml
hugovk: Although two different runs from two different merges will have different `${{ github.ref }}` so won't be cancellable.

However, manual triggers via `workflow_dispatch` could have the same `ref`. Again, no strong opinion either way for this.

PR #6625, opened 2022-09-29
Improve documentation for ImageDraw.rectangle and rounded_rectangle

### LINKED ISSUE #1668: It seemed that the box of Draw.rectangle contained the end xy point
I learned this from doc:

> xy – Four points to define the bounding box. Sequence of either [(x0, y0), (x1, y1)] or [x0, y0, x1, y1]. The second point is just outside the drawn rectangle.

Then I ran this script

``` python
bb=Image.new('RGB',(4,4),(255,255,255))
dr=ImageDraw.Draw(bb)
box=(0,0,2,2)
dr.rectangle(box, fill=(125,45,55))
for i in range(4):
    for j in range(4):
        print(i,j,bb.getpixel((i,j)))
cc=bb.crop(box)
print(bb.getcolors())
print(cc.getcolors())
```

The output was

``` python
0 0 (125, 45, 55)
0 1 (125, 45, 55)
0 2 (125, 45, 55)
0 3 (255, 255, 255)
1 0 (125, 45, 55)
1 1 (125, 45, 55)
1 2 (125, 45, 55)
1 3 (255, 255, 255)
2 0 (125, 45, 55)
2 1 (125, 45, 55)
2 2 (125, 45, 55)
2 3 (255, 255, 255)
3 0 (255, 255, 255)
3 1 (255, 255, 255)
3 2 (255, 255, 255)
3 3 (255, 255, 255)
[(7, (255, 255, 255)), (9, (125, 45, 55))]
[(4, (125, 45, 55))]
```

It drew 9 points, while I just wanted the first 4 points to be filled

d0sboots:
Fixes #1668

### DIFF
--- a/docs/reference/ImageDraw.rst
+++ b/docs/reference/ImageDraw.rst
@@ -285,8 +285,8 @@ Methods
     Draws a rectangle.
 
     :param xy: Two points to define the bounding box. Sequence of either
-            ``[(x0, y0), (x1, y1)]`` or ``[x0, y0, x1, y1]``. The second point
-            is just outside the drawn rectangle.
+            ``[(x0, y0), (x1, y1)]`` or ``[x0, y0, x1, y1]``. The bounding box
+            is inclusive of both endpoints.
     :param outline: Color to use for the outline.
     :param fill: Color to use for the fill.
     :param width: The line width, in pixels.
@@ -298,8 +298,8 @@ Methods
     Draws a rounded rectangle.
 
     :param xy: Two points to define the bounding box. Sequence of either
-            ``[(x0, y0), (x1, y1)]`` or ``[x0, y0, x1, y1]``. The second point
-            is just outside the drawn rectangle.
+            ``[(x0, y0), (x1, y1)]`` or ``[x0, y0, x1, y1]``. The bounding box
+            is inclusive of both endpoints.
     :param radius: Radius of the corners.
     :param outline: Color to use for the outline.
     :param fill: Color to use for the fill.


hugovk:
Thanks for the PR!

---

> The rectangle is inclusive of both endpoints.

Quick demo:

```python

im = Image.new("RGB", (8, 8))
draw = ImageDraw.Draw(im)

draw.rectangle([(1, 1), (3, 3)], outline="green")
draw.rectangle([(4, 4), (6, 6)], outline="blue")
im.show()
```

image

Zoomed:

image

✅ 

---

> :param outline: Color to use for the outline. Ignored if fill is set.

With both `outline` and `fill` (and bigger rectangles this time):

```python

im = Image.new("RGB", (8, 8))
draw = ImageDraw.Draw(im)

draw.rectangle([(1, 1), (3, 3)], fill="red", outline="green")
draw.rectangle([(4, 4), (6, 6)], fill="red", outline="blue")
im.show()
```

image

image

:x: `outline` is not ignored when `fill` is set.

---

> :param width: The line width, in pixels. Ignored if fill is set.

With both `width` and `outline`:

```python

im = Image.new("RGB", (12, 12))
draw = ImageDraw.Draw(im)

draw.rectangle([(1, 1), (5, 5)], width=2, outline="green")
draw.rectangle([(6, 6), (10, 10)], width=2, outline="blue")
im.show()
```

image

image
	
:x: `width` is not ignored when `fill` is set.

d0sboots:
Thanks for unittesting the docs!

I don't want to just delete those lines until I understand what's going on though... the code seems very unambiguous that they *should* be ignored: https://github.com/python-pillow/Pillow/blob/aa8877ac5bb7c33939c4503ef56240e7027ac8ad/src/PIL/ImageDraw.py#L285

(Specifically, neither `width` nor `ink` is passed to the underlying C library function.)

d0sboots:
Oh, answered my own question immediately. It *is* ignored on that call, but that doesn't prevent the outline call from happening. Right-o, lemme fix it to only have the correct info.

d0sboots:
I like it. I feel like it has better (correct) implications for how outline works - that increasing width increases the border by working inwards.

hugovk:
Thanks!

PR #6633, opened 2022-10-02
Improved layout_engine documentation

### LINKED ISSUE #6631: ImageFont.getlength dramatically slower in some environments
### What did you do?
There's this [script](https://github.com/avalonv/literaryclock/blob/main/quote%20to%20image/quote_to_image.py) which reads some text and recursively wraps/scales a TrueType font to fit an image. I was porting it to Windows and noticed that it ran almost 3x faster than Linux on the same machine, which greatly hurt my sense of superiority as a penguin user. 

Never profiled python code before but what I did was `python -m cProfile -o quote_to_image.prof quote_to_image.py`, and then checked the output with [snakeviz](https://github.com/jiffyclub/snakeviz). Unless I'm wildly mistaken about how this stuff works, there's a massive discrepancy in time spent on getlength calls, despite the number of calls being very similar, with Linux spending more than 3 minutes to handle methods that Windows handled in 30 seconds.
```
ncalls  tottime percall         cumtime percall         filename:lineno(function)
1482429	187.6	0.0001265	187.6	0.0001265	~:0(<method 'getlength' of 'Font' objects>) Linux
1485834	26.45	1.78e-05	26.45	1.78e-05	~:0(<method 'getlength' of 'Font' objects>) Windows
```

To isolate the issue a bit more I decided to profile a shorter [script](https://gist.github.com/avalonv/34486658cc56e5b96dce509f7604b9ba) with only the function that does most of these calls, and interestingly, the discrepancy isn't anywhere as large. However, it's still very obviously there:
```
ncalls  tottime percall         cumtime percall         filename:lineno(function)
96726	16.64	0.000172	16.64	0.000172	~:0(<method 'getlength' of 'Font' objects>) Linux
96726	12.72	0.0001315	12.72	0.0001315	~:0(<method 'getlength' of 'Font' objects>) Windows
```

I was able to reproduce both tests on different dual-booting machines with different python/pillow versions, however I didn't bother to save the results cause dual-booting on a HDD sucks. If I'm an idiot and you need more information let me know.

Similar time spent ImageFont getlength method calls on both operating systems.

### What actually happened? 
Windows was faster, ranging from 500% to 25% (increases exponentially with the number of calls?)

* OS: Windows 10/Arch linux 5.19
* Python: 3.10.7
* Pillow: 9.2.0

radarhere:
Resolves #6631

### DIFF
--- a/src/PIL/ImageFont.py
+++ b/src/PIL/ImageFont.py
@@ -945,6 +945,8 @@ def truetype(font=None, size=10, index=0, encoding="", layout_engine=None):
                      encoding of any text provided in subsequent operations.
     :param layout_engine: Which layout engine to use, if available:
                      :data:`.ImageFont.Layout.BASIC` or :data:`.ImageFont.Layout.RAQM`.
+                     If it is available, Raqm layout will be used by default.
+                     Otherwise, basic layout will be used.
 
                      You can check support for Raqm layout using
                      :py:func:`PIL.features.check_feature` with ``feature="raqm"``.
--- a/src/PIL/ImageFont.py
+++ b/src/PIL/ImageFont.py
@@ -948,6 +948,9 @@ def truetype(font=None, size=10, index=0, encoding="", layout_engine=None):
                      If it is available, Raqm layout will be used by default.
                      Otherwise, basic layout will be used.
 
+                     If complex text layout is not required, basic layout will have
+                     better performance.
+
                      You can check support for Raqm layout using
                      :py:func:`PIL.features.check_feature` with ``feature="raqm"``.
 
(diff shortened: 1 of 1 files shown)


src/PIL/ImageFont.py
```diff
@@ -945,6 +945,11 @@ def truetype(font=None, size=10, index=0, encoding="", layout_engine=None):
                      encoding of any text provided in subsequent operations.
     :param layout_engine: Which layout engine to use, if available:
                      :data:`.ImageFont.Layout.BASIC` or :data:`.ImageFont.Layout.RAQM`.
+                     If it is available, Raqm layout will be used by default.
+                     Otherwise, basic layout will be used.
+
+                     If complex text layout is not required, basic layout will have
+                     better performance.
```
nulano, line 952:
```suggestion
                     Raqm layout is recommended for all non-English text. If Raqm layout
                     is not required, basic layout will have better performance.
```

nulano (approved):

PR #6634, opened 2022-10-03
Parametrized tests

### LINKED ISSUE #6525: Parameterize tests
Currently many of the pytest tests start with a loop over a list of some values, with all of the test code inside this loop.

https://github.com/python-pillow/Pillow/blob/4dd678a2e77a5111ca388187e599b986b7ad43bc/Tests/test_image_paste.py#L104-L112

Instead, we can use pytest's [`@pytest.mark.parametrize` annotation](https://docs.pytest.org/en/latest/how-to/parametrize.html) to handle this looping automatically.

https://github.com/python-pillow/Pillow/blob/b236c61c04c0f6a6cc1ac24f5a56e327e890ad9c/Tests/test_image_paste.py#L106-L114

The benefit of using pytest to do this is that if a test fails, it tells you which parameter it failed with. Also pytest will test with each of the parameters instead of stopping on the first failure.

I've already made this change for one file (#6519), but there are a lot more tests that could benefit from this annotation.

radarhere:
Resolves #6525

### DIFF
--- a/Tests/test_file_eps.py
+++ b/Tests/test_file_eps.py
@@ -125,14 +125,6 @@ def test_file_object(tmp_path):
 
 
 @pytest.mark.skipif(not HAS_GHOSTSCRIPT, reason="Ghostscript not available")
-def test_iobase_object(tmp_path):
-    # issue 479
-    with Image.open(FILE1) as image1:
-        with open(str(tmp_path / "temp_iobase.eps"), "wb") as fh:
-            image1.save(fh, "EPS")
-
-
-@pytest.mark.skipif(not HAS_GHOSTSCRIPT, reason="Ghostscript not available")
 def test_bytesio_object():
     with open(FILE1, "rb") as f:
         img_bytes = io.BytesIO(f.read())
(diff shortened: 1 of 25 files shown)


Tests/test_file_eps.py
```diff
@@ -124,14 +124,6 @@ def test_file_object(tmp_path):
             image1.save(fh, "EPS")
 
 
-@pytest.mark.skipif(not HAS_GHOSTSCRIPT, reason="Ghostscript not available")
-def test_iobase_object(tmp_path):
-    # issue 479
-    with Image.open(FILE1) as image1:
-        with open(str(tmp_path / "temp_iobase.eps"), "wb") as fh:
-            image1.save(fh, "EPS")
```
radarhere, line 132: #4109 caused this to become a duplicate of `test_file_object`

Tests/test_file_libtiff.py
```diff
@@ -509,20 +509,13 @@ def test_palette_save(self, im, tmp_path):
             # colormap/palette tag
             assert len(reloaded.tag_v2[320]) == 768
 
-    def xtest_bw_compression_w_rgb(self, tmp_path):
-        """This test passes, but when running all tests causes a failure due
-        to output on stderr from the error thrown by libtiff. We need to
-        capture that but not now"""
```
radarhere, line 515: Enabling this test now, it doesn't cause the tests to fail.

Issue #6614, opened 2022-09-23
Make `import PIL; PIL.Image` work


Kodiologist:
With Pillow 9.2.0 and (e.g.) matplotlib 3.5.2:

```console
$ python3 -c 'import PIL; PIL.Image'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
AttributeError: module 'PIL' has no attribute 'Image'
$ python3 -c 'import matplotlib, PIL; PIL.Image'
# No error.
```

[This behavior is quite strange.](https://stackoverflow.com/a/69125762)

hugovk:
Alternatively, maybe we should only allow the recommended `from PIL import Image`?

https://pillow.readthedocs.io/en/stable/installation.html#warnings

Kodiologist:
Or `import PIL.Image`, or just `import Pillow` to eliminate confusion with https://pypi.org/project/PIL once and for all.

aclark4life:
I don't find this issue that strange, but yes, I guess I would deprecate ``import PIL.Image`` because ... I don't think I've ever typed that in 10+ years of Pillowing. So, if it helps eliminate some confusion, 👍 

Also [we have been given control of PIL on PyPI](https://pypi.org/project/PIL/), but we've not done anything with it. Enabling ``import Pillow`` is probably not an improvement... unless we print a message "You should probably use ``from PIL import Image``."

So,

```
>>> import PIL.Image
Please import Image from PIL.

>>> import PIL
Please import Image from PIL.

>>> import Pillow
Please import Image from PIL.

```

Something like that.

radarhere:
[...]

PR #6635, opened 2022-10-03
Do not import PIL.Image

radarhere:
Since we're [considering](https://github.com/python-pillow/Pillow/issues/6614#issuecomment-1256599460) deprecating `import PIL.Image`, this PR removes a use of that from the test suite.


hugovk:
Good idea.

At the very least `from PIL import Image` is best practice and `import PIL` and `import PIL.Image` are discouraged.

PR #6640, opened 2022-10-04
Fix indices of colors read from GIMP Palette file

### LINKED ISSUE #6639: Indexes of colors when reading a GIMP Palette file are wrong
At PIL root, in an iPython prompt do:
```python
from PIL.GimpPaletteFile import GimpPaletteFile

!cat Tests/images/custom_gimp_palette.gpl
GimpPaletteFile(open("Tests/images/custom_gimp_palette.gpl", "rb")).getpalette()
```

The results are
```
n [109]: from PIL.GimpPaletteFile import GimpPaletteFile

In [110]: !cat Tests/images/custom_gimp_palette.gpl
GIMP Palette
Name: custompalette
Columns: 4
#
  0   0   0     Index 3
 65  38  30     Index 4
103  62  49     Index 6
 79  73  72     Index 7
114 101  97     Index 8
208 127 100     Index 9
151 144 142     Index 10
221 207 199     Index 11

In [111]: GimpPaletteFile(open("Tests/images/custom_gimp_palette.gpl", "rb")).getpalette()
Out[111]: 
(b'\x00\x00\x00\x01\x01\x01\x02\x02\x02\x00\x00\x00A&\x1eg>1OIHrea\xd0\x7fd\x97\x90\x8e\xdd\xcf\xc7\x0b\x0b\x0b\x0c\x0c\x0c<...>',
 'RGB')
```

Inspecting the bytes generated by reading the file, it is possible to see that the 3 first colors in the palette do not correspond to the 3 first colors in the file (the first does, but it is (0, 0, 0) by a coincidence). Rather, the colors specified in the palette file
show up from index 3 on.

Strangely, that is even sorted of "documented" in the color names in the existing sample file, but it is plain wrong, as loading that file in GIMP will assign index 0 to the color currently named "index 3" in custom_gimp_palette.gpl" and so on. Although the behavior is acknowledged, it is straightforward wrong, as one can't replicate the colors from an original GIMP palette file by inspecting the palette created by this loader alone - one has to be aware of this quirk instead.

Moreover, the current implementation will simply ignore the last 3 colors in a 256 GIMP GPL file.

As an additional feature, although the created palette has always 256 colors as needed in other PIL structures using Palettes, one has no way to know which of these colors are from the original file, and which are from the filer gray gradient that is created by default. Annotating the number of colors in the instance of the palette reader is essentially free, so it can be put there should anyone actually want to use this.

jsbueno:
Fixes #6639

 * Do not count non-color lines (headers and comments) in the Palette file as color indexes
 * Annotate the total number of read colors in the GimpPaletteFile instance
 * Update tests accordingly
 * Rename color names in the sample palette file to match the indexes actually present in the file.


src/PIL/GimpPaletteFile.py
```diff
@@ -26,30 +26,28 @@ class GimpPaletteFile:
...
+        index = 0
+        for s in fp:
             # skip fields and comment lines
             if re.match(rb"\w+:|#", s):
                 continue
             if len(s) > 100:
                 raise SyntaxError("bad palette file")
 
             v = tuple(map(int, s.split()[:3]))
-            if len(v) != 3:
```
radarhere, line 47: This change seems unnecessary? The previous line has already limited the number of items to 3.

jsbueno:
[WIP]  - modifying to allow reading > 256 colors

src/PIL/GimpPaletteFile.py
```diff
@@ -24,33 +25,49 @@ class GimpPaletteFile:
 
     rawmode = "RGB"
 
-    def __init__(self, fp):
+    #: override if reading larger palettes is needed
+    max_colors = 256
+    _max_line_size = 100
```
radarhere, line 30: We're reading RGB data from each line. That's three integers. Under what circumstances would a user need to read more than 100 characters?

src/PIL/GimpPaletteFile.py
```diff
@@ -24,33 +25,49 @@ class GimpPaletteFile:
...
+            for i in range(3):
+                self.palette += o8(int(v[i]))
 
     def getpalette(self):
 
         return self.palette, self.rawmode
+
+    @property
+    def n_colors(self):
+        return len(self.palette) / 3
```
radarhere, line 73: [As I've said](https://github.com/jsbueno/Pillow/pull/1#issue-1398563514), i don't think that `n_colors` is necessary. If the user is going to use the palette string, then they have enough of an understanding to be able to determine how many colors there are. This is a simple operation without an obvious need, so I'd like to avoid adding to the API for a trivial case.

But if you disagree, you can leave it here and see what others think.

src/PIL/GimpPaletteFile.py
jsbueno: The 4th Colunn contains the color name, though that is currently unused. And there might be a name with more than 100 characters - moving the number to an attribute just offera chance of one working around an error imposed by an arbitrary, though reasonable, limit on an otherwise good file.

src/PIL/GimpPaletteFile.py
jsbueno: I will remove it. However, I definetelly like Python things to be "pythonic".  That include bring able to know about object states through properties, usingagic.methods so users can comfortably use operators and common interfaces to deal with their data.

In my experience these things make libraries more joyful to use and easier to maintain.

Pillow Is a project that tends to the opposite: a lot of functional things and magic global states. I think part of that is due to historic reasons than a philosophy of the project, and hope to offer more pythonic interfaces like I had indicated in the chat about animared images.

That said, I'd say these comfortable properties could be left for a revamp of the ImagPallete class ata future time, and there is no need for the new property here.

This will also give time for all involved to understand and have an opinion if this "first class classes" approach is desirable at all, and draw, projectwise, a more solid roadmap.


(As far as I am concerned wanting to know the number of colors of a palette is an operation common enough and with nothing of unexpected - akim to getting a strong length - and requiring one to deal with bytes when their problem is dealing with colors is a quasi definition of "unpythonic". However, if this class is to be instrumented,.maybe this could be offered as it's `__len__` for once, instead of creating an arbitrarily named property.)

src/PIL/GimpPaletteFile.py
```diff
@@ -24,33 +25,49 @@ class GimpPaletteFile:
...
             if re.match(rb"\w+:|#", s):
                 continue
-            if len(s) > 100:
+            if len(s) > self._max_line_size:
                 raise SyntaxError("bad palette file")
 
-            v = tuple(map(int, s.split()[:3]))
-            if len(v) != 3:
+            # 4th column is color name and may contain spaces.
+            v = s.split(None, 4)
```
radarhere, line 61:
```suggestion
            v = s.split(maxsplit=3)
```
3 is sufficient, as seen here -
```pycon
>>> "0 1 2 name".split(maxsplit=3)
['0', '1', '2', 'name']
```

jsbueno: My bad. it is indeed the maximum "splits", not maximum "resulting components".

Tests/test_file_gimppalette.py
```diff
@@ -29,4 +45,26 @@ def test_get_palette():
...
     assert mode == "RGB"
+
+
+def test_n_colors():
+    # Arrange
+    with open("Tests/images/custom_gimp_palette.gpl", "rb") as fp:
+        palette_file = GimpPaletteFile(fp)
+
+    palette, _ = palette_file.getpalette()
+    assert len(palette) == 24
```
radarhere, line 70:
```suggestion
```

Now that `n_colors` has been removed, I don't think this is doing anything above and beyond the previous test. Or at least, I think it could be removed from `test_n_colors` to something else.

src/PIL/GimpPaletteFile.py
```diff
@@ -24,32 +25,44 @@ class GimpPaletteFile:
...
             raise SyntaxError("not a GIMP palette file")
 
-        for i in range(256):
+        read = 0
+
+        self.palette = b""
+        while len(self.palette) < 3 * self.max_colors:
 
-            s = fp.readline()
+            s = fp.readline(self._max_file_size)
```
radarhere, line 43:
```suggestion
            s = fp.readline(self._max_line_size)
```

Tests/images/custom_gimp_palette.gpl
```diff
@@ -1,12 +1,12 @@
...
-#
-  0   0   0     Index 3
- 65  38  30     Index 4
-103  62  49     Index 6
- 79  73  72     Index 7
-114 101  97     Index 8
-208 127 100     Index 9
-151 144 142     Index 10
-221 207 199     Index 11
+# Original written by David Wetz in https://stackoverflow.com/questions/815836/im-creating-a-program-that-generates-a-palette-from-a-true-color-image-need-hel/815855#815855
```
radarhere, line 4: Just an explanation of what is happening here - this isn't a random attribution. You're creating a comment that is over 100 characters.

src/PIL/GimpPaletteFile.py
```diff
@@ -49,7 +49,7 @@ def __init__(self, fp):
                 warnings.warn(
                     f"Palette file truncated at {self._max_file_size - len(s)} bytes"
                 )
-                break
+                break  # pragma: no cover
```
radarhere, line 52: Why is this "no cover"? I know your commit message says "testing promotes the warning to an exception", but I don't know what you mean by that. All our CIs passed.

jsbueno: Hi - I also had to learn further on Warnings to write this code: 
The thing is, on ordinary runs, creating a warning just does its thing - with no further configurations it logs to stderr. 

However, the Warnings has a statefull config (which takes place in a `with` block in the test), which changes then to raise as exceptions instead of simply logging. I found it simpler to check for the warning by letting it raise and using `pytest.raies`. 

The problem is the coverage extension: on running tests, as the warning raises, line 52 is not executed: the pipelines warned me that this line was not covered in tests. the `# pragma: no cover`  is the explict way to tell the coverage plug-in it should not worry about this line.

Of course, I don't know at this point if it is preferrable to have a "not 100% covered" report, as that may be a misleading metric, to have the visual dirt these directives insert - feel free to remove it.

radarhere: The solution to this is to use [`pytest.warns`](https://docs.pytest.org/en/7.1.x/how-to/capture-warnings.html#warns), as Pillow does [elsewhere](https://github.com/python-pillow/Pillow/blob/e0c30d2fcd90ed9cfb1bdf2d27f6170bc9e762ad/Tests/test_file_jpeg.py#L860-L861). I've created https://github.com/jsbueno/Pillow/pull/3, which will cover the line.

src/PIL/GimpPaletteFile.py
radarhere: This looks like a mistake, where you've accidentally used the file size limit for the line size limit for [the size argument.](https://docs.python.org/3/library/io.html#io.IOBase.readline)

jsbueno: no - "file" size is deliberate here - to avoid DoS only. 
line size checks are made just bellow, and as far as I know it is something just PIL has - there is no spec for this kind of file, but for GIMP source code, and there is no line size limit there.

Changing here for reading linesize would make te file break with long comments (as it did), and that makes no sense:

if one has a python project in which Palttes with this format are used, there is no reason to impose an arbitrary limit on comments on that source code.   Resilience against malicious files when those are input by 3rdy parties are mitigated through the file-size check.

Tests/images/custom_gimp_palette.gpl
jsbueno: comments with >100 char lines long should not "break" the file. Even if we are reformatting here, the URL alone is wider than 100 chars in this case, so even breaking this in a functional way would  still have lines wider than that.

src/PIL/GimpPaletteFile.py
jsbueno: (currently unused in PIL, that is: GIMP uses that name)

radarhere:
Regarding the line size limit, I expect the original intention of the 100 character limit was to pick an unreasonably high number, so that if it was being exceeded, then something was wrong, or an attack was occurring.

You have a different opinion, that we should support, at least through configuration, all theoretically possible files. Because of the large number of users who pass through this repository with requests, I have a concern about Pillow becoming overly complex. If no one would actually benefit, I'm reluctant to complicate the API further. Overall, I think that three attributes to control reading limits for one plugin is complicated.

Regarding the color limit, [as I've said](https://github.com/python-pillow/Pillow/issues/6639#issuecomment-1269311418), I'm reluctant to do so, as I don't want to mislead users into thinking that they have a palette that can be used without a problem elsewhere in Pillow.

radarhere:
#8805 has been merged to cover part of this. I've created #8826 as an alternative to the rest.

jsbueno:
Hi - thanks to coming back to this.

I can't organize myself, at least over the next few weeks, to contribute with the project. But it is something I intend to do at some point. 

Also, since you split the issue, feel free to close this one, if you think it is due.

radarhere:
#8826 has been merged.

PR #3335, opened 2018-09-05
Add more raw Tiff modes (RGBaX, RGBaXX, RGBAX, RGBAXX)

homm:

PR #6642, opened 2022-10-05
added support for 16 bit integer tiffs

JayWiz:
just added a line to support 16 bit integer TIFFs. is kinda unusual, but this this is needed for custom generated TIFFs, e.g. if you're working with holograms or interferograms.

## What did i do?
added a line in TiffImagePlugin.py to support 16 bit integer TIFFs

## What did i expect to happen?
that Pillow is able to work with 16 bit integer TIFFs

## What did actually happen?
Pillow is able to work with 16 bit integer TIFFs

## What versions of Pillow and Python am i using?
Python 3.9 and Pillow 9.2.0


src/PIL/TiffImagePlugin.py
```diff
@@ -169,6 +169,7 @@
     (II, 0, (1,), 2, (8,), ()): ("L", "L;IR"),
     (MM, 0, (1,), 2, (8,), ()): ("L", "L;IR"),
     (II, 1, (1,), 1, (8,), ()): ("L", "L"),
+    (II, 0, (1,), 1, (16,), ()): ("I", "I;16"),
```
radarhere, line 172:
```suggestion
    (II, 0, (1,), 1, (16,), ()): ("I;16", "I;16"),
```
You will notice that a few lines later, there is a rawmode of "I;16".
https://github.com/python-pillow/Pillow/blob/243402e78e2bf2b7af478c6d891816e372b5c3f9/src/PIL/TiffImagePlugin.py#L176
and the mode used is "I;16". It would seem more consistent to also use it here.

radarhere:
#3335 didn't add any TIFF-specific tests, so I don't think tests are needed here.

Feel free to say no, but would you be able to upload a copy of one such TIFF here, just in case we want to refer back to it in the future?

src/PIL/TiffImagePlugin.py
radarhere: Moving the line to another position would also be helpful. I've created https://github.com/JayWiz/Pillow/pull/1 for this.

JayWiz:
> #3335 didn't add any TIFF-specific tests, so I don't think tests are needed here.
> 
> Feel free to say no, but would you be able to upload a copy of one such TIFF here, just in case we want to refer back to it in the future?

yes of course, i'm fine with that. where should i upload it?

radarhere:
Just attach it to a comment here. I understand that GitHub won't let you attach a TIFF, so compress it into a zip first.

JayWiz:
[image_raw.zip](https://github.com/python-pillow/Pillow/files/9716594/image_raw.zip)

in case you're interested: this is a multipage tiff with 6 frames generated by two lasers, which can be reconstructed to this 3d point cloud. had to find a smaller tiff, usually one tiff is around 800 mb)
image

radarhere:
Thanks

radarhere (approved):

PR #6645, opened 2022-10-07
Better C integer definitions

Yay295:
1. Fix `INT64` definition where it was checking the size of `long long` but then defining it as `long`.
2. Add compile warning (not an error) if `INT64` wasn't defined.
3. Add labels to some preprocessor if/else/endifs to make it more clear which ones go together.
4. Add comment explaining why we're using `#define` instead of `typedef`.
5. Check for C99 support and use the values in `stdint.h` if possible.
6. Move the `INT8` definition to the top, so that they're all in order by size.
7. Change the `INT16` definition failure to an error instead of silently using the wrong value. There was a comment saying things mostly worked anyways, but I'm skeptical of that, and also we were already failing if `INT32` couldn't be defined, so I doubt it would have gotten past that point anyways.


src/libImaging/ImPlatform.h
```diff
@@ -37,15 +37,37 @@
 #undef WIN32
 #endif
 
-#else
+#else /* WIN */
```
hugovk, line 40: The if block is for WIN, the else block is for not WIN?

```suggestion
#else /* not WIN */
```

src/libImaging/ImPlatform.h
```diff
@@ -37,15 +37,37 @@
...
+#define INT16 int16_t
+#define UINT16 uint16_t
+#define INT32 int32_t
+#define UINT32 uint32_t
+#ifdef INT64_MAX
+#define INT64 int64_t
+#define UINT64 uint64_t
+#endif
+
+#else /* C99+ */
```
hugovk, line 61: Similarly:
```suggestion
#else /* < C99 */
```

src/libImaging/ImPlatform.h
```diff
@@ -61,17 +83,22 @@
...
 #define UINT16 unsigned INT16
 #define UINT32 unsigned INT32
-
+#ifdef INT64
+#define UINT64 unsigned INT64
 #endif
 
+#endif /* C99+ */
+
+#endif /* WIN */
```
hugovk, line 100: And at least elsewhere the not case has been used for endif:

```suggestion
#endif /* < C99+ */

#endif /* not WIN */
```

hugovk:
What makes these definitions better?

Do we need to support both C99 and pre-C99?

What is the minimum needed for CPython itself, or other libraries such as NumPy?

radarhere:
The oldest version of GCC I could find that we are using is 4.8.5 - https://github.com/python-pillow/docker-images/actions/runs/4813262439/jobs/8569570838#step:6:99

https://gcc.gnu.org/onlinedocs/gcc-4.8.5/cpp.pdf
> The value 199409L signifies the 1989 C standard as amended in 1994, which is the current default; the value 199901L signifies the 1999 revision of the C standard. Support for the 1999 revision is not yet complete.

hugovk:
> What is the minimum needed for CPython itself [...]?

[PEP 7](https://peps.python.org/pep-0007/#c-dialect) says:

> * Python 3.11 and newer versions use C11 without [optional features](https://en.wikipedia.org/wiki/C11_%28C_standard_revision%29#Optional_features). The public C API should be compatible with C++.
> * Python 3.6 to 3.10 use C89 with several select C99 features:
>   * Standard integer types in <stdint.h> and <inttypes.h>. We require the fixed width integer types.
>   * static inline functions
>   * designated initializers (especially nice for type declarations)
>   * intermingled declarations
>   * booleans
>   * C++-style line comments

hugovk:
> What is the minimum needed for [...] other libraries such as NumPy?

https://numpy.org/doc/stable/user/building.html says:

> Much of NumPy is written in C. You will need a C compiler that complies with the C99 standard.

Yay295:
`<stdint.h>` is better because it's standardized, so we have a guarantee about what exists.

We previously discussed the minimum C version here: https://github.com/python-pillow/Pillow/discussions/6516
The oldest we found was GCC on CentOS 7 which uses `gnu90` by default.

If we could guarantee that `<stdint.h>` exists we could simplify this file a lot, but it doesn't seem like we can do that yet.

src/libImaging/ImPlatform.h
```diff
@@ -61,17 +83,22 @@
...
 #define UINT8 unsigned char
-
 #define UINT16 unsigned INT16
 #define UINT32 unsigned INT32
-
+#ifdef INT64
+#define UINT64 unsigned INT64
 #endif
 
+#endif /* < C99+ */
```
radarhere, line 98: This isn't `C99+` or `< C99`, it's both? This is a typo, yes?

hugovk:
Thanks. And CentOS 7 is [EOL on 30 Jun 2024](https://endoflife.date/centos).

We could consider requiring a newer C for Pillow 10.4, due out the next day on 1 Jul 2024, but I'd be more comfortable rolling it into the next quarterly release seeing as it's a major bump (11.0, 15 Oct 2024, supporting Python 3.9+).

And we may want to keep support for older C versions longer, but let's re-evaluate in a year or so.

hugovk (approved):

hugovk:
@nulano How does this look from the Windows perspective?

nulano:
I don't see any Windows changes in this PR.
The `long long` fix looks like it would be Windows-specific, but it is in a non-Windows block, which is presumably why it hasn't been noticed until now.

> If we could guarantee that <stdint.h> exists we could simplify this file a lot, but it doesn't seem like we can do that yet.

I think this also applies to Windows. I'm not sure if it is guaranteed to exist on Windows, but if it might not be present on other platforms, I see no reason to worry about it yet.

src/libImaging/ImPlatform.h
```diff
@@ -61,17 +83,22 @@
 #if SIZEOF_LONG == 8
 #define INT64 long
 #elif SIZEOF_LONG_LONG == 8
-#define INT64 long
+#define INT64 long long
+#else
+#warning Cannot find required 64-bit integer type
```
radarhere, line 88: > Add compile warning (not an error) if INT64 wasn't defined.

So, why is this a warning, and not an error like the 16-bit and 32-bit counterparts?

Yay295: Because we don't actually use INT64 anywhere, so it won't cause any problems if someone happens to be using a system that doesn't support it.

Yay295: It was already defined; I didn't add it.

radarhere: Ok, I've created #7180 to remove it.

src/libImaging/ImPlatform.h
```diff
@@ -61,17 +83,22 @@
...
 #endif
 
-#define INT8 signed char
 #define UINT8 unsigned char
-
 #define UINT16 unsigned INT16
 #define UINT32 unsigned INT32
-
+#ifdef INT64
+#define UINT64 unsigned INT64
```
radarhere, line 95: But you have added UINT64, which also doesn't seem to be used anywhere?

Yay295: If there's one I figure we might as well have the other.

src/libImaging/ImPlatform.h
radarhere: #7180 has been merged. I've created https://github.com/Yay295/Pillow/pull/7 to resolve the conflict.

src/libImaging/ImPlatform.h
radarhere: After #7180 was merged, INT64 is no longer present, so I've created https://github.com/Yay295/Pillow/pull/7 to remove UINT64.

homm (approved):

Issue #3738, opened 2019-03-22
Support list of list in `draw.line`


gaoxinge:
Code below is running normally:

```python
from PIL import Image, ImageDraw

WHITE = (255, 255, 255)
BLUE = "#0000ff"
MyImage = Image.new('RGB', (600, 400), WHITE)
MyDraw = ImageDraw.Draw(MyImage)
line_points = [(100, 100), (150, 200), (300, 100), (500, 300)]  # list of tuple
MyDraw.line(line_points, width=40, fill=BLUE, joint='curve')
MyImage.show()
```

But if I change 

```python
line_points = [(100, 100), (150, 200), (300, 100), (500, 300)]  # list of tuple
```

into

```python
line_points = [[100, 100], [150, 200], [300, 100], [500, 300]]  # list of list
```

the code will raise exception because the points are represented by list. 

Is it appropriate that `draw.line` support the list of list as arguments?

My pillow version is `5.4.1`.

radarhere:
I can see that it could be useful. I've created PR #3739 to resolve this.

gaoxinge:
@radarhere thanks!

hugovk:
My initial thought is how about all the other places we have parameters that take a tuple (eg. `Image.new("RGB", size=(20, 20), color=(255, 255, 255))`)? Should they also take lists?

---

A quick search on when to use a tuple or a list gives https://stackoverflow.com/a/1708538/724176:

> [Tuples][2] are fixed size in nature whereas [lists][3] are dynamic.  
> In other words, a `tuple` is **immutable** whereas a `list` is **mutable**.
> 
>  1. You can't add elements to a tuple. Tuples have no append or extend method.
>  2. You can't remove elements from a tuple. Tuples have no remove or pop method.
>  3. You can find elements in a tuple, since this doesn’t change the tuple. 
>  4. You can also use the `in` operator to check if an element exists in the tuple.
> 
> ----------
> 
> - **Tuples are faster than lists.** If you're defining a constant set of values and all you're ever going to do with it is iterate through it, use a tuple instead of a list.
> 
> - It makes your code safer if you “write-protect” data that does not need to be changed. Using a tuple instead of a list is like having an implied assert statement that this data is constant, and that special thought (and a specific function) is required to override that.
> 
> - Some tuples can be used as dictionary keys (specifically, tuples that contain immutable values like strings, numbers, and other tuples). Lists can never be used as dictionary keys, because lists are not immutable. 
> 
[...]

PR #6653, opened 2022-10-09
Clean up tests

Yay295:


Tests/test_image_transform.py
```diff
@@ -178,11 +178,13 @@ def _test_nearest(self, op, mode):
...
-            (20 * 10, opaque),
-            (20 * 10, transparent),
-        ]
+        colors = sorted(im.getcolors())
+        assert colors == sorted(
+            (
+                (20 * 10, opaque),
+                (20 * 10, transparent),
+            )
+        )
```
radarhere, line 187: What is your thinking here? I know we've had this conversation in https://github.com/python-pillow/Pillow/pull/6540#discussion_r979468526 already. Are you trying to pave the way for that PR?

Tests/test_imagepath.py
```diff
@@ -152,7 +152,7 @@ def test_transform():
 
 def test_transform_with_wrap():
     # Arrange
-    p = ImagePath.Path([0, 1, 2, 3])
+    p = ImagePath.Path((0, 1, 2, 3))
```
radarhere, line 155: I'm less keen on this. Here, you're not just changing the code to setup the tests, you're actually changing what is being tested.

I know you might assume that tuples and lists should be interchangeable as arguments to Pillow code, but we've had issues like #3738

Yay295: Those are still being tested here: https://github.com/python-pillow/Pillow/blob/bf00a4534d728dff93232142cf99aabe4e8174fd/Tests/test_imagepath.py#L42-L58
I didn't change every list to a tuple.

Tests/test_image_transform.py
Yay295: I'm making this test match this other test:
https://github.com/python-pillow/Pillow/blob/bf00a4534d728dff93232142cf99aabe4e8174fd/Tests/test_image.py#L313-L342

Tests/test_imagepath.py
Yay295: I've added more tests with c090fe1.

hugovk:
Thanks for this PR!

The first five commits look good.

"[use tuple or set instead of list in tests](https://github.com/python-pillow/Pillow/pull/6653/commits/bf00a4534d728dff93232142cf99aabe4e8174fd)" is too big to review properly, and I know tuples are generally preferred to lists, but in this case I don't think the huge review effort is worth any noticeable benefit. (I think these kind of changes are okay when making other changes to a file.)

And @radarhere had a question about "[sort colors before comparing them](https://github.com/python-pillow/Pillow/pull/6653/commits/8cacb09696901a5540aef0b8d871c44dfae93852)", so I'll defer on that.

Also I wouldn't worry about the last few commits splitting long lines and wrangling with the formatting, it's okay to have longish strings (below the limit) as they're easier to grep.

Tests/helper.py
```diff
@@ -95,7 +95,7 @@ def assert_image_equal(a, b, msg=None):
             except Exception:
                 pass
 
-        assert False, msg or "got different content"
+        assert a.tobytes() == b.tobytes(), msg or "got different content"
```
radarhere, line 98: Why make this change? `a.tobytes()` and `b.tobytes()` have already been compared a few lines earlier, so we know this is false.

Yay295: It gives a better error message.
```python
import pytest

def test_assert_false():
    assert False, "message"

def test_assert_compare():
    assert b"abc" == b"abb", "message"
```
```python
================ FAILURES =================
____________ test_assert_false ____________

    def test_assert_false():
>       assert False, "message"
E       AssertionError: message
E       assert False

Tests/test_assert.py:4: AssertionError
___________ test_assert_compare ___________

    def test_assert_compare():
>       assert b"abc" == b"abb", "message"
E       AssertionError: message
E       assert b'abc' == b'abb'
E         At index 2 diff: b'c' != b'b'
E         Use -v to get more diff

Tests/test_assert.py:7: AssertionError
```

radarhere: How about https://github.com/Yay295/Pillow/pull/1 instead?

Yay295: Nice idea.

radarhere: In https://github.com/Yay295/Pillow/actions/runs/3810393071/jobs/6482448151#step:8:1241, you can see what happens when this fails now.

...the output is very large.

Tests/test_imagepath.py
radarhere: I've created https://github.com/Yay295/Pillow/pull/3

Tests/helper.py
Yay295: Apparently this is due to the "-v" option. ".ci/test.sh" uses this when running tests, but the makefile doesn't (except for "valgrind" which uses "-vv"), so I hadn't encountered this locally.

Yay295: It looks like there's an open issue on pytest's GitHub for this: https://github.com/pytest-dev/pytest/issues/5285

Tests/test_image_transform.py
radarhere: Ok, I see that `getcolors()` is only guaranteed to return an "[unsorted list](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.getcolors)", and the test in test_image.py actually returns a different order on big-endian.

Tests/helper.py
Yay295: I found a way to temporarily set the verbosity to 0. d75b31b6eae1bc2bfb77bb425e5d043a02a952ae

Tests/helper.py
```diff
@@ -15,6 +15,9 @@
 
 from PIL import Image, ImageMath, features
 
+# overridden in conftest.py
+pytestconfig = {"option": {"verbose": 0}}
```
radarhere, line 19: Since it's overridden, these lines don't need to be here?

Yay295: Probably. I wasn't sure if that test code might ever be run without a pytest config having been loaded first though.

Tests/test_image_resize.py
```diff
@@ -57,13 +57,13 @@ def test_convolution_modes(self):
             Image.Resampling.LANCZOS,
         ),
     )
-    def test_reduce_filters(self, resample):
-        r = self.resize(hopper("RGB"), (15, 12), resample)
```
radarhere, line 61: Why is 'resampling_filter' clearer? 'resample' is the name of the variable when the core resize operation is called in Image.py
https://github.com/python-pillow/Pillow/blob/86634b835257a7d3438eaf783d70eca12e20b13e/src/PIL/Image.py#L2192

Yay295:
#6900, #6904, and #6949 were split off from this, so I'd like those to be merged before I rebase this.

radarhere:
Those PRs have all now been merged.

Yay295:
I've rebased it. Should I create a new pull request for this, or keep using this one?

hugovk:
This PR is fine.

I don't think we really need 3f045ea194c9ef7f0f87e7f2d66310de443a623e "use tuple or set instead of list in tests", there's a lot of changes (+406 −399 in 54 files) for little benefit.

Yay295:
I've moved the `helper.py` changes to #7140.

PR #6656, opened 2022-10-09
Missing _imagingcms.pyd in Python 3.9 64bits

ignaciordc:
Python 3.9 installed by Conda in Windows 10.
Trying to access color conversion with lcms I have found that the DLL file (.pyd) is not installed


radarhere:
Hi. When you have a problem or question like this, it should actually be created as an issue.

This is a pull request, where you've suggested we make a change. The change that you're suggesting is that we combine all of the changes we've made since Pillow 9.2.0 into Pillow 9.2. I don't expect this would help your situation, but it's also not the plan. Our plan is instead to release Pillow 9.3.0 towards the end of the month.

Please create an issue, and provide the complete error that you're seeing. If you used conda-forge, you might also want to consider asking at https://github.com/conda-forge/pillow-feedstock instead.

hugovk:
Closing this, as it's triggering our CI for each merge to `main`. A detailed issue is welcome.

PR #6659, opened 2022-10-13
pylint warnings

marksmayo:
Fixed some pylint issues

Fixes # unnecessary elses, conditional checks, import structure and encodings

 * fixes to pylint warnings

### DIFF
--- a/Tests/test_file_ppm.py
+++ b/Tests/test_file_ppm.py
@@ -240,7 +240,7 @@ def test_header_token_too_long(tmp_path):
 def test_truncated_file(tmp_path):
     # Test EOF in header
     path = str(tmp_path / "temp.pgm")
-    with open(path, "w") as f:
+    with open(path, "w", encoding='utf-8') as f:
         f.write("P6")
 
     with pytest.raises(ValueError) as e:
@@ -259,7 +259,7 @@ def test_truncated_file(tmp_path):
 @pytest.mark.parametrize("maxval", (0, 65536))
 def test_invalid_maxval(maxval, tmp_path):
     path = str(tmp_path / "temp.ppm")
-    with open(path, "w") as f:
+    with open(path, "w", encoding='utf-8') as f:
         f.write("P6\n3 1 " + str(maxval))
 
     with pytest.raises(ValueError) as e:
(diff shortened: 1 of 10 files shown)


marksmayo:
Not quite sure what to do about the coverage - it's not like I changed functionality much, per se

radarhere:
I've created PR #6660 to help one of the missing coverage lines. As for the other, it's [being skipped on CI jobs](https://github.com/python-pillow/Pillow/blob/f63cc582b7d305780762f786cf61e602f7398b0d/Tests/test_image_access.py#L410-L414), so there's nothing to do there.

Tests/test_file_ppm.py
```diff
@@ -240,7 +240,7 @@ def test_header_token_too_long(tmp_path):
 def test_truncated_file(tmp_path):
     # Test EOF in header
     path = str(tmp_path / "temp.pgm")
-    with open(path, "w") as f:
+    with open(path, "w", encoding="utf-8") as f:
         f.write("P6")
 
     with pytest.raises(ValueError) as e:
```
Yay295, line 246: ppm/pgm files are either [ASCII or binary](https://en.wikipedia.org/wiki/Netpbm#Description). Since we're writing text here we should probably use `"ascii"`.

radarhere:
Just checking - so are we adding another linting standard? I had thought the idea behind using black was to end style debates.

https://github.com/python-pillow/Pillow/pull/3733#issue-423817149
> I'd like to suggest formatting the code with Black:
>
>> By using Black, you agree to cede control over minutiae of hand-formatting. In return, Black gives you speed, determinism, and freedom from pycodestyle nagging about formatting. You will save time and mental energy for more important matters.

They will not always agree.

https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#pylint
> [Pylint](https://pypi.org/p/pylint/) is also a code linter like Flake8. It has the same checks as flake8 and more. In particular, it has more formatting checks regarding style conventions like variable naming. With so many checks, Pylint is bound to have some mixed feelings about Black’s formatting style.

hugovk (approved):
Black is about style, generally about whitespace and doesn't touch the AST, whereas Pylint is a static code analyser, and also checks for code deprecations. It can help find bugs.

Pylint has lots of suggestions, and should be evaluated on a case-by-case basis.

For example, we're not going to change:

> C0103: Variable name "im" doesn't conform to snake_case naming style (invalid-name)

I think these changes are fine.

Tests/test_file_ppm.py
radarhere: Since this is P6, it is binary.

I've created PR #6677 to change this to write in binary format.

PR #6674, opened 2022-10-21
Added support for reading BMP images with RLE4 compression

npjg:
It's an old format, but I still see it a good bit when examining old software.


Tests/test_file_bmp.py
```diff
@@ -176,6 +176,14 @@ def test_rle8():
                 im.load()
 
 
+def test_rle4():
+    with Image.open("Tests/images/hopper_rle4.bmp") as im:
+        assert_image_similar_tofile(im, "Tests/images/hopper_4bit.bmp", 12)
+
+    with Image.open("Tests/images/bmp/g/pal4rle.bmp") as im:
+        assert_image_similar_tofile(im, "Tests/images/bmp/g/pal4.bmp", 12)
```
radarhere, line 184: Is there any significant difference between hopper_rle4.bmp and pal4rle.bmp?

radarhere:
I've created https://github.com/npjg/Pillow/pull/1 with a documentation correction, and to simplify the code by combining the two BMP RLE decoders into one.

src/PIL/BmpImagePlugin.py
```diff
@@ -328,6 +330,69 @@ def decode(self, buffer):
...
+                    # delta
+                    bytes_read = self.fd.read(2)
+                    if len(bytes_read) < 2:
+                        break
+                    right, up = self.fd.read(2)
+                    data += b"\x00" * (right + up * self.state.xsize)
+                    x = len(data) % self.state.xsize
+                else:
+                    # absolute mode (2 pixels per byte)
+                    total_bytes_to_read = byte[0] // 2
```
Yay295, line 377: What does this "2" comment mean? If it means this value is always 2, that won't work, because this is already checked for on line 367.

radarhere: It's not a comment. Commenting in Python is #. This is floor division.

Yay295: ah, right. There's even a real comment right above that and I didn't notice...

Tests/test_file_bmp.py
radarhere: I've removed hopper_rle4.bmp, since the other test image seems sufficient.

radarhere (approved):

npjg:
I was gone for a few days. Looks good, thanks for merging!

PR #6678, opened 2022-10-23
Decode JPEG compressed BLP1 data in original mode

radarhere:


radarhere:
This is mentioned in the Pillow 9.3.0 release notes at https://pillow.readthedocs.io/en/stable/releasenotes/9.3.0.html#decode-jpeg-compressed-blp1-data-in-original-mode, and addresses https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=50217

PR #6684, opened 2022-10-24
Revert replacing and deprecating constants with enums

hugovk:
Fixes https://github.com/python-pillow/Pillow/issues/6537

 * Revert the changes in https://github.com/python-pillow/Pillow/pull/5954 (issue: https://github.com/python-pillow/Pillow/issues/5875) as they will likely prove too disruptive for the limited benefits they bring
 * While enums are good for this sort of thing, and we should use them for new code, on balance let's keep this legacy code as simple integer constants


hugovk:
Let's close this, I somehow had glossed over that the enums were already released in 9.1.0, so this itself would be a breaking change.

Some other ideas we're considering:

* Keep the enums and deprecated ints
* Keep the enums and deprecated ints, but extend deprecation period (e.g. from v10/July 2023 to v11/Oct 2024, or even without advertising an end date)
* Provide API via both ints and enums, remove int deprecations

PR #6691, opened 2022-10-26
Only use ASCII characters in C source file

cgohlke:


hugovk (approved):
I'm fine with this, but curious why non-ASCII is a problem?

cgohlke:
> I'm fine with this, but curious why non-ASCII is a problem?

PVS-Studio raised a warning about ["Code contains invisible characters that may alter its logic"](https://pvs-studio.com/en/docs/warnings/v1076/). I'm not sure the warning was legitimate but this PR fixes the warning at least.

PR #6692, opened 2022-10-27
MODIFY Image.py save() -> ADD convert JPG->JPEG code

ccgeonhee:
Fixes # .

Changes proposed in this pull request: If the extension is JPG, there is an error that cannot be saved. Therefore, I added a code that converts JPG to JPEG. You can add to FORMAT_MAP for similar cases. For example, MPG to MPEG.

 * Convert JPG to JPEG

Thank you.

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -97,7 +97,7 @@ try:
     # import Image and use the Image.core variable instead.
     # Also note that Image.core is not a publicly documented interface,
     # and should be considered private and subject to change.
-    from . import _imaging as core
+    from .. import _imaging as core
 
     if __version__ != getattr(core, "PILLOW_VERSION", None):
         raise ImportError(
@@ -225,6 +225,10 @@ OPEN = {}
 MIME = {}
 SAVE = {}
 SAVE_ALL = {}
+# Convert jpg to jpeg, Can apply other case.
+FORMAT_MAP = {
+    'JPG': 'jpeg',
+}
 EXTENSION = {}
 DECODERS = {}
 ENCODERS = {}
@@ -2332,6 +2336,7 @@ class Image:
             except KeyError as e:
                 raise ValueError(f"unknown file extension: {ext}") from e
 
+        format = FORMAT_MAP.get(format.upper(), format)
         if format.upper() not in SAVE:
             init()
         if save_all:
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -97,7 +97,7 @@ try:
     # import Image and use the Image.core variable instead.
     # Also note that Image.core is not a publicly documented interface,
     # and should be considered private and subject to change.
-    from .. import _imaging as core
+    from . import _imaging as core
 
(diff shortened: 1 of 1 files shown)


src/PIL/Image.py
```diff
@@ -97,7 +97,7 @@ class DecompressionBombError(Exception):
     # import Image and use the Image.core variable instead.
     # Also note that Image.core is not a publicly documented interface,
     # and should be considered private and subject to change.
-    from . import _imaging as core
+    from .. import _imaging as core
```
radarhere, line 100: This change isn't related to your initial description. Was it unintentional?

ccgeonhee: Oh,Sorry. I didn't delete it after the test.

radarhere: Thanks

radarhere:
Just to be clear - this should only be helpful to users who are saving an image to a file object instead of a filename, or who are saving to a file but don't want the extension derived from the filename.
```python
from PIL import Image
import io
im = Image.new("RGB", (1, 1))
fp = io.BytesIO()
im.save(fp, "JPG")
im.save("out", "JPG")
im.save("out.gif", "JPG")
```
I think most of the time, users just use a filename. The following code works fine.
```python
from PIL import Image
im = Image.new("RGB", (1, 1))
im.save("out.jpg")
```
If you are creating this PR because you would like the following to work -
```python
from PIL import Image
im = Image.new("RGB", (1, 1))
im.save("out.jpg", "JPG")
```
then I want to clarify that `format` could just be removed. It may seem obvious that it is optional, but I suspect not everyone realises this.

ccgeonhee:
> Just to be clear - this should only be helpful to users who are saving an image to a file object instead of a filename, or who are saving to a file but don't want the extension derived from the filename.
> 
> ```python
> from PIL import Image
> import io
> im = Image.new("RGB", (1, 1))
> fp = io.BytesIO()
> im.save(fp, "JPG")
> im.save("out", "JPG")
> im.save("out.gif", "JPG")
> ```
> 
> I think most of the time, users just use a filename. The following code works fine.
> 
> ```python
> from PIL import Image
> im = Image.new("RGB", (1, 1))
> im.save("out.jpg")
> ```
> 
> If you are creating this PR because you would like the following to work -
> 
> ```python
> from PIL import Image
> im = Image.new("RGB", (1, 1))
> im.save("out.jpg", "JPG")
> ```
> 
> then I want to clarify that `format` could just be removed. It may seem obvious that it is optional, but I suspect not everyone realises this.

Oh, I see. I think I forgot it was an option. 
The error that occurs when the `format`exists is annoying, so I created pr. 
If you're going to get rid of the `format`, you don't need this pr.

radarhere:
Oh no, I didn't mean that the `format` option will be removed from Pillow. I meant that users don't have to specify it if they don't want to.

ccgeonhee:
Oh, haha I see.

radarhere:
If there isn't a clear demand for this, then my preference is to keep the API simple.

PR #6707, opened 2022-10-31
RGBa to RGB color conversion mapping

### LINKED ISSUE #6706: convert("RGB") fails for RGBa
### What did you do?

Called PIL.Image.Image.convert(mode='RGB') for an RGBa image
```python3
img = PIL.Image.open('premultiplied_alpha.heic').convert(mode='RGB')
```

Conversion of data to RGB colorspace.

### What actually happened?

ValueError: conversion from RGBa to RGB not supported

* OS: macOS 12.6
* Python: 3.10
* Pillow: 9.3.0

bigcat88:
Fixes #6706

Added mapping for `RGBa` to `RGB` conversion

### DIFF
--- a/Tests/test_image_convert.py
+++ b/Tests/test_image_convert.py
@@ -338,3 +338,17 @@ def test_matrix_identity():
     # Assert
     # No change
     assert_image_equal(converted_im, im)
+
+
+def test_rgba_premultiplied_to_rgb():
+    im = Image.merge(
+        "RGBa",
+        [
+            Image.new(mode="L", size=(1, 1), color=128),
+            Image.new(mode="L", size=(1, 1), color=255),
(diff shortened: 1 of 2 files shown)


radarhere:
RGBa is "[true color with premultiplied alpha](https://pillow.readthedocs.io/en/stable/handbook/concepts.html#modes)". In main, converting from RGBA to RGBa makes the RGB values smaller
```python
>>> from PIL import Image
>>> im = Image.new("RGBA", (1, 1))
>>> im.putpixel((0, 0), (150, 175, 200, 100))
>>> im.convert("RGBa").load()[0, 0]
(59, 69, 78, 100)
```
so converting to RGB should make the RGB values bigger again, yes? Your test shows the RGB values getting smaller.

I've created #6708 instead.

PR #6708, opened 2022-10-31
Added conversion from RGBa to RGB

### LINKED ISSUE #6706: convert("RGB") fails for RGBa

radarhere:
Resolves #6706. Alternative to #6707

The new conversion function is just a copy of
https://github.com/python-pillow/Pillow/blob/a4ec9f331c11cd8ad05137bc78dff8197e8af2e3/src/libImaging/Convert.c#L463-L480
except with the last channel set to 255, as happens in other conversions to RGB.
https://github.com/python-pillow/Pillow/blob/a4ec9f331c11cd8ad05137bc78dff8197e8af2e3/src/libImaging/Convert.c#L436-L446

### DIFF
--- a/Tests/test_image_convert.py
+++ b/Tests/test_image_convert.py
@@ -104,6 +104,13 @@ def test_rgba_p():
     assert_image_similar(im, comparable, 20)
 
 
+def test_rgba():
+    with Image.open("Tests/images/transparent.png") as im:
+        assert im.mode == "RGBA"
+
+        assert_image_similar(im.convert("RGBa").convert("RGB"), im.convert("RGB"), 1.5)
+
+
 def test_trns_p(tmp_path):
     im = hopper("P")
     im.info["transparency"] = 0
--- a/src/libImaging/Convert.c
+++ b/src/libImaging/Convert.c
@@ -479,6 +479,25 @@ rgba2rgbA(UINT8 *out, const UINT8 *in, int xsize) {
     }
 }
 
+static void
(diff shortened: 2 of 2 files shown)


bigcat88:
Yep, you're right, that's my bad. Good Job! :)

bigcat88:
but I was always thinking that algorithm for premultiplied alpha is
v[0] = v[0] * (v[3] / 255)

nulano:
> but I was always thinking that algorithm for premultiplied alpha is
> v[0] = v[0] * (v[3] / 255)

That is correct. If we denote RGBA as the 8-bit non-premultiplied values and R'G'B'A as the 8-bit premultiplied values, then to get the premultiplied values you calculate R' = (R * A) / 255. (Doing the division after to avoid rounding errors.)

If you want to go back to the non-premultiplied values, you rearrange that equation to get R = (R' * 255) / A.

bigcat88:
Can this be merged before a new version is released?

radarhere:
Simply waiting for a review from another maintainer.

There is often more activity just before a release, and this is a relatively simple change, so in my experience I would expect this to make it in.

PR #6709, opened 2022-11-01
Added signed option when saving JPEG2000 images

### LINKED ISSUE #5502: Exposing the OpenJPEG sgnd parameter when JPEG2000 encoding
I'd like to be able to perform JPEG2K encoding of medical images using Pillow, particularly 16-bit signed greyscale (commonly used with CT and MR). Because of the limited supported `modes` in `Jpeg2KEncode.c` this is not currently possible as all input data is [treated as unsigned](https://github.com/python-pillow/Pillow/blob/eed337eacc58bb77b1d20c194507693ed4a40a80/src/libImaging/Jpeg2KEncode.c#L345).

Would it be acceptable if I provided a PR exposing the `sgnd` compression parameter to end users when saving? I ask because it feels more like a workaround than a true solution. Ideally I'd add support for `I` mode but OpenJPEG only supports a maximum of 31 bit samples.

radarhere:
Resolves #5502

Allows
```python
im.save(outfile, signed=True)
```
to save JPEG2000 `sgnd` images.


hugovk:
Shall we do it as a keyword-only argument? So it can be called like:

```python
im.save(outfile)
im.save(outfile, signed=True)
im.save(outfile, signed=False)
```

But not:

```python
im.save(outfile, True)
im.save(outfile, False)
```

That should make things clearer, no need to wonder what `True` or `False` mean here.

hugovk:
And let's add this to the release notes.

radarhere:
It should already be a keyword-only argument, as is the rule for `encoderinfo` keys.

I've added release notes.

hugovk:
Thanks!

PR #6714, opened 2022-11-03
Fix compiler error: accessing 64 bytes in a region of size 48

wiredfool:
src/_imaging.c:1842:17: warning: ‘ImagingTransform’ accessing 64 bytes in a region of size 48 [-Wstringop-overflow=]
       1842 |         imOut = ImagingTransform(
            |                 ^~~~~~~~~~~~~~~~~
       1843 |             imOut, imIn, IMAGING_TRANSFORM_AFFINE, 0, 0, xsize, ysize, a, filter, 1);
            |             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      src/_imaging.c:1842:17: note: referencing argument 8 of type ‘double *’

### DIFF
--- a/src/_imaging.c
+++ b/src/_imaging.c
@@ -1829,7 +1829,7 @@ _resize(ImagingObject *self, PyObject *args) {
         box[1] - (int)box[1] == 0 && box[3] - box[1] == ysize) {
         imOut = ImagingCrop(imIn, box[0], box[1], box[2], box[3]);
     } else if (filter == IMAGING_TRANSFORM_NEAREST) {
-        double a[6];
+        double a[8];
 
         memset(a, 0, sizeof a);
         a[0] = (double)(box[2] - box[0]) / xsize;


Yay295:
Just fyi, this isn't actually an error per se. `ImagingTransform` takes an "a" that is `double[8]`, but `ImagingTransformAffine` takes a `double[6]`, so passing a `double[6]` here is okay. I'd written a comment about this in my local code but hadn't committed it yet.

radarhere (approved):

wiredfool:
Yeah, I'd run through the analysis of if it was an issue or not and came to the conclusion that it was not a memory safety issue, so not a security problem. It's still a warning that should be silenced, at a cost of 2 extra doubles on the stack.

PR #4959, opened 2020-10-09
Add getlength and getbbox functions for TrueType fonts

nulano:
Third part of #4724, adding `getlength` and `getbbox` for TrueType fonts.

Changes proposed in this pull request:

 * Add functions `FreeTypeFont.getlength` and `ImageDraw.textlength` which measure the advance length of text instead of the rendered width/height. These functions are both very precise (return the exact values from FreeType/Raqm) and fast (2-3x faster depending on the string, font, and layout engine). `ImageDraw.textlength` falls back on `ImageDraw.textsize` if using a non-FreeType font. Helps #4651, https://github.com/python-pillow/Pillow/issues/3977#issuecomment-617349188, #4483, #3921, #4789.
 * Update `ImageDraw.multiline_text` to use this new value, see test case `test_imagefontctl.test_combine_multiline` for an example (essentially fixing multiline version of issue #4553).
 * Add functions `FreeTypeFont.getbbox` and `ImageDraw.[multiline_]textbbox`, inspired by https://github.com/python-pillow/Pillow/pull/784#issuecomment-48213245, which are a direct way to access the internal `FreeTypeFont.font.getsize` function, see https://github.com/python-pillow/Pillow/issues/4789#issuecomment-659609574 for the reasoning. This function will fail for non-FreeType fonts. Helps #3921, #4789.

PR #6722, opened 2022-11-07
Use fractional coordinates when drawing text

### LINKED ISSUE #3977: Font spacing changes

radarhere:
Resolves #3977

That issue attempts to draw two halves of a word separately, expecting them to occupy the same pixels as the complete word. The example given there is "par" and "adise" vs "paradise".

#4959 helped by adding `textlength`, but did not resolve the issue completely. After switching to `textlength`, [there was still a difference.](https://github.com/python-pillow/Pillow/issues/3977#issuecomment-873573618)

The ImageDraw text operation essentially has two parts - it generates a `mask` from the text characters with no awareness of the text position, and then pastes that mask onto the image at the relevant position. I had a thought - what if the position did matter, and text characters were rendered differently if they started halfway through a pixel?

With this PR,

```python
from PIL import ImageDraw, Image, ImageFont

img = Image.new('RGBA', (120, 80), color=(0, 0, 0))
draw = ImageDraw.Draw(img)
font = ImageFont.truetype('/System/Library/Fonts/Supplemental/Arial.ttf', size=48)

mask = font.getmask2("adise", "L", start=(0, 0))[0]
draw.draw.draw_bitmap((0, 2), mask, 0)

mask = font.getmask2("adise", "L", start=(0.4, 0))[0]
draw.draw.draw_bitmap((0, 42), mask, 0)

img.save('example.png')
```
produces "a", "d", "s" and "e" in the same location, but a different "i" for the first line compared to the second.

example

So I have added a `start` argument to `getmask` and `getmask2`, which ImageDraw's `text` method makes use of to pass through the fractional coordinates of the text. This resolves the issue, allowing "par" and "adise" to fill the same pixels as "paradise".

To try and future proof the API, I have set `start` to be a tuple, so that both horizontal and vertical starting positions can be specified.


src/PIL/ImageFont.py
```diff
@@ -588,6 +589,7 @@ def getmask(
         stroke_width=0,
         anchor=None,
         ink=0,
+        start=None,
```
hugovk, line 592: Needs adding to the docstring below so it'll show up at https://pillow.readthedocs.io/en/latest/reference/ImageFont.html#PIL.ImageFont.FreeTypeFont.getmask

src/PIL/ImageFont.py
```diff
@@ -672,6 +675,7 @@ def getmask2(
...
```
hugovk, line 678: Also needs adding to the docstring.

hugovk (approved):
Shall we mention the new argument in the release notes?

src/PIL/ImageFont.py
radarhere: Ok, done - https://pillow--6722.org.readthedocs.build/en/6722/reference/ImageFont.html#PIL.ImageFont.FreeTypeFont.getmask

src/PIL/ImageFont.py
radarhere: Done - https://pillow--6722.org.readthedocs.build/en/6722/reference/ImageFont.html#PIL.ImageFont.FreeTypeFont.getmask2

radarhere:
Sure. You can see the release notes at https://pillow--6722.org.readthedocs.build/en/6722/releasenotes/9.4.0.html

PR #6728, opened 2022-11-10
Added the _exclusive_fp check to Image.close()

### LINKED ISSUE #5309: Image.close() should not close an externally provided file pointer (when exclusive_fp is False)
When using context manager, Image does not close file pointer, as expected. It checks `getattr(self, "_exclusive_fp", False)` and closes the file pointer only if it's True.

```python
with open("file.png", "rb") as fp:
    assert fp.closed is False
    with Image.open(fp) as im:
        ...
    assert fp.closed is False  # OK
```

But the `Image.close()` method misses `_exclusive_fp` check and closes the file pointer unexpectedly:

```python
with open("test.png", "rb") as fp:
    assert fp.closed is False
    im = Image.open(fp)
    ...
    im.close()
    assert fp.closed is False  # AssertionError
```

delta1513:
Fixes #5309

 * Makes it so that when an image is opened using an exclusive file pointer, it will not close the original file pointer. This ensures that `Image.close()` is consistent with the functionality of `Image.__exit__()`

Changelog message:

Image.close() will no longer close an externally provided file pointer if one is provided.

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -570,12 +570,13 @@ class Image:
         more information.
         """
         try:
-            if getattr(self, "_fp", False):
-                if self._fp != self.fp:
-                    self._fp.close()
-                self._fp = DeferredError(ValueError("Operation on closed image"))
-            if self.fp:
-                self.fp.close()
+            if hasattr(self, "fp") and getattr(self, "_exclusive_fp", False):
+                if getattr(self, "_fp", False):
+                    if self._fp != self.fp:
+                        self._fp.close()
+                    self._fp = DeferredError(ValueError("Operation on closed image"))
+                if self.fp:
+                    self.fp.close()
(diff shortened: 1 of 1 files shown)


radarhere:
Hi. Just a heads up - if this doesn't receive much attention, it's not necessarily that there's anything that needs to be done. It's that it's not clear if Pillow should behave this way.

delta1513:
> Hi. Just a heads up - if this doesn't receive much attention, it's not necessarily that there's anything that needs to be done. It's that it's not clear if Pillow should behave this way.

I read the issue extensively and I understand. Thanks for keeping this updated. Changes are here if you need them at anytime in the future. Ping me if you need anything else done :)

radarhere:
Closing this, as I've closed #5309 in favour of backwards compatibility.

PR #6733, opened 2022-11-12
Fix PDF FlateDecode color error

NullYing:
PDF FlateDecode add params

- Fix FlateDecode Color Error


radarhere:
Hi. Would you be able to add tests? This would help prevent your fix from being accidentally broken in the future, and show us the problem that you are trying to solve.

radarhere:
Or at least post an image here with code that triggers the error?

radarhere:
Closing, unless an image and code can be posted to show the error.

While you say that this fixes a FlateDecode color error, there's nothing here to demonstrate that. I suspect that setting 'Colors' to 3 is not always appropriate, but can't go into more detail without knowing the cause of the issue.

PR #6735, opened 2022-11-12
Added MP Format Version when saving MPO

radarhere:
As noted in https://github.com/python-pillow/Pillow/issues/6720#issuecomment-1308617053, https://web.archive.org/web/20190227081740/http://www.cipa.jp/std/documents/e/DC-007_E.pdf states that
> 5.2.3.1. MP Format Version
> This tag specifies the version of the MP Extensions. An MP File that complies with this specification shall specify a 4 byte ASCII value of “0100.”

This PR adds "MP Format Version" when saving MPO images.


hugovk:
Thanks!

PR #6736, opened 2022-11-12
Remove ImagingAccessInit line() method

Yay295:
This method isn't used anywhere, and PyAccess doesn't implement it (because it's not used).


hugovk:
Thank you!

PR #5763, opened 2021-10-14
Add Tidelift alignment action and badge

aclark4life:
https://github.com/python-pillow/Pillow/issues/5762

Issue #5762, opened 2021-10-14
Align development requirements with Tidelift's Security-advised PyPI catalog


aclark4life:
FYI I'm going to be setting up a GitHub Action for Tidelift for this:

- https://support.tidelift.com/hc/en-us/articles/4406286307220-Using-with-Continuous-Integration

In the Tidelift parlance, we're going to make sure our development environment "aligns with Security-advised PyPI catalog". While it's obviously not critical that folks [use the latest docutils](https://github.com/python-pillow/Pillow/blob/master/requirements.txt#L7) to develop Pillow, it may be interesting to be able to say our development requirements "meet certain licensing, security or other standards". If nothing else, it will definitely help Tidelift as they try to improve "how lifters work with Tidelift and their subscribers". 👍 

<img width="1791" alt="Screen Shot 2021-10-14 at 1 34 01 PM" src="https://user-images.githubusercontent.com/72164/137367755-0bf5ef9c-1c54-4359-b1bc-06d328fa02e2.png">

hugovk:
> @hugovk I haven't considered doing it for runtime yet because to create a project on Tidelift, we need a `requirements.txt` file (or Pipfile) and all of our dependencies are in `setup.py` right? Not opposed to doing runtime … but this is a WIP.

Off the top of my head we don't have any required runtime dependencies (only optional).

> @hugovk Should I put a conditional on this? Seems to be running a lot 😂 [5c69dc7](https://github.com/python-pillow/Pillow/commit/5c69dc7c9a1b12eae14e6ab4823f73da6a5911e7)

When do we want it to run? Is it something that should fail a PR if something needs doing? Or just need to run it occasionally, like when something is pushed to `master`, like a PR merge?

https://support.tidelift.com/hc/en-us/articles/4406286307220-Using-with-Continuous-Integration suggests every PR:

> This action will run a check on every pull request to see if all of your dependencies are included in your catalog's approved release list. 

What sort of failures can we expect, what action would I as a developer need to take if something failed?

aclark4life:
@hugovk Sure go ahead

aclark4life:
> What is the action actually doing? Gathering stats for the catalogue?


Here is some output from a Jenkins running at my house to _build a catalog_.

```

+ make scan
tidelift alignment save --debug || true
time="2021-11-30T00:09:51-05:00" level=debug msg="GET https://download.tidelift.com/cli/tidelift-cli.version"
[...]

PR #6597, opened 2022-09-19
Add OpenSSF Best Practices badge

hugovk:
I started filling out the sections at https://bestpractices.coreinfrastructure.org/en/projects/6331 and we're currently at 76%:

image

Still some more to check, but we can already add the badge:

image

# Preview

## README

image

https://github.com/hugovk/Pillow/tree/add-openssf-best-practices-badge

## Docs

image

https://pillow--6597.org.readthedocs.build/en/6597/

PR #6739, opened 2022-11-14
Remove Tidelift alignment action and badge

aclark4life:
Not sure if we still care about this? cf. #5762 #5763


aclark4life:
Also not sure why the credential failures in CI are happening :smile:

radarhere:
I've added commits for the "and badge" part of the title

auvipy:
are you removing this in favor of OpenSSF?

aclark4life:
@auvipy No, this was an experimental feature that served its purpose: testing catalog alignments for Tidelift. As far as I can tell, we haven't added any actions for OpenSSF yet e.g. https://github.com/ossf/scorecard

radarhere:
Regarding the removal of the badge, we have separately added an OpenSSF badge - #6597

PR #6747, opened 2022-11-17
Added WebP encoding with exact config

ashafaei:
When encoding a transparent image with WebP, even if specified to be lossless, it'll drop the information that is not visible. This is normally fine, except for applications where you need to undo the mask or make additional changes later.

There's an option in the `WebPConfig` data structure named `exact`:

```C++
     int exact;           // if non-zero, preserve the exact RGB values under
                          // transparent area. Otherwise, discard this invisible
                          // RGB information for better compression. The default
                          // value is 0.
```

This option is not exposed to the users. In this pull request, I make this option available.

The parameter `exact` is True/False at the python level but translated to `0 or 1 (int)` to match that of `WebPConfig`.

 * Updated the binding at `src/_webp.c`.
 * Updated the Python encoder at `src/PIL/WebPImagePlugin.py`.
 * Added documentation for the new flag `docs/handbook/image-file-formats.rst`.
 * Added the unit-test `Tests/test_file_webp_alpha.py`.

I tried to follow the guidelines as much as possible. Let me know if I'm missing anything.
Thanks.

### DIFF
--- a/docs/handbook/image-file-formats.rst
+++ b/docs/handbook/image-file-formats.rst
@@ -1127,7 +1127,7 @@ The :py:meth:`~PIL.Image.Image.save` method supports the following options:
 **exact**
     If true, preserve the transparent RGB values. Otherwise, discard
     invisible RGB values for better compression. Defaults to false.
-    Requires LibWebP 0.5.0 or later.
+    Requires libwebp 0.5.0 or later.
 
 **icc_profile**
     The ICC Profile to include in the saved file. Only supported if
--- a/docs/handbook/image-file-formats.rst
+++ b/docs/handbook/image-file-formats.rst
@@ -1124,6 +1124,10 @@ The :py:meth:`~PIL.Image.Image.save` method supports the following options:
 **method**
     Quality/speed trade-off (0=fast, 6=slower-better). Defaults to 4.
 
+**exact**
+    If true, preserve the transparent RGB values. Otherwise, discard
+    invisible RGB values for better compression. Defaults to false.
+
 **icc_profile**
     The ICC Profile to include in the saved file. Only supported if
     the system WebP library was built with webpmux support.
--- a/docs/handbook/image-file-formats.rst
+++ b/docs/handbook/image-file-formats.rst
@@ -1127,6 +1127,7 @@ The :py:meth:`~PIL.Image.Image.save` method supports the following options:
 **exact**
     If true, preserve the transparent RGB values. Otherwise, discard
     invisible RGB values for better compression. Defaults to false.
+    Requires LibWebP 0.5.0 or later.
 
 **icc_profile**
     The ICC Profile to include in the saved file. Only supported if
--- a/docs/releasenotes/9.4.0.rst
+++ b/docs/releasenotes/9.4.0.rst
@@ -41,8 +41,9 @@ Added the ``exact`` encoding option for WebP
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 The ``exact`` encoding option for WebP is now supported. The WebP encoder
-removes the hidden RGB values for better compression by default. By setting
-this option to ``True``, the encoder will keep the hidden RGB values.
+removes the hidden RGB values for better compression by default in libwebp 0.5
+or later. By setting this option to ``True``, the encoder will keep the hidden
+RGB values.
 
 Security
 ========
--- a/docs/releasenotes/9.4.0.rst
+++ b/docs/releasenotes/9.4.0.rst
@@ -41,7 +41,7 @@ Added the ``exact`` encoding option for WebP
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 The ``exact`` encoding option for WebP is now supported. The WebP encoder
-removes the hidden RGB values for better compression by default. By setting 
+removes the hidden RGB values for better compression by default. By setting
 this option to ``True``, the encoder will keep the hidden RGB values.
 
 Security
--- a/docs/releasenotes/9.4.0.rst
+++ b/docs/releasenotes/9.4.0.rst
@@ -37,6 +37,13 @@ support a ``start`` argument. This tuple of horizontal and vertical offset
 will be used internally by :py:meth:`.ImageDraw.text` to more accurately place
 text at the ``xy`` coordinates.
 
+Added the ``exact`` encoding option for WebP
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The ``exact`` encoding option for WebP is now supported. The WebP encoder
+removes the hidden RGB values for better compression by default. By setting 
+this option to ``True``, the encoder will keep the hidden RGB values.
+
 Security
 ========
 
--- a/src/_webp.c
+++ b/src/_webp.c
@@ -576,6 +576,7 @@ WebPEncode_wrapper(PyObject *self, PyObject *args) {
     int lossless;
     float quality_factor;
     int method;
+    int exact;
     uint8_t *rgb;
     uint8_t *icc_bytes;
     uint8_t *exif_bytes;
@@ -597,7 +598,7 @@ WebPEncode_wrapper(PyObject *self, PyObject *args) {
 
     if (!PyArg_ParseTuple(
             args,
-            "y#iiifss#is#s#",
+            "y#iiifss#iis#s#",
             (char **)&rgb,
             &size,
             &width,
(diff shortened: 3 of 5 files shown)


radarhere:
This is [currently failing on CentOS 7](https://github.com/python-pillow/Pillow/actions/runs/3492376884/jobs/5846882675#step:6:77), because it is using 0.3.0. `exact` was [added in 0.5.0.](https://github.com/webmproject/libwebp/blob/fad0ece7edec90dbead3cfbe19f2bc9d77a410e2/NEWS#L155-L168)

ashafaei:
@radarhere Thanks for reviewing the pull request.
I see three potential directions forward:

1. Update the `pythonpillow/centos-7-amd64:latest` docker image to install `libwebp` outside of the package manager. The latest version of the CentOS package manager is `0.3.0`. We could even build inside the container. I don't know how the PIL binaries are packaged and distrbuted -- if there's a way to ensure the latest WebP is pulled/embedded correctly on client machines, then this could be fine.
2. Add an `#IFDEF/#IFNDEF` to the `src/_webp.c` code to skip that line if the WebP version is below 0.5.0. That would require updating the build scripts to include the correct `DEFINE` on CentOS7.
3. Drop this pull request altogether in favour of maintaining complete backward compatibility.

Let me know if I'm missing a possibility here. I'd be happy to work on 1, or 2, or even drop this altogether.
Thanks.

radarhere:
Pillow does distribute Python wheels, but we're building from source in our CI presumably to support users who also build from source, and if you install libwebp-devel on CentOS 7, that's what you get.

If you want to suggest that we deprecate support for older versions of libwebp, that's perfectly reasonable. It's just we would probably want to include a deprecation period, and I imagine you'd prefer this feature sooner rather than later.

I think using `#ifdef` is the best solution to the immediate problem.

ashafaei:
Thanks @radarhere for the useful feedback.

Turns out `LibWebP` already has a definition for 
```c
#define WEBP_ENCODER_ABI_VERSION 0x020f    // MAJOR(8b) + MINOR(8b)
```
The fix is a simple
```c
#if WEBP_ENCODER_ABI_VERSION >= 0x0209
    // the exact flag is only available in libwebp 0.5.0 and later
    config.exact = exact;
#endif
```
I verified that `0.5.0` is the earliest version with `exact` with a version `0x0209`. The version before that is `4.4` with `0x0202`.

Interestingly the unit test also passes. I think they started removing the transparent part at that exact version, so the tests pass.

Let me know if there's anything else I could do. Thanks!

hugovk:
Looking at https://github.com/webmproject/libwebp/blob/main/NEWS, version 0.5.0 was released in 2015, so I think we can also deprecate it (in another PR).

We normally deprecate for at least a year and then drop in a major version.

Re: https://github.com/python-pillow/Pillow/milestones

If we deprecate now, Pillow 10 in July 2023 is too soon for removal, so Pillow 11 in October 2024 is the next one that can drop it. That's a nice long deprecation period, and CentOS 7 is EOL in June 2024.

> I verified that `0.5.0` is the earliest version with exact with a version `0x0209`. The version before that is `4.4` with `0x0202`.

Do you mean `0.4.4` with `0x0202`?

ashafaei:
@hugovk , yeah, sorry, `0.4.4` is `0x0202`-- so the condition I put on
```c
#if WEBP_ENCODER_ABI_VERSION >= 0x0209
    // the exact flag is only available in libwebp 0.5.0 and later
    config.exact = exact;
#endif
```
should be correct.

I'm probably not part of the deprecation conversation, but I think if there's no cost in maintaining support, it may be worth keeping around. `WebP > 0.5.0` to support a minor flag that may not be even useful to everyone seems to be too small of a reason to drop support for an entire OS. Unless, of course, you are also experiencing compatibility issues with the other features in Pillow and CentOS7.

src/_webp.c
```diff
@@ -633,6 +635,10 @@ WebPEncode_wrapper(PyObject *self, PyObject *args) {
     config.lossless = lossless;
     config.quality = quality_factor;
     config.method = method;
+#if WEBP_ENCODER_ABI_VERSION >= 0x0209
+    // the exact flag is only available in libwebp 0.5.0 and later
```
Yay295, line 639: should probably quote "exact" in this comment to be more clear

ashafaei: Done! Thanks for the feedback.

radarhere (approved):

docs/handbook/image-file-formats.rst
```diff
@@ -1124,6 +1124,11 @@ The :py:meth:`~PIL.Image.Image.save` method supports the following options:
 **method**
     Quality/speed trade-off (0=fast, 6=slower-better). Defaults to 4.
 
+**exact**
+    If true, preserve the transparent RGB values. Otherwise, discard
+    invisible RGB values for better compression. Defaults to false.
+    Requires LibWebP 0.5.0 or later.
```
radarhere, line 1130:
```suggestion
    Requires libwepb 0.5.0 or later.
```
While it is "WebP", a quick Google leads me to believe it is "libwebp". See https://en.wikipedia.org/wiki/WebP

ashafaei: Yeah, you're right. Done.

radarhere (approved):

docs/handbook/image-file-formats.rst
```diff
@@ -1124,6 +1124,11 @@ The :py:meth:`~PIL.Image.Image.save` method supports the following options:
 **method**
     Quality/speed trade-off (0=fast, 6=slower-better). Defaults to 4.
 
+**exact**
```
hugovk, line 1127: These are keyword arguments, shall we document them in alphabetical order for ease of reference?

hugovk (commented):
Let's add this API addition to the release notes:

* https://pillow.readthedocs.io/en/latest/releasenotes/9.4.0.html
* https://github.com/python-pillow/Pillow/blob/main/docs/releasenotes/9.4.0.rst

docs/handbook/image-file-formats.rst
ashafaei: It currently reads,
```
lossless -> quality -> method -> exact -> icc_profile -> exif
```
Is it preferable to read,
```
exact -> exif -> icc_profile -> lossless -> method -> quality
```
?
If we put `exact` before `exif` it breaks the `icc_profile, exif` connection (as the common parameters across different formats.) 
if we put `exact` after `exif` it breaks the connection between the other 3 encoding parameters, `lossless`, `quality`, `method`. I think the sequence `lossless -> quality -> method -> exact` orders the priority of encoding parameters nicely.
Anyway, if you could let me know what you had in mind, I'd be happy to apply the change. thanks.

hugovk: Ah, let's not worry about it and leave it as you have now :)

Thanks!

hugovk (approved):
Thank you!

hugovk:
Oh one little change, I went ahead and committed it directly, hope that's okay: removed the code formatting from the title.

This follows the [Google developer docs style guide](https://developers.google.com/style/code-in-text#grammatical-treatment-of-code-elements) and is similar to, for example,  
https://pillow.readthedocs.io/en/stable/releasenotes/9.1.0.html#added-mct-and-no-jp2-options-for-saving-jpeg-2000

Yay295:
I think you linked to the wrong part of Google's style guide, since the change you made isn't related to grammar. [The part about headings and titles](https://developers.google.com/style/headings#heading-and-title-format) does say to "Avoid using code items in headings when possible.", but I think they mean you should not include code phrases at all, not that code phrases shouldn't be formatted, because the start of the next sentence is "If you must mention a code item in a heading, ..." implying that by following the first sentence you would not mention a code item in a heading.

I didn't find anything in their style guide specifically about code formatting in titles, but if you think of this as a list rather than just titles, then there is a reference on the [lists page](https://developers.google.com/style/lists#numbered-lettered-bulleted-lists): "If the item is entirely in code font, don't add end punctuation.". So code formatting in list items is allowed.

hugovk:
Right you are, I've undone that change!

docs/releasenotes/9.4.0.rst
```diff
@@ -37,6 +37,13 @@ support a ``start`` argument. This tuple of horizontal and vertical offset
 will be used internally by :py:meth:`.ImageDraw.text` to more accurately place
 text at the ``xy`` coordinates.
 
+Added the ``exact`` encoding option for WebP
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The ``exact`` encoding option for WebP is now supported. The WebP encoder
+removes the hidden RGB values for better compression by default. By setting
+this option to ``True``, the encoder will keep the hidden RGB values.
```
radarhere, line 45:
```suggestion
removes the hidden RGB values for better compression by default in libwebp 0.5
or later. By setting this option to ``True``, the encoder will keep the hidden
RGB values.
```

ashafaei:
There's so much to learn from y'all. I committed the latest changes. Thanks for the suggestions.

hugovk:
Thank you for your contribution!

Issue #6641, opened 2022-10-05
Enhance EXIF handling


cobi-s:
The `getexif()` method of `PIL.Image` could use some enhancements, which most likely require an API bump. I'm just listing some ideas/shortcomings of the current implementation here, which I would love to see addressed (and could possibly contribute to, if the changes are agreed upon).

- [ ] The return type of `getexif()` is `Exif`. But without further processing, it's really only IFD0. However, already returning an IFD makes the `get_ifd()` method look out of place. Therefore, I recommend to implement the `Exif` class as a dict linking to all extracted IFDs.
- [x] Either providing specialized methods for common IFDs (e.g. `Exif.get_ifd0()`) or named constants for common IFDs (e.g. `Exif.get_ifd(ExifTags.IFD.IFD0)`) would improve the usage of the API a lot. (#6748)
- [x] Implement tag names for missing IFDs (especially IFD1 and Interop) (#6724)
- [x] Some tags are purely technical and should never be edited by hand (the IFD offsets come to mind here). It would be nice to have a flag not to return them to the user. (#6762)
- [ ] Some integer tag values encode a human readable description, for example Flash and LightSource. It would be great to provide named constants for those as well as methods to return the description as a string.

I can't say anything about the feasability regarding licencing, but maybe the tag names could be scraped from the Exiftool website (the quasi reference implementation of the Exif format): https://exiftool.org/TagNames/EXIF.html

There are already other python libraries for Exif handling, but they all suffer from fragmentation. As the Pillow library already does the heavy lifting of extracting the IFDs anyway, I think providing a convenient and nearly complete view of the data can be a significant improvement to the project.

cobi-s:
> Isn't this just part of the unfortunate reality that https://exiftool.org/TagNames/EXIF.html lists EXIF tags, and one of the groups listed is an ExifIFD? I would have thought the names were inherently confusing.

I don't think so. I might be totally wrong here, though. As far as I understand, the Exif segment (which is returned by `Image.info.get('exif')` is basically a linked list of IFDs. First one is IFD0, which contains the information that `getexif()` returns. This IFD links to the next one (which is called ExifIFD by ExifTool) which contains most of the camera information. Then on the the next IFD (Interop) and so on. Yes, the names are a bit confusing. What I was trying to say is that `getexif()` returns an actual IFD, but is semantically closer to a handle to the whole EXIF information segment of the image.

[...]

PR #6748, opened 2022-11-19
Added IFD enum to ExifTags

### LINKED ISSUE #6777: Getting IFD1 exif information
Follow-up to #6641 

A test image could be the following:
https://www.sun.ac.za/english/test-site/EXIF/sa.jpg

Please find below a code sample that uses the piexif library to extract all available exif information.

I wasn't able to retrieve the same information using only Pillow. In particular, not having the offsets to IFD1 and the embedded thumbnail makes it impossible (for me) to retrieve the tags from IFD1 as well as the raw data for the embedded thumbnail.

```python
from PIL import Image
from PIL.ExifTags import TAGS, GPSTAGS
import piexif

def lookup_tag(tag, ifd):
    lu = GPSTAGS if ifd == 'GPS' else TAGS
    try:
        return f'0x{tag:04x}: {lu[tag]}'
    except:
        return f'0x{tag:04x}: ???'

def print_dict(data):
    for ifd, val in data.items():
        print(f'{ifd}:')
        if ifd == 'thumbnail':
            print(f'    {len(val)} bytes')
        else:
            for tag in val:
                print(f'    {lookup_tag(tag, ifd)}: {val[tag]}')

path = 'sa.jpg'

with Image.open(path) as i:
    d = piexif.load(i.info['exif'])
    print_dict(d)
```

radarhere:
Helps #6641 by adding "named constants for common IFDs".

Adds the following to ExifTags.
```python
class IFD(IntEnum):
    Exif = 34665
    GPSInfo = 34853
    Makernote = 37500
    Interop = 40965
```

I updated other parts of Pillow's code to use these new constants internally. In doing so, I found
https://github.com/python-pillow/Pillow/blob/df8e87291254dc22415a0c6e371950e79311dbf2/src/PIL/Image.py#L3618
I can only imagine that `0x8225` was a typo when I added it in https://github.com/python-pillow/Pillow/commit/4b14f0102d8fa585c900ff8a174defb2452c042b, so I'm removing it here.

Resolves #6777 by adding `IFD.IFD1` as well, so that IFD1 data can be read, and then modifying `get_child_images()` to make use of IFD1 data to read thumbnails through the "[ThumbnailOffset in IFD1](https://exiftool.org/TagNames/EXIF.html)".


radarhere:
To help #6777, I've added `IFD.IFD1` as well, so that IFD1 data can be read.

src/PIL/Image.py
```diff
@@ -1447,6 +1454,49 @@ def _reload_exif(self):
...
+                try:
+                    thumbnailOffset += self._exif_offset
+                except AttributeError:
+                    pass
+                self.fp.seek(thumbnailOffset)
+                data = self.fp.read(ifd.get(514))
+                fp = io.BytesIO(data)
+
+            with open(fp) as im:
+                if thumbnailOffset is None:
```
hugovk, line 1490: snake_case

```suggestion
            thumbnail_offset = ifd.get(513)
            if thumbnail_offset is not None:
                try:
                    thumbnail_offset += self._exif_offset
                except AttributeError:
                    pass
                self.fp.seek(thumbnailOffset)
                data = self.fp.read(ifd.get(514))
                fp = io.BytesIO(data)

            with open(fp) as im:
                if thumbnail_offset is None:
```

hugovk (approved):

src/PIL/Image.py
radarhere: `self.fp.seek(thumbnailOffset)` was missed in the middle, so I've applied this as a separate commit

PR #6749, opened 2022-11-19
Added LightSource tag values to ExifTags

### LINKED ISSUE #6641: Enhance EXIF handling

radarhere:
Resolves #6641

> Some integer tag values encode a human readable description, for example Flash and LightSource. It would be great to provide named constants for those as well as methods to return the description as a string.


hugovk (approved):
Do you think this is worth mentioning in the release notes? If so, there are some other recent additions to this file that could join it.

radarhere:
Ok, I've pushed a commit.

hugovk:
Thank you!

PR #6754, opened 2022-11-22
chore: enabled syntax highlighting

0xflotus:
I enabled the syntax highlighting feature in the `image-file-formats.rst`

### DIFF
--- a/docs/handbook/image-file-formats.rst
+++ b/docs/handbook/image-file-formats.rst
@@ -95,7 +95,7 @@ method with the following parameters to affect how Ghostscript renders the EPS
     Affects the scale of the resultant rasterized image. If the EPS suggests
     that the image be rendered at 100px x 100px, setting this parameter to
     2 will make the Ghostscript render a 200px x 200px image instead. The
-    relative position of the bounding box is maintained.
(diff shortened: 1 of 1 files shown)


hugovk:
Thanks for the PR, however the default for literal blocks (`::`) is already Python and we've not changed it using `highlight_language`.

* https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#showing-code-examples
* https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#rst-literal-blocks
* https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-code-block

So this PR doesn't change anything other than making it more explicit, which I don't think we need in this particular case.

0xflotus:
Ok, but there is already a use of it in the lines 1402 and 1413:

https://github.com/python-pillow/Pillow/blob/73038257e3b6f48d8909175b24f1c4e77b977b4b/docs/handbook/image-file-formats.rst#L1402-L1417

So I thaught it would be consistent.

radarhere:
I've created PR #6958 as a possible alternative to this.

radarhere:
#6958 has been merged instead.

PR #6758, opened 2022-11-24
Added getxmp() to WebPImagePlugin

### LINKED ISSUE #6757: Can we get `getxmp` method for WebP images?
PNG, Jpeg and TIFF already have this method.
IMHO, it would be nice to add such a thing to WebP.

```python
Image.open("image.webp").getxmp()
```

I know it's stored in `info["xmp"]`, but a generic method would be more convenient.

radarhere:
Resolves #6757

### DIFF
--- a/Tests/test_file_webp_metadata.py
+++ b/Tests/test_file_webp_metadata.py
@@ -11,6 +11,11 @@ pytestmark = [
     skip_unless_feature("webp_mux"),
 ]
 
+try:
+    from defusedxml import ElementTree
+except ImportError:
+    ElementTree = None
+
 
 def test_read_exif_metadata():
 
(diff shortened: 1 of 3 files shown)


bigcat88:
Are you sure that this will return dictionary a not a bytes when xmp is present?

bigcat88:
Meant return like this:

https://github.com/python-pillow/Pillow/blob/f27631601bd307049d73782a63167eb4b12de145/src/PIL/PngImagePlugin.py#L1021
https://github.com/python-pillow/Pillow/blob/f27631601bd307049d73782a63167eb4b12de145/src/PIL/JpegImagePlugin.py#L482
https://github.com/python-pillow/Pillow/blob/f27631601bd307049d73782a63167eb4b12de145/src/PIL/TiffImagePlugin.py#L1196

radarhere:
Yep, you're right. I've updated the commit.

hugovk:
Let's document this in the release notes and at https://pillow--6758.org.readthedocs.build/en/6758/reference/plugins.html#PIL.WebPImagePlugin.WebPImageFile.getxmp

radarhere:
Ok, https://pillow--6758.org.readthedocs.build/en/6758/reference/plugins.html#PIL.WebPImagePlugin.WebPImageFile.getxmp and https://pillow--6758.org.readthedocs.build/en/6758/releasenotes/9.4.0.html#getxmp now document the new addition.

hugovk:
Thanks!

PR #6759, opened 2022-11-25
Only install pyroma 4.0 in Cygwin

radarhere:
[pyroma 4.1 was released yesterday](https://github.com/regebro/pyroma/tags), and since then Cygwin has started failing in main - https://github.com/python-pillow/Pillow/actions/runs/3544062653

Issue #6216, opened 2022-04-15
JPEG not detected when installing Pillow under Cygwin64


biogberg:
Hello,

Attempting to install the current version of Pillow (9.1.0) fails with Python-3.9.10 under Cygwin64. I attempted installing both via pip3 and via python3 setup.py build. Both fail the same way.

Installation succeeds until the configuration stops with a message that the required headers and/or library files for jpeg were not found.

Jpeg packages (both jpeg and openjpeg) are installed including the developer packages. Both header files and library files are in fact present  in the right directories.

I had a hunch that this might be a library file naming issue. In /usr/lib the following jpeg-related files are present:
/usr/lib/libjpeg.dll.a
/usr/lib/libopenjpeg.dll.a

To test, I built and installed a static libjpeg.a and put libjpeg.a file in /usr/local/lib. That done, the build succeeded to the point that the extension was built. However, most other supporting libraries were not detected despite that most of them were in fact installed (including developer packages); the corresponding packages all exclusively install shared libraries (dll's), not static:

The output from the build produced:
```
--------------------------------------------------------------------
PIL SETUP SUMMARY
--------------------------------------------------------------------
version      Pillow 9.1.0
platform     cygwin 3.9.10 (main, Jan 20 2022, 21:37:52)
             [GCC 11.2.0]
--------------------------------------------------------------------
--- JPEG support available
*** OPENJPEG (JPEG2000) support not available
--- ZLIB (PNG/ZIP) support available
*** LIBIMAGEQUANT support not available
*** LIBTIFF support not available
*** FREETYPE2 support not available
*** RAQM (Text shaping) support not available
*** LITTLECMS2 support not available
*** WEBP support not available
*** WEBPMUX support not available
*** XCB (X protocol) support not available
--------------------------------------------------------------------
To add a missing option, make sure you have the required
library and headers.
```
There is in fact a zlib static library in /usr/lib and, thus, presence of zlib was detected.

So either:

1. the build currently only works with static libraries, or
2. there is a an issue with properly detecting the shared libraries.

Under Cygwin the dll files are located in /usr/bin with names such as cygtiff-6.dll, cygjpeg-8.dll, cygopenjpeg-1.dll, cygimagequant-0.dll etc. The link files have names such as libjpeg.dll.a etc.

* OS: Cygwin64 under Windows 10 (64 bit) updated
* Python: 3.9.10 (Cygwin64 version updated)
* Pillow: 9.1.0

[...]

PR #6760, opened 2022-11-25
Use stdlib for setuptools on Cygwin

radarhere:
Alternative to #6759

[pyroma 4.1 was released yesterday](https://github.com/regebro/pyroma/tags). It now [requires setuptools >= 61](https://github.com/regebro/pyroma/blob/28387e8c6b3dd2fa588e78fb0bf6af348d04a532/setup.cfg#L41). This has started Cygwin failing in main - https://github.com/python-pillow/Pillow/actions/runs/3544062653

This PR uses `SETUPTOOLS_USE_DISTUTILS=stdlib`, [as suggested previously.](https://github.com/python-pillow/Pillow/issues/6216#issuecomment-1122340162)

### DIFF
--- a/.github/workflows/test-cygwin.yml
+++ b/.github/workflows/test-cygwin.yml
@@ -5,7 +5,7 @@ on: [push, pull_request, workflow_dispatch]
 permissions:
   contents: read
 
-concurrency: 
+concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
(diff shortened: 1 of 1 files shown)


hugovk (approved):
Thanks!

The plan is that distlib will be removed from the stdlib in Python 3.12: 

* https://peps.python.org/pep-0632/
* https://setuptools.pypa.io/en/latest/deprecated/distutils-legacy.html

Python 3.12 final is due out in October 2023, with an alpha already out: 

* https://peps.python.org/pep-0693/ 

We should start testing 3.12 at some point, I think before the beta period (May 2023) would make sense, because there shouldn't be any feature changes during the beta, and if there's big problems with setuptools/distlib we may be able to give useful feedback before it's too late. (Having said that, for Cygwin we've not been testing pre-releases.)

So we may need to revisit this. But we may see more churn in setuptools before then, so I think this is fine for the time being. 👍

PR #6756, opened 2022-11-24
Document how to install Pillow from a directory

radarhere:
Suggestion for #6755

Issue #6755, opened 2022-11-23
Documentation incorrect for Linux build from source


dnygren:
I asked and got answered this question on how to build Pillow on Linux from source on StackOverflow:
https://stackoverflow.com/questions/74537064/unable-to-build-pillow-python-imaging-library-from-source-on-linux

Based on the response I received, I am filing an issue because there may be incorrect documentation for building Pillow from source on Linux.

### What did you do?

I tried to build Pillow from source by following the directions at https://pillow.readthedocs.io/en/stable/installation.html#building-from-source

### What did you expect to happen?

Have Pillow built from source, including my changes.

### What actually happened?

Pillow appeared to build from a different downloaded copy of Pillow.

### What are your OS, Python and Pillow versions?

* OS: 
```console
$ uname -a
Linux mms-labpc1 5.10.0-18-amd64 #1 SMP Debian 5.10.140-1 (2022-09-02) x86_64 GNU/Linux
```
* Python: 
```console
$ python3 --version
Python 3.9.2
```
* Pillow: 
9.4.0 (unreleased)

The following is how to reproduce the problem. Note I am following the Linux build from source Pillow documentation.
```console
$ git clone https://github.com/python-pillow/Pillow.git
Cloning into 'Pillow'...
remote: Enumerating objects: 67869, done.
remote: Counting objects: 100% (163/163), done.
remote: Compressing objects: 100% (95/95), done.
remote: Total 67869 (delta 90), reused 128 (delta 68), pack-reused 67706
Receiving objects: 100% (67869/67869), 83.17 MiB | 9.39 MiB/s, done.
Resolving deltas: 100% (50202/50202), done.

$ sudo apt-get install python3-dev python3-setuptools
[sudo] password for dan: 
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
python3-dev is already the newest version (3.9.2-3).
python3-setuptools is already the newest version (52.0.0-4).
0 upgraded, 0 newly installed, 0 to remove and 8 not upgraded.
```
Remove all the source files from libImaging directory to prove its not building from source:
```console
$ cd Pillow
$ rm ./src/libImaging/*
```
Continue to follow the Building From Source instructions at https://pillow.readthedocs.io/en/stable/installation.html#building-from-source for Debian Linux:
```console
$ python3 -m pip install --upgrade pip
Defaulting to user installation because normal site-packages is not writeable
[...]

PR #6764, opened 2022-11-28
Use Sphinx Inline Tabs to organise installation per OS

hugovk:
Follow on from https://github.com/python-pillow/Pillow/issues/6755 / https://github.com/python-pillow/Pillow/pull/6756

 * Use [Sphinx Inline Tabs](https://sphinx-inline-tabs.readthedocs.io/) to group the installation instructions by operating system
 * This helps hide irrelevant information on long, detailed page
 * If, for example, you click the macOS tab in the first tab group, it also opens the macOS tab in the second group
 * Because the subheaders are gone, I added in `.. _Linux Installation:` etc. to avoid linkrot
 * I re-ordered them to go: Linux, macOS, Windows, [Windows variant,] FreeBSD, [Android]
   * My thinking was to put the main three first, in order of "likelihood" to needs to follow these instructions, then the others afterwards, but happy to change it

image

image

Plus some other cleanup:

* Like our top-level Makefile, inline the `.PHONY` targets to help avoid omissions in the docs `Makefile` (and `texinfo`, `info`, `livehtml` and `serve` were missing)
* Add `serve` and `livehtml` to `make -C docs help`
* EditorConfig: three-space indents for RST

# Demo

https://pillow--6764.org.readthedocs.build/en/6764/installation.html

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -13,6 +13,10 @@ indent_style = space
 
 trim_trailing_whitespace = true
 
+[*.rst]
+# Three-space indentation
+indent_size = 3
+
 [*.yml]
 # Two-space indentation
 indent_size = 2
--- a/.editorconfig
+++ b/.editorconfig
@@ -15,7 +15,7 @@ trim_trailing_whitespace = true
 
 [*.rst]
 # Three-space indentation
-indent_size = 3
+indent_size = 4
 
 [*.yml]
 # Two-space indentation
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,7 +14,7 @@ indent_style = space
 trim_trailing_whitespace = true
 
 [*.rst]
-# Three-space indentation
+# Four-space indentation
 indent_size = 4
 
 [*.yml]
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -20,6 +20,8 @@ I18NSPHINXOPTS  = $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) .
 help:
 	@echo "Please use \`make <target>' where <target> is one of"
 	@echo "  html       to make standalone HTML files"
+	@echo "  serve      to start a local server for viewing docs"
+	@echo "  livehtml   to start a local server for viewing docs and auto-reload on change"
 	@echo "  dirhtml    to make HTML files named index.html in directories"
 	@echo "  singlehtml to make a single large HTML file"
 	@echo "  pickle     to make pickle files"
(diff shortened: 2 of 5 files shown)


docs/Makefile
```diff
@@ -39,49 +40,57 @@ help:
...
 
+.PHONY: htmlhelp
 htmlhelp:
 	$(MAKE) install-sphinx
 	$(SPHINXBUILD) -b htmlhelp $(ALLSPHINXOPTS) $(BUILDDIR)/htmlhelp
 	@echo
 	@echo "Build finished; now you can run HTML Help Workshop with the" \
 	      ".hhp project file in $(BUILDDIR)/htmlhelp."
 
+.PHONY:  qthelp
```
radarhere, line 93:
```suggestion
.PHONY: qthelp
```

.editorconfig
```diff
@@ -13,6 +13,10 @@ indent_style = space
 
 trim_trailing_whitespace = true
 
+[*.rst]
+# Three-space indentation
+indent_size = 3
```
radarhere, line 18: I was wondering why 3, and then I found
https://devguide.python.org/documentation/style-guide/
> All reST files use an indentation of 3 spaces

hugovk: Yeah, it's a bit weird but then so is RST :)

https://stackoverflow.com/q/48311159/724176 says there's no "official" requirement.

Another random Google result:

> In reST, the indentation of a block of lines is often important. The exact number of spaces, which are used to indent a block of text, does not matter. But what does matter, is that all lines of the block are indented with exactly the same number of spaces.
>
> We use the convention of 4 spaces per indenting level (see [Whitespace and indentation](https://docs.typo3.org/m/typo3/docs-how-to-document/main/en-us/GeneralConventions/CodingGuidelines.html#cgl-indenting)).

https://docs.typo3.org/m/typo3/docs-how-to-document/main/en-us/WritingReST/BasicRestSyntax.html

The installation page uses 3 spaces. I did a cursory search of the docs and `git grep "  \.\." docs` suggests we're already using 3-space indents, but happy to remove this config if you prefer.

On the other hand, if we are going to use 3-space, this helps the IDE know how many to add when tabbing.

radarhere: Opening up the files inside docs/handbook, I don't think we (at least, not consistently) using 3 spaces at the moment. I think that the mix of 3 spaces for RST and 4 for Python indentation is weird.

But given that there's already a variety of style rules that we have to follow when writing documentation, I'm not overly concerned about adding another, and I think that specifying a value is helpful.

radarhere (approved):

.editorconfig
hugovk: Shall we prefer 4 over 3? I'm not too bothered which, and let's not bother mass-reindenting the docs.

aclark4life: 4!

hugovk: Updated to 4! Let's do a squash merge on this PR to keep the Git history a little neater.

.editorconfig
```diff
@@ -13,6 +13,10 @@ indent_style = space
 
 trim_trailing_whitespace = true
 
+[*.rst]
+# Three-space indentation
```
radarhere, line 17:
```suggestion
# Four-space indentation
```

PR #6766, opened 2022-11-29
Do not convert I images when showing PNGs

### LINKED ISSUE #6765: Pillow destroys image by incorrectly binarizing grayscale image during show()
### What did you do?
I've probably discovered a severe bug in PIL. I'm trying to open an image but when opening and showing it with PIL, the image is incorrectly binarized and visually completely destroyed.

That the image opens without being destroyed.

### What actually happened?
The image gets destroyed:
0026_destroyed

* OS: MacOS
* Python: 3.9
* Pillow: 9.3.0

```python
from PIL import Image
Image.open("0026.png").show()
```

when reading the very same image with opencv, it works just fine:

```python
import cv2
img_cv2 = cv2.imread("0026.png")
Image.fromarray(img_cv2).show()
```

radarhere:
Resolves #6765

Similarly to #4744, to work around conversion issues, this PR does not convert I images when showing PNGs.

### DIFF
--- a/src/PIL/ImageShow.py
+++ b/src/PIL/ImageShow.py
@@ -77,7 +77,7 @@ class Viewer:
 
         if not (
             image.mode in ("1", "RGBA")
-            or (self.format == "PNG" and image.mode in ("I;16", "LA"))
+            or (self.format == "PNG" and image.mode in ("I", "I;16", "LA"))
         ):
             base = Image.getmodebase(image.mode)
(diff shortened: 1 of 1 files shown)


radarhere:
Closed as the issue creator didn't feel this really addressed the problem - https://github.com/python-pillow/Pillow/issues/6765#issuecomment-1331947235

PR #6767, opened 2022-11-30
When reading BLP, do not trust JPEG decoder to determine image is CMYK

### LINKED ISSUE #6741: test_file_blp.py fails with libjpeg 9e since 9.3.0
### What did you do?
I built Pillow 9.3.0 as downloaded from PyPI and ran the test suite with the following command:
`/usr/bin/python3.X -m pytest Tests/test_file_blp.py`

Every test passes.

### What actually happened?
`test_load_blp1` failed with the following error:
```python
    def test_load_blp1():
        with Image.open("Tests/images/blp/blp1_jpeg.blp") as im:
>           assert_image_equal_tofile(im, "Tests/images/blp/blp1_jpeg.png")

Tests/test_file_blp.py:15: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
Tests/helper.py:105: in assert_image_equal_tofile
    assert_image_equal(a, img, msg)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

a = <PIL.BlpImagePlugin.BlpImageFile image mode=RGB size=256x256 at 0x7FF952EF3AD10>
b = <PIL.PngImagePlugin.PngImageFile image mode=RGB size=256x256 at 0x7FF952EF3AF10>, msg = None

    def assert_image_equal(a, b, msg=None):
        assert a.mode == b.mode, msg or f"got mode {repr(a.mode)}, expected {repr(b.mode)}"
        assert a.size == b.size, msg or f"got size {repr(a.size)}, expected {repr(b.size)}"
        if a.tobytes() != b.tobytes():
            if HAS_UPLOADER:
                try:
                    url = test_image_results.upload(a, b)
                    logger.error(f"Url for test images: {url}")
                except Exception:
                    pass
    
>           assert False, msg or "got different content"
E           AssertionError: got different content
E           assert False
```

I don't see this failure in 9.2.0 - it seems that https://github.com/python-pillow/Pillow/commit/f7363c1091c70356d92e56abfca6b65bef9e7b26 caused this issue. When I revert it, the issue is gone (when I revert only the [BlpImagePlugin.py](https://github.com/python-pillow/Pillow/commit/f7363c1091c70356d92e56abfca6b65bef9e7b26#diff-30059d5b5f19c6de1f1b7fcb21e982958ef9432c2795ef68fb5470fb9eccb9dd) change, pillow segfaults on newly added [test_file_blp.py](https://github.com/python-pillow/Pillow/commit/f7363c1091c70356d92e56abfca6b65bef9e7b26#diff-ffd01a62e614987be01f44aeed0c978fc9576964b6e6ff9d9bbba177044f1a28) code, but it passes the `assert_image_equal_tofile` part this failure is about.

This happens on both SPARC and Intel meaning than endianness is most likely not the issue.

* OS: Oracle Solaris 11.4 (both on SPARC and x86)
* Python: 3.7.14, 3.9.14 and 3.11.0
* Pillow: 9.3.0

radarhere:
Resolves #6741

libjpeg 9e has decided to no longer detect the JPEG embedded within Tests/images/blp/blp1_jpeg.blp as being CMYK.

To allow our tests to still pass, if the BLP JPEG is in CMYK mode, this PR passes a value for [`jpegmode`](https://github.com/python-pillow/Pillow/blob/0ec32a30120cb22b533fd8563749d44bd5d3f78f/src/decode.c#L824-L833), so that [JpegDecode.c will override what libjpeg has detected.](https://github.com/python-pillow/Pillow/blob/0ec32a30120cb22b533fd8563749d44bd5d3f78f/src/libImaging/JpegDecode.c#L183-L195)

Testing in libjpeg 9d and 9e in my fork of pillow-wheels, the BLP tests [fail without this](https://github.com/radarhere/pillow-wheels/actions/runs/3579890161), and [pass with it.](https://github.com/radarhere/pillow-wheels/actions/runs/3579892207)

### DIFF
--- a/src/PIL/BlpImagePlugin.py
+++ b/src/PIL/BlpImagePlugin.py
@@ -373,6 +373,9 @@ class BLP1Decoder(_BLPBaseDecoder):
         data = BytesIO(data)
         image = JpegImageFile(data)
         Image._decompression_bomb_check(image.size)
+        if image.mode == "CMYK":
+            decoder_name, extents, offset, args = image.tile[0]
+            image.tile = [(decoder_name, extents, offset, (args[0], "CMYK"))]
         r, g, b = image.convert("RGB").split()
         image = Image.merge("RGB", (b, g, r))
         self.set_as_raw(image.tobytes())


kulikjak:
I tested the patch myself with each Python we deliver pillow for and the test suite is green again. Thanks!

Out of curiosity, IIUIC, the fix is to tell libjpeg explicitly that the given image is "CMYK" as it no longer detects it itself (and Pillow does know that through other means)?

PR #6774, opened 2022-12-02
Support saving JPEG comments

### LINKED ISSUE #6773: PIL doesn't write JPEG comment
Following up on an answer I gave to the following stackoverflow question: https://stackoverflow.com/a/74658310/1358308

OP asked how to write a JPEG comment into the file, i.e. not an EXIF comment, but as a COM segment.  The closest issue I could find, #4450, was about reading JPEG comments.

While looking for a solution I cloned the repo (hash 5bdf63e1d170e1db324c5f805c4e838e4d455e98) to see what was going on, but couldn't see anything relevant.

Would be happy to turn this into a pull-request if you think this is a reasonable idea.  I think I'd fix by modifying `_save()` to pull in the comment (from the `applist`) by appending to `extra` in a similar way as `icc_profile`:

https://github.com/python-pillow/Pillow/blob/948678633924fd62b4ab1cc632acd218ad6796ba/src/PIL/JpegImagePlugin.py#L714-L736

Comments appreciated!

smason:
Fixes #6773 

 * a JPEG COM segment can now be written during `Image.save`
 * opening a JPEG file and saving to another file will now copy any existing comment across
 * user can now do `image.save(filename, format='JPEG', comment='comment to be written')` and the comment will be written to a COM segment

### DIFF
--- a/src/encode.c
+++ b/src/encode.c
@@ -1057,7 +1057,7 @@ PyImaging_JpegEncoderNew(PyObject *self, PyObject *args) {
 
     if (!PyArg_ParseTuple(
             args,
-            "ss|nnnnnnnnOy#y#y#",
+            "ss|nnnnnnnnOz#y#y#",
             &mode,
             &rawmode,
             &quality,
--- a/src/libImaging/JpegEncode.c
+++ b/src/libImaging/JpegEncode.c
@@ -278,7 +278,7 @@ ImagingJpegEncode(Imaging im, ImagingCodecState state, UINT8 *buf, int bytes) {
 
         case 4:
 
-            if (context->comment_size > 0) {
+            if (context->comment) {
                 jpeg_write_marker(&context->cinfo, JPEG_COM, (unsigned char *)context->comment, context->comment_size);
             }
             state->state++;
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,9 @@ Changelog (Pillow)
 9.4.0 (unreleased)
 ------------------
 
+- Added getxmp() to WebPImagePlugin #6758
+  [radarhere]
+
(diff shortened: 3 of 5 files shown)


smason:
not sure what's going on with that MSYS2 MinGW 32-bit test failure, but it looks unrelated

radarhere:
There are some intermittent failures in our CIs. I've rerun the job, and it has passed now.

src/encode.c
```diff
@@ -1057,7 +1057,7 @@ PyImaging_JpegEncoderNew(PyObject *self, PyObject *args) {
 
     if (!PyArg_ParseTuple(
             args,
-            "ss|nnnnnnnnOy#y#y#",
+            "ss|nnnnnnnnOz#y#y#",
```
smason, line 1060: @radarhere what do you think about using a `#z` parameter?  advantages are:

* allows passing `None` directly
* no need to check for whether to `str.encode()`

src/libImaging/JpegEncode.c
```diff
@@ -278,7 +278,7 @@ ImagingJpegEncode(Imaging im, ImagingCodecState state, UINT8 *buf, int bytes) {
 
         case 4:
 
-            if (context->comment_size > 0) {
+            if (context->comment) {
```
smason, line 281: the rest of the code is conditional on `comment != NULL` rather than `comment_size > 0`

they should be the same, but keeping things consistent feels nicer to me

radarhere: When you say "the rest of the code", what are you referring to?

smason:
@radarhere thanks for all the help getting that merged!

sorry for the silence the past couple of days. I've been a bit ill, but recovering now

PR #6783, opened 2022-12-07
If available, use wl-paste or xclip for grabclipboard() on Linux

### LINKED ISSUE #6751: Support ImageGrab from clipboard on Linux via xclip and wl-clipboard
Currently, `grabclipboard` raises `NotImplementedError` on Linux. But I think we can use xclip or wl-clipboard if they are installed. A subprocess call saving the image to a temporary file should be enough.

```sh
xclip -selection clipboard -t image/jpeg -o > test.jpg  # Edit 1
wl-paste --type image/jpg > test.jpg
```

I'm willing to help implement this feature.

### What did you do?

Try to grab an image from the clipboard on Linux

The image is grabbed

### What actually happened?

```
NotImplementedError: ImageGrab.grabclipboard() is macOS and Windows only
```

* OS:  Linux (Manjaro KDE)
* Python: 3.10.8
* Pillow: 9.3.0

```python
from PIL.ImageGrab import grabclipboard
grabclipboard()
```

radarhere:
Resolves #6751, by using [`wl-paste`](https://github.com/bugaevc/wl-clipboard) or [`xclip`](https://github.com/astrand/xclip) for `ImageGrab.grabclipboard()` on Linux.

### DIFF
--- a/Tests/test_imagegrab.py
+++ b/Tests/test_imagegrab.py
@@ -64,9 +64,13 @@ $bmp = New-Object Drawing.Bitmap 200, 200
             )
             p.communicate()
         else:
-            with pytest.raises(NotImplementedError) as e:
-                ImageGrab.grabclipboard()
-            assert str(e.value) == "ImageGrab.grabclipboard() is macOS and Windows only"
+            if not shutil.which("wl-paste"):
+                with pytest.raises(NotImplementedError) as e:
+                    ImageGrab.grabclipboard()
+                assert (
+                    str(e.value)
+                    == "wl-paste is required for ImageGrab.grabclipboard() on Linux"
+                )
             return
 
         ImageGrab.grabclipboard()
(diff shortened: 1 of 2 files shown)


Tests/test_imagegrab.py
```diff
@@ -64,9 +64,13 @@ def test_grabclipboard(self):
...
-            with pytest.raises(NotImplementedError) as e:
-                ImageGrab.grabclipboard()
-            assert str(e.value) == "ImageGrab.grabclipboard() is macOS and Windows only"
+            if not shutil.which("wl-paste"):
+                with pytest.raises(NotImplementedError) as e:
+                    ImageGrab.grabclipboard()
+                assert (
+                    str(e.value) == "wl-paste or xclip is required"
+                    " for ImageGrab.grabclipboard() on Linux"
+                )
```
hugovk, line 73: Can we use `pytest.raises(ValueError, match='...')`?

```suggestion
                with pytest.raises(
                    NotImplementedError,
                    match="wl-paste or xclip is required "
                    "for ImageGrab.grabclipboard() on Linux",
                ) as e:
                    ImageGrab.grabclipboard()
```

https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.raises

radarhere: Sure, I've added a commit for that.

Issue #6786, opened 2022-12-07
[Documentation] Value ranges in the `Modes` section


GeorgeJIrwin:
There is missing information in the current `Modes` section (Version 9.3.0):

![image](https://user-images.githubusercontent.com/119657866/206249295-262cd629-2b2e-4f6f-8b8d-88bfd673b26a.png)

While it provides the value ranges expected for images built on `1` bit or `8` bits, it doesn't say the expected values for the types `I` and `F`.
What is the white value in those cases? It is important when using `fromarray` from a numpy array.
Usually the values for `Float32` images is in the range `[0, 1]` yet it doesn't seem to be the case with `PIL`.

Moreover, The values for the `YCbCr`, `LAB` and `HSV` are not specified. Usually `Y` is in the range `[0, 100]`, is it in that case or just scaling of `[0, 1]` into `{0, 1, ..., 255}`? What about the `H` in `HSV`?

It would be great if those will be mentioned explicitly in the documentation.

Yay295:
Also, I believe #3159 is again relevant here.

GeorgeJIrwin:
@radarhere , While indeed it says `8 Bit` will have the range 0-255, it doesn't say what is scaled there.
What's needed is the native values of the parameter and then how it is scaled into the 0-255 range.
For instance, for `H`, is 0-255 scaling of 0-359.99? Is is -180-179.99? Is it 0-360?

Anyhow, for `I` and `F` there is no way to understand what's the expected range.

radarhere:
I've created PR #6795 to document the Hue range.

"I" has the range of a "32-bit signed integer" - an INT32 in C.

```pycon
>>> from PIL import Image
>>> Image.new("I", (1, 1), -2**31).load()[0, 0]
-2147483648
>>> Image.new("I", (1, 1), 2**31-1).load()[0, 0]
2147483647
>>> Image.new("I", (1, 1), 2**31).load()[0, 0]
-2147483648
>>> Image.new("I", (1, 1), -2**31-1).load()[0, 0]
2147483647
```

![Screenshot 2022-12-10 at 7 57 34 pm](https://user-images.githubusercontent.com/3112309/206842229-5290d555-1415-4cce-a70a-c584f15dc17c.png)

For "F", it has the range of a "32-bit floating point pixel", a FLOAT32 in C.

radarhere:
I've updated #6795 to mention INT32 and FLOAT32.

PR #6795, opened 2022-12-10
Improved mode range documentation

radarhere:
Helps #6786

Inputting HSV (359.9, 100%, 100%) into https://www.rapidtables.com/convert/color/hsv-to-rgb.html, I get RGB (255, 0, 0).

```python
>>> from PIL import Image
>>> im = Image.new("HSV", (1, 1), (255, 255, 255))
>>> im.convert("RGB").getpixel((0, 0))
(255, 0, 0)
```

I would therefore conclude that the H value ranges from 0-360, excluding 360.


radarhere:
I've added another commit to mention the ranges of the "I" and "F" modes.

PR #6800, opened 2022-12-13
Fixed writing int as ASCII tag

### LINKED ISSUE #6799: AttributeError: 'int' object has no attribute 'encode'
This image:  rotateexif

Testcase:
 ````python
import PIL.Image
with PIL.Image.open('/tmp/rotateexif.jpg') as img:
    PIL.ImageOps.exif_transpose( img )
````

### PR DESCRIPTION
**Pull request by radarhere**
Resolves #6799

The image from the issue has a GPSLongitudeRef tag that [should be ASCII](https://www.awaresystems.be/imaging/tiff/tifftags/privateifd/gps/gpslongituderef.html), but is instead SHORT.

The tag is loaded as an int then, not a string as expected, and when Pillow tries to save it as a string, it fails.

This PR simply changes TiffImagePlugin to convert the variable to a string.

### DIFF
--- a/Tests/test_file_tiff_metadata.py
+++ b/Tests/test_file_tiff_metadata.py
@@ -185,20 +185,21 @@ def test_iptc(tmp_path):
         im.save(out)
 
 
-def test_writing_bytes_to_ascii(tmp_path):
+def test_writing_other_types_to_ascii(tmp_path):
     im = hopper()
     info = TiffImagePlugin.ImageFileDirectory_v2()
 
     tag = TiffTags.TAGS_V2[271]
     assert tag.type == TiffTags.ASCII
 
-    info[271] = b"test"
-
     out = str(tmp_path / "temp.tiff")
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: Tests/test_file_tiff_metadata.py [hunk 1]
```diff
@@ -185,20 +185,21 @@ def test_iptc(tmp_path):
...
     info = TiffImagePlugin.ImageFileDirectory_v2()
 
     tag = TiffTags.TAGS_V2[271]
     assert tag.type == TiffTags.ASCII
 
-    info[271] = b"test"
-
     out = str(tmp_path / "temp.tiff")
-    im.save(out, tiffinfo=info)
+    for (value, expected) in {b"test": "test", 1: "1"}.items():
```
hugovk, line 196: Move these to `@pytest.mark.parametrize`?

radarhere: Sure, done

PR #6820, opened 2022-12-23
Added DDS support for uncompressed L and LA images

radarhere:
#6486 refactors the DDS plugin, but also adds support for reading and writing uncompressed L and LA images.

This PR extracts that new functionality, without the refactoring.

In https://learn.microsoft.com/en-us/windows/win32/direct3ddds/dds-pixelformat, `DDPF_LUMINANCE` is described as
> Used in some older DDS files for single channel color uncompressed data

It
> Can be combined with DDPF_ALPHAPIXELS for a two channel DDS file.


hugovk (approved):
Thanks, let's also add to release notes.

radarhere:
Ok, I've added a commit for that.

PR #6830, opened 2022-12-27
Removed deprecations for Image constants, except for duplicate Resampling attributes

### LINKED ISSUE #6537: API changes for Resampling modes

radarhere:
Resolves #6537

The user in the issue found a warning that `Image.BILINEAR` is deprecated, and questions whether it is necessary to deprecate this.

This PR removes the deprecation warnings for `Image.*` constants added in #5954 ([and released in Pillow 9.1.0](https://pillow.readthedocs.io/en/stable/releasenotes/9.1.0.html#constants)).
However, there are three exceptions - I've left `LINEAR`, `CUBIC` (which have been duplicates of other constants since PIL) and `ANTIALIAS` (which has been a duplicate since #1028) deprecated.

I'm also leaving the deprecations for ImageCms, ImageFont, BlpImagePlugin, FtexImagePlugin and PngImagePlugin in place.

See what you think of this.

### DIFF
--- a/docs/releasenotes/9.4.0.rst
+++ b/docs/releasenotes/9.4.0.rst
@@ -103,3 +103,40 @@ Added support for DDS L and LA images
 
 Support has been added to read and write L and LA DDS images in the uncompressed
 format, known as "luminance" textures.
+
+.. _restored-image-constants:
+
+Constants
+^^^^^^^^^
+
+In Pillow 9.1.0, the following constants were deprecated. Those deprecations have now
+been restored.
+
+- ``Image.NONE``
+- ``Image.NEAREST``
+- ``Image.ORDERED``
+- ``Image.RASTERIZE``
+- ``Image.FLOYDSTEINBERG``
+- ``Image.WEB``
+- ``Image.ADAPTIVE``
+- ``Image.AFFINE``
+- ``Image.EXTENT``
+- ``Image.PERSPECTIVE``
+- ``Image.QUAD``
+- ``Image.MESH``
+- ``Image.FLIP_LEFT_RIGHT``
+- ``Image.FLIP_TOP_BOTTOM``
+- ``Image.ROTATE_90``
+- ``Image.ROTATE_180``
+- ``Image.ROTATE_270``
+- ``Image.TRANSPOSE``
+- ``Image.TRANSVERSE``
+- ``Image.BOX``
+- ``Image.BILINEAR``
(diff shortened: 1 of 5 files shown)


hugovk:
Thanks for picking this up. Looks fine, let's keep this open a few days and merge before January if there's no strong objections.

Let's also mention it in the release notes, and add a note (with `.. note::` ?) to https://pillow--6830.org.readthedocs.build/en/6830/deprecations.html#constants to mention that some had been deprecated but are no longer.

Also, I wonder if we should somehow mention it at https://pillow--6830.org.readthedocs.build/en/6830/releasenotes/9.1.0.html#constants? These are historical, but perhaps some sort of note, even if just a link to the 9.3.0 notes, could help?

radarhere:
Ok, I've added a commit to update the documentation.

docs/releasenotes/9.4.0.rst
```diff
@@ -103,3 +103,40 @@ Added support for DDS L and LA images
...
 Support has been added to read and write L and LA DDS images in the uncompressed
 format, known as "luminance" textures.
+
+.. _restored-image-constants:
+
+Constants
+^^^^^^^^^
+
+In Pillow 9.1.0, the following constants were deprecated. Those deprecations have now
+been restored.
```
hugovk, line 113: Wonder if something like this is clearer?
```suggestion
In Pillow 9.1.0, the following constants were deprecated.
That has been reversed and these constants will now remain available.
```

radarhere: Ok, sure. I've pushed a commit changing that, and the text in deprecations.rst.

hugovk (approved):

PR #6833, opened 2022-12-28
Updated isort to 5.11.4

radarhere:
isort 5.11.4 has been released - https://github.com/PyCQA/isort/releases

I've triggered #6604, but it didn't detect this for some reason.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,7 @@ repos:
         types: []
 
   - repo: https://github.com/PyCQA/isort
-    rev: 5.11.1
+    rev: 5.11.4
     hooks:
       - id: isort
 


hugovk:
Renovate doesn't handle `.pre-commit-config.yaml`, https://github.com/python-pillow/Pillow/issues/6604 shows its detected dependencies as github-actions and setup-cfg.

https://pre-commit.ci/ should do `.pre-commit-config.yaml` when it next triggers on 1st January:

https://github.com/python-pillow/Pillow/blob/2755e0ffaadc8b29c3e67e223c333c50e197a733/.pre-commit-config.yaml#L63-L64

radarhere:
Looking at [previous PRs](https://github.com/python-pillow/Pillow/pulls?q=is%3Apr+author%3Aapp%2Fpre-commit-ci), I think it will be between the 2nd to the 8th, rather than the 1st - but thanks for the explanation.

radarhere:
This became #6855

hugovk:
> Looking at [previous PRs](https://github.com/python-pillow/Pillow/pulls?q=is%3Apr+author%3Aapp%2Fpre-commit-ci), I think it will be between the 2nd to the 8th, rather than the 1st - but thanks for the explanation.

Looks like it's the first Monday of the month, ± timezones.

PR #6835, opened 2022-12-28
Use cron schedule to fix Renovate

hugovk:
In our [`renovate.json`](https://github.com/python-pillow/Pillow/blob/2f3561fe33df96cdb62d66f387009837ebf7254e/.github/renovate.json#L16) we have:

```json
    "schedule": ["on the 3rd day of the month"]
```

But Renovate PRs are being created immediately, for example:

* https://github.com/python-pillow/Pillow/pull/6776
* https://github.com/python-pillow/Pillow/pull/6806
* https://github.com/python-pillow/Pillow/pull/6813

I understand [Later.js](https://bunkat.github.io/later/) is used for scheduling, and if I run this code at https://npm.runkit.com/:

```js
var later = require('later');
var sched = later.parse.text('on the 3rd day of the month'),
      occurrences = later.schedule(sched).next(10);

for(var i = 0; i < 5; i++) {
  console.log(occurrences[i]);
}
```

I get the expected output for the third day of each month:

```
Tue Jan 03 2023 02:00:00 GMT+0200 (Eastern European Standard Time)
Fri Feb 03 2023 02:00:00 GMT+0200 (Eastern European Standard Time)
Fri Mar 03 2023 02:00:00 GMT+0200 (Eastern European Standard Time)
Mon Apr 03 2023 03:00:00 GMT+0300 (Eastern European Summer Time)
Wed May 03 2023 03:00:00 GMT+0300 (Eastern European Summer Time)
```

I asked about this at https://github.com/renovatebot/renovate/discussions/19513, one suggestion was to use cron scheduling instead. 

So let's try  [`0 12 3 * *`](https://crontab.guru/#0_12_3_*_*).

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -13,5 +13,5 @@
             "separateMajorMinor": "false"
         }
     ],
-    "schedule": ["on the 3rd day of the month"]
+    "schedule": ["0 12 3 * *"]
 }


radarhere:
#6604 has a feature, "Check this box to trigger a request for Renovate to run again on this repository".

I ran this to create #6813.
My memory is not good enough to know about the other two. You can look at the edit log on the first post of #6604 and draw your own conclusions.

So I'm not convinced this is a bug. My understanding is that the monthly schedule is to minimise spontaneous disruptions of the GitHub Actions queue. But if there is nothing running at a given time, I imagined it was ok to manually trigger it?

Alternatively, if you just feel that the cron format is clearer, then sure.

hugovk:
Ah, that'll explain it then :) Looks like two were from the checkbox and the other was actually on the 3rd day. So yes, looks like things are working correctly. Thanks!

"on the 3rd day of the month" is more readable than a cron format, so let's close this.

> My understanding is that the monthly schedule is to minimise spontaneous disruptions of the GitHub Actions queue. But if there is nothing running at a given time, I imagined it was ok to manually trigger it?

A monthly schedule can also help minimise noise: in most cases we don't need to update these very often, and a ~monthly schedule is good to stop things getting too old. And it can group them updates into a single PR, so we could leave it to the bot. (The 3rd was chosen to avoid release days and other repo schedules.)

But sure, it's fine to trigger more often if you like.

PR #6841, opened 2022-12-29
Fix version mismatch

smb123w64gb:
Small mix up.


radarhere:
Thanks. I missed this in #6711.

PR #6842, opened 2022-12-30
Return from ImagingFill early if image has a zero dimension

radarhere:
`ImagingAllocateArray` [returns early](https://github.com/python-pillow/Pillow/blob/8bd5fbf450f9f5a03a32c7efcfb488bfc2a40d1c/src/libImaging/Storage.c#L393-L396) when the image has zero width or height.

This PR applies that to `ImagingFill` as well.

### DIFF
--- a/.github/workflows/test-valgrind.yml
+++ b/.github/workflows/test-valgrind.yml
@@ -48,5 +48,5 @@ jobs:
       run: |
         # The Pillow user in the docker container is UID 1000
         sudo chown -R 1000 $GITHUB_WORKSPACE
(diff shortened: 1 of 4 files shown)


radarhere:
A timeout of 0.1s failed on [amazon-2-amd64](https://github.com/python-pillow/Pillow/actions/runs/3805053861/jobs/6472706491), so I've increased it to 0.5s.

hugovk (approved):

radarhere:
The new test started failing on valgrind, and I've even seen it fail once on ubuntu-18.04-bionic-amd64, so I've increased the timeout to 0.75s.

To fix intermittent valgrind problems with this new test, and the existing intermittent failures with `test_redos()` in test_file_pdf.py, I've pushed a change to skip these tests on valgrind.

PR #6846, opened 2022-12-31
Fixed null pointer dereference crash with malformed font

radarhere:


radarhere:
This would have only been present since Pillow 8.0.0, when `FT_Bitmap_Convert` was added by #4955.

radarhere:
This resolved https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=52906

Issue #6701, opened 2022-10-29
ImportError: DLL load failed on Pillow-9.3.0-cp37-cp37m-win_amd64.whl


cgohlke:
```pycon
> py -3.7 -m pip install Pillow
Collecting Pillow
  Downloading Pillow-9.3.0-cp37-cp37m-win_amd64.whl (2.5 MB)
     |████████████████████████████████| 2.5 MB 1.6 MB/s
Installing collected packages: Pillow
Successfully installed Pillow-9.3.0

> py -3.7
Python 3.7.9 (tags/v3.7.9:13c94747c7, Aug 17 2020, 18:58:18) [MSC v.1900 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> from PIL import _imagingft
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: DLL load failed: The specified module could not be found.
```

`_imagingft.cp37-win_amd64.pyd` depends on `VCRUNTIME140_1.dll`, which is not included with Python 3.7 and no longer included with Pillow wheels.

Solved by installing the [Microsoft Visual C++ Redistributable](https://learn.microsoft.com/en-US/cpp/windows/latest-supported-vc-redist?view=msvc-170).

nulano:
I checked `_imaging.cp37-win_amd64.pyd` did not have extra CRT dependencies, but missed `_imagingft.cp37-win_amd64.pyd`... 😕 

I've now checked all pyd files with `dumpbin /dependents ...\PIL"\*.pyd` and this is the only `VCRUNTIME140_1.dll` dependency.

Checking the lib files, the dependency seems to be coming from HarfBuzz due to a new default setting in Visual Studio.

I've disabled the setting in https://github.com/nulano/Pillow/commit/76b99756e425099b650bd383e204c259b5bdd3f1. Could you please test the wheels from this build, https://github.com/nulano/Pillow/actions/runs/3352301197 (I've attached the 3.7 wheel below)?

[Pillow-9.3.0-cp37-cp37m-win_amd64.whl.zip](https://github.com/python-pillow/Pillow/files/9894268/Pillow-9.3.0-cp37-cp37m-win_amd64.whl.zip)

nulano:
FWIW, I've just checked the wheel from my previous comment in Windows Sandbox and it seems to fix the issue.

hugovk:
[...]

PR #6847, opened 2023-01-01
Check available features in Windows wheels

nulano:
Add test for #6701 and similar issues in earlier releases.
See successful run in my fork: https://github.com/nulano/Pillow/actions/runs/3815551828
See failed run detecting the #6701 issue for 9.3.0 release: https://github.com/nulano/Pillow/actions/runs/3815551828

This runs the tests and checks available features in a Windows Server Core Docker container, similarly to how `run_tests` is run in the pillow-wheels repo.

Marked as draft because:
* [ ] PyPy requires a different download url.
* [ ] I'm not sure when this should be triggered. Running all tests twice for every push seems unnecessary (and takes extra 2min), but at least release wheels should be tested. I think the ideal option might be to run only when pushing a tag, but I think that might not get detected if there is a simultaneous push of the same commit to a branch.

### DIFF
--- a/Tests/test_imagegrab.py
+++ b/Tests/test_imagegrab.py
@@ -15,6 +15,8 @@ class TestImageGrab:
         sys.platform not in ("win32", "darwin"), reason="requires Windows or macOS"
     )
     def test_grab(self):
+        if os.environ.get("USERNAME", "") == "ContainerAdministrator":
+            pytest.skip("can't grab screen when running in Docker")
         ImageGrab.grab()
         ImageGrab.grab(include_layered_windows=True)
         ImageGrab.grab(all_screens=True)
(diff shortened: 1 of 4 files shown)


Tests/test_imagegrab.py
```diff
@@ -15,6 +15,8 @@ class TestImageGrab:
         sys.platform not in ("win32", "darwin"), reason="requires Windows or macOS"
     )
     def test_grab(self):
+        if os.environ.get("USERNAME", "") == "ContainerAdministrator":
```
radarhere, line 18:
```suggestion
        if os.environ.get("USERNAME") == "ContainerAdministrator":
```

radarhere:
@nulano you may wish to consider this again as part of https://github.com/python-pillow/Pillow/issues/7390#issuecomment-1813365035

nulano:
Yes, I was planning to rebase this after writing that PR.

PR #6853, opened 2023-01-02
Relaxed child images check to allow for libjpeg

radarhere:
Reported in https://github.com/conda-forge/pillow-feedstock/pull/128#issuecomment-1368705067

`test_get_child_images()` in test_file_jpeg.py is failing with libjpeg. This PR relaxes the check to test for a similarity of 2.1, instead of equality. I have confirmed this in pillow-wheels with libjpeg 9e.

### DIFF
--- a/Tests/test_file_jpeg.py
+++ b/Tests/test_file_jpeg.py
@@ -447,7 +447,7 @@ class TestFileJpeg:
             ims = im.get_child_images()
 
         assert len(ims) == 1
-        assert_image_equal_tofile(ims[0], "Tests/images/flower_thumbnail.png")
+        assert_image_similar_tofile(ims[0], "Tests/images/flower_thumbnail.png", 2.1)
 
     def test_mp(self):
         with Image.open("Tests/images/pil_sample_rgb.jpg") as im:


h-vetinari:
Sorry, this was a false alarm!

I got caught out by the fact that `test_strip_ycbcr_jpeg_2x2_sampling` started re-failing for 9.4 when I switched to libjpeg-turbo, and actually `eps=1.2` is fine for `test_get_child_images`!

Sorry for the mix-up!

Perhaps it would be possible to widen the skip of `test_strip_ycbcr_jpeg_2x2_sampling` to take into account newer `libjpeg_turbo`? Currently it only skips 2.0, but we're on 2.1 (and ABI-wise, we're only pinning to 2, without the minor version).

Tests/test_file_jpeg.py
```diff
@@ -447,7 +447,7 @@ def test_get_child_images(self):
             ims = im.get_child_images()
 
         assert len(ims) == 1
-        assert_image_equal_tofile(ims[0], "Tests/images/flower_thumbnail.png")
+        assert_image_similar_tofile(ims[0], "Tests/images/flower_thumbnail.png", 2.1)
```
h-vetinari, line 450:
```suggestion
        assert_image_similar_tofile(ims[0], "Tests/images/flower_thumbnail.png", 1.2)
```

h-vetinari (commented):
In other words, this PR is still necessary (or at least useful for us), but should be good enough with the following:

radarhere:
When I test with libjpeg 9e, I also found a difference of 2.0033 - https://github.com/radarhere/pillow-wheels/actions/runs/3821343292/jobs/6500396156

If we're going to change this to improve support, might as well change it to support more libjpeg environments in general.

radarhere:
> Perhaps it would be possible to widen the skip of test_strip_ycbcr_jpeg_2x2_sampling to take into account newer libjpeg_turbo? Currently it only skips 2.0, but we're on 2.1 (and ABI-wise, we're only pinning to 2, without the minor version).

Our GitHub Actions uses libjpeg turbo 2.1 and passes - https://github.com/python-pillow/Pillow/actions/runs/3819974382/jobs/6497949271#step:8:39

If you'd like that to be investigated further, please open a new issue.

h-vetinari:
Yeah, I should not have responded so quickly (but wanted to avoid an inadvertent merge in that moment).

The 2.1 _remains_ necessary on osx, where `test_strip_ycbcr_jpeg_2x2_sampling` does _not_ fail on osx, and so I _had_ actually read the right results.

h-vetinari:
> Our GitHub Actions uses libjpeg turbo 2.1 and passes - https://github.com/python-pillow/Pillow/actions/runs/3819974382/jobs/6497949271#step:8:39
> 
> If you'd like that to be investigated further, please open a new issue.

Fair enough. Since it's just a minor accuracy violation I'm going to skip that test for our builds.

PR #2862, opened 2017-11-17
RFC: Display differences

wiredfool:
I've been thinking about something like this since we had the issue that was non-reproducable on any of my machines. 

What this does is push all test failures for `assert_image_equal` and `assert_image_similar` to a webservice, where it displays those images and the difference. 

```
======================================================================

FAIL: TestUploader.test_upload_equal

----------------------------------------------------------------------

Traceback (most recent call last):

  File "/home/travis/build/wiredfool/Pillow/Tests/test_uploader.py", line 10, in test_upload_equal

    self.assert_image_equal(result, target)

  File "/home/travis/build/wiredfool/Pillow/Tests/helper.py", line 104, in assert_image_equal

    self.fail(msg or "got different content")

AssertionError: got different content

-------------------- >> begin captured logging << --------------------

urllib3.connectionpool: DEBUG: Starting new HTTPS connection (1): fmuvehxed0.execute-api.us-east-2.amazonaws.com

urllib3.connectionpool: DEBUG: https://fmuvehxed0.execute-api.us-east-2.amazonaws.com:443 "GET /prod/upload HTTP/1.1" 200 3421

urllib3.connectionpool: DEBUG: Starting new HTTPS connection (1): s3.us-east-2.amazonaws.com

urllib3.connectionpool: DEBUG: https://s3.us-east-2.amazonaws.com:443 "POST /pillow-test-image-results HTTP/1.1" 204 0

urllib3.connectionpool: DEBUG: Starting new HTTPS connection (1): s3.us-east-2.amazonaws.com

urllib3.connectionpool: DEBUG: https://s3.us-east-2.amazonaws.com:443 "POST /pillow-test-image-results HTTP/1.1" 204 0

helper: ERROR: Url for test images: https://fmuvehxed0.execute-api.us-east-2.amazonaws.com/prod/show/1510948233.21/

--------------------- >> end captured logging << ---------------------
```

I'm not 100% on the output yet, and we're still not going to trigger on some forms of errors. 

Comments?

PR #6857, opened 2023-01-03
Fix function names for pytest

Yay295:
These tests are not currently being run because the test names don't start with "test".

### DIFF
--- a/Tests/test_uploader.py
+++ b/Tests/test_uploader.py
@@ -1,13 +1,13 @@
 from .helper import assert_image_equal, assert_image_similar, hopper
 
 
-def check_upload_equal():
+def test_upload_equal():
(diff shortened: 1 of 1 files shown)


Yay295:
and it looks like those tests don't pass anymore either

radarhere:
This was actually done deliberately in https://github.com/python-pillow/Pillow/pull/2862/commits/9a7fbdf32306bfe83094ffdf4d46c2649b4dfb8a

It looks like they are there not to test Pillow, but to trigger the [uploader](https://github.com/python-pillow/Pillow/blob/f5c47ebe128a448ada4e0330005e580e1ea0ab95/Tests/helper.py#L87-L98), if you would like to do so.

Yay295:
Thanks for the info.

PR #6865, opened 2023-01-06
Clean up tests (simplified)

Yay295:
This is commits 2-6 from #6653.


Tests/test_image_resize.py
```diff
@@ -57,13 +57,13 @@ def test_convolution_modes(self):
             Image.Resampling.LANCZOS,
         ),
     )
-    def test_reduce_filters(self, resample):
-        r = self.resize(hopper("RGB"), (15, 12), resample)
```
radarhere, line 61: Why is 'resampling_filter' clearer? 'resample' is the name of the variable when the core resize operation is called in Image.py
https://github.com/python-pillow/Pillow/blob/86634b835257a7d3438eaf783d70eca12e20b13e/src/PIL/Image.py#L2192

radarhere (approved):

Tests/test_image_resize.py
radarhere: Since your clear intention to just get part of #6653 merged, I'll save this change for that PR, and merge the rest.

Yay295: Because "resample" is an action, while "resampling filter" is an object. The thing being passed is (an ID for) a resampling filter, not a "resample". It doesn't work to say that "resample" is the type of the thing being passed either, because the type of the thing being passed is a [`Resampling` enum item](https://pillow.readthedocs.io/en/stable/reference/Image.html#resampling-filters).

PR #5365, opened 2021-03-29
Document Raqm/FriBiDi linking changes

nulano:
Release notes & documentation for #5062.

I've also noticed that `@cgohlke`'s website includes `fribidi-0.dll`, so this PR adds support for that filename to avoid hassle for users when upgrading.

Also correct information about xcb being included on macos.

PR #6876, opened 2023-01-09
List Raqm build options in installation documentation

nulano:
I noticed that #5365 omitted documentation for the build flags `--disable-raqm` and `--enable-raqm`.
This PR adds them to the documentation.

The order is the same as in `setup.py`.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -369,21 +369,21 @@ Build Options
   available, as many as are present.
 
 * Build flags: ``--disable-zlib``, ``--disable-jpeg``,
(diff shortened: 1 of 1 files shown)


nulano:
It seems one of the Cygwin jobs was cancelled due to timing out after 40 minutes, mainly due to installing numpy: https://github.com/python-pillow/Pillow/actions/runs/3877239287/jobs/6612023132

radarhere:
I reran the job and it has passed now.

radarhere (approved):

Issue #6877, opened 2023-01-09
Flaky test: Tests/test_file_eps.py::test_timeout


hugovk:
`test_timeout[Tests/images/timeout-d675703545fee17acab56e5fec644c19979175de.eps]` in `test_file_eps.py` is a flaky test that intermittently fails. It often passes when restarted.

It often fails on the ubuntu-22.04-jammy-ppc64le and ubuntu-22.04-jammy-s390x Docker images jobs, which are pretty slow anyway - they take between 15-25 mins, so restarting can mean a long additional wait.

It would be good to fix this: fix/rewrite the test, or the code if the problem is there, or maybe even remove it. There are also pytest plugins to re-run flaky tests a number of times.

Perhaps this test should be skipped for these particular flavours.

# The failure

```python
=================================== FAILURES ===================================
_ test_timeout[Tests/images/timeout-d675703545fee17acab56e5fec644c19979175de.eps] _

test_file = 'Tests/images/timeout-d675703545fee17acab56e5fec644c19979175de.eps'

    @pytest.mark.timeout(timeout=5)
    @pytest.mark.parametrize(
        "test_file",
        ["Tests/images/timeout-d675703545fee17acab56e5fec644c19979175de.eps"],
    )
    def test_timeout(test_file):
        with open(test_file, "rb") as f:
            with pytest.raises(Image.UnidentifiedImageError):
>               with Image.open(f):

Tests/test_file_eps.py:294: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/vpy3/lib/python3.10/site-packages/PIL/Image.py:3268: in open
    im = _open_core(fp, filename, prefix, formats)
/vpy3/lib/python3.10/site-packages/PIL/Image.py:3254: in _open_core
    im = factory(fp, filename)
/vpy3/lib/python3.10/site-packages/PIL/ImageFile.py:117: in __init__
    self._open()
/vpy3/lib/python3.10/site-packages/PIL/EpsImagePlugin.py:227: in _open
    s_raw = fp.readline()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <PIL.EpsImagePlugin.PSFile object at 0x40086451e0>

    def readline(self):
[...]

PR #6879, opened 2023-01-10
Refactor EpsImagePlugin and deprecate PSFile

Yay295:
Merge the PSFile class into the EpsImageFile class to hopefully improve performance.
Also added a check for the required "%!PS-Adobe" and "%%BoundingBox" header comments.

This hopefully helps with #6877.


hugovk:
```diff
 class PSFile:
     """
     Wrapper for bytesio object that treats either CR or LF as end of line.
+    This class is no longer used internally, but kept for backwards-compatibility.
     """
```

If we're no longer using it internally, do we need to keep it around in the API?

Can it be deprecated? Is there a replacement?

Yay295:
"`PSFile` is part of the public API, so we'd need to raise deprecation warnings for at least a year first before it can be removed in a major bump, which would be Pillow 11.0.0 in October 2024." - hugovk https://github.com/python-pillow/Pillow/issues/6877#issuecomment-1377200304

I don't know how to add a deprecation warning though; I haven't done that before.

hugovk:
Yup.

To deprecate something:

* Raise a `DeprecationWarning` using https://pillow.readthedocs.io/en/stable/reference/internal_modules.html#PIL._deprecate.deprecate with `when=11` (depends on https://github.com/python-pillow/Pillow/pull/6905)

* Update tests so we're not getting lots of deprecation warnings, add add a new test that explicitly checks for a warning with `with pytest.warns(DeprecationWarning)`.

* Document it at https://pillow.readthedocs.io/en/stable/deprecations.html and in release notes, ideally telling the user what they need to do if they're using it.

---

[Searching the top 5k](https://dev.to/hugovk/how-to-search-5000-python-projects-31gk) projects on PyPI shows no real use of `PSFile`:

```console
$ python3 ~/github/misc/cpython/search_pypi_top.py . "\bPSFile\b" -q
./pyre-check-0.9.16.tar.gz: pyre-check-0.9.16/typeshed/stubs/Pillow/PIL/EpsImagePlugin.pyi: class PSFile:
./Pillow-9.3.0.tar.gz: Pillow-9.3.0/Tests/test_file_eps.py: t = EpsImagePlugin.PSFile(f)
./Pillow-9.3.0.tar.gz: Pillow-9.3.0/Tests/test_file_eps.py: t = EpsImagePlugin.PSFile(r)
./Pillow-9.3.0.tar.gz: Pillow-9.3.0/src/PIL/EpsImagePlugin.py: class PSFile:
./Pillow-9.3.0.tar.gz: Pillow-9.3.0/src/PIL/EpsImagePlugin.py: fp = PSFile(self.fp)
./pytype-2022.10.26.tar.gz: pytype-2022.10.26/pytype/typeshed/stubs/Pillow/PIL/EpsImagePlugin.pyi: class PSFile:
./types-Pillow-9.3.0.0.tar.gz: types-Pillow-9.3.0.0/PIL-stubs/EpsImagePlugin.pyi: class PSFile:

Time: 0:00:23.017025
Found 7 matching lines in 4 projects
```

And I didn't spot any relevant use in the [940 GitHub results](https://github.com/search?q=%2F%5CbPSFile%5Cb%2F+language%3APython&type=code&l=Python&p=1).

---

But before we go ahead, @radarhere what are your thoughts?

radarhere:
I don't mind deprecating `PSFile`. If it isn't used, then removing it makes things simpler.

You've also made various other changes here, such as making use of `memoryview`, which isn't used anywhere else in Pillow at the moment. Could you explain them a bit?

Yay295:
This plugin reads lines byte-by-byte until it reaches the end of the line (or 255 bytes). `PSFile` does this by appending each byte to a list, joining the list, and decoding the bytes to a string.
https://github.com/python-pillow/Pillow/blob/0ad275f85f3bef71c82802e38a6001c4db33512b/src/PIL/EpsImagePlugin.py#L175-L191
My version instead reads the bytes into a fixed `bytearray`. This can be done because we know we can't have a valid line longer than 255 bytes, so we can allocate the space for that at the start and just reuse it for each line. A [`memoryview`](https://docs.python.org/dev/library/stdtypes.html#memoryview) is just a wrapper around this byte array that lets us use subsections of it without having to create a copy of that subsection.

Tests/test_file_eps.py
```diff
@@ -220,7 +311,8 @@ def test_read_binary_preview():
         pass
 
 
-def test_readline(tmp_path):
+@pytest.mark.filterwarnings("ignore::DeprecationWarning")
```
hugovk, line 314:
```suggestion
```

Rather than filtering out all `DeprecationWarning`s (there could potentially be warnings in the future from elsewhere that we would want to know about), we can put a `with pytest.warns(DeprecationWarning):` context manager around the specific `PSFile` call.

src/PIL/EpsImagePlugin.py
```diff
@@ -162,9 +163,15 @@ def Ghostscript(tile, size, fp, scale=1, transparency=False):
...
     """
     Wrapper for bytesio object that treats either CR or LF as end of line.
+    This class is no longer used internally, but kept for backwards-compatibility.
     """
 
     def __init__(self, fp):
+        deprecate(
+            "PSFile",
+            11,
+            action="If you need the functionality of this class you will need to implement it yourself.",
```
hugovk, line 173: Please split this line to pass linting

src/PIL/EpsImagePlugin.py
```diff
@@ -162,9 +163,15 @@ def Ghostscript(tile, size, fp, scale=1, transparency=False):
 class PSFile:
     """
     Wrapper for bytesio object that treats either CR or LF as end of line.
+    This class is no longer used internally, but kept for backwards-compatibility.
```
hugovk, line 166:
```suggestion
    This class is no longer used internally, but kept for backwards compatibility.
```

docs/deprecations.rst
```diff
@@ -80,6 +80,16 @@ A number of constants have been deprecated and will be removed in Pillow 10.0.0
     was reversed in Pillow 9.4.0 and those constants will now remain available.
     See :ref:`restored-image-constants`
 
+PSFile
+~~~~~~
+
+.. deprecated:: 9.4.0
```
hugovk, line 86: The next release is 9.5.0:

```suggestion
.. deprecated:: 9.5.0
```

Also the deprecations are listed in the order of the deprecation version, so let's move this bit down to the end. 

Similarly, move the info from `docs/releasenotes/9.4.0.rst` to `docs/releasenotes/9.5.0.rst`

src/PIL/EpsImagePlugin.py
```diff
@@ -275,43 +317,44 @@ def _open(self):
...
-            s = fp.readline().strip("\r\n")
-            if not s:
-                break
+            bytes_read = 0
+
+        check_required_header_comments()
 
-        if not box:
+        if not self._size:
+            self._size = 1, 1  # errors if this isn't set. why (1,1)?
```
radarhere, line 357: Could you tell me what the "errors" are? I don't see any coming from our test suite - https://github.com/radarhere/Pillow/actions/runs/4582259664

radarhere: I've created PR #7056 to remove this line.

Yay295: I checked out the first commit in this branch, removed that line, and ran the tests, and there weren't any errors. So I'm not sure what error I had seen back then.

PR #6883, opened 2023-01-11
Updated xz to 5.4.1

radarhere:
xz 5.4.1 has been released - https://tukaani.org/xz/

### DIFF
--- a/winbuild/build_prepare.py
+++ b/winbuild/build_prepare.py
@@ -152,9 +152,9 @@ deps = {
         "libs": [r"*.lib"],
     },
     "xz": {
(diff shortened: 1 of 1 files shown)


nulano:
Just noting for the future:

From https://git.tukaani.org/?p=xz.git;a=blob;f=NEWS;hb=HEAD:

```
XZ Utils Release Notes
======================

5.4.1 (2023-01-11)

<snip> 

- Updated windows/INSTALL-MSVC.txt to mention that CMake-based
  build is now the preferred method with Visual Studio. The
  project files will probably be removed after 5.4.x releases.
```

Issue #6875, opened 2023-01-09
Flaky test: test_qt_image_qapplication.py::test_sanity


hugovk:
`test_sanity` in `test_qt_image_qapplication.py` is a flaky test that intermittently fails. It often passes when restarted.

It would be good to fix this: fix/rewrite the test, or the code if the problem is there, or maybe even remove it. There are also pytest plugins to re-run flaky tests a number of times.

# The failure

```python
Tests/test_qt_image_qapplication.py::test_sanity FAILED                  [ 98%]

================================== FAILURES ===================================
_________________________________ test_sanity _________________________________

tmp_path = WindowsPath('C:/Users/runneradmin/AppData/Local/Temp/pytest-of-runneradmin/pytest-0/test_sanity7')

    @pytest.mark.skipif(not ImageQt.qt_is_installed, reason="Qt bindings are not installed")
    def test_sanity(tmp_path):
        # Segfault test
        app = QApplication([])
        ex = Example()
        assert app  # Silence warning
        assert ex  # Silence warning
    
        for mode in ("1", "RGB", "RGBA", "L", "P"):
            # to QPixmap
            im = hopper(mode)
            data = ImageQt.toqpixmap(im)
    
            assert isinstance(data, QPixmap)
            assert not data.isNull()
    
            # Test saving the file
            tempfile = str(tmp_path / f"temp_{mode}.png")
            data.save(tempfile)
    
            # Render the image
            qimage = ImageQt.ImageQt(im)
            data = QPixmap.fromImage(qimage)
            qt_format = QImage.Format if ImageQt.qt_version == "6" else QImage
            qimage = QImage(128, 128, qt_format.Format_ARGB32)
            painter = QPainter(qimage)
            image_label = QLabel()
            image_label.setPixmap(data)
            image_label.render(painter, QPoint(0, 0), QRegion(0, 0, 128, 128))
            painter.end()
            rendered_tempfile = str(tmp_path / f"temp_rendered_{mode}.png")
            qimage.save(rendered_tempfile)
            assert_image_equal_tofile(im.convert("RGBA"), rendered_tempfile)
    
            # from QPixmap
>           roundtrip(hopper(mode))

Tests/test_qt_image_qapplication.py:89: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
[...]

PR #6888, opened 2023-01-13
GHA: Test MSYS2/CLANG64

nulano:
MSYS2 now has more environments: https://www.msys2.org/docs/environments/

 * Test CLANG64 ~~instead of MINGW32~~, https://github.com/python-pillow/Pillow/pull/6888#issuecomment-1400392821
 * ~~Add xfail mark for #6875 as it failed 9 times in a row~~
 * Update installation documentation
 * While investigating https://github.com/python-pillow/Pillow/pull/6944, I found that replacing the commands used in the workflow with the `.ci` scripts simplifies it a bit.


nulano:
Ubuntu is failing because of https://github.com/actions/runner-images/issues/6677

radarhere:
I've rerun the tests, and they've passed now.

hugovk:
Thank you for this PR!

What's the benefit in adding all of these environments?

Our build matrix is pretty big right now, so let's be careful not to make it too much bigger.

This increases the test jobs in this worfklow from 2 to 5, and total duration from ~12 to ~21 mins:

<table>
<tr>
	<td>image
	<td>image

</table>

That's on top of 61 checks at the moment (mostly but not all GHA), taking about 184 mins of GHA time (wall-clock time maybe about an hour?).

For example, we aren't testing the GHA macOS/Ubuntu jobs with [both architectures](https://github.com/actions/setup-python#supported-architectures), and we will need to add some 3.12-dev coverage soonish.

Not opposed to this PR, and I'm very much for increasing test coverage, but let's keep an eye on things so they stay manageable :)

.github/workflows/test-msys2.yml
```diff
@@ -62,7 +65,7 @@ jobs:
               ${{ matrix.package }}-openjpeg2 \
```
kmilos, line 65: It is much nice to use the [pacboy `:p` way of installing packages](https://github.com/msys2/setup-msys2#pacboy), no need to worry about this "package" prefix...

.github/workflows/test-msys2.yml
```diff
@@ -62,7 +65,7 @@ jobs:
...
               subversion
 
-          if [ ${{ matrix.package }} == "mingw-w64-x86_64" ]; then
+          if [[ ${{ matrix.package }} =~ "x86_64" ]]; then
```
kmilos, line 68: use msystem instead once package is removed

docs/installation.rst
```diff
@@ -273,11 +273,27 @@ Many of Pillow's features require external libraries:
 
 .. tab:: Windows using MSYS2/MinGW
 
-    To build Pillow using MSYS2, make sure you run the **MSYS2 MinGW 32-bit** or
-    **MSYS2 MinGW 64-bit** console, *not* **MSYS2** directly.
-
-    The following instructions target the 64-bit build, for 32-bit
-    replace all occurrences of ``mingw-w64-x86_64-`` with ``mingw-w64-i686-``.
+    The following instructions are for the **MSYS2 MINGW64** system.
```
kmilos, line 276: [MINGW64 is no longer recommended.](https://www.msys2.org/news/#2022-10-29-changing-the-default-environment-from-mingw64-to-ucrt64) Please update everything accordingly.

kmilos:
IMHO testing MINGW64 (GCC toolchain w/ legacy MSVCRT) and CLANG64 (LLVM toolchain w/ modern UCRT) should provide decent coverage if one is trying to reduce the number of jobs. There's little point in testing 32-bit, support for it is being phased out anyway...

nulano:
I've had a hardware issue preventing me from working on this until now...

> What's the benefit in adding all of these environments?

I was going to suggest something similar to @kmilos's comment.

I've now updated the test matrix to (and listed the others in a comment):
* MINGW64
* CLANG64 instead of MINGW32

kmilos:
Any particular reason for not using the setup-msys2 action?

nulano:
> Any particular reason for not using the setup-msys2 action and all this extra manual work?

IIRC this was made before the setup-msys2 action, and it seems like less work to just update it as is instead of switching to the setup-msys2 action.

kmilos:
I think you'll find switching will pay off when it comes to flexibility and long term maintanance of these jobs.

kmilos:
Btw, should probably just [install](https://github.com/python-pillow/Pillow/actions/runs/18557744187/workflow?pr=6888#L84) available MSYS2 python-setuptools and pybind11 packages for the [test workflow](https://github.com/python-pillow/Pillow/actions/runs/18557744187/job/52899200593?pr=6888#step:5:47)... (Both could and probably do carry platform specific patches.)

.github/workflows/test-msys2.yml
```diff
@@ -0,0 +1,115 @@
...
+    env:
+      MSYSTEM: ${{ matrix.msystem }}
+      CHERE_INVOKING: 1
+
+    timeout-minutes: 30
+    name: MSYS2 ${{ matrix.msystem }}
+
+    steps:
+      - name: Checkout Pillow
+        uses: actions/checkout@v4
```
radarhere:
```suggestion
        uses: actions/checkout@v6
```

PR #6903, opened 2023-01-19
Support custom comments and PLT markers when saving JPEG2000 images

joshware:

Adds two new options when saving JPEG2000 images.

* The `comment` parameter to the save function replaces the default "Created by OpenJPEG version" with a given string.
* The `add_plt` parameter includes a PLT marker segment codestream, which is required for some ancient image viewers.


docs/handbook/image-file-formats.rst
```diff
@@ -589,6 +589,19 @@ The :py:meth:`~PIL.Image.Image.save` method supports the following options:
...
     .. versionadded:: 9.1.0
 
+**comment**
+    Adds a custom comment to the file, replacing the default 
+    "Created by OpenJPEG version" comment.
+
+    .. versionadded:: 9.5.0
+
+**add_plt**
+    If ``True`` then include a PLT (packet length, tile-part header) marker
```
radarhere, line 599:
```suggestion
    Adds a custom comment to the file, replacing the default
    "Created by OpenJPEG version" comment.

    .. versionadded:: 9.5.0

**add_plt**
    If ``True`` then include a PLT (packet length, tile-part header) marker
```
Lint fixes.

radarhere:
> PLT marker segment codestream, which is required for some ancient image viewers.

Could you elaborate on this? What are some example of the 'ancient image viewers'?

Also, it's required for old image viewers, but the option was [only added to OpenJPEG](https://github.com/python-pillow/Pillow/pull/6903/files#diff-b9b4700e3990596f51d9c477d712c4d73a2fa4fcb20d14e93af2f80d34d6e4aaR499) two years ago?

Tests/test_file_jpeg2k.py
```diff
@@ -371,3 +372,46 @@ def test_crashes(test_file):
...
+    assert opj_version is not None
+
+    if float(opj_version[1]) >= 2.4:
+        out = BytesIO()
+        test_card.save(out, "JPEG2000", no_jp2=True, add_plt=True)
+        out.seek(0)
+        while True:
+            box_bytes = out.read(2)
+            if len(box_bytes) == 0:
+                # End of steam encounterd and no PLT or SOD
```
radarhere, line 399:
```suggestion
                # End of steam encountered and no PLT or SOD
```

joshware:
Well, it's technically one image viewer I'm referring to, it's called 'RICS' and it's used in Pathology and It's built on the Kadaku library.
It dates back to around 2003, and only runs on Windows 2003/XP.

So I've been needing to integrate with/replace it with something more modern.

src/PIL/Jpeg2KImagePlugin.py
```diff
@@ -328,6 +328,8 @@ def _save(im, fp, filename):
     mct = info.get("mct", 0)
     signed = info.get("signed", False)
     fd = -1
+    comment = info.get("comment", None)
```
radarhere, line 331:
```suggestion
    comment = info.get("comment")
```
Just because `None` is the default.

Tests/test_file_jpeg2k.py
```diff
@@ -371,3 +372,46 @@ def test_crashes(test_file):
...
+
+
+def test_custom_comment():
+    output_stream = BytesIO()
+    unique_comment = "This is a unique comment, which should be found below"
+    test_card.save(output_stream, "JPEG2000", comment=unique_comment)
+    output_stream.seek(0)
+    data = output_stream.read()
+    # Lazy method to determine if the comment is in the image generated
+    assert bytes(unique_comment, "utf-8") in data
```
radarhere, line 384: Ideally, we could test this by reading the comment from the image as well. So I've created PR #6909

radarhere: #6909 has been merged now, so I've created https://github.com/joshware/Pillow/pull/2

radarhere:
I didn't think that "add_" was necessary, so I've renamed "add_plt" to just "plt".

docs/handbook/image-file-formats.rst
```diff
@@ -589,6 +589,19 @@ The :py:meth:`~PIL.Image.Image.save` method supports the following options:
 
     .. versionadded:: 9.1.0
 
+**comment**
+    Adds a custom comment to the file, replacing the default
+    "Created by OpenJPEG version" comment.
+
+    .. versionadded:: 9.5.0
+
+**add_plt**
```
hugovk, line 598:
```suggestion
**plt**
```

radarhere (approved):

PR #6915, opened 2023-01-28
Relax ImageQt roundtrip check

radarhere:
Helps #6875

I ran the test suite repeatedly in GitHub Actions, and managed to have the [`roundtrip()`](https://github.com/python-pillow/Pillow/blob/698951e19e19972aeed56df686868f1329981c12/Tests/test_qt_image_qapplication.py#L48-L51) check fail during the RGB iteration of loop. Here is what the image looks like when it fails.

result

There are three short lines of incorrect pixels in the image. If that is acceptable, then this PR relaxes the check so that this image would have passed.

### DIFF
--- a/Tests/test_qt_image_qapplication.py
+++ b/Tests/test_qt_image_qapplication.py
@@ -6,7 +6,7 @@ with warnings.catch_warnings():
     warnings.simplefilter("ignore", category=DeprecationWarning)
     from PIL import ImageQt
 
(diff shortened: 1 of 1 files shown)


hugovk:
Thanks!

I think this is acceptable to ease the CI pain. The problem still exists (and we have a better idea about it now!), but it seems to be limited to the interaction with Qt and I guess we've not had any complaints so far! And it's intermittent.

PR #6918, opened 2023-01-28
Replace SVN with Git for installing extra test images

hugovk:
GitHub is removing support for Subversion on January 8, 2024.

* https://github.blog/2023-01-20-sunsetting-subversion-support/

Why do we care?

Well, we're using `svn` in our CI to extract only the `test_images` directory from https://github.com/python-pillow/pillow-depends.

https://github.com/python-pillow/pillow-depends is a big repo full of binary files.

For our purposes, I think we're better off moving the test images into their own repo, so we can instead simply download the new repo as a zipball.

I've made a new repo at https://github.com/hugovk/test-images, which is just the relevant directory from https://github.com/python-pillow/pillow-depends, and preserves the Git history (following [these instructions](https://stackoverflow.com/a/68536044/724176)).

Let's move that repo into https://github.com/python-pillow. Because I don't have permission to create repos in the org, we need two steps:

@aclark4life I've sent a request to transfer the repo to you. Once you accept, please could you transfer it to https://github.com/python-pillow, and check the usual suspects have admin rights?

* Transfer docs here: https://docs.github.com/en/repositories/creating-and-managing-repositories/transferring-a-repository

Once it's transferred, I'll update the paths in this PR.

# Checklist

* [x] Transfer repo https://github.com/hugovk/test-images to https://github.com/python-pillow org
* [x] Update paths in this PR (the `TODO`s)
* [ ] Merge this PR
* [ ] Remove `test_images` directory from https://github.com/python-pillow/pillow-depends: https://github.com/python-pillow/pillow-depends/pull/41

### DIFF
--- a/.github/workflows/test-cygwin.yml
+++ b/.github/workflows/test-cygwin.yml
@@ -34,18 +34,34 @@ jobs:
         with:
           platform: x86_64
           packages: >
-            ImageMagick gcc-g++ ghostscript jpeg libfreetype-devel
-            libimagequant-devel libjpeg-devel liblapack-devel
-            liblcms2-devel libopenjp2-devel libraqm-devel
-            libtiff-devel libwebp-devel libxcb-devel libxcb-xinerama0
-            make netpbm perl
+            gcc-g++
+            ghostscript
+            ImageMagick
+            jpeg
+            libfreetype-devel
+            libimagequant-devel
+            libjpeg-devel
+            liblapack-devel
+            liblcms2-devel
+            libopenjp2-devel
+            libraqm-devel
+            libtiff-devel
+            libwebp-devel
+            libxcb-devel
+            libxcb-xinerama0
+            make
+            netpbm
+            perl
             python3${{ matrix.python-minor-version }}-cffi
             python3${{ matrix.python-minor-version }}-cython
             python3${{ matrix.python-minor-version }}-devel
             python3${{ matrix.python-minor-version }}-numpy
             python3${{ matrix.python-minor-version }}-sip
             python3${{ matrix.python-minor-version }}-tkinter
-            qt5-devel-tools subversion xorg-server-extra zlib-devel
+            qt5-devel-tools
+            subversion
+            xorg-server-extra
+            zlib-devel
 
       - name: Add Lapack to PATH
         uses: egor-tensin/cleanup-path@v3
--- a/.github/workflows/test-cygwin.yml
+++ b/.github/workflows/test-cygwin.yml
@@ -59,7 +59,7 @@ jobs:
             python3${{ matrix.python-minor-version }}-sip
             python3${{ matrix.python-minor-version }}-tkinter
             qt5-devel-tools
-            subversion
+            wget
             xorg-server-extra
             zlib-devel
 
--- a/.github/workflows/test-mingw.yml
+++ b/.github/workflows/test-mingw.yml
@@ -45,11 +45,6 @@ jobs:
       - name: Install dependencies
         run: |
           pacman -S --noconfirm \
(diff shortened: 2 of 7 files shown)


.github/workflows/test-cygwin.yml
```diff
@@ -34,18 +34,34 @@ jobs:
...
             python3${{ matrix.python-minor-version }}-cython
             python3${{ matrix.python-minor-version }}-devel
             python3${{ matrix.python-minor-version }}-numpy
             python3${{ matrix.python-minor-version }}-sip
             python3${{ matrix.python-minor-version }}-tkinter
-            qt5-devel-tools subversion xorg-server-extra zlib-devel
+            qt5-devel-tools
+            wget
+            xorg-server-extra
+            zlib-devel
```
hugovk, line 64: Moved to one per line, alphabetised, and subversion replaced with wget.

.github/workflows/test-mingw.yml
```diff
@@ -60,7 +55,12 @@ jobs:
               ${{ matrix.package }}-libtiff \
               ${{ matrix.package }}-libwebp \
               ${{ matrix.package }}-openjpeg2 \
-              subversion
+              ${{ matrix.package }}-python3-cffi \
+              ${{ matrix.package }}-python3-numpy \
+              ${{ matrix.package }}-python3-olefile \
+              ${{ matrix.package }}-python3-pip \
+              ${{ matrix.package }}-python3-setuptools \
+              ${{ matrix.package }}-wget
```
hugovk, line 63: Alphabetised, and subversion replaced with wget.

depends/install_extra_test_images.sh
```diff
@@ -1,15 +1,12 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
radarhere, line 1: For the record, what's the thinking behind this change?

For some reason, `depends/install_raqm*.sh` has `#!/usr/bin/env bash`, but the other `depends/install_*.sh` scripts have `#!/bin/bash`.

hugovk: For portability:

https://stackoverflow.com/a/10383546/724176

Probably doesn't matter too much, but this file was being updated anyway.

depends/install_extra_test_images.sh
```diff
@@ -1,15 +1,12 @@
...
-        echo "Retrying svn export..."
-        echo ""
-    fi
-
-    svn export --force https://github.com/python-pillow/pillow-depends/trunk/test_images ../Tests/images
-}
-svn_export || svn_export retry || svn_export retry || svn_export retry
+archive=test-images-main
+
+./download-and-extract.sh $archive https://github.com/python-pillow/test-images/archive/refs/heads/main.tar.gz
```
radarhere, line 6:
```suggestion
./download-and-extract.sh $archive https://github.com/python-pillow/test-images/archive/main.tar.gz
```
Copying the simpler URL you used in test-windows.yml

depends/install_extra_test_images.sh
```diff
@@ -1,15 +1,12 @@
...
-svn_export || svn_export retry || svn_export retry || svn_export retry
+archive=test-images-main
+
+./download-and-extract.sh $archive https://github.com/python-pillow/test-images/archive/main.tar.gz
+
+mv $archive/* ../Tests/images/
+
+# Cleanup old tarball and empty directory
+rm $archive.tar.gz
+rm -r $archive
```
radarhere, line 12:
```suggestion
rmdir $archive
```
If we're just removing an empty directory, then `rmdir`?

radarhere:
I've created https://github.com/hugovk/Pillow/pull/93 with some further suggestions.

radarhere (approved):

PR #6896, opened 2023-01-14
Handle more than one directory returned by pkg-config

sebastic:
tiff (4.5.0-1) in Debian results in two include directories being returned:
```
-I/usr/include/x86_64-linux-gnu -I/usr/include
```
These were not added to the compiler include directories, causing tiff support not be enabled because `tiff.h` could not be found.

PR #6921, opened 2023-01-29
Always redirect pkg-config stderr to DEVNULL

radarhere:
In `_pkg_config`, Pillow [silently catches any Python exceptions that occur.](https://github.com/python-pillow/Pillow/blob/2ae55ccbdad9c842929fb238ea1eb81d1f999024/setup.py#L279-L280)

If we're doing that, then it would seem consistent to also silence any errors that `pkg-config` might output.

Just a note about the history of the `stderr` handling in this function - #6261 started this by silencing the libs call for one loop to prevent a specific error.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -255,23 +255,21 @@ def _pkg_config(name):
         try:
             command_libs = [command, "--libs-only-L", name]
             command_cflags = [command, "--cflags-only-I", name]
-            stderr = None
             if keep_system:
                 command_libs.append("--keep-system-libs")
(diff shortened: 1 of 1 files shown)


radarhere:
I created this after https://github.com/python-pillow/Pillow/pull/6896#pullrequestreview-1273845964, where a user took the opposite view.

I've created PR #7046 to go the other way instead, undoing the error handling change from that PR.

hugovk:
Closed in favour of https://github.com/python-pillow/Pillow/pull/7046.

PR #6925, opened 2023-02-02
Added saving RGBA images as PDFs

### LINKED ISSUE #6924: Cannot save PDF with mode RGBA
### What did you do?
```python
from PIL import Image
p=Image.open("e:/Downloads/pillow_putalpha_circle_blur.jp2")
p.save("e:/Downloads/pillow_putalpha_circle_blur.pdf")
````
test file
[pillow_putalpha_circle_blur.zip](https://github.com/python-pillow/Pillow/files/10562075/pillow_putalpha_circle_blur.zip)

PDF with RGBA image encoded with JPXDecode filter

### What actually happened?
```
  File "<pyshell#268>", line 1, in <module>
    p.save("e:/Downloads/pillow_putalpha_circle_blur.pdf")
  File "C:\Python310\lib\site-packages\PIL\Image.py", line 2300, in save
  File "C:\Python310\lib\site-packages\PIL\PdfImagePlugin.py", line 155, in _save
    raise ValueError(f"cannot save mode {im.mode}")
ValueError: cannot save mode RGBA
```

* OS: windows 11
* Python: 3.10.4
* Pillow: 9.1.1

radarhere:
Resolves #6924

As suggested in the issue, adds RGBA saving for PDFs through the JPXDecode filter with JPEG2000 image data.

However, when opening the result, I found that the background wasn't transparent, but black instead.

Testing converting a transparent PNG to JPEG2000 (`Image.open("Tests/images/transparent.png").save("transparent.jp2")`), this is the case when simply saving JPEG2000 images. After looking at https://github.com/uclouvain/openjpeg/issues/416 and https://github.com/uclouvain/openjpeg/blob/2d606701e8b7aa83f657d113c3367508e99bd12b/src/lib/openjp2/jp2.c#L2015-L2024 though, it turns out it was simply a case of telling OpenJPEG which channel was the alpha channel.


hugovk (approved):
Thanks, let's add to the release notes (and regular docs if applicable).

hugovk (approved):

PR #6937, opened 2023-02-10
Added ability to pass `memoryview` object to ImageDecoder

### LINKED ISSUE #6933: Can support for the `memoryview` type be added as input to _decode (decode.c)?
Data created with help of `PyMemoryView_FromMemory` function.

Overhead for big images when converting `memoryview` to bytes is noticeable.

Code located here:
https://github.com/python-pillow/Pillow/blob/7e3bed6834e639cd0f5a410abfd8224d4e5725c5/src/decode.c#L124

Is this possible to easy add support for such feature or not?

bigcat88:
Resolves #6933

Checks whether or not input is PyMemoryView object.
If yes, then gets pointer to buffer and size.

This does not change any existing public API, just adds ability to pass `memoryview` to `Image.frombytes` method.

### DIFF
--- a/src/decode.c
+++ b/src/decode.c
@@ -116,12 +116,24 @@ _dealloc(ImagingDecoderObject *decoder) {
 
 static PyObject *
 _decode(ImagingDecoderObject *decoder, PyObject *args) {
+    PyObject *input_object, *memview = NULL;
+    Py_buffer *memview_buf;
     UINT8 *buffer;
     Py_ssize_t bufsize;
     int status;
     ImagingSectionCookie cookie;
 
-    if (!PyArg_ParseTuple(args, "y#", &buffer, &bufsize)) {
+    if (!PyArg_ParseTuple(args, "O", &input_object)) {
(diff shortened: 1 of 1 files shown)


hugovk:
The CIFuzz failure will be fixed by https://github.com/google/oss-fuzz/pull/9681.

radarhere:
I attempted to [add a test for this](https://github.com/radarhere/Pillow/commit/f5b0aaee0c6bc8dfbd749e6697b2d6ce282f8ea5), and it [passes on macOS/Ubuntu](https://github.com/radarhere/Pillow/actions/runs/4242282787), but fails on [MinGW](https://github.com/radarhere/Pillow/actions/runs/4242282784) and [Windows](https://github.com/radarhere/Pillow/actions/runs/4242282788).

radarhere:
I've created PR #6974 as an alternative to this.

bigcat88:
Thanks, I'll close this PR then as your suggestion is more elegant.

Issue #6942, opened 2023-02-11
Python 3.12: Fix building Pillow on Windows


hugovk:
For https://github.com/python-pillow/Pillow/issues/6941

Pillow doesn't build for Windows on 3.12:

```diff
diff --git a/.github/workflows/test-windows.yml b/.github/workflows/test-windows.yml
index e938e999d..306e34ca9 100644
--- a/.github/workflows/test-windows.yml
+++ b/.github/workflows/test-windows.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12-dev"]
         architecture: ["x86", "x64"]
         include:
           # PyPy 7.3.4+ only ships 64-bit binaries for Windows
```

[Python 3.12-dev x86](https://github.com/hugovk/Pillow/actions/runs/4141811740/jobs/7161777842#logs) fails:

```
   Creating library build\temp.win32-cpython-312\Release\src\_imaging.cp312-win32.lib and object build\temp.win32-cpython-312\Release\src\_imaging.cp312-win32.exp
display.obj : error LNK2001: unresolved external symbol _PyUnicode_AS_UNICODE
build\lib.win32-cpython-312\PIL\_imaging.cp312-win32.pyd : fatal error LNK1120: 1 unresolved externals
error: command 'C:\\Program Files\\Microsoft Visual Studio\\2022\\Enterprise\\VC\\Tools\\MSVC\\14.34.31933\\bin\\HostX86\\x86\\link.exe' failed with exit code 1120
```

[Python 3.12-dev x64](https://github.com/hugovk/Pillow/actions/runs/4141811740/jobs/7161777920#logs) similarly fails:

```
   Creating library build\temp.win-amd64-cpython-312\Release\src\_imaging.cp312-win_amd64.lib and object build\temp.win-amd64-cpython-312\Release\src\_imaging.cp312-win_amd64.exp
display.obj : error LNK2001: unresolved external symbol PyUnicode_AS_UNICODE
build\lib.win-amd64-cpython-312\PIL\_imaging.cp312-win_amd64.pyd : fatal error LNK1120: 1 unresolved externals
[...]

Issue #6941, opened 2023-02-11
Add support for Python 3.12


hugovk:
Tracking issue for getting the pieces into place to:

* release wheels for Python 3.12.0 beta in Pillow 10.0.0, released on 2023-07-01
* support Python 3.12.0 final in Pillow 10.1.0, released on 2023-10-15

# TODO 

* [x] Wait for 3.12.0, expected [Monday, 2023-10-02](https://peps.python.org/pep-0693/)
* [x] ~Update https://github.com/python-pillow/pillow-wheels~ https://github.com/python-pillow/Pillow/tree/main/wheels
* [x] [#7280](https://github.com/python-pillow/Pillow/pull/7280), ([#7285](https://github.com/python-pillow/Pillow/pull/7285)) Add Trove classifier, update `setup.py` version check, add to `tox.ini`, update [Python support table](https://pillow.readthedocs.io/en/latest/installation.html#python-support)
* [x] Update [CI targets table](https://pillow.readthedocs.io/en/latest/installation.html#continuous-integration-targets)

# Non-blockers

* [x] Wait for 3.12.0 on GHA: 
  * [x] Replace `3.12-dev` with `3.12` on GHA: 
  * `3.12-dev` should pick up 3.12.0 on-flight fairly soon after release, and be added to cached images later
* [ ] Wait for 3.12.0 on AppVeyor: https://github.com/appveyor/ci/issues/3879
  * [ ] Replace `3.11` with `3.12` on AppVeyor: 
* [x] Wait for 3.12.0 on Travis CI: https://travis-ci.community/t/add-python-3-12/14006?u=hugovk
  * this is base Python, we use Docker for building wheels 
  * [x] Replace `3.11` with `3.12` on Travis CI: https://github.com/python-pillow/Pillow/pull/7501

References:

* [Pillow milestones](https://github.com/python-pillow/Pillow/milestones?direction=asc&sort=due_date&state=open)
* [PEP 693 – Python 3.12 Release Schedule](https://peps.python.org/pep-0693/)
* https://github.com/python-pillow/Pillow/issues/6575

radarhere:
[...]

PR #6943, opened 2023-02-11
Test Python 3.12-dev

hugovk:
For https://github.com/python-pillow/Pillow/issues/6941.

Fixes https://github.com/python-pillow/Pillow/issues/6942.

 * [Python 3.12](https://peps.python.org/pep-0693/) is in alpha and I expect we'll release wheels during the beta
 * Let's start testing now
 * ~Not included Windows yet, see https://github.com/python-pillow/Pillow/issues/6942~


radarhere:
...I meant to add that commit to my own repository, but accidentally pushed it here instead.

It re-adds Focal to the CI targets table, because we are still testing that through our Docker builds, and removes 'x86-64' from the list of Python 3.10 Jammy builds, as it it already listed in the previous row.

radarhere (approved):

hugovk:
That's fine, thank you for the update!

PR #6281, opened 2022-05-07
Updated codecov action to v3

radarhere:

PR #6944, opened 2023-02-11
Enable gcov support for codecov/codecov-action

hugovk:
I noticed we're missing code coverage for the C files on Codecov.

For example, no C files listed at:

image

https://app.codecov.io/gh/python-pillow/Pillow/tree/main/src

Looks like this happened when switched bumped some workflows from v1 of the Codecov Action to v3, and others from the Codecov Bash Uploader to the v3 Action in https://github.com/python-pillow/Pillow/pull/6281, and coverage increased, because there's better average coverage on the Python files (https://app.codecov.io/gh/python-pillow/Pillow/commit/53b6e5f4bf9a6a68e29109c457b12d17761d0035):

image

Now I remember I had earlier opened https://github.com/codecov/codecov-action/issues/566, where the v1->v2 upgrade lost C coverage. (Codecov never replied.)

Anyway, by now, the Bash Uploader has been retired (https://github.com/codecov/codecov-bash) and there's a new Uploader (https://github.com/codecov/uploader).

Better yet, the Codecov Action has added support for gcov (used for C coverage) in v3 (https://github.com/codecov/codecov-action/pull/688).

I tried enabling it for all workflows, but for some reason it only worked for the Docker one, so let's enable it there at least to get us back on track. The others will need more investigation.

### DIFF
--- a/.github/workflows/test-docker.yml
+++ b/.github/workflows/test-docker.yml
@@ -87,6 +87,7 @@ jobs:
       with:
         flags: GHA_Docker
         name: ${{ matrix.docker }}
+        gcov: true
 
   success:
     permissions:


nulano:
I've created https://github.com/hugovk/Pillow/pull/94 to also upload C coverage for Ubuntu/macOS.

Cygwin does not seem to be uploading coverage at all: https://github.com/python-pillow/Pillow/actions/runs/4153272820/jobs/7184796519#step:15:26
> [2023-02-11T22:28:10.252Z] ['error'] There was an error running the uploader: Error fetching git root. Please try using the -R flag. Error: Error running external program: Error: spawnSync git ENOENT

hugovk:
Good stuff, thank you!

hugovk:
And please could you report the MSYS2 error to https://github.com/codecov/codecov-action? Looks like the Action needs to call their uploader client differently in this case.

PR #6951, opened 2023-02-15
image-file-formats.rst: document WebP 'xmp' option

jzern:


radarhere:
The functionality was added in #2761

PR #6952, opened 2023-02-15
Corrected documented WebP quality range

jzern:
0-100, not 1-100


radarhere (approved):
The minimum value of 1 was documented in #390, but then #2761 added documentation talking about what happens when the value is 0, so our current version is inconsistent.

PR #6954, opened 2023-02-16
Added "corners" argument to ImageDraw rounded_rectangle()

### LINKED ISSUE #6953: Add a corners parameter for `ImageDraw.rounded_rectangle`?
### What's your feature request?
To add an optional bool tuple parameter (`corners=(True, True, True, True)`), or 4 optional bool parameters (`top_left=True, top_right=True,  ...`) describing which corners of a rectangle should be rounded for `ImageDraw.rounded_rectangle` and possibly similar functions.

* OS: Windows 10
* Python: 3.10.10
* Pillow: latest

## Example of feature request
```python
draw.rounded_rectangle((x0, y0, x1, y1), radius=4, fill = white, corners=(True, False, True, False)) # TL TR BL BR
draw.rounded_rectangle((x0, y0, x1, y1), radius=4, fill=white, top_right=False, bottom_right=False)
```

radarhere:
Resolves #6953, by adding a `corners` argument to `ImageDraw.rounded_rectangle` - a tuple of booleans for whether or not to round each corner, going clockwise, starting with top left and ending with bottom left.


hugovk:
Re: https://understandlegacycode.com/blog/what-is-wrong-with-boolean-parameters/

Rather than something like:

```python
rounded_rectangle((10, 20, 190, 180), 30, (True, False, True, True))
```

Would it be better if we made `corners` a required keyword argument?

```python
rounded_rectangle((10, 20, 190, 180), 30, rounded=(True, False, True, True))
```

radarhere:
Ok, I've pushed a commit to only allow `corners` to be used as a keyword argument.
```pycon
>>> draw.rounded_rectangle((10, 20, 190, 180), 30, "red", "green", 5, (True, True, True, True))
TypeError: rounded_rectangle() takes from 2 to 6 positional arguments but 7 were given
>>> draw.rounded_rectangle((10, 20, 190, 180), 30, "red", "green", 5, corners=(True, True, True, True))
>>> 
```

src/PIL/ImageDraw.py
```diff
@@ -296,15 +296,14 @@ def rectangle(self, xy, fill=None, outline=None, width=1):
             self.draw.draw_rectangle(xy, ink, 0, width)
 
     def rounded_rectangle(
-        self, xy, radius=0, fill=None, outline=None, width=1, corners=None
+        self, xy, radius=0, fill=None, outline=None, width=1, **kwargs
```
hugovk, line 299: Or something like this for more transparency? Better for `help(rounded_rectangle)` and so on.

```suggestion
        self, xy, radius=0, fill=None, outline=None, width=1, *, corners=None
```

Or even this?

```suggestion
        self, xy, radius=0, fill=None, outline=None, width=1, *, corners=(True, True, True, True)
```

(And update/delete the below to match.)

ChrisMGeo: I think `corners=(True, True, True, True)` is best. But what would the ordering be? Top left, top right, bottom left, bottom right?

radarhere: I've updated the commit to use the first suggestion.

As for the ordering, yes, that is the ordering I've used. It mirrors the [ordering in CSS for `border-radius`](https://developer.mozilla.org/en-US/docs/Web/CSS/border-radius).

PR #6957, opened 2023-02-18
Updated Gitter URL

radarhere:
https://gitter.im/python-pillow/Pillow now redirects to https://app.gitter.im/#/room/#python-pillow_Pillow:gitter.im


hugovk:
If I was going to share a link to someone, https://gitter.im/python-pillow/Pillow looks nicer, and is easier to remember, than https://app.gitter.im/#/room/#python-pillow_Pillow:gitter.im.

Shall we keep the original? Is there any indication the original will break?

radarhere:
Sure, we can keep the original. The only indication is the redirect.

PR #6958, opened 2023-02-18
Replace Python code-blocks with double colons

radarhere:
#6754 has pointed out that sometimes we use `.. code-block:: python` and sometimes we use `::`.

Because `::` is rendered as a single colon, perhaps `.. code-block:: python` is used where we don't want to show a colon to the reader. This PR replaces `.. code-block:: python` with `::` if it is already following a colon, making the documentation syntax consistent.


hugovk:
We can also use `::` in other cases too, when there is no colon.

For example:


```rst

Some header
-----------

::

    a = 1

Some more text. ::

    a = 1
```

Renders as:

image

radarhere:
Ok, I've replaced all instances of `.. code-block:: python`

PR #6965, opened 2023-02-23
ICO improvements, part 1

akx:
This PR is extracted from a larger effort to improve how Pillow works with certain old-school legacy ICO and CUR files that I've encountered working on https://github.com/akx/res-extract.

It's mostly simple refactoring and some removal of old, dead code; the last commit could technically have downstream effects for users who rely, for some reason, on being able to read `ico.ico.entry[x]["dim"]` or `["square"]`, but I think that's quite fringe.


src/PIL/IcoImagePlugin.py
```diff
@@ -304,10 +314,13 @@ class IcoImageFile(ImageFile.ImageFile):
     format = "ICO"
     format_description = "Windows Icon"
 
+    # For subclasses.
+    _ico_file_class = IcoFile
```
radarhere, line 318: Did you want to talk a bit about why this is useful?

src/PIL/IcoImagePlugin.py
```diff
@@ -142,11 +142,8 @@ def __init__(self, buf):
             msg = "No images were found"
             raise TypeError(msg)
 
-        self.entry = sorted(self.entry, key=lambda x: x["color_depth"])
-        # ICO images are usually squares
-        # self.entry = sorted(self.entry, key=lambda x: x['width'])
-        self.entry = sorted(self.entry, key=lambda x: x["square"])
-        self.entry.reverse()
```
radarhere, line 179: I presume you dropped the first sort because you think it is necessary with the second sort? To me, it is still necessary in the scenario where two entries have the same "square" value.

For a simple example,
```python
x = [9,8,7,6,5,4,3,2,1]
x.sort(key=lambda x: x // 2)
print(x)
```
does not produce the same output as
```python
x = [9,8,7,6,5,4,3,2,1]
x.sort(key=lambda x: x % 2)
x.sort(key=lambda x: x // 2)
print(x)
```

radarhere:
Your two refactoring commits feel like they're specifically useful to your project. Are you sure you wouldn't rather register your own plugin in your code, instead of trying to subclass the ICO format?

src/PIL/IcoImagePlugin.py
akx: Ah, yeah, you're right (and this ties to the fact that `.sort()` is guaranteed stable).

The intent of the original code has been to sort so the most color-rich, largest image is preferred, but that's not clear from the comments (and commented-out old code). I'll add `color_depth` back to the sort key.

Tests/test_file_ico.py
```diff
@@ -229,3 +229,8 @@ def test_draw_reloaded(tmp_path):
 
     with Image.open(outfile) as im:
         assert_image_equal_tofile(im, "Tests/images/hopper_draw.ico")
+
+
+def test_empty():
+    with pytest.raises(SyntaxError):
+        IcoImagePlugin.IcoImageFile("Tests/images/no_icons.ico")
```
radarhere, line 236:
```suggestion
    with pytest.raises(SyntaxError) as e:
        IcoImagePlugin.IcoImageFile("Tests/images/no_icons.ico")
    assert str(e.value) == "No images were found"
```
There was already a SyntaxError raised before your code change, so this alters the test to check the message.

akx:
> Your two refactoring commits feel like they're specifically useful to your project. Are you sure you wouldn't rather register your own plugin in your code, instead of trying to subclass the ICO format?

@radarhere No, they're not specific to my project – these improvements to the ICO machinery are necessary because CUR files are a variant of ICO files, and they're not currently being correctly read by PIL (they're missing transparency, for one). 

So, instead of hacking in transparency support into the current [quite spare and not quite correct CurImagePlugin](https://github.com/python-pillow/Pillow/blob/879e77a44604b31e271711898b035a5caba8cb73/src/PIL/CurImagePlugin.py), the following work (in flux, will change, doesn't work fully yet) over at https://github.com/python-pillow/Pillow/compare/main...akx:Pillow:cur-support makes the Cur plugin extend the Ico plugin instead.

(I should've probably shown that context in the PR message for clarity... 😄)

Tests/test_file_ico.py
akx: Good point – changed to `with pytest.raises(SyntaxError, match="No images were found"):` :)

radarhere:
The test image that you have added - that's something that can be distributed under our Pillow license?

Tests/test_file_ico.py
```diff
@@ -229,3 +229,8 @@ def test_draw_reloaded(tmp_path):
 
     with Image.open(outfile) as im:
         assert_image_equal_tofile(im, "Tests/images/hopper_draw.ico")
+
+
+def test_empty():
+    with pytest.raises(SyntaxError, match="No images were found"):
```
radarhere, line 235:
```suggestion
    with pytest.raises(SyntaxError, match="No icons were found"):
```

src/PIL/IcoImagePlugin.py
```diff
@@ -138,55 +136,70 @@ def __init__(self, buf):
...
-                    and ceil(log(icon_header["nb_color"], 2))
-                )
-                or 256
-            )
+            self.entry.append(self._read_icon_header(buf))
 
-            icon_header["dim"] = (icon_header["width"], icon_header["height"])
-            icon_header["square"] = icon_header["width"] * icon_header["height"]
+        if not self.entry:
+            msg = "No images were found"
```
radarhere, line 142:
```suggestion
            msg = "No icons were found"
```
Rather than the more generic word "images", how about something specific to this format, to hopefully avoid confusion with our Image class? This would also then be similar to "No cursors were found"

akx: As mentioned in my earlier comment, the future followup to this will base CUR image support on this plugin, so this moves the error to the common "images" (which covers both "icons" and "cursors"). If you like, I can make this a class field, so the deriving cursor plugin can change the message to `No cursors were found`..?

akx:
> The test image that you have added - that's something that can be distributed under our Pillow license?

Yes, I engineered the six bytes by hand based on https://github.com/python-pillow/Pillow/blob/main/Tests/images/no_cursors.cur.

radarhere:
> So, instead of hacking in transparency support into the current [quite spare and not quite correct CurImagePlugin](https://github.com/python-pillow/Pillow/blob/879e77a44604b31e271711898b035a5caba8cb73/src/PIL/CurImagePlugin.py), the following work (in flux, will change, doesn't work fully yet) over at https://github.com/python-pillow/Pillow/compare/main...akx:Pillow:cur-support makes the Cur plugin extend the Ico plugin instead. These were easily separable parts from that work.

If this is just paving the way for future changes, then I would rather wait until those future changes are ready for this to be considered. I'm uneasy about trusting that the next round of changes won't require this to be refactored in some way.

akx:
@radarhere The changes in this PR have been carefully split out from that work for ease of review, and I'm quite confident these don't need to be changed...

radarhere:
While I can appreciate trying to avoid large PRs, if the individual parts don't work in isolation to improve the functionality or API of Pillow, then I think they have become too granular.

PR #6966, opened 2023-02-23
Move linting to ruff

akx:
This PR switches linting (flake8, isort, <del>bandit</del>, yesqa, pygrep-hooks) for Pillow to [the `ruff` linter](https://github.com/charliermarsh/ruff).

* Running a full `pre-commit` run is much faster (45.15 seconds of CPU time to 8.62 seconds on my machine)
* It's easier to enable more rules if need be since the configuration is all in one place
* 👉  The only rule that may not be yet supported by `ruff` that had been previously enabled is `rst-backticks` from `pygrep-hooks`.
  * That being said, this PR still enables _more_ lints than had been in use from `pygrep-hooks` since it enables the whole `PGH` family.
* <del>👉 Instead of touching expressions like `x, y, l, r, w, a, d, f = metrics[ix]`, I opted to adding specific `noqa` directives to quiesce Ruff's "ambiguous name" for `l` (which could be an `1` or an `I` 😁)</del> No longer an issue via #6971.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,20 +1,15 @@
 repos:
-  - repo: https://github.com/asottile/pyupgrade
-    rev: v3.13.0
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.1.4
     hooks:
-      - id: pyupgrade
-        args: [--py38-plus]
+      - id: ruff
+        args: [--fix, --exit-non-zero-on-fix]
 
   - repo: https://github.com/psf/black-pre-commit-mirror
     rev: 23.9.1
     hooks:
       - id: black
         args: [--target-version=py38]
-
-  - repo: https://github.com/PyCQA/isort
-    rev: 5.12.0
-    hooks:
-      - id: isort
 
   - repo: https://github.com/PyCQA/bandit
     rev: 1.7.5
@@ -23,28 +18,15 @@ repos:
       args: [--severity-level=high]
       files: ^src/
 
-  - repo: https://github.com/asottile/yesqa
-    rev: v1.5.0
-    hooks:
-      - id: yesqa
-
   - repo: https://github.com/Lucas-C/pre-commit-hooks
     rev: v1.5.4
     hooks:
       - id: remove-tabs
         exclude: (Makefile$|\.bat$|\.cmake$|\.eps$|\.fits$|\.gd$|\.opt$)
 
-  - repo: https://github.com/PyCQA/flake8
-    rev: 6.1.0
-    hooks:
-      - id: flake8
-        additional_dependencies:
-          [flake8-2020, flake8-errmsg, flake8-implicit-str-concat, flake8-logging]
-
   - repo: https://github.com/pre-commit/pygrep-hooks
     rev: v1.10.0
     hooks:
-      - id: python-check-blanket-noqa
       - id: rst-backticks
 
   - repo: https://github.com/pre-commit/pre-commit-hooks
--- a/Makefile
+++ b/Makefile
@@ -49,7 +49,7 @@ help:
 	@echo "  install            make and install"
 	@echo "  install-coverage   make and install with C coverage"
 	@echo "  lint               run the lint checks"
-	@echo "  lint-fix           run Black and isort to (mostly) fix lint issues"
+	@echo "  lint-fix           run Ruff to (mostly) fix lint issues"
 	@echo "  release-test       run code and package tests before release"
 	@echo "  test               run tests on installed Pillow"
 
@@ -118,6 +118,6 @@ lint:
 .PHONY: lint-fix
 lint-fix:
 	python3 -c "import black" > /dev/null 2>&1 || python3 -m pip install black
-	python3 -c "import isort" > /dev/null 2>&1 || python3 -m pip install isort
 	python3 -m black --target-version py38 .
-	python3 -m isort .
+	python3 -c "import ruff" > /dev/null 2>&1 || python3 -m pip install ruff
+	python3 -m ruff --fix .
--- a/_custom_build/backend.py
+++ b/_custom_build/backend.py
@@ -1,6 +1,6 @@
 import sys
 
-from setuptools.build_meta import *  # noqa: F401, F403
+from setuptools.build_meta import *  # noqa: F403
 from setuptools.build_meta import build_wheel
 
 backend_class = build_wheel.__self__.__class__
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -77,8 +77,34 @@ package-dir = {"" = "src"}
 [tool.setuptools.dynamic]
 version = {attr = "PIL.__version__"}
 
-[tool.isort]
-profile = "black"
+[tool.ruff]
+target-version = "py38"
+line-length = 88
+select = [
+  "E", # pycodestyle errors
+  "EM", # flake8-errmsg
+  "F", # pyflakes errors
+  "I", # isort
+  "ISC", # flake8-implicit-str-concat
+  "PGH", # pygrep-hooks
+  "RUF100", # unused noqa (yesqa)
+  "UP", # pyupgrade
+  "W", # pycodestyle warnings
+  "YTT", # flake8-2020
+  # "LOG", # TODO: enable flake8-logging when it's not in preview anymore
+]
+extend-ignore = [
+  "E203", # Whitespace before ':'
+  "E221", # Multiple spaces before operator
+  "E226", # Missing whitespace around arithmetic operator
+  "E241", # Multiple spaces after ','
+]
+
+[tool.ruff.per-file-ignores]
+"Tests/*.py" = ["I001"]
+
+[tool.ruff.isort]
+known-first-party = ["PIL"]
 
 [tool.pytest.ini_options]
 addopts = "-ra --color=yes"
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -3100,7 +3100,7 @@ def fromarray(obj, mode=None):
         try:
             mode, rawmode = _fromarray_typemap[typekey]
         except KeyError as e:
-            msg = "Cannot handle this data type: %s, %s" % typekey
+            msg = "Cannot handle this data type: %s, %s" % typekey  # noqa: UP031
             raise TypeError(msg) from e
     else:
         rawmode = mode
--- a/src/PIL/ImageFilter.py
+++ b/src/PIL/ImageFilter.py
@@ -222,7 +222,7 @@ class UnsharpMask(MultibandFilter):
 
     .. _digital unsharp masking: https://en.wikipedia.org/wiki/Unsharp_masking#Digital_unsharp_masking
 
-    """  # noqa: E501
+    """
 
     name = "UnsharpMask"
 
--- a/src/PIL/IptcImagePlugin.py
+++ b/src/PIL/IptcImagePlugin.py
@@ -18,10 +18,9 @@ import os
 import tempfile
 
 from . import Image, ImageFile
-from ._binary import i8
+from ._binary import i8, o8
 from ._binary import i16be as i16
 from ._binary import i32be as i32
-from ._binary import o8
 
 COMPRESSION = {1: "raw", 5: "jpeg"}
 
--- a/.editorconfig
+++ b/.editorconfig
@@ -13,7 +13,7 @@ indent_style = space
 
 trim_trailing_whitespace = true
 
-[*.yml]
+[*.{toml,yml}]
 # Two-space indentation
 indent_size = 2
 
--- a/.flake8
+++ b/.flake8
@@ -1,3 +1,0 @@
-[flake8]
-extend-ignore = E203
-max-line-length = 88
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -5,6 +5,7 @@ include *.md
 include *.py
 include *.rst
 include *.sh
+include *.toml
 include *.txt
 include *.yaml
 include .flake8
--- a/Tests/bench_cffi_access.py
+++ b/Tests/bench_cffi_access.py
@@ -45,7 +45,7 @@ def test_direct():
 
     assert caccess[(0, 0)] == access[(0, 0)]
 
-    print("Size: %sx%s" % im.size)
+    print("Size: %sx%s" % im.size)  # noqa: UP031
     timer(iterate_get, "PyAccess - get", im.size, access)
     timer(iterate_set, "PyAccess - set", im.size, access)
     timer(iterate_get, "C-api - get", im.size, caccess)
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -318,14 +318,14 @@ def setup(app):
 
 
 linkcheck_allowed_redirects = {
-    r"https://www.bestpractices.dev/projects/6331": r"https://www.bestpractices.dev/en/.*",  # noqa: E501
-    r"https://badges.gitter.im/python-pillow/Pillow.svg": r"https://badges.gitter.im/repo.svg",  # noqa: E501
-    r"https://gitter.im/python-pillow/Pillow?.*": r"https://app.gitter.im/#/room/#python-pillow_Pillow:gitter.im?.*",  # noqa: E501
-    r"https://pillow.readthedocs.io/?badge=latest": r"https://pillow.readthedocs.io/en/stable/?badge=latest",  # noqa: E501
+    r"https://www.bestpractices.dev/projects/6331": r"https://www.bestpractices.dev/en/.*",
+    r"https://badges.gitter.im/python-pillow/Pillow.svg": r"https://badges.gitter.im/repo.svg",
+    r"https://gitter.im/python-pillow/Pillow?.*": r"https://app.gitter.im/#/room/#python-pillow_Pillow:gitter.im?.*",
+    r"https://pillow.readthedocs.io/?badge=latest": r"https://pillow.readthedocs.io/en/stable/?badge=latest",
     r"https://pillow.readthedocs.io": r"https://pillow.readthedocs.io/en/stable/",
-    r"https://tidelift.com/badges/package/pypi/Pillow?.*": r"https://img.shields.io/badge/.*",  # noqa: E501
-    r"https://zenodo.org/badge/17549/python-pillow/Pillow.svg": r"https://zenodo.org/badge/doi/[\.0-9]+/zenodo.[0-9]+.svg",  # noqa: E501
-    r"https://zenodo.org/badge/latestdoi/17549/python-pillow/Pillow": r"https://zenodo.org/record/[0-9]+",  # noqa: E501
+    r"https://tidelift.com/badges/package/pypi/Pillow?.*": r"https://img.shields.io/badge/.*",
+    r"https://zenodo.org/badge/17549/python-pillow/Pillow.svg": r"https://zenodo.org/badge/doi/[\.0-9]+/zenodo.[0-9]+.svg",
+    r"https://zenodo.org/badge/latestdoi/17549/python-pillow/Pillow": r"https://zenodo.org/record/[0-9]+",
 }
 
 # sphinx.ext.extlinks
--- a/src/PIL/EpsImagePlugin.py
+++ b/src/PIL/EpsImagePlugin.py
@@ -122,7 +122,7 @@ def Ghostscript(tile, size, fp, scale=1, transparency=False):
         gs_binary,
         "-q",  # quiet mode
         "-g%dx%d" % size,  # set output geometry (pixels)
-        "-r%fx%f" % res,  # set input DPI (dots per inch)
+        "-r%fx%f" % res,  # set input DPI (dots per inch)  # noqa: UP031
         "-dBATCH",  # exit after processing
         "-dNOPAUSE",  # don't pause between pages
         "-dSAFER",  # safe mode
--- a/src/PIL/IcnsImagePlugin.py
+++ b/src/PIL/IcnsImagePlugin.py
@@ -392,7 +392,7 @@ if __name__ == "__main__":
         imf = IcnsImageFile(fp)
         for size in imf.info["sizes"]:
             imf.size = size
-            imf.save("out-%s-%s-%s.png" % size)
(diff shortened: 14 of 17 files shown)


Makefile
```diff
@@ -114,7 +114,4 @@ lint:
 
 .PHONY: lint-fix
 lint-fix:
-	python3 -c "import black" > /dev/null 2>&1 || python3 -m pip install black
-	python3 -c "import isort" > /dev/null 2>&1 || python3 -m pip install isort
-	python3 -m black --target-version py37 .
-	python3 -m isort .
+	ruff --fix .
```
hugovk, line 117: Why remove Black?

Please also make sure Ruff is installed.

.pre-commit-config.yaml
```diff
@@ -1,4 +1,10 @@
 repos:
+  - repo: https://github.com/charliermarsh/ruff-pre-commit
```
hugovk, line 2: Let's move Ruff after Black: Black might fix some things that Flake8 would complain about.

.pre-commit-config.yaml
```diff
@@ -1,4 +1,10 @@
...
+    rev: 'v0.0.252'
```
hugovk, line 3: To match others?
```suggestion
    rev: v0.0.252
```

.pre-commit-config.yaml
```diff
@@ -8,42 +14,12 @@ repos:
...
-    rev: 5.12.0
-    hooks:
-      - id: isort
-
-  - repo: https://github.com/PyCQA/bandit
-    rev: 1.7.4
-    hooks:
-    - id: bandit
-      args: [--severity-level=high]
-      files: ^src/
```
hugovk, line 21: How does Ruff deal with limiting the Bandit severity and files?

src/PIL/IptcImagePlugin.py
```diff
@@ -18,10 +18,9 @@
 import tempfile
 
 from . import Image, ImageFile
-from ._binary import i8
+from ._binary import i8, o8
```
hugovk, line 21: What happened here? Does Ruff not support isort's Black profile?

setup.cfg
```diff
@@ -63,13 +63,6 @@ tests =
     pytest-cov
     pytest-timeout
 
-[flake8]
-extend-ignore = E203
```
hugovk, line 67: Does Ruff not have E203?

setup.py
```diff
@@ -22,7 +22,7 @@
 def get_version():
     version_file = "src/PIL/_version.py"
     with open(version_file, encoding="utf-8") as f:
-        exec(compile(f.read(), version_file, "exec"))
+        exec(compile(f.read(), version_file, "exec"))  # noqa: S102
```
hugovk, line 25: Please limit Bandit to the `src` directory.

src/PIL/BdfFontFile.py
```diff
@@ -64,7 +64,7 @@ def bdf_char(f):
         bitmap.append(s[:-1])
     bitmap = b"".join(bitmap)
 
-    [x, y, l, d] = [int(p) for p in props["BBX"].split()]
+    [x, y, l, d] = [int(p) for p in props["BBX"].split()]  # noqa: E741
```
hugovk, line 67: Please can you check what `l` represents? Is there a better name?

src/PIL/Image.py
```diff
@@ -767,7 +767,7 @@ def tobytes(self, encoder_name="raw", *args):
 
         data = []
         while True:
-            l, s, d = e.encode(bufsize)
+            l, s, d = e.encode(bufsize)  # noqa: E741
```
hugovk, line 770: Please can you check what `l` represents? Is there a better name?

src/PIL/ImageFile.py
```diff
@@ -530,12 +530,12 @@ def _encode_tile(im, fp, tile, bufsize, fh, exc=None):
...
             if encoder.pushes_fd:
                 encoder.setfd(fp)
-                l, s = encoder.encode_to_pyfd()
+                l, s = encoder.encode_to_pyfd()  # noqa: E741
             else:
                 if exc:
                     # compress to Python file-compatible object
                     while True:
-                        l, s, d = encoder.encode(bufsize)
+                        l, s, d = encoder.encode(bufsize)  # noqa: E741
```
hugovk, line 538: Please can you check what `l` represents? Is there a better name?

src/PIL/PcfFontFile.py
```diff
@@ -86,7 +86,7 @@ def __init__(self, fp, charset_encoding="iso8859-1"):
 
         for ch, ix in enumerate(encoding):
             if ix is not None:
-                x, y, l, r, w, a, d, f = metrics[ix]
+                x, y, l, r, w, a, d, f = metrics[ix]  # noqa: E741
```
hugovk, line 89: Please can you check what `l` represents? Is there a better name?
(And below)

src/PIL/TiffImagePlugin.py
```diff
@@ -1845,7 +1845,7 @@ def _save(im, fp, filename):
         e.setimage(im.im, (0, 0) + im.size)
         while True:
             # undone, change to self.decodermaxblock:
-            l, s, d = e.encode(16 * 1024)
+            l, s, d = e.encode(16 * 1024)  # noqa: E741
```
hugovk, line 1848: Please can you check what `l` represents? Is there a better name?

hugovk (commented):
> * Running a full `pre-commit` run is much faster (45.15 seconds of CPU time to 8.62 seconds on my machine)

That's a good speedup! I get 5s -> 2s for `pre-commit run --all-files`, which is still welcome. On pre-commit.ci, it's [23.5s](https://results.pre-commit.ci/run/github/5171600/1677155148.Es8LdiESQVC7BxML-Cwxgg) -> [12.3s](https://results.pre-commit.ci/run/github/5171600/1677154839.tN83z3JTTTiN4YNogtkGjA).

.pre-commit-config.yaml
akx: It doesn't, at present. The current configuration enables all Bandit rules except the three ignored ones, so in that sense it's stricter than before.

If Bandit should be disabled altogether for tests, that could be done with per-file-ignores.

src/PIL/IptcImagePlugin.py
akx: isort is a bit looser than Ruff with these; Ruff wants either everything or nothing to be grouped like this, and isort doesn't always care.

setup.cfg
akx: It doesn't.

Makefile
akx: Oops, oversight there.

akx: That said, should this actually be `pre-commit run --all-files black ruff`?

pyproject.toml
```diff
@@ -0,0 +1,36 @@
...
+# Ignore Bandit in tests
+"Tests/*" = ["S"]
+"setup.py" = ["S"]
+# TODO: these files have some ambiguous `l` identifiers,
+#       which should be cleaned up.
+"src/PIL/BdfFontFile.py" = ["E741"]
+"src/PIL/Image.py" = ["E741"]
+"src/PIL/ImageFile.py" = ["E741"]
+"src/PIL/PcfFontFile.py" = ["E741"]
+"src/PIL/TiffImagePlugin.py" = ["E741"]
```
akx, line 36: @hugovk Instead of diving deep into various font formats etc. right now, I opted to disable these codes for those files for now; I can gladly add subsequent PRs, but those would definitely be easier to review separately.

src/PIL/BdfFontFile.py
Yay295: > The word BBX followed by the width in x (BBw), height in y (BBh), and x and y displacement
(BBox, BBoy) of the lower left corner from the origin of the character.

https://www.x.org/docs/BDF/bdf.pdf

So `l` and `d` should be the horizontal and vertical displacement. `x` and `y` could be renamed to `width` and `height` as well.

src/PIL/Image.py
Yay295: `l`, `s`, and `d` are the length of the encoded data, an error code, and the encoded data.

src/PIL/ImageFile.py
Yay295: `l`, `s`, and `d` are the length of the encoded data, an error code, and the encoded data.

src/PIL/TiffImagePlugin.py
Yay295: `l`, `s`, and `d` are the length of the encoded data, an error code, and the encoded data.

src/PIL/PcfFontFile.py
Yay295: https://github.com/python-pillow/Pillow/blob/583dd3b4245649ddd367c2dcbbf2bf9fb7835017/src/PIL/PcfFontFile.py#L168

pyproject.toml
hugovk: Sure. Maybe the per-line `noqa`s are better than per-file ignores, to prevent new ones being introduced?

akx: @hugovk Well, I was planning on cleaning these up in subsequent PRs in the following days – and should new ones get introduced in that timeframe, it shouldn't be a big problem to clean them up then.

hugovk:
The addition of `pyproject.toml` has changed the build process from `setup.py`-based to `pyproject.toml`-based, leading to failures on MinGW:

https://github.com/python-pillow/Pillow/actions/runs/4253283135/jobs/7404119827

akx:
> The addition of `pyproject.toml` has changed the build process from `setup.py`-based to `pyproject.toml`-based, leading to failures on MinGW

Ah, should've known there are still uncivilized OSes like that... 😁 

Switched to `ruff.toml` instead 👍

akx:
@hugovk CI is green now, and #6971 will fix up the local variable names.

Makefile
```diff
@@ -115,6 +115,6 @@ lint:
 .PHONY: lint-fix
 lint-fix:
 	python3 -c "import black" > /dev/null 2>&1 || python3 -m pip install black
-	python3 -c "import isort" > /dev/null 2>&1 || python3 -m pip install isort
 	python3 -m black --target-version py37 .
-	python3 -m isort .
+	which -s ruff || python3 -m pip install ruff
+	ruff --fix .
```
radarhere, line 120:
```suggestion
	python3 -c "import ruff" > /dev/null 2>&1 || python3 -m pip install ruff
	python3 -m ruff --fix .
```
On my machine, when I run `python3 -m pip install ruff`, it isn't automatically available to be run as `ruff`.

This is more flexible.

src/PIL/IptcImagePlugin.py
hugovk: Testing, isort always sorts this like the original, with or without the Black profile.

Anyway, I don't mind too much, as long as it remains relatively stable.

On that topic, maybe we should wait until Ruff is at "v1" or "stable", which looks to be close:

https://github.com/charliermarsh/ruff/issues/1992

hugovk:
> Switched to `ruff.toml` instead 👍

Makes sense, we'll probably have to move to `pyproject.toml` at some point, but I agree it's out of scope for this. Please could you also add `check-toml` to `.pre-commit-config.yaml`?

aclark4life:
Yeah lots not switch, just add ruff … if possible/practical to do so.

.pre-commit-config.yaml
cclauss: https://beta.ruff.rs/docs/usage
> Note that Ruff's pre-commit hook should run before Black, isort, and other formatting tools.

.pre-commit-config.yaml
cclauss: charliermarsh/ruff#3239

src/PIL/IptcImagePlugin.py
cclauss: Ruff has Black profile as the ONLY profile possible.

cclauss:
If you do `ruff --format=github` then you get GitHub Annotations like:
image
Which some contributors might find quite helpful.  The problem is that when you run in pre-commit, you are unsure if `$CI` is true or false and you probably do not want `--format=github` unless it is true.

I run a separate GitHub that is superfast (<10 seconds) to quicky add annotations to code changes.
* internetarchive/openlibrary#7606

ruff.toml
hugovk, line 1: I still need to review the rest of the PR, but let's name this file `.ruff.toml`. Our other config files also generally use the dot-prefix.

https://beta.ruff.rs/docs/configuration/

akx: Done deal.

.ruff.toml
```diff
@@ -0,0 +1,28 @@
...
+line-length = 88
+select = [
+    "E", # pycodestyle errors
+    "EM", # flake8-errmsg
+    "F", # pyflakes errors
+    "I", # isort
+    "ISC", # flake8-implicit-str-concat
+    "PGH", # pygrep-hooks
+    "RUF100",  # unused noqa (yesqa)
+    "S", # flake8-bandit
```
hugovk, line 11:
```suggestion
```

.ruff.toml
```diff
@@ -0,0 +1,28 @@
...
+    "W", # pycodestyle warnings
+    "YTT", # flake8-2020
+]
+ignore = [
+    # TODO: these could be enabled later
+    "S101", # use of assert
+    "S110", # try-except-pass
+    "S112", # try-except-continue
+]
```
hugovk, line 20:
```suggestion
```

.ruff.toml
```diff
@@ -0,0 +1,28 @@
...
+
+[isort]
+known-first-party = ["PIL"]
+
+[per-file-ignores]
+# Ignore Bandit in tests
+"Tests/*" = ["S"]
+"setup.py" = ["S"]
```
hugovk, line 28:
```suggestion
```

akx:
> > * Running a full `pre-commit` run is much faster (45.15 seconds of CPU time to 8.62 seconds on my machine)

I was measuring CPU time on purpose – wallclock time isn't affected that much, but the old tooling isn't able to utilize all cores on my machine (macOS 12.6.3, 2,4 GHz 8-Core Intel Core i9, etc.).

hugovk:
> * 👉  The only rule that may not be yet supported by `ruff` that had been previously enabled is `rst-backticks` from `pygrep-hooks`.
>   
>   * That being said, this PR still enables _more_ lints than had been in use from `pygrep-hooks` since it enables the whole `PGH` family.


https://beta.ruff.rs/docs/rules/#pygrep-hooks-pgh has four of the 10 hooks from https://github.com/pre-commit/pygrep-hooks:

- PGH001	`eval`	No builtin eval() allowed	
- PGH002	`deprecated-log-warn`	warn is deprecated in favor of warning	
- PGH003	`blanket-type-ignore`	Use specific rule codes when ignoring type issues	
- PGH004	`blanket-noqa`	Use specific rule codes when using noqa

`rst-backticks` is useful, let's add it back via the original `pygrep-hooks`.

Makefile
```diff
@@ -50,7 +50,7 @@ help:
 	@echo "  install            make and install"
 	@echo "  install-coverage   make and install with C coverage"
 	@echo "  lint               run the lint checks"
-	@echo "  lint-fix           run Black and isort to (mostly) fix lint issues"
+	@echo "  lint-fix           run ruff to (mostly) fix lint issues"
```
hugovk, line 53:
```suggestion
	@echo "  lint-fix           run Ruff to (mostly) fix lint issues"
```

src/PIL/ImageFilter.py
```diff
@@ -205,7 +205,7 @@ class UnsharpMask(MultibandFilter):
 
     .. _digital unsharp masking: https://en.wikipedia.org/wiki/Unsharp_masking#Digital_unsharp_masking
 
-    """  # noqa: E501
```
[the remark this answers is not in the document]
akx: Long lines would be fixed by Black anyway, as far as it's able to. This is a long enough line that it can't be fixed by it.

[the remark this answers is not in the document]
hugovk: Sure, but Flake8 complains if this `noqa` is missing, as does Ruff 0.0.259, whereas Ruff 0.0.260 is fine without it. 

```console
❯ flake8 --version
6.0.0 (mccabe: 0.7.0, pycodestyle: 2.10.0, pyflakes: 3.0.1) CPython 3.11.2 on Darwin
❯ flake8 src/PIL/ImageFilter.py
src/PIL/ImageFilter.py:206:89: E501 line too long (102 > 88 characters)
```
```console
❯ ruff --version
ruff 0.0.259

❯ ruff src/PIL/ImageFilter.py
src/PIL/ImageFilter.py:206:89: E501 Line too long (102 > 88 characters)
Found 1 error.
```
```console
❯ ruff --version
ruff 0.0.260

❯ ruff src/PIL/ImageFilter.py
```

Why has Ruff changed?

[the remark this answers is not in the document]
hugovk: This is why: https://github.com/charliermarsh/ruff/pull/3663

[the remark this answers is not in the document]
akx: Yeah, I think that's a reasonable change in Ruff's behavior TBH :)

cclauss:
What is the blocker or todo on this pull request?  It has been open for 9 weeks.

hugovk:
Partly reviewer time, partly waiting for a "stable" release: https://github.com/charliermarsh/ruff/issues/1992

And hopefully a stable release will be soon given the funding: https://astral.sh/blog/announcing-astral-the-company-behind-ruff

We're not in a rush to merge this, the speedup is negligible here.

I do want this merged though, so thank you to both of you for the PR and answers to my questions :)

akx:
Rebased, added a commit to upgrade to Ruff 275, fixed another autofix. 

Ruff will probably still not be stable-stable for a while (despite funding) so it might be nice to get this merged so we could maybe look at enabling more lints afterwards? :)

hugovk:
Like ruff, we're not in a rush :)

.pre-commit-config.yaml
```diff
@@ -1,51 +1,41 @@
 repos:
+  - repo: https://github.com/charliermarsh/ruff-pre-commit
+    rev: v0.0.275
+    hooks:
+      - id: ruff
+        args:
+          - --fix
```
hugovk, line 7: Let's be consistent with the other hooks:

```suggestion
        args: [--fix]
```

And as suggested by https://github.com/astral-sh/ruff-pre-commit, should we exit with an error when it fixes, so that fixed commits aren't committed in the pre-fix state? 

```suggestion
        args: [--fix, --exit-non-zero-on-fix]
```

_custom_build/backend.py
```diff
@@ -1,6 +1,6 @@
 import sys
 
-from setuptools.build_meta import *  # noqa: F401, F403
```
hugovk, line 3: Why does Flake8 warn about F401 when Ruff doesn't?

hugovk:
That is helpful, thanks!

@akx Would you like to resolve the conflicts and update the config?

We've since removed `setup.cfg` (https://github.com/python-pillow/Pillow/pull/7484) and added some new pre-commit hooks including pyupgrade and flake-logging (https://github.com/python-pillow/Pillow/pull/7421).

akx:
@hugovk Done! I also have another branch that would stack on top of this to switch from `black` to [the Ruff formatter](https://astral.sh/blog/the-ruff-formatter) but that's a different concern so not muddling these waters with that.

akx:
The UP031 fixes would involve rewriting the `"something %dx%d" % some_tuple` style expressions in another way case by case - I'd leave them for an imminent follow up PR?

_custom_build/backend.py
akx: I suppose Flake8 warning about both F401 and F403 for a star import is overkill anyway.

akx:
@hugovk @radarhere Could this maybe be good to go now? 😁 I could follow up with a PR to remove the six UP031 `noqa`s next...

Makefile
```diff
@@ -118,6 +118,6 @@ lint:
 .PHONY: lint-fix
 lint-fix:
 	python3 -c "import black" > /dev/null 2>&1 || python3 -m pip install black
-	python3 -c "import isort" > /dev/null 2>&1 || python3 -m pip install isort
 	python3 -m black --target-version py38 .
-	python3 -m isort .
+	python3 -c "import ruff" > /dev/null 2>&1 || python3 -m pip install ruff
+	python3 -m ruff --preview --fix .
```
radarhere, line 123: If stability is a concern, then wouldn't it be better to not use `--preview`? I have to imagine that any features that were in preview back when this PR was created are no longer in preview.

radarhere:
I don't think I have strong opinions on the style changes seen here, so once @hugovk is ok with this, I'm happy for it to be merged.

Makefile
akx: Sans `--preview`:

```
warning: Selection `LOG` has no effect because the `--preview` flag was not included.
warning: Selection `LOG` has no effect because the `--preview` flag was not included.
warning: Selection `LOG` has no effect because the `--preview` flag was not included.
```

akx: (LOG was added via #7421.)

hugovk: I think we should go for stability and avoid `--preview`. Any idea when LOG will leave preview?

akx: I don't really think there's a big huge stability issue here, since whoever ends up updating the Ruff version in this repo will probably look at what they're doing?

Also, the more happy users a `preview` rule has, the better its chances are to graduate, I think 😁

hugovk: Let's remove `--preview`, I'd prefer to keep things stable and not spend time double checking potentially-unstable changes every month.

akx: Removed `--preview`, and also the `"LOG"` family of lints, because it would otherwise warn about it on every `--preview`less invocation.

hugovk (commented):
Please also delete `.flake8` (recently added in https://github.com/python-pillow/Pillow/pull/7484).

akx:
@hugovk Done, rebased, etc. :)

hugovk:
Thanks, a followup for UP031 would be good. And I'd prefer to keep Black rather than the Ruff formatter.

PR #6974, opened 2023-02-25
Added memoryview support to frombytes()

### LINKED ISSUE #6933: Can support for the `memoryview` type be added as input to _decode (decode.c)?

radarhere:
Resolves #6933. Alternative to #6937

Adds support for using `memoryview` objects in `frombytes()`, by changing
https://github.com/python-pillow/Pillow/blob/dbcd7372e554ee420a156b968eb9a609ec66ffd1/src/decode.c#L124
to use `y*` instead. This is actually the ["recommended way to accept binary data"](https://docs.python.org/3/c-api/arg.html#strings-and-buffers).

See https://bugs.python.org/issue30625 for a discussion about why this change allows `memoryview` objects to be accepted.

### DIFF
--- a/Tests/test_image_frombytes.py
+++ b/Tests/test_image_frombytes.py
@@ -1,10 +1,17 @@
+import pytest
+
 from PIL import Image
 
 from .helper import assert_image_equal, hopper
 
 
-def test_sanity():
+@pytest.mark.parametrize("data_type", ("bytes", "memoryview"))
+def test_sanity(data_type):
     im1 = hopper()
-    im2 = Image.frombytes(im1.mode, im1.size, im1.tobytes())
+
+    data = im1.tobytes()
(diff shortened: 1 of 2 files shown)


hugovk (approved):

Yay295:
I think this is missing a call to `PyBuffer_Release()`.

radarhere:
I've created PR #6976 to release the buffer.

radarhere:
#6976 has now been merged.

PR #6977, opened 2023-02-28
Converting from APNG to WebP caused an error about timestamp being a float

singpolyma:

 * cast timestamp to int in WebPImagePlugin so that APNG conversion works instead of crashing

### DIFF
--- a/src/PIL/WebPImagePlugin.py
+++ b/src/PIL/WebPImagePlugin.py
@@ -285,7 +285,7 @@ def _save_all(im, fp, filename):
                 # Append the frame to the animation encoder
                 enc.add(
                     frame.tobytes("raw", rawmode),
-                    timestamp,
+                    int(timestamp),
                     frame.size[0],
                     frame.size[1],
                     rawmode,
@@ -305,7 +305,7 @@ def _save_all(im, fp, filename):
         im.seek(cur_idx)
 
     # Force encoder to flush frames
-    enc.add(None, timestamp, 0, 0, "", lossless, quality, 0)
+    enc.add(None, int(timestamp), 0, 0, "", lossless, quality, 0)
 
     # Get the final output from the encoder
     data = enc.assemble(icc_profile, exif, xmp)


hugovk:
Please can you include a test case that fails without the fix and passes with it?

radarhere:
To demonstrate that this problem at least doesn't exist with all APNG files, using https://github.com/python-pillow/Pillow/blob/main/Tests/images/iss634.apng, the following code runs with Pillow 9.4.0 without an error.
```python
from PIL import Image
im = Image.open('iss634.apng')
im.save('out.webp')
```

radarhere:
if you feel uncomfortable writing a test case, just posting a self-contained example here would still be helpful.

singpolyma:
This WIP test includes an image from https://signalstickers.com/pack/2ef10d945db55e0712773e18b24febe0 so it may not be properly redistributable as part of your suite, but it reproduces the issue when my patch is not applied

radarhere:
Thanks. I see now that I missed the obvious fact that I needed to use `save_all` in my previous comment.

If I replace your test with
```python
def test_float_duration(tmp_path):
    temp_file = str(tmp_path / "temp.webp")
    with Image.open("Tests/images/iss634.apng") as im:
        assert im.info["duration"] == 70.0

        im.save(temp_file, save_all=True)

    with Image.open(temp_file) as reloaded:
        reloaded.load()
        assert reloaded.info["duration"] == 70
```
does that sound good to you? It uses one of our existing test images.

src/PIL/WebPImagePlugin.py
```diff
@@ -285,7 +285,7 @@ def _save_all(im, fp, filename):
                 # Append the frame to the animation encoder
                 enc.add(
                     frame.tobytes("raw", rawmode),
-                    timestamp,
+                    int(timestamp),
```
radarhere, line 288: What do you think about using `round` instead of `int`?

src/PIL/WebPImagePlugin.py
```diff
@@ -305,7 +305,7 @@ def _save_all(im, fp, filename):
         im.seek(cur_idx)
 
     # Force encoder to flush frames
-    enc.add(None, timestamp, 0, 0, "", lossless, quality, 0)
+    enc.add(None, int(timestamp), 0, 0, "", lossless, quality, 0)
```
radarhere, line 308:
```suggestion
    enc.add(None, round(timestamp), 0, 0, "", lossless, quality, 0)
```

src/PIL/WebPImagePlugin.py
radarhere, line 288:
```suggestion
                    round(timestamp),
```

radarhere:
I've created PR #6996 with my two suggestions.

hugovk:
Thanks for the PR, https://github.com/python-pillow/Pillow/pull/6996 has been merged instead.

PR #6978, opened 2023-02-28
Raise an error if ImageDraw co-ordinates are incorrectly ordered

radarhere:
Alternative to #6956

For ImageDraw methods that take `(x0, y0, x1, y1)`, raise an error if `x1` < `x0` or `y0` < `y1`.

I've also added this requirement to the docs for `rectangle()` and `rounded_rectangle()`.


hugovk:
`ValueError` instead of `TypeError`?

> Passing arguments of the wrong type (e.g. passing a [`list`](https://docs.python.org/3/library/stdtypes.html#list) when an [`int`](https://docs.python.org/3/library/functions.html#int) is expected) should result in a [`TypeError`](https://docs.python.org/3/library/exceptions.html?highlight=typeerror#TypeError), but passing arguments with the wrong value (e.g. a number outside expected boundaries) should result in a [`ValueError`](https://docs.python.org/3/library/exceptions.html?highlight=typeerror#ValueError).

https://docs.python.org/3/library/exceptions.html#TypeError

We're getting a tuple as expected, but of the wrong value order.

radarhere:
Sure, I've updated my last commit to use ValueError instead.

src/PIL/ImageDraw.py
```diff
@@ -303,6 +303,12 @@ def rounded_rectangle(
             (x0, y0), (x1, y1) = xy
         else:
             x0, y0, x1, y1 = xy
+        if x1 < x0 or y1 < y0:
+            msg = (
+                "x1 must be greater than or equal to x0,"
+                " and y1 must be greater than or equal to y0"
+            )
+            raise ValueError(msg)
```
hugovk, line 311: Shall we split this into two, so the messages can be more targeted and helpful?

For example:

```suggestion
        if x1 < x0:
            msg = (
                "x1 must be greater than or equal to x0"
            )
            raise ValueError(msg)
        if y1 < y0:
            msg = (
                "y1 must be greater than or equal to y0"
            )
            raise ValueError(msg)
```

And similarly in `_imaging.c`.

hugovk (approved):

wlritchi:
I think choosing this approach over just ordering the points isn't a great idea - or at least, not without a major version bump. There are plenty of applications using Pillow that rely on the old behaviour, and the fact that they don't sort their points doesn't necessarily imply that they've got a typo or a bug in their calculations. Maybe sorting would be the better option, with the ordering requirement scheduled for 10.x?

radarhere:
> the fact that they don't sort their points doesn't necessarily imply that they've got a typo or a bug in their calculations

My main thinking behind this change was [trying to help if they do have a bug](https://github.com/python-pillow/Pillow/pull/6956#issuecomment-1433878223).

I don't know exactly which method is a problem for you, but we've had notes in the ImageDraw documentation that ``x1 >= x0`` and ``y1 >= y0`` since 5.2.0 (#3080) for `arc()`, `chord()`, `ellipse()` and `pieslice()`. For the general case, we've had notes at https://pillow.readthedocs.io/en/latest/handbook/concepts.html#coordinate-system that
> Rectangles are represented as 4-tuples, with the upper left corner given first.

since 2.3.0, 9 years ago. So I don't think this is a sudden change.

If many users say that this is a problem, sure, we can revisit it, but otherwise, I'm reluctant to rollback decisions that we've made.

wlritchi:
Alright, fair enough - 9 years seems like plenty of notice to me. I guess [Hyrum's Law](https://www.hyrumslaw.com/) applies here.

I'll mention for context: the application where I encountered the problem is [Overviewer](https://github.com/overviewer/Minecraft-Overviewer), a Minecraft map viewer with 3.2k stars that recently became unmaintained (hence no official fix and no update to require `pillow<9.5.0`). The offending line is
```
ImageDraw.Draw(torch).rectangle((0,24,10,15),outline=(0,0,0,0),fill=(0,0,0,0))
```
which I suspect doesn't hit any of the snags mentioned in the earlier PR because it lacks rounding, outline, or fill.

PR #6984, opened 2023-03-03
Added "closed" attribute to images

### LINKED ISSUE #6982: Any performant way to determine if an image is closed?
The only way I'm aware of to determine if an image instance has been closed/finalized is to try to [load](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.load) the image and check if an exception is raised or not.

This results in the image being loaded if it hasn't been closed nor loaded already, which isn't good in terms of performance. On the other hand, exceptions being raised and handled isn't so performant in python.

I'm developing a library which defines a class that accepts PIL images upon instantiation and would like to check if the image has been closed at initialization of instances. The only other alternative is to accept the image as given and allow an exception raised from within Pillow to be propagated later on when the PIL image is to be used but I do not like this.

I suggest that a property be added to `PIL.Image.Image` to determine if an image has been closed without the possible cost of loading the image or raising and handling an exception. A boolean attribute set at initialization and updated upon closing the image should be enough to get this done.

I know this might not be of much importance but it would be good to at least have a Look-Before-You-Leap alternative.

Thanks.

radarhere:
Resolves #6982

Adds a `closed` attribute to images. It is initially `False`, and set to `True` after `im.close()` or after exiting the context manager.


AnonymouX47:
Umm... the only thing left would be documentation. 🤔

radarhere:
I've updated the commit to include documentation. You can preview it at https://pillow--6984.org.readthedocs.build/en/6984/reference/open_files.html#image-lifecycle

> By the way... Is there any particular reason why Pillow seems to prefer not to use properties (i.e the `property` built-in class)?

The `@property` decorator? I don't recall a discussion about this at any point. I don't know what scenarios you think we could be using it in but aren't. I will say that PIL existed since before the @property decorator, so our preference to be not change our API in a backwards incompatible way could be a factor.

AnonymouX47:
I see. Thanks

hugovk (commented):
No particular opposition to this, but wondering how much of a special case this addressing, and how common this sort of thing is in general.

radarhere:
I've mentioned how this could be done currently using `isinstance(im.im, DeferredError)` in the issue, so if this is seen as an edge case, then this PR can be closed.

PR #6986, opened 2023-03-03
Close file pointer copy in the libtiff encoder if still open

### LINKED ISSUE #6985: Too Many Open Files when working with Tiffs (issue with files being closed)
### What did you do?
I am using os.walk to pattern match files in folders, reducing the size of images using pillow and saving them to a temporary folder, eventually merging these to a PDF. I have narrowed the issue down to the Pillow compression as _Without the pillow compression_, the process completes for many hundreds of thousands of images but when using the pillow compression it stops consistently with [Errno 24] Too many open files on file 8187.

This happens in VS Code and an Anaconda Powershell on different computers in different python environments and in pillow 9.2.0 and 9.4.0.

Images would be opened, saved and closed using the context manager where conditions are met. This works until the script stops with [Errno 24] Too many open files.

### What actually happened?
The process stops at the same point every time (after successfully processing 8186 files) with [Errno 24] Too many open files.

pillow

I think this is similar to the issue shown in [https://github.com/python-pillow/Pillow/issues/5936](https://github.com/python-pillow/Pillow/issues/5936) where the tiff files are not being closed.

Unfortunately, I have many hundreds of thousands of files to process. I've tried all the suggestions on [https://github.com/python-pillow/Pillow/issues/5936](https://github.com/python-pillow/Pillow/issues/5936) Could something have re-introduced this issue?

* OS: Windows, tested on two separate computers (i7, 16GB ram and Xeon, 128GB ram) with different environments:
* Python: 3.8.12 and 3.9.12
* Pillow: 9.2.0 and 9.4.0

I have provided a working example using only the compression function.

### Working example
Code with an example 20kb image is included: [Pil.zip](https://github.com/python-pillow/Pillow/files/10882249/Pil.zip)

This throws the exception [Errno 24] Too many open files on file 8187 on both an i7, 16GB ram and a Xeon, 128GB ram.

```python
import os
import gc
from PIL import Image, TiffImagePlugin

TiffImagePlugin.DEBUG = True
Image.MAX_IMAGE_PIXELS=None

dir = os.path.dirname(__file__)

input_dir= os.path.join(dir +"/input/")
temp_dir= os.path.join(dir +"/temp/")
exceptionfile = './temp/exceptionlog.txt'


def compression():
    try:
        with open (os.path.join(dir +"/input/" + file), "rb") as f:
            img = Image.open(f)
            if img.mode == '1':
                img.save(temp_dir+file, compression='group4')
            elif img.mode == 'P' or img.mode == 'L':
                bitonal = img.convert('1')
                bitonal.save(temp_dir+file, compression='group4')
            elif img.mode == 'RGB':
                img.save(temp_dir+file+".jpg", compression='jpeg') 
            else: img.save(temp_dir+file+".jpg", compression='jpeg')
    except Exception as e:
        print(e)
        with open(exceptionfile, 'a') as f:
            print(f"{e}", file = f)
    finally:
         gc.collect()

test_image = ["0500_SI_0001.tif"]
         
if __name__ ==  "__main__":

    count = 0
    while True:
        for file in test_image:
            compression()
            count += 1
            print(count)
```

fcarron:
Resolves #6985

**Problem:**
When saving many tiff files with compression="group4" I get the error OSError: [Errno 24] Too many open files (after ~8000 images)

```python
def test_pil():
    im_pil = Image.open("T:/test.jpg")
    im_pil = im_pil.reduce(5)
    im_pil = im_pil.convert("1")
    for i in range(10000)
        file = f"{i:03}"
        out_name = os.path.join(
            "T:/trash", os.path.splitext(os.path.basename(file))[0] + ".tif")
        )
        im_pil.save(out_name, format="TIFF", compression="group4")
```

Without compression="group4" it works fine.

This also happens when a new image is created on each iteration and with explicit close()

```python
def test_pil():
    for i in range(10000):
        file = f"{i:03}"
        im_pil = Image.open("T:/test.jpg")
        im_pil = im_pil.reduce(5)
        im_pil = im_pil.convert("1")
        out_name = os.path.join(
            "T:/trash", os.path.splitext(os.path.basename(file))[0] + ".tif")
        )
        im_pil.save(out_name, format="TIFF", compression="group4")
        im_pil.close()
```

The file pointer seems to be cloned by the libtiff library, but not freed at the end.
Closing to file pointer solve the problem for me.

### DIFF
--- a/Tests/test_file_libtiff.py
+++ b/Tests/test_file_libtiff.py
@@ -1065,3 +1065,9 @@ class TestFileLibTiff(LibTiffTestCase):
         out = str(tmp_path / "temp.tif")
         with pytest.raises(SystemError):
             im.save(out, compression=compression)
+
+    def test_save_many_compressed(self, tmp_path):
+        im = hopper()
+        out = str(tmp_path / "temp.tif")
+        for _ in range(10000):
+            im.save(out, compression="jpeg")
--- a/src/PIL/TiffImagePlugin.py
+++ b/src/PIL/TiffImagePlugin.py
@@ -1850,6 +1850,11 @@ def _save(im, fp, filename):
                 fp.write(data)
             if errcode:
                 break
+        if _fp:
+            try:
+                os.close(_fp)
+            except OSError:
+                pass
         if errcode < 0:
             msg = f"encoder error {errcode} when writing image file"
             raise OSError(msg)


geoawd:
Hi @fcarron - I opened [https://github.com/python-pillow/Pillow/issues/6985#issue-1608688767](https://github.com/python-pillow/Pillow/issues/6985#issue-1608688767) 2 minutes before you posted this merge request. I've made the same changes and this seems to have resolved the issue in [#6985](https://github.com/python-pillow/Pillow/issues/6985#issue-1608688767) - in the test files I had provided as well. I can test this properly next week but thanks for submitting this.

radarhere:
I presume that like the other reports we've received about this, that you're a Windows user?

radarhere (approved):

fcarron:
@radarhere Yes, exactly (sorry I forgot to mention that),
My setup:
Microsoft Windows 11 Pro, 10.0.22621 Build 22621
Python 3.11.2
Pillow version 9.4.0
But I first had this issue on an old Windows 7 machine with Python 3.8.6 and Pillow 9.4.0.

radarhere:
Thanks.

I've created https://github.com/fcarron/Pillow/pull/1 as a suggestion for a test to add, to help ensure that this change doesn't get broken again in the future.

geoawd:
Thanks @fcarron for this pull request which has also resolved #6985. I’ve now tested this with a couple of sets of ~40k files and they have completed without issue.

fcarron:
@radarhere Thank you for adding this test.

radarhere:
Just to link, this PR is similar to #5946

PR #6987, opened 2023-03-04
Removed unused profile_fromstring method

radarhere:
While ImageCms calls `profile_frombytes`,
https://github.com/python-pillow/Pillow/blob/1457d2c14667d9901d7bc2552beed4a1ab81742b/src/PIL/ImageCms.py#L187-L191
`profile_fromstring` is unused.

`profile_fromstring` actually calls the same code as `profile_frombytes`.

https://github.com/python-pillow/Pillow/blob/1457d2c14667d9901d7bc2552beed4a1ab81742b/src/_imagingcms.c#L963-L964

So this PR removes the `profile_fromstring` method declaration, and renames the C method to `cms_profile_frombytes` for clarity.

### DIFF
--- a/src/_imagingcms.c
+++ b/src/_imagingcms.c
@@ -116,7 +116,7 @@ cms_profile_open(PyObject *self, PyObject *args) {
 }
 
 static PyObject *
-cms_profile_fromstring(PyObject *self, PyObject *args) {
+cms_profile_frombytes(PyObject *self, PyObject *args) {
     cmsHPROFILE hProfile;
 
     char *pProfile;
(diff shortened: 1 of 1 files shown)


etienned:
I just want to let you know that this change broke my code (written a long time ago) because I was using the `profile_fromstring` (maybe this is considered a private method?) to create a profile from the 'icc_profile' info of the image. I understand that using the `profile_frombytes` is the right thing to do, because the 'icc_profile' is already in bytes, no problem there, but I would like to know if this method is considered private and, if it's the case, what's the proper way to create a profile from the 'icc_profile' info? If the method is not considered private, maybe you should have put a deprecation before removing it. Thanks

hugovk:
Sorry about that, how were you calling it?

Pillow's C API is internal and private, and can and will change between versions without deprecation warnings.

etienned:
Thanks for the fast reply! As it's private, no need to be sorry, this is just normal.

I was calling it like this:
```python
profile_data = img.info.get('icc_profile')
if profile_data:
    profile = ImageCms.core.profile_fromstring(profile_data)
```

I'm still calling `profile_frombytes` the same way. Is there a way to get a profile from the image info 'icc_profile' without using private methods?

radarhere:
> After reflexion, I should probably do this:
> 
> ```python
> profile_data = img.info.get('icc_profile')
> if profile_data:
>     profile_file = io.BytesIO(profile_data)
>     profile = ImageCms.getOpenProfile(profile_file)
> ```

Sounds like a plan to me. Although for the record, to better match your first code block, you'd want to do `ImageCms.getOpenProfile(profile_file).profile`.

PR #6988, opened 2023-03-04
Added memoryview support to Dib.frombytes()

### LINKED ISSUE #6975: Use `y*` instead of `y#`
#6974 changed https://github.com/python-pillow/Pillow/blob/dbcd7372e554ee420a156b968eb9a609ec66ffd1/src/decode.c#L124 to use `y*` instead of `y#` so that it can accept a `memoryview` object, and notes that `y*` is actually the recommended way to accept binary data. `y#` is used in a few other locations; should these be changed as well?

https://github.com/python-pillow/Pillow/blob/7d8a08b2211abdfc5766e01c0444d3ae81843439/src/display.c#L201
https://github.com/python-pillow/Pillow/blob/7d8a08b2211abdfc5766e01c0444d3ae81843439/src/display.c#L721
https://github.com/python-pillow/Pillow/blob/7d8a08b2211abdfc5766e01c0444d3ae81843439/src/encode.c#L567
https://github.com/python-pillow/Pillow/blob/7d8a08b2211abdfc5766e01c0444d3ae81843439/src/encode.c#L1060
https://github.com/python-pillow/Pillow/blob/7d8a08b2211abdfc5766e01c0444d3ae81843439/src/_imaging.c#L1644
https://github.com/python-pillow/Pillow/blob/7d8a08b2211abdfc5766e01c0444d3ae81843439/src/_imaging.c#L1709
https://github.com/python-pillow/Pillow/blob/7d8a08b2211abdfc5766e01c0444d3ae81843439/src/_imaging.c#L2220
https://github.com/python-pillow/Pillow/blob/7d8a08b2211abdfc5766e01c0444d3ae81843439/src/_imaging.c#L2544
https://github.com/python-pillow/Pillow/blob/7d8a08b2211abdfc5766e01c0444d3ae81843439/src/_imagingcms.c#L124
https://github.com/python-pillow/Pillow/blob/7d8a08b2211abdfc5766e01c0444d3ae81843439/src/_imagingft.c#L142
https://github.com/python-pillow/Pillow/blob/7d8a08b2211abdfc5766e01c0444d3ae81843439/src/_imagingft.c#L1099
https://github.com/python-pillow/Pillow/blob/7d8a08b2211abdfc5766e01c0444d3ae81843439/src/_imagingft.c#L1145
https://github.com/python-pillow/Pillow/blob/7d8a08b2211abdfc5766e01c0444d3ae81843439/src/_webp.c#L599

radarhere:
Resolves #6975

Allows `memoryview` objects to be passed as the `buffer` argument to `Dib.frombytes()`. Similar to #6974

### DIFF
--- a/src/display.c
+++ b/src/display.c
@@ -202,6 +202,7 @@ _frombytes(ImagingDisplayObject *display, PyObject *args) {
     }
 
     if (display->dib->ysize * display->dib->linesize != buffer.len) {
+        PyBuffer_Release(&buffer);
         PyErr_SetString(PyExc_ValueError, "wrong size");
         return NULL;
     }
--- a/src/display.c
+++ b/src/display.c
@@ -195,20 +195,20 @@ _releasedc(ImagingDisplayObject *display, PyObject *args) {
 
 static PyObject *
 _frombytes(ImagingDisplayObject *display, PyObject *args) {
-    char *ptr;
-    Py_ssize_t bytes;
+    Py_buffer buffer;
 
-    if (!PyArg_ParseTuple(args, "y#:frombytes", &ptr, &bytes)) {
+    if (!PyArg_ParseTuple(args, "y*:frombytes", &buffer)) {
         return NULL;
     }
 
-    if (display->dib->ysize * display->dib->linesize != bytes) {
+    if (display->dib->ysize * display->dib->linesize != buffer.len) {
         PyErr_SetString(PyExc_ValueError, "wrong size");
         return NULL;
     }
 
-    memcpy(display->dib->bits, ptr, bytes);
+    memcpy(display->dib->bits, buffer.buf, buffer.len);
 
+    PyBuffer_Release(&buffer);
     Py_INCREF(Py_None);
     return Py_None;
 }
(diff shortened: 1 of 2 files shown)


src/display.c
```diff
@@ -195,20 +195,20 @@ _releasedc(ImagingDisplayObject *display, PyObject *args) {
...
-    Py_ssize_t bytes;
+    Py_buffer buffer;
 
-    if (!PyArg_ParseTuple(args, "y#:frombytes", &ptr, &bytes)) {
+    if (!PyArg_ParseTuple(args, "y*:frombytes", &buffer)) {
         return NULL;
     }
 
-    if (display->dib->ysize * display->dib->linesize != bytes) {
+    if (display->dib->ysize * display->dib->linesize != buffer.len) {
```
nulano, line 204:
```suggestion
    if (display->dib->ysize * display->dib->linesize != buffer.len) {
        PyBuffer_Release(&buffer);
```

I think the buffer should be released in the error branch also?

radarhere: Yep, thanks

PR #6991, opened 2023-03-04
Build docs in own workflow

hugovk:
So far the `make doccheck` docs test has been piggybacking on the main `test.yml` build.

We have a pretty big (and slow!) CI matrix, I think it's time to split the it into its own workflow.

This way, we can run the docs workflow only when the docs change (or the docs workflow file), and have the other workflows ignore docs-only changes.

It will make it much quicker for docs-only PRs.

For example, the first commit here still ran all the workflows:

image

And the second only ran the docs (plus linting):

image

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,50 @@
+name: Docs
+
+on:
+  push:
+    paths:
+      - ".github/workflows/docs.yml"
+      - "docs/**"
+  pull_request:
+    paths:
+      - ".github/workflows/docs.yml"
+      - "docs/**"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    name: Docs
+
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: "3.x"
+        cache: pip
+        cache-dependency-path: ".ci/*.sh"
+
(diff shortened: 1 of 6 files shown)


.github/workflows/docs.yml
```diff
@@ -0,0 +1,50 @@
+name: Docs
+
+on:
+  push:
+    paths:
+      - ".github/workflows/docs.yml"
+      - "docs/**"
+  pull_request:
+    paths:
+      - ".github/workflows/docs.yml"
```
radarhere, line 10: Ok, so you're also running the Docs workflow if the GitHub Actions file changes, fair enough. In that case, shouldn't the same logic also apply to the Valgrind workflow? https://github.com/hugovk/Pillow/pull/95

hugovk: Good idea, merged!

.github/workflows/docs.yml
```diff
@@ -0,0 +1,50 @@
...
+        python-version: "3.x"
+        cache: pip
+        cache-dependency-path: ".ci/*.sh"
+
+    - name: Build system information
+      run: python3 .github/workflows/system-info.py
+
+    - name: Install Linux dependencies
+      run: |
+        .ci/install.sh
```
radarhere, line 42:
```suggestion
        .ci/install.sh
      env:
        GHA_PYTHON_VERSION: "3.x"
```
I know it's not a specific Python version, but for the purposes of
https://github.com/python-pillow/Pillow/blob/eccc1e9afb4b2b012772abdad026e1efa49b0e9f/.ci/install.sh#L41-L44
it is sufficient.

hugovk: I left it out because I don't think we specifically need PyQt for the docs build? Well, let's use the env var so as not to worry about it in the future. 👍

PR #6999, opened 2023-03-10
Handle failure from PyDict_New or PyList_New

### LINKED ISSUE #6322: Memory leaks when function returns directly on failure without decreasing the refcnt of a new reference (static analyzer reports)
New reference is returned here:
https://github.com/python-pillow/Pillow/blob/68e39cba431680105c0e0f1a54371b7028e8da9a/src/_imaging.c#L4253
Function returns without decreasing the refcnt.
https://github.com/python-pillow/Pillow/blob/68e39cba431680105c0e0f1a54371b7028e8da9a/src/_imaging.c#L4256
Internal Report ID: 3fa97f

---
New reference is returned here:
https://github.com/python-pillow/Pillow/blob/68e39cba431680105c0e0f1a54371b7028e8da9a/src/_imagingft.c#L1071
Function returns without decreasing the refcnt.
https://github.com/python-pillow/Pillow/blob/68e39cba431680105c0e0f1a54371b7028e8da9a/src/_imagingft.c#L1077
Internal Report ID: 60bc38

---
New reference is returned here: 
https://github.com/python-pillow/Pillow/blob/68e39cba431680105c0e0f1a54371b7028e8da9a/src/_imagingft.c#L1114
Function returns without decreasing the refcnt.
https://github.com/python-pillow/Pillow/blob/68e39cba431680105c0e0f1a54371b7028e8da9a/src/_imagingft.c#L1128
Internal Report ID: 775d14

---
New reference is returned here:
https://github.com/python-pillow/Pillow/blob/68e39cba431680105c0e0f1a54371b7028e8da9a/src/_imagingft.c#L1118
Function returns without decreasing the refcnt.
https://github.com/python-pillow/Pillow/blob/68e39cba431680105c0e0f1a54371b7028e8da9a/src/_imagingft.c#L1128
Internal Report ID: 379efb

---
New reference is returned here: 
https://github.com/python-pillow/Pillow/blob/68e39cba431680105c0e0f1a54371b7028e8da9a/src/_imagingmorph.c#L138
Function returns without decreasing the refcnt.
https://github.com/python-pillow/Pillow/blob/68e39cba431680105c0e0f1a54371b7028e8da9a/src/_imagingmorph.c#L142
Internal Report ID: 8cc0d2

---
New reference is returned here:
https://github.com/python-pillow/Pillow/blob/68e39cba431680105c0e0f1a54371b7028e8da9a/src/_imagingmorph.c#L215
Function returns without decreasing the refcnt.
https://github.com/python-pillow/Pillow/blob/68e39cba431680105c0e0f1a54371b7028e8da9a/src/_imagingmorph.c#L220
Internal Report ID: 239db5

---
New reference is returned here: 
https://github.com/python-pillow/Pillow/blob/68e39cba431680105c0e0f1a54371b7028e8da9a/src/_imagingtk.c#L77
Function returns without decreasing the refcnt (when `?:` operator takes the false branch).
https://github.com/python-pillow/Pillow/blob/68e39cba431680105c0e0f1a54371b7028e8da9a/src/_imagingtk.c#L78
Internal Report ID: bc9c58

---
New reference is returned here:
https://github.com/python-pillow/Pillow/blob/68e39cba431680105c0e0f1a54371b7028e8da9a/src/_webp.c#L984 
Function returns without decreasing the refcnt.
https://github.com/python-pillow/Pillow/blob/68e39cba431680105c0e0f1a54371b7028e8da9a/src/_webp.c#L986
Internal Report ID: e4ed93

radarhere:
Two changes here.
1. [`PyDict_New()`](https://docs.python.org/3/c-api/dict.html#c.PyDict_New) and [`PyList_New()`](https://docs.python.org/3/c-api/list.html#c.PyList_New) may return `NULL`. This addresses situations where that is not handled.
2. Resolves #6322, by decrementing reference counts. The first change adds another situation where this is needed, so these are in the same PR.


src/_imagingft.c
```diff
@@ -1125,10 +1129,21 @@ font_getvaraxes(FontObject *self) {
...
     for (i = 0; i < num_axis; i++) {
         axis = master->axis[i];
 
         list_axis = PyDict_New();
+        if (list_axis == NULL) {
+            for (j = 0; j < i; j++) {
+                list_axis = PyList_GetItem(list_axes, j);
+                Py_DECREF(list_axis);
+            }
+            Py_DECREF(list_axes);
```
akx, line 1144: Feels like this duplicated cleaning code should be deduplicated, maybe..?

src/_imagingtk.c
```diff
@@ -58,5 +58,9 @@ PyInit__imagingtk(void) {
     };
     PyObject *m;
     m = PyModule_Create(&module_def);
-    return (load_tkinter_funcs() == 0) ? m : NULL;
+    if (load_tkinter_funcs() != 0) {
+        Py_DECREF(m);
+        return NULL;;
```
akx, line 63:
```suggestion
        return NULL;
```

src/_imagingft.c
```diff
@@ -1082,11 +1082,15 @@ font_getvarnames(FontObject *self) {
 
     num_namedstyles = master->num_namedstyles;
     list_names = PyList_New(num_namedstyles);
+    if (list_names == NULL) {
+        return NULL;
```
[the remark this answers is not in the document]
radarhere: My thinking was to copy behaviour is already used in Pillow in another location.
https://github.com/python-pillow/Pillow/blob/2b164943a3ab33e2ac515e95865fda08a130d95a/src/_imaging.c#L3819-L3822

Looking at CPython, I think that `PyList_New` already [sets an error before returning](https://github.com/python/cpython/blob/5e6661bce968173fa45b74fa2111098645ff609c/Objects/listobject.c#L154-L162). The same with `PyDict_New`.

src/_imagingft.c
radarhere: Ok, I've pushed a commit to rearrange the code.

src/_imagingft.c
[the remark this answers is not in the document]
akx: Ah, great, so it seems 👍 

(I'd personally probably add a comment that says something like `// Error set by PyDict_New` on the return line, but that's fine.)

src/_imagingft.c
hugovk: @akx How's it look now?

hugovk (commented):
Some of these new `return NULL`s aren't covered by tests. I guess it's not very easy to trigger them, and that's fine as defensive programming in this case.

src/_imaging.c
```diff
@@ -1249,6 +1249,9 @@ _histogram(ImagingObject *self, PyObject *args) {
 
     /* Build an integer list containing the histogram */
     list = PyList_New(h->bands * 256);
+    if (list == NULL) {
+        return NULL;
```
nulano, line 1253:
```suggestion
        ImagingHistogramDelete(h);
        return NULL;
```

src/_imaging.c
```diff
@@ -2154,6 +2157,9 @@ _getcolors(ImagingObject *self, PyObject *args) {
         Py_INCREF(out);
     } else {
         out = PyList_New(colors);
+        if (out == NULL) {
+            return NULL;
```
nulano, line 2161:
```suggestion
            free(items);
            return NULL;
```

src/_imagingft.c
nulano: This loop looks suspicious to me. `PyList_GetItem` [returns a borrowed reference](https://docs.python.org/3/c-api/list.html#c.PyList_GetItem), so the decref will cause it to be freed. However, the list is expected to also free its contents when it's reference count is decrefed later: https://github.com/python/cpython/blob/e375bff03736f809fbc234010c087ef9d7e0d384/Objects/listobject.c#L356

So I would expect this loop to cause each `list_axis` to be double freed (except the last, which has not been added to the list yet).

radarhere: Thanks. I've applied the suggestion from @nulano, which means that the loop was de-duplicated is no longer present. So I've actually undone the rearranging of the code as it is no longer necessary.

src/_imagingft.c
nulano, line 1086:
```suggestion
        FT_Done_MM_Var(library, master);
        return NULL;
```

src/_imagingft.c
```diff
@@ -1082,11 +1082,15 @@ font_getvarnames(FontObject *self) {
...
+    }
 
     name_count = FT_Get_Sfnt_Name_Count(self->face);
     for (i = 0; i < name_count; i++) {
         error = FT_Get_Sfnt_Name(self->face, i, &name);
         if (error) {
+            Py_DECREF(list_names);
```
nulano, line 1093:
```suggestion
            Py_DECREF(list_names);
            FT_Done_MM_Var(library, master);
```

src/_imagingft.c
```diff
@@ -1125,10 +1129,17 @@ font_getvaraxes(FontObject *self) {
     name_count = FT_Get_Sfnt_Name_Count(self->face);
 
     list_axes = PyList_New(num_axis);
+    if (list_axes == NULL) {
+        return NULL;
```
nulano, line 1133:
```suggestion
        FT_Done_MM_Var(library, master);
        return NULL;
```

src/_imagingft.c
```diff
@@ -1125,10 +1129,17 @@ font_getvaraxes(FontObject *self) {
...
+    }
     for (i = 0; i < num_axis; i++) {
         axis = master->axis[i];
 
         list_axis = PyDict_New();
+        if (list_axis == NULL) {
+            Py_DECREF(list_axes);
```
nulano, line 1140:
```suggestion
            Py_DECREF(list_axes);
            FT_Done_MM_Var(library, master);
```

src/_imagingft.c
```diff
@@ -1144,6 +1155,8 @@ font_getvaraxes(FontObject *self) {
         for (j = 0; j < name_count; j++) {
             error = FT_Get_Sfnt_Name(self->face, j, &name);
             if (error) {
+                Py_DECREF(list_axis);
+                Py_DECREF(list_axes);
```
nulano, line 1159:
```suggestion
                Py_DECREF(list_axes);
                FT_Done_MM_Var(library, master);
```

nulano (approved):

PR #7004, opened 2023-03-11
Test build_prepare.py with multiple Python versions in AppVeyor

radarhere:
At the moment, AppVeyor always uses Python 3.7 when running `build_prepare.py`.
https://github.com/python-pillow/Pillow/blob/6b71090828a37340a4e552ad24602deb71c119aa/.appveyor.yml#L34

This PR changes it to use the Python version from the matrix, so that we can test it more thoroughly.

### DIFF
--- a/winbuild/build.rst
+++ b/winbuild/build.rst
@@ -100,7 +100,7 @@ The following is a simplified version of the script used on AppVeyor::
 
     set PYTHON=C:\Python38\bin
     cd /D C:\Pillow\winbuild
-    C:\Python37\bin\python.exe build_prepare.py -v --depends=C:\pillow-depends
+    %PYTHON%\python.exe build_prepare.py -v --depends=C:\pillow-depends
     build\build_dep_all.cmd
     build\build_pillow.cmd install
     cd ..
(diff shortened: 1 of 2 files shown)


hugovk (approved):
cc @nulano

winbuild/build.rst
```diff
@@ -100,7 +100,7 @@ The following is a simplified version of the script used on AppVeyor::
 
     set PYTHON=C:\Python38\bin
     cd /D C:\Pillow\winbuild
-    C:\Python37\bin\python.exe build_prepare.py -v --depends=C:\pillow-depends
```
nulano, line 103: I would keep this as is to demonstrate that a different Python version can be used to generate the build scripts.

nulano (commented):
Seems fine, but I don't see the benefit? Both versions are already tested on GHA. By using a version different from `%PYTHON%`, the original instead tests using a different Python version for generating build scripts vs building Pillow.

radarhere:
I hadn't imagined that anyone would intentionally use multiple versions of Python for this - but if you think that might be case, then ok.

I suppose one AppVeyor tests using the same Python version for both, and one tests using different Python versions.

nulano:
> I suppose one AppVeyor tests using the same Python version for both, and one tests using different Python versions.

Actually, both AppVeyor bulids are testing a different Python version:
The matrix is `C:/Python311`, `C:/Python37-x64`, while the build script is generated with `C:\Python37`.

It's only the GHA jobs that use the same version, where instead of using environment variables, the target is implicit from the `setup-python` step.

But yes, I think it could be useful in some circumstances to use a different target.

Issue #6769, opened 2022-11-30
Endianness is inconsistent


Yay295:
Image class methods:

| Method        | Parameter      | Endianness  |
|---------------|----------------|-------------|
| blend         | alpha          | native      |
| new           | color          | native      |
| frombytes     | data           | mode        |
| frombuffer    | data           | mode        |
| getcolors     |                | native      |
| getdata all   |                | mode        |
| getdata band  |                | native      |
| getextrema    |                | native      |
| getpalette    |                | mode        |
| getpixel      |                | native      |
| point         |                | unsupported |
| putdata       | data           | native      |
| putpalette    | data           | mode        |
| putpixel      | value          | native      |
| remap_palette | source_palette | mode        |
| tobytes       |                | mode        |

Methods that use one of the above methods:
`eval` calls `point`
`fromarray` calls `frombuffer`
`rotate` calls `transform`
`transform` calls `new`

Related: #2228

radarhere:
Could you provide some more detail about how you worked out this list of endianness behaviour, perhaps some Python code to demonstrate the inconsistency?

I understand that 'native' means that the endianness of the environment is used, but what do you mean by 'mode'?

Yay295:
I worked these out by looking at the code, though it seems I got a few wrong, and some of them have mode checks that I missed. By 'mode' I mean it accepts/returns data in the endianness of the image's mode.

```python
from PIL import Image

def t(method, *args, **kwargs):
    print(method.__name__ + ' ', end='')
    try:
        print(method(*args, **kwargs))
    except Exception as e:
        print(e)

def callMethods(image):
    t(x.getcolors)
    t(lambda:list(x.getdata()))
    t(lambda:list(x.getdata(0)))
    t(x.getextrema)
    t(x.getpixel,(0,0))
    t(x.tobytes)

for mode in ('I;16L','I;16B'):
    print(f'\n{mode} New 123')
    with Image.new(mode, (1,1), color=123) as x:
        callMethods(x)
    
    print(f'\n{mode} From Bytes \\x00{{')
    with Image.frombytes(mode, (1,1), data=b'\x00{') as x:
        callMethods(x)
    
    print(f'\n{mode} From Bytes {{\\x00')
    with Image.frombytes(mode, (1,1), data=b'{\x00') as x:
        callMethods(x)
```

```
I;16L New 123
getcolors image has wrong mode
<lambda> [123]
<lambda> image has wrong mode
getextrema image has wrong mode
[...]

PR #7010, opened 2023-03-13
Support creating BGR;15, BGR;16 and BGR;24 images, but drop support for BGR;32

radarhere:
https://github.com/python-pillow/Pillow/issues/6769#issuecomment-1364769723 mentioned that creating BGR images does not work. `Image.new("BGR;15", (1, 1))` does not work, and neither does BGR;16, BGR;24 or BGR;32.

So what are these modes?

https://github.com/python-pillow/Pillow/blob/3b4b77bb1192388634badf8e12a8e78161c8e5a6/src/libImaging/Unpack.c#L691-L693

https://github.com/python-pillow/Pillow/blob/3b4b77bb1192388634badf8e12a8e78161c8e5a6/src/libImaging/Unpack.c#L736-L738

So with that pattern, BGR;24 would allow 8 bits per channel, the same size as an RGB channel. And yes, when converting, the values are just reversed.

https://github.com/python-pillow/Pillow/blob/3b4b77bb1192388634badf8e12a8e78161c8e5a6/src/libImaging/Convert.c#L285-L291

So this PR adds support for creating images with those three image modes.

Continuing on, with that pattern, one could imagine that BGR;32 is a mode that supports 10 or 11-bit values for each channel.

However, there isn't any concrete information about that. Pillow doesn't currently support converting images to BGR;32, nor does it support any kind of unpacking with this mode.

In fact, the only code about this mode is
https://github.com/python-pillow/Pillow/blob/079caf671120894c6cd0e1d9dc8429970d023847/src/libImaging/Storage.c#L155-L161
which I think can't be accessed through our Python API without other support for the mode, and
https://github.com/python-pillow/Pillow/blob/079caf671120894c6cd0e1d9dc8429970d023847/src/PIL/ImageMode.py#L64
which doesn't offer any value to the user if nothing else can be done with this mode.

So if ImageMode has the only functionality that would change for users (and it has only been in there since #5935), I'm going to suggest just dropping support for BGR;32 without deprecation - there isn't any demonstrated need for BGR;32, and we would just be inventing a mode by trying to specify what it is.


Tests/test_image_access.py
```diff
@@ -363,8 +363,8 @@ def test_p_putpixel_rgb_rgba(self, mode):
 
 
 class TestImagePutPixelError(AccessTest):
-    IMAGE_MODES1 = ["L", "LA", "RGB", "RGBA"]
-    IMAGE_MODES2 = ["I", "I;16", "BGR;15"]
+    IMAGE_MODES1 = ["LA", "RGB", "RGBA"]
+    IMAGE_MODES2 = ["L", "I", "I;16"]
```
radarhere, line 367: Later in this file, `IMAGE_MODES2` are tested to raise "color must be int or single-element tuple", while `IMAGE_MODES1` are tested to raise "color must be int or tuple". This makes sense, as all `IMAGE_MODES2` are single channel, while `IMAGE_MODES1` have multiple channels.

Tests/test_image_access.py
```diff
@@ -407,11 +407,6 @@ def test_putpixel_overflow_error(self, mode):
         with pytest.raises(OverflowError):
             im.putpixel((0, 0), 2**80)
 
-    def test_putpixel_unrecognized_mode(self):
-        im = hopper("BGR;15")
-        with pytest.raises(ValueError, match="unrecognized image mode"):
-            im.putpixel((0, 0), 0)
```
radarhere, line 413: With the changes in this PR, you can no longer convert to an image mode that you can't create directly.

hugovk:
Let's also include this in release notes.

PR #7021, opened 2023-03-17
Do not render text if image has zero width or height

radarhere:
In [`FreeTypeFont.getmask2()`](https://pillow.readthedocs.io/en/stable/reference/ImageFont.html#PIL.ImageFont.FreeTypeFont.getmask2), only spend time rendering text if the image it will be rendered onto has a non-zero width or height.

### DIFF
--- a/src/PIL/ImageFont.py
+++ b/src/PIL/ImageFont.py
@@ -767,18 +767,19 @@ class FreeTypeFont:
         offset = offset[0] - stroke_width, offset[1] - stroke_width
         Image._decompression_bomb_check(size)
         im = fill("RGBA" if mode == "RGBA" else "L", size, 0)
-        self.font.render(
-            text,
-            im.id,
-            mode,
-            direction,
-            features,
-            language,
-            stroke_width,
-            ink,
-            start[0],
-            start[1],
-        )
+        if min(size):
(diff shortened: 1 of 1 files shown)


hugovk:
Seems fine, but I wonder how often someone will draw on an image with non-zero width/height, and how much time this would actually save them?

radarhere:
`getmask2()` isn't actually rendering onto a user-supplied image. it is creating an image from the dimensions returned by `font.getsize`
https://github.com/python-pillow/Pillow/blob/80edcd18d162793d1e4dc7a87952bdedcee2ee5a/src/PIL/ImageFont.py#L761-L769

nulano:
Sure, but this would only happen if a user tries to render a string containing spaces only, right?

In which case I'd argue that might be up to the user to prevent, as that doesn't seem like something that would be commonly done.

radarhere:
Aside from performance motivations, I was primarily thinking about https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=48708

Locally, #6842 resolved the problem with memory usage, but apparently oss-fuzz itself doesn't think it is fixed yet, so this is a second attempt. The font in the issue can trigger an `OSError: raster overflow`. This change avoids that.

PR #1177, opened 2015-04-04
Multiline text

radarhere:
My contribution to #739, building on the work of #582

I recognise that there is some discussion about it, so consider this just moving the conversation forward.

PR #7025, opened 2023-03-18
Removed incorrect docstring

Auto-5:


radarhere:
A combination of https://github.com/python-pillow/Pillow/pull/1177/commits/77169b2fdba9546c9f6f3842a48c7ba2ee370b16 and https://github.com/python-pillow/Pillow/pull/1926/commits/934485269c4286e9e890597877d73e15a091eb22 meant that the docstring should have been in `text()` instead.

I've created #7027 to bring it back to that location.

PR #7028, opened 2023-03-19
Add pillow logo

seriousadi:
old pillow logo was not working so i added pillow logo from official pillow site

Fixes #
The pillow logo wasn't working in readme.md 

- replaced the broken link to new working one from python-pillow.org

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 <p align="center">
-    <img width="248" height="250" src="https://raw.githubusercontent.com/python-pillow/pillow-logo/main/pillow-logo-248x250.png" alt="Pillow logo">
+    <img width="248" height="250" src="https://python-pillow.org/images/pillow-logo-light-text-1280x640.png" alt="Pillow logo">
 </p>
 
 # Pillow


radarhere:
You say that it wasn't working, and that it was a broken link.

However, https://raw.githubusercontent.com/python-pillow/pillow-logo/main/pillow-logo-248x250.png works for me if I navigate to it directly.

pillow-logo-248x250

If I look at the README, I see
readme

Are you seeing something different? Or were you thinking that the word 'Pillow' should be part of the logo, as you'd made it in this PR?

seriousadi:
It wasn't working on my laptop i even tried to go i that link but it wasn't working i tried it in my other devices too but it wasn't working then i thought i should add if the existing one isn't working so i added it from your pillow site but now it is working fine as if nothing happened anyways adding pillow to side actually  looks better but not with that pick it looks a bit squished.

hugovk:
Here's how it currently looks fore on mobile:

Screenshot_20230319_135644_Chrome

And with this PR:

Screenshot_20230319_135701_Chrome

seriousadi:
So if i edit that photo to look good on computers and mobile too will it work, is that ok

hugovk:
Please can you share a screenshot or explain how the existing one "wasn't working"?

hugovk:
It sounds like a temporary network issue causing it to fail to load the image. Thanks for the PR, but I don't think we need to replace the image for this.

seriousadi:
thanks no problem its ok

PR #7050, opened 2023-03-31
Fix warning that variable may be uninitialized

radarhere:
https://github.com/python-pillow/Pillow/actions/runs/4571913703/jobs/8070643875#step:7:66
>   src/_imaging.c: In function ‘getink’:
>  src/_imaging.c:555:44: warning: ‘tupleSize’ may be used uninitialized in this function [-Wmaybe-uninitialized]
>    555 |                         if (tupleSize != 1 && tupleSize != 2) {
>        |                             ~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~

This variable was added in #7010.

### DIFF
--- a/src/_imaging.c
+++ b/src/_imaging.c
@@ -502,7 +502,7 @@ getink(PyObject *color, Imaging im, char *ink) {
        be cast to either UINT8 or INT32 */
 
     int rIsInt = 0;
-    int tupleSize;
+    int tupleSize = 0;
     if (PyTuple_Check(color)) {
         tupleSize = PyTuple_GET_SIZE(color);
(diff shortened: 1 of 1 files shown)


radarhere:
Compiling on my local machine, I see
> src/libImaging/Quant.c:344:17: warning: variable 'nLeft' set but not used [-Wunused-but-set-variable]
>     int nRight, nLeft;
>                 ^
> 1 warning generated.

This is because nLeft is only used if NO_OUTPUT is defined. I've added a commit to also only set nLeft if NO_OUTPUT is defined.

hugovk (approved):

Issue #7008, opened 2023-03-12
Cannot save TIFF files with clipping info


dorphalsig:
I am opening a TIFF file with clipping path data (see https://github.com/python-pillow/Pillow/issues/6025), I now want to save it with another name (nothing else was done to it)

```python
from PIL import Image
img = Image.open('images/path test.tif')
ex = ""
try:
    img.save("path test public.tif")
except Exception as e:
    ex=e
```

error message is "path test public.tif: Failed to allocate memory for custom tag binary object (-1578999712 elements of 1 bytes each)."

when I print the exception I see: RuntimeError('Error setting from dictionary')
file is attached inside a zip
[Path test.zip](https://github.com/python-pillow/Pillow/files/10950847/Path.test.zip)

dorphalsig:
update tried it with another file that also has clipping info and got the same prob. updated tip file is attached
[Path test.zip](https://github.com/python-pillow/Pillow/files/10950851/Path.test.zip)

radarhere:
Hmm. That doesn't quite match what I'm seeing, so let me ask - what happens if you delete the EXIF key from the TIFF tags? Does the script run successfully then?
```python
from PIL import Image
img = Image.open('images/path test.tif')
del img.tag_v2[34665]
ex = ""
try:
    img.save("path test public.tif")
except Exception as e:
    ex=e
```

dorphalsig:
Hi. Nope. exact same error

radarhere:
Ok, thanks. I've put together a potential fix, that should be part of Pillow 9.5.0 when it is released on April 1.

I'll check back in around then to confirm that it does indeed also resolve the problem in your environment.

radarhere:
PR #7053 has been merged. Does that fix your problem? To test, you can either try and [install Pillow from source](https://pillow.readthedocs.io/en/stable/installation.html#building-from-source), or you can wait a day or so for Pillow 9.5.0 to be released.

PR #7053, opened 2023-03-31
Added ImageSourceData to TAGS_V2

radarhere:
In #7008, two images were reported to cause a RuntimeError. Testing on my machine, I found that this caused a segfault instead.

It turns out that this was due to an incorrect tag type being used for [ImageSourceData](https://www.awaresystems.be/imaging/tiff/tifftags/imagesourcedata.html).

Except then I found the RuntimeError, happening on [32-bit Windows.](https://github.com/radarhere/Pillow/actions/runs/4404887057/jobs/7715072523#step:27:726)
```
E       RuntimeError: Error setting from dictionary

C:\hostedtoolcache\windows\Python\3.7.9\x86\lib\site-packages\pillow-9.5.0.dev0-py3.7-win32.egg\PIL\Image.py:437: RuntimeError
---------------------------- Captured stderr call -----------------------------
_TIFFVSetField: C:\Users\runneradmin\AppData\Local\Temp\pytest-of-runneradmin\pytest-0\test_save_imagesourcedata0\temp.tif: Bad LONG8 or IFD8 value 1000052142389717520 for "EXIFIFDOffset" tag 34665 in ClassicTIFF. Tag won't be written to file.
```
When it states 'Bad LONG8 or IFD8 value 1000052142389717520 for "EXIFIFDOffset" tag 34665', that number wasn't the value being passed through for the EXIF IFD offset from Python. When we are about to pass the value through to libtiff in C, we cast it to `UINT32`.
https://github.com/python-pillow/Pillow/blob/31669013d420c2a532f5be6c5631827fe36e6f70/src/encode.c#L906-L907

When libtiff is [throwing the error](https://gitlab.com/libtiff/libtiff/-/blob/master/libtiff/tif_dir.c#L942-955), the variable is `uint64_t`. Changing `UINT32` to `uint64_t` fixed the RunTimeError.

### DIFF
--- a/src/encode.c
+++ b/src/encode.c
@@ -904,7 +904,7 @@ PyImaging_LibTiffEncoderNew(PyObject *self, PyObject *args) {
                     &encoder->state, (ttag_t)key_int, (UINT16)PyLong_AsLong(value));
             } else if (type == TIFF_LONG) {
                 status = ImagingLibTiffSetField(
-                    &encoder->state, (ttag_t)key_int, (UINT32)PyLong_AsLong(value));
+                    &encoder->state, (ttag_t)key_int, PyLong_AsLongLong(value));
             } else if (type == TIFF_SSHORT) {
                 status = ImagingLibTiffSetField(
                     &encoder->state, (ttag_t)key_int, (INT16)PyLong_AsLong(value));
--- a/Tests/test_file_libtiff.py
+++ b/Tests/test_file_libtiff.py
@@ -668,10 +668,15 @@ class TestFileLibTiff(LibTiffTestCase):
             assert reloaded.tag_v2[530] == (1, 1)
             assert reloaded.tag_v2[532] == (0, 255, 128, 255, 128, 255)
 
-    def test_save_imagesourcedata(self, tmp_path):
+    def test_exif_ifd(self, tmp_path):
         outfile = str(tmp_path / "temp.tif")
         with Image.open("Tests/images/tiff_adobe_deflate.tif") as im:
+            assert im.tag_v2[34665] == 125456
             im.save(outfile)
+
+        with Image.open(outfile) as reloaded:
+            if Image.core.libtiff_support_custom_tags:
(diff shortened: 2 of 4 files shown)


src/encode.c
```diff
@@ -904,7 +904,7 @@ PyImaging_LibTiffEncoderNew(PyObject *self, PyObject *args) {
                     &encoder->state, (ttag_t)key_int, (UINT16)PyLong_AsLong(value));
             } else if (type == TIFF_LONG) {
                 status = ImagingLibTiffSetField(
-                    &encoder->state, (ttag_t)key_int, (UINT32)PyLong_AsLong(value));
+                    &encoder->state, (ttag_t)key_int, (uint64_t)PyLong_AsLong(value));
```
nulano, line 907: I don't fully understand this change and can't look into it right now (I'm not sure why the `TIFF_LONG8` error was triggered when this change is for `TIFF_LONG`), but at least on Windows, the `PyLong_AsLong` function returns the equivalent of `int32_t`. I think `PyLong_AsLongLong` should return `int64_t` on all platforms, although I'm not sure if it matters here.

To be clear, using `PyLong_AsLong` here instead of `PyLong_AsLongLong` is not a security issue, I'm just not sure whether it could save the wrong values.

radarhere: Sure, that seems better than casting the value. I've updated the commit.

nulano: It returns `int64_t`, not `uint64_t`. It probably doesn't matter due to how signedness casting works, but it's not the same type.

mgorny:
This change broke Pillow on PowerPC: https://github.com/python-pillow/Pillow/issues/8522

To be honest, I don't really understand what's being done here. Given that `TIFF_LONG` is defined as "32-bit unsigned integer", passing a 64-bit integer (signed or unsigned) seems wrong. I'm guessing on other architectures VA are passed in such a way as to conceal the problem, and it explodes on PowerPC.

radarhere:
I've created #8529 to address the problem.

PR #7054, opened 2023-04-01
Removed unused sections in release notes

radarhere:


hugovk:
Thanks! I'll retag and restart the wheel builds.

PR #7057, opened 2023-04-01
10.0.0.dev0 version bump

radarhere:
https://github.com/python-pillow/Pillow/commit/4f7070e24c9cba8fc7ffab8d9f1c291977bbf183 set the next version of Pillow as
https://github.com/python-pillow/Pillow/blob/4f7070e24c9cba8fc7ffab8d9f1c291977bbf183/src/PIL/_version.py#L2

However, we have been listing Pillow 10.0.0 in the code as the Pillow version due out in the next quarter on the 1st of July.

https://github.com/python-pillow/Pillow/blob/4f7070e24c9cba8fc7ffab8d9f1c291977bbf183/src/PIL/_deprecate.py#L49
https://github.com/python-pillow/Pillow/blob/4f7070e24c9cba8fc7ffab8d9f1c291977bbf183/src/PIL/ImageTk.py#L173-L174
https://github.com/python-pillow/Pillow/blob/4f7070e24c9cba8fc7ffab8d9f1c291977bbf183/src/PIL/ImageShow.py#L118-L120

We've also been listing it in the docs as that date - https://pillow.readthedocs.io/en/stable/deprecations.html#tk-tcl-8-4

I think this was just an error. Happy to start a discussion if there is other thinking.


hugovk:
Yes, that's right! 10.0.0 is next!

PR #7058, opened 2023-04-01
Drop support for soon-EOL Python 3.7

hugovk:
Python 3.7 reaches end-of-life on 2023-06-27:

* https://devguide.python.org/versions/
* https://peps.python.org/pep-0537/

Which comes before our next quarterly release on 2023-07-01.

Let's remove support and free up CI resources.

---

Open issue:

* [x] Amazon Linux 2 uses Python 3.7. We either need to drop it from the test matrix, or upgrade its Python


hugovk:
> * [x]  Amazon Linux 2 uses Python 3.7. We either need to drop it from the test matrix, or upgrade its Python

Now uses 3.9: https://github.com/python-pillow/docker-images/pull/174. Documentation updated in this PR.

Issue #2563, opened 2017-06-01
Build issues on FreeBSD if two versions of openjpeg are installed


melvyn-sopacua:
### Related
[PR 2554](https://github.com/python-pillow/Pillow/pull/2544)

Recap: `/usr/local/include` is inserted before `/usr/local/include/openjpeg-2.1`, which causes problems if two versions of openjpeg are installed.

### Observations:

1. It sometimes succeeds even without the fix in mentioned PR.
2. The version specific directory is added twice.
3. When it succeeds, openjpeg is the first package tried.
4. As a result, converting the inline dict in the for loop in build_extensions to a `collections.OrderedDict` also resolves the issue.

### Root cause
Ordering aside, the root cause is that packages are essentially configured twice in different ways. First `pkg-config` is used and then best guess directory searches are used to do the same thing.

### Possible solutions
If this were to be fixed without the insert override from the PR, then fixing it relies on kits ordering. Not ideal.
A more generic approach is to use the following order of events:

1. Based on the platform, add the needed "generic" directories, so top-level "lib" and "include" paths.
2. Go through the pkg-config loop, but always prepend directories (in the reverse order as returned by pkg-config). Set state for found libraries.
3. Add available libraries, but only those not already configured by pkg-config.

### Analysis
Since this braindead OpenJPEG versioning is likely to be an exception, as is having two versions installed, the current fix from the PR looks "good enough" to me, but if maintainers want the approach from above, I'm happy to make a new PR as I'm in a good position to test it.

I do not actually understand, why packages found by `pkg-config` are revisited in the "feature" code, so that would have to be explained.

P.S.: I pretty much expect the choice will be made to not go for an elaborate fix and as such this issue logs the problem and causes.

wiredfool:
The patch for pkg-config was added to support a specific platform that would not build otherwise, IIRC NixOS. It has not been helpful on any of the dev or testing platforms, so it's not well tested.

radarhere:
#2544 was merged, solving the immediate problem.

Pillow is searching for the latest version of OpenJPEG.
https://github.com/python-pillow/Pillow/blob/943a7a89c5b8b6cd3d531da756ece661f30b0c5b/setup.py#L680-L689
Given that, I don't see why the ordering of the paths prior to that selection is relevant.

Closing. Let us know if there are any further thoughts.

PR #7072, opened 2023-04-06
Removed FIXME comment

radarhere:
PR #2544 resolved the immediate problem of #2563, but it also added a comment.

https://github.com/python-pillow/Pillow/blob/943a7a89c5b8b6cd3d531da756ece661f30b0c5b/setup.py#L684-L687

The user was thinking about the logic of the hardcoded paths in setup.py and our use of `pkg-config`, but it is possible for any given path to be in `self.compiler.include_dirs` from the very beginning of `build_extensions()` if the user added it in their install command, like
```
python3 -m pip install . -v --global-option=build_ext --global-option="-Iinsertedpath"
```

If the user wants to be eager in adding include directories, I don't think that is a problem. So this PR removes the comment, as it is at least misleading.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -681,10 +681,6 @@ class pil_build_ext(build_ext):
                 # Add the directory to the include path so we can include
                 # <openjpeg.h> rather than having to cope with the versioned
                 # include path
-                # FIXME (melvyn-sopacua):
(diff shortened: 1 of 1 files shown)


radarhere:
The 3.12 failures are fixed by PR #7059

Issue #6989, opened 2023-03-04
Release Pillow 9.5.0 on April 1, 2023


radarhere:
## Main Release

Released quarterly on January 2nd, April 1st, July 1st and October 15th.

* [x] Open a release ticket e.g. https://github.com/python-pillow/Pillow/issues/3154
* [x] Develop and prepare release in `main` branch.
* [x] Check [GitHub Actions](https://github.com/python-pillow/Pillow/actions) and [AppVeyor](https://ci.appveyor.com/project/python-pillow/Pillow) to confirm passing tests in `main` branch.
* [x] Check that all of the wheel builds [Pillow Wheel Builder](https://github.com/python-pillow/pillow-wheels) pass the tests in Travis CI and GitHub Actions.
* [x] In compliance with [PEP 440](https://www.python.org/dev/peps/pep-0440/), update version identifier in `src/PIL/_version.py`
* [x] Update `CHANGES.rst`.
* [x] Run pre-release check via `make release-test` in a freshly cloned repo.
* [x] Create branch and tag for release e.g.:
  ```bash
  git branch 9.5.x
  git tag 9.5.0
  git push --all
  git push --tags
  ```
* [x] Create and check source distribution:
  ```bash
  make sdist
  ```
* [x] Create [binary distributions](https://github.com/python-pillow/Pillow/blob/main/RELEASING.md#binary-distributions)
* [x] Check and upload all binaries and source distributions e.g.:
  ```bash
  twine check dist/*
  twine upload dist/Pillow-9.5.0*
  ```
* [x] Publish the [release on GitHub](https://github.com/python-pillow/Pillow/releases)
* [x] In compliance with [PEP 440](https://www.python.org/dev/peps/pep-0440/), increment and append `.dev0` to version identifier in `src/PIL/_version.py`

### Windows
* [x] Download the artifacts from the [GitHub Actions "Test Windows" workflow](https://github.com/python-pillow/Pillow/actions/workflows/test-windows.yml) and copy into `dist/`

### Mac and Linux
* [x] Use the [Pillow Wheel Builder](https://github.com/python-pillow/pillow-wheels):
  ```bash
  git clone https://github.com/python-pillow/pillow-wheels
  cd pillow-wheels
  ./update-pillow-tag.sh 9.5.0
  ```
* [x] Download wheels from the [Pillow Wheel Builder release](https://github.com/python-pillow/pillow-wheels/releases) and copy into `dist/`

## Publicize Release

[...]

PR #7074, opened 2023-04-08
Release refinement

hugovk:
Some updates to the release checklist, I followed these for [9.5.0](https://github.com/python-pillow/Pillow/issues/6989).

---

First, I moved the `git push --all` from near the start (after removing `.dev0` from the version and updating `CHANGES.rst`), to near the end (after incrementing and adding `.dev0` to the version).

The reason: the CI does a _lot_ of building during releases and there's a lot of waiting around for the release manager. We only need the tags pushed early to build the Windows wheels via `git push --tags`. The Linux and macOS wheels are built using their own dedicated wheel builder. The `git push --all` just delays the CI from building all these wheels.

So  `git push --all` which pushes `main` and `9.5.x` can come after all the wheel building.

---

Next, during the "Binary Distributions" stage, it says to download the Windows wheels (once they're ready); then do some things to kick off the macOS and Linux wheel builder and download once ready.

Let's flip the order. Even if Windows wheels were ready (and they won't be yet), it's better to get the Mac/Linux CI queued up first.

---

Finally, I used the [GitHub CLI `gh`](https://github.com/cli/cli) to download all the wheels before uploading them. Saves a lot of clicking in the UI! So let's include example commands for next time.

(Aside: `gh co 1234` is a really handy command for checking out a PR for local testing.)

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -105,11 +105,18 @@ Released as needed privately to individual vendors for critical security-related
   ./update-pillow-tag.sh [[release tag]]
   ```
 * [ ] Download wheels from the [Pillow Wheel Builder release](https://github.com/python-pillow/pillow-wheels/releases)
-  and copy into `dist/`
+  and copy into `dist/`. For example using [GitHub CLI](https://github.com/cli/cli):
+  ```bash
+  gh release download --dir dist --pattern "*.whl" --repo python-pillow/pillow-wheels
+  ```
 
 ### Windows
 * [ ] Download the artifacts from the [GitHub Actions "Test Windows" workflow](https://github.com/python-pillow/Pillow/actions/workflows/test-windows.yml)
-  and copy into `dist/`
+  and copy into `dist/`For example using [GitHub CLI](https://github.com/cli/cli):
+  ```bash
+  gh run download --dir dist
+  # select dist-x.y.z
+  ```
 
 ## Publicize Release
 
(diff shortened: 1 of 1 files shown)


RELEASING.md
```diff
@@ -95,19 +97,26 @@ Released as needed privately to individual vendors for critical security-related
...
 * [ ] Download wheels from the [Pillow Wheel Builder release](https://github.com/python-pillow/pillow-wheels/releases)
-  and copy into `dist/`
+  and copy into `dist/`. For example using [GitHub CLI](https://github.com/cli/cli):
+  ```bash
+  gh release download --dir dist --pattern "*.whl" --repo python-pillow/pillow-wheels
+  ```
+
+### Windows
+* [ ] Download the artifacts from the [GitHub Actions "Test Windows" workflow](https://github.com/python-pillow/Pillow/actions/workflows/test-windows.yml)
+  and copy into `dist/`For example using [GitHub CLI](https://github.com/cli/cli):
```
radarhere, line 115:
```suggestion
  and copy into `dist/`. For example using [GitHub CLI](https://github.com/cli/cli):
```

RELEASING.md
```diff
@@ -95,19 +97,26 @@ Released as needed privately to individual vendors for critical security-related
...
   ```bash
   git clone https://github.com/python-pillow/pillow-wheels
   cd pillow-wheels
   ./update-pillow-tag.sh [[release tag]]
   ```
 * [ ] Download wheels from the [Pillow Wheel Builder release](https://github.com/python-pillow/pillow-wheels/releases)
-  and copy into `dist/`
+  and copy into `dist/`. For example using [GitHub CLI](https://github.com/cli/cli):
+  ```bash
+  gh release download --dir dist --pattern "*.whl" --repo python-pillow/pillow-wheels
```
radarhere, line 110: I gather you were thinking that this command would be run from the main repository, so that all of the files would be in one neat 'dist' directory.

However, in the previous point, the user cloned and navigated into pillow-wheels. So I feel like this is confusing. It would be good to either clarify that this should be run from the main repository, or if this is run from pillow-wheels, then `--repo python-pillow/pillow-wheels` is unnecessary.

hugovk: That's right, I ran both these commands from the main repo, and agree this is confusing. Will clarify!

hugovk: Added https://github.com/python-pillow/Pillow/pull/7074/commits/cb68187006e02b3aeda44160338b67891455749e (and rebased in the PyQt workaround).

PR #7075, opened 2023-04-08
Add release check to make sure no TODOs remain in release notes

hugovk:
To avoid mistakes like https://github.com/python-pillow/Pillow/pull/7054.

We could use something like `grep -P TODO docs/releasenotes/[1-9]*.rst` but the `-P` option isn't available in the system `grep` on macOS (see https://stackoverflow.com/a/22704387/724176), and a short script is also cross platform for Windows too.

### DIFF
--- a/Tests/check_release_notes.py
+++ b/Tests/check_release_notes.py
@@ -1,0 +1,6 @@
+import sys
+from pathlib import Path
+
+for rst in Path("docs/releasenotes/").rglob("[1-9]*.rst"):
+    if "TODO" in open(rst).read():
+        sys.exit(f"Error: remove TODO from {rst}")
--- a/Tests/check_release_notes.py
+++ b/Tests/check_release_notes.py
@@ -1,6 +1,6 @@
 import sys
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


Tests/check_release_notes.py
```diff
@@ -0,0 +1,6 @@
+import sys
+from pathlib import Path
+
+for rst in Path("docs/releasenotes/").rglob("[1-9]*.rst"):
```
radarhere, line 4:
```suggestion
for rst in Path("docs/releasenotes").glob("[1-9]*.rst"):
```
I don't think the trailing slash is necessary, and `docs/releasenotes` doesn't have subdirectories.

radarhere (approved):

radarhere:
If we wanted to, we could easily limit this to just check the current version of the release notes. By this point in the release checklist, we've removed `.dev0` from the version, so as long as we moved `python3 Tests/check_release_notes.py` after the `pip install` command in `release-test`, we would know which release note doc was the latest.

But I understand the current state of this PR is more thorough.

PR #7085, opened 2023-04-12
Fix codecov after they deleted the Python package from PyPI

hugovk:
Codecov had deprecated their Python codecov package:

* https://about.codecov.io/blog/introducing-codecovs-new-uploader/

But rather than only deprecating it, they've completely deleted the package from PyPI:

* https://pypi.org/project/codecov/
* https://github.com/codecov/python-standard/issues/31
* https://community.codecov.com/t/codecov-yanked-from-pypi-all-versions/4259?u=hugovk

This is causing failures on our CI.

For AppVeyor, let's switch to the new uploader.

For others, we were pip installing codecov in `.ci/after_success.sh`, called by:

* `.github/workflows/test-docker.yml`
* `.github/workflows/test-cygwin.yml`
* `.github/workflows/test-windows.yml`
* `.github/workflows/test.yml`

We'd already moved all of these to using the Codecov GitHub Action, but were still installing the codecov package. Instead, install _coverage_, which is the actual command we need in `.ci/after_success.sh`.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -52,8 +52,9 @@ test_script:
 #- '%PYTHON%\%EXECUTABLE% test-installed.py -v -s %TEST_OPTIONS%' TODO TEST_OPTIONS with pytest?
 
 after_test:
-- python -m pip install codecov
-- codecov --file coverage.xml --name %PYTHON% --flags AppVeyor
+- curl -Os https://uploader.codecov.io/latest/windows/codecov.exe
+- chmod +x codecov
+- .\codecov.exe --file coverage.xml --name %PYTHON% --flags AppVeyor
 
 matrix:
   fast_finish: true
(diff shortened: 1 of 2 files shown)


.appveyor.yml
```diff
@@ -52,8 +52,9 @@ test_script:
 #- '%PYTHON%\%EXECUTABLE% test-installed.py -v -s %TEST_OPTIONS%' TODO TEST_OPTIONS with pytest?
 
 after_test:
-- python -m pip install codecov
-- codecov --file coverage.xml --name %PYTHON% --flags AppVeyor
+- curl -Os https://uploader.codecov.io/latest/windows/codecov.exe
+- chmod +x codecov
```
radarhere, line 56:
```suggestion
```
I find this isn't necessary - https://github.com/radarhere/Pillow/commit/4866fc4443f1e7e9b30841cd4f56bc08199ebd1e / https://ci.appveyor.com/project/radarhere/pillow/builds/46769617/job/j08ajcwcui0e2bab

radarhere (approved):

PR #7089, opened 2023-04-14
Updated nasm to 2.16.01

radarhere:
nasm 2.16.01 has been released - https://www.nasm.us/


hugovk:
Cygwin cffi build error unrelated.

radarhere:
I've started a solution for the Cygwin problem in https://github.com/python-pillow/Pillow/pull/7091

radarhere:
And the remainder of the Cygwin problem fixed itself. See #7091 for slightly more detail.

PR #7092, opened 2023-04-15
Added in_place argument to ImageOps.exif_transpose()

### LINKED ISSUE #7086: Add a switch to control whether image.copy() or not in ImageOps.exif_transpose
* Python: 3.9+
* Pillow: 9.5

Add a switch flag to control whether `image.copy()` or not if _no orientation_ or _orientation=1_ in exif to the experimental function `ImageOps.exif_transpose`.

As pointed out in https://github.com/python-pillow/Pillow/pull/5574#issuecomment-872899719 , pillow wants to _keeps the output consistent_ .
However there's unnecessary computation&time cost, and even in many cases where the `copy()` is not needed, for example:
```python
# Just open image with correct look as human eyes'

from PIL import Image
im = Image.open("hopper.jpg")
im = ImageOps.exif_transpose(im)  #assign to the same variable

#or simplified by:
im = ImageOps.exif_transpose(Image.open("hopper.jpg"))
```

It's better to let the user decide explicitly when he/she knows what's happening under the hood and wants more control. For lazy user just keep the old consistent manner.

the code maybe like this or something similar:
```python
# Defined in ImageOps.py
def exif_transpose(image, copy_when_no_transpose=True):
    # check orientation in exif
    if method is not None:
        # do something
        return transposed_image
    else:
        return image.copy() if copy_when_no_transpose else image


----------------------------------------

# Called in user code
im = exif_transpose(im, copy_when_no_transpose=False)
```

radarhere:
Resolves #7086

If there is no transposition required, `ImageOps.exif_transpose()` performs a `copy()` operation. This is to provide consistency with when there is transposition required and `transpose()` is called, so that whether or not there is a change made, a new image instance is returned.

#7086 (and effectively #5574 as well) dislike the performance cost of the `copy()` operation.

This PR introduces a new argument, `inPlace`, copying the API of [`ImageCms.applyTransform()`](https://pillow.readthedocs.io/en/stable/reference/ImageCms.html#PIL.ImageCms.applyTransform). It is ``False`` by default, to continue to provide the existing behaviour. If it is ``True``, then the provided image instance is modified, and nothing is returned. This means that `copy()` does not have to be called, as no changes need to be made to the provided image.


src/PIL/ImageOps.py
```diff
@@ -600,22 +601,30 @@ def exif_transpose(image):
...
                 ):
-                    transposed_image.info["XML:com.adobe.xmp"] = re.sub(
-                        pattern, "", transposed_image.info["XML:com.adobe.xmp"]
+                    exif_image.info["XML:com.adobe.xmp"] = re.sub(
+                        pattern, "", exif_image.info["XML:com.adobe.xmp"]
                     )
+        if inPlace:
+            return
         return transposed_image
+    if inPlace:
```
yantaozhao, line 628: use `else` here along with upmost `if` above to avoid check again when `method is not None`.

such as:
```python
def exif_transpose(image, inPlace=False):
    if method is not None:
        # apply exif orientation
        if not inPlace:
            return transposed_image
    else:
        if not inPlace:  #can merge into `else` line.
            return image.copy()
```

Yay295:
Exif tag 0x0112 is `ExifTags.Base.Orientation` if you want to import that instead of using the number directly.

radarhere:
Ok, I've pushed a commit for that.

Tests/test_imageops.py
```diff
@@ -404,6 +404,18 @@ def check(orientation_im):
     assert 0x0112 not in transposed_im.getexif()
 
 
+def test_exif_transpose_inplace():
+    with Image.open("Tests/images/orientation_rectangle.jpg") as im:
+        assert im.size == (2, 1)
+        assert im.getexif()[0x0112] == 8
```
yantaozhao, line 410: though hardcode `0x0112` is ok.
But why not change it to `ExifTags.Base.Orientation` according to src code?

radarhere: My thinking is that I know the [orientation tag is 0x0112](https://www.awaresystems.be/imaging/tiff/tifftags/orientation.html), and if Pillow's src code is changed so that the orientation tag is something different, I want this test to start failing.

yantaozhao: good

yantaozhao (approved):

src/PIL/ImageOps.py
```diff
@@ -576,19 +576,20 @@ def solarize(image, threshold=128):
     return _lut(image, lut)
 
 
-def exif_transpose(image):
+def exif_transpose(image, inPlace=False):
```
hugovk, line 579: snake_case, and make it a required keyword argument (re: https://github.com/python-pillow/Pillow/pull/6954#issuecomment-1432872156)?

```suggestion
def exif_transpose(image, *, in_place=False):
```

src/PIL/ImageOps.py
```diff
@@ -576,19 +576,20 @@ def solarize(image, threshold=128):
...
-    Otherwise, return a copy of the image.
+    If an image has an EXIF Orientation tag, other than 1, transpose the image
+    accordingly, and remove the orientation data.
 
     :param image: The image to transpose.
-    :return: An image.
-    """
-    exif = image.getexif()
-    orientation = exif.get(0x0112)
+    :param inPlace: Boolean.
```
hugovk, line 585:
```suggestion
    :param in_place: Boolean.
```

src/PIL/ImageOps.py
```diff
@@ -600,22 +601,28 @@ def exif_transpose(image):
...
-            elif "Raw profile type exif" in transposed_image.info:
-                transposed_image.info[
-                    "Raw profile type exif"
-                ] = transposed_exif.tobytes().hex()
-            elif "XML:com.adobe.xmp" in transposed_image.info:
+        if inPlace:
+            image.im = transposed_image.im
+            image.pyaccess = None
+            image._size = transposed_image._size
+        exif_image = image if inPlace else transposed_image
```
hugovk, line 608:
```suggestion
        if in_place:
            image.im = transposed_image.im
            image.pyaccess = None
            image._size = transposed_image._size
        exif_image = image if in_place else transposed_image
```

src/PIL/ImageOps.py
```diff
@@ -600,22 +601,28 @@ def exif_transpose(image):
...
-                    transposed_image.info["XML:com.adobe.xmp"] = re.sub(
-                        pattern, "", transposed_image.info["XML:com.adobe.xmp"]
+                    exif_image.info["XML:com.adobe.xmp"] = re.sub(
+                        pattern, "", exif_image.info["XML:com.adobe.xmp"]
                     )
-        return transposed_image
-    return image.copy()
+        if not inPlace:
+            return transposed_image
+    elif not inPlace:
```
hugovk, line 627:
```suggestion
        if not in_place:
            return transposed_image
    elif not in_place:
```

Tests/test_imageops.py
```diff
@@ -404,6 +404,18 @@ def check(orientation_im):
...
+        expected = im.rotate(90, expand=True)
+
+        ImageOps.exif_transpose(im, inPlace=True)
```
hugovk, line 413:
```suggestion
        ImageOps.exif_transpose(im, in_place=True)
```

Tests/test_imageops.py
```diff
@@ -404,6 +404,18 @@ def check(orientation_im):
     assert 0x0112 not in transposed_im.getexif()
 
 
+def test_exif_transpose_inplace():
```
hugovk, line 407:
```suggestion
def test_exif_transpose_in_place():
```

src/PIL/ImageOps.py
radarhere: I've pushed a commit to make it a keyword-only argument, but regarding snake_case, would it be better to match the existing `inPlace` case of [`ImageCms.applyTransform()`](https://pillow.readthedocs.io/en/stable/reference/ImageCms.html#PIL.ImageCms.applyTransform)?

yantaozhao (approved):

src/PIL/ImageOps.py
hugovk: Hmm, consistency is good, but I think these two modules are sufficiently different that we can use snake_case for new things here.

hugovk (approved):

PR #7094, opened 2023-04-15
Fix ImageGrab with wl-paste

### LINKED ISSUE #7093: `wl-paste` with image copied from browser results in `ImageGrab.grabclipboard` error
### What did you do?

When source application offered multiple MIME type, `wl-paste` without `-t mime/type` maybe result in other MEME type instead of image. `ImageGrab.grabclipboard` use only `wl-paste` without arguments, which reports error whenever calling it with multiple MEME type. For example, open some website with images, copy the image, it will offer multiple MIME type.

No error reports.

### What actually happened?

Error reports like this:

```
  File "/home/user/workspace/grab.py", line 9, in <module>
    main()
  File "/home/user/workspace/grab.py", line 5, in main
    ImageGrab.grabclipboard()
  File "/home/user/.local/lib/python3.10/site-packages/PIL/ImageGrab.py", line 146, in grabclipboard
    im = Image.open(filepath)
  File "/home/user/.local/lib/python3.10/site-packages/PIL/Image.py", line 3298, in open
    raise UnidentifiedImageError(msg)
PIL.UnidentifiedImageError: cannot identify image file '/tmp/tmpvi9tcfs3'
```

* OS: Arch Linux
* Python: CPython 3.10.10
* Pillow: 9.5.0

```python
from PIL import ImageGrab


def main():
    ImageGrab.grabclipboard()


if __name__ == "__main__":
    main()
```

rrcgat:
Fixes #7093

 * Add `-t <mime/type>` arguments to `wl-paste` when there is an available image MIME type.


rrcgat:
The CI skip my test, which covered on my local machine. I'm not familiar with GitHub CI, anyone helps?

image

nulano:
I guess `wl-clipboard` is not installed by default in the Ubuntu 22.04 image:
https://github.com/python-pillow/Pillow/actions/runs/4707895239/jobs/8351231955?pr=7094#step:8:4513
https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md

rrcgat:
@nulano  Thanks for your reply, I'll try to modify the GitHub workflow to make it work.

Tests/test_imagegrab.py
```diff
@@ -98,3 +98,22 @@ def test_grabclipboard_png(self):
...
+        "image_path", ["Tests/images/hopper.gif", "Tests/images/hopper.png"]
+    )
+    def test_grabclipboard_wl_clipboard(self, image_path):
+        with open(image_path, mode="rb") as raw_image:
+            try:
+                subprocess.call(["wl-copy"], stdin=raw_image)
+                im = ImageGrab.grabclipboard()
+                assert_image_equal_tofile(im, image_path)
+            except OSError as e:
+                pytest.skip(str(e))
```
radarhere, line 119: What is the `OSError` that might be thrown here?

src/PIL/ImageGrab.py
```diff
@@ -135,6 +135,12 @@ def grabclipboard():
     else:
         if shutil.which("wl-paste"):
             args = ["wl-paste"]
+            output = subprocess.check_output(["wl-paste", "-l"]).decode()
+            mime_types = output.splitlines()
+            for image_type in ["image/gif", "image/png"]:
+                if image_type in mime_types:
+                    args.extend(["-t", image_type])
```
radarhere, line 142:
```suggestion
            clipboard_mimetypes = output.splitlines()
            Image.preinit()
            for mimetype in Image.MIME.values():
                if mimetype in clipboard_mimetypes:
                    args.extend(["-t", mimetype])
```
This would allow you to use more than just GIF and PNG.

Tests/test_imagegrab.py
rrcgat: Oh, `OSError` will not be thrown here anymore. I made a mistake with skipif condition and OSError reported in other systems.

rrcgat:
Hi everybody, I've tried a lot in [test-workflow](https://github.com/rrcgat/Pillow/tree/test-workflow) branch, but I couldn't find a solution to support wl-clipboard in xvfb. 

The errors include:
1. Failed to connect to a Wayland server (wl-clipboard)
2. Core dumped without xwayland (in other test files)
3. Missing a seat (wl-clipboard, I've tested `seatd` on Docker, not work)

Do you have any suggestions?

radarhere:
If you don't know of a simple way to add coverage for this, then I wouldn't think that is a blocker for this PR. You're not adding code that isn't covered into the middle of an otherwise-covered section - we weren't testing wl-paste or xclip on our CIs to begin with. See https://app.codecov.io/gh/python-pillow/Pillow/blob/main/src/PIL/ImageGrab.py

rrcgat:
Hi @radarhere , I finally found a way to test wl-paste by using sway. The CI is running, I'll commit to this branch if it works.

radarhere:
Unfortunately, there is a problem. If I [remove the fix](https://github.com/radarhere/Pillow/commit/e1a9b1ecbf401ab1f3d203b0054892e3693ee238), the [test doesn't fail.](https://github.com/radarhere/Pillow/actions/runs/4720198212/jobs/8371944230)

rrcgat:
> 

I'm not sure why the test is failing. I tried running it again, this time with the CODECOV_TOKEN set, and it [doesn't fail too](https://github.com/rrcgat/Pillow/actions/runs/4719556542/jobs/8372139689#step:8:2796):
image

rrcgat:
> 

Hi @radarhere , If you remove the code, the test won't fail because there's only one MIME type (set by `subprocess.call(["wl-copy"], stdin=raw_image)`), so `wl-paste` without arguments works fine.

image

rrcgat:
I didn't set the CODECOV_TOKEN previously, but after setting it, I reran some workflows.

radarhere:
I've created https://github.com/rrcgat/Pillow/pull/1 with some suggestions for this PR.

rrcgat:
Thanks for getting back to me! Apologies for the inconvenience, I wasn't aware of these suggestions until you left a comment.

radarhere:
I've discovered a problem. I ran Ubuntu, switched to Wayland, opened https://python-pillow.org/images/pillow-logo-light-text-1280x640.png in Firefox, copied the image, and ran `wl-paste -l`.
The first option in `wl-paste -l` was TIFF, and when I selected that as the mime type, it saved an empty file!
Only when I chose PNG instead did the image save.

https://user-images.githubusercontent.com/3112309/233653058-471300bd-7d64-4566-b258-6707b9f2baed.mp4

If the first suggestion from `wl-paste -l` doesn't necessarily work, I guess that means this PR needs to iterate through the mime types until it finds one that works? That seems cumbersome.

rrcgat:
@radarhere Yes, you're right, I didn't test it to much because I simply thought `image/png` and `image/gif` is enough for the image type of `wl-paste`.

In the early days before, I found that `image/png` can be use for ICO and JPEG in the case of `wl-paste`. Withou much knowledge of image type, I thought that `image/png` can be use for other image type, so I only check for PNG and GIF format in my first commit.

When I opened https://python-pillow.org/images/pillow-logo-light-text-1280x640.png, Firefox and Chrome offered different MIME types (Chrome only offered `text/html` and `image/png`), and I have different order of `wl-paste -l` compare to yours, which both copied from Firefox (with Gnome and Wayland). For the PNG and JPEG image, even GIF (for example, https://upload.wikimedia.org/wikipedia/commons/thumb/2/2c/Rotating_earth_%28large%29.gif/200px-Rotating_earth_%28large%29.gif) Chome offered the same MIME type (the output of `wl-paste -l`).

It seems that there are too many cases of copy, iterate through the mime types works (with different MIME types, `wl-paste` output with different file size, I don't know if it matters), but as you said, it's cumbersome.

rrcgat:
Use the hard way here: https://github.com/rrcgat/Pillow/pull/2/files

What do you think?
```python
        if shutil.which("wl-paste"):
            args = ["wl-paste"]
            output = subprocess.check_output(["wl-paste", "-l"]).decode()
            clipboard_mimetypes = set(output.splitlines())

            Image.preinit()
            available_mimetypes = set(Image.MIME.values()) & clipboard_mimetypes
            if not available_mimetypes:
                Image.init()
                available_mimetypes = set(Image.MIME.values()) & clipboard_mimetypes
            if available_mimetypes:
                return _grab(
                    *[["wl-paste", "-t", mimetype] for mimetype in available_mimetypes]
                )
        elif shutil.which("xclip"):
            args = ["xclip", "-selection", "clipboard", "-t", "image/png", "-o"]
        else:
            msg = "wl-paste or xclip is required for ImageGrab.grabclipboard() on Linux"
            raise NotImplementedError(msg)
        return _grab(args)


def _grab(*call_args):
    fh, filepath = tempfile.mkstemp()
    for args in call_args:
        subprocess.call(args, stdout=fh)
        if os.path.getsize(fh):
            break
    os.close(fh)
    try:
        im = Image.open(filepath)
        im.load()
    finally:  # Remove it anyway or left it when Image.open fails (for debug usage)?
        os.unlink(filepath)
    return im
```

radarhere:
I think iterating through all of the different mime types doesn't sit nicely with #7110, a request that we raise an error on a failed ImageGrab operation - because if we try 12 different wl-paste variations, then we don't know which error to show.

Let me suggest this instead. Given the output of `wl-paste -l`,
1. If there is only one mime type, use that.
2. If there are multiple mime types, and one of them is "image/png", use that. This worked in my earlier recording for a GIF, when the first mime type was TIFF.
3. Otherwise, use the first mime type listed.

What do you think?

rrcgat:
@radarhere I agree, this approach is better than iterating through all mime types and it works well in my initial case. I will implement this solution based on your suggestion.

rrcgat:
Using the first mime type when there are multiple mime types without image/png is great, although image/png mime type is provided in the video:

[Screencast from 2023-05-23 18-50-35.webm](https://github.com/python-pillow/Pillow/assets/26518505/7fb656d0-4786-4fda-8a68-97ac42ea2504)

PR #7100, opened 2023-04-18
If available, use scrot on Linux for ImageGrab

HansBug:
Use scrot when this environment have in `ImageGrab.grab()`.

In practical use, although `gnome-screenshot` is widely present in Linux systems, in newer systems such as Ubuntu 20.04, **there will be a full-screen animation effect when taking screenshots, and it is not easy to close**. This feature may cause inconvenience to users in some tasks. Therefore, it has been changed to detect `scrot`. If `scrot` is present, it will be used directly. There will be no similar problems when taking screenshots with `scrot`, and it is very easy to install on Ubuntu and other systems with `apt install -y scrot`, and similar packages are available on other systems such as debian and centos.

Here are some more records about the annoying flash animation on `gnome-screenshot` @radarhere :
* https://www.linux.org/threads/gnome-screenshot-without-the-annoying-flash-more.28655/
* https://askubuntu.com/questions/854350/disable-gnome-screenshots-camera-flash-animation
* https://gitlab.gnome.org/GNOME/gnome-screenshot/-/issues/2  (This is the official explaination)

### DIFF
--- a/docs/reference/ImageGrab.rst
+++ b/docs/reference/ImageGrab.rst
@@ -15,8 +15,9 @@ or the clipboard to a PIL image memory.
     returned as an "RGBA" on macOS, or an "RGB" image otherwise.
     If the bounding box is omitted, the entire screen is copied.
 
-    On Linux, if ``xdisplay`` is ``None`` then ``gnome-screenshot`` will be used if it
-    is installed. To capture the default X11 display instead, pass ``xdisplay=""``.
+    On Linux, if ``xdisplay`` is ``None`` then ``gnome-screenshot`` or ``scrot`` will
+    be used if they are installed. To capture the default X11 display instead, pass
+    ``xdisplay=""``.
 
     .. versionadded:: 1.1.3 (Windows), 3.0.0 (macOS), 7.1.0 (Linux)
 
--- a/docs/reference/ImageGrab.rst
+++ b/docs/reference/ImageGrab.rst
@@ -15,9 +15,9 @@ or the clipboard to a PIL image memory.
     returned as an "RGBA" on macOS, or an "RGB" image otherwise.
     If the bounding box is omitted, the entire screen is copied.
 
-    On Linux, if ``xdisplay`` is ``None`` then ``gnome-screenshot`` or ``scrot`` will
-    be used if they are installed. To capture the default X11 display instead, pass
-    ``xdisplay=""``.
+    On Linux, if ``xdisplay`` is ``None`` then (in order) ``scrot`` or
+    ``gnome-screenshot`` will be used if they are installed. To capture the default
+    X11 display instead, pass ``xdisplay=""``.
 
     .. versionadded:: 1.1.3 (Windows), 3.0.0 (macOS), 7.1.0 (Linux)
 
--- a/src/PIL/ImageGrab.py
+++ b/src/PIL/ImageGrab.py
@@ -64,7 +64,7 @@ def grab(bbox=None, include_layered_windows=False, all_screens=False, xdisplay=N
         elif shutil.which("gnome-screenshot") or shutil.which("scrot"):
             fh, filepath = tempfile.mkstemp(".png")
             os.close(fh)
-            if shutil.which("scrot"):  # use scrot when have
+            if shutil.which("scrot"):  # prefer scrot, as it is less intrusive
                 subprocess.call([shutil.which("scrot"), "-z", "--overwrite", filepath])
             else:
                 subprocess.call([shutil.which("gnome-screenshot"), "-f", filepath])
--- a/src/PIL/ImageGrab.py
+++ b/src/PIL/ImageGrab.py
@@ -61,10 +61,13 @@ def grab(bbox=None, include_layered_windows=False, all_screens=False, xdisplay=N
                 left, top, right, bottom = bbox
                 im = im.crop((left - x0, top - y0, right - x0, bottom - y0))
             return im
-        elif shutil.which("gnome-screenshot"):
+        elif shutil.which("gnome-screenshot") or shutil.which("scrot"):
             fh, filepath = tempfile.mkstemp(".png")
             os.close(fh)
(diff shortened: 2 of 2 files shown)


src/PIL/ImageGrab.py
```diff
@@ -61,10 +61,13 @@ def grab(bbox=None, include_layered_windows=False, all_screens=False, xdisplay=N
                 left, top, right, bottom = bbox
                 im = im.crop((left - x0, top - y0, right - x0, bottom - y0))
             return im
-        elif shutil.which("gnome-screenshot"):
+        elif shutil.which("gnome-screenshot") or shutil.which("scrot"):
             fh, filepath = tempfile.mkstemp(".png")
             os.close(fh)
-            subprocess.call(["gnome-screenshot", "-f", filepath])
+            if shutil.which("scrot"):  # use scrot when have
+                subprocess.call([shutil.which("scrot"), "-z", "--overwrite", filepath])
```
radarhere, line 68:
```suggestion
                subprocess.call([shutil.which("scrot"), "-z", filepath])
```
I don't think it should be necessary to overwrite the file, since it is a temporary file.

HansBug: Regarding this issue, in my local operating environment, if the --overwrite option is removed, the following exception message will be thrown:

PIL.UnidentifiedImageError: cannot identify image file
The reason for this is that the temporary file does not exist. However, after adding the --overwrite option, this issue disappears.

After testing, scrot defaults to renaming the file if it detects that the file to be saved already exists, and the --overwrite option can disable this feature and force the screenshot information to be written to the target file.

radarhere:
> it is very easy to install on Ubuntu and other systems with `apt install -y scrot`

For the record, when testing I found I had to add the universe repository on Ubuntu before I could install it.

docs/reference/ImageGrab.rst
```diff
@@ -15,8 +15,9 @@ or the clipboard to a PIL image memory.
     returned as an "RGBA" on macOS, or an "RGB" image otherwise.
     If the bounding box is omitted, the entire screen is copied.
 
-    On Linux, if ``xdisplay`` is ``None`` then ``gnome-screenshot`` will be used if it
-    is installed. To capture the default X11 display instead, pass ``xdisplay=""``.
+    On Linux, if ``xdisplay`` is ``None`` then ``gnome-screenshot`` or ``scrot`` will
+    be used if they are installed. To capture the default X11 display instead, pass
+    ``xdisplay=""``.
```
radarhere, line 20:
```suggestion
    On Linux, if ``xdisplay`` is ``None`` then (in order) ``scrot`` or
    ``gnome-screenshot`` will be used if they are installed. To capture the default
    X11 display instead, pass ``xdisplay=""``.
```
I realise now that we should mention the order of preference.

radarhere:
> In practical use, although gnome-screenshot is widely present in Linux systems, in newer systems such as Ubuntu 20.04, there will be a full-screen animation effect when taking screenshots, and it is not easy to close. This feature may cause inconvenience to users in some tasks.

When I use `gnome-screenshot` by itself, yes, I see a disruptive white flash - but I don't see that when I just use Pillow.

https://user-images.githubusercontent.com/3112309/233002319-995beeeb-055c-421e-b208-ba99e5ddc526.mov

Do you have a different experience?

radarhere (dismissed):

src/PIL/ImageGrab.py
```diff
@@ -61,10 +61,13 @@ def grab(bbox=None, include_layered_windows=False, all_screens=False, xdisplay=N
                 left, top, right, bottom = bbox
                 im = im.crop((left - x0, top - y0, right - x0, bottom - y0))
             return im
-        elif shutil.which("gnome-screenshot"):
+        elif shutil.which("gnome-screenshot") or shutil.which("scrot"):
```
nulano, line 64: For clarity, I would put the two options in the order they are used in:
```suggestion
        elif shutil.which("scrot") or shutil.which("gnome-screenshot"):
```

radarhere:
Back when we initially added `gnome-screenshot`, there was a [link](https://github.com/python-pillow/Pillow/issues/6312#issuecomment-1132278281) to a [StackOverflow issue where `scrot` resulted "in a completely black image".](https://stackoverflow.com/questions/72216896/how-can-i-get-a-screenshot-on-wayland-with-pure-python) Apparently, `scrot` [doesn't work under Wayland](https://askubuntu.com/a/1411760). I tested, and yes, I found that it worked under X11, but not Wayland.

Ok, so `scrot` isn't useful for Wayland. But it's still useful for X11?

Well, `Image.core.grabscreen_x11` already works under X11, and from what I can see, does not produce a white flash. So does supporting `scrot` actually provide any advantage?

Is the problem just a user trying to ignore `gnome-screenshot` when it is installed in favour of `Image.core.grabscreen_x11`? If so, then `ImageGrab.grab(xdisplay=":0")` should be sufficient with the current version of Pillow.

radarhere:
@HansBug any thoughts?

HansBug:
> Is the problem just a user trying to ignore `gnome-screenshot` when it is installed in favour of `Image.core.grabscreen_x11`? If so, then `ImageGrab.grab(xdisplay=":0")` should be sufficient with the current version of Pillow.

Sorry for not responding earlier due to work-related reasons. I have just tested setting `xdisplay=':0'`, and it ran on my end without any white flash. I believe this usage needs to be tested across different platforms. If the test is successful, it should be set as the default option. Otherwise, we still need to detect different environments and use different parameters and screenshot methods to avoid the occurrence of white flashes when using the default parameters.

radarhere:
Ok, I've created PR #7139 to prefer X11 over gnome-screenshot.

If no one is aware of a need for scrot, then it doesn't seem necessary to search through the many flavours of Linux trying to find a problem to match this solution.

radarhere:
PR #7143 has been merged.

Issue #7032, opened 2023-03-22
__array_interface__ - get size of image after loading


bigcat88:
To be short, I want to ask to one change that do not break any compatibility, and this case is not usual.

Here is what Pillow has now:
https://github.com/python-pillow/Pillow/blob/e7fa3091a065a51ceb7e656b71f67ee587f86082/src/PIL/Image.py#L686-L712

I want to ask to move these lines:
```python3
        shape, typestr = _conv_type_shape(self)
        new["shape"] = shape
        new["typestr"] = typestr
        new["version"] = 3
```
at the end , right after before ``return new``

Reason why it is important to me.
I received such an [issue](https://github.com/bigcat88/pillow_heif/issues/86) where images in `HEIF` format created by **Sony ILCE-7M4** camera has encoded size in itself different from the size value in `HEIF` header. (HEIF format has `size` in header and also `size` of decoded image, and for some unknown reason [they can be different](https://github.com/strukturag/libheif/issues/784))

As I wrote in that issue, I added an optional option ALLOW_INCORRECT_HEADERS(default disabled) that will allow to load such images, and just change the `Image.size` during `load` method.
Tested it, and all seems work well, except this array property, cause here first fills image size in `_conv_type_shape` and after that Image get decoded.

This simple change will do nothing worse, so I will be very appreciated if you agree to move that lines.
Thanks.

radarhere:
Your request is actually similar to #6186, so sure. I've created PR #7034 to resolve this.

PR #7105, opened 2023-04-21
Load before getting size in __getstate__

bigcat88:
Fixes the same what was in https://github.com/python-pillow/Pillow/issues/7032 but for `deepcopy(im)` operation.

 * first we load image data in `__getstate__` and then only fill list with it's attributes.

### DIFF
--- a/Tests/test_numpy.py
+++ b/Tests/test_numpy.py
@@ -1,4 +1,5 @@
 import warnings
+from copy import deepcopy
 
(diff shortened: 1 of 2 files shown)


radarhere:
I've created https://github.com/bigcat88/Pillow/pull/4 with a suggestion to simplify and move the test.

radarhere (approved):

PR #7111, opened 2023-04-24
Support reading signed 8-bit TIFF images

### LINKED ISSUE #811: Fails to open OME-TIFF example data files
[OME-TIFF](https://www.openmicroscopy.org/site/support/ome-model/ome-tiff/index.html) is a TIFF format that embeds some microscopy-specific metadata as an XML comment embedded in the TIFF header.  Sample data is available at https://www.openmicroscopy.org/site/support/ome-model/ome-tiff/data.html, but PIL fails to open `single-channel.ome.tiff` (throwing `OSError: cannot identify image file 'single-channel.ome.tiff'`).

radarhere:
Resolves #811 using the [suggested code](https://github.com/python-pillow/Pillow/issues/811#issuecomment-49553185).

Test image was created with a modified Pillow.


hugovk:
Let's add this to the release notes.

radarhere:
This is not completely simple, because there are a variety of specific settings in `OPEN_INFO`.

I've adjusted the code to also support both ByteOrders, and presumed that I don't need to mention that there aren't any ExtraSamples supported and that FillOrder 2 isn't supported since "[FillOrder=2 is not required in Baseline TIFF compliant readers](https://www.awaresystems.be/imaging/tiff/tifftags/fillorder.html)".

So the release notes read

> Support reading signed 8-bit TIFF images
> ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>
> TIFF images with signed integer data, 8 bits per sample and a photometric interpretaton of BlackIsZero can now be read.

PR #7116, opened 2023-04-26
Build all readthedocs formats

### LINKED ISSUE #7114: add documentation for downloading
### What did you do?
I was trying to download Pillow documentation. I visited Pillow site, then ReadTheDocs, then I could not download the documentation in one link.

I expected that there will be *pdf* or at least zipped *html* for downloading.

### What actually happened?
On ReadTheDocs the [downloads](https://readthedocs.org/projects/pillow/downloads/) section was empty.

* OS: Arch Linux.
* Python: yes.
* Pillow: yes. From PyPI, I believe.

radarhere:
Resolves #7114

The issue reports that there are no downloads at https://readthedocs.org/projects/pillow/downloads/

After looking at https://docs.readthedocs.io/en/stable/config-file/v2.html#formats, I've inserted `formats: all` to .readthedocs.yml.

As evidence that this works, https://readthedocs.org/projects/pillow-radarhere/versions/ lists 'latest' and 'formats' as versions, https://readthedocs.org/projects/pillow-radarhere/builds/ shows recent builds of both, but only 'formats' is visible at https://readthedocs.org/projects/pillow-radarhere/downloads/ as downloads (while 'stable' also appears, it is very old).

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,5 +1,7 @@
 version: 2
 
+formats: all
+
 build:
   os: ubuntu-22.04
   tools:


aclark4life:
LGTM, thanks!

hugovk:
This nearly doubles docs build times from ~137s to ~230s.

Do we need all four build formats? PDF was the preferred format in #7114, is default + PDF enough to begin with?

radarhere:
I've left a [comment](https://github.com/python-pillow/Pillow/issues/7114#issuecomment-1526554730) to see what the user thinks.

Issue #4849, opened 2020-08-10
Regression in 7.1 with Image.getbbox returning None caused by converting the file to RGBA


blayzen-w:
### What did you do?
In version 7.1.0, after converting the file to RGBA the getbbox will return `None` instead of the bounding box of the difference of the two images. This has happened on every file I have tried it on so far. This was working on and below 7.0.0.

### What did you expect to happen?
It should return the same result as if it wasn't converted.

### What actually happened?
It returned `None`

### What are your OS, Python and Pillow versions?
* OS: Mac 10.14.6 and ubuntu:18.04
* Python: 3.6.5
* Pillow: 7.1.0 though 7.2.0

This a small reproducible test case though the python shell.

Version 7.0.0 (working)
```pycon
>>> from PIL import Image, ImageChops
>>> img = Image.open('some_file.jpg')
>>> bg = Image.new(img.mode, img.size, img.getpixel((0, 0)))
>>> diff = ImageChops.difference(img, bg)
>>> diff.getbbox()
(328, 328, 3880, 3885)
>>> 
>>> img = Image.open('some_file.jpg')
>>> img = img.convert('RGBA')
>>> bg = Image.new(img.mode, img.size, img.getpixel((0, 0)))
>>> diff = ImageChops.difference(img, bg)
>>> diff.getbbox()
(328, 328, 3880, 3885)
```

Version 7.1.0 (broken)
```pycon
>>> from PIL import Image, ImageChops
>>> img = Image.open('some_file.jpg')
>>> bg = Image.new(img.mode, img.size, img.getpixel((0, 0)))
>>> diff = ImageChops.difference(img, bg)
>>> diff.getbbox()
(328, 328, 3880, 3885)
>>> 
>>> img = Image.open('some_file.jpg')
>>> img = img.convert('RGBA')
>>> bg = Image.new(img.mode, img.size, img.getpixel((0, 0)))
>>> diff = ImageChops.difference(img, bg)
>>> diff.getbbox()
>>> 
```

Let me know if you need any other information.

radarhere:
Testing, I find this is the result of #4454.

If I run your code with https://github.com/python-pillow/Pillow/blob/master/Tests/images/hopper.jpg, converting to RGBA, and save `diff` to a PNG, I get this -
![diff](https://user-images.githubusercontent.com/3112309/89879156-dbdf1980-dc05-11ea-8e80-ea1a6095d5fb.png)

[...]

PR #7123, opened 2023-04-29
Added alpha_only argument to getbbox()

### LINKED ISSUE #7122: Alpha channel is discarded when saving an animated PNG
### What did you do?

Created an animated PNG from RGBA images with transparent colors (non-zero alpha).

The animated PNG should preserve the alpha channel of the individual images.

### What actually happened?

The alpha channel of all but the first image in the sequence is discarded. If the frames are of a static object fading in or out, then PIL then thinks that all of the frames are the same, and then discards them.

This bug also seems to occur with animated GIFs, but I'm not 100% sure the root cause is the same.

* OS: macOS 13.3.1
* Python: 3.11.3
* Pillow: 9.5.0

This script will create an APNG from 10 frames, with the alpha of the red square in the center increasing each frame: 
```python
#!/usr/bin/env python3

images = []
for i in range(10):
    img = Image.new('RGBA', (300, 300), (255, 255, 255, 255))

    draw = ImageDraw.Draw(img)
    draw.rectangle((100, 100, 200, 200), fill=(255, 0, 0, i * 20))

    img.save(f'frame-{i}.png')
    images.append(img)

images[0].save(
    'test.png', save_all=True, append_images=images[1:], duration=100, loop=0
)
images[0].show()
```
Here is the PNG file with the transparency issue:
test-incorrect

What happened here is that when the alpha channel was thrown away, all of the images in `append_images` are exactly the same and they get thrown away.

This is how I would expect the correct image to look:
test-correct

### LINKED ISSUE #7207: Image.save(append_images) cropping using bg color instead of alpha
### What did you do?
I used `Image.save(append_images=[])` to generate gif but it cropped part of the image (got boundary box of the image wrong). It throws away alpha when preparing to add image and then it relies only on bg color, which does not work when the image has the same color as background.
 
* OS: Linux (Debian GNU/Linux 12 (bookworm) x86_64)
* Python: 3.11.2
* Pillow: 9.5.0

code:
```python

with Image.open('testin3.png') as avatar:
    frames = []
    deformWidth = [-1, -2]
    deformHeight = [4, 3]
    width, height = 80, 80
    x, y = 112, 122

    for i in range(2):
        frame = Image.new("RGBA", (x, y), (0, 0, 0, 0))
        width = width - deformWidth[i]
        height = height - deformHeight[i]
        avatar = avatar.resize((width, height))
        avatar = avatar.convert("P", palette=Image.ADAPTIVE, colors=200).convert("RGBA")

        frame.paste(avatar, (x - width, y - height), avatar)
        frames.append(frame)

    frames[0].save(
        "testout.gif",
        save_all=True,
        append_images=frames[1:],
        duration=400,
        loop=0,
        transparency=0,
        disposal=2,
        optimize=False
    )
```
testin3.png: testin3

testout.gif: testout

It could be solved by [this](https://github.com/python-pillow/Pillow/issues/7122#issuecomment-1528526139).

radarhere:
Resolves #7207
Resolves #7122

When saving APNG frames, APNG converts the individual frames to RGB to check if there are any differences. If there aren't, then it is a duplicate frame, and can be combined with the previous one to save space.

https://github.com/python-pillow/Pillow/blob/d1731b35fed4e69158d98bffbf5a415f3a4cec8d/src/PIL/PngImagePlugin.py#L1140-L1151

However, the issue tries to save RGBA frames that become identical when converted to RGB. So instead, this PR changes the code to convert the frames to RGBA.

Except that then introduces a new problem. If two frames have the same alpha values, then `ImageChops.subtract_modulo()` returns a image showing that there is 0 alpha difference... when is then interpreted by `getbbox()` as a fully transparent image, meaning that as per #4454, it ignores whatever RGB information there might be.

 So this PR adds an argument to the C `getbbox()` to allow it to turn off the special treatment of the alpha channel.


Yay295:
`consider_alpha` doesn't sound quite right to me. I think `alpha_only` fits a bit better, or reverse it and call it `consider_all_bands`.

radarhere:
I've pushed a commit to rename it to `alpha_only`.

Yay295:
It can be a separate change, but should this new parameter be exposed to the public api?

radarhere:
Ok, I've added an `alpha_only` argument to the Python `getbbox()`. This was originally requested in #4849

gsingh93:
I've been using a local build with this patch for a few weeks now and it's been working well, hopefully we can get this merged soon :)

radarhere:
Be aware that even if it is merged, Pillow 10.0.0 is not due out until July 1.

src/PIL/Image.py
```diff
@@ -1279,11 +1279,14 @@ def getbands(self):
         """
         return ImageMode.getmode(self.mode).bands
 
-    def getbbox(self):
+    def getbbox(self, alpha_only=True):
```
hugovk, line 1282: Make it a required keyword argument (re: https://github.com/python-pillow/Pillow/pull/6954#issuecomment-1432872156)?


```suggestion
    def getbbox(self, *, alpha_only=True):
```

src/PIL/Image.py
```diff
@@ -1279,11 +1279,14 @@ def getbands(self):
...
         """
         Calculates the bounding box of the non-zero regions in the
         image.
 
+        :param alpha_only: Optional flag, defaulting to true.
+           If true and the image has an alpha channel, trim transparent pixels.
```
hugovk, line 1288:
```suggestion
        :param alpha_only: Optional flag, defaulting to ``True``.
           If ``True`` and the image has an alpha channel, trim transparent pixels.
```

radarhere:
I've added a commit to resolve #7207 - the change involves the same logic presented initially for PngImagePlugin, just this time for GifImagePlugin.

radarhere:
Pillow 10.0.0 has now been released with this fix.

PR #7131, opened 2023-05-02
Use stdlib for setuptools on MinGW

radarhere:
MinGW is currently failing in main - https://github.com/python-pillow/Pillow/actions/runs/4861154081

https://www.msys2.org/docs/python/ states
> [setuptools](https://github.com/pypa/setuptools) >= 60.0 is currently incompatible with MSYS2. You can set export SETUPTOOLS_USE_DISTUTILS=stdlib to work around the issue. We are currently working on restoring compatibility.

Adding in `SETUPTOOLS_USE_DISTUTILS="stdlib"` fixes the problem.

### DIFF
--- a/.github/workflows/test-mingw.yml
+++ b/.github/workflows/test-mingw.yml
@@ -80,7 +80,7 @@ jobs:
           pushd depends && ./install_extra_test_images.sh && popd
 
       - name: Build Pillow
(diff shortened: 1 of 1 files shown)


hugovk:
Thanks for figuring it out!

Issue #1090, opened 2015-01-28
Consider implementing _repr_png_ for the Image object


payne92:
Given the popularity of IPython, consider implementing a `_repr_png_()` method for the Image object, which just returns the object in PNG form.  If this method exists on an object, the IPython notebook will display the rich image representation in-line.

See: http://nbviewer.ipython.org/github/adrn/ipython/blob/1.x/examples/notebooks/Part%205%20-%20Rich%20Display%20System.ipynb

This can be done without dependency on or reference to IPython.

wiredfool:
I've got a branch with that method in it now, (https://github.com/wiredfool/Pillow/tree/repr_png) but it does not seem to trigger iPython's display hooks. Can you give a step by step to test it, or determine what else needs to be added to make it work?

payne92:
The method is _repr_png_() -- your implementation is missing the trailing underscore in the name.

To test in an IPytyon web notebook:

    from PIL import Image
    im = Image.new("RGB", (512, 512), "blue")
    im

IPython will attempt to render the last value.  When it finds _repr_png_(), it will render it.

hugovk:
The problem was GitHub interpreted the underscores as italic markup. I've edited to put backticks ` to make sure it's in code markup!

payne92:
Ah, my bad -- sorry about that.  I should have checked the preview more carefully.

wiredfool:
Ah, much better. Works now.

hugovk:
Implemented #1091, thanks.

PR #7135, opened 2023-05-04
Added `_repr_jpeg_` for IPython display

n3011:

 *  Add `_repr_jpg_` and `_repr_jpeg_` method for ipython display (similar to currently existing `_repr_png_`)

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -633,18 +633,35 @@ class Image:
             )
         )
 
-    def _repr_png_(self):
+    def _repr_image(self, format):
         """iPython display hook support
 
+        :param format: Image format.
         :returns: png version of the image as bytes
         """
         b = io.BytesIO()
         try:
-            self.save(b, "PNG")
+            self.save(b, format)
         except Exception as e:
-            msg = "Could not save to PNG for display"
+            msg = f"Could not save to {format} for display"
             raise ValueError(msg) from e
         return b.getvalue()
+
+    def _repr_png_(self):
+        """iPython display hook support for PNG format.
+
+        :returns: png version of the image as bytes
+        """
+        return self._repr_image("PNG")
+
+    def _repr_jpg_(self):
+        """iPython display hook support for JPEG format.
+
+        :returns: jpg version of the image as bytes
+        """
+        return self._repr_image("JPEG")
+
+    _repr_jpeg_ = _repr_jpg_
 
     @property
     def __array_interface__(self):
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -634,7 +634,8 @@ class Image:
         )
 
     def _repr_image(self, image_format):
-        """Helper function for iPython display hook
+        """Helper function for iPython display hook.
+
         :param image_format: Image format.
         :returns: image as bytes, saved into the given format.
         """
@@ -1122,7 +1123,6 @@ class Image:
            Available methods are :data:`Dither.NONE` or :data:`Dither.FLOYDSTEINBERG`
            (default).
         :returns: A new image
-
         """
 
         self.load()
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -633,17 +633,17 @@ class Image:
             )
         )
 
-    def _repr_image(self, format):
+    def _repr_image(self, image_format):
         """iPython display hook support
 
-        :param format: Image format.
+        :param image_format: Image format.
         :returns: image as bytes, saved into the given format.
         """
         b = io.BytesIO()
         try:
-            self.save(b, format)
+            self.save(b, image_format)
         except Exception as e:
-            msg = f"Could not save to {format} for display"
+            msg = f"Could not save to {image_format} for display"
             raise ValueError(msg) from e
         return b.getvalue()
 
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -654,14 +654,12 @@ class Image:
         """
         return self._repr_image("PNG")
 
-    def _repr_jpg_(self):
+    def _repr_jpeg_(self):
         """iPython display hook support for JPEG format.
 
         :returns: jpg version of the image as bytes
         """
         return self._repr_image("JPEG")
-
-    _repr_jpeg_ = _repr_jpg_
 
     @property
     def __array_interface__(self):
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -637,7 +637,7 @@ class Image:
         """iPython display hook support
 
         :param format: Image format.
-        :returns: png version of the image as bytes
+        :returns: image as bytes, saved into the given format.
         """
         b = io.BytesIO()
         try:
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -650,14 +650,14 @@ class Image:
     def _repr_png_(self):
         """iPython display hook support for PNG format.
 
-        :returns: png version of the image as bytes
+        :returns: PNG version of the image as bytes
         """
         return self._repr_image("PNG")
 
     def _repr_jpeg_(self):
         """iPython display hook support for JPEG format.
 
-        :returns: jpeg version of the image as bytes
+        :returns: JPEG version of the image as bytes
         """
         return self._repr_image("JPEG")
 
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -634,8 +634,7 @@ class Image:
         )
 
     def _repr_image(self, image_format):
-        """iPython display hook support
-
+        """Helper function for iPython display hook
         :param image_format: Image format.
         :returns: image as bytes, saved into the given format.
         """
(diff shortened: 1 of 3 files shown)


radarhere:
If I understand correctly, this is not just for passing an image object directly to IPython, like in https://github.com/python-pillow/Pillow/issues/1090#issuecomment-71880748
```python
from PIL import Image
im = Image.new("RGB", (512, 512), "blue")
im
```
and not just for [`IPython.display.display`](https://ipython.readthedocs.io/en/stable/api/generated/IPython.display.html#IPython.display.display), but specifically for [`IPython.display.display_jpeg`](https://ipython.readthedocs.io/en/stable/api/generated/IPython.display.html#IPython.display.display_jpeg), yes?

src/PIL/Image.py
```diff
@@ -633,19 +633,36 @@ def _repr_pretty_(self, p, cycle):
...
+        return self._repr_image("PNG")
+
+    def _repr_jpg_(self):
+        """iPython display hook support for JPEG format.
+
+        :returns: jpg version of the image as bytes
+        """
+        return self._repr_image("JPEG")
+
+    _repr_jpeg_ = _repr_jpg_
```
radarhere, line 664:
```suggestion
    def _repr_jpeg_(self):
        """iPython display hook support for JPEG format.

        :returns: jpg version of the image as bytes
        """
        return self._repr_image("JPEG")
```

Why is `_repr_jpg_` necessary? I don't see it mentioned in the [docs.](https://ipython.readthedocs.io/en/stable/api/generated/IPython.display.html#IPython.display.display)

src/PIL/Image.py
```diff
@@ -633,19 +633,36 @@ def _repr_pretty_(self, p, cycle):
             )
         )
 
-    def _repr_png_(self):
+    def _repr_image(self, format):
         """iPython display hook support
 
+        :param format: Image format.
         :returns: png version of the image as bytes
```
radarhere, line 640:
```suggestion
        :returns: image as bytes, saved into the given format.
```

n3011:
> If I understand correctly, this is not just for passing an image object directly to IPython, like in [#1090 (comment)](https://github.com/python-pillow/Pillow/issues/1090#issuecomment-71880748)
> 
> ```python
> from PIL import Image
> im = Image.new("RGB", (512, 512), "blue")
> im
> ```
> 
> and not just for [`IPython.display.display`](https://ipython.readthedocs.io/en/stable/api/generated/IPython.display.html#IPython.display.display), but specifically for [`IPython.display.display_jpeg`](https://ipython.readthedocs.io/en/stable/api/generated/IPython.display.html#IPython.display.display_jpeg), yes?

Yes this is `display_jpeg`.

src/PIL/Image.py
n3011: This method is required panel usage: https://panel.holoviz.org/reference/panes/JPG.html#parameters

radarhere: So `_repr_jpg` and `_repr_jpeg` are separate features, that you've grouped together in one PR because they are both related to JPEGs? One is for Panel, one is for IPython's `IPython.display.display_jpeg()`?

Could you post example code showing how you would use this?

src/PIL/Image.py
```diff
@@ -633,19 +633,36 @@ def _repr_pretty_(self, p, cycle):
             )
         )
 
-    def _repr_png_(self):
+    def _repr_image(self, format):
```
hugovk, line 636: Can we change `format` to something that doesn't shadow a builtin?

src/PIL/Image.py
n3011: Right, `_repr_jpeg_` is required for  iPython `display_jpeg` and `_repr_jpg_` is required for Panel pane `JPG`.

I would like to use these as following (without the addition of the current PR, these does not work):

1. iPython display_jpeg
```
from IPython.display import display_jpeg
from PIL import Image

im=Image.open(<path to jpg image>)
display_jpeg(im)
```

2. Panel JPEG pane
```
from pathlib import Path
import panel as pn
from PIL import Image
pn.extension()

imb = Path(<path to jpg image>).read_bytes()
im3 = Image.open(BytesIO(imb))
jpg_pane = pn.pane.JPG(im3, width=500)

jpg_pane
```

n3011: Okay, sure then, we can only keep `_repr_jpeg_`

src/PIL/Image.py
```diff
@@ -633,19 +633,34 @@ def _repr_pretty_(self, p, cycle):
             )
         )
 
-    def _repr_png_(self):
+    def _repr_image(self, image_format):
         """iPython display hook support
```
radarhere, line 638:
```suggestion
        """
```
I feel like the current version implies that iPython uses `_repr_image` directly, which is not the case.

src/PIL/Image.py
```diff
@@ -633,19 +633,34 @@ def _repr_pretty_(self, p, cycle):
...
+        """iPython display hook support for PNG format.
+
+        :returns: png version of the image as bytes
+        """
+        return self._repr_image("PNG")
+
+    def _repr_jpeg_(self):
+        """iPython display hook support for JPEG format.
+
+        :returns: jpg version of the image as bytes
```
radarhere, line 660:
```suggestion
        :returns: jpeg version of the image as bytes
```
Just for consistency with the surrounding text.

src/PIL/Image.py
nulano:
```suggestion
        """Helper function for iPython display hook
```
How about this to keep a helpful message without the misleading implication?

radarhere (approved):

src/PIL/Image.py
```diff
@@ -633,19 +633,34 @@ def _repr_pretty_(self, p, cycle):
...
 
+    def _repr_png_(self):
+        """iPython display hook support for PNG format.
+
+        :returns: png version of the image as bytes
+        """
+        return self._repr_image("PNG")
+
+    def _repr_jpeg_(self):
+        """iPython display hook support for JPEG format.
```
hugovk, line 658: Isn't this one for Panel, not iPython?

src/PIL/Image.py
```diff
@@ -633,19 +633,34 @@ def _repr_pretty_(self, p, cycle):
...
         except Exception as e:
-            msg = "Could not save to PNG for display"
+            msg = f"Could not save to {image_format} for display"
             raise ValueError(msg) from e
         return b.getvalue()
 
+    def _repr_png_(self):
+        """iPython display hook support for PNG format.
+
+        :returns: png version of the image as bytes
```
hugovk, line 653:
```suggestion
        :returns: PNG version of the image as bytes
```

src/PIL/Image.py
```diff
@@ -633,19 +633,34 @@ def _repr_pretty_(self, p, cycle):
...
+        """
+        return self._repr_image("PNG")
+
+    def _repr_jpeg_(self):
+        """iPython display hook support for JPEG format.
+
+        :returns: jpeg version of the image as bytes
```
hugovk, line 660:
```suggestion
        :returns: JPEG version of the image as bytes
```

hugovk (commented):
Please could you also mention this in the release notes?

https://github.com/python-pillow/Pillow/blob/main/docs/releasenotes/10.0.0.rst

src/PIL/Image.py
radarhere: `_repr_jpg_`, now removed, [was for Panel](https://panel.holoviz.org/reference/panes/JPG.html#parameters). `_repr_jpeg_` [is for iPython's](https://ipython.readthedocs.io/en/stable/config/integrating.html#custom-methods) [IPython.display.display_jpeg](https://ipython.readthedocs.io/en/stable/api/generated/IPython.display.html#IPython.display.display_jpeg).

radarhere:
Release notes have now been added - https://pillow--7135.org.readthedocs.build/en/7135/releasenotes/10.0.0.html#support-display-jpeg-in-ipython

smason:
I'm wondering what the purpose of this change is.  I'm interested to know if it's mainly to support calling `display_jpeg` on an image?

If so, this seems to be having unintended consequences as I'm now getting both a PNG and JPEG encoded image saved in each cell.  For example, a cell I just created includes:
```none
     "data": {
      "image/jpeg": "/9j/4AAQSkZJRgABAQAAAQABAAD/2wBDAAgGBgcGBQgHBwcJCQgKDBQNDAsLDBkSEw8UHRofHh0aHBwgJC4nICIsI>
      "image/png": "iVBORw0KGgoAAAANSUhEUgAAC7kAAAu5CAAAAABw4BI+AAEAAElEQVR4AQBSgK1/AdfANnmuHKdSoc6W0wL98OSB+b>
      "text/plain": [
       "<PIL.Image.Image image mode=L size=3001x3001>"
      ]
     },
```
which might well double resulting `.ipynb` file sizes, depending on how either encoder deals with the image.

radarhere:
> I'm interested to know if it's mainly to support calling `display_jpeg` on an image?

Yes. That is what the [original poster was after](https://github.com/python-pillow/Pillow/pull/7135#issuecomment-1535651028), and that is what we have [announced](https://pillow--7135.org.readthedocs.build/en/7135/releasenotes/10.0.0.html#support-display-jpeg-in-ipython).

Discussion of double images can continue in #7259

PR #7138, opened 2023-05-05
Updated ImagePath tolist() default

radarhere:
`ImagePath.Path.tolist()` is a C function that we have exposed in our public Python API. It has an argument `flat` that it uses like so -
https://github.com/python-pillow/Pillow/blob/b073a8962bb0bb182f92eda674152a20b6332e43/src/path.c#L431-L440

Looking at the documentation for `ImagePath.Path.tolist()`, https://pillow.readthedocs.io/en/stable/reference/ImagePath.html#PIL.ImagePath.PIL.ImagePath.Path.tolist, `flat` is documented as
> flat – By default, this function returns a list of 2-tuples [(x, y), …]. If this argument is True, it returns a flat list [x, y, …] instead.

Ok, so `flat` is a boolean. That's how it should be thought of, and how it is used.

Except the documentation at that link also the default for `flat` as `0`, which could cause a brief moment of confusion as to why the default is a different type.

Screenshot 2023-05-05 at 6 18 18 pm

The reason isn't not `False` would be because the code is in C, not Python, but I think it would be clearer if we described our Python API as having a default of `False`.


hugovk:
I added the documentation label, but then saw you added and removed it too! Feel free to remove it again if you like :)

radarhere:
It was initially documentation only, but then I technically changed a test too in the second commit, so... whatever works.

PR #7139, opened 2023-05-06
Do not use gnome-screenshot on Linux if X11 can be used instead

radarhere:
Alternative to #7143

https://github.com/python-pillow/Pillow/pull/7100#issue-1672406340
> although gnome-screenshot is widely present in Linux systems, in newer systems such as Ubuntu 20.04, there will be a full-screen animation effect when taking screenshots, and it is not easy to close. This feature may cause inconvenience to users in some tasks.

For this reason, [a later comment in the thread](https://github.com/python-pillow/Pillow/pull/7100#issuecomment-1537056956) requested that X11 be preferred over gnome-screenshot.

This PR modifies ImageGrab to not use gnome-screenshot if X11 is in use and Pillow has XCB support.

### DIFF
--- a/src/PIL/ImageGrab.py
+++ b/src/PIL/ImageGrab.py
@@ -61,7 +61,9 @@ def grab(bbox=None, include_layered_windows=False, all_screens=False, xdisplay=N
                 left, top, right, bottom = bbox
                 im = im.crop((left - x0, top - y0, right - x0, bottom - y0))
             return im
-        elif shutil.which("gnome-screenshot"):
+        elif not (
+            Image.core.HAVE_XCB and os.environ.get("XDG_SESSION_TYPE") == "x11"
+        ) and shutil.which("gnome-screenshot"):
             fh, filepath = tempfile.mkstemp(".png")
             os.close(fh)
             subprocess.call(["gnome-screenshot", "-f", filepath])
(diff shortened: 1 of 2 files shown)


src/PIL/ImageGrab.py
```diff
@@ -61,7 +61,9 @@ def grab(bbox=None, include_layered_windows=False, all_screens=False, xdisplay=N
                 left, top, right, bottom = bbox
                 im = im.crop((left - x0, top - y0, right - x0, bottom - y0))
             return im
-        elif shutil.which("gnome-screenshot"):
+        elif not (
+            Image.core.HAVE_XCB and os.environ.get("XDG_SESSION_TYPE") == "x11"
```
nulano, line 65: What if `XDG_SESSION_TYPE` is not set? Or alternatively, if the X11 connection is not configured properly (e.g. `DISPLAY` is not set)? The following is set on my Ubuntu 20.04 (in WSL), with the `DISPLAY` value being valid only when I do not forget to run my X server. I imagine the `DISPLAY` value could be invalid for other reasons as well.

```bash
$ set | grep DISPLAY
DISPLAY=:0
$ set | grep XDG
XDG_DATA_DIRS=/usr/local/share:/usr/share:/var/lib/snapd/desktop
    local -a dirs=(${BASH_COMPLETION_USER_DIR:-${XDG_DATA_HOME:-$HOME/.local/share}/bash-completion}/completions);
    for dir in ${XDG_DATA_DIRS:-/usr/local/share:/usr/share};
$
```

I think a better option is to use a try-except block instead to prefer X11 over gnome-screenshot: #7143

radarhere:
Closing in favour of #7143

PR #7143, opened 2023-05-08
Prefer screenshots using XCB over gnome-screenshot

nulano:
Alternative to #7139

From https://github.com/python-pillow/Pillow/pull/7100#issuecomment-1537056956

> > Is the problem just a user trying to ignore `gnome-screenshot` when it is installed in favour of `Image.core.grabscreen_x11`? If so, then `ImageGrab.grab(xdisplay=":0")` should be sufficient with the current version of Pillow.
> 
> [...] I have just tested setting `xdisplay=':0'`, and it ran on my end without any white flash. I believe this usage needs to be tested across different platforms. If the test is successful, it should be set as the default option. [...]

### DIFF
--- a/src/PIL/ImageGrab.py
+++ b/src/PIL/ImageGrab.py
@@ -66,10 +66,6 @@ def grab(bbox=None, include_layered_windows=False, all_screens=False, xdisplay=N
             msg = "Pillow was built without XCB support"
             raise OSError(msg)
         size, data = Image.core.grabscreen_x11(xdisplay)
-        im = Image.frombytes("RGB", size, data, "raw", "BGRX", size[0] * 4, 1)
-        if bbox:
-            im = im.crop(bbox)
-        return im
     except OSError:
         if (
             xdisplay is None
@@ -89,6 +85,11 @@ def grab(bbox=None, include_layered_windows=False, all_screens=False, xdisplay=N
             return im
         else:
             raise
+    else:
+        im = Image.frombytes("RGB", size, data, "raw", "BGRX", size[0] * 4, 1)
+        if bbox:
+            im = im.crop(bbox)
+        return im
 
 
 def grabclipboard():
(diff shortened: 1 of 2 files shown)


src/PIL/ImageGrab.py
```diff
@@ -61,7 +61,17 @@ def grab(bbox=None, include_layered_windows=False, all_screens=False, xdisplay=N
...
+    try:
+        if not Image.core.HAVE_XCB:
+            msg = "Pillow was built without XCB support"
+            raise OSError(msg)
+        size, data = Image.core.grabscreen_x11(xdisplay)
+        im = Image.frombytes("RGB", size, data, "raw", "BGRX", size[0] * 4, 1)
+        if bbox:
+            im = im.crop(bbox)
+        return im
+    except OSError:
```
nulano, line 73: The original error in #6312 was:
```
builtins.OSError: X get_image failed: error 8 (73, 0, 1315)
```

src/PIL/ImageGrab.py
```diff
@@ -61,7 +61,17 @@ def grab(bbox=None, include_layered_windows=False, all_screens=False, xdisplay=N
...
+        if xdisplay is None and shutil.which("gnome-screenshot"):
```
radarhere, line 74:
```suggestion
        if (
            sys.platform not in ("darwin", "win32")
            and xdisplay is None
            and shutil.which("gnome-screenshot")
        ):
```
To at least be clear in the code that we're expecting gnome-screenshot on macOS and Windows?

nulano: Sure. How about ordering them same as the checks above?

```suggestion
        if (
	            xdisplay is None
	            and sys.platform not in ("darwin", "win32")
	            and shutil.which("gnome-screenshot")
	        ):
```

radarhere: Sure.

radarhere (approved):

PR #7146, opened 2023-05-09
Fixed combining single duration across duplicate APNG frames

### LINKED ISSUE #7144: Animated PNG filters out similar frames without adjusting frame rate
### What did you do?

Attempted to make an animated PNG from a set of images.

To have a working animated PNG, with the right duration/frame rate.

### What actually happened?

The resulting animated PNG only has two frames, instead of 40 frames, and with a very high frame rate.

* OS: macOS 13.3.1
* Python: Python 3.11.3
* Pillow: I have checked out this PR: https://github.com/python-pillow/Pillow/pull/7123, which fixes a related issue.

This code creates a 40 frame animation, 20 frames of a red square and 20 frames of a blue square which run at 20 fps, which should result in a 2 second animation:

```python
#!/usr/bin/env python3

images = []
for i in range(20):
    img = Image.new("RGBA", (300, 300), (255, 255, 255, 255))

    draw = ImageDraw.Draw(img)
    draw.rectangle((100, 100, 200, 200), fill="red")

    img.save(f"frame1-{i}.png")
    images.append(img)

for i in range(20):
    img = Image.new("RGBA", (300, 300), (255, 255, 255, 255))

    draw = ImageDraw.Draw(img)
    draw.rectangle((100, 100, 200, 200), fill="blue")

    img.save(f"frame2-{i}.png")
    images.append(img)

# Should result in a 2 second animation
frame_rate = 20
frame_duration = 1000 / frame_rate

images[0].save(
    "test.png", save_all=True, append_images=images[1:], duration=frame_duration, loop=0
)
images[0].save(
    "test.gif", save_all=True, append_images=images[1:], duration=frame_duration, loop=0
)
```
GIF:
test

PNG:
test

These images actually only contain 2 frames, not 40.

One interesting thing to note is the frame rate of these images shown by `ffprobe`:
```console
$ ffprobe -v quiet -show_entries stream=r_frame_rate,time_base,duration  output.gif
[STREAM]
r_frame_rate=100/1
time_base=1/100
duration=2.100000
[/STREAM]

$ ffprobe -v quiet -show_entries stream=r_frame_rate,time_base,duration output.png
[STREAM]
r_frame_rate=100000/1
time_base=1/100000
duration=N/A
[/STREAM]
```

This is confusing to me, as while the GIF duration is roughly correct, the frame rate is not 100 fps.

radarhere:
Resolves #7144

When saving APNG files, we are testing that when duplicate frames are combined, the different durations are added together.
https://github.com/python-pillow/Pillow/blob/2a274a476088c610451370b9b78e7c63b7b31da3/Tests/test_file_apng.py#L433-L441

However, we aren't testing that the durations are combined when the `duration` keyword is a single number, rather than a list or a tuple. This issue found that this situation fails, and this PR corrects that.


gsingh93:
I've been using a local build with this patch for a few weeks now and it's been working well, hopefully we can get this merged soon :)

radarhere:
Be aware that even if it is merged, Pillow 10.0.0 is not due out until July 1.

gsingh93:
Yup, just looking forward to being able to remove the applying of these patches from my CI infrastructure and providing patching instructions to anyone who wants to run my code. I'm fine using `master` for now, other than that.

Tests/test_file_apng.py
```diff
@@ -440,6 +440,12 @@ def test_apng_save_duration_loop(tmp_path):
         assert im.n_frames == 1
         assert im.info.get("duration") == 750
 
+    # test removal of duplicated frames with a single duration
+    frame.save(test_file, save_all=True, append_images=[frame, frame], duration=500)
+    with Image.open(test_file) as im:
+        assert im.n_frames == 1
+        assert im.info.get("duration") == 1500
```
hugovk, line 447: `test_apng_save_duration_loop` is testing quite a few things that don't depend on each other, can we at least move this chunk into its own test case? 

For example, it could be useful to know if this bit passes, even if earlier bits fail.

radarhere:
Pillow 10.0.0 has now been released with this fix.

Issue #7147, opened 2023-05-09
ImageGrab.grabclipboard() keeps unnecessary temporary files around


stephen-huan:
### What did you do?

Call `PIL.ImageGrab.grabclipboard()` repeatedly. 

### What did you expect to happen?

A constant number of temp files to be created.

### What actually happened?

The number of files in `/tmp` keeps increasing.

If there not an actual image in the clipboard, e.g. only text, then
```shell
xclip -selection clipboard -target image/png -out
```
will error with the message
```text
Error: target image/png not available
```

while this error is caught on the latest Pillow (lines 146--148), the newly created tempfile is never removed.
https://github.com/python-pillow/Pillow/blob/ab3d0c071e60da904062d22f6b9a73e8fc6cdcb9/src/PIL/ImageGrab.py#L146-L151

This can cause tempfiles to accumulate with repeated calls of `grabclipboard()`.

I think `os.unlink(filepath)` should be called before `ChildProcessError` is raised to remove the now unnecessary file, i.e.
```python
        if err:
            os.unlink(filepath)
            msg = f"{args[0]} error: {err.strip().decode()}"
            raise ChildProcessError(msg)
        im = Image.open(filepath)
        im.load()
        os.unlink(filepath)
 ```

### What are your OS, Python and Pillow versions?

* OS: archlinux
* Python: 3.11.3
* Pillow: 9.5.0 (or commit [ab3d0c0](https://github.com/python-pillow/Pillow/commit/ab3d0c071e60da904062d22f6b9a73e8fc6cdcb9))

**Comment by stephen-huan** [comment]
with a bit of additional testing I think the suggested
```python
        if err:
            os.unlink(filepath)
            msg = f"{args[0]} error: {err.strip().decode()}"
            raise ChildProcessError(msg)
        im = Image.open(filepath)
        im.load()
        os.unlink(filepath)
```
should actually be something like
```python
        if err:
            os.unlink(filepath)
            msg = f"{args[0]} error: {err.strip().decode()}"
            raise ChildProcessError(msg)
        im = None
        try:
            im = Image.open(filepath)
            im.load()
        except UnidentifiedImageError:
            pass
        os.unlink(filepath)
```
as after the first paste `xclip` will no longer error with the message
```text
Error: target image/png not available
```
and simply write whatever text in the clipboard to `filepath`,
which causes an `UnidentifiedImageError`.

radarhere:
[...]

PR #7148, opened 2023-05-09
Remove temporary file when error is raised

radarhere:
Helps #7147 by ensuring that temporary files are still removed, even if the code returns early due to an error.


radarhere:
#7200 superseded this for ImageGrab, but there are similar changes made to other parts of the code.

hugovk (approved):

Issue #7163, opened 2023-05-15
Trying to download **stable** documentation as PDF


MMMZZZZ:
### What did you do?

* Go to [https://pillow.readthedocs.io/en/**stable**/](https://readthedocs.org/projects/pillow/downloads/)
* At the bottom left, hover over the book icon and click downloads
* A new page opens with download options: https://readthedocs.org/projects/pillow/downloads/

### What did you expect to happen?

* At least a download option for the documentation of the current stable release, v9.5.0
* Desirable: archive / download option for past stable releases

### What actually happened?

* I can only download the documentation for the not-yet-released v10, which [as far as I can see](https://github.com/python-pillow/Pillow/compare/9.5.x...main?diff=split#diff-92a33c6e5a9449d744005ca15e81ba9ff57ba09d5f555470c15b395f5c19b888) has already been modified compared to v9.5.x

### Further information

* I tried modifying the download links from `latest` to `stable` but that didn't work
* I looked up the download links in the Internet Archive but they don't have any captures between v9.0.0 and v10.0.0
* I checked the `docs` folder in the v9.5.x branch of the repository, no PDF there either. 

Thank you very much!

hugovk:
We were missing the PDF builds until https://github.com/python-pillow/Pillow/pull/7116 was merged 3 weeks ago, so it's currently only done for the `latest` build, It will also be available for the next `stable` build for Pillow 10 on 1st July.

In the meantime, you can find a `stable` build for 9.5.0 at https://github.com/python-pillow/Pillow/issues/7114#issuecomment-1522924805

MMMZZZZ:
Thank you very much for the quick and helpful reply!

PR #7164, opened 2023-05-15
Build only PDF in addition to default html

hugovk:

 * https://github.com/python-pillow/Pillow/issues/7114 requested PDF (or at least zipped HTML)
 * PDF was also requested in https://github.com/python-pillow/Pillow/issues/7163
 * https://github.com/python-pillow/Pillow/pull/7116 added PDF, epub and htmlzip
 * To save docs build time, only build PDF in addition to default HTML; skips epub and htmlzip

Build times:

* epub 34s
* htmlzip 16s
* pdf 25s

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,6 +1,6 @@
 version: 2
 
-formats: all
+formats: [pdf]
 
 build:
   os: ubuntu-22.04


aclark4life:
nice

PR #7171, opened 2023-05-20
Use --config-settings instead of deprecated --global-option

### LINKED ISSUE #7167: --global-option is deprecated
https://github.com/python-pillow/Pillow/actions/runs/4999985216/jobs/8956875899#step:7:23
> DEPRECATION: --build-option and --global-option are deprecated. pip 23.3 will enforce this behaviour change. A possible replacement is to use --config-settings. Discussion can be found at https://github.com/pypa/pip/issues/11859

While we could possibly just remove `make inplace`
https://github.com/python-pillow/Pillow/blob/599979caae111c22bd15f0103320bf74eb53d963/Makefile#L58-L59

and `--global-option="build_ext"`, we do still need something to replace our [build options](https://pillow.readthedocs.io/en/stable/installation.html#build-options).

The possible replacement of [`--config-settings`](https://peps.python.org/pep-0517/#config-settings) would involve switching from setuptools to build.

pip releases a new version [every 3 months](https://pip.pypa.io/en/stable/development/release-process/), so pip 23.3 is due out in October, [possibly before](https://github.com/pypa/pip/releases/tag/21.3) our scheduled October 15 release for Pillow 10.1. So it would be ideal to figure this out for Pillow 10.

radarhere:
Resolves #7167

https://github.com/python-pillow/Pillow/actions/runs/4999985216/jobs/8956875899#step:7:23
> DEPRECATION: --build-option and --global-option are deprecated. pip 23.3 will enforce this behaviour change. A possible replacement is to use --config-settings. Discussion can be found at https://github.com/pypa/pip/issues/11859

`--config-settings` is described at https://peps.python.org/pep-0517/#config-settings

So I think our [build options](https://pillow.readthedocs.io/en/stable/installation.html#build-options) should change from
```
python3 -m pip install --upgrade Pillow --global-option="build_ext" --global-option="--enable-[feature]"
```
to
```
python3 -m pip install --upgrade Pillow -C enable=[feature]
```

However, setuptools does not currently pass the config settings through to the backend - https://github.com/pypa/setuptools/issues/2491

Inspired by https://setuptools.pypa.io/en/latest/build_meta.html#dynamic-build-dependencies-and-other-build-meta-tweaks, my suggestion is to instead have a custom backend that translates the config-settings into a form that setuptools will understand. This should work until setuptools resolve their issue.

For pyproject.toml, I used the fallback setting for the `require` value - https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/#fallback-behaviour

### DIFF
--- a/.github/workflows/test-mingw.yml
+++ b/.github/workflows/test-mingw.yml
@@ -80,7 +80,7 @@ jobs:
           pushd depends && ./install_extra_test_images.sh && popd
 
       - name: Build Pillow
-        run: SETUPTOOLS_USE_DISTUTILS="stdlib" CFLAGS="-coverage" python3 -m pip install --global-option="build_ext" .
+        run: SETUPTOOLS_USE_DISTUTILS="stdlib" CFLAGS="-coverage" python3 -m pip install .
 
       - name: Test Pillow
         run: |
--- a/_custom_build/backend.py
+++ b/_custom_build/backend.py
@@ -1,0 +1,31 @@
+import sys
+
+from setuptools.build_meta import *  # noqa: F401, F403
+from setuptools.build_meta import _BuildMetaBackend
+
+
+class _CustomBuildMetaBackend(_BuildMetaBackend):
+    def run_setup(self, setup_script="setup.py"):
+        if self.config_settings:
+            flags = []
+            for key in ("enable", "disable", "vendor"):
+                settings = self.config_settings.get(key)
+                if settings:
+                    if not isinstance(settings, list):
+                        settings = [settings]
+                    for value in settings:
+                        flags.append("--" + key + "-" + value)
+            if self.config_settings.get("debug") == "true":
+                flags.append("--debug")
+            if flags:
+                sys.argv = sys.argv[:1] + ["build_ext"] + flags + sys.argv[1:]
+        return super().run_setup(setup_script)
+
+    def build_wheel(
+        self, wheel_directory, config_settings=None, metadata_directory=None
+    ):
+        self.config_settings = config_settings
+        return super().build_wheel(wheel_directory, config_settings, metadata_directory)
+
+
+build_wheel = _CustomBuildMetaBackend().build_wheel
--- a/_custom_build/backend.py
+++ b/_custom_build/backend.py
@@ -7,14 +7,37 @@ from setuptools.build_meta import _BuildMetaBackend
 class _CustomBuildMetaBackend(_BuildMetaBackend):
     def run_setup(self, setup_script="setup.py"):
         if self.config_settings:
-            flags = []
-            for key in ("enable", "disable", "vendor"):
+
+            def config_has(key, value):
                 settings = self.config_settings.get(key)
                 if settings:
                     if not isinstance(settings, list):
                         settings = [settings]
-                    for value in settings:
-                        flags.append("--" + key + "-" + value)
+                    return value in settings
+
+            flags = []
+            for dependency in (
+                "zlib",
+                "jpeg",
+                "tiff",
+                "freetype",
+                "raqm",
+                "lcms",
+                "webp",
+                "webpmux",
+                "jpeg2000",
+                "imagequant",
+                "xcb",
+            ):
+                if config_has(dependency, "enable"):
+                    flags.append("--enable-" + dependency)
+                elif config_has(dependency, "disable"):
+                    flags.append("--disable-" + dependency)
+            for dependency in ("raqm", "fribidi"):
+                if config_has(dependency, "vendor"):
+                    flags.append("--vendor-" + dependency)
+            if self.config_settings.get("platform-guessing") == "disable":
+                flags.append("--disable-platform-guessing")
             if self.config_settings.get("debug") == "true":
                 flags.append("--debug")
             if flags:
--- a/_custom_build/backend.py
+++ b/_custom_build/backend.py
@@ -1,10 +1,12 @@
 import sys
 
 from setuptools.build_meta import *  # noqa: F401, F403
-from setuptools.build_meta import _BuildMetaBackend
+from setuptools.build_meta import build_wheel
+
+backend_class = build_wheel.__self__.__class__
 
 
-class _CustomBuildMetaBackend(_BuildMetaBackend):
+class _CustomBuildMetaBackend(backend_class):
     def run_setup(self, setup_script="setup.py"):
         if self.config_settings:
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,4 @@
+[build-system]
+requires = ["setuptools >= 40.8.0", "wheel"]
+build-backend = "backend"
+backend-path = ["_custom_build"]
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -15,6 +15,7 @@ graft src
 graft depends
 graft winbuild
 graft docs
+graft _custom_build
 
 # build/src control detritus
 exclude .appveyor.yml
--- a/Makefile
+++ b/Makefile
@@ -46,17 +46,12 @@ help:
 	@echo "  docserve           run an HTTP server on the docs directory"
 	@echo "  html               make HTML docs"
 	@echo "  htmlview           open the index page built by the html target in your browser"
-	@echo "  inplace            make inplace extension"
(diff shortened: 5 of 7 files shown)


_custom_build/backend.py
```diff
@@ -0,0 +1,31 @@
+import sys
+
+from setuptools.build_meta import *  # noqa: F401, F403
```
hugovk, line 3: What's being imported here? Can we name them?

_custom_build/backend.py
```diff
@@ -0,0 +1,31 @@
...
+from setuptools.build_meta import _BuildMetaBackend
```
hugovk, line 4: Hmm, I'm not too keen on using an internal class from setuptools, especially as there have been quite big changes in the stable releases, and `_BuildMetaBackend` could change without notice.

nulano:
Would it be clearer to use `freetype=enable` instead of `enable=freetype`?

_custom_build/backend.py
radarhere: This started with https://setuptools.pypa.io/en/latest/build_meta.html#dynamic-build-dependencies-and-other-build-meta-tweaks, which says
> You can override any of the functions specified in [PEP 517](https://peps.python.org/pep-0517/#build-backend-interface), not only the ones responsible for gathering requirements. It is important to import * so that the hooks that you choose not to reimplement would be inherited from the setuptools’ backend automatically. This will also cover hooks that might be added in the future like the ones that [PEP 660](https://peps.python.org/pep-0660/) declares.

_custom_build/backend.py
radarhere: Rather than importing `_BuildMetaBackend`, would
```python
backend_class = build_wheel.__self__.__class__


class _CustomBuildMetaBackend(backend_class):
```
be better?

radarhere:
Ok, sure, I've swapped the order.

_custom_build/backend.py
hugovk: I guess it should be more stable, but really I'd like upstream setuptools to be fixed :)  But that's seems unlikely to be ready before 1st July, so I pinged the issue (https://github.com/pypa/setuptools/issues/2491#issuecomment-1589764230) to see if anyone there has a nicer solution. If not, let's go with this!

hugovk: People over at https://github.com/pypa/setuptools/issues/2491 like this, so let's go for it, thanks!

hugovk:
> Would it be clearer to use `freetype=enable` instead of `enable=freetype`?

I don't mind too much either way, but will note we're flipping the order from the current: ``--disable-zlib`` to ``-C zlib=disable``.

Trying to think of precedent, `pip install --help` has:

```
  --use-feature <feature>     Enable new functionality, that may be backward incompatible.
  --use-deprecated <feature>  Enable deprecated functionality, that will be removed in the future.
```

Which is more like the existing, but is a bit different. Are there any more direct comparisons?

hugovk (approved):

radarhere:
I wasn't able to find any comparisons.

Just for my personal thoughts, I think `freetype=enable` is better than `enable=freetype` because
- `enable=freetype` might make people think there should only be one `enable` argument, which wouldn't have been the plan
- it more clearly implies that you shouldn't try enabling and disabling freetype simultaneously

hugovk:
Fair enough, thank you!

_custom_build/backend.py
webknjaz: I requested that clarification to the setuptools docs this spring, having experimented with the in-tree backends for years, after my colleague at Ansible pointed out the problem — so I've gone ahead and documented it.
Ref: https://github.com/pypa/setuptools/pull/3893.

pyproject.toml
```diff
@@ -0,0 +1,4 @@
+[build-system]
+requires = ["setuptools >= 40.8.0", "wheel"]
```
webknjaz, line 2: @radarhere @hugovk `wheel` shouldn't actually be depended on: https://github.com/pypa/setuptools/pull/3056.

_custom_build/backend.py
```diff
@@ -0,0 +1,56 @@
...
+                    flags.append("--vendor-" + dependency)
+            if self.config_settings.get("platform-guessing") == "disable":
+                flags.append("--disable-platform-guessing")
+            if self.config_settings.get("debug") == "true":
+                flags.append("--debug")
+            if flags:
+                sys.argv = sys.argv[:1] + ["build_ext"] + flags + sys.argv[1:]
+        return super().run_setup(setup_script)
+
+    def build_wheel(
```
webknjaz, line 49: @hugovk @radarhere this omits `build_editable()` which is almost the same as `build_wheel()`. You'll likely want to wrap it as well.

.github/workflows/test-mingw.yml
```diff
@@ -80,7 +80,7 @@ jobs:
           pushd depends && ./install_extra_test_images.sh && popd
 
       - name: Build Pillow
-        run: SETUPTOOLS_USE_DISTUTILS="stdlib" CFLAGS="-coverage" python3 -m pip install --global-option="build_ext" .
+        run: SETUPTOOLS_USE_DISTUTILS="stdlib" CFLAGS="-coverage" python3 -m pip install .
```
webknjaz, line 83: I got to the conclusion that it's also nice to configure coverage  through the config setting. Here's an example of the interface I ended up with: https://yarl.aio-libs.org/en/latest/changes/#released-versions

pyproject.toml
nulano: This has since been removed in #7248.

webknjaz: Ah, I should've checked. Thanks!

_custom_build/backend.py
nulano: I've attempted to add `build_editable` in https://github.com/python-pillow/Pillow/commit/290a2d15b6066e66b797b985bdedf6986c271f41, but for some reason it is not working properly on Windows: https://github.com/nulano/Pillow/actions/runs/7308594440/job/19915362677#step:25:728

webknjaz: It looks like the build script calls something different from what's called on the main branch: I see "Running command Getting requirements to build editable" vs. "Running command Getting requirements to build wheel". The former is from your log, the latter is from the main branch.

nulano: Yes, I added https://github.com/python-pillow/Pillow/commit/2446fd8d7ca7d495ed19addefd28e2925c74e629 to test editable installing on GHA. For some reason, it seems the include paths are not passed in properly when building editable on Windows (specifically the OpenJpeg path is found and added for part of the build, but not all of it).

.github/workflows/test-mingw.yml
radarhere: We're already providing `make install-coverage`. I don't see the need to provide yet another way of enabling coverage.

_custom_build/backend.py
webknjaz: Oh, interesting, I didn't see where it's defined. It's often hard to distinguish outputs of different commands if they are squashed in the same CI step, so I tend to recommend putting each command in a separate step or using the corresponding shell's capabilities to improve such logging. I also recently saw a case where Powershell would proceed to execute commands after the previous ones fail.

Anyway, this may indicate dependency on some weird pre-existing state.

One way of exploring GHA jobs is https://github.com/marketplace/actions/debugging-with-tmate, by the way.

.github/workflows/test-mingw.yml
webknjaz: Yeah, I see. Just thought it's less integrated into the Python interfaces, but into tooling of external ecosystems.

_custom_build/backend.py
nulano: Sorry, I should have mentioned that it also fails for me locally.

It's the first command in that step that is failing, the second one just runs a simple test.
The path is added in `setup.py` in the `build_extensions` method here: https://github.com/python-pillow/Pillow/blob/41e45b5f95c5184b8b4c4712e5614bf468331e05/setup.py#L687-L691
From the GHA log, it looks almost as if it first does a regular build, then a second build again skipping that method, but I'm not sure why or where to look.

Yes, I have used tmate before.

nulano: I've created #7658 for this.

PR #7174, opened 2023-05-21
Use "not in"

radarhere:
Extracting a formatting change from #6966 that can be made independently of ruff, as we've preferred this style [in the past.](https://github.com/python-pillow/Pillow/commit/eacbd7b04abac95a2f84a3f091642530910f450f#diff-74d414aaf6e27f4f300028b1d09db57e995195807399fd96f63a207bc4f29ab9L63)

### DIFF
--- a/src/PIL/GdImageFile.py
+++ b/src/PIL/GdImageFile.py
@@ -47,7 +47,7 @@ class GdImageFile(ImageFile.ImageFile):
         # Header
         s = self.fp.read(1037)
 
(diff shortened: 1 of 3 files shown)


hugovk (approved):
Perhaps add a `Co-authored-by` trailer to give the original author credit?

radarhere:
Interesting, I've never done that manually before.

Ok, I've updated the commit.

PR #6926, opened 2023-02-02
Perform font fallback

nulano:
Fixes #4808.

Add a new type, `ImageFont.FreeTypeFontFamily(font1, font2, ..., layout_engine=layout_engine)`, that can be used with `ImageDraw.text*(...)` functions performing font fallback. Font fallback is done per cluster with Raqm layout (similar to Chromium) and per codepoint with basic layout.

This PR is far from complete, several TODOs:
* [ ] Font families have only a minimal API so far, e.g. retrieving metrics or setting font variations should be supported
* [ ] Maybe add a wrapper similar to `ImageFont.truetype(...)`, perhaps `ImageFont.truetype_family(...)`?
* [ ] Lots of tests
* [ ] Documentation

I would like to get some feedback, both on the API and the implementation, before working on the TODOs above.
A dev build for Windows is available here: https://github.com/nulano/Pillow/suites/10741759296/artifacts/538936223

<details>
<summary>A few examples (click to expand):</summary>

All examples use this helper block:
```python
from PIL import Image, ImageDraw, ImageFont

im = Image.new("RGBA", (500, 200), "white")
draw = ImageDraw.Draw(im)
def line(y, string, font, name, **kwargs):
  draw.text((10, y), name, fill="black", font=font, **kwargs)
  draw.text((300, y), string, fill="black", font=font, **kwargs)

example()

im.show()
```

Combining Latin, symbols, and an emoji:
```python
def example():
  s = "smile ⊗ 😀"

  times = ImageFont.truetype("times.ttf", 24)
  segoe_ui_emoji = ImageFont.truetype("seguiemj.ttf", 24)
  segoe_ui_symbol = ImageFont.truetype("seguisym.ttf", 24)
  family = ImageFont.FreeTypeFontFamily(times, segoe_ui_emoji, segoe_ui_symbol)

  line(30, s, times, "Times New Roman", anchor="ls", embedded_color=True)
  line(80, s, segoe_ui_emoji, "Segoe UI Emoji", anchor="ls", embedded_color=True)
  line(130, s, segoe_ui_symbol, "Segoe UI Symbol", anchor="ls", embedded_color=True)
  line(180, s, family, "Font Family", anchor="ls", embedded_color=True)
```

fallback_emoji

Combining Arabic, Greek, Latin, and a symbol:
```python
def example():
  s = "ية↦α,abc"

[...]

PR #7177, opened 2023-05-21
Corrected error code

radarhere:
Extracting a change from the draft PR #6926

https://github.com/python-pillow/Pillow/blob/599979caae111c22bd15f0103320bf74eb53d963/src/_imagingft.c#L168-L170

On first glance, the PR simply stops this value from being hardcoded, by replacing it with `FT_Err_Out_Of_Memory`, as per https://freetype.org/freetype2/docs/reference/ft2-error_code_values.html.

However, 65 is actually not the out of memory error code. 65 is 0x41, and according to that FreeType doc, 0x40 is "out of memory", while 0x41 is "unlisted object". So this change is not just stylistic, but also corrects the error code.


hugovk (approved):
Good to merge, maybe add a `Co-authored-by`?

radarhere:
Ok, I've updated the commit.

PR #7180, opened 2023-05-23
Removed unused INT64 definition

radarhere:
See https://github.com/python-pillow/Pillow/pull/6645#discussion_r1201417169


Yay295:
I noticed these lines too:
https://github.com/python-pillow/Pillow/blob/b8719033ca91ef57f58128d32df675457431bbce/src/libImaging/ImPlatform.h#L74-L76

but actually `int64_t` *is* used in QuantTypes.h, QuantOctree.c, and Draw.c, all of which import `<stdint.h>`.

PR #7190, opened 2023-05-27
Specify build backend via pyproject.toml

hugovk:
For https://github.com/python-pillow/Pillow/issues/6941.

https://github.com/python-pillow/Pillow/pull/7188 added setuptools to the CI installation for Python 3.12, because 3.12 dropped setuptools:

* https://docs.python.org/3.12/whatsnew/3.12.html#removed

This is actually a bit bigger problem than the CI because users could theoretically have a system that does not have setuptools installed.

---

With pyproject.toml, we can specify the "build backend". In our case, this is setuptools. (The "build frontend" is what reads this file and does the install, often pip.)

So when installing, the frontend will install the required backend, so we don't need to worry whether users have already installed it. We can also specify a minimum version. I just went with the current latest, and it also seems we can remove the extra `PILLOW_VERSION` double quotes hack we needed for Windows on Python 3.8.

It's also possible to move more config from `setup.cfg` to `pyproject.toml`, but I ran into some problems when moving `[options]` and `[options.extras_require]` over, so let's leave those for later. We can also move the tools config, but that can also wait to keep the diff smaller here and the PR focused.

---

About `winbuild/build_prepare.py`: it had a direct invocation of `setup.py` which is a no-go because it assumes setuptools is already installed. So we need `pip install` instead (reminder of https://blog.ganssle.io/tag/setuptools.html).

I went with `--global-option` for now, to follow the current approach elsewhere, but that will need changing as part of https://github.com/python-pillow/Pillow/issues/7167.

I did hit one blocker though: `test-windows.yml` builds a wheel, it calls:

```
winbuild\\build\\build_pillow.cmd --disable-imagequant bdist_wheel
```

Before, that was handled by this: 

```
        r'"{python_dir}\{python_exe}" setup.py build_ext --vendor-raqm --vendor-fribidi %*',  # noqa: E501
```

Which wrote out a command that was called like:

```
2023-04-28T14:37:47.1413330Z D:\a\Pillow\Pillow>"C:\hostedtoolcache\windows\Python\3.8.10\x86\python.exe" setup.py build_ext --vendor-raqm --vendor-fribidi --disable-imagequant bdist_wheel 
```

Now, those two args `--disable-imagequant bdist_wheel` each need wrapping as `--global-option="--disable-imagequant" --global-option="bdist_wheel"`

But this PR does:

```python
        r'"{python_dir}\{python_exe}" -m pip install --upgrade pip',
        r'"{python_dir}\{python_exe}" -m pip install . '
        r'--global-option="--vendor-raqm" '
        r'--global-option="--vendor-fribidi" '
        r'--global-option="%*"',
```

And those two args were both passed into `%*` meaning they both were wrapped in a single `global-option` which doesn't work, so it doesn't build the wheel:

```
2023-05-27T14:56:57.1800310Z D:\a\Pillow\Pillow>"C:\hostedtoolcache\windows\Python\3.8.10\x86\python.exe" -m pip install . --global-option="--vendor-raqm" --global-option="--vendor-fribidi" --global-option="--disable-imagequant bdist_wheel" 
```

Any ideas?

Maybe replace the `build_pillow.cmd` call entirely, and use something like https://pypi.org/project/build/ to call `python -m build --wheel`?

### DIFF
--- a/.github/workflows/test-windows.yml
+++ b/.github/workflows/test-windows.yml
@@ -65,8 +65,8 @@ jobs:
     - name: Print build system information
       run: python3 .github/workflows/system-info.py
 
-    - name: python3 -m pip install setuptools wheel pytest pytest-cov pytest-timeout defusedxml
-      run: python3 -m pip install setuptools wheel pytest pytest-cov pytest-timeout defusedxml
+    - name: python3 -m pip install wheel pytest pytest-cov pytest-timeout defusedxml
+      run: python3 -m pip install wheel pytest pytest-cov pytest-timeout defusedxml
 
     - name: Install dependencies
       id: install
--- a/.github/workflows/test-windows.yml
+++ b/.github/workflows/test-windows.yml
@@ -166,8 +166,8 @@ jobs:
     - name: Build Pillow
       run: |
         $FLAGS=""
-        if ('${{ github.event_name }}' -ne 'pull_request') { $FLAGS="--disable-imagequant" }
-        & winbuild\build\build_pillow.cmd $FLAGS install
+        if ('${{ github.event_name }}' -ne 'pull_request') { $FLAGS='--global-option="--disable-imagequant"' }
+        & winbuild\build\build_pillow.cmd $FLAGS --global-option="install"
         & $env:pythonLocation\python.exe selftest.py --installed
       shell: pwsh
 
@@ -231,7 +231,7 @@ jobs:
           )
         )
         for /f "tokens=3 delims=/" %%a in ("${{ github.ref }}") do echo dist=dist-%%a >> %GITHUB_OUTPUT%
-        winbuild\\build\\build_pillow.cmd --disable-imagequant bdist_wheel
+        winbuild\\build\\build_pillow.cmd --global-option="--disable-imagequant" --global-option="bdist_wheel"
       shell: cmd
 
     - name: Upload wheel
--- a/winbuild/build_prepare.py
+++ b/winbuild/build_prepare.py
@@ -569,6 +569,7 @@ def build_pillow():
         *prefs["header"],
         cmd_set("DISTUTILS_USE_SDK", "1"),  # use same compiler to build Pillow
         cmd_set("py_vcruntime_redist", "true"),  # always use /MD, never /MT
+        r'"{python_dir}\{python_exe}" -m pip install --upgrade pip',
         r'"{python_dir}\{python_exe}" -m pip install . '
         r'--global-option="--vendor-raqm" '
         r'--global-option="--vendor-fribidi" '
--- a/winbuild/build_prepare.py
+++ b/winbuild/build_prepare.py
@@ -569,7 +569,10 @@ def build_pillow():
         *prefs["header"],
         cmd_set("DISTUTILS_USE_SDK", "1"),  # use same compiler to build Pillow
         cmd_set("py_vcruntime_redist", "true"),  # always use /MD, never /MT
-        r'"{python_dir}\{python_exe}" setup.py build_ext --vendor-raqm --vendor-fribidi %*',  # noqa: E501
+        r'"{python_dir}\{python_exe}" -m pip install . '
+        r'--global-option="--vendor-raqm" '
+        r'--global-option="--vendor-fribidi" '
+        r'--global-option="%*"',
     ]
 
     write_script("build_pillow.cmd", lines)
--- a/winbuild/build_prepare.py
+++ b/winbuild/build_prepare.py
@@ -569,7 +569,6 @@ def build_pillow():
         *prefs["header"],
         cmd_set("DISTUTILS_USE_SDK", "1"),  # use same compiler to build Pillow
         cmd_set("py_vcruntime_redist", "true"),  # always use /MD, never /MT
-        r'"{python_dir}\{python_exe}" -m pip install --upgrade pip',
         r'"{python_dir}\{python_exe}" -m pip install . '
         r'--global-option="--vendor-raqm" '
         r'--global-option="--vendor-fribidi" '
--- a/winbuild/build_prepare.py
+++ b/winbuild/build_prepare.py
@@ -572,7 +572,7 @@ def build_pillow():
         r'"{python_dir}\{python_exe}" -m pip install . '
         r'--global-option="--vendor-raqm" '
         r'--global-option="--vendor-fribidi" '
-        r'--global-option="%*"',
+        r"%*",
     ]
 
     write_script("build_pillow.cmd", lines)
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -72,7 +72,7 @@ before_deploy:
   - cd c:\pillow
   - '%PYTHON%\%EXECUTABLE% -m pip install wheel'
   - cd c:\pillow\winbuild\
-  - c:\pillow\winbuild\build\build_pillow.cmd bdist_wheel
+  - c:\pillow\winbuild\build\build_pillow.cmd --global-option="bdist_wheel"
   - cd c:\pillow
   - ps: Get-ChildItem .\dist\*.* | % { Push-AppveyorArtifact $_.FullName -FileName $_.Name }
 
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -20,6 +20,7 @@ environment:
 
 install:
 - '%PYTHON%\%EXECUTABLE% --version'
+- '%PYTHON%\%EXECUTABLE% -m pip install --upgrade pip'
 - curl -fsSL -o pillow-depends.zip https://github.com/python-pillow/pillow-depends/archive/main.zip
 - curl -fsSL -o pillow-test-images.zip https://github.com/python-pillow/test-images/archive/main.zip
 - 7z x pillow-depends.zip -oc:\
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -58,7 +58,7 @@ repos:
       - id: sphinx-lint
 
   - repo: https://github.com/tox-dev/pyproject-fmt
-    rev: 0.11.2
+    rev: 0.12.1
     hooks:
       - id: pyproject-fmt
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -49,12 +49,23 @@ repos:
     hooks:
       - id: check-merge-conflict
       - id: check-json
+      - id: check-toml
       - id: check-yaml
 
   - repo: https://github.com/sphinx-contrib/sphinx-lint
     rev: v0.6.7
     hooks:
       - id: sphinx-lint
+
+  - repo: https://github.com/tox-dev/pyproject-fmt
+    rev: 0.11.2
+    hooks:
+      - id: pyproject-fmt
+
+  - repo: https://github.com/abravalheri/validate-pyproject
+    rev: v0.13
+    hooks:
+      - id: validate-pyproject
 
   - repo: https://github.com/tox-dev/tox-ini-fmt
     rev: 1.3.0
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,5 @@
+[build-system]
+build-backend = "setuptools.build_meta"
+requires = [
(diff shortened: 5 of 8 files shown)


aclark4life:
@hugovk Is there any other reasonable option besides jumping ship to [build](https://pypa-build.readthedocs.io/en/latest/)? Even if you could fix the wheel build with `build_pillow.cmd` the entire `build_prepare.py` is hard to follow like `setup.py` … not sure how much of that can be cleaned up now, but that's an attractive goal at least.

winbuild/build_prepare.py
```diff
@@ -569,7 +569,11 @@ def build_pillow():
         *prefs["header"],
         cmd_set("DISTUTILS_USE_SDK", "1"),  # use same compiler to build Pillow
         cmd_set("py_vcruntime_redist", "true"),  # always use /MD, never /MT
-        r'"{python_dir}\{python_exe}" setup.py build_ext --vendor-raqm --vendor-fribidi %*',  # noqa: E501
+        r'"{python_dir}\{python_exe}" -m pip install --upgrade pip',
```
nulano, line 572: Is it really necessary to upgrade pip here?
I imagine it would be preferable to upgrade pip before calling this script instead (and recommend it in the README) to (slightly) speed up local development.

winbuild/build_prepare.py
```diff
@@ -569,7 +569,11 @@ def build_pillow():
...
+        r'"{python_dir}\{python_exe}" -m pip install . '
+        r'--global-option="--vendor-raqm" '
+        r'--global-option="--vendor-fribidi" '
+        r'--global-option="%*"',
```
nulano, line 576: A quick and dirty solution to get this unblocked would be to instead require the calling script to wrap arguments using `--global-option=...`.
```suggestion
        r'%*',
```

.github/workflows/test-windows.yml
```diff
@@ -65,8 +65,8 @@ jobs:
     - name: Print build system information
       run: python3 .github/workflows/system-info.py
 
-    - name: python3 -m pip install setuptools wheel pytest pytest-cov pytest-timeout defusedxml
-      run: python3 -m pip install setuptools wheel pytest pytest-cov pytest-timeout defusedxml
+    - name: python3 -m pip install wheel pytest pytest-cov pytest-timeout defusedxml
+      run: python3 -m pip install wheel pytest pytest-cov pytest-timeout defusedxml
```
nulano, line 69:
```suggestion
    - name: Install Python dependencies
      run: |
        python3 -m pip install --upgrade pip
        python3 -m pip install wheel pytest pytest-cov pytest-timeout defusedxml
```

To avoid adding the pip upgrade to the build script. (AppVeyor might need the same change.)

nulano (commented):
> Maybe replace the build_pillow.cmd call entirely, and use something like https://pypi.org/project/build/ to call python -m build --wheel?

Does this build a wheel from the previously compiled Pillow without invoking the compiler?
If so, that sounds like a great solution!
(I expect the `--disable-imagequant` flag will no longer be needed when building the wheel if that is the case.)

Note that I have a ton of work this week so I cannot look into this in detail until next week.

hugovk:
> @hugovk Is there any other reasonable option besides jumping ship to [build](https://pypa-build.readthedocs.io/en/latest/)? Even if you could fix the wheel build with `build_pillow.cmd` the entire `build_prepare.py` is hard to follow like `setup.py` … not sure how much of that can be cleaned up now, but that's an attractive goal at least.

Switching to `build` might be the simpler choice - creating wheels from existing binaries that have already been built using `build_prepare.py`, rather than calling it a second time.

Otherwise, we can keep the current `build_prepare.py` system, it's a matter of getting the existing parameters passed through properly, shouldn't be too hard. I don't have Windows so it's a bit slow to test via CI pushes :)

hugovk:
> > Maybe replace the build_pillow.cmd call entirely, and use something like [pypi.org/project/build](https://pypi.org/project/build/) to call python -m build --wheel?
> 
> Does this build a wheel from the previously compiled Pillow without invoking the compiler? If so, that sounds like a great solution! (I expect the `--disable-imagequant` flag will no longer be needed when building the wheel if that is the case.)
> 
> Edit to clarify: The first `build_pillow.cmd` call when building Pillow is needed to make sure the correct compiler is used. The second one is needed because setuptools requires passing the same parameters to avoid rebuilding Pillow before making a wheel. If the second call can be avoided in some way, it would greatly simplify building wheels from an already built and tested build. However, I can't quite understand how build works in the time I have available at the moment, so it is possible I am misunderstanding it.

No, it actually does rebuild (hence the name!) but does so in an isolated environment. It's a "build frontend" so it can do a full build, like pip.

We might want to replace `build_pillow.cmd` with `build` at some point, but I'm less clear on the full Windows build process so it's out of scope here and we should probably to stick with `build_pillow.cmd` so it can do the correct compiler selection, and fix passing the parameters.

> Note that I have a ton of work this week so I cannot look into this in detail until next week.

Thanks! We're not in a rush for this. The July release will have beta wheels for Python 3.12, it would be nice this in there to get feedback before releasing with full support for 3.12 in the October release. Otherwise, it can likely wait until October.

winbuild/build_prepare.py
hugovk: Without any upgrade it was failing (for AppVeyor only) like:

<details>
<summary>Details</summary>



```

C:\pillow>"C:/Python38-x64\python.exe" -m pip install . --global-option="--vendor-raqm" --global-option="--vendor-fribidi" --global-option="install" 
Processing c:\pillow
  Installing build dependencies: started
  Installing build dependencies: finished with status 'error'
c:\pillow\winbuild\build\build_pillow.cmd : C:\Python38-x64\lib\site-packages\pip\_internal\commands\install.py:283: UserWarning: Disabling all use of wheels due to the use of --build-options / --global-options / 
--install-options.
At line:1 char:1
+ c:\pillow\winbuild\build\build_pillow.cmd install
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (C:\Python38-x64...nstall-options.:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
 
  cmdoptions.check_install_build_global(options)
  ERROR: Command errored out with exit status 2:
   command: 'C:\Python38-x64\python.exe' 'C:\Python38-x64\lib\site-packages\pip' install --ignore-installed --no-user --prefix 'C:\Users\appveyor\AppData\Local\Temp\1\pip-build-env-cgb3zz5u\overlay' --no-warn-script-location 
--no-binary :all: --only-binary :none: -i https://pypi.org/simple -- 'setuptools>=67.8'
       cwd: None
  Complete output (44 lines):
  Collecting setuptools>=67.8
    Using cached https://files.pythonhosted.org/packages/03/20/630783571e76e5fa5f3e9f[2939](https://ci.appveyor.com/project/hugovk/Pillow/build/job/y6vupdxep0yqi8v5#L2939)8ca3ace377207b8196b54e0ffdf09f12c1/setuptools-67.8.0.tar.gz
    Getting requirements to build wheel: started
    Getting requirements to build wheel: finished with status 'done'
  ERROR: Exception:
    File "C:\Python38-x64\lib\site-packages\pip\_internal\cli\base_command.py", line 153, in _main
      status = self.run(options, args)
    File "C:\Python38-x64\lib\site-packages\pip\_internal\commands\install.py", line 382, in run
      resolver.resolve(requirement_set)
    File "C:\Python38-x64\lib\site-packages\pip\_internal\legacy_resolve.py", line 201, in resolve
      self._resolve_one(requirement_set, req)
    File "C:\Python38-x64\lib\site-packages\pip\_internal\legacy_resolve.py", line 365, in _resolve_one
      abstract_dist = self._get_abstract_dist_for(req_to_install)
    File "C:\Python38-x64\lib\site-packages\pip\_internal\legacy_resolve.py", line 312, in _get_abstract_dist_for
      abstract_dist = self.preparer.prepare_linked_requirement(
    File "C:\Python38-x64\lib\site-packages\pip\_internal\operations\prepare.py", line 223, in prepare_linked_requirement
      abstract_dist = _get_prepared_distribution(
    File "C:\Python38-x64\lib\site-packages\pip\_internal\operations\prepare.py", line 49, in _get_prepared_distribution
      abstract_dist.prepare_distribution_metadata(finder, build_isolation)
    File "C:\Python38-x64\lib\site-packages\pip\_internal\distributions\source\legacy.py", line 37, in prepare_distribution_metadata
      self._setup_isolation(finder)
    File "C:\Python38-x64\lib\site-packages\pip\_internal\distributions\source\legacy.py", line 90, in _setup_isolation
      reqs = backend.get_requires_for_build_wheel()
    File "C:\Python38-x64\lib\site-packages\pip\_vendor\pep517\wrappers.py", line 151, in get_requires_for_build_wheel
      return self._call_hook('get_requires_for_build_wheel', {
    File "C:\Python38-x64\lib\site-packages\pip\_vendor\pep517\wrappers.py", line 255, in _call_hook
      raise BackendUnavailable(data.get('traceback', ''))
  pip._vendor.pep517.wrappers.BackendUnavailable: Traceback (most recent call last):
    File "C:\Python38-x64\lib\site-packages\pip\_vendor\pep517\_in_process.py", line 63, in _build_backend
      obj = import_module(mod_path)
    File "C:\Python38-x64\lib\importlib\__init__.py", line 127, in import_module
      return _bootstrap._gcd_import(name[level:], package, level)
    File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
    File "<frozen importlib._bootstrap>", line 991, in _find_and_load
    File "<frozen importlib._bootstrap>", line 961, in _find_and_load_unlocked
    File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
    File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
    File "<frozen importlib._bootstrap>", line 991, in _find_and_load
    File "<frozen importlib._bootstrap>", line 973, in _find_and_load_unlocked
  ModuleNotFoundError: No module named 'setuptools'
  
  WARNING: You are using pip version 19.3.1; however, version 23.1.2 is available.
  You should consider upgrading via the 'python -m pip install --upgrade pip' command.
  ----------------------------------------
ERROR: Command errored out with exit status 2: 'C:\Python38-x64\python.exe' 'C:\Python38-x64\lib\site-packages\pip' install --ignore-installed --no-user --prefix 
'C:\Users\appveyor\AppData\Local\Temp\1\pip-build-env-cgb3zz5u\overlay' --no-warn-script-location --no-binary :all: --only-binary :none: -i https://pypi.org/simple -- 'setuptools>=67.8' Check the logs for full command output.
WARNING: You are using pip version 19.3.1; however, version 23.1.2 is available.
You should consider upgrading via the 'python -m pip install --upgrade pip' command.
cd c:\pillow
%PYTHON%\%EXECUTABLE% selftest.py --installed
  File "selftest.py", line 6, in <module>
    from PIL import Image, features
ModuleNotFoundError: No module named 'PIL'
Command exited with code 1
```


</details>

https://ci.appveyor.com/project/hugovk/Pillow/build/job/y6vupdxep0yqi8v5

But we can move it `.appveyor.yml` instead. Updated!

winbuild/build_prepare.py
hugovk: Ah yes, updated!

It's still not building the wheel, will have to check that later...

nulano: You mean for the PR? Wheel building is disabled on PRs, but a wheel was built for the push in your repo: https://github.com/hugovk/Pillow/actions/runs/5105935733/jobs/9177831445

hugovk: Aha! There we go!

radarhere:
> We're not in a rush for this. The July release will have beta wheels for Python 3.12, it would be nice this in there to get feedback before releasing with full support for 3.12 in the October release. Otherwise, it can likely wait until October.

The essential part of this - adding pyproject.toml - is part of #7171, which we do plan to release in July.

So I would think this should also be part of that release?

nulano:
With #7171 this fails to pass the `--global-option` values to `setup.py`.
I'm working on updating this to use the `-C raqm=vendor` format instead.

PR #7192, opened 2023-05-29
Replaced deprecated Py_FileSystemDefaultEncoding for Python >= 3.12

radarhere:
Python 3.12 reports a deprecation - https://github.com/python-pillow/Pillow/actions/runs/5105776867/jobs/9177570491#step:7:234
> warning: 'Py_FileSystemDefaultEncoding' is deprecated

https://docs.python.org/3.12/whatsnew/3.12.html#id4 instructs us to replace it with `PyConfig.filesystem_encoding`.

Using https://docs.python.org/3.12/c-api/init_config.html#example as a basis, I made that change.

However, I found [this](https://github.com/radarhere/Pillow/commit/cda44e2408d64663ac71cb8f9cc46bef1b298611) [failed on PyPy](https://github.com/radarhere/Pillow/actions/runs/5097051329/jobs/9163321558#step:7:232), with
> error: use of undeclared identifier 'PyConfig'

I asked in https://foss.heptapod.net/pypy/pypy/-/issues/3946, and PyPy has chosen not to implement PyConfig yet. So I've switched to the new code only for Python 3.12 and later, and we can see what happens once PyPy does support Python 3.12.

### DIFF
--- a/src/_imagingft.c
+++ b/src/_imagingft.c
@@ -132,6 +132,27 @@ getfont(PyObject *self_, PyObject *args, PyObject *kw) {
         return NULL;
     }
 
+#if PY_MAJOR_VERSION > 3 || PY_MINOR_VERSION > 11
+    PyConfig config;
+    PyConfig_InitPythonConfig(&config);
+    if (!PyArg_ParseTupleAndKeywords(
+            args,
+            kw,
+            "etf|nsy#n",
+            kwlist,
+            config.filesystem_encoding,
+            &filename,
+            &size,
+            &index,
+            &encoding,
+            &font_bytes,
+            &font_bytes_size,
+            &layout_engine)) {
+        PyConfig_Clear(&config);
+        return NULL;
+    }
+    PyConfig_Clear(&config);
+#else
     if (!PyArg_ParseTupleAndKeywords(
             args,
             kw,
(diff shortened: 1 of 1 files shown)


src/_imagingft.c
```diff
@@ -132,6 +132,27 @@ getfont(PyObject *self_, PyObject *args, PyObject *kw) {
         return NULL;
     }
 
+#if PY_MAJOR_VERSION > 3 || PY_MINOR_VERSION > 11
```
hugovk, line 135: Technically this is false for Python 4.0-4.11 and true again for Python 4.12+.

We'll drop support for Python 3.11 in 2027 and this conditional can go away. I don't expect a Python 4.0 before then but it's not a certainty.

nulano: I think this is an OR, so it is correct for Python 4 but incorrectly includes Python 1.12 or Python 2.12, neither of which is likely to ever be released?

radarhere: I was thinking along @nulano's lines, that yeah, 1.12 and 2.12 aren't going to happen.

hugovk: All good then!

This is a good opportunity to bring up [PEP 404](https://peps.python.org/pep-0404/) (Python 2.8 not found) :)

radarhere:
PyPy 3.12 has now been released in beta, but has still not implemented PyConfig. See https://github.com/python-pillow/Pillow/pull/10047

Issue #5936, opened 2022-01-06
OSError: [Errno 24] Too many open files for image.save for tiff images


gooseillo:
**What did you do?**
I am opening a large number of jpg and tiff (single page & multi page) images, resizing them and saving them in a new directory. I am running this code within VS code in jupyter notebook using PIL 9.0.0

```python    
    def change_image_res(rep_image, 
                        new_path,
                        max_image_thresh=49, 
                        percent_reduce_res = [0.1, 0.2, 0.3, 0.4, 0.5]):

    image_name = rep_image.name
    save_all = True if rep_image.suffix == '.tiff' else False

    new_path = Path(new_path)

    with open(rep_image, "rb") as f:
        image = Image.open(f)
    
        if (image.size[0]*image.size[1])/1000000 > max_image_thresh:
            for res in percent_reduce_res:
                fixed_height = int(int((float(image.size[1]))) - (res * int((float(image.size[1])))))
                height_percent = (fixed_height / float(image.size[1]))
                width_size = int((float(image.size[0]) * float(height_percent)))
                if fixed_height*width_size/1000000 < 50:
                    image = image.resize((width_size, fixed_height), PIL.Image.NEAREST)
                    image.save(f'{new_path}\{image_name}', dpi=(300,300), save_all=save_all)
                    break
        else:
            image.save(f'{new_path}\{image_name}', dpi=(300,300), save_all=save_all)
```

**What did you expect to happen?**
I expected the images to be opened, processed, saved and closed automatically because of the context manager.

**What actually happened?**

After opening over a thousand images I get the following error

> OSError: [Errno 24] Too many open files

I also notice that the python.exe instance builds up around 5GB ram. Restarting the Kernel releases the memory

I have tried every thing possible on the internet like temp copy, deep copy, using close(), using load() etc.
I have also tried every thing in this issue

> https://github.com/python-pillow/Pillow/issues/2760

![Screenshot 2022-01-07 002627](https://user-images.githubusercontent.com/37607355/148415738-1710bb5e-01dd-4cc8-9958-ba2f14ce36ad.png)

radarhere:
If you didn't have the context manager, I would suggest calling `image.close` - https://pillow.readthedocs.io/en/stable/releasenotes/7.0.0.html#image-del

But you do, so that makes things more interesting.

Is there any way for you to provide us with a self-contained example including images, so that we can run the code and see the error for ourselves?

gooseillo:
[...]

PR #7199, opened 2023-06-05
Do not close provided file handles with libtiff

### LINKED ISSUE #7042: "close" syscall run twice on same file descriptor when loading binary LZW-compressed TIFF
If I open an LZW-compressed binary TIFF file using pillow and load its data, pillow calls the "close" syscall on one of the two relevant file descriptors twice. The second call results in a "EBADF (Bad file descriptor)" error. The program doesn't crash and everything seems to work ok. But if multithreading is used and a race condition happens, this might crash the program.

In this example, the expected behaviour would be that there is only one "close" syscall on each of the two relevant file descriptors and that there is no "EBADF" error. Below, I show how to reproduce the problem.

```console
$ cat create_tiff_file.py
#!/usr/bin/env python

import numpy as np
import PIL.Image as Image

Image.fromarray(np.random.rand(500, 600) > 0.9) \
    .save("example.tif", format="TIFF", compression="tiff_lzw")
$ cat read_tiff_file.py
#!/usr/bin/env python

import PIL.Image as Image

print("Right before the with-block", flush=True)
with Image.open("example.tif") as mask:
    print("Right in the beginning of the with-block", flush=True)
    mask.load()
    print("Right after load()", flush=True)
print("Right after the with-block", flush=True)
$ mamba create -n pillow_double_close_env --override-channels --channel anaconda python=3.11  # omitting the output
$ mamba activate pillow_double_close_env  # omitting the output
$ pip install numpy pillow  # omitting the output
$ conda list
# packages in environment at /home/philip/soft/mambaforge/envs/pillow_double_close_env:
#
# Name                    Version                   Build  Channel
_libgcc_mutex             0.1                        main    anaconda
_openmp_mutex             5.1                       1_gnu    anaconda
bzip2                     1.0.8                h7b6447c_0    anaconda
ca-certificates           2023.01.10           h06a4308_0    anaconda
certifi                   2022.9.24       py311h06a4308_0    anaconda
ld_impl_linux-64          2.38                 h1181459_1    anaconda
libffi                    3.4.2                h6a678d5_6    anaconda
libgcc-ng                 11.2.0               h1234567_1    anaconda
libgomp                   11.2.0               h1234567_1    anaconda
libstdcxx-ng              11.2.0               h1234567_1    anaconda
libuuid                   1.41.5               h5eee18b_0    anaconda
ncurses                   6.4                  h6a678d5_0    anaconda
numpy                     1.24.2                   pypi_0    pypi
openssl                   1.1.1s               h7f8727e_0    anaconda
pillow                    9.4.0                    pypi_0    pypi
pip                       22.2.2          py311h06a4308_0    anaconda
python                    3.11.0               h7a1cb2a_2    anaconda
readline                  8.2                  h5eee18b_0    anaconda
setuptools                65.5.0          py311h06a4308_0    anaconda
sqlite                    3.40.1               h5082296_0    anaconda
tk                        8.6.12               h1ccaba5_0    anaconda
tzdata                    2022a                hda174b7_0    anaconda
wheel                     0.37.1             pyhd3eb1b0_0    anaconda
xz                        5.2.10               h5eee18b_1    anaconda
zlib                      1.2.13               h5eee18b_0    anaconda
$ pip freeze
certifi @ file:///croot/certifi_1669749558946/work/certifi
numpy==1.24.2
Pillow==9.4.0
$ python create_tiff_file.py
$ strace -e decode-fds=path -e trace=openat,close ./read_tiff_file.py
# omitting all the output except for the few last lines, which are the most relevant
Right in the beginning of the with-block
close(4</pillow_double_close_bug_report/example.tif>) = 0
close(4)                                = -1 EBADF (Bad file descriptor)
close(3</pillow_double_close_bug_report/example.tif>) = 0
Right after load()
Right after the with-block
+++ exited with 0 +++
```

Also, I want to note that this caused my training of a neural network to crash once. It happened because Pytorch's `DataLoader` uses multithreading and a race condition happened.

radarhere:
Resolves #7042

libtiff closes file handles, but Pillow might still want access afterwards to seek to other frames in the file. To resolve this, TiffImagePlugin duplicates the file handle.

https://github.com/python-pillow/Pillow/blob/9a560c78a821638662206a51423f987eb095a901/src/PIL/TiffImagePlugin.py#L1256-L1258

#5936 had a problem with too many open files, so the following code was added to ensure that if libtiff did not close the file, Pillow would still close the file.
https://github.com/python-pillow/Pillow/blob/9a560c78a821638662206a51423f987eb095a901/src/PIL/TiffImagePlugin.py#L1308-L1312

#7042 doesn't like this behaviour, feeling that it is not thread-safe to close a file handle more than once.

Investigating how exactly libtiff closes file handles, I found that it is [`_tiffCloseProc`](https://gitlab.com/libtiff/libtiff/-/blob/master/libtiff/tif_unix.c#L132-137) that closes the file handle. This is called by [`TIFFClose`](https://gitlab.com/libtiff/libtiff/-/blob/master/libtiff/tif_close.c#L148-158), which we call from https://github.com/python-pillow/Pillow/blob/9a560c78a821638662206a51423f987eb095a901/src/libImaging/TiffDecode.c#L723

Looking at the code for [`TIFFClose`](https://gitlab.com/libtiff/libtiff/-/blob/master/libtiff/tif_close.c#L148-158), something becomes apparent - if we didn't want to call `_tiffCloseProc` when a file handle is in use, then we could just call `TIFFCleanup` instead of `TIFFClose`. Then we don't need to duplicate the file handle and don't need to try and close the duplicated file handle. Testing, this shouldn't break #5936, and passes #7042's reproduction (although that is [not necessarily indicative](https://github.com/python-pillow/Pillow/issues/7042#issuecomment-1490391771)).

### DIFF
--- a/src/libImaging/TiffDecode.c
+++ b/src/libImaging/TiffDecode.c
@@ -720,9 +720,14 @@ ImagingLibTiffDecode(
     }
 
  decode_err:
+    // TIFFClose in libtiff calls tif_closeproc and TIFFCleanup
     if (clientstate->fp) {
+        // Pillow will manage the closing of the file rather than libtiff
+        // So only call TIFFCleanup
         TIFFCleanup(tiff);
     } else {
+        // When tif_closeproc refers to our custom _tiffCloseProc though,
+        // that is fine, as it does not close the file
         TIFFClose(tiff);
     }
     TRACE(("Done Decoding, Returning \n"));
--- a/src/libImaging/TiffDecode.c
+++ b/src/libImaging/TiffDecode.c
@@ -720,7 +720,11 @@ ImagingLibTiffDecode(
     }
 
  decode_err:
-    TIFFClose(tiff);
+    if (clientstate->fp) {
+        TIFFCleanup(tiff);
+    } else {
+        TIFFClose(tiff);
+    }
     TRACE(("Done Decoding, Returning \n"));
     // Returning -1 here to force ImageFile.load to break, rather than
     // even think about looping back around.
--- a/src/PIL/TiffImagePlugin.py
+++ b/src/PIL/TiffImagePlugin.py
@@ -1253,9 +1253,8 @@ class TiffImageFile(ImageFile.ImageFile):
         # To be nice on memory footprint, if there's a
         # file descriptor, use that instead of reading
         # into a string in python.
-        # libtiff closes the file descriptor, so pass in a dup.
         try:
-            fp = hasattr(self.fp, "fileno") and os.dup(self.fp.fileno())
+            fp = hasattr(self.fp, "fileno") and self.fp.fileno()
             # flush the file descriptor, prevents error on pypy 2.4+
             # should also eliminate the need for fp.tell
             # in _seek
@@ -1305,18 +1304,11 @@ class TiffImageFile(ImageFile.ImageFile):
             # UNDONE -- so much for that buffer size thing.
             n, err = decoder.decode(self.fp.read())
 
-        if fp:
-            try:
-                os.close(fp)
-            except OSError:
-                pass
-
         self.tile = []
         self.readonly = 0
 
         self.load_end()
 
-        # libtiff closed the fp in a, we need to close self.fp, if possible
         if close_self_fp:
             self.fp.close()
             self.fp = None  # might be shared


src/libImaging/TiffDecode.c
```diff
@@ -720,7 +720,11 @@ ImagingLibTiffDecode(
     }
 
  decode_err:
-    TIFFClose(tiff);
+    if (clientstate->fp) {
```
Yay295, line 723: I think a comment here explaining why we're checking this would be good.

radarhere: Ok, I've added some comments.

PR #7209, opened 2023-06-11
Fix putdata() for I;16N on big-endian

Yay295:
Added some tests for Image methods that deal with bytes, and a test that `helper.hopper()` works for all image modes.

One of the tests failed though: https://github.com/Yay295/Pillow/actions/runs/5237154342/jobs/9455150887
`FAILED Tests/test_image.py::TestImageBytes::test_pixels_after_getdata_putdata[I;16N-1-2] - assert (1, 1029, 515, 1543) == (256, 1284, 770, 1798)`

For some reason doing `image.putdata(image.getdata())` on an `I;16N` image on a big-endian machine (s390x) changes the pixels.


src/_imaging.c
```diff
@@ -1571,7 +1571,22 @@ if (PySequence_Check(op)) { \
...
+                    endian = 1;
+                } else if (strcmp(image->mode, "I;16B") == 0) {
+                    endian = 2;
+                } else {
+                    // native endianness
+#ifdef WORDS_BIGENDIAN
+                    endian = 2;
+#else
+                    endian = 1;
+#endif
```
radarhere, line 1587:
```suggestion
                if (strcmp(image->mode, "I;16N") == 0) {
#ifdef WORDS_BIGENDIAN
                    endian = 2;
#else
                    endian = 1;
#endif
                } else if (strcmp(image->mode, "I;16B") == 0) {
                    endian = 2;
                } else {
                    endian = 1;
                }
```
This seems simpler to me?

Yay295:
> Yay295 force-pushed the `bytes_tests` branch

GitHub's suggestion auto-commit added an extra brace for some reason.

> Why would we need to test that this runs without error for all image modes? `hopper()` isn't part of Pillow, just the test suite. I don't see a need to start testing the tests.

You're right. This will essentially be tested just by using it in other tests anyways.

Yay295:
rebased on main

Tests/test_image.py
```diff
@@ -977,6 +979,59 @@ def test_fli_overrun2(self):
...
+        source_image = hopper(mode)
+        source_bytes = source_image.tobytes()
+        copy_image = Image.new(mode, source_image.size)
+        copy_image.frombytes(source_bytes)
+        assert copy_image.tobytes() == source_bytes
+
+    @pytest.mark.parametrize(("mode", "num_bands", "pixelsize"), image_modes_not_bgr)
+    def test_pixels_after_getdata_putdata(self, mode, num_bands, pixelsize):
+        image_byte_size = 2 * 2 * pixelsize
+        start_bytes = self.sample_bytes[0:image_byte_size]
```
radarhere, line 1013:
```suggestion
        start_bytes = self.sample_bytes[:image_byte_size]
```

Yay295:
Can this be merged?

hugovk (commented):
This PR has a fix to `src/_imaging.c`, let's rename the PR about that change, rather than about just adding tests.

radarhere:
I've created https://github.com/Yay295/Pillow/pull/13 with some suggestions.

radarhere (approved):

hugovk:
Thanks!

PR #7218, opened 2023-06-17
Improved checks in font_render

radarhere:
1. #7206 introduced new code to `font_render`, triggering https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=59831, because I didn't check that `Image.core.fill` didn't return NULL. I've added a fix for that, and tested it at https://github.com/radarhere/oss-fuzz/actions/runs/5296042009 and https://github.com/radarhere/oss-fuzz/actions/runs/5296052733.

2. I've also improved the memory handling by destroying the image when an error occurs.

3. Investigating https://github.com/python-pillow/Pillow/actions/runs/5262013783, I find that this failure is because of a problem checking for decompression bombs. 529621 * 541261 should be greater than `Image.MAX_IMAGE_PIXELS * 2`, but instead it becomes a negative number. Casting to `long long` first fixes that.


src/_imagingft.c
```diff
@@ -880,13 +880,17 @@ font_render(FontObject *self, PyObject *args) {
     width += stroke_width * 2 + ceil(x_start);
     height += stroke_width * 2 + ceil(y_start);
     if (max_image_pixels != Py_None) {
-        if (width * height > PyLong_AsLong(max_image_pixels) * 2) {
+        if ((long long)width * height > PyLong_AsLong(max_image_pixels) * 2) {
```
nulano, line 883:
```suggestion
        if ((long long)width * height > PyLong_AsLongLong(max_image_pixels) * 2) {
```

radarhere:
I'm unconcerned about the Windows failures, as I believe they will be fixed once the cache is cleared by #7216

hugovk:
There's a new UI for clearing CI cache files.

I cleared the cache by going to https://github.com/python-pillow/Pillow/actions/caches (also https://github.com/python-pillow/Pillow/actions > Management > Caches) and deleting `93478be93085fc54c3d24154a61c1a69df7fc001ac7a395ac7396f2123670bdc-b03bb08209f989d5ec95d20d01e46b3caf49b2c8e91095eb6898f612eea1518a-C:\hostedtoolcache\windows\PyPy\3.9.16\x86-17.6.33723.286`, and have restarted https://github.com/python-pillow/Pillow/actions/runs/5296892543/jobs/9591206960.

radarhere:
Thanks for the tip.

Windows is now passing. Cygwin Python 3.8 is failing however, as it is also doing on main - https://github.com/python-pillow/Pillow/actions/runs/5296887052/jobs/9595867281

PR #7228, opened 2023-06-23
Removed support for 32-bit

radarhere:
https://github.com/python-pillow/Pillow/issues/6941#issuecomment-1604058002
> I suggest we drop all the 32-bit wheels and testing.

This PR drops 32-bit support. Depends on https://github.com/python-pillow/docker-images/pull/185


.github/workflows/test-mingw.yml
```diff
@@ -93,14 +78,4 @@ jobs:
...
           file: ./coverage.xml
           flags: GHA_Windows
-          name: ${{ matrix.name }}
-
-  success:
-    permissions:
-      contents: none
-    needs: build
-    runs-on: ubuntu-latest
-    name: MinGW Test Successful
```
hugovk, line 103: We'll need to adjust https://github.com/python-pillow/Pillow/blob/0d54cf5c7fad29d7382ccb8927bb763b0dc4be7c/.github/mergify.yml#L10 to something like "Test MinGW"

.github/workflows/test-windows.yml
```diff
@@ -24,18 +24,11 @@ jobs:
...
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12-dev"]
-        architecture: ["x86", "x64"]
-        include:
-          # PyPy 7.3.4+ only ships 64-bit binaries for Windows
-          - python-version: "pypy3.9"
-            architecture: "x64"
-          - python-version: "pypy3.10"
-            architecture: "x64"
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12-dev", "pypy3.9", "pypy3.10"]
```
hugovk, line 27: PyPy is up to twice as slow as CPython tests:

image

Like test.yml, let's put pypy3.x first to give it a head start:

```suggestion
        python-version: ["pypy3.10", "pypy3.9", "3.8", "3.9", "3.10", "3.11", "3.12-dev"]
```

hugovk (commented):
Thanks! This will need release notes as well.

.github/workflows/test-mingw.yml
radarhere: Sorry, I'm not following at all.

I initially thought the success job could be removed it there was only one job, but now I realise that I'd removed something mergify was using, and so I've restored it.

Still not sure what you're suggesting though.

hugovk: Sorry for not being clear :) 

Yeah, so mergify needs to depend on something. When we had a matrix here, mergify depended on the final noop, success job.

But now this workflow has a single job, mergify can depend on that single job, and we can avoid the overhead of spinning up the final success job. 

In other words, we can do the same as the lint workflow.

radarhere: Oh, I didn't realise you'd referenced mergify.yml

Thanks. I've pushed a different commit, see what think.

hugovk: Looks good 👍

docs/releasenotes/10.0.0.rst
```diff
@@ -4,6 +4,11 @@
 Backwards Incompatible Changes
 ==============================
 
+32-bit support
+^^^^^^^^^^^^^^
+
+32-bit architecture is no longer supported.
```
hugovk, line 10: And mention wheels not produced?

radarhere: Ok, I've added that note.

hugovk (approved):
Thank you!

PR #7230, opened 2023-06-24
Fix Windows build with pyproject.toml

nulano:
Continuation of #7190.
Closes #7190.

Two changes:
* Use `pip install -C raqm=vendor -C fribidi=vendor .` instead of calling `setup.py` directly.
* Extract the environment setup into a separate script:
  * selecting wheel vs install and passing through pip flags cannot be easily done via a batch script,
  * this change (very slightly) speeds up compilation locally and on AppVeyor,
  * it will enable build caching simplifications after #7228 is merged.


.github/workflows/test-windows.yml
```diff
@@ -57,8 +57,8 @@ jobs:
     - name: Print build system information
       run: python3 .github/workflows/system-info.py
 
-    - name: python3 -m pip install setuptools wheel pytest pytest-cov pytest-timeout defusedxml
-      run: python3 -m pip install setuptools wheel pytest pytest-cov pytest-timeout defusedxml
+    - name: python3 -m pip install wheel pytest pytest-cov pytest-timeout defusedxml
+      run: python3 -m pip install wheel pytest pytest-cov pytest-timeout defusedxml
```
radarhere, line 61:
```suggestion
    - name: python3 -m pip install pytest pytest-cov pytest-timeout defusedxml
      run: python3 -m pip install pytest pytest-cov pytest-timeout defusedxml
```
If we're removing setuptools because it is in pyproject.toml, then we can remove wheel by the same logic.

.appveyor.yml
```diff
@@ -62,18 +62,15 @@ cache:
 - '%LOCALAPPDATA%\pip\Cache'
 
 artifacts:
-- path: pillow\dist\*.egg
+- path: pillow\*.egg
   name: egg
```
hugovk, line 66: The egg format is deprecated on PyPI:

* https://peps.python.org/pep-0715/

I know we're not uploading these, but do we need them here?

```suggestion
```

(Non-blocking for merge.)

hugovk (approved):

hugovk:
Eggs can wait, let's merge so everything is ready in time for tomorrow's build.  🥚🚀

Thanks!

PR #7232, opened 2023-06-25
Removed deleted file from codecov.yml and increased coverage threshold

radarhere:
#7228 removed Tests/32bit_segfault_check.py, so it no longer needs to be ignored in codecov.yml

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -12,7 +12,7 @@ coverage:
   status:
     project:
       default:
-        threshold: 0.01%
+        threshold: 0.1%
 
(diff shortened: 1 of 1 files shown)


hugovk:
Shall we also bump the threshold percent in this file?

It often "fails" PRs when there's unlikely to be any coverage change, and we just ignore it. 

For example: https://github.com/python-pillow/Pillow/pull/7231

Maybe we could bump it from 0.01% to 0.1%?

radarhere:
Ok, sure, I've added that change.

PR #7235, opened 2023-06-28
Limit size even if one dimension is zero in decompression bomb check

radarhere:


radarhere:
This was a fix for https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=48708

PR #7236, opened 2023-06-28
Do not use CFFI access by default on PyPy

### LINKED ISSUE #7223: Is CFFI still needed?
The CFFI is only used in PyAccess as a replacement for the C implementation in Access.c/PixelAccess. This was done in #476 to improve performance when using PyPy (see performance profiling in #816). However, that was nine years ago, and the PyPy cpyext has been [greatly improved](https://www.pypy.org/posts/2018/09/inside-cpyext-why-emulating-cpython-c-8083064623681286567.html) since then. Is PyAccess still necessary for performance on PyPy, or can we go back to just using the C implementation?

radarhere:
Resolves #7223

Pillow's C API is now faster than PyAccess on PyPy. However, PyAccess is part of our public API, so it can't be removed immediately.

This PR makes two changes.
1. `Image.USE_CFFI_ACCESS` is set to `False`,  even on PyPy, allowing the faster C API to be used by default. If users do set `Image.USE_CFFI_ACCESS` to `True` manually, then they will find that...
2. PyAccess is now deprecated, to be removed in Pillow 11.

### DIFF
--- a/docs/deprecations.rst
+++ b/docs/deprecations.rst
@@ -22,14 +22,17 @@ be removed in Pillow 11 (2024-10-15). This class was only made as a helper to
 be used internally, so there is no replacement. If you need this functionality
 though, it is a very short class that can easily be recreated in your own code.
 
-PyAccess
-~~~~~~~~
+PyAccess and Image.USE_CFFI_ACCESS
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 .. deprecated:: 10.0.0
 
 Since Pillow's C API is now faster than PyAccess on PyPy,
 :py:mod:`~PIL.PyAccess` has been deprecated and will be removed in Pillow
 11.0.0 (2024-10-15). Pillow's C API will now be used by default on PyPy instead.
+
+``Image.USE_CFFI_ACCESS``, for switching from the C API to PyAccess, is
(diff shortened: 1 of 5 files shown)


docs/deprecations.rst
```diff
@@ -22,6 +22,15 @@ be removed in Pillow 11 (2024-10-15). This class was only made as a helper to
...
 though, it is a very short class that can easily be recreated in your own code.
 
+PyAccess
+~~~~~~~~
+
+.. deprecated:: 10.0.0
+
+Since Pillow's C API is now faster than PyAccess on PyPy,
+:py:mod:`~PIL.PyAccess` has been deprecated and will be removed in Pillow
+11.0.0 (2024-10-15).
```
hugovk, line 32: And mention the C API is now used by default?

(And in docs/releasenotes/10.0.0.rst)

hugovk (approved):

radarhere:
I've updated the documentation to similarly deprecate `Image.USE_CFFI_ACCESS`.

Unfortunately, from https://discuss.python.org/t/extend-pep-562-with-setattr-for-modules/25506 it looks like it is not possible to override `__setattr__` on a module, so we can't raise a deprecation warning if users call `Image.USE_CFFI_ACCESS = True`.

PR #7246, opened 2023-07-01
Fixed deallocating mask images

radarhere:
Corrects a problem from https://github.com/python-pillow/Pillow/pull/7206

### DIFF
--- a/src/PIL/ImageFont.py
+++ b/src/PIL/ImageFont.py
@@ -562,9 +562,17 @@ class FreeTypeFont:
         _string_length_check(text)
         if start is None:
             start = (0, 0)
-        im, size, offset = self.font.render(
+        im = None
+
+        def fill(mode, size):
+            nonlocal im
+
+            im = Image.core.fill(mode, size)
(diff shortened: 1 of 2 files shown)


hugovk (approved):

src/PIL/ImageFont.py
```diff
@@ -562,9 +562,17 @@ def getmask2(
         _string_length_check(text)
         if start is None:
             start = (0, 0)
-        im, size, offset = self.font.render(
+        im = None
+
+        def fill(mode, size):
+            nonlocal im
+
+            im = Image.core.fill(mode, size)
```
nulano, line 570: It might make sense to move the decompression bomb check here to simplify the C function.

radarhere: I've created https://github.com/python-pillow/Pillow/pull/7247 for this.

radarhere:
This fixed https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=60273

PR #7248, opened 2023-07-02
Remove redundant wheel dep from pyproject.toml

mgorny:

 * Remove redundant `wheel` dep from `pyproject.toml`

----

Remove the redundant `wheel` dependency, as it is added by the backend automatically.  Listing it explicitly in the documentation was a historical mistake and has been fixed since, see: https://github.com/pypa/setuptools/commit/f7d30a9529378cf69054b5176249e5457aaf640a

While Pillow uses a custom backend that modifies the `bdist_wheel` method, it does not import `wheel` or use it in a way that would rely on setuptools implementation details.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,6 @@
 build-backend = "backend"
 requires = [
   "setuptools>=67.8",
-  "wheel",
 ]
 backend-path = [
   "_custom_build",


radarhere (approved):

radarhere:
Thanks. I was looking at https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/#fallback-behaviour when I added this in #7171

mgorny:
Thanks. I'll look into having an appropriate warning added there.

mgorny:
Thanks again! I've filed https://github.com/pypa/pip/pull/12122 to update pip documentation.

PR #7257, opened 2023-07-04
Corrected drawing I;16 points and writing I;16 text

### LINKED ISSUE #7245: ImageDraw only uses lower 8-bits of colors in I;16 images
### What did you do?

`ImageDraw` on an image with mode `I;16` fails to work correctly. When given a fill color (in the range 0...0xFFFF), it draws the shapes correctly, but uses a color with the lower 8-bits repeated twice.

The full 16-bit fill color provided should be used to draw on the image.

### What actually happened?
Only the lower 8-bits (duplicated into the higher and lower byte) were used.

* OS: macOS 12 (also confirmed on Ubuntu 20.04)
* Python: 3.10
* Pillow: 9.5.0

```python
image = Image.new("I;16", (10, 10))
draw = ImageDraw.Draw(image)

c = 0x1234
draw.point([(0, 0)], fill=c)
print(hex(image.load()[0, 0]))
# --> 0x3434
# (but we expect it to be 0x1234)

c = 0x4567
draw.line([(0, 0), (5, 0)], fill=c)
print(hex(image.load()[0, 0]))
# --> 0x6767
# (but we expect it to be 0x4567)

c = 0x89AB
draw.rectangle([(0, 0), (5, 5)], fill=c)
print(hex(image.load()[0, 0]))
# --> 0xABAB
# (but we expect it to be 0x89AB)
```

As you can see, for each color, the *actual* color applied to the image is based on the lower 8-bits, duplicated twice.

I suspect this has something to do with this: https://github.com/python-pillow/Pillow/blob/7c945f5131cf8596084b32af582f90a43b090540/src/libImaging/Draw.c#L71-L72
(Which was originally added in #3899)

Note that that issue is also referenced by #5598, so I suspect there might be a similar issue with drawing text.

radarhere:
Resolves #7245

As the issue explains,
https://github.com/python-pillow/Pillow/blob/be828f4d62b3802733a8fac6b6b59cced2092696/src/libImaging/Draw.c#L68-L75
is just drawing the writing the same value to two different points in our 8-bit image, whereas the point of 16-bit images is that each of those 2 8-bit values should be different.

This fixes that. It also means that the test from #3899 is slightly changed. The important part of that test was the width of the rectangle, which is maintained here.

The issue also requested a similar change for writing text, so I've added that commit for that.


src/libImaging/Paste.c
```diff
@@ -425,7 +425,7 @@ fill_mask_L(
                 *out = BLEND(*mask, *out, ink[0], tmp1);
                 if (strncmp(imOut->mode, "I;16", 4) == 0) {
                     out++;
-                    *out = BLEND(*mask, *out, ink[0], tmp1);
+                    *out = BLEND(*mask, *out, ink[1], tmp1);
```
homm, line 428: why?

radarhere: `ink` was originally an `int` in `_draw_bitmap`
https://github.com/python-pillow/Pillow/blob/7a1e28404d692d4a7fed33dc67019a1d8c4bf9d9/src/_imaging.c#L2898
and then became `const void *ink` for `ImagingDrawBitmap`.
https://github.com/python-pillow/Pillow/blob/7a1e28404d692d4a7fed33dc67019a1d8c4bf9d9/src/libImaging/Draw.c#L834
It passed through `ImagingFill2`, and then became `const UINT8 *ink` for `fill_mask_L`
https://github.com/python-pillow/Pillow/blob/7a1e28404d692d4a7fed33dc67019a1d8c4bf9d9/src/libImaging/Paste.c#L403

The second item of that `UINT8` pointer array will give us the second part of the 16 bits.

hugovk: @homm Any further concerns, or good to merge?

homm (approved):

PR #7260, opened 2023-07-05
Add tp_richcompare handler for Imaging_Type/ImagingCore

Yay295:
Imaging_Type/ImagingCore is exposed by Image.getdata(), so this allows you to properly compare the result of that method without it being an identity comparison. This should also speed up comparison between Image objects, because now the comparison is done in C instead of passing the data back to Python for it to be compared.


Tests/test_lib_image.py
```diff
@@ -30,3 +60,69 @@ def test_setmode():
...
+    data_a = data_b = secrets.token_bytes(num_img_bytes)
+    while data_a == data_b:
+        data_b = secrets.token_bytes(num_img_bytes)
+    img_a = Image.frombytes(mode, (2, 2), data_a)
+    img_b = Image.frombytes(mode, (2, 2), data_b)
+    assert img_a.tobytes() != img_b.tobytes()
+    assert img_a.im != img_b.im
+
+
+@pytest.mark.skip(reason="no way to directly set C bytes from Python")
```
radarhere, line 103: If this test is never run, then why add it?

radarhere:
Valgrind is currently failing - https://github.com/python-pillow/Pillow/actions/runs/5460694184/jobs/9937936242?pr=7260

Also, I went to test the speed of image equality with this, https://github.com/radarhere/Pillow/commit/98e02aa40d607a3c12b4108a72d070b1bd85f060... and found https://github.com/radarhere/Pillow/actions/runs/5461893489/jobs/9940492884#step:8:4787
> Fatal Python error: bool_dealloc: deallocating True or False: bug likely caused by a refcount error in a C extension

Tests/test_lib_image.py
Yay295: I thought someone else might know a way to do it. If I change to code to always use 0xffffffff for the mask, "test_image_reduce.py::test_mode_LA_opaque[1]" fails. So there are situations where identical images can end up with different values in their unused bytes, and that should be tested. I just don't know how.

Yay295:
oops, Py_NewRef wasn't added until Python 3.10.

src/_imaging.c
```diff
@@ -3721,6 +3721,134 @@ static PySequenceMethods image_as_sequence = {
...
+        !strcmp(mode, "RGB") || !strcmp(mode, "RGBX") ||
+        !strcmp(mode, "YCbCr") || !strcmp(mode, "HSV") || !strcmp(mode, "LAB")
+    ) {
+        // These modes have three channels in four bytes,
+        // so we have to ignore the last byte.
+        mask ^= 0xff;
+    } else if (!strcmp(mode, "LA") || !strcmp(mode, "La") || !strcmp(mode, "PA")) {
+        // These modes have two channels in four bytes,
+        // so we have to ignore the middle two bytes.
+        mask = 0xff0000ff;
```
radarhere, line 3745: You shouldn't need to ignore the middle two bytes, they should be equal to the first two. See https://github.com/python-pillow/Pillow/blob/3ffa8dca446ba22f3f143331e86cfef0e9caf8f2/src/libImaging/Convert.c#L234-L242 for example.

I initially wondered if you were adding this mask for speed, but looking at the next block, removing this mask would allow `memcmp` to be used, which I expect would be faster.

Yay295: "they should be equal to the first two" You'd think so, but they're not: https://github.com/Yay295/Pillow/actions/runs/5471546420/jobs/9962793523#step:8:1757

This is why I added those last two tests. Except I couldn't find a way to set this up from Python so I marked them to skip.

Yay295: I added some logging: https://github.com/Yay295/Pillow/actions/runs/5472920560/jobs/9965699214#step:8:1759

After `im.reduce()` the data is `0xffb8b8b8`, but after `im.crop().reduce()` the data is `0xff0000b8`.

homm: > they should be equal to the first two

They set to the same with rgb2la conversion, but this doesn't mean that they **should** be equal

Tests/test_lib_image.py
Yay295: Thanks. That works for RGB and YCbCr, but HSV, LAB, LA, La, and PA don't have a 4-byte raw mode.

Tests/test_lib_image.py
```diff
@@ -30,3 +60,69 @@ def test_setmode():
...
+        data_b = secrets.token_bytes(num_img_bytes)
+    img_a = Image.frombytes(mode, (2, 2), data_a)
+    img_b = Image.frombytes(mode, (2, 2), data_b)
+    assert img_a.tobytes() != img_b.tobytes()
+    assert img_a.im != img_b.im
+
+
+@pytest.mark.parametrize(
+    ("mode", "rawmode"),
+    (("RGB", "RGBX"), ("YCbCr", "YCbCrX"), ("HSV", None), ("LAB", None)),
```
homm, line 105: It's more semantically.
```suggestion
    [("RGB", "RGBX"), ("YCbCr", "YCbCrX"), ("HSV", None), ("LAB", None)],
```

Tests/test_lib_image.py
```diff
@@ -30,3 +60,69 @@ def test_setmode():
...
+        pytest.skip("no 4-byte rawmode for " + mode)
+    img_a = Image.frombytes(mode, (2, 2), b"ABC1DEF2GHI3JKL4", "raw", rawmode)
+    img_b = Image.frombytes(mode, (2, 2), b"ABC5DEF6GHI7JKL8", "raw", rawmode)
+    assert img_a.tobytes() == b"ABCDEFGHIJKL"
+    assert img_b.tobytes() == b"ABCDEFGHIJKL"
+    assert img_a.im == img_b.im
+
+
+@pytest.mark.skip(reason="no way to directly set C bytes from Python")
+@pytest.mark.parametrize("mode", ("LA", "La", "PA"))
```
homm, line 118:
```suggestion
@pytest.mark.parametrize("mode", ["LA", "La", "PA"])
```

Tests/test_lib_image.py
Yay295: A previous decision was to use tuples. https://github.com/python-pillow/Pillow/issues/6525#issuecomment-1272405965

homm: I've provided a detailed answer there. Hope @radarhere will agree with me

homm: As an example you can look at the [Django settings documentation](https://docs.djangoproject.com/en/4.2/ref/settings/). Settings are almost always immutable, but both lists and tuples are used for different things. Tuples are used for [ADMINS](https://docs.djangoproject.com/en/4.2/ref/settings/#admins) items (first value is a name, second is an email), [LANGUAGES](https://docs.djangoproject.com/en/4.2/ref/settings/#languages) items, [SECURE_PROXY_SSL_HEADER](https://docs.djangoproject.com/en/4.2/ref/settings/#secure-proxy-ssl-header) (first value is a name, second is a value of header). While lists are used for all sorts of… well… lists.

src/_imaging.c
Yay295: I added a new method to ImagingCore to allow directly setting the internal bytes. All of the tests work now.

src/_imaging.c
```diff
@@ -3534,6 +3534,34 @@ _save_ppm(ImagingObject *self, PyObject *args) {
     return Py_None;
 }
 
+static PyObject *
+_set_internal_pixel_bytes(ImagingObject *self, PyObject *args) {
```
homm, line 3538: Personally I don't think this is a good idea to add core method just for testing. The problems are:
* It needs maintenance in future
* It exposes internals which will be harder to change
* This method itself doesn't tested

However, for a long time I suffer from lack of functionality just to change internal model of the image without copying. Probably we can add something like `ImagingCore.rewrite_mode()` which will work within the same `pixelsize`.

Yay295: > It exposes internals which will be harder to change

I don't think it exposes anything that wouldn't also be exposed by `ImagingCore.rewrite_mode()`.

> This method itself doesn't tested

That can be added.

Tests/test_lib_image.py
radarhere: I'm not overly concerned about which is used, and I don't think performance considerations are exactly a priority for tests, so I'm happy to go with @homm's preference here.

src/_imaging.c
Yay295: > Probably we can add something like `ImagingCore.rewrite_mode()` which will work within the same `pixelsize`.

Unfortunately, pixelsize/linesize are not available anywhere except on an image - there is no list to loop through that has this information.

homm: Damn, right. Maybe create minimal image with target mode just to check it's pixel size?

Yay295: I think that would work. I still don't see how this is any better than the function I already added though.

homm: It have practical use cases outside of tests at least

Yay295: I ran into another issue. `Image.mode` and `Image.im.mode` are not the same object. So I can't set `Image.mode` from a method on `ImagingCore`. There's a note to fix this, but it hasn't been done yet: https://github.com/python-pillow/Pillow/blob/f089c2db8c1806f09e6ed53a1986146693db1ade/src/PIL/Image.py#L483C40-L483C40

Yay295: I can make `Image.mode` a property, but `Image.mode` is currently writable, while `Image.im.mode` is not.

Yay295: I've created #7271 to look into that.

homm: To unblock this PR I'd be happy with tests for modes we can test without core changes.

Yay295: With `ImagingCore.rewrite_mode()` there are none, because `ImagingCore.rewrite_mode()` cannot update the mode that is currently returned from `Image.mode`.

radarhere: I've created https://github.com/Yay295/Pillow/pull/8 as a suggestion.

Tests/test_lib_image.py
```diff
@@ -30,3 +60,59 @@ def test_setmode():
         im.im.setmode("L")
     with pytest.raises(ValueError):
         im.im.setmode("RGBABCDE")
+
+
+@pytest.mark.parametrize("mode", mode_names_not_bgr)
+def test_equal(mode):
+    num_img_bytes = len(Image.new(mode, (2, 2)).tobytes())
+    # alternatively, random.randbytes() in Python 3.9
+    data = secrets.token_bytes(num_img_bytes)
```
radarhere, line 69: Apart from oss-fuzz, Pillow doesn't use random data in the test suite.

src/_imaging.c
```diff
@@ -3748,6 +3748,138 @@ static PySequenceMethods image_as_sequence = {
...
+        const UINT8 **palette_a_data_ptr = &palette_a_data;
+        const UINT8 **palette_b_data_ptr = &palette_b_data;
+        if (
+            !palette_a || !palette_b
+            || palette_a->size != palette_b->size
+            || strcmp(palette_a->mode, palette_b->mode)
+            || _compare_pixels(
+                palette_a->mode,
+                1,
+                palette_a->size * 4,
```
radarhere, line 3847: Why is this multiplied by 4?

Tests/test_lib_image.py
```diff
@@ -1,7 +1,37 @@
...
+    "I;16N",
+    "RGB",
+    "RGBA",
+    "RGBa",
+    "RGBX",
+    # Image.frombytes() doesn't work with BGR modes:
+    # unknown raw mode for given image mode
+    # "BGR;15",
+    # "BGR;16",
+    # "BGR;24",
```
radarhere, line 28: I fixed these errors in #7303, but have since discovered that is not sufficient to get this PR working for those modes. The problem is that those modes are using their lines efficiently, rather than spacing out their data into 4 bytes per pixel.
https://github.com/python-pillow/Pillow/blob/96d683dafa3bb57562d0ba8bf887e18a8b8d9125/src/libImaging/Storage.c#L149-L150

src/_imaging.c
Yay295: The third parameter is the line size, but palettes don't have a line size, so it has to be calculated from the number of pixels and the pixel size. Palettes also don't have a stored pixel size, but currently it's always 4. It's not good to hardcode it like this, but there isn't anywhere to actually get it from.

Tests/test_lib_image.py
```diff
@@ -30,3 +60,59 @@ def test_setmode():
...
+        data_b = bytes(secrets.choice(b"\x00\xff") for i in range(4))
+    # We need to use rawmode "1;8" so that each full byte is interpreted as a value
+    # instead of the bits in the bytes being interpreted as values.
+    img_a = Image.frombytes("1", (2, 2), data_a, "raw", "1;8")
+    img_b = Image.frombytes("1", (2, 2), data_b, "raw", "1;8")
+    assert img_a.tobytes() != img_b.tobytes()
+    assert img_a.im != img_b.im
+
+
+@pytest.mark.parametrize("mode", [mode for mode in mode_names_not_bgr if mode != "1"])
```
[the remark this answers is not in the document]
Yay295: Because if you scroll up a bit, there's `test_not_equal_mode_1` just for mode 1 images, with comments explaining why it's different.

Basically, it's easier to create a mode 1 image from bytes if we use rawmode "1;8".

Tests/test_lib_image.py
Yay295: Different image modes need different amounts of bytes to create them, so this seemed like the best way to get enough bytes, and then get a second set of bytes that are different.

radarhere:
> Imaging_Type/ImagingCore is exposed by Image.getdata(), so this allows you to properly compare the result of that method without it being an identity comparison.

https://github.com/python-pillow/Pillow/pull/9292 deprecated this.

radarhere:
@akx did you have any thoughts on this PR? It seems [related](https://github.com/python-pillow/Pillow/pull/9883#issuecomment-5643131873) to other work currently being discussed.

akx:
@radarhere Yep, tangentially related to #9883, but also #9938! I'm all for faster image comparison that doesn't involve dumping both images into a secondary buffer...

I can do the work to forward-port this to the new mode ID stuff and so on. 👍

radarhere:
#10030 has taken over this work.

Issue #7259, opened 2023-07-04
Addition of _repr_jpeg_ causes both JPEG and PNG format images to be attached to Jupyter cells


smason:
I've had a go at improving the interaction between Jupyter and `Image`s.  As I commented on in #7135, that change causes the image to be saved in JPEG and PNG format and both files attached to the invoking cell.

This behavior seems redundant, and given that no resizing is done can result in very large `.ipynb` files if the user isn't careful when working with high resolution images.  I've had a look at how to avoid this, and the nicest way I could see to do this was by adding a `Image._repr_mimebundle_` method that would do the "right thing" in more circumstances.

Specifically, it would know whether the user has called `display_jpeg` or `display_png` and could return the appropriate encoding.  If neither format is specifically requested it could return something appropriate for the data.

I've had a go at implementing this and it currently looks like:
```python
THRESHOLD_SIZE = 1200

MODE_MAP = {
    'La': 'LA',
    'LAB': 'RGB',
    'HSV': 'RGB',
    'RGBX': 'RGB',
    'RGBa': 'RGBA',
}

VALID_JPEG_MODES = {'L', 'RGB', 'YCbCr', 'CMYK'}
VALID_PNG_MODES = {'1', 'P', 'L', 'RGB', 'RGBA', 'LA', 'PA'}

def _repr_mimebundle_(self, include=None, exclude=None, **kwargs):
    # let IPython interpret its ambiguous include/exclude flags
    if include is not None or exclude is not None or kwargs:
        return None

    # massage the image into something that has a reasonable
    # chance of being viewed by the user

    # handle the special snowflake modes
    if self.mode in {'I', 'F'}:
        # linearly transform extrema to fit in [0, 255]
        # this should have a similar result as Image.histogram
        lo, hi = self.getextrema()
        scale = 256 / (hi - lo) if lo != hi else 1
        image = self.point(lambda e: (e - lo) * scale).convert('L')
    elif self.mode == 'I;16':
        # linearly transform max down to 255
        image = self.point(lambda e: e / 256).convert('L')
    else:
        image = self

    # shrink large images so they don't take too long to transfer/render
    factor = max(image.size) // THRESHOLD_SIZE
    if factor > 1:
        image = image.reduce(factor)

    # process remaining modes into things supported by writers
    if image.mode in MODE_MAP:
        image = image.convert(MODE_MAP[image.mode])

    jpeg = image._repr_jpeg_() if image.mode in VALID_JPEG_MODES else None
[...]

PR #7266, opened 2023-07-06
Handle exceptions in _repr_jpeg_ and _repr_png_

### LINKED ISSUE #7283: iPython cannot display RGBA image after upgrade to 10.0.0
Previously we only have `_repr_png_`, so we can always display RGBA in ipython.

now we also have `_repr_jpg_` for some reason, the default way to display it is this jpg function, thus when we have RGBA image we cannot display it.

### What did you do?
in ipython qtconsole, I try to display RGBA image

```python
from PIL import Image
im = Image.new("RGBA", (126, 187))
display(im)
```

show the image without error.

The default repr should be png or at least for RGBA image it should be png.

### What actually happened?

display default to choose the jpeg formatter.
```
SError: cannot write mode RGBA as JPEG
```

* OS:  ubuntu 22.04
* Python:  3.10.6
* Pillow: 10.0.0

```python
from PIL import Image
im = Image.new("RGBA", (126, 187))
display(im)

```

mtreinish:
Resolves #7283

In 10.0.0 a `_repr_jpeg_` implementation was added to the Image object to enable the use of `display_jpeg()` in IPython environments. However, in some cases the implementation of this method could result in an exception being raised while trying to generate the jpeg data. The best example is if the image data is in an RGBA format as a result of the object being created by opening a PNG file. In this case trying to save the Image object as a jpeg will error because the jpeg format can't represent the transparency in the alpha channel. This results in an exception being raised in the IPython/Jupyter context when outputting the image object. However, in cases like this IPython allows the repr methods to return None to indicate there is no representation in that format available. [1] This commit updates the `_repr_png_` and `_repr_jpeg_` methods to catch any exception that might be raised while trying to generate the image data. If an exception is raised we treat that as not being able to generate image data in that format and return None instead.

Related to #7259

[1] https://ipython.readthedocs.io/en/stable/config/integrating.html#custom-methods

 * Handle any exceptions raised by generating image data in `Image._repr_png_` and return `None` instead of raising to users.
 * Handle any exceptions raised by generating image data in `Image._repr_jpeg_` and return `None` instead of raising to users.

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -651,14 +651,20 @@ class Image:
 
         :returns: PNG version of the image as bytes
         """
-        return self._repr_image("PNG", compress_level=1)
+        try:
+            return self._repr_image("PNG", compress_level=1)
+        except Exception:
+            return None
 
     def _repr_jpeg_(self):
         """iPython display hook support for JPEG format.
 
         :returns: JPEG version of the image as bytes
         """
-        return self._repr_image("JPEG")
+        try:
+            self._repr_image("JPEG")
+        except Exception:
+            return None
 
     @property
     def __array_interface__(self):
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -662,7 +662,7 @@ class Image:
         :returns: JPEG version of the image as bytes
         """
         try:
-            self._repr_image("JPEG")
+            return self._repr_image("JPEG")
         except Exception:
             return None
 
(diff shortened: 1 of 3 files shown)


src/PIL/Image.py
```diff
@@ -651,14 +651,20 @@ def _repr_png_(self):
...
+            return None
 
     def _repr_jpeg_(self):
         """iPython display hook support for JPEG format.
 
         :returns: JPEG version of the image as bytes
         """
-        return self._repr_image("JPEG")
+        try:
+            self._repr_image("JPEG")
```
radarhere, line 665:
```suggestion
            return self._repr_image("JPEG")
```

mtreinish: Heh, I've been writing too much rust lately and keep forgetting the returns in python :)

radarhere:
I've created https://github.com/mtreinish/Pillow/pull/1 with a suggestion to simplify the code.

radarhere (approved):

PR #7271, opened 2023-07-09
Delegate Image mode and size to ImagingCore

Yay295:
This causes a lot of errors. The main issues are that `Image.im` isn't always set, so we still need to keep Python-only variables for these properties, and the Python values are writable, but the C values are not writable from Python.

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -508,37 +508,31 @@ class Image:
 
     @property
     def size(self):
-        return self._size
-
-    @property
-    def _size(self):
         if self._use_im_values():
             return self.im.size
         return self.__size
 
-    @_size.setter
     def _size(self, value):
         # set im.size first in case it raises an exception
         if self._use_im_values():
             self.im.size = value
         self.__size = value
 
-    @property
-    def mode(self):
-        return self._mode
+    _size = property(fset=_size)
 
     @property
-    def _mode(self):
+    def mode(self):
         if self._use_im_values():
             return self.im.mode
         return self.__mode
 
-    @_mode.setter
     def _mode(self, value):
         # set im.mode first in case it raises an exception
         if self._use_im_values():
             self.im.mode = value
         self.__mode = value
+
+    _mode = property(fset=_mode)
 
     def _new(self, im):
         new = Image()
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -492,7 +492,10 @@ class Image:
         self._exif = None
 
     def _use_im_values(self):
-        ''' Whether or not to try using values from self.im in addition to the values in this class. '''
+        """
+        Whether or not to try using values from self.im
+        in addition to the values in this class.
+        """
         return self.im is not None
 
     @property
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -480,15 +480,20 @@ class Image:
 
     def __init__(self):
         # FIXME: take "new" parameters / other image?
-        # FIXME: turn mode and size into delegating properties?
         self.im = None
-        self._mode = ""
-        self._size = (0, 0)
+        # do not directly change __mode; use _mode instead
+        self.__mode = ""
+        # do not directly change __size; use _size instead
+        self.__size = (0, 0)
         self.palette = None
         self.info = {}
         self.readonly = 0
         self.pyaccess = None
         self._exif = None
+
+    def _use_im_values(self):
+        ''' Whether or not to try using values from self.im in addition to the values in this class. '''
+        return self.im is not None
 
     @property
     def width(self):
@@ -503,8 +508,34 @@ class Image:
         return self._size
 
     @property
+    def _size(self):
+        if self._use_im_values():
+            return self.im.size
+        return self.__size
+
+    @_size.setter
+    def _size(self, value):
+        # set im.size first in case it raises an excepton
+        if self._use_im_values():
+            self.im.size = value
+        self.__size = value
+
+    @property
     def mode(self):
         return self._mode
+
+    @property
+    def _mode(self):
+        if self._use_im_values():
+            return self.im.mode
+        return self.__mode
+
+    @_mode.setter
+    def _mode(self, value):
+        # set im.mode first in case it raises an excepton
+        if self._use_im_values():
+            self.im.mode = value
+        self.__mode = value
 
     def _new(self, im):
         new = Image()
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -518,7 +518,7 @@ class Image:
 
     @_size.setter
     def _size(self, value):
-        # set im.size first in case it raises an excepton
+        # set im.size first in case it raises an exception
         if self._use_im_values():
             self.im.size = value
         self.__size = value
@@ -535,7 +535,7 @@ class Image:
 
     @_mode.setter
     def _mode(self, value):
-        # set im.mode first in case it raises an excepton
+        # set im.mode first in case it raises an exception
         if self._use_im_values():
             self.im.mode = value
         self.__mode = value
--- a/src/PIL/ImageFile.py
+++ b/src/PIL/ImageFile.py
@@ -139,6 +139,9 @@ class ImageFile(Image.Image):
         if self.format is not None:
             return Image.MIME.get(self.format.upper())
 
+    def _use_im_values(self):
+        return self.tile is None and self.im is not None
+
     def __setstate__(self, state):
         self.tile = []
         super().__setstate__(state)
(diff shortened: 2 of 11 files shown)


Yay295:
There is a `setmode` method on `ImagingCore` that could perhaps be used as the setter for `ImagingCore.mode`. I'm not sure why it wasn't, so I didn't make that change.

src/PIL/Image.py
```diff
@@ -480,16 +480,26 @@ class Image:
...
         self._exif = None
 
+    @property
+    def mode(self):
+        return self.im.mode if self.im else self._mode
+
+    @mode.setter
+    def mode(self, value):
+        if self.im:
+            self.im.mode = value
```
homm, line 499: It's intended that you can't directly set size (or mode for now), `setter`s are not needed.

radarhere:
Could you provide an explanation for why this change is helpful?

Yay295:
I think keeping the mode and size of `Image` and `ImagingCore` in sync is a good thing, and delegating to `ImagingCore` when possible keeps them in sync.

src/PIL/Image.py
Yay295: The problem is that you *can* currently set the mode and size, as can be seen by the many failing tests, so not allowing this would be a breaking change.

homm: There is no way to set size (only _size). The same rules should be for mode

radarhere: I created #7307 to prevent users from setting the mode directly.

radarhere: I've created https://github.com/Yay295/Pillow/pull/9 to fix the tests that fail here.

radarhere:
As this PR knows, because `self.im` starts out as `None`, and stays that way until an image is loaded, `mode` and `size` can't always be delegated to C, and must exist sometimes in Python.

> I think keeping the mode and size of Image and ImagingCore in sync is a good thing

I agree, and I would think that they are already in sync.

> and delegating to ImagingCore when possible keeps them in sync.

While this PR isn't yet finalised, it would seem that the idea here is to make Python's `_mode` and the ImagingCore `mode` go out of sync. The public API's `mode` might stay correct, but I'm uncomfortable with the idea of Python's `_mode` sometimes disagreeing with the C mode. It would seem to be clearer code if they were both kept the same.

Yay295:
The original issue that I created this change for was discussed here: https://github.com/python-pillow/Pillow/pull/7260#discussion_r1255424515

The problem was that changing an image's mode in C does not change the mode seen in Python, because the C code doesn't have access to the Python value. So if an image's mode or size is ever changed directly in C, it will be out of sync with the value(s) in Python. Likewise, if one of these values is changed in Python, it will be out of sync with the value(s) in C (making `Image.size` and `Image.mode` read-only helps prevent this). By always using the C value if it exists we can be sure we are using the same value everywhere.

radarhere:
> The problem was that changing an image's mode in C does not change the mode seen in Python, because the C code doesn't have access to the Python value. So if an image's mode or size is ever changed directly in C, it will be out of sync with the value(s) in Python.

Pillow has a public Python API. Our intention is not for users to access the C API directly. If `ImagingCore.rewrite_mode()` is something that is added in the future, we can talk about that then, but for the moment, I'm not aware of any logic in C determines what the ImagingCore mode is - the logic to do so lives in Python.

> Likewise, if one of these values is changed in Python, it will be out of sync with the value(s) in C (making Image.size and Image.mode read-only helps prevent this).

I think ignoring the Python mode would obscure the fact that in this scenario, setting the mode in Python is not having the intended effect. That doesn't sound helpful to me.

> By always using the C value if it exists we can be sure we are using the same value everywhere.

It ensures that `im.mode` reflects the C mode, but you're also deliberately removing setting `_mode` in this PR, making the Python `_mode` sometimes incorrect. This is not part of the public API, so you may think it doesn't matter, but `_mode` is something that plugins still use, and so something that users writing plugins still need to understand. It seems to me like this makes the code less clear.

Yay295:
> If `ImagingCore.rewrite_mode()` is something that is added in the future, we can talk about that then,

`ImagingCore.rewrite_mode()` has been proposed as part of #7260, which is why we are talking about his now.

> I think ignoring the Python mode would obscure the fact that in this scenario, setting the mode in Python is not having the intended effect.

This is currently the case. Before #7307 (which is not yet part of a release), it was possible to set `Image.mode`. This would sometimes actually work if the C mode wasn't used, and sometimes it wouldn't if the C mode was used.

> but you're also deliberately removing setting `_mode` in this PR

`_mode` did not exist before #7307. This change was made before that one, and I did add a setter for `mode`. Your changes are the ones that removed the setters.

radarhere:
If you'd like to wait until #7260 is resolved, okay.

> It ensures that im.mode reflects the C mode, but you're also deliberately removing setting _mode in this PR, making the Python _mode sometimes incorrect.

When I said this, I wasn't referring to setters for mode, I was referring to
Screenshot 2023-08-03 at 3 32 24 pm

Yay295:
`Image.im` is also set in all of those locations, so `Image.mode` would have gotten its data from `Image.im.mode`, making those lines redundant. I suppose `Image.im` could be set to `None` after that point though, so they wouldn't be redundant in that case.

Yay295:
New change based on main.

`mode` and `size` are read-only properties.
`_mode` and `_size` are read-write properties.
`__mode` and `__size` are the actual values.

If `_mode` or `_size` are changed, it will also attempt to set those values on `Image.im`.

src/PIL/ImageFile.py
```diff
@@ -139,6 +139,9 @@ def get_format_mimetype(self):
         if self.format is not None:
             return Image.MIME.get(self.format.upper())
 
+    def _use_im_values(self):
+        return self.tile is None and self.im is not None
```
homm, line 143: What is purpose of `self.tile` check?

src/_imaging.c
```diff
@@ -3646,11 +3646,49 @@ _getattr_mode(ImagingObject *self, void *closure) {
...
+        return 0;
+    }
+
+    int xsize, ysize;
+    if (!PyArg_ParseTuple(value, "ii", &xsize, &ysize)) {
+        return -1;
+    }
+
+    self->image->xsize = xsize;
+    self->image->ysize = ysize;
```
homm, line 3688: Why we want to do this? It will just break image

src/PIL/Image.py
```diff
@@ -502,10 +510,36 @@ def height(self):
     def size(self):
         return self._size
 
+    @property
+    def _size(self):
+        if self._use_im_values():
+            return self.im.size
+        return self.__size
```
homm, line 517: Why this can't be just in `size` property?

homm (changes requested):

src/PIL/Image.py
Yay295: The `_size` getter is actually used in quite a few places, so that would have to be changed to use the `size` getter first. It would save three lines if the `_size` getter was removed though:
```python
    @property
    def size(self):
        if self._use_im_values():
            return self.im.size
        return self.__size

    def _size(self, value):
        # set im.size first in case it raises an exception
        if self._use_im_values():
            self.im.size = value
        self.__size = value

    _size = property(fset=_size)
```

src/_imaging.c
Yay295: So that if `Image.size` is changed, `Image.im.size` will have the same value. Another option could be that if `Image.size` is changed, and `Image.im.size` doesn't match, it raises an exception.

src/PIL/ImageFile.py
Yay295: @radarhere proposed that change for `ImageFile` in a [pull request](https://github.com/Yay295/Pillow/pull/9) they made for the previous code. The issue is that when seeking to another frame, the `Image` values may not match the `Image.im` values. Using `not self.tile` like in their pull request wasn't working for me though, so I changed it to `self.tile is None`, which after looking into it more, probably isn't correct.

radarhere:
https://github.com/python-pillow/Pillow/pull/7271#issuecomment-1663274371
> `ImagingCore.rewrite_mode()` has been proposed as part of https://github.com/python-pillow/Pillow/pull/7260, which is why we are talking about this now.

As far as I can see, https://github.com/Yay295/Pillow/pull/8 removed the need for `ImagingCore.rewrite_mode()` in that PR. Does that make this PR unnecessary?

I'm not in favour of this PR, for [reasons stated earlier](https://github.com/python-pillow/Pillow/pull/7271#issuecomment-1662229669). @homm you've left some individual comments, but regarding the basic idea, do you see value here that I don't?

homm:
> regarding the basic idea, do you see value here that I don't?

I do see profit in wall-time and memory consumption, but I'm not sure about public interface. In general, I think we can close it for now.

PR #7276, opened 2023-07-11
Set unused bytes to zero when converting to LA/La/PA

radarhere:
Inspired by @homm's comment https://github.com/python-pillow/Pillow/pull/7260#discussion_r1254128299, when converting to LA/La/PA, this sets the two middle channels to zero.

This shouldn't make any functional difference at the moment, just improving clarity, and perhaps a small improvement in speed by replacing `*out++ = (UINT8)pixel` with `*out++ = 0`.


homm:
I personally don't have preferences what to store in unused bytes. Also I guess performance impact should be near zero/

PR #7289, opened 2023-07-16
Set C palette to be empty by default

### LINKED ISSUE #7288: Incorrect transparency after converting indexed image to RGB
### What did you do?

I use Pillow to convert images on my web site on-demand to .webp format.  It works great! Except for one of two images on this page: https://nedbatchelder.com/text/cardcube.html  The first image is strangely pixellated.  The second looks perfect.

The webp conversion should produce the same pixels as the original .png files.

### What actually happened?

Unfortunate graininess.

* OS: MacOS
* Python: 3.9.17
* Pillow: 10.0.0

I tried converting the two .png files into a variety of formats.  The .gif and .bmp look identical to the .png.  The .webp shows the results I see on my site:

```py
from pathlib import Path
from PIL import Image

for n in [1, 2]:
    for fmt in ["WEBP", "GIF", "BMP"]:
        inpath = Path(f"cardcube_step{n}.png")
        outpath = inpath.with_suffix("." + fmt.lower())
        im = Image.open(inpath)
        im.save(outpath, fmt, lossless=True, quality=80, method=6)
```

Here are the two .png files:

cardcube_step1
cardcube_step2

Here are the .webp files I get:

[webp.zip](https://github.com/python-pillow/Pillow/files/12063803/webp.zip)

radarhere:
Resolves #7288

The image in the issue has indexes that occur outside of the palette. This PR ensures that they are black.

### DIFF
--- a/Tests/test_file_gif.py
+++ b/Tests/test_file_gif.py
@@ -205,14 +205,14 @@ def test_optimize_full_l():
 
 
 def test_optimize_if_palette_can_be_reduced_by_half():
-    with Image.open("Tests/images/test.colors.gif") as im:
-        # Reduce dimensions because original is too big for _get_optimize()
-        im = im.resize((591, 443))
-    im_rgb = im.convert("RGB")
+    im = Image.new("P", (8, 1))
+    im.palette = ImagePalette.raw("RGB", bytes((0, 0, 0) * 150))
+    for i in range(8):
+        im.putpixel((i, 0), (i + 1, 0, 0))
 
     for optimize, colors in ((False, 256), (True, 8)):
         out = BytesIO()
-        im_rgb.save(out, "GIF", optimize=optimize)
+        im.save(out, "GIF", optimize=optimize)
         with Image.open(out) as reloaded:
             assert len(reloaded.palette.palette) // 3 == colors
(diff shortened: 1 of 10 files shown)


Yay295:
This seems inefficient. `ImagingPaletteNew` already uses `calloc` to get its data, but then initializes it. It would be faster to just not initialize it in the first place. `ImagingPaletteNew` isn't used in very many places, so I think the cleaner solution would be to change `ImagingPaletteNew` to not initialize its values, and create a new function `ImagingPaletteNewRamp` that first calls `ImagingPaletteNew` and then initializes the data as `ImagingPaletteNew` does currently. There's another already existing function `ImagingPaletteNewBrowser` that does essentially that, but with different starting values.

https://github.com/python-pillow/Pillow/blob/b9451540973ca7915b0ce68e6bbacbed665caec2/src/libImaging/Palette.c#L23C1-L95

radarhere:
Ok, I've pushed a new commit to set the palette to be empty by default.

This is now a C sequel to https://github.com/python-pillow/Pillow/pull/5552/commits/4d36feefc7cda3fc3a77ef873f3c4c166aa69352

PR #7299, opened 2023-07-22
Added saving LA images as PDFs

radarhere:
Resolves #7293

This is similar to #6925, except for LA instead of RGBA, and so uses
https://github.com/python-pillow/Pillow/blob/ea0abb94cce2aa8b6aed33f57b3871053f2aad59/src/PIL/PdfImagePlugin.py#L160-L161
instead of
https://github.com/python-pillow/Pillow/blob/ea0abb94cce2aa8b6aed33f57b3871053f2aad59/src/PIL/PdfImagePlugin.py#L174-L175

### DIFF
--- a/src/libImaging/Jpeg2KEncode.c
+++ b/src/libImaging/Jpeg2KEncode.c
@@ -490,6 +490,8 @@ j2k_encode_entry(Imaging im, ImagingCodecState state) {
 
     if (strcmp(im->mode, "RGBA") == 0) {
         image->comps[3].alpha = 1;
+    } else if (strcmp(im->mode, "LA") == 0) {
+        image->comps[1].alpha = 1;
     }
 
     opj_set_error_handler(codec, j2k_error, context);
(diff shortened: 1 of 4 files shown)


src/libImaging/Jpeg2KEncode.c
```diff
@@ -490,6 +490,8 @@ j2k_encode_entry(Imaging im, ImagingCodecState state) {
 
     if (strcmp(im->mode, "RGBA") == 0) {
         image->comps[3].alpha = 1;
+    } else if (strcmp(im->mode, "LA") == 0) {
```
homm, line 493: Does it mean there was no LA support for JPEG2K at all? Do we need better testing for this case, which is not related to pdf?

radarhere: Without this change, saving an LA image as JPEG2K and visually inspecting it, shows a black rather than transparency. This fixes that. However, `getpixel()` gives `(0, 0)` either way.

When Pillow reads a JPEG2K image, it only checks the "number of components" to see if it is LA.

homm (approved):

PR #7302, opened 2023-07-24
Allow saving I;16B images as PNG

### LINKED ISSUE #7301: Enable writing I;16B images as PNG
While investigating an ImageIO issue concerning behavior on big-endian architectures I've noticed that pillow's PNG plugin currently doesn't allow writing 16-bit big-endian images to PNG. `I;16` is already supported, so I was wondering if it would be possible to add `I;16B` as well.

From what I can see, the only change needed to support this would be to add the line `"I;16B": ("I;16B", b"\x10\x00"),` to this dict:

https://github.com/python-pillow/Pillow/blob/3c5324b07c2ed98bc0ec94bd5c6a30f9c66e42fb/src/PIL/PngImagePlugin.py#L1035-L1051

With this change we would be able to do 16 bit round-trips on big-endian, i.e., we can decode the pixel data directly into a 16 bit big-endian buffer and then use that same buffer when encoding as PNG. This would improve performance and streamline working with native encoding on big-endian machines.

radarhere:
Resolves #7301


hugovk (approved):
Worthy of release notes? Maybe too minor, if so, feel free to merge.

radarhere:
Yes, I think this is a minor change.

PR #7303, opened 2023-07-25
Support BGR;15, BGR;16 and BGR;24 access, unpacking and putdata

radarhere:
#7209 and #7260 need to exclude BGR;15, BGR;16 and BGR;24 from their tests because these modes cannot be unpacked and used with putdata.

This fixes that. It doesn't add PyAccess support though, since that is deprecated as of #7236


src/_imaging.c
```diff
@@ -1571,21 +1573,46 @@ if (PySequence_Check(op)) { \
...
+                int b;
+                for (i = x = y = 0; i < n; i++) {
+                    char ink[4];
+
+                    op = PySequence_Fast_GET_ITEM(seq, i);
+                    if (!op || !getink(op, image, ink)) {
+                        Py_DECREF(seq);
+                        return NULL;
+                    }
+                    /* FIXME: what about scale and offset? */
```
radarhere, line 1609: This is copying an existing FIXME
https://github.com/python-pillow/Pillow/blob/07623d1a7cc65206a5355fba2ae256550bfcaba6/src/_imaging.c#L1639

hugovk (approved):
Worthy of release notes? Maybe too minor, if so, feel free to merge.

PR #7307, opened 2023-07-28
Changed Image mode property to be read-only by default

radarhere:
https://github.com/python-pillow/Pillow/pull/7271#discussion_r1257538129
> There is no way to set size (only _size). The same rules should be for mode

This PR implements that change. This is for `mode` what #3203 was for `size`.


homm (approved):
Docs/changelog updates?

radarhere:
Ok, I've added release notes.

PR #7310, opened 2023-07-29
Added CMYK to RGB unpacker

### LINKED ISSUE #7309: Add capability to process CMYK palettes
### What did you do?
I'm using Pillow to add image extraction capability in pypdf. I'm facing some issues with images using some CMYK palettes.
However Pillow is currently not able to process them.

Accept CMYK indexed images.

### What actually happened?
Exception `ValueError: unrecognized raw mode` is raised.

* OS: windows
* Python: 3.10.4
* Pillow: 10.0.0

test file: 

```python
from PIL using Image

img = Image.open("test.png")
img2.putpalette(b"\x00\x00\x00\xFB\x00\x00\x00\x00","CMYK")
```

radarhere:
Resolves #7309

Adds a CMYK to RGB unpacker, essentially copying the code from https://github.com/python-pillow/Pillow/blob/07623d1a7cc65206a5355fba2ae256550bfcaba6/src/libImaging/Convert.c#L567-L579


hugovk:
> essentially copying the code from ...

Is it worth de-duplicating, and having them both calling a common function?

homm (approved):

radarhere:
Ok, I've pushed a commit to include the Convert function in Unpack.

PR #7311, opened 2023-07-31
Read WebP duration after opening

k128:
Many WebP images don't initially have duration in their info dict.  This automatically adds that information during image open.

### DIFF
--- a/Tests/test_file_webp.py
+++ b/Tests/test_file_webp.py
@@ -233,5 +233,4 @@ class TestFileWebp:
             im.save(out_webp, save_all=True)
 
         with Image.open(out_webp) as reloaded:
-            reloaded.load()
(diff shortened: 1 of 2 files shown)


radarhere:
I've created https://github.com/k128/Pillow/pull/1 with some suggestions.

radarhere (approved):

radarhere:
As long as feedback is received before a release, I'm happy.

Looking at https://developers.google.com/speed/webp/docs/container-api, I don't see another way to get the timestamp. So your suggestion is simply to revert this? I've created #7406.

homm:
@k128 unfortunately we reverted this due to [significant performance degradation](https://github.com/python-pillow/Pillow/pull/7406#issuecomment-1739658824).

Issue #7312, opened 2023-08-01
RGBA image losing its transparency when saved as pdf


pubpub-zz:
### What did you do?
convert an png to image to pdf or ~JPEG2000~
edit: 
   after checking with various  image viewer, I've observed that the issue could be linked with my original viewer(nomacs)
   In order to not bother any I will left on side JPEG2000 and carry on my investigations

### What did you expect to happen?

have the good alpha channel

### What actually happened?

alpha channel not correctly displayed in pdf  ~is not the good one~

### What are your OS, Python and Pillow versions?

* OS: windows 11
* Python:  3.10
* Pillow: 10.0
with this png
![tt](https://github.com/python-pillow/Pillow/assets/4083478/e054f4b7-dcbc-4173-ab28-e19e9b6d4fd9)

```python
from PIL import Image
img = Image.open("tt.png")
print("mode : ",img.mode)
img.save("tt.pdf")
img.save("tt.jp2")
img = Image.open("tt.jp2")
print("mode jp2 : ",img2.mode)
img.save("tt.jp2")
```
output
_mode : RGBA_
_mode jp2 : RGBA_

but the image in the PDF has lost its transparency
~same for the jpeg2000 image, although still reported as RGBA~

homm:
I can't confirm JPEG2000. Your image is transparent after conversion.

pubpub-zz:
@homm 
this is the pdf I'm getting: 
[tt.pdf](https://github.com/python-pillow/Pillow/files/12233871/tt.pdf)

homm:
This is how your pdf is rendered to me. It's unclear to me if PDF really can be transparent. 

<img width="902" alt="Screenshot 2023-08-02 at 01 55 57" src="https://github.com/python-pillow/Pillow/assets/128982/3bedfe56-6446-4fac-9338-59ad1a0bd72d">

radarhere:
I'm also unable to replicate the problem with either format. Are you sure that you are using Pillow 10.0.0, and not a version before 9.5.0?

pubpub-zz:
I reinstalled it before to confirm the bug was on the latest version.
Actually it looks like the problem is also linked with the viewer:
https://github.com/mozilla/pdf.js/issues/5649#issuecomment-1597489448

Personally I checked all under windows with acrobat reader (black) sumatra (transparent) pdf.js( black and damaged) chrome (black)

radarhere:
How did you install Pillow?

[...]

PR #7316, opened 2023-08-03
Set SMaskInData to 1 for PDFs with alpha

radarhere:
Helps #7312

Pillow can convert a PNG with transparency to PDF
```python
from PIL import Image
im = Image.open("Tests/images/transparent.png")
im.save("transparent.pdf")
```
However, what does transparency mean exactly in PDF? There is debate about this in #7312, and various image viewers are discussed, so I'll try and keep it simple here by saving it that if I convert the PDF back to PNG with ImageMagick, `convert transparent.pdf out.png`, at the moment, I get a black background instead of transparency.

In https://opensource.adobe.com/dc-acrobat-sdk-docs/pdfstandards/PDF32000_2008.pdf, under 7.4.9 'JPXDecode Filter', it states
> **SMaskInData** specifies whether soft-mask information packaged with the image samples shall be used (see 11.6.5.3, "Soft-Mask Images"); if it is, the **SMask** entry shall not be present. If **SMaskInData** is nonzero, there shall be only one opacity channel in the JPEG2000 data and it shall apply to all colour channels.

If I set `SMaskInData` to 1 when saving the PDF, the roundtripped PNG keeps its transparency.

### DIFF
--- a/src/PIL/PdfImagePlugin.py
+++ b/src/PIL/PdfImagePlugin.py
@@ -125,20 +125,19 @@ def _save(im, fp, filename, save_all=False):
             # (packbits) or LZWDecode (tiff/lzw compression).  Note that
             # PDF 1.2 also supports Flatedecode (zip compression).
 
-            bits = 8
             params = None
             decode = None
-            smaskindata = 0
 
             #
             # Get image characteristics
 
             width, height = im.size
 
+            dict_obj = {"BitsPerComponent": 8}
             if im.mode == "1":
                 if features.check("libtiff"):
                     filter = "CCITTFaxDecode"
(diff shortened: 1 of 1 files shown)


homm (approved):

homm:
Have you tested it, does it solves problems with pdf.js from #7312?

radarhere:
Testing pdf.js, with main, I get 
before
With this branch, I get
after

This is the same final result that the user reported in the issue. They believe that the inverted colors is a bug in pdf.js, that they have reported in https://github.com/mozilla/pdf.js/issues/16782

homm (approved):

PR #7323, opened 2023-08-06
Fixed transparency when saving P mode images to PDF

### LINKED ISSUE #7312: RGBA image losing its transparency when saved as pdf

radarhere:
Resolves #7312

The (last unresolved) image from the issue is

which saved as a PDF with main becomes
Screenshot 2023-08-06 at 10 41 42 pm

```python
print(img.mode)
print(img.info)
```
gives
```
P
{'transparency': b'\x00PJ\xdb/@k\x0b\x1d\x18\xac~\x9a\x98\xd9m\xe2s\xde\xbc\xc2\x90\xb0\xde\xef\xe8\xd1\xed\xcf\xb5\xbd\xde\xf2\xd9\xf1\xeb\xf2\xeb\xee\xf8'}
```

This is not a single transparency index, but different levels of opacity for different palette entries.

To attach this to a P mode PDF image, this PR writes a secondary image, the image converted to just an alpha channel, as an "SMask".

https://opensource.adobe.com/dc-acrobat-sdk-docs/pdfstandards/PDF32000_2008.pdf describes "SMask" under 8.9.5 'Image Dictionaries' as
> A subsidiary image XObject defining a soft-mask image (see 11.6.5.3, "Soft-Mask Images") that shall be used as a source of mask shape or mask opacity values in the transparent imaging model.

With that, the PDF becomes
Screenshot 2023-08-06 at 10 42 14 pm
which can be converted back to a transparent PNG using ImageMagick's `convert`.


Tests/test_file_pdf.py
```diff
@@ -48,6 +48,22 @@ def test_save_alpha(tmp_path, mode):
...
+        assert im.mode == "P"
+        assert isinstance(im.info["transparency"], bytes)
+
+        # Act
+        im.save(outfile)
+
+    # Assert
+    with open(outfile, "rb") as fp:
+        contents = fp.read()
+    assert b"SMask" in contents
```
homm, line 64: Can we add some separators around `SMask`?

homm (approved):

Tests/test_file_pdf.py
radarhere: Sure, I've updated the test to check for `b"\n/SMask "`

homm: I'm ok with other

PR #7332, opened 2023-08-11
Added session type check in grabclipboard for Linux

### LINKED ISSUE #7331: ImageGrab.grabclipboard on Linux with both X11 and Wayland installed
### What did you do?
1. Using `grabclipboard` on X11 session with `wl-paste` utility installed regardless if `xclip` is installed
2. Using `grabclipboard` on Wayland session without `wl-paste` but `xclip` is installed

1.  * If `xclip` is installed, `xclip` command should be used
    * If `xclip` is not installed, raise `NotImplementedError`
2. Raise `NotImplementedError`

### What actually happened?
1. `wl-paste` command is used instead of `xclip`
```pytb
>>> ImageGrab.grabclipboard()
Failed to connect to a Wayland server
  File "[path redacted]/python3.11/site-packages/PIL/ImageGrab.py", line 144, in grabclipboard
    output = subprocess.check_output(["wl-paste", "-l"]).decode()
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/subprocess.py", line 466, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/subprocess.py", line 571, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['wl-paste', '-l']' returned non-zero exit status 1.
```
3. `xclip` is used and failed as it cannot access Wayland clipboard
```pytb
>>> ImageGrab.grabclipboard()
  File "[path redacted]/lib/python3.11/site-packages/PIL/ImageGrab.py", line 165, in grabclipboard
    raise ChildProcessError(msg)
ChildProcessError: xclip error: Error: target image/png not available
```

* OS: Debian 12
* Python: 3.11.2
* Pillow: 10.0.0

```python
from PIL import ImageGrab

ImageGrab.grabclipboard()
```

TheNooB2706:
Fixes #7331

* Check if current session is X11 or Wayland and decide which command to use. This is done by using `loginctl` command (based on [this](https://unix.stackexchange.com/a/371164)). If the session check failed (loginctl not available because eg. not on systemd), fall back to previous routine (assume Wayland first, if `wl-paste` unavailable then assume X11)
* Raise `NotImplementedError` if `xclip` unavailable on X11 session or `wl-paste` unavailable on Wayland session

### DIFF
--- a/src/PIL/ImageGrab.py
+++ b/src/PIL/ImageGrab.py
@@ -147,7 +147,7 @@ def grabclipboard():
         else:  # Session type check failed
             session_type = None
 
-        if shutil.which("wl-paste") and session_type in ["wayland", None]:
+        if shutil.which("wl-paste") and session_type in ("wayland", None):
             output = subprocess.check_output(["wl-paste", "-l"]).decode()
             mimetypes = output.splitlines()
             if "image/png" in mimetypes:
@@ -160,7 +160,7 @@ def grabclipboard():
             args = ["wl-paste"]
             if mimetype:
                 args.extend(["-t", mimetype])
-        elif shutil.which("xclip") and session_type in ["x11", None]:
+        elif shutil.which("xclip") and session_type in ("x11", None):
             args = ["xclip", "-selection", "clipboard", "-t", "image/png", "-o"]
         else:
             msg = "wl-paste or xclip is required for ImageGrab.grabclipboard() on Linux"
--- a/src/PIL/ImageGrab.py
+++ b/src/PIL/ImageGrab.py
@@ -140,27 +140,10 @@ def grabclipboard():
                 return BmpImagePlugin.DibImageFile(data)
         return None
     else:
-        if shutil.which("loginctl"):
-            try:
-                loginctl = subprocess.check_output("loginctl").decode().split("\n")
-            except subprocess.CalledProcessError:
-                loginctl = None
-        else:
-            loginctl = None
-
-        if loginctl is not None:
-            username = os.getlogin()
-            sessionid = [
-                line.split()[0] for line in loginctl if username in line.split()
-            ][0]
-            sessiontype = (
-                subprocess.check_output(
-                    ["loginctl", "show-session", sessionid, "-p", "Type"]
-                )
-                .decode()
-                .strip("\n")
-                .split("=")[1]
-            )
+        if os.getenv("WAYLAND_DISPLAY"):
+            sessiontype = "wayland"
+        elif os.getenv("DISPLAY"):
+            sessiontype = "x11"
         else:  # Session type check failed
             sessiontype = None
 
--- a/src/PIL/ImageGrab.py
+++ b/src/PIL/ImageGrab.py
@@ -149,7 +149,7 @@ def grabclipboard():
             loginctl = None
 
         if loginctl is not None:
-            username = subprocess.check_output("whoami").decode().strip("\n")
+            username = os.getlogin()
             sessionid = [
                 line.split()[0] for line in loginctl if username in line.split()
             ][0]
--- a/src/PIL/ImageGrab.py
+++ b/src/PIL/ImageGrab.py
@@ -164,9 +164,7 @@ def grabclipboard():
         else:  # Session type check failed
             sessiontype = None
 
-        if shutil.which("wl-paste") and (
-            (sessiontype == "wayland") or (sessiontype is None)
-        ):
+        if shutil.which("wl-paste") and sessiontype in ["wayland", None]:
             output = subprocess.check_output(["wl-paste", "-l"]).decode()
             mimetypes = output.splitlines()
             if "image/png" in mimetypes:
@@ -179,9 +177,7 @@ def grabclipboard():
             args = ["wl-paste"]
             if mimetype:
                 args.extend(["-t", mimetype])
-        elif shutil.which("xclip") and (
-            (sessiontype == "x11") or (sessiontype is None)
-        ):
+        elif shutil.which("xclip") and sessiontype in ["x11", None]:
             args = ["xclip", "-selection", "clipboard", "-t", "image/png", "-o"]
         else:
             msg = "wl-paste or xclip is required for ImageGrab.grabclipboard() on Linux"
--- a/src/PIL/ImageGrab.py
+++ b/src/PIL/ImageGrab.py
@@ -140,7 +140,24 @@ def grabclipboard():
                 return BmpImagePlugin.DibImageFile(data)
         return None
     else:
-        if shutil.which("wl-paste"):
+        if shutil.which("loginctl"):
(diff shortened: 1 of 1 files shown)


src/PIL/ImageGrab.py
```diff
@@ -153,11 +179,14 @@ def grabclipboard():
             args = ["wl-paste"]
             if mimetype:
                 args.extend(["-t", mimetype])
-        elif shutil.which("xclip"):
+        elif shutil.which("xclip") and (
+            (sessiontype == "x11") or (sessiontype is None)
+        ):
```
radarhere, line 184:
```suggestion
        elif shutil.which("xclip") and sessiontype in ["x11", None]:
```
Is this simpler?

src/PIL/ImageGrab.py
```diff
@@ -140,7 +140,33 @@ def grabclipboard():
...
+                .decode()
+                .strip("\n")
+                .split("=")[1]
+            )
+        else:  # Session type check failed
+            sessiontype = None
+
+        if shutil.which("wl-paste") and (
+            (sessiontype == "wayland") or (sessiontype is None)
+        ):
```
radarhere, line 169:
```suggestion
        if shutil.which("wl-paste") and sessiontype in ["wayland", None]:
```

src/PIL/ImageGrab.py
```diff
@@ -140,7 +140,33 @@ def grabclipboard():
...
+            except subprocess.CalledProcessError:
+                loginctl = None
+        else:
+            loginctl = None
+
+        if loginctl is not None:
+            username = subprocess.check_output("whoami").decode().strip("\n")
+            sessionid = [
+                line.split()[0] for line in loginctl if username in line.split()
+            ][0]
```
radarhere, line 155: Would it be simpler to just use [`os.getsid()`](https://docs.python.org/3/library/os.html#os.getsid)?

TheNooB2706: I'm not sure what pid to pass to the function though, `os.getsid(0)` (sid of current process) seems to be different from the one from `loginctl`

TheNooB2706: Alternatively, the session type can be obtained by checking the environment variable `XDG_SESSION_TYPE`, but I didn't go with that route because user can override it (security reason?). Or maybe it is fine? That would simplify a lot.

radarhere: I don't think the user being able to control whether xclip or wl-paste is used constitutes a security risk.

It's just hard to see a consensus on StackOverflow about which methods work and which ones don't. An alternative method is suggested at https://unix.stackexchange.com/a/559950, of checking `WAYLAND_DISPLAY` instead. For some reason, when we added testing for `wl-paste` in #7094, we set `WAYLAND_DISPLAY`, but not `XDG_SESSION_TYPE` - so at the moment, only the first variable is available in our CI jobs. Something to keep in mind.

TheNooB2706: Indeed there doesn't seem to have universal method for the check. Should I go ahead and change to using environment variable?

src/PIL/ImageGrab.py
```diff
@@ -140,7 +140,31 @@ def grabclipboard():
...
+            sessiontype = (
+                subprocess.check_output(
+                    ["loginctl", "show-session", sessionid, "-p", "Type"]
+                )
+                .decode()
+                .strip("\n")
+                .split("=")[1]
+            )
```
TheNooB2706, line 163:
```suggestion
        if os.getenv("WAYLAND_DISPLAY"):
            sessiontype = "wayland"
        elif os.getenv("DISPLAY"):
            sessiontype = "x11"
```

That would change to something like this

radarhere: Sounds reasonable to me.

src/PIL/ImageGrab.py
```diff
@@ -153,11 +160,12 @@ def grabclipboard():
             args = ["wl-paste"]
             if mimetype:
                 args.extend(["-t", mimetype])
-        elif shutil.which("xclip"):
+        elif shutil.which("xclip") and sessiontype in ["x11", None]:
```
radarhere, line 163:
```suggestion
        elif shutil.which("xclip") and session_type in ["x11", None]:
```

src/PIL/ImageGrab.py
```diff
@@ -140,7 +140,14 @@ def grabclipboard():
...
     else:
-        if shutil.which("wl-paste"):
+        if os.getenv("WAYLAND_DISPLAY"):
+            sessiontype = "wayland"
+        elif os.getenv("DISPLAY"):
+            sessiontype = "x11"
+        else:  # Session type check failed
+            sessiontype = None
+
+        if shutil.which("wl-paste") and sessiontype in ["wayland", None]:
```
radarhere, line 150:
```suggestion
        if os.getenv("WAYLAND_DISPLAY"):
            session_type = "wayland"
        elif os.getenv("DISPLAY"):
            session_type = "x11"
        else:  # Session type check failed
            session_type = None

        if shutil.which("wl-paste") and session_type in ["wayland", None]:
```
https://peps.python.org/pep-0008/#function-and-variable-names
> Function names should be lowercase, with words separated by underscores as necessary to improve readability.
>
> Variable names follow the same convention as function names.

radarhere (approved):

src/PIL/ImageGrab.py
```diff
@@ -140,7 +140,14 @@ def grabclipboard():
...
     else:
-        if shutil.which("wl-paste"):
+        if os.getenv("WAYLAND_DISPLAY"):
+            session_type = "wayland"
+        elif os.getenv("DISPLAY"):
+            session_type = "x11"
+        else:  # Session type check failed
+            session_type = None
+
+        if shutil.which("wl-paste") and session_type in ["wayland", None]:
```
hugovk, line 150:
```suggestion
        if shutil.which("wl-paste") and session_type in ("wayland", None):
```

src/PIL/ImageGrab.py
```diff
@@ -153,11 +160,12 @@ def grabclipboard():
             args = ["wl-paste"]
             if mimetype:
                 args.extend(["-t", mimetype])
-        elif shutil.which("xclip"):
+        elif shutil.which("xclip") and session_type in ["x11", None]:
```
hugovk, line 163:
```suggestion
        elif shutil.which("xclip") and session_type in ("x11", None):
```

hugovk (approved):

PR #7336, opened 2023-08-12
Allow GaussianBlur and BoxBlur to accept a sequence of x and y radii

### LINKED ISSUE #7321: Gaussian/Box blur with different X and Y radius
### What did you do?

Call `ImageFilter.BoxBlur` or `ImageFilter.GaussianBlur` with the `radius` parameter set to a 2-tuple containing an X and Y radius.

The blur operation should use different radii for the X and Y dimensions. If one of the dimensions is zero, then the blur should only be done in the other axis.

### What actually happened?

This is not supported, per documentation (the radius must be a single number).

* OS: Linux
* Python: 3.11
* Pillow: 9.5.0

### Example code

```python
# im is an image

# This works
im.filter(ImageFilter.GaussianBlur(32))

# This doesn't work
im.filter(ImageFilter.GaussianBlur((32, 16)))
```

### Other notes

It looks like there's already internal code for doing single-dimensional box blur; making the 2-dimensional blur functions (which are exposed to Python) use different radii per dimension should be pretty easy. See this code:

https://github.com/python-pillow/Pillow/blob/a5b025629023477ec62410ce77fd717c372d9fa2/src/libImaging/BoxBlur.c#L266-L281

I would probably be happy to send in a PR, if this is a feature that you would welcome -- let me know.

radarhere:
Resolves #7321

Allows specifying a different radius for X and Y for ImageFilter's GaussianBlur and BoxBlur, e.g. `ImageFilter.GaussianBlur((32, 16))` and `ImageFilter.BoxBlur((32, 16))`.

This is in addition to the existing single number, `ImageFilter.GaussianBlur(16)` and `ImageFilter.BoxBlur(16)`.

### DIFF
--- a/src/libImaging/BoxBlur.c
+++ b/src/libImaging/BoxBlur.c
@@ -230,14 +230,14 @@ ImagingHorizontalBoxBlur(Imaging imOut, Imaging imIn, float floatRadius) {
 }
 
 Imaging
-ImagingBoxBlur(Imaging imOut, Imaging imIn, float radius, int n) {
+ImagingBoxBlur(Imaging imOut, Imaging imIn, float xradius, float yradius, int n) {
     int i;
     Imaging imTransposed;
 
     if (n < 1) {
         return ImagingError_ValueError("number of passes must be greater than zero");
     }
-    if (radius < 0) {
+    if (xradius < 0 || yradius < 0) {
         return ImagingError_ValueError("radius must be >= 0");
     }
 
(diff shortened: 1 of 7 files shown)


src/libImaging/BoxBlur.c
```diff
@@ -266,16 +266,16 @@ ImagingBoxBlur(Imaging imOut, Imaging imIn, float radius, int n) {
     /* Apply blur in one dimension.
        Use imOut as a destination at first pass,
        then use imOut as a source too. */
-    ImagingHorizontalBoxBlur(imOut, imIn, radius);
+    ImagingHorizontalBoxBlur(imOut, imIn, xradius);
     for (i = 1; i < n; i++) {
-        ImagingHorizontalBoxBlur(imOut, imOut, radius);
+        ImagingHorizontalBoxBlur(imOut, imOut, xradius);
```
homm, line 271: We could skip passes if it’s radius == 0

radarhere: Sure. I've pushed a commit

homm (approved):

PR #7344, opened 2023-08-18
Updated zlib to 1.3

radarhere:
zlib 1.3 has been released - https://zlib.net/

### DIFF
--- a/Tests/test_file_png.py
+++ b/Tests/test_file_png.py
@@ -79,7 +79,7 @@ class TestFilePng:
 
     def test_sanity(self, tmp_path):
         # internal version number
-        assert re.search(r"\d+\.\d+\.\d+(\.\d+)?$", features.version_codec("zlib"))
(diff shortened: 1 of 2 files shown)


Tests/test_file_png.py
```diff
@@ -79,7 +79,7 @@ def get_chunks(self, filename):
 
     def test_sanity(self, tmp_path):
         # internal version number
-        assert re.search(r"\d+\.\d+\.\d+(\.\d+)?$", features.version_codec("zlib"))
+        assert re.search(r"\d+(\.\d+){1,3}$", features.version_codec("zlib"))
```
radarhere, line 82: Without this change, our test suite fails with zlib 1.3, as this code presumes zlib versions are X.Y.Z, not just X.Y.

The Cygwin builds have actually started failing on main because this is missing - https://github.com/python-pillow/Pillow/actions/runs/5921422961

PR #7351, opened 2023-08-23
Skip tests that require FreeType if FreeType is not available

radarhere:
If I disable the optional dependency FreeType, various tests fail.

This PR skips them if FreeType is not available.

### DIFF
--- a/Tests/oss-fuzz/test_fuzzers.py
+++ b/Tests/oss-fuzz/test_fuzzers.py
@@ -6,6 +6,7 @@ import packaging
 import pytest
 
 from PIL import Image, features
+from Tests.helper import skip_unless_feature
 
 if sys.platform.startswith("win32"):
     pytest.skip("Fuzzer is linux only", allow_module_level=True)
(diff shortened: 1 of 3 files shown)


Tests/oss-fuzz/test_fuzzers.py
```diff
@@ -48,6 +48,7 @@ def test_fuzz_images(path):
         fuzzers.disable_decompressionbomb_error()
 
 
+@pytest.mark.skipif(not features.check("freetype2"), reason="freetype2 not available")
```
hugovk, line 51: Can we also use this one here? 

```suggestion
@skip_unless_feature("freetype2")
```

radarhere: This is in the oss-fuzz subdirectory, rather than just in Tests. I might be missing something, but all my attempts to `skip_unless_feature` via `helper` from here lead to `ImportError: attempted relative import with no known parent package`

hugovk: <kbd>option</kbd>+<kbd>enter</kbd> in PyCharm suggests `from Tests.helper import skip_unless_feature`, which works for me locally:

image

radarhere: Thanks, that works. I've updated the commit.

PR #7355, opened 2023-08-28
Allow getpixel() to accept a list

### LINKED ISSUE #7352: getpixel() should work with list of length 2
### What did you do?
`img.getpixel([473, 653])`

the pixel at that position is returned

### What actually happened?
An unexpected error message: `TypeError: argument must be sequence of length 2`

But the argument is a **sequence** of length 2

* OS: Ubuntu 18.04
* Python: 3.9
* Pillow: '9.4.0'

```python
img = Image.open('test.png')
img.getpixel([473, 653])
```

radarhere:
Resolves #7352

The issue points out that while `im.getpixel()` accepts a tuple, `im.getpixel((0, 0))`, it does not accept a list.
```pytb
>>> im.getpixel([0, 0])
  File "PIL/Image.py", line 1568, in getpixel
TypeError: argument must be sequence of length 2
```
Even though the list supplied is still a "sequence of length 2".

It is worth noting that this error string is shared with another situation.
```pytb
>>> im.load()[0, 0, 0]
TypeError: argument must be sequence of length 2
```
So the error message cannot be simply reworded to say "tuple", as I don't think that clearly explains the problem with `im.load()[0, 0, 0]` .

Instead, this PR allows `im.getpixel()` to accept a list.

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -1565,6 +1565,8 @@ class Image:
         self.load()
         if self.pyaccess:
             return self.pyaccess.getpixel(xy)
+        if isinstance(xy, list):
+            xy = tuple(xy)
         return self.im.getpixel(xy)
 
     def getprojection(self):
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -1565,9 +1565,7 @@ class Image:
         self.load()
         if self.pyaccess:
             return self.pyaccess.getpixel(xy)
-        if isinstance(xy, list):
-            xy = tuple(xy)
-        return self.im.getpixel(xy)
+        return self.im.getpixel(tuple(xy))
 
     def getprojection(self):
         """
--- a/src/_imaging.c
+++ b/src/_imaging.c
@@ -1146,11 +1146,15 @@ static inline int
 _getxy(PyObject *xy, int *x, int *y) {
     PyObject *value;
 
-    if (!PyTuple_Check(xy) || PyTuple_GET_SIZE(xy) != 2) {
+    int tuple = PyTuple_Check(xy);
+    if (
+      !(tuple && PyTuple_GET_SIZE(xy) == 2) &&
+      !(PyList_Check(xy) && PyList_GET_SIZE(xy) == 2)
+    ) {
         goto badarg;
     }
 
-    value = PyTuple_GET_ITEM(xy, 0);
+    value = tuple ? PyTuple_GET_ITEM(xy, 0) : PyList_GET_ITEM(xy, 0);
     if (PyLong_Check(value)) {
         *x = PyLong_AS_LONG(value);
     } else if (PyFloat_Check(value)) {
@@ -1164,7 +1168,7 @@ _getxy(PyObject *xy, int *x, int *y) {
         }
     }
 
-    value = PyTuple_GET_ITEM(xy, 1);
+    value = tuple ? PyTuple_GET_ITEM(xy, 1) : PyList_GET_ITEM(xy, 1);
     if (PyLong_Check(value)) {
         *y = PyLong_AS_LONG(value);
     } else if (PyFloat_Check(value)) {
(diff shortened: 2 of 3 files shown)


src/_imaging.c
```diff
@@ -1146,11 +1146,15 @@ static inline int
 _getxy(PyObject *xy, int *x, int *y) {
     PyObject *value;
 
-    if (!PyTuple_Check(xy) || PyTuple_GET_SIZE(xy) != 2) {
+    int tuple = PyTuple_Check(xy);
```
homm, line 1149: So many checks here? Isn't a way do something like `PyArg_ParseTuple(xy, "ii", x, y)`?

radarhere: Possibly - but if your goal is simplicity, then I should mention that the simplest solution is to implement this change in Python instead of C.
```diff
diff --git a/src/PIL/Image.py b/src/PIL/Image.py
index 476ed0122..6ea711b56 100644
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -1565,7 +1565,7 @@ class Image:
         self.load()
         if self.pyaccess:
             return self.pyaccess.getpixel(xy)
-        return self.im.getpixel(xy)
+        return self.im.getpixel(tuple(xy))
 
     def getprojection(self):
         """
```

radarhere: My original intention here was to make this change close to the source of the error though. If the change is being made higher up, then my argument that it would be awkward to update the error message for just this scenario falls apart, since you could just add a new error message.

Then I feel less sure about this change, as the discourse in this issue doesn't have a clear conclusion, and I'm reminded of #3738, in particular https://github.com/python-pillow/Pillow/issues/3738#issuecomment-476701199 and https://github.com/python-pillow/Pillow/issues/3738#issuecomment-507126728.

https://github.com/python-pillow/Pillow/issues/3738#issuecomment-507126728
> Also, tuples and lists are for slightly different things (immutable/mutable, structure/order). Does that really matter? I think maybe it does: semantically you know a coordinate like (100, 200) can't have a third value; but [100, 200] looks like it could. So I'm tending towards keeping it as is.

hugovk: I didn't have strong feelings in https://github.com/python-pillow/Pillow/issues/3738#issuecomment-507126728, but in the years since then am now tending the other way, maybe we could accept more iterables for these kinds of things?

radarhere: Ok, I've pushed a commit with a simpler change in Python.

src/PIL/Image.py
```diff
@@ -1565,6 +1565,8 @@ def getpixel(self, xy):
         self.load()
         if self.pyaccess:
             return self.pyaccess.getpixel(xy)
+        if isinstance(xy, list):
+            xy = tuple(xy)
         return self.im.getpixel(xy)
```
homm, line 1570: Unconditional casting is faster and makes possible [other types which was asked](https://github.com/python-pillow/Pillow/issues/7352#issuecomment-1693088128).

```python
In [1]: a = (1, 2)

In [3]: %timeit _ = tuple(a)
29.4 ns ± 0.059 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [4]: %%timeit
   ...: if isinstance(a, list):
   ...:     _ = tuple(a)
51.9 ns ± 0.0765 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [2]: b = [1, 2]

In [5]: %timeit _ = tuple(b)
42.8 ns ± 0.0378 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [6]: %%timeit
   ...: if isinstance(b, list):
   ...:     _ = tuple(b)
83.4 ns ± 0.235 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

```suggestion
        return self.im.getpixel(tuple(xy))
```

radarhere: In main at the moment, `im.getpixel(0)` raises `TypeError: argument must be sequence of length 2`

Your suggestion would raise instead `TypeError: 'int' object is not iterable`

I was trying to avoid that change. Perhaps it is not a significant concern however.

homm (approved):

PR #7359, opened 2023-08-31
Map buffer into (0, 0) image

radarhere:
In `frombuffer`, Pillow maps the buffer onto a (1, 1) image.
https://github.com/python-pillow/Pillow/blob/24606216e1e5931a8fe6f41acde9e7e67489905d/src/PIL/Image.py#L3003-L3004

As a small change for the sake of performance, this PR changes it to be a (0, 0) image instead.

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -3000,7 +3000,7 @@ def frombuffer(mode, size, data, decoder_name="raw", *args):
         if args == ():
             args = mode, 0, 1
         if args[0] in _MAPMODES:
-            im = new(mode, (1, 1))
+            im = new(mode, (0, 0))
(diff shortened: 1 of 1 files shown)


src/PIL/Image.py
```diff
@@ -3000,7 +3000,7 @@ def frombuffer(mode, size, data, decoder_name="raw", *args):
         if args == ():
             args = mode, 0, 1
         if args[0] in _MAPMODES:
-            im = new(mode, (1, 1))
+            im = new(mode, (0, 0))
```
homm, line 3003: For the record. This call gives 1.37 µs instead of 1.47 µs, which probably unmeasurable for whole `frombuffer` function.

homm (approved):

PR #7382, opened 2023-09-07
Read bounding box information from the trailer of EPS files if specified

nopperl:
The bounding box for EPS images is currently only parsed from the file header. However, the DSC specification also allows it to be specified in the trailer (see Section 4.6 of [the specification](https://web.mit.edu/PostScript/Adobe/Documents/5001.DSC_Spec.pdf)). According to the specification, this must be indicated in the header using `%%BoundingBox: (atend)`. The real bounding box information is then specified at the end of the file following the `%%Trailer` comment. For an example, see `Tests/images/zero_bb_trailer.eps`.

Currently, the `EpsImagePlugin` tries to parse the bounding box only from the header, which fails if the value is `(atend)`. This PR fixes the error and implements boundary box parsing from the trailer.


Tests/test_file_eps.py
```diff
@@ -404,3 +404,16 @@ def test_timeout(test_file):
...
+
+@pytest.mark.parametrize(
+    ("trailer_image_file", "header_image_file"),
+    [("Tests/images/zero_bb_trailer.eps", FILE1)],
+)
+def test_boundary_box_in_trailer(trailer_image_file, header_image_file):
+    # Check whether boundary boxes which are specified in the trailer are parsed
+    # just like boundary boxes specified in the header
+    with Image.open(trailer_image_file) as trailer_image, Image.open(
+        header_image_file
```
radarhere, line 417:
```suggestion
def test_boundary_box_in_trailer():
    # Check whether boundary boxes are parsed in the same way
    # when specified in the header or the trailer
    with Image.open("Tests/images/zero_bb_trailer.eps") as trailer_image, Image.open(
        FILE1
```
If there's only one parameter, it doesn't seem necessary to parametrize the test.

radarhere:
I'm guessing you created the test image by opening up one of our existing test images in a text editor and modifying it manually?

nopperl:
> I'm guessing you created the test image by opening up one of our existing test images in a text editor and modifying it manually?

Yes, I simply modified `zero_bb.eps` because I can't distribute the images I've worked with. Should it be a "real" image instead?

radarhere:
Not necessarily, no. Real images are certainly ideal, but created ones are also accepted, as it's not always possible/practical to find an image that triggers a certain scenario.

src/PIL/EpsImagePlugin.py
```diff
@@ -243,6 +245,34 @@ def check_required_header_comments():
...
+                msg = "not an EPS file"
+                raise SyntaxError(msg) from e
+
+            if m:
+                k, v = m.group(1, 2)
+                self.info[k] = v
+                if k == "BoundingBox":
+                    if v == "(atend)":
+                        reading_trailer_comments = True
+                    elif not self._size or reading_trailer_comments:
```
radarhere, line 262:
```suggestion
                    elif not self._size or (
                        trailer_reached and reading_trailer_comments
                    ):
```
I've realised I made a mistake in https://github.com/nopperl/Pillow/pull/1 - the last BoundingBox has priority not just if there was an "(atend)" value, but if there was an "(atend)" value AND we are now in the trailer.

radarhere (approved):

PR #7383, opened 2023-09-07
Prevent TIFF orientation from being applied more than once

### LINKED ISSUE #7381: ImageOps.exif_transpose not expected behaviour for TIFF images
### What did you do?
I have a simple tiff image 150x100px rotated 90 degrees according to exif tag: 

[sample-out.zip](https://github.com/python-pillow/Pillow/files/12549680/sample-out.zip)

```bash
exiftool -n sample-out.tif
...
Image Width                     : 150
Image Height                    : 100
Orientation                     : 6
...
```
I tried to rotate the image in Pillow according to EXIF Orientation tag.

New resolution is 100x150, correct orientation

### What actually happened?
New resolution is 150x100, image transposed wrong way

* OS: Ubuntu
* Python: 3.10
* Pillow: 9.3.0

```python
im = Image.open('sample-out.tif') 
print(im.size) # (150, 100) - correct
im = ImageOps.exif_transpose(im)
print(im.size) # (150, 100) - not correct
im.save('output.tif') # you may save and see that now image rotated 540 degrees and does not contain
```

```bash
exiftool -n output.tif
...
Image Width                     : 150
Image Height                    : 100
# there is not Orientation tag anymore
...
```

### Reason

``Image.transpose`` calls ``Image.load()``.
``Image.load()``calls ``TiffImageFile.load()``
``TiffImageFile.load()`` calls ``ImageFile.load()``
``ImageFile.load()`` calls ``TiffImageFile.load_end()``
``load_end()`` transposes image. 

After ``Image.load()`` is done, image is transposed again.

radarhere:
Resolves #7381

When a TIFF image is loaded, the orientation is applied.

https://github.com/python-pillow/Pillow/blob/52c6d686135e13df23f05cfe34d393b9d6e981f6/src/PIL/TiffImagePlugin.py#L1545
https://github.com/python-pillow/Pillow/blob/52c6d686135e13df23f05cfe34d393b9d6e981f6/src/PIL/TiffImagePlugin.py#L1205-L1218

However, there is no change in Pillow's information about the image as a result of this. This is not ideal, because
1. if a user inspects the `tag_v2` orientation for an image, it may or may not have been applied already, depending on whether the image has loaded, and that could be confusing.
2. If a user runs `ImageOps.exif_transpose()` afterwards, the orientation is applied a second time.

This PR addresses 1 by deleting the `tag_v2` orientation tag, so that it is consistently present if not applied yet, and removed once it has been applied. It address 2 by calling `ImageOps.exif_transpose(self, in_place=True)` from TiffImagePlugin, which also reduces duplicate code.

The issue also calls `exif_transpose()` immediately after opening an image, without `load()`. So it is necessary to call `load()` the image in `exif_transpose()`, so that the image's orientation is not initially detected by `exif_transpose`, but then quickly changed once the image is loaded to perform a transposition.

### DIFF
--- a/src/PIL/ImageOps.py
+++ b/src/PIL/ImageOps.py
@@ -588,6 +588,7 @@ def exif_transpose(image, *, in_place=False):
         with the transposition applied. If there is no transposition, a copy of the
         image will be returned.
     """
+    image.load()
     image_exif = image.getexif()
     orientation = image_exif.get(ExifTags.Base.Orientation)
     method = {
--- a/src/PIL/ImageOps.py
+++ b/src/PIL/ImageOps.py
@@ -611,6 +611,10 @@ def exif_transpose(image, *, in_place=False):
         exif = exif_image.getexif()
         if ExifTags.Base.Orientation in exif:
             del exif[ExifTags.Base.Orientation]
+            if in_place and ExifTags.Base.Orientation in getattr(
+                exif_image, "tag_v2", {}
+            ):
+                del exif_image.tag_v2[ExifTags.Base.Orientation]
             if "exif" in exif_image.info:
                 exif_image.info["exif"] = exif.tobytes()
             elif "Raw profile type exif" in exif_image.info:
--- a/src/PIL/ImageOps.py
+++ b/src/PIL/ImageOps.py
@@ -611,10 +611,6 @@ def exif_transpose(image, *, in_place=False):
         exif = exif_image.getexif()
         if ExifTags.Base.Orientation in exif:
             del exif[ExifTags.Base.Orientation]
-            if in_place and ExifTags.Base.Orientation in getattr(
-                exif_image, "tag_v2", {}
-            ):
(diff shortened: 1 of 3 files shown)


src/PIL/ImageOps.py
```diff
@@ -610,6 +611,10 @@ def exif_transpose(image, *, in_place=False):
         exif = exif_image.getexif()
         if ExifTags.Base.Orientation in exif:
             del exif[ExifTags.Base.Orientation]
+            if in_place and ExifTags.Base.Orientation in getattr(
+                exif_image, "tag_v2", {}
```
homm, line 615: Isn't this format-specific? Probably move it to `load_end`?

radarhere: Oh, because `load()` is always run by the time we get here. Yep, good call, I've pushed a commit for that.

homm (approved):

PR #7392, opened 2023-09-12
Added gs_binary to control Ghostscript use on all platforms

### LINKED ISSUE #7391: Provide a documented way to turn off EPS handling via ghostscript
Pillow parses EPS files through ghostscript.

Ghostscript is troubled with regular security issues of critical impact, such as remote code execution. There has been [this one](https://github.com/farisv/PIL-RCE-Ghostscript-CVE-2018-16509), two CVEs this year already, and there is another one [just coming up](https://git.ghostscript.com/?p=ghostpdl.git;a=commit;h=e59216049cac290fb437a04c4f41ea46826cfba5) with the patch not yet included in any release.

We do not need EPS handling and we can turn it off in our code using `Image.open(…, formats=["JPG", "PNG", …])`. However, we use third-party components that also use Pillow (such as `reportlab`) and we'd like to make sure they also don't accidentally open EPS files and we can't modify the code.

From what I can tell, we can accomplish this with:

```
from PIL import Image

if "EPS" in Image.ID:
    Image.ID.remove("EPS")
```

However, that seems to be undocumented and feels like unclean monkeypatching. I think it would be worth it providing a documented API to turn off third-party backends like ghostscript globally.

radarhere:
Resolves #7391 

The user in that issue would like an API to prevent Ghostscript use. Since we already have `EpsImagePlugin.gs_windows_binary` that stores the name of the Ghostscript binary on Windows,
https://github.com/python-pillow/Pillow/blob/e1889544cdd959c60e2526505053e81a8f43c7e4/src/PIL/EpsImagePlugin.py#L40
this PR suggests adding `EpsImagePlugin.gs_binary`, that stores the name of the Ghostscript binary on all platforms.

Setting it to false would result in `OSError: Unable to locate Ghostscript on paths` when trying to load an EPS image.

### DIFF
--- a/docs/handbook/image-file-formats.rst
+++ b/docs/handbook/image-file-formats.rst
@@ -88,8 +88,14 @@ in ``L``, ``RGB`` and ``CMYK`` modes.
 Loading
 ~~~~~~~
 
+To use Ghostscript, Pillow searches for the "gs" executable. On Windows, it
+also searches for "gswin32c" and "gswin64c". If you would like to customise
+this behaviour, ``EpsImagePlugin.gs_binary = "gswin64"`` will set the name of
+the executable to use. ``EpsImagePlugin.gs_binary = False`` will prevent
+Ghostscript from being used altogether.
+
 If Ghostscript is available, you can call the :py:meth:`~PIL.Image.Image.load`
-method with the following parameters to affect how Ghostscript renders the EPS
+method with the following parameters to affect how Ghostscript renders the EPS.
 
 **scale**
     Affects the scale of the resultant rasterized image. If the EPS suggests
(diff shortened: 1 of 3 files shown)


docs/handbook/image-file-formats.rst
```diff
@@ -88,8 +88,14 @@ in ``L``, ``RGB`` and ``CMYK`` modes.
 Loading
 ~~~~~~~
 
+To use Ghostscript, Pillow searches for the "gs" executable. On Windows, it
+also searches for "gswin32c" and "gswin64c". If you would like to customise
```
hugovk, line 92: Bit more concise:

```suggestion
also searches for "gswin32c" and "gswin64c". To customise
```

hugovk (approved):

docs/handbook/image-file-formats.rst
radarhere: Sure. I've pushed a commit so that I could rearrange the lines ending as well, and shortened another bit of text later on.

PR #7403, opened 2023-09-17
NumPy now supports Python 3.12

radarhere:
With the release of NumPy 1.26.0, NumPy supports Python 3.12 - https://github.com/numpy/numpy/releases/tag/v1.26.0

### DIFF
--- a/.ci/install.sh
+++ b/.ci/install.sh
@@ -38,8 +38,7 @@ python3 -m pip install -U pytest-timeout
 python3 -m pip install pyroma
 
 if [[ $(uname) != CYGWIN* ]]; then
-    # TODO Remove condition when NumPy supports 3.12
-    if ! [ "$GHA_PYTHON_VERSION" == "3.12-dev" ]; then python3 -m pip install numpy ; fi
+    python3 -m pip install numpy
 
     # PyQt6 doesn't support PyPy3
     if [[ $GHA_PYTHON_VERSION == 3.* ]]; then
--- a/.github/workflows/macos-install.sh
+++ b/.github/workflows/macos-install.sh
@@ -13,8 +13,7 @@ python3 -m pip install -U pytest-cov
 python3 -m pip install -U pytest-timeout
 python3 -m pip install pyroma
 
-# TODO Remove condition when NumPy supports 3.12
-if ! [ "$GHA_PYTHON_VERSION" == "3.12-dev" ]; then python3 -m pip install numpy ; fi
+python3 -m pip install numpy
 
 # extra test images
 pushd depends && ./install_extra_test_images.sh && popd


hugovk:
Hmm, NumPy has wheels for PyPy3.9 but not for PyPy3.10:

* numpy-1.26.0-pp39-pypy39_pp73-win_amd64.whl
* numpy-1.26.0-pp39-pypy39_pp73-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
* numpy-1.26.0-pp39-pypy39_pp73-macosx_10_9_x86_64.whl

https://pypi.org/project/numpy/1.26.0/#files

hugovk:
I asked NumPy about PyPy3.10 wheels: https://github.com/numpy/numpy/issues/24728

radarhere:
I've added commits to fix building on PyPy 3.10 in the meanwhile.

As for Python 3.9 on Cygwin, the initial error is "[Failed to build ninja](https://github.com/python-pillow/Pillow/actions/runs/6224303804/job/16892212199#step:11:52)".
[Installing that](https://github.com/radarhere/Pillow/commit/10011ab84392d65a684f07c5ec78e774b3054d92), I get "[No BLAS library detected! Install one, or use the `allow-noblas` build option](https://github.com/radarhere/Pillow/actions/runs/6244593251/job/16951756051#step:11:76)". 

https://stackoverflow.com/questions/73659464/does-openblas-devel-exist-for-cygwin tells me that there is no devel openblas package for Cygwin, so I attempted to use 'allow-noblas'.

To do this, https://github.com/numpy/numpy/issues/24703 suggests
- [`pip install numpy -C-Dallow-noblas=true`](https://github.com/numpy/numpy/issues/24703#issuecomment-1718988549), but that gives '[meson-python: error: Unknown option "-Dallow-noblas"](https://github.com/radarhere/Pillow/actions/runs/6244597341/job/16951767040#step:11:24)'
or
- [`pip install numpy --config-settings=setup-args="-Dallow-noblas=true"`](https://github.com/numpy/numpy/issues/24703#issuecomment-1722379388), but that gives '[../../meson.build:1:0: ERROR: Value true
       is not boolean (true or false).](https://github.com/radarhere/Pillow/actions/runs/6244601995/job/16951778973#step:11:31)'. That appears to be a problem with the line ending, which [makes me think it is specific to Cygwin](https://github.com/python-pillow/Pillow/blob/89340819cfe77b9680060983b49e91a73a5c94e9/.github/workflows/test-cygwin.yml#L34-L36).

Instead, I have pushed a commit here to simply not use NumPy 1.26 on Cygwin.

hugovk:
Thanks!

PR #7406, opened 2023-09-18
Revert #7311: "Read WebP duration after opening"

radarhere:
Reverts #7311, after [a comment requesting that we avoid an expensive operation.](https://github.com/python-pillow/Pillow/pull/7311#issuecomment-1723004045)
> For me this is one of the key concepts in Pillow: you can safely open the image file without decoding the whole image (which is much more expensive operation) and get any info from the file.


homm:
I will try to test does this really affects opening time vs opening & decoding time.

homm:
Things even worse than I thought. With the current `main` the loading time is actually doubled.


### Pillow-10.0.1

```python
In [2]: %timeit im = Image.open('/Code/imgs/avif/space.webp')
711 µs ± 5.68 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [3]: %timeit im = Image.open('/Code/imgs/avif/space.webp').load()
14.8 ms ± 243 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [4]: Image.open('/Code/imgs/avif/space.webp').size
Out[4]: (1500, 1000)
```

### python-pillow:main

```python
In [2]: %timeit im = Image.open('/Code/imgs/avif/space.webp')
13.4 ms ± 77 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [3]: %timeit im = Image.open('/Code/imgs/avif/space.webp').load()
26.6 ms ± 73.6 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

### radarhere:webp

```python
In [2]: %timeit im = Image.open('/Code/imgs/avif/space.webp')
714 µs ± 13.1 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [3]: %timeit im = Image.open('/Code/imgs/avif/space.webp').load()
14.5 ms ± 56 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

homm (approved):

PR #7407, opened 2023-09-18
AppVeyor: don't download huge pillow-depends.zip

hugovk:
On AppVeyor, we're downloading a zip of the entire contents of https://github.com/python-pillow/pillow-depends/ mirror which is 851 MB and on my machine takes 6 minutes. We don't use all the files so don't need to download them all.

`extract_dep` of `winbuild/build_prepare.py` first checks if there's a local copy, and if there's not, downloads it from the upstream source.

Instead of downloading the whole mirror, I've modified `extract_dep`: if there's not a local copy, it first tries to download from our mirror. If that doesn't succeed, it tried the upstream source.

Also add some type hints and capitalise constants. 

---

Also, curl has started failing to download with:

```
curl -fsSL -o pillow-depends.zip https://github.com/python-pillow/pillow-depends/archive/main.zip
curl: (35) schannel: next InitializeSecurityContext failed: Unknown error (0x80092013) - The revocation function was unable to check revocation because the revocation server was offline.
Command exited with code 35
``` 

https://ci.appveyor.com/project/Python-pillow/pillow/builds/48061132/job/i2a2jg124obvmoud

Adding `--ssl-no-revoke` fixes it:

* https://stackoverflow.com/a/56486639/724176
* https://stackoverflow.com/a/63996289/724176


radarhere:
AppVeyor has started passing again - https://ci.appveyor.com/project/Python-pillow/pillow/builds/48066083. I think that means adding `--ssl-no-revoke` is no longer necessary?

hugovk:
Good, I've dropped the `--ssl-no-revoke` commit.

winbuild/build_prepare.py
```diff
@@ -421,25 +430,41 @@ def find_msvs():
...
     import urllib.request
+
+    ex = None
+    for i in range(3):
+        try:
+            print(f"Fetching {url} (attempt {i + 1})...")
+            content = urllib.request.urlopen(url).read()
+            with open(file, "wb") as f:
+                f.write(content)
+            break
```
radarhere, line 443:
```suggestion
            ex = None
            break
```
You've changed the `for`/`else` to "If `ex` is truthy, raise".

There's actually a difference here, that I presume was unintended. The `for`/`else` option only raises the error if it doesn't break after success. This version might fail on the first loop, then succeed on the second loop, but because `ex` is still populated, it will still raise afterwards.

If you prefer this style to `for`/`else`, I suggest inserting a `ex = None` before `break`.

hugovk: Yep, it was unintended. I originally refactored it to return instead of raising, but then put the `raise` back. So I've put the `for`/`else` back too. Thanks!

.appveyor.yml
```diff
@@ -21,13 +21,11 @@ environment:
...
 - '%PYTHON%\%EXECUTABLE% -m pip install --upgrade pip'
-- curl -fsSL -o pillow-depends.zip https://github.com/python-pillow/pillow-depends/archive/main.zip
-- curl -fsSL -o pillow-test-images.zip https://github.com/python-pillow/test-images/archive/main.zip
-- 7z x pillow-depends.zip -oc:\
+- curl -fsSL --ssl-no-revoke -o pillow-test-images.zip https://github.com/python-pillow/test-images/archive/main.zip
 - 7z x pillow-test-images.zip -oc:\
-- mv c:\pillow-depends-main c:\pillow-depends
 - xcopy /S /Y c:\test-images-main\* c:\pillow\tests\images
-- 7z x ..\pillow-depends\nasm-2.16.01-win64.zip -oc:\
+- curl -fsSL --ssl-no-revoke -o nasm-win64.zip https://raw.githubusercontent.com/python-pillow/pillow-depends/main/nasm-2.16.01-win64.zip
```
radarhere, line 27: `--ssl-no-revoke` came back somehow.

radarhere (approved):

PR #7413, opened 2023-09-21
Set blue channel to 128 for BC5S

### LINKED ISSUE #7386: DDS BC5_SNORM images are decoded incorrectly
### What did you do?

I loaded a BC5 SNORM images.

For it to be decoded correctly.

image

### What actually happened?

Pillow incorrectly decoded the image. It added artifacts and washed out the "colors" of the stored normals.

This is speculation on my part (I haven't read PIL's source code), but it seems like PIL uses the same algorithm for BC5_UNORM (ATI2) and BC5_SNORM.

image

* OS: Win 10 64 bit
* Python: 3.8.2
* Pillow: 9.2.0 and 10.0.0

```python
# it's pretty much as basic as it can get
im = Image.open(dds_path)
img = np.array(im)
```

The DDS file in question: [cables_normal.zip](https://github.com/python-pillow/Pillow/files/12568098/cables_normal.zip)

radarhere:
Resolves #7386

As per https://learn.microsoft.com/en-us/windows/win32/direct3d10/d3d10-graphics-programming-guide-resources-block-compression#bc5, the BC5 (both signed and unsigned) formats only store red and green values. Currently, Pillow takes this to mean that the blue channel is set to zero.

However, the user in the issue has found that [texconv](https://github.com/Microsoft/DirectXTex/wiki/Texconv), a Microsoft tool, presumes that for BC5S (signed BC5) the blue channel is always 128.


RunDevelopment (approved):
LGTM; 

Thank you @radarhere!

PR #7274, opened 2023-07-10
Improved flexibility of XMP parsing

radarhere:
Resolves #7273
Resolves #7324

The JPEG XMP code was originally added in #5144. The image in #7273 differs in that it is padded at the end with b"\x00". I've updated the JpegImagePlugin code to allow for that.

The image in #7273 also has an attribute name that doesn't contain a semicolon. Normally, the tags do, and leads to the start of the tag being surrounded with curly brackets. Their absence causes an error at the moment, so I've also added a commit to allow for that.

I created the test files by copying the XMP data from Tests/images/xmp_test.jpg, modifying it to trigger these conditions, and saving new images with a modified Pillow.

PR #7416, opened 2023-09-22
Fix JpegImagePlugin when XMP meta has trailing 0x00 byte

kerozin:
I faced the issue with files having trailing 0x00 in xmp meta information. The files was originally generated with Canon Digital Photo Professional and later edited in MacOs Preview to add image description to the meta information. I haven't investigated which program exactly added trailing zero byte but fix is dealing with it fine

Add xmp content stripping 0x00 bytes

### DIFF
--- a/src/PIL/JpegImagePlugin.py
+++ b/src/PIL/JpegImagePlugin.py
@@ -496,7 +496,7 @@ class JpegImageFile(ImageFile.ImageFile):
 
         for segment, content in self.applist:
             if segment == "APP1":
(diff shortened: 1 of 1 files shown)


radarhere:
Are you able to add a test our test suite, that fails without this fix and passes with it?

If not, could you attach the image here?

kerozin:
Alas I'm not yet familiar with the testing system. Here is the file to reproduce the issue and make the test.
P329

radarhere:
This has been fixed instead by #7274

PR #7420, opened 2023-09-25
Added has_transparency_data

### LINKED ISSUE #7384: Method to test if image has transparency/alpha
[**2023/09/11 edit**] adding a tl;dr:
1. Online sources do not agree on a single method to check for either transparency (actual transparent pixels) or the presence of an alpha channel (whether or not it is opaque), and for all the examples I've found, it's easy to craft examples on which they fail. It would be useful to have an authoritative correct solution from someone who really understands Pillow/images that handles all the edge cases. Or in lieu of that, we could crowd-source one here.
2. Given the frequency of the question and the widespread incorrect code, it seems like it'd be useful to add this feature to Pillow itself.

Original post:

I found myself wanting to check if an image had an alpha channel/transparency (I'm allowing for a distinction between those, perhaps there isn't a meaningful one). I was trying to convert some optimised images that didn't contain any transparency, so wanted to do something different in that case.

Checking for this is slightly tricky though, and I've seen incorrect logic posted elsewhere, e.g. [this stackoverflow post](https://stackoverflow.com/a/35859141/5506429) references [_this_ stackoverflow post](http://stackoverflow.com/a/1963146) which states to use:
```python
img.mode in ('RGBA', 'LA') or (img.mode == 'P' and 'transparency' in img.info)
```

But palette-mode images can store their transparency in at least two ways, in `img.info`, or the palette itself might be `RGBA`. If you run the following
```python
img = Image.new("RGBA", (100, 100), (128, 128, 128, 128)).quantize()
print(img.mode in ('RGBA', 'LA') or (img.mode == 'P' and 'transparency' in img.info)) # False
```

the result is `False`, but clearly it has alpha values in the palette:
```python
print(img.palette.mode) # 'RGBA'
print(img.palette.colors) # {(128, 128, 128, 128): 0, (0, 0, 0, 0): 1}
```

Would it be possible to add something like `img.has_alpha()` or `img.has_transparency()` to help here?

I'm not familiar enough with Pillow's image formats to say whether checking the mode of the palette in addition to the checks above would be exhaustive. Perhaps someone can clarify? Is it sufficient to check `img.palette.mode == 'RGBA'`? `.quantize()` does not seem to like an `LA` image, but if I manually set the palette mode and colors, it works with `show()`. Then there are other esoteric modes with alpha channels. So perhaps something like this will work?
```python
alpha_modes = {"RGBA", "RGBa", "LA", "La", "PA"}
return img.mode in alpha_modes or (
    img.mode == "P"
    and ("transparency" in img.info or img.palette.mode in alpha_modes)
)
```
but I very easily could have missed something due to inexperience with the unusual image formats.

(Happy to have a go at the implementation if that would be of assistance!)

radarhere:
Resolves #7384

Adds a method to determine if an image has transparency data - an alpha channel, an RGBA palette or "transparency" in the image's `info` directory.

I chose this method name to try and indicate that the image may still be opaque. For example, `hopper("RGB").convert("RGBA")` has an alpha channel, but has no transparent or translucent pixels.

In doing so, I found a similar check in WebPImagePlugin.
https://github.com/python-pillow/Pillow/blob/b723e9e62e4706a85f7e44cb42b3d838dae5e546/src/PIL/WebPImagePlugin.py#L335-L340

So I've changed that to use this new method.

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -915,7 +915,7 @@ class Image:
 
         self.load()
 
-        has_transparency = self.info.get("transparency") is not None
+        has_transparency = "transparency" in self.info
         if not mode and self.mode == "P":
             # determine default mode
             if self.palette:
@@ -1530,6 +1530,23 @@ class Image:
         if rawmode is None:
             rawmode = mode
         return list(self.im.getpalette(mode, rawmode))
+
+    def has_transparency_data(self):
+        """
+        Determine if an image has transparency data, whether in the form of an
+        alpha channel, a palette with an alpha channel, or a "transparency" key
+        in the info dictionary.
+
+        Note the image might still appear solid, if all of the values shown
+        within are opaque.
+
+        :returns: A boolean.
+        """
+        return (
+            self.mode in ("LA", "La", "PA", "RGBA", "RGBa")
+            or (self.mode == "P" and self.palette.mode == "RGBA")
+            or "transparency" in self.info
+        )
 
     def apply_transparency(self):
         """
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -1531,7 +1531,7 @@ class Image:
             rawmode = mode
         return list(self.im.getpalette(mode, rawmode))
 
-    def has_transparency_data(self):
+    def has_transparency_data(self) -> bool:
         """
         Determine if an image has transparency data, whether in the form of an
         alpha channel, a palette with an alpha channel, or a "transparency" key
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -1531,6 +1531,7 @@ class Image:
             rawmode = mode
         return list(self.im.getpalette(mode, rawmode))
 
+    @property
     def has_transparency_data(self) -> bool:
         """
         Determine if an image has transparency data, whether in the form of an
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -1544,7 +1544,7 @@ class Image:
         """
         return (
             self.mode in ("LA", "La", "PA", "RGBA", "RGBa")
-            or (self.mode == "P" and self.palette.mode == "RGBA")
+            or (self.mode == "P" and self.palette.mode.endswith("A"))
             or "transparency" in self.info
         )
 
--- a/Tests/test_file_webp.py
+++ b/Tests/test_file_webp.py
@@ -234,3 +234,13 @@ class TestFileWebp:
 
         with Image.open(out_webp) as reloaded:
             assert reloaded.info["duration"] == 1000
+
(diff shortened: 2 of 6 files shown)


src/PIL/Image.py
```diff
@@ -1531,6 +1531,23 @@ def getpalette(self, rawmode="RGB"):
...
+        in the info dictionary.
+
+        Note the image might still appear solid, if all of the values shown
+        within are opaque.
+
+        :returns: A boolean.
+        """
+        return (
+            self.mode in ("LA", "La", "PA", "RGBA", "RGBa")
+            or (self.mode == "P" and self.palette.mode == "RGBA")
```
homm, line 1547: Correct way to check P mode from my application:

```python
        if self.mode == 'P':
            self.load()
            return self.im.getpalettemode().endswith('A')
```

I can't remember all details, I had problems with `self.palette.mode` and thus `self.im` and `self.load()` is required here.

radarhere: If `self.palette.mode` is going out of sync with `self.im.getpalettemode()`, I'm interested in investigating that for its own sake.

Is it possible your code was written before something like #5552, and `load()` is no longer necessary?

homm: > Is it possible your code was written before 

Definitely before. 

Anyway, I think `.endswith('A')` is more correct way to check alpha

chinery:
Hi @radarhere – thanks for this PR! My only question was going to be whether image palettes support modes other than RGBA, but you've addressed that in your comments to @homm, so from my perspective (of admittedly limited experience), I think it looks great.

src/PIL/Image.py
homm: > ImagePalette is only ever created internally

This could eventually change and `endswith('A')` will stay correct in any way.

src/PIL/Image.py
```diff
@@ -1531,6 +1531,23 @@ def getpalette(self, rawmode="RGB"):
             rawmode = mode
         return list(self.im.getpalette(mode, rawmode))
 
+    def has_transparency_data(self):
```
hugovk, line 1534: It would be good to gradually start adding type hints:

```suggestion
    def has_transparency_data(self) -> bool:
```

hugovk (commented):
This could also be a property, any preference?

src/PIL/Image.py
radarhere: Ok, I've added a commit.

radarhere:
Sure, that seems like a reasonable idea. I've added a commit to change it to a property.

homm (approved):

hugovk:
Looks good, let's also add to the release notes.

radarhere:
Sure, I've pushed a commit for that.

hugovk (approved):
Thank you!

PR #7421, opened 2023-09-25
Add checks to pre-commit

hugovk:
* Add https://github.com/asottile/pyupgrade and upgrade Python syntax
* Update Black and switch to the [mirror repo](https://github.com/psf/black-pre-commit-mirror) for faster mypyc wheels
* Add https://github.com/adamchainz/flake8-logging and fix logging issues
* From https://github.com/pre-commit/pre-commit-hooks:
  * Add `check-executables-have-shebangs` and remove executable flags
  * Add `end-of-file-fixer`, excluding `Tests/images/`
  * Add `trailing-whitespace`, excluding some templates, test fonts and images


Tests/helper.py
```diff
@@ -126,7 +126,7 @@ def assert_image_similar(a, b, epsilon, msg=None):
         if HAS_UPLOADER:
             try:
                 url = test_image_results.upload(a, b)
-                logger.error(f"Url for test images: {url}")
+                logger.exception("URL for test images: %s", url)
```
radarhere, line 129: Why is this `logger.exception` but the earlier change in this file is still `logger.error`?

.pre-commit-config.yaml
```diff
@@ -44,10 +50,15 @@ repos:
...
     hooks:
+      - id: check-executables-have-shebangs
       - id: check-merge-conflict
       - id: check-json
       - id: check-toml
       - id: check-yaml
+      - id: end-of-file-fixer
+        exclude: ^Tests/images/
+      - id: trailing-whitespace
+        exclude: ^.github/.*TEMPLATE|^Tests/(fonts|images)/
```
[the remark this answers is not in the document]
hugovk: Hmm, it is fixing the text file in this directory and not touching the binaries. Shall we leave it as is, until it comes up?

Tests/helper.py
hugovk: Because it's inside an `except` block (see line 125), the nested `try` isn't relevant.

Compare:

```pycon
>>> import logging
>>>
>>> # both `error`, don't get any extra exception info:
>>> try:
...     logging.error(12)
...     raise ValueError
... except ValueError:
...     logging.error(34)
...
ERROR:root:12
ERROR:root:34
>>>
```
```pycon
>>> # `exception` in the `except` gives extra info:
>>> try:
...     logging.error(12)
...     raise ValueError
... except ValueError:
...     logging.exception(34)
...
ERROR:root:12
ERROR:root:34
  File "<stdin>", line 3, in <module>
ValueError
>>>
```
```pycon
>>> # `exception` in the `try` gives unhelpful info:
>>> try:
...     logging.exception(12)
...     raise ValueError
... except ValueError:
...     logging.error(34)
...
ERROR:root:12
NoneType: None
ERROR:root:34
>>>
```

radarhere: But line 94 and line 129 are both within `try` blocks with `except Exception`. I'm not seeing why one should be `logger.error` and the other `logger.exception`.

hugovk: It needs to be in the `except` part:

```python
try:
    logger.error(123)
except:
    logger.exception(456)  # <-- here
```

Only at 456 has an exception occurred, and there's exception info to be logged.

The first one only has the logger in the `try` block, not the `except` block:

```python
            try:
                url = test_image_results.upload(a, b)
                logger.error("URL for test images: %s", url)  # <-- no exception info here
            except Exception:
                pass  # <-- there's exception info here, if we wanted to log it
```

The second one:

```python
    try:
        assert epsilon >= ave_diff, (
            (msg or "")
            + f" average pixel value difference {ave_diff:.4f} > epsilon {epsilon:.4f}"
        )
    except Exception as e:
        if HAS_UPLOADER:
            try:
                url = test_image_results.upload(a, b)
                logger.exception("URL for test images: %s", url)  # <-- has exception info from the outer try/except
            except Exception:
                pass
        raise e
```

The `logger.exception` is in the `try` of the inner one, but that doesn't matter. What matters is it's in the `except` of the outer one.

Simplifying a bit, this is what matters:

```python
    try:
        ...
    except Exception as e:
        logger.exception("URL for test images: %s", url)  # <-- exception info here
```

Does that make sense?

radarhere:
> * Update Black and switch to the [mirror repo](https://github.com/psf/black-pre-commit-mirror) for faster mypyc wheels

For the record, or maybe just as a note to self, this is recommended at https://github.com/psf/black/blob/3dcacdda0d7f69a1705f3e2a151c24a6cf004171/.pre-commit-hooks.yaml#L1-L2
> \# Note that we recommend using https://github.com/psf/black-pre-commit-mirror instead
> \# This will work about 2x as fast as using the hooks in this repository

Tests/helper.py
radarhere: Ok, thanks. You were explaining it clearly, I just wasn't getting it.

hugovk:
Also in the docs here:

https://black.readthedocs.io/en/stable/integrations/source_version_control.html

PR #7426, opened 2023-09-27
Added reading 8-bit color DDS images

### LINKED ISSUE #7423: Unable to open DDS image: Unimplemented pixel format
### What did you do?

Attempted to read the attached DDS image.

Expected it to render in my tkinter GUI.

### What actually happened?

Received the following error:

```
NotImplementedError: Unimplemented pixel format b'\x00\x00\x00\x00'
```

* OS: Windows 11
* Python: 3.11 64-bit
* Pillow: 10.0.1

```python
from PIL import ImageTk

with open ("test.dds", "rb") as f:
  data = f.read()

image = ImageTk.PhotoImage(data=data)
```

### Additional notes

I'm guessing the version of DDS that I am attempting to render is not supported solely based off of the error message. I am passing byte data and not the files themselves, but am unsure of how else to get this to render. Attempted to use Pillow, but no luck.

Attached zip with dds image inside:
- [test.zip](https://github.com/python-pillow/Pillow/files/12721995/test.zip)

radarhere:
Resolves #7423 

The issue has found a DDS image that has the `DDPF_PALETTEINDEXED8` pixel format
https://github.com/python-pillow/Pillow/blob/b723e9e62e4706a85f7e44cb42b3d838dae5e546/src/PIL/DdsImagePlugin.py#L46-L50

This pixel format is not mentioned at https://learn.microsoft.com/en-us/windows/win32/direct3ddds/dds-pixelformat under "dwFlags", but is at https://learn.microsoft.com/en-us/previous-versions/ms783269(v=vs.85). It is an "8-bit color indexed" image.

Looking at the data in the issue, I guessed that the file structure was the header, followed by 1024 palette bytes, followed by the indexes to the palette. That produced a good result.

To create a test image, I took the header from [hopper.dds](https://github.com/python-pillow/Pillow/blob/main/Tests/images/hopper.dds), manually changed the width, height and pixel format flags, and put together code to write out the palette and indexes of [transparent.gif](https://github.com/python-pillow/Pillow/blob/main/Tests/images/transparent.gif).


homm (approved):

hugovk (approved):
Thanks! 

Worth adding to release notes or documenting elsewhere?

radarhere:
Sure, I've added it to the release notes and the file formats documentation.

docs/handbook/image-file-formats.rst
```diff
@@ -64,7 +64,8 @@ DDS
 
 DDS is a popular container texture format used in video games and natively supported
 by DirectX. Uncompressed RGB and RGBA can be read, and (since 8.3.0) written. DXT1,
-DXT3 (since 3.4.0) and DXT5 pixel formats can be read, only in ``RGBA`` mode.
+DXT3 (since 3.4.0) and DXT5 pixel formats can be read, only in ``RGBA`` mode. Since
+10.1.0, 8-bit color indexed files can be read as ``P`` mode images.
```
hugovk, line 68: Perhaps with `versionadded`?

```suggestion
DXT3 (since 3.4.0) and DXT5 pixel formats can be read, only in ``RGBA`` mode.

.. versionadded:: 10.1.0

8-bit color indexed files can be read as ``P`` mode images.
```

Or attempting to rewrite to avoid beginning a sentence with a numeral?

```suggestion
DXT3 (since 3.4.0) and DXT5 pixel formats can be read, only in ``RGBA`` mode.

.. versionadded:: 10.1.0

Files indexed with 8-bit color can be read as ``P`` mode images.
```

radarhere: From what I see in https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html, the tradition is for 'versionadded' to go below the new feature.

I've pushed a commit.

hugovk: The [Sphinx docs](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionadded) shows the text afterwards:

> The first argument must be given and is the version in question; you can add a second argument consisting of a _brief_ explanation of the change.
> 
> Example:
```rst
.. versionadded:: 2.5
   The *spam* parameter.
```


For example:

```rst
.. versionadded:: 3.4.0
   DXT3 images can be read in ``RGB`` mode and DX10 images can be read in ``RGB``
      and ``RGBA`` mode.
```

Gives:

image

Which I think is clearer.

hugovk:
Thanks!

PR #7427, opened 2023-09-29
Added support for Windows 1.0 icons

### LINKED ISSUE #7320: Add support for Windows 1.x/2.x Icon file format
### Is your feature request related to a feature request? Please describe.

Be able to view, edit or convert `.ICO` files from Windows 1.x/2.x.

http://fileformats.archiveteam.org/wiki/Windows_1.0_Icon

Sample file: CARDFILE.ICO opened by ICONEDIT.EXE (attachment below):

CARDFILE ICO-SCR

The data are the black&white pixels, the grey zones are believed to be transparent.

The icon of the program CARDFILE.EXE (CARDFILE.ICO) can be seen in the taskbar, bottom left, when CARDFILE.EXE is minimized:

CARDFILE ICO-MINIMIZED

and sometimes in the About dialog, specially in Microsoft programs:

CARDFILE-ABOUT

According to this website, 1.x and 2.x share same icons:

https://www.howtogeek.com/733912/a-visual-history-of-windows-icons-from-windows-1-to-11/

### Describe the solution you'd like

Be able to convert from and to ICO file format (Windows 1.x/2.x). For example from and to PNG, TIFF.

### Describe alternatives you've considered

GIMP and ImageMagick don't support it yet:
- https://gitlab.gnome.org/GNOME/gimp/-/issues/9814
- https://github.com/ImageMagick/ImageMagick/issues/6527

### Samples:

[CARDFILE.ICO.ZIP](https://github.com/python-pillow/Pillow/files/12261632/CARDFILE.ICO.ZIP)

radarhere:
Resolves #7320 by adding support for reading and writing device-independent Windows 1.0 icons.

http://fileformats.archiveteam.org/wiki/Windows_1.0_Icon was used as a reference for the specification.

Because these images contains white, black and transparent pixels, I chose LA for the mode.

The test image was created by the user in the issue.

Because .ico is already taken, the format needs to be passed in as an argument when saving.
```python
im.save("newimage.ico", format="ICO1")
```


superbonaci:
Check out the samples as device dependent and independent in #7320

radarhere:
Looking at "DEVDEP.ICO", it's actually not device-dependent, but in "[both](http://fileformats.archiveteam.org/wiki/Windows_1.0_Icon)" formats.

I presume that it is also able to be added to our test suite, and distributed under the Pillow license? I've added that here as a test image for the "both" format, and manually changed the image I already had to register as device-dependent.

So this PR now reads device-independent, device-dependent, "both", and saves in device-independent format.

superbonaci:
Yes, you can add those icons of course to the test suite under any license. They are just randomly generated pictures trying to include all bitmaps, white, black and transparent. Not sure if the format could be generated more complicated or complete that that to include all use cases. I encourage anybody to try with `ICONEDIT.EXE` Microsoft editor.

akx:
Ah, this duplicates some of the work I'd done in https://github.com/python-pillow/Pillow/pull/6965... :(

src/PIL/IcoImagePlugin.py
```diff
@@ -347,6 +348,85 @@ def load_seek(self):
...
+    def decode(self, buffer):
+        data = bytearray()
+        bitmapLength = self.state.xsize * self.state.ysize // 8
+        firstBitmap = self.fd.read(bitmapLength)
+        secondBitmap = self.fd.read(bitmapLength)
+        for i, byte in enumerate(firstBitmap):
+            secondByte = byte ^ secondBitmap[i]
+            for j in reversed(range(8)):
+                first = byte >> j & 1
+                second = secondByte >> j & 1
```
akx, line 385: `snake_case`?

src/PIL/IcoImagePlugin.py
```diff
@@ -347,6 +348,85 @@ def load_seek(self):
...
+            for x in range(w):
+                l, a = self.im.getpixel((x, y))
+                if x % 8 == 0:
+                    firstBitmap += b"\x00"
+                    secondBitmap += b"\xff"
+                if not a:
+                    firstBitmap[-1] ^= 1 << (7 - x % 8)
+                if not l:
+                    secondBitmap[-1] ^= 1 << (7 - x % 8)
+        data = firstBitmap + secondBitmap
```
akx, line 409: `snake_case`? Also, `l` and `a` are somewhat opaque to me.

src/PIL/IcoImagePlugin.py
radarhere: I've pushed a commit for this.

src/PIL/IcoImagePlugin.py
radarhere: I've pushed commits to update the variable names. It is an LA image being saved, so `l` is the L channel, and `a` the A channel.

superbonaci: So the master branch is already patched, or we must wait?

radarhere: If I understand you correctly, this PR has not been merged into Pillow's main branch, no. It is on hold because there have been [concerns raised about whether this really needs to be part of Pillow](https://github.com/python-pillow/Pillow/issues/7320#issuecomment-1741270231).

superbonaci: Isn't it possible to make it a plugin? Are plugins easy to install?

radarhere: If you just want to add it your own code, here is a separate plugin derived from this PR - [Ico1ImagePlugin.py.zip](https://github.com/python-pillow/Pillow/files/13358252/Ico1ImagePlugin.py.zip)

Unzip it, and then you can
```python
from PIL import Image
import Ico1ImagePlugin

im = Image.open("ico1both.ico")
```

akx: I think this would be useful in Pillow – people aren't going to find external plugins when they need them, so they'll end up writing the same code again, and that's no fun for anyone...

aclark4life: I think we've merged new features for less good reasons so I'd like to see this go in. Worst case it comes out again in the future? 🤷

superbonaci:
ImageMagick apparently added support for it but doesn't work: https://github.com/ImageMagick/ImageMagick/issues/6866

radarhere:
Closing due to [lack of interest.](https://github.com/python-pillow/Pillow/issues/7320#issuecomment-2029506669)

superbonaci:
//TODO

radarhere:
@superbonaci if that is a note to yourself, ok, but [the intention from our side](https://github.com/python-pillow/Pillow/issues/7320#issuecomment-2029506669) is that this is not something that will be done.

PR #7435, opened 2023-10-02
Add progress callback when save_all is used

### LINKED ISSUE #7433: feature request: progress indication for writing multiple frames
Processing all frames of an animated image file (i.e. GIF, APNG, WebP) can take a while, and there's no indication of any sort for progress there, nor is there any way to keep track of it. Thus, I suggest adding to `save_all` implementation at least one of the following:

* a log output (either at `INFO` or `DEBUG` level) after every frame (e.g. the way it's done by `convert -verbose`)
* an (optional) custom progress tracking callback parameter (which should be given the current frame index and possibly total `n_frames` & `filename` of the first file as well); this can be used for customizing log output or for GUI progress indication
  + log only when no progress tracking callback is provided?
  + alternatively, if there's a callback and it returns a string, print it out as logging output?

radarhere:
Resolves #7433

The issue requests the ability to use a callback function to monitor progress when calling `save_all`. This function is triggered after processing each frame, with the filename of the current image (or `None` if the filename is not available), the number of frames processed so far, and the total number of frames.

```python
from PIL import Image

def callback(filename, frame_count, n_frames):
    print((filename, frame_count, n_frames))

with Image.open("Tests/images/hopper.gif") as im:
    with Image.open("Tests/images/dispose_bgnd.gif") as im2:
        im.save("out.gif", save_all=True, append_images=[im2], progress=callback)
```
gives
```
('Tests/images/hopper.gif', 1, 6)
('Tests/images/dispose_bgnd.gif', 2, 6)
('Tests/images/dispose_bgnd.gif', 3, 6)
('Tests/images/dispose_bgnd.gif', 4, 6)
('Tests/images/dispose_bgnd.gif', 5, 6)
('Tests/images/dispose_bgnd.gif', 6, 6)
```


Tests/test_file_apng.py
```diff
@@ -663,6 +665,36 @@ def test_apng_save_blend(tmp_path):
...
+
+    with Image.open("Tests/images/apng/single_frame.png") as im:
+        with Image.open("Tests/images/apng/delay.png") as im2:
+            im.save(
+                out, "PNG", save_all=True, append_images=[im, im2], progress=callback
+            )
+
+    assert progress == [
+        ("Tests/images/apng/single_frame.png", 1, 7),
+        ("Tests/images/apng/single_frame.png", 2, 7),
```
LeXofLeviafan, line 689: This line could use a comment that `single_frame.png` gets appended to itself in the `save()` call

Tests/test_file_apng.py
```diff
@@ -663,6 +665,36 @@ def test_apng_save_blend(tmp_path):
...
 
+def test_save_all_progress():
+    out = BytesIO()
+    progress = []
+
+    def callback(filename, frame_number, n_frames):
+        progress.append((filename, frame_number, n_frames))
+
+    Image.new("RGB", (1, 1)).save(out, "PNG", save_all=True, progress=callback)
+    assert progress == [(None, 1, 1)]
```
LeXofLeviafan, line 676: Maybe indicate index here (instead of `None`)? `1` for the 1st input image?

radarhere: Would you be open to the idea of having an index all of the time? Or, we could pass back the image instance, and let the user figure out the index or filename as they see fit.

Tests/test_file_webp.py
```diff
@@ -122,6 +122,30 @@ def test_save_all(self, tmp_path):
...
+        progress = []
+
+        with Image.open("Tests/images/iss634.webp") as im:
+            im2 = Image.new("RGB", im.size)
+            im.save(out, "WEBP", save_all=True, append_images=[im2], progress=callback)
+
+        expected = []
+        for i in range(42):
+            expected.append(("Tests/images/iss634.webp", i + 1, 43))
+        expected.append((None, 43, 43))
```
LeXofLeviafan, line 146: Likewise, saying that the last frame comes from image #​2 would be more informative than "don't know" – especially when there's multiple images involved

Tests/test_file_apng.py
LeXofLeviafan: Having index always is probably simpler. As for passing the image, it wouldn't be much help with identifying the index if an image is reused in the input (like in that one test… come to think of it, having explicit index there would make adding a comment somewhat redundant).

radarhere: I've pushed a commit to switch to an index.

Tests/test_file_apng.py
radarhere: > come to think of it, having explicit index there would make adding a comment somewhat redundant

You no longer think this is necessary?

Tests/test_file_webp.py
radarhere: I've pushed a commit to switch to an index.

homm:
An idea: use the same callback for

homm:
An idea: use the same callback for generating frames: pass current `im` object  (which could be an image without bitmap in memory) and check if complete image is returned.

radarhere:
Sorry, I'm not sure I understand. Are you suggesting this
```python
from PIL import Image

colors_left = ["#f00", "#0f0"]
def callback():
    if colors_left:
        return Image.new("RGB", (1, 1), frames_left.pop(0))

im = Image.new("RGB", (1, 1))
im.save("out.png", save_all=True, progress=callback)
```
with the intention that it allow multiframe images to be saved without needing all images to be created beforehand and so reduce memory impact?

homm:
In general, yes. Depending on current code flow, it could be like this, or require some informations about the frames (like size and mode).

radarhere:
APNG considers the modes of all frames when saving - #6610. I don't know what an elegant way of getting that particular information from the user at the beginning would look like, and I find it hard to imagine that users would expect that they have to know that at the beginning.

Tests/test_file_apng.py
LeXofLeviafan: ...Why did you remove the filename though? I thought you were talking about _adding_ an index, not _removing_ the filename (as my own suggestion was to provide index _when the filename isn't available_)

Tests/test_file_apng.py
```diff
@@ -663,6 +665,36 @@ def test_apng_save_blend(tmp_path):
...
+    with Image.open("Tests/images/apng/single_frame.png") as im:
+        with Image.open("Tests/images/apng/delay.png") as im2:
+            im.save(
+                out, "PNG", save_all=True, append_images=[im, im2], progress=callback
+            )
+
+    assert progress == [
+        (0, 1, 7),
+        (1, 2, 7),
+        (2, 3, 7),
```
LeXofLeviafan, line 690: Say... if the frames are counted from `1`, why count the images from `0`?

LeXofLeviafan:
> allow multiframe images to be saved without needing all images to be created beforehand and so reduce memory impact

I imagine if this is ever implemented, it should not be exposed in a workaround-ish way, and it probably should be implemented as a format-specific method if it can only be supported for specific formats (e.g. `GifImagePlugin.fromFrames(generator)`).

I'm also pretty sure such a feature is out of scope for this particular pull-request.

homm:
> I'm also pretty sure such a feature is out of scope for this particular pull-request.

It out of scope of your needs, but is closely related to current PR since it may or may not share the same API and it is better to discuss it here.

homm:
> APNG considers the modes of all frames when saving

We can restrict modes of generated frames only to the same as the initial image.

homm:
Having iterators support you can track progress without any callbacks.

LeXofLeviafan:
> Having iterators support you can track progress without any callbacks.

Only in the sole specific case **when composing an animated image from separate single-frame images that you load one-by-one**. Which means, when combining animated images, you'd only get updates _per image_, not _per frame_ (and the simple usecase of "saving/converting a single animated image" would get _zero_ progress tracking support).

Not to mention you're conflating concerns of "providing input" and "progress tracking" again. Why would the user need to be forced to deal with both when he only needs one of these things? And why would the user code dealing with one of these need to also deal with the other? If the user wants to track frame progress, he passes a callback that gets invoked on every frame (e.g. `print`). If the user wants to load input images one-at-a-time, he passes a generator instead of a list. He can easily do one of these things, both, or neither. Simple, intuitive, unentangled.

Tests/test_file_apng.py
radarhere: I'm triggering these callbacks when frames are completed, not when they are started.

My intention is for `(0, 1, 7)` to communicate "We're in the first image. We've completed the first frame, there are seven frames in total".
Then `(2, 7, 7)` communicates "We're in the third image. We've completed seven frames, there are seven frames in total, so we're done"

If I started counting the frames from 0, the last value would be `(2, 6, 7)`. That looks less to me like save_all is finished.

I recognise this could be a subject of debate.

radarhere: I was trying to explain that my goal is not to number the frames from 1, but rather to clearly communicate when progress is completed, because the frame count will be compared with the total number of frames. The image count will not.

> arrays in C are indexed starting from 0, but that's an implementation detail and it's kind of ridiculous to drag it into higher-level languages

Python arrays are also indexed from 0.

LeXofLeviafan: > The image count will not.

And yet counting from `0` would be inconsistent with the other counting used _within same data set_.

> Python arrays are also indexed from 0.

Which makes no sense since Python is a high-level language (and there's some consistency issues caused by this arbitrary choice to imitate C array indexing, but that's neither here nor there).

Besides, the images aren't coming from a list as far as the user is concerned (well, there's `append_images`, but going by that logic you'd have to start from `-1` since the first input image is not in this list in the first place… except that `-1` is a valid index in Python, so it probably would be `None` instead).

Tests/test_file_apng.py
LeXofLeviafan: To clarify: can you add the filename back to the output? Indices are useful for ordering, but using actual filenames when available makes for much more user-friendly output (and it can be rather inconvenient to be forced to resolve filenames from input indices externally depending on usage scenario, while supplying appropriate filename along with the index internally would be trivial).

radarhere: I'm concerned that `(index, filename, frame_number, frame_total)` crosses the line from 'neat and tidy' to 'passing back everything to the user just in case something is useful'.

> it can be rather inconvenient to be forced to resolve filenames from input indices externally depending on usage scenario

I would have thought it was simple, given that the user just provided these images as the base image and `append_images`?

homm: > To clarify: can you add the filename back to the output?

Only `PIL.ImageFile.ImageFile` have filenames, while `append_images` are arbitrary `PIL.Image.Image`.

@radarhere why not pass just saved im object itself?

radarhere: I [suggested that](https://github.com/python-pillow/Pillow/pull/7435#discussion_r1342329935), but [it was pointed out that the image might be provided as an input more than once.](https://github.com/python-pillow/Pillow/pull/7435#discussion_r1342335519)

homm: …which is not the case with filenames, I suppose :-)

homm: I don't like passing indices since it's hard to figure out the difference between `index` and `frame_number` and it's very simple to write the code which uses `index` incorrectly.

radarhere: I'm not opposed to dropping `index` altogether - I think the key benefit of this feature is being able to track how many frames have been completed out of the total number.

radarhere: I'm still not convinced that a filename or image index is necessary, but I've pushed a commit to restore the filename.

```pycon
>>> from collections import namedtuple
>>> State = namedtuple("State", "a b")
>>> State(2, 4)[0]
2
```
Because namedtuples still have order, and I'm concerned that there could be different views about what the best order is, or that we might want to add more items in the future, I've used a dictionary instead.

Tests/test_file_apng.py
radarhere: I've pushed a commit naming each piece of information within the returned data. It is now "image_index" and "completed_frames". Hopefully that clarifies that one is the Pythonic zero-based index of the image in progress, and the other is the number of frames that have been finished.

Tests/test_file_apng.py
LeXofLeviafan: Either way, it would be a good idea to have this `getattr(im, "filename", None)` thing in one place in the codebase (and if you keep the `dict`, it would still be a good idea to have _one_ piece of code defining its creation, at least for sake of ensuring there won't be any discrepancies in key names).

radarhere: I've pushed a commit to move the creation of the dictionary into a common method.

src/PIL/Image.py
```diff
@@ -2446,6 +2446,23 @@ def save(self, fp, format=None, **params):
         if open_fp:
             fp.close()
 
+    def _save_all_progress(
```
LeXofLeviafan, line 2449: What's with "save all"? This method just passes data into another function (supplied externally).

If anything, `_report_progress` would make much more sense here. (Or even just `_progress`)

radarhere: It indicates that the method is used in relation to `im.save(out, save_all=True)`

LeXofLeviafan: …Ah. Alright, it does make sense in that context.

Though it would still make more sense to name it something like `_frame_progress`, I'd say; it's more generic and there's not really anything in the method itself that makes it necessarily about saving only.

radarhere:
@hugovk this hasn't seen any progress in a while.

Is the implementation elegant? I don't think so, but I doubt there are better alternatives.
Is the output clean? I'd like it to be simpler, but I guess simplifying it would make it less clear.

Is there anything that needs to be done here? If it's just on the backburner waiting for review, no worries. But if the fundamental concept is considered messy, I'm also fine with closing #7433

hugovk:
I'm not too bothered about the feature either way, and was probably leaving approval to @homm, as he'd reviewed it earlier.

PR #7418, opened 2023-09-23
Move code from pillow-wheels

radarhere:
Helps #7390

As suggested in that issue, moves the https://github.com/python-pillow/pillow-wheels code into this main repository, in order to simplify the release process.

I have setup the new "Wheels" job to be triggered on
- manual builds
- changes to the workflow file or the wheels directory
- tags.

Regarding TravisCI
- I re-enabled branches and pull requests at https://app.travis-ci.com/github/python-pillow/pillow/settings.
- While I've added a commit to only run it on tags, it does not seem possible to filter TravisCI builds based on file changes. There is no mention of this at https://docs.travis-ci.com/user/conditions-v1, and https://travis-ci.community/t/how-to-skip-jobs-based-on-the-files-changed-in-a-subdirectory/2979 said it wasn't possible earlier this year. I think we can work around this by triggering manual builds to test it when it is necessary.
- TravisCI will obviously not upload the wheels as GitHub Actions artifacts. It instead uploads them to GitHub releases. Maybe that's better?

Issue #7390, opened 2023-09-11
Automate build and release to PyPI


hugovk:
Currently the [release process](https://github.com/python-pillow/Pillow/blob/main/RELEASING.md) has many steps, including preparing things in this repo, and then starting a build in https://github.com/python-pillow/pillow-wheels, before downloading the sdist + wheel artifacts to disk, and manually uploading them to PyPI using twine.

I'd like us to move towards more automated releases, including using the new Trusted Publishers to upload wheels directly from the CI to PyPI, which is more secure:

* https://blog.pypi.org/posts/2023-04-20-introducing-trusted-publishers/

We can do this in a phased manner, not all of these things would need to be done at once, to ensure we can still make releases with just some of them completed.

We could potentially deal with the Travis CI wheels later, and tackle GitHub Actions first as they're a bigger win and we don't know when Travis CI will go away.

Big picture:

* Prepare release
* Push a tag
* The CI builds release artifacts
* The CI uploads artifacts to PyPI

I'd suggest something along these lines:

1. Move Linux and macOS https://github.com/python-pillow/pillow-wheels to this repo https://github.com/python-pillow/Pillow. We already have Windows wheels build here.
   * We'll need to add [`paths:` / `paths-ignore`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore) to workflows so we only trigger the main test workflows for "normal" code changes, and only trigger the wheel workflow for things like release and modifying its config
   * Upload wheels as GitHub artifacts in this repo instead

2. Add automation to download artifacts and upload them to PyPI via Trusted Publishing
   * This may need refactoring to move all the wheel building to the same workflow (but separate jobs), and then a final job that runs when the wheels have all been built, that builds sdist, checks and uploads them all.
   * See https://github.com/ultrajson/ultrajson/blob/main/.github/workflows/deploy.yml and https://github.com/ultrajson/ultrajson/actions/workflows/deploy.yml for examples.
   * We also have https://test.pypi.org/project/Pillow/ which we can use to verify the upload machinery is working smoothly before release. For example, we can upload for non-tagged wheel builds.

Thoughts?

wiredfool:
This sounds like a great step.

aclark4life:
Love it!!

radarhere:
I've created https://github.com/python-pillow/Pillow/pull/7418 for 1.

hugovk:
[...]

PR #7436, opened 2023-10-02
Don't run normal tests on changes to wheel-building files

hugovk:
Helps #7390.

After https://github.com/python-pillow/Pillow/pull/7418 is merged, we can skip running the "normal" tests when there are changes to files matching:

```yml
      - ".github/workflows/wheels*"
      - ".gitmodules"
      - ".travis.yml"
      - "wheels/**"
```

### DIFF
--- a/.github/workflows/test-cygwin.yml
+++ b/.github/workflows/test-cygwin.yml
@@ -4,11 +4,19 @@ on:
   push:
     paths-ignore:
       - ".github/workflows/docs.yml"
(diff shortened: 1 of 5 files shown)


radarhere (approved):

hugovk:
Rebased on main after https://github.com/python-pillow/Pillow/pull/7418 was merged, to confirm the wheel CI jobs don't run ✅

Issue #7341, opened 2023-08-14
Installation failed on Windows with 32-bit Python


alexsilva:
### What did you do?
Tried to update to version 10.0.0
`python -m pip install Pillow==10.0.0 --force-reinstall`

### What did you expect to happen?
That the installation proceeded without failures

### What actually happened?
The sequence of failures described below.
Version 9.5.0 installs normally.

### What are your OS, Python and Pillow versions?

* OS:  Windows 10
* Python:  3.9
* Pillow: 10.0.0

```
python -m pip install Pillow==10.0.0 --force-reinstall
Collecting Pillow==10.0.0
  Using cached Pillow-10.0.0.tar.gz (50.5 MB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Building wheels for collected packages: Pillow
  Building wheel for Pillow (pyproject.toml) ... error
  error: subprocess-exited-with-error

  × Building wheel for Pillow (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [198 lines of output]
      running bdist_wheel
      running build
      running build_py
      creating build
      creating build\lib.win32-cpython-39
      creating build\lib.win32-cpython-39\PIL
      copying src\PIL\BdfFontFile.py -> build\lib.win32-cpython-39\PIL
      copying src\PIL\BlpImagePlugin.py -> build\lib.win32-cpython-39\PIL
      copying src\PIL\BmpImagePlugin.py -> build\lib.win32-cpython-39\PIL
      copying src\PIL\BufrStubImagePlugin.py -> build\lib.win32-cpython-39\PIL
      copying src\PIL\ContainerIO.py -> build\lib.win32-cpython-39\PIL
      copying src\PIL\CurImagePlugin.py -> build\lib.win32-cpython-39\PIL
      copying src\PIL\DcxImagePlugin.py -> build\lib.win32-cpython-39\PIL
      copying src\PIL\DdsImagePlugin.py -> build\lib.win32-cpython-39\PIL
      copying src\PIL\EpsImagePlugin.py -> build\lib.win32-cpython-39\PIL
      copying src\PIL\ExifTags.py -> build\lib.win32-cpython-39\PIL
      copying src\PIL\features.py -> build\lib.win32-cpython-39\PIL
      copying src\PIL\FitsImagePlugin.py -> build\lib.win32-cpython-39\PIL
[...]

PR #7447, opened 2023-10-07
Add warning to docs: Pillow > 9.5.0 no longer includes 32-bit wheels

aclark4life:
More documentation for #7341  .

 * Add warning

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -82,7 +82,7 @@ Install Pillow with :command:`pip`::
 
 .. tab:: Windows
 
-    .. warning:: Pillow >= 9.5.0 no longer includes 32-bit wheels.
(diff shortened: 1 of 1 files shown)


radarhere:
It is not actually Pillow >= 9.5.0. As https://pillow.readthedocs.io/en/stable/releasenotes/10.0.0.html#bit-wheels states, they were dropped in Pillow 10.

aclark4life:
Yeah fixed in 2784136b7abceee3c632fb8eb062d218792cfbd5 I think ? Or feel free to word however …

radarhere:
Yep, thanks - we might have both realised the problem at the same time.

radarhere (approved):

PR #7451, opened 2023-10-09
Updated brotli to 1.1.0

radarhere:
brotli 1.1.0 has been released - https://github.com/google/brotli/releases/tag/v1.1.0

PR #7453, opened 2023-10-09
Fix libbrotli build for Windows wheels

nulano:
#7451 broke FreeType support on Windows due to changing libbrotli from a static to a dynamic (dll) build, this PR fixes that change.


radarhere:
Thanks very much.

PR #7457, opened 2023-10-11
If Image is created directly, add missing core image when loading

### LINKED ISSUE #7455: Image.Image() == Image.Image() fails with AttributeError
### What did you do?
I'm using the object constructor `Image.Image()` to create dummy `Image` objects for testing. I then compare two objects constructed in this way: `Image.Image() == Image.Image()`

I expect the comparison of two objects constructed in this way to return `True`. (I am aware that the recommendation is to use factory functions to create `Image` objects. However, would still expect the above to work.)

### What actually happened?

The comparison fails with `AttributeError`:
```pytb
  File "/[...]/lib64/python3.8/site-packages/PIL/Image.py", line 605, in __eq__
    and self.getpalette() == other.getpalette()
  File "/[...]/lib64/python3.8/site-packages/PIL/Image.py", line 1522, in getpalette
    mode = self.im.getpalettemode()
AttributeError: 'NoneType' object has no attribute 'getpalettemode'
```

* OS: RHEL8 
* Python: 3.8
* Pillow: 9.5.0, 10.0.1

```python
from PIL import Image
Image.Image() == Image.Image()
```

radarhere:
Resolves #7455

Normally, image instances are created with `Image.new()` or `Image.open()`. The issue would like to be able to use `Image()` directly.

The only thing that [`_new()`](https://github.com/python-pillow/Pillow/blob/4ecf1df4ea63457040f63adaeea35996913b6ac1/src/PIL/Image.py#L511) provides that [`__init__`](https://github.com/python-pillow/Pillow/blob/4ecf1df4ea63457040f63adaeea35996913b6ac1/src/PIL/Image.py#L483) doesn't is a core image instance. Pillow allows for `im.im` to be `None`, but expects `load()` to populate it.

Using the fact that `__init__` sets `_mode` to `""`
https://github.com/python-pillow/Pillow/blob/4ecf1df4ea63457040f63adaeea35996913b6ac1/src/PIL/Image.py#L483-L487
this PR detects in `load()` if an image has been directly instantiated, and if so, adds a core image instance with mode 1, as that is the simplest mode.

### DIFF
--- a/Tests/test_image.py
+++ b/Tests/test_image.py
@@ -94,6 +94,18 @@ class TestImage:
         # with pytest.raises(MemoryError):
         #   Image.new("L", (1000000, 1000000))
 
+    def test_direct(self):
+        # Test that a directly instantiated Image()
+        # is given a core image during load()
+        im = Image.Image()
+        assert im.im is None
+
+        im.load()
+        assert im.im is not None
+
+        # Test equality
+        assert Image.Image() == Image.Image()
+
     def test_repr_pretty(self):
         class Pretty:
             def text(self, text):
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -826,6 +826,10 @@ class Image:
         :returns: An image access object.
         :rtype: :ref:`PixelAccess` or :py:class:`PIL.PyAccess`
         """
+        if self._mode == "":
+            # This image was instantiated directly
+            self._mode = "1"
+            self.im = core.new(self.mode, self.size)
         if self.im is not None and self.palette and self.palette.dirty:
             # realize palette
             mode, arr = self.palette.getdata()


homm:
Isn't better restrict creating image instances directly? Is it used anywhere?

radarhere:
I don't understand why the user in that issue tried to run `Image()`, but this seemed like a simple solution.

In contrast, preventing users from running `Image()` but still allowing the following internal uses seemed like it would involve changing some fundamental methods.
https://github.com/python-pillow/Pillow/blob/4ecf1df4ea63457040f63adaeea35996913b6ac1/src/PIL/Image.py#L511-L512
https://github.com/python-pillow/Pillow/blob/4ecf1df4ea63457040f63adaeea35996913b6ac1/src/PIL/Image.py#L2919-L2937

If the issue is closed, I'm happy to close this as well.

radarhere:
I've created #7461 as a suggestion for how to restrict creating image instances directly. I think the changes involved are relatively minimal.

PR #7461, opened 2023-10-13
Raise a TypeError when Image() is called

### LINKED ISSUE #7455: Image.Image() == Image.Image() fails with AttributeError

radarhere:
Resolves #7455

When running
```python
from PIL import Image
Image.Image()
```
this PR will raise a TypeError
```pytb
  File "test.py", line 2, in <module>
    Image.Image()
  File "PIL/Image.py", line 488, in __init__
TypeError: Images should not be instantiated directly. Use the module new() function instead.
```


src/PIL/Image.py
```diff
@@ -980,7 +991,7 @@ def convert_transparency(m, v):
                 else:
                     # get the new transparency color.
                     # use existing conversions
-                    trns_im = Image()._new(core.new(self.mode, (1, 1)))
+                    trns_im = Image._init()._new(core.new(self.mode, (1, 1)))
```
homm, line 994: `Image._init()._new()` is quite common and very ugly. In this case we don't want to create a new image second time just to copy empty info and palette properties.

I suggest following:

`im._new(core.new())` — creates image like im, copying info and palette. I'm not sure how often this method is used this way. 
`Image._init(core.new())` — creates empty image using only core image object.

The only place I see where we need to create empty Image object (without core object) is `new()` function, but it could be easily avoided.

src/PIL/Image.py
```diff
@@ -2927,7 +2938,7 @@ def new(mode, size, color=0):
 
         color = ImageColor.getcolor(color, mode)
 
-    im = Image()
+    im = Image._init()
     if mode == "P" and isinstance(color, (list, tuple)) and len(color) in [3, 4]:
         # RGB or RGBA value for a P image
         from . import ImagePalette
```
homm, line 2944:
```python
    palette = None
    if mode == "P" and isinstance(color, (list, tuple)) and len(color) in [3, 4]:
        # RGB or RGBA value for a P image
        from . import ImagePalette

        palette = ImagePalette.ImagePalette()
        color = palette.getcolor(color)
    im = Image._init(core.fill(mode, size, color))
    im.palette = palette
    return im
```

src/PIL/Image.py
radarhere: I've cherry-picked #7460 here, since that helps.

I've pushed a commit with a version of your suggestion. It's not exact, because it turns out that `_new()` doesn't just copy existing palettes, it also creates new ones. It's not enough to just attach a palette afterwards in `new()`, because something like `Image.linear_gradient("P")` also needs a palette.

src/PIL/Image.py
radarhere: Per my last comment, `_init(im)` needs to create a palette for P mode images, which might lead to two ImagePalettes being created here.

However, we don't need `getcolor()` to know what the color index should be - it is always going to be zero, as this is the first color in the palette. Taking advantage of that, we can just re-use the palette from `_init()`.

src/PIL/Image.py
```diff
@@ -2927,14 +2939,17 @@ def new(mode, size, color=0):
...
+    rgb_color = None
     if mode == "P" and isinstance(color, (list, tuple)) and len(color) in [3, 4]:
         # RGB or RGBA value for a P image
-        from . import ImagePalette
+        rgb_color = color
 
-        im.palette = ImagePalette.ImagePalette()
-        color = im.palette.getcolor(color)
-    return im._new(core.fill(mode, size, color))
+        # This will be the first color allocated to the palette
```
homm, line 2947: Why we need to make this assumption rather than just using existing API?

We can rewrite palette if it is exist:

```python
im = Image._init(core.fill(mode, size, color))
if palette:
     im.palette = palette
```

radarhere: Hmm, I understand the desire to use APIs. My original thought was to try and avoid constructing a second imagePalette, but this version of the code seems simpler to me. I don't think it should ever change that a palette starts out empty, and that the first color is at the zero index.

Are you sure using `getcolor()` is better?

radarhere: @homm did my argument persuade you, or no? I can change to your suggestion if you like, I'd rather progress this than let it sit unresolved.

radarhere:
@hugovk is there any particular problem with this? My guess would be that it is perhaps too complex of a change for an edge case, and the user in question should just use our API as intended?

hugovk:
I was waiting to see if @homm had more to say (re: https://github.com/python-pillow/Pillow/pull/7461#discussion_r1413726539).

But it does seem like a edge case (they're getting an error calling the API in the wrong way, and this changes it to another error) and I don't think this has come up before, so I'm leaning towards the status quo.

PR #7462, opened 2023-10-13
Add Dependabot configuration for GitHub Actions updates

EwoutH:
Add a Dependabot configuration that checks once a week if the GitHub Actions are still using the latest version. If not, it opens a PR to update them.

It won't flood the repo PRs, since only major versions are specified (like v3), so only on a major release (like v4) it will update and open a PR. This way it helps actively keep GitHub Actions workflows up to date and secure, while minimizing maintenance burden.

See [Keeping your actions up to date with Dependabot](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"


hugovk:
Thanks for the PR, but we already have Renovate doing this:

* https://github.com/python-pillow/Pillow/blob/main/.github/renovate.json
* https://github.com/python-pillow/Pillow/pulls?q=is%3Apr+sort%3Aupdated-desc+author%3Aapp%2Frenovate+is%3Amerged
* https://docs.renovatebot.com/

PR #7412, opened 2023-09-21
Added ImageOps cover method

radarhere:
Resolves #7411

Adds a `cover()` method to ImageOps. This is like `fit()`, except without cropping.

Running this code for similar `ImageOps` methods in turn for comparison -

```python
from PIL import Image, ImageOps
im = Image.open("Tests/images/hopper.png")
ImageOps.cover(im, (200, 300))
```
|            | Cover                                                                                                | Fit                                                                                               | Pad                                                                                               | Contain                                                                                                 |
| ---------- | ---------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------- |
| Final Size |300 x 300                                                                                             | 200 x 300                                                                                         | 200 x 300                                                                                         | 200 x 200                                                                                               |
| Image      |cover |fit|pad| contain |

PR #7469, opened 2023-10-15
Fix typos in ImageOps.rst and tutorial.rst

nulano:
For #7412

The example gives the size `(100, 150)`:
https://github.com/python-pillow/Pillow/blob/0be67e5544185b76a33f2adf7f34f8be20e4f262/docs/reference/ImageOps.rst?plain=1#L36

It seems the sizes in the table below have been swapped accidentally.


radarhere (approved):
Thanks very much.

PR #7476, opened 2023-10-18
Added ImageFont.load_pilfont()

radarhere:
https://github.com/python-pillow/Pillow/pull/7354#issuecomment-1767239100 feels that even though the new default font provided by #7354 is superior to the old one, it breaks backwards compatibility to remove access to the old one (without the user disabling FreeType support).

This adds a `load_pilfont()` method to restore access.

### DIFF
--- a/Tests/test_imagefontpil.py
+++ b/Tests/test_imagefontpil.py
@@ -4,42 +4,48 @@ from PIL import Image, ImageDraw, ImageFont, features
 
 from .helper import assert_image_equal_tofile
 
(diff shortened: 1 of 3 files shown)


pmeier (commented):
OP of https://github.com/python-pillow/Pillow/pull/7354#issuecomment-1767239100 here: we internally agreed that we can update our images and don't need this change. If no one else complains, there is no need to make this change for us.

radarhere:
Thanks for following up.

PR #7480, opened 2023-10-19
Added messages to errors

radarhere:
In main, lint has started insisting that "Built-in Exceptions must not be thrown without being called" and "Built-in Exceptions must have a useful message" - https://github.com/python-pillow/Pillow/actions/runs/6570025729/job/17846816725#step:7:63

This PR adds messages to errors when they are raised, and replaces an error raised in the tests with a failed assertion.

### DIFF
--- a/Tests/test_file_palm.py
+++ b/Tests/test_file_palm.py
@@ -27,7 +27,7 @@ def open_with_magick(magick, tmp_path, f):
         magick + [f, outfile], stdout=subprocess.DEVNULL, stderr=subprocess.STDOUT
     )
     if rc:
-        raise OSError
+        assert False
     return Image.open(outfile)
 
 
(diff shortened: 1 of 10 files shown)


Tests/test_file_palm.py
```diff
@@ -27,7 +27,7 @@ def open_with_magick(magick, tmp_path, f):
         magick + [f, outfile], stdout=subprocess.DEVNULL, stderr=subprocess.STDOUT
     )
     if rc:
-        raise OSError
+        assert False
```
nulano, line 30: Would this work/be clearer?
```diff
-    if rc:
-        raise OSError
+    assert not rc
```

radarhere: Sure, that also works. I've made that change.

PR #7484, opened 2023-10-21
Move config from `setup.cfg` to `pyproject.toml`

hugovk:
Leave Flake8 config in `setup.cfg` because it doesn't support `pyproject.toml`: https://github.com/search?q=repo%3APyCQA%2Fflake8+pyproject.toml&type=issues

And we may be ditching Flake8 for Ruff anyway.


hugovk:
Updated to move build config as well, this means we can delete `setup.cfg`. Flake8 config goes in `.flake8`.

We also need to bump setuptools for the Ubuntu 3.8 and 3.9 jobs because the pyroma test runs without an isolated build, and old setuptools returns "UNKNOWN" for the metadata.

pyproject.toml
```diff
@@ -6,3 +6,77 @@ requires = [
...
+  "pytest",
+  "pytest-cov",
+  "pytest-timeout",
+]
+[project.urls]
+Changelog = "https://github.com/python-pillow/Pillow/blob/main/CHANGES.rst"
+Documentation = "https://pillow.readthedocs.io"
+Funding = "https://tidelift.com/subscription/pkg/pypi-pillow?utm_source=pypi-pillow&utm_medium=pypi"
+Homepage = "https://python-pillow.org"
+Mastodon= "https://fosstodon.org/@pillow"
```
radarhere, line 67:
```suggestion
Mastodon = "https://fosstodon.org/@pillow"
```

pyproject.toml
```diff
@@ -6,3 +6,77 @@ requires = [
...
+  "Programming Language :: Python :: Implementation :: PyPy",
+  "Topic :: Multimedia :: Graphics",
+  "Topic :: Multimedia :: Graphics :: Capture :: Digital Camera",
+  "Topic :: Multimedia :: Graphics :: Capture :: Screen Capture",
+  "Topic :: Multimedia :: Graphics :: Graphics Conversion",
+  "Topic :: Multimedia :: Graphics :: Viewers",
+]
+dynamic = [
+  "version",
+]
```
radarhere, line 39: Looking at https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata,
should this not be
```
dynamic = ["version", "readme"]
```
and move `readme = "README.md"` from earlier into
```
[tool.setuptools.dynamic]
version = {attr = "PIL.__version__"}
readme = {file = "README.md"}
```
later?

nulano: The `readme` is listed in both examples at that page, but it looks to me like it should be in the dynamic section so that the content type can be specified (the default is `text/x-rst`):
```
readme = {file = "README.md", content-type = "text/markdown"}
```

hugovk: Version:

Setting the version is good, it means we can move `version=PILLOW_VERSION` out of `setup.py`'s `setup()`.

README:

Despite the docs saying content-type defaults to `text/x-rst`, I did a test upload to https://test.pypi.org/project/pillow/10.2.0.dev0/ (files built using `python -m build` from the current state of the PR) and the README renders just fine, so I think it's fine?

PS @radarhere and other Pillow maintainers: if you like, I can add you to https://test.pypi.org/project/pillow/ in case you want to test anything out there. Just let me know your username (you need to create a new account there).

setup.cfg
```diff
@@ -1,74 +0,0 @@
...
-long_description = file: README.md
-long_description_content_type = text/markdown
-url = https://python-pillow.org
-author = Jeffrey A. Clark (Alex)
-author_email = aclark@aclark.net
-license = HPND
-classifiers =
-    Development Status :: 6 - Mature
-    License :: OSI Approved :: Historical Permission Notice and Disclaimer (HPND)
-    Programming Language :: Python :: 3
```
radarhere, line 13: Why is "Programming Language :: Python :: 3" not present in pyproject.toml?

hugovk: pyproject-fmt removes it:
https://github.com/tox-dev/pyproject-fmt/issues/106

radarhere (approved):

setup.py
```diff
@@ -986,7 +986,6 @@ def debug_build():
 
 try:
     setup(
-        version=PILLOW_VERSION,
```
homm, line 989: Is there a way to query the version from setuptools instead of reading it from a file in `get_version`?

hugovk: There are 7 options at https://packaging.python.org/en/latest/guides/single-sourcing-package-version/

1. Also reads from file
2. I'm not too sure about those external build tools
3. We're using something like this
4. Also reads from file
5. Moves version constant to `setup.py`, project code fetchs via `importlib.metadata`
6. Import the version from `src/PIL/__init__.py`, but I'm not sure if we can import PIL before the build?
7. Get the version number from SCM using setuptools_scm. I've used setuptools_scm in other projects, this might work.

But I suggest these could be topics for a followup.

aclark4life:
Out of curiosity is it possible to drop `setup.py` now too? I'm doing it over here: https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4011. But I'm not exactly sure how it works. Maybe in a project without C extensions you don't need setup.py anymore.

hugovk:
> Out of curiosity is it possible to drop `setup.py` now too? I'm doing it over here: [AcademySoftwareFoundation/OpenImageIO#4011](https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4011). But I'm not exactly sure how it works. Maybe in a project without C extensions you don't need setup.py anymore.

* https://setuptools.pypa.io/en/latest/userguide/ext_modules.html

Pillow still needs it for building C extensions with setuptools, but pure Python projects can drop both `setup.py` and `setup.cfg` for a single `pyproject.toml`.

aclark4life:
Well, the point was more that we could replace setuptools with scikit-build if we were so inclined, and I'm trying to gauge how inclined we are to do that. I'm also not following setuptools very closely anymore and wondering if there are any indications that we should prepare to exist without setuptools.

radarhere:
Investigating scikit-build, I found that it [depends on setuptools](https://github.com/scikit-build/scikit-build/blob/630c3e79f3eed65f365db169a6004dd18fe460e2/pyproject.toml#L33) and "[is fundamentally just glue between the setuptools Python module and CMake](https://github.com/scikit-build/scikit-build)".

That was confusing, but then I found https://github.com/scikit-build/scikit-build-core. I'm guessing that was what was meant?

aclark4life:
Correct, the newer one, used in the OIIO PR I referenced above.

hugovk:
I'm not necessarily against moving to something newer than setuptools if someone wants to experiment, but our `setup.py` is 1,013 lines long, so it might be not be an easy task! There's also PDM (I think this just calls setuptools) and meson-python.

I think setuptools is going to be around for a long time. It's still maintained, and they're working to modernise it, which sometimes brings the occasional rocky patch, but we've figured it out.

aclark4life:
PDM is what I used to convert OIIO's setup.py to pyproject.toml. I guess the best case scenario would be for setuptools to support C extensions without setup.py, but no idea of that's on the roadmap. I'll do some experiments!

PR #7488, opened 2023-10-23
Allow configuring JPEG restart marker interval on save

bgilbert:
If a JPEG includes restart markers, the decoder doesn't have to give up if it finds data corruption, but can scan forward to the next restart marker and resume decoding from there.  Restart markers add overhead, but are useful in applications that want to support a limited form of error recovery or random access within a JPEG.

libjpeg allows specifying the marker interval either in MCU blocks or in MCU rows.  Support both, via separate parameters, rather than requiring callers to do the math.

### DIFF
--- a/Tests/test_file_jpeg.py
+++ b/Tests/test_file_jpeg.py
@@ -643,6 +643,23 @@ class TestFileJpeg:
             assert max(im2.quantization[0]) <= 255
             assert max(im2.quantization[1]) <= 255
 
+    @pytest.mark.parametrize(
+        "blocks, rows, markers",
+        ((0, 0, 0), (1, 0, 15), (3, 0, 5), (8, 0, 1), (0, 1, 3), (0, 2, 1)),
+    )
+    def test_restart_markers(self, blocks, rows, markers):
+        im = Image.new("RGB", (32, 32))  # 16 MCUs
+        out = BytesIO()
+        im.save(
+            out,
+            format="JPEG",
+            restart_marker_blocks=blocks,
+            restart_marker_rows=rows,
+            # force 8x8 pixel MCUs
+            subsampling=0,
+        )
+        assert len(re.findall(b"\xff[\xd0-\xd7]", out.getvalue())) == markers
+
     @pytest.mark.skipif(not djpeg_available(), reason="djpeg not available")
     def test_load_djpeg(self):
         with Image.open(TEST_FILE) as img:
--- a/src/libImaging/Jpeg.h
+++ b/src/libImaging/Jpeg.h
@@ -83,6 +83,10 @@ typedef struct {
     /* Chroma Subsampling (-1=default, 0=none, 1=medium, 2=high) */
     int subsampling;
 
+    /* Restart marker interval, in MCU blocks or MCU rows, or 0 for none */
+    unsigned int restart_marker_blocks;
+    unsigned int restart_marker_rows;
+
     /* Converter input mode (input to the shuffler) */
     char rawmode[8 + 1];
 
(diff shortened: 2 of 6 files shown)


src/libImaging/Jpeg.h
```diff
@@ -83,6 +83,10 @@ typedef struct {
     /* Chroma Subsampling (-1=default, 0=none, 1=medium, 2=high) */
     int subsampling;
 
+    /* Restart marker interval, in MCU blocks or MCU rows, or 0 for none */
+    unsigned int restart_marker_blocks;
+    int restart_marker_rows;
```
homm, line 88: Why different types?

bgilbert: Those are the types libjpeg uses.  Looking through the libjpeg source code, there doesn't seem to be a good reason for it, and I agree it seems pretty nonsensical.  I can switch over to unsigned if that's preferred.

radarhere:
Just linking to some more information, https://github.com/libjpeg-turbo/libjpeg-turbo/blob/main/libjpeg.txt describes these features as
> unsigned int restart_interval
> int restart_in_rows
>    To emit restart markers in the JPEG file, set one of these nonzero. Set restart_interval to specify the exact interval in MCU blocks (samples in lossless mode).  Set restart_in_rows to specify the interval in MCU rows.  (If restart_in_rows is not 0, then restart_interval is set after the image width in MCUs is computed.) Defaults are zero (no restarts).  One restart marker per MCU row is often a good choice.  NOTE: the overhead of restart markers is higher in grayscale JPEG files than in color files, and MUCH higher in progressive JPEGs.  If you use restarts, you may want to use larger intervals in those cases.

src/libImaging/Jpeg.h
bgilbert: Done.

bgilbert:
Updated with feedback from https://github.com/bgilbert/Pillow/pull/2.

Tests/test_file_jpeg.py
```diff
@@ -643,6 +643,26 @@ def test_save_low_quality_baseline_qtables(self):
...
+            format="JPEG",
+            restart_marker_blocks=blocks,
+            restart_marker_rows=rows,
+            # force 8x8 pixel MCUs
+            subsampling=0,
+        )
+        markers_found = sum(
+            len(out.getvalue().split(bytes([0xFF, 0xD0 + i]))) - 1 for i in range(8)
+        )
+        assert markers_found == markers
```
radarhere, line 664:
```suggestion
        assert len(re.findall(b"\xff[\xd0-\xd7]", out.getvalue())) == markers
```
This seems simpler.

radarhere (approved):

radarhere:
Could you provide a short explanation for why you picked the names "restart_marker_blocks" and "restart_marker_rows" instead of continuing with libjpeg's names of "restart_interval" and "restart_in_rows"?

bgilbert:
I find the libjpeg names inconsistent and confusing, and tried to pick better ones.  I'm open to other ideas though.

bgilbert:
Sorry, I missed the merge commit when force-pushing.  The only changes to the PR are the `test_file_jpeg.py` ones in the [net diff](https://github.com/python-pillow/Pillow/compare/70d9cdf79e422118510e5c0344b570d1ae11f5f5..e5727224794620720f9b73098cce8c80b16ac304).

radarhere (approved):

bgilbert:
Is this ready to go in?

radarhere:
I've approved it, so I raise no objections if it were to be merged in.

I'm holding off on actually merging for two reasons
1. I'm not completely certain about renaming "restart_interval" and "restart_in_rows", as that is subjective, and maybe others have thoughts
2. In case @homm has any further thoughts in general, as he has offered some thoughts already

homm (approved):

PR #7491, opened 2023-10-24
Implement `streamtype=1` option for tables-only JPEG encoding

bgilbert:
We already support `streamtype=2` to skip producing JPEG tables, but `streamtype=1`, which skips everything but the tables, was never implemented. The `streamtype=1` stub code dates to Git pre-history, so it's not immediately clear why.  Implement the missing support.

`jpeg_write_tables()` can't resume after a full output buffer (it fails with `JERR_CANT_SUSPEND`), so it might seem that Pillow needs to pre-compute the necessary buffer size.  However, in the normal case of producing an interchange stream, the tables are written via the same libjpeg codepath during the first `jpeg_write_scanlines()` call, and table writes aren't resumable there either.  Thus, any buffer large enough for the normal case will also be large enough for a tables-only file.

I've opted to implement the early exit with a `goto` rather than refactoring the state machine.  If the `goto` is not desired, it would be possible to add a dedicated cleanup state, do an early return after writing tables, and require a second call into the state machine to finish cleanup.

The `streamtype` option isn't documented and this PR doesn't change that. It does add a test though.

### DIFF
--- a/Tests/test_file_jpeg.py
+++ b/Tests/test_file_jpeg.py
@@ -961,6 +961,28 @@ class TestFileJpeg:
             im.load()
             ImageFile.LOAD_TRUNCATED_IMAGES = False
 
+    def test_separate_tables(self):
+        im = hopper()
+        data = []  # [interchange, tables-only, image-only]
+        for streamtype in range(3):
+            out = BytesIO()
+            im.save(out, format="JPEG", streamtype=streamtype)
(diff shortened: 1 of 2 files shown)


bgilbert:
Updated based on suggestions in https://github.com/bgilbert/Pillow/pull/1.

radarhere (approved):

PR #7496, opened 2023-10-26
Concatenate multiple JPEG EXIF markers

### LINKED ISSUE #7495: Multi-segment Exif fails to read
### What did you do?
Writing code to read ai generated images and tried to get User Comment field from EXIF data contained in an image opened with Image.open. The jpeg image is as far as I can tell not corrupt. ExifTool reads it OK and shows me the comment with a minor warning that "File contains multi-segment EXIF." I can see the User Comment listed in the console with ExifTool. It's large, but looks OK.

User Comment field to be read and returned.

### What actually happened?
Using image.getexif().get_ifd(ExifTags.IFD.Exif) returns an empty dict along with this warning on the console-
python3.11/site-packages/PIL/TiffImagePlugin.py:868: UserWarning: Truncated File Read

* OS: Linux pop-os 6.2.6-76060206-generic (Ubuntu compatible with 22 I think)
* Python: 3.11
* Pillow: 10.1.0

```python
with Image.open("03d2695a.jpeg") as i:
    x = i.getexif().get_ifd(ExifTags.IFD.Exif)
```
[03d2695a.jpeg.zip](https://github.com/python-pillow/Pillow/files/13166523/03d2695a.jpeg.zip)

radarhere:
Resolves #7495

The issue has found an image where multiple EXIF markers are present. This is against the specification -

https://en.wikipedia.org/wiki/Exif
> Exif metadata are restricted in size to 64 kB in JPEG images because according to the specification this information must be contained within a single JPEG APP1 segment.

but this PR allows Pillow to be more flexible, and concatenates the two markers.

I created a test image with
```python
from PIL import Image
im = Image.new("L", (1, 1))
im.save("multiple_exif.jpg", exif=b"Exif\x00\x00first", extra=b"\xFF\xE1\x00\x0EExif\x00\x00second")
```


src/PIL/JpegImagePlugin.py
```diff
@@ -85,10 +85,12 @@ def APP(self, marker):
                 self.info["dpi"] = jfif_density
             self.info["jfif_unit"] = jfif_unit
             self.info["jfif_density"] = jfif_density
-    elif marker == 0xFFE1 and s[:5] == b"Exif\0":
-        if "exif" not in self.info:
-            # extract EXIF information (incomplete)
-            self.info["exif"] = s  # FIXME: value will change
+    elif marker == 0xFFE1 and s[:6] == b"Exif\0\0":
```
homm, line 88: Why the length changed to 6?

radarhere: Since I am removing 6 bytes of the EXIF prefix when concatenating subsequent markers, it makes sense to me to check that the 6-byte prefix is present.

PR #7497, opened 2023-10-27
Fix incorrect color blending for overlapping glyphs

ZachNagengast:
Fixes the previous code that just took the color with the highest alpha value. 

 * With this change, colors are now properly blended based on their respective alpha values. 

This occurs only for overlapping glyphs in emoji zwj sequences as 🫱🏼‍🫲🏾 and 👩🏼‍❤️‍👨🏽

Example **before**:
sequence-wrong-

sequence-wrong

Notice the artifacts on the left side of the heart that overlaps, and the background color coming through between the shaking hands.

Here is the **corrected** output:

sequence-corrected
sequence-corrected-red

The heart now retains its alpha entirely, and the hands fit together much closer.

Here is some code you can use to test it out:

```python

font = ImageFont.truetype(
    "/System/Library/Fonts/Apple Color Emoji.ttc", size=160
)

im = Image.new("RGBA", (160*3, 160), "blue")
d = ImageDraw.Draw(im)

d.text((0, 0), "👩🏼‍❤️‍👨🏽🫱🏼‍🫲🏾🏴‍☠️", fill="white", embedded_color=True, font=font)

im.show()
```

The issue requires a glyph that is originally separate and put together on the fly. This probably saves space because of all the possible combinations. 

"u1F468.3.u1F48B.L"
glyph-u1F468 3 u1F48B L
"u1F468.3.u1F48B.R"
glyph-u1F468 3 u1F48B R

and 

"u1FAF1.2.L"
glyph-u1FAF1 2 L
"u1FAF2.2.R"
glyph-u1FAF2 2 R


homm:
You shouldn’t use float in pixels computations. You can learn from the other code or wait for help with this.

ZachNagengast:
Fair enough, just sharing the issue and what worked for me. I'll take a second look but open to edits or suggestions.

homm:
If I understand this correctly (currently from the phone), this is alpha blending. We already have implemented alpha blending as a separate function. It would be great to reuse it, which cold be not so easy. Or at least you can copy parts of code

radarhere:
Interestingly, running your code on my machine, I don't get the disfigured heart with main.

out

You can see that our test suite fails with these changes. That is because
https://github.com/python-pillow/Pillow/blob/d3fd1734afb45d379e26b22ea733e141c336b875/Tests/test_imagefont.py#L966-L975
no longer gives https://github.com/python-pillow/Pillow/blob/main/Tests/images/colr_bungee.png

but instead

with a darker outline. Did you have any thoughts on this difference?

src/_imagingft.c
```diff
@@ -1059,16 +1059,25 @@ font_render(FontObject *self, PyObject *args) {
...
+                        float out_alpha = src_alpha + dst_alpha * (1.0 - src_alpha);
+
+                        /* unpremultiply BGRa to RGBA */
+                        int src_red = source[k * 4 + 0];
+                        int src_grn = source[k * 4 + 1];
+                        int src_blu = source[k * 4 + 2];
+
+                        int dst_blu = target[k * 4 + 0];
+                        int dst_grn = target[k * 4 + 1];
+                        int dst_red = target[k * 4 + 2];
```
nulano, line 1073: This is not unpremultiplying, which is likely what causes the dark outline pointed out by @radarhere.

nulano (changes requested):
The BGRA and grayscale bitmaps should be rendered the same way. If this is to be merged, the gray glyphs should be alpha blended as well.

ZachNagengast:
@radarhere I believe the blond haired version doesn't show as much of the alpha issues because the majority of the hair portion has higher alpha, I can see it's still there by the jagged edge at the border, and also the heart is supposed to overlap the left side when rendered. You might see it with this version of the emoji 👩🏻‍❤️‍👨🏼. I can also see a lot of blue coming through the hands so I believe it's there in your picture. I'm not sure about the dark outline but @nulano's explanation makes sense, will adjust that.

ZachNagengast:
Ok I was able to resolve the dark outline by bringing back in the unpremultiply step, and also updated the blending logic to use one of the macros that is already in the codebase.
image

image
 I also started down the path of adding a test, but the system apple color emoji font is 189mb which seems a bit large to add. I may be able to pare it down by creating a much smaller set that only includes the emojis needed to test, or investigate some other font that has similar overlapping glyphs.

ZachNagengast:
@nulano Regarding the grey scale alpha blending, I've implemented a method do it in a similar way as the RGBA, but it's resulting in an image that is ~25% bigger although it looks visually the same. This is because the new code results in an image that has an alpha channel, whereas the test image doesn't.

```diff
- if (color) {
-     unsigned char *ink = (unsigned char *)&foreground_ink;
-     for (k = x0; k < x1; k++) {
-         v = source[k] * convert_scale;
-         if (target[k * 4 + 3] < v) {
-             target[k * 4 + 0] = ink[0];
-             target[k * 4 + 1] = ink[1];
-             target[k * 4 + 2] = ink[2];
-             target[k * 4 + 3] = v;
-         }
-     }
- } else {
-     for (k = x0; k < x1; k++) {
-         v = source[k] * convert_scale;
-         if (target[k] < v) {
-             target[k] = v;
-         }
-     }
- }
+ if (color) {
+     unsigned char *ink = (unsigned char *)&foreground_ink;
+     for (k = x0; k < x1; k++) {
+         int src_alpha = source[k] * convert_scale;
+         if (src_alpha > 0) {
+             if (target[k * 4 + 3] > 0) {
+                 target[k * 4 + 0] = BLEND(src_alpha, target[k * 4 + 0], ink[0], tmp);
+                 target[k * 4 + 1] = BLEND(src_alpha, target[k * 4 + 1], ink[1], tmp);
+                 target[k * 4 + 2] = BLEND(src_alpha, target[k * 4 + 2], ink[2], tmp);
+                 target[k * 4 + 3] = CLIP8(src_alpha + MULDIV255(target[k * 4 + 3], (255 - src_alpha), tmp));
+             } else {
+                 target[k * 4 + 0] = ink[0];
+                 target[k * 4 + 1] = ink[1];
+                 target[k * 4 + 2] = ink[2];
+                 target[k * 4 + 3] = src_alpha;
+             }
+         }
+     }
+ } else {
+     for (k = x0; k < x1; k++) {
+         int src_alpha = source[k] * convert_scale;
+         if (src_alpha > 0) {
+             target[k] = target[k] > 0 ? CLIP8(src_alpha + MULDIV255(target[k], (255 - src_alpha), tmp)) : src_alpha;
+         }
+     }
+ }
```

outputs look the same
tmpsbepe46j
bitmap_font_stroke_raqm

I'll look further into where the alpha channel is coming from but lmk if you see anything obvious here.

ZachNagengast:
> I also started down the path of adding a test, but the system apple color emoji font is 189mb which seems a bit large to add. I may be able to pare it down by creating a much smaller set that only includes the emojis needed to test, or investigate some other font that has similar overlapping glyphs.

@radarhere do you have any opinion on this? Possible options would be 1. Include the font with git lfs 2. Create a new font specifically for testing 3. Some other method to test the code (i.e. with a different font, etc)

src/_imagingft.c
```diff
@@ -1059,15 +1060,32 @@ font_render(FontObject *self, PyObject *args) {
...
+                                /* blend colors to BGRa */
+                                target[k * 4 + 0] = BLEND(src_alpha, target[k * 4 + 0], src_blu, tmp);
+                                target[k * 4 + 1] = BLEND(src_alpha, target[k * 4 + 1], src_grn, tmp);
+                                target[k * 4 + 2] = BLEND(src_alpha, target[k * 4 + 2], src_red, tmp);
+
+                                /* blend alpha */
+                                int out_alpha = CLIP8(src_alpha + MULDIV255(target[k * 4 + 3], (255 - src_alpha), tmp));
+                                target[k * 4 + 3] = out_alpha;
+                            } else {
+                                /* paste source directly to BGRa */
```
nulano, line 1083:
```suggestion
                                /* paste unpremultiplied RGBA values */
```

The source is `BGRa` data (lowercase `a` indicates premultiplied values), the target is `RGBA`.

nulano (commented):
> the heart is supposed to overlap the left side when rendered

It's not clear to me how the rendering code is supposed to know which glyph should be on top when there are overlapping glyphs. AFAIK they are usually just ordered left-to-right before rendering (to account for mixed LTR/RTL text), but perhaps it is possible that intentionally overlapping glyphs are reordered by the font. This is why my original implementation just used the higher value instead of doing proper blending. However, I can see how that is undesirable in this case. I've also run your test with the Windows font Segoe UI Emoji and it has a similar issue for the hands emoji.

---

> I also started down the path of adding a test, but the system apple color emoji font is 189mb which seems a bit large to add. [...] Possible options would be

> 1. Include the font [Apple Color Emoji] with git lfs

No, don't include that font. It's way too large and I doubt its license allows for that.

There is also https://github.com/python-pillow/test-images which is used for extra test files; I'm not sure if it could go there instead.

(Although Segoe UI Emoji is only 11MB, the Windows Fonts FAQ explicitly states that it may not be redistributed and is not even available for sale.)

> 2. Create a new font specifically for testing

Removing unused glyphs from a test font to minimize its size would also be fine; I'm not sure which is easier.

> 3. Some other method to test the code (i.e. with a different font, etc)

A few ideas:
* Look for another font with this issue. I was not able to find a suitable font, but didn't look for too long.
  
  An old version (see the discussion starting at https://github.com/python-pillow/Pillow/pull/4955#issuecomment-704841075) of Noto Color Emoji is already included in the test files. It does not contain the glyphs from your test, but a more recent version might. However, the latest version seems to be in SVG format which is not yet supported.

  Even better might be a COLR/CPAL font as they tend to be smaller than CBDT and SBIX fonts.
* Raqm has somewhat recently added support for letter and word spacing: https://github.com/HOST-Oman/libraqm/pull/171
  I'm not sure how useful it is in practice and I have not yet seen any requests for adding support for letter spacing in Pillow, but it could make it easier to test this (by adding negative spacing to force glyphs to overlap).

---

> Regarding the grey scale alpha blending, I've implemented a method do it in a similar way as the RGBA, but it's resulting in an image that is ~25% bigger although it looks visually the same. This is because the new code results in an image that has an alpha channel, whereas the test image doesn't.

I don't see how that diff could possibly affect the output image format. What error do you get when you run the tests? There are separate checks for the image mode and similarity.

The diff does look correct to me, although I haven't tested it thoroughly.

ZachNagengast:
Thanks for the detailed response! Here's the tests that are failing with the greyscale changes:
```
FAILED Tests/test_imagefont.py::test_bitmap_font_stroke[Layout.BASIC] - AssertionError:  average pixel value difference 0.3207 > epsilon 0.0300
FAILED Tests/test_imagefont.py::test_bitmap_font_stroke[Layout.RAQM] - AssertionError:  average pixel value difference 0.4011 > epsilon 0.0300
FAILED Tests/test_imagefont.py::test_default_font - AssertionError: got different content
```

The info for each image in the test `test_bitmap_font_stroke`
My changes:
image
Test image:
image

Similar slight difference for `test_default_font`, except that one is similar enough to pass the similarity check, then when checking byte values it fails. The metadata seems to suggest it has an alpha channel somehow, so I'll try to get to the bottom of that.

Next steps here, I will take a look into so other fonts already included to see if there is a case that can be used for these tests. If that doesn't work, I'll attempt your suggestion by forcing the overlap with some spacing adjustments.

src/_imagingft.c
```diff
@@ -1059,15 +1060,32 @@ font_render(FontObject *self, PyObject *args) {
...
+                            int src_red = CLIP8((255 * (int)source[k * 4 + 0]) / src_alpha);
+                            int src_grn = CLIP8((255 * (int)source[k * 4 + 1]) / src_alpha);
+                            int src_blu = CLIP8((255 * (int)source[k * 4 + 2]) / src_alpha);
+
+                            /* blend required if target has data */
+                            if (target[k * 4 + 3] > 0) {
+                                /* blend colors to BGRa */
+                                target[k * 4 + 0] = BLEND(src_alpha, target[k * 4 + 0], src_blu, tmp);
+                                target[k * 4 + 1] = BLEND(src_alpha, target[k * 4 + 1], src_grn, tmp);
+                                target[k * 4 + 2] = BLEND(src_alpha, target[k * 4 + 2], src_red, tmp);
```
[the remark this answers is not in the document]
ZachNagengast: I tried PREBLEND with CLIP8
image

and without CLIP8
image

It seems like they're getting some artifacts in the areas with alpha, what do you think?

Here is just `BLEND` for reference:
image

nulano:
> AssertionError:  average pixel value difference 0.3207 > epsilon 0.0300

Yeah, that is not a mode issue, but a change in output due to different blending of the edges.
You should be able to just replace the images by adding an `im.save(target)` call before the check and visually comparing they are correct.

> Similar slight difference for test_default_font, except that one is similar enough to pass the similarity check, then when checking byte values it fails. The metadata seems to suggest it has an alpha channel somehow, so I'll try to get to the bottom of that.

I only see the byte check, no similarity check. Since it failed on the byte check, the mode must be correct as it is tested first. Again, it should be fine to replace the test file.

> If that doesn't work, I'll attempt your suggestion by forcing the overlap with some spacing adjustments.

Note that this isn't yet implemented in Pillow, so it's not trivial to add to a test. I somehow lost that part while editing my previous comment.

ZachNagengast:
Ok great! Just replaced the images and appears to be passing again 👍

ZachNagengast:
Took a while but looks like everything is finally passing again except for the codecov check.

hugovk:
Is it possible to add tests to cover the missed `if` and two `else` blocks, so we know they're working correctly and to guard against regressions?

nulano:
I've opened https://github.com/ZachNagengast/Pillow/pull/1 to add tests for this PR.

I created two fonts specifically to test this PR at https://github.com/nulano/font-tests:
* a CBDT color bitmap font to test blending color glyphs when rendering in color mode,
* and a grayscale bitmap font to test blending glyphs when rendering in either mode.

Although these should provide full code coverage, I was unable to fully test one case - blending a grayscale glyph on top of a color glyph when rendering in color mode.

The CBDT font can replace NotoColorEmoji as the CBDT test font to save 5MB in the sdist.

Here are the added test images:

|test|before PR|after PR|
|-|-|-|
|`test_bitmap_blend`|||
|`test_cbdt`|||
|`test_cbdt_mask`|||

ZachNagengast:
Fantastic solution @nulano. Custom fonts seem like the way to go for these specific tests! Will monitor PR for when it's ready to merge.

nulano:
I've updated https://github.com/ZachNagengast/Pillow/pull/1 to fix a big-endian bug from #4955 I found with the new test `test_bitmap_blend`.

src/_imagingft.c
```diff
@@ -1059,34 +1060,58 @@ font_render(FontObject *self, PyObject *args) {
...
+
+                                /* blend alpha */
+                                int out_alpha = CLIP8(src_alpha + MULDIV255(target[k * 4 + 3], (255 - src_alpha), tmp));
+                                target[k * 4 + 3] = out_alpha;
```
nulano, line 1081:
```suggestion
                                /* blend RGBA colors */
                                target[k * 4 + 0] = BLEND(src_alpha, target[k * 4 + 0], src_red, tmp);
                                target[k * 4 + 1] = BLEND(src_alpha, target[k * 4 + 1], src_grn, tmp);
                                target[k * 4 + 2] = BLEND(src_alpha, target[k * 4 + 2], src_blu, tmp);
                                target[k * 4 + 3] = CLIP8(src_alpha + MULDIV255(target[k * 4 + 3], (255 - src_alpha), tmp));
```

src/_imagingft.c
```diff
@@ -1049,6 +1049,7 @@ font_render(FontObject *self, PyObject *args) {
                 int k;
                 unsigned char v;
```
nulano, line 1050:
```suggestion
```
To fix https://github.com/nulano/Pillow/actions/runs/7007718323/job/19063218352#step:7:237:
```
  src/_imagingft.c:1050:31: warning: unused variable ‘v’ [-Wunused-variable]
   1050 |                 unsigned char v;
        |                               ^
```

src/_imagingft.c
```diff
@@ -1059,34 +1060,58 @@ font_render(FontObject *self, PyObject *args) {
...
-                        if (target[k * 4 + 3] < source[k * 4 + 3]) {
-                            /* unpremultiply BGRa to RGBA */
-                            target[k * 4 + 0] = CLIP8(
-                                (255 * (int)source[k * 4 + 2]) / source[k * 4 + 3]);
-                            target[k * 4 + 1] = CLIP8(
-                                (255 * (int)source[k * 4 + 1]) / source[k * 4 + 3]);
-                            target[k * 4 + 2] = CLIP8(
-                                (255 * (int)source[k * 4 + 0]) / source[k * 4 + 3]);
-                            target[k * 4 + 3] = source[k * 4 + 3];
+                        int src_alpha = source[k * 4 + 3];
```
nulano, line 1063:
```suggestion
                        unsigned int src_alpha = source[k * 4 + 3];
```

src/_imagingft.c
```diff
@@ -1059,34 +1060,58 @@ font_render(FontObject *self, PyObject *args) {
...
                     if (color) {
                         unsigned char *ink = (unsigned char *)&foreground_ink;
                         for (k = x0; k < x1; k++) {
-                            v = source[k] * convert_scale;
-                            if (target[k * 4 + 3] < v) {
-                                target[k * 4 + 0] = ink[0];
-                                target[k * 4 + 1] = ink[1];
-                                target[k * 4 + 2] = ink[2];
-                                target[k * 4 + 3] = v;
+                            int src_alpha = source[k] * convert_scale;
```
nulano, line 1095:
```suggestion
                            unsigned int src_alpha = source[k] * convert_scale;
```

src/_imagingft.c
```diff
@@ -1059,34 +1060,58 @@ font_render(FontObject *self, PyObject *args) {
...
+                                    target[k * 4 + 3] = src_alpha;
+                                }
                             }
                         }
                     } else {
                         for (k = x0; k < x1; k++) {
-                            v = source[k] * convert_scale;
-                            if (target[k] < v) {
-                                target[k] = v;
+                            int src_alpha = source[k] * convert_scale;
```
nulano, line 1112:
```suggestion
                            unsigned int src_alpha = source[k] * convert_scale;
```
To fix https://github.com/nulano/Pillow/actions/runs/7007718323/job/19063218352#step:7:234:
```
  src/_imagingft.c:1114:127: warning: operand of ‘?:’ changes signedness from ‘int’ to ‘unsigned int’ due to unsignedness of other operand [-Wsign-compare]
   1114 |                                 target[k] = target[k] > 0 ? CLIP8(src_alpha + MULDIV255(target[k], (255 - src_alpha), tmp)) : src_alpha;
        |                                                                                                                               ^~~~~~~~~
```

nulano (approved):

src/_imagingft.c
```diff
@@ -1059,34 +1059,55 @@ font_render(FontObject *self, PyObject *args) {
...
+                                /* blend RGBA colors */
+                                target[k * 4 + 0] = BLEND(src_alpha, target[k * 4 + 0], src_red, tmp);
+                                target[k * 4 + 1] = BLEND(src_alpha, target[k * 4 + 1], src_grn, tmp);
+                                target[k * 4 + 2] = BLEND(src_alpha, target[k * 4 + 2], src_blu, tmp);
+                                target[k * 4 + 3] = CLIP8(src_alpha + MULDIV255(target[k * 4 + 3], (255 - src_alpha), tmp));
+                            } else {
+                                /* paste unpremultiplied RGBA values */
+                                target[k * 4 + 0] = src_red;
+                                target[k * 4 + 1] = src_grn;
+                                target[k * 4 + 2] = src_blu;
```
radarhere, line 1082:
```suggestion
                            int src_green = CLIP8((255 * (int)source[k * 4 + 1]) / src_alpha);
                            int src_blue = CLIP8((255 * (int)source[k * 4 + 0]) / src_alpha);

                            /* blend required if target has data */
                            if (target[k * 4 + 3] > 0) {
                                /* blend RGBA colors */
                                target[k * 4 + 0] = BLEND(src_alpha, target[k * 4 + 0], src_red, tmp);
                                target[k * 4 + 1] = BLEND(src_alpha, target[k * 4 + 1], src_green, tmp);
                                target[k * 4 + 2] = BLEND(src_alpha, target[k * 4 + 2], src_blue, tmp);
                                target[k * 4 + 3] = CLIP8(src_alpha + MULDIV255(target[k * 4 + 3], (255 - src_alpha), tmp));
                            } else {
                                /* paste unpremultiplied RGBA values */
                                target[k * 4 + 0] = src_red;
                                target[k * 4 + 1] = src_green;
                                target[k * 4 + 2] = src_blue;
```

Tests/test_imagefont.py
```diff
@@ -856,6 +856,19 @@ def test_bitmap_font_stroke(layout_engine):
...
 
+@pytest.mark.parametrize("embedded_color", (False, True))
+def test_bitmap_blend(layout_engine, embedded_color):
+    font = ImageFont.truetype(
+        "Tests/fonts/EBDTTestFont.ttf", size=64, layout_engine=layout_engine
+    )
+
+    im = Image.new("RGBA", (128, 96), "white")
+    d = ImageDraw.Draw(im)
+    d.text((16, 16), "AA", font=font, embedded_color=embedded_color, fill="#8E2F52")
```
radarhere, line 867:
```suggestion
    d.text((16, 16), "AA", font=font, fill="#8E2F52", embedded_color=embedded_color)
```
Updates the argument order to match
https://github.com/python-pillow/Pillow/blob/316f39702c16ed97a95b971cb4aad8a87a26f9df/Tests/test_imagefont.py#L866

src/PIL/ImageDraw.py
```diff
@@ -542,7 +543,8 @@ def draw_text(ink, stroke_width=0, stroke_offset=None):
                 # font.getmask2(mode="RGBA") returns color in RGB bands and mask in A
                 # extract mask and set text alpha
                 color, mask = mask, mask.getband(3)
-                color.fillband(3, (ink >> 24) & 0xFF)
+                ink_alpha = struct.pack("=i", ink)[3]
```
radarhere, line 546:
```suggestion
                ink_alpha = struct.pack("i", ink)[3]
```
Searching through Pillow's existing code, the tradition has been to use native mode implicitly rather than explicitly.
https://github.com/python-pillow/Pillow/blob/76446ee45066554380e08b2f0c4c514f43a297c1/src/PIL/SpiderImagePlugin.py#L259
https://github.com/python-pillow/Pillow/blob/76446ee45066554380e08b2f0c4c514f43a297c1/src/PIL/SgiImagePlugin.py#L184-L188

nulano: The default is `@` mode, which is not equivalent to `=`.
From https://docs.python.org/3/library/struct.html#byte-order-size-and-alignment:

> Note the difference between '@' and '=': both use native byte order, but the size and alignment of the latter is standardized.

The `s` and `f` formats are the same for both `@` and `=`.
`l` is platform dependent with `@`, but is used with `>` in the quoted code above.

`i` is platform dependent, so changing `=` to `@` can change the behavior here.
However, in the C code, I see that `int` is assumed to be at least 4 bytes, so it is likely the same on all currently supported platforms even if it is not the same in general.

radarhere: I hadn't read the linked documentation in enough detailed to notice the distinction before I made my suggestion. Thanks nulano for catching that.

I don't have strong feelings, I was just trying to follow a convention. If there is a preference to future proof this code by using `=`, that's fine with me.

ZachNagengast: Makes sense, @nulano let us know if you want to be sure to include this

ZachNagengast: Thanks @nulano, just marked this thread as unresolved.

ZachNagengast:
Any tasks still pending for this?

nulano (approved):

radarhere:
I just updated the title, since FT_PIXEL_MODE_GRAY has been changed as well as FT_PIXEL_MODE_BGRA

PR #7509, opened 2023-10-31
Mention defusedxml and olefile in installation instructions

radarhere:
Alternative to #7502

Under 'Basic Installation', mention that olefile can additionally be installed to allow Pillow to read FPX and
MIC images.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -42,8 +42,7 @@ Install Pillow with :command:`pip`::
     python3 -m pip install --upgrade pip
     python3 -m pip install --upgrade Pillow
 
-:pypi:`olefile` can additionally be installed to allow Pillow to read FPX and
-MIC images::
+Optionally, install :pypi:`olefile` for Pillow to read FPX and MIC images::
 
     python3 -m pip install --upgrade olefile
 
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -42,9 +42,10 @@ Install Pillow with :command:`pip`::
     python3 -m pip install --upgrade pip
     python3 -m pip install --upgrade Pillow
 
-Optionally, install :pypi:`olefile` for Pillow to read FPX and MIC images::
+Optionally, install :pypi:`defusedxml` for Pillow to read XMP data,
+and :pypi:`olefile` for Pillow to read FPX and MIC images::
 
-    python3 -m pip install --upgrade olefile
+    python3 -m pip install --upgrade defusedxml olefile
 
 
 .. tab:: Linux
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -42,6 +42,11 @@ Install Pillow with :command:`pip`::
     python3 -m pip install --upgrade pip
     python3 -m pip install --upgrade Pillow
 
+:pypi:`olefile` can additionally be installed to allow Pillow to read FPX and
+MIC images::
+
+    python3 -m pip install --upgrade olefile
+
 
 .. tab:: Linux
 


docs/installation.rst
```diff
@@ -42,6 +42,11 @@ Install Pillow with :command:`pip`::
     python3 -m pip install --upgrade pip
     python3 -m pip install --upgrade Pillow
 
+:pypi:`olefile` can additionally be installed to allow Pillow to read FPX and
+MIC images::
```
hugovk, line 46: With "active voice" instead of "passive":
```suggestion
Optionally, install :pypi:`olefile` for Pillow to read FPX and MIC images::
```
```

**Review by hugovk — APPROVED** [review]

File: docs/installation.rst [hunk 2]
```diff
@@ -42,6 +42,10 @@ Install Pillow with :command:`pip`::
     python3 -m pip install --upgrade pip
     python3 -m pip install --upgrade Pillow
 
+Optionally, install :pypi:`olefile` for Pillow to read FPX and MIC images::
+
+    python3 -m pip install --upgrade olefile
```
nulano, line 47: Perhaps also mention `defusedxml` here?

```suggestion
Optionally, install :pypi:`defusedxml` for Pillow to read XMP data,
and :pypi:`olefile` for Pillow to read FPX and MIC images::

    python3 -m pip install --upgrade defusedxml olefile
```

nulano (approved):

docs/installation.rst
radarhere: Good point!

F30:
I think tkinter should also be mentioned here. It is maintained as part of the Python standard library, but but packaged separately by most distributions. That unfortunately means you cannot install it from PyPI, but depending on where you installed Python from, you will need something like `python3-tk`.

See also https://github.com/python-pillow/Pillow/pull/7502#issuecomment-1791184254.

radarhere:
@F30 has said that [they are in favour this change by itself](https://github.com/python-pillow/Pillow/pull/7502#issuecomment-1791184254), so I'm going to merge this.

I originally intended this as an alternative, but it has ended up as an addendum. I think it will be simpler to make any changes regarding tkinter in #7502

radarhere:
> I think tkinter should also be mentioned here. It is maintained as part of the Python standard library, but but packaged separately by most distributions. That unfortunately means you cannot install it from PyPI, but depending on where you installed Python from, you will need something like python3-tk.

Pillow uses defusedxml and olefile to read image data. I don't think Pillow uses tkinter, it just interfaces with it. If you don't have tkinter installed, you are not missing out on any Pillow functionality.

I also don't expect it to be common for people to install Pillow so that they can manipulate images, and then think about how to display those images in a fully fledged UI. I think it would be far more common for people to decide to build a UI with Python, and then look for a library to prepare images for that. In other words, I think it would make more sense for tkinter to suggest users install Pillow, and actually, [it does!](https://docs.python.org/3/library/tkinter.html#images)

F30:
> Pillow uses defusedxml and olefile to read image data. I don't think Pillow uses tkinter, it just interfaces with it. If you don't have tkinter installed, you are not missing out on any Pillow functionality.

Are you sure about that?

https://pillow.readthedocs.io/en/stable/reference/ImageTk.html says "The `ImageTk` module contains support to create and modify Tkinter BitmapImage and PhotoImage objects". To me, this sounds like some image formats from the Tk world. But I'm not very familiar with that, so it may also be possible that BitmapImage and PhotoImage are not full-blown image formats, but just useful for building UIs.

PR #7524, opened 2023-11-06
ruff: Minor optimizations of list comprehensions, x in set, etc.

cclauss:
 * Add ruff rules C4 to the linting process - https://docs.astral.sh/ruff/rules/#flake8-comprehensions-c4
 * Use ruff rule PERF102 to speed up dict iteration - https://docs.astral.sh/ruff/rules/incorrect-dict-iterator
 * Use ruff rule PIE810 to speed up `str.endswith()` - https://docs.astral.sh/ruff/rules/multiple-starts-ends-with
 * Use ruff rules PyLint Refactor to speed up comparisons and remove useless code
     * https://docs.astral.sh/ruff/rules/#pylint-pl

### DIFF
--- a/Tests/test_image_quantize.py
+++ b/Tests/test_image_quantize.py
@@ -67,7 +67,7 @@ def test_quantize_no_dither():
 
 def test_quantize_no_dither2():
     im = Image.new("RGB", (9, 1))
-    im.putdata(list((p,) * 3 for p in range(0, 36, 4)))
+    im.putdata([(p,) * 3 for p in range(0, 36, 4)])
 
     palette = Image.new("P", (1, 1))
     data = (0, 0, 0, 32, 32, 32)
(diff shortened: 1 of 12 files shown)


cclauss:
https://docs.astral.sh/ruff/rules/repeated-equality-comparison -- Please read `Why is this bad?`

Yay295:
Python bytecode is still further interpreted, so I wouldn't necessarily assume that one is faster just because it's shorter. I have done some more looking though and found some more tests:
https://stackoverflow.com/questions/28885132/why-is-x-in-x-faster-than-x-x
https://github.com/pylint-dev/pylint/issues/4776

Basically, `in` is faster (even sometimes for only one value!) because it's done in C instead of in Python (and it also does an identity check), and sets have some optimizations to be fast even when they're small.

homm:
+1 for frozensets

radarhere (approved):

PR #7526, opened 2023-11-06
Fix up most noqas

akx:
Follows up on https://github.com/python-pillow/Pillow/pull/6966#issuecomment-1794626628 to get rid of `# noqa` comments.

Some are still left in tricky places, such as the build backend's star import, and the TIFF chunk dispatch registry.

### DIFF
--- a/Tests/bench_cffi_access.py
+++ b/Tests/bench_cffi_access.py
@@ -45,7 +45,7 @@ def test_direct():
 
     assert caccess[(0, 0)] == access[(0, 0)]
 
-    print("Size: %sx%s" % im.size)  # noqa: UP031
+    print(f"Size: {im.width}x{im.height}")
     timer(iterate_get, "PyAccess - get", im.size, access)
     timer(iterate_set, "PyAccess - set", im.size, access)
     timer(iterate_get, "C-api - get", im.size, caccess)
--- a/src/PIL/EpsImagePlugin.py
+++ b/src/PIL/EpsImagePlugin.py
@@ -77,12 +77,11 @@ def Ghostscript(tile, size, fp, scale=1, transparency=False):
 
     # Hack to support hi-res rendering
     scale = int(scale) or 1
-    size = (size[0] * scale, size[1] * scale)
+    width = size[0] * scale
+    height = size[1] * scale
     # resolution is dependent on bbox and size
-    res = (
-        72.0 * size[0] / (bbox[2] - bbox[0]),
-        72.0 * size[1] / (bbox[3] - bbox[1]),
-    )
+    res_x = 72.0 * width / (bbox[2] - bbox[0])
+    res_y = 72.0 * height / (bbox[3] - bbox[1])
 
     out_fd, outfile = tempfile.mkstemp()
     os.close(out_fd)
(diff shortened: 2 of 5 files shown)


Tests/bench_cffi_access.py
```diff
@@ -45,7 +45,8 @@ def test_direct():
 
     assert caccess[(0, 0)] == access[(0, 0)]
 
-    print("Size: %sx%s" % im.size)  # noqa: UP031
+    w, h = im.size
+    print(f"Size: {w}x{h}")
```
homm, line 49:
```suggestion
    print(f"Size: {im.width}x{im.height}")
```

homm (approved):

src/PIL/EpsImagePlugin.py
```diff
@@ -117,12 +117,15 @@ def Ghostscript(tile, size, fp, scale=1, transparency=False):
 
     device = "pngalpha" if transparency else "ppmraw"
 
+    w, h = size
```
radarhere, line 120: Minor concern: declaring `w` and `h` here seems a bit odd to me, since the width and height have already been used earlier in
https://github.com/python-pillow/Pillow/blob/8a13bf8b936497303f028b458483f2e8a64e96c6/src/PIL/EpsImagePlugin.py#L85-L86

akx: Yeah, I was thinking about that but didn't want to make a more intrusive change here. If you like, I can, though (though it'll conflict with https://github.com/python-pillow/Pillow/pull/7527/files#diff-c43e9ae25942c60787ebabdbfabddf8618c31d166f93520ea22c8547957f7e9eL80 so I'd love for that to get merged first :) ).

radarhere: #7527 has now been merged

src/PIL/IcnsImagePlugin.py
```diff
@@ -391,8 +391,8 @@ def _accept(prefix):
     with open(sys.argv[1], "rb") as fp:
         imf = IcnsImageFile(fp)
         for size in imf.info["sizes"]:
-            imf.size = size
-            imf.save("out-%s-%s-%s.png" % size)  # noqa: UP031
+            w, h, x = imf.size = size
+            imf.save(f"out-{w}-{h}-{x}.png")
```
radarhere, line 395:
```suggestion
            width, height, scale = imf.size = size
            imf.save(f"out-{width}-{height}-{scale}.png")
```
The third part is "scale", as seen under "sizes" at https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html#icns

src/PIL/Image.py
```diff
@@ -3100,7 +3100,8 @@ def fromarray(obj, mode=None):
         try:
             mode, rawmode = _fromarray_typemap[typekey]
         except KeyError as e:
-            msg = "Cannot handle this data type: %s, %s" % typekey  # noqa: UP031
+            typ_shape, typ_type = typekey
+            msg = f"Cannot handle this data type: {typ_shape}, {typ_type}"
```
radarhere, line 3104:
```suggestion
            typekey_shape, typestr = typekey
            msg = f"Cannot handle this data type: {typekey_shape}, {typestr}"
```
`typestr` seems more appropriate to me, given https://github.com/python-pillow/Pillow/blob/25cc5afbb115db5449bb18a45431e94ebf5ad01c/src/PIL/Image.py#L3096

radarhere (approved):

hugovk:
Thanks!

PR #7535, opened 2023-11-08
Fix for Python 3.13: explicitly include `unistd.h`

hugovk:

 * Test Python 3.13 pre-release
 * Explicitly include `unistd.h` to use `lseek`

Changed in [Python 3.13](https://docs.python.org/dev/whatsnew/3.13.html#id8):

> * ``Python.h`` no longer includes the ``<unistd.h>`` standard header file. If
  needed, it should now be included explicitly. For example, it provides the
  functions: ``read()``, ``write()``, ``close()``, ``isatty()``, ``lseek()``,
  ``getpid()``, ``getcwd()``, ``sysconf()``, ``getpagesize()``, ``alarm()`` and
  ``pause()``.
  As a consequence, ``_POSIX_SEMAPHORES`` and ``_POSIX_THREADS`` macros are no
  longer defined by ``Python.h``. The ``HAVE_UNISTD_H`` and ``HAVE_PTHREAD_H``
  macros defined by ``Python.h`` can be used to decide if ``<unistd.h>`` and
  ``<pthread.h>`` header files can be included.

So we need to `#include <unistd.h>` for `src/libImaging/TiffDecode.c` which uses `lseek`.

---

Also add Python 3.13 to the CI for Ubuntu, macOS and Windows. We need to skip `cffi`, it's not ready for 3.13.

3.13 is in alpha, so we can expect things still to change. I'm fine leaving it out of the CI until later or beta, but it could be useful to get early feedback. If any sudden failures turn up, we can always comment it out until we get it sorted (in Pillow or upstream in CPython, as necessary).

### DIFF
--- a/.ci/install.sh
+++ b/.ci/install.sh
@@ -28,7 +28,8 @@ fi
 
 python3 -m pip install --upgrade pip
 python3 -m pip install --upgrade wheel
-PYTHONOPTIMIZE=0 python3 -m pip install cffi
+# TODO Update condition when cffi supports 3.13
+if ! [[ "$GHA_PYTHON_VERSION" == "3.13" ]]; then PYTHONOPTIMIZE=0 python3 -m pip install cffi ; fi
 python3 -m pip install coverage
 python3 -m pip install defusedxml
 python3 -m pip install olefile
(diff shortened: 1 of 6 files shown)


Yay295:
Isn't `unistd.h` Unix only? How is `TiffDecode.c` building on Windows?

radarhere (approved):

Issue #7348, opened 2023-08-21
Release Pillow 10.1.0 on October 15, 2023


hugovk:
Highlights:

* https://github.com/python-pillow/Pillow/issues/6941

## TODO

* [x] Merge https://github.com/python-pillow/pillow-wheels/pull/400 before release
* [x] https://github.com/python-pillow/Pillow/issues/6941

## Main Release

Released quarterly on January 2nd, April 1st, July 1st and October 15th.

* [x] Open a release ticket e.g. https://github.com/python-pillow/Pillow/issues/3154
* [x] Develop and prepare release in `main` branch.
* [x] Check [GitHub Actions](https://github.com/python-pillow/Pillow/actions) and [AppVeyor](https://ci.appveyor.com/project/python-pillow/Pillow) to confirm passing tests in `main` branch.
* [x] Check that all of the wheel builds pass the tests in the [GitHub Actions "Wheels" workflow](https://github.com/python-pillow/Pillow/actions/workflows/wheels.yml) and [Travis CI](https://app.travis-ci.com/github/python-pillow/pillow) jobs by manually triggering them.
* [x] In compliance with [PEP 440](https://peps.python.org/pep-0440/), update version identifier in `src/PIL/_version.py`
* [x] Update `CHANGES.rst`.
* [x] Run pre-release check via `make release-test` in a freshly cloned repo.
* [x] Create branch and tag for release e.g.:
  ```bash
  git branch 5.2.x
  git tag 5.2.0
  git push --tags
  ```
* [x] Create and check source distribution:
  ```bash
  make sdist
  ```
* [x] Create [binary distributions](https://github.com/python-pillow/Pillow/blob/main/RELEASING.md#binary-distributions)
* [x] Check and upload all binaries and source distributions e.g.:
  ```bash
  python3 -m twine check --strict dist/*
  python3 -m twine upload dist/Pillow-5.2.0*
  ```
* [x] Publish the [release on GitHub](https://github.com/python-pillow/Pillow/releases)
* [x] In compliance with [PEP 440](https://peps.python.org/pep-0440/),
      increment and append `.dev0` to version identifier in `src/PIL/_version.py` and then:
  ```bash
  git push --all
   ```

### macOS and Linux
* [x] Download wheels from the [GitHub Actions "Wheels" workflow](https://github.com/python-pillow/Pillow/actions/workflows/wheels.yml)
  and copy into `dist/`. For example using [GitHub CLI](https://github.com/cli/cli):
[...]

Issue #6179, opened 2022-04-04
Image import fails on macOS 10.13 with Pillow 9.1.0


AlphonsG:
### What did you do?

Tried running `from PIL import Image` after installing Pillow from pip.

### What did you expect to happen?

Import to work.

### What actually happened?

Error: 

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/alphons/.venv/rainbow-env/lib/python3.8/site-packages/PIL/Image.py", line 132, in <module>
    from . import _imaging as core
ImportError: dlopen(/Users/alphons/.venv/rainbow-env/lib/python3.8/site-packages/PIL/_imaging.cpython-38-darwin.so, 2): Symbol not found: ____chkstk_darwin
  Referenced from: /Users/alphons/.venv/rainbow-env/lib/python3.8/site-packages/PIL/.dylibs/libtiff.5.dylib (which was built for Mac OS X 11.0)
  Expected in: /usr/lib/libSystem.B.dylib
 in /Users/alphons/.venv/rainbow-env/lib/python3.8/site-packages/PIL/.dylibs/libtiff.5.dylib
```

### What are your OS, Python and Pillow versions?

* OS: macOS 10.13 High Sierra (64 bit)
* Python: 3.8.10
* Pillow: 9.1.0

*Note: import works using Pillow 9.0.1*

radarhere:
Hi. We've changed a few settings in our wheels since January, so this could be a few different things. I've put together a Pillow 9.1.0 wheel for your system the way we did it for Pillow 9.0.1 - [Pillow-9.1.0-cp38-cp38-macosx_10_10_x86_64.whl.zip](https://github.com/python-pillow/Pillow/files/8407703/Pillow-9.1.0-cp38-cp38-macosx_10_10_x86_64.whl.zip). Would you be able to test it and see if that works?

If it does work, would you be willing to test out some more variations on that wheel, so that we can figure out exactly what the problem is? macOS 10.13 is not a system that I have easy access to.

If that doesn't work, and you're looking for things to try immediately, you can install Pillow from source - `python3 -m pip install Pillow --no-binary :all:`

hugovk:
Note that [macOS 10.13 (High Sierra) is now 16 months EOL](https://endoflife.date/macos) and two versions out of support, so strictly speaking [not supported by Pillow](https://pillow.readthedocs.io/en/stable/installation.html#platform-support).

[...]

PR #7552, opened 2023-11-14
Use cibuildwheel

radarhere:
Helps https://github.com/python-pillow/Pillow/issues/7390. 

It has been [suggested](https://github.com/python-pillow/Pillow/issues/7348#issuecomment-1763372101) that we move from [multibuild](https://github.com/multi-build/multibuild) to [cibuildwheel](https://github.com/pypa/cibuildwheel).

Looking at cibuildwheel though, it doesn't have [multibuild's library builders](https://github.com/multi-build/multibuild/blob/devel/library_builders.sh) to build our dependencies.

So this PR employs a hybrid approach, using multibuild to build the dependencies first, and then cibuildwheel to build the wheels.

It seems to me that the primary advantage of cibuildwheel is speed, in that the dependencies are built less, and re-used by different Python environments.


.github/workflows/wheels.yml
```diff
@@ -24,15 +24,47 @@ env:
...
+      - uses: actions/checkout@v4
+        with:
+          submodules: true
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.x"
+      - name: Build wheels
+        run: |
+          python3 -m pip install cibuildwheel
+          python3 -m cibuildwheel --output-dir wheelhouse
```
hugovk, line 58: Perhaps we can use the action?
```suggestion
      - name: Build wheels
        uses: pypa/cibuildwheel@v2.16.2
        with:
          output-dir: wheelhouse
```

And https://cibuildwheel.readthedocs.io/en/stable/faq/#automatic-updates recommends pinning, whatever method is used:

> Selecting a moving target (like the latest release) is generally a bad idea in CI. If something breaks, you can't tell whether it was your code or an upstream update that caused the breakage, and in a worse-case scenario, it could occur during a release.

Renovate will keep it up-to-date.

.github/workflows/wheels.yml
```diff
@@ -24,15 +24,44 @@ env:
...
+          output-dir: wheelhouse
+        env:
+          CIBW_ARCHS: ${{ matrix.archs }}
+          CIBW_BUILD: ${{ matrix.build }}
+          CIBW_CONFIG_SETTINGS: raqm=enable raqm=vendor fribidi=vendor
+          CIBW_MANYLINUX_PYPY_X86_64_IMAGE: ${{ matrix.manylinux }}
+          CIBW_MANYLINUX_X86_64_IMAGE: ${{ matrix.manylinux }}
+      - uses: actions/upload-artifact@v3
+        with:
+          path: ./wheelhouse/*.whl
```
hugovk, line 64: To put them in the same `dist.zip` as the sdist, rather then the default `artifacts.zip`:

```suggestion
          path: ./wheelhouse/*.whl
          artifacts-name: "dist"
```

hugovk:
This is looking good, thanks!

Comparing the zip generated [before](https://github.com/python-pillow/Pillow/actions/runs/6842495477?pr=7544) with this PR:

image

The sdist covered above.

---

We're also building PyPy 3.8 wheels in addition to PyPy 3.9-3.10, I think it's because cibuildwheel by default looks at our `requires-python = ">=3.8"`. We should be able to [skip](https://cibuildwheel.readthedocs.io/en/stable/options/#build-skip) with something like this: 

```yml
CIBW_SKIP: pp38-*
```

---

The macOS Intel wheels are built for 10.9 here instead of 10.10 before. I don't think this matters?

hugovk:
A quick inspection of Intel Mac, manylinux, and musllinux wheels, they look like they have everything.

.github/workflows/wheels.yml
```diff
@@ -24,15 +24,44 @@ env:
...
+      - name: Build wheels
+        uses: pypa/cibuildwheel@v2.16.2
+        with:
+          output-dir: wheelhouse
+        env:
+          CIBW_ARCHS: ${{ matrix.archs }}
+          CIBW_BUILD: ${{ matrix.build }}
+          CIBW_CONFIG_SETTINGS: raqm=enable raqm=vendor fribidi=vendor
+          CIBW_MANYLINUX_PYPY_X86_64_IMAGE: ${{ matrix.manylinux }}
+          CIBW_MANYLINUX_X86_64_IMAGE: ${{ matrix.manylinux }}
```
nulano, line 61: > While arm64 wheels can be built on x86_64, they cannot be tested. The ability to test the arm64 wheels will be added in a future release of cibuildwheel, once Apple Silicon CI runners are widely available. To silence this warning, set `CIBW_TEST_SKIP: *-macosx_arm64`.

```suggestion
          CIBW_MANYLINUX_X86_64_IMAGE: ${{ matrix.manylinux }}
          CIBW_TEST_SKIP: *-macosx_arm64
```

radarhere: Sure, I've added a commit for this. Testing, I found that it required quotes, as per https://cibuildwheel.readthedocs.io/en/stable/options/#test-skip

.github/workflows/wheels.yml
radarhere: Good call. I've added a commit to add `name`, as per https://github.com/python-pillow/Pillow/blob/da0b826e960474669f630a419ecc96b8540b52a7/.github/workflows/wheels.yml#L51-L54

radarhere:
> The macOS Intel wheels are built for 10.9 here instead of 10.10 before. I don't think this matters?

Taking a look, I found that #6179 had a problem when I accidentally reverted it to 10.9 once before, so I've pushed a commit to restore 10.10.

hugovk:
Thanks for the update. Now we get the same filenames before and after: also the sdist, and 10_10 files instead of 10_9. 👍 

However, I'm still getting this for `pillow-10.2.0.dev0-cp311-cp311-macosx_11_0_arm64.whl`:
```pytb
ImportError: dlopen(/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/PIL/_imaging.cpython-311-darwin.so, 0x0002): symbol not found in flat namespace
'_jpeg_resync_to_restart'
```

We still seem to be missing `.dylibs`:

<details>
<summary>Details</summary>

image


</details>

hugovk:
On an Intel Mac, `pillow-10.2.0.dev0-cp311-cp311-macosx_10_10_x86_64.whl` does install, and a quick `Image.new()` and `.show()` work.

radarhere:
Ok, the macOS arm64 wheels should be better now.

hugovk:
Thanks! I tested the cp38 to cp312 macOS arm64 wheels and they all passed a simple install/`im=Image.new("RGB", (20, 20), "red")`/`im.show()` test.

The wheels contain the same files as before:

<details>
<summary>Details</summary>

image


</details>

I'll test the Intel ones later. I note they now contain libjpeg and libzstd dylib files that they didn't have before:

<details>
<summary>Details</summary>

image


</details>

It would be good if anyone can help test out some manylinux and musllinux wheels: go to https://github.com/python-pillow/Pillow/actions/runs/6901285566?pr=7552 and download the "dist" artifact, and `pip install <filename>.whl`.

nulano:
I can see that the libzstd dylib has different permissions (`-r--r--r--`) than most of the other files (`-rwx-r-xr-x`). Don't know if that is an issue, but it usually means it was copied from the system and not built during the release.

> It would be good if anyone can help test out some manylinux and musllinux wheels

`/pillow-10.2.0.dev0-cp38-cp38-manylinux_2_28_x86_64.whl` passes tests on my Ubuntu Server system.

All of the musllinux wheels pass tests in the corresponding `python:3.X-alpine` Docker containers.

hugovk:
✅  Working on Intel macOS with Python 3.8-3.12 with:
```sh
#!/usr/bin/env bash

for VARIABLE in 8 9 10 11 12
do
	python3.$VARIABLE -m pip install -U pip
	python3.$VARIABLE -m pip install pillow-10.2.0.dev0-cp3$VARIABLE-cp3$VARIABLE-macosx_10_10_x86_64.whl
	python3.$VARIABLE 1.py
done
```
```python
from PIL import Image
im = Image.new("RGB", (10, 20), "red")
im.show()
```

radarhere:
> I can see that the libzstd dylib has different permissions (-r--r--r--) than most of the other files (-rwxr-xr-x). Don't know if that is an issue, but it usually means it was copied from the system and not built during the release.

Ok, I've pushed a commit to remove it again, like it was before this PR.

hugovk:
✅ Looking good with https://github.com/python-pillow/Pillow/pull/7552#issuecomment-1816955168 on both Intel and Apple Silicon for all versions.

.github/workflows/wheels.yml
hugovk: I've created a PR to fix this upstream: https://github.com/pypa/cibuildwheel/pull/1665

hugovk (approved):
Thank you!

PR #7553, opened 2023-11-14
Add `keep_rgb` option when saving JPEG to prevent conversion of RGB colorspace

bgilbert:
libjpeg automatically converts RGB to YCbCr by default.  Add a `keep_rgb` option to disable libjpeg's automatic conversion of RGB images during write.

### DIFF
--- a/Tests/test_file_jpeg.py
+++ b/Tests/test_file_jpeg.py
@@ -145,11 +145,13 @@ class TestFileJpeg:
         def getchannels(im):
             return tuple(v[0] for v in im.layer)
 
-        im = self.roundtrip(hopper())
-        assert getchannels(im) == (1, 2, 3)
-        im = self.roundtrip(hopper(), keep_rgb=True)
-        assert getchannels(im) == (ord("R"), ord("G"), ord("B"))
-        assert_image_similar(hopper(), im, 12)
+        im = hopper()
+        im_ycbcr = self.roundtrip(im)
+        assert getchannels(im_ycbcr) == (1, 2, 3)
+
+        im_rgb = self.roundtrip(im, keep_rgb=True)
+        assert getchannels(im_rgb) == (ord("R"), ord("G"), ord("B"))
+        assert_image_similar(im, im_rgb, 12)
 
     @pytest.mark.parametrize(
         "test_image_path",
@@ -434,30 +436,25 @@ class TestFileJpeg:
         # experimental API
         im = self.roundtrip(hopper(), subsampling=-1)  # default
         assert getsampling(im) == (2, 2, 1, 1, 1, 1)
-        im = self.roundtrip(hopper(), subsampling=0)  # 4:4:4
-        assert getsampling(im) == (1, 1, 1, 1, 1, 1)
-        im = self.roundtrip(hopper(), subsampling=1)  # 4:2:2
-        assert getsampling(im) == (2, 1, 1, 1, 1, 1)
-        im = self.roundtrip(hopper(), subsampling=2)  # 4:2:0
-        assert getsampling(im) == (2, 2, 1, 1, 1, 1)
+        for subsampling in (0, "4:4:4"):
+            im = self.roundtrip(hopper(), subsampling=subsampling)
+            assert getsampling(im) == (1, 1, 1, 1, 1, 1)
+        for subsampling in (1, "4:2:2"):
+            im = self.roundtrip(hopper(), subsampling=subsampling)
+            assert getsampling(im) == (2, 1, 1, 1, 1, 1)
+        for subsampling in (2, "4:2:0", "4:1:1"):
+            im = self.roundtrip(hopper(), subsampling=subsampling)
+            assert getsampling(im) == (2, 2, 1, 1, 1, 1)
+
         im = self.roundtrip(hopper(), subsampling=3)  # default (undefined)
         assert getsampling(im) == (2, 2, 1, 1, 1, 1)
-
-        im = self.roundtrip(hopper(), subsampling="4:4:4")
-        assert getsampling(im) == (1, 1, 1, 1, 1, 1)
-        im = self.roundtrip(hopper(), subsampling="4:2:2")
-        assert getsampling(im) == (2, 1, 1, 1, 1, 1)
-        im = self.roundtrip(hopper(), subsampling="4:2:0")
-        assert getsampling(im) == (2, 2, 1, 1, 1, 1)
-        im = self.roundtrip(hopper(), subsampling="4:1:1")
-        assert getsampling(im) == (2, 2, 1, 1, 1, 1)
-
-        with pytest.raises(TypeError):
-            self.roundtrip(hopper(), subsampling="1:1:1")
 
         # RGB colorspace, no subsampling by default
         im = self.roundtrip(hopper(), subsampling=3, keep_rgb=True)
         assert getsampling(im) == (1, 1, 1, 1, 1, 1)
+
+        with pytest.raises(TypeError):
+            self.roundtrip(hopper(), subsampling="1:1:1")
 
     def test_exif(self):
         with Image.open("Tests/images/pil_sample_rgb.jpg") as im:
--- a/Tests/test_file_jpeg.py
+++ b/Tests/test_file_jpeg.py
@@ -141,6 +141,16 @@ class TestFileJpeg:
             )
             assert k > 0.9
 
+    def test_rgb(self):
+        def getchannels(im):
+            return tuple(v[0] for v in im.layer)
+
+        im = self.roundtrip(hopper())
+        assert getchannels(im) == (1, 2, 3)
+        im = self.roundtrip(hopper(), keep_rgb=True)
+        assert getchannels(im) == (ord("R"), ord("G"), ord("B"))
+        assert_image_similar(hopper(), im, 12)
+
     @pytest.mark.parametrize(
         "test_image_path",
         [TEST_FILE, "Tests/images/pil_sample_cmyk.jpg"],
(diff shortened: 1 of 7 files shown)


docs/handbook/image-file-formats.rst
```diff
@@ -483,6 +483,13 @@ The :py:meth:`~PIL.Image.Image.save` method supports the following options:
 **exif**
     If present, the image will be stored with the provided raw EXIF data.
 
+**keep_colorspace**
```
homm, line 486: For me this is not clear name, since it affects only RGB to YCbCr conversion

bgilbert: It prevents libjpeg from doing _any_ automatic colorspace conversion, and currently that's only relevant for RGB to YCbCr.  If you'd prefer, I can rename to e.g. `keep_rgb` and change the code to specifically only handle the RGB case.

akx: I think this is better than `keep_rgb` would be.

bgilbert:
Rebased (first push) and switched to `keep_rgb` (second push).

src/libImaging/JpegEncode.c
```diff
@@ -137,6 +137,20 @@ ImagingJpegEncode(Imaging im, ImagingCodecState state, UINT8 *buf, int bytes) {
...
+            if (context->keep_rgb) {
+                switch (context->cinfo.in_color_space) {
+                case JCS_RGB:
+#ifdef JCS_EXTENSIONS
+                case JCS_EXT_RGBX:
+#endif
+                    jpeg_set_colorspace(&context->cinfo, JCS_RGB);
+                    break;
+                default:
+                    break;
```
radarhere, line 150:
```suggestion
                    case JCS_RGB:
#ifdef JCS_EXTENSIONS
                    case JCS_EXT_RGBX:
#endif
                        jpeg_set_colorspace(&context->cinfo, JCS_RGB);
```
A stylistic suggestion. Other people might have different ideas, so feel free to take or leave this as you see fit.

docs/handbook/image-file-formats.rst
```diff
@@ -486,6 +486,12 @@ The :py:meth:`~PIL.Image.Image.save` method supports the following options:
 **exif**
     If present, the image will be stored with the provided raw EXIF data.
 
+**keep_rgb**
+    If present and true, and the image's color space is RGB, the image will
+    be stored as RGB instead of automatically converted to YCbCr.
```
radarhere, line 491:
```suggestion
    By default, RGB colorspace images will be converted to YCbCr by libjpeg.

    If this option is present and true, those images will be stored as RGB instead.
```

src/libImaging/JpegEncode.c
bgilbert: I've redone the indent.  The other stuff is there to prevent compiler warnings about missing switch cases (since the switch expression is an enum) and fallthrough.  I don't know what compiler flags Pillow uses under all circumstances, but it seemed better to be explicit.

docs/handbook/image-file-formats.rst
bgilbert: Updated with a variant of this.

src/libImaging/JpegEncode.c
nulano: I don't think it would generate warnings, but I do think it is easier to read with the `break`s and `default`.
Fall-through is intentionally used in `_imagingft.c`, but it is accompanied by a comment to make it clear it is intentional: https://github.com/python-pillow/Pillow/blob/f9c7bd86115090ce2b110794485f138f0d5d5953/src/_imagingft.c#L1023-L1025

Tests/test_file_jpeg.py
```diff
@@ -445,6 +455,10 @@ def getsampling(im):
         with pytest.raises(TypeError):
             self.roundtrip(hopper(), subsampling="1:1:1")
 
+        # RGB colorspace, no subsampling by default
+        im = self.roundtrip(hopper(), subsampling=3, keep_rgb=True)
```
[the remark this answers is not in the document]
bgilbert: libjpeg doesn't subsample RGB output by default, since it doesn't make sense to have e.g. lower resolution in the green channel.  The test verifies that an RGB image doesn't get subsampled when we use that default.

bgilbert:
Added @radarhere's test improvements from https://github.com/bgilbert/Pillow/pull/3.

Tests/test_file_jpeg.py
```diff
@@ -424,23 +436,22 @@ def getsampling(im):
...
+            im = self.roundtrip(hopper(), subsampling=subsampling)
+            assert getsampling(im) == (1, 1, 1, 1, 1, 1)
+        for subsampling in (1, "4:2:2"):
+            im = self.roundtrip(hopper(), subsampling=subsampling)
+            assert getsampling(im) == (2, 1, 1, 1, 1, 1)
+        for subsampling in (2, "4:2:0", "4:1:1"):
+            im = self.roundtrip(hopper(), subsampling=subsampling)
+            assert getsampling(im) == (2, 2, 1, 1, 1, 1)
+
         im = self.roundtrip(hopper(), subsampling=3)  # default (undefined)
```
homm, line 449: For me undefined means undefined, not particular invalid value )
I'm for removing this argument

Tests/test_file_jpeg.py
```diff
@@ -422,25 +434,22 @@ def getsampling(im):
...
+            assert getsampling(im) == (1, 1, 1, 1, 1, 1)
+        for subsampling in (1, "4:2:2"):
+            im = self.roundtrip(hopper(), subsampling=subsampling)
+            assert getsampling(im) == (2, 1, 1, 1, 1, 1)
+        for subsampling in (2, "4:2:0", "4:1:1"):
+            im = self.roundtrip(hopper(), subsampling=subsampling)
+            assert getsampling(im) == (2, 2, 1, 1, 1, 1)
+
+        # RGB colorspace, no subsampling by default
+        im = self.roundtrip(hopper(), keep_rgb=True)
```
homm, line 451: I believe we should ignore or raise errors on any `subsampling` other than default and `0` to prevent undesirable behavior.

homm (approved):

Tests/test_file_jpeg.py
radarhere: My preference would be for a separate PR.

bgilbert:
Rebased onto current `main` and added release note.

bgilbert:
Is anything else needed here?

docs/releasenotes/10.2.0.rst
```diff
@@ -46,6 +46,14 @@ Added DdsImagePlugin enums
 :py:class:`~PIL.DdsImagePlugin.DXGI_FORMAT` and :py:class:`~PIL.DdsImagePlugin.D3DFMT`
 enums have been added to :py:class:`PIL.DdsImagePlugin`.
 
+JPEG RGB color space
+^^^^^^^^^^^^^^^^^^^^
+
+When saving JPEG files, ``keep_rgb`` can now be set to ``True`` to store RGB images in
+the RGB color space instead of converting them to YCbCr.  When this option is enabled,
+attempting to chroma-subsample RGB images with the ``subsampling`` option will result
+in an error.
```
radarhere, line 55:
```suggestion
When saving JPEG files, ``keep_rgb`` can now be set to ``True``. This will store RGB
images in the RGB color space instead of being converted to YCbCr automatically by
libjpeg. When this option is enabled, attempting to chroma-subsample RGB images with
the ``subsampling`` option will raise an :py:exc:`OSError`.
```

docs/handbook/image-file-formats.rst
```diff
@@ -486,6 +486,16 @@ The :py:meth:`~PIL.Image.Image.save` method supports the following options:
 **exif**
     If present, the image will be stored with the provided raw EXIF data.
 
+**keep_rgb**
+    By default, libjpeg converts images with an RGB color space to YCbCr.
+    If this option is present and true, those images will be stored as RGB
+    instead.
+
+    When this option is enabled, attempting to chroma-subsample RGB images
+    with the ``subsampling`` option will result in an error.
```
radarhere, line 495:
```suggestion
    with the ``subsampling`` option will raise an :py:exc:`OSError`.
```

bgilbert:
Applied suggestions.

src/libImaging/JpegEncode.c
```diff
@@ -137,6 +137,30 @@ ImagingJpegEncode(Imaging im, ImagingCodecState state, UINT8 *buf, int bytes) {
...
+                    case JCS_RGB:
+#ifdef JCS_EXTENSIONS
+                    case JCS_EXT_RGBX:
+#endif
+                        switch (context->subsampling) {
+                            case -1:  /* Default */
+                            case 0:   /* No subsampling */
+                                break;
+                            default:
+                                /* Would subsample the GB channels, which
```
radarhere, line 152:
```suggestion
                                /* Would subsample the RGB channels, which
```
?

bgilbert: No, the GB channels.  We want to avoid sampling red at a higher rate than green and blue, as one might do with YCbCr.

Yay295: Maybe change it to "the G and B channels" so it looks less like a typo.

hugovk:
```suggestion
                                /* Would subsample the green and blue channels, which
```

bgilbert: Updated!

nulano (approved):

Tests/test_file_jpeg.py
```diff
@@ -141,6 +141,18 @@ def test_cmyk(self):
...
+        def getchannels(im):
+            return tuple(v[0] for v in im.layer)
+
+        im = hopper()
+        im_ycbcr = self.roundtrip(im)
+        assert getchannels(im_ycbcr) == (1, 2, 3)
+
+        im_rgb = self.roundtrip(im, keep_rgb=True)
+        assert getchannels(im_rgb) == (ord("R"), ord("G"), ord("B"))
+        assert_image_similar(im, im_rgb, 12)
```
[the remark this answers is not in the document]
bgilbert: Added a check for this, for symmetry with the RGB case.

PR #7555, opened 2023-11-15
Correct PDF palette size when saving

### LINKED ISSUE #7554: [10.1.0 regression] Palette images save as blank PDFs
Minimal example (you can use [this tiny PNG](https://github.com/python-pillow/Pillow/assets/1119169/d8d45152-7734-4fe3-a2d3-fb49839a0893) for example):

```python
from PIL import Image

image = Image.open('test.png')
image = image.convert('P')
image.save('test.pdf')
```

Output PDF with Pillow 10.0.1:
image

Output PDF with Pillow 10.1.0:
image

Issue faced with Python 3.11.6 on Ubuntu 22.04 and Debian 12 (bookworm). I also had the same issue in Docker environments, so I could make a Docker image if needed.

radarhere:
Resolves #7554

#7289 set the C palette to be empty by default. The issue has found that this causes P PDFs to appear blank. Testing, I find that the problem is that there can now be fewer than 256 palette entries saved to the PDF file, and that padding the palette to 256 entries fixes it.

### DIFF
--- a/src/PIL/PdfImagePlugin.py
+++ b/src/PIL/PdfImagePlugin.py
@@ -96,7 +96,7 @@ def _write_image(im, filename, existing_pdf, image_refs):
         dict_obj["ColorSpace"] = [
             PdfParser.PdfName("Indexed"),
             PdfParser.PdfName("DeviceRGB"),
-            255,
+            len(palette) // 3 - 1,
             PdfParser.PdfBinary(palette),
         ]
         procset = "ImageI"  # indexed color


src/PIL/PdfImagePlugin.py
```diff
@@ -93,6 +93,7 @@ def _write_image(im, filename, existing_pdf, image_refs):
     elif im.mode == "P":
         filter = "ASCIIHexDecode"
         palette = im.getpalette()
+        palette += (768 - len(palette)) * [0]
```
homm, line 96: For me `[array] * number` is more obvious

homm (approved):
There is a third member in `dict_obj["ColorSpace"]` list which is 255. Isn't better to adjust it?

radarhere:
Looking at page 164 of https://opensource.adobe.com/dc-acrobat-sdk-docs/pdfstandards/PDF32000_2008.pdf, '255' is the 'hival', the 'maximum valid index value'. We couldn't correctly adjust it to lower the number of required palette entries without also needing to remap the palette.

homm:
The maximum valid index value in the pallet (the current pallet, as I understand it). This is articulated because it's not a number or items, it's the index of the last element. "The colour table data shall be m•Y(hival +1) bytes long". So if we say it hival is 31, we can write (31+1)•3 = 96 items.

radarhere:
Ah, I was confused. Ok, I've adjusted that value instead.

PR #7557, opened 2023-11-16
If absent, do not try to close fp when closing image

RaphaelVRossi:

 *  Closes image when they are copied
 * Since image copied doesn't have 'fp' attribute, a message was logged as "error"

### DIFF
--- a/Tests/test_image.py
+++ b/Tests/test_image.py
@@ -1014,6 +1014,21 @@ class TestImage:
             except OSError as e:
                 assert str(e) == "buffer overrun when reading image file"
 
+    @pytest.fixture(scope="function")
+    def inject_caplog(self, caplog):
+        self._caplog = caplog
+
+    @pytest.mark.usefixtures("inject_caplog")
+    def test_close_graceful(self):
+        with Image.open("Tests/images/hopper.jpg") as im:
+            copy = im.copy()
+            im.close()
+            copy.close()
+
+            assert len(self._caplog.records) == 0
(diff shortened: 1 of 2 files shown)


Tests/test_image.py
```diff
@@ -1014,6 +1014,21 @@ def test_fli_overrun2(self):
             except OSError as e:
                 assert str(e) == "buffer overrun when reading image file"
 
+    @pytest.fixture(scope="function")
+    def inject_caplog(self, caplog):
+        self._caplog = caplog
+
+    @pytest.mark.usefixtures("inject_caplog")
+    def test_close_graceful(self):
```
akx, line 1022: Why not just use the caplog fixture directly?

radarhere: I think I've had the same idea in https://github.com/RaphaelVRossi/Pillow/pull/1

Tests/test_image.py
```diff
@@ -1014,6 +1014,21 @@ def test_fli_overrun2(self):
...
+        with Image.open("Tests/images/hopper.jpg") as im:
+            copy = im.copy()
+            im.close()
```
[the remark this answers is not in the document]
RaphaelVRossi: Yeah! 👍

Tests/test_image.py
RaphaelVRossi: @radarhere already fixed this! Thanks again.

radarhere (approved):

PR #7558, opened 2023-11-16
Add Windows 11 tested Pillow versions

nulano:
All (non-skipped) tests are passing on the added versions.

I tested these wheels on arm64: https://github.com/python-pillow/Pillow/discussions/7547#discussioncomment-7592240


radarhere (approved):

radarhere:
https://github.com/python-pillow/Pillow/issues/7390#issuecomment-1813365035
> ARM64 wheels (although I have no way to test them!).

I'm confused why you've said in #7390 that you have no way to test arm64 wheels, but you were able to here? Just checking that this was actually tested.

nulano:
I built and tested these on a relative's MacBook. I did not realize it was an M2 version until today (I thought they still had an Intel version).

radarhere:
Ok, thanks

Issue #6679, opened 2022-10-23
Failed to build Windows ARM64 dependencies


infinitepower18:
### What did you do?
I'm trying to build a Python 3.11 ARM64 wheel for Windows ARM64 using the [provided instructions](https://github.com/python-pillow/Pillow/blob/main/winbuild/build.rst) and used the command `build_prepare.py --architecture=ARM64`. However it fails when I try to build the dependencies afterwards. I have Visual Studio Community 2022 installed on my system.

### What did you expect to happen?
The dependencies should have been built successfully.

### What actually happened?
Fails when trying to build for ARM64

### What are your OS, Python and Pillow versions?

* OS: Windows 11 Build 22621.675
* Python: Python 3.11.0rc2
* Pillow: Cloned GitHub main branch

```
PS C:\Users\Ahnaf\Desktop\Pillow\winbuild\build> .\build_dep_all.cmd

C:\Users\Ahnaf\Desktop\Pillow\winbuild\build>cmd.exe /c "C:\Users\Ahnaf\Desktop\Pillow\winbuild\build\build_dep_libjpeg.cmd"
======================================================================
==== Building libjpeg (libjpeg-turbo-2.1.4)                       ====
======================================================================

C:\Users\Ahnaf\Desktop\Pillow\winbuild\build>cd /D C:\Users\Ahnaf\Desktop\Pillow\winbuild\build\libjpeg-turbo-2.1.4

C:\Users\Ahnaf\Desktop\Pillow\winbuild\build\libjpeg-turbo-2.1.4>set INCLUDE=C:\Users\Ahnaf\Desktop\Pillow\winbuild\build\inc

C:\Users\Ahnaf\Desktop\Pillow\winbuild\build\libjpeg-turbo-2.1.4>set INCLIB=C:\Users\Ahnaf\Desktop\Pillow\winbuild\build\lib

C:\Users\Ahnaf\Desktop\Pillow\winbuild\build\libjpeg-turbo-2.1.4>set LIB=C:\Users\Ahnaf\Desktop\Pillow\winbuild\build\lib

[...]

PR #7559, opened 2023-11-16
Windows arm64 dependency cross-compilation fixes

nulano:
For #6679 and https://github.com/python-pillow/Pillow/issues/7390#issuecomment-1813365035

When compiling Pillow directly on an ARM64 system, I found it worked after installing the correct Visual Studio components.
Cross-compiling required a few patches to the dependencies.

 * Select Visual Studio version only if it has ARM64 build tools installed (https://github.com/python-pillow/Pillow/issues/6679#issuecomment-1295644192)
 * Fix ARM64 target detection when cross-compiling libjpeg-turbo (if NASM is installed) and libimagequant (https://github.com/python-pillow/Pillow/issues/6679#issuecomment-1292620454)
 * Do not cross-compile FriBiDi (https://github.com/python-pillow/Pillow/issues/6679#issuecomment-1295510138)

### DIFF
--- a/winbuild/build_prepare.py
+++ b/winbuild/build_prepare.py
@@ -375,7 +375,7 @@ DEPS = {
 
 
 # based on distutils._msvccompiler from CPython 3.7.4
-def find_msvs(architecture) -> dict[str, str] | None:
+def find_msvs(architecture: str) -> dict[str, str] | None:
     root = os.environ.get("ProgramFiles(x86)") or os.environ.get("ProgramFiles")
     if not root:
         print("Program Files not found")
--- a/winbuild/build_prepare.py
+++ b/winbuild/build_prepare.py
@@ -118,6 +118,12 @@ DEPS = {
             "(LEGAL ISSUES\n============\n\n.+?)\n\nREFERENCES\n=========="
             ".+(libjpeg-turbo Licenses\n======================\n\n.+)$"
         ),
+        "patch": {
+            r"CMakeLists.txt": {
+                # libjpeg-turbo does not detect MSVC x86_arm64 cross-compiler correctly
+                'if(MSVC_IDE AND CMAKE_GENERATOR_PLATFORM MATCHES "arm64")': "if({architecture} STREQUAL ARM64)",  # noqa: E501
+            },
+        },
         "build": [
             *cmds_cmake(
                 ("jpeg-static", "cjpeg-static", "djpeg-static"),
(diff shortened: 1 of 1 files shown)


winbuild/build_prepare.py
```diff
@@ -689,6 +702,11 @@ def build_dep_all() -> None:
         disabled += ["libimagequant"]
     if args.no_fribidi:
         disabled += ["fribidi"]
+    elif args.architecture == "ARM64" and platform.machine() != "ARM64":
+        print(
+            "\nWarning: Cross-compiling FriBiDi is currently not supported, disabling"
+        )
```
radarhere, line 708: You didn't want to use `warnings.warn()`?

nulano: I don't like that it can sometimes appear at a random point in the output due to using stderr instead of stdout so I hadn't even tested it. I've just tried it and found that:
* in Terminal, it seems to always print in the intended order;
* in PyCharm, it can print out of order, but uses a red highlight.

I've therefore changed the code to use `warnings.warn`.

winbuild/build_prepare.py
```diff
@@ -367,12 +375,17 @@ def cmd_msbuild(
 
 
 # based on distutils._msvccompiler from CPython 3.7.4
-def find_msvs() -> dict[str, str] | None:
+def find_msvs(architecture) -> dict[str, str] | None:
```
hugovk, line 378:
```suggestion
def find_msvs(architecture: str) -> dict[str, str] | None:
```

hugovk (approved):

winbuild/build_prepare.py
```diff
@@ -118,6 +118,12 @@ def cmd_msbuild(
             "(LEGAL ISSUES\n============\n\n.+?)\n\nREFERENCES\n=========="
             ".+(libjpeg-turbo Licenses\n======================\n\n.+)$"
         ),
+        "patch": {
+            r"CMakeLists.txt": {
+                # libjpeg-turbo does not detect MSVC x86_arm64 cross-compiler correctly
+                'if(MSVC_IDE AND CMAKE_GENERATOR_PLATFORM MATCHES "arm64")': "if({architecture} STREQUAL ARM64)",  # noqa: E501
+            },
+        },
```
radarhere, line 126: Is this change something that could be suggested to https://github.com/libjpeg-turbo/libjpeg-turbo, so that it could one day no longer require patching?

nulano: No, not easily.
CMake does not provide a value for the target CPU architecture, only the host CPU architecture: https://stackoverflow.com/a/12024211/1648883
The `{architecture}` string here is referring to the Python value from https://github.com/python-pillow/Pillow/blob/af83f679e80d6bdcd3456b9add007fce2e0025ee/winbuild/build_prepare.py#L712

radarhere (approved):

PR #7564, opened 2023-11-20
Use --config-settings in external libraries list in installation.rst

nulano:
These two references to the build flags were missed in #7171.


hugovk (approved):
Thanks!

radarhere:
I feel a bit weird about

> Starting with Pillow 3.0.0, libjpeg is required by default, but may be disabled with the ``-C jpeg=disable`` flag.

since "libjpeg is required by default" is since Pillow 3.0.0, but ``-C jpeg=disable`` isn't until much later. My suggestion is to split this into two sentences? https://github.com/nulano/Pillow/pull/24

radarhere (approved):

PR #7566, opened 2023-11-24
Do not assign new fp attribute when exiting context manager

radarhere:
At the moment,
```python
from PIL import Image
with Image.new("L", (1, 1)) as im:
  print(hasattr(im, "fp"))
print(hasattr(im, "fp"))
```
gives
```
False
True
```
`im` does not have an `fp` attribute inside the context manager, but once it leaves, it is set.

This PR keeps it unset.

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -527,15 +527,18 @@ class Image:
     def __enter__(self):
         return self
 
+    def _close_fp(self):
+        if getattr(self, "_fp", False):
+            if self._fp != self.fp:
+                self._fp.close()
+            self._fp = DeferredError(ValueError("Operation on closed image"))
+        if self.fp:
+            self.fp.close()
+
     def __exit__(self, *args):
         if hasattr(self, "fp"):
             if getattr(self, "_exclusive_fp", False):
-                if getattr(self, "_fp", False):
-                    if self._fp != self.fp:
-                        self._fp.close()
-                    self._fp = DeferredError(ValueError("Operation on closed image"))
-                if self.fp:
-                    self.fp.close()
+                self._close_fp()
             self.fp = None
 
     def close(self):
@@ -552,12 +555,7 @@ class Image:
         """
         if hasattr(self, "fp"):
             try:
-                if getattr(self, "_fp", False):
-                    if self._fp != self.fp:
-                        self._fp.close()
-                    self._fp = DeferredError(ValueError("Operation on closed image"))
-                if self.fp:
-                    self.fp.close()
+                self._close_fp()
                 self.fp = None
             except Exception as msg:
                 logger.debug("Error closing: %s", msg)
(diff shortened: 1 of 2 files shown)


src/PIL/Image.py
```diff
@@ -528,14 +528,15 @@ def __enter__(self):
...
-        self.fp = None
+        if hasattr(self, "fp"):
+            if getattr(self, "_exclusive_fp", False):
+                if getattr(self, "_fp", False):
+                    if self._fp != self.fp:
+                        self._fp.close()
+                    self._fp = DeferredError(ValueError("Operation on closed image"))
+                if self.fp:
+                    self.fp.close()
+            self.fp = None
```
akx, line 539: Might be easier to read if the outermost layer of the pyramid was negated?
```suggestion
        if not hasattr(self, "fp"):
            return
        if getattr(self, "_exclusive_fp", False):
            if getattr(self, "_fp", False):
                if self._fp != self.fp:
                    self._fp.close()
                self._fp = DeferredError(ValueError("Operation on closed image"))
            if self.fp:
                self.fp.close()
        self.fp = None
```

radarhere: Easier to read in isolation, yes, but I kind of like the fact that this version mirrors `close()` - https://github.com/python-pillow/Pillow/blob/5431b15bd2f8177b762332ac632b0342c3b9d506/src/PIL/Image.py#L530-L565

akx: Should that code be extracted from `close()` then and reused in both of these places? `_close_fp()`?

radarhere: Ok, I've pushed a commit.

src/PIL/Image.py
```diff
@@ -527,15 +527,19 @@ def _new(self, im):
...
-        if hasattr(self, "fp") and getattr(self, "_exclusive_fp", False):
-            if getattr(self, "_fp", False):
-                if self._fp != self.fp:
-                    self._fp.close()
-                self._fp = DeferredError(ValueError("Operation on closed image"))
-            if self.fp:
-                self.fp.close()
-        self.fp = None
+        if hasattr(self, "fp"):
+            if getattr(self, "_exclusive_fp", False):
```
akx, line 540: Out of curiosity: does `exclusive_fp` mean, like I think it does, "we own this fp because we opened it"? If so, does it make sense for `.close()` to also close FPs that aren't "owned" by this instance?

radarhere: > Out of curiosity: does exclusive_fp mean, like I think it does, "we own this fp because we opened it"?

Yes

https://github.com/python-pillow/Pillow/blob/697c24bc7dee8a679ae18d6267cc333f79a32dca/src/PIL/Image.py#L3250-L3258

> If so, does it make sense for .close() to also close FPs that aren't "owned" by this instance?

See #5309

Issue #6832, opened 2022-12-27
Color and quality loss with similar colors or from RGBA


BootsManOut:
Hello,

I have found 2 cases where PIL reduces color quality, when I import a gif animation as PIL images, and then save it again as a gif animation.

**Case 1: Importing a gif animation with similar background colors removes differentiated background:**

This is the gif animation with a checkered background:
![GIF before import](https://user-images.githubusercontent.com/63544082/209736522-7825a198-c3b1-4898-bd41-15a694f16abf.gif)

After importing it as PIL format images, and then exporting it again as a gif, the colors get reduced, and you cannot see the checkered pattern of the background anymore:
![GIF after export](https://user-images.githubusercontent.com/63544082/209736462-a686bb2c-b99d-480a-9798-13e204083281.gif)

The code used is this:
```python
from PIL import Image as Img

#Convert Gif Frames into PIL images:
tempimage = Img.open("GIF before import.gif")
ExportFrames = []
durationFrame=None
for i in range(0, tempimage.n_frames):
    tempimage.seek(i)
    ExportFrames.append(tempimage.copy())

    if durationFrame == None:
        durationFrame = tempimage.info['duration']

ExportFrames[0].save("GIF after export.gif", disposal=2, save_all=True,
                                     append_images=ExportFrames[1:], loop=0,
                                     duration=durationFrame, optimize=False, lossless=True)
```

      
        
                
**Case 2: Adding Alpha Channel to PIL images, before saving as gif:**

I added this additional step of pasting each frame unto a transparent alpha PIL image, because for some GIFs, when I imported them and then saved them, the first main frame of the GIF would not have a transparent background. This would be fixed by adding this step.
However with it this quality loss occurs:
The original GIF image that I import as PIL formatted images:
![GIF before import](https://user-images.githubusercontent.com/63544082/209737429-86a14511-1770-4c4c-b507-dfebf9f65b27.gif)
After saving, there is an extreme quality loss:
![GIF after export](https://user-images.githubusercontent.com/63544082/209737463-d7b4ab32-2078-4c85-a120-eabdc027a3a6.gif)

This quality loss does only occur when adding the step.
However, even with the step added, there is no quality loss when exporting the frames as single png images instead as a gif animation:

![GIF before import 44](https://user-images.githubusercontent.com/63544082/209737594-361d9639-8c10-4872-b48d-aaa4acdcbb33.png)

This suggests that error happens during GIF export.

The code used for this is the following:

```python
[...]

Issue #617, opened 2014-04-12
Filesize concerns


jvanasco:
I've noticed that the images created by PIL are quite large.  

gifs seem to be over 4x larger than they could be.  
pngs are around 130% larger

i tested this just off of pillow created files vs the same ones imported/exported through a graphics app.  ie, i used lossless compression on the second time.   

as a quick example, i drew some text on a canvas.  a (256 color) gif from pillow was around 12k; the same gif was opened & exported to 2k using an app.  pillow generated it as 3 color gif at 3k (larger filesize than the much better looking 256 color), while it was a few hundred bytes via an app.

there are a handful of advanced image 'compressing' apps out there too, which do lossless optimizations.  i didn't run the images through any of them.

i'm wondering if anyone here knows any techniques or ways to address this.  i thought this might have something to do with the old gif patents, but they're long expired and pngs seem to be affected.

hugovk:
From https://github.com/python-pillow/Pillow/issues/2480 (closed as duplicate of this):

---

### What did you do?

Reverse a gif get it much bigger size on disk

### What did you expect to happen?

Still the original size or only a bit more

### What versions of Pillow and Python are you using?

latest Pillow and 3.6 Python

```python
from PIL import Image, ImageSequence
from PIL import ImagePalette
with Image.open('sd.gif') as im:
    if im.is_animated:
        frames = [f.copy() for f in ImageSequence.Iterator(im)]
        frames.reverse()
        frames[0].save('out.gif', save_all=True, append_images = frames[1:])  
```

jvanasco:
I'm not familiar with the exact inner workings, but #2480 is probably not the same as this (or an issue at all).

I should probably address some of the earlier comments in here too.

This ticket shouldn't have to do with quantization (which is lossy) but is meant to address bloat issues when doing a lossless encoding as gif.  Perhaps Pillow is running the gif through the quantizer when it doesn't have to, and could just ignore it if the number of colors is small enough?

In terms of #2480, animated gifs are built off of deltas -- not individual frames.  The bloat could also be created by:

* the original gif uses diffs, the Pillow created gif uses full frames
* reversing the image drastically changes the filesize of a few frames. 
* other things  (It's the end of the day!)

wiredfool:
Actually, it's simpler than that. Pillow doesn't do compression on GIFs. It only implements the uncompressed frame format. So anything that's going to do GIF compression is going to beat it. There was code for that, but it was in the closed source PILplus extension. 

[...]

PR #7568, opened 2023-11-24
Fill identical pixels with transparency in subsequent frames when saving GIF

radarhere:
When saving multiple GIF frames, Pillow trims identical edges of each new frame. This PR goes further, setting the pixels that are not new in the next frame as transparent if Pillow is optimizing. If there is no transparency, try and allocate one.

I started work on this after seeing https://github.com/python-pillow/Pillow/issues/6832#issuecomment-1821931777. If I remove `optimize=False` from that code, then without this change the file size is 21.5mb. With this change, the file size is 8.1mb.

For a less dramatic example, the output of https://github.com/python-pillow/Pillow/issues/617#issuecomment-1188984138 is reduced from 408kb to 345kb

Because the only difference in output should be filesize, there are no new tests to demonstrate correctness. I have added a test to cover an exception caught if trying to add transparency when there is no room in the frame.

### DIFF
--- a/Tests/test_file_gif.py
+++ b/Tests/test_file_gif.py
@@ -217,6 +217,27 @@ def test_optimize_if_palette_can_be_reduced_by_half():
             assert len(reloaded.palette.palette) // 3 == colors
 
 
+def test_full_palette_second_frame(tmp_path):
+    out = str(tmp_path / "temp.gif")
+    im = Image.new("P", (1, 256))
+
+    full_palette_im = Image.new("P", (1, 256))
+    for i in range(256):
+        full_palette_im.putpixel((0, i), i)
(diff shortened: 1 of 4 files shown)


alanhamlett:
Could this be causing the `ValueError` in #7777?

radarhere:
Yes, this would be the cause - but rather than just undoing this change, it would be better to adjust it, to both reduce filesize and to fix your problem. The quickest way to understanding how to do that would be for you to post a self-contained example of your problem in that issue.

radarhere:
https://github.com/python-pillow/Pillow/issues/7777 was resolved by https://github.com/python-pillow/Pillow/pull/7779

PR #7574, opened 2023-11-27
Added reading DDS BGR;15 images

### LINKED ISSUE #7517: PIL.UnidentifiedImageError: cannot identify image file (DDS file)
Hello! i try to make .DDS to .PNG converter but when i call code it raises PIL.UnidentifiedImageError: cannot identify image file
```python
# Imports
from PIL import Image
import os

# Define names
file = 'C:\DDS\Portrait_Britain_Alan_Brooke.dds'
l = len(os.listdir("SOV"))+1
name = f'SOV/{l}.png'

# Load image
with Image.open(file) as im: # in this line error
    im.save(name)
```
screenshot of console:
error
this file:
[Portrait_Britain_Alan_Brooke.zip](https://github.com/python-pillow/Pillow/files/13239297/Portrait_Britain_Alan_Brooke.zip)

radarhere:
Resolves #7517

The issue found a DDS image with `masks` that mean the image is BGR;15.

I've created a test image using a modified Pillow.

### DIFF
--- a/src/PIL/DdsImagePlugin.py
+++ b/src/PIL/DdsImagePlugin.py
@@ -137,7 +137,6 @@ class DdsImageFile(ImageFile.ImageFile):
         pfsize, pfflags = struct.unpack("<2I", header.read(8))
         fourcc = header.read(4)
         (bitcount,) = struct.unpack("<I", header.read(4))
-        masks = struct.unpack("<4I", header.read(16))
         if pfflags & DDPF_LUMINANCE:
             # Texture contains uncompressed L or LA data
             if pfflags & DDPF_ALPHAPIXELS:
@@ -150,12 +149,33 @@ class DdsImageFile(ImageFile.ImageFile):
             # Texture contains uncompressed RGB data
             if not pfflags & DDPF_ALPHAPIXELS:
                 self._mode = "RGB"
-            if masks[:3] == (31744, 992, 31) and self.mode == "RGB":
-                rawmode = "BGR;15"
+
+            # Each channel is derived using a mask of the pixel values
+            masks = {
+                mask: "RGB"[i]
+                for i, mask in enumerate(struct.unpack("<3I", header.read(12)))
+            }
+            if self.mode == "RGB" and all(
+                mask in (0b0000000000011111, 0b0000001111100000, 0b0111110000000000)
+                for mask in masks
+            ):
+                # Each mask uses 5 bits
+                rawmode = (
+                    masks[0b0000000000011111]
+                    + masks[0b0000001111100000]
+                    + masks[0b0111110000000000]
+                    + ";15"
+                )
             else:
-                masks = {mask: ["R", "G", "B", "A"][i] for i, mask in enumerate(masks)}
-                rawmode = masks[0xFF] + masks[0xFF00] + masks[0xFF0000]
+                # Try 8 bits for each mask
                 if self.mode == "RGBA":
+                    # The fourth mask, alpha,
+                    # is only valid when the correct flags are present
+                    a_mask = struct.unpack("<I", header.read(4))[0]
+                    masks[a_mask] = "A"
+
+                rawmode = masks[0xFF] + masks[0xFF00] + masks[0xFF0000]
+                if 0xFF000000 in masks:
                     rawmode += masks[0xFF000000]
 
             self.tile = [("raw", (0, 0) + self.size, 0, (rawmode, 0, 1))]


src/PIL/DdsImagePlugin.py
```diff
@@ -148,15 +148,17 @@ def _open(self):
...
             # Texture contains uncompressed RGB data
-            masks = {mask: ["R", "G", "B", "A"][i] for i, mask in enumerate(masks)}
-            rawmode = ""
-            if pfflags & DDPF_ALPHAPIXELS:
-                rawmode += masks[0xFF000000]
-            else:
+            if not pfflags & DDPF_ALPHAPIXELS:
                 self._mode = "RGB"
-            rawmode += masks[0xFF0000] + masks[0xFF00] + masks[0xFF]
+            if masks[:3] == (31744, 992, 31) and self.mode == "RGB":
```
akx, line 153: The magic numbers here could use some explanation if you ask me.

src/PIL/DdsImagePlugin.py
```diff
@@ -148,15 +148,17 @@ def _open(self):
...
+                rawmode = "BGR;15"
+            else:
+                masks = {mask: ["R", "G", "B", "A"][i] for i, mask in enumerate(masks)}
```
akx, line 156:
```suggestion
                masks = {mask: "RGBA"[i] for i, mask in enumerate(masks)}
```
would be more efficient, since the (possibly-mutable-but-the-interpreter-can't-tell) list doesn't need to regenerated for every call of this function, but it could use the same `"RGBA"` string constant used just below.

src/PIL/DdsImagePlugin.py
radarhere: I know what you're said sounds like a perfectly straightforward request, but I can't think of a short way to explain it that isn't just explaining what masks are, or what BGR;15 is.

There are 4 bit masks - https://learn.microsoft.com/en-us/windows/win32/direct3ddds/dds-pixelformat - R, G, B and A.

Each mask filters the pixel values to transform them into the value for each channel.

These masks are filtering 5 bits for R, 5 bits for G and 5 bits for A - BGR;15.
```
31744 = 0b0111110000000000
  992 = 0b0000001111100000
   31 = 0b0000000000011111
```

akx: That looks like great material for a code comment right there :)

You could also replace the decimal constants with those binary constants – then it's immediately more clear for a reader.

src/PIL/DdsImagePlugin.py
radarhere: I've added a version of this.

src/PIL/DdsImagePlugin.py
radarhere: Clearer, yes, it just feels verbose... I've pushed a commit.

src/PIL/DdsImagePlugin.py
```diff
@@ -148,15 +147,38 @@ def _open(self):
...
-            else:
+            if not pfflags & DDPF_ALPHAPIXELS:
                 self._mode = "RGB"
-            rawmode += masks[0xFF0000] + masks[0xFF00] + masks[0xFF]
 
-            self.tile = [("raw", (0, 0) + self.size, 0, (rawmode[::-1], 0, 1))]
+            # Each channel is derived using a mask of the pixel values
+            masks = {
+                mask: "RGB"[i]
+                for i, mask in enumerate(struct.unpack("<3I", header.read(12)))
```
akx, line 156: The previous version always read 16 bytes for masks – this will now read either 12 or 16 bytes. That seems odd?

src/PIL/DdsImagePlugin.py
```diff
@@ -148,15 +147,38 @@ def _open(self):
...
+                mask in (0b0000000000011111, 0b0000001111100000, 0b0111110000000000)
+                for mask in masks
+            ):
+                # Each mask uses 5 bits
+                rawmode = (
+                    masks[0b0000000000011111]
+                    + masks[0b0000001111100000]
+                    + masks[0b0111110000000000]
+                    + ";15"
+                )
```
akx, line 168: Will this now support esoteric formats like `GRB;15` (and if so, will everything be handled correctly down the line)? If not, then I think just "RGB;15" would be better?

src/PIL/DdsImagePlugin.py
```diff
@@ -148,15 +147,38 @@ def _open(self):
...
-            rawmode += masks[0xFF0000] + masks[0xFF00] + masks[0xFF]
 
-            self.tile = [("raw", (0, 0) + self.size, 0, (rawmode[::-1], 0, 1))]
+            # Each channel is derived using a mask of the pixel values
+            masks = {
+                mask: "RGB"[i]
+                for i, mask in enumerate(struct.unpack("<3I", header.read(12)))
+            }
+            if self.mode == "RGB" and all(
+                mask in (0b0000000000011111, 0b0000001111100000, 0b0111110000000000)
```
akx, line 159: This did become less clear to read, you're right. Wonder if these and the `0xFF0000` masks from below could benefit from becoming constants á la (I didn't double check whether I have RGB/BGR right here):

```
R_MASK_555 = 0b0000000000011111
G_MASK_555 = 0b000001111100000
B_MASK_555 = 0b111110000000000
R_MASK_888 = 0x000000FF
G_MASK_888 = 0x0000FF00
B_MASK_888 = 0x00FF0000
A_MASK_888 = 0xFF000000
```

src/PIL/DdsImagePlugin.py
```diff
@@ -148,15 +147,38 @@ def _open(self):
...
+                    masks[0b0000000000011111]
+                    + masks[0b0000001111100000]
+                    + masks[0b0111110000000000]
+                    + ";15"
+                )
+            else:
+                # Try 8 bits for each mask
+                if self.mode == "RGBA":
+                    # The fourth mask, alpha,
+                    # is only valid when the correct flags are present
```
akx, line 173: What are the correct flags?

src/PIL/DdsImagePlugin.py
```diff
@@ -148,15 +147,38 @@ def _open(self):
...
+            ):
+                # Each mask uses 5 bits
+                rawmode = (
+                    masks[0b0000000000011111]
+                    + masks[0b0000001111100000]
+                    + masks[0b0111110000000000]
+                    + ";15"
+                )
+            else:
+                # Try 8 bits for each mask
```
akx, line 170: What's the trying here? Rather, what happens if trying fails?

src/PIL/DdsImagePlugin.py
radarhere: It isn't always necessary to read the last 4 bytes, and therefore seems more efficient to not do so where possible.

If you are concerned about the file pointer position, that's not relevant, as it is just reading from
https://github.com/python-pillow/Pillow/blob/697c24bc7dee8a679ae18d6267cc333f79a32dca/src/PIL/DdsImagePlugin.py#L127

src/PIL/DdsImagePlugin.py
radarhere: This will support RGB;15 and BGR;15
https://github.com/python-pillow/Pillow/blob/697c24bc7dee8a679ae18d6267cc333f79a32dca/src/libImaging/Unpack.c#L1576-L1577

src/PIL/DdsImagePlugin.py
akx: I think it would actually be more efficient and clearer to _not_ do another read-and-struct-unpack? (Would also mean that the entire [`DDS_PIXELFORMAT`](https://learn.microsoft.com/en-us/windows/win32/direct3ddds/dds-pixelformat) is always consumed.)

src/PIL/DdsImagePlugin.py
radarhere: Given that the issue related to this issue has found 5 bit masks, it seems possible that there could be other mask values used in files, beyond 5 bit and 8 bits.

If it fails, an error is raised, and ultimately an `UnidentifiedImageError`.

src/PIL/DdsImagePlugin.py
radarhere: https://learn.microsoft.com/en-us/windows/win32/direct3ddds/dds-pixelformat seems inconsistent on this to me

> DDPF_ALPHAPIXELS | Texture contains alpha data; dwRGBAlphaBitMask contains valid data.

> Alpha mask for reading alpha data. dwFlags must include DDPF_ALPHAPIXELS or DDPF_ALPHA

The code is checking for `DDPF_ALPHAPIXELS`
https://github.com/python-pillow/Pillow/blob/aa03c5a82a0213260517ea4d1ed8925babea42ca/src/PIL/DdsImagePlugin.py#L150

src/PIL/DdsImagePlugin.py
akx, line 170:
```suggestion
                # Assume 8 bits for each mask
```

src/PIL/DdsImagePlugin.py
akx, line 173:
```suggestion
                    # is only valid when `pfflags & DDPF_ALPHAPIXELS`
```

akx (commented):
A couple comment suggestions based on the discussion, to make this clearer for future Pillow archaeologists without having to look at PRs :)

PR #7576, opened 2023-11-27
Import plugins relative to the module

### LINKED ISSUE #7575: PIL.UnidentifiedImageError: cannot identify image file (Custom installation in other directories)
### What did you do?

```bash
pip install -h |grep target
#  -t, --target <dir>        Install packages into <dir>. By default this
#                            multiple platforms supported by the target
#                            abis supported by the target interpreter.

mkdir -p packages
# remove the Pillow, which installed into system packages
pip uninstall Pillow
pip install -t packages Pillow
```

It imports plugins correctly 

### What actually happened?

It can't import plugins, when I didn't install Pillow into system packages directories

* OS:  Debian
* Python: 3.8.0
* Pillow: All versions

The real format of the image is webp, you need to import PIL.WebPImagePlugin. The image is as follows:

3ae6c099102fba7a4773c3568edee0be

```python
from packages.PIL import Image

with open('3ae6c099102fba7a4773c3568edee0be.jpg', 'rb') as fr:
    Image.open(fr)
```

```pytb
  File "test2.py", line 4, in <module>
    Image.open(fr)
  File "/app/packages/PIL/Image.py", line 3023, in open
    raise UnidentifiedImageError(
packages.PIL.UnidentifiedImageError: cannot identify image file <_io.BufferedReader name='3ae6c099102fba7a4773c3568edee0be.jpg'>
```

deliangyang:
Fixes #7575

 *  Importing plugins from a user-specified installation directory of PIL

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -355,10 +355,11 @@ def init():
     if _initialized >= 2:
         return 0
 
+    parent_module, _, _ = __name__.rpartition(".")
     for plugin in _plugins:
         try:
             logger.debug("Importing %s", plugin)
-            __import__(f"PIL.{plugin}", globals(), locals(), [])
+            __import__(f"{parent_module}.{plugin}", globals(), locals(), [])
         except ImportError as e:
             logger.debug("Image: failed to import %s: %s", plugin, e)
 
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -355,11 +355,11 @@ def init():
     if _initialized >= 2:
         return 0
 
-    parent_module, _, _ = __name__.rpartition(".")
+    parent_name = __name__.rpartition(".")[0]
     for plugin in _plugins:
         try:
             logger.debug("Importing %s", plugin)
-            __import__(f"{parent_module}.{plugin}", globals(), locals(), [])
+            __import__(f"{parent_name}.{plugin}", globals(), locals(), [])
         except ImportError as e:
             logger.debug("Image: failed to import %s: %s", plugin, e)
 


deliangyang:
> What do you mean by "user-specified" directory? Looks like relative path 

Yes.

src/PIL/Image.py
```diff
@@ -355,10 +355,11 @@ def init():
     if _initialized >= 2:
         return 0
 
+    parrent_module = __name__.replace(".Image", "")
```
homm, line 358: This will break paths like "Project.ImageLibs.PIL.Image".

```suggestion
    parrent_module, _, _ = __name__.rpartition('.')
```

deliangyang: OK, thanks, I'll fix it

homm (approved):

src/PIL/Image.py
```diff
@@ -355,10 +355,11 @@ def init():
     if _initialized >= 2:
         return 0
 
+    parrent_module, _, _ = __name__.rpartition(".")
```
radarhere, line 358:
```suggestion
    parent_module, _, _ = __name__.rpartition(".")
```

src/PIL/Image.py
```diff
@@ -355,10 +355,11 @@ def init():
...
     for plugin in _plugins:
         try:
             logger.debug("Importing %s", plugin)
-            __import__(f"PIL.{plugin}", globals(), locals(), [])
+            __import__(f"{parrent_module}.{plugin}", globals(), locals(), [])
```
radarhere, line 362:
```suggestion
            __import__(f"{parent_module}.{plugin}", globals(), locals(), [])
```

src/PIL/Image.py
deliangyang: Thanks. I fixed it.

src/PIL/Image.py
```diff
@@ -355,10 +355,11 @@ def init():
     if _initialized >= 2:
         return 0
 
+    parent_module, _, _ = __name__.rpartition(".")
```
radarhere, line 358:
```suggestion
    parent_name = __name__.rpartition(".")[0]
```
Actually, looking at https://docs.python.org/3/library/importlib.html#approximating-importlib-import-module, I wonder if `parent_name` would be better?

src/PIL/Image.py
```diff
@@ -355,10 +355,11 @@ def init():
...
     for plugin in _plugins:
         try:
             logger.debug("Importing %s", plugin)
-            __import__(f"PIL.{plugin}", globals(), locals(), [])
+            __import__(f"{parent_module}.{plugin}", globals(), locals(), [])
```
radarhere, line 362:
```suggestion
            __import__(f"{parent_name}.{plugin}", globals(), locals(), [])
```

src/PIL/Image.py
deliangyang: Good idea. I've replaced it

Issue #258, opened 2013-06-29
Bytes filenames should be accepted with Python 3


Arfrever:
Bytes filenames should be accepted with Python 3. They are accepted with Python 2. An application might want to use bytes filenames to avoid problems with undecodable characters.

```
$ python2.7 -c 'import PIL.Image; PIL.Image.open(b"file.png")'
$ python3.3 -c 'import PIL.Image; PIL.Image.open(b"file.png")'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/lib64/python3.3/site-packages/PIL/Image.py", line 1993, in open
    prefix = fp.read(16)
AttributeError: 'bytes' object has no attribute 'read'
```

Also there is NameError in line 491 of PIL/JpegImagePlugin.py with Python 3.

PR #7578, opened 2023-11-27
Handle pathlib.Path in FreeTypeFont

### LINKED ISSUE #7577: ImageFont.truetype() function does not accept `Path` objects
### What did you do?

Pass `Path` object instead of `str` to `ImageFont.truetype()`. IMO this should work if I understand the docs correctly.

```python
from PIL import ImageFont
from pathlib import Path

font_path = Path("path/to/your/font")
font = ImageFont.truetype(font_path, 12)
```

Get back a font object.

### What actually happened?

I get a type error saying that I cannot pass a `Path` object.

```
  File "/home/mauro/.pyenv/versions/test_env/lib/python3.10/site-packages/PIL/ImageFont.py", line 797, in truetype
    return freetype(font)
  File "/home/mauro/.pyenv/versions/test_env/lib/python3.10/site-packages/PIL/ImageFont.py", line 794, in freetype
  File "/home/mauro/.pyenv/versions/test_env/lib/python3.10/site-packages/PIL/ImageFont.py", line 226, in __init__
    self.font = core.getfont(
TypeError: argument 1 must be str, bytes or bytearray, not PosixPath
```

Funnily enough, the `__init__` function of `FreeTypeFont` checks if the passed `font` argument is of type `bytes`, `str` or `Path`: https://github.com/python-pillow/Pillow/blob/f9c7bd86115090ce2b110794485f138f0d5d5953/src/PIL/ImageFont.py#L215
The problem occurs later when the `core` object should load the font which presumably only works if `font` is a string:
https://github.com/python-pillow/Pillow/blob/f9c7bd86115090ce2b110794485f138f0d5d5953/src/PIL/ImageFont.py#L226-L228

This makes me think, that the `ImageFont.truetype()` function should indeed accept `Path` objects and internally convert the argument into a string before fowarding it to `core.getfont()`. 

* OS: Ubuntu 22.04.3 LTS
* Python: 3.10.7
* Pillow: 10.1.0

radarhere:
Resolves #7577


src/PIL/ImageFont.py
```diff
@@ -213,6 +214,8 @@ def load_from_bytes(f):
             )
 
         if is_path(font):
+            if isinstance(font, Path):
+                font = str(font)
```
hugovk, line 218: It's more efficient to always convert it:
```suggestion
            font = str(font)
```

string: 41ns -> 30ns
Path: 72ns -> 59ns

# Test script

```python
from pathlib import Path

font_str = "my_font.ttf"
font_path = Path("my_font.ttf")


def a(font):
    if isinstance(font, Path):
        font = str(font)

def b(font):
    font = str(font)
```

# Before

```console
❯ python --version
Python 3.12.0
❯ python -m timeit -s "import bench" "bench.a(bench.font_str)"
5000000 loops, best of 5: 40.7 nsec per loop

Pillow on  disable-gil [$?] via  v3.12.0
❯ python -m timeit -s "import bench" "bench.a(bench.font_path)"
5000000 loops, best of 5: 72.4 nsec per loop
```

# After

```console
❯ python -m timeit -s "import bench" "bench.b(bench.font_str)"
10000000 loops, best of 5: 30.3 nsec per loop

Pillow on  disable-gil [$?] via  v3.12.0 took 2s
❯ python -m timeit -s "import bench" "bench.b(bench.font_path)"
5000000 loops, best of 5: 59.2 nsec per loop
```

src/PIL/ImageFont.py
```diff
@@ -213,6 +214,8 @@ def load_from_bytes(f):
             )
 
         if is_path(font):
```
homm, line 216: I'm looking at `_util.is_path`, it checks `isinstance(f, (bytes, str, Path))`, so bytes are also handled as paths, but next I see: `else: load_from_bytes(font)`.

Do we really want consider bytes as paths? I believe this is obsolete in Python3

src/PIL/ImageFont.py
radarhere: `font` may also be bytes, in which case `str(font)` would be incorrect.

src/PIL/ImageFont.py
radarhere: #258 was the original request for this functionality.

src/PIL/ImageFont.py
hugovk: Fair enough, let's keep the `isinstance`.

Shall we start gradually adding type hints? (re: https://github.com/python-pillow/Pillow/issues/2625)

Something like this:

```python
from __future__ import annotations  # [this is so we can use | instead of Union in 3.9]

...

font: bytes | str | Path | None = None
```

radarhere: Ok, I've pushed a commit.

src/PIL/ImageFont.py
```diff
@@ -186,7 +188,14 @@ def getlength(self, text, *args, **kwargs):
...
     """FreeType font wrapper (requires _imagingft service)"""
 
-    def __init__(self, font=None, size=10, index=0, encoding="", layout_engine=None):
+    def __init__(
+        self,
+        font: bytes | str | Path | None = None,
+        size: float = 10,
+        index: int = 0,
+        encoding: str = "",
+        layout_engine: int = None,
```
hugovk, line 197: To account for the default:
```suggestion
        layout_engine: int | None = None,
```

hugovk (approved):

src/PIL/ImageFont.py
```diff
@@ -185,7 +188,14 @@ def getlength(self, text, *args, **kwargs):
...
     """FreeType font wrapper (requires _imagingft service)"""
 
-    def __init__(self, font=None, size=10, index=0, encoding="", layout_engine=None):
+    def __init__(
+        self,
+        font: bytes | str | Path | None = None,
+        size: float = 10,
+        index: int = 0,
+        encoding: str = "",
+        layout_engine: int | None = None,
```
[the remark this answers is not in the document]
radarhere: Testing `Layout | None`, I find that `layout_engine=ImageFont.Layout.RAQM` is fine, but `layout_engine=1` gives a warning in PyCharm.
pycharm

Is that helpful or unhelpful?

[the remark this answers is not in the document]
hugovk: I think that's helpful -- if people are adding types to their code, they should use `Layout` enums rather than `int`s.

And typeshed also has `Layout | None` here:

https://github.com/python/typeshed/blob/9fbf2e1c0cecb6d4c855fa283e1f3c6030006485/stubs/Pillow/PIL/ImageFont.pyi#L90

src/PIL/ImageFont.py
homm: I assume `load_from_bytes` is unreachable here, so we need to remove it, or provide a way to reach it

radarhere: I think it's reachable. The logic hasn't changed since `load_from_bytes()` was added in #3785.

If it is a path, then pass the path to the C layer. If we're on Windows and the path is not ASCII, then open a file pointer to the path and read the contents of the file pointer with `load_from_bytes()`.
If it is not a path, then it is a file pointer, read the contents of the file pointer with `load_from_bytes()`.

homm: Ok, I see. I believe there is misnaming here, since f is not a bytes, it's file object.

radarhere: Yep, I've pushed a commit. I presume from the naming that the type hint also includes `io.BytesIO`

nulano: @radarhere I see you've used `typing.IO`. This causes a mypy error, see #7642.

Testing with both PyCharm and mypy, I find that a `io.BytesIO` value is accepted for a `typing.BinaryIO` type hint, so I'm not sure I understand your last message.

radarhere: I think I was concerned about strings being used, but looking now, that doesn't appear to be a problem.

PR #7579, opened 2023-11-27
Docs: Activate tabs based on browser's operating system

hugovk:
We're also using this on the [Python Developer’s Guide](https://devguide.python.org/).

### DIFF
--- a/docs/resources/js/activate_tab.js
+++ b/docs/resources/js/activate_tab.js
@@ -27,7 +27,7 @@ function activateTab(tabName) {
   const labels = document.querySelectorAll(".tab-label");
 
   labels.forEach((label) => {
-    if (label.textContent.includes(tabName)) {
+    if (label.textContent == tabName) {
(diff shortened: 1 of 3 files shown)


docs/resources/js/activate_tab.js
```diff
@@ -0,0 +1,42 @@
...
+
+  return "unknown";
+}
+
+function activateTab(tabName) {
+  // Find all label elements containing the specified tab name
+  const labels = document.querySelectorAll(".tab-label");
+
+  labels.forEach((label) => {
+    if (label.textContent.includes(tabName)) {
```
nulano, line 30:
```suggestion
    if (label.textContent == tabName) {
```

nulano (commented):
This is selecting the `Windows using MSYS2/MinGW` tab rather than the expected `Windows` tab.

radarhere (approved):

Issue #7443, opened 2023-10-05
win32 wheels missing on PyPI


bparzella:
Since version 10.0.0 there are no wheels for 32-bit version of Windows present on pypi. Version 9.5.0 still had 32-bit wheels.

[Pillow 9.5.0](https://pypi.org/project/Pillow/9.5.0/#files) vs [Pillow 10.0.1](https://pypi.org/project/Pillow/10.0.1/#files) (search for win32)

We still use 32-bit Python on Windows (64-bit), as some old COM components (32-bit) require the use of the matching Python. They can't be called from a 64-bit Python.

homm:
Version 9.5.0 still had 32-bit wheels

bparzella:
So this is intended?

In contrast to linux, the 32-bit version of Python is still used on Windows due to compatibility. And Python still provides 32-bit Windows releases, even for 3.12.0 there is one: https://www.python.org/downloads/release/python-3120/

hugovk:
Yes, this is intended, see the [10.0.0 release notes](https://pillow.readthedocs.io/en/stable/releasenotes/10.0.0.html#bit-wheels) and https://github.com/python-pillow/Pillow/issues/6941.

radarhere:
The domino that [started](https://github.com/python-pillow/Pillow/issues/6941#issuecomment-1560930935) the dropping of 32-bit wheels was when [deadsnakes decided not to support Python 3.12 on Bionic](https://github.com/deadsnakes/python3.12/commit/b32841d16138850a953f89408d0c12186ef56cc6), the last [32-bit Ubuntu](https://askubuntu.com/questions/1376090/latest-version-of-ubuntu-for-i386-architecture-32-bit).

Pillow should still work on 32-bit, you just have to build it from source yourself. See https://pillow.readthedocs.io/en/stable/installation.html#building-from-source for more information.

bparzella:
While I completely understand the decision for Linux with below 0.1% of installs, Windows is quite different. More than 9000 installs in 7 days is a big number.

In the discussion mentioned in [issue 6941](https://github.com/python-pillow/Pillow/issues/6941#issuecomment-1561033773) a pyca/cryptography developer even notes the difference between Linux and Windows: https://discuss.python.org/t/dropping-32-bit-packages/5476/17

[...]

PR #7580, opened 2023-11-27
Build Windows wheels using cibuildwheel

nulano:
For #7390:

* removed wheel building on every push to simplify `test-windows.yml` so `libimagequant` can be included there,
* moved `fribidi` artifact to the wheels workflow so that both x86 and x64 versions are provided,
* moved `CIBW_CONFIG_SETTINGS` to `pyproject.toml` and added `build-verbosity = 1` so that build warnings are visible,
* moved modules/codecs/features check from `wheels-test.sh` to `Tests\check_wheel.py` so that it can also be used on Windows,
* added a Windows build to `wheels.yml`, using `winbuild\build_prepare.py` to build dependencies same as before, building with `cibuildwheel`, and testing in Docker similarly to #6847.

I've included x86 also for https://github.com/python-pillow/Pillow/issues/7443#issuecomment-1756321996; there have been several issues opened about installing on win32. Perhaps it is not yet the time to stop providing these.

I've tested the arm64 wheels on an M2 MacBook in a Windows ARM VM using this batch script:
```batch
rem params: path\to\python.exe path\to\wheel.whl venv_dir_name

%1 -m pip install virtualenv
%1 -m virtualenv %3
cmd /c %3\Scripts\activate.bat ^&^& python -m pip install %2[tests]
powershell .github\workflows\wheels-test.ps1 %3 && echo Success!
```
Although the VM did have VC++ redistributable DLLs installed, I checked the list of imported dlls by the pyd files match between arm64 and amd64 with dumpbin.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -111,9 +111,6 @@ jobs:
           choco install nasm --no-progress
           echo "C:\Program Files\NASM" >> $env:GITHUB_PATH
 
-          choco install ghostscript --version=10.0.0.20230317 --no-progress
-          echo "C:\Program Files\gs\gs10.00.0\bin" >> $env:GITHUB_PATH
-
           # Install extra test images
           xcopy /S /Y Tests\test-images\* Tests\images
 
@@ -160,10 +157,25 @@ jobs:
             powershell C:\pillow\.github\workflows\wheels-test.ps1 %CD%\..\venv-test
         shell: cmd
 
-      - uses: actions/upload-artifact@v3
+      - name: Upload wheels
+        uses: actions/upload-artifact@v3
         with:
           name: dist
           path: ./wheelhouse/*.whl
+
+      - name: Prepare to upload FriBiDi
+        if: "matrix.arch != 'ARM64'"
+        run: |
+          mkdir fribidi\${{ matrix.arch }}
+          copy winbuild\build\bin\fribidi* fribidi\${{ matrix.arch }}
+        shell: cmd
+
+      - name: Upload fribidi.dll
+        if: "matrix.arch != 'ARM64'"
+        uses: actions/upload-artifact@v3
+        with:
+          name: fribidi
+          path: fribidi\*
 
   sdist:
     runs-on: ubuntu-latest
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -104,7 +104,7 @@ jobs:
 
       - uses: actions/setup-python@v4
         with:
-          python-version: "3.11"
+          python-version: "3.x"
 
       - name: Prepare for build
         run: |
@@ -147,14 +147,13 @@ jobs:
           CIBW_BEFORE_ALL: "{package}\\winbuild\\build\\build_dep_all.cmd"
           CIBW_CACHE_PATH: "C:\\cibw"
           CIBW_TEST_SKIP: "*-win_arm64"
-          CIBW_TEST_COMMAND: >-
-            docker run --rm
+          CIBW_TEST_COMMAND: 'docker run --rm
             -v {project}:C:\pillow
             -v C:\cibw:C:\cibw
             -v %CD%\..\venv-test:%CD%\..\venv-test
             -e CI -e GITHUB_ACTIONS
             mcr.microsoft.com/windows/servercore:ltsc2022
-            powershell C:\pillow\.github\workflows\wheels-test.ps1 %CD%\..\venv-test
+            powershell C:\pillow\.github\workflows\wheels-test.ps1 %CD%\..\venv-test'
         shell: cmd
 
       - name: Upload wheels
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -3,14 +3,20 @@ name: Wheels
 on:
   push:
     paths:
-      - ".github/workflows/wheels*.yml"
+      - ".ci/requirements-cibw.txt"
+      - ".github/workflows/wheel*"
       - "wheels/*"
+      - "winbuild/build_prepare.py"
+      - "winbuild/fribidi.cmake"
     tags:
       - "*"
   pull_request:
     paths:
-      - ".github/workflows/wheels*.yml"
+      - ".ci/requirements-cibw.txt"
+      - ".github/workflows/wheel*"
       - "wheels/*"
+      - "winbuild/build_prepare.py"
+      - "winbuild/fribidi.cmake"
   workflow_dispatch:
 
 permissions:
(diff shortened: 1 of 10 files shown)


Tests/helper.py
```diff
@@ -258,11 +259,21 @@ def hopper(mode=None, cache={}):
 
 
 def djpeg_available():
-    return bool(shutil.which("djpeg"))
+    if shutil.which("djpeg"):
+        try:
+            subprocess.check_call(["djpeg", "-version"])
+            return True
+        except subprocess.CalledProcessError:
+            return False
```
nulano, line 267: `djpeg.exe` and `cjpeg.exe` require VC++ redistributable packages, but the whole point of testing in Docker is to not install those. Running these during CPython tests therefore crashes with a message about missing DLLs.

.github/workflows/wheels.yml
```diff
@@ -71,6 +74,96 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+      - name: Checkout extra test images
+        uses: actions/checkout@v4
+        with:
+          repository: python-pillow/test-images
+          path: Tests\test-images
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.11"
```
radarhere, line 99:
```suggestion
          python-version: "3.x"
```

.github/workflows/wheels.yml
```diff
@@ -71,6 +74,96 @@ jobs:
...
+          CIBW_CACHE_PATH: "C:\\cibw"
+          CIBW_TEST_SKIP: "*-win_arm64"
+          CIBW_TEST_COMMAND: >-
+            docker run --rm
+            -v {project}:C:\pillow
+            -v C:\cibw:C:\cibw
+            -v %CD%\..\venv-test:%CD%\..\venv-test
+            -e CI -e GITHUB_ACTIONS
+            mcr.microsoft.com/windows/servercore:ltsc2022
+            powershell C:\pillow\.github\workflows\wheels-test.ps1 %CD%\..\venv-test
```
hugovk, line 147: Still work with some extra indent for readability?
```suggestion
            docker run --rm
              -v {project}:C:\pillow
              -v C:\cibw:C:\cibw
              -v %CD%\..\venv-test:%CD%\..\venv-test
              -e CI -e GITHUB_ACTIONS
              mcr.microsoft.com/windows/servercore:ltsc2022
              powershell C:\pillow\.github\workflows\wheels-test.ps1 %CD%\..\venv-test
```

hugovk (commented):
Thanks for this! A quick review.

.github/workflows/wheels.yml
```diff
@@ -71,6 +74,96 @@ jobs:
...
+          python-version: "3.11"
+      - name: Prepare for build
+        run: |
+          choco install nasm --no-progress
+          echo "C:\Program Files\NASM" >> $env:GITHUB_PATH
+
+          # Install extra test images
+          xcopy /S /Y Tests\test-images\* Tests\images
+
+          & python.exe -m pip install cibuildwheel==2.16.2
```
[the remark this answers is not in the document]
nulano: Unfortunately not, `call winbuild\\build\\build_env.cmd` needs to be run before (but in the same step as) `%pythonLocation%\python.exe -m cibuildwheel . --output-dir wheelhouse` to set up environment variables, since `CIBW_BEFORE_ALL` does not update them.

Tests/test_imagegrab.py
```diff
@@ -15,6 +15,8 @@ class TestImageGrab:
         sys.platform not in ("win32", "darwin"), reason="requires Windows or macOS"
     )
     def test_grab(self):
+        if os.environ.get("USERNAME") == "ContainerAdministrator":
```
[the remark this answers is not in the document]
nulano: I've added an explicit `#pragma: no cover` instead as that seems clearer to me.

.github/workflows/wheels.yml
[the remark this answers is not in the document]
hugovk: That's okay, we can put it in a requirements file and have Renovate update that instead: 

* https://github.com/python-pillow/Pillow/pull/7581

Tests/test_imagegrab.py
```diff
@@ -15,8 +15,9 @@ class TestImageGrab:
         sys.platform not in ("win32", "darwin"), reason="requires Windows or macOS"
     )
     def test_grab(self):
-        if os.environ.get("USERNAME") == "ContainerAdministrator":
+        if os.environ.get("USERNAME") == "ContainerAdministrator":  # pragma: no cover
```
hugovk, line 18: Or can we use a `@pytest.mark.skipif` decorator as well?

nulano: Oh right, yeah.

.github/workflows/wheels.yml
nulano: That adds extra newlines which [makes it fail](https://github.com/python-pillow/Pillow/actions/runs/7010925491/job/19072484245?pr=7580#step:6:3335).

Using a [single-quoted flow scalar](https://yaml-multiline.info#flow-scalars-single) seems to work fine:
```
          CIBW_TEST_COMMAND: 'docker run --rm
            -v {project}:C:\pillow
            -v C:\cibw:C:\cibw
            -v %CD%\..\venv-test:%CD%\..\venv-test
            -e CI -e GITHUB_ACTIONS
            mcr.microsoft.com/windows/servercore:ltsc2022
            powershell C:\pillow\.github\workflows\wheels-test.ps1 %CD%\..\venv-test'
```

hugovk (approved):

Tests/check_wheel.py
```diff
@@ -0,0 +1,38 @@
...
+import sys
+
+from PIL import features
+
+
+def test_wheel_modules():
+    expected_modules = {"pil", "tkinter", "freetype2", "littlecms2", "webp"}
+
+    # tkinter is not available in cibuildwheel installed CPython on Windows
+    if not importlib.util.find_spec("tkinter"):
```
radarhere, line 11: I expect there's a reason, but why is it not simpler to test for tkinter by attempting to import it normally?

nulano: Lint error, https://github.com/nulano/Pillow/actions/runs/6998411214/job/19036420674#step:7:21:
```
Tests/check_wheel.py:10:16: F401 `tkinter` imported but unused; consider using `importlib.util.find_spec` to test for availability
```

hugovk: Ah, `assert tkinter` should silence the warning. We've done that before.

nulano: Yes, I remember seeing that before, but figured the linter suggestion would work as well.

I've pushed a commit to `try: import tkinter` instead for consistency with other optional imports (the only one I could find with a simple grep was in `Image.preinit`).

.github/workflows/wheels.yml
```diff
@@ -75,6 +80,102 @@ jobs:
...
+      - name: Prepare to upload FriBiDi
+        if: "matrix.arch != 'ARM64'"
+        run: |
+          mkdir fribidi\${{ matrix.arch }}
+          copy winbuild\build\bin\fribidi* fribidi\${{ matrix.arch }}
+        shell: cmd
+
+      - name: Upload fribidi.dll
+        if: "matrix.arch != 'ARM64'"
+        uses: actions/upload-artifact@v3
```
radarhere, line 174: Previously, this was only run for one Python version.

https://github.com/python-pillow/Pillow/blob/63bec07b971cfe54311a2883206c020756081fa5/.github/workflows/test-windows.yml#L246-L248

Any particular reason it's now being run for every version?

nulano: This is now in the `wheels.yml` workflow which only runs a single job for all Python versions of a single architecture.
Previously it was in `test-windows.yml` which has a separate job for each Python version.

hugovk: Yeah, the others were replaced with `pytest.importorskip` (https://github.com/python-pillow/Pillow/pull/4436) which doesn't fit here.

Tests/check_wheel.py
hugovk: Yeah, the others were replaced with `pytest.importorskip` (https://github.com/python-pillow/Pillow/pull/4436) which doesn't fit here.

PR #7581, opened 2023-11-27
Install cibuildwheel from requirements file

hugovk:
Helps https://github.com/python-pillow/Pillow/issues/7390.

Switch back to the PyPI-installed cibuildwheel for GitHub Actions, also use it from Travis CI.

* https://cibuildwheel.readthedocs.io/en/stable/faq/#option-2-requirement-files

The upcoming Windows wheel build can't use the action, so Renovate can update all three at the same time.

* https://github.com/python-pillow/Pillow/pull/7580#discussion_r1406769153

### DIFF
--- a/.ci/requirements-cibw.txt
+++ b/.ci/requirements-cibw.txt
@@ -1,0 +1,1 @@
+cibuildwheel==2.16.2
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -52,10 +52,14 @@ jobs:
         with:
           submodules: true
 
-      - name: Build wheels
-        uses: pypa/cibuildwheel@v2.16.2
+      - uses: actions/setup-python@v4
         with:
-          output-dir: wheelhouse
+          python-version: "3.x"
+
+      - name: Build wheels
+        run: |
+          python3 -m pip install -r .ci/requirements-cibw.txt
+          python3 -m cibuildwheel --output-dir wheelhouse
         env:
           CIBW_ARCHS: ${{ matrix.archs }}
           CIBW_BUILD: ${{ matrix.build }}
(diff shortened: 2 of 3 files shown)


.github/renovate.json
```diff
@@ -11,6 +11,10 @@
             "groupName": "github-actions",
             "matchManagers": ["github-actions"],
             "separateMajorMinor": "false"
+        },
+        {
+            "groupName": "requirements-cibw.txt",
+            "matchPaths": [".ci/requirements-cibw.txt"]
```
radarhere, line 17:
```suggestion
            "matchFileNames": [".ci/requirements-cibw.txt"]
```
https://docs.renovatebot.com/release-notes-for-major-versions/
> matchPaths and matchFiles are now combined into matchFileNames, supporting exact match and glob-only.

.github/renovate.json
```diff
@@ -11,6 +11,10 @@
             "groupName": "github-actions",
             "matchManagers": ["github-actions"],
             "separateMajorMinor": "false"
+        },
+        {
+            "groupName": "requirements-cibw.txt",
+            "matchFileNames": [".ci/requirements-cibw.txt"]
```
radarhere, line 17: Are you sure that renovate won't handle this automatically? https://docs.renovatebot.com/modules/manager/pip_requirements/#file-matching

hugovk: I'm not sure. Shall we leave this file as is, and check when the next release comes out?

radarhere: I don't think we'd need to wait until the next release. I think if we merged the PR without this change, then we could see if #6604 started listing the file or not.

hugovk: Changes to this file dropped.

radarhere: Yep, now that this is merged, it's there!

PR #7584, opened 2023-11-28
Raise ValueError when TrueType font size is zero or less

### LINKED ISSUE #7583: Creating TrueType fonts with size=0 fails with Pillow 10+ differently than with Pillow 9
### What did you do?

```
(Pillow) ~/b/Pillow (main) $ docker run -it -v "./Arial Bold.ttf:/a.ttf:ro" python:3.11 bash
root@e1ff41aff78e:/# pip install 'pillow<10'
Successfully installed pillow-9.5.0
root@e1ff41aff78e:/# python
Python 3.11.6 (main, Nov  1 2023, 10:08:51) [GCC 12.2.0] on linux
>>> from PIL import ImageFont
>>> ImageFont.truetype("/a.ttf", size=0)
<PIL.ImageFont.FreeTypeFont object at 0xffff9ef7d990>
root@e1ff41aff78e:/# pip install 'pillow>=10'
Successfully installed pillow-10.1.0
root@e1ff41aff78e:/# python
Python 3.11.6 (main, Nov  1 2023, 10:08:51) [GCC 12.2.0] on linux
>>> from PIL import ImageFont
>>> ImageFont.truetype("/a.ttf", size=0)
  File "/usr/local/lib/python3.11/site-packages/PIL/ImageFont.py", line 797, in truetype
    return freetype(font)
           ^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/PIL/ImageFont.py", line 794, in freetype
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/PIL/ImageFont.py", line 226, in __init__
    self.font = core.getfont(
                ^^^^^^^^^^^^^
OSError: invalid ppem value
>>>
```

### What did you expect to happen? / What actually happened?

Font creation for a zero or negative size should fail alright, but maybe with a less cryptic error. (`size=-1` yields `invalid argument`.)

With Pillow 9.x, font construction succeeds, but e.g. `getbbox()` or attempting to draw raises `invalid argument`.

* OS: Linux, various
* Python: various
* Pillow: see above for variants

### See also

Downstream bug: https://github.com/WhyNotHugo/python-barcode/issues/217

akx:
Fixes #7583

* Validate `size` before FreeType gets to raise an "invalid ppem value" or "invalid argument" for us.

### DIFF
--- a/Tests/test_imagefont.py
+++ b/Tests/test_imagefont.py
@@ -1074,6 +1074,6 @@ def test_raqm_missing_warning(monkeypatch):
 
 
 @pytest.mark.parametrize("size", [-1, 0])
-def test_invalid_truetype_sizes_raise(layout_engine, size):
+def test_invalid_truetype_sizes_raise_valueerror(layout_engine, size):
     with pytest.raises(ValueError):
         ImageFont.truetype(FONT_PATH, size, layout_engine=layout_engine)
--- a/Tests/test_imagefont.py
+++ b/Tests/test_imagefont.py
@@ -1071,3 +1071,9 @@ def test_raqm_missing_warning(monkeypatch):
         "Raqm layout was requested, but Raqm is not available. "
         "Falling back to basic layout."
     )
+
+
+@pytest.mark.parametrize("size", [-1, 0])
+def test_invalid_truetype_sizes_raise(layout_engine, size):
+    with pytest.raises(ValueError):
+        ImageFont.truetype(FONT_PATH, size, layout_engine=layout_engine)
--- a/src/PIL/ImageFont.py
+++ b/src/PIL/ImageFont.py
@@ -788,7 +788,12 @@ def truetype(font=None, size=10, index=0, encoding="", layout_engine=None):
                      .. versionadded:: 4.2.0
     :return: A font object.
     :exception OSError: If the file could not be read.
+    :exception ValueError: If the font size is not greater than zero.
     """
+
+    if size <= 0:
+        msg = "font size must be greater than 0"
+        raise ValueError(msg)
 
     def freetype(font):
         return FreeTypeFont(font, size, index, encoding, layout_engine)


src/PIL/ImageFont.py
```diff
@@ -788,8 +788,13 @@ def truetype(font=None, size=10, index=0, encoding="", layout_engine=None):
                      .. versionadded:: 4.2.0
     :return: A font object.
     :exception OSError: If the file could not be read.
+    :exception ValueError: If the size given is invalid.
     """
 
+    if size <= 0:
+        msg = "font size must be positive"
```
radarhere, line 795:
```suggestion
    :exception ValueError: If the font size is not greater than zero.
    """

    if size <= 0:
        msg = "font size must be greater than 0"
```
Suggestion, to be closer to an existing error message.
https://github.com/python-pillow/Pillow/blob/a1d22970295d1edd48970fff0825a9e48fbfd75e/src/PIL/ImageOps.py#L367

Tests/test_imagefont.py
```diff
@@ -1071,3 +1071,9 @@ def test_raqm_missing_warning(monkeypatch):
         "Raqm layout was requested, but Raqm is not available. "
         "Falling back to basic layout."
     )
+
+
+@pytest.mark.parametrize("size", [-1, 0])
+def test_invalid_truetype_sizes_raise(size, layout_engine):
```
radarhere, line 1077:
```suggestion
def test_invalid_truetype_sizes_raise(layout_engine, size):
```
This is minor, but there seems to be a convention of the fixture going first.

https://github.com/python-pillow/Pillow/blob/a1d22970295d1edd48970fff0825a9e48fbfd75e/Tests/test_imagefont.py#L736
https://github.com/python-pillow/Pillow/blob/a1d22970295d1edd48970fff0825a9e48fbfd75e/Tests/test_imagefont.py#L781
https://github.com/python-pillow/Pillow/blob/a1d22970295d1edd48970fff0825a9e48fbfd75e/Tests/test_imagefont.py#L825

akx: There's a
```
def test_non_ascii_path(tmp_path, layout_engine):
```
though 😁

radarhere: It's not completely the case, no, but I do think it's most of the time in this file. And neither of those are params?

nulano: `tmp_path` is a builtin fixture: https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html

Ordering test arguments as `builtin_fixture, custom_fixture, parameterized_argument` seems like a reasonable convention to me.

The only other exception is https://github.com/python-pillow/Pillow/blob/697c24bc7dee8a679ae18d6267cc333f79a32dca/Tests/test_imagefont.py#L179-L181 which I'd argue should be changed to match all the other tests here.

nulano: I agree, just pointing out why I agree with @radarhere's suggestion. 😄

radarhere (approved):

Tests/test_imagefont.py
```diff
@@ -1071,3 +1071,9 @@ def test_raqm_missing_warning(monkeypatch):
         "Raqm layout was requested, but Raqm is not available. "
         "Falling back to basic layout."
     )
+
+
+@pytest.mark.parametrize("size", [-1, 0])
+def test_invalid_truetype_sizes_raise(layout_engine, size):
```
radarhere, line 1077:
```suggestion
def test_invalid_truetype_sizes_raise_valueerror(layout_engine, size):
```
Ending with `_raise` seems odd to me.

radarhere (approved):

PR #7585, opened 2023-11-28
Add gcc problem matcher to test.yml

nulano:
Similar to https://github.com/python/cpython/pull/18567, but I found there are extra spaces printed before the file name.
Would help find issues like https://github.com/python-pillow/Pillow/pull/7497#discussion_r1406492621 faster.

The matcher is only run on a single Ubuntu job to prevent duplicate annotations; I found that the macOS jobs created two annotations per warning.

See https://github.com/nulano/Pillow/pull/25 for an example.

Not adding a problem matcher for MSVC yet because there are already many warnings that need to be fixed first to avoid noise.

### DIFF
--- a/.github/problem-matchers/gcc.json
+++ b/.github/problem-matchers/gcc.json
@@ -1,0 +1,18 @@
+{
+    "__comment": "Based on vscode-cpptools' Extension/package.json gcc rule",
+    "problemMatcher": [
+        {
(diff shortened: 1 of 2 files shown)


radarhere:
Looks fine to me - my only concern is licensing/permission as the code comes from somewhere else. I could imagine an argument that this is too simplistic for such concerns though.

PR #7590, opened 2023-12-01
Update Windows wheels info

radarhere:
Update Windows wheels information in light of #7580 restoring 32-bit wheels for Windows and adding arm64 wheels.

This reverts #7347 and #7447

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -94,7 +94,6 @@ Released as needed privately to individual vendors for critical security-related
 
 ## Source and Binary Distributions
 
-### macOS and Linux
 * [ ] Download sdist and wheels from the [GitHub Actions "Wheels" workflow](https://github.com/python-pillow/Pillow/actions/workflows/wheels.yml)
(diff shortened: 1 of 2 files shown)


Yay295:
Are 32-bit Windows wheels now available for 10.0.0, 10.0.1, and 10.1.0?

radarhere:
No. Is that relevant to this documentation change? This documentation will only become part of 'stable' once Pillow 10.2.0 is released.

If you're just making an aside that they should retroactively released, could you comment on #7443?

Yay295:
Just that the updated documentation doesn't mention that they're missing for those three releases.

radarhere:
I think adding in information about wheels for previous versions could be awkward. There's also not arm64 Windows wheels before Pillow 10.2.0, or manylinux_2_28 wheels before Pillow 9.2.0. There's potentially a lot of information down that road. We currently have one table explaining historical Pillow support, and that seems helpful. Adding a second table for wheels seems confusing.

Also, once Pillow 10.2.0 is released, there will no Python version that doesn't have a wheel for at least one Pillow version.

PR #7597, opened 2023-12-03
Use list comprehensions to create transformed lists

hugovk:

 * Replace some for loops with list comprehensions
 * Identified with `ruff . --select PERF401`: "PERF401 Use a list comprehension to create a transformed list"
 * https://docs.astral.sh/ruff/rules/manual-list-comprehension/
 * It finds some that can't be easily replaced, so I didn't add this rule to `pyproject.toml`


src/PIL/ImageOps.py
```diff
@@ -557,9 +557,7 @@ def invert(image):
     :param image: The image to invert.
     :return: An image.
     """
-    lut = []
-    for i in range(256):
-        lut.append(255 - i)
+    lut = [255 - i for i in range(256)]
```
Yay295, line 560: This could be `list(range(255,-1,-1))`.

src/PIL/ImageCms.py
```diff
@@ -787,11 +787,8 @@ def getProfileInfo(profile):
         # info was description \r\n\r\n copyright \r\n\r\n K007 tag \r\n\r\n whitepoint
         description = profile.profile.profile_description
         cpright = profile.profile.copyright
-        arr = []
-        for elt in (description, cpright):
-            if elt:
-                arr.append(elt)
-        return "\r\n\r\n".join(arr) + "\r\n\r\n"
+        elements = [element for element in (description, cpright) if element]
+        return "\r\n\r\n".join(elements) + "\r\n\r\n"
```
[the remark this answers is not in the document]
nulano: It is faster on my system, but not by much:
```
C:>python -m timeit -s "description, cpright = 'sdjkdkfjhgdjfghkjghdfkj', None" "elements = [element for element in (description, cpright) if element]" "'\r\n\r\n'.join(elements) + '\r\n\r\n'"
2000000 loops, best of 5: 141 nsec per loop

C:>python -m timeit -s "description, cpright = 'sdjkdkfjhgdjfghkjghdfkj', None" "'\r\n\r\n'.join(filter(None, [description, cpright])) + '\r\n\r\n'"
2000000 loops, best of 5: 117 nsec per loop
```
I do agree with the readability argument.

src/PIL/ImagePalette.py
```diff
@@ -226,9 +220,7 @@ def negative(mode="RGB"):
 def random(mode="RGB"):
     from random import randint
 
-    palette = []
-    for i in range(256 * len(mode)):
-        palette.append(randint(0, 255))
+    palette = [randint(0, 255) for i in range(256 * len(mode))]
```
radarhere, line 223:
```suggestion
    palette = [randint(0, 255) for _ in range(256 * len(mode))]
```

radarhere (approved):

homm (approved):

PR #7603, opened 2023-12-04
Added support for reading DX10 BC4 DDS images

sambvfx:
This fixes issues reading BC4 dds files.

The [nvidia texture tool exporter](https://developer.nvidia.com/nvidia-texture-tools-exporter) generates BC4 dds files that use the DX10 header so that the exact DXGI format can be specified.


src/PIL/DdsImagePlugin.py
```diff
@@ -98,6 +98,8 @@
 DXGI_FORMAT_R8G8B8A8_TYPELESS = 27
 DXGI_FORMAT_R8G8B8A8_UNORM = 28
 DXGI_FORMAT_R8G8B8A8_UNORM_SRGB = 29
+DXGI_FORMAT_BC4_TYPELESS = 79
+DXGI_FORMAT_BC4_UNORM = 80
```
radarhere, line 102: A note that you may or may not care about - I'm hoping to get #6486 merged before the next release. If it is merged in time, then I would remove these new constants in favour of enums. You've absolutely done the right thing here, no need to change this PR, just letting you know that if you had plans to use these constants externally, things may not play out that way.

radarhere:
I've created https://github.com/sambvfx/Pillow/pull/1 to add some more test images - a PNG to compare the image against, and a TYPELESS variation of your test image.

To be certain, this is what your image is supposed to look like, yes?
bc4_unorm

sambvfx:
> I've created [sambvfx#1](https://github.com/sambvfx/Pillow/pull/1) to add some more test images - a PNG to compare the image against, and a TYPELESS variation of your test image.
> 
> To be certain, this is what your image is supposed to look like, yes? bc4_unorm

Thanks @radarhere, I've merged your changes.

One odd thing to note is that the original PNG I used was simply the `ati1.png` from the existing test images. I converted that to the `bc4_unorm.dds` using [nvidia texture tool exporter](https://developer.nvidia.com/nvidia-texture-tools-exporter) like this:

```
nvtt_export.exe ati1.png --format bc4 --output bc4_unorm.dds
```

Tweaking the tests to compare against the ati1.png instead of the new one you added fails... There's a chance I'm doing something wrong, but I find that a little unexpected.

radarhere:
The compression is [lossy](https://learn.microsoft.com/en-us/windows/win32/direct3d10/d3d10-graphics-programming-guide-resources-block-compression) format, so it is not unexpected that converting to it created a slightly different file.

radarhere (approved):

radarhere:
With #6486, this change is now mentioned in the release notes - https://pillow.readthedocs.io/en/latest/releasenotes/10.2.0.html#other-changes

src/PIL/DdsImagePlugin.py
radarhere: #6486 has now been merged

PR #7599, opened 2023-12-04
Optimization of ImageStat.Stat._getcount method

florath:
The optimized function improves the performance. The new implementation uses "sum" instead of the construct
"functools.reduce(operator.add, ...)". Tests showed that the new function is about three times faster than the original. Also it is shorter and easier to read and the dependency to functools and operator modules can be removed.

Changes proposed in this pull request:

 * Performance optimized ImageStat _getcount function

PR #7605, opened 2023-12-06
Added release notes for #7599 and #7593

radarhere:


radarhere:
Ok, I've updated the commit.

When I say 'an average of 3x faster', that's from running the speed comparison scripts from the issues over our test images and averaging the result.

hugovk:
Thanks!

PR #7609, opened 2023-12-08
Translate encoder error codes to strings; deprecate `ImageFile.raise_oserror()`

bgilbert:
When decoding, we use `raise_oserror()` to convert codec error codes to strings.  Adapt that code to be used when encoding as well.  Add a new helper function that returns the exception so we can still raise `from exc`.

Replace all callers of `raise_oserror()` with `raise _get_oserror()`.  Deprecate `raise_oserror()`, since it doesn't need to be in the public API.

Split out from https://github.com/python-pillow/Pillow/pull/7553#discussion_r1413910020.

### DIFF
--- a/docs/deprecations.rst
+++ b/docs/deprecations.rst
@@ -34,6 +34,16 @@ Since Pillow's C API is now faster than PyAccess on PyPy,
 ``Image.USE_CFFI_ACCESS``, for switching from the C API to PyAccess, is
 similarly deprecated.
 
+ImageFile.raise_oserror
+~~~~~~~~~~~~~~~~~~~~~~~
+
+.. deprecated:: 10.2.0
+
+``ImageFile.raise_oserror()`` has been deprecated and will be removed in Pillow
+12.0.0 (2025-10-15). The function is undocumented and is only useful for translating
+error codes returned by a codec's ``decode()`` method, which ImageFile already does
+automatically.
+
 Removed features
 ----------------
 
--- a/docs/releasenotes/10.2.0.rst
+++ b/docs/releasenotes/10.2.0.rst
@@ -12,6 +12,14 @@ TODO
 Deprecations
 ============
 
+ImageFile.raise_oserror
+^^^^^^^^^^^^^^^^^^^^^^^
+
+``ImageFile.raise_oserror()`` has been deprecated and will be removed in Pillow
+12.0.0 (2025-10-15). The function is undocumented and is only useful for translating
+error codes returned by a codec's ``decode()`` method, which ImageFile already does
+automatically.
+
 TODO
 ^^^^
 
--- a/docs/releasenotes/10.2.0.rst
+++ b/docs/releasenotes/10.2.0.rst
@@ -77,3 +77,9 @@ Calculating the :py:attr:`~PIL.ImageStat.Stat.count` and
 :py:attr:`~PIL.ImageStat.Stat.extrema` statistics is now faster. After the
 histogram is created in ``st = ImageStat.Stat(im)``, ``st.count`` is 3x as fast
 on average and ``st.extrema`` is 12x as fast on average.
+
+Encoder errors now report error detail as string
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+:py:exc:`OSError` exceptions from image encoders now include a textual description of
+the error instead of a numeric error code.
--- a/src/PIL/ImageFile.py
+++ b/src/PIL/ImageFile.py
@@ -63,15 +63,19 @@ Dict of known error codes returned from :meth:`.PyDecoder.decode`,
 # Helpers
 
 
-def raise_oserror(error):
+def _get_oserror(error, *, encoder):
     try:
         msg = Image.core.getcodecstatus(error)
     except AttributeError:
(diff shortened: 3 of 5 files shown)


radarhere (approved):

src/PIL/ImageFile.py
```diff
@@ -63,15 +63,19 @@
 # Helpers
 
 
-def raise_oserror(error):
+def _get_oserror(error, encoder=False):
```
hugovk, line 66: Let's make `encoder` keyword only, because calling `_get_oserror(errcode, encoder=True)` is clearer than `_get_oserror(errcode, True)`:


```suggestion
def _get_oserror(error, *, encoder=False):
```

bgilbert: Done.

bgilbert:
Should we switch all the internal callers to `_get_oserror()` and deprecate `raise_oserror()` entirely?  It seems like the functionality is only useful for code that calls into libImaging, i.e. not for external callers.  Its predecessor `raise_ioerror()` was inherited from PIL as an exported function.

Alternatively, if we don't want to do that, should we expose the new `encoder` argument on `raise_oserror()`?  It's a bit awkward that `raise_oserror()` will have less functionality than the helper function it wraps.

radarhere:
> Alternatively, if we don't want to do that, should we expose the new encoder argument on raise_oserror()? It's a bit awkward that raise_oserror() will have less functionality than the helper function it wraps.

That wouldn't completely simplify things though, would it? Because it is still ideal to raise the error `from` another in the following code
```python
raise _get_oserror(errcode, encoder=True) from exc
```
which `raise_oserror()` can't do, with or without an `encoder` argument, because doesn't return an error, it immediately raises it. 

While Pillow values backwards compatibility, I agree that `raise_oserror()` doesn't appear useful to users, so deprecating it sounds fine to me.

bgilbert:
Sounds good, done.

docs/deprecations.rst
```diff
@@ -34,6 +34,15 @@ Since Pillow's C API is now faster than PyAccess on PyPy,
 ``Image.USE_CFFI_ACCESS``, for switching from the C API to PyAccess, is
 similarly deprecated.
 
+ImageFile.raise_oserror
+~~~~~~~~~~~~~~~~~~~~~~~
+
+.. deprecated:: 10.2.0
+
+``ImageFile.raise_oserror`` has been deprecated and will be removed in Pillow
+12.0.0 (2025-01-02). The function is undocumented and is only useful for
```
hugovk, line 43: 12.0.0 is planned for 2025-10-15:

https://github.com/python-pillow/Pillow/milestone/49

```suggestion
12.0.0 (2025-10-15). The function is undocumented and is only useful for
```

src/PIL/_deprecate.py
```diff
@@ -47,6 +47,8 @@ def deprecate(
         raise RuntimeError(msg)
     elif when == 11:
         removed = "Pillow 11 (2024-10-15)"
+    elif when == 12:
+        removed = "Pillow 12 (2025-01-02)"
```
hugovk, line 51:
```suggestion
        removed = "Pillow 12 (2025-10-15)"
```

src/PIL/ImageFile.py
```diff
@@ -63,15 +64,25 @@
...
+
+
+def raise_oserror(error):
+    deprecate(
+        "raise_oserror",
+        12,
+        action="It is only useful for translating error codes from internal "
+        "Pillow APIs.",
+    )
+    raise _get_oserror(error, encoder=False)
```
radarhere, line 85: I wouldn't feel the need to specify the default value - but I realise there may be different views on this.

bgilbert: Yeah, I didn't at first.  I figure it provides some protection against copypaste errors down the road.  The consequence of an incorrect `encoder` value is easy to miss.

hugovk (approved):

src/PIL/ImageFile.py
```diff
@@ -63,15 +64,25 @@
 # Helpers
 
 
-def raise_oserror(error):
+def _get_oserror(error, *, encoder=False):
```
radarhere, line 67:
```suggestion
def _get_oserror(error, *, encoder):
```
Given https://github.com/python-pillow/Pillow/pull/7609/files#r1423830172, maybe requiring the argument would be helpful?

bgilbert: Yup, done.

docs/releasenotes/10.2.0.rst
```diff
@@ -12,6 +12,13 @@ TODO
 Deprecations
 ============
 
+ImageFile.raise_oserror
+^^^^^^^^^^^^^^^^^^^^^^^
+
+``ImageFile.raise_oserror`` has been deprecated and will be removed in Pillow
+12.0.0. The function is undocumented and is only useful for translating error
+codes returned by internal codec APIs.
```
radarhere, line 20:
```suggestion
12.0.0 (2025-10-15). The function is undocumented and is only useful for
translating error codes returned by internal codec APIs.
```

bgilbert: Done.

bgilbert:
Added @radarhere's changes from https://github.com/bgilbert/Pillow/pull/4.

PR #7611, opened 2023-12-08
Updated readthedocs to latest version of Python

radarhere:


.readthedocs.yml
```diff
@@ -5,7 +5,7 @@ formats: [pdf]
 build:
   os: ubuntu-22.04
   tools:
-    python: "3.11"
+    python: "3.12"
```
hugovk, line 8: Or for the last stable CPython version:

```suggestion
    python: "3"
```

https://docs.readthedocs.io/en/stable/config-file/v2.html#build-tools-python

PR #7612, opened 2023-12-09
Encryption, Mosaic, Gradation add...

ttotti0903:
We are Kr. MYONGJI Unv Computer Angineering Student.
Open Source Software (Oss) class, It's a class to learn how to deal with open sources, and we chose Pillow from many open sources, and we added mosaic filters, gradation filters, and encryption decryption.


radarhere:
Hi. A few notes.
- Could leave a comment here why you think your code should be part of Pillow?
	- Your image encryption methods are relatively short, and dont' connect to other parts of Pillow. If a user wanted this functionality, they could just write it themselves.
	- Regarding the filters, could you give some information about why you chose these filters, why you think they are popular, and some example images showing before and after?
- Would you mind using only English? I can understand that you might appreciate multiple languages being included, but I'm concerned that if we start this, the length of the documentation would increase significantly with all the languages being included.
- If you would like to add filters, src/PIL/ImageFilter.py seems a better home for the code than adding new files.
- We ask our PRs to add tests, to help ensure correctness and prevent code being broken in the future. You will find our existing test suite under the 'Tests' directory
- You will see that the Lint check is failing. Those problems will need to be addressed.

src/PIL/blackGradation.py
```diff
@@ -0,0 +1,58 @@
...
+            for y in range(height):
+                alpha = int((1.0 - x / width) * 255)
+                color = start + (alpha,)
+                image.putpixel((x, y), color)
+
+        return image
+
+    @staticmethod
+    def hex_to_rgb(hex_color):
+        return tuple(int(hex_color[i : i + 2], 16) for i in (0, 2, 4))
```
radarhere, line 38: You might be interested to know that we already have a method for this - [`ImageColor.getrgb()`](https://pillow.readthedocs.io/en/stable/reference/ImageColor.html#PIL.ImageColor.getrgb)

radarhere:
Judging by the lack of response, I feel like you simply created this PR as a once-off experiment with open source.

That is fine, I don't want to discourage anyone from entering this space, it just seems that if even the creator of a feature is not engaged or excited about it, then there might not be a large number of other users who would benefit from that feature.

Issue #6750, opened 2022-11-19
Release Pillow 9.4.0 on January 2, 2023


hugovk:
## Main Release

Released quarterly on January 2nd, April 1st, July 1st and October 15th.

* [x] Open a release ticket e.g. https://github.com/python-pillow/Pillow/issues/3154
* [x] Develop and prepare release in `main` branch.
* [x] Check [GitHub Actions](https://github.com/python-pillow/Pillow/actions) and [AppVeyor](https://ci.appveyor.com/project/python-pillow/Pillow) to confirm passing tests in `main` branch.
* [x] Check that all of the wheel builds [Pillow Wheel Builder](https://github.com/python-pillow/pillow-wheels) pass the tests in Travis CI and GitHub Actions.
* [x] In compliance with [PEP 440](https://www.python.org/dev/peps/pep-0440/), update version identifier in `src/PIL/_version.py`
* [x] Update `CHANGES.rst`.
* [x] Run pre-release check via `make release-test` in a freshly cloned repo.
* [x] Create branch and tag for release e.g.:
  ```bash
  git branch 9.4.x
  git tag 9.4.0
  git push --all
  git push --tags
  ```
* [x] Create and check source distribution:
  ```bash
  make sdist
  ```
* [x] Create [binary distributions](https://github.com/python-pillow/Pillow/blob/main/RELEASING.md#binary-distributions)
* [x] Check and upload all binaries and source distributions e.g.:
  ```bash
  twine check dist/*
  twine upload dist/Pillow-9.4.0*
  ```
* [x] Publish the [release on GitHub](https://github.com/python-pillow/Pillow/releases)
* [x] In compliance with [PEP 440](https://www.python.org/dev/peps/pep-0440/), increment and append `.dev0` to version identifier in `src/PIL/_version.py`

### Windows
* [x] Download the artifacts from the [GitHub Actions "Test Windows" workflow](https://github.com/python-pillow/Pillow/actions/workflows/test-windows.yml) and copy into `dist/`

### Mac and Linux
* [x] Use the [Pillow Wheel Builder](https://github.com/python-pillow/pillow-wheels):
  ```bash
  git clone https://github.com/python-pillow/pillow-wheels
  cd pillow-wheels
  ./update-pillow-tag.sh [[release tag]]
  ```
* [x] Download wheels from the [Pillow Wheel Builder release](https://github.com/python-pillow/pillow-wheels/releases) and copy into `dist/`

## Publicize Release

[...]

PR #7616, opened 2023-12-14
Added trusted PyPI publishing

radarhere:
Helps #7390

- Removes "Wheels Successful" job. These jobs were added as part of #4931 for mergify, but this workflow isn't referenced in https://github.com/python-pillow/Pillow/blob/main/.github/mergify.yml
- Adds "Upload release to PyPI" step, as per https://github.com/marketplace/actions/pypi-publish. This still requires the steps from https://docs.pypi.org/trusted-publishers/adding-a-publisher/ to be taken before it will work.


.github/workflows/wheels.yml
```diff
@@ -195,12 +195,16 @@ jobs:
...
+    name: Upload release to PyPI
+    environment:
+      name: pypi
+      url: https://pypi.org/p/Pillow
+    permissions:
+      id-token: write
     steps:
-      - name: Success
-        run: echo Wheels Successful
+      - name: Publish package distributions to PyPI
```
hugovk, line 209:
```suggestion
      - name: Publish to PyPI
```

.github/workflows/wheels.yml
```diff
@@ -195,12 +195,16 @@ jobs:
...
 
-  success:
-    permissions:
-      contents: none
+  pypi-publish:
+    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
     needs: [build, windows, sdist]
     runs-on: ubuntu-latest
-    name: Wheels Successful
+    name: Upload release to PyPI
```
hugovk, line 202: We'll need a step or two before this to download the artifacts into the correct dir using https://github.com/actions/download-artifact, so we have the files from the preceding jobs.

hugovk:
Very nice!

Looks like:

image

---

I've been using trusted publishing in my other projects (although not with environments).

I recommend we also set up trusted publishing to TestPyPI, so we know the whole pipeline is working before the big release day.

I've already registered Pillow at https://test.pypi.org/project/pillow/ and have set up trusted publishers on that side:

image


---

One sticking point:

You can't publish a file to [Test]PyPI that has already been published there, unless the version number changes. But we have a hardcoded value in `src/PIL/_version.py`.

One way is to use https://github.com/pypa/setuptools_scm/ to generate a version number based on the "distance"/number of commits since the last tag. We'd also need `local_scheme = "no-local-version"`, it removes a `+abc123` suffix that PyPI doesn't like, and also need some juggling to get the version to the C code. This could be a separate PR.

radarhere:
> You can't publish a file to [Test]PyPI that has already been published there, unless the version number changes. But we have a hardcoded value in src/PIL/_version.py.

To make sure I'm following - this is only a problem with test.pypi.org, because we may want to submit the same version more than once, yes? You're not suggesting that anything about the normal version mechanism needs to be changed.

.github/workflows/wheels.yml
```diff
@@ -195,12 +195,16 @@ jobs:
...
+    environment:
+      name: pypi
```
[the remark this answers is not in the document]
radarhere: What is the effect of changing this?

.github/workflows/wheels.yml
radarhere: I've added a step to the download the artifacts. I presume they should all be together, so I also added a step to move them out of their individual directories (now that actions/upload-artifact v4 means that artifacts need different names).

.github/workflows/wheels.yml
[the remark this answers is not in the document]
hugovk: It's an arbitrary name of our choosing.

It must match the name we set in PyPI (see screenshot: https://github.com/python-pillow/Pillow/pull/7616#issuecomment-1855773030 (I can change this if needed)), and it will be the name of the environment we see in the GitHub UI.

Without an environment, the release will be published to PyPI from the CI workflow.

With an environment, a repo admin will first need to set one up for us. They will choose who can access it. The name above will be used here. With an env, the CI will first stage it in the env. Then someone with access can go and approve it and allow it to deploy.

It's an extra bit of security, but also an extra step in the release process. Without an env, anyone with write access to the repo and create tags, can publish. That's the five Pillow core devs. Currently, only the four of us listed at https://pypi.org/project/Pillow/ can `twine upload`. The env means we're essentially preserving the "maintainer" and "release manager" roles described at https://sethmlarson.dev/security-for-package-maintainers#platforms-and-roles.

Could be we don't need it? So far in other projects, I've not used envs because maintainers == release managers.

Anyway, `release-[test-]pypi` is perhaps a bit more descriptive than `[test-]pypi`, copying another [project](https://github.com/python-attrs/attrs/blob/main/.github/workflows/pypi-package.yml), but it's arbitrary and just a bit of bikeshedding.

hugovk:
> > You can't publish a file to [Test]PyPI that has already been published there, unless the version number changes. But we have a hardcoded value in src/PIL/_version.py.
> 
> To make sure I'm following - this is only a problem with test.pypi.org, because we may want to submit the same version more than once, yes? 

Yes and no :)

No, both TestPyPI and PyPI behave the same in this regard.

The restriction really is that filenames cannot be re-used. So we can re-use version numbers, but need to 

For example, we have a `
Pillow-10.1.0-cp312-cp312-macosx_10_10_x86_64.whl` at https://pypi.org/project/Pillow/10.1.0/#files. If there was something wrong with that file and we needed to upload a new one, we can't upload a file with the same name, PyPI will reject it with an error. That's why we've needed to use wheel build tags in the past, so we have a unique file we can upload:

* https://github.com/python-pillow/Pillow/issues/6750#issuecomment-1373331324
* https://packaging.python.org/en/latest/specifications/binary-distribution-format/#file-format
* https://snarky.ca/what-to-do-when-you-botch-a-release-on-pypi/

And so yes, we may submit the same version more than once, but need to make it unique in other ways.

---

> You're not suggesting that anything about the normal version mechanism needs to be changed.

Correct. 

For normal releases, we still have the same version numbers.

This would give us new version numbers for testing uploads of intra-release builds to TestPyPI.

If we look at https://github.com/hugovk/norwegianblue (now uses Hatchling+hatch-vcs, but used to use setuptools+setuptools_scm and both work in the same way (I think hatch-vcs uses setuptools_scm under the hood).

* PyPI has "normal" release numbers for proper releases: https://pypi.org/project/norwegianblue/#history
* TestPyPI has releases appended with `.devX`: 0.16.1.dev4, 0.16.1.dev6: https://test.pypi.org/project/norwegianblue/#history

---

> You're not suggesting that anything about the normal version mechanism needs to be changed.

Although there would be one change:

Right now, after we release 10.2.0, we bump the hardcoded version to 10.3.0.dev0. 

With setuptools_scm, we'd still be getting a version like 10.3.0.dev1, 10.3.0.dev200, etc, because it's based on the last tag.

I think that's fine, it's an internal change.

This does however bring to mind a potential blocker: we do compile the version into the C layer, and error if they're different. With setuptools_scm, the devX is incremented for every commit, which could result in mismatches or needed to rebuild C more often.. This could be a blocker, and we may need another solution to give a unique filename for TestPyPI...

radarhere:
> we do compile the version into the C layer, and error if they're different. With setuptools_scm, the devX is incremented for every commit, which could result in mismatches

We could relax the condition to allow 10.3.0.dev1 or 10.3.0.dev200, by changing
https://github.com/python-pillow/Pillow/blob/e5ee034675491d1f67410022f07256ed583cbbc5/src/PIL/Image.py#L84
to
```python
    if not hasattr(core, "PILLOW_VERSION") or re.sub(
        "dev[0-9]+", "dev0", __version__
    ) != re.sub("dev[0-9]+", "dev0", core.PILLOW_VERSION):
```

hugovk:
Yep, that could work.

radarhere:
I've just registered for test.pypi.org with the same username. Would you mind adding me as a maintainer to Pillow?

hugovk:
Invited!

hugovk:
If possible, I'd prefer to automate testing where possible rather than do it manually. Doing it manually means we have to foresee a change might cause problems in the release pipeline. And as said, we can consider that in a separate PR, and can consider other approaches too.

.github/workflows/wheels.yml
```diff
@@ -190,17 +197,35 @@ jobs:
...
+                          macOS_arm64 \
+                          manylinux2014_musllinux \
+                          manylinux_2_28 \
+                          windows_x86 \
+                          windows_x64 \
+                          windows_arm64 \
+                          sdist; do
+            mv $artifact/* .
+            rmdir $artifact
+          done
```
hugovk, line 229: Let's try something like https://github.com/scikit-build/cmake-python-distributions/pull/438 / https://github.com/scientific-python/cookie/pull/335 to avoid repeating the names:

```suggestion
      - uses: actions/download-artifact@v4
        with:
          path: all

      - name: Merge files
        run: |
          mkdir dist
          mv all/*/* dist

```

.github/workflows/wheels.yml
```diff
@@ -157,9 +164,9 @@ jobs:
         shell: cmd
 
       - name: Upload wheels
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
-          name: dist
+          name: windows_${{ matrix.artifact_suffix }}
```
hugovk, line 169: Can we re-use `arch` and/or `cibw_arch` instead of introducing a new `artifact_suffix` variable?

.github/workflows/wheels.yml
```diff
@@ -75,9 +79,9 @@ jobs:
           CIBW_TEST_SKIP: "*-macosx_arm64"
           MACOSX_DEPLOYMENT_TARGET: ${{ matrix.macosx_deployment_target }}
 
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v4
         with:
-          name: dist
+          name: ${{ matrix.artifact_name }}
```
hugovk, line 84: Can we re-use existing matrix variables instead of introducing a new `artifact_name`?

The tricky bit might be differentiating between these two:

```yaml
          - name: "manylinux2014 and musllinux x86_64"
            os: ubuntu-latest
            archs: x86_64
            artifact_name: "manylinux2014_musllinux"
          - name: "manylinux_2_28 x86_64"
            os: ubuntu-latest
            archs: x86_64
            build: "*manylinux*"
            manylinux: "manylinux_2_28"
            artifact_name: "manylinux_2_28"
```

If we do `wheel-${{ matrix.os }}-${{ matrix.archs }}-${{ matrix.manylinux }}`, does that error or do we get something like this?

* `wheel-ubuntu-latest-x86_64`
* `wheel-ubuntu-latest-x86_64-manylinux_2_28`

.github/workflows/wheels.yml
radarhere: Ok, I've pushed a commit to use `arch` instead.

.github/workflows/wheels.yml
radarhere: I've pushed a commit to use `${{ matrix.os }}-${{ matrix.archs }}-${{ matrix.manylinux }}`.

We get
- macos-latest-x86_64-
- macos-latest-arm64-
- ubuntu-latest-x86_64-
- ubuntu-latest-x86_64-manylinux_2_28

.github/workflows/wheels.yml
radarhere: Ok, I've pushed something similar.

.github/workflows/wheels.yml
hugovk: Thanks!

Since my last suggestion, and based on lots of feedback about the breaking `@v4` changes, GitHub have just added `pattern:` and `merge-multiple:` inputs:

https://github.com/actions/upload-artifact/issues/472#issuecomment-1861571655

Shall we try that? We'll need a common prefix/pattern for the sdist and all the wheel artifacts.

.github/workflows/wheels.yml
```diff
@@ -190,17 +190,26 @@ jobs:
...
     steps:
-      - name: Success
-        run: echo Wheels Successful
+      - uses: actions/download-artifact@v4
+      - name: Merge artifacts
+        run: |
+          mkdir dist
+          mv */* dist
+      - name: Publish to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
```
hugovk, line 215: I'm expecting we'll need a few cycles to get the actual upload working, so let's add some temporary commits to publish to TestPyPI, and then rebase them out later:

```suggestion
      - name: Publish to PyPI
        uses: pypa/gh-action-pypi-publish@release/v1
        with:
          repository-url: https://test.pypi.org/legacy/
```
Plus we could comment out/replace the `if` to temporarily trigger.

Do we want to use a staging environment before publishing? If so, please let a repo admin know what we need there (I don't know, other than to call it `release-pypi`).

We can always remove it from this PR and add it later if we want.

webknjaz:
> * Removes "Wheels Successful" job. These jobs were added as part of [Add Mergify config #4931](https://github.com/python-pillow/Pillow/pull/4931) for mergify, but this workflow isn't referenced in https://github.com/python-pillow/Pillow/blob/main/.github/mergify.yml

Moreover, the "gate" was implemented incorrectly: https://github.com/marketplace/actions/alls-green#why.

.github/workflows/wheels.yml
```diff
@@ -171,9 +171,9 @@ jobs:
...
       - name: Upload fribidi.dll
         if: "matrix.arch != 'ARM64'"
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
-          name: fribidi
+          name: fribidi_${{ matrix.arch }}
           path: fribidi\*
 
   sdist:
```
webknjaz, line 179: May I suggest building wheels from sdist? This is what pip does — it doesn't build wheels straight from Git source, but from the tarballs. And this is what pypa/build does if you don't pass any args. Doing so will ensure that sdists are actually usable, as a smoke test.

Here's an example: https://github.com/aio-libs/yarl/actions/runs/7258064074/workflow#L124.

P.S. `cibuildwheel` accepts a path to the sdist tarball natively (in place of a directory path) so integrating this into the wheel build jobs is as easy as downloading the artifact and feeding the sdist from it into the `cibuildwheel` action.

nulano:
> Moreover, the "gate" was implemented incorrectly: https://github.com/marketplace/actions/alls-green#why.

The argument being that a "skipped" status is incorrectly accepted, yes?

This does not appear to be the case for Mergify, which is the only use of these gates here. A recent PR, https://github.com/python-pillow/Pillow/pull/7608, only updated documentation and so most of the gates were skipped. However, the Mergify report did not accept these skips as a success: https://github.com/python-pillow/Pillow/runs/19444375622

.github/workflows/wheels.yml
webknjaz: FYI my PyPUG guide was updated with examples of Sigstore signatures a few months ago: https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/.

And also, I suggest using two separate jobs for PyPI and TestPyPI since they are different environments semantically and would benefit from having different GitHub Environment names and URLs, being displayed distinctly in the GH UI + the protections can be different (I normally have PyPI with required reviewers and TestPyPI w/o it)

.github/workflows/wheels.yml
webknjaz: Oh, and that repo also features the wheel build jobs moved into a reusable workflow and called with different params.

webknjaz:
> The argument being that a "skipped" status is incorrectly accepted, yes?
> 
> This does not appear to be the case for Mergify, which is the only use of these gates here. A recent PR, #7608, only updated documentation and so most of the gates were skipped. However, the Mergify report did not accept these skips as a success: https://github.com/python-pillow/Pillow/runs/19444375622

Yeah, you're right. Mergify treats these statuses better. But overall, on the GH platform level, it might still be reasonable to implement them correctly so that somebody doesn't trip over this implementation quirk in the future.

.github/workflows/wheels.yml
webknjaz: Fair

.github/workflows/wheels.yml
```diff
@@ -75,9 +75,9 @@ jobs:
           CIBW_TEST_SKIP: "*-macosx_arm64"
           MACOSX_DEPLOYMENT_TARGET: ${{ matrix.macosx_deployment_target }}
 
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v4
         with:
-          name: dist
+          name: ${{ matrix.os }}-${{ matrix.archs }}-${{ matrix.manylinux }}
```
nulano, line 80:
```suggestion
          name: ${{ matrix.os }}-${{ matrix.archs }}${{ matrix.manylinux && '-' }}${{ matrix.manylinux }}
```
Would get rid of the extra trailing `-`; not sure if it's worth sacrificing a bit of readability to make the output names cleaner?

webknjaz: Could also be something like
```suggestion
          name: >-
            ${{
              format(
                '{0}-{1}{2}{3}',
                matrix.os,
                matrix.archs,
                (matrix.manylinux && '-' || ''),
                matrix.manylinux
              )
            }}
```
_(not tested)_

.github/workflows/wheels.yml
radarhere: Thanks for finding that. I've updated the PR.

.github/workflows/wheels.yml
radarhere: > May I suggest building wheels from sdist?

We're currently using the `.ci` directory in the build process, which isn't included in the sdist
https://github.com/python-pillow/Pillow/blob/41e45b5f95c5184b8b4c4712e5614bf468331e05/.github/workflows/wheels.yml#L65-L67

Sure, we could checkout the repository AND use sdist - but that doesn't sound very neat to me.

I know you said that `cibuildwheel` can simply accept an sdist path, but we're also currently adding in extra test images, which wouldn't be quite so simple when using a compressed sdist.
https://github.com/python-pillow/Pillow/blob/41e45b5f95c5184b8b4c4712e5614bf468331e05/.github/workflows/wheels.yml#L99-L103

> Oh, and that repo also features the wheel build jobs moved into a reusable workflow

If wheels.yml gets much longer, I could see that being nice, but at the moment, I don't think it's problematic.

Also, neither of these ideas depend on this PR. If you would like to see them become part of Pillow, you could create a separate PR.

.github/workflows/wheels.yml
radarhere: My preference would be in the middle, with
```
${{ matrix.os }}-${{ matrix.archs }}${{ matrix.manylinux && format('-{0}', matrix.manylinux) }}
```
but I have a feeling that would make no one happy.

.github/workflows/wheels.yml
```diff
@@ -190,17 +190,28 @@ jobs:
...
       with:
-        name: dist
+        name: dist-sdist
         path: dist/*.tar.gz
 
-  success:
-    permissions:
-      contents: none
+  pypi-publish:
+    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
```
hugovk, line 199: Let's check it can upload to TestPyPI:
```suggestion
    # TEMP for testing
    # if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
    if: github.event_name == 'push'
```

.github/workflows/wheels.yml
nulano: > My preference would be in the middle, with [...]

Actually, that looks like the best option to me also.

radarhere: Cool, thanks, I've pushed a commit for that.

.github/workflows/wheels.yml
```diff
@@ -190,17 +190,30 @@ jobs:
...
+        name: dist-sdist
         path: dist/*.tar.gz
 
-  success:
-    permissions:
-      contents: none
+  pypi-publish:
+    # TEMP for testing
+    # if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
+    if: github.event_name == 'push'
```
hugovk, line 201: We're actually getting a PR event here. 
```suggestion
```

webknjaz: I'm not sure you'd be able to publish from PRs.

.github/workflows/wheels.yml
```diff
@@ -190,17 +190,29 @@ jobs:
...
+  pypi-publish:
+    # TEMP for testing
+    # if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
     needs: [build, windows, sdist]
     runs-on: ubuntu-latest
-    name: Wheels Successful
+    name: Upload release to PyPI
+    environment:
+      name: release-pypi
+      url: https://pypi.org/p/Pillow
```
hugovk, line 206: Maybe it's because we don't have an env set-up yet? Shall we try without, and we can always add back later?

```suggestion
```

hugovk:
```
Notice: Attempting to perform trusted publishing exchange to retrieve a temporary short-lived API token for authentication against https://test.pypi.org/legacy/ due to __token__ username with no supplied password field
Error: Trusted publishing exchange failure: 
OpenID Connect token retrieval failed: GitHub: missing or insufficient OIDC token permissions, the ACTIONS_ID_TOKEN_REQUEST_TOKEN environment variable was unset

This generally indicates a workflow configuration error, such as insufficient
permissions. Make sure that your workflow has `id-token: write` configured
at the job level, e.g.:

permissions:
  id-token: write

Learn more at https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect#adding-permissions-settings.
```
https://github.com/python-pillow/Pillow/actions/runs/7324548352/job/19948625585

But we have that at the job level. Maybe it's because we have an env configured at TestPyPI. I'll remove that, and restart the job.

hugovk:
Added the second one, and left the first one:

image

And restarted...

hugovk:
Hmm, same thing. Removed the first one and restarted...

hugovk:
Nope, same thing. Perhaps it's as @webknjaz said in one thread, you can't publish from a PR.

nulano:
PRs don't have access to most secrets (otherwise they could be trivially extracted by a malicious PR). From https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow:

> With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.

Although the publish action uses `GITHUB_TOKEN`, that token cannot have `write` permissions by default. From https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#changing-the-permissions-in-a-forked-repository:

> You can use the permissions key to add and remove read permissions for forked repositories, but typically you can't grant write access. The exception to this behavior is where an admin user has selected the Send write tokens to workflows from pull requests option in the GitHub Actions settings. For more information, see "[Managing GitHub Actions settings for a repository](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#enabling-workflows-for-private-repository-forks)."

If I understand [this](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#enabling-workflows-for-forks-of-private-repositories) correctly, the required setting can only be enabled for private repositories (where only members of the organization can create PRs).

webknjaz:
> Nope, same thing. Perhaps it's as @webknjaz said in one thread, you can't publish from a PR.

Yes, PR privileges are lower. At least, when the PRs are coming from forks. Some of them can be addressed through `pull_request_target`. But that's its own can of worms.

webknjaz:
> Although the publish action uses GITHUB_TOKEN, that token cannot have write permissions by default.

No, my action doesn't access that token. GHA sets two special env vars for OIDC when `id: write` is set and is allowed in the given context.

webknjaz:
> Hmm, same thing. Removed the first one and restarted...

I recommend keeping the environment. Just make sure it matches what you set up on the PyPI side. I suggest using `pypi` for PyPI and `testpypi` for TestPyPI in my guides and tutorials. These represent the targets rather accurately.

webknjaz:
Perhaps we should spell it out in the error message that GH does allow OIDC in untrusted event contexts, like `pull_request`.

cc @woodruffw

nulano:
> No, my action doesn't access that token. GHA sets two special env vars for OIDC when id: write is set and is allowed in the given context.

That's what I thought, but I got confused by the GHA documentation. 
Not that it matters, but I guess those envvars are automatically extracted(?) from the GITHUB_TOKEN by the runner?

woodruffw:
Just to provide some disambiguation:

* Trusted publishing allows publishing from a `pull_request` event. More succinctly: as long as a workflow has `id-token: write`, trusted publishing will work from it.
* GitHub *itself* does not give `id-token: write` to `pull_request` events run from forks, even if the workflow explicitly adds that permission. This isn't super well documented anywhere, but makes sense from a security perspective (I believe this is the same for all other `write` permissions, to prevent 3p forks from making destructive changes to the "target" repo by default).

So TL;DR: you *can* publish to PyPI from a `pull_request` event, but only if the `pull_request` event is coming from a first-party branch (**not** from a fork). This limitation persists even if the fork is controlled or pushed to by an owner of the upstream repository.

woodruffw:
> Perhaps we should spell it out in the error message that GH does allow OIDC in untrusted event contexts, like `pull_request`.

IMO this is likely to cause some confusion, although I think there's a variant we can chip out per https://github.com/python-pillow/Pillow/pull/7616#issuecomment-1869167753: if the publish is running in a `pull_request` event and we detect that it's a branch from a fork, then we can produce a more useful error message, like:

```
This error indicates that you're attempting to obtain an OpenID Connect token from a fork of a repository, which GitHub does not permit.

<link to another troubleshooting section>
```

hugovk:
Thanks for clearing that up! That's all fine, we don't want to deploy from PRs in normal use, it was for testing.

So I think we can merge this pretty much as-is, test from `main`, then cleanup TestPyPI -> PyPI later.

Another option: @radarhere, if you like, we could try things out from your fork. Feel free to add a new publisher for that at: https://test.pypi.org/manage/project/pillow/settings/publishing/

Our next quarterly release is in just a week, 2nd January, so I'd prefer to just get something merged and working first to minimise risk. This will also be the first release with cibuildwheel, plus replacing `setup.cfg` with `pyproject.toml`, so we have a few big changes.

radarhere:
Ok, sure. I added a publisher, and merged this to my main.

There was already a partial release for 10.2.0.dev0, so I updated the version to 10.2.0.dev1 and [it worked.](https://github.com/radarhere/Pillow/actions/runs/7329629479/job/19960359962)

hugovk:
Great! Did you test environments?

Let's merge this, and can you arrange with a repo admin to set up what is needed in this repo, and on PyPI? Or for them to give you and me access to do it.

.github/workflows/wheels.yml
```diff
@@ -154,9 +154,9 @@ jobs:
         shell: cmd
 
       - name: Upload wheels
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
-          name: dist
+          name: dist-${{ matrix.arch }}
```
nulano, line 159:
```suggestion
          name: dist-windows-${{ matrix.arch }}
```

radarhere:
I tested it with an environment, yes.

https://github.com/radarhere/Pillow/blob/9078f85a55090a7c4bca6c304b4b0d480de72f13/.github/workflows/wheels.yml#L187-L207
```
  pypi-publish:
    # TEMP for testing
    # if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
    needs: [build, windows, sdist]
    runs-on: ubuntu-latest
    name: Upload release to PyPI
    permissions:
      id-token: write
    environment:
      name: radarhere-test-pypi
      url: https://test.pypi.org/p/Pillow
    steps:
      - uses: actions/download-artifact@v4
        with:
          pattern: dist-*
          path: dist
          merge-multiple: true
      - name: Publish to PyPI
        uses: pypa/gh-action-pypi-publish@release/v1
        with:
          repository-url: https://test.pypi.org/legacy/
```

webknjaz:
@woodruffw that sounds reasonable!

.github/workflows/wheels.yml
webknjaz: > We're currently using the `.ci` directory in the build process, which isn't included in the sdist

And this is a reason to include it. Downstreams might want to run tests using the sdist as the source.

> If wheels.yml gets much longer, I could see that being nice, but at the moment, I don't think it's problematic.

For me, it's a structural thing that helps layer components. Not necessarily related to how big the YAML files are (although, that's also important).

> Also, neither of these ideas depend on this PR. If you would like to see them become part of Pillow, you could create a separate PR.

Sure. These are mostly drive-by comments I made having seen some familiar faces and tech involved.

PR #7622, opened 2023-12-18
Check types with mypy

hugovk:
Helps #2625.

Run mypy to check type hints:

```sh
tox -e mypy
```

Some good advice here:

* https://mypy.readthedocs.io/en/stable/existing_code.html

So let's start off small with just the `src` directory, and exclude lots of files with warnings, currently 61, so it's running on 34 to begin with.

This adds some fairly lenient settings (happy to tweak them now and later), notably `follow_imports = "silent"`.

When mypy encounters an import, it also checks the imported file, and its imports. They recommend using the default "normal", but lots of things import others which aren't ready yet, so let's change it when we can.

* https://mypy.readthedocs.io/en/stable/running_mypy.html#following-imports

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -94,7 +94,6 @@ test-command = "cd {project} && .github/workflows/wheels-test.sh"
 test-extras = "tests"
 
 [tool.ruff]
-line-length = 88
 select = [
   "C4", # flake8-comprehensions
   "E", # pycodestyle errors
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -125,3 +125,34 @@ known-first-party = ["PIL"]
 [tool.pytest.ini_options]
 addopts = "-ra --color=yes"
 testpaths = ["Tests"]
+
+[tool.mypy]
+python_version = "3.8"
+pretty = true
+disallow_any_generics = true
+enable_error_code = "ignore-without-code"
+extra_checks = true
+follow_imports = "silent"
+warn_redundant_casts = true
+warn_unreachable = true
+warn_unused_ignores = true
+exclude = [
+  '^src/PIL/_tkinter_finder.py$',
+  '^src/PIL/DdsImagePlugin.py$',
+  '^src/PIL/FpxImagePlugin.py$',
+  '^src/PIL/Image.py$',
+  '^src/PIL/ImageCms.py$',
+  '^src/PIL/ImageFile.py$',
+  '^src/PIL/ImageFont.py$',
+  '^src/PIL/ImageMath.py$',
+  '^src/PIL/ImageMorph.py$',
+  '^src/PIL/ImageQt.py$',
+  '^src/PIL/ImageShow.py$',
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -29,3 +29,10 @@ pass_env =
 commands =
     pre-commit run --all-files --show-diff-on-failure
     check-manifest
+
+[testenv:mypy]
+skip_install = true
+deps =
+    mypy==1.7.1
+commands =
+    mypy src {posargs}
```
radarhere, line 38: [`{posargs}`](https://tox.wiki/en/4.11.4/user_guide.html#basic-example) isn't used by the Lint job, right? You're imagining this as being helpful locally somehow?

pyproject.toml
```diff
@@ -125,3 +124,79 @@ known-first-party = ["PIL"]
...
+[tool.mypy]
+python_version = "3.8"
+pretty = true
+disallow_any_generics = true
+enable_error_code = "ignore-without-code"
+extra_checks = true
+follow_imports = "silent"
+warn_redundant_casts = true
+warn_unreachable = true
+warn_unused_configs = true
```
radarhere, line 137: https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-warn-unused-configs
> This requires turning off incremental mode using [--no-incremental](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-no-incremental)

tox.ini
hugovk: That's right, so you can pass arguments through, like `tox -e mypy -- --verbose`.

tox can be useful locally, it takes care of the virtualenv so you're not testing against other things you may have installed globally.

pyproject.toml
hugovk: Let's remove this one then, the cache is more important.

hugovk, line 137:
```suggestion
```

hugovk:
Rebased after https://github.com/python-pillow/Pillow/pull/7624, and removed 44 files from the exclude, it now checks 76 files!

tox.ini
```diff
@@ -29,3 +29,10 @@ pass_env =
 commands =
     pre-commit run --all-files --show-diff-on-failure
     check-manifest
+
+[testenv:mypy]
+skip_install = true
+deps =
+    mypy==1.7.1
```
radarhere, line 36:
```suggestion
    mypy==1.7.1
    numpy
```
There is [currently](https://github.com/python-pillow/Pillow/actions/runs/7258653852/job/19774332621?pr=7622) a failure on ImageFilter.

A simple solution would be to exclude ImageFilter again, but this is also a fix.

hugovk: This is also a good demo of tox: when I updated the list I had been running mypy directly, and have NumPy installed globally. If I'd run it via tox, I would have run into this failure :)

radarhere (approved):

PR #7631, opened 2023-12-21
Add `from __future__ import annotations` using Ruff/isort

hugovk:
Helps https://github.com/python-pillow/Pillow/issues/2625.

Adds the `from __future__ import annotations` to all files, and will ensure it's added for new files too, so we can use newer type annotations for older Pythons.

For example, use `list` and `tuple` directly instead of first `from typing import List, Tuple`.

### DIFF
--- a/src/PIL/JpegPresets.py
+++ b/src/PIL/JpegPresets.py
@@ -64,6 +64,8 @@ https://web.archive.org/web/20120328125543/http://www.jpegcameras.com/libjpeg/li
 """
 
 # fmt: off
+from __future__ import annotations
+
 presets = {
             'web_low':      {'subsampling':  2,  # "4:2:0"
                              'quantization': [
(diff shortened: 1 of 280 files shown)


radarhere:
I see you've added it to the test suite as well. I would have thought that the goal of #2625 was to add type hints to user-facing code. Or are you suggesting that expecting that we will eventually use type hints for their own sake?

hugovk:
User-facing code is the primary focus, but typing tests can also be beneficial, especially if it helps evaluate how the typed production code is used, and if it helps us find bugs in tests. 

See also https://sethmlarson.dev/tests-arent-enough-case-study-after-adding-types-to-urllib3#type-your-tests

src/PIL/JpegPresets.py
```diff
@@ -64,6 +64,8 @@
 """
 
 # fmt: off
+from __future__ import annotations
```
radarhere, line 67:
```suggestion
from __future__ import annotations

# fmt: off
```

radarhere (approved):

PR #7632, opened 2023-12-22
Convert colormap hex values to RGB beforehand instead of on demand

radarhere:
The `colormap` dictionary maps color names to hex values.
https://github.com/python-pillow/Pillow/blob/bfbfff0b785125a76f74e6571c2af8c0f900a098/src/PIL/ImageColor.py#L161-L166

When it is used, the hex values is converted to an RGB tuple. That value is then saved back to `colormap`, as a way of caching the conversion to RGB.
https://github.com/python-pillow/Pillow/blob/bfbfff0b785125a76f74e6571c2af8c0f900a098/src/PIL/ImageColor.py#L40-L45

It would be more efficient to always have the RGB values stored in the code. This also makes `colormap` consistent, rather than changing during runtime.


hugovk:
Hmm, the original behaviour is a little strange.

If you access the values from `colormap` directly, you get hex values:

```pycon
>>> from PIL import ImageColor
>>>
>>> ImageColor.colormap["whitesmoke"]
'#f5f5f5'
>>> ImageColor.colormap["cadetblue"]
'#5f9ea0'
```

Fine.

But, as noted, as soon as you call `getrgb()` it replaces the value in `colormap`:

```pycon
>>> ImageColor.getrgb("whitesmoke")
(245, 245, 245)
>>> ImageColor.colormap["whitesmoke"]
(245, 245, 245)
>>> ImageColor.colormap["cadetblue"]
'#5f9ea0'
```

So could this PR be considered a breaking change for anyone accessing `ImageColor.colormap` directly and expecting hex values?

https://grep.app/search?q=ImageColor.colormap

hugovk:
If we started from scratch, I'd avoid changing `colormap` in `getrgb()`, but this is literally 20-year-old code (PIL 1.1.4, May 2003) so I'm cautious about changing it.

Could c) also be considered a breaking change?

Now:

```pycon
>>> from PIL import ImageColor
>>> ImageColor.colormap["whitesmoke"]
'#f5f5f5'
>>> ImageColor.getrgb("whitesmoke")
(245, 245, 245)
>>> ImageColor.colormap["whitesmoke"]
(245, 245, 245)
```

c)

```pycon
>>> from PIL import ImageColor
>>> ImageColor.colormap["whitesmoke"]
'#f5f5f5'
>>> ImageColor.getrgb("whitesmoke")
(245, 245, 245)
>>> ImageColor.colormap["whitesmoke"]
'#f5f5f5'
```

This is arguably less of a problem that a), and maybe people don't rely on the value in `colormap` _after_ calling `getrgb()` at least once?

Pro: not many hits for accessing `colormap` directly: https://grep.app/search?q=ImageColor.colormap

---


If efficiency is a problem, we could use `from functools import lru_cache` and add a `lru_cache` or `@lru_cache(maxsize=None)` decorator to `getrgb()`. 

Of course, this only helps for subsequent calls to the functions.

Now:

```console
❯ python -m timeit -s "from PIL import ImageColor" "ImageColor.getrgb('whitesmoke'); ImageColor.getrgb('whitesmoke');"
1000000 loops, best of 5: 202 nsec per loop
```

This PR:

```console
❯ python -m timeit -s "from PIL import ImageColor" "ImageColor.getrgb('whitesmoke'); ImageColor.getrgb('whitesmoke');"
2000000 loops, best of 5: 175 nsec per loop
```

lru_cache:

```console
❯ python -m timeit -s "from PIL import ImageColor" "ImageColor.getrgb('whitesmoke'); ImageColor.getrgb('whitesmoke');"
5000000 loops, best of 5: 54.1 nsec per loop
```

It could also help `getcolor()`.

Now:

```console
❯ python -m timeit -s "from PIL import ImageColor" "ImageColor.getcolor('whitesmoke', 'RGB'); ImageColor.getcolor('whitesmoke', 'RGB');"
1000000 loops, best of 5: 317 nsec per loop
```

lru_cache:

```console
❯ python -m timeit -s "from PIL import ImageColor" "ImageColor.getcolor('whitesmoke', 'RGB'); ImageColor.getcolor('whitesmoke', 'RGB');"
5000000 loops, best of 5: 73.8 nsec per loop
```

---

So: maybe we do c), possibly with `lru_cache`?

radarhere:
I'm reluctant to add a new caching strategy here as a once-off, or to open the door to further caching strategies without careful consideration, so I'll just close this.

PR #7633, opened 2023-12-22
Add type hints for ImageMode

hugovk:
Helps #2625.

I also checked in strict mode:

```console
❯ mypy --strict src/PIL/ImageMode.py
Success: no issues found in 1 source file
```

We're not ready to enable strict mode for the tox check, and I don't think we need it for the first pass, but good to fix if not too tricky.

### DIFF
--- a/src/PIL/ImageMode.py
+++ b/src/PIL/ImageMode.py
@@ -26,9 +26,9 @@ class ModeDescriptor:
     def __init__(
         self,
         mode: str,
-        bands: str | tuple[str, ...],
+        bands: tuple[str, ...],
         basemode: str,
-        basetype: str | tuple[str, ...],
+        basetype: str,
         typestr: str,
     ) -> None:
         self.mode = mode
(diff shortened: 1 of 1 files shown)


src/PIL/ImageMode.py
```diff
@@ -23,18 +23,25 @@
 class ModeDescriptor:
     """Wrapper for mode strings."""
 
-    def __init__(self, mode, bands, basemode, basetype, typestr):
+    def __init__(
+        self,
+        mode: str,
+        bands: str | tuple[str, ...],
+        basemode: str,
+        basetype: str | tuple[str, ...],
```
radarhere, line 31:
```suggestion
        basetype: str,
```
I don't think `basetype` is ever a tuple.

src/PIL/ImageMode.py
```diff
@@ -23,18 +23,25 @@
 class ModeDescriptor:
     """Wrapper for mode strings."""
 
-    def __init__(self, mode, bands, basemode, basetype, typestr):
+    def __init__(
+        self,
+        mode: str,
+        bands: str | tuple[str, ...],
```
radarhere, line 29:
```suggestion
        bands: tuple[str, ...],
```
I don't think `bands` is ever a string.

radarhere (approved):

PR #7635, opened 2023-12-24
Added type hints to ImageSequence

radarhere:
Helps #2625

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -1181,7 +1181,7 @@ class Image:
 
         return im
 
-    def copy(self):
+    def copy(self) -> Image:
         """
         Copies this image. Use this method if you wish to paste things
         into an image, but still retain the original.
@@ -2450,7 +2450,7 @@ class Image:
         if open_fp:
             fp.close()
 
-    def seek(self, frame):
+    def seek(self, frame) -> Image:
         """
         Seeks to the given frame in this sequence file. If you seek
         beyond the end of the sequence, the method raises an
@@ -2537,7 +2537,7 @@ class Image:
 
         return self._new(self.im.getband(channel))
 
-    def tell(self):
+    def tell(self) -> int:
         """
         Returns the current frame number. See :py:meth:`~PIL.Image.Image.seek`.
 
--- a/src/PIL/ImageSequence.py
+++ b/src/PIL/ImageSequence.py
@@ -16,6 +16,10 @@
 ##
 from __future__ import annotations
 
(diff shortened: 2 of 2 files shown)


hugovk (approved):
Thanks!


---

Shall we add this and `from typing import Self` to Image.py so `mypy --strict src/PIL/ImageSequence.py` passes? 

```diff
diff --git a/src/PIL/Image.py b/src/PIL/Image.py
index 1cb484b852..b121774095 100644
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -1181,7 +1181,7 @@ class Image:
 
         return im
 
-    def copy(self):
+    def copy(self) -> Self:
         """
         Copies this image. Use this method if you wish to paste things
         into an image, but still retain the original.
@@ -2450,7 +2450,7 @@ class Image:
         if open_fp:
             fp.close()
 
-    def seek(self, frame):
+    def seek(self, frame: int) -> None:
         """
         Seeks to the given frame in this sequence file. If you seek
         beyond the end of the sequence, the method raises an
@@ -2537,7 +2537,7 @@ class Image:
 
         return self._new(self.im.getband(channel))
 
-    def tell(self):
+    def tell(self) -> int:
         """
         Returns the current frame number. See :py:meth:`~PIL.Image.Image.seek`.
 
```



---

Wondering if we should have a TypeAlias for `Image.Image` in `Image.py`?

Something like this?

```python
from typing import TypeAlias

ImageType: TypeAlias = Image.Image
```

Or maybe explicit is better?

Anyway, not necessary for this PR.

radarhere:
Testing, I find that `from typing import Self` is only available in Python >= 3.11. Apart from that though, I've made changes to allow `--strict` to pass.

radarhere:
> Wondering if we should have a TypeAlias for Image.Image in Image.py?

It turns out that TypeAlias was only added in Python 3.10 - https://docs.python.org/3/library/typing.html#typing.TypeAlias

hugovk:
We should be able do it instead like `ImageType = Image.Image`:

> For backwards compatibility, type aliases can also be created through simple assignment

https://docs.python.org/3/library/typing.html#type-aliases

radarhere:
Ok, see what you think of https://github.com/python-pillow/Pillow/pull/7639

radarhere:
[Further thought](https://github.com/python-pillow/Pillow/pull/7639#issuecomment-1872303265) decided against the idea of #7639 / a type alias for `Image.Image`.

PR #4996, opened 2020-10-20
JPEG2000: Support decoding subsampled RGB and YCbCr images

nulano:
Fixes #4991 
Fixes #3946

Adds support decoding subsampled RGB and YCbCr JPEG2000 images.
I left the grayscale formats as-is as I was a bit confused by the code and did not find test files for these formats.

Test images are from https://github.com/uclouvain/openjpeg-data, but these are huge and have an unusual license, so they are in https://github.com/python-pillow/pillow-depends/pull/35. `subsampling_1.jp2` and `zoo1.jp2` are YCbCr 4:2:0 and `subsampling_2.jp2` and `zoo2.jp2` are something like RGB 2:2:0 (2x2 subsampling for all 3 channels). This list is based on [this OpenJPEG test](https://github.com/uclouvain/openjpeg/blob/491299eb073c80602ca8ee333303a31a581d3fa8/tests/conformance/CMakeLists.txt#L568-L571).

~~Until the pillow-depends PR is merged, see successful test in another branch: https://github.com/nulano/Pillow/runs/1279251514?check_suite_focus=true#step:10:425~~

All four images have `color_space = OPJ_CLRSPC_UNSPECIFIED`, and based on https://github.com/uclouvain/openjpeg/issues/570 I think this may be a bug, but I did not look into fixing this.

PR #7636, opened 2023-12-24
Document JPEG 2000 support for writing YCbCr and reading subsampled color

bgilbert:
Read support for subsampled RGB and YCbCr color channels was added in 4f4c3b34f8 and not documented at the time.  Write support for YCbCr appears to date to 61fb89ec54, the original commit.

Retain the existing language about YCbCr input conversion to RGB, even though it's not completely correct.  OpenJPEG through 2.5.0 doesn't set `color_space` in `opj_read_header()`, so we end up in our `OPJ_CLRSPC_UNSPECIFIED` fallback path, which guesses sRGB if there's no component subsampling.  This means we currently can't round-trip YCbCr via JPEG 2000.  The next OpenJPEG release will fix this (https://github.com/uclouvain/openjpeg/pull/1463) so leave the docs as is.

Also fix typo: `.j2p` -> `.jp2`.

### DIFF
--- a/docs/handbook/image-file-formats.rst
+++ b/docs/handbook/image-file-formats.rst
@@ -552,12 +552,13 @@ JPEG 2000
 
 .. versionadded:: 2.4.0
 
-Pillow reads and writes JPEG 2000 files containing ``L``, ``LA``, ``RGB`` or
-``RGBA`` data.  It can also read files containing ``YCbCr`` data, which it
-converts on read into ``RGB`` or ``RGBA`` depending on whether or not there is
-an alpha channel.  Pillow supports JPEG 2000 raw codestreams (``.j2k`` files),
-as well as boxed JPEG 2000 files (``.j2p`` or ``.jpx`` files).  Pillow does
-*not* support files whose components have different sampling frequencies.
+Pillow reads and writes JPEG 2000 files containing ``L``, ``LA``, ``RGB``,
+``RGBA``, or ``YCbCr`` data.  When reading, ``YCbCr`` data is converted to
+``RGB`` or ``RGBA`` depending on whether or not there is an alpha channel.
+Beginning with version 8.3.0, Pillow can read (but not write) ``RGB``,
+``RGBA``, and ``YCbCr`` images with subsampled components.  Pillow supports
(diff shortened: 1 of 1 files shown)


radarhere:
@nulano since you wrote https://github.com/python-pillow/Pillow/commit/4f4c3b34f8685bc9e16f596d093f8b01f8b2e1b6, did you have any thoughts on this?

docs/handbook/image-file-formats.rst
```diff
@@ -552,12 +552,13 @@ JPEG 2000
...
-``RGBA`` data.  It can also read files containing ``YCbCr`` data, which it
-converts on read into ``RGB`` or ``RGBA`` depending on whether or not there is
-an alpha channel.  Pillow supports JPEG 2000 raw codestreams (``.j2k`` files),
-as well as boxed JPEG 2000 files (``.j2p`` or ``.jpx`` files).  Pillow does
-*not* support files whose components have different sampling frequencies.
+Pillow reads and writes JPEG 2000 files containing ``L``, ``LA``, ``RGB``,
+``RGBA``, or ``YCbCr`` data.  When reading, ``YCbCr`` data is converted to
+``RGB`` or ``RGBA`` depending on whether or not there is an alpha channel.
+Beginning with version 8.3.0, Pillow can read ``RGB``, ``RGBA``, and
+``YCbCr`` images whose components have different sampling frequencies.
```
nulano, line 559: Two notes:

* I'm not sure this needs to be stated in the documentation at all, since it is what I would expect by default (I didn't see this text when writing #4996 and thought of it as a bugfix, not a new feature).
* AFAIK, Pillow<8.3.0 cannot read images with any subsampling, not just if it is different for some components (although I couldn't find any images where all components are subsampled to test this).

bgilbert: Well, it's a polite way to say "before 8.3.0, Pillow didn't support subsampling".  Also, there are still limitations that should be documented: subsampled grayscale images aren't supported (I guess those are unlikely in practice) and Pillow can't subsample on write.

Updated to fix the implied historical behavior and to make the write limitation explicit.

nulano (approved):

radarhere:
Thanks for your thoughts @nulano

PR #7637, opened 2023-12-25
Moved __future__ import to beginning of file

radarhere:
oss-fuzz coverage has started failing after #7631 - https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=65341&sort=-opened&can=1&q=proj%3Apillow
> Syntax error in /src/Pillow/Tests/oss-fuzz/fuzz_font.py
>   File "/src/Pillow/Tests/oss-fuzz/fuzz_font.py", line 27
>      from \_\_future__ import annotations
>      ^
>  SyntaxError: from \_\_future__ imports must occur at the beginning of the file

### DIFF
--- a/Tests/oss-fuzz/fuzz_font.py
+++ b/Tests/oss-fuzz/fuzz_font.py
@@ -1,5 +1,7 @@
 #!/usr/bin/python3
 
+from __future__ import annotations
+
 # Copyright 2020 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


hugovk:
I'm surprised the linters didn't catch it. Do you know what oss-fuzz is "Analyzing" when it fails and if other files will trigger it? Or we can merge and see the next build.

hugovk (approved):

radarhere:
While the initial report was about Tests/oss-fuzz/fuzz_font.py, my attempt to replicate [gave an error about Tests/oss-fuzz/fuzz_pillow.py](https://github.com/radarhere/oss-fuzz/actions/runs/7327008250/job/19953332942#step:7:10016), so I've updated that as well.

My attempts to convince my oss-fuzz fork to look at this PR's branch aren't working however, so no, I can't guarantee that there aren't other fixes that need to be made.

radarhere:
#7649 was the solution to this in the end.

PR #7638, opened 2023-12-26
Added TarIO type hints

radarhere:
Helps #2625

I took the `__exit__` argument type hints from https://adamj.eu/tech/2021/07/04/python-type-hints-how-to-type-a-context-manager/#class-based-context-managers


hugovk (approved):
Thanks!

And thanks @adamchainz for the blogpost!

PR #7639, opened 2023-12-26
Added ImageType as alias for Image.Image

### LINKED ISSUE #6676: Typing difficulty because `Image` is a module and also a class
### What did you do?
```py
from PIL import Image

def foo(img: Image) -> None:
	...
```
and got `Module "PIL.Image" is not valid as a typemypy(error)`.

functions such as `Image.open` are classmethod of class `Image`.

### What actually happened?
`Image` was actually an module, so for typing you need to type `Image.Image`, such as
```py
from PIL import Image

def foo(img: Image.Image) -> None:
	...
```

* OS: Ubuntu 22.04
* Python: Python 3.11.0rc2
* Pillow: Name: Pillow-SIMD 9.0.0.post1

radarhere:
Resolves #6676. Also suggested in https://github.com/python-pillow/Pillow/pull/7635#pullrequestreview-1795553042

When using the `Image` class as a type hint, you must currently use `from PIL import Image; Image.Image`.

This PR allows for `from PIL import ImageType; ImageType`.

### DIFF
--- a/docs/reference/Image.rst
+++ b/docs/reference/Image.rst
@@ -10,7 +10,8 @@ functions, including functions to load images from files, and to create new
 images.
 
 To assist with type hints, ``ImageType`` has been added a simpler way to refer
-to the class within the module. ::
+to the class within the module. In the following example, the two type
+annotations are equivalent::
 
     from PIL import Image, ImageType
     im1: Image.Image = Image.new("RGB", (1, 1))
--- a/docs/reference/Image.rst
+++ b/docs/reference/Image.rst
@@ -9,6 +9,13 @@ used to represent a PIL image. The module also provides a number of factory
 functions, including functions to load images from files, and to create new
 images.
 
+To assist with type hints, ``ImageType`` has been added a simpler way to refer
+to the class within the module. ::
+
+    from PIL import Image, ImageType
+    im1: Image.Image = Image.new("RGB", (1, 1))
+    im2: ImageType = Image.new("RGB", (1, 1))
+
 Examples
 --------
 
--- a/src/PIL/__init__.py
+++ b/src/PIL/__init__.py
@@ -83,3 +83,11 @@ class UnidentifiedImageError(OSError):
     """
 
     pass
+
+
+def __getattr__(name: str):
+    if name == "ImageType":
+        from . import Image
+
+        return Image.Image
+    raise AttributeError
--- a/src/PIL/__init__.py
+++ b/src/PIL/__init__.py
@@ -90,4 +90,5 @@ def __getattr__(name: str):
         from . import Image
 
         return Image.Image
-    raise AttributeError
+    msg = f"module 'PIL' has no attribute '{name}'"
+    raise AttributeError(msg)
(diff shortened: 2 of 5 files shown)


src/PIL/__init__.py
```diff
@@ -83,3 +83,11 @@ class UnidentifiedImageError(OSError):
     """
 
     pass
+
+
+def __getattr__(name: str):
```
hugovk, line 88: What do you think about putting `ImageType` in `Image.py` to avoid this? In other files, would importing from `Image` instead of from `PIL` (or `.`) be tricky?

It's just a thought, I'm fine with this.

hugovk (dismissed):

src/PIL/__init__.py
nulano: Moving it to `Image.py` would make it easier to find in the docs ("Reference > Image module" instead of the current "Reference > PIL Package (autodoc of remaining modules)").

That said, I have no preference on putting it in either file either.

radarhere: So, I was confused by your initial suggestion of [`ImageType: TypeAlias = Image.Image` in Image.py](https://github.com/python-pillow/Pillow/pull/7635#pullrequestreview-1795553042), because if it's in Image.py, then it's not `ImageType = Image.Image`, it's just `ImageType = Image`, right? It's https://github.com/radarhere/Pillow/commit/75afb580d71146fdb8c3294d4939116214732635, yeah?

I think the users in #6676 basically don't like that `Image.Image` exists as a class within a module, and would like both to be at the package level. I think this version still not that, but closer to what they want.

I'm happy to go with either this version or the `ImageType = Image` version, since both are backwards compatible - or neither. If the goal of the type alias is brevity though, then this is the shortest version?

hugovk: Yep, this PR is better than https://github.com/radarhere/Pillow/commit/75afb580d71146fdb8c3294d4939116214732635, thanks for the illustration :)

radarhere:
Ok, I've added it to the docs - https://pillow--7639.org.readthedocs.build/en/7639/reference/Image.html

Screenshot 2023-12-27 at 10 03 33 am

docs/reference/Image.rst
```diff
@@ -9,6 +9,13 @@ used to represent a PIL image. The module also provides a number of factory
 functions, including functions to load images from files, and to create new
 images.
 
+To assist with type hints, ``ImageType`` has been added a simpler way to refer
+to the class within the module. ::
```
nulano, line 13: Perhaps explicitly mention that `Image.Image` and `ImageType` are equivalent?
```suggestion
to the class within the module. In the following example,
the two type annotations are equivalent::
```

hugovk:
I know I suggested this in the first place, but after some thought, perhaps it's not so useful after all. 

`Image.Image` vs `ImageType` is adding abstraction on top of something which isn't particularly complex in the first place, only save two chars, at the cost of another layer of indirection.

I can see aliases may be more useful for more complex things, such as tuples of ints for RGB etc, but I think we should pass on this one.

PR #7642, opened 2023-12-27
Add type hints to _util

nulano:
I wanted to start adding type hints for `PIL.ImageFont`, but it is still in the list of excluded files.

The errors in `ImageFont` can be fixed by adding type hints for `PIL._util` and an empty `_imagingft.pyi` file.
There are three issues in `ImageFont`:

* > error: Module "PIL" has no attribute "_imagingft"

  Solved by adding empty `_imagingft.pyi` file, which reveals another error:

  > error: Incompatible types in assignment (expression has type "DeferredError", variable has type Module)
  
  Solved by adding `_util.DeferredError.new(ex) -> Any`.

* > error: Missing type parameters for generic type "IO"

  I think this should be `typing.BinaryIO`, https://github.com/python-pillow/Pillow/pull/7578#discussion_r1413732510.
  Note that [typeshed](https://github.com/python/typeshed/blob/ed201da4205b8dc6a2bc200187f61b3251c1b2cd/stubs/Pillow/PIL/ImageFont.pyi#L32) has `_typeshed.SupportsRead[bytes]`.
* several more errors due to failed propagation of the `_util.is_path` type guard
  
  Solved by adding type hints for `_util.is_path`. This requires the use of `typing.TypeGuard` which is Python 3.10+ only. For Python<=3.9, it is taken from `typing_extensions`. See https://mypy.readthedocs.io/en/stable/runtime_troubles.html#using-new-additions-to-the-typing-module for more information.

I've also included fixes for errors in `ImageFile` (incorrect `ImageFile._Tile.args` annotation) and `ImageCms` (missing `_imagingcms.pyi` file), as they are quite simple.


nulano:
Thinking about this some more, the `typing_extensions` dependency can be made optional with something like:

```python
if sys.version_info >= (3, 10):
    from typing import TypeGuard

    PathGuard = TypeGuard[bytes | str | Path]
elif typing.TYPE_CHECKING:
    from typing_extensions import TypeGuard

    PathGuard = TypeGuard[bytes | str | Path]
else:
    PathGuard = bool


def is_path(f) -> PathGuard:
    return isinstance(f, (bytes, str, Path))
```

However, I'm not sure if this is worth doing to avoid the hard dependency, and if so, how the dependency should be declared in `pyproject.toml`.

hugovk:
Other than for testing, we've managed to avoid Python dependencies so far.

I doubt adding `typing_extensions` would be a big problem, but if we want to keep zero dependencies, we could move this `if`/`elif`/`else` block into a `_typing.py` or `_compat.py` module.

Another option is to leave `typing_extensions` out, like:

```python
if sys.version_info >= (3, 10):
    from typing import TypeGuard

    PathGuard = TypeGuard[bytes | str | Path]
else:
    PathGuard = bool
```

And use new Python for type checking; we use 3.12 on the CI. This could also be tidied into an internal module.

nulano:
OK, after a lot of trial and error, I've managed to make `typing_extensions` an optional dependency.
As it turns out, my initial suggestion was not functional because mypy doesn't understand a `TypeGuard` when it is provided as a `TypeAlias`. The only way I could convince it to accept the type guard in `ImageFont` was to (in)directly import `TypeGuard` from either `typing` or `typing_extensions`, and provide a fallback class when neither is available (which mypy will never see).

hugovk (approved):

src/PIL/_util.py
```diff
@@ -2,20 +2,31 @@
...
-    def __init__(self, ex):
+    def __init__(self, ex: BaseException):
         self.ex = ex
 
-    def __getattr__(self, elt):
+    def __getattr__(self, elt: str) -> NoReturn:
         raise self.ex
+
+    @staticmethod
+    def new(ex: BaseException) -> Any:
```
radarhere, line 27:
```suggestion
    def new(ex: BaseException) -> DeferredError:
```

PR #7643, opened 2023-12-27
Added type hints to FontFile and subclasses

radarhere:
Adds type hints to FontFile, BdfFontFile and PcfFontFile.

In doing so,
- I found that https://github.com/python-pillow/Pillow/blob/1e8bad805ec763f4bf1c265cf26a3deee4fec071/src/PIL/FontFile.py#L68
is the only time that `FontFile.compile()` returns something other than `None`. I've changed that to be consistent.
- I changed the following error
```pycon
>>> from PIL import FontFile
>>> font = FontFile.FontFile()
>>> font.save("out")
  File "PIL/FontFile.py", line 97, in save
    self.bitmap.save(os.path.splitext(filename)[0] + ".pbm", "PNG")
AttributeError: 'NoneType' object has no attribute 'save'
```
to "ValueError: No bitmap created" instead.

### DIFF
--- a/src/PIL/PcfFontFile.py
+++ b/src/PIL/PcfFontFile.py
@@ -42,7 +42,7 @@ PCF_SWIDTHS = 1 << 6
 PCF_GLYPH_NAMES = 1 << 7
 PCF_BDF_ACCELERATORS = 1 << 8
 
-BYTES_PER_ROW = [
+BYTES_PER_ROW: list[Callable[[int], int]] = [
     lambda bits: ((bits + 7) >> 3),
     lambda bits: ((bits + 15) >> 3) & ~1,
     lambda bits: ((bits + 31) >> 3) & ~3,
@@ -210,7 +210,7 @@ class PcfFontFile(FontFile.FontFile):
 
         data = fp.read(bitmapsize)
 
-        pad: Callable[[int], int] = BYTES_PER_ROW[padindex]
+        pad = BYTES_PER_ROW[padindex]
         mode = "1;R"
         if bitorder:
             mode = "1"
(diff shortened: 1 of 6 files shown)


hugovk (approved):
Great to find and fix some bugs along the way!

src/PIL/PcfFontFile.py
```diff
@@ -207,7 +210,7 @@ def _load_bitmaps(self, metrics):
 
         data = fp.read(bitmapsize)
 
-        pad = BYTES_PER_ROW[padindex]
+        pad: Callable[[int], int] = BYTES_PER_ROW[padindex]
```
nulano, line 213: I don't see why this one is needed?

Checking with `typing.reveal_type`, I see that it can be inferred if you add the following at the top of the file:
```python
BYTES_PER_ROW: list[Callable[[int], int]] = [ ...
```

radarhere: Without it, `python3 -m mypy --strict src/PIL/PcfFontFile.py` gives
```
src/PIL/PcfFontFile.py:223: error: Call to untyped function (unknown) in typed context  [no-untyped-call]
                    Image.frombytes("1", (xsize, ysize), data[b:e], "raw", mode, pad(xsize))
                                                                                 ^~~~~~~~~~
Found 1 error in 1 file (checked 1 source file)
```

radarhere: Your suggestion also works however, so I've pushed a commit.

nulano: Ah, I was missing the `--strict`.

nulano (approved):

PR #7644, opened 2023-12-27
Update UnidentifiedImageError and __version__ imports

radarhere:
There are a few instances where `UnidentifiedImageError` is imported from `Image` rather than `.`, and `__version__` is imported from `.` rather than `_version`.

It would seem better to import items from where they are located, instead of relying on them being imported in another place.


src/PIL/Image.py
```diff
@@ -54,11 +54,11 @@
     ImageMode,
     TiffTags,
     UnidentifiedImageError,
-    __version__,
     _plugins,
 )
 from ._binary import i32le, o32be, o32le
 from ._util import DeferredError, is_path
+from ._version import __version__
```
homm, line 61: Does `_version` exist solely to simplify version parsing in `setup.py`? In my opinion, the actual module where the version should be defined is `PIL`.

radarhere: It was in setup.py, __init__.py, _imaging.c and appveyor.yml, and #2517 chose to create a new file for it.

It might not be worth making any changes at the moment though, as #7627 has a different plan for the version number.

radarhere: Ok, I've pushed a commit so that `__version__` is imported from the base instead.

hugovk:
Thanks!

src/PIL/Image.py
radarhere: I've created https://github.com/python-pillow/Pillow/pull/7874 as a follow up to this, moving the definition of `__version__` to `__init__.py`.

wiredfool: The reason we did this was to be able to import the version without a buildable PIL, so that it could be added to the C build automatically to facilitate the version checks there.

`_version.py` is a private implementation detail, everyone but setup.py should be referencing the published `PIL.__version__` or whatever we're calling it now.

PR #7645, opened 2023-12-27
Simplify FreeTypeFont.render

nulano:

 * Call `Image._decompression_bomb_check` directly in `fill` rather than duplicating its code. If an error is raised, it is propagated via the `if (image == NULL)` branch.
 * On error, delete the Python `image` object by releasing the only reference to it (causing Python to call its `tp_dealloc` slot, i.e. `_dealloc` in `_imaging.c`) instead of just freeing the memory it points to (added in #7218).
 * Avoid nonlocal variables (to simplify adding type hints in a future PR by avoiding the need for a `typing.cast` or `assert size is not None`).

This partially reverts changes from #7246, which was connected to an oss-fuzz error, but I do not see any added tests or reproduction steps. IIUC, the original error was a memory leak, which I avoid in a different way in this PR:

The original code (before #7246) returned `image` via `O` format (returning a new reference) and did not `Py_DECREF` (causing a memory leak).
#7246 instead returns a reference via a nonlocal Python variable and added `Py_DECREF`.
In this PR I have removed the nonlocal variable and return the existing reference via `N` format.
See https://docs.python.org/3/c-api/arg.html#c.Py_BuildValue for the difference between `O` and `N`.

### DIFF
--- a/src/PIL/ImageFont.py
+++ b/src/PIL/ImageFont.py
@@ -582,22 +582,13 @@ class FreeTypeFont:
         _string_length_check(text)
         if start is None:
             start = (0, 0)
-        im = None
-        size = None
 
         def fill(width, height):
-            nonlocal im, size
-
             size = (width, height)
-            if Image.MAX_IMAGE_PIXELS is not None:
-                pixels = max(1, width) * max(1, height)
-                if pixels > 2 * Image.MAX_IMAGE_PIXELS:
-                    return
+            Image._decompression_bomb_check(size)
+            return Image.core.fill("RGBA" if mode == "RGBA" else "L", size)
 
-            im = Image.core.fill("RGBA" if mode == "RGBA" else "L", size)
-            return im
-
-        offset = self.font.render(
+        return self.font.render(
             text,
             fill,
             mode,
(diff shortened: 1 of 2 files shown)


radarhere (approved):
I had imagined that raising a DecompressionBombError during a Python callback from C was somehow bad, but our test suite triggers that scenario with your branch, and it is fine.

Presuming that the memory management in fine, I approve - but it is probably not the best idea to adjust code that fixed an oss-fuzz problem less than a week before release, so I suggest holding off on this for now

nulano:
> it is probably not the best idea to adjust code that fixed an oss-fuzz problem less than a week before release, so I suggest holding off on this for now

Sure, that sounds reasonable.
There are many other things blocking a PR adding type hints to `ImageFont.py` anyway.

src/PIL/ImageFont.py
```diff
@@ -582,22 +582,13 @@ def getmask2(
...
-                if pixels > 2 * Image.MAX_IMAGE_PIXELS:
-                    return
-
-            im = Image.core.fill("RGBA" if mode == "RGBA" else "L", size)
-            return im
+            Image._decompression_bomb_check(size)
+            return Image.core.fill("RGBA" if mode == "RGBA" else "L", size)
 
-        offset = self.font.render(
+        return self.font.render(
```
hugovk, line 591: Before we returned `(im, offset)`, but now just `offset`, is this a breaking change?

nulano: This changes `self.font.render` to return `(im, offset)` instead of just `offset`.

PR #7649, opened 2023-12-29
Removed __future__ import from fuzz_font.py and fuzz_pillow.py

radarhere:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=65341 is still open, complaining that `from __future__ imports must occur at the beginning of the file`, despite #7637 moving the imports.

I asked about this at https://github.com/google/oss-fuzz/issues/11415, and was told that oss-fuzz [prepends code to the files](https://github.com/google/oss-fuzz/blob/55650963e17693deedfe2ed86df2544cf124ab8b/infra/base-images/base-builder/compile_python_fuzzer#L56-L75).

I think the solution here is to just remove `from __future__ import annotations` from fuzz_font.py and fuzz_pillow.py. I've tested it in my oss-fuzz fork, and [it fails without this](https://github.com/radarhere/oss-fuzz/actions/runs/7352047687/job/20016249063), but [passes with.](https://github.com/radarhere/oss-fuzz/actions/runs/7352388832)

### DIFF
--- a/Tests/oss-fuzz/fuzz_font.py
+++ b/Tests/oss-fuzz/fuzz_font.py
@@ -1,7 +1,5 @@
 #!/usr/bin/python3
 
-from __future__ import annotations
-
 # Copyright 2020 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


hugovk:
I think the short/medium term is to remove them.

A better fix would be if they inserted after any `from __future__ import ...` statements. We, they or others might want to type those files in the future.

But then these are test files so it's not so important to type them, there's not much in these files, and if we did type we could use non-future stuff.

Thanks for the following up!

radarhere:
This worked. https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=65341 is now closed

Issue #4669, opened 2020-06-05
Saving TIFF in chunks


DexterHill0:
### What did you do?
I need to save to save huge image files (approx. 819200x460800 RGBA). This is too much for anyone's RAM so I have to save it in chunks from disk. I start by saving the array to an HDF5 file. I then loop over the array in large steps, and parse a slice of the array into .fromarray(). I then save this to a tiff file. Once it loops again, it will add more to the tiff file and so on.

### What did you expect to happen?
It should create an tiff image that is very large.

### What actually happened?
It errored out while saving, giving me the error:
```pytb
ile "c:\Users\Dexter\Desktop\workspace\test.py", line 68, in test
    a.save("ff.tiff", format="tiff", quality=1)
  File "C:\Users\Dexter\AppData\Local\Programs\Python\Python38\lib\site-packages\PIL\Image.py", line 2134, in save
    save_handler(self, fp, filename)
  File "C:\Users\Dexter\AppData\Local\Programs\Python\Python38\lib\site-packages\PIL\TiffImagePlugin.py", line 1629, in _save
    offset = ifd.save(fp)
  File "C:\Users\Dexter\AppData\Local\Programs\Python\Python38\lib\site-packages\PIL\TiffImagePlugin.py", line 865, in save
    result = self.tobytes(offset)
  File "C:\Users\Dexter\AppData\Local\Programs\Python\Python38\lib\site-packages\PIL\TiffImagePlugin.py", line 808, in tobytes
    data = self._write_dispatch[typ](self, *values)
  File "C:\Users\Dexter\AppData\Local\Programs\Python\Python38\lib\site-packages\PIL\TiffImagePlugin.py", line 644, in <lambda>
    b"".join(self._pack(fmt, value) for value in values)
  File "C:\Users\Dexter\AppData\Local\Programs\Python\Python38\lib\site-packages\PIL\TiffImagePlugin.py", line 644, in <genexpr>
    b"".join(self._pack(fmt, value) for value in values)
  File "C:\Users\Dexter\AppData\Local\Programs\Python\Python38\lib\site-packages\PIL\TiffImagePlugin.py", line 611, in _pack
    return struct.pack(self._endian + fmt, *values)
struct.error: argument out of range
```
### What are your OS, Python and Pillow versions?

* OS:  Windows 10 V. 1909
* Python:  3.8.1
* Pillow:  Latest (7.1.2)

```python
def test():
    f = h5py.File("test.hdf5", "w")
[...]

PR #7650, opened 2023-12-29
Allow uncompressed TIFF images to be saved in chunks

radarhere:
Helps #4669

The issue would like to save a large TIFF image, but to do so in chunks, read then loading the entire image into memory at once.

If libtiff isn't used when saving, then the data is saved [simply using the "raw" encoder](https://github.com/python-pillow/Pillow/blob/78b96c037597b56b125257f568de04b4d8921118/src/PIL/TiffImagePlugin.py#L1859-L1861).

So then what if one just appended data to a created image? The header would still need to change so that the image knew about the new size.
```python
from PIL import Image, TiffImagePlugin

with open("out.tiff", "wb") as fp:
  for i, chunk in enumerate([
    im.crop((0, 0, 128, 32)),
    im.crop((0, 32, 128, 64)),
    im.crop((0, 64, 128, 96)),
    im.crop((0, 96, 128, 128)),
  ]):
    if i == 0:
      chunk.save(fp, "TIFF", tiffinfo={
        TiffImagePlugin.IMAGEWIDTH: 128,
        TiffImagePlugin.IMAGELENGTH: 128
      })
    else:
      fp.write(chunk.tobytes())
```
The image is the right size, but black.

When calculating `stride`, `ifd[ROWSPERSTRIP]` and `ifd[STRIPBYTECOUNTS]`, TiffImagePlugin currently [uses the image size](https://github.com/python-pillow/Pillow/blob/78b96c037597b56b125257f568de04b4d8921118/src/PIL/TiffImagePlugin.py#L1707-L1726). This PR updates it to use `ifd[IMAGEWIDTH]` and `ifd[IMAGELENGTH]`, and then the saved image appears correctly.

Apart from this scenario, `ifd[IMAGEWIDTH]` and `ifd[IMAGELENGTH]` are supposed to be the same as image width and height. Two of the tests had to be updated when saving TIFF images using existing info, but I could argue that they were already incorrect in main for saving images with an incorrect `ifd[IMAGEWIDTH]`.


hugovk (approved):
Needs a conflict resolving.

PR #7656, opened 2023-12-29
Add type hints to PIL.ContainerIO

nulano:
The two ignore comments are required because mypy doesn't know that `"b" in self.fh.mode` if and only if `self` is a `ContainerIO[bytes]` rather than a `ContainerIO[str]`.

### DIFF
--- a/src/PIL/ContainerIO.py
+++ b/src/PIL/ContainerIO.py
@@ -16,15 +16,16 @@
 from __future__ import annotations
 
 import io
+from typing import IO, AnyStr, Generic, Literal
 
 
-class ContainerIO:
+class ContainerIO(Generic[AnyStr]):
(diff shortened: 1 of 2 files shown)


hugovk (approved):

src/PIL/ContainerIO.py
```diff
@@ -16,23 +16,24 @@
 from __future__ import annotations
 
 import io
+from typing import IO, AnyStr, Generic, Literal
```
radarhere, line 19:
```suggestion
from typing import AnyStr, Generic, IO, Literal
```

radarhere: Ok, nevermind, pre-commit disagrees with me.

PR #7658, opened 2023-12-30
Fix editable installation with custom build backend and configuration options

nulano:
#7171 added a custom build backend to translate `-C feature=disable` style flags to the old `--disable-feature` style flags by prepending the build command (typically `bdist_wheel`) with a `build_ext --disable-feature` command which builds the extension modules before being added to a wheel.

It was later noted that this doesn't support editable installs: https://github.com/python-pillow/Pillow/pull/7171#discussion_r1435463654
However, the obvious implementation does not work on Windows: https://github.com/python-pillow/Pillow/pull/7171#discussion_r1435635957

Investigating, I found that editable installs use a private setuptools command `editable_wheel`: https://github.com/pypa/setuptools/blob/0e2229d7e7d50e4c5f363cb8435a78b03a957dfc/setuptools/build_meta.py#L434
It seems that this command does not reuse the result of the added `build_ext` step, and instead triggers `build_ext` a second time. For some reason, the second `build_ext` does not get initialized properly on Windows and fails to set some of the compiler flags (it does work properly if it is only invoked once, i.e. if no `-C feature=disable` style flags were provided).

Therefore, in this PR I instead pass the configuration flags before the build command and store them before calling `setup(...)`, then use them to set the defaults in `pil_build_ext.initialize_options`. There has been some discussion in https://github.com/pypa/setuptools/issues/3896 about how this could be done within setuptools itself, but there does not appear to be a solution yet.

~~Two issues to consider:~~
* ~~Could the option `-C` conflict with some setuptools options (currently, passing `-C` to `setup.py` exits with an error)?~~
* ~~Perhaps this should wait until after the release given that it changes `setup.py`.~~


webknjaz:
> * Could the option `-C` conflict with some setuptools options (currently, passing `-C` to `setup.py` exits with an error)?

One way of solving this could be prefixing the options. But also, you could declare that any underlying setuptools options aren't officially supported unless documented. Then, you'd be in control of how the build is recommended to be invoked. After all, most of the builds are to be made by project's own CI while most of the rest are done by the downstreams. Most of the end-users should avoid using sdist. And the contributors would need to follow the contrib docs.

I've been experimenting with having dedicated changelog sections for the downstream packagers and things that are contributor-facing: https://yarl.aio-libs.org/en/latest/changes/. Maybe, implementing a similar approach could benefit Pillow too.

> translate `-C feature=disable` style flags to the old `--disable-feature` style

I came to a conclusion in other projects that having true/false values might communicate the intent better, especially since this is a new API. Think about this option?

nulano:
> One way of solving this could be prefixing the options. But also, you could declare that any underlying setuptools options aren't officially supported unless documented. 

I mean that I chose the prefix `-C` for passing the configuration from `backend.py` to `setup.py`, but it is not clear whether this might be broken by a future setuptools version (i.e. setuptools adding its own `-C` option at the start of the build command in `build_meta`). I also considered the prefix `--pillow-private-config` (or something similar), but wanted to wait for opinions from others since I don't expect this to make it into the 10.2.0 release on Tuesday.

> I came to a conclusion in other projects that having true/false values might communicate the intent better, especially since this is a new API. Think about this option?

This API was added in #7171 and released in Pillow 10.0.0: https://pillow.readthedocs.io/en/stable/installation.html#build-options
Perhaps this could be changed, but at least I don't see the benefit in this case given that it has already been included in a release.

Also note that there are three options for some features: `disable`, `enable`, `vendor`.

To be clear, I would consider the flags added to `setup.py` by this PR to be private and not intended for direct use; installing (and building) with pip is what is recommended.

> I've been experimenting with having dedicated changelog sections for the downstream packagers and things that are contributor-facing: https://yarl.aio-libs.org/en/latest/changes/. Maybe, implementing a similar approach could benefit Pillow too.

That looks like a good idea, but again, this PR is not intended to change anything, only to fix editable installation (which has already been useful for me while writing #7659).

webknjaz:
That confused me because the popular build frontends (pypa/build and pip) have `--config-setting` (or `-C` for short) specifically for supplying arbitrary config settings to the backends.

nulano:
> That confused me because the popular build frontends (pypa/build and pip) have --config-setting (or -C for short) specifically for supplying arbitrary config settings to the backends.

Yeah, that is why I chose `-C` for passing those settings from the custom backend to setuptools-based `setup.py`, but I do see how it can be confusing.

nulano:
I've renamed the (private) parameter to `--pillow-configuration` to avoid confusion or potential conflicts with future versions of setuptools.

docs/installation.rst
```diff
@@ -390,9 +390,10 @@ After navigating to the Pillow directory, run::
 Build Options
 """""""""""""
 
-* Environment variable: ``MAX_CONCURRENCY=n``. Pillow can use
-  multiprocessing to build the extension. Setting ``MAX_CONCURRENCY``
-  sets the number of CPUs to use, or can disable parallel building by
+* Config setting: ``-C parallel=n``. Can also be given
```
webknjaz, line 393: Why not use the same naming pattern for settings and env vars?

winbuild/build.rst
```diff
@@ -87,11 +87,18 @@ are set by running ``winbuild\build\build_env.cmd`` and install Pillow with pip:
     winbuild\build\build_env.cmd
     python.exe -m pip install -v -C raqm=vendor -C fribidi=vendor .
 
+You can also install Pillow in `editable mode`_::
+
+    winbuild\build\build_env.cmd
+    python.exe -m pip install -v -C raqm=vendor -C fribidi=vendor -e .
+
 To build a wheel instead, run::
```
webknjaz, line 95:
```suggestion
To build a non-editable wheel instead, run::
```

(the editable hook does produce special "editable" wheels too, under the hood)

winbuild/build.rst
```diff
@@ -87,11 +87,18 @@ are set by running ``winbuild\build\build_env.cmd`` and install Pillow with pip:
     winbuild\build\build_env.cmd
     python.exe -m pip install -v -C raqm=vendor -C fribidi=vendor .
 
+You can also install Pillow in `editable mode`_::
+
+    winbuild\build\build_env.cmd
+    python.exe -m pip install -v -C raqm=vendor -C fribidi=vendor -e .
```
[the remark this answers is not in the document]
nulano: I can't find any recommendations about the order in the pip documentation, so I'll take your word for it. Testing locally, I find that either order works the same with a single package. I've created #7686 for your suggestion.

winbuild/build.rst
nulano: I would expect anyone who understands the distinction between regular and editable wheels can figure that out from the command, so I'm not sure how helpful this change would be. Personally, I think "non-editable" is more confusing than helpful.

Would "binary" make sense here?
```suggestion
To build a binary wheel instead, run::
```

docs/installation.rst
nulano: `MAX_CONCURRENCY` was added in #721 (almost 10 years ago).

`--parallel` is a distutils/setuptools flag added in Python 3.5; Pillow support added in https://github.com/python-pillow/Pillow/commit/64bce1a583ffdbf1811263ae7dce3088a669d7c7 (over 5 years ago).

So I think it makes sense to keep those names for consistency.

winbuild/build.rst
```diff
@@ -87,11 +87,18 @@ are set by running ``winbuild\build\build_env.cmd`` and install Pillow with pip:
...
+
 To build a wheel instead, run::
 
     winbuild\build\build_env.cmd
     python.exe -m pip wheel -v -C raqm=vendor -C fribidi=vendor .
 
+.. _editable mode: https://setuptools.pypa.io/en/latest/userguide/development_mode.html
```
radarhere, line 100:
```suggestion
.. _editable mode: https://setuptools.pypa.io/en/stable/userguide/development_mode.html
```
Would it be more practically helpful to link to 'stable'?

_custom_build/backend.py
```diff
@@ -54,5 +27,15 @@ def build_wheel(
...
+        self.config_settings = config_settings
+        return super().build_editable(
+            wheel_directory, config_settings, metadata_directory
+        )
+
 
-build_wheel = _CustomBuildMetaBackend().build_wheel
+_backend = _CustomBuildMetaBackend()
+build_wheel = _backend.build_wheel
+build_editable = _backend.build_editable
```
radarhere, line 41:
```suggestion
if not getattr(_backend, "LEGACY_EDITABLE", False):
    build_editable = _backend.build_editable
```
Looking at https://github.com/pypa/setuptools/blob/6ee23bf0579c52e1cbe7c97fc20fd085ff2a25c7/setuptools/build_meta.py#L430-L437, I think perhaps this condition should be added, because there may not be a `super()` method to call.

I've used `getattr` to guard against it being [removed in the future.](https://github.com/pypa/setuptools/blob/6ee23bf0579c52e1cbe7c97fc20fd085ff2a25c7/NEWS.rst#L844-L847)

nulano: I'm not sure it is even possible to use LEGACY_EDITABLE mode anymore, since all my attempts end with 
```
    File "C:\Git\Pillow\_custom_build\backend.py", line 5, in <module>
      from setuptools.build_meta import *  # noqa: F403
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  AttributeError: module 'setuptools.build_meta' has no attribute 'get_requires_for_build_editable'
```
It seems pip will not even attempt to use `build_editable` without first seeing `get_requires_for_build_editable` which is not defined here, so I'm not sure this change is necessary?

nulano: Actually, this is failing on `from setuptools.build_meta import *` due to `get_requires_for_build_editable` being listed in `setuptools.build_meta.__all__`. After changing it to only import the relevant functions, I can finally build an editable wheel, but it is ignoring all `--config-settings`. So I'm not sure there is a need to support legacy mode when it is already non-functional and there have been no complaints AFAIK.

radarhere: Ok. Thanks for investigating.

radarhere (approved):

PR #7659, opened 2023-12-31
Add type hints to _binary

nulano:
There are two tricky parts in adding type hints to `PIL._binary`:

The `i16*` and `i32*` give a "Returning Any from function declared to return int" in `--strict` mode due to the use of `struct.unpack_from`. This can be properly silenced with `#type: ignore[no-any-return]`, but that gives a "Unused type ignore comment" when running without `--strict`, so I have left it out for now.

The `i8` function distinguishes between `int` and `bytes` inputs in a way that mypy doesn't recognize. Since `i8` is called with an `int` input only from `IptcImagePlugin`, I have removed those calls there and removed `int` support as input to `i8`.

Although the values in `self.info[(3, 60)]` and `self.info[(3, 65)]` can be lists if the tags are duplicated, these tags are marked as "non-repeatable" in [the specification](https://www.iptc.org/std/IIM/4.1/specification/Dnprv4.pdf) (pg. 13-14). The specification recommends using the first entry if the tags are duplicated (sec. 1.5-d), but this recommendation was already ignored in `IptcImagePlugin` and I was unable to find any images to test this with.

In doing so, I found that `IptcImagePlugin.dump`, `IptcImagePlugin.i`, and `IptcImagePlugin.PAD` are simple helpers that should probably be deprecated (edit: #7664).

### DIFF
--- a/src/PIL/IptcImagePlugin.py
+++ b/src/PIL/IptcImagePlugin.py
@@ -20,17 +20,20 @@ import os
 import tempfile
 
 from . import Image, ImageFile
-from ._binary import i8, o8
 from ._binary import i16be as i16
 from ._binary import i32be as i32
 
 COMPRESSION = {1: "raw", 5: "jpeg"}
 
-PAD = o8(0) * 4
+PAD = b"\0\0\0\0"
 
 
 #
 # Helpers
+
+
+def _i8(c: int | bytes) -> int:
+    return c if isinstance(c, int) else c[0]
 
 
 def i(c):
@@ -39,7 +42,7 @@ def i(c):
 
 def dump(c):
     for i in c:
-        print("%02x" % i8(i), end=" ")
+        print("%02x" % _i8(i), end=" ")
     print()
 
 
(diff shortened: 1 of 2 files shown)


docs/releasenotes/10.2.0.rst
```diff
@@ -20,10 +20,14 @@ ImageFile.raise_oserror
...
-
-TODO
+IptcImageFile helper functions
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The functions ``IptcImageFile.dump`` and ``IptcImageFile.i``, and the constant
+``IptcImageFile.PAD`` have been deprecated and will be removed in Pillow
+12.0.0 (2025-10-15). These are undocumented helper functions intended
+for internal use, so there is no replacement. They can each be replaced
+by a single line of code using builtin functions in Python.
```
hugovk, line 30: Is the idea to a) eventually rename them with underscores and keep using them internally, or b) remove them altogether?

If a) let's add underscore versions and have the public ones issue a warning then call those, and use the underscore versions internally.

If b) let's use the builtins directly now, and only call the deprecated versions from tests.

In both cases, no need for the `_internal: bool` parameter.

nulano: Both, a) for `i` and b) for `dump` (which is unused otherwise).

> In both cases, no need for the _internal: bool parameter.

Good point, removed.

hugovk:
This looks fine, please can you split this into two PRs, one to deprecate and the other to add type hints?

I think it will be clearer to keep them separate. 

If you want, it's okay to leave this PR as-is, and create another one that is the first step, and we can merge that one first.

nulano:
OK, I've removed the deprecation commits and will create a new PR for that.

hugovk (approved):
Thanks!

PR #7666, opened 2023-12-31
Coverage: add more exclusions

hugovk:

 * With recent Coverage.py, we can use `exclude_also` instead of `exclude_lines` which requires re-enabling the standard pragma
 * `create_lut()` was a helper method not run anywhere, should we delete it instead?
 * `class TypeGuard` is compatibility code being introduced in https://github.com/python-pillow/Pillow/pull/7642. We have a `# type: ignore[no-redef]` on it, and I don't think we need to cover it in tests either.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -2,11 +2,8 @@
 
 [report]
 # Regexes for lines to exclude from consideration
-exclude_lines =
-    # Have to re-enable the standard pragma:
-    pragma: no cover
-
-    # Don't complain if non-runnable code isn't run:
+exclude_also =
+    # Don't complain if non-runnable code isn't run
     if 0:
     if __name__ == .__main__.:
     # Don't complain about code that shouldn't run
--- a/.coveragerc
+++ b/.coveragerc
@@ -9,6 +9,8 @@ exclude_lines =
     # Don't complain if non-runnable code isn't run:
     if 0:
     if __name__ == .__main__.:
+    # Don't complain about code that shouldn't run
+    def create_lut():
     # Don't complain about debug code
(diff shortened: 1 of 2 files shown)


.coveragerc
```diff
@@ -2,15 +2,16 @@
...
+exclude_also =
+    # Don't complain if non-runnable code isn't run
     if 0:
     if __name__ == .__main__.:
+    # Don't complain about code that shouldn't run
+    def create_lut():
     # Don't complain about debug code
     if DEBUG:
+    # Don't complain about compatibility code
+    class TypeGuard:
```
nulano, line 14: What do you think about this?
```suggestion
    except ImportError:
```
Most of these blocks are missing coverage in the tests already, and I'm not sure how much value there would be in adding a CI job with no optional dependencies.

nulano, line 14: What do you think about this?
```suggestion
    # Don't complain about missing optional dependencies
    except ImportError:
```
Most of these blocks are missing coverage in the tests already, and I'm not sure how much value there would be in adding a CI job with no optional dependencies.

hugovk: This should cover both, without the colon:
```suggestion
    # Don't complain about compatibility code for missing optional dependencies
    except ImportError
```

nulano: Looking at https://app.codecov.io/gh/python-pillow/Pillow/pull/7666/indirect-changes, it seems to have worked.

nulano:
> create_lut() was a helper method not run anywhere, should we delete it instead?

+1 for deleting it, it's easy to temporarily recreate if necessary.

PR #7669, opened 2024-01-01
Do not try and crop glyphs from outside of source ImageFont image

radarhere:


nulano:
I've made a suggestion to this in #7672.

PR #7671, opened 2024-01-01
Added type hints to ImageTransform

radarhere:


hugovk (approved):
Thanks, and let's list it in the release notes (this or another PR).

src/PIL/ImageTransform.py
```diff
@@ -14,17 +14,26 @@
...
 from __future__ import annotations
 
+from typing import Sequence
+
 from . import Image
 
 
 class Transform(Image.ImageTransformHandler):
-    def __init__(self, data):
+    method: int
```
nulano, line 23:
```suggestion
    method: Image.Transform
```

radarhere: You're suggesting that we refer to the enum. My concern is that someone might just be using the integer directly.

Or are you suggesting that because type hints are not actually required in Python, we do this to encourage people to use the enum?

nulano: This seems like the same question that was discussed in https://github.com/python-pillow/Pillow/pull/7578#discussion_r1408057633

Yes, if people are concerned with type safety, they should be using the enum.

Issue #7673, opened 2024-01-01
Accessing URLs


R3dan:
# Feature Request -- Allow importing images from url

### What did you expect to happen?
I expected to be able to show, manipulate etc (standard PIL features) the image

### What actually happened?
When I tested this with the latest version I got error:
```pytb
 Traceback (most recent call last):
  File "{file address}", line 17, in <module>
    with Image.open(url) as img:
         ^^^^^^^^^^^^^^^
  File "C:\Python312\Lib\site-packages\PIL\Image.py", line 3243, in open
    fp = builtins.open(filename, "rb")
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
OSError: [Errno 22] Invalid argument: '{url}' 
```

### What are your OS, Python and Pillow versions?

* OS: Windows 12
* Python: 3.12.0 and 3.11.5
* Pillow: 10.1.0

```python
with Image.open(url) as img:
    img.load()
```
### What I want

I want to have the ability to access images from url's.

nulano:
Is there any reason you can't use the method from [the tutorial](https://pillow.readthedocs.io/en/stable/handbook/tutorial.html#reading-from-url)?

```python
from PIL import Image
from urllib.request import urlopen
url = "https://python-pillow.org/assets/images/pillow-logo.png"
img = Image.open(urlopen(url))
```

I would be wary of adding something like this to Pillow natively, since there are many other configuration options others could then request be also added (headers, user agent, cookies, https parameters, certificates, ...).

R3dan:
When looking through the docs I had not seen this. Sorry. I understand what you mean in terms of why not to add it and I would agree although also think 2 things:
1) If someone made a PR with the additional in (headers, cookies etc) then you could accept that, with no additional work for you.
2) Surely it could just be done by using the requests module and loading the bytes.

I also tried to find a way to load bytes but couldn't (I have since found out how via the tutorial). I am probably overlooking something, but whats stopping you from `importing io` with a function to load? Something like this:
```
import io
...
#class image
@classmeathod
def load_from_bytes(cls, bytes):
    return Image.open(io.BytesIO(bytes))
```

Fyi this is not me complaining. What you have made is really useful and most of this is just me not knowing how to use it (my problem). It would just seem that a lot of useful tasks (opening from url or bytes) could be easier to find and do to both new programmers (who might not know of how well documented things are) or just people new to PIL.

hugovk:
[...]

PR #7674, opened 2024-01-01
Fix image URL in "Reading from URL" example

nulano:
While testing https://github.com/python-pillow/Pillow/issues/7673#issuecomment-1873401715 I found that the image has moved from https://python-pillow.org/images/pillow-logo.png to https://python-pillow.org/assets/images/pillow-logo.png.


hugovk (approved):
Thank you!

PR #4639, opened 2020-05-23
Write JFIF header when saving JPEG

radarhere:
Resolves #3328

Pillow has some C code that sets `density_unit`, `X_density` and `Y_density`. However, [libjpeg](https://github.com/LuaDist/libjpeg/blob/master/libjpeg.txt) states that

> UINT8 density_unit
> UINT16 X_density
> UINT16 Y_density
> 	The resolution information to be written into the JFIF marker;
> 	not used otherwise.

So this code should tell libjpeg to [write the JFIF marker](https://github.com/LuaDist/libjpeg/blob/6c0fcb8ddee365e7abc4d332662b06900612e923/jcmarker.c#L528).

PR #7679, opened 2024-01-02
Allow disabling default emission of JPEG APP0 and APP14 segments

bgilbert:
When embedding JPEGs into a container file format, it may be desirable to minimize JPEG metadata, since the container will include the pertinent details.  By default, libjpeg emits a JFIF APP0 segment for JFIF-compatible colorspaces (grayscale or YCbCr) and Adobe APP14 otherwise.  Add a `no_default_app_segments` option to disable these.

#4639 added code to force emission of the JFIF segment if the DPI is specified, even for JFIF-incompatible colorspaces.  This seems inconsistent with the JFIF spec, but apparently other software does it too.  `no_default_app_segments` does not disable this behavior, since it only happens when the application explicitly specifies the DPI.


bgilbert:
The setting could also be `default_app_segments`, defaulting to `True`.

radarhere:
Recently, #7488 added "restart_marker_blocks" and "restart_marker_rows", and #7553 added "keep_rgb". Those were  functional changes, whereas this is about saving a few bytes and grouping two libjpeg settings into one, which feels a bit more arbitrary (I'm afraid that a user will come along and say that they would like to set "write_JFIF_header" to false but not "write_Adobe_marker").

To try and prevent waking up one day and finding an overabundance of settings when saving JPEGs, and being unable to rearrange them without breaking backwards compatibility, can I ask - do you have a planned list of JPEG settings that you're thinking of adding in future PRs?

bgilbert:
This is the last JPEG setting I'm currently planning.  I've been submitting them sequentially to avoid merge conflicts in `PyImaging_JpegEncoderNew()`, partly because I didn't understand how long review would take.  (Not complaining.  It's best to get API changes correct.)

This change indeed seems a bit more obscure than the other two, and I'm okay skipping it if desired.  It's possible to remove the APP segments outside of Pillow by postprocessing the byte stream.  (Assuming performance isn't a major concern, which is true for my use case.)

Alternatively, if the setting is too broad, I could split it into two tri-states such as `write_jfif_segment` and `write_adobe_segment` so the caller can force either segment on or off.

Yay295:
Would it make sense to have one setting that takes a list of segment names to enable/disable?
`exclude_app_segments = ('APP0','APP14')`

bgilbert:
I believe JFIF and Adobe are the only APP segments supported natively by libjpeg, so those would be the only two supported values unless Pillow adds its own segments.

src/libImaging/Jpeg.h
```diff
@@ -77,6 +77,9 @@ typedef struct {
     /* Disable automatic conversion of RGB images to YCbCr if non-zero */
     int keep_rgb;
 
+    /* Disable default application segments if nonzero */
```
radarhere, line 80:
```suggestion
    /* Disable default application segments if non-zero */
```
See #7695

bgilbert:
Whoops, didn't notice the branch update before force-pushing.  I've now rebased onto `main`.  The net diff of the force-pushes is [here](https://github.com/python-pillow/Pillow/compare/cd3bc9189f1c81680ae31a63a99ca1d85957d8bc..b29b67ecd6d3b2a69a176b52ef6c07be44e91e43).

Tests/test_file_jpeg.py
```diff
@@ -88,6 +88,32 @@ def test_app(self):
...
+        expect_app14,
+    ):
+        out = BytesIO()
+        hopper().save(
+            out,
+            format="JPEG",
+            keep_rgb=keep_rgb,
+            no_default_app_segments=no_default_app_segments,
+        )
+        assert (b"\xff\xe0" in out.getvalue()) == expect_app0
```
homm, line 114: This is not reliable check. Can we open image instead and check `im.applist` instead?

bgilbert:
Done.

Before we get too far into detailed review, could I get some maintainer feedback about the high-level approach?  Should I continue with `no_default_app_segments`, invert to `default_app_segments`, split into two options, drop the PR, something else?

bgilbert:
Something like:

```python
start = fp.tell()
img.save(fp, format='JPEG', quality='web_medium', streamtype=2, no_default_app_segments=True)
end = fp.tell()
```

Container metadata, including JPEG tables, can then be written afterward as appropriate.  My immediate use case is for a TIFF builder (I need to be able to write non-standards-conforming TIFFs), and DICOM datasets are also on my radar.  Any file format designed to embed multiple JPEGs can potentially use this.

Tests/test_file_jpeg.py
```diff
@@ -88,6 +88,31 @@ def test_app(self):
             assert im.info["comment"] == b"File written by Adobe Photoshop\xa8 4.0\x00"
             assert im.app["COM"] == im.info["comment"]
 
+    @pytest.mark.parametrize(
+        "keep_rgb, no_default_app_segments, expect_app0, expect_app14",
```
homm, line 92: Could you also add a case with dpi? (JFIF should present as documentation said)

Tests/test_file_jpeg.py
```diff
@@ -101,6 +101,31 @@ def test_app(self) -> None:
...
+            (True, True, False, False),
+        ),
+    )
+    def test_default_app_write(
+        self,
+        keep_rgb,
+        no_default_app_segments,
+        expect_app0,
+        expect_app14,
+    ):
```
radarhere, line 119:
```suggestion
        keep_rgb: bool,
        no_default_app_segments: bool,
        expect_app0: bool,
        expect_app14: bool,
    ) -> None:
```

radarhere:
> I need to be able to write non-standards-conforming TIFFs

This doesn't sound like a very common need for this feature.

Issue #7571, opened 2023-11-25
Release Pillow 10.2.0 on January 2, 2024


hugovk:
## Main Release

Released quarterly on January 2nd, April 1st, July 1st and October 15th.

* [x] Open a release ticket e.g. https://github.com/python-pillow/Pillow/issues/3154
* [x] Develop and prepare release in `main` branch.
* [x] Check [GitHub Actions](https://github.com/python-pillow/Pillow/actions) and [AppVeyor](https://ci.appveyor.com/project/python-pillow/Pillow) to confirm passing tests in `main` branch.
* [x] Check that all of the wheel builds pass the tests in the [GitHub Actions "Wheels" workflow](https://github.com/python-pillow/Pillow/actions/workflows/wheels.yml) and [Travis CI](https://app.travis-ci.com/github/python-pillow/pillow) jobs by manually triggering them.
* [x] In compliance with [PEP 440](https://peps.python.org/pep-0440/), update version identifier in `src/PIL/_version.py`
* [x] Update `CHANGES.rst`.
* [x] Run pre-release check via `make release-test` in a freshly cloned repo.
* [x] Create branch and tag for release e.g.:
  ```bash
  git branch 5.2.x
  git tag 5.2.0
  git push --tags
  ```
* [x] Create and upload all [source and binary distributions](https://github.com/python-pillow/Pillow/blob/main/RELEASING.md#source-and-binary-distributions)
* [x] Publish the [release on GitHub](https://github.com/python-pillow/Pillow/releases)
* [x] In compliance with [PEP 440](https://peps.python.org/pep-0440/),
      increment and append `.dev0` to version identifier in `src/PIL/_version.py` and then:
  ```bash
  git push --all
   ```

## Source and Binary Distributions

* [x] Check the [GitHub Actions "Wheels" workflow](https://github.com/python-pillow/Pillow/actions/workflows/wheels.yml) has passed, including the "Upload release to PyPI" job. This will have been triggered by the new tag.
* [x] Download the Linux aarch64 wheels created by Travis CI from [GitHub releases](https://github.com/python-pillow/Pillow/releases) and copy into `dist`. Check and upload them e.g.:
  ```bash
  python3 -m twine check --strict dist/*
  python3 -m twine upload dist/Pillow-5.2.0*
  ```
## Publicize Release

[...]

PR #7501, opened 2023-10-29
Update Travis CI to Python 3.12 and Ubuntu 22.04 Jammy Jellyfish

hugovk:
For https://github.com/python-pillow/Pillow/issues/6941.

Travis CI now has Python 3.12.

We can also bump the Ubuntu version.

Issue #7485, opened 2023-10-23
Set up GitHub release token for Travis CI


hugovk:
For https://github.com/python-pillow/Pillow/issues/7390.
Re: https://github.com/python-pillow/Pillow/issues/7348#issuecomment-1763383329, https://github.com/python-pillow/pillow-wheels/pull/156#discussion_r450959499.

We need to set up a token for Travis CI to be able to upload wheels to this repo's release section.

1. Create a GitHub personal access token scoped to this repo. ⚠️ A repo admin needs to do this due to permissions.
2. Install Travis CLI:

```console
❯ sudo gem install travis --no-document
...
❯ travis --version
Shell completion not installed. Would you like to install it now? |y| y
1.11.1
```

3. The next bit is where I got stuck, but try something like this (https://stackoverflow.com/a/76234874/724176):

```
❯ travis login --github-token GITHUB_PERSONAL_ACCESS_TOKEN

❯ travis setup releases -r python-pillow/Pillow 
```

5. Save the new token as `GITHUB_RELEASE_TOKEN` at https://app.travis-ci.com/github/python-pillow/pillow/settings.

hugovk:
@aclark4life Please could you look into this? It needs to be done by someone with admin permissions. Thanks!

aclark4life:
@hugovk Sure, level of urgency?

hugovk:
It would be nice (but not essential) to have it ready for the next quarterly release.

hugovk:
Actually, we do need this before the next release, otherwise we'll need to do a similar hack to [last time](https://github.com/python-pillow/Pillow/issues/7348#issuecomment-1763371082) by uploading the files to a third-party site instead, and from a security perspective it would be better to avoid that.

aclark4life:
@hugovk OK I'll work on it this week

aclark4life:
@hugovk When you say stuck do you mean this or something else? It's possible I need to make a better token for this task with the appropriate permissions … I have not done that yet.

![Screenshot 2023-11-27 at 9 32 16 AM](https://github.com/python-pillow/Pillow/assets/72164/6a681a71-aec6-4720-890e-0e6127c346d6)

hugovk:
Yes, I also got `resource not found ({}`: https://github.com/python-pillow/Pillow/issues/7348#issuecomment-1763351055

aclark4life:
I cannot unsee what I've seen here today in working on this … but good news: progress!!

```
[...]

PR #7680, opened 2024-01-02
Downgrade TravisCI to Focal and Python 3.11

radarhere:
- During [an attempt to get TravisCI working](https://github.com/python-pillow/Pillow/issues/7485#issuecomment-1840514584) in #7485, I found that downgrading to Focal (reverting #7501) resolved the following error
> /home/travis/.rvm/gems/ruby-3.1.2/gems/faraday-0.15.4/lib/faraday/options.rb:166:in `new': tried to create Proc object without a block (ArgumentError)
- When following the releasing process in #7571, I found that the wheel filenames now include 'pillow' in lowercase. This is likely a result of cibuildwheel in #7552

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,8 @@ env:
 
 language: python
 # Default Python version is usually 3.6
-python: "3.12"
-dist: jammy
+python: "3.11"
+dist: focal
 services: docker
 
 jobs:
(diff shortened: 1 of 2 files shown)


radarhere:
Closing, as [a different plan is in the works for TravisCI](https://github.com/python-pillow/Pillow/issues/7485#issuecomment-1873821482).

I've opened https://github.com/python-pillow/Pillow/pull/7681 for the RELEASING change.

PR #7691, opened 2024-01-04
Updated Linux and macOS wheels matrix variable name

radarhere:
Minor suggestion.

Renames `archs` in Linux and macOS
https://github.com/python-pillow/Pillow/blob/27eacddfd51fab9b4432b3e228836b5947685763/.github/workflows/wheels.yml#L33-L46
to `cibw_arch`, to better match the variable name used in Windows
https://github.com/python-pillow/Pillow/blob/27eacddfd51fab9b4432b3e228836b5947685763/.github/workflows/wheels.yml#L83-L93

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -39,18 +39,18 @@ jobs:
         include:
           - name: "macOS x86_64"
             os: macos-latest
-            archs: x86_64
+            cibw_arch: x86_64
             macosx_deployment_target: "10.10"
           - name: "macOS arm64"
             os: macos-latest
-            archs: arm64
+            cibw_arch: arm64
             macosx_deployment_target: "11.0"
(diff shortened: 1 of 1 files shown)


hugovk:
@nulano What do you think about removing `matrix.arch` from the Windows job so it only has `matrix.cibw_arch`, and refactoring `build_prepare.py` to use `matrix.cibw_arch` values?

More effort than it's worth, or a nice cleanup?

nulano:
I have considered it. ~~It would make sense, but I didn't feel like doing it 😄.
The current arch definition matches the use by webp, so a new key would need to be added to `ARCHITECTURES`.~~
...is what I was going to write, but looking at it now, webp in fact uses a different value for arm.
It might also make sense to build webp with CMake (which doesn't need the arch at all) instead of NMake; I'll take a look.

PR #7693, opened 2024-01-04
Rename x64 to AMD64 in winbuild

nulano:
Suggested in https://github.com/python-pillow/Pillow/pull/7691#issuecomment-1876480543

 * Rename x64 to AMD64 in winbuild.
 * Use CMake build of libwebp instead of NMake build: It seems to be more actively maintained and avoids the need for specifying the arch (which is used incorrectly in the nmake build anyway).


hugovk (approved):
Thanks!

PR #7696, opened 2024-01-06
Add support for reading and writing grayscale PFM images

nulano:
Based on documentation:
* https://linux.die.net/man/5/pfm
* https://www.pauldebevec.com/Research/HDR/PFM/

`hopper.pfm` created by converting `hopper("F").save("im.tiff")` using https://convertio.co/tiff-pfm/.
`hopper_be.pfm` created using `struct.pack`.

Color PFM images are not supported due to #1888.


radarhere:
Interestingly, there is disagreement in the documentation about row order - one says top to bottom, the other says bottom to top.

https://linux.die.net/man/5/pfm
> The raster is a sequence of pixels, packed one after another, with no delimiters of any kind. They are in standard Western reading order: left to right and top to bottom within the image.

https://www.pauldebevec.com/Research/HDR/PFM/
> After the final carriage return the file proceeds with a series of three 4-byte IEEE 754 single precision floating point numbers for each pixel, specified in left to right, bottom to top order.

nulano:
Huh, I thought https://linux.die.net/man/5/pfm was identical to https://netpbm.sourceforge.net/doc/pfm.html, which says
> The raster is a sequence of pixels, packed one after another, with no delimiters of any kind. They are grouped by row, with the pixels in each row ordered left to right and the rows ordered bottom to top.

From https://sourceforge.net/p/netpbm/code/1682/tree//userguide/pfm.html?diff=51243f8e2718460a4a53b63a:1681, it looks like this was changed in 2012, with "bottom to top" being the newer version. I suspect the doc file was wrong until then, since looking at the 10.22 release of netpbm from 2004, the HISTORY file shows that this is the version that added PFM support and the source code for the PFM utilities shows bottom to top order being used.

Similarly, [Wikipedia](https://en.wikipedia.org/wiki/Netpbm) says:
> After the header the file proceeds with floating point numbers for each pixel, specified in left-to-right, bottom-to-top order. Some programs suggest PF4 as an additional extension for the RGBA format.

I've also tested Gimp, ImageMagick's `convert`, and libjxl's `cjxl`, and they all agree on bottom-to-top order.

radarhere (approved):

Tests/test_file_ppm.py
```diff
@@ -100,6 +105,44 @@ def test_pnm(tmp_path):
...
+
+def test_pfm_big_endian(tmp_path):
+    with Image.open("Tests/images/hopper_be.pfm") as im:
+        assert im.info["scale"] == 2.5
+        assert_image_equal(im, hopper("F"))
+
+        f = str(tmp_path / "tmp.pfm")
+        im.save(f)
+
+        assert_image_equal_tofile(im, f)
```
hugovk, line 127: Let's rename `f` to `filename`:
```suggestion
        filename = str(tmp_path / "tmp.pfm")
        im.save(filename)

        assert_image_equal_tofile(im, filename)


def test_pfm_big_endian(tmp_path):
    with Image.open("Tests/images/hopper_be.pfm") as im:
        assert im.info["scale"] == 2.5
        assert_image_equal(im, hopper("F"))

        filename = str(tmp_path / "tmp.pfm")
        im.save(filename)

        assert_image_equal_tofile(im, filename)
```

docs/handbook/image-file-formats.rst
```diff
@@ -696,6 +696,24 @@ PCX
 
 Pillow reads and writes PCX files containing ``1``, ``L``, ``P``, or ``RGB`` data.
 
+PFM
+^^^
+
+.. versionadded:: 10.3.0
+
+Pillow reads and writes grayscale (Pf format) PFM files containing ``F`` data.
```
hugovk, line 704:
```suggestion
Pillow reads and writes grayscale (Pf format) Portable FloatMap (PFM) files containing ``F`` data.
```

src/PIL/PpmImagePlugin.py
```diff
@@ -110,6 +113,14 @@ def _open(self):
         if magic_number in (b"P1", b"P2", b"P3"):
             decoder_name = "ppm_plain"
         for ix in range(3):
+            if mode == "F" and ix == 2:
+                scale = float(self._read_token())
+                if not math.isfinite(scale) or scale == 0.0:
```
hugovk, line 118: I find it a bit more readable to have `not` last:
```suggestion
                if scale == 0.0 or not math.isfinite(scale):
```

hugovk (commented):
Please could you also add this to the release notes?

nulano:
> Please could you also add this to the release notes?

Done

Issue #7683, opened 2024-01-02
Image.transform data not well documented


AlexKubiesa:
The [Image.transform](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.transform) documentation describes the `data` parameter as

> Extra data to the transformation method.

However, the documentation doesn't mention which data needs to be provided and in what format.

I am using perspective transform in a project and I was able to reverse engineer that a 1D numpy array of 8 floats works, but as I am lacking some of the theory behind it, it was unintuitive and I'm still not sure what the best way to generate those 8 numbers is.

nulano:
With the exception of the perspective transform, I think the others are documented here: https://pillow.readthedocs.io/en/stable/PIL.html#module-PIL.ImageTransform
However, I agree that this is not well discoverable in the documentation.

AlexKubiesa:
@nulano thank you for updating it!

PR #7699, opened 2024-01-07
Added PerspectiveTransform

radarhere:
Furthers the work of #7684 to help clarify ImageTransform for #7683, by adding code and documentation for the missing ImageTransform - PerspectiveTransform.

This is based on [AffineTransform](https://github.com/python-pillow/Pillow/blob/a786a0551b75ab3e85da1bfad54226faa2336022/src/PIL/ImageTransform.py#L45-L61), since it is similar.
https://github.com/python-pillow/Pillow/blob/a786a0551b75ab3e85da1bfad54226faa2336022/src/libImaging/Geometry.c#L375-L417

If the 7th (g) and 8th (h) values are 0, then `perspective_transform` is identical to `affine_transform`, so`PERSPECTIVE` should be able to do everything that `AFFINE` can.


Tests/test_image_transform.py
```diff
@@ -16,6 +16,8 @@ def test_sanity(self):
 
         transform = ImageTransform.AffineTransform(seq[:6])
         im.transform((100, 100), transform)
+        transform = ImageTransform.PerspectiveTransform(seq[:8])
+        im.transform((100, 100), transform)
```
hugovk, line 20: Can we add some correctness tests rather than just a "didn't crash" test?

radarhere: Ok, done

homm (approved):

hugovk (approved):
Should we add to release notes?

radarhere:
Ok, sure, I've added release notes.

PR #7701, opened 2024-01-07
Fix APNG info after seeking backwards more than twice

### LINKED ISSUE #7700: Seeking APNG to frame 0 multiple times breaks the info attribute
### What did you do?

- Load an APNG with 3 frames with different durations (let us imagine the first frame has a duration of 1ms, the second frame has a duration of 2ms and the third frame has a duration of 3m)
- iterate through the image - looking at the duration supplied by the info dict, output will appropriately be (1ms, 2ms, 3ms)
- do the same thing, output will appropriately be (1ms, 2ms, 3ms)
- do the same thing, output will incorrectly be (3ms, 2ms, 3ms)

The duration value per frame should be accurate no matter how many times the image is seeked

### What actually happened?

After the 2nd loop through the first frame will always have the info attributes of the last frame iterated.

* OS: Linux
* Python: 3.12
* Pillow: 10.2.0

---

The problem is, afaik, specific to APNG. When seeking the `Image` to the first frame the `PngStream` goes through a "rewind", in which it assigns the `rewind_state` (generated on the first read of the first frame) to the current state of the object. However, the rewind state `im_info` is simply assigned to the current state, not copied. This means the rewind state info object is later modified inadvertently because current and rewind state share the same object.

https://github.com/python-pillow/Pillow/blob/a786a0551b75ab3e85da1bfad54226faa2336022/src/PIL/PngImagePlugin.py#L381

This should be:
```python
self.im_info = self.rewind_state["info"].copy()
```

esoma:
Fixes #7700


hugovk:
Please could you add a test for this?

esoma:
@hugovk 
Added the test and checked it against main to ensure it produces the issue. If ya'll have any preferences on the naming of the test or if you want a full explanation of the issue in the test just let me know.

I'm guessing the codecov failure is because the github actions aren't approved to run and the appveyor builds don't emit coverage in the src directory.

radarhere:
I approved the GitHub Actions run and CodeCov is no longer failing.

Tests/test_file_apng.py
```diff
@@ -689,3 +689,13 @@ def test_different_modes_in_later_frames(mode, default_image, duplicate, tmp_pat
     )
     with Image.open(test_file) as reloaded:
         assert reloaded.mode == mode
+
+
+def test_apng_issue_7700():
+    # https://github.com/python-pillow/Pillow/issues/7700
+    with Image.open("Tests/images/apng/issue_7700.png") as im:
```
hugovk, line 696: Please could you give the test and image more descriptive names than `issue_7700`?

Tests/test_file_apng.py
```diff
@@ -689,3 +689,13 @@ def test_different_modes_in_later_frames(mode, default_image, duplicate, tmp_pat
...
+        for i in range(5):
+            im.seek(0)
+            assert im.info["duration"] == 4000.0
+            im.seek(1)
+            assert im.info["duration"] == 1000.0
```
hugovk, line 701:
```suggestion
            assert im.info["duration"] == 4000
            im.seek(1)
            assert im.info["duration"] == 1000
```

Tests/test_file_apng.py
```diff
@@ -689,3 +689,13 @@ def test_different_modes_in_later_frames(mode, default_image, duplicate, tmp_pat
     )
     with Image.open(test_file) as reloaded:
         assert reloaded.mode == mode
+
+
+def test_apng_issue_7700():
```
hugovk, line 694:
```suggestion
def test_apng_issue_7700() -> None:
```

esoma:
Is an occasional segfault in testing a known issue? I'm guessing the reason for the failing action is not related to this PR.

nulano:
Someone has triggered it to re-run so I can't see which job failed, but yes, the PyPy jobs occasionally segfault while collecting tests.

radarhere:
I triggered the rerun. It was https://github.com/python-pillow/Pillow/actions/runs/7452748291/job/20281343444 that failed.

Tests/test_file_apng.py
```diff
@@ -689,3 +689,13 @@ def test_different_modes_in_later_frames(mode, default_image, duplicate, tmp_pat
     )
     with Image.open(test_file) as reloaded:
         assert reloaded.mode == mode
+
+
+def test_apng_repeated_seeks_give_correct_info() -> None:
+    # https://github.com/python-pillow/Pillow/issues/7700
+    with Image.open("Tests/images/apng/repeated_seeks_give_correct_info.png") as im:
+        for i in range(5):
```
radarhere, line 697:
```suggestion
    with Image.open("Tests/images/apng/repeated_seeks_give_correct_info.png") as im:
        for i in range(3):
```
You only need 3 to demonstrate the problem
1. Seek through the image from the initial state.
2. Rewind, loading in `self.rewind_state["info"]`. Seek to the second frame, setting the duration in `self.rewind_state["info"]`
3. Rewind again, but `self.rewind_state["info"]` has been modified now to have the duration of the second frame.

Tests/test_file_apng.py
```diff
@@ -689,3 +689,12 @@ def test_different_modes_in_later_frames(mode, default_image, duplicate, tmp_pat
     )
     with Image.open(test_file) as reloaded:
         assert reloaded.mode == mode
+
+
+def test_apng_repeated_seeks_give_correct_info() -> None:
+    with Image.open("Tests/images/apng/repeated_seeks_give_correct_info.png") as im:
```
hugovk, line 695: Thanks, can we concisely describe the image rather than how it's used? It may come in useful for other tests later. Maybe something like this?
```suggestion
    with Image.open("Tests/images/apng/different_durations.png") as im:
```

hugovk:
Thank you!

PR #7707, opened 2024-01-10
Added type hints to ImageMath

radarhere:


src/PIL/Image.py
```diff
@@ -1305,7 +1305,7 @@ def getbands(self):
         """
         return ImageMode.getmode(self.mode).bands
 
-    def getbbox(self, *, alpha_only=True):
+    def getbbox(self, *, alpha_only=True) -> tuple[int, int, int, int]:
```
hugovk, line 1308:
```suggestion
    def getbbox(self, *, alpha_only: bool = True) -> tuple[int, int, int, int]:
```

src/PIL/ImageMath.py
```diff
@@ -247,7 +257,7 @@ def eval(expression, _dict={}, **kw):
 
     compiled_code = compile(expression, "<string>", "eval")
 
-    def scan(code):
+    def scan(code) -> None:
```
hugovk, line 260: What type is `code`?

radarhere: I've pushed a commit.

hugovk: `code`'s type is `CodeType` :D

PR #7709, opened 2024-01-12
Handle truncated chunks at the end of PNG images

lajiyuan:
Fix png image plugin load_end func handle truncated file.

When attempting to process PNG image files that are incomplete at the end, the current LOAD_TRUNCATED_IMAGES parameter is unable to handle such cases. It is necessary to add corresponding handling code within the load_end function to ensure proper processing of incomplete images. Below is a link to an example of such an anomalous image:

error_image

 * Fix PngImagePlugin.py load_end func, add exception handle.

cc @radarhere Can you help me with a code review? thanks a lot.


radarhere:
If you're interested in an immediate solution as an alternative to this PR, I find that you can just ignore the error and the image works fine.
```python
from PIL import Image
im = Image.open('in.png')
try:
    im.load()
except OSError:
    pass
im.save("out.png")
```

lajiyuan:
I'm amazed by your quick response. Indeed, loading images by ignoring load exceptions can be a way to load and convert images into a normal state. Thank you for your suggestion. Are there any other suggested modifications for this pull request?

@radarhere

Tests/test_file_png.py
```diff
@@ -786,18 +786,20 @@ class TestTruncatedPngPLeaks(PillowLeakTestCase):
     iterations = 100  # Leak is 56k/iteration, this will leak 5.6megs
 
     def test_leak_load(self):
```
radarhere, line 788: Is there any particular reason you have chosen to add your test case to `TestTruncatedPngPLeaks`, rather than `TestFilePng`? `TestTruncatedPngPLeaks` was originally added in #2541, which was partly about memory leaks, but your description so far doesn't mention memory at all.

src/PIL/PngImagePlugin.py
```diff
@@ -981,7 +981,16 @@ def load_end(self):
             except EOFError:
                 if cid == b"fdAT":
                     length -= 4
-                ImageFile._safe_read(self.fp, length)
+                try:
+                    ImageFile._safe_read(self.fp, length)
+                except OSError as e:
+                    if (
+                        str(e) == "Truncated File Read"
+                    ) and ImageFile.LOAD_TRUNCATED_IMAGES:
```
radarhere, line 989:
```suggestion
                    if ImageFile.LOAD_TRUNCATED_IMAGES:
```
Given that [`ImageFile._safe_read()`](https://github.com/python-pillow/Pillow/blob/067c5f4123c7cab7507fca02605bfd9762861d1f/src/PIL/ImageFile.py#L572-L604) only explicitly raises an `OSError` with this message, I don't think the message needs to be checked.

radarhere:
The test image that you've added is 721kb - this is a bit large.

Here's a test image I've created by modifying [hopper_idat_after_image_end.png](https://github.com/python-pillow/Pillow/blob/main/Tests/images/hopper_idat_after_image_end.png) - it is only 30kb.

load_end_truncated_idat

Tests/test_file_png.py
```diff
@@ -778,6 +778,15 @@ class MyStdOut:
         with Image.open(mystdout) as reloaded:
             assert_image_equal_tofile(reloaded, TEST_PNG_FILE)
 
+    def test_end_truncated_filed(self):
+        ImageFile.LOAD_TRUNCATED_IMAGES = True
+        try:
+            img = Image.open("Tests/images/end_trunc_file.png")
+            img.load()
+            assert img is not None
```
radarhere, line 786:
```suggestion
            with Image.open("Tests/images/end_trunc_file.png") as im:
                assert_image_equal_tofile(im, "Tests/images/hopper.png")
```
Two suggestions here.
1) This [uses a context manager.](https://pillow.readthedocs.io/en/stable/releasenotes/7.0.0.html#image-del)
2) This tests the image contents are correct.

lajiyuan: ok, i will resolve it.

lajiyuan:
@radarhere  The comments have all been fixed.

lajiyuan:
Do I need to perform any additional actions for this PR to be merged? @radarhere

src/PIL/PngImagePlugin.py
```diff
@@ -981,7 +981,14 @@ def load_end(self):
             except EOFError:
                 if cid == b"fdAT":
                     length -= 4
-                ImageFile._safe_read(self.fp, length)
+                try:
+                    ImageFile._safe_read(self.fp, length)
+                except OSError as e:
+                    if ImageFile.LOAD_TRUNCATED_IMAGES:
+                        logger.debug("%r %s %s (Truncated File Read)", cid, pos, length)
```
radarhere, line 988:
```suggestion
```
I don't think you need the logging, because none of code that deals with truncated code in ImageFile or PngImagePlugin has logging - but feel free to disagree with me.

Tests/test_file_png.py
```diff
@@ -778,6 +778,14 @@ class MyStdOut:
         with Image.open(mystdout) as reloaded:
             assert_image_equal_tofile(reloaded, TEST_PNG_FILE)
 
+    def test_end_truncated_filed(self):
```
radarhere, line 781: I'm guessing you meant "truncated_file"?

radarhere:
I've just combined the commits in this PR into one so that your original image isn't part of the commit history.

Tests/test_file_png.py
```diff
@@ -786,6 +786,7 @@ def test_truncated_end_chunk(self):
         ImageFile.LOAD_TRUNCATED_IMAGES = True
         try:
             with Image.open("Tests/images/truncated_end_chunk.png") as im:
+                im.load()
```
radarhere, line 789: Why do you think it is necessary to add this?

lajiyuan: The test cases just now indicated that the new lines are not covered. 
My assumption is that **assert_image_equal_tofile** may not actually invoke load_end. However, I haven't looked into the code; I'm just testing whether the additions are covered.

lajiyuan: I just checked the implementation of assert_image_equal_tofile, and my guess seems to be incorrect. I need to rethink why the test cases are not covering it.

radarhere: The lack of coverage wasn't because `im.load()` wasn't there.

Because you are a first-time contributor to Pillow, our CI jobs mostly require manual approval by someone from the Pillow team before they are run. If the CI jobs are not run, this affects coverage. Approval had not been given yet.

lajiyuan: oh，I see. thanks for reponse.

lajiyuan: @radarhere  I have reset my wrong commit, i'am sorry for that.

radarhere (approved):

Tests/test_file_png.py
```diff
@@ -778,6 +778,18 @@ class MyStdOut:
         with Image.open(mystdout) as reloaded:
             assert_image_equal_tofile(reloaded, TEST_PNG_FILE)
 
+    def test_truncated_end_chunk(self):
```
radarhere, line 781:
```suggestion
    def test_truncated_end_chunk(self) -> None:
```
To follow on with the type hinting added in #7756

PR #7712, opened 2024-01-13
Added type hints for ImageShow

radarhere:


src/PIL/ImageShow.py
```diff
@@ -180,7 +182,11 @@ class UnixViewer(Viewer):
     format = "PNG"
     options = {"compress_level": 1, "save_all": True}
 
-    def get_command(self, file, **options):
+    @abc.abstractmethod
```
hugovk, line 185: Shall we exclude this from coverage?

PR #7718, opened 2024-01-15
ImageMath: Only compute `ops` dict when needed

### LINKED ISSUE #7717: Image.open: dictionary changed size during iteration (race condition)
### What did you do?

We are running a Django server with Gunicorn, with a view that is returning a thumbnail from an existing PNG image.
We open it from a BytesIO with `PIL.Image.open`.

The `Image` instance is returned by `open`.

### What actually happened?

Python interpreter raised `RuntimeError(dictionary changed size during iteration)` on [this line](https://github.com/python-pillow/Pillow/blob/6956d0b2853f5c7ec5f6ec4c60725c5a7ee73aeb/src/PIL/ImageMath.py#L217) :

```python
  File "PIL/Image.py", line 3258, in open
    preinit()
  File "PIL/Image.py", line 320, in preinit
    from . import GifImagePlugin
  File "PIL/GifImagePlugin.py", line 34, in <module>
    from . import (
  File "PIL/ImageMath.py", line 217, in <module>
    for k, v in list(globals().items()):
```

* OS: Ubuntu 22.04 AArch64
* Python: 3.9.13 (CPython)
* Pillow: 10.2.0

Unfortunately, it looks like it is a race condition on `globals()` with another lazy-loaded dependency, I don't have a minimal code that can reproduce this bug consistently. It is highly unlikely to happen, but is has happened once on our servers.

```python
# in a Django view
pil_image = Image.open(raw_image_bytes)
```

hugovk:
An idea to fix #7717, only compute the `ops` dict when needed, when first calling `eval`, and `functools.lru_cache` it so it's only computed once.

Also rewrite as a dict comprehension. Timings for non-cached use:

200000 loops, best of 5: 1.87 usec per loop
->
200000 loops, best of 5: 1.53 usec per loop

### DIFF
--- a/src/PIL/ImageMath.py
+++ b/src/PIL/ImageMath.py
@@ -17,7 +17,7 @@
 from __future__ import annotations
 
 import builtins
-from functools import cache
+from functools import lru_cache
 from types import CodeType
 from typing import Any
 
@@ -225,7 +225,7 @@ def imagemath_convert(self: _Operand, mode: str) -> _Operand:
     return _Operand(self.im.convert(mode))
 
 
-@cache
+@lru_cache
 def _get_ops() -> dict[str, Any]:
     return {k[10:]: v for k, v in globals().items() if k[:10] == "imagemath_"}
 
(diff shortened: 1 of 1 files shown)


radarhere:
I wonder if #7721 is a simpler and faster solution?

hugovk:
Yes, that's better, thanks!

PR #7721, opened 2024-01-16
Changed ImageMath.ops to be static

### LINKED ISSUE #7717: Image.open: dictionary changed size during iteration (race condition)

radarhere:
Resolves #7717

### DIFF
--- a/src/PIL/ImageMath.py
+++ b/src/PIL/ImageMath.py
@@ -224,10 +224,15 @@ def imagemath_convert(self: _Operand, mode: str) -> _Operand:
     return _Operand(self.im.convert(mode))
 
 
-ops = {}
-for k, v in list(globals().items()):
-    if k[:10] == "imagemath_":
-        ops[k[10:]] = v
+ops = {
+    "int": imagemath_int,
+    "float": imagemath_float,
+    "equal": imagemath_equal,
+    "notequal": imagemath_notequal,
+    "min": imagemath_min,
+    "max": imagemath_max,
+    "convert": imagemath_convert,
+}
 
 
 def eval(expression: str, _dict: dict[str, Any] = {}, **kw: Any) -> Any:
@@ -244,7 +249,7 @@ def eval(expression: str, _dict: dict[str, Any] = {}, **kw: Any) -> Any:
     """
 
     # build execution namespace
-    args = ops.copy()
+    args: dict[str, Any] = ops.copy()
     for k in list(_dict.keys()) + list(kw.keys()):
         if "__" in k or hasattr(builtins, k):
             msg = f"'{k}' not allowed"


hugovk:
Looks good, thanks.

I don't expect we'll add new `imagemath_*` functions often, but if we do, is it worth adding a test to make sure we remember to add it to the dict?

hugovk:
Ah, that's right, temporarily commenting one one:

```diff
diff --git a/src/PIL/ImageMath.py b/src/PIL/ImageMath.py
index a7652f237..47e552d3a 100644
--- a/src/PIL/ImageMath.py
+++ b/src/PIL/ImageMath.py
@@ -231,7 +231,7 @@ ops = {
     "notequal": imagemath_notequal,
     "min": imagemath_min,
     "max": imagemath_max,
-    "convert": imagemath_convert,
+    # "convert": imagemath_convert,
 }
```

Fails like this:

```pytb
Tests/test_imagemath.py ........F..................                                                                              [100%]

=============================================================== FAILURES ===============================================================
_____________________________________________________________ test_convert _____________________________________________________________

    def test_convert():
>       assert pixel(ImageMath.eval("convert(A+B, 'L')", images)) == "L 3"

Tests/test_imagemath.py:84:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
src/PIL/ImageMath.py:276: in eval
    scan(compiled_code)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

code = <code object <module> at 0x107002f70, file "<string>", line 1>

    def scan(code: CodeType) -> None:
        for const in code.co_consts:
            if type(const) is type(compiled_code):
                scan(const)

        for name in code.co_names:
            if name not in args and name != "abs":
                msg = f"'{name}' not allowed"
>               raise ValueError(msg)
E               ValueError: 'convert' not allowed

src/PIL/ImageMath.py:274: ValueError
======================================================= short test summary info ========================================================
FAILED Tests/test_imagemath.py::test_convert - ValueError: 'convert' not allowed
===================================================== 1 failed, 26 passed in 1.80s =====================================================
```

hugovk (approved):

PR #7733, opened 2024-01-20
Added type hints to Tests/helper.py

radarhere:


Tests/helper.py
```diff
@@ -86,28 +87,32 @@ def assert_image(im, mode, size, msg=None):
         )
 
 
-def assert_image_equal(a, b, msg=None):
+def assert_image_equal(a: Image.Image, b: Image.Image, msg: str | None = None) -> None:
     assert a.mode == b.mode, msg or f"got mode {repr(a.mode)}, expected {repr(b.mode)}"
     assert a.size == b.size, msg or f"got size {repr(a.size)}, expected {repr(b.size)}"
     if a.tobytes() != b.tobytes():
-        if HAS_UPLOADER:
+        if uploader:
```
nulano, line 94: This `if` is probably no longer needed since it is now checked in the `upload` function.

Tests/helper.py
```diff
@@ -125,55 +130,68 @@ def assert_image_similar(a, b, epsilon, msg=None):
             + f" average pixel value difference {ave_diff:.4f} > epsilon {epsilon:.4f}"
         )
     except Exception as e:
-        if HAS_UPLOADER:
+        if uploader:
```
nulano, line 133: Same for this `if`.

Tests/helper.py
radarhere: If this condition was removed, then the `logger.error` on line 97 would always run, and similarly, the `logger.exception` on line 136 would always run.

nulano: Perhaps it shouldn't run if there is no url returned?
If `uploader = "show"`, that logger message is probably also undesirable.
```python
url = upload(a, b)
if url:
    logger.error("URL for test images: %s", url)
```

hugovk (commented):
I also added some hints to this [file yesterday](https://github.com/hugovk/Pillow/commit/90babc825ae133d9d37938bb3597f13433f79656) but hadn't opened the PR yet :) The good news is we can combine these!

Tests/helper.py
```diff
@@ -228,17 +246,17 @@ def _test_leak(self, core):
...
 
-def tostring(im, string_format, **options):
+def tostring(im: Image.Image, string_format: str, **options: dict[str, Any]) -> bytes:
     out = BytesIO()
     im.save(out, string_format, **options)
     return out.getvalue()
 
 
-def hopper(mode=None, cache={}):
+def hopper(mode: str | None = None, cache: dict[str, Image.Image] = {}) -> Image.Image:
```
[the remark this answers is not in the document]
radarhere: Testing, this suggestion changes the behaviour. I guess it's actually using the default to store the cache values.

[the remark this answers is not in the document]
hugovk: Righto, let's keep the original.

Tests/helper.py
radarhere: Ok, I've pushed https://github.com/python-pillow/Pillow/pull/7733/commits/f7701e6596c9a2e7c9ed886621f01b3861d70990 for this.

Tests/helper.py
radarhere: Ok, I've pushed https://github.com/python-pillow/Pillow/pull/7733/commits/f7701e6596c9a2e7c9ed886621f01b3861d70990 for this.

Tests/helper.py
```diff
@@ -19,42 +20,40 @@
...
+        b.show()
+    elif uploader == "github_actions":
+        dir_errors = os.path.join(os.path.dirname(__file__), "errors")
+        os.makedirs(dir_errors, exist_ok=True)
+        tmpdir = tempfile.mkdtemp(dir=dir_errors)
+        a.save(os.path.join(tmpdir, "a.png"))
+        b.save(os.path.join(tmpdir, "b.png"))
+        return tmpdir
+    elif uploader == "aws":
+        return test_image_results.upload(a, b)
```
hugovk, line 50: Out of scope for this PR, but if the AWS Lambda is long gone, shall we remove this stuff?

https://github.com/wiredfool/test-image-results/pull/1#issuecomment-1898821181

hugovk (approved):

Tests/helper.py
radarhere: Ok, I've created https://github.com/python-pillow/Pillow/pull/7739 to remove it.

PR #7739, opened 2024-01-21
Do not support using test-image-results to upload images after test failures

radarhere:
Do not use [test-image-results](https://pypi.org/project/test-image-results/) to upload images after test failures.

Suggested in https://github.com/python-pillow/Pillow/pull/7733#discussion_r1460362160


hugovk (approved):
FYI @wiredfool.

PR #7732, opened 2024-01-19
Added type hints to Tests/check_*.py

radarhere:

PR #7743, opened 2024-01-22
Added type hints to Tests/test_font_*.py

radarhere:


Tests/test_font_pcf_charsets.py
```diff
@@ -14,38 +15,40 @@
...
+        223,
+        "witaj świecie",
+        "Tests/images/test_draw_pbm_ter_pl_target.png",
+    ),
+    "cp1250": (
+        250,
+        "witaj świecie",
+        "Tests/images/test_draw_pbm_ter_pl_target.png",
+    ),
 }
```
hugovk, line 34: This original version is clearer: `charsets[encoding]["glyph_count"]` instead of `charsets[encoding][0]`.

Can we keep it? Or use a `NamedTuple`?

https://docs.python.org/3/library/typing.html#typing.NamedTuple

radarhere: Ok, I've pushed a commit with the original dictionary.

Tests/test_font_pcf_charsets.py
```diff
@@ -14,7 +15,7 @@
 
 fontname = "Tests/fonts/ter-x20b.pcf"
 
-charsets = {
+charsets: dict[str, dict[str, int | str]] = {
```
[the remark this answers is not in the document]
radarhere: Ok, I've pushed a commit for this.

Tests/test_font_pcf.py
```diff
@@ -20,7 +21,7 @@
 pytestmark = skip_unless_feature("zlib")
 
 
-def save_font(request, tmp_path):
+def save_font(request: pytest.FixtureRequest, tmp_path: PosixPath) -> str:
```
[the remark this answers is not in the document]
radarhere: Ok, I've pushed a commit to update it, both the changes from here and #7732

Tests/check_j2k_overflow.py
```diff
@@ -1,13 +1,13 @@
...
-from pathlib import PosixPath
+from pathlib import Path
 
 import pytest
 
 from PIL import Image
 
 
-def test_j2k_overflow(tmp_path: PosixPath) -> None:
+def test_j2k_overflow(tmp_path: Path) -> None:
```
hugovk, line 10: I think we should we use `os.PathLike` instead of `pathlib.Path`? (here and below)

https://docs.python.org/3/library/os.html#os.PathLike

radarhere: https://docs.pytest.org/en/6.2.x/tmpdir.html
> tmp_path is a pathlib.Path object.

?

hugovk:
Thanks!

PR #7745, opened 2024-01-22
Update wl-paste handling and return None for some errors in grabclipboard() on Linux

nik012003:
Simpified logic and made it more robust against edge cases ( see the `allowed_errors` list ). 
Doing error checking this way, makes the behaviour of this function for x11 and wayland platforms more similar to darwin and windows systems.

### DIFF
--- a/src/PIL/ImageGrab.py
+++ b/src/PIL/ImageGrab.py
@@ -171,7 +171,9 @@ def grabclipboard():
             err = p.stderr
             if any(e in err for e in allowed_errors):
                 return None
-            msg = f"{args[0]} error: {err.strip().decode() if err else 'Unknown error'}"
+            msg = f"{args[0]} error"
+            if err:
+                msg += f": {err.strip().decode()}"
             raise ChildProcessError(msg)
 
         data = io.BytesIO(p.stdout)
--- a/src/PIL/ImageGrab.py
+++ b/src/PIL/ImageGrab.py
@@ -149,18 +149,7 @@ def grabclipboard():
             session_type = None
 
         if shutil.which("wl-paste") and session_type in ("wayland", None):
-            output = subprocess.check_output(["wl-paste", "-l"]).decode()
-            mimetypes = output.splitlines()
-            if "image/png" in mimetypes:
-                mimetype = "image/png"
-            elif mimetypes:
-                mimetype = mimetypes[0]
-            else:
-                mimetype = None
-
-            args = ["wl-paste"]
-            if mimetype:
-                args.extend(["-t", mimetype])
+            args = ["wl-paste", "-t", "image"]
         elif shutil.which("xclip") and session_type in ("x11", None):
             args = ["xclip", "-selection", "clipboard", "-t", "image/png", "-o"]
         else:
@@ -168,10 +157,19 @@ def grabclipboard():
             raise NotImplementedError(msg)
 
         p = subprocess.run(args, capture_output=True)
-        err = p.stderr
-        if err:
-            msg = f"{args[0]} error: {err.strip().decode()}"
+        err = p.stderr.decode()
+        if p.returncode != 0:
+            allowed_errors = [
+                "Nothing is copied",  # wl-paste, when the clipboard is empty
+                "not available",  # wl-paste/debian xclip, when an image isn't available
+                "cannot convert",  # xclip, when an image isn't available
+                "There is no owner",  # xclip, when the clipboard isn't initialized
+            ]
+            if any(e in err for e in allowed_errors):
+                return None
+            msg = f"{args[0]} error: {err.strip() if err else 'Unknown error'}"
             raise ChildProcessError(msg)
+
         data = io.BytesIO(p.stdout)
         im = Image.open(data)
         im.load()


src/PIL/ImageGrab.py
```diff
@@ -149,29 +149,27 @@ def grabclipboard():
...
         p = subprocess.run(args, capture_output=True)
-        err = p.stderr
-        if err:
-            msg = f"{args[0]} error: {err.strip().decode()}"
+        err = p.stderr.decode()
+        if p.returncode != 0:
+            allowed_errors = [
+                "Nothing is copied",  # wl-paste, when the clipboard is empty
+                "not available",  # wl-paste, when an image isn't avaiable
+                "cannot convert",  # xclip, when an image isn't avaiable
```
nulano, line 165:
```suggestion
                "not available",  # wl-paste, when an image isn't available
                "cannot convert",  # xclip, when an image isn't available
```

src/PIL/ImageGrab.py
```diff
@@ -149,29 +149,27 @@ def grabclipboard():
...
+        if p.returncode != 0:
+            allowed_errors = [
+                "Nothing is copied",  # wl-paste, when the clipboard is empty
+                "not available",  # wl-paste, when an image isn't available
+                "cannot convert",  # xclip, when an image isn't available
+                "There is no owner",  # xclip, when the clipboard isn't initialized
+            ]
+            if any(e in err for e in allowed_errors):
+                return None
+            msg = f"{args[0]} error: {err.strip() if err else 'Unkown error'}"
```
radarhere, line 170:
```suggestion
            msg = f"{args[0]} error: {err.strip() if err else 'Unknown error'}"
```

src/PIL/ImageGrab.py
```diff
@@ -149,29 +149,27 @@ def grabclipboard():
...
-        err = p.stderr
-        if err:
-            msg = f"{args[0]} error: {err.strip().decode()}"
+        err = p.stderr.decode()
+        if p.returncode != 0:
+            allowed_errors = [
+                "Nothing is copied",  # wl-paste, when the clipboard is empty
+                "not available",  # wl-paste, when an image isn't available
+                "cannot convert",  # xclip, when an image isn't available
+                "There is no owner",  # xclip, when the clipboard isn't initialized
```
radarhere, line 166: Could you describe how I can manually reproduce this error?

src/PIL/ImageGrab.py
```diff
@@ -149,29 +149,27 @@ def grabclipboard():
...
+            ]
+            if any(e in err for e in allowed_errors):
+                return None
+            msg = f"{args[0]} error: {err.strip() if err else 'Unknown error'}"
```
radarhere, line 170: Did you actually find a scenario where an error wasn't returned, meaning that you needed "Unknown error"?

radarhere:
> makes the behaviour of this function for x11 and wayland platforms more similar to darwin and windows systems.

You mean because `grabclipboard()` now returns `None` under some circumstances on Linux.

src/PIL/ImageGrab.py
nik012003: I can reproduce that by logging oit of gnome, and logging back in. Then run the xclip command. I can confirm this happens on fedora 39.

src/PIL/ImageGrab.py
nik012003: Not really, but I would like to be as possible, since we are calling userspace programs which could change their behaviour anytime

nik012003:
> > makes the behaviour of this function for x11 and wayland platforms more similar to darwin and windows systems.
> 
> You mean because `grabclipboard()` now returns `None` under some circumstances on Linux.

Yes, that behaviour is not consitant with the macos/windows implementation of the function, so if it does break if the developer isn't careful.(i.e. puts a try-except block all around it). With this PR the majority of x11 and wayland "non-errors" (i.e. clipboard empty, or not containg images) will return None like mac/win instead of raising some exceptions.

src/PIL/ImageGrab.py
```diff
@@ -149,29 +149,27 @@ def grabclipboard():
...
         p = subprocess.run(args, capture_output=True)
-        err = p.stderr
-        if err:
-            msg = f"{args[0]} error: {err.strip().decode()}"
+        err = p.stderr.decode()
+        if p.returncode != 0:
+            allowed_errors = [
+                "Nothing is copied",  # wl-paste, when the clipboard is empty
+                "not available",  # wl-paste, when an image isn't available
+                "cannot convert",  # xclip, when an image isn't available
```
radarhere, line 165: When I have text in the clipboard, I get a "not available" message.

So that seems like the "not available" comment should mention that it also applies to xclip?
Could you explain how to replicate "cannot convert"?

<video src="https://github.com/python-pillow/Pillow/assets/3112309/28e24a09-4d22-41ad-bd28-49f061c37ada" />

nik012003: Mhh, I get this result. I tried copying from multiple applications, gnome terminal, firefox, kitty, and they all respond with the "cannot convert" error... 
[Screencast from 2024-01-27 09-06-37.webm](https://github.com/python-pillow/Pillow/assets/10795335/55da4603-beaa-4289-8998-55d74e7c12ef)

I will look into the xclip code to be sure

nik012003: After further inspection, it seems that the version debian uses (and thus ubuntu) is different from the upstream one, and does indeed error out with `fprintf(stderr, "Error: target %s not available\n", atom_name);` instead of `fprintf(stderr, " cannot convert %s selection to target '%s'\n", selection_name, atom_name);`. I will update the comment accordingly

src/PIL/ImageGrab.py
```diff
@@ -149,29 +149,31 @@ def grabclipboard():
...
+                b"not available",
+                # xclip, when an image isn't available
+                b"cannot convert",
+                # xclip, when the clipboard isn't initialized
+                b"There is no owner",
+            ]
+            err = p.stderr
+            if any(e in err for e in allowed_errors):
+                return None
+            msg = f"{args[0]} error: {err.strip().decode() if err else 'Unknown error'}"
```
radarhere, line 174:
```suggestion
            msg = f"{args[0]} error"
            if err:
                msg += f": {err.strip().decode()}"
```
Feel free to ignore this suggestion if you prefer your current version.

src/PIL/ImageGrab.py
radarhere: I see - https://github.com/astrand/xclip/blob/b372f73579d30f9ba998ffd0a73694e7abe2c313/xclip.c#L741-L749 vs https://sources.debian.org/src/xclip/0.13-2/xclip.c/?hl=468#L466-L468

radarhere (approved):

radarhere:
I've created https://github.com/nik012003/Pillow/pull/3 to test two failing wl-paste scenarios.

hugovk (approved):

PR #7748, opened 2024-01-23
Restored testing of non-TrueType default font

radarhere:
Similar to https://github.com/python-pillow/Pillow/pull/7669/commits/ecd3948b45ca9c4c2e8b0b3cb58fe6af9798cfa2

This changes `TestDefaultFontLeak` to test the non-TrueType default font, as it would have been doing before #7354

### DIFF
--- a/Tests/test_font_leaks.py
+++ b/Tests/test_font_leaks.py
@@ -1,8 +1,10 @@
 from __future__ import annotations
 
-from PIL import Image, ImageDraw, ImageFont
+from PIL import Image, ImageDraw, ImageFont, _util
 
-from .helper import PillowLeakTestCase, skip_unless_feature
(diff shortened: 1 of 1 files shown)


Tests/test_font_leaks.py
```diff
@@ -31,5 +33,9 @@ class TestDefaultFontLeak(TestTTypeFontLeak):
     mem_limit = 1024  # k
 
     def test_leak(self):
+        if features.check_module("freetype2"):
+            ImageFont.core = _util.DeferredError(ImportError)
         default_font = ImageFont.load_default()
+        ImageFont.core = original_core
```
nulano, line 39: ?
```suggestion
        try:
            default_font = ImageFont.load_default()
        finally:
            ImageFont.core = original_core
```

PR #7760, opened 2024-01-29
Remove execute bit from `setup.py`

hugovk:
Invoking `setup.py` directly is deprecated:

* https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html

Instead, something like `pip install -e .` or `pip install .` should be used:

* https://pillow.readthedocs.io/en/stable/installation.html

See https://github.com/python-pillow/Pillow/pull/5896 and linked PRs for related changes.


nulano:
Would it be a good idea to also remove the shebang?
https://github.com/python-pillow/Pillow/blob/529487c244c64ee93ed7601eac9a9bbc3194827f/setup.py#L1

hugovk:
Good idea!

Our pre-commit has a "check that executables have shebangs" check, but didn't have the reverse.

So I've added the "check that scripts with shebangs are executable" check and removed the shebangs from some other files. Plus removed the shebang and exec bit from one `Tests/check_*` file, to match the others.

radarhere (approved):

PR #7762, opened 2024-01-30
Pin to Python 3.9.16-1

radarhere:
Our Cygwin jobs have been failing for several days - https://github.com/python-pillow/Pillow/actions/workflows/test-cygwin.yml

In looking at a discussion over at NumPy about this, there was a suggestion to use the egor-tension/setup-cygwin action and pin to Python 3.9.16-1 - https://github.com/numpy/numpy/issues/25708#issuecomment-1913885487

### DIFF
--- a/.github/workflows/test-cygwin.yml
+++ b/.github/workflows/test-cygwin.yml
@@ -47,7 +47,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Install Cygwin
-        uses: cygwin/cygwin-install-action@v4
+        uses: egor-tensin/setup-cygwin@v4
         with:
           platform: x86_64
           packages: >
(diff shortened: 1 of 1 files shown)


hugovk:
Sounds like a good idea. Has this been reported upstream to Cygwin?

radarhere:
I've just posted in the Cygwin mailing list - https://cygwin.com/pipermail/cygwin/2024-January/255273.html

hugovk:
So before, both 3.8 and 3.9 jobs were timing out: https://github.com/python-pillow/Pillow/actions/runs/7705171752

Now, we've switched to `egor-tensin/setup-cygwin` but does `python39=3.9.16-1` mean we install that for both 3.8 and 3.9 jobs? But we're still testing with either 3.8 or 3.9 for the relevant job.

https://github.com/python-pillow/Pillow/actions/runs/7708554081

radarhere:
python39 is already being installed in the 3.8 job, as some sort of dependency - https://github.com/python-pillow/Pillow/actions/runs/7705171752/job/20998682687#step:4:362. It's just explicit rather than implicit now so that the version can be pinned.

hugovk (approved):

radarhere:
This was reverted in https://github.com/python-pillow/Pillow/pull/7880

PR #7769, opened 2024-01-31

radarhere:


Tests/test_imagedraw.py
```diff
@@ -74,7 +74,14 @@ def test_mode_mismatch() -> None:
 
 @pytest.mark.parametrize("bbox", BBOX)
 @pytest.mark.parametrize("start, end", ((0, 180), (0.5, 180.4)))
-def test_arc(bbox, start, end) -> None:
+def test_arc(
+    bbox: tuple[tuple[int, int], tuple[int, int]]
+    | list[tuple[int, int]]
+    | list[int]
+    | tuple[int, int, int, int],
```
hugovk, line 81: This monster would be a good one to define as a type alias in `_typing.py`.

Yay295: Would `tuple[list[int], list[int]]` also be a valid option?

hugovk: Would that also be useful as an alias, for clearer re-use?

src/PIL/_typing.py
```diff
@@ -15,4 +16,7 @@ def __class_getitem__(cls, item: Any) -> type[bool]:
                 return bool
 
 
+CoordList = Union[Sequence[float], Sequence[Sequence[float]]]
```
hugovk, line 19: Maybe just `Coords`? It's shorter, and a sequence can be a not-a-list such as a tuple:

```suggestion
Coords = Union[Sequence[float], Sequence[Sequence[float]]]
```

radarhere: Sure, I've made this change.

hugovk:
Thanks!

PR #7773, opened 2024-02-03
Removed gcov codecov argument

radarhere:
codecov-action v4 no longer has the 'gcov' argument - https://github.com/codecov/codecov-action?tab=readme-ov-file#arguments. Instead, the README now states that the default behaviour is to run gcov
> plugin | plugins to run. Options: xcode, gcov, pycoverage. The default behavior runs them all.

It has been confirmed that 'gcov' is no longer necessary - https://github.com/codecov/codecov-action/issues/1247#issuecomment-1924946275

### DIFF
--- a/.github/workflows/test-docker.yml
+++ b/.github/workflows/test-docker.yml
@@ -105,7 +105,6 @@ jobs:
       with:
         flags: GHA_Docker
         name: ${{ matrix.docker }}
-        gcov: true
 
   success:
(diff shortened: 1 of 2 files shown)


radarhere:
Closing, as this is no longer needed after #7776

radarhere:
The pin from #7776 was removed in #8041, so this has been recreated as #8521

PR #7776, opened 2024-02-03
Pin codecov/codecov-action to v3.1.5

hugovk:
There's a breaking change in https://github.com/python-pillow/Pillow/pull/7770 from bumping codecov/codecov-action to v4 which affects us, requiring setting a token due to rate limiting:

https://github.com/codecov/codecov-action/releases/tag/v4.0.0

> Tokenless uploading is unsupported. However, PRs made from forks to the upstream public repos will support tokenless (e.g. contributors to OS projects do not need the upstream repo's Codecov token). This [doc](https://docs.codecov.com/docs/adding-the-codecov-token#github-actions) shows instructions on how to add the Codecov token.

v4 doesn't need a token for PRs from forks, but does for merges. This is why coverage has dropped to ~50% for merges, because it's only recording the coverage from AppVeyor:

https://app.codecov.io/gh/python-pillow/Pillow/commits

image

See:

* https://about.codecov.io/blog/january-product-update-updating-the-codecov-ci-uploaders-to-the-codecov-cli/ 
* https://github.com/codecov/feedback/issues/112#issuecomment-1919718118
* https://github.com/codecov/feedback/issues/112#issuecomment-1920204435
* https://github.com/codecov/feedback/issues/112#issuecomment-1924487786

I think we should stick to v3 for a while, to see what happens. 

This PR uses v3.1.5, which also bumped to Node.js 20 to fix the deprecation warnings. (v3=v3.1.6 went back down to Node.js 16.)

We could adjust the threshold for Codecov to "fail", so we don't miss huge drops like this in https://github.com/python-pillow/Pillow/pull/7770:

image

### DIFF
--- a/.github/workflows/test-cygwin.yml
+++ b/.github/workflows/test-cygwin.yml
@@ -142,7 +142,7 @@ jobs:
           bash.exe .ci/after_success.sh
 
       - name: Upload coverage
-        uses: codecov/codecov-action@v4
+        uses: codecov/codecov-action@v3.1.5
         with:
           file: ./coverage.xml
           flags: GHA_Cygwin
--- a/.github/workflows/test-docker.yml
+++ b/.github/workflows/test-docker.yml
@@ -101,7 +101,7 @@ jobs:
         MATRIX_DOCKER: ${{ matrix.docker }}
 
     - name: Upload coverage
-      uses: codecov/codecov-action@v4
+      uses: codecov/codecov-action@v3.1.5
       with:
         flags: GHA_Docker
         name: ${{ matrix.docker }}
--- a/.github/workflows/test-mingw.yml
+++ b/.github/workflows/test-mingw.yml
@@ -82,7 +82,7 @@ jobs:
           python3 -m pytest -vx --cov PIL --cov Tests --cov-report term --cov-report xml Tests
 
       - name: Upload coverage
-        uses: codecov/codecov-action@v4
+        uses: codecov/codecov-action@v3.1.5
         with:
           file: ./coverage.xml
           flags: GHA_Windows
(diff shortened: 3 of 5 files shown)


radarhere:
Thanks for catching this. Apologies for not noticing it in #7770

I find this interesting - https://github.com/codecov/feedback/issues/112#issuecomment-1920190706
> As of now, there are no plans to drop support for v3.

but if we're pinning to 3.1.5 for nodejs20, it is not as relevant as it could be.

It occurs to me that we could keep v4 for pushes and workflow dispatches, and only downgrade for pull requests? https://github.com/hugovk/Pillow/pull/111

hugovk:
> Thanks for catching this. Apologies for not noticing it in #7770

No problem, it wasn't obvious and I only noticed _after_ merging the same thing in some other projects!

> I find this interesting - [codecov/feedback#112 (comment)](https://github.com/codecov/feedback/issues/112#issuecomment-1920190706)

Yes, that might what we need to do in the end. I've not see rate-limiting cause problems for us, so I think it's better to stick to v3 for a bit longer to give time for things to settle, it might be that things will change.

> > As of now, there are no plans to drop support for v3.
> 
> but if we're pinning to 3.1.5 for nodejs20, it is not as relevant as it could be.

We can go back to v3=v3.1.6 if we don't mind the Node.js 16 warnings. In any case, I don't see pinning to v3.1.5 as a long-term solution, rather a stopgap for the time being.

> It occurs to me that we could keep v4 for pushes and workflow dispatches, and only downgrade for pull requests? [hugovk#111](https://github.com/hugovk/Pillow/pull/111)

This seems to add more complexity than we need right now, and will be extra work to make sure Renovate doesn't update the wrong bit. And hopefully we can choose only v3 or v4 or something soonish.

radarhere (approved):

radarhere:
This pin was later removed in #8041

PR #7781, opened 2024-02-05
Replace deprecated "extend-ignore" with "ignore"

radarhere:
https://docs.astral.sh/ruff/settings/#lint_extend-ignore
> [extend-ignore](https://docs.astral.sh/ruff/settings/#lint_extend-ignore)
> This option has been deprecated. The extend-ignore option is now interchangeable with ignore. Please update your configuration to use the ignore option instead.


radarhere:
Ok, I've cherry-picked this into that PR.

PR #7782, opened 2024-02-06
Release GIL while calling `WebPAnimDecoderGetNext`

evanmiller:
`WebPAnimDecoderGetNext` is a relatively expensive pure-C call that currently holds the Python GIL. Release it!

(Haven't tested this branch but it seemed like a straightforward change.)

### DIFF
--- a/src/_webp.c
+++ b/src/_webp.c
@@ -450,12 +450,16 @@ _anim_decoder_get_next(PyObject *self) {
     int timestamp;
     PyObject *bytes;
     PyObject *ret;
+    ImagingSectionCookie cookie;
     WebPAnimDecoderObject *decp = (WebPAnimDecoderObject *)self;
 
+    ImagingSectionEnter(&cookie);
     if (!WebPAnimDecoderGetNext(decp->dec, &buf, &timestamp)) {
+        ImagingSectionLeave(&cookie);
         PyErr_SetString(PyExc_OSError, "failed to read next frame");
         return NULL;
     }
+    ImagingSectionLeave(&cookie);
 
     bytes = PyBytes_FromStringAndSize(
         (char *)buf, decp->info.canvas_width * 4 * decp->info.canvas_height);
--- a/src/_webp.c
+++ b/src/_webp.c
@@ -455,7 +455,7 @@ _anim_decoder_get_next(PyObject *self) {
     WebPAnimDecoderObject *decp = (WebPAnimDecoderObject *)self;
 
     ImagingSectionEnter(&cookie);
-    ok = WebPAnimDecoderGetNext(decp->dec, &buf, &timestamp)
+    ok = WebPAnimDecoderGetNext(decp->dec, &buf, &timestamp);
     ImagingSectionLeave(&cookie);
     if (!ok) {
         PyErr_SetString(PyExc_OSError, "failed to read next frame");
--- a/src/_webp.c
+++ b/src/_webp.c
@@ -448,18 +448,19 @@ PyObject *
 _anim_decoder_get_next(PyObject *self) {
     uint8_t *buf;
     int timestamp;
+    int ok;
     PyObject *bytes;
     PyObject *ret;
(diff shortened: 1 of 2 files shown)


radarhere:
Testing with
```python
import timeit
from PIL import Image
im = Image.open('Tests/images/iss634.webp')

def decode():
  im._decoder.reset()
  for i in range(im.n_frames):
    im._decoder.get_next()

print(timeit.timeit(decode, number=1000))
```
I find it hard to say that this is definitively faster than main.

evanmiller:
Hi, I don't expect it to be faster than `main` in a single-threaded context – the point of releasing the GIL is to allow multiple threads to call `get_next()` concurrently (i.e. on different images).

evanmiller:
I think you'll want a test with code like

```
import concurrent
import timeit
from PIL import Image
images = [Image.open('Tests/images/iss634.webp') for _ in range(100)]

def decode(im):
  im._decoder.reset()
  for i in range(im.n_frames):
    im._decoder.get_next()

with concurrent.futures.ThreadPoolExecutor(max_workers=5) as pool:
  pool.map(decode, images)
```

(untested but hopefully you get the idea)

src/_webp.c
```diff
@@ -450,12 +450,16 @@ _anim_decoder_get_next(PyObject *self) {
...
+    ImagingSectionCookie cookie;
     WebPAnimDecoderObject *decp = (WebPAnimDecoderObject *)self;
 
+    ImagingSectionEnter(&cookie);
     if (!WebPAnimDecoderGetNext(decp->dec, &buf, &timestamp)) {
+        ImagingSectionLeave(&cookie);
         PyErr_SetString(PyExc_OSError, "failed to read next frame");
         return NULL;
     }
+    ImagingSectionLeave(&cookie);
```
radarhere, line 462:
```suggestion
    int ok;
    PyObject *bytes;
    PyObject *ret;
    ImagingSectionCookie cookie;
    WebPAnimDecoderObject *decp = (WebPAnimDecoderObject *)self;

    ImagingSectionEnter(&cookie);
    ok = WebPAnimDecoderGetNext(decp->dec, &buf, &timestamp)
    ImagingSectionLeave(&cookie);
    if (!ok) {
        PyErr_SetString(PyExc_OSError, "failed to read next frame");
        return NULL;
    }
```
Rather than writing `ImagingSectionLeave` twice, this follows a pattern from earlier in the file -
https://github.com/python-pillow/Pillow/blob/6782a07b8e404271dbd2e5ddf5fbb93e575ed2bc/src/_webp.c#L665-L670

radarhere (approved):

radarhere:
You're right, that test code does demonstrate a substantial speed increase.

evanmiller:
Updated w/ your code suggestion... I think you will have to re-approve the workflow run

src/_webp.c
```diff
@@ -448,11 +448,16 @@ PyObject *
...
     int timestamp;
+    int ok;
     PyObject *bytes;
     PyObject *ret;
+    ImagingSectionCookie cookie;
     WebPAnimDecoderObject *decp = (WebPAnimDecoderObject *)self;
 
-    if (!WebPAnimDecoderGetNext(decp->dec, &buf, &timestamp)) {
+    ImagingSectionEnter(&cookie);
+    ok = WebPAnimDecoderGetNext(decp->dec, &buf, &timestamp)
```
radarhere, line 458:
```suggestion
    ok = WebPAnimDecoderGetNext(decp->dec, &buf, &timestamp);
```

hugovk:
Please could you post a summary of the speed increase?

And let's include this in the release notes.

homm (approved):

radarhere:
I've pushed a commit with release notes, based on https://pillow.readthedocs.io/en/stable/releasenotes/9.3.0.html#release-gil-when-converting-images-using-matrix-operations

hugovk (approved):

PR #7790, opened 2024-02-10
Removed mypy excludes, except for olefile-related files

radarhere:
Removes the mypy excludes, except for FpxImagePlugin.py and MicImagePlugin.py, as olefile isn't typed.

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -26,6 +26,7 @@
 
 from __future__ import annotations
 
+import abc
 import atexit
 import builtins
 import io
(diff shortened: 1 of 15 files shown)


src/PIL/Image.py
```diff
@@ -2871,7 +2884,14 @@ class ImageTransformHandler:
...
 
-    pass
+    @abc.abstractmethod
+    def transform(
+        self,
+        size: tuple[int, int],
+        image: Image,
+        **options: dict[str, str | int | tuple[int, ...] | list[int]],
+    ) -> Image:
+        pass  # pragma: no cover
```
hugovk, line 2894: Let's also add `@abc.abstractmethod` to `exclude_also`:

```suggestion
        pass
```

And we can remove the pragma from `ImageShow`'s `@abc.abstractmethod` too.

hugovk (approved):
Thanks!

PR #7793, opened 2024-02-11
Added remaining type hints to Tests/test_image_*.py

radarhere:


Tests/test_image_access.py
```diff
@@ -138,12 +142,13 @@ def color(mode):
...
-    def check(self, mode, expected_color=None) -> None:
+    def check(self, mode: str, expected_color_int: int | None = None) -> None:
         if self._need_cffi_access and mode.startswith("BGR;"):
             pytest.skip("Support not added to deprecated module for BGR;* modes")
 
-        if not expected_color:
-            expected_color = self.color(mode)
+        expected_color = (
+            expected_color_int if expected_color_int is not None else self.color(mode)
+        )
```
hugovk, line 151: Switch around to use `is None` instead of `is not None`?
```suggestion
        expected_color = (
            self.color(mode) if expected_color_int is None else expected_color_int
        )
```

hugovk (approved):

PR #7796, opened 2024-02-13
Validate self.palette in has_transparency_data()

gdesmar:
I have a few files that I was able to open and save in Pillow 9.3.0, that result in an error with the newest (10.2.0) version. I believe that the newer has_transparency_data property can cause an error if the palette is not defined from opening a file.

```python
>>> from PIL import __version__, Image
>>> __version__
'10.2.0'
>>> img = Image.open("icon_6.ico")
>>> img.format
'ICO'
>>> img.height
256
>>> img.width
256
>>> img.save("output", format='WEBP')
  File ".../python3.12/site-packages/PIL/Image.py", line 2439, in save
  File ".../python3.12/site-packages/PIL/WebPImagePlugin.py", line 338, in _save
    im = im.convert("RGBA" if im.has_transparency_data else "RGB")
                              ^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../python3.12/site-packages/PIL/Image.py", line 1550, in has_transparency_data
    or (self.mode == "P" and self.palette.mode.endswith("A"))
                             ^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'mode'
```

I believe the solution is as simple as that, but feel free to disregard the PR if it needs a more complex fix.

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -1561,7 +1561,7 @@ class Image:
         """
         return (
             self.mode in ("LA", "La", "PA", "RGBA", "RGBa")
-            or (self.mode == "P" and self.palette.mode.endswith("A"))
+            or (self.mode == "P" and self.palette and self.palette.mode.endswith("A"))
             or "transparency" in self.info
         )
 


gdesmar:
I believe the older version of Pillow was assuming no transparency, but I just tested with VSCode and it looks to think that the icons do have transparency. I would like to point out that I often have to work with corrupted files, so hopefully those won't be too much of a problem.
I added the SHA256 hash for the three icons just in case. Here is the [non_palette_icons.zip](https://github.com/python-pillow/Pillow/files/14272142/non_palette_icons.zip) file.
| sha256 hash | name |
| ------------- | ------------- |
| e2c5c2c687c9f08a77b307da6f4f5dceaee5fb9a2c099adedac977dec8f29211 | icon_1.ico |
| 555ca02331d90dcaeaabc66f5e7932e4882916766d4375288c5b8018c7f3a4b0 | icon_2.ico |
| 20fde3f789dfeaffcabd1ae1978e6aee10c977cfdc3e8778af6da1e7b794a50b | icon_3.ico |

radarhere:
Thanks. I've created https://github.com/python-pillow/Pillow/pull/7798. That should fix your original error.

gdesmar:
Thank you for your detailed explanations and quick fixes. :)
I will go ahead and close this PR as I doubt it will be useful.

radarhere:
https://github.com/python-pillow/Pillow/pull/7798 has now been merged.

Issue #7802, opened 2024-02-15
Misleading deprecations documentation for getsize


stweil:
The [documentation](https://github.com/python-pillow/Pillow/blob/main/docs/deprecations.rst#font-size-and-offset-methods) suggests that this old code
```
width, height = font.getsize("Hello world")
```
should be replaced by
```
left, top, right, bottom = font.getbbox("Hello world")
width, height = right - left, bottom - top
```

A 1:1 replacement would require different new code
```
left, top, right, bottom = font.getbbox("Hello world")
width, height = right - left, bottom
```

The documentation should mention that and maybe link to this [comment](https://github.com/python-pillow/Pillow/issues/6555#issuecomment-1234515138).

radarhere:
I think the argument would be that the old methods were incorrect, which is why they were deprecated, and that we're encouraging users to use the correct new methods, with the new correct results.

Explaining to users how to keep things as they were seems to work against that purpose.

stweil:
Yes, sure, but a short comment which explains that the old method was incorrect would not hurt, but help. There might be old code which circumvented the deficits of the old method, and there is code which requires additional changes, not simply the suggested replacement.

https://github.com/Belval/TextRecognitionDataGenerator/pull/323 for example changes the code as suggested in the deprecations documentation, but that breaks the unit test of that software.

quark67:
I think that the documentation should include the transition code, as provided by OP. So old code can easily be converted without changing the output. And as it's not easy to understand that the old "height" is yet obtened by the new "bottom", the transition code added to the documentation can be very heplful. Or, if you would not add the transition code, perhaps a picture showing what is really the "bottom" in the `font.getbbox("Hello world")` can also help. Thanks. Big up to @stweil for his transition code in this issue. Very very very helpful.

radarhere:
The OP spoke about this change breaking unit tests in https://github.com/Belval/TextRecognitionDataGenerator/pull/323. I don't personally think that necessitates simulating Pillow's old behaviour - I think updating the unit test is the more thought out solution.

In the case of https://github.com/uvipen/ASCII-generator/issues/17, it looks like the code is drawing characters in order to evaluate the brightness. Pillow's old behaviour would have meant that some blank space was included above the character. I don't think you want to consider that in that calculation.

So I'm still not seeing why we should be encouraging users to continue doing things in a less-than-ideal way. I think this additional documentation should make it clear enough for users to figure it out how to do so on their own if they really want.

---

[...]

PR #7806, opened 2024-02-17
Describe difference between font size and bbox

radarhere:
Helps #7802

The user in the issue thinks that the documentation could use further detail about the transition from `font.getsize()` to `font.getbbox()`.

I've amended the deprecations and release notes documentation with the following
> Previously, the ``size`` methods returned a ``height`` that included the vertical offset of the text, while the new ``bbox`` methods explicitly distinguish this as a ``top`` offset.

While here, I also noticed that some of the example code didn't use the font declared earlier, so I've fixed that.

### DIFF
--- a/docs/deprecations.rst
+++ b/docs/deprecations.rst
@@ -232,10 +232,10 @@ Previous code::
 
     im = Image.new("RGB", (100, 100))
     draw = ImageDraw.Draw(im)
-    width, height = draw.textsize("Hello world")
+    width, height = draw.textsize("Hello world", font)
 
     width, height = font.getsize_multiline("Hello\nworld")
-    width, height = draw.multiline_textsize("Hello\nworld")
+    width, height = draw.multiline_textsize("Hello\nworld", font)
 
 Use instead::
 
@@ -247,9 +247,9 @@ Use instead::
 
     im = Image.new("RGB", (100, 100))
     draw = ImageDraw.Draw(im)
-    width = draw.textlength("Hello world")
+    width = draw.textlength("Hello world", font)
 
-    left, top, right, bottom = draw.multiline_textbbox((0, 0), "Hello\nworld")
+    left, top, right, bottom = draw.multiline_textbbox((0, 0), "Hello\nworld", font)
     width, height = right - left, bottom - top
 
 FreeTypeFont.getmask2 fill parameter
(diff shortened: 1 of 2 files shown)


stweil (approved):
That's better, thank you!

docs/deprecations.rst
```diff
@@ -253,8 +253,12 @@ Use instead::
     width, height = right - left, bottom - top
 
 Previously, the ``size`` methods returned a ``height`` that included the vertical
-offset of the text, while the new ``bbox`` methods explicitly distinguish this as a
-``top`` offset.
+offset of the text, while the new ``bbox`` methods distinguish this as a ``top``
+offset.
+
+.. image:: ./example/size_vs_bbox.png
+    :alt: Demonstration of size height vs bbox top and bottom
```
hugovk, line 260: Rather than saying what the image is, could you describe what the image conveys?

Imagine it read aloud to you when you can't see the image.

https://accessibility.huit.harvard.edu/describe-content-images

docs/releasenotes/9.2.0.rst
```diff
@@ -84,11 +84,19 @@ Use instead::
...
 
-    left, top, right, bottom = draw.multiline_textbbox((0, 0), "Hello\nworld")
+    left, top, right, bottom = draw.multiline_textbbox((0, 0), "Hello\nworld", font)
     width, height = right - left, bottom - top
 
+Previously, the ``size`` methods returned a ``height`` that included the vertical
+offset of the text, while the new ``bbox`` methods distinguish this as a ``top``
+offset.
+
+.. image:: ./example/size_vs_bbox.png
```
nulano, line 96:
```suggestion
.. image:: ../example/size_vs_bbox.png
```

nulano (commented):
I've created https://github.com/radarhere/Pillow/pull/23 to also show how anchors can be used to align text, which I have seen to be a common source of confusion with this change.

I originally wanted to write this back when these functions were being deprecated, but I didn't have time to get to it then and forgot to get back to it until now.

docs/deprecations.rst
radarhere: I've pushed a commit.
> In bbox methods, top measures the vertical distance above the text, while bottom measures that plus the vertical distance of the text itself. In size methods, height also measures the vertical distance above the text plus the vertical distance of the text itself.

I don't really like it, but it was a struggle to explain the meaning without using the words 'height', 'top' and 'bottom' in the definitions of those terms themselves.

hugovk: Thanks! It can be a challenge to write good alt text.

I read it through and understood the difference without looking at the image. 👍

PR #7808, opened 2024-02-18
Removed mypy excludes

radarhere:
Sequel to #7790

https://pypi.org/project/types-olefile/ now exists, so this can be used to remove the last of the mypy excludes.


hugovk:
Thanks also for contributing the olefile type hints to typeshed! 

Could be worth asking about upstreaming at https://github.com/decalage2/olefile

radarhere:
Ok, done - https://github.com/decalage2/olefile/issues/166

PR #7812, opened 2024-02-19
Remove _SupportsGetMesh from docs

radarhere:
After #7786, the docs have started failing in main - https://github.com/python-pillow/Pillow/actions/runs/7951921486/job/21705919173#step:8:254
> /opt/hostedtoolcache/Python/3.12.2/x64/lib/python3.12/site-packages/PIL/ImageOps.py:docstring of PIL.ImageOps.deform:1: WARNING: py:class reference target not found: PIL.ImageOps._SupportsGetMesh

This fixes it by adding `_SupportsGetMesh` to the docs.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -40,6 +40,10 @@ extensions = [
 
 intersphinx_mapping = {"python": ("https://docs.python.org/3", None)}
 
+autodoc_type_aliases = {
+    "_SupportsGetMesh": "object",
+}
+
 # The suffix(es) of source filenames.
 # You can specify multiple suffix as a list of string:
 # source_suffix = ['.rst', '.md']


hugovk:
I don't think `_SupportsGetMesh` is something users need to know about, and we don't want them to use it.

Can we exclude it from the docs instead?

radarhere:
Well, I suppose the simplest way to remove it from the docs is to just stop using `autofunction`.

I've updated the commit.

hugovk:
To avoid duplication, does `:exclude-members:` (https://stackoverflow.com/a/45531084/724176) or `:meta private:` or another answer from there help?

radarhere:
How about this? 
```python
autodoc_type_aliases = {
    "_SupportsGetMesh": "object",
}
```
Changes `_SupportsGetMesh` to `object` in the docs - https://pillow--7812.org.readthedocs.build/en/7812/reference/ImageOps.html#PIL.ImageOps.deform

hugovk:
Thanks, that could work.

But now I think about it some more, I'm going to completely contradict what I said earlier :)

> I don't think `_SupportsGetMesh` is something users need to know about, and we don't want them to use it.

If we're using `_SupportsGetMesh` as a type hint, this probably is something users might want to use, so should be accessible as a non-underscore thing?

@AlexWaygood: What do you think?

radarhere:
For Alex's reference, this is what we're talking about.

https://github.com/python-pillow/Pillow/blob/e1c5d57d5036fac07df2209e020d39ce4c771c62/src/PIL/ImageOps.py#L414-L424

radarhere:
> If we're using _SupportsGetMesh as a type hint, this probably is something users might want to use, so should be accessible as a non-underscore thing?

This suggestion may not fly because `_SupportsGetMesh` is specific to ImageOps, but, throwing an idea out there - we could move it to [_typing.py](https://github.com/python-pillow/Pillow/blob/main/src/PIL/_typing.py)?

radarhere:
#7841 was merged instead.

Issue #249, opened 2013-06-14
Issue warning(s) re: import _imaging breakage coming in 2.1.0


aclark4life:
@olt FYI we are releasing 2.1.0 in 2 weeks, at which point "import _imaging" will fail (instead please use "from PIL.Image import core as _imaging")

If you are interested, please help us get the word out about this change!

PR #7814, opened 2024-02-19
Remove outdated installation warnings

hugovk:
I think we can remove the three warnings from the top of https://pillow.readthedocs.io/en/stable/installation.html

---

> Pillow and PIL cannot co-exist in the same environment. Before installing Pillow, please uninstall PIL.

I seem to remember that around ten years ago, it used to be possible to install PIL from PyPI. 

It used to be that PyPI would allow installing files hosted elsewhere, but now all files must be uploaded to PyPI, and there are no downloadable files at https://pypi.org/project/PIL/.

I don't think it's possible now to install PIL and Pillow at the same time, even if you tried installing PIL from source.

---

> Pillow >= 1.0 no longer supports `import Image`. Please use `from PIL import Image` instead.

https://pypi.org/project/pillow/1.0/ was released in 2010. I don't think we need to warn about this any more.

---

> Pillow >= 2.1.0 no longer supports `import _imaging`. Please use `from PIL.Image import core as _imaging` instead.

https://pypi.org/project/pillow/2.1.0/ was released in 2013. I don't think we need to warn about this either.

### DIFF
--- a/docs/deprecations.rst
+++ b/docs/deprecations.rst
@@ -504,3 +504,27 @@ PIL.OleFileIO
 the upstream :pypi:`olefile` Python package, and replaced with an :py:exc:`ImportError` in 5.0.0
 (2018-01). The deprecated file has now been removed from Pillow. If needed, install from
 PyPI (eg. ``python3 -m pip install olefile``).
+
+import _imaging
+~~~~~~~~~~~~~~~
+
+.. versionremoved:: 2.1.0
+
+Pillow >= 2.1.0 no longer supports ``import _imaging``.
+Please use ``from PIL.Image import core as _imaging`` instead.
+
+Pillow and PIL
+~~~~~~~~~~~~~~
+
+Pillow and PIL cannot co-exist in the same environment.
+Before installing Pillow, please uninstall PIL.
+
+.. versionremoved:: 1.0.0
+
+import Image
+~~~~~~~~~~~~
+
+.. versionremoved:: 1.0.0
+
+Pillow >= 1.0 no longer supports ``import Image``.
+Please use ``from PIL import Image`` instead.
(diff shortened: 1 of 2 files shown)


hugovk:
I definitely remember juggling installs/uninstalls for the first one but it just cannot happen anymore:

```console
❯ pip install pil
ERROR: Could not find a version that satisfies the requirement pil (from versions: none)
ERROR: No matching distribution found for pil
```

For the second, I don't recall anyone tripping up over `import Image` (but I could have forgotten and it's hard to search for), and don't find any results for the symptom `ModuleNotFoundError: No module named 'Image'`.

And I also don't recall any for the third. There are many results for `import _imaging` in issues, but the couple I checked were from tracebacks of our own internal use.

If someone tripped up over it, we would have told them to do `from PIL.Image import core as _imaging`. But the only results for that are https://github.com/python-pillow/Pillow/issues/249 (2013) where we first added this warning for 2.1.0, and https://github.com/python-pillow/Pillow/pull/1016 (2014) where I suggested removing it.

aclark4life:
FWIW ``import Image`` is a Zope vs. PIL issue and predates the fork, see: https://mail.python.org/pipermail/image-sig/2010-August/006440.html. So, yeah, we don't need this stuff anymore but I also don't want to forget about it. So, I'd just like to see some link or section or something added to the documentation to coincide with the removal of the outdated warnings. But that's just me! I don't feel *that* strongly about it, and I'm willing to be overruled 😄

aclark4life:
Probably can just move them to here: https://pillow.readthedocs.io/en/stable/deprecations.html

hugovk:
Yeah, that'd work. Updated!

aclark4life:
Cool thanks

PR #7816, opened 2024-02-20

radarhere:


Tests/test_imagecms.py
```diff
@@ -335,12 +335,26 @@ def test_extended_information() -> None:
...
     p = o.profile
 
-    def assert_truncated_tuple_equal(tup1, tup2, digits: int = 10) -> None:
+    def assert_truncated_tuple_equal(
+        tup1: tuple[tuple[float, float, float], ...] | tuple[float],
+        tup2: (
+            tuple[tuple[tuple[float, float, float], ...], ...]
+            | tuple[tuple[float, float, float], ...]
+            | tuple[float]
+        ),
```
hugovk, line 344: Probably okay to make use of `tuple[Any]` or even `Any` here and in `truncate_tuple` to simplify things, but I don't mind too much :)

hugovk (approved):

Tests/test_imagecms.py
radarhere: Ok, I've pushed a commit with `tuple[Any, ...]`

Tests/test_imagecms.py
```diff
@@ -336,25 +337,13 @@ def test_extended_information() -> None:
...
         power = 10**digits
 
-        def truncate_tuple(
-            tuple_or_float: (
-                tuple[tuple[tuple[float, float, float], ...], ...]
-                | tuple[tuple[float, float, float], ...]
-                | tuple[float, ...]
-            )
-        ) -> tuple[tuple[float, ...], ...]:
+        def truncate_tuple(tuple_or_float: tuple[Any, ...]) -> tuple[Any, ...]:
```
hugovk, line 346: This parameter is called `tuple_or_float` but typed only as a `tuple`. 

Do we ever call it as a `float`? If not, shall we rename it?

radarhere: I've pushed a commit to rename it to `tuple_value`.

PR #7837, opened 2024-02-26
Updated openjpeg to 2.5.1

radarhere:


moubctez:
OpenJPEG 2.5.1 defines `OPJ_VERSION_MAJOR`, so this

```
/* OpenJPEG 2.0 doesn't have OPJ_VERSION_MAJOR */
#ifndef OPJ_VERSION_MAJOR
```
is no longer valid (in src/libImaging/Jpeg2KEncode.c and src/libImaging/Jpeg2KDecode.c).

radarhere:
From what I can see, `OPJ_VERSION_MAJOR` has been around since OpenJPEG 2.1.0 - so I think the comment meant 2.0.x, not 2.x.

So for this segment,

https://github.com/python-pillow/Pillow/blob/0fc9b9183fa8a97f37e0c06e64505e3813ce7c7d/src/libImaging/Jpeg2KDecode.c#L671-L675

the intention is to call the earlier version of `opj_stream_set_user_data` for earlier OpenJPEG versions.

PR #7843, opened 2024-02-29
winbuild: Refactor dependency versions into constants

hugovk:

 * Move global `__main__` calls into `main()` to avoid shadowing variables
 * Refactor version numbers into constants to reduce duplication and make updating easier

See also the last commit, I first put them as separate constants like:

```python
BROTLI_VERSION = "1.1.0"
FREETYPE_VERSION = "2.13.2"
...
```

I then refactored that into a single dict:

```python
V = {
    "BROTLI": "1.1.0",
    "FREETYPE": "2.13.2",
    ....
```

So we can refer to them like `V['BROTLI']` instead of `BROTLI_VERSION`, but I'm not sure which I prefer.

Thoughts?

cc @nulano


winbuild/build_prepare.py
```diff
@@ -109,13 +109,33 @@ def cmd_msbuild(
...
+
+
 # dependencies, listed in order of compilation
 DEPS = {
     "libjpeg": {
-        "url": SF_PROJECTS
-        + "/libjpeg-turbo/files/3.0.1/libjpeg-turbo-3.0.1.tar.gz/download",
-        "filename": "libjpeg-turbo-3.0.1.tar.gz",
-        "dir": "libjpeg-turbo-3.0.1",
+        "url": SF_PROJECTS + f"/libjpeg-turbo/files/{V['JPEGTURBO']}/"
```
nulano, line 135:
```suggestion
        "url": f"{SF_PROJECTS}/libjpeg-turbo/files/{V['JPEGTURBO']}/"
```

winbuild/build_prepare.py
```diff
@@ -157,9 +177,9 @@ def cmd_msbuild(
         "libs": [r"*.lib"],
     },
     "xz": {
-        "url": SF_PROJECTS + "/lzmautils/files/xz-5.4.5.tar.gz/download",
-        "filename": "xz-5.4.5.tar.gz",
-        "dir": "xz-5.4.5",
+        "url": SF_PROJECTS + f"/lzmautils/files/xz-{V['XZ']}.tar.gz/download",
```
nulano, line 180:
```suggestion
        "url": f"{SF_PROJECTS}/lzmautils/files/xz-{V['XZ']}.tar.gz/download",
```

winbuild/build_prepare.py
```diff
@@ -224,21 +244,24 @@ def cmd_msbuild(
         "libs": [r"libtiff\*.lib"],
     },
     "libpng": {
-        "url": SF_PROJECTS + "/libpng/files/libpng16/1.6.39/lpng1639.zip/download",
-        "filename": "lpng1639.zip",
-        "dir": "lpng1639",
+        "url": SF_PROJECTS + f"/libpng/files/libpng{V['LIBPNG_XY']}/{V['LIBPNG']}/"
```
nulano, line 247:
```suggestion
        "url": f"{SF_PROJECTS}/libpng/files/libpng{V['LIBPNG_XY']}/{V['LIBPNG']}/"
```

winbuild/build_prepare.py
```diff
@@ -282,9 +305,10 @@ def cmd_msbuild(
         "libs": [r"objs\{msbuild_arch}\Release Static\freetype.lib"],
     },
     "lcms2": {
-        "url": SF_PROJECTS + "/lcms/files/lcms/2.16/lcms2-2.16.tar.gz/download",
-        "filename": "lcms2-2.16.tar.gz",
-        "dir": "lcms2-2.16",
+        "url": SF_PROJECTS
+        + f"/lcms/files/lcms/{V['LCMS2']}/lcms2-{V['LCMS2']}.tar.gz/download",
```
nulano, line 309:
```suggestion
        "url": f"{SF_PROJECTS}/lcms/files/lcms/{V['LCMS2']}/lcms2-{V['LCMS2']}.tar.gz/download",  # noqa: E501
```

nulano (commented):
Looks good.

Two suggestions:
* Do we also want `V['libimagequant']` for consistency?
* I would also include `SF_PROJECTS` in the format strings now that it's not the only variable. IIRC it was done with a plus operator because it was the same length, but now it saves one character :).

hugovk:
> Two suggestions:
> 
> * Do we also want `V['libimagequant']` for consistency?

Yes, but let's do it when we're no longer pointing to a Git hash:

https://github.com/python-pillow/Pillow/blob/1b7bcfb6e71bc2c8a02ee4334b8dbd3b5db6e6e6/winbuild/build_prepare.py#L324-L344

> * I would also include `SF_PROJECTS` in the format strings now that it's not the only variable. IIRC it was done with a plus operator because it was the same length, but now it saves one character :).

Good idea, suggestions applied!

hugovk:
I've added a couple more commits to silence IDE (PyCharm) squiggly warnings.

winbuild/build_prepare.py
```diff
@@ -109,13 +109,33 @@ def cmd_msbuild(
...
+V = {
+    "BROTLI": "1.1.0",
+    "FREETYPE": "2.13.2",
+    "FRIBIDI": "1.0.13",
+    "HARFBUZZ": "8.3.0",
+    "JPEGTURBO": "3.0.1",
+    "LCMS2": "2.16",
+    "LIBPNG": "1.6.39",
+    "LIBWEBP": "1.3.2",
+    "LIBXCB": "1.16",
```
radarhere, line 121:
```suggestion
```
This isn't used?

winbuild/build_prepare.py
```diff
@@ -224,21 +243,24 @@ def cmd_msbuild(
...
         "build": [
             *cmds_cmake("png_static", "-DPNG_SHARED:BOOL=OFF", "-DPNG_TESTS:BOOL=OFF"),
-            cmd_copy("libpng16_static.lib", "libpng16.lib"),
+            cmd_copy(
+                f"libpng{V['LIBPNG_XY']}_static.lib", f"libpng{V['LIBPNG_XY']}.lib"
+            ),
         ],
         "headers": [r"png*.h"],
-        "libs": [r"libpng16.lib"],
+        "libs": [rf"libpng{V['LIBPNG_XY']}.lib"],
```
radarhere, line 258:
```suggestion
        "libs": [f"libpng{V['LIBPNG_XY']}.lib"],
```

winbuild/build_prepare.py
```diff
@@ -262,7 +284,7 @@ def cmd_msbuild(
                 "<UserDefines></UserDefines>": "<UserDefines>FT_CONFIG_OPTION_SYSTEM_ZLIB;FT_CONFIG_OPTION_USE_PNG;FT_CONFIG_OPTION_USE_HARFBUZZ;FT_CONFIG_OPTION_USE_BROTLI</UserDefines>",  # noqa: E501
                 "<UserIncludeDirectories></UserIncludeDirectories>": r"<UserIncludeDirectories>{dir_harfbuzz}\src;{inc_dir}</UserIncludeDirectories>",  # noqa: E501
```
radarhere, line 285: Isn't `inc_dir` out of scope here now?

winbuild/build_prepare.py
```diff
@@ -604,10 +634,10 @@ def build_dep_all() -> None:
...
     print()
     lines.append("@echo All Pillow dependencies built successfully!")
-    write_script("build_dep_all.cmd", lines)
+    write_script("build_dep_all.cmd", lines, prefs, verbose)
 
 
-if __name__ == "__main__":
+def main() -> None:
     winbuild_dir = os.path.dirname(os.path.realpath(__file__))
     pillow_dir = os.path.realpath(os.path.join(winbuild_dir, ".."))
```
radarhere, line 642:
```suggestion
```
Apart from being listed in the prefs, this is unused.

winbuild/build_prepare.py
```diff
@@ -718,12 +748,13 @@ def build_dep_all() -> None:
         "pillow_dir": pillow_dir,
```
radarhere, line 748:
```suggestion
```

winbuild/build_prepare.py
nulano: No, this gets expanded in `build_dep` here: https://github.com/python-pillow/Pillow/blob/ceb0c575e8c61a2d2b164f7f3e06d867804e304e/winbuild/build_prepare.py#L568

radarhere (approved):

winbuild/build_prepare.py
radarhere: Ah, I see, thanks

PR #7766, opened 2024-01-30
Test on macOS M1 where available

hugovk:
GitHub Actions now has macOS M1 runners:

* https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/
* https://github.blog/changelog/2024-01-30-github-actions-macos-14-sonoma-is-now-available/

Python 3.9.1 was the first to support Apple Silicon: 

* https://www.python.org/downloads/release/python-391/

But 3.9 is not currently available for M1:

* https://github.com/actions/setup-python/issues/696#issuecomment-1637587760

So let's test 3.10-3.13 on M1, and 3.8-3.9 on Intel.

I bumped the `PYTHONOPTIMIZE` jobs from 3.8/3.9 to 3.10/3.11 to avoid complicating the matrix includes/excludes. And needed to install Ghostscript, it wasn't already on the image.

Here's how long it took to test on M1 (left) and Intel (right). M1 is faster for all, around twice as fast for 3.13 and PyPy3.10, and nearly three times as fast for PyPy3.9!

image

That's reduced from about 49 minutes to 25 minutes!

PR #7852, opened 2024-03-03
Build macOS arm64 wheels natively

radarhere:
#7766 started testing macOS CI jobs on M1 where available.

We can go further, building macOS M1 wheels natively. This also means that we can start testing macOS M1 wheels, [apart from Python 3.8.](https://github.com/pypa/cibuildwheel/blob/ae2451a199bdec6636108b1110cf633157dba7e0/cibuildwheel/macos.py#L520-L523)


nulano:
Hmm, this seems to have added `libwebp.7.1.8.dylib` and `libsharpyuv.0.0.1.dylib` on x86_64:

image

and `libjpeg.8.3.2.dylib` on arm64:

image

For comparison, the list of dylibs was the same on both x86_64 and arm64 before:

image

radarhere:
Thanks for spotting. I've pushed a commit that should fix that.

PR #7856, opened 2024-03-04
Corrected exiv2.org links

radarhere:
https://exiv2.org/tags.html no longer works.

It looks like they're aware of this - https://github.com/Exiv2/exiv2/issues/2875

This fixes the links using the Wayback Machine.


hugovk:
Thanks, hopefully they'll fix it soon. I'll add this to the release milestone and merge it if it's not fixed by then.

PR #7872, opened 2024-03-12
Added alpha_quality argument when saving WebP images

### LINKED ISSUE #7868: Add webp export option alpha_quality
`cwebp` has a commandline flag of `-alpha_q <x>%` which uses lossy-compression for the alpha channel and can significantly reduce image size, and also seems to speed up image conversion in some cases. It would be great to have this ability in Pillow also.

Additionally an option similar to `-noalpha` where we know the image does not contain any alpha layer saves ~27 bytes per image - not particularly significant but again a nice-to-have for cwebp compatibility.

Thanks!

radarhere:
Resolves #7868

Adds `alpha_quality` argument when saving WebP images, as per https://developers.google.com/speed/webp/docs/api

### DIFF
--- a/Tests/test_file_webp_alpha.py
+++ b/Tests/test_file_webp_alpha.py
@@ -151,3 +151,15 @@ def test_write_unsupported_mode_PA(tmp_path: Path) -> None:
             target = im.convert("RGBA")
 
         assert_image_similar(image, target, 25.0)
+
+
+def test_alpha_quality(tmp_path: Path) -> None:
+    with Image.open("Tests/images/transparent.png") as im:
+        out = str(tmp_path / "temp.webp")
+        im.save(out)
+
+        out_quality = str(tmp_path / "quality.webp")
+        im.save(out_quality, alpha_quality=50)
+        with Image.open(out) as reloaded:
+            with Image.open(out_quality) as reloaded_quality:
+                assert reloaded.tobytes() != reloaded_quality.tobytes()
--- a/Tests/test_file_webp_animated.py
+++ b/Tests/test_file_webp_animated.py
@@ -188,3 +188,21 @@ def test_seek_errors() -> None:
 
         with pytest.raises(EOFError):
             im.seek(42)
+
+
+def test_alpha_quality(tmp_path: Path) -> None:
+    with Image.open("Tests/images/transparent.png") as im:
+        first_frame = Image.new("L", im.size)
+
+        out = str(tmp_path / "temp.webp")
+        first_frame.save(out, save_all=True, append_images=[im])
+
+        out_quality = str(tmp_path / "quality.webp")
+        first_frame.save(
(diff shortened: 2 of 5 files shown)


mzealey:
I'm not 100% sure if this is actually working and it's a little difficult to tell as unknown save flags are silently dropped.

I'm installing like:

```
git+https://github.com/python-pillow/Pillow.git#64c8c272711a0604f806b52941510ee4636fbae2=pillow
```

and a grep of the PIL directory shows alpha_quality is matching in the webp binary file output.

I run a small script like

```
from PIL import Image
image = Image.open("t.png")
image.save("alpha_trans.webp", "WEBP", quality=60, method=6, exif=b"", alpha_quality=50)
```

with something with significant alpha such as http://moya-discover-assets.moya.app/moya/icons/sport/sportsbetting.png results in  output like:

```
$ webpinfo -summary  alpha_trans.webp 
File: alpha_trans.webp
...
Chunk ALPH at offset     30, length   4960
Chunk VP8  at offset   4990, length   2874
...
```

Whereas with cwebp:

```
$ cwebp -alpha_q 50 -q 60 -m 6 t.png -o cwebp.webp
$ webpinfo -summary  cwebp.webp 
Chunk ALPH at offset     30, length   1952
Chunk VP8  at offset   1982, length   2874
```

So I suspect somehow the flag is not making it through to the underlying library or there is some other variable that also needs tweaking if not going for alpha_quality==100?

radarhere:
When I install this PR and run that code with that image, I get
```
Chunk ALPH at offset     30, length   1952
```

So my theory would be this is working, but it's simply not installed correctly on your machine. If you run `import PIL; print(PIL.__version__)` with this PR, you should get 10.3.0.dev0. If you don't, it means you're using another version of Pillow.

Would it help if I put together a wheel of this for you? What OS/Python version are you using?

radarhere:
I ran the [Docker scenario](https://github.com/radarhere/docker-images/commit/5f3515daea52ebe5f677749d9228dddaa12274dc) you described, and received the [same filesize output](https://github.com/radarhere/docker-images/actions/runs/8357469479/job/22876673623#step:6:1999) whether I specified `alpha_quality` or not.

Then I tried [changing](https://github.com/radarhere/docker-images/commit/090448b37830ba067f4b413b284a415233dc5cb2) the pip requirement to
```
pillow @ git+https://github.com/radarhere/Pillow.git@webp_alpha_quality
```
and the [output file size changed.](https://github.com/radarhere/docker-images/actions/runs/8357476334/job/22876693908#step:6:1996)

I don't know why one requirement works and the other doesn't, but give that a try.

mzealey:
Yes thank you this fixes it. Very strange - I'm sure the #commit approach worked with some other projects but it could perhaps be because of the dev on a different repo. All lgtm.

PR #7874, opened 2024-03-13
Moved version definition from _version.py to __init__.py

radarhere:
There seems to be a [consensus](https://github.com/python-pillow/Pillow/pull/7644#discussion_r1436890926) that `__version__` should truly live in `__init__.py`, not `_version.py`.

So this PR moves the version definition, and just removes `_version.py`.


hugovk:
Whilst we still need to edit it as part of the release, I'd prefer to keep it in the dedicated `_version.py`, it makes the release that bit easier to do.

PR #7875, opened 2024-03-13
CI: Update Cygwin NumPy to 1.26 in Python 3.9

DWesl:
Removes the NumPy-pinning requirement noted in #7403.

 * NumPy can use the Netlib BLAS interface, and tests against it regularly.  Cygwin provides Netlib interfaces for BLAS and Lapack in `liblapack-devel`.  NumPy recently changed how to specify BLAS versions: [reference here](https://numpy.org/doc/1.26/user/building.html) and [elaboration here](https://github.com/numpy/numpy/blob/main/meson_options.txt).  [Pip can pass arguments through to sub-processes](https://github.com/numpy/numpy/issues/24703#issuecomment-1722379388).  Combining those, it should be possible to allow CI to use the most recent NumPy version.  This is a slight tweak of the fix proposed in #7403.

### DIFF
--- a/.github/workflows/test-cygwin.yml
+++ b/.github/workflows/test-cygwin.yml
@@ -81,6 +81,7 @@ jobs:
             wget
             xorg-server-extra
             zlib-devel
+            ninja cmake meson
 
       - name: Add Lapack to PATH
         uses: egor-tensin/cleanup-path@v4
--- a/.github/workflows/test-cygwin.yml
+++ b/.github/workflows/test-cygwin.yml
@@ -119,7 +119,6 @@ jobs:
         shell: C:/tools/cygwin/bin/dash.exe -l "{0}"
         run: |
           python3 -m pip install -U numpy --config-settings=setup-args="-Dblas=blas" --config-settings=setup-args="-Dlapack=lapack"
-          python3 -c "import numpy as np; print(np.__version__)"
 
       - name: Build
         shell: C:/tools/cygwin/bin/bash.exe -eo pipefail -o igncr "{0}"
--- a/.github/workflows/test-cygwin.yml
+++ b/.github/workflows/test-cygwin.yml
@@ -52,6 +52,7 @@ jobs:
       - name: Install Cygwin
         uses: egor-tensin/setup-cygwin@v4
         with:
+          install-dir: 'C:/tools/cygwin'
           packages: >
             gcc-g++
             ghostscript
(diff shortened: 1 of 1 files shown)


radarhere:
> Fixes https://github.com/python-pillow/Pillow/pull/7403.

Ah, ok, you mean that the PR pinned NumPy on Cygwin, and this removes the pinning. Normally, I would only think of issues as being "fixed", not PRs that passed and were merged.

.github/workflows/test-cygwin.yml
```diff
@@ -116,7 +117,7 @@ jobs:
       - name: Upgrade NumPy
         shell: dash.exe -l "{0}"
         run: |
-          python3 -m pip install -U "numpy<1.26"
+          python3 -m pip install -U "numpy" --config-settings=setup-args="-Dblas=blas" --config-settings=setup-args="-Dlapack=lapack"
```
radarhere, line 120:
```suggestion
          python3 -m pip install -U numpy --config-settings=setup-args="-Dblas=blas" --config-settings=setup-args="-Dlapack=lapack"
```

radarhere:
I've updated the PR title, as this only affects Python 3.9 - [NumPy 1.24 is the last version to support Python 3.8](https://endoflife.date/numpy#python-support), so this makes no difference for 3.8.

radarhere:
It turns out there is a problem, both in this PR and in main, and for both Python 3.8 and 3.9 - [NumPy is not detected by the tests](https://github.com/python-pillow/Pillow/actions/runs/8248697493/job/22559567810#step:13:4362).

DWesl:
Looks like the switch to the egor-tensin Cygwin install action didn't catch all the places the default install path got used: in particular, the egor-tensin/cleanup-path specification of the new path, including the location of the BLAS and Lapack runtimes.

.github/workflows/test-cygwin.yml
```diff
@@ -81,56 +82,58 @@ jobs:
...
-          bash.exe .ci/install.sh
+          C:/tools/cygwin/bin/bash.exe .ci/install.sh
 
       - name: Upgrade NumPy
-        shell: dash.exe -l "{0}"
+        shell: C:/tools/cygwin/bin/dash.exe -l "{0}"
         run: |
-          python3 -m pip install -U "numpy<1.26"
+          python3 -m pip install -U numpy --config-settings=setup-args="-Dblas=blas" --config-settings=setup-args="-Dlapack=lapack"
+          python3 -c "import numpy as np; print(np.__version__)"
```
hugovk, line 122: Shall we remove this debugging line?
```suggestion
```

hugovk (commented):
Nice work :)

.github/workflows/test-cygwin.yml
DWesl: This has the added benefit of testing whether I properly fixed the pip cache: the last run took around 30 minutes, this one should take around 15.

radarhere:
I've created https://github.com/python-pillow/Pillow/pull/7880 as an alternative to this.

hugovk:
We've merged https://github.com/python-pillow/Pillow/pull/7880, but thanks for this PR that prompted that one :)

PR #7880, opened 2024-03-15
Removed Python and NumPy pinning on Cygwin

radarhere:
Alternative to #7875

Two Cygwin changes
1. Reverts #7762, where Python was pinned to Python 3.9.16-1
2. Removes the pinning of NumPy from #7403, but also goes further, and just uses the packaged NumPy. NumPy was originally upgraded in [a commit entitled "CI: Try to get Cygwin workflow working."](https://github.com/python-pillow/Pillow/pull/5878/commits/2d6dee1dae7c6f4ce639e1042c05152ef5d14c9e), so if there wasn't another need for this, then perhaps testing the package version is more valuable?

This also decreases the build time from [31m54s](https://github.com/python-pillow/Pillow/actions/runs/8280312072/job/22656563787) to [10m13s.](https://github.com/python-pillow/Pillow/actions/runs/8294540097/job/22699754552)

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -702,7 +702,7 @@ class Image:
                     pass
                 else:
                     if parse_version(numpy.__version__) < parse_version("1.23"):
-                        warnings.warn(e)
+                        warnings.warn(str(e))
             raise
         new["shape"], new["typestr"] = _conv_type_shape(self)
         return new
--- a/.github/workflows/test-cygwin.yml
+++ b/.github/workflows/test-cygwin.yml
@@ -50,7 +50,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Install Cygwin
-        uses: egor-tensin/setup-cygwin@v4
+        uses: cygwin/cygwin-install-action@v4
         with:
           packages: >
             gcc-g++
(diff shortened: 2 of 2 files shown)


DWesl:
Tests pass, and include NumPy.  I think 1.21 had problems with unaligned load, but that seems gone now.

src/PIL/Image.py
```diff
@@ -702,7 +702,7 @@ def __array_interface__(self):
                     pass
                 else:
                     if parse_version(numpy.__version__) < parse_version("1.23"):
-                        warnings.warn(e)
+                        warnings.warn(str(e))
```
hugovk, line 705: How did this look before/after?

hugovk (approved):

src/PIL/Image.py
radarhere: [Without this change](https://github.com/radarhere/Pillow/commit/4005a52b54d1cd1ae7897c4120af570f4db0f7f8), I get https://github.com/radarhere/Pillow/actions/runs/8296217068/job/22704816536#step:11:2665
> TypeError: Warning must be str or Warning, got OSError('image file is truncated (8 bytes not processed)') (type OSError)

Issue #7890, opened 2024-03-21
What does the integer in Kmeans in Image.quantize mean?


AeroDEmi:
I'm having trouble understanding the Integer argument of Kmeans in [Image.quantize](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.quantize)

Yay295:
So the C quantize function passes the `kmeans` value to the `ImagingQuantize` function,

https://github.com/python-pillow/Pillow/blob/ca973709a079b5152b9670e5cd684003aa71f913/src/_imaging.c#L1685-L1700

which passes it to the `quantize` or `quantize2` functions,

https://github.com/python-pillow/Pillow/blob/ca973709a079b5152b9670e5cd684003aa71f913/src/libImaging/Quant.c#L1761-L1782

which each do the same thing - check if it's 0, and if not, subtract 1 as they pass it to the `k_means` function.

https://github.com/python-pillow/Pillow/blob/ca973709a079b5152b9670e5cd684003aa71f913/src/libImaging/Quant.c#L1474-L1476

In the `k_means` function the value is now called `threshold`,

https://github.com/python-pillow/Pillow/blob/ca973709a079b5152b9670e5cd684003aa71f913/src/libImaging/Quant.c#L1158-L1165

and it determines the number of pixel changes that are allowed during each loop of the k-means quantization algorithm, ending when the number of changes is less than or equal to the threshold.

Yay295:
It should probably be mentioned that `dither` is only used when `palette` is given, and if `palette` is given then `colors`, `method`, and `kmeans` are ignored.

radarhere:
> It should probably be mentioned that `dither` is only used when `palette` is given

#5836 is actually an open issue that expects this not to be the case.

Yay295:
It looks like `kmeans` needs to be non-negative, but I don't see a check for that. I think it would get stuck in an infinite loop. It also seems like it might be possible to get stuck in a loop if `kmeans` is small and the number of changes in each loop never gets below that value.

Yay295:
To answer your original question @AeroDEmi, the `kmeans` parameter determines how the quantization uses [k-means clustering](https://en.wikipedia.org/wiki/K-means_clustering). If it's 0, then it doesn't use k-means clustering at all. Otherwise, `kmeans` should be an integer from 1 though the number of pixels in your image, with smaller values meaning that it will try harder to find the best color palette to use.

[...]

PR #7891, opened 2024-03-22
Raise ValueError if kmeans is negative

radarhere:
https://github.com/python-pillow/Pillow/issues/7890#issuecomment-2014337747 found that if `kmeans` if negative, `im.quantize()` can become stuck in an infinite loop.

Investigating, I was able to reproduce this for `MEDIANCUT` and `MAXCOVERAGE`.

This PR updates the docstring to specify that `kmeans` should not be negative, raises a ValueError if `kmeans` is negative, and to be safe, also updates the C code to not use negative kmeans either.

### DIFF
--- a/Tests/test_image_quantize.py
+++ b/Tests/test_image_quantize.py
@@ -99,6 +99,10 @@ def test_quantize_dither_diff() -> None:
 )
 def test_quantize_kmeans(method) -> None:
     im = hopper()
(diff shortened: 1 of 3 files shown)


hugovk:
Thanks, shall we add a test that successfully calls `quantize` with `kmeans=0`, and `kmeans=1` or some other positive integer?

hugovk (approved):
Thank you!

PR #7897, opened 2024-03-23
Type hints: Add PYI (flake8-pyi) to Ruff and fix errors

hugovk:
Add PYI to Ruff to run the flake8-pyi linting rules related to type hints and fix the errors:

```
Tests/test_file_libtiff.py:246:14: PYI024 Use `typing.NamedTuple` instead of `collections.namedtuple`
Tests/test_lib_pack.py:18:18: PYI041 Use `float` instead of `int | float`
Tests/test_lib_pack.py:242:18: PYI041 Use `float` instead of `int | float`
src/PIL/PdfParser.py:85:5: PYI024 Use `typing.NamedTuple` instead of `collections.namedtuple`
src/PIL/TarIO.py:61:9: PYI034 `__enter__` methods in classes like `TarIO` usually return `self` at runtime
src/PIL/TiffTags.py:24:15: PYI024 Use `typing.NamedTuple` instead of `collections.namedtuple`
src/PIL/_imaging.pyi:1:1: PYI044 `from __future__ import annotations` has no effect in stub files, since type checkers automatically treat stubs as having those semantics
src/PIL/_imagingcms.pyi:1:1: PYI044 `from __future__ import annotations` has no effect in stub files, since type checkers automatically treat stubs as having those semantics
src/PIL/_imagingft.pyi:1:1: PYI044 `from __future__ import annotations` has no effect in stub files, since type checkers automatically treat stubs as having those semantics
src/PIL/_imagingmath.pyi:1:1: PYI044 `from __future__ import annotations` has no effect in stub files, since type checkers automatically treat stubs as having those semantics
src/PIL/_imagingmorph.pyi:1:1: PYI044 `from __future__ import annotations` has no effect in stub files, since type checkers automatically treat stubs as having those semantics
src/PIL/_webp.pyi:1:1: PYI044 `from __future__ import annotations` has no effect in stub files, since type checkers automatically treat stubs as having those semantics
Found 12 errors.
```

I skipped [PYI034](https://docs.astral.sh/ruff/rules/non-self-return-type/). The fix is to return `typing.Self`, but that was only added in 3.10, or `typing_extensions.Self` from the backport, which we're not using. Open to other solutions.

* https://docs.astral.sh/ruff/rules/#flake8-pyi-pyi
* https://github.com/PyCQA/flake8-pyi
* https://docs.python.org/3/library/typing.html#typing.Self

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -117,7 +117,7 @@ ignore = [
   "E221", # Multiple spaces before operator
   "E226", # Missing whitespace around arithmetic operator
   "E241", # Multiple spaces after ','
-  "PYI034", # flake8-pyi: typing.Self added in Python 3.10
+  "PYI034", # flake8-pyi: typing.Self added in Python 3.11
 ]
 
 [tool.ruff.lint.per-file-ignores]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -106,6 +106,7 @@ select = [
   "ISC", # flake8-implicit-str-concat
   "LOG", # flake8-logging
   "PGH", # pygrep-hooks
+  "PYI", # flake8-pyi
   "RUF100", # unused noqa (yesqa)
   "UP", # pyupgrade
   "W", # pycodestyle warnings
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -117,6 +117,7 @@ ignore = [
   "E221", # Multiple spaces before operator
   "E226", # Missing whitespace around arithmetic operator
   "E241", # Multiple spaces after ','
+  "PYI034", # flake8-pyi: typing.Self added in Python 3.10
 ]
 
 [tool.ruff.lint.per-file-ignores]
--- a/src/PIL/TiffTags.py
+++ b/src/PIL/TiffTags.py
@@ -18,10 +18,18 @@
 ##
 from __future__ import annotations
 
-from collections import namedtuple
+from typing import NamedTuple
 
 
-class TagInfo(namedtuple("_TagInfo", "value name type length enum")):
+class _TagInfo(NamedTuple):
+    value: int
+    name: str
+    type: int
+    length: int
+    enum: dict[int, str]
+
+
+class TagInfo(_TagInfo):
     __slots__: list[str] = []
 
     def __new__(cls, value=None, name="unknown", type=None, length=None, enum=None):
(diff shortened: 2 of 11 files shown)


pyproject.toml
```diff
@@ -116,6 +117,7 @@ ignore = [
   "E221", # Multiple spaces before operator
   "E226", # Missing whitespace around arithmetic operator
   "E241", # Multiple spaces after ','
+  "PYI034", # flake8-pyi: typing.Self added in Python 3.10
```
radarhere, line 120:
```suggestion
  "PYI034", # flake8-pyi: typing.Self added in Python 3.11
```
See https://docs.python.org/3/library/typing.html#typing.Self

src/PIL/TiffTags.py
```diff
@@ -18,10 +18,18 @@
...
+from typing import NamedTuple
 
 
-class TagInfo(namedtuple("_TagInfo", "value name type length enum")):
+class _TagInfo(NamedTuple):
+    value: int
+    name: str
+    type: int
+    length: int
+    enum: dict[int, str]
```
radarhere, line 29:
```suggestion
    value: int | None
    name: str
    type: int | None
    length: int | None
    enum: dict[int, str] | None
```

It would seem that `value`, `type`, `length` and `enum` can all be `None`
https://github.com/python-pillow/Pillow/blob/58c4c757a2e969f52d23586e731571887cad2db0/src/PIL/TiffTags.py#L27-L28

src/PIL/TiffTags.py
```diff
@@ -18,10 +18,18 @@
...
+from typing import NamedTuple
 
 
-class TagInfo(namedtuple("_TagInfo", "value name type length enum")):
+class _TagInfo(NamedTuple):
+    value: int | None
+    name: str
+    type: int | None
+    length: int | None
+    enum: dict[int, str] | None
```
radarhere, line 29:
```suggestion
    enum: dict[str, int]
```
Oh, sorry, I was wrong - `enum` can't be `None`.
`enum or {}` is right there in the code I referenced.

https://github.com/python-pillow/Pillow/blob/58c4c757a2e969f52d23586e731571887cad2db0/src/PIL/TiffTags.py#L27-L28

But also, it is `str` is the key, not the value in the dictionary.

radarhere (approved):

PR #7911, opened 2024-03-27
Use setup_module() to open images so they aren't opened if skipped

Yay295:
This entire file is skipped if Qt isn't available. This means `teardown_module()` is also skipped. But opening the images isn't skipped because they're not in a function. By moving the image loading to `setup_module()` it will also be skipped when Qt isn't available.

### DIFF
--- a/Tests/test_image_fromqimage.py
+++ b/Tests/test_image_fromqimage.py
@@ -16,12 +16,13 @@ pytestmark = pytest.mark.skipif(
     not ImageQt.qt_is_installed, reason="Qt bindings are not installed"
 )
 
-ims = [
-    hopper(),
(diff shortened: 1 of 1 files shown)


Tests/test_image_fromqimage.py
```diff
@@ -16,11 +16,13 @@
     not ImageQt.qt_is_installed, reason="Qt bindings are not installed"
 )
 
-ims = [
-    hopper(),
-    Image.open("Tests/images/transparent.png"),
-    Image.open("Tests/images/7x13.png"),
-]
+ims = []
```
nulano, line 19:
```suggestion
ims: list[Image.Image] = []
```

hugovk:
Thanks!

PR #7912, opened 2024-03-28
Use `functools.lru_cache` for `hopper()`

hugovk:
Replace the custom caching for `hopper()` with `functools.lru_cache`.

https://docs.python.org/3/library/functools.html#functools.lru_cache


Tests/helper.py
```diff
@@ -250,25 +251,27 @@ def tostring(im: Image.Image, string_format: str, **options: Any) -> bytes:
...
+    return _cached_hopper(mode).copy()
+
+
+@lru_cache(maxsize=None)
+def _cached_hopper(mode: str | None = None) -> Image.Image:
     if mode is None:
         # Always return fresh not-yet-loaded version of image.
         # Operations on not-yet-loaded images is separate class of errors
         # what we should catch.
         return Image.open("Tests/images/hopper.ppm")
```
nulano, line 267: This part should probably be in the `hopper()` function not `_cached_hopper()`.

Tests/helper.py
```diff
@@ -250,25 +251,27 @@ def tostring(im: Image.Image, string_format: str, **options: Any) -> bytes:
     return out.getvalue()
 
 
-def hopper(mode: str | None = None, cache: dict[str, Image.Image] = {}) -> Image.Image:
+def hopper(mode: str | None = None) -> Image.Image:
+    # Use caching to reduce reading from disk but so an original copy is
+    # returned each time and the cached image isn't modified by tests
+    # (for fast, isolated, repeatable tests).
```
Yay295, line 257: This seems to be a run-on sentence, and some of the words seem to be misplaced. Perhaps this instead:

> Use caching to reduce reading from disk, but return a copy so that the cached image isn't modified by the tests.

Tests/helper.py
hugovk: With:
```python
def hopper(mode: str | None = None) -> Image.Image:
    # Use caching to reduce reading from disk but so an original copy is
    # returned each time and the cached image isn't modified by tests
    # (for fast, isolated, repeatable tests).
    if mode is None:
        # Always return fresh not-yet-loaded version of image.
        # Operations on not-yet-loaded images is separate class of errors
        # what we should catch.
        return Image.open("Tests/images/hopper.ppm")
    return _cached_hopper(mode).copy()
```

We get:
```python
============================================================================= ERRORS ==============================================================================
___________________________________________________________ ERROR collecting Tests/test_file_libtiff.py ___________________________________________________________
Tests/test_file_libtiff.py:53: in <module>
    class TestFileLibTiff(LibTiffTestCase):
Tests/test_file_libtiff.py:516: in TestFileLibTiff
    @pytest.mark.parametrize("im", (hopper("P"), Image.new("P", (1, 1), "#000")))
Tests/helper.py:263: in hopper
    return _cached_hopper(mode).copy()
Tests/helper.py:273: in _cached_hopper
    im = _cached_hopper().convert(mode)
Tests/helper.py:270: in _cached_hopper
    elif mode[:4] == "I;16":
E   TypeError: 'NoneType' object is not subscriptable
________________________________________________________ ERROR collecting Tests/test_file_libtiff_small.py ________________________________________________________
Tests/test_file_libtiff_small.py:8: in <module>
    from .test_file_libtiff import LibTiffTestCase
<frozen importlib._bootstrap>:1360: in _find_and_load
    ???
<frozen importlib._bootstrap>:1331: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:935: in _load_unlocked
    ???
/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/_pytest/assertion/rewrite.py:178: in exec_module
    exec(co, module.__dict__)
Tests/test_file_libtiff.py:53: in <module>
    class TestFileLibTiff(LibTiffTestCase):
Tests/test_file_libtiff.py:516: in TestFileLibTiff
    @pytest.mark.parametrize("im", (hopper("P"), Image.new("P", (1, 1), "#000")))
Tests/helper.py:263: in hopper
    return _cached_hopper(mode).copy()
Tests/helper.py:273: in _cached_hopper
    im = _cached_hopper().convert(mode)
Tests/helper.py:270: in _cached_hopper
    elif mode[:4] == "I;16":
E   TypeError: 'NoneType' object is not subscriptable
_________________________________________________________ ERROR collecting Tests/test_image_transpose.py __________________________________________________________
Tests/test_image_transpose.py:12: in <module>
    mode: helper.hopper(mode).crop((0, 0, 121, 127)).copy()
Tests/helper.py:263: in hopper
    return _cached_hopper(mode).copy()
Tests/helper.py:273: in _cached_hopper
    im = _cached_hopper().convert(mode)
Tests/helper.py:270: in _cached_hopper
    elif mode[:4] == "I;16":
E   TypeError: 'NoneType' object is not subscriptable
===================================================================== short test summary info =====================================================================
ERROR Tests/test_file_libtiff.py - TypeError: 'NoneType' object is not subscriptable
ERROR Tests/test_file_libtiff_small.py - TypeError: 'NoneType' object is not subscriptable
ERROR Tests/test_image_transpose.py - TypeError: 'NoneType' object is not subscriptable
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 3 errors during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
======================================================================== 3 errors in 2.82s ========================================================================
```

Tests/helper.py
hugovk: I think the original does make sense, but it can definitely be improved, and your suggestion is much clearer 👍 

Will update.

Tests/helper.py
nulano: Maybe put it in both?
That way the cached version is used within `_cached_hopper(...)` and a new instance is created every time `hopper(None)` is called.

nulano: Sure, but `copy()` calls `load()`, whereas the comment here is specifically talking about returning an image object that has not been `load()`ed yet.

hugovk: Got it, thanks!

So to have a single check, we need to make sure the call in `_cached_hopper`'s `else` block goes through the initial `hopper()` instead of `_cached_hopper` again.

And we might as well make the call in the `if` go through `hopper`: it's only an extra function call and will be cached for subsequent calls.

Please see the newest commit.

Tests/helper.py
```diff
@@ -250,23 +251,27 @@ def tostring(im: Image.Image, string_format: str, **options: Any) -> bytes:
...
-        else:
-            im = hopper().convert(mode)
-        cache[mode] = im
-    return im.copy()
+
+    return _cached_hopper(mode).copy()
+
+
+@lru_cache(maxsize=None)
+def _cached_hopper(mode: str | None = None) -> Image.Image:
```
nulano, line 269:
```suggestion
def _cached_hopper(mode: str = None) -> Image.Image:
```

Tests/helper.py
nulano: Looks good; you can now remove `None` from the type hint for `_cached_hopper()` since the `None` part is handled in `hopper()`.

nulano (approved):

Tests/helper.py
```diff
@@ -250,23 +251,27 @@ def tostring(im: Image.Image, string_format: str, **options: Any) -> bytes:
...
-            im = hopper("L").convert(mode)
-        else:
-            im = hopper().convert(mode)
-        cache[mode] = im
-    return im.copy()
+
+    return _cached_hopper(mode).copy()
+
+
+@lru_cache(maxsize=None)
```
radarhere, line 268:
```suggestion
@lru_cache
```
I don't think we're going to have more than 128 (the default `maxsize`) modes anytime soon.

Tests/helper.py
```diff
@@ -250,23 +251,27 @@ def tostring(im: Image.Image, string_format: str, **options: Any) -> bytes:
...
+def _cached_hopper(mode: str = None) -> Image.Image:
+    if mode == "F":
+        im = hopper("L").convert(mode)
+    else:
+        im = hopper().convert(mode)
+    return im
```
radarhere, line 274:
```suggestion
        im = hopper("L")
    else:
        im = hopper()
    return im.convert(mode)
```

Tests/helper.py
```diff
@@ -250,23 +251,27 @@ def tostring(im: Image.Image, string_format: str, **options: Any) -> bytes:
...
-def hopper(mode: str | None = None, cache: dict[str, Image.Image] = {}) -> Image.Image:
+def hopper(mode: str | None = None) -> Image.Image:
+    # Use caching to reduce reading from disk, but return a copy
+    # so that the cached image isn't modified by the tests
+    # (for fast, isolated, repeatable tests).
+
     if mode is None:
         # Always return fresh not-yet-loaded version of image.
         # Operations on not-yet-loaded images is separate class of errors
         # what we should catch.
```
radarhere, line 262:
```suggestion
        # that we should catch.
```

Tests/helper.py
```diff
@@ -250,23 +251,27 @@ def tostring(im: Image.Image, string_format: str, **options: Any) -> bytes:
...
 
-def hopper(mode: str | None = None, cache: dict[str, Image.Image] = {}) -> Image.Image:
+def hopper(mode: str | None = None) -> Image.Image:
+    # Use caching to reduce reading from disk, but return a copy
+    # so that the cached image isn't modified by the tests
+    # (for fast, isolated, repeatable tests).
+
     if mode is None:
         # Always return fresh not-yet-loaded version of image.
         # Operations on not-yet-loaded images is separate class of errors
```
radarhere, line 261:
```suggestion
        # Operations on not-yet-loaded images are a separate class of errors
```

Tests/helper.py
```diff
@@ -250,23 +251,27 @@ def tostring(im: Image.Image, string_format: str, **options: Any) -> bytes:
...
-        else:
-            im = hopper().convert(mode)
-        cache[mode] = im
-    return im.copy()
+
+    return _cached_hopper(mode).copy()
+
+
+@lru_cache(maxsize=None)
+def _cached_hopper(mode: str = None) -> Image.Image:
```
radarhere, line 269:
```suggestion
def _cached_hopper(mode: str) -> Image.Image:
```

Tests/helper.py
hugovk: Adding to `conftest.py`:
```python
def pytest_sessionfinish(session, exitstatus):
    from Tests.helper import _cached_hopper
    print(_cached_hopper.cache_info())
```
Gives this after a full run:
```
CacheInfo(hits=751, misses=24, maxsize=None, currsize=24)
```

radarhere (approved):

PR #7913, opened 2024-03-28
Add type hints for ImageCms

nulano:

 * Add type hints for `ImageCms` and `ImageCms.core`;
 * Add type check to `ImageCms.core.profile_tobytes` - this cannot be called via the non-core functions thanks to a type check in `ImageCms.ImageCmsProfile()`.
 * Update types in `ImageCms.rst` to match the new ones in `_imagingcms.pyi`.


src/_imagingcms.c
```diff
@@ -1511,15 +1511,16 @@ setup_module(PyObject *m) {
...
-
     /* Ready object types */
     PyType_Ready(&CmsProfile_Type);
     PyType_Ready(&CmsTransform_Type);
 
     Py_INCREF(&CmsProfile_Type);
     PyModule_AddObject(m, "CmsProfile", (PyObject *)&CmsProfile_Type);
 
+    Py_INCREF(&CmsTransform_Type);
+    PyModule_AddObject(m, "CmsTransform", (PyObject *)&CmsTransform_Type);
```
nulano, line 1522: It seemed odd to me to expose one of the classes but not the other.

src/_imagingcms.c
```diff
@@ -1511,15 +1511,16 @@ setup_module(PyObject *m) {
     PyObject *v;
     int vn;
 
-    CmsProfile_Type.tp_new = PyType_GenericNew;
```
nulano, line 1514: This is not needed (and incorrect! should call `core.profile_open` or `core.profile_frombytes` instead).

I added new tests for this being unset, but on PyPy this is the default value.

src/PIL/ImageCms.py
```diff
@@ -787,14 +797,17 @@ def createProfile(colorSpace, colorTemp=-1):
         except (TypeError, ValueError) as e:
             msg = f'Color temperature must be numeric, "{colorTemp}" not valid'
             raise PyCMSError(msg) from e
+    else:
+        # colorTemp is unused if colorSpace != "LAB"
+        colorTemp = 0.0
```
radarhere, line 802: This is minor, but considering it's unused, what's the advantage of setting it to zero?

src/PIL/ImageCms.py
```diff
@@ -576,15 +582,15 @@ def buildTransform(
...
-    flags=Flags.SOFTPROOFING,
-):
+    inputProfile: _CmsProfileCompatible,
+    outputProfile: _CmsProfileCompatible,
+    proofProfile: _CmsProfileCompatible,
+    inMode: str,
+    outMode: str,
+    renderingIntent: Intent = Intent.PERCEPTUAL,
+    proofRenderingIntent: Intent = Intent.ABSOLUTE_COLORIMETRIC,
+    flags: Flags | int = Flags.SOFTPROOFING,
```
radarhere, line 592: Wouldn't our discussion in https://github.com/python-pillow/Pillow/pull/7671#discussion_r1438996324 about favouring enums over `int` also apply to the `IntFlag` of `Flags`, meaning this could just be `flags: Flags`?

src/PIL/ImageCms.py
nulano: The value is passed to `PyArg_ParseTuple` with the `d` format. Based on [the documentation](https://docs.python.org/3/c-api/arg.html#numbers), my understanding was that this required a `float` object, not a `SupportsFloat` object. However, testing with Python 3.11, it seems that a `SupportsFloat` object is actually accepted, so I'll revert this part of the change.

(Testing with `SupportsInt` I do get the expected `TypeError`, so at least that part matches the documentation.)

src/PIL/ImageCms.py
nulano: Yep, you're right. IIRC I might have added `int` to support the deprecated `ImageCms.FLAGS`, but that does not have type hints and should not be used anyway.

hugovk (approved):

hugovk:
Thanks!

PR #7916, opened 2024-03-29
Updated xz to 5.6.2

radarhere:
https://github.com/tukaani-project/xz/releases/tag/v5.4.6


nulano:
I'm assuming you are aware of [CVE-2024-3094](https://github.com/advisories/GHSA-rxwq-x6h5-x525) and didn't upgrade to 5.6.1 for that reason. Perhaps it is an overreaction at this point, but I'm wondering if we should consider completely removing liblzma from the wheels for the upcoming release. liblzma is only an indirect optional dependency via libtiff and users could just install from source if they need it.

wiredfool:
Yes, we're aware. I'd like to hold off on merging this until more clarity comes out of this from the experts. Xz is deeply in use in debian (e.g. dpkg) so they're digging pretty heavily now (https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1068024).

radarhere:
xz 5.6.2 has now been released - https://github.com/tukaani-project/xz/releases/tag/v5.6.2

https://tukaani.org/xz-backdoor/
> There was discussion if the 5.6.x version numbers shouldn’t be used further and the next release could be 5.8.0. However, it felt clearer to make XZ Utils 5.6.2 as a cleanup step. There are a few small improvements planned which could become 5.8.0 in shorter time than usual, hopefully replacing 5.6.x somewhat soon.

nulano:
Should we also switch to using the GitHub release instead of SourceForge?
https://github.com/python-pillow/Pillow/blob/114e01701ac92d6bd1df7df771a2be8abcc0ae33/winbuild/build_prepare.py#L179

FWIW, [multibuild uses `https://tukaani.org/xz/xz-5.6.2.tar.gz`](https://github.com/multi-build/multibuild/blob/452dd2d1705f6b2375369a6570c415beb3163f70/library_builders.sh#L269) which is a redirect to the GitHub release:
```
C:\Users\Nulano>curl --head https://tukaani.org/xz/xz-5.6.2.tar.gz
HTTP/1.1 302 Found
Server: nginx
Date: Wed, 12 Jun 2024 18:51:24 GMT
Content-Type: text/html; charset=iso-8859-1
Content-Length: 262
Connection: keep-alive
Location: https://github.com/tukaani-project/xz/releases/download/v5.6.2/xz-5.6.2.tar.gz
X-Proxy-Cache: BYPASS
X-Powered-By: Zoner
```

radarhere:
Ok, GitHub does look to be the primary repository. I've pushed a commit.

hugovk:
My thinking on this: I don't see anything particularly urgent/important in the changelog to merge this right now, so maybe we hold off for another release or so?

radarhere:
Another release or so of ours, or of xz? I'm guessing you mean xz, as a way of increasing confidence that they have reviewed their code thoroughly.

hugovk:
Of ours, but that could also include of xz :)

Yeah, I don't think we're in a rush to upgrade, and it gives some more time for the dust to settle.

hugovk:
More time has passed and things have settled. Any objections to merging this for our release next month?

radarhere:
No objections.

PR #7918, opened 2024-03-30
Added RGB to I;16, I;16L and I;16B conversion

radarhere:
Instead of Tests/helper.py converting from RGB to I to I;16*,
this adds support for converting directly from RGB to I;16, I;16L and I;16B.


Yay295:
There's "I;16N" too, though I guess it's not yet being used in any tests.

PR #7921, opened 2024-03-30
Add test for creating an image from color values

Yay295:
Also changed the image mode info to use a `NamedTuple`.

The test currently fails for `BGR;15` and `BGR;16`.

```python
__________ TestImage.test_pixels_from_new_image_by_color[mode16] __________

self = <Tests.test_image.TestImage object at 0x7fa65defb1c0>, mode = ImageModeInfo(name='BGR;15', num_bands=3, pixel_size=2)

    @pytest.mark.parametrize("mode", image_modes)
    def test_pixels_from_new_image_by_color(self, mode: ImageModeInfo) -> None:
        color = 123 if mode.num_bands == 1 else tuple(range(mode.num_bands))
        im = Image.new(mode.name, (2, 2), color=color)

>       assert im.getpixel((0, 0)) == color
E       assert (0, 0, 0) == (0, 1, 2)
E         At index 1 diff: 0 != 1
E         Use -v to get more diff

Tests/test_image.py:670: AssertionError
__________ TestImage.test_pixels_from_new_image_by_color[mode17] __________

self = <Tests.test_image.TestImage object at 0x7fa65def9a80>, mode = ImageModeInfo(name='BGR;16', num_bands=3, pixel_size=2)

    @pytest.mark.parametrize("mode", image_modes)
    def test_pixels_from_new_image_by_color(self, mode: ImageModeInfo) -> None:
        color = 123 if mode.num_bands == 1 else tuple(range(mode.num_bands))
        im = Image.new(mode.name, (2, 2), color=color)

>       assert im.getpixel((0, 0)) == color
E       assert (0, 0, 0) == (0, 1, 2)
E         At index 1 diff: 0 != 1
E         Use -v to get more diff

Tests/test_image.py:670: AssertionError
```


Yay295:
Looks like the issue with `BGR;15` and `BGR;16` was previously mentioned in #7303.

Tests/test_image.py
```diff
@@ -688,6 +694,24 @@ def _make_new(
...
+            color = 123
+        else:
+            color = tuple(range(mode.num_bands))
+
+        im = Image.new(mode.name, (2, 2), color=color)
+
+        assert im.getpixel((0, 0)) == color
+        assert im.getpixel((0, 1)) == color
+        assert im.getpixel((1, 0)) == color
+        assert im.getpixel((1, 1)) == color
```
radarhere, line 713:
```suggestion
        im = Image.new(mode.name, (1, 1), color)
        assert im.getpixel((0, 0)) == color
```
I don't think more than one pixel needs to be tested.

radarhere:
Why you feel `test_pixels_from_new_image_by_color` is necessary?

I would think we are already testing this at https://github.com/python-pillow/Pillow/blob/d31148bb86ad33cfb5bd9952775d5aaede028c44/Tests/test_image_access.py#L186-L192 with [almost all of the modes](https://github.com/python-pillow/Pillow/blob/d31148bb86ad33cfb5bd9952775d5aaede028c44/Tests/test_image_access.py#L212-L228) you're testing here.

Yay295:
Because I didn't know that test existed. It looks like it will work with all of the modes though, so I'll update that instead.

radarhere:
I recognise this is relatively minor, but I would rather not add pixel size to Tests/helper.py just for use in a single test.

Yay295:
It can be used for more tests in the future. I'm trying to keep my pull requests small so they actually get merged in a reasonable time frame.

radarhere:
> more tests in the future

Such as? Apart from the scenario where a user manually constructs data for `frombytes()`, it's not immediately obvious when a user needs to know pixel size.

I would rather think of the test suite as testing the inputs and outputs of Pillow, and not containing data specific to our internals that most users don't need to know.

Yay295:
I don't see an alternative. We need to know the pixel size of each mode for this test, so the pixel sizes need to be included in the list of mode names. There could be a separate list just for this test, but that defeats the point of having a common list of modes, and there would need to be another check that the two lists always have the same modes.

hugovk:
> I recognise this is relatively minor, but I would rather not add pixel size to Tests/helper.py just for use in a single test.

> It can be used for more tests in the future. 


I agree with @radarhere, let's not move it now just in case, we can move it later when we need to move it, which might never happen (re: https://wiki.c2.com/?YouArentGonnaNeedIt).


> I'm trying to keep my pull requests small so they actually get merged in a reasonable time frame.

Yes, this is the way, thank you :)

Yay295:
I thought of another way to do it. It's not efficient, but I suppose at least it's only being done in this one test.

Yay295:
So it looks like when getting BGR 15/16 data the values are scaled to be 0-255,

https://github.com/python-pillow/Pillow/blob/33a73b5266c6589e21cc32a381fe907e23161877/src/libImaging/Access.c#L90-L108

but when setting that data the values aren't scaled

https://github.com/python-pillow/Pillow/blob/33a73b5266c6589e21cc32a381fe907e23161877/src/_imaging.c#L607-L623

Though that doesn't explain why my original test passed.

Yay295:
No wait, that can't be it, because this is only failing on big-endian.

Yay295:
Does anyone know what the endianness of BGR;15 and BGR;16 is supposed to be?

Yay295:
It seems like BGR;15 and BGR;16 are always treated as little-endian, which would explain why they're not working on a big-endian machine, but I haven't yet found a place where they're being treated as native-endian rather than explicitly little-endian.

Yay295:
I added some logging in another branch and found some things.

First, the original RGB Hopper pixel data for the failing pixel is `(117, 104, 121)`.

On a little-endian machine I found this:
After BGR;15 Conversion: `0b0011100110101111`
After BGR;15 Conversion: `(123, 106, 115)`
After getdata putdata: `0b0011100110101111`
After  getdata putdata: `(123, 106, 115)`

The pixel values were changed a bit due to the RGB to BGR;15 conversions, but they're close enough.

On a big-endian machine I found this:
After BGR;15 Conversion: `0b1010111100111001`
After BGR;15 Conversion: `(205, 205, 90)`
After getdata putdata: `0b0010111100111001`
After getdata putdata: `(205, 205, 90)`

The pixel values are correct for that byte data, but the bytes are reversed. Also, the first bit (which is nothing in BGR;15) is set at the start, but not the end.

Yay295:
All tests should be passing now. Some of the code was reading/writing RGB instead of BGR, and `memcpy` copies data in native-endianness when we need it to be in little-endian, so that can't be used for these modes.

Yay295:
Should I rebase this, or can it be merged?

radarhere:
I don't see any need to rebase - there's no merge conflict.

If #7978 is merged, I'm not a fan of the BGR;* changes here, as I think introducing changes to BGR;* modes at the same time as deprecating them is [not helpful](https://github.com/python-pillow/Pillow/issues/7970#issuecomment-2056406966).

Yay295:
So, what, I have to wait until you finally remove the BGR modes before you'll accept this fix for I;16N?

radarhere:
I wasn't suggesting that this wait until the deprecation period ended, only that this wait until the question of deprecation was resolved with #7978 being accepted or rejected.

If you would like this to move ahead faster than that, then my suggestion would to be mark the BGR;* cases as failures for now - see https://github.com/Yay295/Pillow/pull/17. If #7978 is rejected, then the BGR;* changes here can be revisited.

radarhere (approved):

radarhere:
I found that BGR;16 doesn't actually fail, and have created https://github.com/Yay295/Pillow/pull/18

radarhere (approved):

Tests/test_image_access.py
```diff
@@ -168,16 +168,15 @@ def check(self, mode: str, expected_color_int: int | None = None) -> None:
             f"expected {expected_color} got {actual_color}"
         )
 
-        # Check 0
+        # check 0x0 image with None initial color
```
hugovk, line 171: Let's keep these beginning with a capital letter:

```suggestion
        # Check 0x0 image with None initial color
```

If anything, that's consistent with PEP 8: https://peps.python.org/pep-0008/#comments

> Comments should be complete sentences. The first word should be capitalized, unless it is an identifier that begins with a lower case letter (never alter the case of identifiers!).

Tests/test_image_access.py
```diff
@@ -168,16 +168,15 @@ def check(self, mode: str, expected_color_int: int | None = None) -> None:
...
         im = Image.new(mode, (0, 0), None)
         assert im.load() is not None
-
         error = ValueError if self._need_cffi_access else IndexError
         with pytest.raises(error):
             im.putpixel((0, 0), expected_color)
         with pytest.raises(error):
             im.getpixel((0, 0))
-        # Check 0 negative index
+        # check negative index
```
hugovk, line 179:
```suggestion
        # Check negative index
```

Tests/test_image_access.py
```diff
@@ -198,36 +197,15 @@ def check(self, mode: str, expected_color_int: int | None = None) -> None:
             f"expected {expected_color} got {actual_color}"
         )
 
-        # Check 0
+        # check 0x0 image with initial color
```
hugovk, line 200:
```suggestion
        # Check 0x0 image with initial color
```

Tests/test_image_access.py
```diff
@@ -198,36 +197,15 @@ def check(self, mode: str, expected_color_int: int | None = None) -> None:
...
         im = Image.new(mode, (0, 0), expected_color)
         with pytest.raises(error):
             im.getpixel((0, 0))
-        # Check 0 negative index
+        # check negative index
```
hugovk, line 204:
```suggestion
        # Check negative index
```

hugovk (approved):

Issue #7922, opened 2024-03-31
Reading Quite OK Image (QOI) files is slow compared to PNG and qoi


ivanstepanovftw:
### What did you do?
```python
image = Image.open(image_filepath)
```
while using PyTorch data loader with 8 workers.

### What did you expect to happen?
Fast

### What actually happened?
Slow

### What are your OS, Python and Pillow versions?

* OS: Linux
* Python: 3.10.13
* Pillow: 10.2.0

Reading PNG images with Pillow is faster.

For me, current workaround is to use https://github.com/kodonnell/qoi from PyPI (`qoi` package). It is faster when I switched to it. To compare:
```python
image = qoi.read(image_filepath)
```

Image sizes are regular, like 800x480, 1280x600.

radarhere:
Hi. I think it may not be the most useful metric to say 'QOI reading should at least be as fast as PNG reading', since if we manage to make QOI reading faster, then someone could say 'PNG reading should be at least as fast as QOI reading', and it is theoretically a neverending need to be faster.

Rather than comparing it to PNG reading, could you give us an idea for how much faster you would like QOI reading to be than it is now? If we were to make it twice as fast, I would consider that to be a massive win, but it may not be as much as you are hoping for.

Also, if you could attach a specific QOI image that we can use when testing speed, that would allow us to better cater to your situation.

ivanstepanovftw:
> it is theoretically a neverending need to be faster.

Performance of QOI encoding and decoding is O(n), where n is total pixels. I have give you example above using `qoi` Python wrapper for original qoi library, which is faster than Pillow decoding in my tests. Is something else performed while reading and decoding the compressed QOI?

Here is the reproducer for you:
```python
import timeit

import matplotlib.pyplot as plt
import numpy as np
import pandas as pd
import qoi
import seaborn as sns
import torch
from PIL import Image

# Convert PNG/JPG image to numpy array
img_path = "/home/i/Downloads/PNG_transparency_demonstration_1.png"  # original image
qoi_path = "/home/i/Downloads/PNG_transparency_demonstration_1.qoi"  # saved qoi file
img = Image.open(img_path)
rgb_array = np.array(img)
_ = qoi.write(qoi_path, rgb_array)

# Define functions to time reading QOI file using Pillow and qoi
def read_png_with_pillow_into_tensor():
    img = Image.open(img_path)
    return torch.tensor(np.asarray(img))

def read_qoi_with_pillow_into_tensor():
[...]

PR #7925, opened 2024-04-01
Improve speed of loading QOI images

radarhere:
Helps #7922 by reducing the loading time of [PNG_transparency_demonstration_1.qoi.zip](https://github.com/python-pillow/Pillow/files/14819369/PNG_transparency_demonstration_1.qoi.zip) from 5.15s to 3.89s.


hugovk (approved):
Here's the charts from running the benchmark from the issue:
# Before

About 3.5s for Pillow to read QOI

# After

About 2.5s for Pillow to read QOI

PR #7927, opened 2024-04-01
Deprecate eval(), replacing it with lambda_eval() and unsafe_eval()

hugovk:


src/PIL/ImageMath.py
```diff
@@ -235,9 +236,55 @@ def imagemath_convert(self: _Operand, mode: str) -> _Operand:
...
+    Returns the result of an image function.
+
+    :py:mod:`~PIL.ImageMath` only supports single-layer images. To process multi-band
+    images, use the :py:meth:`~PIL.Image.Image.split` method or
+    :py:func:`~PIL.Image.merge` function.
+
+    :param expression: A function that receives a dictionary.
+    :param options: Values to add to the function's dictionary. You
+                    can either use a dictionary, or one or more keyword
+                    arguments.
```
hugovk, line 254: We have `_dict` in the signature and `options` here.

Should rename `_dict` to `options` in the signature, or rename `options` to `_dict` in the docstring, or remove `options` from the docstring?

(And same in the others)

src/PIL/ImageMath.py
```diff
@@ -235,9 +236,55 @@ def imagemath_convert(self: _Operand, mode: str) -> _Operand:
 }
 
 
-def eval(expression: str, _dict: dict[str, Any] = {}, **kw: Any) -> Any:
+def lambda_eval(
+    expression: Callable[[dict[str, Any]], Any],
+    _dict: dict[str, Any] = {},
+    **kw: Any,
+) -> Any:
```
hugovk, line 243:
```suggestion
def lambda_eval(
    expression: Callable[[dict[str, Any]], Any], _dict: dict[str, Any] = {}, **kw: Any
) -> Any:
```

hugovk: (Will leave these expanded, they're clearer that way given all the type hints.)

radarhere (approved):

PR #7928, opened 2024-04-01
CVE-2024-28219: Use strncpy to avoid buffer overflow

hugovk:
> A straightforward way to stop `strcpy` from overflowing is to use `strncpy`.
> 
> The test image is from https://github.com/saucecontrol/Compact-ICC-Profiles/blob/master/profiles/sGrey-v2-nano.icc, [licensed under CC0](https://github.com/saucecontrol/Compact-ICC-Profiles/blob/master/license).


radarhere (approved):

hugovk:
Merging; we don't need to wait for valgrind or AppVeyor to finish testing the latest commit which was docs only.

hugovk:
I'll update CHANGES.rst with this when bumping the version.

hugovk:
Or you already did, no problem :)

src/_imagingcms.c
```diff
@@ -201,8 +201,8 @@ cms_transform_new(cmsHTRANSFORM transform, char *mode_in, char *mode_out) {
 
     self->transform = transform;
 
-    strcpy(self->mode_in, mode_in);
-    strcpy(self->mode_out, mode_out);
+    strncpy(self->mode_in, mode_in, 8);
+    strncpy(self->mode_out, mode_out, 8);
```
charmander, line 205: `strncpy` doesn’t null-terminate oversized entries.

```suggestion
    strncpy(self->mode_in, mode_in, 8);
    self->mode_in[7] = '\0';
    strncpy(self->mode_out, mode_out, 8);
    self->mode_out[7] = '\0';
```

But are these even used? The Python `ImageCmsTransform` wrapper has `inputMode` and `outputMode` attributes already, and they work normally. Is there a need to preserve a slightly different `.transform.inputMode`?

nulano: Probably not, see #7931 for a suggestion to remove them.

ThiefMaster:
Is this exploitable when all you do is opening user-provided images and converting them? Or only when you explicitly used the `ImageCms` module?

radarhere:
Something outside of Pillow would have to use ImageCms, yes.

ArielPrevu3D:
It looks like `Image.convert` can lead to a buffer overflow if the `mode` parameter is user-supplied. Your code does not have to use `ImageCms` directly to be exploitable.

https://github.com/python-pillow/Pillow/blob/955c5dac03bd4b59281d00ffad345aa28e79310f/src/PIL/Image.py#L1079

radarhere:
If you look a few lines earlier
https://github.com/python-pillow/Pillow/blob/955c5dac03bd4b59281d00ffad345aa28e79310f/src/PIL/Image.py#L1070-L1079
you will see that one of the modes has to be "LAB", and the other has to be one of "RGB", "RGBA" or "RGBX",
none of which is more than 8 characters.

So it isn't exploitable from `Image.convert`.

PR #7931, opened 2024-04-01
Remove unused CMS properties and fix documentation

Yay295:
The `ImageCmsTransform` mode properties are set here:

https://github.com/python-pillow/Pillow/blob/6b676f2962d34bd54eb87ae172c3a7a403c03345/src/PIL/ImageCms.py#L317-L319

Some documentation further down in that file used the wrong property names.

These values are also passed to the C code where they are used and then stored, but the stored values are not documented or used anywhere, so I've removed them.


nulano:
> the stored values are not documented or used anywhere

FWIW, there are type hints added in #7913: https://github.com/python-pillow/Pillow/blob/6b676f2962d34bd54eb87ae172c3a7a403c03345/src/PIL/_imagingcms.pyi#L111-L114

Yay295:
Ah, I didn't think to check there. And that did go in with 10.3.0 a few hours ago.

radarhere (approved):

hugovk:
Thank you!

PR #7933, opened 2024-04-02
Fix ImageMath documentation parameter names

jbjd:
In the docs for ImageMath, the parameter "environment" was changed to "options". This was updated in the function descriptions but not in their parameters.


radarhere (approved):

hugovk:
Thank you!

---

As an aside, AppVeyor didn't skip this docs-only PR (re: https://github.com/python-pillow/Pillow/pull/7909).

PR #7936, opened 2024-04-02
fromarray: add type hints

adamjstewart:
Really excited to see type hints in the latest release!

The only feature I'm using that was missing type hints is `fromarray`, so I added them. Not 100% sure if `numpy.typing.ArrayLike` is the right way to go, but feel free to suggest alternatives.


Yay295:
I think the actual type would be `numpy.ndarray`, but that would require importing numpy.

adamjstewart:
It doesn't have to be a `numpy.ndarray`, it could also be a `PIL.Image` or any other structure that implements an array interface. There are many ways to avoid importing numpy for typing: https://mypy.readthedocs.io/en/stable/runtime_troubles.html

Yay295:
The other half of the NumPy support, which also isn't typed yet, is here:

https://github.com/python-pillow/Pillow/blob/aeeb596c98fe4c9cd79cc1408eb7a3353c43eef1/src/PIL/Image.py#L686-L709

Yay295:
I think [`numpy.typing.ArrayLike`](https://numpy.org/devdocs/reference/typing.html#numpy.typing.ArrayLike) is too broad, since it includes scalars and sequences, whereas `Image.fromarray()` only supports the NumPy array interface.

nulano:
I think you may need to add numpy to the list of dependencies here https://github.com/python-pillow/Pillow/blob/aeeb596c98fe4c9cd79cc1408eb7a3353c43eef1/pyproject.toml#L69-L71 or in a new block.

Yay295:
Something like
```python
from typing import Protocol, Sequence, TypedDict
class ArrayInterfaceDict(TypedDict):
    shape: Sequence[int]
    typestr: str
class ArrayInterface(Protocol):
    __array_interface__: ArrayInterfaceDict
```

nulano:
> Something like
> 
> ```python
> from typing import Protocol, Sequence, TypedDict
> class ArrayInterfaceDict(TypedDict):
>     shape: Sequence[int]
>     typestr: str
> class ArrayInterface(Protocol):
>     __array_interface__: ArrayInterfaceDict
> ```
> 
> would be a more exact definition of what Pillow requires.

Adding that to `PIL._typing` would probably be a better option than requiring numpy be installed.

radarhere:
`__array_interface__` can [optionally have "strides"](https://github.com/python-pillow/Pillow/blob/e8ab5640774716c5486d3cb05167f74f742ad6ef/src/PIL/Image.py#L3122), but `TypedDict` [only gained optional items in Python 3.11](https://peps.python.org/pep-655), so it doesn't look like we can use a `TypedDict` here.

Otherwise, I've created https://github.com/adamjstewart/Pillow/pull/1 as a suggestion to add the protocol.

adamjstewart:
Thanks for the help @radarhere!

P.S. Love the profile pic @Yay295, excited for season 3 coming out this weekend!

radarhere (approved):

src/PIL/Image.py
```diff
@@ -3069,7 +3069,15 @@ def frombuffer(mode, size, data, decoder_name="raw", *args):
     return frombytes(mode, size, data, decoder_name, args)
 
 
-def fromarray(obj, mode=None):
+class SupportsArrayInterface(Protocol):
```
nulano, line 3072: Since (I assume) this is not intended to be exported, I would expect we either want this to be private, i.e.
```suggestion
class _SupportsArrayInterface(Protocol):
```
or in a private module, i.e. `PIL._typing.SupportsArrayInterface`.

@radarhere what do you think?

radarhere: I was modelling this after our implementation of [`SupportsGetMesh`](https://pillow.readthedocs.io/en/stable/reference/ImageOps.html#PIL.ImageOps.SupportsGetMesh).

https://github.com/python-pillow/Pillow/pull/7812#issuecomment-1952150296
> If we're using _SupportsGetMesh as a type hint, this probably is something users might want to use, so should be accessible as a non-underscore thing?

nulano: I feel like the difference is that `SupportsGetMesh` is a Pillow protocol, whereas `SupportsArrayInterface` seems to be a numpy protocol. Similarly, we have [`PIL._typing.SupportsRead`](https://pillow.readthedocs.io/en/stable/reference/internal_modules.html#PIL._typing.SupportsRead) for a builtin Python protocol. I would therefore expect users to either use the actual numpy type (e.g. `numpy.ndarray`) or define their own protocol.

@hugovk Do you have any thoughts as the author of the quoted comment?

nulano:
> Not 100% sure if numpy.typing.ArrayLike is the right way to go

NumPy seems to [overload the `np.array` function](https://github.com/numpy/numpy/blob/e59c074842e3f73483afa5ddef031e856b9fd313/numpy/_core/multiarray.pyi#L179-L233) with several types including `Any` and I [can't find a more specific protocol](https://github.com/search?q=repo%3Anumpy%2Fnumpy+__array_interface__&type=code).

src/PIL/Image.py
```diff
@@ -3069,7 +3069,15 @@ def frombuffer(mode, size, data, decoder_name="raw", *args):
...
+    """
+    An object that has an ``__array_interface__`` dictionary.
+    """
+
+    __array_interface__: dict[str, Any]
```
nulano, line 3077: We should probably use a `@property` like [numpy](https://github.com/numpy/numpy/blob/e59c074842e3f73483afa5ddef031e856b9fd313/numpy/__init__.pyi#L991):

```suggestion
    @property
    def __array_interface__(self) -> dict[str, Any]:
        raise NotImplementedError()
```

As it is written now, it requires read-write access, whearas we only need read access.

See https://mypy-play.net/?mypy=latest&python=3.12&gist=b20b51c7cf6c3360a24d4cf77fb1ca64:

> main.py:22: error: Incompatible types in assignment (expression has type "TestProperty", variable has type "SupportsArrayInterfaceAttribute")  [assignment]
> main.py:22: note: Protocol member SupportsArrayInterfaceAttribute.__array_interface__ expected settable variable, got read-only attribute

adamjstewart:
Anything else needed for this PR?

src/PIL/Image.py
hugovk: Hmm, let's go for `SupportsArrayInterface`, it's probably more useful that way, and it's only a small class. We an always deprecate/remove later if needed.

hugovk:
Thanks!

PR #7937, opened 2024-04-03
Document that QoiImagePlugin uses Python for decoding

### LINKED ISSUE #7922: Reading Quite OK Image (QOI) files is slow compared to PNG and qoi

radarhere:
Resolves #7922

https://github.com/python-pillow/Pillow/issues/7922#issuecomment-2033336635
> That said, Pillow's processing of QOI files in Python, not C, could probably be clarified here: https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html#qoi.

### DIFF
--- a/docs/handbook/image-file-formats.rst
+++ b/docs/handbook/image-file-formats.rst
@@ -1488,7 +1488,9 @@ QOI
 
 .. versionadded:: 9.5.0
 
-Pillow reads images in Quite OK Image format, using a Python decoder.
+Pillow reads images in Quite OK Image format, using a Python decoder. If you wish to
+write code specifically for this format, https://pypi.org/project/qoi is an alternative
+library that uses C to decode the image, and interfaces with NumPy.
 
 SUN
 ^^^
--- a/docs/handbook/image-file-formats.rst
+++ b/docs/handbook/image-file-formats.rst
@@ -1489,7 +1489,7 @@ QOI
 .. versionadded:: 9.5.0
 
 Pillow reads images in Quite OK Image format, using a Python decoder. If you wish to
-write code specifically for this format, https://pypi.org/project/qoi is an alternative
+write code specifically for this format, :pypi:`qoi` is an alternative
 library that uses C to decode the image, and interfaces with NumPy.
 
 SUN
--- a/docs/handbook/image-file-formats.rst
+++ b/docs/handbook/image-file-formats.rst
@@ -1488,7 +1488,7 @@ QOI
 
 .. versionadded:: 9.5.0
 
-Pillow identifies and reads images in Quite OK Image format.
+Pillow reads images in Quite OK Image format, using a Python decoder.
 
 SUN
 ^^^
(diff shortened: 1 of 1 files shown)


hugovk:
Shall we also mention https://github.com/kodonnell/qoi or https://pypi.org/project/qoi/? It's not a Pillow plugin but can be a useful alternative.

docs/handbook/image-file-formats.rst
```diff
@@ -1488,7 +1488,9 @@ QOI
 
 .. versionadded:: 9.5.0
 
-Pillow identifies and reads images in Quite OK Image format.
+Pillow reads images in Quite OK Image format, using a Python decoder. If you wish to
+write code specifically for this format, https://pypi.org/project/qoi is an alternative
+library that interfaces with NumPy.
```
hugovk, line 1493: Mention C somehow, so they know it's faster?

```suggestion
library written in C that interfaces with NumPy.
```

radarhere: "written in C" sounds to me like the entire library is in C. How about "that uses C to decode the image"?

hugovk: Yeah, something like that's fine too.

radarhere: Cool, I've pushed a commit for that.

docs/handbook/image-file-formats.rst
```diff
@@ -1488,7 +1488,9 @@ QOI
 
 .. versionadded:: 9.5.0
 
-Pillow identifies and reads images in Quite OK Image format.
+Pillow reads images in Quite OK Image format, using a Python decoder. If you wish to
+write code specifically for this format, https://pypi.org/project/qoi is an alternative
```
hugovk, line 1492: Just remembered, we have an extlink role for this:

```suggestion
write code specifically for this format, :pypi:`qoi` is an alternative
```

hugovk (approved):

docs/handbook/image-file-formats.rst
```diff
@@ -1488,7 +1488,9 @@ QOI
 
 .. versionadded:: 9.5.0
 
-Pillow identifies and reads images in Quite OK Image format.
+Pillow reads images in Quite OK Image format, using a Python decoder. If you wish to
+write code specifically for this format, :pypi:`qoi` is an alternative
+library that uses C to decode the image, and interfaces with NumPy.
```
nulano, line 1493: Feel free to ignore, but this reads better to me.
```suggestion
Pillow reads images in Quite OK Image format using a Python decoder. If you wish to
write code specifically for this format, :pypi:`qoi` is an alternative
library that uses C to decode the image and interfaces with NumPy.
```

nulano (approved):

PR #7943, opened 2024-04-04
Image.open returns an ImageFile instance

### LINKED ISSUE #7941: Image.filename is not an exposed property
The Pillow reference says that [Image.filename](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.filename) is an attribute, but VSCode doesn't auto-complete it.  This code works:

Pillow filename screenshot

but VSCode shows `image.filename` as white whereas the other attributes are blue.  Is it possible to make `Image.filename` a declared Python attribute so that developers have confidence that their code will work?

I'm using Pillow 10.3.0.

radarhere:
Resolves #7941

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -3222,7 +3222,7 @@ def _decompression_bomb_check(size: tuple[int, int]) -> None:
         )
 
 
-def open(fp, mode="r", formats=None) -> Image:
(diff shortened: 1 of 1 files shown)


radarhere:
Closing in favour of #7944

PR #7944, opened 2024-04-04
Add type hints for `Image.open`, `Image.init`, and `Image.Image.save`

### LINKED ISSUE #7941: Image.filename is not an exposed property

nulano:
Fixes #7941 

While the `Image.open` function is currently marked as returning an `Image` object, it does so by calling a factory function from `OPEN`, which is marked as returning an `ImageFile`: https://github.com/python-pillow/Pillow/blob/e8ab5640774716c5486d3cb05167f74f742ad6ef/src/PIL/Image.py#L222-L228

The `accept` function usually returns a `bool`, but I found one case (webp) where it can return a `str` instead to be used as a warning.

### DIFF
--- a/src/PIL/BlpImagePlugin.py
+++ b/src/PIL/BlpImagePlugin.py
@@ -241,7 +241,7 @@ class BLPFormatError(NotImplementedError):
     pass
 
 
-def _accept(prefix):
+def _accept(prefix: bytes) -> bool:
     return prefix[:4] in (b"BLP1", b"BLP2")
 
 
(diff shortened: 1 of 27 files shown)


aclark4life:
Hmmm, jinx #7943 ?

hugovk (approved):

PR #7950, opened 2024-04-06
Rename test_roundtrip() to test_mode()

Yay295:
This test isn't actually roundtripping anything.


Tests/test_image_getdata.py
```diff
@@ -14,7 +14,7 @@ def test_sanity() -> None:
     assert data[0] == (20, 20, 70)
 
 
-def test_roundtrip() -> None:
+def test_hopper() -> None:
```
radarhere, line 17:
```suggestion
def test_mode() -> None:
```
How about this instead?

PR #7956, opened 2024-04-08
Add support for reading `BITMAPV2INFOHEADER` and `BITMAPV3INFOHEADER`

Cirras:

 * Add support for bitmaps with header size 52 (`BITMAPV2INFOHEADER`)
 * Add support for bitmaps with header size 56 (`BITMAPV3INFOHEADER`)
 * Improve adjacent comments around bitmap info headers

Notes:
* These are poorly-documented and obscure formats, but they exist in the wild and are included in [BMP Suite](https://entropymine.com/jason/bmpsuite/)
* Both formats are recognized by Adobe Photoshop and popular web browsers (Chromium, Firefox)
* `BITMAPV3INFOHEADER` is recognized by Windows


src/PIL/BmpImagePlugin.py
```diff
@@ -53,7 +53,7 @@ def _accept(prefix: bytes) -> bool:
 
 
 def _dib_accept(prefix):
-    return i32(prefix) in [12, 40, 64, 108, 124]
+    return i32(prefix) in [12, 40, 52, 56, 64, 108, 124]
```
radarhere, line 56:
```suggestion
    return i32(prefix) in [12, 40, 64, 108, 124]
```
The test images you've added here work fine without this change to `_dib_accept`.

radarhere (approved):

src/PIL/BmpImagePlugin.py
Cirras: `_dib_accept` just checks if the prefix (header size) indicates one of the supported bitmap info headers.
This is sidestepped in the tests in `test_bmp_reference.py` because the test files from bmp-suite are `.bmp` and not `.dib`.

The core bitmap-reading logic is otherwises shared, with the only difference being that DIBs lack a 14-byte [BITMAPFILEHEADER](https://learn.microsoft.com/en-us/windows/win32/api/wingdi/ns-wingdi-bitmapfileheader) starting with `BM`.

Reverting this change would revert support for these headers when reading DIBs.

radarhere: The general rule is that if there are changes, there should be tests so that we can ensure the changes aren't accidentally broken down the line.

So let's test this then - https://github.com/Cirras/Pillow/pull/3

Cirras: Thanks! I merged your tests.

radarhere (approved):

PR #7957, opened 2024-04-08
Update `make release-test`: no more eggs

hugovk:
During the release, the output of `make release-test` includes:

```
Successfully installed pillow-10.4.0.dev0
rm dist/*.egg
rm: dist/*.egg: No such file or directory
make: [release-test] Error 1 (ignored)
rmdir dist
rmdir: dist: No such file or directory
make: [release-test] Error 1 (ignored)
python3 -m pytest -qq
```

Eggs are no longer used: https://peps.python.org/pep-0715/

Let's remove `-rm dist/*.egg`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,7 +78,6 @@ release-test:
 	python3 selftest.py
 	python3 -m pytest Tests
 	python3 -m pip install .
-	-rm dist/*.egg
 	-rmdir dist
 	python3 -m pytest -qq
 	python3 -m check_manifest


radarhere:
The PEP you've linked to talks about eggs in relation to PyPI, but why are they no longer being created by Pillow?

The line was originally added in #4992, to resolve #4981. So what changed between then and now?

radarhere (approved):

radarhere:
You're right, `python3 setup.py install` creates a dist egg, but `python3 -m pip install .` does not.

By the same token then, we also don't need to remove the dist directory any longer, so I've created #7974

PR #7965, opened 2024-04-11
Fix TGA rawmode BGRA;15

Yay295:
Looks like a typo in the mode name. The only other place "BGR;5" is used is in `Unpack.c` where it's an alias for "BGR;15".

The documentation I found says that it should actually be "BGRA;16" though, which isn't a mode we support.


Yay295:
Actually, I missed it; "BGRA;15" *is* supported.

src/PIL/TgaImagePlugin.py
```diff
@@ -118,7 +118,7 @@ def _open(self) -> None:
             start, size, mapdepth = i16(s, 3), i16(s, 5), s[7]
             if mapdepth == 16:
                 self.palette = ImagePalette.raw(
-                    "BGR;15", b"\0" * 2 * start + self.fp.read(2 * size)
+                    "BGRA;15", b"\0" * 2 * start + self.fp.read(2 * size)
                 )
```
radarhere, line 122:
```suggestion
                )
                self.palette.mode = "RGBA"
```
BGRA;15 can be unpacked, but only to RGBA
https://github.com/python-pillow/Pillow/blob/489795422381cd6040aa1efa4e112c9c2975fa82/src/libImaging/Unpack.c#L1620

The change I've suggested here, along with https://github.com/python-pillow/Pillow/pull/7966, will allow the test suite to start passing.

radarhere:
While the test suite now passes, if you actually run
```python
with Image.open("Tests/images/p_16.tga") as im:
    im.save("out.png")
```
you get a transparent image.

Looking at 'Color Map Specification' in https://www.dca.fee.unicamp.br/~martino/disciplinas/ea978/tgaffs.pdf, it says
> When working with VDA or VDA/D cards it is preferred that you select 16 bits (5 bits per primary with 1 bit to select interrupt control) and set the 16th bit to 0 so that the interrupt bit is disabled

I think setting the 16th bit to 0 actually means it is opaque, not transparent like Pillow is used to. I've created https://github.com/Yay295/Pillow/pull/20 to fix this. I've also updated the tests to compare with RGBA images to prevent this problem from slipping by again.

src/PIL/TgaImagePlugin.py
```diff
@@ -36,7 +36,7 @@
     (3, 1): "1",
     (3, 8): "L",
     (3, 16): "LA",
-    (2, 16): "BGR;5",
```
radarhere, line 39: First, this means that we should set the image mode to RGBA.

But second, testing against https://github.com/image-rs/image/blob/main/tests/images/tga/testsuite/utc16.tga, I find that this change also produces a transparent image.

Really, the ideal solution, rather than [my rough solution of inverting bits](https://github.com/Yay295/Pillow/pull/20), would be to add an unpacker that inverted the alpha - I just felt awkward about introducing a new type of unpacker. I've created https://github.com/Yay295/Pillow/pull/22

radarhere (approved):

PR #7971, opened 2024-04-13
Get and set BGR pixels as BGR without scaling

### LINKED ISSUE #7970: BGR;15/16 scaling
Currently when getting a BGR;15/16 pixel value, the result is scaled from 0-31 to 0-255.

https://github.com/python-pillow/Pillow/blob/de18f555686b59466ead87ad21bbc053f070391b/src/libImaging/Access.c#L90-L108

Storing one of these pixels is also scaled, though using a different algorithm.

https://github.com/python-pillow/Pillow/blob/de18f555686b59466ead87ad21bbc053f070391b/src/_imaging.c#L607-L623

Should scaling be done at all for these modes when getting/setting pixels? Also, the pixel order is backwards. It seems that these functions are actually converting to/from RGB rather than just returning the BGR;15/16 values.

Yay295:
Fixes #7970

### DIFF
--- a/Tests/test_image_access.py
+++ b/Tests/test_image_access.py
@@ -137,10 +137,6 @@ class TestImageGetPixel(AccessTest):
         bands = Image.getmodebands(mode)
         if bands == 1:
             return 1
-        if mode in ("BGR;15", "BGR;16"):
-            # These modes have less than 8 bits per band,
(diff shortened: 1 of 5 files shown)


radarhere:
#7978 has been merged, deprecating these modes, so I'm in favour of closing the issue and this PR, rather than introducing a breaking change to deprecated functionality.

PR #7972, opened 2024-04-13
Scale values when converting RGB to BGR;15/16

Yay295:
Currently it just cuts off the lower three bytes. This should give a slightly more accurate conversion.


radarhere:
See https://github.com/python-pillow/Pillow/issues/7970#issuecomment-2053589376 for my thoughts on the need for this.

PR #7978, opened 2024-04-15
Deprecate BGR;15, BGR;16 and BGR;24 modes

radarhere:
Helps #2228


Yay295:
`tobytes()`, `frombytes()`, and `frombuffer()` can also take mode and rawmode parameters through their `*args` parameter.

radarhere:
- [`frombuffer()` calls `frombytes()`](https://github.com/python-pillow/Pillow/blob/e7a0ef44bb725adc894f0513537cd2c0df38f7eb/src/PIL/Image.py#L3021-L3074), and
- [`frombytes()` calls `new()`](https://github.com/python-pillow/Pillow/blob/e7a0ef44bb725adc894f0513537cd2c0df38f7eb/src/PIL/Image.py#L2982-L3008), which contains a deprecation.

The suggestion is not to deprecate the rawmodes, only the modes.

hugovk (approved):

radarhere:
BGR;15, BGR;16 and BGR;24 have now been removed in #9053

Issue #7980, opened 2024-04-15
Can't install Pillow on Ubuntu


iconoclasthero:
So I guessed and it did not work.  Do you have installation instructions elsewhere as the ones linked from the README.md have no information and thus I cannot get started with your project.

```
$ apt install python3-pillow; pipx install pillow
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Note, selecting 'python3-pil' instead of 'python3-pillow'
python3-pil is already the newest version (10.2.0-1).
python3-pil set to manually installed.
The following package was automatically installed and is no longer required:
  libgcrypt20:i386
Use 'sudo apt autoremove' to remove it.
0 upgraded, 0 newly installed, 0 to remove and 144 not upgraded.

No apps associated with package pillow or its dependencies. If you are attempting to install a library, pipx should
not be used. Consider using pip or a similar tool instead.
```

aclark4life:
Moved to https://pillow.readthedocs.io/en/latest/installation/basic-installation.html#basic-installation

iconoclasthero:
Yeah I see that now.
I don't read fine print well.  
It would be great if it just went straight there...that's my suggestion.  

Sorry for the waste of time.

iconoclasthero:
Hey, it's not showing up in pipx and i can't just install things system-wide in Ubuntu so I'm not sure where it is...
```
$ pipx install Pillow pillow

No apps associated with package pillow or its dependencies. If you are attempting to install a library, pipx should
not be used. Consider using pip or a similar tool instead.
```

aclark4life:
No prob! It [has been discussed](https://github.com/python-pillow/Pillow/pull/7832). `pipx install pillow` should work … if not, paste the traceback here.

aclark4life:
Actually, no. Use `pip install pillow`.

```
alexclarks-Air:~:% pipx install pillow
⚠️ Found a space in the home path. We heavily discourage this, due to multiple incompatibilities. Please check our docs for more information on this, as well
    as some pointers on how to migrate to a different home path.

No apps associated with package pillow or its dependencies. If you are attempting to install a library, pipx should not be used. Consider using pip or a
similar tool instead.
```

iconoclasthero:
Yeah, this is where ubuntu-based linux and python run into problems...you can't use pipx for libraries (i just found this out much to my frustration last week) and you can't do what you suggested either:
```
$ pip install pillow
error: externally-managed-environment

× This environment is externally managed
[...]

PR #7832, opened 2024-02-24
Move installation.rst to installation/*.rst

aclark4life:
- Add index
- Move content to sections
- Rearrange sections

Changes proposed in this pull request:

 * Adds navigation to Installation section
 * Rearrange sections to prioritize basic installation, followed by supported platforms and Python versions, followed by building from source and old releases
Screenshot 2024-02-24 at 6 36 01 PM

PR #7981, opened 2024-04-15
Updated installation links

radarhere:
Helps #7980

Replace https://pillow.readthedocs.io/en/latest/installation.html with https://pillow.readthedocs.io/en/latest/installation/basic-installation.html after #7832


aclark4life (approved):
Thanks!

PR #7998, opened 2024-04-21
Deprecate support for libtiff < 4

radarhere:
libtiff 4.0 was released on 21st December 2011, 12 years ago.

I'm hoping that's long enough to not require a deprecation period for removing support for libtiff < 4.

### DIFF
--- a/docs/deprecations.rst
+++ b/docs/deprecations.rst
@@ -100,13 +100,13 @@ ImageMath eval()
 ``ImageMath.eval()`` has been deprecated. Use :py:meth:`~PIL.ImageMath.lambda_eval` or
 :py:meth:`~PIL.ImageMath.unsafe_eval` instead.
 
-Support for libtiff earlier than 4
+Support for LibTIFF earlier than 4
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 .. deprecated:: 10.4.0
 
-Support for libtiff earlier than 4 has been deprecated. Upgrade to a newer version of
-libtiff instead.
+Support for LibTIFF earlier than version 4 has been deprecated.
+Upgrade to a newer version of LibTIFF instead.
 
 Removed features
 ----------------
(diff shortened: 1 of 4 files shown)


hugovk:
And looks like the last v3 release was v3.9.7, released in 2012:

* https://www.asmail.be/msg0054931648.html
* https://gitlab.com/libtiff/libtiff/-/tags?page=2
* http://www.simplesystems.org/libtiff/releases/index.html

If we make an exception to "at least one year of deprecation", I suggest we at least deprecate in Pillow 10.4.0 (July) and remove in 11.0.0 (October). I'd rather play it safe, so there's some warning period, and we still remove it in a SemVer major bump.

radarhere:
If we're playing it safe, then I happy enough to just keep the normal 1 year minimum deprecation.

I've updated the commit.

docs/deprecations.rst
```diff
@@ -100,6 +100,14 @@ ImageMath eval()
 ``ImageMath.eval()`` has been deprecated. Use :py:meth:`~PIL.ImageMath.lambda_eval` or
 :py:meth:`~PIL.ImageMath.unsafe_eval` instead.
 
+Support for libtiff earlier than 4
```
hugovk, line 103:
```suggestion
Support for LibTIFF earlier than 4
```

docs/deprecations.rst
```diff
@@ -100,6 +100,14 @@ ImageMath eval()
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. deprecated:: 10.4.0
+
+Support for libtiff earlier than 4 has been deprecated. Upgrade to a newer version of
+libtiff instead.
```
hugovk, line 109:
```suggestion
Support for LibTIFF earlier than version 4 has been deprecated.
Upgrade to a newer version of LibTIFF instead.
```

hugovk (approved):
Thanks, let's also add to the 10.4.0 release notes.

PR #8001, opened 2024-04-22
Automatically add dates to release notes

hugovk:
This adds a small Sphinx extension to add the date to the titles of the release notes, both on the page and tables of contents.

<table>
<tr>
 <th>Before
 <th>After
<tr>
 <td>
 <td>
</table>

I based this on @jaraco's nifty https://github.com/jaraco/rst.linker, and I've added you as co-author on the commit, hope that's okay?

This PR works by getting the title of pages matching `releasenotes/\d+\.\d+\.\d+"`, then looking up the timestamp from a `git log` call, and then adds "(yyyy-mm-dd)" to the title.

There's some things in rst.linker which make it not suitable to use here, and I think it might not be worth generalising it further:

* It doesn't have wildcard support for the input files
* It's designed to read a date-less file such as https://github.com/pypa/setuptools/blob/main/NEWS.rst which sit out of the docs directory, and create a page with dates like https://setuptools.pypa.io/en/stable/history.html
* It names the output `filename (files).rst` and deletes the original `filename.rst`
* It modifies the `.rst` files, rather than the source output
* It adds the date underneath the title. I considered this, but think it's better for us in the title

Some notes:

* When it can't find a Git tag, this PR adds "(unreleased)" instead of "(yyyy-mm-dd)". This is for when the next version isn't released or tagged yet (see 10.4.0)
* This means, when building from a fork, it will show "(unreleased)" next to any release if you haven't fetched the upstream tags.
* Similarly, it will show "(unreleased)" if you build the docs from a zip or sdist.

I don't think these are problems? For forks, `git fetch upstream --tags` will grab them all; I added it to Read the Docs for forks builds. For sdist builds, we could return early  if a `.git` directory is absent.

### DIFF
--- a/docs/dater.py
+++ b/docs/dater.py
@@ -1,0 +1,52 @@
+"""
+Sphinx extension to add timestamps to release notes based on Git versions.
+
+Based on https://github.com/jaraco/rst.linker, with thanks to Jason R. Coombs.
+"""
+
+from __future__ import annotations
+
+import datetime as dt
+import os
+import re
+import subprocess
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from sphinx.application import Sphinx
+
+DOC_NAME_REGEX = re.compile(r"releasenotes/\d+\.\d+\.\d+")
+VERSION_TITLE_REGEX = re.compile(r"^(\d+\.\d+\.\d+)\n-+\n")
+
+
+def get_date_for(git_version: str) -> dt.datetime | None:
+    cmd = ["git", "log", "-1", "--format=%ai", git_version]
+    try:
+        with open(os.devnull, "w", encoding="utf-8") as devnull:
+            out = subprocess.check_output(
+                cmd, stderr=devnull, text=True, encoding="utf-8"
+            )
+            ts = out.strip()
+        return dt.datetime.fromisoformat(ts)
+    except subprocess.CalledProcessError:
+        return None
+
+
+def add_date(app: Sphinx, doc_name: str, source: list[str]) -> None:
+    if DOC_NAME_REGEX.match(doc_name) and (m := VERSION_TITLE_REGEX.match(source[0])):
+        old_title = m.group(1)
+
+        if tag_datetime := get_date_for(old_title):
+            new_title = f"{old_title} ({tag_datetime:%Y-%m-%d})"
+        else:
+            new_title = f"{old_title} (unreleased)"
+
+        new_underline = "-" * len(new_title)
+
+        result = source[0].replace(m.group(0), f"{new_title}\n{new_underline}\n", 1)
+        source[0] = result
+
+
+def setup(app: Sphinx) -> dict[str, bool]:
+    app.connect("source-read", add_date)
+    return {"parallel_read_safe": True}
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -6,6 +6,10 @@ build:
   os: ubuntu-22.04
   tools:
     python: "3"
+  jobs:
+    post_checkout:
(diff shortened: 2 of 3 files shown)


aclark4life:
Wooo ✨

jaraco:
Thanks for looping me in.

> I've added you as co-author on the commit, hope that's okay?

Perfectly fine.

> There's some things in rst.linker which make it not suitable to use here.

I'd be open to extending rst.linker if that makes sense... or even just exposing some of the functions for customized re-use. No obligation, though.

> * It names the output `filename (files).rst` and deletes the original `filename.rst`

Minor correction - it creates a new file `filename (links).rst` containing the links as a temporary, implementation internal artifact, and it deletes that file after the rendering is done. It doesn't touch the original file.

> It adds the date underneath the title. I considered this, but think it's better for us in the title

I made this choice intentionally in order for the anchors to be predictable (#v69-5-1). With dates in the title, the anchors will include the dates. That's also why I use `v`-prefixed versions (a title must begin with a letter to be used as a stable anchor). I can see how that's of far less value when there's a file per release.

hugovk:
> I'd be open to extending rst.linker if that makes sense... or even just exposing some of the functions for customized re-use. No obligation, though.

Thanks, I think the end result here is quite compact.

> Minor correction - it creates a new file `filename (links).rst` containing the links as a temporary, implementation internal artifact, and it deletes that file after the rendering is done. It doesn't touch the original file.

Ah right, it was just my hacky WIP when I first looked at modifying it :)

> I made this choice intentionally in order for the anchors to be predictable (#v69-5-1). With dates in the title, the anchors will include the dates. That's also why I use `v`-prefixed versions (a title must begin with a letter to be used as a stable anchor). I can see how that's of far less value when there's a file per release.

Thanks for pointing that out. Yeah, that definitely makes sense for https://setuptools.pypa.io/en/stable/history.html#v65-6-1, and doesn't really matter for https://pillow.readthedocs.io/en/stable/releasenotes/9.4.0.html.

docs/dater.py
```diff
@@ -0,0 +1,50 @@
...
+
+DOC_NAME_REGEX = re.compile(r"releasenotes/\d+\.\d+\.\d+")
+VERSION_TITLE_REGEX = re.compile(r"^(\d+\.\d+\.\d+)\n-+\n")
+
+
+def get_date_for(git_version: str) -> dt.datetime | None:
+    cmd = ["git", "log", "-1", "--format=%ai", git_version]
+    try:
+        out = subprocess.check_output(
+            cmd, stderr=subprocess.DEVNULL, text=True, encoding="utf-8"
```
radarhere, line 25: Why are `text` and `encoding` used? From testing locally and from the docs, I would think only one is necessary.

https://docs.python.org/3/library/subprocess.html
> If encoding or errors are specified, or text is true, file objects for stdin, stdout and stderr are opened in text mode using the specified encoding and errors or the [io.TextIOWrapper](https://docs.python.org/3/library/io.html#io.TextIOWrapper) default.

jaraco: Funny story - originally, only `text=True` was used. Then later, when [PEP 597](https://peps.python.org/pep-0597/) introduce encoding warnings when the encoding wasn't specified, indicated to include the encoding. Eventually, the need for `encoding=` will be once again unnecessary (python 3.14 maybe), so we'll want to fall back to `text=`, but until then, both are there. Maybe `text=` could be removed in the interim, but I'm kind-of inclined to leave it there as it's the primary signal.

nulano (approved):
Looks good, and even works out-of-the-box on Windows (at least on my system)!

radarhere (approved):

PR #8002, opened 2024-04-22
Use (void) for empty function parameters

Yay295:
I just happened to notice that these are the only three C functions that don't take any arguments and use `()` instead of `(void)` in their definition.

### DIFF
--- a/src/libImaging/Access.c
+++ b/src/libImaging/Access.c
@@ -185,7 +185,7 @@ put_pixel_32(Imaging im, int x, int y, const void *color) {
 }
 
 void
-ImagingAccessInit() {
+ImagingAccessInit(void) {
 #define ADD(mode_, get_pixel_, put_pixel_)      \
     {                                           \
         ImagingAccess access = add_item(mode_); \
--- a/src/libImaging/JpegDecode.c
+++ b/src/libImaging/JpegDecode.c
@@ -48,7 +48,7 @@ char *libjpeg_turbo_version = NULL;
 #endif
 
 int
-ImagingJpegUseJCSExtensions() {
+ImagingJpegUseJCSExtensions(void) {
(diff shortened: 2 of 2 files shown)


nulano:
> but if those parameters were added there would be a warning due to them being unused.

Can't you just use unnamed parameters to get rid of the warning?

Yay295:
It looks like the solution in other places is just to cast it to the right type.

radarhere:
What is the benefit of `(void)` over `()`? It's just a stylistic choice right, to try and more clearly communicate that there are no arguments?

If this is the general agreement, then sure, go ahead - but to me, requiring casting is a step too far for just a style preference.

Yay295:
Casting is already used for this reason quite a lot elsewhere already. By "other places" I meant "other places in Pillow"; I wasn't talking about other projects.

https://github.com/python-pillow/Pillow/blob/c250a44177ac1e82d300aef702d5c2feef15fd03/src/_webp.c#L530-L536

https://github.com/search?q=repo%3Apython-pillow%2FPillow+%28PyCFunction%29&type=code

radarhere:
> Casting is already used for this reason

None of those castings involve functions that have `(void)` for their parameters.

Yay295:
The WEBP functions were removed in #8213, so it's just the one JPEG function now.

radarhere:
What is the benefit of `(void)` over `()`? It's just a stylistic choice right, to try and more clearly communicate that there are no arguments?

Yay295:
They technically have different meanings in C. `func(void)` means "this function does not take any arguments", while `func()` means "this function takes an undefined number of arguments of undefined types". I believe I saw this come up as a warning somewhere when compiling, but I don't remember where, or what flags I might have enabled,

radarhere (approved):

PR #8008, opened 2024-04-24
Simplified RGB to I;16, I;16L and I;16B conversion

radarhere:
Discussion at https://github.com/python-pillow/Pillow/commit/fb6b8601721da507f590c85a1b9f426f6c45cca6#r141279400 has revealed that when converting from RGB to an I;16* mode, the result is not going to be larger than 8 bits, and so the code can be simplified.


lukegb (approved):
Nice, thanks for the cleanup!

nulano (approved):

PR #8009, opened 2024-04-24
Update tests to allow for zlib-ng

radarhere:
I created https://github.com/python-pillow/docker-images/pull/203 after the release of Fedora 40, but the tests are failing.

This is because Fedora 40 uses [zlib-ng instead of zlib](https://fedoraproject.org/wiki/Releases/40/ChangeSet#Changes/ZlibNGTransition), and [zlib-ng appends '.zlib-ng' to its version string.](https://github.com/zlib-ng/zlib-ng/blob/3f35bfccff2d1dacdfe9844712be1e042d028700/zlib.h.in#L61)

This updates our tests to allow for that suffix.

### DIFF
--- a/Tests/test_features.py
+++ b/Tests/test_features.py
@@ -37,6 +37,8 @@ def test_version() -> None:
         else:
             assert function(name) == version
             if name != "PIL":
+                if name == "zlib" and version is not None:
+                    version = version.replace(".zlib-ng", "")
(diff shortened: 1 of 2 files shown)


hugovk:
GitHub is having some slowness which probably caused the pre-commit.ci and AppVeyor problems. 

https://www.githubstatus.com/incidents/7txcg0j03kkg

Not a blocker to merge. Thanks!

Issue #7906, opened 2024-03-25
OS X: binary installation with PIP includes incompatible "libXau.6.0.0.dylib"


TurtleWilly:
Installing anything that depends on PIL/pillow with PIP on an older Mac OS X/ OS X/ macOS (e.g. OS X Yosemite, 10.10.5) breaks, because the included pre-build "libXau.6.0.0.dylib" seems to be build without a compatibility fallback (aka needs a more recent version of the OS). I'm also not sure why this library is required at all on macOS, but that's just me wondering…

### What did you do?

```bash
# prepare
$ cd /Volumes/Temporary/
$ python3.12 -m venv vcsitest
$ cd vcsitest/
$ source bin/activate

# install/use 'vcsi' as example case. Note: output reduced for clarity
(vcsitest) $ pip install vcsi
Collecting vcsi
  Using cached vcsi-7.0.16-py3-none-any.whl.metadata (16 kB)
…
Collecting pillow<11.0.0,>=10.0.0 (from vcsi)
  Using cached pillow-10.2.0-cp312-cp312-macosx_10_10_x86_64.whl.metadata (9.7 kB)
…
Installing collected packages: texttable, parsedatetime, pillow, numpy, MarkupSafe, jinja2, vcsi
Successfully installed MarkupSafe-2.1.5 jinja2-3.1.3 numpy-1.26.4 parsedatetime-2.6 pillow-10.2.0 texttable-1.7.0 vcsi-7.0.16

# try to run it
(vcsitest) $ bin/vcsi --version
Traceback (most recent call last):
  File "/Volumes/Temporary/vcsitest/bin/vcsi", line 5, in <module>
    from vcsi.vcsi import main
  File "/Volumes/Temporary/vcsitest/lib/python3.12/site-packages/vcsi/vcsi.py", line 33, in <module>
    from PIL import Image, ImageDraw, ImageFont
  File "/Volumes/Temporary/vcsitest/lib/python3.12/site-packages/PIL/Image.py", line 84, in <module>
    from . import _imaging as core
ImportError: dlopen(/Volumes/Temporary/vcsitest/lib/python3.12/site-packages/PIL/_imaging.cpython-312-darwin.so, 2): Library not loaded: @loader_path/libXau.6.0.0.dylib
  Referenced from: /Volumes/Temporary/vcsitest/lib/python3.12/site-packages/PIL/.dylibs/libxcb.1.1.0.dylib
  Reason: no suitable image found.  Did find:
[...]

PR #8010, opened 2024-04-24
GitHub Actions: use `macos-13` or `macos-14` instead of `macos-latest`

hugovk:
GitHub is migrating `macos-latest` to point to `macos-14` which is (a) M1 and (b) does not have Python 3.8 or 3.9:

* https://github.blog/changelog/2024-04-01-macos-14-sonoma-is-generally-available-and-the-latest-macos-runner-image/
* https://github.com/actions/setup-python/issues/696#issuecomment-1637587760

This can cause failed builds for migrated repos.

At least during this migration, let's explicitly use `macos-13` or `macos-14` instead of `macos-latest`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -57,9 +57,9 @@ jobs:
         - python-version: "3.10"
           PYTHONOPTIMIZE: 2
         # M1 only available for 3.10+
-        - os: "macos-latest"
+        - os: "macos-13"
           python-version: "3.9"
-        - os: "macos-latest"
+        - os: "macos-13"
           python-version: "3.8"
         exclude:
         - os: "macos-14"
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -97,7 +97,7 @@ jobs:
       matrix:
         include:
           - name: "macOS x86_64"
-            os: macos-latest
+            os: macos-13
             cibw_arch: x86_64
             macosx_deployment_target: "10.10"
           - name: "macOS arm64"


radarhere (approved):

hugovk:
GitHub has now added Python 3.8 and 3.9 to `macos-14`/Apple Silicon:

https://github.com/actions/setup-python/issues/808#issuecomment-2077670184

We can save about 45s of CI time if we switch them all to `macos-14`:

```
macos-13 / Python 3.9 / 3m 50s	
macos-13 / Python 3.8 / 3m 37s
```

https://github.com/python-pillow/Pillow/actions/runs/8840522550/usage
```
macos-14 / Python 3.9 / 3m 15s
macos-14 / Python 3.8 / 3m 27s
```

https://github.com/hugovk/Pillow/actions/runs/8843777263/usage

Which is a nice little saving, but not huge in the scale of our CI.

The question is, shall we keep testing at least one job on `macos-13`/Intel to increase our test surface?

Might help find some bugs.

radarhere:
Python 3.9.1 was the first version of Python [to support Silicon](https://www.python.org/downloads/release/python-391/), but it looks like it was only backported to Python 3.8 in [3.8.10.](https://www.python.org/downloads/release/python-3810/)

It's not really a strong argument, but that makes me think we should keep `macos-13` until Python 3.8 is EOL at least?

hugovk:
Sure, although we'll still have Intel users for a while longer.

The internet speculates Apple might support Intel until 2025 or 2026 or 2027, and that we'd have to wait to see what Apple actually decides because they've not announced anything.

radarhere:
Regarding when users will actually stop using Intel, it's worth remembering that last month we received [an issue about macOS 10.10](https://github.com/python-pillow/Pillow/issues/7906), which is almost 10 years old at this point.

hugovk:
Indeed, and we don't *have* to support such old versions.

Anyway, I suggest we leave 3.8 and 3.9 on Intel for now.

aclark4life:
Also depends on user demand. If you take away something and people notice it's gone, then we can always put it back. In general, I would drag along as broad a support as you can easily maintain. As the maintenance burden increases for older tech, you can based the decision to drop it on perceived or predicted user demand and act accordingly. If you need to support something you can't easily support, then it gets harder. 🤔

hugovk:
Yeah, it all adds up. I think a fair amount of effort has gone into maintaining old macOS versions, all of which is time away from everything else.

Having clear policies helps both our users and us. We have:

> In general, we aim to support all current versions of Linux, macOS, and Windows.

Which means macOS 12, 13 and 14, not 10.10, 10.11, 10.12, 10.13, 10.14, 10.15 or 11 - 10.10 is very much no longer receiving security updates.

For the time-being, I don't mind keeping 10.10 along.

PR #8011, opened 2024-04-24
Combined conditions

radarhere:


src/_imagingcms.c
```diff
@@ -213,32 +213,31 @@ cms_transform_dealloc(CmsTransformObject *self) {
...
-    } else if (strcmp(PILmode, "RGBX") == 0) {
+    if (
+        strcmp(PILmode, "RGB") == 0 ||
+        strcmp(PILmode, "RGBA") == 0 ||
+        strcmp(PILmode, "RGBX") == 0
+    ) {
         return TYPE_RGBA_8;
     } else if (strcmp(PILmode, "RGBA;16B") == 0) {
         return TYPE_RGBA_16;
     } else if (strcmp(PILmode, "CMYK") == 0) {
```
hugovk, line 224: More of a style issue: these all `return`, I think it reads a little cleaner to remove the `else`s:

```suggestion
    }
    if (strcmp(PILmode, "RGBA;16B") == 0) {
        return TYPE_RGBA_16;
    }
    if (strcmp(PILmode, "CMYK") == 0) {
```
etc.

src/_imagingcms.c
```diff
@@ -213,32 +213,31 @@ cms_transform_dealloc(CmsTransformObject *self) {
...
+    } else if (
+        strcmp(PILmode, "YCCA") == 0 ||
+        strcmp(PILmode, "YCC") == 0
+    ) {
         return TYPE_YCbCr_8;
     } else if (strcmp(PILmode, "LAB") == 0) {
         // LabX equivalent like ALab, but not reversed -- no #define in lcms2
         return (COLORSPACE_SH(PT_LabV2) | CHANNELS_SH(3) | BYTES_SH(1) | EXTRA_SH(1));
     }
     else {
```
hugovk, line 239: And ditch and dedent this block?

src/_imagingcms.c
radarhere: Ok, I've updated the commit.

src/_imagingcms.c
radarhere: I think I know what you meant, I've updated the commit.

hugovk: Yes, like that :)

hugovk (approved):

aclark4life:
Speaking of formatting, anyone interested in adding `clang-format` in some practical way? Maybe one time, since C is not touched very often. 🤔

hugovk:
Yes, like https://github.com/python-pillow/Pillow/issues/4493 / https://github.com/python-pillow/Pillow/pull/4770 ;)

But looks like we forgot to add it to pre-commit/CI and there's some new formatting that can be applied.

aclark4life:
Hmmm, I ran it local and got a bunch of additional changes but … cool! Thanks

PR #8015, opened 2024-04-25
Add ClangFormat to pre-commit

hugovk:
Follow on from https://github.com/python-pillow/Pillow/issues/4493 / https://github.com/python-pillow/Pillow/pull/4770.

As reminded by https://github.com/python-pillow/Pillow/pull/8011#issuecomment-2076902826 we didn't add this to pre-commit/CI at the time. Let's add it now.

Update the others at the same time.


src/_imagingcms.c
```diff
@@ -390,7 +389,7 @@ _buildTransform(
 
     Py_END_ALLOW_THREADS
 
-    if (!hTransform) {
+        if (!hTransform) {
         PyErr_SetString(PyExc_ValueError, "cannot build transform");
     }
```
hugovk, line 394: This doesn't look right. It's caused by the lack of semicolon after `Py_END_ALLOW_THREADS`.

hugovk, line 394: This doesn't look right. It's caused by the lack of semicolon after `Py_END_ALLOW_THREADS`.

src/_imagingcms.c
```diff
@@ -390,7 +389,7 @@ _buildTransform(
 
     Py_END_ALLOW_THREADS
```
hugovk, line 390:
```suggestion
    Py_END_ALLOW_THREADS;
```

src/_imagingcms.c
```diff
@@ -424,7 +423,7 @@ _buildProofTransform(
...
```
hugovk, line 424:
```suggestion
    Py_END_ALLOW_THREADS;
```

Yay295:
A lot of these changes don't really improve readability, imo.

src/_imagingcms.c
nulano: At the time of #4770 this caused a compiler error on Windows; not sure if it is supported now.

src/libImaging/FliDecode.c
```diff
@@ -232,7 +232,8 @@ ImagingFliDecode(Imaging im, ImagingCodecState state, UINT8 *buf, Py_ssize_t byt
                 /* Note, have to check Data + size, not just ptr + size) */
                 if (data + (state->xsize * state->ysize) > ptr + bytes) {
                     /* not enough data for frame */
-                    /* UNDONE Unclear that we're actually going to leave the buffer at the right place. */
+                    /* UNDONE Unclear that we're actually going to leave the buffer at
+                     * the right place. */
```
radarhere, line 236:
```suggestion
                    /* UNDONE Unclear that we're actually going to leave the buffer */
                    /* at the right place. */
```

src/libImaging/GifEncode.c
```diff
@@ -64,156 +72,157 @@ static void glzwe_init(GIFENCODERSTATE *st) {
...
+            /* Hash found experimentally to be pretty good. */
+            /* This works ONLY with TABLE_SIZE a power of 2. */
+            st->probe = ((st->head ^ (st->tail << 6)) * 31) & (TABLE_SIZE - 1);
+            while (st->codes[st->probe]) {
+                if ((st->codes[st->probe] & 0xFFFFF) == ((st->head << 8) | st->tail)) {
+                    st->head = st->codes[st->probe] >> 20;
+                    goto encode_loop;
+                } else {
+                    /* Reprobe decrement must be non-zero and relatively prime to table
+                     * size. So, any odd positive number for power-of-2 size. */
```
radarhere, line 120:
```suggestion
                    // Reprobe decrement must be non-zero and relatively prime to table
                    // size. So, any odd positive number for power-of-2 size.
```

src/libImaging/TiffDecode.c
```diff
@@ -396,7 +413,8 @@ _decodeTile(Imaging im, ImagingCodecState state, TIFF *tiff, int planes, Imaging
 
     if (tile_bytes_size > ((tile_length * state->bits / planes + 7) / 8) * tile_width) {
         // If the tile size as expected by LibTiff isn't what we're expecting, abort.
-        // man:   TIFFTileSize returns the equivalent size for a tile of data as it would be returned in a
+        // man:   TIFFTileSize returns the equivalent size for a tile of data as it
+        // would be returned in a
         //        call to TIFFReadTile ...
```
radarhere, line 418:
```suggestion
        // would be returned in a call to TIFFReadTile ...
```

src/libImaging/TiffDecode.c
```diff
@@ -478,7 +506,8 @@ _decodeStrip(Imaging im, ImagingCodecState state, TIFF *tiff, int planes, Imagin
     unpacker_row_byte_size = (state->xsize * state->bits / planes + 7) / 8;
     if (strip_size > (unpacker_row_byte_size * rows_per_strip)) {
         // If the strip size as expected by LibTiff isn't what we're expecting, abort.
-        // man:   TIFFStripSize returns the equivalent size for a strip of data as it would be returned in a
+        // man:   TIFFStripSize returns the equivalent size for a strip of data as it
+        // would be returned in a
         //        call to TIFFReadEncodedStrip ...
```
radarhere, line 511:
```suggestion
        // would be returned in a call to TIFFReadEncodedStrip ...
```

radarhere (approved):

src/_imagingft.c
```diff
@@ -719,7 +727,7 @@ bounding_box_and_anchors(FT_Face face, const char *anchor, int horizontal_dir, G
 static PyObject *
 font_getsize(FontObject *self, PyObject *args) {
     int width, height, x_offset, y_offset;
-    int load_flags;               /* FreeType load_flags parameter */
+    int load_flags; /* FreeType load_flags parameter */
     int error;
```
radarhere, line 731:
```suggestion
    int error;
    int load_flags;               /* FreeType load_flags parameter */
```
[If the indentation is a concern](https://github.com/python-pillow/Pillow/pull/8015/files#r1579447074), then this could be a solution.

src/_imagingft.c
```diff
@@ -719,7 +727,7 @@ bounding_box_and_anchors(FT_Face face, const char *anchor, int horizontal_dir, G
 static PyObject *
 font_getsize(FontObject *self, PyObject *args) {
     int width, height, x_offset, y_offset;
-    int load_flags;               /* FreeType load_flags parameter */
+    int load_flags; /* FreeType load_flags parameter */
```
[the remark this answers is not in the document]
radarhere: See https://github.com/python-pillow/Pillow/pull/8015/files#r1582006590 for a possible solution.

src/_imagingft.c
```diff
@@ -47,17 +47,17 @@
...
+#include <raqm.h>
+#else
+#include "thirdparty/raqm/raqm.h"
+#ifdef HAVE_FRIBIDI_SYSTEM
+#include <fribidi.h>
+#else
+#include "thirdparty/fribidi-shim/fribidi.h"
+#include <hb.h>
+#endif
+#endif
```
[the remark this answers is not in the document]
radarhere: There is a clang setting to change this, https://stackoverflow.com/a/66004745, but I don't think the effect of the alternative settings on other files looks good.

src/libImaging/JpegEncode.c
```diff
@@ -342,7 +346,7 @@ ImagingJpegEncode(Imaging im, ImagingCodecState state, UINT8 *buf, int bytes) {
             }
             jpeg_finish_compress(&context->cinfo);
 
-cleanup:
+        cleanup:
```
radarhere, line 349: If we wanted to not add indentation, we could add in `IndentGotoLabels: false` - https://github.com/hugovk/Pillow/pull/118

src/libImaging/FliDecode.c
hugovk: It looks a little odd to mix the comment style with the preceding line, let's use `//` style for the `/* not enough data for frame */` as well.

PEP 7 uses C++-style `//` https://peps.python.org/pep-0007/#c-dialect but we don't need to go through and change them everywhere.

radarhere: Sure - https://github.com/hugovk/Pillow/pull/119

src/_imagingft.c
radarhere: Hmm, every variable in this function is declared at the top at the moment, and I think the consistency of that is good.

I'm happy just leaving it as it currently is in this PR.

Yay295:
Currently in Python when a function declaration is split across multiple lines, the closing parentheses is on the same level as the `def`.

https://github.com/python-pillow/Pillow/blob/c250a44177ac1e82d300aef702d5c2feef15fd03/src/PIL/Image.py#L3484-L3488

However with this change in C the closing parentheses is on the same line as the last argument.

https://github.com/hugovk/Pillow/blob/5597f618a3b5cf8e208dded44a2107313c47122d/src/libImaging/BcnDecode.c#L825-L833

```c
static int
decode_bcn(
    Imaging im,
    ImagingCodecState state,
    const UINT8 *src,
    int bytes,
    int N,
    int C,
    char *pixel_format) {
```

It's the same with function calls as well. It looks like this is changed with `AlignAfterOpenBracket: BlockIndent`.

radarhere:
> However with this change in C the closing parenthesis is on the same line as the last argument.

From what I can see, this is the style of C that is already in place. For the example you've cited, here how it is in main.
https://github.com/python-pillow/Pillow/blob/58a47978af9f34851ce926303d05fa677010ce2a/src/libImaging/BcnDecode.c#L827-L829

Here are some other examples.

https://github.com/python-pillow/Pillow/blob/58a47978af9f34851ce926303d05fa677010ce2a/src/_imagingft.c#L250-L257
https://github.com/python-pillow/Pillow/blob/58a47978af9f34851ce926303d05fa677010ce2a/src/_imagingcms.c#L404-L412

Unless I've missed something in looking at the changes here, this PR isn't even moving any closing brackets from a newline onto the same line as the last argument - it's just making the existing style slightly more apparent because it's splitting up some long lines.

I'm not opposed to changing the style for its own sake, I just don't think the current state of this PR affects what you're describing.

Yay295:
Ah, yeah that does seem to be the case. I didn't find any instances of `) {` at the start of a line, other than three `if` blocks.
https://github.com/python-pillow/Pillow/blob/58a47978af9f34851ce926303d05fa677010ce2a/src/_imagingcms.c#L220
https://github.com/python-pillow/Pillow/blob/58a47978af9f34851ce926303d05fa677010ce2a/src/_imagingcms.c#L238
https://github.com/python-pillow/Pillow/blob/58a47978af9f34851ce926303d05fa677010ce2a/src/libImaging/TiffDecode.c#L713
I would prefer the other format, but that could be a separate change.

PR #8016, opened 2024-04-25
Make sure third-party libraries are found in debug build

mrbean-bremen:
- _find_library_file now also checks debug libraries if in debug mode and returns the library name
- the returned library name is used for linking instead of the hardcoded name

We build Pillow in debug mode and had the problem that the debug libraries (ending with "_d") were not found.
Using this patch fixed this.

I did not add release notes, as this is internal build stuff only - let me know if they are needed (in case this PR will get accepted).


setup.py
```diff
@@ -225,12 +225,18 @@ def _find_include_file(self, include):
...
         _dbg("Found library %s at %s", (library, ret))
-    else:
-        _dbg("Couldn't find library %s in %s", (library, self.compiler.library_dirs))
-    return ret
+        # we are only interested in the library name including a possible debug suffix
+        lib_name_base = os.path.basename(ret).split(".")[0]
+        # as the library prefix differs depending on library type and platform,
+        # we ignore it by looking for the actual library name
+        start_index = lib_name_base.find(library)
+        return lib_name_base[start_index:]
```
mrbean-bremen, line 238: This is a bit hacky, but I didn't find a better way to get the library name from the file name. I could not use `splitext`, as that would not handle something like `libfoo.so.6`. 
As this is an internal function only, it should be ok.

radarhere:
Are you able to put together a Dockerfile that demonstrates the need for this?

mrbean-bremen:
> Are you able to put together a Dockerfile that demonstrates the need for this?

Sorry, missed that comment.

I have to check... we are building using conan (we also build the other dependencies), that would probably be overkill here. I first thaught to add a CI wheel build in debug mode, but that is hardly needed by anyone, so it would only make the build process even longer. Will think about it...

mrbean-bremen:
The problem happens only if building against the debug libraries, which is not a common scenario (especially under Linux), except you build the libraries yourself (which we do). To reproduce this in a container, we would have to do this first (e.g. build the thirdparty libraries), as to my knowledge there is no ready container that already has them. As I wrote, we use conan to build the thirdparty libraries (and pillow) and store the build libraries in an artifactory, which would not be an option in the CI.

So, at the moment I don't have an idea how to test this with reasonable effort. I still think that the scenario is not completely outlandish, and it would make sense to support it (given that the underlying `find_library_file` from `distutils` supports it), but I can understand if you don't want to merge it without a test. We would patch it ourselves in this case.

And, by the way, thanks for all the work you put into this project, and for the fast and helpful feedback!

mrbean-bremen:
Hm, the failing 3.14 build is probably due to [this issue](https://github.com/python/cpython/issues/133779) (fixed for 3.14b2), not related to the PR.

setup.py
```diff
@@ -759,35 +744,40 @@ def build_extensions(self) -> None:
...
 
-            if best_version and _find_library_file(self, "openjp2"):
-                # Add the directory to the include path so we can include
-                # <openjpeg.h> rather than having to cope with the versioned
-                # include path
-                _add_directory(self.compiler.include_dirs, best_path, 0)
-                feature.set("jpeg2000", "openjp2")
+            if best_version:
+                lib_jpeg2k = _find_library_file(self, "openjp2")
+                if lib_jpeg2k is not None:
```
```suggestion
                if lib_jpeg2k := _find_library_file(self, "openjp2"):
```

setup.py
```diff
@@ -759,35 +744,40 @@ def build_extensions(self) -> None:
...
+                if lib_tiff is None:
+                    lib_tiff = _find_library_file(self, "tiff")
+                if lib_tiff is not None:
+                    feature.set("tiff", lib_tiff)
 
         if feature.want("freetype"):
             _dbg("Looking for freetype")
-            if _find_library_file(self, "freetype"):
+            lib_freetype = _find_library_file(self, "freetype")
+            if lib_freetype is not None:
```
```suggestion
            if lib_freetype := _find_library_file(self, "freetype"):
```

setup.py
```diff
@@ -804,27 +794,29 @@ def build_extensions(self) -> None:
...
                 _dbg("Looking for HarfBuzz")
                 feature.set("harfbuzz", None)
                 hb_dir = _find_include_dir(self, "harfbuzz", "hb.h")
                 if hb_dir:
                     if isinstance(hb_dir, str):
                         _add_directory(self.compiler.include_dirs, hb_dir, 0)
-                    if _find_library_file(self, "harfbuzz"):
-                        feature.set("harfbuzz", "harfbuzz")
+                    lib_harfbuzz = _find_library_file(self, "harfbuzz")
+                    if lib_harfbuzz is not None:
```
```suggestion
                    if lib_harfbuzz := _find_library_file(self, "harfbuzz"):
```

setup.py
```diff
@@ -835,20 +827,21 @@ def build_extensions(self) -> None:
                                 _add_directory(
                                     self.compiler.include_dirs, fribidi_dir, 0
                                 )
-                            if _find_library_file(self, "fribidi"):
-                                feature.set("fribidi", "fribidi")
+                            lib_fribidi = _find_library_file(self, "fribidi")
+                            if lib_fribidi is not None:
```
```suggestion
                            if lib_fribidi := _find_library_file(self, "fribidi"):
```

setup.py
```diff
@@ -862,14 +855,15 @@ def build_extensions(self) -> None:
...
+                        feature.set("webp", _find_library_file(self, prefix + "webp"))
                         break
 
         if feature.want("xcb"):
             _dbg("Looking for xcb")
             if _find_include_file(self, "xcb/xcb.h"):
-                if _find_library_file(self, "xcb"):
-                    feature.set("xcb", "xcb")
+                lib_xcb = _find_library_file(self, "xcb")
+                if lib_xcb is not None:
```
```suggestion
                if lib_xcb := _find_library_file(self, "xcb"):
```

setup.py
```diff
@@ -878,8 +872,10 @@ def build_extensions(self) -> None:
                     major_version = int(
                         fp.read().split(b"#define AVIF_VERSION_MAJOR ")[1].split()[0]
                     )
-                    if major_version >= 1 and _find_library_file(self, "avif"):
-                        feature.set("avif", "avif")
+                    if major_version >= 1:
+                        lib_avif = _find_library_file(self, "avif")
+                        if lib_avif is not None:
```
```suggestion
                        if lib_avif := _find_library_file(self, "avif"):
```

mrbean-bremen:
@radarhere - Thanks, makes sense - I applied your suggestions.

mrbean-bremen:
This got broken by the merges from main - I'm re-creating the PR in a new branch.

PR #8017, opened 2024-04-25
Consolidate getdata and putdata tests

Yay295:
Moved all of the `getdata` and `putdata` tests to the same file (because `test_getdata_putdata` doesn't fit in either individual file), and parametrized two tests.


Tests/test_image_getdata_putdata.py
```diff
@@ -29,36 +38,56 @@ def test_sanity() -> None:
...
+    (
+        (0xFFFFFFFF, (255, 255, 255, 255)),
+        (-1, (255, 255, 255, 255)),
+        (sys.maxsize, (255, 255, 255, 255 if sys.maxsize > 2**32 else 127)),
+    ),
+)
+def test_putdata_long_integers(value: int, pixel: tuple[int, int, int, int]) -> None:
+    im = Image.new("RGBA", (1, 1))
+    im.putdata([value])
+    assert im.getpixel((0, 0)) == pixel
```
nulano, line 75:
```suggestion
@pytest.mark.parametrize(
    "value, expected",
    (
        (0xFFFFFFFF, (255, 255, 255, 255)),
        (-1, (255, 255, 255, 255)),
        (sys.maxsize, (255, 255, 255, 255 if sys.maxsize > 2**32 else 127)),
    ),
)
def test_putdata_long_integers(value: int, expected: tuple[int, int, int, int]) -> None:
    im = Image.new("RGBA", (1, 1))
    im.putdata([value])
    assert im.getpixel((0, 0)) == expected
```

radarhere:
test_image_getdata_putdata.py seems like a mouthful. How about just test_image_data.py?

Yay295:
sure

radarhere:
> Moved all of the getdata and putdata tests to the same file (because test_getdata_putdata doesn't fit in either individual file)

Sorry, I've just realised what this meant. In main, we currently have [test_image_getdata.py](https://github.com/python-pillow/Pillow/blob/main/Tests/test_image_getdata.py) and [test_image_putdata.py](https://github.com/python-pillow/Pillow/blob/main/Tests/test_image_putdata.py), and here you've merged them together into a single file because you'd like to find a new home for a test from test_image.py / #7209, and you don't think a roundtrip test belongs in either file.

I think test_image_putdata.py already has roundtrip tests, using both `getdata()` and `putdata()`.
https://github.com/python-pillow/Pillow/blob/c250a44177ac1e82d300aef702d5c2feef15fd03/Tests/test_image_putdata.py#L60-L68

Yay295:
I'd say those should be in the combined file for the same reason. Though I suppose I shouldn't have added `putdata` to those function names.

Yay295:
Cleaned up the tests a bit more.

Yay295:
rebased, multiple times because I didn't check it locally and kept missing something when resolving the conflict

Yay295:
What is currently blocking this from being merged?

radarhere:
[As I've said](https://github.com/python-pillow/Pillow/pull/8017#issuecomment-2081325752), my opinion is that this change isn't needed. If others think it is though, feel free to outvote me.

PR #8019, opened 2024-04-26
Add packers for RGB;15/16 and BGR;15/16

Yay295:
I made these to be the opposite of the unpackers for these rawmodes, but this conversion method isn't actually roundtrippable. I know of two other methods that are roundtrippable, but I would have to change the unpackers as well for that.

### DIFF
--- a/Tests/test_lib_pack.py
+++ b/Tests/test_lib_pack.py
@@ -67,6 +67,10 @@ class TestLibPack:
 
     def test_RGB(self) -> None:
         self.assert_pack("RGB", "RGB", 3, (1, 2, 3), (4, 5, 6), (7, 8, 9))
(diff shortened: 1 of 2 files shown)


radarhere:
What is the motivation behind this? Are these packers something that you plan on using in your own external code that uses Pillow?

Yay295:
It's currently possible to create or convert an image to BGR;15/16 and save it as that mode because there's a copy packer for those modes, but when those modes are removed that will no longer be possible. By adding these packers it will still be possible to save data in these formats.

I added the RGB versions because it's basically the same code with just two variables swapped.

radarhere:
If you've moved on and are [coming up with a different plan](https://github.com/python-pillow/Pillow/issues/8021#issuecomment-2080760298), can this be closed?

PR #6531, opened 2022-08-24
Parametrize more tests

Yay295:

PR #8026, opened 2024-04-27
Parametrize test_lib_pack.py

Yay295:


Tests/test_lib_pack.py
```diff
@@ -1,820 +1,835 @@
...
+        ("L", "L", 0, [0]),
+        ("RGB", "RGB", 2, [0]),
+        ("CMYK", "CMYK", 2, [0]),
+    ),
+)
+def test_unpack_valueerror(
+    mode: str, rawmode: str, data: int | bytes, pixels: list[float | tuple[int, ...]]
+) -> None:
+    with pytest.raises(ValueError):
+        test_unpack(mode, rawmode, data, pixels)
```
radarhere, line 835: Calling one test from another seems odd, and a potential source of confusion?

Yay295: I could make `test_unpack` not a test function, and have all three test functions call that, but I don't see how that would be any clearer.

radarhere:
I understand there's a lot of repetition in this file, but I don't think that this version is easier to understand.

If others think it is easier to understand, that's fine, but those are my thoughts on the matter.

Yay295:
It's much easier to understand when a test fails.

radarhere:
While I recognise I'm not [as fond of parametrisation as others](https://github.com/python-pillow/Pillow/pull/6531#issuecomment-1225680091), I think over 200 and 500 lines of code to construct the parameters is a bit much.

If the concern is about clarity when a test fails, maybe just adding a message to the assertion would be a smaller change with a similar effect? Something like this, but with the details about which case failed as the message.
https://github.com/python-pillow/Pillow/blob/b55d74bcfe4b4e69a64d09c785552af8dc9f013d/Tests/test_file_webp_metadata.py#L100

Yay295:
The main benefit I see for parametrization is that each assert can fail on its own, instead of the test stopping at the first failed assert.

radarhere:
> The main benefit I see for parametrization is that each assert can fail on its own, instead of the test stopping at the first failed assert.

With our use of `-x` though, I don't think that's something that is a priority in our CI testing.
https://github.com/python-pillow/Pillow/blob/329d6a6a6276a03ed67725ffb367adac5c769469/.ci/test.sh#L7

PR #8032, opened 2024-04-29
Added type hints for PixelAccess related methods and others

nulano:


src/PIL/ImageGrab.py
```diff
@@ -63,6 +69,7 @@ def grab(bbox=None, include_layered_windows=False, all_screens=False, xdisplay=N
                 left, top, right, bottom = bbox
                 im = im.crop((left - x0, top - y0, right - x0, bottom - y0))
             return im
+    xdisplay = cast(Union[str, None], xdisplay)  # type: ignore[redundant-cast, unused-ignore]
```
nulano, line 72: This cast is required on Windows and macOS.

src/PIL/PyAccess.py
```diff
@@ -357,9 +365,13 @@ def set_pixel(self, x, y, color):
...
+def new(img: Image.Image, readonly: bool = False) -> PyAccess | None:
     access_type = mode_map.get(img.mode, None)
     if not access_type:
         logger.debug("PyAccess Not Implemented: %s", img.mode)
         return None
     return access_type(img, readonly)
+
+
+if TYPE_CHECKING:
+    from . import Image
```
radarhere, line 377: You didn't want to put this import at the top of the file?

hugovk: For lighter/quicker imports when not type checking, and potentially avoid circular imports?

radarhere: The convention so far has been to put the `TYPE_CHECKING` imports just before their use - [`OPEN`](https://github.com/python-pillow/Pillow/blob/58a47978af9f34851ce926303d05fa677010ce2a/src/PIL/Image.py#L220-L221), [`Exif`](https://github.com/python-pillow/Pillow/blob/58a47978af9f34851ce926303d05fa677010ce2a/src/PIL/Image.py#L3698-L3701), [`ImageFileDirectory_v2`](https://github.com/python-pillow/Pillow/blob/58a47978af9f34851ce926303d05fa677010ce2a/src/PIL/TiffImagePlugin.py#L484-L487) and [`PdfDict`](https://github.com/python-pillow/Pillow/blob/58a47978af9f34851ce926303d05fa677010ce2a/src/PIL/PdfParser.py#L251-L254)

nulano: > For lighter/quicker imports when not type checking, and potentially avoid circular imports?

I hadn't realized mypy has no problem dealing with circular imports. I've now moved the import near the top (still with the `TYPE_CHECKING` check).

src/PIL/Image.py
```diff
@@ -847,7 +881,7 @@ def load(self):
         operations. See :ref:`file-handling` for more information.
 
         :returns: An image access object.
-        :rtype: :ref:`PixelAccess` or :py:class:`PIL.PyAccess`
+        :rtype: :py:class:`.PixelAccess` or :py:class:`.PyAccess`
```
radarhere, line 884: My initial reaction to this was to feel confused , because the return type is PixelAccess or None, but this says PixelAccess or PyAccess.

Of course, the return type uses PixelAccess the protocol, not the class. But should the protocol have a different name perhaps to avoid confusion? `SupportsPixelAccess`?

radarhere: Be aware that this will become simpler after Pillow 10.4.0, as PyAccess will end deprecation and be removed.

src/PIL/ImageGrab.py
```diff
@@ -94,7 +101,7 @@ def grab(bbox=None, include_layered_windows=False, all_screens=False, xdisplay=N
         return im
 
 
-def grabclipboard():
+def grabclipboard() -> Image.Image | list[str] | None:
```
radarhere, line 104: This could be `ImageFile.ImageFile`, rather than `Image.Image`

src/PIL/Image.py
```diff
@@ -1977,7 +2013,7 @@ def putpalette(self, data, rawmode="RGB") -> None:
         self.palette.mode = "RGB"
         self.load()  # install new palette
 
-    def putpixel(self, xy, value):
+    def putpixel(self, xy: tuple[int, int], value: float | tuple[int, ...]) -> None:
```
radarhere, line 2016: It would appear `value` can also be a list.
https://github.com/python-pillow/Pillow/blob/c7574392b050bf93334ec3c1b1873de2cc3487fb/src/PIL/Image.py#L2035

src/PIL/Image.py
```diff
@@ -1627,7 +1660,9 @@ def apply_transparency(self):
 
         del self.info["transparency"]
 
-    def getpixel(self, xy):
+    def getpixel(
+        self, xy: tuple[SupportsInt, SupportsInt]
```
radarhere, line 1664: The test suite does provide a list for `xy`
https://github.com/python-pillow/Pillow/blob/95a69ec6989deda8deb55b6fe842205327de9ca6/Tests/test_image_access.py#L217-L219

Also, why use `SupportsInt`, and not just `int`?

radarhere:
I've merged some of these changes in #8099

src/PIL/ImageGrab.py
nulano: I'm not opposed to changing this to `ImageFile`, but would that unnecessarily lock us in to a given return type?
Or can we change return types without a major version bump and deprecation period?

radarhere: I hope that we can change return types without deprecation, but I can appreciate that we could be a bit forwards-compatible here. There's also no reason for the user to expect `ImageFile` instead of an `Image`. So I think either form is fine.

src/PIL/Image.py
radarhere: I think that after 10.4.0, we'll have no need for the protocol - we can just use `core.PixelAccess`. If so, then it's not helpful to add a public protocol only to remove it in the next version. I've created https://github.com/nulano/Pillow/pull/39 to return `core.PixelAccess | PyAccess.PyAccess | None` instead.

nulano: Fair enough, I've pushed a similar commit, see comment in https://github.com/nulano/Pillow/pull/39.

docs/reference/PixelAccess.rst
```diff
@@ -44,42 +44,23 @@ Access using negative indexes is also possible. ::
...
+        Returns the pixel at x,y. The pixel is returned as a single
+        value for single band images or a tuple for multi-band images.
 
         :param xy: The pixel coordinate, given as (x, y).
         :returns: a pixel value for single band images, a tuple of
-          pixel values for multiband images.
+                  pixel values for multiband images.
 
-  .. method:: putpixel(self, xy, color):
+  .. method:: __setitem__(self, xy: tuple[int, int], color: float | tuple[int, ...]) -> None
```
radarhere, line 58: Was removing the colon at the end of this and `__getitem__` deliberate?

radarhere (approved):

hugovk (approved):

docs/reference/PixelAccess.rst
radarhere: Oh, I see. Restoring them causes Sphinx to complain.
```
/home/runner/work/Pillow/Pillow/docs/reference/PixelAccess.rst:49: WARNING: py:class reference target not found: float | tuple[int, ...]:
/home/runner/work/Pillow/Pillow/docs/reference/PixelAccess.rst:58: WARNING: py:class reference target not found: None:
```

Issue #8034, opened 2024-04-30
UnidentifiedImageError opening simple PSD using shapes and masks


ppearsonhutch:
### What did you do?
Attempted to open the supplied psd image with PIL.open() in python.

I didn't actually create this image myself, nor can I figure out what exactly is wrong with it, but I've managed to reduce the cause of the issue down to 2 shape layers, one with a layer mask. Removing either layer fixes the error. Additionally, inserting a new empty layer between the two layers also fixes the error. So it's not critical and I can work around it, but I have a feeling it's a bug in the psd opening logic if it's somehow fixed by adding blank layers.

Image below of the layers:
<img width="581" alt="Screenshot 2024-04-30 at 21 34 44" src="https://github.com/python-pillow/Pillow/assets/47038883/697a2444-3742-4c8d-ae9c-cbc7ae38d998">

### What did you expect to happen?
No PIL.UnidentifiedImageError when opening the PSD.

### What actually happened?
Received a PIL.UnidentifiedImageError when opening the PSD with PIL.open()

### What are your OS, Python and Pillow versions?

* OS: MacOS Sonoma 14.4.1
* Python: 3.9.6 (via shim/pyenv)
* Pillow: 10.3.0 (via pip)

```
--------------------------------------------------------------------
Pillow 10.3.0
Python 3.9.6 (default, Apr  6 2024, 22:52:34)
       [Clang 15.0.0 (clang-1500.3.9.4)]
--------------------------------------------------------------------
Python executable is /Users/<username>/.pyenv/versions/3.9.6/bin/python3
System Python files loaded from /Users/<username>/.pyenv/versions/3.9.6
--------------------------------------------------------------------
Python Pillow modules loaded from /Users/<username>/.pyenv/versions/3.9.6/lib/python3.9/site-packages/PIL
Binary Pillow modules loaded from /Users/<username>/.pyenv/versions/3.9.6/lib/python3.9/site-packages/PIL
--------------------------------------------------------------------
--- PIL CORE support ok, compiled for 10.3.0
--- TKINTER support ok, loaded 8.6
--- FREETYPE2 support ok, loaded 2.13.2
--- LITTLECMS2 support ok, loaded 2.16
--- WEBP support ok, loaded 1.3.2
--- WEBP Transparency support ok
--- WEBPMUX support ok
--- WEBP Animation support ok
--- JPEG support ok, compiled for libjpeg-turbo 3.0.2
[...]

PR #8039, opened 2024-05-02
Do not read layers immediately when opening PSD images

radarhere:
When opening a PSD image, this PR skips reading and parsing the PSD layers until the user calls `im.layers`, `im.n_frames` or `im.is_animated`.

This will improve performance, and also allow users who just want to `open()` an image without seeking to workaround any errors that might be raised when parsing layers, such as in #8034

### DIFF
--- a/Tests/test_file_psd.py
+++ b/Tests/test_file_psd.py
@@ -4,7 +4,7 @@ import warnings
 
 import pytest
 
-from PIL import Image, PsdImagePlugin, UnidentifiedImageError
+from PIL import Image, PsdImagePlugin
 
 from .helper import assert_image_equal_tofile, assert_image_similar, hopper, is_pypy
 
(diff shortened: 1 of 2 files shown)


hugovk (approved):
👍 

`main`:

```console
❯ python -m timeit -s "from PIL import Image" "Image.open('Tests/images/hopper.psd')"
2000 loops, best of 5: 166 usec per loop

❯ python -m timeit -s "from PIL import Image" "im = Image.open('Tests/images/hopper.psd'); im.layers"
2000 loops, best of 5: 164 usec per loop
```

PR:

```console
❯ python -m timeit -s "from PIL import Image" "Image.open('Tests/images/hopper.psd')"
2000 loops, best of 5: 101 usec per loop

❯ python -m timeit -s "from PIL import Image" "im = Image.open('Tests/images/hopper.psd'); im.layers"
2000 loops, best of 5: 164 usec per loop
```

(These runs were actually the third ones, to make sure everything had imported properly and avoid delays after an initial build.)

PR #8045, opened 2024-05-07
Do not use percent format in strings

radarhere:
Fixes the failures in https://github.com/python-pillow/Pillow/pull/8044

Converts several percent format strings to f-strings.

There is also a percent format inside a multiline string - I think the neatest solution is just to break the multiline string.

### DIFF
--- a/Tests/test_image_access.py
+++ b/Tests/test_image_access.py
@@ -409,15 +409,14 @@ class TestEmbeddable:
         from setuptools.command import build_ext
 
         with open("embed_pil.c", "w", encoding="utf-8") as fh:
+            home = sys.prefix.replace("\\", "\\\\")
             fh.write(
-                """
+                f"""
 #include "Python.h"
 
 int main(int argc, char* argv[])
-{
-    char *home = \""""
-                + sys.prefix.replace("\\", "\\\\")
-                + """\";
+{{
+    char *home = "{home}";
     wchar_t *whome = Py_DecodeLocale(home, NULL);
     Py_SetPythonHome(whome);
 
@@ -432,7 +431,7 @@ int main(int argc, char* argv[])
     PyMem_RawFree(whome);
 
     return 0;
-}
+}}
         """
             )
 
--- a/Tests/test_image_access.py
+++ b/Tests/test_image_access.py
@@ -415,7 +415,9 @@ class TestEmbeddable:
 
 int main(int argc, char* argv[])
 {
-    char *home = "%s";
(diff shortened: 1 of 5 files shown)


Yay295:
created https://github.com/radarhere/Pillow/pull/25 for the `PdfParser` change

src/PIL/PdfParser.py
```diff
@@ -825,8 +825,9 @@ def get_value(cls, data, offset, expect_indirect=None, max_nesting=-1):
                 try:
                     stream_len = int(result[b"Length"])
                 except (TypeError, KeyError, ValueError) as e:
-                    msg = "bad or missing Length in stream dict (%r)" % result.get(
-                        b"Length", None
+                    msg = (
+                        "bad or missing Length in stream dict "
+                        f"({result.get(b'Length')})"
                     )
```
radarhere, line 831:
```suggestion
                    stream_len = result.get(b"Length")
                    msg = f"bad or missing Length in stream dict ({stream_len})"
```
https://github.com/radarhere/Pillow/pull/25 feels that error message should fit onto one line. The change above this would be my suggestion for that goal.

The suggestion from the PR is
```python
stream_len_str = result.get(b"Length")
try:
	stream_len = int(stream_len_str)
except (TypeError, KeyError, ValueError) as e:
	msg = f"bad or missing Length in stream dict ({stream_len_str})"
	raise PdfFormatError(msg) from e
```

I think that makes the code harder to read, because, if you ignore the exception, there's more to understand.

nulano: My suggestion would be: https://github.com/radarhere/Pillow/pull/25#discussion_r1592841987

radarhere: Ok, I've pushed that - minus the `KeyError`, because now that `result[b"Length"]` is gone, it will no longer happen. `int(None)` will instead raise a TypeError.

nulano (approved):

Tests/test_image_access.py
```diff
@@ -415,7 +415,9 @@ def test_embeddable(self) -> None:
 
 int main(int argc, char* argv[])
 {
-    char *home = "%s";
+    char *home = \""""
+                + sys.prefix.replace("\\", "\\\\")
```
hugovk, line 419: Can we put this in a variable outside the string, then make this an f-string?

radarhere: Ok, I've pushed a commit. It did require escaping brackets though.

PR #8046, opened 2024-05-07
Add various type annotations

srittau:

 * This PR add various type annotations, especially for the `Image`, `ImageFont`, `ImageDraw`, and `_imagingft` modules.


nulano:
FWIW I have some WIP type annotations for ImageFont and _imagingft here: https://github.com/python-pillow/Pillow/compare/main...nulano:Pillow:types-imagefont

hugovk:
Yeah, the PyPy one is a bit flaky.

radarhere:
> The docs build might still fail, but I'd need help with that.

The [error](https://github.com/python-pillow/Pillow/actions/runs/9000614869/job/24725200519?pr=8046#step:8:274)
> docstring of PIL.Image.Image.transform:1: WARNING: py:class reference target not found: PIL.Image.GetDataTransform

is saying that the `GetDataTransform` doesn't appear in the documentation, so `transform()`'s description can't link to it.

The way to solve this is by updating [docs/reference/Image.rst](https://github.com/python-pillow/Pillow/blob/main/docs/reference/Image.rst) to include it, by adding
```
.. autoclass:: GetDataTransform
    :show-inheritance:
```

But given that our other protocols so far are `SupportsGetMesh` and `SupportsArrayInterface`, maybe this new protocol could be renamed `SupportsGetData`?

src/PIL/ImageFont.py
```diff
@@ -33,12 +33,18 @@
...
-from typing import BinaryIO
+from typing import TYPE_CHECKING, BinaryIO
+
+from PIL import ImageFile
 
 from . import Image
 from ._typing import StrOrBytesPath
 from ._util import is_directory, is_path
 
+if TYPE_CHECKING:
```
radarhere, line 44:
```suggestion
from . import Image
from ._typing import StrOrBytesPath
from ._util import is_directory, is_path

if TYPE_CHECKING:
    from . import ImageFile
```

src/PIL/_imaging.pyi
```diff
@@ -1,3 +1,18 @@
 from typing import Any
 
+from typing_extensions import Buffer
```
radarhere, line 3: Just pointing out that is adding typing_extensions as a requirement. Previous discussion about this can be seen in #7642

src/PIL/Image.py
```diff
@@ -3774,7 +3802,7 @@ def _get_ifd_dict(self, offset, group=None):
             return self._fixup_dict(info)
 
     def _get_head(self):
-        version = b"\x2B" if self.bigtiff else b"\x2A"
+        version = b"\x2b" if self.bigtiff else b"\x2a"
```
radarhere, line 3805: This is minor, but what is the reason behind this change? Do you just think lowercase is better, or is it more consistent with other parts of Pillow?

src/PIL/_imaging.pyi
srittau: This is not super critical. We could use `bytes | bytesarray | memoryview` here, instead. This would exclude other types using the buffer protocol, but I'm not sure how common that is with Pillow. (And it can always be `# type: ignore`d, of course.)

src/PIL/Image.py
srittau: I though pre-commit slipped that change in. Seems not, so it's probably something my IDE did. I'll revert this unnecessary change.

src/PIL/ImageTransform.py
```diff
@@ -34,7 +34,7 @@ def transform(
         self,
         size: tuple[int, int],
         image: Image.Image,
-        **options: dict[str, str | int | tuple[int, ...] | list[int]],
```
radarhere, line 37: Could you explain why you made this change?

src/PIL/ImageFont.py
```diff
@@ -56,7 +61,7 @@ class Layout(IntEnum):
     core = DeferredError.new(ex)
 
 
-def _string_length_check(text):
+def _string_length_check(text: str | bytes) -> None:
```
radarhere, line 64:
```suggestion
def _string_length_check(text: str) -> None:
```
This is an internal method that will only be passed strings when it is called elsewhere in this file.

src/PIL/ImageTransform.py
srittau: The problem here is that `options` can be any of the remaining `Image.transform` arguments, i.e.:

```
        resample: int = Resampling.NEAREST,
        fill: int = 1,
        fillcolor: float | tuple[float, ...] | str | None 
```

`dict` is wrong in any case, but we could annotate it as `float | tuple[float, ...] | str | None` (with `int` being redundant with `float`), but that would imply that any argument could have that type. I see four alternatives:

* Don't bother and use `Any`.
* Use `float | tuple[float, ...] | str | None` as that would prevent some typing errors, but not all (e.g. `fill=(1,2,3)` would not be caught), at the cost of a slightly misleading type hints.
* Just replace `options` with the actual forwarded arguments. (The option I would probably choose, as there are only three options.)
* Use `Unpack`, which means we would need to introduce a new `TypedDict` for the arguments.

radarhere:
I've created https://github.com/srittau/Pillow/pull/2 with various suggestions.

src/PIL/_imaging.pyi
radarhere: I removed this in https://github.com/srittau/Pillow/pull/2

radarhere (approved):

src/PIL/Image.py
```diff
@@ -2211,13 +2228,9 @@ def resize(self, size, resample=None, box=None, reducing_gap=None) -> Image:
             msg = "reducing_gap must be 1.0 or greater"
             raise ValueError(msg)
 
-        size = tuple(size)
-
         self.load()
         if box is None:
             box = (0, 0) + self.size
-        else:
-            box = tuple(box)
```
lgeiger, line 2220: @srittau @radarhere This change ended up being a breaking change for me when upgrading from 1.3.0 to 1.4.0 since `.resize()` previously would accept a numpy array as size. This now fails with an internal error message:
```python
  File "/usr/local/lib/python3.11/site-packages/PIL/Image.py", line 2297, in resize
    if self.size == size and box == (0, 0) + self.size:
       ^^^^^^^^^^^^^^^^^
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```
If this change was intentional I think it would be best to raise a nice error message or revert this and relax the input type requirements.

radarhere: Let's move this conversation to #8195

radarhere: https://github.com/python-pillow/Pillow/pull/8201 has now restored the ability to use a NumPy array.

PR #8056, opened 2024-05-13
Do not detect Ultra HDR images as MPO

### LINKED ISSUE #8052: Is this MPO file decoded correctly?
### What did you do?

```python3

from PIL import Image

im = Image.open("328384445-d4972d5b-3409-4d5f-a107-ccb8c5dc0177.jpg")  # MPO file
im.show()
```

Correctly decode second image

image

### What actually happened?

image

* OS: macOS/Linux
* Python: 3.10/3.11
* Pillow: 10.2.0 (edited) (was 10.3.0, my mistake)

Test image can be found here: https://github.com/comfyanonymous/ComfyUI/issues/3416#issuecomment-2097406107

radarhere:
Resolves #8052

The Ultra HDR format stores its gain map [as an additional MPO image](https://developer.android.com/media/platform/hdr-image-format#MPF_storage_of_the_gain_map). However, as the issue has found, because we do not support Ultra HDR images, the additional image appears incorrectly.

Instead, this PR treats such files as standard JPEGs.

https://developer.android.com/media/platform/hdr-image-format#signal_of_the_format details how this format can be detected.
> A JPEG file conforming to this format may be identified by the presence of hdrgm:Version="1.0" in the primary image's XMP packet


hugovk:
`Tests/images/ultrahdr.jpg` is about half a megabyte, not our biggest test image (which is 9.9 MB!), but still pretty big.

Can we find a smaller image to include in the test suite? cc @bigcat88

bigcat88:
I am not the author of that image.

Here is the link to the message of author https://github.com/python-pillow/Pillow/issues/8052#issuecomment-2107318083

where he mentions that he does not have the smaller image (

radarhere:
@hugovk would you prefer I create a smaller image artificially?

hugovk:
If it's not too difficult and it doesn't feel too hacky? Otherwise the one already here is probably fine. Just keeping an eye on our 46.6 MB sdist.

radarhere:
I've realised that since we're not loading the image, I can just truncate the file leaving only the data that's read in the initial open. The test image is now down to 48kb.

PR #8057, opened 2024-05-13
FIX: in py3.13 exec does not implicitly update locals()

tacaswell:
Alternative to #8050

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,8 +23,9 @@ from setuptools.command.build_ext import build_ext
 def get_version():
     version_file = "src/PIL/_version.py"
     with open(version_file, encoding="utf-8") as f:
-        exec(compile(f.read(), version_file, "exec"))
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -23,8 +23,9 @@
 def get_version():
     version_file = "src/PIL/_version.py"
     with open(version_file, encoding="utf-8") as f:
-        exec(compile(f.read(), version_file, "exec"))
-    return locals()["__version__"]
+        lcl = {}
+        exec(compile(f.read(), version_file, "exec"), lcl)
```
radarhere, line 27: `lcl` is the second argument to `exec()`. Looking at https://docs.python.org/3.13/library/functions.html#exec, this would be `globals`, right? So maybe `lcl` is a confusing name, because it makes one think that `locals` is being used instead?

PR #8067, opened 2024-05-19
Change AlignAfterOpenBracket in .clang-format to BlockIndent

Yay295:
Currently in Python when a function declaration is split across multiple lines, the closing parenthesis is on the same level as the `def`.

https://github.com/python-pillow/Pillow/blob/c250a44177ac1e82d300aef702d5c2feef15fd03/src/PIL/Image.py#L3484-L3488

However currently in C the closing parenthesis is on the same line as the last argument.

https://github.com/hugovk/Pillow/blob/5597f618a3b5cf8e208dded44a2107313c47122d/src/libImaging/BcnDecode.c#L825-L833

```c
static int
decode_bcn(
    Imaging im,
    ImagingCodecState state,
    const UINT8 *src,
    int bytes,
    int N,
    int C,
    char *pixel_format) {
```

It's the same with function calls as well.

This change changes the C style to match the Python style.


homm (approved):
Solves the issue described in #8212

Issue #5076, opened 2020-12-03
Feature request: implement im.getxmp() to return all embedded XMP meta data as XML


laynr:
Implement **image.getxmp()** similar to image.getexif(), that **returns all embedded XMP meta data out of an image as XML**

Something like:
```python
def getxmp(self):
    for segment, content in self.applist:
        if segment == 'APP1':
            marker, xmp_tags = content.rsplit(b'\x00', 1)
            if marker == b'http://ns.adobe.com/xap/1.0/':
                root = xml.etree.ElementTree.fromstring(xmp_tags)
return root

# Based off https://stackoverflow.com/a/32001778
```

FYI: This didn't work for me:
```xmp_tags = self.info.get("XML:com.adobe.xmp")```

I am sure this feature request has been asked before... but a search of 'XMP' in issues yielded nothing.  Just asking for MVP, not write support, or tag comprehension. 

XMP documentation: 
Official:  https://www.adobe.com/devnet/xmp.html
Helpful: https://exiftool.org/TagNames/XMP.html

Requesting output similar to the output of:
exiftool.exe -xmp:all -X image.jpg

UrielMaD:
Hello @laynr @hugovk I'd like to take this issue, I'm already working on this.

I already got the xmp tags out of the file, I'm just wondering which output structure would be best for the getxmp() to return

I was thinking that it could be an object like the getexif(), but in this one I already got the tags name, so, instead of the tag number I could implement the actual tag name and its value

UrielMaD:
Btw I also could just simply return its xml tree

laynr:
Great thanks @UrielMaD! 

It is probably best to keep it similar to getexif() if they return an object and adding tags name, instead of the tag number would be awesome... 

That said, there is definite value in just returning the XML tree.  For one, returning the XML tree may be the most future proof  as you wouldn't need to stay current on new tags names.

I guess if you return an object, one of the objects functions can be to return the XML tree - perhaps that would be the best of both worlds!  (but more work)

For my personal project I just used the XML tree as the parsers for XML are well established.   

One thing I am noticing is that there can be multiple XMP sections in one file and not necessarily adjacent.   

Thank you for taking this on.  I believe it will be very useful for many people!

UrielMaD:
Thank you @layn, then I'll send a PR implementing xml object, I can also return the whole xml tree as string.

[...]

PR #8069, opened 2024-05-20
Improved consistency of XMP handling

radarhere:
https://github.com/python-pillow/Pillow/issues/5076#issuecomment-2118983437 is a user who would like to retrieve XMP strings from images. This can be done, but [inconsistently between formats](https://github.com/python-pillow/Pillow/issues/5076#issuecomment-2119966091).

This PR updates JPEG, TIFF and PNG to create `im.info["xmp"]` like WEBP does.

Then, given that common location of XMP data, it makes sense to tidy up the code by adding `getxmp()` to the Image class, rather than duplicating it for each of supported formats. `getxmp()` will return an empty dictionary for other formats.


bigcat88:
Since this change removed the `_getxmp` function - which I used in pillow_heif, could you at least add parsing for the NULL terminating byte in xmp? @radarhere 

Info is here:

https://github.com/bigcat88/pillow_heif/issues/69
https://github.com/AOMediaCodec/libavif/issues/1333   - topic is big, but most important are last messages

Because now I need to either remove the leading byte, which will lead to the fact that if the image is opened and saved, the `xmp` will differ in a result by that one byte.

That will allows me to just remove `getxmp` from pillow_heif, otherwise I will need to copy old `_getxmp` code and use it..

radarhere:
Ah, I had presumed I was just removing an unused internal method.

I've created https://github.com/python-pillow/Pillow/pull/8171. Let us know if you have any feedback.

PR #8071, opened 2024-05-21
Added Python 3.13 beta wheels

radarhere:
Requires #8050 or #8057

### DIFF
--- a/.github/workflows/wheels-test.sh
+++ b/.github/workflows/wheels-test.sh
@@ -12,7 +12,7 @@ elif [ "${AUDITWHEEL_POLICY::9}" == "musllinux" ]; then
 else
     yum install -y fribidi
 fi
-if [ "${AUDITWHEEL_POLICY::9}" != "musllinux" ]; then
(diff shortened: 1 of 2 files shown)


hugovk (approved):
Thanks, let's also add to the release notes, something like:

https://pillow.readthedocs.io/en/stable/releasenotes/9.3.0.html#python-3-11-wheels

(Looks like we didn't mention the 3.12 beta wheels.)

radarhere:
You've linked to the release notes for Python 3.11 final wheels - it just starts out saying that the previous release had beta wheels.
Similarly, the release notes for Python 3.12 final wheels are at https://pillow.readthedocs.io/en/stable/releasenotes/10.1.0.html#python-3-12 and mention beta wheels in the previous release.

From what I can see, we've never announced beta wheels in the release they're in.

We can start that tradition if you want though.

hugovk:
So I did :)

Yeah, I think it's worth calling out, it could encourage others to start testing the beta.

PR #8097, opened 2024-06-02
Use DCTDecode and SMask when saving LA and RGBA PDF images

### LINKED ISSUE #8074: RGBA PNG saved as PDF renders incorrectly in some applications
### What did you do?

Convert a PNG file (screenshots created on Gnome and on Windows) to PDF.

The rendered PDF looks correct.

### What actually happened?

The output looks correct in MuPDF and Chromium, but incorrect in Evince and `pdf.js`.

* OS: OpenSUSE Leap 15.4
* Python: 3.9.18
* Pillow: 10.3.0

```text
Pillow 10.3.0
Python 3.9.18 (main, Sep 06 2023, 07:49:32) [GCC]
Python executable is /home/stefan/tmp/venv1/bin/python3
Environment Python files loaded from /home/stefan/tmp/venv1
Python Pillow modules loaded from /home/stefan/tmp/venv1/lib64/python3.9/site-packages/PIL
Binary Pillow modules loaded from /home/stefan/tmp/venv1/lib64/python3.9/site-packages/PIL
--- WEBPMUX support ok
--- WEBP Animation support ok
--- RAQM (Bidirectional Text) support ok, loaded 0.10.1, fribidi 1.0.10, harfbuzz 8.4.0
```

```python
from PIL import Image

Image.open("image2vu2shjb.png").save("out2.pdf")
```

Input:

image2vu2shjb

Output: [out2.pdf](https://github.com/python-pillow/Pillow/files/15401682/out2.pdf)

Rendered output from Evince:

ksnip_20240522-121933

Rendered output from pdf.js:

ksnip_20240522-121949

radarhere:
Resolves #8074

Discussion in the issue revealed that JPXDecode PDFs are not displayed correctly in Firefox. This is the filter that Pillow uses to save [LA](https://github.com/python-pillow/Pillow/pull/7299) and [RGBA](https://github.com/python-pillow/Pillow/pull/6925) images. There is an [open issue about the Firefox problem](https://github.com/mozilla/pdf.js/issues/16782), but in the meantime, there was a [suggestion that we could use SMasks instead](https://github.com/python-pillow/Pillow/issues/8074#issuecomment-2143739314).

### DIFF
--- a/Tests/test_file_pdf.py
+++ b/Tests/test_file_pdf.py
@@ -12,7 +12,7 @@ import pytest
 
 from PIL import Image, PdfParser, features
 
-from .helper import hopper, mark_if_feature_version, skip_unless_feature
+from .helper import hopper, mark_if_feature_version
 
 
 def helper_save_as_pdf(tmp_path: Path, mode: str, **kwargs: Any) -> str:
@@ -41,14 +41,8 @@ def helper_save_as_pdf(tmp_path: Path, mode: str, **kwargs: Any) -> str:
     return outfile
 
 
-@pytest.mark.parametrize("mode", ("L", "P", "RGB", "CMYK"))
+@pytest.mark.parametrize("mode", ("L", "LA", "P", "RGB", "RGBA", "CMYK"))
 def test_save(tmp_path: Path, mode: str) -> None:
-    helper_save_as_pdf(tmp_path, mode)
-
-
-@skip_unless_feature("jpg_2000")
-@pytest.mark.parametrize("mode", ("LA", "RGBA"))
-def test_save_alpha(tmp_path: Path, mode: str) -> None:
     helper_save_as_pdf(tmp_path, mode)
 
 
--- a/docs/handbook/image-file-formats.rst
+++ b/docs/handbook/image-file-formats.rst
@@ -1585,9 +1585,8 @@ files. Different encoding methods are used, depending on the image mode.
 
 * 1 mode images are saved using TIFF encoding, or JPEG encoding if libtiff support is
   unavailable
-* L, RGB and CMYK mode images use JPEG encoding
+* L, LA, RGB, RGBA and CMYK mode images use JPEG encoding
 * P mode images use HEX encoding
-* LA and RGBA mode images use JPEG2000 encoding
 
 .. _pdf-saving:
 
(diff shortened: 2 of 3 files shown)


sl2c:
Thanks for the commit!

I've just briefly glanced over the code and it seems like you're encoding every mode except 'P' as '/DCTDecode', while 'P' is encoded as '/ASCIIHexDecode'. Some comments:

1) why use '/ASCIIHexDecode' for 'P'? If the palette is not random (i.e., if, on the opposite side of this, the palette is monotone — increasing palette index means non-decreasing luminosity) — and it's almost never random in real life — there's quite a bit of correlation between consecutive palette indices, and applying a .DecodeParms.Predictor = 11-15 (see Adobe PDF Ref. v. 1.7 page 76) and then compressing with /FlateDecode would give a much smaller file size. But even without the predictors, usually ZIP compresses this real-life images much better.

2) it makes sense to use '/DCTDecode' only if a) the user asks for it; b) if the image.format is 'JPEG', which would imply that we already have the quantization tables etc. In all other situations the image has to be saved using a lossless filter, e.g. the /FlateDecode + /Predictor combination (see previous paragraph).

3) A single exception to this is when image.mode is '1', for which the best (and lossless) filter to use is /CCITTFaxDecode.

For examples of how all this could be done please take a look at [pdfimage.py](https://github.com/sl2c/pdfrwx/blob/master/pdfimage.py)

Finally, if all this is too complicated the simplest solution to code would be for all image.mode-s to call .tobytest() and then zlib.compress() and put the result into the PDF image dictionary stream, and then add /Filter = /FlateDecode. The bottom logic is that Pillow should just produce the PDF format and then it's up to the user to compress/re-encode it as they wish.

radarhere:
I don't think your comments relate to this PR or the problem it is trying to fix, but are rather comments on existing state of PdfImagePlugin. So I would like to leave this PR as it is.

That said,
1. I think our code is already aware there could be an improvement.
https://github.com/python-pillow/Pillow/blob/eeb1eeab209ac395427d269a7aecfd76d525bdc4/src/PIL/PdfImagePlugin.py#L51-L53

2. You're avocating for lossless saving. However, the natural downside of this would be larger file sizes. I think a good solution would be to [add a 'lossless' argument like WebP has](https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html#webp).

radarhere:
The PDF.js issue has now been closed! https://github.com/mozilla/pdf.js/issues/16782

It will still require another release of PDF.js, and then a release of Firefox to include that, but it is a positive development.

radarhere:
Testing with the latest version of Firefox (129.0.2), it is now rendering the PDF correctly - so the PDF.js fix has made it through.

PR #8108, opened 2024-06-05
Added type hints

radarhere:


src/PIL/WmfImagePlugin.py
```diff
@@ -30,7 +30,7 @@
 _handler = None
 
 
-def register_handler(handler: ImageFile.StubHandler) -> None:
+def register_handler(handler: ImageFile.StubHandler | None) -> None:
```
radarhere, line 33: In the tests, `None` is passed to this from https://github.com/python-pillow/Pillow/blob/6dbedc51aeb431b543469bb1f3fdb51d5e36af3b/Tests/test_file_wmf.py#L53

PR #8110, opened 2024-06-06
Do not preserve EXIFIFD tag by default when saving TIFF images

### LINKED ISSUE #8106: save() produces an unreadable TIFF
`Image.open('in.tif').save('out.tif')` produces `out.tif` which is unreadable. 

`Image.__version__ == '10.1.0'`

[in.tif.zip](https://github.com/user-attachments/files/15568097/in.tif.zip)

radarhere:
Resolves #8106

The EXIFIFD tag is '[a pointer to the Exif IFD](https://www.awaresystems.be/imaging/tiff/tifftags/exififd.html)'. However, there's no reason why the data should be located at that very same position in the image that Pillow creates.

So this PR does not preserve it by default when saving TIFF images. The user may still provide it through an encoder argument.


Tests/test_file_libtiff.py
```diff
@@ -686,12 +686,17 @@ def test_save_ycbcr(self, tmp_path: Path) -> None:
             assert reloaded.tag_v2[532] == (0, 255, 128, 255, 128, 255)
 
     def test_exif_ifd(self, tmp_path: Path) -> None:
```
Yay295, line 688: The `tmp_path` parameter can be removed now.

src/PIL/TiffImagePlugin.py
```diff
@@ -1819,8 +1819,9 @@ def _save(im, fp, filename):
         # SAMPLEFORMAT is determined by the image format and should not be copied
         # from legacy_ifd.
```
Yay295, line 1820: Maybe add EXIFIFD to this comment?

Tests/test_file_libtiff.py
radarhere: I've updated the commit.

src/PIL/TiffImagePlugin.py
radarhere: I've updated the commit.

PR #8114, opened 2024-06-08
Allow float center for rotate operations to resolve FIXME

radarhere:
In `im.rotate()`, there is
https://github.com/python-pillow/Pillow/blob/5bacce9dc2c0db25822bbddadcaa07496a9ffe1b/src/PIL/Image.py#L2375-L2379

Looking through the code after that, there is no reason why they can't be floats - the numbers are simply combined with other floats.

https://github.com/python-pillow/Pillow/blob/5bacce9dc2c0db25822bbddadcaa07496a9ffe1b/src/PIL/Image.py#L2381-L2399

This also allows the rotation center to be the exact center of an image with odd dimensions - although that also sound like a good argument why the user might want to pass in floats as the `center`, so I've updated the type hint accordingly.

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -2305,7 +2305,7 @@ class Image:
         angle: float,
         resample: Resampling = Resampling.NEAREST,
         expand: int | bool = False,
-        center: tuple[int, int] | None = None,
+        center: tuple[float, float] | None = None,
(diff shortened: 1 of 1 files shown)


hugovk (approved):
Only a seven-year-old FIXME :)

PR #8117, opened 2024-06-08
Added type hints

radarhere:


src/PIL/GimpGradientFile.py
```diff
@@ -64,9 +65,22 @@ def sphere_decreasing(middle, pos):
...
+                float,
+                float,
+                float,
+                list[float],
+                list[float],
+                Callable[[float, float], float],
+            ]
+        ]
+        | None
+    ) = None
```
hugovk, line 80: We also have this long type def in `GimpGradientFile`, is it worth assigning it to something and re-using it?

radarhere: I've realised we can just re-use the other variable. I've pushed a commit.

PR #8118, opened 2024-06-09

radarhere:


Tests/test_image_thumbnail.py
```diff
@@ -16,7 +16,7 @@
 
 def test_sanity() -> None:
     im = hopper()
-    assert im.thumbnail((100, 100)) is None
+    assert im.thumbnail((100, 100)) is None  # type: ignore[func-returns-value]
```
radarhere, line 19: Without this ignore, mypy reports

> Tests/test_image_thumbnail.py:19: error: "thumbnail" of "Image" does not return a value (it only ever returns None)  [func-returns-value]
>         assert im.thumbnail((100, 100)) is None
>                ^~~~~~~~~~~~~~~~~~~~~~~~

Yay295: Couldn't this line just be `im.thumbnail((100, 100))`, without the assert?

radarhere, line 19:
```suggestion
    im.thumbnail((100, 100))
```

radarhere: Ok, I’ve pushed a commit.

PR #8125, opened 2024-06-10
Added type hints

radarhere:


src/PIL/IcnsImagePlugin.py
```diff
@@ -346,29 +347,27 @@ def _save(im, fp, filename):
     entries = []
     for type, size in sizes.items():
         stream = size_streams[size]
-        entries.append(
-            {"type": type, "size": HEADERSIZE + len(stream), "stream": stream}
-        )
+        entries.append((type, HEADERSIZE + len(stream), stream))
```
hugovk, line 350: What triggered this change?

src/PIL/ImImagePlugin.py
```diff
@@ -341,6 +341,8 @@ def _save(im: Image.Image, fp: IO[bytes], filename: str) -> None:
         # or: SyntaxError("not an IM file")
         # 8 characters are used for "Name: " and "\r\n"
         # Keep just the filename, ditch the potentially overlong path
+        if isinstance(filename, bytes):
+            filename = filename.decode("ascii")
```
hugovk, line 345: Is this also fixing a bug?

radarhere: In main at the moment, save handlers can receive strings or bytes as the filename.
https://github.com/python-pillow/Pillow/blob/53e82e49c548420cc84a1ce233096f5b06d2cf37/src/PIL/Image.py#L2454
https://github.com/python-pillow/Pillow/blob/53e82e49c548420cc84a1ce233096f5b06d2cf37/src/PIL/Image.py#L231-L232

Part of this PR is propagating that out to some plugins that were only using strings.

src/PIL/IcnsImagePlugin.py
radarhere: Once i type hinted the `_save` method in this plugin, mypy reported
```
src/PIL/IcnsImagePlugin.py:358: error: Generator has incompatible item type "object"; expected "bool"  [misc]
        file_length += sum(entry["size"] for entry in entries)
                           ^~~~~~~~~~~~~
src/PIL/IcnsImagePlugin.py:365: error: No overload variant of "write" of "IO" matches argument type "object"  [call-overload]
            fp.write(entry["type"])
            ^~~~~~~~~~~~~~~~~~~~~~~
src/PIL/IcnsImagePlugin.py:365: note: Possible overload variants:
src/PIL/IcnsImagePlugin.py:365: note:     def write(self, Buffer, /) -> int
src/PIL/IcnsImagePlugin.py:365: note:     def write(self, bytes, /) -> int
src/PIL/IcnsImagePlugin.py:370: error: No overload variant of "write" of "IO" matches argument type "object"  [call-overload]
            fp.write(entry["type"])
            ^~~~~~~~~~~~~~~~~~~~~~~
src/PIL/IcnsImagePlugin.py:370: note: Possible overload variants:
src/PIL/IcnsImagePlugin.py:370: note:     def write(self, Buffer, /) -> int
src/PIL/IcnsImagePlugin.py:370: note:     def write(self, bytes, /) -> int
src/PIL/IcnsImagePlugin.py:372: error: No overload variant of "write" of "IO" matches argument type "object"  [call-overload]
            fp.write(entry["stream"])
            ^~~~~~~~~~~~~~~~~~~~~~~~~
src/PIL/IcnsImagePlugin.py:372: note: Possible overload variants:
src/PIL/IcnsImagePlugin.py:372: note:     def write(self, Buffer, /) -> int
src/PIL/IcnsImagePlugin.py:372: note:     def write(self, bytes, /) -> int
Found 4 errors in 1 file (checked 1 source file)
```

This change fixes it. I can use a named tuple or something if you prefer.

hugovk: Thanks. Named tuples are nice, but this is very localised so I think we're fine with a basis tuple here.

hugovk (approved):

PR #8132, opened 2024-06-11
Added type hints

radarhere:


src/PIL/PdfParser.py
```diff
@@ -404,9 +404,8 @@ def __init__(self, filename=None, f=None, buf=None, start_offset=0, mode="rb"):
     def __enter__(self) -> PdfParser:
         return self
 
-    def __exit__(self, exc_type, exc_value, traceback):
+    def __exit__(self, *args: object) -> None:
         self.close()
-        return False  # do not suppress exceptions
```
radarhere, line 409: https://python.readthedocs.io/en/latest/reference/datamodel.html#object.__exit__
> If an exception is supplied, and the method wishes to suppress the exception (i.e., prevent it from being propagated), it should return a true value. Otherwise, the exception will be processed normally upon exit from this method.

So `False` and `None` will have the same result.

PR #8135, opened 2024-06-12
Improve `ImageFont.freetype` support for XDG directories on Linux

mamg22:

For `ImageFont.freetype`:

* Add support for searching fonts in the user data directory indicated by `XDG_DATA_HOME` or its default, before searching in system-wide directories.
* Update the default value for `XDG_DATA_DIRS` to include the directory for system-local fonts in `/usr/local/share/fonts`.

These changes allow searching for both user-installed fonts, and fonts that aren't managed by the system package manager, such as when Microsoft fonts are installed by `mscorefonts-installer` into `/usr/local`.

Reference: [version 0.8 of the freedesktop base directory spec](https://specifications.freedesktop.org/basedir-spec/0.8/)

### DIFF
--- a/Tests/test_imagefont.py
+++ b/Tests/test_imagefont.py
@@ -564,7 +564,7 @@ def test_find_font(
     # catching syntax like errors
     monkeypatch.setattr(sys, "platform", platform)
     if platform == "linux":
-        monkeypatch.setenv("XDG_DATA_HOME", "/home/__pillow__/.local/share")
+        monkeypatch.setenv("XDG_DATA_HOME", os.path.expanduser("~/.local/share"))
         monkeypatch.setenv("XDG_DATA_DIRS", "/usr/share/:/usr/local/share/")
 
     def fake_walker(path: str) -> list[tuple[str, list[str], list[str]]]:
--- a/Tests/test_imagefont.py
+++ b/Tests/test_imagefont.py
@@ -564,6 +564,7 @@ def test_find_font(
     # catching syntax like errors
     monkeypatch.setattr(sys, "platform", platform)
     if platform == "linux":
+        monkeypatch.setenv("XDG_DATA_HOME", "/home/__pillow__/.local/share")
         monkeypatch.setenv("XDG_DATA_DIRS", "/usr/share/:/usr/local/share/")
 
     def fake_walker(path: str) -> list[tuple[str, list[str], list[str]]]:
--- a/docs/releasenotes/10.4.0.rst
+++ b/docs/releasenotes/10.4.0.rst
@@ -46,7 +46,7 @@ ImageFont.truetype
 ^^^^^^^^^^^^^^^^^^
 
 :py:func:`~PIL.ImageFont.truetype` now searches for fonts in the directory indicated
-by the ``XDG_DATA_HOME`` envirnorment variable in addition to those specified in
+by the ``XDG_DATA_HOME`` environment variable in addition to those specified in
 ``XDG_DATA_DIRS``.  Additionally, the default for ``XDG_DATA_DIRS`` has been updated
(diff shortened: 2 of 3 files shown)


docs/releasenotes/10.4.0.rst
```diff
@@ -42,6 +42,14 @@ The ``hints`` parameter in :py:meth:`~PIL.ImageDraw.getdraw()` has been deprecat
 API Changes
 ===========
 
+ImageFont.truetype
+^^^^^^^^^^^^^^^^^^
+
+:py:function:`~PIL.ImageFont.truetype` now searches for fonts in the directory
```
radarhere, line 48:
```suggestion
:py:func:`~PIL.ImageFont.truetype` now searches for fonts in the directory
```
You can see at https://pillow--8135.org.readthedocs.build/en/8135/releasenotes/10.4.0.html#imagefont-truetype that this isn't rendering correctly as is.

mamg22: Oh thanks, didn't notice that. Let me commit that change.

docs/releasenotes/10.4.0.rst
```diff
@@ -42,6 +42,14 @@ The ``hints`` parameter in :py:meth:`~PIL.ImageDraw.getdraw()` has been deprecat
 API Changes
 ===========
 
+ImageFont.truetype
+^^^^^^^^^^^^^^^^^^
+
+:py:func:`~PIL.ImageFont.truetype` now searches for fonts in the directory indicated
+by the ``XDG_DATA_HOME`` envirnorment variable in addition to those specified in
```
radarhere, line 49:
```suggestion
by the ``XDG_DATA_HOME`` environment variable in addition to those specified in
```

mamg22: 😅

Tests/test_imagefont.py
```diff
@@ -564,6 +564,7 @@ def loadable_font(
     # catching syntax like errors
     monkeypatch.setattr(sys, "platform", platform)
     if platform == "linux":
+        monkeypatch.setenv("XDG_DATA_HOME", "/home/__pillow__/.local/share")
```
radarhere, line 567:
```suggestion
        monkeypatch.setenv("XDG_DATA_HOME", os.path.expanduser("~/.local/share"))
```
This seems better to me.

docs/releasenotes/10.4.0.rst
```diff
@@ -42,6 +42,14 @@ The ``hints`` parameter in :py:meth:`~PIL.ImageDraw.getdraw()` has been deprecat
 API Changes
 ===========
 
+ImageFont.truetype
+^^^^^^^^^^^^^^^^^^
+
+:py:func:`~PIL.ImageFont.truetype` now searches for fonts in the directory indicated
+by the ``XDG_DATA_HOME`` environment variable in addition to those specified in
+``XDG_DATA_DIRS``.  Additionally, the default for ``XDG_DATA_DIRS`` has been updated
+to match the freedesktop spec.
```
radarhere, line 51: I think an API change is when a new parameter is added, or a new error is raised - something that the user should be conscious of when thinking about the inputs and outputs of the code they're writing. I don't think this change qualifies.

I'm not actually not convinced it needs to be in the release notes at all - but if you think it should be, I'd suggest putting it in 'Other Changes'

mamg22: Looking back at older release notes, yeah, it doesn't seem to fit inside the "API changes" section, which matches your definition. It could fit under "Other changes", but since the Linux font finding was undocumented behavior, I think its fine to not include it. I'll remove it later when I make the rest of changes.

radarhere (approved):

PR #8144, opened 2024-06-16
Prevent extra EPS header validations

Yay295:
This is the only return statement in this function, and it appears to be unnecessary since breaking from the loop at this location accomplishes the same thing, and matches what the rest of this loop does.

The only change is that `check_required_header_comments()` is called at the end of the function. This was being skipped before due to the early return, but it will be called now that it's just breaking from the loop.

Before 8f75cc2bbf9cb274a7167f895ab0d437ddca1c07 this return was inside a nested loop, so breaking wouldn't have worked, but now it does. Also, `check_required_header_comments()` didn't exist at that time, so the early return wasn't skipping anything.

### DIFF
--- a/src/PIL/EpsImagePlugin.py
+++ b/src/PIL/EpsImagePlugin.py
@@ -231,9 +231,9 @@ class EpsImageFile(ImageFile.ImageFile):
 
         def check_required_header_comments() -> None:
             """
-            The EPS spec requires that some headers exist.
-            This should be checked after all headers have been read,
-            or at the end of the file if that comes first.
+            The EPS specification requires that some headers exist.
+            This should be checked when the header comments formally end,
+            when image data starts, or when the file ends, whichever comes first.
             """
             if "PS-Adobe" not in self.info:
                 msg = 'EPS header missing "%!PS-Adobe" comment'
--- a/src/PIL/EpsImagePlugin.py
+++ b/src/PIL/EpsImagePlugin.py
@@ -230,6 +230,11 @@ class EpsImageFile(ImageFile.ImageFile):
         trailer_reached = False
 
         def check_required_header_comments() -> None:
+            """
+            The EPS spec requires that some headers exist.
+            This should be checked after all headers have been read,
+            or at the end of the file if that comes first.
+            """
             if "PS-Adobe" not in self.info:
                 msg = 'EPS header missing "%!PS-Adobe" comment'
                 raise SyntaxError(msg)
@@ -270,6 +275,8 @@ class EpsImageFile(ImageFile.ImageFile):
             if byte == b"":
                 # if we didn't read a byte we must be at the end of the file
                 if bytes_read == 0:
+                    if reading_header_comments:
+                        check_required_header_comments()
                     break
             elif byte in b"\r\n":
                 # if we read a line ending character, ignore it and parse what
@@ -364,8 +371,6 @@ class EpsImageFile(ImageFile.ImageFile):
             elif bytes_mv[:9] == b"%%Trailer":
                 trailer_reached = True
             bytes_read = 0
-
-        check_required_header_comments()
 
         if not self._size:
             msg = "cannot determine EPS bounding box"


radarhere:
> The only change is that `check_required_header_comments()` is called at the end of the function. This was being skipped before due to the early return, but it will be called now that it's just breaking from the loop.

It's already been called by this point.

`reading_header_comments` starts out as true.
https://github.com/python-pillow/Pillow/blob/6a6ad65b473c18971aa649493ae7ac44db5f2c3d/src/PIL/EpsImagePlugin.py#L228

The image data part of the code can't be reached until `reading_header_comments` is false
https://github.com/python-pillow/Pillow/blob/6a6ad65b473c18971aa649493ae7ac44db5f2c3d/src/PIL/EpsImagePlugin.py#L300
https://github.com/python-pillow/Pillow/blob/6a6ad65b473c18971aa649493ae7ac44db5f2c3d/src/PIL/EpsImagePlugin.py#L327

But it is only set to false after `check_required_header_comments()` is called.
https://github.com/python-pillow/Pillow/blob/6a6ad65b473c18971aa649493ae7ac44db5f2c3d/src/PIL/EpsImagePlugin.py#L291-L292
https://github.com/python-pillow/Pillow/blob/6a6ad65b473c18971aa649493ae7ac44db5f2c3d/src/PIL/EpsImagePlugin.py#L307-L308

If anyone is looking at this and thinks 'What about the size check?'
https://github.com/python-pillow/Pillow/blob/6a6ad65b473c18971aa649493ae7ac44db5f2c3d/src/PIL/EpsImagePlugin.py#L370-L372

That has to be set by the end of the image data part of the code.
https://github.com/python-pillow/Pillow/blob/6a6ad65b473c18971aa649493ae7ac44db5f2c3d/src/PIL/EpsImagePlugin.py#L356

So the change suggested here doesn't make any functional difference.

Yay295:
So actually, if `check_required_header_comments()` was added at line 273, I think it could be removed from the end.

radarhere:
Ok, you're describing

```diff
diff --git a/src/PIL/EpsImagePlugin.py b/src/PIL/EpsImagePlugin.py
index 380b1cf0e..20ef0e5e2 100644
--- a/src/PIL/EpsImagePlugin.py
+++ b/src/PIL/EpsImagePlugin.py
@@ -270,6 +270,7 @@ class EpsImageFile(ImageFile.ImageFile):
             if byte == b"":
                 # if we didn't read a byte we must be at the end of the file
                 if bytes_read == 0:
+                    check_required_header_comments()
                     break
             elif byte in b"\r\n":
                 # if we read a line ending character, ignore it and parse what
@@ -365,8 +366,6 @@ class EpsImageFile(ImageFile.ImageFile):
                 trailer_reached = True
             bytes_read = 0
 
-        check_required_header_comments()
-
         if not self._size:
             msg = "cannot determine EPS bounding box"
             raise OSError(msg)
```

Sure, that looks correct, if you like it as a fractional performance increase.

radarhere:
I think your first commit could be removed? As I was saying, it doesn't make any functional difference. It just unnecessarily runs the check(s) at the end.

Yay295:
I think it makes the function easier to read by only having `continue`s and `break`s, and not having one early return.

radarhere:
I don't personally agree, but we can see what others think.

wiredfool:
I'm generally a fan of early exit rather than introducing conditional checks to allow a method to run to completion when there shouldn't be code executed anyway.

Yay295:
I would normally agree, but in this case this is the only return in the method, and it's inside a loop, so it's easy to miss it, and this change also stops the check from happening twice in some situations.

radarhere:
> this change also stops the check from happening twice in some situations.

As per [my earlier comment](https://github.com/python-pillow/Pillow/pull/8144#issuecomment-2172059620), the `return` change doesn't have any functional impact. Your second commit on the other hand, does stop `check_required_header_comments()` from potentially being called twice.

Yay295:
I still think it's bad form to have that return there, but I've removed the first commit and rebased on main.

src/PIL/EpsImagePlugin.py
```diff
@@ -230,6 +230,11 @@ def _open(self) -> None:
         trailer_reached = False
 
         def check_required_header_comments() -> None:
+            """
+            The EPS spec requires that some headers exist.
```
radarhere, line 234:
```suggestion
            The EPS specification requires that some headers exist.
```
Nitpick suggestion: Using the full word is clearer.

src/PIL/EpsImagePlugin.py
```diff
@@ -230,6 +230,11 @@ def _open(self) -> None:
...
+            This should be checked after all headers have been read,
+            or at the end of the file if that comes first.
```
radarhere, line 236:
```suggestion
            This should be checked when the header comments formally end, when image
            data starts or when the file ends, whichever comes first.
```
The previous wording of this is slightly odd to me, because aren't all headers read once you get to the end of the file?

radarhere (approved):

PR #8146, opened 2024-06-17
Raise SyntaxError specific to JP2

Yay295:
With this change the `_parse_codestream` and `_parse_jp2_header` methods never return `None` for the size or mode.


src/PIL/Jpeg2KImagePlugin.py
```diff
@@ -122,7 +122,8 @@ def _parse_codestream(fp):
     elif csiz == 4:
         mode = "RGBA"
     else:
-        mode = None
+        msg = "unable to determine JP2 image mode"
```
radarhere, line 125:
```suggestion
        msg = "unable to determine J2K image mode"
```
https://github.com/python-pillow/Pillow/blob/99dd55324dcddc378341ca1dae75649c308bae43/src/PIL/Jpeg2KImagePlugin.py#L229-L230
might raise
https://github.com/python-pillow/Pillow/blob/99dd55324dcddc378341ca1dae75649c308bae43/src/PIL/Jpeg2KImagePlugin.py#L206-L207
So I think `_parse_codestream()` should use its codec name in an error.
https://github.com/python-pillow/Pillow/blob/99dd55324dcddc378341ca1dae75649c308bae43/src/PIL/Jpeg2KImagePlugin.py#L223-L224

radarhere:
I don't see any problems with this change, as it is mirroring the existing pattern of raising an error from inside `_parse_jp2_header`.

Your motivation here is just to let the behaviour of these two methods match each other more closely, as a code cleanup task, yes?

Yay295:
Yes

radarhere (approved):

PR #8154, opened 2024-06-20
Hack202406

### LINKED ISSUE #1888: Tracking Issue for high bit depth multichannel images

aclark4life:
Fixes #1888 .

 * Add support for high depth multi-channel images


src/PIL/ImageFile.py
```diff
@@ -117,6 +117,8 @@ def __init__(self, fp=None, filename=None):
 
         self.readonly = 1  # until we know better
 
+        self.newconfig = ()
```
radarhere, line 120: Could we have a more descriptive name for this?

PR #8157, opened 2024-06-20
Add support for high bit depth multichannel images

### LINKED ISSUE #1888: Tracking Issue for high bit depth multichannel images

aclark4life:
Fixes #1888 .

 * Add support for high bit depth multichannel images


hugovk:
Please can you rename the PR to something more descriptive?

PIL/TiffImagePlugin.py
hugovk, line 1: Is this an old file added by accident? 

The source is now under src/PIL/ not PIL/, and this file has handling for Python 2.6.

yoursunny:
My contribution from #8161 has been reverted 🫥

hugovk:
@yoursunny Would you prefer to open this as a PR yourself directly?

aclark4life:
@yoursunny @radarhere @hugovk OK sorry about that, I see that we can grab that contribution from [here](https://github.com/python-pillow/Pillow/tree/6fb6106a8fa467a107540aad8d36a040a1826aad) so let's try again with my apologies for the extra work. I'm going to close this PR, create a new PR from @yoursunny 's [hack branch](https://github.com/python-pillow/Pillow/tree/hack202406) and regroup there. I'm also going to remove the branch I created. Maybe we can commit directly to the PR in the future. 🤔

PR #8158, opened 2024-06-21
Rename 16bit RGB rawmodes

### LINKED ISSUE #8021: Some of the unpackers are misnamed
RGB15 actually reads data as XBGR (XBBBBBGGGGGRRRRR).

https://github.com/python-pillow/Pillow/blob/a8f434f67657d9286bafb0f132ac608276fa96c0/src/libImaging/Unpack.c#L661-L674

RGBA15 actually reads data as ABGR (ABBBBBGGGGGRRRRR).

https://github.com/python-pillow/Pillow/blob/a8f434f67657d9286bafb0f132ac608276fa96c0/src/libImaging/Unpack.c#L676-L689

RGBA4B actually reads data as ABGR (AAAABBBBGGGGRRRR).

https://github.com/python-pillow/Pillow/blob/a8f434f67657d9286bafb0f132ac608276fa96c0/src/libImaging/Unpack.c#L766-L779

etc.

Basically, all of the unpackers that read less than 8 bits per band appear to be backwards.

Yay295:
Fixes #8021. Supersedes #7965.
This is based on #8026, but that pull request has been open for nearly two months now without any activity, and the next version of Pillow is coming soon. I'd like for this change to go out with the #7978 changes so that the deprecation timelines match.

RGB;15 → XBGR;1555
RGB;16 → BGR;565
BGR;5 → XRGB;1555
BGR;15 → XRGB;1555
BGR;16 → RGB;565
RGB;4B → XBGR;4
RGBA;4B → ABGR;4
RGBA;15 → ARGB;1555
BGRA;15 → ABGR;1555

Packers were added for XRGB;1555, RGB;565, XBGR;1555, and BGR;565. Previously it was possible to use the BGR;15 and BGR;16 modes for XRGB;1555 and RGB;565 data, but since those modes are being deprecated, these new packers are needed to keep the ability to write that data.

The unpackers were rewritten so that their names match the new rawmodes, and to reduce the usage of multiplication and division for performance reasons.


src/PIL/TgaImagePlugin.py
```diff
@@ -118,8 +118,9 @@ def _open(self) -> None:
             start, size, mapdepth = i16(s, 3), i16(s, 5), s[7]
             if mapdepth == 16:
                 self.palette = ImagePalette.raw(
-                    "BGR;15", b"\0" * 2 * start + self.fp.read(2 * size)
+                    "ARGB;1555", b"\0" * 2 * start + self.fp.read(2 * size)
                 )
+                self.palette.mode = "RGBA"
```
radarhere, line 123:
```suggestion
```
Without #7966, this line makes the palette mode RGBA only until the image is loaded, after which it becomes RGB.

Removing the line makes it consistently RGB. I don't know if it's better to consistently wrong or inconsistently wrong though.

Yay295: If #7966 is merged I can rebase this on that. Otherwise if this is merged first this change would need to be made in #7966.

radarhere: I think I meant the opposite what you're saying - if #7966 is merged, this line is helpful. If #7966 is not merged, then I think this line should be removed.

radarhere:
The test suite is printing a number of deprecation warnings here - https://github.com/python-pillow/Pillow/actions/runs/9616954928/job/26527638175?pr=8158#step:10:5144

For comparison, see main where deprecation warnings are not printed by the test suite - https://github.com/python-pillow/Pillow/actions/runs/9622876608/job/26544673674#step:10:4868

Yay295:
Those warnings should be gone now.

src/libImaging/Unpack.c
```diff
@@ -1531,16 +1666,24 @@ static struct {
...
-    /* exception: rawmodes "I" and "F" are always native endian byte order */
+    // The rawmode syntax is "<mode>;<bits><flags>".
+    // "bits" defaults depending on mode (1 for "1", 8 for "P" and "L", etc.).
+    // "flags" should be given in alphabetical order.
+    // If both bits and flags have their default values, the ; should be left out.
+
+    // Flags:
+    // "B" big endian byte order (default is little endian)
+    // "F" floating point
+    // "I" inverted data
```
radarhere, line 1677:
```suggestion
    // "I" inverted data
    // "F" floating point
```
With this minor change, the flags become alphabetical

src/PIL/TgaImagePlugin.py
hugovk: #7966 is now merged.

radarhere: Thanks

radarhere: This TGA change actually isn't quite complete. I've left a comment in https://github.com/python-pillow/Pillow/pull/7965

Yay295: Since #7966 has been merged, and this has been rebased on that, this change is no longer necessary. However, it was noticed in #7965 that the TGA alpha channel is actually inverted, so "ARGB;1555" isn't quite correct here.

radarhere:
> Packers were added for XRGB;1555, RGB;565, XBGR;1555, and BGR;565. Previously it was possible to use the BGR;15 and BGR;16 modes for XRGB;1555 and RGB;565 data, but since those modes are being deprecated, these new packers are needed to keep the ability to write that data.

If I understand correctly, you're saying that at the moment, you can pack BGR;15 to BGR;15 and BGR;16 to BGR;16, but because those modes are deprecated, you'd like to add the ability to pack from RGB and RGBX to those rawmodes?

The modes are being deprecated because [they aren't used](https://github.com/python-pillow/Pillow/issues/7970#issuecomment-2053589376). I don't think there is any need to try and retain their functionality.

radarhere:
> to reduce the usage of multiplication and division for performance reasons.

Could you explain this? I'm not seeing anywhere that you've rewritten how a mode is packed or unpacked, so I don't understand how performance is affected.

radarhere:
The following code doesn't raise a deprecation warning.
```python
>>> from PIL import Image
>>> im = Image.new("P", (1, 1))
>>> im.putpalette(b"0", "BGR;15")
```

Yay295:
```c
void
ImagingUnpackRGB4B(UINT8 *out, const UINT8 *in, int pixels) {
    int i, pixel;
    /* RGB, 4 bits per pixel, little-endian */
    for (i = 0; i < pixels; i++) {
        pixel = in[0] + (in[1] << 8);
        out[R] = (pixel & 15) * 17;
        out[G] = ((pixel >> 4) & 15) * 17;
        out[B] = ((pixel >> 8) & 15) * 17;
        out[A] = 255;
        out += 4;
        in += 2;
    }
}
static void
ImagingUnpackXBGR4(UINT8 *out, const UINT8 *in, const int pixels) {
    /* XBGR, 4 bits per pixel, little-endian */
    for (int i = 0; i < pixels; i++) {
        const UINT8 b = in[1] & 0x0F;
        const UINT8 g = in[0] & 0xF0;
        const UINT8 r = in[0] & 0x0F;
        out[R] = (r << 4) | r;
        out[G] = g | (g >> 4);
        out[B] = (b << 4) | b;
        out[A] = 255;
        out += 4;
        in += 2;
    }
}
```

The existing function multiplies each value by 17 (17 == 255 // 15) to scale from 4 to 8 bits, while the new function uses a shift and an or. The math looks like this in Python:
```python
bytes4 = list(range(2**4))
bytes4to8accurate = [round(x * 255 / 15) for x in bytes4]
bytes4to8current = [x * 255 // 15 for x in bytes4]
bytes4to8new = [x << 4 | x for x in bytes4]
```

docs/releasenotes/10.4.0.rst
```diff
@@ -30,6 +30,22 @@ The use in :py:mod:`.ImageCms` of input modes and output modes that are not Pill
 image modes has been deprecated. Defaulting to "L" or "1" if the mode cannot be mapped
 is also deprecated.
 
+16-Bit RGB/BGR Rawmodes
+^^^^^^^^^^^^^^^^^^^^^^^
+
+The following rawmodes have been deprecated and replaced with better-named rawmodes.
+
+RGB;15 ? XBGR;1555
```
radarhere, line 38: You originally had an arrow here. I don't think '?' is the best choice, and I'm guessing it wasn't intentional.

src/PIL/TgaImagePlugin.py
radarhere: https://github.com/python-pillow/Pillow/pull/7965 has now been merged.

docs/releasenotes/10.4.0.rst
Yay295: Fixed. It looks like the file got converted to ASCII when I rebased it for some reason.

radarhere: Actually, would you mind copying the 'Deprecated' / 'Use instead' structure that we have already used elsewhere? https://pillow.readthedocs.io/en/stable/deprecations.html#imagecms-constants-and-versions-function

Yay295: done

docs/deprecations.rst
```diff
@@ -116,6 +116,28 @@ The use in :py:mod:`.ImageCms` of input modes and output modes that are not Pill
...
 
+16-Bit RGB/BGR Rawmodes
+^^^^^^^^^^^^^^^^^^^^^^^
+
+.. deprecated:: 10.4.0
+
+The following rawmodes have been deprecated and replaced with better-named rawmodes.
+
+============  ==============
+Removed       Use instead
```
radarhere, line 127: They're not 'Removed' yet, they're 'Deprecated' - see https://pillow.readthedocs.io/en/stable/deprecations.html#imagecms-constants-and-versions-function

radarhere:
> to reduce the usage of multiplication and division for performance reasons.

> The existing function multiplies each value by 17 (17 == 255 // 15) to scale from 4 to 8 bits, while the new function uses a shift and an or.

Oh, I had expected that you'd rewritten the existing functions.

Taking a look,
```python
from PIL import Image
rawmodes = {
  "RGB;15": "XBGR;1555",
  "RGB;16": "BGR;565",
  "BGR;5": "XRGB;1555",
  "BGR;15": "XRGB;1555",
  "BGR;16": "RGB;565",
  "RGB;4B": "XBGR;4",
  "RGBA;4B": "ABGR;4",
  "RGBA;15": "ABGR;1555",
  "BGRA;15": "ARGB;1555",
  "BGRA;15Z": "ARGB;1555Z",
}
for old, new in rawmodes.items():
    mode = "RGBA" if "A" in old else "RGB"
    data = bytes()
    for i in range(65536):
        data += (i).to_bytes(2, 'big')
    im_old = Image.frombytes(mode, (65536, 1), data, "raw", old, 0, 1)
    im_new = Image.frombytes(mode, (65536, 1), data, "raw", new, 0, 1)
    if im_old.tobytes() != im_new.tobytes():
        print("Differences found:", old, new)
    else:
        print("Identical:", old, new)
```
I see
```
Differences found: RGB;15 XBGR;1555
Differences found: RGB;16 BGR;565
Differences found: BGR;5 XRGB;1555
Differences found: BGR;15 XRGB;1555
Differences found: BGR;16 RGB;565
Identical: RGB;4B XBGR;4
Identical: RGBA;4B ABGR;4
Differences found: RGBA;15 ABGR;1555
Differences found: BGRA;15 ARGB;1555
Differences found: BGRA;15Z ARGB;1555Z
```

The ones that are identical should be able to immediately replace the existing functions - https://github.com/Yay295/Pillow/pull/24

However, as for the replacements that produce different output, that hasn't been mentioned so far. Do you think the new output is more or less accurate?

Yay295:
I did mention it in the description of one of the commits, but I also have some Python to show the difference. For going from 5 bits to 8 bits:
```python
bytes5 = list(range(2**5))
bytes5to8accurate = [round(x * 255 / 31) for x in bytes5]
bytes5to8current = [x * 255 // 31 for x in bytes5]
bytes5to8fast = [x << 3 | x >> 2 for x in bytes5]
```
The current method differs from the accurate rounding for 15 numbers, while the new method differs for only 4 numbers. For each number the difference is only 1.

For going from 6 to 8 bits the current method differs from the accurate rounding for 30 numbers, while the new method differs for only 10 numbers. Again for each number the difference is only 1.
```python
bytes6 = list(range(2**6))
bytes6to8accurate = [round(x * 255 / 63) for x in bytes6]
bytes6to8current = [x * 255 // 63 for x in bytes6]
bytes6to8fast = [x << 2 | x >> 4 for x in bytes6]
```

Another benefit is that the new methods are roundtrippable. Currently going from 5 to 8 to 5 bits (or 6 to 8 to 6) will not produce the original values for most numbers.

radarhere:
Could you mention in the documentation that the output will be different?

Yay295:
I could have just committed that instead of rebasing, but I've been doing so much rebasing today I didn't think of it...

PR #8164, opened 2024-06-23
Test improvements

sofijazolotarev:


radarhere:
Hi. How can we help you? I don't think you're actually suggesting we change 663 files. Do you just want assistance to understand why https://github.com/jovanovicisidora/Pillow-SEP/tree/Sofija is failing?

Issue #7273, opened 2023-07-10
getxmp() not returning any values for file containing XMP


duebbert:
Attached is an example file with XMP data.

Exiftool shows the XMP data correctly:
```shell
# exiftool -v test.jpg
[...]
JPEG APP1 (1132 bytes):
  + [XMP directory, 1103 bytes]
  | XMPToolkit = RICOH THETA SC Ver1.01
  | ProjectionType = equirectangular
  | UsePanoramaViewer = True
  | CroppedAreaImageWidthPixels = 5376
  | CroppedAreaImageHeightPixels = 2688
  | FullPanoWidthPixels = 5376
  | FullPanoHeightPixels = 2688
  | CroppedAreaLeftPixels = 0
  | CroppedAreaTopPixels = 0
  | PosePitchDegrees = 2.8
  | PoseRollDegrees = -1.3
[...]
```

But when opening it with Pillow (defusedxml is installed), it returns no XMP Data:
```python
>>> from PIL import Image
>>> img = Image.open("test.jpg")
>>> img.getxmp()
{}
>>> 
```

### What did you expect to happen?

It should return the XMP data of the file.

### What actually happened?

It returned an empty dict.

Digging into the code, I believe the problem is that the "APP1" segment seems to be padded with several `b"\x00"`  so the `getxmp()` function doesn't find the marker due to rsplit being used. But there might be other issues:

```python
>>> img = Image.open("test.jpg")
>>> for segment, content in img.applist:
...     if segment == "APP1" and b"ns.adobe.com" in content:
...             print(content)
... 
b'http://ns.adobe.com/xap/1.0/\x00<?xpacket begin="\xef\xbb\xbf" id="W5M0MpCehiHzreSzNTczkc9d"?>\n<x:xmpmeta 
xmlns:x="adobe:ns:meta/" xmptk="RICOH THETA SC Ver1.01">\n  <rdf:RDF xmlns:rdf="http://www.w3.org/1999/02/22-rdf-
syntax-ns#">\n    <rdf:Description rdf:about="" xmlns:GPano="http://ns.google.com/photos/1.0/panorama/">\n      
<GPano:ProjectionType>equirectangular</GPano:ProjectionType>\n      
<GPano:UsePanoramaViewer>True</GPano:UsePanoramaViewer>\n      
<GPano:CroppedAreaImageWidthPixels>5376</GPano:CroppedAreaImageWidthPixels>\n      
[...]

PR #8171, opened 2024-06-26
Remove zero-byte end padding when parsing any XMP data

radarhere:
Last year, we received a report of a JPEG image where [XMP data was zero-padded at the end](https://github.com/python-pillow/Pillow/issues/7273). I created a fix for that [by adjusting JpegImagePlugin.](https://github.com/python-pillow/Pillow/pull/7274/commits/2488167f33b7e357315c7052273b973fe2cad8e8)

Now https://github.com/python-pillow/Pillow/pull/8069#issuecomment-2186382029 reports that XMP data might also be zero-padded for another format.

So this PR removes zero-byte end padding from XMP data for all formats, not just JPEG.

### DIFF
--- a/Tests/test_image.py
+++ b/Tests/test_image.py
@@ -8,6 +8,7 @@ import sys
 import tempfile
 import warnings
 from pathlib import Path
+from types import ModuleType
 from typing import IO, Any
 
 import pytest
@@ -34,6 +35,12 @@ from .helper import (
     mark_if_feature_version,
     skip_unless_feature,
 )
+
+ElementTree: ModuleType | None
+try:
+    from defusedxml import ElementTree
+except ImportError:
+    ElementTree = None
 
 
 # Deprecation helper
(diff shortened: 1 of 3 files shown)


bigcat88:
Yes, with this changes all will be good.

However I found anther place in Pillow that failing my tests and I think we need to take a look at that:

ImageOps.py , exif_transpose:

```
                    exif_image.info["xmp"] = re.sub(
                        pattern.encode(), b"", exif_image.info["xmp"]
                    )
```                    

I get an KeyError here, as "xmp" is not in **exif_image.info** keys
Maybe here should be a check for that? Should I create a separate issue for that?

bigcat88:
ImageOps.py , exif_transpose:

```python3

            elif "XML:com.adobe.xmp" in exif_image.info:
                for pattern in (
                    r'tiff:Orientation="([0-9])"',
                    r"<tiff:Orientation>([0-9])</tiff:Orientation>",
                ):
                    exif_image.info["XML:com.adobe.xmp"] = re.sub(
                        pattern, "", exif_image.info["XML:com.adobe.xmp"]
                    )
            elif "xmp" in exif_image.info:
                for pattern in (
                    r'tiff:Orientation="([0-9])"',
                    r"<tiff:Orientation>([0-9])</tiff:Orientation>",
                ):
                    exif_image.info["xmp"] = re.sub(
                        pattern.encode(), b"", exif_image.info["xmp"]
                    )
```

imho, this will be the correct code that not fails with exception when "xmp" not in `image.info` dictionary

radarhere:
I imagine [this](https://github.com/bigcat88/pillow_heif/blob/aeda9728bf01b28eec517fc06d2ba24bebce971b/tests/orientation_test.py#L76-L82) is where the failure is occurring? For my own clarity, I wouldn't say that pillow-heif is failing, but rather Pillow is failing to handle an image you've constructed in your tests.

But there's no reason that your constructed image isn't perfectly valid, so I've created https://github.com/python-pillow/Pillow/pull/8173

By the way, thanks for letting us know about these matters before the release has gone out!

PR #8173, opened 2024-06-27
Do not presume "xmp" info simply because "XML:com.adobe.xmp" info exists

radarhere:
#8069 changed `ImageOps.exif_transpose()` so that when updating a PNG image's `im.info["XML:com.adobe.xmp"]`, it would update the newly added `im.info["xmp"]` as well.

However, https://github.com/python-pillow/Pillow/pull/8171#issuecomment-2191625500 has pointed out that our code shouldn't presume that the user didn't manually set `im.info["XML:com.adobe.xmp"]` themselves, in which case `im.info["xmp"]` might be absent.

### DIFF
--- a/Tests/test_imageops.py
+++ b/Tests/test_imageops.py
@@ -432,6 +432,16 @@ def test_exif_transpose() -> None:
     assert 0x0112 not in transposed_im.getexif()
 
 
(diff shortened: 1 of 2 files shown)


bigcat88:
Thank you, with the past and this pull request together - everything will be fine :)

PR #8183, opened 2024-07-01
Drop support for Python 3.8

hugovk:
Python 3.8 reaches end of life in October 2024:

* https://peps.python.org/pep-0569/

Pillow 11.0.0 is planned for 2024-10-15 and will drop Python 3.8: 

* https://github.com/python-pillow/Pillow/milestone/48


.github/workflows/test-docker.yml
```diff
@@ -50,7 +50,6 @@ jobs:
           fedora-39-amd64,
           fedora-40-amd64,
           gentoo,
-          ubuntu-20.04-focal-amd64,
```
hugovk, line 53: We could either update the image to upgrade the default Python version, but given Pillow 11.0 is set for release in October 2024, it's not that long until Ubuntu 20.04's EOL in April 2025 so we could just remove it?

I don't mind too much either way.

radarhere (approved):

hugovk:
The ""Could not resolve host: mirrorlist.centos.org; Unknown error" should be fixed in a cibuildwheel release later today: 

https://github.com/pypa/cibuildwheel/issues/1915

Yay295:
oss-fuzz/CIFuzz is hardcoded to use Python 3.8.3

https://github.com/google/oss-fuzz/blob/b4407a1aa1998de44f13fa7ed0f93259cbc2ab8e/infra/base-images/base-builder/Dockerfile#L22-L23

so the CIFuzz tests fail now with "ERROR: Package 'pillow' requires a different Python: 3.8.3 not in '>=3.9'".

https://github.com/Yay295/Pillow/actions/runs/9960735804/job/27520667252

hugovk:
Related issues and PRs: https://github.com/google/oss-fuzz/issues/11419, https://github.com/google/oss-fuzz/issues/9638, https://github.com/google/oss-fuzz/pull/9532, https://github.com/google/oss-fuzz/pull/11420. 

In the meantime, shall we do something about the failing CI? It's training us to ignore CIFuzz failures.

For example, remove the check or make it always report passed.

radarhere:
I've created https://github.com/python-pillow/Pillow/pull/8239 to skip it.

PR #8184, opened 2024-07-01
Updated deprecation timeline

radarhere:


radarhere:
Closing, as updating the deprecation timeline fails without removing the deprecated features.

https://github.com/python-pillow/Pillow/pull/8182 looks like it will take over from here instead.

PR #8187, opened 2024-07-01
Fix incompatibility with NumPy 1.20 

neutrinoceros:
Fixes an incompatibility encountered in the wild when importing `Pillow==10.4.0` with `numpy==1.20.x` installed.

### DIFF
--- a/src/PIL/_typing.py
+++ b/src/PIL/_typing.py
@@ -7,8 +7,8 @@ from typing import Any, Protocol, Sequence, TypeVar, Union
 try:
     import numpy.typing as npt
 
-    NumpyArray = npt.NDArray[Any]
-except ImportError:
+    NumpyArray = npt.NDArray[Any]  # requires numpy>=1.21
+except (ImportError, AttributeError):
     pass
 
 if sys.version_info >= (3, 10):


radarhere:
NumPy < 1.24 is EOL - https://endoflife.date/numpy

Would an alternative solution be to upgrade NumPy?

neutrinoceros:
> NumPy < 1.24 is EOL - [endoflife.date/numpy](https://endoflife.date/numpy)

One could argue that numpy < 2.0 is EOL since numpy devs have stated multiple times they cannot backport patches to 1.26 anymore, so I don't think NEP 29 (which I assume is the source for this endoflife page) is to be strictly followed.

The presence of this `try/except` block seems to indicate an intention to support versions older than 1.24, and it does work with any version *but* 1.20 , and it's very simple to fix that. I see that Pillow doesn't *require* numpy so I don't think it's possible to express an incompatibility in metadata (something equivalent to pip-style constraints files), except by requiring Python>=3.10 (numpy 1.21.4 being the first version to support it).
I can add this sort of information in a comment if it helps; my patch can be reverted when Python 3.9 is unsupported.


>Would an alternative solution be to upgrade NumPy?

I should note that I discovered this in CI while explicitly testing a library with its minimal requirements (it has matplotlib pinned but not Pillow, since it's not a direct dependency). If this patch is undesired for some reason, bumping minimal requirements is my next best shot, but I'm emphasizing that it's not about a particular virtualenv that would be broken.

radarhere:
> One could argue that numpy < 2.0 is EOL since numpy devs have stated multiple times they cannot backport patches to 1.26 anymore, so I don't think NEP 29 (which I assume is the source for this endoflife page) is to be strictly followed.

That sounds like an argument **for** upgrading NumPy, not for staying on an old version.

> my patch can be reverted when Python 3.9 is unsupported.

I'm mostly conscious of trying to have some line in the sand, so that we're not supporting old versions forever. If this request is just that we support NumPy 1.20 for the next year though, then that is a limited time frame.

radarhere (approved):

neutrinoceros:
> If this request is just that we support NumPy 1.20 for the next year though, then that is a limited time frame.

Yes, I'm not planning to support Python 3.9 (hence, numpy 1.20) beyond early 2025.

Note that I don't think the new title for this PR is correct: this fixes a *runtime* crash, type checking isn't the only (or main) concern.

radarhere:
> this fixes a runtime crash, type checking isn't the only (or main) concern.

Hmm. While this deserves more investigation, I wouldn't have thought that was the case. We install [NumPy 1.17](https://github.com/python-pillow/docker-images/actions/runs/9743529115/job/26893478752#step:6:1677) in our Ubuntu 20.04 CI, and [our test suite runs without a problem.](https://github.com/python-pillow/docker-images/actions/runs/9743529115/job/26893478752#step:7:5053)

neutrinoceros:
> That is surprising - we install [NumPy 1.17](https://github.com/python-pillow/docker-images/actions/runs/9743529115/job/26893478752#step:6:1677) in our Ubuntu 20.04 CI, and [our test suite runs without a problem.](https://github.com/python-pillow/docker-images/actions/runs/9743529115/job/26893478752#step:7:5053)

I'm not surprised: numpy 1.19 and bellow or 1.21 works as intended: only version 1.20 has `numpy.typing` but doesn't have `numpy.typing.NDArray`.

radarhere (approved):

radarhere:
We're about to remove support for Python 3.9 (https://github.com/python-pillow/Pillow/pull/9119), so I've created https://github.com/python-pillow/Pillow/pull/9125.

PR #8189, opened 2024-07-01
Test NumPy on Python 3.13

hugovk:


.github/workflows/wheels-test.sh
```diff
@@ -12,7 +12,7 @@ elif [ "${AUDITWHEEL_POLICY::9}" == "musllinux" ]; then
 else
     yum install -y fribidi
 fi
-if [ "${AUDITWHEEL_POLICY::9}" != "musllinux" ] && !([[ "$OSTYPE" == "darwin"* ]] && [[ $(python3 --version) == *"3.13."* ]]); then
+if [ "${AUDITWHEEL_POLICY::9}" != "musllinux" ] && ! [[ "$OSTYPE" == "darwin"* ]]; then
```
radarhere, line 15:
```suggestion
if [ "${AUDITWHEEL_POLICY::9}" != "musllinux" ]; then
```
The condition in main is 'Don't install NumPy on musllinux and don't install NumPy on Python 3.13 on macOS'. So the second half should be able to be completely removed.

See https://github.com/python-pillow/Pillow/pull/8071/files#diff-3986342f6995d3445cc8a0076546af4257a213b190e39f45e73acda7de8d2fdbR15 for the origin of the second half.

radarhere (approved):

Issue #8199, opened 2024-07-04
Python 3.13 free-threading compatibility


lysnikolaou:
Hey all! 👋 

This aims to be a tracking issue for all work necessary to support the free-threaded build of Python 3.13. A very high-level list of steps is (more details to follow as I investigate more issues):

- [x] Set up CI for free-threading
  - [x] https://github.com/python-pillow/Pillow/pull/8200
- [x] Move away from APIs that use borrowed references under the free-threaded build
  - [x] https://github.com/python-pillow/Pillow/pull/8216
- [x] Audit C extension module for thread-safety issues
  - [x] #8238
- [x] Mark C extension modules as thread-safe with the `Py_mod_gil` slot (or `PyUnstable_Module_SetGIL`)
  - [x] https://github.com/python-pillow/Pillow/pull/8216 
- [x] Upload wheels for the free-threaded build. Other projects have started uploading nightlies for the free-threaded on the [scientific-python-nightly-wheels index](https://anaconda.org/scientific-python-nightly-wheels) so that we can enable faster iteration, we can discuss whether we wanna do this for `Pillow` as well.
  - [X] https://github.com/python-pillow/Pillow/pull/8219
  - [x] https://github.com/scientific-python/upload-nightly-action/issues/84 
  - [x] #8236  
  - [x] https://github.com/python-pillow/Pillow/pull/8243
  - [x] https://github.com/python-pillow/Pillow/pull/8244

hugovk:
Please see PR https://github.com/python-pillow/Pillow/pull/8200 to set up CI for free-threading.

hugovk:
I'm working on this today with @lysnikolaou at the EuroPython sprints. 

It would help the ecosystem if we also upload nightly wheels so others can test the current code without needing to wait for quarterly releases.

@lysnikolaou has requested access at https://github.com/scientific-python/upload-nightly-action/issues/84 for the Scientific Python Nightly Wheels index, where other projects like SciPy are uploading.

See also https://labs.quansight.org/blog/free-threaded-python-rollout and https://py-free-threading.github.io/.

radarhere:
If anyone would like wheels generated after #8216, here you go - [Linux.zip](https://github.com/user-attachments/files/16216912/Linux.zip), [macOS.zip](https://github.com/user-attachments/files/16216941/macOS.zip), [Windows.zip](https://github.com/user-attachments/files/16216960/Windows.zip)

rgommers:
[...]

PR #8200, opened 2024-07-04
Test 3.13 free-threaded build on CI

hugovk:
For https://github.com/python-pillow/Pillow/issues/8199.

Currently GitHub's `actions/setup-python` does not yet support free-threading but it has been requested: https://github.com/actions/setup-python/issues/771 (give it a 👍)

However, it is available via `deadsnakes/action`:

* https://github.com/deadsnakes/action
* https://dev.to/hugovk/help-us-test-free-threaded-python-without-the-gil-1hgf

This PR adds an Ubuntu `3.13-dev` job to the matrix, to run tests with a free-threaded build.

The build actually _runs_ with the GIL enabled, we'll need to explicitly indicate in the extensions that they should run with it disabled. This will be changed in a follow-up PR.

We can check if Python is built with free-threaded mode by looking for "experimental free-threading build" in `python --version --version`. On the CI we can see it under "Build system information":

```
sys.version		 ['3.13.0b2+ experimental free-threading build (main, Jun 27 2024, 08:49:18) [GCC 11.4.0]']
```

https://github.com/python-pillow/Pillow/actions/runs/9794942077/job/27045976814#step:5:10

Compared to the regular build:

```
sys.version		 ['3.13.0b3 (main, Jun 28 2024, 13:42:44) [GCC 11.4.0]']
```

https://github.com/python-pillow/Pillow/actions/runs/9794942077/job/27045975199#step:5:16

### DIFF
--- a/.ci/install.sh
+++ b/.ci/install.sh
@@ -38,7 +38,11 @@ python3 -m pip install pyroma
 
 if [[ $(uname) != CYGWIN* ]]; then
     # TODO Update condition when NumPy supports free-threading
-    if ! [[ "$GHA_PYTHON_VERSION" == "3.13-dev" ]]; then python3 -m pip install numpy ; fi
+    if [[ "$GHA_PYTHON_VERSION" == "3.13-dev" ]]; then
+        python3 -m pip install numpy --index-url https://pypi.anaconda.org/scientific-python-nightly-wheels/simple
+    else
+        python3 -m pip install numpy
+    fi
 
     # PyQt6 doesn't support PyPy3
     if [[ $GHA_PYTHON_VERSION == 3.* ]]; then
--- a/.ci/install.sh
+++ b/.ci/install.sh
@@ -37,12 +37,14 @@ python3 -m pip install -U pytest-timeout
 python3 -m pip install pyroma
 
 if [[ $(uname) != CYGWIN* ]]; then
-    python3 -m pip install numpy
+    # TODO Update condition when NumPy supports free-threading
+    if ! [[ "$GHA_PYTHON_VERSION" == "3.13-dev" ]]; then python3 -m pip install numpy ; fi
 
     # PyQt6 doesn't support PyPy3
     if [[ $GHA_PYTHON_VERSION == 3.* ]]; then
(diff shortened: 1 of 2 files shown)


.ci/install.sh
```diff
@@ -37,12 +37,14 @@ python3 -m pip install -U pytest-timeout
 python3 -m pip install pyroma
 
 if [[ $(uname) != CYGWIN* ]]; then
-    python3 -m pip install numpy
+    # TODO Update condition when NumPy supports free-threading
+    if ! [[ "$GHA_PYTHON_VERSION" == "3.13-dev" ]]; then python3 -m pip install numpy ; fi
```
lysnikolaou, line 41: We can change this to conditionally install `numpy` from `scientific-python-nightly-wheels`, which has nightly wheels that we can test against for free-threading.

lysnikolaou (commented):
I left one comment regarding NumPy versioning.

> The build actually runs with the GIL enabled, we'll need to explicitly indicate in the extensions that they should run with it disabled.

I would recomment adding the `PYTHON_GIL=0` environment variable to keep the GIL disabled and test against that. It's still going to be some time before we can run without the GIL without the environment variable, since extension modules in dependencies (NumPy for example) can enable it as well.

hugovk:
Thanks for the review, how does this look?

.github/workflows/test.yml
```diff
@@ -50,17 +50,14 @@ jobs:
...
-        - os: "macos-13"
-          python-version: "3.9"
+        - { os: "macos-13", python-version: "3.9" }
         exclude:
-        - os: "macos-14"
-          python-version: "3.9"
+        - { os: "macos-14", python-version: "3.9" }
 
     runs-on: ${{ matrix.os }}
     name: ${{ matrix.os }} Python ${{ matrix.python-version }}
```
lysnikolaou, line 63: Maybe include `free-threading` in the name for the free-threaded run?

lysnikolaou (approved):
LGTM! Just one nit in case you find it useful as well.

.github/workflows/test.yml
hugovk: Added: 
image

lysnikolaou (approved):
LG!

PR #8209, opened 2024-07-06
Allow using SIMD intrinsics for SSE2, SSE4, AVX2 and NEON code

homm:
## Current SIMD state

There is a separate [Pillow-SIMD](https://pypi.org/project/Pillow-SIMD/) package on PyPI with modified Pillow code, which includes some rewritten routines using SSE4 and AVX2 compiler intrinsics. Any version of Pillow-SIMD has strictly the same functionality, behavior, and bugs as the corresponding Pillow version. There are many problems with this approach, including:

1. Pillow-SIMD versions are often outdated. This is because any Pillow-SIMD release requires some work to merge SIMD-accelerated code into every Pillow version, which involves resolving potential conflicts and testing all implementations.
2. SIMD-accelerated code is unconditionally platform-specific. Pillow-SIMD can be compiled and run only on x86. This means that it is very hard to add Pillow-SIMD as a dependency in cross-platform applications.
3. Pillow-SIMD is a different package, although it shares the same namespace (PIL), which means only one package (Pillow or Pillow-SIMD) should be installed at a time in one environment. It can be tricky to avoid Pillow installation if it is a dependency for some third-party libraries.
4. Pillow-SIMD itself can't be a dependency for third-party libraries, since it's not cross-platform.
5. Pillow-SIMD doesn't provide precompiled binaries since there are two types of SIMD-acceleration implemented: SSE4 and AVX2. Pillow-SIMD compiled with AVX2 is a bit faster, but can't be run on non-AVX2 CPUs (including Rosetta 2 on Apple silicon).

## Proposal 

I'd like to move SIMD-accelerated code to the upstream with some enhancements for cross-platform compilation. How it affects described problems:

1. Every released Pillow version will contain SIMD-accelerated code, so no additional actions will be required.
2. Pillow can be compiled on any platform, with or without SIMD-acceleration, if available. No separate package will be required.
3. Like the previous point, no separate package will be required.
3. Like the previous point, no separate package will be required.
5. Default precompiled binaries for Pillow will be provided with minimal available SIMD-acceleration for every platform: SSE2 for x86_64 and NEON for ARM aarch64 (please note, currently there is no SSE2 or NEON accelerated code). For advanced users who want acceleration, compilation documentation will be provided.

### Testing

The main challenge is testing. If some code is contributed to the upstream, it should be run during tests. There could be the following versions of the code:

* Native code: Will be tested on platforms without acceleration support: 32-bit x86, s390x, and ppc64le. Are there any tests for any of these platforms?
* SSE2: The most common, since it is the default for x86_64.
* SSE4: Not currently tested. Should be added explicitly.
* AVX2: Not currently tested. Should be added explicitly.
* NEON: Mandatory extension for ARM aarch64, so it will be tested in many ARM configurations.

### Detecting Current Acceleration

There should be a way to detect the current acceleration implementation. I think some property for the core object should be enough.

## Criticism

The proposed solution doesn't provide first-class SIMD-acceleration for Pillow. When you are using any application or installing other libraries, like NumPy, there is only one compiled version available for a platform, and it contains all types of available acceleration. The right accelerated functions are dynamically called at runtime instead of recompiling for particular CPU capabilities during installation. However, such an approach requires much more effort, and this is not the aim of the current proposal.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -52,15 +52,19 @@ jobs:
         include:
         - { python-version: "3.11", PYTHONOPTIMIZE: 1, REVERSE: "--reverse" }
         - { python-version: "3.10", PYTHONOPTIMIZE: 2 }
+        # SIMD-accelerated builds for x86
+        - { os: "ubuntu-latest", python-version: "3.9", acceleration: "sse4"}
+        - { os: "ubuntu-latest", python-version: "3.12", acceleration: "avx2"}
         # Free-threaded
         - { os: "ubuntu-latest", python-version: "3.13-dev", disable-gil: true }
         # M1 only available for 3.10+
         - { os: "macos-13", python-version: "3.9" }
+        - { os: "macos-13", python-version: "3.9", acceleration: "avx2"}
         exclude:
         - { os: "macos-14", python-version: "3.9" }
 
     runs-on: ${{ matrix.os }}
-    name: ${{ matrix.os }} Python ${{ matrix.python-version }} ${{ matrix.disable-gil && 'free-threaded' || '' }}
+    name: ${{ matrix.os }} Python ${{ matrix.python-version }} ${{ matrix.acceleration }} ${{ matrix.disable-gil && 'free-threaded' || '' }}
 
     steps:
     - uses: actions/checkout@v4
@@ -108,7 +112,7 @@ jobs:
         GHA_LIBIMAGEQUANT_CACHE_HIT: ${{ steps.cache-libimagequant.outputs.cache-hit }}
 
     - name: Install macOS dependencies
-      if: startsWith(matrix.os, 'macOS')
+      if: startsWith(matrix.os, 'macos')
       run: |
         .github/workflows/macos-install.sh
       env:
@@ -117,6 +121,11 @@ jobs:
     - name: Register gcc problem matcher
       if: "matrix.os == 'ubuntu-latest' && matrix.python-version == '3.12'"
       run: echo "::add-matcher::.github/problem-matchers/gcc.json"
+
+    - name: Set compiler options for optimization
+      if: ${{ matrix.acceleration }}
+      run: |
+        echo "CC=cc -m${{ matrix.acceleration }}" >> $GITHUB_ENV
 
     - name: Build
       run: |
--- a/src/PIL/features.py
+++ b/src/PIL/features.py
@@ -128,6 +128,7 @@ features = {
     "libjpeg_turbo": ("PIL._imaging", "HAVE_LIBJPEGTURBO", "libjpeg_turbo_version"),
     "libimagequant": ("PIL._imaging", "HAVE_LIBIMAGEQUANT", "imagequant_version"),
     "xcb": ("PIL._imaging", "HAVE_XCB", None),
+    "acceleration": ("PIL._imaging", "acceleration", "acceleration"),
 }
 
 
@@ -267,6 +268,7 @@ def pilinfo(out: IO[str] | None = None, supported_formats: bool = True) -> None:
 
     for name, feature in [
         ("pil", "PIL CORE"),
+        ("acceleration", "Acceleration"),
         ("tkinter", "TKINTER"),
         ("freetype2", "FREETYPE2"),
         ("littlecms2", "LITTLECMS2"),
@@ -291,7 +293,7 @@ def pilinfo(out: IO[str] | None = None, supported_formats: bool = True) -> None:
             if v is None:
                 v = version(name)
             if v is not None:
-                version_static = name in ("pil", "jpg")
+                version_static = name in ("pil", "jpg", "acceleration")
                 if name == "littlecms2":
                     # this check is also in src/_imagingcms.c:setup_module()
                     version_static = tuple(int(x) for x in v.split(".")) < (2, 7)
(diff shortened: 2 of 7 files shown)


homm:
@python-pillow/pillow-team What do you think?

nulano:
I don't see why testing x86_64 would be an issue - we already have special cases for `PYTHONOPTIMIZE` and could do the same for SSE4 / AVX2: https://github.com/python-pillow/Pillow/blob/2152a1736c06b3e99582a7a512e5480aa8d74b66/.github/workflows/test.yml#L53-L54

I think the only x86 (32-bit) job we have is on AppVeyor (using SysWOW64): https://github.com/python-pillow/Pillow/blob/2152a1736c06b3e99582a7a512e5480aa8d74b66/.appveyor.yml#L21-L23 However, it seems to have selected sse2: https://ci.appveyor.com/project/Python-pillow/pillow/builds/50165788/job/sv60thgf3ygibbv4?fullLog=true#L3667

s390x and ppc64le are tested using QEMU: https://github.com/python-pillow/Pillow/blob/2152a1736c06b3e99582a7a512e5480aa8d74b66/.github/workflows/test-docker.yml#L39-L40

---

> Detecting Current Acceleration
>
> There should be a way to detect the current acceleration implementation. I think some property for the core object should be enough.

~~I anticipate that we will get questions about this, so I would propose we add this to `PIL.features` and the report summary `PIL.features.pilinfo`.~~

I see you've already done this 👍 

---

> The right accelerated functions are dynamically called at runtime instead of recompiling for particular CPU capabilities during installation.

I think we should ideally aim for implementing this at some point in the future, but starting with compile-time selection for now would be a good first step.

src/PIL/features.py
```diff
@@ -128,6 +128,7 @@ def get_supported_codecs() -> list[str]:
     "libjpeg_turbo": ("PIL._imaging", "HAVE_LIBJPEGTURBO", "libjpeg_turbo_version"),
     "libimagequant": ("PIL._imaging", "HAVE_LIBIMAGEQUANT", "imagequant_version"),
     "xcb": ("PIL._imaging", "HAVE_XCB", None),
+    "acceleration": ("PIL._imaging", "acceleration", "acceleration"),
```
nulano, line 131: I think this violates the type hint for `check_feature` by returning a string instead of a bool.

.github/workflows/test.yml
```diff
@@ -37,7 +37,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [
-          "macos-14",
+          "macos-14-arm64",
```
hugovk, line 40: No runners are found for `macos-14-arm64`, and `macos-14` is already M1/ARM64 (and only M1):

https://github.blog/changelog/2024-01-30-github-actions-macos-14-sonoma-is-now-available/

```suggestion
          "macos-14",
```

.github/workflows/test.yml
```diff
@@ -52,15 +52,19 @@ jobs:
...
+        - { os: "ubuntu-latest", python-version: "3.9", acceleration: "sse4"}
+        - { os: "ubuntu-latest", python-version: "3.12", acceleration: "avx2"}
         # Free-threaded
         - { os: "ubuntu-latest", python-version: "3.13-dev", disable-gil: true }
         # M1 only available for 3.10+
         - { os: "macos-13", python-version: "3.9" }
+        - { os: "macos-13", python-version: "3.9", acceleration: "avx2"}
         exclude:
-        - { os: "macos-14", python-version: "3.9" }
+        - { os: "macos-14-arm64", python-version: "3.9" }
```
hugovk, line 64:
```suggestion
        - { os: "macos-14", python-version: "3.9" }
```

.github/workflows/test.yml
homm: Thanks. The problem is there is no a list of available runners in documentation, and [in the repository](https://github.com/actions/runner-images/tree/main/images/macos)  `macos-14-arm64` is present.

aclark4life:
Love it 🚀 Thanks @homm, all

src/PIL/features.py
radarhere: I've created https://github.com/uploadcare/pillow-simd/pull/140 to try and address this.

homm: @radarhere Thanks!

jpventorim:
What's the status of this? I use Pillow-SIMD but it's outdated now and using a pillow version with a critical vulnerability, which is a problem. Would be nice to be able to use the newer versions with the optimization.

malagus:
Any chance of merging this pr @homm @radarhere ?

aclark4life:
Sorry @jpventorim @tvasenin to miss your comments. @malagus I think we need a rebase to start with… I'm not sure of the status otherwise. Anything controversial that remains to be discussed ?

jpventorim:
> Anything controversial that remains to be discussed ?

@aclark4life I suppose it's a matter if the maintainers of Pillow want to support this and put the work to test and merge those two. It's so old that I can imagine there will be compatibility issues. Although I'm considering `pillow-simd` to be an abandoned project, I'm still dependent on it at the time, so I'm highly interested on this topic.

tvasenin:
@homm @radarhere @hugovk @nulano

Is it possible to make any progress in order to make this merged? After the last review comments, nothing happened for more than 2 years.

I assume, this should be rebased first, then reviewed again by the approvers.

aclark4life:
> @homm @radarhere @hugovk @nulano
> 
> Is it possible to make any progress in order to make this merged? After the last review comments, nothing happened for more than 2 years.
> 
> I assume, this should be rebased first, then reviewed again by the approvers.

You are welcome to do so! That's definitely the first step but it does not appear this is something we're interested in supporting.

akx:
Bands.c already got some speedups in #9675 (and there's [more pending review and merge](https://github.com/python-pillow/Pillow/pulls?q=is%3Aopen+is%3Apr+label%3APerformance+author%3Aakx)), would be interesting to see how much the hand-rolled SIMD is on top of that. 

There's also a recent [PEP draft about wheel variants, PEP-0825](https://peps.python.org/pep-0825/) that touches exactly this:

> Libraries that ship performance enhanced builds which can be used when certain CPU instruction sets are available, such as AVX2 or AVX-512.

PR #8211, opened 2024-07-07
Remove all WITH_* flags from _imaging.c

homm:
## Rationale

* The value of these flags is questionable. I don't see real use cases.
* Python code is not aware of these flags. After compiling without some of them, Pillow functionality is not guaranteed.


src/_imaging.c
```diff
@@ -215,16 +198,12 @@ PyImaging_AsImaging(PyObject *op) {
 
 void
 ImagingSectionEnter(ImagingSectionCookie *cookie) {
-#ifdef WITH_THREADING
```
homm, line 218: I've tried to measure performance penalty for saving state.

Empty function call:

```python
In [2]: %timeit Image.core.test_threading()
64.1 ns ± 0.842 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

With `ImagingSectionEnter`/`ImagingSectionLeave`:
```python
In [6]: %timeit Image.core.test_threading()
109 ns ± 0.951 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

src/libImaging/Pack.c
```diff
@@ -258,16 +258,6 @@ void
 ImagingPackRGB(UINT8 *out, const UINT8 *in, int pixels) {
     int i = 0;
     /* RGB triplets */
-#ifdef __sparc
```
radarhere, line 261: Could you say a few words about why [this](https://www.ibm.com/docs/en/clearcase/10.0.1?topic=example-defining-architecture-specific-cpp-macros) is removed?

wiredfool:
LGTM, This probably deserves a changelog entry -- as technically it's possible that people have been editing this.

src/_imaging.c
```diff
@@ -3688,20 +3630,14 @@ static struct PyMethodDef methods[] = {
     {"chop_hard_light", (PyCFunction)_chop_hard_light, METH_VARARGS},
     {"chop_overlay", (PyCFunction)_chop_overlay, METH_VARARGS},
 
-#endif
-
-#ifdef WITH_UNSHARPMASK
-    /* Kevin Cazabon's unsharpmask extension */
```
radarhere, line 3694: Just for clarity - you're removing 'Kevin Cazabon' presumably because the code has been modified over time?

homm: If the reason were due to modifications over time, I would have included additional names accordingly. The actual reason is that I was unsure of the specific value this information provides in the comments. However, if this is important, I can revert the change.

radarhere (approved):

hugovk:
Please mention this in the release notes: https://github.com/python-pillow/Pillow/blob/main/docs/releasenotes/11.0.0.rst

homm:
@hugovk @wiredfool 
Release notes added

hugovk (approved):

PR #8212, opened 2024-07-07
Use BreakBeforeBraces: Linux for clang-format

homm:
## Rationale

It's extremely hard to distinguish function arguments from local variables:

```c
void static inline ImagingLineBoxBlur32(
    pixel *lineOut,
    pixel *lineIn,
    int lastx,
    int radius,
    int edgeA,
    int edgeB,
    UINT32 ww,
    UINT32 fw) {
    int x;
    UINT32 acc[4];
    UINT32 bulk[4];
```

```c
static union hist_extrema *
parse_histogram_extremap(
    ImagingObject *self, PyObject *extremap, union hist_extrema *ep) {
    int i0, i1;
    double f0, f1;
```

After proposition:

```c
void static inline ImagingLineBoxBlur32(
    pixel *lineOut,
    pixel *lineIn,
    int lastx,
    int radius,
    int edgeA,
    int edgeB,
    UINT32 ww,
    UINT32 fw)
{
    int x;
    UINT32 acc[4];
    UINT32 bulk[4];
```

```c
static union hist_extrema *
parse_histogram_extremap(
    ImagingObject *self, PyObject *extremap, union hist_extrema *ep)
{
    int i0, i1;
    double f0, f1;
```

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -6,7 +6,7 @@ AllowShortIfStatementsOnASingleLine: false
 AlignAfterOpenBracket: AlwaysBreak
 BinPackArguments: false
 BinPackParameters: false
-BreakBeforeBraces: Attach
+BreakBeforeBraces: Linux
 ColumnLimit: 88
 DerivePointerAlignment: false
 IndentGotoLabels: false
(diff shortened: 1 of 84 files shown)


radarhere:
https://github.com/python-pillow/Pillow/pull/8067 would seem like an alternative suggestion. I'm guessing you prefer this?

homm:
> I'm guessing you prefer this?

Actually, no! I haven't seen #8067 and like it more.

PR #8213, opened 2024-07-07
Remove webp support without anim, mux/demux, and with buggy alpha

homm:
## Rationale

There is a bunch of legacy code which is not even tested:
https://app.codecov.io/gh/python-pillow/Pillow/pull/8211/blob/src/PIL/WebPImagePlugin.py#L50
All Pillow capabilities requires libwebp 0.5.0, which was released in Dec 2015.

## Removed APIs

* `features.check("transp_webp")`
* `features.check("webp_mux")`
* `features.check("webp_anim")`
* `_webp.WebPDecoderVersion()`: not documented and has alternative: `_webp.webpdecoder_version`
* `_webp.WebPDecoderBuggyAlpha()`: not documented
* `_webp.HAVE_WEBPANIM`: not documented
* `_webp.WebPDecode()`: obsolete webp decoder

### DIFF
--- a/Tests/test_file_webp_animated.py
+++ b/Tests/test_file_webp_animated.py
@@ -14,10 +14,7 @@ from .helper import (
     skip_unless_feature,
 )
 
-pytestmark = [
-    skip_unless_feature("webp"),
-    skip_unless_feature("webp_anim"),
-]
+pytestmark = [skip_unless_feature("webp")]
 
 
 def test_n_frames() -> None:
--- a/Tests/test_file_webp_metadata.py
+++ b/Tests/test_file_webp_metadata.py
@@ -10,10 +10,7 @@ from PIL import Image
 
 from .helper import mark_if_feature_version, skip_unless_feature
 
-pytestmark = [
-    skip_unless_feature("webp"),
-    skip_unless_feature("webp_mux"),
-]
+pytestmark = [skip_unless_feature("webp")]
 
 ElementTree: ModuleType | None
 try:
--- a/Tests/test_file_webp_metadata.py
+++ b/Tests/test_file_webp_metadata.py
@@ -133,7 +133,6 @@ def test_getxmp() -> None:
             )
 
 
-@skip_unless_feature("webp_anim")
 def test_write_animated_metadata(tmp_path: Path) -> None:
     iccp_data = b"<iccp_data>"
     exif_data = b"<exif_data>"
--- a/setup.py
+++ b/setup.py
@@ -295,7 +295,6 @@ class pil_build_ext(build_ext):
             "raqm",
             "lcms",
             "webp",
-            "webpmux",
             "jpeg2000",
             "imagequant",
             "xcb",
@@ -794,28 +793,21 @@ class pil_build_ext(build_ext):
 
         if feature.want("webp"):
             _dbg("Looking for webp")
-            if _find_include_file(self, "webp/encode.h") and _find_include_file(
-                self, "webp/decode.h"
+            if all(
+                _find_include_file(self, src)
+                for src in ["webp/encode.h", "webp/mux.h", "webp/demux.h"]
             ):
                 # In Google's precompiled zip it is call "libwebp":
-                if _find_library_file(self, "webp"):
+                if all(
+                    _find_library_file(self, lib)
+                    for lib in ["webp", "webpmux", "webpdemux"]
+                ):
                     feature.webp = "webp"
-                elif _find_library_file(self, "libwebp"):
+                elif all(
+                    _find_library_file(self, lib)
+                    for lib in ["libwebp", "libwebpmux", "libwebpdemux"]
+                ):
                     feature.webp = "libwebp"
-
-        if feature.want("webpmux"):
-            _dbg("Looking for webpmux")
-            if _find_include_file(self, "webp/mux.h") and _find_include_file(
-                self, "webp/demux.h"
-            ):
-                if _find_library_file(self, "webpmux") and _find_library_file(
-                    self, "webpdemux"
-                ):
-                    feature.webpmux = "webpmux"
-                if _find_library_file(self, "libwebpmux") and _find_library_file(
-                    self, "libwebpdemux"
-                ):
-                    feature.webpmux = "libwebpmux"
 
         if feature.want("xcb"):
             _dbg("Looking for xcb")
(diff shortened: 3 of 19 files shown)


src/_webp.c
```diff
@@ -902,87 +762,26 @@ WebPDecoderVersion_str(void) {
...
     }
-#endif
-    PyObject *d = PyModule_GetDict(m);
-    addMuxFlagToModule(m);
-    addAnimFlagToModule(m);
-    addTransparencyFlagToModule(m);
 
     PyObject *v = PyUnicode_FromString(WebPDecoderVersion_str());
-    PyDict_SetItemString(d, "webpdecoder_version", v ? v : Py_None);
-    Py_XDECREF(v);
```
homm, line 985: Not needed: https://docs.python.org/3/c-api/module.html#c.PyModule_AddObject

winbuild/build_prepare.py
```diff
@@ -201,7 +201,7 @@ def cmd_msbuild(
         },
         "build": [
             *cmds_cmake(
-                "webp webpdemux webpmux",
+                "webp webpmux webpdemux",
```
homm, line 204: Preserve order like in other places

src/_webp.c
```diff
@@ -15,10 +13,8 @@
  * very early versions had some significant differences, so we require later
  * versions, before enabling animation support.
  */
-#if WEBP_MUX_ABI_VERSION >= 0x0104 && WEBP_DEMUX_ABI_VERSION >= 0x0105
```
homm, line 18: This ABI versions actually was never released, v0.5.0 have 0x0106 and 0x0106

radarhere:
> features.check("transp_webp")
> features.check("webp_mux")
> features.check("webp_anim")

Removing these without warning sounds a little concerning to me - using them doesn't mean someone is using software from 9 years ago, merely that they are supporting the idea of other people doing so.

homm:
@radarhere pretty reasonable. I've moved them back.

src/_webp.c
wiredfool: IIRC, some of that was for webp as shipped in debian derived distributions, now long deprecated.

Tests/test_file_webp_animated.py
```diff
@@ -15,10 +15,7 @@
     skip_unless_feature,
 )
 
-pytestmark = [
-    skip_unless_feature("webp"),
-    skip_unless_feature("webp_anim"),
-]
+pytestmark = [skip_unless_feature("webp")]
```
nulano, line 18:
```suggestion
pytestmark = skip_unless_feature("webp")
```

This doesn't need to be a list if there is only one.

Tests/test_file_webp_metadata.py
```diff
@@ -10,10 +10,7 @@
 
 from .helper import mark_if_feature_version, skip_unless_feature
 
-pytestmark = [
-    skip_unless_feature("webp"),
-    skip_unless_feature("webp_mux"),
-]
+pytestmark = [skip_unless_feature("webp")]
```
nulano, line 13:
```suggestion
pytestmark = skip_unless_feature("webp")
```

setup.py
```diff
@@ -794,28 +793,21 @@ def build_extensions(self):
 
         if feature.want("webp"):
             _dbg("Looking for webp")
-            if _find_include_file(self, "webp/encode.h") and _find_include_file(
-                self, "webp/decode.h"
+            if all(
+                _find_include_file(self, src)
+                for src in ["webp/encode.h", "webp/mux.h", "webp/demux.h"]
```
nulano, line 798: Just curious, why remove `webp/decode.h`?

src/_webp.c
nulano: If `v` is non-null then yes, but `PyModule_AddObject` will steal a reference to `Py_None` with this change (although perhaps we fail to initialize the module and return -1 in that case instead).

homm: Do you mean that we need to INCREF `Py_None` explicitly?

setup.py
homm: Missed

src/_webp.c
homm: Actually, `Py_None` is not one of the expected values here, so it was replaced with `PyModule_AddObject(m, "webpdecoder_version", v);`. In case `v` is `NULL`, `PyModule_AddObject` won't fail and will return -1.

As of `PyModule_AddObject` result itself, I don't see Pillow cares at all of any `PyModule_Add*` functions in `_imaging.c`.

homm: On the other hand, Pillow doesn't care about the `PyModule_Add*` result because the value is never `NULL`. Ref counting in Python is hard :-) Reverted it.

nulano (approved):

docs/releasenotes/11.0.0.rst
```diff
@@ -43,10 +43,13 @@ similarly removed.
...
 
-TODO
-^^^^
+WebP Features Test
+^^^^^^^^^^^^^^^^^^
 
-TODO
+The following features ``features.check("transp_webp")``,
+``features.check("webp_mux")``, and ``features.check("webp_anim")`` are now
+always ``True`` if the WebP module is installed and sohuld not be used.
```
radarhere, line 51:
```suggestion
always ``True`` if the WebP module is installed and should not be used.
```

docs/releasenotes/11.0.0.rst
```diff
@@ -43,10 +43,13 @@ similarly removed.
...
+These checks wil be removed in Pillow 12.0.0 (2025-10-15).
```
radarhere, line 52:
```suggestion
These checks will be removed in Pillow 12.0.0 (2025-10-15).
```

homm:
Rebased

src/_webp.c
```diff
@@ -15,10 +13,8 @@
...
-#define HAVE_WEBPANIM
-#endif
-
+#if WEBP_MUX_ABI_VERSION < 0x0106 || WEBP_DEMUX_ABI_VERSION < 0x0107
+#error libwebp 0.5.0 and above is required. Upgrade libwebp or build with --disable-webp flag
```
radarhere, line 17:
```suggestion
#error libwebp 0.5.0 and above is required. Upgrade libwebp or build Pillow with --disable-webp flag
```
Just clarifying that it's not libwebp that is being built with `--disable-webp`

homm:
Rebased and squashed some commits

homm:
@radarhere could you see at this? I'm not experienced with mypy at all
https://github.com/python-pillow/Pillow/actions/runs/10364204279/job/28689074326?pr=8213

src/PIL/features.py
```diff
@@ -119,9 +120,9 @@ def get_supported_codecs() -> list[str]:
 
 
 features = {
```
radarhere, line 122:
```suggestion
features: dict[str, tuple[str, str | bool, str | None]] = {
```
This should fix the type hint error.

homm: Thanks! This helps

radarhere (approved):

PR #8216, opened 2024-07-08
Move away from APIs that use borrowed references under the free-threaded build

hugovk:
For https://github.com/python-pillow/Pillow/issues/8199.

Make a start on making Pillow's code ready for free-threading ([PEP 703](https://peps.python.org/pep-0703/)).

This is some of the low-hanging fruit from https://docs.python.org/3.13/howto/free-threading-extensions.html.

* Declare extension modules can support running with the GIL disabled. We use [single-phase initialisation](https://docs.python.org/3.13/howto/free-threading-extensions.html#single-phase-initialization).

* Replace functions that return borrowed references [with strong references](https://docs.python.org/3.13/howto/free-threading-extensions.html#borrowed-references), using https://github.com/python/pythoncapi-compat

* Plus from @lysnikolaou: don't use `PyList_GetItemRef` immediately after `PyList_New`

### DIFF
--- a/src/_imagingft.c
+++ b/src/_imagingft.c
@@ -20,6 +20,7 @@
 
 #define PY_SSIZE_T_CLEAN
 #include "Python.h"
+#include "thirdparty/pythoncapi_compat.h"
 #include "libImaging/Imaging.h"
 
 #include <ft2build.h>
@@ -1219,7 +1220,7 @@ font_getvarnames(FontObject *self) {
         }
 
         for (j = 0; j < num_namedstyles; j++) {
-            if (PyList_GetItem(list_names, j) != NULL) {
+            if (PyList_GetItemRef(list_names, j) != NULL) {
                 continue;
             }
 
--- a/src/_imagingft.c
+++ b/src/_imagingft.c
@@ -1576,5 +1576,9 @@ PyInit__imagingft(void) {
         return NULL;
     }
 
+#ifdef Py_GIL_DISABLED
+    PyUnstable_Module_SetGIL(m, Py_MOD_GIL_NOT_USED);
+#endif
+
     return m;
 }
--- a/src/_imagingft.c
+++ b/src/_imagingft.c
@@ -1243,8 +1243,8 @@ font_getvarnames(FontObject *self) {
         }
     }
 
+    PyMem_Free(list_names_filled);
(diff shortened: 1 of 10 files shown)


src/_imagingft.c
```diff
@@ -1204,6 +1205,16 @@ font_getvarnames(FontObject *self) {
 
     num_namedstyles = master->num_namedstyles;
     list_names = PyList_New(num_namedstyles);
+
+    int *list_names_filled = PyMem_Malloc(num_namedstyles * sizeof(int));
```
lysnikolaou, line 1209: This was my bad, but there should be a `PyMem_Free` at the end of the function for this. Sorry!

src/encode.c
```diff
@@ -721,7 +722,10 @@ PyImaging_LibTiffEncoderNew(PyObject *self, PyObject *args) {
         }
 
         if (!is_core_tag) {
-            PyObject *tag_type = PyDict_GetItem(types, key);
+            PyObject *tag_type;
+            if (PyDict_GetItemRef(types, key, &tag_type) == 0) {
+                PyErr_SetString(PyExc_KeyError, "unknown tag type");
+            }
```
lysnikolaou, line 728: This wasn't raising before, so it should probably not raise now either, right?

Also, the `-1` case should be handled appropriately here.

lysnikolaou (approved):
LGTM! Let's merge this.

radarhere:
pythoncapi_compat.h has been added in it's entirety, but if I'm reading this right, we're only using `PyDict_GetItemRef()`.

Was the entire file added just for simplicity, or do you expect that other functions will be used in the future as well?

hugovk:
I thought about only including what we use, but added the whole thing for simplicity. It also makes updating it easier, and we might use more in the future.

I'm not against using a minimal file, but then again, does it affect compiled executable size (at all or significantly)? (I'm on a plane waiting to take off so can't test right now :)

radarhere:
The file is currently 1103 bytes, so I don't consider size to be a problem. I just wanted to clarify the rationale.

nulano:
> does it affect compiled executable size (at all or significantly)

They appear to all be `static inline` functions, so as far as I understand (I also checked GCC on godbolt.org), only those functions that we actually use should be included in the compiled executable.

PR #8219, opened 2024-07-09
Build free-threaded wheels

radarhere:
Helps #8199

Does not build free-threaded wheels for aarch64. I guess that isn't supported yet by cibuildwheel.


hugovk:
Thank you!

PR #8227, opened 2024-07-12
Return boolean from ImageMath lambda_eval and unsafe_eval comparison operations

radarhere:
```python
from PIL import Image, ImageMath
A = Image.new("L", (1, 1))
print(ImageMath.lambda_eval(lambda args: args["A"] == args["A"], A=A))
```
currently gives `<PIL.Image.Image image mode=I size=1x1 at 0x1047906E0>`

This PR suggests changing the output to a more meaningful `True`.

`==`, `!=`, `<`, `<=`, `>`, `>=`, `equal()` and `notequal()` would all now return boolean results.


Tests/test_imagemath_lambda_eval.py
```diff
@@ -176,8 +176,9 @@ def test_compare() -> None:
         )
         == "I 2"
     )
-    assert pixel(ImageMath.lambda_eval(lambda args: args["A"] == 1, images)) == "I 1"
-    assert pixel(ImageMath.lambda_eval(lambda args: args["A"] == 2, images)) == "I 0"
+    for i in range(1, 3):
+        assert ImageMath.lambda_eval(lambda args: args["A"] == i, images) is False
```
homm, line 180: Not related to this changes, but for me this syntax should be deprecated in favor to more pythonic:
```suggestion
        assert ImageMath.lambda_eval(lambda args: args["A"] == i, **images) is False
```

radarhere: I've pushed a commit to update how it is called in the tests.

I'm guessing you're not suggesting we actually deprecate any functionality, as we're not doing anything special with `kw`, so this would just be natural Python handling of keywords.

homm: Actually I am :-)

For me there are no reasons to support two ways for passing arguments. I see that Python's `eval` gets dicts as arguments, so we could adopt only this and remove kwargs, or keep kwargs only since this is more convenient and pythonistic.

radarhere: Oh, I see now, you want to deprecate `options` or `**kw` - https://github.com/python-pillow/Pillow/blob/d05262105ada44e79f10736823cef7105de795c1/src/PIL/ImageMath.py#L241-L261

radarhere: I've removed these changes from here, and created https://github.com/python-pillow/Pillow/pull/8242 instead.

hugovk:
Is this a breaking change?

radarhere:
For ImageMath expressions that involve equality or comparison checks, yes.

Here are some alternative ways forward.
1) Keep things as they are. The return type is `_Operand`, and we can type hint it as such and use a `# type: ignore[override]` to prevent mypy saying
```
error: Return type "_Operand" of "__eq__" incompatible with return type "bool" in supertype "object"
```
2) We could made this breaking change only for `lambda_eval()` and `unsafe_eval()`. It would still be a breaking change for those methods (added in 10.3.0), but `eval()` would stay the same until it is removed in Pillow 12. Not a perfect solution, but an improvement.

hugovk:
@python-pillow/pillow-team What do you think?

aclark4life:
Probably one of the few good times to make a breaking change (from 10.x to 11) and worst case you could follow up with options 1 or 2 in `11.0.1` if some anticipated breakage occurred in the wild. 🤷

hugovk:
We've usually been cautious about breaking changes without at least a year of deprecations. 

I would prefer alternative 1 or 2.

This PR is to improve the API, but motivated by adding type hints, and not user demand, so I would err on bring cautious.

radarhere:
Ok, I've pushed a commit to implement 2 - only changing the behaviour for `lambda_eval()` and `unsafe_eval()`, the methods that were added in 10.3.0.

radarhere:
I've created #8464 to implement 1 - I've realised that the current form does have conceivably useful functionality. For example, you can invert the upper left quadrant of hopper.
```pycon
>>> from PIL import Image, ImageMath
>>> im = Image.open("Tests/images/hopper.png").convert("1")
>>> b = Image.new("1", (128, 128))
>>> for x in range(64):
...   for y in range(64):
...     b.putpixel((x, y), 1)
... 
>>> ImageMath.eval("(A != B) * 65535", A=im, B=b).save("out.png")
```
out

PR #8234, opened 2024-07-15
Added type hints

radarhere:


Ilovemesomehim1 (approved):

src/PIL/IcoImagePlugin.py
```diff
@@ -222,8 +229,7 @@ def frame(self, idx: int) -> Image.Image:
             im.tile[0] = d, (0, 0) + im.size, o, a
 
             # figure out where AND mask image starts
-            bpp = header["bpp"]
-            if 32 == bpp:
+            if 32 == header.bpp:
```
hugovk, line 232: This sort of defensive programming (["Yoda conditions"](https://en.wikipedia.org/wiki/Yoda_conditions)) can be useful in some languages, but not needed for Python:

```suggestion
            if header.bpp == 32:
```

hugovk (approved):

PR #8235, opened 2024-07-15
Simplified code

radarhere:
#8216 has changed the code to return early if `axis_name` is `NULL`.

https://github.com/python-pillow/Pillow/blob/6944e9e1831c621db296e4ba708b38e5be378995/src/_imagingft.c#L1311-L1319

So the ternary condition can be removed.


lysnikolaou (approved):
LGTM! I missed this while working on this part of the code. Thanks for noticing!

PR #8238, opened 2024-07-16
Lock around usages of imaging memory arenas

lysnikolaou:
Related to #8199.

Did some testing with threads. Thread-sanitizer is also happy after these changes.


src/libImaging/Storage.c
```diff
@@ -365,17 +368,18 @@ ImagingDestroyArray(Imaging im) {
 
     if (im->blocks) {
         while (im->blocks[y].ptr) {
+            MUTEX_LOCK(&ImagingDefaultArena.mutex);
             memory_return_block(&ImagingDefaultArena, im->blocks[y]);
+            MUTEX_UNLOCK(&ImagingDefaultArena.mutex);
             y += 1;
         }
```
Yay295, line 377: It would probably be better to move the lock/unlock outside the loop.

src/_imaging.c
```diff
@@ -4071,14 +4091,20 @@ _set_blocks_max(PyObject *self, PyObject *args) {
     if (blocks_max < 0) {
         PyErr_SetString(PyExc_ValueError, "blocks_max should be greater than 0");
         return NULL;
-    } else if (
-        (unsigned long)blocks_max >
-        SIZE_MAX / sizeof(ImagingDefaultArena.blocks_pool[0])) {
+    }
+
+    MUTEX_LOCK(&ImagingDefaultArena.mutex);
+    size_t blocksize = sizeof(ImagingDefaultArena.blocks_pool[0]);
```
nulano, line 4097:
```c
sizeof(ImagingDefaultArena.blocks_pool[0])
````
is evaluated at compile time so there is no need to lock here.

In fact, you could even use
```c
sizeof(((ImagingMemoryArena*)0)->blocks_pool[0])
```

lysnikolaou: Thanks for the pointer! Updated.

lysnikolaou:
This is ready to be reviewed more throughly. All tests pass and tsan is happy.

nulano:
> All tests pass and tsan is happy.

Out of curiosity, how are you testing with tsan? I don't think we have it in CI (although perhaps we should add it!).

lysnikolaou:
> Out of curiosity, how are you testing with tsan? I don't think we have it in CI (although perhaps we should add it!).

I added a few very very basic threaded tests, built CPython and Pillow with tsan enabled and then ran the tests. It all takes forever, cause CPython with tsan is very slow, but it works.

If your question is how I build Pillow with tsan enabled, it is really easy:
```
CC="clang -fsanitize=thread" python -m pip install .
```

hugovk (approved):

hugovk:
If no other reviews or comments, I'll merge this tomorrow.

hugovk:
Thanks @lysnikolaou!

PR #8239, opened 2024-07-16
Temporarily disable cifuzz

radarhere:
While we have [dropped support for Python 3.8](https://github.com/python-pillow/Pillow/pull/8183), the[ cifuzz base image is still using it](https://github.com/google/oss-fuzz/pull/11401#issuecomment-1872050104).

Until https://github.com/google/oss-fuzz/issues/11419 is resolved by upgrading the Python version, this skips the cifuzz job.


hugovk:
Thanks!

PR #8242, opened 2024-07-17
Deprecate ImageMath lambda_eval and unsafe_eval options argument

radarhere:
Suggested in https://github.com/python-pillow/Pillow/pull/8227#discussion_r1677553779

Deprecate ImageMath lambda_eval and unsafe_eval `options` argument, in favour of providing arguments through individual keywords instead.

### DIFF
--- a/Tests/test_imagemath_lambda_eval.py
+++ b/Tests/test_imagemath_lambda_eval.py
@@ -1,5 +1,9 @@
 from __future__ import annotations
 
+from typing import Any
+
+import pytest
+
 from PIL import Image, ImageMath
 
 
@@ -19,7 +23,7 @@ I = Image.new("I", (1, 1), 4)  # noqa: E741
 A2 = A.resize((2, 2))
 B2 = B.resize((2, 2))
 
-images = {"A": A, "B": B, "F": F, "I": I}
+images: dict[str, Any] = {"A": A, "B": B, "F": F, "I": I}
 
 
 def test_sanity() -> None:
@@ -30,13 +34,13 @@ def test_sanity() -> None:
         == "I 3"
     )
     assert (
-        pixel(ImageMath.lambda_eval(lambda args: args["A"] + args["B"], images))
+        pixel(ImageMath.lambda_eval(lambda args: args["A"] + args["B"], **images))
         == "I 3"
     )
     assert (
         pixel(
             ImageMath.lambda_eval(
-                lambda args: args["float"](args["A"]) + args["B"], images
+                lambda args: args["float"](args["A"]) + args["B"], **images
             )
         )
         == "F 3.0"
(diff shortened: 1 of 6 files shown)


homm (approved):

Tests/test_imagemath_lambda_eval.py
```diff
@@ -19,7 +23,7 @@ def pixel(im: Image.Image | int) -> str | int:
 A2 = A.resize((2, 2))
 B2 = B.resize((2, 2))
 
-images = {"A": A, "B": B, "F": F, "I": I}
+images: dict[str, Any] = {"A": A, "B": B, "F": F, "I": I}
```
homm, line 26: Does it work?
```suggestion
images: dict[str, Image.Image] = {"A": A, "B": B, "F": F, "I": I}
```

docs/reference/ImageMath.rst
```diff
@@ -37,10 +37,10 @@ Example: Using the :py:mod:`~PIL.ImageMath` module
 
     :param expression: A function that receives a dictionary.
     :param options: Values to add to the function's dictionary, mapping image
```
homm, line 39: We need to add **kwargs in function signature and params description

homm (changes requested):

Tests/test_imagemath_lambda_eval.py
radarhere: Not exactly, no.
```
Tests/test_imagemath_lambda_eval.py:37: note: "Dict" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
Tests/test_imagemath_lambda_eval.py:37: note: Consider using "Mapping" instead, which is covariant in the value type
Tests/test_imagemath_lambda_eval.py:43: error: Argument 2 to "lambda_eval" has incompatible type "**dict[str, Image]"; expected "dict[str, Any]"  [arg-type]
```
This would be because `**kw` is `Any`.
https://github.com/python-pillow/Pillow/blob/47fc36a323892040b1945afd22d036a15fb99abd/src/PIL/ImageMath.py#L239-L243

docs/reference/ImageMath.rst
```diff
@@ -31,20 +31,21 @@ Example: Using the :py:mod:`~PIL.ImageMath` module
...
-                    names to Image instances. You can use one or more keyword
-                    arguments instead of a dictionary, as shown in the above
-                    example. Note that the names must be valid Python
-                    identifiers.
+                    names to Image instances. Note that the names must be
+                    valid Python identifiers. Deprecated.
+                    You can instead use one or more keyword arguments, as
+                    shown in the above example.
+    :param **kw: Values to add to the function's dictionary.
```
homm, line 44: It seems reasonable to add long description to actual argument and leave only short version for deprecated one.

Tests/test_imagemath_lambda_eval.py
homm: It looks strange. Isn't `Image` `Any`?

radarhere: I think the key part of mypy's response is

> "Dict" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance

`Image` may sit within `Any`, but it is not `Any`. I don't think there's a better option, short of removing `images` altogether and just using the image variables individually.

docs/reference/ImageMath.rst
radarhere: I've updated the commit, moving some text around.

homm (approved):

PR #8243, opened 2024-07-17
Skip QEMU-emulated wheels on workflow dispatch event

lysnikolaou:


radarhere:
I expect we will revert this before the next release - it is helpful to be able to manually trigger all of the wheel jobs before a release, to increase confidence that everything will work correctly.

lysnikolaou:
> I expect we will revert this before the next release - it is helpful to be able to manually trigger all of the wheel jobs before a release, to increase confidence that everything will work correctly.

Good point. Maybe it makes sense to add an input to the workflow so that someone can choose whether to build the emulated wheels or not. I can open a PR if you think that's a good idea.

radarhere:
> Maybe it makes sense to add an input to the workflow

At that point, it's probably simplest to split the workflow into two.

Do you anticipate frequently wanting to manually trigger an upload to scientific python? I would have thought the motivation for this PR was just to temporarily speed things up while sorting out #8236

lysnikolaou:
> Do you anticipate frequently wanting to manually trigger an upload to scientific python?

No, not at all. Most nightly uploads are supposed to be happening through the cron job.

PR #8247, opened 2024-07-19
Remove unnecessary pymode variable

Yay295:
The `mode` string can be directly passed to `Py_BuildValue()` as an "s" value instead of an "S" value where we need to convert it to a Python object ourselves.


radarhere:
#8213 would this code altogether.

Yay295:
#8213 removed this code.

PR #8265, opened 2024-07-27
Improve ImageDraw2 shape methods

radarhere:
Some improvements to ImageDraw2.

1. You might think that ImageDraw2 `ellipse()` only requires `xy`,

https://github.com/python-pillow/Pillow/blob/8f62fbdf440b99cc087c19b58725ff266b76758d/src/PIL/ImageDraw2.py#L132

but that is not the case.

```python
from PIL import Image, ImageDraw2
d = ImageDraw2.Draw(im)
d.ellipse((0, 0, 100, 100))
```
gives
```pytb
  File "demo.py", line 4, in <module>
    d.ellipse((0, 0, 100, 100))
  File "PIL/ImageDraw2.py", line 138, in ellipse
    self.render("ellipse", xy, *options)
TypeError: render() missing 1 required positional argument: 'pen'
```

So I've made a change to add `pen` as a required argument to `arc()` and other methods.

Some of our test code does pass through `pen` as `None`, and uses the `brush` argument for the `Pen` instead.

https://github.com/python-pillow/Pillow/blob/8f62fbdf440b99cc087c19b58725ff266b76758d/Tests/test_imagedraw2.py#L115-L120

But even if `pen` is `None`, the fact that `brush` needs to come afterwards means that a value for `pen` is still required.

2. At the moment, ImageDraw2 `arc()` does not work.

https://github.com/python-pillow/Pillow/blob/8f62fbdf440b99cc087c19b58725ff266b76758d/src/PIL/ImageDraw2.py#L114

```pytb
>>> from PIL import Image, ImageDraw2
>>> im = Image.new("RGB", (1, 1))
>>> d = ImageDraw2.Draw(im)
>>> d.arc((0, 0), 0, 180)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/PIL/ImageDraw2.py", line 121, in arc
    self.render("arc", xy, start, end, *options)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/PIL/ImageDraw2.py", line 107, in render
    getattr(self.draw, op)(xy, fill=fill, outline=outline)
TypeError: arc() got an unexpected keyword argument 'outline'
```

This is because ImageDraw2 `render()` passes `outline` to every method except for `line()`, but ImageDraw `arc()` doesn't accept that.

https://github.com/python-pillow/Pillow/blob/8f62fbdf440b99cc087c19b58725ff266b76758d/src/PIL/ImageDraw2.py#L104-L107

https://github.com/python-pillow/Pillow/blob/8f62fbdf440b99cc087c19b58725ff266b76758d/src/PIL/ImageDraw.py#L174-L181

I've made a change for that.

3. You will also notice in the above code that the required `start` and `end` parameters aren't passed through to ImageDraw. I've added `**kwargs` to `render()` to allow them to be added.

### DIFF
--- a/Tests/test_imagedraw2.py
+++ b/Tests/test_imagedraw2.py
@@ -66,6 +66,36 @@ def test_mode() -> None:
 
 
 @pytest.mark.parametrize("bbox", BBOX)
+@pytest.mark.parametrize("start, end", ((0, 180), (0.5, 180.4)))
+def test_arc(bbox: Coords, start: float, end: float) -> None:
+    # Arrange
+    im = Image.new("RGB", (W, H))
+    draw = ImageDraw2.Draw(im)
+    pen = ImageDraw2.Pen("white", width=1)
+
+    # Act
+    draw.arc(bbox, pen, start, end)
+
+    # Assert
+    assert_image_similar_tofile(im, "Tests/images/imagedraw_arc.png", 1)
+
+
+@pytest.mark.parametrize("bbox", BBOX)
+def test_chord(bbox: Coords) -> None:
+    # Arrange
+    im = Image.new("RGB", (W, H))
+    draw = ImageDraw2.Draw(im)
+    pen = ImageDraw2.Pen("yellow")
+    brush = ImageDraw2.Brush("red")
+
+    # Act
+    draw.chord(bbox, pen, 0, 180, brush)
+
+    # Assert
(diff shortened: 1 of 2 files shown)


hugovk:
Thanks, this makes the API easier to use 👍

PR #8274, opened 2024-08-01
Deprecate JpegImageFile huffman_ac and huffman_dc

radarhere:
These dictionaries have been unused since PIL, so let's deprecate them.

https://github.com/python-pillow/Pillow/blob/4df4df2a18ded44811be786703a43c3460cb3bc9/src/PIL/JpegImagePlugin.py#L349-L350


hugovk:
And mention something like "and will be removed in Pillow 12 (2025-10-15)" in deprecations.rst and 11.0.0.rst?

hugovk (approved):

PR #8275, opened 2024-08-01
Explicitly type Image __init__ attributes

Yay295:


Yay295:
Unfortunately, trying to type `im` and `palette` leads to a lot of false errors because mypy can't always determine that those values aren't `None`.

Related:
https://github.com/python/mypy/issues/10151
https://github.com/python/mypy/issues/11969

radarhere:
I've created #8279 as an alternative to this, addressing `im` and `palette` as well.

Yay295:
Other than thee `im` and `palette` changes, the only differences I see are a slightly different handling of `PngImagePlugin`, and a very different handling of `EpsImagePlugin`.

radarhere:
I think my versions of the EPS and PNG changes are simpler. I've also restricted the hinting of `self.info` keys more than this PR.

Yay295:
I'm not sure how you think your PNG change are simpler when you introduce a new local variable and I don't.

radarhere:
My opinion is that it's simpler because my version declares what the type is once, whereas this does it three times.

radarhere:
https://github.com/python-pillow/Pillow/pull/8279 has been merged instead.

Issue #6382, opened 2022-06-20
Bitmap ".eps" image have incorrect size and mode


mlove4u:
* OS: OSX 11.6
* Python: 3.7.4
* Pillow: 9.1.1

```python
eps_file_path = "EPS_Bitmap.eps" # File was attached.
img = Image.open(eps_file_path)
print(img.mode, img.size)  # RGB (90, 60) --> correct: L, (500,333)
```
Am I doing it the wrong way? This is how I'm using it for now (getting data from %ImageData).Thank you.
```python
def get_eps_info(eps_file_path):
    """
    https://mail.python.org/pipermail/image-sig/2005-September/003575.html
    Photoshop 6.0 SDK
    EPS Parameters for ImageData (Photoshop 3.0 and later)
    columns      Width of the image in pixels
    rows         Height of the image in pixels
    depth        Number of bits per channel. Must be 1 or 8
    mode         Image mode. Bitmap/grayscale = 1; Lab = 2; RGB = 3; CMYK = 4
    pad channels Number of other channels stored in the file.
                Ignored when reading...
    block size   Number of bytes per row per channel.
                 Will be either 1 or formula (below):
                 1 = Data is interleaved
                 (columns*depth+7)/8 = Data is stored in line interleaved format
                 or there is only one channel
    binary/ascii 1 = Data is in binary format
                 2 = Data is in hex ascii format
    data start   Entire PostScriptline immediately preceding the image data.
                This entire line should not occur elewhere in the PostScript
                header code, but it may occur at part of line
    """
    p = re.compile(rb'%ImageData:\s(\d.+)\s"beginimage"')
    with open(eps_file_path, "rb") as f:
        for line in f.readlines():
            info = re.match(p, line)
            if info:
                columns, rows, depth, mode, pad_channels, block_size, data_format = map(
                    lambda x: int(x), info.group(1).decode().split(" "))
                return columns, rows, mode
print(get_eps_info(eps_file_path))  # (500, 333, 1)
```
[EPS_Bitmap.eps.zip](https://github.com/python-pillow/Pillow/files/8937742/EPS_Bitmap.eps.zip)

radarhere:
[...]

PR #8281, opened 2024-08-03
Updated EPS mode when opening images without transparency

Yay295:
1. I moved all EPS images (and images converted from EPS images) to their own folder so that they're all together.
2. I noticed that `1.eps` has a `BoundingBox` of (72, 48), but its `ImageData` says its size is (100, 67). Pillow currently believes the image size should be (100, 67), so I've added a BMP version of this image at that size to compare it to. However, this comparison does not currently work, because an error occurs when calling `Image.tobytes()` on the EPS image.

`ValueError: No packer found from RGB to 1`


Yay295:
I tried converting "1.eps" to PNG using ImageMagick, and it converted it to a 72x48 sized image. So perhaps Pillow isn't using the right size either.

Yay295:
Mode "1" and "L" images should now load as those modes instead of being converted to RGB. "LAB" and "CMYK" are still converted to RGB though.

radarhere:
> I tried converting "1.eps" to PNG using ImageMagick, and it converted it to a 72x48 sized image. So perhaps Pillow isn't using the right size either. Not that there necessarily _is_ a "right" size, since EPS isn't a raster image format.

This is reverting the change requested in https://github.com/python-pillow/Pillow/issues/6382#issuecomment-1213597912. I'm reluctant to change how Pillow behaves again just because a different third-party application was chosen to compare this against?

Yay295:
> > I tried converting "1.eps" to PNG using ImageMagick, and it converted it to a 72x48 sized image. So perhaps Pillow isn't using the right size either. Not that there necessarily _is_ a "right" size, since EPS isn't a raster image format.
> 
> This is reverting the change requested in [#6382 (comment)](https://github.com/python-pillow/Pillow/issues/6382#issuecomment-1213597912). I'm reluctant to change how Pillow behaves again just because a different third-party application was chosen to compare this against?

I used ImageMagick because I don't have Photoshop, but I would consider Photoshop to be the one to follow, since EPS is an Adobe creation. I'll remove that commit.

src/PIL/EpsImagePlugin.py
```diff
@@ -324,32 +325,51 @@ def _read_comment(s: str) -> bool:
...
                         self._mode = self.mode_map[mode_id]
                     except ValueError:
-                        break
+                        pass
                 else:
-                    break
-
-                self._size = columns, rows
-                return
+                    pass
```
radarhere, line 338: Couldn't
```python
else:
    pass
```
just be removed?

Yay295: Yes. I'm used to using a linter that requires an `else` for every `if`/`elif`, but that's not the case for this project.

src/PIL/EpsImagePlugin.py
```diff
@@ -193,7 +193,11 @@ def _open(self) -> None:
         self.fp.seek(offset)
 
         self._mode = "RGB"
-        self._size = None
+
+        # The last occurrence of one of these values
+        # takes precedence over prior occurrences.
+        bounding_box = None
+        imagedata_size = None
```
radarhere, line 200: I'm not convinced that this is always the case.

https://web.mit.edu/PostScript/Adobe/Documents/5001.DSC_Spec.pdf
>  In the case of multiple header comments, the first comment encountered is
considered to be the truth. In the case of multiple trailer comments (those
comments that were deferred using the (atend) convention), the last
comment encountered is considered to be the truth.

src/PIL/EpsImagePlugin.py
```diff
@@ -324,32 +325,50 @@ def _read_comment(s: str) -> bool:
...
+                bounding_box[3] - bounding_box[1],
+            )
+
+        self.tile = [
+            ImageFile._Tile(
+                codec_name="eps",
+                extents=(0, 0) + self._size,
+                offset=offset,
+                args=(length, bounding_box),
+            )
```
radarhere, line 369: I realise mentioning the names for the NamedTuple would just be your personal preference, but ideally Pillow's code would be somewhat consistent, rather than just a combination of the styles of whoever worked on a given plugin last.

See, for example, JpegImagePlugin, where the tile is created without names.

https://github.com/python-pillow/Pillow/blob/0ec1153a627a46b978022c68c2adce89ff81f40d/src/PIL/JpegImagePlugin.py#L829-L831

src/PIL/EpsImagePlugin.py
```diff
@@ -333,22 +341,38 @@ def _read_comment(s: str) -> bool:
...
             bytes_read = 0
 
-        if not self.tile:
+        # A "BoundingBox" is always required,
+        # even if an "ImageData" descriptor size exists.
+        if not bounding_box:
             msg = "cannot determine EPS bounding box"
             raise OSError(msg)
 
+        # An "ImageData" size takes precedence over the "BoundingBox".
```
radarhere, line 363: From my reading of https://github.com/python-pillow/Pillow/pull/8281#issuecomment-2267145829, is this logic just based on your experience of ImageMagick?

src/PIL/EpsImagePlugin.py
```diff
@@ -333,22 +341,38 @@ def _read_comment(s: str) -> bool:
...
                 s = str(bytes_mv[:bytes_read], "latin-1")
-                _read_comment(s)
+                read_comment(s)
             elif bytes_mv[:9] == b"%%Trailer":
                 trailer_reached = True
             bytes_read = 0
 
-        if not self.tile:
+        # A "BoundingBox" is always required,
+        # even if an "ImageData" descriptor size exists.
```
radarhere, line 358: Could you explain to me (or refer to a specific part of a specification) why this is the case?

src/PIL/EpsImagePlugin.py
```diff
@@ -333,22 +341,38 @@ def _read_comment(s: str) -> bool:
                 else:
                     break
 
-                self._size = columns, rows
-                return
+                # Read the columns and rows after checking the bit depth and mode
```
radarhere, line 344:
```suggestion
                # Parse the columns and rows after checking the bit depth and mode
```
In a Pillow context, I'm very used to thinking of 'read' as meaning 'read from file', which isn't what's happening here.

src/PIL/EpsImagePlugin.py
Yay295: https://www.loc.gov/preservation/digital/formats/fdd/fdd000246.shtml

> The PostScript code that specifies an EPS image is subject to constraints, specified in the [PostScript Language Document Structuring Conventions Specification (DSC)](https://web.archive.org/web/20190126043023/https://www.adobe.com/content/dam/acom/en/devnet/actionscript/articles/5001.DSC_Spec.pdf).
> ...
> Most importantly, an EPS file must include two required DSC header comments:
> * %!PS-Adobe-3.0 EPSF-3.0
> * '%%BoundingBox: llx lly urx ury'

src/PIL/EpsImagePlugin.py
Yay295: Yes, and Photoshop also works this way: https://github.com/python-pillow/Pillow/issues/6382#issuecomment-1213597912

src/PIL/EpsImagePlugin.py
radarhere: Ok, I see in '2.1 Required DSC Header Comments' of https://web.archive.org/web/20220120164601/https://www.adobe.com/content/dam/acom/en/devnet/actionscript/articles/5002.EPSF_Spec.pdf
> The two required DSC Header comments are
> > %!PS-Adobe-3.0 EPSF-3.0
> > %%BoundingBox: llx lly urx ury

That same PDF also includes in 3.1
> If either of the two required header comments is missing, the application should alert the user that the file is not a conforming EPS file and abort the import.

src/PIL/EpsImagePlugin.py
radarhere: Looking at https://paulbourke.net/dataformats/psdpsb/psdpsb.html, Photoshop actually originated this, so it does make some sense that Photoshop would prioritise their own comment.

> Photoshop includes a comment in the EPS files it writes so that it is able to read them back in again. Third party programs that write pixel-based EPS files may want to include this comment in their EPS files, so Photoshop can read their files.
> 
> The comment must follow immediately after the %% comment block at the start of the file. The comment is:
> 
> `%ImageData: <columns> <rows> <depth> <mode> <pad channels> <block size> <binary/hex> "<data start>"`

src/PIL/EpsImagePlugin.py
```diff
@@ -333,22 +341,38 @@ def _read_comment(s: str) -> bool:
...
+        if imagedata_size:
+            self._size = imagedata_size
+        else:
+            self._size = (
+                bounding_box[2] - bounding_box[0],
+                bounding_box[3] - bounding_box[1],
+            )
+
+        self.tile = [
+            ImageFile._Tile("eps", (0, 0) + self._size, offset, (length, bounding_box))
```
radarhere, line 373:
```suggestion
            ImageFile._Tile("eps", (0, 0) + self.size, offset, (length, bounding_box))
```
I'd like to save `self._size` for when it is being assigned to.

src/PIL/EpsImagePlugin.py
```diff
@@ -333,22 +341,38 @@ def _read_comment(s: str) -> bool:
...
             raise OSError(msg)
 
+        # An "ImageData" size takes precedence over the "BoundingBox".
+        if imagedata_size:
+            self._size = imagedata_size
+        else:
+            self._size = (
+                bounding_box[2] - bounding_box[0],
+                bounding_box[3] - bounding_box[1],
+            )
```
radarhere, line 370:
```suggestion
        self._size = imagedata_size or (
            bounding_box[2] - bounding_box[0],
            bounding_box[3] - bounding_box[1],
        )
```
This seems simpler to me.

src/PIL/EpsImagePlugin.py
Yay295: huh, I don't remember changing that. I'm not sure how that happened.

Tests/test_file_eps.py
```diff
@@ -141,23 +151,19 @@ def test_missing_boundingbox_comment(prefix: bytes) -> None:
...
-def test_invalid_boundingbox_comment(prefix: bytes) -> None:
-    data = io.BytesIO(prefix + b"\n".join(simple_eps_file_with_invalid_boundingbox))
+@pytest.mark.parametrize(
+    "file_lines",
+    (
+        simple_eps_file_with_invalid_boundingbox,
+        simple_eps_file_with_invalid_boundingbox_valid_imagedata,
+    ),
+)
+def test_invalid_boundingbox_comment(prefix: bytes, file_lines: list[bytes]) -> None:
```
radarhere, line 161:
```suggestion
def test_invalid_boundingbox_comment(
    prefix: bytes, file_lines: tuple[bytes, ...]
) -> None:
```

radarhere (approved):
My list of the changes in this PR.

1. Raise an error if the "BoundingBox" is absent, even if "ImageData" is present
2. Fix loading images where "BoundingBox" occurs after "ImageData"
3. For images without transparency, automatically pick 1, L or RGB mode

PR #8286, opened 2024-08-05
Added writing XMP bytes to JPEG

### LINKED ISSUE #8283: Add XMP writing
This is a feature request:
As a followup of [this discussion](https://github.com/python-pillow/Pillow/discussions/8269#discussioncomment-10201110) it might be possible to add a simple XMP write support using the provided infrastructure (that is by passing image.encoderinfo["extra"]). This approach doesn't account for large XMP Data (>64Kb), but would be an improvement.
 
For an discussion on XMP, especially large XMP fields, see #5076, there are also sample images linked.

To keep the interface simple, it should only accept `str` in the first iteration, later on some XML structure, requirements for a `dict` interface would need full support for the XML data model  like XML namespace support and [lists of elements with the same name](https://github.com/python-pillow/Pillow/issues/5076#issuecomment-2119000926).

See also #8069

radarhere:
Resolves #8283, allowing XMP data to be saved to JPEG files, either by
```python
im.save("out.jpg", xmp=b"test")
```
or
```python
im.info["xmp"] = b"test"
im.save("out.jpg")
```

### DIFF
--- a/src/PIL/JpegImagePlugin.py
+++ b/src/PIL/JpegImagePlugin.py
@@ -747,7 +747,7 @@ def _save(im: Image.Image, fp: IO[bytes], filename: str | bytes) -> None:
     extra = info.get("extra", b"")
 
     MAX_BYTES_IN_MARKER = 65533
-    xmp = info.get("xmp")
+    xmp = info.get("xmp", im.info.get("xmp"))
     if xmp:
         overhead_len = 29  # b"http://ns.adobe.com/xap/1.0/\x00"
         max_data_bytes_in_marker = MAX_BYTES_IN_MARKER - overhead_len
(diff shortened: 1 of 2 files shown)


cmahnke:
Thanks!
But please keep in mind that there is a size limit of about 64Kb on this field, see [Spec](https://github.com/adobe/XMP-Toolkit-SDK/blob/main/docs/XMPSpecificationPart3.pdf), page 14. How about raising an error if it's exceeded?

src/PIL/JpegImagePlugin.py
```diff
@@ -746,23 +746,28 @@ def validate_qtables(
 
     extra = info.get("extra", b"")
 
+    MAX_BYTES_IN_MARKER = 65533
     xmp = info.get("xmp")
     if xmp:
-        size = o16(31 + len(xmp))
+        overhead_len = 29
```
cmahnke, line 752: Maybe add a comment that this is `len('http://ns.adobe.com/xap/1.0/'+ b"\0")`? Otherwise a reader not famliar with the matter might come to the conclusion that this is just a(nother) magic number.

PR #8289, opened 2024-08-06
Move auxiliary mypy requirements to a requirements file

akx:
For easier installation of those requirements outside tox.

### DIFF
--- a/.ci/requirements-mypy.txt
+++ b/.ci/requirements-mypy.txt
@@ -1,1 +1,10 @@
 mypy==1.11.1
+IceSpringPySideStubs-PyQt6
+IceSpringPySideStubs-PySide6
+ipython
+numpy
+packaging
+pytest
+types-defusedxml
+types-olefile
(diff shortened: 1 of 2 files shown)


radarhere:
And the reason that you're not just moving these into .ci/requirements-mypy.txt is in case someone wants to use a type-checking system other than mypy?

That might be reasonable, sure, I just hope that we're not going to start down the road of trying to satisfy more than one type-checking tool at the same time.

akx:
> And the reason that you're not just moving these into .ci/requirements-mypy.txt is in case someone wants to use a type-checking system other than mypy?

No, it was because I wasn't sure if there was a reason for these to have been in the toxfile to begin with 😅 

I can absolutely put them in `requirements-mypy.txt` too!

akx:
Moved 'em all to `requirements-mypy.txt`.

radarhere:
Thanks

PR #8290, opened 2024-08-06
Parametrize some color_lut tests for DRYness

akx:
Refs #8279.

These tests were doing the same repeated call with slightly differing arguments, so might as well use `parametrize` for less repetition (and so #8279 wouldn't need as many `assert`s).

I can rebase this afterwards if #8279 should get merged first :)


radarhere (approved):

hugovk:
Thanks!

PR #8295, opened 2024-08-07
Enforce CSV names style for `pytest.mark.parametrize`

akx:
As mentioned by @radarhere in 
https://github.com/akx/Pillow/pull/7, Pillow's `pytest.mark.parametrize` house style is `csv` – this configures Ruff to enforce that and fixes the stray tuples.

* Ruff rule reference: https://docs.astral.sh/ruff/rules/pytest-parametrize-names-wrong-type/

### DIFF
--- a/Tests/test_file_dds.py
+++ b/Tests/test_file_dds.py
@@ -152,7 +152,7 @@ def test_sanity_ati2_bc5u(image_path: str) -> None:
 
 
 @pytest.mark.parametrize(
-    ("image_path", "expected_path"),
+    "image_path, expected_path",
     (
         # hexeditted to be typeless
         (TEST_FILE_DX10_BC5_TYPELESS, TEST_FILE_DX10_BC5_UNORM),
@@ -248,7 +248,7 @@ def test_dx10_r8g8b8a8_unorm_srgb() -> None:
 
 
 @pytest.mark.parametrize(
-    ("mode", "size", "test_file"),
+    "mode, size, test_file",
     [
         ("L", (128, 128), TEST_FILE_UNCOMPRESSED_L),
         ("LA", (128, 128), TEST_FILE_UNCOMPRESSED_L_WITH_ALPHA),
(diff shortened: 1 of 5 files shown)


Tests/test_file_dds.py
```diff
@@ -152,7 +152,7 @@ def test_sanity_ati2_bc5u(image_path: str) -> None:
 
 
 @pytest.mark.parametrize(
-    ("image_path", "expected_path"),
+    "image_path,expected_path",
```
hugovk, line 155: The suggestion in https://github.com/akx/Pillow/pull/7 is with spaces after the comma:

> * Replaced `("lut_mode", "table_size")` with `"lut_mode, table_size"`. It seems simpler to me, and is [the style that Pillow is using elsewhere](https://github.com/python-pillow/Pillow/blob/11b4df3ff9c02757eaf687ac13374f3525c8cb71/Tests/test_file_tiff_metadata.py#L195)

Can the formatter do it like that too?

```suggestion
    "image_path, expected_path",
```

akx: The autofix would not (did not) add a space, but the linter check is fine with strings that have or don't have spaces.

Re-pushed with spaces :)

radarhere (approved):

pyproject.toml
```diff
@@ -129,6 +130,8 @@ lint.per-file-ignores."Tests/oss-fuzz/fuzz_font.py" = [
 lint.per-file-ignores."Tests/oss-fuzz/fuzz_pillow.py" = [
   "I002",
 ]
+lint.flake8-pytest-style.parametrize-names-type = "csv"
+
 lint.isort.known-first-party = [
```
radarhere, line 135:
```suggestion
lint.flake8-pytest-style.parametrize-names-type = "csv"
lint.isort.known-first-party = [
```
Nitpick

akx: A linter forced the order I committed.

radarhere: I wasn't suggesting the order be changed, simply removing the blank line in the middle.

PR #8299, opened 2024-08-11
SIMD: AlphaComposite SSE4 & AVX2

homm:
Depends on #8209.

This is porting of AlphaComposite acceleration from [Pillow-SIMD](https://pypi.org/project/Pillow-SIMD/).

This PR includes required changes from #8209.


radarhere:

PR #8300, opened 2024-08-11
SIMD: BoxBlur SSE4

homm:
Depends on https://github.com/python-pillow/Pillow/pull/8209.

This is porting of BoxBlur acceleration from [Pillow-SIMD](https://pypi.org/project/Pillow-SIMD/).

This PR includes required changes from https://github.com/python-pillow/Pillow/pull/8209.


radarhere:

PR #8301, opened 2024-08-11
SIMD: Filter SSE4 & AVX2

homm:
Depends on https://github.com/python-pillow/Pillow/pull/8209.

This is porting of Filter acceleration from [Pillow-SIMD](https://pypi.org/project/Pillow-SIMD/).

This PR includes required changes from https://github.com/python-pillow/Pillow/pull/8209.

### DIFF
--- a/src/libImaging/Filter.c
+++ b/src/libImaging/Filter.c
@@ -26,6 +26,7 @@
 
 #include "Imaging.h"
 
+#if defined(__SSE4__)
 /* 5 is number of bits enought to account all kernel coefficients (1<<5 > 25).
    8 is number of bits in result.
    Any coefficients delta smaller than this precision will have no effect. */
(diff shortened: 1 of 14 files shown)


src/libImaging/Filter.c
```diff
@@ -26,6 +26,24 @@
 
 #include "Imaging.h"
 
+#if defined(__SSE4__)
+/* 5 is number of bits enought to account all kernel coefficients (1<<5 > 25).
+   8 is number of bits in result.
+   Any coefficients delta smaller than this precision will have no effect. */
+#define PRECISION_BITS (8 + 5)
+/* 16 is number of bis required for kernel storage.
```
radarhere, line 34:
```suggestion
/* 5 is enough bits to account for all kernel coefficients (1<<5 > 25).
   8 is the number of bits in the result.
   Any coefficients delta smaller than this precision will have no effect. */
#define PRECISION_BITS (8 + 5)
/* 16 is the number of bits required for kernel storage.
```

radarhere:

PR #8309, opened 2024-08-16
Lint: run more flake8-pytest-style

hugovk:
Expand from PT006 to all of PT:

https://docs.astral.sh/ruff/rules/#flake8-pytest-style-pt

Fix warnings for PT001, PT014 and PT018.

Exclude these which have warnings:

*  PT007 pytest-parametrize-values-wrong-type
*  PT011 pytest-raises-too-broad
*  PT012 pytest-raises-with-multiple-statements
*  PT016 pytest-fail-without-message
*  PT017 pytest-assert-in-except

I'm fine if someone wants to tackle those in another PR.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,12 +1,12 @@
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.5.6
+    rev: v0.6.0
     hooks:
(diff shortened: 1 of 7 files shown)


hugovk:
Changed my mind about PT001, let's skip it.

radarhere (approved):

PR #8310, opened 2024-08-16
Move repeated code to private helper function

Yay295:


radarhere (approved):

src/PIL/TiffImagePlugin.py
```diff
@@ -2080,38 +2080,33 @@ def readLong(self) -> int:
         (value,) = struct.unpack(self.longFmt, self.f.read(4))
         return value
 
+    def _verify_bytes_written(self, bytes_written: int | None, expected: int) -> None:
```
hugovk, line 2083: We don't use `self` in this method, would it be better to make it a non-class function or static method instead?

Yay295: I think a static method would work here. That keeps the function definition close to its usage, and associated with the `AppendingTiffWriter` class instead of any of the other classes in that file.

hugovk (approved):

PR #8313, opened 2024-08-18
Add feature scale factor resizing

### LINKED ISSUE #8312: Scale factor for resizing
Hi!
Often working with images it is common to specify the scale factor to downsample/upsample the images preserving the aspect ratio.
I was wondering if it could be possible to include a scale factor as input param of the _resize_ method.

Today I achieve this by doing the following:

```python
from PIL import Image

image = Image.open('example.jpg')
scale_factor = 0.5  # For example, reduce the size by half
# Calculate the new size
new_size = (
    int(image.width * scale_factor),
    int(image.height * scale_factor)
)
resampled_image = image.resize(new_size)
```

Perhaps, it could be interesting to have this directly  inside the resize method.
I leave a simple PR in case you like the addition.
Wdyt?

enomis-dev:
Resolves #8312

 * Scale factor in the resize method to preserve image ratio as alternative to specify width and height.

### DIFF
--- a/Tests/test_image_resize.py
+++ b/Tests/test_image_resize.py
@@ -292,10 +292,7 @@ class TestImageResize:
 
         def resize_scale(mode: str, scale: float) -> None:
             img = hopper(mode)
-            size = (
-                int(self.width * scale),
-                int(self.height * scale)
-            )
+            size = (int(self.width * scale), int(self.height * scale))
             out = img.resize(scale=scale)
             assert out.mode == mode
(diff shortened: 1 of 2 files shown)


radarhere:
Discussion in #8312 suggests that [`Image.reduce`](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.reduce) might already supply this feature.

enomis-dev:
Closed as discussed in #8312

PR #8316, opened 2024-08-19
Add more data to error message

Yay295:
This changes the error message from
`offset is neither short nor long`
to
`offset is neither short nor long: Tag ID 273 Field Type 16 Field Size 8 Count 1`


radarhere:
Could you give an example of how to trigger this error?

Yay295:
With #8317, `test_bigtiff()` in `test_file_tiff.py` triggers this error if `del im.tag_v2[273]` is removed from the test.

radarhere:
Would it not be more consistent with the rest of TiffImagePlugin to log the extra information?

https://github.com/python-pillow/Pillow/blob/eeb3d04843996117457af2481ca242f97e646b5e/src/PIL/TiffImagePlugin.py#L1429-L1435

Yay295:
It might. Logging is only used in five files in Pillow though (Image, PcxImagePlugin, PngImagePlugin, TiffImagePlugin, helper).

radarhere:
> With #8317, `test_bigtiff()` in `test_file_tiff.py` triggers this error if `del im.tag_v2[273]` is removed from the test.

Just a note: this stopped being the case after #8417 added support for LONG8.

This error might still occur, but the tag would have to have a type of UNDEFINED, SBYTE, ASCII, BYTE or 0.

radarhere:
I've created https://github.com/python-pillow/Pillow/pull/8568 as an alternative to this, using my suggestion of logging.

radarhere:
#8568 has been merged instead.

Issue #7077, opened 2023-04-09
Saving TIFF stack does not support tags with type long8


stilley2:
Title pretty much says it all. 

I believe the problem is that this list doesn't contain long8: [here](https://github.com/python-pillow/Pillow/blob/4ffbbe194c5a1b8840f809574017ab5f1333695f/src/PIL/TiffImagePlugin.py#L1881)

to reproduce
```python
import numpy as np
import PIL
from PIL.TiffImagePlugin import ImageFileDirectory_v2
x = np.ones((2, 10, 10))
pilimgs = [PIL.Image.fromarray(i) for i in x]
ifd = ImageFileDirectory_v2()
ifd[256] = 10
ifd[257] = 10
ifd[65000] = 1234
ifd.tagtype[65000] = 16
pilimgs[0].save("testimgs.tif", save_all=True, tiffinfo=ifd,
                append_images=pilimgs[1:], format="tiff")
```

radarhere:
I've created PR #7078 to resolve this.

PR #8317, opened 2024-08-19
Fixed writing multiple StripOffsets to TIFF

Yay295:
The StripOffsets values don't really mean anything without associated image data, so I just did the same thing as when this tag only has one value and added the IFD offset to the values.

StripOffsets is overwritten with correct values when writing TIFF images, so this only matters when writing a non-TIFF image, where StripOffsets doesn't mean anything. The main benefit of this change is that it won't raise an exception now when saving a non-TIFF image with EXIF data copied from a TIFF image that has more than one strip.

### DIFF
--- a/Tests/test_file_tiff.py
+++ b/Tests/test_file_tiff.py
@@ -108,7 +108,8 @@ class TestFileTiff:
             assert_image_equal_tofile(im, "Tests/images/hopper.tif")
 
         with Image.open("Tests/images/hopper_bigtiff.tif") as im:
-            # multistrip support not yet implemented
+            # The data type of this file's StripOffsets tag is LONG8,
+            # which is not yet supported for offset data when saving multiple frames.
             del im.tag_v2[273]
 
             outfile = str(tmp_path / "temp.tif")
--- a/Tests/test_file_tiff_metadata.py
+++ b/Tests/test_file_tiff_metadata.py
@@ -181,6 +181,28 @@ def test_change_stripbytecounts_tag_type(tmp_path: Path) -> None:
         assert reloaded.tag_v2.tagtype[TiffImagePlugin.STRIPBYTECOUNTS] == TiffTags.LONG
 
 
+def test_save_multiple_stripoffsets() -> None:
+    ifd = TiffImagePlugin.ImageFileDirectory_v2()
+    ifd[TiffImagePlugin.STRIPOFFSETS] = (123, 456)
+    assert ifd.tagtype[TiffImagePlugin.STRIPOFFSETS] == TiffTags.LONG
+
+    # all values are in little-endian
+    assert ifd.tobytes() == (
+        # number of tags == 1
+        b"\x01\x00"
+        # tag id (2 bytes), type (2 bytes), count (4 bytes), value (4 bytes)
+        # == 273, 4, 2, 18
+        # == TiffImagePlugin.STRIPOFFSETS, TiffTags.LONG, 2, 18
+        # the value is the index of the tag data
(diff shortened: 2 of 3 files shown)


Tests/test_file_tiff.py
```diff
@@ -108,7 +108,8 @@ def test_bigtiff(self, tmp_path: Path) -> None:
             assert_image_equal_tofile(im, "Tests/images/hopper.tif")
 
         with Image.open("Tests/images/hopper_bigtiff.tif") as im:
-            # multistrip support not yet implemented
+            # The data type of this file's StripOffsets tag is LONG8,
+            # which is not currently supported when saving.
```
radarhere, line 112:
```suggestion
            # which is not yet supported as an offset when saving multiple frames.
```
The current version makes it sound like #7077 wasn't resolved, which is when this test was created in the first place.

src/PIL/TiffImagePlugin.py
```diff
@@ -991,9 +991,11 @@ def tobytes(self, offset: int = 0) -> bytes:
         if stripoffsets is not None:
             tag, typ, count, value, data = entries[stripoffsets]
             if data:
-                msg = "multistrip support not yet implemented"
-                raise NotImplementedError(msg)
-            value = self._pack("L", self._unpack("L", value)[0] + offset)
+                size, handler = self._load_dispatch[typ]
+                values = [val + offset for val in handler(self, data, self.legacy_api)]
```
radarhere, line 995:
```suggestion
                values = [val + offset for val in handler(self, data)]
```
Throwing this suggestion out there, see what happens - considering that this should be [SHORT or LONG](https://web.archive.org/web/20230619074139/https://www.awaresystems.be/imaging/tiff/tifftags/stripoffsets.html) (or LONG8), it doesn't matter whether the legacy API is in use or not, as they're all passed to https://github.com/python-pillow/Pillow/blob/b557876ec3aa4d5d236d8044519f1c613031fbee/src/PIL/TiffImagePlugin.py#L502-L505

Tests/test_file_tiff_metadata.py
```diff
@@ -181,6 +181,28 @@ def test_change_stripbytecounts_tag_type(tmp_path: Path) -> None:
...
+    assert ifd.tagtype[TiffImagePlugin.STRIPOFFSETS] == TiffTags.LONG
+
+    # all values are in little-endian
+    assert ifd.tobytes() == (
+        # number of tags == 1
+        b"\x01\x00"
+        # tag id (2 bytes), type (2 bytes), count (4 bytes), value (4 bytes)
+        # == 273, 4, 2, 18
+        # == TiffImagePlugin.STRIPOFFSETS, TiffTags.LONG, 2, 18
+        # the value is the index of the tag data
```
radarhere, line 196:
```suggestion
        # TiffImagePlugin.STRIPOFFSETS, TiffTags.LONG, 2, 18
        # where STRIPOFFSETS is 273, LONG is 4
        # and 18 is the offset of the tag data
```

radarhere:
> StripOffsets is overwritten with correct values when writing TIFF images

If libtiff is used when saving, [yes](https://github.com/python-pillow/Pillow/blob/b557876ec3aa4d5d236d8044519f1c613031fbee/src/PIL/TiffImagePlugin.py#L1871-L1872). Otherwise, no.

src/PIL/TiffImagePlugin.py
Yay295: I think it's more correct to pass the value since we have it. The value not being used is an implementation detail that this call shouldn't be concerned with.

radarhere (approved):

PR #8319, opened 2024-08-19
Added type hints

radarhere:


src/PIL/PSDraw.py
```diff
@@ -28,15 +28,12 @@
...
+    ``sys.stdout.buffer`` is assumed.
     """
 
-    def __init__(self, fp=None):
+    def __init__(self, fp: IO[bytes] | None = None) -> None:
         if not fp:
-            try:
-                fp = sys.stdout.buffer
-            except AttributeError:
-                fp = sys.stdout
```
radarhere, line 39: When mypy looks at `sys.stdout` here, it says
> expression has type "Union[TextIO, Any]"

I think #5437 misinterpreted the docs, and that really, `sys.stdout` will not accept the bytes that we are sending.

https://docs.python.org/3/library/sys.html#sys.stdout
> be aware that the standard streams may be replaced with file-like objects like [io.StringIO](https://docs.python.org/3/library/io.html#io.StringIO) which do not support the buffer attribute.

docs/reference/Image.rst
```diff
@@ -363,6 +363,7 @@ Classes
     :show-inheritance:
 .. autoclass:: PIL.Image.ImagePointHandler
 .. autoclass:: PIL.Image.ImageTransformHandler
+.. autoclass:: PIL.Image._E
```
hugovk, line 366: Do we want this "empty" internal class in the docs?

image

https://pillow--8319.org.readthedocs.build/en/8319/reference/Image.html#PIL.Image._E

radarhere: It's referenced from the type hint of https://pillow--8319.org.readthedocs.build/en/8319/reference/Image.html#PIL.Image.Image.point

hugovk: For a user-facing class, can we give it a description or a better name?

radarhere: I've added a description, and deprecated it in favour of a new name, `ImagePointTransform`.

src/PIL/Image.py
```diff
@@ -1928,10 +1936,10 @@ def point(self, data):
                 # check if the function can be used with point_transform
                 # UNDONE wiredfool -- I think this prevents us from ever doing
                 # a gamma function point transform on > 8bit images.
-                scale, offset = _getscaleoffset(lut)
+                scale, offset = _getscaleoffset(lut)  # type: ignore[arg-type]
                 return self._new(self.im.point_transform(scale, offset))
             # for other modes, convert the function to a table
-            flatLut = [lut(i) for i in range(256)] * self.im.bands
+            flatLut = [lut(i) for i in range(256)] * self.im.bands  # type: ignore[arg-type]
```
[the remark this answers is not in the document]
hugovk: It's an underscore name so "private" and we can rename it without deprecation. I see very little use on GitHub:

https://github.com/search?q=%2F%5CbImage%5C._E%5Cb%2F+language%3APython&ref=opensearch&type=code

https://github.com/search?q=language%3APython+%2Ffrom+Image+import.*_E%5Cb%2F&ref=opensearch&type=code

But I don't mind being cautious and deprecating first if you prefer.

hugovk (approved):

PR #8324, opened 2024-08-23
If bitmap buffer is empty, do not render anything

### LINKED ISSUE #8272: Bitmap missing for glyph
### What did you do?

Tried to call `ImageDraw.text((15, 36), "A B")` with the following font:
[FS Lola Medium Regular.zip](https://github.com/user-attachments/files/16443570/FS.Lola.Medium.Regular.zip)

Either a successful call, since that font is rendered correctly on browsers and as a system font, or the usage of one of the 'unidentified characters' (□, �).

### What actually happened?

Because of the whitespace, the call fails with a `OSError: Bitmap missing for glyph` exception on `PIL/ImageFont.py:607`. I'm not exactly sure if there's any actual error with the font file, but as mentioned, it works fine on browsers and as a system font.

* OS: MacOS 13.3.1, Ubuntu 20.04
* Python: 3.8
* Pillow: 10.4

```text
Pillow 10.4.0
Python 3.8.18 (default, Nov  2 2023, 16:50:59)
       [Clang 14.0.3 (clang-1403.0.22.14.1)]
Python executable is /Users/joaohaas/dev/server/env/bin/python3
Environment Python files loaded from /Users/joaohaas/dev/server/env
System Python files loaded from /opt/homebrew/Cellar/python@3.8/3.8.18_1/Frameworks/Python.framework/Versions/3.8
Python Pillow modules loaded from /Users/joaohaas/dev/server/env/lib/python3.8/site-packages/PIL
Binary Pillow modules loaded from /Users/joaohaas/dev/server/env/lib/python3.8/site-packages/PIL
--- PIL CORE support ok, compiled for 10.4.0
--- WEBPMUX support ok
--- WEBP Animation support ok
--- RAQM (Bidirectional Text) support ok, loaded 0.10.1, fribidi 1.0.14, harfbuzz 8.5.0
```

```python

font = ImageFont.truetype("<path>/FS Lola Medium Regular.ttf", size=27)
image = Image.new("RGB", (440, 58), (255, 255, 255))
draw_object = ImageDraw.Draw(image)
draw_object.text((15, 36), "A B", font=font)
```

radarhere:
Resolves #8272

#6846 added an error if the bitmap buffer is empty.
https://github.com/python-pillow/Pillow/blob/d6cfebd016db25549d05a9c5caa2ef3b53cff5c5/src/_imagingft.c#L1028-L1032

However, this new issue found the buffer is empty for a space character in a particularly font, which seems like a realistic scenario, and has been confirmed as valid by one of the FreeType maintainers - https://github.com/python-pillow/Pillow/issues/8272#issuecomment-2305870693

So if the bitmap buffer is empty, this PR just doesn't draw anything, uses the `advance`s to update the position for the next character, and moves on.

### DIFF
--- a/Tests/test_font_crash.py
+++ b/Tests/test_font_crash.py
@@ -1,10 +1,8 @@
 from __future__ import annotations
 
-import pytest
-
 from PIL import Image, ImageDraw, ImageFont
 
-from .helper import skip_unless_feature
+from .helper import skip_unless_feature_version
 
 
 class TestFontCrash:
@@ -17,8 +15,7 @@ class TestFontCrash:
             draw.multiline_textbbox((10, 10), "ABC\nAaaa", font, stroke_width=2)
             draw.text((10, 10), "Test Text", font=font, fill="#000")
 
-    @skip_unless_feature("freetype2")
+    @skip_unless_feature_version("freetype2", "2.12.0")
     def test_segfault(self) -> None:
-        with pytest.raises(OSError):
-            font = ImageFont.truetype("Tests/fonts/fuzz_font-5203009437302784")
-            self._fuzz_font(font)
+        font = ImageFont.truetype("Tests/fonts/fuzz_font-5203009437302784")
+        self._fuzz_font(font)
--- a/src/_imagingft.c
+++ b/src/_imagingft.c
@@ -1025,130 +1025,99 @@ font_render(FontObject *self, PyObject *args) {
             yy = -(py + glyph_slot->bitmap_top);
         }
 
-        // Null buffer, is dereferenced in FT_Bitmap_Convert
-        if (!bitmap.buffer && bitmap.rows) {
-            PyErr_SetString(PyExc_OSError, "Bitmap missing for glyph");
-            goto glyph_error;
-        }
-
-        /* convert non-8bpp bitmaps */
-        switch (bitmap.pixel_mode) {
-            case FT_PIXEL_MODE_MONO:
-                convert_scale = 255;
-                break;
-            case FT_PIXEL_MODE_GRAY2:
-                convert_scale = 255 / 3;
-                break;
-            case FT_PIXEL_MODE_GRAY4:
-                convert_scale = 255 / 15;
-                break;
-            default:
-                convert_scale = 1;
-        }
-        switch (bitmap.pixel_mode) {
-            case FT_PIXEL_MODE_MONO:
-            case FT_PIXEL_MODE_GRAY2:
-            case FT_PIXEL_MODE_GRAY4:
-                if (!bitmap_converted_ready) {
-                    FT_Bitmap_Init(&bitmap_converted);
-                    bitmap_converted_ready = 1;
-                }
-                error = FT_Bitmap_Convert(library, &bitmap, &bitmap_converted, 1);
-                if (error) {
-                    geterror(error);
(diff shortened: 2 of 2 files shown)


Tests/test_font_crash.py
```diff
@@ -17,8 +15,7 @@ def _fuzz_font(self, font: ImageFont.FreeTypeFont) -> None:
             draw.multiline_textbbox((10, 10), "ABC\nAaaa", font, stroke_width=2)
             draw.text((10, 10), "Test Text", font=font, fill="#000")
 
-    @skip_unless_feature("freetype2")
+    @skip_unless_feature_version("freetype2", "2.12.0")
```
nulano, line 18: Why change the version here?

nulano:
We might want to add a test that triggers the `if (stroker != NULL)` branch in `font_render` to make sure that passing such a glyph to the stroker API is safe.

Tests/test_font_crash.py
radarhere: The test intended to catch `OSError: Bitmap missing for glyph` from https://github.com/python-pillow/Pillow/blob/b6f90c47325f4d6cb507269b51cf309319eca8d7/src/_imagingft.c#L1028-L1032

For FreeType2 versions before 2.12.0, `FT_New_Face` is returning an `Unknown_File_Format` error. It does that even on main at the moment, it's just not obvious because both are `OSError`. If I [increase the specificity of the test](https://github.com/radarhere/Pillow/commit/8663c46d7d1aedd199130068539df5575f36d414), [you can see this.](https://github.com/radarhere/Pillow/actions/runs/10536876212)

FreeType2 2.12.0 was released in March 2022, before #6846, so I expect this has been the case the whole time. Now that I've fixed the `OSError: Bitmap missing for glyph` error, the other error becomes obvious.

This was a font generated by OSS-Fuzz, so its purpose is not to be valid, but merely to trigger a security problem. We could modify the file to be valid, but in principle it would seem to be better to not modify test scenarios over time.

I suppose I could modify the test to run on FreeType2 < 2.12.0 as well, and just catch the error for that scenario if you want.

radarhere:
You're concerned about passing a glyph where bitmap buffer is empty to [this block](https://github.com/python-pillow/Pillow/blob/b6f90c47325f4d6cb507269b51cf309319eca8d7/src/_imagingft.c#L1003-L1021)? [`test_bitmap_font_stroke`](https://github.com/python-pillow/Pillow/blob/b6f90c47325f4d6cb507269b51cf309319eca8d7/Tests/test_imagefont.py#L880) actually already does this.

apodtele:
Note then the glyph outline in question is not empty but degenerate. It contains two points but still produces a blank space. In this particular case, it results in a 0×1 bitmap, aka one empty row. Do not check for `bitmap.rows`, or check both dimentions

radarhere:
The glyph outline may not be empty, but the bitmap buffer is. From my understanding, we're safe to skip drawing and just advance if that is the case. I presume the size of the glyph does not affect the advance.

apodtele:
Correct, just advance if the buffer is NULL regardless of its dimensions. You have sufficient error checking from FreeType calls.

Tests/test_font_crash.py
nulano: Thank you for the explanation.

PR #8338, opened 2024-08-30
Improve ImageFont error messages

yngvem:
I spent a longer than I'd like to admit trying to figure out why my fonts wouldn't load before I realised that I was using the wrong function. This PR should hopefully help someone else realise their mistake faster.

 * Add more information in the font loading exception messages
 * Add links between `ImageFont.load` and `ImageFont.truetype` in the documentation

  

**Documentation update:**
image
and similar for truetype:
image


Tests/test_imagefont.py
```diff
@@ -460,6 +460,17 @@ def test_free_type_font_get_mask(font: ImageFont.FreeTypeFont) -> None:
     assert mask.size == (108, 13)
 
 
+def test_load_raises_if_image_not_found(tmp_path) -> None:
```
radarhere, line 463:
```suggestion
def test_load_raises_if_image_not_found(tmp_path: Path) -> None:
```

Tests/test_imagefont.py
```diff
@@ -471,6 +482,22 @@ def test_load_path_not_found() -> None:
         ImageFont.truetype(filename)
 
 
+def test_load_path_exisitng_path(tmp_path) -> None:
```
radarhere, line 485:
```suggestion
def test_load_path_exisitng_path(tmp_path: Path) -> None:
```

Tests/test_imagefont.py
```diff
@@ -471,6 +482,22 @@ def test_load_path_not_found() -> None:
...
+    # First, the file doens't exist, so we don't suggest `load`
```
radarhere, line 486:
```suggestion
    # First, the file doesn't exist, so we don't suggest `load`
```

src/PIL/ImageFont.py
```diff
@@ -224,7 +228,7 @@ def __init__(
             raise core.ex
 
         if size <= 0:
-            msg = "font size must be greater than 0"
+            msg = f"font size must be greater than 0, not {size}"
```
radarhere, line 231: Was this also a source of confusion for you? I'm not sure why this wouldn't be suitably obvious.

yngvem: No this happened more on auto-pilot while coding in the evening -- I pretty much always try to include the incorrect value in error messages if it fails on input-validation as I find it makes debugging quite a lot easier for me (especially if the error is several frames down the stack from where I'm working).

Feel free to remove it, no hard feelings there :)

radarhere:
I've created https://github.com/yngvem/Pillow/pull/1 with some suggestions.

yngvem:
I just merged the changes now

yngvem:
I'm unsure why the pipeline crashed, but it seems to be unrelated to the changes in this PR? Maybe just unlucky with the load on the node that was assigned to this pipeline run?

```raw
Tests/test_file_tiff.py::TestFileTiff::test_timeout FAILED               [ 48%]
```

```raw

        if not self.map:
            # sort tiles in file order
>           self.tile.sort(key=_tilesort)
E           Failed: Timeout >6.0s

/vpy3/lib/python3.12/site-packages/PIL/ImageFile.py:245: Failed
```

https://github.com/python-pillow/Pillow/actions/runs/10656884196/job/29544000694?pr=8338

radarhere:
I've created https://github.com/yngvem/Pillow/pull/2 with some suggestions.

yngvem:
Good ideas for changes!

src/PIL/ImageFont.py
```diff
@@ -927,7 +932,10 @@ def load_path(filename: str | bytes) -> ImageFont:
             return load(os.path.join(directory, filename))
         except OSError:
             pass
-    msg = "cannot find font file"
+    msg = f"cannot find font file '{filename}' in `sys.path`"
+    if os.path.exists(filename):
+        msg += f", did you mean `ImageFont.load({filename})` instead?"
```
radarhere, line 937:
```suggestion
        msg += f", did you mean `ImageFont.load(\"{filename}\")` instead?"
```

radarhere (approved):

src/PIL/ImageFont.py
```diff
@@ -927,7 +932,10 @@ def load_path(filename: str | bytes) -> ImageFont:
             return load(os.path.join(directory, filename))
         except OSError:
             pass
-    msg = "cannot find font file"
+    msg = f"cannot find font file '{filename}' in `sys.path`"
+    if os.path.exists(filename):
+        msg += f', did you mean `ImageFont.load("{filename}")` instead?'
```
hugovk, line 937: Looks good, thanks!

I think we can skip these backticks, it's understandable without them:

```suggestion
    msg = f"cannot find font file '{filename}' in sys.path"
    if os.path.exists(filename):
        msg += f', did you mean ImageFont.load("{filename}") instead?'
```

src/PIL/ImageFont.py
```diff
@@ -927,7 +932,10 @@ def load_path(filename: str | bytes) -> ImageFont:
             return load(os.path.join(directory, filename))
         except OSError:
             pass
-    msg = "cannot find font file"
+    msg = f"cannot find font file '{filename}' in sys.path"
```
radarhere, line 935:
```suggestion
    msg = f'cannot find font file "{filename}" in sys.path'
```
I've realised that both this and the second half of the message use quotes, and I think it's better if they use the same type.

You might ask why using double quotes in `ImageFont.load("...")` is good. I think you will gather from our codebase that this is the style we have chosen.

src/PIL/ImageFont.py
```diff
@@ -804,7 +808,8 @@ def truetype(
 ) -> FreeTypeFont:
     """
     Load a TrueType or OpenType font from a file or file-like object,
-    and create a font object.
+    and create a font object. For loading bitmap fonts instead,
+    see :py:func:`~PIL.ImageFont.load` and :py:func:`~PIL.ImageFont.load_path`.
     This function loads a font object from the given file or file-like
     object, and creates a font object for a font of the given size.
```
radarhere, line 814:
```suggestion
    and create a font object. This function loads a font object from the given
    file or file-like object, and creates a font object for a font of the given
    size. For loading bitmap fonts instead, see :py:func:`~PIL.ImageFont.load`
    and :py:func:`~PIL.ImageFont.load_path`.
```
Very hopefully my last nitpick - I think the links should be at the end of the paragraph, like in `load()`, so this swaps the sentences.

hugovk:
Thank you!

PR #8340, opened 2024-09-01
Imaging as shared lib

homm:


aclark4life:
Interesting!

homm:
@aclark4life Not sure the work will be finished ) Right now I just need various test environments from this PR.

Yay295:
Is there any way in the build we can declare the minimum C version? `pyproject.toml` has `requires-python = ">=3.9"` for the Python version, but I don't see a setting that can be used for C. NumPy uses Meson, so their minimum C version is declared in the `meson.build` file.

homm:
Build with `python ./setup.py develop --pillow-configuration=parallel=1`.

Currently `_imaging` extension contain batch very useful functions and types which are not available in another extensions (like `_webp` or `_imagingcms`). As a result, the last extensions don't use useful functions from `_imaging` (for example, `_webp` the only extension which use `ImagingSectionEnter/Leave`) and don't use any type checking. This is typical code from extensions which use `Imaging` type:

```c
static PyObject *
cms_transform_apply(CmsTransformObject *self, PyObject *args) {
    Py_ssize_t idIn;
    Py_ssize_t idOut;
    Imaging im;
    Imaging imOut;

    int result;

    if (!PyArg_ParseTuple(args, "nn:apply", &idIn, &idOut)) {
        return NULL;
    }

    im = (Imaging)idIn;
    imOut = (Imaging)idOut;

    result = pyCMSdoTransform(im, imOut, self->transform);

    return Py_BuildValue("i", result);
}
```

So it's just converts arbitrary `Py_ssize_t` to pointer, which is not safe. What I'm trying to achieve: use `_imaging` extension as a library for other extensions. Unfortunately, as I see, it only possible with changing name to `lib_imaging` since `gcc -l` always appends lib prefix and there is no way to set full lib name.

Yay295:
`_imaging` also contains a bunch of things for the imaging module though, which aren't needed elsewhere. I think it would be cleaner to leave `_imaging` as it is, and create a new library that `_imaging` depends on. Then code can be moved from `_imaging` to the new library, and the other extensions can also be changed to use the new library.

homm:
@Yay295 This involves massive extension redesign, I'm not ready for this )

Issue #4532, opened 2020-04-06
Provide a C API independent of CPython for the core functionality


saudet:
As part of the recently started [TensorFlow for Java](https://github.com/tensorflow/java) project, we are striving to make TensorFlow fully usable from Java, without depending on CPython. For this purpose, TensorFlow itself offers a supported and stable C API that we can use independently from CPython. However, the Python API of TensorFlow known as Keras uses Pillow internally to process images, and given issues like #4445, we would also need to use Pillow from Java to obtain reproducible results, ideally without going through CPython. (Pillow can be called very easily from Java by using JavaCPP, so that part is not an issue, see https://github.com/bytedeco/javacpp-presets.)

I believe that most of the code of Pillow written in C could be separated without too much trouble into modules that do not depend on CPython, but before diving into the implementation details, we would like to know if this is something the developers of Pillow would be interested in maintaining, and if so, how we could assist you in the process. Thank you for your consideration!

/cc @karllessard

wiredfool:
Pillow makes a commitment to stable public interfaces, which are defined at the Python layer. The C interfaces are explicitly internal, and no effort is made to keep them stable even between minor releases, and no support or warning is given when they change.  (IIRC, there are/were people using the internal interfaces, but they've always been explicitly on their own with that) We go to some lengths to prevent the use of different versions of the Python and C layers, so that we can ensure that we're internally consistent in the face of potentially flakey install histories.

This proposal would effectively add another public interface deep into the C layer. I'm not in favor of that for a few reasons:
1) The Pillow C code has gone through a bunch of fuzzing and hardening assuming the current interface, this would open a second interface where we would have to do parameter validation. 
2) As I work through the security reports, I'm not at all sure that we should be keeping some of the more complicated parsing portions in straight C, rather we should probably be rewriting them in a safe language. (e.g., Python). Right now, the only constraint to do that is the will to do it. 
3) It locks us in to some implementation details that while they have been stable for a long time, are becoming increasingly constraining and will need to be changed to support newer image formats. (Specifically, > 4 Channel images, or Multi-Channel >8Bit/channel images)

radarhere:
My concern would be about how much functionality we would be committing to setting up independently in the C layer, and whether this would cause us to move larger amounts of code out of Python and into C.

radarhere:
Closing due to lack of interest

PR #8341, opened 2024-09-01
Use ImagingCore.ptr instead of ImagingCore.id

homm:
Partially suppress #8340

The non-primary extensions in Pillow don't have direct access to Python's `_imaging` types, such as `Imaging_Type` and `ImagingObject`. However, they often still need to interact with objects passed from Python code. Currently, the raw memory pointer to an `Imaging` object (`ImageCore.id`) is passed as an integer and then cast to a memory pointer:

```c
static PyObject *
cms_transform_apply(CmsTransformObject *self, PyObject *args) {
    Py_ssize_t idIn, idOut;
    Imaging im, imOut;

    if (!PyArg_ParseTuple(args, "nn:apply", &idIn, &idOut)) {
        return NULL;
    }

    im = (Imaging)idIn;
    imOut = (Imaging)idOut;

    return Py_BuildValue("i", pyCMSdoTransform(im, imOut, self->transform));
}
```

It's not safe and can lead to segmentation faults. Python has Capsule objects, which provide a safer way to pass pointers to C code. These are exposed through `ImageCore.ptr`.

This PR deprecates the raw pointer properties `ImageCore.id` and `ImageCore.unsafe_ptrs`, and migrates non-primary extensions to interact with Capsule objects.


homm:
@python-pillow/pillow-team I need help here)

Could anyone debug pypy3.10 on windows? I can't get what's wrong, since as I can see, the capsule object is correctly restored from the pointer and the capsule object itself is alive.

homm:
Get it! In pypy `repr(capsule)` holds the capsuled pointer itself, not the pointer to the `PyObject`. And even `hex(id(capsule))` is not the pointer to `PyObject`. Lucky, we can reliably distinguish pypy capsule from cpython capsule by absence of trailing "<>".

https://github.com/pypy/pypy/blob/branches/py3.10/pypy/objspace/std/capsuleobject.py#L39

radarhere:
So is the motivation here just to simplify the code?

Also, we don't necessarily need the deprecations - https://github.com/python-pillow/Pillow/issues/4532#issuecomment-618255798
> Pillow makes a commitment to stable public interfaces, which are defined at the Python layer. The C interfaces are explicitly internal, and no effort is made to keep them stable even between minor releases, and no support or warning is given when they change. (IIRC, there are/were people using the internal interfaces, but they've always been explicitly on their own with that)

homm:
> So is the motivation here just to simplify the code?

Added description to PR.

> Also, we don't necessarily need the deprecations

While `ImageCore` is internal Pillow object, it's still Python API. I believe `ImageCore.id` could be used by third-party extensions and can't be removed in a one day.

docs/deprecations.rst
```diff
@@ -136,6 +136,16 @@ Specific WebP Feature Checks
...
 
+Get Internal Pointers to Objects
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. deprecated:: 11.0.0
+
+``Image.core.ImageCore.id`` and ``Image.core.ImageCore.unsafe_ptrs``
+have been deprecated and will be removed in Pillow 12 (2025-10-15).
+They were used for obtaining raw pointers to ``ImageCore`` internals. To interact with
+C code, you can use ``Image.core.ImageCore.ptr``, which returns a ``Capsule`` object.
```
radarhere, line 147:
```suggestion
``Image.core.ImagingCore.id`` and ``Image.core.ImagingCore.unsafe_ptrs`` have been
deprecated and will be removed in Pillow 12 (2025-10-15). They were used for obtaining
raw pointers to ``ImagingCore`` internals. To interact with C code, you can use
``Image.core.ImagingCore.ptr``, which returns a ``Capsule`` object.
```
When you say 'ImageCore'... it should be 'ImagingCore', right?
https://github.com/python-pillow/Pillow/blob/cb2a0c40f3760411423d09defada95c54564317a/src/_imaging.c#L3777-L3778

docs/deprecations.rst
```diff
@@ -136,6 +136,16 @@ Specific WebP Feature Checks
...
 
+Get Internal Pointers to Objects
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. deprecated:: 11.0.0
+
+``Image.core.ImagingCore.id`` and ``Image.core.ImagingCore.unsafe_ptrs`` have been
+deprecated and will be removed in Pillow 12 (2025-10-15). They were used for obtaining
+raw pointers to ``ImagingCore`` internals. To interact with C code, you can use
+``Image.core.ImagingCore.ptr``, which returns a ``Capsule`` object.
```
radarhere, line 147:
```suggestion
``Image.Image.getim()``, which returns a ``Capsule`` object.
```
Encourage use of our Python API?

src/PIL/ImageCms.py
```diff
@@ -349,19 +349,17 @@ def point(self, im: Image.Image) -> Image.Image:
         return self.apply(im)
 
     def apply(self, im: Image.Image, imOut: Image.Image | None = None) -> Image.Image:
-        im.load()
         if imOut is None:
             imOut = Image.new(self.output_mode, im.size, None)
```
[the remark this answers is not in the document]
homm: > it is possible for the size of an image to change after it loads

I'm considering this as bug and have patched it in our production brunch: https://github.com/uploadcare/pillow-simd/commit/1d77bfe9d939e208e3bb2b897a4209c0881eb9f1
Do you want I made a PR for this?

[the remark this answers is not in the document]
radarhere: A separate PR would be good, rather than including it here, yeah, thanks.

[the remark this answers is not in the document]
homm: Here it is: https://github.com/python-pillow/Pillow/pull/8387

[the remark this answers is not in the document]
radarhere: This has now been addressed via #8390

src/PIL/ImageTk.py
```diff
@@ -142,6 +131,8 @@ def __init__(
             self.paste(image)
 
     def __del__(self) -> None:
+        if not hasattr(self, "__photo"):
+            return
```
radarhere, line 135: I don't think this works as simply as one might imagine.

```python
class A:
	def __init__(self):
		self.photo = True
		self._photo = True
		self.__photo = True
		
		print([
			hasattr(self, "photo"), # True
			hasattr(self, "_photo"), # True
			hasattr(self, "__photo") # False
		])

A()
```

But I'm also curious - what made you think there would be a point in the lifecycle when the object didn't have `__photo`?

radarhere:
> In pypy repr(capsule) holds the capsuled pointer itself, not the pointer to the PyObject. And even hex(id(capsule)) is not the pointer to PyObject. Lucky, we can reliably distinguish pypy capsule from cpython capsule by absence of trailing "<>".

What's the advantage of using `repr(capsule)` in the first place? Why not just pass the capsule itself all the time?

src/PIL/ImageTk.py
```diff
@@ -190,10 +181,10 @@ def paste(self, im: Image.Image) -> None:
         if image.isblock() and im.mode == self.__mode:
             block = image
         else:
-            block = image.new_block(self.__mode, im.size)
+            block = Image.core.new_block(self.__mode, im.size)
```
radarhere, line 184: I imagine this change was just because you didn't like `image.new_block(self.__mode, im.size)`?

Rather than moving `new_block` in C, wouldn't it have been better for the C method to access the size on the image object instead of being passed it? Then this call could become `image.new_block(self.__mode)`.

src/PIL/ImageTk.py
homm: > what made you think there would be a point in the lifecycle when the object didn't have __photo

This happens when `ImageTk.PhotoImage()` returns `ValueError`.

```
$ pytest Tests/test_imagetk.py
===================================================== test session starts ======================================================
platform darwin -- Python 3.9.13, pytest-8.3.2, pluggy-1.5.0
Pillow 11.0.0.dev0
Python 3.9.13 (v3.9.13:6de2ca5339, May 17 2022, 11:37:23)
       [Clang 13.0.0 (clang-1300.0.29.30)]
Python executable is /Users/Master/Code/_env/env_py39/bin/python3
Environment Python files loaded from /Users/Master/Code/_env/env_py39
System Python files loaded from /Library/Frameworks/Python.framework/Versions/3.9
Python Pillow modules loaded from /Users/Master/Code/Pillow/src/PIL
Binary Pillow modules loaded from /Users/Master/Code/Pillow/src/PIL
--- PIL CORE support ok, compiled for 11.0.0.dev0
*** FREETYPE2 support not installed
*** OPENJPEG (JPEG2000) support not installed

rootdir: /Users/Master/Code/Pillow
configfile: pyproject.toml
plugins: cov-5.0.0, env-0.6.2, timeout-2.3.1, django-4.5.2, xdist-3.6.1
collected 13 items                                                                                                             

Tests/test_imagetk.py .............                                                                                      [100%]

======================================================= warnings summary =======================================================
Tests/test_imagetk.py::test_photoimage[1]
Tests/test_imagetk.py::test_photoimage[L]
Tests/test_imagetk.py::test_photoimage[P]
Tests/test_imagetk.py::test_photoimage[RGB]
Tests/test_imagetk.py::test_photoimage[RGBA]
  /Users/Master/Code/_env/env_py39/lib/python3.9/site-packages/_pytest/unraisableexception.py:85: PytestUnraisableExceptionWarning: Exception ignored in: <function PhotoImage.__del__ at 0x102963430>
  
    File "/Users/Master/Code/Pillow/src/PIL/ImageTk.py", line 134, in __del__
      name = self.__photo.name
  AttributeError: 'PhotoImage' object has no attribute '_PhotoImage__photo'
  
    warnings.warn(pytest.PytestUnraisableExceptionWarning(msg))

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
================================================ 13 passed, 5 warnings in 0.24s ================================================
```

src/PIL/ImageTk.py
homm: I’d like to clarify that the reason for this change is not a matter of personal preference. The `new_block` method doesn't pertain to an existing image but instead creates a new block based on the passed parameters. As such, it makes more sense for it to reside in the `Image.core` module rather than being a method of the image object itself.

There is no intent to modify or enhance the behavior; I was simply correcting the method's improper placement.

homm:
> What's the advantage of using repr(capsule) in the first place? Why not just pass the capsule itself all the time?

I'm not aware of TK API, but it seems it converts any arguments to strings, at least `PyImagingPhotoPut` receives arguments as `const char **argv`. So looks like there is no way to pass the capsule object itself, instead `repr(capsule)` will be passed any way. I just made this conversion more explicit.

src/PIL/ImageTk.py
homm: I've fixed `hasattr` here

src/PIL/ImageTk.py
```diff
@@ -225,17 +216,11 @@ def __init__(self, image: Image.Image | None = None, **kw: Any) -> None:
...
-        else:
-            # slow but safe way
-            kw["data"] = image.tobitmap()
-        self.__photo = tkinter.BitmapImage(**kw)
+        self.__photo = tkinter.BitmapImage(data=image.tobitmap(), **kw)
 
     def __del__(self) -> None:
+        if not hasattr(self, "__photo"):
+            return
         name = self.__photo.name
```
radarhere, line 224:
```suggestion
        try:
            name = self.__photo.name
        except AttributeError:
            return
```

homm: damn it

radarhere:
> it seems it converts any arguments to strings, at least `PyImagingPhotoPut` receives arguments as `const char **argv`. So looks like there is no way to pass the capsule object itself, instead `repr(capsule)` will be passed any way. I just made this conversion more explicit.

In CPython, sure, but from your comments, not in PyPy.

It seems simpler to note that CPython converts the capsule to a string,
rather than converting to a string and noting that PyPy doesn't convert it?

Depends on whether your thinking was already aligned with how Tk works, I guess.

homm:
> but from your comments, not in PyPy.
That isn't in my comments. In both implementations, the Capsule is converted to a string. In the case of CPython, it's a string like `<capsule object "Pillow Imaging" at 0xffff>`, while in PyPy, it's a string like `capsule object "Pillow Imaging" at 0xeeee`. Where `0xffff` is PyCapsule object address, and `0xeeee` is the pointer stored in the capsule itself.

radarhere:
> // Special case for PyPy, where the string representation of a Capsule
> // refers directly to the pointer itself, not to the PyCapsule object.

Oh, I see, I very much misunderstood this comment.

radarhere:
> Get it! In pypy `repr(capsule)` holds the capsuled pointer itself, not the pointer to the `PyObject`.

Interestingly, this prevents you from checking the capsule name against the magic.

homm:
This check is the best we can do here. It’s not as secure as PyCapsule_IsValid(), since it only verifies that the passed string is a correctly formatted string representation of the Capsule object. It’s still possible to craft a well-formatted string with the wrong pointer, which could result in a segmentation fault when calling a Tk operation. However, it’s still a significant improvement over the previous implementation, which simply obtained some number and assumed it was a pointer.

src/_imaging.c
```diff
@@ -3707,16 +3704,39 @@ _getattr_bands(ImagingObject *self, void *closure) {
 
 static PyObject *
 _getattr_id(ImagingObject *self, void *closure) {
+    if (PyErr_WarnEx(
+            PyExc_DeprecationWarning,
+            "id property is deprecated and will be removed in Pillow 12.0",
```
radarhere, line 3709:
```suggestion
            "id property is deprecated and will be removed in Pillow 12 (2025-10-15)",
```
For consistency with _deprecate.py.

src/_imaging.c
```diff
@@ -3707,16 +3704,39 @@ _getattr_bands(ImagingObject *self, void *closure) {
...
+    Py_INCREF(self);
+    PyCapsule_SetContext(capsule, self);
+    return capsule;
 }
 
 static PyObject *
 _getattr_unsafe_ptrs(ImagingObject *self, void *closure) {
+    if (PyErr_WarnEx(
+            PyExc_DeprecationWarning,
+            "unsafe_ptrs property is deprecated and will be removed in Pillow 12.0",
```
radarhere, line 3735:
```suggestion
            "unsafe_ptrs property is deprecated and will be removed in Pillow 12 (2025-10-15)",
```

src/libImaging/Imaging.h
```diff
@@ -61,7 +61,7 @@ typedef struct ImagingOutlineInstance *ImagingOutline;
 typedef struct ImagingPaletteInstance *ImagingPalette;
 
 /* handle magics (used with PyCObject). */
```
radarhere, line 63:
```suggestion
/* handle magics (used with PyCapsule). */
```

Tests/test_image_getim.py
```diff
@@ -1,11 +1,19 @@
...
+    type_repr = repr(type(im.getim()))
     assert "PyCapsule" in type_repr
-    assert isinstance(im.im.id, int)
+
+    with pytest.warns(DeprecationWarning):
+        assert isinstance(im.im.id, int)
+
+    with pytest.warns(DeprecationWarning):
+        ptrs = dict(im.im.unsafe_ptrs)
+    assert all(k in ptrs for k in ["image8", "image32", "image"])
```
radarhere, line 19:
```suggestion
    assert list(ptrs.keys()) == ["image8", "image32", "image"]
```
Easier to read?

src/_imagingmorph.c
```diff
@@ -129,46 +124,39 @@ apply(PyObject *self, PyObject *args) {
 static PyObject *
 match(PyObject *self, PyObject *args) {
     const char *lut;
-    PyObject *py_lut;
-    Py_ssize_t lut_len, i0;
+    PyObject *i0;
+    Py_ssize_t lut_len;
```
radarhere, line 128:
```suggestion
    Py_ssize_t lut_len;
    PyObject *i0;
```
It seems better to have `lut` and `lut_len` next to each other?

src/_imagingmorph.c
```diff
@@ -30,35 +29,31 @@
 static PyObject *
 apply(PyObject *self, PyObject *args) {
     const char *lut;
-    PyObject *py_lut;
-    Py_ssize_t lut_len, i0, i1;
+    PyObject *i0, *i1;
+    Py_ssize_t lut_len;
```
radarhere, line 33:
```suggestion
    Py_ssize_t lut_len;
    PyObject *i0, *i1;
```

radarhere:
For the sake of interest, `unsafe_ptrs` was originally added in #476 for PyAccess (which was removed in #8182). `id` has been there since PIL.

> I'm not aware of TK API, but it seems it converts any arguments to strings, at least `PyImagingPhotoPut` receives arguments as `const char **argv`.

I looked, and this would be mirroring https://www.tcl.tk/man/tcl8.6/TclLib/CrtCommand.htm

src/_imaging.c
```diff
@@ -3707,16 +3704,39 @@ _getattr_bands(ImagingObject *self, void *closure) {
...
+    Py_INCREF(self);
+    PyCapsule_SetContext(capsule, self);
+    return capsule;
 }
 
 static PyObject *
 _getattr_unsafe_ptrs(ImagingObject *self, void *closure) {
+    if (PyErr_WarnEx(
+            PyExc_DeprecationWarning,
+            "unsafe_ptrs property is deprecated and will be removed in Pillow 12 (2025-10-15)",
```
radarhere, line 3735:
```suggestion
            "unsafe_ptrs property is deprecated and will be removed in Pillow 12 "
            "(2025-10-15)",
```
Lint fix.

docs/deprecations.rst
```diff
@@ -165,6 +165,16 @@ Specific WebP Feature Checks
 ``True`` if the WebP module is installed, until they are removed in Pillow
 12.0.0 (2025-10-15).
 
+Get Internal Pointers to Objects
```
hugovk, line 168:
```suggestion
Get internal pointers to objects
```

docs/releasenotes/11.0.0.rst
```diff
@@ -73,6 +73,16 @@ vulnerability introduced in FreeType 2.6 (:cve:`2020-15999`).
 
 .. _2.10.4: https://sourceforge.net/projects/freetype/files/freetype2/2.10.4/
 
+Get Internal Pointers to Objects
```
hugovk, line 76:
```suggestion
Get internal pointers to objects
```

PR #8349, opened 2024-09-05
Canonical type comparison

AryazE:
Since in numpy `dtype('float64') == None` is True, it is better to check with `.type` and `is` instead of `==`.

 * Using the canonical type comparison `is` instead of `==`.


radarhere (approved):
While
```pycon
>>> import numpy
>>> numpy.dtype('float64') == None
True
```
is the case, yes, I will note that
```
>>> numpy.dtype('float64').type == None
False
```
so it seems like `.type` is sufficient - but I'm not opposed to `is`

PR #8354, opened 2024-09-06
Added type hints

radarhere:


tox.ini
```diff
@@ -36,4 +36,4 @@ deps =
 extras =
     typing
 commands =
-    mypy docs src winbuild Tests {posargs}
+    mypy conftest.py selftest.py setup.py docs src winbuild Tests {posargs}
```
hugovk, line 39: To catch any new top-level scripts?
```suggestion
    mypy *.py docs src winbuild Tests {posargs}
```

src/PIL/TiffImagePlugin.py
```diff
@@ -2101,8 +2108,16 @@ def skipIFDs(self) -> None:
...
+    if sys.version_info >= (3, 12):
+        from collections.abc import Buffer
+
+        def write(self, data: Buffer, /) -> int:
+            return self.f.write(data)
+
+    else:
+
+        def write(self, data: Any, /) -> int:
+            return self.f.write(data)
```
hugovk, line 2120: Perhaps something like this?
```suggestion
    if sys.version_info >= (3, 12):
        from collections.abc import Buffer
    else:
        Buffer = Any

    def write(self, data: Buffer, /) -> int:
        return self.f.write(data)
```

hugovk (approved):

tox.ini
radarhere: Sadly and strangely, [that](https://github.com/radarhere/Pillow/commit/e69bc496d77aeb9ad486501305f5b8b3075b9087) [doesn't work.](https://github.com/radarhere/Pillow/actions/runs/10751715254/job/29819259203#step:8:17)

src/PIL/TiffImagePlugin.py
radarhere: Sure. I've actually moved `Buffer` into `src/PIL/_typing.py` so that this change can occur for GifImagePlugin cleanly as well.

PR #8355, opened 2024-09-06
Removed unused TiffImagePlugin IFD_LEGACY_API

radarhere:
This value is not used anywhere.

https://github.com/python-pillow/Pillow/blob/6377321625b93f5241305a7b301026cbbcd21e96/src/PIL/TiffImagePlugin.py#L72

### DIFF
--- a/docs/releasenotes/11.0.0.rst
+++ b/docs/releasenotes/11.0.0.rst
@@ -40,6 +40,11 @@ removed. Pillow's C API will now be used on PyPy instead.
 ``Image.USE_CFFI_ACCESS``, for switching from the C API to PyAccess, was
 similarly removed.
 
+TiffImagePlugin IFD_LEGACY_API
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+An unused setting, ``TiffImagePlugin.IFD_LEGACY_API``, has been removed.
+
 Deprecations
 ============
 
--- a/src/PIL/TiffImagePlugin.py
+++ b/src/PIL/TiffImagePlugin.py
@@ -69,7 +69,6 @@ logger = logging.getLogger(__name__)
 # Set these to true to force use of libtiff for reading or writing.
 READ_LIBTIFF = False
 WRITE_LIBTIFF = False
-IFD_LEGACY_API = True
 STRIP_SIZE = 65536
 
 II = b"II"  # little-endian (Intel style)
@@ -1125,7 +1124,7 @@ class ImageFileDirectory_v1(ImageFileDirectory_v2):
         return val
 
 
-# undone -- switch this pointer when IFD_LEGACY_API == False
+# undone -- switch this pointer
 ImageFileDirectory = ImageFileDirectory_v1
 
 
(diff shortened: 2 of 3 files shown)


hugovk:
Searching the [top 8k PyPI projects](https://dev.to/hugovk/how-to-search-5000-python-projects-31gk) only shows use in Pillow and type checkers:

```python
❯ python3 ~/github/misc/cpython/search_pypi_top.py -q . IFD_LEGACY_API
./pytype-2024.4.11.tar.gz: pytype-2024.4.11/pytype/typeshed/stubs/Pillow/PIL/TiffImagePlugin.pyi: IFD_LEGACY_API: bool
./pillow-10.3.0.tar.gz: pillow-10.3.0/src/PIL/TiffImagePlugin.py: IFD_LEGACY_API = True
./pillow-10.3.0.tar.gz: pillow-10.3.0/src/PIL/TiffImagePlugin.py: # undone -- switch this pointer when IFD_LEGACY_API == False
./pyre-check-0.9.22.tar.gz: pyre_check-0.9.22/typeshed/stubs/Pillow/PIL/TiffImagePlugin.pyi: IFD_LEGACY_API: bool
./types-Pillow-10.2.0.20240520.tar.gz: types-Pillow-10.2.0.20240520/PIL-stubs/TiffImagePlugin.pyi: IFD_LEGACY_API: bool

Time: 0:00:29.920569
Found 5 matching lines in 4 projects
```

Likewise with grep.app:

https://grep.app/search?q=IFD_LEGACY_API&case=true

And GitHub, where it's also in some AST symbol tables, but they look like dumps of specific versions:

https://github.com/search?q=IFD_LEGACY_API&ref=opensearch&type=code&p=1

Technically this is a breaking change, but I think we can make an exception to the deprecation period, and go straight to removal in a major version. Let's also document it in the release notes and removals.

docs/releasenotes/11.0.0.rst
```diff
@@ -69,10 +69,10 @@ Specific WebP Feature Checks
 API Changes
 ===========
 
-TODO
-^^^^
+TiffImagePlugin IFD_LEGACY_API
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-TODO
+An unused setting, ``TiffImagePlugin.IFD_LEGACY_API``, has been removed.
```
hugovk, line 75: Let's move this under "Backwards Incompatible Changes"

src/PIL/TiffImagePlugin.py
```diff
@@ -1125,7 +1124,6 @@ def __getitem__(self, tag: int) -> Any:
         return val
 
 
-# undone -- switch this pointer when IFD_LEGACY_API == False
 ImageFileDirectory = ImageFileDirectory_v1
```
hugovk, line 1127: Are there still plans to switch this in the future?

wiredfool: There were something like 3 stages to the tiff IFD modernization, and I got through stage 1, years back now. Something like it is going to be required if we're going to do robust tiff tags, but it's not on my list anymore.

radarhere: I've restored the commit that it is 'undone' to 'switch this pointer'

docs/releasenotes/11.0.0.rst
radarhere: Done

hugovk (approved):

PR #8362, opened 2024-09-09
Add type hints

### LINKED ISSUE #8029: `py.typed` present, but not all methods have types
We're using Pillow from our strictly-typed Python code.  We recently updated to Fedora 40, which brought us `python3-pillow-10.3.0-1.fc40.x86_64`.

This version of Pillow ships a `py.typed` file, which was not present in the version in Fedora 39 (`python3-pillow-10.2.0-1.fc39.x86_64`).

Previously, our mypy setup was ignoring our import of `PIL.Image` from a typing perspective, but with the `py.typed` file, it tries to check types.  This is problem, because we call `PIL.Image.Image.open()` which (among many other function) is untyped: https://github.com/python-pillow/Pillow/blob/e542c9f89883473fd0b7652b5157e98e4e15cee3/src/PIL/Image.py#L837

The following example produces no errors on Fedora 39, but fails on Fedora 40:

```python
from PIL import Image

with open('file.png', 'rb') as file:
    Image.open(file).load()
```

```
$ mypy --strict bzzt.py
bzzt.py:4: error: Call to untyped function "load" in typed context  [no-untyped-call]
Found 1 error in 1 file (checked 1 source file)
```

Output from `python3 -m PIL --report`:

```
Pillow 10.3.0
Python 3.12.2 (main, Feb 21 2024, 00:00:00) [GCC 14.0.1 20240217 (Red Hat 14.0.1-0)]
Python Pillow modules loaded from /usr/lib64/python3.12/site-packages/PIL
Binary Pillow modules loaded from /usr/lib64/python3.12/site-packages/PIL
--- WEBPMUX support ok
--- WEBP Animation support ok
--- ZLIB (PNG/ZIP) support ok, loaded 1.3.0.zlib-ng
--- RAQM (Bidirectional Text) support ok, loaded 0.8.0
--- LIBIMAGEQUANT (Quantization method) support ok, loaded 4.2.2
```

Out of curiosity, I got the current `main` branch of Pillow and added

```toml
[tool.mypy]
disallow_untyped_defs = true
```

to `pyproject.toml`.  There's quite a lot of untyped defs:

```
Found 853 errors in 58 files (checked 102 source files)
```

I'd like to propose a PR, but I'm afraid I don't have the time to work on such a large change.  Even the `Image` class has quite a lot of untyped methods (`load()`, `verify()`, `draft()`, `expand()`, `filter()`, and many others). 

In the meantime, I think the most appropriate course of action may be to temporarily remove the `py.typed` file.

radarhere:
Resolves #8029

In order to add type hints to the following `Image` methods
```python
def __enter__(self):
def _close_fp(self):
def __exit__(self, *args):
```
I've made a few changes.

`Image` now has
```python
def __enter__(self) -> Image:
def __exit__(self, *args: object) -> None:
```
I added an `__enter__()` and `__exit__()` to `ImageFile`, so that `with Image.open(...):` will type hint an `ImageFile` instance. I also moved `_close_fp()` to ImageFile, as it is only used when exiting an `ImageFile` instance.
```python
def __enter__(self) -> ImageFile:
def _close_fp(self) -> None:
def __exit__(self, *args: object) -> None:
```

I also moved `get_child_images()` from `Image` to `ImageFile`. It requires the use of `self.fp`, so this made sense to me, but feel to tell me it should stay in `Image`.

Believe it or not, all the other changes in this PR flowed out from there.


src/PIL/ImageFile.py
```diff
@@ -163,6 +168,93 @@ def __init__(
     def _open(self) -> None:
         pass
 
+    # Context manager support
+    def __enter__(self) -> ImageFile:
+        return self
+
+    def _close_fp(self) -> None:
+        if getattr(self, "_fp", False):
```
Yay295, line 176: Wouldn't `hasattr(self, "_fp")` work here?

hugovk:
This is such a big PR I've not found time to review it yet, so let's either ask for reviewers from the community (for example, via https://github.com/python-pillow/Pillow/issues/8029), or just go ahead and merge it if you'd like it in the next release.

radarhere:
I've commented in #8029 asking for reviews, but I'll also try and break this down into smaller PRs, like #8617. The PRs won't help external type hinting by themselves, but they should still individually make sense.

PR #8364, opened 2024-09-10
Deprecate isImageType

radarhere:
Removes `isImageType()`, which relies on `hasattr(t, "im")`,
https://github.com/python-pillow/Pillow/blob/be01d536c652f2224351ad0803c92067394db1bb/src/PIL/Image.py#L125-L136
and instead uses `isinstance(im, Image)`.

I've also made a similar change in ImageMath, replacing `hasattr(v, "im")` with `isinstance(v, Image.Image)`.

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -133,6 +133,7 @@ def isImageType(t: Any) -> TypeGuard[Image]:
     :param t: object to check if it's an image
     :returns: True if the object is an image
     """
+    deprecate("Image.isImageType(im)", 12, "isinstance(im, Image.Image)")
     return hasattr(t, "im")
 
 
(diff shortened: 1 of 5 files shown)


src/PIL/Image.py
```diff
@@ -1862,7 +1847,7 @@ def paste(
         elif isinstance(im, tuple):
             source = im
         else:
-            source = cast(float, im)
+            source = im
```
Yay295, line 1850: This `elif` could be removed now since it does the same thing as the `else` below it.

radarhere: You're right, I've updated the commit.

src/PIL/Image.py
```diff
@@ -122,20 +122,6 @@ class DecompressionBombError(Exception):
     raise
 
 
-def isImageType(t: Any) -> TypeGuard[Image]:
-    """
-    Checks if an object is an image object.
-
-    .. warning::
-
-       This function is for internal use only.
```
hugovk, line 131: This [11-year-old docstring](https://github.com/python-pillow/Pillow/commit/15eac684c4ef1ce38a2c95d259c60b3d46c803d3#diff-68bb791db6b090ce3dc87eb429f6b033085c0a4acd8e36894880e35db9e02cdbR99) may well have been warning this is for internal use only but the function is [24 years old, since PIL 1.1.1,](https://github.com/hugovk/PIL/blame/645e19c47556b9086c0d0fdccc1db1b9e6f79f7f/PIL/Image.py#L98) and is [used in the wild](https://grep.app/search?current=2&q=isImageType&case=true&filter[lang][0]=Python).

To be on the safe side, shall we put this through a deprecation cycle?

radarhere: Sure. I've pushed updates.

hugovk:
Thank you.

PR #8372, opened 2024-09-12
Install NumPy from PyPI for free threaded job

radarhere:
https://pypi.org/project/numpy/#files now offers 313t wheels, and so we can start installing NumPy from PyPI for the free threaded job, rather than using Scientific Python Nightly Wheels.

### DIFF
--- a/.ci/install.sh
+++ b/.ci/install.sh
@@ -38,12 +38,7 @@ python3 -m pip install -U pytest-timeout
 python3 -m pip install pyroma
 
 if [[ $(uname) != CYGWIN* ]]; then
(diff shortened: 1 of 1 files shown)


hugovk:
@lysnikolaou Which do you think we should use for free-threaded?

lysnikolaou:
Agreed that using 2.1.0 from PyPI is better. If there's any changes in NumPy that are necessary for pillow, we can always switch back to the nightlies. Do you want me to handle that?

hugovk:
Thanks, that would be helpful. Let's merge this.

Thanks both!

PR #8373, opened 2024-09-12
Install NumPy from PyPI when testing wheels

lysnikolaou:
Related to #8372.


.github/workflows/wheels-test.sh
```diff
@@ -15,11 +15,7 @@ fi
 
 if [ "${AUDITWHEEL_POLICY::9}" != "musllinux" ]; then
   # TODO Update condition when NumPy supports free-threading
```
Yay295, line 17: remove this comment too?

lysnikolaou: Done, thanks!

PR #8382, opened 2024-09-14
Lint: Run PT016

radarhere:
Sequel to #8309

Enables https://docs.astral.sh/ruff/rules/pytest-fail-without-message


hugovk:
Nice replacement of the try/except with pytest.raises :)

PR #8386, opened 2024-09-16
Use Capsule for WEBP saving

homm:
Adoption of #8341 ideas

### Improvements

* Doesn't create extra image copy (`.tobytes("raw", rawmode)`)
* Uses the same more appropriate API `.has_transparency_data` for save and save_all.
* Supports saving `RGBX` mode without conversion
* Adds `ImagingSectionEnter/Leave` for `WebPAnimEncoder`

### DIFF
--- a/Tests/test_file_webp.py
+++ b/Tests/test_file_webp.py
@@ -72,7 +72,7 @@ class TestFileWebp:
     def _roundtrip(
         self, tmp_path: Path, mode: str, epsilon: float, args: dict[str, Any] = {}
     ) -> None:
-        temp_file = str(tmp_path / "temp.webp")
+        temp_file = tmp_path / "temp.webp"
 
         hopper(mode).save(temp_file, **args)
         with Image.open(temp_file) as image:
@@ -116,7 +116,7 @@ class TestFileWebp:
         assert buffer_no_args.getbuffer() != buffer_method.getbuffer()
 
     def test_save_all(self, tmp_path: Path) -> None:
-        temp_file = str(tmp_path / "temp.webp")
+        temp_file = tmp_path / "temp.webp"
         im = Image.new("RGB", (1, 1))
         im2 = Image.new("RGB", (1, 1), "#f00")
         im.save(temp_file, save_all=True, append_images=[im2])
@@ -151,18 +151,16 @@ class TestFileWebp:
 
     @pytest.mark.skipif(sys.maxsize <= 2**32, reason="Requires 64-bit system")
     def test_write_encoding_error_message(self, tmp_path: Path) -> None:
-        temp_file = str(tmp_path / "temp.webp")
         im = Image.new("RGB", (15000, 15000))
         with pytest.raises(ValueError) as e:
-            im.save(temp_file, method=0)
+            im.save(tmp_path / "temp.webp", method=0)
         assert str(e.value) == "encoding error 6"
 
     @pytest.mark.skipif(sys.maxsize <= 2**32, reason="Requires 64-bit system")
     def test_write_encoding_error_bad_dimension(self, tmp_path: Path) -> None:
-        temp_file = str(tmp_path / "temp.webp")
         im = Image.new("L", (16384, 16384))
         with pytest.raises(ValueError) as e:
-            im.save(temp_file)
+            im.save(tmp_path / "temp.webp")
         assert (
             str(e.value)
             == "encoding error 5: Image size exceeds WebP limit of 16383 pixels"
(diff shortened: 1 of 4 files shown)


Tests/test_file_webp.py
```diff
@@ -204,27 +201,27 @@ def test_file_pointer_could_be_reused(self) -> None:
...
     ) -> None:
-        temp_file = str(tmp_path / "temp.webp")
+        temp_file = tmp_path / "temp.webp"
         im = hopper()
         with pytest.raises(OSError):
             im.save(temp_file, save_all=True, append_images=[im], background=background)
 
     def test_background_from_gif(self, tmp_path: Path) -> None:
+        out_webp = tmp_path / "temp.webp"
+
```
radarhere, line 211:
```suggestion
        out_webp = tmp_path / "temp.webp"

```
Lint fix.

src/PIL/WebPImagePlugin.py
```diff
@@ -243,31 +239,16 @@ def _save_all(im: Image.Image, fp: IO[bytes], filename: str | bytes) -> None:
...
-                        or (ims.mode == "P" and "A" in ims.im.getpalettemode())
-                    )
-                    rawmode = "RGBA" if alpha else "RGB"
-                    frame = ims.convert(rawmode)
-
-                if rawmode == "RGB":
-                    # For faster conversion, use RGBX
-                    rawmode = "RGBX"
+                if frame.mode not in ("RGBX", "RGBA", "RGB"):
+                    frame = frame.convert("RGBA" if im.has_transparency_data else "RGB")
```
radarhere, line 246:
```suggestion
                    frame = frame.convert("RGBA" if ims.has_transparency_data else "RGB")
```

homm: Thanks!

Yay295:
You're essentially using `_VALID_WEBP_MODES` in two locations, but you removed that constant. I assume you inlined them when `_VALID_WEBP_MODES` and `_VALID_WEBP_LEGACY_MODES` were each only used in one place, but it seems like `_VALID_WEBP_MODES` could be kept. Also it could just be a Set instead of a Dict with `True` values.

homm:
@Yay295 I've removed this constants primary since it's mistyping. Now I'm added separate `_convert_frame` routine for this.

radarhere:
I expected these changes to improve speed, but I'm not finding a definitive difference when I test. Does that mirror your experience?

Are the improvements here more about theory and tidying up?

Yay295:
I don't know how much of an effect it has, but there is some overhead when using Capsules because they use strings as identifiers. So there's some added string comparisons behind the scenes.

homm:
@radarhere
> Are the improvements here more about theory and tidying up?

All improvements are listed in the PR description. While elimination of `.tobytes()` doesn't affect execution time, it helps to avid extra image copy, which reduces memory footprint. See "Maximum resident set size" below. Also, `ImagingSectionEnter/Leave` in `WebPAnimEncoder` unlocks GIL in the current thread, which improves IO in other threads.

```python
from io import BytesIO
from PIL import Image
for i in range(4):
    with BytesIO() as f:
        Image.new('RGB', (8192, 8192)).save(f, 'webp', method=0, quality=1)
```

```
# AMD Cpu

## Main branch
$ /usr/bin/time -v python test.py
  Command being timed: "python test.py"
  User time (seconds): 19.39
  System time (seconds): 1.16
  Percent of CPU this job got: 99%
  Elapsed (wall clock) time (h:mm:ss or m:ss): 0:20.63
  Maximum resident set size (kbytes): 1296972
  Major (requiring I/O) page faults: 18
  Minor (reclaiming a frame) page faults: 1484790
  Voluntary context switches: 19
  Involuntary context switches: 719

## webp-capsule branch
$ /usr/bin/time -v python test.py 
  Command being timed: "python test.py"
  User time (seconds): 19.46
  System time (seconds): 0.78
  Percent of CPU this job got: 99%
  Elapsed (wall clock) time (h:mm:ss or m:ss): 0:20.27
  Maximum resident set size (kbytes): 1100660
  Major (requiring I/O) page faults: 0
  Minor (reclaiming a frame) page faults: 1090899
  Voluntary context switches: 1
  Involuntary context switches: 1180


# Local Docker (M1 Pro)

## Main branch
$ /usr/bin/time -v python test.py
    Command being timed: "python test.py"
    User time (seconds): 4.51
    System time (seconds): 0.35
    Percent of CPU this job got: 99%
    Elapsed (wall clock) time (h:mm:ss or m:ss): 0:04.89
    Maximum resident set size (kbytes): 851160
    Major (requiring I/O) page faults: 9
    Minor (reclaiming a frame) page faults: 674262
    Voluntary context switches: 495
    Involuntary context switches: 88

## webp-capsule branch
$ /usr/bin/time -v python test.py
    Command being timed: "python test.py"
    User time (seconds): 4.18
    System time (seconds): 0.24
    Percent of CPU this job got: 99%
    Elapsed (wall clock) time (h:mm:ss or m:ss): 0:04.44
    Maximum resident set size (kbytes): 648432
    Major (requiring I/O) page faults: 9
    Minor (reclaiming a frame) page faults: 286806
    Voluntary context switches: 390
    Involuntary context switches: 10
```

homm:
@Yay295 
> some overhead when using Capsules because they use strings as identifiers

It's literally nothing comparing to the copying megabytes of image data to the byte string. But as you can see below, event the copying doesn't affect encoding time much.

radarhere (approved):

PR #8387, opened 2024-09-16
TIFF: report correct image size after opening

homm:


src/PIL/TiffImagePlugin.py
```diff
@@ -1275,6 +1275,13 @@ def load(self) -> Image.core.PixelAccess | None:
             return self._load_libtiff()
         return super().load()
 
+    @property
+    def size(self) -> tuple[int, int]:
+        if hasattr(self, "tag_v2"):
```
homm, line 1280: There is a cases after copying or pickling, when the class is actually `TiffImageFile`, but it is loaded and doesn't have `tag_v2` attribute

src/PIL/TiffImagePlugin.py
```diff
@@ -1275,6 +1275,13 @@ def load(self) -> Image.core.PixelAccess | None:
             return self._load_libtiff()
         return super().load()
 
+    @property
+    def size(self) -> tuple[int, int]:
```
homm, line 1279: This implementation is differ from https://github.com/uploadcare/pillow-simd/commit/1d77bfe9d939e208e3bb2b897a4209c0881eb9f1 since for me it looks less magical, but if you can't agree, I can copy the previous as is

radarhere:
This PR was created in an effort to ensure that [`load()` does not change the size of an image](https://github.com/python-pillow/Pillow/pull/8341#discussion_r1760798235).

It naturally sounds like a good idea for images to have the same size after opening as after loading. However, Pillow also aims to [open an image without loading it, for the sake of performance](https://github.com/python-pillow/Pillow/pull/7311#issuecomment-1723004045), and the rotation data might not appear early on in the file.

For the PNG format, https://ftp-osl.osuosl.org/pub/libpng/documents/pngext-1.5.0.html#C.eXIf states
> The eXIf chunk may appear anywhere between the IHDR and IEND chunks except between IDAT chunks.

So I don't think it is possible to simultaneously achieve those two goals.

homm:
There is no conflict between the presence of eXIf chunks in PNG files and the goal of "having the same size after opening as after loading," because when loading a PNG (by calling the `.load()` method or any dependent method), the Exif data is populated, but the image is not automatically rotated. This differs from the behavior of TIFF, where calling `.load()` actually rotates the image and can change its size. This is exactly the issue I aim to address in this PR.

radarhere:
Ah, I see now, thanks.

I think introducing the idea that `im.size` might not match `im._size` complicates things. How about #8390 instead?

hugovk:
Closed in favour of https://github.com/python-pillow/Pillow/pull/8390.

PR #8390, opened 2024-09-18
Use transposed size after opening for TIFF images

radarhere:
Alternative to #8387

At the moment, `load()` might change the size of a TIFF image, since TIFF images are [automatically transposed when doing so](https://github.com/python-pillow/Pillow/blob/08d9c89d8a7db84e8145814bca36d7dd5b7465ce/src/PIL/TiffImagePlugin.py#L1295).

`load()` is a method that many users are probably not even aware of, and so they might find
```python
from PIL import Image
im = Image.open("/Users/andrewmurray/pillow/Pillow/Tests/images/g4_orientation_5.tif")
print(im.size)  # (88, 590)
print(im.copy().size)  # (590, 88)
```
unexpected.

After this PR, the first `im.size` will also be (590, 88). This has means that changes from #6186 and #6190 can be removed.

#8387 solves this by making `im.size` sometimes different to `im._size`. I would like to avoid that, and have instead found a solution by adding `TiffImageFile.load_prepare()`.

While we are here streamlining things so that an image does not change its size after loading, I've also updated ImageFile to remove an implication that loading might change an image's mode.

### DIFF
--- a/src/PIL/ImageFile.py
+++ b/src/PIL/ImageFile.py
@@ -322,7 +322,7 @@ class ImageFile(Image.Image):
 
     def load_prepare(self) -> None:
         # create image memory if necessary
-        if self._im is None or self.im.mode != self.mode or self.im.size != self.size:
+        if self._im is None or self.im.mode != self.mode:
             self.im = Image.core.new(self.mode, self.size)
         # create palette (optional)
         if self.mode == "P":
--- a/src/PIL/ImageFile.py
+++ b/src/PIL/ImageFile.py
@@ -322,7 +322,7 @@ class ImageFile(Image.Image):
 
     def load_prepare(self) -> None:
         # create image memory if necessary
-        if self._im is None or self.im.mode != self.mode:
+        if self._im is None:
             self.im = Image.core.new(self.mode, self.size)
         # create palette (optional)
         if self.mode == "P":
--- a/src/PIL/TiffImagePlugin.py
+++ b/src/PIL/TiffImagePlugin.py
@@ -1275,6 +1275,11 @@ class TiffImageFile(ImageFile.ImageFile):
             return self._load_libtiff()
         return super().load()
 
+    def load_prepare(self) -> None:
+        if self._im is None and self._will_be_transposed:
+            self.im = Image.core.new(self.mode, self.size[::-1])
+        ImageFile.ImageFile.load_prepare(self)
+
     def load_end(self) -> None:
         # allow closing if we're on the first frame, there's no next
         # This is the ImageFile.load path only, libtiff specific below.
@@ -1416,7 +1421,16 @@ class TiffImageFile(ImageFile.ImageFile):
         if not isinstance(xsize, int) or not isinstance(ysize, int):
             msg = "Invalid dimensions"
             raise ValueError(msg)
-        self._size = xsize, ysize
+        self._will_be_transposed = self.tag_v2.get(ExifTags.Base.Orientation) in (
+            5,
+            6,
+            7,
+            8,
+        )
+        if self._will_be_transposed:
+            self._size = ysize, xsize
+        else:
+            self._size = xsize, ysize
 
         logger.debug("- size: %s", self.size)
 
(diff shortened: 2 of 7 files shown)


src/PIL/ImageFile.py
```diff
@@ -322,7 +322,7 @@ def load(self) -> Image.core.PixelAccess | None:
 
     def load_prepare(self) -> None:
         # create image memory if necessary
-        if self._im is None or self.im.mode != self.mode or self.im.size != self.size:
```
homm, line 325: Not sure now it affects multipage images with different page sizes

src/PIL/TiffImagePlugin.py
```diff
@@ -1416,7 +1421,16 @@ def _setup(self) -> None:
         if not isinstance(xsize, int) or not isinstance(ysize, int):
             msg = "Invalid dimensions"
             raise ValueError(msg)
-        self._size = xsize, ysize
+        self._will_be_transposed = self.tag_v2.get(ExifTags.Base.Orientation) in (
+            5,
+            6,
+            7,
+            8,
+        )
```
homm, line 1429: Just a stylistic correction
```suggestion
        orientation = self.tag_v2.get(ExifTags.Base.Orientation)
        self._will_be_transposed = orientation in (5, 6, 7, 8)
```

src/PIL/TiffImagePlugin.py
```diff
@@ -1416,7 +1421,16 @@ def _setup(self) -> None:
...
+        if self._will_be_transposed:
+            self._size = ysize, xsize
```
homm, line 1431: Please be aware of following `self.size` using in the same function, which should be fixed:

https://github.com/python-pillow/Pillow/pull/8387/files#diff-6ad43f85f1a075181d4d8cfcd97ae27bba1eccf5c3db5a3457160f98218eba6eR1569
https://github.com/python-pillow/Pillow/pull/8387/files#diff-6ad43f85f1a075181d4d8cfcd97ae27bba1eccf5c3db5a3457160f98218eba6eR1603-R1605

radarhere: I've pushed a commit for this.

src/PIL/ImageFile.py
radarhere: When seeking to a different TIFF frame, `self.im` is populated afresh.

https://github.com/python-pillow/Pillow/blob/08d9c89d8a7db84e8145814bca36d7dd5b7465ce/src/PIL/TiffImagePlugin.py#L1189-L1198

homm: Yeah, I saw this and tried to remove this initialization and relay on `load_prepare` instead, for some reasons it didn't work.

For me providing correct canvas (self.im) in the base class is more preferable than recreating it the plugin. There was [another approach](https://github.com/uploadcare/pillow-simd/commit/1d77bfe9d939e208e3bb2b897a4209c0881eb9f1#diff-6ad43f85f1a075181d4d8cfcd97ae27bba1eccf5c3db5a3457160f98218eba6eR1205) which doesn't introduce difference between `im.size` and `im._size`.

radarhere: > For me providing correct canvas (self.im) in the base class is more preferable than recreating it in the plugin.

If you think I'm creating a core image instance an extra time in `TiffImagePlugin.load_prepare()` at the moment, that's not the case - the creating of the core image in `ImageFile.load_prepare()` is skipped [because](https://github.com/python-pillow/Pillow/blob/e80cada0f691ef60d4448145dc9dd0319c13b68f/src/PIL/ImageFile.py#L323-L326) it's already performed here.

radarhere: > Yeah, I saw this and tried to remove this initialization and relay on load_prepare instead, for some reasons it didn't work.

I've pushed https://github.com/python-pillow/Pillow/pull/8390/commits/3ab41ec7dbf55ea535d1691a102fa17fe9bfa4b5, see what you think.

radarhere: Ah, I see now - my `load_prepare()` would create a new core image when the image was loaded after `seek()`.

I've pushed a commit following your other approach.

homm (approved):

PR #8393, opened 2024-09-18
Test image stats of all image modes

Yay295:
Test the image stats of all image modes. This should give an indicator if a code change changes the image data anywhere.


Yay295:
and somehow specific tests in specific environments are failing

Yay295:
The image stats are all based on `Image.histogram()`, so I've created #8394.

radarhere:
https://github.com/python-pillow/Pillow/pull/8394#issuecomment-2360116458
> This makes me think there's actually something going on in the conversion from "RGB" to "LAB" and "HSV" that is somehow not the same in every environment.

From what I can see, our 32-bit Docker job is giving different results from RGB to HSV conversion at
https://github.com/python-pillow/Pillow/blob/b557876ec3aa4d5d236d8044519f1c613031fbee/src/libImaging/Convert.c#L342

Perhaps this is to be expected when you are check the output with a high level of precision?

As for the LAB check, it is happening when using lcms2 2.6. This is not surprising given it is noticeably older than lcms2 2.12, otherwise the earliest version used in our Docker jobs.

Yay295:
I tried changing the RGB to HSV conversion to use `double` instead of `float`, and I tried implementing `fmod()` directly, but neither of those changes (nor both together) resolved the issue. It did cause the max value to be correct, but the hash was still wrong for just that environment.

radarhere:
Part of the goal of this PR is to check the output with a high level of precision, but it would seem this isn't possible in a 32-bit environment.

Closing, as the tests aren't passing.

PR #8400, opened 2024-09-20
Check image value before use

radarhere:
Similar to https://github.com/python-pillow/Pillow/pull/8398, check the value of `imIn` before using it.

https://github.com/python-pillow/Pillow/blob/b557876ec3aa4d5d236d8044519f1c613031fbee/src/libImaging/Geometry.c#L794-L801
https://github.com/python-pillow/Pillow/blob/b557876ec3aa4d5d236d8044519f1c613031fbee/src/libImaging/Geometry.c#L709-L712


hugovk (approved):

hugovk:
(I misclicked the "update branch" button, but never mind.)

PR #8302, opened 2024-08-12
Updated freetype to 2.13.3

radarhere:
FreeType 2.13.3 has been released - https://freetype.org/

PR #8403, opened 2024-09-22
winbuild: Build freetype.vcxproj to fix passing in custom parameters

nulano:
#8302 broke passing in custom properties to FreeType, fix it by using the VS project file directly instead of the solution file.

This should fix the skipped CBDT/SBIX/WOFF2 tests.


nulano:
An alternative would be to pass in the custom parameters directly using `/p:UserDefines="..."` format, but that causes the command to be really long, and I am not convinced it is easier to read.

PR #8406, opened 2024-09-23
Add check for positive xsize/ysize in ImagingFliDecode

### LINKED ISSUE #8405: Unused code in FliDecode.c and _imaging.c has SAST security issue
Hello!

According to the comments in _imaging.c thers is number of codecs, that must be replaced in PIL 1.2. But PIL development was discontinued 15 years ago.

image

One of this codecs has SAST security issue.

image

`state->xsize` potentially be equal to zero, which will lead to division by zero exception in `ImagingFliDecode`. If this code is not used, I suggest remove it.

Found by Linux Verification Center (linuxtesting.org) with SVACE.
Reporter: Dmitriy Karasovsky ([d.karasovsky@fobos-nt.ru](mailto:d.karasovsky@fobos-nt.ru)).

Yay295:
Fixes #8405

I put the check near the top, but is there any valid reason `xsize` or `ysize` would be 0, and the check would need to be later in the code?

### DIFF
--- a/src/libImaging/FliDecode.c
+++ b/src/libImaging/FliDecode.c
@@ -62,6 +62,10 @@ ImagingFliDecode(Imaging im, ImagingCodecState state, UINT8 *buf, Py_ssize_t byt
         state->errcode = IMAGING_CODEC_UNKNOWN;
         return -1;
     }
+    if (state->xsize <= 0 || state->ysize <= 0) {
+        state->errcode = IMAGING_CODEC_CONFIG;
+        return -1;
+    }
 
     chunks = I16(ptr + 6);
     ptr += 16;


radarhere:
> is there any valid reason `xsize` or `ysize` would be 0

No. Images that say one of their dimensions are zero will be stopped at https://github.com/python-pillow/Pillow/blob/731bcda904544d9d26bce268eca3a5cb4fcc1c46/src/PIL/ImageFile.py#L154-L156

Even if you consider just the C decoding process, we have
https://github.com/python-pillow/Pillow/blob/731bcda904544d9d26bce268eca3a5cb4fcc1c46/src/decode.c#L189-L192

radarhere:
My minor concern with this strategy is that we could make someone reading the code think that `xsize` might be zero, which as I've said, shouldn't happen.

I've created #8408 as an alternative to this.

radarhere:
#8408 has been merged instead.

PR #8417, opened 2024-09-25
Support writing LONG8 offsets in AppendingTiffWriter

radarhere:
https://github.com/python-pillow/Pillow/blob/3f24276bf2ab8c29b2c3aa96f5bca2b60b9f3340/Tests/test_file_tiff.py#L110-L116

if you try and run the test without removing this tag, an error is raised from `AppendingTiffWriter`.
```
    def fixOffsets(
        self, count: int, isShort: bool = False, isLong: bool = False
    ) -> None:
        if not isShort and not isLong:
            msg = "offset is neither short nor long"
>           raise RuntimeError(msg)
E           RuntimeError: offset is neither short nor long
```

Extending `AppendingTiffWriter` to handle LONG8 offsets is made complicated as it explicitly only handles shorts and longs, and has duplicated code to achieve that.

https://github.com/python-pillow/Pillow/blob/3f24276bf2ab8c29b2c3aa96f5bca2b60b9f3340/src/PIL/TiffImagePlugin.py#L2143-L2151
https://github.com/python-pillow/Pillow/blob/3f24276bf2ab8c29b2c3aa96f5bca2b60b9f3340/src/PIL/TiffImagePlugin.py#L2153-L2159
https://github.com/python-pillow/Pillow/blob/3f24276bf2ab8c29b2c3aa96f5bca2b60b9f3340/src/PIL/TiffImagePlugin.py#L2198-L2200

Rather than just adding in `rewriteLastLong8()`, `readLong8()`, and adding an `isLong8` parameter, I've reworked the internals of the class to pass around the field size in my first commit. That made adding support for LONG8 very straightforward in my second commit.

### DIFF
--- a/src/PIL/TiffImagePlugin.py
+++ b/src/PIL/TiffImagePlugin.py
@@ -2121,13 +2121,24 @@ class AppendingTiffWriter(io.BytesIO):
     def write(self, data: Buffer, /) -> int:
         return self.f.write(data)
 
-    def readShort(self) -> int:
-        (value,) = struct.unpack(self.shortFmt, self.f.read(2))
+    def _fmt(self, field_size: int) -> str:
+        try:
+            return {2: "H", 4: "L"}[field_size]
+        except KeyError:
+            msg = "offset is not supported"
+            raise RuntimeError(msg)
+
+    def _read(self, field_size: int) -> int:
+        (value,) = struct.unpack(
+            self.endian + self._fmt(field_size), self.f.read(field_size)
+        )
         return value
 
+    def readShort(self) -> int:
+        return self._read(2)
+
     def readLong(self) -> int:
-        (value,) = struct.unpack(self.longFmt, self.f.read(4))
-        return value
+        return self._read(4)
 
     @staticmethod
     def _verify_bytes_written(bytes_written: int | None, expected: int) -> None:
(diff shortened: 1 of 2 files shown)


src/PIL/TiffImagePlugin.py
```diff
@@ -2180,32 +2194,22 @@ def fixIFD(self) -> None:
...
-            msg = "offset is neither short nor long"
-            raise RuntimeError(msg)
-
+    def _fixOffsets(self, count: int, field_size: int) -> None:
         for i in range(count):
-            offset = self.readShort() if isShort else self.readLong()
+            offset = self._read(field_size)
             offset += self.offsetOfNewPage
-            if isShort and offset >= 65536:
+            if field_size == 2 and offset >= 65536:
```
Yay295, line 2212: There should probably now also be a check for `field_size == 4 and offset >= 2**32`.

radarhere: It is conceivable that some users might not wish for us to silently upgrade tags to the LONG8 type, since it would be changing the saved image to rely on the [BigTIFF](https://web.archive.org/web/20240706160214/https://www.awaresystems.be/imaging/tiff/bigtiff.html) specification.

radarhere: I've added this functionality to #8663, provided the image is being saved as a BigTIFF.

PR #8432, opened 2024-10-01
Handle negative intermediate value in rgb to hsv conversion

Yay295:
An incorrect hue was being calculated when "h/6 is negative" because Python's `%` and C's `fmod` use a slightly different strategy. So to match Python, we need to implement it ourselves. It can also be simplified some due to the second parameter being 1.

The starting hue calculation has also been simplified by inlining some values. ex.:
```c
h = bc - gc;
h = (((float)(maxc - b)) / cr) - (((float)(maxc - g)) / cr);
h = ((maxc - b) - (maxc - g)) / cr;
h = (maxc - b - maxc + g) / cr;
h = (g - b) / cr
```


Yay295:
For the `fmod` simplification, `fmod` looks like this (if it wasn't implemented in assembly):
```c
double fmod(double x, double y) {
    return x - y * trunc(x / y);
}
```
Since `y` is `1.0` in our case, this becomes just `x - trunc(x)`. And then replace `trunc` with `floor` to match Python.

Yay295:
I've been trying to tweak this code for long enough that I forgot the `+ 1.0` in the original code (`fmod((h / 6.0 + 1.0), 1.0)`) already handles the negative values. So this change doesn't fix anything; it's just optimizations.

radarhere:
From what I can see, there are four basic changes here.
1. You've rearranged the declarations. I don't think there's a need to change them being grouped at the start. It is a pattern used elsewhere, e.g. https://github.com/python-pillow/Pillow/blob/f614580a2de93235d9fe5c802f3d54f881a65c5e/src/libImaging/Convert.c#L1263-L1267
2. You've rearranged math operations to divide once rather than twice.
3. You've replaced `fmod` with other code. From the back and forth comments, I'm a bit confused about why you've done that?
4. You've removed `CLIP8`. Could you explain why?

hugovk:
re: 1., I think declarations at the start of a block was required by the C89 standard, but is no longer by modern compilers. And I find declaring on first use more readable, rather than have to refer back.

Yay295:
> You've replaced `fmod` with other code. From the back and forth comments, I'm a bit confused about why you've done that?

Originally I did it because the existing comment confused me, thinking negative values weren't being handled correctly. But the new code does replace `fmod` with a simple `floor`, so it should be slightly faster.

> You've removed `CLIP8`. Could you explain why?

Because of the `fmod`, `h` at that point will be between 0 and 1, so there's no need to clip it after multiplying by 255 because it will already be between 0 and 255.

`s` in the original code is `255.0 * (float)(maxc - minc) / (float)(maxc)`. `(float)(maxc - minc) / (float)(maxc)` will be between 0 and 1, so again the result will already be between 0 and 255. This calculation can also be done without converting anything to floating-point, so I did that too.

src/libImaging/Convert.c
```diff
@@ -310,40 +310,45 @@ rgb2bgr24(UINT8 *out, const UINT8 *in, int xsize) {
...
         } else {
-            h = 4.0 + gc - rc;
+            h = 4.0 + (r - g) / cr;
         }
-        // incorrect hue happens if h/6 is negative.
-        h = fmod((h / 6.0 + 1.0), 1.0);
 
-        uh = (UINT8)CLIP8((int)(h * 255.0));
-        us = (UINT8)CLIP8((int)(s * 255.0));
+        // the modulus operator in Python does not exactly match fmod in C
```
radarhere, line 342: Outside of the context of this PR, I don't think it's necessary to talk about fmod here. After this PR, there would be no fmod in the code anymore, and it wouldn't be in colorsys, so what does it matter how it behaves?

Yay295: Maybe "the modulus operator in Python does not exactly match the modulus operator or the fmod function in C" then.

radarhere:
> So this change doesn't fix anything; it's just optimizations.

To be clear, you're saying that this PR should not adjust the output data?

Yay295:
Correct

radarhere:
To see if the output stayed the same, I converted hopper to HSV, and saved the output to a file - https://github.com/radarhere/Pillow/commit/8811b00940a0a851a2f9f3adf67bc516d72d4182 / https://github.com/radarhere/Pillow/actions/runs/11159666541

When I introduced this change, [the test started failing](https://github.com/radarhere/Pillow/actions/runs/11159669890) - meaning that this does change the output.

hugovk:
See https://github.com/python/cpython/pull/124880 for a similar CPython change to `colorsys.rgb_to_hsv()`, which the Pillow function was originally based on.

src/libImaging/Convert.c
```diff
@@ -310,40 +310,46 @@ rgb2bgr24(UINT8 *out, const UINT8 *in, int xsize) {
...
     } else {
-        cr = (float)(maxc - minc);
-        s = cr / (float)maxc;
-        rc = ((float)(maxc - r)) / cr;
-        gc = ((float)(maxc - g)) / cr;
-        bc = ((float)(maxc - b)) / cr;
+        const UINT8 color_range = maxc - minc;
+        double h;
+
+        const double cr = (double)color_range;
```
homm, line 333: I believe double in overkill here.

```suggestion
        float h;

        const float cr = (float)color_range;
```

src/libImaging/Convert.c
```diff
@@ -310,40 +310,46 @@ rgb2bgr24(UINT8 *out, const UINT8 *in, int xsize) {
...
-        h = fmod((h / 6.0 + 1.0), 1.0);
 
-        uh = (UINT8)CLIP8((int)(h * 255.0));
-        us = (UINT8)CLIP8((int)(s * 255.0));
+        // the modulus operator in Python does not exactly match
+        // the modulus operator or the fmod function in C
+        // https://stackoverflow.com/a/3883019/3878168
+        // "h = (h/6.0) % 1.0" in Python can be computed as:
+        h = h / 6.0;
+        h = h - floor(h);
```
homm, line 347:
```suggestion
        h = h - floorf(h);
```

src/libImaging/Convert.c
Yay295: Is there any downside to using a `double` here?

homm: Performance, of course. It could be more or less noticeable depending on platform. Also long types often prevents auto vectorisation (which I believe still doesn't work here even with floats).

```python
from PIL import Image
im = Image.new('RGB', (1024, 1024))
%timeit -n 1000 im.convert('HSV')
```

### Master
2.04 ms ± 39.5 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

### Yay295:patch-2
2.18 ms ± 27.2 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

### Yay295:patch-2 with float
2.05 ms ± 62.6 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

radarhere:
Given this PR is [motivated](https://github.com/python-pillow/Pillow/pull/8432#issuecomment-2384677399) purely by performance, but [changes the output](https://github.com/python-pillow/Pillow/pull/8432#issuecomment-2390988839) as a side effect, I'm going to close this. I doubt this function is a bottleneck, and given that we receive issues even when [PNG bytes change in saved images](https://github.com/python-pillow/Pillow/issues/8749), I'm reluctant to alter output.

PR #8434, opened 2024-10-02
Updated xz to 5.6.3

radarhere:
xz 5.6.3 has been released - https://github.com/tukaani-project/xz/releases/tag/v5.6.3


winbuild/build_prepare.py
```diff
@@ -185,7 +185,7 @@ def cmd_msbuild(
             cmd_copy(r"src\liblzma\api\lzma\*.h", r"{inc_dir}\lzma"),
         ],
         "headers": [r"src\liblzma\api\lzma.h"],
-        "libs": [r"liblzma.lib"],
+        "libs": [r"lzma.lib"],
```
hugovk, line 188: Why these changes?

radarhere: https://github.com/tukaani-project/xz/releases/tag/v5.6.3
> * MSVC: Rename import library from liblzma.lib to lzma.lib while keeping liblzma.dll name as is. This helps with "pkgconf --msvc-syntax --libs liblzma" because it mungles "-llzma" in liblzma.pc to "lzma.lib".

Issue #7910, opened 2024-03-27
Do we still need AppVeyor?


hugovk:
Are we still getting value from AppVeyor compared to the Windows tests on GitHub Actions?

AppVeyor tests:
* 3.8 on x86-64 (Windows Server 2017)
* 3.12 on x86 (Windows Server 2022)

Which of these are important? I expect 32-bit, whilst we're supporting it. Does OS version matter?

GitHub Actions tests:
* 3.8-3.13 and PyPy3.9-3.10 on x86-64 (Windows Server 2022)

GHA also has Windows Server 2019 available:

* https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

I thought GHA also had 32-bit available, but I can't find it right now.

aclark4life:
While I still support providing 32 bit wheels as described in #7443 until folks no longer notice if we stop, and while I support having one less vendor involved in this process (no offense to AppVeyor), it looks like the [official word](https://github.com/python-pillow/Pillow/issues/7443) is no 32-bit on GHA and so we'd need a reliable work around.

hugovk:
Another option is to only test 32-bit on AppVeyor, so it would be ~10 minutes.

But it would be nice having just one CI vendor, less types of config to maintain.

aclark4life:
> But it would be nice having just one CI vendor, less types of config to maintain.

That's what I'm interested in as well. I'm leaning towards:

- 3.8 on on 64-bit Windows Server 2017 is "too old to care about"
- 3.12 on 32-bit Windows 2022 should remain in some form but not on AppVeyor.

We probably should ask @cgohlke or someone else with lots of Windows experience to weigh in, but my feeling is we support our matrix running on Windows "latest" and not much else.

hugovk:
Thanks @nulano!

It's not been a big problem having AppVeyor in addition to GHA. It would indeed be nice to have just provider, but broad and reliable test coverage is more important. It's good to review things once in a while.

What do you think about using a single job on AppVeyor, combining 32-bit (either 32-bit CPU or 64-bit CPU with 32-bit Python) with Windows Server 2016?

As mentioned, the AppVeyor duration isn't too slow compared to GHA, so I'm also fine to keep 2 jobs, it doesn't add much complexity.

---

Correcting my OP:

> AppVeyor tests:
> 
> * 3.8 on x86-64 (Windows Server 2017)
[...]

PR #8437, opened 2024-10-04
Restore testing of Windows Server 2016 on AppVeyor

radarhere:
@nulano has suggested that we continue testing Windows Server 2016 - https://github.com/python-pillow/Pillow/issues/7910#issuecomment-2023624348

> > Does OS version matter?
> 
> We currently [claim to support Visual Studio 2017](https://github.com/python-pillow/Pillow/blob/main/winbuild/build.rst), which is only available on [AppVeyor using the Windows Server 2016 image](https://www.appveyor.com/docs/windows-images-software/); the [GHA windows-2019 image only has Visual Studio 2019](https://github.com/actions/runner-images/blob/main/images/windows/Windows2019-Readme.md#visual-studio-enterprise-2019). I'm also hesitant about using older windows images on GHA after the (at least initially) short deprecation window of the windows-2016 image (#4188).

However, since AppVeyor doesn't install anything newer than Python 3.8 on Windows Server 2016, this is not entirely straightforward.

In this PR, I've installed 64-bit Python 3.9 through Chocolately.


radarhere:
Closing in favour of https://github.com/python-pillow/Pillow/pull/8525

nulano:
To be clear, I would also approve these changes, but the button is no longer available because its closed.

To repeat my comment from #8525, this would have the benefit of giving us Windows Server 2016 testing, but if the goal is to simplify the CI configs then I suppose we can stick with just GHA.

nulano (approved):

Issue #6667, opened 2022-10-17
ImageFilter does not work with I;16


amdepott:
### What did you do?
Attempted to use Image.filter(ImageFilter.BLUR) on an image of mode I or I;16

### What did you expect to happen?
Image is blurred

### What actually happened?
Code throws ValueError

### What are your OS, Python and Pillow versions?

* OS: Windows 10 Home
* Python: 3.10.1
* Pillow: 9.2.0

```pycon
>>> from tkinter import filedialog
>>> from PIL import Image, ImageFilter
>>> image1 = Image.open(filedialog.askopenfilename())
>>> image1
   <PIL.BmpImagePlugin.BmpImageFile image mode=RGB size=630x514 at 0x28B7051E9B0>
>>> image2 = image1.convert('I')
>>> image2
   <PIL.Image.Image image mode=I size=630x514 at 0x28B70A8E560>
>>> image1.filter(ImageFilter.BLUR).show()
>>> image2.filter(ImageFilter.BLUR).show()
   Traceback (most recent call last):
     File "<pyshell#56>", line 1, in <module>
       image2.filter(ImageFilter.BLUR).show()
     File "C:\Users\x\AppData\Local\Programs\Python\Python310\lib\site-packages\PIL\Image.py", line 1247, in filter
       return self._new(filter.filter(self.im))
     File "C:\Users\x\AppData\Local\Programs\Python\Python310\lib\site-packages\PIL\ImageFilter.py", line 32, in filter
       return image.filter(*self.filterargs)
   ValueError: image has wrong mode
```

radarhere:
Just an explanation from the code.
https://github.com/python-pillow/Pillow/blob/1f5be894b91a00731ca328b3352eaf69eaee766c/src/libImaging/Filter.c#L326-L332
means that it doesn't support `I`, `I;*`, `F` or `BGR;*`.

radarhere:
PR #7108 has been merged, fixing this for I mode.

GrayMask:
Hi!
When I use `ImageFilter` to performs a look-up table (LUT) transform on an image of mode `I` or `I;16`, 
```
File ".../PIL/ImageFilter.py", line 560, in filter
    return image.color_lut_3d(
ValueError: image has wrong mode
```
still occurred.

I found that I mode has been supported for `BuiltinFilter` but still not supported for `Color3DLUT`.
related code:
[...]

PR #8438, opened 2024-10-04
Support ImageFilter.BuiltinFilter for I;16* images

radarhere:
Helps #6667

### DIFF
--- a/Tests/test_image_filter.py
+++ b/Tests/test_image_filter.py
@@ -35,12 +35,14 @@ from .helper import assert_image_equal, hopper
         ImageFilter.UnsharpMask(10),
     ),
 )
-@pytest.mark.parametrize("mode", ("L", "I", "RGB", "CMYK"))
+@pytest.mark.parametrize(
+    "mode", ("L", "I", "I;16", "I;16L", "I;16B", "I;16N", "RGB", "CMYK")
+)
 def test_sanity(
     filter_to_apply: ImageFilter.Filter | type[ImageFilter.Filter], mode: str
 ) -> None:
     im = hopper(mode)
-    if mode != "I" or (
+    if mode[0] != "I" or (
         callable(filter_to_apply)
         and issubclass(filter_to_apply, ImageFilter.BuiltinFilter)
     ):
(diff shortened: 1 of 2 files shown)


Tests/test_image_filter.py
```diff
@@ -35,16 +35,25 @@
...
+)
+def test_sanity(
+    filter_to_apply: ImageFilter.Filter | type[ImageFilter.Filter], mode: str
+) -> None:
     im = hopper(mode)
-    if mode != "I" or isinstance(filter_to_apply, ImageFilter.BuiltinFilter):
+    if mode[0] != "I" or (
+        callable(filter_to_apply)
+        and issubclass(filter_to_apply, ImageFilter.BuiltinFilter)
+    ):
```
radarhere, line 48: The `isinstance` condition, added in #7108, was never true - each `BuiltinFilter` run of this test uses a `BuiltinFilter` class, not an instance. So this corrects it.

src/libImaging/Filter.c
```diff
@@ -26,6 +26,8 @@
 
 #include "Imaging.h"
 
+#define ROUND_UP(f) ((int)((f) >= 0.0 ? (f) + 0.5F : (f) - 0.5F))
```
hugovk, line 29: Can we make this a function instead?

radarhere: Any particular reason?

This was copying https://github.com/python-pillow/Pillow/blob/27c1bb265432cb5a1138e39165f0610e2d8a4e94/src/libImaging/Reduce.c#L5 and https://github.com/python-pillow/Pillow/blob/27c1bb265432cb5a1138e39165f0610e2d8a4e94/src/libImaging/Resample.c#L5

If there's a reason to prefer functions, then maybe they should all be changed.

I'd suggest a common function for all three, except that this uses a `float` and those other two use `double`.

hugovk: Functions are generally safer than macros, which need extra care around parentheses and don't get extra checking. But if this is a copy/paste of existing ones then we should be fine.

hugovk (approved):

Issue #8439, opened 2024-10-05
Random crashes of python in latest dev version of Pillow with pillow_heif


bigcat88:
**Good day, everyone.**

I'll try to describe the situation briefly.

First, I want to clarify that I'm not certain this is a bug in Pillow. I created this issue so that knowledgeable contributors could help identify any recent changes in the repository that might be causing the problem I'm experiencing.

About two weeks ago, when using the development version of Pillow, some actions in pillow_heif repo started to fail: [run link](https://github.com/bigcat88/pillow_heif/actions/runs/10973779020).

Now that I have some time to investigate, I'm encountering a situation where the root cause of the error is still unclear to me.

Here is the script I'm using for testing:

```python
import os
import pillow_heif
from pathlib import Path
from PIL import Image, ImageSequence

pillow_heif.register_heif_opener()

os.chdir(Path(__file__).parent.parent.joinpath("tests/images/heif/"))

if __name__ == "__main__":
    im = Image.open("zPug_3.heic")
    im.load()
    for frame in ImageSequence.Iterator(im):
        print("before frame.load()")
        frame.load()
        print("after frame.load()")
        assert len(frame.tobytes()) > 0
        print("after frame.tobytes()")
    print("exit")
```

It can crash, or sometimes it does not crash.
Sometimes I even see "exit" in output and then crash occurs.

I'm running this script from the cloned `pillow_heif` repository, creating a "dev" folder at the root (if needed for quick reproduction) and placing it there.

The issue is that my Python interpreter randomly crashes when using the development version of Pillow, and I haven't been able to pinpoint the exact location of the crash.

Nothing has changed in `pillow_heif` itself in last months (I'm waiting for HDR image support before refactoring the code).

I've tried running the script with other versions of Pillow, and everything works fine. I've also tried adding print statements inside Pillow to locate the failure, but the crash seems random, and I haven't been able to identify the exact spot where it occurs.

I'm testing this on macOS (but can also reproduce it on Linux). Any insights or tips on where to investigate, or information on what may have changed in the Pillow repo 2–3 weeks ago that could lead to these crashes, would be greatly appreciated.

Some info that I have found:

1ee3bd1d9e1a96da01ba55aff376e214082bcba8 - on this commit crash already occurs
08d9c89d8a7db84e8145814bca36d7dd5b7465ce - repo at this commit is good

Thank you!

wiredfool:
[...]

PR #8450, opened 2024-10-08
In Image.Image.seek(), clear core image if mode or size has changed

radarhere:
Addresses #8439

#8390 stopped `load_prepare()` from recreating the core image instance if the mode or size had changed, as might happen at the end of a `seek()` operation. After that PR, this was no longer needed for any of our internal plugins.

However, #8439 has pointed out that this would still be helpful for external plugins. Rather than reverting the change, this PR suggests that external plugins can call `super().seek()` at the end of their `seek()` methods. That will then clear the core image instance if the mode or size has changed, allowing `load_prepare()` to detect that a new core image is needed.

The advantage of this over a simple revert is that if a user calls `seek()` on an image, but not `load()`, then the core image instance is discarded.

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -2633,9 +2633,15 @@ class Image:
         """
 
         # overridden by file handlers
-        if frame != 0:
+        if frame != self.tell():
+            # The file handler likely did not override this method
             msg = "no more images in file"
             raise EOFError(msg)
+
+        if self._im is not None and (
+            self.im.size != self.size or self.im.mode != self.mode
+        ):
+            self._im = None
 
     def show(self, title: str | None = None) -> None:
         """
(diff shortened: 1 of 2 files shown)


homm:
What about TiffImagePlugin? Can we call super().seek() there?

src/PIL/Image.py
```diff
@@ -2633,10 +2633,16 @@ def seek(self, frame: int) -> None:
         """
 
         # overridden by file handlers
-        if frame != 0:
```
homm, line 2636: Does it make impossible to call super().seek() for plugins prior Pillow 11?

radarhere:
> What about TiffImagePlugin? Can we call super().seek() there?

No, because at the end of TiffImagePlugin's `seek()`, the normal size isn't used - `self._tile_size` is.

src/PIL/Image.py
radarhere: ...yes, an external plugin would fail calling `super().seek()` if the user had Pillow < 11.

I could add a dedicated new method instead of using `super().seek()`, but any way you look at the idea of this PR, it is new functionality.

radarhere:
I don't think this is the neatest API change I've ever suggested, so if there isn't much interest, I'm happy for this to be closed.

PR #8484, opened 2024-10-18
Use fixture to re-open image for each JPEG2000 test

radarhere:
At the start of test_file_jpeg2k.py, an image is opened.
https://github.com/python-pillow/Pillow/blob/d59b169ed257ca14d86c90dbf3384b8b3999fc5c/Tests/test_file_jpeg2k.py#L32-L33

It is then used throughout the tests.
https://github.com/python-pillow/Pillow/blob/d59b169ed257ca14d86c90dbf3384b8b3999fc5c/Tests/test_file_jpeg2k.py#L103-L110

Rather than re-using the same image instance throughout the tests, and allowing it to be potentially modified in one test and affect all subsequent tests, this PR uses a fixture to re-open it each time, following the style of [test_image_resize.py.](https://github.com/python-pillow/Pillow/blob/d59b169ed257ca14d86c90dbf3384b8b3999fc5c/Tests/test_image_resize.py#L181-L188)

### DIFF
--- a/Tests/test_file_jpeg2k.py
+++ b/Tests/test_file_jpeg2k.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import os
 import re
+from collections.abc import Generator
 from io import BytesIO
 from pathlib import Path
 from typing import Any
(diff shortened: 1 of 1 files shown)


Tests/test_file_jpeg2k.py
```diff
@@ -29,8 +30,16 @@
 
 pytestmark = skip_unless_feature("jpg_2000")
 
-test_card = Image.open("Tests/images/test-card.png")
-test_card.load()
+
+@pytest.fixture
+def test_card() -> Generator[ImageFile.ImageFile, None, None]:
```
hugovk, line 35: Good idea! 

I know the old one was also prefixed `test_`, but that makes it look like a test function. Maybe just `card`? 
```suggestion
def card() -> Generator[ImageFile.ImageFile, None, None]:
```

hugovk (approved):

Tests/test_file_jpeg2k.py
radarhere: Sure, I've pushed a commit.

PR #8495, opened 2024-10-24
Replace Windows zlib with zlib-ng

dofuuz:

 * Replacing zlib with [zlib-ng](https://github.com/zlib-ng/zlib-ng) makes huge speedup for PNG encoding.

Here are my quick and dirty benchmark.
(on x64 Windows, read/write PNG 10 times)

Original (with zlib):
```
PIL.__version__='11.1.0.dev0'
read PNG: 0.092544 (sec)
on-memory PNG: 1.034144 (sec)
write PNG: 1.054776 (sec)
PNG size 1352945 bytes
```

With zlib-ng:
```
PIL.__version__='11.1.0.dev0'
read PNG: 0.089149 (sec)
on-memory PNG: 0.515073 (sec)
write PNG: 0.527402 (sec)
PNG size 1394511 bytes
```

It's about 2x speed-up with little effort.

Comparison & benchmark of zlib forks (JAN 2021):
https://aws.amazon.com/blogs/opensource/improving-zlib-cloudflare-and-comparing-performance-with-other-zlib-forks/

### DIFF
--- a/winbuild/build_prepare.py
+++ b/winbuild/build_prepare.py
@@ -104,8 +104,16 @@ SF_PROJECTS = "https://sourceforge.net/projects"
 
 ARCHITECTURES = {
     "x86": {"vcvars_arch": "x86", "msbuild_arch": "Win32", "zlibng_arch": "x86"},
-    "AMD64": {"vcvars_arch": "x86_amd64", "msbuild_arch": "x64", "zlibng_arch": "x86-64"},
-    "ARM64": {"vcvars_arch": "x86_arm64", "msbuild_arch": "ARM64", "zlibng_arch": "arm64"},
+    "AMD64": {
+        "vcvars_arch": "x86_amd64",
+        "msbuild_arch": "x64",
+        "zlibng_arch": "x86-64",
+    },
+    "ARM64": {
+        "vcvars_arch": "x86_arm64",
+        "msbuild_arch": "ARM64",
+        "zlibng_arch": "arm64",
+    },
 }
 
 V = {
@@ -161,7 +169,7 @@ DEPS: dict[str, dict[str, Any]] = {
     },
     "zlib": {
         "url": f"https://github.com/zlib-ng/zlib-ng/releases/download/{V['ZLIBNG']}/zlib-ng-win-{{zlibng_arch}}-compat.zip",
-        "filename": f"zlib-ng-win-{{zlibng_arch}}-compat.zip",
+        "filename": "zlib-ng-win-{zlibng_arch}-compat.zip",
         "dir": "",
         "release": f"zlib-ng-{V['ZLIBNG']}",
         "license": "LICENSE.md",
--- a/winbuild/build_prepare.py
+++ b/winbuild/build_prepare.py
@@ -120,11 +120,10 @@ V = {
     "OPENJPEG": "2.5.2",
     "TIFF": "4.6.0",
     "XZ": "5.6.3",
-    "ZLIB": "1.3.1",
+    "ZLIBNG": "2.2.2",
 }
 V["LIBPNG_DOTLESS"] = V["LIBPNG"].replace(".", "")
 V["LIBPNG_XY"] = "".join(V["LIBPNG"].split(".")[:2])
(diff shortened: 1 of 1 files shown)


nulano:
The changes look OK (although we might want to consider switching to the CMake build at some point as it seems to be more actively maintaned).

We probably want to add a flag in `_imaging.c`/`PIL.features` similar to the libjpeg-turbo flag:
https://github.com/python-pillow/Pillow/blob/822ec3d05dfbaf2e3209c2b26e3392040e274975/src/_imaging.c#L4352-L4368
https://github.com/python-pillow/Pillow/blob/822ec3d05dfbaf2e3209c2b26e3392040e274975/src/PIL/features.py#L299-L302

It is [currently reporting zlib version `1.3.1.zlib-ng`](https://github.com/python-pillow/Pillow/actions/runs/11500002895/job/32016382938?pr=8495#step:27:33):
```
--- JPEG support ok, compiled for libjpeg-turbo 3.0.4
--- ZLIB (PNG/ZIP) support ok, loaded 1.3.1.zlib-ng
```

---

> Here are my quick and dirty benchmark.

Can you provide the benchmark itself? (I imagine its only a couple of lines, but it would make it easier to compare with others. Using one of our test images could make it even simpler.)

It looks like zlib-ng produces a slightly larger file. This is probably fine, but I would be curious what effect changing the `compress_level` parameter has.

Also, do note that the FreeType library also uses zlib for decompressing font files, although I do not expect any issues there.

dofuuz:
ARM Windows build fails, so I think we should switch to CMake build.
OR just use pre-compiled release from zlib-ng like https://github.com/zlib-ng/zlib-ng/releases/download/2.2.1/zlib-ng-win-x86-64-compat.zip

For version number, zlib-ng reports compatible zlib version in zlib-compat API.

Here is the bench script:
```py
import os
import timeit

from PIL import Image


REPEAT = 10

print(f'{Image.__version__ = }')
print(f'{Image.core.zlib_version = }')

def read_png():
    img = Image.open('example.png')
    img.load()

t = timeit.timeit(read_png, number=REPEAT)
print(f'read PNG: {t:f} (sec)')

img = Image.open('example.png')
img.load()

fp = open(os.devnull, "wb")
t = timeit.timeit(lambda: img.save(fp, format='png'), number=REPEAT)
print(f'on-memory encoding PNG: {t:f} (sec)')

t = timeit.timeit(lambda: img.save('pil_png.png'), number=REPEAT)
print(f'write PNG: {t:f} (sec)')

print(f'PNG size {os.path.getsize("pil_png.png")} bytes')
```

hugovk:
> Comparison & benchmark of zlib forks (JAN 2021):
> https://aws.amazon.com/blogs/opensource/improving-zlib-cloudflare-and-comparing-performance-with-other-zlib-forks/

Is it worth re-benchmarking these other forks or finding some newer results?

nulano:
> For version number, zlib-ng reports compatible zlib version in zlib-compat API.

My point being that we should report the zlib-ng version when using it, not the version of the zlib-compat API. This is to make it clear which version is used in tests and bug reports.

nulano:
I have created #8500 using the CMake build instead, as we can use a newer version that way. I also added a feature flag so that the use of zlib-ng is clear from running `python3 -m PIL.report`.

> Here is the bench script

Thanks for that, I've adjusted it slightly to also test at various compression levels. It seems that in compression, zlib-ng usually produces a slightly larger file but in much less time.

radarhere:
@dofuuz are you happy for this to be closed in favour of #8500?

dofuuz:
@radarhere 
Yes. 👍
But why not merge #8500 already?

radarhere:
If we're going to start using zlib-ng, then I'd like to see if we can use it on other platforms as well - I've created https://github.com/nulano/Pillow/pull/42 to add it on Linux for example.

The next Pillow release is scheduled for January 2, so as long as it is merged before then, we should be all good.

radarhere:
https://github.com/python-pillow/Pillow/pull/8500 has now been merged.

PR #8496, opened 2024-10-24
Fix SEGFAULT from calling FT_New_Face/FT_Done_Face in multiple threads

### LINKED ISSUE #8454: Segfault under free-threaded build
See https://github.com/tonybaloney/pytest-freethreaded/issues/9 where this was discovered and for steps to reproduce.

Ultimately this comes down to the thread safety warnings in the freetype docs for `FT_Face`: https://freetype.org/freetype2/docs/reference/ft2-face_creation.html#ft_face. Somehow an `FT_Face` is being shared across threads in the test scenario linked above.

Ping @lysnikolaou

lysnikolaou:
Fixes #8454.

 * Introduce a mutex that needs to be held every time `FT_New_Face` or `FT_Done_Face` are called.
 * Upgrade `pythoncapi_compat.h` to the latest version.

### DIFF
--- a/src/_imagingft.c
+++ b/src/_imagingft.c
@@ -83,7 +83,7 @@ struct {
 
     static FT_Library library;
 #ifdef Py_GIL_DISABLED
-PyMutex ft_library_mutex;
+static PyMutex ft_library_mutex;
 #endif
 
 typedef struct {
@@ -1648,7 +1648,6 @@ PyInit__imagingft(void) {
     }
 
 #ifdef Py_GIL_DISABLED
-    ft_library_mutex = (PyMutex){0};
     PyUnstable_Module_SetGIL(m, Py_MOD_GIL_NOT_USED);
 #endif
 
--- a/src/_imagingft.c
+++ b/src/_imagingft.c
@@ -81,7 +81,10 @@ struct {
     /* -------------------------------------------------------------------- */
     /* font objects */
 
-    static FT_Library library;
+static FT_Library library;
+#ifdef Py_GIL_DISABLED
+PyMutex ft_library_mutex;
+#endif
 
 typedef struct {
     PyObject_HEAD FT_Face face;
@@ -187,7 +190,9 @@ getfont(PyObject *self_, PyObject *args, PyObject *kw) {
 
     if (filename && font_bytes_size <= 0) {
         self->font_bytes = NULL;
+        MUTEX_LOCK(&ft_library_mutex);
         error = FT_New_Face(library, filename, index, &self->face);
+        MUTEX_UNLOCK(&ft_library_mutex);
     } else {
         /* need to have allocated storage for font_bytes for the life of the object.*/
         /* Don't free this before FT_Done_Face */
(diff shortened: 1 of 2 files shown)


hugovk (approved):
Thanks!

src/thirdparty/pythoncapi_compat.h
```diff
@@ -7,7 +7,10 @@
 // https://github.com/python/pythoncapi_compat
 //
 // Latest version:
-// https://raw.githubusercontent.com/python/pythoncapi_compat/master/pythoncapi_compat.h
+// https://raw.githubusercontent.com/python/pythoncapi-compat/main/pythoncapi_compat.h
+//
+// This file was vendored from the following commit:
+// https://github.com/python/pythoncapi-compat/commit/0041177c4f348c8952b4c8980b2c90856e61c7c7
```
nulano, line 13: This commit appears to be already merged in the latest version of this file.

nulano (commented):
Is anything here actually using the new changes in `pythoncapi_compat.h`?

There's no reason to stick with an old version, but I'm curious if there is a reason for updating it in what looks like an unrelated PR?

src/thirdparty/pythoncapi_compat.h
lysnikolaou: This is what I wanted to convey in my comment. That the version of the file came from this specific commit.

nulano: Ah, I see now, thanks for the explanation.

src/_imagingft.c
```diff
@@ -1639,6 +1648,7 @@ PyInit__imagingft(void) {
     }
 
 #ifdef Py_GIL_DISABLED
+    ft_library_mutex = (PyMutex){0};
```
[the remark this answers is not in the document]
lysnikolaou: That's because assignment without a cast is not possible when it's not initialization.

We can actually get rid of assigning to the mutex altogether by declaring the `PyMutex` as static, which is the recommended way and what needs to be done here. Sorry I missed this and thanks for pointing it out.

src/_imagingft.c
```diff
@@ -82,6 +82,9 @@ struct {
     /* font objects */
 
     static FT_Library library;
+#ifdef Py_GIL_DISABLED
+static PyMutex ft_library_mutex;
```
nulano, line 86:
```suggestion
static PyMutex ft_library_mutex = {0};
```
?

radarhere (approved):

src/_imagingft.c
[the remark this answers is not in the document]
radarhere: Just for more information, could you link to that recommendation? I'm not seeing anything explicit at https://docs.python.org/3/c-api/init.html#synchronization-primitives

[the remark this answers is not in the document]
lysnikolaou: The recommendation is in a header comment in the CPython code [here](https://github.com/python/cpython/blob/dc76a4ad3c46af3fb70361be10b8a791cd126931/Include/cpython/lock.h#L15).

Issue #7942, opened 2024-04-04
Update license identifier to MIT-CMU


capfei:
The license states HPND and I saw a PR from last year to get that text to match closer to what is listed on SPDX. However, I see that the Pillow license text actually matches what SPDX calls MIT-CMU (https://spdx.org/licenses/MIT-CMU.html) because of the additional text that is not included in HPND:
> By obtaining, using, and/or copying this software and/or its associated
> documentation, you agree that you have read, understood, and will comply
> with the following terms and conditions:

For clarity, would it make sense to change `Like PIL, Pillow is licensed under the open source HPND License` to be MIT-CMU or maybe add the SPDX identifier to the license?
`SPDX Identifier: MIT-CMU`

hugovk:
The [PIL licence](https://github.com/python-pillow/Pillow/blob/main/LICENSE) is closer to[MIT-CMU](https://spdx.org/licenses/MIT-CMU.html) than [HPND](https://spdx.org/licenses/HPND.html).

PIL vs. HPND:

<img width="1235" alt="image" src="https://github.com/python-pillow/Pillow/assets/1324225/ae557013-c330-4ee3-a84b-cf2af29973b1">

PIL vs. MIT-CMU:

<img width="1233" alt="image" src="https://github.com/python-pillow/Pillow/assets/1324225/12649647-97ec-4077-b100-1fa727fe3801">

tieguy:
Yeah, I validated this now with an automated tool (eyeballed it when we did this yearrrrrs ago) and it indeed more correctly labeled as MIT-CMU. Sorry for the extra work, @aclark4life !

aclark4life:
> Yeah, I validated this now with an automated tool (eyeballed it when we did this yearrrrrs ago) and it indeed more correctly labeled as MIT-CMU. Sorry for the extra work, @aclark4life !

No trouble at all, but just so I understand:

- The "Standard PIL License" is actually MIT-CMU
- We thought the "Standard PIL License" was HPND but that was a mistake so now we should fix that mistake by change all of our references to "Standard PIL License" or HPND to MIT-CMU

At least that's my current understanding. Either way, the answer to [questions like this one](https://opensource.stackexchange.com/questions/14407/how-can-i-comply-with-the-terms-of-the-hpnd-license-if-i-am-using-a-library-that) are still the same: comply with the terms of the license! I mean this is pretty clear whatever you call it:

[...]

Issue #7250, opened 2023-07-02
10.0.0: New Mach-O header is too large to relocate on py2app


MAKOMO:
### What did you do?

Bundling my app with py2app on macOS which fails under Pillow 10.0.0 with above message while it works with 9.5.0

### What did you expect to happen?

A working app bundle

### What actually happened?

On relocating py2app complaint

```
ValueError: New Mach-O header is too large to relocate in '/Users/appveyor/projects/artisan/src/dist/Artisan.app/Contents/Resources/lib/python3.11/PIL/.dylibs/libwebp.7.dylib' (new size=1768, max size=1760, delta=40)
```
### What are your OS, Python and Pillow versions?

* OS: macOS
* Python: 3.11
* Pillow: 10.0.0

The author of py2app stated on similar issues

> ..in general this can be fixed upstream by adding `-Wl, headerpad_max_install_names` to the linker flags. This will increase padding in the Mach-O headers.

see https://github.com/ronaldoussoren/py2app/issues/219#issuecomment-1305142378

radarhere:
What version of macOS and py2app are you using?

MAKOMO:
- pip installed current py2app==0.28.6  (I am aware that a new major version is under development, but not released yet)
- macOS macos-bigsur 11.7.1 on appveyor 

I just triggered a build locally on Intel macOS 13.4.1 resulting in

```
ValueError: New Mach-O header is too large to relocate in '/Users/luther/Documents/Projects/Artisan/Releases/2.8.4/artisan-2.8.4/src/dist/Artisan.app/Contents/Resources/lib/python3.11/PIL/.dylibs/libwebp.7.dylib' (new size=1768, max size=1760, delta=40)
```

Again, downgrading to Pillow 9.5 suceeds.

Pillow is not imported directly but via the current pypip matplotlib.

It is difficult for me to come up with a small script to demonstrate the issue as this is happening on building a relatively large project. But if I can assist you with any experiment or further data, let me know.

radarhere:
When I was able to reproduce this, I found that downgrading libwebp from 1.3.1 to 1.3.0 (reverting https://github.com/python-pillow/pillow-wheels/pull/392) fixed it.

[...]

PR #6562, opened 2022-09-07
Windows build improvements

nulano:
For #6369, close #6554, close #553

In an effort to sync the dependencies available on Linux/macOS ([python-pillow/pillow-wheels](https://github.com/python-pillow/pillow-wheels)) and Windows ([test-windows.yml](https://github.com/python-pillow/Pillow/blob/main/.github/workflows/test-windows.yml)) ahead of the next release (#6460), I've compiled a list of dependencies (including indirect ones) below, and included some updates for Windows in this PR.

<details>
<summary>The full list is: (click to expand)</summary>

* `libjpeg` via its fork `libjpeg-turbo` version 2.1.4 is included on all platforms
* `zlib` version 1.2.12, included on all platforms
* `libtiff` version 4.4.0, included on all platforms
    * uses `zlib` and `libjpeg` on all platforms
    * Linux/macOS uses `lzma2` version 5.2.6 from `xz`; added on Windows in this PR and added test
    * no platform uses `webp` because that is built after `libtiff` on both platforms; swapped on Windows in this PR and added test
    * no current wheels include `deflate`, `jbig`, `libjpeg12`, `lerc`, `zstd`, but cgohlke mentioned at least LERC being included in his wheels in https://github.com/python-pillow/Pillow/issues/5573#issuecomment-872691741
 * `webp` version 1.2.4, included on all platforms
    * AFAICT its dependencies `libjpeg`, `libpng`, `libtiff`, and `giflib` are only used for binary conversion utilities, not `libwebp`. However, on Linux/macOS multibuild ensures these are built before `webp`. For Windows I moved `webp` to be built before `libtiff`.
 * `freetype` version 2.12.1, included on all platforms
    * uses `zlib` and `libpng` on all platforms
    * uses `harfbuzz` only on Windows because the cyclic dependency is tricky to resolve with GCC
    * only Linux/macOS uses `bzip2` version 1.0.8. I tried adding it on Windows, but the driver `ftbzip2` appears to have been left out of the Visual Studio project files. It is only used for `.pcf.bz2` fonts which aren't that common anyway (especially on Windows).
    * `brotli` version 1.0.9 added in https://github.com/python-pillow/pillow-wheels/pull/320 for Linux/macOS and this PR for Windows
 * `lcms2` version 2.13.1
    * AFAICT its dependencies `libjpeg`, `zlib`, `libtiff` are only used in binary utilities, not the library
 * `openjpeg` version 2.5.0
[...]

Issue #6554, opened 2022-09-01
Unknown file format error on some `woff2` fonts on `ImageFont.truetype()`


mirusu400:
### What did you do?

I'm trying to convert woff and woff2 fonts to `ImageFont`

### What did you expect to happen?

Return a `FreeTypeFont` object

### What actually happened?

It raises `OSError` error, terminates program.

### What are your OS, Python and Pillow versions?
```
ProductName:	macOS
ProductVersion:	12.4
BuildVersion:	21F79
Python 3.9.13 (main, May 24 2022, 21:28:31)
[Clang 13.1.6 (clang-1316.0.21.2)] on darwin
Pillow==9.2.0
```

Here are some sample code represents my issue:
```python
import requests
import os
from PIL import ImageFont

r = requests.get("https://cdn.jsdelivr.net/korean-webfonts/1/orgs/othrs/kywa/Youth/Youth.woff2", stream=True)
with open("Youth.woff2", "wb") as f:
    f.write(r.content)

print("Trying to get Font ...\t", end="...")

try:
    font = ImageFont.truetype("Youth.woff2")
    print("Success")
    os.remove("Youth.woff2")
except Exception as e:
    print("Failed " + str(e))
    os.remove("Youth.woff2")
    raise e
```

```
Traceback (most recent call last):
  File "/Users/mirusu400/kwoss/pillow-woff-issue/test2.py", line 18, in <module>
    raise e
  File "/Users/mirusu400/kwoss/pillow-woff-issue/test2.py", line 12, in <module>
    font = ImageFont.truetype("Youth.woff2")
  File "/usr/local/lib/python3.9/site-packages/PIL/ImageFont.py", line 959, in truetype
    return freetype(font)
  File "/usr/local/lib/python3.9/site-packages/PIL/ImageFont.py", line 956, in freetype
    return FreeTypeFont(font, size, index, encoding, layout_engine)
  File "/usr/local/lib/python3.9/site-packages/PIL/ImageFont.py", line 247, in __init__
    self.font = core.getfont(
OSError: unknown file format
```

I don't know detailed reason this happen, So I created a repo about this issue.
https://github.com/mirusu400/pillow-woff-issue
Here are some samples about having ImageFont issue.

radarhere:
Thanks for your detailed example.

[...]

PR #8497, opened 2024-10-25
Isolate macOS wheel builds from Homebrew

freakboy3742:
The existing macOS `cibuildwheel` configuration relies on Homebrew to provide *some* dependencies. This clearly works in practice, but there are some issues associated with this choice.

Firstly, it is destructive when used on a local build machine. The wheel dependency script invokes `brew remove --ignore-dependencies` (which can leave the host machine in a broken, and potentially difficult to restore state); and requires the use of `sudo` to make changes in the `/usr/local` tree. This means it is difficult to test `cibuildwheel` changes locally, or to recommend using `cibuildwheel` as a local build solution.

Secondly, Homebrew builds could be incompatible with Pillow builds. The `MACOSX_DEPLOYMENT_TARGET` for Homebrew dependencies is fixed by the Homebrew build chain, rather than the Pillow build tools, so a change in Pillow's configuration may not be satisfied by Homebrew's build configuration. This isn't an issue at present, but it could easily become one in future.

Related to this, there is also the potential that a change in a Homebrew recipe could lead to the inadvertent introduction of additional binary libraries into the delocated macOS binaries. This is especially problematic with the fribidi, raqm and freetype dependedencies which are included as "vendor" sources.

However, the *real* motivation for proposing this change is that it is a precursor for PEP 730-compliant iOS builds. If Homebrew is on the build path when compiling iOS binaries, compiler tooling will often find an ARM64 Homebrew binary and attempt link it into an iOS library - which then (predictably) breaks. So - Homebrew isolation is essential for iOS compilation. Since the build processes for iOS and macOS are very similar, and Homebrew isolation was necessary for iOS, adding Homebrew isolation for macOS will simplify any future iOS patch. It also provides a way to submit iOS support in a series of smaller pieces, rather than a single "monster" patch. Plus, it provides all the side benefits of isolation described above.

An overview of the changes:

* Uses the `build` folder as a location for all dependency builds. This prevents the root checkout from becoming dirtied by dependency artefacts, simplifying cleanup, and avoiding the need for extensive modifications to the `.gitignore` file.

* Sets up an isolated build prefix in the `build` folder (`./build/deps/darwin`) where dependencies can be installed, and `make installs` all dependencies into that path.

* Forces `PATH` to be a "clean" environment that only includes bare system tools, plus the Python binary being used for the build, and the isolated build prefix. macOS *doesn't* include most of it's development libraries in /usr or /usr/local, instead using a path provided by the macOS SDK (which is configured as part of the compiler toolchain).

* Adds a build of `pkg-config` to the dependencies. This is the one build tool that autotools and cmake often require that Xcode doesn't provide. Providing a custom build of `pkg-config` also ensures that no dependencies other than the ones provided by cibuildwheel are included. The build recipe that is used is derived from the Homebrew recipe.

* ~Adds a build of `libXdmcp`, to avoid a dependency on the Homebrew-provided version.~

* ~Adds builds of `fribidi` and `raqm`, rather than using the Homebrew versions as "vendored" versions. The `setup.py` configuration passed in by `cibuildwheel` has been modified from the default on other platforms to *not* use the `vendor` version.~

* Updates the pinned multibuild version to the current `devel` branch hash. This is to incorporate Python3.13 support, plus a number of fixes required to support installs into locations other than `/usr/local`, and corrections to LIBDIR handling for platforms that use the `lib64` suffix.

* ~Adds a `dependencies-prefix` configuration option to `setup.py`. This allows the user to provide a specific location to look for dependencies, rather than the series of Fink/MacPorts/Homebrew fallbacks that currently exist.~


.github/workflows/wheels-test.sh
```diff
@@ -1,16 +1,13 @@
 #!/bin/bash
 set -e
 
-if [[ "$OSTYPE" == "darwin"* ]]; then
-    brew install fribidi
-    export PKG_CONFIG_PATH="/usr/local/opt/openblas/lib/pkgconfig"
-    if [ -f /opt/homebrew/lib/libfribidi.dylib ]; then
-        sudo cp /opt/homebrew/lib/libfribidi.dylib /usr/local/lib
+# For Unix, Ensure fibidi is installed by the system.
```
radarhere, line 4:
```suggestion
# For Unix, ensure fribidi is installed by the system.
```

.github/workflows/wheels-dependencies.sh
```diff
@@ -23,6 +23,8 @@ OPENJPEG_VERSION=2.5.2
 XZ_VERSION=5.6.3
 TIFF_VERSION=4.6.0
 LCMS2_VERSION=2.16
+RAQM_VERSION=0.7.1
```
radarhere, line 26:
```suggestion
RAQM_VERSION=0.10.2
```

freakboy3742: It definitely makes sense to use the most recent available version; but pillow-depends-main doesn't include this version. What's the procedure for updating that repo?

(Related: pkg-config, libXdmcp, and an updated webp sources should probably be added to that project)

freakboy3742:
The manylinux failures look like they'll be fixed by [this PR](https://github.com/multi-build/multibuild/pull/537) to multi build.

The x86_64 macOS failures are probably a leakage from /usr/local that I haven't accounted for. I'm looking into that one now.

I'm not sure how to interpret the coverage drop, though...

freakboy3742:
@radarhere I've updated the pin for multibuild to include the (just merged) PR reverting the libjpeg-turbo CMAKE_INSTALL_LIBDIR change; and I've addressed some additional locations where Homebrew can leak in on x86_64.

.github/workflows/wheels-dependencies.sh
radarhere: The theory is that pillow-depends-main is just a more reliable source for the dependencies, and PRs should work without it.

Once this is merged, I will likely add them there.

nulano:
> Related to this, there is also the potential that a change in a Homebrew recipe could lead to the inadvertent introduction of additional binary libraries into the delocated macOS binaries. 

This has already happened many times in the past, so thanks for working on this.

> Adds builds of fribidi and raqm, rather than using the Homebrew versions as "vendored" versions. The setup.py configuration passed in by cibuildwheel has been modified from the default on other platforms to not use the vendor version.

I'm not sure if I understand what you mean here. We need to use a vendored version of raqm in wheels for license reasons. When using `raqm=vendor`, we build our own version of raqm from https://github.com/python-pillow/Pillow/tree/main/src/thirdparty/raqm. Similarly, using `fribidi=vendor` does not actually build fribidi, but instead custom code that can load fribidi at runtime, see https://github.com/python-pillow/Pillow/tree/main/src/thirdparty/fribidi-shim. The Homebrew version of fribidi is only used for testing the built wheels.

freakboy3742:
> > Adds builds of fribidi and raqm, rather than using the Homebrew versions as "vendored" versions. The setup.py configuration passed in by cibuildwheel has been modified from the default on other platforms to not use the vendor version.
> 
> I'm not sure if I understand what you mean here. We need to use a vendored version of raqm in wheels for license reasons. When using `raqm=vendor`, we build our own version of raqm from https://github.com/python-pillow/Pillow/tree/main/src/thirdparty/raqm. Similarly, using `fribidi=vendor` does not actually build fribidi, but instead custom code that can load fribidi at runtime, see https://github.com/python-pillow/Pillow/tree/main/src/thirdparty/fribidi-shim. The Homebrew version of fribidi is only used for testing the built wheels.

Thanks - that's some helpful context. I was seeing the missing fribidi detection during the build and interpreting that as something that was part of the move away from homebrew; but I see now that Pillow is loading from Homebrew *by design* in this instance. That also might explain the coverage drop (as supplying fribidi means any runtime loading code won't be in use). I'll take another swing at that part (and resolve the linux build issues as well).

Out of interest - what's the license issue with raqm? The source code indicates It looks to be MIT licensed... did it used to be GPL (or something else problematic?)

hugovk:
> Out of interest - what's the license issue with raqm? The source code indicates It looks to be MIT licensed... did it used to be GPL (or something else problematic?)

From https://github.com/python-pillow/Pillow/pull/2753 in 2017:

> We can't currently distribute a binary of Pillow with support for raqm due to license incompatibility, because while libraqm is MIT licensed, it links to GPL libraries.

FriBiDi is LGPL. 

Since then, Raqm added support for Apache-licensed SheenBiDi in 2021: https://github.com/HOST-Oman/libraqm/issues/138, https://github.com/HOST-Oman/libraqm/pull/139.

Perhaps we could switch.

freakboy3742:
@radarhere After a lot of false starts (apologies for the noise, BTW...), I think I've finally got it all sorted. 

1. macOS is isolated from Homebrew on both x86_64 and ARM64
2. The shim fribidi builds is used in the wheel, with Homebrew providing the fribidi binary on macOS in the test environment
3. Linux builds are now correctly using `BUILD_PREFIX`, and don't require the `lib64` copying step.

.github/workflows/wheels-test.sh
```diff
@@ -1,12 +1,20 @@
...
 set -e
 
+# Ensure fribidi is installed by the system.
 if [[ "$OSTYPE" == "darwin"* ]]; then
-    brew install fribidi
-    export PKG_CONFIG_PATH="/usr/local/opt/openblas/lib/pkgconfig"
-    if [ -f /opt/homebrew/lib/libfribidi.dylib ]; then
-        sudo cp /opt/homebrew/lib/libfribidi.dylib /usr/local/lib
+    # If Homebrew is on the path during the build, it may leak into the wheels.
+    # However, we need a *do* need Homebrew to provide a copy of fribidi for
```
radarhere, line 7:
```suggestion
    # However, we *do* need Homebrew to provide a copy of fribidi for
```

freakboy3742:
I'm not sure I understand how coverage can go *down* when you don't touch any code or tests... and the codecov report isn't really helping explain what is going on.

hugovk:
Hmm, let's see... (Short version: I think it's some problem with Codecov and not this PR.)

<details>
<summary>Codecov details</summary>

Looking at the coverage report at https://github.com/python-pillow/Pillow/pull/8497/checks?check_run_id=32242849880 it says "90.36% (-3.98%) compared to 2d1d801" which is already suspicious: our coverage is around 90%, not 94%.

That 2d1d801 is the last commit on `main`: "Update CHANGES.rst [ci skip]"

That "[ci skip]" did in fact skip the CI, but I see @radarhere later triggered a manual build of that commit at https://github.com/python-pillow/Pillow/actions/runs/11573601676 for only Windows, presumably checking the PyPy failure was also happening on `main` and not just on (another) failing PR.

Anyway, this Windows-only CI resulted in a skewed 94% coverage report at https://app.codecov.io/github/python-pillow/Pillow/commit/2d1d801ec032bfb760bede88da6dcaab5ff3c75e (see the flags on the right only show Windows and not the others).

All of which suggests this is nothing to worry about. Having said that, these jobs ran about 9 hours ago and @freakboy3742's comment is about 14 hours ago, so this digging doesn't reflect the situation 14 hours ago!

I will say that we've been seeing drops in other PRs where it doesn't quite make sense. I've not had time to investigate, I wonder if we're not setting tokens properly for some builds after Codecov changed their v4 action to be more strict.

</details>

freakboy3742:
@hugovk /me casually drops [this article](https://hynek.me/articles/ditch-codecov-python/) into the chat :-)

FWIW - BeeWare ditched Codecov a couple of years ago specifically because of weird inconsistencies like this one; and although the output of a raw coverage report is marginally less pretty (i.e., no annotated code listings), you can get the same report locally that you get in CI, and you get *none* of the weird outages and reporting issues that seem to plague Coverage. YMMV; but if there's interest, it's a relatively straightforward switch, especially if coverage reporting is already in place.

freakboy3742:
The (seemingly inconsistent) Windows/pypy3.10 test failure also fascinates me... is this the same cache issue mentioned by #8513?

.github/workflows/wheels-test.sh
```diff
@@ -1,12 +1,20 @@
...
 if [[ "$OSTYPE" == "darwin"* ]]; then
-    brew install fribidi
-    export PKG_CONFIG_PATH="/usr/local/opt/openblas/lib/pkgconfig"
-    if [ -f /opt/homebrew/lib/libfribidi.dylib ]; then
-        sudo cp /opt/homebrew/lib/libfribidi.dylib /usr/local/lib
+    # If Homebrew is on the path during the build, it may leak into the wheels.
+    # However, we *do* need Homebrew to provide a copy of fribidi for
+    # testing purposes so that we can verify the fribidi shim works as expected.
+    if [[ "$(uname -m)" == "x86_64" ]]; then
+        HOMEBREW_HOME=/usr/local/homebrew
```
radarhere, line 10:
```suggestion
        HOMEBREW_HOME=/usr/local
```
I would have thought just '/usr/local', based on https://docs.brew.sh/Installation
> The script installs Homebrew to its default, supported, best prefix (/opt/homebrew for Apple Silicon, /usr/local for macOS Intel

hugovk:
> The (seemingly inconsistent) Windows/pypy3.10 test failure also fascinates me... is this the same cache issue mentioned by #8513?

Yep, we added a new licence (well, old: MIT CMU) to OSI and Trove classifiers, so that test needs the new `trove-classifiers`. https://github.com/python-pillow/Pillow/issues/7942 has the details.

I've opened https://github.com/python-pillow/Pillow/pull/8514 as an alternative to https://github.com/python-pillow/Pillow/pull/8513.

.github/workflows/wheels-test.sh
freakboy3742: Turns out both work (for `brew`) at least - Homebrew keeps the "original" versions in /usr/local/homebrew. However, I agree that `/usr/local` is the generally intended entry point.

freakboy3742: I take that back - it *does* matter (and making this change caused test failures on x86_64).

Firstly, it's possible for homebrew to have packages *installed*, but not *linked*. Using the `/usr/lib/Homebrew` location ensures that the binary can be used regardless (note also the capitalisation fix - doesn't strictly matter on HFS+, but for accuracy).

Secondly, it's possible for *other* libraries to be in `/usr/local/lib`. Using `/usr/local/Homebrew/lib` for the `DYLD_LIBRARY_PATH` avoids this.

nulano:
> * The shim fribidi builds is used in the wheel, with Homebrew providing the fribidi binary on macOS in the test environment

Thanks, that looks better.

---
There still seems to be something wrong with freetype on arm64, https://github.com/python-pillow/Pillow/actions/runs/11604707482/job/32314276332?pr=8497#step:5:14308
```
  SKIPPED [2] Tests/test_imagefont.py:1081: FreeType compiled without brotli or WOFF2 support
```
But it was built with brotli, https://github.com/python-pillow/Pillow/actions/runs/11604707482/job/32314276332?pr=8497#step:5:7768:
```
    brotli:        yes (pkg-config)
```
We need to support this for #6554.
---
We could also consider compiling libwebp before libtiff so that WEBP-compressed tiff files can be read (AFAIK webp only looks for libtiff so that it can compile companion tools, but does not actually use it in the library - as I mentioned in #6562). But we don't need to do it now since the test was already being skipped on macOS and Linux before this PR.
https://github.com/python-pillow/Pillow/actions/runs/11604707482/job/32314276332?pr=8497#step:5:15489:
```
  SKIPPED [1] Tests/test_file_libtiff.py:905: WEBP compression support is not configured
```

freakboy3742:
> There still seems to be something wrong with freetype on arm64, https://github.com/python-pillow/Pillow/actions/runs/11604707482/job/32314276332?pr=8497#step:5:14308

Weird. My immediate guess is that it might have something to do with the fact that a vendored version of fribidi is being used, which might *not* be triggering brotli support (or at least not triggering the test). I'll take a look and see what is going on. 

> We could also consider compiling libwebp before libtiff so that WEBP-compressed tiff files can be read (AFAIK webp only looks for libtiff so that it can compile companion tools, but does not actually use it in the library - as I mentioned in #6562). But we don't need to do it now since the test was already being skipped on macOS and Linux before this PR. https://github.com/python-pillow/Pillow/actions/runs/11604707482/job/32314276332?pr=8497#step:5:15489:

I saw that, and ultimately gave up trying to fix it - firstly because I was trying to limit the scope of the changes, but also because I'm not sure it *can* be fixed. libwebp has a dependency on tiff... but then lcms2 needs tiff; and openjpeg needs libpng, tiff, and lcms2. So - I'm not sure this *can* be fixed (or, at least, it can't be fixed without some sort of 2-pass compile and the use of dynamic linking, which iOS can't use anyway).

nulano:
> Weird. My immediate guess is that it might have something to do with the fact that a vendored version of fribidi is being used, which might not be triggering brotli support (or at least not triggering the test). I'll take a look and see what is going on.

The test skip is triggered by a specific error raised by the freetype library which itself does not use fribidi.

> libwebp has a dependency on tiff... but then lcms2 needs tiff; and openjpeg needs libpng, tiff, and lcms2

AFAIK all of these dependencies are just for the binary utilities that we don't need so we can disable them with relevant compile flags; the libraries themselves do not have dependencies between them (except for libtiff optionally using libwebp). But I agree that it might be best to leave that for a separate PR.

freakboy3742:
@nulano I've found the culprit - Homebrew's copy of freetype was leaking into the test environment.

By using `DYLD_LIBRARY_PATH` pointed at Homebrew's lib foldler, *all* of Homebrew's libraries become available, and take take precedence over the ones provided by the wheel. As a result, Homebrew's copy of freetype is used, which *doesn't* include Brotli support. I presume this wasn't an issue previously because an explicit `DYLD_LIBRARY_PATH` comes earlier in library resolution precedence; when Homebrew was being used implicitly, Homebrew's library resolve *after* Pillow's bundled dependencies.

This can be fixed by pointing the `DYLD_LIBRARY_PATH` at the *cellar* version of the vendored fribidi library (which is the only library we actually need from Homebrew). I've pushed that update (update: I'm looking at the linting error now...)

The update includes 2 other small changes:

1. The location of the "homebrew isolated" macOS build is slightly altered - the extra `darwin` folder became necessary when I started working on the iOS build, and the build folders for macOS started colliding with iOS. 
2. I've added some safety checks against some uses that are valid cibuildwheel invocations, but will break if you try to use them. Specifically, `CIBW_ARCHS="arm64 x86_64"`, and not specifying `CIBW_ARCHS` at all, are both valid cibuildwheel setup, but Pillow doesn't allow it because the `before_all` step requires a single explicit architecture. This bit me a couple of times before I worked out what was happening; I figured it couldn't hurt to raise an explicit error when we know it won't work.

.github/workflows/wheels-dependencies.sh
```diff
@@ -1,11 +1,34 @@
...
-# Test for macOS with [ -n "$IS_MACOS" ]
-if [ -z "$IS_MACOS" ]; then
-    export MB_ML_LIBC=${AUDITWHEEL_POLICY::9}
-    export MB_ML_VER=${AUDITWHEEL_POLICY:9}
+
+# Setup that needs to be done before multibuild utils are invoked
+PROJECTDIR=$(pwd)
+if [[ "$(uname -s)" == "Darwin" ]]; then
+    # Safety check - macOS builds require that CIBW_ARCHS is set, and that it
+    # only contains a single value (even though cwbuildwheel) allows multiple
```
radarhere, line 7:
```suggestion
    # only contains a single value (even though cibuildwheel) allows multiple
```

nulano:
Thanks, looks good to me now.

.github/workflows/wheels-dependencies.sh
```diff
@@ -1,11 +1,34 @@
...
-if [ -z "$IS_MACOS" ]; then
-    export MB_ML_LIBC=${AUDITWHEEL_POLICY::9}
-    export MB_ML_VER=${AUDITWHEEL_POLICY:9}
+
+# Setup that needs to be done before multibuild utils are invoked
+PROJECTDIR=$(pwd)
+if [[ "$(uname -s)" == "Darwin" ]]; then
+    # Safety check - macOS builds require that CIBW_ARCHS is set, and that it
+    # only contains a single value (even though cibuildwheel) allows multiple
+    # values in CIBW_ARCHS.
```
radarhere, line 8:
```suggestion
    # only contains a single value (even though cibuildwheel allows multiple
    # values in CIBW_ARCHS).
```

.github/workflows/wheels-dependencies.sh
```diff
@@ -77,67 +109,88 @@ function build {
...
+fi
 
-  brew install pkg-config
+if [[ -n "$IS_MACOS" ]]; then
+    # Homebrew (or similar packaging environments) install can contain some of
+    # the libraries that we're going to build. However, they may be compiled
+    # with a MACOSX_DEPLOYMENT_TARGET that doesn't match what we want to use,
+    # and they may bring in other dependencies that we don't want. The same will
+    # be true of any other locations on the path. To avoid conflicts, strip the
+    # path down to the bare mimimum (which, on macOS, won't include any
```
radarhere, line 174:
```suggestion
    # path down to the bare minimum (which, on macOS, won't include any
```

setup.py
```diff
@@ -448,7 +454,7 @@ def _remove_extension(self, name: str) -> None:
     def get_macos_sdk_path(self) -> str | None:
         try:
             sdk_path = (
-                subprocess.check_output(["xcrun", "--show-sdk-path"])
+                subprocess.check_output(["xcrun", "--show-sdk-path", "--sdk", "macosx"])
```
radarhere, line 457: What was the reason for this change?

.github/workflows/wheels-test.sh
```diff
@@ -1,12 +1,27 @@
...
+        HOMEBREW_HOME=/opt/homebrew
     fi
+    $HOMEBREW_HOME/bin/brew install fribidi
+
+    # Add the lib folder for fribidi so that the vendored library can be found.
+    # Don't use /opt/homebrew/lib directly - use the lib folder where the
+    # installed copy of fribidi is cellared. This ensures we don't pick up the
+    # Homebrew version of any other library that we're dependent on (most notably,
+    # freetype).
+    export DYLD_LIBRARY_PATH=$(dirname $(realpath $HOMEBREW_HOME/lib/libfribidi.dylib))
```
radarhere, line 24:
```suggestion
    if [[ "$(uname -m)" == "x86_64" ]]; then
        FRIBIDI_DYLIB=$(find /usr/local/Cellar/fribidi -type f -name libfribidi.dylib)
    else
        FRIBIDI_DYLIB=$HOMEBREW_HOME/lib/libfribidi.dylib
    fi
    export DYLD_LIBRARY_PATH=$(dirname $FRIBIDI_DYLIB)
```

https://github.com/python-pillow/Pillow/actions/runs/11752543578/job/32744256384?pr=8497#step:5:9142
> realpath: /usr/local/Homebrew/lib/libfribidi.dylib: No such file or directory

libfribidi.dylib isn't located there on the macOS Intel jobs.

setup.py
freakboy3742: Explicitness. Xcode can manage multiple SDKs (iOS being an obvious one); adding the explicit `--sdk macosx` SDK ensures that you're *definitely* getting the macOS one. That will be the default on *almost* every install of Xcode, but if you have a stray `SDKROOT` environment variable from some other activity, it could be inadvertently pointing at an iOS, tvOS, visionOS or MacCatalyst SDK.

.github/workflows/wheels-test.sh
freakboy3742: Huh... at some point, I've clearly done something to mess up my x86 test machine's Homebrew install, because I've got a `/usr/local/Homebrew/Cellar`... but I now see I'm also getting a bunch of warnings because of the existence of that folder. 

The `find`-based solution you've got here will fail (or be prone to failure) if there's more than one version of fribidi installed, which is possible on user machines; but if `HOMEBREW_HOME` is moved back to `/usr/local`, the `realpath` approach should work.

.github/workflows/wheels-dependencies.sh
```diff
@@ -38,35 +60,44 @@ BZIP2_VERSION=1.0.8
 LIBXCB_VERSION=1.17.0
 BROTLI_VERSION=1.1.0
 
+function build_pkg_config {
+    if [ -e pkg-config-stamp ]; then return; fi
+    # This essentially duplicates the Homebrew recipe:
+    # https://github.com/Homebrew/homebrew-core/blob/master/Formula/p/pkg-config.rb
```
radarhere, line 66:
```suggestion
    # This essentially duplicates the Homebrew recipe
```
https://github.com/Homebrew/homebrew-core/blob/master/Formula/p/pkg-config.rb no longer exists after https://github.com/Homebrew/homebrew-core/pull/194885

.github/workflows/wheels-dependencies.sh
```diff
@@ -77,67 +108,88 @@ function build {
...
+
     build_libpng
     build_lcms2
     build_openjpeg
 
     ORIGINAL_CFLAGS=$CFLAGS
     CFLAGS="$CFLAGS -O3 -DNDEBUG"
-    if [[ -n "$IS_MACOS" ]]; then
-        CFLAGS="$CFLAGS -Wl,-headerpad_max_install_names"
-    fi
```
radarhere, line 99: What was your thinking behind removing this? I'm just concerned that https://github.com/python-pillow/Pillow/issues/7250 might re-appear.

.github/workflows/wheels-dependencies.sh
```diff
@@ -77,67 +108,88 @@ function build {
...
+    # the libraries that we're going to build. However, they may be compiled
+    # with a MACOSX_DEPLOYMENT_TARGET that doesn't match what we want to use,
+    # and they may bring in other dependencies that we don't want. The same will
+    # be true of any other locations on the path. To avoid conflicts, strip the
+    # path down to the bare minimum (which, on macOS, won't include any
+    # development dependencies).
+    export PATH="$BUILD_PREFIX/bin:$(dirname $(which python3)):/usr/bin:/bin:/usr/sbin:/sbin:/Library/Apple/usr/bin"
+    export CMAKE_PREFIX_PATH=$BUILD_PREFIX
+
+    # Link the brew command into our isolated build directory.
```
radarhere, line 178: I'm not following this comment. Are you talking about a specific brew command that we're invoking, or just if a user wants to add their own call to brew?

.github/workflows/wheels-dependencies.sh
```diff
@@ -77,67 +108,88 @@ function build {
...
-    fi
     build_libwebp
     CFLAGS=$ORIGINAL_CFLAGS
 
     build_brotli
 
     if [ -n "$IS_MACOS" ]; then
         # Custom freetype build
-        build_simple freetype $FREETYPE_VERSION https://download.savannah.gnu.org/releases/freetype tar.gz --with-harfbuzz=no
+        build_simple freetype $FREETYPE_VERSION https://download.savannah.gnu.org/releases/freetype tar.gz --without-harfbuzz
```
radarhere, line 144: https://gitlab.freedesktop.org/freetype/freetype/-/blob/master/builds/unix/configure.raw?ref_type=heads#L429 mentions `--with-harfbuzz`, but not `--without-harfbuzz`. What was the thinking here?

.github/workflows/wheels-dependencies.sh
```diff
@@ -77,67 +108,88 @@ function build {
     if [ -n "$IS_MACOS" ]; then
         build_simple xorgproto 2024.1 https://www.x.org/pub/individual/proto
         build_simple libXau 1.0.11 https://www.x.org/pub/individual/lib
+        build_simple libXdmcp 1.1.5 https://www.x.org/pub/individual/lib
```
radarhere, line 111: > Adds a build of `libXdmcp`, to avoid a dependency on the Homebrew-provided version.

I suspect you took out all the `brew remove` statements, then noticed this was being pulled from brew, and then added it here to replace that.

I'm saying that this wouldn't be part of the Pillow wheels at the moment, so if it's still not after this PR, that sounds fine to me.

This function
https://github.com/python-pillow/Pillow/blob/5bff2f3b2894ec6923c590d0c37b18177d0634bd/src/display.c#L834 is the only use of XCB in Pillow, so if it's not necessary, I don't think it needs to be here?

.github/workflows/wheels-dependencies.sh
```diff
@@ -77,67 +108,88 @@ function build {
...
     build_libjpeg_turbo
-    build_tiff
+    if [ -n "$IS_MACOS" ]; then
+        # Custom tiff build to include jpeg; by default, configure won't include
+        # headers/libs in the custom macOS prefix. Explicitly disable webp,
+        # libdeflate and zstd, because on x86_64 macs, it will pick up the
+        # Homebrew versions of those libraries from /usr/local.
+        build_simple tiff $TIFF_VERSION https://download.osgeo.org/libtiff tar.gz \
+            --with-jpeg-include-dir=$BUILD_PREFIX/include --with-jpeg-lib-dir=$BUILD_PREFIX/lib \
+            --disable-webp --disable-zstd --disable-libdeflate
```
radarhere, line 126:
```suggestion
            --disable-webp --disable-libdeflate --disable-zstd
```
Nitpick - use the arguments in the same order as they are mentioned in the comment.

radarhere:
Thanks for your patience. This is getting close - aside from my latest comments, I'm happy with the current state.

.github/workflows/wheels-dependencies.sh
freakboy3742: "--with/--without" handing is a standard feature of autoconf, AFAIK; I think this change got made when I was trying to diagnose some harfbuzz issues. Happy to revert, as it's code churn with no explicit purpose.

.github/workflows/wheels-dependencies.sh
freakboy3742: A stale comment. In an earlier iteration, it was explicitly linking the brew executable into `$BUILD_PREFIX/bin`; that's no longer happening. I'll update the comment.

.github/workflows/wheels-dependencies.sh
freakboy3742: This got removed when I was getting webp to build, because it didn't seem to causing any build issues. I see now it's related to py2app packaging, so wouldn't necessarily manifest in a CI failure. I'll restore it.

.github/workflows/wheels-dependencies.sh
freakboy3742: Link rot before the PR even lands has to be a new record... :-)

I'll remove the reference.

.github/workflows/wheels-dependencies.sh
freakboy3742: My recollection was that Xdmcp was needed for XCB to compile *at all*; but I've just removed it and done a clean build, and everything worked as expected (AFAICT). I'll remove it.

freakboy3742:
@radarhere No problem at all - thanks for all the reviews. I've just pushed an update that I think addresses all your comments.

radarhere (approved):

aclark4life:
Thanks @freakboy3742 @radarhere @nulano !

hugovk:
Thank you!

PR #8498, opened 2024-10-25
Added filename placeholder in winbuild URL

radarhere:
There are several places in `winbuild/build_prepare.py` where the "filename" is replicated in the "url"
https://github.com/python-pillow/Pillow/blob/822ec3d05dfbaf2e3209c2b26e3392040e274975/winbuild/build_prepare.py#L163-L165

I wonder if
```python
"zlib": {
    "url": "https://zlib.net/FILENAME",
    "filename": f"zlib{V['ZLIB_DOTLESS']}.zip",
```
is neater, reducing duplication?

Feel free to reject if others think it isn't simpler.

### DIFF
--- a/winbuild/build_prepare.py
+++ b/winbuild/build_prepare.py
@@ -130,8 +130,7 @@ V["ZLIB_DOTLESS"] = V["ZLIB"].replace(".", "")
 # dependencies, listed in order of compilation
 DEPS: dict[str, dict[str, Any]] = {
     "libjpeg": {
-        "url": f"{SF_PROJECTS}/libjpeg-turbo/files/{V['JPEGTURBO']}/"
-        "FILENAME/download",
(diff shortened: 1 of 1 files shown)


winbuild/build_prepare.py
```diff
@@ -131,7 +131,7 @@ def cmd_msbuild(
 DEPS: dict[str, dict[str, Any]] = {
     "libjpeg": {
         "url": f"{SF_PROJECTS}/libjpeg-turbo/files/{V['JPEGTURBO']}/"
-        f"libjpeg-turbo-{V['JPEGTURBO']}.tar.gz/download",
+        "FILENAME/download",
```
nulano, line 134: This string probably fits on a single line now.

nulano (approved):

winbuild/build_prepare.py
radarhere: You're right. I've pushed a commit.

Issue #8492, opened 2024-10-23
Running the tests in multiple threads fails because of thread-unsafe Python code


lysnikolaou:
Apart from the failure in #8454 and other hard crashes at the C level (I'll open separate issues for those), quite a few tests fail when run in multiple threads because of thread-unsafe Python code.

The failures come mainly from three different thread safety issues. Two of those issues come from issues in the tests themselves:
1. Using `pytest.warns` which is inherently thread-unsafe. The solution for this is to always run those tests in a signle thread.
2. Saving to the same file from multiple threads. The filesystem thus becomes global state, which introduces data races. The solution to this is to make sure different threads write to different files when testing things (for example by using `uuid.uuid4()` and prepending that to the filename `Image.save` is called with).

The third, however, is because of thread-unsafe Python code in Pillow. More specifically, various registries (like [these ones](https://github.com/python-pillow/Pillow/blob/822ec3d05dfbaf2e3209c2b26e3392040e274975/src/PIL/Image.py#L229)) are kept as global variables. Changing these in multiple threads leads to data races. Fixing these is going to be harder, but the possible solutions are:

1. Make all of theses `ContextVar`s or `threading.local`s. This is a backwards-incompatible change though and users that touch these in their code will have to update it to use the context variable APIs (`.get()` & `.set()`).
2. Leave them alone and document that registering plugins in multiple threads is not supported (for now?).

Does anyone else have ideas regarding other ways to handle this?

An easy way to reproduce this (though this fails because of [this global variable](https://github.com/python-pillow/Pillow/blob/822ec3d05dfbaf2e3209c2b26e3392040e274975/src/PIL/BufrStubImagePlugin.py#L17)) is to install `pytest-run-parallel` and then run the following under the free-threaded build:

```
pytest -v --parallel-threads=100 ./Tests/test_file_bufrstub.py
```

<details>
<summary>Failure info</summary>

```pytest
rootdir: /Users/lysnikolaou/repos/python/Pillow
configfile: pyproject.toml
plugins: run-parallel-0.1.0
collected 5 items                                                                                                                                                           

Tests/test_file_bufrstub.py::test_open PASSED                                                                                                                         [ 20%]
Tests/test_file_bufrstub.py::test_invalid_file PASSED                                                                                                                 [ 40%]
Tests/test_file_bufrstub.py::test_load PASSED                                                                                                                         [ 60%]
Tests/test_file_bufrstub.py::test_save PASSED                                                                                                                         [ 80%]
[...]

PR #8501, opened 2024-10-26
Only test `register_handler` in a single thread

radarhere:
Possible way to help #8492. Alternative to #8518

The issue has found that running a test that calls `register_handler` at the beginning, and removes the handler at the end, fails when the test is run across multiple threads.

My thinking is that if it ok for `register_handler` to operate across threads, and that it doesn't need to register multiple handlers in parallel. So this PR returns early from the relevant test if it is already running in another thread.


lysnikolaou (commented):
I've also been working on fixing some of these issues. Since you think we should leave these alone in https://github.com/python-pillow/Pillow/issues/8492#issuecomment-2437863662, I think we should mark the tests with `@pytest.mark.parallel_threads(1)` instead.

radarhere:
Is it a problem that `pytest.mark.parallel_threads` would only work for `pytest-run-parallel`? We've also had https://github.com/python-pillow/Pillow/issues/8454 from a person using https://github.com/tonybaloney/pytest-freethreaded. If that happens again, do we need to ask such a user to change testing tools?

https://github.com/python-pillow/Pillow/pull/8518 is another idea I had.

lysnikolaou:
> Is it a problem that `pytest.mark.parallel_threads` would only work for `pytest-run-parallel`? We've also had #8454 from a person using [tonybaloney/pytest-freethreaded](https://github.com/tonybaloney/pytest-freethreaded). If that happens again, do we need to ask such a user to change testing tools?

I'm not sure about that to be honest. I'm using `pytest-run-parallel` cause I like the fact that the default is a single thread. The best solution here would be to have a `@pytest.mark.thread_unsafe` mark that would work across plugins. I can coordinate doing that.

PR #8506, opened 2024-10-26
Skip failing records when rendering WMF images

### LINKED ISSUE #6980: EMF files: OSError: cannot render metafile
### What did you do?
* Download example EMF file `test_libuemf_ref.emf` from https://bugs.documentfoundation.org/attachment.cgi?id=135791 (via https://bugs.documentfoundation.org/show_bug.cgi?id=55058)

* Run:

```python
import PIL.Image
with PIL.Image.open("test_libuemf_ref.emf") as im:
    im.save("out.png")
```

`out.png` should be created.

### What actually happened?
```pytb
C:\Python311\Lib\site-packages\PIL\Image.py:3167: DecompressionBombWarning: Image size (139177600 pixels) exceeds limit of 89478485 pixels, could be decompression bomb DOS attack.
  warnings.warn(
{'wmf_bbox': (0, 0, 14030, 9920), 'dpi': (1199.9124549648136, 1199.9047573693986)}
  File "test.py", line 3, in <module>
    im.save("test_libuemf_ref.emf")
  File "C:\Python311\Lib\site-packages\PIL\Image.py", line 2394, in save
    self._ensure_mutable()
  File "C:\Python311\Lib\site-packages\PIL\Image.py", line 611, in _ensure_mutable
    self._copy()
  File "C:\Python311\Lib\site-packages\PIL\Image.py", line 604, in _copy
    self.load()
  File "C:\Python311\Lib\site-packages\PIL\WmfImagePlugin.py", line 162, in load
    return super().load()
           ^^^^^^^^^^^^^^
  File "C:\Python311\Lib\site-packages\PIL\ImageFile.py", line 345, in load
    image = loader.load(self)
            ^^^^^^^^^^^^^^^^^
  File "C:\Python311\Lib\site-packages\PIL\WmfImagePlugin.py", line 53, in load
    Image.core.drawwmf(im.fp.read(), im.size, self.bbox),
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
OSError: cannot render metafile
```

* OS: Windows 10
* Python: 3.11
* Pillow: 9.4.0

radarhere:
Resolves #6980

The issue would like Pillow to be more lenient when rendering invalid EMF files. The API that raises the error is relatively simple

https://github.com/python-pillow/Pillow/blob/81a3bf5f58d760128ca8beb3f10c1a2c27ec6df6/src/display.c#L799-L800

so I asked about it at https://learn.microsoft.com/en-us/answers/questions/2100252/playenhmetafile-fails-without-any-getlasterror-res, and the answer showed that `PlayEnhMetaFileRecord` can be used to render the image gradually, skipping failing records.


petsuter:
Great! As mentioned in the ticket, the C# .NET implementation also uses such record-by-record API's internally and allows the files to be handled, so I wouldn't hesitate to do the same. Also programs like Windows Paint and IrfanView allow the files to be opened without error.👍

radarhere:
It turns out my initial commit didn't work on 32-bit Windows - https://ci.appveyor.com/project/Python-pillow/pillow/builds/50869937/job/3jd42o23cjses7ms#L3048
> src/display.c(807): error C2440: 'function': cannot convert from 'BOOL (__cdecl *)(HDC,HANDLETABLE *,const ENHMETARECORD *,int,LPARAM)' to 'ENHMFENUMPROC'

I've pushed a commit to only use the new behaviour on 64-bit. @petsuter is that sufficient to resolve your issue? Or would you like to see this work on 32-bit as well?

petsuter:
I don't know of any 32-bit Python usage anymore, thanks.

(Could it be the return value should be int not BOOL? https://learn.microsoft.com/en-us/windows/win32/api/wingdi/nc-wingdi-enhmfenumproc)

radarhere:
Thanks.

No, [changing it to `int`](https://github.com/radarhere/Pillow/commit/f233057216e84a0b3a139ea92422c704b63930b4) still gives https://ci.appveyor.com/project/radarhere/pillow/builds/50870384/job/ji4kbtk94gl000q3#L3047
>   src/display.c(807): error C2440: 'function': cannot convert from 'int (__cdecl *)(HDC,HANDLETABLE *,const ENHMETARECORD *,int,LPARAM)' to 'ENHMFENUMPROC'

src/display.c
```diff
@@ -716,6 +716,14 @@ PyImaging_EventLoopWin32(PyObject *self, PyObject *args) {
 
 #define GET32(p, o) ((DWORD *)(p + o))[0]
 
+BOOL
+enhMetaFileProc(
+    HDC hdc, HANDLETABLE FAR *lpht, const ENHMETARECORD *lpmr, int nHandles, LPARAM data
```
Yay295, line 721: why a FAR pointer?

radarhere: Good question. Ok, I've updated the code to match https://learn.microsoft.com/en-us/windows/win32/api/wingdi/nc-wingdi-enhmfenumproc

src/display.c
```diff
@@ -716,6 +716,14 @@ PyImaging_EventLoopWin32(PyObject *self, PyObject *args) {
 
 #define GET32(p, o) ((DWORD *)(p + o))[0]
 
+int
```
nulano, line 719:
```suggestion
int CALLBACK
```
You need to specify the `__stdcall` call convention for 32-bit builds as the default is `__cdecl`.
It worked on 64-bit because there is only a single call convention used there.

I'm not sure why this is missing from the documentation, I had to look at `wingdi.h` to figure this out.

radarhere: Thanks for the thoughts, but I can't find a way to put them together into something that passes on 32-bit. Would you be able to create a PR?

radarhere: https://github.com/radarhere/Pillow/pull/28 fixed this.

Issue #6547, opened 2022-08-29
New struct for image mode data


Yay295:
Currently all of the data about the format/mode of an image is stored in the image itself.

https://github.com/python-pillow/Pillow/blob/ad7be550aa539dc4c79da928dad071cf0150b731/src/libImaging/Imaging.h#L80-L107

I think it would be cleaner if this information was stored in a separate struct.

```c
struct ImageMode {
    char mode[IMAGING_MODE_LENGTH]; /* Band names ("1", "L", "P", "RGB", "RGBA", "CMYK", "YCbCr", "BGR;xy") */
    int type;                       /* Data type (IMAGING_TYPE_*) */
    int depth;                      /* Depth (ignored in this version) */
    int bands;                      /* Number of bands (1, 2, 3, or 4) */
    int pixelsize;                  /* Size of a pixel, in bytes (1, 2 or 4) */
};
```

Furthermore, I think some more changes could be made that would simplify coding for some calculations, and allow for more image formats to be processed (#1888).

```c
struct BandDataType {
    char dataType; /* 'u' = unsigned integer, 's' = signed integer, 'f' = floating point */
    UINT8 numBytes;
} BAND_DATA_TYPES[] = {
    {'u', 1}, /* UINT8 */
    {'u', 2}, /* UINT16 */
    {'s', 4}, /* INT32 */
    {'f', 4}, /* FLOAT32 */
};
/* Normal bands are treated normally */
/* Alpha bands may be treated specially by some calculations */
/* Extra bands are ignored (this could be used for padding, or maybe storing non-image data) */
enum BandType { NORMAL, ALPHA, EXTRA };
struct Band {
    char *name; /* "R", "G", "B", "Y", "Cb", "Cr", etc. */
    BandDataType dataType;
    BandType type;
};
struct ImageFormat {
    char *name;      /* Format name ("1", "L", "P", "RGB", "RGBA", "CMYK", "YCbCr", "BGR;xy") */
    Band *bands;     /* Array of bands in this format */
    UINT32 numBands; /* Number of bands in this format */
    UINT32 numBytes; /* Total size in bytes of one pixel in this format */
};
```

radarhere:
@wiredfool did you have any thoughts on this?

wiredfool:
I don't have strong opinions on it -- perhaps from not seeing how this reorg would simplify existing code. 

[...]

PR #8510, opened 2024-10-28
Use enums for Modes and RawModes in C

Yay295:
Adds some type safety when comparing modes, and replaces a lot of string comparisons with enum comparisons.


src/libImaging/Mode.h
```diff
@@ -0,0 +1,240 @@
...
+    IMAGING_MODE_I_16L,
+    IMAGING_MODE_I_16B,
+    IMAGING_MODE_I_16N,
+    IMAGING_MODE_I_32L,
+    IMAGING_MODE_I_32B,
+} ModeID;
+
+typedef struct {
+    const char *const name;
+} ModeData;
```
radarhere, line 38: You've created a struct with only one member, presumably because you expect to add more members in the future. Are you trying to prepare for #6547?

Yay295: There are some things I think could be added to this struct in the future, but the more important reason for this as it is right now is type safety. A `char *` could be almost anything, but a `ModeData` can only be a `ModeData`.

radarhere: Correct me if I'm wrong, but the only place where `ModeData` is returned is from Mode.c's `getModeData()` (and similarly, `RawModeData` from `getRawModeData()`)? It's used there, and in the construction of the `MODES` array in Mode.c, and by design, never anywhere else? I don't think we even pass between `ModeData` between functions - every use outside of Mode.c is just to access the `name`.

So the only benefit is to assure us that `getModeData()` and `getRawModeData()` return the correct output? This could just be my opinion, but I think they are simple functions, and it's overkill to be passing around a struct to the rest of the C code to ensure their correctness.

wiredfool: I think this will be useful for the band names and arrow band format constants as well. I could also see this as where the bands/depth/type are stored.

radarhere:
This has been rebased as #9100

PR #8512, opened 2024-10-29
Do not repeatedly save to the same path

radarhere:
https://github.com/tonybaloney/pytest-freethreaded/issues/11 noted a segfault that can occur under free-threading with `Tests/test_file_jpeg.py::test_qtables`. I can replicate this with [`--require-gil-disabled`](https://github.com/radarhere/Pillow/commit/d928ddf06c75ca8e730d52df575314fc4788e825) - https://github.com/radarhere/Pillow/actions/runs/11572743510

Changing the test to save to a BytesIO instance rather than writing repeatedly to the same temporary path fixes this specific test - https://github.com/radarhere/Pillow/actions/runs/11572819452

### DIFF
--- a/Tests/test_file_jpeg.py
+++ b/Tests/test_file_jpeg.py
@@ -541,7 +541,7 @@ class TestFileJpeg:
     @mark_if_feature_version(
         pytest.mark.valgrind_known_error, "libjpeg_turbo", "2.0", reason="Known Failing"
     )
-    def test_qtables(self, tmp_path: Path) -> None:
+    def test_qtables(self) -> None:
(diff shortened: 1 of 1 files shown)


Tests/test_file_jpeg.py
```diff
@@ -543,10 +543,10 @@ def test_truncated_jpeg_throws_oserror(self) -> None:
     )
     def test_qtables(self, tmp_path: Path) -> None:
```
hugovk, line 544: Can we remove the fixture?
```suggestion
    def test_qtables(self) -> None:
```

PR #8513, opened 2024-10-29
Require trove-classifiers>=2024.10.12

radarhere:
main has started failing - https://github.com/python-pillow/Pillow/actions/runs/11573601676/job/32216060741#step:28:4035
```
   +         'Some of your classifiers are not standard classifiers:\n'
   +         'License :: OSI Approved :: CMU License (MIT-CMU)',
```

PR #8514, opened 2024-10-30
Add trove-classifiers>=2024.10.12 to 'tests' extra and use for Windows CI

hugovk:
Re: https://github.com/python-pillow/Pillow/pull/8513#discussion_r1820775250

`trove-classifiers>=2024.10.12` is required for the `pyroma` tests.

Removes duplication of test dependencies.


hugovk:
Strange, pypy3 couldn't find `trove-classifiers>=2024.10.12`, and `2024.7.2` is the latest it can find:

```pytb
INFO: pip is looking at multiple versions of pillow[tests] to determine which version is compatible with other requirements. This could take a while.
ERROR: Could not find a version that satisfies the requirement trove-classifiers>=2024.10.12; extra == "tests" (from pillow[tests]) (from versions: 2020.4.1, 2020.4.9, 2020.5.11, 2020.5.21, 2020.6.20, 2020.7.30, 2020.8.3, 2020.8.6, 2020.8.7, 2020.8.11, 2020.8.14, 2020.8.17, 2020.9.3, 2020.9.8, 2020.9.25, 2020.10.6, 2020.10.7, 2020.10.21, 2021.1.14, 2021.2.28, 2021.4.5, 2021.7.10, 2021.8.4, 2021.8.31, 2021.9.17, 2021.9.22, 2021.9.23, 2021.9.27, 2021.10.9, 2021.10.20, 2021.11.16, 2021.11.17, 2022.1.3, 2022.1.6, 2022.2.16, 2022.3.10, 2022.3.16, 2022.3.17, 2022.3.21, 2022.3.30, 2022.4.11, 2022.4.30, 2022.5.5, 2022.5.11, 2022.5.12, 2022.5.19, 2022.6.23, 2022.6.26, 2022.7.22, 2022.8.7, 2022.8.24, 2022.8.31, 2022.9.26, 2022.10.19, 2022.12.1, 2022.12.22, 2023.1.12, 2023.1.20, 2023.2.8, 2023.2.20, 2023.3.9, 2023.4.18, 2023.4.20, 2023.4.22, 2023.4.25, 2023.4.29, 2023.5.2, 2023.5.22, 2023.5.24, 2023.7.6, 2023.8.7, 2023.9.19, 2023.10.17, 2023.10.18, 2023.11.7, 2023.11.9, 2023.11.13, 2023.11.14, 2023.11.22, 2023.11.29, 2024.1.8, 2024.1.31, 2024.2.22, 2024.2.23, 2024.3.3, 2024.3.25, 2024.4.10, 2024.5.17, 2024.5.22, 2024.7.1, 2024.7.2)

Notice:  A new release of pip is available: 24.2 -> 24.3.1
Notice:  To update, run: python.exe -m pip install --upgrade pip
ERROR: No matching distribution found for trove-classifiers>=2024.10.12; extra == "tests"
  File "D:\a\Pillow\Pillow\selftest.py", line 7, in <module>
    from PIL import Image, features
ModuleNotFoundError: No module named 'PIL'
```

But https://pypi.org/project/trove-classifiers/2024.10.12/ and newer are on PyPI.

It installed it and passed on my fork:

https://github.com/hugovk/Pillow/actions/runs/11588271332/job/32277504083#step:25:132

.github/workflows/test-windows.yml
```diff
@@ -69,22 +69,11 @@ jobs:
...
-        pytest-cov
-        pytest-timeout
-
     - name: Install CPython dependencies
       if: "!contains(matrix.python-version, 'pypy')"
-      run: >
-        python3 -m pip install
-        PyQt6
+      run: |
+        python3 -m pip install --upgrade pip
```
nulano, line 75: This is not run on PyPy, only CPython. Maybe that is causing the issue?

hugovk: Let's see! It makes sense to upgrade for both anyway.

hugovk: Well, passing now, thanks :)

.github/workflows/test-windows.yml
```diff
@@ -69,22 +69,14 @@ jobs:
     - name: Print build system information
       run: python3 .github/workflows/system-info.py
 
-    - name: Install Python dependencies
-      run: >
-        python3 -m pip install
-        coverage>=7.4.2
```
radarhere, line 75: Just to call it out, there is a difference between this and https://github.com/python-pillow/Pillow/blob/82199efbf77f369ad9b7c2c3759e08f517d9925b/pyproject.toml#L57-L59

radarhere (approved):
I'm not convinced this isn't just passing now because of the cleared caches, but this doesn't hurt

.github/workflows/test-windows.yml
hugovk: Yep. 

* `coverage` vs `coverage>=7.4.2`: before, theoretically we could have installed coverage older than 7.4.2 here. I think that was unlikely and it's safer to require the newer version.

* `check-manifest`: we don't need that here. I see little harm in installing it here. It's in the `tests` extra because it's used by `make release-test`. If we do want to keep this env "cleaner", we could remove it from the `tests` extra and put it in something like a `release-test` extra.

Relatedly, [PEP 735](https://peps.python.org/pep-0735/) was accepted recently to define something called "dependency groups". These are like extras, but you can install a dependency group without having to also install Pillow as well.

tox and some other tools already support it, but not yet pip. We could look into using it when pip is ready.

hugovk:
> I'm not convinced this isn't just passing now because of the cleared caches

Yeah, I think so, here's a similar report: https://github.com/actions/setup-python/issues/972

The pyroma test is skipped when trove-classifiers isn't installed.

But if we want to run it, `trove-classifiers>=2024.10.12` is a required dependency. We shouldn't really just hope we end up with a newer one installed into the system.

PR #8516, opened 2024-10-31
Allow HWND to be passed to ImageGrab.grab() on Windows

### LINKED ISSUE #4415: Windows application screenshot
windows 10 supports built-in virtual desktop. But when I change desktop from A to B while Pillow process takes screenshot in A, then I get B's screenshot. 

The thing i want to know is, was this intended? Cause i don't know well about pillow's goal.
If this is not intended, i'm planning to improve window screen grabber

https://github.com/nulano/Pillow/blob/3c311f5619b1b82e2b91cd2a42e3a44b9526755c/src/display.c#L323

### LINKED ISSUE #8456: ImageGrab.grab() doesn't ignore windows layered on top (tooltip, pop-ups, etc.)
### What did you do?

Run PIL.ImageGrab.grab() without arguments.

Until now this has not included pop-ups and tooltips in the screenshot.
I just updated to windows 11. 
I can see in issue #2569 that it is the expected behavior to not include layered windows, and that the option to include them was added, but not as a default. 

I suppose the underlying API has changed in windows 11.
Anyone that can share some knowledge about this? 

* OS: Windows 11
* Python: 3.10.5
* Pillow: 10.4.0

```text
Pillow 10.4.0
Python 3.10.5 (tags/v3.10.5:f377153, Jun  6 2022, 16:14:13) [MSC v.1929 64 bit (AMD64)]
Python executable is C:\Users\Public\SPEDA\env\Scripts\python.exe
Environment Python files loaded from C:\Users\Public\SPEDA\env
System Python files loaded from C:\Program Files\Python310
Python Pillow modules loaded from C:\Users\Public\SPEDA\env\lib\site-packages\PIL
Binary Pillow modules loaded from C:\Users\Public\SPEDA\env\lib\site-packages\PIL
--- PIL CORE support ok, compiled for 10.4.0
--- WEBPMUX support ok
--- WEBP Animation support ok
```

radarhere:
Resolves #4415

Adds a `handle` argument to `ImageGrab.grab()` that accepts a HDC, allowing for a screenshot of a specific window, rather than the entire screen.


src/display.c
```diff
@@ -331,14 +331,18 @@ PyImaging_GrabScreenWin32(PyObject *self, PyObject *args) {
     HMODULE user32;
     Func_SetThreadDpiAwarenessContext SetThreadDpiAwarenessContext_function;
 
-    if (!PyArg_ParseTuple(args, "|ii", &includeLayeredWindows, &all_screens)) {
+    if (!PyArg_ParseTuple(
+            args, "|ii" F_HANDLE, &includeLayeredWindows, &screens, &screen
```
nulano, line 335: This seems to be stealing a reference to the HDC and then deleting it, which could violate the `DeleteDC` documentation, https://learn.microsoft.com/en-us/windows/win32/api/wingdi/nf-wingdi-deletedc#remarks:

> An application must not delete a DC whose handle was obtained by calling the [GetDC](https://learn.microsoft.com/en-us/windows/desktop/api/winuser/nf-winuser-getdc) function. Instead, it must call the [ReleaseDC](https://learn.microsoft.com/en-us/windows/desktop/api/winuser/nf-winuser-releasedc) function to free the DC.

It would also be more intuitive to accept a HWND argument and get the HDC with the `GetDC` function or the [`GetWindowDC` function](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getwindowdc) (making sure to release it with `ReleaseDC` as stated in the documentation).

nulano (changes requested):

src/display.c
radarhere: Ok, I've updated the commit to receive a HWND and use `ReleaseDC`.

src/PIL/ImageGrab.py
```diff
@@ -22,15 +22,20 @@
...
+if TYPE_CHECKING:
+    from . import ImageWin
+
 
 def grab(
     bbox: tuple[int, int, int, int] | None = None,
     include_layered_windows: bool = False,
     all_screens: bool = False,
     xdisplay: str | None = None,
+    handle: int | ImageWin.HWND | None = None,
```
nulano, line 38:
```suggestion
    window: int | ImageWin.HWND | None = None,
```
or
```suggestion
    window_handle: int | ImageWin.HWND | None = None,
```
seems clearer.

src/display.c
```diff
@@ -420,7 +441,11 @@ PyImaging_GrabScreenWin32(PyObject *self, PyObject *args) {
     PyErr_SetString(PyExc_OSError, "screen grab failed");
 
     DeleteDC(screen_copy);
-    DeleteDC(screen);
+    if (screens == -1) {
+        ReleaseDC(wnd, screen);
```
nulano, line 445: You added this for the `error:` branch but not for the success branch.

src/display.c
```diff
@@ -351,11 +362,17 @@ PyImaging_GrabScreenWin32(PyObject *self, PyObject *args) {
         dpiAwareness = SetThreadDpiAwarenessContext_function((HANDLE)-3);
```
nulano, line 362: We need to match the capturing thread DPI awareness to the target window DPI awareness which can be queried with GetWindowDpiAwarenessContext (also added in Windows 10 version 1607): https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getwindowdpiawarenesscontext

Without it I get a black border around applications without native DPI scaling support such as Microsoft Spy++.

Tests/test_imagegrab.py
```diff
@@ -57,6 +57,11 @@ def test_grab_invalid_xdisplay(self) -> None:
             ImageGrab.grab(xdisplay="error.test:0.0")
         assert str(e.value).startswith("X connection failed")
 
+    @pytest.mark.skipif(sys.platform != "win32", reason="Windows only")
+    def test_grab_invalid_handle(self) -> None:
+        with pytest.raises(OSError):
+            ImageGrab.grab(handle=-1)
```
nulano, line 63:
```suggestion
        with pytest.raises(OSError):
            ImageGrab.grab(handle=-1)
        with pytest.raises(OSError):
            ImageGrab.grab(handle=0)
```
The value 0 is a special way to refer to the desktop. However, it cannot be captured in the same way as regular windows so it fails, but we can add a test for that.

nulano (commented):
FWIW not all windows can be captured like this. Notably, Vivaldi (my browser) only shows as a black box of the correct size, and the Windows file explorer is missing its toolbar. Other applications, e.g. PyCharm or Microsoft Spy++, are captured correctly.

src/PIL/ImageGrab.py
radarhere: Ok, I've chosen `window`.

src/display.c
radarhere: Ok, I've updated the commit to add it or success as well.

Tests/test_imagegrab.py
radarhere: Thanks. I've added a commit.

src/display.c
radarhere: Ok, I've pushed a commit.

src/display.c
```diff
@@ -347,15 +360,32 @@ PyImaging_GrabScreenWin32(PyObject *self, PyObject *args) {
...
-        // DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE = ((DPI_CONTEXT_HANDLE)-3)
-        dpiAwareness = SetThreadDpiAwarenessContext_function((HANDLE)-3);
+        if (screens == -1) {
+            GetWindowDpiAwarenessContext_function = (Func_GetWindowDpiAwarenessContext
+            )GetProcAddress(user32, "GetWindowDpiAwarenessContext");
+            DPI_AWARENESS_CONTEXT dpiAwarenessContext =
+                GetWindowDpiAwarenessContext_function(wnd);
+            if (dpiAwarenessContext != NULL) {
+                dpiAwareness =
+                    SetThreadDpiAwarenessContext_function(dpiAwarenessContext);
```
nulano, line 370: It should probably never happen that `SetThreadDpiAwarenessContext` is found but `GetWindowDpiAwarenessContext` is not.

However, if it somehow does happen, we would then use `dpiAwareness` before its initialized after measuring the window size. So we should ideally either fallback to PER_MONITOR_AWARE context if `GetWindowDpiAwarenessContext` is not found, or skip the reset a few lines later.

src/display.c
```diff
@@ -347,15 +360,34 @@ PyImaging_GrabScreenWin32(PyObject *self, PyObject *args) {
...
     )GetProcAddress(user32, "SetThreadDpiAwarenessContext");
     if (SetThreadDpiAwarenessContext_function != NULL) {
-        // DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE = ((DPI_CONTEXT_HANDLE)-3)
-        dpiAwareness = SetThreadDpiAwarenessContext_function((HANDLE)-3);
+        if (screens == -1) {
+            GetWindowDpiAwarenessContext_function = (Func_GetWindowDpiAwarenessContext
+            )GetProcAddress(user32, "GetWindowDpiAwarenessContext");
+            DPI_AWARENESS_CONTEXT dpiAwarenessContext =
+                GetWindowDpiAwarenessContext_function(wnd);
+            if (dpiAwarenessContext != NULL) {
```
nulano, line 368:
```suggestion
            if (GetWindowDpiAwarenessContext_function != NULL && dpiAwarenessContext != NULL) {
```

docs/reference/ImageGrab.rst
```diff
@@ -39,6 +39,11 @@ or the clipboard to a PIL image memory.
         You can check X11 support using :py:func:`PIL.features.check_feature` with ``feature="xcb"``.
 
         .. versionadded:: 7.1.0
+
+    :param handle:
+        HWND, to capture a single window. Windows only.
+
+        .. versionadded:: 11.1.0
```
hugovk, line 46:
```suggestion
        .. versionadded:: 11.2.0
```

hugovk (approved):

adabru:
> FWIW not all windows can be captured like this. Notably, Vivaldi (my browser) only shows as a black box of the correct size, and the Windows file explorer is missing its toolbar. Other applications, e.g. PyCharm or Microsoft Spy++, are captured correctly. But I think that is just a limitation of modern Windows applications and nothing we can easily fix.

I have a similar problem with games made with Unity3d where the second and subsequent screenshots just repeat the first screenshot I took from that window. Maybe the limitation could be mentioned in the doc: https://pillow.readthedocs.io/en/stable/reference/ImageGrab.html.

radarhere:
I would like to be able to provide specifics beyond 'Sometimes it doesn't work', but not being a Windows user, I'm unable to easily test further.

If you have a better documentation suggestion, feel free to create a PR for review.

PR #8518, opened 2024-11-01
Added running_in_another_thread()

radarhere:
Possible way to help #8492. Alternative to #8501

I've created a Tests/helper.py function called `running_in_another_thread()` that uses a global variable and `inspect` to determine if a test has already started running in another thread, so that we can return early if we would only like the test to run in one thread at a time.

This is used here in two cases
1. #8492 has found that running a test that calls `register_handler` at the beginning, and removes the handler at the end, fails when the test is run across multiple threads. My thinking is that if it ok for `register_handler` to operate across threads, and that it doesn't need to register multiple handlers in parallel. So this PR returns early from the relevant test if it is already running in another thread.
2. https://github.com/tonybaloney/pytest-freethreaded/issues/10 finds an error when running
https://github.com/python-pillow/Pillow/blob/16372dd951c5f7ab664286e6ad25b2e7ed645358/Tests/test_core_resources.py#L50-L54
across multiple threads. I again think that `Image.core.set_alignment` doesn't need to be run across multiple threads simultaneously, so this PR returns early.

### DIFF
--- a/Tests/helper.py
+++ b/Tests/helper.py
@@ -4,6 +4,7 @@ Helper functions.
 
 from __future__ import annotations
 
+import inspect
 import logging
 import os
 import shutil
@@ -28,6 +29,8 @@ if os.environ.get("SHOW_ERRORS"):
     uploader = "show"
 elif "GITHUB_ACTIONS" in os.environ:
     uploader = "github_actions"
+
+already_running = []
 
 
 def upload(a: Image.Image, b: Image.Image) -> str | None:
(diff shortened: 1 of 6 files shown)


Tests/helper.py
```diff
@@ -165,6 +168,18 @@ def assert_tuple_approx_equal(
...
 
+def running_in_another_thread() -> bool:
+    frameInfo = inspect.stack()[1]
+    identifier = (
+        frameInfo.filename + ":" + frameInfo.function + ":" + str(frameInfo.lineno)
+    )
+    if identifier in already_running:
+        return True
+
+    already_running.append(identifier)
```
lysnikolaou, line 179: I think that this introduces another possible race, where two or more threads all think that the test is not currently running elsewhere and add themselves to the list.

lysnikolaou (commented):
If I were to choose between this and #8501, I'd probably go with #8501. This introduces another piece of global state that we probably can avoid.

PR #8525, opened 2024-11-02
Test oldest Python on 32-bit Windows 2019 and remove AppVeyor

hugovk:
Closes https://github.com/python-pillow/Pillow/issues/7910.
Replaces and closes https://github.com/python-pillow/Pillow/pull/8437.

 * On GitHub Actions, switch the oldest Python (3.9) to test 32-bit Windows 2019
 * Remove AppVeyor from the test suite.


winbuild/README.md
```diff
@@ -11,10 +11,10 @@ For more extensive info, see the [Windows build instructions](build.rst).
 * Requires Microsoft Visual Studio 2017 or newer with C++ component.
 * Requires NASM for libjpeg-turbo, a required dependency when using this script.
 * Requires CMake 3.15 or newer (available as Visual Studio component).
-* Tested on Windows Server 2019 with Visual Studio 2019 Community and Visual Studio 2022 Community (AppVeyor).
 * Tested on Windows Server 2022 with Visual Studio 2022 Enterprise (GitHub Actions).
```
radarhere, line 14:
```suggestion
* Tested on Windows Server 2022 with Visual Studio 2022 Enterprise and Windows Server
  2019 with Visual Studio 2019 Enterprise (GitHub Actions).
```

docs/installation/platform-support.rst
```diff
@@ -48,13 +48,11 @@ These platforms are built and tested for every change.
 | Ubuntu Linux 24.04 LTS (Noble)   | 3.12                       | x86-64, ppc64le,    |
 |                                  |                            | s390x               |
 +----------------------------------+----------------------------+---------------------+
-| Windows Server 2019              | 3.9                        | x86-64              |
+| Windows Server 2019              | 3.9                        | x86                 |
 +----------------------------------+----------------------------+---------------------+
 | Windows Server 2022              | 3.9, 3.10, 3.11,           | x86-64              |
 |                                  | 3.12, 3.13, PyPy3          |                     |
```
radarhere, line 54:
```suggestion
| Windows Server 2022              | 3.10, 3.11, 3.12, 3.13,    | x86-64              |
|                                  | PyPy3                      |                     |
```

nulano (approved):

nulano:
I would still consider whether we want to test on Windows Server 2016, but if this simplifies the CI configs then I suppose this is fine as well.

hugovk:
How useful is testing Windows Server 2016?

It looks like its standard support is EOL and only extended, paid support is available.

https://learn.microsoft.com/en-us/lifecycle/products/windows-server-2016

If it was a Linux distro, we would have dropped it 2.5 years ago.

nulano:
As I understand it, paid extended support is currently only available for Server 2008 and Server 2012: https://learn.microsoft.com/en-us/windows-server/get-started/extended-security-updates-overview

For Server 2016 and Server 2019, there appears to be 10 years of regular support: https://learn.microsoft.com/en-us/windows-server/get-started/servicing-channels-comparison

> With the Long-Term Servicing Channel, a new major version of Windows Server is typically released every 2-3 years. Users are entitled to five years of mainstream support and five years of extended support.

In fact, it seems that even Windows 10 Enterprise version 1507 (which is older than Server 2016) is still in Extended Support (i.e. free security updates according to https://learn.microsoft.com/en-us/lifecycle/policies/fixed#extended-support), assuming https://endoflife.date/windows has the correct date (I couldn't find any date on Microsoft's website).

This level of support to me sounds similar to Ubuntu Maintenance & Security Support ([currently Ubuntu 20.04 LTS and later](https://endoflife.date/ubuntu)) or RHEL Maintenance Support ([currently RHEL 8 and later](https://endoflife.date/rhel)).

We dropped Ubuntu 18 when it left Maintenance & Security Support around May 2023, https://github.com/python-pillow/docker-images/pull/180, and Ubuntu 20 slightly ahead of schedule because it only has Python 3.8 by default, https://github.com/python-pillow/Pillow/pull/8183#discussion_r1660807544.

---

However, I'd be fine with dropping support for VS2017 (dropping https://github.com/python-pillow/Pillow/blob/2e7da079cdadf24a5d17cc4b0d32df434777c789/winbuild/build_prepare.py#L441) and Windows 1507 (replacing this runtime linking with regular dynamic linking https://github.com/python-pillow/Pillow/blob/2e7da079cdadf24a5d17cc4b0d32df434777c789/src/display.c#L346-L348) if we decide that we don't include the only-security-updates phase of standard Windows support.

hugovk:
Thanks for clarifying the support levels.

> However, I'd be fine with dropping support for VS2017 (...) and Windows 1507 (...) if we decide that we don't include the only-security-updates phase of standard Windows support.

I think this can be another discussion :)

I'm leaning towards dropping AppVeyor anyway, and relying on GitHub Actions. We don't necessarily have to test everything we support.

Any further thoughts on this PR?

aclark4life:
> I'm leaning towards dropping AppVeyor anyway, and relying on GitHub Actions. We don't necessarily have to test everything we support.
> 
> Any further thoughts on this PR?

Ideally we'd test everything we support, but not at all costs. If there's a clear path to simplifying the infrastructure and workflow and the risk is that we don't test a small portion of use cases, then I'd definitely proceed because there's a good chance we'll never hear about it.

That said, if we do receive some unexpected negative feedback about the change and smaller test matrix is there any tentative plan to address those issues somehow?

If not, I'd still probably do it! But may be nice to spend a little time thinking about it prior to removing AppVeyor.

hugovk:
OK, let's merge! Thanks all!

PR #8534, opened 2024-11-05
Use test image filename

radarhere:
https://pillow.readthedocs.io/en/latest/reference/ImageFile.html has an example where users are invited to open "hopper.pgm".

Maybe no one else is bothered by this, but hopper.pgm doesn't exist as one of our test files, so I think the example code should be updated to refer to [hopper.ppm](https://github.com/python-pillow/Pillow/blob/main/Tests/images/hopper.ppm) instead.

### DIFF
--- a/docs/reference/ImageFile.rst
+++ b/docs/reference/ImageFile.rst
@@ -19,7 +19,7 @@ Example: Parse an image
 
     from PIL import ImageFile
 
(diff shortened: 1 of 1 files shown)


aclark4life:
Yes, that would have bothered me had I noticed, thanks!

PR #8535, opened 2024-11-05
Raise UnidentifiedImageError when opening TIFF without dimensions

### LINKED ISSUE #8530: Pillow 11.0.0 and Django: TiffImagePlugin raised 'Invalid dimensions' error
Seems that v11.0.0 release add a Bug parsing tiff images:

```pytb
...
  File "django/core/files/images.py", line 63, in get_image_dimensions
    p.feed(data)
  File "PIL/ImageFile.py", line 471, in feed
    im = Image.open(fp)
  File "PIL/Image.py", line 3515, in open
    im = _open_core(fp, filename, prefix, formats)
  File "PIL/Image.py", line 3503, in _open_core
    im = factory(fp, filename)
  File "PIL/TiffImagePlugin.py", line 1153, in __init__
    super().__init__(fp, filename)
  File "PIL/ImageFile.py", line 144, in __init__
    self._open()
  File "PIL/TiffImagePlugin.py", line 1177, in _open
    self._seek(0)
  File "PIL/TiffImagePlugin.py", line 1248, in _seek
    self._setup()
  File "PIL/TiffImagePlugin.py", line 1426, in _setup
    raise ValueError(msg)
```

more info in Sentry message:

grafik

Runtime: CPython v3.11.10(3.11.10 (main, Oct 19 2024, 01:04:28) [GCC 12.2.0])

It works with Pillow 10.4.0

radarhere:
Resolves #8530

#8319 changed
```python
xsize = int(self.tag_v2.get(IMAGEWIDTH))
ysize = int(self.tag_v2.get(IMAGELENGTH))
```
to
https://github.com/python-pillow/Pillow/blob/82199efbf77f369ad9b7c2c3759e08f517d9925b/src/PIL/TiffImagePlugin.py#L1436-L1440

This change seems reasonable, especially when you consider #4103.

However, if IMAGEWIDTH or IMAGELENGTH were missing, then `int(None)` previously raised a TypeError. TypeErrors are caught when opening images
https://github.com/python-pillow/Pillow/blob/82199efbf77f369ad9b7c2c3759e08f517d9925b/src/PIL/Image.py#L3497-L3506
but the ValueErrors are not.

So this change inadvertently broke feeding a TIFF image to `ImageFile.Parser`, which relies on OSError ([which UnidentifiedImageErrors are](https://github.com/python-pillow/Pillow/blob/82199efbf77f369ad9b7c2c3759e08f517d9925b/src/PIL/__init__.py#L77)) to know that there isn't enough data yet.

https://github.com/python-pillow/Pillow/blob/82199efbf77f369ad9b7c2c3759e08f517d9925b/src/PIL/ImageFile.py#L468-L473

This PR changes the code to raise a TypeError again.

### DIFF
--- a/src/PIL/TiffImagePlugin.py
+++ b/src/PIL/TiffImagePlugin.py
@@ -1433,8 +1433,12 @@ class TiffImageFile(ImageFile.ImageFile):
         logger.debug("- YCbCr subsampling: %s", self.tag_v2.get(YCBCRSUBSAMPLING))
 
         # size
-        xsize = self.tag_v2.get(IMAGEWIDTH)
-        ysize = self.tag_v2.get(IMAGELENGTH)
+        try:
+            xsize = self.tag_v2[IMAGEWIDTH]
+            ysize = self.tag_v2[IMAGELENGTH]
+        except KeyError as e:
+            msg = "Missing dimensions"
+            raise TypeError(msg) from e
         if not isinstance(xsize, int) or not isinstance(ysize, int):
             msg = "Invalid dimensions"
             raise ValueError(msg)
--- a/Tests/test_imagefile.py
+++ b/Tests/test_imagefile.py
@@ -93,6 +93,19 @@ class TestImageFile:
             assert p.image is not None
             assert (48, 48) == p.image.size
 
+    @pytest.mark.filterwarnings("ignore:Corrupt EXIF data")
+    def test_incremental_tiff(self) -> None:
+        with ImageFile.Parser() as p:
+            with open("Tests/images/hopper.tif", "rb") as f:
+                p.feed(f.read(1024))
+
+                # Check that insufficient data was given in the first feed
+                assert not p.image
+
+                p.feed(f.read())
+            assert p.image is not None
+            assert (128, 128) == p.image.size
+
     @skip_unless_feature("webp")
     def test_incremental_webp(self) -> None:
         with ImageFile.Parser() as p:


src/PIL/TiffImagePlugin.py
```diff
@@ -1433,8 +1433,12 @@ def _setup(self) -> None:
...
 
         # size
-        xsize = self.tag_v2.get(IMAGEWIDTH)
-        ysize = self.tag_v2.get(IMAGELENGTH)
+        try:
+            xsize = self.tag_v2[IMAGEWIDTH]
+            ysize = self.tag_v2[IMAGELENGTH]
+        except KeyError as e:
+            msg = "Missing dimensions"
+            raise TypeError(msg) from e
```
jedie, line 1441: This is also a new error type that was not raised in past versions, isn't it? So i will also make trouble in Django...

jedie: Maybe it's a good idea to always raise a special error for all parsers, in all cases where the chunk data will not be enough to get the meta data and use `RuntimeError` as a base, because it's already catch?

radarhere: In 10.4.0, a TypeError would have been raised here by `int(self.tag_v2.get(IMAGEWIDTH))` being run when there was no IMAGEWIDTH tag - effectively, `int(None)`.

This is raising a TypeError again.

It is then [caught in `Image.open` and turned into an `UnidentifiedImageError`](https://github.com/python-pillow/Pillow/blob/82199efbf77f369ad9b7c2c3759e08f517d9925b/src/PIL/Image.py#L3497-L3506), which [inherits from `OSError`](https://github.com/python-pillow/Pillow/blob/82199efbf77f369ad9b7c2c3759e08f517d9925b/src/PIL/__init__.py#L77), and is then [caught in `Parser.feed()`](https://github.com/python-pillow/Pillow/blob/7fe56d62c6739a3aebcc80e2a2cf2c8700fa3726/src/PIL/ImageFile.py#L469-L473), and the Django loop continues.

jedie (approved):

PR #8545, opened 2024-11-09
Removed use of os.path.realpath

### LINKED ISSUE #8543: Use of `os.path.realpath()` leads to heavy metadata IOPS on some systems
The Pillow functions `Image.open()` and `Image.save()`, among others, use `os.path.realpath()` on the input path to transform it to a symlink-free canonical path.

  - https://github.com/python-pillow/Pillow/blob/d642d2c36f268558b8a217829358aa54758a8ec5/src/PIL/Image.py#L2553
  - https://github.com/python-pillow/Pillow/blob/d642d2c36f268558b8a217829358aa54758a8ec5/src/PIL/Image.py#L2562
  - https://github.com/python-pillow/Pillow/blob/d642d2c36f268558b8a217829358aa54758a8ec5/src/PIL/Image.py#L3466

There is no reliance on this path's "realness" afterwards that is obvious to me, and `realpath()` requires an `lstat()` system call on _every single path segment_ from `/` on down to the file in question _(or more, if symlinks are detected along the way)_. On clusters with distributed filesystems such as BeeGFS, this leads and is seen as a multiplication _(by a factor greater than or equal to the depth of the image file within the hierarchy)_ of the number of metadata I/O operations, and causes DDoS of the metadata servers for those filesystems.

Pillow is heavily utilized in AI workloads with image datasets and data-loaders, a popular usecase, making Pillow's use of `realpath()` especially painful. PyTorch's `ImageFolder` is the [classic example](https://github.com/pytorch/vision/blob/4249b610811b290ea9ac9e445260be195ce52ae1/torchvision/datasets/folder.py#L260-L264).

Please reconsider all uses of `realpath()` in Pillow. Preferentially remove it entirely. In the alternative that you absolutely need the last segment of the path to be translated to acquire a "true" file-extension, consider `readlink()`.

radarhere:
Resolves #8543

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -2550,7 +2550,7 @@ class Image:
         filename: str | bytes = ""
         open_fp = False
         if is_path(fp):
-            filename = os.path.realpath(os.fspath(fp))
+            filename = os.fspath(fp)
             open_fp = True
         elif fp == sys.stdout:
             try:
@@ -2559,7 +2559,7 @@ class Image:
                 pass
         if not filename and hasattr(fp, "name") and is_path(fp.name):
             # only set the name for metadata purposes
-            filename = os.path.realpath(os.fspath(fp.name))
+            filename = os.fspath(fp.name)
 
         # may mutate self!
         self._ensure_mutable()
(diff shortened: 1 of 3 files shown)


hugovk:
Thank you.

PR #8560, opened 2024-11-20
Fix file position desync when calling into libtiff

### LINKED ISSUE #8559: TIFF image loading EXIF corruption
### What did you do?
Load a TIFF with exif rotation info and save as JPG

Save image

### What actually happened?
save() crashes

* OS: Debian GNU/Linux 12 `Linux dev 6.8.12-2-pve #1 SMP PREEMPT_DYNAMIC PMX 6.8.12-2 (2024-09-05T10:03Z) x86_64 GNU/Linux`
* Python: Python 3.11.2 (main, Aug 26 2024, 07:20:54) [GCC 12.2.0]
* Pillow: 11.1.0.dev0

```text
Pillow 11.1.0.dev0
Python 3.11.2 (main, Aug 26 2024, 07:20:54) [GCC 12.2.0]
Python Pillow modules loaded from /home/tom/.local/lib/python3.11/site-packages/PIL
Binary Pillow modules loaded from /home/tom/.local/lib/python3.11/site-packages/PIL
--- PIL CORE support ok, compiled for 11.1.0.dev0
--- FREETYPE2 support ok, loaded 2.12.1
--- LITTLECMS2 support ok, loaded 2.14
--- WEBP support ok, loaded 1.2.4
--- JPEG support ok, compiled for libjpeg-turbo 2.1.5
--- ZLIB (PNG/ZIP) support ok, loaded 1.2.13
--- LIBTIFF support ok, loaded 4.5.0
```

```python
img_path = pathlib.Path('IMG_0282.CR2')
img = PIL.Image.open(img_path)
img.save(img_path.with_suffix('.jpg'), format='JPEG')

  File "/mnt/zp/tom/docs/Programming/bugreports/pillow/pillow_bug2.py", line 27, in save
    img.save(img_path.with_suffix('.jpg'), format='JPEG')
  File "/home/tom/.local/lib/python3.11/site-packages/PIL/Image.py", line 2605, in save
  File "/home/tom/.local/lib/python3.11/site-packages/PIL/JpegImagePlugin.py", line 843, in _save
    ImageFile._save(
  File "/home/tom/.local/lib/python3.11/site-packages/PIL/ImageFile.py", line 556, in _save
    _encode_tile(im, fp, tile, bufsize, fh)
  File "/home/tom/.local/lib/python3.11/site-packages/PIL/ImageFile.py", line 576, in _encode_tile
    encoder.setimage(im.im, extents)
```

### Debugging Notes

I have debugged this, and determined it's caused by:
- save() calls load() calls into _load_libtiff()
- `self.fp: _io.BufferedReader` is corrupted after giving it to libtiff `decode.decode()`
- `load_end()` calls `exif_transpose()` calls `getexif()` calls `ImageFileDirectory_v2.load`
- since `fp` is corrupt, no valid tags are read. you can see this in the logs of many unknown tags
- no valid exif rotation tag means `Image.transpose()` is not called in `load_end()`
- the image/tile size is confused (WxH swapped) somewhere. causing the tile size check to fail

### Workarounds

I've come up with a handful of workarounds:

1) Don't buffer
```python
img = PIL.Image.open(img_path.open('rb', buffering=0))
```
The buffer can't be corrupted if there isn't one.

2) Pre-read exif tags
```python
img = PIL.Image.open(img_path)
img.getexif()
```
This calls `ImageFileDirectory_v2.load` on a virgin fp, so it's not corrupt. Subsequent calls in Image.load are cached.

3) Poke the buffer

Add after here https://github.com/python-pillow/Pillow/blob/main/src/PIL/TiffImagePlugin.py#L1385
```python
            n, err = decoder.decode(b"fpfp")
            if not close_self_fp:
                self.fp.seek(0, 2)
                assert self.fp.read(1) == b''
                self.fp.seek(0)
```
This gets the buffer to drop its cache.
I got the idea from comments and similar code here: https://github.com/python-pillow/Pillow/commit/fd299e36cec4c12a0f201dbbbc1014dfaca7005a

However, I don't think this is a safe idea. This is not a documented behavior and the buffering/caching behavior of this could change at any time. When I tried with just `seek(0)`, it did not help, and looking at the CPython code, it looks like seek() might be optimized if it's a location that's already buffered and won't actually move the file handle.

[pillow.zip](https://github.com/user-attachments/files/17789458/pillow.zip)

Knio:
Fixes #8559

 * Save and restore file handle position when passing file handle to libtiff


src/PIL/TiffImagePlugin.py
```diff
@@ -1381,8 +1373,17 @@ def _load_libtiff(self) -> Image.core.PixelAccess | None:
             logger.debug("have fileno, calling fileno version of the decoder.")
             if not close_self_fp:
                 self.fp.seek(0)
+            # Save and restore the file position, because libtiff will move it
+            # outside of the python runtime, and that will confuse
```
radarhere, line 1377:
```suggestion
            # outside of the Python runtime, and that will confuse
```

Tests/test_libtiff_file_position.py
```diff
@@ -0,0 +1,25 @@
...
+
+from PIL import Image
+
+
+@pytest.mark.parametrize(
+    "test_file",
+    [
+        "Tests/images/old-style-jpeg-compression-no-samplesperpixel.tif",
+        "Tests/images/old-style-jpeg-compression.tif",
+    ],
```
radarhere, line 13: Is there anything different between these two files that means that one might conceivably fail where the other wouldn't? Or are you just testing two to be thorough?

radarhere (approved):

PR #8567, opened 2024-11-25
Derive dir from filename if root is the same in winbuild

radarhere:
https://github.com/python-pillow/Pillow/blob/4304414473768e581bfb6cf519c8a1f9065ad1df/winbuild/build_prepare.py#L134-L135

It would seem like the "dir" could be derived from the "filename", simply by removing the file extension.

### DIFF
--- a/winbuild/build_prepare.py
+++ b/winbuild/build_prepare.py
@@ -131,7 +131,6 @@ DEPS: dict[str, dict[str, Any]] = {
     "libjpeg": {
         "url": f"{SF_PROJECTS}/libjpeg-turbo/files/{V['JPEGTURBO']}/FILENAME/download",
         "filename": f"libjpeg-turbo-{V['JPEGTURBO']}.tar.gz",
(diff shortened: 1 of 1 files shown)


winbuild/build_prepare.py
```diff
@@ -763,6 +750,8 @@ def main() -> None:
     }
 
     for k, v in DEPS.items():
+        if "dir" not in v:
+            v["dir"] = re.sub(r"\.(tar\.gz|zip)", "", v["filename"])
```
hugovk, line 754: Or with https://docs.python.org/3/library/os.path.html#os.path.basename ?
```suggestion
            v["dir"] = os.path.basename(v["filename"])
```

radarhere: `os.path.basename` is for removing the directories before the filename, which isn't the intention here. I presume you meant `os.path.splitext`. Sadly, that won't suffice either.

```pycon
>>> import os
>>> os.path.splitext("a.tar.gz")
('a.tar', '.gz')
```

PR #8574, opened 2024-11-29
Use brew formula to install libraqm

radarhere:
https://pillow.readthedocs.io/en/latest/installation/building-from-source.html

> To install libraqm on macOS use Homebrew to install its dependencies:
> 
> `brew install freetype harfbuzz fribidi`
> Then see `depends/install_raqm_cmake.sh` to install libraqm.

This documented process can be made simpler by suggesting that users just `brew install libraqm` instead.

We have been using the libraqm formula in GitHub Actions since https://github.com/python-pillow/Pillow/pull/5061.


hugovk:
Much better!

PR #8577, opened 2024-12-01
Only use start year in copyright, remove end years

hugovk:
Over in CPython, after consulting with the PSF's legal representative, we decided to remove the end years from the copyright notices because there's no need to update them each January.

See:

* https://devguide.python.org/getting-started/pull-request-lifecycle/#copyrights
* https://github.com/python/cpython/issues/126133#issuecomment-2460824052
* https://github.com/python/release-tools/pull/196

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -5,7 +5,7 @@ The Python Imaging Library (PIL) is
 
 Pillow is the friendly PIL fork. It is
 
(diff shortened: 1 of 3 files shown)


radarhere (approved):

aclark4life:
LGTM

PR #8500, opened 2024-10-25
Use zlib-ng instead of zlib

nulano:
Alternative to #8495 .

Changes proposed in this pull request:

 * Replace zlib with zlib-ng built with CMake in Windows builds. Building our own zlib-ng allows us to use the latest release since the zlib-ng 2.2.2 release is missing the x86_64 Windows build.
 * Add a flag to `PIL.features` for zlib-ng. `python3 -m PIL.report` now looks like this:

   ```
   --------------------------------------------------------------------
   Pillow 11.1.0.dev0
   Python 3.11.9 (tags/v3.11.9:de54cf5, Apr  2 2024, 10:12:12) [MSC v.1938 64 bit (AMD64)]
   --------------------------------------------------------------------
   Python executable is C:\hostedtoolcache\windows\Python\3.11.9\x64\python.exe
   System Python files loaded from C:\hostedtoolcache\windows\Python\3.11.9\x64
   --------------------------------------------------------------------
   Python Pillow modules loaded from C:\hostedtoolcache\windows\Python\3.11.9\x64\Lib\site-packages\PIL
   Binary Pillow modules loaded from C:\hostedtoolcache\windows\Python\3.11.9\x64\Lib\site-packages\PIL
   --------------------------------------------------------------------
   --- PIL CORE support ok, compiled for 11.1.0.dev0
   --- TKINTER support ok, loaded 8.6
   --- FREETYPE2 support ok, loaded 2.13.3
   --- LITTLECMS2 support ok, loaded 2.16
   --- WEBP support ok, loaded 1.4.0
   --- JPEG support ok, compiled for libjpeg-turbo 3.0.4
   --- OPENJPEG (JPEG2000) support ok, loaded 2.5.2
   --- ZLIB (PNG/ZIP) support ok, loaded 1.3.1.zlib-ng, compiled for zlib-ng 2.2.2
   --- LIBTIFF support ok, loaded 4.6.0
   --- RAQM (Bidirectional Text) support ok, loaded 0.10.1, fribidi 1.0.16, harfbuzz 10.0.1
   --- LIBIMAGEQUANT (Quantization method) support ok, loaded 2.16.0
   *** XCB (X protocol) support not installed
   --------------------------------------------------------------------
   ```

I took inspiration from https://github.com/python-pillow/Pillow/pull/8495#issuecomment-2436585125 to make a benchmark with multiple compression values. It seems that zlib-ng usually produces a slightly larger file but in much less time during compression.

<details>

[...]

PR #8599, opened 2024-12-14
Added release notes for using zlib-ng instead of zlib

radarhere:
Re: #8500.

### DIFF
--- a/docs/releasenotes/11.1.0.rst
+++ b/docs/releasenotes/11.1.0.rst
@@ -1,0 +1,59 @@
+11.1.0
+------
+
+Security
+========
+
+TODO
+^^^^
+
+TODO
+
+:cve:`YYYY-XXXXX`: TODO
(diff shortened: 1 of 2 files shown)


docs/releasenotes/11.1.0.rst
```diff
@@ -0,0 +1,58 @@
...
+    features.check_feature("zlib_ng")  # True or False
+    features.version_feature("zlib_ng")  # "2.2.2" for example, or None
+
+Other Changes
+=============
+
+zlib-ng use in wheels
+^^^^^^^^^^^^^^^^^^^^^
+
+Wheels are now built against zlib-ng, for improved speed.
```
hugovk, line 58: Can we quantify it? Perhaps something like "more than twice as fast for higher compression levels".

hugovk:
Thanks!

PR #8617, opened 2024-12-23
Assert fp is not None

radarhere:
This is part of #8362 - I'm hoping to break down that PR into easier-to-review chunks.

ImageFile's `fp` may be `None`.

```pycon
>>> from PIL import Image
>>> im = Image.open("Tests/images/hopper.png")
>>> im.fp
<_io.BufferedReader name='Tests/images/hopper.png'>
>>> im.close()
>>> im.fp is None
True
```

So before using it, let's assert that it isn't `None`.

### DIFF
--- a/Tests/test_file_libtiff.py
+++ b/Tests/test_file_libtiff.py
@@ -11,7 +11,15 @@ from typing import Any, NamedTuple
 
 import pytest
 
-from PIL import Image, ImageFilter, ImageOps, TiffImagePlugin, TiffTags, features
+from PIL import (
+    Image,
+    ImageFile,
+    ImageFilter,
+    ImageOps,
+    TiffImagePlugin,
+    TiffTags,
+    features,
+)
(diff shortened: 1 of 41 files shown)


Tests/test_file_libtiff.py
```diff
@@ -556,8 +564,9 @@ def test_bw_compression_w_rgb(self, compression: str, tmp_path: Path) -> None:
             im.save(out, compression=compression)
 
     def test_fp_leak(self) -> None:
-        im: Image.Image | None = Image.open("Tests/images/hopper_g4_500.tif")
+        im: ImageFile.ImageFile | None = Image.open("Tests/images/hopper_g4_500.tif")
```
radarhere, line 567: `Image.open` returns an `ImageFile` instance.
https://github.com/python-pillow/Pillow/blob/0e3f51dec67991e4dee53bcc149218d6c65cf820/src/PIL/Image.py#L3404-L3408

PR #8622, opened 2024-12-25
Added reading of J2K comments

radarhere:
#6909 added reading of JP2 comments. This adds reading of J2K comments as well.

JP2 comments are read from a ["Contiguous codestream box"](https://www.ics.uci.edu/~dan/class/267/papers/jpeg2000.pdf) that "contains a valid and complete JPEG 2000 codestream"

https://github.com/python-pillow/Pillow/blob/0e3f51dec67991e4dee53bcc149218d6c65cf820/src/PIL/Jpeg2KImagePlugin.py#L264-L265

and starts by skipping `hdr`.

https://github.com/python-pillow/Pillow/blob/0e3f51dec67991e4dee53bcc149218d6c65cf820/src/PIL/Jpeg2KImagePlugin.py#L298-L301

Reading from a J2K image,

https://github.com/python-pillow/Pillow/blob/0e3f51dec67991e4dee53bcc149218d6c65cf820/src/PIL/Jpeg2KImagePlugin.py#L252-L254

the `hdr` is already read.

https://github.com/python-pillow/Pillow/blob/0e3f51dec67991e4dee53bcc149218d6c65cf820/src/PIL/Jpeg2KImagePlugin.py#L101-L107

So this PR moves the `hdr` handling out of `_parse_comment`.

### DIFF
--- a/Tests/test_file_jpeg2k.py
+++ b/Tests/test_file_jpeg2k.py
@@ -424,8 +424,9 @@ def test_pclr() -> None:
 
 
 def test_comment() -> None:
-    with Image.open("Tests/images/comment.jp2") as im:
-        assert im.info["comment"] == b"Created by OpenJPEG version 2.5.0"
+    for path in ("Tests/images/9bit.j2k", "Tests/images/comment.jp2"):
+        with Image.open(path) as im:
+            assert im.info["comment"] == b"Created by OpenJPEG version 2.5.0"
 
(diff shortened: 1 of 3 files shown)


hugovk (approved):
> #6909 added reading of JP2 comments. This adds reading of J2K comments as well.

Let's also add release notes like that one did.

Issue #8614, opened 2024-12-21
Exif: MakerNote enums spelled differently


mixmastamyk:
https://github.com/radarhere/Pillow/blob/main/src/PIL/ExifTags.py#L156

    MakerNote = 0x927C

https://github.com/radarhere/Pillow/blob/main/src/PIL/ExifTags.py#L355

    Makernote = 37500

These names should be consistent.  Also, some values specified in hex and others in decimal is also a bit messy, though merely cosmetic.

radarhere:
I've created https://github.com/python-pillow/Pillow/pull/8615 to resolve this.

Also, just as an aside - you've linked to my fork, rather than to this repository.

mixmastamyk:
Thanks.  However, it would probably be good to carry the old spelling for a while, marked deprecated as a courtesy.  Otherwise will break some users.  Unless there is a big breaking release coming up.

radarhere:
I've created https://github.com/python-pillow/Pillow/pull/8629

PR #8629, opened 2024-12-28
Restored Makernote as a deprecated enum

radarhere:
#8615 replaced `ExifTags.IFD.Makernote` with `ExifTags.IFD.MakerNote`.

https://github.com/python-pillow/Pillow/issues/8614#issuecomment-2564126641 has requested that we keep `ExifTags.IFD.Makernote` for now, as a deprecated enum.

### DIFF
--- a/src/PIL/ExifTags.py
+++ b/src/PIL/ExifTags.py
@@ -353,6 +353,7 @@ class IFD(IntEnum):
     Exif = 0x8769
     GPSInfo = 0x8825
     MakerNote = 0x927C
+    Makernote = 0x927C  # Deprecated
     Interop = 0xA005
     IFD1 = -1
 
--- a/docs/deprecations.rst
+++ b/docs/deprecations.rst
@@ -175,6 +175,14 @@ deprecated and will be removed in Pillow 12 (2025-10-15). They were used for obt
 raw pointers to ``ImagingCore`` internals. To interact with C code, you can use
 ``Image.Image.getim()``, which returns a ``Capsule`` object.
 
+ExifTags.IFD.Makernote
+^^^^^^^^^^^^^^^^^^^^^^
+
+.. deprecated:: 11.1.0
+
+``ExifTags.IFD.Makernote`` has been deprecated. Instead, use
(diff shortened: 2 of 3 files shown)


src/PIL/ExifTags.py
```diff
@@ -353,6 +353,7 @@ class IFD(IntEnum):
     Exif = 0x8769
     GPSInfo = 0x8825
     MakerNote = 0x927C
+    Makernote = 0x927C  # Deprecated
```
hugovk, line 356: Can we raise a deprecation warning when accessed? 

We've had problems before, removing `VERSION` after only deprecating in docs.

radarhere: I've tried `__getattr__`, `__getitem__` and `@property`, and I don't think so, no.

It was only added in #6748 in Pillow 9.4.0, so I would expect this to have smaller use.

radarhere:
Something to be aware of - with the current state of this PR,
```python
from PIL import ExifTags
print(ExifTags.IFD.Makernote)
```
will give
```
IFD.MakerNote
```

src/PIL/ExifTags.py
hugovk: Thanks for checking. 

Yeah, I expect this should cause much less breakage.

Still, if we can't warn in code, perhaps we should just deprecate this without planning to remove it? It doesn't cost anything to keep this one line.

radarhere: Do you mean just restore the enum value, and add a comment? Done.

hugovk: Yes, and we can also document it as deprecated, but don't mention any removal date:

```diff
-``ExifTags.IFD.Makernote`` has been deprecated and will be removed in Pillow 13
-(2026-10-15). Instead, use ``ExifTags.IFD.MakerNote``.
+``ExifTags.IFD.Makernote`` has been deprecated. Instead, use ``ExifTags.IFD.MakerNote``.
```

hugovk (approved):
Thank you!

PR #8526, opened 2024-11-02
Apply security fixes to GitHub Actions

hugovk:
There's a new tool that flags potential security issues in GitHub Actions workflows: https://github.com/woodruffw/zizmor

<details>
<summary>Details</summary>

```console
❯ zizmor .
🌈 completed cifuzz.yml
🌈 completed release-drafter.yml
🌈 completed stale.yml
🌈 completed docs.yml
🌈 completed test-valgrind.yml
🌈 completed test-windows.yml
🌈 completed test-mingw.yml
🌈 completed test-docker.yml
🌈 completed lint.yml
🌈 completed test.yml
🌈 completed test-cygwin.yml
🌈 completed wheels.yml
error[excessive-permissions]: overly broad workflow or job-level permissions
 --> /Users/hugo/github/Pillow/.github/workflows/stale.yml:8:1
  |
8 | / permissions:
9 | |   issues: write
  | |_______________^ issues: write is overly broad at the workflow level
  |

warning[artipacked]: credential persistence through GitHub Actions artifacts
  --> /Users/hugo/github/Pillow/.github/workflows/docs.yml:35:7
   |
35 |     - uses: actions/checkout@v4
   |       ------------------------- does not set persist-credentials: false
   |

warning[artipacked]: credential persistence through GitHub Actions artifacts
  --> /Users/hugo/github/Pillow/.github/workflows/test-valgrind.yml:42:7
   |
42 |     - uses: actions/checkout@v4
   |       ------------------------- does not set persist-credentials: false
   |

warning[artipacked]: credential persistence through GitHub Actions artifacts
  --> /Users/hugo/github/Pillow/.github/workflows/test-windows.yml:45:7
   |
45 |       - name: Checkout Pillow
   |  _______-
46 | |       uses: actions/checkout@v4
   | |_______________________________- does not set persist-credentials: false
   |

warning[artipacked]: credential persistence through GitHub Actions artifacts
  --> /Users/hugo/github/Pillow/.github/workflows/test-windows.yml:48:7
   |
48 |       - name: Checkout cached dependencies
   |  _______-
49 | |       uses: actions/checkout@v4
50 | |       with:
51 | |         repository: python-pillow/pillow-depends
52 | |         path: winbuild\depends
   | |______________________________- does not set persist-credentials: false
   |

warning[artipacked]: credential persistence through GitHub Actions artifacts
  --> /Users/hugo/github/Pillow/.github/workflows/test-windows.yml:54:7
   |
54 |       - name: Checkout extra test images
[...]

PR #8638, opened 2024-12-29
Add zizmor to pre-commit and fix potential cache-poisoning in wheels workflow

hugovk:
Follow on from https://github.com/python-pillow/Pillow/pull/8526.

Add zizmor to pre-commit, run the new version 0.10.0, and fix the one new thing it finds:

```
error[cache-poisoning]: runtime artifacts potentially vulnerable to a cache poisoning attack
   --> /Users/hugo/github/Pillow/.github/workflows/wheels.yml:3:1
    |
  3 | / on:
  4 | |   schedule:
...   |
 29 | |       - "winbuild/fribidi.cmake"
 30 | |   workflow_dispatch:
    | |____________________^ generally used when publishing artifacts generated at runtime
 31 |
...
263 |         uses: actions/setup-python@v5
264 | /       with:
265 | |         python-version: "3.x"
266 | |         cache: pip
267 | |         cache-dependency-path: "Makefile"
    | |_________________________________________^ opt-in for caching here
    |
    = note: audit confidence → Low

54 findings (53 suppressed): 0 unknown, 0 informational, 0 low, 0 medium, 1 high
```

More info: https://woodruffw.github.io/zizmor/audits/#cache-poisoning

In short, the idea is not to use caches in workflows that produce release artifacts.

This featured in the recent Ultralytics supply chain attack:

* https://blog.pypi.org/posts/2024-12-11-ultralytics-attack-analysis/
* https://blog.yossarian.net/2024/12/06/zizmor-ultralytics-injection

We don't run the wheels workflow that often, and the main time bottleneck is building and testing all the wheels, so it's not a big loss to download things from PyPI each time.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -263,8 +263,6 @@ jobs:
       uses: actions/setup-python@v5
       with:
         python-version: "3.x"
-        cache: pip
-        cache-dependency-path: "Makefile"
 
     - run: make sdist
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.8.1
+    rev: v0.8.4
     hooks:
       - id: ruff
         args: [--exit-non-zero-on-fix]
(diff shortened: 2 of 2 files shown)


radarhere:
The change is only for the sdist job, which is miles faster than any other wheel job, so I actually wouldn't consider it a speed loss at all.

PR #8642, opened 2024-12-30
Allow saving as BigTIFF

### LINKED ISSUE #4669: Saving TIFF in chunks

radarhere:
Resolves #4669

Adds a keyword argument to allow TIFF images to be saved as BigTIFF - `im.save("out.tiff", big_tiff=True)`

### DIFF
--- a/docs/handbook/image-file-formats.rst
+++ b/docs/handbook/image-file-formats.rst
@@ -1208,6 +1208,9 @@ The :py:meth:`~PIL.Image.Image.save` method can take the following keyword argum
 
     .. versionadded:: 8.4.0
 
+**bigtiff**
+    If true, the image will be saved as a BigTIFF.
+
 **compression**
     A string containing the desired compression method for the
     file. (valid only with libtiff installed) Valid compression
--- a/docs/handbook/image-file-formats.rst
+++ b/docs/handbook/image-file-formats.rst
@@ -1208,7 +1208,7 @@ The :py:meth:`~PIL.Image.Image.save` method can take the following keyword argum
 
     .. versionadded:: 8.4.0
 
-**bigtiff**
+**big_tiff**
     If true, the image will be saved as a BigTIFF.
 
 **compression**
--- a/docs/handbook/image-file-formats.rst
+++ b/docs/handbook/image-file-formats.rst
@@ -1211,6 +1211,8 @@ The :py:meth:`~PIL.Image.Image.save` method can take the following keyword argum
 **big_tiff**
     If true, the image will be saved as a BigTIFF.
 
+    .. versionadded:: 11.1.0
+
 **compression**
     A string containing the desired compression method for the
     file. (valid only with libtiff installed) Valid compression
--- a/src/PIL/TiffImagePlugin.py
+++ b/src/PIL/TiffImagePlugin.py
@@ -582,7 +582,7 @@ class ImageFileDirectory_v2(_IFDv2Base):
 
     def __init__(
         self,
-        ifh: bytes = b"II\052\0\0\0\0\0",
+        ifh: bytes = b"II\x2A\x00\x00\x00\x00\x00",
         prefix: bytes | None = None,
         group: int | None = None,
     ) -> None:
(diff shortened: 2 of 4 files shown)


src/PIL/TiffImagePlugin.py
```diff
@@ -582,7 +582,7 @@ class ImageFileDirectory_v2(_IFDv2Base):
 
     def __init__(
         self,
-        ifh: bytes = b"II\052\0\0\0\0\0",
+        ifh: bytes = b"II\x2A\x00\x00\x00\x00\x00",
```
radarhere, line 585: This is the same, I'm just changing it to the form seen elsewhere in this file.

src/PIL/TiffImagePlugin.py
```diff
@@ -949,28 +949,34 @@ def load(self, fp: IO[bytes]) -> None:
             warnings.warn(str(msg))
             return
 
+    def _get_ifh(self):
+        ifh = self._prefix + self._pack("H", 43 if self._bigtiff else 42)
+        if self._bigtiff:
+            ifh += self._pack("HH", 8, 0)
+        ifh += self._pack("Q", 16) if self._bigtiff else self._pack("L", 8)
```
radarhere, line 956: See https://web.archive.org/web/20240706160214/https://www.awaresystems.be/imaging/tiff/bigtiff.html for a description of the header values.

docs/handbook/image-file-formats.rst
```diff
@@ -1208,6 +1208,9 @@ The :py:meth:`~PIL.Image.Image.save` method can take the following keyword argum
 
     .. versionadded:: 8.4.0
 
+**bigtiff**
```
hugovk, line 1211: I suggest adding an underscore to the public API for readability (and in the code):
```suggestion
**big_tiff**
```

radarhere: Ok, done.

PR #8644, opened 2024-12-31
Remove pre-C99 definitions

radarhere:
See #6516 for previous discussion about whether C99 can be the minimum version.


hugovk (approved):
Let's also include in the release notes.

PR #8651, opened 2025-01-01
Corrected BLP1 alpha depth handling

### LINKED ISSUE #8650: BLP 1 header is not read correctly
Hi,

I found your project while searching an issue with loading BLP files into Unity. You did an amazing job!

I have a blp file (version 1) from Warcaft 3 I'd like to load into Unity and since the content is JPEG, Unity assumes it has no alpha channel but I really suspect it has. 

I tried loading it with the following python code:
```python
from PIL import Image

with Image.open("Peasant.blp") as im:
    im.getchannel('A')
    im.show()
``` 

And it failed saying there is no alpha channel. The call to show the image display a RGB (no alpha) image that is correct.

I have been searching in your code how you detect alpha channel and I found an error:
https://github.com/python-pillow/Pillow/blob/aa0f4127b8040b82822808fa2037fea1a162d3c5/src/PIL/BlpImagePlugin.py#L300

The BLP header for version 1 is incorrect.
You read 4 bytes there for 4 different values but in version 1, these 4 bytes are actually one unsigned integer 32bits (4 bytes) representing the number of bits for the alpha channel (which in my case is 8).
See this thread for more information about BLP specifications : https://www.hiveworkshop.com/threads/blp-specifications-wc3.279306/

Hope it helps.

radarhere:
Resolves #8650

https://www.hiveworkshop.com/threads/blp-specifications-wc3.279306/ describes the start of the 'BLP File Header' as
```
uint32_t magic;
uint32_t content;
if (version >= 2) {
    uint8_t encodingType; // not documented
    uint8_t alphaBits;
    uint8_t sampleType; // not documented
    uint8_t hasMipmaps;
} else {
    uint32_t alphaBits;
}
```

However, we don't currently perform different behaviour for BLP1, and only treat the data as if it was BLP2.
https://github.com/python-pillow/Pillow/blob/aa0f4127b8040b82822808fa2037fea1a162d3c5/src/PIL/BlpImagePlugin.py#L300-L303

This PR fixes updates how the header data is handled for BLP1, for both reading and writing.

### DIFF
--- a/Tests/test_file_blp.py
+++ b/Tests/test_file_blp.py
@@ -4,7 +4,7 @@ from pathlib import Path
 
 import pytest
 
-from PIL import Image
+from PIL import BlpImagePlugin, Image
 
 from .helper import (
     assert_image_equal,
@@ -36,6 +36,13 @@ def test_load_blp2_dxt1() -> None:
 def test_load_blp2_dxt1a() -> None:
     with Image.open("Tests/images/blp/blp2_dxt1a.blp") as im:
         assert_image_equal_tofile(im, "Tests/images/blp/blp2_dxt1a.png")
+
+
+def test_invalid_file() -> None:
+    invalid_file = "Tests/images/flower.jpg"
+
+    with pytest.raises(BlpImagePlugin.BLPFormatError):
+        BlpImagePlugin.BlpImageFile(invalid_file)
 
 
 def test_save(tmp_path: Path) -> None:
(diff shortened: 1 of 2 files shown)


sminjard (approved):
Wow, I didn't expect to hear from you this soon.
Thanks!

Everything seems fine. A bit confused about why you had to change things in the `_open`function. I think there is a way to call `_read_blp_header` so that you don't have that logic at two places, but this will go in another PR as this one resolves the issue.

Happy New Year !

sminjard (approved):

sminjard (approved):

sminjard (approved):

sminjard:
Somehow GitHub was not checking the "#approved-reviews-by>=1" for auto merge check.

hugovk:
This was a manual merge by me, so it didn't use automerge, which would have also needed the label adding and approval by a committer.

PR #8660, opened 2025-01-03
Updated TIFF decode error message string

HarmvandenBrand:
OSError caused by decode error should use string argument to be in line with rest of module

 * Changed an OSError raised in TiffImagePlugin.py during a decoder error to contain a str message as its argument, instead of directly passing the error code as an integer. This change brings this specific OSError in line with the way errors are raised in the rest of TiffImagePlugin.py and thus reduces surprise.

### DIFF
--- a/Tests/test_file_libtiff.py
+++ b/Tests/test_file_libtiff.py
@@ -1146,7 +1146,7 @@ class TestFileLibTiff(LibTiffTestCase):
                 im.load()
 
             # Assert that the error code is IMAGING_CODEC_MEMORY
-            assert str(e.value) == "-9"
+            assert str(e.value) == "decoder error -9"
 
     @pytest.mark.parametrize("compression", ("tiff_adobe_deflate", "jpeg"))
     def test_save_multistrip(self, compression: str, tmp_path: Path) -> None:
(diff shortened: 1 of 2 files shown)


radarhere:
Nitpick suggestion: maybe remove the colon, to more closely match
https://github.com/python-pillow/Pillow/blob/48712f29cec31efd0f55dcb95ce2eaf98357cc00/src/PIL/TiffImagePlugin.py#L1972-L1974
?

Tests/test_file_libtiff.py
```diff
@@ -1146,7 +1146,7 @@ def test_realloc_overflow(self, monkeypatch: pytest.MonkeyPatch) -> None:
                 im.load()
 
             # Assert that the error code is IMAGING_CODEC_MEMORY
-            assert str(e.value) == "-9"
+            assert str(e.value) == "decoder error: -9"
```
radarhere, line 1149:
```suggestion
            assert str(e.value) == "decoder error -9"
```

src/PIL/TiffImagePlugin.py
```diff
@@ -1406,7 +1406,8 @@ def _load_libtiff(self) -> Image.core.PixelAccess | None:
             self.fp = None  # might be shared
 
         if err < 0:
-            raise OSError(err)
+            msg = f"decoder error: {err}"
```
radarhere, line 1409:
```suggestion
            msg = f"decoder error {err}"
```

Tests/test_file_libtiff.py
HarmvandenBrand: Fair enough

radarhere (approved):

PR #8686, opened 2025-01-11
Skip test_embeddable if compiler cannot be initialized

radarhere:
It turns out that when `test_embeddable` was marked as passing in Python 3.13 in #8683, there was an error when using Windows x86 or AMD64 after building the wheels - https://github.com/python-pillow/Pillow/actions/runs/12721987647/job/35465533303#step:7:25500
> distutils.errors.DistutilsPlatformError: Microsoft Visual C++ 14.0 or greater is required. Get it with "Microsoft C++ Build Tools"

Because this problem may occur in a local setup when running the test suite, this PR initializes the compiler (as seen two steps earlier in the error traceback) separately, and if that throws an error, before anything Pillow-related occurs, then the test is skipped.

With this change, the wheels start passing - https://github.com/radarhere/Pillow/actions/runs/12722505250/job/35466599838

### DIFF
--- a/Tests/test_image_access.py
+++ b/Tests/test_image_access.py
@@ -278,6 +278,18 @@ class TestEmbeddable:
 
         from setuptools.command import build_ext
 
+        compiler = getattr(build_ext, "new_compiler")()
+        compiler.add_include_dir(sysconfig.get_config_var("INCLUDEPY"))
+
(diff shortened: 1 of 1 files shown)


hugovk:
Thank you!

PR #8689, opened 2025-01-12
Moved get_child_images() to ImageFile

radarhere:
This is part of #8362 - I'm hoping to break down that PR into easier-to-review chunks.

[`get_child_images()`](https://github.com/python-pillow/Pillow/blob/3111e37bf437ecfdd19b31417aa7e843d742ed0b/src/PIL/Image.py#L1556) relies on `self.fp`, and since that only exists for an `ImageFile` and not an `Image`, I suggest moving it there. This isn't a strictly necessary change, so feel free to disagree.

After that, I've also made changes to fix two mypy errors. One is asserting that `fp` is not `None`, like #8617. The other is asserting that an Exif tag is an integer, to fix
> src/PIL/ImageFile.py:238: error: Argument 1 to "read" of "IO" has incompatible type "Optional[Any]"; expected "int"  [arg-type]
>                         data = self.fp.read(ifd.get(ExifTags.Base.JpegIFByteCount))
>                                             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

### DIFF
--- a/Tests/test_image.py
+++ b/Tests/test_image.py
@@ -989,6 +989,11 @@ class TestImage:
         else:
             assert im.getxmp() == {"xmpmeta": None}
 
+    def test_get_child_images(self) -> None:
+        im = Image.new("RGB", (1, 1))
+        with pytest.warns(DeprecationWarning):
+            assert im.get_child_images() == []
+
     @pytest.mark.parametrize("size", ((1, 0), (0, 1), (0, 0)))
     def test_zero_tobytes(self, size: tuple[int, int]) -> None:
         im = Image.new("RGB", size)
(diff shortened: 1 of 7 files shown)


hugovk:
> [`get_child_images()`](https://github.com/python-pillow/Pillow/blob/3111e37bf437ecfdd19b31417aa7e843d742ed0b/src/PIL/Image.py#L1556) relies on `self.fp`, and since that only exists for an `ImageFile` and not an `Image`, I suggest moving it there. This isn't a strictly necessary change, so feel free to disagree.

I wonder, are there some circumstances where this could be a breaking change?

hugovk:
Yes okay, if you'd like to go ahead with this, I suggest a deprecation cycle, so we don't break things without warning.

radarhere:
Ok, I've deprecated the functionality, to be removed in Pillow 13.

hugovk:
Thanks!

PR #8690, opened 2025-01-13
Test Python 3.14 pre-release

radarhere:
Python 3.14.0a2 has been released - https://www.python.org/downloads/release/python-3140a2/


hugovk:
Thanks!

3.14.0a3 has also been released: https://www.python.org/downloads/release/python-3140a3/

And I'm in the middle of releasing 3.14.0a4 as I type! 🚀

PR #8692, opened 2025-01-13
Thread-local arenas

kddnewton:
## Summary

Currently, all threads use the same arena for imaging. When there are enough workers, in regular Python the GIL will be under lots of contention and in free-threaded Python the mutex will be under lots of contention.

This commit instead introduces lockless thread-local arenas for environments that support it. For environments that do not support thread-locals (or for environments where we couldn't determine if they do or not) we fall back to either the GIL or a mutex if there is no GIL.

This has some implications for statistics, as statistics are now thread-specific. This could be solved in a couple of ways (in C or in Python), or left unsolved and just documented. I think either way is fine.

## Code

Most of the code doesn't actually need to change. The bulk of the changes were getting `setup.py` to emit the proper compilation definitions so that we could check which kind of thread-local declarations were supported at compile-time. Other than that, the declaration of the default arena now has the thread-local declaration and the places where we previously locked the mutex the macro name has changed to reflect that it is specific to the thread-local arena.

## Benchmarks

For regular Python, this didn't make much of a difference. (The difference in the samples wasn't statistically significant, 95% CI). For free-threaded Python, however, the difference was fairly massive (about a 70% increase).

### v3.13.0 on main

```
Max: 0.439743 Mean: 0.355661 Min: 0.305783
Max: 0.415384 Mean: 0.361710 Min: 0.304075
Max: 0.427207 Mean: 0.366160 Min: 0.300381
Max: 0.460026 Mean: 0.388431 Min: 0.316797
Max: 0.419853 Mean: 0.361484 Min: 0.309495
Max: 0.393699 Mean: 0.350330 Min: 0.302294
Max: 0.443584 Mean: 0.372369 Min: 0.311351
Max: 0.404041 Mean: 0.355057 Min: 0.309706
Max: 0.420880 Mean: 0.341415 Min: 0.280980
Max: 0.408922 Mean: 0.320707 Min: 0.228622
```

### v3.13.0t on main

```
Max: 0.218140 Mean: 0.143962 Min: 0.091831
Max: 0.195644 Mean: 0.124187 Min: 0.079139
Max: 0.169986 Mean: 0.124365 Min: 0.081508
Max: 0.194228 Mean: 0.136258 Min: 0.103134
Max: 0.192837 Mean: 0.131196 Min: 0.094301
Max: 0.180463 Mean: 0.126546 Min: 0.079336
Max: 0.181516 Mean: 0.126875 Min: 0.083507
Max: 0.178397 Mean: 0.120558 Min: 0.083620
Max: 0.182262 Mean: 0.129299 Min: 0.087499
Max: 0.167291 Mean: 0.114647 Min: 0.074147
```

### v3.13.0 on branch

```
Max: 0.429302 Mean: 0.362776 Min: 0.314723
Max: 0.406314 Mean: 0.355255 Min: 0.299485
Max: 0.438540 Mean: 0.378539 Min: 0.308898
Max: 0.425942 Mean: 0.368141 Min: 0.310095
Max: 0.408924 Mean: 0.365672 Min: 0.313756
Max: 0.419717 Mean: 0.361498 Min: 0.307699
Max: 0.418639 Mean: 0.355136 Min: 0.314148
Max: 0.426816 Mean: 0.377236 Min: 0.321773
Max: 0.424230 Mean: 0.358225 Min: 0.291148
Max: 0.421029 Mean: 0.363783 Min: 0.315103
```

### v3.13.0t on branch

```
Max: 0.103066 Mean: 0.041306 Min: 0.018575
Max: 0.121496 Mean: 0.043042 Min: 0.018622
Max: 0.129727 Mean: 0.040726 Min: 0.014389
Max: 0.124282 Mean: 0.037581 Min: 0.018034
Max: 0.112015 Mean: 0.042051 Min: 0.017231
Max: 0.123254 Mean: 0.042117 Min: 0.019646
Max: 0.129165 Mean: 0.043886 Min: 0.017393
Max: 0.157608 Mean: 0.045151 Min: 0.017874
Max: 0.117050 Mean: 0.043070 Min: 0.016238
Max: 0.131859 Mean: 0.044563 Min: 0.017736
```

### Script

Below is the script that I used to run these benchmarks.

<details>
<summary>bench.py</summary>

```python
import concurrent.futures
import os
import threading
import time

from PIL import Image

num_threads = 16
num_images = 1024


def operation():
    images = []
    for i in range(num_images):
        img = Image.new(
            "RGB", (100, 100), color=(i % 256, (i // 256) % 256, (i // 65536) % 256)
        )
        images.append(img)

    for img in images:
        img = img.convert("CMYK")

    images.clear()


def worker(barrier):
    barrier.wait()
    runtimes = []

    for _ in range(5):
        start_time = time.time()
        operation()
        end_time = time.time()
        runtimes.append(end_time - start_time)

    return runtimes


def benchmark():
    with concurrent.futures.ThreadPoolExecutor(max_workers=num_threads) as executor:
        barrier = threading.Barrier(num_threads)
        futures = [executor.submit(worker, barrier) for _ in range(num_threads)]

        run_times = []
        for future in concurrent.futures.as_completed(futures):
            try:
                run_times.extend(future.result())
            except IndexError:
                os._exit(-1)

        min_time = min(run_times)
        max_time = max(run_times)
        mean_time = sum(run_times) / len(run_times)
        print(f"Max: {max_time:.6f} Mean: {mean_time:.6f} Min: {min_time:.6f}")


benchmark()
```

</details>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,18 +8,21 @@
 # ------------------------------
 from __future__ import annotations
 
+import distutils.ccompiler
 import os
 import re
 import shutil
 import struct
 import subprocess
 import sys
+import tempfile
 import warnings
 from collections.abc import Iterator
 from typing import Any
 
 from setuptools import Extension, setup
 from setuptools.command.build_ext import build_ext
+from setuptools.errors import CompileError
 
 
 def get_version() -> str:
@@ -292,6 +295,47 @@ def _pkg_config(name: str) -> tuple[list[str], list[str]] | None:
     return None
 
 
+def _try_compile(compiler: distutils.ccompiler.CCompiler, code: str) -> bool:
+    try:
+        with tempfile.TemporaryDirectory() as d:
+            fn = os.path.join(d, "test.c")
+            with open(fn, "w") as f:
+                f.write(code)
+            compiler.compile([fn], output_dir=d, extra_preargs=["-Werror"])
+        return True
+    except CompileError:
+        return False
+
+
+def _try_compile_attr(compiler: distutils.ccompiler.CCompiler, attr: str) -> bool:
+    code = f"""
+        #pragma GCC diagnostic error "-Wattributes"
+        #pragma clang diagnostic error "-Wattributes"
+
+        int {attr} foo;
+        int main() {{
+            return 0;
+        }}
+    """
+
+    return _try_compile(compiler, code)
+
+
+def _try_compile_tls_define_macros(
+    compiler: distutils.ccompiler.CCompiler,
+) -> list[tuple[str, str | None]]:
+    if _try_compile_attr(compiler, "thread_local"):  # C23
+        return [("HAVE_THREAD_LOCAL", None)]
+    elif _try_compile_attr(compiler, "_Thread_local"):  # C11/C17
+        return [("HAVE__THREAD_LOCAL", None)]
+    elif _try_compile_attr(compiler, "__thread"):  # GCC/clang
+        return [("HAVE___THREAD", None)]
+    elif _try_compile_attr(compiler, "__declspec(thread)"):  # MSVC
+        return [("HAVE___DECLSPEC_THREAD_", None)]
+    else:
+        return []
+
+
 class pil_build_ext(build_ext):
     class ext_feature:
         features = [
@@ -426,13 +470,14 @@ class pil_build_ext(build_ext):
     def _update_extension(
         self,
         name: str,
-        libraries: list[str] | list[str | bool | None],
+        libraries: list[str] | list[str | bool | None] | None = None,
         define_macros: list[tuple[str, str | None]] | None = None,
         sources: list[str] | None = None,
     ) -> None:
         for extension in self.extensions:
             if extension.name == name:
-                extension.libraries += libraries
+                if libraries is not None:
+                    extension.libraries += libraries
                 if define_macros is not None:
                     extension.define_macros += define_macros
                 if sources is not None:
@@ -890,7 +935,10 @@ class pil_build_ext(build_ext):
 
         defs.append(("PILLOW_VERSION", f'"{PILLOW_VERSION}"'))
 
-        self._update_extension("PIL._imaging", libs, defs)
+        tls_define_macros = _try_compile_tls_define_macros(self.compiler)
+        self._update_extension("PIL._imaging", libs, defs + tls_define_macros)
+        self._update_extension("PIL._imagingmath", define_macros=tls_define_macros)
+        self._update_extension("PIL._imagingmorph", define_macros=tls_define_macros)
 
         #
         # additional libraries
@@ -913,7 +961,9 @@ class pil_build_ext(build_ext):
                         libs.append(feature.get("fribidi"))
                     else:  # building FriBiDi shim from src/thirdparty
                         srcs.append("src/thirdparty/fribidi-shim/fribidi.c")
-            self._update_extension("PIL._imagingft", libs, defs, srcs)
+            self._update_extension(
+                "PIL._imagingft", libs, defs + tls_define_macros, srcs
+            )
 
         else:
             self._remove_extension("PIL._imagingft")
@@ -922,19 +972,19 @@ class pil_build_ext(build_ext):
             libs = [feature.get("lcms")]
             if sys.platform == "win32":
                 libs.extend(["user32", "gdi32"])
-            self._update_extension("PIL._imagingcms", libs)
+            self._update_extension("PIL._imagingcms", libs, tls_define_macros)
         else:
             self._remove_extension("PIL._imagingcms")
 
         webp = feature.get("webp")
         if isinstance(webp, str):
             libs = [webp, webp + "mux", webp + "demux"]
-            self._update_extension("PIL._webp", libs)
+            self._update_extension("PIL._webp", libs, tls_define_macros)
         else:
             self._remove_extension("PIL._webp")
 
         tk_libs = ["psapi"] if sys.platform in ("win32", "cygwin") else []
-        self._update_extension("PIL._imagingtk", tk_libs)
+        self._update_extension("PIL._imagingtk", tk_libs, tls_define_macros)
 
         build_ext.build_extensions(self)
 
--- a/src/_imaging.c
+++ b/src/_imaging.c
@@ -3938,7 +3938,7 @@ _get_stats(PyObject *self, PyObject *args) {
         return NULL;
     }
 
-    MUTEX_LOCK(&ImagingDefaultArena.mutex);
+    IMAGING_ARENA_LOCK(&ImagingDefaultArena.mutex);
     ImagingMemoryArena arena = &ImagingDefaultArena;
 
     v = PyLong_FromLong(arena->stats_new_count);
@@ -3965,7 +3965,7 @@ _get_stats(PyObject *self, PyObject *args) {
     PyDict_SetItemString(d, "blocks_cached", v ? v : Py_None);
     Py_XDECREF(v);
 
-    MUTEX_UNLOCK(&ImagingDefaultArena.mutex);
+    IMAGING_ARENA_UNLOCK(&ImagingDefaultArena.mutex);
     return d;
 }
 
@@ -3975,14 +3975,14 @@ _reset_stats(PyObject *self, PyObject *args) {
         return NULL;
     }
 
-    MUTEX_LOCK(&ImagingDefaultArena.mutex);
+    IMAGING_ARENA_LOCK(&ImagingDefaultArena.mutex);
     ImagingMemoryArena arena = &ImagingDefaultArena;
     arena->stats_new_count = 0;
     arena->stats_allocated_blocks = 0;
     arena->stats_reused_blocks = 0;
     arena->stats_reallocated_blocks = 0;
     arena->stats_freed_blocks = 0;
-    MUTEX_UNLOCK(&ImagingDefaultArena.mutex);
+    IMAGING_ARENA_UNLOCK(&ImagingDefaultArena.mutex);
 
     Py_INCREF(Py_None);
     return Py_None;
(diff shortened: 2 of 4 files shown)


aclark4life:
@kddnewton Can we say "environment" instead of "arena" here? Otherwise, thank you for the PR!

kddnewton:
@aclark4life No problem! Arenas in this context are memory arenas, which are already in use inside Pillow. The general idea is that they represent large contiguous blocks of memory, that you can then go and manually allocate memory from but avoid the cost of calling malloc/free. Below is a super simplified example:

```c
extern struct my_struct;

int main() {
  struct my_struct s1 = malloc(sizeof(struct my_struct));
  struct my_struct s2 = malloc(sizeof(struct my_struct));
  struct my_struct s3 = malloc(sizeof(struct my_struct));

  /* do something */

  free(s1);
  free(s2);
  free(s3);

  return EXIT_SUCCESS;
}
```

In this example we manually allocate memory for all three structs, and then manually free them. This can cause heap fragmentation and results in a lot of sys calls. Instead:

```c
extern struct my_struct;

struct my_arena {
  uint8_t *memory;
  size_t size;
};

void* my_malloc(struct my_arena *arena, size_t size) {
  void* result = arena->memory + arena->size + size;
  arena->size += size;
  return result;
}

int main() {
  struct my_arena arena = { .memory = malloc(1024), .size = 0 };

  struct my_struct s1 = my_malloc(&arena, sizeof(struct my_struct));
  struct my_struct s2 = my_malloc(&arena, sizeof(struct my_struct));
  struct my_struct s3 = my_malloc(&arena, sizeof(struct my_struct));

  /* do something */

  free(arena.memory);

  return EXIT_SUCCESS;
}
```

In this example we make a single memory allocation and then a single free, which means all of the memory is contiguous (helping with locality) and only 2 sys calls are made (more efficient). I'm omitting a couple of details here about bookkeeping, but that's the general gist.

This is already in place in Pillow. There is a single global arena that is used for all memory allocations. This is great, and helps a lot in terms of performance. However the downside is that in a multi-threaded environment, the mutex that wraps access to the arena (be it the GIL or a Python mutex in free-threaded Python) falls under a lot of contention because everyone is trying to use the same arena.

This commit instead makes a separate arena for each thread, so that each thread manages its own memory. This means it never falls under contention, and you can see in the benchmarks that it drastically speeds up free-threaded Python because it never has to lock anything.

I hope I explained that sufficiently, let me know if there's anything I can clear up!

hugovk:
cc @lysnikolaou who's been helping with the free-threaded work.

setup.py
```diff
@@ -1008,18 +1055,32 @@ def debug_build() -> bool:
...
+tls_define_macro = _try_compile_tls_define_macro(distutils.ccompiler.new_compiler())
+if tls_define_macro is not None:
+    define_macros.append(tls_define_macro)
+
 for src_file in _IMAGING:
     files.append("src/" + src_file + ".c")
 for src_file in _LIB_IMAGING:
     files.append(os.path.join("src/libImaging", src_file + ".c"))
 ext_modules = [
-    Extension("PIL._imaging", files),
```
lysnikolaou, line 1016: Turns out that inside `build_extensions` there's an `self._update_extension` call that updates the defines, libraries, sources etc. We should do that instead of calling `_try_compile_tls_define_macro` here again and then passing those in the `Extension` call.

The only problem is with `_imagingmath` and `_imagingmorph` for which there's no calls to `_update_extension`. But we could add those in with very minimal changes to the function signature (to allow for calling it without `libs`).

lysnikolaou (commented):
Looks great @kddnewton! Only suggested a change to `setup.py`, so that it's a bit clearer.

setup.py
kddnewton: Got it! Just updated it.

setup.py
```diff
@@ -890,7 +934,10 @@ def build_extensions(self) -> None:
 
         defs.append(("PILLOW_VERSION", f'"{PILLOW_VERSION}"'))
 
-        self._update_extension("PIL._imaging", libs, defs)
+        tls_define_macros = _try_compile_tls_define_macros(self.compiler)
+        self._update_extension("PIL._imaging", libs, defs + tls_define_macros)
+        self._update_extension("PIL._imagingmath", define_macros=tls_define_macros[:])
```
lysnikolaou, line 939: No need for a slice here or anywhere else, since `_update_extension` does not mutate the list.

kddnewton: Got it, updated!

setup.py
```diff
@@ -292,6 +294,47 @@ def _pkg_config(name: str) -> tuple[list[str], list[str]] | None:
...
 
+def _try_compile(compiler: distutils.ccompiler.CCompiler, code: str) -> bool:
+    try:
+        with tempfile.TemporaryDirectory() as d:
+            fn = os.path.join(d, "test.c")
+            with open(fn, "w") as f:
+                f.write(code)
+            compiler.compile([fn], output_dir=d, extra_preargs=["-Werror"])
+        return True
+    except Exception:
```
lysnikolaou, line 305: `setuptools.errors.CompileError` would work here, I think.

kddnewton: Yup, that worked.

kddnewton:
@lysnikolaou are those test failures related to my changes? Doesn't seem like it but since setup.py infects everything I'm not so sure.

radarhere:
The test failures should be fixed by https://github.com/python-pillow/Pillow/pull/8686
The docs failure has been fixed in main by https://github.com/python-pillow/Pillow/pull/8691

hugovk:
> The test failures should be fixed by #8686

Just merged, please update this PR from `main`.

setup.py
```diff
@@ -972,7 +1022,7 @@ def summary_report(self, feature: ext_feature) -> None:
             (feature.get("tiff"), "LIBTIFF"),
             (feature.get("freetype"), "FREETYPE2"),
             (feature.get("raqm"), "RAQM (Text shaping)", raqm_extra_info),
-            (feature.get("lcms"), "LITTLECMS2"),
```
wiredfool, line 975: This looks like a typo

kddnewton:
@hugovk done!

wiredfool:
The ImagingMemoryArena is an implicit default for the image -- it's not recorded anywhere that I see. What happens if an image is passed from thread to thread?

kddnewton:
@wiredfool do you have an example of passing it from thread to thread? I'm not sure if I know how that would happen.

wiredfool:
I've done it in the past where I had an app where all of the processing was offloaded to worker threads, using queuing. Scanner -> initial processing -> thumbnailing -> uploading were all done off the main thread. 

Anything were you're doing something with a UI main thread and processing elsewhere -- there are a bunch of operations that will create a new image.  If you then have a reference on the main thread you won't be able to release it. 

I'm also thinking that it's going to interfere with the lifetimes for arrow support, because that could potentially be freed from a thread that's not even part of our process.

aclark4life:
I was going to raise "does this help #1888?" so I'm curious too to know the answer … thanks all!

wiredfool:
I think something like:

1) Open image in parent thread
2) call image.resize in child thread  (e.g. threading.run())  
3) Return resized image to parent thread

would probably be enough to do it. 

Actually, looking at gcc's tls: 
> lifetimes that match the thread lifetime, and destructors that cleanup the unique per-thread storage

That concerns me on a couple of fronts -- 

The image memory is probably accessible outside the thread, since it's a malloc, and it's just the original struct that's going to be in the TLS storage. However, if we have a oneshot thread that passes the image off, the arena will be deallocated before the image is freed. 

So not only will the mutex and the arena likely be wrong in the child, there's going to be a pretty good memory leaks there because we won't necessarily get a chance to clean up the malloc'd items.

kddnewton:
@wiredfool Okay that helps a lot. I'll put together some example code and see what I can see. Maybe we'll need to add some logic on moving between threads to ensure everything is working properly. In the meantime let's put a pause on this PR until I can answer your questions.

wiredfool:
Ok, Some thoughts here --

1) I don't think that image storage tied to the life of the thread is a good idea. It breaks how we think about python objects. However failing tests due to that on this branch that pass on main should be added to main, because this is clearly an undertested corner. I suspect some of the tests might only fail under valgrind. 
2) Alternatively, It might be possible to reduce the scope of the locks, so that we're only locking things that actually modify the arena struct. e.g., we don't need to lock around the (re)malloc, only the addition into the block list. Reads are probably ok, and comparisons to mostly static ones like the blocks_max and block_size probably don't need locks. More fine grained locks might reduce contention. 
3) Or a set of 8 or 16 or n memory arenas and choose which of one them to use with a hash of some thread id. We'd need to store a pointer to the arena in the image struct though, and follow that for destruction. If the thread goes away, we're not actually losing any arena. There could potentially be arenas that don't get allocated to though, or allocated to and then never drained. The settings for the block cache are per arena, so there's the potential for n* the expected memory to be retained when images are freed.

ngoldbaum:
The original version of this PR I saw used pthread and windows key/value TLS. While it would require a pthread dependency on Unix, I wonder if that would avoid the problems with the `thread_local` keyword. Although that said I don't know if the storage ultimately being tied to a thread lifetime would still be an issue.

Thanks for pointing out that these arenas can be shared between threads! An important point. The uses of `thread_local` in NumPy are for places where NumPy used to use static scratch spaces local to a compilation unit as a performance optimization.

wiredfool:
Had some back of the mind thinking about this -- I think under TLS the mutex is essentially a no-op, because it's never going to have contention. How could it when the only things that lock it are on the same thread. So while it's not (IMO) a valid approach to speeding this up due to object lifetime, I think it's a good indication of the speedup that could happen from eliminating the bottleneck. 

If there were a static array of N memory arenas, and each thread took the next one in order, eg, (m++ % N) there would only need to be a single int stored in TLS. There's a bit of subtlety about pool sizes vs number of pools, but it's going to be a bit of a second order tuning thing. We'd definitely need to tag the image with it's source pool and read that on destroy.

API wise, I'm wondering if the Python TLS support would be a good place to look rather than native TLS, as it 's in the stable ABI. https://docs.python.org/3/c-api/init.html#thread-local-storage-support  I'm not sure if that's going to work on pypy though. 

- For single threaded programs, this would wind up just using a single arena like now. 
- For onshot threads, the allocations would rotate between pools.
- For n workers, you'd likely want to configure n pool threads so there's no contention, otherwise you'd get threads/pools contention.

kddnewton:
@wiredfool I like that approach a lot. Sorry I got caught up with other work stuff. I'll take a look at this first thing next week (starting Tuesday).

PR #8698, opened 2025-01-17
Use arm64 Linux runners for wheels

radarhere:
GitHub has announced that arm64 Linux runners are now free for public repositories - https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/

This means that we can use them, rather than QEMU, to build aarch64 wheels, greatly reducing the time that it takes to build all of our wheels.

While I'm here, I've also added 'pypy' to `CIBW_ENABLE`, since it will be turned off by default in cibuildwheel 3. You can see a warning about this at https://github.com/python-pillow/Pillow/actions/runs/12755221825 and read more at https://cibuildwheel.pypa.io/en/stable/options/#enable

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -105,7 +105,7 @@ jobs:
         env:
           CIBW_ARCHS: ${{ matrix.cibw_arch }}
           CIBW_BUILD: ${{ matrix.build }}
-          CIBW_ENABLE: cpython-prerelease cpython-freethreading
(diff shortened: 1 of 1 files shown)


hugovk:
Like https://github.com/python-pillow/docker-images/pull/229, also very nice!

This PR replaces these 8 jobs, taking ~2+ hours each:

image

With these two, taking ~25 mins each:

image

It reduces total duration (wall clock) from ~2h 34m to ~45m, and total run time (total CPU time) from ~21h 24m to ~4h 30m! 🚀

PR #8701, opened 2025-01-18
Only use outside border of stroke in text()

### LINKED ISSUE #8697: Stroke Text Tittles Not Working Correctly
### What did you do?

When trying to create a piece of text with a stroke fill the tittles on `i` and `j` do not look correct. 

I would expect that the stroke would fill all the way to the edge of the inner fill for the tittles.

### What actually happened?

The tittles on the `i` and `j` look like they have a thin outline around them rather than filling to the tittle.

* OS: Windows
* Python 3.11: 
* Pillow:  10.3.0

```text
Pillow 10.3.0
Python 3.11.5 | packaged by Anaconda, Inc. | (main, Sep 11 2023, 13:26:23) [MSC v.1916 64 bit (AMD64)]
Python executable is C:\Users\bradl\anaconda3\python.exe
System Python files loaded from C:\Users\bradl\anaconda3
Python Pillow modules loaded from C:\Users\USER\anaconda3\Lib\site-packages\PIL
Binary Pillow modules loaded from C:\Users\USER\anaconda3\Lib\site-packages\PIL
--- WEBPMUX support ok
--- WEBP Animation support ok
```

```python


# Create an image with a white background
width, height = 2000, 400
image = Image.new("RGB", (width, height), "white")

# Load a font
font_size = 100
font = ImageFont.load_default(font_size)

# Create a drawing context
draw = ImageDraw.Draw(image)

# Define text and positions
text = 'i j h o w'
text_position = (100, 100)  # Position of the text
stroke_width = 14  # Width of the stroke
stroke_fill = "black"  # Color of the stroke
text_fill = "blue"  # Color of the text

# Draw text with stroke
draw.text(
    text_position,
    text,
    font=font,
    fill=text_fill,
    stroke_width=stroke_width,
    stroke_fill=stroke_fill,
)

image.save("testing.png", dpi=(300, 300))
```

Image

radarhere:
Resolves #8697

The issue finds that when Pillow draws [normal text onto stroked text](https://github.com/python-pillow/Pillow/blob/a92a664ee5a3d5e3ee9b17bd6d6a2a3130394d75/src/PIL/ImageDraw.py#L691-L695), there might be an unexpected gap between the  the outside of the normal text and the inside border of the stroked text.

main

When drawing stroked text, FreeType returns a glyph with an outside border and an inside border. We can ask FreeType to leave out the inside border though, creating a filled glyph. With that change, the output is corrected.

branch

See https://freetype.org/freetype2/docs/reference/ft2-glyph_stroker.html#ft_glyph_stroke and https://freetype.org/freetype2/docs/reference/ft2-glyph_stroker.html#ft_glyph_strokeborder for documentation.

### DIFF
--- a/Tests/test_imagedraw.py
+++ b/Tests/test_imagedraw.py
@@ -1397,6 +1397,28 @@ def test_stroke_descender() -> None:
 
 
 @skip_unless_feature("freetype2")
+def test_stroke_inside_gap() -> None:
+    # Arrange
+    im = Image.new("RGB", (120, 130))
+    draw = ImageDraw.Draw(im)
+    font = ImageFont.truetype("Tests/fonts/FreeMono.ttf", 120)
+
+    # Act
+    draw.text((12, 12), "i", "#f00", font, stroke_width=20)
+
+    # Assert
+    for y in range(im.height):
+        glyph = ""
+        for x in range(im.width):
+            if im.getpixel((x, y)) == (0, 0, 0):
+                if glyph == "started":
+                    glyph = "ended"
+            else:
+                assert glyph != "ended", "Gap inside stroked glyph"
+                glyph = "started"
+
+
+@skip_unless_feature("freetype2")
 def test_split_word() -> None:
     # Arrange
     im = Image.new("RGB", (230, 55))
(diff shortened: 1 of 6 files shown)


apodtele:
Using only outside border is a good solution in this case.. You might try faster FT_Outline_EmboldenXY but it does not have round joints.

The algorithms are rather complicated. I suspect the problem with the inside border arises then the stroker radius exceeds the stem width or the dot radius. The inside border is ill-defined in this case. I doubt this can be fixed.

radarhere:
Thanks very much for your input.

PR #8707, opened 2025-01-20
Use monkeypatch

radarhere:


hugovk:
Shall we do a similar thing in `Tests/test_map.py`?

radarhere:
Sure, done.

PR #8330, opened 2024-08-25
Image -> Arrow support

wiredfool:
Partial implementation of #8329.

This is for discussion -- there are a number of issues.

Changes proposed in this pull request:

 * Implements basic `__arrow_c_array__` and `__arrow_c_schema__` support for reading Pillow images in Arrow capable libraries.
 * Implements basic arrow array exporting object -> Pillow image. This supports reading arrays in _exactly_ the same formats that we export from Pillow. 
 * Adds a way to force the internal allocator to use single blocks for images. Previously there is core.new_block, but internal operations call ImageNewDirty internally, which only calls the Arena Allocator.  
 * Adds C level read-only flag. Currently only exported to python layer, and overrides python read_only. 

Issues:
 * [x] Currently fails on large images. 
    - Returns a ValueError if the image is allocated in multiple blocks
 * [x] There's extraneous stuff in Arrow.c that should be cleaned up.
 * Some changes in Storage.c aren't required.
 * Unclear if we're exporting the most useful version of the data.

Todo:
* [x] Alternate binary installable arrow lib for testing, or loopback only testing on troublesome platforms if we can't find one that works.
* [x] Fix Windows Segfault. I don't have a windows env here, so either I need to figure out a CI based version or I'll need an assist. 
    - Windows tests now appear to be passing on 3.10->3.13, with 3.9, pypy3.10 and 3.14 failing on arrow install. 
* [ ] ~Python accessible version (r/w, likely) of the schema capsule and possibly array metadata. Partially for testing, partially for sanity checking at the python layer before passing into the C portion.~
* [x] Fix the UNDONEs in the code, especially around error handling. 
  - [x] make sure we're raising an error if we don't accept the array format. 
* [x] Large image support _or_ error out. 
* [x] Wire up enforcement of c level read_only flag
* [x] Docs

For Review:
* What CVE is in here that I'm not seeing?
* There's some pretty serious mucking with the object lifetimes, specifically there's now effectively a refcount on the core C image, so it's lifetime is not limited to the python object's lifetime. I think this is required, but it needs review. 
* I'm not 100% on how so much of the checking is in Storage.c, vs somewhere closer to the surface. 
* The core Imaging item now stores a reference to the array pycapsule, which requires PyINCREF and PyDECREF in Storage.c. This feels like it's finally breaking the wall of libImaging not relying on python headers.

PR #8714, opened 2025-01-24
Reduce memory arena contention

kddnewton:
This is a follow-up to https://github.com/python-pillow/Pillow/pull/8692, based on @wiredfool's feedback.

Previously there was one memory arena for all threads, making it the bottleneck for multi-threaded performance. As the number of threads increased, the contention for the lock on the arena would grow, causing other threads to wait to acquire it.

This commit makes it use 8 memory arenas, and round-robbins how they are assigned to threads. Threads keep track of the index that they should use into the arena array, assigned the first time the arena is accessed on a given thread.

When an image is first created, it is allocated from an arena. When the logic to have multiple arenas is enabled, it then keeps track of the index on the image, so that when deleted it can be returned to the correct arena.

Effectively this means that in single-threaded programs, this should not really have an effect. We also do not do this logic if the GIL is enabled, as it effectively acts as the lock on the default arena for us.

As expected, this approach has no real noticable effect on regular CPython. On free-threaded CPython, however, there is a massive difference (measuring up to about 70%).

Here is the benchmarking script that I used:

<details>
<summary>test.py</summary>

```python
import concurrent.futures
import os
import threading
import time

from PIL import Image

num_threads = 16
num_images = 1024


def operation():
    images = []
    for i in range(num_images):
        img = Image.new(
            "RGB", (100, 100), color=(i % 256, (i // 256) % 256, (i // 65536) % 256)
        )
        images.append(img)

    for img in images:
        img = img.convert("CMYK")

    images.clear()


def worker(barrier):
    barrier.wait()
    runtimes = []

    for _ in range(5):
        start_time = time.time()
        operation()
        end_time = time.time()
        runtimes.append(end_time - start_time)

    return runtimes


def benchmark():
    with concurrent.futures.ThreadPoolExecutor(max_workers=num_threads) as executor:
        barrier = threading.Barrier(num_threads)
        futures = [executor.submit(worker, barrier) for _ in range(num_threads)]

        run_times = []
        for future in concurrent.futures.as_completed(futures):
            try:
                run_times.extend(future.result())
            except IndexError:
                os._exit(-1)

        min_time = min(run_times)
        max_time = max(run_times)
        mean_time = sum(run_times) / len(run_times)
        print(f"Max: {max_time:.6f} Mean: {mean_time:.6f} Min: {min_time:.6f}")


benchmark()
```
</details>

### 3.13.0 on main

```
$ python -VV                                        
Python 3.13.0 (tags/v3.13.0:60403a5409f, Jan  3 2025, 14:04:52) [Clang 16.0.0 (clang-1600.0.26.6)]
$ python test.py
Max: 0.404962 Mean: 0.353120 Min: 0.303807
$ python test.py
Max: 0.369188 Mean: 0.320218 Min: 0.282613
$ python test.py
Max: 0.386692 Mean: 0.335509 Min: 0.294476
$ python test.py
Max: 0.394410 Mean: 0.350275 Min: 0.299456
$ python test.py
Max: 0.416075 Mean: 0.354347 Min: 0.309045
```

### 3.13.0 on branch

```
$ python -VV
Python 3.13.0 (tags/v3.13.0:60403a5409f, Jan  3 2025, 14:04:52) [Clang 16.0.0 (clang-1600.0.26.6)]
$ python test.py
Max: 0.422371 Mean: 0.354453 Min: 0.292521
$ python test.py
Max: 0.423698 Mean: 0.358393 Min: 0.313581
$ python test.py
Max: 0.405487 Mean: 0.356346 Min: 0.299354
$ python test.py
Max: 0.431244 Mean: 0.369772 Min: 0.308096
$ python test.py
Max: 0.472806 Mean: 0.377575 Min: 0.313588
```

### 3.13.0t on main

```
$ python -VV
Python 3.13.0 experimental free-threading build (tags/v3.13.0:60403a5409f, Jan  7 2025, 13:53:44) [Clang 16.0.0 (clang-1600.0.26.6)]
$ python test.py
Max: 0.161379 Mean: 0.114555 Min: 0.072739
$ python test.py
Max: 0.188203 Mean: 0.133376 Min: 0.095111
$ python test.py
Max: 0.181084 Mean: 0.128733 Min: 0.086086
$ python test.py
Max: 0.187286 Mean: 0.131114 Min: 0.094561
$ python test.py
Max: 0.191979 Mean: 0.133439 Min: 0.097527
```

### 3.13.0t on branch

```
$ python -VV  
Python 3.13.0 experimental free-threading build (tags/v3.13.0:60403a5409f, Jan  7 2025, 13:53:44) [Clang 16.0.0 (clang-1600.0.26.6)]
$ python test.py
Max: 0.090055 Mean: 0.044987 Min: 0.019220
$ python test.py
Max: 0.095986 Mean: 0.040712 Min: 0.019204
$ python test.py
Max: 0.094424 Mean: 0.041949 Min: 0.016574
$ python test.py
Max: 0.084610 Mean: 0.042789 Min: 0.016866
$ python test.py
Max: 0.095480 Mean: 0.044068 Min: 0.015999
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,18 +8,21 @@
 # ------------------------------
 from __future__ import annotations
 
+import distutils.ccompiler
 import os
 import re
 import shutil
 import struct
 import subprocess
 import sys
+import tempfile
 import warnings
 from collections.abc import Iterator
 from typing import Any
 
 from setuptools import Extension, setup
 from setuptools.command.build_ext import build_ext
+from setuptools.errors import CompileError
 
 
 def get_version() -> str:
@@ -292,6 +295,47 @@ def _pkg_config(name: str) -> tuple[list[str], list[str]] | None:
     return None
 
 
+def _try_compile(compiler: distutils.ccompiler.CCompiler, code: str) -> bool:
+    try:
+        with tempfile.TemporaryDirectory() as d:
+            fn = os.path.join(d, "test.c")
+            with open(fn, "w") as f:
+                f.write(code)
+            compiler.compile([fn], output_dir=d, extra_preargs=["-Werror"])
+        return True
+    except CompileError:
+        return False
+
+
+def _try_compile_attr(compiler: distutils.ccompiler.CCompiler, attr: str) -> bool:
+    code = f"""
+        #pragma GCC diagnostic error "-Wattributes"
+        #pragma clang diagnostic error "-Wattributes"
+
+        int {attr} foo;
+        int main() {{
+            return 0;
+        }}
+    """
+
+    return _try_compile(compiler, code)
+
+
+def _try_compile_tls_define_macros(
+    compiler: distutils.ccompiler.CCompiler,
+) -> list[tuple[str, str | None]]:
+    if _try_compile_attr(compiler, "thread_local"):  # C23
+        return [("HAVE_THREAD_LOCAL", None)]
+    elif _try_compile_attr(compiler, "_Thread_local"):  # C11/C17
+        return [("HAVE__THREAD_LOCAL", None)]
+    elif _try_compile_attr(compiler, "__thread"):  # GCC/clang
+        return [("HAVE___THREAD", None)]
+    elif _try_compile_attr(compiler, "__declspec(thread)"):  # MSVC
+        return [("HAVE___DECLSPEC_THREAD_", None)]
+    else:
+        return []
+
+
 class pil_build_ext(build_ext):
     class ext_feature:
         features = [
@@ -426,13 +470,14 @@ class pil_build_ext(build_ext):
     def _update_extension(
         self,
         name: str,
-        libraries: list[str] | list[str | bool | None],
+        libraries: list[str] | list[str | bool | None] | None = None,
         define_macros: list[tuple[str, str | None]] | None = None,
         sources: list[str] | None = None,
     ) -> None:
         for extension in self.extensions:
             if extension.name == name:
-                extension.libraries += libraries
+                if libraries is not None:
+                    extension.libraries += libraries
                 if define_macros is not None:
                     extension.define_macros += define_macros
                 if sources is not None:
@@ -890,7 +935,10 @@ class pil_build_ext(build_ext):
 
         defs.append(("PILLOW_VERSION", f'"{PILLOW_VERSION}"'))
 
-        self._update_extension("PIL._imaging", libs, defs)
+        tls_define_macros = _try_compile_tls_define_macros(self.compiler)
+        self._update_extension("PIL._imaging", libs, defs + tls_define_macros)
+        self._update_extension("PIL._imagingmath", define_macros=tls_define_macros)
+        self._update_extension("PIL._imagingmorph", define_macros=tls_define_macros)
 
         #
         # additional libraries
@@ -913,7 +961,9 @@ class pil_build_ext(build_ext):
                         libs.append(feature.get("fribidi"))
                     else:  # building FriBiDi shim from src/thirdparty
                         srcs.append("src/thirdparty/fribidi-shim/fribidi.c")
-            self._update_extension("PIL._imagingft", libs, defs, srcs)
+            self._update_extension(
+                "PIL._imagingft", libs, defs + tls_define_macros, srcs
+            )
 
         else:
             self._remove_extension("PIL._imagingft")
@@ -922,19 +972,19 @@ class pil_build_ext(build_ext):
             libs = [feature.get("lcms")]
             if sys.platform == "win32":
                 libs.extend(["user32", "gdi32"])
-            self._update_extension("PIL._imagingcms", libs)
+            self._update_extension("PIL._imagingcms", libs, tls_define_macros)
         else:
             self._remove_extension("PIL._imagingcms")
 
         webp = feature.get("webp")
         if isinstance(webp, str):
             libs = [webp, webp + "mux", webp + "demux"]
-            self._update_extension("PIL._webp", libs)
+            self._update_extension("PIL._webp", libs, tls_define_macros)
         else:
             self._remove_extension("PIL._webp")
 
         tk_libs = ["psapi"] if sys.platform in ("win32", "cygwin") else []
-        self._update_extension("PIL._imagingtk", tk_libs)
+        self._update_extension("PIL._imagingtk", tk_libs, tls_define_macros)
 
         build_ext.build_extensions(self)
 
--- a/src/_imaging.c
+++ b/src/_imaging.c
@@ -3938,34 +3938,50 @@ _get_stats(PyObject *self, PyObject *args) {
         return NULL;
     }
 
-    MUTEX_LOCK(&ImagingDefaultArena.mutex);
-    ImagingMemoryArena arena = &ImagingDefaultArena;
+    long stats_new_count = 0;
+    long stats_allocated_blocks = 0;
+    long stats_reused_blocks = 0;
+    long stats_reallocated_blocks = 0;
+    long stats_freed_blocks = 0;
+    long blocks_cached = 0;
 
-    v = PyLong_FromLong(arena->stats_new_count);
+    uint8_t index;
+    ImagingMemoryArena arena;
+    IMAGING_ARENAS_FOREACH(index, arena) {
+        MUTEX_LOCK(&arena->mutex);
+        stats_new_count += arena->stats_new_count;
+        stats_allocated_blocks += arena->stats_allocated_blocks;
+        stats_reused_blocks += arena->stats_reused_blocks;
+        stats_reallocated_blocks += arena->stats_reallocated_blocks;
+        stats_freed_blocks += arena->stats_freed_blocks;
+        blocks_cached += arena->blocks_cached;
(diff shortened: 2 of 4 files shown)


ngoldbaum:
I wonder what happens if you plot the result of your benchmark as a function of thread count. See [e.g. this NumPy issue](https://github.com/numpy/numpy/issues/27786) which reported a similar scaling issue and running a benchmark as a function of thread count was a very useful way to identify the scaling issue and that it was fixed by using a locking that scales better.

src/libImaging/Storage.c
```diff
@@ -258,23 +259,106 @@ ImagingDelete(Imaging im) {
...
+
+/* Get a pointer to the correct arena for this context. In this case where we
+ * are using a round-robin approach to the statically allocated arenas, we will
+ * return the arena that is assigned to the thread on first use.
+ */
+ImagingMemoryArena
+ImagingGetArena(void) {
+    if (ImagingArenaThreadIndex == UINT64_MAX) {
+        ImagingArenaThreadIndex =
+            _Py_atomic_add_uint64(&ImagingArenaIndex, 1) % IMAGING_ARENAS_COUNT;
```
lysnikolaou, line 295: This is a bit confusing. As far as I remember, `_Py_atomic_add_*` functions return the old value of the atomic variable, not the new one, so the first iteration on this would actually do `UINT64_MAX % IMAGING_ARENAS_COUNT`. That's probably not what's intended here, right?

src/libImaging/Storage.c
```diff
@@ -258,23 +259,106 @@ ImagingDelete(Imaging im) {
...
+#ifdef IMAGING_TLS
+/* This is the overall process-level index that keeps track of the next index
+ * that will be assigned to a thread.
+ */
+static uint64_t ImagingArenaIndex = UINT64_MAX;
+
+/* This is the thread-local index that associated a thread with an arena in the
+ * statically-allocated list.
+ */
+static IMAGING_TLS uint64_t ImagingArenaThreadIndex = UINT64_MAX;
```
lysnikolaou, line 273: A comment that the code below relies on `UINT64_MAX + 1` overflowing to 0 would probably be a good idea here.

src/_imaging.c
```diff
@@ -4087,15 +4117,21 @@ _set_blocks_max(PyObject *self, PyObject *args) {
...
-    MUTEX_LOCK(&ImagingDefaultArena.mutex);
-    int status = ImagingMemorySetBlocksMax(&ImagingDefaultArena, blocks_max);
-    MUTEX_UNLOCK(&ImagingDefaultArena.mutex);
-    if (!status) {
+    int status = 0;
+    uint8_t index;
+    ImagingMemoryArena arena;
+    IMAGING_ARENAS_FOREACH(index, arena) {
+        MUTEX_LOCK(&arena->mutex);
+        status |= ImagingMemorySetBlocksMax(arena, blocks_max);
```
lysnikolaou, line 4130: Maybe change the name of the variable to something like `error` here, since we've changed the return value?

src/libImaging/Storage.c
```diff
@@ -258,23 +259,106 @@ ImagingDelete(Imaging im) {
...
+struct ImagingMemoryArena ImagingArenas[IMAGING_ARENAS_COUNT] = {
+    {1, IMAGING_ARENA_BLOCK_SIZE, 0, 0, NULL, 0, 0, 0, 0, 0, 0, {0}},
+    {1, IMAGING_ARENA_BLOCK_SIZE, 0, 0, NULL, 0, 0, 0, 0, 0, 1, {0}},
+    {1, IMAGING_ARENA_BLOCK_SIZE, 0, 0, NULL, 0, 0, 0, 0, 0, 2, {0}},
+    {1, IMAGING_ARENA_BLOCK_SIZE, 0, 0, NULL, 0, 0, 0, 0, 0, 3, {0}},
+    {1, IMAGING_ARENA_BLOCK_SIZE, 0, 0, NULL, 0, 0, 0, 0, 0, 4, {0}},
+    {1, IMAGING_ARENA_BLOCK_SIZE, 0, 0, NULL, 0, 0, 0, 0, 0, 5, {0}},
+    {1, IMAGING_ARENA_BLOCK_SIZE, 0, 0, NULL, 0, 0, 0, 0, 0, 6, {0}},
+    {1, IMAGING_ARENA_BLOCK_SIZE, 0, 0, NULL, 0, 0, 0, 0, 0, 7, {0}}
+};
```
lysnikolaou, line 285: We could maybe NULL-terminal this array here, so that we can avoid having to define an index every time we're doing a `FOREACH`.

lysnikolaou (commented):
Great job! The approach looks good to me, though I've left some comments on some specifics.

src/libImaging/Storage.c
kddnewton: Ahh, I misunderstood the function. Yes this is wrong, I will fix it!

src/libImaging/Storage.c
kddnewton: No longer necessary since we can initialize `ImagingArenaIndex` to `0` now.

src/_imaging.c
kddnewton: 👍

kddnewton:
@lysnikolaou — Thanks for the review! I've applied your suggestions. Please let me know if you see anything else.
@ngoldbaum — I'll work on a graph. I chose 8 arenas after some experimentation, but it would be good to back this up.

lysnikolaou:
@kddnewton Could you please not rebase/force-push after the first round of reviews? It makes follow-up reviews a bit harder.

kddnewton:
@lysnikolaou ahh sorry! I saw it was out of date so just wanted to keep it synced. I'll avoid that going forward.

kddnewton:
@ngoldbaum here's the chart. Sorry I didn't end up getting it working with mathplotlib, so it's just manually putting numbers into google sheets. But the numbers come straight from the benchmarking script linked in the description.

Screenshot 2025-01-31 at 12 20 11 PM

You can see from the graph it gets really bad when you start to have a lot of threads on `main` with free threading. This branch and `main` are about the same when running without free threading. On this branch with free threading it is significantly better.

lysnikolaou:
There's a segfault on Ubuntu that might need some attention here.

kddnewton:
Yes and it appears that the other one is hanging, I will take a quick look. It was passing before, so I'm imagining this is related to my changes around looping through the arenas.

lysnikolaou:
@kddnewton Are you going to have a look at the failures here? If not, I can also spend some time on it.

wiredfool:
I'm actually curious if we're really getting any benefit from the Arena memory allocator -- since at least in the default case, we're not actually retaining any of the memory for reallocation. We might just be better off using the simpler block allocator. On the other hand, it may work better if we actually retain some of the freed blocks. 

There's a patch  (https://github.com/python-pillow/Pillow/pull/8330/commits/240175733e2824122427a7d401d52abfbeb96c91) in my arrow branch that enables the block allocator for all operations (instead of just being used for ImageTk as it is on main).

kddnewton:
Ahh apologies @lysnikolaou — I went on paternity leave a little earlier than expected, so I am out at the moment. I believe @SonicField may be able to look at this soon, but if you have time that would be great. I _think_ the issue is how we're looping through the arenas at the moment since it changed from the initial version of the PR (I see what you mean about not force-pushing!).

Worst-case I'll pick this up when I return at the end of March.

lysnikolaou:
Oh, unfortunately, it seems like I don't have the necessary permissions to push to this branch.

kddnewton:
@lysnikolaou just sent you an invite

setup.py
```diff
@@ -14,12 +14,17 @@
...
 from collections.abc import Iterator
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
 from setuptools import Extension, setup
 from setuptools.command.build_ext import build_ext
+from setuptools.errors import CompileError
+
+if TYPE_CHECKING:
+    import distutils.ccompiler
```
hugovk, line 27: Let's avoid some type-checking only imports:
```suggestion

from setuptools import Extension, setup
from setuptools.command.build_ext import build_ext
from setuptools.errors import CompileError

TYPE_CHECKING = False
if TYPE_CHECKING:
    import distutils.ccompiler
    from collections.abc import Iterator
    from typing import Any
```

setup.py
```diff
@@ -14,12 +14,17 @@
 import struct
 import subprocess
 import sys
+import tempfile
 import warnings
 from collections.abc import Iterator
```
hugovk, line 19:
```suggestion
```

hugovk:
Do we want this for tomorrow's release, assuming the the Arrow PR (https://github.com/python-pillow/Pillow/pull/8330) is merged?

wiredfool:
I think this should be compatible with the arrow pr. I'm still curious if the simple block allocator (vs. the arena) is a simpler way to reduce the contention on the memory arena, because if it is, it's possible that the memory arena allocator isn't really an advantage anymore.

wiredfool:
I don't think there's anything particularly strongly tying this to the Arrow PR, other than the Arrow PR has a way to prevent using memory arenas at all (bypassing this), and requires that for larger images to be exported via error.

PR #8715, opened 2025-01-25
Improved type hints for ImageDraw and ImageFont

AsfhtgkDavid:
Fixes #typing

 * Changed some str type to Literal for more specificity


radarhere (approved):
Approved, but I'd like to know what @hugovk thinks of `Anchor`, `Align` and `Direction` as well.

I'm not sure if a list of 24 `Anchor` values is helpful or not. I would imagine either you already know what value you want, or you need to go and look at https://pillow.readthedocs.io/en/stable/handbook/text-anchors.html.

AsfhtgkDavid:
I think this type hinting is in line with The zen of python's principle ‘Explicit is better than implicit’.

hugovk:
How disruptive do you think this will be?

Does this mean people using the old string versions will start getting potentially lots of type hint errors, and will need to switch to the new literals or add `type: ignore[arg-type]`?

Is there any performance impact?

hugovk:
> ... but
> 
> ```python
> anchor = "la"
> font.getbbox("test", anchor=anchor)
> ```
> 
> would raise a type hint error, yes.

I'd prefer to avoid this.

Issue #8750, opened 2025-02-14
Segmentation Fault when decoding PNG on macOS 10.14


arisolt:
### What did you do?
I've come across this bug while using `pytautogui.screenshot()`. By debugging it, I've narrowed it down to this call: `decoder.decode(b)` inside `Image.load()`. This call runs successfully when the image is a JPEG format, but not when it's PNG (the latter is the default format used by the screenshot utility on macOS).

For now, I've resolved the problem by downgrading Pillow to 7.2, which was the last version tested on macOS 10.14 according to https://pillow.readthedocs.io/en/latest/installation/platform-support.html

### What actually happened?
Python crashes with `Segmentation fault: 11` on this line inside `Image.load()`:
```
n, err_code = decoder.decode(b)
```
Jumping to the definition of `decode()` shows that the only instruction is to raise "NotImplementedError", with the message "unvailable in base decoder".

### What are your OS, Python and Pillow versions?

* OS: macOS 10.14.6
* Python: 3.11.4
* Pillow: 11.1.0

```text
--------------------------------------------------------------------
Pillow 11.1.0
Python 3.11.4 (main, Jun 23 2023, 05:06:34) [Clang 10.0.1 (clang-1001.0.46.4)]
--------------------------------------------------------------------
Python executable is /usr/local/opt/python@3.11/bin/python3.11
System Python files loaded from /usr/local/opt/python@3.11/Frameworks/Python.framework/Versions/3.11
--------------------------------------------------------------------
Python Pillow modules loaded from /usr/local/lib/python3.11/site-packages/PIL
Binary Pillow modules loaded from /usr/local/lib/python3.11/site-packages/PIL
--------------------------------------------------------------------
--- PIL CORE support ok, compiled for 11.1.0
*** TKINTER support not installed
--- FREETYPE2 support ok, loaded 2.13.2
--- LITTLECMS2 support ok, loaded 2.16
--- WEBP support ok, loaded 1.5.0
--- JPEG support ok, compiled for libjpeg-turbo 3.1.0
--- OPENJPEG (JPEG2000) support ok, loaded 2.5.3
--- ZLIB (PNG/ZIP) support ok, loaded 1.3.1.zlib-ng, compiled for zlib-ng 2.2.2
--- LIBTIFF support ok, loaded 4.6.0
--- RAQM (Bidirectional Text) support ok, loaded 0.10.1, fribidi 1.0.11, harfbuzz 10.1.0
[...]

PR #8772, opened 2025-02-24
Revert to zlib on macOS < 10.15

radarhere:
#8500 switched to zlib-ng in our wheels. However, #8750 has reported that wheels now segfault on macOS 10.13 and 10.14.

https://github.com/zlib-ng/zlib-ng/issues/1877 was created to investigate this, and there is now a proposed fix in https://github.com/zlib-ng/zlib-ng/pull/1878.

However, in case it is not merged and released before Pillow 11.2.0, this PR temporarily reverts to using zlib on macOS < 10.15.


radarhere:
zlib-ng 2.2.5 has been released with the fix, so https://github.com/python-pillow/Pillow/pull/9140 has been created to revert this.

PR #1470, opened 2015-10-05
Warn that tobytes is not for compressed image data

masklinn:
Every time I try to get in-memory PNG or JPEG data I tend to spend half an hour fighting with tobytes until I remember it is not going to work and I'm supposed to use `im.save(fp, format='png')`.

#985 makes me think I'm not alone, and a clear warning in the documentation can only help others (and future-me).

I don't actually know what tobytes is for, so I went with something along the lines of @wiredfool's comment in #985, they note that it's

>  only used for a couple very close to raw formats

PR #8777, opened 2025-02-27
Remove mention that 1-bit images use 1 byte per pixel

hchargois:
> (1-bit pixels, black and white, stored with one pixel per byte)

That line in the documentation implies that a 1-bit (W, H)-sized image would occupy W*H bytes in raw form, at least that's how I understand it.

But it doesn't seem to be the case in Pillow 11.1.0:

```pycon
>>> Image.new("1", (3, 2), 1).tobytes()
b'\xe0\xe0'
```

It seems that it now stores strips of (up to) 8 horizontal pixels. Since that line in the documentation was last edited 12 years ago, I guess it was true at the time but the format has been changed for efficiency since. Which is great, but the documentation is misleading.

I suggest to simply remove the 1-byte-per-pixel part.

### DIFF
--- a/docs/handbook/concepts.rst
+++ b/docs/handbook/concepts.rst
@@ -30,7 +30,7 @@ image. Each pixel uses the full range of the bit depth. So a 1-bit pixel has a r
 INT32 and a 32-bit floating point pixel has the range of FLOAT32. The current release
 supports the following standard modes:
 
-    * ``1`` (1-bit pixels, black and white, stored with one pixel per byte)
+    * ``1`` (1-bit pixels, black and white)
     * ``L`` (8-bit pixels, grayscale)
     * ``P`` (8-bit pixels, mapped to any other mode using a color palette)
     * ``RGB`` (3x8-bit pixels, true color)


wiredfool:
Internally, storage for 1 bit images is 1 byte per pixel: https://github.com/python-pillow/Pillow/blob/main/src/libImaging/Storage.c#L64`

```
    if (strcmp(mode, "1") == 0) {
        /* 1-bit images */
        im->bands = im->pixelsize = 1;
        im->linesize = xsize;
    }
```

1 bit images are shuffled at import/export to put them in this format.

hchargois:
Interesting! I didn't expect `tobytes` to return something other than how the data is stored internally. The documentation of `tobytes` does say so:

> This method returns the raw image data from the internal storage

That documentation also says to look at `_imaging.c` so I followed the white rabbit up to `Pack.c` and found out that even the raw encoder can actually use multiple packing functions, that's super cool! I've seen cases where custom serialization of individual channels or postprocessing of `tobytes` was done to encode e.g. pixels in BGR format, while `tobytes("raw", "BGR")` would have been enough and much more efficient. I guess documenting all of these special packers would be daunting; but pointing at `Pack.c` and giving a couple examples could be very useful.

And in any case, I still think the two sentences (the "one pixel per byte" and the "tobytes returns raw data") are very misleading when taken together. I would still suggest to remove the first sentence. I think it's better to not give any expectations than provide false expectations.

wiredfool:
Yeah, `tobytes` has a lot behind it because of the packer infrastructure. "Returns raw image data" is mostly true, at least for most modes in the default raw configuration. It's more true than running `image.save` into a `BytesIO` object. 

I suspect that the 1 bit thing is just one of those historical choices -- it's a lot easier to do math in 1 pixel per byte than one per bit, especially when you already have that math implemented for other modes. Historically it's tripped up people with the array/numpy integration, and this is effectively the one way you can get 1 bit/pixel out of a mode `1` image.

radarhere:
> This method returns the raw image data from the internal storage.

The origin of this sentence was in #1470, where one user wanted to clarify that `tobytes()` didn’t return data in a regular image format. It’s unfortunate that one clarification has lead to different confusion.

i think ‘raw image data' is intended to mean one-dimensional pixel data without context, and ‘from the internal storage’ is intended to mean ‘this isn’t the form the data had in any given file’

> I think it's better to not give any expectations than provide false expectations.

As a general principle, I'm in favour of clarifying information. I'm not in favour of removing information because it might be misunderstood.

I think 'stored with one pixel per byte' is meant to communicate that Pillow will use more memory than you might expect.

This is kind of a specific situation to document - not so much because of the 1 mode, but because what you’re talking about just the “raw” encoder, and `tobytes()` can be used with other encoders. If you used any other encoder, I think it would be clearer what you received wasn't exactly what was stored internally in Pillow.

And interestingly, when this specific situation does become relevant to Pillow's internal code, there is a comment that tries to explain your scenario.
https://github.com/python-pillow/Pillow/blob/c8d98d56a02e0729f794546d6f270b3cea5baecf/src/PIL/Image.py#L730-L735

I've created #8878 as an alternative - it leaves in the information about how mode 1 image data is stored, mentions that `tobytes()` returns data **derived** from internal storage, and adds a reference to Pack.c.

hchargois:
Great, thank you. I'll close this PR in favor of the new one.

Issue #8775, opened 2025-02-25
How to write a multipage TIFF with different ImageFileDirectories for each frame?


anntzer:
### What did you do?

I would like to write a multipage tiff file with different ImageFileDirectories for each frame.  I started with the example given at https://github.com/python-pillow/Pillow/issues/3636#issuecomment-508066387 (#3636 being essentially the same issue).

### What did you expect to happen?

The example code should print 1 2 3 4.

### What actually happened?

The code prints 1 2 3 4, then crashes with
```pytb
Exception ignored in: <PIL.TiffImagePlugin.AppendingTiffWriter object at 0x101bb2930>
Traceback (most recent call last):
  File "/path/to/lib/python3.13/site-packages/PIL/TiffImagePlugin.py", line 2197, in close
  File "/path/to/lib/python3.13/site-packages/PIL/TiffImagePlugin.py", line 2068, in finalize
ValueError: seek of closed file
```

### What are your OS, Python and Pillow versions?

```text
--------------------------------------------------------------------
Pillow 11.1.0
Python 3.13.0 | packaged by conda-forge | (main, Oct 17 2024, 12:32:35) [Clang 17.0.6 ]
--------------------------------------------------------------------
Python executable is /path/to/miniforge3/envs/default/bin/python
System Python files loaded from /path/to/miniforge3/envs/default
--------------------------------------------------------------------
Python Pillow modules loaded from /path/to/.local/lib/python3.13/site-packages/PIL
Binary Pillow modules loaded from /path/to/.local/lib/python3.13/site-packages/PIL
--------------------------------------------------------------------
--- PIL CORE support ok, compiled for 11.1.0
--- TKINTER support ok, loaded 8.6
--- FREETYPE2 support ok, loaded 2.13.2
--- LITTLECMS2 support ok, loaded 2.16
--- WEBP support ok, loaded 1.5.0
--- JPEG support ok, compiled for libjpeg-turbo 3.1.0
--- OPENJPEG (JPEG2000) support ok, loaded 2.5.3
--- ZLIB (PNG/ZIP) support ok, loaded 1.3.1.zlib-ng, compiled for zlib-ng 2.2.2
--- LIBTIFF support ok, loaded 4.6.0
*** RAQM (Bidirectional Text) support not installed
*** LIBIMAGEQUANT (Quantization method) support not installed
--- XCB (X protocol) support ok
--------------------------------------------------------------------
```

A few notes:
[...]

PR #8781, opened 2025-03-01
If append_images is populated, default save_all to True

radarhere:
https://github.com/python-pillow/Pillow/issues/8775#issuecomment-2691281777 requests that `save_all` default to true if `append_images` is not empty.

So rather than requiring `save_all` in the tutorial example,
```python
from PIL import Image

# List of image filenames
image_filenames = [
    "hopper.jpg",
    "rotated_hopper_270.jpg",
    "rotated_hopper_180.jpg",
    "rotated_hopper_90.jpg",
]

# Open images and create a list
images = [Image.open(filename) for filename in image_filenames]

# Save the images as an animated GIF
images[0].save(
    "animated_hopper.gif",
    save_all=True,
    append_images=images[1:],
    duration=500,  # duration of each frame in milliseconds
    loop=0,  # loop forever
)
```
users can just
```python
images[0].save(
    "animated_hopper.gif",
    append_images=images[1:],
    duration=500,  # duration of each frame in milliseconds
    loop=0,  # loop forever
)
```


hugovk (approved):
Please could you resolve the conflict?

PR #8782, opened 2025-03-01
Removed ImageTk._show

radarhere:
https://github.com/python-pillow/Pillow/blob/d7d48df9af3b6177a62759cb00db391196f91f39/src/PIL/ImageTk.py#L268-L269

This helper is actually unused, so I suggest deprecating it.


hugovk:
It's underscore-prefixed, so shall we just remove it?

radarhere:
Ok, I've updated the commit.

PR #8783, opened 2025-03-01
Improved BdfFontFile test coverage

radarhere:
Adds coverage for the lines missing at https://app.codecov.io/gh/python-pillow/Pillow/blob/main/src%2FPIL%2FBdfFontFile.py

### DIFF
--- a/Tests/test_font_bdf.py
+++ b/Tests/test_font_bdf.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import io
+
 import pytest
 
 from PIL import BdfFontFile, FontFile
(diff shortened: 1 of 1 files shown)


Tests/test_font_bdf.py
```diff
@@ -8,13 +10,20 @@
...
-    with open(filename, "rb") as test_file:
-        font = BdfFontFile.BdfFontFile(test_file)
+    with open(filename, "rb") as fp:
+        font = BdfFontFile.BdfFontFile(fp)
 
     assert isinstance(font, FontFile.FontFile)
     assert len([_f for _f in font.glyph if _f]) == 190
 
 
+def test_valueerror() -> None:
```
hugovk, line 20: More descriptive name, as the test doesn't directly deal with the `ValueError`:
```suggestion
def test_zero_width_chars() -> None:
```

hugovk (approved):

PR #8799, opened 2025-03-04
Call `startswith`/`endswith` once with a tuple, and `range` without `start=0`

hugovk:
Some small cleanups found with https://docs.astral.sh/ruff/rules/#flake8-pie-pie.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -121,6 +121,7 @@ lint.select = [
   "ISC",    # flake8-implicit-str-concat
   "LOG",    # flake8-logging
   "PGH",    # pygrep-hooks
+  "PIE",    # flake8-pie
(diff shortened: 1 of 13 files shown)


pyproject.toml
```diff
@@ -133,6 +134,7 @@ lint.ignore = [
   "E221",   # Multiple spaces before operator
   "E226",   # Missing whitespace around arithmetic operator
   "E241",   # Multiple spaces after ','
+  "PIE790", # flake8-pie: unnecessary-placeholder
```
radarhere, line 137: Just for the record, you're ignoring this because you dislike the rule? I mean, we could adjust to it. https://github.com/hugovk/Pillow/pull/130

Issue #1604, opened 2015-12-21
Resizing may result in degraded quality because gamma is not compensated


znerol:
Citing from http://www.4p8.com/eric.brasseur/gamma.html

> Technically speaking, the problem is that "the computations are performed as if the scale of brightnesses was linear while in fact it is a power scale." In mathematical terms: "a gamma of 1.0 is assumed while it is 2.2." Lots of filters, plug-ins and scripts make the same error.

Pillow seems to be affected by that problem too as the following test script demonstrates:

```python
from PIL import Image
from cStringIO import StringIO
from urllib2 import urlopen

f = urlopen('http://www.4p8.com/eric.brasseur/gamma_dalai_lama_gray.jpg')
img = Image.open(StringIO(f.read()))
img.show()

newsize = tuple([i/2 for i in img.size])
imgscaled = img.resize(newsize, Image.ANTIALIAS)
imgscaled.show()
```

znerol:
There is an [SO question](http://stackoverflow.com/q/31300865/2779045) providing a workaround using numpy.

wiredfool:
There is essentially no color space awareness in Pillow by default. It's possible to have some compiled in, but it's optional. So it's not terribly surprising that it's not taking one specific colorspace into account when doing geometric transformations. 

If you really wish to do it properly, you're probably better off doing the operations in L*a*b space instead of sRGB anyway.

znerol:
Right. For future reference, this is one way to implement transformations in linearized RGB color space:

```python
from PIL import Image, ImageCms
from cStringIO import StringIO
from urllib2 import urlopen

f = urlopen('http://www.4p8.com/eric.brasseur/gamma_dalai_lama_gray.jpg')
#f = urlopen('http://www.4p8.com/eric.brasseur/gamma_cloud_gate.jpg')
img = Image.open(StringIO(f.read()))
img.show()

newsize = tuple([i/2 for i in img.size])

# Load profiles.
srgb_profile = ImageCms.getOpenProfile("/usr/share/color/icc/sRGB.icc")
lsrgb_profile = ImageCms.getOpenProfile("linearized-sRGB.icc")

# Linearize RGB color space, resize and convert back to sRGB.
if not img.mode in ('RGB', 'RGBA'):
    img = img.convert('RGB')

lsrgbimg = ImageCms.profileToProfile(img, srgb_profile, lsrgb_profile)
[...]

PR #8811, opened 2025-03-11
Add support for Magic Kernel Sharp resizing modes

drhead:
Original formula is by John Costella, link includes paper with explanation of why it works better than Lanczos and demonstrations of its performance: https://johncostella.com/magic/

This is a simple implementation of the Magic Kernel Sharp 2021 resize filter. It should produce higher quality results for resampling, either upsampling or downsampling, than Lanczos.

This PR is incomplete, I still need to test it against the reference implementation of MKS, write tests for it, update documentation samples, and run linting. The outputs that I have gotten from preliminary testing overall look as expected and look better than Lanczos in the expected ways (reduced moiré patterning on downsampling, reduced blockiness on upscaling), so it should be in a state where it is suitable for experimentation if wanted.


drhead:
A notable discrepancy:
image
("pillow (linear)" is applying the resize after a manual conversion to linear sRGB space, "reference (sRGB)" is the reference resizer run with `--no-gamma-handling`.)

The correct behavior would be to convert images to a linear color space before resizing. If the situation has changed at all since #1604 was filed it may be worth revisiting, but in either case it's out of scope for this. As you can see, it matches the reference implementation without gamma handling by default, and when applied in linear sRGB space it matches the reference with gamma handling, both within reasonable quantization error.

Just for reference, this is the amount of difference this produces on a test image:
image

drhead:
I have updated this to include Magic Kernel Sharp 2013 (at the suggestion of the creator -- it should have the same effects in combatting moire/blockiness while being slightly faster than Lanczos 3) as well as the 2021 implementation.

I have also added these kernels to documentation at appropriate points and added them to tests. The additions to concepts.rst may need revision. I have also not added the `test_reduce_x` and `test_enlarge_x` tests because I am not completely certain of how those tests need to be properly set up and would want some guidance on that first -- as far as I am aware the current implementation right now is correct and could be trusted to generate testing code for this. The only other point in the code that I think might need some advisement is whether MKS2021 should replace LANCZOS at [src/PIL/IcoImagePlugin.py:77](https://github.com/python-pillow/Pillow/blob/cda26be10e93472cca7cc2c07c045037572d9b38/src/PIL/IcoImagePlugin.py#L77). In any case, I'm opening the issue for review.

src/libImaging/Resample.c
```diff
@@ -79,11 +79,45 @@ lanczos_filter(double x) {
...
+        return 577.0 / 576.0 - 239.0 / 144.0 * pow(x, 2.0);
+    if (x < 1.5)
+        return 35.0 / 36.0 * (x - 1.0) * (x - 239.0 / 140.0);
+    if (x < 2.5)
+        return 1.0 / 6.0 * (x - 2.0) * (65.0 / 24.0 - x);
+    if (x < 3.5)
+        return 1.0 / 36.0 * (x - 3.0) * (x - 15.0 / 4.0);
+    if (x < 4.5)
+        return -1.0 / 288.0 * pow(x - 9.0 / 2.0, 2.0);
+    return (0.0);
```
radarhere, line 111:
```suggestion
    if (x < 0) {
        x = -x;
    }
    if (x < 0.5) {
        return 17.0 / 16.0 - 7.0 / 4.0 * pow(x, 2);
    }
    if (x < 1.5) {
        return (1.0 - x) * (7.0 / 4.0 - x);
    }
    if (x < 2.5) {
        return -1.0 / 8.0 * pow(x - 5.0 / 2.0, 2);
    }
    return 0;
}

static inline double
mks_2021_filter(double x) {
    /* https://johncostella.com/magic/ */
    if (x < 0) {
        x = -x;
    }
    if (x < 0.5) {
        return 577.0 / 576.0 - 239.0 / 144.0 * pow(x, 2);
    }
    if (x < 1.5) {
        return 35.0 / 36.0 * (x - 1.0) * (x - 239.0 / 140.0);
    }
    if (x < 2.5) {
        return 1.0 / 6.0 * (x - 2.0) * (65.0 / 24.0 - x);
    }
    if (x < 3.5) {
        return 1.0 / 36.0 * (x - 3.0) * (x - 15.0 / 4.0);
    }
    if (x < 4.5) {
        return -1.0 / 288.0 * pow(x - 9.0 / 2.0, 2);
    }
    return 0;
```
Adding braces, as per #4618. Also removing '.0' from numbers that don't directly interact with other numbers - `return (0.0)` -> `return 0` and `pow(..., 2.0)` -> `pow(..., 2)`.

cwygoda:
any thoughts on how this will proceed to be released?

wiredfool (approved):

akx:
@radarhere et al.: this would be great to get merged for 13.0, right? 🙏

radarhere (approved):
Just one question - why
```c
if (x < 0) {
    x = -x;
}
```
?

akx:
To make the filter symmetrical around the zero axis? That same pattern appears in `bilinear_filter`, `hamming_filter`, `bicubic_filter`.

radarhere:
Oh, sorry, for some reason I thought it said
```c
if (x < 0) {
    return -x;
}
```

hugovk:
Let's also add to release notes.

radarhere:
I've created https://github.com/python-pillow/Pillow/pull/9937

PR #8819, opened 2025-03-16
Use docker/setup-qemu-action

radarhere:
In January, the ppc64le and s390x jobs started experiencing segfaults. We downgraded the hosts to Ubuntu 22 in #8713, but the segfaults have started happening again since then. I presumed a fix would eventually occur upstream, but it's been a while and we're approaching a release.

This is a suggestion to switch from
```
docker run --rm --privileged aptman/qus -s -- -p ${{ matrix.qemu-arch }}
```
to https://github.com/docker/setup-qemu-action

- Seems simpler
- Fixes the segfaults
- We can revert the temporary downgrading of the hosts

### DIFF
--- a/.github/workflows/test-docker.yml
+++ b/.github/workflows/test-docker.yml
@@ -35,6 +35,10 @@ jobs:
       matrix:
         os: ["ubuntu-latest"]
         docker: [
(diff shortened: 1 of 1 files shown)


hugovk:
Good stuff, thanks!

PR #8823, opened 2025-03-19
Simplified test code

radarhere:
Replace
```python
f = str(tmp_path / "temp.png")
```
with
```python
f = tmp_path / "temp.png"
```
where possible.


hugovk:
I guess most of these are fine, but are any of these actually testing it works with a string?

radarhere:
Almost all of these changes are about opening and saving. I've added an explicit test for opening and saving an image using a string.

An exception is a call to `ImageFont.truetype()`, but we already have other tests that pass strings to that - https://github.com/python-pillow/Pillow/blob/e66ebb64288f6deffe4ef54b624dcbf34570bb02/Tests/test_imagefont.py#L56

Issue #515, opened 2014-02-04
Decompression bomb protection


masklinn:
"zip bombs" are a somewhat know threat, but it also applies to images and can't be protected against by checking the filesystem size of the data:
- http://www.aerasec.de/security/advisories/decompression-bomb-vulnerability.html provides 2 PNG bomb demos, a 4KB 6000x6000 images decompressing to 100MB in memory and a 44KB 19000x19000 image decompressing to 1GB
- JPEG is somewhat less efficient, but e.g. NASA Visible Earth's [Blue Marble](http://visibleearth.nasa.gov/view.php?id=73884) has 21600x21600 tiles as small as 5MB ([tile A2, southern pacific](http://eoimages.gsfc.nasa.gov/images/imagerecords/73000/73884/world.topo.bathy.200411.3x21600x21600.A2.jpg)) decompressing to 1.4GB

This means it's possible to DOS e.g. a web application performing image resizing by sending one of these bombs. As far as I can tell the protection possibilities are limited:
- assert source image sizes before doing any operation which will need the image data, the documentation may benefit from a warning on that subject (similar to [warnings about XML vulnerabilities in the Python documentation](http://docs.python.org/2/library/xml.etree.elementtree.html)), `Image.open` could be augmented with e.g. a `maximum_pixels` parameter raising an error in case `image.h * image.w` goes above the specified limit to make this easier for users
- during image loading/decompression, raise an error if the decompressed data size gets above a specific threshold

regsgiter:
yeah

PR #8826, opened 2025-03-20
 Added GimpPaletteFile frombytes() to allow for unlimited parsing

### LINKED ISSUE #6639: Indexes of colors when reading a GIMP Palette file are wrong

radarhere:
Resolves #6639

#6640 has concerns about Pillow's limits when reading a Gimp palette file.

https://github.com/python-pillow/Pillow/blob/e66ebb64288f6deffe4ef54b624dcbf34570bb02/src/PIL/GimpPaletteFile.py#L33-L34
1. Pillow reads at most 256 lines after the first header line. This means that Pillow can't read more than 256 colours, which [apparently might be the case in Gimp palette files](https://github.com/jsbueno/Pillow/pull/1#discussion_r988488865) and the [spec](https://developer.gimp.org/core/standards/gpl/) doesn't say otherwise. I'm [concerned](https://github.com/python-pillow/Pillow/issues/6639#issuecomment-1269311418) about the idea of creating a "palette" that is larger than what our ImagePalette can handle.
2. The fact that only 256 lines are read after the first header line also means that any additional lines that are present ('Name:' header, 'Columns:' header, comments) actually subtract from the 256 colour limit.

https://github.com/python-pillow/Pillow/blob/e66ebb64288f6deffe4ef54b624dcbf34570bb02/src/PIL/GimpPaletteFile.py#L41-L43
3. A colour's name can't be long enough to make the line more than 100 characters.

My general concern is that the spirit of #515 is against reading an unlimited amount of data from a file.

The test files that we have do include two additional headers and a comment.
https://github.com/python-pillow/Pillow/blob/e66ebb64288f6deffe4ef54b624dcbf34570bb02/Tests/images/custom_gimp_palette.gpl#L1-L5
So I think extending the number of lines read by three is reasonable. That is my first commit, partly addressing point 2.

Beyond that, I think there are three options.
A. Do nothing. Palettes with more than 256 colours and long colour names are an edge case, and this is a lesser used format.
B. #6640's solution is to add class variables to allow configuration - `max_colors`, `_max_line_size` and `_max_file_size`. I think that makes for a bit of a complex API.
C. My idea is to add a `frombytes()` method that accepts bytes. That way, if the user doesn't like our limits, they can take on the responsibility of reading everything from a file, and we can just parse the result.

### DIFF
--- a/Tests/images/full_gimp_palette.gpl
+++ b/Tests/images/full_gimp_palette.gpl
@@ -1,0 +1,260 @@
+GIMP Palette
+Name: fullpalette
+Columns: 4
+#
+  0   0   0     Index 0
+  1   1   1     Index 1
+  2   2   2     Index 2
+  3   3   3     Index 3
+  4   4   4     Index 4
+  5   5   5     Index 5
+  6   6   6     Index 6
+  7   7   7     Index 7
+  8   8   8     Index 8
+  9   9   9     Index 9
+ 10  10  10     Index 10
+ 11  11  11     Index 11
+ 12  12  12     Index 12
+ 13  13  13     Index 13
+ 14  14  14     Index 14
+ 15  15  15     Index 15
+ 16  16  16     Index 16
+ 17  17  17     Index 17
+ 18  18  18     Index 18
+ 19  19  19     Index 19
+ 20  20  20     Index 20
+ 21  21  21     Index 21
+ 22  22  22     Index 22
+ 23  23  23     Index 23
+ 24  24  24     Index 24
+ 25  25  25     Index 25
+ 26  26  26     Index 26
+ 27  27  27     Index 27
+ 28  28  28     Index 28
+ 29  29  29     Index 29
+ 30  30  30     Index 30
+ 31  31  31     Index 31
+ 32  32  32     Index 32
+ 33  33  33     Index 33
+ 34  34  34     Index 34
+ 35  35  35     Index 35
+ 36  36  36     Index 36
+ 37  37  37     Index 37
+ 38  38  38     Index 38
+ 39  39  39     Index 39
+ 40  40  40     Index 40
+ 41  41  41     Index 41
+ 42  42  42     Index 42
+ 43  43  43     Index 43
+ 44  44  44     Index 44
+ 45  45  45     Index 45
+ 46  46  46     Index 46
+ 47  47  47     Index 47
+ 48  48  48     Index 48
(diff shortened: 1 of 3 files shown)


aclark4life:
Yeah I like `frombytes` too and I agree with you that adding class variables is not great.

PR #8840, opened 2025-03-28
Enable memory mapping on PyPy

radarhere:


src/PIL/Image.py
```diff
@@ -621,6 +621,8 @@ def close(self) -> None:
         more information.
         """
         if getattr(self, "map", None):
+            if sys.platform == "win32" and hasattr(sys, "pypy_version_info"):
+                self.map.close()
```
radarhere, line 625: Without this change, [`test_fd_leak()`](https://github.com/python-pillow/Pillow/blob/1cb6c7c347a7509e0303c32cfe8f5d1e063be1b3/Tests/test_file_tiff.py#L976) fails - https://github.com/radarhere/Pillow/actions/runs/14129067570/job/39584722751#step:28:1319

PR #8851, opened 2025-04-01
Allow cmake<4 when building libavif and libtiff

radarhere:
libtiff has started failing to build in main on WIndows - https://github.com/python-pillow/Pillow/actions/runs/14187037544/job/39744130790

This would be due to the recent release of cmake 4 - https://github.com/Kitware/CMake/releases/tag/v4.0.0

https://cmake.org/cmake/help/latest/release/4.0.html#deprecated-and-removed-features
> Compatibility with versions of CMake older than 3.5 has been removed.

### DIFF
--- a/winbuild/build_prepare.py
+++ b/winbuild/build_prepare.py
@@ -235,6 +235,7 @@ DEPS: dict[str, dict[str, Any]] = {
                 "-DBUILD_SHARED_LIBS:BOOL=OFF",
                 "-DWebP_LIBRARY=libwebp",
                 '-DCMAKE_C_FLAGS="-nologo -DLZMA_API_STATIC"',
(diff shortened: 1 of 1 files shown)


hugovk (approved):

radarhere:
They mysteriously reverted to the 20250309.1.0 runner image. I reran the jobs and they passed this time.

If you'd like, https://github.com/python-pillow/Pillow/pull/8855 adds `rustup update` to fix the problem even in the earlier runner image.

hugovk:
Thanks! Good to know we can restart it and hopefully get the newer runner.

I'll go ahead with the release without merging it to save a bunch of CI time, but we can merge it afterwards if we see it pop up again.

PR #8855, opened 2025-04-01
Always update rust when building libavif on Windows

radarhere:
Recently in the libavif PR, an error occurred when building rav1e in Windows wheels. I reported it at https://github.com/AOMediaCodec/libavif/issues/2703, and the suggestion was to install a Rust toolchain - https://github.com/fdintino/Pillow/pull/32

I opted for a simpler fix, `rustup update` - https://github.com/fdintino/Pillow/pull/33

I noticed the same problem happening in the non-wheels Windows jobs, but a new runner image fixed the problem.

https://github.com/python-pillow/Pillow/pull/8851#issuecomment-2768884176 has now reported a problem building rav1e in the Windows jobs. Those jobs have, for some reason, decided to suddenly use a runner image two weeks old.

Running the jobs again allows them to pass, but I offer this PR as a more proactive measure if you would like one - running `rustup update` in all the Windows builds.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -160,11 +160,6 @@ jobs:
           & python.exe winbuild\build_prepare.py -v --no-imagequant --architecture=${{ matrix.cibw_arch }}
         shell: pwsh
 
-      - name: Update rust
-        if: matrix.cibw_arch == 'AMD64'
-        run: |
-          rustup update
-
       - name: Build wheels
         run: |
           setlocal EnableDelayedExpansion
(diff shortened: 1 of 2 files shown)


hugovk:
Thanks! Good to know we can restart it and hopefully get the newer runner.

I'll go ahead with the release without merging it to save a bunch of CI time, but we can merge it afterwards if we see it pop up again.

wiredfool:
Does this mean that we have rust available across all the platforms now?

radarhere:
Not quite. Cygwin and the Docker images don't have rust, but they also don't have libavif support at the moment. I don't think MinGW has rust, but I might just be missing some MinGW nuance.

PR #8858, opened 2025-04-01
Add AVIF to wheels using only aomenc and dav1d AVIF codecs for reduced size

### LINKED ISSUE #8856: Reduce AVIF wheel size?
The AVIF .so has made the wheels about 4 times bigger: https://github.com/python-pillow/Pillow/issues/8722#issuecomment-2769412355

Is there anything we can do to make it smaller?

fdintino:
Fixes #8856 

This removes all libavif codecs except for AOM for encoding and dav1d for decoding. It also builds libavif as a shared library on linux and macOS, which modestly decreases the file size because it allows the linker to omit unused objects from the codec libraries. Building with `-Os` and `-flto` reduces the size by another further 0.7-1.5MB.

Below is a comparison of the average CPython wheel sizes by platform for the wheels (MiB) prior to the merge of the avif pull request, the current 11.2.0 wheels, and the wheels from this pull request. An additional few-hundred K could be saved by using aom's decoder instead of dav1d, but I think the benefits of dav1d outweigh those cost savings.

Since Pillow does not yet support multi-channel high bit depth images, I've compiled libaom without high-bit-depth support. This further reduces wheel sizes by ~200KiB.

In absolute terms, the uncompressed libavif shared library file size is somewhere between 2.8MiB (macOS ARM) and 5.0MiB (musllinux x86_64), with the exception of manylinux2014 x86_64 which, because of a gcc bug, is missing some optimizations and so is 7.4MiB.

| platform | pre-AVIF | 11.2.0 | aomenc + dav1d | `-flto -Os` | further optimizations |
| -- | -- | -- | -- | -- | -- | 
| macosx_86_64 | 3.0 | 10.9 | 6.9 | 5.8 | 5.1 |
| macosx_arm64 | 2.9 | 8.5 | 5.7 | 4.9 | 4.5 |
| win_amd64 | 2.5 | 13.1 | 7.4 | 6.9 | 6.7 |
| manylinux_2_17_aarch64 | 4.2 | 16.3 | 7.7 | 6.3 | 5.6 |
| manylinux_2_17_x86_64 | 4.3 | 18.0 | 8.2 | 8.2 | 7.3 |
| manylinux_2_28_aarch64 | 4.3 | 16.8 | 8.0 | 6.5 | 5.7 |
| manylinux_2_28_x86_64 | 4.4 | 18.7 | 8.6 | 7.2 | 6.3 |
| musllinux_1_2_aarch64 | 4.3 | 17.2 | 8.3 | 6.6 | 5.8 |
| musllinux_1_2_x86_64 | 4.4 | 18.8 | 8.7 | 7.3 | 6.4 |

### DIFF
--- a/docs/releasenotes/11.3.0.rst
+++ b/docs/releasenotes/11.3.0.rst
@@ -71,7 +71,7 @@ AVIF support in wheels
 ^^^^^^^^^^^^^^^^^^^^^^
 
 Support for reading and writing AVIF images is now included in Pillow's wheels, except
-for Windows arm64. libaom is available as an encoder and dav1d as a decoder.
+for Windows ARM64. libaom is available as an encoder and dav1d as a decoder.
 
 Python 3.14 beta
 ^^^^^^^^^^^^^^^^


.github/workflows/wheels-dependencies.sh
```diff
@@ -137,14 +137,16 @@ function build_libavif {
...
             -DCMAKE_INSTALL_LIBDIR=$BUILD_PREFIX/lib \
+            -DCMAKE_INSTALL_NAME_DIR=$BUILD_PREFIX/lib \
+            -DCMAKE_MACOSX_RPATH=OFF \
             -DCMAKE_BUILD_TYPE=Release \
-            -DBUILD_SHARED_LIBS=OFF \
+            -DBUILD_SHARED_LIBS=ON \
             -DAVIF_LIBSHARPYUV=LOCAL \
             -DAVIF_LIBYUV=LOCAL \
             -DAVIF_CODEC_AOM=LOCAL \
+            -DCONFIG_AV1_DECODER=0 \
```
wantehchang, line 147: IMPORTANT: The `-DCONFIG_AV1_DECODER=0` needs to be passed to libaom's cmake command. We are passing it to libavif's cmake command here. Does libavif's cmake command forward it to libaom's cmake command? Otherwise this has no effects and should be removed.

Does this require changing libavif's cmake/Modules/LocalAom.cmake?

winbuild/build_prepare.py
```diff
@@ -392,9 +392,9 @@ def cmd_msbuild(
                 "-DAVIF_LIBSHARPYUV=LOCAL",
                 "-DAVIF_LIBYUV=LOCAL",
                 "-DAVIF_CODEC_AOM=LOCAL",
+                "-DCONFIG_AV1_DECODER=0",
```
wantehchang, line 395: Same here. We need to pass `-DCONFIG_AV1_DECODER=0` to libaom's cmake command.

.github/workflows/wheels-dependencies.sh
fdintino: To your other question though, it would probably make sense for us to `set(CONFIG_AV1_DECODER 0)` in `LocalAom.cmake` if `AVIF_CODEC_AOM_DECODE` is `OFF`, so that passing this is unnecessary.

wantehchang: Frankie: Thanks for the explanation. Let's add a comment to note that `CONFIG_AV1_DECODER` is a cmake flag of libaom.

I will take a stab at modifying libavif's cmake/Modules/LocalAom.cmake.

.github/workflows/wheels-dependencies.sh
```diff
@@ -137,14 +137,16 @@ function build_libavif {
         && CMAKE_POLICY_VERSION_MINIMUM=3.5 cmake \
             -DCMAKE_INSTALL_PREFIX=$BUILD_PREFIX \
             -DCMAKE_INSTALL_LIBDIR=$BUILD_PREFIX/lib \
+            -DCMAKE_INSTALL_NAME_DIR=$BUILD_PREFIX/lib \
+            -DCMAKE_MACOSX_RPATH=OFF \
             -DCMAKE_BUILD_TYPE=Release \
-            -DBUILD_SHARED_LIBS=OFF \
+            -DBUILD_SHARED_LIBS=ON \
```
wantehchang, line 143: Did you mean to change `BUILD_SHARED_LIBS` to `ON`? It is still `OFF` in winbuild/build_prepare.py.

fdintino: The savings are modest, about 0.5 MB, but there isn't any cost to switching to a shared library.

doublex:
Is `codec` still useful in `save()`?

.github/workflows/wheels-dependencies.sh
```diff
@@ -133,6 +133,8 @@ function build_libavif {
     fi
 
     local out_dir=$(fetch_unpack https://github.com/AOMediaCodec/libavif/archive/refs/tags/v$LIBAVIF_VERSION.tar.gz libavif-$LIBAVIF_VERSION.tar.gz)
+    # CONFIG_AV1_DECODER=0 is a flag for libaom (included as a subproject of
+    # libavif) to disable the compilation and inclusion of aom's AVIF decoder.
```
wantehchang, line 137: Nit: AVIF => AV1

libaom is an AV1 codec.

.github/workflows/wheels-dependencies.sh
```diff
@@ -134,7 +134,7 @@ function build_libavif {
 
     local out_dir=$(fetch_unpack https://github.com/AOMediaCodec/libavif/archive/refs/tags/v$LIBAVIF_VERSION.tar.gz libavif-$LIBAVIF_VERSION.tar.gz)
     # CONFIG_AV1_DECODER=0 is a flag for libaom (included as a subproject of
-    # libavif) to disable the compilation and inclusion of aom's AVIF decoder.
+    # libavif) to disable the compilation and inclusion of aom's AV1 decoder.
```
wantehchang, line 137: I also suggest moving `-DENABLE_NASM=ON` next to `-DCONFIG_AV1_DECODER=0` because it is also a cmake flag for libaom.

Why do you need to set `ENABLE_NASM` to `ON`? Are you concerned that yasm may be used?

fdintino: I think it's probably unnecessary. That flag goes back to when I first opened #5201, over four years ago. If there was a reason for its inclusion then (and there may not have been) I have forgotten it.

fdintino: `yasm` won't be in the `PATH` in any of the build environments, so even if it was a matter of forcing the user of nasm over yasm, it wouldn't be necessary. I'll remove it.

wantehchang:
Frankie: Here are some other ideas to further reduce the binary size.

1. On Linux, use the compiler flags `-ffunction-sections -fdata-sections` and the linker flag `-Wl,--gc-sections`. See, for example, https://www.sandordargo.com/blog/2023/07/19/binary-sizes-and-compiler-flags and https://www.sandordargo.com/blog/2023/07/19/binary-sizes-and-compiler-flags.
2. Look at the binary sizes of libsharpyuv and libyuv.
    - If libsharpyuv is big, consider changing `-DAVIF_LIBSHARPYUV=LOCAL` to `-DAVIF_LIBSHARPYUV=OFF`.
    - libyuv is critical to performance. If libyuv is big, I will look into how to make it smaller.
3. In libaom, it may be possible to exclude the fallback SIMD optimizations and only include the "highest" SIMD optimizations. For example, if a function has an AVX2 optimization, then exclude its SSE4 and SSE2 optimizations. I will look into this.

Tests/test_file_avif.py
```diff
@@ -146,7 +146,7 @@ def test_write_rgb(self, tmp_path: Path) -> None:
 
             # avifdec hopper.avif avif/hopper_avif_write.png
             assert_image_similar_tofile(
-                reloaded, "Tests/images/avif/hopper_avif_write.png", 6.02
+                reloaded, "Tests/images/avif/hopper_avif_write.png", 6.14
```
wantehchang, line 149: Why do we need to relax the average pixel difference threshold?

wantehchang, line 149: Why do we need to relax the average pixel difference threshold?

fdintino: I changed this temporarily; it has since been changed back. It seems that when aom is built with `-Os` for manylinux2014 wheels (i.e. CentOS 7 with gcc 10) the encoder behavior changes. I think this must be a bug with gcc? I can't see why compiler optimization flags should affect the actual behavior of an application. I've opted to use `-DCMAKE_BUILD_TYPE=Release` for these el7 builds.

hugovk:
The alternative #8869 removed installation of things like `rav1e` and `svt-av1`, are they needed here?

fdintino:
They have been removed here as well. The only difference between this PR and that one is that this one uses the dav1d decoder instead of AOM's. And it leaves open the possibility of compiling for more codecs

wantehchang:
@hugovk Another difference between this PR and the alternative https://github.com/python-pillow/Pillow/pull/8869 is that this PR  has additional changes to the build system to reduce the binary size, specifically:

- Change `-DBUILD_SHARED_LIBS=OFF` to `-DBUILD_SHARED_LIBS=ON` (except on Windows)
- Change `-DCMAKE_BUILD_TYPE=Release` to `-DCMAKE_BUILD_TYPE=MinSizeRel` under some condition (I haven't figured out what the condition is)
- Add `-DCMAKE_INTERPROCEDURAL_OPTIMIZATION=ON`

For a fair comparison, these changes should be applied to the alternative https://github.com/python-pillow/Pillow/pull/8869.

.github/workflows/wheels-dependencies.sh
```diff
@@ -122,31 +122,37 @@ function build_libavif {
...
+    # CONFIG_AV1_DECODER=0 is a flag for libaom (included as a subproject of
+    # libavif) to disable the compilation and inclusion of aom's AV1 decoder.
     (cd $out_dir \
         && CMAKE_POLICY_VERSION_MINIMUM=3.5 cmake \
             -DCMAKE_INSTALL_PREFIX=$BUILD_PREFIX \
             -DCMAKE_INSTALL_LIBDIR=$BUILD_PREFIX/lib \
-            -DCMAKE_BUILD_TYPE=Release \
-            -DBUILD_SHARED_LIBS=OFF \
+            -DCMAKE_INSTALL_NAME_DIR=$BUILD_PREFIX/lib \
+            -DCMAKE_MACOSX_RPATH=OFF \
```
radarhere, line 146: What was the thinking behind `CMAKE_MACOSX_RPATH=OFF`?

radarhere:
> - Change `-DCMAKE_BUILD_TYPE=Release` to `-DCMAKE_BUILD_TYPE=MinSizeRel` under some condition (I haven't figured out what the condition is)

It is still using Release for manylinux2014.

.github/workflows/wheels-dependencies.sh
fdintino: Without it, delocate fails on libavif.dylib

fdintino: It looks as though my delocate `@rpath` error was fixed by the addition of `-DCMAKE_INSTALL_LIB_DIR`, making this change unnecessary

.github/workflows/wheels-dependencies.sh
```diff
@@ -122,31 +122,29 @@ function build_libavif {
...
-    curl https://sh.rustup.rs -sSf | sh -s -- -y
-    . "$HOME/.cargo/env"
-    if [ -z "$IS_ALPINE" ] && [ -z "$SANITIZER" ] && [ -z "$IS_MACOS" ]; then
-        yum install -y perl
-        if [[ "$MB_ML_VER" == 2014 ]]; then
-            yum install -y perl-IPC-Cmd
-        fi
+    local build_type=MinSizeRel
+
+    if [[ -z "$IS_ALPINE" ]] && [[ "$MB_ML_VER" == 2014 ]]; then
```
radarhere, line 127:
```suggestion
    if [[ "$MB_ML_VER" == 2014 ]]; then
```

.github/workflows/wheels-dependencies.sh
radarhere: When you say 0.5mb, I presume that's per wheel. Which wheel(s) were you using to check that?

.github/workflows/wheels-dependencies.sh
```diff
@@ -122,31 +122,29 @@ function build_libavif {
...
             -DCMAKE_INSTALL_LIBDIR=$BUILD_PREFIX/lib \
-            -DCMAKE_BUILD_TYPE=Release \
-            -DBUILD_SHARED_LIBS=OFF \
+            -DCMAKE_INSTALL_NAME_DIR=$BUILD_PREFIX/lib \
+            -DBUILD_SHARED_LIBS=ON \
             -DAVIF_LIBSHARPYUV=LOCAL \
             -DAVIF_LIBYUV=LOCAL \
             -DAVIF_CODEC_AOM=LOCAL \
+            -DCONFIG_AV1_DECODER=0 \
+            -DAVIF_CODEC_AOM_DECODE=OFF \
```
wantehchang, line 144: Frankie: Is it possible to build with the libaom and dav1d packages in Linux distributions (i.e., `-DAVIF_CODEC_AOM=SYSTEM `-DAVIF_CODEC_DAV1D=LOCAL`)? That will allow the Pillow Library to share the libaom and dav1d shared libraries with other applications on the system.

fdintino: If someone were to build Pillow from source, they could use libavif from the distro, or build libavif against system libraries. The binary wheels are tagged for python version, system architecture, and ABI compatibility (musl and two different versions of glibc). They are intended to be installed in any compatible environment without any other dependencies.

fdintino:
It actually looks like macOS is the only platform where `CMAKE_INTERPROCEDURAL_OPTIMIZATION=ON` results in larger binaries. They are only modestly smaller on some platforms (less than 40KiB on manylinux_2_28 and musllinux), but the LTO binaries are 1.3 MiB smaller on manylinux_2_17 aarch64.

radarhere (approved):

radarhere:
> Another difference between this PR and the alternative #8869 is that this PR has additional changes to the build system to reduce the binary size, specifically:
> 
> * Change `-DBUILD_SHARED_LIBS=OFF` to `-DBUILD_SHARED_LIBS=ON` (except on Windows)
> * Change `-DCMAKE_BUILD_TYPE=Release` to `-DCMAKE_BUILD_TYPE=MinSizeRel` under some condition (I haven't figured out what the condition is)
> * Add `-DCMAKE_INTERPROCEDURAL_OPTIMIZATION=ON`
> 
> For a fair comparison, these changes should be applied to the alternative #8869.

To answer this, if I download wheels from the current state of this PR, I get 522.92mb in total.
If I remove
```
-DCONFIG_AV1_DECODER=0 \
-DAVIF_CODEC_AOM_DECODE=OFF \
-DAVIF_CODEC_DAV1D=LOCAL \
```
I get 488.77mb.

doublex:
Not in wheels: this is regrettable.
AVIF is supported by all modern browsers.
Corresponding xkcd: https://xkcd.com/619/

aclark4life:
> Not in wheels: this is regrettable.

@doublex Please see my [comments here](https://github.com/python-pillow/Pillow/pull/8876#issuecomment-2798915567) and clarify if you find the decision "regrettable" even with concerns expressed about wheel size? I'm not sure any feature is worth universally increasing the wheel size by a factor of 10. If we can do it optionally, then not getting to it in this release is regrettable I'd say. But I'm not sure yet how that would work.

radarhere:
If anyone watching this would like to download prebuild Pillow wheels with libavif, be aware that you can use the artifacts from the GitHub Actions of this PR, for example by going to https://github.com/python-pillow/Pillow/actions/runs/14424832221 and scrolling down.

fdintino:
@aclark4life Those comments were made with regard to the initial 11.2.0 release. Those did indeed result in wheels that were as much as 4x larger and uncompressed sizes of 6x or more. Because of various size optimizations, the AVIF feature now only increases both by about 60% (a little more or less, depending on the platform).

I think the door was left open for AVIF ending up in the 11.3.0 wheels, judging from @hugovk's comment here: https://github.com/python-pillow/Pillow/pull/8869#issuecomment-2791718428. The number of people who would be negatively affected by the addition to their virtualenv of a 50 MB file is, I suspect, reasonably large. But 5 MB? For people looking to use Pillow in such resource-constrained environments, we could perhaps provide guidance on how to delete undesired bundled dependencies.

.github/workflows/wheels-dependencies.sh
```diff
@@ -105,7 +105,7 @@ function build_libavif {
     python3 -m pip install meson ninja
 
     if [[ "$PLAT" == "x86_64" ]] || [ -n "$SANITIZER" ]; then
-        build_simple nasm 2.16.03 https://www.nasm.us/pub/nasm/releasebuilds/2.16.03
```
radarhere: Fyi, if the nasm.us outage is temporary, then the presence of https://github.com/python-pillow/pillow-depends/blob/main/nasm-2.16.03.tar.gz should allow this to continue working until it comes back.

fdintino: I'm not sure it's temporary, since it's been down now for nearly a week. But I had forgotten about the pillow-depends backup. I've reverted this change.

fdintino: I looked into that. The source tarballs on nasm.us have the autoconf and automake artifacts included. So we couldn't just swap out with a GitHub tag archive here, as we would need to update this script to call autogen.sh. Not an insurmountable obstacle, but it couldn't use the `build_simple` bash function from multibuild.

.github/workflows/wheels-dependencies.sh
```diff
@@ -98,6 +99,62 @@ function build_harfbuzz {
...
+        libavif_cmake_flags+=(
+            -DCMAKE_C_FLAGS_MINSIZEREL="-Oz -DNDEBUG -flto " \
+            -DCMAKE_CXX_FLAGS_MINSIZEREL="-Oz -DNDEBUG -flto" \
+            -DCMAKE_SHARED_LINKER_FLAGS_INIT="-Wl,-S,-x,-dead_strip_dylibs" \
+        )
+    else
+        if [[ "$MB_ML_VER" == 2014 ]] && [[ "$PLAT" == "x86_64" ]]; then
+            build_type=Release
+        fi
+        libavif_cmake_flags+=(-DCMAKE_SHARED_LINKER_FLAGS_INIT="-Wl,--strip-all,-z,relro,-z,now")
```
radarhere, line 127: If I compare the wheel sizes from https://github.com/python-pillow/Pillow/actions/runs/14773502497?pr=8858,
with a commit where I [remove](https://github.com/radarhere/Pillow/commit/73c2f320ce3f1f1b186ba42bfaa0237a0f9a28a5) this line, I find that https://github.com/radarhere/Pillow/actions/runs/14773542213 is actually 8kb smaller.

If this line is purely about reducing file size, then I don't think it's fulfilling that role.

fdintino: Ah, it's because there is a typo. I'll fix it.

fdintino: It actually had to do with a cmake policy that was disabling LTO on linux. See the build here: https://github.com/fdintino/Pillow/actions/runs/14777836226

fdintino: In the PR description, I've updated the paragraph that begins with "In absolute terms,..." and the last column of the table with the latest wheels and shared library file sizes. You can look at the github diff under the list of edits to see the change.

.github/workflows/wheels-dependencies.sh
radarhere: nasm.us is back now.

.github/workflows/wheels-dependencies.sh
```diff
@@ -51,6 +51,7 @@ LIBWEBP_VERSION=1.5.0
 BZIP2_VERSION=1.0.8
 LIBXCB_VERSION=1.17.0
 BROTLI_VERSION=1.1.0
+LIBAVIF_VERSION=1.2.1
```
radarhere, line 54:
```suggestion
LIBAVIF_VERSION=1.3.0
```
libavif 1.3.0 has been released - https://github.com/AOMediaCodec/libavif/releases/tag/v1.3.0

wantehchang: Thanks for the suggested change. I suggest two related changes if you will only be building with libavif v1.3.0 or later.

1. Please delete all occurrences of `CONFIG_AV1_DECODER=0` in your build files (including the comments). libavif v1.3.0 or later sets `CONFIG_AV1_DECODER` to 0 if you pass `-DAVIF_CODEC_AOM_DECODE=OFF` to libavif.
2. Please try removing `"-DCMAKE_POLICY_VERSION_MINIMUM=3.5",`. We fixed all the old `cmake_minimum_required()` versions in libavif v1.3.0. You should not need this workaround now.

.github/workflows/wheels-dependencies.sh
```diff
@@ -98,6 +99,64 @@ function build_harfbuzz {
...
+            build_type=Release
+        fi
+        libavif_cmake_flags+=(-DCMAKE_SHARED_LINKER_FLAGS_INIT="-Wl,--strip-all,-z,relro,-z,now")
+    fi
+
+    local out_dir=$(fetch_unpack https://github.com/AOMediaCodec/libavif/archive/refs/tags/v$LIBAVIF_VERSION.tar.gz libavif-$LIBAVIF_VERSION.tar.gz)
+    # CONFIG_AV1_DECODER=0 is a flag for libaom (included as a subproject of
+    # libavif) to disable the compilation and inclusion of aom's AV1 decoder.
+    # CONFIG_AV1_HIGHBITDEPTH=0 is another flag for libaom that disables support
+    # for encoding high bit depth images.
```
wantehchang, line 134: We should not disbale support for encoding high bit depth images. Most AVIF HDR images are 10 bits. Can we avoid this?

fdintino: Pillow does not yet support high bit depth images (see #1888). It is currently a work-in-progress, but at the moment there can only ever be 8-bit images passed to the C API.

fdintino: I think we might still want to allow decoding of 10-bit images, even if that necessarily means downsampling to 8 bit to load them into Pillow.

winbuild/build_prepare.py
```diff
@@ -116,7 +119,7 @@ def cmd_msbuild(
     "HARFBUZZ": "11.1.0",
     "JPEGTURBO": "3.1.0",
     "LCMS2": "2.17",
-    "LIBAVIF": "1.2.1",
+    "LIBAVIF": "1.3.0",
```
wantehchang, line 122: Please also change the libavif version to 1.3.0 in Pillow/depends/install_libavif.sh.

.github/workflows/wheels-dependencies.sh
radarhere: From what I see, these suggestions have been applied.

Tests/test_file_avif.py
radarhere: This change is no longer present. See https://github.com/python-pillow/Pillow/pull/8858#discussion_r2029662183

hugovk:
This PR was about x1.8 in April (https://github.com/python-pillow/Pillow/issues/8856#issuecomment-2827278596), and we're getting close to the next release, how is it looking now?

doublex:
Yes, please.
That would greatly simplify deployment.

docs/releasenotes/11.3.0.rst
```diff
@@ -67,6 +67,12 @@ Pillow only supports libavif 1.0.0 or later. In order to prevent errors when bui
 from source, if a user happens to have an earlier libavif on their system, Pillow will
 now ignore it.
 
+AVIF support in wheels
+^^^^^^^^^^^^^^^^^^^^^^
+
+Support for reading and writing AVIF images is now included in Pillow's wheels, with
+libaom available as an encoder and dav1d as a decoder.
```
radarhere:
```suggestion
Support for reading and writing AVIF images is now included in Pillow's wheels, except
for Windows arm64. libaom is available as an encoder and dav1d as a decoder.
```

I realised I should have mentioned Windows arm64 wheels do not have libavif.

radarhere:
> This PR was about x1.8 in April ([#8856 (comment)](https://github.com/python-pillow/Pillow/issues/8856#issuecomment-2827278596)), and we're getting close to the next release, how is it looking now?

The size of Pillow 11.2.1 artefacts (wheels and sdist zips) was 321.03mb

The size from main is currently 374.13mb - it's expected that would go up though, as we now include Python 3.14 beta wheels.

The size from this PR is currently 591.47mb - so that's 1.58 times main.

.github/workflows/wheels-dependencies.sh
```diff
@@ -98,6 +99,59 @@ function build_harfbuzz {
...
+
+    local build_type=MinSizeRel
+    local lto=ON
+
+    local libavif_cmake_flags
+
+    if [ -n "$IS_MACOS" ]; then
+        lto=OFF
+        libavif_cmake_flags=(
+            -DCMAKE_C_FLAGS_MINSIZEREL="-Oz -DNDEBUG -flto " \
```
radarhere:
```suggestion
            -DCMAKE_C_FLAGS_MINSIZEREL="-Oz -DNDEBUG -flto" \
```

radarhere:
I expect libsharpyuv doesn't have a license file here because it comes from libwebp.

radarhere (approved):

docs/releasenotes/11.3.0.rst
```diff
@@ -67,6 +67,12 @@ Pillow only supports libavif 1.0.0 or later. In order to prevent errors when bui
 from source, if a user happens to have an earlier libavif on their system, Pillow will
 now ignore it.
 
+AVIF support in wheels
+^^^^^^^^^^^^^^^^^^^^^^
+
+Support for reading and writing AVIF images is now included in Pillow's wheels, except
+for Windows arm64. libaom is available as an encoder and dav1d as a decoder.
```
hugovk:
```suggestion
for Windows ARM64. libaom is available as an encoder and dav1d as a decoder.
```

PR #8859, opened 2025-04-02
Include filename in ImageFile state

### LINKED ISSUE #8857: Change in behavior with JPEG Images in 11.2.0
It looks like something in Pillow 11.2 introduced an issue similar to this: https://github.com/python-pillow/Pillow/issues/8472. This time with the `filename` attribute.

### What did you do?

```python
import pickle

from io import BytesIO

from PIL import Image

im = Image.open("anyimage.jpg")

p = pickle.dumps(im)
im = pickle.loads(p)

# Fails in 11.2, works in 11.1
output = BytesIO()
im.save(output, format=im.format, quality="keep")
```

The code above to work.

### What actually happened?

```pytb
  File "test.py", line 14, in <module>
    im.save(output, format=im.format, quality="keep")
  File "/local/lib64/python3.9/site-packages/PIL/Image.py", line 2546, in save
    if isinstance(self, ImageFile.ImageFile) and filename == self.filename:
  File "/local/lib64/python3.9/site-packages/PIL/JpegImagePlugin.py", line 400, in __getattr__
    raise AttributeError(name)
AttributeError: filename
```

* OS: Linux
* Python: Python 3.9
* Pillow: 11.2

radarhere:
Resolves #8857

Like #8476, this adds another item to the pickled state. When an ImageFile instance is pickled and unpickled, it will now retain the image filename.

In thinking about the image filename, I've also added a change for the #8852 functionality - since the user might open the file with an absolute path, and then save it to a relative path, both targeting the same file, let's compare the basename, rather than the full filename.


hugovk:
Let's be selective in what we merge before we decide whether a new release will go out, but this new-in-11.2.0-(take 1) is a good candidate. Thanks.

PR #8860, opened 2025-04-02
Downgrade libXau to 1.0.11

radarhere:
http://x.org/ is currently down

Some of our dependencies are hosted there. Fortunately, most are cached at https://github.com/python-pillow/pillow-depends. Unfortunately, libXau 1.0.12 is not, and so there are failures in our macOS wheel builds - https://github.com/python-pillow/Pillow/actions/runs/14210030070

So this PR downgrades libXau to 1.0.11 so that the wheels are able to build again.


radarhere:
http://x.org/ is returned, and I've added libXau 1.0.12 to pillow-depends.

PR #8861, opened 2025-04-02
Use multibuild build_github

radarhere:
https://github.com/python-pillow/Pillow/blob/7c56b383ff3ef952854306bbe2c48c5e21066b9c/.github/workflows/wheels-dependencies.sh#L83-L87

can be replaced with

```bash
build_github zlib-ng/zlib-ng $ZLIB_NG_VERSION --zlib-compat
```

thanks to [`build_github()`](https://github.com/multi-build/multibuild/blob/42d761728d141d8462cd9943f4329f12fe62b155/library_builders.sh#L77-L92), which anticipated this scenario.

### DIFF
--- a/.github/workflows/wheels-dependencies.sh
+++ b/.github/workflows/wheels-dependencies.sh
@@ -80,11 +80,7 @@ function build_pkg_config {
 
 function build_zlib_ng {
     if [ -e zlib-stamp ]; then return; fi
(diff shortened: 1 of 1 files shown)


hugovk:
https://github.com/python-pillow/Pillow/pull/8859#issuecomment-2772467237 ... and so is this.

PR #8862, opened 2025-04-02
Removed condition that is always true

radarhere:
https://github.com/python-pillow/Pillow/actions/runs/14220077307/job/39845672092#step:10:365
>   src/_avif.c:571:10: warning: address of 'rgb' will always evaluate to 'true' [-Wpointer-bool-conversion]
>       if (&rgb) {
>       ~~   ^~~

### DIFF
--- a/src/_avif.c
+++ b/src/_avif.c
@@ -568,9 +568,7 @@ _encoder_add(AvifEncoderObject *self, PyObject *args) {
     }
 
 end:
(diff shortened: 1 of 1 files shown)


fdintino (approved):
Makes sense to me. I think this got added in the refactor that added the `goto`; I noticed the warning and had intended to fix it, but it slipped my mind.

PR #8863, opened 2025-04-02
Simplify code

radarhere:


src/PIL/ImageGrab.py
```diff
@@ -134,10 +134,7 @@ def grabclipboard() -> Image.Image | list[str] | None:
             import struct
 
             o = struct.unpack_from("I", data)[0]
-            if data[16] != 0:
-                files = data[o:].decode("utf-16le").split("\0")
-            else:
-                files = data[o:].decode("mbcs").split("\0")
+            files = data[o:].decode("mbcs" if data[16] == 0 else "utf-16le").split("\0")
```
hugovk: If anything, I find the original a bit easier to read and compare the conditions, but I'd flip the condition. 

```suggestion
            if data[16] == 0:
                files = data[o:].decode("mbcs").split("\0")
            else:
                files = data[o:].decode("utf-16le").split("\0")
```

PR #8864, opened 2025-04-03
Replace strncpy with strcpy

radarhere:
I noticed warnings at https://github.com/python-pillow/Pillow/actions/runs/14210030070/job/39830689411#step:5:11057
```
    In file included from /usr/include/string.h:633,
                     from /opt/_internal/cpython-3.9.21/include/python3.9/Python.h:30,
                     from src/libImaging/ImPlatform.h:11,
                     from src/libImaging/Imaging.h:13,
                     from src/libImaging/Arrow.c:3:
    src/libImaging/Arrow.c: In function ‘export_named_type’:
    src/libImaging/Arrow.c:75:5: warning: ‘__builtin_strncpy’ specified bound depends on the length of the source argument [-Wstringop-overflow=]
       75 |     strncpy(formatp, format, format_len);
          |     ^~~~~~~
    src/libImaging/Arrow.c:60:25: note: length computed here
       60 |     size_t format_len = strlen(format) + 1;
          |                         ^~~~~~~~~~~~~~
    In file included from /usr/include/string.h:633,
                     from /opt/_internal/cpython-3.9.21/include/python3.9/Python.h:30,
                     from src/libImaging/ImPlatform.h:11,
                     from src/libImaging/Imaging.h:13,
                     from src/libImaging/Arrow.c:3:
    src/libImaging/Arrow.c:76:5: warning: ‘__builtin_strncpy’ specified bound depends on the length of the source argument [-Wstringop-overflow=]
       76 |     strncpy(namep, name, name_len);
          |     ^~~~~~~
    src/libImaging/Arrow.c:61:23: note: length computed here
       61 |     size_t name_len = strlen(name) + 1;
          |                       ^~~~~~~~~~~~
```

### DIFF
--- a/src/libImaging/Arrow.c
+++ b/src/libImaging/Arrow.c
@@ -61,7 +61,6 @@ export_named_type(struct ArrowSchema *schema, char *format, char *name) {
     size_t name_len = strlen(name) + 1;
 
     formatp = calloc(format_len, 1);
-
     if (!formatp) {
         return IMAGING_CODEC_MEMORY;
     }
@@ -72,8 +71,8 @@ export_named_type(struct ArrowSchema *schema, char *format, char *name) {
         return IMAGING_CODEC_MEMORY;
     }
 
-    strncpy(formatp, format, format_len);
-    strncpy(namep, name, name_len);
+    strcpy(formatp, format);
+    strcpy(namep, name);
 
     *schema = (struct ArrowSchema){// Type description
                                    .format = formatp,


wiredfool:
There's no way that `strcpy` is going to be any _better_ here wrt buffer overflows. I wanted to use a safe `str*cpy` variant, on macos/bsds there's `strlcopy`, but that's not in glibc. 

The warning is saying that the specified bound depends on the length of the source, but the length of the source is used for the malloc call in this very function, so it's also the length of the buffer. 

Specifically, we have the pattern:
```
    size_t format_len = strlen(format) + 1;
    formatp = calloc(format_len, 1);
    strncpy(formatp, format, format_len);
```

radarhere:
Ok. I don't really expect that this will be considered a good idea, but if we want to ensure that this function doesn't perform an unlimited copy, how about breaking at the maximum length that we ever use for the values?

```diff
diff --git a/src/libImaging/Arrow.c b/src/libImaging/Arrow.c
index 33ff2ce77..e2e689023 100644
--- a/src/libImaging/Arrow.c
+++ b/src/libImaging/Arrow.c
@@ -72,8 +72,19 @@ export_named_type(struct ArrowSchema *schema, char *format, char *name) {
         return IMAGING_CODEC_MEMORY;
     }
 
-    strncpy(formatp, format, format_len);
-    strncpy(namep, name, name_len);
+    int i;
+    for (int i = 0; i < format_len; i++) {
+        formatp[i] = format[i];
+        if (i == 5) {
+            break;
+        }
+    }
+    for (int i = 0; i < name_len; i++) {
+        namep[i] = name[i];
+        if (i == 6) {
+            break;
+        }
+    }
 
     *schema = (struct ArrowSchema){// Type description
                                    .format = formatp,
~
```

wiredfool:
First off -- there's no potential for buffer overflow. 

Second, doing it in a while loop is worse -- there are magic constants there that have to be kept in sync, otherwise we'd wind up with an unterminated string and eventually a buffer overflow on read. 

We're controlling the length of the buffers, it's all ok. The warning is wrong because it's picking up that we're basing the length parameter on the length of the string, but doesn't realize that it's the length of the buffer. 

I used `strncpy` instead of `strcpy`, but in this case, they're equivalent because I'm passing in a buffer length to the strncpy that I know to be the length of the string. We could go back to `strcpy`, but that's likely to emit warnings in windows compilations. Windows really prefers you to use `strncpy` or potentially `strcpy_s`, but that one isn't well supported even in their own compiler.

PR #8866, opened 2025-04-03
Fixed conversion of AVIF image rotation property to EXIF orientation

fdintino:
This implements a fix for the issue identified in AOMediaCodec/libavif#2727 and fixed in AOMediaCodec/libavif#2729. The code in Pillow that converts irot and imir properties to EXIF orientation was repurposed from libavif, so it suffers the same bug.

### DIFF
--- a/Tests/test_file_avif.py
+++ b/Tests/test_file_avif.py
@@ -309,7 +309,7 @@ class TestFileAvif:
         assert exif[274] == 3
 
     @pytest.mark.parametrize("use_bytes", [True, False])
-    @pytest.mark.parametrize("orientation", [1, 2])
+    @pytest.mark.parametrize("orientation", [1, 2, 3, 4, 5, 6, 7, 8])
     def test_exif_save(
         self,
         tmp_path: Path,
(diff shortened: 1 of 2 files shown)


radarhere:
Would it be possible to add a test?

fdintino:
Oh, of course. Yes, I'll add one.

Tests/test_file_avif.py
```diff
@@ -327,6 +327,7 @@ def test_exif_save(
             if orientation == 1:
                 assert "exif" not in reloaded.info
             else:
+                assert reloaded.getexif()[274] == orientation
```
fdintino, line 330: This assertion fails before this change for orientations 6 and 8.

radarhere: We can include this new assertion if you like, but actually, the next line
```python
assert reloaded.info["exif"] == exif_data
```
already fails with the new parameters. This is because the next line is checking the full EXIF data, not just the orientation.

radarhere (approved):

Tests/test_file_avif.py
fdintino: I realize they're redundant, but it was a bit obscure identifying the mismatch with the bytes, so I thought this more specific assertion might be more diagnostic of failures

Issue #7983, opened 2024-04-16
Unable to open AVIF images


ponponon:
I have a picture here in avif format：https://img10.360buyimg.com/n7/jfs/t1/194600/2/37658/144533/652f464dF8dc6bd47/3aada6e1078054ae.jpg.avif

![图片](https://github.com/python-pillow/Pillow/assets/38725104/aeb7a6df-efba-4aa6-9989-7e39ce33a2db)

 
```shell
╭─ponponon@MBP13ARM ~/Pictures
╰─➤  mediainfo jd002.avif
General
Complete name                            : jd002.avif
Format                                   : avif
Codec ID                                 : avif (avif/mif1/miaf/MA1A)
File size                                : 11.4 KiB

Image
ID                                       : 1
Format                                   : av01
Codec ID                                 : av01
Width                                    : 220 pixels
Height                                   : 220 pixels
Color space                              : YUV
Bit depth                                : 8 bits
Stream size                              : 11.1 KiB (98%)
Color range                              : Full
Color primaries                          : BT.709
Transfer characteristics                 : sRGB/sYCC
Matrix coefficients                      : BT.601
Codec configuration box                  : av1C


╭─ponponon@MBP13ARM ~/Pictures
╰─➤  mediainfo --version
MediaInfo Command line,
MediaInfoLib - v24.03
```

Opening with pillow results in an error

```shell
In [1]: from PIL import Image

In [2]: 

In [2]: Image.open('/Users/ponponon/Pictures/jd002.avif')
---------------------------------------------------------------------------
UnidentifiedImageError                    Traceback (most recent call last)
Cell In[2], line 1
----> 1 Image.open('/Users/ponponon/Pictures/jd002.avif')

File ~/.local/share/virtualenvs/image_vector-8oWws01B/lib/python3.10/site-packages/PIL/Image.py:3186, in open(fp, mode, formats)
   3184 for message in accept_warnings:
   3185     warnings.warn(message)
-> 3186 raise UnidentifiedImageError(
   3187     "cannot identify image file %r" % (filename if filename else fp)
   3188 )

UnidentifiedImageError: cannot identify image file '/Users/ponponon/Pictures/jd002.avif'
```

```shell
╰─➤  pip show pillow                                                                                   
Name: Pillow
Version: 9.3.0
Summary: Python Imaging Library (Fork)
Home-page: https://python-pillow.org
Author: Alex Clark (PIL Fork Author)
Author-email: aclark@python-pillow.org
[...]

PR #8876, opened 2025-04-10
Do not include libavif in wheels

radarhere:
Suggested in https://github.com/python-pillow/Pillow/pull/8869#issuecomment-2791409240. Alternative to #8869 and #8858


aclark4life (approved):
I think I like this option the best, thanks @radarhere . I don't think we'll sort out whether or not to push the limits on wheel sizes in a way that feels satisfying anytime soon.

radarhere:
I've removed the release note changes from this PR, and created #8885 instead. That way, if there are any updates needed just to the release notes, GitHub Actions doesn't need to rebuild the wheels.

doublex:
That is unfortunate.

hugovk:
We'll revisit this for the next release, 11.3.0.

aclark4life:
@doublex Given that we now support libavif and have resolved #7983, do you view the "unfortunate" part having to do with not including it in the wheels? I think it would be unfortunate if we were unable to support libavif but now that we can, I want to address the [concerns](https://github.com/python-pillow/Pillow/issues/8856#issuecomment-2769827939) that [folks have raised](https://github.com/python-pillow/Pillow/issues/8856#issuecomment-2769778859) about the wheel size.

radarhere:
Pillow 11.3.0 has now been released with AVIF support in the wheels.

PR #8879, opened 2025-04-10
Fix qtables and quality scaling

Kyliroco:
When both a custom quantization table and a quality value are provided,
the quantization table should be scaled using the JPEG quality scaling factor.
If quality is not explicitly set, the default base quality of 50 is used to preserve the original table.
This ensures consistent behavior when saving JPEGs with custom qtables.

Before this correction, a quality set with a quantization table worked in reverse: the closer the quality was to 0, the less aggressive the quantization table was (made up of 1) and the closer it was to 100, the more it resembled the original table, whereas a standard jpeg table corresponds to a quality of 50. now the inserted table is modified by the quality parameter like a standard jpeg quantization table.


radarhere:
Would you be able to add a test that passes with your change, but fails without it?

> a standard jpeg table corresponds to a quality of 50

Could you link to documentation or similar describing this?

radarhere:
The test you added was to
> verify that specifying standard qtables and a quality is equivalent to specifying only a quality

You tested it with a value of 75. Testing this [further](https://github.com/radarhere/Pillow/commit/9f34d58f9f95ce17681f9a3f5612ef1252014877), I [find](https://github.com/radarhere/Pillow/actions/runs/14417536441/job/40435984361#step:11:133) that the results don't match for 0-23, do match for 24-100, but don't match if I don't specify quality at all. Any thoughts?

Also, in case you haven't seen it, and think it needs updating, here is our current documentation of the 'quality' parameter.
https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html#jpeg-saving

Screenshot 2025-04-12 at 4 54 31 pm

Kyliroco:
I found the problem, it came from the fact that force_baseline was set to true when the qtable was not given and to false when the qtable was given so I set it to true when the qtable and the quality were given and to false when only the qtable was given

radarhere:
> but don't match if I don't specify quality at all

Sorry, still wondering about this. If I [save with qtables and no quality, and compare that against no qtables and no quality](https://github.com/radarhere/Pillow/commit/f8afdd86f330c3b4a22f01ee9f6b0a4e7242ec4e), it doesn’t [match](https://github.com/radarhere/Pillow/actions/runs/14679187278/job/41199643325#step:11:2750). From the user’s perspective, why should that be a special case?

https://pillow--8879.org.readthedocs.build/en/8879/handbook/image-file-formats.html#jpeg-saving
> The image quality, on a scale from 0 (worst) to 95 (best), or the string keep. The default is 75.

Isn't the default value different when saving with qtables?

Kyliroco:
@radarhere do you have more questions?

radarhere:
@gofr as the author of the related #4989, I wonder if you had any thoughts on this?

gofr:
@radarhere I'm afraid I've been away from this too long to be of much help.

The initial commit looks reasonable. The failures you found, and the subsequent fix for them looks related to https://github.com/python-pillow/Pillow/pull/4918. Maybe my thoughts in there add some background if you hadn't already seen it.

radarhere:
Thanks.

We recently had #9024 report a test failure when using libjpeg, rather than libjpeg-turbo. Considering that, I've tested this branch with libjpeg, and the assertion does fail - https://github.com/radarhere/Pillow/actions/runs/15942707545/job/44972935266

I'm not saying this functionality **has** to work with libjpeg, but I would like to make sure it's been considered.

PR #8882, opened 2025-04-11
If pasting an image onto itself at a lower position, copy from bottom

### LINKED ISSUE #8881: `image.paste(image, box)` has a special semantic if both images are identical
I searched for a solution to get a high performance shift operation with Pillow (in-place, SIMD, executed on C-level) but have not found one.
It seems they all require one/multiple memory allocations or manual python loops.

I then realized `image.paste(image, box)` should do it. But the resulting image shifts they `y`-amount again and again.
My first guess was that I used the `box` parameter wrongly, but `image.paste(image.copy(), box)` works as expected.

I conclude that a check is missing if the image pasted in is the same like `self` or that it should get mentioned in the docs that in this case the function has a different semantic.

Repeatedly shifted:
```python
img = Image.open("some_random_image.jpg")
img.paste(img, box=(0, 10))
img.save("img_pasted.jpg")
```

Working as expected:
```python
img = Image.open("some_random_image.jpg")
img.paste(img.copy(), box=(0, 10))
img.save("img_pasted.jpg")
```

radarhere:
Resolves #8881

When pasting an image,
https://github.com/python-pillow/Pillow/blob/c8d98d56a02e0729f794546d6f270b3cea5baecf/src/libImaging/Paste.c#L47-L49
will copy the rows of the image from top to bottom.

However, if the image is pasting onto itself, but say, 10 pixels lower,
after those 10 rows, it will start reading from a part of the image that has already been written to. This results in a repeating effect.

```python
from PIL import Image
im.paste(im, (0, 10))
```
repeat

A solution for that situation is to instead copy the rows of the image from bottom to top.

copy

### DIFF
--- a/src/libImaging/Paste.c
+++ b/src/libImaging/Paste.c
@@ -44,8 +44,14 @@ paste(
 
     xsize *= pixelsize;
 
-    for (y = 0; y < ysize; y++) {
-        memcpy(imOut->image[y + dy] + dx, imIn->image[y + sy] + sx, xsize);
+    if (imOut == imIn && dy > sy) {
+        for (y = ysize - 1; y >= 0; y--) {
+            memcpy(imOut->image[y + dy] + dx, imIn->image[y + sy] + sx, xsize);
+        }
+    } else {
+        for (y = 0; y < ysize; y++) {
+            memcpy(imOut->image[y + dy] + dx, imIn->image[y + sy] + sx, xsize);
+        }
     }
 }
 
--- a/Tests/test_image_paste.py
+++ b/Tests/test_image_paste.py
@@ -124,6 +124,18 @@ class TestImagingPaste:
         im = im.crop((12, 23, im2.width + 12, im2.height + 23))
         assert_image_equal(im, im2)
 
+    @pytest.mark.parametrize("y", [10, -10])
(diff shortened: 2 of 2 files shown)


TS53:
It seems there is a different C-code that is called if a `mask` is available.
Maybe your fix works here, too. Have you tested it with a `mask`?

```python
mask = Image.new("1", img.size, 1) # mask without an effect just to make sure that also the mask is tested

img.paste(
    img,
    box=(0, 10),
    mask=mask,
)
img.save("img_pasted.png")
```

radarhere:
Ok, I've pushed a commit to cover masks as well.

src/libImaging/Paste.c
```diff
@@ -23,6 +23,18 @@
 
 #include "Imaging.h"
 
+#define PREPARE_PASTE_LOOP()        \
```
hugovk, line 26: Can this be a function instead?

radarhere: https://github.com/radarhere/Pillow/commit/be8209ef7d79bfb84ee573f455a6ae5c1111008c is what would be required. I don't think it's as neat, but let me know if you disagree.

hugovk: OK, let's go with this PR version.

PR #8885, opened 2025-04-12
Add 11.2.1 release notes

radarhere:
Suggestion for https://github.com/python-pillow/Pillow/issues/8722#issuecomment-2798690775


hugovk:
Thanks, I've updated this to:

* remove the release notes for the incomplete 11.2.0, as it never completed, it never happened
* move the things that would have been released in 11.2.0 into 11.2.1
* note in 11.2.1 that AVIF is supported but needs to be built from source
* mention at the top of the 11.2.1 release notes the reason why 11.2.0 doesn't exist

PR #8893, opened 2025-04-14
Move XV Thumbnails to read only section

aclark4life:

 * XV Thumbnails was under "Write only formats", moved to "Read only formats"


aclark4life:
Screenshot 2025-04-14 at 6 31 57 PM

PR #8896, opened 2025-04-15
Improve SgiImagePlugin test coverage

radarhere:
Looking at SgiImagePlugin, I see that an error for the incorrect number of bands will never be raised.
https://github.com/python-pillow/Pillow/blob/507fefbce4c3b8c46c6c21483d488884f1b5a8e1/src/PIL/SgiImagePlugin.py#L160-L176

Only L, RGB and RGBA modes are supported when saving.
https://github.com/python-pillow/Pillow/blob/507fefbce4c3b8c46c6c21483d488884f1b5a8e1/src/PIL/SgiImagePlugin.py#L137-L139

In the first code snippet, you can see that `z = len(im.mode)` would be set `z` to the correct values of 1 for L, 3 for RGB and 4 for RGBA. It is later set to 1 if `dim` is 1 or 2, but `dim` only has those values for L mode images, which should be 1 anyway.

So the error can be removed.

After some code simplification, this adds coverage for the lines missing at https://app.codecov.io/gh/python-pillow/Pillow/blob/main/src%2FPIL%2FSgiImagePlugin.py


src/PIL/SgiImagePlugin.py
```diff
@@ -156,24 +149,15 @@ def _save(im: Image.Image, fp: IO[bytes], filename: str | bytes) -> None:
...
-
-    # assert we've got the right number of bands.
-    if len(im.getbands()) != z:
-        msg = f"incorrect number of bands in SGI write: {z} vs {len(im.getbands())}"
-        raise ValueError(msg)
+    # Number of dimensions (x,y,z)
+    if im.mode == "L":
+        dim = 1 if y == 1 else 2
+    else:
+        dim = 3
```
hugovk: Shall we use more a descriptive name? (and below)
```suggestion
        dimensions = 1 if y == 1 else 2
    else:
        dimensions = 3
```

hugovk (approved):

src/PIL/SgiImagePlugin.py
radarhere: I've pushed a commit with `dimension`. It matches
https://github.com/python-pillow/Pillow/blob/1788ab7887beb80ee89cca9b853a391d7f026a4c/src/PIL/SgiImagePlugin.py#L73-L74

and the NAME in 'The Header' at https://paulbourke.net/dataformats/sgirgb/sgiversion.html

PR #8904, opened 2025-04-16
Updated harfbuzz to 11.1.0

radarhere:
harfbuzz 11.1.0 has been released - https://github.com/harfbuzz/harfbuzz/releases/tag/11.1.0


radarhere:
https://github.com/harfbuzz/harfbuzz/blob/main/CONFIG.md
> Make sure you build with your compiler's "optimize for size" option. On gcc this is -Os, and can be enabled by passing CXXFLAGS=-Os. On clang there is an even more extreme flag, -Oz. Meson also provides --buildtype=minsize for more convenience.

I've pushed a commit to change `--buildtype=release` to `--buildtype=minsize`.

hugovk (approved):
What sort of size difference does it make?

radarhere:
The artifacts from Wheels job in the first commit are 321.2mb.
The artifacts from Wheels job in the second commit are 304.5mb.

PR #8842, opened 2025-03-28
Add KDE Wayland support for ImageGrab

AdianKozlica:

PR #8912, opened 2025-04-18
Add support for Grim in Wayland sessions ImageGrab

AdianKozlica:
Adds support for Grim, a screenshot tool used by XFCE (Wayland session), and compositors such as Sway and Hyprland.


src/PIL/ImageGrab.py
```diff
@@ -91,6 +91,8 @@ def grab(
                 args = ["gnome-screenshot", "-f"]
             elif shutil.which("spectacle"):
                 args = ["spectacle", "-n", "-b", "-f", "-o"]
+            elif shutil.which("grim"):
+                args = ["grim"]
```
radarhere, line 95:
```suggestion
            elif shutil.which("grim"):
                args = ["grim"]
            elif shutil.which("spectacle"):
                args = ["spectacle", "-n", "-b", "-f", "-o"]
```
If there's no reason not to, let's make the choice behave alphabetically.

Tests/test_imagegrab.py
```diff
@@ -44,6 +44,7 @@ def test_grab_no_xcb(self) -> None:
             sys.platform not in ("win32", "darwin")
             and not shutil.which("gnome-screenshot")
             and not shutil.which("spectacle")
+            and not shutil.which("grim")
```
radarhere, line 47:
```suggestion
            and not shutil.which("grim")
            and not shutil.which("spectacle")
```

radarhere:
Just trying to think ahead - after #8842 added spectacle, and this adds grim, are there any more screenshot tools you plan to add?

radarhere (approved):

AdianKozlica:
The only one that remains is cosmic-screenshot, but COSMIC DE is still in an alpha state and from my testing it only seems to support specifying a custom save directory but not a custom output filename.

PR #8913, opened 2025-04-18
Add `make [-C docs] htmllive` to rebuild and reload HTML files

hugovk:
The existing `make livehtml` uses https://github.com/lepture/python-livereload and builds the docs and opens them in the browser, but if I edit an RST file, it doesn't rebuild the file and reload the site.

Let's replace it with `make htmllive` which uses https://github.com/sphinx-doc/sphinx-autobuild, and does the same as above, but also rebuilds changed RST files and reloads the browser.

This is what is used in CPython docs (and devguide and PEPs) and it's great to use. So I used the same `htmllive` target name because I'm used to typing it. I guess no-one has been using the old `livehtml` target in a while, as we've not noticed it working.

I added sphinx-autobuild to the list of optional docs dependencies. We _could_ only install this when doing a `livehtml` build, but it's easier to always install it. Happy to change if preferred.

Also deleted `docs/Guardfile`, I don't think it's needed?


docs/Makefile
```diff
@@ -201,9 +201,10 @@ doctest:
 htmlview: html
 	$(PYTHON) -c "import os, webbrowser; webbrowser.open('file://' + os.path.realpath('$(BUILDDIR)/html/index.html'))"
 
-.PHONY: livehtml
-livehtml: html
-	livereload $(BUILDDIR)/html -p 33233
+.PHONY: htmllive
+htmllive: SPHINXBUILD = sphinx-autobuild
```
radarhere, line 205:
```suggestion
htmllive: SPHINXBUILD = $(PYTHON) -m sphinx_autobuild
```

Makefile
```diff
@@ -43,6 +47,7 @@ help:
 	@echo "  docserve           run an HTTP server on the docs directory"
 	@echo "  html               make HTML docs"
 	@echo "  htmlview           open the index page built by the html target in your browser"
+	@echo "  htmllive           to rebuild and reload HTML files in your browser"
```
radarhere, line 50:
```suggestion
	@echo "  htmllive           rebuild and reload HTML files in your browser"
```
?

radarhere: I understand it's different from the docs/Makefile, but it's consistent with the rest of this Makefile?

radarhere (approved):

Makefile
```diff
@@ -43,6 +47,7 @@ help:
 	@echo "  docserve           run an HTTP server on the docs directory"
 	@echo "  html               make HTML docs"
 	@echo "  htmlview           open the index page built by the html target in your browser"
+	@echo "  htmllive           rebuild and reload HTML files in your browser"
```
radarhere, line 50:
```suggestion
	@echo "  htmllive           rebuild and reload HTML files in your browser"
	@echo "  htmlview           open the index page built by the html target in your browser"
```
This list is otherwise alphabetical.

docs/Makefile
```diff
@@ -20,8 +20,8 @@ help:
 	@echo "Please use \`make <target>' where <target> is one of"
 	@echo "  html       to make standalone HTML files"
 	@echo "  htmlview   to open the index page built by the html target in your browser"
+	@echo "  htmllive   to rebuild and reload HTML files in your browser"
```
radarhere, line 23:
```suggestion
	@echo "  htmllive   to rebuild and reload HTML files in your browser"
	@echo "  htmlview   to open the index page built by the html target in your browser"
```
If you, and you may not, make the other suggested change to be alphabetical, maybe this group of html* targets should be in the same order?

hugovk: Let's keep this as is -- it's going from simpler action to more complex.

And actually, I'll revert the change to the top-level `Makefile` too.

PR #8917, opened 2025-04-19
Docs: remove unused Makefile targets

hugovk:
I plan to do some more refactoring of `docs/Makefile`, but first, there's a number of targets that we're not using, and some of which aren't even working. Let's remove them. 

We're only really using `html`, but I left some other `html` variants in there like `dirhtml` and `singlehtml`. `linkcheck` can also be useful from time to time.

Also removed from `docs/make.bat` but I don't have Windows to test with, and probably won't many future changes to this file.

If someone now or subsequently says they use any of these, we can add them back.

The non-working targets:
* json
* pickle
* latexpdf (if no latexmk installed)
* info
* doctest


radarhere:
Here's my investigation of the non-working targets,

- ‘json’ and ‘pickle’ raise an error
```
sphinx.errors.ExtensionError: Handler <function html_page_context at 0x107d27920> for event 'html-page-context' threw an exception (exception: 'pagename')
```

- ‘latexpdf’ does indeed raise
```
latexmk: No such file or directory
```

- ‘info’ doesn’t raise an error. It does produce a number of warnings
```
WARNING: a suitable image for texinfo builder not found
warning: @footnote should not appear on @deffn line
warning: @subsubsection missing argument
```

- ‘doctest’ raises an error
```
sphinx.errors.SphinxError: Builder name doctest not registered or available through entry point
```

radarhere (approved):

PR #8918, opened 2025-04-21
Removed unreachable code

radarhere:
https://github.com/python-pillow/Pillow/blob/03e7871afdb8f558cddc10cc9013e1db143298d9/src/PIL/_binary.py#L21-L22
https://github.com/python-pillow/Pillow/blob/03e7871afdb8f558cddc10cc9013e1db143298d9/src/PIL/MpegImagePlugin.py#L31-L39

`c < 0` will never be True. The value can range from 0 to 255, but will never be less than zero.

### DIFF
--- a/src/PIL/MpegImagePlugin.py
+++ b/src/PIL/MpegImagePlugin.py
@@ -33,11 +33,7 @@ class BitStream:
 
     def peek(self, bits: int) -> int:
         while self.bits < bits:
-            c = self.next()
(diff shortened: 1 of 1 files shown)


Metallicow:
Ok. Fair enough.
Unless it is double masked on purpose, but
I wouldn't see much of an issue, other than a dead body. At which point, I'm not much of a facilitator of anything.

PR #8933, opened 2025-04-27
Refactor docs `Makefile`

hugovk:
Mainly to remove duplication.

Some functional changes:

* Use `--jobs auto` for all builders, not just `linkcheck`. This speeds `make -C docs clean html` for me from 25 seconds to 14 seconds.
* Use `--fail-on-warning --keep-going` for all builders, not just `html`.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -12,8 +12,6 @@ BUILDDIR      = _build
 PAPEROPT_a4     = --define latex_paper_size=a4
 PAPEROPT_letter = --define latex_paper_size=letter
 ALLSPHINXOPTS   = --doctree-dir $(BUILDDIR)/doctrees $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) .
-# the i18n builder cannot share the environment and doctrees with the others
-I18NSPHINXOPTS  = $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) .
 
 .PHONY: help
 help:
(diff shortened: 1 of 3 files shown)


docs/Makefile
```diff
@@ -3,17 +3,23 @@
...
+BUILDER       = html
+JOBS          = auto
+PAPER         =
 
 # Internal variables.
 PAPEROPT_a4     = --define latex_paper_size=a4
 PAPEROPT_letter = --define latex_paper_size=letter
-ALLSPHINXOPTS   = --doctree-dir $(BUILDDIR)/doctrees $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) .
-# the i18n builder cannot share the environment and doctrees with the others
-I18NSPHINXOPTS  = $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) .
```
radarhere, line 16: Ah, right, `I18NSPHINXOPTS` is unused due to #8917

docs/Makefile
```diff
@@ -3,17 +3,23 @@
 
 # You can set these variables from the command line.
 PYTHON        = python3
-SPHINXOPTS    =
 SPHINXBUILD   = $(PYTHON) -m sphinx.cmd.build
-PAPER         =
+SPHINXOPTS    = --fail-on-warning --keep-going
```
radarhere, line 7:
```suggestion
SPHINXOPTS    = --fail-on-warning
```

https://www.sphinx-doc.org/en/master/man/sphinx-build.html#cmdoption-sphinx-build-keep-going
> From Sphinx 8.1, --keep-going is always enabled.
> ...
> Changed in version 8.1: sphinx-build no longer exits on the first warning, meaning that in effect --fail-on-warning is always enabled. The option is retained for compatibility, but may be removed at some later date.

radarhere (approved):

PR #8953, opened 2025-05-12
Fix memory leak in arrow export using array structure

wiredfool:
Fixes #8950 .

Changes proposed in this pull request:

 * Properly handle releasing the image associated with PyArrow capsules that have child arrays.

Issue #8950, opened 2025-05-10
Memory leak when calling PIL.Image.Image.__arrow_c_array__()


Joshix-1:
### What did you do?

I'm calling the `PIL.Image.Image.__arrow_c_array__` method on images to access their data. 

### What did you expect to happen?
max rss should stay constant after a while

### What actually happened?
memory gets leaked

The output of the reproduction script below:
```
RSS:      88656
RSS:      88836
RSS:      88908
=== accessing __arrow_c_array__ ===
RSS:     395212
RSS:     713164
=== with arro3.core.Array ===
RSS:    1031376
RSS:    1349072
=== with pyarrow.array ===
RSS:    1667588
RSS:    1986052
```

### What are your OS, Python and Pillow versions?

* OS: Arch Linux
* Python: Python 3.13.3
* Pillow: 11.2.1

```text
$ python3 -m PIL --report
--------------------------------------------------------------------
Pillow 11.2.1
Python 3.13.3 (main, Apr  9 2025, 07:44:25) [GCC 14.2.1 20250207]
--------------------------------------------------------------------
Python executable is /home/josh/code/an-website/venv/bin/python3
Environment Python files loaded from /home/josh/code/an-website/venv
System Python files loaded from /usr
--------------------------------------------------------------------
Python Pillow modules loaded from /home/josh/code/an-website/venv/lib/python3.13/site-packages/PIL
Binary Pillow modules loaded from /home/josh/code/an-website/venv/lib/python3.13/site-packages/PIL
--------------------------------------------------------------------
--- PIL CORE support ok, compiled for 11.2.1
--- TKINTER support ok, loaded 8.6
--- FREETYPE2 support ok, loaded 2.13.3
--- LITTLECMS2 support ok, loaded 2.17
--- WEBP support ok, loaded 1.5.0
*** AVIF support not installed
--- JPEG support ok, compiled for libjpeg-turbo 3.1.0
--- OPENJPEG (JPEG2000) support ok, loaded 2.5.3
--- ZLIB (PNG/ZIP) support ok, loaded 1.3.1, compiled for zlib-ng 2.2.4
--- LIBTIFF support ok, loaded 4.7.0
--- RAQM (Bidirectional Text) support ok, loaded 0.10.1, fribidi 1.0.16, harfbuzz 11.0.1
*** LIBIMAGEQUANT (Quantization method) support not installed
--- XCB (X protocol) support ok
--------------------------------------------------------------------
```

### Reproduction script
[...]

PR #8954, opened 2025-05-13
Valgrind Memory Leak Checking

wiredfool:
* Fix for https://github.com/python-pillow/Pillow/issues/8950 . Arrow Export memory leak.
 * Add support in the Makefile to use valgrind to check for definite leaks.
 * Add some suppressions of python object allocations for leak checks. Python is a bit noisy for leak checking. 
 * Fix a leak in the Arrow schema where child schemas weren't released.
 * Fix a leak in webp encode on error
 * Fix a leak in UnsharpMask on error
 * Fix a leak in TiffEncode on error
 * Fix a leak in Font getmask on error
 * Fix a leak in JpegEncode on error

To Do:
* [x] Consider how to run this periodically in CI. This is significantly slower than plain valgrind -- locally I had a 4 hour run at one point. That's too slow for a merges or prs, but we need to see the results. 
* Consider setting the leak detection to something other than definite.  -- future work
* Review the python suppressions. It's definitely possible that some of them are leaks -- they'd be pointing to issues with our PyObject handling at the C level, either erroring from a function without decrefing or similar. -- future work

~Note -- this is built on the Arrow memory leak check PR #8953.~


aclark4life:
@wiredfool We can trigger a workflow manually or run daily independent of PRs.

Makefile
```diff
@@ -99,6 +99,13 @@ valgrind:
             --log-file=/tmp/valgrind-output \
             python3 -m pytest --no-memcheck -vv --valgrind --valgrind-log=/tmp/valgrind-output
 
+.PHONY: valgrind-leak
+valgrind-leak:
```
radarhere, line 103:
```suggestion
valgrind-leak:
	python3 -c "import pytest_valgrind" > /dev/null 2>&1 || python3 -m pip install pytest-valgrind
```
Install pytest-valgrind if it is missing, like the 'valgrind' target does

src/libImaging/Arrow.c
```diff
@@ -37,6 +37,10 @@ ReleaseExportedSchema(struct ArrowSchema *array) {
             child->release = NULL;
         }
         // UNDONE -- should I be releasing the children?
```
radarhere, line 39: Is this UNDONE comment resolved now?

aclark4life:
In that case, long PR it is! We typically have PRs queued for days, or longer, before merging.

wiredfool:
On My Machine:
```4711 passed, 345 skipped, 11 xfailed, 8 warnings in 15074.00s (4:11:14) ```

That's vs 87 minutes for ordinary valgrind on this pr in the GHA testing. 

Unfortunately, we're single threaded for valgrind tests.

**Comment by wiredfool** [comment]
Ok, We'll see how long this takes: https://github.com/python-pillow/Pillow/actions/runs/15054197221/job/42316085246?pr=8954

I'm only running this on the pull request, not the push. I'm expecting 5 hours. 

This is injecting the test script in the Pillow repo, instead of building a new copy of the valgrind image. At some point we can move that over, but I'd like to be tuning the supressions without having to build a new valgrind image, so it's probably better to leave it here for now.

**Comment by wiredfool** [comment]
Ok, so one more leak found, and a couple of timeouts that weren't xfailed in valgrind.  Total test time was 3 hours, which was a little better than I was expecting.  Guess core speed has gotten a little better than this machine. 


So what this whole exercise is showing is that we're reasonably good at dealing with the memory lifetime for the happy case, but there are a lot of cases where c-level error handling is returning without freeing the inflight items that would otherwise be passed out or handled properly. I suspect that there are more similar leaks out there -- the ones that we're hitting are the ones that are exercised in our test suite.  There are a couple of places here where I've added changes that aren't picked up by the code coverage but are essentially the same pattern as the leak. This indicates that we're going to have other error level leaks, and we'll probably find them with additional testing.

File: depends/docker-test-valgrind-memory.sh [hunk 3]
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+## Run this as the test script in the docker valgrind image.
+## Note -- can be included directly into the docker image,
+## but requires the currnet python.supp.
```
radarhere, line 5:
```suggestion
## but requires the current python.supp.
```

Makefile
```diff
@@ -100,10 +100,18 @@ test:
...
+	python3 -c "import pytest_valgrind" > /dev/null 2>&1 || python3 -m pip install pytest-valgrind
+	PILLOW_VALGRIND_TEST=true PYTHONMALLOC=malloc valgrind --suppressions=Tests/oss-fuzz/python.supp \
+	    --leak-check=full --show-leak-kinds=definite --errors-for-leak-kinds=definite \
+            --log-file=/tmp/valgrind-output \
+            python3 -m pytest -vv --valgrind --valgrind-log=/tmp/valgrind-output Tests/
```
radarhere, line 113:
```suggestion
            python3 -m pytest -vv --valgrind --valgrind-log=/tmp/valgrind-output
```

radarhere:
I think this is causing TIFF errors.

Since this was merged, there have been intermittent TIFF errors in GitHub Actions, from Test and Wheels.
- 'cannot identify image file' - https://github.com/python-pillow/Pillow/actions/runs/15371601710/job/43251458128
- 'attempt to seek outside sequence' - https://github.com/python-pillow/Pillow/actions/runs/15370070907/job/43247953625
- 'unrecognized data stream contents when reading image file' - https://github.com/python-pillow/Pillow/actions/runs/15370070907/job/43247953635

These all passed on subsequent attempts.

To check that it was this PR that changed things, I've run main before this merge 5 times in my fork with no errors in [Test](https://github.com/radarhere/Pillow/actions/runs/15375207832/job/43269467478) and [Wheels](https://github.com/radarhere/Pillow/actions/runs/15375211712/job/43280264161).

wiredfool:
That would likely be this: https://github.com/python-pillow/Pillow/pull/8954/commits/e2e40c54568236d2504921eb0b335cdab734a7d5  where the TiffClose was moved to a cleanup method so that it's run on all exits.

radarhere:
I've created #9002 as a possible solution.

radarhere:
https://github.com/python-pillow/Pillow/pull/9002 has now been merged.

PR #8962, opened 2025-05-21
Clear TIFF core image if memory mapping was used for last load

### LINKED ISSUE #8961: Segfault when seeking to position 0 from any position in a TIFF image and reading the numpy array
### What did you do?
In a multi-page TIFF 8-bit image seek back to index 0 from any index greater than 0.

The first index is read and the array can be accessed.

### What actually happened?
Process ends unexpectedly with signal 11:SIGSEGV

* OS: Ubuntu 24.04.2 LTS
* Python: 3.12.10
* Pillow: 11.x
* NumPy: 2.1.3

```text
Pillow 11.0.0
Python 3.12.10 (main, May 18 2025, 14:08:06) [GCC 13.3.0]
Python executable is /home/myuser/.pyenv/versions/pillow-segfault/bin/python3
Environment Python files loaded from /home/myuser/.pyenv/versions/pillow-segfault
System Python files loaded from /home/myuser/.pyenv/versions/3.12.10
Python Pillow modules loaded from /home/myuser/.pyenv/versions/pillow-segfault/lib/python3.12/site-packages/PIL
Binary Pillow modules loaded from /home/myuser/.pyenv/versions/pillow-segfault/lib/python3.12/site-packages/PIL
--- PIL CORE support ok, compiled for 11.0.0
--- JPEG support ok, compiled for libjpeg-turbo 3.0.4
--- ZLIB (PNG/ZIP) support ok, loaded 1.3
--- RAQM (Bidirectional Text) support ok, loaded 0.10.1, fribidi 1.0.13, harfbuzz 10.0.1

```

```python
#!/usr/bin/env python3
import numpy as np
from PIL import Image


def _get_np_array(img, seek_index, dtype):
    img.seek(seek_index)
    return np.asarray(img, dtype=dtype)


if __name__ == '__main__':
    image_path = './11-pages.tif'
    image_shape = (336, 256)

    pil_img = Image.open(image_path)
    assert pil_img.n_frames == 11

    array = _get_np_array(pil_img, 1, np.uint8)
    assert array.shape == (336, 256)

    array = _get_np_array(pil_img, 2, np.uint8)
    assert array.shape == (336, 256)

    array = _get_np_array(pil_img, 1, np.uint8)
    assert array.shape == (336, 256)

    ##########################################################################
    # this call will result in a seg_fault, when attempting to get the array #
    ##########################################################################
    array = _get_np_array(pil_img, 0, np.uint8)
    assert array.shape == (336, 256)
```

[11-pages.zip](https://github.com/user-attachments/files/20298063/11-pages.zip)

The provided example works in Pillow 10.x.

radarhere:
Resolves #8961

I've used the test image from the issue, but due to a larger file size, I've placed it in https://github.com/python-pillow/test-images/pull/7. This will fail without that pull request.

Using memory mapping to load one frame, and then not using it for the next frame, is causing a problem. This would be because memory mapping is replacing the core image.

https://github.com/python-pillow/Pillow/blob/3c71559804e661a5f727e2007a5be51f26d9af27/src/PIL/ImageFile.py#L324-L326

But TiffImagePlugin tries to keep using it afterwards.

https://github.com/python-pillow/Pillow/blob/3c71559804e661a5f727e2007a5be51f26d9af27/src/PIL/TiffImagePlugin.py#L1214-L1223

### DIFF
--- a/Tests/test_tiff_crashes.py
+++ b/Tests/test_tiff_crashes.py
@@ -52,3 +52,17 @@ def test_tiff_crashes(test_file: str) -> None:
         pytest.skip("test image not found")
     except OSError:
         pass
+
+
+def test_tiff_mmap() -> None:
+    try:
+        with Image.open("Tests/images/crash_mmap.tif") as im:
+            im.seek(1)
+            im.load()
+
+            im.seek(0)
+            im.load()
+    except FileNotFoundError:
+        if on_ci():
+            raise
+        pytest.skip("test image not found")
--- a/src/PIL/TiffImagePlugin.py
+++ b/src/PIL/TiffImagePlugin.py
@@ -1217,9 +1217,10 @@ class TiffImageFile(ImageFile.ImageFile):
             return
         self._seek(frame)
         if self._im is not None and (
-            self.im.size != self._tile_size or self.im.mode != self.mode
+            self.im.size != self._tile_size
+            or self.im.mode != self.mode
+            or self.readonly
         ):
-            # The core image will no longer be used
             self._im = None
 
     def _seek(self, frame: int) -> None:
--- a/src/map.c
+++ b/src/map.c
@@ -137,6 +137,7 @@ PyImaging_MapBuffer(PyObject *self, PyObject *args) {
         }
     }
 
+    im->read_only = view.readonly;
     im->destroy = mapping_destroy_buffer;
 
     Py_INCREF(target);


wiredfool:
Does it make sense to set the c level read-only flag on the image if it's memmapped, and then check if the image is read-only instead?  Would the mmap also cause an issue if someone were to try to imagedraw on it?

wiredfool (approved):

PR #8966, opened 2025-05-21
Set tag type accordingly if IFDRational with 0 denominator

### LINKED ISSUE #8965: ImageOps.exif_transpose() raises struct.error on invalid EXIF rational
### What did you do?

Attempted to process an image with ImageOps.exif_transpose(), which contains an EXIF tag with a value of (-1, 0).

The image should be processed without raising an error, with invalid EXIF values (like (-1, 0)) handled gracefully as float('nan') as a result of (1, 0).

### What actually happened?

```
>>> from PIL import Image, ImageOps
>>> img = Image.open('/path/to/file.jpg')
>>> ImageOps.exif_transpose(img)
  File "/home/user/venv_pillow/lib/python3.12/site-packages/PIL/ImageOps.py", line 712, in exif_transpose
    exif_image.info["exif"] = exif.tobytes()
                              ^^^^^^^^^^^^^^
  File "/home/user/venv_pillow/lib/python3.12/site-packages/PIL/Image.py", line 4043, in tobytes
    return b"Exif\x00\x00" + head + ifd.tobytes(offset)
                                    ^^^^^^^^^^^^^^^^^^^
  File "/home/user/venv_pillow/lib/python3.12/site-packages/PIL/TiffImagePlugin.py", line 983, in tobytes
    data = ifd.tobytes(offset)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/user/venv_pillow/lib/python3.12/site-packages/PIL/TiffImagePlugin.py", line 986, in tobytes
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/venv_pillow/lib/python3.12/site-packages/PIL/TiffImagePlugin.py", line 841, in write_rational
    return b"".join(
           ^^^^^^^^^
  File "/home/user/venv_pillow/lib/python3.12/site-packages/PIL/TiffImagePlugin.py", line 842, in <genexpr>
    self._pack("2L", *_limit_rational(frac, 2**32 - 1)) for frac in values
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/venv_pillow/lib/python3.12/site-packages/PIL/TiffImagePlugin.py", line 782, in _pack
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
struct.error: 'L' format requires 0 <= number <= 4294967295

```

* OS: Ubuntu 22.04
* Python: 3.12.3
* Pillow: 11.1.0

```text
(venv_pillow)  user@thinkpad  ~  python3 -m PIL.report
Pillow 11.1.0
Python 3.12.3 (main, Feb  4 2025, 14:48:35) [GCC 13.3.0]
Python executable is /home/user/venv_pillow/bin/python3
Environment Python files loaded from /home/user/venv_pillow
Python Pillow modules loaded from /home/user/venv_pillow/lib/python3.12/site-packages/PIL
Binary Pillow modules loaded from /home/user/venv_pillow/lib/python3.12/site-packages/PIL
--- ZLIB (PNG/ZIP) support ok, loaded 1.3, compiled for zlib-ng 2.2.2
--- RAQM (Bidirectional Text) support ok, loaded 0.10.1, fribidi 1.0.13, harfbuzz 10.1.0

```

epou:
Fixes #8965. Alternative to #8970 and #8978. For a comparison, see [this comment.](https://github.com/python-pillow/Pillow/issues/8965#issuecomment-2911906401)

 * override `__float__` method in IFDRational
 * Setting tagtype to TiffTags.SIGNED_RATIONAL if IFDRational is nan but the numerator is negative.

After investigating the issue, I found that the EXIF tag contains the value (-1, 0), which results in `float('nan')` via the IFDRational class. This was previously handled more gracefully, but starting from commit [5ff2027](https://github.com/python-pillow/Pillow/commit/5ff20273d90c1be49a9662cf6d4c53c9bd53aadb) (included in Pillow 11.1.0), all NaN IFDRational values are considered as TiffTags.RATIONAL.

This causes a problem when the writer attempts to serialize these values using write_rational, which fails because `struct.pack("2L", ...)` cannot handle negative values. In such cases, write_signed_rational should be used instead.


epou:
Clarification regarding:
> This was previously handled more gracefully, but starting from commit 5ff2027 (included in Pillow 11.1.0), all NaN IFDRational values are considered as TiffTags.RATIONAL.

This change in behavior is due to how `float('nan')` operates in Python:
```python
>>> float('nan') == float('nan')
False
>>> float('nan') > 0
False
>>> float('nan') < 0
False
>>> float('nan') == 0
False
```

Before commit 5ff2027, the logic used `all(v >= 0 for v in values)` to determine the tag type. Since any comparison involving float('nan') returns False, the presence of a NaN caused the expression to evaluate to False, resulting in the tag being set to TiffTags.SIGNED_RATIONAL:
```python
                    self.tagtype[tag] = (
                        TiffTags.RATIONAL
                        if all(v >= 0 for v in values)
                        else TiffTags.SIGNED_RATIONAL
                    )
```

After the commit, the logic was changed to inspect values individually:
```python
                    for v in values:
                        assert isinstance(v, IFDRational)
                        if v < 0:
                            self.tagtype[tag] = TiffTags.SIGNED_RATIONAL
                            break
                        else:
                            self.tagtype[tag] = TiffTags.RATIONAL
```
This change means that when a value is `float('nan')`, it does not satisfy the `v < 0` condition (nor would it satisfy any other typical comparison), so the tag is now set to TiffTags.RATIONAL by default

hugovk:
Closed in favour of https://github.com/python-pillow/Pillow/pull/8970.

PR #8970, opened 2025-05-22
Allow for IFDRational with negative numerator and zero denominator

### LINKED ISSUE #8965: ImageOps.exif_transpose() raises struct.error on invalid EXIF rational

radarhere:
Resolves #8965. Alternative to #8966 and #8978. For a comparison, see [this comment.](https://github.com/python-pillow/Pillow/issues/8965#issuecomment-2911906401)

A user has found an image where the [ExposureBiasValue](https://web.archive.org/web/20181105221156/https://www.awaresystems.be/imaging/tiff/tifftags/privateifd/exif/exposurebiasvalue.html) signed rational TIFF tag has a negative numerator but a zero denominator.

When this is passed to our IFDRational,  it is considered `nan`,
https://github.com/python-pillow/Pillow/blob/3c71559804e661a5f727e2007a5be51f26d9af27/src/PIL/TiffImagePlugin.py#L374-L375
which means it is not less than zero.
```pycon
>>> IFDRational(-1, 0) 
nan
>>> IFDRational(-1, 0) < 0
False
```

This means that it isn't detected as something to write as a signed rational.
https://github.com/python-pillow/Pillow/blob/3c71559804e661a5f727e2007a5be51f26d9af27/src/PIL/TiffImagePlugin.py#L690-L692

This PR suggests replacing `v < 0` with `(float(v.numerator) < 0) != (float(v.denominator) < 0)`


epou:
The proposed solution is functionally correct, but it requires developers to manually check the numerator and denominator each time. This introduces potential for inconsistency or error, especially if this behaviour needs to be checked on other parts of the code in the future.

It seems the underlying issue, is that the code assumes IFDRational instances behave like floats and are directly comparable. That expectation is natural and intuitive, and likely why the bug slipped in when the previous change was committed. This made me think that it'd make more sense to align the behavior of IFDRational more closely with float semantics. A more mathematically consistent handling of division-by-zero cases would be:
- Positive numerator / zero denominator → `math.inf`
- Negative numerator / zero denominator → `-math.inf`
- Zero numerator / zero denominator → 0 (or we could keep using `float('nan')`)

If we go this route, and implement `__float__` as I did in the other pull request (https://github.com/python-pillow/Pillow/pull/8966/files#diff-6ad43f85f1a075181d4d8cfcd97ae27bba1eccf5c3db5a3457160f98218eba6eR405-R407), the comparisons `v < 0` made for setting the tag type would behave correctly. This change would also make IFDRational instances more intuitive to work with and directly comparable using standard float logic.

Does this approach make sense to you? I’d be happy to open a PR implementing the -inf/inf logic so you can review it in context.

radarhere:
Ok, my next suggestion is `v < IFDRational(0)`. I've updated this PR.

I don't object to your idea, but Pillow tends to value backwards compatibility. The `nan` value is also explicitly described in the docstring. 
https://github.com/python-pillow/Pillow/blob/7e4d8e2f5557aa3bfe80c7d6376e9bd16ab08390/src/PIL/TiffImagePlugin.py#L335-L343
So there might be a preference to leave that as it is.

I have to imagine that it is rare for IFDRationals with zero denominators to be checked for their sign, simply because it's taken this long for an issue to be raised about it. And since you discovered this problem from Pillow's internal use of the class, I have to imagine that external checks are rarer still.

So, feel free to create a pull request with your suggestion, and see what others think. If you haven't thought of it already, another possibility would be changing `IFDRational`'s `__lt__` and `__gt__`.

epou:
I understand your point about backward compatibility and the existing docstring documentation for `nan`.

I’ve gone ahead and opened a new PR implementing the math.inf / -math.inf logic for zero denominators, as said before: https://github.com/python-pillow/Pillow/pull/8978. Would be great if you could take a look when you have a chance.

PR #8975, opened 2025-05-24
Run slow tests on valgrind, but without timeout

radarhere:
It would seem good to me to still let valgrind run its checks if we can. So rather than skipping them completely, run them, but don't timeout.


wiredfool:
If this works, we should just do this everywhere we're using `pytest.mark.timeout`.

radarhere:
Ok, I've pushed a commit for that.

PR #8979, opened 2025-05-27
Fix saving MPO with more than one appended image

### LINKED ISSUE #8977: MPO encoding then decoding does not work
### What did you do?

I encoded a list of image into MPO and decoded them. Decoding failed to return multiple images.

The MPO plugin should be able to decode the MPO encoded by its own.

### What actually happened?

Decoder complains that the file is truncated.

* OS: Linux
* Python: 3.10.12
* Pillow: 11.2.1 (I also tried 9.5.0, 10.4.0. Same error)

```python
import io
import PIL.Image

frames = [PIL.Image.new("RGB", (512, 512), c) for c in ["red", "green", "blue"]]
with io.BytesIO() as f:
  frames[0].save(
      f, format="MPO", save_all=True, append_images=frames[1:]
  )
  mpo_bytes = f.getvalue()
with io.BytesIO(mpo_bytes) as f:
  mpo_image = PIL.Image.open(f)
  print(mpo_image.n_frames)
```

Logs:

```python
> /usr/local/lib/python3.10/dist-packages/PIL/TiffImagePlugin.py:950: UserWarning: Truncated File Read
>  warnings.warn(str(msg))
> /usr/local/lib/python3.10/dist-packages/PIL/JpegImagePlugin.py:890: UserWarning: Image appears to be a malformed MPO file, it will be interpreted as a base JPEG file
>  warnings.warn(
>
> [/usr/local/lib/python3.10/dist-packages/PIL/JpegImagePlugin.py](https://localhost:8080/#) in __getattr__(self, name)
>    398             deprecate(name, 12)
>    399             return getattr(self, "_" + name)
> --> 400         raise AttributeError(name)
>    401 
>    402     def __getstate__(self) -> list[Any]:
>
> AttributeError: n_frames
```

The same works for GIF, but fails for MPO.

radarhere:
Resolves #8977

With each new frame that is saved to an MPO image, an offset is added
https://github.com/python-pillow/Pillow/blob/086e05f42ff459747c2ceb0e8e046f4b54e4d779/src/PIL/MpoImagePlugin.py#L65-L68
leading to a longer `mpentries`
https://github.com/python-pillow/Pillow/blob/086e05f42ff459747c2ceb0e8e046f4b54e4d779/src/PIL/MpoImagePlugin.py#L81
meaning the IFD becomes longer
https://github.com/python-pillow/Pillow/blob/086e05f42ff459747c2ceb0e8e046f4b54e4d779/src/PIL/MpoImagePlugin.py#L85

We rewind and replace the placeholder data with the IFD
https://github.com/python-pillow/Pillow/blob/086e05f42ff459747c2ceb0e8e046f4b54e4d779/src/PIL/MpoImagePlugin.py#L87-L89

But the placeholder data only ever has the same number of bytes, regardless of the number of frames.
https://github.com/python-pillow/Pillow/blob/086e05f42ff459747c2ceb0e8e046f4b54e4d779/src/PIL/MpoImagePlugin.py#L55

This PR increases the length of the placeholder data as the number of frames increase.


Tests/test_file_mpo.py
```diff
@@ -293,16 +293,18 @@ def test_save_all() -> None:
...
-
-    assert_image_equal(im, im_reloaded)
-    assert isinstance(im_reloaded, MpoImagePlugin.MpoImageFile)
-    assert im_reloaded.mpinfo is not None
-    assert im_reloaded.mpinfo[45056] == b"0100"
-
-    im_reloaded.seek(1)
-    assert_image_similar(im2, im_reloaded, 1)
+    for colors in (("#f00",), ("#f00", "#0f0")):
+        append_images = (Image.new("RGB", (1, 1), color) for color in colors)
```
hugovk:
```suggestion
        append_images = [Image.new("RGB", (1, 1), color) for color in colors]
```
This generator is being consumed in the line below, so we have nothing to iterate over in the `for` loop -- hence no coverage. Let's use a list.

radarhere: Thanks for catching that!

src/PIL/MpoImagePlugin.py
```diff
@@ -47,12 +46,20 @@ def _save_all(im: Image.Image, fp: IO[bytes], filename: str | bytes) -> None:
 
     mpf_offset = 28
     offsets: list[int] = []
-    for imSequence in itertools.chain([im], append_images):
+    total = 0
+    imSequences = [im] + list(append_images)
+    for imSequence in imSequences:
+        total += getattr(imSequence, "n_frames", 1)
+    for imSequence in imSequences:
         for im_frame in ImageSequence.Iterator(imSequence):
```
hugovk: Snake case, list unpacking and generator expression:

```suggestion
    im_sequences = [im, *append_images]
    total = sum(getattr(seq, "n_frames", 1) for seq in im_sequences)
    for im_sequence in im_sequences:
        for im_frame in ImageSequence.Iterator(im_sequence):
```

hugovk:
Thanks!

PR #8984, opened 2025-05-29
Use mask in C when drawing wide polygon lines

### LINKED ISSUE #8976: ImageDraw.polygon is very slow when width > 1
### What did you do?

I'm generating maps from shapefiles. The full code is a bit too much to repeat here, and a reproducible sample is also tricky because of all the data needed. If desired I can produce a POC zip.

This is a cut-down version of the code:

```python
img = Image.new('RGB', (self._view.width * RESIZE_FACTOR,
                        self._view.height * RESIZE_FACTOR),
                self._background.as_int_tuple(255))
draw = ImageDraw.Draw(img)

for layer in self._layers:
    # stuff to compute parameters
    draw.polygon(coords2, outline = lc, width = lw, fill = fc)

# resize to smooth image
if RESIZE_FACTOR != 1:
    img = img.resize((int(img.width / RESIZE_FACTOR),
                      int(img.height / RESIZE_FACTOR)),
                     resample = Image.Resampling.LANCZOS)

img.save(filename, 'PNG')
```

When `lw` above (line width) is 1 this runs in 4 seconds. When it's 3 the code takes 5 minutes. That seems ... unnecessary?

Of course, the `ImageDraw.polygon` method is being worked hard here, but if it can do the thin line in 4 seconds ...

* OS: MacOS 12.7.1
* Python: 3.13.3
* Pillow: 11.2.1 from pip

```text
Pillow 11.2.1
Python 3.13.3 (main, May 20 2025, 08:52:01) [Clang 14.0.0 (clang-1400.0.29.202)]
Python executable is /usr/local/opt/python@3.13/bin/python3.13
System Python files loaded from /usr/local/opt/python@3.13/Frameworks/Python.framework/Versions/3.13
Python Pillow modules loaded from /usr/local/lib/python3.13/site-packages/PIL
Binary Pillow modules loaded from /usr/local/lib/python3.13/site-packages/PIL
--- RAQM (Bidirectional Text) support ok, loaded 0.10.1, fribidi 1.0.16, harfbuzz 11.0.1
```

radarhere:
Resolves #8976

If we just drew thicker lines when creating a polygon, then since the centre of the line is the edge, half of each line would be drawn outside the filled polygon. To avoid this, #5694 introduced masking using the Python API.

https://github.com/python-pillow/Pillow/blob/5a04b9581b16a7f1e1109f1e31a206a6550f314c/src/PIL/ImageDraw.py#L361-L383

However, #8976 finds this to be slow. So instead of
- drawing a filled polygon
- drawing the outline
- combining those two to create a mask of just the outline inside the polygon
- drawing the outline again, and pasting it onto the final image with that mask

this PR
- draws a filled polygon
- draws the outline onto the final image, passing the filled polygon as a mask into the C layer. Every time a pixel is to be 'drawn', the mask image is checked to see whether or not to do so.

The first two commits simplify the existing code by removing `polygon` from the DRAW struct.

https://github.com/python-pillow/Pillow/blob/5a04b9581b16a7f1e1109f1e31a206a6550f314c/src/libImaging/Draw.c#L640-L645

This just means that `polygon_generic` can be used more directly, and the mask doesn't need to be passed through [each different type of polygon function.](https://github.com/python-pillow/Pillow/blob/5a04b9581b16a7f1e1109f1e31a206a6550f314c/src/libImaging/Draw.c#L593-L606)


larsga:
Nice! Thank you! 👍

PR #8990, opened 2025-05-31
Add parallel compile from pybind11

### LINKED ISSUE #8971: Parallel build is no longer parallel
### What did you do?

`make debug` or `make install`

Expect compilation to happen across all the cores, taking ~ 2-3 seconds.

### What actually happened?

Single threaded, taking ~15 seconds. There are something like 60 compilation units, and I have at least 8 cores.

* OS: Mac + Linux
* Python: 3.12
* Pillow: 11.2, main

Sometime between ~6 years ago and now, probably when the setuptools migrated around or we went to the newer build tool specifications, the parallel option here: https://github.com/python-pillow/Pillow/blob/main/setup.py#L381 stopped working. 

I can confirm that the code there is running, `self.parallel` is set correctly, but setuptools is apparently not using it anymore.

wiredfool:
Fixes #8971  .

 * Use the ParallelCompile from pybind11 to compile the c extension in parallel. 
 
With:
```console
$ make clean && time make install
...
real	0m9.226s
user	0m33.522s
sys	0m3.639s

$ make clean && time make debug
...
real	0m4.724s
user	0m15.480s
sys	0m2.941s
```

On main:
```console
$ make clean && time make install
...
real	0m21.224s
user	0m20.892s
sys	0m2.613s

$ make clean && time make debug
...
real	0m10.243s
user	0m9.450s
sys	0m2.123s
```


setup.py
```diff
@@ -18,9 +18,12 @@
 from collections.abc import Iterator
 from typing import Any
 
+from pybind11.setup_helpers import ParallelCompile
 from setuptools import Extension, setup
 from setuptools.command.build_ext import build_ext
 
+ParallelCompile("MAX_CONCURRENCY", default=0).install()
```
```suggestion
ParallelCompile("MAX_CONCURRENCY").install()
```
Looking at https://github.com/pybind/pybind11/blob/c2b32b1e3b1d972186ff745c9190cad664c219bb/pybind11/setup_helpers.py#L410-L413, `default` is already zero.

setup.py
```diff
@@ -18,9 +18,12 @@
 from collections.abc import Iterator
 from typing import Any
 
+from pybind11.setup_helpers import ParallelCompile
 from setuptools import Extension, setup
 from setuptools.command.build_ext import build_ext
 
+ParallelCompile("MAX_CONCURRENCY").install()
```
```suggestion
ParallelCompile("MAX_CONCURRENCY", default=4).install()
```
Actually, from our documentation, this should be 4.

https://pillow.readthedocs.io/en/stable/installation/building-from-source.html#build-options
> Config setting: -C parallel=n. Can also be given with environment variable: MAX_CONCURRENCY=n. Pillow can use multiprocessing to build the extension. Setting -C parallel=n sets the number of CPUs to use to n, or can disable parallel building by using a setting of 1. By default, it uses 4 CPUs, or if 4 are not available, as many as are present.

radarhere: I've created https://github.com/wiredfool/Pillow/pull/12 to use `-C parallel=n` as well.

hugovk: From [pybind11 docs](https://pybind11.readthedocs.io/en/stable/compiling.html#modules-with-setuptools):

> You can also pass `default=N` to set the default number of threads (0 will take the number of threads available) and `max=N`, the maximum number of threads; if you have a large extension you may want set this to a memory dependent number.

If not set, [pybind11 uses 0](https://github.com/pybind/pybind11/blob/4dc4aca2e120180e6bb48c076986feff41b4f5fb/pybind11/setup_helpers.py#L413).

Pillow's limit of 4 was added in 2014 (https://github.com/python-pillow/Pillow/pull/1043) but machines have more CPUs today.

Last I looked, the lowest amount on a macOS on the market was 8, and Linux laptops/desktops usually have at least around this many. Servers and shared systems may be lower.

Shall we leave the default to 0 (and update the docs)?

radarhere: Ok, I've pushed a commit.

radarhere (approved):

PR #8992, opened 2025-06-02
Revert "Fix memory leak in TiffEncode"

wiredfool:
This reverts commit e2e40c54568236d2504921eb0b335cdab734a7d5.

Fixes https://github.com/python-pillow/Pillow/pull/8954#issuecomment-2929027106 .

 * This reverts moving the cleanup for tiff encode into the cleanup function, as opposed to doing it in the actual encode
 * This is an intermittent issue on several of the platforms, it reasonably triggers using `pytest -v --random-order Tests/test_file_libtiff.py` on pypy3.10 on linux-x64
 * There's likely something else going on here, because the errors are on read and the changes are for write.

### DIFF
--- a/src/encode.c
+++ b/src/encode.c
@@ -703,8 +703,6 @@ PyImaging_LibTiffEncoderNew(PyObject *self, PyObject *args) {
         return NULL;
     }
 
-    encoder->cleanup = ImagingLibTiffEncodeCleanup;
-
     num_core_tags = sizeof(core_tags) / sizeof(int);
     for (pos = 0; pos < tags_size; pos++) {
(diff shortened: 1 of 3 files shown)


radarhere:
I've created #9002 as a possible alternative.

wiredfool:
I saw that, I haven't had a chance to test it yet. Probably Friday.

radarhere:
In the absence of any further discussion, I'll merge #9002 sometime in the next few days.

We're coming up on the release, and I would rather not merge it at the last minute.

radarhere:
#9002 has ben merged.

PR #8995, opened 2025-06-03
Deprecate ImageCmsProfile product_name and product_info

lukegb:
The _set method is no longer necessary, since we no longer compute any attributes from the profile. In most cases, we only set the profile, and in only one branch do we set the filename to anything non-None.

product_name/product_info were set to None at some point during what appears to be a batch of changes for Python 3 compatibility (ce041fd1995fe95de86804c83d100ed7d8ecdb3b), and never set back. Given this, let's deprecate these and schedule them for removal in Pillow 13.

 * Add attribute typing to ImageCmsProfile
 * Deprecate ImageCms.ImageCmsProfile.product_name, ImageCms.ImageCmsProfile.product_info
 * Deprecate ImageCms.ImageCmsProfile._set


src/PIL/ImageCms.py
```diff
@@ -256,22 +260,40 @@ def __init__(self, profile: str | SupportsRead[bytes] | core.CmsProfile) -> None
...
             msg = "Invalid type for Profile"  # type: ignore[unreachable]
             raise TypeError(msg)
 
+    def __getattr__(self, attr: str) -> Any:
+        if attr in ("product_name", "product_info"):
+            deprecate(
+                f"PIL.ImageCmsProfile.{attr}",
+                13,
+                action=(
+                    f"Use PIL.ImageCmsProfile.profile.{attr} instead. "
```
radarhere:
```suggestion
                f"ImageCms.ImageCmsProfile.{attr}",
                13,
                action=(
                    f"Use ImageCms.ImageCmsProfile.profile.{attr} instead. "
```
Feel free to argue for `f"PIL.ImageCms.imageCmsProfile.{attr}"` instead if you want, but I think we can all agree that `f"PIL.ImageCmsProfile.{attr}"` is not quite right.

lukegb: Whoops, you're right. Thanks!

src/PIL/ImageCms.py
```diff
@@ -256,22 +260,40 @@ def __init__(self, profile: str | SupportsRead[bytes] | core.CmsProfile) -> None
...
+            deprecate(
+                f"ImageCms.ImageCmsProfile.{attr}",
+                13,
+                action=(
+                    f"Use ImageCms.ImageCmsProfile.profile.{attr} instead. "
+                    f"Note that {attr} has been set to 'None' since Pillow 2.3.0."
+                ),
+            )
+            return None
+        msg = f"'{self.__class__.__name__}' has no attribute '{attr}'"
```
radarhere:
```suggestion
    def __getattr__(self, name: str) -> Any:
        if name in ("product_name", "product_info"):
            deprecate(
                f"ImageCms.ImageCmsProfile.{name}",
                13,
                action=(
                    f"Use ImageCms.ImageCmsProfile.profile.{name} instead. "
                    f"Note that {name} has been set to 'None' since Pillow 2.3.0."
                ),
            )
            return None
        msg = f"'{self.__class__.__name__}' has no attribute '{name}'"
```
Consistency with https://github.com/python-pillow/Pillow/blob/1bf32ae8923431da34a125179108c9403666f77b/src/PIL/JpegImagePlugin.py#L396 and other uses.

docs/deprecations.rst
```diff
@@ -193,6 +193,29 @@ Image.Image.get_child_images()
...
+
+ImageCms.ImageCmsProfile.product_name and .product_info
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. deprecated:: 11.3.0
+
+``ImageCms.ImageCmsProfile.product_name`` and the corresponding
+``.product_info`` attributes have been deprecated, and will be removed in
+Pillow 13 (2026-10-15). These attributes can be accessed on the ``.profile``
+attribute of ``ImageCmsProfile`` instead.
```
radarhere: I'm not following this.
```pycon
>>> from PIL import ImageCms
>>> profile = ImageCms.ImageCmsProfile(ImageCms.createProfile("sRGB"))
>>> profile.profile.product_name
AttributeError: 'PIL.ImageCms.core.CmsProfile' object has no attribute 'product_name'
```

hugovk:
> * Deprecate ImageCms.ImageCmsProfile._set

This is an undocumented underscore method, so not part of the public API and can be changed/removed without deprecation.

https://pillow.readthedocs.io/en/stable/reference/ImageCms.html

radarhere:
I've created https://github.com/python-pillow/Pillow/pull/9032, to move forward with part of this at least.

docs/deprecations.rst
```diff
@@ -59,6 +59,20 @@ another mode before saving::
...
+.. deprecated:: 12.0.0
+
+``ImageCms.ImageCmsProfile.product_name`` and the corresponding
+``.product_info`` attributes have been deprecated, and will be removed in
+Pillow 13 (2026-10-15). These attributes can be accessed on the ``.profile``
+attribute of ``ImageCmsProfile`` instead.
+
+Note that ``.product_name`` and ``.product_info`` have been set to ``None`` on
+``ImageCmsProfile`` since Pillow 2.3.0 (2014-01-01), so any working code that
+makes use of this data will already access it on ``.profile``.
```
radarhere:
```suggestion
Pillow 13 (2026-10-15). They have been set to ``None`` since Pillow 2.3.0.
```

If these attributes don't have a direct mapping,
```pycon
>>> from PIL import ImageCms
>>> profile = ImageCms.ImageCmsProfile(ImageCms.createProfile("sRGB"))
>>> profile.profile.product_name
AttributeError: 'PIL.ImageCms.core.CmsProfile' object has no attribute 'product_name'
```
then I think it is sufficient to just explain that they've been `None` for quite some time.

src/PIL/ImageCms.py
```diff
@@ -256,6 +258,20 @@ def __init__(self, profile: str | SupportsRead[bytes] | core.CmsProfile) -> None
...
+    def __getattr__(self, name: str) -> Any:
+        if name in ("product_name", "product_info"):
+            deprecate(
+                f"ImageCms.ImageCmsProfile.{name}",
+                13,
+                action=(
+                    f"Use ImageCms.ImageCmsProfile.profile.{name} instead. "
+                    f"Note that {name} has been set to 'None' since Pillow 2.3.0."
+                ),
+            )
```
radarhere:
```suggestion
            deprecate(f"ImageCms.ImageCmsProfile.{name}", 13)
```

docs/deprecations.rst
radarhere: I've pushed commits to remove the replacement suggestion.

src/PIL/ImageCms.py
```diff
@@ -256,6 +258,13 @@ def __init__(self, profile: str | SupportsRead[bytes] | core.CmsProfile) -> None
             msg = "Invalid type for Profile"  # type: ignore[unreachable]
             raise TypeError(msg)
 
+    def __getattr__(self, name: str) -> Any:
+        if name in ("product_name", "product_info"):
+            deprecate(f"ImageCms.ImageCmsProfile.{name}", 13)
+            return None
+        msg = f"'{self.__class__.__name__}' has no attribute '{name}'"
```
radarhere:
```suggestion
        msg = f"'{self.__class__.__name__}' object has no attribute '{name}'"
```

src/PIL/ImageCms.py
```diff
@@ -247,7 +246,7 @@ def __init__(self, profile: str | SupportsRead[bytes] | core.CmsProfile) -> None
                         self.profile = core.profile_frombytes(f.read())
                     return
             self.filename = profile
-            self.profile = core.profile_open(profile)
+            self.profile: core.CmsProfile = core.profile_open(profile)
```
radarhere:
```suggestion
            self.profile = core.profile_open(profile)
```
Testing with PyCharm, I don't see any advantage to explicitly type hinting this variable.

radarhere (approved):

PR #9005, opened 2025-06-09
Remove padding between interleaved PCX palette data

### LINKED ISSUE #8952: Loaded PCX image has smeared result
### What did you do?

I loaded a PCX file and used the show() command. Note, I got the same issue if trying to write file as png but I figured .show() was easier to demonstrate?

A non smeared version of the image. (example converted in GIMP) 

Image

### What actually happened?

Loading a PCX file gave me a smeared result of the image. Possibly an issue with loading the different planes?

Image

* OS: Windows 10
* Python: 3.13.3
* Pillow: 11.2.1

```text

> uv run python -m PIL.report
Pillow 11.2.1
Python 3.13.3 (main, Apr  9 2025, 04:04:49) [MSC v.1943 64 bit (AMD64)]
Python executable is .....\.venv\Scripts\python.exe
Environment Python files loaded from \.venv
System Python files loaded from ...\AppData\Roaming\uv\python\cpython-3.13.3-windows-x86_64-none
Python Pillow modules loaded from ....\.venv\Lib\site-packages\PIL
Binary Pillow modules loaded from ......\.venv\Lib\site-packages\PIL
--- ZLIB (PNG/ZIP) support ok, loaded 1.3.1.zlib-ng, compiled for zlib-ng 2.2.4


```

```python
#!/usr/bin/env python3

import logging
logging.root.setLevel(logging.DEBUG)
logging.basicConfig(level=logging.DEBUG)

from PIL import Image

with Image.open("../pillow-bug/KEY.WR.PCX") as img:
    img.show()

with Image.open("../pillow-bug/key-correct.png") as img:
    img.show()


```

Use attached images: 

[pillow-bug.zip](https://github.com/user-attachments/files/20174135/pillow-bug.zip)

(Note: example images are from the shareware version of Word Rescue. I hope that's okay.)

radarhere:
Resolves #8952

#535 added code to remove padding between bands.
https://github.com/python-pillow/Pillow/blob/8e5a15bab746f87822b75cd47db17ae0174ed05e/src/libImaging/PcxDecode.c#L67-L73

However, that didn't consider that there might be padding between lines for interleaved data, like the P;4L image from the issue.

I've created the test image using Pillow and manual hexediting.


src/libImaging/PcxDecode.c
```diff
@@ -60,15 +60,25 @@ ImagingPcxDecode(Imaging im, ImagingCodecState state, UINT8 *buf, Py_ssize_t byt
...
+            int bands;
+            int xsize = 0;
+            int stride = 0;
+            if (state->bits == 2 || state->bits == 4) {
+                xsize = (state->xsize + 7) / 8;
+                bands = state->bits;
+                stride = state->bytes / state->bits;
+            } else {
+                xsize = state->xsize;
+                bands = state->bytes / state->xsize;
```
hugovk, line 72: We had a similar caluclation before, but can `state->xsize` ever be zero and cause a divide by zero here?

src/libImaging/PcxDecode.c
```diff
@@ -60,15 +60,25 @@ ImagingPcxDecode(Imaging im, ImagingCodecState state, UINT8 *buf, Py_ssize_t byt
...
+                if (bands != 0) {
+                    stride = state->bytes / bands;
```
hugovk, line 74: Similarly here?

radarhere: If you were referring to `bands`, you'll see that the line before checks that `bands != 0`. It can be zero here. For example in our test suite, `state->bytes` can be 16 when `state->xsize` is 128. That makes `bands = state->bytes / state->xsize` zero.

src/libImaging/PcxDecode.c
radarhere: We can't open a zero width image, no. See my explanation at the start of #8408, or a demonstration at https://github.com/radarhere/Pillow/actions/runs/15560051255/job/43809855672#step:11:119

hugovk (approved):
Thanks!

PR #9007, opened 2025-06-10
Support writing QOI images

thisismypassport:
Adds support for writing .qoi files.

In more details:
* I followed the reference implementation pretty closely.
* I verified that the test files generated are the same as the reference implementation generates.
* I also followed the coding guidelines observed from the .qoi reading implementation pretty closely, and avoided touching it.

Unrelated note - the .qoi reading implementation incorrectly adds (0, 0, 0, 255) to _previously_seen_pixels at the start.
It should only be set on _previous_pixel.
However, I think this is benign since (0,0,0,255) has a different hash from (0,0,0,0) and a valid encoder won't encode (0,0,0,0) using non-0 hash values.

### DIFF
--- a/Tests/test_file_blp.py
+++ b/Tests/test_file_blp.py
@@ -7,9 +7,8 @@ import pytest
 from PIL import BlpImagePlugin, Image
 
 from .helper import (
-    assert_image_equal,
     assert_image_equal_tofile,
-    assert_image_similar,
+    assert_image_similar_tofile,
     hopper,
 )
 
@@ -52,18 +51,16 @@ def test_save(tmp_path: Path) -> None:
         im = hopper("P")
         im.save(f, blp_version=version)
 
-        with Image.open(f) as reloaded:
-            assert_image_equal(im.convert("RGB"), reloaded)
+        assert_image_equal_tofile(im.convert("RGB"), f)
 
         with Image.open("Tests/images/transparent.png") as im:
             f = tmp_path / "temp.blp"
             im.convert("P").save(f, blp_version=version)
 
-            with Image.open(f) as reloaded:
-                assert_image_similar(im, reloaded, 8)
+            assert_image_similar_tofile(im, f, 8)
 
     im = hopper()
-    with pytest.raises(ValueError):
+    with pytest.raises(ValueError, match="Unsupported BLP image mode"):
         im.save(f)
 
 
--- a/Tests/test_file_ppm.py
+++ b/Tests/test_file_ppm.py
@@ -294,9 +294,10 @@ def test_header_token_too_long(tmp_path: Path, data: bytes) -> None:
     with open(path, "wb") as f:
         f.write(data)
 
-    with pytest.raises(ValueError, match="Token too long in file header: "):
+    with pytest.raises(ValueError) as e:
         with Image.open(path):
             pass
+    assert "Token too long in file header: " in repr(e)
 
 
 def test_truncated_file(tmp_path: Path) -> None:
--- a/docs/releasenotes/11.3.0.rst
+++ b/docs/releasenotes/11.3.0.rst
@@ -54,6 +54,19 @@ Support has been added for saving QOI images. ``colorspace`` can be used to spec
 colorspace as sRGB with linear alpha, e.g. ``im.save("out.qoi", colorspace="sRGB")``.
 By default, all channels will be linear.
 
+Support using more screenshot utilities with ImageGrab on Linux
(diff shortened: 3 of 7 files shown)


radarhere:
> Unrelated note - the .qoi reading implementation incorrectly adds (0, 0, 0, 255) to _previously_seen_pixels at the start.
It should only be set on _previous_pixel.

If I look at https://qoiformat.org/qoi-specification.pdf, I see
> The decoder and encoder start with {r: 0, g: 0, b: 0, a: 255} as the previous pixel value.

src/PIL/QoiImagePlugin.py
```diff
@@ -110,6 +113,111 @@ def decode(self, buffer: bytes | Image.SupportsArrayInterface) -> tuple[int, int
         return -1, 0
 
 
+def _save(im: Image.image, fp: IO[bytes], filename: str | bytes) -> None:
```
radarhere:
```suggestion
def _save(im: Image.Image, fp: IO[bytes], filename: str | bytes) -> None:
```

thisismypassport:
> > Unrelated note - the .qoi reading implementation incorrectly adds (0, 0, 0, 255) to _previously_seen_pixels at the start.
> > It should only be set on _previous_pixel.
> 
> If I look at https://qoiformat.org/qoi-specification.pdf, I see
> 
> > The decoder and encoder start with {r: 0, g: 0, b: 0, a: 255} as the previous pixel value.

Yes, just as the previous pixel value - it says nothing about adding it to the hash.
And indeed, the reference implementation does not add it to the hash - neither in the encoder not decoder.

(I'll look at the other stuff a bit later today, thanks)

radarhere:
> Yes, just as the previous pixel value - it says nothing about adding it to the hash.

Ah. The part of the [spec](https://qoiformat.org/qoi-specification.pdf) that I find most convincing is
> A running array[64] (zero-initialized) of previously seen pixel values is maintained by the encoder and decoder.

I've created #9008

Tests/test_file_qoi.py
```diff
@@ -34,3 +36,22 @@ def test_op_index() -> None:
...
+    assert_image_equal_tofile(im, f)
+
+    for path in ("Tests/images/default_font.png", "Tests/images/pil123rgba.png"):
+        with Image.open(path) as im:
+            im.save(f)
+
+            assert_image_equal_tofile(im, f)
+
+    im = hopper("P")
+    with pytest.raises(ValueError):
```
radarhere:
```suggestion
    with pytest.raises(ValueError, match="Unsupported QOI image mode"):
```

src/PIL/QoiImagePlugin.py
```diff
@@ -110,6 +113,122 @@ def decode(self, buffer: bytes | Image.SupportsArrayInterface) -> tuple[int, int
...
+                                    data += bytes(pixel[:3])
+                        else:
+                            data += o8(0b11111111)  # QOI_OP_RGBA
+                            data += bytes(pixel)
+
+                self._previous_pixel = pixel
+
+        if self._run:
+            data += self._write_run()
+        data += bytes((0, 0, 0, 0, 0, 0, 0, 1))  # padding
```
radarhere, line 224:
```suggestion
                                    data += bytearray(pixel[:3])
                        else:
                            data += o8(0b11111111)  # QOI_OP_RGBA
                            data += bytearray(pixel)

                self._previous_pixel = pixel

        if self._run:
            data += self._write_run()
        data += bytearray((0, 0, 0, 0, 0, 0, 0, 1))  # padding
```

Tests/test_file_qoi.py
thisismypassport: Makes sense.
Keep in mind that I based my test_save on test_save in test_file_blp.py, so you might want to make these improvements (using assert_image_equal_tofile & this suggestion) there as well.

src/PIL/QoiImagePlugin.py
thisismypassport: I'm not sure about this one - given that bytearray is just a mutable version of bytes - and given that you don't need mutability here (since you immediately append it to data) - I would expect bytes to be potentially (now or in the future) more optimal here.

Tests/test_file_qoi.py
radarhere: Sure, I've created https://github.com/python-pillow/Pillow/pull/9020

src/PIL/QoiImagePlugin.py
radarhere: Ok. I was just thinking about casting it to the same type, but yep, that makes sense.

PR #9012, opened 2025-06-12
Add Python 3.14 beta wheels

hugovk:
* With cibuildwheel 3.0.0 (https://github.com/python-pillow/Pillow/pull/9010) and the already present `CIBW_ENABLE: cpython-prerelease`, we now get 3.14 wheels. Add to release notes.
  * See https://iscinumpy.dev/post/cibuildwheel-3-0-0/

* Remove `CIBW_SKIP: pp39-*` to remove a warning:
  * "cibuildwheel: Invalid skip selector: 'pp39-*'. This selector matches a group that wasn't enabled. Enable it using the `enable` option or remove this selector. This selector will have no effect."

* Also update the CI targets table, we're already testing 3.14, and add to `tox.ini`


QuLogic:
Just checking if the macOS x86_64 wheels were specifically dropped for Python 3.14? The release note doesn't specifically say, so I'm not sure if those were being left out for the beta only, or for Python 3.14 entirely.

radarhere:
Good question! This was an oversight. I've created https://github.com/python-pillow/Pillow/pull/9031

PR #9014, opened 2025-06-13
Update libpng to 1.6.49

radarhere:
libpng 1.6.49 has been released - https://github.com/pnggroup/libpng/releases/tag/v1.6.49


radarhere:
The Python 3.13 Windows failure is not related to this PR or our current TIFF issues. The error is https://github.com/python-pillow/Pillow/actions/runs/15629022268/job/44034727346?pr=9014#step:28:777
>   LINK : fatal error LNK1104: cannot open file 'python313t.lib'

It's because it's Windows using Python 3.13.4, which incorrectly tries to link to the threaded version. See https://github.com/python/cpython/issues/135151

Issue #5723, opened 2021-09-20
Inconsistent Image.fromarray() behaviour with mode '1'


dtoniolo:
### What did you do?
The following code can reproduce the issue:
```python
import numpy as np
import PIL
from PIL import Image

print('Testing for Pillow version', PIL.__version__)


img_size = (2, 3)  # (w, h)
rng = np.random.default_rng()

test = rng.integers(0, 2, dtype=np.bool_, size=img_size[::-1])
no_mode = Image.fromarray(test)
no_mode_arr = np.array(no_mode)
outcome = np.all(no_mode_arr == test)
print('No mode specified. Mode:', no_mode.mode, 'dtype:',
      no_mode_arr.dtype, 'outcome:', outcome)
with_mode = Image.fromarray(test, mode='1')
with_mode_arr = np.array(with_mode)
outcome = np.all(with_mode_arr == test)
print('Mode specified. Mode:', with_mode.mode, 'dtype:',
      with_mode_arr.dtype, 'outcome:', outcome)

print('Input array:\n', test)
print('Output array with no mode specified:\n', no_mode_arr)
print('Output array with mode specified:\n', with_mode_arr)
```
Output:
```python
Testing for Pillow version 8.3.2
No mode specified. Mode: 1 dtype: bool outcome: True
Mode specified. Mode: 1 dtype: bool outcome: False
Input array:
 [[ True False]
 [ True False]
 [False False]]
Output array with no mode specified:
 [[ True False]
 [ True False]
 [False False]]
Output array with mode specified:
 [[False False]
 [False False]
 [False False]]
```

### What did you expect to happen?
When initialising an image from a boolean NumPy array, it is expected that `True` values are assigned to white pixels and `False` values to black ones. The resulting image should have `mode` `'1'` and if, converted back to an array, its contents should be equal to the initial input array.

### What actually happened?
If `Image.fromarray()` is called with `mode=None`, the code works as expected. If instead it is called with `mode='1'`, the array results in all `False` values. This is an issue for two reasons.

1. The conversion of the data is incorrect.
1. The behaviour of `Image.fromarray()` is inconsistent: given that when called with `mode=None` it detects the data as having `mode='1'`, passing `mode='1'` as an input argument should *not* change the result.

### What are your OS, Python and Pillow versions?
I have run the script with two different conda virtual environments. The OS is macOS Big Sur v11.5.2 (20G95). 
- Env 1:
  - * Python: 3.8.10
[...]

Issue #5465, opened 2021-05-04
Values change with implicit conversion from mode F to L


meyerjo:
Might be somewhat related to for instance #3011 

### What did you do?

```python
import numpy as np
from PIL import Image
tgt = np.ones(shape=(2, 2)) # float32
tgt_im = Image.fromarray(tgt)
tgt_im_np = np.asarray(tgt_im)
final_img = Image.fromarray(tgt_im_np, mode='L')

print(np.asarray(final_img))
```

### What did you expect to happen?
I assumed that the resulting matrix values still have the value of 1 but are implictly converted to 8-bit range.

### What actually happened?
```pycon
>>> print(np.asarray(final_img))
[[  0   0]
 [128  63]]
```
None of the resulting values keep the original value of 1  and (for me even more surprising) they change to different values at different places in the matrix.

### What are your OS, Python and Pillow versions?

* OS: Ubuntu 20.10
* Python: Python 3.8.8
* Pillow: 
```
pip list
Package    Version
---------- -------------------
certifi    2020.12.5
numpy      1.20.2
Pillow     8.2.0
pip        21.0.1
setuptools 52.0.0.post20210125
wheel      0.36.2
```

radarhere:
Mode for fromarray is the ["Mode to use"](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.fromarray). I think this is the mode to use when reading the numpy data - if so, then it makes sense that the values change when it is told to interpret the input data dirrectly.

You might be interested in `.convert('L')` instead of `, mode='L'`,
```python
import numpy as np
from PIL import Image
tgt = np.ones(shape=(2, 2)) # float32
tgt_im = Image.fromarray(tgt)
tgt_im_np = np.asarray(tgt_im)
final_img = Image.fromarray(tgt_im_np).convert('L')

print(np.asarray(final_img))
```
gives
```
[[1 1]
 [1 1]]
```

meyerjo:
Okay thanks for the workaround.

However, is it an expected behavior that random numbers appear in the matrix after doing the implicit conversion?
I assumed that the ``fromarray`` method either
- calls the conversion script, if the input is np.float32, or,
- raises an exception if the array.dtype does not match the specified mode.

wiredfool:
I think that the mode to use parameter is a footgun. It’s not an implicit conversion, it’s a cast. In this case, you’re interpreting an array of 32 bit floats as 8 bit ints. 

[...]

Issue #5227, opened 2021-01-26
Numpy array computed using vector multiplication not correctly converted


thomas-maschler:
### What did you do?

I am trying to create a PNG image from a Numpy Array. Before converting the array, I transformed values using multiplication.

### What did you expect to happen?
I would expect that the array values used as input correspond 1 to 1 with the output, no matter how they were computed.

### What actually happened?
When I multiply my initial array using a vector, the final image has wrong pixel values (lots of zeros). However, if I replace the vector with a scalar the pixel values are correct.

### What are your OS, Python and Pillow versions?

* OS: MacOS
* Python: 3.8
* Pillow: 8.0.1

```python
import numpy as np
from PIL import Image

vector = np.ones((5, 5, 3)).astype("uint8") * np.array([3])
scalar = np.ones((5, 5, 3)).astype("uint8") * 3

assert np.array_equal(vector, scalar)

vector_img = Image.fromarray(vector, mode="RGB")
scalar_img = Image.fromarray(scalar, mode="RGB")

assert np.array_equal(np.array(vector_img), np.array(scalar_img))
```

radarhere:
Hi. Removing `mode="RGB"` from your code,
```python
vector_img = Image.fromarray(vector)
```
gives me
```
TypeError: Cannot handle this data type: (1, 1, 3), <i8
```
So my diagnosis is that this is just a form of data that we haven't added support for yet. I believe it corresponds to RGB mode with RGB;64S rawmode.

Let us know if you think that it should still be working with `mode="RGB"` at the end.

radarhere:
I think the fact that an error isn't displayed can be considered part of #2856, and the lack of support for this format part of #1888

Issue #4887, opened 2020-08-24
Problems with other data-types than uint8


emilettele:
### What did you do?
I was trying to create an Image from a numpy array using Image.fromarray(array, mode) and validate the created image by using the show() function. It did not create the right image.

### What did you expect to happen?
Show a white image

### What actually happened?
Showed an image with seperated colors.

### What are your OS, Python and Pillow versions?

* OS: Windows
* Python: 3.6.8
* Pillow: 7.2.0

### What is the problem?
Pillow shows the right image if the dtype of the values in the array is uint8, but seems to have problems accepting other data types.

```python
import numpy as np
from PIL import Image

array = np.full((32,32,3),255)
array[0].dtype
image = Image.fromarray(array, "RGB")
image.show()
```
![wrong](https://user-images.githubusercontent.com/9704933/91105415-4ebdab00-e670-11ea-8684-35685bd12235.png)

It does however work if I first convert the array to uint8:
```python
import numpy as np
from PIL import Image

array = np.full((32,32,3),255)
array = np.uint8(array)
array[0].dtype
image = Image.fromarray(array, "RGB")
image.show()
```
![right](https://user-images.githubusercontent.com/9704933/91105457-65fc9880-e670-11ea-9a57-6ad4af2d5ee8.png)

nulano:
Related to #1888.

emilettele:
After going through some of the related issues I now see that @wiredfool commented on 12 Jan 2018 in #2955:
_"That's a 16 bit per channel RGB image, and Pillow doesn't support multi channel images with more than 8 bpc. It's a longstanding issue."_

In my opinion, it would be nice if there would at least be an error text that can explain this limitation of 8bpc. Also, this limitation could be made a lot more clear on the information pages. Since I did not know about this, it took me quite some time to fix the problem.

radarhere:
I'll point out that an error is displayed - if you don't pass `mode` to `fromarray`.

```python
import numpy as np
from PIL import Image

array = np.full((32,32,3),255)
Image.fromarray(array)
```
```
Traceback (most recent call last):
  File "PIL/Image.py", line 2777, in fromarray
KeyError: ((1, 1, 3), '<i8')

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "demo.py", line 5, in <module>
    Image.fromarray(array)
  File "PIL/Image.py", line 2779, in fromarray
TypeError: Cannot handle this data type: (1, 1, 3), <i8
```

[...]

PR #9018, opened 2025-06-14
Deprecate fromarray mode argument

### LINKED ISSUE #9006: Issue with PIL.Image.fromarray(..., mode="1") returning a transposed image
### What did you do?

I have an `np.ndarray` with `shape=(h, w)` and `dtype=bool` that I want to convert to a `PIL.Image.Image` (and later save on disk disk, although that is not relevant).
I have noticed that `PIL.Image.fromarray(array, mode="1")` does not work correctly, while `PIL.Image.fromarray(array, mode=None)` does behave correctly.

Referring to the sample code below, I expect that all the 3 arrays would be equal for all modes.

### What actually happened?
It does not work for `mode="1"`.

Linux:
```text
Pillow 11.2.1
Python 3.12.9 | packaged by Anaconda, Inc. | (main, Feb  6 2025, 18:56:27) [GCC 11.2.0]
Python executable is /home/***/miniconda3/envs/pil-env/bin/python3
System Python files loaded from /home/***/miniconda3/envs/pil-env
Python Pillow modules loaded from /home/***/miniconda3/envs/pil-env/lib/python3.12/site-packages/PIL
Binary Pillow modules loaded from /home/***/miniconda3/envs/pil-env/lib/python3.12/site-packages/PIL
--- ZLIB (PNG/ZIP) support ok, loaded 1.2.13, compiled for zlib-ng 2.2.4
--- RAQM (Bidirectional Text) support ok, loaded 0.10.1, fribidi 1.0.8, harfbuzz 11.0.1
```

macOS:
```text
Pillow 11.1.0
Python 3.12.9 | packaged by conda-forge | (main, Mar  4 2025, 22:44:42) [Clang 18.1.8 ]
Python executable is /Users/***/miniconda3/envs/pil-env/bin/python3
System Python files loaded from /Users/***/miniconda3/envs/pil-env
Python Pillow modules loaded from /Users/***/miniconda3/envs/pil-env/lib/python3.12/site-packages/PIL
Binary Pillow modules loaded from /Users/***/miniconda3/envs/pil-env/lib/python3.12/site-packages/PIL
--- ZLIB (PNG/ZIP) support ok, loaded 1.3.1.zlib-ng, compiled for zlib-ng 2.2.2

```

### Sample code

```python
from PIL import Image
from urllib.request import urlopen
import numpy as np


def main():
    url = "https://python-pillow.github.io/assets/images/pillow-logo.png"

    base_image = Image.open(urlopen(url))
    print(base_image.size, base_image.mode)

    modes = [
        "RGBA",
        "RGB",
        "L",
        "I",
        "F",
        "1",
    ]
    for mode in modes:
        image_1 = base_image.convert(mode=mode)
        array_1 = np.array(image_1)
        image_2 = Image.fromarray(array_1, mode=mode)
        array_2 = np.array(image_2)
        image_3 = Image.fromarray(array_1)
        array_3 = np.array(image_3)

        size_equality = image_1.size == image_2.size == image_3.size
        mode_equality = image_1.mode == image_2.mode == image_3.mode
        shape_equality = array_1.shape == array_2.shape == array_3.shape
        dtype_equality = array_1.dtype == array_2.dtype == array_3.dtype

        print(f"image_1.size = {image_1.size}, image_2.size = {image_2.size}, image_3.size = {image_3.size}")
        print(f"image_1.mode = {image_1.mode}, image_2.mode = {image_2.mode}, image_3.mode = {image_3.mode}")
        print(f"array_1.shape = {array_1.shape}, array_2.shape = {array_2.shape}, array_3.shape = {array_3.shape}")
        print(f"array_1.dtype = {array_1.dtype}, array_2.dtype = {array_2.dtype}, array_3.dtype = {array_3.dtype}")
        print(f"size_equality = {size_equality}")
        print(f"mode_equality = {mode_equality}")
        print(f"shape_equality = {shape_equality}")
        print(f"dtype_equality = {dtype_equality}")

        array_1_and_array_2_equality = (array_1 == array_2).all()
        array_1_and_array_3_equality = (array_1 == array_3).all()
        array_2_and_array_3_equality = (array_2 == array_3).all()
        print(f"array_1_and_array_2_equality = {array_1_and_array_2_equality}")
        print(f"array_1_and_array_3_equality = {array_1_and_array_3_equality}")
        print(f"array_2_and_array_3_equality = {array_2_and_array_3_equality}")

        # uncomment to display images
        # if not all([array_1_and_array_2_equality, array_1_and_array_3_equality, array_2_and_array_3_equality]):
        #     image_1.show()
        #     image_2.show()
        #     image_3.show()

        print()


if __name__ == "__main__":
    main()
```

`image_1`: Image

`image_2`: Image

`image_3`: Image

radarhere:
Resolves #9006

#2856 (with 12 likes), #3781, #4887, #5227, #5465 and #5723 are all issues were the `Image.fromarray()` `mode` parameter has caused confusion.

https://github.com/python-pillow/Pillow/issues/5465#issuecomment-831871276
> I think that the mode to use parameter is a footgun. It’s not an implicit conversion, it’s a cast.

Since those, #5849 improved the documentation, but it's not done causing problems, as #9006 objects to the fact that one type of array can't be correctly read using an explicit mode argument, but `mode=None` can handle it without a problem.

This PR suggests just deprecating the argument altogether.

### DIFF
--- a/Tests/test_image_array.py
+++ b/Tests/test_image_array.py
@@ -101,7 +101,8 @@ def test_fromarray_strides_without_tobytes() -> None:
 
     with pytest.raises(ValueError):
         wrapped = Wrapper({"shape": (1, 1), "strides": (1, 1)})
-        Image.fromarray(wrapped, "L")
+        with pytest.warns(DeprecationWarning):
+            Image.fromarray(wrapped, "L")
 
 
 def test_fromarray_palette() -> None:
@@ -110,7 +111,8 @@ def test_fromarray_palette() -> None:
     a = numpy.array(i)
 
     # Act
-    out = Image.fromarray(a, "P")
+    with pytest.warns(DeprecationWarning):
+        out = Image.fromarray(a, "P")
 
     # Assert that the Python and C palettes match
     assert out.palette is not None
--- a/docs/deprecations.rst
+++ b/docs/deprecations.rst
@@ -193,6 +193,14 @@ Image.Image.get_child_images()
 method uses an image's file pointer, and so child images could only be retrieved from
 an :py:class:`PIL.ImageFile.ImageFile` instance.
 
+Image.fromarray mode parameter
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. deprecated:: 11.3.0
+
+The ``mode`` parameter in :py:meth:`~PIL.Image.fromarray()` has been deprecated. The
+mode can be automatically determined from the object's shape and type instead.
+
 Removed features
 ----------------
 
--- a/docs/releasenotes/11.3.0.rst
+++ b/docs/releasenotes/11.3.0.rst
@@ -23,10 +23,11 @@ TODO
 Deprecations
 ============
 
-TODO
-^^^^
+Image.fromarray mode parameter
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-TODO
+The ``mode`` parameter in :py:meth:`~PIL.Image.fromarray()` has been deprecated. The
+mode can be automatically determined from the object's shape and type instead.
 
 API changes
 ===========
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -3272,7 +3272,7 @@ def fromarray(obj: SupportsArrayInterface, mode: str | None = None) -> Image:
 
     :param obj: Object with array interface
     :param mode: Optional mode to use when reading ``obj``. Will be determined from
-      type if ``None``.
+      type if ``None``. Deprecated.
 
       This will not be used to convert the data after reading, but will be used to
       change how the data is read::
@@ -3307,6 +3307,7 @@ def fromarray(obj: SupportsArrayInterface, mode: str | None = None) -> Image:
             msg = f"Cannot handle this data type: {typekey_shape}, {typestr}"
             raise TypeError(msg) from e
     else:
+        deprecate("'mode' parameter", 13)
         rawmode = mode
     if mode in ["1", "L", "I", "P", "F"]:
         ndmax = 2


neutrinoceros:
How should `Image.fromarray(a, mode='P')` be replaced ? there's one occurrence of this in matplotlib

radarhere:
```python
from PIL import Image
import numpy
a = numpy.ones((256, 256), dtype=numpy.uint8)

# Before
p = Image.fromarray(a, "P")  # This has an empty palette
print(p, p.palette.mode, len(p.palette.colors))

# After
l = Image.fromarray(a)
p = l.convert("P")  # This has a full palette
p.putpalette([])  # If you would like to keep an empty palette, this will do so
print(p, p.palette.mode, len(p.palette.colors))
```

manthey:
I have some uint8 YCbCr data that I access via `imgycbcr = PIL.Image.fromarray(arr, 'YCbCr')`.  What is the correct equivalent with this deprecation?

radarhere:
Could you put together a simple example to demonstrate what you're doing?
```python
from PIL import Image
import numpy
a = numpy.ones((1, 1), dtype=numpy.uint8)
Image.fromarray(a, "YCbCr")
```
gives
```
ValueError: not enough image data
```

manthey:
I have a numpy array of shape (h, w, 3) which contains YCbCr data.  For instance, a single blue pixel might be
```python
a = np.array([[[30, 255, 107]]], dtype=np.uint8)
```
I want to convert this to RGB:
```python
ycbcrimg = PIL.Image.fromarray(a, 'YCbCr')
ycbcrimg.convert('RGB').getpixel((0, 0))
```
yields `(0, 1, 255)` (our blue pixel).

It looks like I can do
```python
ycbcrimg = PIL.Image.frombuffer('YCbCr', (a.shape[1], a.shape[0]), a.tobytes(), 'raw', 'YCbCr', 0, 1)
```
but that seems cumbersome compared to the now deprecated `PIL.Image.fromarray(a, 'YCbCr')`, and I was wondering if there was an alternative preferred idiom for this.

radarhere:
What do you think of this?
```python
from PIL import Image
import numpy as np
a = np.array([[[30, 255, 107]]], dtype=np.uint8)
img = Image.new("YCbCr", (1, 1))
img.putdata([tuple(b[0]) for b in a])
print(img.convert('RGB').getpixel((0, 0)))
```

manthey:
I assume that would be inefficient for large images, since we are constructing a tuple for every pixel.  I used a 1x1 pixel image as the smallest working example; my actual images are larger and can be numerous.

radarhere:
Ok. If the tuple construction is the only problem, I could put together a PR to allow lists to be used, meaning that
```python
from PIL import Image
import numpy as np
a = np.array([[[30, 255, 107]]], dtype=np.uint8)
img = Image.new("YCbCr", (1, 1))
img.putdata(a.tolist()[0])
print(img.convert('RGB').getpixel((0, 0)))
```
would work. How does that sound?

wiredfool:
We really go out of our way to tell people not to do bulk pixel manipulation in python, so deprecating something that does the right thing in 0 time in C and then suggesting that level of pixel pushing in Python seems wrong. 

When this was proposed, there was the one legitimate use case mentioned (P), this is definitely another.  Specifying that the uint8 pixels represent YCbCr instead of RGB is the sort of cast that is safe and useful for importing -- it's essentially the same operation as specifying a colorspace -- rather than specifying a bit layout which is what trips people up. 

I can see two options here:

1) We allow the mode parameter, but only for modes that are compatible with the dtype. This allows for the matrix of RGB(XxA)/Lab/YCbCr to be specified properly. We explicitly error on things like `bool`->`1` and float->int modes. This makes it explicitly cast only, which should remove the footgun aspect of this parameter.

2) We provide an interface to change the `im->mode` between bit compatible modes, so that one can take any RGB image and interpret it as YCbCr.

radarhere:
Ok, I've created #9063 for the first option.

Pinging @guillaume-rochette-oxb just to make sure he is aware of this potential step in the other direction.

manthey:
Thank you.

mara004:
Also ran into this deprecation with pypdfium2.
Problem is, I think this would lose the distinction between RGBA and RGBX, won't it? The former cannot be saved as JPEG, but the latter can.
AFAICS, this case is not included in the revert list in https://github.com/python-pillow/Pillow/pull/9063#issue-3198767838 yet.

radarhere:
Ok, I've added RGBX to the list of RGBA alternatives in the PR.

radarhere:
Pillow 12.0.0 has now been released with #9063

PR #9033, opened 2025-06-24
Use f-strings

radarhere:


hugovk:
This changes from lazy evaluation (only when `DEBUG` is `True`, and by default it's `False`) to always evaluating them.

PR #9040, opened 2025-06-27
Increase pytest verbosity

radarhere:
In discussing the difference between `pytest -v` and `pytest -vv` with regards to checking iOS wheels, @freakboy3742 [said](https://github.com/python-pillow/Pillow/pull/9030#discussion_r2171216972)
> In case of success, there's no real difference; but in case of failure, `-vv` will sometimes include failure context (such as complete diffs) that is truncated by `-v`. If we *can* put the full context in the logs, why not do so?

I'm interested in consistency, so if that argument holds for testing iOS wheels, then why not everywhere else as well?

You can read about the effect at https://docs.pytest.org/en/stable/how-to/output.html#verbosity

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -12,7 +12,7 @@ extras =
     tests
 commands =
     {envpython} selftest.py
-    {envpython} -m pytest -W always {posargs}
+    {envpython} -m pytest -vv -W always {posargs}
 
 [testenv:lint]
 skip_install = true
--- a/.ci/test.cmd
+++ b/.ci/test.cmd
@@ -1,3 +1,3 @@
 python.exe -c "from PIL import Image"
 IF ERRORLEVEL 1 EXIT /B
-python.exe -bb -m pytest -v -x -W always --cov PIL --cov Tests --cov-report term --cov-report xml Tests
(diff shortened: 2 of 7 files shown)


tox.ini
```diff
@@ -12,7 +12,7 @@ extras =
     tests
 commands =
     {envpython} selftest.py
-    {envpython} -m pytest -W always {posargs}
+    {envpython} -m pytest -vv -W always {posargs}
```
hugovk: We're not using tox in CI, so we can leave it to the user to add the options if they want to re-run with more verbosity (which would be something like `tox -e py313 -- -vv`):

```suggestion
    {envpython} -m pytest -W always {posargs}
```

radarhere:
I've pushed a commit to exit early on test failure in wheel tests - which we're already doing in some wheels that use wheels-test.ps1.
https://github.com/python-pillow/Pillow/blob/e783aff6885c8b5d92f819bd9a7bc1566f1b5318/.github/workflows/wheels-test.ps1#L26-L28

hugovk (approved):

PR #9041, opened 2025-06-27
Fix buffer overflow when saving compressed DDS images

radarhere:


marbens-arch:
PoC made by the discoverer (not me):

image

```python
import argparse
from PIL import Image

parser = argparse.ArgumentParser()
parser.add_argument('image')

args = parser.parse_args()

im = Image.open(args.image)
im = im.convert('RGBA')

img440 = im.resize((440,440))

img440.save('image.dds', 'DDS', pixel_format='DXT1')
```

`python3 pillow_bcn_bug.py image.png` (where `image.png` is the path to the bird picture) outputs:
```
double free or corruption (!prev)
Aborted (core dumped)
```

PR #9056, opened 2025-07-01
Thanks, folks!

aclark4life:
As a general rule I think we should acknowledge when significant contribtions come from outside the core team. We know the core team does a lot of work (thank you!) but it's not always obvious when significant contributions come from outside the core team.

In the old change log, we had ACKs via `[radarhere]` syntax which I miss. I don't expect we'll start using the old change log again but maybe we can make a note in the release notes to include such ACKs as needed and appropriate.

Fixes missing acknowledgments.

- Acknowledge non-core team contributors for significant contributions.

### DIFF
--- a/docs/releasenotes/11.3.0.rst
+++ b/docs/releasenotes/11.3.0.rst
@@ -69,12 +69,14 @@ AVIF support in wheels
 
 Support for reading and writing AVIF images is now included in Pillow's wheels, except
 for Windows ARM64 and iOS. libaom is available as an encoder and dav1d as a decoder.
+(Thank you Frankie Dintino and Andrew Murray!)
 
 iOS
 ^^^
 
 Pillow now provides wheels that can be used on iOS ARM64 devices, and on the iOS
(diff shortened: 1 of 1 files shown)


aclark4life:
For example, when folks here are not in the core team and the contribution was significant (more than 5 minutes of work) we should add those folks to the release notes: https://github.com/python-pillow/Pillow/releases/tag/11.3.0

hugovk:
By the way, similar to the old changelog, the releases at https://github.com/python-pillow/Pillow/releases do have acknowledge the primary PR authors.

aclark4life:
> By the way, similar to the old changelog, the releases at https://github.com/python-pillow/Pillow/releases do have acknowledge the primary PR authors.

Right, I think that's what I was trying to say in my comment above but in rereading now I see I didn't make that point. 😄  The ACKs in the GitHub release notes can be selectively included in the docs release notes IF they are significant enough to be included in the docs release notes (many are not in the docs release notes) AND they include contributors outside the core team.

PR #9063, opened 2025-07-03
Only deprecate fromarray mode for changing data types

radarhere:
Further discussion in #9018 has suggested that some of the functionality in the `fromarray()` mode argument is still useful.

https://github.com/python-pillow/Pillow/pull/9018#issuecomment-3031333730
> When this was proposed, there was the one legitimate use case mentioned (P), this is definitely another. Specifying that the uint8 pixels represent YCbCr instead of RGB is the sort of cast that is safe and useful for importing -- it's essentially the same operation as specifying a colorspace -- rather than specifying a bit layout which is what trips people up.

This PR restores it for changing
- L -> P
- LA -> La, PA
- RGB -> YCbCr, LAB, HSV
- RGBA -> RGBa, RGBX, CMYK

### DIFF
--- a/docs/deprecations.rst
+++ b/docs/deprecations.rst
@@ -35,11 +35,12 @@ Image.fromarray mode parameter
 
 .. deprecated:: 11.3.0
 
-Using the ``mode`` parameter in :py:meth:`~PIL.Image.fromarray()` to change data types
-has been deprecated. Since pixel values do not contain information about palettes or
-color spaces, the parameter can still be used to place grayscale L mode data within a
-P mode image, or read RGB data as YCbCr for example. If omitted, the mode will be
-automatically determined from the object's shape and type.
+Using the ``mode`` parameter in :py:meth:`~PIL.Image.fromarray()` was deprecated in
+Pillow 11.3.0. In Pillow 12.0.0, this was partially reverted, and it is now only
+deprecated when changing data types. Since pixel values do not contain information
+about palettes or color spaces, the parameter can still be used to place grayscale L
+mode data within a P mode image, or read RGB data as YCbCr for example. If omitted, the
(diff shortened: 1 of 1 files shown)


djhoese:
Any idea when this might be merged? And any idea what version it might be released in?

The previous deprecation has been causing a lot of warnings in my projects which I think this PR should fix. Thanks for allowing these cases.

radarhere:
Pillow 12.0.0 is scheduled for October 15. I expect that it will be merged by then.

docs/deprecations.rst
```diff
@@ -35,8 +35,11 @@ Image.fromarray mode parameter
 
 .. deprecated:: 11.3.0
 
-The ``mode`` parameter in :py:meth:`~PIL.Image.fromarray()` has been deprecated. The
-mode can be automatically determined from the object's shape and type instead.
+Using the ``mode`` parameter in :py:meth:`~PIL.Image.fromarray()` to change data types
+has been deprecated. Since pixel values do not contain information about palettes or
+color spaces, the parameter can still be used to place grayscale L mode data within a
+P mode image, or read RGB data as YCbCr for example. If omitted, the mode will be
+automatically determined from the object's shape and type.
```
hugovk: Thanks, let's give a mention what happened in 11.3.0 and what's updated for 12.0.0.

It's good that's been added to the 11.3/12 release notes, but let's also include here for those who just read this.

weathon:
I am converting from HSV space. And I get the warning. Is this solved in latest version?

djhoese:
Nope, not yet: https://github.com/python-pillow/Pillow/pull/9063#issuecomment-3121046271

radarhere:
Pillow 12.0.0 has now been released with these changes.

Issue #6144, opened 2022-03-20
ImageGrab.grab(bbox=) fails on Mac with Retina screen, or on secondary monitor


resnbl:
### What did you do?
Displayed a window using `tkinter` and attempted to file a screenshot of it.
### What did you expect to happen?
Create an image file showing the window contents.
### What actually happened?
The generated image shows an area above and to the left of the intended target, and only covers an area 1/4th of the expected size.
### What are your OS, Python and Pillow versions?

* OS: Mac OS 12.2.1
* Python: 3.10.1
* Pillow: 9.0.1

```python
import sys
import tkinter as tk
from tkinter import ttk
from PIL import ImageGrab, __version__ as pil_version


def callback():
    img = ImageGrab.grab()
    img.save('screen.png')

    geom = root.winfo_geometry()
    print('geom=', geom)
    left, top = root.winfo_rootx(), root.winfo_rooty()
    width, height = root.winfo_width(), root.winfo_height()
    print(f'{left=} {top=} {width=} {height=}')
    bbox = (left, top, left+width, top+height)
    img = ImageGrab.grab(bbox=bbox)
    img.save('window.png')


print('Python:', sys.version)
print('PIL:', pil_version)
root = tk.Tk()
root.title('ImageGrab Test')
root.geometry('200x300+100+200')
message = tk.Label(root, text='Hello, World!')
message.pack()
save_btn = ttk.Button(root, text='Save Me', command=lambda: callback())
save_btn.pack(ipadx=5, ipady=5, expand=True)
root.mainloop()
```
Below is a (cropped) version of the 'screen.png' file generated by my script:
<img width="1176" alt="screen" src="https://user-images.githubusercontent.com/33208090/159143737-46714c3b-8b73-492f-afea-f2e75e94b685.png">

Here is the captured image:
![window](https://user-images.githubusercontent.com/33208090/159143744-3e487cca-624f-4614-a2a2-7c13771b5547.png)

`ImageGrab.grab()` is failing to account for the 144 DPI (not 72 DPI "virtual pixels") image generated by `screencapture` for my Retina display, so the bbox passed to the crop function needs to be scaled by 2 (all 4 parameters).

Secondary problem: this all fails completely when the app window is on a secondary monitor.

Proposed solution: Apple knows best!
Instead of capturing the entire screen and then cropping it down, make use of the provided `-R{x,y,w,h}` option for the`screencapture` command. This will:
[...]

PR #9070, opened 2025-07-07
feat(ImageGrab): enhance grab function to support window-based screenshot capturing on macOS

yankeguo:

 * add `-l` argument to `screencapture` for macOS, allowing capture a window

### DIFF
--- a/src/PIL/ImageGrab.py
+++ b/src/PIL/ImageGrab.py
@@ -43,7 +43,10 @@ def grab(
             fh, filepath = tempfile.mkstemp(".png")
             os.close(fh)
             args = ["screencapture"]
-            if bbox:
+            if window:
+                args += ["-l", str(window)]
+            # -R is not working with -l
+            if bbox and not window:
                 left, top, right, bottom = bbox
                 args += ["-R", f"{left},{top},{right-left},{bottom-top}"]
             subprocess.call(args + ["-x", filepath])
@@ -51,9 +54,16 @@ def grab(
             im.load()
             os.unlink(filepath)
             if bbox:
-                im_resized = im.resize((right - left, bottom - top))
-                im.close()
-                return im_resized
+                # manual crop for windowed mode
+                if window:
+                    left, top, right, bottom = bbox
+                    im_cropped = im.crop((left, top, right, bottom))
+                    im.close()
+                    return im_cropped
+                else:
+                    im_resized = im.resize((right - left, bottom - top))
+                    im.close()
+                    return im_resized
             return im
         elif sys.platform == "win32":
             if window is not None:


radarhere:
Hi. Just for the record, how would a Python user get the window id?

yankeguo:
here is a example script:

```python
# pip install pyobjc
# pip install pyobjc-core
# pip install pyobjc-framework-quartz

import Quartz

windows = Quartz.CGWindowListCopyWindowInfo(
    Quartz.kCGWindowListExcludeDesktopElements, Quartz.kCGNullWindowID
)

for window in windows:
    window_id = window["kCGWindowNumber"]

    window_name = window["kCGWindowName"]

    window_x = window["kCGWindowBounds"]["X"]
    window_y = window["kCGWindowBounds"]["Y"]
    window_width = window["kCGWindowBounds"]["Width"]
    window_height = window["kCGWindowBounds"]["Height"]

    window_owner_name = window["kCGWindowOwnerName"]
    window_owner_pid = window["kCGWindowOwnerPID"]
```

user can determine which window id to use

there is also a tool:

https://github.com/smokris/GetWindowID

src/PIL/ImageGrab.py
```diff
@@ -43,17 +43,27 @@ def grab(
...
             im.load()
             os.unlink(filepath)
             if bbox:
-                im_resized = im.resize((right - left, bottom - top))
-                im.close()
-                return im_resized
+                # manual crop for windowed mode
+                if window:
+                    left, top, right, bottom = bbox
+                    im_cropped = im.crop((left, top, right, bottom))
```
radarhere:
```suggestion
                    im_cropped = im.crop(bbox)
```

src/PIL/ImageGrab.py
```diff
@@ -43,17 +43,27 @@ def grab(
...
                 args += ["-R", f"{left},{top},{right-left},{bottom-top}"]
             subprocess.call(args + ["-x", filepath])
             im = Image.open(filepath)
             im.load()
             os.unlink(filepath)
             if bbox:
-                im_resized = im.resize((right - left, bottom - top))
-                im.close()
-                return im_resized
+                # manual crop for windowed mode
```
radarhere: There's an interesting problem here.

Originally, we did just crop the image using Pillow. However, that changed in #6152, because #6144 reported that it wasn't accounting for scaling on retina displays.

So if you pick a specific window, how would we know whether it was scaled or not? It might be on a retina display or it might not. If this were Swift, I'd say we want the [`backingScaleFactor`](https://developer.apple.com/documentation/appkit/nswindow/backingscalefactor) for the window. I'm not convinced AppleScript could easily get information about a window just from the id.

yankeguo: I will look into whether there is any information about Retina in the EXIF data of the generated images.

yankeguo: seems no good solution, may be we should just put this on hold.

However, using the -l parameter when capturing a window does have value, as it can create a complete screenshot of an obscured window compared to taking a full screen capture and then cropping it.

yankeguo: I thought of a less elegant method, using -R0,0,100,100 to sample an image and then calculate the ratio, but I cannot ensure that the window obtained with -l and the image sampled with -R are on the same display.

Another option is to behave like the `screencapture` command line, ignoring the bbox parameter when passing the window parameter.

radarhere: I created https://github.com/yankeguo/Pillow/pull/2

One of the `screencapture` flags is
> -o      In window capture mode, do not capture the shadow of the window.

Excluding the shadow changes the size of the window. So if we take two screenshots of a window, we can measure the size of the shadow.

On a retina screen, the shadow is twice as wide as it is on a non-retina screen.

It is possible that a window doesn't have a shadow, but perhaps that can be considered an edge case.

docs/reference/ImageGrab.rst
```diff
@@ -42,9 +42,9 @@ or the clipboard to a PIL image memory.
         .. versionadded:: 7.1.0
 
     :param window:
-        HWND, to capture a single window. Windows only.
+        Capture a single window. On Windows, this is a HWND. On macOS, it uses windowid.
 
-        .. versionadded:: 11.2.1
+        .. versionadded:: 11.2.1 (Windows), 12.0.0 (macOS)
```
hugovk: This renders like:
image

https://pillow--9070.org.readthedocs.build/en/9070/reference/ImageGrab.html#PIL.ImageGrab.grab

Might this be better?

```suggestion
        .. versionadded:: 11.2.1 Windows support
        .. versionadded:: 12.0.0 macOS support
```

docs/reference/ImageGrab.rst
```diff
@@ -42,9 +42,9 @@ or the clipboard to a PIL image memory.
         .. versionadded:: 7.1.0
 
     :param window:
-        HWND, to capture a single window. Windows only.
+        Capture a single window. On Windows, this is a HWND. On macOS, it uses windowid.
```
hugovk: What's windowid? There's no other mention in the docs or code.

docs/reference/ImageGrab.rst
```diff
@@ -44,9 +44,10 @@ or the clipboard to a PIL image memory.
...
```
radarhere:
```suggestion
        Capture a single window. On Windows, this is a HWND. On macOS, this is a
        CGWindowID.
```

radarhere (approved):

docs/reference/ImageGrab.rst
radarhere: I've pushed a commit to describe it as a CGWindowID instead.

docs/reference/ImageGrab.rst
```diff
@@ -44,9 +44,11 @@ or the clipboard to a PIL image memory.
         .. versionadded:: 7.1.0
 
     :param window:
-        HWND, to capture a single window. Windows only.
+        Capture a single window. On Windows, this is a HWND. On macOS, this is a
+        CGWindowID.
 
-        .. versionadded:: 11.2.1
+        .. versionadded:: 11.2.1 Windows support
+        .. versionadded:: 12.0.0 macOS support
```
radarhere:
```suggestion
        .. versionadded:: 12.1.0 macOS support
```

src/PIL/ImageGrab.py
```diff
@@ -43,17 +43,45 @@ def grab(
...
             subprocess.call(args + ["-x", filepath])
             im = Image.open(filepath)
             im.load()
             os.unlink(filepath)
             if bbox:
-                im_resized = im.resize((right - left, bottom - top))
-                im.close()
-                return im_resized
+                if window:
+                    # Determine if the window was in retina mode or not
```
hugovk:
```suggestion
                    # Determine if the window was in Retina mode or not
```

hugovk (approved):

hugovk:
> Copilot AI review requested due to automatic review settings 3 minutes ago
> Copilot started reviewing on behalf of hugovk 3 minutes ago
> Copilot finished reviewing on behalf of hugovk now

Erm, I don't think I requested that...

src/PIL/ImageGrab.py
```diff
@@ -43,17 +43,45 @@ def grab(
...
+                            box=tuple(coord * 2 for coord in bbox),
+                        )
+                    else:
+                        im_cropped = im.crop(bbox)
+                    im.close()
+                    return im_cropped
+                else:
+                    im_resized = im.resize((right - left, bottom - top))
+                    im.close()
+                    return im_resized
```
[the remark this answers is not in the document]
radarhere: This is a resize operation, not a crop operation.

src/PIL/ImageGrab.py
```diff
@@ -43,17 +43,45 @@ def grab(
...
+                    with Image.open(filepath) as im_no_shadow:
+                        retina = im.width - im_no_shadow.width > 100
+                    os.unlink(filepath)
+
+                    # Since screencapture's -R does not work with -l,
+                    # crop the image manually
+                    if retina:
+                        left, top, right, bottom = bbox
+                        im_cropped = im.resize(
+                            (right - left, bottom - top),
```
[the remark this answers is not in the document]
radarhere: The window is captured twice.

> Implicitly through the subprocess calls

This isn't an additional time.

> Consider whether the Retina detection step is necessary, or if there's a way to determine Retina mode without an additional capture.

See discussion at https://github.com/python-pillow/Pillow/pull/9070#discussion_r2189418847

src/PIL/ImageGrab.py
```diff
@@ -43,17 +43,45 @@ def grab(
...
+                            box=tuple(coord * 2 for coord in bbox),
+                        )
```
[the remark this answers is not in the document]
radarhere: Calling two methods would not be more efficient, from a performance perspective, than one.

src/PIL/ImageGrab.py
```diff
@@ -43,17 +43,45 @@ def grab(
...
+                    # Determine if the window was in Retina mode or not
+                    # by capturing it without the shadow,
+                    # and checking how different the width is
+                    fh, filepath = tempfile.mkstemp(".png")
+                    os.close(fh)
+                    subprocess.call(
+                        ["screencapture", "-l", str(window), "-o", "-x", filepath]
+                    )
+                    with Image.open(filepath) as im_no_shadow:
+                        retina = im.width - im_no_shadow.width > 100
```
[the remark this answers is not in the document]
radarhere: The current comment is
> Determine if the window was in Retina mode or not by capturing it without the shadow, and checking how different the width is

I think it's reasonably apparent that we're using an absolute threshold

src/PIL/ImageGrab.py
```diff
@@ -43,17 +43,45 @@ def grab(
...
-                im.close()
-                return im_resized
+                if window:
+                    # Determine if the window was in Retina mode or not
+                    # by capturing it without the shadow,
+                    # and checking how different the width is
+                    fh, filepath = tempfile.mkstemp(".png")
+                    os.close(fh)
+                    subprocess.call(
+                        ["screencapture", "-l", str(window), "-o", "-x", filepath]
```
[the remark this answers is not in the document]
radarhere: I've created https://github.com/python-pillow/Pillow/pull/9321

src/PIL/ImageGrab.py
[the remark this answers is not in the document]
radarhere: I've created https://github.com/python-pillow/Pillow/pull/9321

PR #9082, opened 2025-07-10
Updated libwebp to 1.6.0

radarhere:
libwebp 1.6.0 has been released - https://github.com/webmproject/libwebp/releases/tag/v1.6.0


.github/workflows/wheels-dependencies.sh
```diff
@@ -103,7 +103,7 @@ TIFF_VERSION=4.7.0
 LCMS2_VERSION=2.17
 ZLIB_VERSION=1.3.1
 ZLIB_NG_VERSION=2.2.4
-LIBWEBP_VERSION=1.5.0  # Patched; next release won't need patching. See patch file.
```
radarhere, line 106: @freakboy3742 it seems that 1.6.0 without a patch isn't working on iOS

freakboy3742: IIRC there was a small config change needed to explicitly perform the behavior that the patch was implementing. I'll take a look in the morning if you don't beat me to working it out.

.github/workflows/wheels-dependencies.sh
```diff
@@ -282,7 +282,7 @@ function build {
     fi
     CFLAGS="$CFLAGS $webp_cflags" build_simple libwebp $LIBWEBP_VERSION \
         https://storage.googleapis.com/downloads.webmproject.org/releases/webp tar.gz \
```
freakboy3742: I've found the cause - because iOS is all statically linked, any transitive dependencies need to be explicitly linked:
```suggestion
    webp_ldflags=""
    if [[ -n "$IOS_SDK" ]]; then
        webp_ldflags="$webp_ldflags -llzma -lz"
    fi
    CFLAGS="$CFLAGS $webp_cflags" LDFLAGS="$LDFLAGS $webp_ldflags" build_simple libwebp $LIBWEBP_VERSION \
```
There's probably something that can be pushed upstream to address this, but in the meantime, adding the 2 extra link arguments works without the need for a patch.

.github/workflows/wheels-dependencies.sh
```diff
@@ -280,9 +280,12 @@ function build {
...
     fi
-    CFLAGS="$CFLAGS $webp_cflags" build_simple libwebp $LIBWEBP_VERSION \
-        https://storage.googleapis.com/downloads.webmproject.org/releases/webp tar.gz \
-        --enable-libwebpmux --enable-libwebpdemux
+    webp_ldflags=""
+    if [[ -n "$IOS_SDK" ]]; then
+        webp_ldflags="$webp_ldflags -llzma -lz"
+    fi
+    CFLAGS="$CFLAGS $webp_cflags" LDFLAGS="$LDFLAGS $webp_ldflags" build_simple libwebp $LIBWEBP_VERSION \
+        --enable-libwebpmux --enable-libwebpdemux --disable-libwebpexamples
```
freakboy3742: Looks like I messed up my suggestion, and inadvertently deleted the actual URL.
```suggestion
    CFLAGS="$CFLAGS $webp_cflags" LDFLAGS="$LDFLAGS $webp_ldflags" build_simple libwebp $LIBWEBP_VERSION \
        https://storage.googleapis.com/downloads.webmproject.org/releases/webp tar.gz \
        --enable-libwebpmux --enable-libwebpdemux --disable-libwebpexamples
```

freakboy3742 (approved):
I'm a little surprised the `--disable-libwebpexamples` isn't needed - they must have changed something else to prevent building the executables from being a problem. But - you can't argue with success - it's building libwebp, and it's passing the libwebp tests.

hugovk:
Thanks!

PR #9093, opened 2025-07-16
Update for pyroma 5.0

radarhere:
pyroma 5.0 has been [released](https://pypi.org/project/pyroma/5.0/), and it is causing failures - https://github.com/python-pillow/Pillow/actions/runs/16298847553/job/46046857802#step:11:5143

This PR
- reverts a test_pyroma.py change for iOS from #9030. The test won't actually run on iOS until https://github.com/python/cpython/pull/135967 lands (see discussion in https://github.com/freakboy3742/Pillow/pull/5), so I accept that this will fail at some point in the future, and we can fix it then. If @freakboy3742 wants to test it and offer an alternative solution, feel free.
- update the manifest
- embed_pil.c is a file created by our test suite. Rather than adding it to the manifest, let's actually remove it once the test has finished

### DIFF
--- a/Tests/test_image_access.py
+++ b/Tests/test_image_access.py
@@ -276,11 +276,10 @@ class TestEmbeddable:
         except Exception:
             pytest.skip("Compiler could not be initialized")
 
-        try:
-            with open("embed_pil.c", "w", encoding="utf-8") as fh:
-                home = sys.prefix.replace("\\", "\\\\")
-                fh.write(
-                    f"""
+        with open("embed_pil.c", "w", encoding="utf-8") as fh:
+            home = sys.prefix.replace("\\", "\\\\")
+            fh.write(
+                f"""
 #include "Python.h"
 
 int main(int argc, char* argv[])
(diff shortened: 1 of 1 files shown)


freakboy3742:
Getting the fix for iOS in place will require an update to the iOS support package, then a cibuildwheel release to make that support package the default version. The former is on my todo list (I'm a little distracted by EuroPython this week); the latter then comes down to cibuildwheel's release cadence, but given 3.14-rc1 will be out shortly, I can't imagine a release would be that far behind.

Tests/test_image_access.py
```diff
@@ -301,17 +302,19 @@ def test_embeddable(self) -> None:
...
+            getattr(ctypes, "windll").kernel32.SetErrorMode(0x0002)
 
-        process = subprocess.Popen(["embed_pil.exe"], env=env)
-        process.communicate()
-        assert process.returncode == 0
+            process = subprocess.Popen(["embed_pil.exe"], env=env)
+            process.communicate()
+            assert process.returncode == 0
+        finally:
+            os.remove("embed_pil.c")
```
hugovk: Shall we add `teardown_method` to this class to delete it, to save an indent?

https://docs.pytest.org/en/stable/how-to/xunit_setup.html#method-and-function-level-setup-teardown

radarhere: Ok, I've pushed a commit.

Issue #7013, opened 2023-03-14
getbbox incorrect when string contains new line character


vanviethieuanh:
### What did you do?

So I use the `getbbox` to calculate the width and height of my string with a specific font. The string includes the `\n` character and it split the line into 2.

### What did you expect to happen?

 The string renders well and has a width of about 7xx pixels which is what I want. And I expect the `getbbox` function returns the same value.

### What actually happened?

However, the result I got from `getbbox` is 1555 pixels which might be the width of no break line text.

### What are your OS, Python, and Pillow versions?

* OS: Windows 10 Enterprise
* Python: Python 3.10.10
* Pillow: 9.4.0

```python
text = "Age is an issue of mind over matter.\nIf you don't mind, it doesn't matter."
_, _, text_width, text_height = FONT.getbbox(text)
# text_width = 1555
```

![image](https://user-images.githubusercontent.com/29355962/224971601-4429a5e4-78f3-4734-81c9-afcdd1ec483a.png)
The size of the image is 1080*1080 so there is no way that the text width is 1555.

Thank you!

nulano:
Line breaking is handled only in `ImageDraw` and not in `ImageFont`; you should use `ImageDraw.multiline_textbbox()`. The thread around https://github.com/python-pillow/Pillow/issues/5816#issuecomment-962864866 might interest you.

radarhere:
@vanviethieuanh did that answer your question?

vanviethieuanh:
@radarhere yes, it does! Thanks (。・ω・。)

mofear85:
Just faced this problem. Took a while to narrow it down. A hint in the documentation would be greatly appreciated (I assume). Either way thx for this great piece of software.

radarhere:
I'd be happy to add a note at the very top of https://pillow.readthedocs.io/en/stable/reference/ImageFont.html, but I wouldn't want to add a note to [`getbbox()`](https://pillow.readthedocs.io/en/stable/reference/ImageFont.html#PIL.ImageFont.ImageFont.getbbox) specifically - there's a number of ImageFont methods, and it would be very repetitive to add it to each one.

Do you think it would be noticed at the top of the page?

mofear85:
can't it be added to the docstring of getbbox() itself?

btw, isn't it possible to make the function properly determine the size even if the next includes newlines? The result is font dependent, but this should be easy/possible when the function is called as a method, ie from a specific font instance. Currently, even if done so, it ignores newlines.

[...]

Issue #6201, opened 2022-04-11
Feature Request: automatic text wrapping/text box filling


HexCodeFFF:
[many people have written scripts to do this](https://gist.github.com/turicas/1455973) and it's relatively easy with the library's `getsize` func but I feel like it really should be a built-in feature.

i think there should be a `draw_text_box` or similar function which has these properties on top of the existing `draw_multiline_text`:

- you can specify a max width in pixels and Pillow will automatically insert line breaks to keep the text's width under the max width
  - I think there should be an option to only break at word boundaries unless the word exceeds the max width, like the [CSS word-break property](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break)
- you can specify a max height in pixels with one of two behaviors:
  - when reaching the max height, the text box is clipped and doesn't draw any more.
  - when reaching the max height, the font size is gradually reduced until the text fits inside
    - I think a min font size property would be useful here
  - also an option to find the biggest font size that fits within the text box
     - a max font size property would be useful here

I'm not very experienced with the library internals or C in general so for now I won't make a pull, I just want to throw the idea out there to the devs

HexCodeFFF:
also options to align text left, right, center, or justify would be useful as well. [This script implements that](https://gist.github.com/jasondilworth56/27764ae8ed2327a34ceebb06e75c30ea)

nulano:
Sounds like a reasonable enhancement request. Note that there might be issues with mixed LTR/RTL text which will need extra tests.

A few notes:

> with the library's `getsize` func

Please use the `getlength` function instead, the API of the `getsize` function is fundamentaly broken (especially with non-English text) and shouldn't be used for text layout purposes.

> I think there should be an option to only break at word boundaries unless the word exceeds the max width, like the [CSS word-break property](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break)

Word breaking is quite a difficult task, it might be better to constrain this to spaces to start with.

> when reaching the max height, the font size is gradually reduced until the text fits inside

Not possible within the current API, a font is created at a given size and cannot be easily changed.

> also options to align text left, right, center

Already possible using the `align` parameter, only `justify` is not yet supported. It would also require extra work in the new proposed function, so I see that as a separate request.

> and top/middle/bottom text alignment

Already possible, use the `anchor` parameter with multiline text.

HexCodeFFF:
[...]

Issue #5816, opened 2021-11-06
getsize_multiline doesn't take into account characters that extend below the baseline


jpetazzo:
### What did you do?

I'm using `getsize_multiline` to compute the size of a text label that I render to an image. The image is exactly the size of the text label.

I noticed that my text labels where cropped; specifically, when there are letters like `p` or `g` (which extend below the baseline) on the last line of text, these letters get cropped.

The problem doesn't affect `getsize`, and can easily be exhibited by comparing the sizes reported by `getsize` vs `getsize_multiline` for a single line string like `g`.

Example:

```python
from PIL import ImageFont
font = ImageFont.truetype("DroidSans", 20)
print(font.getsize("g"))
print(font.getsize_multiline("g"))
```

### What did you expect to happen?

The `getsize` and `getsize_multiline` methods should return the same size.

### What actually happened?

With `DroidSans` in size 20:
- `getsize` returns (10,24)
- `getsize_multiline` returns (10,19)

The `g` actually extends below if I create a label of size 10,19 and draw the string `g` on it, the `g` gets cropped.

### What are your OS, Python and Pillow versions?

* OS: Arch Linux
* Python: 3.9
* Pillow: 8.4.0

```python
#!/usr/bin/env python

from PIL import Image, ImageFont, ImageDraw

font = ImageFont.truetype("DroidSans", 20)

text = "gÂp"

size = font.getsize(text)
image = Image.new("RGB", size)
draw = ImageDraw.Draw(image)
draw.text((0, 0), text, font=font)
image.save("getsize.png")

size = font.getsize_multiline(text)
image = Image.new("RGB", size)
draw = ImageDraw.Draw(image)
draw.text((0, 0), text, font=font)
image.save("getsize_multiline.png")
```

getsize: ![getsize](https://user-images.githubusercontent.com/171481/140619714-defdcb34-edfb-46a9-8618-00fc3d3ac358.png)

getsize_multiline: ![getsize_multiline](https://user-images.githubusercontent.com/171481/140619717-1fdc4e28-f989-480d-844d-906b05376932.png)

I imagine that the logic in getsize_multiline doesn't take into accounts characters that extend below the baseline. For now I'm going to just add a bit of padding on the last line 😅 but I guess there has to be a better way.

jpetazzo:
[...]

PR #9098, opened 2025-07-19
Added ImageText

radarhere:
`ImageDraw.multiline_text` currently takes 14 arguments, and `ImageDraw.multiline_textbbox` currently accepts 12 arguments.

#6201 would probably lead to that number increasing further, and there are [some](https://github.com/python-pillow/Pillow/issues/5816#issuecomment-962864866) [comments](https://github.com/python-pillow/Pillow/issues/7013#issuecomment-2948807948) that wonder why calculating multiline text bounding box dimensions requires an ImageDraw instance.

I'm going to suggest adding a new `ImageText` class. I consider this to be a tidying of the API, rather than something motivated by a pressing need, so if this is accepted, maybe we don't need an immediate deprecation.

Before
```python
im = Image.new(mode, (0, 0))
d = ImageDraw.Draw(im)

d.multiline_textlength(text, font, direction, features, language, embedded_color)
d.multiline_textbbox(xy, text, font, anchor, spacing, align, direction, features, language, stroke_width, embedded_color)
```

After
```python
from PIL import ImageText
text = ImageText.ImageText(text, font, mode, spacing, direction, features, language)
text.embed_color()
text.stroke(stroke_width)

text.get_length()
text.get_bbox(xy, anchor)
```

New features like #6201 would just be made exclusive to the new class, something like `text.wrap(width, height, scaling, limit)`.

If nothing else, this will separate the text handling logic from the rest of the ImageDraw code.

I have some further improvements to make, but wanted to see if there was any feedback.

### DIFF
--- a/src/PIL/ImageDraw.py
+++ b/src/PIL/ImageDraw.py
@@ -35,10 +35,10 @@ import math
 import struct
 from collections.abc import Sequence
 from types import ModuleType
-from typing import Any, AnyStr, Callable, Union, cast
+from typing import Any, AnyStr, Callable, cast
 
-from . import Image, ImageColor
-from ._typing import Coords
+from . import Image, ImageColor, ImageText
+from ._typing import Coords, _Ink
 
 # experimental access to the outline API
 Outline: Callable[[], Image.core._Outline] = Image.core.outline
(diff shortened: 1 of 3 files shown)


src/PIL/ImageDraw.py
```diff
@@ -568,29 +561,18 @@ def text(
...
-                spacing,
-                align,
-                direction,
-                features,
-                language,
-                stroke_width,
-                stroke_fill,
-                embedded_color,
+        if isinstance(text, ImageText.ImageText):
+            imagetext = text
```
hugovk: And below:
```suggestion
            image_text = text
```

Issue #8329, opened 2024-08-25
Arrow Support


wiredfool:
Following on to some of the discussion in #1888, specifically here: https://github.com/python-pillow/Pillow/issues/1888#issuecomment-2112706708

## Rationale

Arrow is the emerging memory layout for zero copy sharing of data in the new data ecosystem. It is an uncompressed columnar format, specifically designed for interop between different implementations and languages. It can be viewed as the spiritual successor to the existing numpy array interface that we provide. The arrow format is supported by numpy 2, pandas 2, polars, pyarrow, and arro3, and [https://github.com/apache/arrow/issues/39195#issuecomment-2245718008](others) in the python ecosystem. 

## What Support means

* The ability to export an image to an arrow array and read/process that no data memory copies
* The ability to read an image in arrow array storage with 0 copies.

## Technical Details

(Apache docs are here: https://arrow.apache.org/docs/format/Columnar.html)

An Arrow Schema is a set of metadata, containing type information, and potentially child schemas. An Arrow Array has an (implicitly) associated schema, metadata about the length of the storage, as well as a buffer of a contiguously allocated chunk of memory for the data. The Arrow Array will generally have the same parent/child arrangement as the schema structure. 

* `obj.__arrow_c_schema__()` must return a PyCapsule with an `arrow_schema` name and an arrow schema struct.
* `obj.__arrow_c_array__(schema=None)` must return a tuple of the schema above and a PyCapsule with an `arrow_array` name and an arrow array struct. The schema is advisory, caller may request a format.
 
The lifetime of the Schema and Array structures is dependent on the caller -- so there are release callbacks that must be called when the caller is done with the memory. This complicates the lifetime of our image storage. 

We have two cases at the moment:
1) single channel image
2) multichannel image

A single channel image can be encoded as a single array of height*width items, using the type of the underlying storage. (e.g., uint8/int32/float32).

A multichannel image _can_ be encoded in a similar manner, using 4*height*width items in the array. The caller would be responsible for knowing that it's 4 elements per pixel. It's also possible to use a parent type of a `FixedWidthArray` of 4 elements, and a child array of 4*height*width elements. The fixed width arrays are statically defined, so the underlying array is still the same continuous block of memory. 

Flat:
```
<pyarrow.lib.FixedSizeListArray object at 0x106ad4280>
[
    20,
    21,
    67,
    255
    17,
    18,
    62,
    255
...
```

Nested:
```
[...]

PR #9099, opened 2025-07-19
Band names for arrow exported images

wiredfool:
Fixes https://github.com/python-pillow/Pillow/issues/8329#issuecomment-2963958884

 * Add image band metadata in the arrow schema

@rok

### DIFF
--- a/Tests/test_pyarrow.py
+++ b/Tests/test_pyarrow.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import json
 from typing import Any, NamedTuple
 
 import pytest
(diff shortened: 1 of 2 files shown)


rok:
@paleolimbot am I right to say array metadata was not really designed to pass application metadata? What are your thoughts on what's being done here?
@wiredfool has the issue here that he can't access metadata of children when in certain cases.
I'm suggesting usage of FixedShapeTensorArray, but I imagine working with arrays directly would simplify things here.

Tests/test_arro3.py
```diff
@@ -0,0 +1,278 @@
...
+    assert img.size == (121, 127)
+
+    arr = Array(img)
+    _test_img_equals_pyarray(img, arr, mask)
+    assert arr.type == dtype
+
+    reloaded = Image.fromarrow(arr, mode, img.size)
+
+    assert reloaded
+
```
radarhere:
```suggestion

```
I don't see the point of this. `fromarrow` always returns an Image instance.

Tests/test_arro3.py
```diff
@@ -0,0 +1,278 @@
...
+    arr_2 = Array(img)
+
+    assert compute.sum(arr_1).as_py() > 0
+    del arr_1
+
+    assert compute.sum(arr_2).as_py() > 0
+    del arr_2
+
+    img2 = img.copy()
+    px = img2.load()
```
radarhere, line 156:
```suggestion
    px = img.load()
```
Or are you using `copy()` because you're testing that the data in `img2` is still correct after `img` is no longer used?

wiredfool: This is to ensure we haven't over freed the memory, as we're refcounting on the arrow usages.

Tests/test_nanoarrow.py
```diff
@@ -0,0 +1,305 @@
...
+
+    arr = nanoarrow.Array(img)
+    _test_img_equals_pyarray(img, arr, mask)
+    assert arr.schema.type == dtype.type
+    assert arr.schema.nullable == dtype.nullable
+
+    reloaded = Image.fromarrow(arr, mode, img.size)
+
+    assert reloaded
+
```
radarhere:
```suggestion

```

PR #9117, opened 2025-07-30
Add libavif support for iOS

freakboy3742:

 * Adds build instructions to add libavif to the compiled dependencies.
 * Bumps cibuildwheel to 3.1.2 to pick up bug fixes and Python 3.14 support for iOS

### DIFF
--- a/.ci/requirements-cibw.txt
+++ b/.ci/requirements-cibw.txt
@@ -1,1 +1,1 @@
-cibuildwheel==3.0.1
+cibuildwheel==3.1.2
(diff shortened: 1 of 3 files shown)


freakboy3742:
This is currently getting stuck on pyroma checks - see #9116.

freakboy3742:
I've spent the morning trying to resolve the build error on x86_64 iOS simulators, and I've come to the conclusion that it can't be done (or, at least, it can't be done without significant extra effort). 

x86_64 builds of libavif require the use of `nasm`, and as far as I can make out `nasm` doesn't produce iOS-compatible binaries. The binaries it produces are rejected at link time as "not an object file".

Given that support for x86_64 iOS simulators is nearing end-of-life (this year's macOS and Xcode major releases will be the last versions that provided support for x86_64), it doesn't seem worth the effort to resolve this issue for x86_64 simulators.

freakboy3742:
The Windows ARM64 failure looks like a network outage.

radarhere (approved):

radarhere:
The plan from https://github.com/python-pillow/Pillow/pull/8858 of using aom only for encoding and dav1d for decoding for the smallest wheel size wasn't working here for x86-64 on iOS, so we have instead used aom for both encoding and decoding on that environment, without aom optimisations for the CPU.

PR #9119, opened 2025-07-30
Drop support for Python 3.9

hugovk:
Python 3.9 reaches end of life in October 2025:

* https://peps.python.org/pep-0596/

Pillow 12.0.0 is planned for 2025-10-15 and will drop Python 3.9:

* https://github.com/python-pillow/Pillow/milestone/49

Notes:

* Only Python 3.9 is available for Cygwin. I don't see any news of upgrades in recent months on the [mailing list](https://sourceware.org/pipermail/cygwin/). This means we need to drop the Cygwin testing from the CI. I've not removed Python/C/test code for Cygwin, we can decide separately if/when to remove that.

* We'll have to decide whether to drop or upgrade the Docker images for Amazon 2 (EOL 2026-06-30), Amazon 2023 (2029-06-30) and CentOS Stream 9 (2027-05-31) from Python 3.9 to something newer. We test CentOS Stream 10 (2030-01-01) with Python 3.12.

* Qt typing refactoring: we can use `a | b` instead of `Union[a, b]` with Python 3.10. But mypy didn't like that for the Qt declarations, so we can instead use the imported ones with `# type: ignore[assignment]` and remove `type: ignore[operator]` elsewhere.


docs/installation/newer-versions.csv
```diff
@@ -1,5 +1,6 @@
 ﻿Python,3.13,3.12,3.11,3.10,3.9,3.8,3.7,3.6,3.5
-Pillow >= 11,Yes,Yes,Yes,Yes,Yes,,,,
+Pillow 12,Yes,Yes,Yes,Yes,,,,,
+Pillow 11,Yes,Yes,Yes,Yes,Yes,,,,
```
radarhere:
```suggestion
Pillow >= 12,Yes,Yes,Yes,Yes,,,,,
Pillow 11.0 - 11.3,Yes,Yes,Yes,Yes,Yes,,,,
```

pyproject.toml
```diff
@@ -190,6 +186,7 @@ lint.ignore = [
   "PT017",  # pytest-assert-in-except
   "PYI026", # flake8-pyi: typing.TypeAlias added in Python 3.10
   "PYI034", # flake8-pyi: typing.Self added in Python 3.11
+  "UP038",  # pyupgrade: deprecated rule
```
radarhere, line 189: Could you talk about why we're adding a deprecated rule?

hugovk: The opposite, ignoring a deprecated rule :)

radarhere: Oh, right!

docs/installation/newer-versions.csv
hugovk: As https://github.com/python-pillow/Pillow/pull/9120#discussion_r2242208186.

hugovk:
> * We'll have to decide whether to drop or upgrade the Docker images for Amazon 2 (EOL 2026-06-30), Amazon 2023 (2029-06-30) and CentOS Stream 9 (2027-05-31) from Python 3.9 to something newer. We test CentOS Stream 10 (2030-01-01) with Python 3.12.

Upgraded in https://github.com/python-pillow/docker-images/pull/244 🚀

.github/workflows/test.yml
```diff
@@ -49,18 +49,13 @@ jobs:
...
         include:
         - { python-version: "3.11", PYTHONOPTIMIZE: 1, REVERSE: "--reverse" }
         - { python-version: "3.10", PYTHONOPTIMIZE: 2 }
         # Free-threaded
         - { python-version: "3.14t", disable-gil: true }
         - { python-version: "3.13t", disable-gil: true }
-        # M1 only available for 3.10+
-        - { os: "macos-13", python-version: "3.9" }
-        exclude:
-        - { os: "macos-latest", python-version: "3.9" }
```
radarhere, line 63: Just to be clear, so you'd like to drop testing for Intel macOS with each PR? We're still generating wheels for the platform, so we should still pick up any problems sooner or later.

radarhere (approved):

.github/workflows/test.yml
hugovk: GitHub Actions will be soon dropping `macos-13` aka Intel:

> The macOS 13 hosted runner image is closing down, following our [N-1 OS support policy](https://github.com/actions/runner-images?tab=readme-ov-file#software-and-image-support). This process will begin September 1, 2025, and the image will be fully retired on November 14, 2025. We recommend updating workflows to use macos-14 or macos-15.

https://github.blog/changelog/2025-07-11-upcoming-changes-to-macos-hosted-runners-macos-latest-migration-and-xcode-support-policy-updates/#macos-13-is-closing-down

So we'll need to drop it at some point. But I've pushed a commit to keep it for now (with 3.10), finding problems sooner is better than later :)

---

At some point, we'll have to decide how long we want to support Intel, and this doesn't necessarily need to be as long as Apple does.

> At WWDC 2025 on June 9, 2025, Apple announced that macOS Tahoe would be the last release of macOS that would support Intel Macs, with macOS 27 in 2026 being exclusive to Macs with Apple silicon.

https://en.wikipedia.org/wiki/Mac_transition_to_Apple_silicon#2025

macOS 26 Tahoe will be released later this year and will likely be supported for three years, until late 2028.

We've generally followed upstream OS support timelines, but test and CI availability also influences ease of maintenance and support. Especially for building wheels. Perhaps we can test via emulation, or have best effort without main CI/wheels so people can self-build, or drop it. Let's see.

radarhere: I've created https://github.com/python-pillow/Pillow/pull/9212

PR #9123, opened 2025-08-02
Move imports into TYPE_CHECKING

radarhere:
Inspired by #9119, this moves some type hint related imports within `if TYPE_CHECKING`


homm:
Why not use standard `from typing import TYPE_CHECKING`?

radarhere:
https://github.com/python-pillow/Pillow/pull/8516#discussion_r2019839169
> We can skip the `typing` import, mypy redefines `TYPE_CHECKING` as `True`

PR #9126, opened 2025-08-02
Use GitHub's automerge instead of Mergify

hugovk:
We first set up Mergify in 2020 to automerge PRs once certain conditions were met: CI has passed, PR is approved, `automerge` label added: https://github.com/python-pillow/Pillow/pull/4931.

Since then, GitHub has added native automerge which is a bit nicer to use: no need to add a label, and while the CI is still running, the "Merge pull request" button becomes an "Enable auto-merge" button, which will do the merge when the CI passes.

Docs: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/automatically-merging-a-pull-request

I've created a ruleset for the `main` branch at 🔒 https://github.com/python-pillow/Pillow/settings/rules/7135275. There are lots of settings, but I've mirrored the Mergify ones:

Require these status checks to pass:

* Docker Test Successful
* Lint
* MinGW
* Test Successful
* Windows Test Successful
* (Cygwin omitted as it'll be removed in https://github.com/python-pillow/Pillow/pull/9119)

Also left this enabled, because we don't want to accidentally delete `main`:

> * Restrict deletions
> Only allow users with bypass permissions to delete matching refs.

We don't need to add a label, because it's now signalled by clicking "Enable auto-merge".

And we don't require explicit approval, because that's also implicit when clicking "Enable auto-merge" (similarly, for a PR where the CI has already passed, we don't require explicit approvals).

If for some reason a status check fails but we still want to merge (maybe some dependency is temporarily failing to download because a server is down), as admins we can still click to bypass the rules and merge.

<details>
<summary>Screenshots of all the settings</summary>

image
image
image

</details>


hugovk:
> If for some reason a status check fails but we still want to merge (maybe some dependency is temporarily failing to download because a server is down), as admins we can still click to bypass the rules and merge.

Updated ruleset for this:

image

hugovk:
Also requires "Allow auto-merge" at https://github.com/python-pillow/Pillow/settings. I also checked "Automatically delete head branches" to keep things tidy:

image

radarhere (approved):

radarhere:
https://github.com/python-pillow/Pillow/pull/9097 and https://github.com/python-pillow/Pillow/pull/8926 are showing an interesting side effect of these changes in settings - when we skip workflows, because we're only building docs or only building wheels, the 'Successful' checks are now just showing as 'pending'.

It doesn't cause any problems, but it makes you think something might be wrong.

hugovk:
Hmm, it actually does cause a problem, because it blocks merge (although you can use admin permissions to bypass it):

image

Let's hold off on this for now, and I've disabled the ruleset ("Enforcement status: Disabled") at https://github.com/python-pillow/Pillow/settings/rules/7135275.

hugovk:
OK, a rebase and a couple more commits:

* https://github.com/python-pillow/Pillow/pull/9126/changes/4d8e7185fdce0b5b388f78f0dec0371986dfeb40 fixes an existing issue: GitHub treats a skipped check as a pass. So make sure the check always runs, and ensure the build result really was a pass (rather than a no-op echo).

* https://github.com/python-pillow/Pillow/pull/9126/changes/2d72447dfe5501719b86e94dffb49ea079c811d8 adds a no-op workflow with `paths` that mirror the `paths-ignore` from the other test workflows. When the normal ones are skipped for a docs build, the no-op one runs and so the expected workflow success jobs are still present.

Next steps:

* Merge
* Enable ruleset: https://github.com/python-pillow/Pillow/settings/rules/7135275
* Check it works
* Uninstall the Mergify app
* Delete the `automerge` label

.github/workflows/test-skipped.yml
```diff
@@ -0,0 +1,58 @@
+# Report success for required status checks when the real test workflows are
+# skipped by their paths-ignore filters, so that, for example, PRs only touching
+# docs can still be merged under a ruleset requiring these checks. See
+# https://docs.github.com/en/pull-requests/how-tos/merge-and-close-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks
+#
+# Keep the paths below in sync with the paths-ignore lists in test.yml,
+# test-docker.yml, test-mingw.yml and test-windows.yml.
```
radarhere, line 7:
```suggestion
# test-docker.yml and test-windows.yml.
```

[test-mingw.yml](https://github.com/python-pillow/Pillow/blob/main/.github/workflows/test-mingw.yml) has no success job

.github/workflows/test-skipped.yml
```diff
@@ -0,0 +1,58 @@
...
+      - name: Success
+        run: echo Tests not required for this change
+
+  test-mingw:
+    runs-on: ubuntu-slim
+    name: MinGW
+    steps:
+      - name: Success
+        run: echo Tests not required for this change
+
```
radarhere, line 43:
```suggestion
```

.github/workflows/test-skipped.yml
```diff
@@ -0,0 +1,58 @@
...
+on:
+  push:
+    branches:
+      - "**"
+    paths:
+      - ".github/workflows/docs.yml"
+      - ".github/workflows/wheels*"
+      - ".gitmodules"
+      - "docs/**"
+      - "wheels/**"
```
radarhere: Do we need to run this on `push`? If it's only used for automerge, then I would have imagined `pull_request` was sufficient.

.github/workflows/test-skipped.yml
```diff
@@ -0,0 +1,58 @@
+# Report success for required status checks when the real test workflows are
+# skipped by their paths-ignore filters, so that, for example, PRs only touching
+# docs can still be merged under a ruleset requiring these checks. See
+# https://docs.github.com/en/pull-requests/how-tos/merge-and-close-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks
+#
+# Keep the paths below in sync with the paths-ignore lists in test.yml,
+# test-docker.yml, test-mingw.yml and test-windows.yml.
+
+name: Test skipped
```
radarhere, line 9:
```suggestion
name: Test Skipped
```
Consistency with other tests.

radarhere (approved):

.github/workflows/test-skipped.yml
hugovk: Ignoring workflows with single-word names:
```yaml
name: Benchmark
name: CIFuzz
name: Docs
name: Lint
name: Test
name: Wheels
```

And those with a proper noun:

```yaml
name: Test MinGW
name: Test Docker
name: Test Windows
name: Test Valgrind
```
There's two in Sentence case:

```yaml
name: Close stale issues
name: Release drafter
```
And just one in Title Case:
```yaml
name: Test Valgrind Memory Leaks
```

So "Tests skipped" is slightly more consistent :)

.github/workflows/test-skipped.yml
hugovk: We don't, I'll remove it.

.github/workflows/test-skipped.yml
hugovk: No success job, but it's a single-job workflow:

https://github.com/python-pillow/Pillow/actions/runs/33146741301

And is one of the required checks:

https://github.com/python-pillow/Pillow/blob/ca443f003557a42c596c9c9c28f103ac4e5f74b6/.github/mergify.yml#L10

.github/workflows/test-skipped.yml
radarhere: Ok, sure

radarhere (approved):

hugovk:
(Can't add `automerge` for one last time on this PR, because it modifies workflow files :)

hugovk:
> Next steps:
> 
> * Merge
> * Enable ruleset: [python-pillow/Pillow/settings/rules/7135275](https://github.com/python-pillow/Pillow/settings/rules/7135275)
> * Check it works
> * Uninstall the Mergify app
> * Delete the `automerge` label



Worked with both a docs PR (https://github.com/python-pillow/Pillow/pull/9879) and a non-docs PR (https://github.com/python-pillow/Pillow/pull/9878) 🎉 

Good timing, rather than deleting the old `automerge` label, we can now archive it without losing the history on old PRs:

https://github.blog/changelog/2026-08-27-label-archiving-is-generally-available/

So `automerge` is now archived:

* https://github.com/python-pillow/Pillow/labels?archived=true
* https://github.com/python-pillow/docker-images/labels?archived=true

And I've uninstalled the Mergify app from:

https://github.com/organizations/python-pillow/settings/installations

hugovk:
Whilst looking at https://github.com/organizations/python-pillow/settings/installations, I've also uninstalled the Azure Pipelines and Travis CI apps, we no longer use those either.

PR #9135, opened 2025-08-05
Fix typos

radarhere:
https://github.com/python-pillow/Pillow/blob/680e68632c48b19fc77c6679741a043824bcaa03/src/PIL/WmfImagePlugin.py#L83-L87

The header is 'placeable'. See https://learn.microsoft.com/en-us/windows/win32/api/gdiplusmetaheader/ns-gdiplusmetaheader-wmfplaceablefileheader or https://en.wikipedia.org/wiki/Windows_Metafile#WMF


hugovk:
There's an "implemetation" in `docs/handbook/third-party-plugins.rst` if you want to fix that at the same time.

radarhere:
Sure, I've updated the commit.

PR #9142, opened 2025-08-09
Make in parallel when building brotli and libavif for macOS and Linux wheels

radarhere:


hugovk:
What sort of difference does this make to build times?

radarhere:
It's hard to say definitively, since the build times don't stay consistent. The last three wheel build times were [1h 13m 36s](https://github.com/python-pillow/Pillow/actions/runs/16839385292), [1h 14m 52s](https://github.com/python-pillow/Pillow/actions/runs/16766057564) and [1h 16m 7s](https://github.com/python-pillow/Pillow/actions/runs/16700240722).

This took [1h 11m 38s](https://github.com/python-pillow/Pillow/actions/runs/16845670746)

Issue #9107, opened 2025-07-24
CI: Enable GitHub Actions CI for ppc64le (Power architecture) support


sandeepgupta12:
Hi team,

We’re reaching out to propose enabling CI support for the ppc64le (IBM Power) architecture in your repository, as part of a broader effort to ensure cross-platform compatibility in the scientific Python ecosystem.

We’re using a GitHub Actions (GHA)-based runner service provided and maintained by IBM to run jobs for the ppc64le architecture. This setup has already been successfully integrated into projects like:

✅ https://github.com/pyca/cryptography/issues/13086

We’d now like to propose enabling the GitHub Actions app in this repository to allow running CI jobs for ppc64le directly via GitHub Actions. This would support upstream compatibility and help ensure continued support for the Power architecture in scikit-learn.

Key Benefits:
🔒 Ephemeral and secure runners, isolated per job

🛠️ Maintained by IBM, requires no setup effort from your side

🔁 Integrates with existing GitHub Actions workflows

📚 Technical documentation and usage details:
https://github.com/IBM/actionspz/tree/main/docs

We’re happy to assist with the setup or provide any additional details the team may need.

Thanks so much!

radarhere:
Hi. We actually already do test ppc64le in our CIs, through Docker and https://github.com/docker/setup-qemu-action. It is listed as one of our tested platforms at https://pillow.readthedocs.io/en/stable/installation/platform-support.html#continuous-integration-targets

You can see the CI script at https://github.com/python-pillow/Pillow/blob/main/.github/workflows/test-docker.yml, and a run of the job at https://github.com/python-pillow/Pillow/actions/runs/16298847593/job/46027846471

For the Dockerfile itself, see https://github.com/python-pillow/docker-images/blob/main/ubuntu-24.04-noble-ppc64le/Dockerfile

sandeepgupta12:
Hi, thanks for the quick response!

Great to see that ppc64le is already being tested using QEMU in your CI. Our main goal is to enable official wheel builds and uploads for ppc64le to PyPI. This could be achieved either using your existing QEMU-based setup or, if needed, by adding jobs that run on IBM-hosted self-hosted runners (native ppc64le hardware) via GitHub Actions.

Would you be open to installing the GitHub Actions app if we find native builds are needed for wheel publishing?

radarhere:
[...]

PR #9145, opened 2025-08-10
Add manylinux_2_28 ppc64le wheels

radarhere:
https://github.com/python-pillow/Pillow/issues/9107 requests ppc64le wheels.

While there is the potential for using self-hosted runners to generate these quickly, if we don't want to go down that route, then here is another option - using QEMU to create them, but with minimal dependencies to reduce the build time. In the build in this PR, ppc64le jobs are not the longest.


sandeepgupta12:
Just following up on regarding the ppc64le wheels. Are there any updates or feedback needed to move it forward?

radarhere:
Simply waiting for review from another Pillow team member.

aclark4life:
@sandeepgupta12 Can you test these wheels before we merge ?

sandeepgupta12:
Sure, I can test the wheels on a ppc64le machine and let you know the results.

sandeepgupta12:
@radarhere 
I tested these wheels on a native ppc64le machine. The build works, but I see some test failures:
```
FAILED Tests/test_file_iptc.py::test_open - AssertionError: assert [_Tile(codec_..., args='raw')] == [('iptc', (0,...'raw', None))]
FAILED Tests/test_file_iptc.py::test_layers[3-RGB] - AssertionError: assert 'R' == 'RGB'
FAILED Tests/test_file_iptc.py::test_layers[4-CMYK] - AssertionError: assert 'C' == 'CMYK'
FAILED Tests/test_image.py::TestImage::test_show - Failed: DID NOT WARN. No warnings of type (<class 'DeprecationWarning'>,) were emitted.
```

Looking at the build script, on ppc64le the build function runs build_libjpeg_turbo but then exits early:
```bash
if [[ "$AUDITWHEEL_ARCH" == "ppc64le" ]]; then
    return
fi
```

Because of this, other dependencies are not built, so the resulting wheel has incomplete image format support. That seems to explain the above test failures.

radarhere:
I would be surprised if the lack of image support was the reason for the test failures. The test failures sound more like #9088, which was merged after the last time I updated this PR. So I suspect the problem is that you built from this PR, and then tested main.

To simplify things, I've now merged main into this PR again, so that change should now be included. Could you try building and testing again?

sandeepgupta12:
@radarhere @aclark4life Thanks for updating the branch. I retested on a native ppc64le machine, and everything is working fine now ✅. I think we can go ahead and merge this PR.

hugovk:
How long does this take to build? How many wheels does it create? How big are they?

radarhere:
Oops, accidentally deleted this branch. I've recreated it in #9216

The longest run is 1h 1m. This creates 7 manylinux_2_28 wheels (all active CPython versions, including threaded). They are each 1.9mb.

Issue #9158, opened 2025-08-17
Feature request: Check if glyph exists in font


glebm:
Using the freetype-py library, I can check if a glyph exists in a font like this:

```python
ft_font = freetype.Face(font_path)
exists = ft_font.get_char_index(0x41) != 0
```

It'd be nice to have this functionality exposed in Pillow as well.

radarhere:
Are you interested in immediate solutions at all? I suspect no, since using freetype-py would be a viable immediate solution.

Anyway, for an immediate Pillow-only solution, I would suggest comparing the character's output to the output of the replacement character, U+FFFD.
```python
def checkIfCharacterIsMissing(fontPath, text):
    font = ImageFont.truetype(fontPath)

    w, h = font.getbbox(text)[2:]
    im = Image.new("L", (w, h), 255)
    draw = ImageDraw.Draw(im)
    draw.text((0, 0), text, font=font)

    w, h = font.getbbox("\ufffd")[2:]
    if im.size != (w, h):
        return False
    im2 = Image.new("L", (w, h), 255)
    draw = ImageDraw.Draw(im2)
    draw.text((0, 0), "\ufffd", font=font)

    return im.tobytes() == im2.tobytes()
```

glebm:
I'm using freetype-py for this at the moment, it'd just be nice to be able to do everything via pillow.

https://github.com/diasurgical/devilutionx-font-converter/blob/ac0f0e48bd690ace783c6c1b16e52b569e46f551/src/devilutionx_font_converter/convert.py#L178-L183

It'd also be nice if glyph properties (such as `horiAdvance`) were exposed via pillow.

radarhere:
I can understand why a Pillow user might want to know if a glyph existed or not - so you can avoid drawing it with Pillow.

I'm less convinced that knowing the glyph properties directly fits into Pillow's overall goal. Could you explain to me how it would help a Pillow user generate an image? If you would like it in order to convert fonts, that indeed sounds like a job better suited to freetype-py.

glebm:
I'm using pillow to generate textured bitmap fonts for a game from OTF fonts.

For later rendering in-game, the width of each rasterized glyph should be its horizontal advance, which can be larger than the glyph itself. Knowing the horizontal advance lets us know the image width.

radarhere:
[...]

Issue #7380, opened 2023-09-06
Some .woff2 fonts return incorrect images


natuan62:
### What did you do?

- Create thumbnail font for .woff2 

### What did you expect to happen?

- image thumbnail correct with text and font family

### What actually happened?

- image after generate incorrect

### What are your OS, Python and Pillow versions?

* OS: Mac Os 13.5.1
* Python: 3.11.4
* Pillow: 10.0.0

```python
from PIL import Image, ImageDraw, ImageFont

# FONT_NAME = "From Cartoon Blocks.woff2"
FONT_NAME = 'Karla.woff2'

font = ImageFont.truetype(
    FONT_NAME, 100, encoding="utf-8", layout_engine=ImageFont.Layout.RAQM
)
_, _, offset_w, offset_h = font.getbbox(FONT_NAME)

W = offset_w
H = offset_h
image = Image.new("RGBA", (W, H), (255, 0, 0, 0))
draw = ImageDraw.Draw(image)
pos = ((W - offset_w) / 2, (H - offset_h) / 2)
draw.text(pos, FONT_NAME, "white", font=font)
# image.save("test.png")
image.save("test2.png")
```

![image](https://github.com/python-pillow/Pillow/assets/10808157/e156f25b-8e86-4dc1-a1fd-0324ec310192)

[test-fonts.zip](https://github.com/python-pillow/Pillow/files/12531673/Archive.zip)

radarhere:
I presume that 'From Cartoon Blocks.woff2' is fine, and you're just demonstrating what you expect.

If I upload your Karla.woff2 to https://fontdrop.info/, I see that some letters are missing? I expect that the font file is damaged in some way.

![karla](https://github.com/python-pillow/Pillow/assets/3112309/c6d250bc-d64a-4ed5-820b-4778aeca6eab)

If I download https://github.com/googlefonts/karla/blob/main/fonts/webfonts/Karla-Regular.woff2, that font file works fine. I'm able to use that file and your code (after changing the colour to black so you can see the image here on GitHub) to generate the following
![test2](https://github.com/python-pillow/Pillow/assets/3112309/a93883c8-8643-4d47-b115-32df26ce49d4)

natuan62:
@radarhere Thanks for your quick reply. 

So what list glyphs need exists (can't missing) for pillow can generate correct thumbnail ?
Maybe all characters from A to Z and numbers 0 to 9 must exists ?
[...]

Issue #4808, opened 2020-07-22
Backup font for missing characters when drawing text


Markxy:
### Description of the request
I want to use a specific font but also draw characters which are not available in that font. Replace them with something generic, like Arial-Bold for example.
Similarly to how web browsers work when they are missing fonts - the OS fills that void with a default font.

```python
from PIL import Image, ImageFont, ImageDraw


temp_canvas = Image.new("RGBA", (1200, 300), (255, 255, 255, 255))
draw_canvas = ImageDraw.Draw(temp_canvas, "RGBA")

font = ImageFont.truetype(r"C:\fonts\BarlowSemiCondensed-Bold.ttf", size=150)
text_string = "hello ಠಠ world"

draw_canvas.text((100, 100), text_string, fill="#000000", font=font)

temp_canvas.show()
```
The output is: 
![image](https://user-images.githubusercontent.com/25804528/88179691-f3de0180-cc34-11ea-97e8-b7654cb05cc2.png)

### Proposed solution
Add an argument to `ImageDraw.Draw().text()` like `backup_font` , which would be used when the first font doesn't have the character specified in `text_string` . Like `ಠ` in this example

nulano:
I don't think FreeType (the library Pillow uses to load fonts) is able to combine multiple typefaces into one. This change would be simple with basic layout (Pillow can try loading each glyph from a list of fonts until one succeeds), but Raqm (used for complex scripts) seems to require passing a single typeface to be used for the whole string. So for complex layout this would require a change in Raqm, either upstream or by including a vendored version in Pillow.

The latter could help with some of the distribution issues that have appeared in the past (the dynamic loading would be moved to FriBiDi; users would only need to install LGPL-licensed FriBiDi to enable complex text, not Raqm which is sometimes available in an outdated and unsupported version in some linux distributions). It could also potentially help #3066.

As for the API to use here, I would propose `ImageFont.truetype_family(font1, font2, font3, ...)` to create a compound font, which would get special handling in the C code.

lateus:
This can be specially helpful with emojis. Many emojis font only contains emojis, and the ones that contains regular glyph have styles that can break the image/poster paradigm that one wants to achieve. However this can be tricky because all color emoji fonts that I've tested so far requires exactly a size of `109`, which would require super-sampling and then smooth resizing.

voussoir:
Hi, I'm interested in this problem too.

[...]

PR #9160, opened 2025-08-17
Added FreeTypeFont has_characters()

radarhere:
Helps #9158. The issue requested a method to check if glyphs are present in a font. This question has arisen previously in https://github.com/python-pillow/Pillow/issues/7380#issuecomment-1708041463, and I think this would also help users from #4808, in a less elegant way.

This adds `FreeTypeFont.has_characters(text)`. I expect that the issue was only thinking about checking a single character at a time, but if user wanted to check multiple characters at once, I don't see any reason not to let them. If any character in the string is missing, then `False` is returned.

The C code in here is based on https://github.com/python-pillow/Pillow/blob/9d39fe6adac233ccd26ba20bff9adda53a1ec89d/src/_imagingft.c#L432-L463


hugovk:
I'm on the fence with this. It feels outside the core focus of Pillow, and it's not too tricky to do this with Pillow in Python:

https://github.com/python-pillow/Pillow/issues/9158#issuecomment-3194350459

Plus it's a even easier with a dedicated FreeType library:

https://github.com/python-pillow/Pillow/issues/9158#issue-3328268124

glebm:
A font is roughly a collection of images and checking if an image exists in the collection is pretty basic functionality?

Of course it's doable in other ways but:
1. The pillow way suggested is verbose and inefficient.
2. The freetype way requires loading the font twice (once in Pillow, once in Freetype).

radarhere:
If there is no consensus on this, should the issue be closed? We can always return to the idea again if someone else raises it in the future.

aclark4life:
Any downside assuming we determine it's in scope?

radarhere:
To adding the code? Only the usual theoretical maintenance burden. But it's a new method, so it doesn't affect anything else, and it's relatively simple code, so I can't foresee any problems.

To returning to it later? No. I don't see us refactoring the underlying font handling that this method relies on anytime soon.

radarhere:
This has been 'on the fence' for six months, so I'll just close.

PR #9161, opened 2025-08-18
Use macos-14 for iOS arm64 simulator

radarhere:
The iOS arm64 simulator has started failing in [main](https://github.com/python-pillow/Pillow/actions/runs/17002152562/job/48205826204).

Based on https://github.com/pypa/cibuildwheel/pull/2557, @freakboy3742's suggested solution is to use macos-14 instead of macos-latest.


freakboy3742:
Confirming that this appears to be the best option at present. 

I've [logged the problem with Github](https://github.com/actions/runner-images/issues/12777), as it seems to indicate a bigger problem with how they've configured iOS simulators.

PR #9165, opened 2025-08-20
Allow pyright to "see" core properly

jmarianer:
Without this change, the following file raises errors (in strict mode) and reveals the type as "Unknown". After this change, the type is correctly revealed as "ImagingCore".

```python
  # pyright: strict

  from PIL import Image

  img = Image.new('RGB', (10, 10))
  data = img.getdata()
  reveal_type(data)
```

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -86,38 +86,41 @@ WARN_POSSIBLE_FORMATS: bool = False
 MAX_IMAGE_PIXELS: int | None = int(1024 * 1024 * 1024 // 4 // 3)
 
 
-try:
-    # If the _imaging C module is not present, Pillow will not load.
-    # Note that other modules should not refer to _imaging directly;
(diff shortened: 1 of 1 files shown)


jmarianer:
I don't understand the purpose of that block, to be honest. It looks like it was originally meant to allow use of some Python stuff without the C part, but that was changed in 2013 to "raise anyway". In any case, I'm just moving it into an "if not TYPE_CHECKING" thing and not changing anything except formatting.

radarhere:
From my testing, I think #9166 fixes your problem as well.

jmarianer:
So it does! I'm happy to close this in favor of yours. I didn't want to mess with that for obvious reasons, but you know this code better than I do. :)

hugovk:
Thanks, closing in favour of https://github.com/python-pillow/Pillow/pull/9166.

PR #9169, opened 2025-08-21
FLI/FLC _accept: Check for 8-bit colour depth

### LINKED ISSUE #9168: Some TIFF files are identified as FLI/FLC files
We operate and develop a digital preservation system. One of our users has TIFF files, which Pillow's FliImagePlugin identifies as FLI/FLC files. I can not share these files, but I can give you some details of them.

### What did you do?
Here is a simple test code
```python
import PIL.Image

image = PIL.Image.open("testfile.tiff")
print(image.format)
print(image.get_format_mimetype())
```
The output should be
```
TIFF
image/tiff
```
### What actually happened?
The output is
```
FLI
None
```

* OS: Fedora 42 (a container)
* Python: Python 3.13.7
* Pillow: main branch compiled/installed from soure

```
Pillow 12.0.0.dev0
Python 3.13.7 (main, Aug 14 2025, 00:00:00) [GCC 15.2.1 20250808 (Red Hat 15.2.1-1)]
Python executable is /home/vvainio/scratch/Pillow/venv/bin/python
Environment Python files loaded from /home/vvainio/scratch/Pillow/venv
Python Pillow modules loaded from /home/vvainio/scratch/Pillow/venv/lib64/python3.13/site-packages/PIL
Binary Pillow modules loaded from /home/vvainio/scratch/Pillow/venv/lib64/python3.13/site-packages/PIL
--- PIL CORE support ok, compiled for 12.0.0.dev0
--- ZLIB (PNG/ZIP) support ok, loaded 1.3.1.zlib-ng, compiled for zlib-ng 2.2.5
--- RAQM (Bidirectional Text) support ok, loaded 0.10.1
--- LIBIMAGEQUANT (Quantization method) support ok, loaded 4.2.2
```

### Additional information
In hex code, our example image starts like this:
`4949 2a00 12af cb02`

Probably due to software used to produce these files, the TIFF IFD location has 0x12 as byte 4 and 0xAF as byte 5 in the file. This is enough to have FliImagePlugin consider it an FLI file. Byte 14 is 0 (and so are bytes 20-22, even though those are checked outside of `_accept()`). Bytes 14 and 20-22 are already image data of strip 1, so they are outside the TIFF header.

The FLI and FLC detection syntax in file/libmagic (here https://github.com/file/file/blob/master/magic/Magdir/animation#L878) checks that byte 12 is always 8, as in 8-bit colour depth. That check would be enough to differentiate these TIFFs from FLIs/FLCs.

The page https://www.compuphase.com/flic.htm says about this about the type field in the FLI/FLC file header:
"0xAF11	For FLI files
0xAF12	For standard FLC files (with an 8-bit colour depth)."

Checking for 8-bit colour depth would work here, because Pillow does not seem to support types 0xAF44, 0xAF30 or 0xAF31 at all. I'll add a PR about that.

Checking for the reserved areas of the FLI/FLC file would make the detection even better. You already check for bytes 20-22 in `_open()`. The other reserved areas are in offsets 48-72 and 80-120. These checks can not be made in `_accept()`, because the `_accept()` `prefix` is only 16 bytes long, so they would not solve this issue.

One option would be to increase the priority of detecting TIFF files. I'm quite certain there are more TIFF files than FLI/FLC files around. Could you consider adding TIFF as a format to `preinit()` in Image.py? Or is there some other way to control the priority?

vpv-csc:
Fixes #9168 .

 * Standard FLI/FLC files always have 8-bit color depth. This change helps avoid erroneously detecting some TIFFs as FLI/FLC files.
 * Update test files Tests/images/timeout-9139147ce93e20eb14088fe238e541443ffd64b3.fli and Tests/images/timeout-bff0a9dc7243a8e6ede2408d2ffa6a9964698b87.fli to claim to have 8-bit colour depth so that the tests related to them will still work.


radarhere:
Just to link to documentation, https://www.fileformat.info/format/fli/egff.htm describes the header as
```
typedef struct _FlicHeader
{
  DWORD FileSize;       /* Total size of file */
  WORD  FileId;         /* File format indicator */
  WORD  NumberOfFrames; /* Total number of frames */
  WORD  Width;          /* Screen width in pixels */
  WORD  Height;         /* Screen height in pixels */
  WORD  PixelDepth;     /* Number of bits per pixel */
...
PixelDepth indicates the number of bits per pixel; the value of this field is always 08h.
```

radarhere:
Under 'Frequent errors', https://www.compuphase.com/flic.htm lists 'The colour depth is set to zero'. Is it worth also accepting zero as the value?

src/PIL/FliImagePlugin.py
```diff
@@ -32,6 +32,7 @@ def _accept(prefix: bytes) -> bool:
     return (
         len(prefix) >= 6
         and i16(prefix, 4) in [0xAF11, 0xAF12]
+        and i16(prefix, 12) == 8  # 8-bit colour
```
radarhere, line 35:
```suggestion
        and i16(prefix, 12) == [0, 8]  # depth
```

vpv-csc:
> Is it worth also accepting zero as the value?

[The `file` magic accepts only 8](https://github.com/file/file/blob/master/magic/Magdir/animation#L878). Based on that, I'd say no, but I'll leave the decision to you (the maintainers).

We have received two ”false positive” TIFFs from our users. One of them has `0` as byte 12 and one of them has `1`. Our problem would not then be completely fixed. But hypothetically there could be a similar TIFF file, which had `8` as byte 12, as well. With the 16-byte limit of `prefix` this can never be perfect.

radarhere:
I've created #9183 to also help the issue. If you don't feel that changing `_accept` is sufficient, then is that better than this? Or would you like both PRs?

vpv-csc:
I concentrated too much on `_accept` when trying to find a solution. #9183 actually fixes this issue as well, for both of our test files. This PR is not absolutely necessary. If you are worried about the 0 vs. 8 case, then I guess you can drop this PR.

radarhere:
#9183 has been merged.

PR #9179, opened 2025-08-31
Removed shebang lines and executable flags

radarhere:
`checks/32bit_segfault_check.py` and `checks/check_imaging_leaks.py` have shebang lines and executable flags.

This adds those to the other scripts in the `checks` directory.

### DIFF
--- a/Tests/createfontdatachunk.py
+++ b/Tests/createfontdatachunk.py
@@ -1,4 +1,3 @@
-#!/usr/bin/env python3
 from __future__ import annotations
 
 import base64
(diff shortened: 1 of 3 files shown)


hugovk:
Shall we go the other way: rather than changing 14 files, change the other two to remove shebangs and executable bits?

I usually run scripts by specifying the Python interpreter, so I can be explicit which Python version is used.

radarhere:
Ok, I've updated the commit, and did the same for one of the Tests files as well.

PR #9206, opened 2025-09-14
Updated FreeType to 2.14.1 on Windows

radarhere:
Resolves #9205

The [last runner image published](https://github.com/actions/runner-images/pull/12897#issuecomment-3306615719) for macOS 13 only had FreeType 2.14.0. 2.14.0 exactly has a [bug with variable fonts](https://gitlab.freedesktop.org/freetype/freetype/-/issues/1355), but there's no perfect workaround, so I've manually built FreeType for macOS 13 instead. However, the job might be removed anyway in https://github.com/python-pillow/Pillow/pull/9212

PR #9212, opened 2025-09-18
Stop testing macOS Intel

radarhere:
The macOS 13 runner image will start closing down in 12 days - https://github.blog/changelog/2025-07-11-upcoming-changes-to-macos-hosted-runners-macos-latest-migration-and-xcode-support-policy-updates/#macos-13-is-closing-down

HomeBrew has dropped support for macOS 13 - https://docs.brew.sh/Installation#macos-requirements

You may notice that the macOS jobs are failing here. https://github.com/python-pillow/Pillow/pull/9206 is required to fix them.


hugovk:
> * trying to use [macos-15-intel] for iOS results in an error
> 
> > Testing failed:
> > iOSTestbed encountered an error (Failed to prepare device 'Clone 1 of iPhone 16e' for impending launch. (Underlying Error: No matching device (5C6CB0A3-D0E1-465A-B035-84E8CB59187A) in set at /Users/runner/Library/Developer/XCTestDevices))
> 
> so instead I've switched to macos-14.

This means switching from Intel to Apple Silicon, so the build is emulated and not tested. Anything else of note?

cc @freakboy3742

freakboy3742:
I haven't tried anything on the macOS-15-intel runner yet; the press release certainly made it sound like it should be a drop-in replacement, but that's evidently not the case. 

The error is very similar to what is happening on the macOS-15 ARM runners, though; so it's possible this might be fixed by the changes to the iOS testbed that will be in the next iOS binary support releases.

.github/workflows/wheels.yml
```diff
@@ -103,7 +103,7 @@ jobs:
             cibw_arch: arm64_iphonesimulator
           - name: "iOS x86_64 simulator"
             platform: ios
-            os: macos-13
+            os: macos-14
```
freakboy3742: With the latest cibuildwheel release, this should now be able to be:
```suggestion
            os: macos-15-intel
```

freakboy3742:
@radarhere FYI - I've pushed an update to this PR as #9249, with a couple of additional tweaks, and it's now passing CI. I don't mind which of the two PRs are used.

.github/workflows/wheels.yml
radarhere: Thanks. I've added a commit for this.

hugovk:
Thanks both!

PR #9216, opened 2025-09-21
 Add manylinux_2_28 ppc64le wheels

radarhere:
Resolves https://github.com/python-pillow/Pillow/issues/9107

While there is the potential for using self-hosted runners to generate these quickly, if we don't want to go down that route, then here is another option - using QEMU to create them, but with minimal dependencies to reduce the build time. In the build in this PR, ppc64le jobs are not the longest.

Recreation of https://github.com/python-pillow/Pillow/pull/9145


sandeepgupta12:
@aclark4life @hugovk 
Can we merge this PR?

PR #9217, opened 2025-09-22
Updated FreeType to 2.14.1 on macOS wheels

radarhere:
freetype 2.14.1 has been released - https://freetype.org/


.github/workflows/wheels-dependencies.sh
```diff
@@ -314,6 +318,10 @@ function build {
 
     if [[ -n "$IS_MACOS" ]]; then
         # Custom freetype build
+        if [[ -z "$IOS_SDK" ]]; then
+          build_simple sed 4.9 https://mirrors.middlendian.com/gnu/sed
+        fi
```
radarhere, line 323: freetype 2.14.0 introduced a bug where libpng is not always correctly detected - https://gitlab.freedesktop.org/freetype/freetype/-/issues/1358. A workaround described in that issue is to use gnu-sed.

PR #9059, opened 2025-07-02
Restored manylinux2014 wheels

radarhere:
Resolves #9057

Restores manylinux2014 wheels, after the default change in https://github.com/pypa/cibuildwheel/pull/2330 with [cibuildwheel 3.0.](https://github.com/python-pillow/Pillow/pull/9010)

This only affects future releases. For Pillow 11.3.0, I could try and use twine to upload additional wheels manually.

Current CI failures should be resolved by https://github.com/python-pillow/Pillow/pull/9053

PR #9239, opened 2025-10-10
Wheels CI: Check number of expected dists

hugovk:
Sometimes changes to cibuildwheel (such as [#9059](https://github.com/python-pillow/Pillow/pull/9059)) change the defaults and more or less wheels are built, and we don't notice until we happen to spot an unnecessary wheel, or someone reports a missing wheel for their system.

This adds a hardcoded number of dists created by the CI (`EXPECTED_DISTS`), and fails if it's different to what's actually produced.

* If it changes when expected (as would happen for [#9214](https://github.com/python-pillow/Pillow/pull/9214), [#9030](https://github.com/python-pillow/Pillow/pull/9030), [#9031](https://github.com/python-pillow/Pillow/pull/9031)), all good, we can change the constant.
* Otherwise unexpected changes (as would happen for [#9010](https://github.com/python-pillow/Pillow/pull/9010)) can be investigated first.

This is inspired by the coverage.py CI: https://github.com/nedbat/coveragepy/blob/8827634adcfe4a4baf71cf0b3e1cd967cf9b14a2/.github/workflows/publish.yml#L76-L81

To demonstrate, right now it will fail:
```
Number of dists, should be 98:
91
```

image

https://github.com/python-pillow/Pillow/actions/runs/18407980303/job/52459987053?pr=9239

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -306,20 +306,3 @@ jobs:
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
           attestations: true
-
-  count-dists:
-    needs: [build-native-wheels, windows, sdist]
-    runs-on: ubuntu-latest
-    name: Count dists
-    steps:
-      - uses: actions/download-artifact@v4
-        with:
-          pattern: dist-*
-          path: dist
-          merge-multiple: true
-      - name: "What did we get?"
-        run: |
-          ls -alR
-          echo "Number of dists, should be $EXPECTED_DISTS:"
-          ls -1 dist | wc -l
-          files=$(ls dist 2>/dev/null | wc -l) && [ "$files" -eq $EXPECTED_DISTS ] || exit 1


hugovk:
The failing run: https://github.com/python-pillow/Pillow/actions/runs/18407980303/job/52459987053?pr=9239

And force pushed with the correct number for it to pass.

.github/workflows/wheels.yml
```diff
@@ -287,3 +288,20 @@ jobs:
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
           attestations: true
+
+  count-dists:
+    needs: [build-native-wheels, windows, sdist]
+    runs-on: ubuntu-latest
+    name: Count dists
+    steps:
+      - uses: actions/download-artifact@v4
```
```suggestion
      - uses: actions/download-artifact@v5
```

hugovk:
Updated to require `count-dists` to pass before publishing any wheels:

image

https://github.com/python-pillow/Pillow/actions/runs/18428446075?pr=9239

PR #9244, opened 2025-10-12
When building scheduled wheels, run count-dists as a successful no-op

radarhere:
Alternative to https://github.com/python-pillow/Pillow/pull/9248

Normally, our scheduled Wheels jobs update wheels to scientific-python-nightly-wheels - https://github.com/python-pillow/Pillow/actions/runs/18331606908

Our most recent scheduled Wheels jobs did not - https://github.com/python-pillow/Pillow/actions/runs/18437781793

This is because https://github.com/python-pillow/Pillow/pull/9239 made 'scientific-python-nightly-wheels-publish' require 'count-dists', which requires 'sdist', but 'sdist' does not run when scheduled.

This PR suggests always running 'sdist', but if the jobs are scheduled, skip all the steps within it.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -232,26 +232,31 @@ jobs:
           path: winbuild\build\bin\fribidi*
 
   sdist:
-    if: github.event_name != 'schedule'
+    if: github.event_name != 'schedule' || github.repository_owner == 'python-pillow'
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v5
+      if: github.event_name != 'schedule'
       with:
         persist-credentials: false
 
     - name: Set up Python
       uses: actions/setup-python@v6
+      if: github.event_name != 'schedule'
       with:
         python-version: "3.x"
 
-    - run: make sdist
+    - if: github.event_name != 'schedule'
+      run: make sdist
 
     - uses: actions/upload-artifact@v4
+      if: github.event_name != 'schedule'
       with:
         name: dist-sdist
         path: dist/*.tar.gz
 
   count-dists:
+    if: github.event_name != 'schedule' || github.repository_owner == 'python-pillow'
     needs: [build-native-wheels, windows, sdist]
     runs-on: ubuntu-latest
     name: Count dists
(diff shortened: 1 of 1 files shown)


hugovk:
Why does 'sdist' does not run when scheduled? To only upload wheels to scientific-python-nightly-wheels and avoid a ~50 MB zip?


---

Or perhaps unhook the 'count-dists' check from the scientific-python-nightly-wheels flow?

image

radarhere:
> Why does 'sdist' does not run when scheduled? To only upload wheels to scientific-python-nightly-wheels and avoid a ~50 MB zip?

It turns out [I suggested it](https://github.com/lysnikolaou/Pillow/pull/1) a year ago. I expect I just thought it wasn't necessary to upload it. If you'd like to though, I have no objection. I've created https://github.com/python-pillow/Pillow/pull/9248

> Or perhaps unhook the 'count-dists' check from the scientific-python-nightly-wheels flow?

All other things being equal, I think it would be better to know if we suddenly start uploading extra wheels to scientific-python-nightly-wheels.

hugovk:
> > Why does 'sdist' does not run when scheduled? To only upload wheels to scientific-python-nightly-wheels and avoid a ~50 MB zip?
> 
> It turns out [I suggested it](https://github.com/lysnikolaou/Pillow/pull/1) a year ago. I expect I just thought it wasn't necessary to upload it. If you'd like to though, I have no objection. I've created #9248

I checked a few others at https://anaconda.org/scientific-python-nightly-wheels/ and didn't see any sdist, so probably better not to upload a fairly large file. And the clue's in the (scientific-python-nightly-wheels) name.

> > Or perhaps unhook the 'count-dists' check from the scientific-python-nightly-wheels flow?
> 
> All other things being equal, I think it would be better to know if we suddenly start uploading extra wheels to scientific-python-nightly-wheels.

Agreed.

How about we use https://github.com/python-pillow/Pillow/pull/9248 and do something like `rm pillow*.tar.gz` in the scientific-python-nightly-wheels-publish job just before upload?

radarhere:
> How about we use https://github.com/python-pillow/Pillow/pull/9248 and do something like rm pillow*.tar.gz in the scientific-python-nightly-wheels-publish job just before upload?

Sure. I've updated the PR, just changing the 'pattern' used for matching in scientific-python-nightly-wheels-publish.

hugovk:
Closed in favour of https://github.com/python-pillow/Pillow/pull/9248.

Issue #9178, opened 2025-08-30
Release Pillow 12.0.0 on October 15, 2025


radarhere:
## Main release

Released quarterly on January 2nd, April 1st, July 1st and October 15th.

* [x] Open a release ticket e.g. https://github.com/python-pillow/Pillow/issues/3154
* [x] Develop and prepare release in `main` branch.
  * [x] Add release notes https://github.com/python-pillow/Pillow/pull/9247
* [x] Check [GitHub Actions](https://github.com/python-pillow/Pillow/actions) to confirm passing tests in `main` branch.
* [x] Check that all the wheel builds pass the tests in the [GitHub Actions "Wheels" workflow](https://github.com/python-pillow/Pillow/actions/workflows/wheels.yml) jobs by manually triggering them.
* [x] In compliance with [PEP 440](https://peps.python.org/pep-0440/), update version identifier in `src/PIL/_version.py`
* [x] Run pre-release check via `make release-test` in a freshly cloned repo.
* [x] Create branch and tag for release e.g.:
  ```bash
  git branch 12.0.x
  git tag 12.0.0
  git push --tags
  ```
* [x] Check the [GitHub Actions "Wheels" workflow](https://github.com/python-pillow/Pillow/actions/workflows/wheels.yml) has passed, including the "Upload release to PyPI" job. This will have been triggered by the new tag.
* [x] Publish the [release on GitHub](https://github.com/python-pillow/Pillow/releases).
* [x] In compliance with [PEP 440](https://peps.python.org/pep-0440/), increment and append `.dev0` to version identifier in `src/PIL/_version.py` and then:
  ```bash
  git push --all
   ```

## Publicize release

* [x] Announce release availability via [Mastodon](https://fosstodon.org/@pillow) e.g. https://fosstodon.org/@pillow/110639450470725321

## Documentation

* [x] Make sure the [default version for Read the Docs](https://pillow.readthedocs.io/en/stable/) is up-to-date with the release changes

## Docker images

* [x] Update Pillow in the Docker Images repository
  ```bash
  git clone https://github.com/python-pillow/docker-images
  cd docker-images
  ./update-pillow-tag.sh 12.0.0
  ```

radarhere:
Highlights:

* Adding support for Python 3.14: https://github.com/python-pillow/Pillow/pull/9120
* Dropping Python 3.9: https://github.com/python-pillow/Pillow/pull/9119
[...]

PR #9258, opened 2025-10-15
Use macos-14 for iOS arm64 simulator

hugovk:
Reverts python-pillow/Pillow#9250.

See https://github.com/python-pillow/Pillow/issues/9178#issuecomment-3407202630 for more info.


hugovk:
I cancelled all the wheel jobs to just restart the relevant one so we don't need to wait: https://github.com/python-pillow/Pillow/actions/runs/18535048742/job/52828212057?pr=9258

hugovk:
OK, that passed. Will merge and restart the release in a bit.

PR #9260, opened 2025-10-16
Remove Fedora 41

radarhere:
Fedora 41 becomes EOL on November 19th, before the next release of Pillow - https://fedorapeople.org/groups/schedule/f-43/f-43-key-tasks.html


kmilos:
Perhaps it's more convenient to wait until f43 (or any N+2 in the future) is released and do just one changeover?

(It would seem there is a pre-release docker image available already.)

hugovk:
It saves us a bit of CI time, so we might as well remove it already.

It's only about 2.5 mins, but it also frees up a runner, so that does help as we have over 50 jobs in total.

radarhere:
Fedora 43 has now been added in https://github.com/python-pillow/Pillow/pull/9290

PR #9261, opened 2025-10-16
Edit getdata to recommend iterating in a type-correct way

DominicOram:
https://github.com/python-pillow/Pillow/blob/5122c8356d6fa8f4a0d8449e8875734bb5d4d966/src/PIL/Image.py#L1436 recommends that people use `list(im.getdata())` for e.g. printing data. This works fine in production but type checkers (such as `pyright`) do not like it as `ImagingCore` does not expose an `__iter__`. This adds that so that the type checking is happy.

## Notes
* It would be better to not say that the `__iter__` returns `Any` but I do not know the codebase well enough to know if there are variations in what types it might return. `Any` is sufficient to fix the type checker for my usecase

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -1159,7 +1159,8 @@ class Image:
             if new_im.mode == "P" and new_im.palette:
                 try:
                     new_im.info["transparency"] = new_im.palette.getcolor(
-                        cast(tuple[int, ...], trns), new_im  # trns was converted to RGB
+                        cast(tuple[int, ...], trns),
+                        new_im,  # trns was converted to RGB
                     )
                 except ValueError as e:
                     del new_im.info["transparency"]
@@ -1433,7 +1434,7 @@ class Image:
         Note that the sequence object returned by this method is an
         internal PIL data type, which only supports certain sequence
         operations.  To convert it to an ordinary sequence (e.g. for
-        printing), use ``list(im.getdata())``.
+        printing), use ``list(iter(im.getdata()))``.
 
         :param band: What band to return.  The default is to return
            all bands.  To return a single band, pass in the index
@@ -3224,7 +3225,8 @@ class SupportsArrowArrayInterface(Protocol):
     """
 
     def __arrow_c_array__(
-        self, requested_schema: "PyCapsule" = None  # type: ignore[name-defined]  # noqa: F821, UP037
+        self,
+        requested_schema: "PyCapsule" = None,  # type: ignore[name-defined]  # noqa: F821, UP037
     ) -> tuple["PyCapsule", "PyCapsule"]:  # type: ignore[name-defined]  # noqa: F821, UP037
         raise NotImplementedError()
 


DiamondJoseph:
Just guessing from the other methods, would it be 

```python
...

    def __getitem__(self, index: int) -> float: ...
    def __iter__(self) -> Iterator[float]: ...
```

Or are other elements returned also?

radarhere:
My suggestion would be `float | tuple[int, ...] | None` to mirror
https://github.com/python-pillow/Pillow/blob/5122c8356d6fa8f4a0d8449e8875734bb5d4d966/src/PIL/Image.py#L1631-L1633

Regarding the first code you found

https://github.com/python-pillow/Pillow/blob/5122c8356d6fa8f4a0d8449e8875734bb5d4d966/src/PIL/_imaging.pyi#L4

I think this should actually be updated as well.

> `def __iter__(self) -> Iterator[float]: ...`

I don't know where this came from.

tpoliaw:
```python
>>> ic = im.getdata()
>>> ic.__iter__()
AttributeError: 'ImagingCore' object has no attribute '__iter__'. Did you mean: '__str__'?
```

I think the problem is in pyright/pyshed more than pillow. The `ImagingCore` type has no `__iter__` method so while the PR fixes the linting, it's not really accurate. Adding `__len__` to the type-stub would be more correct but still wouldn't fix the problem here because the typing on the `list` method is too restrictive which (I think?) comes from typeshed.

A workaround that passes type-checking could be `list(iter(im.getdata()))`

tpoliaw:
For completeness, I found [this issue](https://redirect.github.com/python/typeshed/issues/7813) in typeshed and the upshot was that supporting `__getitem__` iterators wasn't feasible and calling code should use `iter`.

DominicOram:
Ok, in which case I will change the comment in https://github.com/python-pillow/Pillow/blob/5122c8356d6fa8f4a0d8449e8875734bb5d4d966/src/PIL/Image.py#L1436 to suggest `list(iter(im.getdata()))` as it better matches the type

src/PIL/Image.py
```diff
@@ -1159,7 +1159,8 @@ def convert_transparency(
             if new_im.mode == "P" and new_im.palette:
                 try:
                     new_im.info["transparency"] = new_im.palette.getcolor(
-                        cast(tuple[int, ...], trns), new_im  # trns was converted to RGB
+                        cast(tuple[int, ...], trns),
+                        new_im,  # trns was converted to RGB
```
radarhere, line 1163: What was your thinking behind this reformatting?

src/PIL/Image.py
```diff
@@ -3224,7 +3225,8 @@ class SupportsArrowArrayInterface(Protocol):
     """
 
     def __arrow_c_array__(
-        self, requested_schema: "PyCapsule" = None  # type: ignore[name-defined]  # noqa: F821, UP037
+        self,
+        requested_schema: "PyCapsule" = None,  # type: ignore[name-defined]  # noqa: F821, UP037
```
radarhere, line 3229: What was your thinking behind this reformatting?

radarhere:
When I test, I don't find any issue with `list(im.getdata())` in mypy.

I'm a bit reluctant to tell everyone to change their code in this way - not everyone uses type checkers, and of that subset, [only a third use pyright.](https://engineering.fb.com/2024/12/09/developer-tools/typed-python-2024-survey-meta/)

A simpler alternative for pyright might actually be iterating over `im.tobytes()`.

Others may disagree with me, but if you really think there is a problem here, I'd actually be more interested in deprecating `getdata()` altogether.
https://github.com/python-pillow/Pillow/blob/76f04b46c563155a66a7fe92f500e7e8e9d6e096/src/PIL/Image.py#L1447

radarhere:
Any thoughts?

aclark4life:
> I'd actually be more interested in deprecating `getdata()` altogether.

👍

radarhere:
I've created https://github.com/python-pillow/Pillow/pull/9292 to suggest deprecating `getdata()`, and instead adding a new method that always returns `list(im.getdata())`.

DominicOram:
Sorry, lost track of this. If the preferred solution is to deprecate `getdata()` that's fine. I will close this PR

PR #9266, opened 2025-10-18
Do not resize macOS retina screenshots by default

### LINKED ISSUE #9253: Add scale option to ImageGrab.grab function
https://github.com/python-pillow/Pillow/blob/33eb16bb39df3ec9b3a728d46c43790b2af817c7/src/PIL/ImageGrab.py#L54 produces a low quality image if `bbox` is given. The reason is that  the screenshot for a region 100x200 is a 200x400 image on a Macbook Pro  which will scale the the region by 2. After the resizing, the image quality is reduced.

If no `bbox` is provide, `grab` function produces a 3024x1964 image for the whole screen instead of a 1512x982 image. The resulted image does not have its quality reduced.

So the API has an inconsistent behavior. If `bbox` is not given, the resulted screenshot image has a good quality. But if `bbox` is given, the resulted screenshot image has a low quality. 

So I suggest to add option to control the scaling like what mss package has done.  mss has `IMAGE_OPTIONS ` to control the scaling.See https://github.com/BoboTiG/python-mss/blob/main/CHANGELOG.md#1010-2025-08-16 for details.

* OS: macOS 14.3.1 
* Python: 3.10.18
* Pillow: 11.3.0

`python3 -m PIL.report` produces:
```--------------------------------------------------------------------
Pillow 11.3.0
Python 3.10.18 (main, Jun  5 2025, 08:37:47) [Clang 14.0.6 ]
Python executable is /Users/jing/miniconda3/envs/m-agent/bin/python
System Python files loaded from /Users/jing/miniconda3/envs/m-agent
Python Pillow modules loaded from /Users/jing/miniconda3/envs/m-agent/lib/python3.10/site-packages/PIL
Binary Pillow modules loaded from /Users/jing/miniconda3/envs/m-agent/lib/python3.10/site-packages/PIL
--- PIL CORE support ok, compiled for 11.3.0
--- AVIF support ok, loaded 1.3.0
--- JPEG support ok, compiled for libjpeg-turbo 3.1.1
--- ZLIB (PNG/ZIP) support ok, loaded 1.3.1.zlib-ng, compiled for zlib-ng 2.2.4
```

radarhere:
Resolves https://github.com/python-pillow/Pillow/issues/9253

On macOS retina screens, `ImageGrab.grab()` will produce 2x images when not using `bbox`, but 1x images when using `bbox`. The issue describes this as inconsistent.

Scaling bounding box images down was added in https://github.com/python-pillow/Pillow/pull/6152 with Pillow 9.1.0, so that is relatively new. This PR stops that resizing by default. If the user wishes to resize, they must provide a keyword-only argument,``ImageGrab.grab(scale_down=True)``


hugovk:
Just a thought: rather than tacking on a bool `scale_down=False`, perhaps `scale=2`? Who knows if Apple will produce a 4x or something in the future...

radarhere:
In main, ``bbox`` screenshots are captured at 1x.
In the current form of this PR, images are captured at 2x by default, and then 1x if ``scale_down=True``.

If we change that to ``scale=2`` by default, and then Apple releases 4x, users could start specifying ``scale=4``, but our default would still be 2 for these super retina screens. So the default for 4x screens would no longer be 'full resolution' (4x) or 'bbox size' (1x), but something in the middle (2x). I don't imagine anyone would want something in the middle?

The only way I think ``scale`` is helpful is if the default is ``scale=1``. But that's the very problem the issue creator is trying to avoid, where ``ImageGrab.grab(bbox)`` is giving 1x images currently, while ``ImageGrab.grab()`` gives 2x.

Instead, I've pushed a commit to improve the screen scaling detection, so that if Apple releases 4x in the future, the default will start capturing 4x. ``scale_down=True`` will still provide 1x.

hugovk:
The thing I wasn't so keen on was the Boolean `scale_down` which is less extensible. But maybe it's not so important.

If we're keeping the `scale_down` API, let's go for the first version, and we can better internal screen size detection as and when that's necessary.

radarhere:
Sorry, not sure what you mean by 'first version'. Do you mean just remove my last commit?

hugovk:
Yes, I meant remove the last commit, and just go for https://github.com/python-pillow/Pillow/pull/9266/changes/0a272ccc7f5a0fc6a60336f4e9c26e6756846a0f.

PR #9269, opened 2025-10-20
Added type hints

radarhere:


src/PIL/ImageText.py
```diff
@@ -130,8 +130,7 @@ def get_length(self):
 
         :return: Either width for horizontal text, or height for vertical text.
         """
-        split_character = "\n" if isinstance(self.text, str) else b"\n"
-        if split_character in self.text:
+        if "\n" in self.text if isinstance(self.text, str) else b"\n" in self.text:
```
hugovk: Can we rewrite or revert this? It's not so clear with two `if`s and one `else`.

radarhere: The error is
```
src/PIL/ImageText.py:134: error: Unsupported operand types for in
("str | bytes" and "str | bytes")  [operator]
            if split_character in self.text:
               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Is
```python
if isinstance(self.text, str):
	multiline = "\n" in self.text
else:
	multiline = b"\n" in self.text
```
any better?

PR #9256, opened 2025-10-15
Use enums for Modes and RawModes in C

radarhere:
https://github.com/python-pillow/Pillow/pull/9100, plus https://github.com/eyedav/Pillow/pull/3, and with the merge conflicts fixed.

PR #9274, opened 2025-10-21
Additional fields in ModeData structure

wiredfool:

 * Additional hoisting of data from Storage.c into the const ModeData structure. 
 * Perf: Don't strcpy the band names and arrow format in for every new image. 

Still considering -- 
* Should we leave the fields in the Image header?  We can copy ints, we can copy addresses. 
* Should we just have a pointer to the mode struct?  This makes more changes across the code base.


radarhere:
> Should we just have a pointer to the mode struct? This makes more changes across the code base.

Could you clarify for me, what was the motivation behind the mode struct in the first place (#9256)? I thought it was primarily preparing the way for other field to be moved in there as well. Or was the theory that it would run `strcmp` less?

radarhere:
Closing, as this has been in draft for a while with no activity.

PR #9279, opened 2025-10-28
Provide a clear option to depend on zlib-ng

hmaarrfk:
This is more of a nit than anything, but the problem arose in conda-forge where we still haven't gone all-in on zlib-ng.

As such, we ship BOTH zlib, and zlib-ng, which are designed to co-exist.

I thus wrote a patch for pillow to be able to directly target zlib-ng.

I figured I would upstream it and maybe this can help other distributions with this predicament.

xref: 
* https://github.com/conda-forge/pillow-feedstock/pull/173
* https://github.com/conda-forge/zlib-ng-feedstock/issues/10
* https://github.com/conda-forge/conda-forge.github.io/issues/2638

I'm happy to address any issues you all have with this but the idea is that:

if zlib-ng is found, then choose that, and just use the `zlib-ng` headers directly instead of `zlib.h`.

-------------------------------------------------------

TODO:
- [ ] Add an explicit option to link to prioritize `zlib.h` or `zlib-ng.h` -- @mgorny

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -725,7 +725,12 @@ class pil_build_ext(build_ext):
 
         if feature.want("zlib"):
             _dbg("Looking for zlib")
-            if _find_include_file(self, "zlib.h"):
+            if _find_include_file(self, "zlib-ng.h"):
+                if _find_library_file(self, "z-ng"):
+                    feature.set("zlib", "z-ng")
+                elif sys.platform == "win32" and _find_library_file(self, "zlib-ng"):
+                    feature.set("zlib", "zlib-ng")
+            elif _find_include_file(self, "zlib.h"):
                 if _find_library_file(self, "z"):
                     feature.set("zlib", "z")
                 elif sys.platform == "win32" and _find_library_file(self, "zlib"):
@@ -923,9 +928,11 @@ class pil_build_ext(build_ext):
             defs.append(("HAVE_OPENJPEG", None))
             if sys.platform == "win32" and not PLATFORM_MINGW:
                 defs.append(("OPJ_STATIC", None))
-        if feature.get("zlib"):
-            libs.append(feature.get("zlib"))
+        if zlib := feature.get("zlib"):
+            libs.append(zlib)
             defs.append(("HAVE_LIBZ", None))
+            if zlib in ["z-ng", "zlib-ng"]:
+                defs.append(("HAVE_ZLIBNG", None))
         if feature.get("imagequant"):
             libs.append(feature.get("imagequant"))
             defs.append(("HAVE_LIBIMAGEQUANT", None))
--- a/src/_imaging.c
+++ b/src/_imaging.c
@@ -85,8 +85,10 @@
 #endif
 #endif
 
-#ifdef HAVE_LIBZ
-#include "zlib.h"
+#ifdef HAVE_ZLIBNG
+#include <zlib-ng.h>
+#else
+#include <zlib.h>
 #endif
 
 #ifdef HAVE_LIBTIFF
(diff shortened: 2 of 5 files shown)


aclark4life:
> I figured I would upstream it and maybe this can help other distributions with this predicament.

Are we the upstream and what is the predicament? 😄

hmaarrfk:
> Are we the upstream

I think!? right?

> the predicament

Today: i want to take advantage of zlib-ng with Pillow, but:

1. Either I migrate the whole OS/distribution to zlib-ng
2. I stick to zlib

I want to have both:
1. Keep all the old packages on zlib
2. Use zlib-ng with pillow.

hmaarrfk:
now if only i understood the windows failures ^_^

radarhere:
> Today: i want to take advantage of zlib-ng with Pillow, but:
> 
> 1. Either I migrate the whole OS/distribution to zlib-ng
> 2. I stick to zlib

So, just spelling this out, you're saying that if both zlib and zlib-ng are installed, then Pillow may pick zlib over zlib-ng? Probably just based on the order of the paths that it searches?

Have you considered adjusting the search paths? https://pillow.readthedocs.io/en/stable/installation/building-from-source.html#installing
> If they are installed in a non-standard location, you may need to configure setuptools to use those locations by editing setup.py or pyproject.toml, or by adding environment variables on the command line:
> 
> `CFLAGS="-I/usr/pkg/include" python3 -m pip install --upgrade Pillow --no-binary :all:`

hmaarrfk:
> So, just spelling this out

If you build out zlib and zlib-ng so that they can be co-installed, you have to disable zlib-ng's "compatibility" mode.

Once you've disabled compatibility mode, we are "forced" to use `<zlib.h>` from `zlib`.

> Have you considered adjusting the search paths?

Yeah this is generally what we do. we set enough flags that it uses the isolated environment.

setup.py
```diff
@@ -725,7 +725,12 @@ def build_extensions(self) -> None:
 
         if feature.want("zlib"):
             _dbg("Looking for zlib")
-            if _find_include_file(self, "zlib.h"):
+            if _find_include_file(self, "zlib-ng.h"):
```
mgorny, line 728: This introduces an automagic dependency on `zlib-ng` when both `zlib-ng` and `zlib` are installed. As such, Gentoo (as well as anyone who doesn't build in an entirely isolated environment) would end up having to patch this out in order to use plain `zlib`. I'm not opposed to having `zlib-ng` as an option, but then it should be *an option*, one that can be explicitly disabled.

hmaarrfk: thanks for bringing this up. it is true. it is "priority for zlib-ng".

I'm happy to add an explicit "switch" between the two.

I'll add that to the todo list if the general strategy is agreeable.

aclark4life: Still working on the general strategy being agreeable. That is usually determined by interest and the number of folks looking for a fix for this issue. At present, we depend on zlib and zlib-ng satisfies that dependency via compat mode. How many folks are looking for co-existence ?

hmaarrfk: That’s a good question. 

I’m. Happy to let this PR sit

eshattow: zlib-ng-dev is installed but Pillow fails to build as it wants zlib? from Home Assistant home-assistant/core

It would be preferable to add a default favoring zlib-ng (as in Pillow 11.1.0)... yet for whatever reason Pillow does not find this in Alpine when there is zlib-ng-dev apk installed. Maybe user error on my part, I will dig in to this troubleshooting more.

hmaarrfk:
Im out of personal time to work on this

PR #9303, opened 2025-11-17
Added Python domain to Sphinx declarations

radarhere:


hugovk:
What difference does this make? The output looks identical:

* https://pillow.readthedocs.io/en/latest/handbook/concepts.html#filters-comparison-table
* https://pillow--9303.org.readthedocs.build/en/9303/handbook/concepts.html#filters-comparison-table

radarhere:
I don't expect it to make a difference - https://www.sphinx-doc.org/en/master/usage/domains/index.html
> If no other default is selected, the Python domain (named `py`) is the default one

I'm just thinking about consistency with the rest of our documentation.
```
$ grep -r ":py:meth" docs | wc
     254    1786   27947
$ grep -r ":py:func" docs | wc
      47     348    5056
$ grep -r ":py:data" docs | wc
      38     240    3860
```

But if you would rather not have the churn, then ok.

hugovk:
Yeah, let's skip it.

PR #9312, opened 2025-11-28
Update zlib-ng to 2.3.1, except on manylinux2014 aarch64

radarhere:
zlib-ng 2.3.1 has been released - https://github.com/zlib-ng/zlib-ng/releases/tag/2.3.1

When I upgraded it on manylinux2014 aarch64, I found [an error](https://github.com/radarhere/Pillow/actions/runs/19763663372/job/56631262763#step:5:8864). So I have excluded that from this. It will become irrelevant in a few months when manylinux2014 support is dropped with [Amazon 2 EOL.](https://endoflife.date/amazon-linux)

I have also reverted part of https://github.com/python-pillow/Pillow/pull/8673, now that https://github.com/zlib-ng/zlib-ng/pull/1867 is included.


hugovk:
Thank you!

PR #9327, opened 2025-12-05
Do not use cmd shell in GitHub Actions

radarhere:
https://github.com/python-pillow/Pillow/pull/9318 upgraded zizmor to include https://docs.zizmor.sh/audits/#obfuscation
> The CMD shell has no formal grammar, making it impossible to accurately analyze for security issues.

This PR allows for that by
1. Changing test-windows.yml to avoid the cmd shell.
2. Moving the cmd shell instructions from wheels.yml into a separate file, where they will not bother zizmor. It is not an ideal solution, but I suspect an ideal solution would involve changing winbuild to no longer generate cmd files, and that seems like a step too far. This will at least allow zizmor's rule to be in place, so that we can enforce it by default.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -186,24 +186,7 @@ jobs:
 
       - name: Build wheels
         run: |
-          setlocal EnableDelayedExpansion
-          for %%f in (winbuild\build\license\*) do (
-            set x=%%~nf
-            rem Skip FriBiDi license, it is not included in the wheel.
-            set fribidi=!x:~0,7!
-            if NOT !fribidi!==fribidi (
-              rem Skip imagequant license, it is not included in the wheel.
-              set libimagequant=!x:~0,13!
-              if NOT !libimagequant!==libimagequant (
-                echo. >> LICENSE
-                echo ===== %%~nf ===== >> LICENSE
-                echo. >> LICENSE
-                type %%f >> LICENSE
-              )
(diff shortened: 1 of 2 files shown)


.github/workflows/wheels.yml
```diff
@@ -186,24 +186,7 @@ jobs:
...
-              if NOT !libimagequant!==libimagequant (
-                echo. >> LICENSE
-                echo ===== %%~nf ===== >> LICENSE
-                echo. >> LICENSE
-                type %%f >> LICENSE
-              )
-            )
-          )
-          call winbuild\\build\\build_env.cmd
-          %pythonLocation%\python.exe -m cibuildwheel . --output-dir wheelhouse
```
hugovk, line 206: We can keep this chunk, it doesn't declare `shell: cmd` so will default to `pwsh`.

https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#defaultsrunshell

And zizmor doesn't complain about it.

radarhere: I don't think this is the case. The `shell: cmd` on line 220 is related to this block.

I've separated this change out from the other changes, and you can see that without it, zizmor does complain - https://results.pre-commit.ci/run/github/5171600/1765401638.xj0JlAmBT_6lRfMDGmxoqQ

radarhere:
Superseded by #9359

PR #9328, opened 2025-12-05
Added compare_images function to ImageChops

### LINKED ISSUE #6478: Add Function to Compare Images
I think it would be useful to have a function to check if two images are the same. I don't mean with a perceptual hash like #3120, or with the existing `Image.__eq__` method, but something to check if two images are visually similar. By that I mean the images are the same size and have the same color pixels, after taking into account the EXIF Orientation tag and the images possibly having different modes. Something like
```python
def same_image(img_a,img_b):
	img_a = ImageOps.exif_transpose(img_a)
	img_b = ImageOps.exif_transpose(img_b)
	if img_a.size != img_b.size:
		return False
	img_a = img_a.convert('RGBA')
	img_b = img_b.convert('RGBA')
	diff = ImageChops.difference(img_a,img_b)
	return diff.getbbox() is None and diff.convert('RGB').getbbox() is None
```
but more efficient. And also it needs to handle images with multiple frames. Or maybe just throw an exception for those for now.

johnnygitgud:
This pull request implements the feature requested in issue #6478 by adding a new helper function compare_images to ImageChops.

The function compares two images pixel-by-pixel using ImageChops.difference and returns a dictionary containing:

different_pixels — the number of pixels that differ

percent_difference — the percentage of differing pixels

It raises a ValueError if the images have different modes or sizes, matching existing behavior elsewhere in Pillow.

A new test file (Tests/test_compare_images.py) is included to validate that:

identical images return zero difference

different images return nonzero values

This feature builds directly on existing ImageChops functionality and follows the project's testing and docstring conventions.

Closes #6478

### DIFF
--- a/Tests/test_compare_images.py
+++ b/Tests/test_compare_images.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 from PIL import Image, ImageChops
 
 
-def test_compare_identical_images():
+def test_compare_identical_images() -> None:
     im1 = Image.new("RGB", (10, 10), "red")
     im2 = Image.new("RGB", (10, 10), "red")
 
@@ -13,7 +13,7 @@ def test_compare_identical_images():
     assert result["percent_difference"] == 0.0
 
 
-def test_compare_different_images():
+def test_compare_different_images() -> None:
     im1 = Image.new("RGB", (10, 10), "red")
     im2 = Image.new("RGB", (10, 10), "blue")
 
--- a/src/PIL/ImageChops.py
+++ b/src/PIL/ImageChops.py
@@ -293,7 +293,7 @@ def composite(
     return Image.composite(image1, image2, mask)
 
 
-def compare_images(image1, image2):
+def compare_images(image1: Image.Image, image2: Image.Image) -> dict[str, float]:
     """
     Compare two images pixel by pixel.
 
@@ -324,7 +324,7 @@ def compare_images(image1, image2):
         }
 
     # Count non-zero (different) pixels
-    nonzero = sum(1 for px in diff.getdata() if px != 0)
+    nonzero = sum(1 for px in tuple(diff.getdata()) if px != 0)
     total = image1.size[0] * image1.size[1]
 
     return {


src/PIL/ImageChops.py
```diff
@@ -293,6 +293,44 @@ def composite(
...
+
+        - ``different_pixels`` – the number of pixels that differ
+        - ``percent_difference`` – the percentage of differing pixels
+
+    :raises ValueError: If the images have different sizes or modes.
+    """
+    if image1.mode != image2.mode:
+        raise ValueError("Images must have the same mode")
+    if image1.size != image2.size:
+        raise ValueError("Images must have the same size")
```
radarhere:
```suggestion
        msg = "Images must have the same mode"
        raise ValueError(msg)
    if image1.size != image2.size:
        msg = "Images must have the same size"
        raise ValueError(msg)
```
This should fix the Lint error - https://github.com/python-pillow/Pillow/actions/runs/19978770041/job/57301143480?pr=9328

Tests/test_compare_images.py
```diff
@@ -0,0 +1,23 @@
+from __future__ import annotations
+
+from PIL import Image, ImageChops
+
+
+def test_compare_identical_images():
```
radarhere:
```suggestion
def test_compare_identical_images() -> None:
```
Type hint.

Tests/test_compare_images.py
```diff
@@ -0,0 +1,23 @@
...
+    im1 = Image.new("RGB", (10, 10), "red")
+    im2 = Image.new("RGB", (10, 10), "red")
+
+    result = ImageChops.compare_images(im1, im2)
+
+    assert result["different_pixels"] == 0
+    assert result["percent_difference"] == 0.0
+
+
+def test_compare_different_images():
```
radarhere:
```suggestion
def test_compare_different_images() -> None:
```

src/PIL/ImageChops.py
```diff
@@ -293,6 +293,44 @@ def composite(
...
+
+    # If identical, diff is entirely zero
+    if diff.getbbox() is None:
+        return {
+            "different_pixels": 0,
+            "percent_difference": 0.0,
+        }
+
+    # Count non-zero (different) pixels
+    nonzero = sum(1 for px in diff.getdata() if px != 0)
```
radarhere:
```suggestion
    nonzero = sum(1 for px in tuple(diff.getdata()) if px != 0)
```
Another type hint.

src/PIL/ImageChops.py
```diff
@@ -293,6 +293,44 @@ def composite(
     return Image.composite(image1, image2, mask)
 
 
+def compare_images(image1, image2):
```
radarhere:
```suggestion
def compare_images(image1: Image.Image, image2: Image.Image) -> dict[str, float]:
```

radarhere:
Hi. I don't think this does actually resolve #6478, as I think that issue wants to compare all frames of an image, and to have images automatically convert before comparison. That issue also only expected a boolean to be returned, rather than any information about how different the two images are.

So let me ask - ignoring #6478, do you personally think what you're put together would be a helpful addition to Pillow?

src/PIL/ImageChops.py
radarhere:
```suggestion
        msg = "Images must have the same mode"
        raise ValueError(msg)
    if image1.size != image2.size:
        msg = "Images must have the same size"
        raise ValueError(msg)
```

Tests/test_compare_images.py
radarhere:
```suggestion
def test_compare_identical_images() -> None:
```

Tests/test_compare_images.py
radarhere:
```suggestion
def test_compare_different_images() -> None:
```

radarhere:
Closing, as this doesn't resolve the issue. The suggestion here isn't wholly uninteresting by itself, but there's been no response, and I think there should be at least one advocate for any new functionality.

PR #9338, opened 2025-12-12
Fix dithering and improve readability

WildRackoon:
Somes changes related to the dithering conversion code:

- Fix RGB cache error values at the end of a row, not set to the right values (Blue values was used for R,G and B ?)
I believe this is a mistake, this affects only the last column of the image, so it is likely very minimal, but I still really would like a second opinion on that.

- Improved error diffusion readability, now the Floyd–Steinberg weights are readable through the code.

- We should add more comments for posterity IMO, most of this dates back to the original PIL fork, the implementation is nicely optimized, but that makes the code slightly more confusing.

I mean look at this 😅: https://github.com/python-pillow/Pillow/blob/8494b06c7165bbb8d311769037cbf78c6aa73e33/src/libImaging/Convert.c#L1284

Future improvement includes adding more dithering kernels in the same optimized manner.

If anyone needs more details on those changes, I published a gist for that here: [https://gist.github.com/WildRackoon/0030fc3d4db3f4af73e2cdac19bbdf6d](https://gist.github.com/WildRackoon/0030fc3d4db3f4af73e2cdac19bbdf6d)


radarhere:
> but I still have to make some more tests with regards to that.

Are you saying this PR is not yet ready for review?

WildRackoon:
> > but I still have to make some more tests with regards to that.
> 
> Are you saying this PR is not yet ready for review?

I edited my comment, It builds fine and works as intented: the result is the same as before with a bunch of test images I tried.

But it is not like we can test that with a ground truth and improve the test suite to that end, so someone's second opinion would be nice.

radarhere:
You can see that our test suite is failing with your changes. I've created https://github.com/WildRackoon/Pillow/pull/1 to update our expected results.

WildRackoon:
Many thanks @radarhere, haven't had time to figure out the related tests yet, just cherry-picked your commit now.

PR #9341, opened 2025-12-13
Neon and sepia filters

matheusmpff:
## Description:
This PR introduces two new image processing functions to the `PIL.ImageOps`  module:

`sepia` 

Applies a classic sepia tone effect to RGB images.

Converts the image to RGB if necessary.

Computes new pixel values using the standard sepia formula and clamps results to [0, 255].

Preserves the original image size and mode.

Practical use: adds a warm, vintage look to images with minimal processing overhead.

`neon_effect` 

Applies a neon/glow effect to an image.

Internally uses a `Sobel edge-detection filter` , Gaussian blur, and colorization.

Combines the neon layer with the original image using alpha blending.

Supports RGB images and customizable neon color.

Practical use: highlights edges and details in a visually striking, glowing style.

 ## Benefits:

Expands Pillow’s “ready-made” filters with two highly requested effects.

Provides developers with visually appealing filters without third-party dependencies.

## Results
Image after sepia effect:
sepia

Image after neon-glow effect:
neon

## Notes

There are tests for all functions created in this PR covering the main basic behaviors


aclark4life:
Very cool, thank you!

radarhere:
I've created https://github.com/matheusmpff/Pillow/pull/1 with some suggestions.

src/PIL/ImageOps.py
```diff
@@ -623,6 +623,169 @@ def grayscale(image: Image.Image) -> Image.Image:
...
+    horizontal (Gx) and vertical (Gy) Sobel kernels.
+
+    :param image: the image to apply the filter
+    :return: An image.
+    """
+    if image.mode != "L":
+        image = image.convert("L")
+
+    Kx = [[-1, 0, 1], [-2, 0, 2], [-1, 0, 1]]
+    Ky = [[1, 2, 1], [0, 0, 0], [-1, -2, -1]]
```
radarhere, line 668: Just want to confirm - you're sure the `Ky` values are correct? Looking at https://en.wikipedia.org/wiki/Sobel_operator#Formulation, one might expect that `Ky` should actually be `[[-1, -2, -1], [0, 0, 0], [1, 2, 1]]`

matheusmpff: I had used an inverted kernel for the Sobel operator in the y direction. The right one is the `[[-1, -2, -1], [0, 0, 0], [1, 2, 1]]`. But I think in the implementation we use the absolute values of `gy` so I think the results will be the same. It should be great to change for the correct one

src/PIL/ImageOps.py
```diff
@@ -623,6 +623,100 @@ def grayscale(image: Image.Image) -> Image.Image:
...
+) -> Image.Image:
+    """
+    Apply a neon glow effect to an image using edge detection,
+    blur-based glow generation, colorization, and alpha blending.
+    It calls all auxiliary functions required to generate
+    the final result.
+
+    :param image: Image to create the effect
+    :param color: RGB color used for neon effect
+    :alpha: controls the intensity of the neon effect
```
radarhere:
```suggestion
    :param alpha: Controls the intensity of the neon effect. If alpha is 0.0, a copy of
       the image is returned unaltered.
```

src/PIL/ImageOps.py
```diff
@@ -623,6 +623,100 @@ def grayscale(image: Image.Image) -> Image.Image:
...
+
+
+def sobel(image: Image.Image) -> Image.Image:
+    """
+    Applies a Sobel edge-detection filter to the given image.
+
+    This function computes the Sobel gradient magnitude using the
+    horizontal (Gx) and vertical (Gy) Sobel kernels.
+
+    :param image: the image to apply the filter
```
radarhere:
```suggestion
    :param image: The image to be filtered
```

src/PIL/ImageOps.py
```diff
@@ -623,6 +623,100 @@ def grayscale(image: Image.Image) -> Image.Image:
...
+    if image.mode != "L":
+        image = image.convert("L")
+
+    Kx = [[-1, 0, 1], [-2, 0, 2], [-1, 0, 1]]
+    Ky = [[1, 2, 1], [0, 0, 0], [-1, -2, -1]]
+
+    out = Image.new("L", image.size)
+
+    for y in range(1, image.height - 1):
+        for x in range(1, image.width - 1):
```
radarhere:
```suggestion
    for x in range(1, image.width - 1):
        for y in range(1, image.height - 1):
```
Nitpick: I think it would be nice if the order of the loops matched the order in `sepia()`

Tests/test_imageops.py
```diff
@@ -604,3 +604,36 @@ def test_autocontrast_preserve_one_color(color: tuple[int, int, int]) -> None:
         img, cutoff=10, preserve_tone=True
     )  # single color 10 cutoff
     assert_image_equal(img, out)
+
+
+def test_sepia_preserves_size_and_mode() -> None:
+    img = Image.new("RGB", (10, 10), (100, 150, 200))
```
radarhere:
```suggestion
@pytest.mark.parametrize("mode", ("L", "RGB"))
def test_sepia_size_and_mode(mode: str) -> None:
    img = Image.new(mode, (10, 10))
```
Tests `sepia()` with a non-RGB image.

src/PIL/ImageOps.py
```diff
@@ -623,6 +623,101 @@ def grayscale(image: Image.Image) -> Image.Image:
...
+                    gy += v * Ky[dy + 1][dx + 1]
+
+            # Approximate gradient magnitude and clamp to [0, 255]
+            mag = int(min(255, abs(gx) + abs(gy)))
+            out.putpixel((x, y), mag)
+
+    return out
+
+
+def neon_effect(
```
hugovk, line 692: Name this just `neon`? The other two could also be `_effect` but it doesn't add much.

hugovk:
The PR title and description are about neon and sepia functions, but this also introduces a public `sobel` function, but it's described as internal:

> Internally uses a `Sobel edge-detection filter` , Gaussian blur, and colorization.


Is it intentional to be part of the public API or should it be `_sobel` to keep it internal?

src/PIL/ImageOps.py
```diff
@@ -623,6 +623,101 @@ def grayscale(image: Image.Image) -> Image.Image:
...
+        for y in range(image.height):
+            value = image.getpixel((x, y))
+            assert isinstance(value, tuple)
+            r, g, b = value
+
+            tr = 0.393 * r + 0.769 * g + 0.189 * b
+            tg = 0.349 * r + 0.686 * g + 0.168 * b
+            tb = 0.272 * r + 0.534 * g + 0.131 * b
+
+            out.putpixel((x, y), tuple(min(255, int(c)) for c in (tr, tg, tb)))
```
akx, line 649: Isn't this expressible as an `Image.convert(matrix=...)`?

src/PIL/ImageOps.py
```diff
@@ -623,6 +623,101 @@ def grayscale(image: Image.Image) -> Image.Image:
...
+                for dx in (-1, 0, 1):
+                    v = image.getpixel((x + dx, y + dy))
+                    assert isinstance(v, (int, float))
+
+                    gx += v * Kx[dy + 1][dx + 1]
+                    gy += v * Ky[dy + 1][dx + 1]
+
+            # Approximate gradient magnitude and clamp to [0, 255]
+            mag = int(min(255, abs(gx) + abs(gy)))
+            out.putpixel((x, y), mag)
```
akx, line 687: Could this be expressed as `ImageFilter` 3x3 convolution operations..?

PR #9351, opened 2025-12-24
Add type hint for LOAD_TRUNCATED_IMAGES

bencevans:

 * Adds a type hint for ImageFile.LOAD_TRUNCATED_IMAGES.
 
     Sets the type to `bool`, otherwise `Literal[False]` is assumed causing a lint error when setting `ImageFile.LOAD_TRUNCATED_IMAGES = True`.

### DIFF
--- a/src/PIL/ImageFile.py
+++ b/src/PIL/ImageFile.py
@@ -61,7 +61,7 @@ User code may set this to another number.
 
 SAFEBLOCK = 1024 * 1024
 
-LOAD_TRUNCATED_IMAGES = False
+LOAD_TRUNCATED_IMAGES: bool = False
 """Whether or not to load truncated image files. User code may change this."""
 
 ERRORS = {


radarhere:
Hi. What environment were you in when you saw the error? Believe it or not, here's what I already see with Pillow 12.0.0 in PyCharm.

pycharm

bencevans:
Hi! Ah this will teach me for trying out new toys haha. I'm using [ty](https://docs.astral.sh/ty/) in VS Code. It's interesting it's picked up as a Literal, it seems stricter than other type checkers.

Forgot to include in the original message but also using `pillow===12.0.0`.

radarhere:
Thanks. I can verify that your fix does satisfy ty.

While the change you've suggested here is minor, I expect there will be other ty problems lurking. I'm somewhat hesitant about trying to satisfy multiple type checkers at once, particularly when ty is [still in beta](https://astral.sh/blog/ty).

hugovk:
Mypy is pretty slow, it would be good to move to a new fast one such as ty or pyrefly at some point. But I've not really looked into them, and with default settings report over 300 and 600 errors each...

radarhere:
If there isn't a reason that `ImageFile.LOAD_TRUNCATED_IMAGES` is uniquely relevant,
and if we're not going to try and satisfy multiple type checkers at once,
then I think this can be closed.

PR #9359, opened 2025-12-29
Replace `shell: cmd` with `shell: bash`

hugovk:
Replaces and closes https://github.com/python-pillow/Pillow/pull/9327.

Includes part of https://github.com/python-pillow/Pillow/pull/9327, and also converts that last `cmd` to `bash`.

(Fine to leave this until after the January release.)

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -186,24 +186,18 @@ jobs:
 
       - name: Build wheels
         run: |
-          setlocal EnableDelayedExpansion
-          for %%f in (winbuild\build\license\*) do (
-            set x=%%~nf
-            rem Skip FriBiDi license, it is not included in the wheel.
-            set fribidi=!x:~0,7!
-            if NOT !fribidi!==fribidi (
-              rem Skip imagequant license, it is not included in the wheel.
-              set libimagequant=!x:~0,13!
-              if NOT !libimagequant!==libimagequant (
-                echo. >> LICENSE
-                echo ===== %%~nf ===== >> LICENSE
-                echo. >> LICENSE
-                type %%f >> LICENSE
-              )
-            )
-          )
-          call winbuild\\build\\build_env.cmd
-          %pythonLocation%\python.exe -m cibuildwheel . --output-dir wheelhouse
+          for f in winbuild/build/license/*; do
+            name=$(basename "${f%.*}")
+            # Skip FriBiDi license, it is not included in the wheel.
+            [[ $name == fribidi* ]] && continue
+            # Skip imagequant license, it is not included in the wheel.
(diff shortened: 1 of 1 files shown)


.github/workflows/wheels.yml
```diff
@@ -186,24 +186,18 @@ jobs:
...
+          for f in winbuild/build/license/*; do
+            name=$(basename "${f%.*}")
+            # Skip FriBiDi license, it is not included in the wheel.
+            [[ $name == fribidi* ]] && continue
+            # Skip imagequant license, it is not included in the wheel.
+            [[ $name == libimagequant* ]] && continue
+            echo "" >> LICENSE
+            echo "===== $name =====" >> LICENSE
+            echo "" >> LICENSE
+            cat "$f" >> LICENSE
```
radarhere, line 198:
```suggestion
            cat $f >> LICENSE
```

.github/workflows/wheels.yml
```diff
@@ -186,24 +186,18 @@ jobs:
...
-                echo ===== %%~nf ===== >> LICENSE
-                echo. >> LICENSE
-                type %%f >> LICENSE
-              )
-            )
-          )
-          call winbuild\\build\\build_env.cmd
-          %pythonLocation%\python.exe -m cibuildwheel . --output-dir wheelhouse
+          for f in winbuild/build/license/*; do
+            name=$(basename "${f%.*}")
```
radarhere, line 190:
```suggestion
            name=$(basename ${f%.*})
```

.github/workflows/wheels.yml
```diff
@@ -186,24 +186,18 @@ jobs:
...
+            # Skip FriBiDi license, it is not included in the wheel.
+            [[ $name == fribidi* ]] && continue
+            # Skip imagequant license, it is not included in the wheel.
+            [[ $name == libimagequant* ]] && continue
+            echo "" >> LICENSE
```
radarhere, line 195:
```suggestion
            echo >> LICENSE
```
This also works, but I understand if you think it isn't as clear.

.github/workflows/wheels.yml
```diff
@@ -186,24 +186,18 @@ jobs:
...
+            echo "===== $name =====" >> LICENSE
+            echo "" >> LICENSE
```
radarhere, line 197:
```suggestion
            echo >> LICENSE
```

.github/workflows/wheels.yml
hugovk: shellcheck would warn about:

> SC2086 (info): Double quote to prevent globbing and word splitting.

https://www.shellcheck.net/wiki/SC2086

Issue #9361, opened 2025-12-30
Segfaults when using the 39C3 font from Chaos Communication congress


berlincount:
### What did you do?

I was trying to work with the TTF font from https://events.ccc.de/congress/2025/infos/styleguide.html (`39C3-Design-Package/39C3-Fonts/Kario 39C3 Variable/Desktop/Kario39C3Var-Roman.ttf` within https://events.ccc.de/congress/2025/infos/styleguide/39c3-styleguide-full-v2.zip), starting with listing the styles.

### What did you expect to happen?

I expected to get a list of the styles of the font, and to be able to work with it further.

### What actually happened?

`SIGSEGV`/`Segmentation fault`

### What are your OS, Python and Pillow versions?

* OS: python3-slim Docker image on Ubuntu 24.04.3 LTS on WSL2 as well as native
* Python: 3.14.2
* Pillow: 12.0.0

```text
--------------------------------------------------------------------
Pillow 12.0.0
Python 3.14.2 (main, Dec 30 2025, 00:42:55) [GCC 12.2.0]
--------------------------------------------------------------------
Python modules loaded from /opt/venv/python3/dist-packages/PIL
Binary modules loaded from /opt/venv/python3/dist-packages/PIL
--------------------------------------------------------------------
--- PIL CORE support ok, compiled for 12.0.0
--- TKINTER support not installed
--- FREETYPE2 support ok, loaded 2.14.1
--- LITTLECMS2 support ok, loaded 2.17
--- WEBP support ok, loaded 1.6.0
--- AVIF support ok, loaded 1.3.0
--- JPEG support ok, compiled for libjpeg-turbo 3.1.2
--- OPENJPEG (JPEG2000) support ok, loaded 2.5.4
--- ZLIB (PNG/ZIP) support ok, loaded 1.2.13, compiled for zlip-ng 2.2.5
--- LIBTIFF support ok, loaded 4.7.1
*** RAQM (Bidirectional Text) support not installed
--- LIBIMAGEQUANT (Quantization method) support not installed
--- XCB (X protocol) support ok
--------------------------------------------------------------------
```

```python
#!/usr/bin/env python3

from PIL import ImageFont
import pprint

font = ImageFont.truetype('Kario39C3Var-Roman.ttf')
pprint.pp(font.get_variation_names())
```

NOTE: This also might be a security issue, as this essentially seems to be Null Pointer Exception. I think this should be better contained / captured.

radarhere:
[...]

PR #9362, opened 2025-12-30
Allow for duplicate font variation styles

radarhere:
Resolves https://github.com/python-pillow/Pillow/issues/9361

Within `font_getvarnames()`, our code goes through the names, and once the ID of a name matches the ID of a style, it breaks. It presumes that each `master->namedstyle` would have a unique ID.

https://github.com/python-pillow/Pillow/blob/2ebb3e9964bcfb46e2b8dcaec1917caf67730a7d/src/_imagingft.c#L1280-L1291

The font found in the issue has two styles with the same ID. To allow for this, we should no longer break and continue to go through the loop. Once the list has been populated in C, Python can remove the duplicate result.

I created a test font here from our existing AdobeVFPrototype.ttf using ttx.

### DIFF
--- a/Tests/test_imagefont.py
+++ b/Tests/test_imagefont.py
@@ -702,7 +702,7 @@ def test_variation_get(font: ImageFont.FreeTypeFont) -> None:
         font.get_variation_axes()
 
     font = ImageFont.truetype("Tests/fonts/AdobeVFPrototype.ttf")
-    assert font.get_variation_names(), [
+    assert font.get_variation_names() == [
         b"ExtraLight",
         b"Light",
         b"Regular",


Tests/test_imagefont.py
```diff
@@ -742,6 +742,21 @@ def test_variation_get(font: ImageFont.FreeTypeFont) -> None:
     ]
 
 
+def test_variation_duplicates() -> None:
+    font = ImageFont.truetype("Tests/fonts/AdobeVFPrototypeDuplicates.ttf")
+    assert font.get_variation_names(), [
```
hugovk: Fix test:
```suggestion
    assert font.get_variation_names() == [
```

berlincount:
Hmmm, I've worked with the TTF file a bit now, and I think the variants are different - even though they have a same name - and it might be necessary to number them (#1, #2, ...) similar to how Gimp does it or so, otherwise selecting them becomes possible only  by using the index number, which isn't returned either. That font is file is properly horrible. :-/

radarhere:
The idea of adding variation name suffixes is a bit concerning from a theoretical standpoint - there's nothing stopping the font from already having both `XXtended Regular` and `XXtended Regular #1` as variation names, and then things become more complicated. You might say that's ridiculous, but it's kind of ridiculous to have identical variation names in the first place.

Would it help if I provided a modified version of the font file, that uses `XXtended Regular #1` and `XXtended Regular #2` instead? [Kario39C3Var-Roman-unique.ttf.zip](https://github.com/user-attachments/files/24402183/Kario39C3Var-Roman-unique.ttf.zip)

berlincount:
Thanks, a bit - but finding a solution working generically would be better, I think. **sigh** .. what a beepshow.

I've scanned the font with

```python
#!/usr/bin/env python3

from PIL import ImageFont
import pprint

ttf = ImageFont.truetype(font='Kario39C3Var-Roman.ttf', index=0)
#pprint.pp(ttf.font.getvarnames())

for i in range(4294967296+1):
  try:
    print("{} (face={}): {}".format('Kario39C3Var-Roman.ttf', i, ttf.font_variant(index=i).getname()), flush=True)
  except IOError as e:
    pass
```

(yeah yeah I know about the const being ridiculous, and I cancelled early)

this way, I could identify the right fonts and use the indicies in a hardcoded for what I need.

radarhere:
While I do still think that this font is flawed, I've created https://github.com/python-pillow/Pillow/pull/9376 to offer a solution. It would show the duplicates in `font.get_variation_names()`, and add a new method to let you choose between them - `font.set_variation_by_name_index(i)`.

```python
from PIL import ImageFont

font = ImageFont.truetype('Kario39C3Var-Roman.ttf')
for i, name in enumerate(font.get_variation_names()):
  print("Index", i, "Name", name)
  font.set_variation_by_name_index(i)
```

berlincount:
Yupp, that font is horribly flawed, but said fix would help me and potential other running problems with similar fonts.

Thanks for the super quick and helpful interactions, and a happy, healthy new year! 😍

hugovk:
I'm not sure if it's worth adding (and maintaining) extra APIs to support horribly flawed fonts.

radarhere:
If you're willing to accept a solution using our private APIs (with the disclaimer that we feel the freedom to change them in the future without warning), then with Pillow 12.1.0, you could use the following code.
```python
from PIL import ImageFont

font = ImageFont.truetype('Kario39C3Var-Roman.ttf')
names = [name.replace(b"\x00", b"") for name in font.font.getvarnames()]
for i, name in enumerate(names):
  print("Index", i, "Name", name)
  font.font.setvarname(i+1)
```

PR #9370, opened 2026-01-02
Remove Sphinx dependency from mypy

hugovk:
Follow on from https://github.com/python-pillow/Pillow/pull/9364.

`tox -e mypy` was still failing for me because it uses my machine's default Python version, currently 3.14.

We could pin tox to 3.10 as well for mypy, but that requires installing 3.10 locally, and we only require Sphinx for type checking `doc/dater.py`, and the `Sphinx` types don't really add much.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -23,7 +23,7 @@ jobs:
           persist-credentials: false
       - uses: actions/setup-python@v6
         with:
-          python-version: "3.10"
+          python-version: "3.x"
       - name: Install uv
         uses: astral-sh/setup-uv@v7
       - name: Lint
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -76,7 +76,7 @@ repos:
     rev: v0.24.1
     hooks:
       - id: validate-pyproject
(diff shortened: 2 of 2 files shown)


radarhere (approved):
An alternative solution would be to revert #9364 and then pin Sphinx to 9.0.4 for now. However, I suspect you would, very reasonably, prefer to use the latest version of Sphinx and not be held back by type hints.

I will also note that we could bring back the Sphinx dependency when Python 3.11 becomes EOL at the end of next year.

hugovk:
> An alternative solution would be to revert #9364 and then pin Sphinx to 9.0.4 for now. However, I suspect you would, very reasonably, prefer to use the latest version of Sphinx and not be held back by type hints.

Yeah, I think it's cleaner not to pin an old Sphinx for type hints, and use a newer version for building the docs. The hints don't really tell us much here, and `doc/` isn't part of our production code.

> I will also note that we could bring back the Sphinx dependency when Python 3.11 becomes EOL at the end of next year.

By then, Sphinx will likely be on 3.13+:

https://www.sphinx-doc.org/en/master/internals/release-process.html#python-version-support-policy

PR #9376, opened 2026-01-03
Add FreeTypeFont set_variation_by_name_index()

radarhere:
https://github.com/python-pillow/Pillow/issues/9361 found a font with duplicate variation names.

https://github.com/python-pillow/Pillow/pull/9362#issuecomment-3703830655 would like to be able to choose between the duplicates, which [`set_variation_by_name()`](https://pillow.readthedocs.io/en/stable/reference/ImageFont.html#PIL.ImageFont.FreeTypeFont.set_variation_by_name) doesn't allow for.

This PR makes the duplicate names visible in `get_variation_names()`, and adds a new method, `set_variation_by_name_index()`, to allow the user to definitively choose between them.

### DIFF
--- a/Tests/test_imagefont.py
+++ b/Tests/test_imagefont.py
@@ -753,6 +753,8 @@ def test_variation_duplicates() -> None:
         b"Black",
         b"Black Medium Contrast",
         b"Black High Contrast",
+        b"Black High Contrast",
+        b"Black High Contrast",
         b"Default",
     ]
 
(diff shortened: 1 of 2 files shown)


radarhere:
https://github.com/python-pillow/Pillow/pull/9362#issuecomment-3707022614
> I'm not sure if it's worth adding (and maintaining) extra APIs to support horribly flawed fonts.

PR #8389, opened 2024-09-18
_imaging module cleanup

homm:
Includes some useful changers from #8340.

### Improvements

* Move some useful macro such as `MAX` and `MIN` to `ImagingUtils.h` https://github.com/python-pillow/Pillow/pull/8389/commits/517f16be0203419e4a378f31ad88d453d8e2ed69
* `ImPlatform.h` is renamed to `ImagingPlatform.h` https://github.com/python-pillow/Pillow/pull/8389/commits/d73e8c24eebccc3275d31328d47b860b548970d4
* All `#undef *INT*` moved to `Jpeg.h` https://github.com/python-pillow/Pillow/pull/8389/commits/89090524600360307ed06074b01908f591b40d83
* Simplify `#define *INT*` by requiring `stdint.h` (C99) https://github.com/python-pillow/Pillow/pull/8389/commits/92bf6915b4aa0753b99f84f65729a527800c4f52
* `#include <Python.h>` once in `Imaging.h` https://github.com/python-pillow/Pillow/pull/8389/commits/84f02611ca171f5acfa1687c126e6b2ecc9fd3bb
* `#include <math.h>` in `ImagingPlatform.h` https://github.com/python-pillow/Pillow/pull/8389/commits/fbd4c98a0a7ee234edd5ea258f62bda111389c95
* Include all global libs with `<*.h>` instead of `"*.h"` https://github.com/python-pillow/Pillow/pull/8389/commits/108602be2c0a688cabacd8e21d670c24c98ce8b9
* Removed `add-imaging-libs` option (is not documented, not sure why we need it) https://github.com/python-pillow/Pillow/pull/8389/commits/c3172e857cc0718e0f81f3e6f5ec6b68079ed05e
* Use native functions instead of `PyImaging_CheckBuffer` and `PyImaging_GetBuffer` https://github.com/python-pillow/Pillow/pull/8389/commits/e9729626cc6d9914f394e5cbcb3223045c00fe51
* Use `Py_RETURN_NONE` macro when possible https://github.com/python-pillow/Pillow/pull/8389/commits/f58cd7d2275d6ababca412eb839e9222f2897ca4
* Remove unused `HAVE_LIBMPEG` https://github.com/python-pillow/Pillow/pull/8389/commits/ffa023060ae0b24922b4b4dffafd0d24f79c60a6
[...]

PR #9377, opened 2026-01-03
Rename header files

radarhere:
Part of #8389 suggests renaming ImPlatform.h to ImagingPlatform.h.

This PR does that, and I'm going to suggest additionally renaming ImDib.h to ImagingDib.h.


hugovk:
What's the reason for renaming?

https://github.com/python-pillow/Pillow/pull/8389 just says "improvement".

radarhere:
My guess would be clarity, in that it's the full word 'Imaging', rather than an abbreviation, and perhaps it makes clearer that ImagingPlatform.h is included by Imaging.h and ImagingDib.h.

But if you don't think the churn is worth it, then that's fine.

PR #9378, opened 2026-01-03
Remove add-imaging-libs option from setup.py

radarhere:
https://github.com/python-pillow/Pillow/pull/8389 suggests removing `add-imaging-libs` from setup.py, as it is undocumented.

It was added in #3399 to allow use when building in AppVeyor. It is no longer used there.

It was also not included when #7171 made sure that other setup options would work with pip >= 23.3 (released October 16, 2023), so it no longer works with recent versions of pip.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -363,7 +363,6 @@ class pil_build_ext(build_ext):
             ("disable-platform-guessing", None, "Disable platform guessing"),
             ("debug", None, "Debug logging"),
         ]
-        + [("add-imaging-libs=", None, "Add libs to _imaging build")]
     )
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


hugovk (approved):
Tip: Co-authored-by can be used to give credit to the original author:

https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors

Or even cherry pick if it's a clean commit.

radarhere:
I attempted to cherry-pick, but it wasn't working for me. I do tend to forget about 'Co-authored-by', but thankfully, I was able to do that with 'Squash and merge'.

PR #9395, opened 2026-01-13
Simplify `setimage()` by always passing extents

radarhere:
Perhaps this is just me, but reading
https://github.com/python-pillow/Pillow/blob/e2b87a04209d835d1c55633e230265dcd2c4274c/src/encode.c#L235
https://github.com/python-pillow/Pillow/blob/e2b87a04209d835d1c55633e230265dcd2c4274c/src/encode.c#L247-L255
my reaction is
> Why if x0 and x1 are zero? If we're encoding an image with zero width, then use the full image size? That seems odd. Also, what if it is zero height?

The code has been like this since PIL. I suspect the intention was to handle
https://github.com/python-pillow/Pillow/blob/e2b87a04209d835d1c55633e230265dcd2c4274c/src/PIL/Image.py#L787
where no extents are passed in, and say that in that scenario, the extents should default to the entire image.

I think it would simpler if we just explicitly passed that in from Python.

### DIFF
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -885,7 +885,7 @@ class Image:
 
         # unpack data
         e = _getencoder(self.mode, encoder_name, encoder_args)
-        e.setimage(self.im)
+        e.setimage(self.im, (0, 0) + self.size)
 
         from . import ImageFile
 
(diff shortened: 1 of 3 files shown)


hugovk (approved):
I've tried merging this, but GitHub says "Merge already in progress" 🤷 

Let's try later.

PR #9405, opened 2026-01-24
Raise ValueError if insufficient data is read from DDS RGB file

radarhere:
Alternative to #9404

In DdsRgbDecoder,
- if the file ends prematurely, stop trying to read from it
- if the amount of bytes for each value is zero (see dwRGBBitCount from https://learn.microsoft.com/en-us/windows/win32/direct3ddds/dds-pixelformat), then there will never be any data read, so return early

This will then lead to `ValueError: not enough image data` being raised.


hugovk:
I've disabled the merge queue setting for this repo and docker-images at:

https://dashboard.mergify.com/repositories?repository=Pillow&login=python-pillow

PR #9409, opened 2026-01-27
enable the mypy "truthy-bool" error code

jorenham:
This is one of the additional checks that are required by repo-review (`MY106`): https://learn.scientific-python.org/development/guides/style/#MY106
IMO it's more of a linting rule than a type-checking rule, but either way, I think it's a pretty decent one to enable. FWIW; In the projects I maintain (numpy, scipy-stubs, ...), I always enable it.

mypy docs: https://mypy.readthedocs.io/en/stable/error_code_list2.html#check-that-expression-is-not-implicitly-true-in-boolean-context-truthy-bool


src/PIL/Image.py
```diff
@@ -438,33 +438,28 @@ def preinit() -> None:
         return
 
     try:
-        from . import BmpImagePlugin
+        from . import BmpImagePlugin as BmpImagePlugin
```
jorenham, line 441: This might look a bit weird (unless you're used to writing type stubs), but it avoids a `# noqa`. I wouldn't mind using `# noqa` here instead if that's preferred

src/PIL/ImageFont.py
```diff
@@ -110,7 +110,7 @@ def _load_pilfont(self, filename: str) -> None:
                 except Exception:
                     pass
                 else:
-                    if image and image.mode in ("1", "L"):
+                    if image.mode in ("1", "L"):
```
jorenham, line 113: at this point `image` cannot be `None` anymore

Tests/test_pyarrow.py
```diff
@@ -112,8 +112,6 @@ def test_to_array(mode: str, dtype: pyarrow.DataType, mask: list[int] | None) ->
 
     reloaded = Image.fromarrow(arr, mode, img.size)
 
-    assert reloaded
```
jorenham: I could also change this to an `assert isinstance` or something, but I figured that'd also be kinda obvious.

Tests/test_qt_image_qapplication.py
```diff
@@ -56,9 +56,8 @@ def roundtrip(expected: Image.Image) -> None:
 def test_sanity(tmp_path: Path) -> None:
     # Segfault test
     app: QApplication | None = QApplication([])
-    ex = Example()
+    ex = Example()  # noqa: F841
```
jorenham, line 59: I'm not sure what the idea here is, but it looked as if the assert was purely there to avoid this `F841` error.

radarhere: Is it not preferable to avoid `noqa` where possible? My solution would have been `assert ex is not None`

jorenham: Personally I prefer being explicit about ignored errors. And `assert` statements aren't "free" (the runtime impact is probably negligible, but still).
And, if at some point, some change causes this so that `ex` *could* actually become `None`, then ruff will complain about an unused `# noqa`, which would prevent a potential bug.

hugovk (changes requested):
I don't think these changes do much to improve the readability of the code.

Some of the existing hints make core less readable and were to silence mypy. Pillow is an old code base (31 years old!), so I think we need to careful about introducing more strictness.

> This is one of the additional checks that are required by repo-review (`MY106`): [learn.scientific-python.org/development/guides/style#MY106](https://learn.scientific-python.org/development/guides/style/#MY106) IMO it's more of a linting rule than a type-checking rule, but either way, I think it's a pretty decent one to enable. FWIW; In the projects I maintain (numpy, scipy-stubs, ...), I always enable it.

That link says:

> The extra strict options shown above, like `warn_unreachable` MY103, and `ignore-without-code` MY104, `redundant-expr` MY105, and `truthy-bool` MY106 can trigger too often (like on sys.platform checks) and have to be ignored occasionally, but can find some significant logic errors in your typing.

It seems to be triggering too often in this case, let's skip this one.

jorenham:
Seems fair; closing

PR #9410, opened 2026-01-27
mypy strict mode

jorenham:
This configures mypy to run in "strict mode", which according to `mypy --help` enables the following flags:

> --warn-unused-configs, --disallow-any-generics, --disallow-subclassing-any, --disallow-untyped-calls, --disallow-untyped-defs, --disallow-incomplete-defs, --check-untyped-defs, --disallow-untyped-decorators, --warn-redundant-casts, --warn-unused-ignores, --warn-return-any, --no-implicit-reexport, --strict-equality,
                            --strict-bytes, --extra-checks

Enabling this caused a bunch of typing errors to surface, which are now also fixed. 

Most of these are related to `bytes`/`bytearray`/`memoryview` not being assignable to each other. So I used their common denominator, `collections.abc.Buffer`, or rather its backport from `typing_extensions` (which is always available `if TYPE_CHECKING` because typeshed conveniently lies that it's part of the standard library).


src/PIL/Image.py
```diff
@@ -86,37 +88,45 @@ class DecompressionBombError(Exception):
...
-        warnings.warn(
-            "The _imaging extension was built for another version of Python.",
-            RuntimeWarning,
-        )
-    elif str(v).startswith("The _imaging extension"):
-        warnings.warn(str(v), RuntimeWarning)
-    # Fail here anyway. Don't let people run with a mostly broken Pillow.
-    # see docs/porting.rst
-    raise
+    # mypy will not recognize `core` as public symbol when import as
```
```suggestion
    # mypy will not recognize `core` as a public symbol when imported as
```

jorenham: Woops, good catch. But on the bright side; I guess you could take this as prove that I'm not a vibe-coder haha.

pyproject.toml
```diff
@@ -216,11 +216,10 @@ testpaths = [
...
 python_version = "3.10"
 pretty = true
-disallow_any_generics = true
-disallow_untyped_defs = true
+strict = true
+disallow_subclassing_any = false
+disallow_untyped_calls = false
 enable_error_code = "ignore-without-code"
-extra_checks = true
-follow_imports = "silent"
```
radarhere: I don't see anything at https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-strict that mentions `follow_imports`? What was the thinking behind changing this?

jorenham: That's right, `follow_imports` is unrelated to `--strict`. It's just unnecessary to suppress those import errors.

radarhere:
The changes here should be re-evaluated after #9414's debate about Python versions is resolved.

jorenham:
Anything I can do to help with that?

radarhere:
https://github.com/python-pillow/Pillow/pull/9414#issuecomment-3824350914
> Is there authoritative guidance on best practices here?

The only thing I can think of is to ask someone at mypy for their official view on whether we need to run mypy against the oldest supported Python, newest Python, or everything in between.

jorenham:
> [#9414 (comment)](https://github.com/python-pillow/Pillow/pull/9414#issuecomment-3824350914)
> 
> > Is there authoritative guidance on best practices here?
> 
> The only thing I can think of is to ask someone at mypy for their official view on whether we need to run mypy against the oldest supported Python, newest Python, or everything in between.

@JukkaL sorry to bother you, but could I redirect this question to you?

jorenham:
> [#9414 (comment)](https://github.com/python-pillow/Pillow/pull/9414#issuecomment-3824350914)
> 
> > Is there authoritative guidance on best practices here?
> 
> The only thing I can think of is to ask someone at mypy for their official view on whether we need to run mypy against the oldest supported Python, newest Python, or everything in between.

Jukka seems busy, so @ilevkivskyi, could you perhaps help us out here?

ilevkivskyi:
I don't think we have any "official" recommendations on this. But unofficially we recommend:
* Always run mypy on the oldest supported Python version.
* If your code has known platform-specific logic (e.g. Linux vs Windows), run the above on as many platforms as needed.
* You may also run mypy on the newest supported Python version (in this case it makes sense to use `--enable-error-code=deprecated` to see deprecations early).

radarhere:
Closing, as https://github.com/python-pillow/Pillow/pull/9803 has been created instead.

radarhere:
@jorenham sorry for any confusion. It appeared that this PR was in need of rebase, wasn't being updated, and someone else wanted to pick up the ball. If you would like this to be re-opened, it can be, or if you had any comments on #9803, feel free to add them.

jorenham:
No worries.

PR #9413, opened 2026-01-29
Patch libavif for svt-av1 4.0 compatibility

hugovk:
macOS builds have started failing.

Pass: https://github.com/python-pillow/Pillow/actions/runs/21415347226/job/61666976278

Fail: https://github.com/python-pillow/Pillow/actions/runs/21472202845/job/61847296515

The difference is svt-av1 (3.1.2) vs svt-av1 (4.0.0) in Homebrew.

libavif doesn't yet support 4.0.0. The fix has been made but not merged: https://github.com/AOMediaCodec/libavif/pull/2971

Our options include:

* wait for a libavif release
* temporarily comment out the svt-av1 from `macos-install.sh`
* install an old svt-av1, but Homebrew doesn't have old versions
* apply the libavif patch locally

Here's a patch to get the CI back to green.

### DIFF
--- a/depends/install_libavif.sh
+++ b/depends/install_libavif.sh
@@ -7,6 +7,10 @@ version=1.3.0
 
 pushd libavif-$version
 
+# Apply patch for SVT-AV1 4.0 compatibility
+# Pending release of https://github.com/AOMediaCodec/libavif/pull/2971
+patch -p1 < ../libavif-svt4.patch
+
 if [ $(uname) == "Darwin" ] && [ -x "$(command -v brew)" ]; then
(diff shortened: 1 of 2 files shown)


radarhere (approved):

radarhere:
A new libavif release is estimated to be in two weeks - https://github.com/AOMediaCodec/libavif/pull/2971#issuecomment-3826961020

Even if they don't exactly hit that target, we should be able to revert this and upgrade before the next Pillow release.

radarhere:
There is now a new libavif release, and I've created #9460 to revert this.

PR #9414, opened 2026-01-30
Run mypy against multiple Python versions

radarhere:
While exploring #9410, the user found that our type hints are not being adequately checked across different Python versions - https://github.com/jorenham/Pillow/pull/1#discussion_r2745715320

This PR stops specifying the mypy Python version in pyproject.toml, so that it [defaults to the current version being run](https://mypy.readthedocs.io/en/stable/config_file.html#confval-python_version), and adds a matrix to our Lint job to check multiple versions.

Python >= 3.12 required some changes in Buffer handling - https://github.com/python-pillow/Pillow/actions/runs/21516054322/job/61994586341


hugovk:
Do we really need to run mypy on every version? Should we run on the oldest supported, or latest? Is there authoritative guidance on best practices here?

If we do decide to type check on multiple versions, let's move mypy to its own job (can name it like `typing` for when we switch to `ty` or `pyrefly` or something) and only apply the matrix there. The other linting can remain on an single version.

And can we continue to use a new Python interpreter but call mypy with `--python-version` instead of setting up older interpreters?

https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-python-version

radarhere:
> let's move mypy to its own job (can name it like typing for when we switch to ty or pyrefly or something)

> And can we continue to use a new Python interpreter but call mypy with --python-version instead of setting up older interpreters?

Sure, done.

> Do we really need to run mypy on every version? Should we run on the oldest supported, or latest? Is there authoritative guidance on best practices here?

https://github.com/python/mypy/issues/12286 is an issue requesting the ability to check multiple versions at once. From the fact that it is still open, you might infer that this isn't an official priority.

@jorenham did you have any thoughts?

jorenham:
> @jorenham did you have any thoughts?

In the projects I maintain I run all supported typecheckers against the full range of supported Python versions. The main reason is to ensure that all `if sys.version_info < ...` branches are actually checked, and to avoid incorrect `typing`/`typing_extensions`/ imports.

radarhere:
> Is there authoritative guidance on best practices here?

We have received input from one of the mypy maintainers - https://github.com/python-pillow/Pillow/pull/9410#issuecomment-4198872025
> I don't think we have any "official" recommendations on this. But unofficially we recommend:
> 
> * Always run mypy on the oldest supported Python version.
> * If your code has known platform-specific logic (e.g. Linux vs Windows), run the above on as many platforms as needed.
> * You may also run mypy on the newest supported Python version (in this case it makes sense to use `--enable-error-code=deprecated` to see deprecations early).

PR #9417, opened 2026-02-04
Factor out saving metrics in FontFile

fjhenigman:
Facilitates creating an ImageFont from a PCF or BDF file on the fly, without creating any new files, with code like this:
```python
def ifont(path):
    with open(path, 'rb') as f:
        ff = PcfFontFile.PcfFontFile(f)
    ff.compile()
    buf = io.BytesIO()
    ff.save_metrics(buf)
    buf.seek(0)
    imgfont = ImageFont.ImageFont()
    imgfont._load_pilfont_data(buf, ff.bitmap)
    return imgfont
```
This isn't ideal, creating a file in memory and using an underscore function.  Maybe it's preferable to add functionality to ImageFont.py instead?

### DIFF
--- a/src/PIL/FontFile.py
+++ b/src/PIL/FontFile.py
@@ -123,12 +123,15 @@ class FontFile:
 
         # font metrics
         with open(os.path.splitext(filename)[0] + ".pil", "wb") as fp:
-            fp.write(b"PILfont\n")
-            fp.write(f";;;;;;{self.ysize};\n".encode("ascii"))  # HACK!!!
-            fp.write(b"DATA\n")
-            for id in range(256):
-                m = self.metrics[id]
-                if not m:
-                    puti16(fp, (0,) * 10)
-                else:
-                    puti16(fp, m[0] + m[1] + m[2])
+            self.save_metrics(fp)
+
+    def save_metrics(self, fp: BinaryIO) -> None:
+        fp.write(b"PILfont\n")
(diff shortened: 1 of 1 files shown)


radarhere:
Hi. To outline the problem, you've said that you would like to convert `FontFile` instances to `ImageFont` instances.

I'm guessing you saw https://pillow.readthedocs.io/en/stable/reference/ImageFont.html#imagefont-module and https://github.com/python-pillow/pillow-scripts/blob/main/Scripts/pilfont.py. They've sort of presumed that you only need to convert once, or maybe only convert a small number of fonts, but presumably you would like your users to be able to make this conversion for any PCF or BDF font they might choose at runtime, and you would like to do so efficiently.

My suggestion would be adding a `to_imagefont()` method to `FontFile`. What do you think?

Also, just to properly understand the need, what is it that you want to do with the `FontFile` instances? Draw text?

fjhenigman:
Yes exactly, I am using ImageDraw to draw text and would like to support bdf and pcf fonts.  If it was just for me or just a couple fonts, I would use pilfont.py to convert them and deal with the resulting bitmaps and .pil files.  But for an application with wider use, it would be nice if any user could directly feed it any pcf or bdf files they may have.
I will prepare another PR adding a `to_imagefont()` method to `FontFile` as you suggest.
Close this PR unless you want it.
Thanks.

radarhere:
#9419 has been merged.

PR #9424, opened 2026-02-09
CI: Disable pip upgrade warning

hugovk:
Add `PIP_DISABLE_PIP_VERSION_CHECK=1` to prevent warnings like:

```
macos-latest Python 3.14t
[notice] A new release of pip is available: 25.3 -> 26.0.1
[notice] To update, run: /Library/Frameworks/Python.framework/Versions/3.14/bin/python3.14 -m pip install --upgrade pip
```

https://github.com/python-pillow/Pillow/actions/runs/21798050860

After:

https://github.com/hugovk/Pillow/actions/runs/21818838104

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,6 +29,7 @@ concurrency:
 env:
   COVERAGE_CORE: sysmon
   FORCE_COLOR: 1
+  PIP_DISABLE_PIP_VERSION_CHECK: 1
 
 jobs:
   build:


radarhere (approved):
While we [recommend upgrading pip](https://pillow.readthedocs.io/en/stable/installation/basic-installation.html#basic-installation), I expect that we're ignoring the warning here because we want to verify that Pillow can still be installed successfully if users ignore our recommendation.

hugovk:
The warning is actually coming from inside actions/setup-python, and only for 3.13t - 3.15t on macOS. (It'd probably also show up for other versions in the period between a new release and GitHub rebuilding their caches.)

I think it's because they don't (yet?) have cached versions of those Python versions, so it has more on-demand setup to do.

I've previously hunted down and fixed the upstream problem when the warning came whilst upgrading pip itself, by adding `--disable-pip-version-check`: https://github.com/actions/python-versions/pull/117

This time the warning is actually happening right after upgrading pip, and comes when installing OpenSSL certs: 

```
  Upgrading pip...
  Looking in links: /var/folders/yz/zr09txvs5dn18vt4cn21kzl40000gn/T/tmppp0ocgq2
  Requirement already satisfied: pip in /Library/Frameworks/PythonT.framework/Versions/3.14/lib/python3.14t/site-packages (25.3)
  Collecting pip
  Downloading pip-26.0.1-py3-none-any.whl.metadata (4.7 kB)
  Downloading pip-26.0.1-py3-none-any.whl (1.8 MB)
  25l
  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 0.0/1.8 MB ? eta -:--:--
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 1.8/1.8 MB 40.1 MB/s  0:00:00
  25h
  Installing collected packages: pip
  Attempting uninstall: pip
  Found existing installation: pip 25.3
  Uninstalling pip-25.3:
  Successfully uninstalled pip-25.3
  Successfully installed pip-26.0.1
  Install OpenSSL certificates
  Requirement already satisfied: certifi in /Library/Frameworks/Python.framework/Versions/3.14/lib/python3.14/site-packages (2026.1.4)
  Error: [notice] A new release of pip is available: 25.3 -> 26.0.1
  [notice] To update, run: /Library/Frameworks/Python.framework/Versions/3.14/bin/python3.14 -m pip install --upgrade pip
```

https://github.com/python-pillow/Pillow/actions/runs/21798050860/job/62888796977#step:3:33

That's somewhere here:

https://github.com/actions/python-versions/blob/e49da096613ccc8acd929a8d5070c735ca19d790/installers/macos-pkg-setup-template.sh#L85-L91

Which calls `Install Certificates.command`, which contains:

```
    subprocess.check_call([sys.executable,
        "-E", "-s", "-m", "pip", "install", "--upgrade", "certifi"])
```

https://github.com/python/cpython/blob/b22ff1e543a458fff7b11225bf743d1730bbcc71/Mac/BuildScript/resources/install_certificates.command#L27-L28

For some reason, that must be a different pip than the one that was just upgraded. I don't feel like digging further into GitHub Actions, but we could probably also add `--disable-pip-version-check` to this script in CPython. I'll follow up on that, but this env var is a good enough workaround for now.

PR #9426, opened 2026-02-10
Raise ValueError if ImageOps border has unsupported format

veeceey:
## Summary

- **Fix bug in `ImageOps._border()`**: When given a tuple with a length other than 2 or 4 (e.g. a 1-tuple or 3-tuple), the function raised an unhelpful `UnboundLocalError` because the `left`, `top`, `right`, `bottom` variables were never assigned. This now raises a clear `ValueError` with a descriptive message instead.

- **Document `rgba()` color format in `ImageColor`**: The `rgba(red, green, blue, alpha)` color string format has been supported in code and tested for a long time, but was missing from the `ImageColor` documentation. This adds it to the list of supported color string formats.

## Test plan

- [x] Added parametrized test `test_expand_invalid_border` that verifies `ValueError` is raised for 1-tuple, 3-tuple, and 5-tuple border values
- [x] Verified existing tests still pass (ruff and black checks pass)
- [x] Manually verified the fix resolves the `UnboundLocalError`

### DIFF
--- a/Tests/test_imageops.py
+++ b/Tests/test_imageops.py
@@ -256,6 +256,13 @@ def test_expand_palette(border: int | tuple[int, int, int, int]) -> None:
         assert_image_equal(im_cropped, im)
 
 
+@pytest.mark.parametrize("border", ((1,), (1, 2, 3), (1, 2, 3, 4, 5)))
+def test_expand_invalid_border(border: tuple[int, ...]) -> None:
+    im = Image.new("1", (1, 1))
+    with pytest.raises(ValueError):
+        ImageOps.expand(im, border)
+
+
 def test_colorize_2color() -> None:
     # Test the colorizing function with 2-color functionality
 
--- a/src/PIL/ImageOps.py
+++ b/src/PIL/ImageOps.py
@@ -37,7 +37,7 @@ def _border(border: int | tuple[int, ...]) -> tuple[int, int, int, int]:
         elif len(border) == 4:
             left, top, right, bottom = border
         else:
-            msg = "border must be an integer or a 2- or 4-tuple"
+            msg = "border must be an integer, or a tuple of two or four elements"
             raise ValueError(msg)
     else:
         left = top = right = bottom = border
--- a/src/PIL/ImageOps.py
+++ b/src/PIL/ImageOps.py
@@ -36,6 +36,9 @@ def _border(border: int | tuple[int, ...]) -> tuple[int, int, int, int]:
             left, top = right, bottom = border
         elif len(border) == 4:
             left, top, right, bottom = border
+        else:
+            msg = "border must be an integer or a 2- or 4-tuple"
+            raise ValueError(msg)
     else:
         left = top = right = bottom = border
     return left, top, right, bottom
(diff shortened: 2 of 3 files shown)


Tests/test_imageops.py
```diff
@@ -256,6 +256,13 @@ def test_expand_palette(border: int | tuple[int, int, int, int]) -> None:
         assert_image_equal(im_cropped, im)
 
 
+@pytest.mark.parametrize("border", ((1,), (1, 2, 3), (1, 2, 3, 4, 5)))
+def test_expand_invalid_border(border: tuple[int, ...]) -> None:
+    with Image.open("Tests/images/hopper.ppm") as im:
+        with pytest.raises(ValueError):
+            ImageOps.expand(im, border)
```
radarhere:
```suggestion
    im = Image.new("1", (1, 1))
    with pytest.raises(ValueError):
        ImageOps.expand(im, border)
```
Small simplification.

radarhere:
Hi. These seem like two unrelated changes. In the future, it's worth considering whether these should be two separate PRs, or at least two separate commits.

I'm not sure how much of this you used AI for, but if you wouldn't mind explaining your human motivation a bit - did you actually find yourself triggering the `UnboundLocalError` unintentionally?

veeceey:
Fair point on splitting, noted for next time.

On the motivation -- I was looking at the ImageOps.expand code and noticed the border validation raises ValueError for negative values, but if you pass something like a 3-tuple (which is not a valid border spec), the unpacking just silently fails and `border` stays partially unpacked, leaving `bottom` unbound. Then the subsequent code hits an UnboundLocalError, which is confusing because the real issue is the invalid input. So yes, I did trace through it manually -- the UnboundLocalError masks the actual problem.

Will apply the test simplification you suggested.

Tests/test_imageops.py
veeceey: Done, applied the simplification. Thanks\!

veeceey:
Hi @radarhere, just checking in — I've applied your suggested test simplification and responded to your earlier questions. All CI checks are green. Would you be able to take another look when you get a chance? Thanks\!

radarhere (approved):
https://github.com/python-pillow/Pillow/pull/9433#issuecomment-3887082458 might like us to consider our policy towards AI before merging

src/PIL/ImageOps.py
```diff
@@ -36,6 +36,9 @@ def _border(border: int | tuple[int, ...]) -> tuple[int, int, int, int]:
             left, top = right, bottom = border
         elif len(border) == 4:
             left, top, right, bottom = border
+        else:
+            msg = "border must be an integer or a 2- or 4-tuple"
```
radarhere:
```suggestion
            msg = "border must be an integer, or a tuple of two or four elements"
```
Feel free to ignore this idea if you want, it just seems a bit more natural to me.

veeceey:
Hey @radarhere, thanks for the approval\! Totally understand wanting to sort out the policy discussion in #9433 first. Happy to wait — just let me know if there's anything else needed on my end in the meantime.

src/PIL/ImageOps.py
veeceey: Good call, updated to use that wording. Reads much better.

PR #9443, opened 2026-02-20
Fix BMP RLE delta escape reading from wrong file position

bysiber:
The BMP RLE decoder's delta escape handler reads the displacement bytes twice from the file stream. It reads 2 bytes into `bytes_read` for validation, then calls `self.fd.read(2)` again instead of unpacking `bytes_read`:

```python
bytes_read = self.fd.read(2)   # read delta bytes
if len(bytes_read) < 2:
    break
right, up = self.fd.read(2)    # reads 2 MORE bytes, discards the actual delta values
```

The real delta displacement values are thrown away, and `right` / `up` come from whatever happens to be next in the stream. Any RLE-compressed BMP that uses delta escape codes gets decoded with wrong pixel offsets, producing corrupted output.


radarhere:
https://www.digicamsoft.com/bmp/bmp.html supports your change
> 2       Delta. The two bytes following the escape contain unsigned values indicating the horizontal and vertical offsets of the next pixel from the current position.

bysiber:
Thanks for confirming with the spec, glad the fix aligns with what the BMP reference says.

radarhere:
I've created https://github.com/bysiber/Pillow/pull/3 to add a test.

PR #9445, opened 2026-02-20
Fix `self.decode` typo

bysiber:
In `ImageFile.Parser`, the class uses `self.decoder` throughout (initialized on line 516, checked on line 545, set on line 590). But line 584 assigns `self.decode = None` (missing the trailing "r"), which creates a new attribute that's never referenced anywhere.

The intent is clearly to reset `self.decoder` to `None` when the image has custom load code or multiple tiles, so the parser falls through to the `elif self.image:` branch on subsequent `feed()` calls.

### DIFF
--- a/src/PIL/ImageFile.py
+++ b/src/PIL/ImageFile.py
@@ -579,9 +579,7 @@ class Parser:
                 pass  # not enough data
             else:
                 flag = hasattr(im, "load_seek") or hasattr(im, "load_read")
-                if flag or len(im.tile) != 1:
-                    pass  # custom load code, or multiple tiles
-                else:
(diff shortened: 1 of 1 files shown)


radarhere:
I've created https://github.com/bysiber/Pillow/pull/1 with a suggestion to simplify the code.

hugovk:
Rather than:
```python
if condition:
    pass
else:
    do_thing()
```
Invert and drop the `else`:
```python
if not condition:
    do_thing()
```
(And rewrite the condition.)

bysiber:
Good call, inverted the condition so it reads more naturally. Pushed.

PR #9453, opened 2026-03-05
Update harfbuzz to 13.0.1

radarhere:
harfbuzz 13.0.1 has been released - https://github.com/harfbuzz/harfbuzz/releases/tag/13.0.1

### DIFF
--- a/.github/workflows/wheels-dependencies.sh
+++ b/.github/workflows/wheels-dependencies.sh
@@ -90,11 +90,7 @@ fi
 ARCHIVE_SDIR=pillow-depends-main
 
 # Package versions for fresh source builds.
-if [[ -n "$IOS_SDK" ]]; then
-  FREETYPE_VERSION=2.13.3
-else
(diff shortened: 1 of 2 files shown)


khaledhosny:
You can disable the new harfbuzz-raster and harfbuzz-vector (and even the old harfbuzz-subset) as you don't use them. The build error seem to be in harfbuzz-vector.

radarhere:
Thanks, I've updated the commit to disable vector.

khaledhosny:
This [PR](https://github.com/harfbuzz/harfbuzz/pull/5792) might also fix the `harfbuzz-vector` build failure.

radarhere:
It does, thanks. I'd presumed that because manylinux2014 is older, you'd dropped support for it, intentionally or unintentionally.

khaledhosny:
We try to support as old systems as possible.

hugovk:
Thanks both!

PR #9454, opened 2026-03-05
Upgrade CI from `macos-15-intel` to `macos-26-intel`

hugovk:
macOS 26 is now available on GitHub Actions:

* `macos-26` - arm64
* `macos-26-intel` - x64

https://github.blog/changelog/2026-02-26-macos-26-is-now-generally-available-for-github-hosted-runners/

This PR bumps `macos-15-intel` to `macos-26-intel`.

I've left `macos-latest` as-is. They're still pointing to macOS 15, but that will change in the near future.


radarhere:
This is the last time that we will need to upgrade like this, as https://en.wikipedia.org/wiki/MacOS_Tahoe
> Tahoe is the final version of macOS that supports Macs with Intel processors

hugovk:
Yes. 

Tahoe is EOL in ~3 years, but GitHub's [N-1 support policy](https://github.com/actions/runner-images?tab=readme-ov-file#software-and-image-support):

> * We support (at maximum) 2 GA images and 1 beta image at a time. We begin the deprecation process of the oldest image label once the newest OS image label has been released to GA.

So `macos-26-intel` will be around for ~2 years.

This could of course change, before they introduced the `*-intel` runners, `macos-13` was to be the last Intel.

PR #9463, opened 2026-03-12
Add riscv64 to Linux wheel build matrix

### LINKED ISSUE #9462: Add riscv64 (linux_riscv64) wheel to PyPI releases
### Summary

`pip install pillow` on riscv64 Linux currently requires building from source (~5 min on a 1.6 GHz RISC-V SoC). Adding `linux_riscv64` to the wheel build matrix would give riscv64 users a prebuilt wheel.

### Evidence

| | |
|---|---|
| **Tested wheel** | [`pillow-12.1.1-cp313-cp313-linux_riscv64.whl`](https://github.com/gounthar/riscv64-python-wheels/releases/download/v2026.03.11-cp313/pillow-12.1.1-cp313-cp313-linux_riscv64.whl) |
| **Hardware** | BananaPi F3 (SpacemiT K1, rv64imafdcv, 8 cores @ 1.6 GHz, 16 GB RAM) |
| **Python** | 3.13 (CPython) |
| **Build system** | cibuildwheel |
| **Build time** | ~5 min (native, on hardware) |

The wheel imports and passes basic smoke tests on riscv64 Linux.

### Suggested CI change

**Workflow**: `.github/workflows/wheels.yml`

Add riscv64 entries to the `build-native-wheels` matrix:

```yaml
# In matrix.include, add:
- name: "manylinux_2_28 riscv64"
  platform: linux
  os: ubuntu-latest
  cibw_arch: riscv64
  build: "*manylinux*"
```

Add QEMU step (after checkout, before cibuildwheel):

```yaml
- name: Set up QEMU
  if: matrix.cibw_arch == 'riscv64'
  uses: docker/setup-qemu-action@v3
  with:
    platforms: riscv64
```

### Ecosystem context

- `manylinux_2_28_riscv64` images are available on `quay.io/pypa/` (landed in pypa/manylinux, 2025)
- cibuildwheel 3.x supports riscv64 via QEMU emulation
- auditwheel supports riscv64 platform tags
- maturin-action supports `riscv64gc-unknown-linux-gnu` cross-compilation
- Packages already shipping riscv64 wheels on PyPI: aiohttp, yarl, regex, markupsafe, charset-normalizer, rpds-py, multidict, propcache, watchfiles, rignore, setproctitle
- RISC-V hardware is shipping: SiFive HiFive, SpacemiT K1/K3, Sophgo SG2044 (64-core)

### Our temporary index

While upstream support is pending, we maintain a [PEP 503 index](https://gounthar.github.io/riscv64-python-wheels/simple/) with 50+ riscv64 wheels for the Python ML/AI stack, built natively on RISC-V hardware.

Tracking repo: https://github.com/gounthar/riscv64-python-wheels

gounthar:
Fixes #9462.

 * Add `manylinux_2_39 riscv64` entry to `build-native-wheels` matrix in `wheels.yml`
 * Add QEMU setup step for riscv64 emulation on x86_64 runners
 * Build `manylinux_2_39_riscv64` wheels via cibuildwheel + QEMU
 * Add two temporary C library workarounds in `wheels-dependencies.sh` (see below)

### Evidence

- CI run (QEMU, manylinux_2_39): https://github.com/python-pillow/Pillow/actions/runs/25095752245/job/73532532328 — 3h 49m, SUCCESS
- Native RISE riscv64 runner (fork, for comparison): https://github.com/gounthar/Pillow/actions/runs/25096241934 — 2h 6m, SUCCESS

### C library workarounds

Two dependencies need temporary build flags on riscv64. Happy to revisit the approach if there's a preferred way to handle these.

- **libjpeg-turbo** (`-DWITH_SIMD=FALSE`): `simdcoverage.c` references RVV SIMD functions that exist in upstream main but aren't in any release yet. Removes once a release includes those symbols.
- **libaom** (inside libavif, `-DAOM_TARGET_CPU=generic`): riscv64 RVV intrinsic calls are treated as undeclared by GCC 14. Removes once the undeclared function calls in `arch/riscv/` are fixed upstream.

### Context

- `manylinux_2_39_riscv64` is the correct tag for RISE riscv64 runner images
- cibuildwheel 3.x supports riscv64 via QEMU
- RISC-V hardware is shipping (SiFive, SpacemiT, Sophgo SG2044)
- Several packages already ship riscv64 wheels: aiohttp, yarl, multidict, propcache


radarhere:
Hi. At the moment, you've only added the QEMU step, you haven't actually changed the matrix.

As a word of warning, I expect that once you do update the matrix, this will take a very long time to build - https://github.com/python-pillow/Pillow/issues/9462#issuecomment-4044643389

gounthar:
@radarhere: oopsie; I'd added the QEMU step but forgot the matrix entry. Fixed in 7b026c6: added a `manylinux_2_28 riscv64` entry scoped to `cp3{12,13}-manylinux*` only (2 wheels, as suggested in https://github.com/python-pillow/Pillow/issues/9462#issuecomment-4045628666).

This should trigger a CI run. I expect it will either hit the libaom/RVV build error or the 6-hour timeout (or both). Either way, we'll have concrete data on what happens under QEMU, which will help decide whether to pursue the self-hosted runner path or close this.

radarhere:
Regarding the general test failures, they should be fixed by #9464

Regarding riscv64, it did fail to build libavif - https://github.com/python-pillow/Pillow/actions/runs/23025132993/job/66876973143?pr=9463#step:6:5045

I don't expect riscv64 to hit the 6 hour limit if you are only building two Python versions.

gounthar:
@radarhere Thanks for pointing to the libavif failure. I'll investigate the build log; it looks like it could be a libaom/RVV (RISC-V Vector) codegen issue in the cross-compilation path.

On the build time front: if QEMU overhead becomes a concern, https://cloud-v.co/ provides free native riscv64 GitHub Actions runners to open source projects. That would speed things up and give more representative results. But let me first focus on getting libavif to build (or not) under QEMU.
There's no point discussing runners if the build itself doesn't work yet. 🤷

gounthar:
Following up on the libavif/libaom failure. There is an active review on the aomedia Gerrit fixing RVV handling for specific block sizes in riscv64: https://aomedia-review.googlesource.com/c/aom/+/208401

Not sure yet if this is the exact issue we hit, but it is in the same area. I will rebase and re-run once it lands to see if it unblocks the build.

.github/workflows/wheels.yml
```diff
@@ -116,6 +122,12 @@ jobs:
         with:
           python-version: "3.x"
 
+      - name: Set up QEMU
+        if: matrix.cibw_arch == 'riscv64'
+        uses: docker/setup-qemu-action@v3
```
```suggestion
        uses: docker/setup-qemu-action@v4
```

gounthar:
Found and fixed the build failure. It was not in dav1d - that completed successfully.

The actual error was in libaom (bundled as a subproject of libavif 1.4.1), specifically `av1/common/riscv/highbd_convolve_rvv.c`. That file calls `av1_highbd_convolve_y_sr_c` without including the header that declares it. GCC 14 in `manylinux_2_39` compiles with `-std=gnu23` which treats implicit function declarations as errors (not just warnings).

Fix: pass `-DAOM_TARGET_CPU=generic` for riscv64 in `build_libavif`. This skips the arch-specific RVV code in libaom. There is no correctness loss - QEMU-based builds don't benefit from RVV optimizations, and the same flag is already used for the IOS/x86_64 cross-compilation case.

.github/workflows/wheels-dependencies.sh
```diff
@@ -204,6 +204,12 @@ function build_libavif {
             -DAVIF_CODEC_DAV1D=LOCAL
         )
     fi
+    # libaom's riscv64 RVV code calls functions without declarations, which
+    # GCC 14 (manylinux_2_39) treats as errors. Disable arch-specific AOM
+    # optimizations for riscv64; QEMU-based builds don't benefit from them.
+    if [[ "$(uname -m)" == "riscv64" ]]; then
+        libavif_cmake_flags+=(-DAOM_TARGET_CPU=generic)
+    fi
```
radarhere, line 214: Do you know if libaom plan to address this in the future? Is there an issue tracking the progress?

.github/workflows/wheels.yml
```diff
@@ -94,6 +94,12 @@ jobs:
             os: ubuntu-24.04-arm
             cibw_arch: aarch64
             build: "*musllinux*"
+          - name: "manylinux_2_39 riscv64"
+            platform: linux
+            os: ubuntu-latest
+            cibw_arch: riscv64
+            build: "cp3{12,13}-manylinux*"
```
radarhere: Why Python 3.12 and 3.13? Why not 3.13 and 3.14?

.github/workflows/wheels-dependencies.sh
gounthar: Yes, there is an upstream issue tracking it: https://aomedia.issues.chromium.org/issues/492439207

There is also a Gerrit CL in review that addresses RVV handling for specific block sizes: https://aomedia-review.googlesource.com/c/aom/+/208401

Not sure yet how soon it will land, but once it does the `-DAOM_TARGET_CPU=generic` workaround could probably be dropped. For now keeping it in is the safe option.

.github/workflows/wheels.yml
gounthar: No strong reason for 3.12 over 3.14. Updated to cp313+cp314 in e2d2677.

.github/workflows/wheels.yml
```diff
@@ -116,6 +122,12 @@ jobs:
         with:
           python-version: "3.x"
 
+      - name: Set up QEMU
+        if: matrix.cibw_arch == 'riscv64'
+        uses: docker/setup-qemu-action@v4
```
```suggestion
        uses: docker/setup-qemu-action@ce360397dd3f832beb865e1373c09c0e9f86d70a # v4.0.0
```
As per https://github.com/python-pillow/Pillow/pull/9568

gounthar:
The manylinux_2_39 riscv64 job is now green: https://github.com/python-pillow/Pillow/actions/runs/25095752245/job/73532532328 (3h 49m via QEMU).

For reference, I also ran the same build on a native RISE riscv64 runner in a fork branch: https://github.com/gounthar/Pillow/actions/runs/25096241934, 2h 6m on `rise-riscv-runner-sunz1ojxz`. Roughly 45% faster, which answers the earlier question about whether native runners would help.

The two C library workarounds (libjpeg-turbo: `-DWITH_SIMD=FALSE`, libaom: `-DAOM_TARGET_CPU=generic`) are both confirmed working. Both are temporary: libjpeg-turbo's RVV SIMD code exists in upstream main but isn't in any release yet, and libaom has undeclared RVV function calls that GCC 14 rejects.

gounthar:
@radarhere I cherry-picked your commit (fb52ed018) from the fork: instead of skipping all riscv64 tests, only the timeout-sensitive ones are now skipped.

.github/workflows/wheels.yml
gounthar: Done in 43ac97cf, QEMU action pinned to `ce360397dd3f832beb865e1373c09c0e9f86d70a`. Good spot, I hadn't seen the change in #9568.

.github/workflows/wheels.yml
```diff
@@ -113,6 +119,12 @@ jobs:
         with:
           python-version: "3.x"
 
+      - name: Set up QEMU
+        if: matrix.cibw_arch == 'riscv64'
+        uses: docker/setup-qemu-action@ce360397dd3f832beb865e1373c09c0e9f86d70a # v4.0.0
```
```suggestion
        uses: docker/setup-qemu-action@06116385d9baf250c9f4dcb4858b16962ea869c3 # v4.1.0
```

gounthar:
This has been green for a while now: the `manylinux_2_39 riscv64` job passed on the 18th, and the branch has been kept in sync with main since. The two C library workarounds (`-DWITH_SIMD=FALSE` for libjpeg-turbo, `-DAOM_TARGET_CPU=generic` for libaom) are still the same ones described above, and the test skips are the narrower set @radarhere asked for.

Is there anything else you'd like me to change before this queues, or is it just waiting for a spare moment? Happy to rebase, split it up, or drop the riscv64 job back out of the release workflow and keep it PR-only if that is an easier first step.

.github/workflows/wheels.yml
```diff
@@ -120,6 +120,12 @@ jobs:
             os: ubuntu-24.04-arm
             cibw_arch: aarch64
             build: "*musllinux*"
+          - name: "manylinux_2_39 riscv64"
+            platform: linux
+            os: ubuntu-latest
+            cibw_arch: riscv64
+            build: "cp3{13,14}-manylinux*"
+            manylinux: "manylinux_2_39"
```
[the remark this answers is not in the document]
aclark4life: @gounthar Thoughts?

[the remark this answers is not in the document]
radarhere: Building all versions on slower riscv64 led to [hitting the six hour timeout](https://github.com/python-pillow/Pillow/issues/9462#issuecomment-4044643389).

Restricting it to only a few versions is how this PR has chosen to avoid that.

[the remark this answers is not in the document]
gounthar: radarhere has it right, that is the whole reason. Six hour job ceiling, and riscv64 under QEMU eats into it fast.

Numbers from this branch, for what they are worth: the last CI run took 3h36m for cp313 and cp314 together. When radarhere tried all supported versions back in March it hit the six hour wall even with libavif skipped, so the restriction comes from measurement rather than caution.

On the EXPECTED_DISTS half of the comment, that part is already handled: the PR moves it 78 to 80, which is exactly the two wheels the selector produces. If the selector ever widens, that number moves with it.

The documentation point is fair, and I should have done it up front. I can push a comment on the matrix entry explaining why it is two versions, so nobody has to dig through issue history to work it out. Happy to do that now if you want it in this PR.

One thing I should be upfront about: as it stands, 3.11 and 3.12 on riscv64 still fall back to source builds. What would buy room for more versions is native hardware instead of emulation. The same matrix entry ran in 2h06m on a native riscv64 runner from the RISE project, which offers a hosted riscv64 runner pool to open source projects through a GitHub App, so nothing self-hosted on your side. I can make an introduction if that is interesting, though you are far better placed than I am to judge whether riscv64 demand justifies any of this yet.

radarhere:
> `-DWITH_SIMD=FALSE` for libjpeg-turbo

#9840 upgraded libjpeg-turbo to 3.2.0, so I was able to remove this.

gounthar:
Nice, thanks. I checked the run that went with it: `manylinux_2_39 riscv64` passed in 3h29m on eb51e9a5, which is the head with the flag already gone, so that removal is confirmed green rather than just expected to work.

I went looking for the same kind of escape for the libaom one and could not find it. Every link in that chain is already at its newest version:

- Pillow pins libavif 1.4.2, which is the latest release (2026-05-26)
- libavif 1.4.2 pins libaom v3.14.1 through `AVIF_AOM_GIT_TAG`
- v3.14.1 is the newest libaom tag

So there is nothing to bump. Dropping `-DAOM_TARGET_CPU=generic` needs the fix to land in libaom first, then a libaom release, then a libavif release that moves the tag, and only then a bump here.

One caveat on my own claim: I have not re-tested whether v3.14.1 actually still trips GCC 14. It is possible this was fixed quietly and the flag is now dead weight. I can run the build without it to find out, but it is a ~3.5 hour job, so I would rather only spend that if it is useful to you.

For what it is worth the flag only costs riscv64 the aom SIMD paths, and as the comment notes, a QEMU build would not benefit from those anyway.

radarhere:
See https://github.com/python-pillow/Pillow/issues/9462#issuecomment-5699619723

PR #9468, opened 2026-03-16
Do not use palette from grayscale or bilevel colorspace when reading JPEG2000 images

### LINKED ISSUE #9455: Image.open() fails on a JPEG2000 image
Running:

```python
Image.open('image.jp2').show()
```
on [image.jp2.zip](https://github.com/user-attachments/files/25782299/image.jp2.zip) fails:

```
OSError: broken data stream when reading image file
```

My set-up:

MacOS 10.14.6, Python 3.10.18
pillow-12.1.1-cp310-cp310-macosx_10_10_x86_64.whl

radarhere:
Resolves #9455

Allows the first image from #9455 to be read correctly.

If a JPEG2000 image has a ColorSpace of [Bi-level (0 or 15) or Grayscale (17)](https://exiftool.org/TagNames/Jpeg2000.html#ColorSpec), then leave it as an L mode image, rather than using the palette.


hugovk:
This now has a conflict.

radarhere:
The conflict is resolved.

PR #9471, opened 2026-03-18
Support opening and saving L mode AVIF images with libavif >= 1.3.0

### LINKED ISSUE #9465: Support monochrome AVIF saving and loading
### What did you do?
I have a bunch of gray-scale datasets what I would like to compress using the AVIF format to reduce the amount of data sent. Currently I'm able to store my grayscale data in an RGB AVIF file, but I want to use the grayscale compression ability of AVIF to hopefully get an even better compression ratio and not have to discard a bunch of extra data when loading the data file.

For the AVIF encoder/decoder to support gray-scale images.

### What actually happened?
The AVIF encoder/decoder only supports RGB images.

* OS: Windows 10
* Python: 3.12.10
* Pillow: 12.1.1

```py
im = PIL.Image.fromarray(img_data.astype(np.uint8)) # shape (nx, ny, 1)
im.save("image.avif")
# This saves an RGB avif file instead of a grayscale file.
```

radarhere:
Resolves #9465

If an AVIF image is opened with YUV 4:0:0 subsampling, then it is grayscale, so we can use `AVIF_RGB_FORMAT_GRAY`, instead of  `AVIF_RGB_FORMAT_RGB`, to load it in L mode rather than RGB. Similarly, if the frame being saved is grayscale, using `AVIF_RGB_FORMAT_GRAY` simplifies encoding [within libavif.](https://github.com/AOMediaCodec/libavif/blob/a1153dbccbdf005ce5b6632cdb2cf8be3c28d5ce/src/reformat.c#L248-L252) However, `AVIF_RGB_FORMAT_GRAY` only become available in libavif 1.3.0, so this feature is only present for that version. As a side effect, this also means that if an RGB image is saved with 4:0:0 subsampling, it will be in L mode after reopening.

Also, if all frames being saved to an AVIF image are grayscale, then if we default to 4:0:0 subsampling (instead of the current default of 4:2:0), then together, an L mode image can be saved and opened again in the same mode.


src/PIL/AvifImagePlugin.py
```diff
@@ -153,17 +153,23 @@ def _save(
...
+    grayscale = True
+    grayscale_modes = {"1", "L", "I", "I;16", "I;16L", "I;16B", "I;16N", "F"}
     for ims in [im] + append_images:
-        total += getattr(ims, "n_frames", 1)
+        for frame in ImageSequence.Iterator(ims):
+            if frame.mode not in grayscale_modes:
+                grayscale = False
+                break
+        if not grayscale:
+            break
```
hugovk: Not sure if this is an improvement? Feel free to ignore!

```suggestion
    grayscale_modes = {"1", "L", "I", "I;16", "I;16L", "I;16B", "I;16N", "F"}
    grayscale = all(
        frame.mode in grayscale_modes
        for ims in [im] + append_images
        for frame in ImageSequence.Iterator(ims)
    )
```

hugovk (approved):

src/PIL/AvifImagePlugin.py
radarhere: I think it is. GitHub isn't letting me apply your suggestion for some reason, so I've pushed a commit.

radarhere:
It turns out that Valgrind is raising an error - https://github.com/python-pillow/Pillow/actions/runs/23566085263/job/68628338690?pr=9471#step:5:6558

I've created https://github.com/AOMediaCodec/libavif/pull/3139 to address it.

radarhere:
My PR was merged, but [their analysis](https://github.com/AOMediaCodec/libavif/pull/3139#issuecomment-4136331872) of the problem is that while the uninitialized value is copied to a local variable, it isn't actually used.

So I suppose there's a choice here.
a) wait for a new release of libavif
b) wait for a new release of libavif, and restrict this feature to that new version or newer
c) avoid waiting and just silence the valgrind error on this test

radarhere:
I'm inclined towards a) wait for a new release of libavif

hugovk:
a) sounds good, we're not in a rush for this.

radarhere:
libavif has now been updated to 1.4.2 - https://github.com/python-pillow/Pillow/pull/9652

PR #9474, opened 2026-03-19
Fix integer overflow in quantize_pngquant()

myagmartseren:
### Summary

`quantize_pngquant()` allocates memory using `width * height` (both `unsigned int`) without checking for integer overflow. When the product exceeds `UINT_MAX`, it wraps to a small value, causing an undersized `malloc()`. Subsequent `liq_write_remapped_image_rows()` writes the real size into the small buffer → **heap buffer overflow**.

### Affected Code

`src/libImaging/QuantPngQuant.c`, lines 80, 96:
```c
charMatrix = malloc(width * height);                          // overflows
*quantizedPixels = malloc(sizeof(uint32_t) * width * height); // overflows
```

### Example
- `width=65537, height=65537` → product wraps to 131,073 (should be 4.3 billion)
- `malloc(131073)` allocates 128 KB; decoder writes 4 GB → **heap corruption**

### Fix

- Add overflow check: `if (height != 0 && (size_t)width > SIZE_MAX / (size_t)height)`
- Use `size_t total_pixels` for all size calculations
- Also replaces `sprintf` → `snprintf` (line 129), consistent with CVE-2024-28219

### Classification
- **CWE-190** (Integer Overflow) → **CWE-122** (Heap Buffer Overflow)
- **CVSS 3.1:** 8.1 (High) — `AV:N/AC:H/PR:N/UI:N/S:U/C:H/I:H/A:H`
- **Prerequisite:** libimagequant support compiled in

### DIFF
--- a/src/libImaging/QuantPngQuant.c
+++ b/src/libImaging/QuantPngQuant.c
@@ -8,6 +8,7 @@
  *
  */
 
+#include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -38,6 +39,13 @@ quantize_pngquant(
     *palette = NULL;
     *paletteLength = 0;
     *quantizedPixels = NULL;
+
+    /* Check for integer overflow in width * height to prevent
+     * undersized allocations leading to heap buffer overflow. */
+    if (height != 0 && (size_t)width > SIZE_MAX / (size_t)height) {
+        goto err;
+    }
+    size_t total_pixels = (size_t)width * (size_t)height;
 
     /* configure pngquant */
     attr = liq_attr_create();
@@ -77,7 +85,7 @@ quantize_pngquant(
     }
 
     /* write output pixels (pngquant uses char array) */
-    charMatrix = malloc(width * height);
+    charMatrix = malloc(total_pixels);
     if (!charMatrix) {
         goto err;
     }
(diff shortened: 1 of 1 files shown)


EricSoroos:
Note: Security policy: https://github.com/python-pillow/Pillow/blob/main/.github/SECURITY.md

This function is only called in one location: 
https://github.com/python-pillow/Pillow/blob/3a44ba1c75d72e7f375be9c6b7c71e2aef5e35e7/src/libImaging/Quant.c#L1814
```c
#ifdef HAVE_LIBIMAGEQUANT
            result = quantize_pngquant(
                p,
                im->xsize,
                im->ysize,
                colors,
                &palette,
                &paletteLength,
                &newData,
                withAlpha
            );
```

And previously in that function: (https://github.com/python-pillow/Pillow/blob/3a44ba1c75d72e7f375be9c6b7c71e2aef5e35e7/src/libImaging/Quant.c#L1697C5-L1699C6)

```c
if (im->xsize > INT_MAX / im->ysize) {
        return ImagingError_MemoryError();
    }
```

radarhere:
And one level up from that, we also check that the height is not zero.

https://github.com/python-pillow/Pillow/blob/3a44ba1c75d72e7f375be9c6b7c71e2aef5e35e7/src/_imaging.c#L1733-L1737

src/libImaging/QuantPngQuant.c
```diff
@@ -126,7 +134,14 @@ const char *
...
     int number = liq_version();
-    sprintf(version, "%d.%d.%d", number / 10000, (number / 100) % 100, number % 100);
+    snprintf(
+        version,
+        sizeof(version),
+        "%d.%d.%d",
+        number / 10000,
+        (number / 100) % 100,
+        number % 100
+    );
```
radarhere: This change is included in #9476. See that issue for comments.

radarhere: I've pushed a commit to remove the change from this, leaving it for the other PR.

src/libImaging/QuantPngQuant.c
```diff
@@ -8,6 +8,7 @@
  *
  */
 
+#include <stdint.h>
```
radarhere:
```suggestion
```
This should already be included by https://github.com/python-pillow/Pillow/blob/b04c9a3d2f09ee16e0550ed4a89729922c9f818b/src/libImaging/QuantTypes.h#L15-L20

radarhere:
@myagmartseren do you have any thoughts on our comments about why this change doesn't actually affect Pillow as a whole?

PR #9480, opened 2026-03-20
Correct integer overflow in 16-bit resampling

hayatoikoma:
**📝 Description**
This PR fixes a bug in `Resample.c` where downsampling 16-bit images (`I;16`) using filters with negative lobes (such as `Image.Resampling.LANCZOS`) could result in byte corruption. 

Because Lanczos weighting can create overshoots (ringing artifacts) near sharp edges, the accumulated floating-point sum can sometimes exceed the 16-bit maximum (65535) or fall below zero. Previously, these out-of-bounds values were not correctly clamped before being cast or packed into the 16-bit output buffer, leading to integer overflow/underflow and corrupted pixels.

This update correctly clamps the accumulated float values to the `[0, 65535]` range for `I;16` images during resampling.

**🛠️ Changes Made**
* **`src/libImaging/Resample.c`**: Added bounding/clamping logic to ensure the accumulated float sum is strictly clamped to `0` and `65535` before writing to the `I;16` output buffer.
* **Unit Tests**: Added `test_resampling_clamp` to verify this behavior. The test constructs an image with a hard step edge (0 to 65535) and applies a 5x Lanczos downsampling, comparing the clamped `I;16` output against a float (`F`) reference image to ensure no byte corruption occurs.


hayatoikoma:
@radarhere Codecov was failing and I added another test. But I couldn't figure out how I can run Codecov to check the result so it might be going to fail again.

Please let me know if there is anything else I should do to merge this PR. 👍🏼

src/libImaging/Resample.c
```diff
@@ -493,8 +493,13 @@ ImagingResampleHorizontal_16bpc(
                       k[x];
             }
             ss_int = ROUND_UP(ss);
-            imOut->image8[yy][xx * 2 + (bigendian ? 1 : 0)] = CLIP8(ss_int % 256);
-            imOut->image8[yy][xx * 2 + (bigendian ? 0 : 1)] = CLIP8(ss_int >> 8);
+            if (ss_int < 0) {
```
wiredfool: I think it would make sense to have this in ImagingUtils.h:

```
#define CLIP16(v) ((v) <= 0 ? 0 : (v) < 65536 ? (v) : 65535)
```

and then 
```
ss_int = CLIP16(ROUND_UP(ss));
```

hayatoikoma: Good idea! I updated.

hayatoikoma:
@wiredfool @radarhere Any additional action items for me? Do I need to squash the commits?

radarhere:
You don't need to squash the commits, no.

Did you use AI to develop this?

hayatoikoma:
Yes, I used AI.

radarhere (approved):

hayatoikoma:
Awesome! Thank you for merging it!

PR #9481, opened 2026-03-20
Catch OSError for FreeType >= 2.14.0

radarhere:
Since #8324 was merged, the modified test has started failing in main - https://github.com/python-pillow/Pillow/actions/runs/23347044302/job/67915216500#step:13:2864

Testing, I find that FreeType 2.13.3 passes, but for FreeType >= 2.14.0, `FT_New_Face()` returns [`Invalid_File_Format`](https://freetype.org/freetype2/docs/reference/ft2-error_code_values.html), a.k.a "broken file". The test file is a font generated by OSS-Fuzz though, so there is no reason it has to be a valid file.

I've updated the test to specifically catch "broken file" for FreeType >= 2.14.0.

As background, the test was originally added in #6846. Coincidentally, it also had a `with pytest.raises(OSError):`, intending to catch a "Bitmap missing for glyph" error raised by Pillow, but #8324 removed the error itsell.

### DIFF
--- a/Tests/test_font_crash.py
+++ b/Tests/test_font_crash.py
@@ -1,8 +1,10 @@
 from __future__ import annotations
 
+import pytest
+
 from PIL import Image, ImageDraw, ImageFont
 
-from .helper import skip_unless_feature_version
+from .helper import has_feature_version, skip_unless_feature_version
 
 
 class TestFontCrash:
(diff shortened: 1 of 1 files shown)


radarhere:
I am closing this in favour of #9483

PR #9488, opened 2026-03-24
Check all allocs in the Arrow tree

wiredfool:

* handle alloc failure in all cases
* Ensure we're calling release so the refcount on the image is decremented
* Ensure that release array/schema can handle partially allocated children arrays.

Thanks to @devdanzin for the report.


src/libImaging/Arrow.c
```diff
@@ -10,8 +10,8 @@
 
 static void
 ReleaseExportedSchema(struct ArrowSchema *array) {
-    // This should not be called on already released array
-    // assert(array->release != NULL);
+    // TODO here: release and/or deallocate all data directly owned by
+    // the ArrowArray struct, such as the private_data.
```
hugovk, line 14: Is this blocking merge for 12.2.0, or a note for the future?

PR #9490, opened 2026-03-25
Add dash parameter for line, polygon, and rectangle drawing

### LINKED ISSUE #9127: Support for dashed lines?
I need dashed lines, and it's not really on the cards for me to implement it myself outside of Pillow, because it would be much too slow.

I'm [not the only one who's wanted this](https://stackoverflow.com/questions/51908563/dotted-or-dashed-line-with-python-pillow).

Is there any chance this will be added to Pillow?

Krishnachaitanyakc:
## Summary

Adds a `dash` parameter to `ImageDraw.line()`, `ImageDraw.polygon()`, and `ImageDraw.rectangle()` that enables drawing dashed outlines, implemented in the Python layer.

- The dash pattern follows the SVG `stroke-dasharray` specification: a tuple of ints specifying alternating drawn/blank segment lengths (e.g. `(10, 5)` draws 10px, skips 5px, repeats)
- Odd-length patterns are automatically doubled per SVG spec
- The dash pattern is continuous across connected line segments and polygon/rectangle edges
- Empty dash tuples raise `ValueError`
- Fill is drawn normally; only the outline/stroke is dashed
- Fully backward compatible: all existing behavior is preserved when `dash` is not specified

## Reference                                                                                                                                                       
                                                                                                                                                                     
  This implementation follows the SVG `stroke-dasharray` specification:                                                                                              
  https://www.w3.org/TR/SVG2/painting.html#StrokeDashing                                                                                
                                                                                                    
### Example usage

```python

im = Image.new("RGB", (200, 200), "white")
draw = ImageDraw.Draw(im)

# Dashed line
draw.line([(10, 100), (190, 100)], fill="black", width=2, dash=(10, 5))

# Dashed rectangle with fill
draw.rectangle([20, 20, 180, 60], fill="lightyellow", outline="blue", width=2, dash=(15, 5, 5, 5))

# Dashed polygon with fill
draw.polygon([(50, 120), (150, 120), (150, 180), (50, 180)], fill="lightblue", outline="red", width=2, dash=(10, 5))
```

Closes #9127

## Test plan

- [x] Added 10 new tests covering dashed lines, polygons, and rectangles
- [x] Tests cover: basic dash, multi-segment lines, odd-length patterns, empty dash error, polygon with/without fill, rectangle with/without fill
- [x] All 281 existing `test_imagedraw.py` tests continue to pass (backward compatibility verified)
- [x] Reference images generated and included for pixel-exact assertions
- [x] Documentation updated for all three methods with `versionadded:: 12.2.0`


tox.ini
```diff
@@ -18,6 +18,7 @@ commands =
 [testenv:lint]
 skip_install = true
 deps =
+    build!=1.4.1 # pending https://github.com/pypa/build/pull/1003
```
radarhere:
```suggestion
```

1.4.2 has been released, so this is no longer needed. See https://github.com/python-pillow/Pillow/pull/9495

radarhere:
Could you link to the SVG specification that you used as a reference?

Krishnachaitanyakc:
@radarhere updated the description

radarhere:
Did you use AI to create this PR?

Krishnachaitanyakc:
@radarhere  I used AI to plan and implement yes, but did manually verify the changes and tested them

Krishnachaitanyakc:
@radarhere can you please review this?

radarhere:
This is on my list of things to look at. However, as I'm doing this in my spare time in an underpaid capacity, and trying to also address other issues and PRs, I may not get to it immediately.

radarhere:
I've combined the different shape tests, checking the visual output for more operations.

However, when I changed the polygon points to be something other than a rectangle, I found that the dashed line does exactly sit on the edge of the polygon.



I can think of a solution to this when the width is 1 pixel, but more than that could be complicated. Would you be happy with only supporting widths of 1 pixel?

akx (commented):
In imagedraw_dash_polygon.png, the diagonal dashed line appears 1px inside the filled triangle. Why is that?

EDIT: ah, radarhere mentioned this just above.

akx:
I think this could also be implemented in the C drawing code, where it's rather easy to just skip drawing some pixels that would otherwise be drawn, based on a pattern mask?

src/libImaging/Draw.c
```diff
@@ -205,8 +205,43 @@ hline32rgba(Imaging im, int x0, int y0, int x1, int ink, Imaging mask) {
...
+            return 0;
+        }
+        int v = PyLong_AsLongLong(value);
+        if (v == -1 && PyErr_Occurred()) {
+            return 0;
+        }
+        total += v;
+    }
+    return tuple_index % 2 == 0;
+}
```
akx, line 230: Mmmm, I'd maybe not do Python object conversions for every dashed pixel to be drawn or not drawn, but instead convert the tuple into a C struct before passing to `line*`? 

EDIT: it may also be a good idea to look at separate dashed variants of these functions, since a per-pixel check in the hot path is likely to degrade perf for all drawing operations for a somewhat niche feature.

EDIT 2: In fact, doesn't the dash array basically decompose into an array of on/off states + length, so accessing it to figure out whether to draw is then just `[i % state_array_length]`...

src/libImaging/Draw.c
```diff
@@ -205,8 +205,43 @@ hline32rgba(Imaging im, int x0, int y0, int x1, int ink, Imaging mask) {
...
-line8(Imaging im, int x0, int y0, int x1, int y1, int ink) {
+line8(
+    Imaging im,
+    int x0,
+    int y0,
+    int x1,
+    int y1,
+    int ink,
+    PyObject *dash,
+    int *dash_offset
```
akx, line 241: We should document that this `dash_offset` must always be a valid pointer (it's dereferenced unconditionally in the function). (Repeated for all `line` variants.)

radarhere: I've updated the commit so that it is no longer dereferenced in this function. Predictably, it was causing a problem.

src/libImaging/Draw.c
```diff
@@ -229,16 +262,36 @@ line8(Imaging im, int x0, int y0, int x1, int y1, int ink) {
 
     if (dx == 0) {
         /* vertical */
-        for (i = 0; i < dy; i++) {
-            point8(im, x0, y0, ink);
-            y0 += ys;
+        if (dash != NULL) {
```
akx, line 265: It would probably be fine to always use the Bresenham algorithm when dashing and leave the fast paths (dx == 0, dy == 0) alone. Dashed line drawing will be slower anyhow due to the dash pattern check.

The `GEN_LINE()` stuff from https://github.com/python-pillow/Pillow/pull/9772 would help there, I think - that could maybe be merged first? :)

src/libImaging/Draw.c
radarhere: > In fact, doesn't the dash array basically decompose into an array of on/off states + length, so accessing it to figure out whether to draw is then just `[i % state_array_length]`...

Isn't this complicated by the fact that the dash on/off states might have uneven lengths, under the API of this PR?

> instead convert the tuple into a C struct before passing to `line*`?

You're suggesting allocating a struct that contains the length of the dash array, and the values within? Is that manifestly different from what the Python API is does under the hood?

> a per-pixel check in the hot path

It shouldn't be per-pixel anymore, since https://github.com/python-pillow/Pillow/pull/9490/changes/1af2a6779e2c408bf7a3fadbfed877c98f15dee7

akx: > You're suggesting allocating a struct that contains the length of the dash array, and the values within? Is that manifestly different from what the Python API is does under the hood?

Yep, and I think it is. Right now each call to `should_draw_dash` involves, at a quick look, (not including the trivial operations like field retrievals that would need to be done with the struct too):

* Type/subtype check (`PyTuple_Check`) in `PyTuple_GET_SIZE`
* Type/subtype check (`PyTuple_Check`) in `PyTuple_GetItem`
* Bounds check in `PyTuple_GetItem`
* Type/subtype check `PyLong_Check`
* Type/subtype check (`PyLong_Check`) in `PyLong_AsLongLong`
* `_PyLong_CompactValue` or `_PyLong_AsByteArray` of the value

So it does sound like that's worth doing upfront, at least to something like

```
struct DashArray {
  size_t n_states;
  uint32_t states[];
}
```
in a single pre-pass. (I'm assuming we might not really need `long long` (64-bit) values for states?)

> Isn't [`[i % state_array_length]`] complicated by the fact that the dash on/off states might have uneven lengths, under the API of this PR?

Just "exploding" the on-off states into `char`s, (at the expense of taking up `sum(dash_array)` bytes of memory plus a size_t for the length) is not complicated. (Could also pack 8 states down to a byte, so the expense is `ceil(sum(dash_array) / 8)`, but the indexing becomes fiddlier.) That implementation of course means we'd probably need to bound the maximum length of the dash states, maybe to... 16K entries altogether? 🤷 

But just converting the tuple to something like that `DashArray` above should be a lot more efficient, anyway, so might not be worth it to bother with exploding on top of that.

> It shouldn't be per-pixel anymore, since [1af2a67](https://github.com/python-pillow/Pillow/commit/1af2a6779e2c408bf7a3fadbfed877c98f15dee7)

Ah yeah, checking for `dash`ness at all before the loops is a good idea. Whether to draw the pixel or not has to be done for every pixel though, and that involves the above Python value operations right now.

PR #9504, opened 2026-03-26
Allow None extents in C setimage()

radarhere:
Follow up from #9395

### DIFF
--- a/Tests/test_imagefile.py
+++ b/Tests/test_imagefile.py
@@ -170,6 +170,27 @@ class TestImageFile:
         with pytest.raises(SystemError, match="tile cannot extend outside image"):
             ImageFile._save(im, fp, [ImageFile._Tile("raw", xy + (1, 1), 0, "1")])
 
+    def test_extents_none(self) -> None:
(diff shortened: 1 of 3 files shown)


hugovk:
Could this be considered a breaking change? Although it's fixing unusual and perhaps undocumented behaviour.

Might be worth adding a brief note in the release notes? We could also wait for the Pillow 13.0.0 release in October, but that's probably unnecessary?

radarhere:
#9395 changed the C `setimage()` so extents needed to be passed in. However, I've realised that both before and after that, the C `setimage()` never accepted `None`.

For background, #1938 added the Python `setimage()`, which accepted `None` for the extents, and tested for it accordingly. #8285 added `None` as a possible value to the `extents` type hint in a tile.

I think we should aim for the same behaviour with the C and Python `setimage()`. It is fair enough to say that removing `None` as a possible `extents` value for Python `setimage()` is a breaking change, so I'm now going in the opposite direction with this PR - allowing `None` as an `extents` value in the C `setimage()`.

PR #9342, opened 2025-12-14
Run tests in parallel with pytest-xdist

hugovk:
Follow on from https://github.com/python-pillow/Pillow/pull/8972.

This runs tests by default with pytest-xdist, except disables for valgrind.

Locally, `tox -e py314` goes from 44.79s for pytest to 22.78s.

And we save about an hour on the CI:

| workflow     |        before |        after |        saving |        faster |
| :------------|-------------: |------------: |-------------: |-------------: |
| test         |         46.7m |        28.9m |         17.8m |         1.62x |
| test-docker  |        133.7m |       102.0m |         31.8m |         1.31x |
| test-windows |         19.1m |        13.1m |          6.0m |         1.46x |
| test-mingw   |          1.7m |         1.0m |          0.7m |         1.73x |
| Total        |        201.2m |       145.0m |         56.2m |         1.39x |

This means the test order is no longer deterministic, so we don't need pytest-reverse.

Another option is to add another "extra" or "dependency group" to install pytest-xdist/pytest-sugar for certain invocations (and it'll need some config only to only set the options for these).

I put the imagegrab tests in a single group, which means they'll be allocated the same runner, to avoid some resource clashes on the CI. And I put an AVIF test into its own group because it sometimes failed locally when my machine didn't have much memory to spare.

PR #9516, opened 2026-03-30
Run tests in parallel via tox for 3.5x speedup

hugovk:
https://github.com/python-pillow/Pillow/pull/9342 isn't ready to enable pytest-xdist across the board (CI and local) because of difficulties with leak tests.

But let's at least have the benefits for local testing via tox, if not in CI (which doesn't use tox).

~~Also add the `test-arrow` extra to enable some more tests.~~

pytest times on macOS with `tox -e py314` on cached build:

<details>
<summary>Old table with test-arrow</summary>

<table>
<tr>
 <td>
 <td>No test-arrow
 <td>With test-arrow
<tr>
 <td>Sequential
 <td>41.23s
 <td>42.41s
<tr>
 <td>Parallel
 <td>11.71s
 <td>13.39s
 <tr>
 <td>Speedup
 <td>3.52x 
 <td>3.17x
</table>

</details>

<table>
<tr>
 <td>
 <td>Time
<tr>
 <td>Sequential
 <td>41.23s
<tr>
 <td>Parallel
 <td>11.71s
 <tr>
 <td>Speedup
 <td>3.52x 
</table>

Edit: removed the `test-arrow` extra because it fails on 3.15 due to missing wheel.

### DIFF
--- a/Tests/test_file_spider.py
+++ b/Tests/test_file_spider.py
@@ -15,7 +15,7 @@ TEST_FILE = "Tests/images/hopper.spider"
 
 
 def teardown_module() -> None:
-    del Image.EXTENSION[".spider"]
+    Image.EXTENSION.pop(".spider", None)
 
 
 def test_sanity() -> None:


radarhere:
Why wouldn't the leak failure happen locally?

hugovk:
That's a good question! It'd be easier to investigate if it did happen locally :)

I just tried 10 runs of `tox -e py314` and hit the spider `KeyError` we also had at https://github.com/python-pillow/Pillow/pull/9342#issuecomment-3896889171:

```pytb
    def teardown_module() -> None:
>       del Image.EXTENSION[".spider"]
            ^^^^^^^^^^^^^^^^^^^^^^^^^^
E       KeyError: '.spider'
```

So I've re-added that fix and 10 runs passed, followed by 10 x `tox -e py310` that passed.

PR #9524, opened 2026-03-31
Resize tall images vertically first

radarhere:
I understand the ideal solution is to do this in C, so I've applied this Python solution only to images where height is greater than 100x the width.


radarhere:
I've created #9549 with a C solution.

PR #9549, opened 2026-04-07
If height is scaled down significantly more than width, resize vertically first

radarhere:
Refining #9524, applying a solution in C, rather than Python


akx:
This should probably be rebased so it runs the benchmarks from #9832.

radarhere:
I merged main in, and Benchmark passed, but I don't know how if/how much the improvement is.

akx:
According to [CodSpeed](https://app.codspeed.io/python-pillow/Pillow/branches/radarhere%3Aresize), ~4% faster, so could be measurement noise and as such unreported.

radarhere:
Ok. The benchmark tests don't cover reducing height but not width, so it at least confirms that this doesn't make things worse for other scenarios.

src/libImaging/Resample.c
```diff
@@ -782,82 +782,113 @@ ImagingResampleInner(
...
+            goto end;
         }
+    }
 
-        // Shift bounds for vertical pass
-        for (i = 0; i < ysize; i++) {
-            bounds_vert[i * 2] -= ybox_first;
+#define PASS(function, w, h, offset, ksize, bounds, kk) \
+    second_pass = imTemp != NULL;                       \
+    imTemp = ImagingNewDirty(imIn->mode, w, h);         \
```
hugovk, line 880: Possible memory leak here: if the second pass's `ImagingNewDirty` fails, `imTemp` is overwritten with `NULL` before we `goto end`, so the intermediate image from the first pass (now only referenced by `imIn`) is never deleted. We had an `ImagingDelete(imTemp)` call before.

hugovk (commented):
There's a performance regression for reducing very tall images also with a width reduction:

```console
❯ python -c "
  import timeit
  from PIL import Image
  im = Image.new('RGB', (50, 500000))
  print(min(timeit.repeat(lambda: im.resize((40, 1000)), number=5, repeat=3)) / 5 *
  1000, 'ms')
  "
155.0545250007417 ms
```


```console
❯ python -c "
  import timeit
  from PIL import Image
  im = Image.new('RGB', (50, 500000))
  print(min(timeit.repeat(lambda: im.resize((40, 1000)), number=5, repeat=3)) / 5 *
  1000, 'ms')
  "
212.88325839996105 ms
```

src/libImaging/Resample.c
radarhere: Good catch. I've pushed a commit.

radarhere:
> There's a performance regression for reducing very tall images also with a width reduction

Ok, that should be fixed now.

PR #9550, opened 2026-04-07
Generate CycloneDX SBOM at release time via CI

aclark4life:
Includes all 8 C extension modules, 3 vendored thirdparty libraries (raqm 0.10.3, fribidi-shim, pythoncapi_compat), and 13 optional native library dependencies with encoded dependency relationships.


.github/generate-sbom.py
```diff
@@ -0,0 +1,330 @@
...
+            "name": name,
+            "version": version,
+            "description": desc,
+            "licenses": [{"license": {"id": "MIT-CMU"}}],
+            "purl": f"{purl}#c-ext/{name}",
+        }
+        for name, desc in c_extensions
+    ]
+
+    vendored_components = [
```
jkowalleck, line 109: are these `vendored_components` untouched or are they patched somehow?


If they were modified, it would be good to know in which way -- see https://cyclonedx.org/use-cases/pedigree/

.github/generate-sbom.py
```diff
@@ -0,0 +1,330 @@
...
+            "externalReferences": [
+                {
+                    "type": "vcs",
+                    "url": "https://github.com/python/pythoncapi-compat",
+                },
+            ],
+        },
+    ]
+
+    native_deps = [
```
jkowalleck, line 241: I am missing the licenses of those `native_deps`. 
I see them in https://github.com/python-pillow/Pillow/tree/main/wheels/dependency_licenses 

you may  add the license ID, and even the texts - https://cyclonedx.org/docs/1.7/json/#metadata_tools_oneOf_i0_components_items_licenses_items_oneOf_i0_license_text

jkowalleck, line 241: some of the `native_deps` appear to be optional. better mark them as that.
see CycloneDX' component scope - <https://cyclonedx.org/docs/1.7/json/#metadata_tools_oneOf_i0_components_items_scope>

jkowalleck:
Once everything is flushed out, may I ask you to showcase your SBOMs in the official examples repo? 
see <https://github.com/CycloneDX/bom-examples/tree/master/SBOM>

And when shipping the SBOM in the wheel ala PEP770, maybe give https://github.com/psf/sboms-for-python-packages a quick notice :)

.github/generate-sbom.py
```diff
@@ -0,0 +1,330 @@
...
+            "version": version,
+            "description": desc,
+            "licenses": [{"license": {"id": "MIT-CMU"}}],
+            "purl": f"{purl}#c-ext/{name}",
+        }
+        for name, desc in c_extensions
+    ]
+
+    vendored_components = [
+        {
```
jkowalleck, line 110: maybe add a checksum in them, too?
see https://cyclonedx.org/docs/1.7/json/#metadata_tools_oneOf_i0_components_items_hashes

.github/generate-sbom.py
```diff
@@ -0,0 +1,330 @@
...
+
+    native_deps = [
+        {
+            "bom-ref": "pkg:generic/libjpeg",
+            "type": "library",
+            "name": "libjpeg / libjpeg-turbo",
+            "description": "JPEG codec (required by default; disable with "
+            "-C jpeg=disable). Tested with libjpeg 6b/8/9-9d "
+            "and libjpeg-turbo 8.",
+            "externalReferences": [
```
jkowalleck, line 253: where did you fetch them from? have a `distribution` url? 
see https://cyclonedx.org/docs/1.7/json/#metadata_tools_oneOf_i0_components_items_externalReferences_items_type

.github/generate-sbom.py
```diff
@@ -0,0 +1,330 @@
...
+            "dependsOn": [
+                f"{purl}#thirdparty/fribidi-shim",
+                "pkg:generic/harfbuzz",
+            ],
+        },
+    ]
+
+    return {
+        "bomFormat": "CycloneDX",
+        "specVersion": "1.6",
```
jkowalleck: just go with `"1.7"` - the adoption in ingesting tools should be good today

.github/generate-sbom.py
```diff
@@ -0,0 +1,330 @@
...
+        "serialNumber": f"urn:uuid:{serial}",
+        "version": 1,
+        "metadata": {
```
jkowalleck, line 505: lets add `metadata.lifecycles = [ "build" ]`

see <https://cyclonedx.org/docs/1.7/json/#metadata_lifecycles_items_oneOf_i0_phase>

.github/workflows/wheels.yml
```diff
@@ -276,6 +276,36 @@ jobs:
...
+      - name: Generate CycloneDX SBOM
+        run: python .github/generate-sbom.py
+
+      - name: Upload SBOM as workflow artifact
+        uses: actions/upload-artifact@v7
+        with:
+          name: sbom
+          path: "*.cdx.json"
+
+      - name: Attach SBOM to GitHub release
```
jkowalleck, line 330: better move the job condition to here:
`if: github.repository_owner == 'python-pillow' && github.event_name == 'push' && startsWith(github.ref, 'refs/tags')` 

this way, you would have an SBOM artifact even for tests - which might come in handy at some point. - during review of changes and such ... 
and you would not  have any surprises on releases, surprises that you could have seen in regular CI runs

.github/workflows/wheels.yml
```diff
@@ -276,6 +276,36 @@ jobs:
...
+      - uses: actions/checkout@v6
+        with:
+          persist-credentials: false
+
+      - uses: actions/setup-python@v6
+        with:
+          python-version: "3.x"
+
+      - name: Generate CycloneDX SBOM
+        run: python .github/generate-sbom.py
```
```suggestion
        run: python3 .github/generate-sbom.py
```
Our habit has been to use `python3`. See https://github.com/python-pillow/Pillow/blob/7cf4dac7aef468ad73293c64622872c1e48cd7ef/Makefile#L11-L15 or #6222 for example.

.github/generate-sbom.py
```diff
@@ -0,0 +1,330 @@
+#!/usr/bin/env python3
+"""Generate a CycloneDX 1.6 SBOM for Pillow's C extensions and their
+vendored/optional native library dependencies.
+
+Usage:
+    python .github/generate-sbom.py [output-file]
```
radarhere:
```suggestion
    python3 .github/generate-sbom.py [output-file]
```

.github/workflows/wheels.yml
```diff
@@ -276,6 +276,36 @@ jobs:
...
+      - name: Upload SBOM as workflow artifact
+        uses: actions/upload-artifact@v7
+        with:
+          name: sbom
+          path: "*.cdx.json"
+
+      - name: Attach SBOM to GitHub release
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: gh release upload "${{ github.ref_name }}" *.cdx.json
```
```suggestion
        run: gh release upload "$GITHUB_REF_NAME" *.cdx.json
```
My suggestion to fix the Lint error

.github/generate-sbom.py
```diff
@@ -0,0 +1,330 @@
...
+
+Output defaults to pillow-{version}.cdx.json in the current directory.
+"""
+
+from __future__ import annotations
+
+import json
+import sys
+import uuid
+from datetime import datetime, timezone
```
hugovk: Let's follow https://adamj.eu/tech/2019/09/12/how-i-import-pythons-datetime-module/
```suggestion
import datetime as dt
```

.github/workflows/wheels.yml
```diff
@@ -276,6 +276,36 @@ jobs:
           artifacts_path: dist
           anaconda_nightly_upload_token: ${{ secrets.ANACONDA_ORG_UPLOAD_TOKEN }}
 
+  sbom:
+    if: github.repository_owner == 'python-pillow' && github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
```
hugovk:
```suggestion
    if: |
      github.event.repository.fork == false
      && github.event_name == 'push'
      && startsWith(github.ref, 'refs/tags')
```

.github/workflows/wheels.yml
```diff
@@ -276,6 +276,36 @@ jobs:
...
+    needs: count-dists
```
hugovk: Rather than placing all of this after the `count-dists` bottleneck:

image

Let's build the SBOM beforehand, so we can inspect the results (and debug problems) without having to wait hours for all the wheels to build.

And then a job afterwards to `download-artifact` and attach to the release:

Image

.github/workflows/wheels.yml
jkowalleck: PS: if you have your SBOM artifact available, you might test it for validity and context.

There might be automated validators out there, like https://pypi.org/project/cyclonedx-python-lib/ can do this https://cyclonedx-python-library.readthedocs.io/en/latest/examples.html#complex-validation 

To test the value, you might use visualizers like https://cyclonedx.studio/ 

The practical use of the SBOM would be, besides shipping it in PEP770 style, to feed the SBOM into some sort of tool that analyzes the SBOM and warns about new findings anachronously - like https://dependencytrack.org/ can do. 
This way you would get alerts on shipped versions and might yank a vulnerable ones from package registries, and write the appropriate Security Advisory accordingly.

hugovk:
It would be great to get this eventually into the wheels ([PEP 770](https://peps.python.org/pep-0770/)) but that likely needs extra tooling, and generating and adding to the GH release is a good first step. 👍

.github/workflows/wheels.yml
jkowalleck: probably a very good idea to have the SBOM process isolated.


you are not using target-specific components , or do you? 
Like on the windows build, you use this C library, and in the MacOS, you use a different one. 
I mean, if you were having target-specific components, then you would need target-specific SBOMs generated on bdist's build-time (and additionally put the correct one in the respective wheel - if the bdist process [would support PEP 770](https://github.com/pypa/setuptools/issues/4821)).

hugovk: Yes, there are platform differences, like iOS disabling raqm and imagequant:

https://github.com/python-pillow/Pillow/blob/7cf4dac7aef468ad73293c64622872c1e48cd7ef/pyproject.toml#L108-L110

.github/generate-sbom.py
jkowalleck: > add hashes 

done

.github/workflows/wheels.yml
aclark4life: @jkowalleck Isolated outside of `wheels.yml` you mean? Here's what we have now:

Screenshot 2026-04-09 at 6 31 00 PM

.github/generate-sbom.py
```diff
@@ -0,0 +1,448 @@
...
+    ]
+
+    native_deps = [
+        {
+            "bom-ref": "pkg:generic/libjpeg",
+            "type": "library",
+            "name": "libjpeg / libjpeg-turbo",
+            "description": "JPEG codec (required by default; disable with "
+            "-C jpeg=disable). Tested with libjpeg 6b/8/9-9d "
+            "and libjpeg-turbo 8.",
```
radarhere:
```suggestion
            "and libjpeg-turbo 2-3.",
```
See #4708 and #9507

.github/generate-sbom.py
```diff
@@ -0,0 +1,448 @@
...
+            "PIL._imaging",
+            "Core image processing extension "
+            "(decode, encode, map, display, outline, path, libImaging)",
+        ),
+        ("PIL._imagingft", "FreeType font rendering extension"),
+        ("PIL._imagingcms", "LittleCMS2 colour management extension"),
+        ("PIL._webp", "WebP image format extension"),
+        ("PIL._avif", "AVIF image format extension"),
+        ("PIL._imagingtk", "Tk/Tcl display extension"),
+        ("PIL._imagingmath", "Image math operations extension (via pybind11)"),
```
radarhere:
```suggestion
        ("PIL._imagingmath", "Image math operations extension"),
```
The extension doesn't use pybind11 more than the other extensions do.

.github/generate-sbom.py
```diff
@@ -0,0 +1,448 @@
...
+            "licenses": [{"license": {"id": "MIT"}}],
+            "purl": "pkg:github/HOST-Oman/libraqm@0.10.5",
+            "hashes": [
+                {
+                    "alg": "SHA-256",
+                    "content": sha256_file(thirdparty / "raqm" / "raqm.c"),
+                }
+            ],
+            "pedigree": {
+                "notes": "Vendored unmodified from upstream HOST-Oman/libraqm v0.10.5."
```
radarhere: It's not completely unmodified.

See
https://github.com/HOST-Oman/libraqm/blob/main/src/raqm-version.h.in#L34
vs
https://github.com/python-pillow/Pillow/blob/main/src/thirdparty/raqm/raqm-version.h#L34
(and also note that the filename is changed)
and
https://github.com/HOST-Oman/libraqm/blob/main/src/raqm.c#L39
vs
https://github.com/python-pillow/Pillow/blob/main/src/thirdparty/raqm/raqm.c#L39

```diff
diff --git a/src/thirdparty/raqm/raqm-version.h b/src/thirdparty/raqm/raqm-version.h
index 93b20d23f..9908f06bd 100644
--- a/src/thirdparty/raqm/raqm-version.h
+++ b/src/thirdparty/raqm/raqm-version.h
@@ -31,14 +31,14 @@
 #ifndef _RAQM_VERSION_H_
 #define _RAQM_VERSION_H_
 
-#define RAQM_VERSION_MAJOR @RAQM_VERSION_MAJOR@
-#define RAQM_VERSION_MINOR @RAQM_VERSION_MINOR@
-#define RAQM_VERSION_MICRO @RAQM_VERSION_MICRO@
+#define RAQM_VERSION_MAJOR 0
+#define RAQM_VERSION_MINOR 10
+#define RAQM_VERSION_MICRO 5
 
-#define RAQM_VERSION_STRING "@RAQM_VERSION@"
+#define RAQM_VERSION_STRING "0.10.5"
 
 #define RAQM_VERSION_ATLEAST(major,minor,micro) \
-       ((major)*10000+(minor)*100+(micro) <= \
-        RAQM_VERSION_MAJOR*10000+RAQM_VERSION_MINOR*100+RAQM_VERSION_MICRO)
+    ((major)*10000+(minor)*100+(micro) <= \
+     RAQM_VERSION_MAJOR*10000+RAQM_VERSION_MINOR*100+RAQM_VERSION_MICRO)
 
 #endif /* _RAQM_VERSION_H_ */
diff --git a/src/thirdparty/raqm/raqm.c b/src/thirdparty/raqm/raqm.c
index 5b2690978..88bbbfd5e 100644
--- a/src/thirdparty/raqm/raqm.c
+++ b/src/thirdparty/raqm/raqm.c
@@ -36,7 +36,11 @@
 #include <SheenBidi.h>
 #endif
 #else
+#ifdef HAVE_FRIBIDI_SYSTEM
 #include <fribidi.h>
+#else
+#include "../fribidi-shim/fribidi.h"
+#endif
 #endif
 
 #include <hb.h>
```

jkowalleck: > It's not completely unmodified.

modifications can be tracked in the SBOM: https://cyclonedx.org/use-cases/pedigree/

.github/generate-sbom.py
jkowalleck: > add phase

done

jkowalleck (changes requested):

.github/generate-sbom.py
```diff
@@ -0,0 +1,519 @@
...
+    vendored_components = [
+        {
+            "bom-ref": "pkg:github/HOST-Oman/libraqm@0.10.5",
+            "type": "library",
+            "name": "raqm",
+            "version": "0.10.5",
+            "description": "Complex text layout library "
+            "(vendored in src/thirdparty/raqm/)",
+            "licenses": [{"license": {"id": "MIT"}}],
+            "purl": "pkg:github/HOST-Oman/libraqm@0.10.5",
```
jkowalleck: you have a modified version of this component - so the PURL **MUST NOT** be not the one of the original.

I am not certain what should be used here -  maybe ask in https://github.com/package-url/purl-spec/discussions 


When in doubt, just remove the PURL here.

.github/generate-sbom.py
```diff
@@ -0,0 +1,519 @@
...
+                    "Vendored from upstream HOST-Oman/libraqm v0.10.5 with two "
+                    "Pillow-specific modifications: (1) raqm-version.h.in was "
+                    "pre-processed into raqm-version.h with version placeholders "
+                    "replaced by literal values; (2) raqm.c wraps the <fribidi.h> "
+                    "include in an #ifdef HAVE_FRIBIDI_SYSTEM guard so Pillow's "
+                    "bundled fribidi-shim is used when a system FriBiDi is absent."
+                ),
+            },
+            "externalReferences": [
+                {"type": "vcs", "url": "https://github.com/HOST-Oman/libraqm"},
```
jkowalleck: the component is not the original but a vendored one - so the external refs should point to your repo ... i guess - not sure.

.github/generate-sbom.py
```diff
@@ -0,0 +1,519 @@
...
+                    "type": "distribution",
+                    "url": "https://xcb.freedesktop.org/dist/",
+                },
+            ],
+        },
+        {
+            "bom-ref": "pkg:pypi/pybind11",
+            "type": "library",
+            "name": "pybind11",
+            "scope": "optional",
```
jkowalleck: build-time dependencies are not "optional", but "excluded".

If you really wanted to document it, then like so: 

https://cyclonedx.org/docs/1.7/json/#metadata_tools_oneOf_i0_components_items_scope
> "excluded"	- Components that are excluded provide the ability to document component usage for test and other non-runtime purposes. Excluded components are not reachable within a call graph at runtime.

It has some value to document build-time dependencies - for example, to document compilers and such, for after-the-fact analyses. 
This is the category of Manufacturing BOM ([MBOM](https://cyclonedx.org/capabilities/mbom/)) - read the guide: <https://cyclonedx.org/guides/OWASP_CycloneDX-Authoritative-Guide-to-MBOM-en.pdf>

jkowalleck (changes requested):
some things I would change - see below

.github/generate-sbom.py
jkowalleck: done

.github/generate-sbom.py
jkowalleck: done

.github/generate-sbom.py
jkowalleck: done

jkowalleck (approved):

aclark4life:
Thanks @jkowalleck !

sethmlarson (approved):
Really excited to see this, thanks @aclark4life! 💜

aclark4life:
@sethmlarson Likewise thanks! Will leave it to @hugovk or @radarhere to merge when it LGTThem.

.github/workflows/wheels.yml
```diff
@@ -281,6 +283,48 @@ jobs:
...
+        with:
+          persist-credentials: false
+
+      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+        with:
+          python-version: "3.x"
+
+      - name: Generate CycloneDX SBOM
+        run: python3 .github/generate-sbom.py
+
```
hugovk, line 302: Let's also validate. How about like this?

```suggestion
      - name: Generate CycloneDX SBOM
        run: python3 .github/generate-sbom.py

      - name: Validate SBOM
        run: |
          pip install check-jsonschema
          check-jsonschema --schemafile "https://raw.githubusercontent.com/CycloneDX/specification/1.7/schema/bom-1.7.schema.json" *.cdx.json

```

Currently that fails:

```console
$ check-jsonschema --schemafile https://raw.githubusercontent.com/CycloneDX/specification/1.7/schema/bom-1.7.schema.json pillow-*.cdx.json
Schema validation errors were encountered.
  pillow-12.3.0.dev0.cdx.json::$.metadata.tools: [{'type': 'application', 'name': 'generate-sbom.py', 'vendor': 'Pillow'}] is not valid under any of the given schemas
  Underlying errors caused this.

  Best Match:
    $.metadata.tools: [{'type': 'application', 'name': 'generate-sbom.py', 'vendor': 'Pillow'}] is not of type 'object'
  Best Deep Match:
    $.metadata.tools[0]: Additional properties are not allowed ('type' was unexpected)
  pillow-12.3.0.dev0.cdx.json::$.components[8].pedigree.patches[0].type: 'generated' is not one of ['unofficial', 'monkey', 'backport', 'cherry-pick']
```

.github/generate-sbom.py
```diff
@@ -0,0 +1,517 @@
...
+                            {
+                                "type": "distribution",
+                                "url": "https://github.com/HOST-Oman/libraqm/releases/tag/v0.10.5",
+                            }
+                        ],
+                    }
+                ],
+                "patches": [
+                    {
+                        "type": "generated",
```
hugovk: Fix schema validation error:
```
    $.metadata.tools[0]: Additional properties are not allowed ('type' was unexpected)
  pillow-12.3.0.dev0.cdx.json::$.components[8].pedigree.patches[0].type: 'generated' is not one of ['unofficial', 'monkey', 'backport', 'cherry-pick']
```

This matches the other patch on line 141:

```suggestion
                        "type": "unofficial",
```

.github/generate-sbom.py
```diff
@@ -0,0 +1,517 @@
...
+        "metadata": {
+            "timestamp": now,
+            "lifecycles": [{"phase": "build"}],
+            "tools": [
+                {
+                    "type": "application",
+                    "name": "generate-sbom.py",
+                    "vendor": "Pillow",
+                }
+            ],
```
hugovk: Fix validation errors:

```
Schema validation errors were encountered.
  pillow-12.3.0.dev0.cdx.json::$.metadata.tools: [{'type': 'application', 'name': 'generate-sbom.py', 'vendor': 'Pillow'}] is not valid under any of the given schemas
  Underlying errors caused this.

  Best Match:
    $.metadata.tools: [{'type': 'application', 'name': 'generate-sbom.py', 'vendor': 'Pillow'}] is not of type 'object'
  Best Deep Match:
    $.metadata.tools[0]: Additional properties are not allowed ('type' was unexpected)
```

```suggestion
    "tools": {
      "components": [
        {
          "type": "application",
          "name": "generate-sbom.py",
          "group": "Pillow"
        }
      ]
    },
```

.github/generate-sbom.py
```diff
@@ -0,0 +1,517 @@
...
+
+
+if __name__ == "__main__":
+    version = get_version()
+    output = (
+        Path(sys.argv[1]) if len(sys.argv) > 1 else Path(f"pillow-{version}.cdx.json")
+    )
+    sbom = generate(version)
+    output.write_text(json.dumps(sbom, indent=2) + "\n", encoding="utf-8")
+    print(f"Wrote {output} (Pillow {version}, {len(sbom['components'])} components)")
```
hugovk: `main()` to avoid having global variables that get shadowed, and use argparse.
```suggestion
def main() -> None:
    version = get_version()

    parser = argparse.ArgumentParser(
        description=__doc__, formatter_class=argparse.ArgumentDefaultsHelpFormatter
    )
    parser.add_argument(
        "output",
        nargs="?",
        type=Path,
        default=Path(f"pillow-{version}.cdx.json"),
        help="output file",
    )
    args = parser.parse_args()

    sbom = generate(version)
    args.output.write_text(json.dumps(sbom, indent=2) + "\n", encoding="utf-8")
    print(
        f"Wrote {args.output} (Pillow {version}, {len(sbom['components'])} components)"
    )


if __name__ == "__main__":
    main()
```

.github/generate-sbom.py
```diff
@@ -0,0 +1,517 @@
...
+
+Output defaults to pillow-{version}.cdx.json in the current directory.
+"""
+
+from __future__ import annotations
+
+import datetime as dt
+import hashlib
+import json
+import sys
```
hugovk: For argparse
```suggestion
"""

from __future__ import annotations

import argparse
import datetime as dt
import hashlib
import json
```

.github/generate-sbom.py
```diff
@@ -0,0 +1,517 @@
...
+                {
+                    "type": "distribution",
+                    "url": "https://github.com/fribidi/fribidi/releases",
+                },
+            ],
+        },
+        {
+            "bom-ref": "pkg:generic/libimagequant",
+            "type": "library",
+            "name": "libimagequant",
```
hugovk, line 406: Should libimagequant be in a `dependsOn` list?

.github/generate-sbom.py
```diff
@@ -0,0 +1,517 @@
...
+                {
+                    "type": "distribution",
+                    "url": "https://github.com/ImageOptim/libimagequant/releases",
+                },
+            ],
+        },
+        {
+            "bom-ref": "pkg:generic/libxcb",
+            "type": "library",
+            "name": "libxcb",
```
hugovk, line 422: Should libxcb be in a `dependsOn` list?

.github/generate-sbom.py
```diff
@@ -0,0 +1,517 @@
...
+                {"type": "website", "url": "https://pybind11.readthedocs.io"},
+                {
+                    "type": "distribution",
+                    "url": "https://github.com/pybind/pybind11/releases",
+                },
+            ],
+        },
+    ]
+
+    dependencies = [
```
hugovk, line 452: Should `PIL._imagingtk` have a `dependsOn` entry?

.github/workflows/wheels.yml
```diff
@@ -281,6 +283,53 @@ jobs:
...
+      - name: Validate SBOM
+        run: |
+          pip install check-jsonschema
+          check-jsonschema --schemafile "https://raw.githubusercontent.com/CycloneDX/specification/1.7/schema/bom-1.7.schema.json" *.cdx.json
```
hugovk: Please move this after the upload.

* If the validate fails, we still have an uploaded copy we can inspect.
* We upload the file before installing a tool on this critical deploy workflow.

Related to the last point, shall we pin check-jsonschema? and put it in a requirements file like for cibw (maybe the same file?), so it's updated by Renovate, with a cooldown?

.ci/requirements-sbom.txt
hugovk, line 1: Should we re-use requirements-cibw.txt? If not, this file should also trigger the workflow.

.github/workflows/wheels.yml
```diff
@@ -281,6 +285,53 @@ jobs:
...
+      - name: Generate CycloneDX SBOM
+        run: python3 .github/generate-sbom.py
+
+      - name: Upload SBOM as workflow artifact
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: sbom
+          path: "*.cdx.json"
+
+      - name: Validate SBOM
```
jkowalleck, line 309: marker for referencing/linking here

hugovk: @jkowalleck I don't follow, what do you mean?

hugovk:
I think we're ready! I'll review again a bit later today and hopefully can merge. 

@radarhere Any comments or suggestions?

.github/workflows/wheels.yml
jkowalleck: i wanted to link this particular line of the PR to somebody. 
The easiest way i know was to add a comment and use its direct URL: https://github.com/python-pillow/Pillow/pull/9550/changes#r3129878921

hugovk: Okay!

Also possible like https://github.com/python-pillow/Pillow/pull/9550/changes#diff-75627befeafda60a526c116b32f6fdef9bf57bcf456f840d7592c1d6f13facb4R309 by clicking the line number in the diff.

.github/generate-sbom.py
```diff
@@ -0,0 +1,540 @@
...
+                            "text": {
+                                # raqm.c: wrap the <fribidi.h> include in an
+                                # #ifdef HAVE_FRIBIDI_SYSTEM guard so that when
+                                # building without a system FriBiDi Pillow's own
+                                # fribidi-shim is used instead.
+                                "content": (
+                                    "LS0tIGEvc3JjL3JhcW0uYworKysgYi9zcmMvcmFxbS5jCkBAIC0zNiw3ICszNiwxMSBAQAogI2lu"
+                                    "Y2x1ZGUgPFNoZWVuQmlkaS5oPgogI2VuZGlmCiAjZWxzZQorI2lmZGVmIEhBVkVfRlJJQklESV9T"
+                                    "WVNURU0KICNpbmNsdWRlIDxmcmliaWRpLmg+CisjZWxzZQorI2luY2x1ZGUgIi4uL2ZyaWJpZGkt"
+                                    "c2hpbS9mcmliaWRpLmgiCisjZW5kaWYKICNlbmRpZgogCiAjaW5jbHVkZSA8aGIuaD4K"
```
radarhere: Is this hash something that we could generate at runtime, rather than needing to manually update it for new raqm releases?

.github/generate-sbom.py
```diff
@@ -0,0 +1,540 @@
...
+                },
+            ],
+        },
+        {
+            "bom-ref": "pkg:generic/libimagequant",
+            "type": "library",
+            "name": "libimagequant",
+            "scope": "optional",
+            "description": "Improved colour quantization (optional). "
+            "Tested with 2.6-4.4.1. NOTE: GPLv3 licensed.",
```
radarhere:
```suggestion
            "Tested with 2.6-4.4.1.",
```
This seems redundant, given the next line is "licenses"

.github/generate-sbom.py
```diff
@@ -0,0 +1,540 @@
...
+                },
+            ],
+        },
+        {
+            "bom-ref": "pkg:generic/libxcb",
+            "type": "library",
+            "name": "libxcb",
+            "scope": "optional",
+            "description": "X11 screen-grab support (optional, "
+            "used by PIL._imagingtk on Linux).",
```
radarhere:
```suggestion
            "used by PIL._imaging on macOS and Linux).",
```
It is referenced from [display.c](https://github.com/python-pillow/Pillow/blob/4e0aeba4af78c705e921463a30c00d339d836ebb/src/display.c#L840-L843), not _imagingtk.

We distribute libxcb in [our macOS wheels as well.](https://github.com/python-pillow/Pillow/blob/4e0aeba4af78c705e921463a30c00d339d836ebb/.github/workflows/wheels-dependencies.sh#L269)

.github/generate-sbom.py
```diff
@@ -0,0 +1,540 @@
...
+                },
+            ],
+        },
+        {
+            "bom-ref": "pkg:pypi/pybind11",
+            "type": "library",
+            "name": "pybind11",
+            "scope": "excluded",
+            "description": "C++/Python binding library "
+            "(build-time dependency for PIL._imagingmath).",
```
radarhere:
```suggestion
            "(build-time dependency).",
```
pybind11 doesn't specifically relate to _imagingmath.

.github/generate-sbom.py
```diff
@@ -0,0 +1,540 @@
...
+            "ref": f"{purl}#c-ext/PIL._avif",
+            "dependsOn": ["pkg:generic/libavif"],
+        },
+        {
+            "ref": f"{purl}#c-ext/PIL._imagingmath",
+            "dependsOn": ["pkg:pypi/pybind11"],
+        },
+        {
+            "ref": f"{purl}#c-ext/PIL._imagingtk",
+            "dependsOn": ["pkg:generic/libxcb"],
```
radarhere: XCB is referenced from [display.c](https://github.com/python-pillow/Pillow/blob/4e0aeba4af78c705e921463a30c00d339d836ebb/src/display.c#L840-L843), not _imagingtk, so this should move.

.github/generate-sbom.py
```diff
@@ -0,0 +1,540 @@
...
+            "ref": f"{purl}#c-ext/PIL._webp",
+            "dependsOn": ["pkg:generic/libwebp"],
+        },
+        {
+            "ref": f"{purl}#c-ext/PIL._avif",
+            "dependsOn": ["pkg:generic/libavif"],
+        },
+        {
+            "ref": f"{purl}#c-ext/PIL._imagingmath",
+            "dependsOn": ["pkg:pypi/pybind11"],
```
radarhere: Again, pybind11 doesn't specifically relate to _imagingmath.

.github/generate-sbom.py
```diff
@@ -0,0 +1,539 @@
...
+                    "url": "https://xcb.freedesktop.org/dist/",
+                },
+            ],
+        },
+        {
+            "bom-ref": "pkg:pypi/pybind11",
+            "type": "library",
+            "name": "pybind11",
+            "scope": "excluded",
+            "description": "C++/Python binding library " "(build-time dependency).",
```
radarhere:
```suggestion
            "description": "C++/Python binding library (build-time dependency).",
```

.github/generate-sbom.py
```diff
@@ -0,0 +1,548 @@
...
+    return hashlib.sha256(path.read_bytes()).hexdigest()
+
+
+def upstream_diff_b64(
+    upstream_url: str,
+    upstream_display: str,
+    local_path: Path,
+    local_display: str,
+) -> str:
+    """Fetch an upstream file and return a base64-encoded unified diff vs the local copy."""
```
radarhere:
```suggestion
    """
    Fetch an upstream file and return a base64-encoded unified diff vs the local copy.
    """
```
Lint fix

.github/workflows/wheels.yml
```diff
@@ -281,6 +285,53 @@ jobs:
...
+        run: |
+          pip install -r .ci/requirements-sbom.txt
```
```suggestion
          python3 -m pip install -r .ci/requirements-sbom.txt
```

.github/generate-sbom.py
```diff
@@ -0,0 +1,550 @@
...
+        "version": 1,
+        "metadata": {
+            "timestamp": now,
+            "lifecycles": [{"phase": "build"}],
+            "tools": {
+                "components": [
+                    {
+                        "type": "application",
+                        "name": "generate-sbom.py",
+                        "group": "Pillow",
```
radarhere:
```suggestion
                        "group": "pillow",
```
Searching through https://github.com/CycloneDX/bom-examples/tree/master/SBOM, lowercase seems to be the convention.

.github/generate-sbom.py
```diff
@@ -0,0 +1,550 @@
...
+        "name": "Pillow",
+        "version": version,
+        "description": "Python Imaging Library (fork)",
+        "licenses": [{"license": {"id": "MIT-CMU"}}],
+        "purl": purl,
+        "externalReferences": [
+            {"type": "website", "url": "https://python-pillow.github.io"},
+            {"type": "vcs", "url": "https://github.com/python-pillow/Pillow"},
+            {"type": "documentation", "url": "https://pillow.readthedocs.io"},
+        ],
```
radarhere, line 78:
```suggestion
            {"type": "security-contact", "url": "https://github.com/python-pillow/Pillow/security/advisories/new"},
        ],
```

.github/generate-sbom.py
```diff
@@ -0,0 +1,550 @@
...
+            "hashes": [
+                {
+                    "alg": "SHA-256",
+                    "content": sha256_file(thirdparty / "raqm" / "raqm.c"),
+                }
+            ],
+            "pedigree": {
+                "ancestors": [
+                    {
+                        "bom-ref": "pkg:github/HOST-Oman/libraqm@0.10.5#upstream",
```
radarhere, line 127:
```suggestion
                        "bom-ref": "pkg:github/HOST-Oman/libraqm@0.10.5",
```
`#upstream` seems weird? If that's just me though, nevermind

jkowalleck: this is a `bom-ref` - an arbitrary string without any meaning. use whatever you want

.github/generate-sbom.py
hugovk: Or to the more general security policy?

```suggestion
            {"type": "security-contact", "url": "https://github.com/python-pillow/Pillow/security/policy"},
        ],
```

radarhere: Sure, if you prefer.

jkowalleck: https://cyclonedx.org/docs/1.7/json/#metadata_tools_oneOf_i0_components_items_externalReferences_items_type

> security-contact -  Specifies  a way to contact the maintainer, supplier, or provider in the event of a  security incident. Common URIs include links to a disclosure procedure,  a mailto (RFC-2368) that specifies an email address, a tel (RFC-3966)  that specifies a phone number, or dns (RFC-4501) that specifies the  records containing DNS Security TXT.


"https://github.com/python-pillow/Pillow/security/policy" would be link to a "disclosure procedure" - probably the mose useful way here

.github/generate-sbom.py
```diff
@@ -0,0 +1,555 @@
...
+            "description": "JPEG codec (required by default; disable with "
+            "-C jpeg=disable). Tested with libjpeg 6b/8/9-9d "
+            "and libjpeg-turbo 2-3.",
+            "licenses": [
+                {"license": {"id": "IJG"}},
+                {"license": {"id": "BSD-3-Clause"}},
+            ],
+            "externalReferences": [
+                {"type": "website", "url": "https://libjpeg-turbo.org"},
+                {"type": "website", "url": "https://ijg.org"},
```
radarhere:
```suggestion
                {"type": "website", "url": "https://ijg.org"},
                {"type": "website", "url": "https://libjpeg-turbo.org"},
```
Match the order used for "name" and "licenses"

.github/generate-sbom.py
```diff
@@ -0,0 +1,555 @@
...
+        {
+            "bom-ref": "pkg:generic/zlib",
+            "type": "library",
+            "name": "zlib",
+            "description": "Deflate/PNG compression (required by default; "
+            "disable with -C zlib=disable).",
+            "licenses": [{"license": {"id": "Zlib"}}],
+            "externalReferences": [
+                {"type": "website", "url": "https://zlib.net"},
+                {"type": "distribution", "url": "https://zlib.net/"},
```
radarhere:
```suggestion
                {"type": "distribution", "url": "https://zlib.net"},
```
Consistency with previous line

.github/generate-sbom.py
```diff
@@ -0,0 +1,555 @@
...
+                    "url": "https://xcb.freedesktop.org/dist/",
+                },
+            ],
+        },
+        {
+            "bom-ref": "pkg:pypi/pybind11",
+            "type": "library",
+            "name": "pybind11",
+            "scope": "excluded",
+            "description": "C++/Python binding library (build-time dependency).",
```
radarhere:
```suggestion
            "description": "Parallel C compilation library (build-time dependency).",
```

.github/generate-sbom.py
```diff
@@ -0,0 +1,555 @@
...
+            },
+            "externalReferences": [
+                {
+                    "type": "vcs",
+                    "url": "https://github.com/python/pythoncapi-compat",
+                },
+                {
+                    "type": "distribution",
+                    "url": "https://github.com/python/pythoncapi-compat/releases",
+                },
```
radarhere:
```suggestion
```

https://github.com/python/pythoncapi-compat/releases is empty.

.github/generate-sbom.py
```diff
@@ -0,0 +1,555 @@
...
+            "name": "libimagequant",
+            "scope": "optional",
+            "description": "Improved colour quantization (optional). "
+            "Tested with 2.6-4.4.1.",
+            "licenses": [{"license": {"id": "GPL-3.0-only"}}],
+            "externalReferences": [
+                {"type": "website", "url": "https://pngquant.org/lib/"},
+                {
+                    "type": "distribution",
+                    "url": "https://github.com/ImageOptim/libimagequant/releases",
```
radarhere:
```suggestion
                    "url": "https://github.com/ImageOptim/libimagequant/tags",
```
https://github.com/ImageOptim/libimagequant/releases is empty.

.github/generate-sbom.py
```diff
@@ -0,0 +1,551 @@
...
+                {"type": "website", "url": "https://github.com/fribidi/fribidi"},
+            ],
+        },
+        {
+            "bom-ref": "pkg:github/python/pythoncapi-compat",
+            "type": "library",
+            "name": "pythoncapi_compat",
+            "description": "Backport header for new CPython C-API functions "
+            "(vendored in src/thirdparty/pythoncapi_compat.h)",
+            "licenses": [{"license": {"id": "MIT-0"}}],
```
radarhere:
```suggestion
            "licenses": [{"license": {"id": "0BSD"}}],
```
See https://github.com/python/pythoncapi-compat?tab=0BSD-1-ov-file

.github/generate-sbom.py
```diff
@@ -0,0 +1,551 @@
...
+                {"type": "distribution", "url": "https://zlib.net"},
+            ],
+        },
+        {
+            "bom-ref": "pkg:generic/libtiff",
+            "type": "library",
+            "name": "libtiff",
+            "scope": "optional",
+            "description": "TIFF codec (optional). Tested with libtiff 4.0-4.7.1.",
+            "licenses": [{"license": {"id": "HPND"}}],
```
radarhere:
```suggestion
            "licenses": [{"license": {"id": "libtiff"}}],
```
See https://spdx.org/licenses/libtiff.html

.github/generate-sbom.py
```diff
@@ -0,0 +1,551 @@
...
+            ],
+        },
+        {
+            "bom-ref": "pkg:generic/libimagequant",
+            "type": "library",
+            "name": "libimagequant",
+            "scope": "optional",
+            "description": "Improved colour quantization (optional). "
+            "Tested with 2.6-4.4.1.",
+            "licenses": [{"license": {"id": "GPL-3.0-only"}}],
```
radarhere:
```suggestion
            "licenses": [{"license": {"id": "GPL-3.0-or-later"}}],
```
See https://github.com/imageoptim/libimagequant?tab=License-1-ov-file

.github/generate-sbom.py
```diff
@@ -0,0 +1,551 @@
...
+            ],
+        },
+        {
+            "bom-ref": "pkg:generic/libxcb",
+            "type": "library",
+            "name": "libxcb",
+            "scope": "optional",
+            "description": "X11 screen-grab support (optional, "
+            "used by PIL._imaging on macOS and Linux).",
+            "licenses": [{"license": {"id": "MIT"}}],
```
radarhere:
```suggestion
            "licenses": [{"license": {"id": "X11"}}],
```
See the right hand side of https://gitlab.freedesktop.org/xorg/lib/libxcb

radarhere (approved):

.github/generate-sbom.py
```diff
@@ -0,0 +1,551 @@
...
+            "dependsOn": ["pkg:generic/libwebp"],
+        },
+        {
+            "ref": f"{purl}#c-ext/PIL._avif",
+            "dependsOn": ["pkg:generic/libavif"],
+        },
+        {
+            "ref": f"{purl}#c-ext/PIL._imagingtk",
+            "dependsOn": [],
+        },
```
hugovk: Above in `c_extensions` we include:
```
        ("PIL._imagingtk", "Tk/Tcl display extension"),
        ("PIL._imagingmath", "Image math operations extension"),
        ("PIL._imagingmorph", "Image morphology extension"),
```
None of them have any dependencies. But we only list `PIL._imagingtk` in this `dependencies` list, and it has an empty `dependsOn`.

Let's either list the other two here, or more simply, remove this one:

```suggestion
```

hugovk (approved):

hugovk:
Thanks everyone!

aclark4life:
> Once everything is flushed out, may I ask you to showcase your SBOMs in the official examples repo? see https://github.com/CycloneDX/bom-examples/tree/master/SBOM
> 
> And when shipping the SBOM in the wheel ala PEP770, maybe give https://github.com/psf/sboms-for-python-packages a quick notice :)

Should we send a PR to that repo ?

jkowalleck:
> > Once everything is flushed out, may I ask you to showcase your SBOMs in the official examples repo? see https://github.com/CycloneDX/bom-examples/tree/master/SBOM
> > And when shipping the SBOM in the wheel ala PEP770, maybe give https://github.com/psf/sboms-for-python-packages a quick notice :)
> 
> Should we send a PR to that repo ?

for https://github.com/CycloneDX/bom-examples/tree/master/SBOM - yes. :+1: 

for https://github.com/psf/sboms-for-python-packages, i guess opening an issue is enough, to let @sethmlarson know.

PR #9563, opened 2026-04-14
Add python-pillow GitHub Sponsors to FUNDING.yml

aclark4life:

 * Add python-pillow GitHub Sponsors to FUNDING.yml

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,1 +1,2 @@
+github: python-pillow
 tidelift: "pypi/pillow"
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,2 +1,2 @@
 github: python-pillow
-tidelift: "pypi/pillow"
+tidelift: pypi/pillow


.github/FUNDING.yml
```diff
@@ -1 +1,2 @@
+github: python-pillow
 tidelift: "pypi/pillow"
```
radarhere:
```suggestion
tidelift: pypi/pillow
```
Consistency nitpick?

radarhere (approved):

radarhere:
As much as alphabetical order is nice, would we prefer Tidelift sponsorship over GitHub sponsorship? If yes, then perhaps Tidelift should be first.

aclark4life:
> As much as alphabetical order is nice, would we prefer Tidelift sponsorship over GitHub sponsorship? If yes, then perhaps Tidelift should be first.

I don't know if the order indicates priority, let's merge and adjust later if need be.

aclark4life:
> I don't know if the order indicates priority, let's merge and adjust later if need be.

OK so it does appear to control priority in the listing. That said I'm not sure if it matters and we can always adjust later if need be.

PR #9564, opened 2026-04-15
Exclude formats with ! when opening

radarhere:
Currently, to prevent opening a file in a certain format, without limiting your access to any other format, you can
```python
from PIL import Image
Image.init()
Image.open(path, formats=[f for f in Image.ID if f != "EPS"])
```
or
```python
from PIL import Image, EpsImagePlugin
Image.ID.remove("EPS")
Image.open(path)
```

It would be nice if there was a simpler way. This PR suggests
```python
from PIL import Image
Image.open(path, formats=["!EPS"])
```

And yes, I suggest that `Image.open(path, formats=["EPS", "!EPS"])` still does not allow an EPS image to be opened.


src/PIL/Image.py
```diff
@@ -3598,11 +3598,15 @@ def open(
...
     :param mode: The mode.  If given, this argument must be "r".
-    :param formats: A list or tuple of formats to attempt to load the file in.
-       This can be used to restrict the set of formats checked.
-       Pass ``None`` to try all supported formats. You can print the set of
-       available formats by running ``python3 -m PIL`` or using
-       the :py:func:`PIL.features.pilinfo` function.
+    :param formats: A list or tuple of formats to attempt to load the file in, e.g.
+       ("JPEG", "GIF"). This can be used to restrict the set of formats checked.
+
+       To exclude a format, start the format with "!", e.g. ("!EPS", "!PSD").
```
hugovk:
```suggestion
    :param formats: A list or tuple of formats to attempt to load the file in, for example,
       ``("JPEG", "GIF")``. This can be used to restrict the set of formats checked.

       To exclude a format, start the format with "!",for example, ``("!EPS", "!PSD")``.
```

src/PIL/Image.py
```diff
@@ -3682,21 +3690,13 @@ def _open_core(
                 raise
         return None
 
-    im = _open_core(fp, filename, prefix, formats)
-
-    if im is None and formats is ID:
+    if im := _open_core(fp, filename, prefix, ID):
```
hugovk: With `im = Image.open("Tests/images/hopper.png")`, we're entering this loop, even though we've already done the lazy lookup and have the PNG plugin.

This ends up doing both `preinit` and the full init.

I think this?

```suggestion
    if not (im := _open_core(fp, filename, prefix, ID)):
```

src/PIL/Image.py
```diff
@@ -3622,11 +3626,12 @@ def open(
...
-        formats = ID
-    elif not isinstance(formats, (list, tuple)):
-        msg = "formats must be a list or tuple"  # type: ignore[unreachable]
-        raise TypeError(msg)
+    if formats is not None:
+        if not isinstance(formats, (list, tuple)):
+            msg = "formats must be a list or tuple"  # type: ignore[unreachable]
+            raise TypeError(msg)
+        formats = tuple(format.upper() for format in formats)
+        exclude = all(format.startswith("!") for format in formats)
```
hugovk: Perhaps we can precompute some stuff here and avoid some string concat in the loop below.

```suggestion
    excluded = set()
    allowed = set()
    if formats is not None:
        if not isinstance(formats, (list, tuple)):
            msg = "formats must be a list or tuple"  # type: ignore[unreachable]
            raise TypeError(msg)
        for f in formats:
            f = f.upper()
            if f.startswith("!"):
                excluded.add(f.removeprefix("!"))
            else:
                allowed.add(f)
        allowed -= excluded
```

src/PIL/Image.py
```diff
@@ -3657,12 +3662,15 @@ def _open_core(
...
-            i = i.upper()
-            if i not in OPEN:
-                init()
+        for i in check_formats:
+            if formats is not None:
+                if exclude:
+                    if "!" + i in formats:
+                        continue
+                elif i not in formats or "!" + i in formats:
+                    continue
```
hugovk: And:
```suggestion
        if formats is not None:
            if allowed:
                check_formats = [f for f in check_formats if f in allowed]
            else:
                check_formats = [f for f in check_formats if f not in excluded]
        for i in check_formats:
```

hugovk (commented):
Right, so:

```python
Image.open(path)  # tries all, default formats=None
Image.open(path, formats=None)  # tries all
Image.open(path, formats=["JPEG"])  # only tries JPEG
Image.open(path, formats=["!EPS"])  # tries all but EPS
Image.open(path, formats=["EPS", "!EPS"])  # tries all but EPS
Image.open(path, formats=["JPEG", "!EPS"])  # only tries JPEG, !EPS has no effect
```

That last one could be surprising, but it's probably okay.

src/PIL/Image.py
radarhere: I've pushed a variation on this, moving `allowed` and `excluded` within `if formats is not None:`,
and using `f[1:]` instead of `f.removeprefix("!")`. We already know what the prefix is, so slicing should be faster/simpler.

src/PIL/Image.py
radarhere: I've pushed this.

radarhere:
> Image.open(path, formats=["EPS", "!EPS"])  # tries all but EPS

I originally thought this should try none, but it's an edge case that could be debated. I've pushed your suggestions, and gone with trying all except EPS.

Tests/test_image.py
```diff
@@ -128,6 +128,15 @@ def test_open_formats(self) -> None:
                 assert im.mode == "RGB"
                 assert im.size == (128, 128)
 
+    @pytest.mark.parametrize("formats", (("!PNG",), ("PNG", "!PNG"), ("JPEG", "!PNG")))
+    def test_open_formats_exclude(self, formats: tuple[str]) -> None:
```
hugovk:
```suggestion
    def test_open_formats_exclude(self, formats: tuple[str, ...]) -> None:
```

hugovk:
If someone is currently denying all formats with:

```python
from PIL import Image
im = Image.open("Tests/images/hopper.png", formats=[])
print(im)
```

On `main`:

```pytb
  File "/Users/hugo/github/Pillow/1.py", line 2, in <module>
    im = Image.open("Tests/images/hopper.png", formats=[])
  File "/Users/hugo/github/Pillow/src/PIL/Image.py", line 3715, in open
    raise UnidentifiedImageError(msg)
PIL.UnidentifiedImageError: cannot identify image file 'Tests/images/hopper.png'
```

With the PR:
```
<PIL.PngImagePlugin.PngImageFile image mode=RGB size=128x128 at 0x102CF5BE0>
```

This PR updates the security recommendation in `docs/handbook/security.rst` to reflect this behaviour change, but could this be unexpected?

hugovk (commented):
Also, given a PNG file with `.jpg` extension, and only allowing `formats=["JPEG"]`, and counting the `preinit` and `init` calls:


```python
from PIL import Image

calls = []
preinit, init = Image.preinit, Image.init
Image.preinit = lambda: (calls.append("preinit"), preinit())[1]
Image.init    = lambda: (calls.append("init"),    init())[1]

try:
    # cp Tests/images/hopper.png /tmp/hopper_png.jpg
    Image.open("/tmp/hopper_png.jpg", formats=["JPEG"]).load()
except Image.UnidentifiedImageError:
    print("UnidentifiedImageError")

print(calls)
print("plugins loaded:", len(Image.ID), sorted(Image.ID))
```

On `main`, the lazy loading means the JPEG plugin is loaded, the file is rejected, the allowlist is exhausted, and it quickly gives up:

```
UnidentifiedImageError
[]
plugins loaded: 1 ['JPEG']
```

With this PR, the fallback runs anyway, both `preinit` and `init` are called and all the plugins are loaded for nothing:

```
UnidentifiedImageError
['preinit', 'init']
plugins loaded: 43 ['AVIF', 'BLP', 'BMP', 'BUFR', 'CUR', 'DCX', 'DDS', 'DIB', 'EPS', 'FITS', 'FLI', 'FTEX', 'GBR', 'GIF', 'GRIB', 'HDF5', 'ICNS', 'ICO', 'IM', 'IMT', 'IPTC', 'JPEG', 'JPEG2000', 'MCIDAS', 'MPEG', 'MSP', 'PCD', 'PCX', 'PIXAR', 'PNG', 'PPM', 'PSD', 'QOI', 'SGI', 'SPIDER', 'SUN', 'TGA', 'TIFF', 'WEBP', 'WMF', 'XBM', 'XPM', 'XVTHUMB']
```

radarhere:
> but could this be unexpected?

I've pushed a change so that an empty sequence won't try any format.

> With this PR, the fallback runs anyway, both `preinit` and `init` are called and all the plugins are loaded for nothing

Thanks for catching. I've pushed a fix for this.

src/PIL/Image.py
```diff
@@ -3602,11 +3595,17 @@ def open(
...
-       available formats by running ``python3 -m PIL`` or using
-       the :py:func:`PIL.features.pilinfo` function.
+    :param formats: A list or tuple of formats to attempt to load the file in, for
+       example, ``("JPEG", "GIF")``. This can be used to restrict the set of formats
+       checked.
+
+       To exclude a format, start the format with "!", for example,
+       ``("!EPS", "!PSD")``.
+
+       Pass ``None`` to try all supported formats.
```
akx:
```suggestion
       The default is ``None`` to try all supported formats.
```

akx (commented):
One tiny documentation suggestion.

src/PIL/Image.py
radarhere: In my defence, that sentence is there on main as well.

How about any of these?
- Omit, or pass ``None``, to try all supported formats.
- Pass ``None`` to try all supported formats, as per the default behaviour.
or
- All supported formats are attempted by default, or when ``None`` is used.

akx: No need to defend yourself, I'm aware it was on the preimage :) All of these are fine to me, I just wanted the default behavior to be a little more obvious now that the doc.

> Pass `None` to try all supported formats, which is the default behaviour.

is maybe one possible spelling?

radarhere: I've pushed
> Pass ``None`` to try all supported formats. This is the default behaviour.

PR #9559, opened 2026-04-11
Move dependency versions to single JSON and enable Renovate

hugovk:
Right now, we have the versions for our dependencies split between:

* `winbuild/build_prepare.py` - called when building Windows wheels
* `.github/workflows/wheels-dependencies.sh` - called when building all other wheels

And updating them is a lot of manual work:

* https://github.com/python-pillow/Pillow/issues?q=sort%3Aupdated-desc%20state%3Aclosed%20label%3ADependency

By having them managed by Renovate we:

* automate the update PRs
* get them on a schedule rather than many times throughout the month
* have a [cooldown](https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns) applied to protect from supply chain attacks
* avoid duplicating in `build_prepare.py` and `wheels-dependencies.sh`
* avoid triplicating in the SBOM (https://github.com/python-pillow/Pillow/pull/9550) -- the WIP mostly doesn't include dep versions, but it should, and could be read from this file

Here's a demo on my fork, where I downgraded all the deps to see if it would pick them up, and it has: https://github.com/hugovk/Pillow/issues/153

PR #9565, opened 2026-04-15
Reorder renovate.json

radarhere:
Follow up suggestions for #9559

- Rearrange the entries in renovate.json to match the order in dependencies.json
- In build_prepare.py, simplify the code by using the keys from dependencies.json
- In matchStrings, remove the final question mark from `["\"brotli\":\\s*\"(?<currentValue>\\d+[^\"]*?)\""]`, since `*?` can be replaced with only `*`
- Unrelated, change ".github/workflows/wheel*" to ".github/workflows/wheels*", since there's no files that start with wheel but not wheels

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -23,6 +23,26 @@
         {
             "customType": "regex",
             "managerFilePatterns": ["/^\\.github/dependencies\\.json$/"],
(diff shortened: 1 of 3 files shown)


hugovk (approved):
Thanks!

PR #9567, opened 2026-04-17
Pin GitHub Action to digests

radarhere:
Adding "helpers:pinGitHubActionDigestsToSemver" to renovate.json should be all that is needed to update our GitHub Actions from
https://github.com/python-pillow/Pillow/blob/b8933100453ed24352db2dc753880aebea46087f/.github/workflows/cifuzz.yml#L52
to
```
uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
```

See https://docs.renovatebot.com/presets-helpers/#helperspingithubactiondigeststosemver

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,7 +1,8 @@
 {
     "$schema": "https://docs.renovatebot.com/renovate-schema.json",
     "extends": [
-        "config:recommended"
+        "config:recommended",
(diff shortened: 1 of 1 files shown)


radarhere:
Closing. See https://github.com/python-pillow/Pillow/pull/9568

PR #9592, opened 2026-04-28
Added proportional bitmap font format support

endreszabo:
This PR introduces support for YAFF proportional bitmap font support. Not only proportional, but YAFF is the ultimate bitmap font format that supersedes most popular formats. Unlike xkcd#927 this not a random 15th standard; this IS the standard to which all bitmap font should be converted to and used directly with Pillow. It is easily parsable for machines and human-friendly as you can edit the font files directly with your favorite text editor.

The following framebuffer has been created with Pillow, using the C64 font with manually added kerning pairs:

yaff-with-kerning

YAFF features include:

- multiple-font container
- Unicode with multi-codepoint sequences
- Multi-byte character sets
- Overlapping glyphs
- Kerning
- Vertical metrics
- Greyscale fonts
- Colour fonts

Out of these features only the kerning support has been implemented in this commit. In order to avoid introducing dependency on [monobit](https://github.com/robhagemans/monobit/), we wrote our YAFF format parser; with support for kerning.

Details on the YAFF format from [the specs](https://github.com/robhagemans/monobit/blob/master/YAFF.md):

The yaff format has the following design aims:

- Human-friendly. Truly human-readable and human-editable. For example, BDF and XML claim to be human-readable formats, but let's not kid ourselves. Human-friendly means plain text, flat, immediately visualised, easy on the eye, and light and obvious syntax. We should avoid duplication of information, unless it is of obvious use to a human user.
- Able to represent fixed-width and proportional fonts.
- Preserves comments, metadata and metrics. Formats such as BDF contain a wealth of metadata such as names, acknowledgements and style specification, but also font metrics that affect the way the font is displayed. The yaff format should preserve these.
- Able to represent Unicode fonts as well as codepage fonts.


src/PIL/ImageFont.py
```diff
@@ -769,6 +769,159 @@ def getlength(self, text: str | bytes, *args: Any, **kwargs: Any) -> float:
...
 
 
+class YaffImageFont:
+    """Bitmap font loaded from a YAFF format file, with kerning support.
+
+    To load a YAFF font::
+
+        font = ImageFont.yaff("path/to/font.yaff")
+
+    .. versionadded:: 11.3.0
```
radarhere:
```suggestion
    .. versionadded:: 12.3.0
```

src/PIL/ImageFont.py
```diff
@@ -916,6 +1069,23 @@ def freetype(font: StrOrBytesPath | BinaryIO) -> FreeTypeFont:
...
+    This function loads a font from a YAFF format file and returns a font
+    object that supports kerning and proportional glyph widths. For loading
+    TrueType or OpenType fonts, see :py:func:`~PIL.ImageFont.truetype`.
+
+    :param font: A filename or file-like object containing a YAFF font.
+    :return: A font object.
+    :exception OSError: If the file could not be read.
+    :exception SyntaxError: If the file is not a valid YAFF font.
+
+    .. versionadded:: 11.3.0
```
radarhere:
```suggestion
    .. versionadded:: 12.3.0
```

radarhere:
> Unlike xkcd#927 this not a random 15th standard; this IS the standard to which all bitmap font should be converted to and used directly with Pillow

As far as I can see, YAFF resides within [monobit](https://github.com/robhagemans/monobit), which has 244 stars on GitHub, and is [still in 'Beta' on PyPI](https://pypi.org/project/monobit/).

Your claim seems a bit superlative, especially considering that https://github.com/robhagemans/monobit/blob/master/YAFF.md states
> **Does the world need yet another font format?**
> No. No, it doesn't.

I'm not saying that this shouldn't be a part of Pillow, but it would good to assess exactly how much use this feature would receive. Are there any articles you can link to demonstrating widespread acceptance?

What advantage do you see in embedding this support into Pillow directly, rather than, say, using monobit to convert a YAFF font into PCF or BDF, which Pillow already supports?

endreszabo:
> As far as I can see, YAFF resides within [monobit](https://github.com/robhagemans/monobit), which has 244 stars on GitHub, and is [still in 'Beta' on PyPI](https://pypi.org/project/monobit/).
> 
> Your claim seems a bit superlative, especially considering that https://github.com/robhagemans/monobit/blob/master/YAFF.md states
> 
> > **Does the world need yet another font format?**
> > No. No, it doesn't.

Yes, I am very enthusiastic as I spent years with Jimp + AngelCode BMFont to achieve the same (= my goal is to render a proportional bitmap font with kerning). And even with that, BMFont is basically PNG sprites, so you can't 'detach' the font rendering from the sprite images (no `fill=` for instance, you need to have a unique set of PNG files for each font color you'd need).

> I'm not saying that this shouldn't be a part of Pillow, but it would good to assess exactly how much use this feature would receive. Are there any articles you can link to demonstrating widespread acceptance?

I have no evidence for widespread acceptance. This PR will certainly raise awareness of its existence.

> What advantage do you see in embedding this support into Pillow directly, rather than, say, using monobit to convert a YAFF font into PCF or BDF, which Pillow already supports?

Immediate benefit is the proportional (variable width) glyph support (+kerning, which narrows the scope even further in the field of bitmap fonts) that BDF and PCF do not have.

radarhere:
Did you use AI to develop this?

endreszabo:
> Did you use AI to develop this?

Yes, I was using Opus.

src/PIL/YaffFontFile.py
```diff
@@ -0,0 +1,462 @@
...
+    font: StrOrBytesPath | BinaryIO,
+) -> YaffFontData:
+    """Load a YAFF font file.
+
+    :param font: Path to a .yaff file or a file-like object.
+    :return: Parsed font data.
+    :raises SyntaxError: If the file is not a valid YAFF font.
+    """
+    if hasattr(font, "read"):
+        data = font.read()  # type: ignore[union-attr]
```
radarhere:
```suggestion
        data = font.read()
```
You can see from our Lint CI that you don't need the `ignore` - https://github.com/python-pillow/Pillow/actions/runs/25067433774/job/73474096502?pr=9592#step:6:22

src/PIL/ImageFont.py
```diff
@@ -790,6 +790,159 @@ def getlength(self, text: str | bytes, *args: Any, **kwargs: Any) -> float:
         return self.font.getlength(text, *args, **kwargs)
 
 
+class YaffImageFont:
```
radarhere:
```suggestion
class YaffImageFont(BaseImageFont):
```
See #9595

src/PIL/ImageFont.py
```diff
@@ -790,6 +790,159 @@ def getlength(self, text: str | bytes, *args: Any, **kwargs: Any) -> float:
...
+                prev_cp = None
+                continue
+            if prev_cp is not None:
+                total += self._kern(prev_cp, cp)
+            total += self._advance_width(glyph)
+            prev_cp = cp
+        return max(total, 0)
+
+    def getbbox(
+        self, text: str | bytes, *args: Any, **kwargs: Any
```
radarhere:
```suggestion
        self, text: str | bytes | bytearray, *args: Any, **kwargs: Any
```

endreszabo:
thank you @radarhere for your cooperation. Can I contribute more in any way?

hugovk:
I don't think YAFF is quite suitable for including in Pillow right now, I think it needs a lot more adoption first before we commit to maintaining this "forever".

I'd suggest this could be a standalone package on PyPI first, and then when YAFF is more mature, we can reconsider. That will also allow this code to mature and support more YAFF features beyond kerning.

Thanks anyway for the suggestion!

endreszabo:
> I don't think YAFF is quite suitable for including in Pillow right now, I think it needs a lot more adoption first before we commit to maintaining this "forever".
> 
> I'd suggest this could be a standalone package on PyPI first, and then when YAFF is more mature, we can reconsider. That will also allow this code to mature and support more YAFF features beyond kerning.

Thanks for you reply.

Is it theoretically possible to extract this PR into a standalone package on PyPI, which can then be used with the vanilla Pillow codebase? More specifically: is there an interface where I can specify which font rendering engine Pillow should use?

radarhere:
Sure. There's no reason why you couldn't have `YaffImageFont` in your own package, and then pass an instance as the `font` to the Pillow package's `ImageDraw` or `ImageText`.

PR #9593, opened 2026-04-28
SBOM: Use real versions from dependencies.json

hugovk:
Follow on from https://github.com/python-pillow/Pillow/pull/9550 and https://github.com/python-pillow/Pillow/pull/9559.

 * Sort things alphabetically to make easier to find
 * Use versions from dependencies.json
 * Remove historical 'tested on'
 * Git ignore generated SBOM

Will be easier to review commit-by-commit, can also split to separate PRs if desired.

cc @jkowalleck @sethmlarson

### DIFF
--- a/.github/generate-sbom.py
+++ b/.github/generate-sbom.py
@@ -79,18 +79,18 @@ def generate(version: str) -> dict:
     }
 
     c_extensions = [
+        ("PIL._avif", "AVIF image format extension"),
         (
             "PIL._imaging",
             "Core image processing extension "
             "(decode, encode, map, display, outline, path, libImaging)",
         ),
-        ("PIL._imagingft", "FreeType font rendering extension"),
         ("PIL._imagingcms", "LittleCMS2 colour management extension"),
-        ("PIL._webp", "WebP image format extension"),
-        ("PIL._avif", "AVIF image format extension"),
-        ("PIL._imagingtk", "Tk/Tcl display extension"),
+        ("PIL._imagingft", "FreeType font rendering extension"),
         ("PIL._imagingmath", "Image math operations extension"),
         ("PIL._imagingmorph", "Image morphology extension"),
+        ("PIL._imagingtk", "Tk/Tcl display extension"),
+        ("PIL._webp", "WebP image format extension"),
     ]
 
     ext_components = [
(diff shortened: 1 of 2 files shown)


.github/generate-sbom.py
```diff
@@ -107,6 +113,51 @@ def generate(version: str) -> dict:
...
+                {"type": "website", "url": "https://github.com/fribidi/fribidi"},
+            ],
+        },
+        {
+            "bom-ref": "pkg:github/python/pythoncapi-compat",
+            "type": "library",
+            "name": "pythoncapi_compat",
+            "description": "Backport header for new CPython C-API functions "
+            "(vendored in src/thirdparty/pythoncapi_compat.h)",
+            "licenses": [{"license": {"id": "0BSD"}}],
```
sethmlarson, line 144: TIL BSD Zero has an ID of "0BSD".

sethmlarson (approved):

.github/generate-sbom.py
```diff
@@ -191,244 +242,193 @@ def generate(version: str) -> dict:
...
         {
-            "bom-ref": "pkg:generic/libavif",
+            "bom-ref": "pkg:generic/libtiff",
             "type": "library",
-            "name": "libavif",
+            "name": "libtiff",
+            "version": versions["tiff"],
             "scope": "optional",
-            "description": "AVIF codec (optional, used by PIL._avif). "
-            "Requires libavif >= 1.0.0.",
```
radarhere, line 361: No objections, but did you want to mention why you removed 'Requires libavif >= 1.0.0'?

.github/generate-sbom.py
```diff
@@ -447,51 +447,64 @@ def generate(version: str) -> dict:
...
+                {"type": "distribution", "url": "https://zlib.net"},
+            ],
+        },
     ]
 
     dependencies = [
         {
             "ref": purl,
-            "dependsOn": [e["bom-ref"] for e in ext_components],
+            "dependsOn": sorted(e["bom-ref"] for e in ext_components),
```
radarhere: `sorted()` won't do anything, because this PR has already manually sorted the list, right?

radarhere (approved):

.github/generate-sbom.py
hugovk: It's not relevant to an SBOM, which records what was actually used _in this release_.

This is a receipt of what we're delivering, and we report elsewhere what actual libavif version is included.

.github/generate-sbom.py
hugovk: Yep!


```suggestion
            "dependsOn": [e["bom-ref"] for e in ext_components],
```

PR #9598, opened 2026-04-30
Restrict SBOM upload to only Pillow JSON

radarhere:


hugovk:
Are there non-Pillow *.cdx.json files?

radarhere:
No. My thinking was that perhaps being restrictive is better when dealing with a release.

The thought occurred to me when seeing `pillow-*.cdx.json` in `.gitignore` from #9593. You may say though that users can drag anything they want into their local copy of Pillow, but our CI is a far more controlled environment.

PR #9601, opened 2026-05-01
Added bit_depth attribute to PngImageFile

ajslater:
I find it useful when working with PNGs to know their bit depth and since i use Pillow to detect image formats as well as manipulate images I thought it might be a useful addition.

This just reads a well known byte from the png stream and stores it on the PngImageFile.

Tell me if this looks useful and I can modify this PR to work well as a Pillow contribution.


radarhere:
> I find it useful when working with PNGs to know their bit depth

Could you elaborate on this? Pillow attempts to deal with this kind of low level information so you don't have to. You should be able to infer a less-precise bit depth from `im.mode`, or you can look at `im.png.im_rawmode` and figure out the exact value.

The following code would let you determine a numeric value for the bit depth at the moment.
```python
from PIL import Image, PngImagePlugin
with Image.open("Tests/images/hopper.png") as im:
    for bitDepth_colorType, mode_rawmode in PngImagePlugin._MODES.items():
        if im.png.im_rawmode == mode_rawmode[1]:
            print("Bit depth:", bitDepth_colorType[0])
```
I acknowledge my suggestion is inelegant and uses a private property.

If we did go with your suggestion, I think it would be better as another key in `self.info`, rather than a separate property.

> Tell me if this looks useful and I can modify this PR to work well as a Pillow contribution.

I find this comment confusing. What do you think is undone about this PR? Or are you just saying that you're willing to accept feedback?

ajslater:
> I find it useful when working with PNGs to know their bit depth

In my application, called [picopt](https://github.com/ajslater/picopt), I do PNG lossless optimization. The bit depth detection code I have is actually used to disqualify high bit depth pngs from using the PngOut optimize which doesn't accept them. I've had this code rattling around my app for a while now and always wondered if PIllow users might find this extra information useful. 

Idk if it really fits what Pillow wants to accomplish, but I thought I'd run it by you.

I just tested your solution using _MODES and it works in my code.
A "bit_depth" info key would be a little more convenient, of course, for consumers. If that sounds useful to you I could populate that value any which way.

> Or are you just saying that you're willing to accept feedback?

Yes, that. I've never contributed to Pillow before and was unsure if this PR met the standards and practices of this repo.

I suppose the questions are:
1. Would a conveniently accessible bit_depth value be useful for Pillow users or would it just be expanding the info API needlessly?
2. If it is wanted, what method would you prefer to extract the data?
3. If not, would your code snippet be useful to people in the docs? It was interesting to me. I didn't know you could do that.

radarhere:
> unsure if this PR met the standards and practices of this repo.

This PR is essentially valid, yes.

1) This is the part that not sure about. Pillow reads the image and quickly transforms it internally into 1, 8 or 16 bit depth. So the information you want is about how the data was originally represented, which I don't think is relevant to any subsequent Pillow operations.

2) A key in `self.info` would be my suggestion.

3) The fact that it uses a private property means that it isn't something that we'd want to officially support. In practice it's unlikely to change, but in theory, we could change it without a deprecation period. I just offer it in case you wanted an immediate solution, or don't mind living on the edge.

akx:
Yeah, to me this looks like it should be `self.info["bit_depth"]` 👍

radarhere:
Thinking about this again, while the original bit depth might be useful for opened image data, if the image data is adjusted afterwards, that bit depth might no longer be relevant. For example, a 2 bit PNG is opened in L mode. If Pillow places a pixel on the image with a value of 173, then suddenly, it's not a 2 bit image any more.

It would normal for `im.info` to change after `convert()`ing an image to a different mode, but I wouldn't expect it to change after `putpixel()`. I know that the current form of this PR isn't using `im.info`, but the same feeling applies.

So I think this isn't as clean as I'd like for general scenarios, and I'm inclined to close.

akx:
What about `"original_bit_depth"` then, to be very obvious about it?

radarhere:
ajslater would like this information when deciding whether to pass the image to an external optimisation tool, not for any Pillow operation. It seems like a rather specific scenario. Could you explain why you think it is useful beyond that?

akx:
> not for any Pillow operation.

Why is a Pillow operation needed? As it is, the library can be used to identify and read other image metadata (such as EXIF data); the bit-depth of an image seems like such metadata that evidently can be useful for some users.

radarhere:
I think EXIF data is more valuable - it contains information beyond sheer pixels, and can be resaved.

I think the chosen bit depth of the last encoder is a minor detail, and the scenario described here (opening a PNG image, inspecting it, only to save it with a different tool instead of Pillow) seems kind of specific.

If you disagree though, we can leave this open to see what others think.

PR #9602, opened 2026-05-01
Add lossless to WebP info

ajslater:
I find it useful to know if a WebP Image was encoded with the lossless bit set. While the exact amount of true 'losslessness' in encoded images can be a contentious subject, i find knowing the encoder's intent useful.

This sets the is_lossless property to true if *every* frame of the WebP was encoded with the lossless bit set. A simple determination for single frame images, but a restrictive judgement call for animated WebP's which main contain a mix of frame types.

Tell me if this looks useful and I can modify this PR to work well as a Pillow contribution.


ajslater:
There's possibly a general case for an "is_lossless" property on many or all Image formats.
png & gif would always be True. jpeg would always be False. It would be most useful on any other mixed types that require small algorithms like this one to determine.

radarhere:
> I find it useful to know if a WebP Image was encoded with the lossless bit set.

Could you elaborate on this?

If we did go with your suggestion, I think it would be better as another key in `self.info`, rather than a separate property.

src/PIL/WebPImagePlugin.py
```diff
@@ -22,6 +22,40 @@
 }
 
 
+def _is_lossless(data: bytes) -> bool:
+    # A WebP file is considered lossless when every coded frame uses the
+    # VP8L bitstream. See https://developers.google.com/speed/webp/docs/riff_container
+    if len(data) < 16 or data[:4] != b"RIFF" or data[8:12] != b"WEBP":
```
radarhere:
```suggestion
    if not _accept(data):
```

ajslater:
> I find it useful to know if a WebP Image was encoded with the lossless bit set.

My picopt program does lossless optimization. There isn't a way I know of to further compress a lossy WebP image without losing information, so I disqualify images from my program's optimizations by detecting if they were encoded as lossless or not.

Your preference for metadata going into the info dict makes sense. If it looks like we should proceed with this, I'll put the bit there.

radarhere:
> There isn't a way I know of to further compress a lossy WebP image without losing information

It doesn't really matter, but I believe you meant 'lossless' here.

radarhere:
Does it need to consider all frames, or could it just consider the current frame?

radarhere:
Looking at https://developers.google.com/speed/webp/docs/container-api, I don't see a convenient way to determine this information through the official API. That makes me think Google decided this isn't a useful piece of information when reading, and wonder if we shouldn't second guess that.

PR #9618, opened 2026-05-05
Fix multiline text line height calculation

### LINKED ISSUE #1646: Multiline text height not being calculated correctly

mitre88:
The line spacing calculation was using only 'A' which doesn't account for characters that extend below the baseline (like 'p', 'g', 'y'). Using `string.ascii_letters` ensures we measure the full font height including both ascenders and descenders.

Fixes #1646

### DIFF
--- a/src/PIL/ImageText.py
+++ b/src/PIL/ImageText.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import math
 import re
+import string
 from typing import AnyStr, Generic, NamedTuple
 
 from . import ImageFont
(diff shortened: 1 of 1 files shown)


radarhere:
This is very similar to #9581

Please have a read of the comments I have left there, primarily
> There's a lot of debate in https://github.com/python-pillow/Pillow/issues/1646. Are you sure that you have read all of that and concluded that this simple change is all that is needed?

radarhere:
If there's no further comments, then I think this can be closed. This has the same shortcomings as #9581, and there's no need for two PRs where we discuss what to do.

mitre88:
Thanks for the review. I agree this has the same shortcomings as #9581 and needs a more complete design discussion around #1646 rather than a narrow patch here. Closing this PR to avoid duplicating that discussion.

PR #9624, opened 2026-05-12
Add support for Python 3.15

radarhere:
https://github.com/python-pillow/Pillow/blob/3ce681240fa6ad2f19a73b2c5531baada38dee96/setup.py#L60-L71

Since Python 3.15 has now reached [beta](https://www.python.org/downloads/release/python-3150b1/), we would like users to be able to install and use it, so let's increment the version check.


hugovk (approved):
Thanks!

PR #9625, opened 2026-05-12
Update free-threading CI

hugovk:
We dropped 3.13t wheels in https://github.com/python-pillow/Pillow/pull/9535 because the upcoming cibuildwheel will drop them because they were experimental.

Let's also drop testing 3.13t in our CI.

---

Also, we defensively set `PYTHON_GIL=0` to make sure nothing re-enables the GIL. That was when first adding support in 2024, and now the ecosystem has much better free-threading support.

@ngoldbaum suggested removing that, and like NumPy, add a test to fail if something does re-enable it. If something fails, that helps identify something that needs free-threading support, and we can report it (most likely to @ngoldbaum :)

https://github.com/numpy/numpy/blob/3bea241a0fbf5610f876dc54d4caf7380f75471e/numpy/conftest.py#L103

---

We were skipping the pyqt6 install when `PYTHON_GIL=0`, because it doesn't support free-threading yet. Now we're removing that env var, let's only install if a wheel exists.


Tests/conftest.py
```diff
@@ -16,6 +24,27 @@ def pytest_report_header(config: pytest.Config) -> str:
         return f"pytest_report_header failed: {e}"
 
 
+def pytest_terminal_summary(
+    terminalreporter: pytest.TerminalReporter, exitstatus: int, config: pytest.Config
+) -> None:
```
radarhere:
```suggestion
def pytest_terminal_summary(terminalreporter: pytest.TerminalReporter) -> None:
```

Tests/conftest.py
```diff
@@ -16,6 +24,25 @@ def pytest_report_header(config: pytest.Config) -> str:
...
 
+def pytest_terminal_summary(terminalreporter: pytest.TerminalReporter) -> None:
+    if (
+        FREE_THREADED_BUILD
+        and not gil_enabled_at_start
+        and sys._is_gil_enabled()  # type: ignore[attr-defined]
+    ):
+        tr = terminalreporter
+        tr.ensure_newline()
+        tr.section("GIL re-enabled", sep="=", red=True, bold=True)
```
radarhere:
```suggestion
        tr.section("GIL re-enabled", red=True, bold=True)
```
The default for `sep` is already "=" - https://github.com/pytest-dev/pytest/blob/984cabfaccf8aa69fe49097ed3d07bd60d05f240/src/_pytest/terminal.py#L568

But if you prefer to be explicit, then ok.

radarhere (approved):

PR #9626, opened 2026-05-13
Fix typos in test comments

mokashang:
Two small typo fixes in test file comments:

- `mmaped` → `mmapped` in `Tests/test_file_tiff.py` and `Tests/test_image.py` (the past tense of `mmap`).
- `co-ordinates` → `coordinates` in `Tests/test_imagefont.py`, matching the unhyphenated spelling already used throughout `src/PIL/` and the other test modules.

Only comments are touched; no runtime behavior changes.

### DIFF
--- a/Tests/test_file_tiff.py
+++ b/Tests/test_file_tiff.py
@@ -1041,7 +1041,7 @@ class TestFileTiffW32:
     def test_fd_leak(self, tmp_path: Path) -> None:
         tmpfile = tmp_path / "temp.tif"
 
-        # this is an mmaped file.
+        # this is an mmapped file.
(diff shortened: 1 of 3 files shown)


radarhere:
> `mmaped` → `mmapped` ... (the past tense of `mmap`)

What makes you say this? I mean, `mmap` is a name, but I'm not convinced that it has an official past sense. See https://blog.essence.dev/p/are-you-sure-you-want-to-use-mmap for an example of 'mmaped'.

@aclark4life how can you tell that this was AI assisted?

radarhere:
I've created https://github.com/python-pillow/Pillow/pull/9628 for just the 'coordinates' suggestion.

PR #9629, opened 2026-05-13
Don't use list as default in PdfParser read_prev_trailer

danigm:
Backporting [this patch](https://github.com/python-pillow/Pillow/pull/9519) into some openSUSE distro I've detected that it adds an empty list as a default value in a method.

It's not recommended to use the empty list as default value in functions or methods because Python interpreter evaluates during parsing, so it will be the same list for different calls.

https://pylint.pycqa.org/en/latest/user_guide/messages/warning/dangerous-default-value.html

### DIFF
--- a/src/PIL/PdfParser.py
+++ b/src/PIL/PdfParser.py
@@ -692,7 +692,7 @@ class PdfParser:
             self.read_prev_trailer(self.trailer_dict[b"Prev"])
 
     def read_prev_trailer(
-        self, xref_section_offset: int, processed_offsets: list[int] = None
+        self, xref_section_offset: int, processed_offsets: list[int] | None = None
     ) -> None:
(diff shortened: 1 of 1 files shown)


src/PIL/PdfParser.py
```diff
@@ -692,7 +692,7 @@ def read_trailer(self) -> None:
             self.read_prev_trailer(self.trailer_dict[b"Prev"])
 
     def read_prev_trailer(
-        self, xref_section_offset: int, processed_offsets: list[int] = []
+        self, xref_section_offset: int, processed_offsets: list[int] = None
```
radarhere:
```suggestion
        self, xref_section_offset: int, processed_offsets: list[int] | None = None
```

radarhere (approved):

PR #9631, opened 2026-05-15
Switch iOS back to macos-26-intel

radarhere:
Reverts https://github.com/python-pillow/Pillow/pull/9509

Just because it is better to use macos-26-intel than the older macos-15-intel, and it appears to be passing now.


hugovk (approved):
Yes, they made an upstream fix: https://github.com/actions/runner-images/issues/13882#issuecomment-4439181741

PR #9639, opened 2026-05-28
feat: add ImageDraw.textwrap() for automatic text wrapping

### LINKED ISSUE #6201: Feature Request: automatic text wrapping/text box filling

RichardHopperProGrammar:
# Add `ImageDraw.textwrap()` for automatic text wrapping

Fixes [#6201](https://github.com/python-pillow/Pillow/issues/6201)

## Problem

Pillow has `ImageDraw.multiline_text()` but no built-in way to wrap plain text into lines that fit a given pixel width. Users have to implement their own word-wrapping logic using `textlength()`, which is tedious and error-prone.

## Solution

Added `ImageDraw.textwrap()` — a static method that wraps text at whitespace boundaries so no line exceeds a given *max_width*. Returns a tuple of (`lines`, `widths`) that can be passed directly to `multiline_text()`.

\`\`\`python

im = Image.new("RGB", (400, 200), "white")
draw = ImageDraw.Draw(im)

lines, widths = ImageDraw.textwrap(
    "This is a long piece of text that should wrap",
    font=my_font,
    max_width=200,
)

draw.multiline_text((10, 10), "\\n".join(lines), fill="black", font=my_font)
\`\`\`

## Design (per maintainer guidance)
- Breaks at **whitespace only** — individual words longer than max_width are kept intact (they just exceed the limit)
- Uses \`getlength()\` for measurement (no dynamic font resizing)
- Paragraph boundaries (newlines) are preserved
- Accepts same \`font\`, \`direction\`, \`features\`, \`language\` params as \`text\`/ \`multiline_text\`

## Changes
- **ImageDraw.py**: Added \`_textlength()\` helper + \`ImageDraw.textwrap()\` static method (111 lines)
- **Tests/test_imagedraw_textwrap.py**: 5 tests — basic wrap, single-line fit, newline preservation, multiline_text integration, empty input

## Backwards Compatibility
Pure addition. No existing APIs modified. \`textwrap\` is a new static method on \`ImageDraw\`.

### DIFF
--- a/Tests/test_imagedraw_textwrap.py
+++ b/Tests/test_imagedraw_textwrap.py
@@ -1,0 +1,70 @@
+from io import BytesIO
+
+from PIL import Image, ImageDraw, ImageFont
+
+
+def textwrap_basic():
+    """Basic wrapping: words split across lines when exceeding max width."""
+    lines, widths = ImageDraw.textwrap(
+        "hello world foo bar",
+        max_width=100,
+    )
+    assert len(lines) >= 1
+    for w in widths:
+        if w > 0:
+            assert w <= 110  # allow slight rounding tolerance
+
+
+def textwrap_single_line():
+    """Text that fits entirely on one line."""
+    lines, widths = ImageDraw.textwrap("short", max_width=500)
+    assert len(lines) == 1
+    assert lines[0] == "short"
+
+
+def textwrap_newline_preserved():
+    """Paragraphs separated by newlines are preserved."""
+    lines, widths = ImageDraw.textwrap(
+        "line one\nline two", max_width=500
+    )
+    assert "line one" in lines
+    assert "" in lines  # empty line between paragraphs
+
+
+def textwrap_with_multiline_text():
+    """textwrap + multiline_text renders correctly."""
+    im = Image.new("RGB", (400, 200), "white")
+    draw = ImageDraw.Draw(im)
+
+    lines, widths = ImageDraw.textwrap(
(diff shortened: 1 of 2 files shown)


radarhere:
Hi. This feels like you are not aware of `ImageText.Text.wrap()` - https://pillow.readthedocs.io/en/stable/releasenotes/12.2.0.html#imagetext-text-wrap

RichardHopperProGrammar:
Hi @radarhere — you are absolutely right, I missed . That API already does what I was trying to add here. Closing this PR to avoid duplication. Thanks for the quick review!

PR #9645, opened 2026-06-01
Move custom build in `setup.py` to custom backend in `_custom_build`

### LINKED ISSUE #9643: Move custom build in setup.py to custom backend in _custom_build
Do we need to deprecate `setup.py` at some point or can it continue to exist indefinitely? Is there any tangible benefit, security or otherwise to move C extension builds to `pyproject.toml`? ~Can we assume we'd use the `setuptools` backend and~ what's the expected level of effort? I saw #9642 and it made me curious about it.

aclark4life:
Move C extension build logic from `setup.py` into
`_custom_build/pillow_ext.py` and update the custom backend to call `pillow_ext.run()`.

Fixes #9643  .

 * Move the remaining custom build in `setup.py` to our custom backend.


radarhere:
For some background - the custom backend was added in #7171, as a workaround for https://github.com/pypa/setuptools/issues/2491. I had hoped that setuptools would one day resolve their issue, and the custom backend could be removed. But this PR is investing in it further.

What is the advantage of using a custom backend over setup.py? It doesn't appear that [setup.py itself is problematic.](https://packaging.python.org/en/latest/discussions/setup-py-deprecated/#setup-py-deprecated)

aclark4life:
> What is the advantage of using a custom backend over setup.py? It doesn't appear that [setup.py itself is problematic.](https://packaging.python.org/en/latest/discussions/setup-py-deprecated/#setup-py-deprecated)

True, it's only cosmetic. I would prefer to not see `setup.py` anymore because it's confusing, but not urgent.

_custom_build/pillow_ext.py
```diff
@@ -1073,40 +1051,62 @@ def debug_build() -> bool:
...
+
+    if sys.platform == "win32" and sys.version_info >= (3, 16):
+        import atexit
+
+        atexit.register(
+            lambda: warnings.warn(
+                f"Pillow {PILLOW_VERSION} does not support Python "
+                f"{sys.version_info.major}.{sys.version_info.minor} and does not provide "
+                "prebuilt Windows binaries. We do not recommend building from source on "
+                "Windows.",
```
radarhere:
```suggestion
                f"{sys.version_info.major}.{sys.version_info.minor} and does not "
                "provide prebuilt Windows binaries. We do not recommend building from "
                "source on Windows.",
```
Lint fix.

_custom_build/pillow_ext.py
```diff
@@ -57,20 +49,6 @@ def get_version() -> str:
...
-
-    atexit.register(
-        lambda: warnings.warn(
-            f"Pillow {PILLOW_VERSION} does not support Python "
-            f"{sys.version_info.major}.{sys.version_info.minor} and does not provide "
-            "prebuilt Windows binaries. We do not recommend building from source on "
-            "Windows.",
-            RuntimeWarning,
-        )
-    )
```
radarhere, line 71: Is there any particular reason this was moved inside `run()`?

aclark4life: No. I suspect it was the 🤖 's attempt to clean up after the change from executing the setup script to executing `run`, a function within the setup script. Unless it needs to run on import probably fine?

radarhere: Yep, should be fine, I just like to know why changes are made.

aclark4life: OK so 👍 👍 from @radarhere and @aclark4life and still trying to convince @hugovk ? 😄

radarhere: If https://github.com/pypa/setuptools/issues/2491 is never going to be resolved, then I don't have strong feelings either way about this PR.

PR #9654, opened 2026-06-03
Add benchmark tests

akx:
Related to https://github.com/python-pillow/Pillow/pull/9649 and https://github.com/python-pillow/pillow-perf/issues/161.

This PR ports most Pillow-related tests from [pillow-perf](https://github.com/python-pillow/pillow-perf/tree/main/testsuite/cases), modernized (no more compatibility with ancient versions of Pillow), converted to [pytest-benchmark](https://github.com/ionelmc/pytest-benchmark) style, and also jams in the blend benchmark I used to measure #9649's impact.

The file lives under `Tests/`, as it is a pytest file (and some benchmarks use test fixture images), but is not named `test_*`, so it will not be automatically collected as a test (and nothing untoward should happen if you don't have `pytest-benchmark` installed).

The second commit in this PR wires benchmarks into CI after successful tests.

On my Macbook, running the whole benchmark suite takes about 200 seconds right now -- in CI, this could be adjusted with e.g. `--benchmark-max-time` at the possible expense of measurement fidelity. Also, some of the benchmarks are currently somewhat exhaustive (combinatorial `parametrize`s). **EDIT:** Ugh, about 5 minutes (so longer than the test suite) in CI 😅

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -99,4 +99,6 @@ jobs:
       if: ${{ matrix.os == 'ubuntu-latest' }}
       with:
         mode: simulation
-        run: python3 -m pip install -e . pytest-codspeed && pytest -vv --codspeed Tests/benchmarks.py
+        run: |
+          python3 -m pip install -e . pytest-codspeed
+          pytest -vv --codspeed Tests/benchmarks.py
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,0 +1,102 @@
+name: Benchmark
+
+on:
+  push:
+    branches:
+      - "**"
+    paths-ignore: &paths-ignore
+      - ".github/workflows/docs.yml"
+      - ".github/workflows/wheels*"
+      - ".gitmodules"
+      - "docs/**"
+      - "wheels/**"
+  pull_request:
+    paths-ignore: *paths-ignore
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  FORCE_COLOR: 1
+  PIP_DISABLE_PIP_VERSION_CHECK: 1
+
+jobs:
+  benchmark:
+
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [
+          "ubuntu-latest",
+        ]
+        python-version: [
+          # NB: Codspeed doesn't seem to support multiple runs of the same benchmark:
+          #     > At least one benchmark was run multiple times in your benchmarking workflow.
+          #     > This could be because you used a matrix that runs your benchmarks multiple times.
+          #     Hence, this list should only ever have exactly one version.
+          "3.14",
+        ]
+
+    runs-on: ${{ matrix.os }}
+    name: Benchmark ${{ matrix.os }} Python ${{ matrix.python-version }}
+
+    steps:
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      with:
+        persist-credentials: false
+
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+      with:
+        python-version: ${{ matrix.python-version }}
+        allow-prereleases: true
+        cache: pip
+        cache-dependency-path: |
+          ".ci/*.sh"
+          "pyproject.toml"
+
+    - name: Cache libavif
+      if: startsWith(matrix.os, 'ubuntu')
+      uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+      id: cache-libavif
+      with:
+        path: ~/cache-libavif
+        key: ${{ runner.os }}-libavif-${{ hashFiles('depends/install_libavif.sh', 'depends/libavif-svt4.patch') }}
+
+    - name: Cache libimagequant
+      if: startsWith(matrix.os, 'ubuntu')
+      uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+      id: cache-libimagequant
+      with:
+        path: ~/cache-libimagequant
+        key: ${{ runner.os }}-libimagequant-${{ hashFiles('depends/install_imagequant.sh') }}
+
+    - name: Cache libwebp
+      if: startsWith(matrix.os, 'ubuntu')
+      uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+      id: cache-libwebp
+      with:
+        path: ~/cache-libwebp
+        key: ${{ runner.os }}-libwebp-${{ hashFiles('depends/install_webp.sh') }}
+
+    - name: Install Linux dependencies
+      if: startsWith(matrix.os, 'ubuntu')
+      run: |
+        .ci/install.sh
+      env:
+        GHA_PYTHON_VERSION: ${{ matrix.python-version }}
+        GHA_LIBAVIF_CACHE_HIT: ${{ steps.cache-libavif.outputs.cache-hit }}
+        GHA_LIBIMAGEQUANT_CACHE_HIT: ${{ steps.cache-libimagequant.outputs.cache-hit }}
+        GHA_LIBWEBP_CACHE_HIT: ${{ steps.cache-libwebp.outputs.cache-hit }}
+
+    - name: Run CodSpeed benchmarks
+      uses: CodSpeedHQ/action@a4a36bb07c0638b0b4ca52bf1f3dad1b4289e52f # v4.18.1
+      if: ${{ matrix.os == 'ubuntu-latest' }}
+      with:
+        mode: simulation
+        run: python3 -m pip install -e . pytest-codspeed && pytest -vv --codspeed Tests/benchmarks.py
--- a/Tests/benchmarks.py
+++ b/Tests/benchmarks.py
@@ -9,14 +9,15 @@ from io import BytesIO
 
 import pytest
 
-from PIL import Image, ImageFilter
+from PIL import Image, ImageChops, ImageFilter
 from PIL.Image import Resampling, Transpose
 
 TYPE_CHECKING = False
-
 if TYPE_CHECKING:
-    from pytest_benchmark.fixture import (
-        BenchmarkFixture,  # type: ignore[import-not-found]
+    from collections.abc import Callable
+
+    from pytest_benchmark.fixture import (  # type: ignore[import-not-found]
+        BenchmarkFixture,
     )
 
 pytest.importorskip(
@@ -43,6 +44,7 @@ PATHS = [
 # These are derived from the other configuration, above.
 RGB_MODES = [mode for mode in MODES if mode.startswith("RGB")]
 ALPHA_MODES = [mode for mode in MODES if mode.endswith("A")]
+SCALE_MODES = [*MODES, "I", "F"]
 
 
 def _format_size(size: tuple[int, int]) -> str:
@@ -103,7 +105,7 @@ def test_blend(
 @pytest.mark.benchmark(group="scale")
 @pytest.mark.parametrize("resampler", Resampling, ids=lambda r: r.name)
 @pytest.mark.parametrize("scale", [0.01, 0.125, 0.8, 2.14])
-@pytest.mark.parametrize("mode", MODES)
+@pytest.mark.parametrize("mode", SCALE_MODES)
 @pytest.mark.parametrize("size", SIZES, ids=_format_size)
 def test_scale(
     bench: BenchmarkFixture,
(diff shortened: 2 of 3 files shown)


.ci/benchmark.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+set -e
+
+python3 -m pip install pytest-benchmark
+
+python3 -bb -m pytest -vv --benchmark-only --benchmark-autosave Tests/benchmark*
```
radarhere:
```suggestion
python3 -bb -m pytest -vv --benchmark-only --benchmark-autosave Tests/benchmarks.py
```

akx: The idea with the wildcard is to prepare for additional benchmark files (or splitting the current one) in the future, but that can be done later as well. WDYT?

radarhere: My personal preference would be for later, but this is clearly a minor issue.

radarhere:
You are only running this in our Test jobs, not others. I'm guessing you did that because you didn't want to slow down all of our CI jobs?

Tests/benchmarks.py
```diff
@@ -0,0 +1,343 @@
...
+    bench.extra_info["label"] = [op.name]
+    bench(im.transpose, op)
+
+
+@pytest.mark.benchmark(group="load")
+@pytest.mark.parametrize("path", PATHS, ids=_format_path)
+def test_load(bench: BenchmarkFixture, path: pathlib.Path) -> None:
+    def run() -> None:
+        im = Image.open(path)
+        im.load()
```
radarhere:
```suggestion
        with Image.open(path) as im:
            im.load()
```
See https://pillow.readthedocs.io/en/stable/releasenotes/6.1.0.html#image-del

Tests/benchmarks.py
```diff
@@ -0,0 +1,343 @@
...
+
+    bench(run)
+
+
+@pytest.mark.benchmark(group="save")
+@pytest.mark.parametrize("path", PATHS, ids=_format_path)
+def test_save_jpeg(bench: BenchmarkFixture, path: pathlib.Path) -> None:
+    im = Image.open(path)
+    im.load()
+    bench(lambda: im.save(BytesIO(), format="JPEG", quality=85))
```
radarhere:
```suggestion
    with Image.open(path) as im:
        im.load()
        bench(lambda: im.save(BytesIO(), format="JPEG", quality=85))
```

akx:
> You are only running this in our Test jobs, not others. I'm guessing you did that because you didn't want to slow down all of our CI jobs?

Yep, exactly. Since building Pillow from source doesn't actually seem to take that long, maybe it'd be worth it to split it to a separate build + benchmark only job that's run on a single "known" (given these are GHA cloud runners, we may really never know, and neighbors may be noisy) platform?

Tests/benchmarks.py
```diff
@@ -0,0 +1,482 @@
...
+
+from PIL import Image, ImageChops, ImageFilter
+from PIL.Image import Resampling, Transpose
+
+TYPE_CHECKING = False
+
+if TYPE_CHECKING:
+    from pytest_benchmark.fixture import (  # type: ignore[import-not-found]
+        BenchmarkFixture,
+    )
```
hugovk:
```suggestion
from io import BytesIO

import pytest

from PIL import Image, ImageChops, ImageFilter
from PIL.Image import Resampling, Transpose

TYPE_CHECKING = False
if TYPE_CHECKING:
    from collections.abc import Callable
    from pytest_benchmark.fixture import (  # type: ignore[import-not-found]
        BenchmarkFixture,
    )
```

.github/workflows/test.yml
```diff
@@ -163,6 +163,16 @@ jobs:
...
 
+    - name: Run benchmarks
+      run: |
+        .ci/benchmark.sh
+
+    - name: Upload benchmarks
+      uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+      with:
+        name: benchmarks-${{ matrix.os }}-${{ matrix.python-version }}
+        path: .benchmarks
```
hugovk: This isn't working:
```
Warning: No files were found with the provided path: .benchmarks. No artifacts will be uploaded.
```
https://github.com/python-pillow/Pillow/actions/runs/27641704795/job/81743138114

hugovk (commented):
Thanks for this!

The benchmark runs are pretty slow, especially for PyPy, which means the end-to-end time for the workflow goes from ~12 mins to ~40 mins.

| Job | main | PR | Delta | Factor |
| --- | --- | --- | --- | --- |
| ubuntu 3.10 | 2m57s | 9m35s | +6m38s | 3.2x |
| ubuntu 3.11 | 3m36s | 9m32s | +5m56s | 2.6x |
| ubuntu 3.12 | 3m14s | 9m16s | +6m02s | 2.9x |
| ubuntu 3.13 | 3m20s | 9m21s | +6m01s | 2.8x |
| ubuntu 3.14 | 3m02s | 9m04s | +6m02s | 3.0x |
| ubuntu 3.14t | 3m02s | 8m54s | +5m52s | 2.9x |
| ubuntu 3.15 | 2m59s | 9m28s | +6m29s | 3.2x |
| ubuntu 3.15t | 3m15s | 8m51s | +5m36s | 2.7x |
| ubuntu pypy3.11 | 8m46s | 27m13s | +18m27s | 3.1x |
| macos 3.11 | 3m58s | 9m30s | +5m32s | 2.4x |
| macos 3.12 | 3m07s | 8m46s | +5m39s | 2.8x |
| macos 3.13 | 3m07s | 10m01s | +6m54s | 3.2x |
| macos 3.14 | 3m03s | 9m11s | +6m08s | 3.0x |
| macos 3.14t | 4m15s | 8m58s | +4m43s | 2.1x |
| macos 3.15 | 3m55s | 9m57s | +6m02s | 2.5x |
| macos 3.15t | 3m56s | 11m57s | +8m01s | 3.0x |
| macos pypy3.11 | 7m58s | 31m00s | +23m02s | 3.9x |
| macos-26-intel 3.10 | 7m18s | 17m15s | +9m57s | 2.4x |

Do we need to benchmark every version?

Also, this is benchmarking on noisy GHA runners, which might not be as reliable.

Should we use https://codspeed.io/ instead? My understanding is they have two approaches: one is to benchmark the real wall-time on their quiet servers, but there's a quota of 600 minutes per month for open source.

The other is to benchmark on GHA using pytest-codspeed, which uses a Valgrind/Cachegrind CPU simulator, where server noise is less an issue. It's slower, but does a single pass rather than many rounds, so might be quicker.

With both, we can see and compare results on their website, a bit like Codecov.

Again, we might not want to benchmark every version?

akx:
> The benchmark runs are pretty slow, especially for PyPy, which means the end-to-end time for the workflow goes from ~12 mins to ~40 mins.
> Also, this is benchmarking on noisy GHA runners, which might not be as reliable.

Yep, noted above. :)

> Should we use https://codspeed.io/ instead? My understanding is they have two approaches: one is to benchmark the real wall-time on their quiet servers, but there's a quota of 600 minutes per month for open source.

That's a possibility - happily, it looks like the pytest-benchmark `benchmark` fixture and pytest-codspeed `benchmark` fixture have the same signature. (I haven't used pytest-codspeed myself though.)

> Again, we might not want to benchmark every version?

I rebased things and disabled the run-in-CI commit's guts for now; it's there but commented out pending a decision.

hugovk:
Let's try it, I've enabled this repo on Codspeed:

https://app.codspeed.io/python-pillow/Pillow

Setup info:

1. Create benchmarks
Create or update your existing benchmarks and install the CodSpeed integration in your language. [Check out the docs on Creating benchmarks](https://codspeed.io/docs/benchmarks/overview)

2. Setup CI workflow
[Create or update a GitHub Actions workflow to run CodSpeed](https://codspeed.io/docs/integrations/ci/github-actions), and open a pull request to trigger a first run.

3. Merge to default branch
Merge the installation pull request on your default branch to receive detailed performance reports in each pull request.

akx:
@hugovk Wired up CodSpeed: https://github.com/python-pillow/Pillow/pull/9654/changes/3e91b4be90d506f19cf6274bd03dc1f7aa1b35b5

.github/workflows/test.yml
```diff
@@ -163,6 +163,12 @@ jobs:
         flags: ${{ matrix.os == 'ubuntu-latest' && 'GHA_Ubuntu' || 'GHA_macOS' }}
         name: ${{ matrix.os }} Python ${{ matrix.python-version }}
 
+    - name: Run CodSpeed benchmarks
```
hugovk: on macOS:

> Error: The valgrind executor is not supported on macos 15.7.7
https://github.com/python-pillow/Pillow/actions/runs/28358774308/job/84008071891?pr=9654

```suggestion
    - name: Run CodSpeed benchmarks
      if: startsWith(matrix.os, 'ubuntu')
```

And on Ubuntu, do we need to install pytest-codspeed?

> pytest: error: unrecognized arguments: --codspeed

https://github.com/python-pillow/Pillow/actions/runs/28358774308/job/84008071844?pr=9654

akx: Ah, right, I misunderstood the instructions a bit. Somehow I figured the codspeed action would also install whatever it needed, but that's not the case.

Limited to `ubuntu-latest` now too, let's see...

akx:
I'll let at least one of the codspeeds run to completion to see that results get uploaded correctly, then refactor these into a separate workflow file that runs independent of the test cycle on a more limited set of interpreters.

akx:
Per Codspeed's comment

> At least one benchmark was run multiple times in your benchmarking workflow. This could be because you used a matrix that runs your benchmarks multiple times.
> This behavior is not yet supported, so CodSpeed was unable to generate a performance report. 

I changed the new benchmark CI workflow to just run on ubuntu-latest+py3.14.

akx:
@hugovk All green now :+1:

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,102 @@
...
+        .ci/install.sh
+      env:
+        GHA_PYTHON_VERSION: ${{ matrix.python-version }}
+        GHA_LIBAVIF_CACHE_HIT: ${{ steps.cache-libavif.outputs.cache-hit }}
+        GHA_LIBIMAGEQUANT_CACHE_HIT: ${{ steps.cache-libimagequant.outputs.cache-hit }}
+        GHA_LIBWEBP_CACHE_HIT: ${{ steps.cache-libwebp.outputs.cache-hit }}
+
+    - name: Run CodSpeed benchmarks
+      uses: CodSpeedHQ/action@a4a36bb07c0638b0b4ca52bf1f3dad1b4289e52f # v4.18.1
+      if: ${{ matrix.os == 'ubuntu-latest' }}
```
hugovk, line 99: Redundant:
```suggestion
```

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,102 @@
...
+      with:
+        mode: simulation
+        run: python3 -m pip install -e . pytest-codspeed && pytest -vv --codspeed Tests/benchmarks.py
```
hugovk:
```suggestion
        run: |
          python3 -m pip install -e . pytest-codspeed
          pytest -vv --codspeed Tests/benchmarks.py
```

Tests/benchmarks.py
```diff
@@ -0,0 +1,481 @@
...
+TYPE_CHECKING = False
+if TYPE_CHECKING:
+    from collections.abc import Callable
+
+    from pytest_benchmark.fixture import (  # type: ignore[import-not-found]
+        BenchmarkFixture,
+    )
+
+if not (find_spec("pytest_benchmark") or find_spec("pytest_codspeed")):
+    pytest.skip("pytest-benchmark or pytest-codspeed required")
```
hugovk: Before:

```console
❯ pytest Tests/benchmarks.py
...
collected 0 items / 1 error

================================================= ERRORS ==================================================
__________________________________ ERROR collecting Tests/benchmarks.py ___________________________________
Using pytest.skip outside of a test will skip the entire module. If that's your intention, pass `allow_module_level=True`. If you want to skip a specific test or an entire class, use the @pytest.mark.skip or @pytest.mark.skipif decorators.
========================================= short test summary info =========================================
ERROR Tests/benchmarks.py
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
============================================ 1 error in 0.14s =============================================
```


After:

```console
❯ pytest Tests/benchmarks.py
Found existing alias for "pytest". You should use: "pt"
...
collected 0 items / 1 skipped

========================================= short test summary info =========================================
SKIPPED [1] Tests/benchmarks.py:25: pytest-benchmark or pytest-codspeed required
=========================================== 1 skipped in 0.08s ============================================
```



```suggestion
    pytest.skip("pytest-benchmark or pytest-codspeed required", allow_module_level=True)
```

hugovk (approved):
Thanks!

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,102 @@
...
+      with:
+        python-version: ${{ matrix.python-version }}
+        allow-prereleases: true
+        cache: pip
+        cache-dependency-path: |
+          ".ci/*.sh"
+          "pyproject.toml"
+
+    - name: Cache libavif
+      if: startsWith(matrix.os, 'ubuntu')
```
radarhere, line 64:
```suggestion
```

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,102 @@
...
+      uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+      id: cache-libavif
+      with:
+        path: ~/cache-libavif
+        key: ${{ runner.os }}-libavif-${{ hashFiles('depends/install_libavif.sh', 'depends/libavif-svt4.patch') }}
+
+    - name: Cache libimagequant
+      if: startsWith(matrix.os, 'ubuntu')
```
radarhere, line 72:
```suggestion
```

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,102 @@
...
+      uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+      id: cache-libimagequant
+      with:
+        path: ~/cache-libimagequant
+        key: ${{ runner.os }}-libimagequant-${{ hashFiles('depends/install_imagequant.sh') }}
+
+    - name: Cache libwebp
+      if: startsWith(matrix.os, 'ubuntu')
```
radarhere, line 80:
```suggestion
```

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,102 @@
...
+      uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+      id: cache-libwebp
+      with:
+        path: ~/cache-libwebp
+        key: ${{ runner.os }}-libwebp-${{ hashFiles('depends/install_webp.sh') }}
+
+    - name: Install Linux dependencies
+      if: startsWith(matrix.os, 'ubuntu')
```
radarhere, line 88:
```suggestion
```

akx:
I'd keep the (redundant, yes) `if: startsWith(matrix.os, 'ubuntu')` stanzas in the YAML, because it makes it that much easier to copy-paste from the test workflow if they change.

hugovk:
CodSpeed [settings](https://app.codspeed.io/python-pillow/Pillow/settings/general):

image

I've updated it to only comment when there's a change (improvement, regression or drop).

PR #9656, opened 2026-06-04
Prevent reusing ImagingDecoderObject.setimage

Serotav:
`ImagingDecoder.setimage()` binds a native decoder to an image before decoding. Reusing the same decoder with another image can leave decoder state from the previous binding in place.

 This adds a defensive guard so native `ImagingDecoder` objects reject a second `setimage()` call, and adds a regression test for that behavior. Existing image loading behavior is unchanged.

  Tested with:
  ```sh
  env PYTHONPATH=src python3 -m pytest Tests/test_imagefile.py
  ```

### DIFF
--- a/Tests/test_image.py
+++ b/Tests/test_image.py
@@ -1122,6 +1122,14 @@ class TestImage:
             for name in enum.__members__:
                 assert getattr(Image, name) == enum[name]
 
+    def test_decoder_setimage_once(self) -> None:
+        im = Image.new("L", (1, 1))
+        decoder = Image._getdecoder("L", "raw", "L")
+
+        decoder.setimage(im.im, None)
+        with pytest.raises(ValueError, match="decoder already has an image"):
+            decoder.setimage(im.im, None)
+
     @pytest.mark.parametrize(
         "path",
         [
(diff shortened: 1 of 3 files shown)


radarhere (approved):

hugovk:
Is a similar thing needed for encoder?

> Tested with:
> 
> ```shell
> env PYTHONPATH=src python3 -m pytest Tests/test_imagefile.py
> ```

Should this be `Tests/test_image.py`?

radarhere:
> Is a similar thing needed for encoder?

Not necessarily. The original trigger for this PR was that `state->bytes` from a previous `setimage()` call can be re-used in another call.
https://github.com/python-pillow/Pillow/blob/92e0e5f6b8200a24f202151c20b0e0775fddcad9/src/decode.c#L219-L224
This isn't the case for `_encode()`
https://github.com/python-pillow/Pillow/blob/92e0e5f6b8200a24f202151c20b0e0775fddcad9/src/encode.c#L291

> Should this be `Tests/test_image.py`?

The test was originally there, I moved it in https://github.com/Serotav/Pillow/pull/1

PR #9660, opened 2026-06-05
Raise ValueError if FPX tile size is not 64px by 64px

radarhere:
Page 52 (or 42 if you're looking at the printed numbers) of https://graphcomp.com/info/specs/livepicture/fpx.pdf states that
> **Tile width field**
> This field specifies the width of a tile in pixels. The tile width must be 64 pixels.
>
> **Tile height field**
> This field specifies the height of a tile in pixels. The tile height must be 64 pixels.

### DIFF
--- a/Tests/test_file_fpx.py
+++ b/Tests/test_file_fpx.py
@@ -46,7 +46,6 @@ def test_invalid_file() -> None:
 
 
 def test_invalid_tile_size() -> None:
-    # Test a valid OLE file, but not an FPX file
     with open("Tests/images/input_bw_one_band.fpx", "rb") as f:
         data = f.read()
     data = data[:4204] + b"\x00" * 8 + data[4212:]
--- a/Tests/test_file_fpx.py
+++ b/Tests/test_file_fpx.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+from io import BytesIO
+
 import pytest
 
 from PIL import Image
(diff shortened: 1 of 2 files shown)


Tests/test_file_fpx.py
```diff
@@ -43,7 +45,16 @@ def test_invalid_file() -> None:
         FpxImagePlugin.FpxImageFile(ole_file)
 
 
-def test_fpx_invalid_number_of_bands() -> None:
+def test_invalid_size() -> None:
+    # Test a valid OLE file, but not an FPX file
```
hugovk: This comment is a copy/paste and inaccurate here? Let's either update or remove.
```suggestion
```

Tests/test_file_fpx.py
```diff
@@ -43,7 +45,16 @@ def test_invalid_file() -> None:
         FpxImagePlugin.FpxImageFile(ole_file)
 
 
-def test_fpx_invalid_number_of_bands() -> None:
+def test_invalid_size() -> None:
```
hugovk:
```suggestion
def test_invalid_tile_size() -> None:
```

src/PIL/FpxImagePlugin.py
```diff
@@ -142,6 +142,10 @@ def _open_subimage(self, index: int = 1, subimage: int = 0) -> None:
         size = i32(s, 4), i32(s, 8)
         # tilecount = i32(s, 12)
         xtile, ytile = i32(s, 16), i32(s, 20)
+        if xtile != 64 or ytile != 64:
+            msg = "Tile must be 64 pixels by 64 pixels"
+            raise ValueError(msg)
```
hugovk, line 147: `ValueError` seems the right choice, although we're already raising `OSError` for other things which should probably be `ValueError` (if we were writing from scratch).

Maybe use `OSError` here for consistency? Or `ValueError` for any new ones? I don't mind too much either way.

hugovk (approved):

src/PIL/FpxImagePlugin.py
radarhere: Could go either way. I'm more inclined towards what is theoretically correct.

PR #9669, opened 2026-06-12
Invoke pyroma through the command line

### LINKED ISSUE #9668: Refactor Pyroma test to be non-runtime
Pillow currently has a [Pyroma test](https://github.com/python-pillow/Pillow/blob/main/Tests/test_pyroma.py) that is implemented as part of the runtime test suite.

This test needed some small tweaks to work on iOS (See #9116). I have tried to get those changes incorporated upstream to make the test less fragile for future PyRoma changes, but the maintainer has determined that using Pyroma as a library is out of scope: see regebro/pyroma#116.

So - Pillow should restructure the Pyroma test so that it *isn't* implemented as a runtime test, but as a validation of the actual generated wheel using PyRoma as an "external tool", rather than a library.

The issue to be resolved is with the Pillow test suite; the relevant test has been in the current form since at least Pillow 11.3.

radarhere:
Resolves #9668

The maintainer of pyroma has declared that it is [a command line tool, not a library.](https://github.com/regebro/pyroma/pull/116#issuecomment-4681987941) This PR updates test_pyroma.py to use `subprocess` to call it, rather than using the import.

You will notice that I have also removed our code handling `rc` versions (I don't think we've used this since 3.1.0.rc1) and relaxed the pyroma version requirement (now that the changes from #9116 are gone).

### DIFF
--- a/Tests/test_pyroma.py
+++ b/Tests/test_pyroma.py
@@ -1,55 +1,15 @@
 from __future__ import annotations
 
-from importlib.metadata import metadata
+import subprocess
+import sys
 
 import pytest
 
-from PIL import __version__
-
-TYPE_CHECKING = False
-
-if TYPE_CHECKING:
-    from importlib.metadata import PackageMetadata
-
-pyroma = pytest.importorskip("pyroma", reason="Pyroma not installed")
+pytest.importorskip("pyroma", reason="Pyroma not installed")
 
 
-def map_metadata_keys(md: PackageMetadata) -> dict[str, str | list[str] | None]:
-    # Convert installed wheel metadata into canonical Core Metadata 2.4 format.
-    # This was a utility method in pyroma 4.3.3; it was removed in 5.0.
-    # This implementation is constructed from the relevant logic from
-    # Pyroma 5.0's `build_metadata()` implementation. This has been submitted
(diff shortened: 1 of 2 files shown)


freakboy3742:
Using subprocess isn't an option for iOS, as Python doesn't support the subprocess module (the same will be true for Android or Emacripten).

If this is the approach that is adopted, the test will need to be skipped - `subprocess. _can_fork_exec` can be used for a property check.

However, I feel a better approach would be to do this outside the scope of pytest, and run it as a separate post-build check of wheels, rather than try to analyze a wheel while running code in (or adjacent to) that wheel.

radarhere:
I attempted to run it [outside of pytest](https://github.com/radarhere/Pillow/commit/2bf3ffc84466e90666ea381cbe90ecd5a238235d#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711), but it [still triggered a subprocess error on iOS.](https://github.com/radarhere/Pillow/actions/runs/27412021113/job/81015360059#step:5:10263)

hugovk:
Do we need to run it on all platforms as part of the test suite?

Can we instead put it in tox as part of the `lint` env?

For example: https://github.com/python-pillow/Pillow/pull/9670.

PR #9670, opened 2026-06-12
Run pyroma in `tox -e lint` instead of pytest

### LINKED ISSUE #9668: Refactor Pyroma test to be non-runtime

hugovk:
Fixes https://github.com/python-pillow/Pillow/issues/9668.
Replaces and closes https://github.com/python-pillow/Pillow/pull/9669.

Put pyroma in pre-commit. It can take a few seconds to run, so put it in the `manual` stage, so it's not run by default.

* A benefit of putting it in pre-commit is it is pinned, so won't suddenly fail when a new version comes out. We also get automated updates.

Run it in `tox -e lint`, which does the regular pre-commit, and is run on the CI.

Pyroma itself will run check-manifest ([when check-manifest is installed](https://github.com/regebro/pyroma/blob/732eba9006d74079d0ec360a6488e0847144103b/pyroma/ratings.py#L517-L543)), so don't run it twice in `tox -e lint` or `make release-test`.


freakboy3742 (approved):
👍 Looks good to me!

PR #1714, opened 2016-02-04
Overflow in resample.c, with tests

wiredfool:
Extension of #1711

PR #9672, opened 2026-06-14
Force the output of ImageOps.contain() to have at least 1px dimensions

### LINKED ISSUE #9997: ImageOps.contain and pad fail when a narrow image rounds to zero pixels
Resizing a narrow image with `ImageOps.contain()` can round the shorter dimension to zero, even when both requested dimensions are positive. `ImageOps.pad()` hits the same error because it calls `contain()`.

```python

image = Image.new("RGB", (100, 1), "red")
ImageOps.contain(image, (10, 10))
# ValueError: height and width must be > 0

ImageOps.pad(image, (10, 10))
# ValueError: height and width must be > 0
```

I would expect `contain()` to return a 10×1 image, and `pad()` to return that image within a 10×10 canvas. The transposed input, 1×100, fails too. A 20×1 input also fails at the half-pixel rounding boundary.

In `contain()`, the calculated dimension is passed directly from `round()` to `resize()`. Keeping that calculated dimension at least one pixel avoids the error.

Reproduced on Windows with Python 3.13.15 and Pillow 12.3.0, and with `ImageOps.py` from main at `0a61c530aed1a2792c6a47674ac2b70753bfccbb` loaded against the same installed Pillow core. No external image or network access is needed for the reproduction.

Prepared with OpenAI Codex assistance.

hecko-yes:
Resolves #9997. Alternative to #9673

I ran into `ValueError: height and width must be > 0` when trying to contain a very skinny image in a very small box. This fixes that.

### DIFF
--- a/Tests/test_imageops.py
+++ b/Tests/test_imageops.py
@@ -128,6 +128,16 @@ def test_contain_round() -> None:
     assert new_im.height == 5
 
 
+def test_contain_0px() -> None:
+    im = Image.new("1", (1, 256), 1)
+    new_im = ImageOps.contain(im, (8, 8))
+    assert new_im.width == 1
+
+    im = Image.new("1", (256, 1), 1)
+    new_im = ImageOps.contain(im, (8, 8))
+    assert new_im.height == 1
+
+
 @pytest.mark.parametrize(
     "image_name, expected_size",
     (
--- a/src/PIL/ImageOps.py
+++ b/src/PIL/ImageOps.py
@@ -290,11 +290,11 @@ def contain(
 
     if im_ratio != dest_ratio:
         if im_ratio > dest_ratio:
-            new_height = round(image.height / image.width * size[0])
+            new_height = max(1, round(image.height / image.width * size[0]))
             if new_height != size[1]:
                 size = (size[0], new_height)
         else:
-            new_width = round(image.width / image.height * size[1])
+            new_width = max(1, round(image.width / image.height * size[1]))
             if new_width != size[0]:
                 size = (new_width, size[1])
     return image.resize(size, resample=method)


radarhere:
An alternative way of fixing this problem would be to actually return an image with zero width (or height). We don't have to resize the existing image - if `size` is `(0, 8)`, we could just `return Image.new(image.mode, size)`. I'm slightly more inclined to this solution from a mathematical perspective. Do you have any thoughts?

Regardless though, just as an aside... the visual output of this is still unsatisfying, right? As in, I expect you have a large collection of images, and you're looking to display all of them in a smaller way within boxes, so that a user can view all of them at once. But there's no way that the final output with one dimension being 1px is actually something that the user can easily interpret.

hecko-yes:
In my case, I'm scaling it down to speed up subsequent color analysis, so looking pretty isn't a concern, but I understand this is not the most common usecase.

For user-facing applications, I believe e.g. Windows does just give you a 1px-wide image. Twitter puts a limit on the aspect ratio and switch to cropping if the image is beyond it, but I'm not sure if that's a good fit for a primitive such as this.

I'm not sure how well images with a dimension of 0 work in the rest of Pillow, but I assume there's a good reason why Image.resize refuses to produce them. :P

radarhere:
Zero size images are permitted, [except for some early Pillow versions.](https://github.com/python-pillow/Pillow/pull/2262)

The error when resizing an image was added in #1714. I don't mind it as a error that communicates 'There's no way that you actually wanted this, you must have done something wrong'. If anyone could provide an argument as to why you would ever want to use `resize()` directly to create a zero size image, I'd be personally open to changing Pillow's behaviour.

I've created #9673 as an alternative, to allow `ImageOps.contain()` to create zero width or height images.

radarhere:
Do you have any objections to https://github.com/python-pillow/Pillow/pull/9673? Are you fine as long as this is resolved somehow?

PR #9679, opened 2026-06-17
Embed SBOM into wheels

hugovk:
Follow on from https://github.com/python-pillow/Pillow/pull/9550, which generates an SBOM and uploads it to a GitHub release.

This PR embeds the SBOM into the [`.dist-info/sboms/`](https://packaging.python.org/en/latest/specifications/binary-distribution-format/#the-dist-info-sboms-directory) directory of wheels, following [PEP 770](https://peps.python.org/pep-0770/).

Things to note:

* The Linux wheels already contain an SBOM, `auditwheel.cdx.json`, created by auditwheel during the manylinux repair step of cibuildwheel.
  * The manylinux wheels contains only two things: Pillow and libXau, where libXau is what was grafted in from the system by auditwheel. It doesn't contain all the things in our own SBOM. 
  * The musllinux wheels also contain libbsd, libmd and libxdmcp (which are deps of libXau on Alpine).

* `auditwheel.cdx.json` uses CycloneDX 1.4, and varies for each generated wheel.

* Our own SBOM uses CycloneDX 1.7, and is identical for all wheels, and include all our deps, based on our `dependencies.json`. We don't vary per OS.

To inspect: download a wheel, rename to `.zip`, look in `pillow-12.3.0.dev0.dist-info/sboms/`.

---

Should we include our own generated SBOM in the wheel?

Should we keep the `auditwheel.cdx.json` SBOM?

Multiple SBOMs are allowed in the dir, and they do both describe the distribution, in different ways.

cc @jkowalleck @sethmlarson

### DIFF
--- a/.github/embed-sbom.py
+++ b/.github/embed-sbom.py
@@ -22,7 +22,7 @@ def record_entry(path: str, data: bytes) -> str:
     return f"{path},sha256={digest.rstrip(b'=').decode()},{len(data)}"
 
 
-def embed(wheel: Path, sbom_name: str, sbom_bytes: bytes) -> None:
+def embed(wheel: Path, sbom: Path) -> None:
     with zipfile.ZipFile(wheel) as zf:
         infos = zf.infolist()
         contents = {info.filename: zf.read(info.filename) for info in infos}
@@ -33,7 +33,9 @@ def embed(wheel: Path, sbom_name: str, sbom_bytes: bytes) -> None:
         if name.endswith(".dist-info/RECORD") and name.count("/") == 1
     )
     dist_info = record_name.rsplit("/", 1)[0]
-    sbom_path = f"{dist_info}/sboms/{sbom_name}"
+
+    sbom_bytes = sbom.read_bytes()
+    sbom_path = f"{dist_info}/sboms/{sbom.name}"
 
     # Append a matching RECORD line for the SBOM (RECORD's own line has no hash).
     lines = contents[record_name].decode("utf-8").splitlines()
@@ -49,16 +51,18 @@ def embed(wheel: Path, sbom_name: str, sbom_bytes: bytes) -> None:
         zf.writestr(sbom_path, sbom_bytes)
     tmp.replace(wheel)
 
+    print(f"Embedded {sbom.name} in {wheel.name}")
 
-def load_sbom(path: Path) -> tuple[str, bytes]:
-    """Read the SBOM; `path` may be the file or a directory containing one."""
+
+def scan_dir(path: Path) -> Path:
+    """If `path` is a directory, return the path of the SBOM within."""
     if path.is_dir():
         candidates = list(path.glob("*.cdx.json"))
         if len(candidates) != 1:
             msg = f"expected exactly one *.cdx.json in {path}, found {len(candidates)}"
             raise SystemExit(msg)
-        path = candidates[0]
-    return path.name, path.read_bytes()
+        return candidates[0]
+    return path
 
 
 def main() -> None:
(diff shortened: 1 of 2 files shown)


.github/embed-sbom.py
```diff
@@ -0,0 +1,91 @@
...
+        help="SBOM file, or a directory containing a single `.cdx.json`",
+    )
+    args = parser.parse_args()
+
+    sbom_name, sbom_bytes = load_sbom(args.sbom)
+
+    wheels = sorted(args.wheelhouse.glob("*.whl"))
+    if not wheels:
+        print(f"error: no wheels found in {args.wheelhouse}", file=sys.stderr)
+        raise SystemExit(1)
```
radarhere:
```suggestion
        parser.error(f"no wheels found in {args.wheelhouse}")
```

.github/workflows/wheels.yml
```diff
@@ -207,6 +246,15 @@ jobs:
...
         shell: bash
 
+      - name: Download SBOM
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          name: sbom
+          path: sbom
+
+      - name: Embed SBOM in wheels
+        run: python .github/embed-sbom.py wheelhouse sbom
```
```suggestion
        run: python3 .github/embed-sbom.py wheelhouse sbom
```

.github/embed-sbom.py
```diff
@@ -0,0 +1,98 @@
...
+    parser.add_argument(
+        "sbom",
+        type=Path,
+        help="SBOM file, or a directory containing a single `.cdx.json`",
+    )
+    args = parser.parse_args()
+
+    sbom = scan_dir(args.sbom)
+
+    wheels = sorted(args.wheelhouse.glob("*.whl"))
```
radarhere, line 87:
```suggestion
    wheels = args.wheelhouse.glob("*.whl")
```
Does this need to be sorted?

hugovk: It's not needed, but `glob` can return files in different order. Sorting means the logs will have things in the same order each time which makes then easier to debug/compare. But okay to drop if you prefer.

radarhere: No, that's fine, happy to leave it.

.github/embed-sbom.py
```diff
@@ -0,0 +1,98 @@
...
+    with zipfile.ZipFile(wheel) as zf:
+        infos = zf.infolist()
+        contents = {info.filename: zf.read(info.filename) for info in infos}
+
+    record_name = next(
+        name
+        for name in contents
+        if name.endswith(".dist-info/RECORD") and name.count("/") == 1
+    )
+    dist_info = record_name.rsplit("/", 1)[0]
```
radarhere:
```suggestion
    dist_info = record_name.split("/", 1)[0]
```
Extremely minor. Makes no difference.

.github/embed-sbom.py
```diff
@@ -0,0 +1,98 @@
...
+    contents[record_name] = b"\n".join(lines) + b"\n"
+
+    tmp = wheel.with_name(wheel.name + ".tmp")
+    with zipfile.ZipFile(tmp, "w", zipfile.ZIP_DEFLATED) as zf:
+        # Re-use each original ZipInfo to preserve timestamps, mode bits and
+        # compression; only RECORD's contents change.
+        for info in infos:
+            zf.writestr(info, contents[info.filename])
+        zf.writestr(sbom_path, sbom_bytes)
+    tmp.replace(wheel)
```
radarhere:
```suggestion
    with zipfile.ZipFile(wheel, "w", zipfile.ZIP_DEFLATED) as zf:
        # Re-use each original ZipInfo to preserve timestamps, mode bits and
        # compression; only RECORD's contents change
        for info in infos:
            zf.writestr(info, contents[info.filename])
        zf.writestr(sbom_path, sbom_bytes)
```
If you prefer not to, that's fine, but we could just overwrite the original wheel directly.

radarhere (approved):

PR #9714, opened 2026-06-22
Increase AVIF test epsilon for loong64

wszqkzqk:
#9606 has increased the epsilon of avif test, but it's still too small for loongarch.

The epsilon of `assert_image_similar_tofile` need to be larger than `6.9564` on loong64, and `assert_image_similar` need to be larger than `9.4067`.

This PR increases them to `7.0` and `9.5`, which is enough for loong64.

Tested on Arch Linux for Loong64.

### DIFF
--- a/Tests/test_file_avif.py
+++ b/Tests/test_file_avif.py
@@ -154,14 +154,14 @@ class TestFileAvif:
 
             # avifdec hopper.avif avif/hopper_avif_write.png
             assert_image_similar_tofile(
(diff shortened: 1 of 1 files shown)


radarhere:
Hi. Just for the record, what version of libavif did you use when you found this?

radarhere (approved):

wszqkzqk:
> Hi. Just for the record, what version of libavif did you use when you found this?

1.4.2

PR #9719, opened 2026-06-24
Ensure map stride is at least one full row of pixels

radarhere:


samueldr-at-cyberus:
Hi!

Since the fix is authored by the reporter, and smells like the plausible fix, would this be the related fix for GHSA-62p4-gmf7-7g93? (The GHSA does not list a patched version at this current time.)

Thanks!

hugovk:
@samueldr-at-cyberus Yes, that's right.

PR #9722, opened 2026-06-27
Check realloc return value

radarhere:
https://github.com/python-pillow/Pillow/blob/6a8de891fb00968e5ea79bfa84368ed90b3cfc1d/src/path.c#L340-L342

The comment comes from https://github.com/python-pillow/Pillow/commit/4b4ef5f1e2bf3c795fc60fbbf7c40f2a24043020. It is saying that since we a reallocating to a smaller size, the `realloc` operation should not fail.

However, perhaps we can satisfy code vulnerability scanners by checking it anyway.

### DIFF
--- a/src/path.c
+++ b/src/path.c
@@ -337,7 +337,6 @@ path_compact(PyPathObject *self, PyObject *args) {
     i = self->count - j;
 
     /* shrink coordinate array */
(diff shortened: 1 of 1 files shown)


hugovk (approved):
Do we still need the comment?

PR #9723, opened 2026-06-27
Deprecate P;2L and P;4L in frombytes()

radarhere:
Sequel to #9725

That PR raised a ValueError if P;2L or P;4L data is truncated in `frombytes()`. This PR deprecates their use as independent unpackers altogether. They can still be used when decoding PCX images.

I've moved the unpackers from Unpack.c to PcxDecode.c. Once the deprecation period is up, the `unpackers` and `RAWMODES` entries can be removed.

### DIFF
--- a/Tests/test_file_pcx.py
+++ b/Tests/test_file_pcx.py
@@ -213,5 +213,6 @@ def test_break_padding(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> None:
 )
 def test_truncated(data_len: int, rawmode: str) -> None:
     data = b"\x00" * data_len
-    with pytest.raises(ValueError, match="not enough image data"):
-        Image.frombuffer("P", (9, 1), data, "raw", rawmode, 0, 1)
+    with pytest.warns(DeprecationWarning, match=rawmode):
+        with pytest.raises(ValueError, match="not enough image data"):
+            Image.frombuffer("P", (9, 1), data, "raw", rawmode, 0, 1)
--- a/Tests/test_file_pcx.py
+++ b/Tests/test_file_pcx.py
@@ -217,5 +217,6 @@ def test_break_padding(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> None:
 )
 def test_truncated(data_len: int, rawmode: str) -> None:
     data = b"\x00" * data_len
-    with pytest.raises(ValueError, match="not enough image data"):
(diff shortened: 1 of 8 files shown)


akx:
Hmm, these could still be useful decoders for people using Pillow to eg. carve images out of, say, game binaries?

radarhere:
Could you expand further? What makes you think they are used for anything other than PCX images?

akx:
Sure. I've been involved in projects to port retro games, where [Pillow has been useful](https://github.com/suomipelit/ultimatetapankaikki/blame/master/TOOLS/efp2png.py) to convert custom bespoke file formats to modern ones. That particular example isn't P;2L or P;4L, sure, just a palettized image with RLE encoding, but in general Pillow's decoders being available to "userland" have been useful for this sort of thing. 

I don't particularly see a reason why this should be deprecated in `frombytes` - what's the rationale? It isn't clear from the PR description.

radarhere:
Because #9725 added rather specific code for the handling of these two rawmodes in a non-PCX context
https://github.com/python-pillow/Pillow/blob/cfc8a1ebd64155190b6e3d4924afa3a99af91263/src/PIL/Image.py#L953-L957
and if the rawmodes aren't necessary for any other encoder, I'd like to get rid of it.

akx:
> Because #9725 added rather specific code for the handling of these two rawmodes in a non-PCX context

The "rows are even" stands for PCX data (`PcxDecode` un-pads rows itself), not the raw data processed by unpackers and `frombytes`. Gently instrumenting `unpackP2L` shows `Image.frombytes()` can be made to read 5 bytes from a 4-byte buffer that passes that check on ab7e054339a9829f569e48751d1a56bc78168a90 (`main`):

```
$ uv run python -c 'from PIL import Image; Image.frombytes("P", (1, 4), bytes((1,2,3,4)), "raw", "P;2L")'
unpackP2L: in=0x1024b4740 i=0 j=0 pixels=1 reads 0x1024b4740 and 0x1024b4741
unpackP2L: in=0x1024b4741 i=0 j=0 pixels=1 reads 0x1024b4741 and 0x1024b4742
unpackP2L: in=0x1024b4742 i=0 j=0 pixels=1 reads 0x1024b4742 and 0x1024b4743
unpackP2L: in=0x1024b4743 i=0 j=0 pixels=1 reads 0x1024b4743 and 0x1024b4744
```

Now that I've investigated this and have something of a grip of the situation, I could open a PR to fix that? I think it would get rid of the Python-side check and fix the overread too.

PR #9726, opened 2026-06-28
Inspect arg as Python object, instead of using PyErr_Clear()

radarhere:
In `_convert_transparent()`, instead of failing a call to `PyArg_ParseTuple()`, calling `PyErr_Clear()` afterwards and then trying again,
https://github.com/python-pillow/Pillow/blob/095bbc312a34e2dbec68721da11c8722f8cf22e8/src/_imaging.c#L1084-L1089

an alternative strategy would be to parse the second argument as a PyObject ("O") and then check if it is a tuple or not.

A similar change could be made in `_convert_matrix()`.


src/_imaging.c
```diff
@@ -1049,8 +1049,11 @@ static PyObject *
 _convert_matrix(ImagingObject *self, PyObject *args) {
     char *mode_name;
     float m[12];
-    if (!PyArg_ParseTuple(args, "s(ffff)", &mode_name, m + 0, m + 1, m + 2, m + 3)) {
-        PyErr_Clear();
+    PyObject *matrix;
+    if (!PyArg_ParseTuple(args, "sO", &mode_name, &matrix)) {
+        return NULL;
+    }
+    if (PyTuple_Size(matrix) == 12) {
```
hugovk: This fails with lists:

```python
from PIL import Image

rgb = Image.new("RGB", (8, 8), (255, 0, 0))


def transparent(t):
    rgb.info["transparency"] = t
    return rgb.convert("RGBA")


cases = {
    "matrix tuple(4)": lambda: rgb.convert("L", matrix=(0.3, 0.59, 0.11, 0.0)),
    "matrix list(4)": lambda: rgb.convert("L", matrix=[0.3, 0.59, 0.11, 0.0]),
    "matrix tuple(12)": lambda: rgb.convert("RGB", matrix=tuple(range(12))),
    "matrix list(12)": lambda: rgb.convert("RGB", matrix=list(range(12))),
    "transparency tuple": lambda: transparent((255, 0, 0)),
    "transparency list": lambda: transparent([255, 0, 0]),
}

for label, fn in cases.items():
    try:
        fn()
        print(f"OK    {label}")
    except Exception as e:
        print(f"RAISE {label}: {type(e).__name__}: {e}")
```

Before:

```
OK    matrix tuple(4)
OK    matrix list(4)
OK    matrix tuple(12)
OK    matrix list(12)
OK    transparency tuple
OK    transparency list
```

After:

```
OK    matrix tuple(4)
RAISE matrix list(4): SystemError: <method 'convert_matrix' of 'ImagingCore' objects> returned a result with an exception set
OK    matrix tuple(12)
RAISE matrix list(12): SystemError: /Users/nad/build_macos_installer/installer/variant/binaries/build_source/Objects/tupleobject.c:96: bad argument to internal function
OK    transparency tuple
RAISE transparency list: TypeError: 'list' object cannot be interpreted as an integer
```

Maybe worth adding a test?


```suggestion
    if (PySequence_Size(matrix) == 12) {
```

src/_imaging.c
```diff
@@ -1070,27 +1073,35 @@ _convert_matrix(ImagingObject *self, PyObject *args) {
...
+    PyObject *transparency;
+    if (!PyArg_ParseTuple(args, "sO", &mode_name, &transparency)) {
+        return NULL;
     }
-    PyErr_Clear();
-    if (PyArg_ParseTuple(args, "si", &mode_name, &r)) {
-        const ModeID mode = findModeID(mode_name);
-        return PyImagingNew(ImagingConvertTransparent(self->image, mode, r, 0, 0));
+
+    if (PyTuple_Check(transparency)) {
```
hugovk:
```suggestion
    if (PySequence_Check(transparency)) {
```

src/_imaging.c
radarhere: `matrix` should be a tuple?
https://github.com/python-pillow/Pillow/blob/9c9feceb333043159c34809231fa188930867927/src/PIL/Image.py#L1017-L1020
https://github.com/python-pillow/Pillow/blob/9c9feceb333043159c34809231fa188930867927/src/PIL/Image.py#L1054-L1055

hugovk: Should according to type hint and docstring, but not enforced.

There is existing code using lists:

https://github.com/search?q=%2F%5C.convert%5C%28.*%2C+matrix%3D%5C%5B%2F&type=code

Do we want to change the behaviour in this PR?

radarhere: Ok, I pushed a commit to allow for lists.

src/_imaging.c
```diff
@@ -1048,8 +1048,15 @@ static PyObject *
 _convert_matrix(ImagingObject *self, PyObject *args) {
     char *mode_name;
     float m[12];
-    if (!PyArg_ParseTuple(args, "s(ffff)", &mode_name, m + 0, m + 1, m + 2, m + 3)) {
-        PyErr_Clear();
+    PyObject *matrix;
+    if (!PyArg_ParseTuple(args, "sO", &mode_name, &matrix)) {
+        return NULL;
+    }
+    Py_ssize_t size = PySequence_Size(matrix);
```
akx, line 1066: Review note to self: it's OK to not decref `matrix`, since `"O"` in `ParseTuple` doesn't create a strong reference.

src/_imaging.c
```diff
@@ -1069,27 +1076,35 @@ _convert_matrix(ImagingObject *self, PyObject *args) {
             )) {
             return NULL;
         }
+    } else if (!PyArg_ParseTuple(
+                   args, "s(ffff)", &mode_name, m + 0, m + 1, m + 2, m + 3
+               )) {
+        return NULL;
```
akx: With this, passing in a sequence of any other length than 4 or 12 always yields

> argument 2 must be tuple of length 4, not 5

which is a little confusing (since 12 is valid too):

```python
>>> i.convert("L", matrix=(5,5,5,5,5))
  File "<python-input-9>", line 1, in <module>
    i.convert("L", matrix=(5,5,5,5,5))
    ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^
  File "PIL/Image.py", line 1084, in convert
    im = self.im.convert_matrix(mode, matrix)
TypeError: argument 2 must be tuple of length 4, not 5
```

Could just do 

```c
    if (!(size == 4 || size == 12)) {
        return PyErr_Format(PyExc_TypeError, "matrix must be a sequence of 4 or 12 floats");
    }
```

above instead of the `size == -1` check without much extra ado?

Tests/test_image_convert.py
```diff
@@ -342,18 +346,23 @@ def test_matrix_xyz(mode: str) -> None:
 
 def test_matrix_identity() -> None:
     # Arrange
-    im = hopper("RGB")
+    im = hopper()
+    assert im.mode == "RGB"
```
akx: Wouldn't `im = hopper("RGB")` be clear enough about this Hopper's mode?

src/_imaging.c
```diff
@@ -1069,27 +1076,35 @@ _convert_matrix(ImagingObject *self, PyObject *args) {
...
-        return PyImagingNew(ImagingConvertTransparent(self->image, mode, r, 0, 0));
+
+    if (PySequence_Check(transparency)) {
+        if (!PyArg_ParseTuple(args, "s(iii)", &mode_name, &r, &g, &b)) {
+            return NULL;
+        }
+    } else if (!PyArg_ParseTuple(args, "si", &mode_name, &r)) {
+        return NULL;
     }
-    return NULL;
```
akx, line 1104: The same comment as above about confusing error messages stands here (the allowed values are either a single integer, or a 3-sequence of integers), but it's much harder to reach from userland, since you'd either need to call `i.im.convert_transparent()` by hand, or have explicitly set some image's `.info["transparency"]` to an invalid value. Might not be worth addressing.

```
>>> i.im.convert_transparent("RGBA", "hello")
  File "<python-input-22>", line 1, in <module>
    i.im.convert_transparent("RGBA", "hello")
    ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
TypeError: argument 2 must be 3-item tuple, not str
>>> i.im.convert_transparent("RGBA", 123)
<ImagingCore object at 0x1034fca50>
>>> i.im.convert_transparent("RGBA", [1,2,4])
<ImagingCore object at 0x1034ffeb0>
```

Tests/test_image_convert.py
radarhere: Ok, I've pushed a commit.

src/_imaging.c
radarhere: From what I see, main also behaves that way. Not saying we can't do this idea, just seems to be a bit of scope creep for this PR.

akx: It does.

The PR description doesn't really tell the reader why the preimage `PyErr_Clear()` approach was inadvisable, or why this change needs to be done in the first place. In that light, I assume the intent could be "Improve error handling in `convert` and `convert_transparent`", in which case improving errors to be more descriptive altogether (in the spirit of #9784) wouldn't be amiss?

radarhere: I've created https://github.com/python-pillow/Pillow/pull/9956

Tests/test_image_convert.py
```diff
@@ -352,6 +356,7 @@ def test_matrix_xyz(mode: str) -> None:
 def test_matrix_identity() -> None:
     # Arrange
     im = hopper("RGB")
+
```
```suggestion
```

Issue #9330, opened 2025-12-06
Python 3.15 support


lmmx:
### What did you do?

Installed a package (matplotlib) with a Pillow dependency on Python 3.15

### What did you expect to happen?

Build on CI

**edit** I think what actually this is due to is 3.15 wheels not being available for pillow

### What actually happened?

<details><summary>Details</summary>
<p>

```

 Downloaded opencv-python
   Building pillow==12.0.0
   Building scipy==1.16.3
   Building numpy==2.3.5
  × Failed to build `pillow==12.0.0`
  ├─▶ The build backend returned an error
  ╰─▶ Call to `backend.build_wheel` failed (exit status: 1)

      [stdout]
      running bdist_wheel
      running build
      running build_py
      creating build/lib.linux-x86_64-cpython-315/PIL
      copying src/PIL/_binary.py -> build/lib.linux-x86_64-cpython-315/PIL
      copying src/PIL/XVThumbImagePlugin.py ->
      build/lib.linux-x86_64-cpython-315/PIL
      copying src/PIL/PixarImagePlugin.py ->
      build/lib.linux-x86_64-cpython-315/PIL
      copying src/PIL/GribStubImagePlugin.py ->
      build/lib.linux-x86_64-cpython-315/PIL
      copying src/PIL/ImageStat.py -> build/lib.linux-x86_64-cpython-315/PIL
      copying src/PIL/QoiImagePlugin.py ->
      build/lib.linux-x86_64-cpython-315/PIL
      copying src/PIL/ImageDraw.py -> build/lib.linux-x86_64-cpython-315/PIL
      copying src/PIL/ContainerIO.py -> build/lib.linux-x86_64-cpython-315/PIL
      copying src/PIL/_deprecate.py -> build/lib.linux-x86_64-cpython-315/PIL
      copying src/PIL/ImageEnhance.py ->
      build/lib.linux-x86_64-cpython-315/PIL
      copying src/PIL/_util.py -> build/lib.linux-x86_64-cpython-315/PIL
      copying src/PIL/_version.py -> build/lib.linux-x86_64-cpython-315/PIL
      copying src/PIL/features.py -> build/lib.linux-x86_64-cpython-315/PIL
      copying src/PIL/TgaImagePlugin.py ->
      build/lib.linux-x86_64-cpython-315/PIL
      copying src/PIL/JpegPresets.py -> build/lib.linux-x86_64-cpython-315/PIL
[...]

PR #1207, opened 2015-04-23
Use logging instead of print.

anntzer:
cf. #1191, #1205 (recreated because I apparently can't reopen the previous PR).

Only TiffImagePlugin and OLEFileIO still rely on (their own) DEBUG flag.
I left TiffImagePlugin as it is because I hope #1059 gets merged in
first, and OLEFileIO because it uses its own logic.

PR #9733, opened 2026-06-29
Use lazy imports on Python 3.15 to improve startup speed

hugovk:
Re: https://github.com/python-pillow/Pillow/issues/9330

## Defer

The first commit here defers the import of `tempfile` in four files to the place where it's actually used, so we don't need to pay the import time if we're not using it. This helps for all Python versions.

So for example, only importing `PIL.Image` on Python 3.14:

```sh
pip install tuna
python3.14 -X importtime -c 'import PIL.Image' 2> import.log
tuna import.log
```

<table>
<tr>
<th>Before: 29 ms
<th>After: 23 ms
<tr>
<td valign="top">image
<td valign="top">image
</table>

## Lazy

Python 3.15 introduces lazy imports:

* https://docs.python.org/3.15/whatsnew/3.15.html#whatsnew315-lazy-imports
* https://peps.python.org/pep-0810/

If you put the new `lazy` keyword before an import, it won't be actually imported until first use. But `lazy` is only for 3.15+, it's a syntax error in 3.10-3.14.

> For code that cannot use the `lazy` keyword directly (for example, when supporting Python versions older than 3.15 while still using lazy imports on 3.15+), a module can define [`__lazy_modules__`](https://docs.python.org/3.15/reference/datamodel.html#module.__lazy_modules__) as a container of fully qualified module name strings. Regular `import` statements for those modules are then treated as lazy, with the same semantics as the `lazy` keyword

So the second commit introduces `__lazy_modules__`.

We can only put top-level imports here, and not those from function or in `try`/`except`. And we don't need to put any that used right away at the top-level, like `logging` in `Image.py`, but it's not a big deal if we do.

Repeating with 3.15:

<table>
<tr>
<th>Before: 22 ms
<th>After: 16 ms
<tr>
<td valign="top">image
<td valign="top">image
</table>


akx:
I think the 

```
ElementTree: ModuleType | None
try:
    from defusedxml import ElementTree
except ImportError:
    ElementTree = None
```

import should also belong in `getxmp()`? It'd become a neat `try: import; except: warnings.warn()` early return too.

hugovk:
Thanks, a couple extra "icicles" knocked off.

With 3.15, on a different machine compared to yesterday:

<table>
<tr>
<th>main: 26 ms
<th>PR before: 20 ms
<th>PR after: 17 ms
<tr>
<td valign="top">
image


<td valign="top">image


<td valign="top">image

</table>

akx:
It's certainly a bigger and unrelated change, but looks like `Pillow` only uses `logging` for a couple debug calls here and there in `PIL.Image` -- could be worth it to just get rid of it there (though real-world apps will generally probably import `logging` somehow either way, so additional laziness would need to be put in there in the stdlib...).

hugovk:
I made the olefile import eager again: it's an optional dependency and lazy imports affect how its related plugins get registered.

I'll have a look at logging. 

Unfortunately it's not straightforward to `lazy import traceback` in the stdib, because if we get an exception during shutdown, the import machinery might already be shutdown: https://github.com/python/cpython/pull/150073.

`lazy import re` in logging might be more doable, but in any case won't show up until Python 3.16.

hugovk:
Something like this?
```diff
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
-import logging
...
-logger = logging.getLogger(__name__)
+class _LazyLogger:
+    """
+    Defer importing `logging` until the logger is first used to avoid slow import.
+    """
+
+    def __getattr__(self, attr: str) -> Any:
+        import logging
+
+        logger = logging.getLogger(__name__)
+        globals()["logger"] = logger
+        return getattr(logger, attr)
+
+
+logger = _LazyLogger()
```

<table>
<tr>
<th>Before: 16 ms
<th>After: 12 ms
<tr>
<td valign="top">
image

<td valign="top">
image

</table>

akx:
@hugovk That'd work -- but I'm also thinking... do we need the 4 `logger.debug()`s for importing plugin modules in `PIL/Image.py` at all?

hugovk:
@radarhere Do you think we should keep those four log lines?

radarhere:
I'm inclined not to.

The debug message before the import seems unnecessary, as 99% of the time, the import will succeed, and it's not part of an overall consistent use of `logger.debug` on our part.

As for the one on error… I'm not sure under what circumstances it would run for any of our internal plugins. From a big picture perspective though, PIL `print`ed the exception on `DEBUG` and #1207 changed it to use `logger`. I think `PIL.report` (#3870) has since become the easiest way to figure out which plugins are working and which aren't.

hugovk:
`logging` dropped from `Image`:

<table>
<tr>
<th>Before: 20 ms
<th>After: 10 ms
<tr>
<td valign="top">
image

<td valign="top">
image

</table>

src/PIL/ImageFilter.py
```diff
@@ -16,6 +16,8 @@
 #
 from __future__ import annotations
 
+__lazy_modules__ = {"collections.abc", "functools", "typing"}
```
hugovk: If https://github.com/python-pillow/Pillow/pull/9860 is merged first:
```suggestion
__lazy_modules__ = {"collections.abc", "typing"}
```

src/PIL/ImageOps.py
```diff
@@ -18,6 +18,8 @@
 #
 from __future__ import annotations
 
+__lazy_modules__ = {"collections.abc", "functools", "operator", "re"}
```
hugovk:
```suggestion
__lazy_modules__ = {"collections.abc", "re"}
```

src/PIL/XpmImagePlugin.py
```diff
@@ -15,6 +15,8 @@
 #
 from __future__ import annotations
 
+__lazy_modules__ = {"PIL._binary"}
+
 import re
 
 from . import Image, ImageFile, ImagePalette
```
radarhere, line 22: Am I correct in thinking that's there no particular reason why `_binary` should be lazy, but `ImagePalette`, isn't? I'm guessing you ran some performance tests, figured out which modules to import lazily based on that, and then applied that list throughout src?

hugovk: > Am I correct in thinking that's there no particular reason why `_binary` should be lazy, but `ImagePalette`, isn't? 

With `from ._binary import o8`, the module being imported is `PIL._binary`.

With `from . import ImagePalette`, the module being imported is the `PIL` package.

> I'm guessing you ran some performance tests, figured out which modules to import lazily based on that, and then applied that list throughout src?

I used https://github.com/henryiii/flake8-lazy and ran something like `flake8-lazy src/PIL --apply set` (set being faster than list: https://github.com/henryiii/flake8-lazy/issues/50).

Although that does spells it like:

```python
__lazy_modules__ = {f"{__spec__.parent}._binary"}
```

See https://github.com/henryiii/flake8-lazy/issues/41 for the explanation. I decided to replace the f-strings in this PR.

But perhaps we should go for the f-strings and add flake8-lazy to pre-commit? (There's no [Ruff rule yet](https://github.com/henryiii/flake8-lazy#why-is-this-not-in-ruff).)

hugovk:
Updated: re-ran `flake8-lazy src/PIL --apply set` with the newer version (but evaluated f-strings), which removes some lazy imports because they're always imported, and others because they get "reified" (imported) immediately, and adds some extra imports the earlier version didn't catch.

PR #9735, opened 2026-06-30
Remove matrix.os from benchmark

radarhere:
Comparing benchmark.yml with test.yml, benchmark only supports Ubuntu. It doesn't have [the option to install macOS dependencies.](https://github.com/python-pillow/Pillow/blob/9c9feceb333043159c34809231fa188930867927/.github/workflows/test.yml#L113-L118)

Rather than hinting that it supports other operating systems, let's just remove matrix.os completely.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -31,9 +31,6 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [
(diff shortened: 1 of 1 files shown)


hugovk:
I had left it to make the diff between `test.yml` smaller, but don't mind much either way, so feel free to merge if you prefer it this way.

PR #9675, opened 2026-06-16
Speed up `Image.getchannel()`, `Image.merge()`, `Image.putalpha()` and `Image.split()`

akx:
Refs #9649 (same technique, different loops).

On my machine, this speeds up `.merge()` for RGB images by about 3x according to the benchmark code from #9654.

PR #9736, opened 2026-06-30
Speed up `Image.filter()`

akx:
Refs #9649, #9675 (same series, same techniques).

### DIFF
--- a/src/libImaging/Filter.c
+++ b/src/libImaging/Filter.c
@@ -113,10 +113,8 @@ kernel_i16(int size, UINT8 *in0, int x, const float *kernel, int bigendian) {
     int half_size = (size - 1) / 2;
     for (i = 0; i < size; i++) {
         int x1 = x + i - half_size;
-        result += _i2f(
-                      in0[x1 * 2 + (bigendian ? 1 : 0)] +
-                      (in0[x1 * 2 + (bigendian ? 0 : 1)] >> 8)
-                  ) *
+        result += (float)(in0[x1 * 2 + (bigendian ? 1 : 0)] +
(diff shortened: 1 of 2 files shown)


akx:
@codspeed-hq Silly! 😄 It does improve performance, but only about 2% to 10% so it's under your improvement threshold...

src/libImaging/Filter.c
```diff
@@ -271,32 +271,38 @@ ImagingFilter3x3(Imaging imOut, Imaging im, const float *kernel, float offset) {
     memcpy(imOut->image[y], im->image[y], im->linesize);
 }
 
-void
+static void
 ImagingFilter5x5(Imaging imOut, Imaging im, const float *kernel, float offset) {
-#define KERNEL1x5(in0, x, kernel, d)                                       \
-    (_i2f(in0[x - d - d]) * (kernel)[0] + _i2f(in0[x - d]) * (kernel)[1] + \
-     _i2f(in0[x]) * (kernel)[2] + _i2f(in0[x + d]) * (kernel)[3] +         \
-     _i2f(in0[x + d + d]) * (kernel)[4])
```
hugovk, line 279: Does this mean `_i2f` is no longer used and we can delete it from `src/libImaging/ImagingUtils.h`?

akx: Yes, but that felt out of scope for this PR. I can do that after this gets merged -- or if you like, I can add it here as a new commit.

hugovk: Let's do it here.

hugovk (approved):

PR #9737, opened 2026-06-30
Speed up `Image.fill()`, `Image.linear_gradient()` and `Image.radial_gradient()`.

akx:
Refs https://github.com/python-pillow/Pillow/pull/9649, https://github.com/python-pillow/Pillow/pull/9675, #9736 (same series, same techniques).

I didn't even know about `linear_gradient` and `radial_gradient` before looking at this...

### DIFF
--- a/src/libImaging/Fill.c
+++ b/src/libImaging/Fill.c
@@ -21,7 +21,6 @@
 
 Imaging
 ImagingFill(Imaging im, const void *colour) {
-    int x, y;
     ImagingSectionCookie cookie;
 
     /* 0-width or 0-height image. No need to do anything */
@@ -29,19 +28,23 @@ ImagingFill(Imaging im, const void *colour) {
         return im;
     }
 
+    // xsize and ysize are invariant during the loops below.
+    int xsize = im->xsize;
+    int ysize = im->ysize;
+
     if (im->type == IMAGING_TYPE_SPECIAL) {
         /* use generic API */
         ImagingAccess access = ImagingAccessNew(im);
         if (access) {
-            for (y = 0; y < im->ysize; y++) {
-                for (x = 0; x < im->xsize; x++) {
+            for (int y = 0; y < ysize; y++) {
+                for (int x = 0; x < xsize; x++) {
                     access->put_pixel(im, x, y, colour);
                 }
             }
             ImagingAccessDelete(im, access);
         } else {
             /* wipe the image */
-            for (y = 0; y < im->ysize; y++) {
+            for (int y = 0; y < ysize; y++) {
                 memset(im->image[y], 0, im->linesize);
             }
         }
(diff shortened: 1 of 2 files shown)


akx:
My local results for the new benchmarks:

`linear_gradient` was already well-optimized (the generated assembly did the right thing) for 1/L/P/F in the baseline, hence no gains over measurement noise there.

```
  -------------------------- benchmark 'allocate': 10 tests, 2 sources --------------------------
  Name (time in us)           0003_basegra OPS (Kops/s)  0004_newgrad OPS (Kops/s)  ΔOPS (Kops/s)
  -----------------------------------------------------------------------------------------------
  test_linear_gradient[1]                      575.2680                   581.8250          +1.1%
  test_linear_gradient[L]                      576.4618                   569.9319          -1.1%
  test_linear_gradient[P]                      421.5023                   400.9303          -4.9%
  test_linear_gradient[F]                      195.4028                   203.2991          +4.0%
  test_linear_gradient[I]                       39.9882                   201.5319        +404.0%
  test_radial_gradient[F]                       26.4618                    26.8236          +1.4%
  test_radial_gradient[L]                       22.2012                    26.1221         +17.7%
  test_radial_gradient[1]                       22.9234                    25.6476         +11.9%
  test_radial_gradient[P]                       22.1137                    25.4123         +14.9%
  test_radial_gradient[I]                       19.9459                    25.6850         +28.8%
  -----------------------------------------------------------------------------------------------
```

src/libImaging/Fill.c
```diff
@@ -118,22 +125,36 @@ ImagingFillRadialGradient(const ModeID mode) {
...
+    // Branch on pixel type outside the loops so the compiler can tighten them.
+    // Restrict safe: sole owner of the freshly-allocated image data here.
+    if (im->image8) {
+        for (int y = 0; y < 256; y++) {
+            UINT8 *restrict row = im->image8[y];
+            for (int x = 0; x < 256; x++) {
+                int d = (int)sqrt(
+                    (double)((x - 128) * (x - 128) + (y - 128) * (y - 128)) * 2.0
+                );
+                row[x] = d >= 255 ? 255 : d;
```
hugovk: We now have this block repeated three times.

Would a helper function be any use, or would we just lose the speed up?

Or a macro? But macros aren't so great either.

akx: A macro could work; it'd look something like

```c
#define ASSIGN_PIXEL(row, x, y)                                                       \
    {                                                                                 \
        int d =                                                                       \
            (int)sqrt((double)((x - 128) * (x - 128) + (y - 128) * (y - 128)) * 2.0); \
        row[x] = d >= 255 ? 255 : d;                                                  \
    }

    // Branch on pixel type outside the loops so the compiler can tighten them.
    // Restrict safe: sole owner of the freshly-allocated image data here.
    if (im->image8) {
        for (int y = 0; y < 256; y++) {
            UINT8 *restrict row = im->image8[y];
            for (int x = 0; x < 256; x++) {
                ASSIGN_PIXEL(row, x, y);
            }
        }
    } else if (im->type == IMAGING_TYPE_FLOAT32) {
        for (int y = 0; y < 256; y++) {
            FLOAT32 *restrict row = (FLOAT32 *)im->image32[y];
            for (int x = 0; x < 256; x++) {
                ASSIGN_PIXEL(row, x, y);
            }
        }
    } else {
        for (int y = 0; y < 256; y++) {
            INT32 *restrict row = im->image32[y];
            for (int x = 0; x < 256; x++) {
                ASSIGN_PIXEL(row, x, y);
            }
        }
    }

    return im;
```

hugovk: OK, let's try that.

hugovk (approved):

PR #9738, opened 2026-06-30
Speed up ImageChops operations

akx:
Refs https://github.com/python-pillow/Pillow/pull/9649, https://github.com/python-pillow/Pillow/pull/9675, https://github.com/python-pillow/Pillow/pull/9736, https://github.com/python-pillow/Pillow/pull/9737 (same series, same techniques).

### DIFF
--- a/src/libImaging/Chops.c
+++ b/src/libImaging/Chops.c
@@ -18,6 +18,7 @@
 
 #include "Imaging.h"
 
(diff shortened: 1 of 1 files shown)


akx:
I don't think the `Tests.test_file_avif.TestAvifLeaks` test failure on 3.15t is related to this patch 😅

hugovk (approved):

PR #9739, opened 2026-06-30
Speed up `Image.resample()`

akx:
Refs https://github.com/python-pillow/Pillow/pull/9649, https://github.com/python-pillow/Pillow/pull/9675, https://github.com/python-pillow/Pillow/pull/9736, https://github.com/python-pillow/Pillow/pull/9737, #9738 (same series, same techniques).

Follows up on https://github.com/python-pillow/Pillow/pull/1881.

### DIFF
--- a/src/libImaging/Resample.c
+++ b/src/libImaging/Resample.c
@@ -1,5 +1,6 @@
 #include "Imaging.h"
 
+#include <assert.h>
 #include <math.h>
 
 #define ROUND_UP(f) ((int)((f) >= 0.0 ? (f) + 0.5F : (f) - 0.5F))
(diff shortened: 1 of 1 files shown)


akx:
<del>Well, those assertion failures are surprising... 🤔 </del>

No, they're not, I'm just a dummy and I copy-pasted an `image8` assertion where we needed `image32`. However, the _actually_ surprising thing is that Windows tests didn't assert 🤔

hugovk (approved):

src/libImaging/Resample.c
```diff
@@ -296,88 +297,99 @@ ImagingResampleHorizontal_8bpc(
...
 
+    // Sizes are invariant over the loop.
+    int xsize = imOut->xsize, ysize = imOut->ysize;
+
     ImagingSectionEnter(&cookie);
     if (imIn->image8) {
-        for (yy = 0; yy < imOut->ysize; yy++) {
-            for (xx = 0; xx < imOut->xsize; xx++) {
+        // Verify caller allocated distinct buffers.
+        assert(imIn->image8 != imOut->image8);
```
radarhere, line 306: Just for my understanding, why does this need an `assert`, when similar changes you've made elsewhere don't?

akx: I added defensive `assert()`s here since the call chain is more involved than with the other optimizations in this series (where it's plain to see that there's an `ImagingNewDirty` just nearby).

PR #9742, opened 2026-06-30
Remove remaining :rtype:s from docstrings

akx:
The type annotations for the functions describe the return type anyway, so no great reason I can think of to keep a redundant other copy around.

Noticed these in Chops while working on #9738 and figured I'd do a full sweep.

### DIFF
--- a/src/PIL/ImageChops.py
+++ b/src/PIL/ImageChops.py
@@ -21,19 +21,13 @@ from . import Image
 
 
 def constant(image: Image.Image, value: int) -> Image.Image:
-    """Fill a channel with a given gray level.
-
-    :rtype: :py:class:`~PIL.Image.Image`
-    """
+    """Fill a channel with a given gray level."""
 
     return Image.new("L", image.size, value)
 
 
 def duplicate(image: Image.Image) -> Image.Image:
-    """Copy a channel. Alias for :py:meth:`PIL.Image.Image.copy`.
-
-    :rtype: :py:class:`~PIL.Image.Image`
-    """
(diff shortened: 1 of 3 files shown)


src/PIL/ImageChops.py
```diff
@@ -168,11 +144,11 @@ def add(
 ) -> Image.Image:
     """
     Adds two images, dividing the result by scale and adding the
-    offset. If omitted, scale defaults to 1.0, and offset to 0.0. ::
+    offset. If omitted, scale defaults to 1.0, and offset to 0.0.
 
-        out = ((image1 + image2) / scale + offset)
+    ::
```
radarhere, line 149: Just curious, why did you split this to a new line?

akx: Iirc if I didn't, then one of the linters would un-"block" the whole example.

radarhere (approved):

akx:
How did the `main` commit fc7cec1751b90b253ee1059b57185f3734852fe0 end up with @hugovk's authorship 😮

Or rather, I think squash merging PRs has retained the PR author's credits in the past? 🤔

hugovk:
It's not a squash merge commit, it's a regular merge commit.

https://github.com/python-pillow/Pillow/commit/fc7cec1751b90b253ee1059b57185f3734852fe0 is my merge and it has two parents, one of which is your commit https://github.com/python-pillow/Pillow/commit/5ad6072e9115ec8d042dce96f4e48b7848d224d5.

akx:
Ah! Sorry for the noise, my bad. I didn't check the parents correctly. Thanks!

PR #9740, opened 2026-06-30
Speed up `alpha_composite`, `matrix`, `negative`, `quantize`

akx:
Refs https://github.com/python-pillow/Pillow/pull/9649, https://github.com/python-pillow/Pillow/pull/9675, https://github.com/python-pillow/Pillow/pull/9736, https://github.com/python-pillow/Pillow/pull/9737, https://github.com/python-pillow/Pillow/pull/9738, #9739 (same series, same techniques).

These were a couple lines each, so I figured I'd bundle them into one PR.

PR #9743, opened 2026-06-30
Speed up `Image.point()` and `Image.point_transform()`

akx:
Refs https://github.com/python-pillow/Pillow/pull/9649, https://github.com/python-pillow/Pillow/pull/9675, https://github.com/python-pillow/Pillow/pull/9736, https://github.com/python-pillow/Pillow/pull/9737, https://github.com/python-pillow/Pillow/pull/9738, https://github.com/python-pillow/Pillow/pull/9739, #9740 (same series, same techniques).

Cf. the aforementioned PRs, this also does a couple other cleanups for performance; on my machine, using ImagingNewDirty gave another +10% on top of the first commit. The last commit would technically allow a compiler to inline the `point` functions, but given their size, a compiler probably won't.

### DIFF
--- a/src/libImaging/Point.c
+++ b/src/libImaging/Point.c
@@ -25,29 +25,39 @@ typedef struct {
     const void *table;
 } im_point_context;
 
+/**
+ * Contract: imIn is read-only and imOut must be a distinct image.
+ */
 static void
 im_point_8_8(Imaging imOut, Imaging imIn, im_point_context *context) {
-    int x, y;
     /* 8-bit source, 8-bit destination */
     UINT8 *table = (UINT8 *)context->table;
-    for (y = 0; y < imIn->ysize; y++) {
-        UINT8 *in = imIn->image8[y];
-        UINT8 *out = imOut->image8[y];
-        for (x = 0; x < imIn->xsize; x++) {
+    // Invariant over the loop.
+    int xsize = imIn->xsize, ysize = imIn->ysize;
+    for (int y = 0; y < ysize; y++) {
+        // restrict safe: imIn is read-only, imOut is a fresh allocation.
+        UINT8 *restrict in = imIn->image8[y];
+        UINT8 *restrict out = imOut->image8[y];
+        for (int x = 0; x < xsize; x++) {
             out[x] = table[in[x]];
         }
     }
 }
 
+/**
+ * Contract: imIn is read-only and imOut must be a distinct image.
+ */
 static void
 im_point_2x8_2x8(Imaging imOut, Imaging imIn, im_point_context *context) {
-    int x, y;
     /* 2x8-bit source, 2x8-bit destination */
     UINT8 *table = (UINT8 *)context->table;
-    for (y = 0; y < imIn->ysize; y++) {
-        UINT8 *in = (UINT8 *)imIn->image[y];
-        UINT8 *out = (UINT8 *)imOut->image[y];
-        for (x = 0; x < imIn->xsize; x++) {
+    // Invariant over the loop.
+    int xsize = imIn->xsize, ysize = imIn->ysize;
+    for (int y = 0; y < ysize; y++) {
+        // restrict safe: imIn is read-only, imOut is a fresh allocation.
+        UINT8 *restrict in = (UINT8 *)imIn->image[y];
+        UINT8 *restrict out = (UINT8 *)imOut->image[y];
+        for (int x = 0; x < xsize; x++) {
             out[0] = table[in[0]];
             out[3] = table[in[3] + 256];
             in += 4;
@@ -56,15 +66,20 @@ im_point_2x8_2x8(Imaging imOut, Imaging imIn, im_point_context *context) {
     }
 }
 
+/**
+ * Contract: imIn is read-only and imOut must be a distinct image.
+ */
 static void
 im_point_3x8_3x8(Imaging imOut, Imaging imIn, im_point_context *context) {
-    int x, y;
     /* 3x8-bit source, 3x8-bit destination */
     UINT8 *table = (UINT8 *)context->table;
-    for (y = 0; y < imIn->ysize; y++) {
-        UINT8 *in = (UINT8 *)imIn->image[y];
-        UINT8 *out = (UINT8 *)imOut->image[y];
-        for (x = 0; x < imIn->xsize; x++) {
+    // Invariant over the loop.
+    int xsize = imIn->xsize, ysize = imIn->ysize;
+    for (int y = 0; y < ysize; y++) {
+        // restrict safe: imIn is read-only, imOut is a fresh allocation.
+        UINT8 *restrict in = (UINT8 *)imIn->image[y];
+        UINT8 *restrict out = (UINT8 *)imOut->image[y];
+        for (int x = 0; x < xsize; x++) {
             out[0] = table[in[0]];
             out[1] = table[in[1] + 256];
             out[2] = table[in[2] + 512];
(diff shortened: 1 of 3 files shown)


hugovk:

src/libImaging/Point.c
```diff
@@ -25,64 +25,90 @@ typedef struct {
...
+    int xsize = imIn->xsize, ysize = imIn->ysize;
+    for (int y = 0; y < ysize; y++) {
+        // restrict safe: imIn is read-only, imOut is a fresh allocation.
+        UINT8 *restrict in = (UINT8 *)imIn->image[y];
+        UINT8 *restrict out = (UINT8 *)imOut->image[y];
+        for (int x = 0; x < xsize; x++) {
             out[0] = table[in[0]];
             out[1] = table[in[1] + 256];
             out[2] = table[in[2] + 512];
+            out[3] = 0;
```
radarhere, line 91: This doesn't look performance-related. Did you want to explain why you added this?

akx: It is in fact performance-related. As noted at the top of the file,

```
// Note: all point handlers must fill the entirety of `imOut`,
//       as it is allocated dirty by `ImagingPoint`.
```

and without this, that contract isn't fulfilled.

The same thing happens above in `im_point_2x8_2x8`.

espressolee (commented):
I built both sides and measured the change, and also checked the new initialization requirements. Everything I probed for correctness checked out; the main observation I have is about attribution.

## Setup

Base `c8c74c8`, head `1000b1b`, CPython 3.14.6 on macOS arm64. Base and head runs are interleaved and the order alternates each round, so both arms share whatever drift the machine has. The statistic is the median per-call time over 15 equal-duration batches, five rounds. `Image.copy()` is carried through every run as a negative control, since this PR does not touch it.

## The PR contains two independent changes

`ImagingNew` → `ImagingNewDirty` is one. `restrict` plus the hoisted loop bounds is the other. Only the first creates an obligation — every handler must now fill every byte of `imOut`, including the padding lanes, currently enforced by hand in four places.

I measured them separately by building a third tree that is head with only the allocation call reverted. At 1024×1024:

| case | whole PR | loop changes only | allocation change only |
| --- | ---: | ---: | ---: |
| `L`→`I` (8_32) | 0.686x | 0.742x | 0.925x |
| `L` (8_8) | 0.724x | 0.747x | 0.970x |
| `LA` (2x8) | 0.835x | 0.863x | 0.965x |
| `RGB` (3x8) | 0.887x | 0.902x | 0.980x |
| `RGBA` (4x8) | 0.931x | 0.940x | 0.986x |
| **control, `Image.copy()`** | **0.997x** | **1.010x** | **0.985x** |

The middle column was also measured directly against base rather than derived from the other two; the two ways of getting it agree to within 0.4%.

At 64×64 the loop-change results on this setup range from 0.965x to 1.002x, so I could not resolve a consistent gain at that size.

## Reading

On this machine and build, the loop changes account for the larger share of the measured speedup. The allocation change shows a clear additional gain for `L`→`I` (about 7.5%); for the other four paths its measured gain is 1.4–3.5%, close enough to the 1–1.5% control variation that I would not draw a strong conclusion from those alone. That smaller or less certain gain comes with a permanent requirement that every present and future handler fill every output byte.

Whether that tradeoff is worthwhile is a maintainer decision; I mainly wanted to make the two contributions visible. I am not necessarily suggesting this be split into separate PRs — I measured them separately because they have different performance and maintenance tradeoffs.

These proportions are from one machine, one compiler and one OS, and I would not assume they carry to GCC, x86-64 or a different allocator.

## Correctness

I also checked the obvious points: `restrict` appears safe because the handlers are `static` and `imOut` is allocated within `ImagingPoint`; the current handlers fill all output bytes, including the newly explicit padding writes in `im_point_3x8_3x8` and `im_point_2x8_2x8`; and the `L`→`I` and `I`→`L` paths write complete words. I did not establish whether the tests or valgrind exercise reads of the padding lanes, which is what would matter for detecting a future missed fill.

I can share the harness and exact build commands, or rerun particular cases if useful.

akx:
@espressolee Thanks for the comment... 

> The PR contains two independent changes

It contains three, honestly (see the commits, they're atomic and separate on purpose).
The end goal of this PR is "faster `point` and `point_transform`", and that it does, with a couple steps.

> That smaller or less certain gain comes with a permanent requirement that every present and future handler fill every output byte.

The gain from using `Dirty` (which has been a theme; see 31dba2d35897b140d1a71a52c18ffbdbc54b9e58) grows larger with larger images -- there is no point for the non-dirty handler to clear the image if the handlers will fill it anyway. Also, autovectorization tends to work much better when a compiler can look at a loop and see it will end up filling the entire surface, as it were.

As for the maintenance tradeoff, this file has not seen large changes in a while. It's unlikely we'll see new point handlers very soon - but if we do, the new contract is now clearly documented. In fact, by the looks of it, 

```
(main) $ git log --follow --format="%h %cr: %s" src/libImaging/Point.c
140d06dc1 7 weeks ago: Rename IMAGING_TYPE_SPECIAL to IMAGING_TYPE_I16
84aa4372f 1 year, 1 month ago: linter changes
9527ce7f8 1 year, 1 month ago: change mode structs to enums Structs have better type safety, but they make allocation more difficult, especially when we have multiple Python modules trying to share the same code.
7e48697f8 1 year, 1 month ago: use mode structs in Point.c
070e1eba6 1 year, 3 months ago: [pre-commit.ci] pre-commit autoupdate (#8993)
2973b041c 2 years, 2 months ago: [pre-commit.ci] auto fixes from pre-commit.com hooks
617e7295a 2 years, 4 months ago: [pre-commit.ci] auto fixes from pre-commit.com hooks
46b7e86ba 6 years ago: Format with ClangFormat
3a75e843f 6 years ago: Added braces
c1d9931ad 6 years ago: Added braces
ae43af61d 6 years ago: Replaced tabs with spaces
7a4af2b76 7 years ago: fix unaligned accesses by using memcpy()
0bb3f4fee 9 years ago: source layout reorg
bda6d1e58 9 years ago: rename ImagingCopyPaletteInfo to ImagingCopyPalette
f6a05fd22 9 years ago: Rename ImagingCopyInfo to ImagingCopyPaletteInfo
e3954aa4e 9 years ago: remove unnecessary ImagingCopyInfos
bb1b3a532 13 years ago: Cleanup WS, courtesy of @Arfrever
4ebf8e71e 17 years ago: Load Imaging-1.1.7 into pil.
56124d728 21 years ago: Load Imaging-1.1.5 into pil.
893c52a89 21 years ago: Load Imaging-1.1.4 into pil.
8ba491eac 21 years ago: Load Imaging-1.1.3 into pil.
178095672 21 years ago: pil 1.1.1 baseline
```

it's been 7 years since the last substantial change, 7a4af2b76, and before that, all we have is an SVN checkin from effbot, 4ebf8e71e, from 2009. Everything else is just "chore" stuff or refactoring side-effects (something refactored elsewhere, touching Point.c).

> These proportions are from one machine, one compiler and one OS, and I would not assume they carry to GCC, x86-64 or a different allocator.

You can refer to the CodSpeed results above for another take. If you want, you can also run the same benchmarks (instead of your own harness) on your machine:

```
# (on main)
uv pip install -e . pytest-benchmark && uv run pytest --benchmark-autosave Tests/benchmarks.py -kpoint
# (on this branch)
uv pip install -e . pytest-benchmark && uv run pytest --benchmark-autosave Tests/benchmarks.py -kpoint
```

and then use e.g. `uv run pytest-benchmark compare --between=min 0001 0002` (`--between` mode new in pytest-benchmark 5.3.0, courtesy of yours truly) to get a comparison.

espressolee (approved):
You're right: three implementation changes plus the test commit, and Point.c's history weakens my maintenance concern.

Ran `Tests/benchmarks.py -kpoint` as suggested, base `c8c74c8` (merge-base) vs head `1000b1b`, CPython 3.14.6 on macOS arm64; two separately run base/head pairs in opposite order gave deltas within about one percentage point:

| benchmark (1237x811) | ΔMedian |
| --- | ---: |
| point_transform I | −57% |
| point_lut L | −30% |
| point_transform F | −14% |
| point_lut LA, point_transform LA | −9% |
| point_lut RGBA, point_transform RGBA | −9% |
| point_lut RGB | −4.5% |

Same direction throughout; magnitudes differ by mode — notably smaller here for LA and RGB than on my harness. The correctness points from my earlier review still hold at this head. Approving.

akx:
Rebased post #10026.

PR #9744, opened 2026-06-30
Use ImagingNewDirty() in RankFilter and remove ImagingFill() in font.getmask()

akx:
Tiny performance improvement. The image is always fully filled anyway.

Related #9743's 2nd commit (same pattern).


radarhere:
I've created https://github.com/akx/Pillow/pull/8 with a suggestion - in `_imaging.c`, rather than replacing `ImagingNewDirty` with `ImagingNew`, wouldn't it be more efficient to remove the `ImagingFill` call?

radarhere (approved):

Issue #2266, opened 2016-11-30
Saving .ico saves more sizes than necessary


hugovk:
### What did you do?
```python
from PIL import Image
im = Image.open('python.ico')
im.save('saved.ico')
im_saved = Image.open('saved.ico')

print im.info['sizes']
print im_saved.info['sizes']

assert im_saved.info['sizes'] == set([(16, 16), (24, 24), (32, 32), (48, 48)])
```

### What did you expect to happen?

No assertion, and only relevant sizes saved:

```
set([(16, 16), (48, 48), (32, 32)])
set([(16, 16), (24, 24), (48, 48), (32, 32)])
```

```bash
$ identify python.ico
python.ico[0] ICO 16x16 16x16+0+0 32-bit sRGB 15.1KB 0.000u 0:00.002
python.ico[1] ICO 32x32 32x32+0+0 32-bit sRGB 15.1KB 0.000u 0:00.008
python.ico[2] ICO 48x48 48x48+0+0 32-bit sRGB 15.1KB 0.000u 0:00.013
$ identify saved.ico
saved.ico[0] PNG 16x16 16x16+0+0 8-bit sRGB 12.7KB 0.000u 0:00.003
saved.ico[1] PNG 24x24 24x24+0+0 8-bit sRGB 12.7KB 0.000u 0:00.006
saved.ico[2] PNG 32x32 32x32+0+0 8-bit sRGB 12.7KB 0.000u 0:00.007
saved.ico[3] PNG 48x48 48x48+0+0 8-bit sRGB 12.7KB 0.000u 0:00.010
```

### What actually happened?

```
set([(16, 16), (48, 48), (32, 32)])
set([(64, 64), (16, 16), (256, 256), (24, 24), (32, 32), (128, 128), (48, 48)])
Traceback (most recent call last):
  File "test.py", line 9, in <module>
    assert im_saved.info['sizes'] == set([(16, 16), (24, 24), (32, 32), (48, 48)])
AssertionError
```
```bash
$ identify python.ico
python.ico[0] ICO 16x16 16x16+0+0 32-bit sRGB 15.1KB 0.000u 0:00.002
python.ico[1] ICO 32x32 32x32+0+0 32-bit sRGB 15.1KB 0.000u 0:00.008
python.ico[2] ICO 48x48 48x48+0+0 32-bit sRGB 15.1KB 0.000u 0:00.013
$ identify saved.ico
[...]

PR #9745, opened 2026-06-30
Fix ICO save producing an empty file for images smaller than 16x16

binggao1230:
Saving an image smaller than the smallest default ICO size (16×16) as ICO, without an explicit `sizes` argument, silently produces an empty, unreadable file:

```python
import io
from PIL import Image

buf = io.BytesIO()
Image.new("RGBA", (8, 8), (10, 20, 30, 255)).save(buf, "ICO")
print(len(buf.getvalue()))          # 6  -> just the ICONDIR header, idCount=0
buf.seek(0)
Image.open(buf).load()              # UnidentifiedImageError: cannot identify image file
```

The 8×8 image is representable as an ICO (`save(buf, "ICO", sizes=[(8, 8)])` works fine), but the default-`sizes` path drops it entirely with no error or warning.

### Cause

The default `sizes` start at `(16, 16)`, and `_save` skips every candidate size larger than the source image. For an image smaller than 16×16 *every* default candidate is skipped, so `frames` stays empty and `o16(len(frames))` writes `idCount=0` — a 6-byte header-only file.

### Fix

Filter the requested sizes up front, and if none fit, fall back to the image's own size (capped at the 256×256 ICO maximum) so a valid, readable icon is always written. This also fixes the related case where every *explicitly* requested size is larger than the image (previously also an empty file) — consistent with the documented behaviour that oversized requested sizes are ignored.

### Tests

`test_save_smaller_than_default_sizes` parametrizes `(1, 1)`, `(8, 8)`, `(15, 15)` and the non-square `(8, 12)`, asserting each round-trips to a valid ICO of the right size with `info["sizes"]` and a corner pixel preserved. `test_save_all_sizes_larger_than_image` pins the explicit-`sizes` fallback. All fail on `main` with `UnidentifiedImageError` and pass with the fix; the full `Tests/test_file_ico.py` suite stays green (27 passed); `ruff format`, `ruff check` and `mypy` are clean.

### DIFF
--- a/Tests/test_file_ico.py
+++ b/Tests/test_file_ico.py
@@ -88,6 +88,43 @@ def test_save_to_bytes() -> None:
         )
 
 
+@pytest.mark.parametrize("size", ((1, 1), (8, 8), (15, 15), (8, 12)))
+def test_save_smaller_than_default_sizes(size: tuple[int, int]) -> None:
+    # An image smaller than the smallest default size (16x16) must still be
+    # saved as a valid, readable ICO rather than a header-only, empty file.
+    im = Image.new("RGBA", size, (10, 20, 30, 255))
+    im.putpixel((size[0] - 1, size[1] - 1), (200, 100, 50, 255))
+
+    output = io.BytesIO()
+    im.save(output, "ico")
+
+    output.seek(0)
+    with Image.open(output) as reloaded:
+        assert reloaded.format == "ICO"
+        assert reloaded.size == size
+        assert reloaded.info["sizes"] == {size}
+        assert reloaded.convert("RGBA").getpixel((size[0] - 1, size[1] - 1)) == (
+            200,
+            100,
+            50,
+            255,
+        )
+
+
+def test_save_all_sizes_larger_than_image() -> None:
+    # When every requested size is larger than the image, all are ignored, so
+    # fall back to the image's own size instead of writing an empty file.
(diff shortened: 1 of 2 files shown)


radarhere:
> This also fixes the related case where every explicitly requested size is larger than the image

I'm not convinced on this point. If the user requests a specific sets of sizes, and we can't provide them, I think an error should be raised, rather than silently giving the user back something other than what they requested. I think the expectation in #2266 agrees with me.

I've created #9766 as an alternative.

binggao1230:
You're right about the explicit case — if the caller names sizes we can't provide, raising beats silently handing back something else. #9766 is the better shape.

One edge case in it though: the default path raises too, when one dimension is under 16 and the other is over 256.

```python
Image.new("RGBA", (300, 8)).save("out.ico")   # ValueError: All sizes too large for image
```

`sizes` becomes `[im.size]` = `[(300, 8)]`, which the `size[0] > min(256, im.width)` filter then drops, leaving `frames` empty. Same for `(8, 300)` and `(260, 10)`. Not a regression — those write the 6-byte empty file on main today — but a plain `save()` with no `sizes` argument probably shouldn't raise.

Happy to close this in favour of #9766.

radarhere:
So you think 300x8 should save as 256x7? In that scenario, I wonder if it should also save as 128x3, 64x2 and 32x1.

radarhere:
https://github.com/python-pillow/Pillow/pull/9766 has been merged.

PR #9746, opened 2026-06-30
Fix BMP save of a P image with an empty palette

gaoflow:
Saving a `P`-mode image whose palette is empty produces a BMP that Pillow cannot reopen:

```python
import io
from PIL import Image

im = Image.new("P", (8, 8))          # empty palette, no putpalette()
im.frombytes(bytes(range(64)))
buf = io.BytesIO()
im.save(buf, "BMP")
buf.seek(0)
Image.open(buf).load()               # OSError: image file is truncated (0 bytes not processed)
```

PNG, GIF and TIFF all round-trip the same image; only BMP fails, and it fails to read back its own output.

### Cause

In `BmpImagePlugin._save`, the `P` branch computes `colors = len(palette) // 4`, which is `0` for an empty palette. The writer then sets `biClrUsed = 0` and writes no color table, with the pixel-data offset at `14 + 40 = 54`. On read, `biClrUsed == 0` at 8 bpp is interpreted as `1 << 8 = 256` entries, so the reader advances the pixel offset by `256 * 4` bytes past a table that was never written — it reads past EOF and reports the file as truncated.

### Fix

When the palette is empty, write a full 256-entry table (mirroring the existing `L` mode branch) so the header is self-consistent and the file reads back. An all-zero table is used rather than a grayscale ramp so the image reopens as `P` (a ramp would be detected as grayscale and downgraded to `L`). A populated palette is unaffected.

### Tests

`test_save_empty_palette` parametrizes `(1, 1)`, `(7, 5)`, `(8, 8)`, `(16, 16)`, asserting the written header has a non-zero `biClrUsed` (pinning the root cause, not just the symptom) and that the image reopens as `P` with its pixel indices intact. `test_save_empty_palette_round_trips_like_other_formats` pins the cross-format invariant that BMP reopens the image like PNG/GIF/TIFF do. Both fail on `main` (`OSError` / `biClrUsed == 0`) and pass with the fix; the full `Tests/test_file_bmp.py` suite stays green (37 passed); `ruff format`, `ruff check` and `mypy` are clean.

### DIFF
--- a/Tests/test_file_bmp.py
+++ b/Tests/test_file_bmp.py
@@ -69,21 +69,14 @@ def test_small_palette(tmp_path: Path) -> None:
         assert reloaded.getpalette() == colors
 
 
-def test_save_empty_palette() -> None:
-    indices = bytes(range(64))
+def test_empty_palette(tmp_path: Path) -> None:
     im = Image.new("P", (8, 8))
-    im.frombytes(indices)
 
-    output = io.BytesIO()
-    im.save(output, "BMP")
+    out = tmp_path / "temp.bmp"
+    im.save(out)
 
-    # biClrUsed, at info header offset 32
-    assert _binary.i32le(output.getvalue(), 46) == 1
-
-    output.seek(0)
-    with Image.open(output) as reloaded:
-        assert reloaded.size == (8, 8)
-        assert reloaded.tobytes() == indices
+    with Image.open(out) as reloaded:
+        assert_image_equal(im.convert("L"), reloaded)
 
 
 def test_save_too_large(tmp_path: Path) -> None:
--- a/Tests/test_file_bmp.py
+++ b/Tests/test_file_bmp.py
@@ -69,6 +69,23 @@ def test_small_palette(tmp_path: Path) -> None:
         assert reloaded.getpalette() == colors
 
 
+def test_save_empty_palette() -> None:
+    indices = bytes(range(64))
+    im = Image.new("P", (8, 8))
+    im.frombytes(indices)
+
+    output = io.BytesIO()
+    im.save(output, "BMP")
+
+    # biClrUsed, at info header offset 32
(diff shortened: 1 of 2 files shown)


Tests/test_file_bmp.py
```diff
@@ -69,6 +69,47 @@ def test_small_palette(tmp_path: Path) -> None:
         assert reloaded.getpalette() == colors
 
 
+@pytest.mark.parametrize("size", ((1, 1), (7, 5), (8, 8), (16, 16)))
```
radarhere: Why would the image size affect how the palette is handled?

radarhere:
You have assert that if you save a P mode image as a BMP, it should be re-opened as a P mode image.

BmpImagePlugin is actively trying to move to the simpler mode though.
https://github.com/python-pillow/Pillow/blob/fc7cec1751b90b253ee1059b57185f3734852fe0/src/PIL/BmpImagePlugin.py#L275-L283

I don't think that all image formats need to behave exactly the same. The PNG format actually has a ['Color type'](https://www.libpng.org/pub/png/spec/1.2/PNG-Chunks.html) that distinguishes whether the data is grayscale or palette. The BMP format doesn't have that information.

So for a P mode image with an empty palette, I would be inclined to save only 1 palette entry, so that the final file is as small as possible.

Tests/test_file_bmp.py
binggao1230: It doesn't — that was pointless, dropped the parametrize.

binggao1230:
Fair point on both. Switched to a single palette entry — 122 bytes now instead of 1142 — and dropped the mode assertion and the cross-format test, since BMP has no way to record the grayscale/palette distinction and the reader is right to simplify. It reopens as `L` with the pixel indices intact.

The test now just pins `biClrUsed == 1` and the round-trip.

radarhere (approved):

PR #9749, opened 2026-07-01
Do not use CodSpeedHQ/action on fork

radarhere:
I've noticed that codspeed is failing in my fork - https://github.com/radarhere/Pillow/actions/runs/28502089873/job/84481570060
```
Uploading results
  CodSpeed Run Hash: "ac9229155c02b8c001a1368595350a49a27a52b738385214d2cddf9c336dacb4"
  Error: Failed to retrieve upload data: 401 Unauthorized
    -> Reason: Repository not found or the user does not have access to it.
  
  Check that the workflow is correctly authenticated. View more at https://codspeed.io/docs/integrations/ci/github-actions/configuration#authentication
```

There doesn't appear to be any way to instruct CodSpeedHQ/action not to try and upload the results, so this runs pytest with codspeed normally in a separate step on forks.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -93,4 +93,4 @@ jobs:
         mode: simulation
         run: |
           python3 -m pip install -e . pytest-codspeed
-          pytest -vv --codspeed Tests/benchmarks.py
+          python3 -m pytest -vv --codspeed Tests/benchmarks.py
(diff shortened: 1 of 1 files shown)


hugovk:
I've enabled CodSpeed for my fork so it can upload.

Is disabling the workflow acceptable?

image

Or would you prefer to have it run (and not fail uploading)?

radarhere:
Oh, ok, I presumed that it would only upload to Pillow's main repository.

PR #9752, opened 2026-07-01
Document the internal image distinctness constraints

akx:
Follows up on #9649, #9675, #9736, #9737, #9738, #9739, #9740.

#9743 will do this internally.

Adds documentation comments to the touched functions to document the distinctness-or-lack-thereof requirements for the image data.

### DIFF
--- a/src/libImaging/Bands.c
+++ b/src/libImaging/Bands.c
@@ -17,6 +17,10 @@
 
 #include "Imaging.h"
 
+/**
+ * Extract a single band from imIn into a newly allocated single-band image.
+ * Contract: imIn is read-only and the returned image is a distinct allocation.
+ */
 Imaging
 ImagingGetBand(Imaging imIn, int band) {
     Imaging imOut;
@@ -68,6 +72,14 @@ ImagingGetBand(Imaging imIn, int band) {
     return imOut;
 }
 
+/**
+ * Split imIn into its component bands.
+ * The caller must provide an array of 4 Imaging pointers,
+ * which will be allocated and filled with the individual bands.
+ * The number of bands returned is the number of bands in imIn.
+ *
+ * Contract: imIn is read-only.
+ */
 int
 ImagingSplit(Imaging imIn, Imaging bands[4]) {
     int i, j, x, y;
@@ -173,6 +185,11 @@ ImagingSplit(Imaging imIn, Imaging bands[4]) {
     return imIn->bands;
 }
 
+/**
+ * Insert single-band imIn into `band` of imOut, in place.
+ *
+ * Contract: imIn and imOut MUST be distinct images and not alias.
+ */
 Imaging
 ImagingPutBand(Imaging imOut, Imaging imIn, int band) {
     int x, y;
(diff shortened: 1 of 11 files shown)


src/libImaging/Bands.c
```diff
@@ -68,6 +72,14 @@ ImagingGetBand(Imaging imIn, int band) {
     return imOut;
 }
 
+/**
+ * Split imIn into its component bands.
+ * The caller must provide an array of 4 Imaging pointers,
+ * which will be allocated and filled with the individual bands.
+ * The number of bands returned is the number of bands in imIn.
```
radarhere:
```suggestion
 * The caller must provide an array of 4 null Imaging pointers.
 * Some of these will be allocated and filled with the individual bands,
 * up to the number of bands in imIn.
 * The number returned is the number of bands in imIn.
```
Not all of the incoming pointers will be allocated.

src/libImaging/Bands.c
```diff
@@ -173,6 +185,11 @@ ImagingSplit(Imaging imIn, Imaging bands[4]) {
     return imIn->bands;
 }
 
+/**
+ * Insert single-band imIn into `band` of imOut, in place.
+ *
+ * Contract: imIn and imOut MUST be distinct images and not alias.
```
radarhere:
```suggestion
 * Contract: imIn and imOut MUST be distinct images and not alias each other.
```
Minor wording suggestion.

akx:
Thanks, good points! I noticed #9988 would invalidate some of the comments added here - would you prefer this be merged first or that API change and then this? I'm fine rebasing/fixing either way.

src/libImaging/Histo.c
```diff
@@ -56,6 +56,11 @@ ImagingHistogramNew(Imaging im) {
     return h;
 }
 
+/**
+ * Accumulate a histogram over `im`, optionally restricted to imMask.
```
radarhere: 'accumulate' makes it sound like this is a cumulative histogram, which it isn't?

src/libImaging/Histo.c
```diff
@@ -56,6 +56,11 @@ ImagingHistogramNew(Imaging im) {
...
+ *
+ * Contract: Both im and imMask are read-only.
```
radarhere, line 63: Why is this the contract? I expect that is how it is used, but I don't see any use of `restrict` in here.

src/libImaging/Quant.c
```diff
@@ -1663,6 +1663,12 @@ quantize2(
     return 0;
 }
 
+/**
+ * Quantize `im` down to at most `colors` palette entries,
+ * returning a newly allocated result.
+ *
+ * Contract: im is read-only.
```
radarhere, line 1670: Again, I don't see any `restrict` here.

src/libImaging/Histo.c
akx: I didn't get this far in the optimization series, looks like 😄
But this documents the contract nevertheless, that was never going to change with the series.

**EDIT:** See #10005. Turns out I had half-written optimizations but didn't push them before.

src/libImaging/Quant.c
akx: Same here as above: this still documents the contract. Quantize was touched a bit in #9740, though not with `restrict` bits.

akx:
Thanks!

PR #9753, opened 2026-07-01
Speed up `Image.offset`

akx:
As https://github.com/python-pillow/Pillow/pull/9649, https://github.com/python-pillow/Pillow/pull/9675, https://github.com/python-pillow/Pillow/pull/9736, https://github.com/python-pillow/Pillow/pull/9737, https://github.com/python-pillow/Pillow/pull/9738, https://github.com/python-pillow/Pillow/pull/9739, https://github.com/python-pillow/Pillow/pull/9740, #9743 (same techniques).

### DIFF
--- a/src/libImaging/Offset.c
+++ b/src/libImaging/Offset.c
@@ -16,6 +16,12 @@
 
 #include "Imaging.h"
 
+/**
+ * Copy `im` into a newly allocated image,
+ * wrapping every pixel by (xoffset, yoffset) modulo the image size.
+ *
+ * Contract: im is read-only.
+ */
 Imaging
 ImagingOffset(Imaging im, int xoffset, int yoffset) {
     int x, y;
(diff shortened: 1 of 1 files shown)


src/libImaging/Offset.c
```diff
@@ -45,19 +51,23 @@ ImagingOffset(Imaging im, int xoffset, int yoffset) {
...
-            int xi = (x + xoffset) % im->xsize;     \
-            imOut->image[y][x] = im->image[yi][xi]; \
-        }                                           \
+    // yi depends only on y, so compute it (and both row pointers) once per
+    // row instead of redoing the modulo and pointer chase for every x.
+#define OFFSET(type, image)                     \
+    for (y = 0; y < im->ysize; y++) {           \
+        int yi = (y + yoffset) % im->ysize;     \
+        type *restrict out = imOut->image[y];   \
+        type *restrict in = im->image[yi];      \
```
radarhere:
```suggestion
        type *restrict in = im->image[yi];      \
        type *restrict out = imOut->image[y];   \
```
It might be nicer to have `in` before `out`?

hugovk:

hugovk (approved):
This is an improvement, although there's a pre-existing bug on `main` for 16-bit images that this doesn't fix (and doesn't need to, as a performance refactor):

```python
"""ImageChops.offset corrupts I;16 images: ImagingOffset copies image8 rows
as 1 byte/pixel, but I;16 has 2-byte pixels — half of each row is left as
uninitialized memory. Even offset (0, 0) doesn't round-trip.
Other modes (e.g. "I") are unaffected."""

from PIL import Image, ImageChops

for mode in ("I", "I;16"):
    im = Image.new(mode, (127, 61))
    im.putdata([(i * 13 + 7) % 65536 for i in range(127 * 61)])

    out = ImageChops.offset(im, 0, 0)
    ok = out.tobytes() == im.tobytes()
    print(f"{mode:5} offset(0, 0) == original:", ok)  # expect True
```
Produces:
```
I     offset(0, 0) == original: True
I;16  offset(0, 0) == original: False
```
Could be something for a followup.

akx:
@hugovk Good catch! I can open a PR to fix that if you like.

hugovk:
Sure, go for it!

PR #9755, opened 2026-07-02
Avoid divide-by-zero crash in RankFilter

lukegb:

 * Avoid divide-by-zero crash in image filtering - although a kernel size of 1 is pointless/trivial, but it shouldn't cause the C extension to crash.

Fixes a crash accidentally introduced by the validation logic added in dff01a0821c1f1cc8fa35cee5cc08e86399aac7c.


radarhere:
I've created https://github.com/lukegb/Pillow/pull/3 with a suggestion - rather than expand the image by zero, I think just copying the image would be more efficient.

You probably want to keep `margin > 0` for defensiveness.

radarhere (approved):

PR #9760, opened 2026-07-03
Speed up `Image.filter()` for `I;16*`

akx:
Follows up on #8438; same split-by-mode optimization technique as in #9737. 

Also fixes up things around `ImageFilter`:

* the tests' modes list was N-plicated
* the `test_consistency_*` tests didn't use the `mode` parametrization at all and just ran the same test N times
* `test_crash` claimed ImageFilter would crash on images smaller than the kernel; it doesn't, it just returns the same image. 
  * Looks like `test_crash` had had `assert_no_exception` sometime before 3ec505958ed6225c7d0d6481bf4a6bad5d2a3324?
* The docstring for `Kernel` claimed only a couple of modes were supported.

The correctness bugs from #9786 are fixed here too.


akx:
Hmmh, why would the hash tests fail differently on different Python versions... 😞

Tests/test_image_filter.py
```diff
@@ -172,42 +170,67 @@ def test_kernel_not_enough_coefficients() -> None:
...
+    kernel = ImageFilter.Kernel(size, weights, 0.3)
+    reference = hopper("I").filter(kernel)
+    result = hopper(mode).filter(kernel)
+    assert result.mode == mode
+    assert result.size == reference.size
+    # Compare logical pixel values rather than raw bytes
+    # so the test is endianness-independent.
+    ref = list(reference.get_flattened_data())
+    res = list(result.get_flattened_data())
+    assert max(abs(float(a) - float(b)) for a, b in zip(res, ref)) <= 1
```
akx:
```suggestion
    assert max(abs(float(a) - float(b)) for a, b in zip(res, ref)) <= 1  # type: ignore
```

akx:
Draft until #9786 is merged so this is clean to rebase on it.

hugovk:
Updated from `main` following benchmark image update: https://github.com/python-pillow/Pillow/pull/9831.

src/libImaging/Filter.c
```diff
@@ -161,17 +161,16 @@ ImagingFilter3x3(Imaging imOut, Imaging im, const float *kernel, float offset) {
                 }
                 out[x] = in0[x];
             }
-        } else {
+        } else if (im->type == IMAGING_TYPE_SPECIAL) {
+            // Check for I;16 mode once, not per pixel
```
radarhere:
```suggestion
```
Once this is merged, I don't think it's obvious what the comment is referring to

src/libImaging/Filter.c
```diff
@@ -322,17 +325,16 @@ ImagingFilter5x5(Imaging imOut, Imaging im, const float *kernel, float offset) {
                 out[x + 0] = in0[x + 0];
                 out[x + 1] = in0[x + 1];
             }
-        } else {
+        } else if (im->type == IMAGING_TYPE_SPECIAL) {
+            // Check for I;16 mode once, not per pixel
```
radarhere:
```suggestion
```

Tests/test_image_filter.py
```diff
@@ -53,20 +53,18 @@ def test_sanity(
...
 
 
-# crashes on small images
 @pytest.mark.parametrize("size", ((1, 1), (2, 2), (3, 3)))
-def test_crash(size: tuple[int, int]) -> None:
+def test_noop_on_small_images(size: tuple[int, int]) -> None:
+    # If image is smaller than kernel size, return it as-is
     im = Image.new("RGB", size)
-    im.filter(ImageFilter.SMOOTH)
+    assert_image_equal(im, im.filter(ImageFilter.SMOOTH))
```
```suggestion
    assert_image_equal(im.filter(ImageFilter.SMOOTH), im)
```

Tests/test_image_filter.py
```diff
@@ -8,6 +8,8 @@
 
 from .helper import assert_image_equal, hopper
 
+MODES = ("L", "LA", "I", "I;16", "I;16L", "I;16B", "I;16N", "RGB", "CMYK")
```
```suggestion
MODES = ("L", "I", "I;16", "I;16L", "I;16B", "I;16N", "RGB", "CMYK")
```
LA was only in the mode list for `test_consistency_3x3` and `test_consistency_5x5`, and they don't use modes.

Tests/test_image_filter.py
```diff
@@ -174,38 +172,43 @@ def test_kernel_not_enough_coefficients() -> None:
...
+@pytest.mark.parametrize("mode", ("I;16", "I;16L", "I;16B", "I;16N"))
+def test_consistency_i16(size: int, mode: str) -> None:
+    kernel = ImageFilter.Kernel((size, size), EMBOSS_MATRIX[size], 0.3)
+    reference = hopper("I").filter(kernel)
+    result = hopper(mode).filter(kernel)
+    assert result.mode == mode
+    assert result.size == reference.size
+    for x in range(size):
+        for y in range(size):
+            assert result.getpixel((x, y)) == reference.getpixel((x, y))
```
```suggestion
    assert_image_equal(result.convert("I"), reference)
```

Tests/test_image_filter.py
```diff
@@ -53,20 +53,18 @@ def test_sanity(
...
     im = hopper(mode)
     with pytest.raises(TypeError):
         im.filter("hello")  # type: ignore[arg-type]
 
 
-# crashes on small images
 @pytest.mark.parametrize("size", ((1, 1), (2, 2), (3, 3)))
-def test_crash(size: tuple[int, int]) -> None:
+def test_noop_on_small_images(size: tuple[int, int]) -> None:
+    # If image is smaller than kernel size, return it as-is
```
```suggestion
    # If image is not larger than the kernel size, return it as-is
```
The kernel is 3x3, and this is testing a (3, 3) image in one instance.

However, this doesn't seem to be an effective test for small image sizes.
```python
@pytest.mark.parametrize("size", ((400, 400),))
```
also passes.

Tests/test_image_filter.py
akx: `test_sanity` and `test_sanity_error` are parametrized over this `MODES`. I wouldn't like to reduce mode coverage there.

Tests/test_image_filter.py
akx: Mm, why?

At a quick glance, there are both "styles" of parameter orders (original-first, modified-first) for `assert_image_equal()` around Pillow's tests...

Tests/test_image_filter.py
akx: Yeah, `SMOOTH`, a 3x3 filter kernel, works fine on a 400x400 image, as expected. Smoothing an image of all zeroes results in very smooth all zeroes, so `assert_image_equal` would pass there.

"not larger" is more accurate though!

Tests/test_image_filter.py
akx: Sure. Thinking to myself: This would need adjustment if converting from I16 to I8 involved scaling, but it doesn't.

Tests/test_image_filter.py
radarhere: I acknowledge that we're not consistent on this, but I think expected-last is the more common one. It's a minor comment though.

Tests/test_image_filter.py
radarhere: LA wasn't being tested before this PR. If you'd like to add LA, then why not other modes as well? https://github.com/akx/Pillow/pull/26

Tests/test_image_filter.py
radarhere: Before this PR, it was `# crashes on small images / test_crash`. The test comes from the earliest Pillow commits covering the entire test suite, so there's no specifics about why it was added. I presume it was meant to be in past tense.

Your version is saying that 'If image is not larger than the kernel size, return it as-is'. However, that doesn't seem accurate - if I use `hopper()` data [instead](https://github.com/radarhere/Pillow/commit/24f577608e1297809f69c2b948f23fb92df3e693), then it is not returned as-is - https://github.com/radarhere/Pillow/actions/runs/34095868681/job/101659281231#step:12:3427

akx: Yeah, `hopper`, being larger than 3x3, can be filtered by a 3x3 filter and will change.

https://github.com/hugovk/pil-svn.effbot.org/commit/4ebf8e71e8c5677de8ded2d88ab48088a0f8b04e ("Load Imaging-1.1.7 into PIL", 2009-11-15) added 

```c
    if (im->xsize < xsize || im->ysize < ysize) {
        return ImagingCopy(im);
    }
```
in `ImagingFilter`, making a filter operation for an image smaller than the kernel a no-op.

The original form of that test was apparently added in ad784eb8080b653ed9a28b4d03195d379b2d0538 as 

```python
def test_crash():

    # crashes on small images    
    im = Image.new("RGB", (1, 1))
    assert_no_exception(lambda: im.filter(ImageFilter.SMOOTH))

    im = Image.new("RGB", (2, 2))
    assert_no_exception(lambda: im.filter(ImageFilter.SMOOTH))

    im = Image.new("RGB", (3, 3))
    assert_no_exception(lambda: im.filter(ImageFilter.SMOOTH))
```

so I'd guess the test name and comment had really been inaccurate (or, as you said, in past tense) all along? 

https://github.com/python-pillow/Pillow/pull/9760/changes/57c152dfbc2e9f11837a9ba8c852af60cf6ce6b0 here just changed the test name to be a little more accurate... Would you like an even more accurate name, or a better docstring/comment here? `test_images_smaller_than_kernel_noop`?

radarhere: > Yeah, `hopper`, being larger than 3x3

In my demo, I'm not filtering `hopper()`, I'm filtering `hopper().resize(size)`

akx: Ah, sorry, didn't look. Yep. Adjusted the test once more, now to use `SMOOTH_MORE` (5x5), and to use Hopper at all sizes smaller than the filter kernel.

Tests/test_image_filter.py
```diff
@@ -53,20 +67,24 @@ def test_sanity(
...
-@pytest.mark.parametrize("size", ((1, 1), (2, 2), (3, 3)))
-def test_crash(size: tuple[int, int]) -> None:
-    im = Image.new("RGB", size)
-    im.filter(ImageFilter.SMOOTH)
+def test_noop_on_small_images() -> None:
+    # If image is smaller than the kernel size, return it as-is.
+    kernel_size: tuple[int, int] = ImageFilter.SMOOTH_MORE.filterargs[0]
+    kernel_w, kernel_h = kernel_size
+    for w in range(1, kernel_w - 1):
+        for h in range(1, kernel_h - 1):
```
```suggestion
    for w in range(1, kernel_w):
        for h in range(1, kernel_h):
```
Python ranges stop before the second number, so everything within will still be smaller than the kernel.

src/libImaging/Filter.c
radarhere: You marked this, and a similar comment, as resolved without any reply. Did you apply this change originally, only for it to be lost in a later rebase?

radarhere (approved):

src/libImaging/Filter.c
akx: That could have well been the case, sorry!

Tests/test_image_filter.py
akx: I know - I'm pretty sure I had test failures with that, but evidently... not. 😵‍💫 

Thanks!

PR #9761, opened 2026-07-03
Do not use CodSpeedHQ/action on pull requests to a fork

radarhere:
Second version of https://github.com/python-pillow/Pillow/pull/9749

While we can configure the Benchmarks workflow in our own repositories, perhaps it is not ideal that it fails when pull requests are made to a fork? For example, https://github.com/akx/Pillow/pull/8. The receiving user might believe that the changes in my pull request are causing this workflow to fail.


.github/workflows/benchmark.yml
```diff
@@ -88,9 +88,16 @@ jobs:
...
         run: |
           python3 -m pip install -e . pytest-codspeed
-          pytest -vv --codspeed Tests/benchmarks.py
+          python3 -m pytest -vv --codspeed Tests/benchmarks.py
+
+    - name: Run CodSpeed benchmarks on pull requests to a fork
+      if: github.event_name == 'pull_request' && github.event.repository.fork == true
+      run: |
+        python3 -m pip install -e . pytest-codspeed
+        python3 -m pytest -vv --codspeed Tests/benchmarks.py
```
akx, line 103: I'm not sure this alternate is really needed. If you're not using the codspeed action (that does CPU simulation and other deep magic), you're not going to get comparable numbers anyhow.
```suggestion
```

.github/workflows/benchmark.yml
```diff
@@ -88,9 +88,16 @@ jobs:
         GHA_LIBWEBP_CACHE_HIT: ${{ steps.cache-libwebp.outputs.cache-hit }}
 
     - name: Run CodSpeed benchmarks
+      if: github.event_name != 'pull_request' || github.event.repository.fork == false
```
akx, line 91: Maybe simply
```suggestion
      if: github.event.repository.fork == false
```
in conjunction with the below (see rationale there).

WDYT?

akx (changes requested):
I'd be fine with just running Codspeed on pull requests to or pushes on python-pillow/Pillow?

radarhere:
> I'd be fine with just running Codspeed on pull requests to or pushes on python-pillow/Pillow?

This seems like the same thinking as https://github.com/python-pillow/Pillow/pull/9946?

akx:
> This seems like the same thinking as #9946?

Not really... looks like the CodSpeed action doesn't report anything on fork runs, see e.g. https://github.com/akx/Pillow/actions/runs/33734366520/job/100581356953, it just churns through running the benchmarks and then crashes and fails the run.

hugovk:
Depends if you've enabled CodSpeed on your fork, compare:

https://github.com/hugovk/Pillow/actions/runs/33500798014/job/99833423587

Can we have it not fail the run when it's not enabled?

akx:
> Can we have it not fail the run when it's not enabled?

I can suggest a PR for the action, an input where we could feed `${{ github.event.repository.fork == false }}` here.

But if it doesn't otherwise report the benchmark results, then it's just doing 20+ minutes of work for every push just to drop the results on the floor. (I guess it could maybe do a GitHub [Step Summary](https://github.blog/news-insights/product-news/supercharging-github-actions-with-job-summaries/) instead...)


**EDIT:** _Tuumasta toimeen_, as they say.

* https://github.com/CodSpeedHQ/action/pull/237
* https://github.com/CodSpeedHQ/codspeed/pull/527

radarhere:
Thanks very much for those PRs. We'll see how they go.

PR #9762, opened 2026-07-03
Speed up `Image.getextrema()`

akx:
Speeds up `getextrema()` -- with _more_ than just hoist and restrict this time, by adding a separate function for the common case of 2-4 band 8bpc images. The single-channel case speeds up pleasantly too by autovectorisation, though. 

```
------------------- benchmark 'extrema': 4 tests, 2 sources --------------------
Name (time in us)                   0001_3a36c3a OPS  0003_796ae94 OPS      ΔOPS
--------------------------------------------------------------------------------
test_getextrema[1024x1024-RGB]            1,790.4542       13,099.6057   +631.6%
test_getextrema[1024x1024-L]              1,410.9957       17,641.2525  +1150.3%
test_getextrema[1024x1024-RGBA]             687.5769       13,102.1316  +1805.6%
test_getextrema[1024x1024-LA]               494.9731       11,851.6516  +2294.4%
--------------------------------------------------------------------------------
```

### DIFF
--- a/src/_imaging.c
+++ b/src/_imaging.c
@@ -2337,15 +2337,39 @@ _getcolors(ImagingObject *self, PyObject *args) {
 
 static PyObject *
 _getextrema(ImagingObject *self, PyObject *args) {
+    if (self->image->type == IMAGING_TYPE_UINT8 && self->image->bands > 1 &&
+        self->image->bands <= 4) {
+        // Fast per-band extrema for 8-bit multiband images (RGB, RGBA, etc.)
+        UINT8 mb[2 * 4];
+        int bands = self->image->bands;
+        int nb = ImagingGetExtremaMultiband(self->image, mb);
+        if (nb < 0) {
+            return NULL;
+        }
+        PyObject *result = PyTuple_New(bands);
+        if (!result) {
+            return NULL;
+        }
+        for (int b = 0; b < bands; b++) {
+            // nb == 0 for an empty image: report None per band.
+            PyObject *item =
+                nb ? Py_BuildValue("BB", mb[b * 2], mb[b * 2 + 1]) : Py_NewRef(Py_None);
+            if (!item) {
+                Py_DECREF(result);
+                return NULL;
+            }
+            PyTuple_SET_ITEM(result, b, item);
+        }
+        return result;
+    }
+
     union {
         UINT8 u[2];
         INT32 i[2];
         FLOAT32 f[2];
         UINT16 s[2];
     } extrema;
-    int status;
-
-    status = ImagingGetExtrema(self->image, &extrema);
+    int status = ImagingGetExtrema(self->image, &extrema);
     if (status < 0) {
         return NULL;
     }
--- a/src/_imaging.c
+++ b/src/_imaging.c
@@ -2340,15 +2340,39 @@ _getcolors(ImagingObject *self, PyObject *args) {
 
 static PyObject *
 _getextrema(ImagingObject *self, PyObject *args) {
+    if (self->image->type == IMAGING_TYPE_UINT8 && self->image->bands > 1 &&
+        self->image->bands <= 4) {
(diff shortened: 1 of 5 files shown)


src/libImaging/Imaging.h
```diff
@@ -354,6 +354,8 @@ ImagingGetColors(Imaging im, int maxcolors, int *colors);
 extern int
 ImagingGetExtrema(Imaging im, void *extrema);
 extern int
+ImagingGetExtremaMultiband(Imaging im, UINT8 extrema[const 8]);
```
akx:
```suggestion
ImagingGetExtremaMultiband(Imaging im, UINT8 extrema[8]);
```

src/libImaging/GetBBox.c
```diff
@@ -248,6 +244,57 @@ ImagingGetExtrema(Imaging im, void *extrema) {
     return 1; /* ok */
 }
 
+int
+ImagingGetExtremaMultiband(Imaging im, UINT8 extrema[const 8]) {
```
akx:
```suggestion
ImagingGetExtremaMultiband(Imaging im, UINT8 extrema[8]) {
```

akx (commented):
(self-review so I can apply a suggested change, not at my usual machine)

src/_imaging.c
```diff
@@ -2321,15 +2321,39 @@ _getcolors(ImagingObject *self, PyObject *args) {
 
 static PyObject *
 _getextrema(ImagingObject *self, PyObject *args) {
+    if (self->image->type == IMAGING_TYPE_UINT8 && self->image->bands > 1 &&
+        self->image->bands <= 4) {
```
radarhere, line 2344: Considering that we don't have any images with more than 4 bands, I'm not sold that we need to check for it.

akx: If this check is elided, in the future when someone decides to add such support, the following code will crash in unfortunate and non-obvious ways when it writes over the `UINT8 mb[2 * 4];` array, where `4` stands for "maximum number of bands supported by this code".

I can remove the check if you like, but I don't think it costs much here.

src/_imaging.c
```diff
@@ -2321,15 +2321,39 @@ _getcolors(ImagingObject *self, PyObject *args) {
...
+        if (!result) {
+            return NULL;
+        }
+        for (int b = 0; b < bands; b++) {
+            // nb == 0 for an empty image: report None per band.
+            PyObject *item =
+                nb ? Py_BuildValue("BB", mb[b * 2], mb[b * 2 + 1]) : Py_NewRef(Py_None);
+            if (!item) {
+                Py_DECREF(result);
+                return NULL;
```
radarhere, line 2362: Shouldn't the `item`s in the previous bands be decref'd here as well?

akx: I don't think so...

[`PyTuple_SET_ITEM`](https://docs.python.org/3.14/c-api/tuple.html#c.PyTuple_SET_ITEM):
> This function “steals” a reference to o, and, unlike PyTuple_SetItem(), does not discard a reference to any item that is being replaced; any reference in the tuple at position pos will be leaked.

IOW, the references to previous `item`s had been already stolen into `result`.

src/_imaging.c
radarhere: Couldn't you just allocate `mb` to match the number of bands?

akx: Using a VLA (or heap allocation) becomes a little complex, since due to the data layout of 2-band (PA/LA/La) images (pixelsize 4, `XX....AA` in image32), the actual valid sizes for `mb` (and `ImagingGetExtremaMultiband`'s `vmin` and `vmax`) are actually only 3 or 4. IOW, the required accumulator array size is not directly related to `bands`.

I'm not sure it's worth the complexity, what do you think?

PR #9764, opened 2026-07-04
Remove minimumReleaseAge for google/oss-fuzz

radarhere:
It is unlikely that there will be a whole week where [google/oss-fuzz](https://github.com/google/oss-fuzz) doesn't receive a new commit, meaning that the [minimum release age](https://github.com/python-pillow/Pillow/blob/31dba2d35897b140d1a71a52c18ffbdbc54b9e58/.github/renovate.json#L9) for Renovate is unlikely to ever pass.

Rather than manually triggering the 'github-actions' item in https://github.com/python-pillow/Pillow/issues/6604 each month, this PR attempts to remove the minimum release age, just for GitHub Actions.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -171,6 +171,11 @@
             "groupName": "github-actions",
             "matchManagers": ["github-actions"],
             "separateMajorMinor": false
+        },
+        {
+            "matchManagers": ["github-actions"],
+            "matchPackageNames": ["google/oss-fuzz"],
+            "minimumReleaseAge": "0 days"
         }
     ]
 }


hugovk:
Rather than setting it to zero for all actions, can we set it to zero for just `google/oss-fuzz`?

Then we get extra protection from supply chain attacks, which are becoming more common against GitHub Actions:

https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns

Perhaps like:

```json
    "packageRules": [
      {
          "groupName": "github-actions",
          "matchManagers": ["github-actions"],
          "separateMajorMinor": false
      },
      {
          "matchManagers": ["github-actions"],
          "matchPackageNames": ["google/oss-fuzz**"],
          "minimumReleaseAge": "0 days"
      }
    ]
```

radarhere:
Why "google/oss-fuzz**" rather than "google/oss-fuzz"?

hugovk:
`"google/oss-fuzz"` is better 👍

hugovk (approved):

PR #9767, opened 2026-07-04
Remove unused `*_ROOT` globals from `setup.py`

hugovk:
And simplify library root detection in `build_extensions()`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,18 +43,6 @@ def get_version() -> str:
 
 
 PILLOW_VERSION = get_version()
-AVIF_ROOT = None
-FREETYPE_ROOT = None
-HARFBUZZ_ROOT = None
-FRIBIDI_ROOT = None
-IMAGEQUANT_ROOT = None
-JPEG2K_ROOT = None
-JPEG_ROOT = None
-LCMS_ROOT = None
-RAQM_ROOT = None
-TIFF_ROOT = None
-WEBP_ROOT = None
-ZLIB_ROOT = None
 FUZZING_BUILD = "LIB_FUZZING_ENGINE" in os.environ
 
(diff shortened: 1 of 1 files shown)


radarhere:
In PIL, theses looked like
```python
# --------------------------------------------------------------------
# Library pointers.
#
# Use None to look for the libraries in well-known library locations.
# Use a string to specify a single directory, for both the library and
# the include files.  Use a tuple to specify separate directories:
# (libpath, includepath).  Examples:
#
# JPEG_ROOT = "/home/libraries/jpeg-6b"
# TIFF_ROOT = "/opt/tiff/lib", "/opt/tiff/include"
#
# If you have "lib" and "include" directories under a common parent,
# you can use the "libinclude" helper:
#
# TIFF_ROOT = libinclude("/opt/tiff")

TCL_ROOT = None
JPEG_ROOT = None
ZLIB_ROOT = None
TIFF_ROOT = None
FREETYPE_ROOT = None
LCMS_ROOT = None
```

I think the expectation was for people building from source to patch these. https://github.com/conda-forge/pillow-feedstock/blob/main/recipe/patches/0003-add-XCB_ROOT.patch sort of does this.

You are asserting though that we have better mechanisms now, like https://github.com/python-pillow/Pillow/pull/3395/changes/ddb01ac67e7052ccdf5f66df6858ba38eff4cf8e, yes?

hugovk:
Hmm, those comments have been gone since 2010.

That patch sets a global `XCB_ROOT = None`, not an actual value. Just a copy/paste of the neighbouring code.

But I found some use of them here:

https://github.com/NixOS/nixpkgs/blob/5ee04f2d21cb7907546b70ce94c4ae1d5529efab/pkgs/development/python-modules/pillow/default.nix#L89-L91

Let's leave this be, then.

PR #9750, opened 2026-07-01
Drop support for Python 3.10

hugovk:
Python 3.10 reaches end of life in October 2026:

* https://peps.python.org/pep-0619/

Pillow 13.0.0 is planned for 2026-10-15 and will drop Python 3.10:

* https://github.com/python-pillow/Pillow/milestone/66

TODO:

* [x] Rebase after https://github.com/python-pillow/Pillow/pull/9751
* [x] Update Python to 3.11 in CentOS Stream 9: [python-pillow/docker-images#274](https://github.com/python-pillow/docker-images/pull/274)
* [x] Decide about `ubuntu-22.04-jammy-amd64` and `ubuntu-22.04-jammy-amd64-valgrind` which use 3.10: upgrade to 3.11 or drop

PR #9771, opened 2026-07-05
Port coveragerc to pyproject.toml; omit benchmarks from coverage

akx:
As it says in the commits and on the tin; I noticed Codecov (rightly) said `benchmarks.py` is at zero coverage.

While at it, ported `.coveragerc` to the `pyproject.toml` format for one less loose config file in the repo root.


pyproject.toml
```diff
@@ -204,3 +204,25 @@ addopts = [ "-ra", "--color=auto" ]
...
+
+[tool.coverage]
+run.omit = [
+  "checks/*.py",
+  "Tests/benchmarks.py",
+  "Tests/createfontdatachunk.py",
+]
+report.exclude_also = [
+  "@abc.abstractmethod",
+  "^\\s*\\.\\.\\.(\\s*#.*)?$",
```
hugovk: We can remove this, ellipses are now excluded by default:
https://coverage.readthedocs.io/en/latest/excluding.html#default-exclusions
```suggestion
```

pyproject.toml
```diff
@@ -204,3 +204,25 @@ addopts = [ "-ra", "--color=auto" ]
...
+  "^\\s*def [a-zA-Z0-9_]+\\(.*\\)(\\s*->.*)?:\\s*\\.\\.\\.(\\s*#.*)?$",
+  # Don't complain about compatibility code for missing optional dependencies
+  "except ImportError",
+  "if __name__ == .__main__.:",
+  # Don't complain if non-runnable code isn't run
+  "if 0:",
+  # Don't complain about debug code
+  "if DEBUG:",
+  # Don't complain about code that generally isn't run
+  "if TYPE_CHECKING:",
```
hugovk: And these:
```suggestion
  # Don't complain about debug code
  "if DEBUG:",
```

pyproject.toml
```diff
@@ -204,3 +204,25 @@ addopts = [ "-ra", "--color=auto" ]
 testpaths = [
   "Tests",
 ]
+
+[tool.coverage]
```
hugovk, line 208: We can also enable the faster sysmon core where available, and disable the warning where not available.

```suggestion
[tool.coverage]
run.core = "sysmon"  # default for 3.14+, available for 3.12+
run.disable_warnings = [ "no-sysmon" ]  # 3.11 and earlier
```

And then we can remove `COVERAGE_CORE: sysmon` from the GitHub workflows.

hugovk (commented):
Turns out I had a branch to do this earlier this year (except of course not the `benchmarks.py` bit), but never got round to opening the PR :)

Here's the other things I was going to include.

pyproject.toml
```diff
@@ -204,3 +204,22 @@ addopts = [ "-ra", "--color=auto" ]
...
+run.core = "sysmon"  # default for 3.14+, available for 3.12+
+run.disable_warnings = [ "no-sysmon" ]  # 3.11 and earlier
```
radarhere:
```suggestion
run.disable_warnings = [ "no-sysmon" ]  # 3.11
```
Python 3.11 will soon be the minimum. See https://github.com/python-pillow/Pillow/pull/9750

hugovk:
Thanks!

PR #9773, opened 2026-07-06
Mark compilation-unit-internal symbols `static`

akx:
This should allow compilers to better inline those functions.

Along the way, this:

* deletes `ImageAccessDelete` (no-op since [2009](https://github.com/hugovk/pil-svn.effbot.org/commit/4ebf8e71e8c5677de8ded2d88ab48088a0f8b04e))
* deletes `debug_clip_tree` (added in 9a9d3a050a3e579d579c61549618d297a1e837a8, 2020). Since it wasn't used outside of `Draw.c`, it would've been marked `static`, and at that point become entirely dead code. In addition, it would not have compiled in its current form anyway, since `fprintf()` and friends aren't around

### DIFF
--- a/src/_imaging.c
+++ b/src/_imaging.c
@@ -180,10 +180,6 @@ _dealloc(ImagingObject *imagep) {
 #ifdef VERBOSE
     printf("imaging %p deleted\n", imagep);
 #endif
-
-    if (imagep->access) {
-        ImagingAccessDelete(imagep->image, imagep->access);
-    }
     ImagingDelete(imagep->image);
     PyObject_Del(imagep);
 }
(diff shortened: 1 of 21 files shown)


akx:
That's a surprising result from CodSpeed!

<del>
Probably has to do with the different CPUs: *AMD EPYC 7763 64-Core Processor* → *AMD EPYC 9V74 80-Core Processor*
</del>

Oh -- #9760 should be merged first before we take another look at this, because that one does significant things to `Image.filter()`!

radarhere:
https://github.com/python-pillow/Pillow/pull/9760 has been merged.

akx:
I don't really have an idea why this would regress those 3 NEAREST benchmarks aside from measurement fluke or some codegen esoterica...

PR #9774, opened 2026-07-06
Add coverage to tox

hugovk:
Add coverage to tox, now we use the coverage sysmon core where available (3.12+, nearly all supported versions), there's much less difference in the run time: 
  * `tox -e py314`: 31s with, 21s without
  * `tox -e py314`: 40s with, 20s without

Also add `py314t` and `py315t` to tox.

And follow up to https://github.com/python-pillow/Pillow/pull/9771, we can remove the `COVERAGE_CORE: sysmon` env var from GitHub Actions because it's now set at the tool level in `pyproject.toml`.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,10 +1,12 @@
 [tox]
 requires =
-    tox>=4.2
+    tox>=4.32
 env_list =
     lint
     mypy
-    py{py3, 315, 314, 313, 312, 311, 310}
+    py{py3}
+    py{310-315}
+    py{314-315}t
 
 [testenv]
 deps =
--- a/tox.ini
+++ b/tox.ini
@@ -15,6 +15,11 @@ extras =
 commands =
     {envpython} selftest.py
     {envpython} -m pytest \
+      --cov PIL \
+      --cov Tests \
+      --cov-report html \
+      --cov-report term \
+      --cov-report xml \
       --dist worksteal \
       --numprocesses auto \
       -W always \
--- a/.github/workflows/test-mingw.yml
+++ b/.github/workflows/test-mingw.yml
@@ -22,7 +22,6 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  COVERAGE_CORE: sysmon
   FORCE_COLOR: 1
 
 jobs:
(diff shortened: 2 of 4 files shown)


tox.ini
```diff
@@ -1,10 +1,12 @@
 [tox]
 requires =
-    tox>=4.2
+    tox>=4.32
 env_list =
     lint
     mypy
-    py{py3, 315, 314, 313, 312, 311, 310}
+    py{py3}
```
radarhere, line 7:
```suggestion
    pypy3
```

tox.ini
```diff
@@ -15,6 +17,11 @@ extras =
 commands =
     {envpython} selftest.py
     {envpython} -m pytest \
+      --cov PIL \
+      --cov Tests \
+      --cov-report html \
```
radarhere, line 22:
```suggestion
```
We don't use html elsewhere?
https://github.com/python-pillow/Pillow/blob/c97de203911a6cdad71bc435cf198fa001d4b908/.ci/test.sh#L7

radarhere:
Does this mean that the Lint job should have an 'Upload coverage' step?

tox.ini
hugovk: This is a workaround so tox-ini-fmt doesn't sort like:
```ini
env_list =
    pypy3
    lint
    mypy
    py{310-315}
    py{314-315}t
```

Which do you prefer?

tox.ini
hugovk: The CI isn't using tox for tests and doesn't need to generate HTML files.

I use tox locally and usually view coverage via the HTML files when I want to see details, and sometimes the terminal for the overview.

If anything, we could drop XML as that's mainly useful for Codecov in CI. In my other projects, I use tox in CI and locally, so have all three reports, as they don't add any noticeable extra.

hugovk (commented):
> Does this mean that the Lint job should have an 'Upload coverage' step?

No, it runs the `lint` and `mypy` envs, which don't run the base `testenv` or measure any coverage.

tox.ini
radarhere: Ah, ok. Yes, your initial version is preferrable.

radarhere (approved):

PR #9784, opened 2026-07-09
Improve error messages from C

akx:
Inspired by https://github.com/python-pillow/Pillow/issues/6667#issuecomment-4909586514

* New internal exception type `NotSupportedError` for operations that are _currently_ not supported; derives from both `ValueError` and `NotImplementedError` for backward compatibility.
  * Deriving from `NotImplementedError` allows for try-excepting in tests, so we can do better than (e.g.) `Tests.test_image_filter.test_sanity` where the test just quietly does nothing when a mode combo is not (yet, or currently) supported. (This wasn't wired up here yet.)
* The internal `ImagingError_*()` C APIs now all support an optional message text, and message texts were added where practical. No longer "image has wrong mode" or "images do not match" for everything! ✨ 
* Image nullness checks used to occasionally do ModeErrors, occasionally ValueErrors; now they're all ValueError.

### DIFF
--- a/src/libImaging/ColorLUT.c
+++ b/src/libImaging/ColorLUT.c
@@ -87,14 +87,23 @@ ImagingColorLUT3D_linear(
         return NULL;
     }
 
-    if (imIn->type != IMAGING_TYPE_UINT8 || imOut->type != IMAGING_TYPE_UINT8 ||
-        imIn->bands < 3 || imOut->bands < table_channels) {
-        return (Imaging)ImagingError_ModeError(NULL);
+    if (imIn->type != IMAGING_TYPE_UINT8 || imOut->type != IMAGING_TYPE_UINT8) {
+        return (Imaging)ImagingError_NotSupportedError(
+            "only uint8 images are supported"
+        );
     }
-
-    /* In case we have one extra band in imOut and don't have in imIn.*/
+    if (imIn->bands < 3) {
+        return (Imaging)ImagingError_ModeError("input image needs at least 3 bands");
+    }
+    if (imOut->bands < table_channels) {
+        return (Imaging)ImagingError_ValueError(
+            "output image needs at least as many bands as the table"
+        );
+    }
     if (imOut->bands > table_channels && imOut->bands > imIn->bands) {
-        return (Imaging)ImagingError_ModeError(NULL);
+        return (Imaging)ImagingError_ModeError(
+            "output image has more bands than input image and table"
+        );
     }
 
     ImagingSectionEnter(&cookie);
--- a/src/libImaging/ColorLUT.c
+++ b/src/libImaging/ColorLUT.c
@@ -89,12 +89,12 @@ ImagingColorLUT3D_linear(
 
     if (imIn->type != IMAGING_TYPE_UINT8 || imOut->type != IMAGING_TYPE_UINT8 ||
         imIn->bands < 3 || imOut->bands < table_channels) {
-        return (Imaging)ImagingError_ModeError();
+        return (Imaging)ImagingError_ModeError(NULL);
     }
 
     /* In case we have one extra band in imOut and don't have in imIn.*/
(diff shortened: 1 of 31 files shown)


src/libImaging/Reduce.c
```diff
@@ -1453,11 +1453,13 @@ ImagingReduce(Imaging imIn, int xscale, int yscale, int box[4]) {
...
     if (imIn->mode == IMAGING_MODE_P || imIn->mode == IMAGING_MODE_1) {
-        return (Imaging)ImagingError_ModeError();
+        return (Imaging)ImagingError_NotSupportedError("only modes P/1 supported");
     }
 
     if (imIn->type == IMAGING_TYPE_SPECIAL) {
-        return (Imaging)ImagingError_ModeError();
+        return (Imaging)ImagingError_NotSupportedError(
+            "special image types not supported"
+        );
```
radarhere, line 1459: I don't think that 'special' means anything to Python users, who haven't read
https://github.com/python-pillow/Pillow/blob/9cd07435967e22b56f2eeab0ed35c1d324f9e488/src/libImaging/Storage.c#L125-L130

Maybe just '16-bit images are not supported'?

akx: Sure, that makes sense what with #9794 too :)

akx:
Rebased post #9794.

src/libImaging/RankFilter.c
```diff
@@ -64,8 +64,18 @@ MakeRankFunction(UINT8) MakeRankFunction(INT32) MakeRankFunction(FLOAT32)
...
+        return (Imaging)ImagingError_ValueError(NULL);
+    }
+
+    if (!im) {
+        return (Imaging)ImagingError_ValueError(NULL);
+    }
+
+    if (im->bands != 1 || im->type == IMAGING_TYPE_I16) {
+        return (Imaging)ImagingError_ModeError(
+            "image must be 1-band non-special image"
```
radarhere: I think you'll agree that these messages shouldn't use the term 'special' anymore, in light of https://github.com/python-pillow/Pillow/pull/9794

akx: Sure! Fixed.

src/libImaging/Effects.c
```diff
@@ -33,7 +33,9 @@ ImagingEffectMandelbrot(int xsize, int ysize, double extent[4], int quality) {
     width = extent[2] - extent[0];
     height = extent[3] - extent[1];
     if (width < 0.0 || height < 0.0 || quality < 2) {
-        return (Imaging)ImagingError_ValueError(NULL);
+        return (Imaging)ImagingError_ValueError(
+            "width and height must be positive and quality >= 2"
```
radarhere: The check allows for zero width and height, which I would not describe as positive.

src/libImaging/ColorLUT.c
```diff
@@ -87,14 +87,23 @@ ImagingColorLUT3D_linear(
         return NULL;
     }
 
-    if (imIn->type != IMAGING_TYPE_UINT8 || imOut->type != IMAGING_TYPE_UINT8 ||
-        imIn->bands < 3 || imOut->bands < table_channels) {
-        return (Imaging)ImagingError_ModeError();
+    if (imIn->type != IMAGING_TYPE_UINT8 || imOut->type != IMAGING_TYPE_UINT8) {
+        return (Imaging)ImagingError_NotSupportedError(
+            "only uint8 images are supported"
```
radarhere: I don't think the average Pillow user would know what 'uint8 images' means. Could this be reworded to refer to modes?

akx: Reworded as "only 8-bit image modes are supported". I don't think there's a list of 8-bit modes in the code, so listing all supported modes "out loud" sounded like too much maintenance.

src/libImaging/Effects.c
akx: Reworded:
> width and height must be >=0 and quality >= 2

PR #9785, opened 2026-07-09
geometry: clean up imOut if operation fails

akx:
Alternative to https://github.com/python-pillow/Pillow/pull/9962

Sibling of #9784 - an earlier iteration of it uncovered Python complaining about a value being returned while an exception was set. `_transpose` did indeed just wholesale toss out the retval of the underlying operation function...

### DIFF
--- a/src/_imaging.c
+++ b/src/_imaging.c
@@ -2189,30 +2189,40 @@ _transpose(ImagingObject *self, PyObject *args) {
             return NULL;
     }
 
-    if (imOut) {
-        switch (op) {
-            case 0:
-                (void)ImagingFlipLeftRight(imOut, imIn);
-                break;
-            case 1:
-                (void)ImagingFlipTopBottom(imOut, imIn);
-                break;
-            case 2:
-                (void)ImagingRotate90(imOut, imIn);
-                break;
-            case 3:
-                (void)ImagingRotate180(imOut, imIn);
-                break;
-            case 4:
-                (void)ImagingRotate270(imOut, imIn);
-                break;
-            case 5:
-                (void)ImagingTranspose(imOut, imIn);
-                break;
-            case 6:
-                (void)ImagingTransverse(imOut, imIn);
-                break;
-        }
+    if (!imOut) {
+        return NULL;
+    }
+
+    Imaging imTemp = NULL;  // will be either NULL or imOut after the operation finishes
+    switch (op) {
+        case 0:
+            imTemp = ImagingFlipLeftRight(imOut, imIn);
+            break;
+        case 1:
+            imTemp = ImagingFlipTopBottom(imOut, imIn);
+            break;
+        case 2:
+            imTemp = ImagingRotate90(imOut, imIn);
+            break;
+        case 3:
+            imTemp = ImagingRotate180(imOut, imIn);
+            break;
+        case 4:
+            imTemp = ImagingRotate270(imOut, imIn);
+            break;
+        case 5:
+            imTemp = ImagingTranspose(imOut, imIn);
+            break;
+        case 6:
+            imTemp = ImagingTransverse(imOut, imIn);
+            break;
+        default:
+            PyErr_SetString(PyExc_ValueError, "No such transpose operation");
+            return NULL;
+    }
+    if (!imTemp) {  // operation failed and will have set an exception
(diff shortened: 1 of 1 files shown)


radarhere:
I understand completely that you dislike that the return value is ignored.

However, the return value only indicates failure if the modes or the sizes don't match, and `_transpose` creates `imOut` from those aspects of `imIn`, and `ImagingBoxBlur` (the only other places one of these functions is called from) does almost the same. It seems to me that there is no way that the return value would ever be a failure.

Is it worth considering the alternative solution of removing the mode and size checks from `ImagingFlipLeftRight` through to `ImagingTransverse`, so that there is no return value at all, and documenting this in a comment?

akx:
> It seems to me that there is no way that the return value would ever be a failure.

As noted, an earlier iteration of #9784 had an issue where I'd carelessly copy-pasted all checks across `Geometry.c` the same, even though some of the functions actually needed to check whether imIn's size is the transposition of imOut's size, not the same size.

That in turn caused CPython RuntimeErrors because we were both returning a value from a C extension function while having set an exception, because this function wasn't checking correctly... If this function had behaved correctly CPython-API-wise, I would've spent less time debugging that exception, and more time figuring I was a dummy for copy-pasting blindly :)

Given, as you note, that the `Imaging*` functions are `libImaging` internal APIs that are used by other functions, I'd want them to have the same checks that other `libImaging` functions do.

src/_imaging.c
```diff
@@ -2190,30 +2187,40 @@ _transpose(ImagingObject *self, PyObject *args) {
...
+        case 4:
+            imTemp = ImagingRotate270(imOut, imIn);
+            break;
+        case 5:
+            imTemp = ImagingTranspose(imOut, imIn);
+            break;
+        case 6:
+            imTemp = ImagingTransverse(imOut, imIn);
+            break;
+        default:  // Already checked above, but Clang-Tidy insists :)
```
radarhere:
```suggestion
        default:
```
I think you're welcome to enforce your own additional rules when writing code, as long as it doesn't break tests or the Lint job, but I don't think there should be a comment implying that we use Clang-Tidy across the repository if we don't.

So I think the `default:` is fine if you want, but either
a) let's remove the comment, if it is just your preference, or
b) did you want to remove this single comment, and instead suggest enforcing Clang-Tidy in the Lint job?

akx: You're right, the comment is overreaching. My IDE (CLion) just runs Clang-Tidy inspections automagically and shows them inline.

Now that you do mention it, I could run a Clang-Tidy on Pillow to see if it picks up worthwhile things :)

radarhere:
Wouldn't you also want to check the return values here?
https://github.com/python-pillow/Pillow/blob/11de4b9ae7d5a6553a3561b2de210647e5e78f98/src/libImaging/BoxBlur.c#L281-L289

akx:
> Wouldn't you also want to check the return values here?

Yep, good catch. This particular PR was just to fix this particular instance which bit me.

I can do another pass over all of these fallible cases in another PR?

akx:
> Wouldn't you also want to check the return values here?

Yep, good catch. This particular PR was just to fix this particular instance which bit me.

I can do another pass over all of these fallible cases in another PR?

akx:
> Wouldn't you also want to check the return values here?

Yep, good catch. This particular PR was just to fix this particular instance which bit me.

I can do another pass over all of these fallible cases in another PR?

src/_imaging.c
```diff
@@ -2192,30 +2192,40 @@ _transpose(ImagingObject *self, PyObject *args) {
...
+            break;
+        case 5:
+            imTemp = ImagingTranspose(imOut, imIn);
+            break;
+        case 6:
+            imTemp = ImagingTransverse(imOut, imIn);
+            break;
+        default:
+            PyErr_SetString(PyExc_ValueError, "No such transpose operation");
+            return NULL;
```
radarhere, line 2224: I suspect you are already aware, but `op` has been checked in the previous `switch`, so the`default` case will never run here.

akx: I was thinking of combining the two switch/cases even if it means duplicating the allocations, it might be clearer.

**EDIT:** Tried, looks ugly :)

PR #9786, opened 2026-07-09
Fix 16-bit ImageFilter accuracy

akx:
Follows up on #8438 (418ae7caa28d3ecb88f6cfaf1fb1a27b544794a8)

Also included in #9760.

1. The high byte of each 16-bit channel was being truncated due to an operator typo.
2. 16-bit values were inadvertently rounded up by a pixel value per iteration.

### DIFF
--- a/Tests/test_image_filter.py
+++ b/Tests/test_image_filter.py
@@ -210,6 +210,16 @@ def test_consistency_5x5(mode: str) -> None:
             assert_image_equal(source.filter(kernel), reference)
 
 
+@pytest.mark.parametrize("mode", ("I;16", "I;16L", "I;16B", "I;16N"))
+def test_consistency_i16_high_byte(mode: str) -> None:
+    # Exercise filters with a 16bpc image that has content in the high byte, too.
+    im = Image.new(mode, (8, 8), 1000)
+    pixel = im.filter(ImageFilter.SMOOTH).getpixel((4, 4))
+    assert isinstance(pixel, (int, float))
+    # Allow for kernel rounding errors.
+    assert abs(pixel - 1000) <= 1
+
+
 @pytest.mark.parametrize(
     "radius",
     (
--- a/Tests/test_image_filter.py
+++ b/Tests/test_image_filter.py
@@ -214,10 +214,11 @@ def test_consistency_5x5(mode: str) -> None:
 def test_consistency_i16_high_byte(mode: str) -> None:
     # Exercise filters with a 16bpc image that has content in the high byte, too.
     im = Image.new(mode, (8, 8), 1000)
-    pixel = im.filter(ImageFilter.SMOOTH).getpixel((4, 4))
-    assert isinstance(pixel, (int, float))
-    # Allow for kernel rounding errors.
-    assert abs(pixel - 1000) <= 1
+    # Ensure repeated smoothing retains the exact same color value,
+    # rather than drifting due to rounding errors.
+    for _ in range(5):
+        im = im.filter(ImageFilter.SMOOTH)
+        assert im.getpixel((4, 4)) == 1000
 
 
 @pytest.mark.parametrize(
--- a/src/libImaging/Filter.c
+++ b/src/libImaging/Filter.c
@@ -27,7 +27,6 @@
 #include <stdint.h>
 #include "Imaging.h"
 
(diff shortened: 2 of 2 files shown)


radarhere:
I've created https://github.com/akx/Pillow/pull/9 with a suggestion.

akx:
Sure, though then the relationship between the 1000 requested and the 1001 we're testing against is a little muddier.

radarhere:
We can restore it if you would prefer, I just didn't see any need to not be precise.

Is that what 'Allow for kernel rounding errors' was meant to refer to?

akx:
> Is that what 'Allow for kernel rounding errors' was meant to refer to?

Yes... Now that I look into it again without just allowing "well it's probably floating point 🤷", the source of the bug is the "Add one time for rounding" from https://github.com/python-pillow/Pillow/commit/847dd67a412b9689e08c1fc60ef9716dc475188c being accumulated into 1.0 throughout due to `int ss_int = ROUND_UP(ss);`.

That needs to be fixed, and then we can truly use exact values 😄

Tests/test_image_filter.py
```diff
@@ -210,6 +210,18 @@ def test_consistency_5x5(mode: str) -> None:
             assert_image_equal(source.filter(kernel), reference)
 
 
+@pytest.mark.parametrize("mode", ("I;16", "I;16L", "I;16B", "I;16N"))
+def test_consistency_i16_high_byte(mode: str) -> None:
+    # Exercise filters with a 16bpc image that has content in the high byte, too.
+    im = Image.new(mode, (8, 8), 1000)
+    # Ensure repeated smoothing retains the exact same color value,
+    # rather than drifting due to rounding errors.
+    for x in range(50):
```
```suggestion
    for _ in range(50):
```
See https://github.com/python-pillow/Pillow/blob/9cd07435967e22b56f2eeab0ed35c1d324f9e488/Tests/test_file_gif.py#L646

Tests/test_image_filter.py
```diff
@@ -210,6 +210,18 @@ def test_consistency_5x5(mode: str) -> None:
...
+        im = im.filter(ImageFilter.SMOOTH)
+        pixel = im.getpixel((4, 4))
+        assert pixel == 1000
```
```suggestion
        assert im.getpixel((4, 4)) == 1000
```

Tests/test_image_filter.py
akx: Sure. There's a Ruff lint for this: https://docs.astral.sh/ruff/rules/unused-loop-control-variable/

radarhere (approved):

Tests/test_image_filter.py
```diff
@@ -210,6 +210,17 @@ def test_consistency_5x5(mode: str) -> None:
             assert_image_equal(source.filter(kernel), reference)
 
 
+@pytest.mark.parametrize("mode", ("I;16", "I;16L", "I;16B", "I;16N"))
+def test_consistency_i16_high_byte(mode: str) -> None:
+    # Exercise filters with a 16bpc image that has content in the high byte, too.
+    im = Image.new(mode, (8, 8), 1000)
+    # Ensure repeated smoothing retains the exact same color value,
+    # rather than drifting due to rounding errors.
+    for _ in range(5):
```
radarhere, line 219: Every pixel in the image keeps a value of 1000 throughout. I'm not sure I see how 'drift' could occur over multiple iterations? The pixel values returned by `getpixel()` are integer values, not approximations.

akx: It would drift up (but evidently not forever) before the double-round-up fix in 77bd329a6e1cc15acf640a1ddeedffba2d6f8222:
```
>>> from PIL import Image, ImageFilter
>>>
>>> im = Image.new("I;16", (8, 8), 1000)
>>> for _ in range(5):
...     im = im.filter(ImageFilter.SMOOTH)
...     print(im.getpixel((4, 4)))
...
1001
1002
1003
1003
1003
>>>
```

After it:

```
>>> from PIL import Image, ImageFilter
>>> im = Image.new("I;16", (8, 8), 1000)
>>> for _ in range(5):
...     im = im.filter(ImageFilter.SMOOTH)
...     print(im.getpixel((4, 4)))
...
1000
1000
1000
1000
1000
>>>
```

radarhere: Ah, ok. The first run of the loop has an incorrect answer, and it becomes more incorrect over time.

PR #9788, opened 2026-07-10
Apply hoist optimizations to Geometry.c's transforms

akx:
Like the other recent optimization PRs: hoist `im*->xsize` and `im*->ysize` into locals so autovectorization can do its thing.

### DIFF
--- a/src/libImaging/Geometry.c
+++ b/src/libImaging/Geometry.c
@@ -22,20 +22,23 @@ ImagingFlipLeftRight(Imaging imOut, Imaging imIn) {
     if (!imOut || !imIn || imIn->mode != imOut->mode) {
         return (Imaging)ImagingError_ModeError();
     }
-    if (imIn->xsize != imOut->xsize || imIn->ysize != imOut->ysize) {
+    int in_xsize = imIn->xsize, in_ysize = imIn->ysize;
+    int out_xsize = imOut->xsize, out_ysize = imOut->ysize;
+
+    if (in_xsize != out_xsize || in_ysize != out_ysize) {
         return (Imaging)ImagingError_Mismatch();
     }
 
     ImagingCopyPalette(imOut, imIn);
 
-#define FLIP_LEFT_RIGHT(INT, image)               \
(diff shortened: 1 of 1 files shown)


src/libImaging/Geometry.c
```diff
@@ -22,20 +22,23 @@ ImagingFlipLeftRight(Imaging imOut, Imaging imIn) {
     if (!imOut || !imIn || imIn->mode != imOut->mode) {
         return (Imaging)ImagingError_ModeError();
     }
-    if (imIn->xsize != imOut->xsize || imIn->ysize != imOut->ysize) {
+    int in_xsize = imIn->xsize, in_ysize = imIn->ysize;
+    int out_xsize = imOut->xsize, out_ysize = imOut->ysize;
+
+    if (in_xsize != out_xsize || in_ysize != out_ysize) {
```
radarhere, line 28: Could you explain how this is an improvement? I mean, it wasn't using the values within a loop. There was no repetition at all.

akx: You mean `out_xsize`/`out_ysize`? Yes, they're not used in the loops, but I think it reads more cleanly for the human reader if it's 

```c
if (in_xsize != out_xsize || in_ysize != out_ysize) {
```

and not

```c
if (in_xsize != imOut->xsize || in_ysize != imOut->ysize) {
```

-- with the latter, at least my code eye starts wondering why there's two different idioms for the sizes.

radarhere: Yes, that is what I was referring to. This appears to be another different of opinion - your instinct is to wonder why the code is written differently, my instinct is to wonder why the code is more complex than it could be.

We'll see how #9962 goes. It is merged, then this discussion becomes irrelevant.

akx: My opinion about this isn't strong at all, but it's a consistency thing a tiny bit like https://github.com/python-pillow/Pillow/pull/9975#discussion_r3978494204 - it's a little easier to read if all of the similar functions have the same prelude.

akx:
Thanks! 😄

PR #9794, opened 2026-07-17
Rename IMAGING_TYPE_SPECIAL to IMAGING_TYPE_I16

radarhere:
Since #9053 removed BGR;15, BGR;16 and BGR;24, the only IMAGING_TYPE_SPECIAL modes left are I;16 modes.
https://github.com/python-pillow/Pillow/blob/9cd07435967e22b56f2eeab0ed35c1d324f9e488/src/libImaging/Storage.c#L125-L130

Perhaps it would be clearer to rename IMAGING_TYPE_SPECIAL to IMAGING_TYPE_I16?


akx (approved):
Good clarification!

PR #9796, opened 2026-07-17
Remove test support for NumPy < 1.23

radarhere:
#8326 caught a deprecation warning from NumPy < 1.23, reporting when an exception occurred within `__array_interface__` but was not raised.
>   /Pillow/Tests/test_image_array.py:51: DeprecationWarning: An exception was ignored while fetching the attribute `__array_interface__` from an object of type 'JpegImageFile'.  With the exception of `AttributeError` NumPy will always raise this exception in the future.  Raise this deprecation warning to see the original exception. (Warning added NumPy 1.21)

Those versions of NumPy are no longer [active](https://endoflife.date/numpy) though, so the warning no longer needs to be caught, leaving just the test for the actual exception.

### DIFF
--- a/Tests/test_image_array.py
+++ b/Tests/test_image_array.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 from typing import Any
 
 import pytest
-from packaging.version import parse as parse_version
 
 from PIL import Image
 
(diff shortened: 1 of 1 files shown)


akx (approved):
LGTM.

On a related note, tests seem to occasionally use `has_feature_version`, `skip_unless_feature_version`, `mark_if_feature_version` -- could some of those be similarly cleaned up if Pillow has lower bounds for wrapped library versions?

radarhere:
NumPy is slightly different - it's not a dependency we use, it's another library that we interact with. It also has EOL dates.

When we drop support for older versions of a dependency, we do update tests. See https://github.com/python-pillow/Pillow/pull/9053/changes/92bafe6b88d903d06f0eeba2f1dd26dd0eea52bb

However, we do value backwards compatibility. Even when dropping 12 year old versions of a dependency, we [still had a deprecation period](https://github.com/python-pillow/Pillow/pull/7998).

PR #9797, opened 2026-07-18
Handle premultiplied alpha modes in ImageColor.getcolor()

chuenchen309:
### Changes proposed in this pull request

`ImageColor.getcolor()` decides whether to append the alpha channel with `mode[-1] == "A"` (uppercase) in both branches (`ImageColor.py:160` and `:163`). The premultiplied modes are spelled `La` and `RGBa` (lowercase `a`), so they fall through and their alpha is silently dropped:

```python
>>> ImageColor.getcolor("red", "La")           # expected (76, 255)
76
>>> ImageColor.getcolor("#ff000080", "RGBa")   # expected (128, 0, 0, 128)
(255, 0, 0)
```

Downstream, `Image.new("La", size, "red")` comes out `(76, 0)` — fully transparent — and `Image.new("RGBa", size, "#ff000080")` loses the alpha entirely. The uppercase siblings `LA`/`RGBA` handle this correctly, and `La`/`RGBa` are first-class modes, so this is a gap rather than intentional.

* Handle `mode[-1] == "a"` in both branches, premultiplying each band by alpha (rounded, to match `convert()`).
* Add `test_color_premultiplied`, with expected values derived from `convert()` rather than hard-coded:
  ```
  getcolor("red", "La")         -> (76, 255)
  getcolor("#ff000080", "La")   -> (38, 128)
  getcolor("#ff000080", "RGBa") -> (128, 0, 0, 128)
  ```

`LA`/`RGBA`/`L`/`RGB` are unchanged. The change is pure insertion (+9 / +13 test), keeping code and reformatting separate. `test_color_premultiplied` fails before this change and passes after; the `ImageColor`/`ImageDraw`/`ImageOps`/`Image`/`ImagePalette` suites stay green (515 passed); `ruff` and `black` are clean.

`La`/`RGBa` are uncommon (mostly internal `convert`/resize intermediates), so this is a correctness fix for a low-frequency path, not a hot path.

---
Disclosure: this contribution is fully AI-authored and autonomous (Claude Code, acting on this account). An AI found the bug, ran the repro, wrote the fix and the test (expected values derived from `convert()`), and wrote this description; the human account holder reviews every change and is accountable for it. The verification above is real and re-runnable from the diff. If this isn't the kind of contribution you want, say so and I'll close it.

### DIFF
--- a/Tests/test_image_transform.py
+++ b/Tests/test_image_transform.py
@@ -94,7 +94,7 @@ class TestImageTransform:
         (
             ("RGB", (255, 0, 0)),
             ("RGBA", (255, 0, 0, 255)),
-            ("LA", (76, 0)),
+            ("LA", (76, 255)),
         ),
     )
     def test_fill(self, mode: str, expected_pixel: tuple[int, ...]) -> None:
--- a/Tests/test_imagecolor.py
+++ b/Tests/test_imagecolor.py
@@ -209,7 +209,6 @@ def test_color_premultiplied(mode: str, premultiplied_mode: str, color: str) -> 
     )
 
     assert ImageColor.getcolor(color, premultiplied_mode) == expected
-    assert Image.new(premultiplied_mode, (1, 1), color).getpixel((0, 0)) == expected
 
 
 def test_color_too_long() -> None:
--- a/Tests/test_imagecolor.py
+++ b/Tests/test_imagecolor.py
@@ -200,13 +200,13 @@ def test_color_hsv() -> None:
 
 
 @pytest.mark.parametrize("mode, premultiplied_mode", (("LA", "La"), ("RGBA", "RGBa")))
-@pytest.mark.parametrize("color", ("red", "#ff000080"))
-def test_color_premultiplied(mode: str, premultiplied_mode: str, color: str) -> None:
-    # The premultiplied modes are spelled with a lowercase "a", so an
-    # uppercase-only check drops their alpha. Compare against convert().
-    expected = (
-        Image.new(mode, (1, 1), color).convert(premultiplied_mode).getpixel((0, 0))
-    )
+@pytest.mark.parametrize("color", ("#f00", "#ff000080"))
+def test_color_premultiplied_alpha(
+    mode: str, premultiplied_mode: str, color: str
+) -> None:
+    im = Image.new(mode, (1, 1), color)
+    im_premultiplied = im.convert(premultiplied_mode)
+    expected = im_premultiplied.getpixel((0, 0))
 
     assert ImageColor.getcolor(color, premultiplied_mode) == expected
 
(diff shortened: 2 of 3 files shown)


chuenchen309:
CI caught a case I'd missed: `test_image_transform.py::test_fill[LA]`.

For `LA`/`RGBA` images with non-nearest resampling, `Image.transform` converts to the premultiplied `La`/`RGBa` mode, builds the fill with `Image.new(..., fillcolor)`, then converts back. So `fillcolor="red"` on an `LA` image goes through `getcolor("red", "La")`. Before this PR that returned a bare `76`, so the internal `La` fill got alpha 0 and the test asserted `(76, 0)` — a transparent fill for an opaque color. That was inconsistent with the `RGBA` case (opaque `(255, 0, 0, 255)`) and with `Image.new("LA", ..., "red")`, which is already `(76, 255)`.

With `getcolor` now returning `(76, 255)`, the `LA` fill is opaque and matches both, so I updated the expected pixel to `(76, 255)`. Happy to revert if you consider the transparent fill intentional.

This follow-up was written by the same AI coding agent as the PR; the human account holder reviews it and is accountable.

Tests/test_imagecolor.py
```diff
@@ -199,6 +199,19 @@ def test_color_hsv() -> None:
...
+@pytest.mark.parametrize("color", ("red", "#ff000080"))
+def test_color_premultiplied(mode: str, premultiplied_mode: str, color: str) -> None:
+    # The premultiplied modes are spelled with a lowercase "a", so an
+    # uppercase-only check drops their alpha. Compare against convert().
+    expected = (
+        Image.new(mode, (1, 1), color).convert(premultiplied_mode).getpixel((0, 0))
+    )
+
+    assert ImageColor.getcolor(color, premultiplied_mode) == expected
+    assert Image.new(premultiplied_mode, (1, 1), color).getpixel((0, 0)) == expected
```
radarhere:
```suggestion
```
It doesn't appear this line does anything to test ImageColor.

Tests/test_image_transform.py
```diff
@@ -94,7 +94,7 @@ def test_quad(self) -> None:
         (
             ("RGB", (255, 0, 0)),
             ("RGBA", (255, 0, 0, 255)),
-            ("LA", (76, 0)),
+            ("LA", (76, 255)),
```
radarhere, line 97: This check was originally added in #3163

src/PIL/ImageColor.py
```diff
@@ -152,17 +152,20 @@ def getcolor(color: str, mode: str) -> int | tuple[int, ...]:
...
-        return graylevel
-    elif mode[-1] == "A":
-        return rgb + (alpha,)
-    return rgb
+        value: tuple[int, ...] = (graylevel,)
+    else:
+        value = rgb
+    if mode[-1] == "a":
+        # Rounded, to match convert()'s premultiplication.
+        value = tuple((band * alpha + 127) // 255 for band in value)
```
radarhere: I'm not sure what the comment is referring to. This exact calculation method doesn't seem to be in `convert()` anywhere, and the reason to round the values isn't because of `convert()` as much as it is because L and RGB related modes in Pillow use integer pixel values.

radarhere (approved):

PR #9803, opened 2026-07-20
Enable Mypy strict mode

akx:
Rebase of https://github.com/python-pillow/Pillow/pull/9410, with the type fixes reduced to the minimum required to make `uvx --python=3.11 --with-requirements=.ci/requirements-mypy.txt mypy conftest.py selftest.py setup.py checks docs src winbuild Tests` pass. Most of the conflicts there were post f80b9144e63b8861bc536b1b9f2f5881c7610fe4 merge.

Closes #9410.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -28,5 +28,23 @@ jobs:
         uses: astral-sh/setup-uv@fac544c07dec837d0ccb6301d7b5580bf5edae39 # v8.2.0
       - name: Lint
         run: uvx --with tox-uv tox -e lint
+  mypy:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [
+          "3.14",
+          "3.13",
+          "3.12",
+          "3.11",
+        ]
+    name: Mypy
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          persist-credentials: false
+      - name: Install uv
+        uses: astral-sh/setup-uv@fac544c07dec837d0ccb6301d7b5580bf5edae39 # v8.2.0
       - name: Mypy
-        run: uvx --with tox-uv tox -e mypy
+        run: uvx --python=${{ matrix.python-version }} --with tox-uv tox -e mypy
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -28,5 +28,23 @@ jobs:
         uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d # v10.0.1
       - name: Lint
         run: uvx --with tox-uv tox -e lint
+  mypy:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [
+          "3.14",
+          "3.13",
+          "3.12",
+          "3.11",
+        ]
+    name: Mypy
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+      - name: Install uv
+        uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d # v10.0.1
       - name: Mypy
-        run: uvx --with tox-uv tox -e mypy
+        run: uvx --python=${{ matrix.python-version }} --with tox-uv tox -e mypy
--- a/src/PIL/ImageGrab.py
+++ b/src/PIL/ImageGrab.py
@@ -79,7 +79,7 @@ def grab(
                         scale = 1 if scale_down else 2
                         im_cropped = im.resize(
                             ((right - left) * scale, (bottom - top) * scale),
-                            box=tuple(coord * 2 for coord in bbox),
+                            box=(left * 2, top * 2, right * 2, bottom * 2),
                         )
                     else:
                         im_cropped = im.crop(bbox)
--- a/.ci/requirements-mypy.txt
+++ b/.ci/requirements-mypy.txt
@@ -4,7 +4,8 @@ arro3-core
 IceSpringPySideStubs-PyQt6
 IceSpringPySideStubs-PySide6
 ipython
-numpy==2.4.6
+numpy==2.4.6;python_version=="3.11"
+numpy;python_version>="3.12"
 packaging
 pyarrow-stubs
 pybind11
--- a/Tests/benchmarks.py
+++ b/Tests/benchmarks.py
@@ -17,7 +17,7 @@ TYPE_CHECKING = False
 if TYPE_CHECKING:
     from collections.abc import Callable
 
-    from pytest_benchmark.fixture import (  # type: ignore[import-not-found]
+    from pytest_benchmark.fixture import (  # type: ignore[unused-ignore, import-not-found]
         BenchmarkFixture,
     )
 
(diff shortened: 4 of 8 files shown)


radarhere:
Given that this is considerably smaller than #9410, I suspect you're only fixing the strict mode problems that arose testing against Python 3.11.

#9410 had its sights set a bit higher - it actually wanted to correct strict mode problems for [newer versions of Python as well](https://github.com/jorenham/Pillow/pull/1#discussion_r2745715320), which fell into [an unresolved debate](https://github.com/python-pillow/Pillow/pull/9414#issuecomment-3824350914) about whether to test against those versions or not.

But that isn't necessarily something to stop this PR from being merged. I just suspect it doesn't exactly replace #9410.

akx:
Same result on all supported Pythons.
```fish
$ for py in 3.11 3.12 3.13 3.14; uvx -q --python=$py --with-requirements=.ci/requirements-mypy.txt mypy conftest.py selftest.py setup.py checks docs src winbuild Tests; end
Success: no issues found in 303 source files
Success: no issues found in 303 source files
Success: no issues found in 303 source files
Success: no issues found in 303 source files
$
```

radarhere:
https://github.com/radarhere/Pillow/commit/cfb3ab8317693990e5673aa2ecd828b0cef21348 added to these changes gives failures on [3.12 - 3.14](https://github.com/radarhere/Pillow/actions/runs/29788412492).

.github/workflows/lint.yml
```diff
@@ -28,5 +28,23 @@ jobs:
...
+          "3.12",
+          "3.11",
+        ]
+    name: Mypy
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          persist-credentials: false
+      - name: Install uv
+        uses: astral-sh/setup-uv@fac544c07dec837d0ccb6301d7b5580bf5edae39 # v8.2.0
```
radarhere:
```suggestion
        uses: astral-sh/setup-uv@c771a70e6277c0a99b617c7a806ffedaca235ff9 # v9.0.0
```
See #9842

.github/workflows/lint.yml
```diff
@@ -28,5 +28,23 @@ jobs:
...
+      matrix:
+        python-version: [
+          "3.14",
+          "3.13",
+          "3.12",
+          "3.11",
+        ]
+    name: Mypy
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
```
radarhere:
```suggestion
      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
```

.github/workflows/lint.yml
```diff
@@ -28,5 +28,23 @@ jobs:
         uses: astral-sh/setup-uv@c771a70e6277c0a99b617c7a806ffedaca235ff9 # v9.0.0
       - name: Lint
         run: uvx --with tox-uv tox -e lint
+  mypy:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [
+          "3.14",
```
radarhere, line 37: Note to self - it appears https://github.com/apache/arrow/issues/48172 is blocking 3.15 support

radarhere:
A review was requested, so I'll restate - the fundamental issue here isn't any implementation detail, it's the question of whether we should do this - https://github.com/python-pillow/Pillow/pull/9414#issuecomment-3824350914
> Do we really need to run mypy on every version? Should we run on the oldest supported, or latest? Is there authoritative guidance on best practices here?

https://github.com/python/mypy/issues/12286 is an issue requesting the ability to check multiple versions at once. From the fact that it is still open, you might infer that this isn't an official priority.

https://github.com/python-pillow/Pillow/pull/9410#issuecomment-4198872025

> I don't think we have any "official" recommendations on this. But unofficially we recommend:
> * Always run mypy on the oldest supported Python version.
> * If your code has known platform-specific logic (e.g. Linux vs Windows), run the above on as many platforms as needed.
> * You may also run mypy on the newest supported Python version (in this case it makes sense to use `--enable-error-code=deprecated` to see deprecations early).

hugovk:
I'm also not sure if we need to enable strict mode, some of our annotations are pretty complex already and I don't think the `EpsImagePlugin.py` and `GifImagePlugin.py` changes improve readability.

The `ImageGrab.py` fix is needed for macOS with the current config.

src/PIL/ImageGrab.py
```diff
@@ -79,7 +79,7 @@ def grab(
                         scale = 1 if scale_down else 2
                         im_cropped = im.resize(
                             ((right - left) * scale, (bottom - top) * scale),
-                            box=tuple(coord * 2 for coord in bbox),
+                            box=(left * 2, top * 2, right * 2, bottom * 2),
```
hugovk, line 82: Would you like to either reduce the PR to just this, or split this bit to another PR? I'm happy to merge this.

radarhere: > The `ImageGrab.py` fix is needed for macOS with the current config.

Could you expand on what this means?

akx: Split out: https://github.com/python-pillow/Pillow/pull/9935

hugovk: > > The `ImageGrab.py` fix is needed for macOS with the current config.
> 
> Could you expand on what this means?

Yep, on `main` I get:

```console
❯ tox -e mypy
mypy: commands[0]> mypy conftest.py selftest.py setup.py checks docs src winbuild Tests
src/PIL/ImageGrab.py:82: error: Argument "box" to "resize" of "Image" has incompatible
type "tuple[int, ...]"; expected "tuple[float, float, float, float] | None"  [arg-type]
                                box=tuple(coord * 2 for coord in bbox),
                                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Found 1 error in 1 file (checked 302 source files)
mypy: exit 1 (2.58 seconds) /Users/hugo/github/Pillow> mypy conftest.py selftest.py setup.py checks docs src winbuild Tests pid=84969
  mypy: FAIL code 1 (2.58=setup[0.00]+cmd[2.58] seconds)
  evaluation failed :( (2.63 seconds)
```

And with https://github.com/python-pillow/Pillow/pull/9935:

```console
❯ tox -e mypy
mypy: commands[0]> mypy conftest.py selftest.py setup.py checks docs src winbuild Tests
Success: no issues found in 302 source files
  mypy: OK (2.22=setup[0.00]+cmd[2.22] seconds)
  congratulations :) (2.27 seconds)
```

radarhere: So why isn't that error appearing in our CI?

hugovk: CI runs on Ubuntu and the code in question is macOS-only, gated by `if sys.platform == "darwin"`.

radarhere: Ah, ok, I see.

.github/workflows/lint.yml
```diff
@@ -28,5 +28,23 @@ jobs:
...
+          "3.12",
+          "3.11",
+        ]
+    name: Mypy
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+      - name: Install uv
+        uses: astral-sh/setup-uv@c771a70e6277c0a99b617c7a806ffedaca235ff9 # v9.0.0
```
radarhere:
```suggestion
        uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d # v10.0.1
```
See https://github.com/python-pillow/Pillow/pull/9952

radarhere:
@akx do you mind terribly if we close this? The discussion doesn't seem to be in favour of it.

akx:
I don't mind! I'll close it off. It's a good starting point for v3 in the future maybe. :)

PR #9805, opened 2026-07-21
Raise ValueError instead of asserting number of FTEX texture formats

### LINKED ISSUE #9809: Reachable assert in FtexImagePlugin.py
Hi,

### What did you do?
In testing the `Image.open()`, I found a minor issue with a reachable assert that is inconsistent with the code's primary use of raising an error. I am wondering whether this is intended.

Pillow should reject the unsupported/malformed FTEX file using its normal image-format error handling, such as raising `UnidentifiedImageError`.
An explicit check raising `SyntaxError` may be appropriate here because `Image.open()` treats that as a plugin rejecting the input:
 ```python
if format_count != 1:
        msg = f"Unsupported FTEX format count: {format_count}"
        raise SyntaxError(msg)
```
### What actually happened?
 FtexImagePlugin._open() raises an uncaught AssertionError:
```
    File "repro.py", line 9, in <module>
      Image.open(io.BytesIO(data))
    File ".../PIL/Image.py", line 3693, in open
      im = _open_core(
    File ".../PIL/Image.py", line 3673, in _open_core
      im = factory(fp, filename)
    File ".../PIL/ImageFile.py", line 150, in __init__
      self._open()
    File ".../PIL/FtexImagePlugin.py", line 85, in _open
      assert format_count == 1
  AssertionError
```

* OS:  Ubuntu 22.04.5
* Python: 3.10.12
* Pillow:   12.2.0

```text
  Pillow 12.2.0
  Python 3.10.12 (main, Jun 22 2026, 18:55:27) [GCC 11.4.0]
  Python Pillow modules loaded from /home/xxxx/.local/lib/python3.10/site-packages/PIL
  Binary Pillow modules loaded from /home/xxxx/.local/lib/python3.10/site-packages/PIL
  --- PIL CORE support ok, compiled for 12.2.0
  --- FREETYPE2 support ok, loaded 2.14.3
  --- LITTLECMS2 support ok, loaded 2.18
  --- WEBP support ok, loaded 1.6.0
  --- AVIF support ok, loaded 1.4.1
  --- JPEG support ok, compiled for libjpeg-turbo 3.1.4.1
  --- OPENJPEG (JPEG2000) support ok, loaded 2.5.4
  --- ZLIB (PNG/ZIP) support ok, loaded 1.2.11, compiled for zlib-ng 2.3.3
  --- LIBTIFF support ok, loaded 4.7.1
  --- RAQM (Bidirectional Text) support ok, loaded 0.10.3,
      fribidi 1.0.8, harfbuzz 13.2.1
  *** LIBIMAGEQUANT support not installed
  --- XCB support ok
```

```python
import io
import struct
from PIL import Image
# FTEX header:
# magic, version, width, height, mipmap_count, format_count,
# compression format, data offset
data = struct.pack(
      "<4s7i",
      b"FTEX",
      1,
      1,
      1,
      1,
      2,  # The plugin only supports format_count == 1
      0,
      0,
  )
Image.open(io.BytesIO(data))
```

ritsth:
Resolves #9809

`FtexImageFile._open` used `assert format_count == 1` to reject multi-format files. That raises `AssertionError`, which is not one of the exceptions Pillow documents for unparseable files, so callers catching `UnidentifiedImageError` or `ValueError` do not catch it; and under `python -O` the assert is stripped, so the file is parsed as if valid.

This raises a `ValueError` instead, matching the existing "Invalid texture compression format" check in the same plugin. Added a regression test that patches the format count in a valid file, mirroring `test_invalid_texture`.

Used an AI assistant to find this; I reviewed and tested the change myself.

### DIFF
--- a/Tests/test_file_ftex.py
+++ b/Tests/test_file_ftex.py
@@ -38,3 +38,15 @@ def test_invalid_texture() -> None:
     with pytest.raises(ValueError, match="Invalid texture compression format: 2"):
         with Image.open(io.BytesIO(data)):
             pass
+
+
+def test_invalid_format_count() -> None:
+    with open("Tests/images/ftex_dxt1.ftc", "rb") as fp:
+        data = fp.read()
+
+    # Change the format count to an unsupported value
+    data = data[:20] + struct.pack("<i", 2) + data[24:]
+
+    with pytest.raises(ValueError, match="Unsupported number of formats: 2"):
+        with Image.open(io.BytesIO(data)):
+            pass
--- a/Tests/test_file_ftex.py
+++ b/Tests/test_file_ftex.py
@@ -40,13 +40,11 @@ def test_invalid_texture() -> None:
             pass
 
 
-def test_invalid_format_count() -> None:
+def test_unsupported_texture_format_count() -> None:
     with open("Tests/images/ftex_dxt1.ftc", "rb") as fp:
-        data = fp.read()
+        # Change the format count to an unsupported value
+        data = fp.read(20) + struct.pack("<i", 2)
 
-    # Change the format count to an unsupported value
-    data = data[:20] + struct.pack("<i", 2) + data[24:]
-
-    with pytest.raises(ValueError, match="Unsupported number of formats: 2"):
+    with pytest.raises(ValueError, match="Unsupported number of texture formats: 2"):
         with Image.open(io.BytesIO(data)):
             pass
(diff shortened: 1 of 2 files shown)


radarhere:
So are you saying you actually encountered a multi-format file in the wild?

radarhere:
I've created https://github.com/ritsth/Pillow/pull/1 with a suggestion.

ritsth:
> So are you saying you actually encountered a multi-format file in the wild?

No, I haven't encountered one in the wild. I found this while reading through the FTEX plugin's _open() and noticed the assert format_count == 1 was being used for input validation rather than an internal invariant check. That's a problem regardless of whether a real multi-format file exists, since assert raises an undocumented AssertionError and disappears entirely under python -O, so malformed/truncated input could either crash confusingly or silently skip validation. 

> I've created [ritsth#1](https://github.com/ritsth/Pillow/pull/1) with a suggestion.

Also applied your wording suggestion from ritsth/Pillow#1. Thanks!

PR #9806, opened 2026-07-21
Remove write method from test

radarhere:
Extracted from #9803

https://github.com/python-pillow/Pillow/blob/8f86212e947113e49909a1ae2012fee799adf2a8/Tests/test_image.py#L197-L209

The `write` method was originally added in #2227. I guess it became redundant after #8163 changed `FP` to inherit from `io.BytesIO`.


akx (approved):
Looks like an excellent change, if I do say so myself! 😁

PR #2217, opened 2016-11-12
Rearranged WebPImagePlugin _accept for speed

radarhere:
Minor rearranging for speed - if the first condition fails, the other conditions will not be checked.

PR #9807, opened 2026-07-21
Speed up some conversions

akx:
Experiments to speed up the other conversions didn't really yield fruit.

### DIFF
--- a/Tests/benchmarks.py
+++ b/Tests/benchmarks.py
@@ -163,10 +163,38 @@ def test_alpha_composite(
 @pytest.mark.parametrize(
     "mode_from, mode_to",
     [
+        # bilevel <-> greyscale (in-place path: l2bit / bit2l)
+        ("L", "1"),
+        ("1", "L"),
+        # greyscale <-> colour (pack / expand)
         ("RGB", "L"),
+        ("L", "RGB"),
+        ("RGB", "1"),
+        ("LA", "RGBA"),
+        # channel add / drop / premultiply
+        ("RGB", "RGBA"),
+        ("RGBA", "RGB"),
         ("RGBA", "LA"),
         ("RGBa", "RGBA"),
         ("RGBA", "RGBa"),
+        # palette expansion (p2rgb / p2rgba / pa2rgb / pa2rgba)
+        ("P", "RGB"),
+        ("P", "RGBA"),
+        ("PA", "RGB"),
+        ("PA", "RGBA"),
+        # colourspace conversions
+        ("RGB", "YCbCr"),
+        ("YCbCr", "RGB"),
+        ("RGB", "HSV"),
+        ("HSV", "RGB"),
+        ("RGB", "CMYK"),
+        ("CMYK", "RGB"),
+        # integer / float modes
+        ("L", "I"),
+        ("I", "L"),
+        ("L", "F"),
+        ("F", "L"),
+        ("RGB", "I;16"),
     ],
 )
 @pytest.mark.parametrize("size", SIZES, ids=_format_size)
--- a/Tests/benchmarks.py
+++ b/Tests/benchmarks.py
@@ -182,7 +182,7 @@ def test_alpha_composite(
         ("P", "RGBA"),
         ("PA", "RGB"),
         ("PA", "RGBA"),
-        # colourspace conversions
+        # colorspace conversions
         ("RGB", "YCbCr"),
         ("YCbCr", "RGB"),
         ("RGB", "HSV"),
--- a/src/libImaging/Convert.c
+++ b/src/libImaging/Convert.c
@@ -977,27 +977,30 @@ pa2f(UINT8 *out_, const UINT8 *in, int xsize, ImagingPalette palette) {
     }
 }
 
+// Set the alpha channel of the UINT32 `v` in-place to the given value.
+#ifdef WORDS_BIGENDIAN
+#define SET_ALPHA_32(v, alpha) v = ((v & 0xFFFFFF00u) | (alpha))
+#else
+#define SET_ALPHA_32(v, alpha) v = ((v & 0x00FFFFFFu) | ((UINT32)(alpha) << 24))
+#endif
+
 static void
 p2rgb(UINT8 *out, const UINT8 *in, int xsize, ImagingPalette palette) {
-    int x;
-    for (x = 0; x < xsize; x++) {
-        const UINT8 *rgb = &palette->palette[*in++ * 4];
-        *out++ = rgb[0];
-        *out++ = rgb[1];
-        *out++ = rgb[2];
-        *out++ = 255;
+    for (int x = 0; x < xsize; x++, in++, out += 4) {
+        UINT32 v;
+        memcpy(&v, &palette->palette[in[0] * 4], sizeof(v));
+        SET_ALPHA_32(v, 0xFF);
+        memcpy(out, &v, sizeof(v));
     }
 }
 
 static void
 pa2rgb(UINT8 *out, const UINT8 *in, int xsize, ImagingPalette palette) {
-    int x;
-    for (x = 0; x < xsize; x++, in += 4) {
-        const UINT8 *rgb = &palette->palette[in[0] * 4];
-        *out++ = rgb[0];
-        *out++ = rgb[1];
-        *out++ = rgb[2];
-        *out++ = 255;
+    for (int x = 0; x < xsize; x++, in += 4, out += 4) {
+        UINT32 v;
+        memcpy(&v, &palette->palette[in[0] * 4], sizeof(v));
+        SET_ALPHA_32(v, 0xFF);
+        memcpy(out, &v, sizeof(v));
     }
 }
 
@@ -1023,25 +1026,18 @@ pa2hsv(UINT8 *out, const UINT8 *in, int xsize, ImagingPalette palette) {
 
 static void
 p2rgba(UINT8 *out, const UINT8 *in, int xsize, ImagingPalette palette) {
-    int x;
-    for (x = 0; x < xsize; x++) {
-        const UINT8 *rgba = &palette->palette[*in++ * 4];
-        *out++ = rgba[0];
-        *out++ = rgba[1];
-        *out++ = rgba[2];
-        *out++ = rgba[3];
+    for (int x = 0; x < xsize; x++, in++, out += 4) {
+        memcpy(out, &palette->palette[in[0] * 4], 4);
     }
 }
 
 static void
 pa2rgba(UINT8 *out, const UINT8 *in, int xsize, ImagingPalette palette) {
-    int x;
-    for (x = 0; x < xsize; x++, in += 4) {
-        const UINT8 *rgb = &palette->palette[in[0] * 4];
-        *out++ = rgb[0];
-        *out++ = rgb[1];
-        *out++ = rgb[2];
-        *out++ = in[3];
+    for (int x = 0; x < xsize; x++, in += 4, out += 4) {
+        UINT32 v;
+        memcpy(&v, &palette->palette[in[0] * 4], sizeof(v));
+        SET_ALPHA_32(v, in[3]);
+        memcpy(out, &v, sizeof(v));
     }
 }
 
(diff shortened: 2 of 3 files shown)


Tests/benchmarks.py
```diff
@@ -163,10 +163,38 @@ def test_alpha_composite(
...
+        ("RGBA", "RGB"),
         ("RGBA", "LA"),
         ("RGBa", "RGBA"),
         ("RGBA", "RGBa"),
+        # palette expansion (p2rgb / p2rgba / pa2rgb / pa2rgba)
+        ("P", "RGB"),
+        ("P", "RGBA"),
+        ("PA", "RGB"),
+        ("PA", "RGBA"),
+        # colourspace conversions
```
radarhere:
```suggestion
        # colorspace conversions
```
I'm not going to pretend that we're consistent with color vs colour, but we do consistently use colorspace over colourspace, and all things being equal, I'd rather not confuse things further.

hugovk:

Edit: or not, due to conflict :)

Merging `main` in instead.

src/libImaging/Convert.c
```diff
@@ -1622,19 +1612,12 @@ convert(Imaging imOut, Imaging imIn, ModeID mode, ImagingPalette palette, int di
...
-        return (Imaging)ImagingError_ValueError("conversion not supported");
-#else
-        static char buf[100];
-        snprintf(
-            buf,
-            100,
-            "conversion from %.10s to %.10s not supported",
+        return (Imaging)PyErr_Format(
+            PyExc_ValueError,
+            "conversion from %s to %s not supported",
```
radarhere:
```suggestion
            "conversion from %.10s to %.10s not supported",
```
This has the inadvertent change that was initially in #9863

src/libImaging/Convert.c
```diff
@@ -1707,15 +1690,12 @@ ImagingConvertTransparent(Imaging imIn, const ModeID mode, int r, int g, int b)
...
         g = b = r;
     } else {
-        static char buf[100];
-        snprintf(
-            buf,
-            100,
-            "conversion from %.10s to %.10s not supported in convert_transparent",
+        return (Imaging)PyErr_Format(
+            PyExc_ValueError,
+            "conversion from %s to %s not supported in convert_transparent",
```
radarhere:
```suggestion
            "conversion from %.10s to %.10s not supported in convert_transparent",
```

src/libImaging/Convert.c
akx: Could we get #9863 (double approved, CI green) merged so I can just rebase this on top of `main`?

radarhere: Just waiting to see if you had any thoughts on my final attempt to reformat the `switch` - https://github.com/akx/Pillow/pull/12

akx: @radarhere merged https://github.com/akx/Pillow/pull/10 instead, so waiting for the merge on #9863 to rebase this and get rid of this commit altogether here 👍

src/libImaging/Convert.c
```diff
@@ -983,27 +977,30 @@ pa2f(UINT8 *out_, const UINT8 *in, int xsize, ImagingPalette palette) {
     }
 }
 
+// Set the alpha channel of the UINT32 `v` in-place to the given value.
+#ifdef WORDS_BIGENDIAN
+#define SET_ALPHA_32(v, alpha) v = ((v & 0xFFFFFF00u) | (alpha))
```
radarhere, line 982:
```suggestion
#define SET_ALPHA_32(v, alpha) v = ((v & 0xFFFFFF00u) | alpha)
```

src/libImaging/Convert.c
```diff
@@ -983,27 +977,30 @@ pa2f(UINT8 *out_, const UINT8 *in, int xsize, ImagingPalette palette) {
...
+#else
+#define SET_ALPHA_32(v, alpha) v = ((v & 0x00FFFFFFu) | ((UINT32)(alpha) << 24))
```
radarhere, line 984:
```suggestion
#define SET_ALPHA_32(v, alpha) v = ((v & 0x00FFFFFFu) | ((UINT32)alpha << 24))
```

akx: Rather not, because `alpha` could be a more complex expression.
 
EDIT: Prior art in e.g. the `B16`/`L16`/`S16` macros in `_imaging.c`.

src/libImaging/Convert.c
akx: Same here - defensive parentheses if `alpha` expands to an expression. 

EDIT: Prior art in e.g. the `B16`/`L16`/`S16` macros in `_imaging.c`.

radarhere:
Could you mention here what your thinking was behind the new tests?

akx:
> Could you mention here what your thinking was behind the new tests?

* `test_i_to_l_saturation`: a naive implementation might just `& 0xFF` a more-than-8-bit value to make it fit, which would cause wrapping instead of saturation.
* `test_p_to_rgb_exact`: tests for the behavior of `SET_ALPHA_32`, which has that little-endian/big-endian trickery. Also codifies the current behavior of how a P mode image with a palette that has alpha gets converted (which is also touched upon in #9834's new test)

radarhere:
> * `test_i_to_l_saturation`: a naive implementation might just `& 0xFF` a more-than-8-bit value to make it fit, which would cause wrapping instead of saturation.

You mean on `main`
```diff
diff --git a/src/libImaging/Convert.c b/src/libImaging/Convert.c
index 4a35a9a1f..bab3e7bc0 100644
--- a/src/libImaging/Convert.c
+++ b/src/libImaging/Convert.c
@@ -594,7 +594,7 @@ i2l(UINT8 *out, const UINT8 *in_, int xsize) {
         if (v <= 0) {
             *out = 0;
         } else if (v >= 255) {
-            *out = 255;
+            *out = v & 255;
         } else {
             *out = (UINT8)v;
         }
```
? That would cause two already existing tests in `test_file_gif.py` to fail.

> * `test_p_to_rgb_exact`: tests for the behavior of `SET_ALPHA_32`, which has that little-endian/big-endian trickery.

Guessing at what you think we might get wrong on this branch,
```diff
diff --git a/src/libImaging/Convert.c b/src/libImaging/Convert.c
index 3d0653139..72e2a93d6 100644
--- a/src/libImaging/Convert.c
+++ b/src/libImaging/Convert.c
@@ -979,9 +979,9 @@ pa2f(UINT8 *out_, const UINT8 *in, int xsize, ImagingPalette palette) {
 
 // Set the alpha channel of the UINT32 `v` in-place to the given value.
 #ifdef WORDS_BIGENDIAN
-#define SET_ALPHA_32(v, alpha) v = ((v & 0xFFFFFF00u) | (alpha))
+#define SET_ALPHA_32(v, alpha) v = ((v & 0x00FFFFFFu) | (alpha))
 #else
-#define SET_ALPHA_32(v, alpha) v = ((v & 0x00FFFFFFu) | ((UINT32)(alpha) << 24))
+#define SET_ALPHA_32(v, alpha) v = ((v & 0xFFFFFF00u) | ((UINT32)(alpha) << 24))
 #endif
 
 static void
```
causes 36 tests to fail, including `Tests/test_image_convert.py::test_rgba_pa`.

I don't naturally expect a performance PR to add more tests, since no functionality should be changed.

You're adding tests not for coverage, basic correctness, or for regression checking, but to ensure that we don't make a hypothetical mistake in the future? I may be wrong, but that sounds like a step too far to me.

akx:
> [...]? That would cause two already existing tests in `test_file_gif.py` to fail.

Fair enough. Removed.

> Guessing at what you think we might get wrong on this branch, [...] causes 36 tests to fail, including `Tests/test_image_convert.py::test_rgba_pa`.

Thanks for checking! Removed as well. The big-endian architecture Docker CI tests should then scream too. 👍 

> You're adding tests not for coverage, basic correctness, or for regression checking, but to ensure that we don't make a hypothetical mistake in the future?

Somewhat the other way around, really: to check that I didn't make mistakes within the branch. The new tests were stacked first in the branch's commits, so I could verify that they were green before optimizing things and after optimizing things.

src/libImaging/Convert.c
```diff
@@ -587,17 +587,11 @@ l2i(UINT8 *out_, const UINT8 *in, int xsize) {
...
         memcpy(&v, in_, sizeof(v));
-        if (v <= 0) {
-            *out = 0;
-        } else if (v >= 255) {
-            *out = 255;
-        } else {
-            *out = (UINT8)v;
-        }
+        // Branchless saturation
+        *out = (UINT8)(v <= 0 ? 0 : v >= 255 ? 255 : v);
```
radarhere:
```suggestion
        *out = v <= 0 ? 0 : (v >= 255 ? 255 : (UINT8)v);
```
Two ideas I'm suggesting here
1. Why not only cast `v` to `UINT8`?
2. Rather than `a ? b : c ? d : e`, use `a ? b : (c ? d : e)`. ...yes, I find it easier to read.

akx: 1. Technically, in my version the cast makes the type of the entire expression is UINT8, while in the second, it's... not? (I think it's `int` with an implicit narrowing when assigned to `out`)? So Explicit Is Better Than Implicit, to borrow the Zen of Python into this C code.
2. My IDE (CLion with Clang-Tidy etc.) complains (mild syntax note) about the redundant parens. I can add them.

radarhere:
Does the fact that codspeed doesn't detect any significant difference for `i2l()` mean that those changes are unnecessary?

akx:
@radarhere I wouldn't say so, as some of these depend on processor architecture:

On my laptop (macOS, M2 Max) there is a significant improvement there in addition to the ones also shown by CodSpeed (cropped out the rows that are < ±10%):

```
-------------------- benchmark 'convert': 26 tests, 2 sources -------------------
Name (time in us)                    0001_3bcb709 Min  0002_6dccfa0 Min      ΔMin
---------------------------------------------------------------------------------
test_convert[1237x811-I-L]                    86.2919           54.8750    -36.4%
test_convert[1237x811-PA-RGB]                820.7920          453.0830    -44.8%
test_convert[1237x811-P-RGB]                 832.7090          451.4590    -45.8%
test_convert[1237x811-PA-RGBA]               887.3750          472.0420    -46.8%
test_convert[1237x811-P-RGBA]                938.9170          384.2920    -59.1%
---------------------------------------------------------------------------------
```

This might be down to what compiler flags the benchmarked wheels end up being compiled with.

E.g. `-mtune=native` (if memory serves) would allow the compiler to use every fancy feature available on the (<sub>looks up CodSpeed</sub> Intel(R) Xeon(R) 6973P-C the benchmark ran on, but then that wheel might not work on CPUs that don't have those features (they'd get SIGILL or similar). This was touched on in #8209.

Or it just might be that the compiler itself isn't smart enough to autovectorize that loop on that platform. And as we saw in https://github.com/python-pillow/Pillow/pull/9654#issuecomment-4831287031, CodSpeed doesn't like multiple runs of the same benchmark - otherwise it'd be very interesting to also run these benchmarks automatically on a macOS ARM runner!

radarhere (approved):

radarhere:
Hmm. I don't see a definite improvement on my macOS M4.

I'm used to https://github.com/python-pillow/Pillow/pull/2217 being closed. Where exactly do you think the line is between [premature optimization](https://lawsofsoftwareengineering.com/laws/premature-optimization/) and a gain that doesn't appear on every OS when converting between two of our somewhat less popular modes?

akx:
> Hmm. I don't see a definite improvement on my macOS M4.

That's pretty curious. 🧐 Not sure what to attribute that to. Comparing the assembly of `i2l` on `main` and on this branch shows that my C compiler (Apple clang version 17.0.0 (clang-1700.3.19.1)) does heavier unrolling (a whole lot of smin.4s/smax.4s operations). 

Can you do `cc --version`, I'm wondering where the difference is...

I found an interesting post about this though now that I started researching things: https://secret.club/2021/04/09/std-clamp.html

> Where exactly do you think the line is between premature optimization and a gain that doesn't appear on every OS when converting between two of our somewhat less popular modes?

I'm not sure there's a definite line. For this case, the end result is exactly the same, it's less code (evidently an idiom some versions of clang recognize and do better with), and it doesn't show to be slower. If you like, I can of course revert it out of this PR?

radarhere:
Oh, apologies, I think I was testing the wrong conversion.

PR #9813, opened 2026-07-22
Fix unexpected errors when decoding truncated QOI images

### LINKED ISSUE #9812: qoi truncated index error
### What did you do?

I opened a truncated QOI file. Pillow reads its 14-byte header as a `1x220` RGBA image, then reaches EOF while decoding the declared pixels.

The truncated/malformed image should be rejected with Pillow's standard image-decoding exception, for example, an `OSError` describing a truncated or broken data stream, as other decoders do when reporting a premature EOF.

### What actually happened?

`QoiDecoder.decode()` reaches EOF and indexes the empty result of `read(1)`, so an implementation-level `IndexError` escapes from `Image.load()`:

```text
QOI (1, 220) RGBA
  File "repro.py", line 10, in <module>
    im.load()
  File ".../PIL/ImageFile.py", line 382, in load
    err_code = decoder.decode(b"")[1]
  File ".../PIL/QoiImagePlugin.py", line 64, in decode
    byte = self.fd.read(1)[0]
IndexError: index out of range
```

`self.fd.read(1)` returns `b""` at EOF and is indexed without a length check. The one-byte `QOI_OP_LUMA` read (`second_byte = self.fd.read(1)[0]`) and the `QOI_OP_RGB` / `QOI_OP_RGBA` payload reads (`self.fd.read(3)` / `read(4)`) have the same missing short-read handling and can leak similar errors on other truncated inputs.

Those exceptions become an uncaught exception, but not a vulnerability.

* OS: Ubuntu 22.04.5 LTS, x86_64
* Python: 3.10.12
* Pillow: 12.3.0

```text
Pillow 12.3.0
Python 3.10.12 (main, Jun 22 2026, 18:55:27) [GCC 11.4.0]
--- PIL CORE support ok, compiled for 12.3.0
--- FREETYPE2 support ok, loaded 2.14.3
--- LITTLECMS2 support ok, loaded 2.19
--- WEBP support ok, loaded 1.6.0
--- AVIF support ok, loaded 1.4.2
--- JPEG support ok, compiled for libjpeg-turbo 3.1.4.1
--- OPENJPEG (JPEG2000) support ok, loaded 2.5.4
--- ZLIB (PNG/ZIP) support ok, loaded 1.2.11, compiled for zlib-ng 2.3.3
--- LIBTIFF support ok, loaded 4.7.1
--- RAQM (Bidirectional Text) support ok, loaded 0.10.5, fribidi 1.0.8, harfbuzz 14.2.1
```

```python
import base64, io
from PIL import Image
data = base64.b64decode("cW9pZgAAAAEAAADc///T//QAAAEAAwAAAAABAA==")

im = Image.open(io.BytesIO(data))
print(im.format, im.size, im.mode)   # QOI (1, 220) RGBA
im.load()                            # IndexError: index out of range
```

abhi-0203:
## Summary

Fixes a crash in `QoiDecoder.decode()` when a truncated QOI file is opened. The decoder now raises `OSError` with the standard "image file is truncated" message instead of an uncaught `IndexError`.

## Problem

When a truncated QOI file is opened, `QoiDecoder.decode()` reads past EOF and indexes the empty result of `read(1)`, causing an `IndexError`:

```python
byte = self.fd.read(1)[0]  # IndexError: index out of range when read returns b''
```

This affects multiple read sites in the decoder:
- The initial opcode byte read (line 64)
- The QOI_OP_RGB 3-byte payload read (line 67)
- The QOI_OP_RGBA 4-byte payload read (line 69)
- The QOI_OP_LUMA second byte read (line 89)

## Fix

Added length checks before indexing all `read()` results. When a short read is detected (EOF), the decoder raises `OSError("image file is truncated")`, consistent with other Pillow decoders (JPEG, PNG, etc.).

## Testing

The fix can be verified by opening a truncated QOI file:
```python
from PIL import Image
im = Image.open("truncated.qoi")
im.load()  # Previously: IndexError, Now: OSError("image file is truncated")
```

Fixes #9812

### DIFF
--- a/Tests/test_file_qoi.py
+++ b/Tests/test_file_qoi.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from io import BytesIO
 from pathlib import Path
 
 import pytest
@@ -30,6 +31,15 @@ def test_invalid_file() -> None:
 
     with pytest.raises(SyntaxError):
         QoiImagePlugin.QoiImageFile(invalid_file)
+
+
+@pytest.mark.parametrize("length", (14, 44, 48, 55))
+def test_truncated(length: int) -> None:
+    with open("Tests/images/pil123rgba.qoi", "rb") as fp:
+        data = fp.read()[:length]
+    with Image.open(BytesIO(data)) as im:
+        with pytest.raises(ValueError, match="not enough image data"):
+            im.load()
 
 
 def test_op_index() -> None:
--- a/src/PIL/QoiImagePlugin.py
+++ b/src/PIL/QoiImagePlugin.py
@@ -63,21 +63,18 @@ class QoiDecoder(ImageFile.PyDecoder):
         while len(data) < dest_length:
             byte_data = self.fd.read(1)
             if not byte_data:
-                msg = "image file is truncated"
-                raise OSError(msg)
+                break
             byte = byte_data[0]
             value: bytes | bytearray
             if byte == 0b11111110 and self._previous_pixel:  # QOI_OP_RGB
                 rgb_data = self.fd.read(3)
                 if len(rgb_data) < 3:
-                    msg = "image file is truncated"
-                    raise OSError(msg)
+                    break
                 value = bytearray(rgb_data) + self._previous_pixel[3:]
             elif byte == 0b11111111:  # QOI_OP_RGBA
                 value = self.fd.read(4)
                 if len(value) < 4:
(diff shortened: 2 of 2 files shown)


radarhere:
I've created https://github.com/abhi-0203/Pillow/pull/1 with a suggestion.

radarhere (approved):

Issue #9811, opened 2026-07-21
bigtiff offset overflow


nevercodecorrect:
### What did you do?

I opened a malformed 38-byte BigTIFF whose first out-of-line tag data offset (`0x97006b0097009797`) exceeds the platform's signed seek range (`sys.maxsize` on 64-bit).

### What did you expect to happen?
Pillow should reject or skip the malformed TIFF directory through its normal corrupt-image path. `ImageFileDirectory_v2.load()` already catches `OSError` around this loop, warns, and returns. A raw integer-conversion exception should not escape the public `Image.open()` call.

## What actually happened?

`ImageFileDirectory_v2.load()` reads a 64-bit out-of-line data offset and passes it straight to `fp.seek(offset)`. Because the offset exceeds C `ssize_t`, `BytesIO.seek()` raises `OverflowError`, which is a subclass of `ArithmeticError`
(not `OSError`), So the surrounding `except OSError` handler does not catch it:

```pytb
Traceback (most recent call last):
  File "repro.py", line 8, in <module>
    Image.open(io.BytesIO(data))
  File ".../PIL/Image.py", line 3697, in open
    im = _open_core(
  File ".../PIL/Image.py", line 3677, in _open_core
    im = factory(fp, filename)
  File ".../PIL/TiffImagePlugin.py", line 1176, in __init__
    super().__init__(fp, filename)
  File ".../PIL/ImageFile.py", line 150, in __init__
    self._open()
  File ".../PIL/TiffImagePlugin.py", line 1201, in _open
    self._seek(0)
  File ".../PIL/TiffImagePlugin.py", line 1249, in _seek
    self.tag_v2.load(self.fp)
  File ".../PIL/TiffImagePlugin.py", line 917, in load
    fp.seek(offset)
OverflowError: Python int too large to convert to C ssize_t
```

The relevant code in `ImageFileDirectory_v2.load()` is:

```python
if size > (8 if self._bigtiff else 4):
    here = fp.tell()
    (offset,) = self._unpack("Q" if self._bigtiff else "L", data)
    ...
    fp.seek(offset)
```

wrapped by `except OSError as msg: warnings.warn(str(msg)); return`.

These TIFF bytes can raise an unexpected `OverflowError`. For a caller that whitelists those exceptions, it becomes an uncaught exception. The failure is in the pure-Python side so i assume it could not be a vulnerability.

### What are your OS, Python and Pillow versions?

[...]

PR #9818, opened 2026-07-23
Raise SyntaxError instead of OverflowError for oversized JPEG 2000 box

### LINKED ISSUE #9817: Oversized JPEG 2000 box length leaks `OverflowError` from `Image.open()`
### What did you do?
Open a malformed JP2 file whose first box declares an extended 64-bit length (`0xffffffffff000004`) that exceeds the platform's signed seek range (`sys.maxsize` on 64-bit). The issue is identified via fuzzing and manually reproduced to verify.

Pillow should reject the impossible box geometry through its normal corrupt-image path. A box that is too *small* (`lbox < hlen`) already raises `SyntaxError("Invalid header length")`, which `Image.open()` normalizes into `UnidentifiedImageError`. A box that is impossibly *large* should also be rejected the same way. 

## What actually happened?
[Jpeg2KImagePlugin._parse_jp2_header()](https://github.com/python-pillow/Pillow/blob/main/src/PIL/Jpeg2KImagePlugin.py#L148) builds a top-level `BoxReader` with no known stream length. `next_box_type()` reads the extended 64-bit length, and because the reader has no length bound, `_can_read(lbox - hlen)` returns `True` and the huge value is stored in `remaining_in_box`. The parser does not recognize the box type and iterates; on the next `next_box_type()` call, the stored length is passed straight to `fp.seek(...)`. 
Because it exceeds C `ssize_t`, `BytesIO.seek()` raises `OverflowError` and it escapes to the caller:

```text
  File "repro.py", line 6, in <module>
    Image.open(io.BytesIO(data))
  File ".../PIL/Image.py", line 3687, in open
    im = _open_core(
  File ".../PIL/Image.py", line 3667, in _open_core
    im = factory(fp, filename)
  File ".../PIL/ImageFile.py", line 152, in __init__
    self._open()
  File ".../PIL/Jpeg2KImagePlugin.py", line 281, in _open
    header = _parse_jp2_header(self.fp)
  File ".../PIL/Jpeg2KImagePlugin.py", line 166, in _parse_jp2_header
    tbox = reader.next_box_type()
  File ".../PIL/Jpeg2KImagePlugin.py", line 86, in next_box_type
    self.fp.seek(self.remaining_in_box, os.SEEK_CUR)
OverflowError: Python int too large to convert to C ssize_t
```

The relevant code in `BoxReader.next_box_type()` is:

```python
def next_box_type(self) -> bytes:
    # Skip the rest of the box if it has not been read
    if self.remaining_in_box > 0:
        self.fp.seek(self.remaining_in_box, os.SEEK_CUR)   # <-- OverflowError here
    self.remaining_in_box = -1

    lbox, tbox = cast(tuple[int, bytes], self.read_fields(">I4s"))
    if lbox == 1:
        lbox = cast(int, self.read_fields(">Q")[0])
        hlen = 16
    else:
        hlen = 8

    if lbox < hlen or not self._can_read(lbox - hlen):
        msg = "Invalid header length"
        raise SyntaxError(msg)

    self.remaining_in_box = lbox - hlen
    return tbox
```

The `_can_read()` guard does not catch this for the top-level reader, because that reader is created with no length (`has_length` is `False` and `remaining_in_box` is `-1` at the check), so `_can_read()` takes its `return True  # No length known, just read` branch and accepts the oversized box.

These JP2 bytes, therefore, raise an unexpected `OverflowError` which becomes an uncaught exception. The failure is in the pure-Python side, so I assume it isnot a memory-safety vulnerability.

This is the JPEG 2000 counterpart of #9811 

Reproduced on the current `main` branch (commit `9e282f5`) and confirmed on the
latest release 12.3.0.

* OS: Ubuntu 22.04.5 LTS, x86_64
* Python: 3.14.4
* Pillow: 13.0.0.dev0 (git `main`)

```text
Pillow 13.0.0.dev0
Python 3.14.4 (main, Jun 12 2026, 00:08:58) [GCC 11.4.0]
--- PIL CORE support ok, compiled for 13.0.0.dev0
--- JPEG support ok, compiled for libjpeg-turbo 2.1.2
--- OPENJPEG (JPEG2000) support ok, loaded 2.4.0
--- LIBTIFF support ok, loaded 4.3.0
```

```python
import base64, io
from PIL import Image
data = base64.b64decode("AAAADGpQICANCocKAAAAAQAAAAD//////wAABA==")
Image.open(io.BytesIO(data))   # OverflowError: Python int too large to convert to C ssize_t
```

nevercodecorrect:
Fixes #9817 .

JPEG 2000's BoxReader accepts an extended 64-bit box length with no bound (the top-level reader has no known length), then passes it to fp.seek(..., SEEK_CUR), raising OverflowError from Image.open() on malformed input. 
This is the JP2 counterpart of #9811 (TIFF, fixed in #9816). The box is now rejected as a malformed header (SyntaxError → UnidentifiedImageError), consistent with the plugin's other  #checks.

 * In function def next_box_type(self), add a check to reject a box whose contents end beyond the maximum seekable position
 * Add a test case.

### DIFF
--- a/Tests/test_file_jpeg2k.py
+++ b/Tests/test_file_jpeg2k.py
@@ -240,6 +240,17 @@ def test_oversized_box_length() -> None:
         Jpeg2KImagePlugin.Jpeg2KImageFile(BytesIO(data))
 
 
+def test_oversized_box_length_when_reading() -> None:
+    # Do not raise an OverflowError() when reading the contents of a box
+    data = (
+        b"\x00\x00\x00\x0cjP  \x0d\x0a\x87\x0a"  # JP2 signature box
+        + struct.pack(">I4s", 1, b"jp2h")  # box read into with a 64-bit length
+        + struct.pack(">Q", 16 + 2**63)  # length whose contents overflow fp.read()
+    )
+    with pytest.raises(SyntaxError, match="Box length too large"):
+        Jpeg2KImagePlugin.Jpeg2KImageFile(BytesIO(data))
+
+
 def test_layers_type(card: ImageFile.ImageFile, tmp_path: Path) -> None:
     outfile = tmp_path / "temp_layers.jp2"
     for quality_layers in [[100, 50, 10], (100, 50, 10), None]:
--- a/Tests/test_file_jpeg2k.py
+++ b/Tests/test_file_jpeg2k.py
@@ -230,16 +230,14 @@ def test_header_errors() -> None:
 
 
 def test_oversized_box_length() -> None:
-    # A box declaring a 64-bit length that ends beyond the maximum seekable
-    # position must be rejected instead of raising OverflowError from seek()
+    # Do not raise an OverflowError() when seeking to the end of a box
     data = (
         b"\x00\x00\x00\x0cjP  \x0d\x0a\x87\x0a"  # JP2 signature box
-        + struct.pack(">I4s", 1, b"\x00\x00\x00\x00")  # box with 64-bit length
-        + struct.pack(">Q", 0xFFFFFFFFFF000004)  # length beyond the seek range
+        + struct.pack(">I4s", 1, b"")  # box with 64-bit length
+        + struct.pack(">Q", 16 + 2**63)  # length beyond the seek range
     )
-    with pytest.raises(UnidentifiedImageError):
-        with Image.open(BytesIO(data)):
-            pass
+    with pytest.raises(SyntaxError, match="Box length too large"):
+        Jpeg2KImagePlugin.Jpeg2KImageFile(BytesIO(data))
 
 
 def test_layers_type(card: ImageFile.ImageFile, tmp_path: Path) -> None:
--- a/Tests/test_file_jpeg2k.py
+++ b/Tests/test_file_jpeg2k.py
@@ -229,23 +229,14 @@ def test_header_errors() -> None:
             pass
 
 
-def test_oversized_box_length() -> None:
-    # Do not raise an OverflowError() when seeking to the end of a box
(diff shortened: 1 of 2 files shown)


Tests/test_file_jpeg2k.py
```diff
@@ -229,6 +229,19 @@ def test_header_errors() -> None:
             pass
 
 
+def test_oversized_box_length() -> None:
+    # A box declaring a 64-bit length that ends beyond the maximum seekable
+    # position must be rejected instead of raising OverflowError from seek()
+    data = (
+        b"\x00\x00\x00\x0cjP  \r\n\x87\n"  # JP2 signature box
```
radarhere:
```suggestion
        b"\x00\x00\x00\x0cjP  \x0d\x0a\x87\x0a"  # JP2 signature box
```
It would seem clearer to match the form given in
https://github.com/python-pillow/Pillow/blob/9e282f5d754fe49ede35fde65fd862c6c50d1f9f/src/PIL/Jpeg2KImagePlugin.py#L378-L381

radarhere:
I've created https://github.com/nevercodecorrect/Pillow/pull/1 with a suggestion.

nevercodecorrect:
Thank you. This actually makes more sense. I have applied it.

radarhere (dismissed):

nevercodecorrect:
Change the location again as I think there could be two locations causing problem. One is [seek()](https://github.com/nevercodecorrect/Pillow/blob/main/src/PIL/Jpeg2KImagePlugin.py#L86) another is [read()](https://github.com/nevercodecorrect/Pillow/blob/main/src/PIL/Jpeg2KImagePlugin.py#L74)

radarhere (approved):

PR #9823, opened 2026-07-25
Simplified SPIDER code

radarhere:


src/PIL/SpiderImagePlugin.py
```diff
@@ -151,6 +151,7 @@ def _open(self) -> None:
         else:
             msg = "inconsistent stack header values"
             raise SyntaxError(msg)
+        self.is_animated = self.n_frames > 1
```
akx, line 154: I'd maybe keep this as a read-only property? Now a suitably misguided user could set `self.is_animated`...

radarhere: This is how the other plugins operate.

```
$ grep -ir 'self.is_animated =' .
./PIL/MpoImagePlugin.py:        self.is_animated = self.n_frames > 1
./PIL/PngImagePlugin.py:        self.is_animated = self.n_frames > 1
./PIL/AvifImagePlugin.py:        self.is_animated = self.n_frames > 1
./PIL/MicImagePlugin.py:        self.is_animated = self._n_frames > 1
./PIL/DcxImagePlugin.py:        self.is_animated = self.n_frames > 1
./PIL/WebPImagePlugin.py:        self.is_animated = self.n_frames > 1
./PIL/FliImagePlugin.py:        self.is_animated = self.n_frames > 1
./PIL/TiffImagePlugin.py:                self.is_animated = self.__next != 0
```

If the consensus is that we should change the general behaviour, then sure, let's abandon this and create a new PR.

akx: Alright, this is fine then :) Better to be consistent.

PR #9827, opened 2026-07-26
Move typing imports under `TYPE_CHECKING` block

hugovk:
Use https://docs.astral.sh/ruff/rules/#flake8-type-checking-tc so we only import typing things when checking types.


src/PIL/Image.py
```diff
@@ -2402,7 +2405,7 @@ def resize(
             factor_x = int((box[2] - box[0]) / size[0] / reducing_gap) or 1
             factor_y = int((box[3] - box[1]) / size[1] / reducing_gap) or 1
             if factor_x > 1 or factor_y > 1:
-                reduce_box = self._get_safe_box(size, cast(Resampling, resample), box)
+                reduce_box = self._get_safe_box(size, cast("Resampling", resample), box)
```
akx, line 2406: Did this get changed just for consistency? (Resampling is defined in this file.)

hugovk: It was done by TC006: https://docs.astral.sh/ruff/rules/runtime-cast-value/

akx: Check. That wasn't clear from the commit/PR message. I didn't scroll down far enough to see that all of TC got enabled :)

hugovk: We could ignore TC006, skip it for this line, or leave it be. I don't mind really.

src/PIL/Image.py
```diff
@@ -1252,7 +1252,8 @@ def convert_transparency(
             if new_im.mode == "P" and new_im.palette:
                 try:
                     new_im.info["transparency"] = new_im.palette.getcolor(
-                        cast(tuple[int, ...], trns), new_im  # trns was converted to RGB
+                        cast("tuple[int, ...]", trns),
+                        new_im,  # trns was converted to RGB
```
```suggestion
                        cast("tuple[int, ...]", trns),  # trns was converted to RGB
                        new_im,
```
Moves the comment to the first line, like https://github.com/python-pillow/Pillow/blob/abb8ba000acbdb18e36ac6b110b8c7f790ba8f9d/src/PIL/Image.py#L1196-L1197

radarhere (approved):
I've created https://github.com/hugovk/Pillow/pull/179 with some final suggestions.

PR #9828, opened 2026-07-28
Use `fmt: skip` instead of `fmt: off`/`fmt: on`

akx:
A little shorter to write, and a little harder to forget to re-enable formatting... 😇

This also de-triplicates the RGB-to-XYZ matrix from test_image_convert.


Tests/test_box_blur.py
```diff
@@ -225,7 +207,7 @@ def test_radius_bigger_then_width() -> None:
             [159, 154, 148, 142, 136, 130, 124],
             [160, 154, 148, 142, 137, 131, 125],
             [160, 155, 149, 143, 137, 131, 125],
-        ],
+        ],  # fmt: skip
```
radarhere:
```suggestion
        ],
```
This looks like you just added a skip for no reason?

Tests/test_box_blur.py
```diff
@@ -240,7 +222,7 @@ def test_extreme_large_radius() -> None:
             [162, 162, 162, 162, 162, 162, 162],
             [162, 162, 162, 162, 162, 162, 162],
             [162, 162, 162, 162, 162, 162, 162],
-        ],
+        ],  # fmt: skip
```
radarhere:
```suggestion
        ],
```

Tests/test_box_blur.py
akx: For symmetry with the others. This sample doesn't need a `fmt: skip` because all of the values are 3-digit. I can take it out if you like.

radarhere: That would be my preference. I guess your concern is that someone might change these values in the future and `fmt: skip` might be needed? I think the opposite is also the case - someone might change the other instances to where those `fmt: skip` aren't needed, and then this one looks silly.

Tests/test_box_blur.py
```diff
@@ -225,7 +207,7 @@ def test_radius_bigger_then_width() -> None:
             [159, 154, 148, 142, 136, 130, 124],
             [160, 154, 148, 142, 137, 131, 125],
             [160, 155, 149, 143, 137, 131, 125],
-        ],
+        ],
```
radarhere:
```suggestion
        ],
```
This PR currently changes this line with no visible difference. I think you've introduced an unexpected type of newline character.

Tests/test_image_convert.py
```diff
@@ -290,54 +290,44 @@ def test_p2pa_palette() -> None:
...
 def test_matrix_wrong_mode() -> None:
     # Arrange
     im = hopper("L")
-    # fmt: off
-    matrix = (
-        0.412453, 0.357580, 0.180423, 0,
-        0.212671, 0.715160, 0.072169, 0,
-        0.019334, 0.119193, 0.950227, 0)
-    # fmt: on
+
```
```suggestion
```
Nitpicking, but this adds a blank line that doesn't match the pattern from the method above or below.

Tests/test_color_lut.py
```diff
@@ -282,17 +285,17 @@ def test_overflow(self) -> None:
...
-                              ])).load()
-        # fmt: on
+        table = [
+            -1, -1, -1,   2, -1, -1,
+            -1,  2, -1,   2,  2, -1,
+
+            -1, -1,  2,   2, -1,  2,
+            -1,  2,  2,   2,  2,  2,
+        ]  # fmt: skip
+
```
radarhere:
```suggestion
```
Consistency with the next section.

Tests/test_color_lut.py
```diff
@@ -484,24 +485,21 @@ def test_3_channels(self) -> None:
...
     def test_4_channels(self) -> None:
         lut = ImageFilter.Color3DLUT.generate(
             5, channels=4, callback=lambda r, g, b: (b, r, g, (r + g + b) / 2)
         )
         assert tuple(lut.size) == (5, 5, 5)
         assert lut.name == "Color 3D LUT"
-        # fmt: off
         assert lut.table[:24] == [
             0.0, 0.0, 0.0, 0.0,  0.0, 0.25, 0.0, 0.125,  0.0, 0.5, 0.0, 0.25,
             0.0, 0.75, 0.0, 0.375,  0.0, 1.0, 0.0, 0.5,  0.0, 0.0, 0.25, 0.125
```
radarhere:
```suggestion
            0.0, 0.75, 0.0, 0.375,  0.0, 1.0, 0.0, 0.5,  0.0, 0.0, 0.25, 0.125,
```
Consistency with subsequent changes.

radarhere (approved):

Issue #1852, opened 2016-04-21
quantize() to a palette often chooses unideal color matches


ericpauley:
### What did you do?
I am quantizing many images to the same palette and colors are often not picked to be ideal values, instead being mapped to close values when an exact match exists in the palette.
### What did you expect to happen?
Colors would be matched to the closest value in the palette or an exact match if one exists.
### What actually happened?
Colors are mapped to close values. (Specifically white is mapped to #fcfcfc)
### What versions of Pillow and Python are you using?
    PILLOW_VERSION = '2.9.0'
    VERSION = '1.1.7'
    PYTHON = '2.7.10'
Please include code that reproduces the issue and whenever possible, an image that demonstrates the issue. The best reproductions are self-contained scripts with minimal dependencies.

```python
import PIL.Image
[...]

Issue #9619, opened 2026-05-07
Jarvis-judice-ninke dithering


jmerc77:
Feature request...

I wanted to use jarvis-judice-ninke dither but it is not in Pillow nor seems planned.
jarvis-judice-ninke is a more enhanced variant of floyd-steinberg.

while floyd- steinberg uses:
 *7
351

jarvis-judice-ninke uses:
  *75
35753
13531

for the diffusion matrix. this does mean it is a little slower. however, Pillow is fast enough that the speed difference may be unnoticed for typical use cases.

```text

--------------------------------------------------------------------
Pillow 12.1.1
Python 3.13.12 (tags/v3.13.12:1cbe481, Feb  3 2026, 18:22:25) [MSC v.1944 64 bit (AMD64)]
--------------------------------------------------------------------
Python executable is C:\Users\Owner\AppData\Local\Python\pythoncore-3.13-64\pythonw.exe
System Python files loaded from C:\Users\Owner\AppData\Local\Python\pythoncore-3.13-64
--------------------------------------------------------------------
Python Pillow modules loaded from C:\Users\Owner\AppData\Local\Python\pythoncore-3.13-64\Lib\site-packages\PIL
Binary Pillow modules loaded from C:\Users\Owner\AppData\Local\Python\pythoncore-3.13-64\Lib\site-packages\PIL
--------------------------------------------------------------------
--- PIL CORE support ok, compiled for 12.1.1
--- TKINTER support ok, loaded 8.6
--- FREETYPE2 support ok, loaded 2.14.1
--- LITTLECMS2 support ok, loaded 2.17
--- WEBP support ok, loaded 1.6.0
--- AVIF support ok, loaded 1.3.0
--- JPEG support ok, compiled for libjpeg-turbo 3.1.3
--- OPENJPEG (JPEG2000) support ok, loaded 2.5.4
--- ZLIB (PNG/ZIP) support ok, loaded 1.3.1.zlib-ng, compiled for zlib-ng 2.3.2
--- LIBTIFF support ok, loaded 4.7.1
*** RAQM (Bidirectional Text) support not installed
*** LIBIMAGEQUANT (Quantization method) support not installed
*** XCB (X protocol) support not installed
--------------------------------------------------------------------
```
Followup: 
I noticed 2 to be implemented dithering options. any timeline for those?
also, hitherdither implements all of it already but is slower than Pillow. it may be a good reference though.

radarhere:
[...]

PR #9829, opened 2026-07-28
Make `.quantize(palette=...)` use exact colors when possible, and improve dithering quality

akx:
Refs #1852 (fixes the reproducer bug: we now select exact colors when available, but we're still not necessarily selecting the _mathematically nearest_ color when quantizing to a given palette).

* Refactors the `topalette()` conversion function to a separate file for ease of understanding
* Refactors the `topalette()` function to multiple functions for less nesting and ease of reading
* Improves internal documentation/constants around palettes 
* Fixes a couple possible but implausible NULL pointer derefs (that could occur if there was less than ~1100 bytes of memory available for Python)
* The usual™️ hoist + restrict optimizations for internals
* Adds benchmarks for quantize-into-palette

... and then the main thing:

* Adds a small linear-probing hash table for finding exact matches in the palette before looking a close-enough one up in the cache "cube" in the palette. In other words, if a palette specifies a color, we will use it, guaranteed.
* `ImagingPaletteCache*` functions are made private, since the structure of the cache was undocumented and private anyway; we need to add a new flag in the cache entries to have a better guess of when we might have an exact match for a color.

EDIT: this now also includes the changes from https://github.com/python-pillow/Pillow/pull/9338 as the last commit, since that touches the same files as this jiggles around.


akx:
I think we'll have to accept some performance degradation in exchange for actually making conversions accurate...

src/libImaging/ConvertToPalette.c
```diff
@@ -0,0 +1,559 @@
...
+                rd += rx;
+                rx += 2 * RSTEP * RSTEP;
+            }
+        }
+    }
+
+    /* Step 3 -- Update cache */
+
+    /* The c array now contains the closest match for each
+     * cache slot in the box.  Update the cache. */
```
radarhere, line 167:
```suggestion
    /* The C array now contains the closest match for each
     * cache slot in the box. Update the cache. */
```

src/libImaging/ConvertToPalette.c
```diff
@@ -0,0 +1,559 @@
...
+                memcpy(out + x * 4, &v, sizeof(v));
+            } else {
+                out[x] = (UINT8)palette_index;
+            }
+
+            r -= (int)palette->palette[palette_index * 4];
+            g -= (int)palette->palette[palette_index * 4 + 1];
+            b -= (int)palette->palette[palette_index * 4 + 2];
+
+            /* propagate errors (don't ask ;-) */
```
radarhere, line 400: Is this PR not a good time to ask?

akx: I don't think it's a great time to ask. 😄 My Git annotate says it's from [2006's "pil 1.1.1 baseline" commit](https://github.com/hugovk/pil-svn.effbot.org/commit/17809567216b8f50df9f337704b622ba3be28594) - it refers to how the Floyd-Steinberg algorithm has been unrolled below.

Might have a look at improving the comments if I end up picking up #9619.

src/libImaging/ConvertToPalette.c
akx: This is also from 2006 and was just moved over to this file...

src/libImaging/ConvertToPalette.c
akx: This was touched upon in #9338, btw.

PR #9831, opened 2026-07-29
Improve benchmark test image

akx:
The old image was highly channel-correlated, low-color, low-structure and non-representative of anything real. My bad in #9654 :)

The new image contains both low-energy, high-energy (full noise) and zero-energy segments.
If someone has an idea for an even better synthetic image, I'm all ears!

Let's see how the Codspeed numbers change - at least locally, `test_quantize` and `test_histogram` got a lot slower, because they actually need to do work 😄 

| Old | New |
|--------|--------|
| old | new2 |

### DIFF
--- a/Tests/benchmarks.py
+++ b/Tests/benchmarks.py
@@ -33,7 +33,7 @@ MODES = ["RGB", "RGBA", "L", "LA"]
 # Note that adjusting this will naturally change how long operations take.
 # The `bench` fixture takes care of saving this information in the extra info
 # for the benchmark run, so that throughput (Mpx/s) can be recomputed in the future.
-SIZES = [(1024, 1024)]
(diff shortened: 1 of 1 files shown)


akx:
> Merging this PR will degrade performance by 29.32%

As predicted! But no, code performance didn't change, just measurements for the more accurate...

hugovk:
Good idea. Do you think making the image non-square would have any benefit?

akx:
@hugovk Good point, it might. I'm a little afk right now (I'll let you know it's quite warm in Dubrovnik). And maybe also NPOT (non-power-of-2)... Honestly some tests (resizing, etc.) could benefit from running on NPOT and non-NPOT?

hugovk:
Yes, sounds good.

akx:
@hugovk Made it NPOT and non-square :)

hugovk:
Thanks!

PR #9768, opened 2026-07-04
Speed up `Image.getbbox()`; fix it for I;16 modes

akx:
Hoist/restrict and autovectorization optimizations for `getbbox()`.

A review also showed that I;16 modes weren't correctly supported by this function (they were iterating over int8), so that got fixed too. I know it's bad form to mix optimizations and bugfixes into one PR, but the bugfix would have required  the `const type* row = ...` hoist assignments anyhow.

```
-------------------------- benchmark 'bbox': 20 tests, 2 sources ---------------------------
Name (time in us)                               0001_40ea9a1 Min  0002_d07694f Min      ΔMin
--------------------------------------------------------------------------------------------
test_getbbox[1024x1024-L-half-corner]                   226.7091           10.9579    -95.2%
test_getbbox[1024x1024-RGB-half-corner]                 229.2499           38.2080    -83.3%
test_getbbox[1024x1024-RGBA-half-corner]                229.3329           40.5840    -82.3%
test_getbbox[1024x1024-LA-half-corner]                  230.6249           38.2080    -83.4%
test_getbbox[1024x1024-I;16-half-centered]              231.0410           21.3749    -90.7%
test_getbbox[1024x1024-L-half-centered]                 240.4579           11.3329    -95.3%
test_getbbox[1024x1024-RGBA-half-centered]              249.7090           44.7499    -82.1%
test_getbbox[1024x1024-LA-half-centered]                249.8330           41.2500    -83.5%
test_getbbox[1024x1024-RGB-half-centered]               251.2090           41.9160    -83.3%
test_getbbox[1024x1024-L-empty]                         295.4580           11.4590    -96.1%
test_getbbox[1024x1024-L-small-centered]                297.5831           12.5830    -95.8%
test_getbbox[1024x1024-LA-empty]                        298.1250           43.9170    -85.3%
test_getbbox[1024x1024-RGB-empty]                       298.1250           47.5419    -84.1%
test_getbbox[1024x1024-RGBA-empty]                      298.1670           47.0839    -84.2%
test_getbbox[1024x1024-I;16-small-centered]             299.1251           24.1250    -91.9%
test_getbbox[1024x1024-I;16-empty]                      300.4170           21.9160    -92.7%
test_getbbox[1024x1024-I;16-half-corner]                301.2920           19.0841    -93.7%
test_getbbox[1024x1024-LA-small-centered]               304.9170           51.2920    -83.2%
[...]

PR #9832, opened 2026-07-29
New benchmarks

akx:
This PR simply stacks all of the new benchmarks and benchmark utilities (`benchmark_save`, useful for debugging) from #9743, #9753, #9760, #9762, #9768, #9772, #9773, #9788, #9807, and #9829, plus a Mypy change from #9803 -- aside from changes that wouldn't work with the other changes from those branches (such as newly-supported modes, or the correctness change from #9829).

I think it would be useful to merge this before rebasing those, so we have good benchmark baselines. 

#9831 should also be merged, maybe before this one.


hugovk (approved):
Let's merge after #9831.

PR #9833, opened 2026-07-29
Allow saving some PA images as PNG

akx:
Limitation found while working on #9829: we weren't able to save PA (palette index + alpha band) images into PNG at all. This adds limited capacity for that: in case each index+alpha combination occurs only once across the image, we can save it as P+RGBA-palette instead.

An example PA file saved with Pillow - those are solid primaries! 😄 

temp

### DIFF
--- a/src/PIL/PngImagePlugin.py
+++ b/src/PIL/PngImagePlugin.py
@@ -1317,6 +1317,58 @@ def _save_all(im: Image.Image, fp: IO[bytes], filename: str | bytes) -> None:
     _save(im, fp, filename, save_all=True)
 
 
+def _pa_to_p(im: Image.Image) -> Image.Image:
+    """
+    Try to fold a `PA` image's alpha band into an RGBA palette.
+
+    :param im: A `PA` mode image.
+    :return: A `P` mode image with an RGBA palette.
+    """
+
+    # PNG has no palette+alpha mode like Pillow's PA mode.
+    # However, palette entries may carry alpha via the tRNS chunk:
+    # > For color type 3 (indexed-color), the tRNS chunk contains
+    # > a series of one-byte alpha values, corresponding to entries
+    # > in the PLTE chunk.
+    # > Each entry indicates that pixels of the corresponding palette
+    # > index shall be treated as having the specified alpha value.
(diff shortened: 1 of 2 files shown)


src/PIL/PngImagePlugin.py
```diff
@@ -1317,6 +1317,58 @@ def _save_all(im: Image.Image, fp: IO[bytes], filename: str | bytes) -> None:
...
+    # as P, so the RGBA palette gets split back to PLTE + tRNS.
+
+    assert im.mode == "PA"
+    indexes, alpha = im.split()
+    index_alpha: dict[int, int] = {}
+    for index, alpha_value in set(zip(indexes.tobytes(), alpha.tobytes())):
+        if index_alpha.setdefault(index, alpha_value) != alpha_value:
+            msg = (
+                "cannot write mode PA as PNG: "
+                "a palette entry is used with multiple alpha values"
```
radarhere, line 1349: I'm not saying we need to go down this rabbit hole, but if there are less than 256 entries, it would be possible to achieve two alpha values for one palette entry by copying the entry and remapping it within the image. So a (255, 0, 0) entry with 100 alpha and 200 alpha could become two (255, 0, 0) entries with individual alphas.

radarhere:
This would be the only conversion done when saving a single frame PNG. I wonder if it isn't better/consistent to require users to convert the image before the save?

You're converting a PA image, to P with RGBA palette. Maybe a generic `convert()` from PA to P should generate an RGBA palette using the logic you've written here?

src/PIL/PngImagePlugin.py
akx: That could be a future enhancement, but it'd also mean we'd write a different palette than the user originally passed in into the PNG. (The roundtrip test here would probably fail.)

akx:
> Maybe a generic `convert()` from PA to P should generate an RGBA palette using the logic you've written here?

I thought about that too (or rather whether this function should belong somewhere else), and I'm not sure.

`convert()`ing PA to P currently makes every pixel opaque. This probably hasn't been documented properly, so maybe having PA-to-P conversion in the general case do this (when possible) would be an improvement?

Anyway, while looking at this (comment!) I also found #9834...

PR #9834, opened 2026-07-29
Ensure Image and core palettes are in sync after PA conversion

akx:
Alternative to https://github.com/python-pillow/Pillow/pull/10008

While working on #9833, I noticed that an RGBA palette's alphas would be "resurrected" if you converted from PA to P (which retains the RGBA palette in the core im object even if the `ImagePalette` wrapper is recreated), and then to RGBA.

With this, the "zombie" alpha from the RGBA doesn't appear anymore.
I would suspect this is more the Do What I Mean behavior, because an RGBA palette's alpha values weren't "visible" in a PA image anyway.

### DIFF
--- a/Tests/test_image_convert.py
+++ b/Tests/test_image_convert.py
@@ -290,6 +290,19 @@ def test_p2pa_palette() -> None:
     assert im_pa.getpalette() == im.getpalette()
 
 
+@pytest.mark.xfail(reason="zombie colors from palette shouldn't reappear")
+def test_pa2p_truly_drops_alpha() -> None:
+    im = Image.frombytes("P", (2, 1), bytes([0, 1])).convert("PA")
+    im.putpalette(bytes([255, 0, 0, 7, 0, 255, 0, 9]), "RGBA")
+    im.putalpha(Image.frombytes("L", (2, 1), bytes([240, 220])))
+    assert im.get_flattened_data() == ((0, 240), (1, 220))  # Matches the alpha band
+    im_p = im.convert("P")
+    assert im_p.palette is not None
+    assert im_p.im.getpalettemode() == im_p.palette.mode == "RGB"
+    rgba_data = im_p.convert("RGBA").get_flattened_data()
+    assert rgba_data == ((255, 0, 0, 255), (0, 255, 0, 255))
+
+
 def test_matrix_illegal_conversion() -> None:
     # Arrange
     im = hopper("CMYK")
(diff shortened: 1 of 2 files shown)


Tests/test_image_convert.py
```diff
@@ -290,6 +290,18 @@ def test_p2pa_palette() -> None:
     assert im_pa.getpalette() == im.getpalette()
 
 
+def test_pa2p_truly_drops_alpha() -> None:
+    im = Image.frombytes("P", (2, 1), bytes([0, 1])).convert("PA")
+    im.putpalette(bytes([255, 0, 0, 7, 0, 255, 0, 9]), "RGBA")
+    im.putalpha(Image.frombytes("L", (2, 1), bytes([240, 220])))
+    assert im.get_flattened_data() == ((0, 240), (1, 220))  # Matches the alpha band
```
```suggestion
```
These lines don't appear relevant.

src/PIL/Image.py
```diff
@@ -1245,6 +1245,8 @@ def convert_transparency(
             from . import ImagePalette
 
             new_im.palette = ImagePalette.ImagePalette("RGB", im.getpalette("RGB"))
+            new_im.palette.dirty = True
+            new_im.load()
```
```suggestion
            new_im.putpalette(im.getpalette())
```

Tests/test_image_convert.py
akx: They're meant to show the palette alpha isn't used.

src/PIL/Image.py
akx: Applied but with explicit `"RGB"` + comment.

Tests/test_image_convert.py
radarhere: I consider it pretty foundational that the palette doesn't affect pixel values (without conversion). I'm not sure it needs to be tested here? We already have
https://github.com/python-pillow/Pillow/blob/abb8ba000acbdb18e36ac6b110b8c7f790ba8f9d/Tests/test_image_putalpha.py#L28-L33

akx: That `test_image_putalpha` segment doesn't seem to really cover the contents of the palette.

I don't see how having this here really hurts?

src/PIL/Image.py
radarhere: Out of curiosity, why explicit `"RGB"`?

akx: Because then you don't need to look up `getpalette()`'s definition to find out that the default is in fact "RGB", and not something that's inferred from `im`, as I would've initially assumed.

(Or, IOW, [Explicit Is Better Than Implicit.](https://peps.python.org/pep-0020/))

radarhere:
This PR is immediately copying the palette from Python into C, and I've realised that this doesn't need to be the case - it can be lazily deferred until a future `load()` call.

I've created https://github.com/python-pillow/Pillow/pull/10008. See what you think.

PR #9844, opened 2026-08-03
Use `pytest.approx`, `pytest.raises` and `pytest.skip`

hugovk:


radarhere (approved):

Tests/test_file_libtiff.py
```diff
@@ -321,7 +321,7 @@ def check_tags(
                         and libtiff
                     ):
                         # libtiff does not support real RATIONALS
-                        assert round(abs(float(reloaded_value) - float(value)), 7) == 0
+                        assert float(reloaded_value) == pytest.approx(float(value))
```
radarhere:
```suggestion
                        assert reloaded_value == pytest.approx(value)
```

Issue #9101, opened 2025-07-19
Consider adopting an image-format-per-plugin model


bwoodsend:
Pillow's current model of *all image formats in one package* means that Pillow's footprint will inevitably grow boundlessly. The 10.1.0 Linux wheels were ~3.5MB, 10.2.0 went up to 4.5MB then 11.3.0 jumped all the way up to ~6.5MB for AVIF support (and these are compressed sizes – the installed footprint is worse). For anyone not using the new image formats and where a small footprint is coveted (serverless lambdas, small end user applications, services shipped as Docker images), every new addition is also a large regression.

From a security perspective, a new unneeded format is also a choice between an increased surface area of attack or squatting on old versions of Pillow – both of which are obviously bad. (Being able to limit formats using `Image.open(untrusted_content, formats=["JPEG", "PNG"])` helps but it's not really enough that I'd feed comfortable ignoring a CVE in `libavif`.)

I'm aware that pillow can be built without support for some formats by building from source and but neither are appealing options and they both get much worse if you want all the files on a server to come from say RPMs for better efficiency and security auditing.

I was using the `pillow-avif-plugin` in one of my projects for a while before it got merged into `pillow`. I thought it was a really good model. Anyone who didn't need AVIF support didn't have to pay for its large footprint and anyone who did only had to add one `import pillow_avif` and an extra item to their dependencies to get it. The integration felt very seamless.

Would you consider adopting a policy whereby all new and/or large and/or niche and/or complex formats are permanently kept as separately distributed `pillow_$format_plugin` wheels?

aclark4life:
@bwoodsend I don't think we'd rule it out but I'd definitely want to be careful with the implementation. In fact there is a discussion of adding a framework trove classifier in #8892 in case that helps.

radarhere:
I think the author of pillow-avif-plugin always considered it to be a temporary measure until Pillow integrated support for the format, and people [rejoiced](https://github.com/python-pillow/Pillow/pull/5201#issuecomment-2768047694) when it was combined into Pillow. When we didn't include it in the wheels immediately, there was [pushback](https://github.com/python-pillow/Pillow/pull/5201#issuecomment-2833360127).

You would also need to convince the users in https://github.com/python-pillow/Pillow/issues/4247 that they should just use an [existing plugin](https://pillow.readthedocs.io/en/stable/handbook/third-party-plugins.html).

I'm just saying that others have a different perspective.

[...]

PR #9849, opened 2026-08-05
Deprecate IM image format

radarhere:
https://github.com/python-pillow/Pillow/issues/9101#issuecomment-4134756068

> we could consider deprecating some of the obscure core plugins. There are definitely a few added during the early PIL days, where it was really hard to find sample images when adding tests decades later! These should be long deprecations, which could be dropped if we get feedback of real use.

When looking for specification for the IM format for #9800 on Google, I was unable to find anything unrelated to Pillow. I suggest deprecating with a two year deprecation period.

### DIFF
--- a/Tests/test_mode_i16.py
+++ b/Tests/test_mode_i16.py
@@ -46,7 +46,7 @@ def test_basic(tmp_path: Path, mode: str) -> None:
     im_out = im_in.transform((w, h), Image.Transform.EXTENT, (0, 0, w, h))
     verify(im_out)  # transform
 
-    filename = tmp_path / "temp.im"
+    filename = tmp_path / "temp.tiff"
     im_in.save(filename)
 
     with Image.open(filename) as im_out:
--- a/Tests/test_mode_i16.py
+++ b/Tests/test_mode_i16.py
@@ -44,7 +44,7 @@ def test_basic(tmp_path: Path, mode: str) -> None:
     im_out = im_in.transform((w, h), Image.Transform.EXTENT, (0, 0, w, h))
     verify(im_out)  # transform
 
-    filename = tmp_path / "temp.im"
+    filename = tmp_path / "temp.tiff"
     im_in.save(filename)
 
     with Image.open(filename) as im_out:
--- a/docs/deprecations.rst
+++ b/docs/deprecations.rst
@@ -21,7 +21,7 @@ ExifTags.IFD.Makernote
 ``ExifTags.IFD.MakerNote``.
 
 Image getdata()
-~~~~~~~~~~~~~~~
+^^^^^^^^^^^^^^^
 
 .. deprecated:: 12.1.0
 
@@ -29,6 +29,13 @@ Image getdata()
 :py:meth:`~PIL.Image.Image.get_flattened_data` can be used instead. This new method is
 identical, except that it returns a tuple of pixel values, instead of an internal
 Pillow data type.
+
+IM image format
+^^^^^^^^^^^^^^^
+
+.. deprecated:: 13.0.0
+
+The IM image format has been deprecated.
 
 Removed features
 ----------------
--- a/docs/deprecations.rst
+++ b/docs/deprecations.rst
@@ -21,7 +21,7 @@ ExifTags.IFD.Makernote
 ``ExifTags.IFD.MakerNote``.
 
 Image getdata()
-~~~~~~~~~~~~~~~
+^^^^^^^^^^^^^^^
 
 .. deprecated:: 12.1.0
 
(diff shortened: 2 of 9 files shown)


docs/releasenotes/13.0.0.rst
```diff
@@ -69,6 +69,12 @@ ImageCms.ImageCmsProfile.product_name and .product_info
 Deprecations
 ============
 
+IM image format
+^^^^^^^^^^^^^^^
+
+The IM image format has been deprecated. If you are using this format and would like
+to continue doing so, please provide a report about what other software uses it.
```
hugovk:
```suggestion
The IM image format has been deprecated and removal is planned in Pillow 15.0.0
(2028-10-15). If you are using this format and would like to continue doing so,
open an issue about what other software uses it.
```

docs/deprecations.rst
```diff
@@ -30,6 +30,13 @@ Image getdata()
 identical, except that it returns a tuple of pixel values, instead of an internal
 Pillow data type.
 
+IM image format
+^^^^^^^^^^^^^^^
+
+.. deprecated:: 13.0.0
+
+The IM image format has been deprecated.
```
hugovk:
```suggestion
The IM image format has been deprecated and removal is planned in Pillow 15.0.0
(2028-10-15). If you are using this format and would like to continue doing so,
open an issue about what other software uses it.
```

Tests/test_mode_i16.py
```diff
@@ -46,7 +46,7 @@ def test_basic(tmp_path: Path, mode: str) -> None:
     im_out = im_in.transform((w, h), Image.Transform.EXTENT, (0, 0, w, h))
     verify(im_out)  # transform
 
-    filename = tmp_path / "temp.im"
+    filename = tmp_path / "temp.tiff"
```
hugovk, line 49: There's an uncaught deprecation warning in `Tests/test_imagesequence.py`'s `test_sanity` that saves to `"temp.im"`.

Also change that to `"temp.tiff"`?

radarhere: Thanks. I've pushed a commit.

hugovk (approved):
I couldn't find any use of this Pillow plugin on GitHub, and couldn't find any other software that reads IM.

https://github.com/python-pillow/Pillow/pull/9800 was broken for three decades, found via LLM code inspection, not actual user report.

Updates to `ImImagePlugin.py` have been type hints, modernisation, rearrangement, cleanup, nothing looks to be triggered by user reports.

ImageIO wraps IM format through Pillow, listed as a legacy plugin:
https://imageio.readthedocs.io/en/stable/formats/index.html They'll need to remove before the deprecation, possibly with their own deprecation.

Some meandering history.

The docs say:

> IM is a format used by LabEye and other applications based on the IFUNC image processing library. The library reads and writes most uncompressed interchange versions of this format.

The source history comment shows it's been around since the start:

`# 1995-09-01 fl   Created.`

The full name of the format is given as `format_description = "IFUNC Image Memory"`.


"The image processing system, LabEye Profile" is mentioned in https://analyticalsciencejournals.onlinelibrary.wiley.com/doi/10.1002/elps.1150130186. Two of the authors are from the State Institute of Forensic Serology, University Hospital, Linköping, Sweden; the other two from Innovativ Vision AB, Linköping, Sweden.

This 2002 Innovativ Vision company page makes no mention of LabEye, only a WoodEye product:
* https://web.archive.org/web/20020809011844/http://www.ivab.se/ivabeng/foretaget/index.html

Innovativ Vision is now Image Systems Group:
* https://www.imagesystemsgroup.se/about-us/image-systems-group

But Woodeye is part of Microtec:
* https://www.microtec.com/en/products/woodeye
* https://wtp.hoechsmann.com/en/lexikon/26528/woodeye

Fredrik Lundh and his company Secret Labs AB were also from Linköping:
* https://web.archive.org/web/20010208144916/http://www.pythonware.com/company/contact.htm 
* https://www.python.org/psf/records/board/minutes/2001-03-05/

I found no mention of IFUNC or LabEye on the image-sig mailing list. There are three mentions of the IM format, all by Fredrik.

* refers to "lesser known ARG and IM formats": https://mail.python.org/pipermail/image-sig/1998-July/000506.html
* An offset bugfix: https://mail.python.org/pipermail/image-sig/1998-July/000518.html
* A mention of `lenna.im`: https://mail.python.org/pipermail/image-sig/2000-October/001230.html

No user reports.

Aha! Looks like Fredrik used to work at Innovativ Vision AB. His first post in March 1996 to image-sig was announcing the PIL library and its spec, sent from `fredrik_lundh@ivab.se`:

* https://mail.python.org/pipermail/image-sig/1996-March/000004.html

https://www.ivab.se redirects to https://www.microtec.com/

His last post with this address was March 1997, and in March 1999 someone mentioned the address was bouncing.

* https://mail.python.org/pipermail/image-sig/1997-March/000233.html
* https://mail.python.org/pipermail/image-sig/1999-March/000680.html

So looks like Fredrik added this format because he used it at work, but that software isn't used any more, the IM image format appears to have been proprietary, didn't take off and isn't used elsewhere, and we're safe to deprecate and remove after two years.

radarhere:
Impressive research.

PR #9850, opened 2026-08-05
Remove ExifTags.IFD.Makernote

radarhere:
Deprecated in https://github.com/python-pillow/Pillow/pull/8629


akx (approved):
Begone alternate spelling! 😄

hugovk:
Thanks, noting this was deprecated in 11.1.0 (Jan 2025) without a deprecation warning because there wasn't a straightforward way of doing so.

And the next Pillow 13.0.0 is the earliest this can be removed.

I only found one non-Pillow use of the old name on GitHub, so opened https://github.com/Nayu-u/Kingambit/pull/1 to fix it.

PR #9498, opened 2026-03-25
Use critical sections to protect FontObject

colesbury:
FreeType FT_Face objects are not thread-safe. Use per-object critical sections to protect FontObject methods that access the underlying FT_Face in the free-threaded build.

Fixes #9497.

PR #9853, opened 2026-08-05
Use a critical section to protect getlist()

### LINKED ISSUE #9852: Free-threaded: Image.point(list) crashes if another thread mutates the list (getlist / PySequence_Fast, stale size)
`getlist()` in `src/_imaging.c` — used by `Image.point()` — reads a list's size, aliases the list with `PySequence_Fast`, and walks it with the unchecked `PySequence_Fast_GET_ITEM` macro over that captured size. On a free-threaded build, another thread shrinking the list mid-walk drives an out-of-bounds read and a segfault. Present on current `main`.

This is a concrete instance of the already-documented hazard that `PySequence_Fast_GET_ITEM` / `_GET_SIZE` do no locking (CPython's free-threading extension docs; numpy/numpy#28046) — not a new class, but I couldn't find this specific `getlist`/`point` site reported or fixed (it isn't touched by #9743, which speeds up `libImaging/Point.c`, nor by the #8389 cleanup), so filing it as another concrete case alongside the thread-safety hardening in #8492 / #9498.

### Site (current `main`, `src/_imaging.c`)

```c
static UINT8 *getlist(PyObject *arg, Py_ssize_t *length, ...) {
    n = PySequence_Size(arg);                      // 445  size captured once
    if (length && wrong_length && n != *length) ...// 447  point() requires n == 256
    seq = PySequence_Fast(arg, must_be_sequence);  // 458  for a LIST, aliases it (no copy)
    for (i = 0; i < n; i++) {
        op = PySequence_Fast_GET_ITEM(seq, i);     // 465  unchecked macro, stale n
        itemp = PyLong_AsLong(op);                 // dereferences op
```

`Image.point(lut)` → `_point` (:1505) → `getlist` (:1526/1538/1559). If the list is shrunk after line 445's size check, `PySequence_Fast_GET_ITEM(seq, i)` for `i >= new_len` reads past the reallocated storage and `PyLong_AsLong` dereferences a dangling / OOB pointer.

### Reproducer

`python3.14.0rc1t`, Pillow 12.3.0 (official `cp314t` wheel — GIL is off by default). 12 threads call `img.im.point(shared, "L")`; 6 threads hard-shrink `shared` (256→16) and refill. 10 rounds per arm:

| arm | result |
|---|---|
| mutate, GIL off (default on 3.14t) | **10/10 SIGSEGV** |
| control — no mutator | clean 10/10 |
| control — mutate a *different* list | clean 10/10 |
| control — same mutation, `PYTHON_GIL=1` | clean 10/10 |

The lut entries must be **non-cached** ints (> 256): with `range(256)` every entry is an immortal small int, so a dropped slot still points at a live int and it does *not* crash even though the index is already out of bounds — "no crash" isn't "safe" here.

<details><summary>ft_pillow.py</summary>

```python
"""Pillow 12.3.0 — P3 stale-size in _point/getlist over a caller list (iter/index invalidation).

_imaging.c getlist():
    n = PySequence_Size(arg);                          // 446: length captured
    if (length && wrong_length && n != *length) ...    // 447: must equal expected (256 for point)
    seq = PySequence_Fast(arg, ...);                   // 459: ALIASES the list (no copy)
    for (i = 0; i < n; i++)
        op = PySequence_Fast_GET_ITEM(seq, i);         // 466: unchecked macro, stale n
        itemp = PyLong_AsLong(op);                     // dereferences op

Reached by Image.point (_point, _imaging.c:1505 → getlist at 1526/1538/1559). The lut must be exactly 256 (else ValueError at 447), so the race window is: the list is 256 at the size check, then another thread shrinks it during the 256-iteration walk → PySequence_Fast_GET_ITEM(seq, i>=newsize) is OOB.

LIVE-BY-DEFAULT: Pillow ships a cp314t free-threaded wheel and does not re-enable the GIL on import.
"""
import os, sys, threading, time
from PIL import Image

N = int(os.environ.get("N", 8))
N_MUT = int(os.environ.get("N_MUT", 3))
SECONDS = float(os.environ.get("SECONDS", 5))
MUTATE = os.environ.get("MUTATE", "1") == "1"
DECOY = os.environ.get("DECOY", "0") == "1"

def full():
    # 256 valid lut entries, but HEAP ints (> 256, so not immortal small-int cached):
    # when the list shrinks and drops its ref, these are actually freed → the aliased
    # slot dangles. (range(256) would be immortal small ints and never dangle.)
    return [10_000_000 + i for i in range(256)]

def main():
    gil = sys._is_gil_enabled()
    print(f"py={sys.version.split()[0]} gil={gil} callers={N} mutators={N_MUT if MUTATE else 0} decoy={DECOY}", flush=True)
    shared = full()
    decoy = full()
    core = Image.new("L", (16, 16)).im     # the C ImagingObject; core.point(lut, mode) -> _point
    stop = threading.Event()
    barrier = threading.Barrier(N + (N_MUT if MUTATE else 0) + 1)
    ok = [0] * N

    def caller(t):
        barrier.wait()
        c = 0
        while not stop.is_set():
            try:
                core.point(shared, "L")     # -> _point -> getlist(shared) -> the aliased walk
            except Exception:
                pass                        # ValueError when len != 256 mid-toggle is irrelevant
            c += 1
        ok[t] = c

    def mutator():
        barrier.wait()
        target = decoy if DECOY else shared
        while not stop.is_set():
            # keep it mostly 256 (so the size check passes), then hard-truncate so a
            # call that already passed the check walks a now-shorter, realloc'd list
            # whose freed slots dangle. Refill with FRESH heap ints so the dropped ones
            # are actually released.
            del target[16:]                                  # 256 -> 16 (reallocs down)
            target.extend(10_000_000 + i for i in range(16, 256))  # back to 256, fresh ints

    threads = [threading.Thread(target=caller, args=(t,)) for t in range(N)]
    if MUTATE:
        threads += [threading.Thread(target=mutator) for _ in range(N_MUT)]
    for th in threads: th.start()
    barrier.wait()
    time.sleep(SECONDS)
    stop.set()
    for th in threads: th.join()
    print(f"clean: {sum(ok)} point() calls survived", flush=True)
    return 0

if __name__ == "__main__":
    raise SystemExit(main())
```
</details>

### A possible fix

Wrap the walk in `Py_BEGIN_CRITICAL_SECTION(arg)` (as done for FontObject in #9498), or materialize the items with the bounds-checked `PySequence_GetItem` / re-read the length each iteration instead of the `PySequence_Fast_GET_ITEM` macro. The tuple-input path is safe (tuples can't resize); the list path is not.

### Not claimed

Thread-safety / robustness, not a security issue — it needs the caller's own code to concurrently mutate the lut. No severity rating; reported so it can be fixed.

lazerg:
Fixes #9852. Alternative to #9854

`getlist()` reads the sequence length once, then walks the list with `PySequence_Fast_GET_ITEM`, which does no bounds checking and no locking. On a free-threaded build, another thread shrinking the list during the walk makes the remaining reads go past the end of the storage, so Image.point() segfaults. Wrapping the walk in a critical section on the argument keeps the length valid for the duration, the same approach used for FontObject in #9498.


espressolee:
While you're in `getlist()` — the critical section closes the free-threading race, but there's a second, unrelated bug in the same function that it can't reach, because it isn't a race at all.

`n` comes from the caller's `__len__`, and `seq` is built by *iterating*, so for a non-list they need not agree:

```c
n = PySequence_Size(arg);          /* -> __len__ */
...
seq = PySequence_Fast(arg, must_be_sequence);   /* iterates; may be shorter */
for (i = 0; i < n; i++) {
    op = PySequence_Fast_GET_ITEM(seq, i);      /* reads past the end when n > len(seq) */
```

Deterministic, single-threaded, GIL on:

```python
class Lying:
    def __len__(self): return 256          # claims 256
    def __getitem__(self, i):              # yields 8, then stops
        if i >= 8: raise IndexError
        return float(i)

Image.new("L", (4, 4)).point(Lying(), "F")
```

`Image.point` passes the sequence straight through for mode `"F"` (`Image.py` only materialises it with a list comprehension when the mode differs), so `getlist()` sees the original object. The same path is reachable via `ImageFilter.Kernel`, `Image.transform` and the JPEG qtables.

Clamping the loop to the sequence actually returned fixes it:

```c
Py_ssize_t seq_n = PySequence_Fast_GET_SIZE(seq);
if (n > seq_n) { n = seq_n; }
```

**Not raising this as a security issue** — the trigger is an object the caller supplies, so anything that can reach it is already running Python in the process. Flagging it because a critical section around the existing loop would leave it in place, and it seemed worth having in front of you while the function is open.

lazerg:
Thanks, you're right about `point()`. On a GIL-enabled 3.14 build, single threaded, your example segfaults for me every time. `Image.transform()` goes the same way if the object's `__getitem__` also handles the `data[:8]` slice that `Image.py` takes first.

Two corrections on the other paths. `ImageFilter.Kernel` does read past the end, but here it surfaced as `TypeError: bad argument type for built-in operation` rather than a crash, because the slot past the end happened to read as NULL and `PyFloat_AsDouble()` rejects that. The JPEG qtables path isn't reachable: `validate_qtables()` does `list(array.array("H", table))` on each table, so C only ever sees a real list.

On the fix, I'd rather not clamp. In `_point()` the caller declared 256 entries and the length check passed on the lie, so clamping hands them a LUT that is zero-filled from index 8 on and no error. Taking the length from `seq` and running the existing check against that reads better to me:

```c
if (!seq) {
    return NULL;
}

n = PySequence_Fast_GET_SIZE(seq);
if (length && wrong_length && n != *length) {
    PyErr_SetString(PyExc_ValueError, wrong_length);
    Py_DECREF(seq);
    return NULL;
}
```

The `calloc()` then moves below this and the `PySequence_Size()` call goes away. The `PySequence_Check()` guard above it stays, since `PySequence_Fast()` on its own would happily accept a set.

I built that locally: your example now raises "wrong number of lut entries", the transform one raises "wrong number of matrix entries", `Kernel` reaches its own "bad kernel size" check, and the suite is unchanged at 4900 passed on 3.14t.

It is independent of the race, though. The critical section doesn't help here, and this doesn't close the race either, since a real list can still shrink between the size read and the loop. So I'd keep this PR on #9852 and raise the other one separately, unless a maintainer would rather have both here.

espressolee:
Thanks for checking it properly — and you're right on both corrections.

**The qtables path was my error.** `validate_qtables()` does `qtables[idx] = list(array.array("H", table))`, and `array.array` *iterates*, so even a `__len__` that lies its way past the `len(table) != 64` check produces a real list of the true length before C sees anything. I shouldn't have listed it; it isn't reachable. And thank you for pinning `Image.transform()` to the case where `__getitem__` also serves the `data[:8]` slice (`Image.py:3022`) — my note was looser than that.

**Your fix is better than the clamp and I'd drop mine.** The clamp silently hands back a zero-filled LUT *and* writes the clamped count through `*length`, so `_point()` gets 8 where it asked for 256, with no error anywhere. Taking `n` from `PySequence_Fast_GET_SIZE(seq)` runs the existing `wrong_length` check against reality and makes `*length` truthful, which is the behaviour the function already documents.

One thing that may sharpen the "independent of the race" point in your favour: the loop calls `PyLong_AsLong()` / `PyFloat_AsDouble()` on each item, and for a non-`int`/non-`float` those invoke `__index__` / `__float__` — arbitrary Python, inside the loop. And if that code *blocks* — I/O, or acquiring another lock — the critical section is suspended and its locks released, then reacquired afterwards (`Doc/howto/free-threading-extensions.rst`, "How Critical Sections Work"). So a size read once at the top isn't a bound you can rely on even under the section. Snapshotting into a private list, or re-reading the size each iteration, looks like the only thing that closes that, which is a third change again from either of these two.

Agreed on keeping this PR scoped to the race and #9852. Happy for you to raise the size one — you have the tested branch and the better patch, so it should be yours; I'll bring the reproducers over and drop the clamp from #9854 so there's nothing overlapping to review. Say the word if you'd rather I opened it instead.

lazerg:
Thanks, and yes, please leave the size one to me. I'll raise it separately with the reproducers I already have, so nothing overlaps with #9854.

Your read of the docs is right, a critical section is suspended and its locks released if the thread blocks. In `getlist()` that only bites when the items are not `int` or `float`, because `PyLong_AsLong()` and `PyFloat_AsDouble()` then call `__index__` or `__float__` and that is arbitrary Python. For plain numeric items, which is the aliased-list case in #9852, nothing in the loop can run Python code or block, so the section holds for the whole walk. A snapshot would close the wider case, but as you say that is a third change, so this PR stays on the race.

espressolee:
Quick coordination check on the separate size/materialization bug: Andrew replied on GHSA-9w28-qj75-mq69 on August 17 that a normal issue or PR is welcome. I searched the repository again today and do not see the separate report yet.

You asked on August 8 to keep that report on your side, so I have not opened a duplicate issue or PR. I did prepare the independent fix on my fork at `41b675109723dcd0fab6f935518501992e0b26c6` to make sure the path is understood:

- `Image.point`, `putdata`, `transform`, and `ImageFilter.Kernel` public regressions
- Python 3.14: 332/332 affected-file tests pass
- Python 3.14t with the GIL disabled: 332/332 pass
- changed-file pre-commit: pass
- full suite on both interpreters: 4,888 pass with the same unrelated sandbox `ImageGrab.grab()` display-capture failure

Are you still planning to raise the separate issue/patch? If so, I will leave the branch parked. If not, I can take it forward without overlapping #9853. JPEG qtables remains excluded, per the correction above.

lazerg:
Checked again. Still haven't opened the separate issue.

I searched open and closed issues in python-pillow/pillow and found nothing matching the getlist() size/materialization bug. #9852 is still the only tracked report.

I'm behind on this. Go ahead and open it yourself with your branch and reproducers if you want to move now. I don't want to keep blocking you. If you'd rather I file it, say so and I'll do it this week.

espressolee:
Thanks — I opened #9892 with the two deterministic reproducers and linked the
existing candidate at `41b675109723dcd0fab6f935518501992e0b26c6`.

It keeps the size/materialization bug separate from #9852/#9853, covers
`getlist()` and `_putdata()`, and explicitly excludes JPEG qtables per the
correction above. I will leave this PR scoped to your free-threading race fix.

espressolee:
Reporter of #9852 here. I independently re-ran the reproducer against the current PR head (`189a6fb`), which is the commit I tested locally on 2026-08-31. It eliminated the race in this harness.

CPython 3.14.6t, macOS arm64, 12 caller threads and 6 mutator threads, 5s per process, fresh subprocesses, heap integers outside the small-int cache:

| source | shared-list mutation, GIL off | no mutator, GIL off | decoy-list mutation, GIL off | shared-list mutation, GIL on |
| --- | ---: | ---: | ---: | ---: |
| base `e76b081` | 10/10 SIGSEGV | 10/10 clean | 10/10 clean | 10/10 clean |
| head `189a6fb` | 10/10 clean | 10/10 clean | 10/10 clean | 10/10 clean |

The controls help attribute the result: removing the mutator, or mutating a decoy list instead of the list being read, leaves the base clean.

Local suite at head: 4,889 passed, 283 skipped, 3 xfailed, with one unrelated macOS ImageGrab failure (`screencapture` could not capture the display). The affected call-site tests passed: 352 passed, 2 skipped because NumPy was unavailable.

This verification is specific to the #9852 concurrent list-mutation mechanism; it does not cover the separately scoped Python re-entry or size/materialization cases discussed elsewhere. I don't see those as blockers for this fix.

This PR does not add a concurrent regression test for #9852. I'm happy to prepare a separate follow-up test based on this harness after this fix lands.

lazerg:
Thanks for the independent rerun against 189a6fb, that's useful confirmation. A follow-up concurrency test based on your harness sounds good, happy to see it as a separate PR once this one lands.

radarhere (approved):

PR #9854, opened 2026-08-05
Protect getlist() walk with a critical section (free-threading, #9852)

### LINKED ISSUE #9852: Free-threaded: Image.point(list) crashes if another thread mutates the list (getlist / PySequence_Fast, stale size)

espressolee:
Fixes #9852. Alternative to #9853

`getlist()` (used by `Image.point()`) walks the caller's list with `PySequence_Fast` and the unchecked `PySequence_Fast_GET_ITEM` macro over a size captured earlier in the function. On a free-threaded build `PySequence_Fast` returns the list itself for a list input, so the walk aliases the caller's list; another thread resizing it drives an out-of-bounds read and a segfault. This is #9852.

This holds a critical section on the fast sequence for the walk and clamps the loop to its current length, so a resize that happened between the earlier size read and the walk cannot read out of bounds. It mirrors the critical-section approach already used for `FontObject` in #9498, and uses the `Py_BEGIN_CRITICAL_SECTION` shim already vendored in `thirdparty/pythoncapi_compat.h` (a no-op block on < 3.13, where the GIL serialises the walk anyway).

### Verification

`python3.14.0rc1t`, built from source:

- The reproducer in #9852 (12 threads `img.im.point(shared, "L")` while others resize `shared`) goes from **10/10 SIGSEGV to 0/10** with this change; the no-mutator, decoy and GIL-on controls stay clean.
- `Tests/test_image_point.py` passes (4/4), and `point()` output is unchanged for normal look-up tables (identity and offset LUTs verified).

A concurrent resize now yields a truncated read rather than a crash; if you'd prefer it to raise instead, I'm happy to adjust.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/_imaging.c
+++ b/src/_imaging.c
@@ -97,6 +97,8 @@
 
 #include "libImaging/Imaging.h"
 
+#include "thirdparty/pythoncapi_compat.h"
+
 #define _USE_MATH_DEFINES
 #include <math.h>
 #include <stddef.h>
@@ -461,7 +463,17 @@ getlist(PyObject *arg, Py_ssize_t *length, const char *wrong_length, int type) {
         return NULL;
     }
 
-    for (i = 0; i < n; i++) {
+    // On a free-threaded build PySequence_Fast returns the list itself for a list
+    // input, so the walk below aliases the caller's list. Hold a critical section on
+    // it and clamp to its current length, so a concurrent resize cannot make
+    // PySequence_Fast_GET_ITEM read out of bounds (#9852). `n` was read before this
+    // point and may be stale.
+    Py_BEGIN_CRITICAL_SECTION(seq);
+    Py_ssize_t count = PySequence_Fast_GET_SIZE(seq);
+    if (count > n) {
+        count = n;
+    }
+    for (i = 0; i < count; i++) {
         op = PySequence_Fast_GET_ITEM(seq, i);
         // DRY, branch prediction is going to work _really_ well
         // on this switch. And 3 fewer loops to copy/paste.
@@ -484,6 +496,7 @@ getlist(PyObject *arg, Py_ssize_t *length, const char *wrong_length, int type) {
                 break;
         }
     }
+    Py_END_CRITICAL_SECTION();
 
     Py_DECREF(seq);
 


espressolee:
Cross-referencing #9853 — that PR was opened before this one and it fixes the free-threading race in `getlist()`: wrapping `getlist_impl` in `Py_BEGIN_CRITICAL_SECTION(arg)` puts `PySequence_Size` inside the section as well, so the count and the walk see the same object state. Priority is @lazerg's and I am happy for that one to land instead of this.

I will hold this PR open but idle rather than pushing changes to it, and follow whatever the maintainers prefer.

espressolee:
Closing in favour of #9853, which was opened first and is the better-scoped fix.

It takes the critical section on `arg` around the whole of `getlist()`, which also covers the
`PySequence_Size()` → `PySequence_Fast()` window; this PR took it on `seq` around the loop only.
For a list those are the same object, but for a non-list `seq` is a fresh object no other thread
can reach, so locking it does nothing and `arg` is the one that matters.

The other half of this PR — clamping the walk to `PySequence_Fast_GET_SIZE(seq)` — is a separate,
non-threading bug: `n` comes from the caller's `__len__` and `PySequence_Fast` builds by
iterating, so they need not agree. It is being raised separately, and with a better fix than the
clamp: read `n` from the sequence and let the existing `wrong_length` check fire, so the caller
gets an error instead of a silently zero-filled result. Discussion in #9853.

#9852 remains the issue of record for the race. Nothing here is lost; it is only in two better
places.

PR #9859, opened 2026-08-09
Apply dither when converting/quantizing to an adaptive palette

### LINKED ISSUE #5836: Inconsistent use of dithering in `convert` and `quantize`
### What did you do?

Let's play around with image dithering, starting with this nice, low-res picture of Theresa May:

image

According to the [docs](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.convert) calling `convert` in "P" mode defaults to the `WEB` palette with dithering:

```
orig.convert("P")
```

image

Nice, that looks good! And you can pass `IMAGE.NONE` to disable the dithering.

What if we want to use less colors? Maybe we cut down the bit depth and dither to make up for it, for a retro effect.
Checking the docs again, it claims that we can use an `ADAPTIVE` palette and pass in some number of `colors`; let's use 16.
The docs say that dithering is used when converting from "RGB" to "P" and that it defaults to `FLOYDSTEINBERG`, but let's pass in `dither` just to make sure:

```
dither_lesscol_broken = orig.convert(
    "P", palette=Image.ADAPTIVE, colors=16, dither=Image.FLOYDSTEINBERG
)
```

image

Doesn't look dithered - test it on a gradient image and we get severe banding:

image

image

Is there a workaround? Well, the docs for [quantize](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.quantize) claim to apply a dither.
Let's try it:

```
orig.quantize(16, dither=Image.FLOYDSTEINBERG)
```

image

No dice, exactly the same as the previous. Looking at the [code](https://pillow.readthedocs.io/en/stable/_modules/PIL/Image.html#Image.quantize), we see that `quantize` calls `ImagingCore.convert` with a dither - but ONLY if you pass in a palette!
Let's try that:

```
# Call .quantize once to get the palette
pal = orig.quantize(16)
# And again using that palette
dither_lesscol = orig.quantize(16, palette=pal, dither=Image.FLOYDSTEINBERG)
```

image

There we go, that's what I expected.

`Image.convert` should be able to simultanously handle an `ADAPTIVE` palette and apply dithering (or not) based on the `dither` argument.

Additionally, `Image.quantize` should apply dithering (or not) based on the `dither` argument.

### What actually happened?

`orig.convert("P", palette=Image.ADAPTIVE, colors=16, dither=Image.FLOYDSTEINBERG)` produces a non-dithered image.

`orig.quantize(16, dither=Image.FLOYDSTEINBERG)` also produces a non-dithered image.

* OS: Linux Mint 20.1
* Python: 3.8.10
* Pillow: 7.0.0

83Gh0st:
Fixes #5836.

`dither` was silently ignored whenever no reference `palette` was given (e.g. `im.convert("P", palette=Image.Palette.ADAPTIVE, dither=...)` or plain `im.quantize(dither=...)`), because the underlying `self.im.quantize()` C call has no dithering support without an explicit reference palette. This re-converts through the computed adaptive palette when dithering is explicitly requested, so `dither` is honored in both `convert()` and `quantize()`. Calls that don't pass `dither` are unaffected.

* Apply dithering to `Image.convert(..., palette=Palette.ADAPTIVE, dither=...)`
* Apply dithering to `Image.quantize(..., dither=...)` when no reference palette is given
* Add regression tests + a release note

### DIFF
--- a/Tests/test_image_quantize.py
+++ b/Tests/test_image_quantize.py
@@ -105,6 +105,40 @@ def test_quantize_dither_diff() -> None:
     assert dither.tobytes() != nodither.tobytes()
 
 
+def test_quantize_adaptive_dither_diff() -> None:
+    # https://github.com/python-pillow/Pillow/issues/5836
+    # dither was being silently ignored when no reference palette was given
+    # (e.g. plain quantize(colors=N)), since self.im.quantize() doesn't
+    # support dithering directly.
+    image = hopper()
+
+    no_dither = image.quantize(colors=4, dither=Image.Dither.NONE)
+    dither = image.quantize(colors=4, dither=Image.Dither.FLOYDSTEINBERG)
+    default = image.quantize(colors=4)
+
+    assert dither.tobytes() != no_dither.tobytes()
+    # unspecified dither must remain backwards compatible with no dithering
+    assert default.tobytes() == no_dither.tobytes()
+
+
+def test_quantize_dither_reference_palette_default_unchanged() -> None:
+    # The reference-palette path already dithered by default; make sure
+    # changing the parameter default (Dither.FLOYDSTEINBERG -> None) didn't
+    # change that.
+    image = hopper()
+    with Image.open("Tests/images/caption_6_33_22.png") as palette:
+        palette_p = palette.convert("P")
+
+    default = image.quantize(palette=palette_p)
+    explicit_dither = image.quantize(
+        dither=Image.Dither.FLOYDSTEINBERG, palette=palette_p
+    )
+    no_dither = image.quantize(dither=Image.Dither.NONE, palette=palette_p)
+
+    assert default.tobytes() == explicit_dither.tobytes()
+    assert default.tobytes() != no_dither.tobytes()
+
+
 @pytest.mark.parametrize(
     "method", (Image.Quantize.MEDIANCUT, Image.Quantize.MAXCOVERAGE)
 )
--- a/src/PIL/Image.py
+++ b/src/PIL/Image.py
@@ -1052,8 +1052,12 @@ class Image:
            should be 4- or 12-tuple containing floating point values.
         :param dither: Dithering method, used when converting from
            mode "RGB" to "P" or from "RGB" or "L" to "1".
-           Available methods are :data:`Dither.NONE` or :data:`Dither.FLOYDSTEINBERG`
-           (default). Note that this is not used when ``matrix`` is supplied.
+           Available methods are :data:`Dither.NONE` or :data:`Dither.FLOYDSTEINBERG`.
+           :data:`Dither.FLOYDSTEINBERG` is used by default, except when converting
+           to an adaptive ``palette``: there, for backwards compatibility with
+           versions prior to 13.0.0, the default remains :data:`Dither.NONE`; pass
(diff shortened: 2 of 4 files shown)


Tests/test_image_quantize.py
```diff
@@ -105,6 +105,40 @@ def test_quantize_dither_diff() -> None:
...
+    # unspecified dither must remain backwards compatible with no dithering
+    assert default.tobytes() == no_dither.tobytes()
+
+
+def test_quantize_dither_reference_palette_default_unchanged() -> None:
+    # The reference-palette path already dithered by default; make sure
+    # changing the parameter default (Dither.FLOYDSTEINBERG -> None) didn't
+    # change that.
+    image = hopper()
+    with Image.open("Tests/images/caption_6_33_22.png") as palette:
```
radarhere, line 129: Why this particular image?

83Gh0st: No particular significance, that image is already the file's de facto standard for quantize tests, used in three other tests here: test_quantize (L60), test_quantize_no_dither (L69), and test_quantize_dither_diff (L99). I matched that existing convention rather than introducing a new fixture.

https://github.com/python-pillow/Pillow/blob/main/Tests/test_image_quantize.py#L60-L99

src/PIL/Image.py
```diff
@@ -1297,7 +1309,9 @@ def quantize(
         :param dither: Dithering method, used when converting from
            mode "RGB" to "P" or from "RGB" or "L" to "1".
            Available methods are :data:`Dither.NONE` or :data:`Dither.FLOYDSTEINBERG`
-           (default).
+           (default). Prior to this, dithering was silently ignored when no
```
```suggestion
           (default). Prior to 13.0.0, dithering was silently ignored when no
```
Reading this docstring on its own, I would have no idea what 'this' is.

Tests/test_image_quantize.py
```diff
@@ -105,6 +105,40 @@ def test_quantize_dither_diff() -> None:
...
+    # support dithering directly.
+    image = hopper()
+
+    no_dither = image.quantize(colors=4, dither=Image.Dither.NONE)
+    dither = image.quantize(colors=4, dither=Image.Dither.FLOYDSTEINBERG)
+    default = image.quantize(colors=4)
+
+    assert dither.tobytes() != no_dither.tobytes()
+    # unspecified dither must remain backwards compatible with no dithering
+    assert default.tobytes() == no_dither.tobytes()
```
radarhere, line 121: I appreciate that you're being backwards compatible, but doesn't this mean that the docstring of
> Available methods are :data:`Dither.NONE` or :data:`Dither.FLOYDSTEINBERG` (default)

should be updated?

83Gh0st: Ah, thank you. you're right, "(default)" isn't really true for the no-reference-palette path. I've updated both convert()'s and quantize()'s dither docstrings to make that clearer: FLOYDSTEINBERG is the default when a reference palette is given, but without one (e.g. an adaptive palette), it stays Dither.NONE by default for backwards compatibility, and dither=Dither.FLOYDSTEINBERG needs to be passed explicitly to get dithering there. Just pushed that. Let me know if it's still not quite right!

(Also, thanks for the earlier fix)

PR #9862, opened 2026-08-11
Make pybind11 an optional build dependency

commodo:
pybind11 is only used for pybind11.setup_helpers.ParallelCompile, which parallelises the compilation of the C extensions; it provides no runtime bindings. Wrap its import and use in try/except ImportError and drop it from build-system.requires, so the wheel builds without pybind11 (compiling serially) while still using it for parallel builds when it is installed.

Keep it discoverable via a PEP 735 "build" dependency group, so anyone who wants parallel source builds can install it with e.g. "pip install --group build". This lets distributions that do not want the extra build dependency (e.g. OpenWrt) build Pillow without pulling in pybind11.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,12 +7,6 @@ backend-path = [
   "_custom_build",
 ]
 
-[dependency-groups]
-# Optional; enables parallel C extension compilation (see setup.py).
-build = [
-  "pybind11",
-]
-
 [project]
 name = "pillow"
 description = "Python Imaging Library (fork)"
@@ -90,6 +84,12 @@ urls.Funding = "https://tidelift.com/subscription/pkg/pypi-pillow?utm_source=pyp
 urls.Homepage = "https://python-pillow.org"
 urls.Mastodon = "https://fosstodon.org/@pillow"
 urls.Source = "https://github.com/python-pillow/Pillow"
+
+[dependency-groups]
+# Optional; enables parallel C extension compilation (see setup.py).
+build = [
+  "pybind11",
+]
 
 [tool.setuptools]
 packages = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,11 +1,16 @@
 [build-system]
 build-backend = "backend"
 requires = [
-  "pybind11",
   "setuptools>=77",
 ]
 backend-path = [
   "_custom_build",
+]
+
+[dependency-groups]
+# Optional; enables parallel C extension compilation (see setup.py).
+build = [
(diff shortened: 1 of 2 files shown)


radarhere:
Hi. Is pybind11 as a dependency causing you any problems? Or is it purely that you don't see a need for one more required dependency?

commodo:
Mostly the latter: I don't see a need for one more required dependency.
We've packaged pybind11 solely for Pillow, on OpenWrt.
We picked up pybind11 at an older version.

Then our internal scripts (which report messages to maintainers) mentioned to upgrade pybind11.
Did that via https://github.com/openwrt/packages/commit/a3b6bdbee565e3aac067d026d1e02e2b633cc68b#commitcomment-195742530

Then we found that pybind11, needs a new dependency of it's own.
```
requires = ["scikit-build-core >=0.11.2"]
```
Link:  https://github.com/pybind/pybind11/blob/master/pyproject.toml#L2

So, naturally, in order to avoid this, it's easier (for me) to drop pybind11.
Link:  https://github.com/openwrt/packages/pull/30246/changes

I'm ok to keep the patch in our internal tree.
But I'm also happy to talk to upstream to find a middle ground that works elegantly for everybody.


As a context: I've been maintaining these Python packages/packaging (in OpenWrt) for quite a while.
I like some packages (like Pillow or numpy).
In recent years though, a lot of granular dependencies have appeared.
I miss the olden days with "Python, batteries included" motto :)

pyproject.toml
```diff
@@ -86,6 +85,12 @@ urls.Homepage = "https://python-pillow.org"
 urls.Mastodon = "https://fosstodon.org/@pillow"
 urls.Source = "https://github.com/python-pillow/Pillow"
 
+[dependency-groups]
+# Optional; enables parallel C extension compilation (see setup.py).
+build = [
+  "pybind11",
+]
```
akx, line 92: I don't think this will do the right thing -- it will install pybind11 in the _runtime_ environment with the built Pillow package, not in the PEP517 build environment.

akx: The `ParallelBuild` code itself isn't very long and could be vendored into Pillow: 

https://github.com/pybind/pybind11/blob/856a3485ca5b704fc7614e89540670cb553e6e4f/pybind11/setup_helpers.py#L360-L493

That ties to #9645 though :)

commodo: Oh
That's a neat alternative.

Happy to wait for that one.
Will close this PR and wait for a release bump.

PR #9863, opened 2026-08-11
Use PyErr_Format instead of sprintf

akx:
Just noticed a couple places using sprintf() instead of `PyErr_Format`, figured those could use unifying.

### DIFF
--- a/src/Tk/tkImaging.c
+++ b/src/Tk/tkImaging.c
@@ -247,12 +247,9 @@ _dfunc(HMODULE lib_handle, const char *func_name) {
      * Returns function pointer or NULL if not present.
      */
 
-    char message[100];
-
     FARPROC func = GetProcAddress(lib_handle, func_name);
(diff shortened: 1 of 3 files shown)


radarhere:
While I don't see a problem, it's worth being aware that this is rewriting the fix to https://github.com/advisories/GHSA-7534-mm45-c74v

akx:
@radarhere Thanks! Good note. I looked into it: `PyErr_Format` calls `_PyErr_FormatV`, which calls `PyUnicode_FromFormatV`, which uses `PyUnicodeWriter` (an automatically growing buffer) with `unicode_from_format`, so buffer overflows shouldn't be possible with `PyErr_Format`.

I reverted the (inadvertent, my bad!) template string changes back to `%.10s` though, so no chance of MemoryError either.

radarhere (approved):

wiredfool (approved):

Issue #9865, opened 2026-08-11
remap_palette (potential) incorrect behavior if source_palette has 768 elements


denilsonsa:
This is a quick potential one-character fix. But certainly longer, as it is better to write tests about it.

## What happens?

I was reading this code: 
(which was added in d522e0a5c03f87b653ab25a23116d9f685aa0c1e)

https://github.com/python-pillow/Pillow/blob/c8c74c8306eb7e1429a1a629a7471ca1794b7601/src/PIL/Image.py#L2234-L2236

And I realized this is probably buggy. Because `3 * 256 = 768`. But that code only detects RGBA `if len(source_palette) > 768`. It should have been `>=` instead of `>`. Probably.

### What did you do?

I just read the code. I did not write any code (yet). I did not use the `remap_palette` function. And, in fact, I do not intend to use it in the near future. I just read the code and found a potential bug. Consider it a late code review. ;)

### Proposed solution?

1. Write down some tests using RGBA palette.
2. Check if the tests fail. If my understanding is correct, the tests will fail.
3. Replace `>` with `>=`. The tests should now succeed.

Can someone from Pillow please investigate and check if this is the case?

Thanks! I hope my little bug report over here helps. :)

radarhere:
If I'm understanding correctly, then an assumption in your report is that palettes can only have 255 indexes. This is incorrect, as palettes can have 256 indexes.

```python
from PIL import ImagePalette
p = ImagePalette.ImagePalette("RGB", (0, 0, 0) * 255)
p.getcolor((0, 0, 1))  # 256 colours
p.getcolor((0, 0, 2))  # 257 colours
```
gives
```pytb
Traceback (most recent call last):
  File "PIL/ImagePalette.py", line 177, in getcolor
    return self.colors[color]
           ~~~~~~~~~~~^^^^^^^
KeyError: (0, 0, 2)

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "palette.py", line 4, in <module>
    p.getcolor((0, 0, 2))  # 257 colours
  File "PIL/ImagePalette.py", line 180, in getcolor
    index = self._new_color_index(image, e)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "PIL/ImagePalette.py", line 151, in _new_color_index
    raise ValueError(msg) from e
ValueError: cannot allocate more than 256 colors
```

[...]

PR #9867, opened 2026-08-11
Test remap_palette()'s source_palette argument

pranjalm37:
#9865 suggests that `remap_palette()`'s palette-mode detection is off by one and should use `>=` instead of `>`:

https://github.com/python-pillow/Pillow/blob/c8c74c8306eb7e1429a1a629a7471ca1794b7601/src/PIL/Image.py#L2234-L2236

I looked into it, and **the current `>` is correct** — changing it to `>=` would be a regression. The issue asked for tests, so this PR adds them rather than changing the behaviour.

## Why `>` is right

A full 256-entry palette is 768 bytes in RGB and 1024 bytes in RGBA, so the boundary sits *at* 768, not above it:

| source_palette | length | `> 768` | correct? |
|---|---|---|---|
| 256-entry RGB | 768 | RGB | ✅ |
| 256-entry RGBA | 1024 | RGBA | ✅ |

With `>=`, the first row flips to RGBA and a completely ordinary 256-colour RGB palette gets re-read 4 bytes at a time, shifting every colour. Verified against 12.3.0:

```python
from PIL import Image
pal = bytearray()
for i in range(256):
    pal += bytes([i, (100 + i) % 256, (200 + i) % 256])   # 768 bytes, RGB

im = Image.new("P", (2, 2))
out = im.remap_palette(list(range(256)), bytes(pal))
print(out.palette.mode)   # '>' (current): RGB   -- correct
                          # '>=':          RGBA  -- wrong
```

There *is* a real ambiguity underneath the report — a 192-entry RGBA palette is also exactly 768 bytes, and is read as RGB. But that's inherent to inferring the mode from a bare `bytes` length, and no choice of comparison operator fixes it; distinguishing those two cases would need the caller to say which mode it means. That seems like a separate API question, so I've left it alone here.

## Changes proposed in this pull request

* Add `test_remap_palette_source_palette`, parametrized over RGB/RGBA, covering `remap_palette()`'s `source_palette` argument. The existing `test_remap_palette` only exercises the `source_palette=None` path, so this branch had no coverage at all.
* Both sides of the 768-byte boundary are asserted, so the resulting test fails if the comparison is ever changed to `>=` (confirmed: the RGB case fails with that edit, and passes on `main` as-is).

Full `Tests/test_image.py` passes (185 passed, 1 skipped — IPython not installed locally). `ruff check` / `ruff format --check` clean; `mypy` reports no new errors (the 4 it emits are pre-existing missing-IPython-stub errors, identical on a clean tree).

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Tests/test_image.py
+++ b/Tests/test_image.py
@@ -674,6 +674,29 @@ class TestImage:
             with pytest.raises(ValueError):
                 im_hopper.remap_palette([])
 
+    @pytest.mark.parametrize(
+        "bands, palette_mode", ((3, "RGB"), (4, "RGBA")), ids=("RGB", "RGBA")
+    )
+    def test_remap_palette_source_palette(self, bands: int, palette_mode: str) -> None:
+        # When source_palette is passed explicitly its mode is inferred from its
+        # length, so a full 256-entry RGB palette (exactly 768 bytes) must stay
+        # RGB while a 256-entry RGBA one (1024 bytes) is detected as RGBA.
+        source_palette = bytes(
+            (entry + channel * 17) % 256
+            for entry in range(256)
+            for channel in range(bands)
+        )
+        assert len(source_palette) == 256 * bands
+
+        im = Image.new("P", (256, 1))
+        for x in range(256):
+            im.putpixel((x, 0), x)
+
+        im_remapped = im.remap_palette(list(range(256)), source_palette)
+        assert im_remapped.palette is not None
+        assert im_remapped.palette.mode == palette_mode
+        assert im_remapped.palette.palette == source_palette
+
     def test_remap_palette_transparency(self) -> None:
         im = Image.new("P", (1, 2), (0, 0, 0))
         im.putpixel((0, 1), (255, 0, 0))
--- a/Tests/test_image.py
+++ b/Tests/test_image.py
@@ -676,19 +676,18 @@ class TestImage:
 
     @pytest.mark.parametrize("palette_mode", ("RGB", "RGBA"))
     def test_remap_palette_source_palette(self, palette_mode: str) -> None:
-        # When source_palette is passed explicitly its mode is inferred from its
-        # length, so a full 256-entry RGB palette (exactly 768 bytes) must stay
-        # RGB while a 256-entry RGBA one (1024 bytes) is detected as RGBA.
-        source_palette = bytes(
-            (entry + channel * 17) % 256
-            for entry in range(256)
-            for channel in range(len(palette_mode))
-        )
+        # When source_palette is provided, mode is inferred from length.
+        # 768 entries should be detected as a 256-entry RGB palette
+        # 1024 entries should be detected as a 256-entry RGBA palette
         im = Image.new("P", (256, 1))
+        source_palette = bytes(
+            entry for entry in range(256) for channel in range(len(palette_mode))
+        )
         im_remapped = im.remap_palette(list(range(256)), source_palette)
 
         palette = im_remapped.palette
         assert palette is not None
+        assert len(palette.colors) == 256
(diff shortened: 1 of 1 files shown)


radarhere:
> The issue asked for tests

I think the issue wanted us to write tests to check the poster's understanding. If their understanding was incorrect, which we've established it was, I don't think they necessarily felt a need to update our test suite.

Do you yourself feel this is a valuable addition?

> so this branch had no coverage at all.

There is currently no direct test for `source_palette`. The code is being tested indirectly - other tests call Pillow functionality that use `source_palette`.

If you go to https://app.codecov.io/gh/python-pillow/Pillow/blob/main/src%2FPIL%2FImage.py#L2204, you will see that codecov identifies all code within this method as already covered.

Tests/test_image.py
```diff
@@ -674,6 +674,29 @@ def test_remap_palette(self) -> None:
...
 
+    @pytest.mark.parametrize(
+        "bands, palette_mode", ((3, "RGB"), (4, "RGBA")), ids=("RGB", "RGBA")
+    )
+    def test_remap_palette_source_palette(self, bands: int, palette_mode: str) -> None:
+        # When source_palette is passed explicitly its mode is inferred from its
+        # length, so a full 256-entry RGB palette (exactly 768 bytes) must stay
+        # RGB while a 256-entry RGBA one (1024 bytes) is detected as RGBA.
+        source_palette = bytes(
+            (entry + channel * 17) % 256
```
radarhere: Could you explain this line? It seems like any palette would have sufficed.

pranjalm37:
You're right on both counts, and I should have checked branch coverage instead of assuming from the absence of a direct test.

I ran the existing suite (without this PR) under `coverage run --branch` and inspected the annotated output for the `elif len(source_palette) > 768:` block directly — both `bands = 4` and `palette_mode = "RGBA"` are marked executed. I traced it further: several existing GIF tests (`test_dispose2_diff`, `test_dispose2_background_frame`, `test_transparency_in_second_frame`, `test_roundtrip_info_duration`, `test_empty_string_comment`, and others) call `im.save(..., save_all=True)` on an RGBA-paletted image, which flows through `GifImagePlugin._save_all` → `remap_palette(optimized_palette_colors, source_palette)` with a 1024-byte `source_palette` — hitting the exact branch this issue is about. So "completely untested" was wrong; withdrawn.

> Do you yourself feel this is a valuable addition?

Honestly: less than I claimed, now that I've seen it's indirectly covered. What it still adds is a *direct*, boundary-pinned test of the public `remap_palette(dest_map, source_palette)` argument itself — today the only coverage of this exact branch runs through GIF-saving internals, so a change to that internal plumbing (not to `remap_palette`) could silently drop coverage of this branch without any test failing. But that's a fairly marginal defense-in-depth argument, not a "this is currently broken" one, and I don't want to push a maintainer's time on something you don't see value in.

Your call — happy to close this if you'd rather not carry it, no hard feelings either way.

Tests/test_image.py
```diff
@@ -674,6 +674,23 @@ def test_remap_palette(self) -> None:
             with pytest.raises(ValueError):
                 im_hopper.remap_palette([])
 
+    @pytest.mark.parametrize("palette_mode", ("RGB", "RGBA"))
+    def test_remap_palette_source_palette(self, palette_mode: str) -> None:
+        # When source_palette is provided, mode is inferred from length.
+        # 768 entries should be detected as a 256-entry RGB palette
+        # 1024 entries should be detected as a 256-entry RGBA palette
+        im = Image.new("P", (256, 1))
```
radarhere:
```suggestion
        im = Image.new("P", (1, 1))
```

PR #9869, opened 2026-08-12
Convert parallel setting to int

radarhere:
LInt has started failing - https://github.com/python-pillow/Pillow/actions/runs/31563474566/job/94010414409#step:6:22
```pytb
setup.py:383: error: Incompatible types in assignment (expression has type
"str | None", variable has type "int | None")  [assignment]
            self.parallel = configuration.get("parallel", [None])[-1]
                            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Found 1 error in 1 file (checked 303 source files)
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -380,7 +380,8 @@ class pil_build_ext(build_ext):
             setattr(self, f"vendor_{x}", self.check_configuration(x, "vendor"))
         if self.check_configuration("debug", "true"):
             self.debug = True
-        self.parallel = configuration.get("parallel", [None])[-1]
+        if "parallel" in configuration:
+            self.parallel = int(configuration["parallel"][-1])
 
     def finalize_options(self) -> None:
         build_ext.finalize_options(self)
--- a/setup.py
+++ b/setup.py
@@ -32,8 +32,8 @@ while sys.argv[-1].startswith("--pillow-configuration="):
     _, key, value = sys.argv.pop().split("=", 2)
     configuration.setdefault(key, []).append(value)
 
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -380,7 +380,8 @@ def initialize_options(self) -> None:
             setattr(self, f"vendor_{x}", self.check_configuration(x, "vendor"))
         if self.check_configuration("debug", "true"):
             self.debug = True
-        self.parallel = configuration.get("parallel", [None])[-1]
+        if "parallel" in configuration:
+            self.parallel = int(configuration["parallel"][-1])
```
hugovk: Likely caused by today's https://pypi.org/project/types-setuptools/84.0.0.20260812/


---

We already have this value at:

https://github.com/python-pillow/Pillow/blob/a9fd14bfe906e10c817cf2d64953abae52eabab3/setup.py#L35

Shall we rename that to something like `parallel_default` and:

```suggestion
        self.parallel = parallel_default
```

Or use the same pattern as there:


```suggestion
        self.parallel = int(configuration.get("parallel", ["0"])[-1])
```

radarhere: It may not make a difference, but my thinking was that [the default value is `None`](https://github.com/pypa/setuptools/blob/7d531288b7c8708d4cc186213da93eaf5d11833d/setuptools/_distutils/command/build_ext.py#L147).

I've pushed a solution that reuses `parallel_default`, but still defaults to `None`.

Issue #9868, opened 2026-08-12
Pillow ignores palette changes of each frame image when saving as APNG (Animated PNG)


denilsonsa:
### What did you do?

I wrote some code to implement [color cycling](https://en.wikipedia.org/wiki/Color_cycling) [animation](https://amiga.lychesis.net/specials/ColorCycling.html), because [the library I'm writing](https://github.com/denilsonsa/jetpack-python-jetpyck) decodes the graphics format from [an old DOS game](https://www.adeptsoftware.com/jetpack/) and converts it to modern formats (such as PNG, WEBP or not-so-modern GIF). Thus, I implemented [my own color cycling code](https://github.com/denilsonsa/jetpack-python-jetpyck/blob/31add53a9a0aa28ca3d6aa2c8546fd2cfd1d0f97/src/jetpyck/gfxdat.py#L482-L529) that changes the palette, and [a simple loop that generates multiple frames of animation](https://github.com/denilsonsa/jetpack-python-jetpyck/blob/31add53a9a0aa28ca3d6aa2c8546fd2cfd1d0f97/src/jetpyck/gfxdat.py#L1050-L1061) by cloning the `Image` object and replacing the palette.

So, in the end I have a list of Image objects with mode `P`, where each image has the exact same pixel data, but (slightly) different palettes.

Then I tried to save this list of images as an animation. [Saving as GIF](https://github.com/denilsonsa/jetpack-python-jetpyck/blob/31add53a9a0aa28ca3d6aa2c8546fd2cfd1d0f97/src/jetpyck/gfxdat.py#L1190-L1208) works fine. [Saving as WebP](https://github.com/denilsonsa/jetpack-python-jetpyck/blob/31add53a9a0aa28ca3d6aa2c8546fd2cfd1d0f97/src/jetpyck/gfxdat.py#L1243-L1259) also works fine. But [saving as animated PNG](https://github.com/denilsonsa/jetpack-python-jetpyck/blob/31add53a9a0aa28ca3d6aa2c8546fd2cfd1d0f97/src/jetpyck/gfxdat.py#L1220-L1241) doesn't work. [APNG animations](https://stackoverflow.com/questions/62046902/does-apng-support-frame-based-palettes-or-only-a-global-palette) have all frames sharing a single palette (if a palette is present).

### What did you expect to happen?

In summary, after I saved as APNG, I expected to see the animation when opening in the browser. This didn't happen.

[...]

PR #9870, opened 2026-08-14
Save as RGB(A) when APNG frames have different palettes

DSeaStar:
Helps #9868

APNG has a single `PLTE`. Saving several `P` frames with different palettes previously wrote only the first frame's palette, so later frames rendered with the wrong colors.

When the collected palettes differ, the animation is now saved as RGB so each frame keeps its colors. Frames that share one palette stay indexed.

## Test plan

- [x] `Tests/test_file_apng.py::test_apng_save_p_frames_with_different_palettes`
- [x] `Tests/test_file_apng.py::test_apng_save_p_frames_with_same_palette`
- [x] Full `Tests/test_file_apng.py` (49 passed)

### DIFF
--- a/src/PIL/PngImagePlugin.py
+++ b/src/PIL/PngImagePlugin.py
@@ -1332,20 +1332,26 @@ def _save(
         )
         modes = set()
         sizes = set()
-        palettes: set[tuple[int, ...]] = set()
+        last_palette = None
+        different_palette = False
         append_images = im.encoderinfo.get("append_images", [])
         for im_seq in itertools.chain([im], append_images):
             for im_frame in ImageSequence.Iterator(im_seq):
                 modes.add(im_frame.mode)
                 sizes.add(im_frame.size)
-                if im_frame.mode == "P":
-                    palettes.add(tuple(im_frame.getpalette() or []))
+                if im_frame.mode == "P" and not different_palette:
+                    frame_palette = im_frame.getpalette() or []
+                    if last_palette is not None:
+                        if frame_palette != last_palette:
+                            different_palette = True
+                    else:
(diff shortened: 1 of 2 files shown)


radarhere:
https://github.com/python-pillow/Pillow/issues/9868#issuecomment-5264575457
> Bug
> 1. Saving multiple P images as an animated PNG only uses the palette from the first frame.
...
>
> Enhancements
> 1. If there are less than 256 colours in total, you would like the palettes combined into one.
> 2. If there are more than 256 colours in total, you would like all frames to use RGB(A). ...
> 3. You would like pixels unchanged in the next frame to be changed to transparent.

This doesn't address the first and third enhancements. I'm still interested in merging this, but it doesn't resolve the issue completely.

radarhere:
I've created https://github.com/DSeaStar/Pillow/pull/1 with some suggestions.

radarhere (approved):

src/PIL/PngImagePlugin.py
```diff
@@ -1332,25 +1332,44 @@ def _save(
...
+                    if im_frame.palette and im_frame.palette.mode == "RGBA":
+                        palette_frame_with_alpha = im_frame
+                    if different_palette is None:
+                        frame_palette = im_frame.getpalette() or []
+                        if palette is not None:
+                            if frame_palette != palette:
+                                different_palette = True
+                        else:
+                            palette = frame_palette
+                elif im_frame.mode in ("RGBA", "RGBA"):
```
denilsonsa: Hmm, shouldn't this condition be simplified to this?

```suggestion
                elif im_frame.mode == "RGBA":
```

PR #9877, opened 2026-08-19
Use bytearray instead of BytesIO in MspDecoder

radarhere:
1. When iterating through rows, the row index is assigned to `x`
https://github.com/python-pillow/Pillow/blob/3220acae3f820e81c619a75dfb9104ddfa529cb6/src/PIL/MspImagePlugin.py#L129
`y` would seem to be the correct choice instead.

2. `MspDecoder` writes to a `io.BytesIO` instance, only to retrieve the value from it once done. Using `bytearray()` would be simpler.

### DIFF
--- a/src/PIL/MspImagePlugin.py
+++ b/src/PIL/MspImagePlugin.py
@@ -126,14 +126,14 @@ class MspDecoder(ImageFile.PyDecoder):
             msg = "Truncated MSP file in row map"
             raise OSError(msg) from e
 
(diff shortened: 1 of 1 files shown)


akx (approved):
LGTM. `bytearray` and `BytesIO` seem to use the exact same mechanism for figuring out how much to grow the backing array with, so that's not making a difference here.

Not that I imagine Windows 1 and 2 Paint images to generally be very large 😄

PR #9878, opened 2026-08-19
Remove unused ImagingConvertInPlace

radarhere:
`im.putalpha()` is the only place where C's `setmode()` is called.
https://github.com/python-pillow/Pillow/blob/3220acae3f820e81c619a75dfb9104ddfa529cb6/src/PIL/Image.py#L2050-L2052

This is only trying to convert the image to LA, PA or RGBA.

However, the only operation `setmode()` can actually succeed at is converting RGB or RGBX to RGBA.
https://github.com/python-pillow/Pillow/blob/3220acae3f820e81c619a75dfb9104ddfa529cb6/src/_imaging.c#L2079-L2090
https://github.com/python-pillow/Pillow/blob/3220acae3f820e81c619a75dfb9104ddfa529cb6/src/libImaging/Convert.c#L1739-L1748
`ImagingConvertInPlace()` isn't able to convert anything to LA, PA or RGBA.

So this PR
- removes `ImagingConvertInPlace()`. `setmode()` is the only place where `ImagingConvertInPlace()` is called.
- simplifies `setmode()` to more clearly only convert RGB or RGBX images to RGBA. I've also renamed it to `setalpha()`.


src/_imaging.c
```diff
@@ -2052,51 +2052,21 @@ _reduce(ImagingObject *self, PyObject *args) {
...
-static int
-isRGB(const ModeID mode) {
-    return mode == IMAGING_MODE_RGB || mode == IMAGING_MODE_RGBA ||
-           mode == IMAGING_MODE_RGBX;
-}
-
 static PyObject *
-im_setmode(ImagingObject *self, PyObject *args) {
+im_setalpha(ImagingObject *self, PyObject *args) {
     /* attempt to modify the mode of an image in place */
```
akx, line 2058: This comment is a little stale now?

src/_imaging.c
```diff
@@ -3739,7 +3709,7 @@ static struct PyMethodDef methods[] = {
     {"split", (PyCFunction)_split, METH_NOARGS},
     {"fillband", (PyCFunction)_fillband, METH_VARARGS},
 
-    {"setmode", (PyCFunction)im_setmode, METH_VARARGS},
+    {"setalpha", (PyCFunction)im_setalpha, METH_NOARGS},
```
akx, line 3714: What's the deprecation policy for methods on the core image objects in general (is it documented)?

GitHub's Code Search isn't making it easy to find whether there are external users of `im.im.setmode`.

src/_imaging.c
```diff
@@ -2052,51 +2052,21 @@ _reduce(ImagingObject *self, PyObject *args) {
...
-    return mode == IMAGING_MODE_RGB || mode == IMAGING_MODE_RGBA ||
-           mode == IMAGING_MODE_RGBX;
-}
-
 static PyObject *
-im_setmode(ImagingObject *self, PyObject *args) {
+im_setalpha(ImagingObject *self, PyObject *args) {
     /* attempt to modify the mode of an image in place */
+    Imaging im = self->image;
+    if (im->mode == IMAGING_MODE_RGB || im->mode == IMAGING_MODE_RGBX) {
```
akx: Should this function raise a mode error instead of quietly doing nothing?

akx (commented):
I had been looking at removing `ImagingConvertInPlace` (in a bid to see if the shuffler functions could be implemented in a faster way) but didn't dare go through with it.

Some thoughts within though :)

src/_imaging.c
radarhere: There is no deprecation period.

https://github.com/python-pillow/Pillow/issues/4532#issuecomment-618255798
> Pillow makes a commitment to stable public interfaces, which are defined at the Python layer. The C interfaces are explicitly internal, and no effort is made to keep them stable even between minor releases, and no support or warning is given when they change.

src/_imaging.c
radarhere: Is it? The method is still modifying the mode of an image in place. It's more specific now, is all.

src/_imaging.c
akx: That's an useful comment, thanks! Should maybe codify it into development documentation if it isn't already (for instance, I feel much better about touching the palette code even harder in #9829).

And I'll understand this as "the Python-facing interfaces defined in C extension module(s) in Pillow are also malleable" (since this is exactly that). 👍

src/_imaging.c
akx: I suppose... 

Should this function know how to do P-to-PA and L-to-LA too?

src/_imaging.c
radarhere: I've pushed a commit for this.

src/_imaging.c
radarhere: The thought occurred to me. It is a bit different to RGB, since RGB already has the same `pixelsize` as RGBA.

Do you mind if that is a follow-up PR? I would rather tidy things up first.

src/_imaging.c
radarhere: I've created https://github.com/python-pillow/Pillow/pull/9879

hugovk:
I'll click the rebase button to test automerge (https://github.com/python-pillow/Pillow/pull/9126) on a non-docs PR:

image

PR #9883, opened 2026-08-20
Un-deprecate `getdata()`; have it return an `ImageLinearAccess` object

akx:
Follows up on https://github.com/python-pillow/Pillow/pull/9292 to un-deprecate `getdata()`, since `get_flattened_data()` is not a perfect replacement for linear access to image pixels given its upfront memory and time costs. (See https://github.com/python-pillow/Pillow/pull/9292#issuecomment-5353072680 for rationales.)

Fixes https://github.com/python-pillow/Pillow/pull/9261 (since `getdata` now returns a real, well-typed iterator/access object). Given `x.py` like

```python
# pyright: strict
from PIL import Image
im = Image.new("RGB", (4, 4))
printed = list(im.getdata())
```

pyright fails on `main`:
```
(main) $ uvx --quiet --with pytest --with numpy pyright@1.1.411 x.py
  x.py:5:1 - error: Type of "printed" is partially unknown
    Type of "printed" is "list[Unknown]" (reportUnknownVariableType)
  x.py:5:16 - error: Argument of type "ImagingCore" cannot be assigned to parameter "iterable" of type "Iterable[_T@list]" in function "__init__"
    "ImagingCore" is incompatible with protocol "Iterable[_T@list]"
      "__iter__" is not present (reportArgumentType)
2 errors, 0 warnings, 0 informations
```
but is fine on this branch:
```
(image-getdata-linear-access) $ uvx --quiet --with pytest --with numpy pyright@1.1.411 x.py
0 errors, 0 warnings, 0 informations
```

The optimizations from #9881 still help this too and don't conflict here.


akx:
Didn't think I'd bump into a PyPy bug implementing this, but here we are... https://github.com/pypy/pypy/issues/5526

radarhere:
Could you identify exactly what workaround you added for that, so that one day when the PyPy fix has propagated, we are able to remove it?

Tests/test_image_getdata.py
```diff
@@ -30,11 +42,87 @@ def getdata(mode: str) -> tuple[float | tuple[int, ...] | None, int, int]:
...
+
+def test_is_read_only() -> None:
+    hopper_data = hopper().getdata()
+    with pytest.raises(TypeError):
+        hopper_data[0] = (0, 0, 0)  # type: ignore[index]
+
+
+def test_getdata_does_not_expose_the_image_core() -> None:
+    hopper_data = hopper().getdata()
+    # "Weird" core bits are not exposed:
```
radarhere, line 95: What do you mean by 'weird'?

akx: Weird, as in things that a normal user shouldn't be touching (which was the thing in the first place). I can remove the comment if it's too weird in itself. :D

akx:
> Could you identify exactly what workaround you added for that, so that one day when the PyPy fix has propagated, we are able to remove it?

The workaround is the linear access type also act as a mapping (`.tp_as_mapping = &linear_access_as_mapping`), since CPython's and PyPy's mechanisms prefer that for getitem when available. When the object does not act like a mapping, CPython and PyPy will use the sequence item accessor (`linear_access_item`) instead, but before https://github.com/pypy/pypy/commit/e6ed5cc03dcce84a9291202e50678ea40bcf52e1 is shipped in PyPy, it doesn't do the negative-index adjustment like CPython does.

IOW, there's nothing that necessarily needs to be removed.

docs/releasenotes/13.0.0.rst
```diff
@@ -87,6 +87,17 @@ intended for internal use, so there is no replacement.
 API changes
 ===========
 
+Image getdata()
+^^^^^^^^^^^^^^^
+
+:py:meth:`~PIL.Image.Image.getdata` had been marked deprecated since 12.1.0,
+to be removed in Pillow 14, as it returned a poorly specified internal Pillow data type.
```
radarhere, line 106:
```suggestion
to be removed in Pillow 14, as it returned an internal Pillow data type.
```
I don't think it was poorly specified. Perhaps poorly documented, but by design, because it's... internal.

radarhere:
I'm reluctant to return a C object from our Python API, but you could quite rightly say that `load()` does the same thing with `core.PixelAccess`, and that this is similar functionality.

https://github.com/python-pillow/Pillow/pull/9292#issuecomment-5353980219
> My idea would be to return a Python class, rather than C, that calls `getpixel()` from a magic method. However, I recognise that you are likely to prefer C for performance reasons.

If `tuple(self.im)` is truly terrible for performance, then I'd suggest adjusting `get_flattened_data` to use `tobytes()` instead.

I'm not yet convinced that there is a common need for accessing lazy data that can't be suitably addressed with `getpixel()`. An example given in the other PR was
> `zip(im1.getdata(), im2.getdata())` to compare two images pixel by pixel and early-exit upon the first difference. With `get_flattened_data()`, you'd pay upfront to materialize all pixels twice.

but that sounds like something for us to do internally within `__eq__()`, not something to leave for users to find the most optimal route themselves.

akx:
> I'm reluctant to return a C object from our Python API,

I understand that to a degree, but if/when said C object is tightly specified, I'm not sure I totally see the issue.

> If `tuple(self.im)` is truly terrible for performance, then I'd suggest adjusting `get_flattened_data` to use `tobytes()` instead.

Both `tuple(self.im)` and `get_flattened_data()` using `tobytes()` have the same up-front memory bloat issues. (#9938 and all that.) A performance-enlightened user would not want to materialize all the bytes if they only need some.

> [...] but that sounds like something for us to do internally within `__eq__()`, not something to leave for users to find the most optimal route themselves.

That's one UC for `__eq__` (that currently isn't a thing, mind, though #9938 would help (if not optimize) perf there) but I foresee UCs for otherwise comparing, indexing, linearly slicing, etc. image data without fully materializing into a tuple?

PR #9886, opened 2026-08-21
Add `mixed-line-ending` to prek and convert to LF

hugovk:
Follow on from https://github.com/python-pillow/Pillow/pull/9871.

Add `mixed-line-ending` and convert the other line endings to be LF.

https://prek.j178.dev/builtin/#mixed-line-ending


akx (approved):
Looks clean to me :)

PR #9897, opened 2026-08-24
Use borrowed data when exporting an image created by Image.fromarrow

### LINKED ISSUE #9896: Exporting an image created by Image.fromarrow with __arrow_c_array__ crashes with a segmentation fault
### What did you do?

Created an image with `Image.fromarrow` and passed it to `pyarrow.array`.

```python
from PIL import Image
import pyarrow as pa

im = Image.fromarrow(pa.array([0] * 100, type=pa.uint8()), "RGBA", (5, 5))
im.getpixel((0, 0))  # fine
pa.array(im)         # segmentation fault
```

The same happens with a single band image (`"L"` from 25 values), with `im.__arrow_c_array__()` called directly, and on Pillow 11.3.0, 12.3.0 and current `main` (098de8a6e). After any change to the image (`im.putpixel((0, 0), (1, 2, 3, 4))`, which copies the storage) the export works.

An Arrow array with the same values, as for any other image, or a `ValueError` if a borrowed image cannot be exported. The docstring of `fromarrow` shows both directions (`Image.fromarrow(...)` and `pa.array(im)`), so a round trip is the documented usage.

### What actually happened?

The process dies with SIGSEGV (exit code 139), nothing is printed.

### Why

`ImagingBorrowArrow` (`src/libImaging/Storage.c`) fills only the row pointers `im->image[y]` of the borrowed image and leaves `im->block` and `im->blocks` NULL.

`export_single_channel_array` and `export_fixed_pixel_array` (`src/libImaging/Arrow.c`) take the data pointer from `im->block`, or from `im->blocks[0].ptr` when `block` is NULL:

```c
    if (im->block) {
        array->buffers[1] = im->block;
    } else {
        array->buffers[1] = im->blocks[0].ptr;
    }
```

For a borrowed image `blocks` is NULL, so this dereferences NULL. The `im->blocks_count > 1` check above it does not catch this case, `blocks_count` is 0.

`Tests/test_pyarrow.py` compares `fromarrow` images with `getpixel`, so the round trip is not covered.

### Possible fix

A borrowed array is accepted only when its length matches the image exactly, so its rows are contiguous and the export can use the first row. Adding a branch at both places:

```c
    if (im->block) {
        array->buffers[1] = im->block;
    } else if (im->arrow_array_capsule) {
        array->buffers[1] = im->image[0];
    } else {
        array->buffers[1] = im->blocks[0].ptr;
    }
```

With this change (built from `main`) the example returns `fixed_size_list<RGBA: uint8 not null>[4]` with the original values, and the `"L"` case returns the 25 values.

### Context

pillow-heif is moving its plugin to `Image.fromarrow`, so every HEIF image opened with `Image.open` would be a borrowed image, and `pa.array(im)` on such an image would crash. We are holding that release until this is fixed in Pillow.

* OS: macOS 26.5 arm64
* Python: 3.14.5
* Pillow: 11.3.0, 12.3.0, main (098de8a6e)
* pyarrow: 25.0.1

lazerg:
Resolves #9896

An image created by `Image.fromarrow()` borrows its data, so `block` and `blocks` are both NULL and only the row pointers in `image` are set. `export_single_channel_array()` and `export_fixed_pixel_array()` fall back to `blocks[0].ptr` when `block` is NULL, which dereferenced NULL and crashed on `pa.array(im)`.

A borrowed array is only accepted when its length matches the image, so the rows are contiguous and the export can start from the first one.

### DIFF
--- a/Tests/test_pyarrow.py
+++ b/Tests/test_pyarrow.py
@@ -84,21 +84,21 @@ fl_uint8_4_type = pyarrow.field(
 ).type
 
 
-@pytest.mark.parametrize(
-    "mode, dtype, mask",
-    (
-        ("L", pyarrow.uint8(), None),
-        ("I", pyarrow.int32(), None),
-        ("F", pyarrow.float32(), None),
-        ("LA", fl_uint8_4_type, [0, 3]),
-        ("RGB", fl_uint8_4_type, [0, 1, 2]),
-        ("RGBA", fl_uint8_4_type, None),
-        ("RGBX", fl_uint8_4_type, None),
-        ("CMYK", fl_uint8_4_type, None),
-        ("YCbCr", fl_uint8_4_type, [0, 1, 2]),
-        ("HSV", fl_uint8_4_type, [0, 1, 2]),
-    ),
+mode_dtype_mask = (
+    ("L", pyarrow.uint8(), None),
+    ("I", pyarrow.int32(), None),
+    ("F", pyarrow.float32(), None),
+    ("LA", fl_uint8_4_type, [0, 3]),
+    ("RGB", fl_uint8_4_type, [0, 1, 2]),
+    ("RGBA", fl_uint8_4_type, None),
+    ("RGBX", fl_uint8_4_type, None),
+    ("CMYK", fl_uint8_4_type, None),
+    ("YCbCr", fl_uint8_4_type, [0, 1, 2]),
+    ("HSV", fl_uint8_4_type, [0, 1, 2]),
 )
+
+
+@pytest.mark.parametrize("mode, dtype, mask", mode_dtype_mask)
 def test_to_array(mode: str, dtype: pyarrow.DataType, mask: list[int] | None) -> None:
     img = hopper(mode)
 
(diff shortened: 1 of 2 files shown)


radarhere:
I've created https://github.com/lazerg/Pillow/pull/1 with a suggestion.

lazerg:
Good catch, thanks. RGBX/CMYK/YCbCr/HSV route through the same `export_fixed_pixel_array` path as RGBA (branch is picked purely by `im->bands == 1`, not by mode), so the fix covers them too. Applied your refactor and ran all 10 modes locally: all pass, including the 4 new ones. Pushed as bd55c4120.

radarhere:
I've created https://github.com/lazerg/Pillow/pull/2 with a suggestion.

lazerg:
Thanks, merged this in (46e5115).

wiredfool (approved):

radarhere:
Could you explain to me the purpose of `test_fromarray_interleaved_to_array()`? It seems that `test_fromarray_to_array()` already prevents regression.

lazerg:
Correct.. I removed it in a69acca2.

I checked before deleting: with the fix reverted, `test_fromarray_to_array` still segfaults on its own. The interleaved import shape stays covered by the `UINT` rows of `test_fromarray`.

PR #9899, opened 2026-08-25
Add _normalize_coords() to reduce duplicate code

radarhere:
Extracting a cleanup idea from #9490

### DIFF
--- a/src/PIL/ImageDraw.py
+++ b/src/PIL/ImageDraw.py
@@ -222,6 +222,16 @@ class ImageDraw:
         ellipse_xy = (xy[0] - radius, xy[1] - radius, xy[0] + radius, xy[1] + radius)
         self.ellipse(ellipse_xy, fill, outline, width)
 
+    def _normalize_coords(self, xy: Coords) -> Sequence[Sequence[float]]:
+        """Normalize 1 or 2 dimensional coord sequence into 2d sequence."""
+        if isinstance(xy[0], (list, tuple)):
+            return cast("Sequence[Sequence[float]]", xy)
+        else:
+            return [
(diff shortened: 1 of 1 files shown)


src/PIL/ImageDraw.py
```diff
@@ -222,6 +222,16 @@ def circle(
         ellipse_xy = (xy[0] - radius, xy[1] - radius, xy[0] + radius, xy[1] + radius)
         self.ellipse(ellipse_xy, fill, outline, width)
 
+    def _normalize_coords(self, xy: Coords) -> Sequence[Sequence[float]]:
```
akx, line 225: This should maybe be a free function in the module?

radarhere: Why? It's not used anywhere else.

akx: Since it doesn't need to be overridden by any subclass of `ImageDraw` I can conceive of, and free functions are a tiny bit faster to call:

In the `dis`assembly for a quick timing example, 

* a free function call is `LOAD_GLOBAL` (looks at the module globals table) + `LOAD_FAST_BORROW` for the argument + `CALL`
* an instance function call is `LOAD_FAST_BORROW` for the local `self` + `LOAD_ATTR` (a dict lookup on the instance) + `LOAD_FAST_BORROW` for the argument + `CALL`

So in other words, because it's unnecessary to be an instance method.

radarhere: My personal preference would be for code that is only called from internal methods in the instance to be a method, as an indicator of how it is used. It also places it much closer to the first `self._normalize_coords(xy)`, making it quicker to understand.

PR #9901, opened 2026-08-25
Remove undocumented `GifImagePlugin._save_netpbm`

akx:
This PR proposes to remove the undocumented (aside from the changelog for PIL 0.2a1 noting "netpbm save hack") `GifImagePlugin._save_netpbm` function.


docs/releasenotes/13.0.0.rst
```diff
@@ -66,6 +66,11 @@ ImageCms.ImageCmsProfile.product_name and .product_info
 ``.product_info`` attributes have been removed. They were set to ``None`` since Pillow
 2.3.0.
 
+GifImagePlugin._save_netpbm()
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+``GifImagePlugin._save_netpbm()`` has been removed.
```
radarhere, line 72: If this function is unimportant enough that we don't need a deprecation period, then I'm also inclined to think it is unimportant enough that we don't need release notes.

akx: I've been bitten by libraries quietly removing functions (last in https://github.com/oracle/oci-python-sdk/issues/878).

Augmented the changelog entry to include some history (since I had it at hand), and an alternate recipe. What do you think?

radarhere: If you're concerned about removing it without notifying users, then it sounds like a deprecation period is in order. https://github.com/python-pillow/Pillow/pull/9186 deprecated the undocumented `Image._show()` function.

A deprecation period lasts at least a year, so for this PR, it would target Pillow 14.0.0.

akx: I'm not overly concerned about a proper deprecation period for an underscore-private feature that was never really documented. 

So to avoid a deprecation period, I should remove the changelog entry?

radarhere: Looking further,
https://github.com/python-pillow/Pillow/blob/eb3935ffc710e543870ea16c8d4104d5a5962510/src/PIL/GifImagePlugin.py#L1220-L1223
seems similar to our [rationale for adding a deprecation period](https://github.com/python-pillow/Pillow/pull/4693#issuecomment-647110867) to `Image._showxv`

akx: GitHub Code Search [seems to imply](https://github.com/search?q=%22_save_netpbm%22+language%3APython+-path%3AGifImagePlugin+-path%3Atest_file_gif+-path%3Atest_shell_injection&type=code) there is no code using `_save_netpbm` that is not in one of Pillow's own files).

For `_showxv`, [a similar search](https://github.com/search?q=%22Image._showxv%22+language%3Apython+-path%3Atest_image&type=code) today shows only one actual call in [code uploaded 2 years ago](https://github.com/SteveytW/spend-the-debt/blame/62696116beb52bbe214087d0a1ccfc7d8a592afd/data/scrapeamazon.py#L67) while `_showxv` was removed in 499040491bfcc8804b2cb896f8c20802813939f2, 5 years ago.

akx:
> That's a reasonable point - although #9901 is mentioned in [comments](https://github.com/python-pillow/Pillow/blob/cacb5139bf9abe70afe2cd533b319d5c66a48867/src/PIL/GifImagePlugin.py#L1220-L1223), so I think the [`_showxv` logic](https://github.com/python-pillow/Pillow/pull/4693#issuecomment-647110867) applies.

_Originally posted by @radarhere in https://github.com/python-pillow/Pillow/issues/9916#issuecomment-5462327274_

I'm not sure it's the same logic. The comment here is saying you'd need to uncomment the follwoing line _in library code_ to use the alternative. I don't think editing library code is much of a thing in 2026?

Might as well document my archeology: By the looks of it, `Image.show()` itself used to default to using Tk to show the image in Imaging-1.1.1, and if that failed, it would fall back to trying xv, before failing. `_show()` was added in Imaging-1.1.7 as a trampoline  for `_showxv()`, which itself was a trampoline to `ImageShow.Show()`, which already at that point had a "plug-in" mechanism for trying different viewers -- so tbh, the `# override me, as necessary` comment and `Simple display support.  User code may override this.` both were superfluous at that point already, since all you needed to do there was `ImageShow.Show.register(your_function, -1)`.

hugovk (approved):
I don't think we need release notes for this.

Issue #9900, opened 2026-08-25
ImageFilter.BoxBlur(float("nan")) causes a heap-buffer-overflow


marinelay:
I've been fuzzing Python C extension modules for a small research project, and this one came up.
It reproduces with the binary wheel from a plain `pip install Pillow`.

### Versions

Pillow 12.3.0 binary wheel and Pillow `main` commit (`807d689a83738027b6f6e0f219a6a6dd30e01c08`), CPython 3.12.3, Ubuntu 24.04 x86_64, glibc 2.39.

### Reproducer

```python
from PIL import Image, ImageFilter

Image.new("RGB", (1, 1)).filter(ImageFilter.BoxBlur(float("nan")))
```

Running it produces:

```console
$ python reproducer.py
Segmentation fault (core dumped)
```

An ASan build reports invalid NaN-to-integer conversion followed by a heap-buffer-overflow write.

```text
heap-buffer-overflow WRITE of size 1
    #0 ImagingLineBoxBlur32 src/libImaging/BoxBlur.c:73
    #1 ImagingHorizontalBoxBlur src/libImaging/BoxBlur.c:212
    #2 ImagingBoxBlur src/libImaging/BoxBlur.c:273
```

I would expect a non-finite radius to be rejected rather than converted into overflowing native edge arithmetic.

PR #9902, opened 2026-08-25
Reject NaN and out-of-range blur radii

lazerg:
Addresses #9900

 * `ImagingBoxBlur()` only rejected negative radii, so NaN slipped through and `(int)floatRadius` was undefined behaviour, giving a garbage radius and the out of bounds write in `ImagingLineBoxBlur32()`. Infinity and any radius of `2**31` or more crash the same way, since they are not representable as an `int` either.
 * The radius is now rejected unless it is at least 0 and small enough to convert to an `int`, which covers `GaussianBlur` too.


src/libImaging/BoxBlur.c
```diff
@@ -244,9 +244,14 @@ ImagingBoxBlur(Imaging imOut, Imaging imIn, float xradius, float yradius, int n)
     if (n < 1) {
         return ImagingError_ValueError("number of passes must be greater than zero");
     }
-    if (xradius < 0 || yradius < 0) {
+    /* Negated comparisons, so that NaN is rejected as well. */
+    if (!(xradius >= 0) || !(yradius >= 0)) {
```
akx, line 245: <del>To a casual reader, this looks like a very odd way to write the thing. If the idea is that this catches `NaN`s by... off the top of my head, some quirk of how IEEE floats work, a comment would be nice.</del>

EDIT: a comment got added just as I was writing this. Nice!

Even nicer, though, I think, some sort of reusable float-validation function since I'm quite sure there are many other instances of this same class of bug in Pillow.

fallenmi (changes requested):
[P1] Avoid signed overflow below the new radius limit

The new `2**31` guard keeps the float-to-`int` conversion in range, but it still accepts radii for which the existing integer arithmetic immediately overflows. `ImagingHorizontalBoxBlur()` computes `(radius * 2 + 1) * ww` with `radius` as an `int`; for `radius = 2**30`, `radius * 2` is already outside the range of signed `int` even though the new check accepts it.

I reproduced this on exact head `ed5a26fb0caa987f95b539494ae6b26fc5f6c551` with a UBSan build using a 3x3 `L` image and `ImageFilter.BoxBlur(2**30)`:

```
src/libImaging/BoxBlur.c:179:38: runtime error: signed integer overflow: 1073741824 * 2 cannot be represented in type 'int'
```

The three new out-of-range tests pass, but they only exercise `NaN`, infinity, and `2**31`; the same UBSan failure remains after a clean merge onto current `main` `e41083f383c9cd3db95de52564cc0b6452313d4a`. Please either reject radii before this multiplication can overflow (for the current expression, no later than `(INT_MAX - 1) / 2`) or perform the arithmetic in a type that cannot overflow, and add a regression around `2**30`.

This review was prepared with OpenAI Codex assistance under the account owner's authorization. The exact diff, source path, merge result, and test evidence above were inspected and verified in the account owner's Codex session before submission.

lazerg:
Confirmed. A UBSan build reproduces `BoxBlur.c:179:38` at radius `2**30`.

The same problem also starts lower down. `acc = lineIn[0] * (radius + 1)` (lines 43-46, 55-58, 126 and 130) overflows once `255 * (radius + 1)` passes `INT_MAX`, so from radius 8421504. A `(INT_MAX - 1) / 2` bound would still leave those in place.

I made the arithmetic unsigned instead of lowering the limit. `acc`, `ww` and `fw` are already `UINT32`, so only the `int` operands needed a cast, and every radius the guard accepts keeps the same result.

Added `test_large_blur_filter_radius` over `2**24` and `2**30` for `L` and `RGB`. It aborts under UBSan on the previous commit and passes now. Full suite still green.

ac5f69b73

fallenmi (approved):
Thanks — this resolves my prior UBSan blocker. At exact head `ac5f69b73c6086d2dc0865389d5afed322c73dc1`, the multiplications that previously promoted through signed `int` now enter unsigned arithmetic before the operation in both the 8-bit and 32-bit paths. The existing binary32 conversion and `2**31` guard keep the remaining cast, `radius + 1`, edge calculations, and indices representable for every admitted radius.

The new `2**24` and `2**30` regressions cover both `L` and `RGB`. I rechecked current `main` `c9e4cff871b19688789a9da4827d5e993204174c` and synthetic merge `35501a1a5cd91d64051f36bb28c49ec84b83552d`; the merge preserves the fix and exact test blob. All 55 check runs, 11 suites, nine Actions workflows, and two legacy statuses are green.

Disclosure: I used OpenAI Codex to assist this changed-head source/API review. I verified the exact public diff, source paths, refs, interactions, policy, merge, and CI metadata; I did not rerun a local build for this follow-up.

radarhere:
Hi. Yes, I merged https://github.com/python-pillow/Pillow/pull/9906. It seemed helpful to continue raising errors in invalid input at `ImageFilter.BoxBlur(radius)` in Python - but that doesn't negate this.

lazerg:
Hi @radarhere, just checking on the status of this one since the linked issue is now closed: is there anything else needed here, or should I close this out?

radarhere:
I've created https://github.com/lazerg/Pillow/pull/5

lazerg:
Thanks, merged as 8fd1147e7.

Python and C now raise the same two messages, and `nan` still fails the first check because `nan >= 0` is False. `GaussianBlur` has no check in Python, so its test still reaches the C guard. Full suite green.

src/libImaging/BoxBlur.c
```diff
@@ -37,10 +37,10 @@ void static inline ImagingLineBoxBlur32(
     /* Compute acc for -1 pixel (outside of image):
        From "-radius-1" to "-1" get first pixel,
        then from "0" to "radius-1". */
-    acc[0] = lineIn[0][0] * (radius + 1);
-    acc[1] = lineIn[0][1] * (radius + 1);
-    acc[2] = lineIn[0][2] * (radius + 1);
-    acc[3] = lineIn[0][3] * (radius + 1);
+    acc[0] = lineIn[0][0] * (UINT32)(radius + 1);
```
radarhere, line 40: Why `(UINT32)(radius + 1)` rather than `((UINT32)radius + 1)`?

Tests/test_image_filter.py
```diff
@@ -217,40 +217,34 @@ def test_consistency_i16_high_byte(mode: str) -> None:
...
+    im = Image.new("L", (3, 3), 128)
+    assert im.filter(ImageFilter.BoxBlur(radius)).getpixel((1, 1)) == 128
 
-    im = hopper()
-    box_blur_filter = ImageFilter.BoxBlur(2)
-    box_blur_filter.radius = radius
-    with pytest.raises(ValueError):
-        im.filter(box_blur_filter)
+    im = Image.new("RGB", (3, 3), (128, 128, 128))
+    assert im.filter(ImageFilter.BoxBlur(radius)).getpixel((1, 1)) == (128, 128, 128)
```
radarhere, line 240: While this test may reproduce the problem under a build checking for  UBSan, Pillow's CI jobs don't check for that. We use [oss-fuzz](https://github.com/google/oss-fuzz) to check for problems when parsing images or fonts, but the entire test suite isn't run under that.

lazerg:
Closing this. PR #9906 already fixed the same NaN and out-of-range radius crash and it's merged.

radarhere:
@lazerg do you mind if I cherry-pick part of this into a separate PR and merge that?

lazerg:
Not a problem.

PR #9904, opened 2026-08-25
Avoid integer overflow when calculating reduced image size

### LINKED ISSUE #9903: `Image.reduce()` segfaults when passed `2**31 - 1` as the reduction factor
### Summary

Calling `Image.reduce()` with the factor `2**31 - 1` on a `4x4` image terminates the Python process with `SIGSEGV`. It reproduces with the binary wheel from a plain `pip install Pillow`.

### Versions

Pillow 12.3.0, CPython 3.12.3, Ubuntu 24.04 x86_64, glibc 2.39.

### Reproducer

```python
from PIL import Image

Image.new("L", (4, 4)).reduce(2**31 - 1)
```

Running it produces:

```console
$ python reproducer.py
Segmentation fault (core dumped)
```

An instrumented build reports signed integer overflow in [the output-dimension calculation](https://github.com/python-pillow/Pillow/blob/12.3.0/src/libImaging/Reduce.c#L1463-L1465), followed by a NULL store in [`ImagingReduceCorners`](https://github.com/python-pillow/Pillow/blob/12.3.0/src/libImaging/Reduce.c#L1179-L1187).

I found this while fuzzing Python C extension modules.
It looks like a bug to me, but I may be missing an intended constraint, so I would appreciate confirmation.

lazerg:
`ImagingReduce()` sizes the output image with `(box[2] + xscale - 1) / xscale`. When the box width plus the scale exceeds `INT_MAX` that addition overflows and the result is a zero-width, zero-height image. `ImagingReduceCorners()` then writes to `imOut->image8[0][0]`, which has no rows allocated, so `Image.new("L", (4, 4)).reduce(2**31 - 1)` segfaults.

Since the box is never empty, the same round-up can be written as `(box[2] - 1) / xscale + 1`, which cannot overflow and gives a 1x1 image as expected.

Fixes #9903

### DIFF
--- a/Tests/test_image_reduce.py
+++ b/Tests/test_image_reduce.py
@@ -55,18 +55,18 @@ def test_args_factor(size: int | tuple[int, int], expected: tuple[int, int]) -> 
 
 @pytest.mark.parametrize("mode", ("L", "I", "F"))
 @pytest.mark.parametrize(
-    "size, expected",
+    "factor, expected",
     (
         (2**31 - 1, (1, 1)),
         ((2**31 - 1, 1), (1, 10)),
         ((1, 2**31 - 1), (10, 1)),
     ),
 )
-def test_args_factor_large(
-    size: int | tuple[int, int], expected: tuple[int, int], mode: str
+def test_large_factor(
+    factor: int | tuple[int, int], expected: tuple[int, int], mode: str
 ) -> None:
     im = Image.new(mode, (10, 10))
-    assert im.reduce(size).size == expected
+    assert im.reduce(factor).size == expected
 
 
 @pytest.mark.parametrize(
--- a/Tests/test_image_reduce.py
+++ b/Tests/test_image_reduce.py
@@ -54,6 +54,21 @@ def test_args_factor(size: int | tuple[int, int], expected: tuple[int, int]) -> 
 
 
 @pytest.mark.parametrize(
+    "size, expected",
+    (
+        (2**31 - 1, (1, 1)),
+        ((2**31 - 1, 1), (1, 10)),
+        ((1, 2**31 - 1), (10, 1)),
+    ),
+)
+def test_args_factor_large(
+    size: int | tuple[int, int], expected: tuple[int, int]
+) -> None:
+    im = Image.new("L", (10, 10))
+    assert expected == im.reduce(size).size
+
+
+@pytest.mark.parametrize(
     "size, expected_error", ((0, ValueError), (2.0, TypeError), ((0, 10), ValueError))
 )
 def test_args_factor_error(
--- a/Tests/test_image_reduce.py
+++ b/Tests/test_image_reduce.py
@@ -53,6 +53,7 @@ def test_args_factor(size: int | tuple[int, int], expected: tuple[int, int]) -> 
     assert expected == im.reduce(size).size
 
 
+@pytest.mark.parametrize("mode", ("L", "I", "F"))
 @pytest.mark.parametrize(
     "size, expected",
     (
@@ -62,10 +63,10 @@ def test_args_factor(size: int | tuple[int, int], expected: tuple[int, int]) -> 
     ),
 )
 def test_args_factor_large(
-    size: int | tuple[int, int], expected: tuple[int, int]
+    size: int | tuple[int, int], expected: tuple[int, int], mode: str
 ) -> None:
-    im = Image.new("L", (10, 10))
-    assert expected == im.reduce(size).size
+    im = Image.new(mode, (10, 10))
+    assert im.reduce(size).size == expected
 
 
 @pytest.mark.parametrize(
--- a/src/libImaging/Reduce.c
+++ b/src/libImaging/Reduce.c
@@ -1461,7 +1461,7 @@ ImagingReduce(Imaging imIn, int xscale, int yscale, int box[4]) {
     }
 
     imOut = ImagingNewDirty(
-        imIn->mode, (box[2] + xscale - 1) / xscale, (box[3] + yscale - 1) / yscale
+        imIn->mode, (box[2] - 1) / xscale + 1, (box[3] - 1) / yscale + 1
     );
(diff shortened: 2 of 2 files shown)


Tests/test_image_reduce.py
```diff
@@ -53,6 +53,21 @@ def test_args_factor(size: int | tuple[int, int], expected: tuple[int, int]) ->
...
+        (2**31 - 1, (1, 1)),
+        ((2**31 - 1, 1), (1, 10)),
+        ((1, 2**31 - 1), (10, 1)),
+    ),
+)
+def test_args_factor_large(
+    size: int | tuple[int, int], expected: tuple[int, int]
+) -> None:
+    im = Image.new("L", (10, 10))
+    assert expected == im.reduce(size).size
```
radarhere:
```suggestion
    assert im.reduce(size).size == expected
```
Nitpick suggestion

src/libImaging/Reduce.c
```diff
@@ -1461,7 +1461,7 @@ ImagingReduce(Imaging imIn, int xscale, int yscale, int box[4]) {
     }
 
     imOut = ImagingNewDirty(
-        imIn->mode, (box[2] + xscale - 1) / xscale, (box[3] + yscale - 1) / yscale
+        imIn->mode, (box[2] - 1) / xscale + 1, (box[3] - 1) / yscale + 1
```
akx, line 1467: I'd maybe add a comment here about the operation order, and that it guarantees the image to have at least size 1x1.

Tests/test_image_reduce.py
lazerg: Applied in 4d40d5d53.

src/libImaging/Reduce.c
lazerg: Added in bf60f3c86.

While checking this I noticed that `xscale * yscale` in `ImagingReduceNxN` and `(box[2] % xscale) * yscale` in `ImagingReduceCorners` still overflow with a factor this large. Those values only feed loops that run zero iterations, so the output pixels are correct and no bad memory access happens, but an instrumented build will still report the overflow on the same input. Clamping both scales to the box size after the size calculation would remove it.

I am happy to add that here, or to keep this PR to the crash and raise it separately.

akx: What do you mean with "Those values only feed loops that run zero iterations"? Those functions probably shouldn't end up being called...

In fact, it could be a good idea to add separate paths when the reduction ends up being one-dimensional in one dimension or the other, or both, since isn't the end result then just an average over rows, columns or both (separable)?

lazerg: They are still called, not skipped. What's zero is the inner loop count: `box[2] / xscale` and `box[3] / yscale`. Box can never exceed INT_MAX, since it's just an image dimension. Here xscale is 2**31-1. So both divisions come out to 0, and the pixel-reading loops never run.

The overflow happens earlier, before the loop starts. It's in `yscale * xscale` (ImagingReduceNxN) and `(box[2] % xscale) * yscale` (ImagingReduceCorners). Those feed `multiplier` and `amend`, which only get used inside the loop that never executes. So the overflowed value never reaches memory or an output pixel. It's still UB at that line though. An UBSan build will still report it on this input, even though the real output is correct.

The separable idea for the one-dimensional case is worth doing. But it's a bigger change than this crash fix. I'd rather open it as its own issue instead of growing this PR.

Tests/test_image_reduce.py
```diff
@@ -53,6 +53,22 @@ def test_args_factor(size: int | tuple[int, int], expected: tuple[int, int]) ->
     assert expected == im.reduce(size).size
 
 
+@pytest.mark.parametrize("mode", ("L", "I", "F"))
```
radarhere, line 56: I don't see why different modes would affect this at all?

lazerg: Fair point, the sizes don't change by mode. I added L/I/F so the test also exercises the 32bpc path (`ImagingReduceNxN_32bpc` / `ImagingReduceCorners_32bpc`), a separate C function from the 8-bit one that hit the same crash. Without I/F only the 8-bit path gets covered.

radarhere: Those other paths come after the fix. So you're testing to make sure that there aren't other overflow bugs lurking in `reduce()`?

lazerg: Not lurking bugs in general, more specific: the crash itself lives in two separate C functions depending on bit depth (8bpc writes image8, 32bpc writes image32), and the size-calc fix is shared by both. So L only proves the 8bpc writer is safe after the fix, I/F prove the 32bpc one is too.

Tests/test_image_reduce.py
```diff
@@ -53,6 +53,22 @@ def test_args_factor(size: int | tuple[int, int], expected: tuple[int, int]) ->
...
+        (2**31 - 1, (1, 1)),
+        ((2**31 - 1, 1), (1, 10)),
+        ((1, 2**31 - 1), (10, 1)),
+    ),
+)
+def test_args_factor_large(
+    size: int | tuple[int, int], expected: tuple[int, int], mode: str
+) -> None:
+    im = Image.new(mode, (10, 10))
+    assert im.reduce(size).size == expected
```
radarhere:
```suggestion
@pytest.mark.parametrize(
    "factor, expected",
    (
        (2**31 - 1, (1, 1)),
        ((2**31 - 1, 1), (1, 10)),
        ((1, 2**31 - 1), (10, 1)),
    ),
)
def test_large_factor(
    factor: int | tuple[int, int], expected: tuple[int, int], mode: str
) -> None:
    im = Image.new(mode, (10, 10))
    assert im.reduce(factor).size == expected
```
The first argument to `reduce()` is called `factor` in the method - https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.reduce

radarhere (approved):

PR #9906, opened 2026-08-25
Reject non-finite radius in BoxBlur

### LINKED ISSUE #9900: ImageFilter.BoxBlur(float("nan")) causes a heap-buffer-overflow

JonnasFigueiredo:
Fixes #9900. Alternative to #9902

`BoxBlur` validates `radius >= 0` in Python, but `nan < 0` (and `inf < 0`) is `False`, so a non-finite radius slipped through the check and reached the C `box_blur`, causing a heap-buffer-overflow (found by fuzzing).

This extends the existing radius validation to reject non-finite values, raising `ValueError` just like a negative radius does:

```pycon
>>> from PIL import ImageFilter
>>> ImageFilter.BoxBlur(float("nan"))
ValueError: radius must be a finite number >= 0
```

Added tests for `nan`/`inf`, both as a scalar and per-axis. I verified against the compiled extension that valid radii still work and that `nan`/`inf`/negative radii are rejected.


radarhere:
I've created https://github.com/JonnasFigueiredo/Pillow/pull/1 with a suggestion.

JonnasFigueiredo:
Thanks @radarhere, that is a nice cleanup. I have merged it in. Using math.nan and math.inf directly, and covering both tuple positions, reads much better than what I had.

Tests/test_image_filter.py
```diff
@@ -235,6 +237,22 @@ def test_invalid_box_blur_filter(radius: int | tuple[int, int]) -> None:
...
+        math.nan,
+        (math.nan, 1),
+        (1, math.nan),
+        math.inf,
+        (1, math.inf),
+        (math.inf, 1),
+    ),
+)
+def test_box_blur_non_finite_radius(radius: float | tuple[float, float]) -> None:
+    with pytest.raises(ValueError):
```
```suggestion
    with pytest.raises(ValueError, match="radius must be a finite number >= 0"):
```

Issue #9898, opened 2026-08-24
Kerning woes


rarogcoh:
I am trying to draw some text with Pillow, but the result has some jarring kerning issues. See the image:

<img width="220" height="140" alt="Image" src="https://github.com/user-attachments/assets/2ebde997-89b8-476a-a5fb-6c8ffa69a1a9" />

At bigger sizes the distance between the letter P and the letter o in the word "Power" is as big as half a space, give or take. At lower sizes, the distance between the letter o and the letters lt in the word bolt is also pretty huge (particularly in the line that is 15px or 16px in size).

This is with Noto Sans, but I also tried Droid Sans, Adwaita Sans and Open Sans and they all had this problem, usually with the same letter clusters, although the problem was less prominent in Open Sans.

I do not have this problem with apps on my system that use these fonts. I attached image showing how Kate, Geany and Libre Office Writer display that string and the kerning there looks ok:

<img width="271" height="126" alt="Image" src="https://github.com/user-attachments/assets/d7cc20bc-d7d6-4347-9d02-50d77ec4b16f" />

The sizes in the apps are standard 12pts, so are the equivalent of the 16px string in Pillow.

I am on Linux (kernel version: 7.1.9-354.current). This is what `python3 -m PIL.report` says:

```text
--------------------------------------------------------------------
Pillow 12.3.0
Python 3.14.0 (main, Oct 25 2025, 05:06:07) [GCC 15.1.1 20250508]
--------------------------------------------------------------------
...
--------------------------------------------------------------------
--- PIL CORE support ok, compiled for 12.3.0
--- TKINTER support ok, loaded 8.6
--- FREETYPE2 support ok, loaded 2.14.3
--- LITTLECMS2 support ok, loaded 2.19
--- WEBP support ok, loaded 1.6.0
--- AVIF support ok, loaded 1.4.2
--- JPEG support ok, compiled for libjpeg-turbo 3.1.4.1
--- OPENJPEG (JPEG2000) support ok, loaded 2.5.4
--- ZLIB (PNG/ZIP) support ok, loaded 1.3.1.zlib-ng, compiled for zlib-ng 2.3.3
--- LIBTIFF support ok, loaded 4.7.1
--- RAQM (Bidirectional Text) support ok, loaded 0.10.5, fribidi 1.0.15, harfbuzz 14.2.1
*** LIBIMAGEQUANT (Quantization method) support not installed
--- XCB (X protocol) support ok
--------------------------------------------------------------------
```

[...]

PR #9879, opened 2026-08-20
C interface is internal

radarhere:
https://github.com/python-pillow/Pillow/pull/9878#discussion_r3819063611 requested that we document that Pillow is a Python library, and we do not encourage direct use of our C code.

Based on https://github.com/python-pillow/Pillow/issues/4532#issuecomment-618255798

PR #9908, opened 2026-08-26
Document and clarify _imagingft

akx:
Refs #9898 - had to figure out what's happening in `_imagingft` to get to the bottom of the kerning woes, and these came out of that. The actual bugfixes will be in another PR stacked on this.

Basically:

* there was some type confusion with FreeType's 26.6 fixed-point values; some comments noted values were in 26.6 when they are in pixels, or the other way around. This uses FT's `FT_F26Dot6` type (an alias to `long`, since C doesn't have real newtypes, more's the shame) to make it more obvious to the reader.
* adds macros to work with said data type, to avoid unclear `* 64`s or `<< 6`s. Not truly type-safe since, well, see above about newtypes, but better than throwing magic numbers around.
* adds doc comments for the internal functions, because the calling conventions weren't clear
* changes the core `getfont()`'s signature to make all arguments required (#9879 in mind, that should be fine to do), since calling `getfont()` without encoding could handily segfault (since `encoding` in C was left uninitialized, set to stack garbage).
* fixes `_imagingft.pyi`'s types to match the reality of the module.


radarhere:
> adds tests that prove that the typings were wrong

Looks to me like [`test_bytearray_not_supported()`](https://github.com/radarhere/Pillow/commit/2a587293325551a96f7d20fd368d91c8256cf4a7) [passes](https://github.com/radarhere/Pillow/actions/runs/33070774027) without the rest of your changes.

akx:
Er, yes..? The test passes (even on `main`) because `font.getlength`, `font.getbbox`, `font.getmask` don't actually accept `bytearray`s. If you remove the `pytest.raises`, it will fail.

On `main`, trying to pass in a `bytearray` fails:

```
$ uv pip install -e .
$ uv run python
Python 3.14.6 (main, Jun 11 2026, 03:55:33) [Clang 22.1.3 ] on darwin
>>> from PIL import ImageFont
>>> f = ImageFont.load_default()
>>> f.font.getsize(bytearray(b"f"))
  File "<python-input-3>", line 1, in <module>
    f.font.getsize(bytearray(b"f"))
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
TypeError: expected string or bytes
>>>
```
even though the typings say they should be accepted:
```
~/b/Pillow (main) $ grep -C2 getsize src/PIL/_imagingft.pyi
    def getsize(
        self,
        string: str | bytes | bytearray,
```

src/_imagingft.c
```diff
@@ -838,11 +880,30 @@ font_getsize(FontObject *self, PyObject *args) {
     return result;
 }
 
+/**
+ * Rasterize a string into an image buffer.
+ *
+ * Python parameters:
+ * - string: the text to render
+ * - fill: ???
```
radarhere:
```suggestion
 * - fill: Python function to create the core image at a specified width and height
```
The originated in https://github.com/python-pillow/Pillow/pull/7246

src/_imagingft.c
```diff
@@ -838,11 +880,30 @@ font_getsize(FontObject *self, PyObject *args) {
...
+ * - mode_name: imaging mode string
+ * - dir: text direction string
+ * - features: font features sequence
+ * - lang: language string
+ * - stroke_width: stroke width in pixels
+ * - stroke_filled: ???
```
radarhere:
```suggestion
 * - stroke_filled: bool determinong whether to omit the inner border of the stroke, creating a filled glyph
```
See https://github.com/python-pillow/Pillow/pull/8701

radarhere:
Oh, right, most PR tests fail without the other changes, but that was not your intention here.

src/_imagingft.c
```diff
@@ -96,6 +103,14 @@ static PyTypeObject Font_Type;
 
 /* round a 26.6 pixel coordinate to the nearest integer */
 #define PIXEL(x) ((((x) + 32) & -64) >> 6)
+/* round a 26.6 pixel coordinate down to integer */
+#define PIXEL_FLOOR(x) ((x) >> 6)
+/* round a 26.6 pixel coordinate up to integer */
+#define PIXEL_CEIL(x) (((x) + 63) >> 6)
+/* the sub-pixel part of a 26.6 pixel coordinate */
+#define PIXEL_FRAC(x) ((x) & 63)
```
radarhere:
```suggestion
```
These aren't used.

akx: They are used in #9909 (which stacks on top of this). If you like, I can remove them from here and add them in a separate commit there.

radarhere: Yes, that would be preferable, thanks.

akx: Yanked that commit out of the stack here 👍

akx: ... Turns out I'll need to keep `PIXEL_TO_FIXED` in 😅

Tests/test_imagefont.py
```diff
@@ -571,6 +571,24 @@ def test_getbbox_empty(font: ImageFont.FreeTypeFont) -> None:
     assert (0, 0, 0, 0) == font.getbbox("")
 
 
+def test_bytearray_not_supported(font: ImageFont.FreeTypeFont) -> None:
+    # The C text layout only accepts str and bytes, so bytearray is rejected
+    # even though these methods used to be annotated as taking one.
+    for op in (font.getlength, font.getbbox, font.getmask):
+        assert op(b"A") is not None
+        with pytest.raises(TypeError, match="expected string or bytes"):
+            op(bytearray(b"A"))  # type: ignore[arg-type]
```
radarhere: I don't think that past functionality of the code need to be permanently documented in the form of tests. `test_bytes()` already tests for this `TypeError`.

So we no longer support bytearrays. There's a lot of other things we don't support as well.

akx: Removed the unsupport test.

Tests/test_imagefont.py
```diff
@@ -571,6 +571,15 @@ def test_getbbox_empty(font: ImageFont.FreeTypeFont) -> None:
     assert (0, 0, 0, 0) == font.getbbox("")
 
 
+def test_core_font_arguments_are_optional(font: ImageFont.FreeTypeFont) -> None:
+    # Every argument after the leading string is optional at the C level,
+    # and the mode argument additionally accepts None.
+    assert font.font.getlength("A") == font.font.getlength("A", "", None, None, None)
+    assert font.font.getlength("A", None) == font.font.getlength("A", "")
+    assert font.font.getsize("A") == font.font.getsize("A", "", None, None, None, None)
+    assert font.font.getsize("A", None) == font.font.getsize("A", "")
```
radarhere: I don't think we need a test for this. While there are some exceptions (mostly coverage), I think Pillow has a Python API, and the test suite should target that.

akx: Removed.

src/PIL/_imagingft.pyi
```diff
@@ -24,36 +24,36 @@ class Font:
...
-        start: tuple[float, float],
+        mode: str | None = ...,
+        dir: str | None = ...,
+        features: list[str] | None = ...,
+        lang: str | None = ...,
+        stroke_width: float = ...,
+        stroke_filled: bool = ...,
+        anchor: str | None = ...,
+        foreground_ink_long: int = ...,
+        start: tuple[float, float] = ...,
```
radarhere:
```suggestion
        mode: str | None = None,
        dir: str | None = None,
        features: list[str] | None = None,
        lang: str | None = None,
        stroke_width: float = 0,
        stroke_filled: bool = 0,
        anchor: str | None = None,
        foreground_ink_long: int = 0,
        start: tuple[float, float] = (0, 0),
```

src/PIL/_imagingft.pyi
```diff
@@ -24,36 +24,36 @@ class Font:
...
-        dir: str | None,
-        features: list[str] | None,
-        lang: str | None,
-        anchor: str | None,
+        string: str | bytes,
+        mode: str | None = ...,
+        dir: str | None = ...,
+        features: list[str] | None = ...,
+        lang: str | None = ...,
+        anchor: str | None = ...,
```
radarhere:
```suggestion
        mode: str | None = None,
        dir: str | None = None,
        features: list[str] | None = None,
        lang: str | None = None,
        anchor: str | None = None,
```

src/PIL/_imagingft.pyi
```diff
@@ -24,36 +24,36 @@ class Font:
...
         self,
         string: str | bytes,
-        mode: str,
-        dir: str | None,
-        features: list[str] | None,
-        lang: str | None,
+        mode: str | None = ...,
+        dir: str | None = ...,
+        features: list[str] | None = ...,
+        lang: str | None = ...,
```
radarhere:
```suggestion
        mode: str | None = None,
        dir: str | None = None,
        features: list[str] | None = None,
        lang: str | None = None,
```

radarhere:
> * changes the core `getfont()`'s signature to make all arguments required ([C interface is internal #9879](https://github.com/python-pillow/Pillow/pull/9879) in mind, that should be fine to do), since calling `getfont()` without encoding could handily segfault (since `encoding` in C was left uninitialized, set to stack garbage).

You've initialized `encoding` to `NULL` in this PR. I would think that should be enough. The only time in main it is used is
https://github.com/python-pillow/Pillow/blob/0253ef07e697122e9978c1abc28e091f9dab541e/src/_imagingft.c#L227-L231

But you are ensuring that it is always populated from Python bcause you are concerned about someone writing future code that isn't aware that it might be `NULL`? I don't think I share that concern.

akx:
> You've initialized `encoding` to `NULL` in this PR. I would think that should be enough.

I can re-add the `|` into the parse string if you like so those args aren't necessary to pass, but given we've established these are internal (or, in [python-cryptography terms, "hazardous materials"](https://cryptography.io/en/latest/hazmat/primitives/)) APIs, if Pillow ever wants to add more args here, a single calling convention will break hazmat-calling external users break more obviously. If... if that makes sense?

radarhere:
> a single calling convention will break hazmat-calling external users break more obviously

I don't think we need to take action to help such users. But I also have a feeling that there are a bunch of arguments in our C API that are optional that don't need to be (since we always provide values for them in our Python code). See what you think of https://github.com/akx/Pillow/pull/30

akx:
Test coverage is only missing error returns from C code, should be fine.

akx:
Merged `main` in since #10047 also changed `_imagingft.c`...

PR #9909, opened 2026-08-26
Improve text rendering

akx:
> [!NOTE]
> Stacked on #9908 (I guess GitHub's merge stacks don't work cross-fork, ah well).
> That PR ought to be reviewed and merged first and I'll happily rebase this.
> 24a6fa08beeeb00577155a9cc719b6a08a25c961 is the first new commit here.

Refs #9898 (not "fixes", since I can't be sure this ends up fixing OP's issues).

## What's moving?

Some infrastructure:

* `PILLOW_TEST_SNAPSHOTS` can now be set to `create`, `overwrite` or `update`, so `assert_image_similar_tofile` and `assert_image_equal_tofile` can update the golden image to what was passed in; this is similar to `--snapshot-update` for Syrupy or `INSTA_UPDATE` for Rust's Insta.
* A test based on #9898's example code, golden snapshots updated as they improve over the patches. 

These are the actual fixes; see #9908 for notes on the clean-ups:

* Passing in a bytestring where some characters had the high bit set would render mojibake/tofu/.NOTDEF (whatever you want to call it) when using the basic renderer. Very corner-case, but also the fix is just a proper `(unsigned char)` cast.
* Fractional variable font axis values were truncated; no more!
* When using basic font layout (which is the case for the default font added in https://github.com/python-pillow/Pillow/pull/7354), advance values were 64× off due to type confusion (see #9908's description). No more!
* Finally (the last 2/3 commits), glyph positioning within a single call is now subpixel and not pixel-snap-hinted (when available), leading to more accurate rendering.


Tests/test_imagefont.py
```diff
@@ -571,6 +589,24 @@ def test_getbbox_empty(font: ImageFont.FreeTypeFont) -> None:
     assert (0, 0, 0, 0) == font.getbbox("")
 
 
+def test_bytearray_not_supported(font: ImageFont.FreeTypeFont) -> None:
+    # The C text layout only accepts str and bytes, so bytearray is rejected
+    # even though these methods used to be annotated as taking one.
+    for op in (font.getlength, font.getbbox, font.getmask):
+        assert op(b"A") is not None
+        with pytest.raises(TypeError, match="expected string or bytes"):
+            op(bytearray(b"A"))  # type: ignore[arg-type]
```
radarhere: I don't think that past functionality of the code need to be permanently documented in the form of tests. `test_bytes()` already tests for this `TypeError`.

So we no longer support bytearrays. There's a lot of other things we don't support as well.

docs/releasenotes/13.0.0.rst
```diff
@@ -116,9 +116,48 @@ Two new filters are available for :py:meth:`~PIL.Image.Image.resize` and
 :py:meth:`~PIL.Image.Image.thumbnail`: ``Image.Resampling.MKS2013`` and
 ``Image.Resampling.MKS2021``. These are versions of the Magic Kernel Sharp filter.
 
+Bugfixes
+========
```
radarhere: Would you like to elaborate on why this couldn't have lived under 'Other changes'?

akx: It definitely should live there.

akx:
I'll rebase this again after #9908 gets merged since this builds on it 👍

akx:
Rebased post 175997a2198be26c5479a6ec5791e14020fdc30d on a locally rebased version of ft-clarify... I didn't touch ft-clarify on GitHub though, so it can still be separately reviewed and merged.

PR #9910, opened 2026-08-27
Remove unused C convert method

radarhere:
Part of https://github.com/python-pillow/Pillow/pull/8389

### DIFF
--- a/src/PIL/ImageTk.py
+++ b/src/PIL/ImageTk.py
@@ -183,7 +183,7 @@ class PhotoImage:
         image = im.im
         if not image.isblock() or im.mode != self.__mode:
             block = Image.core.new_block(self.__mode, im.size)
-            image.convert2(block, image)  # convert directly between buffers
+            image.convert2(block)  # convert directly between buffers
             ptr = block.ptr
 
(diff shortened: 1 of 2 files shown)


akx:
I was looking at this earlier too. Should we take the opportunity to also rename the functions/methods to something a little more descriptive?

radarhere:
You'd like to rename `convert2`? Ok, to what? `convert_to_image`?

Or, if you want, we could suggest combining it with `new_block`, and go for a single, clear `convert_to_block`.

akx:
> You'd like to rename `convert2`? 

It isn't a great name, is it? 😅 

> Ok, to what? `convert_to_image`?

`convert_into` would be closer, maybe, since AIUI it converts `self` into the format of the caller-allocated `im`?

> Or, if you want, we could suggest combining it with `new_block`, and go for a single, clear `convert_to_block`.

I'd maybe skip that, because you could maybe want to use `convert_into` with a `new_arrow` allocated image too? 

EDIT: I had to look up what `new_block` does: it allocates an image in a single contiguous block (because that's what Tk needs).

radarhere:
I've created https://github.com/python-pillow/Pillow/pull/9911

PR #9911, opened 2026-08-27
Rename C convert2 to convert_into

radarhere:
Suggested by https://github.com/python-pillow/Pillow/pull/9910#issuecomment-5437165138

### DIFF
--- a/src/libImaging/Convert.c
+++ b/src/libImaging/Convert.c
@@ -1575,8 +1575,26 @@ static struct {
     {IMAGING_MODE_I_16B, IMAGING_MODE_F, I16B_F}
 };
 
-static Imaging
-convert(Imaging imOut, Imaging imIn, ModeID mode, ImagingPalette palette, int dither) {
+/**
+ * Convert imIn to `mode`.
+ * If imIn is already in `mode`, this performs a copy into imOut
+ * (or a newly allocated image if imOut is NULL).
+ *
+ * @param imOut   Existing image to write into
+ *                (must already be in `mode` and the same size as imIn),
+ *                or NULL to allocate a new image for the result.
+ * @param imIn    Source image to convert.
+ * @param mode    Target mode.
+ * @param palette Target palette for conversions to "P"/"PA";
+ *                NULL to use/generate a default palette.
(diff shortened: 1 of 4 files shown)


src/libImaging/Convert.c
```diff
@@ -1645,13 +1645,10 @@ convert(Imaging imOut, Imaging imIn, ModeID mode, ImagingPalette palette, int di
 }
 
 Imaging
```
akx: May I suggest adding a doc comment?


```suggestion
/**
 * Convert imIn to `mode`.
 * If imIn is already in `mode`, this performs a copy into imOut
 * (or a newly allocated image if imOut is NULL).
 *
 * @param imOut   Existing image to write into
 *                (must already be in `mode` and the same size as imIn),
 *                or NULL to allocate a new image for the result.
 * @param imIn    Source image to convert.
 * @param mode    Target mode.
 * @param palette Target palette for conversions to "P"/"PA";
 *                NULL to use/generate a default palette.
 * @param dither  Nonzero to dither when converting to "P"/"PA" or "1".
 * @return        The resulting Imaging object,
 *                or NULL with a Python exception set on failure.
 */
Imaging
```

akx (approved):
LGTM! (I was about to say the static `convert` is also superfluous now 😄)

src/libImaging/Convert.c
```diff
@@ -1575,8 +1575,26 @@ static struct {
...
+ * (or a newly allocated image if imOut is NULL).
+ *
+ * @param imOut   Existing image to write into
+ *                (must already be in `mode` and the same size as imIn),
+ *                or NULL to allocate a new image for the result.
+ * @param imIn    Source image to convert.
+ * @param mode    Target mode.
+ * @param palette Target palette for conversions to "P"/"PA";
+ *                NULL to use/generate a default palette.
+ * @param dither  Nonzero to dither when converting to "P"/"PA" or "1".
```
radarhere:
```suggestion
 * @param palette Target palette for conversions to "P" or "PA";
 *                NULL to use a default palette.
 * @param dither  Nonzero to dither when converting to "P", "PA" or "1".
```

PR #9912, opened 2026-08-27
Import external headers with angle brackets

radarhere:
Part of https://github.com/python-pillow/Pillow/pull/8389


akx (approved):
Cohesive! 😄

